PR #12249, opened 2025-02-03
feat: Add DecodedVector::sharedBase()

Yuhta:
Summary: Sometimes we need to take shared ownership of the base value vector of a dictionary.  The current `DecodedVector` only keeps reference to a raw pointer so there is no way to get hold of the `shared_ptr`.  We add `DecodedVector::sharedBase()` and overload of `DecodedVector::decode` to take `shared_ptr` so that we can get the shared ownership.

Differential Revision: D69081492

### DIFF
--- a/velox/vector/DecodedVector.cpp
+++ b/velox/vector/DecodedVector.cpp
@@ -24,6 +24,7 @@ namespace facebook::velox {
 uint64_t DecodedVector::constantNullMask_{0};
 
 namespace {
+
 std::vector<vector_size_t> makeConsecutiveIndices(size_t size) {
   std::vector<vector_size_t> consecutiveIndices(size);
   for (vector_size_t i = 0; i < consecutiveIndices.size(); ++i) {
@@ -31,6 +32,23 @@ std::vector<vector_size_t> makeConsecutiveIndices(size_t size) {
   }
   return consecutiveIndices;
 }
+
+const VectorPtr& getLoadedVector(const VectorPtr& vector) {
+  return BaseVector::loadedVectorShared(vector);
+}
+
+const BaseVector* getLoadedVector(const BaseVector* vector) {
+  return vector->loadedVector();
+}
+
+const VectorPtr& getValueVector(const VectorPtr& vector) {
+  return vector->valueVector();
+}
+
+const BaseVector* getValueVector(const BaseVector* vector) {
+  return vector->valueVector().get();
+}
+
 } // namespace
 
 const std::vector<vector_size_t>& DecodedVector::consecutiveIndices() {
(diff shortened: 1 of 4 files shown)


velox/vector/DecodedVector.cpp
```diff
@@ -155,14 +192,15 @@ void DecodedVector::combineWrappers(
   int32_t levelCounter = 0;
   for (;;) {
     if (numLevels != -1 && ++levelCounter == numLevels) {
-      baseVector_ = values;
+      baseVectorShared_ = values;
```
bikramSingh91, line 195: Its a bit unclear what the contract here is between different decode APIs. Like, will only the API that takes a VectorPtr save a reference to the base?
If yes, then there are some instances like this one where a reference is always taken. This can then affect existing use cases that currently use the BaseVector API and do not expect the base to outlive the original vector. One such case that is causing test failures is in the ExprTest suite. This can be fixed by clearing the decodedVector by calling reset() before adding it to the decodedVector cache. https://github.com/facebookincubator/velox/blob/ccb5094211888fbbd0633dad2a25db864e1c6c3d/velox/expression/EvalCtx.h#L797-L801

velox/vector/DecodedVector.h
```diff
@@ -395,6 +407,8 @@ class DecodedVector {
   // after constant and dictionary vectors have been peeled off.
   const BaseVector* baseVector_ = nullptr;
 
+  VectorPtr baseVectorShared_;
```
bikramSingh91, line 410: nit: Can you mention under what conditions this will be populated. Also, add checks in its getter method to ensure those invariants hold

velox/vector/DecodedVector.cpp
Yuhta: I am also thinking about avoid holding ownership of the smart pointer, but the cast from specific `shared_ptr` (e.g. `FlatVectorPtr`) to `VectorPtr` is annoying.  Maybe I will just keep the ownership exclusively through the new method, and avoid holding onto it in any old calls.

velox/vector/DecodedVector.cpp
```diff
@@ -374,25 +448,23 @@ BufferPtr copyNullsBuffer(
 } // namespace
 
 DecodedVector::DictionaryWrapping DecodedVector::dictionaryWrapping(
-    const BaseVector& wrapper,
+    memory::MemoryPool& pool,
     vector_size_t size) const {
-  VELOX_CHECK(!isIdentityMapping_);
```
bikramSingh91, line 379: do we plan to use dictionaryWrapping() when these conditions are not met?

Edit: Just noticed you added a test for these case, curious why we would want to use this API for these cases as it would end up creating inefficient dictionaries. Is it because we plan to add nulls on top of that?

Also, can you update the method comment to remove the part about `Requires
  /// isIdentityMapping() == false and isConstantMapping() == false.` if we plan to keep this?

bikramSingh91 (approved):

PR #12251, opened 2025-02-04
fix: Remove support for evaluating sub-optimial filters in joins

bikramSingh91:
Summary:
This enables the hash join operator to throw an error when a filter is applied exclusively
to columns from one side of the join. This validation ensures the detection of
suboptimal plans, where filters are evaluated at the join operator instead of being
pushed to the upstream scan node of the respective join side for optimization.

Also adds a private API to ensure an internal state variable which is a buffer is
unique before any mutations are applied to it.

Differential Revision: D69085909

### DIFF
--- a/velox/exec/HashProbe.h
+++ b/velox/exec/HashProbe.h
@@ -101,7 +101,11 @@ class HashProbe : public Operator {
   // the hash table.
   void asyncWaitForHashTable();
 
-  // Sets up 'filter_' and related members.
+  // Sets up 'filter_' and related members. Throws if filter is applied
+  // exclusively to columns from one side of the join to ensure the detection
+  // of suboptimal plans, where filters are evaluated at the join operator
+  // instead of being pushed to the upstream scan node of the respective join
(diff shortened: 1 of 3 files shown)


velox/exec/HashProbe.h
```diff
@@ -327,6 +331,15 @@ class HashProbe : public Operator {
   // of an invalidated table as we always spill the entire table.
   std::optional<RowColumn::Stats> columnStats(int32_t columnIndex) const;
 
+  // Returns the raw pointer to the output row mapping buffer.
+  // Throws exception if the buffer is not mutable.
+  // Note: Use this method close to the block that modifies the buffer to ensure
+  // mutability does not change in-between.
+  vector_size_t* getRawMutableOuputRowMapping() {
+    VELOX_CHECK(outputRowMapping_->isMutable());
```
Yuhta, line 339: This is check is failing because we keep 2 references: One in `outputRowMapping_` and one in filter input.  Maybe loose the condition a little bit?

PR #12253, opened 2025-02-04
fix: Avoid invalid access into random region in memory in OperatorUtils

yuandagits:
Summary:
There are 2 bugs in processEncodedFilterResults:

1. We may call bits::isBitSet(values, index) and then rows.isValid(i), this means we may be passing in an invalid index to `bits::isBitSet(values, index)`. We need to call `rows.isValid(i)` before `bits::isBitSet(values, index)`. Because index may give a bogus value if it is not valid 

2. We may access an invalid value in index via `auto index = indices[i];` if i is not valid.

Fix is to check if the row is valid before processing

Differential Revision: D69095863


spershin (approved):
Nice catch!

Issue #11601, opened 2024-11-20
Spark fuzzer failure: Ordering nulls is not supported


rui-mo:
### Description

Haven't reproduced this issue locally. Will try more runs.

### Error Reproduction

I1119 09:53:13.425566    40 Compression.cpp:627] Initialized zstd compressor with compression level 7
E20241119 10:01:06.825817    40 FuzzerToolkit.cpp:116] Only simplified path threw exception:
terminate called after throwing an instance of 'facebook::velox::VeloxUserError'
  what():  Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Ordering nulls is not supported
Retriable: False
Function: compareNulls
/__w/velox/velox/velox/./velox/vector/BaseVector.h
Line: 858
Stack trace:
Stack trace has been disabled. Use --velox_exception_user_stacktrace_enabled=true to enable it.

### Relevant logs

```bash
https://github.com/facebookincubator/velox/actions/runs/11909559360/job/33189964369?pr=8863
```

rui-mo:
cc @zhli1142015 If you have any clue. Thanks!

czentgr:
I saw it in my PR too: https://github.com/facebookincubator/velox/actions/runs/13370145693/job/37338218557?pr=12262

zhli1142015:
I got a repro in my local, will work on the fix, thanks.

PR #12262, opened 2025-02-05
misc: Enforce using TLS1.2 or later for tz curl client

czentgr:
The tz database is not read using curl the way that the code is currently built. However, if this were switched make sure to enforce using TLS1.2 or higher.


czentgr:
@majetideepak FYI

majetideepak (approved):

czentgr:
@pedroerp this fuzzer failure is unrelated. An issue for this exists: https://github.com/facebookincubator/velox/issues/11601 .

PR #12263, opened 2025-02-05
fix: Fix merge storageStats_  in IoStatistics

kewang1024:


velox/common/io/IoStatistics.cpp
```diff
@@ -149,7 +149,11 @@ void IoStatistics::merge(const IoStatistics& other) {
     const auto& otherStorageStats = other.storageStats();
     std::lock_guard<std::mutex> storageStatsLock(storageStatsMutex_);
     for (auto& item : otherStorageStats) {
-      storageStats_[item.first].merge(item.second);
+      if (storageStats_.count(item.first) == 0) {
```
Yuhta, line 152: nit: This can reduce the look up in hash table by 1 in case of hit:
```c++
auto it = storageStats_.find(item.first);
if (it == storageStats_.end()) {
  storageStats_.emplace(item);
} else {
  it->second.merge(item.second);
}
```

PR #12265, opened 2025-02-05
fix(build): Change to 32-core-ubuntu 

kgpai:
The PyVelox build seems to be ooming out. Temporarily increasing the machine specs while we work on reducing the build scope for PyVelox.


kevinwilfong (approved):

kKPulla (approved):

pedroerp (approved):
Thank you

PR #12268, opened 2025-02-05
feat(aggregate): Add complex type to map_union_sum

peterenescu:
Pull Request resolved: https://github.com/facebookincubator/velox/pull/12268

Adds complex type to Presto function map_union_sum. Addition required some additional surgery in order to make primitives/strings accumulator forward compatible with ComplexType accumulator, namely functions extract/addValues.

In order to have local aggregate fuzzers ran, I added support for timestamp, timezone and moved boolean to match other aggregate formalities.

### DIFF
--- a/velox/functions/prestosql/aggregates/MapUnionSumAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/MapUnionSumAggregate.cpp
@@ -13,10 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include "velox/exec/AddressableNonNullValueList.h"
 #include "velox/exec/Aggregate.h"
 #include "velox/exec/Strings.h"
 #include "velox/expression/FunctionSignature.h"
-#include "velox/functions/lib/CheckedArithmeticImpl.h"
 #include "velox/functions/prestosql/aggregates/AggregateNames.h"
 #include "velox/vector/FlatVector.h"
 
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/aggregates/MapUnionSumAggregate.cpp
```diff
@@ -396,7 +524,10 @@ void registerMapUnionSumAggregate(
         auto& mapType = argTypes[0]->asMap();
         auto keyTypeKind = mapType.keyType()->kind();
         auto valueTypeKind = mapType.valueType()->kind();
+
         switch (keyTypeKind) {
+          case TypeKind::BOOLEAN:
```
bikramSingh91, line 529: can you also mention in the description that this adds these other types too.
Also, can you add a test for boolean type? Timestamp and varbinary would be directed to int and stringView respectively so am not too worried about them but would be good to have atleast a sanity check for the boolean type

bikramSingh91: Can you also mention that you ran the aggregate fuzzer <insert val> num of hours successfully.

peterenescu: Hi Bikram, yes, aggregate fuzzer was run for hours before this release without failure. Also, re:boolean, I didn't add boolean, functionality, it was added by another diff: https://github.com/facebookincubator/velox/pull/12273 which adds a unit test. I just moved it around to follow other aggregate type orderings. Nonetheless, from my runs, timestamp, varbinary and boolean were all tested by the fuzzer with no problems

bikramSingh91: > aggregate fuzzer was run for hours before this release without failure.

Thanks for confirming, would be good to add the testing effort done in the PR description

bikramSingh91 (approved):

PR #12272, opened 2025-02-05
fix: Fix handling of user errors in json_parse

kevinwilfong:
Summary:
json_parse currently only properly propagates user errors that come from 
normalizedSizeForJsonParse or from simd when parsing the JSON.

User errors can come from elsewhere in the UDF, e.g. the generateViews logic, which are not
caught and thus fail the query even when the UDF is wrapped in a "try".

To fix this I've wrapped the bulk of the logic that gets executed on a single row in a try catch that
handles user errors, propagating them to the EvalCtx.

Differential Revision: D69211907

### DIFF
--- a/velox/functions/prestosql/JsonFunctions.cpp
+++ b/velox/functions/prestosql/JsonFunctions.cpp
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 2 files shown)


kgpai:
Make sure to rebase against latest main , so we can get full fuzzer run. Thanks !

PR #12053, opened 2025-01-10
feat: Add spatial support

wraymo:
Summary:
This draft PR introduces spatial data support (#11814) to Velox, including the following key updates:

+ Added support for geometry types within Velox.
+ Introduced the GEOS library as a dependency for spatial operations.
+ Implemented serialization/deserialization for geometry types using GEOS.
+ Added cast operators to convert between `VARCHAR` (WKT) or `VARBINARY` (WKB) and geometry types.
+ Implemented some spatial functions, including `ST_Point and `ST_Contains`.

PR #12243, opened 2025-02-03
feat: Add GEOS library as an optional dependency

wraymo:
Summary
This PR introduces GEOS as an optional dependency, marking the first step in splitting #12053 to enable geospatial support.

PR #12274, opened 2025-02-06
feat: Add Geometry Presto type

wraymo:
Summary

This PR is a split of #12053 to introduce Geometry Presto type in Velox. It focuses on the type declaration without serde code. As suggested, it uses Varbinary as the underlying type. Cast operators can be added later if needed, as Presto Java handles the conversion by functions `ST_GeometryFromText`(`ST_GeomFromBinary`), and `ST_AsText` (`ST_AsBinary`).

### DIFF
--- a/velox/docs/develop/types.rst
+++ b/velox/docs/develop/types.rst
@@ -173,6 +173,7 @@ TIMESTAMP WITH TIME ZONE  BIGINT
 UUID                      HUGEINT
 IPADDRESS                 HUGEINT
 IPPREFIX                  ROW(HUGEINT,TINYINT)
+GEOMETRY                  VARBINARY
 ========================  =====================
 
 TIMESTAMP WITH TIME ZONE represents a time point in milliseconds precision
--- a/velox/expression/tests/CustomTypeTest.cpp
+++ b/velox/expression/tests/CustomTypeTest.cpp
@@ -231,7 +231,8 @@ TEST_F(CustomTypeTest, getCustomTypeNames) {
           "IPADDRESS",
           "IPPREFIX",
           "BINGTILE",
-          "TDIGEST"}),
+          "TDIGEST",
+          "GEOMETRY"}),
       names);
 
   ASSERT_TRUE(registerCustomType(
@@ -248,7 +249,8 @@ TEST_F(CustomTypeTest, getCustomTypeNames) {
           "IPPREFIX",
           "BINGTILE",
           "FANCY_INT",
-          "TDIGEST"}),
+          "TDIGEST",
+          "GEOMETRY"}),
       names);
 
   ASSERT_TRUE(unregisterCustomType("fancy_int"));
--- a/velox/functions/prestosql/TypeOf.cpp
+++ b/velox/functions/prestosql/TypeOf.cpp
@@ -15,6 +15,7 @@
  */
 #include "velox/expression/VectorFunction.h"
 #include "velox/functions/prestosql/types/BingTileType.h"
+#include "velox/functions/prestosql/types/GeometryType.h"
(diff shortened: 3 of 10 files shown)


jagill:
This generally LGTM.  I don't know which parts of Geospatial we should restrict with VELOX_ENABLE_GEO; the definition doesn't depend on the library of course.

wraymo:
Thanks @jagill for the comment. Do you want to add `VELOX_ENABLE_GEO` to the source code after we merge the dependency PR #12243 ?

jagill:
@wraymo: Sorry for the delay here.  There's been some internal refactoring since this PR was made.  Could we update it to match the new organization?

Specifically:
1. In `functions/prestosql/types`, we have extracted out registration code out the FooType.h  into FooRegistration.{h,cpp}.  Check the current directory for examples.
2. Please perform the actual type registration in `functions/prestosql/registration/GeospatialFunctionsRegistration.cpp`. BingTiles are already there and we can just add Geometry.
3. Everywhere you add a file to CMakeLists.txt please also add it (if appropriate) to the TARGETS file in the same directory.  That will allow Buck builds to work too.

Thanks!

wraymo:
@jagill Thanks for getting back! One potential concern is the `VELOX_ENABLE_GEO` (`Enable Geospatial support`) option—do we want to use it for the Geometry type, even though it doesn’t require `geos` (the functions will require)? And if we put BingTile functions in `GeospatialFunctionsRegistration.cpp`, would it a bit confusing with this option?

Another consideration is the large number of BingTile and Geometry functions. In Java, we separated them into two files: `BingTileFunctions` and `GeoFunctions`. Would it make sense to follow the same approach here?

jagill:
Good questions.

1. Re: Should we guard it behind `VELOX_ENABLE_GEO`?  I don't have a strong opinion here, since it doesn't require the library. But some code quickly will, like the `cast` operator.  IDK if it's cleaner to just guard all of it or some of it.

2. Re: `GeospatialFunctionsRegistration`.  My inclination is to have them in one file, partly because there are BingTile functions that also require Geometry (eg `geometry_to_bing_tiles`).  So far, things like JSON/etc are in just one file, which is why I've started down that path, but I'm willing to do whatever is suggested for Velox.

Thoughts, @bikramSingh91 ?

wraymo:
I'm a bit concerned about the impact of setting `VELOX_ENABLE_GEO` to false. If we put the BingTile functions into the GeospatialFunctions, do we still want them to be compiled when geospatial support is disabled?

For types, this shouldn't be an issue since I didn't find any cast operators for the Geometry type. We can handle casting within `ST_GeometryFromText` instead.

wraymo:
@jagill Just updated the code

jagill:
This LGTM, up to the question about whether to hide it behind the flag. I'll see if I can discuss it internally tomorrow.

assignUser (approved):
CMake :+1:

wraymo:
@majetideepak Would you like to take a look?

kgpai (approved):
LGTM.

kgpai:
Can you rebase against latest main - the build failures do not look related to your changes.

wraymo:
@kgpai Thanks for the review. I created #12702 to fix the ci issue and it's not merged. Can you take a look? After that, I'll rebase the main branch

kgpai:
@wraymo The CI issue is being fixed here https://github.com/facebookincubator/velox/pull/12709 , rebase after that PR lands.

jagill:
@wraymo I chatted with the team, and they agree with you that we should have two separate GeometryFunctions and BingTileFunctions files.  Let's land this PR and #12708 then refactor.

wraymo:
@jagill Sure

PR #12277, opened 2025-02-06
feat: Remote UDF, only serialize and process selected rows

vudung45:
Summary:
- In this diff, we are adding an optimization to Remote UDF, to only serialize and process selected rows.
- This could greatly help in cases where the vector is heavily filtered.
- The previous approach could also potentially have bugs where function would throw exception if it doesn't expect to process un-selected rows (e.g null conditions, unexpected arguments, etc.)

Differential Revision: D69231717

### DIFF
--- a/velox/functions/remote/client/Remote.cpp
+++ b/velox/functions/remote/client/Remote.cpp
@@ -17,6 +17,7 @@
 #include "velox/functions/remote/client/Remote.h"
 
 #include <folly/io/async/EventBase.h>
+#include "velox/exec/OperatorUtils.h"
 #include "velox/expression/Expr.h"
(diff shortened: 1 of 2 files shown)


pedroerp:
Nice, thanks for looking into this!

Instead of wrapping the entire vector in a dictionary, a more efficient way would be to keep the vector as-is, but only serialize the rows active in the rowSet. That API doesn't do that today, but it would be better here to create a new version of rowVectorToIOBuf() in VectorStream.h that takes the rowSet and only serializes the rows alive. Inside that function, you will only have to convert the rowSet to a vector of IndexRange, and call the appropriate code, so it would actually be less code. There may be an utility already in the codebase to transform a rowSet into a vector of ranges.

PR #12281, opened 2025-02-07
fix: Wildcard support in json extract using simdjson

bikramSingh91:
Summary:
This change addresses 2 bugs in the wildcard support of json_extract:

Bug 1: Failure to iterate on values inside an object, resulting in
empty arrays.
For eg: calling JSON_EXTRACT(
   JSON '{"foo": {"a" : {"c": 1 , "d": 3} }, "b": 2}', '$.foo.a.[*]')
will return an empty array whereas we expect [1, 3]

Bug 2: Inconsistent handling of indefinite results, always returning a
JSON array even when the token is not evaluated.
or eg: JSON_EXTRACT('({"a": {"b": [123, 456]}})', '$.a.c[*]') should
return NULL instead of [] as the key 'c' does not exist.

Differential Revision: D69281233

### DIFF
--- a/velox/functions/prestosql/json/tests/SIMDJsonExtractorTest.cpp
+++ b/velox/functions/prestosql/json/tests/SIMDJsonExtractorTest.cpp
@@ -31,8 +31,12 @@ simdjson::error_code simdJsonExtract(
     const std::string& path,
     TConsumer&& consumer) {
   auto& extractor = SIMDJsonExtractor::getInstance(path);
+  bool isDefinitePath = true;
   return simdJsonExtract(
-      velox::StringView(json), extractor, std::forward<TConsumer>(consumer));
+      velox::StringView(json),
+      extractor,
+      std::forward<TConsumer>(consumer),
+      isDefinitePath);
 }
 
 class SIMDJsonExtractorTest : public testing::Test {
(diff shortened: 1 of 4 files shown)


velox/functions/prestosql/json/tests/SIMDJsonExtractorTest.cpp
```diff
@@ -177,6 +181,17 @@ TEST_F(SIMDJsonExtractorTest, generalJsonTest) {
...
+  // Wildcard over object's value elements
+  testExtract(
+      json,
+      "$.store.book[0].[*]",
+      std::vector<std::string>{
+          "\"Nigel Rees\"",
+          "\"ayings of the Century\"",
+          "\"reference\"",
+          "8.95"});
+  testExtract(json, "$.store.[*].price", std::vector<std::string>{"19.95"});
```
kevinwilfong, line 193: Can you have an empty object?  E.g. {"a": {}}

Would be good to add a test for that with a wildcard (I guess that would return an empty array?)

kevinwilfong (approved):
LGTM, a couple small suggestions

kevinwilfong:
Could you fix the title before landing, I think the w in "wildcard" needs to be upper case

PR #12283, opened 2025-02-07
misc(fuzzer): Rename ArgGenerator and argsOverrideFuncs to ArgTypesGenerator and argValuesGenerators

kagamiori:
Summary: Rename ArgGenerator and argsOverrideFuncs used in expression fuzzer to ArgTypesGenerator and argValuesGenerators to differentiate generators for types and values of arguments.

Differential Revision: D69286324


kevinwilfong (approved):
Thanks this is much clearer!

PR #12286, opened 2025-02-08
fix(fuzzer): Stabalize expression fuzzer for nightly run by removing known test failures

peterenescu:
Summary: Removes functions already noted as test failures that crash fuzzer run in order to stabalize fuzzer for nightly QA run

Differential Revision: D69337160


velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -102,6 +102,52 @@ int main(int argc, char** argv) {
       // from_unixtime can generate timestamps out of the supported range that
       // make other functions throw VeloxRuntimeErrors.
       "from_unixtime",
+      "round", // existing errors noted
+      "json_size",
```
kagamiori, line 106: Hi @peterenescu, thank you for putting together this initial skip function list. Could you also add the link of github issue in the comment for each skipped function? That would help us remember the reason for skipping them.

kagamiori (approved):
LGTM. Thank you!

PR #12287, opened 2025-02-08
fix: Add support for scalar only jsons to json_extract

bikramSingh91:
Summary:
Currently, for scalar only json like '1' or 'null', json_extract
returns NULL for any json path except the root only path '$'.
This change ensure it properly supports these jsons.

Differential Revision: D69337714


bikramSingh91:
the failed CI run is unrelated and fixed in #12297 will rebase once its merged

kevinwilfong (approved):
Thanks!

PR #12289, opened 2025-02-09
feat: Add API to return function signatures by name

mbasmanova:
Summary:
Add velox::getFunctionSignatures(name) API to get a list of function signatures by name.

Use this API to optimize lambda function resolution logic.

Differential Revision: D69368657


pedroerp (approved):
Looks good to me. Thanks @mbasmanova

PR #12296, opened 2025-02-10
feat(python): Make output-schema optional on tableWrite

pedroerp:
Summary:
output schema should be optional; by default is uses whatever schema
is produced by the upstream operator.

Differential Revision: D69406743


kostasxmeta:
LGTM. Ignoring the test timeouts as they are unrelated to this diff and the fix for those is on its way to landing.

kostasxmeta (approved):
LGTM! The failing tests are unrelated to this diff, failed were due to timeouts.

PR #12298, opened 2025-02-10
fix(unittest): Adjust assertion for random skip to be platform/architecture-dependent

anlowee:
**Summary**: Adjust assertion in ReaderTest.projectColumnsMutation based on platform. The assertion for the random skip is assuming it is using `__gnu_cxx::sfmt19937`, however if using `std::mt19937` then the result will be different and the unit test will fail due to this.


Yuhta (approved):

anlowee:
Hey @Yuhta , I failed `velox_exec_test` and it said it was timeout. Is is because of the GH workflow does not have enough compute resource? How can I rerun it?

PR #12299, opened 2025-02-10
fix(json): GetEscapedChar should ignore bad escaped chars

kgpai:
Summary: Fix JsonStringUtils::getEscapedChar to ignore bad escaping and to conform to Presto java behavior. For e.g. previously we would throw when trying to sort keys like '\\&page=20' and with this change we wont.

Differential Revision: D69419721

### DIFF
--- a/velox/functions/prestosql/tests/JsonFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/JsonFunctionsTest.cpp
@@ -374,12 +374,12 @@ TEST_F(JsonFunctionsTest, jsonParse) {
     velox::test::assertEqualVectors(expected, result);
   }
 
-  // Test try with invalid escape sequence.
+  // Test try with invalid json unicode sequences.
   {
     data = makeRowVector({makeFlatVector<StringView>({
         // The logic for sorting keys checks the validity of escape sequences
-        // and will throw a user error in this case.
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/tests/JsonFunctionsTest.cpp
```diff
@@ -374,12 +374,10 @@ TEST_F(JsonFunctionsTest, jsonParse) {
     velox::test::assertEqualVectors(expected, result);
   }
 
-  // Test try with invalid escape sequence.
```
kevinwilfong, line 377: There's still a potential exception that gets thrown at line 239 in velox/functions/prestosql/json/JsonStringUtil.cpp

Is that something Presto would ignore too?  If not could we update these tests to trigger that exception?

kgpai: That error is thrown due to invalid json escape sequences and Presto shows similar behavior. For e.g 
```
SELECT json_parse(x) from (values ('{"\\&page=20": "a", "\\&page=26": "c"}'), ('{"k\\i":"abc","k2":"xyz"}'), ('{"\/&\"\f\r\n": "a", "\/&page=26": "c"}'), ('"abc\u4FE"')) as t(x);
``` 

I modified the test case to conform to this behavior.

kevinwilfong (approved):
Thanks for updating the tests!

PR #12301, opened 2025-02-11
fix[RowContainer]: Fix bug in RowContainer::extractValuesWithNulls

spershin:
Summary:
Bug happens when:
We have a target (result) vector which has some rows and not a single null
(null buffer is not allocated).
We extract to that target vector a column from a RowContainer and that column
has null in at least one row.

Bug manifests in all rows that the target vector had before extracting our
RowContainer into it becoming nulls.

This is because the null buffer was allocated with 'setNotNull' set to false.

We fix it by supplying true when 'resultOffset' is positive (meaning that
target vector already has some rows).

Differential Revision: D69445179

### DIFF
--- a/velox/vector/BaseVector.h
+++ b/velox/vector/BaseVector.h
@@ -256,8 +256,12 @@ class BaseVector {
     return const_cast<uint64_t*>(rawNulls_);
   }
 
-  BufferPtr& mutableNulls(vector_size_t size) {
-    ensureNullsCapacity(size);
+  /// Ensures the vector has capacity for the nulls and returns the shared
+  /// pointer to the buffer containing them.
+  /// Optional parameter 'setNotNull' is passed to ensureNullsCapacity() and is
+  /// used to ensure all the rows will be 'not nulls' if set to true.
+  BufferPtr& mutableNulls(vector_size_t size, bool setNotNull = false) {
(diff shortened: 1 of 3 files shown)


velox/vector/BaseVector.h
```diff
@@ -256,8 +256,12 @@ class BaseVector {
     return const_cast<uint64_t*>(rawNulls_);
   }
 
-  BufferPtr& mutableNulls(vector_size_t size) {
-    ensureNullsCapacity(size);
+  /// Ensures the vector has capacity for the nulls and returns the shared
+  /// pointer to the buffer containing them.
+  /// Optional parameter 'setNotNull' is passed to ensureNullsCapacity() and is
+  /// used to ensure all the rows will be 'not nulls'.
+  BufferPtr& mutableNulls(vector_size_t size, bool setNotNull = false) {
```
kevinwilfong, line 263: It's oddly inconsistent that mutableRawNulls calls ensureNulls() which always sets "setNotNull" to true and mutableNulls (prior to this change) calls ensureNullsCapacity directly which defaults "setNotNull" to false.

This is probably the safest approach to addressing it, given how core this is to Vectors, but it seems like it was an accident waiting to happen.

Yuhta (approved):

PR #12305, opened 2025-02-11
fix: Regexp_split fails in empty match pattern

HolyLow:
The implementation of Re2RegexpSplit has a bug, that it might fail to split the string when the pattern is a empty string itself.

For example, in the function calling: regexp_split("abcd", "").
The expected result is {"", "a", "b", "c", "d", ""}, but the current implementation would throw error.

This testcase comes from presto https://github.com/prestodb/presto/blob/099bd42eba287b1ea25bf55404c7a18882e0f6d5/presto-main/src/test/java/com/facebook/presto/operator/scalar/AbstractTestRegexpFunctions.java#L231 

See detailed description in https://github.com/facebookincubator/velox/issues/12304.

This PR fix this bug in Re2RegexpSplit implementation.

### DIFF
--- a/velox/functions/lib/Re2Functions.h
+++ b/velox/functions/lib/Re2Functions.h
@@ -408,13 +408,14 @@ struct Re2RegexpSplit {
 
     const auto re2String = re2::StringPiece(string.data(), string.size());
 
-    size_t pos = 0;
+    size_t currPos = 0;
+    size_t lastPos = 0;
     const char* start = string.data();
 
     re2::StringPiece subMatches[1];
     while (re->Match(
         re2String,
-        pos,
+        currPos,
         string.size(),
         RE2::Anchor::UNANCHORED,
         subMatches,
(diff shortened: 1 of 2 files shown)


HolyLow:
@amitkdutta @spershin @kagamiori Any comments? Thanks.

amitkdutta (approved):
Looks good. Thanks @HolyLow

HolyLow:
@amitkdutta @mbasmanova What should I do to help merging this PR? Seems no update for a while.

Any suggestion is welcome, thanks a lot.

mbasmanova:
@HolyLow Thank you for the ping. This PR somehow slipped through the cracks. Let me try to merge.

amitkdutta:
> @HolyLow Thank you for the ping. This PR somehow slipped through the cracks. Let me try to merge.

Thanks @mbasmanova for importing it. I was supposed to land it but due to PTO/shipit plugin not working, couldn't merge it. This is a great fix. Thanks @HolyLow

mbasmanova:
@HolyLow Got "This diff is either too old, or sits on an old base revision, making it unsafe to land - try rebasing first." Would you rebase?

Yuhta:
I am not sure splitting by empty separator is a well defined behavior.  If we want to match the Presto Java behavior, let's put this behind a flag and only allow this in prestosql.

Yuhta (changes requested):
Put this behind a flag and opt-in in only prestosql

HolyLow:
@Yuhta seems there is another user case that hits the bug. I think this fix should be used by default, rather than under some flag. Maybe we could discuss this further.
cc @mbasmanova

Yuhta:
@HolyLow Velox is used cross engine and not just matching Presto, so let's confine this controversial behavior inside Presto.

mbasmanova:
Superceeded by #13644

PR #12306, opened 2025-02-11
feat: Add convenience API to evalute constant expression

mbasmanova:
Summary: Useful for constant folding.

Differential Revision: D69463896

### DIFF
--- a/velox/expression/Expr.cpp
+++ b/velox/expression/Expr.cpp
@@ -2014,4 +2014,20 @@ core::ExecCtx* SimpleExpressionEvaluator::ensureExecCtx() {
   return execCtx_.get();
 }
 
+VectorPtr evaluateConstantExpression(
+    const core::TypedExprPtr& expr,
+    memory::MemoryPool* pool) {
+  auto data = BaseVector::create<RowVector>(ROW({}), 1, pool);
+
+  auto queryCtx = velox::core::QueryCtx::create();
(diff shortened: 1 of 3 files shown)


mbasmanova:
CI failure is unrelated: #12307

velox/expression/Expr.cpp
```diff
@@ -2014,4 +2014,20 @@ core::ExecCtx* SimpleExpressionEvaluator::ensureExecCtx() {
   return execCtx_.get();
 }
 
+VectorPtr evaluateConstantExpression(
+    const core::TypedExprPtr& expr,
+    memory::MemoryPool* pool) {
+  auto data = BaseVector::create<RowVector>(ROW({}), 1, pool);
+
+  auto queryCtx = velox::core::QueryCtx::create();
```
Yuhta, line 2022: Should we consider pass in `ExpressionEvaluator` for these?  We have something similar in https://github.com/facebookincubator/velox/blob/04bfdffcf4f528f33e03fb3a74e103ffab36f222/velox/expression/ExprToSubfieldFilter.cpp#L27-L44

mbasmanova: @Yuhta This is a convenience API. Adding ExpressionEvaluator argument makes it less convenient to use since the caller needs to get ExpressionEvaluator from somewhere.

Yuhta (approved):

PR #12312, opened 2025-02-12
misc(fuzzer): Add custom data generation for JSON path

kagamiori:
Summary:
Add JsonPathGenerator that generates valid JSON paths for JSON of 
a given type. JsonPathGenerator::makeRandomVariation_ controls 
whether it generates JSON paths not supported by Velox 
JsonExtractor.

Differential Revision: D68921793

### DIFF
--- a/velox/common/fuzzer/ConstrainedGenerators.cpp
+++ b/velox/common/fuzzer/ConstrainedGenerators.cpp
@@ -78,7 +78,7 @@ variant JsonInputGenerator::generate() {
   const auto object = objectGenerator_->generate();
   const folly::dynamic jsonObject = convertVariantToDynamic(object);
   auto jsonString = folly::json::serialize(jsonObject, opts_);
-  if (makeRandomVariation_ && coinToss(rng_, 0.5)) {
+  if (makeRandomVariation_) {
     makeRandomVariation(jsonString);
   }
   return variant(jsonString);
@@ -258,8 +258,12 @@ std::unique_ptr<AbstractInputGenerator> getRandomInputGeneratorPrimitive(
   return generator;
 }
 
-std::unique_ptr<AbstractInputGenerator>
-getRandomInputGenerator(size_t seed, const TypePtr& type, double nullRatio) {
+std::unique_ptr<AbstractInputGenerator> getRandomInputGenerator(
+    size_t seed,
+    const TypePtr& type,
+    double nullRatio,
+    const std::vector<variant>& mapKeys,
+    size_t maxContainerSize) {
   std::unique_ptr<AbstractInputGenerator> generator;
   if (type->isPrimitiveType()) {
     return VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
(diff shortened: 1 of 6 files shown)


velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -375,4 +388,52 @@ class JsonInputGenerator : public AbstractInputGenerator {
...
+/// @param jsonType The type of data represented by the JSON.
+/// @param mapKeys Candidate key values of maps in the JSON. All maps in the
+/// JSON are assumed to share the same key type and candidate key values.
+/// @param maxContainerLength The maximum length of a container (array or map)
+/// in the JSON.
+/// @param makeRandomVariation If true, the generator will generate JSON path
+/// not supported by JsonExtractor.
+/// This generator generates the following JSON paths:
+///  - On root: $
+///  - On strings: .length()
```
bikramSingh91, line 402: `.length()` is not supported in velox yet, can we remove this for now?

velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -375,4 +388,52 @@ class JsonInputGenerator : public AbstractInputGenerator {
...
+///  - On arrays: .length(), [index], [], [begin:end:step]
```
bikramSingh91, line 403: - `[]` => is invalid, we can keep this in case the intention here was to deliberately have invalid path.
- `[begin:end:step]` => we currently do not support this, can we remove it?
- we can also add negative indices
- `$.1` => is also valid where `1` is the index
- we can also use the wildcard operator `*` for arrays

lets start with valid paths, then we can incrementally add invalid paths

velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -375,4 +388,52 @@ class JsonInputGenerator : public AbstractInputGenerator {
...
+///  - On objects: .length(), [key], ['key'], [], [*], .key
```
bikramSingh91, line 404: how are keys generated?
There are some symbols like `:` and `,` for this path type `[key]` which should not be valid and other symbols for `.key` which are not valid. As a follow up we can choose to include those to cover unsupported cases.

velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -375,4 +388,52 @@ class JsonInputGenerator : public AbstractInputGenerator {
...
+
+/// Generates a JSON path for JSON of a given type.
+/// @param jsonType The type of data represented by the JSON.
+/// @param mapKeys Candidate key values of maps in the JSON. All maps in the
+/// JSON are assumed to share the same key type and candidate key values.
+/// @param maxContainerLength The maximum length of a container (array or map)
+/// in the JSON.
+/// @param makeRandomVariation If true, the generator will generate JSON path
+/// not supported by JsonExtractor.
+/// This generator generates the following JSON paths:
```
bikramSingh91, line 400: the recursive operator `..` is now also supported. But we can always add it as a follow up.
Also, let add the ability to generated longer paths later like `$.key[index]..[index]`

velox/common/fuzzer/ConstrainedGenerators.cpp
```diff
@@ -222,19 +226,149 @@ getRandomInputGenerator(size_t seed, const TypePtr& type, double nullRatio) {
...
+        path += fmt::format("['{}']", selectedKey);
+        generateImpl(path, type->childAt(1));
+      } else if (makeRandomVariation_ && coinToss(rng_, 0.15)) {
+        path += "[]";
+        generateImpl(path, type->childAt(1));
+      } else if (coinToss(rng_, 0.15)) {
+        path += "[*]";
+        generateImpl(path, type->childAt(1));
+      } else if (coinToss(rng_, 0.15)) {
+        path += fmt::format(".{}", selectedKey);
```
bikramSingh91, line 364: as a follow up we can add the option of appending a key after brackets without a dot which is valid syntax but not currently supported in velox, eg,`$.[2]payload`

kagamiori: @bikramSingh91 Could you give an example of what `$.[2]payload` means? Is it that the `$` is an array of object, so `[2]` selects the second object in the array, then `payload` selects a key inside this object?

velox/common/fuzzer/ConstrainedGenerators.h
kagamiori: A list of candidate keys are passed to `JsonPathGenerator` and `JsonPathGenerator` just randomly select one key from the list. `JsonPathGenerator` can assume that the JSON-string used together with the generated JSON path also only use these candidate keys in objects.

I don't fully understand your comment about symbols like `:` and `,` in `[key]`. Do you mean `key` can contain `:` or `,` and that should be invalid?

velox/common/fuzzer/ConstrainedGenerators.h
kagamiori: Hi @bikramSingh91, what's the syntax of recursive operator supported in JsonExtractor today? I locally made the generator to generate the following paths but JsonExtractor.cpp:61 throws "Invalid JSON path" error on them:
$..[2]
$..['k2']
$..k1
$..["k1"]

velox/common/fuzzer/ConstrainedGenerators.cpp
bikramSingh91: Yes, its equivalent to `$.[2].payload`

velox/common/fuzzer/ConstrainedGenerators.h
bikramSingh91: The syntax is correct. Can you double check if you are on the latest?
My change adds a bunch of tests that should fail if this one is not working for you. You can take a look at the test examples here: https://github.com/facebookincubator/velox/pull/12568

velox/common/fuzzer/ConstrainedGenerators.h
bikramSingh91: > I don't fully understand your comment about symbols like : and , in [key]. Do you mean key can contain : or , and that should be invalid?

For eg.
`$.k:y` is valid and is equivalent to `$.["k:y"]`
however,
`$.[k:y]` is not valid

bikramSingh91 (approved):

PR #12314, opened 2025-02-12
fix(python): Remove .pyi annotations

pedroerp:
Summary:
Removing .pyi annotations by properly setting interface includes and
dependencies.

Differential Revision: D69538934


kevinwilfong (approved):
Thanks!

PR #12316, opened 2025-02-13
fix: Fix operator finishTiming metrics

zhli1142015:
Issue: We record the execution time of `nextOp->noMoreInput()` to `op`'s 
finishTiming metrics. This issue is noticeable in the OrderBy, Window, and 
HashBuild operators, where computationally intensive tasks are executed within 
the `noMoreInput()` function—for example, sorting occurs in `noMoreInput()`. As 
a result, the recorded wall time is incorrectly added to upstream operators.

### DIFF
--- a/velox/exec/Driver.cpp
+++ b/velox/exec/Driver.cpp
@@ -671,7 +671,7 @@ StopReason Driver::runInternal(
               });
               if (finished) {
                 withDeltaCpuWallTimer(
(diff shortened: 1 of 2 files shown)


zhli1142015:
@pedroerp @xiaoxmeng @Yuhta could you please help review this? Thanks.

Yuhta (approved):

PR #12318, opened 2025-02-13
feat(python): Register TPC-H connector

pedroerp:
Summary:
Enable PyVelox users to register TPC-H connector. Also refactoring the
connector register/unregister methods to a separate compilation unit.

Differential Revision: D69604556


velox/py/runner/PyConnectors.cpp
```diff
@@ -0,0 +1,83 @@
...
+std::unordered_set<std::string>& connectorRegistry() {
+  static std::unordered_set<std::string> registry;
+  return registry;
+}
+
+} // namespace
+
+void registerHive(const std::string& connectorId) {
+  connector::registerConnectorFactory(
+      std::make_shared<connector::hive::HiveConnectorFactory>(
```
kKPulla, line 33: at least for these two connectors, looks like the only difference we have currently in implementation is the connector factory? If we see this will be a pattern, can we templatize this so we can reuse?

kKPulla (approved):
Lgtm, one minor comment about code duplication.

velox/py/runner/PyConnectors.cpp
pedroerp: nice!

Issue #10860, opened 2024-08-27
xsimd failures with build


aditi-pandit:
### Problem description

Build velox from main today (8/27) after running setup_macos.sh script. 

```
In file included from /Users/aditipandit/ahana_dev/velox_new/velox/velox/exec/prefixsort/tests/PrefixSortAlgorithmTest.cpp:21:
In file included from /Users/aditipandit/ahana_dev/velox_new/velox/./velox/exec/prefixsort/PrefixSortAlgorithm.h:23:
In file included from /Users/aditipandit/ahana_dev/velox_new/velox/./velox/common/base/SimdUtil.h:24:
In file included from /opt/homebrew/include/xsimd/xsimd.hpp:62:
/opt/homebrew/include/xsimd/types/xsimd_batch.hpp:115:9: error: static assertion failed due to requirement '!std::is_same<bool, bool>::value': use xsimd::batch_bool<T, A> instead of xsimd::batch<bool, A>
        static_assert(!std::is_same<T, bool>::value, "use xsimd::batch_bool<T, A> instead of xsimd::batch<bool, A>");
        ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/aditipandit/ahana_dev/velox_new/velox/./velox/vector/ConstantVector.h:463:55: note: in instantiation of template class 'xsimd::batch<bool>' requested here
  std::conditional_t<can_simd, xsimd::batch<T>, char> valueBuffer_;
                                                      ^
/Users/aditipandit/ahana_dev/velox_new/velox/./velox/vector/FlatVector-inl.h:227:23: note: in instantiation of template class 'facebook::velox::ConstantVector<bool>' requested here
    T value = constant->valueAt(0);
                      ^
/Users/aditipandit/ahana_dev/velox_new/velox/./velox/vector/FlatVector.h:256:5: note:
```

My build picked up system xsimd with following version 

```
Aditis-MacBook-Pro:velox aditipandit$ brew info xsimd
==> xsimd: stable 13.0.0 (bottled)
Modern, portable C++ wrappers for SIMD intrinsics
https://xsimd.readthedocs.io/en/latest/
Installed
/opt/homebrew/Cellar/xsimd/13.0.0 (97 files, 1.4MB) *
 Poured from bottle using the formulae.brew.sh API on 2024-08-22 at 13:04:07
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/x/xsimd.rb
[...]

PR #10670, opened 2024-08-05
Refactor setup scripts

czentgr:
Use a common file to download and install external dependencies.
Extract versions for each library.

This also addresses https://github.com/facebookincubator/velox/issues/10860
xsimd is removed from brew and instead installed using the install function. The issue is caused by xsimd being newer than works for Velox at this point.

PR #292, opened 2021-09-23
Add erase method to HashTable

oerling:
Adds erase to HashTable for use in spilling selected groups from a
group by hash table.

A deleted entry is marked by a tombstone in the tags. This counts as
occupied for probing and as empty for inserting. If a group of tags
has at least one empty, then an erase is represented by setting the
tag to empty. If all tags are non-empty, the erase is represented by
setting the tag to a tombstone. This is done so that an erased entry
will not interrupt the probing when a tag is not found in a group of
tags. Note that probing stops when there is no hit and there is at
least one free in a group of tags.

- Adds eraseRows method to RowContainer.

- Fixes offset of probedFlag bit in RowContainer.

PR #12321, opened 2025-02-13
refactor: Setup scripts

anandamideShakyan:
Continuation of : https://github.com/facebookincubator/velox/pull/10670 
Use a common file to download and install external dependencies.
Extract versions for each library.

This also addresses https://github.com/facebookincubator/velox/issues/10860
xsimd is removed from brew and instead installed using the install function. The issue is caused by xsimd being newer than works for Velox at this point.


.github/workflows/macos.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         env:
           HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK: "TRUE"
         run: |
-          source scripts/setup-macos.sh
+          scripts/setup-macos.sh
```
czentgr, line 70: This will run the full script and not just the functions below. The purpose of this was to get most dependencies from the bundle and not rely on the system installation of the dependencies. You can see the functions below only run a subset.

Please read the comment from @majetideepak in my original PR about this. We need more changes to refactor this build script or we leave the original intent and separate the bundling vs non-bundling between the versions of macOS later.

scripts/setup-common.sh
```diff
@@ -0,0 +1,287 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SCRIPTDIR=$(dirname "${BASH_SOURCE[0]}")
+source $SCRIPTDIR/setup-helper-functions.sh
+source $SCRIPTDIR/setup-versions.sh
+
+NPROC=$(getconf _NPROCESSORS_ONLN)
```
czentgr, line 20: This comes from the setup-helper-functions sourcing.

scripts/setup-common.sh
```diff
@@ -0,0 +1,287 @@
...
+}
+
+function install_fbthrift {
+  wget_and_untar https://github.com/facebook/fbthrift/archive/refs/tags/${FB_OS_VERSION}.tar.gz fbthrift
+  cmake_install_dir fbthrift -Denable_tests=OFF -DBUILD_TESTS=OFF -DBUILD_SHARED_LIBS=OFF
+}
+
+function install_duckdb {
+  if $BUILD_DUCKDB ; then
+    echo 'Building DuckDB'
```
czentgr, line 64: Not needed.

scripts/setup-common.sh
```diff
@@ -0,0 +1,287 @@
...
+
+  # install azure-storage-common
+  cmake_install sdk/storage/azure-storage-common -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF
+
+  # install azure-storage-blobs
+  cmake_install sdk/storage/azure-storage-blobs -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF
+
+  # install azure-storage-files-datalake
+  cmake_install sdk/storage/azure-storage-files-datalake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF
+}
```
czentgr, line 287: Needs newline.

scripts/setup-macos.sh
```diff
@@ -30,24 +30,18 @@ set -x # Print commands that are executed.
 
 SCRIPTDIR=$(dirname "${BASH_SOURCE[0]}")
 export INSTALL_PREFIX=${INSTALL_PREFIX:-"$(pwd)/deps-install"}
-source $SCRIPTDIR/setup-helper-functions.sh
+source $SCRIPTDIR/setup-common.sh
```
czentgr, line 33: This overrides the INSTALL_PREFIX from the previous line to possibly the default `/usr/local`.
INSTALL_PREFIX is likely something that should be set in each OS separately although the Linux ones share the same location.

scripts/setup-macos.sh
```diff
@@ -30,24 +30,18 @@ set -x # Print commands that are executed.
...
-MACOS_BUILD_DEPS="ninja cmake"
-FB_OS_VERSION="v2024.07.01.00"
-FMT_VERSION="10.1.1"
-BOOST_VERSION="boost-1.84.0"
-STEMMER_VERSION="2.2.0"
-DUCKDB_VERSION="v0.8.1"
+MACOS_VELOX_DEPS="bison flex gflags glog googletest icu4c libevent libsodium lz4 lzo openssl protobuf@21 simdjson snappy thrift xz zstd"
+MACOS_BUILD_DEPS="ninja cmake ccache"
+
+SUDO="${SUDO:-""}"
```
czentgr, line 44: Is this needed?

scripts/setup-versions.sh
```diff
@@ -0,0 +1,49 @@
...
+STEMMER_VERSION="2.2.0"
+
+# Adapter related versions.
+ABSEIL_VERSION="20240116.2"
+GRPC_VERSION="v1.48.1"
+CRC32_VERSION="1.1.2"
+NLOHMAN_JSON_VERSION="v3.11.3"
+GOOGLE_CLOUD_CPP_VERSION="v2.22.0"
+HADOOP_VERSION="2.10.1"
+AZURE_SDK_VERSION="12.8.0"
```
czentgr, line 49: new line needed

czentgr (changes requested):
Thanks for continuing the work on my original PR.
I should get co-authorship on this.

Please also note you have build failures. Please investigate.

assignUser (changes requested):
> Thanks for continuing the work on my original PR.
I should get co-authorship on this.

@anandamideShakyan  It's absolutely not ok to copy @czentgr's work without proper attribution. Please base your changes on the previous commits so authorship is properly tracked by Github. In addition, it's considered rude to takeover a PR without pinging the previous author.

czentgr:
@assignUser It is ok, we asked him to continue the work on this given it's on my plate for a while and hadn't come back to it. He is in our team and it wasn't random.
Thanks for speaking up though in case it was not proper.

@anandamideShakyan Looks like some things are still broken though.

assignUser:
Ah good to hear, thanks for the info :+1: I could have checked your profile, sorry for the gruf tone @anandamideShakyan 

I'll review this in the coming days  :)

anandamideShakyan:
@assignUser it's okay. I should have given some context about it in the description section.

scripts/setup-macos.sh
anandamideShakyan: We are using SUDO in install_boost and install_stemmer function calls.

scripts/setup-macos.sh
anandamideShakyan: Sure. I will make the changes.

anandamideShakyan:
Either both macOS builds  13 &14 or 13 or 14 are failing from time to time. I see in the logs that it starts using system installed version of fmt whereas it is specifically mentioned in the macos.yml to use the BUNDLED version for both the builds. This is a bit odd as sometimes both the builds are successful as well. 

Screenshot 2025-03-10 at 9 17 53 AM

czentgr:
@anandamideShakyan The issue that folly is using the bundled version. And when it builds it doesn't know you also want to use the bundled version of fmt to built it. You can see the problem isn't occurring when building a velox file but when building the folly bundle:
```
_deps/folly-build/CMakeFiles/folly_base.dir/folly/IPAddress.cpp.o
```
Can you check other PR logs as to why folly builds ok there? I don't think this should be a consequence of the refactoring. You removed the explicit folly bundling for macOS but it is still bundled because it is not installed into the system.
```
-- Setting folly source to AUTO
  By not providing "Findfolly.cmake" in CMAKE_MODULE_PATH this project has
-- [folly] Building Folly from source
  asked CMake to find a package configuration file provided by "folly", but
  CMake did not find one.
```
How come other macOS builds succeed? Can you please check?

anandamideShakyan:
Screenshot 2025-03-16 at 5 15 11
@czentgr I was resolving some merge conflicts but did not make any change to setup-macos.sh and pushed. This time macos13 and 14 both builds ran successfully. Not sure what is the issue here, also when I looked at the earlier logs of 13 and 14 builds I couldn't find much of  a difference between the two it was just that 14 build was using system installation of fmt at the time of build failure.

czentgr:
Please investigate the CI failures.

anandamideShakyan:
@czentgr I have resolved the CI failures, the only remaining one 'Ubuntu debug with resolve_dependency' is due to SPILL_LIMIT_EXCEEDED and not because of the code changes.

czentgr:
Thanks, I will also review again. @assignUser Please take a look again as well.

scripts/setup-ubuntu.sh
```diff
@@ -306,11 +169,62 @@ function install_cuda {
     cuda-nvrtc-dev-$dashed
 }
 
-function install_geos {
-  if [[ "$BUILD_GEOS" == "true" ]]; then
-    wget_and_untar https://github.com/libgeos/geos/archive/${GEOS_VERSION}.tar.gz geos
-    cmake_install_dir geos -DBUILD_TESTING=OFF
-  fi
+function install_thrift {
```
assignUser, line 172: this is redundant from centos

scripts/setup-centos9.sh
```diff
@@ -59,6 +46,21 @@ function install_clang15 {
   dnf_install clang15 gcc-toolset-13-libatomic-devel
 }
 
+function install_thrift {
```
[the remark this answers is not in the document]
assignUser: yeah just found a copy in ubuntu.sh

scripts/setup-macos.sh
```diff
@@ -195,6 +134,7 @@ function install_velox_deps {
   run_and_time install_wangle
```
assignUser, line 134: As mentioned in centos I think thrift and arrow should be in common and also be installed here

assignUser (commented):
Thanks for pushing this forward, looks good. 
I just have one note about thrift/arrow not being in common and missing from  macos.

scripts/setup-centos9.sh
```diff
@@ -59,6 +46,21 @@ function install_clang15 {
...
+     cd ${DEPENDENCY_DIR}/thrift
+     ./bootstrap.sh
+     EXTRA_CXXFLAGS="-O3 -fPIC"
+     # Clang will generate warnings and they need to be suppressed, otherwise the build will fail.
+     if [[ ${USE_CLANG} != "false" ]]; then
+       EXTRA_CXXFLAGS="-O3 -fPIC -Wno-inconsistent-missing-override -Wno-unused-but-set-variable"
+     fi
+     ./configure --prefix=${INSTALL_PREFIX} --enable-tests=no --enable-tutorial=no --with-boost=${INSTALL_PREFIX} CXXFLAGS="${EXTRA_CXXFLAGS}"
+     make "-j${NPROC}" install
+   )
```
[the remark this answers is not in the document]
czentgr: This is actually neat. I think I tried to use cmake originally but there was some problem. I don't recall. Great if this works.

scripts/adapters.dockerfile
```diff
@@ -15,18 +15,18 @@
...
+    cd build && \
+    source /opt/rh/gcc-toolset-12/enable && \
+    source /setup-centos9.sh && \
+    install_adapters && \
+    install_cuda 12.8; \
+}" && \
+rm -rf build && dnf remove -y conda && dnf clean all
 
 # put CUDA binaries on the PATH
 ENV PATH /usr/local/cuda/bin:${PATH}
```
czentgr, line 32: These warning are fixed by https://github.com/facebookincubator/velox/pull/12789.

scripts/setup-centos9.sh
```diff
@@ -80,7 +82,7 @@ function install_velox_deps_from_dnf {
   dnf_install libevent-devel \
     openssl-devel re2-devel libzstd-devel lz4-devel double-conversion-devel \
     libdwarf-devel elfutils-libelf-devel curl-devel libicu-devel bison flex \
-    libsodium-devel zlib-devel
+    libsodium-devel zlib-devel gtest-devel gmock-devel gflags-devel
```
czentgr, line 85: Why is gflags-devel added here? This is installed later via install function and this version is removed.
gtest and gmock are test only so that's fine here.

scripts/setup-common.sh
```diff
@@ -0,0 +1,308 @@
...
+
+function install_re2 {
+  wget_and_untar https://github.com/google/re2/archive/refs/tags/${RE2_VERSION}.tar.gz re2
+  cmake_install_dir re2 -DRE2_BUILD_TESTING=OFF
+}
+
+function install_gflags {
+  # Remove an older version if present.
+  dnf remove -y gflags
```
czentgr, line 111: This is not a common function. We should have the function in setup-centos9 and that calls the dnf and then calls the common install build function.

scripts/setup-ubuntu.sh
anandamideShakyan: Removed.

scripts/setup-macos.sh
anandamideShakyan: Ok. I will add them here. The current macos.sh script doesn't install arrow and thrift so I thought those won't be needed here.

scripts/setup-centos9.sh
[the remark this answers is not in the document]
anandamideShakyan: Done.

scripts/adapters.dockerfile
anandamideShakyan: Rebased.

scripts/setup-centos9.sh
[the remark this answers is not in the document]
anandamideShakyan: It is working for centos as well, CI checks are passing.

anandamideShakyan:
@assignUser I have made the changes as per your feedback, can you please review again? Thanks.

czentgr:
@anandamideShakyan This looks good so far. There was one test `velox_serializer_test` that did not complete. I've started a re-run. Let's check if it occurs elsewhere too or if this is only for this PR.

assignUser (commented):
While testing I ran into this line that wasn't changed in this PR but breaks when using a venv (because it runs sudos pip which is not the venv one) 
https://github.com/facebookincubator/velox/blob/a62f2940c3a9a18590b2a8e5a626eaa070020406/scripts/setup-ubuntu.sh#L91
This should just use pip3 imo.

I am investigating the test failure, as it happend again it might be related.

anandamideShakyan:
@czentgr Yes, velox_serializer_test is getting timeout. But I don't think it is related to my changes.
`403/404 Test #292: velox_serializer_test ..................................................***Timeout 1500.32 sec`

assignUser:
I build it locally and ran the serializer test standalone, it took 80 minutes! So no wonder it was timed out.

I am totally unclear how any change here could have caused it though but I don't see it happening on main, maybe try rebasing it could have been fixed already.

scripts/adapters.dockerfile
```diff
@@ -15,18 +15,18 @@
...
+COPY scripts/setup-common.sh /
+COPY scripts/setup-versions.sh /
+COPY scripts/setup-helper-functions.sh /
+RUN mkdir build && bash -c "{ \
+    cd build && \
+    source /opt/rh/gcc-toolset-12/enable && \
+    source /setup-centos9.sh && \
+    install_adapters && \
+    install_cuda 12.8; \
+}" && \
```
czentgr, line 28: Nit: indent the two lines here to indicate that this is continuing the run command.

scripts/setup-common.sh
```diff
@@ -0,0 +1,328 @@
...
+    wget_and_untar https://github.com/duckdb/duckdb/archive/refs/tags/${DUCKDB_VERSION}.tar.gz duckdb
+    cmake_install_dir duckdb -DBUILD_UNITTESTS=OFF -DENABLE_SANITIZER=OFF -DENABLE_UBSAN=OFF -DBUILD_SHELL=OFF -DEXPORT_DLL_SYMBOLS=OFF -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+  fi
+}
+
+function install_boost {
+  wget_and_untar https://github.com/boostorg/boost/releases/download/${BOOST_VERSION}/${BOOST_VERSION}.tar.gz boost
+  (
+    cd ${DEPENDENCY_DIR}/boost
+    if [[ ${USE_CLANG} != "false" ]]; then
```
czentgr, line 75: Maybe we should have a comment or check that this applies to Linux only. It doesn't apply if this is macOS but the function is now common.

scripts/setup-common.sh
```diff
@@ -0,0 +1,328 @@
...
+   wget_and_untar https://github.com/apache/thrift/archive/${THRIFT_VERSION}.tar.gz thrift
+
+   EXTRA_CXXFLAGS="-O3 -fPIC"
+   # Clang will generate warnings and they need to be suppressed, otherwise the build will fail.
+   if [[ ${USE_CLANG} != "false" ]]; then
+     EXTRA_CXXFLAGS="-O3 -fPIC -Wno-inconsistent-missing-override -Wno-unused-but-set-variable"
+   fi
+
+   CXX_FLAGS="$EXTRA_CXXFLAGS" cmake_install_dir thrift \
+     -DBUILD_SHARED_LIBS=OFF \
```
czentgr, line 169: Do we want to turn on arrow flight inclusion by default?
On the prestissimo side there is a build flag that can turn on the arrow flight connector but the arrow lib installed here doesn't support it. So a user will have to make changes to get it going. So perhaps its not too bad to turn on now?
@assignUser @majetideepak

assignUser: Is prestissimo still using our setup scripts?^^

majetideepak: Yes, Prestissimo still depends on Velox scripts.
Should we add an EXTRA_ARROW_OPTIONS flag here and enable arrow flight in Prestissimo?

scripts/adapters.dockerfile
anandamideShakyan: Done.

scripts/setup-common.sh
anandamideShakyan: Done. Added check for Linux.

scripts/setup-common.sh
anandamideShakyan: I have added EXTRA_ARROW_OPTIONS and turned on arrow flight.

scripts/setup-common.sh
```diff
@@ -0,0 +1,329 @@
...
+function install_thrift {
+   wget_and_untar https://github.com/apache/thrift/archive/${THRIFT_VERSION}.tar.gz thrift
+
+   EXTRA_CXXFLAGS="-O3 -fPIC"
+   # Clang will generate warnings and they need to be suppressed, otherwise the build will fail.
+   if [[ ${USE_CLANG} != "false" ]]; then
+     EXTRA_CXXFLAGS="-O3 -fPIC -Wno-inconsistent-missing-override -Wno-unused-but-set-variable"
+   fi
+
+   CXX_FLAGS="$EXTRA_CXXFLAGS" EXTRA_ARROW_OPTIONS="-DARROW_FLIGHT=ON" \
```
czentgr, line 168: This is wrong.
Remove `EXTRA_ARROW_OPTIONS="-DARROW_FLIGHT=ON" \` and keep this what it was before.
Then add at the end
```
-DWITH_PYTHON=OFF \
-DWITH_QT5=OFF \
-DWITH_ZLIB=OFF \
${EXTRA_ARROW_OPTIONS}
```

and then at the top of the script add a definition to be read from outside or be an empty string
```
EXTRA_ARROW_OPTIONS=${EXTRA_ARROW_OPTIONS:-" "}
```

```

File: scripts/setup-common.sh (hunk 17 above)
Comment (czentgr) [reply scripts/setup-common.sh:170]: I was confused because arrow-flight is an adapter in prestissimo too but we should be able to handle it here.
And we can via the environment variable that users of velox outside of prestissimo don't have to set.

**Comment by czentgr** [comment]
@anandamideShakyan can we find out what's happenign with macOS 13? We can see it installs fmt11 into the system. But that is why fmt is bundled along with folly. So folly shouldn't see the fmt11 headers. Likewise, for current PRs this problem doesn't appear so something has changed in the refactoring.

**Comment by anandamideShakyan** [comment]
@czentgr I rebased and force pushed the macos 13 error was gone, and macos 14 is now throwing the same error that I was getting [earlier](https://github.com/facebookincubator/velox/pull/12321#issuecomment-2709359529). You can look at the latest screenshots and compare. Apart from this we are not able to install protobuf in both macos13 and 14, I am looking into why that is happening. I will give an update on that soon.

Screenshot 2025-04-24 at 11 36 03 AM
Screenshot 2025-04-24 at 11 36 09 AM

**Comment by czentgr** [comment]
Lets rebase again. The folly changes were undone.

**Comment by anandamideShakyan** [comment]
@czentgr All CI checks are passing after the rebase.

**Comment by czentgr** [comment]
@anandamideShakyan Thanks. One more thing. Did you build the centos and adapters image and then tried a built with these with a temp commit on the source of the linux CI build script?
Use the docker/podman compose command (https://github.com/facebookincubator/velox/blob/6d1820549b997013d863bd6f42313edad2786a5a/docker-compose.yml#L97) for the centos based build (centos-cpp) for the base image and then use the base image to build the adapters image (update the image used for the adapters-cpp https://github.com/facebookincubator/velox/blob/6d1820549b997013d863bd6f42313edad2786a5a/docker-compose.yml#L51 to the new centos image). Upload the adapters image to a container registry (e.g. your github account container registry or docker hub if you have an account there) and make a temp commit to this image to get a CI build from that image here https://github.com/facebookincubator/velox/blob/6d1820549b997013d863bd6f42313edad2786a5a/.github/workflows/linux-build-base.yml#L32. The CI doesn't use a new image built.

Otherwise, this looks good.
@assignUser Please take another round. If the image build is good I think this is good to go. What do you think?

**Review by assignUser — APPROVED** [review]
CI looks good with the regenerated adapters image. Nice! 

Yes, I think we should be ready to merge @czentgr  but let's do it next week so we can react if something goes haywire. That will also give me time to fully test the macos script, as our CI only partially runs those.

**Comment by assignUser** [comment]
On macos `geos` fails to build because it's missing the `cstddef` patch that is added in the bundled build and bison/flex from brew have to be added to path manually. That should either be documented or handled by the script.

**Comment by anandamideShakyan** [comment]
Hi @assignUser, I will handle these in the script where do I need to include `cstdef` ?

**Comment by assignUser** [comment]
@anandamideShakyan  @czentgr is handling this in #13354

File: .github/workflows/linux-build-base.yml [hunk 19]
```diff
@@ -29,7 +29,7 @@ jobs:
     # prevent errors when forks ff their main branch
     if: ${{ github.repository == 'facebookincubator/velox' }}
     runs-on: 8-core-ubuntu-22.04
-    container: ghcr.io/facebookincubator/velox-dev:adapters
+    container: anandamide/velox-dev:adapters
```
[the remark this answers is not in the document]
assignUser: cc @kgpai @majetideepak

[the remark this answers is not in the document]
anandamideShakyan: Sure.

majetideepak:
Sorry for breaking this PR again. We had to push an important fix. @anandamideShakyan can you please rebase?

anandamideShakyan:
@majetideepak I have rebased.

scripts/setup-versions.sh
```diff
@@ -0,0 +1,52 @@
...
+# Build dependencies versions.
+FB_OS_VERSION="v2025.04.28.00"
+FMT_VERSION="10.1.1"
+BOOST_VERSION="boost-1.84.0"
+ARROW_VERSION="15.0.0"
+DUCKDB_VERSION="v0.8.1"
+PROTOBUF_VERSION="21.8"
+MINIO_VERSION="2022-05-26T05-48-41Z"
+MINIO_BINARY_NAME="minio-2022-05-26"
+AWS_SDK_VERSION="1.11.321"
```
majetideepak, line 29: nit: above three should go to adapters list below.

majetideepak (approved):
Thanks. 2 nits. We can fix them in a follow-up PR.

anandamideShakyan:
@majetideepak I have raised a follow up PR with the changes. https://github.com/facebookincubator/velox/pull/13466

PR #12329, opened 2025-02-13
refactor(parquet): Use velox parquet writer in Metadatatest

### LINKED ISSUE #12194: Use Velox Parquet Writer in the Parquet Writer tests
### Description

The current writer tests use a copy of the Arrow Parquet Writer. We can use the Velox Parquet Writer instead and remove the Arrow Parquet Writer, for files like `velox/dwio/parquet/tests/writer/MetadataTest.cpp`

jkhaliqi:
fixes #12194


majetideepak:
@jkhaliqi Changing the Writer code is more involved. Let's table this for now. Let's continue to remove the dependency on the Arrow Parquet Reader first.

PR #12331, opened 2025-02-14
feat: Add option to  HiveDataSink to always create a file

kevinwilfong:
Summary:
This change adds an option to HiveDataSink (plumbed via HiveInsertTableHandle) to always
create a file.  This is useful for query engines that write bucketed tables by running a Task per
output bucket rather than by setting the HiveBucketProperty. In this case, even if there is no data
to process, Velox will still produce an empty file for the output bucket.

Note that this option is not possible to support when the HiveBucketProperty is set or there a
partition keys in the data, as this means that the number of files/directories respectively is
determined at runtime from the input data. Checks for this are enforced at plan generation time.

Differential Revision: D69629165

### DIFF
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -412,6 +412,13 @@ HiveDataSink::HiveDataSink(
       "Unsupported commit strategy: {}",
       commitStrategyToString(commitStrategy_));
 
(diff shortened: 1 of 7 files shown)


Yuhta (approved):
Did you verify if we write empty files in case of bucketing in Prestissimo?

pedroerp:
@kevinwilfong mentioned that in Prestissimo the empty files are created in the coordinator instead.

PR #12337, opened 2025-02-14
fix: Reduce fuzzer expression depth due to bias run failure

peterenescu:
Summary: Reduce bias run fuzzer depth due to a recent failure caused by array_top_n merge. The error is a result of byte limitation imposed by JVM that C++ doesn't have, which causes fuzzer tests to failure for nested calls.

Differential Revision: D69674802

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -992,6 +992,7 @@ jobs:
                 --stderrthreshold=2 \
                 --log_dir=/tmp/presto_only_bias_function_fuzzer_repro/logs \
                 --repro_persist_path=/tmp/presto_only_bias_function_fuzzer_repro \
+                --velox_fuzzer_max_level_of_nesting=4 \
           && echo -e "\n\nPresto Fuzzer run finished successfully."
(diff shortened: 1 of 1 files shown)


bikramSingh91 (approved):

.github/workflows/scheduled.yml
```diff
@@ -992,6 +992,7 @@ jobs:
                 --stderrthreshold=2 \
                 --log_dir=/tmp/presto_only_bias_function_fuzzer_repro/logs \
                 --repro_persist_path=/tmp/presto_only_bias_function_fuzzer_repro \
+                --velox_fuzzer_max_level_of_nesting=4 \
```
kgpai, line 995: I think it will be better if we dont change it. The failure found in array_top_n is a good thing and it will only retrigger if array_top_n is modified again. We should aim to find more such bugs.

bikramSingh91: One thing to be mindful of is that this will fail the fuzzer and prevent potential correctness issues to be tested, so if we encounter this in the future, we should encourage the author to run this fuzzer with reduced level of nesting in order to surface actual correctness bugs

Issue #12349, opened 2025-02-15
SEGV in reading parquet data if column type mismatch occurs


czentgr:
### Bug description

The issue is easily reproducible:

make a table that contains a single varchar column
```
create table test (c0 varchar) with (format='PARQUET');
insert into test values ('abc');
```

Save the resulting parquet file and delete the table.
Create a new table:
```
create table test (c0 integer) with (format='PARQUET');
```
and copy the data file into the storage path of the new table.

This simulates when data is created externally and a new table is created to process it.
If the table definition is incorrect then a SEGV can occur.

A basic read query:
```
select * from test;
```

causes the following SEGV:

```
I20250214 11:04:53.720890 67128642 TaskManager.cpp:588] Starting task 20250214_032653_00006_jukim.1.0.0.0 with 16 max drivers.
*** Aborted at 1739549093 (unix time) try "date -d @1739549093" if you are using GNU date ***
PC: @                0x0 facebook::velox::tpch::dbgen::name_bits
*** SIGSEGV (@0xc) received by PID 44213 (TID 0x16ebd7000) stack trace: ***
    @        0x187e06e04 _sigtramp
    @        0x10c929e34 facebook::velox::ByteOutputStream::appendOne<>()
    @        0x10c929e34 facebook::velox::ByteOutputStream::appendOne<>()
    @        0x10c911150 facebook::velox::serializer::presto::detail::VectorStream::appendLength()
    @        0x10c926f2c facebook::velox::serializer::presto::detail::VectorStream::appendLengths<>()
I20250214 11:04:53.750250 67134468 TaskManager.cpp:634] No more splits for 20250214_032653_00006_jukim.1.0.0.0 for node 0
    @        0x10c926bd8 facebook::velox::serializer::presto::detail::(anonymous namespace)::appendStrings()
    @        0x10c9206fc facebook::velox::serializer::presto::detail::(anonymous namespace)::serializeFlatVector<>()
    @        0x10c91e744 _ZZZN8facebook5velox10serializer6presto6detail15serializeColumnERKNSt3__110shared_ptrINS0_10BaseVectorEEERKN5folly5RangeIPKiEEPNS3_12VectorStreamERNS0_7ScratchEENK3$_0clEvENKUlvE_clEv
    @        0x10c90f564 facebook::velox::serializer::presto::detail::serializeColumn()::$_0::operator()()
    @        0x10c90f3ac facebook::velox::serializer::presto::detail::serializeColumn()
[...]

PR #5962, opened 2023-08-02
Support struct column reading with different schemas

rui-mo:
When reading struct type column, the user-specified output schema can be different with the actual data schema in Parquet. For the missing fields, `null` occupies the position. Below table summarizes the cases and the expected result.

| Parquet column schema | User-specified output schema | Result |
| ------------- | ------------- | ------------- |
| row({"a", "c"}) | row({"a", "b", "c"})  | row(a_val, null, c_val) |
| row({"a", "c"}) | row({"b"})  | row(null) |
| row({"a", "c"}) | row({"b", "d"})  | null |
| row({"a", "c"}) | row({})  | empty |

This PR supports above cases and adds unit tests for them.

PR #12350, opened 2025-02-15
fix(parquet): Avoid SEGV if table column type does not match file column type

czentgr:
If a user defines a table, for example in Hive, where the column types don’t match the file column types SEGV might occur. Specifically, the SEGV has been observed if the parquet file contains a VARCHAR column but the table defines an INTEGER column instead and the data is accessed via TableScan using a basic select * .

The resulting vector is of type string but it doesn’t match the table metadata and can cause a SEGV in the PartitionedOutput operator.

This also prevents issues and error coming from the readers when they encounter types that are not part of the switch, for example, defining a VARCHAR type column when the file column is an INTEGER.

Fixes: https://github.com/facebookincubator/velox/issues/12349

### DIFF
--- a/velox/dwio/parquet/reader/ParquetReader.cpp
+++ b/velox/dwio/parquet/reader/ParquetReader.cpp
@@ -715,62 +715,70 @@ TypePtr ReaderBase::convertType(
           schemaElement.__isset.type_length,
       "FIXED_LEN_BYTE_ARRAY requires length to be set");
 
+  static std::string_view kTypeMappingErrorFmtStr =
+      "Converted type {} is not allowed for requested type {}";
   if (schemaElement.__isset.converted_type) {
     switch (schemaElement.converted_type) {
       case thrift::ConvertedType::INT_8:
-        VELOX_CHECK_EQ(
-            schemaElement.type,
-            thrift::Type::INT32,
-            "INT8 converted type can only be set for value of thrift::Type::INT32");
-        return TINYINT();
-
-      case thrift::ConvertedType::INT_16:
-        VELOX_CHECK_EQ(
-            schemaElement.type,
-            thrift::Type::INT32,
-            "INT16 converted type can only be set for value of thrift::Type::INT32");
-        return SMALLINT();
-
-      case thrift::ConvertedType::INT_32:
-        VELOX_CHECK_EQ(
-            schemaElement.type,
-            thrift::Type::INT32,
-            "INT32 converted type can only be set for value of thrift::Type::INT32");
-        return INTEGER();
-
-      case thrift::ConvertedType::INT_64:
-        VELOX_CHECK_EQ(
-            schemaElement.type,
-            thrift::Type::INT64,
-            "INT64 converted type can only be set for value of thrift::Type::INT64");
-        return BIGINT();
-
       case thrift::ConvertedType::UINT_8:
         VELOX_CHECK_EQ(
             schemaElement.type,
             thrift::Type::INT32,
-            "UINT_8 converted type can only be set for value of thrift::Type::INT32");
+            "{} converted type can only be set for value of thrift::Type::INT32",
+            schemaElement.converted_type);
+        VELOX_CHECK(
+            !requestedType || requestedType->kind() == TypeKind::TINYINT ||
+                requestedType->kind() == TypeKind::SMALLINT ||
+                requestedType->kind() == TypeKind::INTEGER ||
+                requestedType->kind() == TypeKind::BIGINT,
+            kTypeMappingErrorFmtStr,
+            "TINYINT",
+            requestedType->toString());
         return TINYINT();
 
+      case thrift::ConvertedType::INT_16:
       case thrift::ConvertedType::UINT_16:
         VELOX_CHECK_EQ(
             schemaElement.type,
             thrift::Type::INT32,
-            "UINT_16 converted type can only be set for value of thrift::Type::INT32");
+            "{} converted type can only be set for value of thrift::Type::INT32",
+            schemaElement.converted_type);
+        VELOX_CHECK(
+            !requestedType || requestedType->kind() == TypeKind::SMALLINT ||
+                requestedType->kind() == TypeKind::INTEGER ||
+                requestedType->kind() == TypeKind::BIGINT,
+            kTypeMappingErrorFmtStr,
+            "SMALLINT",
+            requestedType->toString());
         return SMALLINT();
 
+      case thrift::ConvertedType::INT_32:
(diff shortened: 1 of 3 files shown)


czentgr:
New E2E test output instead of SEGV:

```
presto:default> select var from segv_int_var;

Query 20250215_035213_00000_xqdxq, FAILED, 1 node
Splits: 2 total, 0 done (0.00%)
[Latency: client-side: 0:01, server-side: 0:01] [0 rows, 0B] [0 rows/s, 0B/s]

Query 20250215_035213_00000_xqdxq failed: veloxType->equivalent(*requestedType) Converted type VARCHAR does not match the requested type INTEGER Split [Hive: file:/Users/czentgr/presto/data/hive_data/default/segv_int_var/20250214_032315_00002_jukim.1.0.0.0_0_5_3fb8b160-a7fe-4c52-855a-1c77c857333c.parquet 0 - 354] Task 20250215_035213_00000_xqdxq.1.0.0.0 Operator: TableScan[0] 0
```

majetideepak (approved):
Thanks, @czentgr

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -644,6 +644,15 @@ std::unique_ptr<ParquetTypeWithId> ReaderBase::getParquetColumnInfo(
   } else { // leaf node
     name = columnNames.at(curSchemaIdx);
     const auto veloxType = convertType(schemaElement, requestedType);
+    // Check if the requestedType matches the veloxType created from the schema
+    // if provided.
+    if (requestedType) {
+      VELOX_CHECK(
+          veloxType->equivalent(*requestedType),
```
Yuhta, line 651: This does not need to be exact match, some schema evolution should be supported (e.g. INTEGER->BIGINT).  You should add individual checks inside `convertType` for each converted type.

majetideepak: This makes sense! We should confirm what schema evolution is currently supported.

Yuhta: At file level we support these:
* Struct field rename
* Additional field at end of struct
* Type widening: all integer types can be widen; REAL can be widen to DOUBLE; there is no conversion between floating point types and integer types

Check out `TableEvolutionFuzzer` to see some examples (ideally we want enable it for Parquet as well): https://github.com/facebookincubator/velox/blob/main/velox/exec/tests/TableEvolutionFuzzerTest.cpp#L140

majetideepak: Struct field rename PR is here https://github.com/facebookincubator/velox/pull/5962
We need an overall design for schema evolution.

In this PR, we should at least throw a reasonable unsupported error instead of SEGV.

> You should add individual checks inside convertType() for each converted type.

Let's do this as a starting point. We can error out for unsupported schema evolution.

pedroerp:
@czentgr @majetideepak any progress on this one?

majetideepak:
@pedroerp we will prioritize this. Should have a PR next week.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -715,69 +715,90 @@ TypePtr ReaderBase::convertType(
...
-            "INT64 converted type can only be set for value of thrift::Type::INT64");
-        return BIGINT();
-
       case thrift::ConvertedType::UINT_8:
         VELOX_CHECK_EQ(
             schemaElement.type,
             thrift::Type::INT32,
-            "UINT_8 converted type can only be set for value of thrift::Type::INT32");
+            "{} converted type can only be set for value of thrift::Type::INT32",
+            schemaElement.converted_type == thrift::ConvertedType::INT_8
```
majetideepak, line 728: Simply `schemaElement.converted_type` should do here. I see that there is a fmt formatter for `facebook::velox::parquet::thrift::ConvertedType::type` inside ParquetThriftTypes.h. Or you can use `to_string(schemaElement.converted_type)`

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -786,19 +807,49 @@ TypePtr ReaderBase::convertType(
...
             "DECIMAL requires a length and scale specifier!");
-        return DECIMAL(schemaElement.precision, schemaElement.scale);
+        if (requestedType) {
+          VELOX_CHECK(
+              requestedType->isDecimal(),
+              kTypeMappingErrorFmtStr,
+              "DECIMAL",
+              requestedType->toString());
+          if (requestedType->isShortDecimal()) {
+            const auto& shortDecimalType = requestedType->asShortDecimal();
```
majetideepak, line 830: Does it work if we want to read a short decimal in Parquet to a long decimal in Velox?

czentgr: Good point. No it doesn't. I added a restriction so short only allows to short decimal and long to long. We can only adjust the precision and not the scale which must be the same.

I'm adding a test to that effect in the table scan test. In the test to read short into a long decimal we get this type of error:
```
[ RUN      ] ParquetTableScanTest.shortDecimalToLongDecimalRead
/Users/czentgr/gitspace/velox/velox/vector/tests/utils/VectorTestBase.cpp:151: Failure
Value of: expected->equalValueAt(actual.get(), i, i)
  Actual: false
Expected: true
at 0: expected 100.01, but got 1845043342252429352732.33
```
Source is DECIMAL(5,2) and target type is DECIMAL(20,2) when read from the test file.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -786,19 +799,65 @@ TypePtr ReaderBase::convertType(
...
+        // restricted.
+        auto type = DECIMAL(schemaElementPrecision, schemaElementScale);
+        if (requestedType) {
+          VELOX_CHECK(
+              requestedType->isDecimal(),
+              kTypeMappingErrorFmtStr,
+              "DECIMAL",
+              requestedType->toString());
+          // Reading short decimals with a long decimal requested type is not
+          // yet possible. To allow for correct interpretation of the values,
```
majetideepak, line 826: scale of the requestedType can be larger than the file column.
We only need to ensure the physical type byte counts are compatible.

majetideepak: We discussed this offline. The implementation ensures file and requested decimal types match.

majetideepak:
@pedroerp, @Yuhta this is ready for review.

PR #12355, opened 2025-02-17
feat: Support to specify partition key order in TableWrite operator

JkSelf:
Fix https://github.com/apache/incubator-gluten/issues/8663

The `HiveDataSink `generates partition directories based on the partition index derived from the input `RowVector`. Currently, the partition index is constructed by traversing the `RowVector `to determine if a column is partitioned. This approach can lead to a mismatch in partition key order if it differs from the order in the `RowVector`. For instance, if the input `RowVector `is` (a, b, c)` and the partition key is set as `(b, a)`, Velox will create directories as` (a={}/b={}) `based on the existing logic. This does not align with Spark's partition directory format, which would be `(b={}/a={})`. To address this, we have introduced a partitionKey parameter in the `HiveInsertTableHandle`. This allows us to generate the partition index according to the specified partitionKey order, ensuring alignment with the user's desired partition key sequence.

### DIFF
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -85,7 +85,7 @@ std::vector<column_index_t> getPartitionChannels(
 
   if (insertTableHandle->partitionKeyOrder().size() > 0) {
     auto partitionKeys = insertTableHandle->partitionKeyOrder();
-    for (auto partitionKeyName : partitionKeys) {
+    for (const auto& partitionKeyName : partitionKeys) {
       for (column_index_t j = 0; j < insertTableHandle->inputColumns().size();
            j++) {
         if (insertTableHandle->inputColumns()[j]->name() == partitionKeyName) {
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -83,8 +83,8 @@ std::vector<column_index_t> getPartitionChannels(
     const std::shared_ptr<const HiveInsertTableHandle>& insertTableHandle) {
   std::vector<column_index_t> channels;
 
-  if (insertTableHandle->partitionKeys().size() > 0) {
-    auto partitionKeys = insertTableHandle->partitionKeys();
+  if (insertTableHandle->partitionKeyOrder().size() > 0) {
+    auto partitionKeys = insertTableHandle->partitionKeyOrder();
     for (auto partitionKeyName : partitionKeys) {
       for (column_index_t j = 0; j < insertTableHandle->inputColumns().size();
            j++) {
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -83,10 +83,23 @@ std::vector<column_index_t> getPartitionChannels(
     const std::shared_ptr<const HiveInsertTableHandle>& insertTableHandle) {
   std::vector<column_index_t> channels;
 
-  for (column_index_t i = 0; i < insertTableHandle->inputColumns().size();
-       i++) {
-    if (insertTableHandle->inputColumns()[i]->isPartitionKey()) {
-      channels.push_back(i);
+  if (insertTableHandle->partitionKeys().size() > 0) {
+    auto partitionKeys = insertTableHandle->partitionKeys();
+    for (auto partitionKeyName : partitionKeys) {
+      for (column_index_t j = 0; j < insertTableHandle->inputColumns().size();
+           j++) {
+        if (insertTableHandle->inputColumns()[j]->name() == partitionKeyName) {
+          channels.push_back(j);
+          break;
+        }
+      }
+    }
+  } else {
+    for (column_index_t i = 0; i < insertTableHandle->inputColumns().size();
+         i++) {
+      if (insertTableHandle->inputColumns()[i]->isPartitionKey()) {
+        channels.push_back(i);
+      }
     }
   }
 
--- a/velox/connectors/hive/HiveDataSink.h
+++ b/velox/connectors/hive/HiveDataSink.h
@@ -246,13 +246,11 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
           partitionKeyOrder_.begin(), partitionKeyOrder_.end());
       for (const auto& inputColumn : inputColumns_) {
         if (inputColumn->isPartitionKey()) {
-          VELOX_CHECK(partitionKeyNames.erase(inputColumn->name()) == 1);
+          VELOX_CHECK(
+              partitionKeyNames.erase(inputColumn->name()) == 1,
+              "partitionKeyOrder does not contain the partition key.");
         }
       }
-
-      VELOX_CHECK(
-          partitionKeyNames.empty(),
-          "Ensure the partitionKeyOrder contains all the partition keys in inputColumns_");
     }
   }
 
--- a/velox/connectors/hive/HiveDataSink.h
+++ b/velox/connectors/hive/HiveDataSink.h
@@ -242,18 +242,17 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
     }
 
     if (partitionKeyOrder_.size() > 0) {
-      // Ensure the partitionKeyOrder contains all the partition keys in
-      // inputColumns_.
-      std::string partitionKeyNames;
+      folly::F14FastSet<std::string> partitionKeyNames(
+          partitionKeyOrder_.begin(), partitionKeyOrder_.end());
       for (const auto& inputColumn : inputColumns_) {
         if (inputColumn->isPartitionKey()) {
-          partitionKeyNames.emplace_back(inputColumn->name());
+          VELOX_CHECK(partitionKeyNames.erase(inputColumn->name()) == 1);
         }
       }
 
       VELOX_CHECK(
-          partitionKeyNames.size() == partitionKeyOrder_.size(),
-          "partition key order size is not equal with the partition column size in inputColumns_");
+          partitionKeyNames.empty(),
+          "Ensure the partitionKeyOrder contains all the partition keys in inputColumns_");
     }
   }
 
--- a/velox/connectors/hive/HiveDataSink.h
+++ b/velox/connectors/hive/HiveDataSink.h
@@ -210,7 +210,7 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
       // if there's no data. This is useful when the table is bucketed, but the
       // engine handles ensuring a 1 to 1 mapping from task to bucket.
       const bool ensureFiles = false,
-      const std::vector<std::string> partitionKeys = {})
+      const std::vector<std::string> partitionKeyOrder = {})
       : inputColumns_(std::move(inputColumns)),
         locationHandle_(std::move(locationHandle)),
         storageFormat_(storageFormat),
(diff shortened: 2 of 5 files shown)


JkSelf:
@majetideepak Can you help to review this PR? Thanks for your help.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -300,6 +306,7 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
   const std::unordered_map<std::string, std::string> serdeParameters_;
   const std::shared_ptr<dwio::common::WriterOptions> writerOptions_;
   const bool ensureFiles_;
+  const std::vector<std::string> partitionKeys_;
```
Yuhta, line 309: Where is this used?  I don't think Velox needs to be aware of these, it's not Velox who is creating the directories.

JkSelf: @Yuhta Thanks for your review. 

The `HiveDataSink `generates partition directories based on the partition index derived from the input `RowVector`. Currently, the partition index is constructed by traversing the `RowVector `to determine if a column is partitioned. This approach can lead to a mismatch in partition key order if it differs from the order in the `RowVector`. For instance, if the input `RowVector` is `(a, b, c)` and the partition key is set as (b, a), Velox will create directories as (`a={}/b={})` based on the existing logic. This does not align with Spark's partition directory format, which would be `(b={}/a={})`. To address this, we have introduced a partitionKey parameter in the `HiveInsertTableHandle`. This allows us to generate the partition index according to the specified partitionKey order, ensuring alignment with the user's desired partition key sequence.

Yuhta: I see, can we rename this method to `partitionKeyOrder()`?  And comment that if specified, we use this order, otherwise we use the column order

velox/connectors/hive/HiveDataSink.h
```diff
@@ -209,15 +209,17 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
...
       : inputColumns_(std::move(inputColumns)),
         locationHandle_(std::move(locationHandle)),
         storageFormat_(storageFormat),
         bucketProperty_(std::move(bucketProperty)),
         compressionKind_(compressionKind),
         serdeParameters_(serdeParameters),
         writerOptions_(writerOptions),
-        ensureFiles_(ensureFiles) {
+        ensureFiles_(ensureFiles),
+        partitionKeys_(partitionKeys) {
```
Yuhta, line 222: Add a validation that if this is non-empty, we make sure every partition column appears here

velox/connectors/hive/HiveDataSink.h
JkSelf: @Yuhta Yes. Updated the method name to `partitionKeyOrder()`. And also added the related comments.

velox/connectors/hive/HiveDataSink.h
JkSelf: @Yuhta Added.

JkSelf:
@Yuhta I have resolved all your comments. Can you help to review again? Thanks.

majetideepak:
@JkSelf There are build failures. Can you take a look?

velox/connectors/hive/HiveDataSink.h
```diff
@@ -238,6 +240,21 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
             "ensureFiles is not supported with partition keys in the data");
       }
     }
+
+    if (partitionKeyOrder_.size() > 0) {
+      // Ensure the partitionKeyOrder contains all the partition keys in
+      // inputColumns_.
+      std::string partitionKeyNames;
```
Yuhta, line 247: Better to also check there is no repetition.  Something like
```c++
folly::F14FastSet<std::string> partitionKeyNames(partitionKeyOrder_.begin(), partitionKeyOrder_.end());
//...
VELOX_CHECK(partitionKeyNames.erase(inputColumn->name()) == 1);
//...
VELOX_CHECK(partitionKeyNames.empty());
```

Also in Presto we have convention to put all partitioning columns at the beginning of the `RowVector` in the right order.  Is there anything in Spark preventing us from doing so?

JkSelf: @Yuhta There is no such convention in Spark to put the partition columns at the beginning of the input columns.

JkSelf:
@Yuhta @majetideepak Can you help to review again? Thanks.

JkSelf:
@Yuhta @majetideepak Can you help to review again? Thanks.

JkSelf:
@Yuhta @majetideepak Please help to review again. Thanks.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -238,6 +240,20 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
             "ensureFiles is not supported with partition keys in the data");
       }
     }
+
+    if (partitionKeyOrder_.size() > 0) {
+      folly::F14FastSet<std::string> partitionKeyNames(
+          partitionKeyOrder_.begin(), partitionKeyOrder_.end());
+      for (const auto& inputColumn : inputColumns_) {
+        if (inputColumn->isPartitionKey()) {
+          VELOX_CHECK(partitionKeyNames.erase(inputColumn->name()) == 1);
```
majetideepak, line 249: I prefer sorting and comparing vs this deleting approach. But since the number of partition columns are often small, this should be okay.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -238,6 +240,20 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
...
+        }
+      }
+
+      VELOX_CHECK(
+          partitionKeyNames.empty(),
+          "Ensure the partitionKeyOrder contains all the partition keys in inputColumns_");
```
majetideepak, line 255: `partitionKeyOrder does not contain all the partition keys` should be sufficient.

velox/exec/tests/TableWriteTest.cpp
```diff
@@ -2126,108 +2126,122 @@ TEST_P(AllTableWriterTest, commitStrategies) {
...
+            partitionName = fmt::format(
+                "p1=str_{}{}/p0={}", i, expectedCharsAfterEscape.at(i % 15), i);
+          }
+          expectedPartitionDirectories.emplace(
+              fs::path(outputDirectory->getPath()) / partitionName);
+        }
+        EXPECT_EQ(actualPartitionDirectories, expectedPartitionDirectories);
+      };
+  testSpecialPartitionName({"p0", "p1"}, {VARCHAR(), INTEGER()});
+  testSpecialPartitionName({"p1", "p0"}, {INTEGER(), VARCHAR()});
```
majetideepak, line 2244: Can we add a test for the error condition? `partitionKeyOrder does not contain all the partition keys`

JkSelf: @majetideepak Yes. Added the error condition test.

JkSelf:
@majetideepak I have resolved all your comments. Can you help to review again? Thanks.

velox/connectors/hive/HiveDataSink.cpp
```diff
@@ -83,10 +83,23 @@ std::vector<column_index_t> getPartitionChannels(
     const std::shared_ptr<const HiveInsertTableHandle>& insertTableHandle) {
   std::vector<column_index_t> channels;
 
-  for (column_index_t i = 0; i < insertTableHandle->inputColumns().size();
-       i++) {
-    if (insertTableHandle->inputColumns()[i]->isPartitionKey()) {
-      channels.push_back(i);
+  if (insertTableHandle->partitionKeyOrder().size() > 0) {
+    auto partitionKeys = insertTableHandle->partitionKeyOrder();
+    for (auto partitionKeyName : partitionKeys) {
```
majetideepak, line 88: nit: const auto

velox/connectors/hive/HiveDataSink.h
```diff
@@ -238,6 +240,18 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
...
     }
+
+    if (partitionKeyOrder_.size() > 0) {
+      folly::F14FastSet<std::string> partitionKeyNames(
+          partitionKeyOrder_.begin(), partitionKeyOrder_.end());
+      for (const auto& inputColumn : inputColumns_) {
+        if (inputColumn->isPartitionKey()) {
+          VELOX_CHECK(
+              partitionKeyNames.erase(inputColumn->name()) == 1,
+              "partitionKeyOrder does not contain the partition key.");
```
majetideepak, line 251: nit: remove full-stop after `key`.

velox/exec/tests/TableWriteTest.cpp
```diff
@@ -2126,108 +2140,154 @@ TEST_P(AllTableWriterTest, commitStrategies) {
...
-  for (auto i = 0; i < numPartitions; ++i) {
-    // url encoded
-    auto partitionName = fmt::format(
-        "p0={}/p1=str_{}{}", i, i, expectedCharsAfterEscape.at(i % 15));
-    expectedPartitionDirectories.emplace(
-        fs::path(outputDirectory->getPath()) / partitionName);
-  }
-  EXPECT_EQ(actualPartitionDirectories, expectedPartitionDirectories);
+        if (partitionKeys.size() != partitionKeyOrder.size()) {
+          VELOX_ASSERT_THROW(
```
majetideepak, line 2214: We can test this outside the function and remove the if-else here

`VELOX_ASSERT_THROW(testSpecialPartitionName({"p1", "p0"}, {"p1"}, {INTEGER(), VARCHAR()}), "partitionKeyOrder does not contain the partition key");`

JkSelf: @majetideepak Yes. Updated.

JkSelf:
@majetideepak Can you help to review again? Thanks for your help.

majetideepak:
@JkSelf did you forget to push a commit? I do not see your changes.

JkSelf:
@majetideepak My apologies for the mistake. I have push the updates in the latest commit. Please help to review again. Thanks.

velox/exec/tests/TableWriteTest.cpp
majetideepak: I don't see this resolved.

JkSelf: @majetideepak Sorry. It seems there is something wrong with my local env. Re-submit the code. Please help to review again. Thanks.

Yuhta:
@JkSelf Looking again at this, why does the optimizer generating a RowVector that has different order on the partition keys?  Presto is able to always put the ordered partition key columns in order at the beginning of the data to the TableWrite, can Gluten also do so?

JkSelf:
> @JkSelf Looking again at this, why does the optimizer generating a RowVector that has different order on the partition keys? Presto is able to always put the ordered partition key columns in order at the beginning of the data to the TableWrite, can Gluten also do so?
@Yuhta If so, should we consider adding a project operator before the TableWrite operator to re-generate the output schema based on the partition key?

Yuhta:
@JkSelf Yes that's probably what Presto does, and it makes the code cleaner

Yuhta:
@JkSelf There will be no performance impact as it's just moving around columns and no touch to the data.  `TableWriter` must accept the columns in order of partitioning columns + file columns in order, not arbitrary order supplied by user query.  If the input is not obeying the file layout, the file written is also wrong () and we cannot read it back using table schema.  So this must be fixed at plan level.

Yuhta (changes requested):
Data columns out of order is also a problem, and we prefer fixing it in plan.

JkSelf:
@Yuhta Thank you for your detailed explanation. I will try to implement this feature in plan level. Thanks for all your review. And I will close this PR.

Issue #12358, opened 2025-02-17
Crash due to dual linking of gflags


zhztheplayer:
### Bug description

```
ERROR: something wrong with flag 'flagfile' in file '/root/velox4j/src/main/cpp/build/_deps/gflags-src/src/gflags.cc'.  One possibility: file '/root/velox4j/src/main/cpp/build/_deps/gflags-src/src/gflags.cc' is being linked both statically and dynamically into this executable.
```

This is because [this workaround](https://github.com/facebookincubator/velox/blob/730842b484e541430e3bb2962473c17c415d8072/CMake/resolve_dependency_modules/gflags.cmake#L45) is not working when `BUILD_SHARED_LIBS` is cached.

Changing the workaround to `unset(BUILD_SHARED_LIBS CACHE)` will solve the issue though it may affect cache reads on the variable afterwards.

### System information

```
Velox System Info v0.0.2
Commit: fd3ab044a600b8d0e2b5605bb4ca823968bf84a5
CMake Version: 3.28.3
System: Linux-5.4.0-200-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.2.1
C Compiler: /usr/bin/cc
C Compiler Version: 11.2.1
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib64/python3.6/site-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt
```

PR #12359, opened 2025-02-17
fix: Fix dual linking of gflags when CMake variable `BUILD_SHARED_LIBS` is cached

zhztheplayer:
This fixes https://github.com/facebookincubator/velox/issues/12358.

### DIFF
--- a/CMake/resolve_dependency_modules/gflags.cmake
+++ b/CMake/resolve_dependency_modules/gflags.cmake
@@ -51,7 +51,7 @@ endif()
 FetchContent_MakeAvailable(gflags)
 
 # Workaround for https://github.com/gflags/gflags/issues/277
-if(DEFINED CACHED_BUILD_SHARED_LIBS AND NOT DEFINED CACHE{BUILD_SHARED_LIBS})
+if(DEFINED CACHED_BUILD_SHARED_LIBS)
   set(BUILD_SHARED_LIBS
       ${CACHED_BUILD_SHARED_LIBS}
       CACHE BOOL "Restored after setting up gflags")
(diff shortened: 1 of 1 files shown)


majetideepak:
I will let @assignUser review this since he was the original author.

CMake/resolve_dependency_modules/gflags.cmake
```diff
@@ -43,7 +43,20 @@ set(GFLAGS_IS_SUBPROJECT ON)
...
 unset(BUILD_SHARED_LIBS)
+if(DEFINED CACHE{BUILD_SHARED_LIBS})
+  set(CACHED_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+  unset(BUILD_SHARED_LIBS CACHE)
+endif()
+
 FetchContent_MakeAvailable(gflags)
+
+# Workaround for https://github.com/gflags/gflags/issues/277
+if(DEFINED CACHED_BUILD_SHARED_LIBS AND NOT DEFINED CACHE{BUILD_SHARED_LIBS})
```
assignUser, line 54: Don't we want to always reset to the value in `CACHED_BUILD_SHARED_LIBS`, as that being defined means we changed the original value/unset it?

assignUser (approved):
Thanks for the improvement, a minor comment.

CMake/resolve_dependency_modules/gflags.cmake
zhztheplayer: Would you mean to remove `AND NOT DEFINED CACHE{BUILD_SHARED_LIBS}`? It's reasonable to me if so.

I was trying to consider the case that `FetchContent_MakeAvailable(gflags)` internally caches `BUILD_SHARED_LIBS` again. But apparently if that happens, we cannot change the cached value without `FORCE` so it's anyway a no-op even jumping into line 55.

assignUser: Thanks, yes that's right. We should force it then. Otherwise it could leak into the main build, unlikely but can cause fun to debug issues ^^

mbasmanova:
@zhztheplayer Got "This diff is either too old, or sits on an old base revision, making it unsafe to land - try rebasing first." Would you rebase?

zhztheplayer:
> @zhztheplayer Got "This diff is either too old, or sits on an old base revision, making it unsafe to land - try rebasing first." Would you rebase?

Thanks, rebased.

PR #12361, opened 2025-02-17
feat(types): Dedup types during serialization

mbasmanova:
Summary:
Serializing plans used in AI/ML produces very large JSONs (multiple GBs). Most of it is taken by large RowTypes serialized repeatedly. This change introduces a cache of serialized types that is used to dedup type instances during serialization. This allows to reduce serialized size 1000x from GBs to MBs.

By default, cache is disabled.

Cache is not thread safe.

To enable the cache, access thread local instance and call 'enable()' API: serializedTypeCache().enable(). Then, serialize types, expressions or plan nodes as usual. Make sure to serialize and store the cache itself using 'serializedTypeCache().serialize()' API. Clear the cache when done.

To deserialize, restore the cache using 'deserializedTypeCache().deserialize()' API. Then, deserialize types, expressions or plan nodes as usual. Clear the cache when done.

See TEST(TypeTest, serdeCache) for an example of usage.

Differential Revision: D69755598

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -160,6 +160,11 @@ std::vector<TypePtr> deserializeChildTypes(const folly::dynamic& obj) {
 } // namespace
 
 TypePtr Type::create(const folly::dynamic& obj) {
+  if (obj.find("ref") != obj.items().end()) {
+    const auto id = obj["ref"].asInt();
+    return deserializedTypeCache().get(id);
+  }
+
   std::vector<TypePtr> childTypes;
   if (obj.find("cTypes") != obj.items().end()) {
     childTypes = deserializeChildTypes(obj);
@@ -212,8 +217,7 @@ void Type::registerSerDe() {
   auto& registry = velox::DeserializationRegistryForSharedPtr();
   registry.Register(
       Type::getClassName(),
-      static_cast<std::shared_ptr<const Type> (*)(const folly::dynamic&)>(
-          Type::create));
+      static_cast<TypePtr (*)(const folly::dynamic&)>(Type::create));
 
   registry.Register("IntervalDayTimeType", IntervalDayTimeType::deserialize);
   registry.Register(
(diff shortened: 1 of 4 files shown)


velox/type/Type.cpp
```diff
@@ -519,12 +523,102 @@ std::string RowType::toString() const {
...
+folly::dynamic makeTypeRef(int32_t id) {
+  folly::dynamic ref = folly::dynamic::object;
+  ref["name"] = "Type";
+  ref["ref"] = id;
+  return ref;
+}
+} // namespace
+
 folly::dynamic RowType::serialize() const {
+  const bool useCache = size() > 10 && serializedTypeCache().isEnabled();
```
kKPulla, line 603: Can we separate this condition into some sort of Options for cache instead so that it does not hide behind?

velox/type/Type.cpp
```diff
@@ -160,6 +160,11 @@ std::vector<TypePtr> deserializeChildTypes(const folly::dynamic& obj) {
 } // namespace
 
 TypePtr Type::create(const folly::dynamic& obj) {
+  if (obj.find("ref") != obj.items().end()) {
```
kKPulla, line 163: nit: "ref" seems to be the key connecting between serialization and deserialization. Can we define a constant to reuse?

kKPulla (commented):
Not sure if we have any plan serialization tests (I did not find any based on my search) but would be nice to have one with cache enabled and disabled to ensure it works.

mbasmanova:
> Not sure if we have any plan serialization tests

@kKPulla See velox/exec/tests/PlanNodeSerdeTest.cpp

velox/type/Type.cpp
mbasmanova: Not sure I understand "it does not hide behind". Would you clarify a bit?

velox/type/Type.cpp
mbasmanova: Indeed. However, this is an existing practice where json field names are hard-coded. It might be helpful to refactor to use constants, but it is probably not a good idea to start using constants for some names, while keeping hard-coded strings for others.

velox/type/Type.cpp
kKPulla: Sorry, will clarify. I meant the condition that we only use cache when the row child count is > 10 is not immediately obvious for the user when they enable cache as it is neither explicitly provided nor visible directly in the serialized cache layer. 

My suggestion is if we can create an Options struct and pass during the SerializedTypeCache creation where this is set (like minRowChildCountForCaching?) and defaulted to 10.

velox/type/Type.cpp
kKPulla: Got it, let's keep it as is for now then.

kKPulla:
> > Not sure if we have any plan serialization tests
> 
> @kKPulla See velox/exec/tests/PlanNodeSerdeTest.cpp

Perfect, thanks for pointing me to it. I guess we can make this parameterized test w/ cache flag toggling?

mbasmanova:
@kKPulla Kk, thank you for review. Addressed your comments. Would you take another look?

kKPulla (approved):
Thanks for addressing comments. Lgtm.

velox/type/Type.cpp
```diff
@@ -519,12 +523,114 @@ std::string RowType::toString() const {
...
+  folly::dynamic values = folly::dynamic::array;
+
+  for (const auto& pair : cacheEntries) {
+    keys.push_back(pair.first);
+    values.push_back(*pair.second);
+  }
+
+  folly::dynamic cacheObj = folly::dynamic::object;
+  cacheObj["keys"] = keys;
+  cacheObj["values"] = values;
```
xiaoxmeng, line 568: Shall we clear cache_ after serialize call for reuse?

xiaoxmeng (commented):
@mbasmanova LGTM % one question: do we support to reuse the local serialize/deserialize type caches? Thanks!

velox/type/Type.cpp
```diff
@@ -519,12 +523,114 @@ std::string RowType::toString() const {
...
+  return cacheObj;
+}
+
+SerializedTypeCache& serializedTypeCache() {
+  thread_local SerializedTypeCache cache;
+  return cache;
+}
+
+void DeserializedTypeCache::deserialize(const folly::dynamic& obj) {
+  VELOX_CHECK(cache_.empty());
```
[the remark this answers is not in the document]
mbasmanova: It needs to be reset at the start and end of serialization. See PlanNodeSerdeTest.cpp for an example.

velox/type/Type.cpp
```diff
@@ -519,12 +523,114 @@ std::string RowType::toString() const {
...
+
+  const auto& keys = obj["keys"];
+  const auto size = keys.size();
+
+  const auto& values = obj["values"];
+  VELOX_CHECK_EQ(size, values.size());
+  for (auto i = 0; i < size; ++i) {
+    auto type = velox::ISerializable::deserialize<Type>(values[i]);
+    cache_.emplace(keys[i].asInt(), type);
+  }
```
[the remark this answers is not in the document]
mbasmanova: @xiaoxmeng Good suggestion. Added: [#12385](https://github.com/facebookincubator/velox/pull/12385)

PR #12362, opened 2025-02-17
fix: Optimize ContainerRowSerde deserialization for string, array and map

Yuhta:
Summary:
1. Optimize `deserializeString` in case of inlined strings
2. Optimize `deserializeArray` to avoid heap allocation for intermediate null-like data
3. Inline `FlatVector<StringView>::setNoCopy`

Differential Revision: D69750091


kevinwilfong (approved):
LGTM, thanks!

PR #12363, opened 2025-02-17
fix: Reduce lock cost in AsciiInfo updates

Yuhta:
Summary:
In some cases (e.g. row-wise deserialization) we need to resize the
string vector frequently.  In that case none of the ASCII info is valid, but we
still pay the price to lock and check the state.  Fix the cost by using an
atomic boolean for fast path check.

Differential Revision: D69750090


kevinwilfong (approved):
Thanks! LGTM

PR #12364, opened 2025-02-17
feat: Add HashStringAllocator::InputStream

Yuhta:
Summary:
When we get `ByteInputStream` from `HashStringAllocator`, we used to
have to materialize all the byte ranges in a vector, which is not efficient.
This change improve the efficiency by creating a `ByteInputStream` directly over
the linked list of a multi-part allocation.

Differential Revision: D69750088

### DIFF
--- a/velox/common/memory/HashStringAllocator.h
+++ b/velox/common/memory/HashStringAllocator.h
@@ -233,13 +233,7 @@ class HashStringAllocator : public StreamArena {
     return header->size() + kHeaderSize;
   }
 
-  /// Returns ByteInputStream over the data in the range of 'header' and
-  /// possible continuation ranges.
-  /// @param maxBytes If provided, the returned stream will cover at most that
-  /// many bytes.
-  static std::unique_ptr<ByteInputStream> prepareRead(
-      const Header* header,
-      size_t maxBytes = std::numeric_limits<size_t>::max());
+  class InputStream;
 
   /// Returns the number of payload bytes between 'header->begin()' and
   /// 'position'.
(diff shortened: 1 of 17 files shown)


velox/common/memory/HashStringAllocator.h
```diff
@@ -516,6 +510,148 @@ class HashStringAllocator : public StreamArena {
...
+      header = header->nextContinued();
+    }
+    return total;
+  }
+
+  bool atEnd() const final {
+    return range_.position == range_.size && !header_->isContinued();
+  }
+
+  std::streampos tellp() const final {
```
kevinwilfong, line 552: nit: Do we end up calling this often? It would probably be cheaper to just maintain a counter instead of counting it on the fly.

velox/common/memory/HashStringAllocator.h
```diff
@@ -516,6 +510,148 @@ class HashStringAllocator : public StreamArena {
...
+    int64_t pos = 0;
+    while (header != header_) {
+      pos += header->usableSize();
+      header = header->nextContinued();
+    }
+    return pos + range_.position;
+  }
+
+  void seekp(std::streampos pos) final {
+    header_ = begin_;
```
kevinwilfong, line 563: updates to header_ and range_ are tightly bound, everywhere you call resetRange(), the line above sets header_ to some new value.

it looks like it'd be easier/safer to make resetRange into resetHeader (or something like that)

  ```
void resetHeader(const Header* header) {
    VELOX_DCHECK_GT(header->usableSize(), 0);
    header_ = header;
    range_.buffer = reinterpret_cast<uint8_t*>(header_->begin());
    range_.size = header_->usableSize();
    range_.position = 0;
  }
```

File: velox/exec/prefixsort/PrefixSortEncoder.h [hunk 3]
```diff
@@ -294,9 +294,9 @@ FOLLY_ALWAYS_INLINE void PrefixSortEncoder::encodeNoNulls(
   } else {
     // 'data' is stored in non-contiguous allocation pieces in the row
     // container, we only read prefix size data out.
-    auto stream = HashStringAllocator::prepareRead(
+    HashStringAllocator::InputStream stream(
         HashStringAllocator::headerOf(value.data()));
-    stream->readBytes(dest, copySize);
+    stream.ByteInputStream::readBytes(dest, copySize);
```
kevinwilfong, line 299: You have to add this prefix because the InputStream class was forward declared inside the HashStringAllocator, right?

Could you inline the definition in side of HashStringAllocator, or define it outside the class in the same header? HashStringAllocatorInputStream would be just as clear as HashStringAllocator::InputStream, slightly shorter, and then we don't need all these ByteInputStream:: prefixes.

kevinwilfong (approved):

velox/common/memory/HashStringAllocator.h
Yuhta: No I don't even see this got called in real query

PR #12365, opened 2025-02-18
fix(wave): Benchmark failure

majetideepak:
Failing with `Device context not found. Inconsistent state.`

### DIFF
--- a/velox/experimental/wave/common/Cuda.cu
+++ b/velox/experimental/wave/common/Cuda.cu
@@ -87,6 +87,7 @@ Device* setDriverDevice(int32_t deviceId) {
     if (cnt == 0) {
       waveError("No Cuda devices found");
     }
+    driverInited = true;
   }
   if (deviceId >= contexts.size()) {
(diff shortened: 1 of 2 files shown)


velox/experimental/wave/common/Cuda.cu
```diff
@@ -135,15 +136,7 @@ Device* currentDevice() {
 }
 
 Device* getDevice(int32_t deviceId) {
-  Device* save = nullptr;
-  if (driverInited) {
-    save = currentDevice();
-  }
-  auto* dev = setDriverDevice(deviceId);
-  if (save) {
-    setDevice(save);
```
Yuhta, line 144: We are changing the current device to `deviceId` with this change, can we keep the logic to restore it back to it's old value?

majetideepak: Make sense! Fixed! I felt that getDevice means to set it as the current device but not always true.

Yuhta (approved):

PR #12370, opened 2025-02-18
Fix: Flakey memory pool already exists failure in TableScanReplayerTest

### LINKED ISSUE #12360: CI: TableScanReplayerTest fails with "Memory pool TableScan_replayer_0 already exists"
### Bug description

```
participants_.count(pool->name()) == 0 (1 vs. 0) Memory pool TableScan_replayer_0 already exists, Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
```
https://github.com/facebookincubator/velox/actions/runs/13345770538/job/37276039027

### System information

https://github.com/facebookincubator/velox/actions/runs/13345770538/job/37276039027

### Relevant logs

```bash

```

duanmeng:
The `OperatorReplayerBase` creates query context for each replay query,
in which it would create a memory pool, we should add a thread-safe 
auto increase ID in it as CI would run unit tests concurrently, and trigger
the flakey failure.

Resolve #12360

### DIFF
--- a/velox/tool/trace/OperatorReplayerBase.cpp
+++ b/velox/tool/trace/OperatorReplayerBase.cpp
@@ -113,8 +113,9 @@ core::PlanNodePtr OperatorReplayerBase::createPlan() {
 }
 
 std::shared_ptr<core::QueryCtx> OperatorReplayerBase::createQueryCtx() {
+  static std::atomic_uint64_t replayQueryId{0};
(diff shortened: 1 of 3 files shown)


majetideepak (approved):

xiaoxmeng (approved):
@duanmeng thanks for the fix!

PR #12380, opened 2025-02-18
feat: Remove more unnecessary PTX code from Wave

dreveman:


CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ option(
 option(VELOX_MONO_LIBRARY "Build single unified library." OFF)
 option(ENABLE_ALL_WARNINGS "Enable -Wall and -Wextra compiler warnings." ON)
 option(VELOX_BUILD_SHARED "Build Velox as shared libraries." OFF)
+option(VELOX_BUILD_BRANCH_KERNEL_TEST "Enable Wave branch kernel test." ON)
```
Yuhta, line 82: Can we define it as an opt-out instead of opt-in?  Something like `VELOX_SKIP_WAVE_BRANCH_KERNEL_TEST`

dreveman: Yes, done in latest version of PR.

Yuhta (approved):

dreveman:
@Yuhta any reason for why this is taking so long to be merged?

mbasmanova:
@dreveman Seeing format error:
Screenshot 2025-02-27 at 12 53 18 AM

@kgpai @majetideepak @assignUser Wondering why this doesn't show up in CI.

majetideepak:
@mbasmanova Likely a clang-format version difference. Can you share the version being used internally?

mbasmanova:
@majetideepak clang-format version 18.1.3

assignUser:
> clang-format version 18.1.3

That's what's in our container as well. I think the reason is that we only run the check on touched files and the file in question wasn't touched by this PR. Maybe the internal linter works on a dir basis and not a file basis or something?

mbasmanova:
@assignUser velox/experimental/wave/common/Cuda.cu is modified in this PR. Do we run format-check on .cu files?

assignUser:
Yes now it is but only to fix the lint. Or did I have a weird diff few?

dreveman:
Sorry, not sure how that format change become part of that commit. Must have run the formatting script locally with the wrong clang-format version. Rebased and removed that format change from latest version of PR.

assignUser:
I looked at the docs and tested something but was  unable to pin down the issue. My local lsp (using the same clang format version) enforces one space after the ctor initializer list and uses the same .clang-format file as `make format-fix` :shrug:

@mbasmanova maybe you can diff you internal clang-format with the public one in the repo to find the misaligned setting. But the manual fix works for now too I guess ^^

mbasmanova:
@kgpai 

> maybe you can diff you internal clang-format with the public one in the repo to find the misaligned setting.

Krishna, is this something you could help with?

PR #12381, opened 2025-02-18
feat(python): TPC-H dbgen

pedroerp:
Summary:
Support generation of TPC-H datasets using the Python API. The idea is
to use this as a pre-step that generates datasets in the target file format,
then use the generates file for benchmarks.

Differential Revision: D69809891

### DIFF
--- a/velox/py/plan_builder/PyPlanBuilder.cpp
+++ b/velox/py/plan_builder/PyPlanBuilder.cpp
@@ -18,7 +18,9 @@
 
 #include <folly/executors/GlobalExecutor.h>
 #include <pybind11/stl.h>
+#include "velox/connectors/hive/HiveConnectorSplit.h"
 #include "velox/connectors/hive/TableHandle.h"
+#include "velox/connectors/tpch/TpchConnectorSplit.h"
 #include "velox/core/PlanNode.h"
 #include "velox/dwio/dwrf/RegisterDwrfReader.h"
 #include "velox/dwio/dwrf/RegisterDwrfWriter.h"
@@ -29,6 +31,7 @@
 #include "velox/functions/prestosql/registration/RegistrationFunctions.h"
 #include "velox/parse/TypeResolver.h"
 #include "velox/py/vector/PyVector.h"
+#include "velox/tpch/gen/TpchGen.h"
 
 namespace facebook::velox::py {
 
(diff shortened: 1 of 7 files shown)


velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -198,6 +197,26 @@ class PyPlanBuilder {
       const std::string& filter,
       core::JoinType joinType);
 
+  /// Generates TPC-H data using dbgen. Note that generating data on the
+  /// fly is not terribly efficient, so one should generate TPC-H data,
+  /// write them to storage files, (Parquet, ORC, or similar), then
+  /// benchmark a query plan that reads those files.
```
kostasxmeta, line 203: It's not entirely clear from this wether this operator generates data on the fly or not (mainly because it also talks about generating files). Let's make it more clear.

velox/py/plan_builder/plan_builder.cpp
```diff
@@ -175,5 +175,29 @@ PYBIND11_MODULE(plan_builder, m) {
...
+          py::arg("table_name"),
+          py::arg("columns") = std::vector<std::string>{},
+          py::arg("scale_factor") = 1,
+          py::arg("num_parts") = 1,
+          py::arg("connector_id") = "tpch",
+          py::doc(R"(
+        Generates TPC-H data using dbgen. Note that generating data on the
+        fly is not terribly efficient, so one should generate TPC-H data,
+        write them to storage files, (Parquet, ORC, or similar), then
+        benchmark a query plan that reads those files.
```
kostasxmeta, line 191: Based on the comment above we should probably update this as well.

velox/py/runner/PyLocalRunner.cpp
```diff
@@ -95,9 +95,9 @@ py::iterator PyLocalRunner::execute() {
   }
 
   // Add any files passed by the client during plan building.
-  for (const auto& [scanId, scanPair] : *scanFiles_) {
-    for (const auto& inputFile : scanPair.second) {
-      addFileSplit(inputFile, scanId, scanPair.first);
+  for (auto& [scanId, splits] : *scanFiles_) {
+    for (auto& split : splits) {
+      cursor_->task()->addSplit(scanId, exec::Split(std::move((split))));
```
kostasxmeta, line 100:
```suggestion
      cursor_->task()->addSplit(scanId, exec::Split(std::move(split)));
```

kostasxmeta (approved):
LGTM, just a few small comments.

velox/py/plan_builder/PyPlanBuilder.cpp
```diff
@@ -201,9 +204,16 @@ PyPlanBuilder& PyPlanBuilder::tableScan(
...
       .endTableScan();
 
-  // Store the id of the scan along with any files that were passed.
-  (*scanFiles_)[planBuilder_.planNode()->id()] =
-      std::make_pair(connectorId, inputFiles.value_or(std::vector<PyFile>{}));
+  // Store the id of the scan and the respective splits.
+  std::vector<std::shared_ptr<connector::ConnectorSplit>> splits;
+  if (inputFiles.has_value()) {
+    for (const auto& inputFile : *inputFiles) {
+      splits.push_back(std::make_shared<connector::hive::HiveConnectorSplit>(
```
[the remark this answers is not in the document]
pedroerp: Good question. This may need to be better decoupled in the future, but for now we are making the same assumption that PlanBuilder does, that if you are using the tableScan() node, it implies you are using HiveConnector. 

In this PR I didn't change the assumption, just moved the code where the split was generated.

velox/py/plan_builder/PyPlanBuilder.h
pedroerp: Rephrased it.

pedroerp (commented):
@kostasxmeta thanks for the review!

PR #12383, opened 2025-02-19
misc(fuzzer): Add stats about verification against reference DB in expression fuzzer

kagamiori:
Summary: This diffs add stats reported by the expression fuzzer about how many test cases are verified against Presto and how many fail in both Velox and the reference DB.

Differential Revision: D69819014

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzerVerifier.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerVerifier.cpp
@@ -293,11 +293,13 @@ void ExpressionFuzzerVerifier::retryWithTry(
   }
 
   std::vector<ResultOrError> tryResults;
+  std::vector<test::ExpressionVerifier::VerificationState>
+      tryVerificationStates;
 
   // The function throws if anything goes wrong except
   // UNSUPPORTED_INPUT_UNCATCHABLE errors.
   try {
-    tryResults = verifier_.verify(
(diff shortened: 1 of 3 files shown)


velox/expression/fuzzer/ExpressionFuzzerVerifier.cpp
```diff
@@ -462,8 +479,13 @@ void ExpressionFuzzerVerifier::go() {
   }
   logStats();
 
-  LOG(ERROR) << "Total iterations: " << i;
-  LOG(ERROR) << "Total failed: " << numFailed;
+  LOG(ERROR) << "Total test cases: " << totalTestCases;
```
peterenescu, line 482: Should these be considered error messages even though they're just runs stats?

kagamiori: Hi @peterenescu, thank you for reviewing! This logging to stderr is actually intended, so that we can see the stats directly on screen in Github CI jobs (because we save stdout to files and only show stderr in screen for CI jobs). See the recent change in https://github.com/facebookincubator/velox/pull/12328.

peterenescu: Yea makes sense I noticed that a lot of the logs are routed to stderr, which I suppose makes sense if we wanted to make stdout cleaner!

peterenescu (approved):

PR #12384, opened 2025-02-19
fix: Fix NullHandlingMode for Spark min/max agg function

zhli1142015:
This issue occurs because `kCompareFlags` in `compare` function is initialized with `kNullAsIndeterminate`.
https://github.com/facebookincubator/velox/blob/d118b1ed4309511b2a8a7730976f55800bf8f46d/velox/functions/lib/aggregates/MinMaxAggregateBase.cpp#L610
The fuzzer detects this issue when min(string) is evaluated before min(array_with_nulls).
UT is updated to reflect this case.
Fix: #11601.

### DIFF
--- a/velox/functions/lib/aggregates/Compare.cpp
+++ b/velox/functions/lib/aggregates/Compare.cpp
@@ -22,14 +22,8 @@ int32_t compare(
     const SingleValueAccumulator* accumulator,
     const DecodedVector& decoded,
     vector_size_t index,
-    CompareFlags::NullHandlingMode nullHandlingMode) {
-  static const CompareFlags kCompareFlags{
-      true, // nullsFirst
-      true, // ascending
-      false, // equalsOnly
-      nullHandlingMode};
-
-  auto result = accumulator->compare(decoded, index, kCompareFlags);
+    CompareFlags compareFlags) {
+  auto result = accumulator->compare(decoded, index, compareFlags);
   VELOX_USER_CHECK(
       result.has_value(),
       fmt::format(
--- a/velox/functions/lib/aggregates/MinMaxAggregateBase.cpp
+++ b/velox/functions/lib/aggregates/MinMaxAggregateBase.cpp
@@ -396,6 +396,12 @@ class MinMaxAggregateBase : public exec::Aggregate {
       return;
     }
 
+    static const CompareFlags kCompareFlags{
+        true, // nullsFirst
+        true, // ascending
+        false, // equalsOnly
+        nullHandlingMode};
+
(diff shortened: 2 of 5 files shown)


rui-mo (commented):
> This issue occurs because kCompareFlags in compare function is initialized with kNullAsIndeterminate, https://github.com/facebookincubator/velox/blob/main/velox/functions/lib/aggregates/MinMaxAggregateBase.cpp#L610

Thanks for working on the fix! Shall we handle to the null handling mode like below?

https://github.com/facebookincubator/velox/blob/000939f119644a2d1b60de1b1ffcbd6ea2da00f5/velox/functions/lib/aggregates/MinMaxAggregateBase.cpp#L614-L623

zhli1142015:
Now, all types' min aggregation functions use the same `kCompareFlags` instance (which is defined in `compare` function). However, min aggregation functions for different types may require different null-handling behavior. I believe it's better to pass this as a parameter rather than using a static instance.
For string, I think `kNullAsIndeterminate` here is a default value, NullHandlingMode is not in use.
https://github.com/facebookincubator/velox/blob/15f4ae6f504b90d24035654bfb26af98f730b34c/velox/exec/ContainerRowSerde.cpp#L419C24-L419C31
Thanks.

velox/functions/lib/aggregates/Compare.cpp
```diff
@@ -22,14 +22,8 @@ int32_t compare(
     const SingleValueAccumulator* accumulator,
     const DecodedVector& decoded,
     vector_size_t index,
-    CompareFlags::NullHandlingMode nullHandlingMode) {
-  static const CompareFlags kCompareFlags{
```
rui-mo, line 26: Another way to fix might be to make 'kCompareFlags' non-static, how do you think?

zhli1142015: This will cause a CompareFlags to be constructed for each comparison, which is an unnecessary overhead.

velox/functions/lib/aggregates/MinMaxAggregateBase.cpp
```diff
@@ -396,6 +396,12 @@ class MinMaxAggregateBase : public exec::Aggregate {
       return;
     }
 
+    static const CompareFlags kCompareFlags{
+        true, // nullsFirst
+        true, // ascending
+        false, // equalsOnly
+        nullHandlingMode};
```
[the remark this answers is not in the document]
zhli1142015: > Using static for a variable that depends on a template parameter might not be a good practice

Updated, thanks.

rui-mo (approved):

PR #12385, opened 2025-02-19
refactor(types): Harden DeserializedTypeCache::deserialize

mbasmanova:
Summary: Make sure there are no duplicate IDs in the serialized cache.

Differential Revision: D69835376


xiaoxmeng (approved):
@mbasmanova thanks for the followup.

PR #12388, opened 2025-02-19
fix: Fix flaky abort injection in MemoryArbitrationFuzzer

tanjialiang:
Summary: The tracking of abort requests is non-differentiable from different queries (threads), causing one query's verification incorrectly impacted by another query's abort request. The PR makes it thread safe.

Differential Revision: D69860780


xiaoxmeng (approved):

PR #12391, opened 2025-02-20
misc(fuzzer): Remove verifyWindow() in AggregationFuzzer

kagamiori:
Summary:
We used to test window operations in aggregaiton fuzzer before the window fuzzer was built. Since 
we now have a window fuzzer that has much better coverage, remove the verifyWindow method in 
AggregationFuzzer.

Differential Revision: D69886065

### DIFF
--- a/velox/exec/fuzzer/AggregationFuzzer.cpp
+++ b/velox/exec/fuzzer/AggregationFuzzer.cpp
@@ -34,11 +34,6 @@ DEFINE_bool(
     true,
     "When true, generates plans with aggregations over sorted inputs");
 
-DEFINE_bool(
-    enable_window_reference_verification,
-    false,
-    "When true, the results of the window aggregation are compared to reference DB results");
-
 using facebook::velox::fuzzer::CallableSignature;
 using facebook::velox::fuzzer::SignatureTemplate;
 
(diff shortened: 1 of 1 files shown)


velox/exec/fuzzer/AggregationFuzzer.cpp
```diff
@@ -374,110 +358,85 @@ void AggregationFuzzer::go() {
       std::vector<TypePtr> argTypes = signature.args;
       std::vector<std::string> argNames = makeNames(argTypes.size());
 
-      // 10% of times test window operator.
+      const bool sortedInputs = FLAGS_enable_sorted_aggregations &&
+          canSortInputs(signature) && vectorFuzzer_.coinToss(0.2);
+
```
natashasehgal, line 363: Why was the removal of 10% window fuzzer replaced by 0.2 coin tosses ?

kagamiori: Hi @natashasehgal, github doesn't show the code diff very well. You can see in [D69886065](https://www.internalfb.com/diff/D69886065) that I actually remove the entire if-body of the 10%-window-fuzzer testing.

PR #12392, opened 2025-02-20
feat(fuzzer): Allow functions to be tested only with sorted input in aggregation fuzzer

kagamiori:
Summary: The tdigest_agg() function needs to be tested only with input of determined order so that Velox result is comparable with Presto's. This is because different ordering of input values affects the content in TDigest. Therefore, this diff makes it possible to specify a set of functions to be tested only with ordered inputs in aggregation fuzzer. Aggregation fuzzer will then use `order by arg0, arg1, ...` inside the aggregaiton function calls of this function.

Differential Revision: D69886108


natashasehgal (approved):
LGTM!
Nit - Meta internal diff details in Github description should be removed

PR #12397, opened 2025-02-20
misc: Remove the unnecessary merge in addSingleGroupRawInput of VarianceAggregate

zhli1142015:
On lines 235 and 240, we performed an unnecessary Accumulator merge 
 operation. Instead, we can call `updateNonNullValue(char* group, T value)` 
 to update the inputs to the result Accumulator directly.

 We noticed this issue when a [Spark unit test](https://github.com/apache/spark/blob/f447c43af4a138b73c3caef9859c6014cf28bcfa/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/WindowQuerySuite.scala#L85) on our side failed due 
to floating-point precision differences (18.475208614068027 -> 18.475208614068023). 
Upon investigation, we found that the extra merge was the root cause.
image

### DIFF
--- a/velox/functions/prestosql/aggregates/VarianceAggregates.cpp
+++ b/velox/functions/prestosql/aggregates/VarianceAggregates.cpp
@@ -230,14 +230,12 @@ class VarianceAggregate : public exec::Aggregate {
       });
     } else if (!exec::Aggregate::numNulls_ && decodedRaw_.isIdentityMapping()) {
       const T* data = decodedRaw_.data<T>();
-      VarianceAccumulator accData;
-      rows.applyToSelected([&](vector_size_t i) { accData.update(data[i]); });
-      updateNonNullValue<false>(group, accData);
-    } else {
-      VarianceAccumulator accData;
       rows.applyToSelected(
-          [&](vector_size_t i) { accData.update(decodedRaw_.valueAt<T>(i)); });
-      updateNonNullValue(group, accData);
+          [&](vector_size_t i) { updateNonNullValue(group, data[i]); });
+    } else {
+      rows.applyToSelected([&](vector_size_t i) {
+        updateNonNullValue(group, decodedRaw_.valueAt<T>(i));
+      });
     }
   }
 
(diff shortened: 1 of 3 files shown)


zhli1142015:
@Yuhta and @mbasmanova could you help to take a look? Thanks.

velox/functions/prestosql/aggregates/VarianceAggregates.cpp
```diff
@@ -230,14 +230,12 @@ class VarianceAggregate : public exec::Aggregate {
       });
     } else if (!exec::Aggregate::numNulls_ && decodedRaw_.isIdentityMapping()) {
       const T* data = decodedRaw_.data<T>();
-      VarianceAccumulator accData;
-      rows.applyToSelected([&](vector_size_t i) { accData.update(data[i]); });
```
mbasmanova, line 234: I believe this is a performance optimization. It is cheaper to aggregate all values "locally" first and update accumulator once.

zhli1142015: Thanks for pointing this out. Do you think it’s necessary to proceed with this change, such as creating relevant benchmarks to evaluate its performance impact?

mbasmanova: If we want to proceed, it would be helpful to understand perf impact.

mbasmanova: @zhli1142015 Thank you for measuring perf impact. Would you clarify what do we gain at the expense of 10% perf regression?

zhli1142015: This change aligns the computation behavior more closely with Spark. It only affects very small decimal values. So far, it primarily helps our unit tests pass and is unlikely to have any real-world impact.

zhli1142015: Yes, will update the test case in our side.

zhli1142015: @mbasmanova , thank you for reviewing and pointing out the performance impact of this change.

PR #12401, opened 2025-02-20
fix MapTopN Ensure Output Order Matches Input Order

duxiao1212:
Summary:
In Presto, the entries of the output preserve the original order, while in Velox it’s ascending.
 {F1975293608} 

However, in Velox, we are following the descending order.

Differential Revision: D69710376


zacw7 (approved):
Thanks for fixing it.

PR #12402, opened 2025-02-20
refactor: Support parameters in custom type resolving

Yuhta:
Summary:
Support `TypeParameter` when creating custom types.  For `tdigest` this
is not absolutely necessary, since the only parameter supported is `double`, but
it would be nice to have this ready so we can use it with `qdigest(T)`.

Differential Revision: D69879669

### DIFF
--- a/velox/expression/tests/CustomTypeTest.cpp
+++ b/velox/expression/tests/CustomTypeTest.cpp
@@ -55,7 +55,8 @@ class FancyIntType : public OpaqueType {
 
 class FancyIntTypeFactories : public CustomTypeFactories {
  public:
-  TypePtr getType() const override {
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK(parameters.empty());
     return FancyIntType::get();
   }
 
@@ -145,7 +146,8 @@ struct FancyPlusFunction {
 
 class AlwaysFailingTypeFactories : public CustomTypeFactories {
  public:
-  TypePtr getType() const override {
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK(parameters.empty());
     VELOX_UNSUPPORTED();
   }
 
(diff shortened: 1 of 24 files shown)


velox/expression/tests/CustomTypeTest.cpp
```diff
@@ -256,7 +258,7 @@ TEST_F(CustomTypeTest, nullConstant) {
```
natashasehgal, line 258: Does CustomTypeTest.getCustomTypeNames [test](https://github.com/facebookincubator/velox/blob/main/velox/expression/tests/CustomTypeTest.cpp#L221) need to be updated to use another function? T-Digest isn't returned from [Type.getCustomNames()](https://github.com/facebookincubator/velox/blob/main/velox/type/Type.cpp#L1044). 

Is that expected?

Yuhta: We probably are not registering the type factory during test initialization.  That's super minor.

velox/functions/prestosql/types/TDigestType.cpp
```diff
@@ -17,6 +17,18 @@
...
 
+folly::dynamic TDigestType::serialize() const {
+  folly::dynamic obj = folly::dynamic::object;
+  obj["name"] = "Type";
+  obj["type"] = name();
+  folly::dynamic children = folly::dynamic::array;
+  for (auto& param : parameters_) {
+    children.push_back(param.type->serialize());
+  }
+  obj["cTypes"] = children;
```
xiaoxmeng, line 28:
```
nit: s/cTypes/childTypes/
```

velox/functions/prestosql/types/TDigestType.h
```diff
@@ -73,8 +74,10 @@ using TDigest = CustomType<TDigestT>;
 
 class TDigestTypeFactories : public CustomTypeFactories {
  public:
-  TypePtr getType() const override {
-    return TDIGEST();
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK_EQ(parameters.size(), 1);
+    VELOX_CHECK(parameters[0].kind == TypeParameterKind::kType);
```
xiaoxmeng, line 79: can we use VELOX_CHECK_EQ?

xiaoxmeng (approved):
@Yuhta LGTM. Thanks!

velox/functions/prestosql/types/TDigestType.cpp
Yuhta: That name is already used by the serialization framework, not sure how dangerous is if we change it now

velox/functions/prestosql/types/TDigestType.h
Yuhta: No there is a build error on GitHub complaining don't know how to convert `kType` to string

Issue #12278, opened 2025-02-06
Avx2 check makes unit tests cannot be passed on ARM64 machine


anlowee:
### Bug description

# Description
When running unit test `E2EFilterTest.stringDirect` (in `velox_dwrf_e2e_filter_test`), the code execution path in ARM64 diverse from AMD64 (w/ Avx2) [here](https://github.com/facebookincubator/velox/blob/main/velox/dwio/dwrf/reader/SelectiveStringDirectColumnReader.cpp#L428). Since ARM64 does not have Avx2 so that this condition cannot be passed and the code under this branch will not be executed, and eventually the unit test failed because the `facebook::velox::dwrf::SelectiveStringDirectColumnReader::try8ConsecutiveSmall` won't be called in this case.

Similarly, when running unit test `E2EFilterTest.stringDictionary` (in `velox_dwrf_e2e_filter_test`), the code execution path in ARM64 diverse from AMD64 (w/ Avx2) [here](https://github.com/facebookincubator/velox/blob/1eb79878dc1f1d3fbf2942220ea2e749a33ce1d2/velox/dwio/common/DecoderUtil.h#L410). And this eventually makes the `addValue` functions incorrectly overloaded ([here](https://github.com/facebookincubator/velox/blob/1eb79878dc1f1d3fbf2942220ea2e749a33ce1d2/velox/dwio/common/tests/utils/E2EFilterTestBase.h#L43-L73)), the `value` is always 0 so it always goes to the first `addValue`.

And there are some other unit tests in `velox_parquet_e2e_filter_test` failed due to the incorrect overloading `addValue` seems to be caused by the similar reason.

Some questions about this:
1. Is it safe to skip these two unit tests on ARM64?
2. If it is not safe, then any suggestions to get around with this?

# Reproduction
Running the problematic unit test:
```
_build/debug/velox/dwio/dwrf/test/velox_dwrf_e2e_filter_test
_build/debug/velox/dwio/parquet/tests/reader/velox_parquet_e2e_filter_test
```

### System information

Velox System Info v0.0.2
Commit: 9f4419e201808bd3bc9fbe85415121c15651f341
CMake Version: 3.30.3
System: Linux-6.1.112+
Arch: aarch64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 18.1.8
C Compiler: /usr/bin/cc
C Compiler Version: 18.1.8
[...]

PR #12403, opened 2025-02-20
test: Skip some unit tests that cannot be passed without AVX2

anlowee:
# Summary
In #12278 , some unit tests cannot be passed on machine without AVX2 support, for example, basically all ARM machines. The root cause of these failures turn out to be that [this function](https://github.com/facebookincubator/velox/blob/main/velox/common/process/ProcessBase.cpp#L111-L117) will return false due to lack of AVX2, so that the code execution path is different with those have AVX2 support, so these unit tests can be passed only on machines with AVX2.

Since currently these unit tests are not compatible with AVX2, as suggested in the [issue comment](https://github.com/facebookincubator/velox/issues/12278#issuecomment-2641530766), we can simply skip them if there is no AVX2 on the machine.

# More details about the failed unit tests
Here are some logs of those failed unit tests.
### E2EFilterTest.integerDirect
```
# Most of failures be like
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:214: Failure
Value of: loadWithHook(batches, rowVector, i, child, hitRows, rowIndex)
  Actual: false
Expected: true
Google Test trace:
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:150: Read with filter
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:288: Lazy
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:324: No row group skip

# The last failure be like
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:358: Failure
Expected: (0) < (runtimeStats_.skippedStrides), actual: 0 vs 0
Google Test trace:
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:340: Row group skip
```
### E2EFilterTest.integerDictionary
```
# The failures are a mix of the following two types of logs
# First one
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:214: Failure
Value of: loadWithHook(batches, rowVector, i, child, hitRows, rowIndex)
  Actual: false
Expected: true
Google Test trace:
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:150: Read with filter
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:288: Lazy
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:324: No row group skip

# Second one
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:214: Failure
Value of: loadWithHook(batches, rowVector, i, child, hitRows, rowIndex)
  Actual: false
Expected: true
Google Test trace:
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:150: Read with filter
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:288: Lazy
${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.cpp:364: Subfield pruning
```
### E2EFilterTest.map
The above four unit tests have the same failure logs as **E2EFilterTest.integerDictionary**.
### E2EFilterTest.varbinaryDictionary
All of the three unit tests above failed by triggering a `VELOX_FAIL()`. The `process::hasAVX2()` gets called [here](https://github.com/facebookincubator/velox/blob/1eb79878dc1f1d3fbf2942220ea2e749a33ce1d2/velox/dwio/common/DecoderUtil.h#L410), to determine whether to use the fast path ([here](https://github.com/facebookincubator/velox/blob/1eb79878dc1f1d3fbf2942220ea2e749a33ce1d2/velox/dwio/dwrf/common/RLEv1.h#L264-L267)). If not, then this [line](https://github.com/facebookincubator/velox/blob/1eb79878dc1f1d3fbf2942220ea2e749a33ce1d2/velox/dwio/dwrf/common/RLEv1.h#L293) is eventually executed and calls the `addValue`. However, the `value_` is 0, but the its type is somehow set to the `folly::StringPiece`, so the type check in the `addValue` cannot be passed and the `VELOX_FAIL()` gets triggered. This was also described in #12278 .

The error message is:
```
E20250218 17:43:46.591727 15575629 Exceptions.h:66] Line: ${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.h:47, Function:addValue, Expression:  , Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
Error Source: RUNTIME
Error Code: INVALID_STATE
Retriable: False
Function: addValue
File: ${VELOX_REPO_DIR}/velox/dwio/common/tests/utils/E2EFilterTestBase.h
Line: 47
```
## velox_dwrf_e2e_filter_test
There are two unit tests failed and all were described in #12278 .


velox/dwio/dwrf/test/E2EFilterTest.cpp
```diff
@@ -242,6 +242,7 @@ TEST_F(E2EFilterTest, floatAndDouble) {
 }
 
 TEST_F(E2EFilterTest, stringDirect) {
+#ifdef __AVX2__
```
Yuhta, line 245: These tests should work with either SSE4.1 or NEON. Let's use that as a guard instead (`#if XSIMD_WITH_SSE4_1 || XSIMD_WITH_NEON || XSIMD_WITH_NEON64`). AVX2 is too much requirements here.

Yuhta (changes requested):

velox/dwio/dwrf/test/E2EFilterTest.cpp
anlowee: Just tried this on my M1 Macbook and a Neoverse_V2 ARM server, where the `XSIMD_WITH_SSE4_1` is 0, `XSIMD_WITH_NEON` and `XSIMD_WITH_NEON64` are 1, so the tests will be executed but failed with the same error message. Are there any other scripts to run or dependencies to install before I run these tests? Currently all preparing work I do is just following the setting up instructions in the [README](https://github.com/facebookincubator/velox/blob/main/README.md)

Yuhta: Maybe there is some bug with NEON implementation.  At least we need to make sure SSE4.1 is working (and I suspect this will also fix NEON as well).  Can we try to fix it instead of skipping them?

anlowee: Yea sure, needs to look into it deeper.

PR #12406, opened 2025-02-20
misc(json): Add jsoncons as a vendored library

bikramSingh91:
Summary:
This is part of an effort to add support for complete set of json path tokens (See https://github.com/facebookincubator/velox/discussions/12394 for more details).
As a first step we add jsoncons as a vendored library and in the upcoming changes we will start to incrementally use it.

The code for the library has been pulled from https://github.com/danielaparker/jsoncons with some changes. The changes include:

- Only files relevant to json and json parsing have been included
- The #include directives have been updated to point to local paths
- Prepend Apache license while preserving the existing boost license
- The namespaces have been prepended with facebook::velox to avoid namespace conflict if velox is included in projects that already use jsoncons
- Made changes to the json path parser to conform with jayway library which is a json library currently used by presto java.
- Added a flag to the json path parser that returns an error for paths not currently supported by velox.

Changes made to the parser include:
- Added support for unquoted identifier after bracket `$[<identifier>]` that can also start with a number like `$[Key]` or `$[30day]`
- Added support for bracket (`[`) after dot `$.['key']` (redundant use of dot operator)
- Added support for dash (`-`) in the middle of an unquoted string
- Added support for prepending by default `$.` in not included at the beginning of the path. This is to support cases like `[0]` or `key` which will now be equivalent to `$.[0]` or `$.key`

Differential Revision: D68564596

### DIFF
--- a/velox/external/jsoncons/allocator_holder.hpp
+++ b/velox/external/jsoncons/allocator_holder.hpp
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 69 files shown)


kevinwilfong:
Do you want to include patches with the changes you made to these files (e.g. see the patches in velox/external/date) otherwise it's really hard to tell what changed which
a) makes reviewing difficult
b) more importantly will make it difficult to reapply those changes if we ever need to import a more recent version

xiaoxmeng (approved):
@bikramSingh91 thanks!

PR #12407, opened 2025-02-20
feat(python): Add order_by() and limit() nodes

pedroerp:
Summary: Add support for order_by and limit nodes.

Reviewed By: hongtaoy

Differential Revision: D69879403

### DIFF
--- a/velox/py/plan_builder/PyPlanBuilder.h
+++ b/velox/py/plan_builder/PyPlanBuilder.h
@@ -174,18 +174,34 @@ class PyPlanBuilder {
   /// Add a single-stage aggregation given a set of group keys, and
   /// aggregations. Aggregations are specified as SQL expressions and currently
   /// use DuckDB's parser semantics.
-  PyPlanBuilder& singleAggregation(
+  PyPlanBuilder& aggregate(
       const std::vector<std::string>& groupingKeys,
       const std::vector<std::string>& aggregations);
+
+  /// Sorts the input based on the values of sorting keys.
+  ///
+  /// @param keys List of columns to order by. The strings can be column
+  /// names and optionally contain the sort orientation ("col" or "col DESC").
+  /// @param is_partial
+  PyPlanBuilder& orderBy(const std::vector<std::string>& keys, bool isPartial);
+
+  /// Limit how many rows from the input to produce as output.
+  ///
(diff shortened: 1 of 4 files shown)


velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -174,18 +174,34 @@ class PyPlanBuilder {
...
   /// use DuckDB's parser semantics.
-  PyPlanBuilder& singleAggregation(
+  PyPlanBuilder& aggregate(
       const std::vector<std::string>& groupingKeys,
       const std::vector<std::string>& aggregations);
 
+  /// Sorts the input based on the values of sorting keys.
+  ///
+  /// @param keys List of columns to order by. The strings can be column
+  /// names and optionally contain the sort orientation ("col" or "col DESC").
```
kostasxmeta, line 184: Let's also add a note here about what the default is if a sort orientation is not provided (ASC ?)

velox/py/plan_builder/plan_builder.cpp
```diff
@@ -157,6 +157,37 @@ PYBIND11_MODULE(plan_builder, m) {
...
+          &velox::py::PyPlanBuilder::orderBy,
+          py::arg("keys"),
+          py::arg("is_partial") = false,
+          py::doc(R"(
+        Sorts the input based on the values of sorting keys.
+
+        Args:
+          keys: List of columns to order by. The strings can be column names
+                and optionally contain the sort orientation ("col" or
+                "col DESC").
```
kostasxmeta, line 171: Same comment here as above about default sort orientation if it's not provided

velox/py/tests/test_runner.py
```diff
@@ -64,6 +64,30 @@ def test_runner_with_values(self):
             total_size += vector.size()
         self.assertEqual(total_size, 100)
 
+    def test_runner_with_values_order_limit(self):
+        vectors = []
+        batch_size = 10
+        num_batches = 10
+
+        for i in range(num_batches):
```
kostasxmeta, line 72: nit: A short comment here about what the test data we are generating looks like would be useful. It would also help understand why the expected result is what it is a bit more easily.

kostasxmeta (approved):
LGTM! Just a couple of small comments

PR #12409, opened 2025-02-20
fix(json): Ensure input is parsed as json for json extract functions

bikramSingh91:
Summary:
We recently noticed as detailed in https://github.com/prestodb/presto/issues/24563
that some json functions that take varchar as input and output a json type can skip
parsing the input. This can result in non-canonicalized result and inconsistent behavior.
This change ensures that json_extract and json_extract_scalar always parse the input
in these instances.

Differential Revision: D69611952


mbasmanova (commented):
@bikramSingh91 Would you update the documentation to clarify the "fixed" behavior?

kgpai (approved):

PR #12410, opened 2025-02-20
misc: Add partitioned writer count metrics to hive data sink

tanjialiang:
large number of partitioned writers may cause issues like large number of open files and memory consumption. Adding this runtime metrics to disclose this critical stats for more efficient debugging related issues.

### DIFF
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -683,6 +683,9 @@ void HiveDataSink::abort() {
 }
 
 void HiveDataSink::closeInternal() {
+  addThreadLocalRuntimeStat(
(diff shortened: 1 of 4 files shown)


bikramSingh91 (approved):

velox/connectors/hive/HiveDataSink.cpp
```diff
@@ -806,6 +806,8 @@ uint32_t HiveDataSink::appendWriter(const HiveWriterId& id) {
               .stats = ioStats_.back().get(),
           }),
       options);
+  addThreadLocalRuntimeStat(
```
xiaoxmeng, line 809: Shall we report once for each table write operator so we can see the distribution better?

PR #12411, opened 2025-02-20
[Do not review]fix: Remove timestamp check for TableScan for FuzzerUtil

aditi-pandit:


aditi-pandit:
@kagamiori : So I saw the error in RowNumberFuzzer after removing the timestamp check in isTableScanSupported.

I don't see presto_url configured for the RowNumber Fuzzer workflow here https://github.com/facebookincubator/velox/blob/main/.github/workflows/scheduled.yml#L724

We should change that. It might just fix the problem.

Issue #12412, opened 2025-02-20
Experimental RAPIDS cuDF Backend for Velox


bdice:
### Description

This issue tracks the development of a RAPIDS cuDF backend for Velox. The project will be implemented in a series of incremental pull requests described below. Our initial goal is to run the [TPC-H query plans defined by Velox](https://github.com/facebookincubator/velox/blob/main/velox/benchmarks/tpch/TpchBenchmark.cpp), thereby covering a broad range of basic data analytics operations end-to-end on GPU.

All of the work described below currently exists in a fork of this repository, which we hope to upstream piece by piece. I plan to update this issue frequently as we make progress on the initial stages of this work.

cuDF team: @karthikeyann @devavret @mhaseeb123 @GregoryKimball

cc: @pedroerp @oerling @Yuhta @assignUser

## High level design

- Build:
  - A new Docker container with CUDA 12.8 is needed for developers (based on the existing Ubuntu 22.04 container). 🐳
  - cuDF is fetched and built if the CMake option `VELOX_ENABLE_CUDF` is true. The dependency logic will go in `CMake/resolve_dependency_modules/cudf.cmake`. (cuDF only links to the CUDA runtime, not the CUDA driver library.)
  - I plan to work with @assignUser on CI testing. We will probably want to run GPU builds on CPU nodes, and only run GPU CI tests if the cuDF code has changed.
- Run:
  - A Velox `DriverAdapter` is used to replace CPU operators with GPU operators that call cuDF's C++ code. This `DriverAdapter` can be registered at startup by the application using Velox to enable the cuDF backend.
    - Each operator will have a cuDF equivalent. For example, `OrderBy` will be replaced by `CudfOrderBy`.
  - In between CPU operators and GPU operators, another conversion operator is inserted to handle CPU->GPU and GPU->CPU data movement. This allows cuDF operators to be used alongside existing Velox operators.
    - The conversion currently uses Arrow (Velox to Arrow, then Arrow to cuDF). A direct Velox-to-cuDF interop without Arrow may be built in the future for higher performance.
  - Currently, no custom CUDA kernels are needed for this code. All functionality is implemented in pure C++ calling cuDF, which implements the CUDA kernels.
  - There is a lot more to say about tuning for performance (GPU batch sizes, CUDA streams, number of Velox drivers, ...) but I'm leaving that out of this document for the moment.
- Tests:
[...]

PR #12413, opened 2025-02-20
build: Add Docker images for CUDA development

bdice:
This PR is the first step towards adding a RAPIDS cuDF backend (#12412). It adds CUDA to the `adapters` CI images. This container will allow us to share a development environment that has CUDA compilers and libraries along with the existing Velox container infrastructure.


assignUser:
As discussed we will likely run the build in the 'adapters' job which is based on a centos image, so installing 12.8 there by default would be the way to go (currently installed in the workflow). If you want these for local development that's fine or is there a reason to prefer ubuntu?

bdice:
Thanks @assignUser. I am more familiar with Ubuntu but I think we can make that change. I’ll work on that early next week.

bdice:
@assignUser It seems like the adapters images don't use CUDA. I would need to change the base image to something deriving from `nvidia/cuda` (probably the base flavor, and install our own CUDA Toolkit) for the best driver compatibility. Currently this derives from `ghcr.io/facebookincubator/velox-dev:centos9`. I guess I would need to copy the setup from that `centos` container on top of an `nvidia/cuda` CentOS image, and then add the rest of the adapters. Is that still the desired path here?

assignUser:
Hm,if you want to use a cuda base image that makes things a bit trickier. We install cuda using this function, currently in the workflow as it's so fast that I haven't had a reason to move it into the dockerfile but we can do that. Unless that is lacking in some way compared to a proper cuda base image?

bdice:
There's some additional logic in the `nvidia/cuda` base images that makes driver compatibility much broader than what a "bare" image would have (using `cuda-compat`, `NVIDIA_REQUIRE_CUDA` environment variables, etc.).

scripts/ubuntu-22.04-cuda-12.8-cpp.dockerfile
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
```
jinchengchenghh, line 2: I meet this exception
```
docker build -t cudf-velox -f scripts/ubuntu-22.04-cuda-12.8-cpp.dockerfile ./

[+] Building 2.2s (2/2) FINISHED                                                                                                                                                                docker:default
 => [internal] load build definition from ubuntu-22.04-cuda-12.8-cpp.dockerfile                                                                                                                           0.0s
 => => transferring dockerfile: 1.23kB                                                                                                                                                                    0.0s
 => ERROR [internal] load metadata for docker.io/nvidia/cuda:12.8.0-devel-ubuntu22.04                                                                                                                     2.0s
------
 > [internal] load metadata for docker.io/nvidia/cuda:12.8.0-devel-ubuntu22.04:
------
ubuntu-22.04-cuda-12.8-cpp.dockerfile:18
--------------------
  16 |     ARG tz="Europe/Madrid"
  17 |     
  18 | >>> FROM ${base}
  19 |     
  20 |     SHELL ["/bin/bash", "-o", "pipefail", "-c"]
--------------------
ERROR: failed to solve: nvidia/cuda:12.8.0-devel-ubuntu22.04: failed to resolve source metadata for docker.io/nvidia/cuda:12.8.0-devel-ubuntu22.04: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/nvidia/cuda/manifests/sha256:54f18e2a8e1b3d03f77b9a6dc905533da46ac93a5513f10e8ba8e560db9fa5ab: 429 Too Many Requests - Server message: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```
I know it is because docker hub as throttling for pulling the images. Is there any chance to avoid this exception?

assignUser:
@bdice I haven't had time to come to a final conclusion but I looked at image sizes and the rocky cuda image is ~500M vs ~250M for stream9. While that is a big difference until we work on our final images sizes it doesn't really matter xD As the centos9 image is ~2G and the adapters one is a whopping 7G. I'll have to investigate if we can reduce that size (#12527). 

But for this issue it means that imo we can just use the cuda base image for centos.dockerfile (with what ever additions you need added to the adapters image?). Rocky should be a good 1:1 replacement for stream9 afaik?

bdice:
Great! I will work on that.

> Rocky should be a good 1:1 replacement for stream9 afaik?

Yes, it should be functionally equivalent.

assignUser:
Cc @majetideepak @kgpai any concerns?

bdice:
I should clarify, now that I'm looking at the numbers in your comment again -- cuDF needs the `devel` flavor of `nvidia/cuda` in order to build. The image [`nvidia/cuda:12.8.0-devel-rockylinux9`](https://hub.docker.com/layers/nvidia/cuda/12.8.0-devel-rockylinux9/images/sha256-986f4aae8ec124172ae3321575ad7e67180171e1d29143888871087eaddb8e76) is about 4.62 GB compressed.

Given that, would my original approach of having a separate image for CUDA development (rather than increasing the size of `adapters` image) be worth considering?

assignUser:
Ah, hm. Well my concern regarding the size is mostly about the time it takes CI to fetch it but using ghcr.io in gha is well cdn'ed.

Adding a separate build for cudf would likely be much more expensive ci time wise... (As you will still build most of velox I assume?).

A separate build would be fine if we limit it to run when changes to the relevant source is made or something, though that might  not catch issues on general velox PRs that impact cudf stuff.

I am mostly bringing this up, because we need larger runners to reliably build velox due to the high RAM usage and CI cost has been a concern before, maybe @kgpai or @pedroerp can chime in?

docker-compose.yml
```diff
@@ -29,6 +29,35 @@ services:
...
+    #   or
+    #   docker-compose run -e NUM_THREADS=<NUMBER_OF_THREADS_TO_USE> --rm ubuntu-cuda-cpp
+    #   to set the number of threads used during compilation
+    image: ghcr.io/facebookincubator/velox-dev:ubuntu-22.04-cuda-12.8
+    build:
+      context: .
+      dockerfile: scripts/ubuntu-22.04-cuda-12.8-cpp.dockerfile
+    environment:
+      NUM_THREADS: 8 # default value for NUM_THREADS
+      VELOX_DEPENDENCY_SOURCE: BUNDLED # Build dependencies from source
```
kgpai, line 50: If we are building an image here, would be better to take advantage of that and have dependencies be AUTO/SYSTEM (to not add to the build times) ? (any concerns with that @assignUser )

kgpai:
> I am mostly bringing this up, because we need larger runners to reliably build velox due to the high RAM usage and CI cost has been a concern before, maybe @kgpai or @pedroerp can chime in?

Lets try this out on a trial basis and I can keep an eye on the costs / stability etc. If it goes beyond our budget relative to the signal we get then maybe we can run it against main periodically rather than on every diff (presuming thats what we are planning to use this image for). I dont have any concerns if we are going to use it for periodic jobs.

kgpai:
@assignUser, @bdice : Just an fyi , our largest CI costs are still Mac builds, and if this isnt going to use a dedicated gpu instance the increase in image size should mostly be ok imo (but I will keep an eye to be sure).

docker-compose.yml
```diff
@@ -52,13 +52,48 @@ services:
...
       CCACHE_DIR: "/velox/.ccache"
-      EXTRA_CMAKE_FLAGS: -DVELOX_ENABLE_PARQUET=ON 
+      EXTRA_CMAKE_FLAGS: -DVELOX_ENABLE_PARQUET=ON
                          -DVELOX_ENABLE_S3=ON
     volumes:
       - .:/velox:delegated
     working_dir: /velox
     command: /velox/scripts/docker-command.sh
 
+  adapters-cuda:
```
bdice, line 62: This is copy of `adapters-cpp`, but with GPU support configured. See the comment below.

docker-compose.yml
```diff
@@ -52,13 +52,48 @@ services:
...
+      EXTRA_CMAKE_FLAGS: -DVELOX_ENABLE_PARQUET=ON
+                         -DVELOX_ENABLE_S3=ON
+    privileged: true
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: 1
+              capabilities: [gpu]
```
bdice, line 90: This section is the difference between `adapters-cpp` and `adapters-cuda`.

assignUser:
Or that ^^

scripts/adapters.dockerfile
```diff
@@ -17,8 +17,22 @@ FROM $image
 
 COPY scripts/setup-helper-functions.sh /
 COPY scripts/setup-adapters.sh /
-RUN mkdir build && ( cd build &&  source /opt/rh/gcc-toolset-12/enable && \
-    bash /setup-adapters.sh ) && rm -rf build && dnf remove -y conda && dnf clean all
+COPY scripts/setup-centos9.sh /
```
bdice, line 20: This file should already be present, because this image is based on the CentOS image (and I don't think it is deleted from that image). Should we re-copy it or just assume it is present? I went with copying it again since we are re-copying `setup-helper-functions.sh` above.

scripts/adapters.dockerfile
```diff
@@ -17,8 +17,22 @@ FROM $image
...
+RUN mkdir build && \
+    ( \
+      cd build && \
+      source /opt/rh/gcc-toolset-12/enable && \
+      bash /setup-adapters.sh && \
+      bash /setup-centos9.sh install_cuda 12.8 \
```
devavret, line 26: Should installing cuda be hidden behind an arg to this?

assignUser (approved):
Thanks! The existing code seems to work well with 12.8, adapters build is :green_circle:

Yuhta:
@bdice @assignUser Is this error relevant? https://github.com/facebookincubator/velox/actions/runs/13770933756/job/38511459315?pr=12413

scripts/adapters.dockerfile
assignUser: The version maybe but the image is for our ci where we always need cuda:)

assignUser, line 26: It seems that the version isn't picked up by the function? 1971.9 Error: Unable to find a match: cuda-compat- cuda-driver-devel- cuda-minimal-build- cuda-nvrtc-devel-

bdice: I fixed this to call `install_cuda` like the CI workflow does: https://github.com/facebookincubator/velox/blob/c2eb47b599c42e80edbb2d9d339ebecd4edbcc09/.github/workflows/linux-build-base.yml#L58-L59

PR #12415, opened 2025-02-20
misc(test_utils): Add file and directory operations to HiveConnectorTestBase

kKPulla:
Summary:
Currently HiveConnectorTestBase has util method for writing to a file, given a file path. Added

1. Create/Remove ops for directories
2. Remove op for files

Differential Revision: D69955818


pedroerp:
@kKPulla we usually use `velox/exec/tests/utils/TempDirectoryPath.h` to create/delete temporary directories for tests. Is what you're trying to achieve not covered by that class?

kagamiori (approved):
LGTM.

PR #12416, opened 2025-02-21
refactor: Remove redundant target spill file size check

tanjialiang:


xiaoxmeng (commented):
As chat offline, we don't need this as spill config already handle this

xiaoxmeng (approved):

PR #12417, opened 2025-02-21
feat: Allow Spark get_json_object function to parse incomplete json

leoluan2009:
Spark get_json_object function allows incomplete json, for example: "select 
get_json_object('{"hello": "3.5"},', '$.'hello)" returns 3.5, but Velox returns 
null. This PR adds 'simdjsonParseIncomplete' to allow incomplete json input 
for Spark.

### DIFF
--- a/CMake/resolve_dependency_modules/simdjson.cmake
+++ b/CMake/resolve_dependency_modules/simdjson.cmake
@@ -34,3 +34,5 @@ if(${VELOX_SIMDJSON_SKIPUTF8VALIDATION})
 endif()
 
 FetchContent_MakeAvailable(simdjson)
+target_compile_definitions(simdjson
+                           PUBLIC SIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON)
--- a/velox/functions/prestosql/json/SIMDJsonUtil.h
+++ b/velox/functions/prestosql/json/SIMDJsonUtil.h
@@ -54,4 +54,9 @@ simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParse(
 simdjson::padded_string_view reusePaddedStringView(
     const std::string_view& json);
 
+/// Parse the input json string using a thread local on demand parser. Allow
+/// incomplete json input.
+simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParseIncomplete(
+    const simdjson::padded_string_view& json);
+
 } // namespace facebook::velox
--- a/velox/functions/prestosql/json/SIMDJsonUtil.cpp
+++ b/velox/functions/prestosql/json/SIMDJsonUtil.cpp
@@ -38,6 +38,12 @@ simdjson::padded_string_view reusePaddedStringView(
     const std::string_view& json) {
   return simdjson::padded_string_view(
       json.data(), json.length(), json.length() + simdjson::SIMDJSON_PADDING);
-};
+}
+
+simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParseIncomplete(
+    const simdjson::padded_string_view& json) {
+  thread_local simdjson::ondemand::parser parser;
+  return parser.iterate_allow_incomplete_json(json);
+}
 
 } // namespace facebook::velox
--- a/velox/functions/sparksql/GetJsonObject.h
+++ b/velox/functions/sparksql/GetJsonObject.h
@@ -54,7 +54,7 @@ struct GetJsonObjectFunction {
     }
     simdjson::ondemand::document jsonDoc;
     simdjson::padded_string paddedJson(json.data(), json.size());
-    if (simdjsonParse(paddedJson).get(jsonDoc)) {
+    if (simdjsonParseIncomplete(paddedJson).get(jsonDoc)) {
       return false;
     }
     const auto formattedJsonPath = jsonPath_.has_value()
--- a/velox/functions/sparksql/tests/GetJsonObjectTest.cpp
+++ b/velox/functions/sparksql/tests/GetJsonObjectTest.cpp
@@ -119,5 +119,34 @@ TEST_F(GetJsonObjectTest, nullResult) {
       std::nullopt);
 }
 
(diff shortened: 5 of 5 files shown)


leoluan2009:
@PHILO-HE , can you help to review? cc: @rui-mo

velox/functions/prestosql/json/SIMDJsonUtil.h
```diff
@@ -47,4 +47,9 @@ void simdjsonErrorsToExceptions(
 simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParse(
     const simdjson::padded_string_view& json);
 
+/// Parse the input incomplete json string using a thread local on demand
```
PHILO-HE, line 50: /// Parse the input json string using a thread local on demand parser. Allow incomplete json input.

PHILO-HE (commented):
Pasted simdjson's pr link for this incomplete json support: https://github.com/simdjson/simdjson/pull/2189. It's an experimental functionality.

Not sure whether there will be any side effect for Spark workload. It would be better to add more incomplete json cases in test and verify the result with vanilla Spark.

CMake/resolve_dependency_modules/simdjson.cmake
```diff
@@ -33,4 +33,5 @@ if(${VELOX_SIMDJSON_SKIPUTF8VALIDATION})
   set(SIMDJSON_SKIPUTF8VALIDATION ON)
 endif()
 
+add_definitions(-DSIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON)
```
assignUser, line 36: Please add this to the `simdjson::simdjson` target, no need to polute all compiler invocations with this definition.

assignUser (changes requested):

velox/functions/prestosql/json/SIMDJsonUtil.h
leoluan2009: Done

CMake/resolve_dependency_modules/simdjson.cmake
assignUser: After makeavailalbe add `target_compile_definitions(simdjson::simdjson PUBLIC SIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON)`
[Docs](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html#command:target_compile_definitions) We need to add it as public because the option is used in the public header and not only at compile time of simdjson.

assignUser (approved):
CMake :+1:

PHILO-HE (approved):
Let's land this firstly to see any Spark user's feedback.

PHILO-HE:
@leoluan2009, please revise the pr title, like

feat: Allow Spark get_json_object function to parse incomplete json

leoluan2009:
> @leoluan2009, please revise the pr title, like
> 
> feat: Allow Spark get_json_object function to parse incomplete json

Done, Thanks

velox/functions/prestosql/json/SIMDJsonUtil.h
```diff
@@ -47,4 +47,9 @@ void simdjsonErrorsToExceptions(
 simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParse(
     const simdjson::padded_string_view& json);
 
+/// Parse the input json string using a thread local on demand parser. Allow
+/// incomplete json input.
+simdjson::simdjson_result<simdjson::ondemand::document> simdjsonParseIncomplete(
+    const simdjson::padded_string_view& json);
```
rui-mo, line 53: Can we add this function in Spark folder as it is only used by Spark function?

rui-mo (commented):
Thanks.

velox/functions/prestosql/json/SIMDJsonUtil.h
PHILO-HE: @rui-mo, this experimental feature in simdjson was proposed by someone to fix semantic inconsistency for presto sql. The added code may be re-usable by presto in the future. So I think it's ok to keep it here.

rui-mo: cc: @Yuhta

rui-mo (commented):
Would you rebase this change? Thanks.

rui-mo (approved):
Would you rebase to bypass the build failures?

CMake/resolve_dependency_modules/simdjson.cmake
```diff
@@ -34,3 +34,5 @@ if(${VELOX_SIMDJSON_SKIPUTF8VALIDATION})
 endif()
 
 FetchContent_MakeAvailable(simdjson)
+target_compile_definitions(simdjson
+                           PUBLIC SIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON)
```
kgpai, line 38: Hi @leoluan2009, I am working on landing this but I am concerned  if this change also changes presto behavior ?
Are the changes affected by this flag only visible if iterate_allow_incomplete_json is called ?

kgpai: Looks like it only gates the allow_incomplete_json function : https://github.com/search?q=repo%3Asimdjson%2Fsimdjson%20SIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON&type=code

Yohahaha:
```
spark-sql (default)> select get_json_object('{"hello": "3.5",', '$.hello');
NULL
```

but this incomplete json return null, I prefer add a flag to call `allow_incomplete_json`

philo-he:
> ```
> spark-sql (default)> select get_json_object('{"hello": "3.5",', '$.hello');
> NULL
> ```
> 
> but this incomplete json return null, I prefer add a flag to call `allow_incomplete_json`

@Yohahaha, do you mean the result is "3.5" with this patch reverted? I am afraid that even if we introduce a flag, it should be always enabled to ensure the compatibility on those relevant cases. Then, I feel the flag becomes unnecessary. Is it possible to fix this case based on the current implementation?

PR #9096, opened 2024-03-15
Add support for DECIMAL types to Simple Function API

mbasmanova:
Use the new functionality to re-write decimal plus, minus, multiple, divide, between, negate, floor and round.

**type/Type.h**

	Add P1, P2, P3, P4, S1, S2, S3, S4 types to specify precision and scale parameters for decimal types during function registration.

	Add LongDecimal<P, S> and ShortDecimal<P, S> templates to specify decimal argument and return types during function registration.

```
  registerFunction<
      DecimalAddFunction,
      LongDecimal<P3, S3>,
      LongDecimal<P1, S1>,
      LongDecimal<P2, S2>>({"plus"}, constraints);
```

**expression/UdfTypeResolver.h**

Define arg_type and out_type for LongDecimal and ShortDecimal.

```
	arg_type<LongDecimal> = int128_t
	out_type<LongDecimal> = int128_t

	arg_type<ShortDecimal> = int64_t
	out_type<ShortDecimal> = int64_t
```

**functions/Registerer.h**

	Add optional ‘constraints’ parameter to registerFunction template. This allows to specify rules for calculating precision and scale for decimal return types.

```
template <template <class> typename Func, typename TReturn, typename... TArgs>
void registerFunction(
    const std::vector<std::string>& aliases = {},
    const std::vector<exec::SignatureVariable>& constraints = {})
```

	Here is how we can specify calculation of precision and scale for the return type of plus(decimal, decimal).

```
  std::vector<exec::SignatureVariable> constraints = {
      exec::SignatureVariable(
          P3::name(),
          fmt::format(
              "min(38, max({a_precision} - {a_scale}, {b_precision} - {b_scale}) + max({a_scale}, {b_scale}) + 1)",
              fmt::arg("a_precision", P1::name()),
              fmt::arg("b_precision", P2::name()),
              fmt::arg("a_scale", S1::name()),
              fmt::arg("b_scale", S2::name())),
          exec::ParameterType::kIntegerParameter),
      exec::SignatureVariable(
          S3::name(),
          fmt::format(
              "max({a_scale}, {b_scale})",
              fmt::arg("a_scale", S1::name()),
              fmt::arg("b_scale", S2::name())),
          exec::ParameterType::kIntegerParameter),
  };
```

**core/SimpleFunctionMetadata.h**

Extend SimpleFunctionMetadata to store physical types (TypeKind) of input arguments and return type in addition to signature.

Decimal “plus” function has a single signature: 

```
[...]

PR #12423, opened 2025-02-21
Parse Presto Data Size

natashasehgal:
Differential Revision: D69878928


velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,106 @@
...
+      {"ZB", Unit::ZETTABYTE},
+      {"YB", Unit::YOTTABYTE}};
+  auto it = unitMap.find(unitString);
+  VELOX_USER_CHECK(it != unitMap.end(), "Invalid data size: '{}'", unitString);
+  return it->second;
+}
+
+inline double parseValue(const std::string& value) {
+  try {
+    return std::stod(value);
```
bikramSingh91, line 55: presto java uses BigDecimal to convert this value, using double here can result in double precision issues.

An equivalent solution here would meant that we pull out the cast ([code pointer](https://github.com/facebookincubator/velox/blob/c6cbeeeaf0f708904b3e693b1a33d36299f6cf56/velox/expression/CastExpr-inl.h#L96)) and arithmetic ([code pointer](https://github.com/facebookincubator/velox/blob/c6cbeeeaf0f708904b3e693b1a33d36299f6cf56/velox/functions/sparksql/DecimalArithmetic.cpp#L332)) implementations for Decimal type into utility functions and them here directly.

@kgpai @Yuhta Any thoughts on whether its worth the effort ?

velox/functions/prestosql/registration/DataSizeFunctionsRegistration.cpp
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+#include "velox/functions/Registerer.h"
+#include "velox/functions/prestosql/DataSizeFunctions.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+void registerSimpleFunctions(const std::string& prefix) {
```
bikramSingh91, line 23: please update the docs in conversion.rst to add an entry for this

velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,106 @@
...
+  } catch (const std::invalid_argument&) {
+    VELOX_USER_FAIL("Invalid data size: '{}'", value);
+  } catch (const std::out_of_range&) {
+    VELOX_USER_FAIL("Value out of range: '{}'", value);
```
HeidiHan0000, line 59: Can this one also be formatted like "Value out of range: '{}' ('{}B')"?

velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,106 @@
...
+         (isdigit(dataSize[valueLength]) || dataSize[valueLength] == '.')) {
+    valueLength++;
+  }
+  VELOX_USER_CHECK_GT(valueLength, 0, "Invalid data size: '{}'", dataSize);
+  double value;
+  Unit unit;
+  try {
+    value = parseValue(dataSize.substr(0, valueLength));
+  } catch (const std::exception&) {
+    throw std::invalid_argument("Invalid data size: '" + dataSize + "'");
```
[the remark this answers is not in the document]
natashasehgal: Thanks, updated this

velox/functions/prestosql/DataSizeFunctions.h
natashasehgal: Updated this

velox/functions/prestosql/registration/DataSizeFunctionsRegistration.cpp
natashasehgal: Updated Velox documents to match Presto documents

velox/functions/prestosql/registration/DataSizeFunctionsRegistration.cpp
```diff
@@ -0,0 +1,33 @@
...
+  registerFunction<ParsePrestoDataSizeFunction, int128_t, Varchar>(
```
bikramSingh91, line 24: check this PR out for all you need to know about decimal types and simple functions (the description is pretty good) https://github.com/facebookincubator/velox/pull/9096

You need to specify LongDecimal as output type and also specify a constraint since this function only outputs a decimal type with a specific precision and scale

Finally in the simple function definition, you need to use the correct out_type like  `out_type<LongDecimal<P1, S1>>`

velox/functions/prestosql/DataSizeFunctions.h
bikramSingh91: @natashasehgal whatever way gets decided here, the output would either be in double or decimal types(initially converted from string) and in either case it would be good to take a look at the cast implementation to make sure type conversion is accurate. Code pointer: https://github.com/facebookincubator/velox/blob/c6cbeeeaf0f708904b3e693b1a33d36299f6cf56/velox/expression/CastExpr-inl.h#L629

bikramSingh91: @natashasehgal I synced up with @kgpai offline and he agrees we need to convert the input to a decimal and use decimal arithmetic when scaling it up to the specified Unit similar to what presto java does.

Can you take a look at the code pointers i mentioned in the previous comments for how you can convert and add decimal types?

natashasehgal: I copied this code. One issue is `toDecimalValue(StringView(str), precision, scale, result)` expects result to be of type int128_t (this results in a loss of precision). There's an error thrown if I try to set result to DecimalType/Long Decimal due from [line](https://github.com/facebookincubator/velox/blob/c6cbeeeaf0f708904b3e693b1a33d36299f6cf56/velox/expression/CastExpr-inl.h#L185)

Any thoughts about converting value to int128_t (e.g. 2.2 -> 22), and rounding down while multiplying? Is overflow a concern here?

cc @karteekmurthys @bikramSingh91

velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,446 @@
...
+  VELOX_USER_CHECK_GT(precision, 0, "Invalid data size: '{}'", decimalString);
+  // TODO - Change parameters use calculated precision, scale. Using output
+  // parameters (38,0) here truncates information after decimal point.
+  std::string valueStr = decimalString.substr(0, precision + 1);
+  LongDecimal<P1, S1> value =
+      parseDecimal<P1, S1>(decimalString, valueStr, precision, scale);
+  // TODO - any better way to convert int128_t to LongDecimal<P1, S1> ?
+  Unit unit = parseUnit(decimalString, precision + 1);
+  int128_t unitValue = static_cast<int128_t>(unit);
+  std::string unitStr = boost::lexical_cast<std::string>(unitValue);
```
natashasehgal, line 374: @karteekmurthys I found DecimalUtil.toString() to convert `int128_t` to std::string in a Velox standard way. 

However, I did not find any function to extract precision and scale from a string.  There's [one](https://github.com/facebookincubator/velox/blob/main/velox/type/Type.cpp#L122) from short or long decimal

velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,104 @@
...
+    return std::stod(value);
+  } catch (const std::exception&) {
+    VELOX_USER_FAIL("Invalid data size: '{}'", dataSize);
+  }
+}
+
+inline int128_t getDecimal(const std::string& dataSize) {
+  size_t valueLength = 0;
+  while (valueLength < dataSize.length() &&
+         (isdigit(dataSize[valueLength]) || dataSize[valueLength] == '.')) {
```
kgpai, line 71: nit: just want to be sure we are always supposed to get ascii digits here right?

velox/functions/prestosql/tests/ParsePrestoDataSizeTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+ protected:
+  void assertVectors(
+      const std::string& expression,
+      const VectorPtr& arg0,
+      const VectorPtr& expected) {
+    auto result = evaluate(expression, makeRowVector({arg0}));
+    test::assertEqualVectors(expected, result);
+  }
+
+  void assertExpression(std::string input, const std::string& outputStr) {
```
kgpai, line 40: nit: you arent asserting a general expression here ; you are asserting the size , maybe renamed this function to assertSize ?

velox/functions/prestosql/tests/ParsePrestoDataSizeTest.cpp
```diff
@@ -0,0 +1,116 @@
...
+  assertExpression("4GB", "4294967296");
+  assertExpression("4TB", "4398046511104");
+  assertExpression("5PB", "5629499534213120");
+  assertExpression("6EB", "6917529027641081856");
+  assertExpression("8YB", "9671406556917033397649408");
+  assertExpression("7ZB", "8264141345021879123968");
+  assertExpression("8YB", "9671406556917033397649408");
+  assertExpression(
+      "6917529027641081856EB", "7975367974709495237422842361682067456");
+  assertExpression(
```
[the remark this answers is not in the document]
natashasehgal: These are the Presto unit test cases. Just verified values on Presto and Prestissimo clusters

velox/functions/prestosql/DataSizeFunctions.h
natashasehgal: Good question. I tried some examples and found Presto implementation is working with non-Ascii as well as combination of non-Ascii and Ascii

Java's Character.isDigit handles non-Ascii

SELECT
    PARSE_PRESTO_DATA_SIZE(x)
FROM (
    VALUES
        ('١٢٣B'),
        ('١٢٣1MB')
) AS t(x);

Response -> 
123
1290797056

natashasehgal: Discussed - We dont support that well in c++ right now for digits . Throw on non ascii for now.

velox/functions/prestosql/DataSizeFunctions.h
```diff
@@ -0,0 +1,111 @@
...
+  FOLLY_ALWAYS_INLINE void callAscii(
+      out_type<LongDecimal<P1, S1>>& result,
+      const arg_type<Varchar>& input) {
+    result = getDecimal(input);
+  }
+  // Fail Non-ASCII characters in input.
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<LongDecimal<P1, S1>>& result,
+      const arg_type<Varchar>& input) {
+    VELOX_USER_FAIL("Invalid data size: '{}'", input);
```
kgpai, line 107: nit: Maybe say non ascii digits are not supported in the failure message.

kgpai (approved):

PR #12427, opened 2025-02-22
fix(tpch): Properly generate rows for lineitem

pedroerp:
Summary:
When using dbgen, the number of lineitems generated for each order is
chosen at random. To be able to control rows ids and generate data in parallel
given the definition on splits, lineitem data is generated based on the number
of associated orders, so it needs special handling logic on the reader. Adding
the required logic so that lineitem data can be generated in parallel as
expected.

Differential Revision: D70029685


xiaoxmeng (approved):
@pedroerp thanks!

PR #12428, opened 2025-02-22
feat(python): Support printPlanWithStats API

pedroerp:
Summary: Add Python API to print the plan executed along with task stats.

Differential Revision: D70029859


velox/py/runner/PyLocalRunner.h
```diff
@@ -54,6 +54,10 @@ class PyLocalRunner {
   /// Execute the task and returns an iterable to the output vectors.
   pybind11::iterator execute();
 
+  /// Prints a descriptive debug message containing plan and execution stats.
+  /// Needs to called after the task has finished.
```
kostasxmeta, line 58: Can we  add here what happens if it is called before the task has finished?

kostasxmeta (approved):

velox/py/runner/PyLocalRunner.h
pedroerp: +1. I added to the comment.

PR #5553, opened 2023-07-07
Add TableWriteMergeNode in Velox

xiaoxmeng:
This is part of multi table writer driver support: [5546](https://github.com/facebookincubator/velox/issues/5546)

PR #12429, opened 2025-02-22
fix: Fix ambiguous `sources` serde logic for `TableWriteNode` and `TableWriteMergeNode`

zhztheplayer:
Related to #5553 and #5176

`PlanNode::serialize()` already does serialization for common member variable `sources` in a `PlanNode`. So serialization code 
 like `sources_.front()->serialize()` is not needed.

This is a benign issue in velox's code base itself because `obj["sources"] = sources_.front()->serialize()` will overwrite `obj["sources"] = serializedSources` (in `PlanNode::serialize()`) so nothing wrong from velox's UT. However a user application / a downstream project may need certain JSON field mapping to make serde integration work correctly. For example, [velox4j](https://github.com/velox4j/velox4j) uses the Jackson JSON library which has harder checks for field ambiguities, and it will raise `Conflicting getter definitions for property "sources"` when I was trying to add a `TableWriteNode` because the definition of [PlanNode](https://github.com/velox4j/velox4j/blob/main/src/main/java/io/github/zhztheplayer/velox4j/plan/PlanNode.java#L20-L22) in velox4j has a fixed `sources` JSON getter already.

The patch is a simple fix by just making `TableWriteNode` / `TableWriteMergeNode`'s serde code align with other plan nodes on the field `sources`.

### DIFF
--- a/velox/core/PlanNode.cpp
+++ b/velox/core/PlanNode.cpp
@@ -2176,7 +2176,6 @@ void TableWriteNode::addDetails(std::stringstream& stream) const {
 
 folly::dynamic TableWriteNode::serialize() const {
   auto obj = PlanNode::serialize();
-  obj["sources"] = sources_.front()->serialize();
   obj["columns"] = columns_->serialize();
   obj["columnNames"] = ISerializable::serialize(columnNames_);
   if (aggregationNode_ != nullptr) {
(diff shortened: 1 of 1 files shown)


xiaoxmeng (approved):

PR #10903, opened 2024-08-30
Allow file name to be supplied to non-bucketing table write node

Yuhta:
Summary:
For some non Presto/Spark use cases, we need to use `PlanBuilder` to
construct plan and supply file name directly to table write.

Differential Revision: D62032439

PR #12431, opened 2025-02-22
feat: Add JSON SerDe code for member field LocationHandle::targetFileName_

zhztheplayer:
JSON SerDe code for the field LocationHandle::targetFileName_ is missing. The patch adds it.

Related to https://github.com/facebookincubator/velox/pull/10903.


xiaoxmeng (approved):

PR #12438, opened 2025-02-24
build: FormatFractionOfSecond: suppress a -Wvla-cxx-extension warning

meyering:
Summary:
This avoids the following errors:

  velox/functions/lib/DateTimeFormatter.cpp:629:14: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]

Reviewed By: dmm-fb

Differential Revision: D70068347


bikramSingh91 (approved):

bikramSingh91 (changes requested):
the mac build seems to fail due to the folly macros, can you apply my previous suggestion of using std::string instead of char?

Issue #12389, opened 2025-02-19
SIGABRT from the JsonParseFunction::sortIndices


kagamiori:
### Description

Expression fuzzer found a SIGABRT from the JsonParseFunction::sortIndices.

### Error Reproduction

On the latest main branch, you can run the following command to reproduce the error.
`expression_fuzzer_test --enable_variadic_signatures --velox_fuzzer_enable_complex_types --lazy_vector_generation_ratio 0.2 --common_dictionary_wraps_generation_ratio=0.3 --velox_fuzzer_enable_column_reuse --velox_fuzzer_enable_expression_reuse --max_expression_trees_per_step 2 --retry_with_try --enable_dereference --duration_sec 1200 --minloglevel=0 --stderrthreshold=2 --only="json_parse,array_constructor" --seed=1136302951`

### Relevant logs

```bash
I0219 10:27:00.563302 3315260 ExpressionFuzzerVerifier.cpp:382] ==============================> Started iteration 0 (seed: 1136302951)
I0219 10:27:00.712445 3315260 ExpressionVerifier.cpp:142] Executing expression 0 : json_parse("c0")
I0219 10:27:00.712646 3315260 ExpressionVerifier.cpp:142] Executing expression 1 : json_parse("c1")
I0219 10:27:00.712694 3315260 ExpressionVerifier.cpp:142] Executing expression 2 : "row_number"
I0219 10:27:00.713757 3315260 ExpressionVerifier.cpp:189] Executing test case: 0
fbcode/third-party-buck/platform010/build/libgcc/include/c++/trunk/string_view:233: const_reference std::basic_string_view<char>::operator[](size_type) const [_CharT = char, _Traits = std::char_traits<char>]: Assertion '__pos < this->_M_len' failed.
*** Aborted at 1739989620 (Unix time, try 'date -d @1739989620') ***
*** Signal 6 (SIGABRT) (0x37b3c0032963c) received by PID 3315260 (pthread TID 0x7f6c61456e40) (linux TID 3315260) (maybe from PID 3315260, UID 228156) (code: -6), stack trace: ***
    @ 000000000000f607 folly::symbolizer::(anonymous namespace)::innerSignalHandler(int, siginfo_t*, void*)
                       ./fbcode/folly/debugging/symbolizer/SignalHandler.cpp:453
    @ 000000000000dd81 folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo_t*, void*)
[...]

PR #12442, opened 2025-02-24
fix(json): Add backslash to unescaped character list

peterenescu:
Summary:
https://github.com/facebookincubator/velox/issues/12389

Fix json_parse error due to lacking switch case for backslash.

Differential Revision: D70132773


kagamiori:
LGTM. Thank you for the fix!

kagamiori (approved):

PR #12444, opened 2025-02-25
feat: Support array_join(Unknown)

xiaodouchen:
Support array_join(Unknown). 

Note: `array_join()` has already supported json type. The CI/CD test
failed and threw an exception "No default Hive type provided for
unsupported Hive type: json" in Presto fuzzer test. For more detailed
information, please refer to https://github.com/facebookincubator/velox/issues/12452. 
We fix it by adding this function in `skipFunctions` and it will lose 
fuzzer coverage.

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -200,6 +200,7 @@ int main(int argc, char** argv) {
         "combine_hash_internal",
         "map_keys_by_top_n_values", // requires
                                     // https://github.com/prestodb/presto/pull/24570
+        "array_join", // https://github.com/facebookincubator/velox/issues/12452
     });
 
     referenceQueryRunner = std::make_shared<PrestoQueryRunner>(
--- a/velox/functions/prestosql/tests/ArrayJoinTest.cpp
+++ b/velox/functions/prestosql/tests/ArrayJoinTest.cpp
@@ -268,4 +268,15 @@ TEST_F(ArrayJoinTest, jsonTest) {
       false,
       true);
 }
+
+TEST_F(ArrayJoinTest, unknownTest) {
+  testArrayJoinNoReplacement<UnknownValue>(
+      {std::nullopt, std::nullopt, std::nullopt}, ","_sv, ""_sv);
+  testArrayJoinReplacement<UnknownValue>(
+      {std::nullopt, std::nullopt, std::nullopt},
+      ","_sv,
+      "apple"_sv,
+      "apple,apple,apple"_sv);
+}
+
 } // namespace
--- a/velox/functions/prestosql/ArrayFunctions.h
+++ b/velox/functions/prestosql/ArrayFunctions.h
@@ -226,6 +226,10 @@ struct ArrayJoinFunction {
     result += inputValue.toString(options_);
   }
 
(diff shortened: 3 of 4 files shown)


xiaodouchen:
@kgpai @laithsakka @mbasmanova Please help review this PR when you have time.

mbasmanova (commented):
@xiaodouchen There are CI failures. Would you take a look?

velox/functions/prestosql/tests/ArrayJoinTest.cpp
```diff
@@ -268,4 +268,12 @@ TEST_F(ArrayJoinTest, jsonTest) {
       false,
       true);
 }
+
+TEST_F(ArrayJoinTest, unknownTest) {
+  testArrayJoinNoReplacement<UnknownValue>(
+      {std::nullopt, std::nullopt, std::nullopt}, ","_sv, ""_sv);
+  testArrayJoinReplacement<UnknownValue>(
```
mbasmanova, line 275: This test case doesn't seem right. I'd expect all NULL values to be replaced with 'apple'.

mbasmanova (changes requested):

xiaodouchen:
> @xiaodouchen There are CI failures. Would you take a look?

`array_join` supports json type and throws exception "No default Hive type provided for unsupported Hive type: json" in Presto fuzzer test. It is a known issue in https://github.com/facebookincubator/velox/pull/12286. We fix it by adding this function in `skipFunctions`.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -198,6 +198,8 @@ int main(int argc, char** argv) {
         "current_date", // Non-deterministic
         "xxhash64_internal",
         "combine_hash_internal",
+        "array_join", // No default Hive type provided for unsupported Hive
```
mbasmanova, line 201: Would you add a link to GitHub issue that provides more details?

It seems that with this change we lose Fuzzer coverage... is there a way to keep the coverage? If not, let's at least call this out in PR description.

CC: @kgpai @kagamiori

xiaodouchen: @xiaodouchen 
1. Yes, added ( https://github.com/facebookincubator/velox/issues/12452)
2. Sorry, I don't know how to keep the coverage. I have mentioned this limitation in the PR description.

mbasmanova: @bikramSingh91 @amitkdutta @kagamiori @kgpai Krishna, Wei, Bikram, would you comment on whether we are Ok with loosing Fuzzer coverage for array_join.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -200,6 +200,7 @@ int main(int argc, char** argv) {
         "combine_hash_internal",
         "map_keys_by_top_n_values", // requires
                                     // https://github.com/prestodb/presto/pull/24570
+        "array_join", // https://github.com/facebookincubator/velox/issues/12452
```
kgpai, line 203: This should be separate PR and not part of this change since it would be good to ensure that your changes are being tested in the fuzzer.

PR #12448, opened 2025-02-26
feat: Support array_intersect(Unknown)

xiaodouchen:


mbasmanova (commented):
@xiaodouchen Would this function run into same / similar limitation in Fuzzer as array_join?

CC: @kgpai @kagamiori

velox/functions/prestosql/ArrayIntersectExcept.cpp
```diff
@@ -622,7 +622,10 @@ std::shared_ptr<exec::VectorFunction> createTypedArraysIntersectExcept(
   VELOX_CHECK_EQ(inputArgs.size(), 2);
   const BaseVector* rhs = inputArgs[1].constantValue.get();
 
-  if (elementType->providesCustomComparison()) {
+  if constexpr (kind == TypeKind::UNKNOWN) {
```
mbasmanova, line 625: UNKNOWN values can only be nulls. Intersecting arrays of NULLs can be done faster by looking at just the array sizes, no?

This change affects array_except function as well? If so, can we add test coverage?

velox/functions/prestosql/tests/ArrayIntersectTest.cpp
```diff
@@ -735,3 +735,50 @@ TEST_F(ArrayIntersectTest, nullInDictionary) {
       makeNullableArrayVector<int64_t>({{{1}}, std::nullopt, {{7}}});
   testExpr(expected, "array_intersect(c0)", {arrays});
 }
+
+TEST_F(ArrayIntersectTest, unknownArrays) {
+  auto array1 = makeNullableArrayVector<UnknownValue>(
```
mbasmanova, line 740: This is a bit hard to read. Is there a way to extend makeArrayVectorFromJson to work with Unknown?

mbasmanova (changes requested):

Issue #12450, opened 2025-02-26
Build on arm64: undefined reference to `__folly_memcpy_aarch64'


zhztheplayer:
### Problem description

The following error occurs when building Velox on GitHub-hosted arm64 runner:

```
2025-02-25T14:44:28.9768977Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memcpy_aarch64'
2025-02-25T14:44:28.9775911Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memmove_aarch64'
2025-02-25T14:44:28.9790740Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memcpy_aarch64_sve'
2025-02-25T14:44:28.9831022Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memcpy_aarch64_simd'
2025-02-25T14:44:28.9843564Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memmove_aarch64_simd'
2025-02-25T14:44:28.9870386Z ld: /home/runner/work/velox4j/velox4j/src/main/cpp/build/dist/lib/libfolly.so.0.58.0-dev: undefined reference to `__folly_memmove_aarch64_sve'
```

See [velox4j's CI failure](https://github.com/velox4j/velox4j/actions/runs/13522598432/job/37785138937?pr=66).

This is related to a known issue in folly: https://github.com/facebook/folly/issues/2251.

The folly issue was resolved in https://github.com/facebook/folly/commit/c30d49dcdc877b38d99b253b8c66ad1853085e09 which was included in folly release [v2024.07.15.00](https://github.com/facebook/folly/releases/tag/v2024.07.15.00). While currently Velox depends on [v2024.07.01.00](https://github.com/facebook/folly/releases/tag/v2024.07.01.00). A FBOS version bump may solve this Velox issue.

### System information

GitHub hosted runner `ubuntu-24.04-arm`

### CMake log

```bash

```

zhztheplayer:
[...]

PR #12451, opened 2025-02-26
fix: Update FBOS to v2024.07.15 to fix arm64 + Linux build

zhztheplayer:
Fixes https://github.com/facebookincubator/velox/issues/12450


zhztheplayer:
Confirmed to fix #12450.

zhztheplayer:
[CI Error](https://github.com/facebookincubator/velox/actions/runs/13535987189/job/37827677443?pr=12451) seems related to https://github.com/facebook/folly/commit/581e09a3a1a454167bc5f65ac408d85394915837

zhztheplayer:
The folly patch will no longer be needed after bumping to version >= [v2024.08.19.00](https://github.com/facebook/folly/releases/tag/v2024.08.19.00) in future.

czentgr:
The version of folly was picked to support both C++17 and C++20 standard. At the time, using a newer folly caused compatibility issues where you couldn't switch between c++17 and c++20 standard and have all tests pass.
I'm working on the c++20 support but got bogged down with compilation issues on Linux with missing symbols in the iceberg code.

zhztheplayer:
> The version of folly was picked to support both C++17 and C++20 standard. At the time, using a newer folly caused compatibility issues where you couldn't switch between c++17 and c++20 standard and have all tests pass. I'm working on the c++20 support but got bogged down with compilation issues on Linux with missing symbols in the iceberg code.

@czentgr Thank you for the inputs. I could expect that there was good reason to stay in v2024.07.01 so the PR is totally optional for the community. 

> The version of folly was picked to support both C++17 and C++20 standard. 

Given this is a small bump (v2024.07.01 -> v2024.07.15), I assume it's not complicated to have the both standards supported in 07.15? But perhaps it's not worth it since I already added a folly patch in this PR.

Another option is to add the [patch](https://github.com/facebook/folly/commit/c30d49dcdc877b38d99b253b8c66ad1853085e09) to current version of folly code directly, without version bumping. Would you think that's feasible?

czentgr:
@zhztheplayer Thanks. The problem with a patch is that it applies only to the bundled version of folly. But most deployments will use the system version (aka the one from the setup scripts). Unless the fix-thread-local.patch patch is applied at the setup script level as well after folly has been downloaded. 

The thing about the folly versions was that I tried to find a version that worked for both at the same time. The proposed version also had the odd issue. I tried to find the most recent version that worked for both and this was it.
I would hope new versions of folly actually work with both c++ versions too.

PR #12456, opened 2025-02-26
Fix DateTimeFormatter to set weekDateFormat only if it has week and day of week along with year.

kgpai:
Summary:
This is a targeted fix that ensures that date_parse formats which contain %a but without additional specifiers like %v does not set weekDateFormat. An example of such a format is  '%d-%b-%Y (%a)' which when parsing a string like '11-Mar-2024 (Mon)' will incorrectly ignore the month and day and instead return a date like '2024-01-01' . 

There are likely other cases where additional or inconsistent specifiers are handled incorrectly. A full fix of DateTimeFormatter is beyond scope of this PR, but I will send out additional PR's in the future to address these issues. 

Mitigates S493660.

Differential Revision: D70253624


velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -1637,6 +1638,10 @@ Expected<DateTimeResult> DateTimeFormatter::parse(
 
   // Convert the parsed date/time into a timestamp.
   Expected<int64_t> daysSinceEpoch;
+
+  // Ensure you use week date format only when you have year and atleast week.
```
pedroerp, line 1642: nit: "at least"

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -4530,6 +4530,12 @@ TEST_F(DateTimeFunctionsTest, dateParse) {
   EXPECT_EQ(
       Timestamp(-66600, 0), dateParse("1969-12-31+11:00", "%Y-%m-%d+%H:%i"));
 
+  setQueryTimeZone("America/Los_Angeles");
+  // Tests if it uses weekdateformat if %v not present but %a is present.
+  EXPECT_EQ(
+      Timestamp(1730707200, 0),
+      dateParse("04-Nov-2024 (Mon)", "%d-%b-%Y (%a)"));
```
pedroerp, line 4537: does Joda throw is the day of the week doesn't match the rest of the date being parsed?

pedroerp (approved):
Thanks, looks good to me!

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
kgpai: No joda doesnt throw , the behavior depends on the specifier  ; Its why we need a fuzzer to ensure we are in sync and have behavior similar to that shown by Joda.

PR #12457, opened 2025-02-26
misc: Let VectorStream::children_ use StlAllocator

tanjialiang:
Summary: For wide structure (thousands of columns) shuffle, the basic std structures quickly sucks in large amount of memory directly from uncontrolled Jemalloc to cause server OOM crash. Absorb these memory to velox controlled memory to avoid this situation.

Differential Revision: D70130662


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #12460, opened 2025-02-26
refactor: Clean up Registration includes from Presto Type headers

kevinwilfong:
Summary:
This is a follow up to https://github.com/facebookincubator/velox/pull/12393

The header files in Presto were fixed in https://github.com/prestodb/presto/pull/24622 so we are
now free to separate the Type and Registration files so that Registration depends on Type and
not the other way around.

Differential Revision: D70275631


kagamiori (approved):
LGTM. Thank you!

PR #12462, opened 2025-02-27
feat(python): Support for MergeSort

pedroerp:
Summary: Adding support for mege sort in PyVelox plan builder API.

Differential Revision: D70294152

### DIFF
--- a/velox/py/plan_builder/PyPlanBuilder.cpp
+++ b/velox/py/plan_builder/PyPlanBuilder.cpp
@@ -282,6 +282,22 @@ PyPlanBuilder& PyPlanBuilder::mergeJoin(
   return *this;
 }
 
+PyPlanBuilder& PyPlanBuilder::sortedMerge(
+    const std::vector<std::string>& keys,
+    const std::vector<std::optional<PyPlanNode>>& pySources) {
+  std::vector<core::PlanNodePtr> sources;
+  sources.reserve(pySources.size());
+
+  for (const auto& pySource : pySources) {
+    if (pySource.has_value()) {
+      sources.push_back(pySource->planNode());
+    }
+  }
+
+  planBuilder_.localMerge(keys, std::move(sources));
+  return *this;
+}
+
 PyPlanBuilder& PyPlanBuilder::tpchGen(
     const std::string& tableName,
     const std::vector<std::string>& columns,
--- a/velox/py/plan_builder/PyPlanBuilder.h
+++ b/velox/py/plan_builder/PyPlanBuilder.h
@@ -213,6 +213,15 @@ class PyPlanBuilder {
       const std::string& filter,
       core::JoinType joinType);
 
+  /// Takes N sorted `source` subtrees and merges them into a sorted output.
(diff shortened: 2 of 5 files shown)


velox/py/plan_builder/PyPlanBuilder.cpp
```diff
@@ -282,6 +282,22 @@ PyPlanBuilder& PyPlanBuilder::mergeJoin(
   return *this;
 }
 
+PyPlanBuilder& PyPlanBuilder::mergeSort(
```
kostasxmeta, line 285: nit: IMO this name is a bit confusing because the data is already sorted and we are just merging them. we aren't doing any sorting per-se. Something like sortedMerge would be more descriptive. I'll leave it up to you though.

velox/py/plan_builder/PyPlanBuilder.cpp
```diff
@@ -282,6 +282,22 @@ PyPlanBuilder& PyPlanBuilder::mergeJoin(
...
+    const std::vector<std::string>& keys,
+    const std::vector<std::optional<PyPlanNode>>& pySources) {
+  std::vector<core::PlanNodePtr> sources;
+  sources.reserve(pySources.size());
+
+  for (const auto& pySource : pySources) {
+    if (pySource.has_value()) {
```
kostasxmeta, line 292: In what situation would anyone pass in an empty PyPlanNode? Wondering why this is a vector of optionals.

velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -213,6 +213,15 @@ class PyPlanBuilder {
       const std::string& filter,
       core::JoinType joinType);
 
+  /// Takes N sorted `source` subtrees and merges them into a sorted output.
+  /// Assumes that all sources are sorted in `keys`.
```
kostasxmeta, line 217:
```suggestion
  /// Assumes that all sources are sorted on `keys`.
```

velox/py/plan_builder/plan_builder.cpp
```diff
@@ -207,6 +207,19 @@ PYBIND11_MODULE(plan_builder, m) {
...
           filter: Optional join filter expression.
       )"))
+      .def(
+          "merge_sort",
+          &velox::py::PyPlanBuilder::mergeSort,
+          py::arg("keys"),
+          py::arg("sources"),
+          py::doc(R"(
+        Takes N sorted `source` subtrees and merges them into a sorted output.
+        Assumes that all sources are sorted in `keys`.
```
kostasxmeta, line 217:
```suggestion
        Assumes that all sources are sorted on `keys`.
```

kostasxmeta (approved):

velox/py/plan_builder/PyPlanBuilder.cpp
pedroerp: That is because `plan_builder.get_plan_node()` returns an optional. It could be done if the plan builder is empty (if no nodes were added yet).

PR #12463, opened 2025-02-27
fix: Add a patch to fix folly build on arm64 + Linux

zhztheplayer:
Chery-pick [this change](https://github.com/facebook/folly/commit/c30d49dcdc877b38d99b253b8c66ad1853085e09) as a folly patch to fix https://github.com/facebookincubator/velox/issues/12450 (Linux + arm64 build).

This is an alternative solution to https://github.com/facebookincubator/velox/pull/12451.

### DIFF
--- a/CMake/resolve_dependency_modules/folly/CMakeLists.txt
+++ b/CMake/resolve_dependency_modules/folly/CMakeLists.txt
@@ -38,8 +38,10 @@ FetchContent_Declare(
   folly
   URL ${VELOX_FOLLY_SOURCE_URL}
   URL_HASH ${VELOX_FOLLY_BUILD_SHA256_CHECKSUM}
-  PATCH_COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/folly-no-export.patch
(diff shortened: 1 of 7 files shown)


bdice:
@zhztheplayer Should I close #12903 in favor of this? Can we finalize this PR now, or are there still blockers? (Apologies if this has been answered, I've lost track.)

zhztheplayer:
Hi @bdice thanks for following up. 

Velox's CI docker images has some `COPY` statements only on the setup scripts, so the patch I add in this PR will not be brought into the docker, so file-not-found errors will happen in CI. Currently I don't have a ideal solution for this except we copy the patch into the images as well...

Should we wait for https://github.com/facebookincubator/velox/pull/13016 at first? Or else I can also update the PR with a solution as I mentioned.

bdice:
That is fine. It looks like @czentgr is actively working on #13016. Thanks to both of you for the effort!

czentgr:
We are getting closer with #13016 . It is under review and CI was clean. I need to make at least one more minor change to separate out changes but otherwise should be good to go after that if approved.

zhztheplayer:
Thank you @bdice @czentgr

PR #12464, opened 2025-02-27
feat: Register Spark sign function

zhli1142015:
Returns -1.0, 0.0 or 1.0 as input is negative, 0 or positive.

Spark's implementation: https://github.com/apache/spark/blob/v3.5.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L758

### DIFF
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -266,7 +266,7 @@ Mathematical Functions
 
     Returns the secant of ``x``.
 
-.. function:: sign(x) -> double
+.. spark:function:: sign(x) -> double
 
     Returns the signum function of ``x``. It returns:
     * 0.0 if the argument is 0.0,
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -268,7 +268,7 @@ Mathematical Functions
 
 .. spark:function:: sign(x) -> double
 
-    Returns the signum function of ``x``. Supported type for ``x`` is DOUBLE.
+    Returns the signum of ``x``. Supported type for ``x`` is DOUBLE.
     It returns:
     * 0.0 if the argument is 0.0,
     * 1.0 if the argument is greater than 0.0,
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -266,6 +266,16 @@ Mathematical Functions
 
     Returns the secant of ``x``.
 
+.. function:: sign(x) -> double
+
+    Returns the signum function of ``x``. It returns:
+    * 0.0 if the argument is 0.0,
+    * 1.0 if the argument is greater than 0.0,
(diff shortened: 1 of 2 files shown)


zhli1142015:
cc @rui-mo @jinchengchenghh , thanks.

velox/docs/functions/spark/math.rst
```diff
@@ -266,6 +266,16 @@ Mathematical Functions
 
     Returns the secant of ``x``.
 
+.. function:: sign(x) -> double
```
jinchengchenghh, line 269: spark:function::

velox/docs/functions/spark/math.rst
```diff
@@ -266,6 +266,16 @@ Mathematical Functions
...
+
+    Returns the signum function of ``x``. It returns:
```
jinchengchenghh, line 271: We can copy the presto function document
```
.. function:: sign(x) -> [same as x]

    Returns the signum function of ``x``. For both integer and floating point arguments, it returns:
    * 0 if the argument is 0,
    * 1 if the argument is greater than 0,
    * -1 if the argument is less than 0.

    For double arguments, the function additionally return:
    * NaN if the argument is NaN,
    * 1 if the argument is +Infinity,
    * -1 if the argument is -Infinity.
```

zhli1142015: We can't use this directly, in Spark this function only supports double and internal types.

jinchengchenghh: In that case, we should add the Supported types for ``x`` are Double

velox/docs/functions/spark/math.rst
```diff
@@ -266,6 +266,17 @@ Mathematical Functions
 
     Returns the secant of ``x``.
 
+.. spark:function:: sign(x) -> double
+
+    Returns the signum function of ``x``. Supported type for ``x`` is DOUBLE.
+    It returns:
+    * 0.0 if the argument is 0.0,
```
rui-mo, line 273: Is there any difference on +0.0 and -0.0 for Spark?

zhli1142015: No, both return 0.0

velox/docs/functions/spark/math.rst
```diff
@@ -266,6 +266,17 @@ Mathematical Functions
 
     Returns the secant of ``x``.
 
+.. spark:function:: sign(x) -> double
+
+    Returns the signum function of ``x``. Supported type for ``x`` is DOUBLE.
```
rui-mo, line 271: > signum function

remove "function"?

rui-mo (approved):
Would you run the fuzzer test to make sure it works? Thanks.

rui-mo:
Screenshot 2025-03-03 at 09 05 39

The document does not render correctly for `*`, would you please take a look?

zhli1142015:
Fixed, thanks.

PR #12466, opened 2025-02-27
fix: Untangle some strange code in `PlanBuilder::finalAggregation()`

zhztheplayer:
The problematic code was brought in #1419.

The code block is actually 

```cpp
if (!a) {
  VELOX_CHECK(a);
}
```

which is apparently wrong. The code was intentionally to find an aggregation with `Step::kPartial`. The patch recovers the right 
 readable logic.

### DIFF
--- a/velox/exec/tests/utils/PlanBuilder.cpp
+++ b/velox/exec/tests/utils/PlanBuilder.cpp
@@ -760,14 +760,10 @@ PlanBuilder& PlanBuilder::finalAggregation() {
   if (!exec::isRawInput(aggNode->step())) {
     // If aggregation node is not the partial aggregation, keep looking again.
     aggNode = findPartialAggregation(aggNode->sources()[0].get());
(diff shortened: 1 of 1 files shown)


velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -760,15 +760,11 @@ PlanBuilder& PlanBuilder::finalAggregation() {
   if (!exec::isRawInput(aggNode->step())) {
     // If aggregation node is not the partial aggregation, keep looking again.
     aggNode = findPartialAggregation(aggNode->sources()[0].get());
-    if (!exec::isRawInput(aggNode->step())) {
```
xiaoxmeng, line 763: Can you keep VELOX_CHECK_NOT_NULL?

xiaoxmeng (commented):

xiaoxmeng (approved):

PR #12470, opened 2025-02-27
feat: Add printPlanWithStats and planFragment APIs for Task

majetideepak:
Client such as Prestissimo create the Task with planNode and access the Task
 at various points. The planNode is not accessible everywhere.
Since the planNode is stored in the Task, it helps to add printPlanWithStats API to Task.

### DIFF
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -2435,11 +2435,9 @@ ContinueFuture Task::taskDeletionFuture() {
 }
 
 std::string Task::printPlanWithStats(bool includeCustomStats) const {
-  if (planFragment_.planNode) {
-    return exec::printPlanWithStats(
-        *planFragment_.planNode, taskStats_, includeCustomStats);
-  }
-  return "";
+  VELOX_CHECK_NOT_NULL(planFragment_.planNode, "PlanFragment has no planNode");
+  return exec::printPlanWithStats(
+      *planFragment_.planNode, taskStats_, includeCustomStats);
 }
 
 std::string Task::toString() const {
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -30,6 +30,7 @@
 #include "velox/exec/NestedLoopJoinBuild.h"
 #include "velox/exec/OperatorUtils.h"
 #include "velox/exec/OutputBufferManager.h"
+#include "velox/exec/PlanNodeStats.h"
 #include "velox/exec/Task.h"
 #include "velox/exec/TraceUtil.h"
 
@@ -2431,6 +2432,14 @@ ContinueFuture Task::taskDeletionFuture() {
       fmt::format("Task::taskDeletionFuture {}", taskId_));
   taskDeletionPromises_.emplace_back(std::move(promise));
   return std::move(future);
+}
+
+std::string Task::printPlanWithStats(bool includeCustomStats) const {
+  if (planFragment_.planNode) {
+    return exec::printPlanWithStats(
+        *planFragment_.planNode, taskStats_, includeCustomStats);
+  }
+  return "";
 }
 
 std::string Task::toString() const {
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -149,9 +149,7 @@ std::unordered_map<core::PlanNodeId, SplitsState> buildSplitStates(
     const std::shared_ptr<const core::PlanNode>& planNode) {
   std::unordered_set<core::PlanNodeId> allIds;
   std::unordered_map<core::PlanNodeId, SplitsState> splitStateMap;
-  if (planNode.get() != nullptr) {
-    buildSplitStates(planNode.get(), allIds, splitStateMap);
-  }
+  buildSplitStates(planNode.get(), allIds, splitStateMap);
   return splitStateMap;
 }
 
@@ -249,6 +247,7 @@ std::shared_ptr<Task> Task::create(
     Consumer consumer,
     int32_t memoryArbitrationPriority,
     std::function<void(std::exception_ptr)> onError) {
+  VELOX_CHECK_NOT_NULL(planFragment.planNode);
   return Task::create(
       taskId,
       std::move(planFragment),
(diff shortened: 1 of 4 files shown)


xiaoxmeng (approved):
@majetideepak LGTM. Thanks!

velox/exec/Task.h
```diff
@@ -116,6 +116,8 @@ class Task : public std::enable_shared_from_this<Task> {
     spillDirectoryCallback_ = std::move(spillDirectoryCallback);
   }
 
+  std::string printPlanWithStats(bool includeCustomStats = false) const;
```
mbasmanova, line 119: Would you document this API?

velox/exec/Task.cpp
```diff
@@ -2433,6 +2434,14 @@ ContinueFuture Task::taskDeletionFuture() {
   return std::move(future);
 }
 
+std::string Task::printPlanWithStats(bool includeCustomStats) const {
+  if (planFragment_.planNode) {
+    return exec::printPlanWithStats(
+        *planFragment_.planNode, taskStats_, includeCustomStats);
+  }
+  return "";
```
mbasmanova, line 2442: Users of the API may be surprised to get an empty string? When would this happen? Should this be an error?

velox/exec/Task.cpp
```diff
@@ -2433,6 +2434,14 @@ ContinueFuture Task::taskDeletionFuture() {
   return std::move(future);
 }
 
+std::string Task::printPlanWithStats(bool includeCustomStats) const {
+  if (planFragment_.planNode) {
```
mbasmanova, line 2438: Should we check Task state and fail if task is not running or completed?

mbasmanova (changes requested):

velox/exec/Task.cpp
majetideepak: A task state check is not required. On Prestissimo, I am using this on task completion. I would also think this is helpful on Task failure for debugging.

velox/exec/Task.h
majetideepak: Done!

majetideepak:
@mbasmanova I addressed your comments. Thanks for the review!

velox/exec/Task.cpp
```diff
@@ -2433,6 +2434,12 @@ ContinueFuture Task::taskDeletionFuture() {
   return std::move(future);
 }
 
+std::string Task::printPlanWithStats(bool includeCustomStats) const {
+  VELOX_CHECK_NOT_NULL(planFragment_.planNode, "PlanFragment has no planNode");
```
mbasmanova, line 2438: Would you add some tests that trigger this check and tests that invoke this API when task is in different states (before start, in progress, failed, succeeded)?

majetideepak: @mbasmanova I added some tests to address this. Thanks.

majetideepak: TaskTest.cpp also covers TaskStats update during Task cancel. This API only depends on planNode and TaskStats. We should have good test coverage now.

velox/exec/tests/PrintPlanWithStatsTest.cpp
```diff
@@ -375,3 +376,52 @@ TEST_F(PrintPlanWithStatsTest, tableWriterWithTableScan) {
        {"        totalRemainingFilterTime\\s+sum: .+, count: .+, min: .+, max: .+"},
        {"        totalScanTime    [ ]* sum: .+, count: .+, min: .+, max: .+"}});
 }
+
+TEST_F(PrintPlanWithStatsTest, taskAPI) {
+  // Test nullptr PlanNode.
+  core::PlanFragment planFragment;
+  auto taskWithNoPlanNode = exec::Task::create(
```
mbasmanova, line 383: Should Task::create API fail if there is no plan? Are we missing such check?

velox/exec/tests/PrintPlanWithStatsTest.cpp
```diff
@@ -375,3 +376,52 @@ TEST_F(PrintPlanWithStatsTest, tableWriterWithTableScan) {
...
+        ::testing::UnitTest::GetInstance()->current_test_info()->name(),
+        task->printPlanWithStats(),
+        {{"-- Aggregation\\[1\\]\\[SINGLE \\[c0\\] a0 := sum\\(ROW\\[\"c1\"\\]\\)\\] -> c0:BIGINT, a0:BIGINT"},
+         {"   Output: .+, Cpu time: .+, Wall time: .+, Blocked wall time: .+, Peak memory: .+, Memory allocations: .+, CPU breakdown: B/I/O/F (.+/.+/.+/.+)"},
+         {"  -- TableScan\\[0\\]\\[table: hive_table\\] -> c0:BIGINT, c1:BIGINT"},
+         {"     Input: .+, Output: .+, Cpu time: .+, Wall time: .+, Blocked wall time: .+, Peak memory: .+, Memory allocations: .+, CPU breakdown: B/I/O/F (.+/.+/.+/.+)"}});
+  };
+
+  const auto data = makeRowVector({
+      makeFlatVector<int64_t>(50, [](auto row) { return row; }),
```
mbasmanova, line 404: nit: folly::identity

mbasmanova (commented):
> The planNode is not accessible everywhere.
> Since the planNode is stored in the Task

Should we add a getter for the plan node to Task than?

velox/exec/tests/PrintPlanWithStatsTest.cpp
majetideepak: There is no check currently. We can add a check and remove planNode checks everywhere.
I was not sure if there was a such use case inside Meta. Looks like not.

majetideepak:
> Should we add a getter for the plan node to Task than?

This is another option. I felt `task->printPlanWithStats` was more user-friendly.

velox/exec/tests/PrintPlanWithStatsTest.cpp
mbasmanova: > We can add a check in Task::create and remove planNode checks everywhere.

That would be nice. I assume Task created with no plan is not usable.

majetideepak: Should `struct PlanFragment` have this check instead?

majetideepak:
@mbasmanova I added `planFragment()` API since that is the input to create::Task. Thanks.

mbasmanova (approved):
@majetideepak Thank you for the improvements.

PR #12474, opened 2025-02-27
feat: Add special form expression optimizations

pramodsatya:
Optimizes special form expressions `AND`, `IF`, `COALESCE`, and `IF` during compilation. 

When any of the inputs to `AND` conjunct expression is `false` or has been constant folded to `false` during compilation, it will be simplified to `false`. Similarly, the result of `OR` conjunct expression will be simplified to `true` if any of its inputs is `true` or has been constant folded to `true` during compilation. For these cases, the `ConjunctExpr` will be simplified to a `ConstantExpr` during compilation. 
The inputs to conjunct expressions `AND` and `OR` that are or have been constant folded to `true` and `false` respectively will be removed from the expression’s inputs since they will not affect the result. If there is only one input expression left after pruning such redundant inputs, the `ConjunctExpr` will be simplified to this input expression. Pruning inputs will simplify the evaluation of the conjunct expression. 

Any inputs to `COALESCE` that are `NULL` will be removed so they are skipped during expression evaluation. The `COALESCE` expression will also be simplified to the first non `NULL` constant input if all inputs seen before were `NULL`. 
`IF` conditional expression has three inputs: `IF(condition, value1, value2)`. If the condition is or has been constant folded to a boolean value during compilation, the conditional expression will be simplified to `value1` or `value2`.

### DIFF
--- a/velox/expression/CoalesceExpr.cpp
+++ b/velox/expression/CoalesceExpr.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "velox/expression/CoalesceExpr.h"
+#include "velox/expression/ConstantExpr.h"
 
 namespace facebook::velox::exec {
 
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
   return CoalesceExpr::resolveType(argTypes);
 }
 
+ExprPtr CoalesceCallToSpecialForm::optimize(
+    std::vector<ExprPtr>& compiledChildren) {
+  auto numChildren = compiledChildren.size();
+  std::vector<column_index_t> nullIndices;
+  for (auto i = 0; i < numChildren; i++) {
+    if (auto constantExpr = std::dynamic_pointer_cast<exec::ConstantExpr>(
+            compiledChildren[i])) {
+      if (constantExpr->value()->isNullAt(0)) {
+        nullIndices.push_back(i);
+      } else if (nullIndices.size() == i) {
+        return constantExpr;
+      }
+    }
+  }
+
+  if (!nullIndices.empty()) {
+    if (nullIndices.size() == compiledChildren.size()) {
+      auto nullChild = compiledChildren.at(nullIndices.front());
+      return std::dynamic_pointer_cast<exec::ConstantExpr>(nullChild);
+    } else {
+      auto nullCount = nullIndices.size();
+      for (vector_size_t j = nullCount - 1; j >= 0; j--) {
+        compiledChildren.erase(compiledChildren.begin() + nullIndices[j]);
+      }
+    }
+  }
+  return nullptr;
+}
+
 ExprPtr CoalesceCallToSpecialForm::constructSpecialForm(
     const TypePtr& type,
     std::vector<ExprPtr>&& compiledChildren,
     bool /* trackCpuUsage */,
     const core::QueryConfig& /*config*/) {
+  auto children = std::move(compiledChildren);
+  if (auto result = optimize(children)) {
+    return result;
+  }
+
   bool inputsSupportFlatNoNullsFastPath =
-      Expr::allSupportFlatNoNullsFastPath(compiledChildren);
+      Expr::allSupportFlatNoNullsFastPath(children);
   return std::make_shared<CoalesceExpr>(
-      type, std::move(compiledChildren), inputsSupportFlatNoNullsFastPath);
+      type, std::move(children), inputsSupportFlatNoNullsFastPath);
 }
 } // namespace facebook::velox::exec
--- a/velox/expression/CoalesceExpr.h
+++ b/velox/expression/CoalesceExpr.h
@@ -53,6 +53,9 @@ class CoalesceCallToSpecialForm : public FunctionCallToSpecialForm {
       std::vector<ExprPtr>&& compiledChildren,
       bool trackCpuUsage,
       const core::QueryConfig& config) override;
+
+ private:
+  ExprPtr optimize(std::vector<ExprPtr>& compiledChildren);
 };
 
 } // namespace facebook::velox::exec
--- a/velox/expression/ConjunctExpr.cpp
+++ b/velox/expression/ConjunctExpr.cpp
@@ -15,6 +15,7 @@
  */
 #include "velox/expression/ConjunctExpr.h"
 #include "velox/expression/BooleanMix.h"
+#include "velox/expression/ConstantExpr.h"
 #include "velox/expression/FieldReference.h"
 #include "velox/expression/ScopedVarSetter.h"
 
@@ -415,18 +416,89 @@ TypePtr ConjunctCallToSpecialForm::resolveType(
   return ConjunctExpr::resolveType(argTypes);
 }
 
+ExprPtr ConjunctCallToSpecialForm::optimizeAnd(
+    std::vector<ExprPtr>& compiledChildren) {
+  auto numChildren = compiledChildren.size();
+  for (vector_size_t i = numChildren - 1; i >= 0; i--) {
+    if (auto constantExpr = std::dynamic_pointer_cast<exec::ConstantExpr>(
+            compiledChildren[i])) {
+      if (!constantExpr->value()->isNullAt(0)) {
+        if (!constantExpr->value()->as<ConstantVector<bool>>()->valueAt(0)) {
+          return constantExpr;
+        } else {
+          if (i == 0 && compiledChildren.size() == 1) {
+            return constantExpr;
+          } else {
+            compiledChildren.erase(compiledChildren.begin() + i);
+          }
+        }
+      }
+    }
+  }
+
+  if (compiledChildren.size() == 1) {
+    return compiledChildren.front();
+  }
+  return nullptr;
+}
+
+ExprPtr ConjunctCallToSpecialForm::optimizeOr(
+    std::vector<ExprPtr>& compiledChildren) {
+  auto numChildren = compiledChildren.size();
+  for (vector_size_t i = numChildren - 1; i >= 0; i--) {
+    if (auto constantExpr = std::dynamic_pointer_cast<exec::ConstantExpr>(
+            compiledChildren[i])) {
+      if (!constantExpr->value()->isNullAt(0)) {
+        if (constantExpr->value()->as<ConstantVector<bool>>()->valueAt(0)) {
+          return constantExpr;
+        } else {
+          if (i == 0 && compiledChildren.size() == 1) {
+            return constantExpr;
+          } else {
+            compiledChildren.erase(compiledChildren.begin() + i);
+          }
+        }
+      }
+    }
+  }
+
+  if (compiledChildren.size() == 1) {
+    return compiledChildren.front();
+  }
+  return nullptr;
+}
+
+// During expression compilation, the inputs to 'AND', 'OR' conjunct expressions
+// could be constant folded to 'true' or 'false'. For instance, the inputs to
+// expression '(1 = 2) and (2 != 3)' are compiled with constant folding enabled
+// to 'false' and 'true'. When a conjunct expression is compiled, certain
+// optimizations are possible when the inputs are constants. When 'AND' conjunct
+// expression has 'false' as an input, it is simplified to a constant boolean
+// expression with value 'false'. When 'OR' conjunct expression has 'true' as an
+// input, it is simplified to a constant boolean expression with value 'true'.
+// Similarly, inputs to 'AND' and 'OR' conjunct expressions that are 'true' or
+// 'false' respectively will be pruned from the conjunct expression. If the
+// conjunct expression has only one input after pruning these constants, this
+// input expression will be returned.
 ExprPtr ConjunctCallToSpecialForm::constructSpecialForm(
     const TypePtr& type,
     std::vector<ExprPtr>&& compiledChildren,
     bool /* trackCpuUsage */,
     const core::QueryConfig& /*config*/) {
-  bool inputsSupportFlatNoNullsFastPath =
-      Expr::allSupportFlatNoNullsFastPath(compiledChildren);
+  auto children = std::move(compiledChildren);
+  if (isAnd_) {
+    if (auto result = optimizeAnd(children)) {
+      return result;
+    }
+  } else {
+    if (auto result = optimizeOr(children)) {
+      return result;
+    }
+  }
 
+  bool inputsSupportFlatNoNullsFastPath =
+      Expr::allSupportFlatNoNullsFastPath(children);
   return std::make_shared<ConjunctExpr>(
-      type,
-      std::move(compiledChildren),
-      isAnd_,
-      inputsSupportFlatNoNullsFastPath);
+      type, std::move(children), isAnd_, inputsSupportFlatNoNullsFastPath);
 }
 } // namespace facebook::velox::exec
--- a/velox/expression/tests/ExprTest.cpp
+++ b/velox/expression/tests/ExprTest.cpp
@@ -343,12 +343,12 @@ class ExprTest : public testing::Test, public VectorTestBase {
 
   std::exception_ptr assertError(
       const std::string& expression,
-      const VectorPtr& input,
+      const RowVectorPtr& input,
       const std::string& context,
       const std::string& additionalContext,
       const std::string& message) {
     try {
-      evaluate(expression, makeRowVector({input}));
+      evaluate(expression, input);
       EXPECT_TRUE(false) << "Expected an error";
     } catch (VeloxException& e) {
       EXPECT_EQ(context, trimInputPath(e.context()));
@@ -2541,7 +2541,7 @@ struct AlwaysThrowsStdExceptionFunction {
 
 /// Verify exception context for the case when function throws std::exception.
 TEST_P(ParameterizedExprTest, stdExceptionContext) {
-  auto data = makeFlatVector<int64_t>({1, 2, 3});
+  auto data = makeRowVector({makeFlatVector<int64_t>({1, 2, 3})});
 
   registerFunction<AlwaysThrowsStdExceptionFunction, int64_t, int64_t>(
       {"throw_invalid_argument"});
@@ -2956,14 +2956,14 @@ TEST_P(ParameterizedExprTest, tryWithConstantFailure) {
 TEST_P(ParameterizedExprTest, castExceptionContext) {
   assertError(
       "cast(c0 as bigint)",
-      makeFlatVector<std::string>({"1a"}),
+      makeRowVector({makeFlatVector<std::string>({"1a"})}),
       "Top-level Expression: cast((c0) as BIGINT)",
       "",
       "Cannot cast VARCHAR '1a' to BIGINT. Non-whitespace character found after end of conversion: \"\"");
 
   assertError(
       "cast(c0 as timestamp)",
-      makeFlatVector(std::vector<int8_t>{1}),
+      makeRowVector({makeFlatVector(std::vector<int8_t>{1})}),
       "Top-level Expression: cast((c0) as TIMESTAMP)",
       "",
       "Cannot cast TINYINT '1' to TIMESTAMP. Conversion to Timestamp is not supported");
@@ -2972,14 +2972,15 @@ TEST_P(ParameterizedExprTest, castExceptionContext) {
 TEST_P(ParameterizedExprTest, switchExceptionContext) {
   assertError(
       "case c0 when 7 then c0 / 0 else 0 end",
-      makeFlatVector(std::vector<int64_t>{7}),
+      makeRowVector({makeFlatVector(std::vector<int64_t>{7})}),
       "divide(c0, 0:BIGINT)",
       "Top-level Expression: switch(eq(c0, 7:BIGINT), divide(c0, 0:BIGINT), 0:BIGINT)",
       "division by zero");
 }
 
 TEST_P(ParameterizedExprTest, conjunctExceptionContext) {
-  auto data = makeFlatVector<int64_t>(20, [](auto row) { return row; });
+  auto data = makeRowVector(
+      {makeFlatVector<int64_t>(20, [](auto row) { return row; })});
 
   assertError(
       "if (c0 % 409 < 300 and c0 / 0 < 30, 1, 2)",
@@ -2990,12 +2991,12 @@ TEST_P(ParameterizedExprTest, conjunctExceptionContext) {
 }
 
 TEST_P(ParameterizedExprTest, lambdaExceptionContext) {
-  auto array = makeArrayVector<int64_t>(
-      10, [](auto /*row*/) { return 5; }, [](auto row) { return row * 3; });
+  auto input = makeRowVector({makeArrayVector<int64_t>(
+      10, [](auto /*row*/) { return 5; }, [](auto row) { return row * 3; })});
 
   assertError(
       "filter(c0, x -> (x / 0 > 1))",
-      array,
+      input,
       "divide(x, 0:BIGINT)",
       "Top-level Expression: filter(c0, (x) -> gt(divide(x, 0:BIGINT), 1:BIGINT))",
       "division by zero");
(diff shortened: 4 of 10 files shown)


Yuhta:
Why not doing it when we compile the expressions?  `exec::Expr` is meant to be runtime expressions and constant folding should be finished before it.

pramodsatya:
> Why not doing it when we compile the expressions? `exec::Expr` is meant to be runtime expressions and constant folding should be finished before it.

Thanks for the suggestion @Yuhta, but I don't quite follow. The virtual function `constructSpecialForm` in `FunctionCallToSpecialForm` is called by the [expression compiler](https://github.com/facebookincubator/velox/blob/3b9239bac2ce4ae9df72a0658f65566b7bc62b94/velox/expression/ExprCompiler.cpp#L416) in `ExprCompiler.cpp -> compileExpressions() -> compileRewrittenExpression()`. When an expression like `(1 = 2) and (2 != 4)` is compiled for instance, it will have as inputs `ConstantExpr`s with values `false` and `true` because the input expressions to `core::CallTypedExpr AND` are constant folded during compilation. So when the function `ConjunctCallToSpecialForm::constructSpecialForm()` is called by the expression compiler, this optimization intends to return a `ConstantExpr` with value `false` instead of the AND `ConjunctExpr` when any of the inputs is (or has been constant folded to) `false`. 
Could you please elaborate where in the expression compiler the changes should be moved to?

Yuhta:
Oh I see it is during expression compilation

pramodsatya:
> Oh I see it is during expression compilation. Can you check why `tryFoldIfConstant` is not working for this case and try to fix it instead?

Thanks for the suggestion @Yuhta, certain cases which were being handled by `tryFoldIfConstant` are now reverted from this PR. The other possible optimizations are outlined in the linked document, could you please take a look? 
I feel it is better for these changes to be in `constructSpecialForm`, since only the constant folded input expressions are needed for these optimizations and they are accessible in this function. These optimizations are also not related to constant folding so it would be better to keep them in `constructSpecialForm`. Alternatively, these changes could also be moved to `evalSpecialForm` so they are applied when the function `tryFoldIfConstant` is called. What do you suggest?

pramodsatya:
@aditi-pandit, @majetideepak, @czentgr, could you please help review these changes?

velox/expression/CoalesceExpr.h
```diff
@@ -53,6 +53,9 @@ class CoalesceCallToSpecialForm : public FunctionCallToSpecialForm {
       std::vector<ExprPtr>&& compiledChildren,
       bool trackCpuUsage,
       const core::QueryConfig& config) override;
+
+ private:
+  ExprPtr optimize(std::vector<ExprPtr>& compiledChildren);
```
mbasmanova, line 58: Assuming this method doesn't access member variables, it can be moved to a free-standing function in .cpp file. Ditto other similar methods.

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
...
+      } else if (nullIndices.size() == i) {
+        return constantExpr;
+      }
+    }
+  }
+
+  if (!nullIndices.empty()) {
+    if (nullIndices.size() == compiledChildren.size()) {
+      auto nullChild = compiledChildren.at(nullIndices.front());
+      return std::dynamic_pointer_cast<exec::ConstantExpr>(nullChild);
```
mbasmanova, line 129: This cast is not needed. Just return nullChild

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
...
+    } else {
```
mbasmanova, line 130: drop 'else' after 'return' for readability

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
...
+      auto nullCount = nullIndices.size();
+      for (vector_size_t j = nullCount - 1; j >= 0; j--) {
```
mbasmanova, line 132: vector_size_t -> auto

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
   return CoalesceExpr::resolveType(argTypes);
 }
 
+ExprPtr CoalesceCallToSpecialForm::optimize(
```
mbasmanova, line 111: The logic here is not easy to read. Would you add a comment to explain what this function does to help future readers?

velox/expression/ConjunctExpr.cpp
```diff
@@ -415,18 +416,89 @@ TypePtr ConjunctCallToSpecialForm::resolveType(
...
+    }
+  }
+
+  if (compiledChildren.size() == 1) {
+    return compiledChildren.front();
+  }
+  return nullptr;
+}
+
+ExprPtr ConjunctCallToSpecialForm::optimizeOr(
```
mbasmanova, line 445: This method is almost a copy-paste of the optimizeAnd method. The only difference is that AND looks for constant TRUE, while OR looks for constant FALSE. Let's refactor to reduce copy-paste.

velox/expression/ConjunctExpr.cpp
```diff
@@ -415,18 +416,89 @@ TypePtr ConjunctCallToSpecialForm::resolveType(
...
+    }
+  }
+
+  if (compiledChildren.size() == 1) {
+    return compiledChildren.front();
+  }
+  return nullptr;
+}
+
+// During expression compilation, the inputs to 'AND', 'OR' conjunct expressions
```
mbasmanova, line 471: This is a nice comment. Please, move it to the header file.

Consider including examples to help readers understand the optimization.

- f(a) AND 1 = 2 => false
- f(b) OR 1 = 1 => true
- f(a) AND 1 = 1 => f(a)
- f(a) OR 1 = 2 => f(a)

velox/expression/tests/ExprTest.cpp
```diff
@@ -2956,14 +2956,14 @@ TEST_P(ParameterizedExprTest, tryWithConstantFailure) {
 TEST_P(ParameterizedExprTest, castExceptionContext) {
   assertError(
       "cast(c0 as bigint)",
-      makeFlatVector<std::string>({"1a"}),
+      makeRowVector({makeFlatVector<std::string>({"1a"})}),
```
mbasmanova, line 2959: Would it be possible to extract this refactoring into a separate PR to allow reviewers to focus on material changes?

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "velox/functions/prestosql/registration/RegistrationFunctions.h"
+#include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+#include "velox/parse/TypeResolver.h"
+
+using namespace facebook::velox;
+using namespace facebook::velox::test;
+using namespace facebook::velox::exec;
+using functions::test::FunctionBaseTest;
+
+class SpecialFormExprOptimizationTest : public FunctionBaseTest {
```
mbasmanova, line 32: Put the test into anonymous namespace within the namespace that contains code under test.

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+ protected:
+  static void SetUpTestCase() {
+    parse::registerTypeResolver();
```
mbasmanova, line 35: use FunctionBaseTest::SetUpTestCase()

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+    exec::registerFunctionCallToSpecialForms();
+    functions::prestosql::registerAllScalarFunctions();
+    memory::MemoryManager::testingSetInstance({});
+  }
+
+  const core::QueryConfig config_{{}};
```
mbasmanova, line 41: Is this needed?

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+TEST_F(SpecialFormExprOptimizationTest, conjunct) {
+  // true AND false AND a -> false
+  auto compiledExpr =
+      compileExpression("(1 = 1) AND (2 != 2) AND a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs()
+          .front();
+  ASSERT_TRUE(compiledExpr->isConstant());
+  ASSERT_EQ(compiledExpr->type(), BOOLEAN());
+  assertEqualVectors(
+      vectorMaker_.constantVector<bool>({false}),
```
mbasmanova, line 53: makeConstant(false, 1)

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
...
+    std::vector<ExprPtr>& compiledChildren) {
+  auto numChildren = compiledChildren.size();
+  std::vector<column_index_t> nullIndices;
+  for (auto i = 0; i < numChildren; i++) {
+    if (auto constantExpr = std::dynamic_pointer_cast<exec::ConstantExpr>(
```
mbasmanova, line 116: `compiledChildren[i]->as<ConstantExpr>()`

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+      compiledExpr->as<ConstantExpr>()->value());
+
+  // false OR true OR a -> true
+  compiledExpr =
+      compileExpression("(1 != 1) OR (2 = 2) OR a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs()
+          .front();
+  ASSERT_TRUE(compiledExpr->isConstant());
```
mbasmanova, line 61: The code to verify that compiledExpr is a constant repeats a couple times. Consider refactoring.

```
assertConstantTrue(compiledExpr);
assertConstantFalse(compiledExpr);
...
void assertConstantTrue(expr) {
   assertConstantBoolean(expr, true);
}
```

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+  ASSERT_EQ(compiledExpr->type(), BOOLEAN());
+  assertEqualVectors(
+      vectorMaker_.constantVector<bool>({false}),
+      compiledExpr->as<ConstantExpr>()->value());
+
+  // false OR true OR a -> true
+  compiledExpr =
+      compileExpression("(1 != 1) OR (2 = 2) OR a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs()
```
mbasmanova, line 59: ->exprs().front() repeats a few times;  consider refactoring

```
auto rowType = ROW({{"a", BOOLEAN()}});
auto expr = compileOne("...", rowType);
...
expr = compileOne("...", rowType);
```

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+  ASSERT_TRUE(compiledExpr->type()->isBoolean());
+  assertEqualVectors(
+      vectorMaker_.constantVector<bool>({true}),
+      compiledExpr->as<ConstantExpr>()->value());
+
+  // true AND true AND a -> a
+  auto compiledExprs =
+      compileExpression("(2 != 3) AND (1 = 1) AND a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs();
+  ASSERT_EQ(compiledExprs.size(), 1);
```
mbasmanova, line 71: consistency: we do not assert the same in other tests

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+  // if(false, a, b) -> b
+  compiledExpr =
+      compileExpression(
+          "if((1 = 2), a, b)", ROW({"a", "b"}, {INTEGER(), INTEGER()}))
+          ->exprs()
+          .front();
+  ASSERT_TRUE(compiledExpr->type()->isInteger());
+  ASSERT_EQ(compiledExpr->name(), "b");
+
+  // coalesce((NULL OR false), (NULL AND true), a) -> a
```
mbasmanova, line 105: Move coalesce tests into separate method.

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+      compileExpression("(2 = 3) OR (1 != 1) OR a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs();
+  ASSERT_EQ(compiledExprs.size(), 1);
+  compiledExpr = compiledExprs.front();
+  ASSERT_TRUE(compiledExpr->type()->isBoolean());
+  ASSERT_EQ(compiledExpr->name(), "a");
+}
+
+TEST_F(SpecialFormExprOptimizationTest, conditional) {
+  // if(true, a, b) -> a
```
mbasmanova, line 87: Are there similar optimizations for switch?

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+  ASSERT_EQ(input->name(), "a");
+
+  // coalesce((NULL OR false), (NULL AND true), a, NULL, b) -> COALESCE(a, b)
+  compiledExpr =
+      compileExpression(
+          "coalesce((cast(NULL as BOOLEAN) OR false), (cast(NULL as BOOLEAN) AND true), a, cast(NULL as BOOLEAN), b)",
+          ROW({"a", "b"}, {BOOLEAN(), BOOLEAN()}))
+          ->exprs()
+          .front();
+  ASSERT_EQ(compiledExpr->inputs().size(), 2);
```
mbasmanova, line 124: Are we missing equals method on Expr?

mbasmanova (changes requested):
@pramodsatya These are nice optimizations. Please, update PR summary to explain these to avoid having the reader to navigate to the Google doc.

High level observation:

Constant folding and subsequent optimizations can be done on ITypedExpr. Is there a particular reason we would prefer to apply these to executable Expr?

CC: @oerling

velox/expression/CoalesceExpr.cpp
```diff
@@ -107,14 +108,48 @@ TypePtr CoalesceCallToSpecialForm::resolveType(
   return CoalesceExpr::resolveType(argTypes);
 }
 
+ExprPtr CoalesceCallToSpecialForm::optimize(
+    std::vector<ExprPtr>& compiledChildren) {
+  auto numChildren = compiledChildren.size();
+  std::vector<column_index_t> nullIndices;
+  for (auto i = 0; i < numChildren; i++) {
```
mbasmanova, line 115: Do we also optimize coalesce(a, null, b, 5, c, d) -> coalesce(a, b, 5)? Let's add a test.

velox/expression/tests/SpecialFormExprOptimizationTest.cpp
```diff
@@ -0,0 +1,131 @@
...
+      compileExpression("(2 != 3) AND (1 = 1) AND a", ROW({"a"}, {BOOLEAN()}))
+          ->exprs();
+  ASSERT_EQ(compiledExprs.size(), 1);
+  compiledExpr = compiledExprs.front();
+  ASSERT_EQ(compiledExpr->name(), "a");
+  ASSERT_EQ(compiledExpr->type(), BOOLEAN());
+
+  // false OR false OR a -> a
+  compiledExprs =
+      compileExpression("(2 = 3) OR (1 != 1) OR a", ROW({"a"}, {BOOLEAN()}))
```
mbasmanova, line 78: Let's also add negative tests for:

- a AND NULL 
- a OR NULL

These should not be optimized.

pramodsatya:
> @pramodsatya These are nice optimizations. Please, update PR summary to explain these to avoid having the reader to navigate to the Google doc.
> 
> High level observation:
> 
> Constant folding and subsequent optimizations can be done on ITypedExpr. Is there a particular reason we would prefer to apply these to executable Expr?
> 
> CC: @oerling

Thanks for the feedback @mbasmanova, updated the PR summary accordingly. 
Constant folding currently happens in the function `tryFoldIfConstant` in [ExprCompiler.cpp](https://github.com/facebookincubator/velox/blob/3b9239bac2ce4ae9df72a0658f65566b7bc62b94/velox/expression/ExprCompiler.cpp#L302), all subtrees of the `TypedExprPtr` are recursively constant folded when function [compileExpressions](https://github.com/facebookincubator/velox/blob/3b9239bac2ce4ae9df72a0658f65566b7bc62b94/velox/expression/ExprCompiler.h#L28) is invoked. The optimizations in this PR rely on a subtree of the special form expression having been constant folded prior to the `TypedExprPtr` representing the special form expression being compiled to an executable `Expr`. 

`tryFoldIfConstant` attempts to constant fold expressions by calling the `eval` function from `exec::Expr` (for special form expressions function `evalSpecialForm` will be called). `ITypedExpr` does not have a similar `eval` function that could be used to constant fold expressions during compilation. So it seems better to apply these optimizations before an executable `Expr` is compiled for special form expressions. Would that be fine?

mbasmanova:
@pramodsatya You latest comment suggests you may not have seen https://github.com/facebookincubator/velox/pull/12474#issuecomment-2693637290 If that's the case, please, take a look.

pramodsatya:
> @pramodsatya You latest comment suggests you may not have seen [#12474 (comment)](https://github.com/facebookincubator/velox/pull/12474#issuecomment-2693637290) If that's the case, please, take a look.

@mbasmanova, sorry I missed that comment. I will introduce an API in `ITypedExpr` and update the PR soon, thanks!

pramodsatya:
> @pramodsatya I wonder what't the overall goal you are working towards. I'm thinking you might be looking for a way to optimize Presto's expressions using Velox eval for constant folding. If that's the case, the workflow you might be putting together is something like:
> 
> Presto's RowExpression -> Velox's ITypedExpr -> Expr -> Presto's RowExpression
> 
> If so, the last leg that converts Expr to RowExpression is problematic as it treats Expr as public API which it is not.
> 
> I further assume that evaluateConstantExpression API is not sufficient as you want to optimize expressions that are not 100% constant, i.e. a * (1 + 2) -> a * 3.
> 
> I believe you need an API that takes ITypedExpr and returns an optimized ITypedExpr. That API should perform constant folding + optimizations in this PR + some more.
> 
> Note that doing constant folding during compilation of ITypedExpr into Expr is not right anyway. This happens after CSE detection and blocks CSE in cases like: f(a, 1 + 2) AND f(a, 3).
> 
> Hence, I suggest to introduce an API to optimize ITypedExpr and use that instead of the approach in this PR.
> 
> It would be great to also consider the following optimizations:
> 
> * f(a, b, NULL, c) -> NULL, where f has default null behavior
> * a + 0 -> a
> * a * 0 -> 0
> * a * 1 -> a
> * a / 1 -> a
> * a / 0 -> fail("Division by zero")
> * a + 1 + 2 -> a + 3 (this doesn't happen today if expression is represented as plus(plus(a, 1), 2))
> * a + 1 - 2 -> a - 1
> * concat(a, 'foo', 'bar') -> concat(a, 'foobar')
> * concat(concat(a, 'foo'), 'bar') -> concat(a, 'foobar')
> * cast('test' as integer) -> fail("...")
> * try(a / 0) -> NULL
> 
> Note that these optimizations work even when 'a' is non-deterministic, i.e. rand() * 0 -> 0.

@mbasmanova, we discussed this internally and had a question, say the API we would introduce to take in an `ITypedExpr` and return an optimized `ITypedExpr` is `optimizeTypedExpr()`. Could we use [compileExpressions()](https://github.com/facebookincubator/velox/blob/8d460174cf57a0c0ac7e1ed3059e8f8c6f860c16/velox/expression/ExprCompiler.h#L28) in `optimizeTypedExpr()` to get a constant folded `exec::Expr`, and add an API to convert this constant folded `exec::Expr` back to an `ITypedExpr`? Or do we want to remove constant folding from `compileExpressions()` and move it to `optimizeTypedExpr()` instead?
Also, would it be fine if we take up the additional optimizations you suggested in a follow-up PR after working on `optimizeTypedExpr()`?
cc: @aditi-pandit @czentgr

mbasmanova:
> Could we use compileExpressions() in optimizeTypedExpr() to get a constant folded exec::Expr, and add an API to convert this constant folded exec::Expr back to an ITypedExpr? 

No, we should not convert Expr to ITypedExpr.

> Or do we want to remove constant folding from compileExpressions() and move it to optimizeTypedExpr() instead?

This is what we want. Initially, it is Ok to keep constant folding logic in ExprCompiler as is and introduce new logic that works on ITypedExpr. Eventually, we'll want to consolidate.

> Also, would it be fine if we take up the additional optimizations you suggested in a follow-up PR after working on optimizeTypedExpr()?

That would be fine.

aditi-pandit:
> > Could we use compileExpressions() in optimizeTypedExpr() to get a constant folded exec::Expr, and add an API to convert this constant folded exec::Expr back to an ITypedExpr?
> 
> No, we should not convert Expr to ITypedExpr.
> 
@mbasmanova : You were mentioning 

"I believe you need an API that takes ITypedExpr and returns an optimized ITypedExpr. That API should perform constant folding + optimizations in this PR + some more."

So we are interested in the result beyond constant folding (that term is being used very broadly). All the other optimizations Velox does for flattening AND and ORs, CSE as well. We will take that result and use it in the rest of Presto optimizer. 

Given that expectation it seemed worthwhile to take the Velox executable expression and convert it to ITypedExpr. I'm keen to understand what are the reservations for that ? Do you have any other options for how to accomplish what we would like ?

aditi-pandit:
@mbasmanova : Pinging on the question. Am really interested to hear your thoughts on the Expr -> ITypedExpr conversion or other ideas you might have. We can try your approaches.

mbasmanova:
@aditi-pandit Hi Aditi, apologize for the delay. It was a bit too busy lately.

> So we are interested in the result beyond constant folding (that term is being used very broadly). All the other optimizations Velox does for flattening AND and ORs, CSE as well. We will take that result and use it in the rest of Presto optimizer.

I would love to understand the overall problem / design a bit better. I can see how it is helpful to flatten AND, OR, concat and similar and I think it is reasonable to include these optimizations into the new API that takes ITypedExpr and produces optimized ITypedExpr. However, CSE is not an optimization that can or should be included. This is because CSE optimization is not a pure plan / expression re-write. It is an execution-specific optimization. 

One may think that CSEs can be extracted into a separate Project node to be executed before the node they were extracted from. This is not the case. For example, it is not safe to extract CSE from under conditional statements (IF, AND, OR, COALESCE, etc.) because CSE may not apply to all rows. 

```
- Project x := if (a = 0, 0, 100 / a), y := if (a = 0 or b = 0, 0, 100 / b + 100 / a) 
   - Scan(t)
```

In this query, 100/a is a CSE, but it cannot be extracted easily.

```
- Project x := if (a = 0, 0, cse), y := if (a = 0 or b = 0, 0, 100 / b + cse) 
   - Project cse := 100 / a
   - Scan(t) -> a, b
```

On this dataset, the original query succeeds, but the re-write fails: division by zero

```
a | b
-----
0, 10
1, 10
```

In Velox, CSE execution includes some non-trivial partial execution logic that is not expressible in the optimized plans (at least not easily).

This is just an example of why in general it is error prone to try to go from an executable expression to a "definition" of the expression.

aditi-pandit:
@mbasmanova : No problem. Thanks for your explanation. I get it now why going from Expr -> TypedExpr is subtle.

We will explore more ideas internally and also share a doc with you.

pramodsatya:
will be implemented in https://github.com/facebookincubator/velox/pull/13424.

PR #10654, opened 2024-08-05
Ignore time zones that are not recognizable by OS when building time zone database

zhztheplayer:
Related to discussion https://github.com/facebookincubator/velox/pull/10577#discussion_r1694501747

The patch fixes fatal error `<zone id> not found in timezone database` when at least one of the timezone IDs in [file](https://github.com/facebookincubator/velox/blob/main/velox/type/tz/TimeZoneDatabase.cpp) don't exist in OS's supported timezone list.

PR #12475, opened 2025-02-27
feat: Throw invalid_time_zone exception from tzdb [4/4]

kevinwilfong:
Summary:
[https://github.com/facebookincubator/velox/pull/10654](https://l.facebook.com/l.php?u=https%3A%2F%2Fgithub.com%2Ffacebookincubator%2Fvelox%2Fpull%2F10654&h=AT018Tv8ICqN3VlCht6ujAqWzlqLbnauv-z1qpzAk8aOuUig1GigRbxXQ0hQXKWH68_oFh66JJz8AyYdyTbLSO-BoSRYA3v8Oa5cn4p-J5ZyKo7DfjktIr-o9Buea_vpQU0ngoPZtePtyDpakDk3U7A) introduced an invalid_timezone exception
to the date library we pulled in.  This PR makes an analogous change to the time zone library we
pulled in from LLVM.  It also restores code/tests that were disabled because this exception was
missing.

Differential Revision: D70344546


velox/external/date/CMakeLists.txt
```diff
@@ -10,7 +10,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-velox_add_library(velox_external_date tz.cpp)
-velox_include_directories(velox_external_date SYSTEM PUBLIC velox/external)
-velox_compile_definitions(velox_external_date PRIVATE USE_OS_TZDB)
+add_library(velox_external_date INTERFACE)
+velox_include_directories(velox_external_date INTERFACE velox/external/date)
```
assignUser, line 14:
```suggestion
target_include_directories(velox_external_date INTERFACE velox/external/date)
```

assignUser (commented):
There is a bit of an inconsistent use of our custom cmake functions here. They are poorly documented, so it's on me ^^

The difference can only be seen when building with `VELOX_MONO_LIBRARY=ON` otherwise `velox_add_library` etc. are transparent wrappers around `add_library` etc..

When `VELOX_MONO_LIBRARY=ON` is set  `velox_add_library` combines the sources and includes of all targets created with it into one libvelox.a . So we only want to use `velox_add_library` for targets that we want to be part of the main velox library itself, which means we don't want to use it for fuzzer, benchmarks, tests and other dev/test utils. 

For the stuff in `external` it depends on if they are velox build time deps and the user can  build against libvelox.a without them, then they can just be static external targets which the linker pulls the symbols from we need -> good for binary size.

If symbols from these libraries are included in out public api we will need to install these external targets to the user system, in that case making them part of libvelox might be a convenient shortcut for now. (maybe needs munging of symbols though to avoid clashes but that's not this PR).

Please choose one approach and use it consistently as mixing normal targets with `velox_*` functions leads to weird results.

velox/external/date/CMakeLists.txt
kevinwilfong: Thanks for the explanation, I'll update https://github.com/facebookincubator/velox/pull/12422 with your suggestion.

PR #12477, opened 2025-02-27
refactor: Move alias for column_index_t to BaseVector.h

aditi-pandit:
Makes it simpler to use column_index_t alias in functions.


aditi-pandit:
@majetideepak : I'm not fixated on this change, but it could be nice for WindowFunctions and TableFunctions use that have a buffer of Vectors without using ComplexVector.h

velox/vector/BaseVector.h
```diff
@@ -51,6 +51,11 @@ class VectorPool;
 class BaseVector;
 using VectorPtr = std::shared_ptr<BaseVector>;
 
+using column_index_t = uint32_t;
```
Yuhta, line 54: This sounds like something at type level, can we move it into `Type.h`?

aditi-pandit: Thanks @Yuhta. That works as well.

Yuhta (approved):

PR #12481, opened 2025-02-28
test: Add InputGenerator for TimestampWithTimeZone

kevinwilfong:
Summary:
This diff adds an implementation of AbstractInputGenerator for the TimestampWithTimeZone 
type.  It uses a random int64_t for the number of milliseconds (there may be overflow since we
only have 52 bits for the milliseconds but that doesn't matter since it's random anyway) and
selects a random time zone ID from a list of valid time zone IDs provided by TimeZoneMap.

Differential Revision: D70362561

### DIFF
--- a/velox/functions/prestosql/types/fuzzer_utils/CMakeLists.txt
+++ b/velox/functions/prestosql/types/fuzzer_utils/CMakeLists.txt
@@ -11,31 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-velox_add_library(
-  velox_presto_types
-  BingTileRegistration.cpp
-  HyperLogLogRegistration.cpp
-  IPAddressRegistration.cpp
-  IPPrefixRegistration.cpp
-  JsonCastOperator.cpp
-  JsonRegistration.cpp
-  TDigestRegistration.cpp
-  TimestampWithTimeZoneRegistration.cpp
-  UuidRegistration.cpp)
+velox_add_library(velox_presto_types_fuzzer_utils
+                  TimestampWithTimeZoneInputGenerator.cpp)
 
 velox_link_libraries(
-  velox_presto_types
+  velox_presto_types_fuzzer_utils
   velox_type
-  velox_memory
-  velox_expression
-  velox_functions_util
-  velox_functions_json
-  velox_functions_lib_date_time_formatter
-  velox_constrained_input_generators
-  velox_presto_types_fuzzer_utils)
+  velox_common_fuzzer_util
+  velox_presto_types
+  velox_type_tz)
 
 if(${VELOX_BUILD_TESTING})
   add_subdirectory(tests)
 endif()
-
-add_subdirectory(fuzzer_utils)
--- a/velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
+++ b/velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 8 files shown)


velox/functions/prestosql/types/fuzzer_utils/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+velox_add_library(velox_presto_types_fuzzer_utils
+                  TimestampWithTimeZoneInputGenerator.cpp)
+
+velox_link_libraries(
```
assignUser, line 17:
```suggestion
add_library(velox_presto_types_fuzzer_utils
                  TimestampWithTimeZoneInputGenerator.cpp)

target_link_libraries(
```

A fuzzer util library doesn't need to be part of the main libvelox.

kevinwilfong: I tried to, but (correct me if I'm wrong) since the AbstractInputGenerator is part of the CustomTypeFactories, TimestampWithTimeZoneRegistration.cpp needs to link against TimestampWithTimeZoneInputGenerator.h in order to construct the TimestampWithTimeZoneInputGenerator, so if it's not included in the main libvelox we run into linker errors.  Similarly velox/common/fuzzer is included in the main libvelox and it contains ConstrainedGenerators.h which is just a bunch of fuzzers (AbstractInputGenerators).

velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
```diff
@@ -0,0 +1,47 @@
...
+  if (coinToss(rng_, nullRatio_)) {
+    return variant::null(type_->kind());
+  }
+
+  int16_t timeZoneId = timeZoneIds_
+      [boost::random::uniform_int_distribution<size_t>()(rng_) %
+       timeZoneIds_.size()];
+
+  return pack(
+      boost::random::uniform_int_distribution<int64_t>()(rng_), timeZoneId);
```
kagamiori, line 45: Would we want to occasionally generate invalid TimestampWithTimezone to test invalid cornercases? (Similar to JsonInputGenerator::makeRandomVariation().)

velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
```diff
@@ -0,0 +1,47 @@
...
+          nullRatio),
+      timeZoneIds_(tz::getTimeZoneIDs()) {}
+
+variant TimestampWithTimeZoneInputGenerator::generate() {
+  if (coinToss(rng_, nullRatio_)) {
+    return variant::null(type_->kind());
+  }
+
+  int16_t timeZoneId = timeZoneIds_
+      [boost::random::uniform_int_distribution<size_t>()(rng_) %
```
kagamiori, line 41: nit: You can use `rand(FuzzerGenerator& rng, T min, T max)` in velox/common/fuzzer/Utils.h

velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
kagamiori, line 45: nit: you can use `rand<int64_t>(rng_, DataSpec())` in velox/common/fuzzer/Utils.h

kagamiori (approved):
LGTM. I left a few nits.

velox/functions/prestosql/types/fuzzer_utils/TimestampWithTimeZoneInputGenerator.cpp
kevinwilfong: I'm not sure if we need that given that TSwTZ is an internal type (it can only occur if it's the output of another UDF, it cannot be part of the input).

The only way a TSwTZ could be invalid is if its timeZoneId is some non-existent value. The only way an invalid TSwTZ could be produced is if a UDF is producing them. As long as we have a UDF that is capable of producing them (from_unixtimestamp comes to mind) included in fuzzing we will cover such cases in fuzzing without having it explicitly here.

If we have a UDF (like json_parse with JsonInputGenerator) where we explicitly want to test invalid values we can
extend this to support it.

kevinwilfong: Actually, looking closer at JsonInputGenerator it produces either a Varchar or a Json type, JsonInputGenerator::makeRandomVariation() is used when outputting a Varchar that is the input to json_parse.  So a better analog would be some yet to be written TimeZoneIdInputGenerator for from_unixtimestamp where we would want to occasionally return an invalid time zone ID for the SmallInt argument.

velox/functions/prestosql/types/fuzzer_utils/CMakeLists.txt
assignUser: I thought we moved `velox/common/fuzzer` out recently but I'm probably mixing something up. Hm, i'll have a look.

PR #12482, opened 2025-02-28
build: Correct bash condition for BUILD_DUCKDB in the setup script

wraymo:
Summary
Fixes an issue in the the setup script where `BUILD_DUCKDB` was incorrectly used in the if condition.

### DIFF
--- a/scripts/setup-centos9.sh
+++ b/scripts/setup-centos9.sh
@@ -175,7 +175,7 @@ function install_mvfst {
 }
 
 function install_duckdb {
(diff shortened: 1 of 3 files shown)


assignUser (approved):
Good catch, thanks

assignUser:
Out of interest I dove into this a bit as I was surprised my local shell check didn't catch it. While the previous version works there are [several false positives possible.](https://stackoverflow.com/a/21210966)

PR #12485, opened 2025-02-28
fix: Disable dynamic filter push-down for Hive tables that do not support it

NEUpanning:
`TableScan#canAddDynamicFilter` always returns true for Hive table regardless of whether the format reader supports this feature. If the filters are pushed down to a format reader that does not support filter push-down, they would not be applied to any rows. While all hive format readers in Velox currently support this feature, this case could happen in the future.

Relates: [gluten#issue](https://github.com/apache/incubator-gluten/issues/8787#issuecomment-2674100600)

### DIFF
--- a/velox/exec/TableScan.h
+++ b/velox/exec/TableScan.h
@@ -15,6 +15,7 @@
  */
 #pragma once
 
+#include "velox/connectors/hive/TableHandle.h"
 #include "velox/core/PlanNode.h"
 #include "velox/exec/Operator.h"
 #include "velox/exec/ScaledScanController.h"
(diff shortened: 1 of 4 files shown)


NEUpanning:
@JkSelf Could you help to take a look?

JkSelf:
@NEUpanning Thanks for your work. Can you help to resolve the failure in the build stage firstly?

Yuhta (changes requested):
Filters in scan spec must be applied for all format readers, it's a bug in the file reader

NEUpanning:
> Filters in scan spec must be applied for all format readers, it's a bug in the file reader. Use projectColumns as a fallback approach in your reader

@Yuhta Thanks for sharing this information. After implementing this approach in my format reader, the issue has been resolved. So I will close this PR.

PR #12489, opened 2025-02-28
fix: Force use index-based column mapping for old ORC file

ccat3z:
An ORC file written by an old version has no field names in the physical schema, always map column names by index.


Yuhta (changes requested):
This is hacky, your metadata service should keep this information and set `useColumnNamesForColumnMapping=false` accordingly for these files.

ccat3z:
cc @kecookier

ccat3z:
> This is hacky, your metadata service should keep this information and set `useColumnNamesForColumnMapping=false` accordingly for these files.

It's impossible to check if it's old ORC files without reading it. And it's a logical error to use `useColumnNamesForColumnMapping` on these ORC files without column names. Spark will determine whether it is an old version when actually reading the ORC, perhaps an option like `autoUseColumnPositionForOldOrc` can be added to support this behavior?

Issue #10354, opened 2024-07-01
Velox doesn't support legacy date format behavior in Spark SQL


NEUpanning:
### Bug description

For backward capability of parsing/formatting of timestamp/date strings expression behavior and adoption of new behaviors. Spark add a setting `spark.sql.legacy.timeParserPolicy` to indicate whether using legacy behavior. If it is not set, Spark doesn't performs strict checking of expression's input. I see Velox doesn't support this when using `unix_timestamp` function, there is a related gluten issue : [link](https://github.com/apache/incubator-gluten/issues/6227). Maybe there are more result mismatch cases.

### System information

\

### Relevant logs

```bash
\
```

NEUpanning:
Maybe we should add support for legacy date format behavior. @pedroerp @rui-mo What do you think? Thanks!

PHILO-HE:
@NEUpanning, is Spark's legacy policy widely used in your production environment?

NEUpanning:
Yes. In Spark upgrades, we need to use legacy policy for backward compatibility rather than pushing users to accept the new behavior. Here is part of legacy policies we are using :
```
spark.sql.storeAssignmentPolicy Legacy
spark.sql.legacy.createArrayOneBasedIndex false
spark.sql.legacy.setCommandRejectsSparkCoreConfs false
spark.sql.legacy.ctePrecedencePolicy LEGACY
spark.sql.legacy.createEmptyCollectionUsingStringType true
spark.sql.legacy.exponentLiteralAsDecimal.enabled true
spark.sql.legacy.parser.havingWithoutGroupByAsWhere true
spark.sql.legacy.timeParserPolicy LEGACY
spark.sql.legacy.dataset.nameNonStructGroupingKeyAsValue true
spark.sql.legacy.fromDayTimeString.enabled true
spark.sql.legacy.allowHashOnMapType true
spark.sql.legacy.allowNegativeScaleOfDecimal true
spark.sql.legacy.followThreeValuedLogicInArrayExists false
spark.sql.legacy.json.allowEmptyString.enabled true
spark.sql.legacy.setopsPrecedence.enabled true
spark.sql.legacy.literal.pickMinimumPrecision false
spark.sql.legacy.sizeOfNull true
spark.sql.legacy.typeCoercion.datetimeToString.enabled true
spark.sql.legacy.doLooseUpcast true
spark.sql.legacy.addMonthsAdjustLastDay true
spark.sql.legacy.statisticalAggregate true
```

[...]

PR #12491, opened 2025-02-28
fix(function): Fix getTimestamp to parse fractional seconds correctly for Spark

NEUpanning:
In `getTimestamp` function, the fractional seconds of "1970-01-01 00:00:00.2"  
should be parsed as 2 milliseconds instead of 200 milliseconds to be consistent 
with Spark.

Part of: #10354  
Fixes: [gluten issue](https://github.com/apache/incubator-gluten/issues/8858)

### DIFF
--- a/velox/functions/lib/DateTimeFormatter.cpp
+++ b/velox/functions/lib/DateTimeFormatter.cpp
@@ -903,6 +903,8 @@ int32_t parseFromPattern(
         ++cur;
         ++count;
       }
+      // If the number of digits is less than 3, a simple formatter interprets
+      // it as the whole number; otherwise, it pads the number with zeros.
       if (type != DateTimeFormatterType::STRICT_SIMPLE &&
           type != DateTimeFormatterType::LENIENT_SIMPLE) {
         number *= std::pow(10, 3 - count);
--- a/velox/functions/lib/DateTimeFormatter.cpp
+++ b/velox/functions/lib/DateTimeFormatter.cpp
@@ -903,7 +903,10 @@ int32_t parseFromPattern(
         ++cur;
         ++count;
       }
-      number *= std::pow(10, 3 - count);
+      if (type != DateTimeFormatterType::STRICT_SIMPLE &&
+          type != DateTimeFormatterType::LENIENT_SIMPLE) {
+        number *= std::pow(10, 3 - count);
+      }
     } else if (
         (curPattern.specifier == DateTimeFormatSpecifier::YEAR ||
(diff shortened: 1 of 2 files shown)


velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -903,7 +903,10 @@ int32_t parseFromPattern(
         ++cur;
         ++count;
       }
-      number *= std::pow(10, 3 - count);
+      if (type != DateTimeFormatterType::STRICT_SIMPLE &&
+          type != DateTimeFormatterType::LENIENT_SIMPLE) {
```
rui-mo, line 907: Perhaps include some comments for this branch.

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -847,13 +847,18 @@ TEST_F(DateTimeFunctionsTest, getTimestamp) {
...
-  enableLegacyFormatter();
+
   // Empty format.
   EXPECT_EQ(getTimestamp("0", ""), std::nullopt);
   // Unsupported specifier.
   EXPECT_EQ(getTimestamp("0", "l"), std::nullopt);
+
+  EXPECT_EQ(
+      getTimestamp("1970-01-01 00:00:00.2", "yyyy-MM-dd HH:mm:ss.SSS"),
+      Timestamp::fromMillis(2));
```
rui-mo, line 861: Perhaps also add one test for the non-legacy behaviour.

velox/functions/lib/DateTimeFormatter.cpp
NEUpanning: Thanks. Updated.

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -847,13 +851,19 @@ TEST_F(DateTimeFunctionsTest, getTimestamp) {
       getTimestamp("2023-07-13 21:34", "yyyy-MM-dd HH:II"),
       "Specifier I is not supported");
 
+  enableLegacyFormatter();
   // Returns null for invalid datetime format when legacy date formatter is
   // used.
-  enableLegacyFormatter();
```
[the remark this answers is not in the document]
NEUpanning: I thought this would be more readable, as the legacy formatter has more tests beyond just "invalid datetime format".

NEUpanning:
@rui-mo Could you take a look again? Thanks.

PR #12496, opened 2025-02-28
fix: TDigest type utility for simple function interface

Yuhta:
Differential Revision: D70401497


natashasehgal (approved):
LGTM, can merge after tests pass

PR #12169, opened 2025-01-24
fix: Make TPCH dbgen text buffer size consistent with Presto Java

minhancao:
Changed text buffer size to be 300 MB for Velox's dbgen to match with Java Presto TPCH dbgen's text buffer size. The text buffer size is used in randomly generating offset and length to grab a chunk from the overall text for each row. This fixed the difference in the comment column for the tables in TPCH. 

Java:
https://github.com/trinodb/tpch/blob/master/src/main/java/io/trino/tpch/TextPool.java#L35
```
private static final int DEFAULT_TEXT_POOL_SIZE = 300 * 1024 * 1024;
```

C++:
https://github.com/facebookincubator/velox/blob/main/velox/tpch/gen/DBGenIterator.cpp#L40
```
load_dists(
        10 * 1024 * 1024, &dbgenCtx); // 10 MB buffer size for text generation.
```

Resolves: https://github.com/prestodb/presto/issues/24011

PR #12499, opened 2025-03-01
feat(tpch): Add flag velox_tpch_text_pool_size_mb for text pool size

minhancao:
Add flag velox_tpch_text_pool_size_mb for TPC-H text pool size

As per discussion in https://github.com/facebookincubator/velox/pull/12169#issuecomment-2677090713, this PR adds a flag velox_tpch_text_pool_size_mb for TPC-H dbgen text pool size. Default value for this config is 300 MB to match with Presto's. 

Reverted TpchConnectorTest.simple's expected output back to what it was for 10 MB buffer size.

### DIFF
--- a/velox/connectors/tpch/tests/TpchConnectorTest.cpp
+++ b/velox/connectors/tpch/tests/TpchConnectorTest.cpp
@@ -22,6 +22,8 @@
 #include "velox/exec/tests/utils/OperatorTestBase.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
 
+DECLARE_int32(velox_tpch_text_pool_size_mb);
+
 namespace {
 
 using namespace facebook::velox;
@@ -35,6 +37,7 @@ class TpchConnectorTest : public exec::test::OperatorTestBase {
   const std::string kTpchConnectorId = "test-tpch";
 
   void SetUp() override {
+    FLAGS_velox_tpch_text_pool_size_mb = 10;
     OperatorTestBase::SetUp();
     connector::registerConnectorFactory(
         std::make_shared<connector::tpch::TpchConnectorFactory>());
@@ -97,11 +100,11 @@ TEST_F(TpchConnectorTest, simple) {
       makeFlatVector<int64_t>({0, 1, 1, 1, 4}),
       // n_comment
       makeFlatVector<StringView>({
-          " haggle. carefully final deposits detect slyly agai",
-          "al foxes promise slyly according to the regular accounts. bold requests alon",
-          "y alongside of the pending deposits. carefully special packages are about the ironic forges. slyly special ",
-          "eas hang ironic, silent packages. slyly regular packages are furiously over the tithes. fluffily bold",
-          "y above the carefully unusual theodolites. final dugouts are quickly across the furiously regular d",
+          "furiously regular requests. platelets affix furious",
+          "instructions wake quickly. final deposits haggle. final, silent theodolites ",
+          "asymptotes use fluffily quickly bold instructions. slyly bold dependencies sleep carefully pending accounts",
+          "ss deposits wake across the pending foxes. packages after the carefully bold requests integrate caref",
+          "usly ironic, pending foxes. even, special instructions nag. sly, final foxes detect slyly fluffily ",
       }),
   });
   test::assertEqualVectors(expected, output);
--- a/velox/flag_definitions/flags.cpp
+++ b/velox/flag_definitions/flags.cpp
@@ -131,3 +131,9 @@ DEFINE_bool(
     velox_ssd_verify_write,
     false,
     "Read back data after writing to SSD");
+
+// Used in /connectors/tpch
+DEFINE_int32(
+    velox_tpch_text_pool_size_mb,
+    300,
+    "TPC-H DBGen text pool size in MB");
--- a/velox/tpch/gen/DBGenIterator.cpp
+++ b/velox/tpch/gen/DBGenIterator.cpp
@@ -37,7 +37,7 @@ class DBGenBackend {
     // structures required by dbgen are populated.
     DBGenContext dbgenCtx;
     load_dists(
-        300 * 1024 * 1024,
+        FLAGS_velox_tpch_text_pool_size_mb * 1024 * 1024,
         &dbgenCtx); // 300 MB buffer size for text generation.
 
     // Initialize global dbgen buffers required to generate data.
--- a/velox/tpch/gen/DBGenIterator.h
+++ b/velox/tpch/gen/DBGenIterator.h
@@ -16,11 +16,14 @@
 
 #pragma once
 
+#include <gflags/gflags.h>
 #include <memory>
 #include <mutex>
 
 #include <velox/tpch/gen/dbgen/include/dbgen/dss.h>
 #include <velox/tpch/gen/dbgen/include/dbgen/dsstypes.h>
+
+DECLARE_int32(velox_tpch_text_pool_size_mb);
 
 namespace facebook::velox::tpch {
 


minhancao:
@amitkdutta @pedroerp @aditi-pandit
Please help review this PR when you guys can, thank you!

I also saw that there's some other Presto tpch connector configs like `tpch.splits-per-node` and `tpch.column-naming` that are not in Velox yet, do we use them in Prestissimo? Should I bring them over in this PR as well?
https://prestodb.io/docs/current/connector/tpch.html#general-configuration-properties

velox/connectors/tpch/CMakeLists.txt
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(velox_tpch_connector OBJECT TpchConnector.cpp)
+velox_add_library(velox_tpch_config OBJECT TpchConfig.cpp)
```
aditi-pandit, line 15: Library velox_tpch_config is not required. Just adding TpchConfig to velox_tpch_connector is sufficient. Please delete lines 15 and 16.

velox/connectors/tpch/TpchConfig.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/tpch/TpchConfig.h"
+#include "velox/common/config/Config.h"
+#include "velox/core/QueryConfig.h"
```
aditi-pandit, line 19: Is QueryConfig include needed ?

velox/connectors/tpch/TpchConfig.h
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <optional>
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::velox::config {
```
aditi-pandit, line 21: "velox/common/config/Config.h" can be directly included here. There isn't a need to forward declare ConfigBase this way.

velox/connectors/tpch/TpchConfig.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace facebook::velox::connector::tpch {
+
+int32_t TpchConfig::textPoolSizeMb() const {
+  return config_->get<int32_t>(kTextPoolSizeMb, 10);
```
aditi-pandit, line 24: Add a comment here giving information about the text pool size for the correct query results for tpch.

velox/connectors/tpch/TpchConfig.cpp
minhancao: No, removed.

velox/connectors/tpch/TpchConfig.cpp
minhancao: Added

```
// For correct query results matching with Presto, use 300 MB for the 
// text pool size.
```

czentgr (commented):
Thanks, just a few minor comments.

velox/connectors/tpch/TpchConfig.h
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  static constexpr const char* kTextPoolSizeMb = "tpch.text-pool-size-mb";
+
+  int32_t textPoolSizeMb() const;
+
+  TpchConfig(std::shared_ptr<const config::ConfigBase> config) {
+    VELOX_CHECK_NOT_NULL(
+        config, "Config is null for TpchConfig initialization");
+    config_ = std::move(config);
+    // TODO: add sanity check
```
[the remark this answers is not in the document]
minhancao: I was creating TpchConfig.h based from https://github.com/facebookincubator/velox/blob/a9999905cc28b49b9d51f5170512168f1398ad0b/velox/connectors/hive/HiveConfig.h#L268

Will remove this // TODO comment.

majetideepak:
A config is an overkill and introduces new API. We should treat the textBufferSize as an argument to TpchTableHandle similar to scaleFactor.
Clients such as Prestissimo can set this there.

czentgr:
@majetideepak @minhancao implemented @amitkdutta 's suggestion of "We can also expose a worker specific config property that can set the flag value." to mean a connector config as shown here. 
But, yes,, suing a connector config might be overkill if we don't anticipate a dynamic change of the value by a user (as it will be hard coded by prestissimo to match java).

minhancao:
@amitkdutta @pedroerp Any preference on how this text buffer size config should be implemented?

majetideepak:
@minhancao We should follow the scaleFactor plumbing. I don't see why the text buffer size would be user configurable.

velox/tpch/gen/DBGenIterator.cpp
```diff
@@ -50,7 +54,8 @@ static folly::Singleton<DBGenBackend> DBGenBackendSingleton;
 
 } // namespace
 
-DBGenIterator::DBGenIterator(double scaleFactor) {
+DBGenIterator::DBGenIterator(double scaleFactor, int32_t textPoolSizeMb) {
+  globalTextPoolSizeMb = textPoolSizeMb;
```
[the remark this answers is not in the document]
minhancao: Changed to:
```
folly::Singleton<DBGenBackend>* getDBGenBackendSingleton(int32_t textPoolSizeMb) {
  static folly::Singleton<DBGenBackend> DBGenBackendSingleton(
    [textPoolSizeMb]() {
        return new DBGenBackend(textPoolSizeMb);
    }
  );
  return &DBGenBackendSingleton;
}
```

velox/connectors/tpch/tests/TpchConnectorTest.cpp
```diff
@@ -97,11 +97,11 @@ TEST_F(TpchConnectorTest, simple) {
       makeFlatVector<int64_t>({0, 1, 1, 1, 4}),
       // n_comment
       makeFlatVector<StringView>({
-          " haggle. carefully final deposits detect slyly agai",
-          "al foxes promise slyly according to the regular accounts. bold requests alon",
-          "y alongside of the pending deposits. carefully special packages are about the ironic forges. slyly special ",
-          "eas hang ironic, silent packages. slyly regular packages are furiously over the tithes. fluffily bold",
-          "y above the carefully unusual theodolites. final dugouts are quickly across the furiously regular d",
+          "furiously regular requests. platelets affix furious",
```
majetideepak, line 100: Can we add a test for 300MB as well?
We can specify the TableHandle. I also think we need to extend TableHandle for Prestissimo? 
```
.tableHandle(std::make_shared<TpchTableHandle>(
                      kTpchConnectorId, Table::TBL_SUPPLIER, scaleFactor))
```

velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -179,11 +179,15 @@ class PlanBuilder {
...
+  /// @param textPoolSizeMb The TPCH-H text pool size in MB.
+  /// For correct query results matching with Presto, use 300 MB for the
+  /// text pool size instead of the default 10 MB.
   PlanBuilder& tpchTableScan(
       tpch::Table table,
       std::vector<std::string> columnNames,
       double scaleFactor = 1,
-      std::string_view connectorId = kTpchDefaultConnectorId);
+      std::string_view connectorId = kTpchDefaultConnectorId,
+      int32_t textPoolSizeMb = 10);
```
majetideepak, line 190: nit: move this after scaleFactor

velox/connectors/tpch/TpchConnector.h
```diff
@@ -126,6 +135,9 @@ class TpchDataSource : public DataSource {
   size_t completedRows_{0};
   size_t completedBytes_{0};
 
+  // For correct query results matching with Presto, use 300 MB for the
+  // text pool size instead of the default 10 MB.
+  int32_t textPoolSizeMb_{10};
```
majetideepak, line 140: Can you move this above after scaleFactor?

majetideepak (commented):
@minhancao Looks great. Two minor comments.
Can you also verify that these changes are sufficient for Prestissimo?

majetideepak:
@minhancao Please fix the PR title and description as well.

minhancao:
@majetideepak I noticed that Pedro added support for generating TPC-H dataset using Python API about 1 month ago: https://github.com/facebookincubator/velox/pull/12381/, I will add the textPoolSizeMb member variable to this as well. 

cc @pedroerp

velox/tpch/gen/DBGenIterator.cpp
```diff
@@ -32,29 +32,54 @@ using namespace dbgen;
...
 
-DBGenIterator::DBGenIterator(double scaleFactor) {
-  auto dbgenBackend = DBGenBackendSingleton.try_get();
+DBGenIterator::DBGenIterator(double scaleFactor, int32_t textPoolSizeMb) {
+  auto dbgenBackend = getDBGenBackendSingleton(textPoolSizeMb)->try_get();
+  if (dbgenBackend != nullptr) {
+    int32_t currentTextPoolSizeMb = dbgenBackend->getTextPoolSizeMb();
+    // This is to ensure not loading dists again when new text pool size
+    // is the same.
+    if (currentTextPoolSizeMb != textPoolSizeMb) {
```
majetideepak, line 79: This is not a clean approach.
Why do we need to call `dbgenBackend->load_new_dists(textPoolSizeMb);` again below?

minhancao: When I was running TpchConnectorTest.simple and then TpchConnectorTest.simpleTextPoolSize300MB, the text pool size for the dbgen singleton still remained 10 MB. 

```
// This is to ensure not loading dists again when new text pool size
  // is the same.
  if (currentTextPoolSizeMb != textPoolSizeMb) {
    dbgenBackend->load_new_dists(textPoolSizeMb);
  }
```

This code is here to allow the test TpchConnectorTest.simpleTextPoolSize300MB to change the text pool size of the dbgen singleton to be 300MB.

majetideepak: > When I was running TpchConnectorTest.simple and then TpchConnectorTest.simpleTextPoolSize300MB in succession, the text pool size for the dbgen singleton still remained 10 MB.

That is expected as DBgenBackend is a singleton and static. You have to create another test executable to test both.
Now I see that textPoolSizeMb should be a GlobalConfig. We are going to add that feature back. Let's add a check that the textPoolSize is always the same.
Pedro here https://github.com/facebookincubator/velox/pull/12169#issuecomment-2659953288 said that 300MB initialization takes a long time. Is it still true? If so, we should remove that test.

minhancao: @majetideepak Yes, the 300MB initialization takes a long time.
```
[ RUN      ] TpchConnectorTest.simple
[       OK ] TpchConnectorTest.simple (119 ms)
```
```
[ RUN      ] TpchConnectorTest.simpleTextPoolSize300MB
[       OK ] TpchConnectorTest.simpleTextPoolSize300MB (2086 ms)
```

I will remove the test and not add a separate executable.

minhancao: @majetideepak I'm a bit confused by "We must only test for the check that the textPoolSize once initialized, is always the same." Can you maybe elaborate/clarify more on what you mean?

majetideepak: Once we initialize the singleton, we cannot change its value. 
We can check that the initial value is the same as subsequent invocations.
`VELOX_CHECK_EQ(dbgenBackend->getTextPoolSizeMb(), textPoolSizeMb)`

This option should ideally be a GlobalConfig as mentioned earlier. Let me add that class first.
We won't have this issue.

majetideepak: @minhancao I don't want to block this on the GlobalConfig change.
Can you add a GFlag for now? `FLAGS_velox_tpch_text_pool_size_mb` in velox/flags_definitions/flags.cpp will have the definitions. You can Define the flag in the tpch folder. You can default it to 300MB. For the test, you can set the flag to 10MB.

velox/flag_definitions/flags.cpp
```diff
@@ -131,3 +131,6 @@ DEFINE_bool(
     velox_ssd_verify_write,
     false,
     "Read back data after writing to SSD");
+
+// Used in /connectors/tpch
+DEFINE_int32(velox_tpch_text_pool_size_mb, 300, "TPC-H pool size in MB.");
```
majetideepak, line 136: nit: `"TPC-H DBGen text pool size in MB"`  Other flag comments don't end in a fullstop.

majetideepak (commented):
This looks nice!
One minor comment.

velox/connectors/tpch/tests/TpchConnectorTest.cpp
```diff
@@ -97,11 +100,11 @@ TEST_F(TpchConnectorTest, simple) {
...
```
majetideepak, line 103: @minhancao Why did the output change if we kept 10MB?

minhancao: @majetideepak I had to change the expected output back to what it was for 10MB.

My PR before https://github.com/facebookincubator/velox/pull/12169 changed the expected output for 300 MB.

majetideepak: Thanks!

majetideepak (approved):

PR #10732, opened 2024-08-13
build: Add option to build monolithic, shared library

assignUser:
This introduces the option to build a shared version of the monolithic `libvelox.a`, the result is a much smaller build tree, library and executables. They are now so small that moving them is much easier, e.g. for test sharding. 

I also fixed some miscellaneous things in this PR:
- Removed `FILESYSTEM`, the link is not necessary with GCC >= 9 which is our minimum.
- Replaced `FOLLYBENCHMARK` with the proper target
- Improved bundled folly's use of glog and gflags (when also bundled)
- [Cleaned up the usage of an addition proto wrapper target](https://github.com/facebookincubator/velox/pull/10732/files#diff-6e1ef8f965a01fe12b35484f26064a6e30207eb3e8722eabe552dba288fbf934R18) so we don't link generated proto files twice (which doesn't cause issues in a static build) 
- Added a marker in the log so it's clear when we are running velox cmake vs. dependency cmake:
```
-- [xsimd] xsimd v10.0.0
-- [stemmer] Building stemmer from source
-- Setting Arrow source to AUTO
-- [Arrow] Checking for module 'thrift'
```
Initially I looked at forcing a bundled build of folly and gflags to make sure they are shared but that introduces weird rpath issues when the dependencies are also on the system (like our CI image), instead I just changed the build script to build folly as shared in the image, we maybe want to make that an option so people can keep folly as static (though even for static build shared folly will likely make the binary size much smaller.).  If folly is build from source it will be built static or shared matching the velox build.

I have changed the wrapper function so that only the mono library target is explicitly created as shared, the few other velox utils, test libraries etc. will be build statically to avoid linking and test issues.

PR #12503, opened 2025-03-02
misc: Revert GlobalConfig to use gflags again

kgpai:
Summary:
X-link: https://github.com/facebookincubator/nimble/pull/147

Reverts the use of GlobalConfig and switches back to using gflags. Unfortunately using globalConfig requires us to call translateFlagsToGlobalConfig immediately after ParseCommandLineFlags to initialize it right. Because other libraries also use velox , transitively there are many such places. Thus reverting this change till we figure out a better mechanism to remove gflags dependencies.

Differential Revision: D70463150


kgpai:
cc: @majetideepak

kgpai:
Our internal linter doesnt have any issues with this change . Will investigate that separately.

bikramSingh91 (approved):

majetideepak:
@kgpai I was not expecting this to be reverted on short notice during the weekend.
The global config change was an important step toward releasing the Velox library.
This was done to avoid facing similar issues due to global gflags symbols in the folly library.
CC: @pedroerp, @mbasmanova

majetideepak:
@mbasmanova Yes, GFlags are not recommended to be used in libraries since they have to be declared in the global namespace and can lead to symbol conflicts. We see this issue with folly when build the velox shared library https://github.com/facebookincubator/velox/pull/10732#issuecomment-2295588282 and also when working with the dynamic library registration https://github.com/facebookincubator/velox/pull/11439/.

They also carry the same static initialization order problem if they are being used to initialize other global static variables.

majetideepak:
I see that both DuckDB and Arrow do not use GFlags in the core library and use config instead.

assignUser:
Additionally my understanding is that using gflags forces anyone that wants to use Velox (or parts of velox that require the flags at least?) to build against gflags when building their application and add ` gflags::ParseCommandLineFlags(&argc, &argv, true);` to their `main()` (see e.g. MergeBenchmark.cpp). Which given that the last gflags release was 6 years ago would annoy me if I wanted to use for example a more up to date alternative.

So I agree with @majetideepak that  we should remove the usage of gflags in libvelox (no problem using it in tests etc. ofc).

kgpai:
@majetideepak Firstly apologies for reverting the change - all blame for this should lie on my doorstep. The reason I had to work on removing it was because Velox is used directly or transitively in many thousands of packages internally and consequently again in many services at meta. Currently unfortunately without the translateFlags option it is not possible to change the behavior and considering the number of services that use Velox , this was a ticking time bomb and as oncall I felt it prudent to revert .  Please consider this as a temporary setback - I will work with you in ensuring that the next solution doesnt have these problems and works for all of us.

kgpai:
@assignUser Currently though there are many thousands of places that ParseCommandLineFlags is used internally and there is a large percentage of these that use Velox. Gflags is pretty endemic and updating all the places where its used and might be a pathway for Velox use puts a large load on the Velox team.

PR #12505, opened 2025-03-03
feat: BingTile cast to/from Bigint

jagill:
Summary:
This just implements the casting functionality.  Methods
to actually access the properties of the BingTile will come later.

Differential Revision: D69942803

### DIFF
--- a/velox/functions/prestosql/types/BingTileType.h
+++ b/velox/functions/prestosql/types/BingTileType.h
@@ -80,6 +80,91 @@ class BingTileType : public BigintType {
     obj["type"] = name();
     return obj;
   }
+
+  static constexpr uint8_t kBingTileVersion = 0;
+  static constexpr uint8_t kBingTileMaxZoomLevel = 23;
+  static constexpr uint8_t kBingTileZoomBitWidth = 5;
+  static constexpr uint8_t kBingTileVersionOffset = 63 - kBingTileZoomBitWidth;
+  static constexpr uint8_t kBingTileZoomOffset = 31 - kBingTileZoomBitWidth;
+  static constexpr uint64_t kBits23Mask = (1 << 24) - 1;
+  static constexpr uint64_t kBits5Mask = (1 << 6) - 1;
+
+  static inline uint64_t
+  bingTileCoordsToInt(uint32_t x, uint32_t y, uint8_t zoom) {
+    uint64_t tile = (static_cast<uint64_t>(zoom) << kBingTileZoomOffset) |
+        (static_cast<uint64_t>(x) << 32) | static_cast<uint64_t>(y);
+    return tile;
+  }
+
+  /// Returns the version of the tile (as uint64)
+  static inline uint8_t bingTileVersion(uint64_t tile) {
+    return (tile >> kBingTileVersionOffset) & kBits5Mask;
+  }
+
+  /// Returns the zoom of the tile (as uint64)
+  static inline uint8_t bingTileZoom(uint64_t tile) {
+    return (tile >> kBingTileZoomOffset) & kBits5Mask;
+  }
+
+  /// Returns the x-coordinate of the tile (as uint64)
+  static inline uint32_t bingTileX(uint64_t tile) {
+    return (tile >> 32) & kBits23Mask;
+  }
+
+  /// Returns the y-coordinate of the tile (as uint64)
+  static inline uint32_t bingTileY(uint64_t tile) {
+    return tile & kBits23Mask;
+  }
+
+  /// Returns true if the tile (as uint64) is valid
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/types/BingTileType.h
```diff
@@ -20,6 +20,50 @@
...
+
+/// Returns the y-coordinate of the tile (as uint64)
+inline uint8_t bingTileY(uint64_t tile) {
+  return tile & BITS_23_MASK;
+}
+
+/// Returns true if the tile (as uint64) is valid
+inline bool isBingTileIntValid(uint64_t tile) {
+  uint64_t coordinateBound = 1 << bingTileZoom(tile);
+  return (uint8_t)(bingTileVersion(tile) == BING_TILE_VERSION) &
```
bikramSingh91, line 61: nit: just curious, is there any benefit to using `cast to (uint8_t) and using bitwise &` instead of using `booleans and logical &`

I would personally prefer the latter in favor of readability and avoiding of c-style casts.

velox/functions/prestosql/types/tests/BingTileTypeTest.cpp
```diff
@@ -40,4 +40,36 @@ TEST_F(BingTileTypeTest, serde) {
   testTypeSerde(BINGTILE());
 }
 
+TEST_F(BingTileTypeTest, packing) {
+  uint64_t tile_x123_y111_z7 = bingTileCoordsToInt(123, 111, 7);
```
bikramSingh91, line 44: nit: can you please update the variable names to use camel case to conform to velox's [naming convention](https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md#naming-conventions)

velox/functions/prestosql/types/tests/BingTileTypeTest.cpp
```diff
@@ -40,4 +40,36 @@ TEST_F(BingTileTypeTest, serde) {
...
+
+  uint64_t tile_z24 = bingTileCoordsToInt(0, 0, 24);
+  ASSERT_FALSE(isBingTileIntValid(tile_z24));
+  ASSERT_EQ(
+      bingTileInvalidReason(tile_z24), "Zoom 24 is greater than max zoom 23");
+
+  uint64_t tile_x0_y1_z0 = bingTileCoordsToInt(0, 1, 0);
+  ASSERT_FALSE(isBingTileIntValid(tile_x0_y1_z0));
+  ASSERT_EQ(
+      bingTileInvalidReason(tile_x0_y1_z0),
```
bikramSingh91, line 71: do we care about the zero bits (5-6 and 36-40 bit positions)? If not, can we add a test case to show that its valid for completeness

velox/functions/prestosql/types/tests/BingTileTypeTest.cpp
```diff
@@ -40,4 +40,36 @@ TEST_F(BingTileTypeTest, serde) {
...
+      "Y coordinate 1 is greater than max coordinate 0 at zoom 0");
```
bikramSingh91, line 72: can you add a case for non-zero zoom level with invalid x or y coordinate?

bikramSingh91 (commented):
Looks good, just a couple of nits

velox/functions/prestosql/types/BingTileType.h
jagill: This way makes the operation branchless and only using SIMD-friendly instructions (bit shifts and comparisions).  Since we are doing this for all elements of a vector, it should make this very cheap: the cache lines will be prefetched and the instructions can be pipelined.

However there's a valid concern of premature optimization, so I'm happy to use standard short-circuiting `&&` and then only move to this if we have a benchmark that improves.

velox/functions/prestosql/types/tests/BingTileTypeTest.cpp
jagill: They are unused in V1 (which likely is the only version): nothing depends on them. Would you want a test showing they are zero?

velox/functions/prestosql/types/BingTileType.h
bikramSingh91: Thanks for the explanation. I don't really have a strong preference either way so this works for me. Can you add a small comment explaining why we chose this, would be good for folks reading this later who are not familiar with this pattern.

velox/functions/prestosql/types/tests/BingTileTypeTest.cpp
bikramSingh91: got it, in that case we can skip adding a test.

bikramSingh91 (approved):

Issue #12508, opened 2025-03-03
Unsafe JSON deserialization code of `HiveConnectorSplit::bucketConversion`


zhztheplayer:
### Bug description

Related code: 

https://github.com/facebookincubator/velox/blob/686f75ecbb8e218db135f15cb061fbe6001b961f/velox/connectors/hive/HiveConnectorSplit.cpp#L146-L153

JSON to deserialize:

```json
{
      "name" : "HiveConnectorSplit",
      "connectorId" : "connector-hive",
      "splitWeight" : 0,
      "cacheable" : false,
      "filePath" : "/tmp/velox4j-5040470243569280667/test/velox4j-test-2500722834285987400.tmp",
      "fileFormat" : "parquet",
      "start" : 0,
      "length" : 3103,
      "partitionKeys" : { },
      "tableBucketNumber" : null,
      "bucketConversion" : null,
      "customSplitInfo" : { },
      "extraFileInfo" : null,
      "serdeParameters" : { },
      "storageParameters" : { },
      "infoColumns" : { },
      "properties" : null,
      "rowIdProperties" : null
    }
```

Error:

```
TypeError: expected dynamic type `object', but had type `null'
```

zhztheplayer:
Reason:

`obj.count("bucketConversion")` counts on null `bucketConversion` values.

PR #12509, opened 2025-03-03
fix: Fix unsafe deserialization code of `HiveConnectorSplit::bucketConversion`

zhztheplayer:
API `folly::dynamic::count` counts on null values as well so it's not enough if the subsequent code assumes the value is not null. Add `!obj["bucketConversion"].isNull()` for the checking.

This fixes https://github.com/facebookincubator/velox/issues/12508.


zhztheplayer:
cc @xiaoxmeng thanks

xiaoxmeng (approved):

PR #12517, opened 2025-03-03
build: GCC13 compilation failures in Experimental CI

### LINKED ISSUE #12497: Fix build failures with GCC13
### Problem description

Velox has build failures with GCC13 on ubuntu 24.04.
https://github.com/facebookincubator/velox/actions/runs/13589489006/job/37991866384?pr=12399

There are some discussion on this in https://github.com/facebookincubator/velox/pull/12102#pullrequestreview-2557910172 and https://github.com/facebookincubator/velox/pull/12399#issuecomment-2690861908, and we need fix gcc13 build issues now.

### System information

Velox System Info v0.0.2
Commit: d118b1ed4309511b2a8a7730976f55800bf8f46d
CMake Version: 3.28.3
System: Linux-6.8.0-53-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 13.3.0
C Compiler: /usr/bin/cc
C Compiler Version: 13.3.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

### CMake log

```bash

```

xin-zhang2:
Unblock Experimental CI jobs which use Ubuntu 24.04

Fixes #12497

### DIFF
--- a/velox/connectors/tpch/TpchConnector.cpp
+++ b/velox/connectors/tpch/TpchConnector.cpp
@@ -87,12 +87,7 @@ TpchDataSource::TpchDataSource(
         toTableName(tpchTable_));
 
     auto handle = std::dynamic_pointer_cast<TpchColumnHandle>(it->second);
-    VELOX_CHECK_NOT_NULL(
-        handle,
-        "ColumnHandle must be an instance of TpchColumnHandle "
-        "for '{}' on table '{}'",
-        handle->name(),
-        toTableName(tpchTable_));
+    VELOX_CHECK_NOT_NULL(handle, "ColumnHandle can't be null.");
 
     auto idx = tpchTableSchema->getChildIdxIfExists(handle->name());
     VELOX_CHECK(
--- a/velox/connectors/tpch/TpchConnector.cpp
+++ b/velox/connectors/tpch/TpchConnector.cpp
@@ -87,7 +87,12 @@ TpchDataSource::TpchDataSource(
         toTableName(tpchTable_));
 
     auto handle = std::dynamic_pointer_cast<TpchColumnHandle>(it->second);
-    VELOX_CHECK_NOT_NULL(handle, "ColumnHandle can't be null.");
+    VELOX_CHECK_NOT_NULL(
+        handle,
+        "ColumnHandle must be an instance of TpchColumnHandle "
+        "for '{}' on table '{}'",
+        it->second->name(),
+        toTableName(tpchTable_));
 
     auto idx = tpchTableSchema->getChildIdxIfExists(handle->name());
     VELOX_CHECK(
--- a/velox/dwio/common/compression/Compression.h
+++ b/velox/dwio/common/compression/Compression.h
@@ -84,9 +84,9 @@ struct CompressionOptions {
     struct {
       bool isHadoopFrameFormat;
     } lz4_lzo;
-  } format;
+  } format{};
 
-  uint32_t compressionThreshold;
+  uint32_t compressionThreshold{0};
 };
 
 /**
--- a/velox/dwio/common/compression/Compression.h
+++ b/velox/dwio/common/compression/Compression.h
@@ -84,9 +84,9 @@ struct CompressionOptions {
     struct {
       bool isHadoopFrameFormat;
     } lz4_lzo;
-  } format{};
+  } format;
 
-  uint32_t compressionThreshold{0};
+  uint32_t compressionThreshold;
 };
 
 /**
--- a/velox/dwio/dwrf/common/Compression.h
+++ b/velox/dwio/dwrf/common/Compression.h
@@ -38,7 +38,7 @@ inline CompressionOptions getDwrfOrcCompressionOptions(
     uint32_t compressionThreshold,
     int32_t zlibCompressionLevel,
     int32_t zstdCompressionLevel) {
-  CompressionOptions options;
+  CompressionOptions options{};
   options.compressionThreshold = compressionThreshold;
 
   if (kind == velox::common::CompressionKind_ZLIB ||
@@ -87,7 +87,7 @@ inline std::unique_ptr<dwio::common::BufferedOutputStream> createCompressor(
 
 inline CompressionOptions getDwrfOrcDecompressionOptions(
     common::CompressionKind kind) {
-  CompressionOptions options;
+  CompressionOptions options{};
   if (kind == common::CompressionKind_ZLIB ||
       kind == common::CompressionKind_GZIP) {
     options.format.zlib.windowBits = Compressor::DWRF_ORC_ZLIB_WINDOW_BITS;
--- a/velox/dwio/dwrf/common/Compression.h
+++ b/velox/dwio/dwrf/common/Compression.h
@@ -87,7 +87,7 @@ inline std::unique_ptr<dwio::common::BufferedOutputStream> createCompressor(
 
 inline CompressionOptions getDwrfOrcDecompressionOptions(
     common::CompressionKind kind) {
-  CompressionOptions options{};
+  CompressionOptions options;
   if (kind == common::CompressionKind_ZLIB ||
       kind == common::CompressionKind_GZIP) {
     options.format.zlib.windowBits = Compressor::DWRF_ORC_ZLIB_WINDOW_BITS;
--- a/velox/dwio/dwrf/common/Compression.h
+++ b/velox/dwio/dwrf/common/Compression.h
@@ -87,7 +87,7 @@ inline std::unique_ptr<dwio::common::BufferedOutputStream> createCompressor(
 
 inline CompressionOptions getDwrfOrcDecompressionOptions(
     common::CompressionKind kind) {
-  CompressionOptions options;
+  CompressionOptions options{};
   if (kind == common::CompressionKind_ZLIB ||
       kind == common::CompressionKind_GZIP) {
     options.format.zlib.windowBits = Compressor::DWRF_ORC_ZLIB_WINDOW_BITS;
--- a/velox/expression/tests/ExpressionRunner.cpp
+++ b/velox/expression/tests/ExpressionRunner.cpp
@@ -230,10 +230,7 @@ void ExpressionRunner::run(
       auto inputVector =
           std::dynamic_pointer_cast<RowVector>(restoreVectorFromFile(
               inputPathsList[i].c_str(), deserializerPool.get()));
-      VELOX_CHECK_NOT_NULL(
-          inputVector,
-          "Input vector is not a RowVector: {}",
-          inputVector->toString());
+      VELOX_CHECK_NOT_NULL(inputVector, "Input vector can't be null.");
       VELOX_CHECK_GT(inputVector->size(), 0, "Input vector must not be empty.");
       if (inputSelectivityPaths.size() > i) {
         inputTestCases.push_back(
--- a/velox/expression/tests/ExpressionRunner.cpp
+++ b/velox/expression/tests/ExpressionRunner.cpp
@@ -227,10 +227,13 @@ void ExpressionRunner::run(
     std::vector<std::string> inputSelectivityPaths =
         split(inputSelectivityVectorPaths);
     for (int i = 0; i < inputPathsList.size(); i++) {
-      auto inputVector =
-          std::dynamic_pointer_cast<RowVector>(restoreVectorFromFile(
-              inputPathsList[i].c_str(), deserializerPool.get()));
-      VELOX_CHECK_NOT_NULL(inputVector, "Input vector can't be null.");
+      auto vector = restoreVectorFromFile(
+          inputPathsList[i].c_str(), deserializerPool.get());
+      auto inputVector = std::dynamic_pointer_cast<RowVector>(vector);
+      VELOX_CHECK_NOT_NULL(
+          inputVector,
+          "Input vector is not a RowVector: {}",
+          vector->toString());
       VELOX_CHECK_GT(inputVector->size(), 0, "Input vector must not be empty.");
       if (inputSelectivityPaths.size() > i) {
         inputTestCases.push_back(
--- a/velox/expression/tests/GenericViewTest.cpp
+++ b/velox/expression/tests/GenericViewTest.cpp
@@ -19,6 +19,7 @@
 #include "velox/common/base/CompareFlags.h"
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/common/testutil/OptionalEmpty.h"
 #include "velox/expression/VectorReaders.h"
 #include "velox/functions/Udf.h"
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
@@ -38,7 +39,7 @@ class GenericViewTest : public functions::test::FunctionBaseTest {
       std::vector<std::optional<std::vector<std::optional<int64_t>>>>;
 
   array_data_t arrayData1 = {
-      std::make_optional<std::vector<std::optional<int64_t>>>({}),
+      common::testutil::optionalEmpty,
       {{{{std::nullopt}}}},
       {{std::nullopt, 1}},
       {{std::nullopt, std::nullopt, std::nullopt}},
@@ -56,7 +57,7 @@ class GenericViewTest : public functions::test::FunctionBaseTest {
   };
 
   array_data_t arrayData2 = {
-      std::make_optional<std::vector<std::optional<int64_t>>>({}),
+      common::testutil::optionalEmpty,
       {{{{std::nullopt}}}},
       {{std::nullopt, 1}},
       {{std::nullopt, std::nullopt, std::nullopt}},
--- a/velox/expression/tests/GenericViewTest.cpp
+++ b/velox/expression/tests/GenericViewTest.cpp
@@ -38,7 +38,7 @@ class GenericViewTest : public functions::test::FunctionBaseTest {
       std::vector<std::optional<std::vector<std::optional<int64_t>>>>;
 
   array_data_t arrayData1 = {
-      {{}},
+      std::make_optional<std::vector<std::optional<int64_t>>>({}),
       {{{{std::nullopt}}}},
       {{std::nullopt, 1}},
       {{std::nullopt, std::nullopt, std::nullopt}},
@@ -56,7 +56,7 @@ class GenericViewTest : public functions::test::FunctionBaseTest {
   };
 
   array_data_t arrayData2 = {
-      {{}},
+      std::make_optional<std::vector<std::optional<int64_t>>>({}),
       {{{{std::nullopt}}}},
       {{std::nullopt, 1}},
       {{std::nullopt, std::nullopt, std::nullopt}},
--- a/velox/vector/fuzzer/VectorFuzzer.h
+++ b/velox/vector/fuzzer/VectorFuzzer.h
@@ -145,13 +145,24 @@ class VectorFuzzer {
     fuzzer::DataSpec dataSpec{false, false};
   };
 
+  // Suppress spurious warnings in GCC 12.4 and later
+#if defined(__GNUC__) && \
+    ((__GNUC__ > 12) || (__GNUC__ == 12 && __GNUC_MINOR__ >= 4))
+  VELOX_SUPPRESS_STRINGOP_OVERFLOW_WARNING
+#endif
   VectorFuzzer(
       const VectorFuzzer::Options& options,
       memory::MemoryPool* pool,
       size_t seed = 123456)
-      : opts_(options), pool_(pool), rng_(seed) {}
+      : opts_(options),
+        pool_(pool),
+        rng_(seed){}
+#if defined(__GNUC__) && \
+    ((__GNUC__ > 12) || (__GNUC__ == 12 && __GNUC_MINOR__ >= 4))
+        VELOX_UNSUPPRESS_STRINGOP_OVERFLOW_WARNING
+#endif
 
-  void setOptions(const VectorFuzzer::Options& options) {
+        void setOptions(const VectorFuzzer::Options& options) {
     opts_ = options;
   }
 
--- a/velox/common/testutil/OptionalEmpty.h
+++ b/velox/common/testutil/OptionalEmpty.h
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <optional>
+#include <vector>
+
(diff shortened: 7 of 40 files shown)


velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -145,11 +145,14 @@ class VectorFuzzer {
     fuzzer::DataSpec dataSpec{false, false};
   };
 
+  // Spurious warnings in GCC 13
+  VELOX_SUPPRESS_STRINGOP_OVERFLOW_WARNING
```
xin-zhang2, line 149: @majetideepak I suppressed the stringop_overflow warning here as there is a failure in the initializaiton of opts_.
```
usr/include/c++/13/bits/stl_algobase.h:437:30: error: ‘void* __builtin_memmove(void*, const void*, long unsigned int)’ writing between 5 and 9223372036854775807 bytes into a region of size 4 overflows the destination [-Werror=stringop-overflow=]
  437 |             __builtin_memmove(__result, __first, sizeof(_Tp) * _Num);
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/c++/13/bits/c++allocator.h:33,
                 from /usr/include/c++/13/bits/allocator.h:46,
                 from /usr/include/c++/13/string:43:
In member function ‘_Tp* std::__new_allocator<_Tp>::allocate(size_type, const void*) [with _Tp = facebook::velox::fuzzer::UTF8CharList]’,
    inlined from ‘static _Tp* std::allocator_traits<std::allocator<_CharT> >::allocate(allocator_type&, size_type) [with _Tp = facebook::velox::fuzzer::UTF8CharList]’ at /usr/include/c++/13/bits/alloc_traits.h:482:28,
    inlined from ‘std::_Vector_base<_Tp, _Alloc>::pointer std::_Vector_base<_Tp, _Alloc>::_M_allocate(std::size_t) [with _Tp = facebook::velox::fuzzer::UTF8CharList; _Alloc = std::allocator<facebook::velox::fuzzer::UTF8CharList>]’ at /usr/include/c++/13/bits/stl_vector.h:381:33,
    inlined from ‘void std::_Vector_base<_Tp, _Alloc>::_M_create_storage(std::size_t) [with _Tp = facebook::velox::fuzzer::UTF8CharList; _Alloc = std::allocator<facebook::velox::fuzzer::UTF8CharList>]’ at /usr/include/c++/13/bits/stl_vector.h:398:44,
    inlined from ‘std::_Vector_base<_Tp, _Alloc>::_Vector_base(std::size_t, const allocator_type&) [with _Tp = facebook::velox::fuzzer::UTF8CharList; _Alloc = std::allocator<facebook::velox::fuzzer::UTF8CharList>]’ at /usr/include/c++/13/bits/stl_vector.h:335:26,
    inlined from ‘std::vector<_Tp, _Alloc>::vector(const std::vector<_Tp, _Alloc>&) [with _Tp = facebook::velox::fuzzer::UTF8CharList; _Alloc = std::allocator<facebook::velox::fuzzer::UTF8CharList>]’ at /usr/include/c++/13/bits/stl_vector.h:603:61,
    inlined from ‘facebook::velox::VectorFuzzer::Options::Options(const facebook::velox::VectorFuzzer::Options&)’ at /root/velox/./velox/vector/fuzzer/VectorFuzzer.h:78:10,
    inlined from ‘facebook::velox::VectorFuzzer::VectorFuzzer(const Options&, facebook::velox::memory::MemoryPool*, size_t)’ at /root/velox/./velox/vector/fuzzer/VectorFuzzer.h:153:9,
    inlined from ‘ExchangeFuzzer::ExchangeFuzzer()’ at /root/velox/velox/exec/tests/ExchangeFuzzer.cpp:96:22:
```
The code in VectorFuzzer.h:153 is `: opts_(options), pool_(pool), rng_(seed) {}`.

This failure should be caused by a bug in GCC 13 and it only occurs in the release build.
I also found some similiar issues online like [114758](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114758) and [111415](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111415).

How do you think we should resolve this issue?

czentgr: Is it known which versions have the bug? You could add a define to activate only of the offending versions of gcc are used to compile.
This is only ion the fuzzer so I don't think it's a big issue adding the warning suppression here.
We've had a few problems like that. So if you add compiler and version specific check around it it should be ok.

xin-zhang2: @czentgr 
It looks the issue was introduced in GCC 12.4, and has not been fixed yet in the latest 14.2.
https://godbolt.org/z/d7PK58hxT

I didn't reproduce it with GCC trunk, so it is expected to be fixed in the next GCC release.
I have added the condition to check the compiler and its version.

velox/connectors/tpch/TpchConnector.cpp
```diff
@@ -87,12 +87,7 @@ TpchDataSource::TpchDataSource(
         toTableName(tpchTable_));
 
     auto handle = std::dynamic_pointer_cast<TpchColumnHandle>(it->second);
-    VELOX_CHECK_NOT_NULL(
-        handle,
-        "ColumnHandle must be an instance of TpchColumnHandle "
-        "for '{}' on table '{}'",
-        handle->name(),
-        toTableName(tpchTable_));
+    VELOX_CHECK_NOT_NULL(handle, "ColumnHandle can't be null.");
```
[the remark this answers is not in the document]
xin-zhang2: When handle is nullptr, `handle->name()` in the previous message will cause a null pointer dereference.
```
/root/velox/velox/connectors/tpch/TpchConnector.cpp:94:21: error: ‘this’ pointer is null [-Werror=nonnull]
   94 |         handle->name(),
      |         ~~~~~~~~~~~~^~
/root/velox/./velox/common/base/Exceptions.h:169:60: note: in definition of macro ‘_VELOX_THROW_IMPL’
  169 |     auto message = ::facebook::velox::detail::errorMessage(__VA_ARGS__); \
      |                                                            ^~~~~~~~~~~
/root/velox/./velox/common/base/Exceptions.h:196:3: note: in expansion of macro ‘_VELOX_CHECK_AND_THROW_IMPL’
  196 |   _VELOX_CHECK_AND_THROW_IMPL(                                      \
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/root/velox/./velox/common/base/Exceptions.h:271:32: note: in expansion of macro ‘_VELOX_CHECK_IMPL’
  271 | #define VELOX_CHECK(expr, ...) _VELOX_CHECK_IMPL(expr, #expr, ##__VA_ARGS__)
      |                                ^~~~~~~~~~~~~~~~~
/root/velox/./velox/common/base/Exceptions.h:279:38: note: in expansion of macro ‘VELOX_CHECK’
  279 | #define VELOX_CHECK_NOT_NULL(e, ...) VELOX_CHECK(e != nullptr, ##__VA_ARGS__)
      |                                      ^~~~~~~~~~~
/root/velox/velox/connectors/tpch/TpchConnector.cpp:90:5: note: in expansion of macro ‘VELOX_CHECK_NOT_NULL’
   90 |     VELOX_CHECK_NOT_NULL(
      |     ^~~~~~~~~~~~~~~~~~~~
```

velox/dwio/dwrf/common/Compression.h
```diff
@@ -87,7 +87,7 @@ inline std::unique_ptr<dwio::common::BufferedOutputStream> createCompressor(
 
 inline CompressionOptions getDwrfOrcDecompressionOptions(
     common::CompressionKind kind) {
-  CompressionOptions options;
+  CompressionOptions options{};
```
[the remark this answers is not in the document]
xin-zhang2: Initialize `options` explicitly to avoid using it uninitialized.
```
/root/velox/./velox/dwio/dwrf/common/Compression.h:99:10: error: ‘options’ is used uninitialized [-Werror=uninitialized]
   99 |   return options;
      |          ^~~~~~~
/root/velox/./velox/dwio/dwrf/common/Compression.h: In member function ‘virtual void TestSeek_uncompressed_Test::TestBody()’:
/root/velox/./velox/dwio/dwrf/common/Compression.h:90:22: note: ‘options’ declared here
   90 |   CompressionOptions options;
      |                      ^~~~~~~
```

velox/expression/tests/ExpressionRunner.cpp
```diff
@@ -230,10 +230,7 @@ void ExpressionRunner::run(
       auto inputVector =
           std::dynamic_pointer_cast<RowVector>(restoreVectorFromFile(
               inputPathsList[i].c_str(), deserializerPool.get()));
-      VELOX_CHECK_NOT_NULL(
-          inputVector,
-          "Input vector is not a RowVector: {}",
-          inputVector->toString());
+      VELOX_CHECK_NOT_NULL(inputVector, "Input vector can't be null.");
```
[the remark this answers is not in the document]
xin-zhang2: When inputVector is nullptr, `inputVector->toString()` in the previous message will cause a null pointer dereference.

velox/dwio/dwrf/common/Compression.h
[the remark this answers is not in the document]
majetideepak: Let's initialize the fields in the `struct CompressionOptions`

velox/connectors/tpch/TpchConnector.cpp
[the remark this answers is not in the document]
majetideepak: `handle->name()` must be replaced with `it->second->name()`.
But nice find by GCC 13.

velox/expression/tests/ExpressionRunner.cpp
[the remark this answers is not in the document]
majetideepak:
```
auto vector = restoreVectorFromFile(inputPathsList[i].c_str(), deserializerPool.get());
auto inputVector = std::dynamic_pointer_cast<RowVector>(vector);

use vector->toString() in the null check argument.
```

velox/connectors/tpch/TpchConnector.cpp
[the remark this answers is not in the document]
xin-zhang2: Updated.

velox/expression/tests/ExpressionRunner.cpp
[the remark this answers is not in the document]
xin-zhang2: Updated.

velox/dwio/dwrf/common/Compression.h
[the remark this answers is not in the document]
xin-zhang2: Initialized the fields.
There is no default value for `CompressionOptions.format`, so I left it empty initialized, which results in `format.zlib.windowBits` and `format.zlib.compressionLevel` being initialized to 0.
As for `CompressionOptions.compressionThreshold`, it is currently used only by `DwrfOrcCompression`, where its value is configured via `orc.compression.threshold`, with a default of 256. However, since `CompressionOptions` is part of the common package, I believe it would be more appropriate to initialize `compressionThreshold` to 0.
Does the approach seem reasonable?

velox/dwio/common/compression/Compression.h
```diff
@@ -84,9 +84,9 @@ struct CompressionOptions {
     struct {
       bool isHadoopFrameFormat;
     } lz4_lzo;
-  } format;
+  } format{};
```
majetideepak, line 87: We must default each field to 0.

majetideepak: `{}` is not required

xin-zhang2: Will default all scalar fields to 0.
We need `{}` either after `zlib` or `format` to make `format` default intialized, otherwise we have to explicitly implement a constructor for the initialization of `format`.

majetideepak: If the fields are initialized, we should be good no?

xin-zhang2: yes, both the union field `format` and scalar field `compressionThreshold` should be initialized.

If all scalar fields default to 0,  the simplest way is to only add `{}` after `format`. It will execute a [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) that zero-intialize the union, thus no initializers required for structs in this union, and the code is like
```
struct CompressionOptions {
  union Format {
    struct {
      int windowBits;
      int32_t compressionLevel;
    } zlib;

    struct {
      int32_t compressionLevel;
    } zstd;

    struct {
      bool isHadoopFrameFormat;
    } lz4_lzo;
  } format{};

  uint32_t compressionThreshold{0};
};
```

If we don't want zero-initialize `format`, it can be [default initialized](https://en.cppreference.com/w/cpp/language/default_initialization) without using `{}` after `format`. In this case,  exactly one member of the union should be initialized. The code is like
```
struct CompressionOptions {
  union Format {
    struct {
      int windowBits{0};
      int32_t compressionLevel{0};
    } zlib {};

    struct {
      int32_t compressionLevel;
    } zstd;

    struct {
      bool isHadoopFrameFormat;
    } lz4_lzo;
  } format;

  uint32_t compressionThreshold{0};
};
```

czentgr (approved):

majetideepak (approved):

majetideepak:
@xin-zhang2 The experimental CI job that uses GCC 13 needs to be triggered to test this. Can you add a comment somewhere in .github/workflows/experimental.yml  to trigger the CI? Thanks.

xin-zhang2:
@majetideepak I've added a comment to trigger the CI, but it looks the Experimental Fuzzer Job didn't pick my code.
For example, the error message in the job is 
```
/home/runner/work/velox/velox/velox/velox/vector/tests/VectorMakerTest.cpp:477:65: error: converting to ‘std::optional<std::vector<std::optional<int>, std::allocator<std::optional<int> > > >’ from initializer list would use explicit constructor ‘constexpr std::optional<_Tp>::optional(std::in_place_t, _Args&& ...) [with _Args = {}; typename std::enable_if<__and_v<std::is_constructible<_Tp, _Args ...> >, bool>::type <anonymous> = false; _Tp = std::vector<std::optional<int>, std::allocator<std::optional<int> > >]’
  477 |   VectorPtr expectedVector = maker_.arrayVectorNullable<int32_t>({
      |                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
  478 |       {{1, 2}},
      |       ~~~~~~~~~                                                  
  479 |       {{2, 3, 4}},
      |       ~~~~~~~~~~~~                                               
  480 |       {{std::nullopt, 7}},
      |       ~~~~~~~~~~~~~~~~~~~~                                       
  481 |       {{1, 3, 7, 9}},
      |       ~~~~~~~~~~~~~~~                                            
  482 |       {{}},
      |       ~~~~~                                                      
  483 |       {{std::nullopt}},
      |       ~~~~~~~~~~~~~~~~~                                          
  484 |       {{1, 2, std::nullopt}},
      |       ~~~~~~~~~~~~~~~~~~~~~~~                                    
  485 |       {{}},
      |       ~~~~~                                                      
  486 |       std::nullopt,
      |       ~~~~~~~~~~~~~                                              
  487 |       {{1, 2, 3}},
      |       ~~~~~~~~~~~~                                               
  488 |       {{}},
      |       ~~~~~                                                      
  489 |       {{4, 5}},
      |       ~~~~~~~~~                                                  
  490 |   });
      |   ~~          
```

while I have already made the following change in my PR.
image

Any idea on this?

majetideepak:
@xin-zhang2 CI is checking out the main branch. Are you able to test this locally? If yes, we an undo the comment.

xin-zhang2:
@majetideepak  Yes, the build runs successfully on my local ubuntu 24.04 enviroment.
I have reverted the comment commit.

majetideepak:
@xin-zhang2 thanks for this fix. Meta will import and merge this.

velox/connectors/tpch/TpchConnector.cpp
```diff
@@ -91,7 +91,7 @@ TpchDataSource::TpchDataSource(
         handle,
         "ColumnHandle must be an instance of TpchColumnHandle "
         "for '{}' on table '{}'",
-        handle->name(),
+        it->second->name(),
```
Yuhta, line 94: Why do we need to change this?  Looks like a readability regression to me

velox/dwio/common/compression/Compression.h
Yuhta, line 87: I honestly don't see the value zeroing out these stuff.  People need to initialize them when they want to use them.  In this case it's fine to pay the extra initialization cost, but that's not a general rule we should follow in this code base.

velox/expression/tests/ExpressionRunner.cpp
```diff
@@ -227,13 +227,13 @@ void ExpressionRunner::run(
...
-          std::dynamic_pointer_cast<RowVector>(restoreVectorFromFile(
-              inputPathsList[i].c_str(), deserializerPool.get()));
+      auto vector = restoreVectorFromFile(
+          inputPathsList[i].c_str(), deserializerPool.get());
+      auto inputVector = std::dynamic_pointer_cast<RowVector>(vector);
       VELOX_CHECK_NOT_NULL(
           inputVector,
           "Input vector is not a RowVector: {}",
-          inputVector->toString());
+          vector->toString());
```
Yuhta, line 236: Again why we cannot call `toString` from `inputVector`?  Looks like a bug in compiler.

velox/expression/tests/GenericViewTest.cpp
```diff
@@ -38,7 +38,7 @@ class GenericViewTest : public functions::test::FunctionBaseTest {
       std::vector<std::optional<std::vector<std::optional<int64_t>>>>;
 
   array_data_t arrayData1 = {
-      {{}},
+      std::make_optional<std::vector<std::optional<int64_t>>>({}),
```
Yuhta, line 41: This `make_optional` thing looks really like a readability regression, can we avoid doing it?

Yuhta (changes requested):

velox/connectors/tpch/TpchConnector.cpp
xin-zhang2: `handle` is cast from `it->second` using
```
auto handle = std::dynamic_pointer_cast<TpchColumnHandle>(it->second);
```
If the cast fails, `handle` will be nullptr, and `handle->name()` will cause a nullptr dereference.

velox/expression/tests/ExpressionRunner.cpp
xin-zhang2: The reason is same as `handle` above. 
If `inputVector` is nullptr, `toString` will cause nullptr dereference.

velox/dwio/common/compression/Compression.h
xin-zhang2: CompressionOptions is used uninitialized in code like velox/dwio/dwrf/common/Compression.h
```
inline CompressionOptions getDwrfOrcDecompressionOptions(
    common::CompressionKind kind) {
  CompressionOptions options;
  if (kind == common::CompressionKind_ZLIB ||
      kind == common::CompressionKind_GZIP) {
    options.format.zlib.windowBits = Compressor::DWRF_ORC_ZLIB_WINDOW_BITS;
  } else if (
      kind == common::CompressionKind_LZ4 ||
      kind == common::CompressionKind_LZO) {
    options.format.lz4_lzo.isHadoopFrameFormat = false;
  }
  return options;
}
```
We made this change to default initialize it to avoid the following compliation errors.
```
    inlined from ‘std::unique_ptr<facebook::velox::dwio::common::SeekableInputStream> TestSeek::createTestDecompressor(facebook::velox::common::CompressionKind, std::unique_ptr<facebook::velox::dwio::common::SeekableInputStream>, uint64_t)’ at /root/velox/velox/dwio/dwrf/test/TestDecompression.cpp:873:73,
    inlined from ‘virtual void TestSeek_uncompressed_Test::TestBody()’ at /root/velox/velox/dwio/dwrf/test/TestDecompression.cpp:962:20:
/root/velox/./velox/dwio/dwrf/common/Compression.h:99:10: error: ‘options’ is used uninitialized [-Werror=uninitialized]
   99 |   return options;
      |          ^~~~~~~
/root/velox/./velox/dwio/dwrf/common/Compression.h: In member function ‘virtual void TestSeek_uncompressed_Test::TestBody()’:
/root/velox/./velox/dwio/dwrf/common/Compression.h:90:22: note: ‘options’ declared here
   90 |   CompressionOptions options;
      |                      ^~~~~~~
cc1plus: all warnings being treated as errors
```

majetideepak: > I honestly don't see the value zeroing out these stuff.  
but that's not a general rule we should follow in this code base.

Initializing variables is one of the basic security principles. Velox code would not pass most security scans if this is not done. We see this a lot in IBM code scans. How does Meta handle security scans?

majetideepak: I have to clarify that without the defaults, there is a good chance users might not initialize variables properly.

velox/expression/tests/GenericViewTest.cpp
xin-zhang2: Passing an initializer list to the std::optional<T> constructor in GCC13 will fail because the constructor is explicit declared. The error (not warning) is 
```
/root/velox/velox/expression/tests/GenericViewTest.cpp:56:3: error: converting to ‘std::optional<std::vector<std::optional<long int> > >’ from initializer list would use explicit constructor ‘constexpr std::optional<_Tp>::optional(std::in_place_t, _Args&& ...) [with _Args = {}; typename std::enable_if<__and_v<std::is_constructible<_Tp, _Args ...> >, bool>::type <anonymous> = false; _Tp = std::vector<std::optional<long int> >]’
   56 |   };
      |   ^
/root/velox/velox/expression/tests/GenericViewTest.cpp:56:3: error: converting to ‘std::in_place_t’ from initializer list would use explicit constructor ‘constexpr std::in_place_t::in_place_t()’
```
That's why we have to explicitly construct it with `std::make_optional`.

Tested it in https://godbolt.org/z/qxK6b5jbY and the error starts from GCC13.1.

Yuhta: This is apparently a compiler bug that it should match `optional(U&& value)` signature but it tries to match the `in_place_t` tag version instead.  Clang is not suffering from this bug and it should be fixed in GCC.

Yuhta: In the meantime, you can do this to mitigate the readability regression while we waiting for the GCC fix (put this in `common/testutil` as it should be used in only tests):
```c++
// Workaround for GCC to avoid matching optional(std::in_place_t, ...)
struct OptionalEmptyT {
    template <typename T>
    operator std::optional<std::vector<T>>() const {
        return std::make_optional<std::vector<T>>({});
    }
};

constexpr OptionalEmptyT optionalEmpty;

int main() {
    std::vector<std::optional<std::vector<int>>> v = {optionalEmpty};
    return 0;
}
```

majetideepak:
@xin-zhang2 is this ready for review?

xin-zhang2:
@majetideepak Yes, I have updated this PR based on @Yuhta 's review.
1. Introduced `OptionalEmptyT` in `common/testutil` to workaround the GCC `std::optional<T>` constructor signature mismatch.
2. Initialize `CompressionOptions` at use site instead of definiton.

@Yuhta Can you help review the changes? Thanks!

Yuhta (approved):

Yuhta:
@xin-zhang2 Can you rebase to the newest commit?  Seems there is some conflicts

xin-zhang2:
@Yuhta rebased to the latest commit, and didn't encounter conflict issues. Can you try re-import it? Thanks.

PR #12534, opened 2025-03-05
feat(s3): Add log location configuration

czentgr:
The S3 filesystem uses the default logging configuration which creates the file in the current directory where an executable is called from. Most deployments have a custom location to write logs.
This change allows the user to specify a path that is used as the new log prefix which results in an absolute path being used for the log file creation. The default AWS log file prefix is used.


velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -54,6 +54,9 @@ class S3Config {
   /// S3FileSystem default identity.
   static constexpr const char* kDefaultS3Identity = "s3-default-identity";
 
+  /// Log location of AWS C++ SDK.
+  static constexpr const char* kS3LogLocation = "hive.s3.log-location";
```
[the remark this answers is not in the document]
czentgr: The default logger uses actually the input aws a prefix and adds a date and rotates the files every hour or so.
So it would be't be the user file name plus what the AWS client adds.

velox/connectors/hive/storage_adapters/s3fs/RegisterS3FileSystem.cpp
```diff
@@ -80,7 +80,9 @@ std::shared_ptr<FileSystem> fileSystemGenerator(
 
         auto logLevel =
             properties->get(S3Config::kS3LogLevel, std::string("FATAL"));
-        initializeS3(logLevel);
+        auto logLocation =
```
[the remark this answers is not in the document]
czentgr: Yes, I had this initially but things got complicated. Perhaps I wasn't using it correctly.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -525,14 +527,29 @@ struct AwsInstance {
...
+
  private:
-  void doInitialize(std::string_view logLevel) {
+  void doInitialize(
+      std::string_view logLevel,
+      std::optional<std::string_view> logLocation) {
     awsOptions_.loggingOptions.logLevel = inferS3LogLevel(logLevel);
+    std::filesystem::path prefix;
+    if (logLocation.has_value()) {
+      prefix = std::string(logLocation.value())
```
czentgr, line 546: Change to fmt and avoid placing a second /.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -547,6 +564,7 @@ struct AwsInstance {
   Aws::SDKOptions awsOptions_;
   std::atomic<bool> isInitialized_;
   std::atomic<bool> isFinalized_;
+  std::string logPrefix_;
```
majetideepak, line 567: We don't need this class variable.

velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemTest.cpp
```diff
@@ -182,6 +184,27 @@ TEST_F(S3FileSystemTest, logLevel) {
   checkLogLevelName("INFO");
 }
 
+TEST_F(S3FileSystemTest, logPrefix) {
```
majetideepak, line 187: nit: `logLocation`

majetideepak (commented):
@czentgr two small comments. Thanks

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
majetideepak: Makes sense. We need a buffer backing the awsOptions_  object.

PR #12536, opened 2025-03-05
fix(json): ParseHex error

kgpai:
Summary: Fixes json_parse parseHex asan error. This happens when a string is generated with an incomplete escape sequence. Also inlined the call to parseHex.

Differential Revision: D70595015


kevinwilfong (approved):
Thanks! Can you fix the title, I think it doesn't like the space in "fix (json_parse)"

PR #12537, opened 2025-03-05
misc: Add ability to remove a function from all registries

bikramSingh91:
Summary:
This addresses an internal use-case where, during initialization,
all Presto functions are registered before some are overwritten
with custom implementations. Since vector functions are prioritized
over simple functions, this approach enables replacing a previously
registered vector function with a simple function. To simplify the
API and prevent confusion between removing specific signatures versus
all, this API is designed to remove all signatures, providing a clean
slate for the user.

Reviewed By: kgpai

Differential Revision: D70594709

### DIFF
--- a/velox/expression/SimpleFunctionRegistry.cpp
+++ b/velox/expression/SimpleFunctionRegistry.cpp
@@ -68,6 +68,11 @@ bool SimpleFunctionRegistry::registerFunctionInternal(
   });
 }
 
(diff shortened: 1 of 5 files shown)


bikramSingh91:
Linux build failed due to unrelated test failing, filed a bug report for it: https://github.com/facebookincubator/velox/issues/12546

kgpai (approved):

PR #12540, opened 2025-03-05
build: Fix breeze build

dreveman:
gtest_SOURCE_DIR is now googletest_SOURCE_DIR.


dreveman:
@Yuhta Fyi, things like this can fail in breeze now and go unnoticed as we only build breeze if breeze code changes. I think that's fine for now but maybe something to address in the future. Building a single test target would be enough to protect against this.

Yuhta (approved):

dreveman:
what's need to merge this change?

PR #12541, opened 2025-03-05
feat: Minor optimize lamdba with std::move

mapleFU:


velox/expression/LambdaExpr.cpp
```diff
@@ -122,9 +122,9 @@ class ExprCallable : public Callable {
       VELOX_DCHECK(!isLazyNotLoaded(*values));
       if (wrapCapture) {
         values = BaseVector::wrapInDictionary(
-            BufferPtr(nullptr), wrapCapture, size, values);
+            BufferPtr(nullptr), wrapCapture, size, std::move(values));
       }
-      allVectors.push_back(values);
+      allVectors.push_back(std::move(values));
```
Yuhta, line 127: This is a bug, if `values` got moved 2 lines above, we get nullptr here.  `childAt` returns a reference and we should keep it that way (`auto&`), so we only copy it in `wrapInDictionary` and `push_back`

Yuhta (changes requested):

velox/expression/LambdaExpr.cpp
mapleFU: > `childAt` returns a reference

Since we don't uses

```
auto& values = capture_->childAt(index);
```

`values` type is `std::shared_ptr<>`, and this just avoid that?

Yuhta: Exactly

PR #12559, opened 2025-03-05
feat: Add support for Unknown Type in Set Union 

natashasehgal:
Add support for missing type Unknown in Set Union function


amitkdutta (approved):
Thanks @natashasehgal

PR #12561, opened 2025-03-06
refactor: Remove "extern void registerXxxAggregate" declarations from RegisterAggregateFunctions.cpp

### LINKED ISSUE #12538: Remove "extern void registerXxxAggregate" declarations from
### Description

Remove "extern void registerXxxAggregate" declarations from functions/prestosql/aggregates/RegisterAggregateFunctions.cpp in favor of #include <agg>.h

See #12528 for an example.

This would allow to cherry-pick which aggregate functions to register and no longer require using linker hack https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries

CC: @kgpai

jkhaliqi:
Registering all aggregate functions increases build size too much.
closes #12538


kgpai (approved):
Thank you for this change, I tried to be as thorough as possible but want to be sure that this only makes the headers public and removes the externs. If so looks good.

PR #12563, opened 2025-03-06
fix: HashProbe load LazyVector before wrapping

### LINKED ISSUE #12484: lazy vector: An unloaded lazy vector cannot be wrapped by two different top level vectors in hash join
### Bug description

```
    withTable("t1", "t2") {
      Seq((1, "a"), (2, "b")).toDF("c1", "c2").write.saveAsTable("t1")
      Seq(1, 2, 3).toDF("c1").write.saveAsTable("t2")
      val query =
        """
          |select t3.* from
          |(select c1, c2 as a,c2 as b from t1) t3
          |join t2
          |on t3.c1 = t2.c1
          |""".stripMargin
      sql(query).show()
    }
```
ERROR:
An unloaded lazy vector cannot be wrapped by two different top level vectors.

The vector of c2 is rewrapped here, and there seems to be HashProbe do not load LazyVector before wrapping.

### System information

-

### Relevant logs

```bash
Reason: An unloaded lazy vector cannot be wrapped by two different top level vectors.
Retriable: False
Expression: dictionaryValues_->markAsContainingLazyAndWrapped()
Context: Operator: HashProbe[3] 2
Function: setInternalState
File: /opt/gluten/ep/build-velox/build/velox_ep/./velox/vector/DictionaryVector-inl.h
Line: 35
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorEPKcEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox16DictionaryVectorINS0_10StringViewEE16setInternalStateEv
# 4  _ZN8facebook5velox16DictionaryVectorINS0_10StringViewEEC1EPNS0_6memory10MemoryPoolEN5boost13intrusive_ptrINS0_6BufferEEEmSt10shared_ptrINS0_10BaseVectorEESA_RKNS0_17SimpleVectorStatsIS2_EESt8optionalIiESJ_SI_IbESJ_SJ_
# 5  _ZN8facebook5veloxL13addDictionaryILNS0_8TypeKindE7EEESt10shared_ptrINS0_10BaseVectorEEN5boost13intrusive_ptrINS0_6BufferEEES9_mS5_
# 6  _ZN8facebook5velox10BaseVector16wrapInDictionaryEN5boost13intrusive_ptrINS0_6BufferEEES5_iSt10shared_ptrIS1_Eb
# 7  _ZN8facebook5velox4exec9wrapChildEiN5boost13intrusive_ptrINS0_6BufferEEERKSt10shared_ptrINS0_10BaseVectorEES5_
# 8  _ZN8facebook5velox4exec9HashProbe10fillOutputEi
# 9  _ZN8facebook5velox4exec9HashProbe17getOutputInternalEb
# 10 _ZN8facebook5velox4exec9HashProbe9getOutputEv
# 11 _ZZN8facebook5velox4exec6Driver11runInternalERSt10shared_ptrIS2_ERS3_INS1_13BlockingStateEERS3_INS0_9RowVectorEEENKUlvE3_clEv
# 12 _ZN8facebook5velox4exec6Driver11runInternalERSt10shared_ptrIS2_ERS3_INS1_13BlockingStateEERS3_INS0_9RowVectorEE
# 13 _ZN8facebook5velox4exec6Driver4nextEPN5folly10SemiFutureINS3_4UnitEEE
# 14 _ZN8facebook5velox4exec4Task4nextEPN5folly10SemiFutureINS3_4UnitEEE
# 15 _ZN6gluten24WholeStageResultIterator4nextEv
# 16 Java_org_apache_gluten_vectorized_ColumnarBatchOutIterator_nativeHasNext
# 17 0x00007f97526413b0
```

zml1206:
Fixes #12484


Yuhta (changes requested):
This will destroy the performance.  Can you explain what problem you are trying to solve and we can find another way to solve it.

zml1206:
> This will destroy the performance. Your optimizer should remove duplicated projections instead.


Although it does not make sense to duplicate the same fields like `select a as b, a as c`, this is also allowed by standard SQL. If the user uses this as the final output, we cannot modify it in the optimizer. So should Velox support this scenario?

Yuhta:
@zml1206 No the optimizer should use a project node to duplicate the result again for the user

zml1206:
Thank you for your answer, I will fix it in gluten @Yuhta

PR #12570, opened 2025-03-06
fix(s3): Handle trailing slash for AWS endpoint values

czentgr:
For flexibility of handling user input the region extraction should handle if the endpoint for for AWS ends in a trailing slash. Users may accidentally copy&paste the endpoint including a slash which may not be easily fixable in deployments.


majetideepak (approved):
Thanks, @czentgr

PR #12571, opened 2025-03-07
feat(fuzzer): Expand expression fuzzer skip list for fb functions

peterenescu:
Summary:
Add additional skip list
Updates ExpressionFuzzer to not error out during default null behavior mismatch

Reviewed By: kagamiori

Differential Revision: D70593746


kagamiori (approved):
LGTM. Thanks!

PR #12574, opened 2025-03-07
fix: Use shortest scientific notation for cast(real|double as varchar)

ccat3z:
Use shortest scientific notation for cast(real|double as varchar). Previously, the conversion to scientific notation used a fixed precision which led some results incompatible with IEEE754. Here is some example.

```
5.957E-4 -> 5.9570000000000001E-4
1.0E-6 -> 9.9999999999999995E-7
-7.639E-4 -> 7.6389999999999997E-4
```

Change to use `dragonbox` in fmt library instead of `fmt::format("{:.16E}")`. It provides shortest digits and better performance. I added a benchmark to compare the performance of several scientific notation conversions. DoubleToFmtExpPrecision is original impl and DoubleToScientificNotation is new impl.

```
[...]type/tests/FloatingPointBenchmark.cpp     relative  time/iter   iters/s
FloatToScientificNotation                                  99.59ms     10.04
FloatToFmt                                      84.754%   117.51ms      8.51
FloatToFmtExp                                   60.285%   165.20ms      6.05
FloatToFmtExpPrecision                          58.467%   170.34ms      5.87
DoubleToScientificNotation                                145.36ms      6.88
DoubleToFmt                                     101.58%   143.10ms      6.99
DoubleToFmtExp                                  84.720%   171.58ms      5.83
DoubleToFmtExpPrecision                         77.518%   187.52ms      5.33
```

Also added a fuzzer test to compare the results between java and velox. Due to https://bugs.openjdk.org/browse/JDK-4511638, java <= 18 will produce incorrect or longer decimal. The fuzzer ignored differences caused by this bug.


Yuhta:
@kgpai Would this cause behavior mismatch with Presto?

kgpai:
@Yuhta  There is always noise related to doubles - whether this results in more noise , Its hard to say but probably yes. Might be good to write some test and compare using presto fuzzer and see if it results in more or less matches.

ccat3z:
@kgpai I added a fuzzer test to compare results between java and velox. Due to https://bugs.openjdk.org/browse/JDK-4511638, java <= 18 will produce incorrect or longer decimal. The fuzzer ignored differences caused by this bug.

ccat3z:
@assignUser I've updated the code, could you help to review it?

ccat3z:
@majetideepak @kgpai Could you help review it?

assignUser (approved):
CMake :+1:

kgpai:
@ccat3z Now that we have migrated to Prestissimo, it might be a more opportune time to reopen this PR, if interested. Let me know if you would want to land this.

PR #12577, opened 2025-03-07
refactor: Add macro to refactor array functions registration

jinchengchenghh:


jinchengchenghh:
Can you help review this PR? Thanks! @rui-mo @zhli1142015

velox/functions/prestosql/registration/ArrayFunctionsRegistration.cpp
```diff
@@ -212,96 +201,42 @@ void registerArrayFunctions(const std::string& prefix) {
...
-  registerArrayRemoveFunctions<double>(prefix);
-  registerArrayRemoveFunctions<bool>(prefix);
-  registerArrayRemoveFunctions<Timestamp>(prefix);
-  registerArrayRemoveFunctions<Date>(prefix);
-  registerArrayRemoveFunctions<Varbinary>(prefix);
-  registerArrayRemoveFunctions<Generic<T1>>(prefix);
+  REGISTER_SCALAR_FUNCTIONS_WITHOUT_VARCHAR(
+      registerArrayRemoveFunctions, prefix);
   registerFunction<
       ArrayRemoveFunctionString,
```
zhli1142015, line 207: Do you know the reason why we need to implement a separate function specifically for the VARCHAR type?

jinchengchenghh: The StringWriter push_back API uses `copy_from` which is a deep copy, usually in the simple function case, it should be a reference copy.

jinchengchenghh: So we use setNoCopy instead of push_back for VARCHAR

velox/functions/lib/RegistrationHelpers.h
```diff
@@ -83,6 +83,36 @@ void registerUnaryNumeric(const std::vector<std::string>& aliases) {
   registerUnaryFloatingPoint<T>(aliases);
 }
 
+#define REGISTER_NUMERIC_FUNCTIONS(functionName, prefix) \
```
rui-mo, line 86: Should bool be considered numeric type?

velox/functions/lib/RegistrationHelpers.h
```diff
@@ -83,6 +83,36 @@ void registerUnaryNumeric(const std::vector<std::string>& aliases) {
...
+  }
+
+#define REGISTER_SCALAR_FUNCTIONS(functionName, prefix) \
+  {                                                     \
+    REGISTER_NUMERIC_FUNCTIONS(functionName, prefix)    \
+    functionName<Varchar>(prefix);                      \
+    functionName<Varbinary>(prefix);                    \
+    functionName<bool>(prefix);                         \
+    functionName<Timestamp>(prefix);                    \
+    functionName<Date>(prefix);                         \
```
rui-mo, line 104: Could call the `REGISTER_SCALAR_FUNCTIONS_WITHOUT_VARCHAR` and add a varchar registration.

velox/functions/lib/RegistrationHelpers.h
```diff
@@ -83,6 +83,36 @@ void registerUnaryNumeric(const std::vector<std::string>& aliases) {
...
+  }
+
+#define REGISTER_SCALAR_FUNCTIONS_WITHOUT_VARCHAR(functionName, prefix) \
```
rui-mo, line 107: Is it a common case that varchar and varbinary not being registered simultaneously?

rui-mo (commented):
Thanks.

velox/functions/lib/RegistrationHelpers.h
jinchengchenghh: No, I find in Spark, bool does not consider as numeric type

velox/functions/lib/RegistrationHelpers.h
jinchengchenghh: Yes, because this issue https://github.com/facebookincubator/velox/issues/12731, usually, VARCHAR is optimized while VARBINARY not.

jinchengchenghh: But I think both of them need to be optimized

PR #12580, opened 2025-03-07
feat: BingTile construction and property functions

jagill:
Summary:
This adds the basic BingTile functions that construct a BingTile from
coordinates (x, y, zoom) and also accesses these same coordinates, as well as
basic documentation.

Differential Revision: D70561190

### DIFF
--- a/velox/functions/prestosql/GeospatialFunctions.h
+++ b/velox/functions/prestosql/GeospatialFunctions.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/GeospatialFunctions.h
```diff
@@ -0,0 +1,79 @@
...
+
+template <typename T>
+struct BingTileZoomLevelFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<int8_t>& result,
+      const arg_type<BingTile>& tile) {
+    int64_t tileInt = (int64_t)tile;
+    result = (int8_t)BingTileType::bingTileZoom((uint64_t)tileInt);
```
bikramSingh91, line 61: nit: would be good to avoid using c-style casts here and other places in this file. If its absolutely required (like at L34) then please use static_cast<>

bikramSingh91 (approved):
Looks good, just one nit. Accepting to unblock but please address the nit before merging. Thanks!

PR #12583, opened 2025-03-07
fix: Presto UDF split_part does not match Presto behavior for empty string delimiter

kevinwilfong:
Summary:
This fixes a bunch of behavior inconsistencies between Presto Java's and Velox's split_part UDFs.

In Presto Java, when split_part is called on a string X with an empty string delimiter and index i, it 
returns the i'th character of X or null if X has fewer than i characters. It also validates that every 
character up to and including the i'th character is valid UTF-8 (or all characters if it returns null).

In Velox today, it will return the entire string X if the i is 1 and the empty string otherwise.

In Velox today, we do not check that the index is a positive value, unlike Presto Java.

In Velox today, we do not check that the search string contains valid UTF-8 characters. Note that
Presto does not check this condition when the delimiter is non-empty, Velox maintains this 
behavior after this change.

Differential Revision: D70810224

### DIFF
--- a/velox/functions/lib/string/StringImpl.h
+++ b/velox/functions/lib/string/StringImpl.h
@@ -27,6 +27,7 @@
 #include "folly/Likely.h"
 #include "velox/common/base/Exceptions.h"
 #include "velox/external/md5/md5.h"
+#include "velox/functions/lib/Utf8Utils.h"
 #include "velox/functions/lib/string/StringCore.h"
 #include "velox/type/StringView.h"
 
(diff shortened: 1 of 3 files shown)


velox/functions/lib/string/StringImpl.h
```diff
@@ -415,23 +416,50 @@ FOLLY_ALWAYS_INLINE int endsWithUnicodeWhiteSpace(
...
+      int codePoint = 0;
+      while (curPos < inputSv.size()) {
+        auto codePointSize = tryGetUtf8CharLength(
+            input.data() + curPos, input.size() - curPos, codePoint);
+        VELOX_USER_CHECK(
+            codePointSize > 0 && codePointSize <= input.size() - curPos,
+            "Invalid UTF-8 encoding in characters: {}",
+            StringView(
+                input.data() + curPos,
+                std::min(input.size() - curPos, curPos + 12)));
```
yuandagits, line 449: curious, why + 12?

IIUC, if we hit the end of the iteration, we are assuming that the remaining codePointSize shall be equal to the left over input?

kevinwilfong: I copied it from the check in the stringToCodePoints function earlier in this file, I assumed it's to give some context so that maybe you can figure out what's wrong with the string (e.g. maybe some stray bytes got inserted into the string).  It does admittedly seem sort of arbitrary.

kevinwilfong: > if we hit the end of the iteration, we are assuming that the the current codePointSize shall be equal to the left over input

or that's the most context we can give.

yuandagits (approved):

PR #12584, opened 2025-03-07
feat: Velox Array Join Unknown support

natashasehgal:
Add support for Unknown type in Velox Array Join Function


amitkdutta (approved):
Thanks @natashasehgal

PR #12600, opened 2025-03-11
fix: Fix incorrect result from casting double to decimal

zhli1142015:
When converting 0.03456789 to DECIMAL(38, 33), the result generated by Velox is 
inconsistent with Spark and Presto. This discrepancy is caused by precision 
errors due to long double multiplication. I propose fixing this issue by first 
converting the result to the target type(int64_t / int128_t) and then applying rescaling.

Velox
```
spark.sql("select cast(0.03456789d as decimal(38, 33))").collect
res18: Array[org.apache.spark.sql.Row] = Array([0.034567889999999999999845079187456])
```
Spark
```
spark.sql("select cast(0.03456789d as decimal(38, 33))").collect
res16: Array[org.apache.spark.sql.Row] = Array([0.034567890000000000000000000000000])
```
Presto
```
presto> select cast(0.03456789 as decimal(38, 33));
                _col0
-------------------------------------
 0.034567890000000000000000000000000
(1 row)
```

### DIFF
--- a/velox/type/DecimalUtil.h
+++ b/velox/type/DecimalUtil.h
@@ -271,16 +271,11 @@ class DecimalUtil {
       if (isOverflow) {
         return Status::UserError("Result overflows.");
       }
-    } else if (fractionDigits > scale) {
+    } else {
       const auto scalingFactor =
           DecimalUtil::kPowersOfTen[fractionDigits - scale];
-      const TOutput remainder = rescaledValue % scalingFactor;
-      rescaledValue /= scalingFactor;
-      if (rescaledValue >= 0 && remainder >= scalingFactor / 2) {
-        ++rescaledValue;
-      } else if (remainder <= -scalingFactor / 2) {
-        --rescaledValue;
-      }
+      divideWithRoundUp<TOutput, TOutput, int128_t>(
+          rescaledValue, rescaledValue, scalingFactor, false, 0, 0);
     }
 
     if (!valueInPrecisionRange<TOutput>(rescaledValue, precision)) {
--- a/velox/type/DecimalUtil.h
+++ b/velox/type/DecimalUtil.h
@@ -258,17 +258,31 @@ class DecimalUtil {
     // LONG_DOUBLE_MAX.
     long double scaledValue = std::round(
         (long double)value * DecimalUtil::kPowersOfTen[fractionDigits]);
-    if (scale > fractionDigits) {
-      scaledValue *= DecimalUtil::kPowersOfTen[scale - fractionDigits];
-    } else {
-      scaledValue /= DecimalUtil::kPowersOfTen[fractionDigits - scale];
-    }
-
-    const auto result = folly::tryTo<TOutput>(std::round(scaledValue));
+    const auto result = folly::tryTo<TOutput>(scaledValue);
     if (result.hasError()) {
       return Status::UserError("Result overflows.");
     }
-    const TOutput rescaledValue = result.value();
+    TOutput rescaledValue = result.value();
+    if (scale > fractionDigits) {
+      bool isOverflow = __builtin_mul_overflow(
+          rescaledValue,
+          DecimalUtil::kPowersOfTen[scale - fractionDigits],
+          &rescaledValue);
+      if (isOverflow) {
+        return Status::UserError("Result overflows.");
+      }
+    } else if (fractionDigits > scale) {
+      const auto scalingFactor =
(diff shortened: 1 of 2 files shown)


mbasmanova:
@zhli1142015 

> The CI failure is unrelated.

Is there an issue about this failure? If not, would you create one and reference it here?

zhli1142015:
> @zhli1142015
> 
> > The CI failure is unrelated.
> 
> Is there an issue about this failure? If not, would you create one and reference it here?

https://github.com/facebookincubator/velox/issues/12603
Thanks.

velox/type/DecimalUtil.h
```diff
@@ -258,17 +258,31 @@ class DecimalUtil {
...
     long double scaledValue = std::round(
         (long double)value * DecimalUtil::kPowersOfTen[fractionDigits]);
-    if (scale > fractionDigits) {
-      scaledValue *= DecimalUtil::kPowersOfTen[scale - fractionDigits];
-    } else {
-      scaledValue /= DecimalUtil::kPowersOfTen[fractionDigits - scale];
-    }
-
-    const auto result = folly::tryTo<TOutput>(std::round(scaledValue));
+    const auto result = folly::tryTo<TOutput>(scaledValue);
```
rui-mo, line 261: If we convert `scaledValue` as target type before rescaling, will unexpected overflow happen? For example, a value fits in int64_t after dividing the rescale factor but overflows before the rescaling.

velox/type/DecimalUtil.h
```diff
@@ -258,17 +258,31 @@ class DecimalUtil {
...
+      }
+    } else if (fractionDigits > scale) {
+      const auto scalingFactor =
+          DecimalUtil::kPowersOfTen[fractionDigits - scale];
+      const TOutput remainder = rescaledValue % scalingFactor;
+      rescaledValue /= scalingFactor;
+      if (rescaledValue >= 0 && remainder >= scalingFactor / 2) {
+        ++rescaledValue;
+      } else if (remainder <= -scalingFactor / 2) {
+        --rescaledValue;
```
rui-mo, line 282: Perhaps also add test for `rescaledValue` being negative.

rui-mo (commented):
Thanks.

velox/type/DecimalUtil.h
```diff
@@ -258,17 +258,31 @@ class DecimalUtil {
...
+      bool isOverflow = __builtin_mul_overflow(
+          rescaledValue,
+          DecimalUtil::kPowersOfTen[scale - fractionDigits],
+          &rescaledValue);
+      if (isOverflow) {
+        return Status::UserError("Result overflows.");
+      }
+    } else if (fractionDigits > scale) {
+      const auto scalingFactor =
+          DecimalUtil::kPowersOfTen[fractionDigits - scale];
```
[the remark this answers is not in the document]
zhli1142015: I think the most appropriate function to use here is rescaleWithRoundUp, but we need to return different error messages. Therefore, this logic can only be implemented here.

velox/type/DecimalUtil.h
zhli1142015: Taking double as an example, it allows up to 17 digits, while int64_t allows 19 digits. So, I think this is safe and will not cause an overflow.

rui-mo: > long double scaledValue = std::round(
        (long double)value * DecimalUtil::kPowersOfTen[fractionDigits]);
 
If `value` is `kMaxDoubleBelowInt64Max`, after multiplying 10^15, the result will overflow for int64_t, no?

velox/type/DecimalUtil.h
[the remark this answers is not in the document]
jinchengchenghh: So we could use divideWithRoundUp for `fractionDigits > scale`, it won't throw exception when `aRescale = 0`

velox/type/DecimalUtil.h
zhli1142015: I think in this case the fractionDigits is 0.
`const auto fractionDigits = std::max(digits - integralDigits, 0);`

zhli1142015: a test case for cast (kMaxDoubleBelowInt64Max as short decimal) is added.

rui-mo (approved):

jinchengchenghh (approved):

zhli1142015:
@Yuhta , could you please help review this PR?

Yuhta (approved):

PR #12601, opened 2025-03-11
fix(dwio): Buffer corruption when new capacity is less than old size

nmahadevuni:
In BufferUtil.h, ensureCapacity() method, when new buffer size is less than the old size, and clearBits is set to true, the buffer gets corrupted because `capacity - oldSize` will be negative.


nmahadevuni:
@yingsu00 @Yuhta Can you please review this?

Yuhta (approved):

PR #12602, opened 2025-03-11
feat: Extract hive connector ID from task meta

duanmeng:
Extracts the hive connector ID from the task meta and registers
the hive connector using it when replaying the `TableScan` operator.
Hence, the query trace replayer can register the right hive connector
instead of the default 'test-hive' connector.

Resolve #12586

### DIFF
--- a/velox/exec/TraceUtil.h
+++ b/velox/exec/TraceUtil.h
@@ -88,6 +88,14 @@ std::string getNodeName(
     const std::shared_ptr<filesystems::FileSystem>& fs,
     memory::MemoryPool* pool);
 
+/// Extracts the hive connector from the trace metadata file, return empty if
+/// it does not exist.
+std::string getHiveConnectorId(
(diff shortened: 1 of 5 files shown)


velox/exec/TraceUtil.h
```diff
@@ -88,6 +88,14 @@ std::string getNodeName(
     const std::shared_ptr<filesystems::FileSystem>& fs,
     memory::MemoryPool* pool);
 
+/// Extracts the hive connector from the trace metadata file, return empty if
+/// it does not exist.
+std::string getHiveConnectorId(
```
xiaoxmeng, line 93: It seems that we parse task metadata in multiple utilities. Can we consolidate this to parse once and return all the required fields? Thanks!

xiaoxmeng (approved):
@duanmeng LGTM. Thanks!

PR #12608, opened 2025-03-11
refactor: move json_extract_scalar back to simple function

bikramSingh91:
Summary:
We initially moved json_extract_scalar to a vector function
implementation so that we can reuse the JSON parsing logic and apply
it to VARCHAR inputs. JSON parsing achieves the following for VARCHAR
input: (1) it formats spaces, (2) it sorts the keys inside objects,
(3) it flags invalid JSON, and (4) it unescapes string scalars. For
json_extract_scalar, we only care about (3) and (4) since the output
is a scalar. Therefore, we can avoid the costly operation of
rewriting the large input JSON in a normalized form and only apply
the operations we require. To achieve (3), SIMD JSON can be used to
quickly check for valid JSON, and for (4), SIMD JSON provides an API
to extract a string scalar after removing escaping.

This change moves json_extract_scalar back to simple function
implementation and ensures (3) and (4) are applied.

Differential Revision: D70978059


velox/functions/prestosql/JsonFunctions.h
```diff
@@ -169,6 +169,86 @@ struct JsonArrayGetFunction {
...
+// (boolean, number or string)
+template <typename T>
+struct JsonExtractScalarFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE bool call(
+      out_type<Varchar>& result,
+      const arg_type<Json>& json,
+      const arg_type<Varchar>& jsonPath) {
+    return callImpl(result, json, jsonPath) == simdjson::SUCCESS;
```
kgpai, line 184: nit: about the return type from callImpl being bool and comparing that to simdjson::returntype .

kgpai (approved):
Please fix nits, looks good otherwise.

PR #12609, opened 2025-03-11
fix: Prevent memory leak caused by cyclic reference of driver in DriverBlockingState

zhztheplayer:
`Task::DriverBlockingState::setDriverFuture` captures `driver_->shared_from_this` then pass to lambdas. In practice this could easily cause cyclic references on `std::shared_ptr<Driver>` because the physical operators could at the same time hold the associated promises that refer to the `driverFuture` instances. Hence, in the case, the reference link will be as following:

Operator -> Promise -> [Driver](https://github.com/facebookincubator/velox/blob/c550daba47db2a17221c8ec0144450f32bd9a56a/velox/exec/Task.cpp#L3242-L3255) -> Query Plan -> Operator

If the operator may have some cleanup logic in its destructor, say, to close the remote connection then to set the promise to unblock the future, this reference cycle will cause the cleanup logic never to run. Then memory leak will be led.

The patch reduces the capture granularity to make the driver be always correctly destructed. Then the query plan can be destructed, then the cleanup logic can be run, etc.

### DIFF
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -3239,7 +3239,7 @@ void Task::DriverBlockingState::setDriverFuture(ContinueFuture& driverFuture) {
   }
   std::move(driverFuture)
       .via(&folly::InlineExecutor::instance())
-      .thenValue([&, holder = shared_from_this()](auto&& /* unused */) {
+      .thenValue([this, holder = shared_from_this()](auto&& /* unused */) {
         std::vector<std::unique_ptr<ContinuePromise>> promises;
         {
           std::lock_guard<std::mutex> l(mutex_);
@@ -3254,8 +3254,9 @@ void Task::DriverBlockingState::setDriverFuture(ContinueFuture& driverFuture) {
       })
       .thenError(
           folly::tag_t<std::exception>{},
-          [&, holder = shared_from_this(), taskId = driver_->task()->taskId()](
-              std::exception const& e) {
+          [this,
+           holder = shared_from_this(),
+           taskId = driver_->task()->taskId()](std::exception const& e) {
             std::lock_guard<std::mutex> l(mutex_);
             VELOX_CHECK(blocked_);
             VELOX_CHECK_NULL(error_);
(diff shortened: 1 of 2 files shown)


zhztheplayer:
Would someone help review this? Thanks!

@xiaoxmeng @Yuhta @kagamiori

velox/exec/Task.cpp
```diff
@@ -3239,31 +3239,30 @@ void Task::DriverBlockingState::setDriverFuture(ContinueFuture& driverFuture) {
...
-              VELOX_CHECK(blocked_);
-              VELOX_CHECK_NULL(error_);
-              promises = std::move(promises_);
-              blocked_ = false;
-            }
-            for (auto& promise : promises) {
-              promise->setValue();
-            }
-          })
+      .thenValue([&, holder = shared_from_this()](auto&& /* unused */) {
```
xiaoxmeng, line 3242:
```
s/&/this/
```

xiaoxmeng (commented):
@zhztheplayer LGTM. Please make sure when don't access driver state in driver block state future callback. Thanks % init.

xiaoxmeng (approved):

xiaoxmeng (changes requested):
@zhztheplayer I think it is intended to hold a reference to the driver from the async event to ensure the driver is still there? Why the async event relies on operator close to cleanup? It should be self-cleanup by timeout or other mechanism. Doesn't seem right use pattern? Thanks!

zhztheplayer:
> @zhztheplayer I think it is intended to hold a reference to the driver from the async event to ensure the driver is still there? Why the async event relies on operator close to cleanup? It should be self-cleanup by timeout or other mechanism. Doesn't seem right use pattern? Thanks!

@xiaoxmeng Thank you for the insights.

There is `Operator::close` to place the cleanup logic either. Though it doesn't propagate to [DataSource](https://github.com/facebookincubator/velox/blob/main/velox/connectors/Connector.h#L230). Do you think we can add a close API to `DataSource`? So I could put the cleanup logic to it to avoid this PR's change.

zhztheplayer:
In addition, to address the comments:

> Why the async event relies on operator close to cleanup? It should be self-cleanup by timeout or other mechanism.

I feel this is very normal practice of [structured concurrency](https://en.wikipedia.org/wiki/Structured_concurrency). Operator should gurantee the async events it started have finished before the operator is destructed. E.g., task gets cancelled by user, then the async events should be also ceaced.

> I think it is intended to hold a reference to the driver from the async event to ensure the driver is still there?

I think operators already have enough flexibility to capture the driver into its async events by themselves?

zhztheplayer:
Replaced by https://github.com/facebookincubator/velox/pull/12701.

PR #12611, opened 2025-03-11
feat: Support prefetch in index lookup join

xiaoxmeng:
Summary:
This PR adds prefetch for index join with query config. The index join operator can prefetch up to the
configured prefetch limit to enable: (1) parallel prefetches  at backend for parallel execution (1-1) in case
of multiple backend shards or (1-2) enable backend to batch multiple requests to improve throughput;
(2) pipeline the table scan and index lookup execution in the same driver pipeline. The table scan is sync
executed while index lookup is async. This achieve pipelining without relying exchange which might cause
non-deterministic execution and Meta internal use case needs deterministic execution for checkpointing.
With Meta internal testing, this can achieve 2x throughput improvement (measured in rows per second)
with 33% memory overhead with up to 4 batches prefetch.

The follow is to add memory based prefetch throttling to integrate with Meta internal
ML use case and memory pool wiring to ease performance (memory overhead) analysis

Differential Revision: D70909786

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -166,6 +166,11 @@ Generic Configuration
      - false
      - By default, if a key is found in multiple given maps, that key's value in the resulting map comes from the last one of those maps.
        If true, throws exception when duplicate keys are found. This configuration is needed by Spark functions `CreateMap`, `MapFromArrays`, `MapFromEntries`, `StringToMap`, `MapConcat`, `TransformKeys`.
+   * - index_lookup_join_max_prefetch_batches
+     - integer
+     - 0
+     - Specifies the max number of input batches to prefetch to do index lookup ahead. If it is zero,
+       then process one input batch at a time.
 
 .. _expression-evaluation-conf:
 
--- a/velox/exec/IndexLookupJoin.h
+++ b/velox/exec/IndexLookupJoin.h
@@ -29,19 +29,14 @@ class IndexLookupJoin : public Operator {
 
   BlockingReason isBlocked(ContinueFuture* future) override;
 
-  bool needsInput() const override {
-    if (noMoreInput_ || input_ != nullptr) {
-      return false;
-    }
-    return true;
-  }
+  bool needsInput() const override;
 
   void addInput(RowVectorPtr input) override;
 
   RowVectorPtr getOutput() override;
 
   bool isFinished() override {
-    return noMoreInput_ && input_ == nullptr;
+    return noMoreInput_ && (numInputBatches() == 0);
   }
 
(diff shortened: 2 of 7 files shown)


velox/docs/configs.rst
```diff
@@ -166,6 +166,11 @@ Generic Configuration
      - false
      - By default, if a key is found in multiple given maps, that key's value in the resulting map comes from the last one of those maps.
        If true, throws exception when duplicate keys are found. This configuration is needed by Spark functions `CreateMap`, `MapFromArrays`, `MapFromEntries`, `StringToMap`, `MapConcat`, `TransformKeys`.
+   * - index_lookup_join_max_prefetch_batches
+     - integer
+     - 0
```
zacw7, line 171: Is the default value 0? But in QueryConfig.h the default value is 4 right?

velox/exec/IndexLookupJoin.h
```diff
@@ -84,6 +114,35 @@ class IndexLookupJoin : public Operator {
...
+  }
+
+  // Returns the current processing batch.
+  InputBatchState& currentInputBatch() {
+    return inputBatches_[startBatchIndex_ % maxNumInputBatches_];
+  }
+
+  const InputBatchState& currentInputBatch() const {
+    return inputBatches_[startBatchIndex_ % maxNumInputBatches_];
+  }
```
zacw7, line 144: n00b q: why do we need to `currentInputBatch()`?

velox/exec/IndexLookupJoin.h
```diff
@@ -107,8 +167,16 @@ class IndexLookupJoin : public Operator {
...
-  // The reused row vector for lookup input.
-  RowVectorPtr lookupInput_;
+
+  // The input batches to process with ranges pointed by 'startBatchIndex_' and
+  // 'endBatchIndex_'.
+  std::vector<InputBatchState> inputBatches_;
+
+  // Points to the input batches in 'inputBatches_' for processing with range of
+  // '[startBatchIndex_ % maxNumInputBatches_, endBatchIndex_ %
+  // maxNumInputBatches_ - 1)'.
```
zacw7, line 177: Could you pls help me understand this part?

If `startBatchIndex_` is 5 and `endBatchIndex_` is 7 and `index_lookup_join_max_prefetch_batches` is set to 3.
Does that mean `[startBatchIndex_ % maxNumInputBatches_, endBatchIndex_ % maxNumInputBatches_ - 1)` is `[5%3, 7%(3-1))` -> `[2, 1)`?

zacw7 (approved):

PR #12622, opened 2025-03-12
feat: Add support for Unknown Value in map_entries

natashasehgal:
Add support for Unknown type in Velox function map_entries

Prestissimo output matches Presto
 Following query match Presto output. The error is the same for 3 invalid values that are commented out below.
```
SELECT
           ->     MAP_ENTRIES(MAP(x, y))
           -> FROM (
           ->     VALUES
           ->
           ->         (NULL, ARRAY[]),
           ->         (NULL, ARRAY[3]),
           ->         (NULL, ARRAY[3, 4, 5]),
           ->         (ARRAY[], NULL),
           ->         (ARRAY[], ARRAY[]),
           ->         (ARRAY[1, 2, 3], ARRAY[4, 5, 6])
           ->         -- (NULL) -- not a valid case
           ->         -- (ARRAY[NULL], ARRAY[1]) -- Key cannot be null
           ->         -- (ARRAY[NULL, 2], ARRAY[1,2]) -- Key cannot be null
           -> ) AS t(x, y);
```

> Output
>                              _col0
> NULL
> NULL
> NULL
> NULL
> []
> [{field0=1, field1=4}, {field0=2, field1=5}, {field0=3, field1=6}]

### DIFF
--- a/velox/functions/prestosql/MapEntries.cpp
+++ b/velox/functions/prestosql/MapEntries.cpp
@@ -51,7 +51,7 @@ class MapEntriesFunction : public exec::VectorFunction {
   static std::vector<std::shared_ptr<exec::FunctionSignature>> signatures() {
     // map(K,V) -> array(row(K,V))
     return {exec::FunctionSignatureBuilder()
-                .knownTypeVariable("K")
+                .typeVariable("K")
                 .typeVariable("V")
                 .returnType("array(row(K,V))")
                 .argumentType("map(K,V)")
--- a/velox/functions/prestosql/tests/MapEntriesTest.cpp
+++ b/velox/functions/prestosql/tests/MapEntriesTest.cpp
@@ -125,6 +125,25 @@ TEST_F(MapEntriesTest, constant) {
   test::assertEqualVectors(expected, result);
 }
 
+TEST_F(MapEntriesTest, unknownType) {
+  auto mapVector = makeMapVector<UnknownValue, UnknownValue>({
+      {},
+  });
+  auto result = evaluate("map_entries(c0)", makeRowVector({mapVector}));
+  auto keyVector = makeNullableFlatVector<UnknownValue>({});
+  auto valueVector = makeNullableFlatVector<UnknownValue>({});
+  auto elementVector = makeRowVector({keyVector, valueVector});
+  auto arrayVector = makeSingleRowArrayVector(elementVector);
+  test::assertEqualVectors(arrayVector, result);
+}
+
+TEST_F(MapEntriesTest, unknownTypeNonNullValue) {
+  auto mapVector = makeAllNullMapVector(1, UNKNOWN(), BIGINT());
+  auto result = evaluate("map_entries(c0)", makeRowVector({mapVector}));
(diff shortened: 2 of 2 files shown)


velox/functions/prestosql/MapEntries.cpp
```diff
@@ -51,7 +51,7 @@ class MapEntriesFunction : public exec::VectorFunction {
   static std::vector<std::shared_ptr<exec::FunctionSignature>> signatures() {
     // map(K,V) -> array(row(K,V))
     return {exec::FunctionSignatureBuilder()
-                .knownTypeVariable("K")
```
kgpai, line 54: Are we sure about this ? Maps cannot have null keys and Unknown's represent null values right ? Can you add an example query from Presto which using map_entries which works with unknown type ?  Thanks !

natashasehgal: Hmm I'm not getting an error using null keys, even if the value is non-null. Is this a bug on Presto? 

 SELECT MAP_ENTRIES(MAP(x,y)) FROM (VALUES (NULL,ARRAY[])) AS t(x,y);
 SELECT MAP_ENTRIES(MAP(x,y)) FROM (VALUES (NULL,ARRAY[3])) AS t(x,y);
 SELECT MAP_ENTRIES(MAP(x,y)) FROM (VALUES (NULL,ARRAY[3,4,5])) AS t(x,y);
 
 The result is NULL

natashasehgal: I can modify the signature to keep knownType as Key and Value, and add another signature for Unknown Key and Value. 

But surprised these queries are working if a map can't have null keys. Are there any other examples I should try on Presto?

kgpai: TIL: So unknown is supported in a map but null keys are not : map() -> map<unknown, unknown>() . 
I wanted to be sure about this, so thanks for digging into this.

velox/functions/prestosql/tests/MapEntriesTest.cpp
```diff
@@ -125,6 +125,18 @@ TEST_F(MapEntriesTest, constant) {
...
 
+TEST_F(MapEntriesTest, unknownValue) {
+  auto mapVector = makeMapVector<UnknownValue, UnknownValue>({
+      {},
+  });
+  auto result = evaluate("map_entries(c0)", makeRowVector({mapVector}));
+  auto keyVector = makeNullableFlatVector<UnknownValue>({});
+  auto valueVector = makeNullableFlatVector<UnknownValue>({});
+  auto elementVector = makeRowVector({keyVector, valueVector});
+  auto arrayVector = makeSingleRowArrayVector(elementVector);
```
kgpai, line 136: Can you also add a test related to your query : `SELECT MAP_ENTRIES(MAP(x,y)) FROM (VALUES (NULL,ARRAY[3])) AS t(x,y);` and ensure that map_entires returns null in this case.

natashasehgal: Added a test case for this below

kgpai (approved):
LGTM. Thanks !

PR #12624, opened 2025-03-12
fix: Improve global arbitration to have fairer abort

tanjialiang:
Summary: When a small participant requesting for a large amount of capacity, we shall take into consideration of the requesting capacity make it more prone for being killed by global arbitration. This would make global arbitration victim selection more fair.

Differential Revision: D71076051

### DIFF
--- a/velox/common/memory/ArbitrationParticipant.h
+++ b/velox/common/memory/ArbitrationParticipant.h
@@ -260,6 +260,14 @@ class ArbitrationParticipant
   /// waiting operation to start execution if there is one.
   void finishArbitration(ArbitrationOperation* op);
 
+  /// Invoked to set the capacity 'this' participant attempts to grow through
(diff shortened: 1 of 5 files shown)


velox/common/memory/ArbitrationParticipant.h
```diff
@@ -260,6 +260,12 @@ class ArbitrationParticipant
   /// waiting operation to start execution if there is one.
   void finishArbitration(ArbitrationOperation* op);
 
+  void setGlobalArbitrationGrowCapacity(int64_t growCapacity);
```
xiaoxmeng, line 263:
```
void setPendingArbitrationGrowCapacity()
```

xiaoxmeng (approved):
@tanjialiang LGTM. Have you tested with internal workload? Thanks!

Issue #12620, opened 2025-03-12
Out of range vector access for complicated Parquet types in getParquetColumnInfo


wraymo:
### Bug description

Suppose we create this table and insert some data using Java workers.
```
CREATE TABLE test_array_row ( id INT, data ARRAY(ROW(field1 INT, field2 VARCHAR)) ) WITH ( format = 'PARQUET' );
INSERT INTO test_array_row (id, data) VALUES (1, ARRAY[ROW(10, 'Alice'), ROW(20, 'Bob')]), (2, ARRAY[ROW(30, 'Charlie'), ROW(40, 'David')])
```
When running a simple query using Velox workers and column index based mapping
```
select * from test_array_row;
```
I got error 
```
Query 20250312_145339_00016_enrin failed:  Operator::getOutput failed for [operator: TableScan, plan node ID: 0]: vector::_M_range_check: __n (which is 5) >= this->size() (which is 4)
```
But if I create the table using velox workers, everything works fine.

A direct solution would be adding more parquet reserved words

### System information

Velox System Info v0.0.2
Commit: 6010f956c3735fdadf6a4a0d24a2e4e87f5ea9c6
CMake Version: 3.28.3
System: Linux-6.1.112+
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 12.3.0
C Compiler: /usr/bin/cc
C Compiler Version: 12.3.0
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib/python3.9/dist-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

Yuhta:
@majetideepak Would you take a look?

majetideepak:
@Yuhta Yes, I will take a look.

PR #12625, opened 2025-03-12
fix: Add Parquet reserved keywords

wraymo:
Summary
This PR adds Parquet reserved keywords to handle Parquet files generated by the Presto Java workers.

Fixes: #12620.

References
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/convert/HiveSchemaConverter.java#L154
https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/serde/ParquetHiveSerDe.java#L78


majetideepak:
@wraymo Thanks for the fix. Can you add a test?

wraymo:
@majetideepak Just added a test. Could you take a look?

majetideepak (approved):
@wraymo Thanks!

PR #12629, opened 2025-03-13
feat: Throw unsupported type error for unregistered type.

amitkdutta:
Summary: Throwing user error when unregistered type is found and fail on unsupported type is true. This helps to fail a presto query with user errors instead of generic errors, thus helping better classificaiton of user and system errors.

Differential Revision: D71096915


Yuhta (approved):
We might need some new flag on error to indicate missing features.  User error is meant to represent some error user made in query and they can fix it by modifying their query.

kgpai (approved):

PR #12632, opened 2025-03-13
Added memchr.S which implements memchr optimized for Grace CPU

izard:
Added memchr.S which implements memchr optimized for Grace CPU.
Changed handling of some strings operations to use this memchr. This adds ~15% and ~25% performance benefits to two Like subbenchmarks. This change was discussed with Yifan Yuan last week.

Downsides of this patch:
1. short string handling becomes somewhat slower, but I could not measure more than 1% perf difference on other benchmarks.
2. This won't compile on AARCH64 CPUs that do not implement SHA512 extensions. (Though I don't know any modern datacenter ARM CPUs that lack it).
3. I am not very familiar with Velox structure so I might have chosen a wrong place to put optimized memchr.S file.


assignUser (changes requested):
Please approve and run CI before merging @YifanYuan3

PR #12544, opened 2025-03-05
feat: Add UUID to Varbinary casting

xin-zhang2:
Fix #10585 .

PR #12633, opened 2025-03-13
fix(UUID): Incorrect results when cast between UUID and VARBINARY

xin-zhang2:
The support for cast between UUID and VARBINARY was added in  #12544. However, there is a mismatch in the results between Prestissimo and Presto Java, which results in the following e2e test failure in https://github.com/prestodb/presto/pull/23301.
```
java.lang.AssertionError: For query:
SELECT CAST(CAST(c_uuid AS uuid) AS VARBINARY) FROM tmp_presto_68ac2ed72d41449bba7c9de1f0530413
actual column types:
[varbinary]
expected column types:
[varbinary]

not equal
Actual rows (4 of 4 extra rows shown, 5 rows in total):
[java.nio.HeapByteBuffer[pos=0 lim=36 cap=36]]
[java.nio.HeapByteBuffer[pos=0 lim=36 cap=36]]
[java.nio.HeapByteBuffer[pos=0 lim=36 cap=36]]
[java.nio.HeapByteBuffer[pos=0 lim=36 cap=36]]
Expected rows (4 of 4 missing rows shown, 5 rows in total):
[java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]]
[java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]]
[java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]]
[java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]]
```

The reason is that #12544 called the castToString and castFromString functions for VARBINARY which is incorrect. Although VARCHAR and VARBINARY are both represented by StringView, we don't need the lexical cast for VARBINARY.
This PR will fix this issue.

### DIFF
--- a/velox/functions/prestosql/types/UuidRegistration.cpp
+++ b/velox/functions/prestosql/types/UuidRegistration.cpp
@@ -73,34 +73,6 @@ class UuidCastOperator : public exec::CastOperator {
 
  private:
   template <TypeKind KIND>
-  static void castFromTypeKind(
-      const BaseVector& input,
-      exec::EvalCtx& context,
-      const SelectivityVector& rows,
-      BaseVector& result) {
-    auto* flatResult = result.as<FlatVector<int128_t>>();
-    const auto* uuidInput = input.as<SimpleVector<StringView>>();
-
-    context.applyToSelectedNoThrow(rows, [&](auto row) {
-      const auto uuidValue = uuidInput->valueAt(row);
-
-      int128_t u;
-
-      if (KIND == TypeKind::VARCHAR) {
-        auto uuid = boost::lexical_cast<boost::uuids::uuid>(uuidValue);
-        memcpy(&u, &uuid, 16);
-      } else if (KIND == TypeKind::VARBINARY) {
-        memcpy(&u, uuidValue.data(), 16);
-      }
-
-      // Convert a big endian value from Boost to native byte-order.
-      u = DecimalUtil::bigEndian(u);
-
-      flatResult->set(row, u);
-    });
-  }
-
-  template <TypeKind KIND>
   static void castToTypeKind(
       const BaseVector& input,
       exec::EvalCtx& context,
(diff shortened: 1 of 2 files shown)


xin-zhang2:
@aditi-pandit @amitkdutta @kevinwilfong @jkhaliqi 
Could you help review this PR?

velox/functions/prestosql/types/UuidRegistration.cpp
```diff
@@ -116,6 +118,28 @@ class UuidCastOperator : public exec::CastOperator {
     });
   }
 
+  static void castToVarbinary(
```
aditi-pandit, line 121: There is a lot of repetition between castTo(From)Varbinary and castTo(From)String. Consider rewriting this code with a template parameter for the type.

aditi-pandit (commented):
@xin-zhang2 : Thanks for debugging this issue. Have one high level comment.

velox/functions/prestosql/types/UuidRegistration.cpp
xin-zhang2: @aditi-pandit I've made an update. Could you please review it? Thanks.

aditi-pandit (commented):
@xin-zhang2 : One more refactoring comment.

velox/functions/prestosql/types/UuidRegistration.cpp
```diff
@@ -79,58 +82,67 @@ class UuidCastOperator : public exec::CastOperator {
...
-
-        offset += 2;
-        if (i == 3 || i == 5 || i == 7 || i == 9) {
-          result.data()[offset] = '-';
-          offset++;
+      const auto* uuidBytes = reinterpret_cast<const uint8_t*>(&uuid);
+
+      exec::StringWriter writer(flatResult, row);
+
+      if constexpr (KIND == TypeKind::VARCHAR) {
```
aditi-pandit, line 91: It would be great to abstract the part in the conditional as a function. This smaller templated function (with specialized implementations of the different type KIND) can be called from the bigger outer function.

xin-zhang2: @aditi-pandit Thanks for the review. 
I've updated the code based on your comment. Could you take a look? Thank!

aditi-pandit (approved):
Thanks @xin-zhang2. Code looks good.

amitkdutta (approved):
Looks good. Thanks @xin-zhang2

PR #12640, opened 2025-03-13
fix: Constructor of std::vector requires type to be complete

yuxuanchen1997:
Summary:
This default argument in `std::vector<IcebergDeleteFile> deletes = {}` causes `vector::~vector` to be instantiated. libc++ enforces that the type argument must be complete by the time its members are referenced. This code should have been ill-formed. 

This is fixable by just including the necessary headers.

Differential Revision: D71142041


pedroerp (approved):
Looks good to me assuming all builds are healthy. Thank you

PR #10866, opened 2024-08-27
feat(build): Switch to use C++20 standard

czentgr:
Fixes as a result:
- remove usage of volatile variables where not needed
- replace usage of deprecated std::is_pod_v
- fix deprecated usage of implicit capture of ‘this’ via ‘[=]’
- fix ambiguity when calling == operator for LambdaTypedExpr
- rearrange some link libraries order to prevent undefined symbols
- handle AppleClang versions overloading operator<< for sys_seconds input
- handle AppleClang format_to function ambiguity between fmt and system headers
- disable folly coroutine headers because the library is not built to support it

Resolves: https://github.com/facebookincubator/velox/issues/10814

PR #12644, opened 2025-03-14
fix: Suppress template-id-cdtor warnings

kou:
We can't specify template ID in constructor/destructor in C++20.

Here is a sample warning message with g++ 14.2.0:

    velox/expression/VectorReaders.h:118:36: warning: template-id not allowed for constructor in C++20 [-Wtemplate-id-cdtor]
      118 |   explicit ConstantVectorReader<T>(ConstantVector<exec_in_t>& vector) {
          |                                    ^~~~~~~~~~~~~~~~~~~~~~~~~
    velox/expression/VectorReaders.h:118:36: note: remove the '< >'

### DIFF
--- a/velox/common/fuzzer/ConstrainedGenerators.h
+++ b/velox/common/fuzzer/ConstrainedGenerators.h
@@ -68,7 +68,7 @@ template <typename T>
 class RandomInputGenerator<T, std::enable_if_t<std::is_same_v<T, StringView>>>
     : public AbstractInputGenerator {
  public:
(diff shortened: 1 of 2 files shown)


kou:
FYI: https://github.com/facebookincubator/velox/pull/10866 is also related to C++20 but it doesn't include this change.

kou:
Thanks. I close this in favor of #10866.

czentgr:
@kou Thank you. if you would be so kind and also review #10866 because you've appear to have looked into C++20 as well.

kou:
OK! I'll do it!

PR #12649, opened 2025-03-14
feat: Support encodings in inputs of MapVector::update

Yuhta:
Differential Revision: D71196967

This is to support the Meta internal delta update.


pedroerp:
@Yuhta can you update the title and summary here to reflect the external changes?

pedroerp (approved):
Thanks for adding the tests!

PR #12650, opened 2025-03-14
misc: Throw user error for reduce function limitation

amitkdutta:
Differential Revision: D71208389


xiaoxmeng (approved):
@amitkdutta thanks!

PR #12652, opened 2025-03-14
feat: Implement array_max_by and array_min_by in Velox

HeidiHan0000:
Summary: As title

Differential Revision: D71105239

### DIFF
--- a/velox/functions/prestosql/ArrayMinMaxBy.cpp
+++ b/velox/functions/prestosql/ArrayMinMaxBy.cpp
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <boost/range/irange.hpp>
+#include <folly/container/F14Set.h>
+#include <velox/type/SimpleFunctionApi.h>
+#include "velox/expression/EvalCtx.h"
+#include "velox/expression/Expr.h"
+#include "velox/expression/VectorFunction.h"
+#include "velox/functions/lib/LambdaFunctionUtil.h"
+#include "velox/functions/lib/RowsTranslationUtil.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+template <bool isMaxBy>
+class ArrayMinMaxByFunction : public exec::VectorFunction {
+ public:
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& /*outputType*/,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    VELOX_CHECK_EQ(args.size(), 2);
+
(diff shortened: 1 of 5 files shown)


velox/functions/prestosql/ArrayMinMaxBy.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::vector<int> resultindices;
+    const SelectivityVector inputElementRows =
+        toElementRows(inputElements.size(), rows, &inputArray);
+    exec::LocalDecodedVector decodedElements(
+        context, inputElements, inputElementRows);
+    const auto* baseElementsVector = decodedElements->base();
+
+    // Allocate new vectors for indices.
+    BufferPtr indices = allocateIndices(inputElements.size(), context.pool());
+    vector_size_t* rawIndices = indices->asMutable<vector_size_t>();
```
kevinwilfong, line 116: It looks like you're only using this as a temporary so you can iterate over it. We already depend on boost you could use an irange instead:

```
      const auto range = boost::irange(offset, offset + size);
      int minMaxIndex =
          *std::max_element(range.begin(), range.end(), compareLogic);
      resultindices.push_back(minMaxIndex);
```

velox/functions/prestosql/ArrayMinMaxBy.cpp
```diff
@@ -0,0 +1,179 @@
...
+    BufferPtr resultIndices = allocateIndices(rows.size(), context.pool());
+    auto* rawResultIndices = resultIndices->asMutable<vector_size_t>();
+
+    BufferPtr resultNulls = allocateNulls(rows.size(), context.pool());
+    auto* mutableNulls = resultNulls->asMutable<uint64_t>();
+
+    auto indices = getMaxOrMinIndices(rows, *flatArray, *newElements, context);
+    rows.applyToSelected([&](auto row) {
+      auto offset = flatArray->offsetAt(row);
+      if (decodedElements->isNullAt(offset)) {
```
kevinwilfong, line 90: should this be decodedArray? I'm not sure why you're checking if the first element of the array is null 

and if that's the case, could you instead call
localResult->addNulls(decodedArray->nulls(), rows)

decodedArray->nulls() returns all the nulls after merging the encodings, and addNulls sets the nulls in localResult anywhere where rows is set and decodedArray->nulls() is null

velox/functions/prestosql/ArrayMinMaxBy.cpp
```diff
@@ -0,0 +1,179 @@
...
+        bits::setNull(mutableNulls, row);
+      }
+      rawResultIndices[row] = indices[row];
```
kevinwilfong, line 93: Why not just populate rawResultIndices inside getMaxOrMinIndices, then if my suggestion above applies, you can just get rid of this whole applyToSelected

velox/functions/prestosql/ArrayMinMaxBy.cpp
kevinwilfong: * You'll have to update the lambda to take it's arguments by value instead of by reference too, but they're vector_size_t so no performance concerns there

velox/functions/prestosql/ArrayMinMaxBy.cpp
HeidiHan0000: @kevinwilfong Thanks so much for the review! For this one, I was trying to cover the case where the input element in the array isn't null, but after the lambda is applied, it returns null. For instance an array of maps and the lambda returns the value at some key and the key does not exist in any of the maps, in which case I want the function to return null as per the Java implementation. But setting localResult nulls from decodedArray->nulls() doesn't seem to work for this case, how can I work with this for such cases?

HeidiHan0000: Changed to auto offset = indices[row]; since the above was looking at the offset of the first element. Fixed by looking at each max/min index instead

kevinwilfong (approved):
Thanks, this looks good to me

PR #12655, opened 2025-03-14
fix: Fix map_top_n_keys being unnecessarily orderable 

duxiao1212:
Summary:
map_top_n_keys returns the top N keys of the given map by sorting the keys in descending order according to the natural ordering of its keys. The values are unnecessarily orderable

partial function support https://fb.prod.workplace.com/groups/presto.users/permalink/28036948789260370/

Scalar function presto.default.map_top_n_keys not registered with arguments: (MAP<VARCHAR,MAP<VARCHAR,VARCHAR>>, BIGINT)

Differential Revision: D71176927


zacw7 (approved):
LGTM. Thanks for the fix. 

Please revise the title: fix ~~(velox)~~: Fix map_top_n_keys being unnecessarily orderable

PR #12657, opened 2025-03-14
refactor: Clean up some error messages and file structure

jagill:
Summary:
In addition to some text changes and extra invariant checks, BingTileType will soon
get a bunch more logic.  So I've extracted the non-inline `bingTileInvalidReason`
from BingTileType.h to BingTileType.cpp.  This is the only type in prestosql that
has a cpp file, so I'd like to check that's the right direction to go.

Differential Revision: D71222628

### DIFF
--- a/velox/functions/prestosql/GeospatialFunctions.h
+++ b/velox/functions/prestosql/GeospatialFunctions.h
@@ -30,6 +30,19 @@ struct BingTileFunction {
       const arg_type<int32_t>& x,
       const arg_type<int32_t>& y,
       const arg_type<int8_t>& zoom) {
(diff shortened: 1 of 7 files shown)


bikramSingh91 (approved):
> This is the only type in prestosql that has a cpp file, so I'd like to check that's the right direction to go.

Keeping the implementation in the header file allows the compiler to potentially inline those implementations that can potentially improve performance. The keyword here is "potentially" so it'll be hard to judge the perf impact, if any, for this move unless we write a micro-benchmark where most rows are error-ing out (and therefore taking the codepath that calls this static function).

If there is no strong reason to move it into a cpp file, we can just keep it in the header file.

PR #12662, opened 2025-03-15
feat: Add TopNRowNumberFuzzer runs to GitHub workflows

aditi-pandit:


aditi-pandit:
@kagamiori : This PR adds the TopNRowNumberFuzzer to CI. PTAL.

assignUser (approved):
The addition looks fine but with the growing number of fuzzers I am wondering if we should be a bit smarter about which to run when. @kgpai @mbasmanova @pedroerp 

We could run only a selection on PRs and do a full run nightly etc.?

kagamiori:
Hi @aditi-pandit, I looked at the saved log of the TopNRowNumberFuzzer but found that most iterations have the query failed in Presto. (You can find the log in topn-row-fuzzer-failure-artifacts at https://github.com/facebookincubator/velox/actions/runs/13869917052/.) You may want to take a look why that happens. (One possibility might be that --req_timeout_ms is too small.)

aditi-pandit:
Hi @kagamiori 

I see many error messages like below in velox_topn_row_number_fuzzer.INFO, but not in velox_topn_row_number_fuzzer.b769e0d338aa.root.log.INFO.20250314-230043.588. So not sure what to make of it.

In general I thought it failed since there are no data files in t_0. Is that not the case ?

Can you give me more info about the failure messages you are seeing ?


```
I20250314 23:00:52.469944   588 PrestoQueryRunner.cpp:734] Execute presto sql: DROP TABLE IF EXISTS t_0
I20250314 23:00:52.484997   588 PrestoQueryRunner.cpp:734] Execute presto sql: CREATE TABLE t_0(p0, p1, p2, p3, p4, s0, s1, s2, row_id) WITH (format = 'DWRF') AS SELECT cast(null as TINYINT), cast(null as REAL), cast(null as BIGINT), cast(null as BOOLEAN), cast(null as varchar), cast(null as BIGINT), cast(null as BIGINT), cast(null as DOUBLE), cast(null as INTEGER)
I20250314 23:00:52.592557   588 PrestoQueryRunner.cpp:734] Execute presto sql: SELECT "$path" FROM t_0
I20250314 23:00:52.720891   588 PrestoQueryRunner.cpp:734] Execute presto sql: DELETE FROM t_0
W20250314 23:00:52.762187   588 PrestoQueryRunner.cpp:716] Query failed in Presto
```

aditi-pandit:
Hi @kagamiori :

  After increasing the timeout I don't see the Exception that happened previously, but I'm still investigating the DELETE error.

```
E20250320 14:28:46.872663   587 Exceptions.h:66] Line: /__w/velox/velox/velox/velox/exec/fuzzer/PrestoQueryRunner.cpp:784, Function:fetchNext, Expression: response.status_code == 200 (0 vs. 200) GET from http://127.0.0.1:8080/v1/statement/queued/20250320_212845_00002_cweq6/2?slug=x40f4f32efa4744e0ae0cb457ee588b0c&binaryResults=true failed: Operation timed out after 1001 milliseconds with 0 bytes received, Source: RUNTIME, ErrorCode: INVALID_STATE
W20250320 14:28:46.873955   587 PrestoQueryRunner.cpp:713] Query failed in Presto
```

aditi-pandit:
Hi @kagamiori : The timeout errors are fixed now. Also, the DELETE errors go when setting hive.allow-drop-table in Hive config. I see the same error in other fuzzers as well.

Please can you take a look at this PR and let me know if anything else is needed from my side.

Thanks.

kagamiori (approved):

PR #12675, opened 2025-03-17
misc: Fix a typo in CMakeLists.txt

kou:
```diff
-qury
+query
```


kou:
If we use #12678, we don't need this. Because #12678 removes this comment.

kou:
We still need this even if we use #12678. Because #12678 doesn't change this eventually.

majetideepak (approved):

PR #12684, opened 2025-03-17
fix: Make type check reflect the corresponding logical type

PHILO-HE:
In Type.h, `is##NAME()` was defined to handily check data type based on `TypeKind` used in its
implementation. It may not reflect the logical type. For example, `isBigint()` returns true for
ShortDecimalType, IntervalDayTimeType and BigintType. It can potentially cause bugs if developers
don't notice this. In this pr, we make this type check function reflect the logical type instead of its
implementation type (when using this check function, developers generally don't care about its
implementation type). Thus, `isBigint()` will return true only for BigintType.

### DIFF
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -385,12 +385,15 @@ struct TypeTraits<TypeKind::OPAQUE> {
 template <TypeKind KIND>
 struct TypeFactory;
 
-#define VELOX_FLUENT_CAST(NAME, KIND)                                     \
+#define VELOX_TYPE_AS(NAME, KIND)                                         \
   const typename TypeTraits<TypeKind::KIND>::ImplType& as##NAME() const { \
     return this->as<TypeKind::KIND>();                                    \
-  }                                                                       \
-  virtual bool is##NAME() const {                                         \
-    return this->kind() == TypeKind::KIND;                                \
+  }
+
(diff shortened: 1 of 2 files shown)


Yuhta (commented):
This is a dangerous change both in terms of correctness and performance.  Can we add new methods instead of changing the behavior of widely used methods?

PHILO-HE:
> This is a dangerous change both in terms of correctness and performance. Can we add new methods instead of changing the behavior of widely used methods?

@Yuhta, thanks for your comment! I found in existing Velox code, the logical type is always expected when this type check function is called. And if we don't fix this, it can easily cause issue I think. For example, in the below code, the order of `isLongDecimal` and `isHugeint` matters since `isHugeint` returns true also for LongDecimal type. I prefer directly changing this function if we expect it is checking logical types instead of the underlying implementation type. 

```
VextorFuzzer.cpp
...
    } else if constexpr (std::is_same_v<TCpp, int128_t>) {
      if (vector->type()->isLongDecimal()) {
        flatVector->set(i, randLongDecimal(vector->type(), rng));
      } else if (vector->type()->isHugeint()) {
        flatVector->set(i, rand<int128_t>(rng, opts.dataSpec));
      } else {
        VELOX_NYI();
      }
```

PR #12685, opened 2025-03-17
fix: Use memory pool to track raw_vector in selective column readers

Yuhta:
Differential Revision: D71318586

### DIFF
--- a/velox/common/memory/MemoryAllocator.cpp
+++ b/velox/common/memory/MemoryAllocator.cpp
@@ -135,10 +135,13 @@ MemoryAllocator::SizeMix MemoryAllocator::allocationSize(
 bool MemoryAllocator::isAlignmentValid(
     uint64_t allocateBytes,
     uint16_t alignmentBytes) {
+  // Check if the allocateBytes is conforming to the alignmentBytes.
+  // alignmentBytes must be a power of two, so we can replace the expensive
+  // modulo operation with bitwise and.
   return (alignmentBytes == kMinAlignment) ||
       (alignmentBytes >= kMinAlignment && alignmentBytes <= kMaxAlignment &&
-       allocateBytes % alignmentBytes == 0 &&
-       (alignmentBytes & (alignmentBytes - 1)) == 0);
+       bits::isPowerOfTwo(alignmentBytes) &&
+       (allocateBytes & (alignmentBytes - 1)) == 0);
 }
 
 void MemoryAllocator::alignmentCheck(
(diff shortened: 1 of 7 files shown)


velox/common/memory/MemoryAllocator.cpp
```diff
@@ -137,8 +137,8 @@ bool MemoryAllocator::isAlignmentValid(
     uint16_t alignmentBytes) {
   return (alignmentBytes == kMinAlignment) ||
       (alignmentBytes >= kMinAlignment && alignmentBytes <= kMaxAlignment &&
-       allocateBytes % alignmentBytes == 0 &&
-       (alignmentBytes & (alignmentBytes - 1)) == 0);
+       isPowerOfTwo(alignmentBytes) &&
+       (allocateBytes & (alignmentBytes - 1)) == 0);
```
tanjialiang, line 141: Is this change equivalent? allocateBytes must be equal to alignmentBytes?

velox/common/memory/MemoryAllocator.h
```diff
@@ -347,6 +347,10 @@ class MemoryAllocator : public std::enable_shared_from_this<MemoryAllocator> {
 
   virtual std::string toString() const = 0;
 
+  static bool isPowerOfTwo(uint16_t size) {
```
tanjialiang, line 350: There is a helper in BitUtil.h

velox/common/memory/MemoryPool.cpp
```diff
@@ -491,7 +491,9 @@ void* MemoryPoolImpl::allocate(
     std::optional<uint32_t> alignment) {
   if (alignment.has_value()) {
     const auto alignmentValue = alignment.value();
-    if (FOLLY_UNLIKELY(alignment_ % alignmentValue != 0)) {
+    if (FOLLY_UNLIKELY(
+            !(MemoryAllocator::isPowerOfTwo(alignmentValue) &&
+              alignmentValue <= alignment_))) {
```
tanjialiang, line 496: ditto BitUtil.h

And I think you mean to put || condition?

Yuhta: It's &&, alignment is only allowed to be power of two, and the alignment of allocation must be guaranteed by alignment of the pool

velox/common/memory/MemoryAllocator.h
Yuhta: That one looks quite terrible...Let me see if I can fix that one

velox/common/memory/MemoryAllocator.cpp
tanjialiang: You are right, the check matches

velox/common/memory/MemoryPool.cpp
tanjialiang: Right, I missed a bracket, thinking the ! is applied to only the first condition.

tanjialiang (approved):

tanjialiang:
Thanks @Yuhta . Could you help to add a note document to MemoryPool::alignment_ saying that it is enforced to be pow2?

PR #12686, opened 2025-03-17
fix(json): Compute capacity correctly when casting to complex jsons

kgpai:
Summary: The current JsonCastOperator will incorrectly compute capacity of its json buffer because finalize() gets called twice - once in line 705 and the other when writer->commit() is called. This causes it throw eventually when it gets close to capacity. We fix this by just removing the unnecessary call to finalize in line 705.

Differential Revision: D71322529

### DIFF
--- a/velox/functions/prestosql/tests/JsonCastTest.cpp
+++ b/velox/functions/prestosql/tests/JsonCastTest.cpp
@@ -1542,6 +1542,7 @@ TEST_F(JsonCastTest, castFromJsonWithEscaping) {
   });
   auto resultMap = evaluate(
       "cast(json_parse(c0) as map(varchar, json))", makeRowVector({svData}));
+
   test::assertEqualVectors(expected, resultMap);
 
   // Test cast from Json to ARRAY(JSON) gets escaped correctly.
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/types/JsonCastOperator.cpp
```diff
@@ -702,7 +702,6 @@ struct CastFromJsonTypedImpl {
           auto size = unescapeSizeForJsonCast(json.data(), json.size());
           vectorWriter.resize(size);
           unescapeForJsonCast(json.data(), json.size(), vectorWriter.data());
-          vectorWriter.finalize();
```
kevinwilfong, line 705: Where does commit get called? (In particular will this work for scalar strings not wrapped inside a complex value too?)

velox/functions/prestosql/tests/JsonCastTest.cpp
```diff
@@ -1581,4 +1582,24 @@ TEST_F(JsonCastTest, castFromJsonWithEscaping) {
   auto resultVarchar =
       evaluate("cast(json_parse(c0) as varchar)", makeRowVector({svData}));
   test::assertEqualVectors(expectedVarchar, resultVarchar);
+
+  // Create a large vector to ensure vectors string buffer has its capacity
+  // computed correctly.
+  {
+    auto largeVector = makeFlatVector<StringView>(1000, [](auto row) {
```
kevinwilfong, line 1589: nit: row isn't used (lint warning)

velox/functions/prestosql/tests/JsonCastTest.cpp
```diff
@@ -1581,4 +1582,24 @@ TEST_F(JsonCastTest, castFromJsonWithEscaping) {
...
+      return R"({"someKey": "some large enough string 😀"})"_sv;
+    });
+    auto resultLarge = evaluate(
+        "cast(json_parse(c0) as map(varchar, json))",
+        makeRowVector({largeVector}));
+    auto resultMap = resultLarge->as<MapVector>();
```
kevinwilfong, line 1595: nit: this shadows another variable resultMap defined in the test (lint warning)

kevinwilfong (approved):

velox/functions/prestosql/tests/JsonCastTest.cpp
kgpai: Yes but I like this name - i will remove the auto..

PR #12689, opened 2025-03-17
misc(fuzzer): Log input test cases in expression fuzzer

kagamiori:
Summary:
This diff makes the expression fuzzer log values in the input vectors when `-v=1` is 
specified. This diff also makes PrestoQueryRunner function signatures that has JSON 
argument type because JSON cannot be written directly to HIVE.

Differential Revision: D71333637


natashasehgal (approved):
LGTM

Issue #12558, opened 2025-03-05
escape: Velox throws while Presto outputs


peterenescu:
### Description

Presto and Velox behave different for `escape`.

### Error Reproduction

Velox
```
presto:di> SELECT c0 like c1 escape c2 from (values (varchar 'N#w&{~>|.v^''%Ed{H^v6/\k::<,pf5-p?@g;1MsE$!Y0&9[hks,eHvrTv>H', varchar 'oEP}kS-lLtMzK_9-|Q&p0zA17[]6-"ZeKB&d(3B!I0Wp7Xu%\0xg&dH', varchar '') ) t(c0, c1, c2);
Query 20250305_225027_43941_nqej6 failed: escapeChar.size() == 1 (0 vs. 1) Escape string must be a single character Top-level Expression: presto.default.like(field, field_0, field_1)
```
Presto
```
presto> SELECT c0 like c1 escape c2 from (values (varchar 'N#w&{~>|.v^''%Ed{H^v6/\k::<,pf5-p?@g;1MsE$!Y0&9[hks,eHvrTv>H', varchar 'oEP}kS-lLtMzK_9-|Q&p0zA17[]6-"ZeKB&d(3B!I0Wp7Xu%\0xg&dH', varchar '') ) t(c0, c1, c2);
 _col0 
-------
 false 
```

### Relevant logs

```bash

```

peterenescu:
Might be solved by https://github.com/facebookincubator/velox/pull/12592/files

kagamiori:
Related: https://github.com/prestodb/presto/issues/24747

PR #12691, opened 2025-03-17
fix(escape): Update escape to accept empty string as a valid character

peterenescu:
Summary:
Update behavior of function `escape` to accept empty string as a valid input character. This change should increase function parity with Presto identified by: https://github.com/facebookincubator/velox/issues/12558

Currently Presto's behavior is identified by the following:
```
presto> SELECT c0 like c1 escape c2 from (values (varchar 'N#w&{~>|.v^''%Ed{H^v6/\k::<,pf5-p?g;1MsE$!Y0&9[hks,eHvrTv>H', varchar 'oEP}kS-lLtMzK_9-|Q&p0zA17[]6-"ZeKB&d(3B!I0Wp7Xu%\0xg&dH', varchar '') ) t(c0, c1, c2);
 _col0 
-------
 false 
```
Velox, on the other hand:
```
presto:di> SELECT c0 like c1 escape c2 from (values (varchar 'N#w&{~>|.v^''%Ed{H^v6/\k::<,pf5-p?g;1MsE$!Y0&9[hks,eHvrTv>H', varchar 'oEP}kS-lLtMzK_9-|Q&p0zA17[]6-"ZeKB&d(3B!I0Wp7Xu%\0xg&dH', varchar '') ) t(c0, c1, c2);
Query 20250305_225027_43941_nqej6 failed: escapeChar.size() == 1 (0 vs. 1) Escape string must be a single character Top-level Expression: presto.default.like(field, field_0, field_1)
```

Differential Revision: D71139264

### DIFF
--- a/velox/functions/lib/Re2Functions.cpp
+++ b/velox/functions/lib/Re2Functions.cpp
@@ -1015,7 +1015,7 @@ class LikeGeneric final : public exec::VectorFunction {
       context.applyToSelectedNoThrow(rows, [&](auto row) {
         vectorWriter.setOffset(row);
         auto escapeChar = escapeReader[row];
-        VELOX_USER_CHECK_EQ(
+        VELOX_USER_CHECK_LE(
             escapeChar.size(), 1, "Escape string must be a single character");
         vectorWriter.current() = applyRow(
             inputReader[row], patternReader[row], escapeChar.data()[0]);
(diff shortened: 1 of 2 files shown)


kagamiori:
Presto is having inconsistent behaviors with the LIKE operator with empty escape string: https://github.com/prestodb/presto/issues/24747. Let's allow Presto to clarify the correct behavior before changing Velox, since it's not clear to me that Velox behavior is wrong.

PR #12694, opened 2025-03-17
fix: Expand the range of supported years

kevinwilfong:
Summary:
Presto Java supports any timestamp that can fit in the 52 bits well allocate for millis UTC in
the TimestampWithTimeZone type.  Likewise the Timestamp type can support dates with
years past 200 million.

In Velox currently we limit the year to what can be represented with 16 bits (this is what
std::chrono applies in C++20).

This change adds support for years up to +/-2^22. This is well beyond what is needed to
support the full range of timestamps Presto Java supports in TimestampWithTimeZone.
This falls short of what Presto Java supports for the Timestamp type (this only supports
years a little past 4 million) but the algorithms we use to convert between year_month_day
and days since epoch fall apart beyond that point.

Differential Revision: D71349259

### DIFF
--- a/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
+++ b/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
@@ -337,12 +337,10 @@ TEST_F(DateTimeFunctionsTest, unixTimestampDateInput) {
   EXPECT_EQ(1727766000, unixTimestamp(parseDate("2024-10-01")));
   EXPECT_EQ(-126065866022, unixTimestamp(parseDate("-2025-02-18")));
   EXPECT_EQ(2398320000, unixTimestamp(parseDate("2045-12-31")));
-
-  // Test invalid inputs.
-  VELOX_ASSERT_THROW(
-      unixTimestamp(kMax), "Timepoint is outside of supported year range");
-  VELOX_ASSERT_THROW(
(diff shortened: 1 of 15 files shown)


yuandagits (approved):

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -337,12 +337,10 @@ TEST_F(DateTimeFunctionsTest, unixTimestampDateInput) {
...
-
-  // Test invalid inputs.
-  VELOX_ASSERT_THROW(
-      unixTimestamp(kMax), "Timepoint is outside of supported year range");
-  VELOX_ASSERT_THROW(
-      unixTimestamp(kMin), "Timepoint is outside of supported year range");
+  EXPECT_EQ(
+      185542587126000, unixTimestamp(std::numeric_limits<int32_t>::max()));
+  EXPECT_EQ(
+      -185542587158822, unixTimestamp(std::numeric_limits<int32_t>::min()));
```
zhouyuan, line 343: this changed the behavior on the invalid timestmp handling in SparkSQL. Could we add a config to control this in Spark function?
Cc: @rui-mo

rui-mo: Fixing the unix_timestamp in https://github.com/facebookincubator/velox/pull/12881.

PR #12701, opened 2025-03-18
feat: Add API DataSource::cancel() to allow implementation run cleanup logic while TableScan::close() is called

zhztheplayer:
This is another simpler approach to allow downstream project deal with the memory leak issue mentioned in https://github.com/facebookincubator/velox/pull/12609.

See also [comment](https://github.com/facebookincubator/velox/pull/12609#issuecomment-2726568760).

### DIFF
--- a/velox/connectors/Connector.h
+++ b/velox/connectors/Connector.h
@@ -297,6 +297,13 @@ class DataSource {
   virtual std::shared_ptr<wave::WaveDataSource> toWaveDataSource() {
     VELOX_UNSUPPORTED();
   }
+
+
+  /// Invoked by table scan close to cancel any inflight async operations
+  /// running inside the data source. This is the best effort and the actual
+  /// connector implementation decides how to support the cancellation if
+  /// needed.
+  virtual void cancel() {}
 };
 
 class IndexSource {
--- a/velox/connectors/Connector.h
+++ b/velox/connectors/Connector.h
@@ -298,7 +298,6 @@ class DataSource {
     VELOX_UNSUPPORTED();
   }
 
-
   /// Invoked by table scan close to cancel any inflight async operations
   /// running inside the data source. This is the best effort and the actual
   /// connector implementation decides how to support the cancellation if
--- a/velox/exec/TableScan.cpp
+++ b/velox/exec/TableScan.cpp
@@ -436,6 +436,10 @@ void TableScan::addDynamicFilter(
 void TableScan::close() {
   Operator::close();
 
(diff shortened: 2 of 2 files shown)


zhztheplayer:
@xiaoxmeng Would you like to take a look? Thanks.

BTW the CI failures don't seem to be related.

velox/exec/TableScan.cpp
```diff
@@ -436,6 +436,10 @@ void TableScan::addDynamicFilter(
 void TableScan::close() {
   Operator::close();
 
+  if (dataSource_ != nullptr) {
+    dataSource_->close();
```
Yuhta, line 440: There are also data sources in splits, it's not convenient to close it explicitly, can we just add the cleanup logic in destructor?
https://github.com/facebookincubator/velox/blob/756f104e27d25c3ac29f7f392516f8231f422cff/velox/connectors/Connector.h#L65

zhztheplayer: > can we just add the cleanup logic in destructor?

https://github.com/facebookincubator/velox/pull/12609 is a fix when destructor is relied for the purpose. While the reason it's not that applicable is explained [here](https://github.com/facebookincubator/velox/pull/12609#pullrequestreview-2687358329).

> There are also data sources in splits, it's not convenient to close it explicitly

There is no code in `DataSource::close` yet by default. Do you mean we'd take care of async close as well in the default implementation?

xiaoxmeng: @Yuhta do you mean async split source? Each data source is assumed to have one split?

xiaoxmeng: Chat with @zhztheplayer offline, can we provide a cancel API to data source for best effort cancellation support and default is do nothing. @zhztheplayer use this in a customized connector for a streaming service.

velox/connectors/Connector.h
```diff
@@ -297,6 +297,13 @@ class DataSource {
   virtual std::shared_ptr<wave::WaveDataSource> toWaveDataSource() {
     VELOX_UNSUPPORTED();
   }
+
+  /// Cancel the ongoing works in the data source. The function will be
```
xiaoxmeng, line 301:
```
  /// Invoked by table scan close to cancel any inflight async operations running inside the data source. This is a best effort and the actual connector implementation decides how to support the cancellation if needs.
```

xiaoxmeng (commented):
@zhztheplayer LGTM % nit.

PR #12706, opened 2025-03-18
feat(fuzzer): Add comman eval execution logging

peterenescu:
Summary: Minor update to ExpressionVerifier to add logging during the common eval try/catch block. I've noticed on my local runs, it really helps to quickly identify whether the value/behavior mismatch occurred because Velox failed and was caught or outputted successfully and failed on Presto.

Differential Revision: D71404365


kagamiori (approved):
LGTM. Thanks!

PR #12711, opened 2025-03-18
misc(backout): Combine low selectivity vectors generated by the hash join filter"

yuandagits:
Summary:
I strongly believe a bug was introduced in D68024717 (where, I am not sure yet). Some where in that diff, I believe the filter may be messing/corrupting a buffer for certain vector encodings.

The symptom is that we get a crash in [BitUtils](https://github.com/facebookincubator/velox/blob/dfadae6dc23e9f2678eb2230277f921dc75dad20/velox/vector/DecodedVector.h#L449); however, I suspect that this logic is okay. What seems to be incorrect is higher in the stack. It's possible that the BufferPtr in the FlatVector is incorrect.

Observation, the Dictionary encoded vector is "flattened" and the value ptr is given to the decodedVector. The decodedVector receives a pointer to data_ (0x00007f5fa4efe840, 140048765806656) of capacity=6048 and size=5995.

When we call decodedVector.valueAt<bool>(idx=450), index_ returns  value of 48641, this is used to access the bits ptr array, idx / (sizeof(bits[0]) *  8 ), with a value of 760. bits[760] crashes. Which is why I think the BufferPtr is incorrect somehow.

See P1759606983

Differential Revision: D71432459


xiaoxmeng (approved):
@zhli1142015 fyi, this might break meta internal production again.

zhli1142015:
Thanks for the heads-up! Would it be possible to provide a repro so I can investigate?

yuandagits:
> Thanks for the heads-up! Would it be possible to provide a repro so I can investigate?

Example query.
```
SELECT
    foo_bar_id,
    "map_union_sum"(interactions) interactions
FROM (
        (
            SELECT
                foo_bar_id,
                interactions,
                log_ds
            FROM (
                SELECT
                    *
                FROM "some_table_1:foobar"
                WHERE
                    (
                        (ds = '2025-03-16')
                        AND (version = 'v4')
                    )
            )
        )
        RIGHT JOIN (
            SELECT
                *
            FROM some_table_2
            WHERE
                (ds = '2025-03-16')
        ) USING (foo_bar_id)
    )
WHERE
    (
        "date_diff"(
            'day',
            "date"(log_ds),
            COALESCE(some_time, "date"('2025-03-16'))
        ) BETWEEN 0 AND 5
    )
GROUP BY
    1;
  ```
  
  There is some issue in HashProbe during the filtering

abhguptg:
Hi @yuandagits, @xiaoxmeng or @Yuhta, I was unable to repro with the above query since it seems to be on internal data. Is it possible to provide the conditions of the repro (data size, queries, etc.) or an open source repro to investigate this further?

Additionally, was this on Prestissimo or Gluten?

abhguptg:
@yuandagits, @xiaoxmeng or @Yuhta, just bumping up the above comment: I was unable to repro with the above query since it seems to be on internal data. Is it possible to provide the conditions of the repro (data size, queries, etc.) or an open source repro to investigate this further?

Additionally, was this on Prestissimo or Gluten?

Issue #12715, opened 2025-03-19
Throw invalid time zone instead of runtime error when timezone not found


kevinw66:
### Description

When we build timezone database in `TimeZoneMap.cpp`, we use the following code which try to catch invalid_time_zone error when timezone not found.
```cpp
      try {
        zone = locateZoneImpl(entry.second);
      } catch (const tzdb::invalid_time_zone& err) {
        // When this exception is thrown, it typically means the time zone name
        // we are trying to locate cannot be found from OS's time zone database.
        // Thus, we just command a "continue;" to skip the creation of the
        // corresponding TimeZone object.
        //
        // Then, once this time zone is requested at runtime, a runtime error
        // will be thrown and caller is expected to handle that error in code.
        LOG(WARNING) << "Invalid time zone [" << entry.second
                     << "]: " << err.what();
        continue;
      }
```

But we throws [Runtime Error](https://github.com/facebookincubator/velox/blob/2f48b5772875f06f1fa8d47f2f298eea0c6cb29f/velox/external/tzdb/tzdb.h#L54) after change timezone to llvm implementation, which cannot be catch by this and cause lots of failure when we use early version of `tzdata`.
I think maybe we should keep throw invalid_time_zone like before.

PR #12716, opened 2025-03-19
misc: Throw invalid time zone instead of runtime error when timezone not found

kevinw66:
fixes: #12715

### DIFF
--- a/velox/external/tzdb/tzdb.h
+++ b/velox/external/tzdb/tzdb.h
@@ -51,7 +51,7 @@ struct tzdb {
     if (const time_zone* __result = __locate_zone(__name))
       return __result;
 
-    std::__throw_runtime_error("tzdb: requested time zone not found");
+    __throw_invalid_time_zone(std::string(__name));
   }
 
   [[nodiscard]] const time_zone* current_zone() const {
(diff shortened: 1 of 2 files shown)


kevinw66:
@kevinwilfong Could you please help review this? thanks!

kevinwilfong:
Thanks for the fix! Could you add a unit test?

kevinw66:
> Thanks for the fix! Could you add a unit test?

Thanks @kevinwilfong , I'll add one in a few days!

kevinwilfong (approved):
Thanks again!

kevinw66:
@kevinwilfong Done, thanks!

zhouyuan:
Hi @kevinw66 @kevinwilfong  
Thanks for the fix. I also run into similar issue with Gluten running on Centos7(It's EOL but still commonly used in some production env) and I made another patch here(https://github.com/facebookincubator/velox/pull/12722). To fix the zone missing issue, we may also need to fix below line to guard on reading the non-exits file otherwise it will report `basic_ios::clear`
https://github.com/facebookincubator/velox/blob/main/velox/external/tzdb/tzdb.cpp#L711


Thanks, -yuan

PR #12728, opened 2025-03-20
fix: Print CMAKE_CXX_FLAGS at the final stage in CMakeLists.txt

wraymo:
Summary
This PR moves the `message("FINAL CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}")` call to occur after the final flag is added.


assignUser (approved):
Makes sense, though we should really provide a better overview of the build status... I'll open an Issue.

PR #12729, opened 2025-03-20
fix: Fix the async direct coalesce load memory leak in case of early task failure in pyspark

xiaoxmeng:
Summary:
Here is the race condition can cause memory leak in race between async direct coalesce load and early task failure:
T1. file reader triggers stripe load which does coalesce prefetch
T2. coalesce prefetch kicks off and pass the cancellation check
T3. coalesce load does the memory allocations and do read from storage
T4. before table scan do on-demand load, the task fails or prefetch data is skipped by filtering and task finishes.
T5. task destruction frees the memory pool hits the memory leak check failure.
T6. if we disable the memory leak check failure (with memory leak metric reporting in Meta production), then
the buffer free will throw with bad memory pool pointer.

Verified the fix with unit test that reproduce the race condition. This is exposed by pyspark use case.

Differential Revision: D71529929

### DIFF
--- a/velox/dwio/common/DirectBufferedInput.cpp
+++ b/velox/dwio/common/DirectBufferedInput.cpp
@@ -17,11 +17,13 @@
 #include "velox/dwio/common/DirectBufferedInput.h"
 #include "velox/common/memory/Allocation.h"
 #include "velox/common/process/TraceContext.h"
+#include "velox/common/testutil/TestValue.h"
 #include "velox/dwio/common/DirectInputStream.h"
 
 DECLARE_int32(cache_prefetch_min_pct);
 
 using ::facebook::velox::common::Region;
+using facebook::velox::common::testutil::TestValue;
 
 namespace facebook::velox::dwio::common {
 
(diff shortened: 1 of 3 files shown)


tanjialiang (approved):
Thanks for the quick fix

fzhedu:
@xiaoxmeng 
why not let the main thread wait all IO threads to be finished? one possible case happen that the main thread ~memoryManager, but the scan pool is still held by the IO threads, causing core dump.
```
  ~DirectBufferedInput() override {
    std::vector<folly::Future<bool>> futures;
    for (auto& load : coalescedLoads_) {
      load->cancel();
      futures.push_back(load->getCompletionFuture());
    }
    if (!futures.empty()) {
      LOG(WARNING) << "wait " << futures.size() << " loads to be completed";
      folly::collectAll(futures).wait();
    }
```

j7nhai:
> @xiaoxmeng why not let the main thread wait all IO threads to be finished? one possible case happen that the main thread ~memoryManager, but the scan pool is still held by the IO threads, causing core dump.
> 
> ```
>   ~DirectBufferedInput() override {
>     std::vector<folly::Future<bool>> futures;
>     for (auto& load : coalescedLoads_) {
>       load->cancel();
>       futures.push_back(load->getCompletionFuture());
>     }
>     if (!futures.empty()) {
>       LOG(WARNING) << "wait " << futures.size() << " loads to be completed";
>       folly::collectAll(futures).wait();
>     }
> ```

Actually, the core dump is caused by memory leak. refer to my PR fix: https://github.com/facebookincubator/velox/pull/14722

PR #12736, opened 2025-03-20
refactor: Remove unused bufferEnd_ field from parquet decoders

Yuhta:
Summary: We keep it in the constructor just in case we need it someday.

Differential Revision: D71554837


pedroerp (approved):
Thank you

PR #12740, opened 2025-03-20
feat: Adding reader tests for orc files

kevins963:
Summary:
Test Changes in `velox/dwio/orc/test/ReaderTest` other files are resource files

Adding ORC format reading tests using example files provided by the apache orc project.  These are predefined orc files and known stripe footer checks and reading outputs.

This validates the ability to read all existing ORC file formats, schemas and types.

Before the ORC reader functionality most of the orc formated files would fail to read with VeloxExceptions being thrown

Differential Revision: D71527920


Yuhta (approved):

majetideepak (changes requested):
We should try to avoid checking in test files.
Can we download these files when needed instead?
We have a similar repo for parquet files https://github.com/apache/parquet-testing

majetideepak:
These files seem to be coming from here https://github.com/apache/orc/tree/main/examples
Let's download this just like Arrow does.

kevins963:
> We should try to avoid checking in test files. Can we download these files when needed instead? We have a similar repo for parquet files https://github.com/apache/parquet-testing

@majetideepak removed the files for testing all combinations of compressions, file info, column types, etc...

For now only including two files to validate the majority of the column types and orc file/footer/stripe_footer info.  If we do want to validate all the combinations we can find out an alternative way such as download test files as needed.

Ideally once we have Writer support of ORC we can remove all the testing files and just use the Writer/Reader for E2E testing without the need for test files.

majetideepak:
@kevins963 Thanks! Did you miss pushing a commit? I still see files in this PR.

velox/dwio/orc/test/ReaderTest.cpp
```diff
@@ -277,3 +287,2869 @@ TEST_F(OrcReaderTest, testOrcRlev2) {
...
+      std::ostream& stream,
+      OrcFileDescription const& obj);
+};
+
+std::ostream& operator<<(std::ostream& stream, OrcFileDescription const& obj) {
+  stream << obj.filename;
+  return stream;
+}
+
+std::map<std::string, std::string> makeMetadata();
```
majetideepak, line 341: @kevins963 Is this used anywhere? The implementation is a huge > 2000 line blob.

majetideepak (approved):

PR #12746, opened 2025-03-20
test: Add IntermediateTypeTransforms for supporting intermediate types in PrestoQueryRunner

kevinwilfong:
Summary:
Presto supports a number of what I call "intermediate types" that are supported in the query
engine but not in the file formats.  This means we can't currently use fuzzers with the 
PrestoQueryRunner with these types because there's no way to ensure we don't generate
plans with these types as inputs (which have to be written to a DWRF file), and if we did
exclude them, that would prevent us from e.g. testing them in keys/aggregates in the 
AggregationFuzzer which doesn't do any projections.

To solve this, I add IntermediateTypeTransforms which provides functions transform and 
projectionExpr. Transform converts values of an "intermediate type" to values of a type that
is supported in the file format. projectionExpr generates an expression tree that can convert
the types output by transform back into the "intermediate type".

In a follow up diff I integrate this into the fuzzers that support PrestoQueryRunner so that
they can begin fuzzing "intermediate types".

In this diff I provide implementations for Array, Map, and Row in cases where they contain an
"intermediate type", as well as an implementation for TimestampWithTimeZone as the
inaugural "intermediate type".

Differential Revision: D71585451

### DIFF
--- a/velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
+++ b/velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
@@ -1,0 +1,444 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.h"
+#include "velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.h"
+#include "velox/expression/VectorWriters.h"
+#include "velox/functions/prestosql/types/TimestampWithTimeZoneType.h"
+#include "velox/parse/Expressions.h"
+#include "velox/vector/ComplexVector.h"
+#include "velox/vector/DecodedVector.h"
+#include "velox/vector/fuzzer/Utils.h"
+
+namespace facebook::velox::exec::test {
+namespace {
+class ArrayTransform {
+ public:
+  VectorPtr transform(const VectorPtr& vector, const SelectivityVector& rows)
+      const;
+
+  core::ExprPtr projectionExpr(
+      const TypePtr& type,
+      const core::ExprPtr& inputExpr,
+      const std::string& columnAlias) const;
+};
+
+class MapTransform {
+ public:
+  VectorPtr transform(const VectorPtr& vector, const SelectivityVector& rows)
+      const;
+
+  core::ExprPtr projectionExpr(
+      const TypePtr& type,
+      const core::ExprPtr& inputExpr,
+      const std::string& columnAlias) const;
+};
+
+class RowTransform {
+ public:
+  VectorPtr transform(const VectorPtr& vector, const SelectivityVector& rows)
+      const;
+
+  core::ExprPtr projectionExpr(
+      const TypePtr& type,
+      const core::ExprPtr& inputExpr,
+      const std::string& columnAlias) const;
+};
+
+const ArrayTransform kArrayTransform;
+const MapTransform kMapTransform;
+const RowTransform kRowTransform;
+
+const std::unordered_map<TypePtr, std::shared_ptr<IntermediateTypeTransform>>&
+intermediateTypeTransforms() {
+  static std::unordered_map<TypePtr, std::shared_ptr<IntermediateTypeTransform>>
+      intermediateTypeTransforms{
+          {TIMESTAMP_WITH_TIME_ZONE(),
+           std::make_shared<TimestampWithTimeZoneTransform>()}};
+  return intermediateTypeTransforms;
+}
+
+const std::shared_ptr<IntermediateTypeTransform>& getIntermediateTypeTransform(
+    const TypePtr& type) {
+  auto it = intermediateTypeTransforms().find(type);
+
+  if (it == intermediateTypeTransforms().end()) {
+    VELOX_FAIL(
+        "Attempting to transform unsupported intermediate only type: {}",
+        type->toString());
+  }
+
+  return it->second;
+}
+
+VectorPtr transformIntermediateOnlyType(
+    const VectorPtr& vector,
+    const SelectivityVector& rows) {
+  const auto& type = vector->type();
+  if (type->isArray()) {
+    return kArrayTransform.transform(vector, rows);
+  } else if (type->isMap()) {
+    return kMapTransform.transform(vector, rows);
+  } else if (type->isRow()) {
+    return kRowTransform.transform(vector, rows);
+  } else {
+    const auto& transform = getIntermediateTypeTransform(type);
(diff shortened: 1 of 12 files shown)


velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -0,0 +1,381 @@
...
+          std::vector<core::ExprPtr>{inputExpr}));
+    }
+  }
+
+  // We deliberately make a copy here so we can move it.
+  std::vector<TypePtr> childTypes = rowType.children();
+  std::vector<std::string> childNames;
+  for (int i = 1; i <= childTypes.size(); i++) {
+    childNames.push_back(fmt::format("c{}", i));
+  }
```
kagamiori, line 330: If the rowType already has child names, shall we use that?

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -0,0 +1,381 @@
...
+  return std::make_shared<core::CallExpr>(
+      "switch",
+      std::vector<core::ExprPtr>{
+          std::make_shared<core::CallExpr>(
+              "is_null", std::vector<core::ExprPtr>{inputExpr}, std::nullopt),
```
kagamiori, line 335: I think I don't fully understand this part. When would is_null(inputExpr) happen?

velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.cpp
```diff
@@ -0,0 +1,125 @@
...
+  str.resize(maxResultSize);
+  const auto resultSize =
+      jodaDateTime_->format(timestamp, timezonePtr, maxResultSize, str.data());
+  str.resize(resultSize);
+
+  return str;
+}
+
+// Convert a Vector of TimestampWithTimeZone to a RowVector of BIGINT and
+// VARCHAR (millisUtc and time_zone).
```
kagamiori, line 41: I don't find the piece of code that create a Bigint vector. Could you help point which line of code does this?

velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+  if (!decoded.isIdentityMapping()) {
+    result = decoded.wrap(result, *vector, vector->size());
+  }
+
+  return result;
+}
+
+// Applies from_unixtime to a RowVector of BIGINT and VARCHAR (millisUtc and
+// time_zone) to produce values of type TimestampWithTimeZone.
```
kagamiori, line 87: Does this comment needs to be updated since the expression actually uses parse_datetime?

kagamiori (commented):
Hi @kevinwilfong, thank you for adding the support for intermediate types! I like the recursive construction of vectors and expressions according to data types. I left some comments.

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -0,0 +1,381 @@
...
+        "transform_values",
+        std::vector<core::ExprPtr>{
+            expr,
+            std::make_shared<core::LambdaExpr>(
+                std::vector<std::string>{"k", "v"},
+                getIntermediateOnlyTypeProjectionExpr(
+                    valuesType,
+                    std::make_shared<core::FieldAccessExpr>("v", "v"),
+                    "v"))},
+        columnAlias);
```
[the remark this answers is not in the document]
kevinwilfong: Yes, in that case the input to transform_values is the transform_keys expression (since we overwrite expr)

velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.cpp
kevinwilfong: My bad, I rewrote the code to produce just a VARCHAR with the formatted datetime (my original approach that this describes could not capture timestamps with millisecond precision), let me fix the comment.

velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.cpp
kevinwilfong: Same here, thanks for catching it.

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
kevinwilfong: If the value of this RowType expression is null in a particular row. Since we call this expression on every row in a batch some of the values may be null so we need to handle that case.

velox/exec/fuzzer/PrestoQueryRunnerTimestampWithTimeZoneTransform.cpp
```diff
@@ -0,0 +1,126 @@
...
+// Applies parse_datetime to a Vector of VARCHAR (formatted timestamps with time
+// zone) to produce values of type TimestampWithTimeZone.
+core::ExprPtr TimestampWithTimeZoneTransform::projectionExpr(
+    const TypePtr& type,
+    const core::ExprPtr& inputExpr,
+    const std::string& columnAlias) const {
+  VELOX_CHECK(isTimestampWithTimeZoneType(type));
+
+  // format_datetime with the ZZZ pattern produces time zones that need to be
+  // parsed with either the ZZZ or ZZ pattern, to handle this we try parsing
```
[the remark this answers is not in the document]
kevinwilfong: I just refer to the Presto documentation (since that's what we should be emulating) and it links to the documentation here https://joda-time.sourceforge.net/apidocs/org/joda/time/format/DateTimeFormat.html

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+  return std::make_shared<core::CallExpr>(
+      "switch",
+      std::vector<core::ExprPtr>{
+          std::make_shared<core::CallExpr>(
+              "is_null", std::vector<core::ExprPtr>{inputExpr}, std::nullopt),
+          std::make_shared<core::ConstantExpr>(
+              type, variant::null(TypeKind::ROW), std::nullopt),
+          std::make_shared<core::CastExpr>(
+              type,
```
kagamiori, line 400: So this `type` argument is the RowType that contains intermediate types, while `inputExpr` is of a RowType that contains input types? Could we add comments in the header file to document these arguments?

velox/vector/fuzzer/ConstrainedVectorGenerator.cpp
```diff
@@ -137,7 +58,10 @@ VectorPtr ConstrainedVectorGenerator::generateFlat(
       writer.commitNull();
     } else {
       VELOX_DYNAMIC_TYPE_DISPATCH(
-          writeOne, type->kind(), variant, writer.current());
+          generator_spec_utils::writeOne,
```
kagamiori, line 61: nit: It looks like writeOne is not relevant to generator_spec_utils, so could we move it out of this namespace?

kagamiori (approved):
LGTM. Thank you for adding this support!

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
kevinwilfong: I can just get rid of type from the public interface, we only need it for converting complex types that contain intermediate types, which only happens within this cpp file.

PR #12747, opened 2025-03-21
build(pyvelox)!: Build System Modernization and Package Architecture Improvements

assignUser:
This PR addresses critical linking issues with PyArrow libraries while modernizing the entire Python build system. The work evolved from fixing dynamic linking problems to a comprehensive restructuring of the PyVelox package architecture.

### Key Changes

- **Dynamic linking against PyArrow libraries** via a custom `Findpyarrow.cmake` module and RPATH manipulation
  - Fixed wheel portability on macOS with a custom fork of `delocate-wheel` to work around a bug in the exclude functionality

- **Replaced legacy `setup.py` with modern `pyproject.toml`** using [`scikit-build-core`](https://scikit-build-core.readthedocs.io/en/latest/)
  - Comprehensive configuration for build, test, and packaging
  - Dynamic version generation using `setuptools_scm` based on git tags
  - Simplified dependency management

- **Consolidated all PyVelox code under `/python`** 
  - Better structure following Python conventions
  - Preparation for future decoupling of PyVelox from Velox (using installed libvelox)
  - Renamed `pyvelox.pyvelox` module to `pyvelox` for better clarity and removal of an 'empty' root module

- **Added centralized configuration of pybind11 modules** with wrapper function
  - Consistent handling of module properties
  - Proper RPATH configuration for both Linux and macOS

- **Improved CI and development workflow**
  - New `VELOX_PYTHON_LEGACY_ONLY` option to build only module needed for signature verification
  - Updated Makefile to work with new build system and optional `uv` support
  - Fixed test discovery and execution

- **Fixed bugs**
  - Memory manager double initialization issue
  - State carryover between tests with `unregister_all`
  - Several compiler warnings


assignUser:
I don't see the errors in https://github.com/facebookincubator/velox/actions/runs/14118870688/job/39555199264?pr=12747 locally. Will work on it once this is merged see #12836

CMake/Findpyarrow.cmake
```diff
@@ -0,0 +1,69 @@
...
+import pyarrow
+print(pyarrow.get_include())
+"
+  OUTPUT_VARIABLE _pyarrow_include_dir
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+execute_process(
+  COMMAND
+    "${Python_EXECUTABLE}" -c "\
+import pyarrow
```
kgpai, line 29: Should we do some version check for pyarrow ? Also what happens if you change venv's etc . Curious what other native python packages which have to depend on pyarrow.

CMakeLists.txt
```diff
@@ -205,8 +206,11 @@ if(${VELOX_BUILD_PYTHON_PACKAGE})
   set(VELOX_ENABLE_EXEC ON)
   set(VELOX_ENABLE_AGGREGATES ON)
   set(VELOX_ENABLE_SPARK_FUNCTIONS ON)
-  set(VELOX_BUILD_RUNNER OFF)
-  set(VELOX_BUILD_TEST_UTILS ON) # required for velox_exec_test_lib
+
+  if(NOT VELOX_PYTHON_LEGACY_ONLY)
```
kgpai, line 210: Thanks !

Makefile
```diff
@@ -219,12 +221,40 @@ help:					#: Show the help messages
...
+		${VENV} $(PYTHON_VENV); \
+	fi
+
+check-pip-version: python-venv # We need a recent pip for '-C'
+	if [ "$(PIP)" == "uv pip" ]; then \
+		exit 0; \
+	fi; \
+	source .venv/bin/activate; \
+	found=0; \
+	for pydir in $(PYTHON_VENV)/lib/python*/site-packages; do \
```
kgpai, line 235: nit  - cant we just run pip -v in the shell , and then use that to get major / minor version .

python/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+**This library is currently in Alpha stage and does not have a stable release. The API and implementation may change based on
+user feedback or performance. Future changes may not be backward compatible.
+If you have suggestions on the API or use cases you'd like to be covered, please open a
+GitHub issue. We'd love to hear thoughts and feedback.**
+
+
+## Prerequisites
+
+You will need Python 3.9 or later. We recommend using `uv` for Python Environment Management [uv](https://github.com/astral-sh/uv) is a fast Python package installer and environment manager written in Rust.
```
kgpai, line 11: Thanks for updating this too !

python/src/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+    velox_vector
+    velox_core
+    velox_exec
+    velox_parse_parser
+    velox_functions_prestosql
+    velox_functions_spark
+    velox_aggregates
+    velox_functions_spark_aggregates)
+
+if(VELOX_PYTHON_LEGACY_ONLY)
```
kgpai, line 37: Is it possible to print out what is dependent on this so we can flag PYTHON_LEGACY_ONLY at the top level ?

python/src/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+endif()
+
+add_library(velox_py_init init/PyInit.cpp)
+
+target_link_libraries(
+  velox_py_init
+  PRIVATE velox_memory)
+
+# pyvelox.type library:
+add_library(velox_py_type_lib type/PyType.cpp)
```
kgpai, line 49: Neat !

python/src/init/PyInit.cpp
```diff
@@ -14,23 +14,17 @@
...
-  velox::memory::initializeMemoryManager({});
-}
-
 void initializeVeloxMemory() {
-  // Initialize Velox once per process.
-  folly::call_once(initOnceFlag, initializeVeloxMemoryOnce);
+  if (not velox::memory::MemoryManager::testInstance()) {
+    // Enable full Velox stack trace when exceptions are thrown.
+    FLAGS_velox_exception_user_stacktrace_enabled = true;
+    velox::memory::initializeMemoryManager({});
```
kgpai, line 26: Shouldnt we add this under folly::call_once so that it isnt called multiple times ? Is library init guaranteed to be single threaded ?

kgpai (approved):
Thanks for this massive change @assignUser !

CMake/Findpyarrow.cmake
assignUser: If there is a version bound we can add it to our pyproject.toml, no problem. 

Well pyarrow is a dependency so will always be installed when building the package/installing the package through python means, this will just give us the paths to link against while building. When installing the package pyarrow will also be installed and we then jsut find it through rpath.

Pretty sure they do this, after all that's why pyarrow added these functions ^^ though I think it might make sense to have them generate a cmake config so we can drop the `Findpyarrow.cmake` :D

Makefile
assignUser: uh, yeah I guess that would work too. For some reason I wanted to get it from a file.

python/src/CMakeLists.txt
assignUser: I want to reorganize the options soon, it's getting pretty unwildy

python/src/init/PyInit.cpp
assignUser: That was the previous solutio but that didn't work and threw runtime errors due to double init. I didn't investigate the reason as I don't think the minor overhead to `testInstance` is harming anything in the context of python. I would guess it's single threaded but I don't actually know.

python/src/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+  velox_cursor
+  velox_hive_connector
+  velox_aggregates
+  velox_functions_prestosql
+  velox_parse_expression
+  velox_exec_test_lib
+  velox_tpch_connector
+  velox_dwio_dwrf_reader
+  velox_dwio_dwrf_common
+  velox_dwio_dwrf_writer
```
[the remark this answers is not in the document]
assignUser: Ah I guess we'll have to rebase this  after those changes then.

assignUser:
Working on rebasing this now.

python/src/CMakeLists.txt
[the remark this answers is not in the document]
pedroerp: I ended up removing Parquet to add as its own PR, so this should be good.

Issue #12734, opened 2025-03-20
Make the parquet writer option configurable


anlowee:
### Background
When evaluating the compression ratio between the Velox parquet writer and the Presto parquet writer, I created a table with only one column of `SMALLINT` type and inserted 10K random generated records. Both writers enabled the dictionary encoding by default, the Presto parquet writer seemed to trigger the fallback mechanism and switch to `RLE/BIT_PACKED` encoding, while the Velox parquet writer didn't trigger the fallback mechanism implemented in the `arrow` library [here](https://github.com/apache/arrow/blob/main/cpp/src/parquet/column_writer.cc#L1683-L1696).

### Issues
I found that there are three configurations hardcoded in the `WriterOptions` [here](https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/writer/Writer.h#L91-L93). 

The first one `enableDictionary` is hardcode to `true`. Although there is code for handling non-dictionary case, but since here it is hardcoded, those code is never used. Or is it determined dynamically in the runtime? (I probably missed something).

The second one `dataPageSize` is hardcoded  to 1MB. However, in Presto, it is configurable [here](https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/ParquetFileWriterConfig.java#L49-L59).

The third one `dictionaryPageSizeLimit` is also hardcoded to 1MB (the default value used by `arrow` is also [1MB](https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/writer/arrow/Properties.h#L209), but it is weird to use a magic number here but no actual code using this variable), which is supposed to be used to trigger `arrow`'s fallback mechanism to ensure the dictionary won't go too large. But **there are no other places using this variable** except unit test, this might be the root case of the difference of the behaviours when I was evaluating compression ratio between Presto and Velox because they might use a different threshold.

### Potential Enhancement
I think the enhancement would be simple as two steps:
- **Step 1**: Add three options in `hive.properties` to make the `enableDictionary`, `dataPageSize` and `dictionaryPageSizeLimit` configurable in Velox.
- **Step 2**: Correctly pass `dictionaryPageSizeLimit` to the `arrow` API, so that the fallback mechanism can work correctly.
If it looks good I can submit a PR.

PR #12755, opened 2025-03-21
feat: Make page size and batch size configurable for velox parquet writer

anlowee:
#12734 
In this issue, I found that some options in the `WriterOptions` are hardcoded but there are setters for them so it does not make sense that they are not configurable. **In this PR, I mainly made the page size and the batch size configurable.** The page size is configurable in Presto [here](https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/ParquetFileWriterConfig.java#L49-L59), so we could make it configurable in velox as well.

However I think it is not enough to just make the page size configurable, because batch size could also affect the page size since velox parquet writer writes the data in batch, and each batch at most take one page. So if the batch size is too big, the actual page size could be far larger than the page size set by user. So user might also want to set the batch size.

### DIFF
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -142,7 +142,9 @@ std::shared_ptr<WriterProperties> getArrowParquetWriterOptions(
         getArrowParquetCompression(columnCompressionValues.second));
   }
   properties = properties->encoding(options.encoding);
-  properties = properties->data_pagesize(options.dataPageSize);
+  properties =
+      properties->data_pagesize(options.dataPageSize.value_or(
+        facebook::velox::parquet::arrow::kDefaultDataPageSize));
   properties = properties->max_row_group_length(
       static_cast<int64_t>(flushPolicy->rowsInRowGroup()));
   properties = properties->codec_options(options.codecOptions);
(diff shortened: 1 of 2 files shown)


anlowee:
I have added a unit test for this, it would be great if someone can take a look

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -260,6 +263,58 @@ std::optional<bool> getParquetDataPageVersion(
...
+  }
+  return std::nullopt;
+}
+
+std::optional<int64_t> getParquetBatchSize(
+    const config::ConfigBase& config,
+    const char* configKey) {
+  if (const auto batchSize = config.get<std::string>(configKey)) {
+    try {
+      int64_t value = std::stoll(
```
majetideepak, line 304: Velox uses `folly::to<int64_t>` at other places.

majetideepak (commented):
Can we add a test for incorrect values as well?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -260,6 +265,50 @@ std::optional<bool> getParquetDataPageVersion(
   return std::nullopt;
 }
 
+std::optional<int64_t> getParquetPageSize(
```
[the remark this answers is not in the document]
anlowee: Fixed, now using `config::toCapacity`

velox/dwio/parquet/writer/Writer.cpp
anlowee: Fixed

anlowee:
Fixed all comments. I added another two cases: default config and incorrect config.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -147,6 +147,143 @@ TEST_F(ParquetWriterTest, compression) {
...
+      incorrectPageSizeConfigFromFile = {
+          {parquet::WriterOptions::kParquetHiveConnectorWritePageSize, "NaN"},
+      };
+  const std::unordered_map<std::string, std::string>
+      incorrectPageSizeSessionPropertiesFromFile = {
+          {parquet::WriterOptions::kParquetSessionWritePageSize, "NaN"},
+      };
+
+  // Values cannot be parsed so that the exception is thrown
+  EXPECT_THROW(
```
majetideepak, line 262: Using VELOX_ASSERT_THROW allows us to test the exact error message. Same below.

velox/dwio/parquet/writer/Writer.h
```diff
@@ -122,6 +123,14 @@ struct WriterOptions : public dwio::common::WriterOptions {
       "hive.parquet.writer.datapage_version";
   static constexpr const char* kParquetHiveConnectorDataPageVersion =
       "hive.parquet.writer.datapage-version";
+  static constexpr const char* kParquetSessionWritePageSize =
+      "hive.parquet.writer.page_size";
```
majetideepak, line 127: Please also document these here `velox/docs/configs.rst` under `Parquet File Format Configuration`

anlowee:
Fixed latest comments.

majetideepak (approved):

anlowee:
Seems this PR failed the internal lint check, how to know what the error is? I currently have run the `make format-fix`.

anlowee:
@majetideepak Anything I can do at this stage? It seems there is a "Facebook Internal - Linter" error.

majetideepak:
Meta imported this. They will let us know if something needs to be fixed.

kgpai:
@anlowee Have to fix a few breaking tests internally before I can land this..

PR #12760, opened 2025-03-22
feat(python): Generate multiple files from TPC-H dbgen

pedroerp:
Summary:
Adding a change to PyVelox to allow dbgen generation of TPC-H data to
produce multiple files, according to num_parts specified by the user. To do so,
the following changes are requires:
- Query configs needs to be more general so we can increase the maximum number
  of writers.
- Create a shared context structure across related plan builders.
- maxDrivers needs to be configurable by the user
- TableWriter needs to take a path in addition to the file name.

Differential Revision: D71663703


velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -35,7 +35,23 @@ class PyVector;
 using TScanFiles = std::unordered_map<
     core::PlanNodeId,
     std::vector<std::shared_ptr<connector::ConnectorSplit>>>;
-using TScanFilesPtr = std::shared_ptr<TScanFiles>;
+using TQueryConfigs = std::unordered_map<std::string, std::string>;
+
+// Stores the context for a particular plan generation, since a single plan may
```
kostasxmeta, line 40: nit: Let's use /// for documentation on the struct

velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -66,24 +84,26 @@ class PyPlanNode {
     return planNode_->toString(detailed, recursive);
   }
 
-  const TScanFilesPtr& scanFiles() const {
-    return scanFiles_;
+  const TScanFiles& scanFiles() const {
```
kostasxmeta, line 87: Just for my information, why are we prefixing these names with T? What does it stand for?

velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -66,24 +84,26 @@ class PyPlanNode {
...
 class PyPlanBuilder {
  public:
-  /// Constructs a new PyPlanBuilder. If provided, the planNodeIdGenerator is
-  /// used; otherwise a new one is created.
-  PyPlanBuilder(
-      const std::shared_ptr<core::PlanNodeIdGenerator>& generator = nullptr,
-      const TScanFilesPtr& scanFiles = nullptr);
+  /// Constructs a new PyPlanBuilder. If provided, the planContext is used;
+  /// otherwise a new one is created.
+  PyPlanBuilder(const std::shared_ptr<PyPlanContext>& planContext = nullptr);
```
kostasxmeta, line 106: So in the case where we are creating the context implicitly the user doesn't have access to it right? It's only for cases where we have a straightforward plan that e.g. doesn't branch out?

velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -93,9 +113,8 @@ class PyPlanBuilder {
   // Returns a new builder sharing the plan node id generator, such that the new
```
kostasxmeta, line 113: nit: Let's use /// here for the documentation

velox/py/plan_builder/plan_builder.cpp
```diff
@@ -108,14 +108,21 @@ PYBIND11_MODULE(plan_builder, m) {
...
+          py::arg("output_file") = std::nullopt,
+          py::arg("output_path") = std::nullopt,
           py::arg("connector_id") = "hive",
           py::arg("output_schema") = std::nullopt,
           py::doc(R"(
         Adds a table write node to the plan.
 
         Args:
           output_file: Name of the file to be written.
+          output_path: The output path where output files will be written.
```
kostasxmeta, line 120: Btw this is currently the only way to write multiple files right? There is no way to specify multiple file names and write those out for example. Maybe it's worth explicitly specifying this somewhere.

velox/py/plan_builder/PyPlanBuilder.h
```diff
@@ -35,7 +35,23 @@ class PyVector;
...
+      std::make_shared<core::PlanNodeIdGenerator>()};
+
+  // For API convenience, we allow clients to specify the files themselves when
+  // they add scans, so that we can automatically create and attach splits to
+  // the task. If specified in the .tableScan() call, store these files here.
+  TScanFiles scanFiles;
+
+  // As we create the plan, we may capture query configs that will need to be
+  // passed to the runner while executing the task.
+  TQueryConfigs queryConfigs;
```
kostasxmeta, line 53: Do these take precedence versus query configs that are specified e.g. in the localRunner?

kostasxmeta (approved):

velox/py/plan_builder/PyPlanBuilder.h
pedroerp: the opposite. Configs explicitly added in local runner take precedence. Will add to the coment.

velox/py/plan_builder/PyPlanBuilder.h
pedroerp: in Velox we usually prefix type names (typedefs) by T by convention. "T" stand for "type".

velox/py/plan_builder/PyPlanBuilder.h
pedroerp: Yes, these are not visible to the user (in the Python land). The common case is that this is null (so one is internally created), but if you have nested plan builders (for joins and similar), plan_builder->new_builder() will take care of sharing the context.

Issue #12673, opened 2025-03-17
Errors when pushdown simple aggregation(sum) with AVX2 optimization disabled


kevinw66:
### Reproduce
**SQL statement**: select sum(l_linenumber), l_orderkey from lineitem where l_orderkey > 59800
**Parquet file**: [lineitem.parquet](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/test/resources/tpch-data-parquet/lineitem/part-00000-6c374e0a-7d76-401b-8458-a8e31f8ab704-c000.snappy.parquet)
**Code changes**: 
Delete these lines (Disable avx2 optimization):
```cpp
// RleBpDataDecoder.h

template <bool hasNulls, typename Visitor>
void readWithVisitor(const uint64_t* nulls, Visitor visitor) {
  // processRle is not implemented for signed char ColumnVisitor
  if constexpr (!std::is_same_v<typename Visitor::DataType, signed char>) { // <-- delete
   if (dwio::common::useFastPath<Visitor, hasNulls>(visitor)) {             // <-- delete
      fastPath<hasNulls>(nulls, visitor);                                   // <-- delete
      return;                                                               // <-- delete
    }                                                                       // <-- delete
  }                                                                         // <-- delete
  int32_t current = visitor.start();
  skip<hasNulls>(current, 0, nulls);
  int32_t toSkip = 0;
  // ignore others
}
```
and print logs on here:
```cpp
// ColumnVisitors.h

template <typename T, typename TFilter, typename ExtractValues, bool isDense>
inline void ColumnVisitor<T, TFilter, ExtractValues, isDense>::addResult(
    T value) {
  values_.addValue(rowIndex_ + numValuesBias_, value); // <-- print rowIndex_ and numValuesBias_
}
```
```cpp
// AggregationHook.h
// SumHook

template <typename T>
void addValueImpl(vector_size_t row, T value) {
  auto group = findGroup(row); // <- print group, row and value
  clearNull(group);
  add(*reinterpret_cast<TAggregate*>(group + offset_), value);
}
```

### Logs
```
2025-03-15T04:54:43.5781289Z I20250315 04:54:43.561138   988 ColumnVisitors.h:522] rowIndex_: 0, numValuesBias_: 0
2025-03-15T04:54:43.5782444Z I20250315 04:54:43.561159   988 AggregationHook.h:143] group: 0x7f443c3ea638, row: 0, value: 1
[...]

PR #11010, opened 2024-09-16
Parquet LazyVector support

Yuhta:
Summary:
Parquet reader was not generating `LazyVector` at all.  Fix this and
address the bugs along the way:

1. Fix the `rowIndex` passed to hook by considering `numValuesBias_`.
2. Add the missing `setNumValues` call in Parquet `StringDecoder`.
3. Fix the Parquet string dictionary column visitor call and reuse the same code
   in DWRF string dictionary column reader.
4. Fix write over boundary in `VectorLoader::load`.

Fix https://github.com/facebookincubator/velox/issues/9563

Differential Revision: D62724551

PR #12761, opened 2025-03-22
fix: Group index should not reset when pushdown aggregation

kevinw66:
fixes: #12673

### DIFF
--- a/velox/exec/AggregationHook.h
+++ b/velox/exec/AggregationHook.h
@@ -45,7 +45,7 @@ class AggregationHook : public ValueHook {
 
  protected:
   inline char* findGroup(vector_size_t row) {
-    return groups_[row];
+    return groups_[currentRow_++];
   }
 
   inline bool clearNull(char* group) {
--- a/velox/dwio/common/ColumnVisitors.h
+++ b/velox/dwio/common/ColumnVisitors.h
@@ -84,28 +84,28 @@ class ExtractToHook {
   static constexpr bool kSkipNulls = THook::kSkipNulls;
 
   explicit ExtractToHook(ValueHook* hook)
-      : hook_(*reinterpret_cast<THook*>(hook)) {}
+      : hook_(reinterpret_cast<THook*>(hook)) {}
 
   bool acceptsNulls() {
-    return hook_.acceptsNulls();
+    return hook_->acceptsNulls();
   }
 
   template <typename T>
   void addNull(vector_size_t rowIndex) {
-    hook_.addNull(rowIndex);
+    hook_->addNull(rowIndex);
   }
 
   template <typename V>
   void addValue(vector_size_t rowIndex, V value) {
-    hook_.addValueTyped(rowIndex, value);
+    hook_->addValueTyped(rowIndex, value);
   }
 
-  auto& hook() {
-    return hook_;
+  THook& hook() {
+    return *hook_;
   }
 
  private:
-  THook hook_;
+  THook* hook_;
 };
 
 class ExtractToGenericHook {


velox/exec/AggregationHook.h
```diff
@@ -45,7 +45,7 @@ class AggregationHook : public ValueHook {
 
  protected:
   inline char* findGroup(vector_size_t row) {
-    return groups_[row];
+    return groups_[currentRow_++];
```
Yuhta, line 48: We probably should remove `currentRow_` here to make it stateless.  Can you explain more why we cannot use `row` here?  We should probably fix where it is calculated.

kevinw66: I've tried incrementing the numValues of the reader. This value will be set to numValueBias_ when reading new page data: https://github.com/facebookincubator/velox/blob/215213a6292d4014eb2374381cf20e0d35534e5d/velox/dwio/parquet/reader/PageReader.h#L551-L552, which can solve this specific issue. But it will throw SIGSEGV for other SQL statements. I didn't dig into the reason. So, I'm thinking whether the reader is specifically designed for `ExtractToReader`. Moreover, I didn't see the use of the row variable in ExtractToReader either(it uses `numValues` which maintain by itself): https://github.com/facebookincubator/velox/blob/215213a6292d4014eb2374381cf20e0d35534e5d/velox/dwio/common/ColumnVisitors.h#L67-L70. So I followed a similar approach and used currentRow_ in ExtractToHook and ExtractToGenericHook.

kevinw66: @Yuhta Hi, any thoughts here?

Yuhta: You probably need to find the exact `addValue` call and fix the argument there

kevinw66: @Yuhta Regarding to this SQL statement: `select sum(l_linenumber), l_orderkey from lineitem where l_orderkey > 59800`, the `hasFilter` will be false and `numValuesBias_` will use the value of `numValues_` in the reader in here: https://github.com/facebookincubator/velox/blob/215213a6292d4014eb2374381cf20e0d35534e5d/velox/dwio/parquet/reader/PageReader.h#L551-L552 https://github.com/facebookincubator/velox/blob/215213a6292d4014eb2374381cf20e0d35534e5d/velox/dwio/parquet/reader/PageReader.h#L362-L370 If my understanding of this code is correct, if we want to increment this `numValues_`, we also need to store the corresponding value into reader, which will result in additional memory consumption. Otherwise, there will be illegal memory access. Or do you have other suggestions here?

kevinw66: @Yuhta Hi, I updated the code and used the same approach as that applied to `StringDecoder.h` in #11010

kevinw66:
@Yuhta Could you please help review this again? many thanks!

Yuhta:
@kevinw66 Would you add a test for it?

kevinw66:
@Yuhta Sorry for the late reply, while I'm trying to add one on `ParquetTableScanTest`, I found there is no parquet file under `velox/dwio/parquet/tests/examples` can reproduce the bug, should I add one(which described in the issue, but it's quite a large file compare to existing ones)?

Yuhta:
@kevinw66 Ideally we want to minimize the testing data that can repro the issue, but if the problem requires large data by nature, it's also ok to add large data.  I don't think this one need though, you might need to tweak the paging policy a little bit when you generate test data.

kevinw66:
@Yuhta Done, thanks!

velox/dwio/common/DirectDecoder.h
```diff
@@ -77,6 +78,10 @@ class DirectDecoder : public IntDecoder<isSigned> {
             this->template skip<false>(toSkip, current, nullptr);
           }
           if (atEnd) {
+            if constexpr (Visitor::kHasHook) {
+              visitor.setNumValues(
+                  Visitor::kHasFilter ? numValues : visitor.numRows());
```
Yuhta, line 83: `kHasHook` and `kHasFilter` cannot be both true.  Also is this file used in Parquet?  If not we don't need to change this file.

kevinw66: @Yuhta 
> `kHasHook` and `kHasFilter` cannot be both true

So this is different from [here](https://github.com/facebookincubator/velox/blame/main/velox/dwio/parquet/reader/StringDecoder.h#L79)? I saw it also check both `kHasHook` and `kHasFilter`.

> Also is this file used in Parquet?

Yes, in the ut, if column `b` is `PLAIN`, it will use this, if it's `RLE_DICTIONARY`, it will use `RleBpDataDecoder`, but I didn't test `DeltaBpDecoder`.

kevinw66: > `kHasHook` and `kHasFilter` cannot be both true. Also is this file used in Parquet? If not we don't need to change this file.

@Yuhta removed `kHasFilter` check, also removed for [StringDecoder.h](https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/reader/StringDecoder.h#L79), thanks!

zml1206:
Can you continue this PR? I have encountered similar problems in the ARM environment. Thanks @kevinw66  @Yuhta 
I picked this PR to solve the problem, but I'm not sure if it will cause other problems.

FelixYBW:
@pedroerp such bug fixes are not reviewed and merged in time then closed. It may seriously affect contributors' motivation. Then it wasted others time to identify and debug time.

@kevinw66 would you like to continue the PR? Looks https://github.com/facebookincubator/velox/pull/14332 can fix the same issue.

pedroerp:
@FelixYBW does #14332 supersede this PR or do we need both? We need to get more people ramped up and able to review  Parquet code.

FelixYBW:
Yes, #14332 can fix the bug. Thank you for reviewing.

PR #12764, opened 2025-03-23
feat: Add Spark CAST(bool as timestamp)

ArnavBalyan:
Add bool to timestamp cast support for Spark.

Spark's implementation: https://github.com/apache/spark/blob/v3.5.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L625.


velox/expression/CastExpr-inl.h
```diff
@@ -285,6 +285,14 @@ void CastExpr::applyCastKernel(
       return;
     }
 
+    if constexpr (
+        (FromKind == TypeKind::BOOLEAN) && ToKind == TypeKind::TIMESTAMP) {
+      const auto castResult =
+          hooks_->castBooleanToTimestamp((bool)inputRowValue);
```
rui-mo, line 291: Do we need the cast as 'inputRowValue' should be of bool type?

velox/functions/sparksql/specialforms/SparkCastHooks.cpp
```diff
@@ -77,6 +77,12 @@ Expected<std::optional<Timestamp>> SparkCastHooks::castDoubleToTimestamp(
   return castNumberToTimestamp(value);
 }
 
+Expected<Timestamp> SparkCastHooks::castBooleanToTimestamp(bool val) const {
+  // Spark true evaluates to 1970-01-01 00:00:00.000001 and false to epoch (0 microseconds).
+  // Use fromMicrosNoError to interpret '1' as 1 microsecond offset from epoch.
```
rui-mo, line 82: Perhaps move the comments to the header.

rui-mo (commented):
Thanks.

velox/expression/CastExpr-inl.h
rui-mo: I assume the case is different because above `castIntToTimestamp` only supports int64_t and the `inputRowValue` could be int8_t, int16_t, int32_t, int64_t so it needs a cast, while the input for `castBooleanToTimestamp` must be bool.

velox/expression/PrestoCastHooks.cpp
```diff
@@ -71,6 +71,11 @@ Expected<int32_t> PrestoCastHooks::castStringToDate(
   return util::fromDateString(dateString, util::ParseMode::kPrestoCast);
 }
 
+Expected<Timestamp> PrestoCastHooks::castBooleanToTimestamp(bool seconds) const {
+  return folly::makeUnexpected(
+      Status::UserError("Conversion to Timestamp is not supported"));
```
rui-mo, line 76: Would you check if it's indeed unsupported in Presto?

velox/functions/sparksql/specialforms/SparkCastHooks.h
```diff
@@ -57,6 +57,10 @@ class SparkCastHooks : public exec::CastHooks {
   /// whitespaces before cast.
   StringView removeWhiteSpaces(const StringView& view) const override;
 
+  // Spark true evaluates to 1970-01-01 00:00:00.000001 and false to epoch (0 microseconds).
+  // Use fromMicrosNoError to interpret '1' as 1 microsecond offset from epoch.
```
rui-mo, line 61: > // Use fromMicrosNoError to interpret '1' as 1 microsecond offset from epoch.

Perhaps drop this comment as we don't repeat the implementation in the documentation.

velox/docs/functions/spark/conversion.rst
```diff
@@ -277,4 +277,20 @@ Valid examples
...
+
+From boolean
+^^^^^^^^^^^^
+
+Casting from boolean to timestamp is supported.
+
+* ``true`` is cast to **1970-01-01 00:00:00.000001**
+* ``false`` is cast to **1970-01-01 00:00:00** (epoch)
+
+This behavior matches Spark's semantics for boolean-to-timestamp casts.
```
rui-mo, line 289: Perhaps drop this sentence as this is the document for Spark casts, so the behaviours all match with Spark.

rui-mo (commented):
Thanks. Looks good overall.

rui-mo (commented):
Thanks. Looks good! Would you fix the format issue?

rui-mo:
@ArnavBalyan Looks like there are some conflicts need to be resolved, thanks.

rui-mo (approved):

velox/expression/PrestoCastHooks.cpp
```diff
@@ -78,6 +78,12 @@ Expected<int32_t> PrestoCastHooks::castStringToDate(
   return util::fromDateString(dateString, util::ParseMode::kPrestoCast);
 }
 
+Expected<Timestamp> PrestoCastHooks::castBooleanToTimestamp(
+    bool seconds) const {
+  return folly::makeUnexpected(
+      Status::UserError("Conversion to Timestamp is not supported"));
```
rui-mo, line 84: nit: consider including the from-type in the error message.

PR #12765, opened 2025-03-23
misc: Avoid data copy for duplicate result and simplify the impementation

xiaoxmeng:
Summary:
We don't need to copy data for duplicate result for inputs with the same key. Extend row serde to support serialize string view.
This optimize cpu/memory and simplify the implementation
Also cover the key-dedup in index join test.

Reviewed By: wenqiwooo

Differential Revision: D71670857


duanmeng (approved):
@xiaoxmeng LGTM, thanks.

PR #12766, opened 2025-03-23
feat(parquet): Add config for enabling dictionary and dictionary page size

anlowee:
In this PR, I mainly made enabling dictionary and dictionary page size. There are completed handlers for non-dictionary case in the [encoder](https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/writer/arrow/Encoding.cpp#L4155-L4227), so it should make sense that this is a configurable option.

This PR also makes the dictionary page size configurable, similar to #12755 , they share the exactly same function `getParquetPageSize` to read from the config file. So that user can customize the dictionary page size threshold to fall back to plain encoding if dictionary page size grows too big.

But the dictionary page size is a soft restriction, because the data is written in batch. If a batch contains too many values to write once, the actual dictionary page size might be far bigger than the preset threshold. So the batch size should also be configurable, which is implemented in #12755  

Fixes: #12734

### DIFF
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -268,15 +268,6 @@ std::optional<bool> isParquetEnableDictionary(
   return std::nullopt;
 }
 
-std::optional<int64_t> getParquetPageSize(
-    const config::ConfigBase& config,
-    const char* configKey) {
-  if (const auto pageSize = config.get<std::string>(configKey)) {
-    return config::toCapacity(pageSize.value(), config::CapacityUnit::BYTE);
-  }
-  return std::nullopt;
-}
-
 std::optional<bool> getParquetDataPageVersion(
     const config::ConfigBase& config,
     const char* configKey) {
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -131,15 +131,15 @@ std::shared_ptr<WriterProperties> getArrowParquetWriterOptions(
     const std::unique_ptr<DefaultFlushPolicy>& flushPolicy) {
   auto builder = WriterProperties::Builder();
   WriterProperties::Builder* properties = &builder;
-  if (options.enableDictionary.value_or(facebook::velox::parquet::arrow::DEFAULT_IS_DICTIONARY_ENABLED)) {
+  if (options.enableDictionary.value_or(
+          facebook::velox::parquet::arrow::DEFAULT_IS_DICTIONARY_ENABLED)) {
     properties = properties->enable_dictionary();
   } else {
     properties = properties->disable_dictionary();
   }
-  properties =
-      properties->dictionary_pagesize_limit(
-        options.dictionaryPageSizeLimit.value_or(
-            facebook::velox::parquet::arrow::DEFAULT_DICTIONARY_PAGE_SIZE_LIMIT));
+  properties = properties->dictionary_pagesize_limit(
+      options.dictionaryPageSizeLimit.value_or(
+          facebook::velox::parquet::arrow::DEFAULT_DICTIONARY_PAGE_SIZE_LIMIT));
   properties = properties->compression(getArrowParquetCompression(
       options.compressionKind.value_or(common::CompressionKind_NONE)));
   for (const auto& columnCompressionValues : options.columnCompressionsMap) {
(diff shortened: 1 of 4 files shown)


anlowee:
I have added the unit test, if anyone can take a look of this, it would be great.

majetideepak:
I will review this after https://github.com/facebookincubator/velox/pull/12755 gets merged.

anlowee:
Hi @majetideepak do you have comments for this PR? I guess I have fixed the all similar issues from #12755 in this PR.

majetideepak:
@anlowee I need to look at the tests. We can may be refactor some common code after https://github.com/facebookincubator/velox/pull/12755 lands.

majetideepak:
@anlowee can you rebase with the latest main? thanks.

anlowee:
> @anlowee can you rebase with the latest main? thanks.

Done. @majetideepak

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -131,9 +131,15 @@ std::shared_ptr<WriterProperties> getArrowParquetWriterOptions(
...
   auto builder = WriterProperties::Builder();
   WriterProperties::Builder* properties = &builder;
-  if (!options.enableDictionary) {
+  if (options.enableDictionary.value_or(
+          facebook::velox::parquet::arrow::DEFAULT_IS_DICTIONARY_ENABLED)) {
+    properties = properties->enable_dictionary();
+  } else {
     properties = properties->disable_dictionary();
   }
+  properties = properties->dictionary_pagesize_limit(
```
majetideepak, line 140: Should we set this below `properties = properties->enable_dictionary();` above?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -248,6 +254,20 @@ std::optional<std::string> getTimestampTimeZone(
   return std::nullopt;
 }
 
+std::optional<bool> isParquetEnableDictionary(
+    const config::ConfigBase& config,
+    const char* configKey) {
+  try {
```
majetideepak, line 260: Do we need the try-catch here? `config.get<bool>()` should throw any error?

anlowee: Yea, but I remembered that in #12755 ([this comment](https://github.com/facebookincubator/velox/pull/12755#discussion_r2024983416)), we want to use `VELOX_ASSET_THROW` to check the exception by its error message, but folly only throw `folly::ConversionError`, which is not a `VeloxException` so that we cannot check its error message by `VELOX_ASSERT_THROW`. Otherwise, we have to use `EXPECT_THROW`.

Do we want to remove the try-catch here and use `EXPECT_THROW` in the unit test?

velox/dwio/parquet/writer/Writer.cpp
anlowee: Fixed this one.

velox/dwio/parquet/writer/Writer.cpp
majetideepak: Yes, let's use EXPECT_THROW. Thanks.

majetideepak: Does Velox not hit this check?
https://github.com/facebookincubator/velox/blob/main/velox/common/config/Config.h#L62

majetideepak: It is odd to get a folly error. We should fix this if that is the case.
Do you have a reproducer?

anlowee: Yea, it didn't hit that one, because the `get` I called here is actually [this one](https://github.com/facebookincubator/velox/blob/main/velox/common/config/Config.h#L115-L127) (where you can see it calls `folly::to<T>` and I think this is why it throws a folly error). But the one hit the check you mentioned is [this one](https://github.com/facebookincubator/velox/blob/main/velox/common/config/Config.h#L107-L112) (where you can see it calls `entry.toT`.

anlowee: My reproducer for this error is in my unit test and [here](https://github.com/facebookincubator/velox/pull/12766/files#diff-1dfe222f0cb53bc084be4b1050bfab5dec06e65f0eba5d7533901d427c62fb67R230-R237) is my code. Since I used try-catch only to catch `folly::ConversionError` so if this assert passes, then it means it didn't hit the check you mentioned.

majetideepak: This is fine. We want VeloxException. Thanks.

majetideepak (approved):

Issue #12614, opened 2025-03-12
array_sort_desc(array(boolean)): Value mismatch between Presto and Velox


peterenescu:
### Description

Value mismatch between Presto and Velox for function `array_sort_desc`. My presumption is this is a result of our special boolean handling.

### Error Reproduction

Presto
```
presto:tpch> select cast(array_sort_desc(c0) as array(SMALLINT)) from (values array[false, true, true]) t(c0);
   _col0   
-----------
 [1, 1, 0] 
(1 row)

presto:tpch> select cast(array_sort_desc(c0) as array(SMALLINT)) from (values array[false, true, false, true, false, false]) t(c0);
       _col0        
--------------------
 [1, 1, 0, 0, 0, 0] 
(1 row)

presto:tpch> select array_sort_desc(c0) from (values array[false, true, false, true, false, false]) t(c0);
                  _col0                   
------------------------------------------
 [true, true, false, false, false, false] 
(1 row)

Query 20250312_000107_00001_7t8j6, FINISHED, 1 node
Splits: 17 total, 17 done (100.00%)
[Latency: client-side: 49ms, server-side: 43ms] [0 rows, 0B] [0 rows/s, 0B/s]
```
Velox
```
presto:di> select cast(array_sort_desc(c0) as array(SMALLINT)) from (values array[false, true, true]) t(c0);
   _col0              
-----------
 [1, 0, 0] 
(1 row)
presto:di> select cast(array_sort_desc(c0) as array(SMALLINT)) from (values array[false, true, false, true, false, false]) t(c0);
       _col0          
--------------------
 [1, 1, 0, 1, 0, 0] 
(1 row)
presto:di> select array_sort_desc(c0) from (values array[false, true, false, true, false, false]) t(c0);
                  _col0                  
-----------------------------------------
 [true, true, false, true, false, false] 
(1 row)
```

### Relevant logs

```bash

```

PR #12770, opened 2025-03-24
fix: Fix the bug related to sorting of boolean in array_sort_desc

prashantgolash:
Summary:
https://github.com/facebookincubator/velox/issues/12614. 

It looks we have minor bug in sort desc of boolean type. Fixed the issue

Differential Revision: D71689192


kagamiori (approved):
LGTM. Thank you for the fix!

PR #12771, opened 2025-03-24
feat: Add serialization and deserialization of Geometry type to/from WKT and WKB

wraymo:
Summary
This PR is a split from #12053 with additional improvements. It introduces serialization and deserialization support for the Geometry type to/from WKT and WKB formats, along with corresponding UDFs. Unit tests are included for both valid and invalid cases, and end-to-end testing has been performed using `presto-cli`.

As discussed in #12274, the new UDFs are added to `GeometryFunctions` and their registration is added to `GeometryFunctionsRegistration`. `GeospatialFunctions` and `GeospatialFunctionsRegistration` will be renamed.


wraymo:
@jagill Please take a look when you get a chance. Thanks!

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+  };
+
+  enum class EsriShapeType : int {
+    POINT = 1,
+    POLYLINE = 3,
+    POLYGON = 5,
+    MULTI_POINT = 8
+  };
+
+  class SliceInput {
```
jagill, line 69: Could you use https://github.com/facebookincubator/velox/blob/main/velox/common/base/IOUtils.h instead?  This seems like it's re-implementing that.

jagill (changes requested):
This is looking good!  Once we ship this we can parallelize on many geometry functions.  Great tests, btw.

First, please rebase past #12778.
Second, let's use common stream de/serializing functions.  I'd rather improve the common utils than make new ones here.
Third, for the thread_local variables, I'm wary of Static Initialization Order Fiasco.  Is it needed because the function creates a static GeometryFactor?  How do those two parts interact?

Thanks!

jagill:
Also, let's catch the exceptions from GEOS and wrap them in PrestoUserErrors with better messages.  I've done this in BingTileFunctions if you want a code pointer.

jagill:
We should also include the GEOS dependency in `functions/prestosql/CMakeLists.txt`

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -0,0 +1,92 @@
...
+template <typename T>
+struct StGeomFromBinaryFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE bool call(
+      out_type<Geometry>& result,
+      const arg_type<Varbinary>& wkb) {
+    thread_local geos::io::WKBReader reader;
+    auto geosGeometry = reader.read(
+        reinterpret_cast<const unsigned char*>(wkb.data()), wkb.size());
```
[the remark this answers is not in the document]
wraymo: Sure

wraymo:
> We should also include the GEOS dependency in `functions/prestosql/CMakeLists.txt`

Do you mean we'll create files like `BingTileFunctions.cpp` or `GeometryFunctions.cpp` that depend on GEOS in this directory?

jagill:
> I think most of the exceptions come from GEOS and get caught by Velox as `VeloxUserError` (I’ve added test cases for those). Some others come from our serde code—like out-of-bounds, unknown types, or type mismatches.
> 
> I checked out `BingTileFunctions`, and I can add a `try...catch` block to handle those two kinds of errors separately and return Status::UserError. But from what I can tell, the actual error message wouldn’t really change.

I think the messages that come from GEOS are unclear, especially without context.  Take the test example of "Expected word but encountered end of stream".  How is the author expected to know that's from a WKT, much less a geometry function?  At the very least we prefix the errors with the function or action, like "Error converting WKT to Geometry: {errorMessage}".

> Do you mean we'll create files like BingTileFunctions.cpp or GeometryFunctions.cpp that depend on GEOS in this directory?

I was thinking about `functions/prestosql/GeometryFunctions.h` already includes headers from GEOS, like `geos/io/WKBReader.h`.  But your method is to include it in the top-level CMake files, so the headers will be available.  I like to have dependencies declared close to where they are used, but having it top-level makes sense too.

CMakeLists.txt
```diff
@@ -631,6 +631,8 @@ if(VELOX_ENABLE_ARROW)
 endif()
 
 if(VELOX_ENABLE_GEO)
+  add_definitions(-DVELOX_ENABLE_GEO)
+
```
assignUser, line 635: please use the target based version of the function and apply it to the target like you did the extra link here https://github.com/facebookincubator/velox/pull/12771/files#diff-dd02356b3e62048cff4aea44c6b7202764835996dd5dfa16666cca020a9505afR52-R55

assignUser (changes requested):

CMakeLists.txt
wraymo: Sure, I've updated the code. I noticed several other global definitions in this CMake file—perhaps we should consider handling them similarly in the future.

wraymo:
@jagill Thanks for the comments. To my understanding, `thread_local` in this case will not cause Static Initialization Order Fiasco issue, but yeah it's not necessary. I removed them in `GeometryFunctions` in the latest commit. I also added `try...catch` block for better error messages. Regarding `SliceInput`, do you want me to replace it by `InputByteStream`?

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+    std::vector<int> partLengths(partCount);
+    if (partCount > 1) {
+      partLengths[0] = startIndexes[1];
+      for (int i = 1; i < partCount - 1; ++i) {
+        partLengths[i] = startIndexes[i + 1] - startIndexes[i];
+      }
+    }
+    partLengths[partCount - 1] = pointCount - startIndexes[partCount - 1];
+
+    std::vector<std::unique_ptr<geos::geom::LineString>> lineStrings;
```
jagill, line 566: please fix: Let's reserve the capacity here before pushing.

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      lineStrings.push_back(GEOMETRY_FACTORY->createLineString(
+          readCoordinates(input, partLengths[i])));
+    }
+
+    if (multiType) {
+      return GEOMETRY_FACTORY->createMultiLineString(std::move(lineStrings));
+    }
+
+    if (lineStrings.size() != 1) {
+      VELOX_USER_FAIL(
```
jagill, line 577: please fix: This would be a system error, not a user error.  This meant that _we_ messed up this serialization.

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+    partLengths[partCount - 1] = pointCount - startIndexes[partCount - 1];
+
+    std::unique_ptr<geos::geom::LinearRing> shell = nullptr;
+    std::vector<std::unique_ptr<geos::geom::LinearRing>> holes;
+    std::vector<std::unique_ptr<geos::geom::Polygon>> polygons;
+
+    for (int i = 0; i < partCount; i++) {
+      auto coordinates = readCoordinates(input, partLengths[i]);
+
+      if (isClockwise(coordinates, 0, coordinates->size())) {
```
jagill, line 620: Please add a comment indicating that this means a new polygon has started.

velox/functions/prestosql/GeometryUtils.h
wraymo: That makes sense. Yeah, we probably don’t need the boundary check. As for serialization, I initially implemented it the same way, but then I refactored it to avoid the extra copy. If you'd prefer the original approach, I can revert the code back to that.

jagill (commented):
I checked over the Geometry serde, and the logic looks good.  I have some please fix and suggestion items, but elsewise the logic is good to go.

I still want @kgpai's opinion on the custom byte reader/writer class.  I also would like better error handling, but that's something we can improve as it becomes an issue.

wraymo:
@jagill Thanks for the update! I’ve also updated the comment regarding the serialization helper here: [link](https://github.com/facebookincubator/velox/pull/12771#discussion_r2031459957). It would be great if you and @kgpai could take a look when you have a chance. I’ll address the issues you mentioned with the serialization logic.

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+#include <velox/common/base/Exceptions.h>
+#include <velox/type/StringView.h>
+
+namespace facebook::velox::functions {
+
+class GeometryUtils {
+ public:
+  template <typename StringWriter>
+  static void serialize(
+      const std::unique_ptr<geos::geom::Geometry>& geometry,
```
kgpai, line 32: nit: not sure if its ok  to pass a const ref to a unique_ptr and then pass the raw ptr to serialize ? (maybe shared if geometry is to be reused or move it ?)

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+    }
+
+   private:
+    template <typename T>
+    T read() {
+      if (position_ + sizeof(T) > size_) {
+        VELOX_USER_FAIL("SliceInput read exceeds buffer size.");
+      }
+      T value;
+      std::memcpy(&value, &data_[position_], sizeof(T));
```
kgpai, line 118: wont a reinterpret cast be better than doing a copy to value here?

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      default:
+        VELOX_USER_FAIL(
+            "Unrecognized geometry type: {}",
+            static_cast<uint8_t>(geometryType));
+        break;
+    }
+    return nullptr;
+  }
+
+  static bool isEsriNaN(double d) {
```
kgpai, line 238: nit: Is there a special NaN encoding for Geometric types ??

jagill: Yes, ESRI has a magic float value that it considers a NaN.  I don't know why, but I could guess to have decent equality/ordering semantics.  I don't want to support that long term (as we move of ESRI), but for the first stage we should stay byte-wise compatible with the format, because the optimizer might do constant evaluation with the Java code (which uses the old format).

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      const geos::geom::Geometry* collection,
+      SliceOutput<T>& output) {
+    output.writeByte(
+        static_cast<uint8_t>(GeometrySerializationType::GEOMETRY_COLLECTION));
+
+    for (size_t geometryIndex = 0;
+         geometryIndex < collection->getNumGeometries();
+         ++geometryIndex) {
+      auto* geometry = collection->getGeometryN(geometryIndex);
+      // Use a temporary buffer to serialize the geometry and calculate its
```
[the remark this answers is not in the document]
jagill: The second largest that I've seen is 200MB (multipolygon of Japan's boundaries). The largest was Siberia, which was substantially larger but I don't remember the size.

People will almost always need to simplify these polygons in order to fit into a Presto page, but the answer to "how large" is "as large as people can fit".

Geometries have an extremely high variance wrt size and computational complexity, so data skew is a real issue.  It's often the case that all workers are finished, except for the one working on Siberia and that times out.

jagill:
@wraymo I've extracted the core GEOS-requiring part of this PR in #13003.  We need to do some internal Buck file changes for it to land, and we can land it with less-good serialization (WKB) while we work out the serialization details here.

velox/functions/prestosql/registration/CMakeLists.txt
```diff
@@ -48,6 +49,16 @@ endif()
 velox_link_libraries(velox_functions_prestosql velox_functions_prestosql_impl
                      velox_is_null_functions simdjson::simdjson)
 
+if(VELOX_ENABLE_GEO)
```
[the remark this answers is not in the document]
wraymo: If the user doesn't need the geospatial feature?

velox/functions/prestosql/GeometryUtils.h
wraymo: Not exactly. `reinterpret_cast` could sometimes run into memory alignment crashes.

velox/functions/prestosql/GeometryUtils.h
wraymo: @kgpai I agree that for deserialization, boundary checks aren't necessary. However, as I mentioned earlier, for serialization we can't predict the final size in advance, so pre-allocating space isn't feasible. While it's possible to use `InputByteStream`, we can't directly use `OutputByteStream` in this case. Do you have any suggestions or alternatives we could consider?

velox/functions/prestosql/GeometryUtils.h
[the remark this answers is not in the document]
wraymo: @jagill We can achieve that by introducing a new I/O helper with raw data access. The current helper and `OutputByteStream` are append-only—do you think that’s a significant limitation? From what I’ve seen, geometry collections are rare in our internal data, and multipolygons don’t go through this method anyway.

jagill:
@wraymo: Sorry for the delay! #13003 has finally landed, after some great work on the build from @kgpai and @assignUser .  Please rebase on top of that and then we can get this moving again :).

wraymo:
@jagill Thanks for the update! I'll go ahead and update this PR today.

I noticed you've submitted several PRs for the geometry functions—great to see that progress! I've implemented many of those internally as well, and I’m planning to open-source them soon. Since you've already started contributing, do you have any thoughts on how we might coordinate or split up the work to avoid duplicating efforts?

velox/functions/prestosql/GeometryUtils.h
bikramSingh91: > @kgpai I agree that for deserialization, boundary checks aren't necessary. However, as I mentioned earlier, for serialization we can't predict the final size in advance, so pre-allocating space isn't feasible. While it's possible to use `InputByteStream`, we can't directly use `OutputByteStream` in this case. Do you have any suggestions or alternatives we could consider?

@wraymo This approach makes sense to me. Let's switch to using InputByteStream for deserialization and at the end we can do a sanity check for bounds by comparing offset() to the size of the input (this should be cheap as its only done once per input). For serialization, SliceOutput works well, but I have a minor suggestion: could we rename it to something like `VarbinaryWriter`?

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      sum += (p2.x - p1.x) * (p2.y + p1.y);
+    }
+    return sum > 0.0;
+  }
+
+  static void reverse(
+      const std::unique_ptr<geos::geom::CoordinateSequence>& coordinates,
+      int start,
+      int end) {
+    for (int i = 0; i < (end - start) / 2; ++i) {
```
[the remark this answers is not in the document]
bikramSingh91: similarly for `isClockwise`, a dcheck to ensure end>start

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      coordinates->setAt(temp, end - 1 - i);
+    }
+  }
+
+  template <typename T>
+  static void writeEnvelope(
+      const geos::geom::Geometry* geometry,
+      SliceOutput<T>& output) {
+    if (geometry->isEmpty()) {
+      output.writeDouble(std::nan(""));
```
bikramSingh91, line 272: just curious, do we need to set an similar magic NAN values here like those that `isEsriNaN` identifies?

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+      output.writeInt(length);
+      output.writeBytes(tempBuffer.data(), tempBuffer.size());
+    }
+  }
+
+  inline static thread_local auto GEOMETRY_FACTORY =
+      geos::geom::GeometryFactory::create();
+
+  static void skipEsriType(SliceInput& input) {
+    input.skip(4); // Esri type is an integer
```
bikramSingh91, line 489: nit: maybe use `sizeof(int)` instead? similarly in skipEnvelope()

velox/functions/prestosql/GeometryUtils.h
```diff
@@ -0,0 +1,685 @@
...
+  static std::unique_ptr<geos::geom::Geometry> readMultiPoint(
+      SliceInput& input) {
+    skipEsriType(input);
+    skipEnvelope(input);
+    int pointCount = input.readInt();
+    auto coords = readCoordinates(input, pointCount);
+    std::vector<std::unique_ptr<geos::geom::Point>> points;
+    for (size_t i = 0; i < coords->size(); ++i) {
+      points.push_back(
+          std::unique_ptr<geos::geom::Point>(GEOMETRY_FACTORY->createPoint(
```
bikramSingh91, line 530: is the memory returned by GEOMETRY_FACTORY->createPoint() owned by it? or is the caller responsible for it?

bikramSingh91 (changes requested):
can you please rebase and address comments? Thanks!

bikramSingh91:
> @jagill Thanks for the update! I'll go ahead and update this PR today.
> 
> I noticed you've submitted several PRs for the geometry functions—great to see that progress! I've implemented many of those internally as well, and I’m planning to open-source them soon. Since you've already started contributing, do you have any thoughts on how we might coordinate or split up the work to avoid duplicating efforts?

@wraymo Thats great, let me talk to folks internally, maybe we can come up with someway to sync up our efforts

bikramSingh91:
> @jagill Thanks for the update! I'll go ahead and update this PR today.
> 
> I noticed you've submitted several PRs for the geometry functions—great to see that progress! I've implemented many of those internally as well, and I’m planning to open-source them soon. Since you've already started contributing, do you have any thoughts on how we might coordinate or split up the work to avoid duplicating efforts?

@wraymo We are planning to use a google doc to coordinate efforts for adding the functions. Can you please share your email address and I will add you to the doc.

wraymo:
@bikramSingh91 Thanks for your reply! My email is raywangv@gmail.com

wraymo:
> Right, InputByteStream expects the caller is aware of the size instead. Fortunately, it looks like there is only one spot where we need this check, that is, in readGeometryCollection, for this we can pass the size as an input to this function and check against offset() for the while loop condition you mentioned.
@bikramSingh91 There might be another issue with `InputByteStream`, it uses `T value = *reinterpret_cast<const T*>(data_ + offset_);` which could cause undefined behaviors or even crashes for some platforms because of unaligned memory access

velox/functions/prestosql/geospatial/GeometrySerde.h
```diff
@@ -17,17 +17,346 @@
...
+  /// within GEOS_TRY macro: it will catch the exceptions that need to bubble
+  /// up.
+  static std::unique_ptr<geos::geom::Geometry> deserialize(
+      const StringView& geometryString) {
+    velox::common::InputByteStream inputStream(geometryString.data());
+    return deserialize(inputStream, geometryString.size());
+  }
+
+ private:
+  inline static auto geometryFactory_ = geos::geom::GeometryFactory::create();
```
bikramSingh91, line 315: this would be used simultaneously among threads. will that usage be thread safe?

velox/functions/prestosql/geospatial/GeometrySerde.h
```diff
@@ -17,17 +17,346 @@
...
+  VarbinaryWriter(StringWriter& stringWriter) : stringWriter_(stringWriter) {}
+  VarbinaryWriter() = delete;
+
+  void write(const char* data, size_t size) {
+    stringWriter_.append(std::string_view(data, size));
+  }
+
+  template <typename T>
+  void write(const T& value) {
+    VELOX_CHECK(
```
bikramSingh91, line 67: since this only depends on the template param, can we make this a static assert instead?, like:

```
static_assert(std::is_trivially_copyable_v<T>, "T must be trivially copyable");
```

bikramSingh91 (commented):
This looks good to me, just one last note on testing.
Can you tests that verify round trips of all these geometry types, just to make sure we have sanity tests for all code paths. Also, do you think its worth adding tests that verify the expected binary (serialized) value?

cc: @jagill any thoughts on testing?

bikramSingh91:
> > This looks good to me, just one last note on testing. Can you tests that verify round trips of all these geometry types, just to make sure we have sanity tests for all code paths. Also, do you think its worth adding tests that verify the expected binary (serialized) value?
> > cc: @jagill any thoughts on testing?
> 
> 1. All the tests in GeometryFunctionsTest are using some combination of de/serialization, so if the current tests pass that gives high confidence this works.
> 2. GeometryFunctionsTest has a good number of explicit serde tests.  @wraymo, could we add a couple more to test a few cases:
>    a. MultiPolygon with polygons with holes
>    b. Nested GeometryCollections
>    c. GeometryCollection with MultiLineStrings and MultiPolygons
> 
> It'd also be nice to test this against very large geometries, but I don't think we want to have those inline. @bikramSingh91, is there a good way to have a resource file with large geometries (> 1MB), or is that too slow/painful to test?

I am not sure if there is a good way to use resource files, but for the time being, we can move ahead with this PR once the tests that you suggested are added. Then we can consider testing large geometries as a follow up as.

cc @wraymo

wraymo:
@jagill @bikramSingh91 Thanks for the suggestions! I'll update it tomorrow

velox/functions/prestosql/geospatial/GeometrySerde.h
bikramSingh91: > this would be used simultaneously among threads. will that usage be thread safe?

@wraymo any thoughts on this?

wraymo: Based on my understanding, `GeometryFactory` is generally thread-safe for creating geometries. However, there is a potential issue related to reference counting. Each time a geometry is created, the factory increments a reference count, and when a geometry is destroyed, it decrements the count. These operations are not protected by locks, which could lead to undefined behavior in multithreaded environments if auto-destroy is enabled.

However, auto-destroy is disabled by default and in our case, we use `GeometryFactory` for the entire lifetime of the program, so this is unlikely to cause problems.

wraymo: But if we want to eliminate potential issues entirely, we can use a thread-local instance. That's my initial implementation.

bikramSingh91: @wraymo thanks for the explanation. there seems to be concerns around thread safety for sure and there are no explicit guarantees from the library. My suggestion would be to move back to using a thread_local variable.

To address @jagill 's concerns about `Static Initialization Order Fiasco` we can create an accessor method instead of directly accessing the factory object, something like:

```
geos::geom::GeometryFactory* getGeometryFactory() {
  thread_local static geos::geom::GeometryFactory::Ptr geometryFactory =
      geos::geom::GeometryFactory::create();
  return geometryFactory.get();
}
```

velox/functions/prestosql/geospatial/GeometrySerde.cpp
```diff
@@ -14,28 +14,242 @@
...
+    return geometryFactory_->createMultiPolygon(std::move(polygons));
+  }
+
+  if (polygons.size() != 1) {
+    VELOX_FAIL("Expected exactly one polygon, but found multiple.");
+  }
+  return std::move(polygons[0]);
+}
+
+std::unique_ptr<geos::geom::Geometry> GeometryDeserializer::readEnvelope(
```
jagill, line 210: We should have a function that returns a `geos::geom::Envelope`.  It could be this one, or a separate one.

wraymo:
@jagill @bikramSingh91 I've updated this PR to address the issues you pointed out and added more test cases. I had to remove two tests because for the newly upgraded GEOS version they triggered exceptions during geometry creation—before `ST_isValid` was even called.

I also reverted to using a thread-local `GeometryFactory` and introduced a `getGeometryFactory()` helper. While we weren’t hitting any static initialization order issues in my opinion, this change reduces the overhead of repeatedly creating factory instances.

Once this PR is merged, I’ll follow up with a cleanup PR to replace usages like
`geos::geom::GeometryFactory::Ptr factory = geos::geom::GeometryFactory::create();`
with
`auto factory = geospatial::getGeometryFactory();`.

bikramSingh91:
@wraymo do you mind rebasing on latest, there are some issues with the current branch that is preventing me to merge internally

wraymo:
@bikramSingh91 updated

PR #12776, opened 2025-03-24
fix: Move only one region to prefetch if the regions overlap

Yuhta:
Summary:
When same stream is fetched multiple times (e.g. shared dictionary),
currently if it falls in the coalesce gap, we move all of them to prefetch, but
only one of them can be read without issue, others would get EOF error.  Fix
this by moving only one of them.  We also pick the largest one if the sizes are
different, although such case probably would not show up in real world data.

Differential Revision: D71711902

### DIFF
--- a/velox/dwio/common/BufferedInput.h
+++ b/velox/dwio/common/BufferedInput.h
@@ -199,8 +199,13 @@ class BufferedInput {
         }
         while (i < noPrefetch.size() &&
                getRegionEnd(noPrefetch[i]) <= coalesceEnd) {
-          prefetch[k++] = noPrefetch[i++];
-          ++numMoved;
+          if (getRegionOffset(noPrefetch[i]) >= coalesceStart) {
+            coalesceStart = getRegionEnd(noPrefetch[i]);
+            prefetch[k++] = noPrefetch[i++];
+            ++numMoved;
+          } else {
+            noPrefetch[l++] = noPrefetch[i++];
+          }
         }
         prefetch[k++] = oldPrefetch[j++];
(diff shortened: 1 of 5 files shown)


velox/dwio/common/CachedBufferedInput.cpp
```diff
@@ -151,8 +151,9 @@ uint64_t getOffset(const CacheRequest& request) {
 }
 
 template <bool kSsd>
-std::pair<uint64_t, uint64_t> toRegion(const CacheRequest& request) {
-  return std::make_pair(getOffset<kSsd>(request), request.size);
+std::pair<uint64_t, int64_t> toRegion(const CacheRequest& request) {
+  return std::make_pair(
```
xiaoxmeng, line 155: Why we use minus here? Shall we comment?

velox/dwio/common/BufferedInput.h
```diff
@@ -199,8 +199,13 @@ class BufferedInput {
         }
         while (i < noPrefetch.size() &&
                getRegionEnd(noPrefetch[i]) <= coalesceEnd) {
-          prefetch[k++] = noPrefetch[i++];
-          ++numMoved;
+          if (getRegionOffset(noPrefetch[i]) >= coalesceStart) {
```
xiaoxmeng, line 202: (getRegionOffset(noPrefetch[i]) >= coalesceStart) is always false given the check at l197?

xiaoxmeng (approved):
@Yuhta LGTM. Thanks for offline explanation!

PR #12777, opened 2025-03-24
fix(parquet): SkippedStrides runtime stats

majetideepak:
skippedStrides must only count the RowGroups that were skipped in the split range.
Currently, this value is calculated using the RowGroup count in the entire file.

### DIFF
--- a/velox/dwio/parquet/reader/ParquetReader.cpp
+++ b/velox/dwio/parquet/reader/ParquetReader.cpp
@@ -1002,11 +1002,16 @@ class ParquetRowReader::Impl {
       if (rowGroupInRange && !isExcluded && !isEmpty) {
         rowGroupIds_.push_back(i);
         firstRowOfRowGroup_.push_back(rowNumber);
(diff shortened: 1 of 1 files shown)


xiaoxmeng (approved):

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1107,6 +1112,7 @@ class ParquetRowReader::Impl {
   const thrift::RowGroup* currentRowGroupPtr_{nullptr};
   uint64_t rowsInCurrentRowGroup_;
   uint64_t currentRowInGroup_;
+  uint32_t skippedStrides_{0};
```
rui-mo, line 1115: nit: shall we use 64-bit type as below?
https://github.com/facebookincubator/velox/blob/0b8c32298a879c748d5df43ca1407108ab7a17e5/velox/dwio/common/Statistics.h#L549

rui-mo (commented):
Thanks!

PR #12779, opened 2025-03-24
feat: Impl split_to_multimap Null Args

duxiao1212:
Summary:
In Presto, when either entryDelimiter or keyValueDelimiter is null, the SPLIT_TO_MULTIMAP function returns null instead of throwing an exception. 
We should maintain the consistency between presto and prestissimo

{F1976364311}

Differential Revision: D71688028


duxiao1212:
> The build failure seems related?

Yeah, I just realized I used C++20 syntax.

zacw7 (approved):
LGTM. Thanks!

PR #10359, opened 2024-07-01
Add SparkSql function to_pretty_string

jinchengchenghh:
Spark implement: https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/ToPrettyString.scala

This is an internal spark function.

PR #12780, opened 2025-03-24
feat: Add Spark to_pretty_string function

ArnavBalyan:
- Add support for to_pretty_string used by Spark.
 - Builds upon @jinchengchenghh's contribution here https://github.com/facebookincubator/velox/pull/10359

### DIFF
--- a/velox/docs/functions/spark/string.rst
+++ b/velox/docs/functions/spark/string.rst
@@ -3,9 +3,9 @@ String Functions
 ====================================
 
 .. note::
-    
+
     Unless specified otherwise, all functions return NULL if at least one of the arguments is NULL.
-    
+
     These functions assume that input strings contain valid UTF-8 encoded Unicode code points.
     The behavior is undefined if they are not.
 
@@ -16,7 +16,7 @@ String Functions
 .. spark:function:: bit_length(string/binary) -> integer
 
     Returns the bit length for the specified string column. ::
-        
+
         SELECT bit_length('123'); -- 24
 
 .. spark:function:: chr(n) -> varchar
@@ -47,7 +47,7 @@ String Functions
 .. spark:function:: contains(left, right) -> boolean
 
     Returns true if 'right' is found in 'left'. Otherwise, returns false. ::
-        
+
         SELECT contains('Spark SQL', 'Spark'); -- true
         SELECT contains('Spark SQL', 'SPARK'); -- false
         SELECT contains('Spark SQL', null); -- NULL
@@ -167,7 +167,7 @@ String Functions
         SELECT lower('SparkSql'); -- sparksql
 
 .. spark:function:: lpad(string, len, pad) -> string
-    
+
     Returns ``string``, left-padded with pad to a length of ``len``. If ``string`` is
     longer than ``len``, the return value is shortened to ``len`` characters or bytes.
     If ``pad`` is not specified, ``string`` will be padded to the left with space characters
@@ -238,7 +238,7 @@ String Functions
 
 .. spark:function:: repeat(input, n) -> varchar
 
-    Returns the string which repeats ``input`` ``n`` times. 
+    Returns the string which repeats ``input`` ``n`` times.
     Result size must be less than or equal to 1MB.
     If ``n`` is less than or equal to 0, empty string is returned. ::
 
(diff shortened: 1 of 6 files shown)


velox/docs/functions/spark/string.rst
```diff
@@ -390,6 +390,15 @@ String Functions
         SELECT substring_index('aaaaa', 'aa', 5); -- "aaaaa"
         SELECT substring_index('aaaaa', 'aa', -5); -- "aaaaa"
 
+.. spark:function:: to_pretty_string(x) -> varchar
```
rui-mo, line 393: The document doesn't appear to render properly. Could you take a look?

Screenshot 2025-03-25 at 15 24 03

velox/functions/sparksql/ToPrettyString.h
```diff
@@ -0,0 +1,206 @@
...
+        result.setNoCopy(*input);
+        return Status::OK();
+      }
+      if constexpr (std::is_same_v<TInput, int32_t>) {
+        if (inputType_->isDate()) {
+          try {
+            auto output = DATE()->toString(*input);
+            result.append(output);
+          } catch (const std::exception& e) {
+            return Status::Invalid(e.what());
```
rui-mo, line 59: It appears the date-to-string cast throws for runtime error,  but here we return "NULL" for both user error and runtime error. Which is aligned with Spark?
https://github.com/facebookincubator/velox/blob/ccd765d3d9e70ddc4cb222b00433b0922915b6cc/velox/expression/CastExpr.cpp#L183-L185

velox/functions/sparksql/ToPrettyString.h
```diff
@@ -0,0 +1,206 @@
...
+      Timestamp inputValue(*timestamp);
+      try {
+        if (options_.timeZone) {
+          inputValue.toTimezone(*(options_.timeZone));
+        }
+        result.reserve(timestampRowSize_);
+        const auto stringView =
+            Timestamp::tsToStringView(inputValue, options_, result.data());
+        result.resize(stringView.size());
+      } catch (const std::exception& e) {
```
rui-mo, line 143: `CastExpr::applyToSelectedNoThrowLocal` adopts the below logic for NULL.

https://github.com/facebookincubator/velox/blob/ccd765d3d9e70ddc4cb222b00433b0922915b6cc/velox/expression/CastExpr-inl.h#L206-L214

rui-mo (commented):
Thanks.

rui-mo:
Would you run the Spark expression fuzzer test for 2 minutes to make sure it works? You could specify `--only to_pretty_string` to test this function only.

The command for reference: https://github.com/facebookincubator/velox/blob/main/.github/workflows/scheduled.yml#L587-L596

rui-mo:
@jinchengchenghh Do you have further comment? Thanks.

jinchengchenghh:
LGTM @rui-mo 
Please fix the code format @ArnavBalyan

velox/functions/sparksql/ToPrettyString.h
```diff
@@ -0,0 +1,208 @@
...
+    if (input) {
+      // One byte spares 2 char, and with the spaces and the boxes.
+      // Byte size: 2 * input->size(), spaces size: input->size() - 1, boxes
+      // size: 2, its sum is 1 + 3 * input->size().
+      result.resize(1 + 3 * input->size());
+      char* const startPosition = result.data();
+      char* pos = startPosition;
+      *pos++ = '[';
+      for (auto i = 0; i < input->size(); i++) {
+        int count = std::sprintf(pos, "%02X", input->data()[i]);
```
assignUser, line 102: Could use fmt::format_to, sprintf throws a warning on macos.

assignUser (changes requested):
Please make sure CI passes, currently a test is failing. I'd recommend rebasing/merging main too.

velox/functions/sparksql/ToPrettyString.h
assignUser: I don't see a change?

velox/functions/sparksql/ToPrettyString.h
jinchengchenghh: Do you mean match the VeloxException, then catch std::exception? @rui-mo This is not addressed. @ArnavBalyan

rui-mo:
@ArnavBalyan You could apply the below diff to fix the format issue.

https://github.com/facebookincubator/velox/actions/runs/14532602668/attempts/2#summary-40925598950

assignUser (approved):
Thanks!

velox/functions/sparksql/ToPrettyString.h
rui-mo: Current implementation catches all the std::exception and converts them as user error, while in the cast, exception is thrown if not a user error, and set null result if it is. If the only differences with cast are the listed ones, we might align with cast here.

rui-mo:
@ArnavBalyan It seems your updates are not pushed in, including https://github.com/facebookincubator/velox/pull/12780#discussion_r2012367324, could you confirm?

PR #12785, opened 2025-03-24
fix: Change sanity check in TDigest::mergeDeserialized()

spershin:
Summary:
TDigest::mergeDeserialized() has a strict check on the sum to be the same as the computed sum (within small epsilon).
That fails some real world queries, whith large sums and small, but not as epsilon, differences).
Example 6178557350588.282 vs 6178557350588.292 resulted in 0.009765625 which is larger than 0.001.

This change adds relative error algorithm that somewhat offsets the epsilon based on the magnitude of the sums.
Presto uses the same algorithm in the Verifier to compare floating point columns.

Differential Revision: D71770400

### DIFF
--- a/velox/functions/lib/TDigest.h
+++ b/velox/functions/lib/TDigest.h
@@ -109,10 +109,11 @@ class TDigest {
     return compression_;
   }
 
- private:
   static constexpr int8_t kSerializationVersion = 1;
   static constexpr double kEpsilon = 1e-3;
+  static constexpr double kRelativeErrorEpsilon = 1e-4;
 
+ private:
   void mergeNewValues(std::vector<int16_t>& positions, double compression);
 
   void merge(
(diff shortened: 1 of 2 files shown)


velox/functions/lib/TDigest.h
```diff
@@ -460,22 +461,48 @@ void TDigest<A>::mergeDeserialized(
...
+    // Sanity check for the sum.
     if (version >= 1) {
       double actualSum = 0;
-      for (int i = 0; i < numNew; ++i) {
+      for (auto i = 0; i < numNew; ++i) {
         actualSum += weights[i] * means[i];
       }
-      VELOX_CHECK_LT(std::abs(actualSum - sum), kEpsilon);
+
+      // Simple check for sums matching first.
```
Yuhta, line 481: Can we extract this to a separate method?

Yuhta (approved):

PR #12788, opened 2025-03-24
feat(fuzzer): Add input generation for varchar and json cast to generic type

peterenescu:
Summary:
Generate valid input for `cast(V as T)` where `V` is either varchar or json and T can be any other type such as boolean, int, real, timestamp, complex, etc.

We can accomplish so by implementing a new class `CastVarcharAndJsonInputGenerator` that generates the aforementioned valid input and invoking it from Expression Fuzzer.

This will increase our test coverage to allow for json and varchar casting.

Differential Revision: D71510669

### DIFF
--- a/velox/common/fuzzer/ConstrainedGenerators.cpp
+++ b/velox/common/fuzzer/ConstrainedGenerators.cpp
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-#include "velox/common/fuzzer/ConstrainedGenerators.h"
-
 #include <boost/random/uniform_int_distribution.hpp>
+#include <cfloat>
+
+#include "velox/common/fuzzer/ConstrainedGenerators.h"
 #include "velox/common/fuzzer/Utils.h"
 #include "velox/functions/lib/TDigest.h"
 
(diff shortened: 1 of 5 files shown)


velox/common/fuzzer/ConstrainedGenerators.cpp
```diff
@@ -286,4 +285,90 @@ getRandomInputGenerator(size_t seed, const TypePtr& type, double nullRatio) {
...
+          std::to_string(rand<int128_t>(rng_, INT64_MIN, INT64_MAX)));
+    case TypeKind::REAL:
+      return variant(std::to_string(rand<float>(rng_, FLT_MIN, FLT_MAX)));
+    case TypeKind::DOUBLE:
+      return variant(std::to_string(rand<double>(rng_, FLT_MIN, FLT_MAX)));
+    case TypeKind::TIMESTAMP:
+      return variant(std::to_string(
+          randTimestamp(rng_, FuzzerTimestampPrecision::kMicroSeconds)));
+    case TypeKind::VARCHAR:
+      return generateRandomGarbageString();
```
kagamiori, line 349: Just call randString() here. At the end of CastVarcharInputGenerator::generate(), we can call makeRandomStrVariation with small probabilities. This would be better because even for the valid input strings, we want sometimes to do a small twist to check whether Velox handle the slightly-wrong input properly.

velox/expression/fuzzer/ArgValuesGenerators.cpp
```diff
@@ -183,4 +183,55 @@ std::vector<core::TypedExprPtr> PhoneNumberArgValuesGenerator::generate(
...
+    const auto representedType = facebook::velox::randType(rng, 0);
+    const auto seed = rand<uint32_t>(rng);
+    const auto nullRatio = options.nullRatio;
+
+    // Populate state.customInputGenerators_ and inputExpressions for inputs
+    // that require custom input generators. Casting to complex type should
+    // require JSON as input. Primitive conversion should require VARCHAR.
+    if (signature.returnType->kind() == TypeKind::ROW ||
+        signature.returnType->kind() == TypeKind::ARRAY ||
+        signature.returnType->kind() == TypeKind::MAP) {
```
kagamiori, line 213: The choice of using JsonInputGenerator or CastVarcharInputGenerator should depend on whether signature.args[0] is json or varchar type. Json type can be identified through the isJsonType() method.

velox/expression/fuzzer/ArgValuesGenerators.cpp
```diff
@@ -183,4 +183,55 @@ std::vector<core::TypedExprPtr> PhoneNumberArgValuesGenerator::generate(
...
+      state.customInputGenerators_.emplace_back(
+          std::make_shared<fuzzer::JsonInputGenerator>(
+              seed,
+              signature.args[0],
+              nullRatio,
+              fuzzer::getRandomInputGenerator(seed, representedType, nullRatio),
```
kagamiori, line 219: This representedType should also match signature.returnType.

velox/expression/fuzzer/ArgValuesGenerators.cpp
```diff
@@ -183,4 +184,53 @@ std::vector<core::TypedExprPtr> PhoneNumberArgValuesGenerator::generate(
...
+          std::make_shared<fuzzer::CastVarcharInputGenerator>(
+              seed, signature.args[0], nullRatio, signature.returnType));
+    }
+
+    VELOX_CHECK_GE(signature.args.size(), 1);
+    inputExpressions.emplace_back(std::make_shared<core::FieldAccessTypedExpr>(
+        signature.args[0],
+        signature.args.size() == 1
+            ? state.inputRowNames_.back()
+            : state.inputRowNames_[state.inputRowNames_.size() - 2]));
```
kagamiori, line 231: Cast always have exactly 1 argument. So we don't need the else branch here.

kagamiori (commented):
Hi @peterenescu, thank you for the update. I left a few comments.

kgpai (approved):

PR #12794, opened 2025-03-25
feat: Add std-like wrappers around breeze atomics

dreveman:
This can be used to replace current libcudacxx usage and as a result make the code more portable and faster to compile.

### DIFF
--- a/velox/experimental/wave/common/Atomic.cuh
+++ b/velox/experimental/wave/common/Atomic.cuh
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


velox/experimental/wave/common/Atomic.cuh
```diff
@@ -0,0 +1,117 @@
...
+
+#include <breeze/platforms/platform.h>
+#include <breeze/utils/types.h>
+#include <breeze/platforms/cuda.cuh>
+#include "velox/experimental/wave/common/BitUtil.cuh"
+#include "velox/experimental/wave/common/CompilerDefines.h"
+
+namespace facebook::velox::wave {
+
+enum class MemoryScope { kDevice };
```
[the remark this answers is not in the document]
dreveman: Yes, we can add that if needed. I added this enum class here so it's easy to extend it for that in the future and it also matches the libcudacxx api better. Didn't add it as part of this PR as I don't see a use of anything but device scope atomics in current wave code.

velox/experimental/wave/common/Atomic.cuh
```diff
@@ -0,0 +1,159 @@
...
+
+enum class MemoryOrder {
```
Yuhta, line 29: Do we have full barrier (something similar to MFENCE)?

dreveman: Not as part of this API at this time. You can use `__threadfence*` family of CUDA builtins directly but I'd be happy to add it to the breeze abstraction as a follow up if that makes sense.

PR #12795, opened 2025-03-25
build: Remove usage of cstdint in CUDA code when using RTC

dreveman:
Move these types to a StdInt.h header where cstdint types are only used when not using RTC.


dreveman:
@Yuhta I marked this as a draft as https://github.com/facebookincubator/velox/pull/12823 it probably a better solution for now. If we're interested in removing usage of jitify then something like this might be needed.

PR #12800, opened 2025-03-25
build: Add xxd hex dump util to dependencies

pramodsatya:
Adds dependencies `vim-common` to `setup-centos` and `xxd` to `setup-ubuntu`.
This lets us use the `xxd` command to create hex dump from file 
`velox/tpcds/gen/include/tpcds.idx` in https://github.com/facebookincubator/velox/pull/12714.
Required for https://github.com/facebookincubator/velox/pull/12714.

### DIFF
--- a/scripts/setup-centos9.sh
+++ b/scripts/setup-centos9.sh
@@ -80,7 +80,7 @@ function install_velox_deps_from_dnf {
   dnf_install libevent-devel \
     openssl-devel re2-devel libzstd-devel lz4-devel double-conversion-devel \
     libdwarf-devel elfutils-libelf-devel curl-devel libicu-devel bison flex \
-    libsodium-devel zlib-devel
+    libsodium-devel zlib-devel vim-common
 
   # install sphinx for doc gen
   pip install sphinx sphinx-tabs breathe sphinx_rtd_theme
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -150,7 +150,8 @@ function install_velox_deps_from_apt {
     bison \
     flex \
     libfl-dev \
-    tzdata
+    tzdata \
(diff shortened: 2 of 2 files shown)


czentgr (commented):
On macOS we don't need to do anything?

pramodsatya:
> On macOS we don't need to do anything?

`xxd` is supported on standard macOS by default so no additional changes are needed.

assignUser:
See:  https://github.com/facebookincubator/velox/pull/12714/files#r2017800589

majetideepak:
@czentgr and @assignUser, any other comments?

czentgr (approved):
Thanks.

assignUser:
I still think that we should check in the text version of the file, while the file is bigger it'll also get compressed by git so we  are kind of doing git's job with xxd but worse as git is bad with binary files. Also given the small space binary files take up in the PR view this could be a nice space to sneak in some hidden malicious payload, maybe a bit far fetched but in the time of xz....

But if @majetideepak and @czentgr think we should go with the binary+xxd version I'll relent.

majetideepak:
@assignUser are you suggesting we also check-in tpcds_idx.hpp instead of generating it in this PR?https://github.com/facebookincubator/velox/pull/12714
This generated text file is around 4.1MB.

assignUser:
Yes, but git is good a text compression and bad with binaries. Duckdb also checks on the full file.

majetideepak:
@pramodsatya Can we check-in the generated file, as Jacob suggested?

pramodsatya:
> @pramodsatya Can we check-in the generated file, as Jacob suggested?

Sure, thanks @majetideepak, @assignUser, closing this PR. Could you please help review: https://github.com/facebookincubator/velox/pull/12714 ?

PR #12801, opened 2025-03-25
feat: Add ip_prefix_subnets function for ipprefix

yuandagits:
Summary:
Implement ip_prefix_subnets similar to Java function https://github.com/prestodb/presto/blob/33ff9521dbd25e2b7b1181cf01391cb899857d65/presto-main/src/main/java/com/facebook/presto/operator/scalar/IpPrefixFunctions.java#L303.

This is done by first ensurin the ipprefix and the new length are valid. 

We find the number of new prefix ip counts and new prefix counts, this is used to compute the array of new ipaddresses/prefix based on the new prefix length.

The implementation is the same as Java.

Differential Revision: D71826712


yuandagits:
cc @mohsaka

zacw7 (approved):
looks good. thx!

PR #12802, opened 2025-03-25
refactor(parquet): Remove arrow parquet reader in PageIndexTest

jkhaliqi:
We can remove TestReadTypedColumnIndex test, since Velox does not support ColumnIndex.
Remove unused headers.


velox/dwio/parquet/writer/arrow/tests/PageIndexTest.cpp
```diff
@@ -18,279 +18,11 @@
 
 #include "velox/dwio/parquet/writer/arrow/PageIndex.h"
 
-#include <gtest/gtest.h>
```
majetideepak, line 21: We still need the gtest header.

majetideepak (approved):

PR #12806, opened 2025-03-25
fix(json): Special case certain unicode characters to escape when casting from json

kgpai:
Summary:
When casting from a json to another json complex type , json -> array(json) etc, we by default unescape unicode. However certain unicode sequences in the basic latin (https://www.compart.com/en/unicode/block/U+0000) are not unescaped. This is not a problem generally but if subsequently this json is again casted to another complex type json (i.e you go from json-> array(json) -> Row(x : json)) then simdjson throws requiring certain characters in this sequence to be escaped.

Note we previously added a fix to unescape json's here : D71078371 but it did not cover this edge case.

Differential Revision: D71834496


bikramSingh91 (approved):
Thanks for the fix, this was a pretty tricky edge case

PR #12809, opened 2025-03-25
misc: Decrease row count for presto serializer large buffer test

yuandagits:
We need a buffer that is large enough to have the zstd codec return a chained iobuf; however, this was causing a time out in our parameterized tests.

Decrease the from 800k -> 200k rows


kevins963 (approved):
LGTM

PR #12810, opened 2025-03-26
feat: Add QuantileDigest data structure

kagamiori:
Summary: This diff adds the QuantileDigest data structure to be used for qdigest functions. The serialization and deserialization of this QuantileDigest are compatiable with Presto-Java.

Differential Revision: D71521497

### DIFF
--- a/velox/functions/lib/QuantileDigest.h
+++ b/velox/functions/lib/QuantileDigest.h
@@ -1,0 +1,1193 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/Bits.h>
+
+#include "velox/common/base/CheckedArithmetic.h"
+#include "velox/common/base/Exceptions.h"
+#include "velox/common/base/Portability.h"
+#include "velox/common/memory/HashStringAllocator.h"
+
+namespace facebook::velox::functions {
+
+namespace qdigest {
+
+constexpr double kZeroWeightThreshold = 1.0E-5;
+
+/// Implementation of Q-Digest that matches Presto Java behavior. The
+/// serialization format is same as Java. There is one performance improvement:
+/// mergeSerialized(const char* serialized) allows merging another serialized
+/// digest into the current one without having to deserialize it first. This
+/// optimization allows us to perform the merge with one pass instead of two,
+/// and skip memory allocation for storing the deserialized digest. The Java
+/// behavior is the same as testingMerge.
+///
+/// Java implementation can be found at:
+/// https://github.com/airlift/airlift/blob/57a1bb0182f6336cb03a365be018a73490d9b410/stats/src/main/java/io/airlift/stats/QuantileDigest.java
+template <typename T, typename Allocator = StlAllocator<T>>
+class QuantileDigest {
+  static_assert(
+      std::is_same_v<T, int64_t> || std::is_same_v<T, double> ||
+      std::is_same_v<T, float>);
+
+ public:
+  explicit QuantileDigest(const Allocator& allocator, double maxError);
+
+  QuantileDigest(const Allocator& allocator, const char* serialized);
+
+  void add(T value, double weight);
+
+  void mergeSerialized(const char* other);
+
+  double getCount() const;
+
+  void scale(double scaleFactor);
+
+  void compress();
+
+  /// Estimate the values of the given quantiles and write the results in order
+  /// directly into 'result'.
+  void estimateQuantiles(const std::vector<double>& quantiles, T* result);
+
+  T estimateQuantile(double quantile);
+
+  int64_t serializedByteSize() const;
+
+  int64_t serialize(char* out);
+
+  T getMin();
+
+  T getMax();
+
+  // For testing only. Calling this method with 'other' being constructed from
+  // QuantileDigest(const Allocator& allocator, const char* serialized) is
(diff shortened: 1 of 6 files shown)


velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+          if (rights_[node] != -1) {
+            nodeStructure = static_cast<int8_t>(nodeStructure | 2);
+          }
+          write<int8_t>(nodeStructure, out);
+          write<double>(counts_[node], out);
+          write<int64_t>(values_[node], out);
+
+        } else {
+          auto nodeStructure = static_cast<int8_t>(
+              std::max((levels_[node] == 32 ? 64 : levels_[node]) - 1, 0) << 2);
```
kagamiori, line 827: If levels_[node] == 32, it means the left and right children of node share 0 contiguous common bits from the most significant side, which means their sign bits are different. Since Java store the values always in int64_t, when the sign bit are different, levels_[node] would be 64 - 0 = 64. So we when we see levels_[node] == 32, we adjust it back to 64 to match Presto's serialization.

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+} // namespace qdigest
+
+template <typename T>
+class QuantileDigest {
+  static_assert(
+      std::is_same_v<T, int64_t> || std::is_same_v<T, double> ||
+      std::is_same_v<T, float>);
+
+ public:
+  using U = std::conditional_t<
```
Yuhta, line 40: This type should be private

velox/functions/lib/tests/StatisticalDigestTestBase.h
Yuhta, line 1: Rename this to `QuantileDigestTestBase`

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+
+  int32_t mergeRecursive(int32_t node, QuantileDigest other, int32_t otherNode);
+
+  int32_t copyRecursive(QuantileDigest other, int32_t otherNode);
+
+  U lowerBound(int32_t node) const;
+
+  U upperBound(int32_t node) const;
+
+  double maxError_{0.01};
```
Yuhta, line 138: Remove default value

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+template <typename T>
+int32_t QuantileDigest<T>::createLeaf(U value, double count) {
+  return createNode(value, 0, count);
+}
+
+template <typename T>
+int32_t QuantileDigest<T>::createNode(U value, int8_t level, double count) {
+  auto node = popFree();
+  if (node == -1) {
+    if (nextNode_ == counts_.size()) {
```
Yuhta, line 420: We don't need to do this, just do
```c++
counts_.push_back(count);
levels_.push_back(level);
// ...
```

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+QuantileDigest<T>::U QuantileDigest<T>::getBranchMask(int8_t level) {
+  return static_cast<U>(1) << (level - 1);
+}
+
+template <typename T>
+int32_t QuantileDigest<T>::makeSiblings(int32_t first, int32_t second) {
+  auto firstValue = values_[first];
+  auto secondValue = values_[second];
+  int32_t parentLevel;
+  if constexpr (std::is_same_v<U, int64_t>) {
```
Yuhta, line 466: This can be extracted to its own method.  Basically anything we need to dispatch on `U` extract that to an helper

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+  U lowerBound(int32_t node) const;
+
+  U upperBound(int32_t node) const;
+
+  double maxError_{0.01};
+  double weightedCount_;
+  U min_;
+  U max_;
+  int32_t root_;
+  int32_t nextNode_;
```
Yuhta, line 143: No need for this field, this is just `counts_.size()`

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+      /*alpha*/ sizeof(double) + /*landmarkInSeconds*/ sizeof(int64_t) +
+      /*min*/ sizeof(int64_t) + /*max*/ sizeof(int64_t) +
+      /*nodeCount*/ sizeof(int32_t) +
+      (sizeof(typename decltype(counts_)::value_type) * nodeCount) +
+      (sizeof(typename decltype(levels_)::value_type) * nodeCount) +
+      /*values*/ (sizeof(int64_t) * nodeCount);
+}
+
+template <typename T>
+void write(T value, char*& out) {
```
Yuhta, line 793: `detail` namespace

velox/functions/lib/QuantileDigest.h
Yuhta: This should only happen when `U = int32_t`, let's add that condition and extract this to a separate method

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+              (values_[node] ^ std::numeric_limits<int32_t>::min()) ^
+                  std::numeric_limits<int64_t>::min(),
+              out);
+        }
+
+        return true;
+      },
+      lefts_,
+      rights_);
+  VELOX_CHECK_LE(out - outStart, estimatedInMemorySizeInBytes());
```
Yuhta, line 846: `VELOX_CHECK_EQ`

velox/functions/lib/QuantileDigest.h
```diff
@@ -0,0 +1,850 @@
...
+
+  explicit QuantileDigest(
+      HashStringAllocator* allocator,
+      double maxError = 0.01);
+
+  QuantileDigest(HashStringAllocator* allocator, const char* serialized);
+
+  void add(T value, double weight);
+
+  void merge(const QuantileDigest& other);
```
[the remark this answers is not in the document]
kagamiori: Discussed offline. The serialize() method writes values in post-order, while mergeRecursive() and copyRecursive() access values in pre-order, so we need to find a way to change the logic in mergeRecursive() and copyRecursive(). Before that, I leave merge() unchanged for now.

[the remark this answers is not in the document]
Yuhta: We can remove this now

Yuhta (approved):

PR #12811, opened 2025-03-26
feat: Add decimal support in histogram aggregate

singcha:
Summary: Add decimal support in histogram aggregate. This is top contributor to Sapphire query failing in PROD

Differential Revision: D71860678


amitkdutta (approved):
Thanks @singcha

PR #12812, opened 2025-03-26
fix(vector): use current vector's buffer as the source

YannByron:


YannByron:
@xiaoxmeng can you take a look to check if i understand correctly.

Yuhta (commented):
Looks like a legit bug, can you add a test for it

YannByron:
> Looks like a legit bug, can you add a test for it

Of course.

YannByron:
@Yuhta The test can't be constructed based on existing data types and their implementations. However, if an empty destructor is added to the `StringView` class, like:
```  ~StringView() {}```
 then `StringView` is not `Buffer::is_pod_like_v<StringView>` at all, and this pr can be checked based the test I added in `VectorTest`.

PR #12816, opened 2025-03-26
isAscii: add support for AVX512F SIMD architecture

ManManson:
Use `_mm512_test_epi32_mask` intrinsic for efficient comparison.

Its behavior differs from `_mm256_testz_si256` in the following way: 

`_mm256_testz_si256` returns the value of `ZF` flag, which is 1 if ALL bits of the result mask are zero, and 0 otherwise.

While the `_mm512_test_epi32_mask` returns the bitwise-AND mask itself (containing 1's where bitwise AND result is 1). So, effectively, this can be treated as a `uint16_t` integer, which would be `!= 0` if any bits are set (that is, if there are `0x80` bytes in the source batch to indicate non-ASCII characters).


Yuhta (commented):
The change itself is good, but turning on AVX512 on the code base would likely trigger a lot of errors else where.  Basically the code base is not tuned for AVX512.

ManManson:
Yup. I know. It just happened so that this header got pulled into the translation unit and messed up the build when I enabled `-mavx512f` flag for the `Bridge.cpp` in our private fork.

So, I think an isolated fix is still better than nothing: one less place to fix when the code base gets adapted for AVX512 :)

PR #12818, opened 2025-03-26
test: Replace GTEST_SKIP with DISABLED for cases that are expected or we don't have intention to fix

Yuhta:
Summary:
`GTEST_SKIP` is treated as a code quality issue that the root cause
needs to be fixed.  If the case is unsupported as expected, we should not use
it.

Differential Revision: D71902104


pedroerp (approved):
Thank you

PR #12819, opened 2025-03-26
fix(tpch): Concurrency issues in dbgen

pedroerp:
Summary:
Moving the intialization of buffers out of the mk_* functions
themselves (since they can be called concurrently) to a separate initialization
function called once when the Velox dbgen singleton gets constructed. This
concurrency issue was caught by TSAN.

Differential Revision: D71902408


Yuhta (approved):

xiaoxmeng (approved):
@pedroerp thanks for the fix!

PR #12825, opened 2025-03-26
fix(json): Fix casting of NaN's in json

kgpai:
Summary: We use folly::tryTo to cast json strings . This is quite permissive for certain strings such as "nan"'s etc which are allowed in folly but not in presto java.

Differential Revision: D71936263


amitkdutta (approved):
Thanks @kgpai

PR #12834, opened 2025-03-27
misc: Add 1 to Type::hashKind() to reduce collisions

kevinwilfong:
Summary:
A users noted they were seeing a lot of collisions in Type::hashKind(). Although we don't
provide any guarantees here, the fact that BOOLEAN maps to 0 definitely increases the chances of collisions.

E.g. ROW({...}) where ... is any number of BOOLEANs will always hash to the same value.

Differential Revision: D72005465


kgpai (approved):
LGTM - Please make sure nothing inadvertently depends on these hash values.

PR #12838, opened 2025-03-28
feat: Add function prefix flags in TraceReplayRunner

duanmeng:
Add function prefix flags in `TraceReplayRunner` to support
adding function prefix for trace replay.

Resolve #12815


xiaoxmeng (approved):
@duanmeng thanks for the fix!

PR #12840, opened 2025-03-28
feat: Add ColumnReaderOptions

rui-mo:
This PR adds column reader options created from reader options, and pass it to 
the Parquet and Dwrf column readers. This allows customization of the column 
readers based on the options.

### DIFF
--- a/velox/dwio/common/Options.cpp
+++ b/velox/dwio/common/Options.cpp
@@ -70,4 +70,11 @@ std::string_view toString(FileFormat fmt) {
   }
 }
 
+ColumnReaderOptions makeColumnReaderOptions(const ReaderOptions& options) {
+  ColumnReaderOptions columnReaderOptions;
+  columnReaderOptions.useColumnNamesForColumnMapping_ =
+      options.useColumnNamesForColumnMapping();
+  return columnReaderOptions;
+}
+
 } // namespace facebook::velox::dwio::common
--- a/velox/dwio/common/Options.h
+++ b/velox/dwio/common/Options.h
@@ -681,6 +681,15 @@ struct WriterOptions {
   virtual ~WriterOptions() = default;
 };
 
+// Options for creating a column reader.
+struct ColumnReaderOptions {
+  // Whether to map table field names to file field names using names, not
+  // indices.
+  bool useColumnNamesForColumnMapping_{false};
+};
+
+ColumnReaderOptions makeColumnReaderOptions(const ReaderOptions& options);
+
 } // namespace facebook::velox::dwio::common
 
 template <>
--- a/velox/dwio/dwrf/reader/DwrfReader.cpp
+++ b/velox/dwio/dwrf/reader/DwrfReader.cpp
@@ -44,7 +44,8 @@ class DwrfUnit : public LoadUnit {
       uint32_t stripeIndex,
       std::shared_ptr<dwio::common::ColumnSelector> columnSelector,
       const std::shared_ptr<BitSet>& projectedNodes,
-      RowReaderOptions options)
+      RowReaderOptions options,
+      const dwio::common::ColumnReaderOptions& columnReaderOptions)
       : stripeReaderBase_{stripeReaderBase},
         strideIndexProvider_{strideIndexProvider},
         columnReaderStatistics_{&columnReaderStatistics},
@@ -52,6 +53,7 @@ class DwrfUnit : public LoadUnit {
         columnSelector_{std::move(columnSelector)},
         projectedNodes_{projectedNodes},
         options_{std::move(options)},
+        columnReaderOptions_{columnReaderOptions},
         stripeInfo_{
             stripeReaderBase.getReader().footer().stripes(stripeIndex_)} {}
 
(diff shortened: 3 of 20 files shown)


velox/connectors/hive/SplitReader.cpp
```diff
@@ -333,7 +333,8 @@ void SplitReader::createRowReader(
       hiveConfig_,
       connectorQueryCtx_->sessionProperties(),
       baseRowReaderOpts_);
-  baseRowReader_ = baseReader_->createRowReader(baseRowReaderOpts_);
+  baseRowReader_ =
+      baseReader_->createRowReader(hiveConfig_, baseRowReaderOpts_);
```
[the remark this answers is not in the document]
rui-mo: Got it. I will update this PR following your suggestion, thanks.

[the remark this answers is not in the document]
majetideepak: @rui-mo We already have some fields in readerOptions. May be readerOptions?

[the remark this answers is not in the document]
rui-mo: Yes, the reader options contains the `useColumnNamesForColumnMapping` field. Thanks for the suggestion.

rui-mo:
@majetideepak I updated this PR following your comment. Would you help take another look? The CI failure is being fixed by https://github.com/facebookincubator/velox/pull/12985. Thanks!

velox/dwio/dwrf/reader/DwrfReader.cpp
```diff
@@ -90,6 +92,7 @@ class DwrfUnit : public LoadUnit {
   const std::shared_ptr<dwio::common::ColumnSelector> columnSelector_;
   const std::shared_ptr<BitSet> projectedNodes_;
   const RowReaderOptions options_;
+  const dwio::common::ReaderOptions readerOptions_;
```
majetideepak, line 95: Can this be a reference?

majetideepak (commented):
@rui-mo one comment.

Yuhta:
My feeling on this is it breaks the abstraction between file reader and column readers.  We definitely not want to pass free form `HiveConfig` into low level object like file readers.  We should create expilicit option and flags that is fully understandable by the lower level components, instead of free form key values.

rui-mo:
@Yuhta Thanks for helping review! I was developing https://github.com/facebookincubator/velox/pull/5962 where the `useColumnNamesForColumnMapping` option needs to be passed through readers. To allow customization of the column readers based on the options and make the chain more general, we proposed this PR to pass the readers options rather than a single bool value.

majetideepak:
@Yuhta I also felt creating yet another `Config` object is may be an overkill. But I agree that a clean approach is to use a 
`struct ColumnOptions{};` with strict options. Do you suggest we add that?

majetideepak:
@rui-mo Let's add `struct ColumnOptions{}` which should make the API clean.

rui-mo:
I updated this PR to add ColumnOptions and passed it through column readers. @majetideepak @Yuhta Would you like to take another look? Thanks.

velox/dwio/common/OptionsUtil.h
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/dwio/common/Options.h"
+
+namespace facebook::velox::dwio::common {
+
+void configureColumnOptions(
```
majetideepak, line 23: Let's rename this to `updateColumnOptionsFromReaderOptions`. Can we keep these inside `Options.h/Options.cpp`? We don't need these new files.

velox/dwio/common/Options.h
```diff
@@ -660,6 +660,23 @@ struct WriterOptions {
   virtual ~WriterOptions() = default;
 };
 
+// Options for creating a column reader.
+struct ColumnOptions {
+ public:
```
majetideepak, line 665: Simple POD will do.
```
struct ColumnOptions {
  bool useColumnNamesForColumnMapping_{false};
}
```

majetideepak (approved):

majetideepak:
@Yuhta any thoughts on this?

velox/dwio/common/Options.h
```diff
@@ -660,6 +660,17 @@ struct WriterOptions {
   virtual ~WriterOptions() = default;
 };
 
+// Options for creating a column reader.
+struct ColumnOptions {
```
Yuhta, line 664: Let's call it `ColumnReaderOptions`

velox/dwio/common/Options.cpp
```diff
@@ -70,4 +70,11 @@ std::string_view toString(FileFormat fmt) {
   }
 }
 
+void updateColumnOptionsFromReaderOptions(
```
Yuhta, line 73: Could be just `ColumnReaderOptions fromReaderOptions(const ReaderOptions& options)`

velox/dwio/common/Options.h
rui-mo: Updated, thanks.

velox/dwio/common/Options.cpp
rui-mo: Thanks for the suggestion. Updated.

rui-mo:
@Yuhta I fixed above comments. Would you like to take another look? Thanks for your help.

Yuhta (approved):

PR #12841, opened 2025-03-28
fix: Fix UnsafeRowDeserializer Map value type Row(Array)

jinchengchenghh:


velox/row/UnsafeRowFast.cpp
```diff
@@ -1083,9 +1083,13 @@ RowVectorPtr deserializeRows(
       std::vector<char*> nestedData(numRows);
       std::vector<size_t> nestedOffsets(numRows, 0);
       for (auto row = 0; row < numRows; ++row) {
-        const auto offset =
-            readInt32(data[row] + offsets[row] + sizeof(int32_t));
-        nestedData[row] = data[row] + offset;
+        const auto isTopLevelNull =
+            rawNulls != nullptr && bits::isBitNull(rawNulls, row);
```
rui-mo, line 1087: nit: could drop `!= nullptr`.

rui-mo (commented):
Would you perhaps expand the PR description to provide more information about the issue to be resolved? Also considering revising the PR title, perhaps like `Fix deserializeRows when the nested complex data contains null`.

velox/row/UnsafeRowFast.cpp
jinchengchenghh: The rawNulls might be null, we can not check `bits::isBitNull` in that case.

rui-mo: I assume `rawNulls != nullptr && bits::isBitNull(rawNulls, row)` is equal to `rawNulls && bits::isBitNull(rawNulls, row)`, no?

rui-mo (approved):

velox/row/UnsafeRowFast.cpp
jinchengchenghh: updated, thanks!

jinchengchenghh:
Hi @mbasmanova Can you help review this PR? Thanks!

velox/row/UnsafeRowFast.cpp
```diff
@@ -1083,9 +1083,12 @@ RowVectorPtr deserializeRows(
       std::vector<char*> nestedData(numRows);
       std::vector<size_t> nestedOffsets(numRows, 0);
       for (auto row = 0; row < numRows; ++row) {
-        const auto offset =
-            readInt32(data[row] + offsets[row] + sizeof(int32_t));
-        nestedData[row] = data[row] + offset;
+        const auto isTopLevelNull = rawNulls && bits::isBitNull(rawNulls, row);
```
mbasmanova, line 1086: Would you add a test case that specifically targets this scenario? The fuzzer test is useful, but it doesn't guarantee that this path is hit on every run.

velox/row/tests/UnsafeRowFuzzTest.cpp
```diff
@@ -235,5 +293,23 @@ TEST_F(UnsafeRowFuzzTests, fast) {
       });
 }
 
+TEST_F(UnsafeRowFuzzTests, nestedMaps) {
```
mbasmanova, line 296: I assume this is not a fuzz test, is it? Why is it part of UnsafeRowFuzzTests?

jinchengchenghh: We don't have UnsafeRowTest, do I need to add a new File UnsafeRowTest?

jinchengchenghh: I would suggest to rename the `UnsafeRowFuzzTest` to `UnsafeRowTest`.

velox/row/tests/UnsafeRowTest.cpp
```diff
@@ -29,13 +29,13 @@ namespace {
 
 using namespace facebook::velox::test;
 
-class UnsafeRowFuzzTests : public ::testing::Test {
+class UnsafeRowTests : public ::testing::Test, public VectorTestBase {
```
mbasmanova, line 32: UnsafeRowTests -> UnsafeRowTest (to match file name and other tests)

velox/row/tests/UnsafeRowTest.cpp
```diff
@@ -235,5 +293,23 @@ TEST_F(UnsafeRowFuzzTests, fast) {
...
+           {{4, 5}},
+           {{1}},
+           std::nullopt,
+           {{3, 2, 4, 5}},
+           {{6}},
+       }),
+       makeNullableFlatVector<int32_t>({1, 2, 3, std::nullopt, 5, 6})},
+      [](auto row) { return row == 3; });
+  auto keys = makeFlatVector<int32_t>({1, 2, 3, 4, 5, 6});
+  auto values = makeMapVector({0, 2, 2, 3, 4, 5}, keys, innerMaps, {1});
```
mbasmanova, line 309: Wondering if using makeMapVectorFromJson would make the test more readable.

velox/row/tests/UnsafeRowTest.cpp
```diff
@@ -94,6 +94,63 @@ class UnsafeRowFuzzTests : public ::testing::Test {
     }
   }
 
+  void testRoundTrip(const RowVectorPtr& data) {
```
mbasmanova, line 97: Is there copy-paste from the existing code in this file? If so, would it be possible to refactor?

CC: @rui-mo @pedroerp

velox/row/tests/UnsafeRowTest.cpp
jinchengchenghh: No, makeMapVectorFromJson can only generate the map vector with primitive key and value data type. The value type is ROW in this test.

velox/row/tests/UnsafeRowTest.cpp
```diff
@@ -67,43 +56,92 @@ class UnsafeRowFuzzTests : public ::testing::Test {
...
+
+      VELOX_CHECK_EQ(size, row.rowSize(i), "Row {}: {}", i, data->toString(i));
+    }
+
+    VELOX_CHECK_EQ(offset, totalSize);
+
+    if constexpr (std::is_same_v<T, std::optional<std::string_view>>) {
+      // Deserialize previous bytes back to row vector
+      VectorPtr outputVector =
+          UnsafeRowDeserializer::deserialize(serialized, rowType, pool_.get());
```
mbasmanova, line 131: Do we still need UnsafeRowDeserializer? Can we replace it with UnsafeRowFast?

mbasmanova (approved):
Thank you for iterating.

kKPulla:
@jinchengchenghh looks like there are some conflict that need to be resolved from the tip. Can you resolve and update the branch for me to finish the merge?

jinchengchenghh:
Updated, and CI pass. Can you help import it? Thanks! @kKPulla

kKPulla:
> Updated, and CI pass. Can you help import it? Thanks! @kKPulla

@jinchengchenghh It still does not let me import as it complains about conflict on the tip of the github main branch. Can you rebase this on the tip and update and I will get this imported today (Sorry, was busy with VeloxCon for the last few days)

jinchengchenghh:
The failure is not related to this PR, create a new issue to track https://github.com/facebookincubator/velox/issues/13067

PR #12846, opened 2025-03-29
fix(parquet): Fix inserting decimal type values precision and scale

jkhaliqi:
Prestissimo decimal type is incorrect when inserting values in Parquet. Decimal precision, scale are written incorrectly to Parquet. Need to change presto velox parquet-tools inspect output on parquet tools created by with decimal to be similar to presto java.

To Reproduce
```
create table decimaltest(col0 DECIMAL(10, 2));
CREATE TABLE
insert into decimaltest values(1200.00);
INSERT: 1 row

parquet meta file.parquet
```
Before the PR the output for prestissimo would be showing the `col0` type as `DECIMAL(6,2)` instead of `DECIMAL(10,2)`.
```
Created by: parquet-mr version 2.6.0 (build parquet-cpp-velox)
Properties: (none)
Schema:
message schema {
  optional int32 col0 (DECIMAL(6,2));
}
```
Expected behavior
Java writer does not have this issue. It uses the right precision, scale. AND with this PR we are getting the same output for the DECIMAL type and scale.

```
Created by: parquet-mr version 1.13.1 (build db4183109d5b734ec5930d870cdae161e408ddba)
Properties: (none)
Schema:
message hive_schema {
  optional fixed_len_byte_array(5) col0 (DECIMAL(10,2));
}
```


velox/dwio/parquet/writer/Writer.h
```diff
@@ -109,6 +110,9 @@ struct WriterOptions : public dwio::common::WriterOptions {
   std::optional<std::string> parquetWriteTimestampTimeZone;
   bool writeInt96AsTimestamp = false;
   std::optional<bool> useParquetDataPageV2;
+  std::optional<bool> storeDecimalAsInteger;
```
majetideepak, line 113: We don't have to change the physical type and the ParquetVersion.
Java can use FLBA, Velox can use integers (more efficient).
We only need to ensure the Decimal logical type is the same as specified by the client (Presto in this case).

velox/exec/TableWriter.cpp
```diff
@@ -70,15 +70,17 @@ TableWriter::TableWriter(
 void TableWriter::setTypeMappings(
     const std::shared_ptr<const core::TableWriteNode>& tableWriteNode) {
   auto outputNames = tableWriteNode->columnNames();
-  auto outputTypes = tableWriteNode->columns()->children();
+  auto tableTypes = tableWriteNode->insertTableHandle()
```
majetideepak, line 73: How are `tableWriteNode->columns()` different from those that are in the `insertTableHandle`? Can you explore how Prestissimo sets the column types in these two?

velox/dwio/parquet/writer/arrow/Properties.h
```diff
@@ -323,10 +323,10 @@ class PARQUET_EXPORT WriterProperties {
           write_batch_size_(DEFAULT_WRITE_BATCH_SIZE),
           max_row_group_length_(DEFAULT_MAX_ROW_GROUP_LENGTH),
           pagesize_(kDefaultDataPageSize),
-          version_(ParquetVersion::PARQUET_2_6),
+          version_(ParquetVersion::PARQUET_1_0),
           data_page_version_(ParquetDataPageVersion::V1),
           created_by_(
-              std::string("parquet-mr version 2.6.0 (build ") +
+              std::string("parquet-mr version 1.13.1 (build ") +
```
majetideepak, line 329: This is incorrect. This should be `parquet-cpp-velox version 0.0.0`. We can add the version number once when we have Velox versioning.
We should also allow users to be able to specify a string for this. See https://github.com/apache/arrow/blob/c56ec12a2f38ed6fce2d87ea65345b6edf4234f5/cpp/src/parquet/properties.h#L751

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -365,9 +373,11 @@ dwio::common::StripeProgress getStripeProgress(
  * This method assumes each input `ColumnarBatch` have same schema.
  */
 void Writer::write(const VectorPtr& data) {
-  VELOX_USER_CHECK(
-      data->type()->equivalent(*schema_),
-      "The file schema type should be equal with the input rowvector type.");
+  // make it check this but for decimal type it does not have to be the same...
```
majetideepak, line 376: We should check that the input decimal data type fits in the schema decimal type.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -216,6 +219,11 @@ std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
         arrowMapType->item_field(), *mapType.valueType());
     return newField->WithType(
         ::arrow::map(newKeyField->type(), newValueField->type()));
+  } else if (name != "" && type.isDecimal()) {
+    auto newField = field->WithName(name);
+    auto precisionScale = getDecimalPrecisionScale(type);
+    return newField->WithType(
+        ::arrow::decimal(precisionScale.first, precisionScale.second));
```
majetideepak, line 226: Why handle decimals specially here? How are they handled without this change?

velox/connectors/Connector.h
```diff
@@ -155,6 +155,8 @@ class ConnectorInsertTableHandle : public ISerializable {
 
   virtual std::string toString() const = 0;
 
+  virtual const std::vector<TypePtr> getHiveColumnHandleDataType() const = 0;
```
czentgr, line 158: We need to leave "hive" out of the name here. This is a generic connector handle.

velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
```diff
@@ -707,10 +710,10 @@ TEST_F(E2EFilterTest, combineRowGroup) {
   EXPECT_EQ(parquetReader.numberOfRows(), 5);
 }
 
-TEST_F(E2EFilterTest, writeDecimalAsInteger) {
+TEST_F(E2EFilterTest, writeDecimalAsFLBA) {
   auto rowVector = makeRowVector(
-      {makeFlatVector<int64_t>({1, 2}, DECIMAL(8, 2)),
-       makeFlatVector<int64_t>({1, 2}, DECIMAL(10, 2)),
+      {makeFlatVector<double>({1.23, 203.43}, DECIMAL(8, 2)),
```
czentgr, line 715: Do we need an error test where the literal does not fit into the column? Is that stopped in the writer?

velox/dwio/parquet/writer/Writer.cpp
czentgr: This is so we make sure that the parquet type is set from the column type rather than inferred from the field data?

velox/dwio/parquet/writer/arrow/Properties.h
```diff
@@ -323,10 +323,10 @@ class PARQUET_EXPORT WriterProperties {
           write_batch_size_(DEFAULT_WRITE_BATCH_SIZE),
           max_row_group_length_(DEFAULT_MAX_ROW_GROUP_LENGTH),
           pagesize_(kDefaultDataPageSize),
-          version_(ParquetVersion::PARQUET_2_6),
+          version_(ParquetVersion::PARQUET_1_0),
```
czentgr, line 326: Are we mixing changes for separate PRs into one here?

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: It is handled here because the arrowSchema is `auto arrowSchema = ::arrow::ImportSchema(&schema).ValueOrDie();` is passed in as the schema and here the decimal value for the fields is whatever the input is. So if it is 1200.00 it will be 6,2 and if it is 120.00 it will be 5,2. Instead of this since we want it to match whatever the table is we now use the `*schema_->childAt(i)` so if the table is 10,2 decimal it will always be that value and not based on the input.
Without the change it will go down to the else if (name != ""), and the schema decimal will be based on whatever  that input is.

velox/exec/TableWriter.cpp
jkhaliqi: `tableWriteNode->columns()` is set from `toRowType(node->columns, typeParser_)` and its based on the input so if we call it with 120.00 it will be decimal(5,2) and 1200.00 it will be decimal (6,2).  insertTableHandle comes from `auto insertTableHandle = std::make_shared<core::InsertTableHandle>(
      connectorId, connectorInsertHandle);` where `connectorInsertHandle = std::shared_ptr(std::move(veloxHandle));` and it's got both the `dataType_` and `hiveType_` which correspond with how the table was created with whatever type so if table is created with DECIMAL(10,2) and even if we insert 1200.00 it will still be DECIMAL(10,2) instead of the inserted DECIMAL(6,2).

velox/dwio/parquet/writer/arrow/Properties.h
jkhaliqi: Yeah I believe the issue mentions `The Created by: for Prestissimo needs to be fixed as well.` But makes sense to make that into another PR, so i'll switch it out

velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
jkhaliqi: When decimal does not fit in the column it fails and this is done in java code before it comes to the writer.

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: Check is done in the Java code before it comes here so I believe it should be safe to remove this check here.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1141,6 +1141,41 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
...
+  auto insertTableHandle = createHiveInsertTableHandle(
+      rowType,
+      "/path/to/test",
+      dwio::common::FileFormat::PARQUET,
+      {"c5", "c6"},
+      bucketProperty);
+  ASSERT_EQ(
+      0,
+      strcmp(
+          insertTableHandle->getColumnHandleDataType()[0]->name(), "DECIMAL"));
```
czentgr, line 1165: Instead of name, why don't you check on `toString` of the `TypePtr`? That will give you precision and scale for DECIMAL.
That way you don't need to check the table later which could have changes that you really care about.

velox/dwio/parquet/writer/Writer.cpp
czentgr: Lets add a comment to this effect here.

czentgr (commented):
Looks good so far.
Can you please add more description to the commit and description in the PR as to the problem solved here?
We don't have an OSS issue for this that describes the problem? If so lets link it.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -411,10 +418,6 @@ dwio::common::StripeProgress getStripeProgress(
  * This method assumes each input `ColumnarBatch` have same schema.
  */
 void Writer::write(const VectorPtr& data) {
-  VELOX_USER_CHECK(
```
czentgr, line 414: Is this too strong to remove this here? So this would trigger because the type is DECIMAL(8,2) but the schema type is DECIMAL(6,2)? For other types we may not have this problem. Perhaps for DECIMAL we need a separate chat to check that we allow a larger precision?

jkhaliqi: Check is done in the Java code before it comes here so I believe it should be safe to remove this check here.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1141,6 +1141,33 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
       "HiveInsertTableHandle [dwrf zstd], [inputColumns: [ HiveColumnHandle [name: c0, columnType: Regular, dataType: BIGINT, requiredSubfields: [ ]] HiveColumnHandle [name: c1, columnType: Regular, dataType: INTEGER, requiredSubfields: [ ]] HiveColumnHandle [name: c2, columnType: Regular, dataType: SMALLINT, requiredSubfields: [ ]] HiveColumnHandle [name: c3, columnType: Regular, dataType: REAL, requiredSubfields: [ ]] HiveColumnHandle [name: c4, columnType: Regular, dataType: DOUBLE, requiredSubfields: [ ]] HiveColumnHandle [name: c5, columnType: PartitionKey, dataType: VARCHAR, requiredSubfields: [ ]] HiveColumnHandle [name: c6, columnType: PartitionKey, dataType: BOOLEAN, requiredSubfields: [ ]] ], locationHandle: LocationHandle [targetPath: /path/to/test, writePath: /path/to/test, tableType: kNew, tableFileName: ], bucketProperty: \nHiveBucketProperty[<HIVE_COMPATIBLE 4>\n\tBucket Columns:\n\t\tc5\n\tBucket Types:\n\t\tVARCHAR\n\tSortedBy Columns:\n\t\t[COLUMN[c5] ORDER[DESC NULLS LAST]]\n]\n, fileNameGenerator: HiveInsertFileNameGenerator]");
 }
 
+TEST_F(HiveDataSinkTest, insertTableHandleGetColumnHandleDataType) {
+  const int32_t numBuckets = 4;
```
majetideepak: nit:  `const int32_t  kNumBuckets = 4`

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -226,6 +226,13 @@ std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
         arrowMapType->item_field(), *mapType.valueType());
     return newField->WithType(
         ::arrow::map(newKeyField->type(), newValueField->type()));
+  } else if (name != "" && type.isDecimal()) {
```
majetideepak: nit: `!name.empty()`

velox/dwio/parquet/writer/Writer.cpp
majetideepak: We should validate in Velox as well since it's a general-purpose library.
For decimals, it should be enough to check if the data decimal value fits in the decimal table type.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -226,6 +226,13 @@ std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
...
+    auto newField = field->WithName(name);
+    auto precisionScale = getDecimalPrecisionScale(type);
+    // Parquet type is set from the column type rather than inferred from the
+    // field data.
+    return newField->WithType(
```
majetideepak: Don't we have to rescale the value? If the data is 1200.00 and say the precision, scale is (6, 2), the value will be `120'000`. If the Table type is (7, 3) the value must be `1'200'000`?

jkhaliqi: Added in test `scaleDecimal` to test out the values getting rescaled with parquet file that was created with values having less scale being inserted into table with higher scale example 12.34(DECIMAL(4,2)) becomes 123400 in the parquet file.

majetideepak: No need to check in a file. Can't we write and read with this change?

majetideepak: Where does the rescale happen on write?

jkhaliqi: This change just sets the precision and scale of the decimal type based on the table that was created, rather than the value that was inserted.
I believe the rescaling is happening before any changes in the PR since the first part hit is this file `TableWriter.cpp` and `const auto& inputType = tableWriteNode->sources()[0]->outputType();` already has the data rescaled coming from `LocalPlanner.cpp auto planNode = planNodes[i];` If the table is (10,4) and I insert `12.12` inputType will read as `10,4` but if I insert `12.1234` then inputType will be `6,4`, but will look into this value rescaling some more to make sense of it...

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -424,9 +431,8 @@ dwio::common::StripeProgress getStripeProgress(
  */
 void Writer::write(const VectorPtr& data) {
   VELOX_USER_CHECK(
-      data->type()->equivalent(*schema_),
+      data->type()->equivalentWithSimilarDecimal(*schema_),
```
czentgr: This doesn't apply to other types, only to decimals.
So I suugest to change this to something where it is more clear.
```
if (data->type()->isDecimal()) {
 // do the new decimal specific check implemented for the DecimalType.
 // might need to convert type to short or long decimal and then call the function. There is no function to just create a pointer to the DecimalType.
}
else {
// current check
}

```

The VELOX_CHECK error line is also different for a Decimal.

velox/type/Type.cpp
```diff
@@ -672,6 +672,33 @@ bool Type::kindEquals(const TypePtr& other) const {
...
+  const auto& otherTyped = other.asRow();
+  if (otherTyped.size() != size()) {
+    return false;
+  }
+  for (size_t i = 0; i < size(); ++i) {
+    if (otherTyped.childAt(i)->isDecimal()) {
+      auto presicisonScaleSchema =
+          getDecimalPrecisionScale(*otherTyped.childAt(i));
+      auto precisionScaleData = getDecimalPrecisionScale(*childAt(i));
+      if (presicisonScaleSchema.first < precisionScaleData.first ||
```
czentgr: Other type must have same or less precision with matching scale to be considered similar enough to allow equivalence.
Maybe add a comment. Allow a larger precision but require same scale.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -337,6 +337,14 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
 
   std::string toString() const override;
 
+  const std::vector<TypePtr> getColumnHandleDataType() const override {
```
czentgr: A column handle represents a single column.
So this function doesn't return the data type for a single handle but a vector of types for each handle.
So the name is kinda odd. Should this not be plural `getColumnHandleDataTypes`?

velox/dwio/parquet/writer/Writer.cpp
czentgr: One question here, I suppose the literal is checked on the coordinator as well in the sense that it is checked that it will fit into the table column type.
For example, `12.1234` won't fit into `DECIMAL(10,2)`.

But for your comment, the plan will have the table column type along with the literal value that is passed down (not sure which plan node the literal is - maybe a VALUES node).

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -476,6 +483,34 @@ void Writer::write(const VectorPtr& data) {
   arrowContext_->stagingBytes += bytes;
 }
 
+bool Writer::equivalentWithSimilarDecimal(
```
czentgr: Does this need to be a member function?
Why can't it be a helper function in the anonymous namespace? No specific Writer information is needed.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -337,6 +337,14 @@ class HiveInsertTableHandle : public ConnectorInsertTableHandle {
 
   std::string toString() const override;
 
+  const std::vector<TypePtr> getColumnHandleDataTypes() const override {
```
majetideepak: We already have `inputColumns()`. We don't need to add this API again.
Let't make this an inline function inside TableWriter.cpp

velox/experimental/cudf/connectors/parquet/ParquetDataSink.h
```diff
@@ -254,6 +254,14 @@ class ParquetInsertTableHandle : public ConnectorInsertTableHandle {
 
   static void registerSerDe();
 
+  const std::vector<TypePtr> getColumnHandleDataTypes() const override {
```
majetideepak: Remove this as well.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1143,6 +1143,33 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
       "HiveInsertTableHandle [dwrf zstd], [inputColumns: [ HiveColumnHandle [name: c0, columnType: Regular, dataType: BIGINT, requiredSubfields: [ ]] HiveColumnHandle [name: c1, columnType: Regular, dataType: INTEGER, requiredSubfields: [ ]] HiveColumnHandle [name: c2, columnType: Regular, dataType: SMALLINT, requiredSubfields: [ ]] HiveColumnHandle [name: c3, columnType: Regular, dataType: REAL, requiredSubfields: [ ]] HiveColumnHandle [name: c4, columnType: Regular, dataType: DOUBLE, requiredSubfields: [ ]] HiveColumnHandle [name: c5, columnType: PartitionKey, dataType: VARCHAR, requiredSubfields: [ ]] HiveColumnHandle [name: c6, columnType: PartitionKey, dataType: BOOLEAN, requiredSubfields: [ ]] ], locationHandle: LocationHandle [targetPath: /path/to/test, writePath: /path/to/test, tableType: kNew, tableFileName: ], bucketProperty: \nHiveBucketProperty[<HIVE_COMPATIBLE 4>\n\tBucket Columns:\n\t\tc5\n\tBucket Types:\n\t\tVARCHAR\n\tSortedBy Columns:\n\t\t[COLUMN[c5] ORDER[DESC NULLS LAST]]\n]\n, fileNameGenerator: HiveInsertFileNameGenerator]");
 }
 
+TEST_F(HiveDataSinkTest, insertTableHandleGetColumnHandleDataTypes) {
```
majetideepak: Let's remove this test as well.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -318,6 +325,34 @@ std::optional<std::string> getParquetCreatedBy(
...
+    const TypePtr dataType,
+    const Type& schemaType) {
+  const auto& schemaTyped = schemaType.asRow();
+  if (schemaTyped.size() != dataType->size()) {
+    return false;
+  }
+  if (!(typeid(*dataType) == typeid(schemaType))) {
+    return false;
+  }
+  for (size_t i = 0; i < dataType->size(); ++i) {
```
majetideepak: This does not handle nested types where the decimal type could be inside an array for example.

velox/connectors/hive/HiveDataSink.h
majetideepak: I realized this will require making hive connector as a dependency to Parquet which is not ideal. This API works.

velox/connectors/Connector.h
```diff
@@ -163,6 +163,8 @@ class ConnectorInsertTableHandle : public ISerializable {
 
   virtual std::string toString() const = 0;
 
+  virtual const std::vector<TypePtr> getColumnHandleDataTypes() const = 0;
```
majetideepak: I am thinking if we should add this API to the TableWriteNode. This interface seems to have minimal API calls so far.

majetideepak: Or we should fix `tableWriteNode->columns()` to use HiveInsertTableHandle to infer the output columns.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1143,6 +1143,33 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
...
+      std::vector<std::string>{"c5"},
+      std::vector<TypePtr>{VARCHAR()},
+      std::vector<std::shared_ptr<const HiveSortingColumn>>{
+          std::make_shared<HiveSortingColumn>(
+              "c5", core::SortOrder{false, false})});
+  auto rowType =
+      ROW({"c0", "c5", "c6"}, {DECIMAL(10, 2), VARCHAR(), BOOLEAN()});
+  auto insertTableHandle = createHiveInsertTableHandle(
+      rowType,
+      "/path/to/test",
```
PingLiuPing:
```suggestion
      TempDirectoryPath::create()->getPath();
```

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1143,6 +1143,33 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
...
+      dwio::common::FileFormat::PARQUET,
+      {"c5", "c6"},
+      bucketProperty);
+  ASSERT_EQ(
+      insertTableHandle->getColumnHandleDataTypes()[0]->toString(),
```
PingLiuPing: This works, but I suggest to explicitly compare the precision and scale.

PingLiuPing (commented):
I just happened to hit this decimal issue, wondering what's the status of this PR? Thanks.

jkhaliqi:
@PingLiuPing Have launched an issue to discuss design that can be found here: https://github.com/facebookincubator/velox/issues/14446.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1143,6 +1143,41 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
...
+              "c5", core::SortOrder{false, false})});
+  auto rowType =
+      ROW({"c0", "c5", "c6"}, {DECIMAL(10, 2), VARCHAR(), BOOLEAN()});
+  auto insertTableHandle = createHiveInsertTableHandle(
+      rowType,
+      TempDirectoryPath::create()->getPath(),
+      dwio::common::FileFormat::PARQUET,
+      {"c5", "c6"},
+      bucketProperty);
+  auto precision = insertTableHandle->getColumnHandleDataTypes()[0]
```
PingLiuPing:
```suggestion
auto [precision, scale] = getDecimalPrecisionScale(insertTableHandle->getColumnHandleDataTypes()[0]);
```

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -229,7 +229,18 @@ std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
...
     return newField->WithType(
         ::arrow::map(newKeyField->type(), newValueField->type()));
-  } else if (name != "") {
+  } else if (type.isDecimal()) {
+    // Parquet type is set from the column type rather than inferred from the
+    // field data.
+    auto precisionScale = getDecimalPrecisionScale(type);
+    if (!name.empty()) {
+      auto newField = field->WithName(name);
+      return newField->WithType(
```
PingLiuPing, line 286: You don't need to return twice in this if branch. The first return can be omitted.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -425,7 +502,7 @@ dwio::common::StripeProgress getStripeProgress(
  */
 void Writer::write(const VectorPtr& data) {
   VELOX_USER_CHECK(
-      data->type()->equivalent(*schema_),
+      equivalentWithSimilarDecimal(*data->type(), *schema_),
```
PingLiuPing: Can you please explain why we implement a new function here?
Should we just enhance `equivalent` for DecimalType?
And the function name `equivalentWithSimilarDecimal` is misleading. Should not include decimal.

velox/exec/TableWriter.cpp
PingLiuPing: I suspect something we can investigate is can we have a look at how `node->columns` is been constructed in Presto Java. Can we send the original DECIMAL(10, 2) ?

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: The current `equivalent ` checks for a match for everything and in Decimal case it will check if the scale and precision is the same. Which will fail for our Decimal case. For this `equivalentWithSimilarDecimal` check we can have a Decimal table column value like `Decimal(10,2)` and insert something like `1234.56` which is `Decimal(6,2)`. So the precision and scale are not the same but the precision fits inside the table and the scale is the same, which is what we are checking for now in this case. It is named `equivalentWithSimilarDecimal ` since we check for equivalence, but also similar decimal case is handled instead of exact match for decimal. The current `equivalent` needs to stay for cases that we are checking an exact match for both precision and scale on Decimal values.

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: The first return is for a decimal type that does not have a name since we need to give it its name and the types, but the second return is if it already has a name then we just return it with the types. I believe they are both needed for this case. Can you explain why the first return can be omitted?

PingLiuPing:
```
    auto[p, s] = getDecimalPrecisionScale(type);
    if (!name.empty()) {
      field = field->WithName(name);
    }
    return field->WithType(
        ::arrow::decimal(p, s));
```

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Yes, understood. Here is the common code path, but you just need special logic for decimal. So the ideal way is re-implement equivalent in DecimalType class. Or at least you can check the data type here and only do special handling for decimal type. Adding a new function for all data type is not necessary.

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: Seems like field is passed in as 
`const std::shared_ptr<::arrow::Field>& field`
so it can't be reassigned

PingLiuPing: Oh, sorry, didn't noticed that.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: As discussed, let's leave this code as is. The problem should be fixed in upstream.

velox/dwio/parquet/writer/Writer.cpp
majetideepak: Seeing your comment here https://github.com/facebookincubator/velox/issues/14446#issuecomment-3259681344 the cast will ensure the values fit. So we are good here.

velox/dwio/parquet/writer/Writer.cpp
majetideepak: https://github.com/facebookincubator/velox/issues/14446#issuecomment-3259681344 explains this

velox/dwio/parquet/writer/Writer.cpp
majetideepak: We need a similar `isTypeOnlyCoercion` as Java
https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/type/TypeCoercer.java#L87C20-L87C38

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -1204,6 +1204,37 @@ TEST_F(HiveDataSinkTest, insertTableHandleToString) {
...
 }
 
+TEST_F(HiveDataSinkTest, insertTableHandleGetColumnHandleDataTypes) {
+  const int32_t kNumBuckets = 4;
+  auto bucketProperty = std::make_shared<HiveBucketProperty>(
```
majetideepak: Do we need the bucketProperty?

jkhaliqi: Not needed, removed from test.

velox/exec/TableWriter.cpp
jkhaliqi: @PingLiuPing Updated to send original table values with this PR: https://github.com/prestodb/presto/pull/26737

velox/dwio/parquet/writer/Writer.cpp
czentgr: So we could now keep this check I suppose. With the other change we will use column types of the table and thus they should be equivalent.

czentgr (commented):
Now we likely can re-add the one check.

velox/dwio/parquet/writer/Writer.cpp
jkhaliqi: Seems like the data that is passed into this function is still coming from the input type so it will still be a mismatch, since input can be not exact match with the decimal still?

`auto sourceNode = toVeloxQueryPlan(fragment.root, tableWriteInfo, taskId);`
`auto inputType = sourceNode->outputType();`

where `sourceNode->outputType` was never changed and comes from 
```
    const auto outputType = toRowType(
      generateOutputVariables(
          {node->rowCountVariable,
           node->fragmentVariable,
           node->tableCommitContextVariable},
          node->statisticsAggregation),
      typeParser_);
  return std::make_shared<core::TableWriteNode>(
      node->id,
      toRowType(node->columns, tableTypes),
      node->columnNames,
      columnStatsSpec,
      std::move(insertTableHandle),
      node->partitioningScheme != nullptr,
      outputType,
      getCommitStrategy(),
      sourceVeloxPlan);
```

jkhaliqi: in [PrestoToVeloxQueryPlan.cpp](https://github.com/prestodb/presto/pull/26737/changes#diff-b596594147f956ff5aadb512a14f1538d5975c551dad6ba773ac266762187734). We now cast the `sourceVeloxPlan->outputType` to the table type if there is a mismatch so placed back the check and works just fine.

czentgr (commented):
Can we please add a test for this slightly different behavior?

czentgr (approved):
Thanks!

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -661,6 +662,55 @@ TEST_F(ParquetWriterTest, updateWriterOptionsFromHiveConfig) {
...
+           [](auto row) { return row * 10000 + 9999; },
+           nullptr,
+           DECIMAL(10, 2)),
+       makeFlatVector<int128_t>(
+           kRows,
+           [](auto row) { return HugeInt::build(row, row * 1000000000LL); },
+           nullptr,
+           DECIMAL(38, 10))});
+
+  auto* sinkPtr = write(data);
```
czentgr: The write function is coming from the ParquetTestBase. It generates a Writer with the schema being derived from the `data->type()`. Because the data vector already has DECIMAL(10,2) etc it will use that. But we should create data of type DECIMAL(4,2) and set the schema to DECIMAL(10,2) when creating the writer.

Youc an add a new optional parameter to the write for the schema that defaults to `data->type())` but can be changed. And the change would occur in this call to differ the writer schema from the data schema.

jkhaliqi:
Functionality and testing done in presto PR: https://github.com/prestodb/presto/pull/26737.

karthikeyann (approved):

PR #12848, opened 2025-03-31
feat: Support date type for Spark from_json function

zhli1142015:
https://github.com/apache/spark/blob/c8c8c9a22edc3ae81ffdb8ebf35e8a677a0a4cca/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JacksonParser.scala#L380


zhli1142015:
@rui-mo and @jinchengchenghh , could you please help review?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -99,7 +100,57 @@ struct ExtractJsonTypeImpl {
...
+          } else {
+            return simdjson::INCORRECT_TYPE;
+          }
+        } else {
+          auto days =
+              util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+          if (!days.hasError()) {
+            day = days.value();
+          } else {
+            // remove 'GMT'.
```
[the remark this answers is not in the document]
zhli1142015: Updated.
There is already a test case for GMT.

velox/functions/sparksql/tests/FromJsonTest.cpp
```diff
@@ -167,6 +167,35 @@ TEST_F(FromJsonTest, basicDouble) {
...
+       std::nullopt,
+       std::nullopt,
+       std::nullopt},
+      DATE());
+  auto input = makeFlatVector<std::string>(
+      {R"({"a": "2021-07-01T"})",
+       R"({"a": "2021-07-01"})",
+       R"({"a": "2021-7"})",
+       R"({"a": "1970"})",
+       R"({"a": "2021-07-01GMT"})",
```
jinchengchenghh, line 189: Is the GMT only be the end? If that, we can use StringView(dateStr.data(), size - gmt_size)

velox/docs/functions/spark/json.rst
```diff
@@ -24,14 +24,16 @@ JSON Functions
...
-    REAL, DOUBLE, VARCHAR, ARRAY, MAP and ROW. When casting to ARRAY or MAP, 
+    REAL, DOUBLE, DATE, VARCHAR, ARRAY, MAP and ROW. When casting to ARRAY or MAP, 
     the element type of the array or the value type of the map must be one of 
     these supported types, and for maps, the key type must be VARCHAR. Casting 
     to ROW supports only JSON objects. ::
         
         SELECT from_json('{"a": true}'); -- {'a'=true} // Output type: ROW({"a"}, {BOOLEAN()})
         SELECT from_json('{"a": 1}'); -- {'a'=1} // Output type: ROW({"a"}, {INTEGER()})
         SELECT from_json('{"a": 1.0}'); -- {'a'=1.0} // Output type: ROW({"a"}, {DOUBLE()})
+        SELECT from_json('{"a":"2021-7-1T"}''); -- {'a'="2021-07-01"} // Output type: ROW({"a"}, {DATE()})
```
jinchengchenghh, line 35:
```
spark-sql (default)> SELECT from_json('{"a":"2021-7-1T"}'');

[PARSE_SYNTAX_ERROR] Syntax error at or near ''': extra input '''.(line 1, pos 36)

== SQL ==
SELECT from_json('{"a":"2021-7-1T"}'');
------------------------------------^^^
```
Changed to this but another exception, looks like we need to specify the schema for it.
```
spark-sql (default)> SELECT from_json('{"a":"2021-7-1T"}');
[WRONG_NUM_ARGS.WITHOUT_SUGGESTION] The `from_json` requires [2, 3] parameters but the actual number is 1. Please, refer to 'https://spark.apache.org/docs/latest/sql-ref-functions.html' for a 
```

```
spark-sql (default)> SELECT from_json('{"a":"2021-7-1T"}', 'a Date');
{"a":2021-07-01}
Time taken: 1.172 seconds, Fetched 1 row(s)
```

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -99,7 +101,63 @@ struct ExtractJsonTypeImpl {
...
+        if (s.size() < 4) {
+          auto result =
+              folly::tryTo<int32_t>(std::string_view(s.data(), s.size()));
+          if (!result.hasError()) {
+            day = result.value();
+          } else {
+            return simdjson::INCORRECT_TYPE;
+          }
+        } else {
+          auto days =
```
jinchengchenghh, line 127: Is here same with `CAST string as date`?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -99,7 +101,63 @@ struct ExtractJsonTypeImpl {
...
+              util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+          if (!days.hasError()) {
+            day = days.value();
+          } else {
+            if (stringImpl::stringPosition<true /*isAscii*/>(
```
jinchengchenghh, line 132: Do we need to find the GMT and then replace, can we replace directly?

zhli1142015: The following replace function does not perform in-place replacement. A copy is always made, which is why it first checks whether replacement is needed.

I have tried in-place replacement, but it seems that data copying is unavoidable.
https://github.com/facebookincubator/velox/blob/9f993aa6c927620cac01bcd0b96f92edcd81ab3e/velox/functions/lib/string/tests/StringImplTest.cpp#L518

velox/docs/functions/spark/json.rst
zhli1142015: This function has different signatures in Velox and Spark.

In Spark, the second parameter is the schema of the result type.
In Velox, the result type is inferred from the expression, so the second parameter is not needed.

jinchengchenghh: Can you add this explanation to document?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -99,7 +101,63 @@ struct ExtractJsonTypeImpl {
...
+        std::string_view s;
+        switch (type) {
+          case simdjson::ondemand::json_type::string: {
+            SIMDJSON_ASSIGN_OR_RAISE(s, value.get_string());
+            break;
+          }
+          default:
+            return simdjson::INCORRECT_TYPE;
+        }
+        int32_t day = 0;
```
jinchengchenghh, line 115: Can we extract the convertStringToDate logic to a new function, looks like we may
 support other conversions in the future, here will be much logic

velox/docs/functions/spark/json.rst
zhli1142015: Done, thanks.

velox/functions/sparksql/tests/FromJsonTest.cpp
zhli1142015: Done, thanks

velox/functions/sparksql/specialforms/FromJson.cpp
zhli1142015: Updated, thanks.

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,62 @@ struct ExtractJsonTypeImpl {
...
+      }
+    } else {
+      auto days =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!days.hasError()) {
+        day = days.value();
+      } else {
+        if (stringImpl::stringPosition<true /*isAscii*/>(
+                std::string_view(s.data(), s.size()), kGMT, 1) > 0) {
+          // remove 'GMT'.
```
jinchengchenghh, line 307: Comment starts with uppercase

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,62 @@ struct ExtractJsonTypeImpl {
...
+        break;
+      }
+      default:
+        return simdjson::INCORRECT_TYPE;
+    }
+    int32_t day = 0;
+    // If there are fewer than four digits, the value is treated as the
+    // number of days since 1970-01-01.
+    if (s.size() < 4) {
+      auto result = folly::tryTo<int32_t>(std::string_view(s.data(), s.size()));
```
jinchengchenghh, line 293: ditto

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,62 @@ struct ExtractJsonTypeImpl {
...
+    } else {
+      auto days =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!days.hasError()) {
+        day = days.value();
+      } else {
+        if (stringImpl::stringPosition<true /*isAscii*/>(
+                std::string_view(s.data(), s.size()), kGMT, 1) > 0) {
+          // remove 'GMT'.
+          std::string dateStr;
```
jinchengchenghh, line 308: Maybe char[s.size()] is better

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,62 @@ struct ExtractJsonTypeImpl {
...
+        return simdjson::INCORRECT_TYPE;
+      }
+    } else {
+      auto days =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!days.hasError()) {
+        day = days.value();
+      } else {
+        if (stringImpl::stringPosition<true /*isAscii*/>(
+                std::string_view(s.data(), s.size()), kGMT, 1) > 0) {
```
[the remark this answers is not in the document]
zhli1142015: Good catch!

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,56 @@ struct ExtractJsonTypeImpl {
...
+      exec::GenericWriter& writer) {
+    SIMDJSON_ASSIGN_OR_RAISE(auto type, value.type());
+    std::string_view s;
+    switch (type) {
+      case simdjson::ondemand::json_type::string: {
+        SIMDJSON_ASSIGN_OR_RAISE(s, value.get_string());
+        break;
+      }
+      default:
+        return simdjson::INCORRECT_TYPE;
```
rui-mo, line 287: Can we use if-else if only two cases?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +273,56 @@ struct ExtractJsonTypeImpl {
...
+      }
+    } else {
+      auto days =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!days.hasError()) {
+        day = days.value();
+      } else {
+        if (stringImpl::stringPosition<true /*isAscii*/>(s, kGMT, 1) > 0) {
+          // Remove 'GMT'.
```
rui-mo, line 306: Are several "GMTs" that are present at any position valid in Spark?  Is there any other allowed string?

rui-mo (commented):
Thanks.

velox/docs/functions/spark/json.rst
```diff
@@ -24,14 +24,18 @@ JSON Functions
...
     determined by the expression. Returns NULL, if the input string is unparsable.
     Supported element types include BOOLEAN, TINYINT, SMALLINT, INTEGER, BIGINT, 
-    REAL, DOUBLE, VARCHAR, ARRAY, MAP and ROW. When casting to ARRAY or MAP, 
+    REAL, DOUBLE, DATE, VARCHAR, ARRAY, MAP and ROW. When casting to ARRAY or MAP, 
     the element type of the array or the value type of the map must be one of 
     these supported types, and for maps, the key type must be VARCHAR. Casting 
-    to ROW supports only JSON objects. ::
+    to ROW supports only JSON objects.
+    Note,  this function's signature differs from Spark's because the result type 
+    is inferred from the expression. ::
```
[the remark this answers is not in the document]
zhli1142015: I added this because the function's signature in Velox is different from that in Spark — Spark's function signature includes the result type information. In velox, we can infer it from expression.

velox/functions/sparksql/specialforms/FromJson.cpp
zhli1142015: Updated.

velox/functions/sparksql/specialforms/FromJson.cpp
zhli1142015: The Spark PR mentioned that all instances of GMT need to be removed, but the only example I found in the Spark repo is 2000-01-01T00:00GMT+01:00.
https://github.com/apache/spark/commit/979a8eb04ae1dec2928cc49028b3df03b4ba78ce
https://github.com/apache/spark/blob/418cfd1f78014698ac4baac21156341a11b771b3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L72

velox/docs/functions/spark/json.rst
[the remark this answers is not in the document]
rui-mo: In Velox the special from also uses the result type passed from Spark, no?
https://github.com/facebookincubator/velox/blob/1a282f0bda7a39ae30dd5f034a6529aa96298550/velox/functions/sparksql/specialforms/FromJson.cpp#L388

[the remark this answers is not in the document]
zhli1142015: ​The from_json function signatures differ between Spark and Velox. In Spark, from_json determines the return type using the schema parameter. In Velox, this parameter is not required. If you're only familiar with Spark's function, the following examples might be confusing and won't run directly in Spark. They are provided here to illustrate this difference. If you believe this explanation is unnecessary, I'm open to removing it. Please let me know your preference. Thank you.

Spark: from_json(jsonStr, schema[, options])​
Velox: from_json(jsonStr)​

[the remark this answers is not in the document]
zhli1142015: I think so — the `dataType` property of this expression (which corresponds to the Velox result type) is the same as the `schema`.
https://github.com/apache/spark/blob/92f5d38ac397190dd866405b94bb8c98ae31aaa3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/jsonExpressions.scala#L306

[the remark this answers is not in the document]
jinchengchenghh: We may also need to add one SparkSql query as example.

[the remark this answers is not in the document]
zhli1142015: Updated.

zhli1142015:
The CI failure is not related to this PR. https://github.com/facebookincubator/velox/issues/12970

velox/docs/functions/spark/json.rst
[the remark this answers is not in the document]
rui-mo: > In Spark, from_json determines the return type using the schema parameter. In Velox, this parameter is not required.

Thanks for the explanation! Does this mean we have simplified the implementation in Velox?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +272,52 @@ struct ExtractJsonTypeImpl {
...
+      if (!result.hasError()) {
+        day = result.value();
+      } else {
+        return simdjson::INCORRECT_TYPE;
+      }
+    } else {
+      auto days =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!days.hasError()) {
+        day = days.value();
```
rui-mo, line 298: nit: the names of `day` and `days` might be confusing as their difference is not the singular or plural form.

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +272,52 @@ struct ExtractJsonTypeImpl {
...
+      } else {
+        if (stringImpl::stringPosition<true /*isAscii*/>(s, kGMT, 1) > 0) {
```
rui-mo, line 300: Can we combine them as `else if`?

velox/docs/functions/spark/json.rst
```diff
@@ -24,14 +24,18 @@ JSON Functions
...
         
         SELECT from_json('{"a": true}'); -- {'a'=true} // Output type: ROW({"a"}, {BOOLEAN()})
         SELECT from_json('{"a": 1}'); -- {'a'=1} // Output type: ROW({"a"}, {INTEGER()})
         SELECT from_json('{"a": 1.0}'); -- {'a'=1.0} // Output type: ROW({"a"}, {DOUBLE()})
+        SELECT from_json('{"a":"2021-7-1T"}'); -- {'a'="2021-07-01"} // Output type: ROW({"a"}, {DATE()})
+        SELECT from_json('{"a":"1"}'); -- {'a'="1970-01-02"} // Output type: ROW({"a"}, {DATE()})
```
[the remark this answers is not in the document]
zhli1142015: I'm also quite confused — Velox doesn't provide any SQL interface, yet we're using SQL statements as an example. I think this can only demonstrate the interfaces we expose to upper layers, but it's not actually usable. If we're only referring to Spark's interface, we might not need to add extra documentation and could just link to Spark's own docs directly.

velox/docs/functions/spark/json.rst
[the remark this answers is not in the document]
zhli1142015: I think this just illustrates that the function signatures exposed by Velox are different from those expected by Spark.

velox/docs/functions/spark/json.rst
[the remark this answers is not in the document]
rui-mo: I suppose the examples illustrate how we use the functions in Spark or Presto SQL, so we need to provide Spark-compatible queries.

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +272,50 @@ struct ExtractJsonTypeImpl {
...
+        day = result.value();
+      } else {
+        return simdjson::INCORRECT_TYPE;
+      }
+    } else {
+      auto castResult =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!castResult.hasError()) {
+        day = castResult.value();
+      } else if (stringImpl::stringPosition<true /*isAscii*/>(s, kGMT, 1) > 0) {
```
rui-mo, line 299: nit: `/*isAscii=*/true`

velox/docs/functions/spark/json.rst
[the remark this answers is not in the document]
zhli1142015: >  so we need to provide Spark-compatible queries.

Thanks for the clarification. The examples are now updated to be consistent with Spark.

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +272,51 @@ struct ExtractJsonTypeImpl {
...
+        day = result.value();
+      } else {
+        return simdjson::INCORRECT_TYPE;
+      }
+    } else {
+      auto castResult =
+          util::fromDateString(StringView(s), util::ParseMode::kSparkCast);
+      if (!castResult.hasError()) {
+        day = castResult.value();
+      } else if (stringImpl::stringPosition</*isAscii*/ true>(s, kGMT, 1) > 0) {
```
rui-mo, line 299: nit: also add a `=` after the variable name according to the code style at https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md#function-arguments.

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -267,6 +272,51 @@ struct ExtractJsonTypeImpl {
...
+        // Try converting the string view as a date value after cleaning up the
+        // legacy timestamp string by removing the "GMT" string.
+        char dateStr[s.size()];
+        auto size = stringCore::replace</*ignoreEmptyReplaced*/ true>(
```
rui-mo, line 303: ditto

rui-mo (approved):

PR #12855, opened 2025-03-31
refactor: Cleanup toConstantSql

mbasmanova:
Summary:
- Use std::quoted instead of custom code: https://en.cppreference.com/w/cpp/io/manip/quoted
- Reduce copy-paste.

Differential Revision: D72165464


xiaoxmeng (approved):
@mbasmanova thanks!

PR #12856, opened 2025-03-31
refactor: Harden is-null check in toConstantSql

mbasmanova:
Summary:
> constant->toString() == "null"

toString() API is not designed to return stable value and should not be used for anything other than logging and debugging.

Replace with constant->isNull().

Differential Revision: D72166016


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

Issue #12830, opened 2025-03-27
Compatibility enhancements on AggregationNode


zhztheplayer:
## Background

Velox's `AggregationNode` was likely designed following Presto style, where [step](https://github.com/facebookincubator/velox/blob/e45fc7383576e42d68ee624b587c52bf53686678/velox/core/PlanNode.h#L596-L605) is a property of the aggregation operator. As a comparison, Spark's [AggregateMode](https://github.com/apache/spark/blob/cbcc298970fe1b5221bb1eef69446e7b26e2b934/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/interfaces.scala#L101) is bound to a specific `AggreagateFunction`, making it possible that one Spark aggregate operator has aggregate functions that are with different modes (step). There was also an old [discussion](https://github.com/facebookincubator/velox/issues/4412) around this difference.

Apache Gluten has been relying on some [tricks](https://github.com/apache/incubator-gluten/pull/4130) to align Spark and Velox on this: the planner interprets Spark's aggregation modes into different types of Velox companion functions, then assign `single` step to Velox's `AggregationNode` constantly. By doing this we can get rid of a lot of relevant issues caused of the mismatch. For example, we can naturally support spill-able partial aggregation which is unsupported by Velox. We can also avoid unwanted flushing causing result mismatches because of some specific query plans generated by Spark.

The solution basically worked as expected until we met that sometimes the **intermediate** / **final** companion functions are not resolvable given intermediate input types for some functions. Because the functions accept different types of input but use the same type of intermediate data. For example [this one](https://github.com/facebookincubator/velox/blob/e45fc7383576e42d68ee624b587c52bf53686678/velox/functions/sparksql/aggregates/AverageAggregate.cpp#L102-L108). 

The issue is literally hard to fix completely because, the companion functions are all treated as normal aggregation functions in Velox, even we have the input types of the original aggregation function from Spark, we can't always use them to find **intermediate** / **final** companion functions accurately because they are resolved with intermediate data types in Velox. Although Velox provides suffixed version of final companion functions to distinguish between them, but theoretically they're not reliable since in SQL world overloaded functions can only be distinguished by function name and input data types.

## Proposed Changes

1. Move `AggregationNode::Step` to `AggregationNode::Aggregate::Step` and make it work.
2. Add a flag to `AggregationNode` to allow user disable flushing actively.

[...]

PR #12861, opened 2025-03-31
Move AggregationNode::Step to AggregationNode::Aggregate::Step

zhztheplayer:
This is an API preparation for supporting executing aggregation functions with different steps in the same operator to make Velox compatible with Spark and to fix https://github.com/facebookincubator/velox/issues/12830.

This contains change to move `AggregationNode::Step` to `AggregationNode::Aggregate::Step` to make each aggregation function call have its own step, and to add a `allowFlush` flag to `AggregationNode` to control whether the aggression is flushable or not.

The patch mainly focuses on API so after the change Velox will throw when mixed function steps are specified.


velox/exec/tests/AggregationTest.cpp
```diff
@@ -837,6 +843,49 @@ TEST_F(AggregationTest, allKeyTypes) {
...
+              {"c0"},
+              {},
+              {"sum(c1)"},
+              {},
+              {core::AggregationNode::Aggregate::Step::kSingle},
+              false,
+              true)
+          .planNode(),
+      VeloxException);
+}
```
zhztheplayer, line 887: These test are added to cover the mixed-steps aggregation cases.

velox/exec/AggregateCompanionAdapter.cpp
```diff
@@ -261,14 +261,14 @@ bool CompanionFunctionsRegistrar::registerPartialFunction(
...
-                 core::AggregationNode::Step step,
+                 core::AggregationNode::Aggregate::Step step,
                  const std::vector<TypePtr>& argTypes,
                  const TypePtr& resultType,
                  const core::QueryConfig& config)
                  -> std::unique_ptr<Aggregate> {
                if (auto func = getAggregateFunctionEntry(name)) {
                  if (!exec::isRawInput(step)) {
-                   step = core::AggregationNode::Step::kIntermediate;
+                   step = core::AggregationNode::Aggregate::Step::kIntermediate;
```
mbasmanova, line 271: @zhztheplayer 

It is a bit difficult to see the material changes because of all the renaming of core::AggregationNode::Step to core::AggregationNode::Aggregate::Step. 

Would it be possible to create a version of this PR that doesn't do the rename, just moves core::AggregationNode::Step property from the plan node to Aggregate?

velox/exec/GroupingSet.cpp
```diff
@@ -103,9 +99,34 @@ GroupingSet::GroupingSet(
...
+    hasPartialInput_ = hasPartialInput;
+    hasPartialOutput_ = hasPartialOutput;
+    allPartialInput_ = allPartialInput;
+    allPartialOutput_ = allPartialOutput;
+  }
+
   sortedAggregations_ =
       SortedAggregations::create(aggregates_, inputType, &pool_);
-  if (isPartial_) {
+  if (hasPartialOutput_) {
```
mbasmanova, line 129: This check doesn't seem correct. I assume it should be fine if none of sortedAggregations_ is partial, no?

velox/exec/tests/PlanNodeToStringTest.cpp
```diff
@@ -198,7 +198,7 @@ TEST_F(PlanNodeToStringTest, aggregation) {
 
   ASSERT_EQ("-- Aggregation[1]\n", plan->toString());
   ASSERT_EQ(
-      "-- Aggregation[1][PARTIAL a := sum(ROW[\"c0\"]), b := avg(ROW[\"c1\"]), c := min(ROW[\"c2\"]), a3 := count(ROW[\"c1\"]) distinct] "
+      "-- Aggregation[1][ALLOW FLUSH PARTIAL a := sum(ROW[\"c0\"]), PARTIAL b := avg(ROW[\"c1\"]), PARTIAL c := min(ROW[\"c2\"]), PARTIAL a3 := count(ROW[\"c1\"]) distinct] "
```
mbasmanova, line 201: This makes the plan harder to read. Perhaps, when all aggs use same step we could keep the original format and use new format only when there is a mix of steps.

velox/exec/AggregateCompanionAdapter.cpp
zhztheplayer: @mbasmanova 

That would be possible. But I'd clarify would you use that PR just for reviewing or you actually prefer that approach?

mbasmanova: Thanks. At this point, I'm still trying to understand the proposal and the changes. I do not have a clear preference yet. Hope this makes sense.

zhztheplayer: Yes I saw your latest comments in https://github.com/facebookincubator/velox/issues/12830. I'll follow up on that.

velox/exec/GroupingSet.cpp
zhztheplayer: > I assume it should be fine if none of sortedAggregations_ is partial

This is true because the check will pass if non of sortedAggregations_ is partial as `hasPartialOutput_` = `false`.

But would you intentionally mean "it should be fine if SOME of sortedAggregations_ are partial"? Which will suggest the change

```
  if (allPartialOutput_) {
    VELOX_USER_CHECK_NULL(
        sortedAggregations_,
        "Partial aggregations over sorted inputs are not supported");
  }
```

? I think it makes sense but can be done in the future because in the first PR I am only changing the APIs. So I put the strongest restriction here.

velox/exec/tests/PlanNodeToStringTest.cpp
zhztheplayer: fixe in https://github.com/facebookincubator/velox/pull/13024

PR #12862, opened 2025-03-31
fix: Fix potential hanging issue in case of a right/full join and empty build table

xiaoxmeng:
Currently NLJ start mismatch build processing when receives no more input signal in noMoreInput()
or when advance the last probe input. We don't need mismatch build processing for right/full
join if the build side is empty. However, currently we only check that in advance probe input site
but not in noMoreInput(), this might cause potential hanging as the last probe wait expect all the probe
peers participate in the mismatch build processing wait

Differential Revision: D72147392

### DIFF
--- a/velox/exec/NestedLoopJoinProbe.h
+++ b/velox/exec/NestedLoopJoinProbe.h
@@ -236,7 +236,7 @@ class NestedLoopJoinProbe : public Operator {
     return (buildIndex_ >= buildVectors_.value().size());
   }
 
-  /// Cross joins are translated into NLJ's without a join conditition.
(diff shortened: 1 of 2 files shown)


velox/exec/NestedLoopJoinProbe.h
```diff
@@ -276,7 +276,8 @@ class NestedLoopJoinProbe : public Operator {
     state_ = state;
   }
 
- private:
```
rui-mo, line 279: nit: do we need this change?

rui-mo (approved):
Thanks!

Yuhta (approved):

PR #12872, opened 2025-04-01
build: Use breeze functions instead of cub functions for dwio

dreveman:
No change in behavior. Just using breeze instead of cub to avoid unnecessary external dependencies.

### DIFF
--- a/velox/experimental/wave/dwio/decode/GpuDecoder-inl.cuh
+++ b/velox/experimental/wave/dwio/decode/GpuDecoder-inl.cuh
@@ -1012,32 +1012,24 @@ __device__ void setRowCountNoFilter(GpuDecode::RowCountNoFilter& op) {
 }
 
 template <int32_t kBlockSize, int32_t kWidth>
-inline __device__ void reduceCase(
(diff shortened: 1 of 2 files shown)


velox/experimental/wave/dwio/decode/GpuDecoder-inl.cuh
```diff
@@ -256,29 +260,57 @@ __device__ inline void decodeDictionaryOnBitpack(GpuDecode& plan) {
   }
 }
 
+template <int32_t kBlockSize>
+struct BlockScanStorage {
```
Yuhta, line 264: Can this be inferred from `BlockScanT`?

dreveman: We can do something more precise for each stage where a scan is needed. I tried to keep it as consistent with existing code as possible in this PR, where we use a maximum size required even if the stage might technically use less. Let me know if this is something that we should improve as part of this change or OK to do as a follow up?

dreveman:
@Yuhta ping for review

PR #12879, opened 2025-04-01
Replaces 2x string.find(..) with custom one using optimized folly:memchr

izard:
This provides performance boost to two Like sub-benchmarks.
Depends on Folly PR: https://github.com/facebook/folly/pull/2409

### DIFF
--- a/velox/functions/lib/string/StringCore.h
+++ b/velox/functions/lib/string/StringCore.h
@@ -312,7 +312,7 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
   return utf8Position;
 }
 
-constexpr size_t long_string_find(
+constexpr size_t longStringFind(
     std::string_view str, const std::string_view substr, size_t start_pos
 ) noexcept
 {
@@ -362,7 +362,7 @@ static inline int64_t findNthInstanceByteIndexFromStart(
     return -1;
   }
 
-  auto byteIndex = long_string_find(string, subString, startPosition);
+  auto byteIndex = longStringFind(string, subString, startPosition);
   // Not found
   if (byteIndex == std::string_view::npos) {
     return -1;
--- a/velox/functions/lib/Re2Functions.cpp
+++ b/velox/functions/lib/Re2Functions.cpp
@@ -674,7 +674,7 @@ bool matchSuffixPattern(
 bool matchSubstringPattern(
     const StringView& input,
     const std::string& fixedPattern) {
-    std::size_t found = facebook::velox::functions::stringCore::long_string_find(
+    std::size_t found = facebook::velox::functions::stringCore::longStringFind(
                             std::string_view(input), std::string_view(fixedPattern), (size_t)0);
   return found != std::string::npos;
 }


embg:
This is a great find (pun intended). Will pull this in after we merge the folly PR.

velox/functions/lib/string/StringCore.h
```diff
@@ -311,6 +312,41 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
...
+
+    const char first_char = substr[0];
+    const char* current = str + start_pos;
+    const char* const end = str + str_len;
+    size_t remaining = str_len - start_pos;
+
+    while (remaining >= substr_len)
+    {
+        // Use optimized memchr variant
+        current = static_cast<const char*>(folly::__folly_memchr(current, first_char, remaining - substr_len + 1));
```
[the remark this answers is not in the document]
embg: Even if we did directly call `__folly_memchr`, how would this code run on x86? The folly PR only adds an implementation for arm64.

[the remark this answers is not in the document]
izard: Thank you very much for spotting! Will fix first thing tomorrow morning.

embg:
Is this PR still a win when using the stdlib memchr? If so, we can merge it now as long as you replace the folly memchr with std::memchr. Wondering if we should do this since the folly PR might take a while to merge.

izard:
> Is this PR still a win when using the stdlib memchr?
No, performance will stay exactly the same with stdlib memchr, all the speedup only comes from faster memory scanning for the first byte.

embg:
@izard Understood. Let's revisit here after we figure out the situation with folly, then. If the win here is sufficiently large, we will find a way to merge it one way or another.

Btw, it would be helpful if tomorrow you can share your benchmark numbers for the sub-benchmarks :)

izard:
> @izard Understood. Let's revisit here after we figure out the situation with folly, then. If the win here is sufficiently large, we will find a way to merge it one way or another.
> Btw, it would be helpful if tomorrow you can share your benchmark numbers :)

Among Velox benchmarks, the only visible difference is ~20% and ~25% uplift in Like/substring and Like/substring and Like/strpos. Everything else is just the same.

izard:
Now using folly::memchr_long explicitly as discussed in folly PR discussion thread.

embg:
I will work on pulling in the folly PR. Meantime, could you benchmark on x86 and verify that there is no regression? Since we only have optimized memchr_long for Arm.

velox/functions/lib/string/StringCore.h
```diff
@@ -311,6 +312,41 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
   return utf8Position;
 }
 
+constexpr size_t long_string_find(
+    const char* str, size_t str_len,
```
embg, line 316: This is C++ code, do not take raw ptr parameters, prefer to take `std::string_view` as a parameter.

izard: Fixed. Thank you!

velox/functions/lib/string/StringCore.h
```diff
@@ -313,20 +313,20 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
 }
 
 constexpr size_t long_string_find(
-    const char* str, size_t str_len,
-    const char* substr, size_t substr_len,
-    size_t start_pos
+    const std::string_view& str, const std::string_view substr, size_t start_pos
```
embg, line 316: nit: `std::string_view` should be passed by value, not by const reference.

izard: fixed

izard:
Just added a commit fixing a bug I introduced earlier, this bug was not affecting the tests results but would backfire in real application.

embg:
@izard 

> Just added a commit fixing a bug I introduced earlier

Thank you!

> this bug was not affecting the tests results but would backfire in real application.

Would it be possible to add a test case? Or is it hard to expose in a test?

embg:
There's a very simple issue which was caught by internal CI. Should be trivial to fix.

```
folly/folly/FollyMemchr.cpp:23:21: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
   23 |   return std::memchr(ptr, ch, count);
      |          ~~~~~~~~~~~^~~~~~~~~~~~~~~~
      |                     |
      |                     const void*
folly/folly/FollyMemchr.cpp: At global scope:
folly/folly/FollyMemchr.cpp:41:1: error: expected declaration before ‘}’ token
   41 | } // namespace folly
```

```
folly/folly/FollyMemchr.cpp:23:10: error: cannot initialize return object of type 'void *' with an rvalue of type 'const void *'
   23 |   return std::memchr(ptr, ch, count);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
folly/folly/FollyMemchr.cpp:41:1: error: extraneous closing brace ('}')
   41 | } // namespace folly
      | ^
```

velox/functions/lib/string/StringCore.h
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <string_view>
 #include "folly/CPortability.h"
+#include "folly/FollyMemchr.h"
```
pedroerp, line 22: how recent does folly need to be to provide the feature needed?

Cc: @kgpai @assignUser

velox/functions/lib/string/StringCore.h
```diff
@@ -311,6 +312,41 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
   return utf8Position;
 }
 
+constexpr size_t long_string_find(
```
pedroerp, line 315: can we add unit tests to exercise this function?

velox/functions/lib/string/StringCore.h
```diff
@@ -311,6 +312,41 @@ cappedByteLengthUnicode(const char* input, size_t size, int64_t maxChars) {
...
+
+    if (start_pos >= str_len)
+        return std::string_view::npos;
+
+    const char first_char = substr.data()[0];
+    const char* current = str.data() + start_pos;
+    const char* const end = str.data() + str_len;
+    size_t remaining = str_len - start_pos;
+
+    while (remaining >= substr_len)
```
pedroerp, line 332: please also review our contributing and coding style guidelines:

https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md
https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md

velox/functions/lib/string/StringCore.h
izard: Very recent, depends on https://github.com/facebook/folly/pull/2409 which is not yet even integrated.

velox/functions/lib/string/StringCore.h
izard: Will do.

assignUser:
I have converted this to draft as the necessary folly PR isn't even landed.

PR #12881, opened 2025-04-01
fix: Fix Spark unix_timestamp function for date type

rui-mo:
Spark unix_timestamp function throws if the microseconds overflows. Correct the 
return type in documentation.


zhli1142015 (approved):

pedroerp (approved):
Thank you

PR #12884, opened 2025-04-01
feat(python): Add file schema inspection API

pedroerp:
Summary:
Add get_schema() API to velox.py.file to enable users to inspect file
schemas and understand out the column available.

Differential Revision: D72245329


xiaoxmeng (approved):
@pedroerp LGTM. Thanks!

PR #12886, opened 2025-04-01
feat: Add isSupportedCompression

xin-zhang2:
We’re introducing support for mutiple compression codecs for Presto exchange in prestodb/presto#24670.
It is unnessary to cast the codec session property string to CompressionKind when creating the query context, but we do need to validate whether it is supported by Velox. To address this, I added the isSupportedCompression function in this PR.
A discussion is in https://github.com/prestodb/presto/pull/24670#discussion_r2015273633.


xin-zhang2:
@yingsu00 
Can you help take a review? Thanks!

PR #12888, opened 2025-04-01
fix(fuzzer): Add SQL generation support for type DATE

peterenescu:
Pull Request resolved: https://github.com/facebookincubator/velox/pull/12888

Allow for SQL generation of input types using DATE to be explicitly specified as DATE rather than INTEGER. Without this time functions using DATE as input generated by their signature will fail.

A regression was identified by the continuous fuzzer run that broke the teset: https://www.internalfb.com/chronos/job/silver/1125908496757416
This was a result of changing `type->toString()` to `mapTypeKindToName(type->kind())` that omits `DATE`.
https://github.com/facebookincubator/velox/issues/12891

### DIFF
--- a/velox/exec/fuzzer/ToSQLUtil.cpp
+++ b/velox/exec/fuzzer/ToSQLUtil.cpp
@@ -28,6 +28,12 @@ void appendComma(int32_t i, std::stringstream& sql) {
 
 // Returns the SQL string of the given type.
 std::string toTypeSql(const TypePtr& type) {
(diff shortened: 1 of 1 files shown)


mbasmanova (changes requested):
@peterenescu Would you add a test?

mbasmanova (approved):
Please, create a GitHub issue to describe the problem this is addressing and mention in the PR summary.

peterenescu:
@mbasmanova will take it up as part of https://github.com/facebookincubator/velox/issues/12891

Issue #12893, opened 2025-04-02
Missing runtime stats for Filter node


mbasmanova:
### Bug description

Velox translates FilterNode followed by ProjectNode into a single FilterProject operator. That operator reports its stats using ProjectNode ID. This results in FilterNode not having any stats.

CPU time and memory usage are being collected and reported by the Driver which doesn't have visibility into the operator details and therefore cannot reasonably split these values between the two nodes. 

However, cardinality stats can be trivially "split". Project doesn't change cardinality, hence, all cardinality change can be attributed to Filter.

CC: @pedroerp @Yuhta @xiaoxmeng 

### System information

n/a

### Relevant logs

```bash

```

PR #12895, opened 2025-04-02
fix: Attribute cardinality change in FilterProject to Filter, not Project node

mbasmanova:
Summary:
Partial fix for https://github.com/facebookincubator/velox/issues/12893

FilterProject operator implements the logic describe by 2 plan nodes: Filter followed by Project. It reports runtime stats on Project node ID. Thus, Filter node ID has no stats.

This change introduces an option for an operator to provide a function to split stats among multiple plan nodes. FilterProject uses that mechanism to attribute cardinality change to Filter node. CPU and memory stats are kept on the Project node.

Differential Revision: D72317735

### DIFF
--- a/velox/exec/FilterProject.cpp
+++ b/velox/exec/FilterProject.cpp
@@ -38,6 +38,34 @@ bool checkAddIdentityProjection(
 
   return false;
 }
+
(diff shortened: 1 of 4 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #12897, opened 2025-04-02
fix: Impl split_to_map Null Args

duxiao1212:
Summary:
Presto  `split_to_map` supports `NULL` as EntryDelimiter and KeyValueDelimiter, we should make the behavior consistent in prestissimo 
 {F1976601386}

Differential Revision: D71905670


zacw7 (approved):
LGTM.

PR #12898, opened 2025-04-02
fix: Impl ends_with Null Args

duxiao1212:
Summary:
Presto  `ends_with` supports NULL as `X` and `Y`, we should make the behavior consistent in prestissimo 

{F1976608403}

Differential Revision: D72192579


zacw7 (approved):
LGTM. Please fix the title.

PR #12904, opened 2025-04-02
feat: Streamline toConstantSql for BIGINT and VARCHAR

mbasmanova:
Summary:
BIGINT and VARCHAR constants in Presto SQL do not require explicit type.

- 10 is the same as BIGINT '10'
- 'foo' is the same as VARCHAR 'foo'

Differential Revision: D72344105


xiaoxmeng (approved):
@mbasmanova thanks!

PR #12907, opened 2025-04-03
feat: Estimate input bytes for Filter node using cardinality change

mbasmanova:
Differential Revision: D72359615

Part of https://github.com/facebookincubator/velox/issues/12893

### DIFF
--- a/velox/exec/FilterProject.cpp
+++ b/velox/exec/FilterProject.cpp
@@ -54,8 +54,12 @@ std::vector<OperatorStats> splitStats(
   filterStats.inputBytes = projectStats.inputBytes;
   filterStats.inputPositions = projectStats.inputPositions;
 
-  // TODO Record filter output bytes.
-  filterStats.outputBytes = projectStats.outputBytes;
+  // Estimate Filter's output bytes based on cardinality change.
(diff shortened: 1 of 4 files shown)


velox/exec/FilterProject.cpp
```diff
@@ -54,8 +54,11 @@ std::vector<OperatorStats> splitStats(
   filterStats.inputBytes = projectStats.inputBytes;
   filterStats.inputPositions = projectStats.inputPositions;
 
-  // TODO Record filter output bytes.
-  filterStats.outputBytes = projectStats.outputBytes;
+  const double filterRate = projectStats.inputPositions > 0
+      ? (projectStats.outputPositions * 1.0 / projectStats.inputPositions)
+      : 1.0;
+
+  filterStats.outputBytes = filterStats.inputBytes * filterRate;
```
xiaoxmeng, line 61: So this is an estimation.

xiaoxmeng (approved):
@mbasmanova thanks!

velox/exec/FilterProject.cpp
mbasmanova: Yup.

PR #12909, opened 2025-04-03
feat: Report input/output batches for Filter node

mbasmanova:
Differential Revision: D72388036


xiaoxmeng (approved):
@mbasmanova thanks!

PR #12910, opened 2025-04-03
feat(hive): Support decimal as partition key type in Hive write and read

jinchengchenghh:
Presto does not support decimal as partition key, https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java
For Hive, it supports decimal, https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java#L754, construct BigDecimal, and use the java function BigDecimal.hashCode() to get the hashValue, https://github.com/apache/hive/blob/master/storage-api/src/java/org/apache/hadoop/hive/common/type/FastHiveDecimalImpl.java#L4106
For Spark, it aligns with hive implementation https://github.com/apache/spark/blob/master/common/unsafe/src/main/java/org/apache/spark/sql/catalyst/expressions/HiveHasher.java

### DIFF
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -32,7 +32,6 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
   uint64_t absValue =
       isNegative ? -static_cast<uint64_t>(value) : static_cast<uint64_t>(value);
 
-  // Split into 32-bit words (big-endian)
   int32_t high = static_cast<int32_t>(absValue >> 32);
   int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
 
@@ -57,7 +56,7 @@ inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
   words[2] = static_cast<int32_t>(absValue >> 32);
   words[3] = static_cast<int32_t>(absValue);
 
-  int hash = 0;
+  int32_t hash = 0;
   for (auto i = 0; i < 4; i++) {
     hash = 31 * hash + words[i];
   }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -172,7 +172,7 @@ void hashPrimitive(
         const uint32_t hash = values.isNullAt(row)
             ? 0
             : hashDecimal(
-                  values.valueAt<typename TypeTraits<kind>::NativeType>(i),
+                  values.valueAt<typename TypeTraits<kind>::NativeType>(row),
                   scale);
         mergeHash(mix, hash, hashes[row]);
       });
@@ -199,7 +199,7 @@ void hashPrimitive(
       const uint32_t hash = values.isNullAt(row)
           ? 0
           : hashOne<kind>(
-                values.valueAt<typename TypeTraits<kind>::NativeType>(i));
+                values.valueAt<typename TypeTraits<kind>::NativeType>(row));
       mergeHash(mix, hash, hashes[row]);
     });
   }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -35,7 +35,7 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
   int32_t high = static_cast<int32_t>(absValue >> 32);
   int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
 
-  unt32_t hash = 31 * high + low;
+  uint32_t hash = 31 * high + low;
   if (isNegative) {
     hash = -hash;
   }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -155,8 +155,31 @@ void hashPrimitive(
     bool mix,
     std::vector<uint32_t>& hashes) {
   const auto& type = values.base()->type();
-  const auto scale =
-      type->isDecimal() ? getDecimalPrecisionScale(*type).second : 0;
+  if (type->isDecimal()) {
+    const auto scale = getDecimalPrecisionScale(*type).second;
+    if (rows.isAllSelected()) {
+      vector_size_t numRows = rows.size();
+      for (auto i = 0; i < numRows; ++i) {
+        const uint32_t hash = values.isNullAt(i)
+            ? 0
+            : hashDecimal(
+                  values.valueAt<typename TypeTraits<kind>::NativeType>(i),
+                  scale);
+        mergeHash(mix, hash, hashes[i]);
+      }
+    } else {
+      rows.applyToSelected([&](auto row) INLINE_LAMBDA {
+        const uint32_t hash = values.isNullAt(i)
+            ? 0
+            : hashDecimal(
+                  values.valueAt<typename TypeTraits<kind>::NativeType>(i),
+                  scale);
+        mergeHash(mix, hash, hashes[row]);
+      });
+    }
+    return;
+  }
+
   if (rows.isAllSelected()) {
     // The compiler seems to be a little fickle with optimizations.
     // Although rows.applyToSelected should do roughly the same thing, doing
@@ -165,43 +188,18 @@ void hashPrimitive(
     // benchmarks.
     vector_size_t numRows = rows.size();
     for (auto i = 0; i < numRows; ++i) {
-      int32_t hash;
-      if (values.isNullAt(i)) {
-        hash = 0;
-      } else if constexpr (
-          kind == TypeKind::BIGINT || kind == TypeKind::HUGEINT) {
-        if (type->isDecimal()) {
-          hash = hashDecimal(
-              values.valueAt<typename TypeTraits<kind>::NativeType>(i), scale);
-        } else {
-          hash = hashOne<kind>(
-              values.valueAt<typename TypeTraits<kind>::NativeType>(i));
-        }
-      } else {
-        hash = hashOne<kind>(
-            values.valueAt<typename TypeTraits<kind>::NativeType>(i));
-      }
+      const uint32_t hash = values.isNullAt(i)
+          ? 0
+          : hashOne<kind>(
+                values.valueAt<typename TypeTraits<kind>::NativeType>(i));
       mergeHash(mix, hash, hashes[i]);
     }
   } else {
     rows.applyToSelected([&](auto row) INLINE_LAMBDA {
-      int32_t hash;
-      if (values.isNullAt(row)) {
-        hash = 0;
-      } else if constexpr (
-          kind == TypeKind::BIGINT || kind == TypeKind::HUGEINT) {
-        if (type->isDecimal()) {
-          hash = hashDecimal(
-              values.valueAt<typename TypeTraits<kind>::NativeType>(row),
-              scale);
-        } else {
-          hash = hashOne<kind>(
-              values.valueAt<typename TypeTraits<kind>::NativeType>(row));
-        }
-      } else {
-        hash = hashOne<kind>(
-            values.valueAt<typename TypeTraits<kind>::NativeType>(row));
-      }
+      const uint32_t hash = values.isNullAt(i)
+          ? 0
+          : hashOne<kind>(
+                values.valueAt<typename TypeTraits<kind>::NativeType>(i));
       mergeHash(mix, hash, hashes[row]);
     });
   }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -32,8 +32,8 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
   uint64_t absValue =
       isNegative ? -static_cast<uint64_t>(value) : static_cast<uint64_t>(value);
 
-  int32_t high = static_cast<int32_t>(absValue >> 32);
-  int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
+  uint32_t high = absValue >> 32;
+  uint32_t low = absValue & 0xFFFFFFFF;
 
   uint32_t hash = 31 * high + low;
   if (isNegative) {
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -169,7 +169,7 @@ void hashPrimitive(
       }
     } else {
       rows.applyToSelected([&](auto row) INLINE_LAMBDA {
-        const uint32_t hash = values.isNullAt(i)
+        const uint32_t hash = values.isNullAt(row)
             ? 0
             : hashDecimal(
                   values.valueAt<typename TypeTraits<kind>::NativeType>(i),
@@ -196,7 +196,7 @@ void hashPrimitive(
     }
   } else {
     rows.applyToSelected([&](auto row) INLINE_LAMBDA {
-      const uint32_t hash = values.isNullAt(i)
+      const uint32_t hash = values.isNullAt(row)
           ? 0
           : hashOne<kind>(
                 values.valueAt<typename TypeTraits<kind>::NativeType>(i));
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -63,7 +63,8 @@ inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
   if (isNegative) {
     hash = -hash;
   }
-  return hash * 31 + scale;
+  __builtin_mul_overflow(hash, 31, &hash);
+  return hash + scale;
 }
 
 #if defined(__has_feature)
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -63,6 +63,8 @@ inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
   if (isNegative) {
     hash = -hash;
   }
+  // Ignore the overflow to align with java implementation, only for lint error
+  // fix.
   __builtin_mul_overflow(hash, 31, &hash);
   return hash + scale;
 }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -56,17 +56,14 @@ inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
   words[2] = static_cast<int32_t>(absValue >> 32);
   words[3] = static_cast<int32_t>(absValue);
 
-  int32_t hash = 0;
+  uint32_t hash = 0;
   for (auto i = 0; i < 4; i++) {
     hash = 31 * hash + words[i];
   }
   if (isNegative) {
     hash = -hash;
   }
-  // Ignore the overflow to align with java implementation, only for lint error
-  // fix.
-  __builtin_mul_overflow(hash, 31, &hash);
-  return hash + scale;
+  return hash * 31 + scale;
 }
 
 #if defined(__has_feature)
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -33,7 +33,7 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
       isNegative ? -static_cast<uint64_t>(value) : static_cast<uint64_t>(value);
 
   uint32_t high = absValue >> 32;
-  uint32_t low = absValue & 0xFFFFFFFF;
+  uint32_t low = absValue;
 
   uint32_t hash = 31 * high + low;
   if (isNegative) {
@@ -48,13 +48,13 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
 // Returns java BigDecimal#hashCode()
 template <>
 inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
-  int32_t words[4];
+  uint32_t words[4];
   bool isNegative = value < 0;
   uint128_t absValue = isNegative ? -value : value;
-  words[0] = static_cast<int32_t>(absValue >> 96);
-  words[1] = static_cast<int32_t>(absValue >> 64);
-  words[2] = static_cast<int32_t>(absValue >> 32);
-  words[3] = static_cast<int32_t>(absValue);
+  words[0] = absValue >> 96;
+  words[1] = absValue >> 64;
+  words[2] = absValue >> 32;
+  words[3] = absValue;
 
   uint32_t hash = 0;
   for (auto i = 0; i < 4; i++) {
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -35,7 +35,7 @@ inline int32_t hashDecimal(T value, uint8_t scale) {
   int32_t high = static_cast<int32_t>(absValue >> 32);
   int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
 
-  int32_t hash = 31 * high + low;
+  unt32_t hash = 31 * high + low;
   if (isNegative) {
     hash = -hash;
   }
--- a/velox/connectors/hive/HivePartitionFunction.cpp
+++ b/velox/connectors/hive/HivePartitionFunction.cpp
@@ -26,6 +26,47 @@ int32_t hashInt64(int64_t value) {
   return ((*reinterpret_cast<uint64_t*>(&value)) >> 32) ^ value;
 }
 
+template <typename T>
+inline int32_t hashDecimal(T value, uint8_t scale) {
+  bool isNegative = value < 0;
+  uint64_t absValue =
+      isNegative ? -static_cast<uint64_t>(value) : static_cast<uint64_t>(value);
+
+  // Split into 32-bit words (big-endian)
+  int32_t high = static_cast<int32_t>(absValue >> 32);
+  int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
+
+  int32_t hash = 31 * high + low;
+  if (isNegative) {
+    hash = -hash;
+  }
+
+  return 31 * hash + scale;
+}
+
+// Simulates Hive's hashing function from Hive v1.2.1
+// org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils#hashcode()
+// Returns java BigDecimal#hashCode()
+template <>
+inline int32_t hashDecimal<int128_t>(int128_t value, uint8_t scale) {
+  int32_t words[4];
+  bool isNegative = value < 0;
+  uint128_t absValue = isNegative ? -value : value;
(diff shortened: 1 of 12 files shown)


jinchengchenghh:
Hi, @majetideepak , Can you help review this PR?

velox/connectors/hive/SplitReader.cpp
```diff
@@ -56,6 +56,30 @@ VectorPtr newConstantFromString(
         pool, size, false, type, std::move(days));
   }
 
+  if constexpr (std::is_same_v<T, int64_t> || std::is_same_v<T, int128_t>) {
+    if (type->isDecimal()) {
+      auto [precision, scale] = getDecimalPrecisionScale(*type);
+      T result;
+      const auto& str = value.value();
+      if (scale > 0) {
```
majetideepak, line 64: We should probably add a stringToDecimal helper in DecimalUtil.h

velox/type/DecimalUtil.h
```diff
@@ -384,6 +384,28 @@ class DecimalUtil {
     return writePosition - startPosition;
   }
 
+  /// @brief Convert the string to unscaled decimal.
+  /// @tparam T The type of return value.
+  /// @param str The input string that represents the decimal, e.g. 12.34.
+  /// @param scale The scale of decimal.
+  /// @return The decimal unscaled value.
+  template <typename T>
+  static T stringToDecimal(const std::string& str, uint8_t scale) {
```
rui-mo, line 393: I wonder if there's certain difference with the 'toDecimalValue' in the 'velox/expression/CastExpr-inl.h'. If no, can we reuse that one?

rui-mo: Not sure if it makes sense to keep 'toDecimalValue' in the CastExpr-inl.h and reuse it.

zhli1142015: I'm currently working on adding support for the decimal type in the `from_json` function. I realized that I might need to move the logic mentioned here from CastExpr-inl.h to DecimalUtil.h. If you can make this change in your PR, then I won't duplicate the effort. Otherwise, I'm happy to include it in my own PR.

jinchengchenghh: I think the move maybe a separate PR. @zhli1142015 If this one is really needed, I think we can use the function to replace the new added  function. Can you help to move it? @zhli1142015

jinchengchenghh: The CastExpr-inl.h is an implement header file, do we include it in other files? This may make the code mixed.

rui-mo: But I assume it would be too heavy to make SplitReader.cpp rely on the CastExpr-inl.h, so I would prefer to move it to the `DecimalUtil.h` for reuse.

jinchengchenghh: Agree.

jinchengchenghh: And I think the DecimalUtil.h is also heavy, maybe we need DecimalCastUtil.h to handle such case, if we think it should do, we can refactor it in another PR. But I'm not sure we can do that.

zhli1142015: Please see my PR, I need to use the related function in a different folder. I think it's ok to move it to DecimalUtil.h.
https://github.com/facebookincubator/velox/pull/13052

jinchengchenghh: I still have a concern, here we need to parse a decimal string partition value which is always a valid decimal with format `12.30` for decimal(10, 2), but the semantics of cast string as decimal is different with it, suppose an edge case, the partition is corrupted, we may not distinguish.

majetideepak: #13052 seems reasonable as long as the interpreted values are correct always.

> suppose an edge case, the partition is corrupted, we may not distinguish.

@jinchengchenghh Can you explain this further? What can't we distinguish? Is the expected value if the partition is corrupted NULL?

jinchengchenghh: For example, the valid partition should be `a=12.34`, user changes manually to `a=e2`(or corrupt to this one), expects it is not the expected partition, but now we recognize it. This is really rarely even not happen. Need to pay attention to the semantic is changed as https://github.com/facebookincubator/velox/pull/12910#discussion_r2047284340 describes

jinchengchenghh: I find spark also parse in this way, it can also parse `a=e2`, so now it is ok for me to use `cast string to decimal`.
```
case _: DecimalType => Literal(new JBigDecimal(value)).value
```
https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningUtils.scala#L556

majetideepak: Okay. Should we get #13052 merged first?

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -596,6 +596,68 @@ TEST_F(HiveDataSinkTest, basicBucket) {
...
+    auto path = listFiles(rootPath + "/c2=" + value)[0];
+    splits.push_back(
+        makeHiveConnectorSplits(
+            path, 1, dwio::common::FileFormat::PARQUET, partitionKeys)
+            .back());
+  };
+  partitionPath("0.0001");
+  partitionPath("0.0340");
+  partitionPath("23.4567");
+  partitionPath("-0.0345");
```
rui-mo, line 645: Would you add some tests for non-standard decimal partition value?

velox/type/DecimalUtil.h
PingLiuPing: Presto recognise `1e2` but not `e2`.

```
presto:tpch_sf10_iceberg> describe identity_t1;
   Column    |     Type      |     Extra     | Comment
-------------+---------------+---------------+---------
 c_int       | integer       | partition key |
 c_bigint    | bigint        |               |
 c_decimal   | decimal(18,3) | partition key |
 c_varchar   | varchar       |               |
 c_varbinary | varbinary     |               |
(5 rows)


presto:tpch_sf10_iceberg> insert into identity_t1 values(1, -123, cast('e2' AS decimal(18,3)), 'ABCD QWERT', x'455843454C4C454E54');
Query 20250422_151721_00013_tciim failed: Cannot cast VARCHAR 'e2' to DECIMAL(18, 3). Value is not a number.

presto:tpch_sf10_iceberg> insert into identity_t1 values(1, -123, cast('1e2' AS decimal(18,3)), 'ABCD QWERT', x'455843454C4C454E54');
INSERT: 1 row

Query 20250422_152119_00014_tciim, FINISHED, 1 node
presto:tpch_sf10_iceberg> insert into identity_t1 values(1, -123, cast('1e-2' AS decimal(18,3)), 'ABCD QWERT', x'455843454C4C454E54');
INSERT: 1 row

Query 20250422_152140_00016_tciim, FINISHED, 1 node

presto:tpch_sf10_iceberg> insert into identity_t1 values(1, -123, cast('1e0' AS decimal(18,3)), 'ABCD QWERT', x'455843454C4C454E54');
INSERT: 1 row

Query 20250422_153958_00033_tciim, FINISHED, 1 node

presto:tpch_sf10_iceberg> select * from identity_t1;
 c_int | c_bigint |   c_decimal   | c_varchar  |        c_varbinary
-------+----------+---------------+------------+----------------------------
     1 |     -123 | 1.000     | ABCD QWERT | 45 58 43 45 4c 4c 45 4e 54
     1 |     -123 | 0.010         | ABCD QWERT | 45 58 43 45 4c 4c 45 4e 54
     1 |     -123 | 123456789.124 | ABCD QWERT | 45 58 43 45 4c 4c 45 4e 54
     1 |     -123 | 100.000       | ABCD QWERT | 45 58 43 45 4c 4c 45 4e 54
```


```
➜  c_int=1 ll -rht
total 0
drwxr-xr-x@ 4 pingliu  staff   128B 22 Apr 16:10 c_decimal=123456789.124
drwxr-xr-x@ 4 pingliu  staff   128B 22 Apr 16:21 c_decimal=100.000
drwxr-xr-x@ 4 pingliu  staff   128B 22 Apr 16:21 c_decimal=1.000
drwxr-xr-x@ 4 pingliu  staff   128B 22 Apr 16:21 c_decimal=0.010
```

majetideepak: I think Presto Java is missing support for this. Let's look at what Hive and other tools support.

https://github.com/prestodb/presto/blob/0510d1d74ef87e3e52ff4b617e02db2f6e8f3b06/presto-main-base/src/main/java/com/facebook/presto/type/DecimalCasts.java#L580

jinchengchenghh:
Could you help review again? Thanks! @rui-mo @majetideepak

velox/connectors/hive/SplitReader.cpp
```diff
@@ -54,6 +54,18 @@ VectorPtr newConstantFromString(
...
+    if (type->isDecimal()) {
+      auto [precision, scale] = getDecimalPrecisionScale(*type);
+      T result;
+      const auto status = DecimalUtil::castFromString<T>(
+          StringView(value.value()), precision, scale, result);
+      VELOX_USER_CHECK(status.ok(), status.message());
+      return std::make_shared<ConstantVector<T>>(
+          pool, size, false, type, std::move(result));
+    }
+  }
```
rui-mo, line 67: Would you please add relevant tests for decimal partition table in https://github.com/facebookincubator/velox/blob/main/velox/exec/tests/TableScanTest.cpp#L1798?

rui-mo (commented):
Thanks!

velox/connectors/hive/tests/HiveDataSinkTest.cpp
jinchengchenghh: We cannot add non-standard decimal partition value here because we don't write this format partition value in velox, so as presto and spark. Spark only supports read this format partition value. I never know which engine writes scientific format partition value. If we need to add the test, we must generate the folder manually, do you think we need to add such test?

jinchengchenghh:
TableScanTest failed by 
```
[ RUN      ] TableScanTest.partitionedTableDecimalKey
E20250724 07:08:51.212296 84436 Exceptions.h:66] Line: /__w/velox/velox/velox/dwio/dwrf/writer/ColumnWriter.cpp:2161, Function:create, Expression: format == DwrfFormat::kOrc (0 vs. 1) Decimal is not supported for DWRF., Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (0 vs. 1) Decimal is not supported for DWRF.
Retriable: False
Expression: format == DwrfFormat::kOrc
Function: create
File: /__w/velox/velox/velox/dwio/dwrf/writer/ColumnWriter.cpp
Line: 2161
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox4dwrf16BaseColumnWriter6createERNS1_13WriterContextERKNS0_4dwio6common10TypeWithIdEjSt8functionIFvRNS1_12IndexBuilderEEENS1_10DwrfFormatE
# 4  _ZN8facebook5velox4dwrf16BaseColumnWriter6createERNS1_13WriterContextERKNS0_4dwio6common10TypeWithIdEjSt8functionIFvRNS1_12IndexBuilderEEENS1_10DwrfFormatE
# 5  _ZN8facebook5velox4dwrf6WriterC2ESt10unique_ptrINS0_4dwio6common8FileSinkESt14default_deleteIS6_EERKNS1_13WriterOptionsESt10shared_ptrINS0_6memory10MemoryPoolEE
# 6  _ZN8facebook5velox4dwrf6WriterC1ESt10unique_ptrINS0_4dwio6common8FileSinkESt14default_deleteIS6_EERKNS1_13WriterOptionsE
# 7  
```
Because dwrf does not support decimal type and table scan uses dwrf type to test, could I remove it?

velox/exec/tests/TableScanTest.cpp
```diff
@@ -1858,6 +1858,15 @@ TEST_F(TableScanTest, partitionedTableDoubleKey) {
   testPartitionedTable(filePath->getPath(), DOUBLE(), "3.5");
 }
 
+TEST_F(TableScanTest, partitionedTableDecimalKey) {
+  auto rowType = ROW({"c0", "c1"}, {BIGINT(), DECIMAL(20, 4)});
```
rui-mo: > https://github.com/facebookincubator/velox/pull/12910#issuecomment-3112402455

I assume you don't need to add a decimal column in the data, which would require decimal write of the DWRF format and that's unsupported. There's a `partitionValue` parameter in the `testPartitionedTableImpl` which could be used to test reading partition key.

velox/duckdb/conversion/DuckConversion.cpp
```diff
@@ -48,6 +53,22 @@ Variant decimalVariant(const Value& val) {
     case ::duckdb::PhysicalType::INT64: {
       return Variant(val.GetValueUnsafe<int64_t>());
     }
+    case ::duckdb::PhysicalType::VARCHAR: {
```
rui-mo: This issue could probably be resolved by casting string as decimal in the DuckDB query.

https://github.com/jinchengchenghh/velox/blob/5d7d1bcf266e47a4cba71bac442d78d0f4fd4971/velox/exec/tests/utils/TableScanTestBase.cpp#L174-L175


When comparing Velox to the DuckDB result, I'm not sure if using the cast-from-string implementation of Velox is a good practice, which would compare Velox to itself.

velox/connectors/hive/tests/HiveDataSinkTest.cpp
rui-mo: I see, thanks.

velox/duckdb/conversion/DuckConversion.cpp
jinchengchenghh: I try to debug the float data type, and I find it is also VARCHAR datachunk. Do I only add cast for decimal type?

jinchengchenghh: Updated, thanks!

rui-mo (approved):
Thanks! LGTM.

jinchengchenghh:
@majetideepak Would you like to take a look? Thanks!

jinchengchenghh:
Could you help take a look? Thanks! @Yuhta

jinchengchenghh:
Could you mark it as ready to merge? Thanks! @rui-mo

rui-mo:
@majetideepak @Yuhta Do you have any further comment on this change? Thanks.

velox/connectors/hive/HivePartitionFunction.cpp
```diff
@@ -26,6 +26,10 @@ int32_t hashInt64(int64_t value) {
   return ((*reinterpret_cast<uint64_t*>(&value)) >> 32) ^ value;
 }
 
+int32_t hashInt128(int128_t value) {
+  return hashInt64(static_cast<uint128_t>(value) >> 64 ^ value);
```
Yuhta: Just to confirm we have this fully identical to Hive and Presto? @jinchengchenghh

kevinwilfong:
Removing the ready-to-merge tag as it looks like there's still ongoing discussion, feel free to add it back once it's been approved by one of the Hive Connector maintainers.

velox/connectors/hive/HivePartitionFunction.cpp
jinchengchenghh: Presto does not support decimal as partition key, https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/HiveBucketing.java
For hive, it supports decimal, https://github.com/apache/hive/blob/master/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java#L754, and use a different method, construct BigDecimal, and use the java function BigDecimal.hashCode(), https://github.com/apache/hive/blob/master/storage-api/src/java/org/apache/hadoop/hive/common/type/FastHiveDecimalImpl.java#L4106
I implement hive read  in `makePartitionKeyValueString`, but meets exception in test which required hive write, do you have any suggestion? @Yuhta

jinchengchenghh: Thanks for your review, I have updated it to Hive same implementation. Could you help review again? Thanks!

velox/connectors/hive/HivePartitionFunction.cpp
```diff
@@ -122,18 +165,43 @@ void hashPrimitive(
...
-      const uint32_t hash = values.isNullAt(i)
-          ? 0
-          : hashOne<kind>(
-                values.valueAt<typename TypeTraits<kind>::NativeType>(i));
+      int32_t hash;
+      if (values.isNullAt(i)) {
+        hash = 0;
+      } else if constexpr (
+          kind == TypeKind::BIGINT || kind == TypeKind::HUGEINT) {
+        if (type->isDecimal()) {
```
Yuhta: Can we move this check outside the loop?  `BIGINT` is the top frequent partitioning column and we don't want to do this check for each row.

jinchengchenghh: Thanks! Updated, do you have further comments?

Yuhta: We need to move the `isDecimal` check outside of loop (the reason is that it's dangerous to slow down BIGINT hashing; some compiler might be able to optimize this away but it's not 100% sure):
```c++
if ((kind == TypeKind::BIGINT || kind == TypeKind::HUGEINT) && type->isDecimal()) {
  for (auto i = 0; i < numRows; ++i) {
    // ...
  }
} else {
  for (auto i = 0; i < numRows; ++i) {
    // ...
  }
}
```

velox/connectors/hive/HivePartitionFunction.cpp
```diff
@@ -122,18 +165,43 @@ void hashPrimitive(
...
   } else {
     rows.applyToSelected([&](auto row) INLINE_LAMBDA {
-      const uint32_t hash = values.isNullAt(row)
-          ? 0
-          : hashOne<kind>(
-                values.valueAt<typename TypeTraits<kind>::NativeType>(row));
+      int32_t hash;
+      if (values.isNullAt(row)) {
+        hash = 0;
+      } else if constexpr (
```
Yuhta: Some change is needed there:
```c++
if ((kind == TypeKind::BIGINT || kind == TypeKind::HUGEINT) && type->isDecimal()) {
  rows.applyToSelected([&](auto row) INLINE_LAMBDA {
    // ...
  });
} else {
  rows.applyToSelected([&](auto row) INLINE_LAMBDA {
    // ...
  });
}
```

velox/connectors/hive/tests/HiveDataSinkTest.cpp
```diff
@@ -598,6 +598,68 @@ TEST_F(HiveDataSinkTest, basicBucket) {
...
+  const auto outputDirectory = TempDirectoryPath::create();
+
+  connectorSessionProperties_->set(
+      HiveConfig::kSortWriterFinishTimeSliceLimitMsSession, "1");
+  const auto rowType =
+      ROW({"c0", "c1", "c2"}, {BIGINT(), DECIMAL(14, 3), DECIMAL(20, 4)});
+  auto dataSink = createDataSink(
+      rowType,
+      outputDirectory->getPath(),
+      dwio::common::FileFormat::PARQUET,
```
Yuhta: Can we change the tests to use DWRF?  PARQUET is not available in Meta internal CI

jinchengchenghh: Updated, thanks! Could you help merge it?

Yuhta:
@jinchengchenghh Can you fix the lint warning?

jinchengchenghh:
Fixed, thanks! @Yuhta

Yuhta:
@jinchengchenghh There are some overflow failure with UBSAN:
```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from HivePartitionFunctionTest
[ RUN      ] HivePartitionFunctionTest.longDecimal

fbcode/velox/connectors/hive/HivePartitionFunction.cpp:66:15: runtime error: signed integer overflow: -647708581 * 31 cannot be represented in type 'int'
    #0 0x7f5bf63e9c36 in int facebook::velox::connector::hive::(anonymous namespace)::hashDecimal<__int128>(__int128, unsigned char) fbcode/velox/connectors/hive/HivePartitionFunction.cpp:66
    #1 0x7f5bf63e90b7 in void facebook::velox::connector::hive::(anonymous namespace)::hashPrimitive<(facebook::velox::TypeKind)10>(facebook::velox::DecodedVector const&, facebook::velox::SelectivityVector const&, bool, std::vector<unsigned int, std::allocator<unsigned int>>&) fbcode/velox/connectors/hive/HivePartitionFunction.cpp:166
    #2 0x7f5bf63e8d9c in void facebook::velox::connector::hive::HivePartitionFunction::hashTyped<(facebook::velox::TypeKind)10>(facebook::velox::DecodedVector const&, facebook::velox::SelectivityVector const&, bool, std::vector<unsigned int, std::allocator<unsigned int>>&, unsigned long) fbcode/velox/connectors/hive/HivePartitionFunction.cpp:288
    #3 0x7f5bf6406b32 in facebook::velox::connector::hive::HivePartitionFunction::hash(facebook::velox::DecodedVector const&, facebook::velox::TypeKind, facebook::velox::SelectivityVector const&, bool, std::vector<unsigned int, std::allocator<unsigned int>>&, unsigned long)::$_0::operator()() const::'lambda'()::operator()() const fbcode/velox/connectors/hive/HivePartitionFunction.cpp:532
    #4 0x7f5bf6405d58 in facebook::velox::connector::hive::HivePartitionFunction::hash(facebook::velox::DecodedVector const&, facebook::velox::TypeKind, facebook::velox::SelectivityVector const&, bool, std::vector<unsigned int, std::allocator<unsigned int>>&, unsigned long)::$_0::operator()() const fbcode/velox/connectors/hive/HivePartitionFunction.cpp:532
    #5 0x7f5bf63f915e in facebook::velox::connector::hive::HivePartitionFunction::hash(facebook::velox::DecodedVector const&, facebook::velox::TypeKind, facebook::velox::SelectivityVector const&, bool, std::vector<unsigned int, std::allocator<unsigned int>>&, unsigned long) fbcode/velox/connectors/hive/HivePartitionFunction.cpp:532
    #6 0x7f5bf63d926f in facebook::velox::connector::hive::HivePartitionFunction::partition(facebook::velox::RowVector const&, std::vector<unsigned int, std::allocator<unsigned int>>&) fbcode/velox/connectors/hive/HivePartitionFunction.cpp:570
    #7 0x0000008c0852 in HivePartitionFunctionTest::assertPartitions(std::shared_ptr<facebook::velox::BaseVector> const&, int, std::vector<unsigned int, std::allocator<unsigned int>> const&) fbcode/velox/connectors/hive/tests/HivePartitionFunctionTest.cpp:49
    #8 0x00000088ca99 in HivePartitionFunctionTest_longDecimal_Test::TestBody() fbcode/velox/connectors/hive/tests/HivePartitionFunctionTest.cpp:161
    #9 0x7f5bf926c366 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) fbsource/src/gtest.cc:2734
    #10 0x7f5bf926be7d in testing::Test::Run() fbsource/src/gtest.cc:2751
    #11 0x7f5bf926f6d3 in testing::TestInfo::Run() fbsource/src/gtest.cc:2897
    #12 0x7f5bf927600b in testing::TestSuite::Run() fbsource/src/gtest.cc:3075
    #13 0x7f5bf92a4c1d in testing::internal::UnitTestImpl::RunAllTests() fbsource/src/gtest.cc:6066
    #14 0x7f5bf92a3702 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) fbsource/src/gtest.cc:2734
    #15 0x7f5bf92a2e55 in testing::UnitTest::Run() fbsource/src/gtest.cc:5606
    #16 0x000000889eb0 in RUN_ALL_TESTS() fbsource/gtest/gtest.h:2337
    #17 0x000000889d8e in main fbcode/velox/connectors/hive/tests/HiveDataSinkTest.cpp:1448
    #18 0x7f5bdfc2c656 in __libc_start_call_main /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #19 0x7f5bdfc2c717 in [__libc_start_main@GLIBC_2.2.5](mailto:__libc_start_main@GLIBC_2.2.5) /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/csu/../csu/libc-start.c:409:3
    #20 0x00000053ab30 in _start /home/engshare/third-party2/glibc/2.34/src/glibc-2.34/csu/../sysdeps/x86_64/start.S:116

SUMMARY: UndefinedBehaviorSanitizer: signed-integer-overflow fbcode/velox/connectors/hive/HivePartitionFunction.cpp:66:15
```

jinchengchenghh:
Could you check if this fix works? Thanks! @Yuhta

velox/connectors/hive/HivePartitionFunction.cpp
```diff
@@ -26,6 +26,49 @@ int32_t hashInt64(int64_t value) {
...
+  bool isNegative = value < 0;
+  uint128_t absValue = isNegative ? -value : value;
+  words[0] = static_cast<int32_t>(absValue >> 96);
+  words[1] = static_cast<int32_t>(absValue >> 64);
+  words[2] = static_cast<int32_t>(absValue >> 32);
+  words[3] = static_cast<int32_t>(absValue);
+
+  int32_t hash = 0;
+  for (auto i = 0; i < 4; i++) {
+    hash = 31 * hash + words[i];
```
Yuhta, line 61: We still hit the overflow here.  Changing `hash` to `uint32_t` should get rid of it

jinchengchenghh: Changed as your suggestion, and the tests can also pass. Thanks! Updated.

velox/connectors/hive/HivePartitionFunction.cpp
```diff
@@ -26,6 +26,46 @@ int32_t hashInt64(int64_t value) {
...
+template <typename T>
+inline int32_t hashDecimal(T value, uint8_t scale) {
+  bool isNegative = value < 0;
+  uint64_t absValue =
+      isNegative ? -static_cast<uint64_t>(value) : static_cast<uint64_t>(value);
+
+  int32_t high = static_cast<int32_t>(absValue >> 32);
+  int32_t low = static_cast<int32_t>(absValue & 0xFFFFFFFF);
+
+  int32_t hash = 31 * high + low;
```
Yuhta: This one also need to be changed (with `high` as well)

jinchengchenghh: Updated, thanks!

jinchengchenghh:
Could you help import again? Thanks! @Yuhta

jinchengchenghh:
Also fixed the uint32_t high, could you help import again? Thanks! @Yuhta

PR #12913, opened 2025-04-03
build(cmake): Remove Profiler from core Velox library

majetideepak:
Profiler need not be part of the core Velox library.
This also excludes it from the GlobalConfig change.


assignUser (approved):

assignUser:
Nice thanks, the test fail might be related? I can't check at the moment:)

majetideepak:
One failure was related :). Fixed it.

PR #12917, opened 2025-04-03
fix(util): Wrap between and in parantheses

peterenescu:
Summary: Wraps `between ... and` in paranthesis due to ambiguous SQL processing.

Differential Revision: D72420747


kevinwilfong (approved):
Thanks!

PR #12920, opened 2025-04-04
test(util): Create unit tests for function toTypeSql of PrestoSql

peterenescu:
Summary: Create unit tests for function toTypeSql of PrestoSql. This class generates Sql for our query runners during fuzzer runs

Differential Revision: D72428441


kagamiori (approved):
LGTM. Thanks!

PR #12923, opened 2025-04-04
fix: Fix the crash cause by NestedLoopJoinBuild sum overflow

kewang1024:
Summary: As titled

Differential Revision: D72441788


velox/exec/NestedLoopJoinBuild.h
```diff
@@ -58,9 +58,9 @@ class NestedLoopJoinBuild : public Operator {
     Operator::close();
   }
 
- private:
```
Yuhta, line 61: Should we comment this is only used outside the class for testing purpose?

Yuhta (approved):

PR #12928, opened 2025-04-04
Optimize buffer class atomics to improve performance on Arm CPUs

NikhilTarte-Arm:
Summary:
Updated Buffer's reference counting methods to use C++ defined memory orders in Buffer class:
  - addRef() now uses std::memory_order_relaxed. The atomic increment operation only needs to ensure atomicity since its primary role is to indicate that a buffer view is present.
  - refCount() now uses std::memory_order_relaxed for load. Since this method simply returns the reference count and does not synchronize dependent operations, relaxed ordering is sufficient.
  - release() now uses std::memory_order_acq_rel for fetch_sub. This means that all memory writes made before the fetch_sub call are guaranteed to complete (release semantics) before the reference count is decremented, and any thread that reads the updated count will see those writes (acquire semantics). In other words, this ordering prevents instructions from being rearranged around the fetch_sub, ensuring that resources are deallocated safely and that the buffer is not destroyed prematurely while still in use.

Background:
The default memory_order_seq_cst is overly strict for this use case. Arm CPUs have a weak memory ordering model that allows using C++ defined relaxed memory ordering constraints without sacrificing correctness. On x86 systems, which enforce strong memory ordering, these modifications have no noticeable impact.

Results:
VectorSlice benchmark shows up to 12% performance improvement, as measured on Arm Neoverse V1.

Change-Id: I812ee27bd0f9e3a440533778c704ee12273e093d


velox/buffer/Buffer.h
```diff
@@ -64,15 +64,15 @@ class Buffer {
   virtual ~Buffer() {}
 
   void addRef() {
-    referenceCount_.fetch_add(1);
+    referenceCount_.fetch_add(1, std::memory_order_relaxed);
```
Yuhta, line 67: `referenceCount_` is used for mutability check (single referenced buffer is considered mutable).  Reordering is dangerous in this use case.

Yuhta (changes requested):
I don't think these change is safe because we consider `refCount() == 1` as mutable, and allowing reordering would break this assumption.

NikhilTarte-Arm:
Thanks for addressing this.

If I understood correctly, and as you updated in the main comment, the mutability check is handled using refCount() function, and using relax in returning the refCount value could prove dangerous. I think fetch_add should be safe.

 int refCount() const {
    return referenceCount_.load(std::memory_order_relaxed); <- Using relax here seems to dangerous here
  }
What are your thoughts if we use
return referenceCount_.load(std::memory_order_acq_rel); //LINE 71

Really appreciate if you can help to provide your insights around this concept.
Thanks.

Yuhta:
@NikhilTarte-Arm Yes `unique` and `refCount` need to be strict.  `addRef` maybe we can relax it and `release` could be `memory_order_acq_rel`.  However we need complete unit tests to make sure we are not doing anything wrong here since such things are really hard to debug.

NikhilTarte-Arm:
Thanks @Yuhta 
I do agree with you. I'll make sure everything is thoroughly verified.

Will update a new patch based on that.

PR #12933, opened 2025-04-04
fix: ExpressionFuzzer never selects function signatures that return a type variable

kevinwilfong:
Summary:
I was trying to run the expression fuzzer on the newly added array_min_by/array_max_by
functions and discovered it would never select either of these functions when generating an 
expression.

I discovered that when we look up function templates as candidates, we use
`typeToExpressionList_[kTypeParameterName]`
Likewise we use kTypeParameterName when looking up a signature in 
expressionToTemplatedSignature_ with chooseRandomSignatureTemplate.

However, when we add entries to these two maps, we first sanitize the type name, which just
converts it to lower case. kTypeParameterName is "T" which will clearly never match an all
lower case string. So in practice we never recognize functions that return a type variable as a
viable candidate.

This diff update kTypeParameterName to sanitize the string we choose for it, to ensure it's
consistent with the keys in the maps we use it for lookups in.

I also noticed that when we iterate over the list of signatureTemplates_, we check if the return
type is a type variable, but again we sanitize it before doing this check.  Since simple 
functions use strings like __user_T1 for type variable names, and it's common practice to use
capital letters like T and U in Vector functions, this likewise never matches.

I updated this logic to do the check against the unsanitized type name as well.

These bugs were preventing ExpressionFuzzer from ever covering array_min_by, 
array_max_by, subscript, find_first at all, among other functions that are only being partially
covered.

Differential Revision: D72474230

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzer.h
+++ b/velox/expression/fuzzer/ExpressionFuzzer.h
@@ -299,23 +299,24 @@ class ExpressionFuzzer {
   // Returns random integer between min and max inclusive.
   int32_t rand32(int32_t min, int32_t max);
 
-  static const inline std::string kTypeParameterName = "T";
-
   const Options options_;
 
   std::vector<CallableSignature> signatures_;
   std::vector<SignatureTemplate> signatureTemplates_;
 
+  // IMPORTANT: this needs to be sanitized because ExpressionFuzzer sanitizes
+  // type names before adding them to the maps below.
+  static const inline std::string kTypeParameterName = exec::sanitizeName("T");
+
   /// Maps the base name of a return type signature to the function names that
-  /// support that return type. Base name could be "T" if the return type is a
-  /// type variable.
+  /// support that return type. Base name could be `kTypeParameterName` if the
+  /// return type is a type variable.
   std::unordered_map<std::string, std::vector<std::string>>
       typeToExpressionList_;
 
   /// Maps the base name of a *concrete* return type signature to the function
   /// names that support that return type. Those names then each further map
-  /// to a list of CallableSignature objects that they support. Base name
-  /// could be "T" if the return type is a type variable.
+  /// to a list of CallableSignature objects that they support.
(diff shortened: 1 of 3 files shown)


velox/expression/fuzzer/ExpressionFuzzer.h
```diff
@@ -299,23 +299,25 @@ class ExpressionFuzzer {
...
+  /// support that return type. Base name could be `kTypeParameterName` if the
+  /// return type is a type variable.
   std::unordered_map<std::string, std::vector<std::string>>
       typeToExpressionList_;
 
   /// Maps the base name of a *concrete* return type signature to the function
   /// names that support that return type. Those names then each further map
   /// to a list of CallableSignature objects that they support. Base name
-  /// could be "T" if the return type is a type variable.
+  /// could be `kTypeParameterName` if the return type is a type variable.
```
kagamiori, line 320: nit: Could you help check whether expressionToSignature_ can contain kTypeParameterName as a key? From the code, it seems like those keys are added to expressionToTemplatedSignature_ instead. If so, let's correct this comment.

velox/expression/fuzzer/ExpressionFuzzer.h
```diff
@@ -299,23 +299,25 @@ class ExpressionFuzzer {
...
-  static const inline std::string kTypeParameterName = "T";
-
   const Options options_;
 
   std::vector<CallableSignature> signatures_;
   std::vector<SignatureTemplate> signatureTemplates_;
 
+  // IMPORTANT: this needs to be sanitized because ExpressionFuzzer sanitizes
+  // type names before adding them to the maps below.
+  static const inline std::string kTypeParameterName = exec::sanitizeName("T");
```
kagamiori, line 309: Should we update the check at line 773 as well?

kagamiori (approved):
LGTM. Thank you for catching and fixing this bug in fuzzer to expand the test coverage!

PR #12935, opened 2025-04-04
misc: Add array_agg and arbitrary benchmarks with Streaming Aggregation

yunwu52:
Differential Revision: D72486965

### DIFF
--- a/velox/exec/benchmarks/CMakeLists.txt
+++ b/velox/exec/benchmarks/CMakeLists.txt
@@ -124,3 +124,16 @@ target_link_libraries(
   velox_vector_test_lib
   velox_window
   Folly::follybenchmark)
+
+add_executable(velox_streaming_aggregation_benchmark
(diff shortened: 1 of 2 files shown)


velox/exec/benchmarks/CMakeLists.txt
```diff
@@ -123,3 +123,11 @@ target_link_libraries(
   velox_vector_test_lib
   velox_window
   Folly::follybenchmark)
+
+add_executable(velox_streaming_aggregation_benchmark
+               StreamingAggregationBenchmark.cpp)
+
+target_link_libraries(
+  velox_streaming_aggregation_benchmark
+  velox_exec
```
assignUser, line 132: Please be explicit about your includes and don't rely on transitive includes. Looking at the header this at least needes velox_memory but probably also other targets for the planbuilder include etc.

mbasmanova (approved):
Thank you!

PR #12945, opened 2025-04-07
feat: Add percentage based min reclaim for shared arbitrator

tanjialiang:
Summary:
Changes the minimum reclaim amount to be the max of the absolute min reclaim bytes and the min reclaim percent of the current pool capacity. This is to reclaim relatively large chunk wrt total capacity to 
1) avoid more frequent reclaims and 
2) secure arbitration requestor's capacity so that the allocations do not fill the newly reclaimed memory too fast.

Differential Revision: D72431267


xiaoxmeng (approved):
@tanjialiang thanks!

PR #12946, opened 2025-04-07
fix(velox): Resolve runtime error when using ARRAY<UNKNOWN> as checksum argument

duxiao1212:
Summary:
We encountered numerous GENERIC_INTERNAL_ERROR issues originating from the error "not a known type kind: UNKNOWN Operator: Aggregation[10] 1." 
You can find more details at this link: https://fburl.com/scuba/presto_queries/zqwioxmt.

This problem arises because an UNKNOWN(aka null) is passed into the checksum hash function from ARRAY<>. To resolve this, we need to ensure that ARRAY[] defaults to returning 0(the behavior can be found in [java]( https://www.internalfb.com/code/fbsource/[1b646bfae4ec]/fbcode/github/presto-deprecated/presto-common/src/main/java/com/facebook/presto/common/type/UnknownType.java?lines=95)) before the hash function is called.

Differential Revision: D72474507

### DIFF
--- a/velox/functions/prestosql/aggregates/ChecksumAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/ChecksumAggregate.cpp
@@ -77,6 +77,17 @@ class ChecksumAggregate : public exec::Aggregate {
     }
   }
 
+  bool isNullOrNullArray(const TypePtr& type) {
+    if (type->isUnKnown()) {
+      return true;
+    }
+    // Only supports null array type for now.
+    if (type->kind() == TypeKind::ARRAY) {
+      return type->asArray().elementType()->isUnKnown();
+    }
+    return false;
+  }
+
   void addRawInput(
       char** groups,
       const SelectivityVector& rows,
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/aggregates/ChecksumAggregate.cpp
```diff
@@ -77,18 +77,36 @@ class ChecksumAggregate : public exec::Aggregate {
     }
   }
 
+  bool isNullOrNullArray(const TypePtr& type) {
+    if (type->isUnKnown()) {
+      return true;
+    }
+    // only support null array type for now.
```
zacw7, line 84: If it a sentence (ends with `.`). It should be capitalized.

```suggestion
    // Only supports null array type for now.
```

velox/functions/prestosql/aggregates/ChecksumAggregate.cpp
```diff
@@ -77,18 +77,36 @@ class ChecksumAggregate : public exec::Aggregate {
...
+    switch (type->kind()) {
```
zacw7, line 85: Let's use `if` instead of `switch` here.

zacw7 (approved):
LG % nits. Thanks for the fix!

PR #12947, opened 2025-04-07
fix(util): Update default case for SQL type generation

peterenescu:
Summary: Fix bug introduced by https://github.com/facebookincubator/velox/pull/12858 that doesn't allow for SQL generation of logical types (for example, DATE and JSON). It would be best to not special case these and minimize the switch-block logic.

Differential Revision: D72597782


kevinwilfong (approved):
Thanks! I saw you're adding unit tests for this function in https://github.com/facebookincubator/velox/pull/12920

PR #12948, opened 2025-04-07
fix(tableScan): Convert remainingFilterSubfields to member variable

kewang1024:
Summary:
remainingFilterSubfields is a local variable in the HiveDataSource constructor, and its memory will be freed after the constructor finishes executing. However, we are storing a reference to it in the member variable subfields_, which will become invalid once the constructor finishes.

This causes a problem in the setupBucketConversion() function, where we try to access the content of subfields_. Since the memory for remainingFilterSubfields has already been freed, this results in undefined behavior and ultimately leads to a crash.

Crashed Query ASAN log: P1778761122

Differential Revision: D72578859


zacw7 (approved):
Great catch thanks for the fix. Is it possible to add a unit test?

majetideepak (approved):

PR #11964, opened 2024-12-26
fix(parquet): Support converted type in timestamp reader

rui-mo:
Gluten observed a UT failure at the code location below. The cause is that the 
timestamp schema element lacks a logical type and is composed of type, name, 
and converted_type. This PR allows parsing converted type from schema elements 
and uses it in the timestamp reader to determine the timestamp precision if 
logical type is not supplied.

https://github.com/facebookincubator/velox/blob/87c558e094043c5451659c59d38ffc1bd6de781a/velox/dwio/parquet/reader/TimestampColumnReader.h#L99

PR #12953, opened 2025-04-08
fix(parquet): Fix INT64 timestamp precision conversion

liujiayi771:
In #11964, the assignment of `needsConversion_` in `TimestampColumnReader.h` has
been removed, causing `needsConversion_` to always be false. We need to update
the `needsConversion_` if `filePrecision_` is different with
`requestedPrecision_`.

### DIFF
--- a/velox/dwio/parquet/reader/TimestampColumnReader.h
+++ b/velox/dwio/parquet/reader/TimestampColumnReader.h
@@ -119,6 +119,9 @@ class TimestampColumnReader : public IntegerColumnReader {
       } else {
         VELOX_NYI("Logical type and converted type are not provided.");
       }
+      if (filePrecision_ != requestedPrecision_) {
+        needsConversion_ = true;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


liujiayi771:
@rui-mo @zuyu Could you help to review?

rui-mo (commented):
This looks like a bug. We need to update the `needsConversion_` if `filePrecision_` is different with `requestedPrecision_`.  The test `ParquetTableScanTest.timestampPrecisionMicrosecond` ensures nanos could be read as micros for int96 timestamp, while the int64 timestamp path is not guaranteed with a test.

liujiayi771:
@rui-mo I would add a test for this case.

rui-mo:
> @rui-mo I would add a test for this case.

Can you also simply the `ParquetTableScanTest.timestampPrecisionMicrosecond`? The `CursorParameters` and `addSplits` logics were added for hive config setting, but now we can set it via the `AssertQueryBuilder` as below.

```
AssertQueryBuilder(plan, duckDbQueryRunner_)
            .connectorSessionProperty()
```

rui-mo (approved):
Thanks! Looks nice.

rui-mo:
@majetideepak @Yuhta Would you like to take a look? Thanks!

PR #12955, opened 2025-04-08
build(python): Fix failure

majetideepak:


majetideepak:
We likely need more. I see a new error with this change
```
 -- Configuring done (5.9s)
  CMake Error at python/CMakeLists.txt:85 (target_link_libraries):
    Target "velox_py_plan_builder_lib" links to:

      Folly::folly

    but the target was not found.  Possible reasons include:

      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.
```

PR #12957, opened 2025-04-08
fix(util): Generate SQL for function subscript

peterenescu:
Summary:
Generate SQL for function `subscript`.

As part of Presto 0.291, Presto now supports subscripting arrays (https://prestodb.io/docs/current/functions/array.html):
```
SELECT my_array[1] AS first_element
```

Because we have not yet created a special case for subscript, and treat it like any other function, we generate as the following, which causes fuzzer failures:
```
Executing expression 0 : subscript("c0",2059751923) # Velox expression eval
Execute presto sql: SELECT subscript(c0, INTEGER '2059751923') as p0, row_number as p1 FROM (t_values) # SQL executed in Presto
Presto query failed: 1 line 1:8: Function subscript not registered
```
This review adds proper SQL generation for subscripts

Differential Revision: D72651898

### DIFF
--- a/velox/exec/fuzzer/CMakeLists.txt
+++ b/velox/exec/fuzzer/CMakeLists.txt
@@ -161,3 +161,7 @@ target_link_libraries(
 target_link_libraries(
   velox_cache_fuzzer_lib velox_dwio_common velox_temp_path
   velox_vector_test_lib)
+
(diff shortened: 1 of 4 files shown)


kagamiori (approved):
LGTM. Thanks.

assignUser (approved):

PR #12965, opened 2025-04-08
feat(parquet): Add created by config in parquet writer

jkhaliqi:
Need to change presto velox parquet-tools inspect output on parquet tools created by to contain the velox version. Also adding in a config for created by similar to arrow.

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -725,6 +725,11 @@ Each query can override the config by setting corresponding query session proper
      - integer
      - 1024
      - Batch size used when writing into Parquet through Arrow bridge.
+   * - hive.parquet.writer.created-by
+     -
+     - string
+     - parquet-cpp-velox version 0.0.0
+     - Created-by value used when writing to Parquet.
 
 ``Amazon S3 Configuration``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -164,6 +164,9 @@ std::shared_ptr<WriterProperties> getArrowParquetWriterOptions(
     properties =
         properties->data_page_version(arrow::ParquetDataPageVersion::V1);
   }
+  if (options.createdBy.has_value()) {
+    properties = properties->created_by(options.createdBy.value());
+  }
   return properties->build();
 }
 
@@ -302,6 +305,15 @@ std::optional<int64_t> getParquetBatchSize(
     }
   } catch (const folly::ConversionError& e) {
     VELOX_USER_FAIL("Invalid parquet writer batch size: {}", e.what());
+  }
+  return std::nullopt;
+}
+
+std::optional<std::string> getParquetCreatedBy(
+    const config::ConfigBase& config,
+    const char* configKey) {
+  if (config.get<std::string>(configKey).has_value()) {
+    return config.get<std::string>(configKey).value();
   }
   return std::nullopt;
 }
@@ -577,6 +589,11 @@ void WriterOptions::processConfigs(
         ? getParquetBatchSize(session, kParquetSessionWriteBatchSize)
         : getParquetBatchSize(
               connectorConfig, kParquetHiveConnectorWriteBatchSize);
+  }
+
+  if (!createdBy) {
+    createdBy =
+        getParquetCreatedBy(connectorConfig, kParquetHiveConnectorCreatedBy);
   }
 }
 
--- a/velox/dwio/parquet/writer/arrow/Properties.h
+++ b/velox/dwio/parquet/writer/arrow/Properties.h
@@ -36,6 +36,8 @@
 
 // Define the parquet created by version.
 #define CREATED_BY_VERSION "parquet-cpp-velox"
+// Velox has no versioning yet. Set default 0.0.0.
+#define VELOX_VERSION "0.0.0"
 
 namespace facebook::velox::parquet::arrow {
 
@@ -326,8 +328,7 @@ class PARQUET_EXPORT WriterProperties {
           version_(ParquetVersion::PARQUET_2_6),
           data_page_version_(ParquetDataPageVersion::V1),
           created_by_(
-              std::string("parquet-mr version 2.6.0 (build ") +
-              DEFAULT_CREATED_BY + ")"),
+              DEFAULT_CREATED_BY + std::string(" version ") + VELOX_VERSION),
           store_decimal_as_integer_(false),
           page_checksum_enabled_(false) {}
     virtual ~Builder() {}
--- a/velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
+++ b/velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
@@ -428,6 +428,9 @@ TEST(Metadata, TestAddKeyValueMetadata) {
   }
   // Verify keys that were added after file writer was closed are not present.
   EXPECT_FALSE(reader->fileMetaData().keyValueMetadataContains("test_key_4"));
+  ASSERT_EQ(
+      CREATED_BY_VERSION + std::string(" version ") + VELOX_VERSION,
+      reader->fileMetaData().createdBy());
 }
 
 // TODO: disabled as they require Arrow parquet data dir.
@@ -502,10 +505,12 @@ TEST(Metadata, TestSortingColumns) {
     sortingColumns.push_back(sortingColumn);
   }
 
+  auto createdBy = CREATED_BY_VERSION + std::string(" version 1.0");
   auto sink = CreateOutputStream();
   auto writerProps = WriterProperties::Builder()
                          .disable_dictionary()
                          ->set_sorting_columns(sortingColumns)
+                         ->created_by(createdBy)
                          ->build();
 
   EXPECT_EQ(sortingColumns, writerProps->sorting_columns());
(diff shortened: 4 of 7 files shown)


velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
```diff
@@ -261,6 +261,7 @@ TEST_F(E2EFilterTest, integerDictionary) {
 TEST_F(E2EFilterTest, timestampInt64Direct) {
   options_.enableDictionary = false;
   options_.dataPageSize = 4 * 1024;
+  options_.version = parquet::arrow::ParquetVersion::type::PARQUET_2_6;
```
majetideepak, line 264: Why do we need this change?

velox/dwio/parquet/writer/arrow/Properties.h
```diff
@@ -323,10 +323,10 @@ class PARQUET_EXPORT WriterProperties {
           write_batch_size_(DEFAULT_WRITE_BATCH_SIZE),
           max_row_group_length_(DEFAULT_MAX_ROW_GROUP_LENGTH),
           pagesize_(kDefaultDataPageSize),
-          version_(ParquetVersion::PARQUET_2_6),
+          version_(ParquetVersion::PARQUET_1_0),
           data_page_version_(ParquetDataPageVersion::V1),
           created_by_(
-              std::string("parquet-mr version 2.6.0 (build ") +
+              std::string("parquet-cpp-velox version 0.0.0 (build ") +
```
majetideepak, line 329: What is the DEFAULT_CREATED_BY?

velox/dwio/parquet/writer/arrow/Properties.h
```diff
@@ -323,10 +323,10 @@ class PARQUET_EXPORT WriterProperties {
           write_batch_size_(DEFAULT_WRITE_BATCH_SIZE),
           max_row_group_length_(DEFAULT_MAX_ROW_GROUP_LENGTH),
           pagesize_(kDefaultDataPageSize),
-          version_(ParquetVersion::PARQUET_2_6),
+          version_(ParquetVersion::PARQUET_1_0),
```
majetideepak, line 326: why this change?

velox/dwio/parquet/writer/arrow/Properties.h
jkhaliqi: parquet-cpp-velox

velox/dwio/parquet/writer/arrow/Properties.h
jkhaliqi: If the created by had to change to be similar to java output the java output is 
`Created by: parquet-mr version 1.13.1 (build db4183109d5b734ec5930d870cdae161e408ddba)`
where c++ is
`created_by: parquet-mr version 2.6.0 (build parquet-cpp-velox)`
So making that change I was assuming the version would also have to be 1 instead of 2 to match the above?

czentgr: Why is this wrong? This looks good?
```
created_by: parquet-mr version 2.6.0 (build parquet-cpp-velox)
```

Weuse thew Arrow parquet writer and use version 2.6 which enables newer features which we may or may not be using.

majetideepak: parquet-mr is the name of the old Java writer. (https://github.com/apache/parquet-java)

velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
jkhaliqi: Not needed

velox/dwio/parquet/writer/arrow/Properties.h
```diff
@@ -325,9 +325,7 @@ class PARQUET_EXPORT WriterProperties {
           pagesize_(kDefaultDataPageSize),
           version_(ParquetVersion::PARQUET_2_6),
           data_page_version_(ParquetDataPageVersion::V1),
-          created_by_(
-              std::string("parquet-mr version 2.6.0 (build ") +
-              DEFAULT_CREATED_BY + ")"),
+          created_by_(DEFAULT_CREATED_BY + std::string(" version 0.0.0")),
```
majetideepak, line 328: Let's add a config option similar to arrow.
Add a code comment: `Velox has no versioning yet. Set default 0.0.0`.

velox/docs/configs.rst
```diff
@@ -703,6 +703,11 @@ Each query can override the config by setting corresponding query session proper
      - integer
      - 1024
      - Batch size used when writing into Parquet through Arrow bridge.
+    * - hive.parquet.writer.created-by
```
majetideepak, line 706: This is not aligned properly.
You can see from "View file" that this table is now empty because of the misalignment.

velox/docs/configs.rst
```diff
@@ -703,6 +703,11 @@ Each query can override the config by setting corresponding query session proper
...
+      -
+      - string
+      - parquet-cpp-velox version 0.0.0
+      - Created by parquet velox version when writing Parquet.
```
majetideepak, line 710: Created-by value used when writing to Parquet.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -157,6 +157,8 @@ std::shared_ptr<WriterProperties> getArrowParquetWriterOptions(
     properties =
         properties->data_page_version(arrow::ParquetDataPageVersion::V1);
   }
+  properties = properties->created_by(options.createdBy.value_or(
+      CREATED_BY_VERSION + std::string(" version ") + VELOX_VERSION));
```
majetideepak, line 161: We don't need to set a default here again since arrow::WriterProperties::Builder sets the default?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -484,6 +485,58 @@ TEST_F(ParquetWriterTest, updateWriterOptionsFromHiveConfig) {
       TimestampPrecision::kMilliseconds);
 }
 
+TEST_F(ParquetWriterTest, parquetCreatedBy) {
```
majetideepak, line 488: Let's combine this with an existing metadata test or refactor to reuse.

velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
```diff
@@ -334,6 +334,18 @@ TEST(Metadata, TestV1Version) {
...
+  ASSERT_EQ(
+      CREATED_BY_VERSION + std::string(" version ") + VELOX_VERSION,
+      f_accessor->created_by());
+
+  std::shared_ptr<WriterProperties> propsCreatedBy =
+      prop_builder.created_by("parquet-cpp-velox version 1.0")->build();
+  auto fBuilderCreatedBy = FileMetaDataBuilder::Make(&schema, propsCreatedBy);
+
+  // Read the metadata
+  auto fAccessorCreatedBy = fBuilderCreatedBy->Finish();
```
majetideepak, line 346: This won't test the end-to-end plumbing. Can you add this to the config for some other test, and verify after reading the metadata from file.

jkhaliqi: Thank you, checked the default for another test after reading the file and for another test added in this config and checked after reading the metadata from the file as well.

velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
```diff
@@ -537,6 +541,8 @@ TEST(Metadata, TestSortingColumns) {
   EXPECT_EQ(sortingColumns[0].column_idx, rowGroup.sortingColumnIdx(0));
   EXPECT_EQ(sortingColumns[0].descending, rowGroup.sortingColumnDescending(0));
   EXPECT_EQ(sortingColumns[0].nulls_first, rowGroup.sortingColumnNullsFirst(0));
+  ASSERT_EQ(
```
czentgr, line 544: Can we modify another test to test the default value?
So this replaces the full string including the Velox version? So it is either the user specifies the entire thing or a default with currently `0.0.0` in it?

jkhaliqi: `TestAddKeyValueMetadata` above tests the default value where it becomes `CREATED_BY_VERSION + std::string(" version ") + VELOX_VERSION,` which is `parquet-cpp-velox version 0.0.0`. The created-by property would have to be the full string specified to be changed, so anything there would become the new created_by. Since this test we are only changing the version i'll update the code to use the previous defaults as well.

velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
```diff
@@ -503,10 +506,12 @@ TEST(Metadata, TestSortingColumns) {
...
   auto sink = CreateOutputStream();
-  auto writerProps = WriterProperties::Builder()
-                         .disable_dictionary()
-                         ->set_sorting_columns(sortingColumns)
-                         ->build();
+  auto writerProps =
+      WriterProperties::Builder()
+          .disable_dictionary()
+          ->set_sorting_columns(sortingColumns)
+          ->created_by(CREATED_BY_VERSION + std::string(" version 1.0"))
```
majetideepak, line 513: nit: create a variable to ensure an exact match at the input and output validation.
`auto createdBy = CREATED_BY_VERSION + std::string(" version 1.0")`

majetideepak (approved):
@jkhaliqi one nit.

velox/dwio/parquet/writer/arrow/tests/MetadataTest.cpp
majetideepak: Please rebase when you resolve this.

majetideepak (approved):

mbasmanova:
Seeing build failures:

```
[2157/2841] Linking CXX executable velox/buffer/tests/velox_buffer_test
FAILED: velox/buffer/tests/velox_buffer_test 

/opt/rh/gcc-toolset-12/root/usr/libexec/gcc/x86_64-redhat-linux/12/ld: lib/libvelox.so: undefined reference to `unsigned int facebook::velox::functions::remote::RemoteFunctionRequest::serializedSizeZC<apache::thrift::BinaryProtocolWriter>(apache::thrift::BinaryProtocolWriter const*) const'
collect2: error: ld returned 1 exit status
```

majetideepak:
@mbasmanova The build failure is unrelated to this PR. It is being fixed here https://github.com/facebookincubator/velox/pull/13664

PR #12975, opened 2025-04-09
feat: Add Aggregate::addRawClusteredInput and streaming_aggregation_eager_flush

Yuhta:
Summary:
Add fast path for streaming aggregation where we have input rows from same group located together.  For certain functions, we can leverage this property to reduce the number of copy calls and create larger and fewer ranges for copy.  This brings 3x improvements for a specific query shape common in data loading for AI training.

We implement this optimization for `arbitrary` and `array_agg`.  For `arbitrary`, if the input is clustered, we just keep a reference to the input vector and index that is selected; when we extract values from the container, we group all copies from same vector to one `copyRange` call so the cost is minimized.  For `array_agg`, we do similar thing, only track the range (offset and size) where the input will be taken for each group, and do the copy in bulk when we extract value.

There is another optimization to flush the streaming aggregation output whenever there is result available, via a new query config `streaming_aggregation_eager_flush`.  This allows us to minimize the memory used by accumulators.

Differential Revision: D72677410

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -531,6 +531,17 @@ class QueryConfig {
   static constexpr const char* kRequestDataSizesMaxWaitSec =
       "request_data_sizes_max_wait_sec";
 
+  /// If this is false (the default), in streaming aggregation, wait until we
+  /// have enough number of output rows to produce a batch of size specified by
+  /// Operator::outputBatchRows.
+  ///
+  /// If this is true, we put the rows in output batch, as soon as the
+  /// corresponding groups are fully aggregated.  This is useful for reducing
+  /// memory consumption, if the downstream operators are not sensitive to small
+  /// batch size.
+  static constexpr const char* kStreamingAggregationEagerFlush =
+      "streaming_aggregation_eager_flush";
+
   bool selectiveNimbleReaderEnabled() const {
     return get<bool>(kSelectiveNimbleReaderEnabled, false);
   }
@@ -974,6 +985,10 @@ class QueryConfig {
 
   bool throwExceptionOnDuplicateMapKeys() const {
     return get<bool>(kThrowExceptionOnDuplicateMapKeys, false);
+  }
+
+  bool streamingAggregationEagerFlush() const {
+    return get<bool>(kStreamingAggregationEagerFlush, false);
   }
 
   template <typename T>
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -31,14 +31,6 @@ Generic Configuration
      - integer
      - 5000
      - TableScan operator will exit getOutput() method after this many milliseconds even if it has no data to return yet. Zero means 'no time limit'.
-   * - abandon_partial_aggregation_min_rows
-     - integer
-     - 100,000
-     - Number of input rows to receive before starting to check whether to abandon partial aggregation.
-   * - abandon_partial_aggregation_min_pct
-     - integer
-     - 80
-     - Abandons partial aggregation if number of groups equals or exceeds this percentage of the number of input rows.
    * - abandon_partial_topn_row_number_min_rows
      - integer
      - 100,000
@@ -398,6 +390,34 @@ Spilling
      - integer
      - 0
      - Percentage of aggregation or join input batches that will be forced to spill for testing. 0 means no extra spilling.
+
+Aggregation
+-----------
+.. list-table::
+   :widths: 20 10 10 70
+   :header-rows: 1
+
+   * - Property Name
+     - Type
+     - Default Value
+     - Description
+   * - abandon_partial_aggregation_min_rows
+     - integer
+     - 100,000
+     - Number of input rows to receive before starting to check whether to abandon partial aggregation.
+   * - abandon_partial_aggregation_min_pct
+     - integer
+     - 80
+     - Abandons partial aggregation if number of groups equals or exceeds this percentage of the number of input rows.
+   * - streaming_aggregation_eager_flush
+     - bool
+     - false
+     - If this is false (the default), in streaming aggregation, wait until we
+       have enough number of output rows to produce a batch of size specified by
+       Operator::outputBatchRows.  If this is true, we put the rows in output
+       batch, as soon as the corresponding groups are fully aggregated.  This is
+       useful for reducing memory consumption, if the downstream operators are
+       not sensitive to small batch size.
 
 Table Scan
 ------------
--- a/velox/exec/Aggregate.h
+++ b/velox/exec/Aggregate.h
@@ -171,6 +171,35 @@ class Aggregate {
       const std::vector<VectorPtr>& args,
       bool mayPushdown) = 0;
 
+  /// Called by aggregation operator to set whether the input data is eligible
+  /// for clustered input optimization.  This is turned off, in cases for
+  /// example if the input rows from same group are not contiguous, or the
+  /// aggregate is sorted or distinct.
+  void setClusteredInput(bool value) {
+    clusteredInput_ = value;
+  }
+
+  /// Whether the function itself supports clustered input optimization.
+  ///
+  /// When this returns true, `addRawClusteredInput` should be implemented.
+  virtual bool supportsAddRawClusteredInput() const {
+    return false;
+  }
+
+  /// Fast path for the function when the input rows from same group are
+  /// clustered together. `groups`, `rows`, and `args` are the same as
+  /// `addRawInput`, `groupBoundaries` is the indices into `groups` indicating
+  /// the row after last row of each group.
+  ///
+  /// Will only be called when `supportsAddRawClusteredInput` returns true.
+  virtual void addRawClusteredInput(
+      char** /*groups*/,
+      const SelectivityVector& /*rows*/,
+      const std::vector<VectorPtr>& /*args*/,
+      const folly::Range<const vector_size_t*>& /*groupBoundaries*/) {
+    VELOX_NYI("Unimplemented: {} {}", typeid(*this).name(), __func__);
+  }
+
   // Updates final accumulators from intermediate results.
   // @param groups Pointers to the start of the group rows. These are aligned
   // with the 'args', e.g. data in the i-th row of the 'args' goes to the i-th
@@ -468,6 +497,8 @@ class Aggregate {
   std::vector<vector_size_t> pushdownCustomIndices_;
 
   bool validateIntermediateInputs_ = false;
+
+  bool clusteredInput_ = false;
 };
 
 using AggregateFunctionFactory = std::function<std::unique_ptr<Aggregate>(
--- a/velox/exec/StreamingAggregation.cpp
+++ b/velox/exec/StreamingAggregation.cpp
@@ -32,7 +32,10 @@ StreamingAggregation::StreamingAggregation(
               : "Aggregation"),
       outputBatchSize_{outputBatchRows()},
       aggregationNode_{aggregationNode},
-      step_{aggregationNode->step()} {
+      step_{aggregationNode->step()},
+      eagerFlush_{operatorCtx_->driverCtx()
+                      ->queryConfig()
+                      .streamingAggregationEagerFlush()} {
   if (aggregationNode_->ignoreNullKeys()) {
     VELOX_UNSUPPORTED(
         "Streaming aggregation doesn't support ignoring null keys yet");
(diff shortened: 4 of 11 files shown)


velox/core/QueryConfig.h
```diff
@@ -531,6 +531,17 @@ class QueryConfig {
...
 
+  /// If this is false (the default), in streaming aggregation, wait until we
+  /// have enough number of output rows to produce a batch of size specified by
+  /// Operator::outputBatchRows.
+  ///
+  /// If this is true, we put the rows in output batch, as soon as the
+  /// corresponding groups are fully aggregated.  This is useful for reducing
+  /// memory consumption, if the downstream operators are not sensitive to small
+  /// batch size.
+  static constexpr const char* kStreamingAggregationEagerFlush =
```
mbasmanova, line 542: Would you update the documentation to add this new config?

velox/exec/Aggregate.h
```diff
@@ -171,6 +171,38 @@ class Aggregate {
...
+
+  /// Called by aggregation operator to set the flag.
+  void setClusteredInput(bool value) {
+    clusteredInput_ = value;
+  }
+
+  /// Whether the function itself supports clustered input optimization.
+  ///
+  /// When this returns true, `addRawClusteredInput` should be implemented.
+  virtual bool supportAddRawClusteredInput() const {
```
mbasmanova, line 189: nit: support -> supports (to match supportsToIntermediate)

velox/type/Type.cpp
```diff
@@ -488,6 +488,14 @@ bool RowType::equals(const Type& other) const {
   return true;
 }
 
+size_t RowType::hashKind() const {
```
mbasmanova, line 491: This seems like unrelated change? Would it be possible to extract it into a separate PR and add a test?

velox/vector/BaseVector.h
```diff
@@ -430,6 +430,11 @@ class BaseVector {
     vector_size_t sourceIndex;
     vector_size_t targetIndex;
     vector_size_t count;
+
+    bool mergeable(const CopyRange& next) const {
```
mbasmanova, line 434: Would it be possible to add documentation and test for this new API?

velox/exec/StreamingAggregation.cpp
```diff
@@ -256,7 +279,14 @@ void StreamingAggregation::evaluateAggregates() {
     }
 
     if (isRawInput(step_)) {
-      function->addRawInput(inputGroups_.data(), rows, args, false);
+      if (function->clusteredInput() &&
+          function->supportAddRawClusteredInput()) {
+        VELOX_DCHECK(rows.isAllSelected());
```
mbasmanova, line 284: Does that mean we do not support masked aggregations with this optimization? Would it be possible to remove this restriction?

velox/vector/BaseVector.h
Yuhta: I will add some document.  This is probably too trivial to be tested individually.

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
```diff
@@ -258,23 +320,46 @@ class NonNumericArbitrary : public exec::Aggregate {
...
+          std::destroy_at(accumulator);
+        } else {
+          auto* accumulator = value<SingleValueAccumulator>(group);
+          accumulator->destroy(allocator_);
+        }
       }
     }
   }
+
+  void addOrMergeCopyRange(const BaseVector::CopyRange& range) {
```
rui-mo, line 351: It appears this function could be made private.

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
```diff
@@ -172,14 +180,37 @@ class NonNumericArbitrary : public exec::Aggregate {
...
+          copyRanges_.clear();
+        }
+        currentSource = &accumulator->vector;
+        addOrMergeCopyRange({accumulator->index, i, 1});
+      }
+      if (currentSource) {
+        result->get()->copyRanges(currentSource->get(), copyRanges_);
+        copyRanges_.clear();
+      }
+
```
rui-mo, line 204: nit: perhaps drop this empty line.

rui-mo (commented):
Thanks. Added two nits.

velox/exec/StreamingAggregation.cpp
```diff
@@ -215,6 +229,15 @@ void StreamingAggregation::assignGroups() {
       }
     }
   }
+
+  groupBoundaries_.clear();
+  for (vector_size_t i = 1; i < numInput; ++i) {
+    if (inputGroups_[i] != inputGroups_[i - 1]) {
+      groupBoundaries_.push_back(i);
```
mbasmanova, line 236: @Yuhta My reading of this code suggests that this comment is incorrect. It looks like 'groupBoundaries' contains index of the first row in the new group, not index of the last row in the 'current' group. Am I missing something?

> `groupBoundaries` is the indices into `groups` indicating the last row of each group.

velox/exec/StreamingAggregation.cpp
```diff
@@ -256,7 +279,14 @@ void StreamingAggregation::evaluateAggregates() {
     }
 
     if (isRawInput(step_)) {
-      function->addRawInput(inputGroups_.data(), rows, args, false);
+      if (function->clusteredInput() &&
```
mbasmanova, line 282: It is a bit strange to ask the function whether input is clustered. I believe this operator should know and the answer would be the same for all functions, no?

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
```diff
@@ -152,17 +152,25 @@ inline int32_t ArbitraryAggregate<int128_t>::accumulatorAlignmentSize() const {
...
-  // We use singleValueAccumulator to save the results for each group. This
-  // struct will allow us to save variable-width value.
   int32_t accumulatorFixedWidthSize() const override {
-    return sizeof(SingleValueAccumulator);
+    return clusteredInput_ ? sizeof(ClusteredNonNumericAccumulator)
+                           : sizeof(SingleValueAccumulator);
+  }
+
+  bool accumulatorUsesExternalMemory() const override {
+    return true;
```
mbasmanova, line 173: Is this an unrelated fix?

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
```diff
@@ -213,6 +245,35 @@ class NonNumericArbitrary : public exec::Aggregate {
...
+      const std::vector<VectorPtr>& args) override {
+    VELOX_CHECK(clusteredInput_);
+    decoded_.decode(*args[0]);
+    VELOX_DCHECK(copyRanges_.empty());
+    VELOX_DCHECK(accumulators_.empty());
+    vector_size_t groupStart = 0;
+    for (auto groupEnd : groupBoundaries) {
+      auto* accumulator =
+          value<ClusteredNonNumericAccumulator>(groups[groupEnd - 1]);
+      if (!accumulator->vector) {
```
mbasmanova, line 264: Can you help me understand this check? Why 'else' branch is missing? When would this check return 'false'?

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
```diff
@@ -258,23 +320,46 @@ class NonNumericArbitrary : public exec::Aggregate {
...
+      ++copyRanges_.back().count;
+    } else {
+      copyRanges_.push_back(range);
+    }
+  }
+
+ private:
+  DecodedVector decoded_;
+  std::vector<BaseVector::CopyRange> copyRanges_;
+  std::vector<ClusteredNonNumericAccumulator*> accumulators_;
```
mbasmanova, line 362: Would it be possible to document these variables? Where do we use accumulators_?

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
```diff
@@ -97,26 +113,74 @@ class ArrayAggAggregate : public exec::Aggregate {
...
+        vector->mutableOffsets(numGroups)->asMutable<vector_size_t>();
+    auto* resultSizes =
+        vector->mutableSizes(numGroups)->asMutable<vector_size_t>();
+    auto& elements = vector->elements();
+    vector_size_t arrayOffset = 0;
+
+    if (useClusteredAccumulator()) {
+      vector_size_t numTotalElements = 0;
+      for (int32_t i = 0; i < numGroups; ++i) {
+        auto* accumulator = value<ClusteredAccumulator>(groups[i]);
```
mbasmanova, line 126: Is it possible for accumulator to store 'null' result?

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
```diff
@@ -97,26 +113,74 @@ class ArrayAggAggregate : public exec::Aggregate {
...
-      auto& values = value<ArrayAccumulator>(groups[i])->elements;
-      auto arraySize = values.size();
-      if (arraySize) {
-        clearNull(rawNulls, i);
-
-        ValueListReader reader(values);
-        for (auto index = 0; index < arraySize; ++index) {
-          reader.next(*elements, offset + index);
+    } else {
+      elements->resize(countElements(groups, numGroups));
```
mbasmanova, line 167: nit: Would it make sense to enhance countElements to consider useClusteredAccumulator() and reuse it in both branches (or even outside of the if-else block)?

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
```diff
@@ -143,6 +208,25 @@ class ArrayAggAggregate : public exec::Aggregate {
     });
   }
 
+  bool supportAddRawClusteredInput() const override {
+    return !ignoreNulls_;
```
mbasmanova, line 212: Would it be possible to document this limitation somewhere? PR description and a comment here explaining why this is not supported.

velox/exec/StreamingAggregation.cpp
Yuhta: Yes you are correct, let me correct the comment

velox/exec/StreamingAggregation.h
```diff
@@ -83,6 +83,12 @@ class StreamingAggregation : public Operator {
   // Initialize the aggregations setting allocator and offsets.
   void initializeAggregates(uint32_t numKeys);
 
+  bool eagerFlush() const {
+    return operatorCtx_->driverCtx()
```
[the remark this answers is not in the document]
Yuhta: I would prefer method over variable unless we see performance issue.  More state is potentially more bug.

velox/exec/StreamingAggregation.cpp
mbasmanova: Can we drop function->clusteredInput() check and rely solely on function->supportAddRawClusteredInput()?

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
Yuhta: It's kind of related, the new `ClusteredNonNumericAccumulator` needs its destructor called, but `SingleValueAccumulator` actually does not, in most of the cases the memory is released along with the HSA.

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
Yuhta: Let me see if I can remove this limitation along with the one on aggregate mask

velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
mbasmanova: Wondering why would this check return false? Would this happen than a group continues into the next batch?

Yuhta: Yes if a group is crossing multiple batches

mbasmanova: Got it. Would it be possible to add a short comment to clarify?

velox/docs/configs.rst
```diff
@@ -399,6 +399,26 @@ Spilling
      - 0
      - Percentage of aggregation or join input batches that will be forced to spill for testing. 0 means no extra spilling.
 
+Aggregation
```
mbasmanova, line 402: Thank you for starting a new section. Would you move abandon_partial_aggregation_min_rows and other agg-specific config here?

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
```diff
@@ -26,13 +27,28 @@ struct ArrayAccumulator {
   ValueList elements;
 };
 
+struct SourceRange {
+  VectorPtr vector;
+  vector_size_t offset;
+  vector_size_t size;
+};
+
+struct ClusteredAccumulator {
```
mbasmanova, line 36: Would you document this accumulator?

mbasmanova (approved):
Thank you, Jimmy.

PR #12976, opened 2025-04-09
Fix get_json_object parsing errors being ignored

j7nhai:


j7nhai:
@PHILO-HE

j7nhai:
> ### Merging is blocked
> * You're not authorized to push to this branch. Visit https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches for more information.
> 
> [@j7nhai](/j7nhai)

@PHILO-HE  I think consistency is important and so as performance. Maybe we can add config for whether to use strict mode for json parsing through spark conf. It can be non-strict by default and can be turned on for strict scenarios.

j7nhai:
> @j7nhai, in the current implementation, simdjson ondemand parser is used, which only validates the JSON part that is consumed. It's different from the parser you proposed. For per. consideration, it's better to keep using ondemand parser, but it indeed can produce different result for some invalidate input as you encountered.
> 
> Is this inconsistency unacceptable on your side?

@PHILO-HE  I think consistency is very important so as performance. Maybe we can add config for whether to use strict mode for json parsing through spark conf. It can be non-strict by default and can be turned on for strict scenarios.

PHILO-HE:
@j7nhai, I prefer adding a config. Could you firstly help check simdjson doc to see the perf. diff between ondemand parser and regular parser?

j7nhai:
> @j7nhai, I prefer adding a config. Could you firstly help check simdjson doc to see the perf. diff between ondemand parser and regular parser?

Ok. I will work on it.

j7nhai:
> This pull request has been automatically marked as stale because it has not had recent activity. If you'd still like this PR merged, please comment on the PR, make sure you've addressed reviewer comments, and rebase on the latest main. Thank you for your contributions!

still working

Issue #12973, opened 2025-04-09
macOS SDK 15.4 (clang 17) issues


czentgr:
### Problem description

There are two issues after upgrading to macOS SDK 15.4 which introduces Clang 17 to macOS.

1. version of folly used results in build error due to removed traits for strings. This has been fixed in newer folly versions.
```
FAILED: CMakeFiles/folly_base.dir/folly/Fingerprint.cpp.o
[...]

PR #12980, opened 2025-04-09
build: Fix macOS SDK 15.4+ Clang17 build issues

czentgr:
This PR fixes:
- issue with older ABSL and incorrect addition of x86 flags to arm64 builds
- issue with older arrow and incorrect documentation
- instances of error: variable length arrays in C++ are a Clang extension

Partially fixes: https://github.com/facebookincubator/velox/issues/12973


assignUser (approved):
Great, thanks for the quick fix!

scripts/setup-adapters.sh
```diff
@@ -63,7 +60,11 @@ function install_gcs-sdk-cpp {
   # https://github.com/googleapis/google-cloud-cpp/blob/main/doc/packaging.md#required-libraries
 
   # abseil-cpp
+  ABSOLUTE_SCRIPTDIR=$(realpath ${SCRIPTDIR})
   github_checkout abseil/abseil-cpp 20240116.2 --depth 1
+  if [[ "$OSTYPE" == darwin* ]]; then
+    git apply "${ABSOLUTE_SCRIPTDIR}/../CMake/resolve_dependency_modules/absl/absl-macos.patch"
```
majetideepak, line 66: nit: can you add a comment for this patch. Why do we need this? Does a newer version fix this?

majetideepak (commented):
one nit. thanks @czentgr

scripts/setup-adapters.sh
czentgr: Yes, a newer version fixes this (that's where the patch comes from - the commit is in the issue). But because this version is a dependency from the gcs adapter we need to look at upgrading everything and hopefully this would then use a newer version of abseil.
All of the patches made here are fixed in newer versions (some very recently). So they can get cleaned up later once we get the versions that fix this. I believe the build will fail (git apply will fail) if it cannot apply the patch which is likely once the version is upgraded so we likely won't miss it.

majetideepak: Can we add a code comment for future reference?

CMake/resolve_dependency_modules/arrow/CMakeLists.txt
```diff
@@ -73,7 +73,8 @@ if(VELOX_ENABLE_ARROW)
     CMAKE_ARGS ${ARROW_CMAKE_ARGS}
     BUILD_BYPRODUCTS ${ARROW_LIBDIR}/libarrow.a
                      ${ARROW_LIBDIR}/libarrow_testing.a ${THRIFT_LIB}
-    PATCH_COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/thrift-download.patch)
+    PATCH_COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/thrift-download.patch &&
+                  git apply ${CMAKE_CURRENT_LIST_DIR}/arrow-documentation.patch)
```
majetideepak, line 77: @czentgr Why do we need to patch the documentation?

majetideepak: I found the reason. https://github.com/apache/arrow/pull/46012
Patches are hard to deal with. Can we add `-Wno-documentation` here instead?

czentgr: Building resulted in a documentation error. I can suppress the documentation error instead of patching it.

majetideepak (approved):
Thanks, @czentgr

czentgr:
Re-running the RowNumber fuzzer test. I created issue https://github.com/facebookincubator/velox/issues/13107 to see if this is a problem or needs a change in adding the default MemoryReclaimer (we used this for a similar problem earlier for the parquet writers).

czentgr:
@kgpai The internal linter failed. Is there anything to be done on my side to fix this?
Re-running the RowNumber fuzzer was successful.

PR #12985, opened 2025-04-09
build: Use cmake from pip for macOS

czentgr:


scripts/setup-macos.sh
```diff
@@ -87,7 +87,7 @@ function install_build_prerequisites {
     echo "Creating Python Virtual Environment at ${PYTHON_VENV}"
     python3 -m venv ${PYTHON_VENV}
   fi
-  source ${PYTHON_VENV}/bin/activate; pip3 install cmake-format regex pyyaml
+  source ${PYTHON_VENV}/bin/activate; pip3 install cmake-format regex pyyaml cmake==3.31.6
```
kgpai, line 90: So is the idea is to use install cmake under a python venv to fix this ? Not sure if that will work for a lot of our users ..

PR #12988, opened 2025-04-09
fix(util): Wrap "between ... and" in parantheses

peterenescu:
Summary:
Wraps `between ... and` in parentheses due to ambiguous SQL processing. Without this change, SQL calls like the following, generated by the fuzzer, will fail:
```
presto> SELECT (BIGINT '275307804988797837' between BIGINT '7718543791979137217' and BIGINT '5284232489947293814' < BOOLEAN 'false');
Query 20250407_214527_03957_zrrnp failed: line 1:107: mismatched input '<'. Expecting: <expression>
SELECT (BIGINT '275307804988797837' between BIGINT '7718543791979137217' and BIGINT '5284232489947293814' < BOOLEAN 'false')
```
Adding the parentheses around the generated SQL for between/and will succeed:
```
presto> SELECT ((BIGINT '275307804988797837' between BIGINT '7718543791979137217' and BIGINT '5284232489947293814') < BOOLEAN 'false');
 _col0
-------
 false
```
A full fuzzer run is detailed in the below Test section.

Differential Revision: D72751488


kagamiori (approved):
LGTM. Thanks for the fix!

PR #12989, opened 2025-04-10
fix: Fix SQL building for dereference queries

peterenescu:
Summary:
Currently the Presto expression fuzzer is broken for dereference queries and we have subsequently removed the flag `--enable_dereference` from continuous runs to stabilize the fuzzer. This issue comes from how test Presto SQL is constructed, namely two particular issues:
1. How field names are attached to their parent
2. How rows with particular field names are constructed (and then dereferenced)
---
For the former (point #1), the fuzzer is attempting to execute the following expression:
```
if("c0","c1")["row_field0"]
```
This expression aims to access `row_field0` of row `c1` if `c0` is true. Unfortunately the SQL isn't properly constructed and fails because `row_field0` is undefined: 
```
Execute presto sql: SELECT row_field0 as p0, row_number as p1 FROM (t_values)
```
This review updates the logic in both PrestoSql (toCallInputsSql function) and the visit plan node function of the PrestoQuery runner to add the missing SQL.

---

For the latter (point #2), the fuzzer attempts to build a row and immediately access a particular field, as an example:
```
CONCAT(c0)[row_field0]
```
or, in essence
```
row(c0).row_field0
```
The problem is two-fold. Firstly, the default field name is of the form `field0`, `field1`, etc., so we will need to update the field name. Secondly, the above is not possible in Presto; it will fail:
```
ROW('hello', 5).field0
```
The way to make the above work is to cast and explicitly define the field names and types:
```
select cast(row('hello', 15) as row(field0 varchar, field1 integer)).field0
```
This PR adds this functionality (see test plan for example).

Differential Revision: D72755054

### DIFF
--- a/velox/exec/tests/PlanBuilderTest.cpp
+++ b/velox/exec/tests/PlanBuilderTest.cpp
@@ -15,9 +15,12 @@
  */
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/core/Expressions.h"
 #include "velox/exec/WindowFunction.h"
 #include "velox/functions/prestosql/aggregates/RegisterAggregateFunctions.h"
 #include "velox/functions/prestosql/registration/RegistrationFunctions.h"
+#include "velox/parse/Expressions.h"
+#include "velox/parse/IExpr.h"
 #include "velox/parse/TypeResolver.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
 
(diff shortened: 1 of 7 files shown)


kagamiori (commented):
Hi @enescu, this change looks good to me. Could you add some unit tests for the changed code?

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -250,4 +254,60 @@ TEST_F(PlanBuilderTest, missingOutputType) {
...
+          .tableScan("tmp", data)
+          .projectExpressions(
+              {std::make_shared<core::FieldAccessExpr>("c0", std::nullopt)})
+          .planNode()
+          ->toString(true, true, nullptr),
+      PlanBuilder()
+          .tableScan("tmp", data)
+          .project({"c0"})
+          .planNode()
+          ->toString(true, true, nullptr)); // simple field access
```
kagamiori, line 271: nit: You can use a string as the expected value. See the examples in TEST_F(PlanBuilderTest, windowFunctionCall). Same for the check below.

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -250,4 +254,60 @@ TEST_F(PlanBuilderTest, missingOutputType) {
...
+      .projectExpressions(
+          {std::make_shared<core::FieldAccessTypedExpr>(BIGINT(), "c0")})
+      .planNode();
+  PlanBuilder()
+      .tableScan("tmp", ROW({"c0"}, {ROW({BIGINT()})}))
+      .projectExpressions({std::make_shared<core::FieldAccessTypedExpr>(
+          VARCHAR(),
+          std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0"),
+          "field0")})
+      .planNode();
```
kagamiori, line 310: Let's also check that these plans are as expected too, e.g., by checking the toString() result equals to an expected string.

kagamiori (commented):
Hi @peterenescu, this PR looks good to me except two nits. Thanks!

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -250,4 +253,50 @@ TEST_F(PlanBuilderTest, missingOutputType) {
...
+  // Non-typed Expression
+  auto data = ROW({"c0"}, {BIGINT()});
+  VELOX_CHECK_EQ(
+      PlanBuilder()
+          .tableScan("tmp", data)
+          .projectExpressions(
+              {std::make_shared<core::FieldAccessExpr>("c0", std::nullopt)})
+          .planNode()
+          ->toString(true, false),
+      "-- Project[1][expressions: (c0:BIGINT, ROW[\"c0\"])] -> c0:BIGINT\n"); // simple field access
```
kagamiori, line 266: nit: Move comment to above the code as a separate line. E.g., 
```
// Non-typed Expression. Simple field access.
auto data = ROW({"c0"}, {BIGINT()});
...
```

Same for the comment at line 279 below.

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -250,4 +253,50 @@ TEST_F(PlanBuilderTest, missingOutputType) {
...
+      PlanBuilder()
+          .tableScan("tmp", ROW({"c0"}, {ROW({BIGINT()})}))
+          .projectExpressions(
+              {std::make_shared<core::FieldAccessTypedExpr>(BIGINT(), "c0")})
+          .planNode()
+          ->toString(true, false),
+      "-- Project[1][expressions: (p0:BIGINT, \"c0\")] -> p0:BIGINT\n");
+  VELOX_CHECK_EQ(
+      PlanBuilder()
+          .tableScan("tmp", ROW({"c0"}, {ROW({BIGINT()})}))
```
kagamiori, line 292: nit: Impossible to dereference a varchar field from a ROW({BIGINT()}). Would be better to change it to ROW({VARCHAR()}).

kagamiori (approved):
LGTM. Thanks!

PR #12993, opened 2025-04-10
fix: Fix Spark try_cast and cast function error case handling

zhli1142015:
Currently, the semantics of Spark functions are implemented with ANSI mode disabled. 
Correct the corresponding behavior for:
1. Spark cast function should return null if invalid format or invalid characters in the inputs. 
2. Spark try_cast function should return null if the value overflows.

Given there are other behavioral variations besides the NULL handling between 
Spark try_cast and cast, this PR renames the `nullOnFailure` to `isTryCast`.

### DIFF
--- a/velox/core/Expressions.h
+++ b/velox/core/Expressions.h
@@ -685,7 +685,8 @@ class CastTypedExpr : public ITypedExpr {
   static TypedExprPtr create(const folly::dynamic& obj, void* context);
 
  private:
-  // Whether this expression is used for `try_cast`.
+  // Whether this expression is used for `try_cast`. When true, Presto cast
+  // suppresses exception and return null on failure to case.
   const bool isTryCast_;
 };
 
--- a/velox/core/Expressions.h
+++ b/velox/core/Expressions.h
@@ -623,18 +623,15 @@ class CastTypedExpr : public ITypedExpr {
   /// expresion.
   /// @param input Single input. The type of input is referred to as from-type
   /// and expected to be different from to-type.
-  /// @param nullOnFailure Whether to suppress cast errors and return null.
-  CastTypedExpr(
-      const TypePtr& type,
-      const TypedExprPtr& input,
-      bool nullOnFailure)
-      : ITypedExpr{type, {input}}, nullOnFailure_(nullOnFailure) {}
+  /// @param isTryCast Whether to suppress cast errors and return null.
+  CastTypedExpr(const TypePtr& type, const TypedExprPtr& input, bool isTryCast)
+      : ITypedExpr{type, {input}}, isTryCast_(isTryCast) {}
 
   CastTypedExpr(
       const TypePtr& type,
       const std::vector<TypedExprPtr>& inputs,
-      bool nullOnFailure)
-      : ITypedExpr{type, inputs}, nullOnFailure_(nullOnFailure) {
+      bool isTryCast)
+      : ITypedExpr{type, inputs}, isTryCast_(isTryCast) {
     VELOX_USER_CHECK_EQ(
         1, inputs.size(), "Cast expression requires exactly one input");
   }
@@ -643,11 +640,11 @@ class CastTypedExpr : public ITypedExpr {
       const std::unordered_map<std::string, TypedExprPtr>& mapping)
       const override {
     return std::make_shared<CastTypedExpr>(
-        type(), rewriteInputsRecursive(mapping), nullOnFailure_);
+        type(), rewriteInputsRecursive(mapping), isTryCast_);
   }
 
   std::string toString() const override {
-    if (nullOnFailure_) {
+    if (isTryCast_) {
       return fmt::format(
           "try_cast {} as {}", inputs()[0]->toString(), type()->toString());
     } else {
@@ -658,7 +655,7 @@ class CastTypedExpr : public ITypedExpr {
 
   size_t localHash() const override {
     static const size_t kBaseHash = std::hash<const char*>()("CastTypedExpr");
-    return bits::hashMix(kBaseHash, std::hash<bool>()(nullOnFailure_));
+    return bits::hashMix(kBaseHash, std::hash<bool>()(isTryCast_));
   }
 
   void accept(
@@ -672,15 +669,15 @@ class CastTypedExpr : public ITypedExpr {
     }
     if (inputs().empty()) {
       return type() == otherCast->type() && otherCast->inputs().empty() &&
-          nullOnFailure_ == otherCast->nullOnFailure();
+          isTryCast_ == otherCast->isTryCast();
     }
     return *type() == *otherCast->type() &&
         *inputs()[0] == *otherCast->inputs()[0] &&
-        nullOnFailure_ == otherCast->nullOnFailure();
+        isTryCast_ == otherCast->isTryCast();
   }
 
-  bool nullOnFailure() const {
-    return nullOnFailure_;
+  bool isTryCast() const {
+    return isTryCast_;
   }
 
   folly::dynamic serialize() const override;
@@ -689,7 +686,7 @@ class CastTypedExpr : public ITypedExpr {
 
  private:
   // Suppress exception and return null on failure to cast.
-  const bool nullOnFailure_;
+  const bool isTryCast_;
 };
 
 using CastTypedExprPtr = std::shared_ptr<const CastTypedExpr>;
--- a/velox/core/Expressions.h
+++ b/velox/core/Expressions.h
@@ -623,7 +623,7 @@ class CastTypedExpr : public ITypedExpr {
   /// expresion.
   /// @param input Single input. The type of input is referred to as from-type
   /// and expected to be different from to-type.
-  /// @param isTryCast Whether to suppress cast errors and return null.
+  /// @param isTryCast Whether this expression is used for `try_cast`.
   CastTypedExpr(const TypePtr& type, const TypedExprPtr& input, bool isTryCast)
       : ITypedExpr{type, {input}}, isTryCast_(isTryCast) {}
 
@@ -685,7 +685,7 @@ class CastTypedExpr : public ITypedExpr {
   static TypedExprPtr create(const folly::dynamic& obj, void* context);
 
  private:
-  // Suppress exception and return null on failure to cast.
+  // Whether this expression is used for `try_cast`.
   const bool isTryCast_;
 };
 
--- a/velox/docs/functions/spark/conversion.rst
+++ b/velox/docs/functions/spark/conversion.rst
@@ -5,21 +5,25 @@ Conversion Functions
 .. spark:function:: cast(value AS type) -> type
 
     Explicitly cast a ``value`` to a specified ``type``.
-    Behavior when ANSI mode is disabled:
+    Follows the behavior when Spark ANSI mode is disabled, and does not support 
+    the behavior when ANSI is turned on:
     
-    * If the ``value`` exceeds the range of the ``type``, no error is raised. Instead, the ``value`` is "wrapped" around.
+    * If the ``value`` exceeds the range of the ``type``, no error is raised. 
+      Instead, the ``value`` is "wrapped" around.
 
-    * If the ``value`` has an invalid format or contains characters incompatible with the target ``type``, the cast function returns NULL. ::
+    * If the ``value`` has an invalid format or contains characters incompatible 
+      with the target ``type``, the cast function returns NULL. ::
 
         SELECT cast(128 as tinyint); -- -128
         SELECT cast('2012-Oct-23' as date); -- NULL
 
 .. spark:function:: try_cast(value AS type) -> type
 
-    Returns the ``value`` cast to data type ``type`` if possible, or NULL if not possible.
-    try_cast's behavior is consistent regardless of the ANSI mode setting, it behaves similarly to cast with ANSI mode enabled, 
-    but returns NULL instead of throwing errors.
-    try_cast differs from cast function with ANSI mode disabled in following case: 
+    Returns the ``value`` cast to ``type`` if possible, or NULL if not possible.
+    Its behavior is independent of the ANSI mode setting, and it acts identically 
+    to cast with ANSI mode enabled but returns NULL rather than throwing errors 
+    for failure to cast.
+    ``try_cast`` differs from ``cast`` function with ANSI mode disabled in following case: 
     
     * If the ``value`` cannot fit within the domain of ``type``, the result is NULL. ::
 
--- a/velox/docs/functions/spark/conversion.rst
+++ b/velox/docs/functions/spark/conversion.rst
@@ -5,13 +5,13 @@ Conversion Functions
 .. spark:function:: cast(value AS type) -> type
 
     Explicitly cast a ``value`` to a specified ``type``.
-    Follows the behavior when Spark ANSI mode is disabled, and does not support 
+    Follows the behavior when Spark ANSI mode is disabled, and does not support
     the behavior when ANSI is turned on:
-    
-    * If the ``value`` exceeds the range of the ``type``, no error is raised. 
+
+    * If the ``value`` exceeds the range of the ``type``, no error is raised.
       Instead, the ``value`` is "wrapped" around.
 
-    * If the ``value`` has an invalid format or contains characters incompatible 
+    * If the ``value`` has an invalid format or contains characters incompatible
       with the target ``type``, the cast function returns NULL. ::
 
         SELECT cast(128 as tinyint); -- -128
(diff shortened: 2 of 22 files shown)


zhli1142015:
cc @rui-mo  and @jinchengchenghh , thanks.

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,29 @@
 Conversion Functions
 ====================
 
+Conversion Functions
```
jinchengchenghh, line 5: The name is same with higher header

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,29 @@
...
+    instead will “wrap” the result. The ``value`` with an invalid format or invalid characters for 
+    ``type`` will result in a NULL. ::
+
+      SELECT cast(128 as tinyint); -- -128
+      SELECT cast('2012-Oct-23' as date); -- NULL
+
+.. spark:function:: try_cast(value AS type) -> type
+
+    Returns the ``value`` cast to data type ``type`` if possible, or NULL if not possible.
+    try_cast differs from cast function in following cases: If the ``value`` cannot fit within
```
jinchengchenghh, line 20: try_cast differs from cast function in following cases: 

- If the ``value`` cannot fit within

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,29 @@
 Conversion Functions
 ====================
 
+Conversion Functions
+--------------------
+
+.. spark:function:: cast(value AS type) -> type
+
+    Explicitly cast a value as a type. As ANSI mode is OFF, an overflow will not cause an error but
```
jinchengchenghh, line 10: When ANSI mode is disabled, explicit type casting will allow overflow values to "wrap" around rather than causing an error.

jinchengchenghh (approved):

rui-mo (commented):
The expr compilation seems brings this confusion if the null behaviour is not the only difference between the Spark 'cast' and 'try_cast'. 

https://github.com/facebookincubator/velox/blob/946684195d45f4e2bd6dc8162091508df0c41385/velox/expression/ExprCompiler.cpp#L411

I wonder if it makes more sense to add a param 'isTryCast' for the cast expr, and keep the 'nullOnFailure' for ANSI mode. Before that we may also need to confirm Presto's behaviour. How do you think?

zhli1142015:
> The expr compilation seems brings this confusion if the null behaviour is not the only difference between the Spark 'cast' and 'try_cast'.
> 
> https://github.com/facebookincubator/velox/blob/946684195d45f4e2bd6dc8162091508df0c41385/velox/expression/ExprCompiler.cpp#L411
> 
> I wonder if it makes more sense to add a param 'isTryCast' for the cast expr, and keep the 'nullOnFailure' for ANSI mode. Before that we may also confirm Presto's behaviour. How do you think?

I believe a single flag should be sufficient, since Spark only provides two functions: `cast` and `try_cast` — there’s no such function as `cast_with_ansi`. And the flag is already used to distinguish whether the function is `cast` or `try_cast`.
Moreover, ANSI mode is controlled via configuration.
As I understand it, in Spark, `try_cast` is essentially implemented as `try { cast with ANSI mode }`.
If we need to support ANSI mode in the future, we only need to determine the behavior of the `cast` function when creating the expression based on the config: `spark.sql.ansi.enabled`.
Please let me know if this sounds ok to u.

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,30 @@
...
+
+        SELECT cast(128 as tinyint); -- -128
+        SELECT cast('2012-Oct-23' as date); -- NULL
+
+.. spark:function:: try_cast(value AS type) -> type
+
+    Returns the ``value`` cast to data type ``type`` if possible, or NULL if not possible.
+    try_cast differs from cast function in following case: 
+    
+    * If the ``value`` cannot fit within the domain of ``type``, the result is NULL. ::
```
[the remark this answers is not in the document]
zhli1142015: As I understand it, Spark's try_cast is implemented as `try { cast with ANSI mode on }`.

In Spark's implementation, only the following type conversions have different execution paths when ANSI mode is enabled.
```
string --> timestamp,
double/Float --> timestamp, 
string --> date,
string --> decimal,
string --> double,
string --> float,
string --> bool,
string --> long,
numeric --> long,
string --> int,
numeric --> int,
string --> short,
numeric --> short,
string --> byte,
numeric --> byte
```
For try_cast, only these conversions behave differently, and all of them are related to overflow handling when casting as integral types. As there is only one behavior differece, i think it's ok to list it here. Please let me know if this is ok to you.
```
string --> long,
numeric --> long,
string --> int,
numeric --> int,
string --> short,
numeric --> short,
string --> byte,
numeric --> byte
```

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,30 @@
...
+    Behavior when ANSI mode is disabled:
+    
+    * If the ``value`` exceeds the range of the ``type``, no error is raised. Instead, the ``value`` is "wrapped" around.
+
+    * If the ``value`` has an invalid format or contains characters incompatible with the target ``type``, the cast function returns NULL. ::
+
+        SELECT cast(128 as tinyint); -- -128
+        SELECT cast('2012-Oct-23' as date); -- NULL
+
+.. spark:function:: try_cast(value AS type) -> type
```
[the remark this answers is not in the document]
zhli1142015: Updated, thanks.

rui-mo:
@zhli1142015 I assume the problem here is, as you mentioned, in some case 'cast' returns a value while the 'try_cast' expects NULL, which is not like ANSI that controls whether to throw of return NULL. So to use the `nullOnFailure()` to choose between them might not comply to the actual semantics.

zhli1142015:
> So to use the nullOnFailure() to choose between them might not comply to the actual semantics.

I think it's just the name of the flag that's confusing—perhaps renaming it to `isTryCast` would be more reasonable.
However, I’m not sure if this change is really necessary, since it would mean changing the contract.
Do you have any suggestions?

rui-mo:
> I think it's just the name of the flag that's confusing—perhaps renaming it to isTryCast would be more reasonable.

@zhli1142015 I feel we may change the name to avoid the confusion that this flag can be used for the ANSI support.

velox/docs/functions/spark/conversion.rst
```diff
@@ -74,13 +102,13 @@ Invalid examples
...
 ::
 
-  SELECT cast('1234567' as tinyint); -- Out of range
-  SELECT cast('1a' as tinyint); -- Invalid argument
-  SELECT cast('' as tinyint); -- Invalid argument
-  SELECT cast('1,234,567' as bigint); -- Invalid argument
-  SELECT cast('1'234'567' as bigint); -- Invalid argument
-  SELECT cast('nan' as bigint); -- Invalid argument
-  SELECT cast('infinity' as bigint); -- Invalid argument
+  SELECT cast('1234567' as tinyint); -- NULL
```
rui-mo, line 105: Can we keep the reason of NULL by using result like `NULL (Out of range)`?

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,34 @@
...
+
+    * If the ``value`` has an invalid format or contains characters incompatible with the target ``type``, the cast function returns NULL. ::
+
+        SELECT cast(128 as tinyint); -- -128
+        SELECT cast('2012-Oct-23' as date); -- NULL
+
+.. spark:function:: try_cast(value AS type) -> type
+
+    Returns the ``value`` cast to data type ``type`` if possible, or NULL if not possible.
+    try_cast's behavior is consistent regardless of the ANSI mode setting, it behaves similarly to cast with ANSI mode enabled,
```
rui-mo, line 20: > it behaves similarly to cast with ANSI mode enabled

In wonder what is Spark's behavior when the ``value`` exceeds the range of the ``type`` when ANSI is enabled. Does it align with the try_cast except that it throws error?

velox/core/Expressions.h
```diff
@@ -625,18 +625,15 @@ class CastTypedExpr : public ITypedExpr {
   /// expresion.
   /// @param input Single input. The type of input is referred to as from-type
   /// and expected to be different from to-type.
-  /// @param nullOnFailure Whether to suppress cast errors and return null.
-  CastTypedExpr(
-      const TypePtr& type,
-      const TypedExprPtr& input,
-      bool nullOnFailure)
-      : ITypedExpr{type, {input}}, nullOnFailure_(nullOnFailure) {}
+  /// @param isTryCast Whether to suppress cast errors and return null.
```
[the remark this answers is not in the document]
zhli1142015: Updated, thanks.

velox/docs/functions/spark/conversion.rst
zhli1142015: I think try_cast behaviors `try {cast with ansi mode} catch {null}`, please see the spark logic below:

When `evalMode == EvalMode.TRY`, `ansiEnabled` = true.
https://github.com/apache/spark/blob/aa6b5ed99ee33122bd638936207c99e1db4ad210/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L497

When `isTryCast == true`, try cacth block is added.
https://github.com/apache/spark/blob/aa6b5ed99ee33122bd638936207c99e1db4ad210/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L1183

Below there is special handling for converting from string type to FractionalType and DatetimeType. This is done solely for performance reasons, and the semantics remain the same as described above.
https://github.com/apache/spark/blob/aa6b5ed99ee33122bd638936207c99e1db4ad210/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L1194

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,34 @@
 Conversion Functions
 ====================
 
+.. spark:function:: cast(value AS type) -> type
+
+    Explicitly cast a ``value`` to a specified ``type``.
+    Behavior when ANSI mode is disabled:
```
rui-mo, line 8: > Behavior when ANSI mode is disabled:

Comment suggestion:

Follows the behavior when Spark ANSI mode is disabled, and does not support the behavior when ANSI is turned on.

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,34 @@
...
+    * If the ``value`` exceeds the range of the ``type``, no error is raised. Instead, the ``value`` is "wrapped" around.
+
+    * If the ``value`` has an invalid format or contains characters incompatible with the target ``type``, the cast function returns NULL. ::
+
+        SELECT cast(128 as tinyint); -- -128
+        SELECT cast('2012-Oct-23' as date); -- NULL
+
+.. spark:function:: try_cast(value AS type) -> type
+
+    Returns the ``value`` cast to data type ``type`` if possible, or NULL if not possible.
```
rui-mo, line 19: nit: remove "data type" for readability.

velox/docs/functions/spark/conversion.rst
rui-mo, line 20: Comment suggestion:

> Its behavior is independent of the ANSI mode setting, and it acts identically to cast with ANSI mode enabled but returns NULL rather than throwing errors for failure to cast.

velox/docs/functions/spark/conversion.rst
```diff
@@ -2,6 +2,34 @@
...
+    try_cast's behavior is consistent regardless of the ANSI mode setting, it behaves similarly to cast with ANSI mode enabled, 
+    but returns NULL instead of throwing errors.
+    try_cast differs from cast function with ANSI mode disabled in following case:
```
rui-mo, line 22: nit: perhaps use a reference to the function name `try_cast` and `cast`.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -32,8 +32,8 @@ exec::ExprPtr SparkCastCallToSpecialForm::constructSpecialForm(
       type,
       std::move(compiledChildren[0]),
       trackCpuUsage,
-      false,
-      std::make_shared<SparkCastHooks>(config));
+      true,
+      std::make_shared<SparkCastHooks>(config, false));
```
rui-mo, line 36: Why we are passing true to the SparkCastExpr as the isTryCast?

rui-mo (commented):
Thanks. Added some comments on the doc.

velox/type/Conversions.h
```diff
@@ -46,6 +46,12 @@ struct SparkCastPolicy {
   static constexpr bool throwOnUnicode = false;
 };
 
+struct SparkTryCastPolicy {
+  static constexpr bool truncate = false;
```
rui-mo, line 50: How is the truncate flag utilized for the Spark try_cast policy?

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
zhli1142015: Spark’s cast function (with ANSI mode disabled) and its try_cast function must both return NULL when given invalid input, which is why isTryCast needs to be set to true. The difference between them is how they treat overflow: the regular cast does not consider overflow as invalid input, whereas try_cast does treat overflow as illegal input. This behavior is determined by the specific cast policy in use.

velox/type/Conversions.h
zhli1142015: This truncate flag is used to decide whether an overflow is treated as a failure or is wrapped.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
rui-mo: It appears to be a bit conflicting as we are setting `isTryCast` in CastExpr as true, but in SparkCastHooks as false. The reason could be that Spark's try_cast and Velox's TryCast have distinct semantics.

> The difference between them is how they treat overflow: the regular cast does not consider overflow as invalid input, whereas try_cast does treat overflow as illegal input. This behavior is determined by the specific cast policy in use.

Would it make sense to rename the Spark try-cast parameter as allowOverflow?

zhli1142015: I renamed the isTryCast parameter in SparkCastHooks to allowOverflow. Since under the current semantics isTryCast is always true, I removed this parameter from the SparkCastExpr constructor and added a comment explaining the change. I thought the confusion stems from the semantics of Spark’s cast (ANSI off) functions, and there may not be a cleaner mapping from spark cast to velox cast.

velox/docs/functions/spark/conversion.rst
```diff
@@ -74,13 +106,13 @@ Invalid examples
...
 ::
 
-  SELECT cast('1234567' as tinyint); -- Out of range
-  SELECT cast('1a' as tinyint); -- Invalid argument
-  SELECT cast('' as tinyint); -- Invalid argument
-  SELECT cast('1,234,567' as bigint); -- Invalid argument
-  SELECT cast('1'234'567' as bigint); -- Invalid argument
-  SELECT cast('nan' as bigint); -- Invalid argument
-  SELECT cast('infinity' as bigint); -- Invalid argument
+  SELECT cast('1234567' as tinyint); -- NULL(Out of range)
```
jinchengchenghh, line 109: Please describe the NULL is cast while the content in () is for ANSI mode off try_cast

rui-mo (approved):
Thanks for iterating! Just two nits.

rui-mo:
@Yuhta Would you like to take a look? Thanks.

zhli1142015:
@Yuhta , could you please help to review this PR?
Thanks.

rui-mo:
Hi @Yuhta @xiaoxmeng, this PR introduces a rename from "nullOnFailure" to "isTryCast". Would you like to take a look? Thanks!

zhli1142015:
> @zhli1142015, as you know, we Gluten community have a long-term plan to support ANSI mode. I am curious whether the proposed changes cause any conflicts on future ANSI support. `nullOnFailure=true` may partially align with ANSI off mode. If `nullOnFailure` is replaced by `isTryCast`, how will we distinguish ANSI on and ANSI off for regular cast ( not try-cast)?

As I understand it, `nullOnFailure` here is only used to distinguish try_cast from cast. Whether ANSI mode is enabled affects only the behavior of cast, not try_cast, because try_cast is implemented as: `try { cast with ANSI mode }  catch { return null } ` in Spark.
If we need to support ANSI mode, we may pass that parameter to cast function via query config, just as Spark does.

rui-mo:
> If nullOnFailure is replaced by isTryCast, how will we distinguish ANSI on and ANSI off for regular cast ( not try-cast)?

@PHILO-HE The point proposed in this PR, from my understanding, is that the previous `nullOnFailure` flag could not be used to distinguish the ANSI ON and OFF, because there are other behavioral variations besides the NULL handling. Given this, we rename this flag from `nullOnFailure` to `isTryCast`, and as @zhli1142015 mentioned, to support ANSI we will need an extra config.

rui-mo:
@zhli1142015 Could you rebase this PR? We may mark it as ready-to-merge if no further comment from the community. Thanks.

zhli1142015:
> @zhli1142015 Could you rebase this PR? We may mark it as ready-to-merge if no further comment from the community. Thanks.

Done, thanks.

Issue #12542, opened 2025-03-05
Result mismatch with Spark in statistical aggregation function when DivideByZero occurs during expression evaluation


NEUpanning:
### Bug description

In Spark, if the configuration setting `spark.sql.legacy.statisticalAggregate` is set to `true`, statistical aggregation function will return `Double.NaN` instead of `NULL` when DivideByZero occurs during expression evaluation. For example, stddev(2.0) will return `NaN` in Spark but return `NULL` in Velox. Statistical aggregation function includes stddev, stddev_samp, variance, var_samp, skewness, kurtosis, covar_samp, corr.

Spark PR: https://github.com/apache/spark/pull/29983

### System information

/

### Relevant logs

```bash

```

NEUpanning:
This legacy configuration is fully supported, see https://facebookincubator.github.io/velox/configs.html#spark-specific-configuration. Thank you so much for helping to review @rui-mo

PR #12994, opened 2025-04-10
feat: Support Spark legacy behavior for covariance functions

NEUpanning:
In Spark corr and covar_samp functions, the result should be Double.NaN
instead of NULL if spark.legacy_statistical_aggregate is set to true when
dividing by zero during expression evaluation.

Therefore, it is necessary to move CovarianceAggregate to the functions/lib
directory for reuse by both Spark and Presto. Spark and Presto can then
implement their own TAccumulator, TIntermediateInput, TIntermediateResult
and TResultAccessor.

Spark corr:
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Corr.scala#L116
Spark variance:
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Covariance.scala#L129

Part of: https://github.com/facebookincubator/velox/issues/12542


NEUpanning:
@rui-mo Could you help to take a look? Thanks.

NEUpanning:
@rui-mo This is the last PR for https://github.com/facebookincubator/velox/issues/12542. Would you help review it? Thanks!

rui-mo:
@NEUpanning Thanks for your efforts. I will take a look.

velox/functions/lib/aggregates/CovarianceAggregatesBase.h
```diff
@@ -0,0 +1,392 @@
...
+ */
+#pragma once
+
+#include "velox/exec/Aggregate.h"
+#include "velox/vector/DecodedVector.h"
+#include "velox/vector/FlatVector.h"
+
+namespace facebook::velox::functions::aggregate {
+template <typename T>
+SimpleVector<T>* asSimpleVector(
```
rui-mo, line 24: Consider moving it to `namespace detail` to avoid exposing it as an API. Ditto for the other methods, structs etc. not intended to be exposed as API.

velox/functions/lib/aggregates/CovarianceAggregatesBase.h
```diff
@@ -0,0 +1,392 @@
...
+    const RowVector* rowVector,
+    int32_t childIndex) {
+  return rowVector->childAt(childIndex)->as<SimpleVector<T>>();
+}
+
+template <typename T>
+T* mutableRawValues(const RowVector* rowVector, int32_t childIndex) {
```
rui-mo, line 31: ditto

velox/functions/lib/aggregates/CovarianceAggregatesBase.h
```diff
@@ -0,0 +1,392 @@
...
+  int32_t count;
+  int32_t meanX;
+  int32_t meanY;
+  int32_t c2;
+};
+constexpr CovarIndices kCovarIndices{1, 2, 3, 0};
+
+// Indices into RowType representing intermediate results of corr. Columns
+// appear in alphabetical order.
+struct CorrIndices : public CovarIndices {
```
rui-mo, line 49: ditto

rui-mo (commented):
Thanks. Added some initial comments.

velox/functions/lib/aggregates/CovarianceAggregatesBase.h
NEUpanning: Thanks. Updated.[](url)

rui-mo (commented):
Thanks. Just several nits.

velox/functions/lib/aggregates/CovarianceAggregatesBase.h
rui-mo: I see, thanks.

rui-mo:
cc: @Yuhta @kagamiori  If you'd like to take a look, thanks.

rui-mo (approved):

NEUpanning:
@rui-mo Can you help to merge this PR? Thanks!

PR #12999, opened 2025-04-10
fix: Optimize RowType::hashKind

Yuhta:
Summary:
Avoid recurively rehash for `RowType`.  This reduces the expression
compilation time significantly for flatmap types commonly seen in ML workload.

Differential Revision: D72803509


mbasmanova (approved):
@Yuhta Are we caching hash for ROW type? Any reason not to compute it in the ctor?

Yuhta:
@mbasmanova Not sure how often this is used.  Seems only used for expression deduplication so we don't need to pay for this at runtime.

PR #13000, opened 2025-04-10
fix: Remove public from VELOX_DEFINE_CLASS_NAME

amberMZ:
Summary:
When this macro is used in the middle of the class definition, the `public` specifier will make everything below also public, which is unintended. This diff removes `public` specifier so that the macro doesn't change variable visibility for the rest of the class.

Because member visibility will changed, this diff also moves VELOX_DEFINE_CLASS_NAME into public.

Reviewed By: mbasmanova

Differential Revision: D72421576

### DIFF
--- a/velox/common/base/ClassName.h
+++ b/velox/common/base/ClassName.h
@@ -16,7 +16,6 @@
 
 #pragma once
 #define VELOX_DEFINE_CLASS_NAME(name)          \
- public:                                       \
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
Thanks.

pedroerp:
You need to fix the PR title pattern to get rid of the CI failure.

PR #13004, opened 2025-04-11
feat: Support intergral types for the index parameter of Spark get function

zhli1142015:
Support tinyint and smallint types for the index parameter of Spark get function.

Spark's implementation: https://github.com/apache/spark/blob/9bc8cd51b671aeab7f4652698c00a72991e216fb/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/complexTypeExtractors.scala#L248


zhli1142015:
cc @rui-mo  and @jinchengchenghh , thanks

jinchengchenghh (approved):
Thanks~

PR #13013, opened 2025-04-11
misc: Avoid acquiring lock in QueryCtx::checkUnderArbitration if unnecessary

kagamiori:
Summary: When the number of Driver is large, the lock in QueryCtx::checkUnderArbitration() can cause lock contention. This change makes checkUnderArbitration() acquire the lock only if underArbitration_ is true.

Differential Revision: D72753248

### DIFF
--- a/velox/core/QueryCtx.cpp
+++ b/velox/core/QueryCtx.cpp
@@ -123,7 +123,12 @@ uint64_t QueryCtx::MemoryReclaimer::reclaim(
 
 bool QueryCtx::checkUnderArbitration(ContinueFuture* future) {
   VELOX_CHECK_NOT_NULL(future);
+  if (!underArbitration_) {
+    return false;
+  }
+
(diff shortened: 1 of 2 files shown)


velox/core/QueryCtx.cpp
```diff
@@ -123,9 +123,13 @@ uint64_t QueryCtx::MemoryReclaimer::reclaim(
...
 bool QueryCtx::checkUnderArbitration(ContinueFuture* future) {
   VELOX_CHECK_NOT_NULL(future);
+  if (!underArbitration_) {
+    return false;
+  }
+
   std::lock_guard<std::mutex> l(mutex_);
+  // Check again under the lock to avoid data race.
   if (!underArbitration_) {
-    VELOX_CHECK(arbitrationPromises_.empty());
```
xiaoxmeng, line 128: Can you keep the check here?

xiaoxmeng (approved):

PR #13020, opened 2025-04-14
misc: Refactor table scan and add batch size metric for monitoring

xiaoxmeng:
Summary:
Current table scan get output is a big function handles get split, add split, get next batch and a couple of edge case condition
handing. This diff simplify the logic by moving the get split into a separate function to ease extension for feature like task barrier.
Also remove cur stats monitoring logic which is a step by step debugging message which is not required as the code becomes
mature. 
This PR also adds batch size metric for monitoring in non-presto use case.

Differential Revision: D72946387

### DIFF
--- a/velox/exec/TableScan.h
+++ b/velox/exec/TableScan.h
@@ -28,8 +28,6 @@ class TableScan : public SourceOperator {
       DriverCtx* driverCtx,
       const std::shared_ptr<const core::TableScanNode>& tableScanNode);
 
-  folly::dynamic toJson() const override;
-
   RowVectorPtr getOutput() override;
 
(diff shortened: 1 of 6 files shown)


velox/exec/TableScan.h
```diff
@@ -74,6 +72,9 @@ class TableScan : public SourceOperator {
   // terminated.
   bool shouldStop(StopReason taskStopReason) const;
 
+  // Returns true if a new split is fetched from the task otherwise false.
+  bool getSplit();
```
tanjialiang, line 76: s/getSplit/fetchSplit/

tanjialiang (approved):

PR #13021, opened 2025-04-14
fix: Fix merge source promise setting out of lock

xiaoxmeng:
Summary:
The current merge source signal promises within merge source lock which is prone to the potential deadlock if the inline continuation
tries to access the merge source lock again. This PR moves this out of the lock.

Differential Revision: D72946526


JkSelf (approved):
LGTM.

Yuhta (approved):

PR #11772, opened 2024-12-06
fix: Fix the MergeSource data lost issue

JkSelf:
When running the TPCH Q21 query, we found that performing a left semi join followed by a left anti join in the same stage resulted in incorrect results. Upon investigation, we discovered that MergeSource was losing data in such complex join scenarios. This PR addresses the issue by placing the data from MergeSource into a queue, ensuring that the next method call only ends when the queue is empty.

PR #11771, opened 2024-12-06
fix: Fix semi join and anti join result mismatch issue

JkSelf:

PR #13039, opened 2025-04-16
fix: MergeJoin is buggy with RIGHT OUTER JOINs where NULLs are present in the keys

kevinwilfong:
Summary:
While working on another change I discovered that JoinFuzzer does not test MergeJoin with
RIGHT OUTER JOINs. Enabling it locally (I'll publish a change to enable it separately) I
discovered a bug. While trying to add a unit test for it, I discovered a few more.

They fit into two classes:
1) Skipping over rows on the right side with NULL keys. This is the correct thing to do for 
INNER and LEFT OUTER JOINs but we need to output misses for these rows in RIGHT OUTER
JOINs (they can't hit given our NULL semantics).
2) Writing off the end of the output buffer trying to write out this misses. We need to make
sure the size of output_ hasn't yet reached outputBatchSize_ before writing misses to it.

This diff fixes the bugs I found and adds unit tests covering NULL keys (I didn't see any prior
to this change).

Differential Revision: D73077550

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -312,12 +312,14 @@ inline void addNull(
   target->setNull(index, true);
 }
 
-void MergeJoin::addOutputRowForLeftJoin(
-    const RowVectorPtr& leftBatch,
-    vector_size_t leftRow) {
+bool MergeJoin::tryAddOutputRowForLeftJoin() {
   VELOX_USER_CHECK(
       isLeftJoin(joinType_) || isAntiJoin(joinType_) || isFullJoin(joinType_));
-  rawLeftOutputIndices_[outputSize_] = leftRow;
+  if (outputSize_ == outputBatchSize_) {
+    return false;
+  }
+
+  rawLeftOutputIndices_[outputSize_] = leftRowIndex_++;
 
   for (const auto& projection : rightProjections_) {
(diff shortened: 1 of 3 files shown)


xiaoxmeng (approved):

zhouyuan:
@kevinwilfong @xiaoxmeng would you please also help to review below two merge join fixes? There were discovered in Spark/Gluten tests and fixed by @JkSelf 
https://github.com/facebookincubator/velox/pull/11771
https://github.com/facebookincubator/velox/pull/11772

Thanks, -yuan

pedroerp:
Right outer join is indeed not very stable. We discussed that ideally we should only implement a left outer join, and for right outer join we just swap the inputs and run the same left outer join path. So at least we would have a single code path to test.

JkSelf:
@pedroerp @xiaoxmeng @kevinwilfong 
The internal testing for Gluten has already enabled sort merge join and incorporated these two PRs https://github.com/facebookincubator/velox/pull/11771
https://github.com/facebookincubator/velox/pull/11772 into our internal code branch https://github.com/oap-project/velox/commit/588820905fe6633a7b0584a4380879680193f327. Recently, there have been numerous refactors and fixes related to merge join in the community, which have caused issues with our internal testing and rebasing. Therefore, if possible, could you please prioritize reviewing these two PRs? Thanks for your help.

PR #13041, opened 2025-04-16
feat: Enable the hash join to accept a pre-built hash table for joining

JkSelf:
In Spark, the hash table is constructed only once in the driver and then broadcasted to each executor. In Gluten, we replace the broadcast hash join with a hash join, which leads to performance issues when the broadcast threshold increases. This is because each task must build its own hash table when using hash join.

This PR enables `HashBuild `to accept pre-built `HashTable`, thereby bypassing the hash table construction process. Additionally, it modifies `HashProbe ` to avoid clearing the shared hash table after use.

### DIFF
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -3107,7 +3107,8 @@ class HashJoinNode : public AbstractJoinNode {
       TypedExprPtr filter,
       PlanNodePtr left,
       PlanNodePtr right,
-      RowTypePtr outputType)
+      RowTypePtr outputType,
+      void* reusedHashTableAddress = nullptr)
       : AbstractJoinNode(
             id,
             joinType,
@@ -3117,9 +3118,9 @@ class HashJoinNode : public AbstractJoinNode {
             std::move(left),
             std::move(right),
             std::move(outputType)),
-        nullAware_{nullAware} {
+        nullAware_{nullAware},
+        reusedHashTableAddress_(reusedHashTableAddress) {
     validate();
-
     if (nullAware) {
       VELOX_USER_CHECK(
           isNullAwareSupported(joinType),
@@ -3202,6 +3203,10 @@ class HashJoinNode : public AbstractJoinNode {
     return nullAware_;
   }
 
+  void* reusedHashTableAddress() const {
+    return reusedHashTableAddress_;
+  }
+
   folly::dynamic serialize() const override;
 
   static PlanNodePtr create(const folly::dynamic& obj, void* context);
@@ -3210,6 +3215,8 @@ class HashJoinNode : public AbstractJoinNode {
   void addDetails(std::stringstream& stream) const override;
 
   const bool nullAware_;
+
+  void* reusedHashTableAddress_;
 };
 
 using HashJoinNodePtr = std::shared_ptr<const HashJoinNode>;
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -34,6 +34,21 @@ class PlanNodeVisitorContext;
 
 using PlanNodeId = std::string;
 
+struct ReusedHashTableInfo {
+ public:
+  ReusedHashTableInfo() {}
+
+  ReusedHashTableInfo(bool joinHasNullKeys)
+      : joinHasNullKeys_(joinHasNullKeys) {}
+
+  bool joinHasNullKeys() const {
+    return joinHasNullKeys_;
+  }
+
+ private:
+  bool joinHasNullKeys_{false};
+};
+
 /// Generic representation of InsertTable
 struct InsertTableHandle {
  public:
@@ -3108,6 +3123,7 @@ class HashJoinNode : public AbstractJoinNode {
       PlanNodePtr left,
       PlanNodePtr right,
       RowTypePtr outputType,
+      std::shared_ptr<const ReusedHashTableInfo> reusedHashTableInfo = nullptr,
       void* reusedHashTableAddress = nullptr)
       : AbstractJoinNode(
             id,
@@ -3119,7 +3135,8 @@ class HashJoinNode : public AbstractJoinNode {
             std::move(right),
             std::move(outputType)),
         nullAware_{nullAware},
-        reusedHashTableAddress_(reusedHashTableAddress) {
+        reusedHashTableAddress_(reusedHashTableAddress),
+        reusedHashTableInfo_(std::move(reusedHashTableInfo)) {
     validate();
     if (nullAware) {
       VELOX_USER_CHECK(
@@ -3203,6 +3220,11 @@ class HashJoinNode : public AbstractJoinNode {
     return nullAware_;
   }
 
+  const std::shared_ptr<const ReusedHashTableInfo>& reusedHashTableInfo()
+      const {
+    return reusedHashTableInfo_;
+  }
+
   void* reusedHashTableAddress() const {
     return reusedHashTableAddress_;
   }
@@ -3217,6 +3239,8 @@ class HashJoinNode : public AbstractJoinNode {
   const bool nullAware_;
 
   void* reusedHashTableAddress_;
+
+  const std::shared_ptr<const ReusedHashTableInfo> reusedHashTableInfo_;
 };
 
 using HashJoinNodePtr = std::shared_ptr<const HashJoinNode>;
--- a/velox/exec/HashBuild.cpp
+++ b/velox/exec/HashBuild.cpp
@@ -46,6 +46,75 @@ BlockingReason fromStateToBlockingReason(HashBuild::State state) {
 }
 } // namespace
 
+void HashBuild::reuseHashTable(
+    exec::BaseHashTable* baseHashTable,
+    std::shared_ptr<const core::ReusedHashTableInfo> reusedHashTableInfo) {
+  VELOX_CHECK_NOT_NULL(joinBridge_);
+  joinBridge_->start();
+  setState(State::kFinish);
+
+  if (reusedHashTableInfo->joinHasNullKeys() && isAntiJoin(joinType_) &&
+      nullAware_ && !joinNode_->filter()) {
+    joinBridge_->setAntiJoinHasNullKeys();
+  } else {
+    VELOX_CHECK_NOT_NULL(joinBridge_);
+    std::unique_ptr<
+        exec::BaseHashTable,
+        std::function<void(exec::BaseHashTable*)>>
+        hashTable(nullptr, [](exec::BaseHashTable* ptr) { /* Do nothing */ });
+
+    if (auto hasheTableWithNullKeys =
+            dynamic_cast<exec::HashTable<true>*>(baseHashTable)) {
+      hashTable.reset(hasheTableWithNullKeys);
+    } else if (
+        auto hasheTableWithoutNullKeys =
+            dynamic_cast<exec::HashTable<false>*>(baseHashTable)) {
+      hashTable.reset(hasheTableWithoutNullKeys);
+    } else {
+      VELOX_UNREACHABLE("Unexpected HashTable {}", baseHashTable->toString());
+    }
+    joinBridge_->setHashTable(
+        std::move(hashTable), reusedHashTableInfo->joinHasNullKeys(), true);
+  }
+}
+
+void HashBuild::buildHashTable() {
+  const auto& inputType = joinNode_->sources()[1]->outputType();
+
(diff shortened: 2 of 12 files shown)


JkSelf:
@xiaoxmeng @majetideepak Can you help to review this PR? I will add the related unit tests later. Thanks.

JkSelf:
@xiaoxmeng @majetideepak Can you help to review this PR? Thanks.

velox/exec/HashBuild.cpp
```diff
@@ -65,46 +65,83 @@ HashBuild::HashBuild(
...
+      joinBridge_->setAntiJoinHasNullKeys();
+    } else {
+      baseHashTable->prepareJoinTable(
+          {}, BaseHashTable::kNoSpillInputStartPartitionBit);
+
+      VELOX_CHECK_NOT_NULL(joinBridge_);
+      std::unique_ptr<
+          exec::BaseHashTable,
+          std::function<void(exec::BaseHashTable*)>>
+          hashTable(nullptr, [](exec::BaseHashTable* ptr) { /* Do nothing */ });
```
JkSelf: The `HashTable ` will not be deleted because it is shared.

velox/exec/HashTable.cpp
```diff
@@ -1705,61 +1708,65 @@ void HashTable<ignoreNullKeys>::prepareJoinTable(
...
-  }
-
-  bool useValueIds = mayUseValueIds(*this);
-  if (useValueIds) {
-    for (auto& other : otherTables_) {
-      if (!mayUseValueIds(*other)) {
-        useValueIds = false;
-        break;
+  std::lock_guard<std::mutex> l(mutex_);
+  if (!prepared_) {
```
JkSelf: Ensure the HashTable only be prepared once.

JkSelf:
@rui-mo @jinchengchenghh @zhli1142015 Can you help to review this PR? Thanks.

velox/core/PlanNode.h
```diff
@@ -2957,7 +2957,8 @@ class HashJoinNode : public AbstractJoinNode {
       TypedExprPtr filter,
       PlanNodePtr left,
       PlanNodePtr right,
-      RowTypePtr outputType)
+      RowTypePtr outputType,
+      void* reusedHashTableAddress = nullptr)
```
zhouyuan, line 3127: @xiaoxmeng it adds one more parameter(with default value) for `HashJoinNode`, instead of modifying the existing API, is it better to add a new set of one (like the `IndexLookupJoinNode`)?

velox/exec/HashBuild.cpp
```diff
@@ -65,46 +65,83 @@ HashBuild::HashBuild(
...
-    dependentChannels_.reserve(numDependents);
-    decoders_.reserve(numDependents);
-  }
-  for (auto i = 0; i < inputType->size(); ++i) {
-    if (keyChannelMap_.find(i) == keyChannelMap_.end()) {
-      dependentChannels_.emplace_back(i);
-      decoders_.emplace_back(std::make_unique<DecodedVector>());
+  if (reusedHashTableAddress_ != nullptr) {
+    auto baseHashTable =
+        reinterpret_cast<exec::BaseHashTable*>(reusedHashTableAddress_);
```
rui-mo: Do we need to ensure the cast can succeed with `dynamic_cast`?

velox/exec/HashBuild.cpp
```diff
@@ -65,46 +65,83 @@ HashBuild::HashBuild(
...
+        hashTable.reset(hasheTableWithNullKeys);
+      } else if (
+          auto hasheTableWithoutNullKeys =
+              dynamic_cast<exec::HashTable<false>*>(baseHashTable)) {
+        hashTable.reset(hasheTableWithoutNullKeys);
+      } else {
+        VELOX_UNREACHABLE("Unexpected HashTable {}", baseHashTable->toString());
+      }
+      joinBridge_->setHashTable(
+          std::move(hashTable), hashTable->joinHasNullKeys());
```
rui-mo: Would you like to include some comments in this section to clarify how it's done? Also consider extending the documentation for the design of pre-built hash table.

https://github.com/facebookincubator/velox/blob/main/velox/docs/develop/joins.rst

rui-mo (commented):
Thanks!

zhouyuan:
@tanjialiang Hi, could you please also help to take a look on this patch? It's also related with hash join optimization in Gluten. Thanks.

velox/core/PlanNode.h
xiaoxmeng: Can we pass a pre-built hash table object here instead of void pointer? Thanks!

velox/exec/HashBuild.cpp
```diff
@@ -66,46 +66,83 @@ HashBuild::HashBuild(
...
-    // keys with the same build side key: SELECT * FROM t LEFT JOIN u ON t.k1 =
-    // u.k AND t.k2 = u.k.
-    dependentChannels_.reserve(numDependents);
-    decoders_.reserve(numDependents);
-  }
-  for (auto i = 0; i < inputType->size(); ++i) {
-    if (keyChannelMap_.find(i) == keyChannelMap_.end()) {
-      dependentChannels_.emplace_back(i);
-      decoders_.emplace_back(std::make_unique<DecodedVector>());
+  if (reusedHashTableAddress_ != nullptr) {
```
xiaoxmeng, line 147: Can we skip the hash table prepare and the first hash build to finalize the table?

velox/exec/HashBuild.cpp
```diff
@@ -66,46 +66,83 @@ HashBuild::HashBuild(
...
+    auto baseHashTable =
+        reinterpret_cast<exec::BaseHashTable*>(reusedHashTableAddress_);
+
+    VELOX_CHECK_NOT_NULL(joinBridge_);
+    joinBridge_->start();
+
+    if (baseHashTable->joinHasNullKeys() && isAntiJoin(joinType_) &&
```
xiaoxmeng: Can we separate the path for pre-build hash table and the regular path? Thanks!

xiaoxmeng (commented):
@JkSelf we might need a hash table view which is immutable but do we support right join on Gluten? Thanks!

velox/exec/HashTable.h
```diff
@@ -328,6 +328,12 @@ class BaseHashTable {
   /// side. This is used for sizing the internal hash table.
   virtual uint64_t numDistinct() const = 0;
 
+  virtual bool reused() const = 0;
+
+  virtual bool joinHasNullKeys() const = 0;
+
+  virtual void setJoinHasNullKeys() = 0;
```
xiaoxmeng: Instead of saving the join has null keys in hash table, can we have struct pass to plan node which contains the pre-built hash table object + join related table properties such as table has null keys?

tanjialiang:
> @tanjialiang Hi, could you please also help to take a look on this patch? It's also related with hash join optimization in Gluten. Thanks.

Could you please address @xiaoxmeng comments? I have already reviewed this diff and @xiaoxmeng  is helping with the last pass.

JkSelf:
@xiaoxmeng @tanjialiang Thanks for your review. I will resolve all the comments this week. Thanks.

velox/core/PlanNode.h
JkSelf: @xiaoxmeng `HashTable ` object exists in the `velox/exec `package and cannot be accessed from the `velox/core` package because `velox/exec` depends on `velox/core`. Accessing it would create a circular dependency.

velox/exec/HashBuild.cpp
JkSelf: Moved the hash table preparation logic into Gluten after the hash table has been built. And set the hash table state to `kFinish`.

velox/exec/HashBuild.cpp
```diff
@@ -108,6 +115,53 @@ HashBuild::HashBuild(
...
+    reused = (baseHashTable != nullptr);
+
+    if (!reused) {
+      LOG(WARNING)
+          << "Failed to reuse hash table due to type mismatch and need to build hash table";
+    }
+  }
+
+  if (reused) {
+    reuseHashTable(baseHashTable, joinNode_->reusedHashTableInfo());
```
JkSelf, line 159: @xiaoxmeng Add reuseHashTable and buildHashTable to seperate different code path.

velox/exec/HashTable.h
JkSelf: Yes. Add  `ReusedHashTableInfo` struct https://github.com/facebookincubator/velox/pull/13041/files#diff-500373b9b3a6faf282b12f2b04061bcb21a20794b800da36fe3022fbaf6cdaeeR37-R51

JkSelf:
@xiaoxmeng Gluten transforms right joins into left joins, then broadcasts the small table for hash table construction, while maintaining the integrity of the hash table's build state.

JkSelf:
@xiaoxmeng @tanjialiang  Resolve your comments. Please help to review again. Thanks for your help.

velox/exec/HashBuild.cpp
JkSelf: @rui-mo Will add the broadcast hash join implementation doc in separate PR. Thanks.

velox/exec/HashBuild.cpp
```diff
@@ -46,32 +46,39 @@ BlockingReason fromStateToBlockingReason(HashBuild::State state) {
...
-      joinBridge_(operatorCtx_->task()->getHashJoinBridgeLocked(
-          operatorCtx_->driverCtx()->splitGroupId,
-          planNodeId())),
-      keyChannelMap_(joinNode_->rightKeys().size()) {
-  VELOX_CHECK(pool()->trackUsage());
+void HashBuild::reuseHashTable(
+    exec::BaseHashTable* baseHashTable,
+    std::shared_ptr<const core::ReusedHashTableInfo> reusedHashTableInfo) {
   VELOX_CHECK_NOT_NULL(joinBridge_);
+  joinBridge_->start();
```
JkSelf, line 53: @xiaoxmeng The `HashJoinBridge::start() `method must be called before the `setHashTable()` method. The existing hash join implementation also calls `HashJoinBridge::start()` in the `Task::createDriversLocked()` method.

FelixYBW:
@xiaoxmeng We will use this way in Gluten. Does it conflict with https://github.com/facebookincubator/velox/pull/15754? If not can you merge both ways?

cc @pedroerp

JkSelf:
> @xiaoxmeng We will use this way in Gluten. Does it conflict with #15754? If not can you merge both ways?
> 
> cc @pedroerp

@xiaoxmeng @pedroerp @FelixYBW @zhouyuan 

I believe these two implementations(https://github.com/facebookincubator/velox/pull/15754 and https://github.com/facebookincubator/velox/pull/13041) are not contradictory; rather, they represent distinct optimizations tailored for Spark and Presto's respective architectures.

**For Spark**: BroadcastHashJoin constructs the Hash Table during the BroadcastExchange phase and distributes the built table to each executor. This ensures that the Hash Table is built only once globally, and that every executor utilizes the same instance.

**For Presto (Java)**: The Hash Table construction occurs within the HashJoin operator. Typically, the first Task to arrive performs the build, while subsequent tasks reuse the same instance.

Based on our evaluation, [15754](https://github.com/facebookincubator/velox/pull/15754) implementation is unsuitable for Spark scenarios for the following reasons:

1. **Reduced Resource Utilization:** In Spark, building the Hash Table during BroadcastExchange allows the Stage containing the HashJoin to remain in a **Pending** state. This enables the Spark scheduler to allocate resources to other parallel Stages, maximizing cluster utilization. In contrast, under [15754](https://github.com/facebookincubator/velox/pull/15754) approach, the Stage containing the HashJoin is already active and occupying resources. Having executors idle while waiting for the Hash Table build would waste cluster resources and block other parallel stages, ultimately degrading overall performance.
2. **Incompatibility with Spark's Reuse Mechanism:** Spark relies on the ReuseBroadcastExchange rule for optimization. If the build logic is moved into the HashJoin operator, Gluten will be unable to leverage this feature. Currently, [15754](https://github.com/facebookincubator/velox/pull/15754) caches the Hash Table within the QueryCtx pool, which prevents Tasks from different Stages from reusing a Hash Table that has already been created.

**Conclusion:** Due to these architectural discrepancies, Gluten cannot adopt Presto's implementation in https://github.com/facebookincubator/velox/pull/15754. To support efficient Hash Table reuse, we need merge the current code to enable this capability within Gluten.

@xiaoxmeng Could you help to review this PR again? Thanks for your help.

PR #13046, opened 2025-04-16
refactor: Refactor the logic for constructing Join plans from JoinFuzzer into JoinMaker

kevinwilfong:
Summary:
JoinFuzzer has become monolithic (>1500 lines) and functional. We have a lot of functions to 
construct join plans that take the same or very similar long lists of arguments. These functions 
also do a lot of the same things but do not share code. It makes it very difficult to answer
basic questions like: what join plans are we testing in JoinFuzzer?

To try to improve this I've added a JoinMaker class. This class holds the state that's needed to
construct the various plan alternatives, e.g. the inputs, the join type, the filter, and the output 
columns. This way we can call clear functions like makeHashJoin and the arguments specify
just what adjustments we're making to the join plan (are we using flat inputs, are we reversing 
the order of the inputs, are we partitioning the inputs, etc.)

In addition I saw we were passing a lot of the same arguments for the probe and build sides of 
the join. So I added a JoinSource class to capture that (the input data, the key columns, the
type of the data, etc.). Besides making it cleaner I think this will help support fuzzing plans
that join more than 2 tables eventually.

Differential Revision: D73044135

### DIFF
--- a/velox/exec/fuzzer/JoinFuzzer.cpp
+++ b/velox/exec/fuzzer/JoinFuzzer.cpp
@@ -17,15 +17,13 @@
 #include <boost/random/uniform_int_distribution.hpp>
 #include "velox/common/file/FileSystems.h"
 #include "velox/connectors/hive/HiveConnector.h"
-#include "velox/connectors/hive/HiveConnectorSplit.h"
-#include "velox/connectors/hive/PartitionIdGenerator.h"
 #include "velox/dwio/dwrf/RegisterDwrfReader.h"
 #include "velox/dwio/dwrf/RegisterDwrfWriter.h"
-#include "velox/exec/OperatorUtils.h"
+#include "velox/exec/Spill.h"
 #include "velox/exec/fuzzer/FuzzerUtil.h"
+#include "velox/exec/fuzzer/JoinMaker.h"
 #include "velox/exec/fuzzer/ReferenceQueryRunner.h"
 #include "velox/exec/tests/utils/AssertQueryBuilder.h"
-#include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/exec/tests/utils/TempDirectoryPath.h"
 #include "velox/vector/fuzzer/VectorFuzzer.h"
 
@@ -82,40 +80,6 @@ class JoinFuzzer {
 
   void go();
 
-  struct PlanWithSplits {
-    core::PlanNodePtr plan;
-    core::PlanNodeId probeScanId;
-    core::PlanNodeId buildScanId;
-    std::unordered_map<core::PlanNodeId, std::vector<velox::exec::Split>>
-        splits;
-    core::ExecutionStrategy executionStrategy{
-        core::ExecutionStrategy::kUngrouped};
-    int32_t numGroups;
-
-    explicit PlanWithSplits(
-        const core::PlanNodePtr& _plan,
-        const core::PlanNodeId& _probeScanId = "",
-        const core::PlanNodeId& _buildScanId = "",
-        const std::unordered_map<
-            core::PlanNodeId,
-            std::vector<velox::exec::Split>>& _splits = {},
-        core::ExecutionStrategy _executionStrategy =
-            core::ExecutionStrategy::kUngrouped,
-        int32_t _numGroups = 0)
-        : plan(_plan),
-          probeScanId(_probeScanId),
-          buildScanId(_buildScanId),
-          splits(_splits),
-          executionStrategy(_executionStrategy),
-          numGroups(_numGroups) {}
-  };
-
-  static core::PlanNodePtr tryFlipJoinSides(const core::HashJoinNode& joinNode);
-  static core::PlanNodePtr tryFlipJoinSides(
-      const core::MergeJoinNode& joinNode);
-  static core::PlanNodePtr tryFlipJoinSides(
-      const core::NestedLoopJoinNode& joinNode);
-
  private:
   static VectorFuzzer::Options getFuzzerOptions() {
     VectorFuzzer::Options opts;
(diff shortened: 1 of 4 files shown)


velox/exec/fuzzer/JoinMaker.h
```diff
@@ -0,0 +1,271 @@
...
+    std::vector<RowVectorPtr> flatVectors;
+    for (const auto& vector : vectors) {
+      auto flat = BaseVector::create<RowVector>(
+          vector->type(), vector->size(), vector->pool());
+      flat->copy(vector.get(), 0, 0, vector->size());
+      flatVectors.push_back(flat);
+    }
+
+    return flatVectors;
+  }
```
kagamiori, line 95: nit: Move the implementation to .cpp file?

velox/exec/fuzzer/JoinMaker.h
```diff
@@ -0,0 +1,271 @@
...
+        sources_(sources),
+        outputColumns_(outputColumns),
+        filter_(filter) {
+    VELOX_CHECK(
+        sources.size() == 2, "For now we only support joining 2 tables");
+  }
+
+  /// What sort of input to use in the join, either the original encoded
+  /// Vectors, or flattened copies.
+  enum InputType { ENCODED, FLAT };
```
kagamiori, line 144: nit: enum class InputType {...}; Same for below.

velox/exec/fuzzer/JoinMaker.h
```diff
@@ -0,0 +1,271 @@
...
+ private:
+  // Get the sources in the order they should be read according to `joinOrder`.
+  std::vector<std::shared_ptr<JoinSource>> getJoinSources(
+      const JoinOrder joinOrder) const;
+
+  // Get the JoinType to use given `joinOrder`. Throws if the joinType does not
+  // support the given JoinOrder.
+  core::JoinType getJoinType(const JoinOrder joinOrder) const;
+
+  std::string makeNestedLoopJoinFilter(
```
kagamiori, line 240: nit: Can we rename these private helper method to be makeNestedLoopJoinFilter**Impl**(), etc. to differentiate them from the public APIs?

velox/exec/fuzzer/JoinMaker.h
```diff
@@ -0,0 +1,271 @@
...
+  const std::vector<RowVectorPtr> batches_;
+  const std::vector<core::ExprPtr> projections_;
+  const std::string splitsDir_;
+  const std::shared_ptr<memory::MemoryPool> writerPool_;
+
+  // The following are used to cache `batches_` when it is converted to other
+  // forms.
+  mutable std::optional<std::vector<RowVectorPtr>> flatBatches_;
+  mutable std::optional<std::vector<Split>> splits_;
+  mutable std::unordered_map<int32_t, std::vector<Split>> groupedSplits_;
```
kagamiori, line 111: Why would we prefer to mark them `mutable` instead of making functions changing them be non-const? Especially splits_ and groupedSplits_.

velox/exec/fuzzer/JoinFuzzer.cpp
```diff
@@ -1296,49 +739,48 @@ void JoinFuzzer::verify(core::JoinType joinType) {
...
-      probeProjections,
-      buildProjections,
-      outputColumns,
-      altPlans,
-      filter);
+  if (joinMaker.supportsTableScan()) {
+    const int32_t numGroups = randInt(1, probeSource->batches().size());
+
+    altPlans.push_back(joinMaker.makeHashJoinWithTableScan(
+        std::nullopt, // Use ungrouped execution.
```
kagamiori, line 751: Why not just use 0 or -1 to represent ungrouped execution?

kagamiori (commented):
Hi @kevinwilfong, thank you for making this refactoring to improve readability! It looks mostly good to me. I left a few comments.

velox/exec/fuzzer/JoinMaker.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+// Breaks the batches from joinSource (respecting `inputType`) into up to 4
+// partitions. And returns PlanNodes to read each of those partitions, applying
+// any necessary projections.
+std::vector<core::PlanNodePtr> makeSourcesForPartitionedJoinPlan(
+    const std::shared_ptr<JoinSource> joinSource,
+    const JoinMaker::InputType inputType,
+    const std::shared_ptr<core::PlanNodeIdGenerator>& planNodeIdGenerator) {
+  const auto& batches = getBatches(joinSource, inputType);
+  auto numSources = std::min<size_t>(4, batches.size());
```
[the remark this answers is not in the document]
kevinwilfong: good catch, that's a bug in the original code I copied here, I misread the code and wrote the author's intention and not what it does, this should be std::max, i % numSources below doesn't do anything

velox/exec/fuzzer/JoinMaker.h
kevinwilfong: The core state of the class is immutable and I wanted the code to reflect that, these just hold alternate representations of batches_ that are cached so we don't have to recreate them every time their respective getter methods are called. This is an implementation detail of the class that I didn't want to expose in the external API.

velox/exec/fuzzer/JoinMaker.h
```diff
@@ -0,0 +1,271 @@
...
+  std::vector<std::shared_ptr<JoinSource>> getJoinSources(
+      const JoinOrder joinOrder) const;
+
+  // Get the JoinType to use given `joinOrder`. Throws if the joinType does not
+  // support the given JoinOrder.
+  core::JoinType getJoinType(const JoinOrder joinOrder) const;
+
+  std::string makeNestedLoopJoinFilter(
+      const std::shared_ptr<JoinSource>& left,
+      const std::shared_ptr<JoinSource>& righ) const;
```
kevinwilfong, line 242: typo: right

velox/exec/fuzzer/JoinFuzzer.cpp
kagamiori, line 751: nit: Move the comment to a separate line above the code? Ditto for the comment below.

velox/exec/fuzzer/JoinMaker.cpp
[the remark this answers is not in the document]
kagamiori: Wait, I read the code again, and now think that using std::min is actually correct.

kagamiori (approved):
LGTM. There are red internal signals in the diff about join fuzzer test, please take a look too.

velox/exec/fuzzer/JoinMaker.cpp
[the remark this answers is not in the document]
kevinwilfong: You're right, I don't know what I was thinking.

And the comment is right, it breaks up the batches into at most 4 partitions.

PR #13048, opened 2025-04-16
fix: Inverse_poisson_cdf doesn't handle case where results exceed bounds of int32_t

kevinwilfong:
Summary:
inverse_poisson_cdf doesn't handle the case where results exceed the bounds of int32_t, so
it throws an exception (at least with sanitizers enabled) when converting the result to an 
int32_t.

Presto Java handles this by bounding the result to Integer.MAX_VALUE. The change applies
the same restriction to Velox's implementation.

Differential Revision: D73143995


skyelves (commented):
LGTM

skyelves (approved):

PR #10802, opened 2024-08-21
build(cmake): Add VELOX_BUILD_FUZZERS option

majetideepak:
Fixes: https://github.com/facebookincubator/velox/issues/11797

PR #13051, opened 2025-04-17
fix(build): Add option to turn off Spark aggregate fuzzer

Yohahaha:
Spark aggregate fuzzer needs to build with many deps(absl, c-ares, grpc, spark), this patch add `VELOX_BUILD_SPARK_AGGREGATE_FUZZER` allow user who don't care about this feature to turn it off.


rui-mo (commented):
Perhaps we could follow https://github.com/facebookincubator/velox/pull/10802 to add a CMake VELOX_BUILD_FUZZERS flag to build fuzzers optionally, thanks.

assignUser (commented):
I'd agree that it probably makes more sense to generalize as in #10802 but iirc @majetideepak mentioned that some of the fuzzers still need to be decoupled from gtest, like this one. Then we can build fuzzers without building tests and vice versa.

majetideepak:
> if(${VELOX_BUILD_TESTING} AND ${VELOX_BUILD_SPARK_AGGREGATE_FUZZER})

The decoupling is almost near completion. We can then not depend on VELOX_BUILD_TESTING to enable fuzzers and make them independent, similar to benchmarks.

PR #13054, opened 2025-04-17
feat: Optimize Hash Table atomic bools

joeg:
Summary:
HashTable repeatedly sets std::atomic<bool> hasDuplicates_ for each row. This flag can only go from false to true. Every subsequent call to this set is likely redundant. 

Microbenchmarks show a 28000x improvement by using a simple wrapper that avoids the atomic once the flag is already set. TPCH Sf100 on a 88c Bergamo node improved by a factor of >4x while enabling significantly higher cpu utilization. 

A cursory look in Velox shows several other "one way atomic flags" that we could likely replace as well : 

MemoryPool::aborted_
AsyncDataCache::isFirstUse_;
AsyncDataCache::ssdSavable_;
Driver::isEnqueued_
Driver::isTerminated_
ExchangeDataSource::requestPending_
Task::terminateRequested_

Differential Revision: D73003608

### DIFF
--- a/velox/exec/OneWayStatusFlag.h
+++ b/velox/exec/OneWayStatusFlag.h
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/synchronization/SanitizeThread.h>
+#include <atomic>
+
+namespace facebook::velox::exec {
+
+/// A simple one way status flag that uses a non atomic flag to avoid
+/// unnecessary atomic operations.
+class OneWayStatusFlag {
+ public:
+  bool check() const {
+#if defined(__x86_64__)
+    /// This flag is can only go from false to true, and must is only checked at
+    /// the end of a loop. Given that once a flag is true it can never go back
+    /// to false, we are ok to use this in a non synchronized manner to avoid
+    /// the overhead. As such we consciously exempt ourselves here from TSAN
+    /// detection.
+    folly::annotate_ignore_thread_sanitizer_guard g(__FILE__, __LINE__);
+    return fastStatus_ || atomicStatus_.load();
+#else
+    return atomicStatus_.load(std::memory_order_relaxed) ||
+        atomicStatus_.load();
+#endif
+  }
+
+  void set() {
+#if defined(__x86_64__)
+    if (!fastStatus_) {
+      atomicStatus_.store(true);
+      fastStatus_ = true;
+    }
+#else
+    if (!atomicStatus_.load(std::memory_order_relaxed)) {
+      atomicStatus_.store(true);
+    }
+#endif
+  }
+
+  /// Operator overload to convert OneWayStatusFlag to bool
+  operator bool() const {
+    return check();
+  }
+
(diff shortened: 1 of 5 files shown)


Yuhta (approved):

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/synchronization/SanitizeThread.h>
+#include <atomic>
+namespace facebook::velox::exec {
+
+// A simple one way status flag that uses a non atomic flag to avoid
```
xiaoxmeng, line 23: /// for public comments.

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,67 @@
...
+// unnecessary atomic operations.
+class OneWayStatusFlag {
+ public:
+  bool check() const {
+#if defined(__x86_64__)
+    // This flag is can only go from fals to true, and must is only checked at
```
xiaoxmeng, line 29:
```
s/fals/false/
```

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/synchronization/SanitizeThread.h>
+#include <atomic>
+namespace facebook::velox::exec {
```
xiaoxmeng, line 21: Leave an empty line here

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,67 @@
...
+  // Operator overload to convert OneWayStatusFlag to bool
+  operator bool() const {
+    return check();
+  }
+
+ private:
+#if defined(__x86_64__)
+  bool fastStatus_{false};
+#endif
+  std::atomic<bool> atomicStatus_{false};
```
xiaoxmeng, line 64:
```
std::atomic_bool ...
```

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
```
xiaoxmeng, line 18: Can we add unit test for this? Thanks!

xiaoxmeng (commented):
@joeg LGTM % minors. Thanks!

xiaoxmeng (approved):

velox/exec/OneWayStatusFlag.h
```diff
@@ -0,0 +1,68 @@
...
+#if defined(__x86_64__)
+    /// This flag is can only go from false to true, and must is only checked at
+    /// the end of a loop. Given that once a flag is true it can never go back
+    /// to false, we are ok to use this in a non synchronized manner to avoid
+    /// the overhead. As such we consciously exempt ourselves here from TSAN
+    /// detection.
+    folly::annotate_ignore_thread_sanitizer_guard g(__FILE__, __LINE__);
+    return fastStatus_ || atomicStatus_.load();
+#else
+    return atomicStatus_.load(std::memory_order_relaxed) ||
```
MBkkt, line 38: Hi, can you explain this for me?
I don't understand what sense it makes for C++ memory model.

There's two option as I see:
1) there's no need in happens before between check and set, in such case just relaxed store and load should be ok
2) there's needed happens before between check and set, at least acquire for load and release for store is needed

Another thing:
In x86 there's no real difference between seq cst load and non atomic read (except compiler optimization).
And same about x86 store up to release memory order.

So I think your real improvement is non touching atomic after it set first time.
And it makes sense because cache line won't be invalidated in such case.


So instead of this code you can just write
check:
return atomic.load(acquire);

set:
if (!atomic.load(relaxed)) atomic.store(true, release);


And I think it will perform same way

PR #13063, opened 2025-04-18
fix: Correct typos and grammar in velox/expression files

wraymo:
Summary
This PR corrects spelling and grammatical errors specifically within the code comments in the `velox/expression` directory (excluding subdirectories). This enhances the clarity and readability of the codebase.

### DIFF
--- a/velox/expression/VectorWriters.h
+++ b/velox/expression/VectorWriters.h
@@ -62,7 +62,7 @@ struct VectorWriter : public VectorWriterBase {
 
   void copyCommit(const exec_out_t& data) {
     // this code path is called for copying nested structures to slices
-    // in future, we want to eliminate this so all writes go directly to their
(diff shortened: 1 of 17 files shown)


velox/expression/VectorWriters.h
```diff
@@ -62,8 +62,8 @@ struct VectorWriter : public VectorWriterBase {
 
   void copyCommit(const exec_out_t& data) {
     // this code path is called for copying nested structures to slices
```
jkhaliqi, line 64: `This code path is called for copying nested structures to slices. In the future, we want to eliminate this, so all writes go directly to their slice.`

Should we also change line 76 in this file to be
`This code path is called when the slice is top-level.`

wraymo: Thanks. Updated

PR #13065, opened 2025-04-18
fix: Fix the right join result mismatch issue

JkSelf:
```
left    right
null    null
null    null
3       3
```

When performing a right join operation using above data, if we directly set rightRowIndex_ to 0 [here](https://github.com/facebookincubator/velox/pull/13039/files#diff-8279da91a596a4b97a73bc4d9aa635da1baaf254dcdb2f7f8049bd21e7a9551eL833), it will result in all final outcomes being empty in left side. This is because during the comparison process, the first value of the right input is null, leading to a comparison result of -1 consistently. Consequently, all the left-side data will be fully traversed, and the input will be set to null columns, resulting in the final output returning only the right-side columns, with all left-side columns set to null. And this PR addresses the above issue by skipping the null row index prior to comparing the results.

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -861,6 +861,33 @@ RowVectorPtr MergeJoin::getOutput() {
   }
 }
 
+std::pair<bool, RowVectorPtr> MergeJoin::handleRightSideNullRows() {
+  auto rightFirstNonNullIndex = firstNonNull(rightInput_, rightKeyChannels_);
+  if ((isRightJoin(joinType_) || isFullJoin(joinType_)) &&
+      rightFirstNonNullIndex > rightRowIndex_) {
+    if (prepareOutput(nullptr, rightInput_)) {
+      output_->resize(outputSize_);
+      return std::make_pair(true, std::move(output_));
+    }
+    for (int i = rightRowIndex_; i < rightFirstNonNullIndex; i++) {
+      if (!tryAddOutputRowForRightJoin()) {
+        rightRowIndex_ = i;
+        return std::make_pair(true, std::move(output_));
+      }
+
+      if (finishedRightBatch()) {
+        // Ran out of rows on the right side.
+        rightInput_ = nullptr;
+        return std::make_pair(true, nullptr);
+      }
+    }
+
+    rightRowIndex_ = rightFirstNonNullIndex;
+  }
+
+  return std::make_pair(false, nullptr);
+}
+
 RowVectorPtr MergeJoin::doGetOutput() {
   // Check if we ran out of space in the output vector in the middle of the
   // match.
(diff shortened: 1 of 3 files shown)


velox/exec/MergeJoin.cpp
```diff
@@ -1060,6 +1060,29 @@ RowVectorPtr MergeJoin::doGetOutput() {
     return nullptr;
   }
 
+  auto firstNonNullIndex = firstNonNull(rightInput_, rightKeyChannels_);
```
xiaoxmeng, line 1063: Shall we just revert the relevant change in https://github.com/facebookincubator/velox/pull/13039

JkSelf: @xiaoxmeng No, The issue in this PR can be resolved by reverting #13039. However, a bug still persists in PR #13039. We need to verify whether the `prepareOutput ` function returns true. If it does, we should return the output. Otherwise, we must set the rows to null for the range `[0, firstNotNullIndex)`.

velox/exec/MergeJoin.cpp
```diff
@@ -861,6 +861,33 @@ RowVectorPtr MergeJoin::getOutput() {
   }
 }
 
+std::pair<bool, RowVectorPtr> MergeJoin::handleRightSideNullRows() {
+  auto rightFirstNonNullIndex = firstNonNull(rightInput_, rightKeyChannels_);
```
xiaoxmeng, line 865:
```
const auto rightFirstNonNullIndex =
```

velox/exec/MergeJoin.cpp
```diff
@@ -861,6 +861,33 @@ RowVectorPtr MergeJoin::getOutput() {
...
+  if ((isRightJoin(joinType_) || isFullJoin(joinType_)) &&
+      rightFirstNonNullIndex > rightRowIndex_) {
+    if (prepareOutput(nullptr, rightInput_)) {
+      output_->resize(outputSize_);
+      return std::make_pair(true, std::move(output_));
+    }
+    for (int i = rightRowIndex_; i < rightFirstNonNullIndex; i++) {
```
xiaoxmeng, line 872: ++i

velox/exec/MergeJoin.cpp
```diff
@@ -1060,6 +1087,11 @@ RowVectorPtr MergeJoin::doGetOutput() {
     return nullptr;
   }
 
+  auto pair = handleRightSideNullRows();
```
xiaoxmeng, line 1090:
```
const auto output = handleRightSideNullRows();
if (output != nullptr || rightInput_ == nullptr) {
  return output;
}
VELOX_CHECK_NOTNULL(rightInput_);
```

JkSelf:
@xiaoxmeng Resolved all your comments. Please help to review again. Thanks.

xiaoxmeng (approved):
@JkSelf thanks for the update!

kevinwilfong (approved):
Thanks for the fix!

PR #13069, opened 2025-04-18
feat(variant): Allow hashing opaque variants

pedroerp:
Summary:
Since variants are used in some parts of expression compilation,
enabling a simple hashing strategy for opaque types. To make it consistent to
other places where we hash opaque types (e.g. FlatVector::hashAll()), allowing
hashes of opaque types using the physical pointer value.

This is sufficient to generate different hashes for different objects, but it
does not give users the flexibility of providing a function that could enable
different objects that semantically represent the same concept from hashing the
same. This could be added by enabling users to provide a custom hash function
for opaque types, if the use case arrives in the future.

Differential Revision: D73265149

### DIFF
--- a/velox/type/Variant.cpp
+++ b/velox/type/Variant.cpp
@@ -866,11 +866,6 @@ uint64_t variant::hash<TypeKind::MAP>() const {
       folly::Hash{}, combinedKeyHash, combinedValueHash);
 }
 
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
Thank you, @pedroerp

PR #13072, opened 2025-04-18
feat: Conditionally include Geospatial capabilities

jagill:
Summary: This strips out Geospatial capabilities unless VELOX_ENABLE_GEO is defined.  It reaches into a set of files; which is some extra complexity.  But it's a stepping stone before we have the velox-full and velox-core builds.

Differential Revision: D73278117


velox/exec/tests/FunctionResolutionTest.cpp
```diff
@@ -21,12 +21,15 @@
 #include "velox/functions/Udf.h"
 #include "velox/functions/prestosql/registration/RegistrationFunctions.h"
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
-#include "velox/functions/prestosql/types/BingTileType.h"
 #include "velox/functions/prestosql/types/HyperLogLogType.h"
 #include "velox/functions/prestosql/types/JsonType.h"
 #include "velox/functions/prestosql/types/TDigestType.h"
 #include "velox/functions/prestosql/types/TimestampWithTimeZoneType.h"
 
+#ifdef VELOX_ENABLE_GEO
```
czentgr, line 29: The file BingTileType.h is empty if GEO is not set (it has the same macro check) and only has `pragma once` set. So maybe we don't need this change and let it include the empty file? Might be easier as everything is controlled within the h/cpp files if anything is implemented?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -0,0 +1,139 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <geos/io/WKBReader.h>
```
czentgr, line 19: Do we want to also wrap this code in `#ifdef VELOX_ENABLE_GEO`?

velox/functions/prestosql/registration/GeometryFunctionsRegistration.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <string>
+#include "velox/functions/Registerer.h"
+#include "velox/functions/prestosql/GeometryFunctions.h"
+#include "velox/functions/prestosql/types/GeometryRegistration.h"
+
+namespace facebook::velox::functions {
+void registerGeometryFunctions(const std::string& prefix) {
+  registerGeometryType();
+  registerFunction<StGeometryFromTextFunction, Geometry, Varchar>(
+      {{prefix + "ST_GeometryFromText"}});
```
czentgr, line 28: Should this not be `ST_GeomFromText` to follow the standard?

velox/functions/prestosql/TypeOf.cpp
```diff
@@ -86,9 +88,11 @@ std::string typeName(const TypePtr& type) {
       if (isHyperLogLogType(type)) {
         return "HyperLogLog";
       }
+#ifdef VELOX_ENABLE_GEO
```
czentgr, line 91: We could avoid this in the header by exposing this inline function in an else to return always false instead of the implementation when GEO is enabled.
Just thinking on how to reduce these types of additions and contain them within the files that contain the implementation and otherwise if not supported return defaults (false).
Same for the earlier BingTileType.

PR #13076, opened 2025-04-19
fix(build): Build Wave dwio tests only when enabled

zhouyuan:
If gpu is enabled, dwio/tests will be always build. This patch fixes the build behavior


assignUser (approved):
Thanks!

kevinwilfong:
@zhouyuan could you rebase? There are those broken CIs, I suspect they're not related to your change, but it'd be nice to have a clean run just to confirm

zhouyuan:
@kevinwilfong Just rebased. Thanks for notifying! -yuan

PR #13077, opened 2025-04-20
feat: Add SpillPartitionIdLookup for fast partitioning

tanjialiang:
Summary: 'SpillPartitionIdLookup' provides efficient mapping from hash values to 'SpillPartitionId'. It's used in spill operations to quickly determine which partition a row should be assigned to based on its hash value.

Differential Revision: D73289833


velox/exec/Spill.h
```diff
@@ -281,6 +285,48 @@ inline std::ostream& operator<<(std::ostream& os, SpillPartitionId id) {
...
+  //
+  // The key (index to lookup_ vector) has the following bit layout:
+  // |...unused (set 0)...|...lookup bits (4 * numPartitionBits bits)...|
+  //
+  // A hash value is used with the following bit layout:
+  // |...unused...|...lookup bits...|...unused...|
+  //
+  // When a hash value is used to partition, its lookup bits range will be used
+  // as index to check against 'lookup_' and find the corresponding id.
+  std::vector<SpillPartitionId> lookup_;
```
xiaoxmeng, line 327: This should be a const?

velox/exec/Spill.h
```diff
@@ -281,6 +285,48 @@ inline std::ostream& operator<<(std::ostream& os, SpillPartitionId id) {
...
+  void generateLookupHelper(
+      const SpillPartitionId& id,
+      uint32_t startBit,
+      uint32_t endBit,
+      uint64_t prevLookup,
+      uint64_t fixedBits);
+
+  const uint64_t partitionBitsMask_;
+
+  // Index is the extracted hash value key, value is the corresponding spill
```
xiaoxmeng, line 315:
```
// The 'lookup_' array index is ...
```

velox/exec/Spill.h
```diff
@@ -281,6 +285,48 @@ inline std::ostream& operator<<(std::ostream& os, SpillPartitionId id) {
...
+  // partition id. The 'lookup_' vector is sized to accommodate all possible
```
xiaoxmeng, line 316:
```
s/The 'lookup_' vector/The vector/
```

velox/exec/Spill.h
```diff
@@ -281,6 +285,48 @@ inline std::ostream& operator<<(std::ostream& os, SpillPartitionId id) {
...
+  // combinations of partition bits (2^numLookupBits entries).
+  //
+  // The key (index to lookup_ vector) has the following bit layout:
```
xiaoxmeng, line 319:
```
The index key has the following bit layout
```

velox/exec/Spill.h
```diff
@@ -281,6 +285,48 @@ inline std::ostream& operator<<(std::ostream& os, SpillPartitionId id) {
...
+  // |...unused (set 0)...|...lookup bits (4 * numPartitionBits bits)...|
+  //
+  // A hash value is used with the following bit layout:
```
xiaoxmeng, line 322: This description on the bit layout is a bit confusing. I think a high level is pretty clear within the context.

velox/exec/Spill.cpp
```diff
@@ -339,13 +339,15 @@ void FileSpillMergeStream::close() {
   spillFile_.reset();
 }
 
+const SpillPartitionId SpillPartitionId::kEmptyId{};
```
xiaoxmeng, line 342: use inline then you don't need the storage in cc here.

velox/exec/Spill.cpp
```diff
@@ -430,6 +435,84 @@ uint32_t SpillPartitionId::encodedId() const {
...
+      sizeof(uint64_t) * 8 - numLookupBits,
+      "Insufficient lookup bits.");
+  lookup_.resize(1UL << numLookupBits, SpillPartitionId::kEmptyId);
+
+  for (const auto& id : spillPartitionIds) {
+    generateLookup(id, startPartitionBit, numPartitionBits);
+  }
+}
+
+void SpillPartitionIdLookup::generateLookup(
```
xiaoxmeng, line 461: This can be static method? Not sure if we want to compute the max spill level in spillPartitionIds to save space? Otherwise we could std::array as its size is prior known.

velox/exec/Spill.cpp
```diff
@@ -430,6 +435,84 @@ uint32_t SpillPartitionId::encodedId() const {
...
+  const auto enumStartBit =
+      startPartitionBit + (id.spillLevel() + 1) * numPartitionBits;
+  const auto enumEndBit = startPartitionBit +
+      (SpillPartitionId::kMaxSpillLevel + 1) * numPartitionBits;
+
+  // Calculate the spill level partition bits range bits which are fixed.
+  uint64_t fixedBits{0};
+  for (auto i = 0; i <= id.spillLevel(); ++i) {
+    const auto partitionNum = id.partitionNumber(i);
+    VELOX_CHECK_LT(
```
xiaoxmeng, line 480: VELOX_DCHECK_LT and id ctor should has already ensure that.

velox/exec/Spill.cpp
```diff
@@ -430,6 +435,84 @@ uint32_t SpillPartitionId::encodedId() const {
...
+        partitionNum,
+        1UL << numPartitionBits,
+        "partition number exceeds max partition number");
```
xiaoxmeng, line 483:
```
s/partition/Partition/
```

xiaoxmeng (commented):
@tanjialiang thanks for the change % comments.

velox/exec/Spill.h
```diff
@@ -249,6 +251,8 @@ class SpillPartitionId {
   uint32_t encodedId() const;
 
  private:
+  SpillPartitionId() : encodedId_(0xFFFFFFFF){};
```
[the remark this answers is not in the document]
tanjialiang: We can by default set this value, instead of putting it in ctor

PR #13081, opened 2025-04-21
feat: add imds enabled to s3 config

Super-long:
When I tested using s3fs to read data from s3, I found that the first read performance was very poor. The total time minus the actual IO latency was 5-6 seconds. I finally located it to be the ClientConfiguration structure. This modification configures the opening of IMDS. When IMDS is closed, the first reading of the ClientConfiguration structure can be reduced from 5-6 seconds to the us level.

Aws SDK has a similar discussion before https://github.com/aws/aws-sdk-cpp/discussions/2398
The IMDS option does not make sense to enable it in instances outside EC2, but in order to maintain semantics, it is currently enabled by default


velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -627,7 +627,9 @@ class S3FileSystem::Impl {
  public:
   Impl(const S3Config& s3Config) {
     VELOX_CHECK(getAwsInstance()->isInitialized(), "S3 is not initialized");
-    Aws::S3::S3ClientConfiguration clientConfig;
+    Aws::Client::ClientConfigurationInitValues initValues;
```
majetideepak, line 630: The S3Client constructor with Aws::Client::ClientConfiguration argument is deprecated. Use `Aws::S3::S3ClientConfiguration`. I see there is an option to disable IMDS for that as well.

zhouyuan:
Cc @JkSelf

velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -114,6 +115,8 @@ class S3Config {
              std::make_pair("use-proxy-from-env", "false")},
             {Keys::kCredentialsProvider,
              std::make_pair("aws-credentials-provider", std::nullopt)},
+            {Keys::kIMDSEnabled,
+             std::make_pair("aws-imds-enabled", "true")},
```
majetideepak, line 119: Can you please add a doc for this config inside `velox/docs/configs.rst`?

majetideepak (commented):
Thanks for this fix!

velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -243,6 +246,12 @@ class S3Config {
     return config_.find(Keys::kCredentialsProvider)->second;
   }
 
+  /// Use AWS EC2 Instance Metadata Service
```
JkSelf: Maybe we can make the comment a bit clearer by rephrasing it as: `Returns true if IMDS is enabled in the configuration settings`.

JkSelf:
@Super-long Could you help to resolve the conflict? And i will add the ready to merge label after resolving the conflicts. Thanks.

PR #13083, opened 2025-04-21
fix: Change velox buffer copy and serialization interfaces to take size in i64 instead of i32

yuandagits:
Summary:
Many internal queries crash at places like crc computation, or serialization because we have an restriction on the size of the buffer since the input to many of these functions are int32_t. Let's change this to be int64_t and let the application specific logic take care of the large buffer if need be.

The components/functions changed are:
- A few simd utils (memcopies)
- newView API in byteStream
- process_bytes in the crc computation
- byteRange struct
- Functions in ByteStream to internally use int64_t
- StreamArena's newRange and allocate

For example, Presto queries that would crash at crc computation or serialization time would now throw memory exceeded or spill error similar to Java.

Differential Revision: D73310051

### DIFF
--- a/velox/common/base/Crc.h
+++ b/velox/common/base/Crc.h
@@ -22,7 +22,7 @@ namespace facebook::velox::bits {
 // A boost compatible CRC32 calculator.
 class Crc32 {
  public:
-  void process_bytes(const void* data, int32_t size) {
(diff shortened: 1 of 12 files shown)


kewang1024 (approved):
Thank you @yuandagits

PR #13092, opened 2025-04-21
fix(parser): Change Velox parser errors to user errors

kKPulla:
Summary:
This will enable engines to handle unsupported grammar related errors (and rethrow w/ context to users if need be) as these are non-retriable anyways.

Added a unit test for specific case like CHAR(10) which is Presto syntax but not supported in Velox.

Differential Revision: D73383288

### DIFF
--- a/velox/type/parser/TypeParser.yy
+++ b/velox/type/parser/TypeParser.yy
@@ -38,7 +38,7 @@
 %token <long long>   NUMBER
 %token YYEOF         0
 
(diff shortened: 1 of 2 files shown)


amitkdutta (approved):
Thanks @kKPulla

majetideepak (approved):

pedroerp:
Looks good to me as well. Thanks @kKPulla

PR #13095, opened 2025-04-21
fix: Eliminate comparison operator ambiguity in ComplexViewTypes

ispeters:
Summary:
As identified by @r-barnes, the comparison operators defined in
`velox/expression/ComplexViewTypes.h` are ambiguous in C++20. This
diff fixes them by making both parameters constant references to the derived
type instead of a mix of the base and derived types.

Differential Revision: D73390221


kevinwilfong (approved):
LGTM

PR #13096, opened 2025-04-21
fix: Fix the semi merge join with duplicate match vectors

xiaoxmeng:
Summary:
When semi-join with multiple matched vectors, it produce redundant matched rows with one per each matched vector but we actually just
need to produce exact one matched row. This PR fixes this and verified with unit test

Differential Revision: D73397297

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -608,27 +608,27 @@ bool MergeJoin::addToOutputForLeftJoin() {
           : rightMatch_->startRowIndex;
 
       const auto numRightBatches = rightMatch_->inputs.size();
-      for (size_t r = firstRightBatch; r < numRightBatches; ++r) {
+      // TODO: Since semi joins only require determining if there is at least
+      // one match on the other side, we could explore specialized algorithms
+      // or data structures that short-circuit the join process once a match
+      // is found.
+      for (size_t r = isLeftSemiFilterJoin(joinType_) ? numRightBatches - 1
+                                                      : firstRightBatch;
+           r < numRightBatches;
+           ++r) {
         const auto rightBatch = rightMatch_->inputs[r];
-        const auto rightStartRow =
-            r == firstRightBatch ? rightStartRowIndex : 0;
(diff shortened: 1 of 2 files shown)


velox/exec/MergeJoin.cpp
```diff
@@ -608,29 +608,28 @@ bool MergeJoin::addToOutputForLeftJoin() {
...
+      // is found.
+      for (size_t r = isLeftSemiFilterJoin(joinType_) ? numRightBatches - 1
+                                                      : firstRightBatch;
+           r < numRightBatches;
+           ++r) {
         const auto rightBatch = rightMatch_->inputs[r];
-        const auto rightStartRow =
-            r == firstRightBatch ? rightStartRowIndex : 0;
+        auto rightStartRow = r == firstRightBatch ? rightStartRowIndex : 0;
         auto rightEndRow = r == numRightBatches - 1 ? rightMatch_->endRowIndex
```
kevinwilfong, line 621: nit: this can be const now

velox/exec/MergeJoin.cpp
```diff
@@ -688,11 +687,22 @@ bool MergeJoin::addToOutputForRightJoin() {
...
+      // or data structures that short-circuit the join process once a match
+      // is found.
+      for (size_t l = isRightSemiFilterJoin(joinType_) ? numLeftBatches - 1
+                                                       : firstLeftBatch;
+           l < numLeftBatches;
+           ++l) {
         const auto leftBatch = leftMatch_->inputs[l];
-        const auto leftStartRow = l == firstLeftBatch ? leftStartRowIndex : 0;
+        auto leftStartRow = l == firstLeftBatch ? leftStartRowIndex : 0;
         auto leftEndRow = l == numLeftBatches - 1 ? leftMatch_->endRowIndex
```
kevinwilfong, line 700: nit: same here

kevinwilfong (approved):
Thanks for the fix!

JkSelf:
@xiaoxmeng @kevinwilfong This issue has fixed in https://github.com/facebookincubator/velox/pull/11771. And https://github.com/facebookincubator/velox/pull/11771 also fixed the anti and full outer join issues. Can you help to review if you have time? Thanks.

Issue #10571, opened 2024-07-25
map_subset: Presto throws while Velox doesn't


kagamiori:
### Description

Query: 
```
SELECT
    map_subset(c0, c1)
FROM (
    VALUES
        (
            MAP(ARRAY[ARRAY[1, 2]], ARRAY[1]),
            ARRAY[ARRAY[1, NULL], ARRAY[1, NULL]]
        )
) t(c0, c1);
```
Presto-0.289: Throws an exception "ARRAY comparison not supported for arrays with null elements".
Velox: Returns a result {}

juwentus1234:
This seems not trivial as the way that presto java did for compare is very different in the sense that it  uses block compare and position also rewrite the hash open addressing

Presto Java accepted 
SELECT
    map_subset(c0, c1)
FROM (
    VALUES
        (
            MAP(ARRAY[ARRAY[1, 2]], ARRAY[1]),
            ARRAY[ARRAY[1, NULL], ARRAY[3, NULL]]
        )
) t(c0, c1)
limit 100

but not 
SELECT
    map_subset(c0, c1)
FROM (
    VALUES
        (
            MAP(ARRAY[ARRAY[1, 2]], ARRAY[1]),
            ARRAY[ARRAY[1, NULL], ARRAY[2, NULL]]
        )
) t(c0, c1)
limit 100

rewrited Hash and Get slot position of element at {@code position} of {@code block} in TypeSet.java and compare in the ArrayType.java in this case for Array type  
    private int getHashPositionOfElement(Block block, int position)
    {
        int hashPosition = getMaskedHash(hashPosition(elementType, block, position));
        while (true) {
            int blockPosition = blockPositionByHash.get(hashPosition);
            // Doesn't have this element
            if (blockPosition == EMPTY_SLOT) {
                return hashPosition;
            }
            // Already has this element
            else if (isContainedAt(block, position, blockPosition)) {
                return hashPosition;
            }

            hashPosition = getMaskedHash(hashPosition + 1);
        }
    }

kgpai:
@juwentus1234 
I think you should also file an issue with Presto about inconsistency with Presto java. 

```
Presto Java accepted
SELECT
map_subset(c0, c1)
FROM (
VALUES
(
MAP(ARRAY[ARRAY[1, 2]], ARRAY[1]),
ARRAY[ARRAY[1, NULL], ARRAY[3, NULL]]
)
) t(c0, c1)
limit 100
```

and 

```
SELECT
[...]

PR #13099, opened 2025-04-21
fix: MapSubset ARRAY comparison should not be supported for null elements

juwentus1234:
Summary:
Presto java and Velox should be consistent on handling null elements compare for query such as:

```
SELECT
    map_subset(c0, c1)
FROM (
    VALUES
        (
            MAP(ARRAY[ARRAY[1, 2]], ARRAY[1]),
            ARRAY[ARRAY[1, NULL], ARRAY[1, NULL]]
        )
) t(c0, c1)
LIMIT
    1000
```
https://github.com/facebookincubator/velox/issues/10571

the diff will fix so that velox also throw out exception.

link to newly created issue https://github.com/prestodb/presto/issues/25011

Differential Revision: D73242028

### DIFF
--- a/velox/functions/prestosql/MapSubset.h
+++ b/velox/functions/prestosql/MapSubset.h
@@ -161,6 +161,21 @@ struct MapSubsetVarcharFunction {
   std::vector<std::string> searchKeyStrings_;
 };
 
+struct MapSubsetFunctionEqualComparator {
+  bool operator()(const exec::GenericView& lhs, const exec::GenericView& rhs)
+      const {
+    static constexpr auto kEqualValueAtFlags = CompareFlags::equality(
+        CompareFlags::NullHandlingMode::kNullAsIndeterminate);
+
+    auto result = lhs.compare(rhs, kEqualValueAtFlags);
+
+    VELOX_USER_CHECK(
+        result.has_value(), "Comparison on null elements is not supported");
+
+    return result.value() == 0;
+  }
+};
+
 /// Generic implementation. Doesn't provide an optimization for constant search
 /// keys.
 template <typename TExec>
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/MapSubset.h
```diff
@@ -161,6 +161,22 @@ struct MapSubsetVarcharFunction {
...
+  bool operator()(const exec::GenericView& lhs, const exec::GenericView& rhs)
+      const {
+    static constexpr auto kEqualValueAtFlags = CompareFlags::equality(
+        CompareFlags::NullHandlingMode::kNullAsIndeterminate);
+
+    auto result = lhs.compare(rhs, kEqualValueAtFlags);
+
+    VELOX_USER_CHECK(
+        result.has_value(),
+        "ARRAY comparison not supported for arrays with null elements");
```
kagamiori, line 174: Just "Comparison on null elements is not supported". The map key is not necessarily of ARRAY type.

kagamiori (commented):
Hi @juwentus1234, thank you for making the fix! Could you also run expression fuzzer with Presto as the source of truth with --only="map_subset" for 1 hour? If there is not other mismatch, we'd like to unblock "map_subset" from the fuzzer test too.

velox/functions/prestosql/MapSubset.h
juwentus1234: updated thanks:)

juwentus1234:
> Hi @juwentus1234, thank you for making the fix! Could you also run expression fuzzer with Presto as the source of truth with --only="map_subset" for 1 hour? If there is not other mismatch, we'd like to unblock "map_subset" from the fuzzer test too.

It has this error https://www.internalfb.com/phabricator/paste/view/P1793346446 with/without my change though, I debugged and it looks other errors which does not going into this change.

kagamiori (approved):
LGTM. Thank you for the fix!

PR #11689, opened 2024-11-28
fix: Unsigned index underflow in trimWhiteSpace

Yuhta:
Summary: We should prefer signed type for numeric values whenever possible.  Unsigned types are only for opaque bits.

Differential Revision: D66587108

PR #13103, opened 2025-04-22
misc: Use unsigned integers consistently for bitwise operations

andersdellien-arm:
This change decreases the risk of unintended signed/unsigned conversions and also allows the compiler to generate more efficient code. By avoiding signed integers there is no need for sign extension, which saves a few instructions when looping over bit arrays. This change gives around 2% performance improvement for the "Copy" sub-benchmarks in VectorSlice, as measured on an AWS C7G machine.


velox/common/base/BitUtil.cpp
```diff
@@ -26,8 +26,8 @@ namespace facebook::velox::bits {
 namespace {
 // Naive implementation that does not rely on BMI2.
 void scatterBitsSimple(
-    int32_t numSource,
-    int32_t numTarget,
+    uint32_t numSource,
```
Yuhta, line 29: This particular change could cause problem when `numTarget` is 0.  And it won't improve performance (only when we do division on unsigned number we will see perf gain).  We should leave them as signed.

velox/common/base/BitUtil.cpp
```diff
@@ -108,8 +108,8 @@ std::string toString(const void* bits, int offset, int size) {
 }
 
 void scatterBits(
-    int32_t numSource,
-    int32_t numTarget,
+    uint32_t numSource,
```
Yuhta, line 111: ditto

velox/common/base/BitUtil.cpp
```diff
@@ -118,13 +118,16 @@ void scatterBits(
     return;
   }
 #ifdef __BMI2__
-  int32_t highByte = numTarget / 8;
-  int32_t highBit = numTarget & 7;
-  int lowByte = std::max(0, highByte - 7);
+  uint32_t highByte = numTarget / 8;
```
Yuhta, line 121: You can `static_cast<uint32_t>(numTarget)` if we keep `numTarget` signed to still get perf gain here.

velox/common/base/BitUtil.cpp
```diff
@@ -118,13 +118,16 @@ void scatterBits(
...
+  uint32_t highBit = numTarget & 7;
+  uint32_t lowByte = 0;
+  if (highByte >= 7) {
```
Yuhta, line 124: Can we check we generate optimal instruction for this?  Should not introduce any branching.  Maybe just use `?:` to be sure

velox/common/base/BitUtil.cpp
```diff
@@ -143,27 +146,30 @@ void scatterBits(
...
       *targetPtr = (*targetPtr & ~writeMask) | (newBits & writeMask);
     }
-    VELOX_DCHECK_GE(numSource, 0);
     if (!lowByte) {
       break;
     }
     highByte = lowByte;
     highBit = 0;
-    lowByte = std::max(lowByte - 8, 0);
+    if (lowByte >= 8) {
```
Yuhta, line 168: Use `?:` to be safe and more intentional for branchless

velox/common/base/BitUtil.h
```diff
@@ -257,19 +257,19 @@ forEachWord(int32_t begin, int32_t end, PartialWordFunc partialWordFunc) {
 // outside of begin ... end by up to kWidth - 1 bits but using the
 // mask for example for load with mask will scope the operation to
 // valid values only.
-template <int8_t kWidth, typename Callable>
+template <uint8_t kWidth, typename Callable>
```
Yuhta, line 260: This has no effect, compiler will just plug in the exact literal and there is no instruction generated.  Using signed number is safer in general.

Yuhta (commented):
I would say we just need to make sure the operand is unsigned when we do division and everywhere else we can leave it as signed, it's safer that way (see #11689).

andersdellien-arm:
Hi @Yuhta 
Thanks for looking at this. Just to clarify the cause of the performance gains here - there is nothing inherently faster with unsigned vs signed variables, however when converting from signed to unsigned the compiler needs to add extra instructions to sign-extend the value (SXTW on Arm). 
As some of the lower-level utility functions have uint64_t parameters (introduced in #11986) there are some extra instructions added in hot loops when they are called with int32_t. Changing to uint32_t removes the need for the SXTW instructions and gives a performance gain.
Please let me know if this makes sense on an overall level and I will go ahead and work on your other comments.
Thanks

Yuhta:
@andersdellien-arm Do we know where the exactly sign extension happen?  We should only cast at the very place where it happens.  As I mentioned earlier using unsigned number at higher level could introduce tricky underflow bugs like #11689 so we want to avoid them as much as possible.

andersdellien-arm:
@Yuhta I share your concern about mixing signed/unsigned, however this seems to me to be somewhat built in to the codebase as the vector sizes are signed but the underlying bit-manipulation functions (setBit/clearBit) take unsigned parameters. 

For example, in the call sequence copyRangesImpl -> applyToEachRow -> setNull -> clearBit we have an implicit conversion in the last step in the chain.

So in general any piece of code that manipulates vectors will risk having this issue.

It is certainly possible to type cast the parameters to avoid this problem, however this means that there will be a lot of type casts spread out in the code. In this updated patch, I have just changed the set/clearNull parameters to avoid (some of the) type conversions with a minimal patch.

Let me know if this make sense, otherwise I can try the type-casting approach instead.

andersdellien-arm:
@Yuhta Did you have time to take another look at this?

PR #13104, opened 2025-04-22
fix: Fix the merge join source testvalue naming

xiaoxmeng:
Summary: The current naming uses merge source but not merge join source,

Differential Revision: D73444570


tanjialiang (approved):

JkSelf (approved):
Good catch.

PR #13105, opened 2025-04-22
feat: Add  variant::tryOpaque method to cast without throwing

breeze1990:
Summary: need a lightweight non-throwing cast method

Differential Revision: D73414779


mbasmanova (approved):
Thanks.

PR #13108, opened 2025-04-22
feat(json): Add support for field_names_in_json_cast_enabled when casting rows to json

kgpai:
Summary:
This diff adds support for field_names_in_json_cast_enabled session property which when enabled produces a json object with the field names having same key as the child name in the row. 

For e.g: 

```
presto:gen_ai> set session field_names_in_json_cast_enabled=true;
SET SESSION           
presto:gen_ai> select cast(cast(row(1, array[1,2], 3.0, row('a', 5)) AS row(x int, y array<int>, z double, r row(a varchar, b int))) as json);
                     _col0                     
-----------------------------------------------
 {"r":{"a":"a","b":5},"x":1,"y":[1,2],"z":3.0} 
```

Velox will produce canonicalized json even though currently Presto java doesnt.

Differential Revision: D73468351


velox/core/QueryConfig.h
```diff
@@ -542,6 +542,12 @@ class QueryConfig {
   static constexpr const char* kStreamingAggregationEagerFlush =
       "streaming_aggregation_eager_flush";
 
+  /// If this is true, then we it allows you to get the struct field names
```
jkhaliqi, line 545: nit: small typo can get rid of the 'we'

amitkdutta (approved):
Looks great. Thanks so much @kgpai for unblocking an important feature.

PR #13110, opened 2025-04-23
fix: Fix raw_vector allocation bytes overflow

tanjialiang:
Summary:
Change int32_t type to int64_t in raw_vector class to:
1) fix the overflow it caused when allocation more than 2GB of memory
2) align with allocator API (int64_t)

Differential Revision: D73484257


xiaoxmeng (approved):

Issue #9499, opened 2024-04-16
Add `encoding` parquet writer option test


czentgr:
### Description

The ParquetWriter tests were added testing the compression option but other options also need to be tested and added.

The test file is https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp.
The writer options are here: https://github.com/facebookincubator/velox/blob/0680169a239f9dd84d4eb8c17b01a8df6249f8f9/velox/dwio/parquet/writer/Writer.h#L89

four1er:
I want to do it. Pls assign it to me, thx.

PR #13111, opened 2025-04-23
feat: Allow ParquetWriter to specify column encoding

Super-long:
In this issue https://github.com/facebookincubator/velox/issues/9499 , I noticed that Parquet Write did not have any encoding-related tests. When I added relevant tests, I found that I could not specify the encoding of columns. I improved the logic here and added relevant tests.


Super-long:
Cc @majetideepak

PR #9559, opened 2024-04-22
misc: Enable Spark query runner as reference in aggregation fuzzer test

rui-mo:
- Enables Spark query runner as reference DB in Spark aggregation fuzzer test. 
- Removes duplicate Spark aggregation fuzzer test from experimental run.
- Implements `executeAndReturnVector` in Spark query runner and removes 
the below two APIs from the reference query runner as they are not in use since
https://github.com/facebookincubator/velox/commit/8519fad1170975ab32ee75e186a88f91d3098495.

```
virtual std::multiset<std::vector<velox::variant>> execute(
      const std::string& /*sql*/,
      const std::vector<velox::RowVectorPtr>& /*input*/,
      const velox::RowTypePtr& /*resultType*/) {
    VELOX_UNSUPPORTED();
  }

virtual std::vector<RowVectorPtr> executeVector(
      const std::string& /*sql*/,
      const std::vector<RowVectorPtr>& /*input*/,
      const RowTypePtr& /*resultType*/) {
    VELOX_UNSUPPORTED();
  }
```

- Removes the Spark source file download. Instead, photo files are introduced 
in Velox and need to be updated for every Spark version. Currently, these files
are imported from Spark 3.5.1 version.

Fixes https://github.com/facebookincubator/velox/issues/9270.

PR #13118, opened 2025-04-23
fix: Support get Spark from VELOX_SPARK_URL env

liujiayi771:
Velox supports using the `velox_resolve_dependency_url` method to set a custom package download url, which can be configured to a local file. In offline compilation scenarios, this avoids downloading the Spark package. To use this method, it is necessary to define the `SOURCE_URL` and `SHA256_CHECKSUM` variables according to specific rules.


liujiayi771:
cc @rui-mo

rui-mo (commented):
Hi @liujiayi771, since the download of Spark is going to be removed in https://github.com/facebookincubator/velox/pull/9559, we may not need to improve the download process further.

PR #13119, opened 2025-04-23
Change the default encoding of boolean type from PLAIN to RLE

Super-long:
The current version already supports RLE encoding, changing the default encoding of bool from PLAIN to RLE.


Super-long:
Cc @majetideepak

PR #13121, opened 2025-04-23
fix: Fix semi join result mismatch with filter and multi duplicated matched rows

JkSelf:
In a semi join, only the first matching record is retained when there are multiple matching rows on the other side. [ PR#13096](https://github.com/facebookincubator/velox/pull/13096) addresses this issue by selecting only the last matched row for the final output, which resolves result mismatch issue when no filter expression is present. However, this approach can not result mismatches if a filter is applied.
For example:
```
Left Record	    Right Record
a      b	       c      d
2      5	       2      4
                       2      5
```
With the join condition `a == c `and the filter `b > d`, selecting only the last matched row results in right side `(2, 5)` which does not satisfy the filter, leading to an empty result. The correct result should be `(2, 5, 2, 4)`.

This PR ensures that all matched rows are retained when a filter expression exists. During the filtering process, only the first matched row is kept.

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -20,6 +20,25 @@
 
 namespace facebook::velox::exec {
 
+namespace {
+void copyRow(
+    const RowVectorPtr& source,
+    vector_size_t sourceIndex,
+    const RowVectorPtr& target,
+    vector_size_t targetIndex,
+    const std::vector<IdentityProjection>& projections) {
+  for (const auto& projection : projections) {
+    const auto& sourceChild = source->childAt(projection.inputChannel);
+    const auto& targetChild = target->childAt(projection.outputChannel);
+    targetChild->copy(sourceChild.get(), targetIndex, sourceIndex, 1);
+  }
+}
+
+bool isSemiFilterJoin(core::JoinType joinType) {
+  return isLeftSemiFilterJoin(joinType) || isRightSemiFilterJoin(joinType);
+}
+} // namespace
+
 MergeJoin::MergeJoin(
     int32_t operatorId,
     DriverCtx* driverCtx,
@@ -91,8 +110,7 @@ void MergeJoin::initialize() {
 
     if (joinNode_->isLeftJoin() || joinNode_->isAntiJoin() ||
         joinNode_->isRightJoin() || joinNode_->isFullJoin() ||
-        joinNode_->isLeftSemiFilterJoin() ||
-        joinNode_->isRightSemiFilterJoin()) {
+        isSemiFilterJoin(joinType_)) {
       joinTracker_ = JoinTracker(outputBatchSize_, pool());
     }
   } else if (joinNode_->isAntiJoin()) {
@@ -275,25 +293,6 @@ bool MergeJoin::findEndOfMatch(
   match.complete = true;
   return true;
 }
-
-namespace {
-void copyRow(
-    const RowVectorPtr& source,
-    vector_size_t sourceIndex,
-    const RowVectorPtr& target,
-    vector_size_t targetIndex,
-    const std::vector<IdentityProjection>& projections) {
-  for (const auto& projection : projections) {
-    const auto& sourceChild = source->childAt(projection.inputChannel);
-    const auto& targetChild = target->childAt(projection.outputChannel);
-    targetChild->copy(sourceChild.get(), targetIndex, sourceIndex, 1);
-  }
-}
-
-bool isSemiFilterJoin(core::JoinType joinType) {
-  return isLeftSemiFilterJoin(joinType) || isRightSemiFilterJoin(joinType);
-}
-} // namespace
 
 inline void addNull(
     VectorPtr& target,
@@ -1354,11 +1353,11 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
       }
     };
 
-    auto onMatch = [&](auto row, bool hasMatched_) {
+    auto onMatch = [&](auto row, bool firstMatch) {
       const bool isNonSemiAntiJoin =
           !isSemiFilterJoin(joinType_) && !isAntiJoin(joinType_);
 
-      if ((isSemiFilterJoin(joinType_) && hasMatched_) || isNonSemiAntiJoin) {
+      if ((isSemiFilterJoin(joinType_) && firstMatch) || isNonSemiAntiJoin) {
         rawIndices[numPassed++] = row;
       }
     };
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -90,7 +90,9 @@ void MergeJoin::initialize() {
     initializeFilter(joinNode_->filter(), leftType, rightType);
 
     if (joinNode_->isLeftJoin() || joinNode_->isAntiJoin() ||
-        joinNode_->isRightJoin() || joinNode_->isFullJoin()) {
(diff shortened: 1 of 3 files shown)


JkSelf:
@rui-mo @jinchengchenghh @zhli1142015 Can you help to review this PR? Thanks.

JkSelf:
@pedroerp @xiaoxmeng Can you help to review this PR? Thanks.

velox/exec/MergeJoin.h
```diff
@@ -394,11 +394,12 @@ class MergeJoin : public Operator {
     // rows that correspond to a single left-side row. Use
     // 'noMoreFilterResults' to make sure 'onMiss' is called for the last
     // left-side row.
-    template <typename TOnMiss>
+    template <typename TOnMiss, typename TOnMatch>
     void processFilterResult(
         vector_size_t outputIndex,
         bool passed,
-        TOnMiss onMiss) {
+        TOnMiss onMiss,
```
xiaoxmeng:
```
const TOnMiss& onMiss,
const TOnMatch& onMatch
```

velox/exec/MergeJoin.cpp
```diff
@@ -1371,26 +1384,27 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
...
+    // 1. If leftMatch_ is nullopt, there for sure the next buffer will
+    // contain a different key match.
     //
     // 2. leftMatch_ may not be nullopt, but may be related to a different
     // (subsequent) left key. So we check if the last row in the batch has the
     // same left row number as the last key match.
     if (!leftMatch_ || !joinTracker_->isCurrentLeftMatch(numRows - 1)) {
       joinTracker_->noMoreFilterResults(onMiss);
     }
+
```
xiaoxmeng: remove this?

velox/exec/MergeJoin.h
```diff
@@ -407,12 +408,18 @@ class MergeJoin : public Operator {
...
+        firstMatched_ = false;
       } else {
         currentRow_ = outputIndex;
       }
 
       if (passed) {
         currentRowPassed_ = true;
+
+        if (!firstMatched_) {
+          onMatch(outputIndex);
```
xiaoxmeng: Can we pass firstMatched_ flag to onMatch and let each join type to decide how to proceed inside onMatch call instead of here? Thanks!

xiaoxmeng (commented):
@JkSelf thanks for the fix. LGTM % one minor.

velox/exec/MergeJoin.h
JkSelf: @xiaoxmeng Yes. Updated.

velox/exec/MergeJoin.cpp
JkSelf: Removed.

velox/exec/MergeJoin.h
JkSelf: Updated.

JkSelf:
@xiaoxmeng Thanks for your review. Have resolved all your comments. Can you help to review again? Thanks.

velox/exec/MergeJoin.cpp
```diff
@@ -1346,18 +1351,29 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
       }
     };
 
+    auto onMatch = [&](auto row, bool& firstMatched_) {
```
xiaoxmeng: Let's control the firstMatched_ flag in caller context. Also let's add matching rows here for other join types if possible?

JkSelf: Yes. Updated.

velox/exec/MergeJoin.cpp
```diff
@@ -1346,21 +1351,27 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
       }
     };
 
+    auto onMatch = [&](auto row, bool firstMatched_) {
+      bool isSemiJoin =
```
xiaoxmeng: Can you provide an anonymous function say isSemiFilterJoin?

velox/exec/MergeJoin.cpp
```diff
@@ -1346,21 +1351,27 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
...
+          isLeftSemiFilterJoin(joinType_) || isRightSemiFilterJoin(joinType_);
+      bool isNonSemiAntiJoin = !isSemiJoin && !isAntiJoin(joinType_);
```
xiaoxmeng: const bool ...

velox/exec/MergeJoin.h
```diff
@@ -407,12 +408,15 @@ class MergeJoin : public Operator {
         currentRow_ = outputIndex;
         currentLeftRowNumber_ = rowNumber;
         currentRowPassed_ = false;
+        firstMatched_ = false;
       } else {
         currentRow_ = outputIndex;
       }
 
       if (passed) {
```
xiaoxmeng, line 415:
```
      if (passed) {
        onMatch(outputIndex, !currentRowPassed_);
        currentRowPassed_ = true;
      }
```
I don't think we need firstMatched_?

JkSelf: @xiaoxmeng Yes. Removed `firstMatched_`

velox/exec/MergeJoin.cpp
```diff
@@ -1346,21 +1354,26 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
       }
     };
 
+    auto onMatch = [&](auto row, bool hasMatched_) {
```
xiaoxmeng:
```
s/hasMatched_/firstMatch/
```

velox/exec/MergeJoin.cpp
```diff
@@ -90,7 +90,9 @@ void MergeJoin::initialize() {
     initializeFilter(joinNode_->filter(), leftType, rightType);
 
     if (joinNode_->isLeftJoin() || joinNode_->isAntiJoin() ||
-        joinNode_->isRightJoin() || joinNode_->isFullJoin()) {
+        joinNode_->isRightJoin() || joinNode_->isFullJoin() ||
+        joinNode_->isLeftSemiFilterJoin() ||
```
xiaoxmeng: Can you replace here?

velox/exec/MergeJoin.h
```diff
@@ -412,6 +413,7 @@ class MergeJoin : public Operator {
...
+        onMatch(outputIndex, !currentRowPassed_);
```
xiaoxmeng:
```
/*firstMatch=*/!currentRowPassed_
```

JkSelf:
@xiaoxmeng Thank you for helping review and land this PR!

FelixYBW:
@JkSelf with this PR merged, do we still need #11771 

Can we remove oap/velox fix? https://github.com/facebookincubator/velox/commit/a57553543da1c0c42448de195569ba610193b2b7

JkSelf:
> @JkSelf with this PR merged, do we still need #11771
> 
> Can we remove oap/velox fix? [a575535](https://github.com/facebookincubator/velox/commit/a57553543da1c0c42448de195569ba610193b2b7)

@FelixYBW Yes. We still need https://github.com/facebookincubator/velox/pull/11771 to fix the anti and full outer join result mismatch issue. And I have rebased https://github.com/facebookincubator/velox/pull/11771 and resolved the conflicts.

PR #13122, opened 2025-04-23
feat: add memory allocator size stats to periodic stats reporter

Super-long:
This PR implements adding MemoryAllocator SizeStat statistics to PeriodicStatsReporter. Because there are 20*4=80 SizeStat related information, macros are used for implementation;

Currently commented, unit testing has not been implemented yet, and will be improved after confirming that the code implementation idea is correct;

I think the following need to be aligned:
1. Naming rules in counters.h.
2. Whether it is necessary to add a switch in PeriodicStatsReporter.Options to allow external shutdown of SizeStat related reporting.


Super-long:
Cc @majetideepak

PR #13123, opened 2025-04-23
fix: Fix anti join result mismatch with filter and multi duplicated rows

JkSelf:
For an anti join, if a row in the left table has multiple matched rows in the right table, and these matched rows are distributed across different batches, we need to traverse all matched rows in different batches to determine the final result.
For example, in the case below, the left table has columns a and b, and the right table has columns c and d. The join condition is a == c and b < d:

```
a b    c  d
1 3    1  2
       1  2
       1  4
```
Meanwhile, the output batch size is 2, so the first batch1 is:
```
a b c d
1 3 1 2
1 3 1 2
```
If we do not wait for the subsequent batch, such as `1 3 1 4`, the result for batch1 would output a record `1 3`. However, the final result should be null because `1 3 1 4` satisfies `b < d`.

The current PR relies on the joinTracker mechanism to ensure that the final result is output only after all matched rows have been traversed.

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -1281,7 +1281,7 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
     // If all matches for a given left-side row fail the filter, add a row to
     // the output with nulls for the right-side columns.
     const auto onMiss = [&](auto row) {
-      if (isAntiJoin(joinType_) || isSemiFilterJoin(joinType_)) {
+      if (isSemiFilterJoin(joinType_)) {
         return;
       }
       rawIndices[numPassed++] = row;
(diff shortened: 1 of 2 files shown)


jinchengchenghh:
Is this one discarded?

JkSelf:
> Is this one discarded?

No. this pr is to resolve the anti join result mismatch issue. Could you help to review? Thanks.

velox/exec/MergeJoin.cpp
```diff
@@ -1368,12 +1368,6 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
             decodedFilterResult_.valueAt<bool>(i);
 
         joinTracker_->processFilterResult(i, passed, onMiss, onMatch);
-
-        if (isAntiJoin(joinType_)) {
```
xiaoxmeng, line 1372: What's the fix? It seems that we move the add rows into miss callback?

JkSelf: @xiaoxmeng Yes, placing the code to add rows within the onMiss function ensures that a row is added after all matched rows have been processed.

xiaoxmeng: Thanks for the offline explanation!

PR #13126, opened 2025-04-23
feat: Enable maximum elements for repeat configurable

juwentus1234:
The diff will make maxElementsSize configurable so that user could formally use large number.

Enable maximum elements for sequence configurable will be sent in a subsequent diff.

Differential Revision: D73493321

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -27,6 +27,10 @@ Generic Configuration
      - 10000
      - Max number of rows that could be return by operators from Operator::getOutput. It is used when an estimate of
        average row size is known and preferred_output_batch_bytes is used to compute the number of output rows.
+   * - max_elements_size_in_repeat_and_sequence
+     - integer
+     - 10000
+     - Max number of elements that can be set in `repeat` and `sequence` functions.
    * - table_scan_getoutput_time_limit_ms
      - integer
      - 5000
--- a/velox/functions/lib/Repeat.cpp
+++ b/velox/functions/lib/Repeat.cpp
@@ -23,10 +23,9 @@ class RepeatFunction : public exec::VectorFunction {
  public:
   // @param allowNegativeCount If true, negative 'count' is allowed
   // and treated the same as zero (Spark's behavior).
-  explicit RepeatFunction(bool allowNegativeCount)
-      : allowNegativeCount_(allowNegativeCount) {}
-
-  static constexpr int32_t kMaxResultEntries = 10'000;
+  explicit RepeatFunction(bool allowNegativeCount, int32_t maxElementsSize)
+      : allowNegativeCount_(allowNegativeCount),
+        maxElementsSize_(maxElementsSize) {}
 
   void apply(
       const SelectivityVector& rows,
@@ -48,7 +47,7 @@ class RepeatFunction : public exec::VectorFunction {
 
  private:
   // Check count to make sure it is in valid range.
-  static int32_t checkCount(int32_t count, bool allowNegativeCount) {
+  int32_t checkCount(int32_t count, bool allowNegativeCount) const {
     if (count < 0) {
       if (allowNegativeCount) {
         return 0;
@@ -58,10 +57,12 @@ class RepeatFunction : public exec::VectorFunction {
           count,
           0);
     }
+
     VELOX_USER_CHECK_LE(
         count,
-        kMaxResultEntries,
-        "Count argument of repeat function must be less than or equal to 10000");
+        maxElementsSize_,
+        "Count argument of repeat function must be less than or equal to {}",
+        maxElementsSize_);
     return count;
   }
 
@@ -122,9 +123,11 @@ class RepeatFunction : public exec::VectorFunction {
     exec::DecodedArgs decodedArgs(rows, args, context);
     auto countDecoded = decodedArgs.at(1);
     int32_t totalCount = 0;
+
     context.applyToSelectedNoThrow(rows, [&](auto row) {
       auto count =
           countDecoded->isNullAt(row) ? 0 : countDecoded->valueAt<int32_t>(row);
+
       count = checkCount(count, allowNegativeCount_);
       totalCount += count;
     });
@@ -181,6 +184,7 @@ class RepeatFunction : public exec::VectorFunction {
   }
 
   const bool allowNegativeCount_;
+  const int32_t maxElementsSize_;
 };
 } // namespace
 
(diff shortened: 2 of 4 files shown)


velox/functions/lib/Repeat.cpp
```diff
@@ -125,7 +131,7 @@ class RepeatFunction : public exec::VectorFunction {
     context.applyToSelectedNoThrow(rows, [&](auto row) {
       auto count =
           countDecoded->isNullAt(row) ? 0 : countDecoded->valueAt<int32_t>(row);
-      count = checkCount(count, allowNegativeCount_);
+      count = checkCount(count, allowNegativeCount_, context);
```
pedroerp, line 134: maybe here it would be better to save the limit as a local/member variable outside of the loop, just in case?

velox/functions/lib/tests/RepeatTest.cpp
```diff
@@ -97,6 +97,23 @@ TEST_F(RepeatTest, repeat) {
...
+      {over_default_limit,
+       over_default_limit,
+       over_default_limit,
+       std::nullopt});
+  VectorPtr expected;
+
+  testExpressionWithError(
+      "repeat(C0, C1)",
+      {elementVector, countVector},
+      "Count argument of repeat function must be less than or equal to 10000");
```
pedroerp, line 114: it would be nice if the test case changed the query option, then ensured that the query now passes.

velox/functions/lib/Repeat.cpp
```diff
@@ -48,7 +46,8 @@ class RepeatFunction : public exec::VectorFunction {
 
  private:
   // Check count to make sure it is in valid range.
-  static int32_t checkCount(int32_t count, bool allowNegativeCount) {
+  static int32_t
+  checkCount(int32_t count, bool allowNegativeCount, exec::EvalCtx& context) {
```
kagamiori, line 50: The makeRepeat() method receives a core::QueryConfig. Can we just get the maxElementsSizeInRepeatAndSequence() from this query config there and pass it to the constructor of RepeatFunction? We can have a member variable in RepeatFunction to keep this max size.

velox/docs/configs.rst
```diff
@@ -27,6 +27,10 @@ Generic Configuration
      - 10000
      - Max number of rows that could be return by operators from Operator::getOutput. It is used when an estimate of
        average row size is known and preferred_output_batch_bytes is used to compute the number of output rows.
+   * - max_elements_size_in_repeat_and_sequence
+     - integer
+     - 10000
+     - Max number of elements can be set in repeat function and sequence function.
```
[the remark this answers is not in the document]
juwentus1234: updated

velox/functions/lib/Repeat.cpp
juwentus1234: updated using local variable

velox/functions/lib/tests/RepeatTest.cpp
juwentus1234: updated, not so easy to add it to query option in this test.

velox/functions/lib/Repeat.cpp
juwentus1234: I feel like in previous case in reduce, it was just used a local variable though. Since it could take out from QueryConfig, I kind of feel like member variable is not needed?

velox/functions/lib/tests/RepeatTest.cpp
```diff
@@ -97,6 +97,33 @@ TEST_F(RepeatTest, repeat) {
...
+  const auto elementVector =
+      makeNullableFlatVector<float>({0.0, 2.0, std::nullopt, 3.0});
+  const int32_t over_default_limit = 15'200;
+  auto countVector = makeNullableFlatVector<int32_t>(
+      {over_default_limit,
+       over_default_limit,
+       over_default_limit,
+       std::nullopt});
+
+  auto valuesMap = execCtx_.queryCtx()->queryConfig().rawConfigsCopy();
```
pedroerp, line 110: you can probably use `testingOverrideConfigUnsafe()`:

https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/tests/utils/FunctionBaseTest.h#L52

velox/functions/lib/Repeat.cpp
```diff
@@ -122,10 +128,15 @@ class RepeatFunction : public exec::VectorFunction {
     exec::DecodedArgs decodedArgs(rows, args, context);
     auto countDecoded = decodedArgs.at(1);
     int32_t totalCount = 0;
+    const int32_t maxElementsSize = context.execCtx()
```
pedroerp, line 131: nit: perhaps wrap as a tiny helper method

`auto maxElementsSize = getMaxElementsSize(context);`

velox/functions/lib/tests/RepeatTest.cpp
juwentus1234: updated, thanks Pedro!

velox/functions/lib/Repeat.cpp
juwentus1234: updated

velox/functions/lib/tests/RepeatTest.cpp
```diff
@@ -97,6 +97,31 @@ TEST_F(RepeatTest, repeat) {
...
+  valuesMap["max_elements_size_in_repeat_and_sequence"] = "15000";
+  execCtx_.queryCtx()->testingOverrideConfigUnsafe(std::move(valuesMap));
```
pedroerp, line 112: nit: you can probably have this inline:

```
execCtx_.queryCtx()->testingOverrideConfigUnsafe({
  {"max_elements_size_in_repeat_and_sequence", "15000"},
});
```

pedroerp: also, please use core::QueryConfig::kMaxElementsSizeInRepeatAndSequence instead of repeating the literal string.

velox/functions/lib/Repeat.cpp
```diff
@@ -79,9 +88,14 @@ class RepeatFunction : public exec::VectorFunction {
       return BaseVector::createNullConstant(outputType, numRows, pool);
     }
 
+    const auto maxElementsSize = context.execCtx()
```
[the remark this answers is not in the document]
juwentus1234: updated, not sure why this was not being exported initially.

velox/functions/lib/tests/RepeatTest.cpp
juwentus1234: updated

velox/functions/lib/tests/RepeatTest.cpp
```diff
@@ -97,6 +97,31 @@ TEST_F(RepeatTest, repeat) {
...
+
+  auto data = makeRowVector({elementVector, countVector});
+  auto typedExpr = makeTypedExpr("repeat(C0, C1)", asRowType(data->type()));
+
+  exec::ExprSet exprSet({typedExpr}, &execCtx_);
+  exec::EvalCtx evalCtx(&execCtx_, &exprSet, data.get());
+
+  VELOX_ASSERT_THROW(
+      evaluate(exprSet, data),
+      "Count argument of repeat function must be less than or equal to 15000");
```
kagamiori, line 122: We should also check that repeat() doesn't throw when count argument is between 10000 and 15000.

velox/functions/lib/Repeat.cpp
pedroerp: Agreed with that, but I think we only get a pointer to QueryConfig when the apply() function is called. We don't have that in the constructor; or do we?

velox/functions/lib/tests/RepeatTest.cpp
juwentus1234: updated adding the case

velox/functions/lib/Repeat.cpp
```diff
@@ -48,7 +47,8 @@ class RepeatFunction : public exec::VectorFunction {
 
  private:
   // Check count to make sure it is in valid range.
-  static int32_t checkCount(int32_t count, bool allowNegativeCount) {
+  static int32_t
```
pedroerp, line 50: probably better to make this function non-static now so you can access the member and not having to pass it as a parameter.

pedroerp (approved):
Small comments, but looks good!

velox/functions/lib/Repeat.cpp
juwentus1234: updated

PR #13128, opened 2025-04-23
fix: Add SQL generation case for CallExpr with function switch

peterenescu:
Summary:
Add case to function `toCallSql` of PrestoSql to generate special SQL syntax for function `switch`. In Velox expression evaluation, the function is registered as `switch`, while in Presto Java there is no such conception. `switch` in Presto refers to case blocks. For example,
`case when true then "hello" end`
`case when true then "hello" when false then "goodbye" else "wave" end`
This PR adds SQL syntax for the above. It also handles the special case where two arguments passed to `switch` is treated as an `if` clause. This will fix our cogwheel runs that invoke special form `switch` and are currently failing due to the errors noted below.

Differential Revision: D73517749

### DIFF
--- a/velox/exec/fuzzer/PrestoSql.cpp
+++ b/velox/exec/fuzzer/PrestoSql.cpp
@@ -309,6 +309,24 @@ std::string toCallSql(const core::CallTypedExprPtr& call) {
     sql << "[";
     toCallInputsSql({call->inputs()[1]}, sql);
     sql << "]";
+  } else if (call->name() == "switch") {
+    VELOX_CHECK_GE(
+        call->inputs().size(),
+        2,
+        "Expected at least two arguments to function 'switch'");
+    sql << "case";
+    int i = 0;
+    for (; i < call->inputs().size() - 1; i += 2) {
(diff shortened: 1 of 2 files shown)


velox/exec/fuzzer/PrestoSql.cpp
```diff
@@ -309,6 +309,33 @@ std::string toCallSql(const core::CallTypedExprPtr& call) {
...
+        2,
+        "Expected at least two arguments to function 'switch'");
+    // If two inputs, handle like an "if"
+    if (call->inputs().size() == 2) {
+      sql << "if(";
+      toCallInputsSql({call->inputs()[0]}, sql);
+      sql << ", ";
+      toCallInputsSql({call->inputs()[1]}, sql);
+      sql << ")";
+    } else {
```
kagamiori, line 324: This special case doesn't seem to be needed. For switch with two inputs, we can still convert it to "case when C then B1 end".

kagamiori (approved):
LGTM. Thank you for fixing the SQL translation for switch.

PR #13130, opened 2025-04-23
feat: Add insertable iteratable spill partition set

tanjialiang:
Summary: Adding a IterableSpillPartitionSet that is used in recursive spilling scenario. Capable of handling spill partition insertion while handling the partition retrieving order such that deeper and lower partition number partitions are retrieved first. It also preserves the entire leaf spill partition set that could be retrieved by calling 'extractAll()'.

Differential Revision: D73414276


xiaoxmeng (approved):
@tanjialiang thanks for the update % left some comments

PR #13131, opened 2025-04-23
feat: Pass mapping from PlanNodeId to ExchangeClient to TaskListener::onCompletion

kagamiori:
We're using TaskListener to log query plan fragments so that we can rerun the query 
later. To do that, we need to know the relation among plan fragments, i.e., the 
producers and consumers of shuffles. This diff makes TaskListener::onCompletion 
receives a map from consumer PlanNode id to a ExchangeClient with which we can 
log the mapping from consumer PlanNode id to a list of producer Task id in 
TaskListener.

Differential Revision: D69828745


kevinwilfong (approved):
Thanks!

PR #13137, opened 2025-04-24
misc: Allow using Task::addSplitWithSequence through AssertQueryBuilder

kagamiori:
This diff extends the API of AssertQueryBuilder to allow exercising the Task::addSplitWithSequence code path.

Differential Revision: D73052401

### DIFF
--- a/velox/exec/tests/utils/AssertQueryBuilder.cpp
+++ b/velox/exec/tests/utils/AssertQueryBuilder.cpp
@@ -178,6 +178,12 @@ AssertQueryBuilder& AssertQueryBuilder::splits(
   return *this;
 }
 
+AssertQueryBuilder& AssertQueryBuilder::addSplitWithSequence(
+    bool addWithSequence) {
+  addSplitWithSequence_ = addWithSequence;
+  return *this;
+}
+
 std::shared_ptr<Task> AssertQueryBuilder::assertResults(
     const std::string& duckDbSql,
     const std::optional<std::vector<uint32_t>>& sortingKeys) {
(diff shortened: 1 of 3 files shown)


velox/exec/tests/utils/AssertQueryBuilder.h
```diff
@@ -160,39 +160,49 @@ class AssertQueryBuilder {
   /// duckDbQueryRunner to be provided in the constructor.
   std::shared_ptr<Task> assertResults(
       const std::string& duckDbSql,
-      const std::optional<std::vector<uint32_t>>& sortingKeys = std::nullopt);
+      const std::optional<std::vector<uint32_t>>& sortingKeys = std::nullopt,
```
pedroerp, line 163: since this uses the= builder pattern, would it be cleaner to have addSplitWithSequence as a separate API, instead of carrying that as parameter all over? Something like:

```
AssertQueryBuilder().addSplitsWithSequence().assertResults(expected);
```

?

kagamiori: Hi @pedroerp, thank you for the suggestion. I've updated the PR. Could you take another look?

velox/exec/tests/utils/AssertQueryBuilder.cpp
```diff
@@ -178,6 +178,12 @@ AssertQueryBuilder& AssertQueryBuilder::splits(
   return *this;
 }
 
+AssertQueryBuilder& AssertQueryBuilder::addSplitWithSequence(
+    bool addWithSequence) {
```
pedroerp, line 182: maybe set the default value to true, for convenience?

pedroerp (approved):
Thank you

PR #13138, opened 2025-04-24
fix: Fix the full outer join result mismatch issue with multi duplicated match

JkSelf:
Assume the left table has columns a and b:
```
a       b
2	100
2	1
2	1
```
The right table has columns c and d:
```
c       d
2	3
2	-1
2	-1
2	3
```

The two tables are joined using a full outer join on the condition a == c and b < d. During the doGetOutput phase, the result is matched using a left join, resulting in 3 * 4 = 12 records:

```
No      a        b       c      d
0	2	100	 2	3
1	2	100	 2	-1
2	2	100	 2	-1
3	2	100	 2	3
4	2	1	 2	3
5	2	1	 2	-1
6	2	1	 2	-1
7	2	1	 2	3
8	2	1	 2	3
9	2	1	 2	-1
10	2	1	 2	-1
11	2	1	 2	3
```

Then, in the filter method, the records are filtered based on the condition b < d, resulting in the following:

```
No	a	b	c	d	matched
0	2	100	2	3	FALSE
1	2	100	2	-1	FALSE
2	2	100	2	-1	FALSE
3	2	100	2	3	FALSE
4	2	1	2	3	TRUE
5	2	1	2	-1	FALSE
6	2	1	2	-1	FALSE
7	2	1	2	3	TRUE
8	2	1	2	3	TRUE
9	2	1	2	-1	FALSE
10	2	1	2	-1	FALSE
11	2	1	2	3	TRUE
```

Finally, records from the left table that do not have a match are filled with nulls, resulting in the following final output:
```
No	a	b	c	 d
0	2	100	null null
1	2	1	2	 3
2	2	1	2	 3
3	2	1	2	 3
4	2	1	2	 3
```
The above result is incorrect because it is missing rows from the right table that do not have a match. Among the 12 rows above, rows 0, 4, and 8 correspond to the first record (2, 3) from the right table, rows 1, 5, and 9 correspond to the second record (2, -1) from the right table, rows 2, 6, and 10 correspond to the third record (2, -1) from the right table, and rows 3, 7, and 11 correspond to the fourth record (2, 3) from the right table. From the matching results above, rows 1, 5, and 9, as well as rows 2, 6, and 10, are all false, meaning that the third and fourth records from the right table do not have matching rows. Therefore, the final result is missing rows from the right table that do not have matches. The correct final result should be:
```
No	a	b	c	 d
0	2	100	null   null
1	2	1	2	 3
2	2	1	2	 3
3	2	1	2	 3
4	2	1	2	 3
5       null    null    2       -1
6       null    null    2       -1
```

This PR calls the filter function when the keys are the same to filter out rows from the right table that do not have matches. If a row from the right table does not have a match, a new record is inserted with the corresponding columns from the left table set to null.

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -113,9 +113,10 @@ void MergeJoin::initialize() {
         isSemiFilterJoin(joinType_)) {
       joinTracker_ = JoinTracker(outputBatchSize_, pool());
     }
-  } else if (joinNode_->isAntiJoin()) {
+  } else if (joinNode_->isAntiJoin() || joinNode_->isFullJoin()) {
     // Anti join needs to track the left side rows that have no match on the
-    // right.
+    // right. Full outer join needs to track the right side rows that have no
+    // match on the left.
     joinTracker_ = JoinTracker(outputBatchSize_, pool());
   }
 
@@ -392,7 +393,8 @@ bool MergeJoin::tryAddOutputRow(
     const RowVectorPtr& leftBatch,
     vector_size_t leftRow,
     const RowVectorPtr& rightBatch,
-    vector_size_t rightRow) {
+    vector_size_t rightRow,
+    bool isRightJoinForFullOuter) {
   if (outputSize_ == outputBatchSize_) {
     return false;
   }
@@ -426,12 +428,15 @@ bool MergeJoin::tryAddOutputRow(
         filterRightInputProjections_);
 
     if (joinTracker_) {
-      if (isRightJoin(joinType_) || isRightSemiFilterJoin(joinType_)) {
+      if (isRightJoin(joinType_) || isRightSemiFilterJoin(joinType_) ||
+          (isFullJoin(joinType_) && isRightJoinForFullOuter)) {
         // Record right-side row with a match on the left-side.
-        joinTracker_->addMatch(rightBatch, rightRow, outputSize_);
+        joinTracker_->addMatch(
+            rightBatch, rightRow, outputSize_, isRightJoinForFullOuter);
       } else {
         // Record left-side row with a match on the right-side.
-        joinTracker_->addMatch(leftBatch, leftRow, outputSize_);
+        joinTracker_->addMatch(
+            leftBatch, leftRow, outputSize_, isRightJoinForFullOuter);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


velox/exec/MergeJoin.cpp
```diff
@@ -18,6 +18,8 @@
 #include "velox/exec/Task.h"
 #include "velox/expression/FieldReference.h"
 
+#include <iostream>
```
jinchengchenghh, line 21: Unexpected change?

velox/exec/MergeJoin.cpp
```diff
@@ -92,7 +94,7 @@ void MergeJoin::initialize() {
         joinNode_->isRightJoin() || joinNode_->isFullJoin()) {
       joinTracker_ = JoinTracker(outputBatchSize_, pool());
     }
-  } else if (joinNode_->isAntiJoin()) {
+  } else if (joinNode_->isAntiJoin() || joinNode_->isFullJoin()) {
     // Anti join needs to track the left side rows that have no match on the
```
jinchengchenghh, line 98: Please also update the comment

jinchengchenghh (commented):
Can you a brief description to the PR description?

JkSelf:
@jinchengchenghh Resolved all your comments. Can you help to review again? Thanks.

jinchengchenghh (commented):
Since I'm not familiar with the join, I can only do the code review, not the logic.

JkSelf:
@jinchengchenghh Thanks for your review. Resolved all your comments. Thanks.

jinchengchenghh:
LGTM, Thanks!

velox/exec/MergeJoin.h
```diff
@@ -459,6 +471,9 @@ class MergeJoin : public Operator {
     BufferPtr leftRowNumbers_;
     vector_size_t* rawLeftRowNumbers_;
 
+    BufferPtr rightJoinRows_;
```
jinchengchenghh, line 474: Is it rightRowNumbers? If that, rename it to match the above leftRowNumbers_

JkSelf:
@jinchengchenghh Thanks for your review. I will close this pr and fix this issue in https://github.com/facebookincubator/velox/pull/13907. Please help to review. Thanks.

PR #13142, opened 2025-04-24
fix(arrow): Use `VELOX_DCHECK*` instead of `DCHECK*`

kou:
`DCHECK*` are for internal use. Users should not use them.

FYI: `DCHECK*` will be hidden since Apache Arrow C++ 20.0.0: https://github.com/apache/arrow/pull/46015

We should use `VELOX_DCHECK*` instead of `DCHECK*` provided by Apache Arrow because we forked Apache Parquet C++ to adapt our use case.

Here are basic conversion rules for `DCHECK*` -> `VELOX_DCHECK*`:

* `DCHECK*` -> `VELOX_DCHECK*`
* `DCHECK*() << "XXX"` -> `VELOX_DCHECK*(..., "XXX")`
* `DCHECK_EQ(XXX, nullptr)` -> `VELOX_DCHECK_NULL(XXX)`
* `DCHECK_NE(XXX, nullptr)` -> `VELOX_DCHECK_NOT_NULL(XXX)`

We need some `format_as()`s for `VELOX_DCHECK_*(ENUM_VALUE1, ENUM_VALUE2)`.

This also converts `DCHECK(XXX != nullptr)` to `VELOX_DCHECK_NOT_NULL(XXX)` for better error message.

### DIFF
--- a/velox/dwio/parquet/writer/arrow/Properties.cpp
+++ b/velox/dwio/parquet/writer/arrow/Properties.cpp
@@ -23,7 +23,6 @@
 
 #include "arrow/io/buffered.h"
 #include "arrow/io/memory.h"
-#include "arrow/util/logging.h"
 #include "arrow/util/thread_pool.h"
 
 namespace facebook::velox::parquet::arrow {
--- a/velox/dwio/parquet/writer/arrow/Schema.cpp
+++ b/velox/dwio/parquet/writer/arrow/Schema.cpp
@@ -27,7 +27,6 @@
 #include <type_traits>
 #include <utility>
 
-#include "arrow/util/logging.h"
 #include "velox/dwio/parquet/writer/arrow/Exception.h"
 
 using facebook::velox::parquet::thrift::SchemaElement;
--- a/velox/dwio/parquet/writer/arrow/util/Hashing.h
+++ b/velox/dwio/parquet/writer/arrow/util/Hashing.h
@@ -45,6 +45,8 @@
 #include "arrow/util/macros.h"
 #include "arrow/util/ubsan.h"
 
+#include "velox/common/base/Exceptions.h"
+
 #define XXH_INLINE_ALL
 
 #include <xxhash.h>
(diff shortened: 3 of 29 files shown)


kou:
We may want to use `VELOX_DCHECK*` not `ARROW_DCHECK*`.

pedroerp:
AFAIK, the Parquet code was vendored (forked) from Arrow, so we don't necessarily want to use Arrow as a library, but rather incorporate the changes to Velox. So I suppose we should move these to VELOX_DCHECK? Cc: @majetideepak

majetideepak:
> So I suppose we should move these to VELOX_DCHECK

+ 1 for this.
As and when we have been touching the ported Arrow code, we have been migrating to Velox primitives.

kou:
OK. I've changed `ARROW_DCHECK*` to `VELOX_DCHECK*`. Please see the updated PR description how to convert them.

velox/dwio/parquet/writer/arrow/util/Hashing.h
```diff
@@ -252,14 +254,15 @@ class HashTable {
...
+    VELOX_DCHECK_NOT_NULL(pool);
     // Minimum of 32 elements
     capacity = std::max<uint64_t>(capacity, 32UL);
     capacity_ = ::arrow::bit_util::NextPower2(capacity);
     capacity_mask_ = capacity_ - 1;
     size_ = 0;
 
-    DCHECK_OK(UpsizeBuffer(capacity_));
+    auto status = UpsizeBuffer(capacity_);
+    VELOX_DCHECK(status.ok(), status.ToString());
```
kou, line 265: We may want to define `VELOX_DCHECK_OK()` that is a specialized `DCHECK` for `arrow::Status`.

kou:
The CI failure is tracked by https://github.com/facebookincubator/velox/issues/13153 .

velox/dwio/parquet/writer/arrow/Properties.cpp
```diff
@@ -23,9 +23,10 @@
 
 #include "arrow/io/buffered.h"
 #include "arrow/io/memory.h"
-#include "arrow/util/logging.h"
 #include "arrow/util/thread_pool.h"
 
+#include "velox/common/base/Exceptions.h"
```
majetideepak, line 28: nit: Is this header used?

velox/dwio/parquet/writer/arrow/Schema.cpp
```diff
@@ -27,7 +27,7 @@
 #include <type_traits>
 #include <utility>
 
-#include "arrow/util/logging.h"
+#include "velox/common/base/Exceptions.h"
```
majetideepak, line 30: Is this header used?

majetideepak (approved):
Thanks, @kou.
2 nits.

velox/dwio/parquet/writer/arrow/Properties.cpp
kou: This and other `#include` are needless. Sorry. I missed them.

PR #13143, opened 2025-04-24
misc: Pass by value in Checked Arithmetic functions

andersdellien-arm:
Passing by value instead of references gives a performance improvement because the parameters are small (integers and/or floats).

For VectorSlice, this gives about 1% on some of the sub-benchmarks (measured on AWS C7G machine)

### DIFF
--- a/velox/common/base/CheckedArithmetic.h
+++ b/velox/common/base/CheckedArithmetic.h
@@ -24,7 +24,7 @@
 namespace facebook::velox {
 
 template <typename T>
-T checkedPlus(const T& a, const T& b, const char* typeName = "integer") {
+T checkedPlus(T a, T b, const char* typeName = "integer") {
   T result;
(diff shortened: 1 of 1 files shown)


velox/common/base/CheckedArithmetic.h
```diff
@@ -24,7 +24,7 @@
 namespace facebook::velox {
 
 template <typename T>
-T checkedPlus(const T& a, const T& b, const char* typeName = "integer") {
+T checkedPlus(const T a, const T b, const char* typeName = "integer") {
```
Yuhta, line 27: Can you remove the `const` as well?

Yuhta (approved):

velox/common/base/CheckedArithmetic.h
andersdellien-arm: Sure, it's removed in the updated patch

PR #13150, opened 2025-04-24
refactor: Use SpillPartitionId in SpillState to replace simple integer

tanjialiang:
Summary: Use SpillPartitionId in SpillState to replace the originally used integer partition number. This is to generalize spilling framework on supporting the newly introduced SpillPartitionId in both single level spilling and nested level spilling. The changes are limited to Spill.* and Spiller.* and simple temporary adaptors were introduced in some operators to help to limit the changes' propagation to be too much. The next refactor will continue the change.

Differential Revision: D73562773


xiaoxmeng (approved):
@tanjialiang thanks!

PR #13151, opened 2025-04-24
fix: Load lazy input before order by operation

tanjialiang:
Summary: Order by operator is under unreclaimable state when loading lazy input vector. We load input vector at the beginning of addInput() call to fix this corner case scenario. Also extracted the reclaimable lazy loading to a common utility in base Operator class.

Differential Revision: D73606256


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks for the quick fix and test!

PR #13154, opened 2025-04-25
feat(json): Add support for unknown in maps and arrays when casting to json

kgpai:
Summary:
Currently casting an unknown type in map/array will fail in Prestissimo, but succeed in Presto. 
This change adds this support: 
```
SELECT
    CAST(x AS JSON)
FROM (
    VALUES
        CAST(MAP(ARRAY[], ARRAY[]) AS MAP<UNKNOWN, UNKNOWN>)
) AS t(x);
```

Differential Revision: D73629889


amitkdutta (approved):
Thanks @kgpai

PR #13158, opened 2025-04-25
fix: Spark lower function on unicode character

rui-mo:
The behaviour of lower function differs between Presto and Spark for 'İ' 
(\u0130). Presto returns \u0069 and Spark returns \u0069 + \u0307. This PR
adds a template parameter `forSpark` for the `lowerUnicode` function and
`UpperLowerTemplateFunction` class to customize Spark's behaviour.

Spark's behaviour:
```
scala> val df = spark.sql("SELECT lower(CAST(X'C4B0' AS STRING))")
df: org.apache.spark.sql.DataFrame = [lower(CAST(X'C4B0' AS STRING)): string]

scala> df.foreach(row => println(row.getString(0).map(_.toInt.toHexString).mkString("\\u"))) 
69\u307
```

Presto's behaviour:

```
presto> SELECT to_hex(CAST(lower(U&'\0130') AS varbinary)) AS hex_result;
 hex_result 
------------
 69         
```


velox/functions/lib/string/StringCore.h
```diff
@@ -194,6 +195,18 @@ FOLLY_ALWAYS_INLINE size_t lowerUnicode(
     }
 
     inputIdx += size;
+
+    if constexpr (forSpark) {
+      // Handle Turkish-specific case for İ (U+0130).
+      if (UNLIKELY(nextCodePoint == 0x0130)) {
```
jinchengchenghh, line 201: Do you try other combination dot above?https://www.compart.com/en/unicode/U+0307 ,such as T with dot above.

rui-mo:
@jinchengchenghh Would you help take another review? Thanks.

zhli1142015:
will '\u03A3' also trigger different result?
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/StringUTF16.java#L1021C29-L1021C35

velox/functions/prestosql/UpperLower.h
```diff
@@ -0,0 +1,96 @@
...
+#include "velox/functions/lib/string/StringImpl.h"
+
+namespace facebook::velox::functions {
+
+/// Function to convert string to upper or lower case. Ascii and unicode
+/// conversion are supported.
+/// @tparam isLower Instantiate for upper or lower.
+/// @tparam forSpark When true, Spark's specific behavior is considered, e.g.
+/// for 'İ' Spark's lower case is 'i̇' and Presto's is 'i'.
+template <bool isLower, bool forSpark>
```
jinchengchenghh, line 29: Please also add the test when forSpark is false
Do we need to move this file to directory velox/functions/lib

rui-mo:
> will '\u03A3' also trigger different result? https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/StringUTF16.java#L1021C29-L1021C35

Thanks for providing the details. I find the result is consistent with Spark and added unit test for it.

https://github.com/facebookincubator/velox/pull/13158/files#diff-e00fdf6c345591c1152c6b9545963cb04e8c795266e2215b2f52df6140b83ddbR89

velox/functions/prestosql/UpperLower.h
rui-mo: The tests of `forSpark` being false are covered by the `StringImplTest`. Do we need to repeat them?

https://github.com/facebookincubator/velox/pull/13158/files#diff-df8512514193690325471bddf44afcd54a1b37493a9ad8b607f8c0d1cd290a95R41-R47

Moved UpperLower.h to the lib, thanks.

jinchengchenghh: It does not cover I with dot above case, this is the different code path

jinchengchenghh: I got it now.

velox/functions/lib/string/tests/StringImplTest.cpp
```diff
@@ -42,7 +42,33 @@ class StringImplTest : public testing::Test {
         {"àáâãäåæçèéêëìíîïðñòóôõöøùúûüýþ", "ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞ"},
         {"αβγδεζηθικλμνξοπρςστυφχψ", "ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΣΤΥΦΧΨ"},
         {"абвгдежзийклмнопрстуфхцчшщъыьэюя",
-         "АБВГДЕЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ"}};
+         "АБВГДЕЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ"},
+        {"\u0069", "\u0049"},
+        {"\u03C3", "\u03A3"},
+        {"i\xCC\x87", "I\xCC\x87"},
```
jinchengchenghh, line 48: What do you think about use this format instead of using original I with dot above?

velox/functions/lib/string/tests/StringImplTest.cpp
```diff
@@ -42,7 +42,33 @@ class StringImplTest : public testing::Test {
...
+        {"\u010B", "\u010A"},
+        {"\u0117", "\u0116"},
+        {"\u0121", "\u0120"},
+        {"\u017C", "\u017B"},
```
jinchengchenghh: If this is not affected by this change, I would prefer to remove these tests

rui-mo:
@zhli1142015 @jinchengchenghh Do you have any further comment for this change? Thanks.

zhli1142015 (approved):
LGTM, thanks.

rui-mo:
@zhli1142015 Could you help mark it as ready-to-merge if no comment? Thanks!

jinchengchenghh (approved):

velox/functions/lib/string/tests/StringImplTest.cpp
rui-mo: It is a bit hard to read if we use original format. I assume this way would make it much clearer.

velox/functions/lib/string/tests/StringImplTest.cpp
rui-mo: It's OK for me to remove them as these cases are also covered in the function's test.

PR #13159, opened 2025-04-25
fix: Replace space with underscore in generated task id from cursor

tanjialiang:
Summary: The space in current task id generated from task cursor might introduce issues when we use it to generate other stuffs like dir path. Replacing it with '_'.

Differential Revision: D73678898


kostasxmeta (approved):

pedroerp (approved):
Thank you! I was always annoyed by this whitespace but never enough to actually change it :)

PR #13160, opened 2025-04-25
feat: Enable maximum elements for sequence configurable

juwentus1234:
Summary: Enable maximum elements for sequence function configurable, similar to repeat function configurable

Differential Revision: D73682674

### DIFF
--- a/velox/functions/prestosql/Sequence.cpp
+++ b/velox/functions/prestosql/Sequence.cpp
@@ -96,8 +96,6 @@ int128_t getStepCount(Timestamp start, Timestamp end, int32_t step) {
 template <typename T, typename K>
 class SequenceFunction : public exec::VectorFunction {
  public:
-  static constexpr int32_t kMaxResultEntries = 10'000;
-
   void apply(
       const SelectivityVector& rows,
       std::vector<VectorPtr>& args,
@@ -131,7 +129,11 @@ class SequenceFunction : public exec::VectorFunction {
           stepVector,
           row,
           isDate,
-          isIntervalYearMonth);
+          isIntervalYearMonth,
+          context.execCtx()
+              ->queryCtx()
+              ->queryConfig()
+              .maxElementsSizeInRepeatAndSequence());
       numElements += rawSizes[row];
     });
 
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/Sequence.cpp
```diff
@@ -131,7 +129,11 @@ class SequenceFunction : public exec::VectorFunction {
           stepVector,
           row,
           isDate,
-          isIntervalYearMonth);
+          isIntervalYearMonth,
+          context.execCtx()
```
pedroerp, line 133: why don't you capture this in the constructor, like what you did for `repeat`?

juwentus1234: Because here only one place is using it which is minimum change. If we put in the constructor, line 296, 298, 306, 308, all need to change, also change also all static method to non-static, similar usage example in reduce.cpp which uses a local variable. Besides, a member variable takes the memory life longer than a local one. Thus, I don't see a reason that we should use this as member variable. Let me know if you still want this as a member variable:)

velox/functions/prestosql/tests/SequenceTest.cpp
```diff
@@ -125,6 +125,48 @@ TEST_F(SequenceTest, exceedMaxEntries) {
   testExpression("try(sequence(C0, C1))", {startVector, stopVector}, expected);
 }
 
+TEST_F(SequenceTest, sequenceWithEntriesWithMaxElementsSize) {
+  const int32_t within_limit = 13000;
```
pedroerp, line 129: we use camel case for local variables. Check out coding style doc:

https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md

velox/functions/prestosql/Sequence.cpp
pedroerp: Ok. My only point is that the constructor is called once per query, while the apply method is called once per batch. But this is so cheap it probably wouldn't make a difference.

pedroerp (approved):

velox/functions/prestosql/tests/SequenceTest.cpp
juwentus1234: udpated

Issue #13089, opened 2025-04-21
Presto query scheduling hangs due to velox task reporting blocked split as queued split


kewang1024:
### Description

In presto, Coordinator relies on table scan running splits + queued splits signal from worker to decide whether to schedule more split or block. 
There are two threshold configs correspondingly:
- node-scheduler.max-splits-per-node
- node-scheduler.max-pending-splits-per-task

However, on worker, the split and driver models are different between java and cpp, causing the pending splits reporting to be different:
-Java reports queued splits for the "runnable" but "queued" splits, which does not contain "blocked" split
-CPP reports queued splits for any split that is not running, which contain both "blocked" and "queued"

In cases like colocated join, for example we have scheduled 100 splits for both probe and build sides. 
<img width="690" alt="Image" src="https://github.com/user-attachments/assets/9041fb52-e048-4357-9a20-fab9c1730e5e" />
-Java reports 90 queued splits
-CPP reports 180 queued splits, and as build side progresses, probe side receive more splits and queued splits grow more than the threshold, causing the query to hang

PR #13163, opened 2025-04-26
fix: Report task split status based on driver status

kewang1024:
In presto, table scan split status has three types: running, queued and blocked. And presto worker only reports running and queued split info to coordinator scheduler.

Velox calculates running and queued split differently and it’s based on addSplitLocked, getSplitLocked and splitFinished API:
- addSplitLocked will put the split into splitStore then queued++
- getSplitLocked will fetch split from splitStore then running++ and queued--
- splitFinished running—

As a result:
- queued contains both queued and blocked state
- running contains both running and blocked state

Change to calculate split status based on driver status:
- Count splits which are on running drivers as runningSplit
- Count running table scan node's remaining splits in splitStore as queuedSplit

In this way, we will be aligned with presto’s reporting and in the future, we can add blocked splits if needed

Issue
https://github.com/facebookincubator/velox/issues/13089

### DIFF
--- a/velox/exec/TableScan.h
+++ b/velox/exec/TableScan.h
@@ -51,6 +51,10 @@ class TableScan : public SourceOperator {
       column_index_t outputChannel,
       const std::shared_ptr<common::Filter>& filter) override;
 
+  int64_t splitWeight() {
+    return currentSplitWeight_;
+  }
+
   /// The name of runtime stats specific to table scan.
   /// The number of running table scan drivers.
   ///
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -473,6 +473,59 @@ const std::string& Task::getOrCreateSpillDirectory() {
   return spillDirectory_;
 }
 
+Task::SplitStatus Task::getSplitStatus() {
+  SplitStatus splitStatus;
+
+  // Iterate drivers to
+  // 1. Collect running splits
+  // 2. Collect runningTableScanGroups: a map of <tableScan nodeId, [group ids]>
+  std::unordered_map<std::string, std::unordered_set<uint32_t>>
+      runningTableScanGroups;
+  for (auto driver : drivers_) {
+    if (driver == nullptr || driver->operators().empty() ||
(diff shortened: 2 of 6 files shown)


velox/exec/TableScan.h
```diff
@@ -51,6 +51,10 @@ class TableScan : public SourceOperator {
       column_index_t outputChannel,
       const std::shared_ptr<common::Filter>& filter) override;
 
+  int64_t splitWeight() {
```
spershin, line 54: This method should be `const`.

velox/exec/Task.cpp
```diff
@@ -1422,6 +1475,7 @@ std::unique_ptr<ContinuePromise> Task::addSplitLocked(
   ++taskStats_.numTotalSplits;
   ++taskStats_.numQueuedSplits;
 
+  // Remove after presto switches to use the new API
```
spershin, line 1478: These comments would be hard to find later.
1. Do we need these comments? Would simply removing the old class members highlight these code sites with compilation errors, so no comment needed?
2. Could add some 'searchable key' to the comments, such as GH issue ID: `issueNNNNN`.

velox/exec/Task.cpp
```diff
@@ -473,6 +473,59 @@ const std::string& Task::getOrCreateSpillDirectory() {
...
 
+Task::SplitStatus Task::getSplitStatus() {
+  SplitStatus splitStatus;
+
+  // Iterate drivers to
+  // 1. Collect running splits
+  // 2. Collect runningTableScanGroups: a map of <tableScan nodeId, [group ids]>
+  std::unordered_map<std::string, std::unordered_set<uint32_t>>
+      runningTableScanGroups;
+  for (auto driver : drivers_) {
```
spershin, line 484: Most of the Task's members cannot be accessed w/o taking a lock first.
The Task's methods that access members are usually called after a lock is acquired and have `Locked` suffix.

spershin (commented):
@kewang1024 
Curious, what exactly this change is trying to fix?

PR #13177, opened 2025-04-28
feat(fuzzer): Add custom input generation for URL functions

peterenescu:
Summary: Add custom input generation for URL functions. In this diff, we add custom input for URLs by constructing a valid URL (using standard protocol and facebook hostname/domain). We then randomly add ports, fragements and queries to generate a valid random URL. Finally, we make small permutations on the input randomly to occasionally invalidate input and see how the results look.

Differential Revision: D73807620

### DIFF
--- a/velox/common/fuzzer/ConstrainedGenerators.cpp
+++ b/velox/common/fuzzer/ConstrainedGenerators.cpp
@@ -631,4 +631,216 @@ variant CastVarcharInputGenerator::generate() {
   return variant(input);
 }
 
+// URLInputGenerator creates URL input data for URL functions.
+URLInputGenerator::URLInputGenerator(
+    size_t seed,
+    const TypePtr& type,
+    double nullRatio,
+    std::string functionName,
+    std::vector<std::string> functionsToSkipForMailTo,
+    std::vector<std::string> functionsToSkipForTruncate)
+    : AbstractInputGenerator(seed, type, nullptr, nullRatio),
+      functionName_{std::move(functionName)},
+      functionsToSkipForMailTo_{std::move(functionsToSkipForMailTo)},
+      functionsToSkipForTruncate_{std::move(functionsToSkipForTruncate)} {}
+
+URLInputGenerator::~URLInputGenerator() = default;
+
+std::shared_ptr<RuleList> URLInputGenerator::generateURLRules() {
+  auto url = RuleList({
+      std::make_shared<ChoiceRule>(
+          rng_,
+          std::make_shared<RuleList>(std::vector<std::shared_ptr<Rule>>{
+              std::make_shared<ConstantRule>("http"),
+              std::make_shared<ConstantRule>("https"),
+          })),
+      std::make_shared<ConstantRule>("://"),
+      std::make_shared<WordRule>(rng_), // domain
+      std::make_shared<ConstantRule>("."),
+      std::make_shared<WordRule>(rng_, 2, 3, true),
+      std::make_shared<OptionalRule>( // port
+          rng_,
+          std::make_shared<RuleList>(std::vector<std::shared_ptr<Rule>>{
+              std::make_shared<ConstantRule>(":"),
+              std::make_shared<NumRule>(rng_, 3, 7, true)})),
+      std::make_shared<RepeatingRule>( // path
+          rng_,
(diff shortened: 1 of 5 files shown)


velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -480,6 +480,23 @@ class CastVarcharInputGenerator : public AbstractInputGenerator {
...
+
+  variant generate() override;
+
+ private:
+  const std::string kProtocol = "https";
+  const std::string kDomain = "www.facebook.com";
+  const std::string kPort = "8080";
+  const std::string kPath = "/test";
+  const std::string kQuery = "a=1&b=2";
+  const std::string kFragment = "test";
```
kagamiori, line 497: I wonder whether we should randomly generate these strings. E.g., could you check whether valid vs. invalid protocol string affects the behavior of the url_xxx functions in Presto/Velox? Same for the other fields.

kevinwilfong: Perhaps outside the scope of this diff, but something that would be really useful for generating URLs, JSON, JSON paths, regexes, etc. would be a tool that could take a grammar and generate random strings according to that grammar.

Short of that though, randomly generating strings (perhaps from fixed character sets) would offer better coverage.

velox/common/fuzzer/ConstrainedGenerators.cpp
```diff
@@ -20,6 +20,8 @@
 #include "velox/functions/lib/TDigest.h"
 #include "velox/functions/prestosql/types/BingTileType.h"
 
+#include <cfloat>
```
kevinwilfong: why do we need this?

velox/common/fuzzer/ConstrainedGenerators.cpp
```diff
@@ -631,4 +633,216 @@ variant CastVarcharInputGenerator::generate() {
...
+            std::find(
+                functionsToSkipForMailTo_.begin(),
+                functionsToSkipForMailTo_.end(),
+                functionName_) != functionsToSkipForMailTo_.end()
+                ? 0.0
+                : 0.1});
+  }
+
+  // Intentionally ignore these cases due to intentional differences, see
+  // T222820058
```
kevinwilfong: can you file a GitHub Issue instead, these task numbers are meaningless to open source folks

velox/common/fuzzer/ConstrainedGenerators.h
```diff
@@ -480,6 +480,32 @@ class CastVarcharInputGenerator : public AbstractInputGenerator {
...
+  variant generate() override;
+
+ private:
+  std::shared_ptr<RuleList> generateURLRules();
+  std::shared_ptr<RuleList> generateMailToRules();
+  std::shared_ptr<RuleList> generateChromeExtensionRules();
+
+  const std::string functionName_;
+  // Particular UDFs are known to have mismatches for mailto and trucated input.
+  // Let's skip those test cases for now. More info can be found in T222820058.
```
kevinwilfong: likewise

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -210,6 +211,15 @@ int main(int argc, char** argv) {
           {"cast", std::make_shared<CastVarcharAndJsonArgValuesGenerator>()},
           {"json_parse", std::make_shared<JsonParseArgValuesGenerator>()},
           {"json_extract", std::make_shared<JsonExtractArgValuesGenerator>()},
+          {"scale_tdigest",
+           std::make_shared<TDigestArgValuesGenerator>("scale_tdigest")},
```
kevinwilfong, line 220: why are you adding this one?

peterenescu: I think it was part of the rebasing, I did not add this, this is currently in mainline

velox/common/fuzzer/ConstrainedGenerators.cpp
peterenescu: I think it was necessary for some double::max things, let me try to take it out and see if it complains

velox/common/fuzzer/ConstrainedGenerators.cpp
peterenescu: good idea, adding

PR #13180, opened 2025-04-28
fix(json): Ensure invalid escaped unicode characters are ignored by simdjson

kgpai:
Summary: Currently Illformed and illegal escaped unicode characters cause simdjson to throw - this behavior is different from Presto Java , especially for escape characters like : \uD83E which form the first part of a surrogate pair.

Differential Revision: D73815909


amitkdutta (approved):
Thanks @kgpai

PR #13188, opened 2025-04-29
refactor: Use new recursive spill structure in hash join

tanjialiang:
Summary: Use SpillPartitionId to replace the number based spill id in hash join. This is the preparation change for mixed grouped execution mode hash join spill.

Differential Revision: D73820671


xiaoxmeng (approved):
@tanjialiang thanks and the code looks cleaner.

PR #13201, opened 2025-04-30
feat: Add client result value tracer and benchmark tool

xiaoxmeng:
Summary: Add client tracer to collect the lookup result value in ss client and persistent for offline replay to analyze decode overhead. This PR also integrate with ss connector and make it configurable through ss connector config, and provide a replay tool

Differential Revision: D73898602


tanjialiang (approved):
LGTM thanks

PR #13205, opened 2025-05-01
feat: Add mixed execution mode support for hash join

tanjialiang:
Summary: Add support for mixed grouped execution mode spill for hash join. Mixed grouped execution mode for a join means the probe side pipeline is grouped executed, and the build size pipeline is ungrouped executed.

Differential Revision: D73866216


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #13206, opened 2025-05-01
build: Update FBOS to v2025.04.28.00

czentgr:
This PR changes a few things due to changes in the new dependent libraries
- namespace issues
- coroutine support causing segv in remote functions test
- updated patch files
- new dependencies for FBOS (xxhash and fast float)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -74,6 +74,7 @@ jobs:
             run_and_time install_fmt
             run_and_time install_protobuf
             run_and_time install_boost
+            run_and_time install_fast_float
             run_and_time install_folly
             run_and_time install_stemmer
             run_and_time install_thrift
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -80,6 +80,13 @@ jobs:
             run_and_time install_thrift
             run_and_time install_arrow
 
+      # Folly is built with a new flag that the cache may not have. Ensure velox build and folly have
+      # used the correct flags (disabled the coroutines)
+      - name: TEMP: Force folly rebuild and update cache
+        run: |
+            source velox/scripts/setup-ubuntu.sh
+            run_and_time install_folly
+
       - name: Save Dependencies
         if: ${{ steps.restore-deps.outputs.cache-hit != 'true' }}
         uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
--- a/CMake/FindXxhash.cmake
+++ b/CMake/FindXxhash.cmake
@@ -1,0 +1,35 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# Try to find Facebook xxhash library This will define Xxhash_FOUND
+# Xxhash_INCLUDE_DIR Xxhash_LIBRARY
+#
+
+find_path(Xxhash_INCLUDE_DIR NAMES xxhash.h)
+
+find_library(Xxhash_LIBRARY_RELEASE NAMES xxhash)
+
+include(SelectLibraryConfigurations)
+select_library_configurations(Xxhash)
+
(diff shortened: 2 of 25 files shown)


kgpai:
@czentgr Can you post an update. 
cc: @majetideepak

CMake/resolve_dependency_modules/README.md
```diff
@@ -32,16 +32,17 @@ by Velox. See details on bundling below.
...
-| fizz              | v2024.07.01.00  | No       |
-| wangle            | v2024.07.01.00  | No       |
-| mvfst             | v2024.07.01.00  | No       |
-| fbthrift          | v2024.07.01.00  | No       |
+| folly             | v2025.03.31.00  | Yes      |
+| fizz              | v2025.03.31.00  | No       |
+| wangle            | v2025.03.31.00  | No       |
+| mvfst             | v2025.03.31.00  | No       |
+| fbthrift          | v2025.03.31.00  | No       |
 | libstemmer        | 2.2.0           | Yes      |
```
majetideepak, line 40: let's add thrift version as well.

majetideepak:
@czentgr and I investigated this further. The root cause does not seem to be thrift and fbthrift conflict since the test fails even with just fbthrift installed. 
The segfault is related to a virtual table thunk offset being incorrect for a destructor.

czentgr:
@majetideepak @kgpai @assignUser Finally, a new PR with the fix for the Remote Functions that also upgrades folly and passes all tests.
Currently, the Linux action that uses the dependency image were run with a new dependency image I made from the changes to the setups script for centos.

```
  /usr/bin/docker --config /home/runner/work/_temp/.docker_66f62f0f-b7f5-405c-9e3f-462c8b230bd0 login ghcr.io -u czentgr --password-stdin
  /usr/bin/docker --config /home/runner/work/_temp/.docker_66f62f0f-b7f5-405c-9e3f-462c8b230bd0 pull ghcr.io/czentgr/czentgr-test:adapters
```

The other change I had to make was to force a folly rebuild for the benchmarks because the there is a cache for the dependencies that are installed (including folly) that had an old version before we added the new flag and it caused a link error because folly has some checks to make sure that the library and user executable used the same flags in some cases it appears.

Anyway, I will remove the commit that changed the dependency image and the forced folly rebuild once we get a review done.

majetideepak (approved):
Thanks! @czentgr

majetideepak:
@assignUser can you take anothter look at the CMake changes?

velox/benchmarks/basic/ComparisonConjunct.cpp
```diff
@@ -176,7 +176,7 @@ BENCHMARK(conjunctsNested) {
 
 int main(int argc, char* argv[]) {
   folly::Init init{&argc, &argv};
-  gflags::ParseCommandLineFlags(&argc, &argv, true);
+  ::gflags::ParseCommandLineFlags(&argc, &argv, true);
```
assignUser, line 179: I am still unsure why this needed the explicit namespace to disambiguate, maybe because of all the macro stuff in folly? I was unable to repro on [godbolt](https://godbolt.org/z/zhc7s8bfa)

assignUser (approved):
Thanks for solving this!

P.S. We should probably document in the resolve README which dependencies are direct dependencies and which are just transitively used by folly etc.

assignUser:
Do we want to merge this first or #12321? In either case someone will have to solve conflicts.

CMake/FindXxhash.cmake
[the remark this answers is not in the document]
czentgr: Let me add a comment at the top.

czentgr:
> Do we want to merge this first or #12321? In either case someone will have to solve conflicts.

Yes. We will have to resolve the conflicts. Will we get the other PR merged faster than this one? This one is blocking arm builds. I'd say let this one go through and refactor the other PR as other people become unblocked. But then lets focus on the refactor PR to get the refactoring over the line. We have to double check the versions used. Turns out one hadoop version was incorrect which was caught by a test because the script was copied from my old PR where it had changed in the mean time.

I'm undoing the image change. Looks like we should keep the folly build to change the cache around for a bit longer.

.github/workflows/benchmark.yml
```diff
@@ -74,11 +74,19 @@ jobs:
...
             run_and_time install_boost
+            run_and_time install_fast_float
             run_and_time install_folly
             run_and_time install_stemmer
             run_and_time install_thrift
             run_and_time install_arrow
 
+      # Folly is built with a new flag that the cache may not have. Ensure velox build and folly have
+      # used the correct flags (disabled the coroutines)
+      - name: TEMP: Force folly rebuild and update cache
```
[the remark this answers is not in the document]
czentgr: Ok, I will remove the dependency image change only.

velox/benchmarks/basic/ComparisonConjunct.cpp
assignUser: > If using folly::gflags it would complain that the namespace was not found 

So weird. Well, easy enough fix for now.

PR #13209, opened 2025-05-01
fix: Virtual env should be created with Python version of user's system

### LINKED ISSUE #13208: Issue creating venv while running setup-ubuntu.sh
### Bug description

When setup-ubuntu script tries to create a virtual environment, if fails to create it because of the absence of ensurepip.

### System information

Velox System Info v0.0.2
Commit: 0d3d818fad7c4055abe136fa94639f018ffb1a07
CMake Version: 3.28.3
System: Linux-6.8.0-1027-aws
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.4.0

### Relevant logs

```bash
Creating Python Virtual Environment at ./scripts/../.venv
+ python3 -m venv ./scripts/../.venv
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.10-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.

Failing command: /velox/.venv/bin/python3
```

ajeyabsf:
This PR fixes issue of venv creation.

### Related issue

Fixes #13208

### DIFF
--- a/scripts/setup-ubuntu.sh
+++ b/scripts/setup-ubuntu.sh
@@ -44,6 +44,9 @@ USE_CLANG="${USE_CLANG:-false}"
 export INSTALL_PREFIX=${INSTALL_PREFIX:-"/usr/local"}
 DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)/deps-download}
 VERSION=$(cat /etc/os-release | grep VERSION_ID)
+PYTHON=$(command -v python3)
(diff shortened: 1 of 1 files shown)


ajeyabsf:
@assignUser @majetideepak your views on this change?

assignUser (commented):
I don't think this should be necessary. If you `sudo apt install python3 python3-venv` they will have the same version. If there is a mismatch I'd  say that's a local misconfiguration?

PR #13213, opened 2025-05-01
refactor(fuzzer): Unlink MemoryArbitrationFuzzer from GTest

xin-zhang2:

PR #13218, opened 2025-05-02
refactor: Separate TableEvolutionFuzzer from TableEvolutionFuzzerTest to allow for easier inheritance

peterenescu:
Separate TableEvolutionFuzzer from its corresponding test. This will allow for easier enhancements and subpackage inheritance. One planned such inheriting will be for Nimble support.

Also added file format name parser to fuzzer implements to reuse and add command-line arguments for

Differential Revision: D74042138


majetideepak:
We should unlink gtest from the fuzzer as well.
Similar to https://github.com/facebookincubator/velox/pull/13213

PR #13219, opened 2025-05-02
refactor: Remove unused error_source SYSTEM 

acvictor:
Remove unused error_source SYSTEM.


acvictor:
@Yuhta can you please review this PR? Thanks!

velox/common/base/VeloxException.h
```diff
@@ -50,10 +50,6 @@ inline constexpr auto kErrorSourceUser = "USER"_fs;
 /// in the system.
 inline constexpr auto kErrorSourceRuntime = "RUNTIME"_fs;
 
-/// Errors where the root cause of the problem is some unreliable aspect of the
```
pedroerp, line 53: I'm not sure it's worth it to remove this though. Sounds like a legit class of error that may be useful in the future. Since all these things leak through the API. it may be even the case this is used by some other project using Velox.

PR #13220, opened 2025-05-02
fix: Add alias substring for SubstrFunction

yabinma:
This is for fixing presto error 
Error in log presto.default.substring, called with arguments: (VARCHAR, BIGINT, BIGINT)

### DIFF
--- a/velox/functions/prestosql/registration/StringFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/StringFunctionsRegistration.cpp
@@ -61,16 +61,16 @@ void registerSimpleFunctions(const std::string& prefix) {
       {prefix + "trail"});
 
   registerFunction<SubstrFunction, Varchar, Varchar, int64_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
   registerFunction<SubstrFunction, Varchar, Varchar, int64_t, int64_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
 
   // TODO Presto doesn't allow INTEGER types for 2nd and 3rd arguments. Remove
   // these signatures.
   registerFunction<SubstrFunction, Varchar, Varchar, int32_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
   registerFunction<SubstrFunction, Varchar, Varchar, int32_t, int32_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
 
   registerFunction<SubstrVarbinaryFunction, Varbinary, Varbinary, int64_t>(
       {prefix + "substr"});
--- a/velox/functions/prestosql/tests/StringFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/StringFunctionsTest.cpp
@@ -91,13 +91,23 @@ class StringFunctionsTest : public FunctionBaseTest {
       int stringVectorIndex = 0) {
     auto row = makeRowVector(args);
     auto stringVector = args[stringVectorIndex];
-    int refCountBeforeEval =
-        bufferRefCounts(stringVector->asFlatVector<StringView>());
+    auto flatStringArg = stringVector->asFlatVector<StringView>();
+
+    int refCountBeforeEval = bufferRefCounts(flatStringArg);
     auto result = evaluate<FlatVector<StringView>>(query, row);
 
-    int refCountAfterEval =
-        bufferRefCounts(stringVector->asFlatVector<StringView>());
-    EXPECT_EQ(refCountAfterEval, 2 * refCountBeforeEval) << "at " << query;
+    int refCountAfterEval = bufferRefCounts(flatStringArg);
+
+    int numNonNullBuffersInStringArg = 0;
+    for (const auto& buf : flatStringArg->stringBuffers()) {
+      if (buf != nullptr) {
+        numNonNullBuffersInStringArg++;
+      }
+    }
+
+    EXPECT_EQ(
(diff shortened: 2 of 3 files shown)


velox/functions/prestosql/registration/StringFunctionsRegistration.cpp
```diff
@@ -61,16 +61,16 @@ void registerSimpleFunctions(const std::string& prefix) {
...
   // TODO Presto doesn't allow INTEGER types for 2nd and 3rd arguments. Remove
   // these signatures.
   registerFunction<SubstrFunction, Varchar, Varchar, int32_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
   registerFunction<SubstrFunction, Varchar, Varchar, int32_t, int32_t>(
-      {prefix + "substr"});
+      {prefix + "substr", prefix + "substring"});
 
   registerFunction<SubstrVarbinaryFunction, Varbinary, Varbinary, int64_t>(
```
aditi-pandit, line 75: Can alias these versions of substr as well.

aditi-pandit (commented):
@yabinma : Thanks for this PR.

One more change:
Add tests with using function name substring in https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/tests/StringFunctionsTest.cpp#L360 and https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/tests/StringFunctionsTest.cpp#L668

velox/functions/prestosql/tests/StringFunctionsTest.cpp
```diff
@@ -388,18 +400,31 @@ TEST_F(StringFunctionsTest, substrVariable) {
     // Checking the null results
     if (expectNullString(i) || expectNullStart(i) || expectNullLength(i)) {
       EXPECT_TRUE(result->isNullAt(i)) << "expected null at " << i;
+      EXPECT_TRUE(substringResult->isNullAt(i)) << "expected null at " << i;
```
aditi-pandit, line 403: You can create a lambda function say validateResult does the code in the body of the loop and call it with result and substringResult to avoid the code repetition.

velox/functions/prestosql/tests/StringFunctionsTest.cpp
```diff
@@ -97,7 +97,12 @@ class StringFunctionsTest : public FunctionBaseTest {
 
     int refCountAfterEval =
         bufferRefCounts(stringVector->asFlatVector<StringView>());
-    EXPECT_EQ(refCountAfterEval, 2 * refCountBeforeEval) << "at " << query;
+
+    if (query.find("substring(") != std::string::npos) {
```
aditi-pandit, line 101: Why is this behavior different ?

aditi-pandit (approved):
Thanks @yabinma

aditi-pandit:
Hi Wei, Krishna, Please an you help with the review and merge for this PR.

velox/functions/prestosql/tests/StringFunctionsTest.cpp
yabinma: done in new commit.

velox/functions/prestosql/tests/StringFunctionsTest.cpp
yabinma: We added the test case for substring alias, there are more calls compared with substr only.   I made the changes so that buffer reference count works for both of substr and substring.

velox/functions/prestosql/tests/StringFunctionsTest.cpp
```diff
@@ -91,13 +91,23 @@ class StringFunctionsTest : public FunctionBaseTest {
...
-        bufferRefCounts(stringVector->asFlatVector<StringView>());
+    auto flatStringArg = stringVector->asFlatVector<StringView>();
+
+    int refCountBeforeEval = bufferRefCounts(flatStringArg);
     auto result = evaluate<FlatVector<StringView>>(query, row);
 
-    int refCountAfterEval =
-        bufferRefCounts(stringVector->asFlatVector<StringView>());
-    EXPECT_EQ(refCountAfterEval, 2 * refCountBeforeEval) << "at " << query;
+    int refCountAfterEval = bufferRefCounts(flatStringArg);
```
aditi-pandit, line 99: @yabinma : Why did you change the EXPECT condition ? Is EXPECT_EQ(refCountAfterEval, 2 * refCountBeforeEval) not valid anymore ?

yabinma: @aditi-pandit ,  yes,  the previous EXPECT_EQ does not work anymore after adding test case for substring(evaluateSubstr is called twice after the new substring case added, buffer reference count is difference with twice calls).
https://github.com/facebookincubator/velox/pull/13220#discussion_r2077360039

kgpai (commented):
Should we add this to the docs ?

kgpai (approved):

kgpai:
@yabinma please update the docs

yabinma:
> @yabinma please update the docs

@kgpai , doc added.  please review.   All check passed .

aditi-pandit:
@yabinma : Please use the instructions at https://github.com/facebookincubator/velox/tree/main/velox/docs#velox-documentation to build the docs and check the result to ensure the formating is consistent.

yabinma:
> @yabinma : Please use the instructions at https://github.com/facebookincubator/velox/tree/main/velox/docs#velox-documentation to build the docs and check the result to ensure the formating is consistent.

OK. Here is what it looks like,
image

aditi-pandit:
@yabinma Looks good. Thanks !

PR #13221, opened 2025-05-02
misc: Remove simd::memcpy and simd::memset

andersdellien-arm:
These implementations are slower than std::memcpy in most cases, so can be removed.


majetideepak:
@Yuhta can you take look?

Yuhta:
@andersdellien-arm Can we keep the implementation and just replace the usages and remove the implementation after a while if we see no regression?  Also can you share the benchmark on `memset`?

PR #13232, opened 2025-05-05
fix: Remove duplicated assignements

juwentus1234:
Summary: Remove duplicated assignments in initialization list, also removed unused include headers

Differential Revision: D74187332


velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -83,7 +82,6 @@ PlanBuilder& PlanBuilder::tableScan(
       .subfieldFilters(subfieldFilters)
       .remainingFilter(remainingFilter)
       .dataColumns(dataColumns)
-      .assignments(assignments)
```
Yuhta, line 86: We should not remove this line

Yuhta (approved):

PR #13237, opened 2025-05-06
feat: Support any match function following binary-valued logical

ccat3z:
In spark, the behavior of `array_exists` (`any_match` in velox) can be changed by `spark.sql.legacy.followThreeValuedLogicInArrayExists`. For three-valued boolean logic, if the predicate returns any nulls and no true is obtained, then exists returns null instead of false. For example, exists(array(1, null, 3), x -> x % 2 == 0) is null. For binary-valued boolean logic, the result is false instead of null. This PR add a config `kSparkLegacyFollowThreeValuedLogicInArrayExists` to support this behavior.

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -342,6 +342,9 @@ class QueryConfig {
   static constexpr const char* kSparkLegacyStatisticalAggregate =
       "spark.legacy_statistical_aggregate";
 
+  static constexpr const char* kSparkLegacyFollowThreeValuedLogicInArrayExists =
+      "spark.follow_three_valued_logic_in_array_exists";
+
   /// The number of local parallel table writer operators per task.
   static constexpr const char* kTaskWriterCount = "task_writer_count";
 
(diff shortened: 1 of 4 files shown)


ccat3z:
@rui-mo Could you help review this pr?

velox/functions/prestosql/ArrayAndMapMatch.cpp
```diff
@@ -25,7 +25,11 @@ namespace facebook::velox::functions {
 namespace {
 
 // @tparam TContainer Either ArrayVector or MapVector.
-template <typename TContainer, bool initialValue, bool earlyReturn>
+template <
+    typename TContainer,
+    bool initialValue,
+    bool earlyReturn,
+    bool followNull = true>
```
rui-mo, line 32: Hi, we may need to move the implementation to the lib folder to support the Spark-compatible behavior, thanks.

velox/core/QueryConfig.h
```diff
@@ -342,6 +342,9 @@ class QueryConfig {
   static constexpr const char* kSparkLegacyStatisticalAggregate =
       "spark.legacy_statistical_aggregate";
 
+  static constexpr const char* kSparkLegacyFollowThreeValuedLogicInArrayExists =
+      "spark.follow_three_valued_logic_in_array_exists";
```
rui-mo, line 346: Would you please document this config in https://github.com/facebookincubator/velox/blob/main/velox/docs/configs.rst#spark-specific-configuration?

Issue #13061, opened 2025-04-18
Some AWS S3 requests should be retryable


czentgr:
### Description

An issue observed is if the S3 access over network is somewhat unreliable. Errors reading or writing to S3 can be observed that lead to query failure.
Instead of failing the query the request should be retried. Example, MakeRequest call in the S3 filesystem.

An observed error had the following stack:

```
[...]

PR #13239, opened 2025-05-06
fix(s3): Add retry logic for retryable errors

nmahadevuni:
Fixes https://github.com/facebookincubator/velox/issues/13061
This PR adds
1) Retry logic for client errors where the SDK returns the error with RETRYABLE set to true and any custom errors that we want to be retried.
2) Default values for `retry-mode` and `max-attempts` that will be used by the aws-sdk to retry in case of some failures.

### DIFF
--- a/velox/connectors/hive/storage_adapters/s3fs/S3Config.h
+++ b/velox/connectors/hive/storage_adapters/s3fs/S3Config.h
@@ -76,6 +76,7 @@ class S3Config {
     kRetryMode,
     kUseProxyFromEnv,
     kCredentialsProvider,
+    kMaxClientRetries,
     kEnd
   };
 
@@ -114,6 +115,8 @@ class S3Config {
              std::make_pair("use-proxy-from-env", "false")},
             {Keys::kCredentialsProvider,
              std::make_pair("aws-credentials-provider", std::nullopt)},
+            {Keys::kMaxClientRetries,
+             std::make_pair("max-client-retries", std::nullopt)},
         };
     return config;
   }
@@ -224,6 +227,15 @@ class S3Config {
   /// Retry mode for a single http client.
   std::optional<std::string> retryMode() const {
     return config_.find(Keys::kRetryMode)->second;
+  }
+
+  /// Maximum retry attempts for a single http client.
+  std::optional<uint32_t> maxClientRetries() const {
+    auto val = config_.find(Keys::kMaxClientRetries)->second;
+    if (val.has_value()) {
+      return folly::to<uint32_t>(val.value());
+    }
+    return std::optional<uint32_t>();
   }
 
   bool useProxyFromEnv() const {
--- a/velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
@@ -48,6 +48,20 @@
 #include <aws/s3/model/HeadObjectRequest.h>
 #include <aws/s3/model/UploadPartRequest.h>
 
+namespace {
+bool shouldRetryRequest(
+    Aws::Client::AWSError<Aws::S3::S3Errors>& error,
+    uint32_t retries,
+    uint32_t maxErrorRetries) {
+  if (retries < maxErrorRetries &&
+      (error.ShouldRetry() ||
+       error.GetErrorType() == Aws::S3::S3Errors::UNKNOWN)) {
+    return true;
+  }
+  return false;
+}
+} // namespace
+
 namespace facebook::velox::filesystems {
 namespace {
 // Reference: https://issues.apache.org/jira/browse/ARROW-8692
@@ -98,8 +112,11 @@ std::shared_ptr<Aws::Auth::AWSCredentialsProvider> getCredentialsProviderByName(
 
 class S3ReadFile final : public ReadFile {
  public:
-  S3ReadFile(std::string_view path, Aws::S3::S3Client* client)
-      : client_(client) {
+  S3ReadFile(
+      std::string_view path,
+      Aws::S3::S3Client* client,
+      uint32_t maxClientRetries)
+      : client_(client), maxClientRetries_(maxClientRetries) {
     getBucketAndKeyFromPath(path, bucket_, key_);
   }
 
@@ -216,11 +233,22 @@ class S3ReadFile final : public ReadFile {
         AwsWriteableStreamFactory(position, length));
     RECORD_METRIC_VALUE(kMetricS3ActiveConnections);
     RECORD_METRIC_VALUE(kMetricS3GetObjectCalls);
-    auto outcome = client_->GetObject(request);
-    if (!outcome.IsSuccess()) {
+
+    Aws::S3::Model::GetObjectOutcome outcome;
+    uint32_t retries = 0;
+    Aws::Client::AWSError<Aws::S3::S3Errors> getObjectError;
+    do {
+      outcome = client_->GetObject(request);
+      if (outcome.IsSuccess()) {
+        break;
+      }
+      getObjectError = outcome.GetError();
+      auto s3ClientRetryCount = outcome.GetRetryCount();
       RECORD_METRIC_VALUE(kMetricS3GetObjectErrors);
-    }
-    RECORD_METRIC_VALUE(kMetricS3GetObjectRetries, outcome.GetRetryCount());
+      retries += s3ClientRetryCount > 0 ? s3ClientRetryCount : 1;
+    } while (shouldRetryRequest(getObjectError, retries, maxClientRetries_));
+
+    RECORD_METRIC_VALUE(kMetricS3GetObjectRetries, retries);
     RECORD_METRIC_VALUE(kMetricS3ActiveConnections, -1);
     VELOX_CHECK_AWS_OUTCOME(outcome, "Failed to get S3 object", bucket_, key_);
   }
@@ -229,6 +257,7 @@ class S3ReadFile final : public ReadFile {
   std::string bucket_;
   std::string key_;
   int64_t length_ = -1;
+  uint32_t maxClientRetries_;
 };
 
 Aws::Utils::Logging::LogLevel inferS3LogLevel(std::string_view logLevel) {
@@ -690,6 +719,10 @@ class S3FileSystem::Impl {
 
     auto credentialsProvider = getCredentialsProvider(s3Config);
 
+    if (s3Config.maxClientRetries().has_value()) {
+      maxClientRetries_ = s3Config.maxClientRetries().value();
+    }
+
     client_ = std::make_shared<Aws::S3::S3Client>(
         credentialsProvider, nullptr /* endpointProvider */, clientConfig);
     ++fileSystemCount;
@@ -837,8 +870,13 @@ class S3FileSystem::Impl {
     return getAwsInstance()->getLogPrefix();
   }
 
+  uint32_t getMaxClientRetries() const {
+    return maxClientRetries_;
+  }
+
  private:
   std::shared_ptr<Aws::S3::S3Client> client_;
+  uint32_t maxClientRetries_;
 };
 
 S3FileSystem::S3FileSystem(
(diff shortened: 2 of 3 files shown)


velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -226,6 +229,15 @@ class S3Config {
     return config_.find(Keys::kRetryMode)->second;
   }
 
+  /// Maximum retry attempts for a single http client.
+  std::optional<int32_t> maxClientRetries() const {
+    auto val = config_.find(Keys::kMaxClientRetries)->second;
+    if (val.has_value()) {
+      return folly::to<int32_t>(val.value());
+    }
+    return std::optional<int32_t>(5);
```
czentgr, line 238: This should technically never come here because the default is defined with 5. So just like the others return `()` here.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -96,10 +96,25 @@ std::shared_ptr<Aws::Auth::AWSCredentialsProvider> getCredentialsProviderByName(
   });
 }
 
+bool shouldRetryRequest(
```
czentgr, line 99: This is not a member function and we don't need to expose it. Put it in an anonymous namespace.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -229,6 +257,7 @@ class S3ReadFile final : public ReadFile {
   std::string bucket_;
   std::string key_;
   int64_t length_ = -1;
+  int32_t maxClientRetries_;
```
czentgr, line 260: Why is this an int32_t? When later changed to a uint32_t to compare with a uint32_t retries in `shouldRetryRequest`. Lets use the same types throughout to avoid any type/cast issues.
Lets make the retries counter and maxRetries either use all signed or unsigned.

nmahadevuni:
Thanks @czentgr . I have addressed your comments. Please have a look.

czentgr (approved):
Thanks, looks good.

velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -108,12 +109,14 @@ class S3Config {
              std::make_pair("socket-timeout", std::nullopt)},
             {Keys::kMaxConnections,
              std::make_pair("max-connections", std::nullopt)},
-            {Keys::kMaxAttempts, std::make_pair("max-attempts", std::nullopt)},
-            {Keys::kRetryMode, std::make_pair("retry-mode", std::nullopt)},
+            {Keys::kMaxAttempts, std::make_pair("max-attempts", "10")},
```
majetideepak, line 112: Let's keep the library defaults since they probably suit the average case. These can be set in the deployment.

velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -108,12 +109,14 @@ class S3Config {
...
+            {Keys::kRetryMode, std::make_pair("retry-mode", "standard")},
             {Keys::kUseProxyFromEnv,
              std::make_pair("use-proxy-from-env", "false")},
             {Keys::kCredentialsProvider,
              std::make_pair("aws-credentials-provider", std::nullopt)},
+            {Keys::kMaxClientRetries,
+             std::make_pair("max-client-retries", "5")},
```
majetideepak, line 119: Let's leave the default to 0, which is the current behavior.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -48,6 +48,20 @@
 #include <aws/s3/model/HeadObjectRequest.h>
 #include <aws/s3/model/UploadPartRequest.h>
 
+namespace {
+bool shouldRetryRequest(
```
majetideepak, line 52: This can be a private function of S3ReadFile.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -216,11 +233,24 @@ class S3ReadFile final : public ReadFile {
...
+    while (shouldRetry) {
+      outcome = client_->GetObject(request);
+      if (outcome.IsSuccess()) {
+        break;
+      }
+      auto error = outcome.GetError();
+      auto s3ClientRetryCount = outcome.GetRetryCount();
       RECORD_METRIC_VALUE(kMetricS3GetObjectErrors);
+      retries += s3ClientRetryCount > 0 ? s3ClientRetryCount : 1;
+      shouldRetry = shouldRetryRequest(error, retries, maxClientRetries_);
```
majetideepak, line 250:
```
do {
  ....
} while(shouldRetryRequest(error, retryCount));
```

velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemMetricsTest.cpp
```diff
@@ -168,6 +168,34 @@ TEST_F(S3FileSystemMetricsTest, metrics) {
   EXPECT_EQ(1, s3Reporter->counterMap[std::string{kMetricS3GetObjectCalls}]);
 }
 
+// This test takes two to three minutes to run.
```
majetideepak, line 171: Where does the time go? We should aim to reduce this.

majetideepak: I see the delay is from the time out `Couldn't connect to server`. Can we reduce this timeout?

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -216,11 +233,24 @@ class S3ReadFile final : public ReadFile {
...
+
+    while (shouldRetry) {
+      outcome = client_->GetObject(request);
+      if (outcome.IsSuccess()) {
+        break;
+      }
+      auto error = outcome.GetError();
+      auto s3ClientRetryCount = outcome.GetRetryCount();
       RECORD_METRIC_VALUE(kMetricS3GetObjectErrors);
+      retries += s3ClientRetryCount > 0 ? s3ClientRetryCount : 1;
```
majetideepak, line 249: I am also lost here. If we use the standard retry strategy, with max attempts, does it not retry up to max attempts?
Where do we need `max-client-retries`?

majetideepak: The PR description mentions custom errors. Can you elaborate?

velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemMetricsTest.cpp
```diff
@@ -168,6 +168,34 @@ TEST_F(S3FileSystemMetricsTest, metrics) {
...
+      "Dance me through the panic till I'm gathered safely in"
+      "Lift me like an olive branch and be my homeward dove"
+      "Dance me to the end of love";
+  writeFile->append(kDataContent);
+  writeFile->close();
+  EXPECT_EQ(1, s3Reporter->counterMap[std::string{kMetricS3SuccessfulUploads}]);
+
+  auto readFile = s3fs.openFileForRead(s3File);
+  minioServer_->stop();
+  VELOX_ASSERT_THROW(
```
czentgr, line 194: This is a connect timeout. Usually, these are not retried by re-attempting to connect but instead instead increasing the connect timeout. 
I discussed with Deepak and agree that we need to understand the different retry types that we can pass to the S3 client first a bit.
There are many reasons why a request failed. Both at network and application level.

velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemMetricsTest.cpp
nmahadevuni: I have tried reducing socket and connect timeout values, but still the runtime is the same.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
nmahadevuni: Aws sdk uses default retry strategy with 10 attempts but for some errors, the retry count is 0 and is marked RETRYABLE in the error returned. An example is below.
```
error = {Aws::S3::S3Error} 
 Aws::Client::AWSError<Aws::S3::S3Errors> = {Aws::Client::AWSError<Aws::S3::S3Errors>} 
  m_errorType = {Aws::S3::S3Errors} NETWORK_CONNECTION
  m_exceptionName = {Aws::String} ""
  m_message = {Aws::String} "curlCode: 7, Couldn't connect to server"
  m_remoteHostIpAddress = {Aws::String} ""
  m_requestId = {Aws::String} ""
  m_responseHeaders = {Aws::Http::HeaderValueCollection} size=0
  m_responseCode = {Aws::Http::HttpResponseCode} REQUEST_NOT_MADE
  m_errorPayloadType = {Aws::Client::ErrorPayloadType} NOT_SET
  m_xmlPayload = {Aws::Utils::Xml::XmlDocument} 
  m_jsonPayload = {Aws::Utils::Json::JsonValue} 
  m_retryableType = {Aws::Client::RetryableType} RETRYABLE
success = {bool} false
retryCount = {unsigned int} 0
```

Custom errors are just errors we want to retry irrespective of what SDK says. For example, in the issue Christian reported, the error is UNKNOWN and its not a retryable error. I have added that as a retryable error.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
nmahadevuni: I want to keep it outside so we may use it for other calls too.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
majetideepak: As Christian mentioned below, network errors are not retried.

nmahadevuni:
Reopen

nmahadevuni:
@czentgr @JkSelf @yma11 Can you please review this?

velox/connectors/hive/storage_adapters/s3fs/S3Config.h
```diff
@@ -226,6 +229,15 @@ class S3Config {
     return config_.find(Keys::kRetryMode)->second;
   }
 
+  /// Maximum retry attempts for a single http client.
+  std::optional<uint32_t> maxClientRetries() const {
+    auto val = config_.find(Keys::kMaxClientRetries)->second;
+    if (val.has_value()) {
```
JkSelf, line 235: Since `kMaxClientRetries ` already has a default value, we can directly return folly::to<uint32_t>(val.value())  without this check.

velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.cpp
```diff
@@ -294,6 +294,10 @@ class S3FileSystem::Impl {
 
     auto credentialsProvider = getCredentialsProvider(s3Config);
 
+    if (s3Config.maxClientRetries().has_value()) {
```
JkSelf, line 297: ditto. No need this check.

velox/connectors/hive/storage_adapters/s3fs/S3ReadFile.cpp
```diff
@@ -24,6 +24,30 @@
...
+      return true;
+    default:
+      return false;
+  }
+}
+bool shouldRetryRequest(
+    Aws::Client::AWSError<Aws::S3::S3Errors>& error,
+    uint32_t retries,
+    uint32_t maxErrorRetries) {
+  if (retries < maxErrorRetries && !error.ShouldRetry() &&
```
JkSelf, line 41: Why the condition is `!error.ShouldRetry()`?

JkSelf (commented):
Thanks for your work. Just some nits.

alexr17:
We are also hitting `Code:-1, S3 Service:'AmazonS3', Message:'Failed to flush response stream (eof: 0, bad: 1)'`, perhaps due to s3 throttling.

@nmahadevuni Thanks for your work! Hope this lands soon :)

PR #13245, opened 2025-05-06
fix: Casting complex types is only supposed to cast recursively if a child's type will change

kevinwilfong:
Summary:
There is logic in CastExpr so that when we cast a row or map to a different row or map type, 
we only apply the cast to the children recursively if their types are different in the new type.

However, this comparison is done by comparing shared_ptrs (TypePtrs) which compares the
pointer addresses.  This only works for primitive types that have singleton Types. For complex
types this needs to compare the Type objects themselves.

This came up in fuzzer testing when casting a Row that contained an Array(Timestamp) field
that only changed the names of the fields not the types of any of the fields. The array types
were the same but had different addresses.  Array correctly assumes that if it's being casted
to another array type it's children must be different (otherwise the type would be the same
and cast should never have been called). So we attempted to cast a Timestamp to a
Timestamp which is not supported.

Differential Revision: D74260775

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -382,7 +382,7 @@ VectorPtr CastExpr::applyMap(
 
   // Cast keys
   VectorPtr newMapKeys;
-  if (fromType.keyType() == toType.keyType()) {
+  if (*fromType.keyType() == *toType.keyType()) {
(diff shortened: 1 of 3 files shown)


kKPulla (approved):
Good catch. Thanks for fixing!

PR #11554, opened 2024-11-15
(Do not review) feat: TopNRank optimization

aditi-pandit:
Design doc : https://docs.google.com/document/d/1WQfNigR9bVrbM-PqY7F0mswcetN_tdNahzD9ENye-Q0/edit?usp=sharing

https://github.com/facebookincubator/velox/discussions/9404

e2e Presto PR (with changes in the Presto optimizer as well) https://github.com/prestodb/presto/pull/24138

Latency for SF1K TPC-DS Q67 fell from 399s to 146s with this change.

(I also started working on a fuzzer in https://github.com/facebookincubator/velox/pull/12103 which I will enhance for the rank and dense_rank functions added here).

PR #13248, opened 2025-05-06
feat(TopNRank): Add rank function field to TopNRowNumber node

aditi-pandit:
This is the first step to support TopNRowNumber node to support rank() and dense_rank() function.

ref https://github.com/facebookincubator/velox/pull/11554

### DIFF
--- a/velox/core/PlanNode.cpp
+++ b/velox/core/PlanNode.cpp
@@ -2228,8 +2228,41 @@ PlanNodePtr RowNumberNode::create(const folly::dynamic& obj, void* context) {
       source);
 }
 
+namespace {
+std::unordered_map<TopNRowNumberNode::RankFunction, std::string>
+rankFunctionNames() {
+  return {
+      {TopNRowNumberNode::RankFunction::kRowNumber, "row_number"},
+      {TopNRowNumberNode::RankFunction::kRank, "rank"},
+      {TopNRowNumberNode::RankFunction::kDenseRank, "dense_rank"},
+  };
+}
+} // namespace
+
+// static
+const char* TopNRowNumberNode::rankFunctionName(
+    TopNRowNumberNode::RankFunction function) {
+  static const auto kFunctionNames = rankFunctionNames();
+  auto it = kFunctionNames.find(function);
+  VELOX_CHECK(
+      it != kFunctionNames.end(),
+      "Invalid rank function {}",
+      static_cast<int>(function));
+  return it->second.c_str();
+}
+
+// static
+TopNRowNumberNode::RankFunction TopNRowNumberNode::rankFunctionFromName(
+    std::string_view name) {
+  static const auto kFunctionNames = invertMap(rankFunctionNames());
+  auto it = kFunctionNames.find(name.data());
+  VELOX_CHECK(it != kFunctionNames.end(), "Invalid rank function {}", name);
+  return it->second;
+}
+
 TopNRowNumberNode::TopNRowNumberNode(
     PlanNodeId id,
+    RankFunction function,
     std::vector<FieldAccessTypedExprPtr> partitionKeys,
     std::vector<FieldAccessTypedExprPtr> sortingKeys,
     std::vector<SortOrder> sortingOrders,
@@ -2237,6 +2270,7 @@ TopNRowNumberNode::TopNRowNumberNode(
     int32_t limit,
     PlanNodePtr source)
     : PlanNode(std::move(id)),
+      function_(function),
       partitionKeys_{std::move(partitionKeys)},
       sortingKeys_{std::move(sortingKeys)},
       sortingOrders_{std::move(sortingOrders)},
@@ -2274,6 +2308,8 @@ TopNRowNumberNode::TopNRowNumberNode(
 }
 
 void TopNRowNumberNode::addDetails(std::stringstream& stream) const {
+  stream << rankFunctionName(function_) << " ";
+
   if (!partitionKeys_.empty()) {
     stream << "partition by (";
     addFields(stream, partitionKeys_);
@@ -2289,6 +2325,7 @@ void TopNRowNumberNode::addDetails(std::stringstream& stream) const {
 
 folly::dynamic TopNRowNumberNode::serialize() const {
   auto obj = PlanNode::serialize();
+  obj["function"] = rankFunctionName(function_);
   obj["partitionKeys"] = ISerializable::serialize(partitionKeys_);
   obj["sortingKeys"] = ISerializable::serialize(sortingKeys_);
   obj["sortingOrders"] = serializeSortingOrders(sortingOrders_);
@@ -2310,6 +2347,7 @@ PlanNodePtr TopNRowNumberNode::create(
     const folly::dynamic& obj,
     void* context) {
   auto source = deserializeSingleSource(obj, context);
+  auto function = rankFunctionFromName(obj["function"].asString());
   auto partitionKeys = deserializeFields(obj["partitionKeys"], context);
   auto sortingKeys = deserializeFields(obj["sortingKeys"], context);
 
@@ -2322,6 +2360,7 @@ PlanNodePtr TopNRowNumberNode::create(
 
   return std::make_shared<TopNRowNumberNode>(
       deserializePlanNodeId(obj),
+      function,
       partitionKeys,
       sortingKeys,
       sortingOrders,
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -4781,19 +4781,55 @@ class MarkDistinctNode : public PlanNode {
 
 using MarkDistinctNodePtr = std::shared_ptr<const MarkDistinctNode>;
 
-/// Optimized version of a WindowNode for a single row_number function with a
-/// limit over sorted partitions.
-/// The output of this node contains all input columns followed by an optional
+/// Optimized version of a WindowNode for a single row_number, rank or
+/// dense_rank function with a limit over sorted partitions. The output of this
+/// node contains all input columns followed by an optional
 /// 'rowNumberColumnName' BIGINT column.
+/// TODO: This node will be renamed to TopNRank or TopNRowNode once all the
+/// support for handling rank and dense_rank is committed to Velox.
 class TopNRowNumberNode : public PlanNode {
  public:
+  enum class RankFunction {
+    kRowNumber,
+    kRank,
+    kDenseRank,
+  };
+
+  static const char* rankFunctionName(TopNRowNumberNode::RankFunction function);
+
+  static RankFunction rankFunctionFromName(std::string_view name);
+
+  /// @param rankFunction RanksFunction (row_number, rank, dense_rank) for TopN.
   /// @param partitionKeys Partitioning keys. May be empty.
   /// @param sortingKeys Sorting keys. May not be empty and may not intersect
   /// with 'partitionKeys'.
   /// @param sortingOrders Sorting orders, one per sorting key.
   /// @param rowNumberColumnName Optional name of the column containing row
-  /// numbers. If not specified, the output doesn't include 'row number'
-  /// column. This is used when computing partial results.
+  /// numbers or rank or dense_rank. If not specified, the output doesn't
+  /// include 'row_number' column. This is used when computing partial results.
+  /// @param limit Per-partition limit. The number of
+  /// rows produced by this node will not exceed this value for any given
+  /// partition. Extra rows will be dropped.
+  TopNRowNumberNode(
+      PlanNodeId id,
+      RankFunction function,
+      std::vector<FieldAccessTypedExprPtr> partitionKeys,
+      std::vector<FieldAccessTypedExprPtr> sortingKeys,
+      std::vector<SortOrder> sortingOrders,
+      const std::optional<std::string>& rowNumberColumnName,
+      int32_t limit,
+      PlanNodePtr source);
+
+#ifdef VELOX_BACKWARD_COMPATIBILITY
+  /// Note : This constructor is for backwards compatibility. Remove it after
+  /// migrating Prestissimo to use the new constructor.
+  /// @param partitionKeys Partitioning keys. May be empty.
+  /// @param sortingKeys Sorting keys. May not be empty and may not intersect
+  /// with 'partitionKeys'.
+  /// @param sortingOrders Sorting orders, one per sorting key.
+  /// @param rowNumberColumnName Optional name of the column containing row
+  /// numbers (or rank and dense_rank). If not specified, the output doesn't
+  /// include 'row number' column. This is used when computing partial results.
   /// @param limit Per-partition limit. The number of
   /// rows produced by this node will not exceed this value for any given
   /// partition. Extra rows will be dropped.
@@ -4804,7 +4840,17 @@ class TopNRowNumberNode : public PlanNode {
       std::vector<SortOrder> sortingOrders,
       const std::optional<std::string>& rowNumberColumnName,
       int32_t limit,
-      PlanNodePtr source);
+      PlanNodePtr source)
+      : TopNRowNumberNode(
+            id,
+            RankFunction::kRowNumber,
+            partitionKeys,
+            sortingKeys,
+            sortingOrders,
+            rowNumberColumnName,
+            limit,
+            source) {}
+#endif
 
   class Builder {
    public:
@@ -4821,10 +4867,16 @@ class TopNRowNumberNode : public PlanNode {
       limit_ = other.limit();
       VELOX_CHECK_EQ(other.sources().size(), 1);
       source_ = other.sources()[0];
+      function_ = other.rankFunction();
     }
 
     Builder& id(PlanNodeId id) {
       id_ = std::move(id);
+      return *this;
+    }
+
+    Builder& function(RankFunction function) {
+      function_ = function;
       return *this;
     }
 
@@ -4879,6 +4931,7 @@ class TopNRowNumberNode : public PlanNode {
 
       return std::make_shared<TopNRowNumberNode>(
           id_.value(),
+          function_.has_value() ? function_.value() : RankFunction::kRowNumber,
           partitionKeys_.value(),
           sortingKeys_.value(),
           sortingOrders_.value(),
@@ -4889,6 +4942,7 @@ class TopNRowNumberNode : public PlanNode {
 
    private:
     std::optional<PlanNodeId> id_;
+    std::optional<RankFunction> function_;
     std::optional<std::vector<FieldAccessTypedExprPtr>> partitionKeys_;
     std::optional<std::vector<FieldAccessTypedExprPtr>> sortingKeys_;
     std::optional<std::vector<SortOrder>> sortingOrders_;
@@ -4932,6 +4986,10 @@ class TopNRowNumberNode : public PlanNode {
     return limit_;
   }
 
+  RankFunction rankFunction() const {
+    return function_;
+  }
+
   bool generateRowNumber() const {
     return outputType_->size() > sources_[0]->outputType()->size();
   }
@@ -4946,6 +5004,8 @@ class TopNRowNumberNode : public PlanNode {
 
  private:
   void addDetails(std::stringstream& stream) const override;
+
+  const RankFunction function_;
 
   const std::vector<FieldAccessTypedExprPtr> partitionKeys_;
 
@@ -5096,5 +5156,16 @@ template <>
 struct fmt::formatter<facebook::velox::core::JoinType> : formatter<int> {
   auto format(facebook::velox::core::JoinType s, format_context& ctx) const {
     return formatter<int>::format(static_cast<int>(s), ctx);
+  }
+};
+
+template <>
+struct fmt::formatter<facebook::velox::core::TopNRowNumberNode::RankFunction>
+    : formatter<std::string> {
+  auto format(
+      facebook::velox::core::TopNRowNumberNode::RankFunction f,
+      format_context& ctx) const {
+    return formatter<std::string>::format(
+        facebook::velox::core::TopNRowNumberNode::rankFunctionName(f), ctx);
   }
 };
--- a/velox/exec/TopNRowNumber.cpp
+++ b/velox/exec/TopNRowNumber.cpp
@@ -140,6 +140,9 @@ TopNRowNumber::TopNRowNumber(
           node->sortingOrders(),
           data_.get()),
       decodedVectors_(inputType_->size()) {
+  VELOX_CHECK_EQ(
+      node->rankFunction(), core::TopNRowNumberNode::RankFunction::kRowNumber);
+
   const auto& keys = node->partitionKeys();
   const auto numKeys = keys.size();
 
--- a/velox/exec/tests/PlanNodeSerdeTest.cpp
+++ b/velox/exec/tests/PlanNodeSerdeTest.cpp
@@ -89,6 +89,26 @@ class PlanNodeSerdeTest : public testing::Test,
     }
   }
 
+  void topNRankSerdeTest(std::string_view function) {
+    auto plan = PlanBuilder()
+                    .values({data_})
+                    .topNRank(function, {}, {"c0", "c2"}, 10, false)
+                    .planNode();
+    testSerde(plan);
+
+    plan = PlanBuilder()
+               .values({data_})
+               .topNRank(function, {}, {"c0", "c2"}, 10, true)
+               .planNode();
+    testSerde(plan);
+
+    plan = PlanBuilder()
+               .values({data_})
+               .topNRank(function, {"c0"}, {"c1", "c2"}, 10, false)
+               .planNode();
+    testSerde(plan);
+  }
+
   std::vector<RowVectorPtr> data_;
 };
 
@@ -628,23 +648,15 @@ TEST_F(PlanNodeSerdeTest, scan) {
 }
 
 TEST_F(PlanNodeSerdeTest, topNRowNumber) {
-  auto plan = PlanBuilder()
-                  .values({data_})
-                  .topNRowNumber({}, {"c0", "c2"}, 10, false)
-                  .planNode();
-  testSerde(plan);
+  topNRankSerdeTest("row_number");
+}
 
-  plan = PlanBuilder()
-             .values({data_})
-             .topNRowNumber({}, {"c0", "c2"}, 10, true)
-             .planNode();
-  testSerde(plan);
+TEST_F(PlanNodeSerdeTest, topNRank) {
+  topNRankSerdeTest("rank");
+}
 
-  plan = PlanBuilder()
-             .values({data_})
-             .topNRowNumber({"c0"}, {"c1", "c2"}, 10, false)
-             .planNode();
-  testSerde(plan);
+TEST_F(PlanNodeSerdeTest, topNDemseRank) {
+  topNRankSerdeTest("dense_rank");
 }
 
 TEST_F(PlanNodeSerdeTest, write) {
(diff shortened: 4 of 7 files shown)


velox/core/PlanNode.h
```diff
@@ -4683,31 +4683,73 @@ class MarkDistinctNode : public PlanNode {
...
-/// Optimized version of a WindowNode for a single row_number function with a
-/// limit over sorted partitions.
-/// The output of this node contains all input columns followed by an optional
+/// Optimized version of a WindowNode for a single row_number, rank or
+/// dense_rank function with a limit over sorted partitions. The output of this
+/// node contains all input columns followed by an optional
 /// 'rowNumberColumnName' BIGINT column.
 class TopNRowNumberNode : public PlanNode {
  public:
+  enum class RankFunction {
```
JkSelf, line 4692: @aditi-pandit The `TopNRowNumberNode ` was previously used to handle the `row_number()` function. We now support the `row_number()`, `rank()`, and `dense_rank()` functions. Therefore, can we rename `TopNRowNumberNode `to `TopNRankLikeNode`? Additionally, should we consider changing `RankFunction `to `RankLikeFunction`?

aditi-pandit: @JkSelf : I do want to change the PlanNode and Operator both to use "TopNRank" in naming. But felt I should do the rename in one go after all the code changes are submit. wdyt ?

JkSelf: @aditi-pandit I'm fine with this. Could we add a TODO note for these changes? Thanks.

JkSelf (approved):
LGTM. Thanks for your work.

JkSelf:
@xiaoxmeng  @majetideepak  
After upgrading the Spark version to 3.5 in Gluten, we encountered a fallback issue with the `TopNRowNumber ` operator due to the absence of support for `rank `and `dense_rank `functions. We are eagerly need this support in Gluten. If you have some time available, could you please help to review this PR? Thanks.

velox/core/PlanNode.h
```diff
@@ -4683,31 +4683,75 @@ class MarkDistinctNode : public PlanNode {
...
+/// handling rank and dense_rank is committed to Velox.
 class TopNRowNumberNode : public PlanNode {
  public:
+  enum class RankFunction {
+    kRowNumber,
+    kRank,
+    kDenseRank,
+  };
+
+  static const char* rankFunctionName(TopNRowNumberNode::RankFunction function);
```
czentgr, line 4700: Will these be used later?  Or just convenience functions that are externalized as API to be used by anyone using this plan node?

aditi-pandit: @czentgr : These are used in TopNRowNumber::addDetails and TopNRowNumber::create(..) in TopNRowNumber.cpp

czentgr: Yes, just wondering why they are public if they are only used within the class itself.

velox/exec/tests/PlanNodeSerdeTest.cpp
```diff
@@ -571,24 +571,35 @@ TEST_F(PlanNodeSerdeTest, scan) {
   testSerde(plan);
 }
 
-TEST_F(PlanNodeSerdeTest, topNRowNumber) {
-  auto plan = PlanBuilder()
-                  .values({data_})
-                  .topNRowNumber({}, {"c0", "c2"}, 10, false)
-                  .planNode();
+#define TOPN_SERDE_TEST(funcname)                        \
```
czentgr, line 574: Just two nits about the code generation macros.
- we should wrap them in do-while(false)
- undefine it after usage in the tests.

czentgr (commented):
Just nits.

velox/core/PlanNode.h
aditi-pandit: Got your question now... I just followed the practise of other PlanNodes in this file.

aditi-pandit:
@czentgr : Have addressed the review comments. PTAL.

czentgr (approved):
Thanks!

velox/exec/tests/PlanNodeSerdeTest.cpp
```diff
@@ -571,26 +571,41 @@ TEST_F(PlanNodeSerdeTest, scan) {
   testSerde(plan);
 }
 
+#define TOPN_SERDE_TEST(funcname)                          \
```
jinchengchenghh: Please use _nodeName as argument as VELOX_ASSERT_THROW

velox/core/PlanNode.h
```diff
@@ -4811,6 +4860,7 @@ class TopNRowNumberNode : public PlanNode {
 
       return std::make_shared<TopNRowNumberNode>(
           id_.value(),
+          function_.has_value() ? function_.value() : RankFunction::kRowNumber,
```
jinchengchenghh, line 4934: Please use function_.value(), after the new change, this field is always required.

velox/exec/tests/PlanNodeToStringTest.cpp
```diff
@@ -927,39 +927,60 @@ TEST_F(PlanNodeToStringTest, rowNumber) {
...
-  auto plan = PlanBuilder()
-                  .tableScan(rowType)
-                  .topNRowNumber({}, {"a DESC"}, 10, false)
-                  .planNode();
-
-  ASSERT_EQ("-- TopNRowNumber[1]\n", plan->toString());
-  ASSERT_EQ(
-      "-- TopNRowNumber[1][order by (a DESC NULLS LAST) limit 10] -> a:BIGINT, b:VARCHAR\n",
-      plan->toString(true, false));
+#define TOPN_PLANNODE_TO_STRING_TEST(nodename, functionname)                                                             \
```
jinchengchenghh: ditto

velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -1420,6 +1436,13 @@ class PlanBuilder {
       const std::vector<std::string>& windowFunctions,
       bool inputSorted);
 
+  PlanBuilder& topNRowNumberBase(
```
jinchengchenghh: Move this one to private scope, and I'm also not sure if we should have 3 functions to interpret the nodes.

velox/core/PlanNode.h
```diff
@@ -4713,31 +4713,75 @@ class MarkDistinctNode : public PlanNode {
...
-/// Optimized version of a WindowNode for a single row_number function with a
-/// limit over sorted partitions.
-/// The output of this node contains all input columns followed by an optional
+/// Optimized version of a WindowNode for a single row_number, rank or
+/// dense_rank function with a limit over sorted partitions. The output of this
+/// node contains all input columns followed by an optional
 /// 'rowNumberColumnName' BIGINT column.
+/// TODO: This node will be renamed to TopNRank once all the support for
+/// handling rank and dense_rank is committed to Velox.
 class TopNRowNumberNode : public PlanNode {
```
jinchengchenghh, line 4790: Do we need to have a TopNWindowNode as base class and extends it by TopNRowNumberNode, TopNRankNode, xxx? Not sure for that. CC @rui-mo

velox/exec/TopNRowNumber.cpp
```diff
@@ -140,6 +140,10 @@ TopNRowNumber::TopNRowNumber(
           node->sortingOrders(),
           data_.get()),
       decodedVectors_(inputType_->size()) {
+  VELOX_CHECK(
```
rui-mo: nit: VELOX_CHECK_EQ

velox/core/PlanNode.h
rui-mo: I think it would be good if we could leverage much of the TopNRank node to handle several functions. Also, the row_number, rank, and dense_rank appear to be more function-level than operator-level distinctions.

@rui-mo : See answer above.

rui-mo (commented):
Thanks.

velox/core/PlanNode.h
jinchengchenghh: Do you have suggest for the node name? And in that case, do we need 3 functions for PlanBuilder? https://github.com/facebookincubator/velox/pull/13248/files#r2134892580

velox/exec/TopNRowNumber.cpp
aditi-pandit: There are compilation errors due to that on account of the enum usage:
`error occurred here in instantiation of function template specialization 'fmt::detail::make_arg<true, fmt::basic_format_context<fmt::appender, char>, const facebook::velox::core::TopNRowNumberNode::RankFunction, 0>' requested here in instantiation of function template specialization 'fmt::format_arg_store<fmt::basic_format_context<fmt::appender, char>, facebook::velox::core::TopNRowNumberNode::RankFunction, facebook::velox::core::TopNRowNumberNode::RankFunction>:... in instantiation of function template specialization 'fmt::make_format_args<fmt::basic_format_context<fmt::appender, char>, const facebook::velox::core::TopNRowNumberNode::RankFunction, const facebook::velox::core::TopNRowNumberNode::Ra... in instantiation of function template specialization 'facebook::velox::detail::errorMessage<facebook::velox::core::TopNRowNumberNode::RankFunction, facebook::velox::core::TopNRowNumberNode::RankFunction>' requested here template is declared here`

velox/core/PlanNode.h
aditi-pandit: @jinchengchenghh @rui-mo There is only one PlanNode TopNRowNumberNode (and one operator TopNRowNumber) with function field for row_number, rank, dense_rank.

The 3 methods are in PlanBuilder which is a test utility to make it simple for users to write plans. I tried to follow same style as window and streamingWindow methods which use the same underlying WindowNode but with different options.

Am I misunderstanding something here ?

jinchengchenghh: Your comment says `TODO: This node will be renamed to TopNRank once all the support for`, so I try to find a reasonable new name.
if we have only one node, I would prefer PlanBuilder has 1 function.

aditi-pandit: @jinchengchenghh : I intend to rename TopNRowNumberNode to TopNRank in the future as that is more representative of the functionality.

So we generally don't want the user to type internal PlanNode enum values (as would be needed for the function here) if possible. So one option is to take a string "row_number", "rank" or "dense_rank"" function name in the single PlanBuilder method. In order to avoid parsing/matching especially for errors, and since they were only 3 values I went with current solution : to have 3 PlanBuilder functions (especially since we had an example of window/streamingWindow) .... Don't think we have a rule that PlanBuilder functions should be be 1-1 with PlanNodes.

So that's my reasoning of the current design. wdyt ?

velox/exec/tests/utils/PlanBuilder.h
aditi-pandit: This is already in private space.

velox/core/PlanNode.h
jinchengchenghh: I don't have strong opinion for it. One node for one function in PlanBuilder. Maybe the name TopNRowNode is more reasonable.

aditi-pandit: @jinchengchenghh : TopNRowNode is a reasonable suggestion. I intend to change the node name as a single change post the main code changes for this PR.

aditi-pandit:
@xiaoxmeng @kagamiori : This is the first change for supporting rank, dense_rank based topn filtering in TopNRowNumber operator, and has got internal reviews already. Please can you help with the review and merge.

velox/core/PlanNode.h
jinchengchenghh: Please update the comment

> /// TODO: This node will be renamed to TopNRank once

aditi-pandit: @jinchengchenghh : Trino uses the name TopNRankingNode for this PlanNode. Presto uses TopNRowNumberNode (but it doesn't have rank, dense_rank). Does Spark have an equivalent node and what is it called there ?

I feel we should choose a name consistent with the engines.

JkSelf: Spark use WindowGroupLimitExec name https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/window/WindowGroupLimitExec.scala#L44

aditi-pandit: Ok.. lets stick with TopNRankNode or TopNRowNode in the future.

velox/core/PlanNode.h
jinchengchenghh: Why mark it as resolved?

jinchengchenghh: We should use the same format as `partitionKeys`, it is required, so not check if it hasValue. If have check, we should use VELOX_CHECK(function_.has_value()) before access the value

aditi-pandit: Sorry @jinchengchenghh... happened by accident.

My thinking was that since the class is called TopNRowNumberNode::Builder()... to make concession that if the user didn't set function(), then by default we use RowNumber. So wrote the code that way.

wdyt ?

jinchengchenghh: So you need to update here after you rename the node to `TopNRowNode` or `TopNRankNode`.

jinchengchenghh (approved):

velox/core/PlanNode.h
aditi-pandit: Yes

aditi-pandit:
@xiaoxmeng @kagamiori : This PR is towards add TopNRank optimization ref https://github.com/facebookincubator/velox/pull/11554. Multiple maintainers have reviewed this PR. Please can you help with review and merge.

velox/core/PlanNode.h
```diff
@@ -4790,31 +4790,75 @@ class MarkDistinctNode : public PlanNode {
...
+
+  /// @param rankFunction RanksFunction (row_number, rank, dense_rank) for TopN.
   /// @param partitionKeys Partitioning keys. May be empty.
   /// @param sortingKeys Sorting keys. May not be empty and may not intersect
   /// with 'partitionKeys'.
   /// @param sortingOrders Sorting orders, one per sorting key.
   /// @param rowNumberColumnName Optional name of the column containing row
-  /// numbers. If not specified, the output doesn't include 'row number'
-  /// column. This is used when computing partial results.
+  /// numbers (or rank and dense_rank). If not specified, the output doesn't
```
majetideepak: nit: row_number or rank or dense_rank.

velox/core/PlanNode.h
```diff
@@ -4790,31 +4790,75 @@ class MarkDistinctNode : public PlanNode {
...
+  /// include 'row number' column. This is used when computing partial results.
```
majetideepak: row_number

velox/core/PlanNode.h
```diff
@@ -4790,31 +4790,75 @@ class MarkDistinctNode : public PlanNode {
...
  public:
+  enum class RankFunction {
+    kRowNumber,
+    kRank,
+    kDenseRank,
+  };
+
+  static const char* rankFunctionName(TopNRowNumberNode::RankFunction function);
+
+  static RankFunction rankFunctionFromName(const std::string& name);
```
majetideepak: nit: use `std::string_view`

velox/core/PlanNode.h
```diff
@@ -4790,31 +4790,75 @@ class MarkDistinctNode : public PlanNode {
...
+  /// @param sortingKeys Sorting keys. May not be empty and may not intersect
+  /// with 'partitionKeys'.
+  /// @param sortingOrders Sorting orders, one per sorting key.
+  /// @param rowNumberColumnName Optional name of the column containing row
+  /// numbers (or rank and dense_rank). If not specified, the output doesn't
+  /// include 'row number' column. This is used when computing partial results.
+  /// @param limit Per-partition limit. The number of
+  /// rows produced by this node will not exceed this value for any given
+  /// partition. Extra rows will be dropped.
+  TopNRowNumberNode(
```
majetideepak, line 4836: Can you put this under `#ifdef VELOX_BACKWARD_COMPATIBILITY` as well?

velox/exec/TopNRowNumber.cpp
majetideepak: You can add a `fmt::formatter<facebook::velox::core::TopNRowNumberNode::RankFunction>`
similar to others here
https://github.com/facebookincubator/velox/blob/0d771e9eeab06500db2e874a6ed195575c6a4afb/velox/core/PlanNode.h#L5162

velox/exec/tests/PlanNodeToStringTest.cpp
```diff
@@ -927,39 +927,60 @@ TEST_F(PlanNodeToStringTest, rowNumber) {
...
-  auto plan = PlanBuilder()
-                  .tableScan(rowType)
-                  .topNRowNumber({}, {"a DESC"}, 10, false)
-                  .planNode();
-
-  ASSERT_EQ("-- TopNRowNumber[1]\n", plan->toString());
-  ASSERT_EQ(
-      "-- TopNRowNumber[1][order by (a DESC NULLS LAST) limit 10] -> a:BIGINT, b:VARCHAR\n",
-      plan->toString(true, false));
+#define TOPN_PLANNODE_TO_STRING_TEST(_nodename, _functionname)                                                           \
```
majetideepak: Can we make this a function and add a small switch to return the planNode based on the `__nodename`
We are duplicating this code otherwise for each test.
`_functionname` can be an argument.

velox/core/PlanNode.h
aditi-pandit: We are printing the string (concatenating it at some point) which implies adding a conversion from string_view to string, so it seemed safer to pass the std::string initially itself.

velox/core/PlanNode.h
aditi-pandit: done.

velox/core/PlanNode.h
aditi-pandit: done.

velox/core/PlanNode.h
aditi-pandit: done.

velox/exec/TopNRowNumber.cpp
aditi-pandit: Yes, that should work. Thanks.

velox/exec/tests/PlanNodeToStringTest.cpp
aditi-pandit: @majetideepak : It might be simpler to change the PlanBuilder API to have a single function for TopNRowNumber which takes function name as an argument instead of having separate ones for topNRowNumberNode, topNRank and topNDenseRank then. The function to return a PlanBuilder for different function name is pretty much the same code.

The code duplication here is not very big imo, and its sort of done for all the other tests as well.

wdyt ?

velox/core/PlanNode.cpp
```diff
@@ -2228,15 +2228,49 @@ PlanNodePtr RowNumberNode::create(const folly::dynamic& obj, void* context) {
...
+      static_cast<int>(function));
+  return it->second.c_str();
+}
+
+// static
+TopNRowNumberNode::RankFunction TopNRowNumberNode::rankFunctionFromName(
+    const std::string& name) {
+  static const auto kFunctionNames = invertMap(rankFunctionNames());
+  auto it = kFunctionNames.find(name);
+  VELOX_CHECK(it != kFunctionNames.end(), "Invalid rank function " + name);
```
majetideepak: VELOX_CHECK(it != kFunctionNames.end(), "Invalid rank function {}", name);

velox/core/PlanNode.h
majetideepak: We can skip the concat. I posted another comment.

velox/exec/tests/PlanNodeToStringTest.cpp
majetideepak: > single function for TopNRowNumber

This works too.

> The code duplication here is not very big imo, and its sort of done for all the other tests as well.

The Velox binary sizes are already high. We should clean up others as well.

velox/core/PlanNode.cpp
aditi-pandit: done

velox/exec/tests/PlanNodeToStringTest.cpp
aditi-pandit: I thought this test was separate from velox_exec_test, but its a part of it. Have changed the PlanBuilder API and changed the test.

velox/exec/tests/PlanNodeSerdeTest.cpp
```diff
@@ -83,6 +83,26 @@ class PlanNodeSerdeTest : public testing::Test,
     }
   }
 
+  void topNRowNumberSerdeTest(const std::string& function) {
```
majetideepak: `std::string_view function`

velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -2282,6 +2282,7 @@ PlanBuilder& PlanBuilder::rowNumber(
 }
 
 PlanBuilder& PlanBuilder::topNRowNumber(
+    const std::string& function,
```
majetideepak: `std::string_view function`

velox/core/PlanNode.cpp
```diff
@@ -2228,15 +2228,49 @@ PlanNodePtr RowNumberNode::create(const folly::dynamic& obj, void* context) {
...
+  VELOX_CHECK(
+      it != kFunctionNames.end(),
+      "Invalid rank function {}",
+      static_cast<int>(function));
+  return it->second.c_str();
+}
+
+// static
+TopNRowNumberNode::RankFunction TopNRowNumberNode::rankFunctionFromName(
+    const std::string_view& name) {
```
majetideepak: No need to pass string_view by reference. We pass by value.
`std::string_view name`

velox/core/PlanNode.h
```diff
@@ -4790,31 +4790,77 @@ class MarkDistinctNode : public PlanNode {
...
  public:
+  enum class RankFunction {
+    kRowNumber,
+    kRank,
+    kDenseRank,
+  };
+
+  static const char* rankFunctionName(TopNRowNumberNode::RankFunction function);
+
+  static RankFunction rankFunctionFromName(const std::string_view& name);
```
majetideepak: remove reference.

velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -1278,14 +1278,23 @@ class PlanBuilder {
...
   PlanBuilder& topNRowNumber(
       const std::vector<std::string>& partitionKeys,
       const std::vector<std::string>& sortingKeys,
       int32_t limit,
       bool generateRowNumber);
 
+  /// Add a TopNRowNumberNode to compute row_number, rank or dense_rank window
+  /// function with a limit applied to sorted partitions.
+  PlanBuilder& topNRowNumber(
+      const std::string& function,
```
majetideepak: `std::string_view`

velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -1278,14 +1278,23 @@ class PlanBuilder {
...
+  /// function with a limit applied to sorted partitions.
   PlanBuilder& topNRowNumber(
       const std::vector<std::string>& partitionKeys,
       const std::vector<std::string>& sortingKeys,
       int32_t limit,
       bool generateRowNumber);
 
+  /// Add a TopNRowNumberNode to compute row_number, rank or dense_rank window
+  /// function with a limit applied to sorted partitions.
+  PlanBuilder& topNRowNumber(
```
majetideepak: Should we name this `topNRow` to make it generic?

velox/core/PlanNode.h
majetideepak: I feel we can start using the new name for the newer API i.e. in the PlanBuilder.

velox/exec/tests/PlanNodeToStringTest.cpp
```diff
@@ -927,38 +927,58 @@ TEST_F(PlanNodeToStringTest, rowNumber) {
       plan->toString(true, false));
 }
 
-TEST_F(PlanNodeToStringTest, topNRowNumber) {
+namespace {
+void topNPlanNodeToStringTest(const std::string& function) {
```
majetideepak: topNRowNodeToStringTest?

velox/exec/tests/PlanNodeSerdeTest.cpp
majetideepak: TopNRowNodeSerdeTest?

velox/exec/tests/utils/PlanBuilder.h
aditi-pandit: This class of functions is called ranking functions (so row_number is a ranking function), so I feel topNRank is more appropriate. I'm open to change PlanBuilder APIs at this point.

The original topNRowNumber can stay the same as such, to cover backward compatible calls.

velox/exec/tests/PlanNodeSerdeTest.cpp
aditi-pandit: done.

velox/exec/tests/utils/PlanBuilder.cpp
aditi-pandit: done.

velox/core/PlanNode.cpp
aditi-pandit: done

velox/core/PlanNode.h
aditi-pandit: done.

velox/exec/tests/utils/PlanBuilder.h
aditi-pandit: done

velox/exec/tests/PlanNodeToStringTest.cpp
aditi-pandit: Changed to topNRankPlanNodeToStringTest.

velox/exec/tests/PlanNodeSerdeTest.cpp
aditi-pandit: changed to topNRankSerdeTest

aditi-pandit:
@majetideepak : Have addressed review comments. PTAL.

majetideepak (approved):

majetideepak:
@aditi-pandit can you please take a look at the failures? Thanks!

aditi-pandit:
@majetideepak : The test timed out. There weren't any failures in the new tests introduced. I've rebased and pushed again.

majetideepak (approved):

majetideepak:
@xiaoxmeng can you please help review this? The CI failures are unrelated and are being fixed separately.

aditi-pandit:
@majetideepak @xiaoxmeng : The test failures are fixed because of Christians change to increase timeouts. This PR CI is clean now.

majetideepak:
@xiaoxmeng can you please take a look at this? Thanks!

aditi-pandit:
@xiaoxmeng : Gentle ping. Please can you help with review and merge.

velox/core/PlanNode.h
```diff
@@ -4781,31 +4781,77 @@ class MarkDistinctNode : public PlanNode {
...
+/// support for handling rank and dense_rank is committed to Velox.
 class TopNRowNumberNode : public PlanNode {
  public:
+  enum class RankFunction {
+    kRowNumber,
+    kRank,
+    kDenseRank,
+  };
+
+  static const char* rankFunctionName(TopNRowNumberNode::RankFunction function);
```
xiaoxmeng, line 4798: TopNRowNumberNode is not required here.

aditi-pandit:
@xiaoxmeng : Gentle ping. Please can you help with the review and merge.

PR #13261, opened 2025-05-07
refactor(simd): Upgrade xsimd to 13.1 (from 10.0)

serge-sans-paille:
This update requires a refactor of the way xsimd::batch<bool> are handled. They use to actually encode bitset, so instead of batch<bool>, which is no longer supported by xsimd, use batch<uint8_t>. This actually make the code cleaner in several places, so go for it.


serge-sans-paille:
gentle ping :-)

serge-sans-paille:
@majetideepak : ping?

assignUser:
mhm, we need to make sure this doesn't clash with the version of xsimd used at meta internally.cc @kKPulla I assume internally xsimd is still 10.x?

assignUser:
Sorry, that was pretty short. First off thanks for the contribution! 

For dependencies we are constrained by the versions used in the meta monorepo, at least for upgrades that require changes to the actual source, as a version update like that  in the monorepo is quite the undertaking.

In special cases it might be necessary to diverge and use  `#ifdef`s but that would need good arguments. (I don't think we have done that yet?)

serge-sans-paille:
It's possible that the patch also works for the previous version of xsimd, I just pushed a patch to check that. Could you give CI a try?

serge-sans-paille:
All looks good then, do you want me to manually squash everything?

assignUser:
Ok, looks like this works with both versions, nice! @Yuhta could you review, or ping someone else familiar with the simd code?

serge-sans-paille:
@assignUser / @Yuhta : gentle ping?

assignUser:
I send a link to this PR asking for a cpp review in the velox slack.

kgpai:
@Yuhta

velox/common/base/SimdUtil.h
```diff
@@ -25,6 +25,10 @@
 
 namespace facebook::velox::simd {
 
+template <typename T, typename A = xsimd::default_arch>
+using xbatch =
+    xsimd::batch<std::conditional_t<std::is_same_v<T, bool>, uint8_t, T>, A>;
```
Yuhta: Can we use `uint64_t` as the element, aligning with the non-vectorized case?

serge-sans-paille: not quite : that would mean only two elements per batch for SSEx, and only four on AVX

serge-sans-paille: Plus the C++ standard does not give any guarantee considering `sizeof(bool)`. On my x86_64 Intel hardware `sizeof(bool) == 1`, so I don't quite understand why you mention `uint64_t` here?

serge-sans-paille:
@Yuhta : gentle ping :-)

serge-sans-paille:
@Yuhta : gentle ping :-) I can rebase the PR, but if I do so, I'd like to make sur it gets merged and doesn't stale for another 6 months or so /o\

velox/common/base/SimdUtil.h
Yuhta: Booleans in velox are represented using bits, not `bool` types.  In scalar code we use `uint64_t` so it's 64 elements per word; in SIMD it should be 128 for SSE and 256 for AVX, no matter what element type we use.  Using `uint64_t` is just for consistency with scalar version.

serge-sans-paille: I pushed an update following that idea, let's see what CI thinks of it.

serge-sans-paille:
@Yuhta : I don't have the rights to trigger CI

Yuhta (commented):
I kicked off the CI, it should be good if all tests pass

serge-sans-paille:
@Yuhta : these was an error triggered by a path I did not cover locally. I just pushed an update.

ravigumm:
Can this PR please be merged ? Thanks

mbasmanova:
CI is red. @serge-sans-paille Can you take a look? @ravigumm Would you like to take over?

serge-sans-paille:
@mbasmanova : can you retrigger CI?

mbasmanova (changes requested):
@serge-sans-paille Sorry for the long wait on this PR. CI has been retriggered. Please also update the xsimd version in `CMake/resolve_dependency_modules/README.md`.

serge-sans-paille:
Thanks! I'll submit a fixup on CI finishes. Should I worry about the fuzzing failure?

mbasmanova:
> Should I worry about the fuzzing failure?

No. @kgpai is fixing these.

serge-sans-paille:
mmmh, failing test `velox_exec_test_group7`. Does that ring any bell?

mbasmanova:
@serge-sans-paille Looks like CI flakiness. I re-ran. Please update the xsimd version in CMake/resolve_dependency_modules/README.md and I'll merge.

mbasmanova:
@serge-sans-paille If CI re-run times out again, check https://github.com/facebookincubator/velox/pull/13261#issuecomment-4463719164

mbasmanova:
@serge-sans-paille CI re-run passed. Please update the xsimd version in CMake/resolve_dependency_modules/README.md and I'll merge.

serge-sans-paille:
fixup added! Thanks

mbasmanova (approved):

mbasmanova:
@serge-sans-paille Heads up on the landing path: the Meta-internal fbsource third-party/xsimd is on 9.0.1, while OSS is on 10.0. This PR targets 13.1 but only updates the OSS CMake — the internal build will compile this code against the old xsimd, which will fail.

Previous xsimd upgrades were coordinated between internal third-party imports and the OSS side. We'll need to do the same here.

@kgpai @Yuhta Can you advise on how to coordinate the internal xsimd bump with this PR?

serge-sans-paille:
@mbasmanova ok. I originally opened that PR to fix an issue in xsimd bug tracker - the issue happened to be on velox side. I think I've done my part there, I'm leaving the remaining parts to you (FYI: xsimd current release is now 14.2)

mbasmanova:
> I originally opened that PR to fix an issue in xsimd bug tracker 

Thank you for the context. Mind sharing link to that issue?

mbasmanova (changes requested):
@serge-sans-paille 

The `xbatch<bool>` alias is a nice idea, but `xbatch<bool>::size` returns 4 (uint64 lane count) instead of 256 (the conceptual bool element count). That contradicts the alias's own comment — "we use one bit to represent one element" — and any caller that asks "how many elements does this batch hold?" gets the wrong answer.

I hit two manifestations in fbcode CI (OSS CI was green because `VELOX_ENABLE_LOAD_SIMD_VALUE_BUFFER` is fbcode-only, so the entire `loadSIMDValueBufferAt` body and the `SimpleVectorTest` SIMD instantiations are `#ifdef`-d out in OSS):

1. **`FlatVector-inl.h:79`** — `return xsimd::load_unaligned(mem)` deduced `batch<uint8_t>` from `uint8_t* mem`, but the return type is now `xbatch<bool>` = `batch<uint64_t>` — no implicit conversion, compile error. Fixed by `reinterpret_cast<uint64_t*>(mem)`.
2. **`SimpleVectorTest.cpp`** — `width = xbatch<T>::size` silently dropped from 256 to 4 for bool, so the bool case checked only 4 elements per load instead of 256 and produced incoherent comparisons against `expected[]`. Fixed by `width = is_bool ? 256 : xbatch<T>::size`, but that's a symptom — every future caller of `xbatch<bool>::size` will hit the same trap.

Could you replace the alias with a small wrapper struct that exposes `size = 256` for bool? That keeps `xbatch<T>::size` honest across all `T` and avoids per-caller special-casing.

serge-sans-paille:
The original issue was https://github.com/xtensor-stack/xsimd/issues/1113

serge-sans-paille:
> @serge-sans-paille
> 
> The `xbatch<bool>` alias is a nice idea, but `xbatch<bool>::size` returns 4 (uint64 lane count) instead of 256 (the conceptual bool element count). That contradicts the alias's own comment — "we use one bit to represent one element" — and any caller that asks "how many elements does this batch hold?" gets the wrong answer.
> 
> I hit two manifestations in fbcode CI (OSS CI was green because `VELOX_ENABLE_LOAD_SIMD_VALUE_BUFFER` is fbcode-only, so the entire `loadSIMDValueBufferAt` body and the `SimpleVectorTest` SIMD instantiations are `#ifdef`-d out in OSS):
> 
>     1. **`FlatVector-inl.h:79`** — `return xsimd::load_unaligned(mem)` deduced `batch<uint8_t>` from `uint8_t* mem`, but the return type is now `xbatch<bool>` = `batch<uint64_t>` — no implicit conversion, compile error. Fixed by `reinterpret_cast<uint64_t*>(mem)`.
> 
>     2. **`SimpleVectorTest.cpp`** — `width = xbatch<T>::size` silently dropped from 256 to 4 for bool, so the bool case checked only 4 elements per load instead of 256 and produced incoherent comparisons against `expected[]`. Fixed by `width = is_bool ? 256 : xbatch<T>::size`, but that's a symptom — every future caller of `xbatch<bool>::size` will hit the same trap.
> 
> 
> Could you replace the alias with a small wrapper struct that exposes `size = 256` for bool? That keeps `xbatch<T>::size` honest across all `T` and avoids per-caller special-casing.

I have to apologize, but I'm not eager to invest more time on this, fixing a bug on sources I don't have access to on my hobby time is not a thrilling activity :-)

Yuhta:
@mbasmanova Upgrading xsimd internally is a larger effort than item fits in an oncall cycle. Basically
1. Find out all the projects depending on the code
2. Analyze the incompatible changes (this step is easer now with AI; used to be very hard or impossible)
3. If the changes are all backward compatible, we may just import and bump it (still a few days of work)
4. If not, we need to set up some communication channels for the downstream projects that we are going to change this and fix their projects
5. Then we do the same as 3, it actually contains 2 steps, importing the code (but it co-exists with previous version, controlled by some link in filesystem), here we only import the code, fix the build, and do not flip the link
6. In a commit fix the downstream dependencies and flip the link

So this is something needs a place in roadmap planning.  Another option is we should take all our dependencies and vendor them.  I think the latter is we probably should go if we want to be GitHub first.  PyTorch is doing the similar thing.  Otherwise every dependency change is a backward incompatible change.

PR #13264, opened 2025-05-07
refactor: Enhance RowComparator for rank function comparisons

aditi-pandit:
TopNRank comparisons require an API to return if the comparisons are <, > or =. The previous code only returned a bool for = or not.


velox/exec/TopNRowNumber.cpp
```diff
@@ -249,25 +249,52 @@ void TopNRowNumber::initializeNewPartitions() {
...
-    newRow = data_->newRow();
+  if (partition.topRank < limit_) {
+    newRow = processRowBeforeLimit(index, partition);
   } else {
     char* topRow = topRows.top();
 
-    if (!comparator_(decodedVectors_, index, topRow)) {
-      // Drop this input row.
+    auto result = comparator_.compare(decodedVectors_, index, topRow);
+    if (result < 0) {
```
JkSelf, line 284: Can we combine these conditions as following?
```
if (result <= 0) {
return;
} else {
newRow = processRowBeyondLimit(index, partition);
}
```

velox/exec/TopNRowNumber.h
```diff
@@ -84,6 +84,10 @@ class TopNRowNumber : public Operator {
     std::priority_queue<char*, std::vector<char*, StlAllocator<char*>>, Compare>
         rows;
 
+    // This is the highest rank seen so far in the input rows. It is compared
+    // with the limit for the operator.
+    int64_t topRank = 0;
```
JkSelf, line 89: `topRank `-> `topNRowNumbers`?

velox/exec/TopNRowNumber.h
```diff
@@ -97,6 +101,14 @@ class TopNRowNumber : public Operator {
   // Decodes and potentially loads input if lazy vector.
   void prepareInput(RowVectorPtr& input);
 
+  // Handles input row when the partition has not yet accumulated 'limit' rows.
+  // Returns a pointer to the row to add to the partition accumulator.
+  char* processRowBeforeLimit(vector_size_t index, TopRows& partition);
```
JkSelf, line 106: How about renaming to `processRowsWithinLimit `and `processRowsExceedingLimit`?

velox/exec/TopNRowNumber.h
aditi-pandit: I'm trying to update this code so that there are few diffs when enhancing for rank functions. So named accordingly. I'll revise this comment indicating that.

velox/exec/TopNRowNumber.cpp
aditi-pandit: Same as below. rank and dense_rank have some difference in behavior, so retaining it to avoid further diffs.

aditi-pandit:
@JkSelf : Have addressed the comments. PTAL.

velox/exec/TopNRowNumber.cpp
```diff
@@ -249,25 +249,52 @@ void TopNRowNumber::initializeNewPartitions() {
...
+  if (partition.topRank < limit_) {
+    newRow = processRowsWithinLimit(index, partition);
   } else {
     char* topRow = topRows.top();
 
-    if (!comparator_(decodedVectors_, index, topRow)) {
-      // Drop this input row.
+    auto result = comparator_.compare(decodedVectors_, index, topRow);
+    if (result < 0) {
+      // The new row is higher than the top rank so far, so this row is ignored.
```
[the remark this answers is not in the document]
aditi-pandit: Actually this comparison is reverse... result < 0 if topRank < decodedVectors_[index]... I should change the order of the arguments to make it look consistent. Let me try that.

velox/exec/RowContainer.cpp
```diff
@@ -1275,36 +1275,47 @@ RowComparator::RowComparator(
...
+      return result;
     }
   }
-  return false;
+  return 0;
 }
 
-bool RowComparator::operator()(
+bool RowComparator::operator()(const char* lhs, const char* rhs) {
+  return compare(lhs, rhs) <= 0;
```
majetideepak: If `lhs = rhs`, then we `return 0 <= 0`, which is `true`. It is supposed to be `false`.
should be `compare(lhs, rhs) < 0;`

majetideepak: Don't we have a test to catch this?

velox/exec/RowContainer.cpp
```diff
@@ -1275,36 +1275,47 @@ RowComparator::RowComparator(
...
+}
+
+int32_t RowComparator::compare(
+    const char* lhs,
     const std::vector<DecodedVector>& decodedVectors,
-    vector_size_t index,
-    const char* rhs) {
+    vector_size_t index) {
```
majetideepak: Can we not continue with the existing API? It seems less risky not to change.
``` 
    const std::vector<DecodedVector>& decodedVectors,
      vector_size_t index,
      const char* rhs);
```

aditi-pandit: In the original PR I had retained the current API.

The usage of the current API is a bit awkward (as rhs is used as lhs in the comparison) so its counterintuitive, and I also changed the way comparisons were done to make it easier to use. This came as a result of a review comment.

The code is used only in TopN and TopNRowNumber exercised by Fuzzer.

Its definitely less risky to keep the original API. And I will cheange to it.

velox/exec/RowContainer.cpp
aditi-pandit: See comment below.

velox/exec/RowContainer.cpp
```diff
@@ -1302,9 +1306,16 @@ bool RowComparator::operator()(
             decodedVectors[key.first],
             index,
             {key.second.isNullsFirst(), key.second.isAscending(), false})) {
-      return result > 0;
+      return -result;
```
majetideepak, line 1309: nit: let's `return result` and use `compare(decodedVectors, index, rhs) > 0` below to minimize change.

majetideepak: I see this allows `< 0 for decodedVectors[index] < rhs`. This is fine then.

velox/exec/TopNRowNumber.cpp
```diff
@@ -249,25 +249,52 @@ void TopNRowNumber::initializeNewPartitions() {
...
-      // Drop this input row.
+    auto result = comparator_.compare(decodedVectors_, index, topRow);
+    if (result > 0) {
+      // The new row is bigger than the top rank so far, so this row is ignored.
       return;
     }
 
-    // Replace existing row.
-    topRows.pop();
+    if (result == 0) {
```
majetideepak, line 289: Let's combine both `if (result >= 0) { return; }` and clarify the semantics in the comments.

majetideepak (approved):

majetideepak:
@xiaoxmeng can you please take a look? Thanks!

velox/exec/TopNRowNumber.cpp
aditi-pandit: I specifically added this way as the rank functions only change the logic for if (result == 0) case...Agree this can be combined for a bit, but feel it will help the final PR.

velox/exec/RowContainer.h
```diff
@@ -1873,12 +1873,22 @@ class RowComparator {
   /// Returns true if lhs < rhs, false otherwise.
   bool operator()(const char* lhs, const char* rhs);
 
-  /// Returns true if decodeVectors[index] < rhs, false otherwise.
+  /// Returns 0 for equal, < 0 for lhs < rhs, > 0 otherwise.
+  int compare(const char* lhs, const char* rhs);
```
xiaoxmeng, line 1877: static method?

velox/exec/RowContainer.h
```diff
@@ -1873,12 +1873,22 @@ class RowComparator {
   /// Returns true if lhs < rhs, false otherwise.
   bool operator()(const char* lhs, const char* rhs);
```
xiaoxmeng, line 1874: ditto?

velox/exec/RowContainer.h
```diff
@@ -1873,12 +1873,22 @@ class RowComparator {
...
 
-  /// Returns true if decodeVectors[index] < rhs, false otherwise.
+  /// Returns 0 for equal, < 0 for lhs < rhs, > 0 otherwise.
+  int compare(const char* lhs, const char* rhs);
+
+  /// Returns true if decodedVectors[index] < rhs, false otherwise.
   bool operator()(
       const std::vector<DecodedVector>& decodedVectors,
       vector_size_t index,
       const char* rhs);
```
xiaoxmeng:
```
nit: s/rhs/other/
```

velox/exec/RowContainer.h
```diff
@@ -1873,12 +1873,22 @@ class RowComparator {
...
 
+  /// Returns 0 for equal, < 0 for decodedVectors[index] < rhs,
+  /// > 0 otherwise.
+  int32_t compare(
+      const std::vector<DecodedVector>& decodedVectors,
+      vector_size_t index,
+      const char* rhs);
```
xiaoxmeng: ditto

velox/exec/TopNRowNumber.cpp
```diff
@@ -249,25 +249,52 @@ void TopNRowNumber::initializeNewPartitions() {
   }
 }
 
+char* TopNRowNumber::processRowWithinLimit(
+    vector_size_t /*index*/,
+    TopRows& partition) {
+  // row_number accumulates the new row in the partition, and the top rank is
+  // incremented by 1 as row_number increases by 1 at each new row.
+  partition.topRank++;
```
xiaoxmeng: nit: ++partition.topRank;

velox/exec/TopNRowNumber.cpp
```diff
@@ -249,25 +249,52 @@ void TopNRowNumber::initializeNewPartitions() {
...
-  if (topRows.size() < limit_) {
-    newRow = data_->newRow();
+  if (partition.topRank < limit_) {
+    newRow = processRowWithinLimit(index, partition);
   } else {
     char* topRow = topRows.top();
 
-    if (!comparator_(decodedVectors_, index, topRow)) {
-      // Drop this input row.
+    auto result = comparator_.compare(decodedVectors_, index, topRow);
```
xiaoxmeng:
```
nit: const auto result = 
```

xiaoxmeng (commented):
@aditi-pandit LGTM % nits. Thanks!

velox/exec/RowContainer.h
aditi-pandit: @xiaoxmeng : These methods access non-static fields of the RowComparator, so don't think they can be static.

velox/exec/RowContainer.h
aditi-pandit: same as above

aditi-pandit:
@xiaoxmeng : Thanks for the review Have addressed your comments. PTAL.

xiaoxmeng (approved):
@aditi-pandit thanks for the update!

Issue #13260, opened 2025-05-07
Add spill support for sort merge join operator to prevent OOM when merge a large number of local sources


xiaoxmeng:
### Description

To address the query OOM issue when clustering hourly data partitions to daily partitions, we propose adding spill support to the Merge operator. This solution involves extending the Merge operator to process multiple local sources at a time, leveraging existing spilling facilities to store intermediate results, and introducing lazy start processing for merge sources:
1. Extend Merge Operator
    Process up to N local sources at a time
    Perform M/N merge operations to produce M/N sorted streams (where M is the total number of sources)
    Recursively merge results until fewer than N local sources remain
2. Lazy Start Processing for Merge Sources
    Allow merge sources to wait for a start signal from the Merge operator
    Prevent auto-triggering of read operations from merge sources during parallelized execution mode
3. Auto-Detect N Based on Memory Usage
    Leverage actual memory usage measurements from memory pool to find an optimal balance between memory and latency
4. Extend Spill File Input Stream 
    Handle multiple sorted spilled files produced by the same Merge operator
    Utilize a concatenate iterator to read from these files one at a time, in order

PR #13269, opened 2025-05-08
feat: Zip the sorting column indices and compare flags

duanmeng:
Currently, the sorted spiller moves the sorting columns to the front
and uses the number of sorting keys to access them. Zip the sorting
column indices and compare flags into a vector of pairs, decoupling
the RowVector layout from the sorting process. Hence, it can handle
both cases: when the sorting keys are adjusted and when no adjustment
is made.

Part of #13260

### DIFF
--- a/velox/exec/Spill.cpp
+++ b/velox/exec/Spill.cpp
@@ -23,20 +23,6 @@
 using facebook::velox::common::testutil::TestValue;
 
 namespace facebook::velox::exec {
-namespace {
-// Returns the CompareFlags vector whose size is equal to numSortKeys. Fill in
-// with default CompareFlags() if 'compareFlags' is empty.
-const std::vector<CompareFlags> getCompareFlagsOrDefault(
-    const std::vector<CompareFlags>& compareFlags,
-    int32_t numSortKeys) {
-  VELOX_DCHECK(compareFlags.empty() || compareFlags.size() == numSortKeys);
-  if (compareFlags.size() == numSortKeys) {
-    return compareFlags;
-  }
-  return std::vector<CompareFlags>(numSortKeys);
-}
-} // namespace
-
 void SpillMergeStream::pop() {
   VELOX_CHECK(!closed_);
   if (++index_ >= size_) {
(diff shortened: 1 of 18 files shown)


velox/exec/Spill.cpp
```diff
@@ -117,6 +84,45 @@ SpillState::SpillState(
       pool_(pool),
       stats_(stats) {}
 
+std::vector<std::pair<column_index_t, CompareFlags>>
+SpillState::makeSortingKeys(
+    size_t numKeys,
```
aditi-pandit, line 89: @duanmeng : If numKeys is always compareFlags.size, then we can avoid that parameter entirely. Is there something else happening here ?

velox/exec/Spiller.cpp
```diff
@@ -60,6 +58,10 @@ SpillerBase::SpillerBase(
           spillStats,
           spillConfig->fileCreateConfig) {
   TestValue::adjust("facebook::velox::exec::SpillerBase", this);
+  compareFlags_.reserve(sortingKeys.size());
```
xiaoxmeng, line 61: Can we make this a const member and provide an anonymous function to convert from sortkeys to compare flags? Also we should have sanity check that the sort key channels are prefix of input?

velox/exec/Spill.cpp
```diff
@@ -117,6 +84,45 @@ SpillState::SpillState(
...
+    const std::vector<CompareFlags>& compareFlags) {
+  if (numKeys == 0) {
+    VELOX_CHECK(compareFlags.empty());
+    return std::vector<std::pair<column_index_t, CompareFlags>>{};
+  }
+
+  std::vector<std::pair<column_index_t, CompareFlags>> sortingKeys;
+  sortingKeys.reserve(numKeys);
+  if (compareFlags.empty()) {
+    for (auto i = 0; i < numKeys; i++) {
```
xiaoxmeng, line 99:
```
s/i++/++i/
```

velox/exec/Spill.cpp
xiaoxmeng: I think compare flags could be empty as defaults. See code below @duanmeng @tanjialiang do we have such use cases? Thanks!

velox/exec/Spill.cpp
```diff
@@ -117,6 +84,45 @@ SpillState::SpillState(
...
+      sortingKeys.emplace_back(i, CompareFlags{});
+    }
+    return sortingKeys;
+  }
+
+  VELOX_CHECK_EQ(numKeys, compareFlags.size());
+  for (auto i = 0; i < numKeys; i++) {
```
xiaoxmeng, line 106: nit: ++i

xiaoxmeng (approved):
@duanmeng LGTM % nits. Thanks!

velox/exec/Spiller.h
```diff
@@ -160,6 +159,8 @@ class SpillerBase {
   // Collects the rows to spill for each partition.
   folly::F14FastMap<SpillPartitionId, SpillRun> spillRuns_;
 
+  std::vector<CompareFlags> compareFlags_;
```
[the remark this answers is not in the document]
duanmeng: It is used by `SpillerBase::ensureSorted` to sort.

xiaoxmeng (approved):
@duanmeng thanks for the update. It looks nice!

PR #9079, opened 2024-03-14
Store duplicate row address in vector for join probe

zhli1142015:
Problem
When there are a large number of rows with the same key in the build side, 
the `listJoinResults` function becomes very time-consuming.

Design
`appendNextRow`
Create a next-row-vector if it doesn't exist. Append the row address to
the next-row-vector, and store the address of the next-row-vector in the
`nextOffset_` slot for all duplicate rows.

`listJoinResults`
To retrieve the addresses of all rows with the same keys, we first obtain the 
address of the first row using the hash function, then, by the `nextOffset_`, 
we retrieve the address of the next-row-vector. Then, we iterate through the 
next-row-vector to obtain the addresses of the remaining rows. We can utilize 
SIMD instructions to accelerate the next-row-vector access. 

When a row needs to be erased, if value in `nextOffset_` slot is not null, then it
will be removed from corresponding next-row-vector and set it's `nextOffset_` 
slot as null.

The current design is applicable to all hash modes.

Benchmark
The results indicate that this PR can accelerate the `listJoinResults` function,
with the acceleration effect becoming more pronounced as the proportion of 
rows with the same key increases.

Fixes #9078

PR #13271, opened 2025-05-08
fix: Revert #9079

arhimondr:
Summary:
#9079 optimizes listJoinResults when the number of duplicates for a certain row
is high (>5).

However the optimization increases cost of building a hash table significnatly
for cases when the number of duplicates for each entry is low (<5) and the
number of entries with duplicates is high. Allocating vectors for every entry
is expensive, both memory and CPU wise.

Differential Revision: D74405114


arhimondr:
TPC-H, sf100, CBO Disabled

**Trunk**

Screenshot 2025-05-08 at 11 50 24 AM


The total runtime is more than 60s. Q21 filed with out of memory.

**With #9079 reverted**

Screenshot 2025-05-08 at 11 50 35 AM

The total runtime is 35s. Q21 succeeded.


Observation

With CBO disabled Lineitem table often ends up on build side. When joined on a key such as order_key, for each order_key there are several entries in the Lineitem table. The engine ends up spending a lot of time allocating and re-allocating vectors to store pointers to duplicates.

arhimondr:
While the vector approach provides superior performance of the listJoinResults method when the number of duplicates is high the cost is significant.

It is worth exploring an alternative approach.

For example for keys with the number of duplicates greater than 5 it might be worth caching the list of duplicates on retrieval. 

Caching on retrieval allows to optimize allocations. For example a single contiguous vector can be used instead of multiple tiny vectors. It would also allow to get best of two words. For entires with low number of duplicates the linked list approach tends to perform better.

arhimondr:
CC: @zhli1142015

zhli1142015:
Thank you for raising this issue. I agree with your point. Internally, we keep both approaches and choose the appropriate one—vector or linked list—based on certain conditions.

Specifically, we use vector approach when row count > 2* estimate of cardinality || table is in array mode and row count > some threshold.
We use `capacity_` as estimate of cardinality, but this value is typically much larger than the actual cardinality, so for array-mode tables with many rows, we assume there will also be many duplicate rows.

In practice, we validate this heuristic using TPC-H Q21 and TPC-DS Q72(CBO disabled also), and it yields good performance for both queries. If you think this makes sense and would help your use case, I can push our changes to velox.

Also, you mentioned using a single contiguous vector as an alternative. I may not have fully understood that approach—could you share more details?

arhimondr:
> Also, you mentioned using a single contiguous vector as an alternative. I may not have fully understood that approach—could you share more details?

I was thinking about the following approach

1. During HashBuild link duplicates via a linked list (nextRow)
2. During HashProbe create cache of duplicates for entires with more than N (5?) duplicates

The cache can be structured as following

`map<pointer_to_first_row, cache_entry_struct>`

Where

cache_entry_struct is as following

```
number_of_duplicates
pointer_to_list_of_duplicates
```

Where 

pointer_to_list_of_duplicates points to a position in a single contiguous vector

The vector can be expanded when adding new cache entries

arhimondr:
> I thought a key has many build‐side duplicates doesn’t mean it will appear often in the probe relation.

If the key does not appear often (at least twice) on the probe side do you think it may still cause a significant impact on overall query performance?

arhimondr:
@zhli1142015 I'm going to land this change for now, but let's continue discussion and come up with a solution that works great for both use cases.

xiaoxmeng (approved):

arhimondr:
@zhli1142015 Let's keep the discussion open and find a solution that works for both workflows

PR #13273, opened 2025-05-08
feat: Add task barrier support for streaming aggregation

xiaoxmeng:
Summary: Add to support barrier support for streaming aggregation. Also fix the output batch size respect after receive no more input signal. Some sequence storage test improvement with more barrier execution mode coverage.

Differential Revision: D74369703

### DIFF
--- a/velox/exec/StreamingAggregation.h
+++ b/velox/exec/StreamingAggregation.h
@@ -43,6 +43,8 @@ class StreamingAggregation : public Operator {
     return true;
   }
 
+  bool startDrain() override;
+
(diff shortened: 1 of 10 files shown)


velox/exec/StreamingAggregation.h
```diff
@@ -43,6 +43,8 @@ class StreamingAggregation : public Operator {
     return true;
   }
 
+  bool startDrain() override;
```
tanjialiang, line 46: Can you change Operator::startDrain() method to not return? It already throws.

tanjialiang (approved):
LGTM

PR #13280, opened 2025-05-08
fix: Rewrite equality check operator for LambdaTypedExpr to fix type ambiguity issue in C++20

kKPulla:
Summary: Added a member ``operator==(const LambdaTypedExpr&)`` to explicitly check two lambda exprs for equality. Call this from the generic one that takes ``ITypedExpr`` after casting successfully.

Differential Revision: D74434139


pedroerp (approved):
Nice and clean :)

Thanks @kKPulla

PR #13292, opened 2025-05-09
build: Allow custom arrow build during setup

czentgr:
This allows a user to reuse the functions and pass in separate arrow options, e.g. to add arrow flight while at the same time ensuring that the same version of arrow and the same options needed for velox are bing used.


FelixYBW:
@PHILO-HE

PHILO-HE:
Looks good! Do we also need to modify the file below? https://github.com/facebookincubator/velox/blob/main/CMake/resolve_dependency_modules/arrow/CMakeLists.txt

czentgr:
@PHILO-HE Good question. Initially, I intended the variable to be used for setups scripts (specifically there is a use case for Prestissimo). But if the library is picked up from bundling could an upstream project also inject additional options?
I suspect, we could add this capability.

czentgr:
Solved differently.

PR #13297, opened 2025-05-09
test: Enable fuzz tests for BingTile functions (#12924)

bikramSingh91:
Summary:

Previously, fuzz tests would raise errors for functions that
took BingTile arguments, because the fuzzer would default to bigint,
and most bigints are not valid BingTiles.  Thus we'd get runtime errors.

This creates a BingTile input type generator that only creates valid
BingTiles.  Since BingTiles are validated on creation, if the argument
is already a BingTile, it is valid.  Thus our fuzzer should also use
valid BingTiles.

Reviewed By: kgpai

Differential Revision: D72434294

### DIFF
--- a/velox/common/fuzzer/CMakeLists.txt
+++ b/velox/common/fuzzer/CMakeLists.txt
@@ -18,8 +18,12 @@ velox_link_libraries(velox_common_fuzzer_util velox_type velox_exception)
 
 velox_add_library(velox_constrained_input_generators ConstrainedGenerators.cpp)
 
-velox_link_libraries(velox_constrained_input_generators Folly::folly velox_type
-                     velox_common_fuzzer_util)
+velox_link_libraries(
+  velox_constrained_input_generators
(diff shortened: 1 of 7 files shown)


velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -331,6 +331,11 @@ VectorPtr VectorFuzzer::fuzzConstant(
     if (type->isUnKnown()) {
       return BaseVector::createNullConstant(type, size, pool_);
     } else {
+      auto inputGenerator = customGenerator;
+      if (!inputGenerator && customTypeExists(type->name())) {
```
kgpai, line 335: Isnt this from your previous change ?

kgpai: NVM. stacked diffs..

kgpai (approved):

PR #13298, opened 2025-05-09
test: Add EnsureWritableVectorTest for lazy vector

SuperMario210:
Summary: This diff adds a new unit test to verify the fix introduced in D74420826.

Differential Revision: D74502244


pedroerp (approved):
Thanks for the retroactive test.

PR #13316, opened 2025-05-12
fix: FloatingPointRange::testDoubleRange

Yuhta:
Summary:
When either side of the filter range is unbounded, and the
corresponding upper/lower is set to a finite value,
`FloatingPointRange::testDoubleRange` would use the upper/lower value and ignore
the fact that it's unbounded on that side.  In Prestissimo the upper/lower is
set to infinity or negative infinity if the side is unbounded, so we do not see
this bug there.

Differential Revision: D74538066

### DIFF
--- a/velox/type/Filter.h
+++ b/velox/type/Filter.h
@@ -1416,7 +1416,9 @@ class FloatingPointRange final : public AbstractRange {
       return true;
     }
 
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@Yuhta thanks for the fix!

PR #13322, opened 2025-05-12
fix: Added type validation for intermediate and output vectors

kletkavrubashku:
Summary:
X-link: https://github.com/facebookexperimental/verax/pull/11

Fixed a crash in D74199165. Adding validation to Velox to simplify future debugging.

TL;DR We don't validate output types in aggregate functions.
For example, if intermediateType in Coordinator is `Row<int>` and in Velox it's `Row<double>`, no validation will occur. Dynamic cast for FlatVectors doesn't catch this mismatch either, because `RowVector` isn't a template. Without validation we can write by wrong offset in the result vector.

Differential Revision: D74435817

### DIFF
--- a/velox/exec/Aggregate.cpp
+++ b/velox/exec/Aggregate.cpp
@@ -244,6 +244,32 @@ std::vector<CompanionSignatureEntry> getCompanionSignaturesWithSuffix(
   return entries;
 }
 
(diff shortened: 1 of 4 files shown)


kletkavrubashku:
- Renamed returnType to finalType
- Extracted some common code to helper function

Yuhta (approved):

PR #13325, opened 2025-05-12
feat: Add dynamic_filter_enabled property

arhimondr:
Summary: To allow disabling dynamic filtering when necessary

Differential Revision: D74606332

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -598,6 +598,9 @@ class QueryConfig {
   static constexpr const char* kFieldNamesInJsonCastEnabled =
       "field_names_in_json_cast_enabled";
 
+  /// Enable dynamic filter for broadcast joins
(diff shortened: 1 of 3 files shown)


velox/core/QueryConfig.h
```diff
@@ -598,6 +598,9 @@ class QueryConfig {
   static constexpr const char* kFieldNamesInJsonCastEnabled =
       "field_names_in_json_cast_enabled";
 
+  /// Enable dynamic filter for broadcast joins
+  static constexpr const char* kDynamicFilterEnabled = "dynamic_filter_enabled";
```
pedroerp, line 602: nit: should we call this "dynamic filter pushdown" to be more explicit?

arhimondr:
Probably not needed after https://github.com/facebookincubator/velox/pull/13784

PR #13326, opened 2025-05-12
feat: Add rule based input generation

peterenescu:
Summary:
Add rule or grammar based input generation.

This will allow us to have an extensible, easy-to-use, structured format to generate input that can be used by our fuzzer custom input generators.

Although, we can expand upon this list, I've created several classes that can:
* Repeat
* Be randomly included (or excluded)
* Define basic strings like numbers and constants
* Choose randomly from a list.

Differential Revision: D74610765


igs-raiderler:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #13328, opened 2025-05-13
fix: Enable cache fuzz test

zhouyuan:
This patch re-enabled cache fuzz test. It was disabled due to flaky and should be fixed now


zhouyuan:
@kgpai would you please help to take a look? Thanks!

PR #13332, opened 2025-05-13
feat: Geometry overlay operations

jagill:
Summary:
This implements the four core overlay operations: Union, Intersection, Difference, and Symmetric Difference.
For Union, this only implements the binary operation, leaving `geometry_union(array<geometry>)` for a later PR.

Differential Revision: D74671626

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -142,6 +142,10 @@ int main(int argc, char** argv) {
       "st_overlaps",
       "st_touches",
       "st_within",
+      "st_difference",
+      "st_intersection",
+      "st_symdifference",
+      "st_union",
   };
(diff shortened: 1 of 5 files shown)


velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -132,6 +132,15 @@ int main(int argc, char** argv) {
...
       "st_asbinary",
+      "st_relate",
+      "st_contains",
+      "st_crosses",
+      "st_disjoint",
+      "st_equals",
+      "st_intersects",
+      "st_overlaps",
+      "st_touches",
+      "st_within",
```
wraymo, line 143: It seems that this PR is based on https://github.com/facebookincubator/velox/pull/13330. Can you also add overlay functions here since they can geometry types as the input

bikramSingh91 (approved):

PR #13082, opened 2025-04-21
build(gpu): Decouple Wave and cuDF

majetideepak:
Also fix VELOX_CUDF_GPU_BATCH_SIZE_ROWS -> velox.cudf.gpu_batch_size_rows

PR #13334, opened 2025-05-13
build(gpu): Re-enable CUDF and WAVE build in CI

czentgr:
PR #13082 separated VELOX_ENABLE_GPU into two different flags. We missed to update the CI to include the build for these parts (CUDF and WAVE) once VELOX_ENABLE_GPU had no more effect.


.github/workflows/linux-build-base.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       GTest_SOURCE: BUNDLED
       simdjson_SOURCE: BUNDLED
       xsimd_SOURCE: BUNDLED
+      cudf_SOURCE: BUNDLED
```
assignUser, line 42: Hm did we neve add cudf to the container? We probably should...

majetideepak (approved):

.github/workflows/linux-build-base.yml
czentgr: Once we are done with the refactor of the setup scripts we can add them there too which then adds them to the runtime, right?

Issue #10397, opened 2024-07-04
An unsupported nested encoding was found.


JkSelf:
### Bug description

When we execute the SQL statement below using Gluten, it raises an exception with the message 'An unsupported nested encoding was found.
```
sql("create table map_table(a map<bigint, string>) using parquet")
sql("insert into table map_table select map(1, 'hello')")
```

### System information

Velox System Info v0.0.2
Commit: 467812fb0a0220fd6b479502dd89b21973a4a3b3
CMake Version: 3.28.3
System: Linux-5.4.0-167-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 9.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 9.4.0
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib/python3.8/dist-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash
Caused by: org.apache.gluten.exception.GlutenException: java.lang.RuntimeException: Exception: VeloxRuntimeError
 Error Source: RUNTIME
 Error Code: INVALID_STATE
 Reason: An unsupported nested encoding was found.
 Retriable: False
 Expression: vec.valueVector() == nullptr || vec.wrappedVector()->isFlatEncoding()
 Context: Operator: TableWrite[2] 2
 Function: exportFlattenedVector
 File: /__w/incubator-gluten/incubator-gluten/ep/build-velox/build/velox_ep/velox/vector/arrow/Bridge.cpp
 Line: 884
 Stack trace:
 # 0  _ZN8facebook5velox7process10StackTraceC1Ei
 # 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
 # 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorEPKcEEvRKNS1_18VeloxCheckFailArgsET0_
 # 3  _ZN8facebook5velox12_GLOBAL__N_121exportFlattenedVectorERKNS0_10BaseVectorERKNS1_9SelectionERK12ArrowOptionsR10ArrowArrayPNS0_6memory10MemoryPoolERNS1_24VeloxToArrowBridgeHolderE
[...]

PR #13338, opened 2025-05-14
fix(parquet): Flatten complex-type vectors when writing

rui-mo:
Parquet write relies on Bridge to convert Velox vector as Arrow array. In 
https://github.com/facebookincubator/velox/commit/98308143adc593fabbbf23f1b9a12c02d462fed6, 'flattenDictionary' and 'flattenConstant' are set as true for Parquet 
write, while Bridge export cannot handle the flattening of dictionary-encoded 
complex-type vectors. To fix this issue, this PR flattens the input vector
before exporting it as Arrow array in Parquet write.

https://github.com/facebookincubator/velox/issues/10397

### DIFF
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -427,10 +427,15 @@ void Writer::write(const VectorPtr& data) {
       data->type()->equivalent(*schema_),
       "The file schema type should be equal with the input rowvector type.");
 
+  VectorPtr exportData = data;
+  if (needFlatten(exportData)) {
+    BaseVector::flattenVector(exportData);
+  }
+
   ArrowArray array;
   ArrowSchema schema;
-  exportToArrow(data, array, generalPool_.get(), options_);
-  exportToArrow(data, schema, options_);
+  exportToArrow(exportData, array, generalPool_.get(), options_);
+  exportToArrow(exportData, schema, options_);
 
   // Convert the arrow schema to Schema and then update the column names based
   // on schema_.
@@ -512,6 +517,22 @@ void Writer::setMemoryReclaimers() {
   // TODO https://github.com/facebookincubator/velox/issues/8190
   pool_->setReclaimer(exec::MemoryReclaimer::create());
   generalPool_->setReclaimer(exec::MemoryReclaimer::create());
+}
+
+bool Writer::needFlatten(const VectorPtr& data) const {
+  auto rowVector = std::dynamic_pointer_cast<RowVector>(data);
+  VELOX_CHECK_NOT_NULL(
+      rowVector, "Arrow export expects a RowVector as input data.");
+
+  const auto& children = rowVector->children();
+  return std::any_of(children.begin(), children.end(), [](const auto& child) {
+    bool isNestedWrapped =
+        (child->encoding() == VectorEncoding::Simple::DICTIONARY ||
+         child->encoding() == VectorEncoding::Simple::CONSTANT) &&
+        child->valueVector() && !child->wrappedVector()->isFlatEncoding();
+    bool isComplex = !child->isScalar();
+    return isNestedWrapped || isComplex;
+  });
 }
 
 std::unique_ptr<dwio::common::Writer> ParquetWriterFactory::createWriter(
--- a/velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
+++ b/velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
@@ -711,7 +711,7 @@ TEST_F(E2EFilterTest, writeDecimalAsInteger) {
   auto rowVector = makeRowVector(
       {makeFlatVector<int64_t>({1, 2}, DECIMAL(8, 2)),
        makeFlatVector<int64_t>({1, 2}, DECIMAL(10, 2)),
-       makeFlatVector<int64_t>({1, 2}, DECIMAL(19, 2))});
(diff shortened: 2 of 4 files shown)


JkSelf:
@rui-mo It appears that there are some failed unit tests associated with this PR.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -415,10 +415,22 @@ void Writer::write(const VectorPtr& data) {
       data->type()->equivalent(*schema_),
       "The file schema type should be equal with the input rowvector type.");
 
+  // If the input data contains complex vector, flatten the input to make it
+  // compatible with Arrow export.
+  VectorPtr exportable = data;
```
majetideepak: nit: do we need this `exportable` variable? Can't we simply use `data`?

rui-mo: The `flattenVector` function requires `VectorPtr&`, while `data` is `const VectorPtr&`, so we need the `exportable` variable. Thanks.

rui-mo:
Hi @majetideepak @Yuhta, would you please take another look? Thanks!

velox/dwio/parquet/writer/Writer.cpp
majetideepak: @rui-mo thanks. If the data is passed by reference, is it safe to modify it by removing the const? Should we be making a copy of the data here?

rui-mo: The `flattenVector` calls `ensureWritable`, which copies-on-write and creates a new vector for `exportable`. The original data stay unmodified. Do you think this makes sense? Thanks.


https://github.com/facebookincubator/velox/blob/a7c95360ccc2632e4793e730096e000f28b9a09f/velox/vector/BaseVector.cpp#L634-L640

wenfang6:
Hi, @rui-mo There is still a scenario where this error occurs.
for example
vanilla spark3.5.4 run this code. 
note: We need to ensure that the data in table t2 are in the same file.
```
create table t_1(id int,name string);
insert overwrite table t_1 values(1,'wukong'),(1,'wukong'),(2,'bajie'),(3,'tanseng'),(2,'bejie'),(4,'shaheshang'),(4,'shaheshang'),(5,'balongma')
create table t_2 as select * from t_1
```
gluten 1.3.0  run this code
```
insert overwrite table g_test_3
select 
    case name
      when 'wukong' then '悟空'
      when 'baajie' then '八戒' end as chinese_name
      from t_2
      where name in ('wukong','baajie')
```

error
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: An unsupported nested encoding was found.
Retriable: False
Expression: vec.valueVector() == nullptr || vec.wrappedVector()->isFlatEncoding()
Context: Operator: TableWrite[3] 2
Function: exportFlattenedVector
/data/wen/github/velox/velox/vector/arrow/Bridge.cpp

rui-mo:
Hi @wenfang6, would you please paste the full stack if you have, and is this PR included in your test? Because I did not find a call to `exportToArrow` except for the `Writer::write`, I'm not sure where this exception was thrown. Please feel free to contact me and provide more details on this issue through slack or email.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -427,10 +427,22 @@ void Writer::write(const VectorPtr& data) {
...
```
majetideepak: @rui-mo just a minor comment on the name here. Thanks.
Can we use `exportData` or `outputData` instead of `exportable`

rui-mo:
Hi @wenfang6, the issue you met might be the case of nested wrapped vector. I made a fix to it and added a test case `dictionary encoded constant vector of scalar type` to cover. Thanks for the pointer!

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -415,10 +415,14 @@ void Writer::write(const VectorPtr& data) {
       data->type()->equivalent(*schema_),
       "The file schema type should be equal with the input rowvector type.");
 
+  // Flatten the input data to make it compatible with Arrow export.
+  VectorPtr flattened = data;
+  BaseVector::flattenVector(flattened);
```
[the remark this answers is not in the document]
rui-mo: Hi @majetideepak, the nested wrapped vector also needs to flattened. I made a fix, and would you like to take another review? Thank you for the help.

velox/dwio/parquet/writer/Writer.cpp
rui-mo: Updated as `exportData`. And you may want to review the new update on https://github.com/facebookincubator/velox/pull/13338#discussion_r2145520483. Thanks!

wenfang6:
> Hi @wenfang6, the issue you met might be the case of nested wrapped vector. I made a fix to it and added a test case `dictionary encoded constant vector of scalar type` to cover. Thanks for the pointer!

Hi @rui-mo ,Thank you very much! I used this PR, and it fixed my issue.

majetideepak (approved):

PingLiuPing:
Thanks a lot. This PR also resolves writing struct data to parquet file.
But I don't see such test case. Maybe you can consider to add a struct column case in a following PR.

rui-mo:
@PingLiuPing I updated this PR to add struct and array test besides the map type, thanks.

rui-mo:
Hi @Yuhta, could you please help import this PR if no further comment? Thanks.

rui-mo:
Thank you for helping review and land this PR!

Issue #13339, opened 2025-05-14
geos 3.10.2 build failure with gcc13


xin-zhang2:
### Problem description

geos 3.10.2 has build issues with gcc13,  which were fixed in 3.10.6.
https://github.com/libgeos/geos/releases/tag/3.10.6

### System information

Velox System Info v0.0.2
Commit: 7de743b488ffbdefb34b8c09ded2fb9ca7ea41c6
CMake Version: 3.28.3
System: Linux-6.8.0-59-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 13.3.0
C Compiler: /usr/bin/cc
C Compiler Version: 13.3.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

### CMake log

```bash
FAILED: CMakeFiles/geos.dir/src/geomgraph/Edge.cpp.o
/usr/bin/c++ -DDLL_EXPORT -DGEOS_INLINE -DNDEBUG -DUSE_UNSTABLE_GEOS_CPP_API -Dgeos_EXPORTS -I/root/deps-download/geos/include -I/root/deps-download/geos/_build/include -I/root/deps-download/geos/src/deps -I/root/deps-download/geos/_build/src/deps -mavx2 -mfma -mavx -mf16c -mlzcnt  -mbmi2 -O3 -DNDEBUG -std=c++11 -fPIC -ffp-contract=off -MD -MT CMakeFiles/geos.dir/src/geomgraph/Edge.cpp.o -MF CMakeFiles/geos.dir/src/geomgraph/Edge.cpp.o.d -o CMakeFiles/geos.dir/src/geomgraph/Edge.cpp.o -c /root/deps-download/geos/src/geomgraph/Edge.cpp
In file included from /root/deps-download/geos/include/geos/geomgraph/Label.h:26,
                 from /root/deps-download/geos/include/geos/geomgraph/GraphComponent.h:28,
                 from /root/deps-download/geos/include/geos/geomgraph/Edge.h:29,
                 from /root/deps-download/geos/src/geomgraph/Edge.cpp:29:
/root/deps-download/geos/include/geos/geomgraph/TopologyLocation.h:99:52: error: ‘uint32_t’ has not been declared
   99 |     bool isEqualOnSide(const TopologyLocation& le, uint32_t locIndex) const;
      |                                                    ^~~~~~~~
[...]

PR #13133, opened 2025-04-23
build: Downgrade geos to 3.10.2

jagill:
Summary:
Upgrading GEOS to 3.13.1 in Meta infra is proving challenging.
This is an experiment to see how easy it is to downgrade the CMake version
to 3.10.2

Differential Revision: D73531553

PR #13340, opened 2025-05-14
fix(geos): Upgrade geos to 3.10.7 to fix geos build issues with gcc13

xin-zhang2:
Fix https://github.com/facebookincubator/velox/issues/13339


xin-zhang2:
@jagill @assignUser
geos was downgraded to 3.10.2 in https://github.com/facebookincubator/velox/pull/13133 and we're now facing build issues with gcc13. Can we upgrade it to 3.10.7 for the fix? If so, I'll make changes to the patches as some are imcompatible with 3.10.7.

kou:
Building with Clang 17 is also failed by this.

Could you remove `CMake/resolve_dependency_modules/geos/geos/geos-build.patch` to fix the following error?

https://github.com/facebookincubator/velox/actions/runs/15018052562/job/42200523222?pr=13340#step:5:1155

```text
[geos] error: patch failed: _build/Debug/_deps/geos-src/capi/geos_ts_c.cpp:2168
[geos] error: _build/Debug/_deps/geos-src/capi/geos_ts_c.cpp: patch does not apply
[geos] error: patch failed: _build/Debug/_deps/geos-src/include/geos/io/ByteOrderDataInStream.h:22
[geos] error: _build/Debug/_deps/geos-src/include/geos/io/ByteOrderDataInStream.h: patch does not apply
[geos] error: patch failed: _build/Debug/_deps/geos-src/tests/unit/math/DDTest.cpp:160
[geos] error: _build/Debug/_deps/geos-src/tests/unit/math/DDTest.cpp: patch does not apply
[geos] FAILED: geos-populate-prefix/src/geos-populate-stamp/geos-populate-patch /Users/runner/work/velox/velox/_build/Debug/_deps/geos-subbuild/geos-populate-prefix/src/geos-populate-stamp/geos-populate-patch 
[geos] cd /Users/runner/work/velox/velox/_build/Debug/_deps/geos-src && git apply /Users/runner/work/velox/velox/CMake/resolve_dependency_modules/geos/geos-cmakelists.patch && git apply /Users/runner/work/velox/velox/CMake/resolve_dependency_modules/geos/geos-build.patch && /usr/local/Cellar/pipx/1.7.1_1/venvs/cmake/lib/python3.13/site-packages/cmake/data/bin/cmake -E touch /Users/runner/work/velox/velox/_build/Debug/_deps/geos-subbuild/geos-populate-prefix/src/geos-populate-stamp/geos-populate-patch
[geos] ninja: build stopped: subcommand failed.
[geos] 
CMake Error at /usr/local/Cellar/pipx/1.7.1_1/venvs/cmake/lib/python3.13/site-packages/cmake/data/share/cmake-3.31/Modules/FetchContent.cmake:1918 (message):
  Build step for geos failed: 1
```

kgpai:
@jagill Can we have 3.10.7 externally and 3.10.2 internally ?

assignUser:
Some error messages changed and were causing test errors, I updated them. This should now be ready to merge @kgpai

xin-zhang2:
Thanks @assignUser.
I resolved some conflicts caused by the recent refactor of the setup scripts.

@kgpai @jagill Could you please take a look?

jagill:
We discussed and decided it was better to upgrade OSS to 3.10.7 and keep the internal version at 3.10.2, scheduling an internal upgrade to 3.10.7 in 2025H2.

This LGTM with the caveat that this is testing against the older error messages (as @assignUser mentioned).  Can we rebase to the latest aned omit the error message changes?  Then this is good to go.

bikramSingh91 (changes requested):
@xin-zhang2 
Since we need to support both versions (like @jagill mentioned), can we remove the error message checks in the tests as they are different between the two versions.

wraymo:
Is there any challenge to upgrade to 3.13.x?

assignUser:
> Is there any challenge to upgrade to 3.13.x?

Yes, that's blocked at meta internally.

assignUser:
I couldn't push to the branch anymore so I have continured this in #13613 with a fix to handle both versions with the error tests

xin-zhang2:
Thanks @assignUser @wraymo 
The current 3.10.2 was downgraded from 3.13.1 in https://github.com/facebookincubator/velox/pull/13133#issue-3015088240.
If two separate versions are decided to be used within Meta and the open source, would it be possible to upgrade the oss version to 3.13.x?

assignUser:
@xin-zhang2  I would suspect that the version bump that big would introduce changes to the API that would need to be handled through ifdefs?

I think if we can keep the actual cpp code the same having two different versions would work. But I assume you are planning to use new functionality? That would require conditional including etc. which gets messy fast.

wraymo:
I wrote serde code for both versions. Most of the code remains the same. The main difference is [CoordinateArraySequence](https://github.com/libgeos/geos/blob/3.10/include/geos/geom/CoordinateArraySequence.h) was removed in 3.13. So as @assignUser said, we’ll likely need to use #ifdefs to handle this version difference.

xin-zhang2:
@assignUser I'm not sure about how this library is used, I only noticed that 3.13.1 was previously used and later downgraded to 3.10.2. 
As @wraymo has pointed out, there are API changes between 3.10.x and 3.13.x, so keeping both versions 3.10.x should avoid introuducing those ifdefs.

assignUser:
Ok, in that case I am also in favour of sticking with 3.10.x! 
@xin-zhang2 do you want to pick the diff to fix the tests from https://github.com/facebookincubator/velox/pull/13613 or should we just continue with that PR? (I don't mind either way)

xin-zhang2:
@assignUser 
Let's continue with #13613 as it has included your latest changes.

xin-zhang2:
Will continue with #13613.

PR #13075, opened 2025-04-19
refactor: Move driver creation into task init from the first next call

xiaoxmeng:
Summary: This is required for task barrier processing.

Differential Revision: D73299498

PR #13341, opened 2025-05-14
feat: Validate the cudf plan

jinchengchenghh:
In Gluten, construct a Task to get the operators after DriverAdapter adapts, then validate the operators if it satisfies the offload condition, now only check it can be fully offloaded to cudf without extra data format conversion cost.
Futher more, we can compare the cost of the plan offload to cudf may with fallback and cpu operators.

The cudf and wave uses operators instead of PlanNode to validate the plan, it benefits Project validation which can be easily validated by FilterProject::exprsAndProjection.

This PR extracts the DriverFactory initialization from Task function `next` to Task constructor under mode Task::ExecutionMode::kSerial, then we can get the operators Driver after task is created.

### DIFF
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -284,6 +284,10 @@ std::shared_ptr<Task> Task::create(
       std::move(onError)));
   task->initTaskPool();
   task->addToTaskList();
+  if (mode == Task::ExecutionMode::kSerial) {
+    task->initDriverFactory();
+  }
+
   return task;
 }
 
@@ -502,6 +506,45 @@ void Task::initTaskPool() {
       fmt::format("task.{}", taskId_.c_str()), createTaskReclaimer());
 }
 
+void Task::initDriverFactory() {
+  VELOX_CHECK_NULL(
+      consumerSupplier_,
+      "Serial execution mode doesn't support delivering results to a "
+      "callback");
+
+  taskStats_.executionStartTimeMs = getCurrentTimeMs();
+  LocalPlanner::plan(
+      planFragment_, nullptr, &driverFactories_, queryCtx_->queryConfig(), 1);
+  exchangeClients_.resize(driverFactories_.size());
+
+  // In Task::next() we always assume ungrouped execution.
+  for (const auto& factory : driverFactories_) {
+    VELOX_CHECK(factory->supportsSerialExecution());
+    numDriversUngrouped_ += factory->numDrivers;
+    numTotalDrivers_ += factory->numTotalDrivers;
+    taskStats_.pipelineStats.emplace_back(
+        factory->inputDriver, factory->outputDriver);
+  }
+
+  // Create drivers.
+  createSplitGroupStateLocked(kUngroupedGroupId);
+  std::vector<std::shared_ptr<Driver>> drivers =
+      createDriversLocked(kUngroupedGroupId);
+  if (pool_->reservedBytes() != 0) {
+    VELOX_FAIL(
+        "Unexpected memory pool allocations during task[{}] driver initialization: {}",
+        taskId_,
+        pool_->treeMemoryUsage());
+  }
+
+  drivers_ = std::move(drivers);
+  driverBlockingStates_.reserve(drivers_.size());
+  for (auto i = 0; i < drivers_.size(); ++i) {
+    driverBlockingStates_.emplace_back(
+        std::make_unique<DriverBlockingState>(drivers_[i].get()));
+  }
+}
+
 velox::memory::MemoryPool* Task::getOrAddNodePool(
     const core::PlanNodeId& planNodeId) {
   if (nodePools_.count(planNodeId) == 1) {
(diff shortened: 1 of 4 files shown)


velox/exec/tests/TaskTest.cpp
```diff
@@ -1156,16 +1156,17 @@ TEST_F(TaskTest, supportSerialExecutionMode) {
...
-  auto task = Task::create(
-      "single.execution.task.0",
-      plan,
-      0,
-      core::QueryCtx::create(),
-      Task::ExecutionMode::kSerial);
 
   // PartitionedOutput does not support serial execution mode, therefore the
   // task doesn't support it either.
-  ASSERT_FALSE(task->supportSerialExecutionMode());
```
jinchengchenghh, line 1303: Do we need to drop supportSerialExecutionMode? After this change, we can not test whether supportSerialExecutionMode

velox/exec/Task.cpp
```diff
@@ -502,6 +505,47 @@ void Task::initTaskPool() {
       fmt::format("task.{}", taskId_.c_str()), createTaskReclaimer());
 }
 
+void Task::initDriverFactory() {
+  VELOX_CHECK_NULL(
+      consumerSupplier_,
+      "Serial execution mode doesn't support delivering results to a "
+      "callback");
+
+  taskStats_.executionStartTimeMs = getCurrentTimeMs();
```
jinchengchenghh, line 514: The task stats executionStartTimeMs moves the Task constructor, not sure for that.

jinchengchenghh:
Hi, @Yuhta Can you help review this PR? Do you agree this design?

jinchengchenghh:
I'm not sure for that, or GPU should validates the PlanNode, and then if validate ok, convert to the operator. In substrait to velox plan, we have the similar discussion, validate then execute or directly convert to planNode and throw exception when conversion failed. In cudf, same question, validate the PlanNode or directly convert to CudfOperator and throws exception if conversion failed.

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -219,4 +222,8 @@ bool cudfIsRegistered() {
   return isCudfRegistered;
 }
 
+bool isCudfOperator(const exec::Operator* op) {
+  return isAnyOf<NvtxHelper>(op);
+}
```
devavret, line 227: You mention in your PR description that you want to 
>check it can be fully offloaded to cudf without extra data format conversion cost

But the cudf conversion operators also inherit from `NvtxHelper`. Did you mean to make a different function that checks that `CudfFromVelox` and `CudfToVelox` don't appear in the task?

jinchengchenghh: Partly yes, and I will check the replaced operators, I allow TableSan and ValueStream node with CudfFromVelox and CudfToVelox. ValueStream accepts shuffle reader input.

velox/exec/Task.cpp
zhztheplayer: This code has similar purpose with https://github.com/facebookincubator/velox/pull/13075?

velox/exec/Task.h
```diff
@@ -747,6 +747,9 @@ class Task : public std::enable_shared_from_this<Task> {
     return cancellationSource_.getToken();
   }
 
+  /// Returns the driver by driverIdx in drivers_.
+  std::shared_ptr<Driver> getDriver(uint32_t driverIdx) const;
+
```
zhztheplayer, line 752: I guess it's a big change to expose the drivers. There's `testingVisitDrivers` can be used if it's just to work around something.

jinchengchenghh: Thanks for this information, I will try.

velox/exec/Task.cpp
jinchengchenghh: Yes, if https://github.com/facebookincubator/velox/pull/13075 can merge, it's ok.

jinchengchenghh:
Hi, @pedroerp would you like to take a look? This is essential for Gluten to validate the plan of one stage and measure if offload the plan to GPU

velox/exec/Task.cpp
```diff
@@ -717,44 +761,8 @@ RowVectorPtr Task::next(ContinueFuture* future) {
...
-    }
-
-    drivers_ = std::move(drivers);
-    driverBlockingStates_.reserve(drivers_.size());
-    for (auto i = 0; i < drivers_.size(); ++i) {
-      driverBlockingStates_.emplace_back(
-          std::make_unique<DriverBlockingState>(drivers_[i].get()));
-    }
+  VELOX_CHECK(!driverFactories_.empty());
+  if (numDriversUnderBarrier_ == 0) {
```
Yuhta, line 817: @xiaoxmeng Can you check if this is correct?

pedroerp:
@jinchengchenghh could you rebase?

@xiaoxmeng could you take a look at the task changes?

xiaoxmeng (commented):
@jinchengchenghh as we discussed, this might break Meta internal streaming service. I need to check if they fix their use case.

jinchengchenghh:
Moved to https://github.com/facebookincubator/velox/pull/14988

PR #13350, opened 2025-05-15
fix: Selective flatmap column reader read offset not updated when all rows filtered out

Yuhta:
Summary:
When we have a filter on flatmap column that filtering all the rows
out, the read offset on the column is not updated.  This was triggering a sanity
check and throwing exception on next batch.

Differential Revision: D74774715


pedroerp (approved):
Thanks for the fix.

PR #13360, opened 2025-05-15
fix: Join condition should be not-exists for TPCH 22

yuandagits:
Summary: TPCH - 22's join should be an anti join that is not null-aware to support "NOT EXISTS" semantics. We need to `nullAware=false`

Differential Revision: D74550644


amitkdutta (approved):
Looks good

PR #13372, opened 2025-05-16
feat: Implement ST_Point ST_X ST_Y

jagill:
Summary: These construct a point and access its members.

Differential Revision: D74898391

### DIFF
--- a/velox/functions/prestosql/GeometryFunctions.h
+++ b/velox/functions/prestosql/GeometryFunctions.h
@@ -16,10 +16,14 @@
 
 #pragma once
 
+#include <geos/geom/Coordinate.h>
 #include <geos/io/WKBReader.h>
 #include <geos/io/WKBWriter.h>
 #include <geos/io/WKTReader.h>
 #include <geos/io/WKTWriter.h>
+#include <geos/util/AssertionFailedException.h>
+#include <geos/util/UnsupportedOperationException.h>
+#include <cmath>
 
 #include <velox/type/StringView.h>
 #include "velox/functions/Macros.h"
@@ -105,6 +109,32 @@ struct StAsBinaryFunction {
           result = outputStream.str();
         },
         "Failed to write WKB");
+    return Status::OK();
+  }
+};
+
+template <typename T>
+struct StPointFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE Status call(
+      out_type<Geometry>& result,
+      const arg_type<double>& x,
+      const arg_type<double>& y) {
+    if (!std::isfinite(x) || !std::isfinite(y)) {
+      return Status::UserError(fmt::format(
+          "ST_Point requires finite coordinates, got x={} y={}", x, y));
+    }
+    GEOS_TRY(
+        {
+          geos::geom::GeometryFactory::Ptr factory =
+              geos::geom::GeometryFactory::create();
+          geos::geom::Point* point =
+              factory->createPoint(geos::geom::Coordinate(x, y));
+          result = geospatial::serializeGeometry(*point);
+          factory->destroyGeometry(point);
+        },
+        "Failed to create point geometry");
     return Status::OK();
   }
 };
@@ -391,6 +421,8 @@ struct StUnionFunction {
   }
 };
 
+// Accessors
+
 template <typename T>
 struct StAreaFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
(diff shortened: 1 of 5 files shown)


jagill:
cc @czentgr

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -109,6 +113,28 @@ struct StAsBinaryFunction {
...
+      out_type<Geometry>& result,
+      const arg_type<double>& x,
+      const arg_type<double>& y) {
+    if (!std::isfinite(x) || !std::isfinite(y)) {
+      return Status::UserError(fmt::format(
+          "ST_Point requires finite coordinates, got x={} y={}", x, y));
+    }
+    geos::geom::GeometryFactory::Ptr factory =
+        geos::geom::GeometryFactory::create();
+    geos::geom::Point* point =
```
bikramSingh91, line 130: just curious, does `factory` own the memory allocated for point? like, would the memory be destroyed along with `factory` or whether its just a heap allocated naked pointer

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -391,4 +417,53 @@ struct StUnionFunction {
...
+      const arg_type<Geometry>& geometry) {
+    std::unique_ptr<geos::geom::Geometry> geosGeometry =
+        geospatial::deserializeGeometry(geometry);
+    if (geosGeometry->getGeometryTypeId() !=
+        geos::geom::GeometryTypeId::GEOS_POINT) {
+      throw Status::UserError(fmt::format(
+          "ST_Y requires a Point geometry, found {}",
+          geosGeometry->getGeometryType()));
+    }
+    if (geosGeometry->isEmpty()) {
```
bikramSingh91, line 460: i noticed a test case with input x coordinate is null, would that exercise this code path? If not, can you please add one?

velox/functions/prestosql/registration/GeometryFunctionsRegistration.cpp
```diff
@@ -55,24 +69,19 @@ void registerOverlayOperations(const std::string& prefix) {
       {{prefix + "ST_Union"}});
 }
 
+void registerAccessors(const std::string& prefix) {
+  registerFunction<StXFunction, double, Geometry>({{prefix + "ST_X"}});
```
bikramSingh91, line 73: can you also add the new functions to documentation at: velox/docs/functions/presto/geospatial.rst

velox/functions/prestosql/GeometryFunctions.h
jagill: Good catch.  I'll add explicit tests for empty points.

velox/functions/prestosql/GeometryFunctions.h
jagill: In 3.10.2, it returns a raw pointer that needs to be explicitly destroyed.  In 3.13.1, it returns an owned `std::unique_ptr`.

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -109,6 +113,28 @@ struct StAsBinaryFunction {
...
+
+  FOLLY_ALWAYS_INLINE Status call(
+      out_type<Geometry>& result,
+      const arg_type<double>& x,
+      const arg_type<double>& y) {
+    if (!std::isfinite(x) || !std::isfinite(y)) {
+      return Status::UserError(fmt::format(
+          "ST_Point requires finite coordinates, got x={} y={}", x, y));
+    }
+    geos::geom::GeometryFactory::Ptr factory =
```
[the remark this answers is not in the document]
jagill: I've wrapped the Point creation, but the x/y coordinate access should be safe.

bikramSingh91 (approved):

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -109,6 +113,32 @@ struct StAsBinaryFunction {
...
+      out_type<Geometry>& result,
+      const arg_type<double>& x,
+      const arg_type<double>& y) {
+    if (!std::isfinite(x) || !std::isfinite(y)) {
+      return Status::UserError(fmt::format(
+          "ST_Point requires finite coordinates, got x={} y={}", x, y));
+    }
+    GEOS_TRY(
+        {
+          geos::geom::GeometryFactory::Ptr factory =
```
wraymo, line 130: I think we can reuse the factory, either making it a local variable here or have a global instance

velox/functions/prestosql/GeometryFunctions.h
wraymo: We are setting the version to 3.13.1. But for compatibility, we can probably do `auto point = std::unique_ptr...()`

velox/functions/prestosql/GeometryFunctions.h
wraymo: Put up the PR here https://github.com/facebookincubator/velox/pull/13525

PR #13376, opened 2025-05-17
feat: Expose maxBucketCount as a hive config

kostasxmeta:
Differential Revision: D74924060

### DIFF
--- a/velox/connectors/hive/HiveConfig.h
+++ b/velox/connectors/hive/HiveConfig.h
@@ -49,6 +49,10 @@ class HiveConfig {
   static constexpr const char* kMaxPartitionsPerWritersSession =
       "max_partitions_per_writers";
 
+  /// Maximum number of buckets allowed to output by the table writers.
+  static constexpr const char* kMaxBucketCount = "hive.max-bucket-count";
+  static constexpr const char* kMaxBucketCountSession = "hive.max_bucket_count";
+
   /// Whether new data can be inserted into an unpartition table.
   /// Velox currently does not support appending data to existing partitions.
   static constexpr const char* kImmutablePartitions =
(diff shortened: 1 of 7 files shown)


velox/connectors/hive/HiveConfig.h
```diff
@@ -49,6 +49,10 @@ class HiveConfig {
   static constexpr const char* kMaxPartitionsPerWritersSession =
       "max_partitions_per_writers";
 
+  /// Maximum number of buckets allowed to output by the table writers.
+  static constexpr const char* kMaxBucketCount = "max_bucket_count";
```
majetideepak, line 53: Config values must use `-`. i/e. `hive.max-bucket-count`.

majetideepak: Please add documentation for this.

kostasxmeta: @majetideepak Addressed this and added documentation -- I see that most configs don't explicitly have `hive.` prefix, so I am following that here (e.g. `max-partitions-per-writers, file-column-names-read-as-lower-case, allow-null-partition-keys, ...` etc). Let me know if that's ok.

kostasxmeta: @majetideepak actually, after talking with @xiaoxmeng , ended up using "hive." prefix anyways. Let me know if there are any issues.

majetideepak (approved):
Thanks, @kostasxmeta

velox/connectors/hive/HiveConfig.h
majetideepak: Yes, we need to fix the other configs as well. Thanks!

PR #13377, opened 2025-05-17
refactor: Correct the name of operatorSupplier

duanmeng:
Some variable types and method return types are actually
OperatorSupplier, we should use the correct name.


xiaoxmeng (approved):
@duanmeng thanks!

zacw7 (approved):

PR #13378, opened 2025-05-19
fix: Handle extreme float-point values in Spark from_json by mapping the value to INF

zhli1142015:
When handling double and float values, the current Velox implementation checks 
whether the values fall within the range [minValue, maxValue]. This behavior is 
inconsistent with Spark's implementation and may lead to the following 
mismatches. This PR fixes this issue.

Spark
```
scala> spark.conf.set("spark.gluten.enabled","false")

scala> spark.sql("""select from_json('{"a": 3.4028235E39}', 'a FLOAT')""").collect
res10: Array[org.apache.spark.sql.Row] = Array([[Infinity]])

scala> spark.sql("""select from_json('{"a": 1.7976931348623158e+309}', 'a DOUBLE')""").collect
res11: Array[org.apache.spark.sql.Row] = Array([[Infinity]])
```
Velox
```
scala> spark.conf.set("spark.gluten.enabled","true")

scala> spark.sql("""select from_json('{"a": 3.4028235E39}', 'a FLOAT')""").collect
25/06/04 17:13:00 WARN IndicatorVectorPool: There are still unreleased native columnar batches during ending the task. Will close them automatically however the batches should be better released manually to minimize memory pressure.
res13: Array[org.apache.spark.sql.Row] = Array([[null]])

scala> spark.sql("""select from_json('{"a": 1.7976931348623158e+309}', 'a DOUBLE')""").collect
25/06/04 17:13:03 WARN IndicatorVectorPool: There are still unreleased native columnar batches during ending the task. Will close them automatically however the batches should be better released manually to minimize memory pressure.
res14: Array[org.apache.spark.sql.Row] = Array([[null]])
```

### DIFF
--- a/velox/functions/sparksql/specialforms/FromJson.cpp
+++ b/velox/functions/sparksql/specialforms/FromJson.cpp
@@ -454,8 +454,28 @@ struct ExtractJsonTypeImpl {
     SIMDJSON_ASSIGN_OR_RAISE(auto type, value.type());
     switch (type) {
       case simdjson::ondemand::json_type::number: {
-        SIMDJSON_ASSIGN_OR_RAISE(auto num, value.get_double());
-        return convertIfInRange<T>(num, writer);
+        auto result = value.get_double();
+        if (result.error() == simdjson::SUCCESS) {
+          auto num = result.value_unsafe();
+          writer.castTo<T>() = num;
+        } else if (result.error() == simdjson::NUMBER_ERROR) {
+          // simdjson parses floating point numbers in the range
+          // [std::numeric_limits<double>::lowest(),
+          // std::numeric_limits<double>::max()], i.e., from approximately
+          // -1.7976e308 to 1.7975e308. Values outside this range
+          // (<= -1e308 or >= 1e308) are rejected and simdjson returns
+          // NUMBER_ERROR. However, our expected behavior is to convert such
+          // extreme values to -INF or +INF, so we add extra logic here to
+          // handle NUMBER_ERROR and perform the conversion.
+          auto castResult = util::Converter<TypeKind::DOUBLE>::tryCast(
+              value.raw_json_token());
+          if (!castResult.hasError()) {
+            writer.castTo<T>() = castResult.value();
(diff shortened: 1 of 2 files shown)


zhli1142015:
cc @rui-mo  and @jinchengchenghh , could you please help review this PR?

velox/functions/sparksql/specialforms/FromJson.cpp
```diff
@@ -349,8 +349,28 @@ struct ExtractJsonTypeImpl {
...
+        if (result.error() == simdjson::SUCCESS) {
+          auto num = result.value_unsafe();
+          writer.castTo<T>() = num;
+        } else if (result.error() == simdjson::NUMBER_ERROR) {
+          // simdjson parses floating point numbers in the range
+          // [std::numeric_limits<double>::lowest(),
+          // std::numeric_limits<double>::max()], i.e., from approximately
+          // -1.7976e308 to 1.7975e308. Values outside this range
+          // (<= -1e308 or >= 1e308) are rejected and simdjson returns
+          // NUMBER_ERROR. However, our expected behavior is to convert such
```
rui-mo, line 467: Is value-out-of-range the only cause of the `simdjson::NUMBER_ERROR`?

rui-mo (commented):
Thanks.

velox/functions/sparksql/specialforms/FromJson.cpp
zhli1142015: Thank you for the reminder. The following cases will all cause NUMBER_ERROR to be reported, and I have added corresponding test cases to ensure they are handled correctly.
```
       R"({"a": 1.1.0})", // Multiple decimal points.
       R"({"a": 1.})", // Missing fraction digits after a decimal point.
       R"({"a": 01})", // Leading zero.
       R"({"a": 1e})", // Missing exponent digits after ‘e’ or ‘E’.
       R"({"a": .e10})", // Missing digits.
       R"({"a": -.})", // Missing digits entirely.
```

rui-mo (approved):

PR #13380, opened 2025-05-19
fix(vector)!: Add check for BaseVector::length_

sdruzkin:
Summary:
Add a non-negative constraint for BaseVector::length_ to prevent it from being set to
negative values in case of int overflows and to expose places causing such overflows.

BREAKING CHANGE:
This change will break code that relies on ability to pass negative value as a length, and
later on updating it to a correct value. Such places should be fixed to not rely on passing
negative vector size.

Differential Revision: D75005693

### DIFF
--- a/velox/vector/FlatVector.h
+++ b/velox/vector/FlatVector.h
@@ -533,6 +533,7 @@ class FlatVector final : public SimpleVector<T> {
   }
 
   void unsafeSetSize(vector_size_t newSize) {
+    VELOX_CHECK_GE(newSize, 0, "Size must be non-negative.");
     this->length_ = newSize;
   }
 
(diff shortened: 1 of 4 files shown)


velox/vector/FlatVector.h
```diff
@@ -533,6 +533,7 @@ class FlatVector final : public SimpleVector<T> {
   }
 
   void unsafeSetSize(vector_size_t newSize) {
+    VELOX_CHECK_GE(newSize, 0, "Size must be non-negative.");
```
sdruzkin, line 536: Not 100% confident about this place because it's an `unsafe` method.

pedroerp (approved):
Thanks @sdruzkin.

My only concern with this is that this in practice breaks backward compatibility, in case users rely on the buggy behavior. Let's keep an eye on open source and internal CI to see if anything gets triggered. Could you add something about this to the PR summary?

sdruzkin:
> We also use a "!" in the title to signal that this PR may break backward compatibility. This should also be a "fix" rather than "misc" ("fix!: Add check ..."):

Thanks, updated the PR title.

sdruzkin:
Fuzzer test seems to be failing with OOM

PR #13385, opened 2025-05-19
feat: Enable prefetch and barrier feature for indexLookupJoin barrier test

juwentus1234:
Summary: Enable prefetch and barrier for indexLookupJoinTest.barrier and disable the pre-materialization if prefetch is not enabled

Differential Revision: D75016998


xiaoxmeng (approved):
@juwentus1234 thanks for the fix.

PR #13386, opened 2025-05-19
feat: Add support for memory pool priority

singcha:
Summary:
Current memory arbitration process doesn't know(hence consider) priority of a query. This results into high priority queries getting killed.
Add support for killing low priority queries in same memory bucket first.
Followup presto PRs will add the functionality of computing the priority and passing it down to workers.

Differential Revision: D74895697


spershin (commented):
Looks good to me. Few comments.

PR #13393, opened 2025-05-20
fix: Minor cleanups for spill partition bits

zhztheplayer:
1. Fix out-of-date documentations for option `spiller_num_partition_bits`
2. Add a meaningful error message when the spill partition bits and hash bits are overlapping in hash table


xiaoxmeng (approved):
@zhztheplayer LGTM. Thanks!

PR #13408, opened 2025-05-20
refactor: Rename copyPreserveEncodings to testingCopyPreserveEncodings

YiChengLee03:
Summary: [T222142639] copyPreserveEncodings are only used in test file with the exception of facebook::velox::exec::Values::initialize. We also replaced the call in Values::initialize() as it was not called elsewhere in the codebase.

Differential Revision: D75084110


zacw7 (approved):
LGTM. Thanks!

PR #13412, opened 2025-05-20
fix: Increase abbreviation length for DateTimeFormatter maxLength function

peterenescu:
Summary:
Increase maximum length of abbreviations in maxLength function. This function estimates the absolute maximum length of a timestamp with timezone, but incorrectly assumes maximum abbreviation length (i.e. PST or -08:00) is 5, when it should actually be 6 (in the case of positive or negative numeric timezones).

Increasing this value to 6 fixes "off-by-one" allocation error found in expression fuzzer runs:

https://www.internalfb.com/servicelab/experiment/5301127199/trial/5301676062/twlogs?logHandle=tsp_atn%2Ftwindtunnel%2Fcogwheel_velox_expression_fuzzer.cogwheel_test.5301676062.a&taskHandle=0&file=stderr

Differential Revision: D75105347


kevinwilfong (approved):
Thanks for the fix!

PR #13415, opened 2025-05-21
build(ci): Fix broken yaml syntax in benchmark workflow

assignUser:
The name is interpreted as invalid yaml, because it looks like a `KEY: <value>`, making the job silently fail. These kind of failures will finally be prevented by the new hooks added in #13361.


kgpai (approved):

majetideepak:
@assignUser It is possible we no longer need this step. Can you verify?

PR #13416, opened 2025-05-21
feat: Register Spark sqrt function

yaooqinn:


velox/functions/sparksql/Arithmetic.h
```diff
@@ -349,6 +349,13 @@ struct Log10Function {
   }
 };
 
+template <typename T>
+struct SqrtFunction {
+  FOLLY_ALWAYS_INLINE void call(double& result, double a) {
+    result = std::sqrt(a);
```
zhli1142015, line 355: We may not need to re-implement this function from scratch; we can simply register the Presto implementation directly in Spark.
https://github.com/facebookincubator/velox/blob/7fbd47bd05e237c03a3803a8f20e672f9999bcd5/velox/functions/prestosql/Arithmetic.h#L391

yaooqinn:
```git
git grep "struct Log2Function"

velox/functions/prestosql/Arithmetic.h:struct Log2Function {
velox/functions/sparksql/Arithmetic.h:struct Log2Function {
```

Hi @zhli1142015, thank you for the suggestion. However, I just follow the current implementation logic, which duplicates Presto/Spark expressions.

zhli1142015: The function above has different semantics in Spark and Presto, so there are two separate implementations. However, your implementation looks exactly the same as Presto’s sqrt function, so I don’t think it’s necessary to duplicate it.

yaooqinn: I see, @zhli1142015, thank you for the hint & sorry for the bad example.

However, I do see functions that are exactly the same being duplicated

```scala
template <typename T>
struct CoshFunction {
  template <typename TInput>
  FOLLY_ALWAYS_INLINE void call(TInput& result, TInput a) {
    result = std::cosh(a);
  }
};

```

zhli1142015 (approved):
LGTM, thanks.
Please also add the corresponding tests in Gluten to ensure that the expression is correctly offloaded.
https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/test/scala/org/apache/gluten/functions/MathFunctionsValidateSuite.scala

yaooqinn:
Thank you @zhli1142015

PR #13418, opened 2025-05-21
fix: Make partially started hash probes reclaimable

tanjialiang:
Summary:
Following state is currently not reclaimable whereas it actually can be reclaimable:
Build has finished. Some probe has started and has transferred the table ownership from join bridge to itself. But some probe has not started and hence still in the wait for build state, not having the hash table.
Change the probe to be reclaimable in this state to cover more corner cases in hash join spilling.

Differential Revision: D75122904


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks for the fix!

PR #13423, opened 2025-05-21
feat(functions): Support named rows as return type

pedroerp:
Summary:
Changing signature binder to carry the name of row children while
binding types, so that users can use functions that return rows with named
children, e.g `row(id bigint)`.

Differential Revision: D75146919

### DIFF
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -397,7 +397,7 @@ class Type;
 using TypePtr = std::shared_ptr<const Type>;
 
 enum class TypeParameterKind {
-  /// Type. For example, element type in the array type.
+  /// Type. For example, element type in the array, map, or row children type.
   kType,
   /// Integer. For example, precision in a decimal type.
   kLongLiteral,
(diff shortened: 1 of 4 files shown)


velox/type/Type.h
```diff
@@ -407,24 +407,32 @@ struct TypeParameter {
...
+  /// null or unset (other than rowFieldName).
   const TypePtr type;
 
   /// Must be set when kind is kLongLiteral. All other properties should be null
   /// or unset.
   const std::optional<int64_t> longLiteral;
 
+  /// If this parameter is a child of another parent row type, it can optionally
+  /// have a name, e.g, "id" for `row(id bigint)`. Only set when kind is kType
+  const std::optional<std::string> rowFieldName;
```
Yuhta, line 419: Hopefully this won't increase memory pressure

Yuhta (approved):

velox/type/Type.h
pedroerp: this is only really used during query/task planning (not during execution), so it shouldn't change things much.

PR #13424, opened 2025-05-21
feat(expression): Add ExpressionOptimizer

pramodsatya:


velox/expression/ExpressionOptimizer.h
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <set>
```
czentgr: Lets move this to the cpp file. It is not needed here.

velox/expression/tests/ExpressionOptimizerTest.cpp
```diff
@@ -0,0 +1,273 @@
...
+#include "velox/expression/ExpressionOptimizer.h"
+#include "velox/expression/ExprCompiler.h"
+#include "velox/expression/VectorFunction.h"
+#include "velox/functions/prestosql/registration/RegistrationFunctions.h"
+#include "velox/parse/Expressions.h"
+#include "velox/parse/ExpressionsParser.h"
+#include "velox/parse/TypeResolver.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
+
+#include <gtest/gtest.h>
```
czentgr: Move this to the top.

velox/expression/tests/ExpressionOptimizerTest.cpp
```diff
@@ -0,0 +1,273 @@
...
+      "33");
+  testExpression(
+      "case when 100 * 10 = 1200 + 10 * 3 + 4 then 1200 else -1000 end",
+      "-1000");
+  testExpression("case when true then 1238 - 4 else 100 end", "1234");
+  testExpression("case when false then 1 else 1230 + 4 end", "1234");
+}
+
+TEST_F(ExpressionOptimizerTest, conjunct) {
+  // AND
```
czentgr: Nit. In other comments (last test) there are periods at the end. So I think we either remove them or add them for consistency reasons. These are not full sentences.
Do we need them when we have the blocks separating the test blocks?

czentgr (commented):
Some initial minor comments.

pramodsatya:
Thanks @czentgr, addressed the comments. @aditi-pandit, @majetideepak, could you please help review this PR?

velox/expression/tests/ExpressionOptimizerTest.cpp
pramodsatya: removed these comments.

czentgr:
@mbasmanova FYI, our work on the expression optimizer.

velox/expression/ExpressionOptimizer.cpp
```diff
@@ -0,0 +1,471 @@
...
+      flattenInput(child, flattenCall, flat);
+    }
+  } else {
+    flat.emplace_back(input);
+  }
+}
+
+void registerExpressionOptimizations(
+    const std::vector<exec::ExpressionRewrite>& customRewrites) {
+  exec::registerExpressionRewrite(
```
aditi-pandit: Add a function registerCommonExpressionRewrite with the basic rewrites. You could also separate them into a single file, so that this file becomes smaller.

velox/expression/ExpressionOptimizer.cpp
```diff
@@ -0,0 +1,471 @@
...
+  }
+}
+
+// Input expression should be of form: IF(condition, valueIfTrue, valueIfFalse).
+core::TypedExprPtr optimizeIfExpression(
+    const core::TypedExprPtr& input,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  auto expr = std::dynamic_pointer_cast<const core::CallTypedExpr>(input);
+  if (expr == nullptr || expr->name() != "if" || expr->inputs().size() != 3) {
```
aditi-pandit: Abstract a constant for "if"

velox/expression/ExpressionOptimizer.cpp
```diff
@@ -0,0 +1,471 @@
...
+}
+
+// Input expression should be of form: SWITCH(condition1, value1, condition2,
+//   value2, ...., defaultValue).
+core::TypedExprPtr optimizeSwitchExpression(
+    const core::TypedExprPtr& input,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  auto expr = std::dynamic_pointer_cast<const core::CallTypedExpr>(input);
+  if (expr == nullptr || expr->name() != "switch") {
```
aditi-pandit: Nit : Abstract a constant for "switch"

velox/expression/ExpressionOptimizer.cpp
```diff
@@ -0,0 +1,471 @@
...
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  // If a case evaluates to true, it will be the new else clause.
+  bool hasOptimizedElseValue = false;
+  for (auto i = 0; i < numInputs - 1; i += 2) {
+    const auto condition = inputs.at(i);
+    const auto foldedCondition = tryConstantFold(condition, queryCtx, pool);
+    const auto value = expr->inputs().at(i + 1);
+    const auto foldedValue = tryConstantFold(value, queryCtx, pool);
+    if (auto constantExpr =
```
aditi-pandit:
```
if (auto constantExpr =
            std::dynamic_pointer_cast<const core::ConstantTypedExpr>(
                foldedCondition)) {
      if (auto constVector = constantExpr->toConstantVector(pool)) {
        if (!constVector->isNullAt(0) &&
            constVector->as<ConstantVector<bool>>()->valueAt(0)) {
```

This seems to be a common code snippet in this logic. Consider making it a function.

velox/expression/ExpressionOptimizer.cpp
```diff
@@ -0,0 +1,471 @@
...
+// to be of type IN(value, arrayVector<literal1, ....., literalN>). When any
+// input literal in IN-list is non-constant, the expression is expected to be
+// of type IN(value, literal1, ....., literalN). The latter case is optimized
+// by this function and the former is handled during constant folding.
+core::TypedExprPtr optimizeInExpression(
+    const core::TypedExprPtr& input,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  auto expr = std::dynamic_pointer_cast<const core::CallTypedExpr>(input);
+  if (expr == nullptr || expr->name() != "in" || expr->inputs().size() < 2) {
```
aditi-pandit: Abstract a constant for "in"

velox/expression/ExpressionOptimizer.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+#pragma once
+
+#include "velox/core/Expressions.h"
+#include "velox/core/QueryCtx.h"
+#include "velox/expression/VectorFunction.h"
+
+namespace facebook::velox::expression {
+
+bool allInputTypesEquivalent(const core::TypedExprPtr& expr);
```
aditi-pandit: You can consider moving this to ITYpedExpr.h, as a member function of ITypedExpr ?

velox/expression/ExpressionOptimizer.h
```diff
@@ -0,0 +1,42 @@
...
+
+void flattenInput(
```
aditi-pandit: Consider moving this to ITypedExpr.h as well

velox/expression/ExprCompiler.cpp
```diff
@@ -586,6 +530,18 @@ std::unordered_set<std::string> collectFlatteningCandidates(
 }
 } // namespace
 
+core::TypedExprPtr rewriteExpression(
```
aditi-pandit: You can consider moving rewriteExpression to ExpressionOptimizer.h and make it the first class method exposed by it.

velox/expression/fuzzer/ExpressionFuzzerVerifier.cpp
```diff
@@ -401,6 +403,15 @@ void ExpressionFuzzerVerifier::go() {
     expressions.push_back(
         std::make_shared<core::FieldAccessTypedExpr>(BIGINT(), "row_number"));
 
+    if (options_.expressionFuzzerOptions.enableExpressionOptimizer) {
+      for (auto& expression : expressions) {
+        auto rewritten =
```
aditi-pandit: This function should actually be the first-class function exposed by ExpressionOptimizer.h

velox/expression/ExprCompiler.h
```diff
@@ -25,6 +25,11 @@ namespace facebook::velox::exec {
 class Expr;
 class ExprSet;
 
+core::TypedExprPtr rewriteExpression(
```
aditi-pandit: It would be great to document the differences between what rewriteExpression and compileExpressions do and are used for.

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,49 @@
...
+namespace facebook::velox::expression {
+
+/// Allows users to register additional expression rewrites.
+void registerExpressionOptimizations(
+    const std::vector<exec::ExpressionRewrite>& customRewrites);
+
+/// Rewrites registered expressions to optimize the expression and
+/// constant folds the result.
+/// The function returns:
+///   - a new constant expression (if constant folded)
```
aditi-pandit: Its more readable if you present the options as 

///   - a new constant expression (if constant folded)
///   - a fail function call expression due to an error during constant
///   evaluation,
///     for example, if constant folding resulted in an error, for example if
///     the expression was 5/0)
///   - the original expression (if it cannot be constant folded, for example,
///   due to presence of a field expression)
///   - a new rewritten expression (if it didn't apply to be constant folded)

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,49 @@
...
+///   - a new rewritten expression (if it didn't apply to be constant folded)
```
aditi-pandit: Its not clear what this means... can you give an example ?

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,49 @@
...
+///   - the original expression (if it cannot be constant folded, for example,
```
aditi-pandit: Do we always just return the original expression if there is a field ? If there is a part of the original expression that can be constant folded, do we not constant fold it ?

velox/expression/VectorFunction.cpp
```diff
@@ -204,4 +204,8 @@ void registerExpressionRewrite(ExpressionRewrite rewrite) {
   expressionRewrites().emplace_back(rewrite);
 }
 
+void unregisterExpressionRewrites() {
```
aditi-pandit: This could be part of ExprOptimizer.h (or ExprCompiler.h) as that is where rewrites are registered.

aditi-pandit:
@mbasmanova : This PR is a follow up from our side for a new first-class optimizeExpressions API from Velox that we can use in Prestissimo while constant folding and expression simplification. Please can you take a look and help review.

@pedroerp

velox/expression/VectorFunction.cpp
pramodsatya: Thanks for the suggestion, since `std::vector<ExpressionRewrite>& expressionRewrites();` is declared in `VectorFunction.h`, it might be better to keep it here, since expression rewrites are primarily used by vector functions?

pramodsatya: Moved expression rewrite related APIs to `ExprRewriter.h/cpp`, since they are used in `ExprCompiler`, `VectorFunction`, and now in `ExprOptimizer`.

velox/expression/ExprCompiler.cpp
pramodsatya: Moved to `ExprRewriter.h`.

aditi-pandit:
@pramodsatya : Can you rebase this PR ? I requested Masha in todays Velox OSS sync to review this code.

pramodsatya:
> @pramodsatya : Can you rebase this PR ? I requested Masha in todays Velox OSS sync to review this code.

Rebased, thanks @aditi-pandit.

mbasmanova:
> Simplify COALESCE to the first non NULL constant input if all inputs seen before were NULL.

I assume we can simply COALESCE to drop any input after first constant non-NULL. Are we already doing that?

mbasmanova:
> IF expression has three inputs: IF(condition, value1, value2)

Let's use 'then' and 'else' for the names of IF inputs.

mbasmanova:
> SWITCH.. If conditionN can be constant folded to true

I assume we can drop inputs after that one. Are we already doing that?

mbasmanova:
@pramodsatya @aditi-pandit Thank you for the detailed summary. 

This is a large PR. Based on the summary, I assume there is code to setup the framework for the optimizations + a number of specific optimizations. Would it be possible to extract the framework + maybe one or two optimizations into one PR and put the rest of the optimizations into follow-up PRs?

mbasmanova:
@pramodsatya @aditi-pandit Would it be possible to update the summary to add a section that describes the design of the optimizer? Thanks.

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+#pragma once
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::expression::utils {
+
+static const std::string kFail = "fail";
+
+using namespace facebook::velox;
```
mbasmanova: Please, no "using namespace" in header files.

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::expression::utils {
+
+static const std::string kFail = "fail";
```
mbasmanova: Where do we define constants for names of other well-known functions? Can we move this there? If there is no such place, then let's make one.

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,47 @@
...
+
+using namespace facebook::velox;
+
+bool isCall(const core::TypedExprPtr& expr, const std::string& name);
```
mbasmanova, line 34: Please, document all public APIs.

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/expression/ExprUtils.h"
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::expression::utils {
+
+using namespace facebook::velox;
```
mbasmanova: Do we need this? I assume not.

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+bool isCall(const core::TypedExprPtr& expr, const std::string& name) {
+  if (auto call = std::dynamic_pointer_cast<const core::CallTypedExpr>(expr)) {
```
mbasmanova: dynamic_cast is expensive; check expr->isCallKind() first, then use `expr->asUnchecked<CallTypedExpr>()`

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+core::TypedExprPtr tryConstantFold(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  try {
+    auto results =
+        exec::tryEvaluateConstantExpression(expr, pool, queryCtx, false);
+    if (nullptr != results) {
```
mbasmanova: if (auto results = ...) {

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+      return std::make_shared<core::ConstantTypedExpr>(results);
+    } else {
```
mbasmanova: drop else after return

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+      // Return the expression unevaluated.
+      return expr;
+    }
+
+  } catch (VeloxUserError& e) {
+    const auto error = std::string(e.what());
```
mbasmanova: unused variable?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    return std::make_shared<core::CallTypedExpr>(
```
mbasmanova: Let's simplify to 

```
std::make_shared<core::CallTypedExpr>(UNKNOWN(), kFail, std::make_shared<core::ConstantTypedExpr>(VARCHAR(), e.what()));
```

by adding one more ctor to CallTypedExpr

```
  template <typename... TypedExprs>
  CallTypedExpr(
      velox::TypePtr type,
      std::string name,
      TypedExprs... inputs)
      : CallTypedExpr(
            std::move(type),
            std::vector<TypedExprPtr>{std::forward<TypedExprs>(inputs)...},
            std::move(name)) {}
```

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+            {std::make_shared<core::ConstantTypedExpr>(VARCHAR(), e.what())}),
+        kFail);
+  }
+}
+
+core::TypedExprPtr constantFold(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  core::TypedExprPtr result;
```
mbasmanova: Move this variable closer to where it is used for the first time.

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+  std::vector<core::TypedExprPtr> foldedInputs;
+  for (auto& input : expr->inputs()) {
+    foldedInputs.push_back(constantFold(input, queryCtx, pool));
+  }
+
+  bool isField = false;
+  if (auto callExpr =
```
mbasmanova: Let's add a new API to ITypedExpr:

```
TypedExprPtr withInputs(std::vector<TypedExprPtr> newInputs)
```

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+    result = concatExpr;
+  } else {
+    result = expr;
+  }
+
+  auto folded = !isField ? tryConstantFold(result, queryCtx, pool) : result;
+  return folded;
+}
+
+// Recursively flattens nested ANDs, ORs or eligible callable expressions into a
```
mbasmanova: This comment goes into the header file.

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+// in: (a OR b) OR (c OR d)
+// out: [a, b, c, d]
+//
+// flattenCall: concat
+// in: (array1, concat(array2, concat(array2, intVal))
+// out: [array1, array2, concat(array2, intVal)]
+void flattenInput(
+    const core::TypedExprPtr& input,
+    const std::string& flattenCall,
+    std::vector<core::TypedExprPtr>& flat) {
```
mbasmanova, line 62: Any particular reason to use an output parameter instead of a return value?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+  if (isCall(input, flattenCall) && allInputTypesEquivalent(input)) {
+    for (auto& child : input->inputs()) {
+      flattenInput(child, flattenCall, flat);
+    }
+  } else {
+    flat.emplace_back(input);
+  }
+}
+
+// Utility method to check eligibility for flattening.
```
mbasmanova: This comment goes into the header file, but it is not very descriptive. Please, consider clarifying.

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+bool allInputTypesEquivalent(const core::TypedExprPtr& expr) {
```
mbasmanova, line 72: Wondering why do we need this logic? Functions that are eligible for flattening always have inputs of the same type? When would this check ever fail?

velox/expression/ExprRewriter.h
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/core/Expressions.h"
+#include "velox/core/QueryCtx.h"
+
+namespace facebook::velox::exec {
```
mbasmanova: Should this be facebook::velox::expression?

velox/expression/ExprRewriter.h
```diff
@@ -0,0 +1,67 @@
...
+
+/// An expression re-writer that takes an expression and returns an equivalent
+/// expression or nullptr if re-write is not possible.
+using ExpressionRewrite = std::function<core::TypedExprPtr(
+    const core::TypedExprPtr,
```
mbasmanova: const&, not just const

velox/expression/ExprRewriter.h
```diff
@@ -0,0 +1,67 @@
...
+std::vector<ExpressionRewrite>& expressionRewrites();
+
+/// Appends a 'rewrite' to 'expressionRewrites'.
+///
+/// The logic that applies re-writes is very simple and assumes that all
+/// rewrites are independent. Re-writes are applied to all expressions starting
+/// at the root and going down the hierarchy. For each expression, rewrites are
+/// applied in the order they were registered. The first rewrite that returns
+/// non-null result terminates the re-write for this particular expression.
+void registerExpressionRewrite(ExpressionRewrite rewrite);
```
mbasmanova: Wondering if it would be helpful for debugging to give each re-write a name.

velox/expression/ExprRewriter.h
```diff
@@ -0,0 +1,67 @@
...
+
+/// Clears all registered expression re-writes.
+void unregisterExpressionRewrites();
```
mbasmanova: Let's move these functions into a class. Say, ExpressionOptimizer or ExpressionRewriteRegistry.

velox/expression/ExprRewriter.cpp
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriter.h"
+
+namespace facebook::velox::exec {
+std::vector<ExpressionRewrite>& expressionRewrites() {
+  static std::vector<ExpressionRewrite> rewrites;
```
mbasmanova: naming: kRewrites

Do we need to synchronize access? I assume these are accessed concurrently from multiple threads.

velox/expression/ExprRewriter.cpp
```diff
@@ -0,0 +1,58 @@
...
+core::TypedExprPtr rewriteExpression(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  for (auto& rewrite : expressionRewrites()) {
+    if (auto rewritten = rewrite(expr, queryCtx, pool)) {
+      return rewritten;
+    }
+  }
+  return expr;
```
mbasmanova: Should we return nullptr to allow the caller to distinguish between the case where expr was rewritten and when it was left unchanged?

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include "velox/core/Expressions.h"
+#include "velox/core/QueryCtx.h"
+#include "velox/expression/ExprRewriter.h"
+#include "velox/expression/VectorFunction.h"
+
+namespace facebook::velox::expression {
+
+/// Allows users to register additional expression rewrites.
+void registerExpressionOptimizations(
```
mbasmanova: Let's use the same terms consistently. "ExpressionOptimization" is the same as "ExpressionRewrite", right? Let's use "rewrite" and not introduce new term.

Why do we need this API when we already have similar APIs elsewhere?

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+constexpr std::string_view kAnd = "and";
+constexpr std::string_view kOr = "or";
+constexpr std::string_view kSwitch = "switch";
+constexpr std::string_view kIn = "in";
+constexpr std::string_view kIf = "if";
+constexpr std::string_view kCoalesce = "coalesce";
+
+} // namespace
+
+// Input expression should be of form: IF(condition, valueIfTrue, valueIfFalse).
```
mbasmanova: > valueIfTrue, valueIfFalse

then, else

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  auto condition = expr->inputs().at(0);
+  auto foldedCondition = utils::tryConstantFold(condition, queryCtx, pool);
+  // The folded expression could be the fail function. In this case,
+  // we don't want to futher analyze the expression and instead return the fail
+  // function expression.
+  if (utils::isCall(foldedCondition, utils::kFail)) {
+    return foldedCondition;
+  }
```
mbasmanova: This logic seems to be shared across the re-writes. Would it make sense to pull this out?

Also, would it be nicer to use a visitor?

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+  if (auto constantExpr =
+          std::dynamic_pointer_cast<const core::ConstantTypedExpr>(
+              foldedCondition)) {
+    if (auto constVector = constantExpr->toConstantVector(pool)) {
```
mbasmanova: Make sure to handle the case when constantExpr stores a Variant. isNull check can be done generically.

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+    return nullptr;
+  }
+
+  const auto inputs = expr->inputs();
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  // If a case evaluates to true, it will be the new else clause.
+  bool hasOptimizedElseValue = false;
+  for (auto i = 0; i < numInputs - 1; i += 2) {
+    const auto condition = inputs.at(i);
```
mbasmanova: const &

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+    const auto foldedCondition =
+        utils::tryConstantFold(condition, queryCtx, pool);
+    if (utils::isCall(foldedCondition, utils::kFail)) {
+      return foldedCondition;
+    }
+
+    const auto value = expr->inputs().at(i + 1);
```
mbasmanova: const &

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+    const auto foldedValue = utils::tryConstantFold(value, queryCtx, pool);
+    if (utils::isCall(foldedValue, utils::kFail)) {
+      return foldedValue;
+    }
+
+    if (auto constantExpr =
+            std::dynamic_pointer_cast<const core::ConstantTypedExpr>(
+                foldedCondition)) {
+      if (auto constVector = constantExpr->toConstantVector(pool)) {
```
mbasmanova: Check for constant value repeats multiple times. Please, refactor to avoid copy-paste.

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,382 @@
...
+        if (!constVector->isNullAt(0) &&
+            constVector->as<ConstantVector<bool>>()->valueAt(0)) {
+          if (optimizedInputs.empty()) {
+            return foldedValue;
+          } else {
```
mbasmanova: drop else after return

aditi-pandit:
> @pramodsatya @aditi-pandit Would it be possible to update the summary to add a section that describes the design of the optimizer? Thanks.

@mbasmanova : Thanks for your review. Have updated the summary with a section on the design of the optimizer.

mbasmanova:
@aditi-pandit Aditi, thank you for sharing the design. This is helpful.

Q: I believe expression rewrites and constant folding need to be interleaved. I imagine we would process the expression tree bottom up. For each node, we would apply re-writes, followed by constant folding. Is this how that works? Would it be possible to add a section to the summary to clarify?

@mbasmanova : Agree that expression rewrites and constant folding are interleaved. 
The Expression tree is being traversed top-down in most logical rewrites. The rewrites are typically invoking constant folding of their input arguments before doing the rewrite at the node. 

The final constant folding is a pure aggressive constant fold that tries to collapse each sub-tree of each node into a constant if possible.

Have updated the summary with these clarifications.

velox/expression/ExprUtils.h
czentgr: For expressions they are defined in the corresponding header. For example, CastExpr.h defines `cast` and `try_cast` - however, the constants are not used because the header is not included in ExprCompiler.cpp.
Similarly with with coalesce, conjunct, etc where the individual header defines the constant for the name. 
So we could make a new file called ExprConstants.h and define all the well known functions there (namespace facebook::velox:expression) and then use it appropriately.

velox/expression/ExprUtils.cpp
czentgr: This was a function moved from ExprCompiler.cpp (but missed to actually remove it from there - it is still there) to be used in multiple places as a utility function. Don't know why the original author used an output parameter instead of a return here. We can refactor.

velox/expression/ExprUtils.cpp
czentgr: This was another function used in ExprCompiler.cpp and moved to ExprUtils.cpp due to the usage in multiple places.
The original use is in `shouldFlatten()` in the the ExprCompiler.cpp. It points out that 
    // Currently only supports the most common case for flattening where all
    // inputs are of the same type.
It is then used again in the actual flattening process. 
Besides AND and OR there are other functions and they are identified by their metadata `supportsFlattening` flag. So perhaps it is a check to make sure a function isn't incorrectly defined as supports flattening by checking the actual input types?

velox/expression/ExprRewriter.h
czentgr: I guess we can decide on that. For example, lots of classes such as expression compiler are in exec. The newer code with the optimizer is in expression. The idea is that the rewriter (like the compiler) is in exec while the expression logic itself is in expression. We can go either way.

velox/expression/ExprRewriter.cpp
czentgr: Multiple threads will have read access when processing expressions. But the vector and functions are registered once on startup.
Things would look different if there are means to register custom optimizations after initialization (registration) of the functions. Maybe with UDFs this can happen if they are not registered at startup.

velox/expression/ExprOptimizer.cpp
czentgr: A visitor like using the ITypedExprVisitor class? Currently, the visit functions returns void - it might need to return an ITypedExpr instead? As in a refactor of how the rewrite is applied to an expression (expression tree?).

I'm trying to understand how this could work. We aren't really traversing an expression tree. We currently get an expression and look at some of its arguments and in some cases do check those for common expression types (and/or and coalesce descend further into the arguments). 
This might also check how compileExpression works So not sure how it would fit together.

velox/expression/ExprRewriter.h
czentgr: The ExpressionOptimizer is supposed to be the external API that can be used via endpoint. It calls the rewrite on a vector of expressions so it is just a thin layer on top, with rewrite being internal only. 
We wanted to keep this separate from the actual implementation of the rewrite and also logically it could different things than just rewrite itself, including that this endpoint would be used by Presto.

aditi-pandit:
@mbasmanova : Would be great to get a round of your comments for this PR.

mbasmanova:
@aditi-pandit I might be confused, but I didn't see replies to comments I left last time. There were some responses to some of the comments, but many we left unanswered. Did I miss the replies / updates?

aditi-pandit:
> @aditi-pandit I might be confused, but I didn't see replies to comments I left last time. There were some responses to some of the comments, but many we left unanswered. Did I miss the replies / updates?

@mbasmanova : Sorry for the confusion. 

Seems like https://github.com/facebookincubator/velox/pull/13424#discussion_r2228162110 was a big design point. So we are keen to carry that discussion further.

The rest of the comments are more doable, so we will resolve them before the next round of review.

mbasmanova:
@aditi-pandit @czentgr Who is the main developer working on this functionality? Should we sync up over VC to come up with a plan to land this?

czentgr:
@mbasmanova @pramodsatya is the main developer with me helping out in addressing review comments (I've done some refactoring as well). We've addressed most/all comments and are about to push up a new version asap.

FYI, there is a sub-PR for the constants: https://github.com/facebookincubator/velox/pull/14493
We were thinking to maybe we need to break up the PR into multiple pieces.

aditi-pandit:
@mbasmanova : Thanks for your comment. Lets sync over VC if possible. 

- There were many questions about the comment for using a visitor interface and it would be great to get some clarity.
- Lets decide a plan on how to land this. There are many sub-parts and it would be great for us to work out the essential bits for each party.

pramodsatya:
Completed in other PRs.

PR #13426, opened 2025-05-21
feat: Add streaming aggregation test inside AggregationReplayerTest

juwentus1234:
Summary:
Add a test case which uses pregrouped (aka streaming aggregation operator) to use AggregationReplayer.

The replayer does go into streaming aggregation operator logic and because the tracing part is done on operator level, it seems the trace will work for streamingaggregation operator.

Differential Revision: D75158168


xiaoxmeng (approved):
@juwentus1234 LGTM. Thanks!

PR #13427, opened 2025-05-21
build: Add flags for NVIDIA Grace

krenzland:
This PR adds the correct flags for NVIDIA Grace CPUs, which have a Neoverse V2 core with additional extensions.
This not only activates cryptography extensions and related codepaths guarded by ifdef clauses but also enables compilers to emit some instructions that can be beneficial for general purpose code.
Validation: This leads to the correct flags for NVIDIA Grace and falls back to Neoverse V2 flags for non-Grace environments.


assignUser (approved):
We don't have matching CI, so going to rely on you to test this.

majetideepak (approved):

PR #13428, opened 2025-05-21
feat: Making TableWriter commitStrategy configurable

kostasxmeta:
Summary:
Currently we hard-code `kNoCommit` as the commit strategy, which directly writes to the final file instead of writing to a temporary file instead before "committing". 

With this option we will be able to set this to avoid inconsistent behavior when running velox in a distributed environment (Pyspark/Velox).

Differential Revision: D75159377

### DIFF
--- a/velox/exec/tests/utils/PlanBuilder.h
+++ b/velox/exec/tests/utils/PlanBuilder.h
@@ -471,6 +471,15 @@ class PlanBuilder {
       return *this;
     }
 
+    /// Specifies commitStrategy for writing to the connector.
+    /// @param commitStrategy The commit strategy to use for the table write
+    /// operation.
+    TableWriterBuilder& commitStrategy(
+        connector::CommitStrategy commitStrategy) {
+      commitStrategy_ = commitStrategy;
+      return *this;
+    }
+
     /// Stop the TableWriterBuilder.
     PlanBuilder& endTableWriter() {
       planBuilder_.planNode_ = build(planBuilder_.nextPlanNodeId());
@@ -502,6 +511,8 @@ class PlanBuilder {
     common::CompressionKind compressionKind_{common::CompressionKind_NONE};
 
     bool ensureFiles_{false};
+    connector::CommitStrategy commitStrategy_{
+        connector::CommitStrategy::kNoCommit};
   };
 
   /// Start a TableWriterBuilder.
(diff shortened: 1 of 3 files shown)


velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -442,6 +442,14 @@ class PlanBuilder {
       return *this;
     }
 
+    /// @param commitStrategy The commit strategy to use for the table write
+    /// operation.
+    TableWriterBuilder& commitStrategy(
```
JkSelf, line 447: nit: `Specifies commitStrategy  for writing to the connector.`

JkSelf (commented):
Do you need to add some unit test? Thanks.

kostasxmeta:
> Do you need to add some unit test? Thanks.

Thanks for the review @JkSelf -- addressed your other comment and added a unit test.

velox/exec/tests/TableWriterTest.cpp
```diff
@@ -189,6 +189,66 @@ TEST_F(BasicTableWriterTestBase, targetFileName) {
       .assertResults(data);
 }
 
+TEST_F(BasicTableWriterTestBase, commitStrategyParameter) {
+  auto data = makeRowVector({makeFlatVector<int64_t>(10, folly::identity)});
+  auto directory = TempDirectoryPath::create();
+
+  // Create a plan with explicit commitStrategy parameter
```
JkSelf, line 196: Can you simplify the code using a lambda function to pass different commitStrategy parameters  to test?

velox/exec/tests/TableWriterTest.cpp
```diff
@@ -189,6 +189,75 @@ TEST_F(BasicTableWriterTestBase, targetFileName) {
       .assertResults(data);
 }
 
+TEST_F(BasicTableWriterTestBase, commitStrategyParameter) {
```
kevinwilfong, line 192: one minor thing, this should probably be in PlanBuilderTest.cpp

PlanBuilder is a test util, and TableWriterTest uses it to generate plans in its tests, but the primary goal of tests in this file is to test the TableWriter Operator, and the PlanBuilder is just a utility to do so.

PlanBuilderTest is where we explicitly test PlanBuilder.

kevinwilfong (approved):
Looks good to me unless @JkSelf has any more comments.

velox/exec/tests/TableWriterTest.cpp
kostasxmeta: Thanks @kevinwilfong, I moved it!

JkSelf:
@kostasxmeta @kevinwilfong Apologies for the delayed response due to the Dragon Boat Festival in China. I have refactored the PlanBuilderTest#commitStrategyParameter to simplify the code https://github.com/facebookincubator/velox/pull/13609. Could you please review it? Thank you.

PR #13445, opened 2025-05-22
fix(json): Bug with duplicate keys leading to incorrect writes

kgpai:
Currently when we cast json to row's if we encounter a duplicate key, we throw, however this throw isnt caught in the json cast row calling function , instead the throw is subsumed by the applyToselectedNoThrow call. This causes a problem because the previous write operation is not cancelled and is still valid, often leading to invalid data being read. This diff fix this bug. This bug is very evident when the key duplicated has a string value  , it might be harder to replicate if the key duplicated is against a complex type , however the fix should work for all types.

Differential Revision: D75251964

### DIFF
--- a/velox/functions/prestosql/types/JsonCastOperator.cpp
+++ b/velox/functions/prestosql/types/JsonCastOperator.cpp
@@ -1229,21 +1229,25 @@ void JsonCastOperator::castFromJson(
     maxSize = std::max(maxSize, input.size());
   });
   paddedInput_.resize(maxSize + simdjson::SIMDJSON_PADDING);
-  context.applyToSelectedNoThrow(rows, [&](auto row) {
-    writer.setOffset(row);
-    if (inputVector->isNullAt(row)) {
-      writer.commitNull();
-      return;
-    }
-    auto& input = inputVector->valueAt(row);
-    memcpy(paddedInput_.data(), input.data(), input.size());
-    simdjson::padded_string_view paddedInput(
-        paddedInput_.data(), input.size(), paddedInput_.size());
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/types/JsonCastOperator.cpp
```diff
@@ -1229,21 +1229,28 @@ void JsonCastOperator::castFromJson(
...
+        memcpy(paddedInput_.data(), input.data(), input.size());
+        simdjson::padded_string_view paddedInput(
+            paddedInput_.data(), input.size(), paddedInput_.size());
+        if (auto error = castFromJsonOneRow<kind>(paddedInput, writer)) {
+          context.setVeloxExceptionError(row, errors_[error]);
+          writer.commitNull();
+        }
+      },
+      [&](auto row) INLINE_LAMBDA {
+        context.setVeloxExceptionError(row, std::current_exception());
```
bikramSingh91, line 1250: error setting is taken care of by applyToSelectedNoThrow so you only need to commit the writer

https://github.com/facebookincubator/velox/blob/main/velox/expression/EvalCtx.h#L306-L309

bikramSingh91 (commented):
can you also mention that this only happens with **string values** because the writer has internal state unlike other types.

on the same note, are there any other value types that can have this same behavior, like complex types (Array , map etc)?

velox/functions/prestosql/types/JsonCastOperator.cpp
kgpai: Fixed.

kevinwilfong (approved):
Thanks for the fix!

PR #13447, opened 2025-05-23
feat(type): Add support for BIGINT noise scale

oliver6782:
Summary:
### Summary
Add another function signature to support BIGINT as a valid variable type for noiseScale in NOISY_COUNT_IF_GAUSSIAN(col, noise_scale) in Velox. Also provide unit test for this feature.

#### Changes
*   Modified `NoisyCountIfGaussianAggregate.cpp` to include BIGINT as a valid noise scale argument type.
*   Added test cases to `NoisyCountIfGaussianAggregationTest.cpp` for BIGINT noise scale support.

Differential Revision: D75258579


amitkdutta (approved):
Thanks @oliver6782

Issue #13452, opened 2025-05-23
Core dump while build use gcc11 with tree-loop-vectorize optimize enabled


lifulong:
### Bug description

we use velox from gluten, velox version is 2025_03_26 branch from https://github.com/oap-project/velox.git

in my test, rewrite the loop line 337 at LzoDecompressor.cpp for prevent loop vectorization optimization, core dump disapears
https://github.com/facebookincubator/velox/blob/main/velox/common/compression/LzoDecompressor.cpp#L337

use `#pragma GCC optimize ("no-tree-loop-vectorize")` before lzoDecompress function at LzoDecompressor.cpp, core dump also disapears

I suspect that vectorization optimization might be accessing a larger memory range than intended, thereby triggering an out-of-bounds memory access.

I'd like to get some suggestions for further troubleshooting, root cause analysis, and fixes to identify and resolve the issue.

### System information

gcc 11

### Relevant logs

```bash

```

lifulong:
https://github.com/facebookincubator/velox/blob/main/velox/common/compression/LzoDecompressor.cpp#L337
```
          while (output < matchOutputLimit) {
              *reinterpret_cast<int64_t*>(output) =
                  *reinterpret_cast<int64_t*>(matchAddress);
              matchAddress += SIZE_OF_LONG;
              output += SIZE_OF_LONG;
            }
```

i have add log, and get some address info while core dump, but i can not reproduce this segment fault error use simple c++ code fragment.
while core dump, 
output-matchAddress is 13
outputLimit-output is 49
matchOutputLimit-output is 1

lifulong:
![Image](https://github.com/user-attachments/assets/845327d9-ee18-4552-b090-3316ed89ff7b)
by add log and analyze the core dump exception stack, i find tree-loop-vectorize optimize for `while loop` above lead to error buffer result, and lead to core dump in subsequent code logic, it is not direct core dump in the `while loop`

PR #13453, opened 2025-05-23
fix core dump issue while do tree-loop-vectorize optimize for lzoDeco…

lifulong:
Fixes: https://github.com/facebookincubator/velox/issues/13452

this is not the final fix, temp commit for show the issue code bock and this change can prevent the core dump

### DIFF
--- a/velox/common/compression/LzoDecompressor.cpp
+++ b/velox/common/compression/LzoDecompressor.cpp
@@ -60,6 +60,8 @@ class MalformedInputException : public dwio::common::ParseError {
 };
 } // namespace
 
+#pragma GCC push_options
+#pragma GCC optimize ("no-tree-loop-vectorize")
 uint64_t lzoDecompress(
     const char* inputAddress,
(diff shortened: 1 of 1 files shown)


velox/common/compression/LzoDecompressor.cpp
```diff
@@ -60,6 +60,8 @@ class MalformedInputException : public dwio::common::ParseError {
 };
 } // namespace
 
+#pragma GCC push_options
+#pragma GCC optimize ("no-tree-loop-vectorize")
```
Yuhta, line 64: Does it happen on both GCC and clang or only one of them?  Since it is likely a compiler bug we want to limit the workaround to only the compilers that have this bug.

lifulong: i have test scan with the issue files, clang is ok, only gcc core dump

lifulong:
@Yuhta  can you help review this pr again

PR #13459, opened 2025-05-23
fix: Incorrect filter result when the range is outside of old type during schema evolution

Yuhta:
Summary:
When we add filter support to table evolution fuzzer, we find a bug in
`BigintRange` that when the type is evolved, and when the filter range is
completely outside of old type, we incorrectly mark the result as passed sometimes.

Differential Revision: D75301749


juwentus1234 (approved):

juwentus1234:
Would this happen to the case of TinyInt and smallInt as well?

PR #13485, opened 2025-05-26
misc: Remove unused dwrf writer reference from HiveDataSink

yingsu00:
Remove unused dwrf writer reference from HiveDataSink


rui-mo (approved):
Thanks.

PR #13488, opened 2025-05-26
build(ci): Fix failing benchmark and fuzzer jobs

assignUser:
- I missed one line in the makefile in #13345 which causes the benchmark job to fail
- We used to build simdjson from source but #12321 changed that, this showed that we added the compile definition to turn on the experimental feature at the wrong place.


majetideepak (approved):

assignUser:
The fuzzer job will fail until this is merged as it pulls from main.

PR #13494, opened 2025-05-27
feat: Support timestamp as hive PartitionId

duxiao1212:
Summary:
https://fburl.com/scuba/presto_queries/c8e9g59t
We need to support queries failing for timestamp partition exec::VectorHasher::typeKindSupportsValueIds( inputType->childAt(channel)->kind()) Unsupported partition type: TIMESTAMP.

Differential Revision: D75156810

### DIFF
--- a/velox/connectors/hive/HivePartitionUtil.cpp
+++ b/velox/connectors/hive/HivePartitionUtil.cpp
@@ -28,6 +28,7 @@ namespace facebook::velox::connector::hive {
       case TypeKind::BIGINT:                                                \
       case TypeKind::VARCHAR:                                               \
       case TypeKind::VARBINARY:                                             \
+      case TypeKind::TIMESTAMP:                                             \
         return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(                          \
             TEMPLATE_FUNC, typeKind, __VA_ARGS__);                          \
       default:                                                              \
@@ -45,6 +46,18 @@ inline std::string makePartitionValueString(T value) {
 template <>
 inline std::string makePartitionValueString(bool value) {
   return value ? "true" : "false";
+}
+
+template <>
+inline std::string makePartitionValueString(Timestamp value) {
+  value.toTimezone(Timestamp::defaultTimezone());
+  TimestampToStringOptions options;
+  options.dateTimeSeparator = ' ';
+  // Set the precision to milliseconds, and enable the skipTrailingZeros match
+  // the timestamp precision and truncation behavior of Presto.
+  options.precision = TimestampPrecision::kMilliseconds;
+  options.skipTrailingZeros = true;
+  return value.toString(options);
 }
 
 template <TypeKind Kind>
--- a/velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
+++ b/velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
@@ -16,6 +16,7 @@
 
 #include "velox/connectors/hive/PartitionIdGenerator.h"
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/type/TimestampConversion.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
 
 #include "gtest/gtest.h"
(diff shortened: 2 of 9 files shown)


kewang1024 (commented):
Can you help also check if writer fuzzer is able to catch the case for timestamp?

velox/connectors/hive/HivePartitionUtil.cpp
```diff
@@ -47,6 +48,15 @@ inline std::string makePartitionValueString(bool value) {
   return value ? "true" : "false";
 }
 
+template <>
+inline std::string makePartitionValueString(Timestamp value) {
+  value.toTimezone(Timestamp::defaultTimezone());
+  TimestampToStringOptions options;
+  options.dateTimeSeparator = ' ';
+  options.precision = TimestampPrecision::kMilliseconds;
```
kewang1024: How do we make sure this is align with Java behavior?

velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
```diff
@@ -236,6 +237,78 @@ TEST_F(PartitionIdGeneratorTest, limitOfPartitionNumber) {
       fmt::format("Exceeded limit of {} distinct partitions.", maxPartitions));
 }
 
+TEST_F(PartitionIdGeneratorTest, timestampPartitionKeyComparasion) {
```
kewang1024, line 240: What does this one test? It seems redundant as `TEST_F(HivePartitionUtilTest, partitionName) {`

velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
```diff
@@ -236,6 +237,78 @@ TEST_F(PartitionIdGeneratorTest, limitOfPartitionNumber) {
...
+      makeFlatVector<Timestamp>({timestampResult.value()}),
+  });
+  raw_vector<uint64_t> testTimeIds;
+  idGenerator.run(input, testTimeIds);
+  EXPECT_EQ(
+      idGenerator.partitionName(testTimeIds[0]),
+      "timestamp_col=2025-01-01 16%3A00%3A00.000");
+}
+
+TEST_F(PartitionIdGeneratorTest, timestampPartitionKey) {
```
kewang1024, line 255: What does this one test?

velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
duxiao1212: There are some overlap but. PartitionIdGenerator test validates the entire workflow of generating partition IDs 
HivePartitionUtil test focuses specifically on extracting and formatting partition key values for timestamp

velox/connectors/hive/tests/PartitionIdGeneratorTest.cpp
duxiao1212: it tests that the same timestamp value will always be assigned to the same partition;

velox/connectors/hive/HivePartitionUtil.cpp
duxiao1212: I tested my change by creating a table with timestamp as a partitioned column using the c++ cluster. I then inserted a row using the cpp cluster and another row using Java, and read the table using both C++ and Java to compare the results. Please suggest other tests I can perform to further validate my change?

duxiao1212:
> Could U also check if writer fuzzer is able to catch the case for timestamp?

Yes, timestamp as partitioned is covered by RowNumberFuzzer::generatePartitionKeys()

Yuhta (approved):

velox/exec/fuzzer/RowNumberFuzzer.cpp
```diff
@@ -68,7 +68,12 @@ class RowNumberFuzzer : public RowNumberFuzzerBase {
 RowNumberFuzzer::RowNumberFuzzer(
     size_t initialSeed,
     std::unique_ptr<test::ReferenceQueryRunner> referenceQueryRunner)
-    : RowNumberFuzzerBase(initialSeed, std::move(referenceQueryRunner)) {}
+    : RowNumberFuzzerBase(initialSeed, std::move(referenceQueryRunner)) {
+  // Set timestamp precision as milliseconds, as timestamp may be used as
+  // paritition key, and presto doesn't supports nanosecond precision.
+  vectorFuzzer_.getMutableOptions().timestampPrecision =
+      fuzzer::FuzzerTimestampPrecision::kMilliSeconds;
```
kagamiori, line 75: FYI, @duanmeng, please let us know if you have concerns about this change.

@duxiao1212, please briefly explain what was wrong when the timestamp precision is kNanoSeconds (i.e., the seg fault).

duxiao1212: Thanks @kagamiori ,
Hi @duanmeng , when a timestamp has nanosecond precision, since presto doesn't support it, the valueId will be invalid(~0UL), and causing the deep recursion in `HashTable::prepareForGroupProbe` that leads to the stack overflow.

duanmeng: @duxiao1212 @kagamiori LGTM

PingLiuPing:
@duxiao1212  What's the valid value range of Timestamp? Thanks.

duxiao1212:
> @duxiao1212 What's the valid value range of Timestamp? Thanks.

I'm not sure if I fully understand your question. Were you asking about the valid range for the velox timestmap? I think that depends on the integer overflow. but anyway, this PR is not related to the Timestamp itself; it's about how to generate a partition Id for a timestamp.

PingLiuPing:
Thanks @duxiao1212 , I'm asking what is the valid supported timestamp range that can be hashed by this PR. The timestamp range is [INT64_MIN/1000 - 1, INT64_MAX/1000], but does any timestamp within this range can be hashed?

velox/functions/sparksql/fuzzer/SparkAggregationFuzzerTest.cpp
```diff
@@ -134,8 +134,10 @@ int main(int argc, char** argv) {
   options.skipFunctions = skipFunctions;
   options.customVerificationFunctions = customVerificationFunctions;
   options.orderableGroupKeys = true;
+  // Set timestamp precision as milliseconds, as timestamp may be used as
+  // paritition key, and presto doesn't supports nanosecond precision
   options.timestampPrecision =
-      facebook::velox::VectorFuzzer::Options::TimestampPrecision::kMicroSeconds;
+      facebook::velox::VectorFuzzer::Options::TimestampPrecision::kMilliSeconds;
```
rui-mo, line 140: Hi @duxiao1212, since the precision of Spark's timestamp is microsecond, we will need to revert this change.
For the comment https://github.com/facebookincubator/velox/pull/13494/files#r2124629633, do you think this fix https://github.com/facebookincubator/velox/pull/14383 can avoid the issue you mentioned? Thanks.

FelixYBW: Just want to share the story. @rui-mo added a check in case fuzzer test is changed in future.

The change made fuzzer passed and got PR merged, which eventually caused 5 people involved and days to fix.

PR #13498, opened 2025-05-27
feat: Add missing function definition

YiChengLee03:
Summary: Added createRowBatch and next method

Differential Revision: D75468157

### DIFF
--- a/velox/dwio/common/Reader.h
+++ b/velox/dwio/common/Reader.h
@@ -47,6 +47,27 @@ class RowReader {
   virtual ~RowReader() = default;
 
   /**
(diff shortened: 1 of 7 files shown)


velox/dwio/common/Reader.h
```diff
@@ -46,6 +46,10 @@ class RowReader {
 
   virtual ~RowReader() = default;
 
+  virtual std::unique_ptr<BaseVector> createRowBatch(uint64_t size) const = 0;
```
zacw7, line 49: Could you pls add some comments what are functions for?

PR #13507, opened 2025-05-27
fix(geos): Add support for bing_tile_parent(bing_tile, integer) signature

kgpai:
Summary: Added  bing_tile_parent(bing_tile, integer)  which is the signature used by Presto when calling bing_tile_parent.

Differential Revision: D75481392


jagill (approved):
LGTM; this will allow us to maintain compatibility with Presto Java.

PR #13509, opened 2025-05-27
fix: Update corrupted file handling from GENERIC_INTERNAL_ERROR to USER ERROR

duxiao1212:
Summary:
For partitioned tables, when some partitions are  in RCFile format, Velox is unable to read them. Since we are deprecating RCFile formats, it's necessary to update the error category for this type of error from internal error to user error. This change will prevent unnecessary panic or mis calculated UER.

link: https://fburl.com/scuba/presto_queries/hvnkoz74

Reviewed By: amitkdutta

Differential Revision: D75488333


amitkdutta (approved):
Thanks @duxiao1212

PR #13524, opened 2025-05-28
feat: Implement ST_IsValid and ST_IsSimple

jagill:
Summary:
These two functions are core functions in testing geometries: if
a geometry is not valid or simple, algorithms are not required to work correctly
or at all.

Differential Revision: D75552740


bikramSingh91 (approved):
The failures look unrelated as I've seen them on other PRs too, looking into them but this should be good to go

PR #13525, opened 2025-05-28
refactor: Refactor ST_Point to reuse GeometryFactory instance

wraymo:
Summary: 
Previously, a new GeometryFactory was created and destroyed for each call. This change moves the factory to a member variable initialized once per function instance.


jagill (approved):
LGTM!  Thanks for the cleanup.  cc @bikramSingh91 for final review and merging.

bikramSingh91 (approved):

Issue #11878, opened 2024-12-16
velox_local_runner_test is flaky


kagamiori:
### Bug description

Example failure: https://github.com/facebookincubator/velox/actions/runs/12333843341/job/34423296887.

### System information

CI

### Relevant logs

```bash
2024-12-15T00:48:38.5214561Z 377/384 Test #382: velox_local_runner_test ..............................................Subprocess aborted***Exception:   9.40 sec
2024-12-15T00:48:38.5215885Z [==========] Running 3 tests from 1 test suite.
2024-12-15T00:48:38.5216613Z [----------] Global test environment set-up.
2024-12-15T00:48:38.5217136Z [----------] 3 tests from LocalRunnerTest
2024-12-15T00:48:38.5217560Z [ RUN      ] LocalRunnerTest.count
2024-12-15T00:48:38.5218254Z I1215 00:48:29.939764 94011 Compression.cpp:627] Initialized zstd compressor with compression level 7
2024-12-15T00:48:38.5218996Z [       OK ] LocalRunnerTest.count (2944 ms)
2024-12-15T00:48:38.5219425Z [ RUN      ] LocalRunnerTest.error
2024-12-15T00:48:38.5221297Z E20241215 00:48:33.879781 94011 Exceptions.h:66] Line: /home/runner/work/velox/velox/velox/velox/common/memory/SharedArbitrator.cpp:308, Function:shutdown, Expression: participants_.size() == 0 (1 vs. 0) Unexpected alive participants on destruction, Source: RUNTIME, ErrorCode: INVALID_STATE
2024-12-15T00:48:38.5223445Z terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
2024-12-15T00:48:38.5224125Z   what():  Exception: VeloxRuntimeError
2024-12-15T00:48:38.5224525Z Error Source: RUNTIME
2024-12-15T00:48:38.5224845Z Error Code: INVALID_STATE
2024-12-15T00:48:38.5225295Z Reason: (1 vs. 0) Unexpected alive participants on destruction
2024-12-15T00:48:38.5225803Z Retriable: False
2024-12-15T00:48:38.5226129Z Expression: participants_.size() == 0
2024-12-15T00:48:38.5226523Z Function: shutdown
2024-12-15T00:48:38.5227093Z File: /home/runner/work/velox/velox/velox/velox/common/memory/SharedArbitrator.cpp
2024-12-15T00:48:38.5227997Z Line: 308
[...]

PR #13535, opened 2025-05-29
feat: Add Spark timestampdiff function

zml1206:
Adds Spark timestampdiff function. Extracts `diffTimestamp` to 
'velox/functions/lib' for reuse between Presto and Spark. Microsecond unit is 
supported for Spark, and a parameter `respectLastDay` is added to specify 
whether the last day of a year-month should be respected.

Spark's implementation:
https://github.com/apache/spark/blob/v3.5.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala#L3242


zml1206:
@rui-mo Thank you.

zml1206:
@jinchengchenghh @zhli1142015 Can you help take a look, thanks.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,59 @@ struct MillisToTimestampFunction {
...
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const arg_type<Varchar>* unitString,
+      const arg_type<Timestamp>* /*timestamp1*/,
+      const arg_type<Timestamp>* /*timestamp2*/) {
+    if (unitString != nullptr) {
+      unit_ = fromDateTimeUnitString(
+          *unitString, /*throwIfInvalid=*/false, /*allowMicro=*/true);
```
zhli1142015: Should we throw an exception if the unit is invalid?

zhli1142015: https://github.com/apache/spark/blob/ebf5a2aef403e14906ecc9a9e56863bad957942e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/DateTimeUtils.scala#L787

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,59 @@ struct MillisToTimestampFunction {
...
+    }
+    const auto unit = unit_.value();
+    if (LIKELY(sessionTimeZone_ != nullptr)) {
+      // sessionTimeZone not null means that the config
+      // adjust_timestamp_to_timezone is on.
+      Timestamp fromZonedTimestamp = timestamp1;
+      fromZonedTimestamp.toTimezone(*sessionTimeZone_);
+
+      Timestamp toZonedTimestamp = timestamp2;
+      if (isTimeUnit(unit)) {
```
zhli1142015: Should we support the micro seconds in `isTimeUnit`？

zhli1142015 (commented):
Thanks

velox/functions/lib/TimeUtils.h
```diff
@@ -162,4 +162,151 @@ Timestamp truncateTimestamp(
...
+          std::chrono::duration_cast<date::days>(toTimepoint - fromTimepoint)
+              .count() /
+          7;
+    }
+    default:
+      break;
+  }
+
+  // Spark support microsecond unit.
+  if (unit == DateTimeUnit::kMicrosecond) {
```
jinchengchenghh: Why not combine it to the switch case?

velox/functions/lib/TimeUtils.h
```diff
@@ -162,4 +162,151 @@ Timestamp truncateTimestamp(
...
+    const std::chrono::time_point<std::chrono::system_clock>
+        fromMicrosecondpoint(std::chrono::microseconds(
```
jinchengchenghh: Do you make sure Presto does not support kMicrosecond? If that, Check the unit in Presto implementation, better aligns with Presto error message.

zml1206: Presto does not support kMicrosecond, allowMicro of fromDateTimeUnitString defaults to false, https://github.com/facebookincubator/velox/blob/aad382dbe8570821e5b55ac62dbcfd1617c9f81b/velox/functions/prestosql/DateTimeFunctions.h#L1164

velox/functions/sparksql/DateTimeFunctions.h
zml1206: Thanks, If the unit is illegal, Spark will report an error during analyze. 
Therefore, it is theoretically unreachable, but it is more suitable to keep it consistent, updated.

velox/functions/sparksql/DateTimeFunctions.h
zml1206: updated.

velox/functions/lib/TimeUtils.h
zml1206: Thank you, updated.

velox/functions/lib/TimeUtils.h
jinchengchenghh: OK, Presto throws in fromDateTimeUnitString as before

velox/functions/lib/TimeUtils.h
zml1206: Combine will cause Presto UT Timestamp(9223372036854775, 0) toMicro overflow, how about separating first? @jinchengchenghh

zml1206:
Addressed all the comments, could you help review again, thanks! @zhli1142015 @jinchengchenghh

velox/functions/lib/TimeUtils.h
jinchengchenghh: Could you describe why it overflows? Presto UT should not test kMicrosecond

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,53 @@ struct MillisToTimestampFunction {
...
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const arg_type<Varchar>* unitString,
+      const arg_type<Timestamp>* /*timestamp1*/,
+      const arg_type<Timestamp>* /*timestamp2*/) {
+    VELOX_USER_CHECK_NOT_NULL(unitString);
+    unit_ = fromDateTimeUnitString(
```
jinchengchenghh, line 965: Should we support the micro seconds in fromDateTimeUnitString? If that, this issue https://github.com/facebookincubator/velox/pull/13535/files#r2168040923 needs to be resolved

zml1206: No problem here, fromDateTimeUnitString supports micro seconds. Presto and spark are distinguished by the parameter allowMicro.

velox/functions/lib/TimeUtils.h
zml1206: Combine needs to be uniformly converted to micro processing, `std::min(fromTimestamp, toTimestamp).toMicros())` will overflow.

jinchengchenghh: Because you move change the toMills to toMicros, keep the toMills in L182, and combine them

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,53 @@ struct MillisToTimestampFunction {
...
+        *unitString, /*throwIfInvalid=*/true, /*allowMicro=*/true);
+    sessionTimeZone_ = getTimeZoneFromConfig(config);
+  }
+
+  FOLLY_ALWAYS_INLINE void call(
```
jinchengchenghh, line 970: Looks like only the initialize function is different, could we use a base function like `UnixTimestampParseFunction`?

zml1206: The unit of spark is a constant, so the processing in call is slightly different. I extracted the main logic into a general function.

velox/functions/lib/TimeUtils.h
```diff
@@ -162,4 +168,178 @@ Timestamp truncateTimestamp(
...
+    const Timestamp& fromTimestamp,
+    const Timestamp& toTimestamp) {
+  // TODO(gaoge): Handle overflow and underflow with 64-bit representation
+  if (fromTimestamp == toTimestamp) {
+    return 0;
+  }
+
+  const int8_t sign = fromTimestamp < toTimestamp ? 1 : -1;
+
+  // Spark support microsecond unit, treat it separately because it needs
```
jinchengchenghh: Comment if convert by .toMicros() for all the unit, may cause overflow, so treat it dependently.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,34 @@ struct MillisToTimestampFunction {
...
+      int64_t& result,
+      const arg_type<Varchar>& /*unitString*/,
+      const arg_type<Timestamp>& timestamp1,
+      const arg_type<Timestamp>& timestamp2) {
+    const auto unit = unit_.value();
```
jinchengchenghh, line 975: I remember in the test, we can not call the unit_.value() in function call, so we always check the unit_.hasValue(), am I right?  @rui-mo So Presto use
```
 const auto unit = unit_.value_or(
        fromDateTimeUnitString(unitString,)
```  here? But the test passed.

Comment (zml1206) [reply velox/functions/sparksql/DateTimeFunctions.h:975]: Different from presto's dateDiff, if there is no value, an exception will be thrown during initialize.

Comment (jinchengchenghh) [reply velox/functions/sparksql/DateTimeFunctions.h:975]: velox function test has some restrictions before, maybe this restriction has been removed.

File: velox/functions/lib/TimeUtils.h [hunk 9]
```diff
@@ -162,4 +168,178 @@ Timestamp truncateTimestamp(
     Timestamp timestamp,
     DateTimeUnit unit,
     const tz::TimeZone* timeZone);
+
+// Returns toTimestamp - fromTimestamp expressed in terms of unit.
+FOLLY_ALWAYS_INLINE int64_t diffTimestamp(
```
jinchengchenghh: Don't include this implementation in TimeUtils.h, use a new file DateTimeImpl.h CC @rui-mo

jinchengchenghh (approved):

velox/docs/functions/spark/datetime.rst
```diff
@@ -271,6 +271,19 @@ These functions support TIMESTAMP and DATE input types.
...
 
+.. spark:function:: timestampdiff(unit, timestamp1, timestamp2) -> bigint
+
+    Returns ``timestamp2`` - ``timestamp1`` expressed in terms of ``unit``.
+    Throws exception if ``unit`` is invalid.
+    ``unit`` is case insensitive and must be one of the following:
+    ``YEAR``, ``QUARTER``, ``MONTH``, ``WEEK``, ``DAY``, ``HOUR``, ``MINUTE``, ``SECOND``,
+    ``MILLISECOND``, ``MICROSECOND``. ::
+
+        SELECT timestampdiff(YEAR, '2020-02-29 10:00:00.500', '2030-02-28 10:00:00.500'); -- 10
```
rui-mo: The output was 9 for this query in my test. Could you please confirm?

```
spark.sql("SELECT timestampdiff(YEAR, '2020-02-29 10:00:00.500', '2030-02-28 10:00:00.500');").show(false)
+---------------------------------------------------------------------+
|timestampdiff(YEAR, 2020-02-29 10:00:00.500, 2030-02-28 10:00:00.500)|
+---------------------------------------------------------------------+
|9                                                                    |
+---------------------------------------------------------------------+
```

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <velox/type/Timestamp.h>
```
rui-mo: nit: use "" to include the header in Velox.

velox/functions/lib/TimeUtils.h
```diff
@@ -31,6 +31,12 @@ inline constexpr int64_t kSecondsInDay = 86'400;
 inline constexpr int64_t kDaysInWeek = 7;
 extern const folly::F14FastMap<std::string, int8_t> kDayOfWeekNames;
 
+FOLLY_ALWAYS_INLINE bool isTimeUnit(const DateTimeUnit unit) {
+  return unit == DateTimeUnit::kMillisecond || unit == DateTimeUnit::kSecond ||
+      unit == DateTimeUnit::kMinute || unit == DateTimeUnit::kHour ||
+      unit == DateTimeUnit::kMicrosecond;
```
rui-mo: nit: perhaps reorder the `kMicrosecond` before `kMillisecond` to improve readability.

velox/functions/lib/TimeUtils.h
```diff
@@ -31,6 +31,12 @@ inline constexpr int64_t kSecondsInDay = 86'400;
 inline constexpr int64_t kDaysInWeek = 7;
 extern const folly::F14FastMap<std::string, int8_t> kDayOfWeekNames;
 
+FOLLY_ALWAYS_INLINE bool isTimeUnit(const DateTimeUnit unit) {
```
rui-mo: It seems this functions corresponds to the `DateTimeUnit` in 'DateTimeFormatter.h', because nanosecond is also time unit but not considered. Perhaps move it near the `DateTimeUnit`.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
...
+          fromZonedTimestamp.getSeconds();
+      toZonedTimestamp = Timestamp(
+          toZonedTimestamp.getSeconds() - offset, toZonedTimestamp.getNanos());
+    } else {
+      toZonedTimestamp.toTimezone(*timeZone);
+    }
+    return diffTimestamp(unit, fromZonedTimestamp, toZonedTimestamp);
+  }
+  return diffTimestamp(unit, fromTimestamp, toTimestamp);
+}
```
rui-mo: nit: add empty line below.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
+/*
```
rui-mo: Consider the name `DateTimeDiff.h` if the file is mainly about timestamp differences.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
...
+#include "velox/functions/lib/TimeUtils.h"
+
+namespace facebook::velox::functions {
+
+// Returns toTimestamp - fromTimestamp expressed in terms of unit.
+FOLLY_ALWAYS_INLINE int64_t diffTimestamp(
+    const DateTimeUnit unit,
+    const Timestamp& fromTimestamp,
+    const Timestamp& toTimestamp) {
+  // TODO(gaoge): Handle overflow and underflow with 64-bit representation
```
rui-mo: nit: not your change, but we may remove the name after TODO to follow the recommended code style.
Also add `.` at the end for comments.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
...
+  if (fromTimestamp == toTimestamp) {
+    return 0;
+  }
+
+  const int8_t sign = fromTimestamp < toTimestamp ? 1 : -1;
+
+  // Spark support microsecond unit, if convert by .toMicros() for all the
+  // unit, may cause overflow, so treat it dependently.
```
rui-mo: Sorry, I did not understand this comment. Would you please clarify a bit?

rui-mo (commented):
Thanks!

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,199 @@
...
+              .count();
+    }
+    case DateTimeUnit::kWeek: {
+      return sign *
+          std::chrono::duration_cast<date::days>(toTimepoint - fromTimepoint)
+              .count() /
+          7;
+    }
+    default:
+      break;
```
zhli1142015: I think you should mark default branch as unreachable, please use VELOX_UNREACHABLE()

velox/docs/functions/spark/datetime.rst
zml1206: Thanks for finding it. Spark is different from Presto. Spark does not respect the last day of a year-month. I will add a parameter to distinguish it.

velox/functions/lib/DateTimeImpl.h
zml1206: This class will add other common methods in the future, such as addTimestamp.

velox/functions/lib/DateTimeImpl.h
zml1206: If all are converted to micro computing, Presto UT Timestamp(9223372036854775, 0) will overflow when do  `std::min(fromTimestamp, toTimestamp).toMicros())`.

velox/functions/lib/DateTimeImpl.h
zml1206: No, here does not include kMonth, kQuarter and kYear.

zhli1142015: All of the other units are handled in the same switch statement—why aren’t kMonth, kQuarter, and kYear handled there as well?

zml1206: This is the existing logic, moved from presto/DateTimeImpl.h. Month, quarter and year do not have fixed conversion ratio. Ex. a month can have 28, 29, 30 or 31 days. A year can have 365 or 366 days.so additional processing is required.

velox/functions/lib/DateTimeImpl.h
rui-mo: How about `DateTimeUtil.h` to show this file contains a collection of reusable date/time utility functions?

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+namespace facebook::velox::functions {
+
+// Returns toTimestamp - fromTimestamp expressed in terms of unit. Unlike
+// Presto, Spark does not respect the last day of a year-month.
+FOLLY_ALWAYS_INLINE int64_t diffTimestamp(
+    const DateTimeUnit unit,
+    const Timestamp& fromTimestamp,
+    const Timestamp& toTimestamp,
+    const bool respectLastDay = true) {
+  // TODO(gaoge): Handle overflow and underflow with 64-bit representation.
```
rui-mo: nit: remove the name `(gaoge)` after TODO.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+#include "velox/external/date/date.h"
+#include "velox/functions/lib/DateTimeFormatter.h"
+#include "velox/functions/lib/DateTimeImpl.h"
+#include "velox/functions/lib/TimeUtils.h"
+#include "velox/type/Timestamp.h"
+
+namespace facebook::velox::functions {
+
+// Returns toTimestamp - fromTimestamp expressed in terms of unit. Unlike
+// Presto, Spark does not respect the last day of a year-month.
```
rui-mo: Use /// for multiple-line comments of public API.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+FOLLY_ALWAYS_INLINE int64_t diffTimestamp(
+    const DateTimeUnit unit,
```
rui-mo: nit: drop 'const' when passing by value.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+    const Timestamp& fromTimestamp,
+    const Timestamp& toTimestamp,
+    const bool respectLastDay = true) {
```
rui-mo: ditto

velox/functions/lib/DateTimeImpl.h
rui-mo: I see, thanks. Comment suggestion:

> // The microsecond unit is handled independently to prevent overflow while converting all timestamps to microseconds for each unit.

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+        (fromMonth == toMonth && fromDay > toDay &&
+         (!respectLastDay || toDay != toLastYearMonthDay)) ||
+        (fromMonth == toMonth && fromDay == toDay &&
+         fromTimeInstantOfDay > toTimeInstantOfDay)) {
+      diff--;
+    }
+    return sign * diff;
+  }
+
+  VELOX_UNREACHABLE("Unsupported datetime unit");
```
rui-mo: > * Use `VELOX_UNREACHABLE()` when a particular branch/block should never be
  executed, such as in a switch statement with an invalid `default:` block.
> * Use `VELOX_NYI()` for features or code paths that are not implemented yet. (throws runtime error)

`VELOX_UNSUPPORTED` throws user error for unsupported features.

Which is the case here?

velox/functions/lib/DateTimeImpl.h
```diff
@@ -0,0 +1,204 @@
...
+#include "velox/functions/lib/DateTimeFormatter.h"
+#include "velox/functions/lib/DateTimeImpl.h"
+#include "velox/functions/lib/TimeUtils.h"
+#include "velox/type/Timestamp.h"
+
+namespace facebook::velox::functions {
+
+// Returns toTimestamp - fromTimestamp expressed in terms of unit. Unlike
+// Presto, Spark does not respect the last day of a year-month.
+FOLLY_ALWAYS_INLINE int64_t diffTimestamp(
```
rui-mo: Comment suggestion:

> /// Returns toTimestamp - fromTimestamp expressed in terms of unit.
/// @param respectLastDay If true, the last day of a year-month is respected.
///   For example, the difference between '2020-01-31' and '2020-02-01' is 1 day.
///   If false, the last day of a year-month is not respected. For example, the
///   difference between '2020-01-31' and '2020-02-01' is 0 day.
///   This is useful for Spark compatibility, as Spark does not respect the last
///   day of a year-month.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -951,4 +951,35 @@ struct MillisToTimestampFunction {
...
+    result =
+        diffTimestamp(unit, timestamp1, timestamp2, sessionTimeZone_, false);
```
rui-mo: nit: `/*respectLastDay=*/false`

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -1287,5 +1287,282 @@ TEST_F(DateTimeFunctionsTest, trunc) {
...
+  EXPECT_EQ(
+      1,
+      timestampDiff(
+          "year",
+          parseTimestamp("2019-02-28 10:00:00.500"),
+          parseTimestamp("2020-02-28 10:00:00.500")));
+
+  /// Test for daylight saving. Daylight saving in US starts at 2021-03-14
+  /// 02:00:00 PST.
+  /// When adjust_timestamp_to_timezone is off, Daylight saving occurs in UTC.
```
rui-mo: nit: use // for comments of private code.

velox/docs/functions/spark/datetime.rst
```diff
@@ -271,6 +271,19 @@ These functions support TIMESTAMP and DATE input types.
 
         SELECT second('2009-07-30 12:58:59'); -- 59
 
+.. spark:function:: timestampdiff(unit, timestamp1, timestamp2) -> bigint
+
+    Returns ``timestamp2`` - ``timestamp1`` expressed in terms of ``unit``.
```
rui-mo: Would you please document how the fraction part is handled? Spark mentions the fraction part is truncated.

https://github.com/apache/spark/blob/v3.5.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/datetimeExpressions.scala#L3210

Please also run the fuzzer test with `--only` to specify this function for 2 minutes to ensure it works, thanks.

velox/functions/lib/DateTimeImpl.h
zml1206: Google's standard seems to be like this, other places in velox also use TODO comments like this. https://zh-google-styleguide.readthedocs.io/en/latest/google-cpp-styleguide/comments.html

rui-mo: > Do not include the author’s username. If required, this can always be retrieved from git blame.

 This is documented in Velox's code style. Not your change, just a minor cleanup suggestion. Thanks.
https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md#comments

velox/functions/lib/DateTimeImpl.h
zml1206: The first one, all possible units have been processed, similar to default: block.

rui-mo: If so, you could just call `VELOX_UNREACHABLE()` without message.

velox/docs/functions/spark/datetime.rst
zml1206: image

rui-mo (approved):
Thanks for iterating!

zml1206:
@kagamiori Could you help import this, thank you.

rui-mo:
The CI failure is irrelevant and tracked in https://github.com/facebookincubator/velox/issues/11878.

rui-mo:
Could you please help merge this PR if no further comment? Thanks! @xiaoxmeng @Yuhta

zhouyuan:
@kagamiori @bikramSingh91 Could you please help to merge this diff?

PR #13542, opened 2025-05-29
feat: Add bulk string buffer method to StringViewBufferholder

joeg:
Summary: add a addOwnedBuffer method to StringViewBufferHolder to allow addition of larger owned buffer. When compared to adding a value at a time via getOwnedValue we can allocate significantly fewer buffers when constructing large vectors ( kInitialStringReservation is 8kb).

Differential Revision: D75619581

### DIFF
--- a/velox/buffer/StringViewBufferHolder.h
+++ b/velox/buffer/StringViewBufferHolder.h
@@ -72,6 +72,17 @@ class StringViewBufferHolder {
     return stringBuffers_;
   }
 
+  /// Add a buffer to the list of buffers. This is used to allow bulk addition
+  /// of values with fewer overall buffers vs adding a value at a time via
+  /// getOwnedValue. The buffer must be allocated on the same underlying memory
+  /// pool.
+  void addOwnedBuffer(BufferPtr&& inBuffer) {
+    VELOX_CHECK(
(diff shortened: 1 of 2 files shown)


velox/buffer/StringViewBufferHolder.h
```diff
@@ -72,6 +72,13 @@ class StringViewBufferHolder {
     return stringBuffers_;
   }
 
+  /// Add a buffer to the list of buffers. This is used to allow bulk addition
+  /// of values with fewer overall buffers vs adding a value at a time via
+  /// getOwnedValue
```
kagamiori: nit: Is `inBuffer` supposed to be owned by `pool_`? If so, could you specify this in the comment and add a `VELOX_CHECK_EQ(inBuffer->pool(), pool_)` in the method?

velox/buffer/StringViewBufferHolder.h
```diff
@@ -72,6 +72,13 @@ class StringViewBufferHolder {
...
+  void addOwnedBuffer(BufferPtr&& inBuffer) {
```
kagamiori, line 79: Can we add a unit test in StringViewBufferHolderTest.cpp?

PR #13544, opened 2025-05-29
fix: Provide SQL generation support for type INTERVAL

peterenescu:
Summary:
There is no tailored support for INTERVAL type SQL generation; it is treated as all other types, so, during casts, PrestoSql outputs `INTERVAL YEAR TO MONTH some_value`, which is invalid.

This diff adds a case for INTERVAL type in PrestoSql's constantSql generation function and properly outputs `INTERVAL some_value YEAR TO MONTH` (and `... DAY TO SECOND`)

Differential Revision: D75644402


kagamiori (approved):
LGTM. Thanks for the fix.

PR #13553, opened 2025-05-30
refactor(Task): Add '#undef' after using 'CLEAR' marco

lingbin:
As a best practice, we should explicitly '#undef' a macro after we 
no longer use it.

No functional changes.


lingbin:
@bikramSingh91  Could you help review this PR?  This is a simple change, but it has been waiting for a review for a very long time.  Already rebased, Thanks.

PR #13559, opened 2025-05-30
Use Faiss utils to compute cosine simarity

zhichenxu-meta:
Differential Revision: D75618542

### DIFF
--- a/velox/functions/prestosql/Arithmetic.h
+++ b/velox/functions/prestosql/Arithmetic.h
@@ -26,6 +26,7 @@
 #include <system_error>
 #include <type_traits>
 
+#include "faiss/utils/distances.h"
 #include "folly/CPortability.h"
 #include "velox/common/base/Doubles.h"
 #include "velox/common/base/Exceptions.h"
@@ -659,27 +660,6 @@ template <typename T>
 struct CosineSimilarityFunctionMap {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
-  double normalize(const null_free_arg_type<Map<Varchar, double>>& map) {
-    double norm = 0.0;
-    for (const auto& [key, value] : map) {
-      norm += (value * value);
-    }
-    return std::sqrt(norm);
-  }
-
-  double dotProduct(
-      const null_free_arg_type<Map<Varchar, double>>& leftMap,
-      const null_free_arg_type<Map<Varchar, double>>& rightMap) {
-    double result = 0.0;
-    for (const auto& [key, value] : leftMap) {
-      auto it = rightMap.find(key);
-      if (it != rightMap.end()) {
-        result += value * it->second;
-      }
-    }
-    return result;
-  }
-
   void callNullFree(
       out_type<double>& result,
       const null_free_arg_type<Map<Varchar, double>>& leftMap,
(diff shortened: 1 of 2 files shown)


majetideepak:
@zhichenxu-meta can you share how faiss improves the computation over the current implementation?

zhichenxu-meta:
> @zhichenxu-meta can you share how faiss improves the computation over the current implementation?

@majetideepak The benefits will come in subsequent changes. The steps are: (1) add faiss as a dependency, get the build scripts to work, (2) introduce different distance functions besides cosine, (3) Introduce faiss index to improve performance for bulk lookup and joins.

majetideepak:
@zhichenxu-meta One of the goals is to keep faiss features and the dependency optional. This change is going to make it core to Velox since prestosql functions are involved. Can we add this feature as optional?
CC: @mbasmanova, @assignUser, @kgpai

assignUser:
+1 on that, with this change faiss becomes a core dependency and I don't think that's a change we should make this easily.

assignUser (changes requested):
Just to formalize: I'm against merging this without further discussion about dependency bloat and what core velox should be.

zhichenxu-meta:
I agree that making Faiss optional, at least initially, is a sensible approach. I'm updating this PR to conditionally enable Faiss-based distance functions using the VELOX_ENABLE_FAISS flag.

As the project evolves, we'll continue to refine and optimize the logic as needed. We can draw inspiration from how geographical search was introduced into Velox, and adapt similar strategies for our implementation.

zhichenxu-meta:
@assignUser @majetideepak @amitkdutta Please see my comment above, and let me know your suggestions, and thank you!

amitkdutta:
> Just to formalize: I'm against merging this without further discussion about dependency bloat and what core velox should be.

@assignUser @majetideepak We are going to build a lot machine learning functionalities in velox that's need faiss. I am not sure whats stopping us not to extend velox with new possibilities. What do you mean by core velox?

zhichenxu-meta:
> +1 on that, with this change faiss becomes a core dependency and I don't think that's a change we should make this easily.

@assignUser @amitkdutta @FYI, I've made the Faiss-based distance functions conditional on the VELOX_ENABLE_FAISS flag.

> I agree that making Faiss optional, at least initially, is a sensible approach. I'm updating this PR to conditionally enable Faiss-based distance functions using the VELOX_ENABLE_FAISS flag.
> 
> As the project evolves, we'll continue to refine and optimize the logic as needed. We can draw inspiration from how geographical search was introduced into Velox, and adapt similar strategies for our implementation.

@majetideepak @assignUser @amitkdutta 
FYI, I've made the Faiss-based distance functions conditional on the VELOX_ENABLE_FAISS flag.  Also, will improve the logic in subsequent diffs (limit number of changes in on PR).

velox/functions/prestosql/CMakeLists.txt
```diff
@@ -63,18 +63,24 @@ velox_add_library(
...
-  velox_functions_json
-  velox_functions_lib
-  velox_expression
-  velox_external_md5
-  velox_type_tz
-  velox_presto_types
-  velox_functions_util
-  Folly::folly
-  stemmer::stemmer)
+set(PRESTOSQL_IMPL_LIBS
```
majetideepak: velox_link_libraries() will append to the list. You don't need to do this.

velox/functions/prestosql/CMakeLists.txt
```diff
@@ -63,8 +63,7 @@ velox_add_library(
   Zip.cpp
   ZipWith.cpp)
 
-velox_link_libraries(
-  velox_functions_prestosql_impl
+velox_link_libraries(velox_functions_prestosql_impl
```
zhichenxu-meta: @majetideepak changed to let velox_link_libraries to append the faiss lib. Thank you for the review.  Will fix formatting in subsequent changes.

velox/functions/prestosql/CMakeLists.txt
zhichenxu-meta: done, and thank you

assignUser:
It seems you are still very activly working on this. I will convert this to a draft as otherwise each push creates a notification for me. Let me know when this is review ready.

zhichenxu-meta:
I have done making the structural changes to make Faiss dependency optional. You are welcome to review the PR now.  I will make c++ logic improvements in future PRs.
________________________________

velox/docs/functions/presto/math.rst
```diff
@@ -51,6 +51,19 @@ Mathematical Functions
 
         SELECT cosine_similarity(ARRAY[], ARRAY[]); -- NaN
 
+.. function:: l2_squared(array(double), array(double)) -> double
```
amitkdutta: Lets update the commit message with l2 distance. May be "Use faiss utils to compute distance functions"

amitkdutta:
Thanks so much @zhichenxu-meta for iterating over it. Looks good to me. @assignUser let us know if you have any issue on the PR.

amitkdutta (approved):

amitkdutta (changes requested):
Looks like fuzzer failure
https://github.com/facebookincubator/velox/actions/runs/15714350527/job/44280219823?pr=13559

zhichenxu-meta:
@assignUser @majetideepak Could either of you please take another look at PR, and let me know if you have any suggestions, and thank you!

PR #13563, opened 2025-05-30
misc: Log failing plan at error level in AggregationFuzzer/WindowFuzzer so it's visible in CI

kevinwilfong:
Summary:
In CI only logs at ERROR level or above are written to stderr, and only logs to stderr are displayed directly on 
the results page.  To get the rest of the logs we have to download the archive.

Currently we log all plans that we run at the INFO level, so when we see an error, e.g. a mismatch in results, 
it's difficult to figure out what's failing without downloading the archive.

We can't log all plans to stderr without exceeding the UIs limit, but we can at least log the failing plan to
speed up debugging.

Differential Revision: D75698709


amitkdutta (approved):
Thanks @kevinwilfong

PR #13569, opened 2025-05-30
feat: Add ST_XMin, ST_XMax, ST_YMin, and ST_YMax

jkhaliqi:
Get the max or minimum x or y value of the geometry’s bounding box.


aditi-pandit (commented):
Thanks @jkhaliqi. Have few comments. Will leave the main review to James.

aditi-pandit:
@wraymo... Not able to add you as a reviewer. Please can you help with this review.

velox/docs/functions/presto/geospatial.rst
```diff
@@ -184,12 +184,32 @@ Accessors
     ``null`` if the geometry is empty.  Raises an error if the geometry is
     not a Point and not empty.
 
+.. function:: ST_XMin(geometry: Geometry) -> x: double
```
jagill: nit: Move these after ST_Y.  The reasoning is that people tend to want ST_X and ST_Y in the same operation, and `ST_{X,Y}{Min,Max}` in the same operation.

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -521,4 +556,38 @@ struct StYFunction {
   }
 };
 
+template <typename T>
+struct StYMinFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE Status
+  call(out_type<double>& result, const arg_type<Geometry>& geometry) {
+    std::unique_ptr<geos::geom::Geometry> geosGeometry =
```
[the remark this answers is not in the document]
jagill: Agreed, particularly when #12771 lands, we'll deserialize the envelope in a different (and far more efficient) way.  Let's make a function in GeometrySerde that deserializes and returns a `geos::geom::Envelope` using this logic.  Then after #12771 lands, we can just modify that one function.

velox/functions/prestosql/tests/GeometryFunctionsTest.cpp
```diff
@@ -1360,3 +1360,81 @@ TEST_F(GeometryFunctionsTest, testSimplifyGeometry) {
...
+      3.0);
+  assertPointMin(
+      "GEOMETRYCOLLECTION (POINT (5 1), LINESTRING (3 4, 4 4))", 3.0, 1.0);
+  assertPointMin(std::nullopt, std::nullopt, std::nullopt);
+  std::optional<double> nullX =
+      evaluateOnce<double>("ST_XMin(ST_GeometryFromText('POLYGON EMPTY'))");
+  EXPECT_FALSE(nullX.has_value());
+  std::optional<double> nullY =
+      evaluateOnce<double>("ST_YMin(ST_GeometryFromText('POLYGON EMPTY'))");
+  EXPECT_FALSE(nullY.has_value());
```
jagill:
```suggestion
  assertPointMin("POLYGON EMPTY", std::nullopt, std::nullopt);
```

velox/functions/prestosql/tests/GeometryFunctionsTest.cpp
```diff
@@ -1360,3 +1360,81 @@ TEST_F(GeometryFunctionsTest, testSimplifyGeometry) {
...
+      10.0);
+  assertPointMax(
+      "GEOMETRYCOLLECTION (POINT (5 1), LINESTRING (3 4, 4 4))", 5.0, 4.0);
+  assertPointMax(std::nullopt, std::nullopt, std::nullopt);
+  std::optional<double> nullX =
+      evaluateOnce<double>("ST_XMax(ST_GeometryFromText('POLYGON EMPTY'))");
+  EXPECT_FALSE(nullX.has_value());
+  std::optional<double> nullY =
+      evaluateOnce<double>("ST_YMax(ST_GeometryFromText('POLYGON EMPTY'))");
+  EXPECT_FALSE(nullY.has_value());
```
jagill:
```suggestion
  assertPointMax("POLYGON EMPTY", std::nullopt, std::nullopt);
```

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -542,6 +557,58 @@ struct StYFunction {
...
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE Status
+  call(out_type<double>& result, const arg_type<Geometry>& geometry) {
+    return computeFromEnvelope(
+        result, geometry, [](const geos::geom::Envelope* env) {
+          return env->getMinX();
+        });
+  }
+};
```
jagill: This this can return `null`, we actually need to:

1. Change the return type to `bool`, `true` indicating non-`null` (valid) and `false` indicating `null`.
2. Use VELOX_USER_FAIL or VELOX_FAIL instead of returning status.

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -542,6 +557,58 @@ struct StYFunction {
...
+template <typename T>
+struct StYMinFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE Status
+  call(out_type<double>& result, const arg_type<Geometry>& geometry) {
+    return computeFromEnvelope(
+        result, geometry, [](const geos::geom::Envelope* env) {
+          return env->getMinY();
+        });
```
jagill: I actually find it easier to read like 
```suggestion
    // maybe std::unique_ptr<...::Envelope> instead
    geos::geom::Envelope envelope = geospatial::readEnvelope(geometry); 
    return envelope->getMinY();
```

We'll use envelopes in other computations later, so this gives us more optionality.

jagill (changes requested):
This looks nice!  Let's just get it into a form that we can build off of for other UDFs.

velox/functions/prestosql/geospatial/GeometrySerde.cpp
```diff
@@ -38,4 +38,14 @@ std::string serializeGeometry(const geos::geom::Geometry& geosGeometry) {
...
 
+const std::optional<geos::geom::Envelope> getEnvelopeFromGeometry(
+    const StringView& geometry) {
+  std::unique_ptr<geos::geom::Geometry> geosGeometry =
+      geospatial::deserializeGeometry(geometry);
+  if (geosGeometry->isEmpty()) {
+    return std::nullopt;
+  }
+  return *geosGeometry->getEnvelopeInternal();
+}
```
jagill: Actually, empty geometries will return an empty envelope, which will behave nicely in many cases.  For example, intersection tests will be correct without additional code.

```suggestion
const std::unique_ptr<geos::geom::Envelope> getEnvelopeFromGeometry(
    const StringView& geometry) {
  std::unique_ptr<geos::geom::Geometry> geosGeometry =
      geospatial::deserializeGeometry(geometry);
  return std::unique_ptr<geos::geom::Envelope>(geosGeometry->getEnvelopeInternal());
}
```

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -542,6 +543,66 @@ struct StYFunction {
...
+};
+
+template <typename T>
+struct StYMaxFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  bool call(out_type<double>& result, const arg_type<Geometry>& geometry) {
+    const std::optional<geos::geom::Envelope> env =
+        geospatial::getEnvelopeFromGeometry(geometry);
+    if (!env.has_value()) {
```
jagill: With the change in getEnvelopeFromGeometry below...

```suggestion
    const std::unique_ptr<geos::geom::Envelope> env =
        geospatial::getEnvelopeFromGeometry(geometry);
    if (env.isNull()) {
```

PR #13572, opened 2025-05-30
Simplify Split

Magoja:
Summary:
`Split` is a POD, and unnecessarily have constructors and all assignment operators.

We can simply delete them all to do the same thing.

Differential Revision: D75712160


Magoja:
Too much of changes required for this.

PR #13575, opened 2025-05-30
Add Faiss library as an optional dependency

zhichenxu-meta:
Summary: This PR introduces FAISS as an optional dependency, making the first step to support Faiss vector search in velox.

Differential Revision: D75707431


xiaoxmeng (approved):

CMakeLists.txt
```diff
@@ -160,6 +160,7 @@ option(
   OFF)
 option(VELOX_SIMDJSON_SKIPUTF8VALIDATION
        "Skip simdjson utf8 validation in JSON parsing" OFF)
+option(VELOX_ENABLE_FAISS "Build faiss vector search support" OFF)
```
kgpai, line 163: Lets enable this by default.

scripts/setup-common.sh
```diff
@@ -232,6 +233,95 @@ function install_geos {
   fi
 }
 
+function install_faiss_deps {
+  local uname_s=$(uname -s)
+  echo "Detected OS: $uname_s"
+
+  if [[ "$uname_s" == "Darwin" ]]; then
+    brew list openblas &>/dev/null || brew install openblas
```
kgpai: Its probably not best to mix arch's here. Can we move the os specific stuff to their respective scripts. 
Another thing to do is also probably add a flag for them, since FAISS itself depends on dependencies that arent used anywhere else.

zhichenxu-meta:
> Sorry for the terse ❌ but there is a lot going on here that we need to reconsider. I will follow up tomorrow.

Thank you and look forward to your suggestions

.github/workflows/linux-build-base.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       GTest_SOURCE: BUNDLED
       cudf_SOURCE: BUNDLED
       CUDA_VERSION: "12.8"
+      faiss_SOURCE: BUNDLED
+      FAISS_VERSION: "1.11.0"
```
assignUser: This should be/is set in `setup-versions.sh`
```suggestion
```

.github/workflows/linux-build-base.yml
```diff
@@ -98,6 +106,10 @@ jobs:
             "-DVELOX_ENABLE_WAVE=ON"
             "-DVELOX_MONO_LIBRARY=ON"
             "-DVELOX_BUILD_SHARED=ON"
+            "-DFAISS_ENABLE_PYTHON=OFF"
+            "-DFAISS_ENABLE_GPU=OFF"
+            "-DFAISS_ENABLE_GPU_TESTS=OF"
+            "-DFAISS_ENABLE_BENCHMARKS=OFF"
```
assignUser: This should be set in the resolve_dependency module or setup script.
```suggestion
```

.github/workflows/linux-build-base.yml
```diff
@@ -128,6 +140,13 @@ jobs:
     env:
       CCACHE_DIR: "${{ github.workspace }}/ccache"
       USE_CLANG: "${{ inputs.use-clang && 'true' || 'false' }}"
+    run: |
+      EXTRA_CMAKE_FLAGS=(
+        "-DFAISS_ENABLE_PYTHON=OFF"
+        "-DFAISS_ENABLE_GPU=OFF"
+        "-DFAISS_ENABLE_GPU_TESTS=OF"
+        "-DFAISS_ENABLE_BENCHMARKS=OFF"
+      )
```
assignUser: Same here but this also completely breaks the[ workflow](https://github.com/facebookincubator/velox/actions/runs/15447826834)

.github/workflows/macos.yml
```diff
@@ -98,7 +98,10 @@ jobs:
                 -DVELOX_ENABLE_PARQUET=ON \
                 -DVELOX_MONO_LIBRARY=ON \
                 -DVELOX_BUILD_SHARED=ON \
-                -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+                -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
+                -DFAISS_ENABLE_GPU=OFF \
+                -DFAISS_ENABLE_GPU_TESTS=OFF \
+                -DFAISS_ENABLE_PYTHON=OFF
```
assignUser: Again, shouldn't be necessary.
```suggestion
                -DCMAKE_BUILD_TYPE=$BUILD_TYPE
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+      CACHE STRING "FAISS source archive SHA256 checksum")
+
+  # Builds the URL to download the FAISS source tarball for the specified
+  # version.
+  set(VELOX_FAISS_SOURCE_URL
+      "https://github.com/facebookresearch/faiss/archive/refs/tags/v${VELOX_FAISS_BUILD_VERSION}.tar.gz"
+      CACHE STRING "FAISS source URL to download")
+endfunction()
+
+configure_faiss()
```
assignUser: I don't understand what the use for the function here is. It's called once, directly under the function. In addition functions have their own variable scope which explains why you resorted to setting everything as cache variables.

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+
+# Prevents this CMake file from being included more than once in the same build,
+# avoiding duplicate definitions.
+include_guard(GLOBAL)
+
+function(find_and_requre_openmp)
+  # Enable OpenMP for FAISS
+  set(FAISS_ENABLE_OPENMP
+      ON
+      PARENT_SCOPE)
```
assignUser: This variable doesn't even [exist](https://github.com/search?q=repo%3Afacebookresearch%2Ffaiss%20FAISS_ENABLE_OPENMP&type=code).

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+            "HOMEBREW_PREFIX environment variable is not set and Homebrew is not installed! Please install Homebrew or set HOMEBREW_PREFIX manually."
+        )
+      endif()
+    endif()
+
+    # Set architecture-specific flags if needed
+    set(ARCH_FLAGS "")
+    if(DEFINED ENV{CPU_TARGET} AND "$ENV{CPU_TARGET}" STREQUAL "arm64")
+      message(STATUS "Detected Apple Silicon (arm64) via CPU_TARGET")
+      set(ARCH_FLAGS "-mcpu=apple-m1+crc")
```
assignUser: This is handled through the main cml

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Prevents this CMake file from being included more than once in the same build,
+# avoiding duplicate definitions.
+include_guard(GLOBAL)
+
+function(find_and_requre_openmp)
```
assignUser: Both this function and `find_and_require_blas` are not needed. Let faiss handle their own dependencies, there are FindModules [for](https://cmake.org/cmake/help/latest/module/FindOpenMP.html) [both](https://cmake.org/cmake/help/latest/module/FindBLAS.html) that handle this in a standardized manner which faiss [uses](https://github.com/facebookresearch/faiss/blob/043534c4c2b80d07af8f0d9ede66f22a4fc041f0/faiss/CMakeLists.txt#L380). 

The only thing we might want to is add the relvant paths hints but tbh that should be handled by the user through the cmake prefix path.

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+        git
+        apply
+        "${_patch}")
+    endif()
+  endforeach()
+  # Return the command as a list to the caller
+  set(${out_var}
+      "${_faiss_patch_command}"
+      PARENT_SCOPE)
+endfunction()
```
assignUser: I assume you started out by copying geos.cmake (which is fine!). We are not applying any patches, just delete that part... no need for an elaborate function that applies patches if they exists. We have full control about that.

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+      CACHE STRING "FAISS source URL to download")
+endfunction()
+
+configure_faiss()
+# Resolve the dependency URL for FAISS
+velox_resolve_dependency_url(FAISS)
+
+generate_patch_command(FAISS_PATCH_COMMAND)
+find_and_requre_openmp()
+find_and_require_blas()
```
assignUser:
```suggestion
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+# Tells CMake to fetch FAISS from the specified URL, verify its checksum, and
+# apply any patches if needed.
+FetchContent_Declare(
+  faiss
+  URL ${VELOX_FAISS_SOURCE_URL}
+  URL_HASH ${VELOX_FAISS_BUILD_SHA256_CHECKSUM}
+  PATCH_COMMAND ${FAISS_PATCH_COMMAND})
+
+# Append FAISS cmake module path
+list(APPEND CMAKE_MODULE_PATH "${faiss_SOURCE_DIR}/cmake")
```
assignUser: No, we don't need that, faiss will do that itself.
```suggestion
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+# Set build options
+set(PREVIOUS_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+set(BUILD_SHARED_LIBS ${VELOX_BUILD_SHARED})
+set(PREVIOUS_BUILD_TYPE ${CMAKE_BUILD_TYPE})
+set(CMAKE_BUILD_TYPE Release)
+set(PREVIOUS_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -Wno-nonnull ")
+
+# Disables developer-specific warnings in FAISS.
+set(FAISS_BUILD_DEVELOPER OFF)
```
assignUser: This was a geos specific variable...
```suggestion
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+
+# Add specific compiler flags for GNU
+if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -Wno-dangling-pointer")
+endif()
```
assignUser: From geos, not needed her.
```suggestion
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+
+# Make FAISS available
+FetchContent_MakeAvailable(faiss)
+# Link BLAS to FAISS target
+if(TARGET faiss)
+  target_link_libraries(faiss PRIVATE ${BLAS_LIBRARIES} OpenMP::OpenMP_CXX)
+endif()
```
assignUser: No, faiss does that.
```suggestion
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+# Append FAISS cmake module path
+list(APPEND CMAKE_MODULE_PATH "${faiss_SOURCE_DIR}/cmake")
+
+# Set build options
+set(PREVIOUS_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+set(BUILD_SHARED_LIBS ${VELOX_BUILD_SHARED})
+set(PREVIOUS_BUILD_TYPE ${CMAKE_BUILD_TYPE})
+set(CMAKE_BUILD_TYPE Release)
+set(PREVIOUS_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -Wno-nonnull ")
```
assignUser: block() is a new'ish command we can use to make this much simpler
```suggestion
block()
set(BUILD_SHARED_LIBS OFF)
set(CMAKE_BUILD_TYPE Release)
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+# Create an alias for the FAISS library
+add_library(FAISS::faiss ALIAS faiss)
+
+# Reset build options
+set(BUILD_SHARED_LIBS ${PREVIOUS_BUILD_SHARED_LIBS})
+unset(PREVIOUS_BUILD_SHARED_LIBS)
+set(CMAKE_CXX_FLAGS ${PREVIOUS_CMAKE_CXX_FLAGS})
+unset(PREVIOUS_CMAKE_CXX_FLAGS)
+set(CMAKE_BUILD_TYPE ${PREVIOUS_BUILD_TYPE})
+unset(PREVIOUS_BUILD_TYPE)
```
assignUser:
```suggestion
endblock()
```

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,238 @@
...
+find_and_requre_openmp()
+find_and_require_blas()
+
+# Tells CMake to fetch FAISS from the specified URL, verify its checksum, and
+# apply any patches if needed.
+FetchContent_Declare(
+  faiss
+  URL ${VELOX_FAISS_SOURCE_URL}
+  URL_HASH ${VELOX_FAISS_BUILD_SHA256_CHECKSUM}
+  PATCH_COMMAND ${FAISS_PATCH_COMMAND})
```
assignUser: We don't  need the patch but marking as system makes sense.
```suggestion
  SYSTEM EXCLUDE_FROM_ALL)
```

CMake/resolve_dependency_modules/faiss.cmake
assignUser, line 1: Can we in general please also get rid of the redundant LLM comments ala
```
# Does thing
do_thing()
```
They just create noise.

CMakeLists.txt
```diff
@@ -229,6 +230,14 @@ if(${VELOX_ENABLE_DUCKDB})
   velox_resolve_dependency(DuckDB)
 endif()
 
+if(${VELOX_ENABLE_FAISS})
+  if(NOT DEFINED faiss_SOURCE)
+    set(faiss_SOURCE "BUNDLED")
+  endif()
```
assignUser: set_source handles this
```suggestion
```

CMakeLists.txt
```diff
@@ -229,6 +230,14 @@ if(${VELOX_ENABLE_DUCKDB})
   velox_resolve_dependency(DuckDB)
 endif()
 
+if(${VELOX_ENABLE_FAISS})
```
assignUser: You can just use the variable name in a condition, looks cleaner imo.
```suggestion
if(VELOX_ENABLE_FAISS)
```

CMakeLists.txt
```diff
@@ -160,6 +160,7 @@ option(
   OFF)
 option(VELOX_SIMDJSON_SKIPUTF8VALIDATION
        "Skip simdjson utf8 validation in JSON parsing" OFF)
+option(VELOX_ENABLE_FAISS "Build faiss vector search support" ON)
```
assignUser: This should not default to on.
```suggestion
option(VELOX_ENABLE_FAISS "Build faiss vector search support" OFF)
```

scripts/setup-centos9.sh
```diff
@@ -127,6 +130,27 @@ function install_adapters {
...
+  install_faiss_deps_from_dnf
+}
+
+function install_faiss_deps_from_dnf {
+  echo "Installing dependencies with yum..."
+  if [[ $EUID -ne 0 ]]; then
+    SUDO=sudo
+  else
+    SUDO=""
+  fi
```
assignUser: We already set this at the top of the script in a way that allows the user to override it.
```suggestion
```

scripts/setup-centos9.sh
```diff
@@ -127,6 +130,27 @@ function install_adapters {
...
+  rpm -q openblas-devel &>/dev/null || $SUDO dnf install -y openblas-devel
+  rpm -q libomp &>/dev/null || $SUDO dnf install -y libomp
```
assignUser: That check seems unnecessarily verbose, also use the existing helper function and are we sure faiss won't need the libomp development headers?
```suggestion
  dnf_install -y openblas-devel libomp-devel
```

scripts/setup-centos9.sh
```diff
@@ -127,6 +130,27 @@ function install_adapters {
...
+  echo "Checking OpenBLAS library location:"
+  ls -l /usr/lib/x86_64-linux-gnu/libopenblas* 2>/dev/null \
+    || ls -l /usr/lib/aarch64-linux-gnu/libopenblas* 2>/dev/null \
+    || ls -l /usr/lib64/libopenblas* 2>/dev/null \
+    || ls -l /usr/lib/libopenblas* 2>/dev/null \
+  || echo "OpenBLAS not found in standard locations."
```
assignUser: Is this a scenario that we really need to check after an install with dnf? Seems overly cautious, unless this is a known issue or something?
```suggestion
```

scripts/setup-helper-functions.sh
```diff
@@ -196,11 +196,28 @@ function wget_and_untar {
...
+        ;;
+    esac
+  done
+  if [ "$quote_args" = true ]; then
+    cmake_install "${args[@]}"
+  else
+    # shellcheck disable=SC2068
+    cmake_install ${args[@]}
+  fi
+  popd || return 1
```
assignUser: Why? just keep `$@` and add the additional args via variables in the call site

scripts/setup-centos9.sh
```diff
@@ -69,6 +69,9 @@ function install_velox_deps_from_dnf {
     libdwarf-devel elfutils-libelf-devel curl-devel libicu-devel bison flex \
     libsodium-devel zlib-devel gtest-devel gmock-devel xxhash-devel
 
+  # install faiss deps (openblas, openmp)
+  install_faiss_deps_from_dnf
+
```
assignUser: Why are you calling this explicitly, shouldn't it be called through install_faiss automatically?

scripts/setup-common.sh
```diff
@@ -21,6 +21,7 @@ VELOX_BUILD_SHARED=${VELOX_BUILD_SHARED:-"OFF"} #Build folly and gflags shared f
 CMAKE_BUILD_TYPE="${BUILD_TYPE:-Release}"
 DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)}
 BUILD_GEOS="${BUILD_GEOS:-true}"
+BUILD_FAISS="${BUILD_FAISS:-true}"
```
assignUser, line 25: Should default to false probably and then be explicitly enabled in the centos or adapters dockerfile? Not sure, depends on build time and I suspect that is long in this case?

scripts/setup-common.sh
```diff
@@ -232,6 +233,42 @@ function install_geos {
...
+
+    local cmake_args
+    cmake_args=(
+      -DFAISS_ENABLE_GPU=OFF
+      -DFAISS_ENABLE_PYTHON=OFF
+      -DFAISS_ENABLE_REMOTE=OFF
+      -DFAISS_ENABLE_GPU_TESTS=OFF
+      -DFAISS_ENABLE_BENCHMARKS=OFF
+      -DFAISS_ENABLE_OPENMP=ON
+    )
```
assignUser: No need to over-complicate it. Just add these to the call  in L267 directly

scripts/setup-common.sh
```diff
@@ -232,6 +233,42 @@ function install_geos {
...
+
+    if [[ -f /etc/os-release ]] && (grep -qi 'ubuntu' /etc/os-release || grep -qi 'centos' /etc/os-release); then
+      # Ubuntu or CentOS
+      cmake_args+=(
+        "-DCMAKE_CXX_FLAGS=-fopenmp"
+        "-DCMAKE_EXE_LINKER_FLAGS=-fopenmp"
+      )
+    else
+      echo "Warning: Unknown OS. You may need to set OpenMP flags manually."
+    fi
```
assignUser: faiss' CMake will handle these.
```suggestion
```

scripts/setup-macos.sh
```diff
@@ -94,6 +95,8 @@ function install_velox_deps_from_brew {
   do
     install_from_brew ${pkg}
   done
+  # Install faiss deps (openblas and libomp)
+  install_faiss_deps_from_brew
```
assignUser: Again, should be called implicitly through install_faiss
```suggestion
```

scripts/setup-macos.sh
```diff
@@ -123,6 +126,53 @@ function install_adapters {
   run_and_time install_hdfs
 }
 
+function install_faiss_deps_from_brew {
+  brew list openblas &>/dev/null || brew install openblas
+  brew list libomp &>/dev/null || brew install libomp
+}
+
+function install_faiss_deps {
+  install__faiss_deps_from_brew
```
assignUser: That typo made it not work implicitly I assume^^
```suggestion
  install_faiss_deps_from_brew
```

scripts/setup-macos.sh
```diff
@@ -123,6 +126,53 @@ function install_adapters {
...
+        "-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I${libomp_prefix}/include"
+        -DOpenMP_CXX_LIB_NAMES=omp
+        "-DOpenMP_omp_LIBRARY=${libomp_prefix}/lib/libomp.dylib"
+    )
+
+    local uname_m
+    uname_m=$(uname -m)
+    if [[ "$uname_m" == "arm64" ]]; then
+      # Apple Silicon: add arch-specific flags if needed
+      cmake_args+=("-DCMAKE_CXX_FLAGS=-mcpu=apple-m1+crc")
```
assignUser: This is handled in `cmake_install` which is used by `cmake_install_dir`

scripts/setup-macos.sh
```diff
@@ -123,6 +126,53 @@ function install_adapters {
...
+      -DFAISS_ENABLE_GPU=OFF
+    )
+
+    local libomp_prefix
+    libomp_prefix=$(brew --prefix libomp)
+    cmake_args+=(
+        "-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I${libomp_prefix}/include"
+        -DOpenMP_CXX_LIB_NAMES=omp
+        "-DOpenMP_omp_LIBRARY=${libomp_prefix}/lib/libomp.dylib"
+    )
```
assignUser: This is probably not necessary either, if cmake doesn't find libomp from hombrew on it's own we can set the relevant path hints for the find module.

scripts/setup-macos.sh
```diff
@@ -123,6 +126,53 @@ function install_adapters {
...
+function install_faiss_deps_from_brew {
+  brew list openblas &>/dev/null || brew install openblas
+  brew list libomp &>/dev/null || brew install libomp
+}
+
+function install_faiss_deps {
+  install__faiss_deps_from_brew
+}
+
+function install_faiss {
```
assignUser, line 142: You can just use the one from common.sh, see my comments why the modifications are not necessary.

scripts/setup-macos.sh
```diff
@@ -123,6 +126,53 @@ function install_adapters {
   run_and_time install_hdfs
 }
 
+function install_faiss_deps_from_brew {
+  brew list openblas &>/dev/null || brew install openblas
+  brew list libomp &>/dev/null || brew install libomp
```
assignUser: Again, why the checking? homebrew won't error without it...

scripts/setup-ubuntu.sh
```diff
@@ -220,12 +232,15 @@ function install_velox_deps {
...
   run_and_time install_geos
+  run_and_time install_faiss
 }
 
 function install_apt_deps {
   install_build_prerequisites
   install_format_prerequisites
   install_velox_deps_from_apt
+  # Install dependencies for faiss (openblas, openmp)
+  install_faiss_deps_from_apt
```
assignUser:
```suggestion
```

scripts/setup-ubuntu.sh
```diff
@@ -201,6 +201,18 @@ function install_adapters {
   run_and_time install_hdfs
 }
 
+function install_faiss_deps_from_apt {
+  echo "Updating apt-get and installing faiss dependencies..."
+  dpkg -s libopenblas-dev &>/dev/null || sudo apt-get install -y libopenblas-dev
+  dpkg -s libomp-dev &>/dev/null || sudo apt-get install -y libomp-dev
+  echo "Checking OpenBLAS library location:"
+  ls -l /usr/lib/x86_64-linux-gnu/libopenblas* || echo "OpenBLAS not found in standard location."
```
assignUser:
```suggestion
  sudo apt-get install -y libopenblas-dev libomp-dev
```

.github/workflows/linux-build-base.yml
zhichenxu-meta: Removed FAISS_VERSION

.github/workflows/linux-build-base.yml
zhichenxu-meta: Thanks, and added these in  faiss.cmake

.github/workflows/linux-build-base.yml
zhichenxu-meta: removed

.github/workflows/macos.yml
zhichenxu-meta: removed

CMake/resolve_dependency_modules/faiss.cmake
zhichenxu-meta: removed

scripts/setup-ubuntu.sh
zhichenxu-meta: Applied suggested change

scripts/setup-macos.sh
zhichenxu-meta: Removed checks

CMake/resolve_dependency_modules/faiss.cmake
zhichenxu-meta: Removed function and cache variables

scripts/setup-helper-functions.sh
zhichenxu-meta: Simplified to keep @

scripts/setup-macos.sh
zhichenxu-meta: If I remove:     local libomp_prefix
    libomp_prefix=$(brew --prefix libomp)
    cmake_args+=(
        "-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I${libomp_prefix}/include"
        -DOpenMP_CXX_LIB_NAMES=omp
        "-DOpenMP_omp_LIBRARY=${libomp_prefix}/lib/libomp.dylib"
    )
I get the following error when build on macos (local):
CMake Error at /Users/zhichenxu/presto/presto-native-execution/.venv/lib/python3.12/site-packages/cmake/data/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
  /Users/zhichenxu/presto/presto-native-execution/.venv/lib/python3.12/site-packages/cmake/data/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /Users/zhichenxu/presto/presto-native-execution/.venv/lib/python3.12/site-packages/cmake/data/share/cmake-3.31/Modules/FindOpenMP.cmake:616 (find_package_handle_standard_args)
  faiss/CMakeLists.txt:379 (find_package)

scripts/setup-macos.sh
zhichenxu-meta: Thank you Jacob for the detailed review and it was super helpful!

I have incorporated most of your suggestions, and there are several remaining problems would appreciate your suggestions:

1) macOs Build does not seem to build Faiss, even though I have -DVELOX_ENABLE_FAISS=ON in macos.yml, and BUILD_FAISS to true in setup-common.sh

2) I am getting the following error in Build with GCC / Linux release with adapters
-- [faiss] Could NOT find MKL (missing: MKL_LIBRARIES) 
-- [faiss] Looking for sgemm_
-- [faiss] Looking for sgemm_ - not found
CMake Error at /usr/local/lib64/python3.9/site-packages/cmake/data/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find BLAS (missing: BLAS_LIBRARIES)
  /usr/local/lib64/python3.9/site-packages/cmake/data/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/lib64/python3.9/site-packages/cmake/data/share/cmake-3.30/Modules/FindBLAS.cmake:1417 (find_package_handle_standard_args)
  _build/release/_deps/faiss-src/faiss/CMakeLists.txt:395 (find_package)

zhichenxu-meta: 3) For macos (locally on my laptop), if I do not include:   local libomp_prefix=$(brew --prefix libomp)
     cmake_args+=(
       "-DOpenMP_CXX_FLAGS=-Xpreprocessor -fopenmp -I${libomp_prefix}/include"
       -DOpenMP_CXX_LIB_NAMES=omp
       "-DOpenMP_omp_LIBRARY=${libomp_prefix}/lib/libomp.dylib"

I will get: 
 Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES)
  /Users/zhichenxu/presto/presto-native-execution/.venv/lib/python3.12/site-packages/cmake/data/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /Users/zhichenxu/presto/presto-native-execution/.venv/lib/python3.12/site-packages/cmake/data/share/cmake-3.31/Modules/FindOpenMP.cmake:616 (find_package_handle_standard_args)
  faiss/CMakeLists.txt:379 (find_package)

CMakeLists.txt
assignUser: Really? Why? I feel like vector search is clearly an optional feature similar to adapters?

scripts/setup-macos.sh
assignUser: 1. I'll have to look at a finished run currently it's still building
2. Yeah, looks like adding blas (and probably openmp too looking at 3) to the prefix path is required, see https://github.com/facebookincubator/velox/blob/main/CMakeLists.txt#L432

scripts/setup-common.sh
zhichenxu-meta: I am following what is done for GEOS

scripts/setup-macos.sh
zhichenxu-meta: @assignUser Hi Jacob, 
Please take another look when you get a chance. I ended up have to install blas (vi calling install_faiss_deps_from_brew in macos.yml) for  "build Macos" to work.

scripts/setup-centos9.sh
```diff
@@ -129,6 +129,20 @@ function install_adapters {
   run_and_time install_hdfs
 }
 
+function install_faiss_deps {
+  install_faiss_deps_from_dnf
+}
+
+function install_faiss_deps_from_dnf {
+  echo "Installing dependencies with yum..."
```
kgpai: nit: remove..

CMakeLists.txt
kgpai: For the same reason we have enabled GEOS by default so that we get some testing that this builds. Ideally we need another build job if these are turned off so that we are sure that they build and tests pass. At the very least this should be enabled in this PR, so we are sure that builds dont suddenly fail when this is turned on.

scripts/setup-centos9.sh
zhichenxu-meta: I have removed install_faiss_deps_from_dnf, and we need install_faiss_deps.

CMakeLists.txt
assignUser: It's enabled in the adapters build: https://github.com/facebookincubator/velox/pull/13575/files#diff-085554acc2b3df5c85e2ca47e37f39ba130544f3f51f97b991ed5c05532ae539R100

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,66 @@
...
+    COMMAND brew --prefix openblas
+    RESULT_VARIABLE BREW_OPENBLAS_RESULT
+    OUTPUT_VARIABLE BREW_OPENBLAS_PREFIX
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  if(BREW_OPENBLAS_RESULT EQUAL 0 AND EXISTS "${BREW_OPENBLAS_PREFIX}")
+    list(APPEND CMAKE_PREFIX_PATH "${BREW_OPENBLAS_PREFIX}")
+  endif()
+  if(NOT BLAS_LIBRARIES)
+    set(BLAS_LIBRARIES "${BREW_OPENBLAS_PREFIX}/lib/libopenblas.dylib")
+  endif()
```
assignUser: I don't understand the purpose of this? This will always be true unless we manually feed in `-DBLAS_LIBRARIES` via cli as `find_package(BLAS)` wasn't called. This should be handled by the   find_package call in faiss
```suggestion
```

.github/workflows/linux-build-base.yml
```diff
@@ -67,6 +68,9 @@ jobs:
             install_adapters
```
assignUser: This step should allow us to test with faiss_SOURCE=SYSTEM! But let's the current run finish first.

CMake/resolve_dependency_modules/faiss.cmake
zhichenxu-meta: Let me try remove it to see if the test will pass.

CMakeLists.txt
zhichenxu-meta: I will keep it OFF since it is enabled in adapters build.

.github/workflows/linux-build-base.yml
zhichenxu-meta: This only installs the dependencies of faiss (not faiss), would SYSTEM still work?

scripts/setup-macos.sh
zhichenxu-meta: I have simplified the code, so the above comments are no longer accurate.

I ended up needing to call install_faiss_deps to install libomp and blas.

.github/workflows/linux-build-base.yml
assignUser: It should because you added it to `install_velox_deps` which is run in this step at the top

zhichenxu-meta: Changed faiss_SOURCE=SYSTEM to testout

CMake/resolve_dependency_modules/faiss.cmake
```diff
@@ -0,0 +1,63 @@
...
+  if(BREW_OPENBLAS_RESULT EQUAL 0 AND EXISTS "${BREW_OPENBLAS_PREFIX}")
+    list(APPEND CMAKE_PREFIX_PATH "${BREW_OPENBLAS_PREFIX}")
+  endif()
+endif()
+
+FetchContent_Declare(
+  faiss
+  URL ${VELOX_FAISS_SOURCE_URL}
+  URL_HASH ${VELOX_FAISS_BUILD_SHA256_CHECKSUM}
+  SYSTEM EXCLUDE_FROM_ALL)
```
assignUser:
```suggestion
  SYSTEM) # Once there are targets depending on FAISS we can add `EXCLUDE_FROM_ALL` back in
```
Can you temporarily remove this, as this stops faiss from being build bundled (because there are no targets depending on it currently), so in the macos job it's only being configured but not build.

Once you add targets that need it we can add this back in.

zhichenxu-meta: Done

zhichenxu-meta: @assignUser Jacob, would you please let me know if there are any additional changes you would suggest, and thank you  again for your great suggestions.

zhichenxu-meta:
@assignUser, all tests have passed. Could you please review and approve the PR? I appreciate your feedback and suggestions, which were extremely helpful. Thank you again for your time and expertise!

scripts/setup-ubuntu.sh
```diff
@@ -227,6 +227,10 @@ function install_adapters {
   run_and_time install_hdfs
 }
 
+function install_faiss_deps {
+  sudo apt-get install -y libopenblas-dev libomp-dev
```
anandamideShakyan, line 231: @zhichenxu-meta We should be using $SUDO in place of sudo since this causes error while building docker image of presto worker.
```
#11 829.3 real	0m0.000s
#11 829.3 user	0m0.000s
#11 829.3 sys	0m0.000s
#11 ERROR: process "/bin/sh -c mkdir build &&     (cd build && ../scripts/setup-ubuntu.sh &&                          apt install -y rpm &&                  ../velox/scripts/setup-ubuntu.sh install_adapters &&                  ../scripts/setup-adapters.sh ) &&     rm -rf build" did not complete successfully: exit code: 127
------
 > [6/6] RUN mkdir build &&     (cd build && ../scripts/setup-ubuntu.sh &&                          apt install -y rpm &&                  ../velox/scripts/setup-ubuntu.sh install_adapters &&                  ../scripts/setup-adapters.sh ) &&     rm -rf build:
829.3 + run_and_time install_faiss
829.3 + install_faiss
829.3 + [[ true == \t\r\u\e ]]
829.3 + install_faiss_deps
829.3 + sudo apt-get install -y libopenblas-dev libomp-dev
829.3 ../scripts/../velox/scripts/setup-ubuntu.sh: line 223: sudo: command not found
829.3
829.3 real	0m0.000s
829.3 user	0m0.000s
829.3 sys	0m0.000s
------
ubuntu-22.04-dependency.dockerfile:34
--------------------
  33 |     # install rpm needed for minio install.
  34 | >>> RUN mkdir build && \
  35 | >>>     (cd build && ../scripts/setup-ubuntu.sh && \
  36 | >>>                          apt install -y rpm && \
  37 | >>>                  ../velox/scripts/setup-ubuntu.sh install_adapters && \
  38 | >>>                  ../scripts/setup-adapters.sh ) && \
  39 | >>>     rm -rf build
  40 |
--------------------
ERROR: failed to solve: process "/bin/sh -c mkdir build &&     (cd build && ../scripts/setup-ubuntu.sh &&                          apt install -y rpm &&                  ../velox/scripts/setup-ubuntu.sh install_adapters &&                  ../scripts/setup-adapters.sh ) &&     rm -rf build" did not complete successfully: exit code: 127
```

PR #13588, opened 2025-06-02
refactor: Replace 'typedef' with 'using'

lingbin:
In modern C++, `using` is preferred over `typedef` for defining aliases.

There is also a " modernize-use-using " rule in clang-tidy, see [1].

No functional changes.

[1] https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-using.html


lingbin:
@mbasmanova  Could you please help review this PR? It's a very small change, with no functional changes. Thanks.

PR #13590, opened 2025-06-02
test: Exclude bing_tile functions from expression fuzzer

Yuhta:
Summary: Some fuzzer tests are failing: https://github.com/facebookincubator/velox/actions/runs/15370535592/job/43249537201

Differential Revision: D75762689

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -153,6 +153,16 @@ int main(int argc, char** argv) {
       "st_issimple",
       "geometry_invalid_reason",
       "simplify_geometry",
+      // bing_tile functions:
+      // https://github.com/facebookincubator/velox/issues/13593
+      "bing_tile",
+      "bing_tile_zoom_level",
+      "bing_tile_coordinates",
+      "bing_tile_parent",
+      "bing_tile_children",
+      "bing_tile_quadkey"
+      "bing_tile_at",
+      "bing_tiles_around",
   };
   size_t initialSeed = FLAGS_seed == 0 ? std::time(nullptr) : FLAGS_seed;
 


kgpai (approved):

kagamiori (approved):

kagamiori:
Hi @bikramSingh91 and @kgpai, the error message looks like a real bug. Do you have context about this function? Should we take a look and fix it?
```
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (-2147483648 vs. 0) Size must be non-negative.
Retriable: False
Expression: size >= 0
Context: bing_tile_children(c0, c1)
Additional Context: Top-level Expression: try(bing_tile_children(c0, c1))
Function: resize
File: /__w/velox/velox/velox/velox/vector/BaseVector.cpp
Line: 114
```

bikramSingh91:
> Hi @bikramSingh91 and @kgpai, the error message looks like a real bug. Do you have context about this function? Should we take a look and fix it?
> 
> ```
> Error Source: RUNTIME
> Error Code: INVALID_STATE
> Reason: (-2147483648 vs. 0) Size must be non-negative.
> Retriable: False
> Expression: size >= 0
> Context: bing_tile_children(c0, c1)
> Additional Context: Top-level Expression: try(bing_tile_children(c0, c1))
> Function: resize
> File: /__w/velox/velox/velox/velox/vector/BaseVector.cpp
> Line: 114
> ```

@kagamiori Almost done with a fix for it. the root cause is as follows:

for the version of bing_tile_children that takes a target zoom level, if the difference in current and target zoom level is large, it can translate to an enormous number of children for each row. For eg, if the difference is 10 then we will have 2^20 children for a single row. For velox, this results in a velox vector that goes over the max size limit (max 32-bit integer) and errors out (fortunately it hit a CHECK). For presto java, I tried it out using (select BING_TILE_CHILDREN(BING_TILE(0,0,1), 20);) and it just timed out.

The fix here is that we should have a max limit of the difference between the target and current zoom level, and error out if we go over that limit. For this solution we are starting off with a limit of 5 and make it tunable via a debug query option. We should also probably apply the same limit on java side cause it seems like it also does not have any enforced as well (see [BingTile.java](http://bingtile.java/) L174)

EDIT: Here is the fix: #13604

Issue #13215, opened 2025-05-01
Duckdb fails to build with Apple clang 17


kgpai:
### Problem description

I get the following build error when building DDB with Apple clang 17.  As we are still on a relatively old version of DDB (0.8.1) , this build problem seems more relevant to Velox than ddb.

``` 
FAILED: src/core_functions/aggregate/distributive/CMakeFiles/duckdb_aggr_distributive.dir/ub_duckdb_aggr_distributive.cpp.o 
[...]

PR #13598, opened 2025-06-02
build(macos): DuckDB on Apple Clang17

majetideepak:
Fixes: https://github.com/facebookincubator/velox/issues/13215


majetideepak:
@kKPulla can you please take a look at this import?

PR #13601, opened 2025-06-02
build(CUDA): Install GPUDirect Storage dependency

zoltan:
we need to install libcufile in order for GDS-enabled libraries (such as kvikio used by cudf) to pick up GDS support during compile time

### DIFF
--- a/scripts/setup-centos9.sh
+++ b/scripts/setup-centos9.sh
@@ -93,7 +93,8 @@ function install_cuda {
     cuda-driver-devel-$dashed \
     cuda-minimal-build-$dashed \
     cuda-nvrtc-devel-$dashed \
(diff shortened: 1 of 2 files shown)


majetideepak:
@devavret, @bdice any feedback?

bdice (approved):
This looks fine. @zoltan You may also need libnuma — please check if GDS support is working with/without libnuma if you can.

zoltan:
@bdice thanks for the suggestion - I've added libnuma as a dependency as well; I don't think we need the development headers. interestingly enough, ldd on libcufile doesn't show it needs numa, however, I saw there are kvikio issues open where the libnuma dependency is evident. thanks!

majetideepak (approved):

PR #13604, opened 2025-06-02
fix: Add config to cap max zoom level difference in bing_tile_children

bikramSingh91:
Summary:
Added a new query configuration that applies to the
bing_tile_children UDF to cap the maximum zoom level difference
between the parent and the child, with a default value set to 5. This
configuration helps control the number of children generated by the
function, preventing excessively large array outputs that could exceed
the size limits of the elements vector in the Velox array vector.

Differential Revision: D75820054

### DIFF
--- a/velox/functions/prestosql/BingTileFunctions.h
+++ b/velox/functions/prestosql/BingTileFunctions.h
@@ -16,6 +16,7 @@
 #pragma once
 
 #include "velox/common/base/Status.h"
+#include "velox/core/QueryConfig.h"
 #include "velox/functions/Macros.h"
 #include "velox/functions/prestosql/types/BingTileType.h"
 
@@ -156,6 +157,17 @@ template <typename T>
 struct BingTileChildrenFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
+  uint8_t maxZoomShift = 5;
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& inputTypes,
+      const core::QueryConfig& config,
+      const arg_type<BingTile>* /* tile */,
+      const arg_type<int8_t>* /* childZoom */) {
+    maxZoomShift =
+        std::max<uint8_t>(config.debugBingTileChildrenMaxZoomShift(), 1);
+  }
+
   FOLLY_ALWAYS_INLINE Status
   call(out_type<Array<BingTile>>& result, const arg_type<BingTile>& tile) {
     uint64_t tileInt = tile;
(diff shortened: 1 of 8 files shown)


velox/functions/prestosql/tests/BingTileFunctionsTest.cpp
```diff
@@ -446,6 +447,25 @@ TEST_F(BingTileFunctionsTest, bingTileChildrenZoom) {
       "Cannot call bing_tile_children with negative zoom");
   VELOX_ASSERT_USER_THROW(
       testBingTileChildren(0, 0, 2, 1), "Child zoom 1 must be >= tile zoom 2");
+  VELOX_ASSERT_USER_THROW(
+      testBingTileChildren(0, 0, 1, 7),
+      "Difference between parent zoom (1) and child zoom (7) must be <= 5");
```
kgpai, line 452: nit: must be <= {}  (Dont hard code value if possible here).

velox/functions/prestosql/tests/utils/FunctionBaseTest.h
```diff
@@ -106,6 +106,23 @@ class FunctionBaseTest : public testing::Test,
     return evaluate(typedExpr, data, rows);
   }
 
+  VectorPtr evaluateWithCustomQueryConfig(
```
kgpai: You can just set this in your test case using : 

``` 
queryCtx_->testingOverrideConfigUnsafe({ .. })
``` 

instead of creating this.

kgpai (approved):
Looks great - some nits.

velox/functions/prestosql/BingTileFunctions.h
```diff
@@ -156,6 +157,17 @@ template <typename T>
 struct BingTileChildrenFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
+  uint8_t maxZoomShift = 5;
```
kagamiori, line 160: nit: Why is 5 needed here? It'll be assigned config.debugBingTileChildrenMaxZoomShift(), right?

bikramSingh91: not strictly required as it will be assigned in the constructor anyway. The alternative here would be to set it to a default value of 1 for it to work for the version that does not take the target zoom level as input. I dont really have a strong opinion either way, let me know if you feel strongly to change it

kagamiori: @bikramSingh91 I see. Then I think we can set it to the default value of 1 to avoid confusion.

PR #13610, opened 2025-06-03
feat: Extract subfield filters using FiltersAsNode in TableEvolutionFuzzer

juwentus1234:
Summary: Extract expected path using FiltersAsNode path in TableEvolutionFuzzer.

Differential Revision: D75842056

### DIFF
--- a/velox/exec/tests/TableEvolutionFuzzer.cpp
+++ b/velox/exec/tests/TableEvolutionFuzzer.cpp
@@ -418,10 +418,10 @@ void TableEvolutionFuzzer::run() {
 
   std::vector<std::unique_ptr<TaskCursor>> scanTasks(2);
 
-  scanTasks[0] =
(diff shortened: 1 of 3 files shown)


velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -605,6 +606,7 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeScanTask(
                             /*remainingFilter=*/"",
                             tableSchema,
                             {})
+                        .filtersAsNode(useFiltersAsNode)
```
Yuhta: This need to be before `tableScanWithPushDown` otherwise the filter is still push down.  Also it worths print some log to double check it's not pushdown on expected side.

juwentus1234: updated

PR #13612, opened 2025-06-03
feat: Support single node tracing

duanmeng:
The TaskTraceMetadataWriter serializes and writes the entire plan
fragment to the task metadata file. Then, the TaskTraceMetadataReader
deserializes the task metadata, extracts the target trace node, and constructs
a replay node by using it. During this process, there may be some plan nodes
that do not support serialization and deserialization in the whole plan fragment,
and it would break the tracing of the target tracing node. Hence, we need only
to trace the target node and replace its source nodes with DummySourceNode
to retain the trace node's input types.

### DIFF
--- a/velox/exec/TraceUtil.cpp
+++ b/velox/exec/TraceUtil.cpp
@@ -18,10 +18,12 @@
 
 #include <folly/json.h>
 
-#include <numeric>
+#include <utility>
+
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/file/File.h"
 #include "velox/common/file/FileSystems.h"
+#include "velox/exec/TableWriter.h"
 #include "velox/exec/Trace.h"
 
 namespace facebook::velox::exec::trace {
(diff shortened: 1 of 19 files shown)


velox/exec/TraceUtil.cpp
```diff
@@ -242,4 +285,145 @@ bool canTrace(const std::string& operatorType) {
...
+        tableWriteNode->aggregationNode(),
+        tableWriteNode->insertTableHandle(),
+        tableWriteNode->hasPartitioningScheme(),
+        TableWriteTraits::outputType(tableWriteNode->aggregationNode()),
+        tableWriteNode->commitStrategy(),
+        std::make_shared<DummySourceNode>(
+            tableWriteNode->sources().front()->outputType()));
+  }
+
+  VELOX_UNSUPPORTED();
```
xiaoxmeng: Log unsupported trace node info?

velox/exec/TraceUtil.h
```diff
@@ -135,4 +135,11 @@ folly::dynamic getTaskMetadata(
 
 /// Checks whether the operator can be traced.
 bool canTrace(const std::string& operatorType);
+
+/// Copy the target plan node, and replay the source node with DummySourceNode.
+core::PlanNodePtr copyTraceNode(
```
xiaoxmeng:
```
/// Gets the specified the trace node from 'plan'. In the returned trace node, we replace its source nodes with DummySourceNode for replay.
s/copyTraceNode/getTraceNode.
s/node/plan/
```

xiaoxmeng (approved):
@duanmeng LGTM % minors. Thanks!

gggrace14 (approved):

PR #11934, opened 2024-12-23
feat: Add map_from_entries Spark function

PHILO-HE:

PR #13614, opened 2025-06-03
feat: Register map_from_entries function for Spark

Surbhi-Vijay:
Presto implementation for MapFromEntries function can be used for Spark as well. The only difference is in duplicate key handling. The current behavior in velox is to throw exception when duplicate keys are encountered.

Spark has two behavior: 1. Exception 2. LAST_WIN.
There will be separate PR for refactoring and handling duplicate key for Spark.

### DIFF
--- a/velox/functions/sparksql/registration/RegisterMap.cpp
+++ b/velox/functions/sparksql/registration/RegisterMap.cpp
@@ -29,7 +29,8 @@ void registerSparkMapFunctions(const std::string& prefix) {
   //   function expression corresponds to body, arguments to signature
   VELOX_REGISTER_VECTOR_FUNCTION(udf_map_filter, prefix + "map_filter");
   VELOX_REGISTER_VECTOR_FUNCTION(udf_map_entries, prefix + "map_entries");
-  VELOX_REGISTER_VECTOR_FUNCTION(udf_map_from_entries, prefix + "map_from_entries");
+  VELOX_REGISTER_VECTOR_FUNCTION(
+      udf_map_from_entries, prefix + "map_from_entries");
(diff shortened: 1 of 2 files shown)


Surbhi-Vijay:
cc: @rui-mo @zhli1142015

velox/functions/sparksql/registration/RegisterMap.cpp
```diff
@@ -29,6 +29,8 @@ void registerSparkMapFunctions(const std::string& prefix) {
   //   function expression corresponds to body, arguments to signature
   VELOX_REGISTER_VECTOR_FUNCTION(udf_map_filter, prefix + "map_filter");
   VELOX_REGISTER_VECTOR_FUNCTION(udf_map_entries, prefix + "map_entries");
+  VELOX_REGISTER_VECTOR_FUNCTION(
```
zhli1142015, line 32: Could you add UTs under spark directory?

zhli1142015:
Would you like to open a PR in Gluten to enable this function and relevant Spark UTs?

Surbhi-Vijay:
> Would you like to open a PR in Gluten to enable this function and relevant Spark UTs?

I have opened a PR here: https://github.com/apache/incubator-gluten/pull/9854.

zhli1142015:
And I find there is already a PR for this function.
https://github.com/facebookincubator/velox/pull/11934
BTW, from the above pr, we can't reuse this function due to differences in null handling.

Surbhi-Vijay:
> And I find there is already a PR for this function. #11934 BTW, from the above pr, we can't reuse this function due to differences in null handling.

Thanks for sharing, I will close this PR.

PR #13620, opened 2025-06-03
fix: Add config for requested type check in ReaderBase::convertType

rui-mo:
A check on the requested type against the element converted type has been 
introduced in https://github.com/facebookincubator/velox/commit/5ad65e4a487ea4a43dc808802726a479b017b1b8.  However, some check failures are acceptable in Spark
scan. For example:
- The bigint data could be read as integer type in Spark.
- When the element converted type in scalar type, the check requires the
requested type must not be array. However, an unannotated array in Parquet
is a repeated field that is not explicitly marked as a LIST logical type, and it
depends on the reader to decide whether the data should be interpreted as
scalar values or array.

To avoid the type check failures and allow more compatibility with engines like
Spark, this PR adds a configuration to enable or disable the type check.

### DIFF
--- a/velox/connectors/hive/HiveConfig.cpp
+++ b/velox/connectors/hive/HiveConfig.cpp
@@ -235,6 +235,13 @@ bool HiveConfig::readStatsBasedFilterReorderDisabled(
       config_->get<bool>(kReadStatsBasedFilterReorderDisabled, false));
 }
 
+bool HiveConfig::isRequestedTypeCheckEnabled(
+    const config::ConfigBase* session) const {
+  return session->get<bool>(
+      kEnableRequestedTypeCheckSession,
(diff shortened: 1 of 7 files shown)


Yuhta:
These checks are pretty much tied to the reader implementation not higher level engines.  Unless we have the corresponding support added (e.g. BIGINT -> INTEGER conversion), we should not remove these checks.

rui-mo:
@Yuhta I understand. Thank you. I'll get to work on supporting the cases we found.

FelixYBW:
@pedroerp Can you help to review and merge the PR? It's essential one for Gluten and currently picked by ibm/velox.

rui-mo:
@FelixYBW This PR is only for Gluten to cherry-pick. Per @Yuhta's suggestion, we are adding the support in https://github.com/facebookincubator/velox/pull/13864.

PR #13629, opened 2025-06-04
misc: Remove unused connector reference

yingsu00:
Remove the unused Hive dependencies from targets


assignUser (approved):
Very nice, thanks!

PR #13633, opened 2025-06-04
fix(fuzzer): Fix ExpressionFuzzerUnitTest.reproduceExpressionWithSeed

kagamiori:
Summary:
This diff fixes ExpressionFuzzerUnitTest.reproduceExpressionWithSeed by making 
sure the JsonInputGenerator always sort object keys in a deterministic order.

Differential Revision: D75960981

### DIFF
--- a/velox/common/fuzzer/ConstrainedGenerators.cpp
+++ b/velox/common/fuzzer/ConstrainedGenerators.cpp
@@ -49,13 +49,21 @@ folly::json::serialization_opts getSerializationOptions(
   folly::json::serialization_opts opts;
   opts.allow_non_string_keys = true;
   opts.allow_nan_inf = true;
+  opts.sort_keys = true;
   if (makeRandomVariation) {
     opts.convert_int_keys = rand<bool>(rng);
     opts.pretty_formatting = rand<bool>(rng);
     opts.pretty_formatting_indent_width = rand<uint32_t>(rng, 0, 4);
     opts.encode_non_ascii = rand<bool>(rng);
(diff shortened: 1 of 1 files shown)


velox/common/fuzzer/ConstrainedGenerators.cpp
```diff
@@ -49,13 +49,21 @@ folly::json::serialization_opts getSerializationOptions(
   folly::json::serialization_opts opts;
   opts.allow_non_string_keys = true;
   opts.allow_nan_inf = true;
+  opts.sort_keys = true;
```
kKPulla, line 52: Just for my understanding, this means that we will always generate json inputs with keys as sorted in our fuzzer, and the variation is either asc/desc right? Asking because, this feels to me like we will not coverage for jsons where keys are not sorted? Please correct me if I am wrong.

kagamiori: @kKPulla, 
> this means that we will always generate json inputs with keys as sorted in our fuzzer, and the variation is either asc/desc right?

Yes, your understanding is correct. I think this is enough to discover behavior mismatch in Velox vs. Presto, such as one engine sort the keys in ascending order and the other one doesn't (because one engine always sort the keys in a single way, whereas we test two possible orders). If we don't set this option, the input generation will serialize object keys with non-deterministic order and cause fuzzer failures not reproducible via the seed.

kKPulla: I see, that makes sense. Thanks!

kKPulla (approved):

PR #13642, opened 2025-06-05
fix: Enable constant folding for lambda functions

mbasmanova:
Summary:
ExprCompiler used to not constant fold lambda functions. transform(array[1, 2, 3], x -> x * 2) would not be constant folded.

This was due to Expr::isConstant returning false for such an expression. Expr::isConstant returned true iff expression is deterministic and all input are ConstantExpr.

A fix is to modify Expr::isConstant to return true iff expression is deterministic and has no dependencies (distinctFields_ is empty).

Also, added convenience API tryEvaluateConstantExpression to compliment existing evaluateConstantExpression. The new API can be safely called on any expression without ensuring the expression is constant-foldable.

Differential Revision: D76004362

### DIFF
--- a/velox/expression/Expr.cpp
+++ b/velox/expression/Expr.cpp
@@ -1192,20 +1192,36 @@ void Expr::evalWithMemo(
   VectorPtr base;
   distinctFields_[0]->evalSpecialForm(rows, context, base);
 
+  // evalWithNulls may throw an exception. If this happens during constant
+  // folding, the exception is suppressed and the Expr object may be reused.
+  // Hence, it is important to update state in way that ensure "valid" state in
+  // case of exceptions.
+  //
+  // Also, note that the same expression running on same data may pass or may
+  // fail depending on whether it runs under TRY or not.
+  //
+  // An example expression that triggers these edge cases:
+  //
+  //    try(coalesce(array_min_by(array[1, 2, 3], x -> x / 0), 0::INTEGER))
+
   if (base.get() != baseOfDictionaryRawPtr_ ||
       baseOfDictionaryWeakPtr_.expired()) {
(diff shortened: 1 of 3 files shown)


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

velox/expression/Expr.cpp
```diff
@@ -1195,17 +1195,21 @@ void Expr::evalWithMemo(
...
     baseOfDictionaryRepeats_ = 0;
-    baseOfDictionaryWeakPtr_ = base;
-    baseOfDictionaryRawPtr_ = base.get();
+    baseOfDictionaryWeakPtr_.reset();
+    baseOfDictionaryRawPtr_ = nullptr;
     context.releaseVector(baseOfDictionary_);
     context.releaseVector(dictionaryCache_);
+
     evalWithNulls(rows, context, result);
+    baseOfDictionaryWeakPtr_ = base;
```
bikramSingh91, line 1216: nit: can you add context on why its important to set the state here. These important details can sometimes get lost  during refactors if only mentioned in the PR description.

mbasmanova: @bikramSingh91 Adding the following comment:

```
  // evalWithNulls may throw an exception. If this happens during constant
  // folding, the exception is suppressed and the Expr object may be reused.
  // Hence, it is important to update state in way that ensure "valid" state in
  // case of exceptions.
  //
  // Also, note that the same expression running on same data may pass or may
  // fail depending on whether it runs under TRY or not.
  //
  // An example expression that triggers these edge cases:
  //
  //    try(coalesce(array_min_by(array[1, 2, 3], x -> x / 0), 0::INTEGER))
```

PR #13643, opened 2025-06-05
feat: support namespace prefixes in cudf

zoltan:
Prestissimo uses a custom namespace prefix for its functions which is currently not understood by the cuDF code. this change enables Prestissimo (or other applications using Velox) to register the prefix.

### DIFF
--- a/velox/experimental/cudf/exec/CudfHashAggregation.cpp
+++ b/velox/experimental/cudf/exec/CudfHashAggregation.cpp
@@ -30,6 +30,8 @@
 #include <cudf/stream_compaction.hpp>
 #include <cudf/unary.hpp>
 
+#include <boost/algorithm/string.hpp>
+
 namespace {
 
 using namespace facebook::velox;
@@ -394,19 +396,19 @@ std::unique_ptr<cudf_velox::CudfHashAggregation::Aggregator> createAggregator(
     bool isGlobal) {
   // Companion function may be count_merge_extract or count_partial or others,
   // so use this to map
-  if (kind.rfind("sum", 0) == 0) {
+  if (boost::algorithm::ends_with(kind, "sum")) {
     return std::make_unique<SumAggregator>(
         step, inputIndex, constant, isGlobal);
-  } else if (kind.rfind("count", 0) == 0) {
+  } else if (boost::algorithm::ends_with(kind, "count")) {
     return std::make_unique<CountAggregator>(
         step, inputIndex, constant, isGlobal);
-  } else if (kind.rfind("min", 0) == 0) {
+  } else if (boost::algorithm::ends_with(kind, "min")) {
     return std::make_unique<MinAggregator>(
         step, inputIndex, constant, isGlobal);
-  } else if (kind.rfind("max", 0) == 0) {
+  } else if (boost::algorithm::ends_with(kind, "max")) {
     return std::make_unique<MaxAggregator>(
         step, inputIndex, constant, isGlobal);
-  } else if (kind.rfind("avg", 0) == 0) {
+  } else if (boost::algorithm::ends_with(kind, "avg")) {
     return std::make_unique<MeanAggregator>(
         step, inputIndex, constant, isGlobal);
   } else {
--- a/velox/experimental/cudf/exec/ExpressionEvaluator.cpp
+++ b/velox/experimental/cudf/exec/ExpressionEvaluator.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "velox/experimental/cudf/exec/ExpressionEvaluator.h"
+#include "velox/experimental/cudf/exec/ToCudf.h"
 
 #include "velox/expression/ConstantExpr.h"
 #include "velox/expression/FieldReference.h"
@@ -29,6 +30,8 @@
 #include <cudf/strings/slice.hpp>
 #include <cudf/table/table.hpp>
 #include <cudf/transform.hpp>
+
+#include <boost/algorithm/string.hpp>
 
 namespace facebook::velox::cudf_velox {
 namespace {
@@ -249,8 +252,18 @@ const std::unordered_set<std::string> supportedOps = {
 
 namespace detail {
 
+inline std::string strip_prefix(
+    const std::string& input,
+    const std::string& prefix) {
+  if (boost::algorithm::starts_with(input, prefix)) {
+    return input.substr(prefix.size());
+  }
+  return input;
+}
+
 bool canBeEvaluated(const std::shared_ptr<velox::exec::Expr>& expr) {
-  const auto& name = expr->name();
+  const auto& name =
+      strip_prefix(expr->name(), CudfOptions::getInstance().prefix());
   if (supportedOps.count(name) || binaryOps.count(name) ||
       unaryOps.count(name)) {
     return std::all_of(
@@ -334,7 +347,8 @@ cudf::ast::expression const& AstContext::multipleInputsToPairWise(
     const std::shared_ptr<velox::exec::Expr>& expr) {
   using Operation = cudf::ast::operation;
 
-  const auto& name = expr->name();
+  const auto& name =
+      detail::strip_prefix(expr->name(), CudfOptions::getInstance().prefix());
   auto len = expr->inputs().size();
   // Create a simple chain of operations
   auto result = &pushExprToTree(expr->inputs()[0]);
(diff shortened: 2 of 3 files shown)


velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -394,19 +396,19 @@ std::unique_ptr<cudf_velox::CudfHashAggregation::Aggregator> createAggregator(
     bool isGlobal) {
   // Companion function may be count_merge_extract or count_partial or others,
   // so use this to map
-  if (kind.rfind("sum", 0) == 0) {
+  if (boost::algorithm::ends_with(kind, "sum")) {
```
devavret: As the comment explains, we also want to catch kind that starts with "sum". 

Also nit: you don't need a boost algorithm when you can use stl.

devavret (changes requested):
Mostly LGTM, I think this one change in HashAggs might break gluten-libcudf. @jinchengchenghh can weigh in.

zoltan:
definitely don't want to break gluten, but right now it's broken for prestissimo :-)

the other option would be to refactor the code more so we register the functions with the prefix already, or I keep the current code and strip is there as well, that might be easier. let me do a prestissimo run, maybe that would be the easiest course.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -249,8 +252,18 @@ const std::unordered_set<std::string> supportedOps = {
 
 namespace detail {
 
+inline std::string strip_prefix(
```
jinchengchenghh: Move to anonymous namespace, and rename to stripPrefix

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -249,8 +252,18 @@ const std::unordered_set<std::string> supportedOps = {
...
+    const std::string& input,
+    const std::string& prefix) {
+  if (boost::algorithm::starts_with(input, prefix)) {
```
jinchengchenghh: Do not use boost

velox/experimental/cudf/exec/ToCudf.h
```diff
@@ -47,16 +47,36 @@ class CompileState {
...
-  static CudfOptions defaultOptions() {
-    return CudfOptions();
+class CudfOptions {
+ public:
+  static CudfOptions& getInstance() {
+    static CudfOptions instance;
+    return instance;
+  }
+
+  const bool cudfEnabled;
```
jinchengchenghh: Place the member after the functions

velox/experimental/cudf/exec/ToCudf.h
```diff
@@ -47,16 +47,36 @@ class CompileState {
...
+  const std::string cudfMemoryResource;
+
+  void setPrefix(const std::string& prefix) {
```
jinchengchenghh, line 57: Do you set the prefix in runtime?

zoltan: yes, prestissimo sets this (or whoever else is consuming it) for now. same as for registering functions however there the prefix is not kept, just used in runtime.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
zoltan: done.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
zoltan: done.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
zoltan: changed the code to make sure the default behaviour is not changed.

zoltan:
in the new patch I've made sure that the default behaviour is the same as the old one and have incorporated all PR feedback.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -247,10 +248,23 @@ const std::unordered_set<std::string> supportedOps = {
     "substr",
     "like"};
 
+namespace {
+inline std::string stripPrefix(
```
jinchengchenghh: Please combine with other anonymous namespace https://github.com/facebookincubator/velox/pull/13643/files#diff-de07a6a4d087bea3a9ffb4edcd0ad27339278cb71f8cad9e68907365e8d90b93R201

jinchengchenghh: Do not need to use inline

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -247,10 +248,23 @@ const std::unordered_set<std::string> supportedOps = {
...
+  }
+  return input;
+}
+} // anonymous namespace
+
 namespace detail {
 
 bool canBeEvaluated(const std::shared_ptr<velox::exec::Expr>& expr) {
-  const auto& name = expr->name();
+  const auto& name =
```
jinchengchenghh: const auto name

devavret (approved):
Looks like it'll work for gluten now.

zoltan:
@jinchengchenghh and @devavret thanks for the reviews! I have hopefully incorporated all the feedback now into the latest version.

@majetideepak FYI

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -394,19 +395,21 @@ std::unique_ptr<cudf_velox::CudfHashAggregation::Aggregator> createAggregator(
     bool isGlobal) {
   // Companion function may be count_merge_extract or count_partial or others,
   // so use this to map
-  if (kind.rfind("sum", 0) == 0) {
+  auto prefix =
+      facebook::velox::cudf_velox::CudfOptions::getInstance().prefix();
```
majetideepak: nit: `cudf_velox::CudfOptions::getInstance().prefix()` should suffice here

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -394,19 +395,21 @@ std::unique_ptr<cudf_velox::CudfHashAggregation::Aggregator> createAggregator(
...
+  if (kind.rfind(prefix + "sum", 0) == 0) {
```
majetideepak, line 397: why not use `kind.compare(prefix + "sum") == 0`?

velox/experimental/cudf/exec/ToCudf.h
```diff
@@ -47,16 +47,38 @@ class CompileState {
...
+    prefix_ = prefix;
   }
+
+  const std::string& prefix() const {
+    return prefix_;
+  }
+
+  const bool cudfEnabled;
```
majetideepak, line 65: should these be private too?

majetideepak (commented):
@zoltan minor comments.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
zoltan: there could be extra things after the op (gluten does this for example), so we can't compare for exact match.

the original code from @devavret was with rfind, and in the default case where prefix is empty, the code path behaves exactly the same as before the modification.

velox/experimental/cudf/exec/ToCudf.h
zoltan: this way we don't need getters, and since they are const, their value is set during instance creation which is exactly once - so same effect and lifecycle as the command line flags they represent.

the only reason prefix_ is private is because it can change during the lifetime, so it has a setter/getter.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
zoltan: indeed! changed.

majetideepak (approved):

jinchengchenghh (approved):

bdice (approved):

kevinwilfong:
@zoltan It looks like there's a conflict in velox/experimental/cudf/exec/CudfHashAggregation.cpp could you address that, then I'll import it.

zoltan:
@kevinwilfong thanks, it should be good to go now!

PR #13644, opened 2025-06-05
fix: Handle empty matches in regexp_split

xhs7700:
### Summary
This PR addresses a critical bug in the `regexp_split` function that causes a crash when the regex pattern matches empty slices at the end of input strings. The issue stems from an invalid `StringView` construction when position pointer exceeds the string length.

### Problem
When executing queries like:
```sql
SELECT regexp_split(c0, '\s*[a-z]*\s*') 
FROM (VALUES ('1a 2b 14m')) t(c0);
```

The following error occurs:
```
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (-1 vs. 0)
Retriable: False
Expression: len >= 0
Context: Top-level Expression: regexp_split(c0, \s*[a-z]*\s*:VARCHAR)
Function: StringView
File: /Users/xxx/velox/./velox/type/StringView.h
```

**Root Cause:** When empty slices are matched at the end of input strings, the position pointer (`pos`) gets incremented beyond the string's size before constructing a `StringView`, leading to invalid memory access in:
```cpp
StringView(string.data() + pos, string.size() - pos)
```

### Solution
1. **Bounds Check before StringView Construction**: Added validation before creating `StringView` instances to prevent invalid memory access.
2. **Align Behavior with Presto for Empty Matches**: Introduced a `lastEnd` pointer to track the end of the last match. This matches Presto's logic and produces more intuitive results.
3. **Update Documentation**: Clarified the function's behavior in the documentation.

### DIFF
--- a/velox/docs/functions/presto/regexp.rst
+++ b/velox/docs/functions/presto/regexp.rst
@@ -114,3 +114,8 @@ limited to 20 different expressions per instance and thread of execution.
     array. Trailing empty strings are preserved::
 
         SELECT regexp_split('1a 2b 14m', '\s*[a-z]+\s*'); -- [1, 2, 14, ]
+
+    Note: Similar with Presto's implementation, When the pattern matches an empty string,
+    the function advances by one character and includes the skipped character in the result::
+
+        SELECT regexp_split('1a 2b 14m', ''); -- [, 1, 'a', ' ', 2, 'b', ' ', 1, 4, 'm', ]
--- a/velox/docs/functions/presto/regexp.rst
+++ b/velox/docs/functions/presto/regexp.rst
@@ -115,7 +115,7 @@ limited to 20 different expressions per instance and thread of execution.
 
         SELECT regexp_split('1a 2b 14m', '\s*[a-z]+\s*'); -- [1, 2, 14, ]
 
-    Note: Similar with Presto's implementation, When the pattern matches an empty string,
-    the function advances by one character and includes the skipped character in the result::
+    Note: When the regular expression ``pattern`` matches an empty string, the result array contains an
+    empty string, a single-character string for each character in the input string, and another empty string::
 
         SELECT regexp_split('1a 2b 14m', ''); -- [, 1, 'a', ' ', 2, 'b', ' ', 1, 4, 'm', ]
--- a/velox/functions/lib/Re2Functions.h
+++ b/velox/functions/lib/Re2Functions.h
@@ -409,6 +409,7 @@ struct Re2RegexpSplit {
     const auto re2String = re2::StringPiece(string.data(), string.size());
 
     size_t pos = 0;
+    size_t lastEnd = 0;
     const char* start = string.data();
 
     re2::StringPiece subMatches[1];
@@ -423,18 +424,24 @@ struct Re2RegexpSplit {
       const auto offset = fullMatch.data() - start;
       const auto size = fullMatch.size();
 
-      out.add_item().setNoCopy(StringView(string.data() + pos, offset - pos));
+      out.add_item().setNoCopy(
+          StringView(string.data() + lastEnd, offset - lastEnd));
 
-      pos = offset + size;
+      lastEnd = offset + size;
       if (UNLIKELY(size == 0)) {
-        ++pos;
+        pos = lastEnd + 1;
+      } else {
+        pos = lastEnd;
       }
     }
 
-    auto lastSlice = LIKELY(pos <= string.size())
-        ? StringView(string.data() + pos, string.size() - pos)
-        : StringView(nullptr, 0);
-    out.add_item().setNoCopy(lastSlice);
+    if (LIKELY(pos <= string.size())) {
+      out.add_item().setNoCopy(
+          StringView(string.data() + pos, string.size() - pos));
+    } else {
+      static const StringView kEmptyString(nullptr, 0);
+      out.add_item().setNoCopy(kEmptyString);
+    }
   }
 
  private:
--- a/velox/functions/lib/Re2Functions.h
+++ b/velox/functions/lib/Re2Functions.h
@@ -431,8 +431,10 @@ struct Re2RegexpSplit {
       }
     }
 
-    out.add_item().setNoCopy(
-        StringView(string.data() + pos, string.size() - pos));
+    auto lastSlice = LIKELY(pos <= string.size())
+        ? StringView(string.data() + pos, string.size() - pos)
+        : StringView(nullptr, 0);
+    out.add_item().setNoCopy(lastSlice);
   }
 
  private:
--- a/velox/functions/lib/tests/Re2FunctionsTest.cpp
+++ b/velox/functions/lib/tests/Re2FunctionsTest.cpp
@@ -21,6 +21,7 @@
 #include <functional>
 #include <optional>
 #include <string>
+#include <vector>
 
 #include "velox/common/base/VeloxException.h"
 #include "velox/common/base/tests/GTestUtils.h"
@@ -1560,6 +1561,14 @@ TEST_F(Re2FunctionsTest, split) {
       {"", "a", "b", ""},
       {""},
       {"a", "b"},
+  });
+  assertEqualVectors(expected, result);
+  result = evaluate("regexp_split(c0, '\\s*[a-z]*\\s*')", input);
+  expected = makeArrayVector<std::string>({
+      std::vector<std::string>(9, ""),
+      std::vector<std::string>(8, ""),
+      std::vector<std::string>(2, ""),
+      std::vector<std::string>(7, ""),
   });
   assertEqualVectors(expected, result);
 }
--- a/velox/functions/lib/tests/Re2FunctionsTest.cpp
+++ b/velox/functions/lib/tests/Re2FunctionsTest.cpp
@@ -1563,12 +1563,22 @@ TEST_F(Re2FunctionsTest, split) {
       {"a", "b"},
   });
   assertEqualVectors(expected, result);
+
+  result = evaluate("regexp_split(c0, '')", input);
+  expected = makeArrayVector<std::string>({
+      {"", "1", "a", " ", "2", "b", " ", "1", "4", "m", ""},
+      {"", "1", "a", " ", "2", "b", " ", "1", "4", ""},
+      {"", ""},
+      {"", "a", "1", "2", "3", "b", ""},
+  });
+  assertEqualVectors(expected, result);
+
   result = evaluate("regexp_split(c0, '\\s*[a-z]*\\s*')", input);
   expected = makeArrayVector<std::string>({
-      std::vector<std::string>(9, ""),
-      std::vector<std::string>(8, ""),
-      std::vector<std::string>(2, ""),
-      std::vector<std::string>(7, ""),
+      {"", "1", "", "2", "", "1", "4", "", ""},
+      {"", "1", "", "2", "", "1", "4", ""},
+      {"", ""},
+      {"", "", "1", "2", "3", "", ""},
   });
   assertEqualVectors(expected, result);
 }
--- a/velox/functions/lib/tests/Re2FunctionsTest.cpp
+++ b/velox/functions/lib/tests/Re2FunctionsTest.cpp
@@ -1564,6 +1564,7 @@ TEST_F(Re2FunctionsTest, split) {
   });
   assertEqualVectors(expected, result);
 
+  // Test for empty matches
   result = evaluate("regexp_split(c0, '')", input);
   expected = makeArrayVector<std::string>({
       {"", "1", "a", " ", "2", "b", " ", "1", "4", "m", ""},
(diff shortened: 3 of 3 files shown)


xhs7700:
@mbasmanova Could you please review this PR when you have a moment? Thanks for your help!

mbasmanova:
@xhs7700 Thank you for reporting and fixing this issue.

@amitkdutta @kgpai  Amit, Krishna, I just confirmed that this issue is present in our deployment as well. However, not sure if it is present in Java.

```
presto:di> SELECT regexp_split(c0, '\s*[a-z]*\s*')
        -> FROM (VALUES ('1a 2b 14m')) t(c0);
Query 20250605_125639_58711_2gduw failed: len >= 0 (-1 vs. 0) Top-level Expression: presto.default.regexp_split(field, \s*[a-z]*\s*:VARCHAR). Input data: /tmp/velox_vector_45uNVS. SQL expression: /tmp/velox_sql_RvDO73. All SQL expressions: /tmp/velox_allExprSql_uBtbQU.  Operator: FilterProject[4] 1

presto:di> SELECT regexp_split('1a 2b 14m', '\s*[a-z]*\s*') ;
Query 20250605_125709_82769_5tkf9 failed: len >= 0 (-1 vs. 0) presto.default.regexp_split(1a 2b 14m:VARCHAR, \s*[a-z]*\s*:VARCHAR)
```

velox/functions/lib/tests/Re2FunctionsTest.cpp
```diff
@@ -1562,6 +1563,14 @@ TEST_F(Re2FunctionsTest, split) {
       {"a", "b"},
   });
   assertEqualVectors(expected, result);
+  result = evaluate("regexp_split(c0, '\\s*[a-z]*\\s*')", input);
```
mbasmanova, line 1578: I'm seeing this issue is a simpler case as well:

```
presto:di> SELECT regexp_split('1a 2b 14m', '');
Query 20250605_130112_83321_5tkf9 failed: len >= 0 (-1 vs. 0) presto.default.regexp_split(1a 2b 14m:VARCHAR, :VARCHAR)
```

Wondering if it makes sense to replace with that case (or add that case).

It is a bit hard to read these regexes and figure out what they "mean". It would be helpful to add a comment. In particular it is hard to see the difference between L1548 and L1566 (+ vs. * after [a-z]).

Please, add an empty line before L1566.

velox/functions/lib/Re2Functions.h
```diff
@@ -431,8 +431,10 @@ struct Re2RegexpSplit {
       }
     }
 
-    out.add_item().setNoCopy(
-        StringView(string.data() + pos, string.size() - pos));
+    auto lastSlice = LIKELY(pos <= string.size())
```
mbasmanova: I ternary operator here is a bit hard to read. Would you replace with if-else?

```
if (LIKELY(pos <= string.size())) {
   out.add_item().setNoCopy(StringView(string.data() + pos, string.size() - pos)));
} else {
   static const StringView kEmptyString(nullptr, 0);
   out.add_item().setNoCopy(kEmptyString);
}
```

mbasmanova (commented):
@xhs7700 Looks good % a few nits.

@rui-mo Rui, I remember there was a similar fix earlier, but can't find it easily. Do you remember by any chance?

mbasmanova:
Here is relevant PR: #12305

CC: @HolyLow @Yuhta

HolyLow:
@mbasmanova Seems like an similar issue, if not the same. The former pr is not merged as there are some disagreements, and it seems the bug is encountered by multi users and cases. Maybe we should merge it this time.

mbasmanova:
I believe this change affects both Presto and Spark. 

@rui-mo Rui, would you check in that change is desirable for Spark or whether we need to keep it Presto-only?

rui-mo:
@mbasmanova Thanks for the remainder! Spark has a separate implementation for the split function and handles the empty delimiter specially, so this change would not affect Spark. After testing the Spark implementation, I find `split('1a 2b 14m', '')` yields the right result, but `split('1a 2b 14m', '\\s*[a-z]*\\s*')` does not match Spark and might need to be fixed.

https://github.com/facebookincubator/velox/blob/d02351911f5f8ca2b00242d6010859baaf513813/velox/functions/sparksql/Split.h#L83-L88

amitkdutta:
Agreed with @mbasmanova and @rui-mo. We should merge this time. It can cause severe mismatch between C++ and Java.

mbasmanova:
@rui-mo Rui, thank you for confirming that this change won't affect Spark and therefore would be safe to land as long it provides correct Presto behavior.

It might be helpful to move this code our of the shared lib into presto_sql to avoid future confusion.

CC: @kgpai @amitkdutta

mbasmanova:
@amitkdutta Amit, do you have easy access to Presto Java instance? If so, can you show the output of the following queries?

- ```SELECT regexp_split('1a 2b 14m', '\s*[a-z]*\s*') ;```
- ```SELECT regexp_split('1a 2b 14m', '');```

amitkdutta:
> * SELECT regexp_split('1a 2b 14m', '');

Here in JONI (java)

```
presto:di> SELECT regexp_split('1a 2b 14m', '\s*[a-z]*\s*') ;
         _col0          
------------------------
 [, 1, , 2, , 1, 4, , ] 
(1 row)
```


```
presto:di> SELECT regexp_split('1a 2b 14m', '');
              _col0              
---------------------------------
 [, 1, a,  , 2, b,  , 1, 4, m, ] 
(1 row)
```

mbasmanova:
@amitkdutta Thank you for Amit.

@xhs7700 Would you take a look at Presto Java results Amit shared... it looks like Velox implementation no longer throws, but results are different from Presto.

xhs7700:
@mbasmanova Thank you for the feedback! The discrepancy with Presto stems from an implementation difference. Unlike Velox's current logic, Presto's `regexp_split` uses an additional pointer `lastEnd` to track the end of the last match, which avoids advancing positions on empty matches (see [implementation details](https://github.com/prestodb/presto/blob/1d46dc8a3fcfbb03df3559c71b7c56a792be59a5/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/JoniRegexpFunctions.java?spm=a2ty_o01.29997173.0.0.711ec921Pf8wgL#L299) ).

I believe Presto's handling of empty matches produces more reasonable output in such cases. I've revised the code in my second commit to align with Presto's behavior by adopting this approach. Let me know if you'd like further clarification or adjustments! Thanks again for your time and suggestions.

velox/functions/lib/tests/Re2FunctionsTest.cpp
xhs7700: I have added the case of `regexp_split('1a 2b 14m', '')`, comments and an empty line. Thanks for your suggestion.

velox/functions/lib/Re2Functions.h
xhs7700: The logic has been updated to use if-else. Thanks for your suggestion.

mbasmanova:
@xhs7700 Thank you for checking Presto's behavior and adjusting. Would you also update the documentation to clarify this behavior?

https://facebookincubator.github.io/velox/functions/presto/regexp.html#regexp_split

mbasmanova:
> Add bounds checking before StringView construction

@xhs7700 Would you also update PR description to describe changes more fully?

mbasmanova (commented):
@xhs7700 Code changes look good to me. Please, update the documentation and PR description and I'll approve this PR. Thanks.

mbasmanova:
@HolyLow Any chance you could also review the changes? I'm going to close #12305 in favor of this one. Sorry we didn't follow through on that PR in a timely manner back in Feb.

mbasmanova:
Created #13673 to move this code from lib to prestosql. 

@xhs7700 Any chance you could help with that?

xhs7700:
> @xhs7700 Code changes look good to me. Please, update the documentation and PR description and I'll approve this PR. Thanks.

@mbasmanova The documentation and PR description have been updated. Thanks for your feedback.

xhs7700:
> Created #13673 to move this code from lib to prestosql.
> 
> @xhs7700 Any chance you could help with that?

@mbasmanova Thank you for the invitation. Unfortunately, I’m not very familiar with Velox’s overall structure and worry I might not be the best fit for this task. Let me know if there’s anything else I can help with!

velox/docs/functions/presto/regexp.rst
```diff
@@ -114,3 +114,8 @@ limited to 20 different expressions per instance and thread of execution.
     array. Trailing empty strings are preserved::
 
         SELECT regexp_split('1a 2b 14m', '\s*[a-z]+\s*'); -- [1, 2, 14, ]
+
+    Note: Similar with Presto's implementation, When the pattern matches an empty string,
```
mbasmanova: > Note: Similar with Presto's implementation

This is redundant. All Presto functions must match Presto behavior 100%. Let's change to 

> the function advances by one character and includes the skipped character in the result

I wonder if users would understand this description. 

How about,

```
When the pattern matches an empty string, the result array contains an empty string, a single-character string for each character in the input string, and another empty string.
```

Also, I would expect regexp_split(.., '') to match split(.., ''), but I see split returning a different result. Wondering if this is a bug in Presto or in Velox implementation of split or regexp_split. CC: @amitkdutta 

```
presto:di> select split('1a 2b 14m', '');
             _col0
-------------------------------
 [1, a,  , 2, b,  , 1, 4, m, ]
(1 row)
```

@xhs7700 Would you check Presto Java results for regexp_split and split with empty delimeter?

amitkdutta: @mbasmanova The behavior seems different in Presto java split and regexp_split

```
presto:di> select split('1a 2b 14m', '');                                                                                                                                    
             _col0                                                                                                                                                           
-------------------------------                                                                                                                                              
 [1, a,  , 2, b,  , 1, 4, m, ]                                                                                                                                               
(1 row)
presto:di> select cardinality(split('1a 2b 14m', ''));
 _col0                
-------
    10 
(1 row

presto:di> select regexp_split('1a 2b 14m', '');
              _col0              
---------------------------------
 [, 1, a,  , 2, b,  , 1, 4, m, ] 
(1 row)
presto:di> select cardinality(regexp_split('1a 2b 14m', ''));
 _col0                
-------
    11 
(1 row)
```

`regexp_split` returns 1 extra item.

mbasmanova: @amitkdutta Should we open an issue about this in Presto? I imagine such discrepancy would be confusing to users, no?

xhs7700: > > Note: Similar with Presto's implementation
> 
> This is redundant. All Presto functions must match Presto behavior 100%. Let's change to
> 
> > the function advances by one character and includes the skipped character in the result
> 
> I wonder if users would understand this description.
> 
> How about,
> 
> ```
> When the pattern matches an empty string, the result array contains an empty string, a single-character string for each character in the input string, and another empty string.
> ```
> 
> Also, I would expect regexp_split(.., '') to match split(.., ''), but I see split returning a different result. Wondering if this is a bug in Presto or in Velox implementation of split or regexp_split. CC: @amitkdutta
> 
> ```
> presto:di> select split('1a 2b 14m', '');
>              _col0
> -------------------------------
>  [1, a,  , 2, b,  , 1, 4, m, ]
> (1 row)
> ```
> 
> @xhs7700 Would you check Presto Java results for regexp_split and split with empty delimeter?

The description has been updated in my latest commit, according to your kind suggestion. Thanks.

velox/functions/lib/tests/Re2FunctionsTest.cpp
```diff
@@ -1562,6 +1563,26 @@ TEST_F(Re2FunctionsTest, split) {
...
+      {"", "1", "a", " ", "2", "b", " ", "1", "4", "m", ""},
+      {"", "1", "a", " ", "2", "b", " ", "1", "4", ""},
+      {"", ""},
+      {"", "a", "1", "2", "3", "b", ""},
+  });
+  assertEqualVectors(expected, result);
+
+  // Test for another case of empty matches
+  result = evaluate("regexp_split(c0, '\\s*[a-z]*\\s*')", input);
+  expected = makeArrayVector<std::string>({
```
mbasmanova, line 1579: This is the same as in L1569, right? Let's not repeat and make it clear that both patterns produce same results.

mbasmanova (approved):
Looks great. Thank you for the fix.

mbasmanova:
@xhs7700 Would you please rebase this PR so I can try merging?

xhs7700:
> @xhs7700 Would you please rebase this PR so I can try merging?

Rebased successfully without conflicts.

mbasmanova:
> > @xhs7700 Would you please rebase this PR so I can try merging?
> 
> Rebased successfully without conflicts.

Thanks!

mbasmanova:
@xhs7700 CI is red. Would you take a look?

Issue #13646, opened 2025-06-05
Lambda function calls with non-deterministic lambdas are marked deterministic


mbasmanova:
### Bug description

An expression like "transform(c0, x -> x + rand())" is non-deterministic, but Expression engine considers it deterministic.

CC: @bikramSingh91 @aditi-pandit @kagamiori 

Repro:

```
TEST_F(ExprTest, isDeterministic) {
  auto isDeterministic = [&](const std::string& sql) {
    SCOPED_TRACE(sql);
    auto exprSet = compileExpression(sql, ROW({"c0"}, {ARRAY(DOUBLE())}));
    return exprSet->expr(0)->isDeterministic();
  };

  EXPECT_TRUE(isDeterministic("cardinality(c0) + 5"));
  EXPECT_TRUE(isDeterministic("transform(c0, x -> x + 5.0)"));
  EXPECT_TRUE(isDeterministic("filter(c0, x -> (x < 5.0))"));

  EXPECT_FALSE(isDeterministic("rand()"));
  EXPECT_FALSE(isDeterministic("c0[1] + rand()"));
  EXPECT_FALSE(isDeterministic("transform(c0, x -> x + rand())"));
  EXPECT_FALSE(isDeterministic("filter(c0, x -> (x < rand()))"));
}

[ RUN      ] ExprTest.isDeterministic
fbcode/velox/expression/tests/ExprTest.cpp:5013: Failure
Value of: isDeterministic("transform(c0, x -> x + rand())")
  Actual: true
Expected: false

fbcode/velox/expression/tests/ExprTest.cpp:5014: Failure
Value of: isDeterministic("filter(c0, x -> (x < rand()))")
  Actual: true
Expected: false
```

### System information

n/a

### Relevant logs

```bash

```

PR #13647, opened 2025-06-05
fix: Fix Expr::isDeterministic for lambda functions

mbasmanova:
Summary: 

An expression like "transform(c0, x -> x + rand())" is non-deterministic, but Expression engine used to consider it deterministic.

Fixes https://github.com/facebookincubator/velox/issues/13646

Differential Revision: D76041612

### DIFF
--- a/velox/expression/Expr.cpp
+++ b/velox/expression/Expr.cpp
@@ -1192,20 +1192,36 @@ void Expr::evalWithMemo(
   VectorPtr base;
   distinctFields_[0]->evalSpecialForm(rows, context, base);
 
+  // evalWithNulls may throw an exception. If this happens during constant
+  // folding, the exception is suppressed and the Expr object may be reused.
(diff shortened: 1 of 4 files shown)


velox/expression/Expr.cpp
```diff
@@ -234,7 +235,11 @@ void Expr::computeMetadata() {
   // a special form, and all its inputs are also deterministic.
   deterministic_ = isCurrentFunctionDeterministic();
```
bikramSingh91, line 236: should we fix isCurrentFunctionDeterministic() for LambdaExpr instead? We can then avoid special handling of LambdaExpr in L238 as the input LambdaExpr would then have its deterministic_ set correctly already

mbasmanova: @bikramSingh91 Discussed offline and decided to keep this PR short. Bikram may follow-up with more changes.

PR #13651, opened 2025-06-05
feat(noisy_count): Add new unit test to noisy_count_gaussian(col, noise_scale)

oliver6782:
Summary:
**Summary**
This diff adds more unit tests so that the noisy function is more robust.

**Changes**
* Added tests that aggregate on non-scalar inputs such as array and map.
* Added edge case where the input is empty.

Differential Revision: D75916564


amitkdutta (approved):
Thnkas @oliver6782

PR #13653, opened 2025-06-05
feat(noisy_count): Implement noisy_count_gaussian(col, noise_scale, random_seed)

oliver6782:
Summary:
### Summary
This diff introduces random seed to the `noisy_count_gaussian` function  to seed the random number generator.

### Changes
* Added a new test cases to `NoisyCountGaussianAggregationTest.cpp`.
* Updated the `noisy_count_gaussian` function documentation to include the `random_seed` parameter.
* Modified the `NoisyCountGaussianAggregate.cpp` file to decode the `random_seed` argument and use it to seed the random number generator.

### Impact
This diff enhances the `noisy_count_gaussian` function by providing an optional `random_seed` parameter, allowing users to reproduce the same noisy count results. The new test case ensures that the function works correctly with multiple aggregates and groups.

Differential Revision: D75926651


amitkdutta:
@oliver6782  Looks like this branch has conflict.

amitkdutta (approved):
Thanks @oliver6782

Issue #13648, opened 2025-06-05
parseTypeSignature fails to parse named structs with parentheses in the field names


mbasmanova:
### Bug description

To reproduce:

velox/expression/signature_parser/tests/ParseTypeSignatureTest.cpp

```
TEST_F(ParseTypeSignatureTest, row) {
  parseTypeSignature("row(\"a (b)\" INTEGER)");
...

Reason: Failed to parse type signature [row("a (b)" INTEGER)]: syntax error, unexpected WORD, expecting RPAREN or COMMA

File: buck-out/v2/gen/fbcode/ba5e74bcde559f36/velox/expression/signature_parser/__velox_signature_parser__SignatureParser.yy__/out/SignatureParser.yy.cc
Line: 1466
```

CC: @pedroerp @majetideepak @bikramSingh91 

### System information

n/a

### Relevant logs

```bash

```

mbasmanova:
@majetideepak Deepak, I believe this is an issue with the Flex / Bison grammar we use for signature parser.

Would you have any ideas about how to fix this?

majetideepak:
@mbasmanova I will take a look.

majetideepak:
@mbasmanova, here is the fix https://github.com/facebookincubator/velox/pull/13658

mbasmanova:
@majetideepak Deepak, thank you for the super quick fix!

mbasmanova:
@majetideepak Deepak, looks like there are more issues. The following signature also fails: `row("a {b}" INTEGER)`.

majetideepak:
@mbasmanova I will add support for that as well.

majetideepak:
@mbasmanova is `row("a [b]" INTEGER)` allowed as well?

mbasmanova:
Thank you, Deepak.

> is row("a [b]" INTEGER) allowed as well?

I believe we want any name that can be used as a top-level column name to also be used a struct field name.

majetideepak:
> I believe we want any name that can be used as a top-level column name to also be used a struct field name.

I see this from Presto https://github.com/prestodb/presto/blob/master/presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4#L933

mbasmanova:
@majetideepak Is the right way to interpret Presto's grammar to say that any character other than double quote is allowed between double quotes?

```
QUOTED_IDENTIFIER
    : '"' ( ~'"' | '""' )* '"'
    ;
```

[...]

PR #13658, opened 2025-06-05
feat: Add support for names with parentheses in parseTypeSignature

majetideepak:
Fixes: https://github.com/facebookincubator/velox/issues/13648

### DIFF
--- a/velox/expression/signature_parser/SignatureParser.ll
+++ b/velox/expression/signature_parser/SignatureParser.ll
@@ -34,7 +34,7 @@ Y   [Y|y]
 Z   [Z|z]
 
 WORD              ([[:alnum:]_]*)
-QUOTED_ID         (['"'][[:alnum:][:space:]_]*['"'])
+QUOTED_ID         (['"'][[:alnum:][:space:]_()]*['"'])
 ROW               (ROW|STRUCT)
 
 %%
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
@majetideepak Wow! This was quick!

Thank you so much, Deepak!!!

majetideepak:
@yunwu52 can you take a look? I will fix the identifier in a separate PR.

yunwu52 (approved):

yunwu52:
@majetideepak Thanks for the super quick change!

majetideepak:
@yunwu52, @mbasmanova  I observed that the Presto fragment does not escape the double quote when sending to the worker. This means, Velox should not unescape or error on a single double quote. The Presto parser takes care of the validation. Velox must accept as-is.
`"dataColumns":[{"name":"c\"h","type":"int"}],`

mbasmanova:
@majetideepak Deepak, the issue here comes from using named struct types in the function signatures, i.e.

```
exec::FunctionSignatureBuilder()
            .returnType("<name> INTEGER")
            ...
```

Before your changes, names that used parentheses, braces, etc. would generate an error. With your changes we expect to be able to create such signatures successfully.

Overall use case is a query engine used to read and pre-process data for AI training.

majetideepak:
@mbasmanova thanks for the clarification. This change should help.

PR #13659, opened 2025-06-05
feat: Add dry run trace mode for crash debugging

xiaoxmeng:
Summary: Add dry run mode for query trace collection for crash debugging so we are able to collect the input that triggers the crash

Differential Revision: D76019158

### DIFF
--- a/velox/connectors/Connector.cpp
+++ b/velox/connectors/Connector.cpp
@@ -173,4 +173,18 @@ folly::dynamic ConnectorTableHandle::serializeBase(
 folly::dynamic ConnectorTableHandle::serialize() const {
   return serializeBase("ConnectorTableHandle");
 }
+
+// static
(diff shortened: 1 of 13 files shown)


tanjialiang (approved):

velox/connectors/Connector.cpp
```diff
@@ -173,4 +173,18 @@ folly::dynamic ConnectorTableHandle::serializeBase(
 folly::dynamic ConnectorTableHandle::serialize() const {
   return serializeBase("ConnectorTableHandle");
 }
+
+// static
+ConnectorTableHandlePtr ConnectorTableHandle::create(
```
mbasmanova, line 178: @xiaoxmeng @majetideepak This doesn't seem right. Why do we need this?

mbasmanova: Removing in #14679

Issue #13663, opened 2025-06-05
Linux Release with adapters build broken.


kgpai:
### Problem description

The adapters release build is broken. Might be due to folly upgrade. 

```
[2157/2841] Linking CXX executable velox/buffer/tests/velox_buffer_test
FAILED: velox/buffer/tests/velox_buffer_test 

facebook::velox::functions::remote::RemoteFunctionRequest::serializedSizeZC<apache::thrift::BinaryProtocolWriter>(apache::thrift::BinaryProtocolWriter const*) const'
collect2: error: ld returned 1 exit status
```

### System information

https://github.com/facebookincubator/velox/actions/runs/15477820175/job/43577448585?pr=13642

### CMake log

```bash

```

kgpai:
cc: @majetideepak @assignUser

assignUser:
Yeah, that's my suspicion as well, less folly directly but that PR bumped the entire fb_os collection to the new version, so likely a change in fbthrift?

kgpai:
@assignUser Unfortunately these things go hand in hand and we might have even more issues if we peg some on one version and some on other.

kgpai:
I am going to revert the folly change then, and I will also try and build locally to confirm this.

PR #13664, opened 2025-06-06
fix: Down grade folly to v2025.04.28.00

kgpai:
Summary:
Reverting the folly upgrade due to issue : https://github.com/facebookincubator/velox/issues/13663 . 
Note the build will still fail since the docker images will only be upgraded on land. 

Original commit changeset: 9eb111fefc57

Original Phabricator Diff: D75487583

Differential Revision: D76092968


assignUser (approved):
Thanks for the quick revert!

kKPulla (approved):

PR #13665, opened 2025-06-06
build(ci): Reinstall dependencies when setup scripts have been touched

assignUser:
We have had issues with updates to dependencies breaking builds once the new image is generated, because the CI would still use the old image. 

This PR adds code to install the dependencies again if the relevant files have been touched in the PR.


assignUser:
This would also apply for the fuzzer build but we would need to install the deps on all 15 or so run jobs as well, (or collect dependencies and move them over too, or link purely statically) which seems a bit overkill at the moment.

majetideepak (approved):

majetideepak:
@assignUser Can you update the FB_OS_DEPENDENCIES and confirm we are seeing the build failure we saw yesterday?

assignUser:
@majetideepak good idea! I have reverted the revert xD

assignUser:
@majetideepak yes it shows up: https://github.com/facebookincubator/velox/actions/runs/15500437053/job/43649229526?pr=13665#step:8:6817 going to remove that commit again

Issue #12294, opened 2025-02-10
NestedLoopJoin is significantly  slower than vanilla spark


lifulong:
### Bug description

same data and same stage (nestdloopjoin operator), vanilla spark cost 3min per task, where gluten (velox) cost 1h+ per task 

![Image](https://github.com/user-attachments/assets/d52985a9-2bb8-481a-9e84-a1b21eb097ef)

the above graph is flame graph of velox

### System information

Velox System Info v0.0.2
Commit: 976a5b72a3a068cd1c70cc92ab64cfedae3649a1
CMake Version: 3.28.3
System: Linux-6.10.14-linuxkit
Arch: x86_64
C++ Compiler: /opt/rh/devtoolset-11/root/usr/bin/c++
C++ Compiler Version: 11.2.1
C Compiler: /opt/rh/devtoolset-11/root/usr/bin/gcc
C Compiler Version: 11.2.1
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib64/python3.6/site-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

pedroerp:
Can you share more details about the query? e.g is this an equi-join? How large are the rows?

pedroerp:
Cc: @Yuhta @xiaoxmeng @mbasmanova @aditi-pandit

Yuhta:
Are you building Velox without `NDEBUG`?

lifulong:
> Thanks for reporting this.
> 
> Can you share more details about the query? e.g is this an equi-join? How large are the rows?

![Image](https://github.com/user-attachments/assets/d9c4b134-a1a5-45f2-a296-69a0d52db69a)
sql query join condition like this， i can't supply whole sql code, because it' our company production env code
join probe side has 150 billion records， build side has 92 records, has supply this info in the issue

lifulong:
> Are you building Velox without `NDEBUG`?

i will retry without NDEBUG

lifulong:
![Image](https://github.com/user-attachments/assets/da9da657-695d-4e98-9a1d-09d97c68b705)

new flame graph with NDEBUG defined

Yuhta:
It would be nice to get the flame graph with Release build, otherwise it is not useful

lifulong:
> It would be nice to get the flame graph with Release build, otherwise it is not useful

[...]

PR #10832, opened 2024-08-26
Fix LEFT and ANTI joins to preserve probe order

Yuhta:
Summary:
Currently for LEFT and ANTI joins, when there is a carryover missing
row from previous batch, we keep it and add it at end of current batch if there
is capacity, which breaks the order of probe side rows.

To fix this we simplify the logic of `NoMatchDetector` to emit rows in order.
As a result of that, the carryover row will be added as soon as we can decide it
needs to be added to output; this means we need to make one row extra space on
output buffers for the carryover case.  In the same case we also need to use
temporary buffers to avoid data being overwritten before we read them.

Differential Revision: D61795406

PR #13666, opened 2025-06-06
fix: Allow NestedLoopJoin to output rows not preserving input order if optimizer does not require it

lifulong:
Fixes: https://github.com/facebookincubator/velox/issues/12294

optimize NestedLoopJoin with conditions by build vector size, do not  always do probe one by one
new execute time is still long than vanilla spark, but has been speed 10x+ compare with before

add config kPreferredNLJOutputProbeOrder, default is true, deal with probe one by one, but for  scenarios where preserving the order of the probe side is not required, we can set kPreferredNLJOutputProbeOrder to false for speed up

i have add some unit tests for new optimize, which is join with condition for one build row, join with condition for one build vector, join with condition for multi build vector

execute time before optimize (use from spark+gluten)
image

execute time after optimize (use from spark+gluten)
image

execute time use vanilla spark
image

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -208,6 +208,11 @@ class QueryConfig {
   static constexpr const char* kPreferredOutputBatchRows =
       "preferred_output_batch_rows";
 
+  /// Preferred produce output same order with probe data at NestedLoopJoin
+  /// operator
+  static constexpr const char* kPreferredNLJOutputProbeOrder =
+      "preferred_nlj_output_probe_order";
+
   /// Max number of rows that could be return by operators from
   /// Operator::getOutput. It is used when an estimate of average row size is
   /// known and kPreferredOutputBatchBytes is used to compute the number of
@@ -744,6 +749,11 @@ class QueryConfig {
     const uint32_t batchRows = get<uint32_t>(kPreferredOutputBatchRows, 1024);
     VELOX_USER_CHECK_LE(batchRows, std::numeric_limits<vector_size_t>::max());
     return batchRows;
+  }
+
+  bool preferredNLJOutputProbeOrder() const {
+    const bool probeOrder = get<bool>(kPreferredNLJOutputProbeOrder, true);
+    return probeOrder;
   }
 
   vector_size_t maxOutputBatchRows() const {
--- a/velox/exec/NestedLoopJoinProbe.cpp
+++ b/velox/exec/NestedLoopJoinProbe.cpp
@@ -63,6 +63,7 @@ NestedLoopJoinProbe::NestedLoopJoinProbe(
   auto buildType = joinNode_->sources()[1]->outputType();
   identityProjections_ = extractProjections(probeType, outputType_);
   buildProjections_ = extractProjections(buildType, outputType_);
+  outputProbeOrder_ = driverCtx->queryConfig().preferredNLJOutputProbeOrder();
 }
 
 void NestedLoopJoinProbe::initialize() {
@@ -183,6 +184,9 @@ void NestedLoopJoinProbe::addInput(RowVectorPtr input) {
     probeSideEmpty_ = false;
   }
   VELOX_CHECK_EQ(buildIndex_, 0);
+  if (needsProbeMismatch(joinType_)) {
+    probeMatched_.resizeFill(input_->size(), false);
+  }
 }
 
 void NestedLoopJoinProbe::noMoreInput() {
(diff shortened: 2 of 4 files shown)


velox/core/QueryConfig.h
```diff
@@ -208,6 +208,11 @@ class QueryConfig {
   static constexpr const char* kPreferredOutputBatchRows =
       "preferred_output_batch_rows";
 
+  /// Preferred produce output same order with probe data at NestedLoopJoin
+  /// operator
+  static constexpr const char* kPreferredNLJOutputProbeOrder =
```
Yuhta, line 213: This should not be a query config but should be some property on plan node.  Whether the order on probe side is preserved would affect how optimizer generates the plan so it is part of the plan property, and should not be toggled by user at runtime (could generate wrong result if the plan assume the order is preserved and user toggle it off).

lifulong: Do you mean that the optimizer determines the node properties and decides whether to preserve the probe order? I'm not familiar with the background of previously preserving the probe order. In which scenarios is it necessary to ensure the output results are consistent with the probe order

mbasmanova:
@lifulong I find PR description hard to understand. I tried reading the linked issue, but there are so many comments that it is also hard to follow. Would it be possible to summary the issue and the proposed solution a bit more clearly?

velox/core/QueryConfig.h
mbasmanova: @lifulong Currently, there is a guarantee that all INNER, LEFT and ANTI joins preserve the order of the probe-side rows. As @Yuhta mentioned we cannot drop that guarantee with a config.

fbcode/velox/core/PlanNode.h

```
  bool isPreservingProbeOrder() const {
    return isInnerJoin() || isLeftJoin() || isAntiJoin();
  }
```

lifulong: @mbasmanova  In many cases, as long as the result is correct, is maintaining the output order consistent with the probe side necessary in all user case ？Can users be given the option to enable/disable order-preserving requirements ?

mbasmanova: @lifulong The user here is the Optimizer, not the actual query operator. The optimizer can control the behavior via PlanNode properties. Hence, to allow the Optimizer to specify whether the order of probe side rows needs to be preserved we would need to introduce a property in the XxxJoinNodes.

lifulong:
> @lifulong I find PR description hard to understand. I tried reading the linked issue, but there are so many comments that it is also hard to follow. Would it be possible to summary the issue and the proposed solution a bit more clearly?
@mbasmanova Here’s a summary:

While using Spark+Gluten with Velox, I noticed that the performance of NestedLoopJoin is significantly worse compared to vanilla Spark. Upon investigation, I found that when the NestedLoopJoin operator has filtering conditions, Velox currently processes one probe-side record at a time—joining it with the build-side vector, generating temporary results, applying the filter, and then appending the results to the output. This execution strategy is inefficient.

In cases where the build-side vector is small, performance could be improved by processing a batch of probe-side records at once, based on the outputBatchSize and the size of the build-side vector. This optimization is similar to the approach taken in the isCrossJoin()=true branch in the code, and actual testing confirms that it delivers much better efficiency.

However, processing a batch of probe-side records at once may not maintain the output order consistent with the probe-side sequence in cases like ​left join​ or ​full join

mbasmanova:
@lifulong 

It happens so that Presto Optimizer today assumes that joins preserve the order of probe rows. Breaking this assumption leads to wrong results.

See https://github.com/facebookincubator/velox/pull/10832

@Yuhta Jimmy, do you have a link to the issue that talks about wrong results?

velox/core/QueryConfig.h
lifulong: @mbasmanova Got it.

lifulong:
> @lifulong
> 
> It happens so that Presto Optimizer today assumes that joins preserve the order of probe rows. Breaking this assumption leads to wrong results.
> 
> See #10832
> 
> @Yuhta Jimmy, do you have a link to the issue that talks about wrong results?

Got it, I'll study this PR.

aditi-pandit:
@lifulong : Presto issue we found as well https://github.com/prestodb/presto/issues/22585

@mbasmanova

lifulong:
> @lifulong : Presto issue we found as well [prestodb/presto#22585](https://github.com/prestodb/presto/issues/22585)
> 
> @mbasmanova

@aditi-pandit  Got it.

PR #13667, opened 2025-06-06
fix(parquet): Reserve memory for thrift allocation

### LINKED ISSUE #12252: Fix Unreported Heap memory allocations while deserializing footers
### Description

When reading Parquet and DWRF footers, the Thrift/Protobuf deserializer allocates memory on the heap, potentially causing OOM errors. Also it affects fair arbitration. The problem is especially worse with Parquet, since all metadata is in the footer. 
One simple solution would be to enable dynamic byte reporting in the memory allocator:  we could report an initial estimate, adjust it based on actual usage, and then free the reported memory in the memory allocator. This would make sure that the memory is at least reported in the standard memory.
Another option is to replace the default STL allocator with one with velox memory management. Lastly ofcource reimplementing the deserializers in velox

For parquet the thrift deserializations seems to be happening [here](https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/reader/ParquetReader.cpp#L218) 

Thanks @majetideepak  for all the background

jaystarshot:
Fix #12252
```
 FileMetaData Statistics for 
  Footer length: 129.85MB):
  FileMetaData object size: 959.77MB
  Schema Statistics:
    Number of schema elements: 6879
    Maximum schema depth: 1 (Element: 'hive_schema', Size: 367B)
    Largest schema element:  (Size: 462B)
    Number of leaf nodes: 0
    Number of group nodes: 0
    Average string length: 0
    Maximum string length: 0
  Row Group Statistics:
    Number of row groups: 147
    Total columns: 1011066
    Average columns per row group: 6878
    Total encodings: 6066480
    Distinct encodings: 4
    Encoding distribution:
      PLAIN: 430661 occurrences
      PLAIN_DICTIONARY: 580447 occurrences
      RLE: 1011066 occurrences
      BIT_PACKED: 1011066 occurrences
    Total paths: 1011066
    Total key-value pairs: 0
```

We have a table with a large footer and queries on this cause OOMs as the thrift memory is not reported.

With this change it is reported and queries begin to fail

```
Top 10 leaf memory pool usages:
    op.4.0.6.Aggregation usage 23.74GB reserved 23.74GB peak 23.74GB
    op.0.0.6.TableScan usage 1.58GB reserved 1.59GB peak 2.81GB
    op.0.0.0.TableScan usage 1.18GB reserved 1.19GB peak 1.67GB
    op.0.0.1.TableScan usage 1.11GB reserved 1.12GB peak 4.86GB
    op.0.0.7.TableScan usage 1.10GB reserved 1.11GB peak 2.70GB
    op.0.0.5.TableScan usage 1.07GB reserved 1.07GB peak 2.22GB
    op.0.0.3.TableScan usage 1.02GB reserved 1.03GB peak 2.16GB
    op.0.0.4.TableScan usage 590.64MB reserved 592.00MB peak 1.65GB
    op.466.1.5.Exchange usage 39.19MB reserved 40.00MB peak 46.05MB
    exchangeClient.466.1 usage 32.25MB reserved 36.00MB peak 88.54MB
```

### DIFF
--- a/velox/common/base/tests/StatsReporterTest.cpp
+++ b/velox/common/base/tests/StatsReporterTest.cpp
@@ -251,6 +251,10 @@ class TestMemoryPool : public memory::MemoryPool {
     return nullptr;
   }
 
+  void reportAllocation(int64_t size) override {
+    return;
+  }
+
   void* allocateZeroFilled(int64_t /* unused */, int64_t /* unused */)
       override {
     return nullptr;
@@ -264,6 +268,8 @@ class TestMemoryPool : public memory::MemoryPool {
   }
 
   void free(void* /* unused */, int64_t /* unused */) override {}
+
+  void reportFree(int64_t /* unused */) override {}
 
   void allocateNonContiguous(
       memory::MachinePageCount /* unused */,
--- a/velox/common/memory/MemoryPool.h
+++ b/velox/common/memory/MemoryPool.h
@@ -241,6 +241,8 @@ class MemoryPool : public std::enable_shared_from_this<MemoryPool> {
       int64_t size,
       std::optional<uint32_t> alignment = std::nullopt) = 0;
 
+  virtual void reportAllocation(int64_t size) = 0;
+
   /// Allocates a zero-filled buffer with capacity that can store 'numEntries'
   /// entries with each size of 'sizeEach'.
   virtual void* allocateZeroFilled(int64_t numEntries, int64_t sizeEach) = 0;
@@ -258,6 +260,7 @@ class MemoryPool : public std::enable_shared_from_this<MemoryPool> {
   transferTo(MemoryPool* /*dest*/, void* /*buffer*/, uint64_t /*size*/) {
     return false;
   }
+  virtual void reportFree(int64_t size) = 0;
 
   /// Allocates one or more runs that add up to at least 'numPages', with the
   /// smallest run being at least 'minSizeClass' pages. 'minSizeClass' must be
@@ -617,6 +620,8 @@ class MemoryPoolImpl : public MemoryPool {
   void* allocate(int64_t size, std::optional<uint32_t> alignment = std::nullopt)
       override;
 
+  void reportAllocation(int64_t size) override;
+
   void* allocateZeroFilled(int64_t numEntries, int64_t sizeEach) override;
 
   void* reallocate(void* p, int64_t size, int64_t newSize) override;
@@ -624,6 +629,8 @@ class MemoryPoolImpl : public MemoryPool {
   void free(void* p, int64_t size) override;
 
   bool transferTo(MemoryPool* dest, void* buffer, uint64_t size) override;
+
+  void reportFree(int64_t size) override;
 
   void allocateNonContiguous(
       MachinePageCount numPages,
--- a/velox/connectors/hive/HiveConfig.h
+++ b/velox/connectors/hive/HiveConfig.h
@@ -140,6 +140,12 @@ class HiveConfig {
   /// request
   static constexpr const char* kFooterEstimatedSize = "footer-estimated-size";
 
+  /// The byte size threshold beyond which the Parquet reader estimates and
+  /// reports the memory usage of deserialized Thrift objects to the memory
+  /// pool.
+  static constexpr const char* kParquetFooterTrackThriftMemoryThreshold =
+      "hive.parquet.track-footer-thrift-memory-threshold";
+
   /// The threshold of file size in bytes when the whole file is fetched with
   /// meta data together. Optimization to decrease the small IO requests
   static constexpr const char* kFilePreloadThreshold = "file-preload-threshold";
@@ -268,6 +274,8 @@ class HiveConfig {
       const config::ConfigBase* session) const;
 
   uint64_t footerEstimatedSize() const;
+
+  uint64_t parquetFooterTrackThriftMemoryThreshold() const;
 
   uint64_t filePreloadThreshold() const;
 
--- a/velox/dwio/common/Options.h
+++ b/velox/dwio/common/Options.h
@@ -521,6 +521,9 @@ class ReaderOptions : public io::ReaderOptions {
   static constexpr uint64_t kDefaultFilePreloadThreshold =
       1024 * 1024 * 8; // 8MB
 
+  static constexpr uint64_t kDefaultParquetFooterTrackThriftMemoryThreshold =
+      std::numeric_limits<uint64_t>::max(); // Disabled by default
+
   explicit ReaderOptions(velox::memory::MemoryPool* pool)
       : io::ReaderOptions(pool),
         tailLocation_(std::numeric_limits<uint64_t>::max()),
@@ -564,6 +567,10 @@ class ReaderOptions : public io::ReaderOptions {
 
   ReaderOptions& setFooterEstimatedSize(uint64_t size) {
     footerEstimatedSize_ = size;
+    return *this;
+  }
+  ReaderOptions& setParquetFooterTrackThriftMemoryThreshold(uint64_t size) {
+    parquetFooterTrackThriftMemoryThreshold_ = size;
     return *this;
   }
 
@@ -626,6 +633,10 @@ class ReaderOptions : public io::ReaderOptions {
 
   uint64_t footerEstimatedSize() const {
     return footerEstimatedSize_;
+  }
+
+  uint64_t parquetFooterTrackThriftMemoryThreshold() const {
+    return parquetFooterTrackThriftMemoryThreshold_;
   }
 
   uint64_t filePreloadThreshold() const {
@@ -709,6 +720,8 @@ class ReaderOptions : public io::ReaderOptions {
   bool adjustTimestampToTimezone_{false};
   bool selectiveNimbleReaderEnabled_{false};
   bool allowEmptyFile_{false};
+  uint64_t parquetFooterTrackThriftMemoryThreshold_{
+      kDefaultParquetFooterTrackThriftMemoryThreshold};
 };
 
 struct WriterOptions {
--- a/velox/dwio/parquet/reader/Metadata.cpp
+++ b/velox/dwio/parquet/reader/Metadata.cpp
@@ -363,4 +363,157 @@ std::string FileMetaDataPtr::createdBy() const {
   return thriftFileMetaDataPtr(ptr_)->created_by;
 }
 
+size_t calculateColumnMetadataSize(const thrift::ColumnChunk& column) {
+  size_t size = 0;
+  size += sizeof(thrift::ColumnChunk);
+  size += sizeof(thrift::ColumnMetaData);
+  size += column.meta_data.encodings.size() * sizeof(thrift::Encoding::type);
+  size += column.meta_data.path_in_schema.size() * sizeof(std::string);
+  for (const auto& path : column.meta_data.path_in_schema) {
+    size += path.capacity();
+  }
+  size += column.meta_data.key_value_metadata.size() * sizeof(thrift::KeyValue);
+  for (const auto& kv : column.meta_data.key_value_metadata) {
+    size += kv.key.capacity();
+    size += kv.value.capacity();
+  }
+
+  if (column.meta_data.__isset.statistics) {
+    const auto& stats = column.meta_data.statistics;
+    size += sizeof(thrift::Statistics);
+    if (stats.__isset.min)
+      size += stats.min.capacity();
+    if (stats.__isset.max)
+      size += stats.max.capacity();
+    if (stats.__isset.min_value)
+      size += stats.min_value.capacity();
+    if (stats.__isset.max_value)
+      size += stats.max_value.capacity();
+    if (stats.__isset.null_count)
+      size += sizeof(int64_t);
+    if (stats.__isset.distinct_count)
+      size += sizeof(int64_t);
+  }
+  return size;
+}
+
+size_t analyzeFileMetadata(
+    const thrift::FileMetaData& metadata,
+    const std::string& filename,
+    uint64_t footerLength) {
+  size_t totalSize = sizeof(thrift::FileMetaData);
+
+  size_t numSchemaElements = metadata.schema.size();
+  size_t maxSchemaDepth = 0;
+  size_t currentDepth = 0;
+  size_t numLeafNodes = 0;
+  size_t numGroupNodes = 0;
+  size_t totalStringLength = 0;
+  size_t maxStringLength = 0;
+  size_t maxDepth = 0;
+  size_t maxDepthElementSize = 0;
+  std::string maxDepthElementName;
+  size_t maxElementSize = 0;
+  std::string maxElementName;
+  currentDepth = 0;
+
+  for (const auto& schema : metadata.schema) {
+    size_t elementSize = sizeof(schema) + schema.name.capacity() +
+        sizeof(schema.type) + sizeof(schema.type_length) +
+        sizeof(schema.repetition_type) + sizeof(schema.num_children) +
+        sizeof(schema.converted_type) + sizeof(schema.scale) +
+        sizeof(schema.precision) + sizeof(schema.field_id);
+    if (schema.__isset.logicalType) {
+      elementSize += sizeof(schema.logicalType);
+    }
+    if (elementSize > maxElementSize) {
+      maxElementSize = elementSize;
+      maxElementName = schema.name;
+    }
+    if (schema.__isset.num_children) {
+      currentDepth++;
+      if (currentDepth > maxDepth) {
+        maxDepth = currentDepth;
+        maxDepthElementName = schema.name;
+        maxDepthElementSize = elementSize;
+      }
+    }
+  }
+  size_t numRowGroups = metadata.row_groups.size();
+  size_t totalColumns = 0;
+  size_t totalEncodings = 0;
+  size_t totalPaths = 0;
+  size_t totalKeyValues = 0;
+
+  totalSize += metadata.row_groups.size() * sizeof(thrift::RowGroup);
+  for (const auto& rowGroup : metadata.row_groups) {
+    totalColumns += rowGroup.columns.size();
+    totalSize += rowGroup.columns.size() * sizeof(thrift::ColumnChunk);
+    for (const auto& column : rowGroup.columns) {
+      totalSize += calculateColumnMetadataSize(column);
+      totalEncodings += column.meta_data.encodings.size();
+      totalPaths += column.meta_data.path_in_schema.size();
+      totalKeyValues += column.meta_data.key_value_metadata.size();
+    }
+  }
+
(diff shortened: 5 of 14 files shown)


velox/common/memory/MemoryPool.h
```diff
@@ -234,6 +234,8 @@ class MemoryPool : public std::enable_shared_from_this<MemoryPool> {
       int64_t size,
       std::optional<uint32_t> alignment = std::nullopt) = 0;
 
+  virtual void fakeAllocate(int64_t size) = 0;
```
majetideepak: Why not call this `virtual void reserve(int64_t size) = 0;` and use it inside `allocate` and `allocateZeroFilled` as well?

velox/common/memory/MemoryPool.h
```diff
@@ -245,6 +247,8 @@ class MemoryPool : public std::enable_shared_from_this<MemoryPool> {
   /// Frees an allocated buffer.
   virtual void free(void* p, int64_t size) = 0;
 
+  virtual void fakeFree(int64_t size) = 0;
```
majetideepak: Similarly, call this `release`.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -36,6 +36,29 @@ bool isParquetReservedKeyword(
       ? true
       : false;
 }
+
+size_t calculateColumnMetadataSize(const thrift::ColumnChunk& column) {
+  size_t size = 0;
+  // Add size of column metadata
+  size += sizeof(thrift::ColumnChunk);
```
majetideepak: There are statistics as well here, which can be significant for string columns.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -108,6 +137,15 @@ class ReaderBase {
   /// the data still exists in the buffered inputs.
   bool isRowGroupBuffered(int32_t rowGroupIndex) const;
 
+  size_t analyzeFileMetadata(
+      const thrift::FileMetaData& metadata,
+      const std::string& filename,
+      uint64_t footerLength);
+
+  bool thriftMemoryTracked_ = false;
```
majetideepak: Let's add a Parquet Config for this?

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1122,6 +1352,10 @@ class ParquetRowReader::Impl {
           // reduce the memory consumption. ColumnChunks consume the most
           // memory. Skip the 0th RowGroup as it is used by estimatedRowSize().
           rowGroups_[i].columns.clear();
+          for (const auto& column : rowGroups_[i].columns) {
```
majetideepak: Put this under the config as well.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -220,6 +258,11 @@ void ReaderBase::loadFileMetaData() {
       thriftTransport);
   fileMetaData_ = std::make_unique<thrift::FileMetaData>();
   fileMetaData_->read(thriftProtocol.get());
+  // Only analyze metadata if footer length is greater than 20MB
+  if (footerLength > 20 * 1024 * 1024) {
```
majetideepak: Use the config and remove this.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1027,6 +1070,191 @@ bool ReaderBase::isRowGroupBuffered(int32_t rowGroupIndex) const {
...
+  for (const auto& [encoding, count] : encodingCounts) {
+    encodingStats += fmt::format(
+        "      {}: {} occurrences\n", encodingToString(encoding), count);
+
+    // Record metric for each encoding type
+  }
+
+  // Only log if totalSize is more than 8x footerLength
+  if (totalSize > 8 * footerLength) {
+    LOG(INFO) << fmt::format(
```
majetideepak: VLOG(1)

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1027,6 +1070,191 @@ bool ReaderBase::isRowGroupBuffered(int32_t rowGroupIndex) const {
   return inputs_.count(rowGroupIndex) != 0;
 }
 
+size_t ReaderBase::analyzeFileMetadata(
```
majetideepak: Can we move this to the `Metadata.h/.cpp` please. This file is already huge.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -36,6 +36,29 @@ bool isParquetReservedKeyword(
       ? true
       : false;
 }
+
+size_t calculateColumnMetadataSize(const thrift::ColumnChunk& column) {
```
majetideepak: Move to the Metadata.h/.cpp file please.

majetideepak (commented):
@jaystarshot thanks for this fix. Some comments.

velox/dwio/parquet/reader/ParquetReader.cpp
jaystarshot: This needs to be false initially and we need to set this to true only if the reserve() succeeds i.e we should't release()  without reserving(causes crashes)

majetideepak: If `reserve()` fails, then the query fails correct?

jaystarshot: Correct but the destructor is called and that shouldn't call release()

majetideepak: Can we not control the call to `release()` under the same config?
I feel one config, say `hive.parquet.thrift.track-memory-use` can control everything.

majetideepak: I see that the destructor always calls release(). We can keep this bool. Probably rename as ` bool thriftMemoryReseved_ = false;`

majetideepak: Probably need a simple RAII object inside the FileMetaData class itself for this.

velox/common/base/tests/StatsReporterTest.cpp
```diff
@@ -299,6 +303,8 @@ class TestMemoryPool : public memory::MemoryPool {
 
   void free(void* /* unused */, int64_t /* unused */) override {}
 
+  void reportFreedAllocation(int64_t /* unused */) override {}
```
majetideepak: nit: `reportFree` to match `free` above.

velox/connectors/hive/HiveConfig.h
```diff
@@ -139,6 +139,17 @@ class HiveConfig {
   /// request
   static constexpr const char* kFooterEstimatedSize = "footer-estimated-size";
 
+  /// The byte size threshold beyond which the Parquet reader estimates and
+  /// reports the memory usage of deserialized Thrift objects to the memory
+  /// pool.
+  static constexpr const char* kparquetFooterTrackThriftMemoryThreshold =
```
majetideepak: `kParquet`

velox/dwio/common/Options.h
```diff
@@ -572,6 +586,14 @@ class ReaderOptions : public io::ReaderOptions {
     return footerEstimatedSize_;
   }
 
+  uint64_t parquetFooterTrackThriftMemoryThreshold() const {
```
majetideepak, line 638: I think we can do away with one config. We can default value to uint64_t::max to disable it.

velox/connectors/hive/HiveConfig.h
```diff
@@ -139,6 +139,17 @@ class HiveConfig {
...
+      "hive.parquet.track-footer-thrift-memory-threshold";
+
+  // Flag indicating if the Parquet reader should estimate and
+  // report the memory usage of deserialized Thrift objects to the memory pool.
+  static constexpr const char* kParquetFooterTrackThriftMemoryEnabled =
+      "hive.parquet.track-footer-thrift-memory-enabled";
```
czentgr: Slight inconsistency in the variable name and the parameter name.
```
kParquetTrackFooterThriftMemoryEnabled
```

velox/connectors/hive/HiveConfig.h
czentgr: Slight inconsistency in the variable name and the parameter name.
```
kParquetTrackFooterThriftMemoryThreshold
```

velox/dwio/common/Options.h
```diff
@@ -465,6 +465,11 @@ class ReaderOptions : public io::ReaderOptions {
   static constexpr uint64_t kDefaultFilePreloadThreshold =
       1024 * 1024 * 8; // 8MB
 
+  static constexpr uint64_t kDefaultParquetFooterTrackThriftMemoryThreshold =
```
czentgr, line 524: Same issue with the name.

velox/dwio/common/Options.h
```diff
@@ -465,6 +465,11 @@ class ReaderOptions : public io::ReaderOptions {
...
+      20 * 1024 * 1024; // 20MB
+
+  static constexpr bool kDefaultParquetFooterTrackThriftMemoryEnabled = true;
```
czentgr: Same issue with the name.

velox/dwio/parquet/reader/Metadata.h
```diff
@@ -160,4 +161,230 @@ class FileMetaDataPtr {
   const void* ptr_;
 };
 
+// returns the size of the thrift::ColumnChunk
```
czentgr: Comment not needed. The function name explains it.

velox/dwio/parquet/reader/Metadata.h
```diff
@@ -160,4 +161,230 @@ class FileMetaDataPtr {
...
+static size_t calculateColumnMetadataSize(const thrift::ColumnChunk& column) {
```
czentgr: Why are we "inlining" the functions in the header and not put them into Metadata.cpp?
We can remove the static and declare the function in an anonymous namespace (and implement it in the cpp in an anonymous namespace).

velox/dwio/parquet/reader/Metadata.h
```diff
@@ -160,4 +161,230 @@ class FileMetaDataPtr {
...
+  // Add size of column_orders
+  totalSize += metadata.column_orders.size() * sizeof(thrift::ColumnOrder);
+
+  // Add size of encryption_algorithm
+  totalSize += sizeof(thrift::EncryptionAlgorithm);
+
+  // Add size of footer_signing_key_metadata
+  totalSize += metadata.footer_signing_key_metadata.capacity();
+
+  auto encodingToString = [](thrift::Encoding::type encoding) -> std::string {
```
czentgr: I guess we don't have access to `EncodingToString`.

velox/dwio/common/Options.h
czentgr: Basically, get rid of the `parquetFooterTrackThriftMemoryEnabled_` and keep threshold with threshold set to `uint64_t::max` to disable it.

velox/dwio/parquet/reader/Metadata.h
```diff
@@ -160,4 +161,230 @@ class FileMetaDataPtr {
   const void* ptr_;
 };
 
+// returns the size of the thrift::ColumnChunk
+static size_t calculateColumnMetadataSize(const thrift::ColumnChunk& column) {
+  size_t size = 0;
+  // Add size of column metadata
```
czentgr: Nit: please see Velox comment guidelines for the functions here.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -221,6 +231,12 @@ void ReaderBase::loadFileMetaData() {
       thriftTransport);
   fileMetaData_ = std::make_unique<thrift::FileMetaData>();
   fileMetaData_->read(thriftProtocol.get());
+  // Only analyze metadata if footer length is greater than 20MB
```
czentgr: Do we need this comment here? It seems straight forward. Please see the Velox guideline, otherwise this is a configurable threshold and not 20MB with missing period.

czentgr (commented):
A few minor things.

velox/dwio/parquet/reader/ParquetReader.cpp
jaystarshot: thriftSize_ > 0 is the best way to do that since we removed the enabled flag.

velox/dwio/parquet/reader/Metadata.h
jaystarshot: Why is it better than having a static function which is simpler and less code change?

jaystarshot: ahh i thought its already there nvm

velox/connectors/hive/HiveConfig.h
```diff
@@ -139,6 +139,12 @@ class HiveConfig {
   /// request
   static constexpr const char* kFooterEstimatedSize = "footer-estimated-size";
 
+  /// The byte size threshold beyond which the Parquet reader estimates and
+  /// reports the memory usage of deserialized Thrift objects to the memory
+  /// pool.
+  static constexpr const char* kParquetFooterTrackThriftMemoryThreshold =
```
majetideepak, line 146: We need to add a doc here `velox/docs/configs.rst ` under `Parquet File Format Configuration`

velox/dwio/common/Options.h
majetideepak: Default is `std::numeric_limits<uint64_t>::max()` ?

velox/dwio/parquet/reader/Metadata.cpp
```diff
@@ -363,4 +363,202 @@ std::string FileMetaDataPtr::createdBy() const {
...
+    }
+  }
+
+  std::string encodingStats;
+  for (const auto& [encoding, count] : encodingCounts) {
+    encodingStats += fmt::format(
+        "      {}: {} occurrences\n", encodingToString(encoding), count);
+  }
+
+  if (totalSize > 8 * footerLength) {
```
majetideepak: Why do we need this check?

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1131,6 +1153,12 @@ class ParquetRowReader::Impl {
 
       rowNumber += rowGroups_[i].num_rows;
     }
+
+    if (totalClearedColumns > 0 && readerBase_->thriftSize_ > 0) {
```
majetideepak, line 1298: why do we check for `totalClearedColumns > 0`? What if all the RowGroups are read?

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -221,6 +231,10 @@ void ReaderBase::loadFileMetaData() {
       thriftTransport);
   fileMetaData_ = std::make_unique<thrift::FileMetaData>();
   fileMetaData_->read(thriftProtocol.get());
+  if (footerLength > options().parquetFooterTrackThriftMemoryThreshold()) {
+    thriftSize_ = analyzeFileMetadata(
+        *fileMetaData_, input_->getReadFile()->getName(), footerLength);
+  }
```
majetideepak, line 253: `thriftMemoryTracked_ = true;` here?

jaystarshot: no we need to set it when the reserve succeeds, will rename this variable

velox/dwio/parquet/reader/ParquetReader.cpp
jaystarshot: @majetideepak  nice catch, correct that should just be for the line 1158

velox/dwio/parquet/reader/Metadata.cpp
jaystarshot: just for some logging, will remove it

jaystarshot:
@majetideepak @czentgr   I also added a UT, can you please review? Sorry for the delay in addressing the comments

majetideepak:
@jaystarshot CI is red. Can you take a look?

steveburnett (approved):
LGTM! (docs)

Pull branch, local doc build.

velox/common/memory/MemoryPool.h
```diff
@@ -241,6 +241,8 @@ class MemoryPool : public std::enable_shared_from_this<MemoryPool> {
       int64_t size,
       std::optional<uint32_t> alignment = std::nullopt) = 0;
 
+  virtual void reportAllocation(int64_t size) = 0;
```
aditi-pandit, line 244: Can you add a comment explaining usage for these method as its external by nature ?

velox/dwio/parquet/reader/Metadata.cpp
```diff
@@ -363,4 +363,70 @@ std::string FileMetaDataPtr::createdBy() const {
...
+  }
+  totalSize += metadata.row_groups.size() * sizeof(thrift::RowGroup);
+  for (const auto& rowGroup : metadata.row_groups) {
+    totalSize += rowGroup.columns.size() * sizeof(thrift::ColumnChunk);
+    for (const auto& column : rowGroup.columns) {
+      totalSize += calculateColumnMetadataSize(column);
+    }
+  }
+  totalSize += metadata.key_value_metadata.size() * sizeof(thrift::KeyValue);
+  for (const auto& kv : metadata.key_value_metadata) {
```
aditi-pandit, line 421: This key value computation looks similar across both computations. Maybe add a common method for it.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
```diff
@@ -1786,6 +1786,33 @@ TEST_F(ParquetReaderTest, readerWithSchema) {
...
+  {
+    auto reader = createReader(sample, readerOptions);
+    EXPECT_EQ(reader->numberOfRows(), 20ULL);
+    auto rowReaderOpts = getReaderOpts(sampleSchema());
+    auto scanSpec = makeScanSpec(sampleSchema());
+    rowReaderOpts.setScanSpec(scanSpec);
+    auto rowReader = reader->createRowReader(rowReaderOpts);
+    memoryAfterRowReader = leafPool_->usedBytes();
+    EXPECT_GT(memoryAfterRowReader, initialUsage);
+    auto result = BaseVector::create(sampleSchema(), 0, leafPool_.get());
```
aditi-pandit, line 1808: Nit : Add newline before this line.

aditi-pandit (commented):
This is a good change @jaystarshot. Thanks for the code.

steveburnett:
Please resolve the file conflict to move this PR closer to merge status.

jja725:
@jaystarshot do you mind fix the ci and after that @aditi-pandit or @Yuhta can give a another look

mbasmanova (changes requested):
per @jja725 's comment

PR #13670, opened 2025-06-06
misc: Use contains() to replace in-list index join condition to ease SQL parsing

xiaoxmeng:
Summary: Use contains() syntax for contains (or previously in-list) index join condition

Differential Revision: D76112220


mbasmanova:
@xiaoxmeng Is this PR still valid? If not, let's close it.

PR #13672, opened 2025-06-06
add ARM support for CUDA install

yabinma:


scripts/setup-centos9.sh
```diff
@@ -86,7 +86,20 @@ function install_gflags {
...
+  local repo_url
+
+  if [[ "$arch" == "x86_64" ]]; then
+    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo"
+  elif [[ "$arch" == "aarch64" ]]; then
+    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/cuda-rhel9.repo"
+  else
+    echo "Unsupported architecture: $arch" >&2
+    return 1
+  fi
```
assignUser: Could be solved with a case statement but I personally find those lack readability in bash, so this is good :)

majetideepak (approved):

scripts/setup-centos9.sh
```diff
@@ -86,7 +86,20 @@ function install_gflags {
 
 function install_cuda {
   # See https://developer.nvidia.com/cuda-downloads
-  dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
+  #dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
```
bdice: This commented line can be removed.
```suggestion
```

scripts/setup-manylinux.sh
```diff
@@ -262,7 +262,21 @@ function install_arrow {
 
 function install_cuda {
   # See https://developer.nvidia.com/cuda-downloads
-  dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
+  #dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
```
bdice: This commented line can be removed.
```suggestion
```

scripts/setup-manylinux.sh
```diff
@@ -262,7 +262,21 @@ function install_arrow {
...
+  local arch=$(uname -m)
+  local repo_url
+
+  if [[ "$arch" == "x86_64" ]]; then
+    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo"
+  elif [[ "$arch" == "aarch64" ]]; then
+    # Using SBSA (Server Base System Architecture) repository for ARM64 servers
```
bdice, line 271: This is a near copy of the changes in the centos9 script but this copy has a comment. It is probably worth matching (with comment both places).

scripts/setup-ubuntu.sh
```diff
@@ -156,12 +156,43 @@ function install_conda {
...
+  #   $SUDO apt update
+  # fi
+  local arch=$(uname -m)
+  local os_ver
+
+  if [[ ${VERSION} =~ "22.04" ]]; then
+    os_ver="ubuntu2204"
+  elif [[ ${VERSION} =~ "24.04" ]]; then
+    os_ver="ubuntu2404"
+  elif [[ ${VERSION} =~ "20.04" ]]; then
```
bdice, line 166: Let’s keep these in order. Probably 24.04, 22.04, 20.04 is best.

Also note that Ubuntu 20.04 is no longer receiving security updates. I recommend that Velox should drop it completely if there are no constraints preventing that.

scripts/setup-ubuntu.sh
```diff
@@ -156,12 +156,43 @@ function install_conda {
 
 function install_cuda {
   # See https://developer.nvidia.com/cuda-downloads
+  # if ! dpkg -l cuda-keyring 1>/dev/null; then
```
bdice: As above, delete this commented section.

bdice (changes requested):

scripts/setup-centos9.sh
yabinma: done in new commit.

scripts/setup-manylinux.sh
yabinma: done in new commit.

scripts/setup-manylinux.sh
yabinma: done in new commit.

scripts/setup-ubuntu.sh
yabinma: done in new commit.

scripts/setup-ubuntu.sh
yabinma: updated in new commit.  Please help review again. 

This is not the only place to deal with Ubuntu versions.   It may handle it in a separate issue.  20.04 is rarely used  I believe. However, it may discuss more if droppint it at all(saying 20.04 is not supported) or a default behavior.

bdice (approved):
The manylinux images should use RHEL 8 instead of RHEL 9 in order to match the desired tag of `manylinux_2_28` (RHEL 8 uses glibc 2.28). But that isn't a new problem in this PR so I am approving as-is.

https://github.com/facebookincubator/velox/blob/c51d8642f53ba88b11e40480810ce7b547753325/scripts/docker/pyvelox.dockerfile#L15

bdice:
@assignUser @kgpai Just a heads-up on these two comments:

- Velox could [drop Ubuntu 20.04](https://github.com/facebookincubator/velox/pull/13672#discussion_r2133356538) (past EOL for security updates: https://endoflife.date/ubuntu)
- manylinux images [should be based on RHEL 8](https://github.com/facebookincubator/velox/pull/13672#pullrequestreview-2907602645), not RHEL 9, in order to adhere to the glibc 2.28 requirements

assignUser:
Agreed let's drop 20.04! 

The pyvox image uses the official many Linux image as a base which is iirc Alma Linux 8, so that should already be what we need :)

bdice:
Okay, then we should modify this to pull RHEL 8 CUDA stuff.

scripts/setup-manylinux.sh
```diff
@@ -262,7 +262,20 @@ function install_arrow {
 
 function install_cuda {
   # See https://developer.nvidia.com/cuda-downloads
-  dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo
+  local arch=$(uname -m)
+  local repo_url
+
+  if [[ "$arch" == "x86_64" ]]; then
+    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo"
```
bdice:
```suggestion
    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64/cuda-rhel8.repo"
```

scripts/setup-manylinux.sh
```diff
@@ -262,7 +262,20 @@ function install_arrow {
...
+  elif [[ "$arch" == "aarch64" ]]; then
+    # Using SBSA (Server Base System Architecture) repository for ARM64 servers
+    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/cuda-rhel9.repo"
```
bdice:
```suggestion
    repo_url="https://developer.download.nvidia.com/compute/cuda/repos/rhel8/sbsa/cuda-rhel8.repo"
```

bdice (commented):
Here are a couple manylinux fixes, then this should be fine.

PR #13677, opened 2025-06-06
misc: Add QueryCtx::started_ member

spershin:
Summary:
This member will be used to avoid queuing tasks for a query that
already has tasks started.

Differential Revision: D76137271

### DIFF
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -857,6 +857,8 @@ void Task::recordBatchEndTime() {
 }
 
 void Task::start(uint32_t maxDrivers, uint32_t concurrentSplitGroups) {
+  queryCtx()->setStarted();
+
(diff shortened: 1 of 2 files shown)


velox/exec/Task.cpp
```diff
@@ -857,6 +857,8 @@ void Task::recordBatchEndTime() {
 }
 
 void Task::start(uint32_t maxDrivers, uint32_t concurrentSplitGroups) {
+  queryCtx()->setStarted();
```
xiaoxmeng, line 860: queryCtx is used to provide the query execution context. Why we provide an action method? I have a change to split init from start: https://github.com/facebookincubator/velox/pull/13075. But it breaks meta internal streaming use case. We can work with Meta team if we really needs to support this for server pushdown mechanism? Thanks!

PR #13694, opened 2025-06-09
feat: Reduce HashTable load factor from 0.875 to 0.7

arhimondr:
Summary:
Load factor of 7/8 is too high. When table is 7/8 full a significant performance
degradation is observed. Highly selective use cases (when value is not
present in the hash table) are impacted the most.

According to the simulation when table is 7/8 full a single probe operation has
to visit 2.88 buckets on average.

When double hashing algorithm is used the average number of bucket to visit is
reduced to 1.88 with the idential load.

The optimal load factor seems to be at 0.7. Past that point performance starts
to deteriorate fast. Performance deteriorates much faster for linear probing
algorithm.

Differential Revision: D76267085

### DIFF
--- a/velox/exec/HashTable.h
+++ b/velox/exec/HashTable.h
@@ -127,6 +127,9 @@ class BaseHashTable {
 
   using MaskType = uint16_t;
 
+  /// The load factor of the hash table.
+  static constexpr double kHashTableLoadFactor = 0.7;
+
   /// 2M entries, i.e. 16MB is the largest array based hash table.
   static constexpr uint64_t kArrayHashMaxSize = 2L << 20;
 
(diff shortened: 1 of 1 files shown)


arhimondr:
Discovered when running a modified TPC-H Q4 query over a sf1000 schema on a single node cluster:

```
SELECT 
  o_orderpriority, 
  COUNT(*) AS order_count 
FROM 
  lineitem, 
  orders 
WHERE 
  o_orderdate >= CAST('1993-07-01' AS DATE)
  AND o_orderdate < CAST('1993-10-01' AS DATE)
  AND l_orderkey = o_orderkey 
  AND l_commitdate < l_receiptdate 
GROUP BY 
  o_orderpriority 
ORDER BY 
  o_orderpriority
```

Join CPU:

Load factor 7/8: 504.42s
Load factor 1/2: 183.23s
Load factor 0.7: 215.4s

arhimondr:
Simulation code: https://gist.github.com/arhimondr/d35863f0548e9ad2717e29c4591e6666

Screenshot 2025-06-09 at 11 53 53 AM
Screenshot 2025-06-09 at 11 53 40 AM

PR #13695, opened 2025-06-09
feat: Add a CudfTpchBenchmark

devavret:
This PR adds a CudfTpchBenchmark, similar to TpchBenchmark and makes some changes to TpchBenchmark to allow it to be extended.

Depends on https://github.com/facebookincubator/velox/pull/14462


velox/benchmarks/QueryBenchmarkBase.cpp
```diff
@@ -195,6 +237,40 @@ void QueryBenchmarkBase::initialize() {
...
+          cudf_velox::connector::parquet::ParquetConnectorFactory::
+              kParquetConnectorName)
+          ->newConnector(
+              cudf_velox::exec::test::kParquetConnectorId,
+              parquetProperties,
+              ioExecutor_.get());
+  connector::registerConnector(parquetConnector);
+
+  // Enable cuDF operators
+  cudf_velox::registerCudf();
```
bdice: I would like to be able to build the benchmarks with cuDF support but only enable cuDF if a flag is passed to the benchmark executable. Can we use gflags to conditionally register cuDF, rather than having to make that decision at build time? It will make it easier to compare performance if we can build once with GPU support and then run with/without cuDF.

In other words, what I am looking for is:

```cpp
// Add a runtime flag:

#ifdef VELOX_ENABLE_CUDF
DEFINE_bool(
    enable_cudf,
    false,
    "Run with cuDF enabled");
#endif

// Then in the initialize function:

#ifdef VELOX_ENABLE_CUDF
if (FLAGS_enable_cudf) {
  // Call logic to register cuDF
}
#endif
```

velox/benchmarks/QueryBenchmarkBase.cpp
```diff
@@ -195,6 +237,40 @@ void QueryBenchmarkBase::initialize() {
   connector::registerConnector(hiveConnector);
   parquet::registerParquetReaderFactory();
   dwrf::registerDwrfReaderFactory();
+
```
jinchengchenghh, line 217: Is it convenient for you to add a new class `CudfQueryBenchmarkBase` and extends the current class `QueryBenchmarkBase`?  Then `ifdef VELOX_ENABLE_CUDF` is reduced.

velox/benchmarks/QueryBenchmarkBase.cpp
```diff
@@ -211,7 +287,30 @@ QueryBenchmarkBase::listSplits(
   return result;
 }
 
+#ifdef VELOX_ENABLE_CUDF
+std::vector<std::shared_ptr<connector::ConnectorSplit>>
```
jinchengchenghh: Could you move it to anonymous namespace?

velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -276,14 +313,32 @@ core::PlanNodePtr PlanBuilder::TableScanBuilder::build(core::PlanNodeId id) {
...
-        connectorId_,
-        tableName_,
-        true,
-        (subfieldFiltersMap_.empty()) ? std::move(filters)
-                                      : std::move(subfieldFiltersMap_),
-        remainingFilterExpr,
-        dataColumns_);
+    tableHandle_ = [&]() -> std::shared_ptr<connector::ConnectorTableHandle> {
+#ifdef VELOX_ENABLE_CUDF
+      // if cudfIsRegistered, then use cudftableScan tableHandle_ here.
```
jinchengchenghh: if -> If

velox/benchmarks/QueryBenchmarkBase.cpp
devavret: This already works like you expect. The cudf driver adapter only "activates" when the gflag `velox_cudf_enabled` is set to true. What happens here is that the gflag determines the default args for `CudfOptions` https://github.com/facebookincubator/velox/blob/6756dee29521b0bf933daa0f43254e269b76dd50/velox/experimental/cudf/exec/ToCudf.h#L50-L59 As for the reasoning behind this design, Gluten wants to be able to enable/disable cudf by passing arguments to registerCudf https://github.com/facebookincubator/velox/pull/12735#discussion_r2013394732. 

I do think this is a code smell and would prefer if users like gluten would check the cudf enable/disable condition outside of calling `registerCudf` and simply not call it when cudf execution isn't needed.

velox/benchmarks/QueryBenchmarkBase.cpp
devavret: Our aim here is to use `TpchBenchmark` which derives from `QueryBenchmarkBase`. If I add a `CudfQueryBenchmarkBase` then I'll have to repeat the code in TpchBenchmark.cpp for cudf.

majetideepak: @devavret Another way is to create a test util API for cudf and add the `#ifdef VELOX_ENABLE_CUDF` calls in the test util. The `#else` part will be empty.
This way, we can keep this file clean and reuse the test util for other places like examples, etc.
We follow this pattern for registering file system adapters. The client code is clean this way.

jinchengchenghh: We may also have CudfTpchBenchmark, in case of we may add wave, spark tpch benchmark or other extensions.
`class CudfTpchBenchmark : publicTpchBenchmark, public CudfQueryBenchmarkBase`

devavret: How about now https://github.com/facebookincubator/velox/pull/13695/files/e4915773e30d8f63b826771f68af438fc782e845..2ea6ec09725f399d7afe655752ddb7864879173b

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+    QueryBenchmarkBase::initialize();
+    initQueryBuilder();
+  }
+
+  void shutdown() override {
+    QueryBenchmarkBase::shutdown();
+    queryBuilder.reset();
+  }
+
+  void runMain(std::ostream& out, facebook::velox::RunStats& runStats)
```
jinchengchenghh: Please move the implement to cpp file.

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+#include <gflags/gflags.h>
+
+#include "velox/benchmarks/QueryBenchmarkBase.h"
+
+DECLARE_int32(run_query_verbose);
+DECLARE_int32(io_meter_column_pct);
+
+class TpchBenchmark : public facebook::velox::QueryBenchmarkBase {
+ public:
+  void initialize() override {
```
jinchengchenghh: ditto

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+
+  void runQuery(int32_t queryId) {
+    const auto planContext = queryBuilder->getQueryPlan(queryId);
+    run(planContext, queryConfigs);
+  }
+
+  void initQueryBuilder();
+
+ private:
+  std::shared_ptr<facebook::velox::exec::test::TpchQueryBuilder> queryBuilder;
```
jinchengchenghh: queryBuilder -> queryBuilder_

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+
+ protected:
```
jinchengchenghh: Place the protected member before private member

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+          << std::endl;
+    }
+  }
+
+  void runQuery(int32_t queryId) {
+    const auto planContext = queryBuilder->getQueryPlan(queryId);
+    run(planContext, queryConfigs);
+  }
+
+  void initQueryBuilder();
```
jinchengchenghh: Move to private scope

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,83 @@
...
+
+ private:
+  std::shared_ptr<facebook::velox::exec::test::TpchQueryBuilder> queryBuilder;
+
+ protected:
+  std::unordered_map<std::string, std::string> queryConfigs;
```
jinchengchenghh: ditto

velox/experimental/cudf/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 add_subdirectory(exec)
+add_subdirectory(benchmarks)
```
jinchengchenghh: Please respect VELOX_ENABLE_BENCHMARKS

velox/experimental/cudf/benchmarks/CudfTpchBenchmark.cpp
```diff
@@ -0,0 +1,125 @@
...
+    connector::unregisterConnectorFactory(
+        cudf_velox::connector::parquet::ParquetConnectorFactory::
+            kParquetConnectorName);
+    TpchBenchmark::shutdown();
+  }
+};
+
+int main(int argc, char** argv) {
+  std::string kUsage(
+      "This program benchmarks TPC-H queries. Run 'velox_tpch_benchmark -helpon=TpchBenchmark' for available options.\n");
```
bdice:
```suggestion
      "This program benchmarks TPC-H queries. Run 'velox_cudf_tpch_benchmark -helpon=TpchBenchmark' for available options.\n");
```

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,35 @@
  */
 #pragma once
 
+#include <gflags/gflags.h>
+
+#include "velox/benchmarks/QueryBenchmarkBase.h"
+
+DECLARE_int32(run_query_verbose);
+DECLARE_int32(io_meter_column_pct);
```
majetideepak: Do we need this in the header? We can then remove gflags.h from the header.

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,35 @@
  */
 #pragma once
 
+#include <gflags/gflags.h>
+
+#include "velox/benchmarks/QueryBenchmarkBase.h"
+
+DECLARE_int32(run_query_verbose);
```
majetideepak: Do we need this in the header?

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,35 @@
...
+DECLARE_int32(io_meter_column_pct);
+
+class TpchBenchmark : public facebook::velox::QueryBenchmarkBase {
+ public:
+  void initialize() override;
+
+  void shutdown() override;
+
+  void runMain(std::ostream& out, facebook::velox::RunStats& runStats) override;
```
majetideepak, line 27: Since we are not overriding this for cudf, we can keep this local for now.

velox/benchmarks/tpch/TpchBenchmark.cpp
```diff
@@ -57,178 +57,164 @@ DEFINE_int32(
...
 
 BENCHMARK(q1) {
-  const auto planContext = queryBuilder->getQueryPlan(1);
-  benchmark.run(planContext);
+  benchmark->runQuery(1);
 }
 
 BENCHMARK(q2) {
-  const auto planContext = queryBuilder->getQueryPlan(2);
-  benchmark.run(planContext);
```
majetideepak, line 119: Let's ensure `VELOX_NOT_NULL(benchmark)` here and below?

velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -276,14 +313,31 @@ core::PlanNodePtr PlanBuilder::TableScanBuilder::build(core::PlanNodeId id) {
...
+            cudf_velox::connector::parquet::ParquetTableHandle>(
+            cudf_velox::exec::test::kParquetConnectorId,
+            tableName_,
+            subfieldFilterExpr != nullptr,
+            subfieldFilterExpr,
+            remainingFilterExpr,
+            dataColumns_);
+      }
+#endif
+      return std::make_shared<HiveTableHandle>(
```
majetideepak: We want to avoid `VELOX_ENABLE_CUDF` in the PlanBuilder.
1) Can we convert `HiveTableHandle` -> `udf_velox::connector::parquet::ParquetTableHandle` on the cudf side?
2) We could adopt a factory for the TableHandle. Some related work here https://github.com/facebookincubator/velox/pull/13798/files

majetideepak: PlanBuilder is expected to be moved to core Velox. Adding VELOX_ENABLE_CUDF complicates releasing Velox libraries in the future.

jinchengchenghh (approved):

velox/experimental/cudf/tests/TableScanTest.cpp
```diff
@@ -327,14 +327,27 @@ TEST_F(TableScanTest, filterPushdown) {
...
+      std::make_shared<core::ConstantTypedExpr>(BIGINT(), int64_t(0)));
+
+  auto c3Expr = std::make_shared<core::CallTypedExpr>(
+      BOOLEAN(),
+      "eq",
+      std::make_shared<core::FieldAccessTypedExpr>(BOOLEAN(), "c3"),
+      std::make_shared<core::ConstantTypedExpr>(BOOLEAN(), true));
+
+  auto subfieldFilterExpr =
+      std::make_shared<core::CallTypedExpr>(BOOLEAN(), "and", c1Expr, c3Expr);
```
devavret: this looks like a merge issue. we now want SubfieldFilters

karthikeyann (changes requested):
Since this PR is stale for long, it needs cleanup, and addressing review comments.
I merged main to this PR, and made sure velox_cudf_tpch_benchmark builds.

velox/benchmarks/tpch/TpchBenchmark.h
```diff
@@ -15,4 +15,31 @@
  */
 #pragma once
 
+#include "velox/benchmarks/QueryBenchmarkBase.h"
+#include "velox/exec/tests/utils/TpchQueryBuilder.h"
+
+class TpchBenchmark : public facebook::velox::QueryBenchmarkBase {
```
jinchengchenghh, line 21: Add TpchBenchmark virtual destructor since CudfTpchBenchmark inherits it.

majetideepak (approved):

velox/benchmarks/tpch/TpchBenchmark.h
devavret: QueryBenchmarkBase already has a virtual destructor

jinchengchenghh:
Looks good to me

karthikeyann:
@kagamiori is the merging blocked due to failing tests?
The changes in this PR are unrelated to Presto Fuzzer.

PR #13700, opened 2025-06-09
fix: Fix binomial_cdf disparity

skyelves:
Summary:
Fix binomial_cdf disparity

T226113911

In java, if value >= numberOfTrials, directly return 1.0
```
public double cumulativeProbability(int x) {
        double ret;
        if (x < 0) {
            ret = (double)0.0F;
        } else if (x >= this.numberOfTrials) {
            ret = (double)1.0F;
        } else {
            ret = (double)1.0F - Beta.regularizedBeta(this.probabilityOfSuccess, (double)x + (double)1.0F, (double)(this.numberOfTrials - x));
        }

        return ret;
    } 
```

In cpp, it throws an error in boost lib

``` 
if(k > N)
{
       *result = policies::raise_domain_error<RealType>(
          function,
          "Number of Successes argument is %1%, but must be <= Number of Trials !", k, pol);
       return false;
}
```

Differential Revision: D76301965

### DIFF
--- a/velox/functions/prestosql/Probability.h
+++ b/velox/functions/prestosql/Probability.h
@@ -97,6 +97,11 @@ struct BinomialCDFFunction {
       return;
     }
 
+    if (value >= numOfTrials) {
+      result = 1.0;
+      return;
+    }
+
     boost::math::binomial_distribution<> dist(numOfTrials, successProb);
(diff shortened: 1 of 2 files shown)


amitkdutta (approved):
Thanks @skyelves

Issue #12664, opened 2025-03-15
Implement createNegatedBytesValuesFilter for Iceberg equality delete


yingsu00:
### Description

In velox/connectors/hive/iceberg/FilterUtil.cpp, we create a NOT IN filter as the domain filter when the equality delete only contains one field. VARCHAR and VARBINARY need to be supported

Joe-Abraham:
closing since this is handled in  https://github.com/facebookincubator/velox/pull/17076

PR #11088, opened 2024-09-25
Support reading Iceberg split with equality deletes

yingsu00:
This PR introduces EqualityDeleteFileReader, which is used to read
Iceberg splits with equality delete files.

PR #13721, opened 2025-06-10
feat(connector): Add support for varchar and varbinary to Iceberg equality delete

Joe-Abraham:
Fixes https://github.com/facebookincubator/velox/issues/12664

Extension of 
- https://github.com/facebookincubator/velox/pull/11088

### DIFF
--- a/velox/connectors/hive/iceberg/FilterUtil.cpp
+++ b/velox/connectors/hive/iceberg/FilterUtil.cpp
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/connectors/hive/iceberg/FilterUtil.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+using namespace facebook::velox::exec;
+using namespace facebook::velox::core;
+
+template <typename T>
+std::unique_ptr<common::Filter> createNegatedBigintValuesFilter(
+    const VectorPtr& valuesVector,
+    vector_size_t offset,
+    vector_size_t size) {
+  auto valuesPair =
+      common::deDuplicateValues<int64_t, T>(valuesVector, offset, size);
+
+  const auto& values = valuesPair.first;
+  bool hasNull = valuesPair.second;
+
+  return common::createNegatedBigintValues(values, !hasNull);
+}
+
+std::unique_ptr<common::Filter> createNotInFilter(
+    const VectorPtr& elements,
+    vector_size_t offset,
+    vector_size_t size,
+    TypeKind type) {
+  std::unique_ptr<common::Filter> filter;
+  switch (type) {
+    case TypeKind::HUGEINT:
+      // TODO: createNegatedHugeintValuesFilter is not implemented yet.
+      VELOX_NYI("createNegatedHugeintValuesFilter is not implemented yet");
+    case TypeKind::BIGINT:
+      filter = createNegatedBigintValuesFilter<int64_t>(elements, offset, size);
+      break;
+    case TypeKind::INTEGER:
+      filter = createNegatedBigintValuesFilter<int32_t>(elements, offset, size);
+      break;
+    case TypeKind::SMALLINT:
+      filter = createNegatedBigintValuesFilter<int16_t>(elements, offset, size);
+      break;
+    case TypeKind::TINYINT:
+      filter = createNegatedBigintValuesFilter<int8_t>(elements, offset, size);
+      break;
+    case TypeKind::BOOLEAN:
+      // Hack: using BIGINT filter for bool, which is essentially "int1_t".
+      filter = createNegatedBigintValuesFilter<bool>(elements, offset, size);
+      break;
+    case TypeKind::TIMESTAMP:
+      filter =
+          createNegatedBigintValuesFilter<Timestamp>(elements, offset, size);
+      break;
+    case TypeKind::VARCHAR:
+    case TypeKind::VARBINARY:
+      // TODO: createNegatedBytesValuesFilter is not implemented yet.
+      VELOX_NYI("createNegatedBytesValuesFilter is not implemented yet");
+    case TypeKind::REAL:
+    case TypeKind::DOUBLE:
+    case TypeKind::UNKNOWN:
+      [[fallthrough]];
+    case TypeKind::ARRAY:
+      [[fallthrough]];
+    case TypeKind::MAP:
+      [[fallthrough]];
+    case TypeKind::ROW:
+      [[fallthrough]];
+    default:
+      VELOX_USER_FAIL(
+          "Iceberg equality delete column cannot be of type ", type);
+  }
+  return filter;
+}
+} // namespace facebook::velox::connector::hive::iceberg
--- a/velox/connectors/hive/iceberg/FilterUtil.cpp
+++ b/velox/connectors/hive/iceberg/FilterUtil.cpp
@@ -20,6 +20,21 @@ namespace facebook::velox::connector::hive::iceberg {
 using namespace facebook::velox::exec;
 using namespace facebook::velox::core;
 
+std::unique_ptr<common::Filter> createNegatedBytesValuesFilter(
+    const VectorPtr& valuesVector,
+    vector_size_t offset,
+    vector_size_t size) {
+  auto [values, hasNull] =
+      facebook::velox::common::deDuplicateValues<std::string, StringView>(
+          valuesVector, offset, size);
+
+  VELOX_USER_CHECK(
+      !values.empty() || hasNull,
+      "NOT IN filter must contain at least one non-null value");
+
+  return std::make_unique<common::NegatedBytesValues>(values, !hasNull);
+}
+
 template <typename T>
 std::unique_ptr<common::Filter> createNegatedBigintValuesFilter(
     const VectorPtr& valuesVector,
@@ -66,8 +81,8 @@ std::unique_ptr<common::Filter> createNotInFilter(
       break;
     case TypeKind::VARCHAR:
     case TypeKind::VARBINARY:
-      // TODO: createNegatedBytesValuesFilter is not implemented yet.
-      VELOX_NYI("createNegatedBytesValuesFilter is not implemented yet");
+      filter = createNegatedBytesValuesFilter(elements, offset, size);
+      break;
     case TypeKind::REAL:
     case TypeKind::DOUBLE:
     case TypeKind::UNKNOWN:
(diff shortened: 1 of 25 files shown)


velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -1086,6 +1086,72 @@ TEST_F(HiveIcebergTest, equalityDeletesSingleFileColumn1) {
   assertEqualityDeletes(equalityDeleteVectorMap, equalityFieldIdsMap);
 }
 
+TEST_F(HiveIcebergTest, equalityDeletesWithNegatedVarcharValues) {
```
yingsu00: Can you please update 
```
assertEqualityDeletes(
      const std::unordered_map<int8_t, std::vector<std::vector<int64_t>>>&
          equalityDeleteVectorMap,
      const std::unordered_map<int8_t, std::vector<int32_t>>&
          equalityFieldIdsMap,
      std::string duckDbSql = "",
      std::vector<RowVectorPtr> dataVectors = {})
```
to 
```
template <typename T>
assertEqualityDeletes(
      const std::unordered_map<int8_t, std::vector<std::vector<T>>>&
          equalityDeleteVectorMap,
      const std::unordered_map<int8_t, std::vector<int32_t>>&
          equalityFieldIdsMap,
      std::string duckDbSql = "",
      std::vector<RowVectorPtr> dataVectors = {})
```
then use it accordingly? It has multiple tests built in.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -1228,6 +1228,36 @@ TEST_F(HiveIcebergTest, equalityDeletesMultipleFiles) {
...
+  equalityFieldIdsMap.insert({{0, {1}}, {1, {2}}});
+
+  // Delete rows {0, 1} from c0, {2, 3} from c1, with two equality delete files
+  equalityDeleteVectorMap.insert({{0, {{0, 1}}}, {1, {{2, 3}}}});
+  VELOX_ASSERT_THROW(
+      assertEqualityDeletes(equalityDeleteVectorMap, equalityFieldIdsMap),
+      "Iceberg does not allow DOUBLE or REAL columns as the equality delete columns: c1 : REAL")
+}
+
+TEST_F(HiveIcebergTest, equalityDeletesDoubleThrowsError) {
```
yingsu00: Can you please merge these two as a single test?

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -1307,4 +1309,49 @@ TEST_F(HiveIcebergTest, TestSubFieldEqualityDelete) {
   assertEqualityDeletes(
       icebergSplits.back(), ROW({"c_bigint"}, {BIGINT()}), duckDbSql);
 }
+
+TEST_F(HiveIcebergTest, equalityDeletesWithNegatedVarbinaryValues) {
```
yingsu00: The test cases are too few. Could you please add more tests in analogous to the integer type tests? Also add a test for reading delete files with 2 columns, one integer, one varchar/varbinary.

velox/connectors/hive/iceberg/FilterUtil.cpp
```diff
@@ -20,6 +20,21 @@ namespace facebook::velox::connector::hive::iceberg {
...
+std::unique_ptr<common::Filter> createNegatedBytesValuesFilter(
+    const VectorPtr& valuesVector,
+    vector_size_t offset,
+    vector_size_t size) {
+  auto [values, hasNull] =
+      facebook::velox::common::deDuplicateValues<std::string, StringView>(
+          valuesVector, offset, size);
+
+  VELOX_USER_CHECK(
+      !values.empty() || hasNull,
```
nmahadevuni, line 32: Why is this check required? Its not present in createNegatedBigintValuesFilter. If the values is empty and hasNull is true, we still create the filter which goes against the error message here. If it hasNull, then shouldn't check the values vector size is greater than one?

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -234,16 +234,21 @@ class HiveIcebergTest : public HiveConnectorTestBase {
     ASSERT_TRUE(it->second.peakMemoryBytes > 0);
   }
 
+  template <TypeKind KIND>
```
nmahadevuni, line 228: `template <typename T>` should be enough.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -496,32 +505,113 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
+      T& startingValue) {
     std::vector<RowVectorPtr> vectors;
     vectors.reserve(vectorSizes.size());
 
     vectors.reserve(vectorSizes.size());
     for (int j = 0; j < vectorSizes.size(); j++) {
       auto data = makeContinuousIncreasingValues(
           startingValue, startingValue + vectorSizes[j]);
-      VectorPtr c0 = makeFlatVector<int64_t>(data);
+      VectorPtr c0 = makeFlatVector<T>(data);
```
nmahadevuni: `makeContinuousIncreasingValues` takes only integer arguments. Add a check whether T is integral type here too.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -363,32 +505,113 @@ class HiveIcebergTest : public HiveConnectorTestBase {
     return deleteFilePaths;
   }
 
+  template <typename T>
   std::vector<RowVectorPtr> makeVectors(
-      std::vector<int64_t> vectorSizes,
-      int64_t& startingValue) {
+      std::vector<T> vectorSizes,
```
yingsu00: vectorSizes should not be of type T. It contains the number of elements each vector shall contain and should be an integral type

yingsu00: I know this is an existing issue, but I there are two makeVectors() functions and they seem to have repeating logics. This one makes continuously increasing values for one column, and it allows the number of values in the vectors to be different. The other one on line 907 can construct vectors for multiple columns, but the number of rows in each vector can only be a single value `count`. Do you think it's possible to condense the two makeVectors() functions into one? If it works, please remember to add a comment to explain what it does too. You can create a new commit before introducing the templates. Thanks!

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -697,30 +860,34 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
-    RowVectorPtr deleteFileVectors = makeRowVector(names, vectors);
+    RowVectorPtr const deleteFileVectors = makeRowVector(names, vectors);
 
     auto deleteFilePath = TempFilePath::create();
     writeToFile(deleteFilePath->getPath(), deleteFileVectors);
 
     return deleteFilePath;
   }
 
+  template <TypeKind KIND>
```
yingsu00: Similarly, there are two writeDataFiles() functions that do similar stuff. Could you please condense them into one? Please also add a /// comment for it. Thanks!
```
template <TypeKind KIND>
std::map<std::string, std::shared_ptr<TempFilePath>> writeDataFiles(
      std::map<std::string, std::vector<int64_t>> rowGroupSizesForFiles,  // This contains 1) number of files, 2) each file's name 3) number of rows in each RowGroup in this file
      int32_t numColumns = 1)

File: velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp [hunk 9]
```diff
@@ -546,95 +635,161 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
 
-  std::vector<int64_t> flattenAndDedup(
-      const std::vector<std::vector<int64_t>>& deletePositionVectors,
-      int64_t baseFileSize) {
-    std::vector<int64_t> deletePositionVector;
-    for (auto vec : deletePositionVectors) {
-      for (auto pos : vec) {
-        if (pos >= 0 && pos < baseFileSize) {
-          deletePositionVector.push_back(pos);
+  template <typename T>
```
yingsu00: flattenAndDedup() is only used on row indexes and does not need to be of typename T

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: I think its best to keep the TypeKind here, since the all the parameterization is done based on the TypeKind and make it easier to throw, if it is unsupported kind

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -546,95 +635,161 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
         flattenedDeletePosVectorsForAllBaseFiles;
     int64_t totalNumDeletePositions = 0;
     for (auto deleteVectorsForBaseFile : deletePosVectorsForAllBaseFiles) {
       auto baseFileName = deleteVectorsForBaseFile.first;
       auto deletePositionVectors = deleteVectorsForBaseFile.second;
-      std::vector<int64_t> deletePositionVector =
-          flattenAndDedup(deletePositionVectors, baseFileSizes[baseFileName]);
+
+      // Use the base file size for validation
+      T baseFileSize = baseFileSizes[baseFileName];
```
yingsu00: File sizes do not need to be of type T

yingsu00 (commented):
There are two `protected:` sections in the same class. Could you please squash them together in a separate commit? Thanks!

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -546,95 +595,134 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
 
-    return deletePositionVector;
+    return result;
   }
 
-  std::string makeNotInList(const std::vector<int64_t>& deletePositionVector) {
+  template <TypeKind KIND>
+  std::string makeNotInList(
+      const std::vector<typename TypeTraits<KIND>::NativeType>&
+          deletePositionVector) {
```
yingsu00: deletePositionVector -> deleteValues

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
yingsu00: Could you please remove makeContinuousIncreasingValues() and replace it with makeSequenceValues<T>()?

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: Done.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: Reverted the code.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: Reverted the code.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: Combining the two makeVectors removed this.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
Joe-Abraham: Removed makeContinuousIncreasingValues().

PR #13730, opened 2025-06-11
fix: Fix TableWriterArbitrationTest.reclaimFromTableWriter

tanjialiang:
Summary:
The test was "broken" by D75535212 which changes folly's randomness generation. The test broke because with previous version of folly random, the generated data for table writer test is just under the size of memory capacity, not triggering cap exceed errors. Now with the new folly random, the size went above the capacity, triggering the cap exceed error.

The test is fixed by reducing the writer memory footprint by reducing the generated string size.

Differential Revision: D76387734


kagamiori (approved):
Thank you for the quick fix!

Issue #13731, opened 2025-06-11
AVG(decimal38) does not round up in overflow cases


xiaodouchen:
### Bug description

When I tested avg(decimal38), I found the result is wrong.
```
CREATE TABLE test_agg_decimal38 (
  decimal38 decimal(38, 7);
)

insert into test_agg_decimal38 values
(9999999999999999999999999999999.9999999),
(9999999999999999999999999999999.9999998),
(9999999999999999999999999999999.9999997),
(9999999999999999999999999999999.9999995),
(9999999999999999999999999999999.9999995),
(9999999999999999999999999999999.9999995);

SELECT avg(decimal38) from test_agg_decimal38;
```
[Expected behavior] 
9999999999999999999999999999999.9999997

[actual behavior].
9999999999999999999999999999999.9999996

I found that in `DecimalUtil::computeAverage`, the calculation of `remTotal` does not perform a round up, but it should theoretically.

### System information

Velox System Info v0.0.2
Commit: a968af52ef136158eaf3c6521c9b669809371826
CMake Version: 3.28.3
System: Linux-4.9.168-019.ali3000.alios7.x86_64
Arch: x86_64
C++ Compiler: /usr/bin/g++-11
C++ Compiler Version: 11.4.0
C Compiler: /usr/bin/gcc-11
C Compiler Version: 11.4.0
CMake Prefix Path: /usr/local;/usr;/;/code/cmake-3.28.3-linux-x86_64;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

mbasmanova:
CC: @amitkdutta

PR #13732, opened 2025-06-11
fix: Fix DecimalUtil::computeAverage rounding in overflow cases

xiaodouchen:
In DecimalUtil::computeAverage(), the calculation of remTotal does not round up, but it should theoretically round up to avoid accuracy issues in the results.

Fixes https://github.com/facebookincubator/velox/issues/13731

### DIFF
--- a/velox/docs/functions/presto/aggregate.rst
+++ b/velox/docs/functions/presto/aggregate.rst
@@ -31,11 +31,20 @@ General Aggregate Functions
     inputs if :doc:`presto.array_agg.ignore_nulls <../../configs>` is set
     to false.
 
-.. function:: avg(x) -> double|real
+.. function:: avg(x) -> double|real|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
     When x is of type REAL, the result type is REAL.
-    For all other input types, the result type is DOUBLE.
+    When x is an integer or a DOUBLE, the result is DOUBLE.
+    When x is of type DECIMAL(p, s), the result type is DECIMAL(p, s).
+    Note: For the overflow cases, Velox returns a result when Presto throws "Decimal overflow". ::
+        SELECT AVG(col)
+        FROM ( VALUES
+        	  (CAST(9999999999999999999999999999999.9999999 AS DECIMAL(38,7))),
+              (CAST(9999999999999999999999999999999.9999999 AS DECIMAL(38,7)))
+             ) AS t(col);
+        -- Velox: 9999999999999999999999999999999.9999999
+        -- Presto: Decimal overflow
 
 .. function:: bool_and(boolean) -> boolean
 
--- a/velox/type/DecimalUtil.cpp
+++ b/velox/type/DecimalUtil.cpp
@@ -98,6 +98,17 @@ void DecimalUtil::computeAverage(
     divideWithRoundUp<int128_t, int128_t, int64_t>(
         avg, sum, count, false, 0, 0);
   } else {
+    // The average is computed as `(sum + overflow * kOverflowMultiplier) /
+    // count`:
+    //  1. sum / count → quotient quotSum, remainder remSum.
+    //  2. (overflow * kOverflowMultiplier) / count → quotient quotMul,
+    //  remainder remMul.
+    //  3. (remSum + remMul) / count -> remTotal
+    //  4. final result = quotSum + quotMul + remTotal
+    // Since we only use quotients (no rounding needed, set "noRoundUp = true")
+    // and combine remainders to compute the final average, rounding is applied
+    // based on the combined remainder values (set "noRoundUp = false"), which
+    // ensures accuracy.
     VELOX_DCHECK_LE(overflow, count);
     __uint128_t quotMul, remMul;
     __int128_t quotSum, remSum;
@@ -109,7 +120,7 @@ void DecimalUtil::computeAverage(
     remSum = DecimalUtil::divideWithRoundUp<__int128_t, __int128_t, int64_t>(
         quotSum, sum, count, true, 0, 0);
     DecimalUtil::divideWithRoundUp<__int128_t, __int128_t, int64_t>(
-        remTotal, remMul + remSum, count, true, 0, 0);
+        remTotal, remMul + remSum, count, false, 0, 0);
     avg = quotMul + quotSum + remTotal;
   }
 }
--- a/velox/functions/prestosql/aggregates/tests/AverageAggregationTest.cpp
+++ b/velox/functions/prestosql/aggregates/tests/AverageAggregationTest.cpp
@@ -482,6 +482,24 @@ TEST_F(AverageAggregationTest, avgDecimal) {
       {makeRowVector(
           {makeFlatVector(std::vector<int64_t>{337}, DECIMAL(3, 2))})});
 
+  // Round-up average when sum overflows the max int128_t limit.
+  testAggregations(
+      {makeRowVector({makeFlatVector<int128_t>(
+          {HugeInt::parse("99999999999999999999999999999999999999"),
+           HugeInt::parse("99999999999999999999999999999999999998"),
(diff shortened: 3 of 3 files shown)


xiaodouchen:
@Yuhta @mbasmanova @karteekmurthys Please help review it.

mbasmanova:
CC: @majetideepak 

Deepak, Amit, do you know what's Presto Java behavior?

mbasmanova:
@xiaodouchen Would you rebase to clear Fuzzer failure?

CC: @kagamiori @bikramSingh91

xiaodouchen:
> @xiaodouchen Would you rebase to clear Fuzzer failure?
> 
> CC: @kagamiori @bikramSingh91

Done.

mbasmanova:
@xiaodouchen Would you check Presto Java's behavior? Does it match the new behavior in this PR?

xiaodouchen:
> @xiaodouchen Would you check Presto Java's behavior? Does it match the new behavior in this PR?

@mbasmanova  It will throw "Decimal overflow" in Presto Java. It does not match the old/new behavior in velox.

```
presto> SELECT AVG(col1) FROM (
     ->   VALUES
     ->     (CAST(9999999999999999999999999999999.9999999 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999998 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999997 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7)))
     -> ) AS t(col1);

Query 20250616_022601_00004_huyj9, FAILED, 1 node
Splits: 1 total, 0 done (0.00%)
[Latency: client-side: 121ms, server-side: 68ms] [0 rows, 0B] [0 rows/s, 0B/s]

Query 20250616_022601_00004_huyj9 failed: Decimal overflow
```


However, in Presto Native, we can get the result as follows.
```
presto> SELECT AVG(col1) FROM (
     ->   VALUES
     ->     (CAST(9999999999999999999999999999999.9999999 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999998 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999997 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7))),
     ->     (CAST(9999999999999999999999999999999.9999995 AS DECIMAL(38,7)))
     -> ) AS t(col1);
                  _col0                  
-----------------------------------------
 9999999999999999999999999999999.9999996 
(1 row)
```

mbasmanova:
@xiaodouchen Thank you for checking Presto's behavior. I see that Velox deviates and returns a result when Presto throws an overflow error. I guess Velox behavior makes more sense and we would accept this deviation. @amitkdutta Amit, any thoughts?

@xiaodouchen Let's update documentation to clarify this discrepancy.

velox/type/DecimalUtil.cpp
```diff
@@ -109,7 +109,7 @@ void DecimalUtil::computeAverage(
     remSum = DecimalUtil::divideWithRoundUp<__int128_t, __int128_t, int64_t>(
         quotSum, sum, count, true, 0, 0);
     DecimalUtil::divideWithRoundUp<__int128_t, __int128_t, int64_t>(
-        remTotal, remMul + remSum, count, true, 0, 0);
+        remTotal, remMul + remSum, count, false, 0, 0);
```
mbasmanova, line 123: This code calls divideWithRoundUp 2 times. Twice with 'true' and once with 'false'. The reader of this could would have a hard time understanding why. Would it be possible to add comments to explain how we choose the value for that boolean?

xiaodouchen:
> @xiaodouchen Let's update documentation to clarify this discrepancy.

@mbasmanova Sure, I can add a clarification to the documentation. Could you please let me know which file would be more appropriate — decimal.rst or aggregate.rst?

mbasmanova:
@xiaodouchen I think we need to update these docs: https://facebookincubator.github.io/velox/functions/presto/aggregate.html#avg

velox/type/DecimalUtil.cpp
xiaodouchen: @mbasmanova I've added comments to explain the boolean logic. Could you review them and let me know if any adjustments are needed?
```
    // The average is computed as `(sum + overflow * kOverflowMultiplier) /
    // count`:
    //  1. sum / count → quotient quotSum, remainder remSum.
    //  2. (overflow * kOverflowMultiplier) / count → quotient quotMul,
    //  remainder remMul.
    //  3. (remSum + remMul) / count -> remTotal
    //  4. final result = quotSum + quotMul + remTotal
    // Since we only use quotients (no rounding needed, set "noRoundUp = true")
    // and combine remainders to compute the final average, rounding is applied
    // based on the combined remainder values (set "noRoundUp = false"), which
    // ensures accuracy.
```

mbasmanova: @xiaodouchen This is helpful. Thanks!

velox/docs/functions/presto/aggregate.rst
```diff
@@ -31,11 +31,13 @@ General Aggregate Functions
     inputs if :doc:`presto.array_agg.ignore_nulls <../../configs>` is set
     to false.
 
-.. function:: avg(x) -> double|real
+.. function:: avg(x) -> double|real|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
     When x is of type REAL, the result type is REAL.
-    For all other input types, the result type is DOUBLE.
+    When x is of type ALL INTs / DOUBLE, the result type is DOUBLE.
```
mbasmanova: > ALL INTs / DOUBLE

Let's clarify: When x is an integer or a DOUBLE, the result is DOUBLE.

> When x is of type DECIMAL, the result type is DECIMAL.

Let's clarify about precision and scale: When x is of type DECIMAL(p, s), the result type is DECIMAL(p, s).

velox/docs/functions/presto/aggregate.rst
```diff
@@ -31,11 +31,20 @@ General Aggregate Functions
...
-.. function:: avg(x) -> double|real
+.. function:: avg(x) -> double|real|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
     When x is of type REAL, the result type is REAL.
-    For all other input types, the result type is DOUBLE.
+    When x is an integer or a DOUBLE, the result is DOUBLE.
+    When x is of type DECIMAL(p, s), the result type is DECIMAL(p, s).
+    Note: For the overflow cases, Velox returns a result when Presto throws "Decimal overflow". ::
+        SELECT AVG(col)
```
mbasmanova, line 41: @xiaodouchen Thank you for adding an example. Would you generate the docs to verify that formatting is looking good?

velox/docs/functions/presto/aggregate.rst
xiaodouchen: @mbasmanova Done. 
In addition, I've added an example. Could you review it again?

velox/docs/functions/presto/aggregate.rst
xiaodouchen: @mbasmanova Yes, I generate the docs on mac and it looks good.
image

mbasmanova: Perfect. Thanks.

amitkdutta:
> Thank you for checking Presto's behavior. I see that Velox deviates and r

@mbasmanova This deviation is fine. Velox behavior is the `default` behavior for Presto, also it's correct here.

PR #13734, opened 2025-06-11
fix: Support constant value for lead/lag function

philo-he:
Example: `lag(100, c1, c2)`, where the first arg is not column reference.

### DIFF
--- a/velox/functions/prestosql/window/LeadLag.cpp
+++ b/velox/functions/prestosql/window/LeadLag.cpp
@@ -30,7 +30,11 @@ class LeadLagFunction : public exec::WindowFunction {
       bool ignoreNulls,
       velox::memory::MemoryPool* pool)
       : WindowFunction(resultType, pool, nullptr), ignoreNulls_(ignoreNulls) {
-    valueIndex_ = args[0].index.value();
+    if (args[0].constantValue) {
+      constantValue_ = args[0].constantValue;
+    } else {
+      valueIndex_ = args[0].index.value();
+    }
 
     initializeOffset(args);
     initializeDefaultValue(args);
@@ -44,6 +48,12 @@ class LeadLagFunction : public exec::WindowFunction {
     ignoreNullsForPartition_ = false;
 
     if (ignoreNulls_) {
+      if (constantValue_) {
+        // The special ignoreNulls processing is not required if it's not null
+        // constant.
+        ignoreNullsForPartition_ = constantValue_->isNullAt(0);
+        return;
+      }
       auto partitionSize = partition_->numRows();
       AlignedBuffer::reallocate<bool>(&nulls_, partitionSize);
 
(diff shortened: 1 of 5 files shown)


velox/functions/prestosql/window/LeadLag.cpp
```diff
@@ -30,7 +30,11 @@ class LeadLagFunction : public exec::WindowFunction {
       bool ignoreNulls,
       velox::memory::MemoryPool* pool)
       : WindowFunction(resultType, pool, nullptr), ignoreNulls_(ignoreNulls) {
-    valueIndex_ = args[0].index.value();
+    if (args[0].constantValue) {
+      constantValue_ = args[0].constantValue;
```
jinchengchenghh, line 34: Could we use std::move here?

jinchengchenghh (commented):
Thanks
Do you need to update the document? https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/presto/window.rst

aditi-pandit:
@PHILO-HE : Thanks for this code.

I'm trying to understand where your optimization is applicable...Is this when the lead has a constant argument instead of a column ? Is that allowed ? It would mean simply projecting that value itself. I think a reasonable alternative is to error on constant argument also. Please can you explain.

philo-he:
@aditi-pandit, thanks for your comment!
> Is this when the lead has a constant argument instead of a column ?

Yes, this pr is not for optimization purpose, but supporting constant 1st argument for lead/lag. It is permitted in Spark, and our users have workloads that include such use cases.

velox/functions/prestosql/window/LeadLag.cpp
philo-he: Seems not quite necessary to use std::move?

philo-he:
> Thanks Do you need to update the document? https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/presto/window.rst

@jinchengchenghh, just updated the corresponding Spark doc. Thanks!

jinchengchenghh:
The code changed Prestosql behavior, does presto also has this case?

aditi-pandit:
@jinchengchenghh : Verified that Presto supports it as well....

select lead(5, 0) over (partition by orderkey order by orderdate) from orders; returned me a column with all rows just '5'

@PHILO-HE : Can we ensure that the WindowFuzzer generates such expressions though ? Would make it easier to review.

jinchengchenghh:
Thanks for your test. @aditi-pandit 
Do you also need to update presto document? @PHILO-HE

aditi-pandit:
@PHILO-HE : Can you update the WindowFuzzer in this PR to ensure it generates lead/lag sql that matches this usage ? Would be great to see the logs from your run showing these kind of queries executing.

philo-he:
I also did a local test. Yes, Presto supports such literal constant input. 

> Do you also need to update presto document?

@jinchengchenghh, yes, I just updated. Thanks!

> Can we ensure that the WindowFuzzer generates such expressions though ? Would make it easier to review.

@aditi-pandit, in the existing window fuzzer code, constant input can be produced, but such input is represented by constant vector which is still accessed via column reference. I just added some code to cover such cases which requires CI validation. Thanks!

jinchengchenghh:
Is the CI failure related to your change?

philo-he:
> Is the CI failure related to your change?

@jinchengchenghh, it's related. I will fix it. Thanks!

philo-he:
FYI: duplicate to https://github.com/facebookincubator/velox/pull/16915, which has already been merged.

PR #13735, opened 2025-06-11
fix: Apply query config to constant folding

mbasmanova:
Summary: Modify velox::exec::tryEvaluateConstantExpression to require the user to pass velox::core::QueryCtx. Using 'default' query context is incorrect as some functions may depend on configs for execution. These functions may either produce wrong results and fail outright with default query context.

Differential Revision: D76423739


xiaoxmeng (approved):
@mbasmanova thanks for the quick fix!

PR #13746, opened 2025-06-12
fix(noisy_count_if): Fix some consistency issue and check overflow

oliver6782:
Summary:
#### Fixed noisy_count_if Function

The change fixes the `noisy_count_if` function to address consistency issues and prevent overflow. it adds checks to ensure the rounded noise value is within the range of `int64_t` before casting. This prevents potential overflow when casting from a double to an `int64_t`. The fix is applied in the `NoisyCountIfGaussianAggregate.cpp` file.

Differential Revision: D76479277


amitkdutta (approved):
Thanks @oliver6782

PR #13755, opened 2025-06-12
fix: Fix resource release for memoizing constant folding expression

rui-mo:
The `memoizingExprs_` of ExprSet is used to  track expressions that have been 
evaluated with dictionary memoization. This ensures that when the ExprSet is 
cleared, it can call `clearMemo` on each memoizing expression to release cached 
resources.

However, the constant folding uses an uninitialized ExprSet for eval. This 
breaks the invariant that `memoizingExprs_` relies on, which is that the Expr 
pointers will be alive for the lifetime of the ExprSet.

To fix this issue, this PR clears the execution state, including the
`memoizingExprs_`, of the ExprSet to avoid dangling Expr pointers being cached.

For reference, the stack was:

```
C  [libvelox.dylib+0x19350]  long std::__1::__libcpp_atomic_refcount_decrement[abi:ne180100]<long>(long&)+0x18
C  [libvelox.dylib+0x1929c]  std::__1::__shared_weak_count::__release_shared[abi:ne180100]()+0x1c
C  [libvelox.dylib+0x301b8]  std::__1::shared_ptr<facebook::velox::BaseVector>::~shared_ptr[abi:ne180100]()+0x40
C  [libvelox.dylib+0x30168]  std::__1::shared_ptr<facebook::velox::BaseVector>::~shared_ptr[abi:ne180100]()+0x1c
C  [libvelox.dylib+0xe7edc]  std::__1::shared_ptr<facebook::velox::BaseVector>::operator=[abi:ne180100](std::__1::shared_ptr<facebook::velox::BaseVector>&&)+0x40
C  [libvelox.dylib+0xf1c4f4]  facebook::velox::exec::Expr::clearMemo()+0x5c
C  [libvelox.dylib+0xfc032c]  facebook::velox::exec::ExprSet::clear()+0x70
C  [libvelox.dylib+0x8bb0fc0]  facebook::velox::exec::FilterProject::close()+0x44
C  [libvelox.dylib+0x8b5616c]  facebook::velox::exec::Driver::closeOperators()+0x84
C  [libvelox.dylib+0x8b557b0]  facebook::velox::exec::Driver::close()+0xac
```

### DIFF
--- a/velox/expression/Expr.h
+++ b/velox/expression/Expr.h
@@ -220,9 +220,9 @@ class Expr {
 
   void clearMemo() {
     baseOfDictionaryRepeats_ = 0;
-    baseOfDictionary_.reset();
-    baseOfDictionaryWeakPtr_.reset();
     baseOfDictionaryRawPtr_ = nullptr;
+    baseOfDictionaryWeakPtr_.reset();
+    baseOfDictionary_.reset();
     dictionaryCache_ = nullptr;
     cachedDictionaryIndices_ = nullptr;
   }
--- a/velox/expression/Expr.cpp
+++ b/velox/expression/Expr.cpp
@@ -1970,8 +1970,7 @@ void ExprSet::clear() {
   for (auto* memo : memoizingExprs_) {
     memo->clearMemo();
   }
-  distinctFields_.clear();
-  multiplyReferencedFields_.clear();
+  memoizingExprs_.clear();
 }
 
 void ExprSet::clearCache() {
--- a/velox/expression/ExprCompiler.cpp
+++ b/velox/expression/ExprCompiler.cpp
@@ -520,12 +520,21 @@ ExprPtr compileRewrittenExpression(
 
   result->computeMetadata();
 
+  ExprPtr compiled;
   // If the expression is constant folding it is redundant.
-  auto folded = enableConstantFolding && !isConstantExpr
-      ? tryFoldIfConstant(result, scope)
-      : result;
-  scope->visited[expr.get()] = folded;
-  return folded;
+  if (enableConstantFolding && !isConstantExpr) {
+    compiled = tryFoldIfConstant(result, scope);
+    // Constant folding uses an uninitialized ExprSet for eval. This breaks the
+    // invariant that 'memoizingExprs_' relies on, which is that the Expr
+    // pointers will be alive for the lifetime of the ExprSet. Clear the
+    // execution state of the ExprSet to avoid this.
+    scope->exprSet->clear();
+  } else {
+    compiled = result;
+  }
+
+  scope->visited[expr.get()] = compiled;
+  return compiled;
 }
 
 ExprPtr compileExpression(
--- a/velox/expression/tests/ExprTest.cpp
+++ b/velox/expression/tests/ExprTest.cpp
@@ -5086,5 +5086,21 @@ TEST_F(ExprTest, peelingOnDeterministicFunctionInNonDeterministicExpr) {
   ASSERT_EQ(stats["plus"].numProcessedRows, input->size() / 2);
 }
 
+TEST_F(ExprTest, lambdaConstantFolded) {
+  const auto makeTypedExpr =
+      [this](const std::string& text, const RowTypePtr& rowType) {
+        auto untyped = parse::parseExpr(text, {});
+        return core::Expressions::inferTypes(untyped, rowType, pool());
+      };
+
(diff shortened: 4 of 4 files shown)


velox/exec/tests/FilterProjectTest.cpp
```diff
@@ -482,3 +482,20 @@ TEST_F(FilterProjectTest, barrier) {
...
+      makeNestedArrayVectorFromJson<std::string>({"[[\"h\"]]"}),
+  });
+
+  auto plan =
+      PlanBuilder(pool_.get())
+          .values({data})
+          .project(
+              {"reduce(regexp_split('a,b,c,d,e,f,g,h',','), array[array['']], (acc, x) -> array[array[x]], (id) -> id) as a"})
+          .planNode();
+  AssertQueryBuilder(plan).assertResults(expected);
```
rui-mo: For some reason, the core dump issue cannot be reproduced in the unit test, but we verified this PR could fix the core dump issue in Gluten.

velox/expression/Expr.h
```diff
@@ -220,9 +220,9 @@ class Expr {
 
   void clearMemo() {
     baseOfDictionaryRepeats_ = 0;
-    baseOfDictionary_.reset();
-    baseOfDictionaryWeakPtr_.reset();
     baseOfDictionaryRawPtr_ = nullptr;
+    baseOfDictionaryWeakPtr_.reset();
+    baseOfDictionary_.reset();
```
rui-mo, line 225: This is a minor improvement for the release order.

rui-mo:
Hi @mbasmanova, would you like to help review this fix? Thanks!

mbasmanova:
CC: @bikramSingh91

velox/expression/Expr.h
```diff
@@ -770,6 +774,12 @@ class ExprSet {
 
   // Exprs which retain memoized state, e.g. from running over dictionaries.
   std::unordered_set<Expr*> memoizingExprs_;
+
+  // Exprs that will be used for constant folding during expression compilation.
+  // Stores them to extend their lifetime so that resources can be correctly
+  // freed when their raw pointers are memoized.
+  std::vector<std::shared_ptr<Expr>> unfoldedExprs_;
```
bikramSingh91: I was looking at ways to avoid adding additional state to the ExprSet to manage, in that process I looked at how the ExprCompiler uses ExprSet and noticed a peculiarity: It uses an uninitialized ExprSet for eval of constant folding. This breaks the invariant that memoizingExprs_ relies on, which is that the Expr pointers will be alive for the lifetime of the ExprSet.
Using a half initialized Expression set to evaluate is risky in on itself. In that context, we should probably clear execution state out of the ExprSet after it is used in this manner. How about we modify Expr::clear() to serve that function. And let it clear this execution state, that is, clear shared sub-expression and memoized state, along with clearing memoizingExprs_ as its populated during eval. Clearing distinctFields_ and multiplyReferencedFields_ can be left out since they would vanish anyways after ExprSet is deleted and its metadata not accumilated during execution. We can then call ExprSet::clear() after using it for constant folding.

rui-mo: Thanks for your inputs! This can indeed solve the issue without maintaining additional state in the ExprSet. I updated this PR following your suggestion and verified it in Gluten.

velox/exec/tests/FilterProjectTest.cpp
bikramSingh91: @rui-mo 

I was able to repro the crash using the following test, can you please replace this test with the following and add it to ExprTest.
Thanks

```
TEST_F(ExprTest, lambdaConstantFolded) {
 // <add context about test here>
  std::string expression =
      "reduce(regexp_split('a,b,c,d,e,f,g,h',','), array[array['']], (acc, x) -> array[array[x]], (id) -> id)";
  auto typedExpr = makeTypedExpr(expression, ROW({"c0"}, {INTEGER()}));
  exec::ExprSet exprSet({typedExpr}, &execCtx_);
  exprSet.clear();
}
```

rui-mo: @bikramSingh91 Thanks for the suggestion! I added this test to ExprTest, but I'm still unable to repro the crash on my end using this test.

bikramSingh91: can you try running in in a debug build? Sometimes optimized build dont crash on these issues. In any case, let me approve the change.

PR #4307, opened 2023-03-16
Allow passing context to ISerializable::deserialize

mbasmanova:
A follow-up PR adds serialization for expressions. To serialize
ConstantTypeExpr, we use VectorSaver to serialize the constant vector. To
deserialize, we need to reconstruct the constant vector, which requires a
memory pool. Hence, we need a way to pass memory pool to
ISerializable::deserialize.

Part of #4303

PR #13758, opened 2025-06-12
Pass context to ISerializable::deserialize from the parent deserialize

MnO2:
In this [PR](https://github.com/facebookincubator/velox/pull/4307) the support to pass context to ISerializable::deserialize is added. However, For the Expr that could contain a `ConstantTypeExpr` we need to pass the context to the nested call to ISerializable::deserialize, so that VectorSaver has a context to use.

Rollback Plan:

Differential Revision: D76525497

### DIFF
--- a/velox/core/PlanNode.cpp
+++ b/velox/core/PlanNode.cpp
@@ -439,12 +439,13 @@ folly::dynamic AggregationNode::Aggregate::serialize() const {
 AggregationNode::Aggregate AggregationNode::Aggregate::deserialize(
     const folly::dynamic& obj,
     void* context) {
-  auto call = ISerializable::deserialize<CallTypedExpr>(obj["call"]);
+  auto call = ISerializable::deserialize<CallTypedExpr>(obj["call"], context);
   auto rawInputTypes =
       ISerializable::deserialize<std::vector<Type>>(obj["rawInputTypes"]);
   FieldAccessTypedExprPtr mask;
   if (obj.count("mask")) {
-    mask = ISerializable::deserialize<FieldAccessTypedExpr>(obj["mask"]);
+    mask =
+        ISerializable::deserialize<FieldAccessTypedExpr>(obj["mask"], context);
   }
   auto sortingKeys = deserializeFields(obj["sortingKeys"], context);
   auto sortingOrders = deserializeSortingOrders(obj["sortingOrders"]);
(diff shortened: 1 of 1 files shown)


velox/core/tests/PlanNodeSerdeTest.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/core/PlanNode.h"
+#include "velox/vector/fuzzer/VectorFuzzer.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
+
+#include <gtest/gtest.h>
+
+namespace facebook::velox::core::test {
+
+class PlanNodeSerDeTest : public testing::Test,
```
mbasmanova: @MnO2 There is an existing velox/exec/tests/PlanNodeSerdeTest.cpp . Would you extend that?

velox/exec/tests/PlanNodeSerdeTest.cpp
```diff
@@ -84,6 +85,12 @@ class PlanNodeSerdeTest : public testing::Test,
   }
 
   std::vector<RowVectorPtr> data_;
+
+  const std::shared_ptr<const core::ValuesNode> source =
```
mbasmanova: @MnO2 This file is using PlanBuilder. Can you use that as well? Is there a particular reason you are using something else?

velox/exec/tests/PlanNodeSerdeTest.cpp
```diff
@@ -622,6 +677,23 @@ TEST_F(PlanNodeSerdeTest, tableWriteMerge) {
...
+              .values({data_})
+              .partialAggregation({"c0"}, {"count(ARRAY[1,2])"})
+              .finalAggregation()
+              .planNode());
+
+  plan = PlanBuilder(pool_.get())
+             .values(data_)
+             .tableWrite("targetDirectory")
+             .localPartition(std::vector<std::string>{})
+             .tableWriteMerge(std::const_pointer_cast<core::AggregationNode>(
```
mbasmanova: Unrelated to this PR, but it is strange that we need std::const_pointer_cast here. Is there a bug in the 'tableWriteMerge' API? CC: @xiaoxmeng

mbasmanova: Here is a fix: https://github.com/facebookincubator/velox/pull/13847

PR #13762, opened 2025-06-13
feat: Optimize nested loop other join types with small build side

zml1206:
follow #10726, optimize nested loop other join types with small build side, include inner, left, right, full outer and left semi join. 
Local test performance improved several times, in the following example, it takes 10 seconds before PR and 2 seconds after PR.
```
TEST_F(NestedLoopJoinTest, innerJoin) {
  auto probeVectors = {
    makeRowVector({sequence<int32_t>(1000000, 10)}),
    };

  auto buildVectors = {
    makeRowVector({sequence<int32_t>(10, 111)}),
    };

  createDuckDbTable("t", {probeVectors});
  createDuckDbTable("u", {buildVectors});

  auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
  auto testOuterJoin = [&](core::JoinType joinType) {
    auto op = PlanBuilder(planNodeIdGenerator)
    .values({probeVectors})
    .nestedLoopJoin(
    PlanBuilder(planNodeIdGenerator)
    .values({buildVectors})
    .project({"c0 AS u_c0"})
    .planNode(),
    "c0 > u_c0",
    {"c0", "u_c0"},
    joinType)
    .singleAggregation({}, {"count(*)"})
    .planNode();
    assertQuery(
    op,
    fmt::format(
        "SELECT count(*) FROM t {} join u on t.c0>u.c0",
        core::joinTypeName(joinType)));
  };
  time_t start_time = time(NULL);
  testOuterJoin(core::JoinType::kInner);
  time_t end_time = time(NULL);
  double time_difference = difftime(end_time, start_time);
  std::cout << "Time difference: " << time_difference << " seconds" << std::endl;
}
```

### DIFF
--- a/velox/exec/NestedLoopJoinProbe.cpp
+++ b/velox/exec/NestedLoopJoinProbe.cpp
@@ -328,7 +328,102 @@ void NestedLoopJoinProbe::handleLeftSemiProjectNoCondition() {
       ->asFlatVector<bool>()
       ->set(numOutputRows_ - 1, true);
   buildIndex_ = buildVectors_.value().size();
+  filterResultRow_ = 0;
+}
+
+bool NestedLoopJoinProbe::addSingleProbeRowOutput(
+    const RowVectorPtr& buildVector) {
+  for (size_t i = filterResultRow_; i < decodedFilterResult_.size(); ++i) {
+    buildRow_ = i;
+
+    if (!isJoinConditionMatch(i)) {
+      if (buildRow_ + 1 != buildVector->size() || !checkProbeMismatchRow()) {
+        continue;
+      }
+    } else {
+      if (needsBuildMismatch(joinType_)) {
+        buildMatched_[buildIndex_].setValid(buildRow_, true);
+      }
+      addOutputRow();
+      ++numOutputRows_;
+      probeRowHasMatch_ = true;
+
+      if (isLeftSemiProjectJoin(joinType_)) {
+        output_->childAt(outputType_->size() - 1)
+            ->asFlatVector<bool>()
+            ->set(numOutputRows_ - 1, true);
+        buildIndex_ = buildVectors_.value().size();
+        filterResultRow_ = 0;
+        buildRow_ = 0;
+        return true;
+      }
+    }
+
+    if (numOutputRows_ == outputBatchSize_) {
+      filterResultRow_ = i + 1;
+      copyBuildValues(buildVector);
+      return false;
+    }
+  }
+
+  copyBuildValues(buildVector);
+  ++buildIndex_;
+  filterResultRow_ = 0;
   buildRow_ = 0;
+  return true;
+}
+
+bool NestedLoopJoinProbe::addMultiProbeRowsOutput(
+    const RowVectorPtr& buildVector) {
+  const auto buildRowCount = buildVector->size();
+  probeRowHasMatch_ = buildRow_ == 0 ? false : probeRowHasMatch_;
+
+  for (size_t i = filterResultRow_; i < decodedFilterResult_.size(); ++i) {
+    if (!isJoinConditionMatch(i)) {
+      if (buildRow_ + 1 == buildRowCount) {
+        checkProbeMismatchRow();
+      }
+    } else {
+      if (needsBuildMismatch(joinType_)) {
+        buildMatched_[buildIndex_].setValid(buildRow_, true);
+      }
+      addOutputRow();
+      ++numOutputRows_;
+      probeRowHasMatch_ = true;
+
+      if (isLeftSemiProjectJoin(joinType_)) {
+        output_->childAt(outputType_->size() - 1)
+            ->asFlatVector<bool>()
+            ->set(numOutputRows_ - 1, true);
+        filterResultRow_ = i + (buildRowCount - buildRow_);
+        ++probeRow_;
+        buildRow_ = 0;
+        probeRowHasMatch_ = false;
+        return true;
+      }
+    }
+
+    ++buildRow_;
+    if (buildRow_ == buildRowCount) {
+      buildRow_ = 0;
+      ++probeRow_;
+      probeRowHasMatch_ = false;
+    }
+
+    if (numOutputRows_ == outputBatchSize_) {
+      filterResultRow_ = i + 1;
+      copyBuildValues(buildVector);
+      return false;
+    }
+  }
+
+  copyBuildValues(buildVector);
+  ++buildIndex_;
+  filterResultRow_ = 0;
+  probeRow_ = filterProbeRow_;
+  probeRowHasMatch_ = false;
+  buildRow_ = 0;
+  return true;
 }
 
 // Main join loop.
@@ -351,7 +446,7 @@ bool NestedLoopJoinProbe::addToOutput() {
     // Empty build vector; move to the next.
     if (currentBuild->size() == 0) {
       ++buildIndex_;
-      buildRow_ = 0;
+      filterResultRow_ = 0;
       continue;
     }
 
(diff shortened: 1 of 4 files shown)


zml1206:
@Yuhta @mbasmanova Can you help take a look? Thank you. Also cc @lifulong

mbasmanova (commented):
@zml1206 Is this an extension to #10726.  If so, do comments in NestedLoopJoinProbe.h need updating? 

Is there documentation about this optimization? I can't find it in https://facebookincubator.github.io/velox/develop/joins.html Would you add it? Otherwise, it is a bit hard to understand what's going on.

Is there a benchmark that can be used to verify the optimization? If not, would you add?

velox/exec/NestedLoopJoinProbe.cpp
```diff
@@ -373,7 +373,9 @@ bool NestedLoopJoinProbe::addToOutput() {
         continue;
       }
 
-      addOutputRow(i);
+      vector_size_t probeRow = i / currentBuild->size() + probeRow_;
```
Yuhta: Can you maintain `probeRow` and `buildRow` and reset them to 0 when reaching `currentBuild->size()`, to avoid putting division inside per-row loop?

velox/exec/NestedLoopJoinProbe.cpp
```diff
@@ -502,13 +504,11 @@ RowVectorPtr NestedLoopJoinProbe::getNextCrossProductBatch(
     const std::vector<IdentityProjection>& buildProjections) {
   VELOX_CHECK_GT(buildVector->size(), 0);
 
-  // TODO: For now we only enable the build optimizations in cross-joins, but we
-  // should allow it for other join types as well.
-  if (isCrossJoin() && isSingleBuildRow()) {
+  if (supportSingleBuild() && isSingleBuildRow()) {
```
Yuhta: Can you make sure the following case is covered in unit test?  If yes paste the unit test names here, if not add the missing ones:
1. Single row build side inner join with filter
2. Single batch build side inner join with filter

zml1206: Single batch already exists in `NestedLoopJoinTest.outputOrder`, added single row.

zml1206:
> @zml1206 Is this an extension to #10726. If so, do comments in NestedLoopJoinProbe.h need updating?
> 
> Is there documentation about this optimization? I can't find it in https://facebookincubator.github.io/velox/develop/joins.html Would you add it? Otherwise, it is a bit hard to understand what's going on.
> 
> Is there a benchmark that can be used to verify the optimization? If not, would you add?

Yes, it is follow up #10726, the main logic has not changed, and there is no need to update the NestedLoopJoinProbe.h document.

velox/exec/NestedLoopJoinProbe.cpp
zml1206: Done,  also add `filterResultRow_` instead of `buildRow_` to  indicate the matching progress of `decodedFilterResult_`. Can you help take a look again? Thanks.

zml1206:
I extended this PR to support all join types of nested loop join.

zml1206:
Join Fuzzer occasionally fails, not sure if it is related to this PR, do you know the reason? @Yuhta

ccat3z:
> Join Fuzzer occasionally fails, not sure if it is related to this PR, do you know the reason? @Yuhta

I found the mismatch case. `buildCopyRanges_` incorrectly accumulate after some mismatch rows. Here is the ut.

```
TEST_F(NestedLoopJoinTest, addOutputRowWithContinuesBuildRow) {
  auto probeVector = makeRowVector(
      {"l1", "l2"},
      {
          makeNullableFlatVector<int64_t>({1, 8, 0}),
          makeFlatVector<StringView>({"a", "b", "c"}),
      });
  auto buildVector = makeRowVector(
      {"r1", "r2"},
      {
          makeNullableFlatVector<int64_t>({1, 1, 0}),
          makeFlatVector<StringView>({"z", "x", "y"}),
      });

  createDuckDbTable("t", {probeVector});
  createDuckDbTable("u", {buildVector});

  auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
  auto op = PlanBuilder(planNodeIdGenerator)
                .values({probeVector})
                .nestedLoopJoin(
                    PlanBuilder(planNodeIdGenerator)
                        .values({buildVector})
                        .planNode(),
                    "l1 = r1",
                    {"l1", "l2", "r1", "r2"},
                    core::JoinType::kLeft)
                .planNode();

  assertQuery(
      op, "SELECT l1, l2, r1, r2 FROM t LEFT JOIN u ON l1 = r1");
}
```

Fix patch: https://github.com/ccat3z/velox/commit/a46ff497c686b6a666123bc28513f97f8883c4aa

rui-mo (commented):
Thanks! Looks good overall. It might be helpful to document the optimization for small build-side inputs, similar to what’s described here: https://facebookincubator.github.io/velox/develop/joins.html#empty-build-side.

zml1206:
> It might be helpful to document the optimization for small build-side inputs, similar to what’s described here: https://facebookincubator.github.io/velox/develop/joins.html#empty-build-side.

@rui-mo Added, thanks.

rui-mo (approved):

zml1206:
@mbasmanova @Yuhta Could you please take another look? Thanks.

mbasmanova:
Initial review — haven't looked at every detail yet, but a few items to address first:

1. **The `addToOutput` loop is hard to follow** — It now tracks three separate indices (`filterResultRow_`, `buildRow_`, `probeRow_`) with complex interactions between them. The `probeRow_ -= probeRowCount_ - 1` at the end of the build vector loop is particularly subtle. Adding comments alone won't fix this — please consider restructuring to make the logic clearer. For example, could the single-probe-row and multi-probe-row paths be separated more explicitly?

2. **Test coverage for all join types** — The tests cover inner and left joins with single row/single batch fast paths. But the PR description says it supports right, full outer, and left semi joins as well. Please add explicit tests for right and full outer joins with single-row and single-vector build sides. Don't rely on the fuzzer — the fuzzer already missed the `buildCopyRanges_` accumulation bug found by @ccat3z.

3. **`isBuildSideEmpty` check** — The mismatch handling changed from unconditionally calling `checkProbeMismatchRow()` at the end of `addToOutput` to only calling it when `isBuildSideEmpty()`. What about the case where build side is not empty but no build rows matched the current probe row? Is this handled correctly by the inline `checkProbeMismatchRow()` call inside the loop?

4. **Tests need restructuring** — The `outputOrder` test uses numbered names throughout (`buildVector1`, `buildVector2`, `buildVector3`, `createPlan1`, `createPlan2`) and mixes multiple scenarios in one long method. Please break this into focused test cases with descriptive names.

5. **Use `assertResults` instead of `assertEqualVectors`** — The `outputOrder` test manually constructs expected vectors, which is fragile — if the expected output is wrong, the test passes with incorrect results. Use `AssertQueryBuilder(..., duckDbQueryRunner_).assertResults("SELECT ...")` to verify against DuckDB, like `addOutputRowWithContinuesBuildRow` already does.

6. **Commit history** — Please squash into a clean commit.

zml1206:
> Initial review — haven't looked at every detail yet, but a few items to address first:
> 
> 1. **The `addToOutput` loop is hard to follow** — It now tracks three separate indices (`filterResultRow_`, `buildRow_`, `probeRow_`) with complex interactions between them. The `probeRow_ -= probeRowCount_ - 1` at the end of the build vector loop is particularly subtle. Adding comments alone won't fix this — please consider restructuring to make the logic clearer. For example, could the single-probe-row and multi-probe-row paths be separated more explicitly?
> 2. **Test coverage for all join types** — The tests cover inner and left joins with single row/single batch fast paths. But the PR description says it supports right, full outer, and left semi joins as well. Please add explicit tests for right and full outer joins with single-row and single-vector build sides. Don't rely on the fuzzer — the fuzzer already missed the `buildCopyRanges_` accumulation bug found by @ccat3z.
> 3. **`isBuildSideEmpty` check** — The mismatch handling changed from unconditionally calling `checkProbeMismatchRow()` at the end of `addToOutput` to only calling it when `isBuildSideEmpty()`. What about the case where build side is not empty but no build rows matched the current probe row? Is this handled correctly by the inline `checkProbeMismatchRow()` call inside the loop?
> 4. **Tests need restructuring** — The `outputOrder` test uses numbered names throughout (`buildVector1`, `buildVector2`, `buildVector3`, `createPlan1`, `createPlan2`) and mixes multiple scenarios in one long method. Please break this into focused test cases with descriptive names.
> 5. **Use `assertResults` instead of `assertEqualVectors`** — The `outputOrder` test manually constructs expected vectors, which is fragile — if the expected output is wrong, the test passes with incorrect results. Use `AssertQueryBuilder(..., duckDbQueryRunner_).assertResults("SELECT ...")` to verify against DuckDB, like `addOutputRowWithContinuesBuildRow` already does.
> 6. **Commit history** — Please squash into a clean commit.

Thanks, updated.

velox/exec/NestedLoopJoinProbe.cpp
```diff
@@ -328,7 +328,102 @@ void NestedLoopJoinProbe::handleLeftSemiProjectNoCondition() {
...
+
+bool NestedLoopJoinProbe::addMultiProbeRowsOutput(
+    const RowVectorPtr& buildVector) {
+  const auto buildRowCount = buildVector->size();
+  probeRowHasMatch_ = buildRow_ == 0 ? false : probeRowHasMatch_;
+
+  for (size_t i = filterResultRow_; i < decodedFilterResult_.size(); ++i) {
+    if (!isJoinConditionMatch(i)) {
+      if (buildRow_ + 1 == buildRowCount) {
+        checkProbeMismatchRow();
```
zml1206, line 384: Regarding the third point, I have addressed it here. @mbasmanova

mbasmanova (changes requested):
Thank you for addressing the feedback!

**`addSingleProbeRowOutput` and `addMultiProbeRowsOutput` have significant duplication.** The match handling, left semi project logic, build mismatch tracking, output buffer management, and `copyBuildValues` are nearly identical — only the probe row advancement differs. Can these share a common implementation parameterized on whether there's a single probe row?

**Missing test for left semi project with small build side.** Both `addSingleProbeRowOutput` and `addMultiProbeRowsOutput` have left semi project handling, but there's no test exercising this path with single-row or single-vector build sides.

mbasmanova (approved):
@zml1206, Thank you for the updates! Looks good.

PR #13765, opened 2025-06-13
feat: Add utility FilterToExpression

juwentus1234:
Summary: Add utility method FilterToExpression which will be used to convert Expression from corresponding Filter, initial draft focusing on getTypedExprFromSubfield

Differential Revision: D76608202

### DIFF
--- a/velox/core/FilterToExpression.cpp
+++ b/velox/core/FilterToExpression.cpp
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


velox/core/FilterToExpression.cpp
```diff
@@ -0,0 +1,98 @@
...
+        // Create a field access expression for the nested field
+        expr = std::make_shared<core::FieldAccessTypedExpr>(
+            nestedType, expr, nestedName);
+        break;
+      }
+
+      case common::SubfieldKind::kLongSubscript:
+      case common::SubfieldKind::kStringSubscript:
+      case common::SubfieldKind::kAllSubscripts:
+        VELOX_NYI(
```
Yuhta, line 80: This could be `VELOX_UNREACHABLE` since no one push down anything other than struct subfields

Yuhta (approved):

PR #13772, opened 2025-06-14
fix: Fix QuantileDigest malformed tree structure

kagamiori:
Summary:
This diff fixes a bug in QuantileDigest::mergeSerializedRecursive that causes
malformed tree structure where one node has more than one parent.

Differential Revision: D76645144

### DIFF
--- a/velox/functions/lib/QuantileDigest.h
+++ b/velox/functions/lib/QuantileDigest.h
@@ -68,17 +68,17 @@ class QuantileDigest {
 
   /// Estimate the values of the given quantiles and write the results in order
   /// directly into 'result'.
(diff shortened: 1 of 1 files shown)


spershin (approved):
Thank you!

spershin:
@kagamiori 
Should we add a unit test that creates malformed tree before the fix and test it with validate?

kagamiori:
> @kagamiori Should we add a unit test that creates malformed tree before the fix and test it with validate?

@spershin The existing unit test QuantileDigestTest.mergeWithJava already triggers failure during validateDigest without the fix. With the fix, this unit test passes.

PR #13773, opened 2025-06-14
fix: Reduce the table scan metrics monitoring cost in some ai use cases

xiaoxmeng:
Summary: Changes the metrics type from histogram to average

Differential Revision: D76648134

### DIFF
--- a/velox/common/base/Counters.cpp
+++ b/velox/common/base/Counters.cpp
@@ -645,15 +645,12 @@ void registerVeloxMetrics() {
       kMetricIndexLookupBlockedWaitTimeMs, 32, 0, 16L << 10, 50, 90, 99, 100);
 
   /// ================== Table Scan Counters =================
-  // The time distribution of table scan batch processing time in range of [0,
-  // 16s] with 512 buckets and reports P50, P90, P99, and P100.
-  DEFINE_HISTOGRAM_METRIC(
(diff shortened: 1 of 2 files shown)


velox/common/base/Counters.cpp
```diff
@@ -645,15 +645,12 @@ void registerVeloxMetrics() {
       kMetricIndexLookupBlockedWaitTimeMs, 32, 0, 16L << 10, 50, 90, 99, 100);
 
   /// ================== Table Scan Counters =================
-  // The time distribution of table scan batch processing time in range of [0,
-  // 16s] with 512 buckets and reports P50, P90, P99, and P100.
-  DEFINE_HISTOGRAM_METRIC(
-      kMetricTableScanBatchProcessTimeMs, 32, 0, 16L << 10, 50, 90, 99, 100);
+  // Tracks the averaged table scan batch processing time in milliseconds.
+  DEFINE_METRIC(
```
czentgr, line 649: In the table scan itself it still collects a histogram metric:
```
          RECORD_HISTOGRAM_METRIC_VALUE(
              velox::kMetricTableScanBatchBytes, data->estimateFlatSize());
```

We've seen this cause SEGVs with Prometheus.
I've created a PR https://github.com/facebookincubator/velox/pull/14210 to fix this. The SEGV does not occur anymore in our local tests.

PR #13771, opened 2025-06-13
feat: Add a method in GeometryDeserializer to efficiently get envelope from a geometry

wraymo:
Summary
In Presto's internal geometry representation, the envelope of a geometry is stored alongside most geometry types. As noted in #13674, it is unnecessary to fully deserialize a geometry to compute its envelope — the envelope can be read directly. This PR introduces a method in `GeometryDeserializer` that efficiently retrieves the envelope without requiring full deserialization.

PR #13778, opened 2025-06-15
feat: Add envelope-based short-circuit optimizations for geometry functions

wraymo:
This PR introduces envelope-based short-circuit optimizations to various geometry functions.

By checking the envelopes of input geometries before performing full deserialization and expensive GEOS operations, the functions can now avoid unnecessary computation when the result is trivially determined by the envelopes alone.

**Note:** This PR depends on #13771 and can be landed after it is merged.

### DIFF
--- a/velox/functions/prestosql/GeometryFunctions.h
+++ b/velox/functions/prestosql/GeometryFunctions.h
@@ -618,7 +618,7 @@ struct StXMinFunction {
 
   bool call(out_type<double>& result, const arg_type<Geometry>& geometry) {
     const std::unique_ptr<geos::geom::Envelope> env =
-        geospatial::getEnvelopeFromGeometry(geometry);
+        geospatial::GeometryDeserializer::deserializeEnvelope(geometry);
     if (env->isNull()) {
       return false;
     }
@@ -633,7 +633,7 @@ struct StYMinFunction {
 
   bool call(out_type<double>& result, const arg_type<Geometry>& geometry) {
     const std::unique_ptr<geos::geom::Envelope> env =
-        geospatial::getEnvelopeFromGeometry(geometry);
+        geospatial::GeometryDeserializer::deserializeEnvelope(geometry);
     if (env->isNull()) {
       return false;
     }
@@ -648,7 +648,7 @@ struct StXMaxFunction {
 
   bool call(out_type<double>& result, const arg_type<Geometry>& geometry) {
     const std::unique_ptr<geos::geom::Envelope> env =
-        geospatial::getEnvelopeFromGeometry(geometry);
+        geospatial::GeometryDeserializer::deserializeEnvelope(geometry);
     if (env->isNull()) {
       return false;
     }
@@ -663,7 +663,7 @@ struct StYMaxFunction {
 
   bool call(out_type<double>& result, const arg_type<Geometry>& geometry) {
     const std::unique_ptr<geos::geom::Envelope> env =
-        geospatial::getEnvelopeFromGeometry(geometry);
+        geospatial::GeometryDeserializer::deserializeEnvelope(geometry);
     if (env->isNull()) {
       return false;
     }
--- a/velox/functions/prestosql/GeometryFunctions.h
+++ b/velox/functions/prestosql/GeometryFunctions.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <geos/geom/Coordinate.h>
+#include <geos/geom/CoordinateArraySequence.h>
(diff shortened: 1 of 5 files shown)


velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -403,8 +459,23 @@ struct StSymDifferenceFunction {
       out_type<Geometry>& result,
       const arg_type<Geometry>& leftGeometry,
       const arg_type<Geometry>& rightGeometry) {
-    // TODO: When #12771 is merged, check envelopes and short-circuit
-    // if envelopes are disjoint
+    // If envelopes are disjoint, return union of envelopes directly.
```
jagill, line 462: I don't think we gain anything from this block.  We still have to deserialize things, and SymDifference almost certainly notices disjoint cases and handles them well.  Please remove this block.

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -378,8 +424,18 @@ struct StIntersectionFunction {
       out_type<Geometry>& result,
       const arg_type<Geometry>& leftGeometry,
       const arg_type<Geometry>& rightGeometry) {
-    // TODO: When #12771 is merged, check envelopes and short-circuit
-    // if envelopes are disjoint
+    auto leftType =
```
jagill, line 427: The types seem unused.  Are they left over from something?  Should we remove them?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -378,8 +424,18 @@ struct StIntersectionFunction {
...
+        geospatial::GeometryDeserializer::deserializeType(leftGeometry);
+    auto rightType =
+        geospatial::GeometryDeserializer::deserializeType(rightGeometry);
+
+    if (!geospatial::checkEnvelopePredicate(
+            leftGeometry, rightGeometry, geospatial::EnvelopeOp::Intersects)) {
+      geospatial::GeometrySerializer::serialize(
+          *geospatial::getGeometryFactory()->createPolygon(), result);
```
jagill, line 435: Current Presto implementation returns a MULTIPOLYGON EMPTY.  Does this return that or POLYGON EMPTY?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -428,8 +499,30 @@ struct StUnionFunction {
...
       const arg_type<Geometry>& rightGeometry) {
-    // TODO: When #12771 is merged, check envelopes and short-circuit if
-    // one/both are empty
+    auto leftEnvelope =
+        geospatial::GeometryDeserializer::deserializeEnvelope(leftGeometry);
+    auto rightEnvelope =
+        geospatial::GeometryDeserializer::deserializeEnvelope(rightGeometry);
+
+    // If both are empty, return empty polygon.
+    if (!leftEnvelope || leftEnvelope->isNull()) {
```
jagill, line 508: Can we separate this into three disjoint blocks?  It'd be easier to read.  Ie,

```
bool leftEmpty = !leftEnvelope || leftEnvelope->isNull();
bool rightEmpty = !rightEnvelope || rightEnvelope->isNull();

if (leftEmpty && rightEmpty) { ... }
if (leftEmpty) { ... }
if (rightEmpty) { ... }
```

velox/functions/prestosql/geospatial/GeometrySerde.h
```diff
@@ -114,24 +114,44 @@ class GeometrySerializer {
     }
   }
 
+  template <typename StringWriter>
```
jagill, line 117: Maybe include these in the early PR?  They seem related.

velox/functions/prestosql/geospatial/GeometryUtils.cpp
```diff
@@ -29,6 +30,41 @@ geos::geom::GeometryFactory* getGeometryFactory() {
...
 }
 
+bool checkEnvelopePredicate(
+    const StringView& leftGeometry,
+    const StringView& rightGeometry,
+    EnvelopeOp op) {
+  auto leftEnvelope = GeometryDeserializer::deserializeEnvelope(leftGeometry);
+  auto rightEnvelope = GeometryDeserializer::deserializeEnvelope(rightGeometry);
+
+  if (!leftEnvelope || !rightEnvelope) {
```
jagill, line 40: This is a good place to ask: What does a null envelope mean?  Should we have a contract that deserializeEnvelope never returns a nullptr, but instead would return an empty envelope?

What are the semantics we want here?

PR #13785, opened 2025-06-16
fix(noisy_count): change noise overflow check

oliver6782:
Summary:
**Noisy Count Gaussian Aggregate Fix**

This diff addresses an issue with the noisy count overflow check in the `NoisyCountGaussianAggregate` class. The change ensures that the added noise does not exceed the range of `int64_t` to prevent overflow.

Specifically, the updated code:

* Generates a raw noise value using a Gaussian distribution
* Checks that the raw noise is within the valid range for `int64_t` using `VELOX_USER_CHECK_GE` and `VELOX_USER_CHECK_LE`
* If the noise is too large, raises an error with a message to reduce the noise scale

These changes improve the robustness and reliability of the `NoisyCountGaussianAggregate` function.

**Changed Files**

* `fbcode/velox/functions/prestosql/aggregates/NoisyCountGaussianAggregate.cpp`

Differential Revision: D76699264


amitkdutta (approved):
Thanks @oliver6782

PR #13806, opened 2025-06-17
feat: Support to configure split index lookup output

xiaoxmeng:
Summary: Add query config to avoid splitting index lookup output to optimize the followup operator processing like streaming aggregation which might be able to receive the entire lookup results for a given input or batch.

Differential Revision: D76243219

### DIFF
--- a/velox/exec/IndexLookupJoin.cpp
+++ b/velox/exec/IndexLookupJoin.cpp
@@ -140,7 +140,10 @@ IndexLookupJoin::IndexLookupJoin(
           "IndexLookupJoin"),
       // TODO: support to update output batch size with output size stats during
       // the lookup processing.
-      outputBatchSize_{outputBatchRows()},
+      outputBatchSize_{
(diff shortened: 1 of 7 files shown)


velox/exec/IndexLookupJoin.cpp
```diff
@@ -636,13 +638,20 @@ RowVectorPtr IndexLookupJoin::produceOutputForLeftJoin(
...
+      batch.lookupResult->size() - nextOutputResultRow_ + batch.input->size() -
+          startProcessInputRow);
+  prepareOutputRowMappings(maxOutputRows);
   VELOX_CHECK_NOT_NULL(rawLookupOutputNulls_);
-
   size_t numOutputRows{0};
   size_t totalMissedInputRows{0};
-  int32_t lastProcessedInputRow = lastProcessedInputRow_.value_or(-1);
-  for (; numOutputRows < outputBatchSize_ &&
+  for (; numOutputRows < maxOutputRows &&
```
zacw7, line 654: curious why don't we just write "while" here?

PR #13809, opened 2025-06-17
fix(build): Fix location of EXTRA_ARROW_OPTIONS usage

czentgr:
The scripts refactor moved the variable from install_arrow to install_thrift which is the incorrect dependency.


czentgr:
@kevinwilfong Can we help with the internal facebook fails? Seems odd that this breaks something on that side?

PR #13812, opened 2025-06-17
misc: Optimize streaming agg performance

xiaoxmeng:
Summary: Adds a config in streaming aggregation to try to split the group output at the input batch boundary if possible. With this config on, the streaming aggregation operator tries to produce the groups created from the previous input batch. This gives the array aggregation to avoid copy data when does array aggregation. The array aggregation avoids copy data if the array element comes from a single source input.

Differential Revision: D76790683

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -623,6 +623,13 @@ class QueryConfig {
   static constexpr const char* kStreamingAggregationMinOutputBatchRows =
       "streaming_aggregation_min_output_batch_rows";
 
+  /// If true, the streaming aggregation accumulates at least two input batches
+  /// and produce all the groups created from the previous input batch except
+  /// the last group.
+  static constexpr const char*
+      kStreamingAggregationTrySplitOutputAtInputBoundary =
+          "streaming_aggregation_try_split_output_at_input_boundary";
+
   /// TODO: Remove after dependencies are cleaned up.
   static constexpr const char* kStreamingAggregationEagerFlush =
       "streaming_aggregation_eager_flush";
@@ -1163,6 +1170,10 @@ class QueryConfig {
 
   int32_t streamingAggregationMinOutputBatchRows() const {
     return get<int32_t>(kStreamingAggregationMinOutputBatchRows, 0);
+  }
+
+  bool streamingAggregationTrySplitOutputAtInputBoundary() const {
+    return get<bool>(kStreamingAggregationTrySplitOutputAtInputBoundary, false);
   }
 
   bool isFieldNamesInJsonCastEnabled() const {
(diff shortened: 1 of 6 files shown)


velox/core/QueryConfig.h
```diff
@@ -623,6 +623,13 @@ class QueryConfig {
   static constexpr const char* kStreamingAggregationMinOutputBatchRows =
       "streaming_aggregation_min_output_batch_rows";
 
+  /// If true, the streaming aggregation accumulates at least two input batches
+  /// and produce all the groups created from the previous input batch except
+  /// the last group.
+  static constexpr const char*
+      kStreamingAggregationTrySplitOutputAtInputBoundary =
+          "streaming_aggregation_try_split_output_at_input_boundary";
```
mbasmanova, line 631: In general, it is error prone to introduce configs that produce incorrect results if set incorrectly. Configs should only affect performance, but not correctness.

CC: @amitkdutta @xiaoxmeng @oerling @zacw7

velox/exec/StreamingAggregation.cpp
```diff
@@ -341,15 +345,20 @@ RowVectorPtr StreamingAggregation::getOutput() {
...
-        std::min(numGroups_ - 1, static_cast<size_t>(maxOutputBatchSize_)));
+  if (trySplitOutputAtInputBoundary_) {
+    if ((numPrevGroups != 0) && (numGroups_ > minOutputBatchSize_)) {
+      output = createOutput(std::min(numGroups_ - 1, numPrevGroups));
+    }
+  } else {
+    if (numGroups_ > minOutputBatchSize_) {
+      output = createOutput(
+          std::min(numGroups_ - 1, static_cast<size_t>(maxOutputBatchSize_)));
+    }
```
tanjialiang, line 361: I think we can extract this section out as a separate batching method that deals with different batching strategies based on settings.

velox/exec/StreamingAggregation.cpp
```diff
@@ -341,15 +345,20 @@ RowVectorPtr StreamingAggregation::getOutput() {
...
   initializeNewGroups(numPrevGroups);
   evaluateAggregates();
 
   RowVectorPtr output;
-  if (numGroups_ > minOutputBatchSize_) {
-    output = createOutput(
-        std::min(numGroups_ - 1, static_cast<size_t>(maxOutputBatchSize_)));
+  if (trySplitOutputAtInputBoundary_) {
+    if ((numPrevGroups != 0) && (numGroups_ > minOutputBatchSize_)) {
+      output = createOutput(std::min(numGroups_ - 1, numPrevGroups));
```
tanjialiang, line 355: Could there be any case that the last group of prev groups is partially filled. This getOutput call uses the current input_ to fill out that group, and a few new groups. Then the current result batch is not strictly cut by input boundary?

velox/core/QueryConfig.h
xiaoxmeng: @mbasmanova there is no correctness issue here as we still keep the last group in the streaming aggregation operator.

PR #13821, opened 2025-06-18
fix: Mark qdigest_agg accumulator as aligned and not fixed size

kagamiori:
Differential Revision: D76907097

### DIFF
--- a/velox/functions/prestosql/aggregates/QDigestAggAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/QDigestAggAggregate.cpp
@@ -32,6 +32,9 @@ namespace {
 
 template <typename T, typename Func, bool DefaultMaxError>
 struct QDigestAccumulator {
+  static constexpr bool is_fixed_size_ = false;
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/aggregates/QDigestAggAggregate.cpp
```diff
@@ -32,6 +32,9 @@ namespace {
 
 template <typename T, typename Func, bool DefaultMaxError>
 struct QDigestAccumulator {
+  static constexpr bool is_fixed_size_ = false;
+  static constexpr bool is_aligned_ = true;
```
Yuhta, line 36: We should remove this flag and align all the simple aggregate adapters

Yuhta (approved):

velox/functions/prestosql/aggregates/QDigestAggAggregate.cpp
kagamiori: @Yuhta Good suggestion. I can make that change in a separate diff later.

PR #13824, opened 2025-06-18
feat: Add Textfile read range, projection and filter feature

YiChengLee03:
Summary:
- Allow users to set range of bytes to read (row aligned)
- Support skip rows (i.e. headers)
- Support projection (column selection)
- Support filter

Differential Revision: D76486983

### DIFF
--- a/velox/dwio/text/TextReader.cpp
+++ b/velox/dwio/text/TextReader.cpp
@@ -82,9 +82,9 @@ const std::shared_ptr<const TypeWithId>& TextReader::typeWithId() const {
 
 std::unique_ptr<RowReader> TextReader::createRowReader(
     const RowReaderOptions& options) const {
-  auto rowReader = reader_->createRowReader(options);
+  auto rowReaderImpl = reader_->createRowReader(options);
   return std::make_unique<TextRowReader>(
-      std::move(rowReader), options_.memoryPool(), options.scanSpec());
+      std::move(rowReaderImpl), options_.memoryPool(), options.scanSpec());
 }
 
 std::unique_ptr<Reader> TextReaderFactory::createReader(
--- a/velox/dwio/text/reader/TextReaderImpl.cpp
+++ b/velox/dwio/text/reader/TextReaderImpl.cpp
@@ -139,8 +139,8 @@ uint64_t TextRowReaderImpl::next(
       }
 
       DelimType delim = DelimTypeNone;
-      auto const& ct = t->childAt(i);
-      auto const& rct = reqT->childAt(i);
+      const auto& ct = t->childAt(i);
+      const auto& rct = reqT->childAt(i);
 
       /// TODO: PROJECTION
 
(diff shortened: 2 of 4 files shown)


velox/dwio/text/TextReader.cpp
```diff
@@ -82,9 +82,9 @@ const std::shared_ptr<const TypeWithId>& TextReader::typeWithId() const {
 
 std::unique_ptr<RowReader> TextReader::createRowReader(
     const RowReaderOptions& options) const {
-  auto rowReader = reader_->createRowReader(options);
+  auto rowReaderImpl = reader_->createRowReader(options);
```
zacw7: Let's keep it as "rowReader"

velox/dwio/text/reader/TextReaderImpl.cpp
```diff
@@ -64,29 +71,43 @@ TextRowReaderImpl::TextRowReaderImpl(
...
-  fileLength_ = getStreamLength();
   // seek to first line at or after the specified region.
-  auto offset = opts.offset();
   if (contents_->compression == CompressionKind::CompressionKind_NONE) {
-    if (offset != 0) {
-      pos_ = offset;
+    /**
+    streamPosition_ = pos_ == 0 ? 0 : --pos_ to account for '\n' at end of
+    line, but kept as it is to be consistent with presto reader
+    */
```
zacw7: Could you elaborate a bit? Not sure I fully understand this comment.

velox/dwio/text/reader/TextReaderImpl.cpp
```diff
@@ -64,29 +71,43 @@ TextRowReaderImpl::TextRowReaderImpl(
...
       atEOL_{false},
       atEOF_{false},
       atSOL_{false},
       depth_{0},
+      limit_{opts.limit()},
+      fileLength_{getStreamLength()},
       stringViewBuffer_{StringViewBufferHolder(&contents_->pool)} {
-  limit_ = options_.limit();
-  fileLength_ = getStreamLength();
   // seek to first line at or after the specified region.
```
zacw7: nit: if the comment is a proper sentence, let's make sure it has right format - ends with `.` with first letter capitalized.

velox/dwio/text/reader/TextReaderImpl.cpp
```diff
@@ -1000,22 +1022,22 @@ void TextRowReaderImpl::readElement(
...
       if ((atEOF_ && atSOL_) || (data == nullptr)) {
         return;
       }
 
-      auto flatVector = data->asChecked<FlatVector<Timestamp>>();
+      auto flatVector =
+          data ? data->asChecked<FlatVector<Timestamp>>() : nullptr;
       if (!flatVector) {
-        VELOX_FAIL("Expected FlatVector but got {}", typeid(*data).name());
+        VELOX_FAIL("Vector for column type does not match");
```
zacw7: Can we somehow keep the mismatched type name in the exception message so we have more visibility into what goes wrong?

velox/dwio/text/reader/TextReaderImpl.cpp
YiChengLee03: Sure!

Issue: When reading from byte offset > 0, we skip rows inclusively at the start position, but when reading from byte 0, no rows are skipped. This creates inconsistent behavior where a row at the boundary may be skipped when it should be included.

Example: If pos_ = 10 is the first byte of row 2, that entire row gets skipped, even though it should be read.

Proposed fix: streamPosition_ = (pos_ == 0) ? 0 : --pos_; 
This would skip rows exclusively of pos_, ensuring consistent behavior.

Note: Updated the comments in the code too.

zacw7 (approved):

PR #13827, opened 2025-06-18
feat: Make global arbitration consider query priority

tanjialiang:
Summary: similarly as we consider query priority when pushback by abort, we should also consider priority when global arbitration.

Differential Revision: D76935498


xiaoxmeng (approved):
@tanjialiang LGTM. Thanks!

PR #13829, opened 2025-06-18
feat(vector): Add support for flat map encoding in vector function map_filter

peterenescu:
Summary:
Add support for flat map encoding (FlatMapVector.h) in function map_filter. As part of the effort to support feature reaping in Velox, let's add support for flat map vectors in vector function `map_filter`. This will enable our pipelines to efficiently project and filter features given this new vector encoding.

Differential Revision: D76938846


pedroerp (approved):
Looks good to me. Thank you

PR #13830, opened 2025-06-19
misc: Pass query context to WarmStorage reads

kletkavrubashku:
Summary:
In Java we pass a context with extra "dw.*" fields about presto query to warmstorage. Basically it's just a `map<string, string>`. We need something similar in Prestissimo to improve debugging and monitoring.
https://fburl.com/scuba/warm_storage_file_service_activity/n84zevam
___
This is just a refactoring, I'm passing session properties through the connector stack for reads. The plan is to pass DW specific properties inside.

The most questionable part is probably changes in `CachedFactory.h`.
Currently `CachedFactory::generate` has 3 template params,
- `Key`, used as string for the path to file in FileHandleGenerator
- `Properties`, used for FileProperties struct. Looks like a const file properties according to this [comment](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/HiveConnectorSplit.h?fbclid=IwY2xjawLASuFleHRuA2FlbQIxMQBicmlkETFlR2JVSktON2Y5eFRJT2lqAR4M46I6rtXwOiI0V8DukkBF_jXUaktPB_4qwzckyvYZweWaV3BVlcHWwj9fVA_aem_aVoQh-IF6xUzfJbZK-PPdQ#L67-L69).
- `Statistics`, used for statistics

I needed to pass SessionProperties there, so I added separate parameter `Context` to that class, but LMK if it can be done better.
___
WarmStorage writes and spills are not covered in this diff, those changes will be published in the next diffs.

Differential Revision: D76921538

### DIFF
--- a/velox/connectors/hive/SplitReader.cpp
+++ b/velox/connectors/hive/SplitReader.cpp
@@ -294,7 +294,8 @@ void SplitReader::createReader() {
     fileHandleCachePtr = fileHandleFactory_->generate(
         hiveSplit_->filePath,
         hiveSplit_->properties.has_value() ? &*hiveSplit_->properties : nullptr,
-        fsStats_ ? fsStats_.get() : nullptr);
+        fsStats_ ? fsStats_.get() : nullptr,
+        connectorQueryCtx_->sessionProperties());
     VELOX_CHECK_NOT_NULL(fileHandleCachePtr.get());
   } catch (const VeloxRuntimeError& e) {
     if (e.errorCode() == error_code::kFileNotFound &&
--- a/velox/common/caching/CachedFactory.h
+++ b/velox/common/caching/CachedFactory.h
@@ -147,6 +147,7 @@ template <
     typename Generator,
     typename Properties = void,
     typename Statistics = void,
+    typename Context = void,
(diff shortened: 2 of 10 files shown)


velox/connectors/hive/SplitReader.cpp
```diff
@@ -294,7 +294,8 @@ void SplitReader::createReader() {
     fileHandleCachePtr = fileHandleFactory_->generate(
         hiveSplit_->filePath,
         hiveSplit_->properties.has_value() ? &*hiveSplit_->properties : nullptr,
-        fsStats_ ? fsStats_.get() : nullptr);
+        fsStats_ ? fsStats_.get() : nullptr,
```
jaystarshot, line 297: This is a filehandle cache (fileHandleCachePtr) and this caches by path. Which means that if 2 users query the same path or 2nd query with different session properties only the first users session properties will be shown

wraymo:
The cache is keyed by the file path, so even if session properties change, the cached file handle is still reused (it will not be generated again). And if we plan to add `Context` to `CachedFactory`, why not use `ConnectorQueryContext` directly?

PR #13833, opened 2025-06-19
fix(build): Replace shellcheck pre-commit with python version

czentgr:
Previously, the shellcheck used docker based check that doesn’t work in all environments. Switch to a python based version instead.


assignUser (approved):
Nice improvement, thanks!

Issue #13843, opened 2025-06-23
Fix build failures in TextReader


xin-zhang2:
### Problem description

`TextReaderImpl.cpp` is not built because the `reader` subdirectory is not included in the `CMakeList.txt` under `text` directory.

I got the following build error when adding the `reader` subdirectory in the `CMakeList.txt`.

```
FAILED: velox/dwio/text/reader/CMakeFiles/velox_dwio_text_reader.dir/TextReaderImpl.cpp.o
[...]

PR #13845, opened 2025-06-23
fix(build): Fix build failures in TextReader

xin-zhang2:
Fix https://github.com/facebookincubator/velox/issues/13843

### DIFF
--- a/velox/dwio/text/reader/TextReaderImpl.cpp
+++ b/velox/dwio/text/reader/TextReaderImpl.cpp
@@ -500,7 +500,7 @@ T TextRowReaderImpl::getInteger(
   int64_t v = 0;
   uint64_t scanPos = 0;
   errno = 0;
-  auto scanCount = sscanf(s.data(), "%" SCNd64 "%" SCNu64, &v, &scanPos);
+  auto scanCount = sscanf(s.data(), "%" SCNd64 "%lln", &v, &scanPos);
   if (scanCount != 1 || errno == ERANGE) {
     isNull = true;
     return 0;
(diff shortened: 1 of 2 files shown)


xin-zhang2:
@YiChengLee03 PTAL.

YiChengLee03:
Thanks for checking on build files and system portability!

velox/dwio/text/reader/TextReaderImpl.cpp
```diff
@@ -500,7 +500,7 @@ T TextRowReaderImpl::getInteger(
   int64_t v = 0;
   uint64_t scanPos = 0;
   errno = 0;
-  auto scanCount = sscanf(s.data(), "%ld%ln", &v, &scanPos);
+  auto scanCount = sscanf(s.data(), "%" SCNd64 "%" SCNu64, &v, &scanPos);
```
YiChengLee03: While using "%" SCNd64 "%" SCNu64 is great for system compatibility, it is more restrictive. Using "%ld%ln" allows us to parse 1.123 as an INTEGER into 1, but this change does not allow it to be parsed as expected in Presto. The former behavior is desired.

xin-zhang2: Thanks for the review. This issus was caused by incorrectly replacing `%ln` with `SCNu64`. I've update it with `%lln` to ensure compatibility with most platforms. The `readFloatAsInt` test now passes.

xin-zhang2: Also changed `scanPos` to `unsigned long long`, because `uint64_t` might be defined as either `unsigned long` or `unsigned long long` depending on the platform, and `unsigned long long` is 64 bit in most platforms.

YiChengLee03: Thanks for the quick fix!

zacw7 (approved):

PR #13846, opened 2025-06-23
fix(TextReader): Resize rowVecPtr when EOF is true

xin-zhang2:
The size of rowVectorPtr is initialized with the `rows` argument and should be resized to the actual value when an EOF is reached.


xin-zhang2:
@YiChengLee03 
Can you take a review ?

YiChengLee03:
@xin-zhang2 Thanks for catching this. LGTM

YiChengLee03:
cc: @zacw7

zacw7 (approved):
thanks for the fix

PR #13847, opened 2025-06-23
fix: PlanBuilder::tableWriteMerge API

mbasmanova:
Summary:
PlanBuilder::tableWriteMerge should take `std::shared_ptr<const AggregationNode>`, not `std::shared_ptr<AggregationNode>`. 

See https://github.com/facebookincubator/velox/pull/13758#discussion_r2161523058

Differential Revision: D77148564


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

Issue #13850, opened 2025-06-23
Build fails when set(VELOX_ENABLE_GEO OFF)


yingsu00:
### Problem description

If VELOX_ENABLE_GEO is set to OFF, build would fail with the following errors:
```
Undefined symbols for architecture arm64:
  "facebook::velox::functions::registerGeometryFunctions(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)", referenced from:
      facebook::velox::functions::prestosql::registerGeometryFunctions(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in libvelox_functions_prestosql.a[21](RegistrationFunctions.cpp.o)
ld: symbol(s) not found for architecture arm64
...
Undefined symbols for architecture arm64:
  "geos::io::WKBReader::read(unsigned char const*, unsigned long)", referenced from:
      void facebook::velox::SelectivityVector::applyToSelected<void facebook::velox::exec::EvalCtx::applyToSelectedNoThrow<void....
```
This is because the function registration would still try to register the geo functions in velox/functions/prestosql/registration/RegistrationFunctions.cpp

### System information

Velox System Info v0.0.2
Commit: 14f24d9c487955a1fd4903c1cb94720a8e35ce20
CMake Version: 3.31.0
System: Darwin-23.4.0
Arch: arm64
C++ Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
C++ Compiler Version: 15.0.0.15000309
C Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
C Compiler Version: 15.0.0.15000309
CMake Prefix Path: /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr;/opt/homebrew;/usr/local;/usr;/;/usr/local;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local
  

### CMake log

```bash
yingsu@loaclhost velox % make debug
/Library/Developer/CommandLineTools/usr/bin/make cmake BUILD_DIR=debug BUILD_TYPE=Debug
mkdir -p _build/debug && \
	cmake  -B \
		"_build/debug" \
[...]

PR #13541, opened 2025-05-29
fix: Guard GEOS dependencies with VELOX_ENABLE_GEO

wraymo:
Summary:
Currently, the build fails when `VELOX_ENABLE_GEO` is turned off.

This PR adds conditional compilation guards using the `VELOX_ENABLE_GEO` macro to exclude compiling files that requires GEOS dependency when geo support is disabled. =

PR #13851, opened 2025-06-23
fix: Fix build error when VELOX_ENABLE_GEO is disabled

yingsu00:
Fixes https://github.com/facebookincubator/velox/issues/13850

When VELOX_ENABLE_GEO is set to OFF, build would fail with error "use of undeclared identifier 'registerGeometryFunctions'". This commit fixes this problem by checking VELOX_ENABLE_GEO when registerGeometryFunctions is called.


wraymo:
Not sure if it's related to https://github.com/facebookincubator/velox/pull/13541

yingsu00:
> Not sure if it's related to #13541

@wraymo Thanks for your information. I think we are both trying to solve the same problem, but I chose to disable the whole GEO function registration, and I didn't set VELOX_ENABLE_GEO to false.  @assignUser what do you think?

assignUser:
I prefer that we continue in #13541 as it checks it in CI.

 But removing the sourcefile from the target instead of wrapping it in an #ifdef might be the better solution... so maybe a combination of these two PRs using `velox_sources` to selectively add the geos source file to the target but with the CI checks from #13541?

yingsu00:
> I prefer that we continue in #13541 as it checks it in CI.
> 
> But removing the sourcefile from the target instead of wrapping it in an #ifdef might be the better solution... so maybe a combination of these two PRs using `velox_sources` to selectively add the geos source file to the target but with the CI checks from #13541?

Thanks @assignUser. I agree that checking it in CI is good, and also it's better to set default VELOX_ENABLE_GEO to OFF as it seems not very stable now. I picked the mentioned changes from #13541 to this PR and added wraymo as a co-author. Also updated the CMakeLists.txt to use `velox_sources`. Your review is appreciated again. 
If you prefer to continue #13541, @wraymo please feel free to pick part of my commit to your commit and add me as a co-author.

CMakeLists.txt
```diff
@@ -675,6 +675,7 @@ if(VELOX_ENABLE_ARROW)
 endif()
 
 if(VELOX_ENABLE_GEO)
+  add_definitions(-DVELOX_ENABLE_GEO)
```
assignUser: Please use target/file based version and move it into this [if](https://github.com/facebookincubator/velox/pull/13851/files#diff-dd02356b3e62048cff4aea44c6b7202764835996dd5dfa16666cca020a9505afR42)
You can see how it was done for faiss just below.

velox/functions/prestosql/registration/CMakeLists.txt
```diff
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
assignUser: please keep the newline

velox/functions/prestosql/types/CMakeLists.txt
```diff
@@ -26,6 +25,10 @@ velox_add_library(
   TimestampWithTimeZoneRegistration.cpp
   UuidRegistration.cpp)
 
+if(VELOX_ENABLE_GEO)
```
assignUser, line 28: also add the definition here if required

velox/functions/prestosql/types/tests/CMakeLists.txt
```diff
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+add_library(velox_presto_types_test_lib TypeTestBase.cpp)
+target_link_libraries(
+  velox_presto_types_test_lib GTest::gtest)
```
assignUser: Not sure what the purpose is of having this extra lib? why not just conditionally add the geometrytypetest.cpp?

yingsu00: > Not sure what the purpose is of having this extra lib? why not just conditionally add the geometrytypetest.cpp?

@assignUser Thanks for reviewing. TypeTestBase is needed by velox_buffer_test, however it used to be in velox_presto_types_test, which was a test executable and therefore not linkable:
```
add_executable(
  velox_presto_types_test
  BingTileTypeTest.cpp
  HyperLogLogTypeTest.cpp
  JsonTypeTest.cpp
  TimestampWithTimeZoneTypeTest.cpp
  TDigestTypeTest.cpp
  TypeTestBase.cpp           # in an executable not a lib
  QDigestTypeTest.cpp
  TypeTestBase.cpp
  UuidTypeTest.cpp
  IPAddressTypeTest.cpp
  IPPrefixTypeTest.cpp)
```
So I extracted TypeTestBase.cpp into a separate library and link it to velox_buffer_test:
```
target_link_libraries(
  velox_buffer_test
  velox_buffer
  velox_memory
  velox_functions_test_lib
  velox_presto_types_test_lib # TypeTestBase needed by velox_buffer_test
  velox_test_util
  velox_vector_test_lib
  GTest::gtest
  GTest::gtest_main
  GTest::gmock
  glog::glog
  gflags::gflags
  pthread)
```
Please let me know if you have a better solution for this.

assignUser: But `TypeTestBase.h` is only included in files in `types/test` so why would the buffer test need it? Are there unclear transitive dependencies? What is the error you get if this is not added?

Sorry that I am poking at this but it seems like a unnecessary coupling between tests. (I might very well be wrong on this as I am only going of the build stuff ^^)

yingsu00: > But TypeTestBase.h is only included in files in types/test so why would the buffer test need it? Are there unclear transitive dependencies? What is the error you get if this is not added?

@assignUser Good question! After a day I finally found it was because the `velox_sources` was used on the test targets, specifically velox_presto_types_test:
```
add_executable(velox_presto_types_test …)
if(VELOX_ENABLE_GEO)
  velox_sources(velox_presto_types_test PRIVATE GeometryTypeTest.cpp)
endif()
```
velox_sources logic:
```
function(velox_sources TARGET)
  if(VELOX_MONO_LIBRARY)
    # IGNORE the TARGET name, always append sources into 'velox'
    target_sources(velox ${ARGN})
  else()
    target_sources(${TARGET} ${ARGN})
  endif()
endfunction()
```
This would add GeometryTypeTest.cpp into the monolithic library `velox`. Then when `velox_buffer_test` was linking with `velox_buffer`, it actually links to the monolithic library `velox`, which includes GeometryTypeTest.cpp but doesn't have `TypeTestBase` definitions. The fix was to change velox_sources on tests to target_sources and I have updated this PR accordingly.

Just as a side thought, I wonder if we could add some documentation to the `velox_add_libraries`, `velox_link_libraries`, `velox_sources`, etc. that they should only be used on production code, not test code? Or alternatively, throwing a warning when they were used on tests or test libs? That will prevent misusage and save potential debug times.

I do have a question about the `velox_link_libraries` implementation though. In that function you added `velox_exec_test_lib` into the `target_link_libraries(velox ${_arg})` whitelist, but I did a global search and didn't find any production targets ever links with it. It seems it's a NOOP there. Was there a specific reason it was added in the whitelist?

yingsu00:
@assignUser One more thing, I added the commit "fix: Fix parsing EXTRA_CMAKE_FLAGS in CI" to this PR to fix [CI doesn't honor EXTRA_CMAKE_FLAGS](https://github.com/facebookincubator/velox/issues/13991#top). I also have a standalone PR  https://github.com/facebookincubator/velox/pull/13992#top that only includes this commit. Let me know which PR you prefer it to come in. Thanks!

velox/functions/prestosql/types/tests/CMakeLists.txt
assignUser: Ah, sorry I should have caught that... :see_no_evil:  great work on tracking it down!

> Just as a side thought, I wonder if we could add some documentation 

Yes, I have been neglecting that, sorry.

> I do have a question about the velox_link_libraries implementation though. 

Hm, yes seems you are correct. I added that with the initial PR where it was necessary but it probably was refactored away at some point. I don't know the exact target it was used on anymore ^^

.github/workflows/scheduled.yml
```diff
@@ -203,8 +203,9 @@ jobs:
 
       - name: Build
         env:
-          EXTRA_CMAKE_FLAGS: >
+          EXTRA_CMAKE_FLAGS: |
             -DVELOX_ENABLE_ARROW=ON \
+            -DVELOX_ENABLE_GEO=ON \
```
assignUser: Ah good catch! I removed the `\` in a later commit but must have lost that when rebasing :facepalm: 
While your fix works removing the `\` is the more idiomatic approach imo. Would you mind changing it to that? I think we can keep the change in this PR :)

yingsu00: > Ah good catch! I removed the \ in a later commit but must have lost that when rebasing 🤦 While your fix works removing the \ is the more idiomatic approach imo. Would you mind changing it to that? I think we can keep the change in this PR :)

@assignUser Sure happy to do so. I've updated the first commit with your recommended change. Your review is very appreciated!

yingsu00:
@assignUser comments addressed. Could you please review again? Thanks!

.github/workflows/scheduled.yml
```diff
@@ -204,11 +204,7 @@ jobs:
       - name: Build
         env:
           EXTRA_CMAKE_FLAGS: >
-            -DVELOX_ENABLE_ARROW=ON \
-            -DVELOX_MONO_LIBRARY=ON \
-            -DVELOX_BUILD_PYTHON_PACKAGE=ON \
-            -DVELOX_PYTHON_LEGACY_ONLY=ON \
-            ${{ inputs.extraCMakeFlags }}
+            -DVELOX_ENABLE_ARROW=ON -DVELOX_ENABLE_GEO=ON -DVELOX_MONO_LIBRARY=ON -DVELOX_BUILD_PYTHON_PACKAGE=ON -DVELOX_PYTHON_LEGACY_ONLY=ON ${{ inputs.extraCMakeFlags }}
```
assignUser: Sorry I didn't phrase that well, we can keep the newlines but can drop the  `\`. The `>` will eat the newlines and make this into one line in the env. So we have both readability and function!
```suggestion
            -DVELOX_ENABLE_ARROW=ON
            -DVELOX_ENABLE_GEO=ON
            -DVELOX_MONO_LIBRARY=ON
            -DVELOX_BUILD_PYTHON_PACKAGE=ON
            -DVELOX_PYTHON_LEGACY_ONLY=ON
             ${{ inputs.extraCMakeFlags }}
```

CMakeLists.txt
assignUser: This should now be removed as you added a private flag to the target, no?

assignUser:
There are some lints and the test failure also seems related: https://github.com/facebookincubator/velox/actions/runs/16138478190/job/45540132846?pr=13851#step:11:1001

velox/common/base/tests/CMakeLists.txt
```diff
@@ -80,4 +80,6 @@ target_link_libraries(
     velox_exception
     velox_time
     Folly::folly
-    gflags::gflags)
+    gflags::gflags
+    GTest::gtest
+    GTest::gmock)
```
kou: It seems that we don't need this change because `Memcpy.cpp` isn't changed in this PR and `Memcpy.cpp` doesn't include `<gtest/gtest.h>` nor `<gmock/gmock.h>`.

velox/functions/prestosql/types/CMakeLists.txt
```diff
@@ -26,6 +25,11 @@ velox_add_library(
...
+  velox_compile_definitions(velox_functions_prestosql PRIVATE VELOX_ENABLE_GEO)
```
kou, line 29:
```suggestion
  velox_compile_definitions(velox_presto_types PRIVATE VELOX_ENABLE_GEO)
```

velox/expression/tests/CustomTypeTest.cpp
```diff
@@ -221,39 +221,30 @@ TEST_F(CustomTypeTest, customType) {
...
+      "HYPERLOGLOG",
+      "TIMESTAMP WITH TIME ZONE",
+      "UUID",
+      "IPADDRESS",
+      "IPPREFIX",
+      "BINGTILE",
+      "TDIGEST",
+      "QDIGEST"};
+
+#ifdef VELOX_ENABLE_GEO
```
kou, line 235: We need the following for this:

```diff
diff --git a/velox/expression/tests/CMakeLists.txt b/velox/expression/tests/CMakeLists.txt
index 036c13fc3..fbb0021ad 100644
--- a/velox/expression/tests/CMakeLists.txt
+++ b/velox/expression/tests/CMakeLists.txt
@@ -48,6 +48,9 @@ add_executable(
   TryExprTest.cpp
   VariadicViewTest.cpp
   VectorReaderTest.cpp)
+if(VELOX_ENABLE_GEO)
+  velox_compile_definitions(velox_expression_test PRIVATE VELOX_ENABLE_GEO)
+endif()
 
 add_test(
   NAME velox_expression_test
```

yingsu00: > We need the following for this:
> 
> ```diff
> diff --git a/velox/expression/tests/CMakeLists.txt b/velox/expression/tests/CMakeLists.txt
> index 036c13fc3..fbb0021ad 100644
> --- a/velox/expression/tests/CMakeLists.txt
> +++ b/velox/expression/tests/CMakeLists.txt
> @@ -48,6 +48,9 @@ add_executable(
>    TryExprTest.cpp
>    VariadicViewTest.cpp
>    VectorReaderTest.cpp)
> +if(VELOX_ENABLE_GEO)
> +  velox_compile_definitions(velox_expression_test PRIVATE VELOX_ENABLE_GEO)
> +endif()
>  
>  add_test(
>    NAME velox_expression_test
> ```

This doesn't work. I tried a few times in prior force pushes but they all failed no matter if I added it before or after. Using the following worked instead:
```
if(VELOX_ENABLE_GEO)
  add_definitions(-DVELOX_ENABLE_GEO)
#  velox_compile_definitions(velox_expression_test PRIVATE VELOX_ENABLE_GEO)  #doesn't work in velox_expression_test
endif()
```
I don't know why it didn't work here, although it seemed to have worked at the other places in this PR. I have added `add_definitions(-DVELOX_ENABLE_GEO)` and `remove_definitions(-DVELOX_ENABLE_GEO)` in velox/expression/tests/CMakeLists.txt. Can we merge as is and you guys can continue investigate why `velox_compile_definitions` doesn't work? I'm on vacation this week and don't want to cause delay in merging this PR.

Actually, I think adding a global VELOX_ENABLE_GEO definition makes more sense. It affects widely used core components including types and functions that can't be hidden under a common interface. On the contrary, VELOX_ENABLE_PARQUET or VELOX_ENABLE_XXX_CONNECTOR should be scoped after we decouple them from the other modules and provide file format-agnostic dwio interface and connector-agnostic connector interface.

kou: @yingsu00 Can I take over this? I also want to fix this as soon as possible for Nimble. If it's OK, I'll open a new PR with your commits and my fixes.

yingsu00: > @yingsu00 Can I take over this? I also want to fix this as soon as possible for Nimble. If it's OK, I'll open a new PR with your commits and my fixes.

@kou yes please go ahead. thanks!

kou: Thanks! I've opened #14103 .

kou:
We can close this because we merged https://github.com/facebookincubator/velox/pull/14103 .

PR #13855, opened 2025-06-23
fix: TestIndexSource::initOutputProjections

mbasmanova:
Summary:
TestIndexSource didn't support using column assignments in TableScan plan node to project columns under different names.

Also, introduce TestIndexConnectorFactory::registerConnector helper method to reduce boiler plate.

Differential Revision: D77173098


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #13858, opened 2025-06-24
feat: Support MAP Type in TextWriter

YiChengLee03:
Summary:
- Add support for complex type writes
- [] Fix deeply nested complex types (verification with reader is required)

Differential Revision: D77191837

### DIFF
--- a/velox/dwio/text/tests/writer/TextWriterTest.cpp
+++ b/velox/dwio/text/tests/writer/TextWriterTest.cpp
@@ -19,7 +19,6 @@
 #include "velox/common/file/FileSystems.h"
 #include "velox/dwio/text/RegisterTextReader.h"
 #include "velox/dwio/text/RegisterTextWriter.h"
-#include "velox/dwio/text/reader/TextReader.h"
 #include "velox/dwio/text/tests/writer/FileReaderUtil.h"
 #include "velox/exec/tests/utils/TempDirectoryPath.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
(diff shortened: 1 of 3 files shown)


velox/dwio/text/common/TextFileTraits.h
```diff
@@ -0,0 +1,69 @@
...
+  char getMapKeyDelim() {
+    return mapKeyDelim_;
+  }
+
+ private:
+  std::string kNullData_;
+  char kNewLine_;
+  char fieldDelim_;
+  char collectionDelim_;
+  char mapKeyDelim_;
```
kewang1024: we can define those as constant?

velox/dwio/text/common/TextFileTraits.h
```diff
@@ -0,0 +1,69 @@
...
+#include <string>
+
+class TextFileTraits {
+ public:
+  //// The following constants define the delimiters used by TextFile format.
+  /// Each row is separated by 'kNewLine'.
+  /// Each column is separated by 'kSOH' within each row.
+
+  TextFileTraits(
+      const std::string& nullData = "\\N",
```
kewang1024: Curious: how is text reader using this right now?

kewang1024 (commented):
Maybe we can put the extraction logic for TextFileTraits.h to a new PR which is better for review and have a separate PR for only complex types

velox/dwio/text/common/TextFileTraits.h
YiChengLee03: TextReader uses SerDeOptions for the delimiters. I already made the change in another diff that isn't ready for review but I can just create a new PR for TextFileTraits.h with the use of SerDeOptions

velox/dwio/text/common/TextFileTraits.h
YiChengLee03: Will use SerDeOptions instead

YiChengLee03:
> Maybe we can put the extraction logic for TextFileTraits.h to a new PR which is better for review and have a separate PR for only complex types

https://github.com/facebookincubator/velox/pull/13880 is a separate PR for refactoring TextFileTraits

velox/dwio/text/writer/TextWriter.h
```diff
@@ -78,14 +68,31 @@ class TextWriter : public dwio::common::Writer {
 
   void abort() override;
 
+  void setDelimiters(const std::vector<char>& delimiters);
```
kewang1024: This is not used? could we remove?

velox/dwio/text/writer/TextWriter.h
```diff
@@ -16,38 +16,26 @@
...
 struct WriterOptions : public dwio::common::WriterOptions {
   int64_t defaultFlushCount = 10 << 10;
 };
 
-// TODO: move to a separate file to be shared with text reader once it is in oss
-class TextFileTraits {
- public:
-  //// The following constants define the delimiters used by TextFile format.
-  /// Each row is separated by 'kNewLine'.
-  /// Each column is separated by 'kSOH' within each row.
```
kewang1024: is it possible to move the comment to Options.h as well?

velox/dwio/common/Options.h
```diff
@@ -88,11 +89,13 @@ class SerDeOptions {
       uint8_t collectionDelim = '\2',
       uint8_t mapKeyDelim = '\3',
       uint8_t escape = '\\',
-      bool isEscapedFlag = false)
+      bool isEscapedFlag = false,
+      uint8_t newLine = '\n')
```
kewang1024: Can we move the comment here? or those can also be used by non-text format?

kewang1024 (commented):
Haven't started the main logic review yet, some early comment

velox/dwio/text/writer/TextWriter.h
YiChengLee03: Ah yes, users should use set delimiters through SerDeOptions, not here.

velox/dwio/common/Options.h
YiChengLee03: I think only TEXT is using it but I can write a more generic comment here.

velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -189,6 +344,39 @@ void TextWriter::writeCellValue(
...
+      VectorEncoding::Simple::ROW,
+      "Text writer expects row vector input");
+
+  /**
+  TODO: verify if we could allow downcasting in the future. Current Java
+  implementation (coordinator layer) allows upcasting (i.e. from int32_t to
+  int64_t) but fails the query for downcasts. We need to check if for
+  upcastings, does coordinator have another layer of casting (changes schema
+  type before reaching worker nodes) or do we have to handle it in writer
+  */
```
kewang1024: Let's remove this comment for now and track it in your doc instead?

velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -189,6 +344,39 @@ void TextWriter::writeCellValue(
...
+  VELOX_CHECK(
+      data->type()->equivalent(*schema_),
+      "The file schema type should be equal with the input row vector type.");
+
```
kewang1024: Add a check for separators's size > 0

YiChengLee03:
The fuzzer failure is being fixed in #13945

velox/dwio/text/tests/writer/FileReaderUtil.cpp
```diff
@@ -18,6 +18,8 @@
 
 namespace facebook::velox::text {
 
+using dwio::common::SerDeOptions;
+
 std::string readFile(const std::string& name) {
   std::ifstream file(name);
```
zacw7: velox filesystem is the preferred way for file io

velox/dwio/text/tests/writer/TextWriterTest.cpp
```diff
@@ -143,6 +145,166 @@ TEST_F(TextWriterTest, write) {
...
+  const auto serDeOptions = dwio::common::SerDeOptions('\x01', '|', '#');
+  auto writer = std::make_unique<TextWriter>(
+      schema,
+      std::move(sink),
+      std::make_shared<text::WriterOptions>(writerOptions),
+      serDeOptions);
+  writer->write(data);
+  writer->close();
+
+  std::vector<std::vector<std::string>> result = parseTextFile(filePath);
```
zacw7: this is great verifying individual column values but can we also add a check on the raw file content just to make sure all delimiters are written as expected.

YiChengLee03: We will have a diff that uses TextReader to check if resulting vector is the same as the data vector that was written into the Textfile. I think we can leave it as it is now?

velox/dwio/common/Options.h
```diff
@@ -72,7 +72,27 @@ enum class SerDeSeparator {
...
  public:
+  /// The following members control how data is separated TEXT file:
+  ///
+  /// - 'separators': An array of separator characters used to delimit columns
+  ///   and nested data.
+  ///     - 'separators[0]' separates top-level columns.
+  ///     - 'separators[depth_]' separates nested ComplexType columns.
+  /// - 'newLine': The character used to separate rows in the file.
+  ///
+  /// Suppose we have a schema: ROW(MAP(VARCHAR(), ARRAY(BIGINT())))
```
zacw7: Does this mean this table has a single column which is a MAP?

velox/dwio/common/Options.h
```diff
@@ -72,7 +72,27 @@ enum class SerDeSeparator {
...
+  /// With the following configuration:
+  ///   - separators = [',', ':', ';']
+  ///   - newLine = '\n'
+  ///   - nullString = "NULL"
+  ///
+  /// A sample text file would look like this:
+  /// key1:varchar1;bigint2,bigint3\n
```
zacw7: Not sure if I follow how this row maps to the schema. Wondering where should I put "varchar1"?
```JSON
{
  "key1" : [2, 3]
}
```

YiChengLee03: Yes, thanks for catching this. This example is wrong, will have to update it. 
With separators as 
`[',' , ':', ';', '#']` 

our write should actually be 
`key1;2#3:key2;4#5`. 

`separators[0]` is column-level separation (not in use here)
`separators[1]` is pair-level separation (separating different keys)
`separators[2]` is key-value separation (differentiates between key and value)
`separators[3]` is key-complex-type separation (value is complex type, separates values)

kewang1024:
Synced up offline, two suggestions:
1. Move the refactoring change (move TextWriteTrait to Options) to different diff
2. Simplify the whole logic as 
```
case TypeKind::ARRAY: {
      const auto& arrVecPtr = decodedColumnVector->base()->as<ArrayVector>();
      const auto& indices = decodedColumnVector->indices();
      const auto& offset = arrVecPtr->offsetAt(indices[row]);
      const auto& size = arrVecPtr->sizeAt(indices[row]);

      auto slice = arrVecPtr->elements()->slice(offset, size);
      auto decodedElement =
          std::make_shared<DecodedVector>(DecodedVector(*slice));
      for (int i = offset; i < offset + size; ++i) {
        if (i == offset) {
          delimiter = std::nullopt;
        } else {
          delimiter = std::optional(TextFileTraits::kSOH);
        }
        writeCellValue(decodedElement, arrVecPtr->typeKind(), i, delimiter);
      }
      return;
    }
```

YiChengLee03:
> Synced up offline, two suggestions:
> 
> 1. Move the refactoring change (move TextWriteTrait to Options) to different diff
> 2. Simplify the whole logic as
> 
> ```
> case TypeKind::ARRAY: {
>       const auto& arrVecPtr = decodedColumnVector->base()->as<ArrayVector>();
>       const auto& indices = decodedColumnVector->indices();
>       const auto& offset = arrVecPtr->offsetAt(indices[row]);
>       const auto& size = arrVecPtr->sizeAt(indices[row]);
> 
>       auto slice = arrVecPtr->elements()->slice(offset, size);
>       auto decodedElement =
>           std::make_shared<DecodedVector>(DecodedVector(*slice));
>       for (int i = offset; i < offset + size; ++i) {
>         if (i == offset) {
>           delimiter = std::nullopt;
>         } else {
>           delimiter = std::optional(TextFileTraits::kSOH);
>         }
>         writeCellValue(decodedElement, arrVecPtr->typeKind(), i, delimiter);
>       }
>       return;
>     }
> ```

Refactoring diff is here. #14034

kewang1024 (commented):
Could you rebase to pick up the latest?

velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -85,7 +85,7 @@ char TextWriter::getDelimiterForDepth(uint8_t depth) const {
       serDeOptions_.separators.size(),
       "Depth {} exceeds maximum supported depth",
       depth);
-  return serDeOptions_.separators[depth];
+  return (char)serDeOptions_.separators[depth];
```
zacw7: is it possible to change the return type instead?

zacw7 (commented):
Thanks for the detailed unit tests.

velox/dwio/text/writer/TextWriter.cpp
YiChengLee03: Yes, I can pass uint8_t all the way until bufferedWriterSink_->write. Will do the same for other complex type supports as well.

kewang1024 (approved):
Same comment as @zacw7, otherwise LGTM, thanks @YiChengLee03

PR #13859, opened 2025-06-24
misc: Make PrestoQueryReplayRunner identify and skip unsupported plans

kagamiori:
Summary:
The current PrestoQueryReplayRunner doesn't support grouped execution and 
scaled writer. This diff makes it identify unsupported plans and return a 
kUnsupported error code.

Differential Revision: D77200027


velox/runner/tests/PrestoQueryReplayRunner.h
```diff
@@ -43,10 +45,11 @@ class PrestoQueryReplayRunner {
...
 
-  /// Runs a query with the given serialized plan fragments and returns the
-  /// results. The serialized plan fragments should have the same query id as
-  /// 'queryId'.
-  std::vector<RowVectorPtr> run(
+  /// Runs a query with the given serialized plan fragments and returns a pair
+  /// of the results and execution status. The result is meaningless and can be
+  /// empty if the status is not kSuccess. The serialized plan fragments should
+  /// have the same query id as 'queryId'.
+  std::pair<std::vector<RowVectorPtr>, Status> run(
```
kevinwilfong: what about returning

`std::pair<std::optional<std::vector<RowVectorPtr>>, Status>`

that way if status is not success, you can just return std::nullopt which is clearly meaningless

velox/runner/tests/PrestoQueryReplayRunner.cpp
```diff
@@ -320,22 +348,32 @@ PrestoQueryReplayRunner::deserializeConnectorSplits(
...
 
-  auto result = readCursor(localRunner, pool_);
-  localRunner->waitForCompletion(kWaitTimeoutUs);
-
-  return result;
+  std::vector<RowVectorPtr> result;
+  try {
+    result = readCursor(localRunner, pool_);
+    localRunner->waitForCompletion(kWaitTimeoutUs);
+  } catch (...) {
```
kevinwilfong: could you log the exception? might make debugging easier in the future

velox/runner/tests/PrestoQueryReplayRunner.cpp
```diff
@@ -217,7 +217,32 @@ std::vector<std::string> PrestoQueryReplayRunner::getTaskPrefixes(
   return taskPrefixes;
 }
 
-MultiFragmentPlanPtr PrestoQueryReplayRunner::deserializePlan(
+bool isSupportedImpl(const core::PlanNodePtr& node) {
+  // We don't support arbitrary partitioning yet.
+  if (auto partitionedOutput =
+          std::dynamic_pointer_cast<const core::PartitionedOutputNode>(node)) {
+    return !partitionedOutput->isArbitrary();
```
kevinwilfong: You're not visiting the source of the PartitionedOutputNode.

It doesn't matter now since you're only checking PartitionedOutputNode and that can't be the source for anything, but if you add checks for other nodes in the future they may be missed.

kevinwilfong (approved):

velox/runner/tests/PrestoQueryReplayRunner.cpp
kagamiori: Good catch. I've updated this method to check the sources for PartitionedOutputNode when it's not arbitrary. Thanks!

PR #13860, opened 2025-06-24
refactor(TopNRowNumber): Abstract computeNextRankInMemory(InSpill) functions in getOutput() logic

aditi-pandit:
Last refactoring towards https://github.com/facebookincubator/velox/pull/11554

### DIFF
--- a/velox/exec/TopNRowNumber.cpp
+++ b/velox/exec/TopNRowNumber.cpp
@@ -213,6 +213,10 @@ void TopNRowNumber::addInput(RowVectorPtr input) {
     // Otherwise, check if row should replace an existing row or be discarded.
     processInputRowLoop(numInput);
 
+    // It is determined that the TopNRowNumber (as a partial) is not rejecting
+    // enough input rows to make the duplicate detection worthwhile. Hence,
+    // abandon the processing at this partial TopN and let the final TopN do
+    // the processing.
     if (abandonPartialEarly()) {
       abandonedPartial_ = true;
       addRuntimeStat("abandonedPartial", RuntimeCounter(1));
@@ -356,9 +360,14 @@ void TopNRowNumber::updateEstimatedOutputRowSize() {
 }
 
 TopNRowNumber::TopRows* TopNRowNumber::nextPartition() {
+  auto setNextRank = [&](const TopRows& partition) {
+    nextRank_ = partition.topRank;
+  };
+
   if (!table_) {
     if (!outputPartitionNumber_) {
       outputPartitionNumber_ = 0;
+      setNextRank(*singlePartition_);
       return singlePartition_.get();
     }
     return nullptr;
@@ -384,7 +393,15 @@ TopNRowNumber::TopRows* TopNRowNumber::nextPartition() {
     }
   }
 
-  return &partitionAt(partitions_[outputPartitionNumber_.value()]);
+  auto partition = &partitionAt(partitions_[outputPartitionNumber_.value()]);
+  setNextRank(*partition);
+  return partition;
+}
+
+void TopNRowNumber::computeNextRankInMemory(
+    const TopRows& /*partition*/,
+    vector_size_t /*rowIndex*/) {
+  nextRank_ -= 1;
 }
 
 void TopNRowNumber::appendPartitionRows(
@@ -398,10 +415,13 @@ void TopNRowNumber::appendPartitionRows(
   for (auto i = 0; i < numRows; ++i) {
     const auto index = outputOffset + i;
     if (rowNumbers) {
-      rowNumbers->set(index, rowNumber--);
+      rowNumbers->set(index, nextRank_);
     }
     outputRows_[index] = partition.rows.top();
     partition.rows.pop();
+    if (!partition.rows.empty()) {
+      computeNextRankInMemory(partition, index);
+    }
   }
 }
 
@@ -417,7 +437,7 @@ RowVectorPtr TopNRowNumber::getOutput() {
       return output;
     }
 
-    // We may have input accumulated in 'data_'.
+    // There could be older rows accumulated in 'data_'.
     if (data_->numRows() > 0) {
       return getOutputFromMemory();
     }
@@ -426,6 +446,7 @@ RowVectorPtr TopNRowNumber::getOutput() {
       finished_ = true;
     }
 
+    // There is no data to return at this moment.
     return nullptr;
   }
 
@@ -433,6 +454,8 @@ RowVectorPtr TopNRowNumber::getOutput() {
     return nullptr;
   }
 
+  // All the input data is received, so the operator can start producing
+  // output.
   RowVectorPtr output;
   if (merge_ != nullptr) {
     output = getOutputFromSpill();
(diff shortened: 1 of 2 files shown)


velox/exec/TopNRowNumber.h
```diff
@@ -150,7 +156,22 @@ class TopNRowNumber : public Operator {
...
+      vector_size_t startColumn,
+      vector_size_t endColumn);
+
+  // Computes next rank value for spill output.
+  void computeNextRankInSpill(
+      const RowVectorPtr& output,
+      vector_size_t index,
+      const SpillMergeStream* next);
 
   // Sets nextRowNumber_ to rowNumber. Checks if next row in 'merge_' belongs to
```
JkSelf: @aditi-pandit Can you also update this commentto use `nextRank_ `to replace the `nextRowNumber_ `?

JkSelf (commented):
LGTM. Except one small nit. Thanks.

velox/exec/TopNRowNumber.cpp
```diff
@@ -353,9 +357,12 @@ void TopNRowNumber::updateEstimatedOutputRowSize() {
 }
 
 TopNRowNumber::TopRows* TopNRowNumber::nextPartition() {
+  auto setNextRank = [&](TopRows& partition) { nextRank_ = partition.topRank; };
```
jinchengchenghh: const TopRows&

velox/exec/TopNRowNumber.cpp
```diff
@@ -527,37 +550,53 @@ bool TopNRowNumber::isNewPartition(
...
   }
 
-  if (isNewPartition(output, output->size(), lookAhead)) {
-    nextRowNumber_ = 0;
+  nextRank_ += 1;
+  return;
+}
+
+void TopNRowNumber::setupNextOutput(const RowVectorPtr& output) {
+  auto resetNextRank = [this]() { nextRank_ = 1; };
```
jinchengchenghh, line 580: Why you reset the nextRank_ = 1 and use <= to replace < before?

velox/exec/TopNRowNumber.h
aditi-pandit: Done.

velox/exec/TopNRowNumber.cpp
aditi-pandit: @jinchengchenghh I will be updating this logic for both rank and dense_rank computation in the next PR. 

In rank and dense_rank the result value changes only if order by keys are different between  current row and next. By setting starting rank as 0 and then incrementing in place if different from adjacent row is complicated logic...Its simpler to set current rank as 1 and increment it when advancing to the next row. So I changed how the result value is computed.

velox/exec/TopNRowNumber.h
```diff
@@ -260,7 +281,8 @@ class TopNRowNumber : public Operator {
   // Used to sort-merge spilled data.
   std::unique_ptr<TreeOfLosers<SpillMergeStream>> merge_;
 
-  // Row number for the first row in the next output batch from the spiller.
-  int32_t nextRowNumber_{0};
+  // Row number ( or rank or dense_rank in the future) for the next row being
```
jinchengchenghh: ( or -> (or

aditi-pandit: done.

jinchengchenghh (approved):
Approve for the code, I'm not familiar with TopN

velox/exec/TopNRowNumber.cpp
czentgr: I'm not too familiar with this operator.
We can use 1 because we have at least one row result (if it doesn't match any other row) that has rank/row_number 1.

aditi-pandit: @czentgr : Yes.

aditi-pandit:
@xiaoxmeng : Please can you help with the review. Have bunch of approvals from maintainers already.

velox/exec/TopNRowNumber.cpp
```diff
@@ -512,13 +535,15 @@ RowVectorPtr TopNRowNumber::getOutputFromMemory() {
   return output;
 }
 
-bool TopNRowNumber::isNewPartition(
+bool TopNRowNumber::compareSpillRowColumns(
     const RowVectorPtr& output,
     vector_size_t index,
-    SpillMergeStream* next) {
+    const SpillMergeStream* next,
+    vector_size_t startColumn,
```
xiaoxmeng, line 542: Why we pass startColumn and endColumn but not 0, numPartitionKeys_? And why call this as compareSpillRowColumns?

velox/exec/TopNRowNumber.cpp
```diff
@@ -583,24 +624,20 @@ RowVectorPtr TopNRowNumber::getOutputFromSpill() {
...
     auto next = merge_->next();
     if (next == nullptr) {
       break;
     }
 
-    // Check if this row comes from a new partition.
-    if (index > 0 && isNewPartition(output, index, next)) {
-      rowNumber = 0;
+    if (index > 0) {
+      computeNextRankInSpill(output, index, next);
```
xiaoxmeng:
```
s/computeNextRankInSpill/computeNextRankFromSpill/
```

velox/exec/TopNRowNumber.cpp
```diff
@@ -530,37 +555,53 @@ bool TopNRowNumber::isNewPartition(
...
-  if (isNewPartition(output, output->size(), lookAhead)) {
-    nextRowNumber_ = 0;
+  nextRank_ += 1;
+  return;
+}
+
+void TopNRowNumber::setupNextOutput(const RowVectorPtr& output) {
+  auto resetNextRank = [this]() { nextRank_ = 1; };
+
+  auto* lookAhead = merge_->next();
```
xiaoxmeng, line 582: Let's just call this next?

velox/exec/TopNRowNumber.cpp
```diff
@@ -530,37 +555,53 @@ bool TopNRowNumber::isNewPartition(
   return false;
 }
 
-void TopNRowNumber::setupNextOutput(
+// Compares the partition keys for new partitions.
+bool TopNRowNumber::isNewPartition(
```
xiaoxmeng, line 559: Let's merge isNewPartition and compareSpillRowColumns? And call this isNewPartitionFromSpill?

xiaoxmeng (commented):
@aditi-pandit LGTM % minors. Thanks!

velox/exec/TopNRowNumber.cpp
aditi-pandit: @xiaoxmeng : This code will be enhanced to do the topn optimization for rank functions as well... and those require to compare the rows on the order_by columns. So I abstracted the function for compareSpillRowColumns which will be reused to compare the order by columns as well.

velox/exec/TopNRowNumber.cpp
aditi-pandit: Same response as first question. We need another function for comparing order_by columns for rank and so compareSpillRowColumns is abstracted.

velox/exec/TopNRowNumber.cpp
aditi-pandit: The loop below uses the next variable naming at line 595, so I stuck with the original lookAhead name here.

velox/exec/TopNRowNumber.cpp
aditi-pandit: Done.

aditi-pandit:
@xiaoxmeng : Have addressed your review comments. PTAL.

velox/exec/TopNRowNumber.cpp
xiaoxmeng: I mean let's just use next why have two variables: next and lookAhead?

PR #13862, opened 2025-06-24
feat: Support spill write batch size limit

### LINKED ISSUE #13861: Support spill write batch size limit
### Description

Currently, spill write only supports the buffer size limit. When the row size is small, it may result in a row vector with a large number of rows. This may cause the subsequent operator calculation performance to deteriorate. The scenario I encountered was that `DecodedVector::fillInIndices()` cache missing caused creation of the indices vector to be expensive.

https://github.com/facebookincubator/velox/blob/8ee33c1b82a40dfe37d8cb4bc6ed9d3d47769e2e/velox/vector/DecodedVector.cpp#L301-L304

After adding this limit (4096) the performance improved significantly. (test on spark with gluten+velox)

before:

Image

Image

after:

Image

Image

wForget:
closes #13861

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -312,12 +312,11 @@ class QueryConfig {
       "spill_write_buffer_size";
 
   /// Specifies the maximum number of rows to buffer before flushing spill data
-  /// to storage. The spiller will flush the buffer if either the total serialized
-  /// data size reaches 'spill_write_buffer_size' bytes or the number of buffered
-  /// rows reaches this threshold, whichever comes first. If set to zero, then spill write
-  /// buffering is disabled.
-  static constexpr const char* kSpillWriteBatchSize =
-      "spill_write_batch_size";
+  /// to storage. The spiller will flush the buffer if either the total
+  /// serialized data size reaches 'spill_write_buffer_size' bytes or the number
+  /// of buffered rows reaches this threshold, whichever comes first. If set to
+  /// zero, then spill write buffering is disabled.
+  static constexpr const char* kSpillWriteBatchSize = "spill_write_batch_size";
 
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
Thanks!
Please update the configs.rst

wForget:
> Thanks! Please update the configs.rst

Thank you, updated

wForget:
@xiaoxmeng @majetideepak Could you please take a look?

wForget:
Gentle ping @xiaoxmeng @majetideepak, sorry to bother you again, could you please review it if you have time?

majetideepak (commented):
@wForget can you share the motivation behind this feature? 
How would users configure both the buffer size and batch size in general?

wForget:
> @wForget can you share the motivation behind this feature?

My motivation is to make the size of the RowVector after deserializing SpillFile closer to this batch size. As observed in #13861, excessively large row sizes may lead to performance regression.

> How would users configure both the buffer size and batch size in general?

The spill buffer will be flushed when either the total serialized size reaches `spill_write_buffer_size` or the number of buffered rows reaches `spill_write_batch_size`. The default value for `spill_write_buffer_size` is `4MB`, and the default value for `spill_write_batch_size` is `4096` (Or should we let it be closer to `preferred_output_batch_rows` ?).

velox/common/base/SpillConfig.h
```diff
@@ -109,6 +110,10 @@ struct SpillConfig {
   /// storage system for io efficiency.
   uint64_t writeBufferSize;
 
+  /// Specifies the batch rows size to buffer the serialized spill data before
+  /// write to storage system
+  uint64_t writeBatchSize;
```
jinchengchenghh, line 115: uint32_t?

jinchengchenghh (approved):

PR #13872, opened 2025-06-24
feat: Implement quantile_at_value(QDigest) function

duxiao1212:
Summary: as title

Differential Revision: D76951311


natashasehgal (approved):
LGTM

Before merge, can you confirm that the accuracy change in QDigestInputGenerator from 0.05 to 1.0E-3 is acceptable for QDigest and matches library?

PR #13877, opened 2025-06-25
fix: Fix arbitrary streaming aggregation optimization issue and reland

xiaoxmeng:
Summary: There is a bug in previous non-numeric arbitrary streaming aggregation implementation. If the input is from a single source and copy range is one, then we return a slice of the input source as the result. This is not always correct as we might have head or tail nulls. So we can only take slice if (1) single source; (2) copy range is one; (3) there is no nulls at head or tail. Note if nulls appears in the middle, then copy range won't be one. Verified with prestissimo query that crash as well as unit test. The unit test leads incorrect result.

Differential Revision: D77284259


Yuhta (approved):

spershin (approved):
Thank you!

PR #13878, opened 2025-06-25
refactor: Clean up unnecessary class abstraction

YiChengLee03:
Summary: Clean up an extra layer of abstraction

Differential Revision: D76927989


YiChengLee03:
Thanks for the review @Yuhta

PR #13880, opened 2025-06-25
refactor: TextFileTraits

YiChengLee03:
Summary: Refactor TextFileTraits and use SerDeOptions for delimiters

Differential Revision: D77317489


kewang1024:
Thank you for the refactoring, could we also include the changes in TextWriter to adopt it?

kewang1024:
Discussed offline, merge with SerDeOptions which is commonly shared between Text reader and writer
cc: @zacw7

PR #13881, opened 2025-06-25
feat: Add quantileAtValue in QuantileDigest

duxiao1212:
Summary: To support the quantile_at_value(Qdigest) expression, expose the quantileAtValue function in the QuantileDigest.h

Differential Revision: D77314427

### DIFF
--- a/velox/functions/lib/QuantileDigest.h
+++ b/velox/functions/lib/QuantileDigest.h
@@ -86,6 +86,10 @@ class QuantileDigest {
   // on 'serialized'.
   void testingMerge(const QuantileDigest& other);
 
+  // Returns nullopt when the digest is empty or the value is out of the
+  // min/max range.
+  std::optional<double> quantileAtValue(T value) const;
+
  private:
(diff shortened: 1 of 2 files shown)


velox/functions/lib/QuantileDigest.h
```diff
@@ -1266,6 +1268,57 @@ int64_t QuantileDigest<T, Allocator>::serialize(char* out) {
   return out - outStart;
 }
 
+template <typename T, typename Allocator>
+std::optional<double> QuantileDigest<T, Allocator>::quantileAtValue(
```
kagamiori, line 1274: This algorithm can be simpler. Please follow this logic used in Presto-Java: https://github.com/airlift/airlift/blob/0423dad37e464351f23d4480d0332332f40cf022/stats/src/main/java/io/airlift/stats/QuantileDigest.java#L449-L487. (This code can be further simplified since the `bucketUpperBounds` would only have one element. Also, `normalizationFactor` is 1.0.)

kagamiori (approved):

PR #13882, opened 2025-06-25
misc: Add a defensive check for nullptr from allocation

HuamengJiang:
Summary:
Per suggestion from Yuhta, we can be more defensive in OOMing scenarios. Though typically we shouldn't see nullptr being returned by malloc or realloc. This might mean we have an edge case in error handling that needs follow-up.

We recently saw crashes in OOM scenarios such as T227957911.

Differential Revision: D77318087


amitkdutta:
Wondering if it's possible to have a unit test on why this code is added. Or we are just guessing it can happen?

PR #13885, opened 2025-06-25
misc: Add dummy memory reclaimer for sys root for the query ctx use case

xiaoxmeng:
Summary: Meta internal use might need to create query ctx with default system root pool which doesn't expect memory arbitration. However the query ctx will set memory reclaimer for memory arbitration integration. This PR adds a dummy memory reclaimer for system root pool just in case if memory arbitration is unexpected triggered to catch.

Differential Revision: D77331518


mbasmanova (approved):
Thanks.

PR #13887, opened 2025-06-25
misc: Move arrow logging include to avoid potential macro errors

czentgr:
Arrow logging defines macro DCHECK which can be redefined and result in an error if VeloxExceptions is included after it.

This PR removes the logging.h from Hashing.h because it is not used there.
It adds it to Encoding.cpp due to use of some arrow macros. VeloxExceptions
is not used and doesn’t cause a problem there.


czentgr:
@majetideepak FYI.

majetideepak (approved):

PR #13894, opened 2025-06-26
fix: Do not crash when reading struct child larger than number of top level row

Yuhta:
Summary:
When we read file that has child vector larger than struct parent
vector (unlike in-memory case, this is suboptimal in file because the extra rows
in child would never be used), and when we apply filter on it in fallback mode
(e.g. delta update), we use the size of child vector instead of parent and
result in out of boundary access (because the result bitmask is created using top level row count).

Differential Revision: D77337719

### DIFF
--- a/velox/dwio/common/ScanSpec.cpp
+++ b/velox/dwio/common/ScanSpec.cpp
@@ -528,6 +528,7 @@ void filterRows(
     const Filter& filter,
     vector_size_t size,
     uint64_t* result) {
+  VELOX_CHECK_LE(size, vector.size());
   switch (vector.typeKind()) {
     case TypeKind::ARRAY:
     case TypeKind::MAP:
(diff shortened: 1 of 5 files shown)


velox/dwio/common/ScanSpec.h
```diff
@@ -358,7 +358,10 @@ class ScanSpec {
   /// Apply filter to the input `vector' and set the passed bits in `result'.
   /// This method is used by non-selective reader and delta update, so it
   /// ignores the filterDisabled_ state.
-  void applyFilter(const BaseVector& vector, uint64_t* result) const;
+  void applyFilter(
+      const BaseVector& vector,
+      vector_size_t size,
```
bikramSingh91, line 367: nit: can you update the description to add context about size and why it is necessary

velox/dwio/common/ScanSpec.cpp
```diff
@@ -558,9 +558,12 @@ void filterRows(
...
 } // namespace
 
-void ScanSpec::applyFilter(const BaseVector& vector, uint64_t* result) const {
+void ScanSpec::applyFilter(
+    const BaseVector& vector,
+    vector_size_t size,
+    uint64_t* result) const {
   if (filter_) {
-    filterRows(vector, *filter_, vector.size(), result);
+    filterRows(vector, *filter_, size, result);
```
bikramSingh91, line 567: do we handle the case where the `size` can be larger than child->size() ? I noticed  at L540-L542 we use the `vector.isNullAt(i)` which can be problematic.
can you also add a test case for this?

bikramSingh91 (approved):

PR #13895, opened 2025-06-26
fix: Unloaded lazy vector in duplicate field project

rui-mo:
The following error was noticed when running a query as below, where the filter 
expressions were for one field and duplicate projects were for another field.

> SELECT id, id FROM tmp WHERE name = 'John'

The lazy vector is not preloaded because it’s only referenced in identity 
projections, not in expressions, and the `fillOutput` calls `wrapInDictionary` 
multiple times on the same lazy vector. The lazy vector is still unloaded at 
the time, and thus, the dictionary vector wraps the same underlying lazy vector 
multiple times while unloaded.

This PR fixes this issue by caching the already wrapped children to avoid 
wrapping the same child multiple times.

```
E20250626 11:53:16.482720 23136090 Exceptions.h:66] Line: /Users/velox/./velox/vector/DictionaryVector-inl.h:38, Function:setInternalState, Expression: dictionaryValues_->markAsContainingLazyAndWrapped() An unloaded lazy vector cannot be wrapped by two different top level vectors., Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: An unloaded lazy vector cannot be wrapped by two different top level vectors.
Retriable: False
Expression: dictionaryValues_->markAsContainingLazyAndWrapped()
Context: Operator: FilterProject[2] 1
Function: setInternalState
File: /Users/velox/./velox/vector/DictionaryVector-inl.h
Line: 38
Stack trace:
```

Benchmark results:

Before fix:

```
duplicate_project_integer                                   7.88ms    126.97
duplicate_project_string                                   57.75ms     17.32
filter_integers_duplicate_project_integers                 10.19ms     98.15
filter_strings_duplicate_project_strings                   47.57ms     21.02
```

After fix:

```
duplicate_project_integer                                   7.68ms    130.17
duplicate_project_string                                   55.27ms     18.09
filter_integers_duplicate_project_integers                 11.24ms     89.00
filter_strings_duplicate_project_strings                   47.55ms     21.03
```

### DIFF
--- a/velox/exec/OperatorUtils.cpp
+++ b/velox/exec/OperatorUtils.cpp
@@ -505,12 +505,23 @@ void projectChildren(
     const std::vector<IdentityProjection>& projections,
     int32_t size,
     const BufferPtr& mapping) {
+  // Cache for already wrapped children to avoid wrapping the same child
+  // multiple times.
+  std::unordered_map<int32_t, VectorPtr> wrappedChildren;
+
   for (auto [inputChannel, outputChannel] : projections) {
     if (outputChannel >= projectedChildren.size()) {
       projectedChildren.resize(outputChannel + 1);
     }
-    projectedChildren[outputChannel] =
-        wrapChild(size, mapping, src[inputChannel]);
+    auto it = wrappedChildren.find(inputChannel);
+    if (it == wrappedChildren.end()) {
+      auto wrapped = wrapChild(size, mapping, src[inputChannel]);
+      wrappedChildren[inputChannel] = wrapped;
+      projectedChildren[outputChannel] = wrapped;
+    } else {
+      // Reuse wrapped child.
+      projectedChildren[outputChannel] = it->second;
+    }
   }
 }
 
--- a/velox/exec/benchmarks/CMakeLists.txt
+++ b/velox/exec/benchmarks/CMakeLists.txt
@@ -137,3 +137,12 @@ target_link_libraries(
   velox_vector_fuzzer
   velox_vector_test_lib
   Folly::follybenchmark)
+
+add_executable(velox_exec_bm_duplicate_project DuplicateProjectBenchmark.cpp)
+
+target_link_libraries(
+  velox_exec_bm_duplicate_project
+  velox_exec_test_lib
+  velox_hive_connector
+  velox_vector_fuzzer
+  Folly::follybenchmark)
--- a/velox/exec/benchmarks/DuplicateProjectBenchmark.cpp
+++ b/velox/exec/benchmarks/DuplicateProjectBenchmark.cpp
@@ -1,0 +1,138 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <folly/Benchmark.h>
+#include <folly/init/Init.h>
+
+#include "velox/common/memory/Memory.h"
+#include "velox/common/memory/SharedArbitrator.h"
+#include "velox/exec/tests/utils/AssertQueryBuilder.h"
+#include "velox/exec/tests/utils/HiveConnectorTestBase.h"
+#include "velox/exec/tests/utils/PlanBuilder.h"
+#include "velox/functions/prestosql/registration/RegistrationFunctions.h"
+
+DEFINE_int64(fuzzer_seed, 99887766, "Seed for random input dataset generator");
+
+using namespace facebook::velox;
+using namespace facebook::velox::exec;
+using namespace facebook::velox::test;
+using namespace facebook::velox::exec::test;
+
+namespace {
+
+static constexpr int32_t kNumVectors = 50;
+static constexpr int32_t kRowsPerVector = 10'000;
+
+class DuplicateProjectBenchmark : public HiveConnectorTestBase {
+ public:
+  explicit DuplicateProjectBenchmark() {
+    HiveConnectorTestBase::SetUp();
+
+    inputType_ = ROW({
+        {"a", INTEGER()},
+        {"b", INTEGER()},
+        {"c", VARCHAR()},
+        {"d", VARCHAR()},
+    });
+
+    VectorFuzzer::Options opts;
+    opts.vectorSize = kRowsPerVector;
+    opts.nullRatio = 0.2;
+    VectorFuzzer fuzzer(opts, pool_.get(), FLAGS_fuzzer_seed);
+    std::vector<RowVectorPtr> inputVectors;
+    for (auto i = 0; i < kNumVectors; ++i) {
+      std::vector<VectorPtr> children;
+      children.emplace_back(fuzzer.fuzzFlat(INTEGER()));
+      children.emplace_back(fuzzer.fuzzFlat(INTEGER()));
+      children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));
+      children.emplace_back(fuzzer.fuzzFlat(VARCHAR()));
+
+      inputVectors.emplace_back(makeRowVector(inputType_->names(), children));
+    }
+
+    sourceFilePath_ = TempFilePath::create();
+    writeToFile(sourceFilePath_->getPath(), inputVectors);
+  }
+
+  ~DuplicateProjectBenchmark() override {
+    HiveConnectorTestBase::TearDown();
+  }
+
+  void TestBody() override {}
+
+  void run(const std::string& filter, const std::vector<std::string>& project) {
+    auto plan = PlanBuilder()
+                    .tableScan(inputType_)
+                    .filter(filter)
+                    .project(project)
+                    .planNode();
+    auto result = exec::test::AssertQueryBuilder(plan)
+                      .split(makeHiveConnectorSplit(sourceFilePath_->getPath()))
+                      .copyResults(pool_.get());
+    auto numResultRows = result->size();
+    folly::doNotOptimizeAway(numResultRows);
+  }
+
+  void addBenchmarks() {
+    folly::addBenchmark(
+        __FILE__, "filter_integer_duplicate_project_integer", [this]() {
+          run("a > 99999", {"b", "b"});
+          return 1;
+        });
+    folly::addBenchmark(
+        __FILE__, "filter_string_duplicate_project_string", [this]() {
+          run("length(c) > 10", {"d", "d"});
+          return 1;
+        });
+    folly::addBenchmark(__FILE__, "duplicate_project_integer", [this]() {
+      run("true", {"b", "b"});
+      return 1;
+    });
+    folly::addBenchmark(__FILE__, "duplicate_project_string", [this]() {
+      run("true", {"d", "d"});
+      return 1;
+    });
+    folly::addBenchmark(
+        __FILE__, "filter_integers_duplicate_project_integers", [this]() {
+          run("a > 9999 and b < 10000", {"a", "a", "b", "b"});
+          return 1;
+        });
+    folly::addBenchmark(
+        __FILE__, "filter_strings_duplicate_project_strings", [this]() {
+          run("length(c) > 10 and length(d) < 20", {"c", "c", "d", "d"});
+          return 1;
+        });
+  }
+
+ private:
+  RowTypePtr inputType_;
+  std::shared_ptr<TempFilePath> sourceFilePath_;
+};
+
+} // namespace
+
+int main(int argc, char** argv) {
+  folly::Init init(&argc, &argv);
+  facebook::velox::memory::MemoryManager::initialize(
+      facebook::velox::memory::MemoryManager::Options{});
+  memory::SharedArbitrator::registerFactory();
+  functions::prestosql::registerAllScalarFunctions();
+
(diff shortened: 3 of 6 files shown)


rui-mo:
@bikramSingh91 Would you like to help review this fix? Thank you.

velox/exec/FilterProject.cpp
```diff
@@ -134,12 +134,28 @@ void FilterProject::initialize() {
       auto fieldIndex = inputType->getChildIdx(field->name());
       distinctFieldIndices.insert(fieldIndex);
     }
+
+    // Collect indices of fields that are multiply referenced in identity
+    // projections. These are fields that are referenced in both expressions and
+    // identity projections, or are referenced multiple times in the identity
+    // projections themselves.
+    std::unordered_set<uint32_t> multiplyReferencedFieldIndices;
+    std::unordered_map<column_index_t, uint32_t> identityCounts;
```
leesf: Thanks for the fix, will this fix cause any performance degradation? we found that the performance degrade after this patch.

rui-mo: Could you elaborate what kind of case you notice perf regression in? I'd like to benchmark to check. Thanks.

leesf: > Could you elaborate what kind of case you notice perf regression in? I'd like to benchmark to check. Thanks.

In our users workflow, it contains thousands of sqls and hard to find specific sql perf regression, we just find the end-to-end time increased.

FelixYBW: Can the pr fix https://github.com/apache/incubator-gluten/issues/9965?

rui-mo: @FelixYBW Yes, this PR fixes this issue, confirmed with her offline.

rui-mo: @leesf After adding a benchmark, I discovered that the prior fix may preload duplicate fields when unnecessary, which can result in a performance degradation of about 10%. I updated this PR with a new fix, so performance shouldn't be impacted going forward. Thanks for the pointer.

leesf (approved):
LGTM

rui-mo:
@mbasmanova @pedroerp Would you like to take review? Thanks!

velox/exec/OperatorUtils.cpp
```diff
@@ -504,12 +504,23 @@ void projectChildren(
     const std::vector<IdentityProjection>& projections,
     int32_t size,
     const BufferPtr& mapping) {
+  // Cache for already wrapped children to avoid wrapping the same child
```
mbasmanova, line 508: Would you update the documentation for projectChildren API in the header file and add a unit test?

velox/exec/tests/TableScanTest.cpp
```diff
@@ -5995,5 +5995,28 @@ TEST_F(TableScanTest, textfileLarge) {
...
+          makeFlatVector<std::string>({"Alice", "John"}),
+      })};
+
+  auto file = TempFilePath::create();
+  writeToFile(file->getPath(), vectors);
+  createDuckDbTable(vectors);
+
+  auto plan = PlanBuilder()
+                  .tableScan(ROW({"id", "name"}, {INTEGER(), VARCHAR()}))
+                  .filter("name is not null and name = 'John'")
```
mbasmanova: Do we need this filter? Can the issue be reproduced with just 'project'?

velox/exec/tests/TableScanTest.cpp
```diff
@@ -5995,5 +5995,28 @@ TEST_F(TableScanTest, textfileLarge) {
...
+          makeFlatVector<int32_t>({1, 2}),
+          makeFlatVector<std::string>({"Alice", "John"}),
+      })};
+
+  auto file = TempFilePath::create();
+  writeToFile(file->getPath(), vectors);
+  createDuckDbTable(vectors);
+
+  auto plan = PlanBuilder()
+                  .tableScan(ROW({"id", "name"}, {INTEGER(), VARCHAR()}))
```
mbasmanova: nit: use vectors.front()->rowType()

It might be helpful to add overrides to writeToFile and createDuckDbTable to allow passing a single vector so we won't be forced to create an std::vector of size 1.

velox/exec/tests/TableScanTest.cpp
rui-mo: Without the filter, this issue does not arise because `wrapResults` is false and [wrapChild](https://github.com/facebookincubator/velox/blob/934c29dc5bfd61574ee847c7653b8f76dc1dd01d/velox/exec/OperatorUtils.cpp#L355) does not wrap the child in dictionary.

https://github.com/facebookincubator/velox/blob/934c29dc5bfd61574ee847c7653b8f76dc1dd01d/velox/exec/Operator.cpp#L305

> Is it fair to say that the issue can be reproduced with a simple query SELECT a, a, a FROM t

For the reason above, filter is still needed to reproduce this problem. Thanks.

velox/exec/OperatorUtils.cpp
rui-mo: Updated the documentation and added a test case in the `TEST_F(OperatorUtilsTest, projectChildren)`.

velox/exec/tests/TableScanTest.cpp
rui-mo: There's the below `writeToFile` implementation which accepts single row vector.
```
void HiveConnectorTestBase::writeToFile(
    const std::string& filePath,
    RowVectorPtr vector)
```

I change to use it and call `createDuckDbTable({vector})`, so `vector->rowType()` could be used to get the table scan type.

rui-mo:
@mbasmanova I appreciate you helping review this change. The above comments were addressed. Could you please take another look?

velox/exec/benchmarks/CMakeLists.txt
```diff
@@ -137,3 +137,12 @@ target_link_libraries(
   velox_vector_fuzzer
   velox_vector_test_lib
   Folly::follybenchmark)
+
+add_executable(velox_duplicate_project_benchmark DuplicateProjectBenchmark.cpp)
```
mbasmanova: To avoid accidental name collisions, I suggest to use file path to make the name

velox_exec_bm_duplicate_project

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -529,6 +532,47 @@ TEST_F(OperatorUtilsTest, projectChildren) {
           srcRowVector->childAt(projection.inputChannel).get());
     }
   }
+
+  {
+    // Test wrapping an unloaded lazy vector in dictionary vector multiple
```
mbasmanova: This method appears to be very long. Would it be possible to put the new test in a separate method?

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -529,6 +532,47 @@ TEST_F(OperatorUtilsTest, projectChildren) {
...
+    auto lazyVector = fuzzer.wrapInLazyVector(fuzzer.fuzzFlat(BIGINT(), size));
+    std::vector<VectorPtr> children = {lazyVector};
+    auto rowVector = std::make_shared<RowVector>(
+        pool(), ROW({"c0"}, {BIGINT()}), nullptr, size, std::move(children));
+
+    std::vector<IdentityProjection> identicalProjections{};
+    identicalProjections.emplace_back(0, 0);
+    identicalProjections.emplace_back(0, 1);
+
+    BufferPtr mapping = allocateIndices(size, pool());
```
mbasmanova: consider using makeIndices helper function from VectorTestBase.h

```
BufferPtr makeIndices(
    vector_size_t size,
    std::function<vector_size_t(vector_size_t)> indexAt,
    memory::MemoryPool* pool);
```

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -529,6 +532,47 @@ TEST_F(OperatorUtilsTest, projectChildren) {
...
     }
   }
+
+  {
+    // Test wrapping an unloaded lazy vector in dictionary vector multiple
+    // times.
+    const auto size = 1'000;
+    auto lazyVector = fuzzer.wrapInLazyVector(fuzzer.fuzzFlat(BIGINT(), size));
+    std::vector<VectorPtr> children = {lazyVector};
+    auto rowVector = std::make_shared<RowVector>(
```
mbasmanova: why not makeRowVector ?

velox/exec/tests/TableScanTest.cpp
mbasmanova: Got it. Would it be possible to update PR description to show simplest SQL query that triggers the issue?

velox/exec/benchmarks/DuplicateProjectBenchmark.cpp
```diff
@@ -0,0 +1,138 @@
...
+using namespace facebook::velox::exec;
+using namespace facebook::velox::test;
+using namespace facebook::velox::exec::test;
+
+namespace {
+
+static constexpr int32_t kNumVectors = 50;
+static constexpr int32_t kRowsPerVector = 10'000;
+
+class DuplicateProjectBenchmark : public HiveConnectorTestBase {
```
mbasmanova, line 39: Curious, what's the motivation for adding this benchmark.

rui-mo: I was trying to fix this issue by preloading the field if it's multiple referenced in identity projections, but as 
https://github.com/facebookincubator/velox/pull/13895#discussion_r2191272890 pointed out, there was some perf regression and this benchmark was added to verify that. After benchmarking, this new fix is adopted to avoid affecting the perf, seeing https://github.com/facebookincubator/velox/pull/13895#discussion_r2222870935.

velox/exec/tests/TableScanTest.cpp
rui-mo: Added the query in the PR description, thanks.

velox/exec/benchmarks/CMakeLists.txt
rui-mo: Updated, thanks.

velox/exec/tests/OperatorUtilsTest.cpp
rui-mo: Added a test `TEST_F(OperatorUtilsTest, projectDuplicateChildren)`.

velox/exec/tests/OperatorUtilsTest.cpp
rui-mo: Thanks for the pointer! Updated.

velox/exec/tests/OperatorUtilsTest.cpp
rui-mo: Updated, thanks!

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -531,6 +531,42 @@ TEST_F(OperatorUtilsTest, projectChildren) {
   }
 }
 
+TEST_F(OperatorUtilsTest, projectDuplicateChildren) {
+  // Test wrapping an unloaded lazy vector in dictionary vector multiple
+  // times.
+  VectorFuzzer::Options options;
+  options.nullRatio = 0.1;
+  VectorFuzzer fuzzer(options, pool());
```
mbasmanova: Do we need to use a fuzzer here? Can we just make a simple vector of 1,2,3,4?

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -531,6 +531,42 @@ TEST_F(OperatorUtilsTest, projectChildren) {
...
+
+  const auto size = 1'000;
```
mbasmanova: Does it need to be 1K rows? Is 10 rows not enough?

velox/exec/tests/OperatorUtilsTest.cpp
```diff
@@ -531,6 +531,42 @@ TEST_F(OperatorUtilsTest, projectChildren) {
...
+  std::vector<VectorPtr> children = {
+      fuzzer.wrapInLazyVector(fuzzer.fuzzFlat(BIGINT(), size))};
+  auto rowVector = makeRowVector(std::move(children));
+
+  std::vector<IdentityProjection> identicalProjections{};
```
mbasmanova: consistency: Identity vs. identical

drop {}

mbasmanova (approved):
@rui-mo Thanks.

velox/exec/tests/OperatorUtilsTest.cpp
rui-mo: I updated to make a vector `{1, std::nullopt, 3, 4, 5}` and creates a lazy vector with it. The use `VectorFuzzer::wrapInLazyVector` was also removed, so we don't need fuzzer here.

velox/exec/tests/OperatorUtilsTest.cpp
rui-mo: Yes, using smaller size can repro this issue. Thanks for this suggestion.

kKPulla:
@rui-mo I see a test failure in the signals - https://github.com/facebookincubator/velox/actions/runs/16654290065/job/47135141462?pr=13895 Can you please fix it and I will update the PR to merge. Thanks!

rui-mo:
@kKPulla Thanks for the pointer. I fixed the test failure on debug build. Now the CI fails due to https://github.com/facebookincubator/velox/issues/14309.

kKPulla:
@rui-mo In the latest run, I see fuzzer failures like this https://github.com/facebookincubator/velox/actions/runs/16692404168/job/47253252340?pr=13895 Can you please take a look? Thanks!

rui-mo:
@kKPulla I'm unable to repro this issue because the fuzzer test fails due to https://github.com/facebookincubator/velox/issues/14221#issuecomment-3155893170 currently.

> https://github.com/facebookincubator/velox/actions/runs/16692404168/job/47253252340?pr=13895 

I opened https://github.com/facebookincubator/velox/issues/14335 to track the above mismatch. Thanks.

PR #13902, opened 2025-06-27
feat: Add prefix key test in IndexLookupJoin to make sure it supports prefix keys

juwentus1234:
Summary:
Add the prefix key test in IndexLookupJoinTest as the first step to verify that IndexLookupJoin Operator is opache to support prefix key indexLookupJoin.

The core logic of handling should be determined by the concrete connector implementation, in this case, the SequenceStorageConnector, and further SequenceStorageIndexDataSource and client library to coomunicate with zippy.

Next step is to write test in SequenceStorageIndexDataSourceTest to find out logic where need to be changed and further more is the library communicate with zippy.

Differential Revision: D77414916

### DIFF
--- a/velox/exec/tests/IndexLookupJoinTest.cpp
+++ b/velox/exec/tests/IndexLookupJoinTest.cpp
@@ -1538,6 +1538,387 @@ TEST_P(IndexLookupJoinTest, inJoinCondition) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


xiaoxmeng (approved):
@juwentus1234 LGTM. Thanks!

PR #13904, opened 2025-06-27
misc: Remove SCOPE_EXIT from `Merge::getOutputFromSource`

duanmeng:
Remove SCOPE_EXIT from `Merge::getOutputFromSource`.

### DIFF
--- a/velox/exec/Merge.cpp
+++ b/velox/exec/Merge.cpp
@@ -229,29 +229,28 @@ RowVectorPtr Merge::getOutputFromSource() {
   VELOX_CHECK_NULL(spillMerger_);
   bool atEnd = false;
   output_ = sourceMerger_->getOutput(sourceBlockingFutures_, atEnd);
-  SCOPE_EXIT {
-    if (!atEnd) {
-      return;
-    }
-
-    finishMergeSourceGroup();
-    if (numStartedSources_ < sources_.size()) {
-      return;
-    }
-
-    if (numSpilledRows_ > 0) {
-      setupSpillMerger();
-      return;
-    }
-    finished_ = true;
-  };
-
   if (needSpill()) {
     spill();
-    return nullptr;
+    VELOX_CHECK_NULL(output_);
   }
 
-  VELOX_CHECK_EQ(numSpilledRows_, 0);
+  if (!atEnd) {
+    return std::move(output_);
+  }
+
+  finishMergeSourceGroup();
+  if (numStartedSources_ < sources_.size()) {
+    VELOX_CHECK_NULL(output_);
+    return std::move(output_);
+  }
+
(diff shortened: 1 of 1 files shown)


velox/exec/Merge.cpp
```diff
@@ -229,28 +229,27 @@ RowVectorPtr Merge::getOutputFromSource() {
...
 
-    if (numSpilledRows_ > 0) {
-      setupSpillMerger();
-      return;
-    }
-    finished_ = true;
-  };
+  finishMergeSourceGroup();
+  if (numStartedSources_ < sources_.size()) {
+    return std::move(output_);
```
xiaoxmeng, line 244:
```
VELOX_CHECK_NULL(output_);
return std::move(output_);
```

velox/exec/Merge.cpp
```diff
@@ -229,28 +229,27 @@ RowVectorPtr Merge::getOutputFromSource() {
...
-  if (needSpill()) {
-    spill();
+  if (numSpilledRows_ > 0) {
+    setupSpillMerger();
+    VELOX_CHECK_NULL(output_);
     return nullptr;
   }
 
+  finished_ = true;
   VELOX_CHECK_EQ(numSpilledRows_, 0);
```
xiaoxmeng: remove this check

velox/exec/Merge.cpp
```diff
@@ -229,29 +229,28 @@ RowVectorPtr Merge::getOutputFromSource() {
   VELOX_CHECK_NULL(spillMerger_);
   bool atEnd = false;
   output_ = sourceMerger_->getOutput(sourceBlockingFutures_, atEnd);
-  SCOPE_EXIT {
```
spershin, line 232: Not sure what this change is about.

But, it is SCOPE_EXIT on line 179 can cause crash when exception is thrown in the function and then scope gets destroyed, calls sourceMerger_->isBlocked(), which hits the same exception again, which triggers SIGABRT.
Kevin found it.

@xiaoxmeng, @duanmeng

spershin (changes requested):

velox/exec/Merge.cpp
xiaoxmeng: similar but https://github.com/facebookincubator/velox/pull/13905 is the fix. This one fixes the spill path which is not enabled yet but only used by pyspark for now.

spershin:
One note: test failed, could be due to this change?
31 - ArbitrationParticipantTest.concurrentAbort (Subprocess aborted)

xiaoxmeng:
> One note: test failed, could be due to this change? 31 - ArbitrationParticipantTest.concurrentAbort (Subprocess aborted)

It doesn't seem related. @tanjialiang could you help take a look? Thanks!

PR #13905, opened 2025-06-27
fix: Fix the server crash caused by remote exchange error

xiaoxmeng:
Summary: The remote exchange might run into error so merge operator check if a remote merge source is blocked or not, it might trigger merge source to fetch data from the exchange client which in turn try to dequeue from the exchange source. The dequeue operation might throw if there is already an error. The recent merge spill support moves the merge source is blocked check inside SCOPE_EXIT which converts an throw to a server crash as we can't throw inside a dtor. This changes move the merge source is blocked check out of SCOPE_EXIT and add a unit test to reproduce and verify

Differential Revision: D77423282

### DIFF
--- a/velox/exec/ExchangeQueue.cpp
+++ b/velox/exec/ExchangeQueue.cpp
@@ -16,6 +16,10 @@
 #include "velox/exec/ExchangeQueue.h"
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


tanjialiang (approved):
Thanks for the fix

duanmeng (approved):

PR #13906, opened 2025-06-27
fix: Fix the result map size setting

xiaoxmeng:
Summary: The result map size is incorrectly set null bytes if the input has nulls. This is discovered by presto unit test which covers AIDN reconstruction udf

Differential Revision: D77424024


duanmeng (approved):
LGTM, thanks.

tanjialiang (approved):

PR #13907, opened 2025-06-27
feat: Fix the full outer join result mismatch issue

JkSelf:
# Problem
Assume the left table has columns a and b:

| a | b   |
|--:|----:|
| 2 | 100 |
| 2 | 1   |
| 2 | 1   |

The right table has columns c and d:

| c | d  |
|--:|---:|
| 2 | 3  |
| 2 | -1 |
| 2 | -1 |
| 2 | 3  |

The two tables are joined using a full outer join on the condition a == c and b < d. During the doGetOutput phase, the result is matched using a left join, resulting in 3 * 4 = 12 records:

| No | a | b   | c | d  |
|---:|--:|----:|--:|---:|
| 0  | 2 | 100 | 2 | 3  |
| 1  | 2 | 100 | 2 | -1 |
| 2  | 2 | 100 | 2 | -1 |
| 3  | 2 | 100 | 2 | 3  |
| 4  | 2 | 1   | 2 | 3  |
| 5  | 2 | 1   | 2 | -1 |
| 6  | 2 | 1   | 2 | -1 |
| 7  | 2 | 1   | 2 | 3  |
| 8  | 2 | 1   | 2 | 3  |
| 9  | 2 | 1   | 2 | -1 |
| 10 | 2 | 1   | 2 | -1 |
| 11 | 2 | 1   | 2 | 3  |

Then, in the filter method, the records are filtered based on the condition a == c and b < d, resulting in the following:

| No | a | b   | c | d  | matched |
|---:|--:|----:|--:|---:|---------|
| 0  | 2 | 100 | 2 | 3  | FALSE   |
| 1  | 2 | 100 | 2 | -1 | FALSE   |
| 2  | 2 | 100 | 2 | -1 | FALSE   |
| 3  | 2 | 100 | 2 | 3  | FALSE   |
| 4  | 2 | 1   | 2 | 3  | TRUE    |
| 5  | 2 | 1   | 2 | -1 | FALSE   |
| 6  | 2 | 1   | 2 | -1 | FALSE   |
| 7  | 2 | 1   | 2 | 3  | TRUE    |
| 8  | 2 | 1   | 2 | 3  | TRUE    |
| 9  | 2 | 1   | 2 | -1 | FALSE   |
| 10 | 2 | 1   | 2 | -1 | FALSE   |
| 11 | 2 | 1   | 2 | 3  | TRUE    |

Finally, records from the left table that do not have a match are filled with nulls, resulting in the following final output:

| No | a | b   | c    | d    |
|---:|--:|----:|------|------|
| 0  | 2 | 100 | null | null |
| 1  | 2 | 1   | 2    | 3    |
| 2  | 2 | 1   | 2    | 3    |
| 3  | 2 | 1   | 2    | 3    |
| 4  | 2 | 1   | 2    | 3    |

The above result is incorrect because it is missing rows from the right table that do not have a match. Among the 12 rows above, rows 0, 4, and 8 correspond to the first record (2, 3) from the right table, rows 1, 5, and 9 correspond to the second record (2, -1) from the right table, rows 2, 6, and 10 correspond to the third record (2, -1) from the right table, and rows 3, 7, and 11 correspond to the fourth record (2, 3) from the right table. From the matching results above, rows 1, 5, and 9, as well as rows 2, 6, and 10, are all false, meaning that the third and fourth records from the right table do not have matching rows. Therefore, the final result is missing rows from the right table that do not have matches. The correct final result should be:

| No | a    | b    | c    | d    |
|---:|------|------|------|------|
| 0  | 2    | 100  | null | null |
| 1  | 2    | 1    | 2    | 3    |
| 2  | 2    | 1    | 2    | 3    |
| 3  | 2    | 1    | 2    | 3    |
| 4  | 2    | 1    | 2    | 3    |
| 5  | null | null | 2    | -1   |
| 6  | null | null | 2    | -1   |

# How to Fix

## Step 1: `doGetOutput` emits regular match rows (Cartesian product 3×4 = 12)

This corresponds to the 12-row table in the PR. At this point:

- `isProvisionalRightRow = 0` for all rows
- `rightRowId` identifies the physical right-side row that produced the output row
- after this change, `rightRowId` is treated as a stable right-row identity, not as a per-match local batch index

For this example there is only one physical right batch, so the 4 right rows map to:

- `(1, 0) -> 4294967296`
- `(1, 1) -> 4294967297`
- `(1, 2) -> 4294967298`
- `(1, 3) -> 4294967299`

| outRow | a | b   | c | d  | rightRowId |
|------:|--:|----:|--:|---:|-----------:|
| 0  | 2 | 100 | 2 | 3  | 4294967296 |
| 1  | 2 | 100 | 2 | -1 | 4294967297 |
| 2  | 2 | 100 | 2 | -1 | 4294967298 |
| 3  | 2 | 100 | 2 | 3  | 4294967299 |
| 4  | 2 | 1   | 2 | 3  | 4294967296 |
| 5  | 2 | 1   | 2 | -1 | 4294967297 |
| 6  | 2 | 1   | 2 | -1 | 4294967298 |
| 7  | 2 | 1   | 2 | 3  | 4294967299 |
| 8  | 2 | 1   | 2 | 3  | 4294967296 |
| 9  | 2 | 1   | 2 | -1 | 4294967297 |
| 10 | 2 | 1   | 2 | -1 | 4294967298 |
| 11 | 2 | 1   | 2 | 3  | 4294967299 |

## Step 2: `doGetOutput` appends provisional right-only rows (one per right row)

After all regular match rows are emitted for this equal-key window, we append one
provisional row for each right row:

- `(a, b)` are set to `NULL`
- `(c, d)` keep the right-side values
- `isProvisionalRightRow = 1`
- `rightRowId` is the same stable identity used by the regular match rows

| outRow | a    | b    | c | d   | rightRowId | isProvisionalRightRow |
|------:|------|------|--:|----:|-----------:|----------------------:|
| 12 | null | null | 2 | 3   | 4294967296 | 1 |
| 13 | null | null | 2 | -1  | 4294967297 | 1 |
| 14 | null | null | 2 | -1  | 4294967298 | 1 |
| 15 | null | null | 2 | 3   | 4294967299 | 1 |

At this point, `doGetOutput` has produced 16 rows total:

- 12 regular match rows
- 4 provisional right-only rows

## Step 3: `applyFilter` evaluates only regular match rows

Predicate:

```sql
a = c AND b < d
```

Within this equal-key window, `a = c` is already satisfied, so the effective check is
just `b < d`.

`applyFilter()` evaluates only the regular match rows. The provisional right-only rows
(12–15) are not included in `filterRows`, so they are not evaluated by the filter.

| outRow | a | b   | c | d   | rightRowId | isProvisionalRightRow | matched (`b < d`) |
|------:|--:|----:|--:|----:|-----------:|----------------------:|------------------:|
| 0  | 2 | 100 | 2 | 3   | 4294967296 | 0 | F |
| 1  | 2 | 100 | 2 | -1  | 4294967297 | 0 | F |
| 2  | 2 | 100 | 2 | -1  | 4294967298 | 0 | F |
| 3  | 2 | 100 | 2 | 3   | 4294967299 | 0 | F |
| 4  | 2 | 1   | 2 | 3   | 4294967296 | 0 | T |
| 5  | 2 | 1   | 2 | -1  | 4294967297 | 0 | F |
| 6  | 2 | 1   | 2 | -1  | 4294967298 | 0 | F |
| 7  | 2 | 1   | 2 | 3   | 4294967299 | 0 | T |
| 8  | 2 | 1   | 2 | 3   | 4294967296 | 0 | T |
| 9  | 2 | 1   | 2 | -1  | 4294967297 | 0 | F |
| 10 | 2 | 1   | 2 | -1  | 4294967298 | 0 | F |
| 11 | 2 | 1   | 2 | 3   | 4294967299 | 0 | T |

## Step 4: Build `matchedRightRowIds_`

We scan all regular match rows where the filter passed and collect their `rightRowId`.

- rows with `matched = true`: `4, 7, 8, 11`
- corresponding `rightRowId`: `4294967296, 4294967299, 4294967296, 4294967299`

So:

| setName | values |
|---------|--------|
| `matchedRightRowIds_` | `{ 4294967296, 4294967299 }` |

Interpretation:

| rightRowId | right row `(c, d)` | has any passing match? |
|-----------:|--------------------|-----------------------:|
| 4294967296 | `(2, 3)`  | Yes |
| 4294967297 | `(2, -1)` | No  |
| 4294967298 | `(2, -1)` | No  |
| 4294967299 | `(2, 3)`  | Yes |

This is the key idea of the fix:

- keep a stable identity for each physical right row
- record which right rows have at least one passing match
- later drop only provisional rows whose `rightRowId` is in `matchedRightRowIds_`

## Step 5: Left-side fill

If all regular match rows for one left row fail the filter, we keep one row and null
out the right-side columns.

Left row `(2, 100)` corresponds to outRows `0..3`, and all of them are `F`, so we keep:

| output row | a | b   | c    | d    |
|------------|--:|----:|------|------|
| left-fill  | 2 | 100 | null | null |

Left rows `(2, 1)` have passing matches, so we keep all passing rows `(2, 1, 2, 3)`.

If we only applied the left-side rule, the kept rows would be:

| No | a | b   | c    | d    |
|---:|--:|----:|------|------|
| 0 | 2 | 100 | null | null |
| 1 | 2 | 1   | 2    | 3    |
| 2 | 2 | 1   | 2    | 3    |
| 3 | 2 | 1   | 2    | 3    |
| 4 | 2 | 1   | 2    | 3    |

This is the incorrect 5-row result because it is still missing the right-side rows that
had no passing match.

## Step 6: Decide which provisional right-only rows should be kept

Rule for provisional rows:

- if `isProvisionalRightRow = 1` and `rightRowId ∈ matchedRightRowIds_` => drop
- otherwise => keep

Decision for outRows `12..15`:

| outRow | a    | b    | c | d   | rightRowId | isProvisionalRightRow | in `matchedRightRowIds_`? | keep? |
|------:|------|------|--:|----:|-----------:|----------------------:|--------------------------:|------:|
| 12 | null | null | 2 | 3   | 4294967296 | 1 | Yes | No  |
| 13 | null | null | 2 | -1  | 4294967297 | 1 | No  | Yes |
| 14 | null | null | 2 | -1  | 4294967298 | 1 | No  | Yes |
| 15 | null | null | 2 | 3   | 4294967299 | 1 | Yes | No  |

So we keep the provisional rows for `4294967297` and `4294967298`, i.e. the two
`(2, -1)` rows.

## Step 7: Final output (correct result)

Kept rows:

- left fill: `(2, 100, NULL, NULL)` × 1
- passing matches: `(2, 1, 2, 3)` × 4
- right fill: `(NULL, NULL, 2, -1)` × 2

One possible final ordering:

| No | a    | b    | c    | d    |
|---:|------|------|------|------|
| 0 | 2    | 100  | null | null |
| 1 | 2    | 1    | 2    | 3    |
| 2 | 2    | 1    | 2    | 3    |
| 3 | 2    | 1    | 2    | 3    |
| 4 | 2    | 1    | 2    | 3    |
| 5 | null | null | 2    | -1   |
| 6 | null | null | 2    | -1   |

## Additional note on the current fix

In addition to using a stable `rightRowId`, this change also updates the
`filterInput_` aliases when right projections are flattened. This ensures that
filter evaluation always reads the current right-side values even when output
crosses right-batch boundaries.

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -282,6 +282,7 @@ int32_t MergeJoin::compare(
 bool MergeJoin::findEndOfMatch(
     const RowVectorPtr& input,
     const std::vector<column_index_t>& keys,
+    uint64_t inputBatchId,
     Match& match) {
   if (match.complete) {
     return true;
@@ -299,6 +300,7 @@ bool MergeJoin::findEndOfMatch(
 
   if (endRow == numInputRows) {
     match.inputs.push_back(input);
+    match.inputBatchIds.push_back(inputBatchId);
     match.endRowIndex = endRow;
     return false;
   }
@@ -306,6 +308,7 @@ bool MergeJoin::findEndOfMatch(
   if (endRow > 0) {
     // Match ends here, no need to pre-load lazies.
     match.inputs.push_back(input);
+    match.inputBatchIds.push_back(inputBatchId);
     match.endRowIndex = endRow;
   }
   match.complete = true;
@@ -346,7 +349,7 @@ bool MergeJoin::tryAddOutputRowForLeftJoin() {
   rawLeftOutputIndices_[outputSize_] = leftRowIndex_++;
   if (rawRightRowIds_ != nullptr) {
     rawRightRowIds_[outputSize_] = 0;
-    rawRightRowIsCandidate_[outputSize_] = 0;
+    rawIsProvisionalRightRow_[outputSize_] = 0;
   }
 
   for (const auto& projection : rightProjections_) {
@@ -376,7 +379,7 @@ bool MergeJoin::tryAddOutputRowForRightJoin() {
 
   if (rawRightRowIds_ != nullptr) {
     rawRightRowIds_[outputSize_] = 0;
-    rawRightRowIsCandidate_[outputSize_] = 0;
+    rawIsProvisionalRightRow_[outputSize_] = 0;
   }
 
   if (!isRightFlattened_) {
@@ -417,7 +420,20 @@ void MergeJoin::flattenRightProjections() {
     newFlat->copy(currentVector.get(), 0, 0, outputSize_);
     children[projection.outputChannel] = std::move(newFlat);
   }
+
+  if (filterInput_ != nullptr) {
+    for (const auto& [filterChannel, outputChannel] :
+         filterInputToOutputChannel_) {
+      filterInput_->childAt(filterChannel) = children[outputChannel];
+    }
+  }
   isRightFlattened_ = true;
+}
+
+uint64_t MergeJoin::rightRowId(
+    uint64_t rightBatchId,
+    vector_size_t rightRowIndex) const {
+  return (rightBatchId << 32) | static_cast<uint32_t>(rightRowIndex);
 }
 
 bool MergeJoin::tryAddOutputRow(
@@ -425,16 +441,14 @@ bool MergeJoin::tryAddOutputRow(
     vector_size_t leftRow,
     const RowVectorPtr& rightBatch,
     vector_size_t rightRow,
-    size_t rightBatchIndex) {
+    uint64_t rightBatchId) {
   if (outputSize_ == outputBatchSize_) {
     return false;
   }
 
   if (rawRightRowIds_ != nullptr) {
-    rawRightRowIds_[outputSize_] =
-        (static_cast<uint64_t>(rightBatchIndex) << 32) |
-        static_cast<uint32_t>(rightRow);
-    rawRightRowIsCandidate_[outputSize_] = 0;
+    rawRightRowIds_[outputSize_] = rightRowId(rightBatchId, rightRow);
+    rawIsProvisionalRightRow_[outputSize_] = 0;
   }
 
   // All left side projections share the same dictionary indices (leftIndices_).
@@ -570,14 +584,14 @@ bool MergeJoin::prepareOutput(
   if (isFullJoin(joinType_) && filter_) {
     rightRowIds_ = AlignedBuffer::allocate<uint64_t>(outputBatchSize_, pool());
     rawRightRowIds_ = rightRowIds_->asMutable<uint64_t>();
-    rightRowIsCandidate_ =
-        AlignedBuffer::allocate<uint64_t>(outputBatchSize_, pool());
-    rawRightRowIsCandidate_ = rightRowIsCandidate_->asMutable<uint64_t>();
+    isProvisionalRightRow_ =
+        AlignedBuffer::allocate<uint8_t>(outputBatchSize_, pool());
+    rawIsProvisionalRightRow_ = isProvisionalRightRow_->asMutable<uint8_t>();
   } else {
     rightRowIds_.reset();
     rawRightRowIds_ = nullptr;
-    rightRowIsCandidate_.reset();
-    rawRightRowIsCandidate_ = nullptr;
+    isProvisionalRightRow_.reset();
+    rawIsProvisionalRightRow_ = nullptr;
   }
 
   if (filterInput_ != nullptr) {
@@ -638,11 +652,12 @@ bool MergeJoin::addToOutput() {
 template <bool IsLeftJoin>
 bool MergeJoin::addToOutputImpl() {
   if constexpr (IsLeftJoin) {
-    if (isFullJoin(joinType_) && filter_ && rightCandidateCursor_.has_value()) {
-      if (appendRightCandidates()) {
+    if (isFullJoin(joinType_) && filter_ &&
+        provisionalRightOutputCursor_.has_value()) {
+      if (appendProvisionalRightRows()) {
         return true;
       }
-      passedRightRows_.clear();
+      matchedRightRowIds_.clear();
       leftMatch_.reset();
       rightMatch_.reset();
       return outputSize_ == outputBatchSize_;
@@ -723,6 +738,7 @@ bool MergeJoin::addToOutputImpl() {
         }
 
         if (IsLeftJoin) {
+          VELOX_CHECK_LT(innerBatchIndex, innerMatch->inputBatchIds.size());
           for (auto innerRow = innerStartRow; innerRow < innerEndRow;
                ++innerRow) {
             if (!tryAddOutputRow(
@@ -730,7 +746,7 @@ bool MergeJoin::addToOutputImpl() {
                     outerRow,
                     rightBatch,
                     innerRow,
-                    innerBatchIndex)) {
+                    innerMatch->inputBatchIds[innerBatchIndex])) {
               outerMatch->setCursor(outerBatchIndex, outerRow);
               innerMatch->setCursor(innerBatchIndex, innerRow);
               return true;
@@ -740,7 +756,11 @@ bool MergeJoin::addToOutputImpl() {
           for (auto innerRow = innerStartRow; innerRow < innerEndRow;
                ++innerRow) {
             if (!tryAddOutputRow(
-                    leftBatch, innerRow, rightBatch, outerRow, 0)) {
+                    leftBatch,
+                    innerRow,
+                    rightBatch,
+                    outerRow,
+                    0 /* rightBatchId */)) {
               outerMatch->setCursor(outerBatchIndex, outerRow);
               innerMatch->setCursor(innerBatchIndex, innerRow);
               return true;
@@ -753,10 +773,10 @@ bool MergeJoin::addToOutputImpl() {
 
   if constexpr (IsLeftJoin) {
     if (isFullJoin(joinType_) && filter_) {
-      if (appendRightCandidates()) {
+      if (appendProvisionalRightRows()) {
         return true;
       }
-      passedRightRows_.clear();
+      matchedRightRowIds_.clear();
     }
   }
 
@@ -765,7 +785,39 @@ bool MergeJoin::addToOutputImpl() {
   return outputSize_ == outputBatchSize_;
 }
 
-bool MergeJoin::appendRightCandidates() {
+void MergeJoin::appendProvisionalRightRow(
+    const RowVectorPtr& rightBatch,
+    vector_size_t rightRowIndex,
+    uint64_t rightRowId) {
+  if (!isRightFlattened_) {
+    rawRightOutputIndices_[outputSize_] = rightRowIndex;
+  } else {
+    copyRow(rightBatch, rightRowIndex, output_, outputSize_, rightProjections_);
+  }
+
+  rawLeftOutputIndices_[outputSize_] = 0;
+  for (const auto& projection : leftProjections_) {
+    auto& target = output_->childAt(projection.outputChannel);
+    addNull(
+        target,
+        outputSize_,
+        leftOutputIndices_,
+        outputBatchSize_,
+        currentLeft_);
+  }
+
+  if (joinTracker_) {
+    joinTracker_->addMiss(outputSize_);
+  }
+
+  VELOX_CHECK_NOT_NULL(rawRightRowIds_);
+  rawRightRowIds_[outputSize_] = rightRowId;
+  rawIsProvisionalRightRow_[outputSize_] = 1;
+
+  ++outputSize_;
+}
+
+bool MergeJoin::appendProvisionalRightRows() {
   VELOX_CHECK(isFullJoin(joinType_));
   VELOX_CHECK_NOT_NULL(filter_);
   VELOX_CHECK(rightMatch_);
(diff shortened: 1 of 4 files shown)


JkSelf:
@xiaoxmeng Could you help to review this PR? Thanks for your help.

velox/exec/fuzzer/JoinFuzzer.cpp
```diff
@@ -63,7 +63,7 @@ DEFINE_bool(
 
 DEFINE_double(
     filter_ratio,
-    0,
+    0.5,
```
shrinidhijoshi, line 66: Is this needed ?

velox/exec/MergeJoin.h
```diff
@@ -562,6 +572,11 @@ class MergeJoin : public Operator {
   vector_size_t* rawLeftOutputIndices_;
   vector_size_t* rawRightOutputIndices_;
 
+  BufferPtr rightRowIds_;
+  uint64_t* rawRightRowIds_{nullptr};
+  BufferPtr rightRowIsCandidate_;
+  uint64_t* rawRightRowIsCandidate_{nullptr};
```
shrinidhijoshi: s/rightRowisCandidate_/isProvisionalRightRow_

velox/exec/MergeJoin.h
```diff
@@ -260,7 +262,15 @@ class MergeJoin : public Operator {
       const RowVectorPtr& leftBatch,
       vector_size_t leftRow,
       const RowVectorPtr& rightBatch,
-      vector_size_t rightRow);
+      vector_size_t rightRow,
+      size_t rightBatchIndex);
+
+  struct RightCandidateCursor {
```
shrinidhijoshi: s/RightCandidateCursor/ProvisionalRightOutputCursor

velox/exec/MergeJoin.h
```diff
@@ -644,6 +659,8 @@ class MergeJoin : public Operator {
 
   // A set of rows with matching keys on the right side.
   std::optional<Match> rightMatch_;
+  std::optional<RightCandidateCursor> rightCandidateCursor_;
```
shrinidhijoshi: s/rightCandidateCursor_/provisionalRightOutputCursor_

velox/exec/MergeJoin.h
```diff
@@ -644,6 +659,8 @@ class MergeJoin : public Operator {
...
+  folly::F14FastSet<uint64_t> passedRightRows_;
```
shrinidhijoshi: s/passedRightRows_/matchedRightRows_

velox/exec/MergeJoin.cpp
```diff
@@ -699,11 +751,99 @@ bool MergeJoin::addToOutputImpl() {
...
+      passedRightRows_.clear();
+    }
+  }
+
   leftMatch_.reset();
   rightMatch_.reset();
   return outputSize_ == outputBatchSize_;
 }
 
+bool MergeJoin::appendRightCandidates() {
```
shrinidhijoshi: s/appendRightCandidates/appendProvisionalRightRows

velox/exec/MergeJoin.cpp
```diff
@@ -688,7 +739,8 @@ bool MergeJoin::addToOutputImpl() {
         } else {
           for (auto innerRow = innerStartRow; innerRow < innerEndRow;
                ++innerRow) {
-            if (!tryAddOutputRow(leftBatch, innerRow, rightBatch, outerRow)) {
+            if (!tryAddOutputRow(
+                    leftBatch, innerRow, rightBatch, outerRow, 0)) {
```
shrinidhijoshi:
```C++
 tryAddOutputRow(
      leftBatch,
      innerRow,
      rightBatch,
      outerRow,
      0 /* rightBatchIndex */);
```

velox/exec/MergeJoin.h
```diff
@@ -260,7 +262,15 @@ class MergeJoin : public Operator {
...
-      vector_size_t rightRow);
+      vector_size_t rightRow,
+      size_t rightBatchIndex);
+
+  struct RightCandidateCursor {
+    size_t rightBatchIndex{0};
+    vector_size_t rightRowIndex{0};
+  };
+
+  bool appendRightCandidates();
```
shrinidhijoshi: Please add some function comments

velox/exec/MergeJoin.cpp
shrinidhijoshi: Also, let's extract some common/helper logic as smaller functions ?

```C++
// common logic for constructing rightRowId 
 uint64_t MergeJoin::rightRowId(
  }
```

```C++
// logic for appending a single right row
 void MergeJoin::appendProvisionalRightRow(
      const RowVectorPtr& rightBatch,
      vector_size_t rightRowIndex,
      uint64_t rightRowId)
```

velox/exec/MergeJoin.h
shrinidhijoshi: Do we need `uint64_t` here ? Can we get away with `uint8_t` or `bool` ?

velox/exec/MergeJoin.cpp
```diff
@@ -411,11 +424,19 @@ bool MergeJoin::tryAddOutputRow(
...
-    vector_size_t rightRow) {
+    vector_size_t rightRow,
+    size_t rightBatchIndex) {
   if (outputSize_ == outputBatchSize_) {
     return false;
   }
 
+  if (rawRightRowIds_ != nullptr) {
+    rawRightRowIds_[outputSize_] =
+        (static_cast<uint64_t>(rightBatchIndex) << 32) |
```
shrinidhijoshi: Using rightBatchIndex means that if 2 keys in 2 different batches have index 0, and the batches are small enough that they both fit in single output batch, then their rightRowIds will collide and produce incorrect result.

Can you add a unit test to confirm/solve this scenario .. 

```C++
    // Both physical right batches have a row at local index zero.
  auto rightBatch1 = makeRowVector(
      {"c", "d"},
      {
          makeFlatVector<int32_t>({1, 1, 1, 1}),
          makeFlatVector<int32_t>({10, 10, 10, 10}),
      });

  auto rightBatch2 = makeRowVector(
      {"c", "d"},
      {
          makeFlatVector<int32_t>({2, 2, 2, 2}),
          makeFlatVector<int32_t>({0, 0, 0, 0}),
      });
```

shrinidhijoshi (commented):
@JkSelf Thanks for identifying the bug and submitting the fix. Overall approach seems  reasonable.

We might have to enhance the rightRowId generation logic as current approach has possibility of collision in some cases.

Thanks!

velox/exec/MergeJoin.cpp
JkSelf: @shrinidhijoshi Good catch, thanks. I added a unit test for this exact cross-batch collision scenario.

The root cause was that rightRowId used the per-match local rightBatchIndex , which is not globally unique across physical right batches. I fixed this by assigning a stable id to each physical right batch and encoding rightRowId from (rightBatchId, rowIndex) .

velox/exec/MergeJoin.h
JkSelf: @shrinidhijoshi Yes. I have updated `uint64_t`to `uint8_t`.

velox/exec/fuzzer/JoinFuzzer.cpp
JkSelf: @shrinidhijoshi  This is unrelated to the main fix. The 0.5 value is used to enable filter expressions in JoinFuzzerTest. Since this is a small update, I included it in the same PR.

JkSelf:
@shrinidhijoshi Thanks for your comments. I have resolved and could you help to review again? Thanks.

JkSelf:
@shrinidhijoshi Could you help to review again? Thanks for your help.

shrinidhijoshi:
LGTM! Thanks for addressing the comments.

We can merge this now to close correctness gap.

A good (not needed in this PR) follow up would be get the filter eval closer to the Cartesian prod generation and just use bitmap for tracking right row matches

Makes it closer to HashJoin semantics and easier to reason about, then we can skip provisional row generation.

shrinidhijoshi:
@JkSelf Can you also rebase ? Thanks!

JkSelf:
> A good (not needed in this PR) follow up would be get the filter eval closer to the Cartesian prod generation and just use bitmap for tracking right row matches
> 
> Makes it closer to HashJoin semantics and easier to reason about, then we can skip provisional row generation.

@shrinidhijoshi Thanks, that makes sense. Moving filter evaluation closer to the match-row generation and tracking right-row matches via a bitmap would simplify the logic. I’ll follow up with further optimizations after this PR is merged.

shrinidhijoshi (approved):

JkSelf:
@shrinidhijoshi Thanks for your review and landing.

PR #13913, opened 2025-06-27
docs: Fix segfault dependency blog date

majetideepak:
Fix date from 2025-07-07 to 2025-06-26


majetideepak:
@kagamiori can you please help import this? Thanks.

majetideepak:
This is not required anymore. The previous date now matches.

Issue #13914, opened 2025-06-27
Support user level authentication in velox filesystem


jaystarshot:
### Description

Currently Velox does not support user level authentication in the hive connector.

If we want to add support there are 2 places we need to change

1. FileHandleCache -  Filehandles are cached here and they are cached by file path. This cache needs to have user info. (Link `[here](https://github.com/facebookincubator/velox/blob/5d477166b82e88240debfa0b688fa29af9854570/velox/connectors/hive/FileHandle.cpp#L61)`)

2. Support FileSystem cache or a Client cache per filesystem. 
First proposal is to have multiple filesystem per user. (Like done [here](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.cpp#L45))

Second is to have a client cache inside filesystem and let the filesystem instance create read and write files based on the user info (basically return different ReadFile based on client cache [here](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp#L91))

jaystarshot:
https://github.com/facebookincubator/velox/issues/13875 talks about 2.1

jaystarshot:
cc: @majetideepak

Yuhta:
Auth token should not be part of file system object.  It should be an object (`TokenProvider`) live with QueryCtx and passed down to file system when createing/opening new files.

jaystarshot:
Thank you! So, just to confirm, we should go with option 2.2 and let the filesystem handle how ReadFiles etc are created, correct?

Yuhta:
@jaystarshot Yes, it's not necessarily a cache implementation though, a well designed multi-tenant file system should have a single client to handle all users correctly.

jaystarshot:
Thanks! That’s what I would have expected as well, but from my initial look at the GCS C++ client, it seems like you need to create a new client for each token, which is a bit surprising. I need to double-check to confirm.

majetideepak:
We are limited to what the S3 and GCS SDKs provide. Yes, you have to create a new client per user/token for a file system instance on S3/GCS. For ABFS, you create a client per ReadFile/WriteFile instance

majetideepak:
You also want to handle token refresh. One idea is to allow users to customize the identity key generation used to cache the file-system instance based on their need.

jaystarshot:
[...]

Issue #13875, opened 2025-06-25
Pass username to filesystems for impersonation


wraymo:
### Description

Currently, Velox interacts with filesystems with only the file path. For certain filesystems like GCS and ABFS, a single instance is used, while for others like S3 and HDFS, instances are cached based on the file path (bucket or `hdfsIdentity`). Similarly, file handles are cached using the file path as the key.

The username is already available within the `sessionProperties` of `ConnectorQueryCtx`. We need a mechanism to propagate this username to the underlying filesystem implementations so that Velox can perform operations of that user.

wraymo:
### Problem Breakdown
This problem can be divided into two main parts:

1. How to pass the username to the filesystem?
2. How to make filesystem operations user-aware?

#### Passing Username to the Filesystem

Since file handles are cached, the caching mechanism must be made user-aware. Otherwise, when a query is submitted by a different user, it might incorrectly reuse a cached handle associated with a different user.

A straightforward solution is to include the username in the cache key:

```cpp
namespace facebook::velox {
struct FileHandleCacheKey {
  std::string path;
  std::string user;

  bool operator==(const FileHandleCacheKey& other) const {
    return path == other.path && user == other.user;
  }
};
} // namespace facebook::velox

namespace std {
template <>
struct hash<facebook::velox::FileHandleCacheKey> {
  size_t operator()(const facebook::velox::FileHandleCacheKey& key) const noexcept {
    return folly::hash::hash_combine(key.path, key.user);
  }
};
} // namespace std
```
This ensures that the file handle cache differentiates entries based on both path and user.

Then in the [FileHandle generator](https://github.com/facebookincubator/velox/blob/5d477166b82e88240debfa0b688fa29af9854570/velox/connectors/hive/FileHandle.cpp#L42), this key can be constructed and passed down to the underlying filesystem. It's also possible to apply some encoding and still use `std::string` as a key.

#### Making Filesystem Operations User-Aware

There are two main approaches to propagating the username:

##### Option A: Pass Username to `getFileSystem`

We can either:

1. Introduce a new `username` parameter (less desirable due to API impact), or
2. Embed the `username` into existing parameters like `filePath` or `properties`.

Since `getFileSystem` is used extensively, embedding the username into existing parameters is preferred to avoid changing function signatures.

[...]

PR #13919, opened 2025-06-28
feat: Add TokenProvider to ConnectorQueryCtx and include it in the cache key for FileHandle

wraymo:
Summary
As discussed in #13875 and #13914, we need to provide user information to the filesystem to support authentication and impersonation. This requirement can be generalized with a TokenProvider abstraction.

This PR introduces an abstract `TokenProvide`r class and integrates it into both `ConnectorQueryCtx` and `FileOptions`. It also becomes part of the `FileHandle` cache key.

To use it, users simply need to implement the TokenProvider interface with hash, equals, and getToken methods. The filesystem can then access relevant user or token information when reading from or writing to files.

### DIFF
--- a/velox/common/file/FileSystems.h
+++ b/velox/common/file/FileSystems.h
@@ -17,6 +17,7 @@
 
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/base/RuntimeMetrics.h"
+#include "velox/common/file/TokenProvider.h"
 #include "velox/common/memory/MemoryPool.h"
 
 #include <functional>
@@ -84,6 +85,9 @@ struct FileOptions {
   /// A hint to the file system for which region size of the file should be
   /// read. Specifically, the read length.
   std::optional<int64_t> readRangeHint{std::nullopt};
+
+  /// A token provider that can be used to get tokens for accessing the file.
+  std::shared_ptr<TokenProvider> tokenProvider{nullptr};
 };
 
 /// Defines directory options
--- a/velox/common/file/TokenProvider.h
+++ b/velox/common/file/TokenProvider.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
+
+namespace facebook::velox::filesystems {
+
+/// Identifier for the file systems to implement to differentiate different
+/// tokens needed in the same query (user).  Such information usually needs to
+/// be passed down and stored in the ReadFile/WriteFile object of the specific
+/// file system.
+class AccessTokenKey {
+  virtual ~AccessTokenKey() = default;
+};
+
+/// Abstract token each file system can implement and cast.
+class AccessToken {
+  virtual ~AccessToken() = default;
+};
+
+/// Interface for providing access tokens to file systems.
+class TokenProvider {
+ public:
+  virtual ~TokenProvider() = default;
+
+  virtual bool equals(const TokenProvider& other) const = 0;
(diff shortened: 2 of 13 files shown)


velox/connectors/hive/FileHandle.h
```diff
@@ -59,7 +60,46 @@ struct FileHandleSizer {
...
+namespace std {
+template <>
+struct hash<facebook::velox::FileHandleKey> {
+  size_t operator()(const facebook::velox::FileHandleKey& key) const noexcept {
+    size_t filenameHash = std::hash<std::string>()(key.filename);
+    size_t tokenProviderHash = 0;
+    if (key.tokenProvider) {
+      tokenProviderHash = key.tokenProvider->hash();
+    }
+    return filenameHash ^ (tokenProviderHash << 1);
```
Yuhta: Let's use `bits::hashMix` here

velox/connectors/hive/iceberg/PositionalDeleteFileReader.cpp
```diff
@@ -92,8 +92,11 @@ PositionalDeleteFileReader::PositionalDeleteFileReader(
       /*tableParameters=*/{},
       deleteReaderOpts);
 
-  auto deleteFileHandleCachePtr =
-      fileHandleFactory_->generate(deleteFile_.filePath);
+  FileHandleCachedPtr fileHandleCachePtr;
```
Yuhta: This is not used

velox/common/file/TokenProvider.h
```diff
@@ -0,0 +1,33 @@
...
+namespace facebook::velox::filesystems {
+
+class TokenProvider {
+ public:
+  virtual ~TokenProvider() = default;
+
+  virtual bool equals(const TokenProvider& other) const = 0;
+  virtual size_t hash() const = 0;
+
+  virtual std::string getToken(const std::string& key) const = 0;
```
Yuhta: Token is not necessarily a string.  @zzhao0 Can our tokens be serialized and deserialized quickly?

Yuhta: After discussion, let's make it more flexible:
```c++
/// Identifier for the file systems to implement to differentiate different tokens needed in the same query (user).  Such information usually needs to be passed down and stored in the ReadFile/WriteFile object of the specific file system.
class AccessTokenKey {
  virtual ~AccessTokenKey() = default;
};

/// Abstract token each file system can implement and cast.
class AccessToken {
  virtual ~AccessToken() = default;
};

class TokenProvider {
  virtual std::shared_ptr<AccessToken> getToken(const AccessTokenKey& key) const = 0;
};
```

Yuhta (approved):

velox/connectors/Connector.h
```diff
@@ -511,6 +514,10 @@ class ConnectorQueryCtx {
     selectiveNimbleReaderEnabled_ = value;
   }
 
+  std::shared_ptr<filesystems::TokenProvider> tokenProvider() const {
```
xiaoxmeng: Can we rename it to fsTokenProvider to make it clear this is for fs access?

velox/connectors/Connector.h
```diff
@@ -528,6 +535,7 @@ class ConnectorQueryCtx {
   const bool adjustTimestampToTimezone_;
   const folly::CancellationToken cancellationToken_;
   bool selectiveNimbleReaderEnabled_{false};
+  const std::shared_ptr<filesystems::TokenProvider> tokenProvider_;
```
xiaoxmeng: Put const members first?

velox/common/file/FileSystems.h
```diff
@@ -84,6 +85,9 @@ struct FileOptions {
   /// A hint to the file system for which region size of the file should be
   /// read. Specifically, the read length.
   std::optional<int64_t> readRangeHint{std::nullopt};
+
+  /// A token provider that can be used to get tokens for accessing the file.
+  std::shared_ptr<filesystems::TokenProvider> tokenProvider{nullptr};
```
xiaoxmeng: filesystems prefix is not required here?

velox/connectors/hive/iceberg/PositionalDeleteFileReader.cpp
```diff
@@ -92,8 +92,10 @@ PositionalDeleteFileReader::PositionalDeleteFileReader(
       /*tableParameters=*/{},
       deleteReaderOpts);
 
-  auto deleteFileHandleCachePtr =
-      fileHandleFactory_->generate(deleteFile_.filePath);
+  FileHandleKey const fileHandleKey{
```
xiaoxmeng:
```
const FileHandleKey fileHandleKey
```

xiaoxmeng (commented):
@wraymo LGTM % nits. Thanks!

PR #13940, opened 2025-06-30
feat: Support comparing MapVectors and FlatMapVectors

kevinwilfong:
Summary:
Given that MapVectors and FlatMapVectors have the same Type, we may find ourselves
needing to compare maps in Vectors of each encoding. E.g. this comes up frequently in
testing.

This diff adds support for this to FlatMapVector's compare function. For the sake of
consolidation MapVector simply flips the order and invokes FlatMapVector's compare 
function.

Differential Revision: D77563385

### DIFF
--- a/velox/vector/ComplexVector.cpp
+++ b/velox/vector/ComplexVector.cpp
@@ -22,6 +22,7 @@
 #include "velox/vector/BaseVector.h"
 #include "velox/vector/ComplexVector.h"
 #include "velox/vector/DecodedVector.h"
+#include "velox/vector/FlatMapVector.h"
 #include "velox/vector/LazyVector.h"
 #include "velox/vector/SimpleVector.h"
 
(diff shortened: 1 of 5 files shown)


velox/vector/ComplexVector.cpp
```diff
@@ -1308,41 +1309,51 @@ std::optional<int32_t> MapVector::compare(
...
+    return compareArrays(
+        *values_, *otherMap->values_, leftIndices, rightIndices, flags);
+  } else if (otherValue->encoding() == VectorEncoding::Simple::FLAT_MAP) {
+    auto otherFlatMap = otherValue->as<FlatMapVector>();
+
+    // Reverse the order and compare the flat map to the map, this way we can
+    // reuse the implementation in FlatMapVector.
+    CompareFlags reversedFlags = flags;
+    reversedFlags.ascending = !reversedFlags.ascending;
+    return otherFlatMap->compare(this, wrappedOtherIndex, index, reversedFlags);
```
pedroerp: nit: we could provide a helper function:
```
return otherFlatMap->compare(this, wrappedOtherIndex, index, reverseFlags(flags));
```

velox/vector/FlatMapVector.cpp
```diff
@@ -313,80 +313,144 @@ std::optional<int32_t> FlatMapVector::compare(
...
+          return valueResult;
+        }
+      }
+    }
+
+    // If one map was smaller than the other.
     int result = leftIndices.size() - rightIndices.size();
     return flags.ascending ? result : result * -1;
-  }
+  } else if (otherValue->encoding() == VectorEncoding::Simple::MAP) {
```
pedroerp: nit: perhaps we should pull these comparisons as a separate helper function so they are more readable (compareToMap()  / compareToFlatMap())

velox/vector/FlatMapVector.cpp
```diff
@@ -313,80 +313,144 @@ std::optional<int32_t> FlatMapVector::compare(
...
+    }
+
+    // We first get sorted key indices for both maps.
+    auto leftIndices = sortedKeyIndices(index);
+    auto rightIndices = otherMap->sortedKeyIndices(wrappedOtherIndex);
+
+    // If equalsOnly and maps have different sizes, we can bail fast.
+    if (flags.equalsOnly && leftIndices.size() != rightIndices.size()) {
+      int result = leftIndices.size() - rightIndices.size();
+      LOG(ERROR) << "Compare of maps of different sizes: " << leftIndices.size()
```
pedroerp: wouldn't this be a legit comparison if equalsOnly is false? Why the error log message?

velox/vector/FlatMapVector.cpp
```diff
@@ -313,80 +313,144 @@ std::optional<int32_t> FlatMapVector::compare(
...
+                 << " " << rightIndices.size();
+      for (int index : leftIndices) {
+        LOG(ERROR) << distinctKeys_->toString(index);
```
pedroerp: same. Assuming this is leftover from debugging?

kevinwilfong: Sorry, ya these are both left over debugging stuff.

pedroerp (approved):
Sorry for the delay in getting back at this. Looks good to me. Thanks!

Issue #13941, opened 2025-06-30
Window Fuzzer with Presto as source of truth fails on TDigest


mbasmanova:
### Bug description

https://github.com/facebookincubator/velox/actions/runs/15984885549/job/45089189293?pr=13938

```
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (8192 vs. 0.001)
Retriable: False
Expression: std::abs(weightSoFar - totalWeight) < kEpsilon
Context: Operator: Window[2] 2
Function: mergeImpl
File: /__w/velox/velox/velox/./velox/functions/lib/TDigest.h
Line: 341
```

CC: @kagamiori @amitkdutta @bikramSingh91 

### System information

n/a

### Relevant logs

```bash

```

kagamiori:
It throws from TDigestAggregate::extractCommon according to the stack trace.
cc @natashasehgal

natashasehgal:
This was closed in https://github.com/facebookincubator/velox/pull/13945

PR #13945, opened 2025-07-01
fix: TDigest Window Fuzzer

natashasehgal:
Issue: https://github.com/facebookincubator/velox/issues/13941#issuecomment-3021189846
Fix TDigest Window Fuzzer with Presto as source of truth 

Differential Revision: D77564909

### DIFF
--- a/velox/functions/lib/TDigest.h
+++ b/velox/functions/lib/TDigest.h
@@ -338,7 +338,11 @@ void TDigest<A>::mergeImpl(
   }
   weightSoFar += weights_[j];
   ++numMerged_;
-  VELOX_CHECK_LT(std::abs(weightSoFar - totalWeight), kEpsilon);
+  // Use relative epsilon to handle floating-point precision issues
+  // with large totalWeight values
+  double relativeEpsilon =
+      std::max(kEpsilon, totalWeight * kRelativeErrorEpsilon);
+  VELOX_CHECK_LT(std::abs(weightSoFar - totalWeight), relativeEpsilon);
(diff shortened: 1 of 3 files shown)


velox/functions/lib/TDigest.h
```diff
@@ -338,7 +338,11 @@ void TDigest<A>::mergeImpl(
   }
   weightSoFar += weights_[j];
   ++numMerged_;
-  VELOX_CHECK_LT(std::abs(weightSoFar - totalWeight), kEpsilon);
+  // Use relative epsilon to handle floating-point precision issues
+  // with large totalWeight values
+  double relativeEpsilon =
+      std::max(kEpsilon, totalWeight * kRelativeErrorEpsilon);
+  VELOX_CHECK_LT(std::abs(weightSoFar - totalWeight), relativeEpsilon);
```
kagamiori, line 345: Hi @natashasehgal, thank you for making this fix. The original fuzzer error says `weightSoFar - totalWeight` was 8192 which was large. Is that normal, e.g., is it because totalWeight was a huge number such that 8192 was smaller than `totalWeight * kRelativeErrorEpsilon`?

Also, is it possible to add a unit test so that we won't break it in further code changes?

kagamiori: Sorry, somehow I didn't submit this ^ review 2 days ago while I thought I did.

natashasehgal: Added a test for this situation. Yes, relative error is required when the weight is a huge number, the acceptable error scales

PR #13956, opened 2025-07-01
fix: Spark corr aggregate function result

rui-mo:
Fixes the below result issue discovered by fuzzer test. When `m2X` or `m2Y` is 
zero, corr function used to return NULL, but Spark expects NaN if `c2` is NaN.

```
2 extra rows, 2 missing rows
2 of extra rows:
	15545 | null | 109 | null
	15545 | 13951 | null | null

2 of missing rows:
	15545 | null | 109 | "NaN"
	15545 | 13951 | null | "NaN"

SELECT g0, g1, g2, corr(c0, c1) as a0 FROM (SELECT c0 as c0, c1 as c1, g0 as g0, g1 as g1, g2 as g2 FROM (t_0)) GROUP BY g0, g1, g2
```


velox/functions/sparksql/aggregates/CovarianceAggregate.cpp
```diff
@@ -47,24 +47,31 @@ struct CovarSampResultAccessor {
...
       if (accumulator.count() == 1) {
         return true;
       }
-      return accumulator.m2X() != 0 && accumulator.m2Y() != 0;
     }
+    return hasNaN || notZero;
   }
 
   static double result(const CorrAccumulator& accumulator) {
     if (accumulator.count() == 1) {
```
zhli1142015, line 70: nit: 
```
if constexpr (!nullOnDivideByZero) {
if (accumulator.count() == 1) {
}
}
```

velox/functions/sparksql/aggregates/CovarianceAggregate.cpp
```diff
@@ -47,24 +47,31 @@ struct CovarSampResultAccessor {
...
         return false;
       }
-      return accumulator.m2X() != 0 && accumulator.m2Y() != 0;
     } else {
       if (accumulator.count() == 1) {
         return true;
       }
-      return accumulator.m2X() != 0 && accumulator.m2Y() != 0;
     }
+    return hasNaN || notZero;
```
zhli1142015: nit:
```
return std::isnan(accumulator.m2Y()) ||
        std::isnan(accumulator.m2X()) || std::isnan(accumulator.c2()) || 
        (accumulator.m2Y() != 0 && accumulator.m2X() != 0);
```

velox/functions/sparksql/aggregates/CovarianceAggregate.cpp
rui-mo: For the case where `nullOnDivideByZero` is true and `accumulator.count() == 1`, the `hasResult` returns false and `result` should not be called. This check is to ensure that so `if constexpr (!nullOnDivideByZero)` is not added. Do you think it makes sense?

jinchengchenghh: Only adjust the order, move nullOnDivideByZero first.

jinchengchenghh: This looks a bit complicated, I would suggest keep current implementation
if constexpr (!nullOnDivideByZero) {
if (count == 1) {
return NAN
} else {
exception
}
} else {
if (count == 1) {
}

jinchengchenghh (approved):

velox/functions/sparksql/aggregates/CovarianceAggregate.cpp
rui-mo: @jinchengchenghh The below might look clearer. Do you agree we change to it?

```
    if (accumulator.count() == 1) {
      if constexpr (!nullOnDivideByZero) {
        return std::numeric_limits<double>::quiet_NaN();
      } else {
        VELOX_UNREACHABLE();
      }
    }
```

velox/functions/sparksql/aggregates/CovarianceAggregate.cpp
rui-mo: Updated, thanks.

zhli1142015 (approved):

rui-mo:
Hi @bikramSingh91, I rebased this PR and resolved the conflict. May need your help to re-import. Thanks!

bikramSingh91:
@rui-mo Can you please rebase this change, it seems to have merge conflicts that need to be resolved with SparkAggregationFuzzerTest.cpp

rui-mo:
@bikramSingh91 I rebased this change. Thanks!

PR #13968, opened 2025-07-01
feat: Implement ST_StartPoint, ST_EndPoint

Sullivan-Patrick:
Summary:
Adds 2 UDFs.

ST_StartPoint:
```
Returns the first point of a LineString geometry as a Point. This is a shortcut for ST_PointN(geometry, 1).
```

ST_EndPoint:
```
Returns the last point of a LineString geometry as a Point. This is a shortcut for ST_PointN(geometry, ST_NumPoints(geometry)).
```

Differential Revision: D77607205

### DIFF
--- a/velox/functions/prestosql/tests/GeometryFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/GeometryFunctionsTest.cpp
@@ -1776,3 +1776,53 @@ TEST_F(GeometryFunctionsTest, testStPointN) {
       testStPointNFunc("POINT (1 2)", -1, std::nullopt),
       "ST_PointN only applies to LineString. Input type is: Point");
 }
+
+TEST_F(GeometryFunctionsTest, testStStartPoint) {
+  const auto testStStartPointFunc =
+      [&](const std::optional<std::string>& wkt,
+          const std::optional<std::string>& expected) {
+        std::optional<std::string> result = evaluateOnce<std::string>(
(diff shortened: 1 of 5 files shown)


velox/functions/prestosql/tests/GeometryFunctionsTest.cpp
```diff
@@ -1776,3 +1776,51 @@ TEST_F(GeometryFunctionsTest, testStPointN) {
...
+          ASSERT_TRUE(result.has_value());
+          ASSERT_EQ(result.value(), expected.value());
+        } else {
+          ASSERT_FALSE(result.has_value());
+        }
+      };
+
+  testStStartPointFunc("LINESTRING (8 4, 4 8, 5 6)", "POINT (8 4)");
+  testStStartPointFunc("LINESTRING (8 2, 4 12, 0 0)", "POINT (8 2)");
+  testStStartPointFunc("LINESTRING (0 0, 4 12, 2 2)", "POINT (0 0)");
```
bikramSingh91, line 1797: can you add a test case for an empty LINESTRING? I think it returns null, in that case, can you also update the description in the docs to specify this behavior?

Both here and for ST_ENDPOINT

Sullivan-Patrick: Thanks, good catch. It should be nullable, but it wasn't initially. Fixed this and added tests/docs.

bikramSingh91 (approved):

PR #13970, opened 2025-07-01
feat: Add scale_qdigest UDF

natashasehgal:
Differential Revision: D77614317


velox/functions/prestosql/QDigestFunctions.h
```diff
@@ -158,4 +158,108 @@ struct QuantileAtValueFunction {
...
 
+template <typename TExec>
+struct ScaleQDigestDoubleFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<SimpleQDigest<double>>& result,
+      const arg_type<SimpleQDigest<double>>& input,
+      const arg_type<double>& scaleFactor) {
+    VELOX_USER_CHECK(scaleFactor > 0, "Scale factor should be positive.");
```
kagamiori: nit: This can be removed. It's already checked in QuantileDigest::scale().

velox/functions/prestosql/tests/QDigestFunctionsTest.cpp
```diff
@@ -460,3 +460,78 @@ TEST_F(QDigestFunctionsTest, quantileAtValueNullInputs) {
...
+TEST_F(QDigestFunctionsTest, scaleQDigestNull) {
+  const auto scaleQDigest = [&](const std::optional<std::string>& input,
+                                const std::optional<double>& scale) {
+    return evaluateOnce<std::string>(
+        "scale_qdigest(c0, c1)", QDIGEST_DOUBLE, input, scale);
+  };
+
+  ASSERT_EQ(std::nullopt, scaleQDigest(std::nullopt, 2.0));
+  ASSERT_EQ(std::nullopt, scaleQDigest("digest", std::nullopt));
+}
```
kagamiori, line 537: It would be nice for QuantileDigest to have a test verifying the Velox result matches Java. Could you add a test case in QuantileDigestTest.cpp that calls QuantileDigest::scale() with a digest containing a relative large number of elements and checks that the Velox result is the same as Java's result of scale_qdigest()? The input qdigest can be these ones: https://github.com/facebookincubator/velox/blob/3826a76de89e06c64f2d14f078b9ec4e59bbf28f/velox/functions/lib/tests/QuantileDigestTest.cpp#L724-L737

velox/functions/prestosql/QDigestFunctions.h
natashasehgal: Done

velox/functions/prestosql/tests/QDigestFunctionsTest.cpp
natashasehgal: Thanks for the examples, added this

velox/functions/prestosql/QDigestFunctions.h
```diff
@@ -158,4 +158,100 @@ struct QuantileAtValueFunction {
...
+      const arg_type<double>& scaleFactor) {
+    std::allocator<float> allocator;
+    qdigest::QuantileDigest<float, std::allocator<float>> digest(
+        allocator, input.data());
+    digest.scale(scaleFactor);
+    int64_t size = digest.serializedByteSize();
+    result.resize(size);
+    digest.serialize(result.data());
+  }
+};
```
kagamiori: @natashasehgal  Do you still need these three separate classes since you're combining them into the ScaleQDigestFunction class below? The rest of this PR looks good to me.

natashasehgal: I was facing issues using ScaleQDigestFunction with TExec or TExec and T templating. Either in registration or overloading. Let me remove this one and keep the three separate classes

kagamiori: Hi @natashasehgal, I think the combined ScaleQDigestFunction would be better since it reduces the code duplicates. What issue did you come across? The templating with TExec and T would be similar to this method:
https://github.com/facebookincubator/velox/blob/96f0d4ac539dc596e62ab1ac3dc483c1e8b67945/velox/functions/prestosql/Arithmetic.h#L251-L258
with this registration:
https://github.com/facebookincubator/velox/blob/96f0d4ac539dc596e62ab1ac3dc483c1e8b67945/velox/functions/prestosql/registration/MathematicalFunctionsRegistration.cpp#L57

natashasehgal: Thanks for the code reference. I think the issue is because SimpleQDigest<T> is a complex type. If I use T and TInput, and use TInput& instead of arg_type I get 

E.g. 
error: no member named 'udf' in 'facebook::velox::functions::ScaleQDigestFunction<double>'
   47 |   using funcClass = typename Func::template udf<exec::VectorExec>;
      |                              ~~~~~~         ^

natashasehgal: Done!

Btw, on Meta using just template T worked but on Github there was a complaint about using C++ 20 functionality. So I moved to a template with T, ResultT, InputT

In file included from /__w/velox/velox/velox/functions/prestosql/registration/QDigestFunctionsRegistration.cpp:17:
/__w/velox/velox/./velox/functions/prestosql/QDigestFunctions.h:165:12: error: use of 'auto' in parameter declaration only available with '-std=c++20' or '-fconcepts' [-Werror]
  165 |   callImpl(auto& result, const auto& input, auto scaleFactor) {

kagamiori (approved):
LGTM. Thank you for adding this function and helping enhance QuantileDigestTest.

Issue #13767, opened 2025-06-13
array_min: Behavior mismatch between Presto and Velox due to signature differences


peterenescu:
### Description

`array_min` has behavior differences between Presto and Velox due to the signature differences between Presto and Velox. Presto will fail for type `array(hyperloglog)` while Velox permits such a type. As it currently stands Velox is more permissible, but we should decide if `hyperloglog` is indeed orderable.

### Error Reproduction

Presto execution:
```
SELECT cardinality(array_min(c0)) as p0, row_number as p1 FROM (SELECT transform(c0, (x) -> cast(x as HYPERLOGLOG)) as c0, row_number as row_number FROM (t_values))
```
```
Presto query failed: 1 line 1:20: Unexpected parameters (array(HyperLogLog)) for function array_min. Expected: array_min(array(T)) T:orderable, Source: RUNTIME, ErrorCode: INVALID_STATE
```

We should add this unit test to verify these changes eventually:
```
 void testHyperLogLog() {
    using S = StringView;

    auto arrayVector = makeNullableArrayVector<StringView>(
        {
            {S("0"), S("1")},
            {std::nullopt, S("1"), S("2"), S("3")},
            {},
            {S("1"), S("2"), S("3")},
        },
        ARRAY(HYPERLOGLOG()));
    auto expected = makeNullableFlatVector<StringView>(
        {S("0"), std::nullopt, std::nullopt, S("1")}, HYPERLOGLOG());
    testExpr<StringView>(expected, "array_min(C0)", {arrayVector});
  }
```
Currently it passes and is consistent with the Velox behavior, but it should throw (and we can change the test to expect it)

### Relevant logs

```bash

```

kagamiori:
The problem is that HyperLogLog type in Presto inherits AbstractType::isOrderable that is false (https://github.com/prestodb/presto/blob/b38dcc0e13c1c8739fe6058cf4236bc11912a096/presto-common/src/main/java/com/facebook/presto/common/type/AbstractType.java#L54). While HyperLogLogType in Velox inherits VarbinaryType::isOrderable that is true. We need to override HyperLogLogType::isOrderable to return false.

peterenescu:
@kagamiori We will need to do the same for JSON. JSON is also marked as orderable and will fail

PR #13972, opened 2025-07-01
fix: Update orderable-ness for custom types

peterenescu:
Summary:
In Presto, custom types JSON, for example, are not orderable because of their AbstractType inheritance, which is defaulted as false. In Velox, since we inherit directly from scalar types, they are marked as orderable because they inherit a true isOrderable field, which allows them to be called by UDFs with orderable type signatures:

https://github.com/facebookincubator/velox/issues/13767

Let's update the orderable-ness of these types. These include: Bingtile, Geometry, Json, Q/TDigest and Hyperloglog. Additionally, in order to test these functions, the fuzzer has been updated to use the internal array sort for the expression transform in order to sort non-orderable types.

Differential Revision: D77615175


kagamiori (approved):
LGTM. Thank you for the fix!

kagamiori (approved):

PR #13978, opened 2025-07-02
fix: Handling of empty nulls buffer in MapVector's Update API

bikramSingh91:
Summary:
This fixes a bug in MapVector's Update API which does not handle the case
where the calling vector itself does not have a nulls buffer. This results in a
Segfault when it tries to access the nulls buffer.

Differential Revision: D77628699

### DIFF
--- a/velox/common/testutil/OptionalEmpty.h
+++ b/velox/common/testutil/OptionalEmpty.h
@@ -30,4 +30,13 @@ struct OptionalEmptyT {
 
 inline constexpr OptionalEmptyT optionalEmpty;
 
+struct EmptyT {
+  template <typename T>
+  operator std::vector<T>() const {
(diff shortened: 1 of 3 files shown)


kevinwilfong (approved):
Good catch, thanks for the fix!

velox/common/testutil/OptionalEmpty.h
```diff
@@ -30,4 +30,13 @@ struct OptionalEmptyT {
 
 inline constexpr OptionalEmptyT optionalEmpty;
 
+struct EmptyT {
+  template <typename T>
+  operator std::vector<T>() const {
+    return std::vector<T>({});
```
kevinwilfong: nit: could you just do this?

return {};

PR #13903, opened 2025-06-27
feat: Add statistics unit tests

YiChengLee03:
Summary:
- Fix bugs in reader due to multithreaded environment, invalid flags for EOL and EOF
- Exec test for rawReadBytes statistics

Differential Revision: D77421942

PR #13987, opened 2025-07-02
fix: Add atPhysicalEOF_ in TextReader

xin-zhang2:
Currently, the EOF flag is set to true in two cases.
1. when the end of the file is reached.
2. when the limit in the RowReaderOptions is reached.

These two scenarios result it different behaviors in the `next` function. In case 1, the number of rows read is returned, while in case 2, it returns rows read + 1 because the next row has been scanned.

Now we determine the case 1 by comparing `pos_` with `getLength()`. However, for compressed input streams, the exact length is not known until the entire stream is read. Therefore, it might be more reasonable to determine it by the return value of the `Next` function of the input stream.

### DIFF
--- a/velox/dwio/text/reader/TextReader.cpp
+++ b/velox/dwio/text/reader/TextReader.cpp
@@ -117,6 +117,7 @@ TextRowReader::TextRowReader(
       atEOL_{false},
       atEOF_{false},
       atSOL_{false},
+      atPhysicalEOF_{false},
       depth_{0},
       unreadData_{""},
       unreadIdx_{0},
@@ -234,13 +235,13 @@ uint64_t TextRowReader::next(
       }
     }
     (void)skipLine();
+
+    if (atPhysicalEOF_) {
+      break;
+    }
+
     ++currentRow_;
     ++rowsRead;
-
-    if (pos_ >= getLength()) {
-      // disable further chunk reads but parse the remainder of the line
-      atEOF_ = true;
-    }
 
     // handle empty file
     if (initialPos == pos_ && atEOF_) {
@@ -594,11 +595,14 @@ char TextRowReader::getByteUncheckedOptimized(DelimType& delim) {
       int length;
       const void* buffer;
       if (!contents_->inputStream->Next(&buffer, &length)) {
+        atPhysicalEOF_ = true;
         setEOF();
         delim = DelimTypeEOR;
         return '\0';
       }
-
+      if (length == 0) {
+        return '\0';
+      }
       unreadData_ = std::string(reinterpret_cast<const char*>(buffer), length);
       unreadIdx_ = 0;
     }
--- a/velox/dwio/text/reader/TextReader.h
+++ b/velox/dwio/text/reader/TextReader.h
@@ -207,6 +207,7 @@ class TextRowReader : public RowReader {
   bool atEOL_;
   bool atEOF_;
   bool atSOL_;
+  bool atPhysicalEOF_;
   uint8_t depth_;
   std::string unreadData_;
   int unreadIdx_;


xin-zhang2:
@YiChengLee03 @zacw7 
Can you please help take a look? Thanks.

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -580,14 +583,22 @@ char TextRowReader::getByteUncheckedOptimized(DelimType& delim) {
   try {
     char v;
     if (unreadData_.empty()) {
```
YiChengLee03: Can we also check for unreadIdx_  >= unreadData_.size() here? (note: Missed it in my previous commits but is planned for a fix at #13903)

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -580,14 +583,22 @@ char TextRowReader::getByteUncheckedOptimized(DelimType& delim) {
...
+      }
+      if (length == 0) {
+        return '\0';
+      }
       unreadData_ = std::string(reinterpret_cast<const char*>(buffer), length);
       unreadIdx_ = 0;
     }
 
-    v = unreadData_[unreadIdx_++];
+    v = unreadData_[0];
```
YiChengLee03: Use unreadIdx_ for character read. We want to avoid erasing string as Next takes in ~8MB per call, erasing at idx = 0 could be a big hit on performance.

YiChengLee03:
Could you also elaborate on case 2? By "return rowsRead + 1" do you mean if our limit is in the middle of the row, we would also include that row in our Vector? Thanks!

xin-zhang2:
> Could you also elaborate on case 2? By "return rowsRead + 1" do you mean if our limit is in the middle of the row, we would also include that row in our Vector? Thanks!

@YiChengLee03 I thought it was an expected behavior as I noticed in [readRanges test](https://github.com/facebookincubator/velox/blob/main/velox/dwio/text/tests/reader/TextReaderTest.cpp#L883), the limit_ is set to read 5 rows but the expected return value of next is 6.

YiChengLee03:
> > Could you also elaborate on case 2? By "return rowsRead + 1" do you mean if our limit is in the middle of the row, we would also include that row in our Vector? Thanks!
> 
> I thought it is an expected behavior as I noticed in [readRanges test](https://github.com/facebookincubator/velox/blob/main/velox/dwio/text/tests/reader/TextReaderTest.cpp#L883), the limit_ is set to read 5 rows but the expected return value of next is 6.

Not quite. It is an odd case but its not universally assumed that we will return rowsRead + 1. There are a couple of things that we need to keep in mind for range reads. 
1. Reads are done in rows. This means that even if our limit is at the first character of the row, that row WILL be included in the output. 
2. In [getDelimType](https://github.com/facebookincubator/velox/blob/04e76d948c738d3ff30a875b52857967b1941562/velox/dwio/text/reader/TextReader.cpp#L509) we only set atEOF_ to true when its past the limit_. While this might not be logically sound, it is kept as is to be consistent with Presto. 
3. In the readRange test, our limit is byte 50 (file bytes are 0-indexed) which puts byte 50 at row 6. Following the logic from point 1, the 6th row will be parsed as our output. 

Hope this clears up any confusion on this odd case!

xin-zhang2:
@YiChengLee03 Thanks for the clarification. 

It aligns with my understanding, and my previous description of returning `rowsRead + 1` was inaccurate. Because of the check `pos_ > limit_`, we have to read the first character beyond the range to set the EOF to true, and include the row containing that character in the result.

That is also why I added `atPhysicalEOF_`. If EOF is set to true because we hit the limit_, we need to increment rowsRead to include that row. However,  if we reach the actual end of file during readElement, it means there's no more data to fill all fields of the current row, so we shouldn't include it.

Does this make sense?

YiChengLee03:
@xin-zhang2 Yes that makes sense! Thanks for the explanation.

YiChengLee03 (approved):

zacw7:
Could you please squash the commits?

xin-zhang2:
> Could you please squash the commits?

@zacw7 Done. Thanks for the review.

xin-zhang2:
> It seems TableScanTest.textfileEscape is failing?

@zacw7  yes, seems the logic of atPhysicalEOF_ is incorrect.

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -594,11 +595,14 @@ char TextRowReader::getByteUncheckedOptimized(DelimType& delim) {
...
       const void* buffer;
       if (!contents_->inputStream->Next(&buffer, &length)) {
+        atPhysicalEOF_ = true;
         setEOF();
         delim = DelimTypeEOR;
         return '\0';
       }
-
+      if (length == 0) {
+        return '\0';
```
YiChengLee03, line 604: Can you check if we have to manually set delim before returning here?

xin-zhang2:
@YiChengLee03 @zacw7 
Currently, we rely on `pos_ > getLength()` to set `atEOF_` flag to stop reading the next row. However, this approach doesn’t work for compressed input since we can’t determine the exact length until the stream is fully read. 
I introduced `atPhysicalEOF_` in this PR as an attempt to provide a unified solution for both compressed and uncompressed inputs based on the return value of `Next`. But it doesn’t quite work—by the time `Next` returns false, we’ve already inside the loop reading the next row, and it‘s uncertain whether that row should be included in the result. (If it occurs when we try to read the first byte of the new row, it means there's no more data and the row shouldn‘t be included, otherwise it should be).

I will think a bit more to see if there's another way. Let me know if you have any thoughts. Thanks!

xin-zhang2:
Created a PR for decompression https://github.com/facebookincubator/velox/pull/14003. Closing this one.

PR #13990, opened 2025-07-02
fix: Spark covar_samp aggregate function result

rui-mo:
Fixes the below result issue discovered by fuzzer test. When c2 is Inf,
covar_samp function used to return Inf, but Spark expects NaN.

```
1 extra rows, 1 missing rows
1 of extra rows:
	null | null | "-Infinity"

1 of missing rows:
	null | null | "NaN"

SELECT g0, g1, covar_samp(c0, c1) as a0 FROM (SELECT c0 as c0, c1 as c1, g0 as g0, g1 as g1 FROM (t_0)) GROUP BY g0, g1

```


zhli1142015 (approved):

rui-mo:
@jinchengchenghh Updated. Thanks.

PR #13993, opened 2025-07-02
fix: Respect Greek final sigma rule in Spark lower function

rui-mo:
Respects the Greek final sigma rule in Spark lower function. For the uppercase 
letter Σ, if it appears at the end of a word, it becomes ς. In all other 
positions, it becomes σ. If false, it is always converted to σ.

Renames the template parameter of lowerUnicode as `turkishCasing` and adds it 
in the `initcap`.

Fixes https://github.com/facebookincubator/velox/issues/14155.

### DIFF
--- a/velox/functions/lib/string/StringCore.h
+++ b/velox/functions/lib/string/StringCore.h
@@ -43,6 +43,30 @@
 #endif
 
 namespace facebook::velox::functions {
+namespace detail {
+
+// Helper function to check if a character is cased. Compatible with the
+// 'isCased' implementation in 'ConditionalSpecialCasting.java' of JDK, which is
+// used by 'toLowerCase' function in Spark SQL.
+FOLLY_ALWAYS_INLINE bool isCased(utf8proc_int32_t ch) {
+  auto type = utf8proc_category(ch);
+  // Lowercase letter, uppercase letter or titlecase letter.
+  if (type == UTF8PROC_CATEGORY_LL || type == UTF8PROC_CATEGORY_LU ||
+      type == UTF8PROC_CATEGORY_LT) {
+    return true;
+  }
+  // Modifier letters and special cases.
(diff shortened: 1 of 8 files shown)


rui-mo:
@zhli1142015  Would you like to review this minor refactor? Thanks.

velox/functions/lib/string/StringCore.h
```diff
@@ -43,6 +43,29 @@
 #endif
 
 namespace facebook::velox::functions {
+namespace detail {
+
+// Helper function to check if a character is cased. Compatible with the isCased
+// function of jdk used by 'toLowerCase' in Spark SQL.
+FOLLY_ALWAYS_INLINE bool isCased(int ch) {
```
liujiayi771: Could you add a comment to mark that this logic comes from the `isCased` method in `ConditionalSpecialCasting.java` of the JDK? This makes it easier to trace the source of the code logic.

rui-mo: Updated, thanks for the suggestion.

velox/functions/lib/string/StringImpl.h
```diff
@@ -770,7 +776,7 @@ FOLLY_ALWAYS_INLINE bool initcapUtf8Impl(
       outputBytes += upperSize;
       isStartOfWord = false;
     } else {
-      auto lowerSize = lowerUnicode<strictSpace>(
+      auto lowerSize = lowerUnicode<turkishCasing, greekFinalSigma>(
```
jinchengchenghh, line 779: The strictSpace is removed?

rui-mo: `strictSpace` param is only needed for the `isSpaceUtf8` call, this PR adds `turkishCasing` and `greekFinalSigma` to avoid confusion.

rui-mo:
The test failure is irrelevant with this change and on track in https://github.com/facebookincubator/velox/issues/14265.

liujiayi771 (approved):

PR #13994, opened 2025-07-02
fix: Check for null partitionFunctionSpec in PartitionedOutputNode constructor

kevinwilfong:
Summary:
The PartitionedOutputNode allows the user to specify a null partitionFunctionSpec.  This leads to a crash in the constructor for PartitionedOutput which attempts to dereference the pointer when the number of destinations is more than 1.

This change addresses the crash by:
1) Adding a check in the PartitionedOutputNode that if the number of destinations is > 1, the partitionFunctionSpec is non-null.  This will fail early with an exception rather than a crash.
2) Adding a check where we dereference partitionFunctionSpec that it is non-null as an added layer of safety.

Differential Revision: D77670782


kagamiori (approved):
LGTM. Thank you for the fix.

PR #13995, opened 2025-07-02
feat: Add OOM injector into TableEvolutionFuzzer

juwentus1234:
Summary: This implements the logic to enable OOM injection into the TableEvolutionFuzzer

Differential Revision: D77678185

### DIFF
--- a/velox/exec/tests/TableEvolutionFuzzer.cpp
+++ b/velox/exec/tests/TableEvolutionFuzzer.cpp
@@ -19,10 +19,20 @@
 #include "velox/dwio/common/tests/utils/FilterGenerator.h"
 #include "velox/exec/Cursor.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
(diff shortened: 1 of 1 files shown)


Yuhta (approved):

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -109,6 +119,18 @@ std::vector<std::vector<RowVectorPtr>> runTaskCursors(
           results.push_back(std::move(result));
         }
         promise.setValue(std::move(results));
+      } catch (VeloxRuntimeError& e) {
+        if (FLAGS_enable_oom_injection &&
+            e.errorCode() == facebook::velox::error_code::kMemCapExceeded &&
```
Yuhta, line 124: Sometimes a different exception is also allowed as long as there is no crash.  Let's see what we get with this though.

PR #13998, opened 2025-07-02
fix(typeparser): Handle presence of special characters in quoted string names

maniloya:
### Summary:
Fixes an issue in the type parser where special characters in quoted string names were not handled correctly.
### Details:
- Modified FLEX lexer rules to accommodate special characters within quoted strings.
### Testing:
- Added test cases for strings with special characters.
- Verified that all existing and new tests pass.
### Impact:
- No impact on other functionalities.

### DIFF
--- a/velox/type/parser/TypeParser.ll
+++ b/velox/type/parser/TypeParser.ll
@@ -34,7 +34,7 @@ Y   [Y|y]
 Z   [Z|z]
 
 WORD              ([[:alpha:][:alnum:]_]*)
-QUOTED_ID         (['"'][[:alnum:][:space:]_]*['"'])
+QUOTED_ID         (['"']([^"\n]|"")*['"'])
 NUMBER            ([[:digit:]]+)
 VARIABLE          (VARCHAR|VARBINARY)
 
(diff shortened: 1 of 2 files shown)


velox/type/parser/TypeParser.ll
```diff
@@ -34,7 +34,7 @@ Y   [Y|y]
 Z   [Z|z]
 
 WORD              ([[:alpha:][:alnum:]_]*)
-QUOTED_ID         (['"'][[:alnum:][:space:]_]*['"'])
+QUOTED_ID         (['"'][[:alnum:][:space:]_+-=:\\[\]]*['"'])
```
gggrace14: Hi @maniloya would you be able to point me at the corresponding rule from Presto Java? I have a random question, can '+' be the first character of QUOTED_ID?

maniloya: https://github.com/prestodb/presto/blob/b566980241fe4ef65f0f3db3e0c00148033136ef/presto-parser/src/main/antlr4/com/facebook/presto/sql/parser/SqlBase.g4#L932

based this syntax quote_identifier matches any thing within double qoutes.

maniloya: I have modified the `QUOTED_ID` to match the rules from Java code. @gggrace14

gggrace14: Hi @maniloya , I actually like your previous version that's minimum change to what we have today to this critical unit. Do you want to change it back? Other than that, everything looks good to me. I will approve after changing it back.

maniloya: discussed offline

gggrace14 (approved):

PR #14003, opened 2025-07-03
fix: Add decompression for TextReader

xin-zhang2:
Add decompression support for TextReader, including gzip, raw deflate, and zstd formats.


xin-zhang2:
@zacw7 @YiChengLee03 
A bit concern about the `blockSize` used when creating the decompressor. It is currently set to the max supported value of the decompressor to ensure enough buffer for decompression. Ideally, it should be set to the uncompressed length of the input stream, but it is unknown until we fully read the stream.

YiChengLee03:
> @zacw7 @YiChengLee03 A bit concern about the `blockSize` used when creating the decompressor. It is currently set to the max supported value of the decompressor to ensure enough buffer for decompression. Ideally, it should be set to the uncompressed length of the input stream, but it is unknown until we fully read the stream.

I think this is fair. We are doing the same thing with our internal text reader and I don't think we have had any issues by initializing with `std::numeric_limits<uint64_t>::max()` and you're absolutely right that we have no clue about the length of the compressed file. 

We could potentially optimize the buffer size of decompression if there are "worst case" compression rate (say worst case compression is 0.5x of the uncompressed fileSize, allocating 2x of the compressed fileSize will guarantee a buffer that is large enough) but that can be something to iterate on in the future as our focus is on rolling out this feature. 

If you are worried about `std::numeric_limits<uint64_t>::max()` not being big enough for the buffer size, I think internally we haven't experience this problem in the past, but that is a valid point of concern.

xin-zhang2:
@YiChengLee03  @zacw7 
I introduced `preLoadedUnreadData_` here as a way to fix the issue mentioned in https://github.com/facebookincubator/velox/pull/13987#issuecomment-3033229533.
Could you help take a look?

velox/dwio/text/tests/reader/TextReaderTest.cpp
```diff
@@ -1387,6 +1396,120 @@ TEST_F(TextReaderTest, tripleNestedArraysWithCustomDelimiters) {
...
+    EXPECT_TRUE(result->equalValueAt(expected.get(), i, 5 + i));
+  }
+  ASSERT_EQ(rowReader->next(5, result), 3);
+  for (int i = 0; i < 3; ++i) {
+    EXPECT_TRUE(result->equalValueAt(expected.get(), i, 9 + i));
+  }
+  ASSERT_EQ(rowReader->next(10, result), 0);
+}
+
+std::vector<TestCompressionParam> params = {
```
zacw7, line 1804: To double check: we have covered `.gz` in TextReaderTest::basic right? Shall we just move that here?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -87,6 +89,28 @@ void resizeVector(
...
+      str.compare(str.size() - suffix.size(), suffix.size(), suffix) == 0;
+}
+
+void setCompressionSettings(
+    const std::string& filename,
+    CompressionKind& kind,
+    compression::CompressionOptions& compressionOptions) {
+  if (endWith(filename, TEXTFILE_COMPRESSION_EXTENSION_GZIP)) {
+    kind = CompressionKind::CompressionKind_GZIP;
+    compressionOptions.format.zlib.windowBits = 15;
```
zacw7: n00b q: what does `15` and `-15` mean here?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -160,6 +186,30 @@ TextRowReader::TextRowReader(
       atEOF_ = true;
     }
     limit_ = std::numeric_limits<uint64_t>::max();
+
+    const auto blockSize =
+        (contents_->compression == CompressionKind::CompressionKind_ZLIB ||
+         contents_->compression == CompressionKind::CompressionKind_GZIP)
+        ? static_cast<uint64_t>(
+              UINT_MAX) // zlib requires blockSize being cast to an uInt
+        : limit_;
```
zacw7: it seems `limit_` will always be `std::numeric_limits<uint64_t>::max()`, why don't we just use the latter instead?

zacw7 (commented):
LG. Will leave it to @YiChengLee03 and @Yuhta for a final pass.

velox/dwio/text/tests/reader/TextReaderTest.cpp
YiChengLee03: Yes, DISABLED_basic, DISABLED_projectColumns, DISABLED_compressedProjectNone and DISABLED_shrinkBatch all uses .gz files. They should be enabled for this PR.

YiChengLee03 (commented):
LGTM. @Yuhta can you take a look at this PR too? Thanks!

velox/dwio/text/tests/reader/TextReaderTest.cpp
xin-zhang2: Yes, TextReaderTest::basic currently covers `.gz`. We could move it to TextReaderDecompressionTest to group all compression tyes, or keep it as is if we consider `.gz` a basic capability of the next reader.

@YiChengLee03 The last three cases are failing for uncompressed input now. We might need fix it before enabling.
```
Reason: Vector for column type does not match: expected FlatVector<StringView>, got null
```

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: windowBits determines the history buffer size, and here is the detailed description in zlib manual.

> The windowBits parameter is the base two logarithm of the maximum window size (the size of the history buffer). It should be in the range 8..15 for this version of the library. The default value is 15 if inflateInit is used instead. windowBits must be greater than or equal to the windowBits value provided to deflateInit2() while compressing, or it must be equal to 15 if deflateInit2() was not used.
> 
> windowBits can also be zero to request that inflate use the window size in the zlib header of the compressed stream.
> 
> windowBits can also be –8..–15 for raw inflate. In this case, -windowBits determines the window size. inflate() will then process raw deflate data, not looking for a zlib or gzip header, not generating a check value, and not looking for any check values for comparison at the end of the stream. This is for use with other formats that use the deflate compressed data format such as zip. Those formats provide their own check values. 
> 
> windowBits can also be greater than 15 for optional gzip decoding. Add 32 to windowBits to enable zlib and gzip decoding with automatic header detection, or add 16 to decode only the gzip format (the zlib format will return a Z_DATA_ERROR). 

We are using deflateInit2 in ZlibDecompressor, and 32 is added for gzip to enable the header detection. 
https://github.com/facebookincubator/velox/blob/17d3540c8c776014259c6cacf165d86f280c0984/velox/dwio/common/compression/Compression.cpp#L153-L156
So in our case, windowBits can be 0 or 8..15 for gzip, and -8..-15 for raw deflate. I kept them as 15 and -15 respectivlely, since they are typically the default values and also consistent with the ones we use for parquet and orc compression.
https://github.com/facebookincubator/velox/blob/17d3540c8c776014259c6cacf165d86f280c0984/velox/dwio/common/compression/Compression.h#L27-L32

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: Sure, I've updated the code to used `std::numeric_limits<uint64_t>::max()` directly.

velox/dwio/text/tests/reader/TextReaderTest.cpp
YiChengLee03: I'll take a look at it. Thanks for pointing that out!

velox/dwio/text/reader/TextReader.cpp
zacw7: Thanks for the detailed explanation! Can we add the const to Compression like `DWRF_ORC_ZLIB_WINDOW_BITS`/`PARQUET_ZLIB_WINDOW_BITS`?

velox/dwio/text/tests/reader/TextReaderTest.cpp
YiChengLee03: @xin-zhang2 #14072 and #14074 addresses the fix and add more comprehensive tests for uncompressed projectColums and shrinkBatch. There is already a working test for uncompressed projectNone [here](https://github.com/facebookincubator/velox/blob/db33557af3e00819455fca969adfc8f461d093c8/velox/dwio/text/tests/reader/TextReaderTest.cpp#L608).

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: I'm not sure its a good idea to add them to Compression. Since Compression.h is under dwio/common, it's probably better to keep format specifc constants out of it.

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -25,9 +27,9 @@ namespace facebook::velox::dwio::common {
 
 using common::CompressionKind;
 
-const std::string TEXTFILE_CODEC = "org.apache.hadoop.io.compress.GzipCodec";
-const std::string TEXTFILE_COMPRESSION_EXTENSION = ".gz";
-const std::string TEXTFILE_COMPRESSION_EXTENSION_RAW = ".deflate";
+const std::string TEXTFILE_COMPRESSION_EXTENSION_GZIP = ".gz";
```
Yuhta: The coding convention is to use `constexpr const char*` for compile time constant strings and camel case names (`kTextfileCompressionExtensionGzip`)

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -586,16 +641,37 @@ char TextRowReader::getByteUncheckedOptimized(DelimType& delim) {
...
     char v;
+    if (contents_->compression != CompressionKind::CompressionKind_NONE &&
+        preLoadedUnreadData_.empty()) {
+      int length = 0;
+      const void* buffer = nullptr;
+      atPhysicalEOF_ =
+          !contents_->decompressedInputStream->Next(&buffer, &length);
+      if (!atPhysicalEOF_) {
+        preLoadedUnreadData_ =
+            std::string(reinterpret_cast<const char*>(buffer), length);
```
Yuhta: This makes a new allocation, can we avoid it?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -156,6 +181,30 @@ TextRowReader::TextRowReader(
       atEOF_ = true;
     }
     limit_ = std::numeric_limits<uint64_t>::max();
+
+    const auto blockSize =
+        (contents_->compression == CompressionKind::CompressionKind_ZLIB ||
+         contents_->compression == CompressionKind::CompressionKind_GZIP)
+        ? static_cast<uint64_t>(
```
Yuhta: This should be `ReaderOptions::loadQuantum()`

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: Thanks for pointing this out. I changed the names to follow the convention.

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: Yes, I changed preLoadedUnreadData_ to string_view to avoid its allocation.

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: `ReaderOptions::loadQuantum()` specifies the maximun number of bytes to read from input.

In this case, `blockSize` is passed to the constructor of the decompressor to indicate the expected uncompressed length of the input. 

In ZlibDecompressor, it determines the available space in the output buffer before decompression.
https://github.com/facebookincubator/velox/blob/3826a76de89e06c64f2d14f078b9ec4e59bbf28f/velox/dwio/common/compression/Compression.cpp#L570-L575
In ZstdDecompressor, it is used as a fallback for uncompressed length when `ZSTD_getDecompressedSize` fails to provide a valid value.
https://github.com/facebookincubator/velox/blob/3826a76de89e06c64f2d14f078b9ec4e59bbf28f/velox/dwio/common/compression/Compression.cpp#L431-L446

As mentioned in https://github.com/facebookincubator/velox/pull/14003#issuecomment-3032742235, since the uncompressed length is unknown at read time, I set `blockSize` to `std::numeric_limits<uint64_t>::max()` or `UINT_MAX` to ensure sufficient buffer is allocated for the output.  

Please correct me if I misunderstood anything, or let me know if you have any suggestion for handling this. Thanks!

Yuhta: `UINT_MAX` just does not make sense here, it's essentially the same as `VELOX_FAIL`.  In 99% of the case the value won't be used though, as all common compressions have ways to figure out the decompressed size.  For a fallback, maybe put something reasonable like 4M and document that is a fallback and should not be trigger in normal case.

For `zstream_.avail_out` it probably can be changed to use the allocated buffer size since we already know it.

xin-zhang2: @Yuhta 
> as all common compressions have ways to figure out the decompressed size.

That might not always be the case. The `getDecompressedLength` method is defined in `Decompressor` and by default returns `blockSize_`. ZstdDecompressor and SnappyDecompressor override this method with their specific logic. However, ZlibDecompressor didn't override it because the decompressed length is generally unknown. Raw deflate format doesn't include any fields indicating the uncompressed length. While gzip format has an ISIZE field in the trailer represents the uncompressed length modulo 2^32, we don't make use of it in our code as there's no special logic for handling gzip in ZlibDecompressor now.

There is a default value `kDefaultCompressionBlockSize` defined in `Compression.h` which is 256K, but I'm unsure whether it is appropriate to set `blockSize` to it here as a fallback value. If the buffer size is insufficient to hold the decompressed data, we might get a failure as we currently don't handle this buffer overflow. 

For zlib decompression, the output buffer size is the same as `blockSize_`, since the return value of  `getDecompressedLength` is used to determine the allocation size.
https://github.com/facebookincubator/velox/blob/a53ef7ed5a10d24b09dce267ab6c92af0d1b41a0/velox/dwio/common/compression/PagedInputStream.cpp#L21-L25

libianoss:
@zacw7 @Yuhta can you continue to review the further change from @xin-zhang2 ? Thanks.

velox/dwio/text/reader/TextReader.cpp
Yuhta: I see, in this case let's just set it to `INT32_MAX` here, and document only zlib will use this value to allocate buffer, the other format will use the actual decompressed size.

xin-zhang2: @Yuhta  Thanks for the review. 

I added a comment explaining in detail how `blockSize` is used in zlib and other compression format. 
For zlib, it is assgined to `avail_out` via `zstream_.avail_out = folly::to<uInt>(blockSize)`, so I set it to `std::numeric_limits<unsigned int>::max()` to make it compatible with zlib `uInt`.
Can you please take another look? Thanks.

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include "velox/dwio/text/reader/TextReader.h"
 #include "velox/common/encode/Base64.h"
+#include "velox/dwio/common/compression/Compression.h"
+#include "velox/dwio/common/compression/PagedInputStream.h"
```
zacw7: is this #include needed?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -91,6 +93,28 @@ void resizeVector(
   }
 }
 
+bool endWith(const std::string& str, const std::string& suffix) {
```
zacw7: nit: endsWith()

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -91,6 +93,28 @@ void resizeVector(
...
+      str.compare(str.size() - suffix.size(), suffix.size(), suffix) == 0;
+}
+
+void setCompressionSettings(
+    const std::string& filename,
+    CompressionKind& kind,
+    dwio::common::compression::CompressionOptions& compressionOptions) {
+  if (endWith(filename, kTextfileCompressionExtensionGzip)) {
+    kind = CompressionKind::CompressionKind_GZIP;
+    compressionOptions.format.zlib.windowBits = 15;
```
zacw7: Thanks for the explanations. Could you please add some comments for the magic numbers?

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: Thanks for pointing it out. They are not needed and I've removed them.

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: The numbers represent the base-2 logarithm of the window size (the history buffer) used in the DEFLATE algorithm. A value of 15 means the size is 2^15 bytes, while −15 indicates the same size (2^15 bytes) but with raw deflate (no header or trailer).  I’ve added comments to explain them.

velox/dwio/text/reader/TextReader.cpp
xin-zhang2: Done. Thanks.

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -91,6 +91,30 @@ void resizeVector(
   }
 }
 
+bool endsWith(const std::string& str, const std::string& suffix) {
+  return str.size() >= suffix.size() &&
+      str.compare(str.size() - suffix.size(), suffix.size(), suffix) == 0;
+}
+
+void setCompressionSettings(
```
zacw7, line 99: shall we add a sanity check if filename.empty()

xin-zhang2: I think the check might be unnecessary here. If the filename is empty, the compression kind will be set to NONE as all `endsWith` return false.

Issue #14004, opened 2025-07-03
ConstantTypedExpr::equals and toString are buggy


mbasmanova:
### Bug description

ConstantTypedExpr can be created either using a VectorPtr or a Variant. It should not matter which way is used to create the expression. However, toString() and equals() method do not work correctly when one expression is created using Vector and other using Variant. 

The following test should pass, but it doesn't.

```
TEST_F(ExprTest, xxx) {
  Variant value{"test"};

  auto a = std::make_shared<core::ConstantTypedExpr>(VARCHAR(), value);

  auto b = std::make_shared<core::ConstantTypedExpr>(
      BaseVector::createConstant(VARCHAR(), value, 1, pool()));

  EXPECT_TRUE(a->equals(*b));

Value of: a->equals(*b)
  Actual: false
Expected: true

  EXPECT_TRUE(b->equals(*a));

Value of: b->equals(*a)
  Actual: false
Expected: true

  EXPECT_EQ(a->hash(), b->hash());

Expected equality of these values:
  a->hash()
    Which is: 13340254087754004103
  b->hash()
    Which is: 782655427638633022

  EXPECT_EQ(a->toString(), b->toString());

Expected equality of these values:
  a->toString()
    Which is: "\"test\""
  b->toString()
    Which is: "test"
```

CC: @aditi-pandit @bikramSingh91 @czentgr @oerling 

### System information

n/a

### Relevant logs

```bash

```

mbasmanova:
The difference in hashes is due to Vector::hashValueAt is hashing StringView (non-zero terminated string), while Variant::hash hases zero-terminated std::string.

CC: @bikramSingh91 @kKPulla

mbasmanova:
The difference in toString output for arrays:

```
Variant::array({1, 2, 3})

Expected equality of these values:
  a->toString()
    Which is: "[1,2,3]"
  b->toString()
    Which is: "3 elements starting at 0 {1, 2, 3}"

Expected equality of these values:
  a->hash()
    Which is: 11283352413530666588
  b->hash()
    Which is: 15052984279813004003
```

mbasmanova:
Fixed in https://github.com/facebookincubator/velox/pull/14006

PR #14006, opened 2025-07-03
fix: ConstantTypedExpr::toString

mbasmanova:
Summary:
ConstantTypedExpr can be created from a Vector or a Variant.

https://github.com/facebookincubator/velox/issues/14004 reports that the result of toString(), hash() and equals() depends on whether the expressions was created from a vector or a variant.

This fix is for ConstantTypedExpr::toString() to produce consistent result regardless of how the expression was created.

Differential Revision: D77746095


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

velox/core/Expressions.cpp
```diff
@@ -125,6 +127,335 @@ TypedExprPtr ConstantTypedExpr::create(
...
+bool equalsNoNulls<TypeKind::ARRAY>(
+    const VectorPtr& vector,
+    vector_size_t index,
+    const Variant& value) {
+  auto* wrappedVector = vector->wrappedVector();
+  VELOX_CHECK_EQ(VectorEncoding::Simple::ARRAY, wrappedVector->encoding());
+
+  auto* arrayVector = wrappedVector->asUnchecked<ArrayVector>();
+
+  const auto offset = arrayVector->offsetAt(index);
```
bikramSingh91, line 243: should we translate the index incase there is an encoding layer in-between?
This can break if the ConstantTypedExpr is created with an ConstantVector where the constantIndex does not point to zero OR if this is a nested complex type of the form CONST(ARRAY(DICT(ARRAY)))) and the dictionary layer adds an indirection.

velox/core/Expressions.cpp
```diff
@@ -125,6 +127,335 @@ TypedExprPtr ConstantTypedExpr::create(
...
+  auto* mapVector = wrappedVector->asUnchecked<MapVector>();
+
+  const auto size = mapVector->sizeAt(index);
+
+  const auto& mapValue = value.value<TypeKind::MAP>();
+  if (size != mapValue.size()) {
+    return false;
+  }
+
+  const auto sortedIndices = mapVector->sortedKeyIndices(index);
```
bikramSingh91, line 279: vector's sortedKeyIndices() uses CompareFlags to modify comparison behavior, do we know if the map variant that returns std::map uses the same comparison logic?

velox/core/Expressions.cpp
mbasmanova: @bikramSingh91 This is a good point. Let me fix and add a test.

velox/core/Expressions.cpp
mbasmanova: The Variant map uses this comparison logic. Looks compatible with nullsFirst, ascending, kNullAsValue flags used by sortedKeyIndices.

```
template <TypeKind KIND>
bool Variant::lessThan(const Variant& other) const {
  using namespace facebook::velox::util::floating_point;
  if (isNull() && !other.isNull()) {
    return true;
  }
  if (isNull() || other.isNull()) {
    return false;
  }
  using T = typename TypeTraits<KIND>::NativeType;

  if constexpr (kindCanProvideCustomComparison<KIND>::value) {
    if (usesCustomComparison_) {
      return customComparisonType<KIND>()->compare(
                 value<KIND>(), other.value<KIND>()) < 0;
    }
  }

  if constexpr (std::is_floating_point_v<T>) {
    return NaNAwareLessThan<T>{}(value<KIND>(), other.value<KIND>());
  }
  return value<KIND>() < other.value<KIND>();
}

bool Variant::operator<(const Variant& other) const {
  if (other.kind_ != this->kind_) {
    return other.kind_ < this->kind_;
  }
  return VELOX_DYNAMIC_TYPE_DISPATCH_ALL(lessThan, kind_, other);
}
```

velox/core/Expressions.cpp
mbasmanova: @bikramSingh91 Updated the PR. Would you take another look?

bikramSingh91 (approved):

Issue #14018, opened 2025-07-05
Variant::hash doesn't respect the structure of the map


mbasmanova:
### Bug description

The hash function uses commutative hash to combine all keys and all values separately. This results in unnecessary hash collisions:

```
TEST(VariantTest, xxx) {
  auto a = Variant::map({{1, 10}, {2, 20}});
  auto b = Variant::map({{1, 20}, {2, 10}});

  LOG(ERROR) << a.hash();
  LOG(ERROR) << b.hash();
}

E0705 14:05:50.692492 1516777 VariantTest.cpp:149] 9577158183860959173
E0705 14:05:50.692611 1516777 VariantTest.cpp:150] 9577158183860959173
```

CC: @bikramSingh91 @oerling 

### System information

n/a

### Relevant logs

```bash

```

Issue #14017, opened 2025-07-05
ArrayVector::hashValueAt ignores the order of elements in the array


mbasmanova:
### Bug description

Array values hash the same regardless of the order of elements. hash([1, 2, 3]) == hash([3, 2, 1]). 

```
TEST_F(ExprTest, xxx) {
  auto a =
      makeArrayVectorFromJson<int32_t>({"[1, 2, 3]", "[3, 2, 1]", "[1, 3, 2]"});

  LOG(ERROR) << a->hashValueAt(0);
  LOG(ERROR) << a->hashValueAt(1);
  LOG(ERROR) << a->hashValueAt(2);
}

E0705 03:47:01.547931 759211 ExprTest.cpp:2601] 4705813931562929230
E0705 03:47:01.548061 759211 ExprTest.cpp:2602] 4705813931562929230
E0705 03:47:01.548077 759211 ExprTest.cpp:2603] 4705813931562929230
```

This happens due to use of commutative hash (bits::commutativeHashMix):

```
uint64_t hashArray(
    uint64_t hash,
    const BaseVector& elements,
    vector_size_t offset,
    vector_size_t size) {
  for (auto i = 0; i < size; ++i) {
    auto elementHash = elements.hashValueAt(offset + i);
    hash = bits::commutativeHashMix(hash, elementHash);
  }
  return hash;
}
```

We should use non-commutative hash.

Similar issue affects MapVector::hashValueAt.

CC: @bikramSingh91 @oerling @Yuhta 

### System information

n/a

### Relevant logs

```bash

```

PR #14019, opened 2025-07-05
Fix Variant::hash and BaseVector::hashValueAt for arrays and maps

mbasmanova:
Summary:
Commutative hashes ignore the order in which values are added to the hash. 

When hashing arrays, one must not use commutative hash, but ArrayVector::hashValueAt used that. This caused unnecessary hash collisions as hashes of two arrays that contain same elements but in different order were always the same. hash([1, 2, 3]) was the same as hash([3, 2, 1]).

When hashing maps, we need to use non-commutative hash to combine hashes of each key and value in a pair, but then combine hashes of pairs using commutative hash. 

Variant::hash used commutative hash to combine hashes of all keys, commutative hash to combine hashes of all values and non-commutative hash to combine these the results. Thus, hash({1: 10, 2: 20}) was same as hash({2: 10, 1: 20}). MapVector::hashValueAt had even more severe problem as it used commutative hash to combine hashes of all keys and values together. Thus, hash({1: 10, 2: 20}) was the same as hash({10: 20, 2: 1}).

Fixes https://github.com/facebookincubator/velox/issues/14018

Fixes https://github.com/facebookincubator/velox/issues/14017

Differential Revision: D77823844

### DIFF
--- a/velox/type/Variant.cpp
+++ b/velox/type/Variant.cpp
@@ -850,20 +850,17 @@ uint64_t Variant::hash<TypeKind::TIMESTAMP>() const {
 template <>
 uint64_t Variant::hash<TypeKind::MAP>() const {
   auto hasher = folly::Hash{};
-  auto& mapVariant = value<TypeKind::MAP>();
-  uint64_t combinedKeyHash = 0, combinedValueHash = 0;
-  uint64_t singleKeyHash = 0, singleValueHash = 0;
-  for (auto it = mapVariant.begin(); it != mapVariant.end(); ++it) {
-    singleKeyHash = it->first.hash();
-    singleValueHash = it->second.hash();
-    combinedKeyHash = folly::hash::commutative_hash_combine_value_generic(
-        combinedKeyHash, hasher, singleKeyHash);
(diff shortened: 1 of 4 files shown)


velox/exec/ContainerRowSerde.cpp
```diff
@@ -1000,18 +1010,39 @@ uint64_t hashOne(ByteInputStream& in, const Type* type) {
   const auto& keyType = type->childAt(0);
   const auto& valueType = type->childAt(1);
 
-  uint64_t hash;
+  auto size = in.read<int32_t>();
+
+  std::vector<uint64_t> keyHashes;
```
Yuhta, line 1015: This may have some performance implication.  Can we loop over keys and values at same time?

Yuhta (approved):
One suggestion

velox/exec/ContainerRowSerde.cpp
mbasmanova: Not very easily since we don't know where the 'values' start in the stream until after we have read the 'keys'.

PR #14024, opened 2025-07-06
fix(iceberg): Use nextRowNumber to update baseReadOffset

zhli1142015:
Parquet row group filtering causes row numbers to be non‑contiguous,
leading to baseReadOffset_ + splitOffset_ misalign with row indices.
This discrepancy causes Position Delete File entries to map to wrong
rows, as shown in unit tests: a filtered‑out middle row group results in
deletion records for the middle row group applying to the third row
group instead.
To fix this, baseReadOffset_ is now updated before each read using
nextRowNumber(), ensuring offsets stay in sync with actual row numbers.


zhli1142015:
@yingsu00 @Yuhta and @xiaoxmeng , could you help to review this PR? Thanks

zhli1142015:
Kindly ping~
@majetideepak could you please help to review this PR?

zhli1142015:
@rui-mo and @jinchengchenghh could you help to review this PR?

jinchengchenghh:
@yingsu00 Could you help review this PR? Thanks!

velox/connectors/hive/iceberg/tests/CMakeLists.txt
```diff
@@ -57,5 +57,13 @@ if(NOT VELOX_DISABLE_GOOGLETEST)
...
     GTest::gtest
     GTest::gtest_main)
+  if(VELOX_ENABLE_PARQUET)
+
+    target_link_libraries(velox_hive_iceberg_test velox_dwio_parquet_reader)
+
+  endif()
+
+  file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/examples
+       DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
rui-mo: nit: do we only need to copy them when VELOX_ENABLE_PARQUET in enabled?

rui-mo (approved):
LGTM. Thanks.

velox/connectors/hive/iceberg/tests/CMakeLists.txt
zhli1142015: Yes, the test cases that reference this file will only run when Parquet is enabled.

rui-mo: In this case, we might need to move the copy inside the `if(VELOX_ENABLE_PARQUET)`, right?

zhli1142015: sure.

majetideepak (approved):

majetideepak:
CC: @PingLiuPing

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -273,6 +284,54 @@ class HiveIcebergTest : public HiveConnectorTestBase {
     return splits;
   }
 
+#ifdef VELOX_ENABLE_PARQUET
+  std::vector<std::shared_ptr<ConnectorSplit>> createParquetDeleteFileAndSplits(
+      const std::string& path,
+      const std::vector<int64_t>& deletePositionsVec,
+      int32_t deletedPositionSize,
```
PingLiuPing, line 291: nit: This parameter might not necessary.

PingLiuPing (approved):
Thanks.

@yingsu00 , Just FYI, this is an example that we might miss some fix in upstream.

PR #14028, opened 2025-07-07
refactor: Decouple SQL parsing from type resolution

mbasmanova:
Summary: SQL is parsed into IExpr (defined in velox/parse), then IExpr is type resolved into ITypedExpr (defined in velox/core).

Differential Revision: D77864254

### DIFF
--- a/velox/parse/PlanNodeIdGenerator.h
+++ b/velox/parse/PlanNodeIdGenerator.h
@@ -15,7 +15,8 @@
  */
 #pragma once
 
-#include "velox/core/PlanNode.h"
+#include <fmt/format.h>
+#include <string>
 
 namespace facebook::velox::core {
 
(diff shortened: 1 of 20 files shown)


velox/parse/PlanNodeIdGenerator.h
```diff
@@ -25,7 +26,7 @@ class PlanNodeIdGenerator {
  public:
   explicit PlanNodeIdGenerator(int startId = 0) : nextId_{startId} {}
 
-  PlanNodeId next() {
+  std::string next() {
```
xiaoxmeng, line 29: We don't want to use PlanNodeId as alias here?

xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

velox/parse/PlanNodeIdGenerator.h
mbasmanova: It creates an unnecessary dependency on velox/core.

PR #14030, opened 2025-07-07
refactor: Prepare the writeCellValue to be recursive for TextWriter

kewang1024:
Summary: As titled

Differential Revision: D77848914

### DIFF
--- a/velox/dwio/text/writer/TextWriter.cpp
+++ b/velox/dwio/text/writer/TextWriter.cpp
@@ -94,13 +94,19 @@ void TextWriter::write(const VectorPtr& data) {
     decodedColumnVectors.push_back(std::move(decodedColumnVector));
   }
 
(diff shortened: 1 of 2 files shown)


velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -94,13 +94,19 @@ void TextWriter::write(const VectorPtr& data) {
     decodedColumnVectors.push_back(std::move(decodedColumnVector));
   }
 
+  std::optional<char> delimiter;
```
zacw7, line 97: Shall we switch to `uint8_t` as the delimiter type for consistency?

kewang1024: Yes, I will leave that part to @ycliu613 once he lands his refactor code of moving TextWriterTrait to Options

PR #14035, opened 2025-07-07
fix: Handle integer overflow in abs Presto function

pramodsatya:
Fixes `abs` Presto function implementation to throw in case of Integer overflow.
Presto fixes this with a similar [check](https://github.com/prestodb/presto/blob/821731906c7801a860397757e84cf0e16871659e/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java#L128) for Integer minimum bounds.

### DIFF
--- a/velox/functions/prestosql/ArithmeticImpl.h
+++ b/velox/functions/prestosql/ArithmeticImpl.h
@@ -156,12 +156,6 @@ T negate(const T& arg) {
 }
 
 template <typename T>
-T abs(const T& arg) {
-  T results = std::abs(arg);
-  return results;
-}
-
-template <typename T>
 T floor(const T& arg) {
   T results = std::floor(arg);
   return results;
--- a/velox/functions/prestosql/tests/ArithmeticTest.cpp
+++ b/velox/functions/prestosql/tests/ArithmeticTest.cpp
@@ -181,6 +181,11 @@ class ArithmeticTest : public functions::test::FunctionBaseTest {
     assertExpression("c0 * c1", op1, doubleOp, expected);
     assertExpression("c1 * c0", op1, doubleOp, expected);
   }
+
+  template <typename T>
+  std::optional<T> evaluateAbs(const T& input) {
+    return evaluateOnce<T>("abs(c0)", std::make_optional<T>(input));
+  }
 };
 
 TEST_F(ArithmeticTest, plus) {
@@ -1040,6 +1045,36 @@ TEST_F(ArithmeticTest, wilsonIntervalUpper) {
   VELOX_ASSERT_THROW(
       wilsonIntervalUpper(1, 3, -kInf), "z-score must not be negative");
   EXPECT_DOUBLE_EQ(wilsonIntervalUpper(1, 3, kInf).value(), 1.0);
+}
+
+TEST_F(ArithmeticTest, abs) {
+  ASSERT_EQ(
+      evaluateAbs(std::numeric_limits<int8_t>::min() + 1),
+      std::numeric_limits<int8_t>::max());
+  VELOX_ASSERT_THROW(
+      evaluateAbs(std::numeric_limits<int8_t>::min()),
+      "Out of range integer input to abs: -128.");
+
+  ASSERT_EQ(
+      evaluateAbs(std::numeric_limits<int16_t>::min() + 1),
+      std::numeric_limits<int16_t>::max());
+  VELOX_ASSERT_THROW(
+      evaluateAbs(std::numeric_limits<int16_t>::min()),
+      "Out of range integer input to abs: -32768.");
+
+  ASSERT_EQ(
+      evaluateAbs(std::numeric_limits<int32_t>::min() + 1),
+      std::numeric_limits<int32_t>::max());
+  VELOX_ASSERT_THROW(
+      evaluateAbs(std::numeric_limits<int32_t>::min()),
+      "Out of range integer input to abs: -2147483648.");
+
+  ASSERT_EQ(
+      evaluateAbs(std::numeric_limits<int64_t>::min() + 1),
+      std::numeric_limits<int64_t>::max());
+  VELOX_ASSERT_THROW(
+      evaluateAbs(std::numeric_limits<int64_t>::min()),
+      "Out of range integer input to abs: -9223372036854775808.");
 }
 
 } // namespace
--- a/velox/functions/prestosql/Arithmetic.h
+++ b/velox/functions/prestosql/Arithmetic.h
@@ -218,8 +218,14 @@ struct FloorFunction {
 template <typename TExec>
 struct AbsFunction {
   template <typename T>
-  FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
-    result = abs(a);
+  FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
+    if constexpr (std::is_integral_v<T>) {
+      if (a == std::numeric_limits<T>::min()) {
+        return Status::UserError("Out of range integer input to abs: {}.", a);
+      }
+    }
+    result = std::abs(a);
+    return Status::OK();
   }
 };
 


pramodsatya:
@majetideepak, @aditi-pandit, could you please help review this fix?

aditi-pandit:
@pramodsatya : Its interesting that this hasn't been caught by ExpressionFuzzer with PrestoQueryRunner... Can you check it includes these expressions and the exception is thrown in a consistent way ?

velox/functions/prestosql/tests/ArithmeticTest.cpp
```diff
@@ -1042,5 +1048,31 @@ TEST_F(ArithmeticTest, wilsonIntervalUpper) {
   EXPECT_DOUBLE_EQ(wilsonIntervalUpper(1, 3, kInf).value(), 1.0);
 }
 
+TEST_F(ArithmeticTest, abs) {
+  ASSERT_EQ(getAbs<int8_t>(-127), 127);
+  VELOX_ASSERT_THROW(
+      getAbs<int8_t>(std::numeric_limits<int8_t>::min()),
+      fmt::format(
+          "Value -128 is out of range for abs({})", CppToType<int8_t>::name));
```
majetideepak: Can you put the exact string for testing purposes?

majetideepak (commented):
@pramodsatya thanks for the fix. One minor comment.

@aditi-pandit the odds for the fuzzer to hit this is low since it is specific to `std::numeric_limits<T>::min()`

pramodsatya:
> @pramodsatya : Its interesting that this hasn't been caught by ExpressionFuzzer with PrestoQueryRunner... Can you check it includes these expressions and the exception is thrown in a consistent way ?

Thanks for pointing it out @aditi-pandit, agree with @majetideepak that the fuzzer would likely not have caught this corner case. Even if the expression `abs(INT_MIN)` were to be generated by the fuzzer and it throws in Presto, such failures would be classified as `unsupported in the reference DB` and are likely tolerated. Would it be fine to check the expression fuzzer behavior and potentially enhance it to catch such corner cases in a follow-up issue/PR?

velox/functions/prestosql/tests/ArithmeticTest.cpp
pramodsatya: Updated, thanks.

aditi-pandit:
@pramodsatya : This was an interesting function since it seems like Velox generated incorrect results in the past, and fuzzer is run for hours to generally touch these corner cases.

You can add a follow up for the fuzzer.

aditi-pandit (approved):
Thanks @pramodsatya

velox/functions/prestosql/ArithmeticImpl.h
```diff
@@ -157,6 +157,12 @@ T negate(const T& arg) {
 
 template <typename T>
 T abs(const T& arg) {
+  if constexpr (std::is_integral_v<T>) {
+    if (arg == std::numeric_limits<T>::min()) {
+      VELOX_USER_FAIL(
```
mbasmanova: Can we use non-throwing API?

Let's make sure that runtime information is at the end of the error message t?

Please, don't use CppToType.

velox/functions/prestosql/tests/ArithmeticTest.cpp
```diff
@@ -181,6 +181,12 @@ class ArithmeticTest : public functions::test::FunctionBaseTest {
     assertExpression("c0 * c1", op1, doubleOp, expected);
     assertExpression("c1 * c0", op1, doubleOp, expected);
   }
+
+  template <typename T>
+  std::optional<T> getAbs(const T& input) {
```
mbasmanova: This is only used in one test, hence, better make it a local lambda.

velox/functions/prestosql/ArithmeticImpl.h
pramodsatya: Updated to return `Status::UserError` and reverted use of `CppToType`.
(The initial error message was borrowed from Presto to maintain consistency)

velox/functions/prestosql/tests/ArithmeticTest.cpp
pramodsatya: I tried to use a lambda initially but looks like templated lambdas are supported from C++20 onwards. Would it be fine to retain this helper function for now, till Velox supports C++20?

mbasmanova: I see.

Then, let's rename to something like abs or evaluateAbs.

velox/functions/prestosql/tests/ArithmeticTest.cpp
```diff
@@ -181,6 +181,12 @@ class ArithmeticTest : public functions::test::FunctionBaseTest {
...
+    const auto a = std::make_optional<T>(input);
```
mbasmanova: Can we remove this temp variable?

```
return evaluateOnce("abs(c0)", input);
```

velox/functions/prestosql/tests/ArithmeticTest.cpp
```diff
@@ -1042,5 +1048,24 @@ TEST_F(ArithmeticTest, wilsonIntervalUpper) {
   EXPECT_DOUBLE_EQ(wilsonIntervalUpper(1, 3, kInf).value(), 1.0);
 }
 
+TEST_F(ArithmeticTest, abs) {
+  ASSERT_EQ(getAbs<int8_t>(-127), 127);
+  VELOX_ASSERT_THROW(
+      getAbs<int8_t>(std::numeric_limits<int8_t>::min()),
+      "Out of range integer input to abs: -128.");
+  ASSERT_EQ(getAbs<int16_t>(-32767), 32767);
```
mbasmanova: Would you add empty lines to separate tests for different types for readability?

I assume we can drop template parameter.

```
getAbs(std::numeric_limits<int8_t>::min())
```

pramodsatya: Updated accordingly, thanks.

velox/functions/prestosql/tests/ArithmeticTest.cpp
```diff
@@ -1042,5 +1047,35 @@ TEST_F(ArithmeticTest, wilsonIntervalUpper) {
   EXPECT_DOUBLE_EQ(wilsonIntervalUpper(1, 3, kInf).value(), 1.0);
 }
 
+TEST_F(ArithmeticTest, abs) {
+  ASSERT_EQ(
+      evaluateAbs(std::make_optional(std::numeric_limits<int8_t>::min() + 1)),
```
mbasmanova: Can you remove std::make_optional ?

pramodsatya: Sure, could I modify `evaluateAbs` like:
```
 template <typename T>
  std::optional<T> evaluateAbs(const T& input) {
    return evaluateOnce<T>("abs(c0)", std::make_optional<T>(input));
  }
```
or use a temp variable like before:
```
 template <typename T>
  std::optional<T> evaluateAbs(const T& input) {
    const auto a = std::make_optional<T>(input);
    return evaluateOnce<T>("abs(c0)", a);
  }
```

pramodsatya: Updated with first option, please let me know if this is fine, thanks.

zhouyuan:
@rui-mo @zhli1142015 
It seems this patch changed the behavior on Spark + Non-ANSI mode, shall we add a Spark version for?
https://github.com/apache/incubator-gluten/actions/runs/16209987045/job/45769900999?pr=10166

rui-mo:
@zhouyuan I've been noticing failures for abs in the expr fuzzer test with Spark query runner. I'll take a look whether we need to re-implement it.

mbasmanova:
@zhouyuan This change is in prestosql folder. Do we have a dependency from Spark on that folder? If so, let's decouple to avoid having such dependency.

rui-mo:
@mbasmanova Spark registers some Presto functions for reuse, like the Abs function. You're right, we might need to have Spark's own version.

https://github.com/facebookincubator/velox/blob/260698993bce966dd314e69c94eebbe6e84cedf3/velox/functions/sparksql/registration/RegisterMath.cpp#L28

PR #14037, opened 2025-07-07
misc: Add ability to track stats on non-special form expressions

bikramSingh91:
Summary:
Recently a query config `operator_track_expression_stats` was added to expose
expression stats via operator stats, however, the stats exposed turned out to
included every expression (constants and field references, etc) and became
unwieldy. For the use-case that motivated this change, we are only interested
in expressions that evaluate a UDF and therefore encompass all non-special
form expressions. We therefore update the query config to only track these
kind of expression. In the future, if we decide to expose more options, we can
enhance this config to support those.

Differential Revision: D77903975


mbasmanova (approved):
Thanks.

PR #14038, opened 2025-07-08
fix(vector): EnsureWritable should copy buffer views

pedroerp:
Summary:
Ensure that ensureWritable() copies buffer views in complex types used
for offsets and lengths. BaseVector already does the right thing for null
buffers, but complex types previouslyu only checked whether offsets and
lengths were unique, but not if they were not views (buffer->unique() vs.
buffer->isMutable()). Adding the fix plus unit tests.

Differential Revision: D77904499

### DIFF
--- a/velox/buffer/Buffer.h
+++ b/velox/buffer/Buffer.h
@@ -686,6 +686,13 @@ class BufferView : public Buffer {
     return result;
   }
 
+  // Helper method to create a buffer view referencing another existing Buffer.
+  static BufferPtr
+  create(BufferPtr innerBuffer, Releaser releaser, bool podType = true) {
+    return create(
+        innerBuffer->as<uint8_t>(), innerBuffer->size(), releaser, podType);
+  }
+
   ~BufferView() override {
     releaser_.release();
   }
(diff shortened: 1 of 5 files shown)


kKPulla (approved):
Good catch. Thanks for fixing this.

velox/buffer/Buffer.h
```diff
@@ -686,6 +686,13 @@ class BufferView : public Buffer {
     return result;
   }
 
+  // Helper method to create a buffer view referencing another existing Buffer.
+  static BufferPtr
+  create(BufferPtr innerBuffer, Releaser releaser, bool podType = true) {
```
Yuhta, line 691: In the case `releaser` is the `innerBuffer` itself, there is no need to pass it separately
```c++
  static BufferPtr
  create(const BufferPtr& innerBuffer, bool podType = true) {
    return create(innerBuffer->as<uint8_t>(), innerBuffer->size(), innerBuffer, podType);
  }
```

pedroerp: @Yuhta innerBuffer is not necessarily a `BufferView`. It could be just a regular `Buffer`, and in that case it doesn't contain a `releaser`.

PR #14039, opened 2025-07-08
feat: Add bits::negateBit function

oliver6782:
Summary: Add a function to negate a single bit.

Differential Revision: D77906047

### DIFF
--- a/velox/common/base/tests/BitUtilTest.cpp
+++ b/velox/common/base/tests/BitUtilTest.cpp
@@ -541,6 +541,21 @@ TEST_F(BitUtilTest, getAndClearLastSetBit) {
   EXPECT_EQ(bits, 0);
 }
 
+TEST_F(BitUtilTest, negateBit) {
+  char data[35];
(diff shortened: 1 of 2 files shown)


velox/common/base/tests/BitUtilTest.cpp
```diff
@@ -541,6 +541,21 @@ TEST_F(BitUtilTest, getAndClearLastSetBit) {
...
+  char data[35];
+  for (int32_t i = 0; i < 100; i++) {
+    setBit(data, i, true);
+  }
+  std::vector<uint64_t> indices = {0, 1, 2, 3, 4, 5, 6, 7};
+  for (auto i : indices) {
+    negateBit(data, i);
+    EXPECT_EQ(isBitSet(data, i), false);
+  }
+  for (int32_t i = 8; i < 100; i++) {
```
mbasmanova, line 554: nit: perhaps, verify that first 8 bits are set as well to make sure that negateBit(5) doesn't affect bits other that '5'.

PR #10904, opened 2024-08-31
Fix fmt V11 build errors 

yingsu00:
Recently brew updated fmt version from 10.1 to 11.0. In order to prevent build errors from happening after we upgrade Folly, the following changes are made:

1. Error "'this' argument to member function 'format' has type 'const fmt::formatter<facebook::velox::exec::Spiller::Type>', but function is not marked const".
The new fmt v11 forces the const check. To fix this error, the "const" keyword was added to the `format()` function declarations.

2. Error "error: static assertion failed due to requirement 'formattable_char': Mixing character types is disallowed." The new fmt v11 added a check if the arguments of fmt::make_format_args() have mixing different character types (e.g., char vs wchar_t or std::string vs folly::Range<const char*>). To fix this, we need to convert the folly::StringPiece to string.

These changes will be needed after we upgrade Folly, which is also dependent on fmt v11. The changes are backward compatible for fmt v10.

Fixes https://github.com/facebookincubator/velox/issues/10886

PR #14040, opened 2025-07-08
fix: Fmt V11 build errors

zml1206:
Similar to #10904, fix build error if install fmt via brew:
```In file included from /opt/homebrew/include/fmt/format.h:41:
/opt/homebrew/include/fmt/base.h:2275:20: error: 'this' argument to member function 'format' has type 'const fmt::formatter<facebook::velox::memory::ArbitrationOperation::State>', but function is not marked const
 2275 |     ctx.advance_to(cf.format(*static_cast<qualified_type*>(arg), ctx));```


zml1206:
cc @assignUser @majetideepak Thanks.

PR #14041, opened 2025-07-08
feat: Support ARRAY Type in TextWriter

YiChengLee03:
Summary: Add support for ARRAY complex type

Differential Revision: D77912304

### DIFF
--- a/velox/dwio/text/tests/writer/TextWriterTest.cpp
+++ b/velox/dwio/text/tests/writer/TextWriterTest.cpp
@@ -15,28 +15,51 @@
  */
 
 #include "velox/dwio/text/writer/TextWriter.h"
-#include <gtest/gtest.h>
+#include "velox/buffer/Buffer.h"
 #include "velox/common/file/FileSystems.h"
+#include "velox/dwio/text/RegisterTextReader.h"
+#include "velox/dwio/text/RegisterTextWriter.h"
+#include "velox/dwio/text/reader/TextReader.h"
 #include "velox/dwio/text/tests/writer/FileReaderUtil.h"
 #include "velox/exec/tests/utils/TempDirectoryPath.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
 
+#include <gtest/gtest.h>
+
+/// TODO: Add fuzzer test.
+
 namespace facebook::velox::text {
-// TODO: add fuzzer test once text reader is move in OSS
 class TextWriterTest : public testing::Test,
                        public velox::test::VectorTestBase {
  public:
   void SetUp() override {
     velox::filesystems::registerLocalFileSystem();
-    dwio::common::LocalFileSink::registerFactory();
+    registerTextWriterFactory();
+    dwio::common::registerTextReaderFactory();
     rootPool_ = memory::memoryManager()->addRootPool("TextWriterTests");
     leafPool_ = rootPool_->addLeafChild("TextWriterTests");
     tempPath_ = exec::test::TempDirectoryPath::create();
   }
 
+  void TearDown() override {
+    dwio::common::unregisterTextReaderFactory();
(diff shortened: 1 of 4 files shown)


velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -139,73 +150,95 @@ void TextWriter::writeCellValue(
         serDeOptions_.nullString.data(), serDeOptions_.nullString.length());
     return;
   }
+
+  ++depth_;
```
kewang1024: Let's not put depth in member variable but use a local variable in `write` function

velox/dwio/text/tests/writer/TextWriterTest.cpp
```diff
@@ -82,74 +106,104 @@ TEST_F(TextWriterTest, write) {
...
-  EXPECT_EQ(result[1][0], "true");
-  EXPECT_EQ(result[2][0], "true");
-
-  // tinyint
-  EXPECT_EQ(result[0][1], "1");
-  EXPECT_EQ(result[1][1], "2");
-  EXPECT_EQ(result[2][1], "3");
-
-  // smallint
-  EXPECT_EQ(result[0][2], "1");
```
kewang1024: I would suggest keeping the original test and add your own test of verifying by reader in a new test. 

The reason for that is TextWriter and TextReader could share certain logic or options, if bug was introduced in the common component, then it would be neglected. Having a non-dynamic expected value would be helpful in this case.

If the array printing could be large, in this test, you can use a smaller array or helper function to verify

YiChengLee03: That makes sense. I think I can also just compare with the expected text line by line. This helps verify the correctness of delimiters written as well.

velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -139,73 +151,96 @@ void TextWriter::writeCellValue(
...
           dataStr.value().data(), dataStr.value().length());
-      return;
+      break;
     }
     case TypeKind::TINYINT: {
       auto dataStr = toTextStr(decodedColumnVector->valueAt<int8_t>(row));
       bufferedWriterSink_->write(
           dataStr.value().data(), dataStr.value().length());
-      return;
+      break;
```
kewang1024: can change back to `return`

kewang1024 (approved):
change the PR and commit name: 
feat: Support ARRAY Type in TextWriter

please address the comment before landing

velox/dwio/text/tests/writer/TextWriterTest.cpp
```diff
@@ -137,9 +161,9 @@ TEST_F(TextWriterTest, write) {
   EXPECT_EQ(result[2][6], "3.100000");
 
   // timestamp
-  EXPECT_EQ(result[0][7], "1970-01-01 00:00:00.000");
-  EXPECT_EQ(result[1][7], "1970-01-01 00:00:01.001");
-  EXPECT_EQ(result[2][7], "1970-01-01 00:00:02.002");
+  EXPECT_EQ(result[0][7], "1969-12-31 16:00:00.000");
+  EXPECT_EQ(result[1][7], "1969-12-31 16:00:01.001");
+  EXPECT_EQ(result[2][7], "1969-12-31 16:00:02.002");
```
kewang1024, line 166: I think you need to wait for your previous diff/pr land before landing this one

velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -139,73 +151,96 @@ void TextWriter::writeCellValue(
         serDeOptions_.nullString.data(), serDeOptions_.nullString.length());
     return;
   }
+
+  ++depth;
+
```
kewang1024: Remove extra line L156

PR #14042, opened 2025-07-08
fix: Arrow patch name error

zml1206:
follow up #14016.


zml1206:
cc @assignUser @majetideepak Thank you.

assignUser (approved):
Oh, thanks for catching that!

Issue #13144, opened 2025-04-24
Found incompatible return types for 'parse_presto_data_size'


rui-mo:
### Description

ExpressionVerifier.cpp:117] ExprSet: Exceptions other than VeloxRuntimeError of UNSUPPORTED_INPUT_UNCATCHABLE are not allowed.
terminate called after throwing an instance of 'facebook::velox::VeloxUserError'
  what():  Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Found incompatible return types for 'parse_presto_data_size' (DECIMAL(38, 0) vs. DECIMAL(24, 8)) for input types (VARCHAR).
Retriable: False
Expression: resultType->equivalent(*simpleFunctionEntry->type().get())
Function: compileRewrittenExpression
/home/runner/work/velox/velox/velox/velox/expression/ExprCompiler.cpp
Line: 436

### Relevant logs

```bash
https://github.com/facebookincubator/velox/actions/runs/14638746294/job/41079267105?pr=9559
```

Issue #12546, opened 2025-03-05
Flakiness in ScaleWriterLocalPartitionTest.partitionBasic


bikramSingh91:
### Bug description

ScaleWriterLocalPartitionTest.partitionBasic failed during CI

### System information

Failed in CI
https://github.com/facebookincubator/velox/actions/runs/13665973578/job/38207294497?pr=12537

### Relevant logs

```bash

```

mbasmanova:
Still seeing this:

https://github.com/facebookincubator/velox/actions/runs/15973254294/job/45049134481?pr=13920

```
[ RUN      ] ScaleWriterLocalPartitionTest.partitionBasic
/__w/velox/velox/velox/exec/tests/ScaleWriterLocalPartitionTest.cpp:891: Failure
Expected equality of these values:
  planStats.at(exchnangeNodeId) .customStats.count( ScaleWriterPartitioningLocalPartition::kScaledPartitions)
    Which is: 0
  1
```

@bikramSingh91 @xiaoxmeng Would it possible to prioritize a fix?

mbasmanova:
Still seeing this: https://github.com/facebookincubator/velox/actions/runs/17409995796/job/49424503987

xiaoxmeng:
> Still seeing this: https://github.com/facebookincubator/velox/actions/runs/17409995796/job/49424503987

I can reproduce locally and will look into this week @gggrace14 @kewang1024

mbasmanova:
Still seeing failures: https://github.com/facebookincubator/velox/actions/runs/17600046045/job/50000273645?pr=14809

```
[----------] 2 tests from ScaleWriterLocalPartitionTest
[ RUN      ] ScaleWriterLocalPartitionTest.unpartitionBasic
/__w/velox/velox/velox/exec/tests/ScaleWriterLocalPartitionTest.cpp:696: Failure
Expected equality of these values:
  planStats.at(exchnangeNodeId) .customStats.count(ScaleWriterLocalPartition::kScaledWriters)
    Which is: 1
  0
Google Test trace:
/__w/velox/velox/velox/exec/tests/ScaleWriterLocalPartitionTest.cpp:616: numProducers 4, numConsumers 4, rebalanceProcessBytesThreshold 0B, scaleWriterRebalanceMaxMemoryUsageRatio 0.1, holdBufferBytes 128.00MB, producerBufferedMemoryRatio 0.8, consumerBufferedMemoryRatio 0.6, expectedRebalance false
```

[...]

PR #14044, opened 2025-07-08
misc: Enable decimal expression fuzzer test in workflow

rui-mo:
The decimal expression fuzzer test had been enabled in the experiment workflow, 
which was removed with https://github.com/facebookincubator/velox/commit/331b82de2b423af464ced0422df61d7009d6cce8. This PR enables it in the scheduled workflow.

Fixes the fuzzer failure issue for Presto `parse_presto_data_size` function. 
When `parse_presto_data_size` is nested inside another function `A`,the return 
type of `parse_presto_data_size` is decided when fuzzing argument types for `A`,
which is random and causes a mismatch with the expected DECIMAL(38, 0).
This PR fixes this issue by respecting integer constraints in the
`checkOrSetIntegerParameter` so as to skip the test for this function.

Includes ParsePrestoDataSizeTest.cpp in the CMakeLists.

Fixes https://github.com/facebookincubator/velox/issues/13144.

### DIFF
--- a/velox/expression/SignatureBinder.cpp
+++ b/velox/expression/SignatureBinder.cpp
@@ -163,6 +163,12 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
     return false;
   }
 
+  const auto& constraint = variables().at(parameterName).constraint();
+  if (isPositiveInteger(constraint) && atoi(constraint.c_str()) != value) {
+    // Return false if the actual value does not match the constraint.
+    return false;
+  }
+
   if (integerVariablesBindings_.count(parameterName)) {
     // Return false if the parameter is found with a different value.
     if (integerVariablesBindings_[parameterName] != value) {
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -414,6 +414,7 @@ jobs:
                 --seed ${random_seed} \
                 --enable_variadic_signatures \
                 --velox_fuzzer_enable_complex_types \
+                --velox_fuzzer_enable_decimal_type \
                 --lazy_vector_generation_ratio 0.2 \
                 --common_dictionary_wraps_generation_ratio=0.3 \
                 --velox_fuzzer_enable_column_reuse \
@@ -542,6 +543,7 @@ jobs:
                 --duration_sec 3600 \
                 --enable_variadic_signatures \
                 --velox_fuzzer_enable_complex_types \
+                --velox_fuzzer_enable_decimal_type \
                 --velox_fuzzer_enable_column_reuse \
                 --velox_fuzzer_enable_expression_reuse \
                 --max_expression_trees_per_step 2 \
@@ -633,6 +635,7 @@ jobs:
           ./spark_expression_fuzzer_test \
                 --seed ${random_seed} \
                 --duration_sec $DURATION \
+                --velox_fuzzer_enable_decimal_type \
                 --minloglevel=0 \
                 --stderrthreshold=2 \
                 --log_dir=/tmp/spark_bias_fuzzer_repro/logs \
@@ -678,6 +681,7 @@ jobs:
                 --max_expression_trees_per_step 2 \
                 --retry_with_try \
                 --enable_dereference \
+                --velox_fuzzer_enable_decimal_type \
                 --duration_sec $DURATION \
                 --minloglevel=0 \
                 --stderrthreshold=2 \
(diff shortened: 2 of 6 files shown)


rui-mo:
@mbasmanova @kagamiori Would you like to review this fuzzer change? Thanks!

velox/functions/prestosql/registration/DataSizeFunctionsRegistration.cpp
```diff
@@ -20,16 +20,36 @@
...
+
+  static std::string name() {
+    return fmt::format("{}", id);
+  }
+};
+
+// Expression fuzzer cannot bind integer variable of the return type directly,
+// so we use LongDecimal with the names being precision and scale.
+// 'tryResolveLongLiteral' will resolve the variable to the corresponding
+// integer value.
```
kagamiori: Hi @rui-mo, could you share a bit more (description or code pointer) of where in fuzzer this binding is needed but doesn't work? I wonder whether we could fix it there?

rui-mo: Hi @kagamiori, I added more context for this issue, thanks.

---
For example, in the case where `parse_presto_data_size` is nested inside `least`, the return type of `parse_presto_data_size` is decided by the `ArgumentTypeFuzzer::fuzzArgumentTypes` of the `least` function, which is unaware of the constraints of the `parse_presto_data_size`. 

At the below point for the `least` function, the integer bindings are decided and are used to gen the the return type of `parse_presto_data_size` which is random and not DECIMAL(38, 0).

https://github.com/facebookincubator/velox/blob/3826a76de89e06c64f2d14f078b9ec4e59bbf28f/velox/expression/fuzzer/ArgumentTypeFuzzer.cpp#L182

The constraints of the `parse_presto_data_size` function is detectable when fuzzing the argument type for itself, but it's too late for the return type. The reason is that in the fuzzer the outer function is responsible for deciding its argument type, which is the return type of the inner function in the same time.

kagamiori: Hi @rui-mo, thank you for the explanation. I understand the problem now. It seems that expression fuzzer should refuse `parse_presto_data_size` when it try to pick a function signature to generate `least`'s argument. Is my understand correct? In this situation, I think we should fix it here: https://github.com/facebookincubator/velox/blob/96f0d4ac539dc596e62ab1ac3dc483c1e8b67945/velox/expression/fuzzer/ExpressionFuzzer.cpp#L975-L988

This piece of code is exactly for this purpose. What's missing now is that ReverseSignatureBinder::tryBind() doesn't check the constraint on the integer parameter. I think we should make ReverseSignatureBinder::tryBind() check that and return false for `parse_presto_data_size`. Then the expression fuzzer will look for another function signature to generate `least`'s argument.

rui-mo: Hi @kagamiori, thanks for sharing your thoughts. I understand your point and updated this PR. Would you take another look?

velox/expression/SignatureBinder.cpp
```diff
@@ -163,6 +163,12 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
     return false;
   }
 
+  const auto& constraint = variables().at(parameterName).constraint();
+  if (isPositiveInteger(constraint) && atoi(constraint.c_str()) != value) {
+    // Return false if the actual value does not match the constraint.
+    return false;
+  }
```
kagamiori, line 170: Hi @rui-mo, let's add a unit test for this change since it not only affects fuzzer. By the way, since checkOrSetIntegerParameter() is used in both ReverseSignatureBinder and the regular SignatureBinder, let's add unit tests for both. The rest of this PR looks good to me.

rui-mo: I updated this PR with corresponding tests. Could you please have another look? Thanks!

kagamiori:
Hi @rui-mo, there are some test failures in this PR. Could you take a look? Or rebase onto the latest main if those are irrelevant to your change?

rui-mo:
Hi @kagamiori, I also noticed the test failures in another PR and opened two issues https://github.com/facebookincubator/velox/issues/14199 and https://github.com/facebookincubator/velox/issues/14200 to track. The third issue is on track in https://github.com/facebookincubator/velox/issues/12546. Thanks.

kagamiori:
Hi @rui-mo, sorry but could you please rebase onto the latest main again (required by internal merge tool)? The diff is approved internally and can be merged once we rebase. Thanks!

PR #14045, opened 2025-07-08
feat(abfs): Add DynamicSasTokenClientProvider to support dynamic SAS token renewal

marin-ma:
In the Java implementation, user can implement and configure the 
`SasTokenProvider` for SAS token generation with interface 
https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-azure/src/main/java/org/apache/hadoop/fs/azurebfs/extensions/SASTokenProvider.java. 
Once the generated token has expired or is near expiry, it will be 
renewed by acquiring a new token from the provider. 

Currently in Velox, the SAS token is configured as a fixed token, 
and expiration is not considered. This patch adds 
`DynamicSasTokenClientProvider` to handle the SAS token renewal 
logic by updating the reader/writer client with a renewed SAS and 
adds `AbfsSasTokenProvider` interface to provide the SAS token. 
User can choose to register the `DynamicSasTokenClientProvider` 
with a custom `AbfsSasTokenProvider` implementation for 
accounts that requires dynamic SAS token renewal.

relies on https://github.com/facebookincubator/velox/pull/14281

### DIFF
--- a/velox/connectors/hive/storage_adapters/abfs/DynamicSasTokenClientProvider.cpp
+++ b/velox/connectors/hive/storage_adapters/abfs/DynamicSasTokenClientProvider.cpp
@@ -24,42 +24,42 @@ namespace {
 
 constexpr int64_t kDefaultSasTokenRenewPeriod = 120; // in seconds
 
-Azure::DateTime getExpiry(const std::string& token) {
+Azure::DateTime getExpiry(const std::string_view& token) {
   if (token.empty()) {
     return Azure::DateTime::clock::time_point::min();
   }
 
-  const std::string signedExpiry = "se=";
-  const int32_t signedExpiryLen = 3;
+  static constexpr std::string_view kSignedExpiry{"se="};
+  static constexpr int32_t kSignedExpiryLen = 3;
 
-  auto start = token.find(signedExpiry);
+  auto start = token.find(kSignedExpiry);
   if (start == std::string::npos) {
     return Azure::DateTime::clock::time_point::min();
   }
-  start += signedExpiryLen;
+  start += kSignedExpiryLen;
 
   auto end = token.find("&", start);
-  std::string seValue = (end == std::string::npos)
-      ? token.substr(start)
-      : token.substr(start, end - start);
+  auto seValue = (end == std::string::npos)
+      ? std::string(token.substr(start))
+      : std::string(token.substr(start, end - start));
 
   seValue = Azure::Core::Url::Decode(seValue);
   auto seDate =
       Azure::DateTime::Parse(seValue, Azure::DateTime::DateFormat::Rfc3339);
 
-  const std::string signedKeyExpiry = "ske=";
-  const int32_t signedKeyExpiryLen = 4;
+  static constexpr std::string_view kSignedKeyExpiry = "ske=";
+  static constexpr int32_t kSignedKeyExpiryLen = 4;
 
-  start = token.find(signedKeyExpiry);
+  start = token.find(kSignedKeyExpiry);
   if (start == std::string::npos) {
     return seDate;
   }
-  start += signedKeyExpiryLen;
+  start += kSignedKeyExpiryLen;
 
   end = token.find("&", start);
-  std::string skeValue = (end == std::string::npos)
-      ? token.substr(start)
-      : token.substr(start, end - start);
+  auto skeValue = (end == std::string::npos)
+      ? std::string(token.substr(start))
+      : std::string(token.substr(start, end - start));
 
   skeValue = Azure::Core::Url::Decode(skeValue);
   auto skeDate =
@@ -127,7 +127,7 @@ class DynamicSasTokenDataLakeFileClient final : public AzureDataLakeFileClient {
   DataLakeFileClient* getWriteClient() {
     if (writeClient_ == nullptr ||
         isNearExpiry(writeSasExpiration_, sasTokenRenewPeriod_)) {
-      const auto sas = sasKeyGenerator_->getSasToken(
+      const auto& sas = sasKeyGenerator_->getSasToken(
           abfsPath_->fileSystem(), abfsPath_->filePath(), kAbfsWriteOperation);
       writeSasExpiration_ = getExpiry(sas);
       writeClient_ = std::make_unique<DataLakeFileClient>(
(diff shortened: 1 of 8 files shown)


marin-ma:
@zhli1142015 Could you help to review? Thanks!

zhli1142015:
Thanks for your change.
I’ve noticed that the SAS‑token logic is scattered across many files. Could you extract a general interface, for example as below. Then we can encapsulate all the SAS token details inside a single ClientProvider.
```
class BlobClientProvider {
 public:
  virtual std::shared_ptr<AzureBlobClient> getBlobClient(AbfsConfig abfsConfig) = 0;
  std::string getAccountIdentifier()= 0；
};
```
```
class BlobClientProviderFactory {
 public:
  static void registerProvider(std::shared_ptr<BlobClientProvider> provider);

  static bool providerRegistered(std::string account);

  static std::shared_ptr<AzureBlobClient> getBlobClient(
      std::string_view path,
      AbfsConfig abfsConfig);
};
```

velox/connectors/hive/storage_adapters/abfs/AbfsConfig.cpp
```diff
@@ -53,14 +124,191 @@ class DataLakeFileClientWrapper final : public AzureDataLakeFileClient {
...
+
+  std::string GetUrl() override {
+    return blobClient_->GetUrl();
+  }
+
+ private:
+  std::unique_ptr<Azure::Storage::Blobs::BlobClient> blobClient_;
+};
+
+class DynamicSasKeyBlobClient : public AzureBlobClient {
```
zhli1142015: Instead of creating a client wrapper, please use `BlobClientOptions.PerRetryPolicies` to inject the SAS token renewal logic.
FYI, https://github.com/Azure/azure-sdk-for-cpp/blob/a332a845d0ac2bf270f4d536a76bb0b21f46bfeb/sdk/storage/azure-storage-blobs/src/blob_client.cpp#L62

zhli1142015: https://github.com/Azure/azure-sdk-for-cpp/blob/a332a845d0ac2bf270f4d536a76bb0b21f46bfeb/sdk/storage/azure-storage-common/inc/azure/storage/common/internal/shared_key_policy.hpp#L15

marin-ma: @zhli1142015 Could you elaborate a bit more on how to update the SAS token by adding this injection? From the [blob sas example](https://github.com/Azure/azure-sdk-for-cpp/blob/a332a845d0ac2bf270f4d536a76bb0b21f46bfeb/sdk/storage/azure-storage-blobs/samples/blob_sas.cpp#L72-L77) Seems like we have to create a new `BlobClient` if the sas token is renewed.

majetideepak:
Please see this general discussion on token renewal 
https://github.com/facebookincubator/velox/issues/13914

velox/connectors/hive/storage_adapters/abfs/AbfsConfig.cpp
zhli1142015: In this injection, we can only operate on the HTTP request's properties. Although the SAS token is just a query string in the request, it’s indeed not that straightforward. Either approach is ok to me.

marin-ma:
> I’ve noticed that the SAS‑token logic is scattered across many files. Could you extract a general interface, for example as below. Then we can encapsulate all the SAS token details inside a single ClientProvider.

@zhli1142015 I'm working on addressing this comment and find this approach introduces some code refactoring. The changes have now been made in the latest commits, including:

1. Currently the Azure read/write clients are created from the `AbfsConfig`. With introducing the `AzureClientProvider`  interface for creating read/write client, the part of logic in `AbfsConfig` to handle the auth configurations should be extracted out. Therefore I moved this part of logic into the `DefaultAzureClientProvider`. For account that is not registered with a `AzureClientProvider` implementation, it will use the `DefaultAzureClientProvider`.
2. Rename `AbfsConfig` to `AbfsPath`, which only preserves the logic that parses the abfs file path.
3. Adds `DynamicSasTokenClientProvider` implementation for the purpose of this PR. For use cases that uses SAS but do not provide a fixed SAS token, the caller/user needs to provide their `SasTokenProvider` implementation to use the DynamicSasTokenClientProvider`, which will handle the SAS token renew logic internally based on the token expiration.

As the refactoring brings lots of changes to this patch, I can extract 1 and 2 to another patch first. Please let me know if you have any advice to the modifications. Thanks!

marin-ma:
@zhli1142015 Could you spare some time to take a look at this change? Thanks!

zhli1142015:
Thanks for your change.
I had a quick look and I think we should create a dedicated ClientProvider for each authentication method instead of lumping all the logic into DefaultAzureClientProvider. That will make the design much clearer.
Of course, if you feel that’s too big of a change at once, you can break it into several smaller changes.

marin-ma:
@zhli1142015 Thanks! 

> I had a quick look and I think we should create a dedicated ClientProvider for each authentication method instead of lumping all the logic into DefaultAzureClientProvider.

Let me separate 1 and 2 to another patch first, and I will address this one in that patch.

marin-ma:
@zhli1142015 The refactor patch has been merged. Could you help to review this one? Thanks!

marin-ma:
@zhli1142015 Could you help to review? Thanks!

zhli1142015 (approved):
LGTM, thanks.

marin-ma:
@majetideepak @rui-mo Could you help to review? Thanks!

velox/connectors/hive/storage_adapters/abfs/DynamicSasTokenClientProvider.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+#include <azure/core/datetime.hpp>
+
+namespace facebook::velox::filesystems {
+
+namespace {
+
+constexpr int64_t kDefaultSasTokenRenewPeriod = 120; // in seconds
+
+Azure::DateTime getExpiry(const std::string& token) {
```
majetideepak: nit: string_view.

marin-ma:
@Yuhta Could you provide the error message for the linter? I can't see the internal error message. Thanks!

Yuhta:
@marin-ma No need to worry about that

marin-ma:
@Yuhta This patch is still not merged yet. Is there any follow up I need to do?

PR #14046, opened 2025-07-08
build: Refactor Docker Images and Workflow to facilitate multi-arch Images

assignUser:
This PR implements multi-architecture support for our Centos based Docker images and lead incidentally to a docker build system refactor.

### Multi-Stage Docker Builds
- Refactored CentOS and Java images into multi-stage builds, providing:
  - **Smaller final images** by excluding build tools and intermediate artifacts
  - **Better layer caching** and re-use
  - **Stable Caching** of rarely changing artifacts (Spark, Presto) 

### Docker Buildx Bake Implementation
- Added `docker-bake.hcl` as a central build configuration:
  - **Build matrices** allow for easy parameterization of targets, e.g. based on arch
  - **Multi-platform support** through specific targets and build caches per arch
  - **Efficient caching** through registry cache configuration, pushing _to_ the cache is disabled by default for local builds
  - **Parallel building** of independent images, this provides a massive speed up!
  - **Improved maintainability**, it's much easier to add new targets to the workflow once they are properly defined in the bake file
### Miscellaneous Changes
- Updated CI workflows to handle multi-architecture builds 
- Removed conda from adapters image 
- Removed `SHELL` for better compatibility with `podman`
- Streamlined Docker Compose configuration by removing redundant elements via `extends`
- I found various issues with our install scripts while working on this which where hidden because run_and_time in it's current form eats errors and let's the script continue.


assignUser:
I am adding pip back in to make the other workflows work again, I will switch that over in a follow up PR. I'll have to rebuild the images in my fork for the testing :)

assignUser:
Re-building the images again to fix some issues with the presto image

assignUser:
The ubuntu dockerfile could also be turned into a multi stage build but that isn't necessary for this build so i'll leave that for a follow up.

assignUser:
Working on reproducing the test failures locally, I have a feeling that this is an issue with the rebuild dependencies that happens due to changes to the setup script.

Edit: We were just missing `VELOX_BUILD_SHARED`.

assignUser:
I have found additional issues with our setup scripts but the tests do pass on a clean container, I will skip the verification step after the next rebuild.

assignUser:
This [commit](https://github.com/facebookincubator/velox/pull/14046/commits/902d0fe25f0e58082055475d03412ea43524047e) has the full green (except a normal fuzzer fail)run with the new images and the current run is with the old images.

(edit by bdice: fixed link)

docker-bake.hcl
```diff
@@ -0,0 +1,117 @@
...
+}
+
+target "ubuntu-amd64" {
+  inherits   = ["ubuntu-cpp"]
+  cache-to   = cache-to-arch("ubuntu", "amd64")
+  cache-from = cache-from-arch("ubuntu", "amd64")
+}
+
+group "ubuntu-arm64" {
+  # We don't actually want to build the ubuntu arm image, this is a trick to simplify CI
```
bdice, line 112: Is there a technical reason why the ARM image wouldn't work? Or just that it's not needed in the current CI matrix?

I think we would like ARM support for all the CI images, if it's just a matter of building and uploading.

bdice (approved):
Looks good to me.

Also, I highly recommend checking out https://github.com/facebookincubator/velox/pull/14046/files and trying to enable zstd compression for these images. We did this recently for our RAPIDS CI images and we saw very significant reductions in CI job startup time (like 6 minutes to 1.5 minutes) because the zstd compressed images are smaller and extract much faster.

docker-bake.hcl
assignUser: Yeah, the ubuntu image is not used in CI so I didn't think it necessary. But I want to rework it into multistage in a follow up anyway, so if there is interest to have this as multi-platform to we can add it.

scripts/docker/centos-multi.dockerfile
```diff
@@ -0,0 +1,166 @@
...
+FROM base-image AS centos9
+
+# Install tools needed for CI
+RUN /bin/bash -c "source /setup-centos9.sh && \
+      dnf_install 'dnf-command(config-manager)' && \
+      dnf config-manager --add-repo 'https://cli.github.com/packages/rpm/gh-cli.repo' && \
+      dnf_install gh jq && \
+      dnf clean all"
+
+ENV CC=/opt/rh/gcc-toolset-12/root/bin/gcc \
```
czentgr, line 89: Do we need this here if we call the enable in 92? The enable should move this to the forefront.

assignUser: eh probabl not? I'd have to test it, lets do it in a follow up, it also doesn;t hurt so :shrug:

assignUser:
@kgpai @majetideepak would you mind reviewing?

scripts/setup-centos-adapters.sh
```diff
@@ -0,0 +1,112 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# shellcheck source-path=SCRIPT_DIR
+
+# This script installs addition dependencies for the adapters build
+# of Velox. The scrip expects base dependencies to already be installed.
+#
+# This script is split of from setup-centos9.sh to improve docker caching
```
majetideepak, line 20: This will break Presto C++ builds. Can we avoid that by keeping the install_adapters functions as is inside `setup-centos9.sh` and sourcing them in this file? That way we also keep it consistent across setup-macos.sh and setup-ubuntu.sh.

assignUser:
@majetideepak yes I had an entire green run, due to force push it's a bit hard to track down but since than I only changed things to make the jobs compatible with both images so we don't have issues when merging/rebuilding the images

scripts/setup-centos-adapters.sh
assignUser: That breaks the purpose of the separate file. As it  would invalidate the base image stage and the centos stage,  when changes to the adapters function are made.

majetideepak: I see. Thanks.

majetideepak (approved):

assignUser:
@kgpai rebased :+1:

.github/workflows/docker.yml
```diff
@@ -34,71 +35,136 @@ concurrency:
...
-        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+      - name: Bake Images
+        id: bake
+        uses: docker/bake-action@37816e747588cb137173af99ab33873600c46ea8 # v6.8.0
+        env:
+          DOCKER_UPLOAD_CACHE: ${{ github.repository == 'facebookincubator/velox' && github.event_name != 'pull_request'}}
+        with:
+          targets: ${{ matrix.target }}-${{ matrix.os.platform }}
+          push: ${{ github.repository == 'facebookincubator/velox' && github.event_name != 'pull_request'}}
+          load: ${{  github.event_name == 'pull_request' && matrix.target == 'ci' && matrix.os.platform == 'amd64' }}
```
kgpai, line 82: nit: can you add a comment on why only amd atm ?

.github/workflows/docker.yml
```diff
@@ -34,71 +35,136 @@ concurrency:
...
+        # registry when not running in a PR
+        if: ${{ github.event_name == 'pull_request' && matrix.target == 'ci' && matrix.os.platform == 'amd64' }}
+        run: |
+          # Update tag to what the dependent images expect
+          docker tag "$BASE_NAME/velox-dev:centos9-amd64" "$BASE_NAME/velox-dev:centos9"
+          docker save -o centos9.tar "$BASE_NAME/velox-dev:centos9"
 
-      - name: Build and Push
-        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
+      - name: Upload Base Image
```
kgpai, line 93: nit: Are we only uploading x86 images ?

.github/workflows/scheduled.yml
```diff
@@ -177,8 +178,10 @@ jobs:
         if: ${{ github.event_name != 'schedule' && steps.get-sig.outputs.stash-hit != 'true' }}
         working-directory: velox_main
         run: |
+          # TODO convert to uv with script defined deps after 14046 is merged
           source .venv/bin/activate
-          python3 -m pip install deepdiff
+          PIP=$(command -v uv > /dev/null 2>&1 && echo "uv pip" || echo "python3 -m pip")
+          $PIP install deepdiff
```
kgpai, line 184: nit can we bake deepdiff into the image ?

docker-compose.yml
```diff
@@ -11,77 +11,52 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-version: '3.5'
 
+#TODO document usage
```
kgpai: nit: Can we do this as a part of this change ?

scripts/ci/bm-report/result
```diff
@@ -0,0 +1 @@
+/nix/store/41zr5k0jf1af3dc8fp153433ksjswhgz-nix-shell
```
kgpai: nit: whats this for ?

kgpai (approved):
Some nits..

.github/workflows/docker.yml
assignUser: It's explained in the export step but I'll add a comment to what this does.

scripts/ci/bm-report/result
assignUser: A remnant from building the report locally, have added it to gitignore

docker-compose.yml
assignUser: Oh yeah!

.github/workflows/scheduled.yml
assignUser: Not really because we use a venv here so a system install wouldn't work. Also the install takes 2 seconds with pip and is likely even faster with uv so we wouldn't gain much from it :)

.github/workflows/docker.yml
assignUser: Uploading as an artifact for use in the follow up job to build the presto and spark containers , yes. Publishing to the registry, no, we publish full multi-arch images for cento9, adapters and pyvelox.

PR #14049, opened 2025-07-08
feat: Add nonDecimalNumericTypeOnly constraint to type parameters

pramodsatya:


pramodsatya:
@aditi-pandit, @majetideepak, could you please help review this change?

velox/expression/FunctionSignature.cpp
```diff
@@ -189,17 +189,20 @@ SignatureVariable::SignatureVariable(
     ParameterType type,
     bool knownTypesOnly,
     bool orderableTypesOnly,
-    bool comaprableTypesOnly)
+    bool comaprableTypesOnly,
```
aditi-pandit: Change spelling to comparableTypesOnly.

aditi-pandit (commented):
Thanks @pramodsatya for this code.

Can we add some tests in SignatureBinderTest.cpp and FunctionSignatureBuilderTest as well ?

velox/expression/SignatureBinder.cpp
```diff
@@ -88,6 +88,12 @@ bool checkNamedRowField(
   return true;
 }
 
+bool isNonDecimalNumericType(const TypePtr& type) {
+  return (type == BIGINT() || type == INTEGER() || type == SMALLINT() ||
+          type == TINYINT()) ||
```
czentgr, line 104: Why not just one list of OR? Why are there two blocks? For readability?

pramodsatya: I had used two blocks just to maintain compatibility with the check in Presto, and to make the distinction between integral and floating point types clear. Would it be better to collapse to single OR check?

pramodsatya:
Thanks for the feedback @aditi-pandit @czentgr, added unit tests. Could you please take another look?

aditi-pandit:
@pramodsatya : Please add in your PR description references to nonDecimalNumeric type constraint usage in Presto Java for more context.

Another question : Is map_union_sum(map<varchar, decimal(10,2)>) supported in Velox correctly ? A quick look at the function implementation seems like it will have issues. Do we intend to support this in Velox correctly ? Because then we should let native engine pick it up.

@mbasmanova : Is the intent of map_union_sum in Velox to support decimal values ? Or we will retain the Presto behavior to error on them ? Asking because a quick search if map_union_sum supports decimal values shows that some databases including MySQL support it. (This function is not part of the standard though mostly).

pramodsatya:
> @pramodsatya : Please add in your PR description references to nonDecimalNumeric type constraint usage in Presto Java for more context.
> 
> Another question : Is map_union_sum(map<varchar, decimal(10,2)>) supported in Velox correctly ? A quick look at the function implementation seems like it will have issues. Do we intend to support this in Velox correctly ? Because then we should let native engine pick it up.
> 
> @mbasmanova : Is the intent of map_union_sum in Velox to support decimal values ? Or we will retain the Presto behavior to error on them ? Asking because a quick search if map_union_sum supports decimal values shows that some databases including MySQL support it. (This function is not part of the standard though mostly).

Thanks @aditi-pandit, this constraint is only used by `map_union_sum` in Presto and updated PR description. `map_union_sum(map<varchar, decimal(10,2)>)` is not supported in Velox currently, queries with this function signature run successfully only when the sidecar is enabled in Presto C++, because the signature is resolved to `map_union_sum(map<varchar, real>)` by coercion (adding the `nonDecimalNumeric` constraint will prevent this).

mbasmanova:
@aditi-pandit I would expect map_union_sum to support same type as sum. Why would it not support decimal values? There could be bugs, but I do not see any reason not to support this. Why is Presto Java not supporting that?

In general a change like this requires some design discussion. Would it be possible to create an issue to describe the problem and proposed solution?

CC: @amitkdutta

pramodsatya:
Thanks @mbasmanova, opened an issue to add this support in Presto: https://github.com/prestodb/presto/issues/26659. Closing this PR. cc: @aditi-pandit @czentgr.

PR #14050, opened 2025-07-08
fix: Fix RowContainer setAllNull() setting initialized flags bug

tanjialiang:
Summary: RowContainer::setAllNull() sets the accumulators' initialized bits to 1 as well, falsely marking uninitialized accumulators initialized. This made cleanup procedure to destroy these uninitialized accumulators, causing crash. This diff modified setAllNull() to set only null bits, leaving initialized bits untouched. The diff also corrected the comments stating the bit order of initialized and null to be aligned with the code implementation (that null bit is before initialized bit).

Differential Revision: D77951091

### DIFF
--- a/velox/exec/RowContainer.h
+++ b/velox/exec/RowContainer.h
@@ -328,13 +328,7 @@ class RowContainer {
   /// Sets all fields, aggregates, keys and dependents to null. Used when making
   /// a row with uninitialized keys for aggregates with no-op partial
   /// aggregation.
-  void setAllNull(char* row) {
-    removeOrUpdateRowColumnStats(row, /*setToNull=*/true);
-    if (!nullOffsets_.empty()) {
-      memset(row + nullByte(nullOffsets_[0]), 0xff, initialNulls_.size());
-      bits::clearBit(row, freeFlagOffset_);
-    }
-  }
+  void setAllNull(char* row);
 
   /// The row size excluding any out-of-line stored variable length values.
   int32_t fixedRowSize() const {
@@ -1512,7 +1506,7 @@ class RowContainer {
   // Indicates if this row container has rows with duplicate keys. This only
   // applies if 'nextOffset_' is set.
   tsan_atomic<bool> hasDuplicateRows_{false};
-  // Bit position of null bit  in the row. 0 if no null flag. Order is keys,
+  // Bit position of null bit in the row. 0 if no null flag. Order is keys,
   // accumulators, dependent.
   std::vector<int32_t> nullOffsets_;
   // Position of field or accumulator. Corresponds 1:1 to 'nullOffset_'.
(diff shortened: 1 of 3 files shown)


velox/exec/RowContainer.h
```diff
@@ -1543,9 +1541,11 @@ class RowContainer {
...
   // 0 after deciding not to use normalized keys.
   int normalizedKeySize_;
-  // Copied over the null bits of each row on initialization. Keys are
-  // not null, aggregates are null.
-  std::vector<uint8_t> initialNulls_;
+  // Initial null flags state, with all null and initialized bits unset.
+  std::vector<uint8_t> initialNullFlagsMask_;
+  // The mask to set all null flags, with each corresponding null bit set,
+  // initialized bit unset.
+  std::vector<uint8_t> allNullFlagsMask_;
```
Yuhta: I don't think we need this, let's just do
```c++
void RowContainer::setAllNull(char* row) {
  removeOrUpdateRowColumnStats(row, /*setToNull=*/true);
  if (!nullOffsets_.empty()) {
    for (auto i : nullOffsets_) {
      row[nullByte(i)] |= nullMask(i);
    }
    bits::clearBit(row, freeFlagOffset_);
  }
}
```

velox/exec/RowContainer.h
```diff
@@ -1543,9 +1541,11 @@ class RowContainer {
   // Extra bytes to reserve before  each added row for a normalized key. Set to
   // 0 after deciding not to use normalized keys.
   int normalizedKeySize_;
-  // Copied over the null bits of each row on initialization. Keys are
-  // not null, aggregates are null.
-  std::vector<uint8_t> initialNulls_;
+  // Initial null flags state, with all null and initialized bits unset.
+  std::vector<uint8_t> initialNullFlagsMask_;
```
Yuhta: Is this really used?  It seems just kept all 0, maybe we can change the `memcpy` from this into a `memset` and get rid of this?

velox/exec/RowContainer.h
tanjialiang: nullMask() does not differentiate if it is initialized bit. Note that nullOffsets_ also stores the initialized bits, too. So doing this will also set the initialized bits.

velox/exec/RowContainer.h
tanjialiang: I think we can get rid of initialNulls_(renamed as initialNullFlagsMask_). But for easier null set we might want to keep allNullFlagsMask_

velox/exec/RowContainer.h
Yuhta: I see, let's separate `initializedFlagOffsets_` out from the `nullOffsets_` to make it more readable then (and less error prone than this bit mask approach)

tanjialiang: Yeah actually we don't even need the initialized offsets. We can just not add the init bits to the nulloffsets_ vec

velox/exec/RowContainer.h
tanjialiang: ditto I think we can remove both once we do not add initialized offsets to nulloffsets_

skyelves (approved):
thanks

PR #14052, opened 2025-07-08
fix: getDouble and getFloat

YiChengLee03:
Summary: Fix production verifier mismatch. StringView.data() does not guarantee end with a null terminator but sscanf's behaviour is undefined without a null terminator.

Differential Revision: D77963869


velox/dwio/text/reader/TextReader.cpp
```diff
@@ -888,28 +896,33 @@ TextRowReader::getDouble(TextRowReader& th, bool& isNull, DelimType& delim) {
   if (strView.empty()) {
     isNull = true;
   }
+
   if (isNull) {
-    return 0;
+    return 0.0;
```
zacw7, line 909: why is this needed?

zacw7 (commented):
LG. thanks for figuring this out.

zacw7:
Do we need to add the dependency to CMakeLists.txt?

velox/dwio/text/reader/TextReader.cpp
YiChengLee03: Technically its the same, just thought that its better to keep it consistent across the file when float/double isNull, return 0.0 instead of 0

YiChengLee03:
> Do we need to add the dependency to CMakeLists.txt?

I don't think we need to add another dependency here since TextReader.h imports StringViewBufferHolder.h which is already linked with Folly::folly. The build also seem to be working now.

YiChengLee03:
@zacw7 @Yuhta I realized that there were still some edge cases with parsing integers but found a workaround using ownedString_.c_str(). Could you help check if there will be any performance hit or adhoc string allocations happening?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -189,8 +194,8 @@ uint64_t TextRowReader::next(
   auto reqChildCount = reqT->size();
 
   // create top level RowVector
-  auto rowVecPtr = std::static_pointer_cast<RowVector>(
-      BaseVector::create(reqT->type(), rows, &contents_->pool));
+  auto rowVecPtr = std::static_pointer_cast<RowVector>(BaseVector::create(
```
Yuhta: You can do `create<RowVector>`

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -189,8 +194,8 @@ uint64_t TextRowReader::next(
...
+      reqT->type(), static_cast<vector_size_t>(rows), &contents_->pool));
```
Yuhta: There is no need to case

Yuhta (approved):

velox/dwio/text/reader/TextReader.cpp
YiChengLee03: Casting rows because it is a narrowing conversion from uint64_t to int here. Will Use BaseVector::create<RowVector> without pointer casting.

PR #14058, opened 2025-07-09
refactor: Reduce unneeded copies

jagill:
Summary:
Lint identified several places where we are implicitly copying variables.
Replace these patterns with ones that don't copy.

Also reserve a vector to reduce unnecessary allocations.

Differential Revision: D78016694


xiaoxmeng (approved):
@jagill thanks!

PR #14062, opened 2025-07-09
fix: Cast to JSON has wrong output size estimation for some invalid Unicode input

Yuhta:
Summary: When the input string is not valid Unicode, the size estimation skips all 4 bytes but we should only skip 1 byte and re-parse the input from second byte.  The actual conversion is done correctly, but the size estimation is wrong due to this mismatch.  This is causing out of boundary write and crashes.

Differential Revision: D78028993


xiaoxmeng (approved):
@Yuhta thanks!

Issue #14059, opened 2025-07-09
Spark add_months function result


rui-mo:
### Description

Spark expects an "integer overflow" exception to be thrown for `add_months(add_months(DATE('1922-05-05'), INT('1874490038')), c0)` when c0's value is NULL, but Velox returns NULL.

Use the below test case to repro.

```
std::optional<int32_t> inputValue = std::nullopt;
  VELOX_ASSERT_THROW(
      evaluateOnce<int32_t>(
          "add_months(add_months('1922-05-05'::DATE, 1874490038::INTEGER),c0)",
          {INTEGER()},
          inputValue),
      "Integer overflow");
```

Reasons
1) Constant folding suppresses the user error thrown by`AddMonthsFunction`, and we can change to use `VELOX_CHECK_UNSUPPORTED_INPUT_UNCATCHABLE` to avoid this issue.
https://github.com/facebookincubator/velox/blob/17d3540c8c776014259c6cacf165d86f280c0984/velox/expression/ExprCompiler.cpp#L338-L339

2) In the expression fuzzer test, NULL is still returned with the change mentioned above. The reason is expression verifier will recreate `exec::ExprSet` with constant folding disabled but the results w/o constant folding are different.
https://github.com/facebookincubator/velox/blob/17d3540c8c776014259c6cacf165d86f280c0984/velox/expression/tests/ExpressionVerifier.cpp#L150-L159
**Without** constant folding, an optimization happens and NULL rows are removed, so the integer overflow does not occur and NULL is returned.
https://github.com/facebookincubator/velox/blob/17d3540c8c776014259c6cacf165d86f280c0984/velox/expression/Expr.cpp#L1157
We may modify the fuzzer to get around this problem because Velox has internal constant folding enabled and uses it as the code path for runtime tests.

### Relevant logs

```bash

```

Issue #14047, opened 2025-07-08
Failure of expression fuzzer without Presto sot


kagamiori:
### Description

I saw a failure of Presto expression fuzzer on a PR that makes no change to Presto functions: https://github.com/facebookincubator/velox/actions/runs/16135292999/job/45534413063?pr=13199.

The failure was due to common path throws while it should not.

This failure needs investigation.

### Relevant logs

```bash

```

PR #14064, opened 2025-07-09
feat: Support ROW Type in TextWriter

YiChengLee03:
Summary: Support nested row complex type.

Differential Revision: D77966020

### DIFF
--- a/velox/dwio/text/writer/TextWriter.cpp
+++ b/velox/dwio/text/writer/TextWriter.cpp
@@ -276,7 +276,42 @@ void TextWriter::writeCellValue(
 
       return;
     }
-    case TypeKind::ROW:
+    case TypeKind::ROW: {
+      const RowVector* rowVecPtr = decodedColumnVector->base()->as<RowVector>();
+      const auto& indices = decodedColumnVector->indices();
+      const auto actualRowIndex = indices[row];
+
+      std::vector<std::shared_ptr<DecodedVector>> decodedColumnVectors;
+      const auto numColumns = rowVecPtr->childrenSize();
+      for (size_t column = 0; column < numColumns; ++column) {
+        auto decodedColumnVector =
+            std::make_shared<DecodedVector>(DecodedVector(
+                *rowVecPtr->childAt(column),
+                SelectivityVector(rowVecPtr->size())));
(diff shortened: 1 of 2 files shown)


velox/dwio/text/writer/TextWriter.cpp
```diff
@@ -240,13 +240,37 @@ void TextWriter::writeCellValue(
...
+            depth,
+            nestedRowDelimiter);
+      }
+      return;
+    }
     case TypeKind::MAP:
       [[fallthrough]];
-    case TypeKind::ROW:
-      [[fallthrough]];
-    case TypeKind::UNKNOWN:
```
zacw7: wondering why remove TypeKind::UNKNOWN?

YiChengLee03: We are not supporting TypeKind::UNKNOWN for Reader, so it is defaulted to fallthrough and throw an exception anyway.

zacw7: maybe keep it to make it explicit that `UNKNOWN` is not supported and will fall through to the exception.

YiChengLee03: @zacw7 Sure, should I do the same for FUNCTION, OPAQUE, and INVALID?

kewang1024:
Looks like there is writer fuzzer failing?

YiChengLee03:
> Looks like there is writer fuzzer failing?

This is an odd one. This Fuzzer only calls on DWRF reader [here](https://github.com/facebookincubator/velox/blob/3b2b10dfae23d1260c8e0e410fac88de3a586145/velox/exec/tests/utils/PlanBuilder.h#L706), not TEXTFILE Reader. This failure is likely transient or caused by other PRs but I don't see any issues created on this yet. Will rebase and retry.

Note: There have been issues with Presto as Truth Expression Fuzzer #14047 recently as well.

YiChengLee03:
@kewang1024 Expression Fuzzer (#14047) and Spark Fuzzer(#14059) are known failures.

kewang1024:
@duxiao1212 is helping exposing more error code in the writer fuzzer: https://github.com/facebookincubator/velox/pull/14088

kewang1024 (approved):

PR #14066, opened 2025-07-10
feat(vector): EnsureWritable support for FlatMapVector

pedroerp:
Summary:
Adding proper support for ensureWritable for flat map vectors,
making them writable and copying on write buffers when appropriate.

Reviewed By: kevinwilfong

Differential Revision: D77979157


kevinwilfong (approved):
Thanks!

PR #14068, opened 2025-07-10
fix: Fix the installed version of RE2 to match the bundled version

zml1206:
Fix the installed version of RE2 to match the bundled version. It can also solve gluten build error on macOS:
```
[100%] Linking CXX shared library ../releases/libvelox.dylib
Undefined symbols for architecture arm64:
  "bool re2::re2_internal::Parse<std::__1::basic_string_view<char, std::__1::char_traits<char>>>(char const*, unsigned long, std::__1::basic_string_view<char, std::__1::char_traits<char>>*)", referenced from:
      bool re2::RE2::Arg::DoParse3ary<std::__1::basic_string_view<char, std::__1::char_traits<char>>>(char const*, unsigned long, void*) in libvelox.a[334](Re2Functions.cpp.o)
  "re2::RE2::FullMatchN(std::__1::basic_string_view<char, std::__1::char_traits<char>>, re2::RE2 const&, re2::RE2::Arg const* const*, int)", referenced from:
      facebook::velox::config::toCapacity(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, facebook::velox::config::CapacityUnit) in libvelox.a[34](Config.cpp.o)
      facebook::velox::config::toDuration(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in libvelox.a[34](Config.cpp.o)
      facebook::velox::memory::MemoryPoolImpl::needRecordDbg(bool) in libvelox.a[56](MemoryPool.cpp.o)
      facebook::velox::RowType::printChildren(std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>&, std::__1::basic_string_view<char, std::__1::char_traits<char>>) const in libvelox.a[113](Type.cpp.o)
      facebook::velox::RowType::printChildren(std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>&, std::__1::basic_string_view<char, std::__1::char_traits<char>>) const in libvelox.a[113](Type.cpp.o)
      facebook::velox::functions::PatternMetadata::parseSubstrings(std::__1::basic_string_view<char, std::__1::char_traits<char>> const&) in libvelox.a[334](Re2Functions.cpp.o)
```


scripts/setup-versions.sh
```diff
@@ -30,7 +30,7 @@ SIMDJSON_VERSION="3.9.3"
 CPR_VERSION="1.10.5"
 DOUBLE_CONVERSION_VERSION="v3.1.5"
 RANGE_V3_VERSION="0.12.0"
-RE2_VERSION="2022-02-01"
+RE2_VERSION="2024-07-02"
```
czentgr, line 33: Please update the README with the version: https://github.com/facebookincubator/velox/blob/main/CMake/resolve_dependency_modules/README.md.

czentgr (commented):
Technically this is a fix resolving a version discrepancy between the bundled and installed version of RE2 as opposed to a new feature.
At first I didn't understand what was meant by "Keep the re2 version of macOS consistent with velox".
But it is meant to say "fix the installed version of RE2 to match the bundled version".

scripts/setup-versions.sh
zml1206: Thanks, pr title and README updated.

czentgr:
On a side note. On Linux the Re2 OS packages are used in the setup scripts which means that version of Re2 is likely older than what is provided by the bundle. Re2 is installed on macOS using the setup scripts only.

assignUser:
The [test fail ](https://github.com/facebookincubator/velox/actions/runs/16222934525/job/45807773685?pr=14068#step:10:865) seems a bit suspicious? 
`Expected: (opStats.at("HashBuild").runtimeStats[Operator::kSpillWrites].sum) > (0), actual: 0 vs 0`

zml1206:
@assignUser  Thanks, this is an occasional problem, and it ran successfully after rebase. I don't know why ut failed, but I think it is not related to this PR, because the installed version of RE2 is currently only used for macOS and does not affect workerflow "Ubuntu debug with resolve_dependency".

assignUser (approved):

philo-he:
> On a side note. On Linux the Re2 OS packages are used in the setup scripts which means that version of Re2 is likely older than what is provided by the bundle. Re2 is installed on macOS using the setup scripts only.

@czentgr, yes, older version of RE2 is installed through using re2-devel package. As this package is still used in ubuntu/contos-9 setup script, instead of building from source, does it mean that older version is still compatible? If so, it may cause confusion among users that Minimum Version in the doc has been updated to 2024-07-02 in this PR. cc @zml1206

zml1206:
Close it because it has been merged.

PR #14072, opened 2025-07-10
test: Add coverage for primitive type projection

YiChengLee03:
Summary:
- Test projection for primitive types
- Test projection with swapped columns

Differential Revision: D78102873


zacw7 (approved):
Thanks for adding the test.

PR #14073, opened 2025-07-10
docs: Update Spark scalar function doc to include try_subtract and try_multiply

nimesh1601:
Update Spark scalar function doc to include try_subtract and try_multiply, which are already supported. Added by using velox_sparksql_coverage


nimesh1601:
@rui-mo Can you take a look ?

nimesh1601:
Thanks, @rui-mo, for pointing this out. I was also going through the code to understand it, and your point makes sense. But if we just keep a few try_xxx like try_add, try_divide and don't include others, it gives the perception other try_xxx are not supported. Ideally, as try_xxx itself is not supported by Velox, I think we should remove all try_xxx to avoid that confusion. Does that make sense ?

rui-mo:
> Ideally, as try_xxx itself is not supported by Velox, I think we should remove all try_xxx to avoid that confusion.

@nimesh1601 To remove them makes sense to me. However, the coverage document may not need to be updated for every function. Instead, it will be updated in a bunch. Would you like to track this update in an issue first?

PR #14079, opened 2025-07-10
refactor: Rename TextReader namespace

YiChengLee03:
Summary: Keeping TextReader namespace consistent with readers for other file formats

Differential Revision: D78130504


zacw7 (approved):
LG. Thanks for the refactor.

PR #14081, opened 2025-07-11
refactor: Delete velox/type/TypeParser.h

mbasmanova:
Summary:
velox/type/TypeParser.h is not needed.

It adds confusion since there is another TypeParser.h in velox/type/parse.

Differential Revision: D78159510


velox/type/fbhive/HiveTypeParser.h
```diff
@@ -23,8 +23,7 @@
 #include <vector>
 
 #include <folly/Range.h>
-
-#include "velox/type/TypeParser.h"
+#include "velox/type/Type.h"
```
czentgr, line 26: Question: could you forward declare TypePtr here and then include Type.h only in the HiveTypeParser.cpp?

mbasmanova: We could do that, but that would be a separate change.

pedroerp: Wouldn't that be only two lines of change (remove the include, forward declare it)?

pedroerp (commented):
I guess the intention here that was by having a base class, this could be specialized to parse other non-hive type naming patterns. Do we foresee us having two parse other patterns of types in the future?

mbasmanova:
> I guess the intention here that was by having a base class, this could be specialized to parse other non-hive type naming patterns. Do we foresee us having two parse other patterns of types in the future?

We can't predict the future, but the present state is confusing.

velox/type/fbhive/HiveTypeParser.h
mbasmanova: Not really. This code won't compile. 

```
  template <TokenType KIND, velox::TypeKind TYPEKIND>
  void setupMetadata(std::vector<std::string>&& tokens) {
    static constexpr bool isPrimitive =
        velox::TypeTraits<TYPEKIND>::isPrimitiveType;
    metadata_[makeTokenId(KIND)] = std::make_unique<TokenMetadata>(
        KIND, TYPEKIND, std::move(tokens), isPrimitive);
  }
```

I would like to avoid spending extra time on this now.

PR #14082, opened 2025-07-11
fix: Access null value in makeInFilter

rui-mo:
If the value at idx is null, the access to the underlying bit is undefined 
behavior, and can lead to a crash, invalid memory access, or other 
unpredictable behavior. This PR fixes the access to null value in the
`makeInFilter` by first checking if it is null.

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -106,8 +106,16 @@ std::vector<int64_t>
 toInt64List(const VectorPtr& vector, vector_size_t start, vector_size_t size) {
   auto ints = vector->as<SimpleVector<T>>();
   std::vector<int64_t> values;
-  for (auto i = 0; i < size; i++) {
-    values.push_back(ints->valueAt(start + i));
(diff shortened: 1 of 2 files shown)


rui-mo:
@Yuhta Would you please help review this fix? Thanks.

FelixYBW:
@Yuhta Can you take a look? It fixed the Gluten UT.

rui-mo:
@majetideepak @Yuhta Could you please help review this fix? Thanks!

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -368,31 +392,41 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeInFilter(
   auto elements = arrayVector->elements();
 
   auto elementType = arrayVector->type()->asArray().elementType();
+  bool hasNull = false;
```
Yuhta: Remove

rui-mo: Nice catch. I fixed this, thanks.

PR #14085, opened 2025-07-11
fix: HiveDataSink should materialize the input before any writes

wecharyu:
`HiveDataSink` does not materialize input for unpartitioned and unbucked tables, when writing to parquet table, it would encounter this error:
```bash
function:asUnchecked, Expression: dynamic_cast<const T*>(this) != nullptr Wrong type cast expected N8facebook5velox9MapVectorE, but got N8facebook5velox10LazyVectorE, Source: RUNTIME, ErrorCode: INVALID_STATE
```

The root cause is that the input may be lazy vector, it should be loaded before writing to parquet.

### DIFF
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -474,6 +474,9 @@ bool HiveDataSink::canReclaim() const {
 void HiveDataSink::appendData(RowVectorPtr input) {
   checkRunning();
 
+  // Lazy load all the input columns.
+  input->loadedVector();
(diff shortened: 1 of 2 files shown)


wecharyu:
@majetideepak could you help review this PR? The gluten job would fail before this patch:
```sql
create table tbl1 (id int, v map<string, string>) using parquet;
insert into tbl1 select 1, map('k', 'v');

create table tbl2 (id int, v map<string, string>) using parquet;
insert overwrite tbl2 select * from tbl; -- failure
```

PR #14090, opened 2025-07-12
Reader registry 3

yingsu00:


yingsu00:
@assignUser Hi Jacob, could you please review this PR? Thanks!

yingsu00:
@pedroerp @Yuhta Could you please review this PR? Thanks!

yingsu00:
@pedroerp @Yuhta @mbasmanova @kgpai @assignUser Could you please review this PR? It cleans up the DWIO interface and allows for more test coverage on multiple file formats in the future. For example, we will be able to test multiple file formats with the existing TableScanTest and TableWriterTest, if we make file format a test parameter.

velox/dwio/RegisterReaders.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace facebook::velox::dwio {
+
+void registerReaderFactories() {
+  dwrf::registerDwrfReaderFactory();
+  orc::registerOrcReaderFactory();
+  text::registerTextReaderFactory();
+#ifdef VELOX_ENABLE_PARQUET
+  parquet::registerParquetReaderFactory();
+#endif
+}
```
rui-mo, line 35: Would you mind sharing the specific issue you're experiencing with the separate registrations? I assume Spark SQL typically does not register the DWRF readers and writers by default, and prior to this change, it was possible to register only the necessary components as needed.

yingsu00: @rui-mo Thanks for reviewing! There are a few motivations:

1. We wanted the tests to be able to cover both Dwrf and Parquet in a more elegant way, e.g. TableScanTest, TableWriterTest, HiveDataSinkTest, etc. These tests only work on a specific format, mostly DWRF. These tests, being in exec or core, which are more core engine stuff, should stay file format agnostic. We shall also be able to use these tests to test on Parquet or other file formats. I believe this is a well known design principle and that's why the ReaderFactory and WriterFactory were introduced in the past. But there were many tests not following this paradigm, making reusing these tests for other file formats hard. With the changes in this PR, we can make the tests parameterized with the dwio::common::FileFormat enums like this 
```
class OperatorTestBase : public ::testing::TestWithParam<dwio::common::FileFormat>
                         public velox::test::VectorTestBase  
```
That way we don't have to introduce new tests specifically for Parquet or new file format, but just turning the tests to TEST_P tests that can automatically iterate through the file format collection.

2. It will also help to add more file readers in a modular way. We will need to add AVRO sometime in the future. If the new AVRO reader can folllow the DWIO interface, then it can just be plugged in and the engine automatically works because it's file format agnostic.

3. This will help us to make the Velox Parquet reader shareable with other communities like the Iceberg community.

I added a paragraph in the design doc and the PR message to explain why this is a good thing: https://github.com/facebookincubator/velox/issues/14514

yingsu00: @rui-mo Also, the single file format registry is still supported. So there is no backward compatibility issue for Gluten.

rui-mo: Thanks for explaining the benefits. Regarding the first point, I find `TableScanTest` is in the `velox/exec`, and I agree with you to make it format agnostic.

Yuhta (approved):

velox/dwio/CMakeLists.txt
```diff
@@ -35,5 +35,29 @@ add_subdirectory(common)
...
 add_subdirectory(dwrf)
 add_subdirectory(orc)
-add_subdirectory(parquet)
 add_subdirectory(text)
+
+if(VELOX_ENABLE_PARQUET)
+  add_subdirectory(parquet)
+endif()
+
+add_library(velox_dwio_registered_readers STATIC RegisterReaders.cpp)
```
assignUser, line 44: Why do you force it to be static?

assignUser (commented):
Please resolve conflicts/rebase so CI can run.

velox/dwio/CMakeLists.txt
yingsu00: @assignUser Thanks for reviewing! I recall that it forces CMake to build libvelox_dwio_registered_readers.a and help Linux system to build correctly. It is not an umbrella target (purely transitive deps, no actual objects) where  "INTERFACE" is usually for. It actually does have code in RegisterReaders.cpp, so it would be either STATIC, OBJECT or SHARED. Since we want outside users to link with it so it cannot be "OBJECT" here. It is manual registration, and it is never supposed to be loaded dynamically on its own, so it won't be "SHARED" either. I believe we are not doing dynamic loading anyways, so by default it is still "STATIC", I just explicitly specified it out to make it clear.

assignUser: Sorry, not sure how I missed this before but this should be part of the main library as it's not a utility for tests or fuzzers, so please use our wrapper functions:
- `velox_add_library` without any library type keyword 
- `velox_link_libraries` instead of `target_link_libraries`. 

Sorry, this isn't well documented yet (#14941).

assignUser (changes requested):

PR #14091, opened 2025-07-13
fix: Fix timestamp partition formatting for Presto compatibility

duxiao1212:
Summary:
When the timestamp is used as the hive partition ID, there was a mismatch pattern observed between Velox and Presto: Velox shows p0=2000-02-20 10%3A02%3A40 while Presto shows p0=2000-02-20 10%3A02%3A40.0.

This fix aligns the discrepancy between the two.

Differential Revision: D78245497


kewang1024:
Is the expression fuzzer failure related?

duxiao1212:
> Is the expression fuzzer failure related?

i don't think so, the previous failure was due to  `FIND_FIRST finds NULL as match`.

kewang1024:
Can you add a bit more details in Summary on how presto is behaving and velox is behaving?

PR #14095, opened 2025-07-14
feat: Extend unnest to support left join semantics by producing output for input row which has empty/null unnest value

xiaoxmeng:
Summary: Add option in Unnest plan node to support yield one output row for the input row with all empty or null unnest values. If configured, the output has one additional column to indicate if the output row has empty unnest value or not. If it is empty, then the corresponding unnest columns are all set to null for that row. This is to extend to support left join semantics for unnest operation which only supports inner join semantics. The left join semantics is required for Meta internal AI index join use case.

Differential Revision: D78238081

### DIFF
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -3954,13 +3954,36 @@ class UnnestNode : public PlanNode {
   /// names must appear in the same order as unnestVariables.
   /// @param ordinalityName Optional name for the ordinality columns. If not
   /// present, ordinality column is not produced.
+  /// @param emptyUnnestValueName Optional name for column which indicates an
+  /// output row has empty unnest value or not. If not present, emptyUnnestValue
+  /// column is not provided and the unnest operator also skips producing output
+  /// rows with empty unnest value.
   UnnestNode(
       const PlanNodeId& id,
       std::vector<FieldAccessTypedExprPtr> replicateVariables,
       std::vector<FieldAccessTypedExprPtr> unnestVariables,
       std::vector<std::string> unnestNames,
       std::optional<std::string> ordinalityName,
+      std::optional<std::string> emptyUnnestValueName,
       const PlanNodePtr& source);
+
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
+  UnnestNode(
+      const PlanNodeId& id,
+      std::vector<FieldAccessTypedExprPtr> replicateVariables,
+      std::vector<FieldAccessTypedExprPtr> unnestVariables,
+      std::vector<std::string> unnestNames,
+      std::optional<std::string> ordinalityName,
+      const PlanNodePtr& source)
+      : UnnestNode(
+            id,
+            std::move(replicateVariables),
+            std::move(unnestVariables),
+            std::move(unnestNames),
+            std::move(ordinalityName),
+            std::nullopt,
+            source) {}
+#endif
 
   class Builder {
    public:
@@ -4008,6 +4031,12 @@ class UnnestNode : public PlanNode {
       return *this;
     }
 
+    Builder& emptyUnnestValueName(
+        std::optional<std::string> emptyUnnestValueName) {
+      emptyUnnestValueName_ = std::move(emptyUnnestValueName);
+      return *this;
+    }
+
     std::shared_ptr<UnnestNode> build() const {
       VELOX_USER_CHECK(id_.has_value(), "UnnestNode id is not set");
       VELOX_USER_CHECK(
@@ -4026,6 +4055,7 @@ class UnnestNode : public PlanNode {
           unnestVariables_.value(),
           unnestNames_.value(),
           ordinalityName_,
+          emptyUnnestValueName_,
           source_.value());
     }
 
@@ -4035,6 +4065,7 @@ class UnnestNode : public PlanNode {
     std::optional<std::vector<FieldAccessTypedExprPtr>> unnestVariables_;
     std::optional<std::vector<std::string>> unnestNames_;
     std::optional<std::string> ordinalityName_;
+    std::optional<std::string> emptyUnnestValueName_;
     std::optional<PlanNodePtr> source_;
   };
 
@@ -4072,8 +4103,16 @@ class UnnestNode : public PlanNode {
     return ordinalityName_;
   }
 
-  bool withOrdinality() const {
+  bool hasOrdinality() const {
     return ordinalityName_.has_value();
+  }
+
+  const std::optional<std::string>& emptyUnnestValueName() const {
+    return emptyUnnestValueName_;
+  }
+
+  bool hasEmptyUnnestValue() const {
+    return emptyUnnestValueName_.has_value();
   }
 
   std::string_view name() const override {
@@ -4091,6 +4130,7 @@ class UnnestNode : public PlanNode {
   const std::vector<FieldAccessTypedExprPtr> unnestVariables_;
   const std::vector<std::string> unnestNames_;
   const std::optional<std::string> ordinalityName_;
+  const std::optional<std::string> emptyUnnestValueName_;
   const std::vector<PlanNodePtr> sources_;
   RowTypePtr outputType_;
 };
--- a/velox/exec/Unnest.cpp
+++ b/velox/exec/Unnest.cpp
@@ -41,11 +41,12 @@ Unnest::Unnest(
           operatorId,
           unnestNode->id(),
           "Unnest"),
-      withOrdinality_(unnestNode->withOrdinality()),
+      withOrdinality_(unnestNode->hasOrdinality()),
+      withEmptyUnnestValue_(unnestNode->hasEmptyUnnestValue()),
       maxOutputSize_(
           driverCtx->queryConfig().unnestSplitOutput()
               ? outputBatchRows()
-              : std::numeric_limits<uint32_t>::max()) {
+              : std::numeric_limits<vector_size_t>::max()) {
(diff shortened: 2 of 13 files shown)


mbasmanova (commented):
CC: @rui-mo

velox/core/PlanNode.h
```diff
@@ -4076,6 +4107,14 @@ class UnnestNode : public PlanNode {
     return ordinalityName_.has_value();
   }
 
+  const std::optional<std::string>& emptyUnnestValueName() const {
+    return emptyUnnestValueName_;
+  }
+
+  bool withEmptyUnnestValue() const {
```
mbasmanova: withXxx -> hasXxx for consistency

velox/exec/tests/UnnestTest.cpp
```diff
@@ -285,6 +285,172 @@ TEST_P(UnnestTest, arrayWithOrdinality) {
   assertQuery(params, expectedInDict);
 }
 
+TEST_P(UnnestTest, arrayWithEmptyUnnestValue) {
+  const auto array = vectorMaker_.arrayVectorNullable<int32_t>(
```
mbasmanova: Consider using makeArrayVectorFromJson for readability.

velox/exec/tests/UnnestTest.cpp
```diff
@@ -285,6 +285,172 @@ TEST_P(UnnestTest, arrayWithOrdinality) {
...
+          "hasOrdinality: {}, hasEmptyUnnestValue: {}, input: {}, expected: {}",
+          hasOrdinality,
+          hasEmptyUnnestValue,
+          input->toString(0, input->size()),
+          expected->toString(0, expected->size()));
+    }
+  } testSettings[] = {
+      {false, false, input, expected},
+      {true, false, input, expectedWithOrdinality},
+      {false, true, input, expectedWithEmptyUnestValue},
```
mbasmanova: typo: expectedWithEmptyUnestValue -> ***Unnest***

velox/exec/tests/UnnestTest.cpp
```diff
@@ -285,6 +285,172 @@ TEST_P(UnnestTest, arrayWithOrdinality) {
...
+            3.3,
+            5.5,
+            std::nullopt,
+            std::nullopt,
+            std::nullopt}),
+       makeNullableFlatVector<int32_t>(
+           {1, 2, std::nullopt, 4, 5, 6, std::nullopt, 7, 8, 9})});
+
+  auto expectedWithOrdinality = makeRowVector(
+      {makeNullableFlatVector<double>(
```
mbasmanova: Copy-pasting these long lines of code makes it difficult to maintain the test. Would it be possible to re-use expected->childAt(0) and expected->childAt(1) instead of copying?

mbasmanova (commented):
@xiaoxmeng Thank you for adding this functionality. Still need to read changes in Unnest.cpp. Here are some initial comments.

velox/exec/Unnest.cpp
```diff
@@ -58,9 +59,16 @@ Unnest::Unnest(
   }
   unnestDecoded_.resize(unnestVariables.size());
 
+  column_index_t checkOutputChannel = outputType_->size() - 1;
+  if (withEmptyUnnestValue_) {
+    VELOX_CHECK_EQ(
+        outputType_->childAt(checkOutputChannel--),
```
mbasmanova: nit: perhaps, separate decrementing checkOutputChannel from the check; it is a bit hard to see this change when it appears in the middle of the check

velox/exec/Unnest.cpp
```diff
@@ -162,22 +168,35 @@ RowVectorPtr Unnest::getOutput() {
...
   vector_size_t numInnerRows{0};
   std::optional<vector_size_t> lastInnerRowEnd;
+  bool hasEmptyUnnestValue{false};
   for (auto inputRow = nextInputRow_; inputRow < inputSize; ++inputRow) {
     const bool isFirstRow = (inputRow == nextInputRow_);
-    const vector_size_t remainingInnerRows = isFirstRow
+    vector_size_t remainingInnerRows = isFirstRow
         ? rawMaxSizes_[inputRow] - firstInnerRowStart_
         : rawMaxSizes_[inputRow];
+    if (FOLLY_UNLIKELY(rawMaxSizes_[inputRow] == 0)) {
```
mbasmanova: Do we need FOLLY_UNLIKELY here? I assume it doesn't help with performance here, but it adds code.

velox/exec/Unnest.cpp
```diff
@@ -198,28 +217,57 @@ Unnest::RowRange Unnest::extractRowRange(vector_size_t inputSize) const {
...
 void Unnest::generateRepeatedColumns(
     const RowRange& range,
     std::vector<VectorPtr>& outputs) {
   // Create "indices" buffer to repeat rows as many times as there are elements
   // in the array (or map) in unnestDecoded.
-  auto repeatedIndices = allocateIndices(range.numInnerRows, pool());
-  auto* rawRepeatedIndices = repeatedIndices->asMutable<vector_size_t>();
-  vector_size_t index = 0;
-  VELOX_DCHECK_GT(range.numInputRows, 0);
+  BufferPtr repeatedIndices = allocateIndices(range.numInnerRows, pool());
```
mbasmanova: Why is this change from auto to specific type?

velox/exec/Unnest.cpp
```diff
@@ -198,28 +217,57 @@ Unnest::RowRange Unnest::extractRowRange(vector_size_t inputSize) const {
...
-            rawRepeatedIndices + index, rawRepeatedIndices + index + size, row);
-        index += size;
-      },
-      rawMaxSizes_,
-      firstInnerRowStart_);
+  if (generateEmptyUnnestValue) {
+    range.forEachRow(
+        [&](vector_size_t row, vector_size_t /*start*/, vector_size_t size) {
+          if (FOLLY_UNLIKELY(size == 0)) {
+            rawRepeatedIndices[index++] = row;
```
mbasmanova, line 249: How do we account for these extra rows in allocateIndices(range.numInnerRows, pool()); ? Does range.numInnerRows includes these somehow?

velox/exec/Unnest.cpp
```diff
@@ -284,32 +335,77 @@ const Unnest::UnnestChannelEncoding Unnest::generateEncodingForChannel(
...
+        rawMaxSizes_,
+        firstInnerRowStart_);
+  } else {
+    range.forEachRow(
+        [&](vector_size_t /*row*/, vector_size_t start, vector_size_t size) {
+          if (FOLLY_LIKELY(size > 0)) {
+            std::iota(rawOrdinality, rawOrdinality + size, start + 1);
+            rawOrdinality += size;
+          } else {
+            // Set ordinality to 0 for output row with empty unnest value.
```
mbasmanova, line 363: Why 0? Shouldn't this be null? Either way, let's document this behavior explicitly.

velox/exec/Unnest.cpp
```diff
@@ -284,32 +335,77 @@ const Unnest::UnnestChannelEncoding Unnest::generateEncodingForChannel(
...
 
+  if (!range.hasEmptyUnnestValue) {
+    return BaseVector::createConstant(
+        BOOLEAN(), false, range.numInnerRows, pool());
+  }
+
+  auto emptyVector = BaseVector::create<FlatVector<bool>>(
+      BOOLEAN(), range.numInnerRows, pool());
+  // Set each output row has empty unnest values or not.
+  auto* const rawEmpty = emptyVector->mutableRawValues<uint64_t>();
```
mbasmanova, line 398: It might be cleaner to assume that most rows will be false and allocate values buffer that's 'false' in all rows. Then, flip the value for rows that are empty.

velox/exec/Unnest.cpp
xiaoxmeng: It is a for loop and mostly likely it is not zero. Maybe not removed.

velox/exec/Unnest.cpp
xiaoxmeng: No specific :)

velox/exec/Unnest.cpp
xiaoxmeng: For the non-empty case, the ordinality starts with 1 so we can set to zero to avoid allocating nulls. I add documents there.

velox/exec/tests/UnnestTest.cpp
```diff
@@ -285,6 +285,172 @@ TEST_P(UnnestTest, arrayWithOrdinality) {
   assertQuery(params, expectedInDict);
 }
 
+TEST_P(UnnestTest, arrayWithEmptyUnnestValue) {
```
aditi-pandit, line 288: @xiaoxmeng : Do we have a SQL query equivalent with LEFT outer join which can generate the same results ? It might be simpler to check test results using it with sql validation in assertQuery. 

Would be simpler for us to understand semantics as well.

velox/exec/Unnest.cpp
xiaoxmeng: extractRowRange has taken care of the extra output rows with empty unnest values in this PR at l197

velox/exec/tests/UnnestTest.cpp
xiaoxmeng: The duckDB doesn't have this feature so we can't test with duckDB SQL

xiaoxmeng (commented):
@mbasmanova thanks for the review. Comments addressed.

mbasmanova (approved):
Thanks.

PR #14097, opened 2025-07-14
fix: Use int32_t for bing tile udfs

Sullivan-Patrick:
Summary: These UDFs were using tinyint signature, but were registered as integer in presto java. This caused production issues with bing tile function rollout.

Differential Revision: D78198453


kgpai:
Bias signature failure is expected since signature was modified, if instead we added a new one, it would have succeeded.

PR #14099, opened 2025-07-14
feat: Implement name interface for TpchConnector

Zhiying12:
Summary: The presto server extends support for the tpch connector, where it can read the catalog config and register its connector. The tests are also included in a similar way as other connectors.

Differential Revision: D78260713


kewang1024:
The title format is failing

kewang1024 (approved):

PR #14100, opened 2025-07-14
fix: Make BingTile type non-orderable

Sullivan-Patrick:
Summary:
BingTile type is not orderable in presto, this diff ensures the same for velox BingTile type. The discrepancy caused some fuzzing errors (https://github.com/facebookincubator/velox/actions/runs/16262347645/job/45913071254?pr=14091)

We may want to add ordering for BingTile in the future, but for now we should aim to achieve the same behavior as presto.

Differential Revision: D78291421

### DIFF
--- a/velox/functions/prestosql/BingTileFunctions.h
+++ b/velox/functions/prestosql/BingTileFunctions.h
@@ -22,6 +22,20 @@
 
 namespace facebook::velox::functions {
 
+FOLLY_ALWAYS_INLINE Status checkBingTileZoom(int32_t zoom) {
+  if (FOLLY_UNLIKELY(zoom < 0)) {
+    return Status::UserError(
+        fmt::format("Bing tile zoom {} cannot be negative", zoom));
+  }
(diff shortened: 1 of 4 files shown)


kgpai (approved):

velox/functions/prestosql/BingTileFunctions.h
```diff
@@ -39,9 +53,10 @@ struct BingTileFunction {
       return Status::UserError(
           fmt::format("Bing tile Y coordinate {} cannot be negative", y));
     }
-    if (FOLLY_UNLIKELY(zoom < 0)) {
-      return Status::UserError(
-          fmt::format("Bing tile zoom {} cannot be negative", zoom));
+
+    auto zoomCheck = checkBingTileZoom(zoom);
+    if (FOLLY_UNLIKELY(!zoomCheck.ok())) {
+      return std::move(zoomCheck);
```
majetideepak, line 59: This is causing the fuzzer builds to fail.
https://github.com/facebookincubator/velox/actions/workflows/scheduled.yml

PR #14104, opened 2025-07-15
fix(gcs): Optional config hive.gcs.endpoint is accessed when not set

PingLiuPing:
When gcs config properties `hive.gcs.endpoint` is not present, query or insert from gcs leads to following error.

prestodb.exceptions.PrestoQueryError: PrestoQueryError(type=INTERNAL_ERROR, name=GENERIC_INTERNAL_ERROR, message=" Operator::addInput failed for [operator: TableWrite, plan node ID: 5]: Empty Optional cannot be unwrapped", query_id=20250714_062014_00309_swbg9)

### DIFF
--- a/velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
+++ b/velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
@@ -30,6 +30,8 @@ namespace facebook::velox {
 // upload this buffer with zero copies if possible.
 auto constexpr kUploadBufferSize = 256 * 1024;
 
+constexpr const char* kGcsDefaultCacheKeyPrefix = "gcs-default-key";
+
 namespace {
 constexpr const char* kSep{"/"};
 constexpr std::string_view kGcsScheme{"gs://"};
--- a/velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.cpp
@@ -16,6 +16,7 @@
 
 #ifdef VELOX_ENABLE_GCS
 #include "velox/common/config/Config.h"
+#include "velox/connectors/hive/HiveConfig.h"
 #include "velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.h" // @manual
 #include "velox/connectors/hive/storage_adapters/gcs/GcsUtil.h" // @manual
 #include "velox/dwio/common/FileSink.h"
@@ -46,7 +47,9 @@ gcsFileSystemGenerator() {
         setBucketAndKeyFromGcsPath(file, bucket, object);
         auto cacheKey = fmt::format(
             "{}-{}",
-            properties->get<std::string>("hive.gcs.endpoint").value(),
+            properties->get<std::string>(
+                connector::hive::HiveConfig::kGcsEndpoint,
+                kGcsDefaultCacheKeyPrefix),
             bucket);
 
         // Check if an instance exists with a read lock (shared).
--- a/velox/connectors/hive/storage_adapters/gcs/tests/GcsEmulator.h
+++ b/velox/connectors/hive/storage_adapters/gcs/tests/GcsEmulator.h
@@ -22,6 +22,7 @@
 #include "gtest/gtest.h"
 
 #include "velox/common/config/Config.h"
+#include "velox/connectors/hive/HiveConfig.h"
 #include "velox/connectors/hive/storage_adapters/gcs/GcsUtil.h"
 #include "velox/exec/tests/utils/PortUtil.h"
 
@@ -92,7 +93,7 @@ class GcsEmulator : public testing::Environment {
       const std::unordered_map<std::string, std::string> configOverride = {})
       const {
     std::unordered_map<std::string, std::string> config(
-        {{"hive.gcs.endpoint", endpoint_}});
+        {{connector::hive::HiveConfig::kGcsEndpoint, endpoint_}});
 
     // Update the default config map with the supplied configOverride map
     for (const auto& [configName, configValue] : configOverride) {
--- a/velox/connectors/hive/storage_adapters/gcs/tests/GcsFileSystemTest.cpp
+++ b/velox/connectors/hive/storage_adapters/gcs/tests/GcsFileSystemTest.cpp
@@ -18,6 +18,7 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/common/file/File.h"
 #include "velox/connectors/hive/storage_adapters/gcs/GcsUtil.h"
+#include "velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.h"
 #include "velox/connectors/hive/storage_adapters/gcs/tests/GcsEmulator.h"
 #include "velox/exec/tests/utils/TempFilePath.h"
 
@@ -249,5 +250,27 @@ TEST_F(GcsFileSystemTest, credentialsConfig) {
   VELOX_ASSERT_THROW(
       gcfs.openFileForRead(gcsFile), "Invalid ServiceAccountCredentials");
 }
+
+TEST_F(GcsFileSystemTest, defaultCacheKey) {
+  registerGcsFileSystem();
+  std::unordered_map<std::string, std::string> configWithoutEndpoint = {};
+  auto hiveConfigDefault = std::make_shared<const config::ConfigBase>(
+      std::move(configWithoutEndpoint));
+  const auto gcsFile1 = gcsURI(
+      emulator_->preexistingBucketName(), emulator_->preexistingObjectName());
+  // FileSystem should be cached by the default key.
+  auto defaultGcs = filesystems::getFileSystem(gcsFile1, hiveConfigDefault);
+
+  std::unordered_map<std::string, std::string> configWithEndpoint = {
+      {connector::hive::HiveConfig::kGcsEndpoint, kGcsDefaultCacheKeyPrefix}};
+  auto hiveConfigCustom =
+      std::make_shared<const config::ConfigBase>(std::move(configWithEndpoint));
+  const auto gcsFile2 = gcsURI(emulator_->preexistingBucketName(), "dummy.txt");
+  auto customGcs = filesystems::getFileSystem(gcsFile2, hiveConfigCustom);
+  // The same FileSystem should be cached by the value of key
+  // kGcsDefaultCacheKeyPrefix.
+  ASSERT_EQ(customGcs, defaultGcs);
+}
+
 } // namespace
 } // namespace facebook::velox::filesystems


velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.cpp
```diff
@@ -44,10 +44,14 @@ gcsFileSystemGenerator() {
         std::string bucket;
         std::string object;
         setBucketAndKeyFromGcsPath(file, bucket, object);
-        auto cacheKey = fmt::format(
-            "{}-{}",
-            properties->get<std::string>("hive.gcs.endpoint").value(),
-            bucket);
+
+        std::string cacheKey = bucket;
+        if (properties->valueExists("hive.gcs.endpoint")) {
```
czentgr: Can you please make "hive.gcs.endpoint" a constant? It is used multiple times.
If "hive.gcs.endpoint" is not present we need to set the default endpoint.

Can you also please add to the unit test to check the default is set correctly?

majetideepak: nit: improvement by removing the default initialization.
```
std::string cacheKey;
if (properties->valueExists(kGcsEndpoint)) {
   cacheKey = fmt::format(
              "{}-{}",
              properties->get<std::string>(kGcsEndpoint).value(),
              bucket);
} else {
  cacheKey = bucket;
}
```

PingLiuPing: Thanks @czentgr , what should be the correct default endpoint? 
In this code change I use the input bucket as the default endpoint. I see similar logic from S3.

PingLiuPing: > nit: improvement by removing the default initialization.
> 
> ```
> std::string cacheKey;
> if (properties->valueExists(kGcsEndpoint)) {
>    cacheKey = fmt::format(
>               "{}-{}",
>               properties->get<std::string>(kGcsEndpoint).value(),
>               bucket);
> } else {
>   cacheKey = bucket;
> }
> ```

Thanks @majetideepak , I will change it back, I originally use this format.

majetideepak: There is no default endpoint. The GCS library takes care of the default endpoint.
We could add a unit test here without the endpoint to test the registration.
See `TEST_F(S3FileSystemRegistrationTest, cacheKey)` for reference.

PingLiuPing: Ok, I will just check the existence of the config key, and not adding default value.

PingLiuPing:
@czentgr  @majetideepak  Can you please have a look again?
The default `entry` is not been used in GCS library, it is just a key to store the FileSystem.

velox/common/config/tests/ConfigTest.cpp
```diff
@@ -157,6 +157,10 @@ TEST_F(ConfigTest, immutableConfig) {
   ASSERT_TRUE(config->get<int32_t>(TestConfig::kInt32Entry.key).has_value());
   ASSERT_EQ(config->get<int32_t>(TestConfig::kInt32Entry.key).value(), -3200);
   ASSERT_FALSE(config->get<int32_t>("wrong_int32_key").has_value());
+  ASSERT_EQ(config->get<int32_t>("wrong_int32_key", 100), 100);
```
JkSelf: Could you clarify why this test is necessary in the current PR? I might have overlooked something. Thanks.

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
```diff
@@ -30,6 +30,9 @@ namespace facebook::velox {
 // upload this buffer with zero copies if possible.
 auto constexpr kUploadBufferSize = 256 * 1024;
 
+constexpr const char* kGcsEndpoint = "hive.gcs.endpoint";
```
JkSelf: The kGcsEndpoint seems to be already defined in the following location: https://github.com/facebookincubator/velox/blob/876c49cc57b801386484659991e640d427fd48cc/velox/connectors/hive/HiveConfig.h#L62. Can we use this existing definition?

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
```diff
@@ -204,6 +205,26 @@ TEST_F(GcsMultipleEndpointsTest, baseEndpoints) {
...
 
+TEST_F(GcsMultipleEndpointsTest, defaultCacheKey) {
+  std::unordered_map<std::string, std::string> configWithoutEndpoint = {};
+  auto hiveConfigDefault = std::make_shared<const config::ConfigBase>(
+      std::move(configWithoutEndpoint));
+  const auto gcsFile1 = gcsURI(
+      gcsEmulatorOne_->preexistingBucketName(),
+      gcsEmulatorOne_->preexistingObjectName());
+  // FileSystem should be cached by the default key
+  auto defaulGcs = filesystems::getFileSystem(gcsFile1, hiveConfigDefault);
```
JkSelf: nit: `defaulGcs  ` to `defaultGcs `

velox/common/config/tests/ConfigTest.cpp
PingLiuPing: Thanks @JkSelf , I don't find any case to cover the default value in `get` method. So, I added two here.

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
PingLiuPing: Thanks, I notice this too, I'm not sure if it is reasonable to include header from connectors layer to storage_adpaters.

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
PingLiuPing: Probably should be Ok, HiveConfig been used from storage_adapters/gcs/tests/ already.

velox/common/config/tests/ConfigTest.cpp
majetideepak: The get method is from the HiveConfig class and is tested elsewhere `velox/connectors/hive/tests/HiveConfigTest.cpp` We can remove this.

velox/common/config/tests/ConfigTest.cpp
```diff
@@ -157,6 +157,10 @@ TEST_F(ConfigTest, immutableConfig) {
...
+  ASSERT_EQ(
```
majetideepak: Remove this as well.

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
```diff
@@ -30,6 +30,8 @@ namespace facebook::velox {
 // upload this buffer with zero copies if possible.
 auto constexpr kUploadBufferSize = 256 * 1024;
 
+constexpr const char* kGcsDefaultEndpoint = "https://storage.googleapis.com";
```
majetideepak: Default is handled by the GCS library.

PingLiuPing: Thanks, the default entry is not been used in GCS library, it is just a key to store the FileSystem.

majetideepak: This is a valid endpoint value. If the library changes the default in the future, then this will conflict. We can keep the semantics clean by adding `constexpr const char* kGcsDefaultCacheKeyPrefix = "gcs-default-key"` instead.

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <folly/executors/IOThreadPoolExecutor.h>
 #include <folly/init/Init.h>
+#include "velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.h"
```
czentgr: Move this to the next block where the other velox includes are (in alphabetical order).

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
```diff
@@ -204,6 +205,26 @@ TEST_F(GcsMultipleEndpointsTest, baseEndpoints) {
...
+  auto defaultGcs = filesystems::getFileSystem(gcsFile1, hiveConfigDefault);
+
+  std::unordered_map<std::string, std::string> configWithEndpoint = {
+      {connector::hive::HiveConfig::kGcsEndpoint, kGcsDefaultEndpoint}};
+  auto hiveConfigCustom =
+      std::make_shared<const config::ConfigBase>(std::move(configWithEndpoint));
+  const auto gcsFile2 =
+      gcsURI(gcsEmulatorOne_->preexistingBucketName(), "dummy.txt");
+  auto customGcs = filesystems::getFileSystem(gcsFile2, hiveConfigCustom);
+  ASSERT_EQ(defaultGcs, customGcs);
```
czentgr: What is the expected vs the test? The "customGcs" is the expected one because we set it up with the explicit (default) endpoint?

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
PingLiuPing: Thanks, changed.

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
PingLiuPing: Thanks, good catch.

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
PingLiuPing: Thanks, this is strange though, it should be formatted by clang already.
Changed.

velox/connectors/hive/storage_adapters/gcs/GcsUtil.h
PingLiuPing: @czentgr  I have change the type back because the test is failed. Seems the get method doesn't work well with string_view.

czentgr: @PingLiuPing Did you push the changes correctly? I don't see the updates to the name/value and other changes.

PingLiuPing: @czentgr  It is correct now.

velox/connectors/hive/storage_adapters/gcs/tests/GcsMultipleEndpointsTest.cpp
```diff
@@ -204,6 +205,26 @@ TEST_F(GcsMultipleEndpointsTest, baseEndpoints) {
...
+  auto defaultGcs = filesystems::getFileSystem(gcsFile1, hiveConfigDefault);
+
+  std::unordered_map<std::string, std::string> configWithEndpoint = {
+      {connector::hive::HiveConfig::kGcsEndpoint, kGcsDefaultCacheKeyPrefix}};
+  auto hiveConfigCustom =
+      std::make_shared<const config::ConfigBase>(std::move(configWithEndpoint));
+  const auto gcsFile2 =
+      gcsURI(gcsEmulatorOne_->preexistingBucketName(), "dummy.txt");
+  auto customGcs = filesystems::getFileSystem(gcsFile2, hiveConfigCustom);
+  ASSERT_EQ(customGcs, defaultGcs);
```
czentgr: This is testing that the cache is correct looked up by providing the same (default) value for the endpoint in the customGcs?
Maybe add a comment to that effect?

Also this test, what was the reason to add it here instead of GcsFileSystemTest? Seems it would fit there a bit better?

PingLiuPing: Ok, let me try again, Initially I add this case to GcsFileSystemTest, but I get compiler error said that the parquet related symbol cannot be found.

PingLiuPing: To move this case to GcsFileSystemTest, need to call `registerGcsFileSystem();` to register filesystem first.

majetideepak (approved):

majetideepak:
@czentgr, @JkSelf  Any other feedback?

czentgr (approved):
Looks good now.

JkSelf (approved):
LGTM. Thanks for your fix.

PR #14108, opened 2025-07-15
rework the parquet connector registration

zoltan:
there is a draft PR open on the presto side as well to bring in the cudf parquet connector. this is the velox side.

I'd prefer to have this here so that PR can be built with this PR together until we figure out how to completely rework it.


majetideepak:
We will use the CudfHiveConnector. Closing this.

PR #14112, opened 2025-07-16
fix: Add Spark abs function

rui-mo:
After 7316a8c, the overflow behavior of Presto abs function does not comply 
with Spark ANSI off mode. This PR fixes this by adding the Spark abs function.

### DIFF
--- a/velox/functions/sparksql/tests/ArithmeticTest.cpp
+++ b/velox/functions/sparksql/tests/ArithmeticTest.cpp
@@ -218,6 +218,11 @@ class ArithmeticTest : public SparkFunctionBaseTest {
   }
 
   template <typename T>
+  std::optional<T> abs(const std::optional<T> input) {
+    return evaluateOnce<T>("abs(c0)", input);
+  }
+
+  template <typename T>
   void assertErrorForCheckedArithmetic(
       const std::string& func,
       const std::optional<T> a,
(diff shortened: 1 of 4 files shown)


zhouyuan:
Cc @zhli1142015 @jinchengchenghh PTAL

rui-mo:
@jinchengchenghh Could you help mark as ready-to-merge if no further comments? Thanks.

velox/functions/sparksql/tests/ArithmeticTest.cpp
```diff
@@ -729,6 +734,33 @@ TEST_F(ArithmeticTest, checkedDivide) {
...
+  EXPECT_EQ(abs<int8_t>(-127), 127);
+  EXPECT_EQ(
+      abs<int8_t>(std::numeric_limits<int8_t>::min()),
+      std::numeric_limits<int8_t>::min());
+  EXPECT_EQ(abs<int16_t>(-32767), 32767);
+  EXPECT_EQ(
+      abs<int16_t>(std::numeric_limits<int16_t>::min()),
+      std::numeric_limits<int16_t>::min());
+  EXPECT_EQ(abs<int32_t>(-2147483647), 2147483647);
+  EXPECT_EQ(
```
bikramSingh91, line 747: not sure why tests are not failing in CI on github, but this test case is throwing an error internally as 2147483648 is outside the range of int32:
```
/velox/functions/sparksql/Arithmetic.h:39:23: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
```
@rui-mo can you please take a look?

rui-mo: @bikramSingh91 Thanks for the pointer. I fixed this, and would you please help re-import?

rui-mo:
The fuzzer failures are not relevant to the Spark abs function. Opened https://github.com/facebookincubator/velox/issues/14199 and  https://github.com/facebookincubator/velox/issues/14200 to track.

velox/functions/sparksql/tests/ArithmeticTest.cpp
bikramSingh91: @rui-mo This is still failing. The issue is that int32 has a range of -2,147,483,648 to 2,147,483,647 with`std::numeric_limits<int32_t>::min()` = -2,147,483,648 . Therefore the abs value of this cannot be represented in an int32. You will either have to remove this test case or upcast to int64

rui-mo: HI @bikramSingh91, I really appreciate your help. Since `abs(-2147483648)` in Spark returns itself (`-2147483648`), this PR aims to align Velox behavior with Spark, and this test is added to ensure that.
To fix the internal issue, I added special handling for the min value, and I assume the internal issue was caused by the undefined behavior of `std::abs`. Could you please help re-import this PR? Apology for any inconvenience.

PR #14113, opened 2025-07-16
feat: Add initial support for coercions

mbasmanova:
Summary:
Coersion is an implicit type conversion that the query engine may apply to resolve a function call. Given a function that takes a bigint and a function call that passes an integer, the engine may apply an implicit cast from integer to bigint to compile that function call.

Given a function `foo: (bigint) -> bigint`, and an expression `foo(1::int)`, the query engine applies coersion to produce `foo(cast(1::int as bigint))`.

Velox doesn't support coersions because it expects the application to fully resolve types before calling Velox APIs. However, it is useful to provide coersion support as a building block for applications to use if they choose to.

This change introduces velox::resolveFunctionWithCoersions API to resolve functions with coersions.

```
/// Like 'resolveFunction', but with support for applying type conversions if no
/// signature matches 'argTypes' exactly.
///
/// param coersions A list of optional type coersions that were applied to
/// resolve the function successfully. Contains one entry per argument. The
/// entry is null if no coersion is required for that argument. The entry is not
/// null if coersions is necessary.
///
/// Example, given functin plus(bigint, bigint) -> bigint and arguments
/// (integer, bigint), returns bigint with coersions = {bigint, null}. The first
/// argument needs to be coersed to bigint, while the second argument doesn't
/// require coersion.
///
/// TODO: Add support for coersion for complex types, signatures with generic
/// types and variadic arguments.
TypePtr resolveFunctionWithCoersions(
    const std::string& functionName,
    std::vector<TypePtr>& coersions);
```

Supported coersions:

- TINYINT -> SMALLINT, INTEGER, BIGINT, REAL, DOUBLE
- SMALLINT -> INTEGER, BIGINT, REAL, DOUBLE
- INTEGER -> REAL, DOUBLE
- BIGINT -> DOUBLE
- REAL -> DOUBLE

If there is a signature that matches input types exactly, that signature is preferred.

If there are multiple signatures that match input types with coersions, the signature with the lowers "coersion cost" is preferred. 

Each coersion is assigned a cost: 
- TINYINT -> SMALLINT is 1
- TINYINT -> INTEGER is 2
- TINYINT -> BIGINT is 3
etc.

Overall cost for a list of coersions is computed as a sum of individual costs.

In case of a tie, when 2+ signatures have the same lowest cost, the call is considered ambiguous and type resolution fails.

Coersion rules are complex. We'll build them out incrementally adding functionality as the need arises. Check out C++ rules for coersions to get a feel for what they may look like: https://en.cppreference.com/w/cpp/language/implicit_conversion.html

Pre-existing wrinkle that affects functions present in both simple and vector registries and not being addressed here:
- Function signature selection is not consistent between simple and vector functions. Simple functions choose a signature with lowest priority, while vector functions pick first match.
- Type resolution looks at simple and vector function signatures separately. It picks simple signature with lowest priority among other simple signatures without considering priority of vector signature.

Differential Revision: D78251314

### DIFF
--- a/velox/expression/VectorFunction.cpp
+++ b/velox/expression/VectorFunction.cpp
@@ -88,6 +88,58 @@ TypePtr resolveVectorFunction(
   return nullptr;
 }
 
+namespace {
+bool hasCoercion(const std::vector<Coercion>& coercions) {
+  for (const auto& coercion : coercions) {
+    if (coercion.type != nullptr) {
+      return true;
+    }
+  }
+
+  return false;
+}
+} // namespace
+
+TypePtr resolveVectorFunctionWithCoercions(
+    const std::string& functionName,
+    const std::vector<TypePtr>& argTypes,
+    std::vector<TypePtr>& coercions) {
+  coercions.clear();
+
+  auto optionalType = applyToVectorFunctionEntry<TypePtr>(
+      functionName,
+      [&](const auto& /*name*/, const auto& entry) -> std::optional<TypePtr> {
+        std::vector<std::pair<std::vector<Coercion>, TypePtr>> candidates;
+        for (const auto& signature : entry.signatures) {
+          exec::SignatureBinder binder(*signature, argTypes);
+          std::vector<Coercion> requiredCoercions;
+          if (binder.tryBindWithCoercions(requiredCoercions)) {
+            auto type = binder.tryResolveReturnType();
+            if (!hasCoercion(requiredCoercions)) {
+              coercions.resize(argTypes.size(), nullptr);
+              return type;
+            }
+
+            candidates.emplace_back(requiredCoercions, type);
+          }
+        }
+
+        if (auto index = Coercion::pickLowestCost(candidates)) {
+          const auto& requiredCoercions = candidates[index.value()].first;
+          coercions.reserve(requiredCoercions.size());
+          for (const auto& coercion : requiredCoercions) {
+            coercions.push_back(coercion.type);
+          }
+
+          return candidates[index.value()].second;
+        }
+
+        return std::nullopt;
+      });
+
+  return optionalType.value_or(nullptr);
+}
+
 std::optional<std::pair<TypePtr, VectorFunctionMetadata>>
 resolveVectorFunctionWithMetadata(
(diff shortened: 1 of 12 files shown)


oerling (approved):

aditi-pandit:
@mbasmanova : Thanks for this code. Have a small nit, but it could be a bit of a usability thing... The spelling is usually coer**C**ion (c instead of your usage of s). Have you used coersion intentionally ?

mbasmanova:
@aditi-pandit Sorry for the typo. Thank you for catching. Will fix.

velox/expression/VectorFunction.cpp
```diff
@@ -88,6 +88,58 @@ TypePtr resolveVectorFunction(
...
+            coersions.push_back(coersion.type);
+          }
+
+          return candidates[index.value()].second;
+        }
+
+        return std::nullopt;
+      });
+
+  return optionalType.has_value() ? optionalType.value() : TypePtr(nullptr);
```
xiaoxmeng:
```
nit: optionalType.valueOr(nullptr);
```

velox/expression/VectorFunction.cpp
```diff
@@ -88,6 +88,58 @@ TypePtr resolveVectorFunction(
...
+      functionName,
+      [&](const auto& /*name*/, const auto& entry) -> std::optional<TypePtr> {
+        std::vector<std::pair<std::vector<Coersion>, TypePtr>> candidates;
+        for (const auto& signature : entry.signatures) {
+          exec::SignatureBinder binder(*signature, argTypes);
+          std::vector<Coersion> requiredCoersions;
+          if (binder.tryBindWithCoersions(requiredCoersions)) {
+            auto type = binder.tryResolveReturnType();
+            if (!hasCoersion(requiredCoersions)) {
+              coersions.resize(argTypes.size());
```
xiaoxmeng:
```
nit: how about make it clear here: coersions.resize(argTypes.size(), nullptr);
```

xiaoxmeng (approved):
@mbasmanova LGTM % nits. Thanks!

PR #14117, opened 2025-07-16
feat: Implement noisy_approx_sfm function

oliver6782:
Summary:
This diff introduces two new aggregate functions based on the SFM sketch. Reference of the function can be found at https://prestodb.io/docs/current/functions/noisy.html#approximate-distinct-counting-sketching

*  `noisy_approx_distinct_sfm(col, epsilon[, buckets[, precision]]) -> bigint()`
* `noisy_approx_set_sfm(col, epsilon[, buckets[, precision]]) -> SfmSketch()`

Differential Revision: D77810376


velox/docs/functions/presto/aggregate.rst
```diff
@@ -809,6 +809,32 @@ Counts, Sums, and Averages
     If provided, random_seed is used to seed the random number generator.
     Otherwise, noise is drawn from a secure random.
 
+.. function:: noisy_approx_set_sfm(col, epsilon[, buckets[, precision]]) -> SfmSketch
```
mbasmanova, line 812: Why do we need _sfm suffix for these functions? Seems noisy_ prefix is sufficient.

velox/functions/prestosql/aggregates/tests/NoisyApproxSfmAggregationTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+       makeConstant<double>(8.0, numElements),
+       makeConstant<int64_t>(8192, numElements)});
+
+  auto result = AssertQueryBuilder(
+                    PlanBuilder()
+                        .values({vectors})
+                        .singleAggregation(
+                            {}, {"noisy_approx_distinct_sfm(c0, c1, c2)"}, {})
+                        .planNode(),
+                    duckDbQueryRunner_)
```
mbasmanova: No need to pass duckDbQueryRunner_ since you are not using DuckDB to produce expected results.

velox/functions/prestosql/aggregates/tests/NoisyApproxSfmAggregationTest.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+  auto serializedView =
+      result->childAt(0)->asFlatVector<StringView>()->valueAt(0);
+  auto deserialized =
+      SfmSketch::deserialize(serializedView.data(), &allocator_);
+  ASSERT_NEAR(deserialized.cardinality(), numElements, numElements * 0.25);
+}
+
+TEST_F(NoisyApproxSfmAggregationTest, distinctBuckets) {
+  vector_size_t numElements = 100000;
```
mbasmanova: use 100'000 for readability

velox/docs/functions/presto/aggregate.rst
oliver6782: We typically add the technique or algorithm for adding noise at the end. The suffix "sfm" helps differentiate this function from other approximate distinct counting functions that may use different sketch algorithms or noise mechanisms. We also have functions such as noisy_count_gaussian which uses gaussian distribution to add noise, perhaps we'll have noisy_count_laplace in the future.

mbasmanova: I see. Wondering if it would be better to pass algorithm as an optional argument instead of having it in the name.

oliver6782: Thanks for the suggestion! I agree that both approaches have their trade-offs. Since we're currently supporting only one algorithm, using a suffix in the function name keeps things simple and explicit, and it really doesn't affect the code right now. If we introduce more algorithms down the line, we might revisit this with Presto team and weigh the benefits of using an optional parameter versus maintaining descriptive function names.

kgpai:
Can you look into the Git CI failures ?

kgpai:
Aggregate fuzzer fails because functions are part of skip list which thus means after application of only there are no signatures for it to run against and hence fails. I am going to fix this so that it ignores function signatures from only if they are in the skip list.

kgpai (approved):

PR #14123, opened 2025-07-17
refactor: Extract vector-to-variant logic from QueryAssertions.cpp

mbasmanova:
Differential Revision: D78457314


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #14131, opened 2025-07-17
feat(cudf): Update to cuDF 25.08 release

mhaseeb123:
This PR updates RAPIDS repo dependencies to 25.08 release version in order to pull in some important bug fixes and features.

### DIFF
--- a/CMake/resolve_dependency_modules/cudf.cmake
+++ b/CMake/resolve_dependency_modules/cudf.cmake
@@ -27,7 +27,7 @@ velox_resolve_dependency_url(rapids_cmake)
 
 set(VELOX_rmm_VERSION 25.08)
 set(VELOX_rmm_BUILD_SHA256_CHECKSUM
-  f2d7a64a3dcfe9b49231375b6df72b3be8a10e0828404463fd5176a9d75a1d4f)
+    f2d7a64a3dcfe9b49231375b6df72b3be8a10e0828404463fd5176a9d75a1d4f)
 set(VELOX_rmm_SOURCE_URL
     "https://github.com/rapidsai/rmm/archive/29dd32302eb7c3e16fb837a1cfe4baac98071512.tar.gz"
 )
--- a/CMake/resolve_dependency_modules/cudf.cmake
+++ b/CMake/resolve_dependency_modules/cudf.cmake
@@ -45,23 +45,6 @@ set(VELOX_cudf_VERSION
     25.08
     CACHE STRING "cudf version")
 
-# cudf 25.08+ requires C++20
-if(VELOX_cudf_VERSION VERSION_GREATER_EQUAL "25.08")
-  set(VELOX_CUDF_CXX_STANDARD
-      20
-      CACHE STRING "C++ standard for cudf modules")
-  set(VELOX_CUDF_CXX_STANDARD_REQUIRED
-      ON
-      CACHE BOOL "C++ standard required for cudf modules")
-else()
-  set(VELOX_CUDF_CXX_STANDARD
-      ${CMAKE_CXX_STANDARD}
-      CACHE STRING "C++ standard for cudf modules")
-  set(VELOX_CUDF_CXX_STANDARD_REQUIRED
-      ON
-      CACHE BOOL "C++ standard required for cudf modules")
-endif()
-
 set(VELOX_cudf_BUILD_SHA256_CHECKSUM
     01d8bd30e8b953b97c71adb2bfc9d83be440b3df6f822f92af95a471ebf001da)
 set(VELOX_cudf_SOURCE_URL
@@ -110,11 +93,6 @@ FetchContent_MakeAvailable(cudf)
 target_compile_options(
   cudf PRIVATE -Wno-non-virtual-dtor -Wno-missing-field-initializers
                -Wno-deprecated-copy -Wno-restrict)
-
-set_target_properties(
-  cudf
-  PROPERTIES CXX_STANDARD ${VELOX_CUDF_CXX_STANDARD}
-             CXX_STANDARD_REQUIRED ${VELOX_CUDF_CXX_STANDARD_REQUIRED})
 
 unset(BUILD_SHARED_LIBS)
 endblock()
--- a/CMake/resolve_dependency_modules/cudf.cmake
+++ b/CMake/resolve_dependency_modules/cudf.cmake
@@ -19,25 +19,25 @@ cmake_minimum_required(VERSION 3.30.4)
 
 set(VELOX_rapids_cmake_VERSION 25.08)
 set(VELOX_rapids_cmake_BUILD_SHA256_CHECKSUM
-    f0e5484d49d9365b84eb41799db4fb1c5c8532f251ace734ead168bbbc1111c7)
+    38f9e374e726b9c30098c723ba97cde4574f411228453f0d247457406143ae20)
 set(VELOX_rapids_cmake_SOURCE_URL
-    "https://github.com/rapidsai/rapids-cmake/archive/492d474a62a91dd61bd6b91994eec2125e12fab8.tar.gz"
+    "https://github.com/rapidsai/rapids-cmake/archive/a0349d5b0eff1c68c399526b512aae754420a5b0.tar.gz"
 )
 velox_resolve_dependency_url(rapids_cmake)
 
 set(VELOX_rmm_VERSION 25.08)
 set(VELOX_rmm_BUILD_SHA256_CHECKSUM
-    9dc014c44191b0a23430f1bd7ebe38663643cdbbedd47a5a8e2079fdbbd4eb2e)
+  f2d7a64a3dcfe9b49231375b6df72b3be8a10e0828404463fd5176a9d75a1d4f)
 set(VELOX_rmm_SOURCE_URL
-    "https://github.com/rapidsai/rmm/archive/a68fdfe26e0d1e2e37feddf53eda1b7b5044f9f4.tar.gz"
+    "https://github.com/rapidsai/rmm/archive/29dd32302eb7c3e16fb837a1cfe4baac98071512.tar.gz"
 )
 velox_resolve_dependency_url(rmm)
 
 set(VELOX_kvikio_VERSION 25.08)
 set(VELOX_kvikio_BUILD_SHA256_CHECKSUM
-    8a5251ee1dff576578b0f96130ef87bfd8693de4769c08d33f203b6ea6750bab)
+    456126c106830666398b49fbeca572717c0b7f8f612ad7a6c96c7f63a6a1ad98)
 set(VELOX_kvikio_SOURCE_URL
-    "https://github.com/rapidsai/kvikio/archive/54c420a652bf29e3c7de9ac2e2e19af07de7c256.tar.gz"
+    "https://github.com/rapidsai/kvikio/archive/a2bbfeb0de49c29245da15d9df0ae7619c0a7531.tar.gz"
 )
 velox_resolve_dependency_url(kvikio)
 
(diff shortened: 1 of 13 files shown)


CMake/resolve_dependency_modules/cudf.cmake
```diff
@@ -17,35 +17,55 @@ include_guard(GLOBAL)
...
+set(VELOX_cudf_VERSION
+    25.08
+    CACHE STRING "cudf version")
+
+# cudf 25.08+ requires C++20
+if(VELOX_cudf_VERSION VERSION_GREATER_EQUAL "25.08")
+  set(VELOX_CUDF_CXX_STANDARD
+      20
+      CACHE STRING "C++ standard for cudf modules")
+  set(VELOX_CUDF_CXX_STANDARD_REQUIRED
```
bdice: It looks like this line is happening in both the if and else clauses. Should we pull it out of the conditional?

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -22,7 +22,6 @@
 #include "velox/exec/Task.h"
 
 #include <cudf/copying.hpp>
-#include <cudf/join.hpp>
```
bdice, line 25: This header was used for `cudf::hash_join`. Following “include what you use” we should keep the needed headers, even if they are transitively included by `CudfHashJoin.h`. The names may need to change like in the corresponding header.

velox/experimental/gpu/tests/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 list(APPEND CMAKE_PREFIX_PATH "${CUDAToolkit_LIBRARY_DIR}/cmake")
 find_package(CUB REQUIRED)
+find_package(Thrust REQUIRED)
```
bdice, line 16: I am wondering if we need find_package for CUB and Thrust in the cudf CMake code? The errors we were seeing were indeed related to a mismatched CUB/Thrust but I think this file isn’t the right place to fix it. I think this is only included if building tests, and this `gpu` section of code is independent of the cuDF CMake code (it is part of the Wave GPU backend).

CMake/resolve_dependency_modules/cudf.cmake
```diff
@@ -17,35 +17,55 @@ include_guard(GLOBAL)
...
 )
 velox_resolve_dependency_url(kvikio)
 
-set(VELOX_cudf_VERSION 25.04)
+set(VELOX_cudf_VERSION
+    25.08
+    CACHE STRING "cudf version")
+
+# cudf 25.08+ requires C++20
+if(VELOX_cudf_VERSION VERSION_GREATER_EQUAL "25.08")
```
bdice: I don’t expect us to support versions earlier than those pinned here. I think it might be safe to eliminate this conditional entirely and always use C++20. See also the comments below.

CMake/resolve_dependency_modules/cudf.cmake
```diff
@@ -17,35 +17,55 @@ include_guard(GLOBAL)
...
+  set(VELOX_CUDF_CXX_STANDARD
```
bdice: We could also set VELOX_CUDF_CUDA_STANDARD to 20. That way we ensure things propagate correctly if we used the CUDA compiler anywhere. What do you think?

https://cmake.org/cmake/help/latest/prop_tgt/CUDA_STANDARD.html

velox/experimental/gpu/tests/CMakeLists.txt
mattgara: You're right to question the location. My thinking is that this is the most direct place to fix the dependency resolution issue for this specific target.

When I look at the failing compilation command, I see an include chain that seems to be the core of the issue: `HashTableTest.cu` -> `cccl-src/.../cub` -> `cuda-12.8/.../thrust/`. Thrust appears to be being added as an (indirect) dependency through this chain. I've included a sample failure below for reference. 

The root of the problem is a bit complex:

1. The `CUB::CUB` target from the bundled cudf build is defined via a `FetchContent`, which makes it globally visible to all other sub-projects.

2. This global target supersedes any later attempts by this test's CMake to `find_package(CUB)` on its own (e.g., to find the system version). The CUB find package CMake module simply exits if the target `CUB::CUB` is already defined.

3. So, the test is forced to use the bundled `CUB`, but because we haven't specified a `Thrust` dependency, it falls back to the system's Thrust, causing the version mismatch.

I did consider another approach by isolating the cudf sub-project's dependencies. Based on looking into this, that would likely require changing cudf and other dependencies to use `ExternalProject*` instead of `FetchContent*`. That's a much more significant/complex change with more trade-offs to consider.

I also tried your recommendation of a find_package in the cudf CMake code, but it hit the same issues described above.

What are your thoughts?

Failing compilation sample:
```
[530/3095] Building CUDA object velox/experimental...s/velox_gpu_hash_table_test.dir/HashTableTest.cu.o                                       
FAILED: velox/experimental/gpu/tests/CMakeFiles/velox_gpu_hash_table_test.dir/HashTableTest.cu.o                                              
/usr/bin/ccache /usr/local/cuda-12.8/bin/nvcc -forward-unknown-to-host-compiler -DBOOST_ATOMIC_DYN_LINK -DBOOST_ATOMIC_NO_LIB -DBOOST_CONTEXT_
DYN_LINK -DBOOST_CONTEXT_NO_LIB -DBOOST_FILESYSTEM_DYN_LINK -DBOOST_FILESYSTEM_NO_LIB -DBOOST_PROGRAM_OPTIONS_DYN_LINK -DBOOST_PROGRAM_OPTIONS
_NO_LIB -DBOOST_REGEX_DYN_LINK -DBOOST_REGEX_NO_LIB -DBOOST_SYSTEM_DYN_LINK -DBOOST_SYSTEM_NO_LIB -DBOOST_THREAD_DYN_LINK -DBOOST_THREAD_NO_LI
B -DGFLAGS_IS_A_DLL=0 -DVELOX_ENABLE_ABFS -DVELOX_ENABLE_GCS -DVELOX_ENABLE_HDFS -DVELOX_ENABLE_PARQUET -DVELOX_ENABLE_S3 -I/velox/. -I/velox/
velox/external/xxhash -I/velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub -I/velox/_build/release/_deps/cccl-src/lib/cmake/libcu
dacxx/../../../libcudacxx/include -isystem /velox/velox -isystem /velox/velox/external -isystem /usr/include/libdwarf-0 -isystem /usr/local/cu
da-12.8/targets/x86_64-linux/include -ccbin /opt/rh/gcc-toolset-12/root/usr/bin -O3 -DNDEBUG -std=c++20 "--generate-code=arch=compute_70,code=
[compute_70,sm_70]" -Xcompiler=-fPIE -fdiagnostics-color=always -MD -MT velox/experimental/gpu/tests/CMakeFiles/velox_gpu_hash_table_test.dir/
HashTableTest.cu.o -MF velox/experimental/gpu/tests/CMakeFiles/velox_gpu_hash_table_test.dir/HashTableTest.cu.o.d -x cu -c /velox/velox/experi
mental/gpu/tests/HashTableTest.cu -o velox/experimental/gpu/tests/CMakeFiles/velox_gpu_hash_table_test.dir/HashTableTest.cu.o                 
nvcc warning : Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release (Use -Wno-
deprecated-gpu-targets to suppress warning).                                                                                                  
In file included from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/system/cuda/detail/execution_policy.h:40,                      
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/detail/device_system_tag.h:31,                        
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/detail/iterator_facade_category.h:30,                 
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/iterator_facade.h:45,                                 
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/detail/iterator_adaptor_base.h:31,                    
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/iterator_adaptor.h:44,                                
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/counting_iterator.h:42,                               
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/detail/discard_iterator_base.h:28,                    
                 from /usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/iterator/discard_iterator.h:32,                                
                 from /velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/util_type.cuh:49,                                   
                 from /velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/util_ptx.cuh:47,                                    
                 from /velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/block/block_adjacent_difference.cuh:45,             
                 from /velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/cub.cuh:48,                                         
                 from /velox/velox/experimental/gpu/tests/HashTableTest.cu:21:                                                                
/usr/local/cuda-12.8/targets/x86_64-linux/include/thrust/system/cuda/config.h:122:2: error: #error The version of CUB in your include path is 
not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. De
fine THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.                                                                                          
  122 | #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit,
 so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.                        
      |  ^~~~~                                                                                          
In file included from /velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/cub.cuh:64:   
/velox/_build/release/_deps/cccl-src/lib/cmake/cub/../../../cub/cub/device/device_for.cuh:48:10: fatal error: thrust/type_traits/unwrap_contig
uous_iterator.h: No such file or directory                                                               
   48 | #include <thrust/type_traits/unwrap_contiguous_iterator.h>                                      
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                      
compilation terminated.                                                         
```

bdice: That seems reasonable to me.

karthikeyann:
I removed VELOX_CUDF_CXX_STANDARD  related changes. Let's have this changes as another PR.
I need this PR merged for https://github.com/facebookincubator/velox/pull/14321 to be ready.

CMake/resolve_dependency_modules/cudf.cmake
mattgara: @karthikeyann removed this as it's no longer necessary after [10866](https://github.com/facebookincubator/velox/pull/10866) landed, it looks like all the warnings arising from that were fixed too, this applies to all the other comments related to C++20 changes.

CMake/resolve_dependency_modules/cudf.cmake
mattgara: This is a good idea, I think as @karthikeyann recommended we can do this in a separate PR.

mattgara: Update: I've pushed a branch with this here: https://github.com/bdice/velox/tree/cudf-cuda-std, we can choose to merge it later, likely after this PR lands.

velox/experimental/cudf/exec/CudfHashJoin.cpp
bdice: I’d like this fixed in a follow-up PR but don’t want to block this one since it is needed for other work to proceed.

mattgara: Sorry I completely misread the recent changeset, this should be fixed in this branch now :)

bdice:
RAPIDS 25.08 has been tagged and released now. Should we update again to use the final commits/hashes before merging this?

karthikeyann:
@bdice Yes. please update the hashes.

PR #14137, opened 2025-07-17
refactor(fuzzer): Unlink TableEvolutionFuzzer from GTest

majetideepak:


velox/exec/fuzzer/CMakeLists.txt
```diff
@@ -180,6 +180,18 @@ target_link_libraries(
   velox_vector_test_lib
   velox_vector_fuzzer)
 
+# Table Evolution Fuzzer
+add_library(velox_table_evolution_fuzzer_lib TableEvolutionFuzzer.cpp)
```
assignUser: Is this being used somewhere else? Otherwise there isn't really a need to split that off as a library, just add the source to the executable.

majetideepak: Fixed!

kgpai (approved):

pedroerp:
Cc: @Yuhta and @juwentus1234 who were recently work on it.

peterenescu:
Little late to the game, saw this was already approve. I think we shouldn't unlink TableEvolutionFuzzer right now from unit tests, we use it internally to verify Nimble support and we will lose test coverage. I wonder if we can extract it to another unit test suite if we are concerned about velox_exec_test or perhaps prioritizing longer-running tests.
cc: @kgpai @Yuhta @bikramSingh91

majetideepak:
@peterenescu Yes, we can add it to another unit test under `exec/fuzzers/tests`. Will that work?

peterenescu:
> @peterenescu Yes, we can add it to another unit test under `exec/fuzzers/tests`. Will that work?

I think that should work. Any concerns @bikramSingh91 @Yuhta ?

bikramSingh91:
> > @peterenescu Yes, we can add it to another unit test under `exec/fuzzers/tests`. Will that work?
> 
> I think that should work. Any concerns @bikramSingh91 @Yuhta ?

makes sense to me

bikramSingh91:
@majetideepak Just checking in. Please tag me once you have posted the changes and I'll move it through the merge queue quickly. Thanks!

PR #14139, opened 2025-07-17
Add now() and current_timestamp function in Velox

anandamideShakyan:
The now() scalar function is currently implemented only on the Presto Java side. However, this function is required on the native side to support queries originating from the BOLT pipeline. This function needs to be added on the Velox side.


aditi-pandit:
@amitkdutta

anandamideShakyan:
The fuzzer jobs fail because presto expects the current_timestamp function without (), but  the fuzzer is still generating current_timestamp() with empty parentheses, which Presto rejects as invalid SQL. 
Screenshot 2025-07-21 at 11 46 51 PM

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1580,6 +1580,39 @@ struct CurrentDateFunction {
...
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  TimeZoneKey timeZoneKey_ = 0;
+  Timestamp ts_;
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>&,
+      const core::QueryConfig& config) {
+    try {
+      timeZoneKey_ =
```
czentgr: getTimeZoneID can throw a Velox exception. Why are we catching this and then re-throw another one?

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1580,6 +1580,39 @@ struct CurrentDateFunction {
...
+      VELOX_USER_FAIL("Invalid session time zone: {}", e.what());
+    } catch (const std::exception& e) {
+      VELOX_FAIL(
+          "Unexpected error initializing current_timestamp: {}", e.what());
+    }
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<TimestampWithTimezone>& result) {
+    try {
+      result = pack(ts_, timeZoneKey_);
```
czentgr: Other functions using this also don't wrap this in try/catch. We can assume this already throws a proper Velox exception if it fails. So no need to recode it.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Removed the try-catch as they are not present at other places where it is used.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1580,6 +1580,26 @@ struct CurrentDateFunction {
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>&,
+      const core::QueryConfig& config) {
+    timeZoneKey_ =
+        tz::getTimeZoneID(config.sessionTimezone(), /*failOnError=*/true);
+    ts_ = Timestamp::now();
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<TimestampWithTimezone>& result) {
+    result = pack(ts_, timeZoneKey_);
```
aditi-pandit: You can do pack(ts_, timeZoneKey_) in initialize() as well and just return result each time.

As an aside, I wonder if we can optimize by implementing this as a Vector function that returns a ConstantVector instead of using the simple function API.

aditi-pandit: @anandamideShakyan : The pack(...) call can be moved to initialize() as well...
Maybe you haven't updated your code post the conversation.

aditi-pandit (commented):
Thanks @anandamideShakyan

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,28 @@ struct CurrentDateFunction {
...
+
+  TimeZoneKey timeZoneKey_{0};
+  Timestamp ts_;
+  int64_t packed_{0};
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>&,
+      const core::QueryConfig& config) {
+    timeZoneKey_ =
+        tz::getTimeZoneID(config.sessionTimezone(), /*failOnError=*/true);
```
czentgr: Why is fail on error set to true? The session timezone is checked in the QueryConfig already so we should never run into an error here.

czentgr (commented):
Also this function leads to fuzzer failures. Please check.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Ok. I will remove it.

anandamideShakyan: @czentgr The fuzzer tests are failing even after removing this. I think it is due to the fact that the function is non deterministic so there is a delay of nano seconds between expected and actual values. Do you have a suggestion on how to fix this? I think if fuzzer tests don't pick up now() and current_timestamp functions then also we are good, but I am not sure how to do that.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -126,6 +126,8 @@ std::unordered_map<std::string, std::shared_ptr<ArgValuesGenerator>>
 // testing. Use function signature to exclude only a specific signature.
 std::unordered_set<std::string> skipFunctions = {
     "noisy_empty_approx_set_sfm", // Non-deterministic because of privacy.
+    "current_timestamp",
```
czentgr: Please add a comment why they are excluded.

czentgr (commented):
How come there are GCS and S3 SEGvs? I don't see this in other PRs.
Maybe rebase?

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,27 @@ struct CurrentDateFunction {
   }
 };
 
+template <typename T>
+struct CurrentTimestampFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  TimeZoneKey timeZoneKey_{0};
```
czentgr: Why do we need `timeZoneKey_` and `ts_`?
Later we use the `packed_` value only. So do we need the others?

anandamideShakyan: You are right. Earlier I was calling the pack() function inside the call block. I shifted it to initialise block later. Totally forgot to shift the declaration for `timeZoneKey_ ` and `ts_` inside initialise block.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,25 @@ struct CurrentDateFunction {
...
 };
 
+template <typename T>
+struct CurrentTimestampFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  int64_t packed_{0};
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>&,
```
aditi-pandit: Typically we have a variable name which is commented for an unused parameter

`const std::vector<TypePtr>& /* type */,`

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,25 @@ struct CurrentDateFunction {
   }
 };
 
+template <typename T>
+struct CurrentTimestampFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  int64_t packed_{0};
```
aditi-pandit: Rename variable to result_

aditi-pandit (commented):
Thanks @anandamideShakyan. Have 2 minor comments else rest of the code is looking good.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,30 @@ struct CurrentDateFunction {
...
+struct CurrentTimestampFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  int64_t result_{0};
+  const tz::TimeZone* timeZone_ = nullptr;
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    Timestamp ts = Timestamp::now();
```
kgpai: Presto java gets current timestamp from session start time, just like it gets timezone from the config.

aditi-pandit: @kgpai : +1. Agree that would be better semantics, since then now() would give the same answer no matter the worker chosen.

Velox queryConfig has sessionStartTime https://github.com/facebookincubator/velox/blob/main/velox/core/QueryConfig.h#L926

Java decompiled object code shows current_timetamp takes sessionState.get().getQueryCurrentTimestamp

```
public ObjectInspector initialize(ObjectInspector[] arguments) throws UDFArgumentException {
        if (arguments.length != 0) {
            throw new UDFArgumentLengthException("The function CURRENT_TIMESTAMP does not take any arguments, but found " + arguments.length);
        } else {
            if (this.currentTimestamp == null) {
                this.currentTimestamp = new TimestampWritable(SessionState.get().getQueryCurrentTimestamp());
            }

            return PrimitiveObjectInspectorFactory.writableTimestampObjectInspector;
        }
    }
```
Are you suggesting we add wiring for that as well ?

kgpai: Yes, I believe we have to fix all of these to use session start time.

anandamideShakyan: I have added a [PR](https://github.com/prestodb/presto/pull/26173/files) in presto that passes sessionStartTime from presto to velox. Also updated the function code to use `kSessionStartTime`.

velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
```diff
@@ -89,6 +89,10 @@ void registerSimpleFunctions(const std::string& prefix) {
       {prefix + "to_unixtime"});
 
   registerFromUnixtime(prefix + "from_unixtime");
+  registerFunction<CurrentTimestampFunction, TimestampWithTimezone>(
```
pramodsatya, line 95: `{prefix + "now"}` can be added as an alias in the same call to `registerFunction`.

anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,30 @@ struct CurrentDateFunction {
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    Timestamp ts = Timestamp::fromMillis(config.sessionStartTimeMs());
+    timeZone_ = getTimeZoneFromConfig(config);
+    if (timeZone_ != nullptr) {
+      result_ = pack(ts, timeZone_->id());
+      return;
+    }
+    result_ = pack(ts, Timestamp::defaultTimezone().id());
```
aditi-pandit: @kgpai : I feel we should error here instead of pick up the default timezone. We always want the value on the co-ordinator and not the local timezone.
wdyt ?

anandamideShakyan: Done. @aditi-pandit Please take another look. Thanks.

velox/docs/functions/presto/datetime.rst
```diff
@@ -188,6 +188,19 @@ Date and Time Functions
 
         SELECT current_timezone;   -- Asia/Kolkata
 
+.. function:: current_timestamp() -> timestamp with time zone
```
aditi-pandit, line 191: @anandamideShakyan : We generally order by function names in rst file but I don't see this convention being followed in this file. 

Maybe do a follow up to reorganize these documentation in sorted order of function names.

aditi-pandit:
@anandamideShakyan : Please can you fix the pre-commit errors.

@kgpai, @kagamiori : Hi Krishna, Wei : Please can you help with the review and merge for these functions.

velox/docs/functions/presto/datetime.rst
anandamideShakyan: Sure.

czentgr:
I removed the `ready-to-merge` There were pushes and the CI is failing.
@anandamideShakyan Did you rebase and now we have a broken Velox build?

czentgr:
@anandamideShakyan Please rebase again. There was a commit that temporarily added duplicate code form the looks of it and has been reverted.

anandamideShakyan:
I have rebased, all ci checks are passing except for one which seems to be unrelated to my code changes. Please take a look @czentgr.

czentgr:
@bikramSingh91 I see you imported the PR but there are Meta internal failures. Can we help with this?

kKPulla:
@czentgr Bikram is away right now, I can take a stab at this and try to merge this today or tomorrow. I will keep you posted.

PR #14141, opened 2025-07-17
Add TopNRank optimization

aditi-pandit:


Enochack:
We really need this PR. When is it expected to be done?

aditi-pandit:
@Enochack : We have now merged all the underlying refactoring PRs. So I will rebase this one for review now.

velox/exec/TopNRowNumber.cpp
```diff
@@ -260,7 +325,62 @@ char* TopNRowNumber::processRowWithinLimit(
...
+    core::TopNRowNumberNode::RankFunction::kRank>(
+    vector_size_t index,
+    TopRows& partition) {
+  auto& topRows = partition.rows;
+  // The new row < top rank
+  // For rank, the new row gets assigned its rank as per its position in the
+  // queue. But the ranks of all subsequent rows increment by 1.
+  // So we can remove the rows at the top rank as its rank > limit now.
+  char* topRow = partition.removeTopRankRows();
+  char* newRow = data_->initializeRow(topRow, true /* reuse */);
```
jinchengchenghh: `/*reuse=*/true`

velox/exec/TopNRowNumber.cpp
```diff
@@ -251,7 +272,51 @@ void TopNRowNumber::initializeNewPartitions() {
   }
 }
 
-char* TopNRowNumber::processRowWithinLimit(
+template <>
+char* TopNRowNumber::processRowWithinLimit<
+    core::TopNRowNumberNode::RankFunction::kRank>(
+    vector_size_t index,
+    TopRows& partition) {
+  // The topRanks queue is not filled yet.
```
jinchengchenghh, line 281: Move the comment to L281

velox/exec/TopNRowNumber.cpp
```diff
@@ -251,7 +272,51 @@ void TopNRowNumber::initializeNewPartitions() {
...
+  auto& topRows = partition.rows;
+  if (topRows.empty()) {
+    partition.topRank = 1;
+  } else {
+    // rank assigns all peer rows the same rank, but the rank increments by
```
jinchengchenghh: rank -> Rank

velox/exec/TopNRowNumber.cpp
```diff
@@ -251,7 +272,51 @@ void TopNRowNumber::initializeNewPartitions() {
...
+    // the number of peers when moving between peers. So when adding a new
+    // row:
+    // If row == top rank then top rank is unchanged.
+    // If row < top rank then top rank += 1.
+    // If row > top, then rank += number of peers of top rank.
+    auto topRow = topRows.top();
```
jinchengchenghh: auto*

velox/exec/TopNRowNumber.cpp
```diff
@@ -251,7 +272,51 @@ void TopNRowNumber::initializeNewPartitions() {
...
+    auto result = comparator_.compare(decodedVectors_, index, topRow);
```
jinchengchenghh, line 293: What if result == 0?

velox/exec/TopNRowNumber.cpp
```diff
@@ -260,7 +325,62 @@ char* TopNRowNumber::processRowWithinLimit(
...
+    vector_size_t index,
+    TopRows& partition) {
+  char* newRow = nullptr;
+  // The new row < top rank
+  // For dense_rank:
+  // i) If the row is a peer of an existing row in the queue, then it
+  // has the same rank as it. The ranks of other rows are unchanged. So its
+  // only added to the queue.
+  // ii) If the row is a distinct new value in the queue, then it is assigned
+  // a rank as per its position, and the ranks of all subsequent rows +=1.
```
jinchengchenghh: rows +=1 -> rows += 1.

velox/exec/TopNRowNumber.cpp
```diff
@@ -260,7 +325,62 @@ char* TopNRowNumber::processRowWithinLimit(
...
+  // So the current top rank rows can be removed from the queue as their new
+  // rank > limit.
+  if (partition.isDuplicate(decodedVectors_, index)) {
+    newRow = data_->newRow();
+  } else {
+    char* topRow = partition.removeTopRankRows();
+    newRow = data_->initializeRow(topRow, true /* reuse */);
```
jinchengchenghh: ditto

velox/exec/TopNRowNumber.cpp
```diff
@@ -260,7 +325,62 @@ char* TopNRowNumber::processRowWithinLimit(
...
+  // has the same rank as it. The ranks of other rows are unchanged. So its
+  // only added to the queue.
+  // ii) If the row is a distinct new value in the queue, then it is assigned
+  // a rank as per its position, and the ranks of all subsequent rows +=1.
+  // So the current top rank rows can be removed from the queue as their new
+  // rank > limit.
+  if (partition.isDuplicate(decodedVectors_, index)) {
+    newRow = data_->newRow();
+  } else {
+    char* topRow = partition.removeTopRankRows();
```
jinchengchenghh, line 376: wrap the topRow since it is only used only once

velox/exec/TopNRowNumber.cpp
```diff
@@ -562,29 +734,71 @@ bool TopNRowNumber::isNewPartition(
...
     const RowVectorPtr& output,
     vector_size_t index,
     const SpillMergeStream* next) {
   if (isNewPartition(output, index, next)) {
     nextRank_ = 1;
+    numPeers_ = 1;
+    return;
+  }
+
+  // The rank always += 1 for row_number.
```
jinchengchenghh: Remove the comment

velox/exec/TopNRowNumber.cpp
```diff
@@ -824,4 +1039,66 @@ void TopNRowNumber::setupSpiller() {
       &spillConfig_.value(),
       spillStats_.get());
 }
+
+// Using the underlying vector of the priority queue for the algorithms to
+// check duplicates and count the number of top rank rows. This makes the
+// algorithms O(n). There could be other approaches to make the
+// algorithms O(1), but would trade memory efficiency.
+namespace {
+template <class T, class S, class C>
```
jinchengchenghh, line 1065: Move to the start of the file with other functions in anonymous space

velox/exec/TopNRowNumber.cpp
```diff
@@ -824,4 +1039,66 @@ void TopNRowNumber::setupSpiller() {
...
+    rows.pop();
+  }
+  return topRow;
+}
+
+vector_size_t TopNRowNumber::TopRows::numTopRankRows() {
+  VELOX_CHECK(!rows.empty());
+  char* topRow = rows.top();
+  vector_size_t numRows = 0;
+  const std::vector<char*, StlAllocator<char*>> partitionRowsVector =
```
jinchengchenghh, line 1096: Why use StlAllocator<char*>?

velox/exec/tests/TopNRowNumberTest.cpp
```diff
@@ -29,12 +29,28 @@ namespace {
...
+      : function_(function),
+        functionName_(core::TopNRowNumberNode::rankFunctionName(function_)) {}
+
+  void SetUp() override {
+    exec::test::OperatorTestBase::SetUp();
     filesystems::registerLocalFileSystem();
   }
+
+ protected:
+  const core::TopNRowNumberNode::RankFunction function_;
```
jinchengchenghh: This is not used in the test, remove it.

velox/exec/tests/TopNRowNumberTest.cpp
```diff
@@ -29,12 +29,28 @@ namespace {
...
+  explicit TopNRowNumberTest(core::TopNRowNumberNode::RankFunction function)
+      : function_(function),
+        functionName_(core::TopNRowNumberNode::rankFunctionName(function_)) {}
+
+  void SetUp() override {
+    exec::test::OperatorTestBase::SetUp();
     filesystems::registerLocalFileSystem();
   }
+
+ protected:
```
jinchengchenghh: remove, it has been in the protected scope

velox/exec/TopNRowNumber.cpp
aditi-pandit: partition.topRank is unchanged. Comment on line 289.

velox/exec/TopNRowNumber.cpp
aditi-pandit: I'm bit divided on this review comment. I think the current code is cleaner to understand.

velox/exec/TopNRowNumber.cpp
aditi-pandit: This is closer to its usage in the functions below. We do add functions in anonymous namespace in the middle of files as well, especially if the anonymous function is just used in the next few functions.

velox/exec/TopNRowNumber.cpp
aditi-pandit: StlAllocator is a Velox class for a STL like interface over HashStringAllocator. This is used widely in Velox code.

Is there any particular concern you have with its use ?

aditi-pandit:
Thanks @jinchengchenghh for the review. Have addressed the comments. PTAL.

pedroerp:
@aditi-pandit and I chatted offline. Let's get the fuzzer changes in first and ensure we have better test coverage before we proceed with this one. 

I can help review/merge the fuzzer changes.

Enochack:
We tested a query which has a `TopNRowNumber` operator before a `Window` operator in our in-house query engine. And the `Window` doesn't work with streaming mode ( `inputsSorted = true`). So shall we sort the paritioning keys and sorting keys together to avoid sorting again in the following `Window`? NOTE that the following `Window` has more window expressions so the `Window` can't be eliminated.

Thanks!

aditi-pandit:
> We tested a query which has a `TopNRowNumber` operator before a `Window` operator in our in-house query engine. And the `Window` doesn't work with streaming mode ( `inputsSorted = true`). So shall we sort the paritioning keys and sorting keys together to avoid sorting again in the following `Window`? NOTE that the following `Window` has more window expressions so the `Window` can't be eliminated.
> 
> Thanks!

@Enochack : It would be better to not set inputsSorted for the Window operator. Without inputsSorted, WindowOperator will do its own sorting. Did you try that ?

Enochack:
> > We tested a query which has a `TopNRowNumber` operator before a `Window` operator in our in-house query engine. And the `Window` doesn't work with streaming mode ( `inputsSorted = true`). So shall we sort the paritioning keys and sorting keys together to avoid sorting again in the following `Window`? NOTE that the following `Window` has more window expressions so the `Window` can't be eliminated.
> > Thanks!
> 
> @Enochack : It would be better to not set inputsSorted for the Window operator. Without inputsSorted, WindowOperator will do its own sorting. Did you try that ?

Yes, that works. But is it better to do sorting only once in `TopNRowNumber`?

aditi-pandit:
@Enochack Well, TopNRowNumber outputs all the partition rows in rank order. So there is a partial order, the output maintains the intra-partition ordering of rows. If you want to use PartitionStreamingWindowBuild for further processing and avoid sorting, it might need some changes.Mostly in the logic for demarcating WindowPartition. What are the issues you saw when you run your test ? Can you share the error message ?

Enochack:
> @Enochack Well, TopNRowNumber outputs all the partition rows in rank order. So there is a partial order, the output maintains the intra-partition ordering of rows. If you want to use PartitionStreamingWindowBuild for further processing and avoid sorting, it might need some changes.Mostly in the logic for demarcating WindowPartition. What are the issues you saw when you run your test ? Can you share the error message ?

@aditi-pandit
There's no error message.

The issue is the output of `TopNRank -> Window(inputsSorted = true)` is incorrect. But the output of `TopNRank -> Window(inputsSorted = false)` is correct. However, it needs to do sorting in both `TopNRank` and `Window`.

So what I expect is `TopNRank -> Window(inputsSorted = true)` produces a correct output like `TopNRank -> Window(inputsSorted = false)`, which means `TopNRank` needs to produce a output which is sorted first by partition value and then by sorting key so that `Window` doesn't need to sort.

Is this feasible?

aditi-pandit:
@Enochack : So TopN uses a HashTable to separate partitions. Its not really possible to sort the entire output across partitions for that case. We could add more data structures to guarantee an ordering of partition values on output. But that kind of data structure could break if we decided to spill to disk.

There could be 2 options:
i) Use Window inputSorted=false. It would almost as expensive as adding a new sort in TopNRowNumber.
ii) Figure any assumption in PartitionStreamingWindowBuild (either related how to disambiguate partitions or output ordering) that is causing wrong results. And then remove those restrictions,

aditi-pandit:
> @aditi-pandit and I chatted offline. Let's get the fuzzer changes in first and ensure we have better test coverage before we proceed with this one.
> 
> I can help review/merge the fuzzer changes.

@pedroerp : I added the fuzzer changes to this PR now. 
Though it would be good to get the changes in https://github.com/facebookincubator/velox/pull/15249 and https://github.com/facebookincubator/velox/pull/15255 merged so that there are only minimal changes in this PR (TopNRowNumberFuzzer).

aditi-pandit:
@pedroerp @xiaoxmeng : This code is updated with the fuzzer changes. Would be great to get some review comments.

velox/exec/TopNRowNumber.h
```diff
@@ -84,13 +84,28 @@ class TopNRowNumber : public Operator {
...
     // This is the highest rank (this code will be enhanced for rank, dense_rank
     // soon) seen so far in the input rows. It is compared
     // with the limit for the operator.
     int64_t topRank = 0;
 
+    // Number of rows with the highest rank in the partition.
+    vector_size_t numTopRankRows();
+
+    // Remove all rows with the highest rank in the partition.
+    char* removeTopRankRows();
```
pedroerp, line 153: can you comment on what this returns?

velox/exec/TopNRowNumber.h
```diff
@@ -84,13 +84,28 @@ class TopNRowNumber : public Operator {
...
+
+    // Returns true if the row at decodedVectors[index] has the same order by
+    // keys as another row in the partition's top rows.
+    bool isDuplicate(
+        const std::vector<DecodedVector>& decodedVectors,
+        vector_size_t index);
```
pedroerp, line 160: is index the index on the vector, or on the DecodedVector?

velox/exec/TopNRowNumber.h
```diff
@@ -158,6 +186,14 @@ class TopNRowNumber : public Operator {
       vector_size_t index,
       const SpillMergeStream* next);
 
+  // Returns true if 'next' row is a new peer (rows differ on order by keys)
```
pedroerp: maybe I misunderstood things, but isn't a `peer` a row with the same order by key values?

velox/exec/fuzzer/TopNRowNumberFuzzer.cpp
```diff
@@ -184,11 +199,14 @@ TopNRowNumberFuzzer::makeDefaultPlan(
   projectFields.emplace_back("row_number");
 
   int32_t limit = randInt(1, FLAGS_batch_size);
-  auto plan = test::PlanBuilder()
-                  .values(input)
-                  .topNRowNumber(partitionKeys, sortKeys, limit, true)
-                  .project(projectFields)
-                  .planNode();
+  int32_t rankFunction = randInt(0, 2); // 0: ROW_NUMBER, 1: RANK, 2: DENSE_RANK
```
pedroerp: probably forgot to remove this?

velox/exec/TopNRowNumber.h
```diff
@@ -84,13 +84,28 @@ class TopNRowNumber : public Operator {
     std::priority_queue<char*, std::vector<char*, StlAllocator<char*>>, Compare>
         rows;
 
+    RowComparator& rowComparator;
```
pedroerp, line 141: Can you update the comment on the beginning to the class to add more context on how this works now?

velox/exec/TopNRowNumber.h
aditi-pandit: Yeah... I see how this is confusing. I changed this to isNewRank to indicate that it will have a different rank.

velox/exec/TopNRowNumber.h
aditi-pandit: Its index in DecodedVector. Made the function comment more clear on this.

velox/exec/TopNRowNumber.h
aditi-pandit: Added a comment about the TopRows structure.

aditi-pandit:
> Thanks @aditi-pandit, went over about half of it and made some initial comments.
> 
> Could you add a header comment to the TopNRowNumber class to explains how it works internally, and the different implementation required to implemented the three functions? I could kind of infer from reading the code, but this should be documented.

Done.

@pedroerp : Have updated the code as per your comments. PTAL.

velox/exec/TopNRowNumber.h
```diff
@@ -22,20 +22,58 @@
...
+/// and then computes the ranks of each row using computeNextRankInMemory().
+/// The logic of the next rank differs based on the ranking function.
+///
+/// If the rows are in the spill, then the spiller iterates over each spilled
+/// partition in order of the ranks. For each row from the spill, the next
+/// rank is computed using computeNextRankInSpill() function. The logic of
+/// the next rank differs based on the ranking function.
+/// Note : The spill could have > limit rows for a partition as each spill
+/// resets the TopRows for the partition. So stop outputting rows after
+/// reaching the limit for each partition.
```
pedroerp, line 75: this is extremely helpful. Thanks for adding the detailed documentation!

velox/exec/fuzzer/TopNRowNumberFuzzer.cpp
pedroerp: shouldn't you remove this line since this is generated inside generateRankFunction?

velox/exec/TopNRowNumber.cpp
```diff
@@ -289,14 +412,18 @@ void TopNRowNumber::processInputRow(vector_size_t index, TopRows& partition) {
       return;
     }
 
+    // This row has the same value as the top rank row. row_number rejects
+    // such rows, but are added to the queue for rank and dense_rank. The top
+    // rank remains unchanged.
```
pedroerp, line 417: nit: maybe an `else if ()` would be clearer

velox/exec/TopNRowNumber.cpp
```diff
@@ -360,15 +488,46 @@ void TopNRowNumber::updateEstimatedOutputRowSize() {
...
+// If there are many peer rows for the highest rank, then topRank could
+// oscillate between the two cases of topRank < limit and topRank > limit
+// as rows are added
+// E.g. If the input rows are 0, 0, 0, 5, 0, 0, 6 and we want rank < = 5, then
+// at 0, 0, 0, 5 :
+// topRows.pq - 0, 0, 0, 5 topRank -> 4
+// 0 is added.
+// topRows.pq - 0, 0, 0, 0, 5 topRank -> 5
+// topRank = limit now.
+// So when the next 0 is added, the last 5 is popped from TopRowsand 0 is added
```
pedroerp: nit: missing space

velox/exec/TopNRowNumber.cpp
```diff
@@ -360,15 +488,46 @@ void TopNRowNumber::updateEstimatedOutputRowSize() {
   }
 }
 
+// This function handles a special case when determining the starting
+// rank value for the 'rank' function.
+// If there are many peer rows for the highest rank, then topRank could
+// oscillate between the two cases of topRank < limit and topRank > limit
+// as rows are added
+// E.g. If the input rows are 0, 0, 0, 5, 0, 0, 6 and we want rank < = 5, then
```
pedroerp: no space between < and =

velox/exec/TopNRowNumber.cpp
```diff
@@ -493,22 +675,30 @@ RowVectorPtr TopNRowNumber::getOutputFromMemory() {
       }
     }
 
-    const auto numOutputRowsLeft = outputBatchSize_ - offset;
+    auto numOutputRowsLeft = outputBatchSize_ - offset;
```
pedroerp: why removing the const?

pedroerp (approved):
Thanks @aditi-pandit, the documentation really helped. I went over the entire thing, and other than a few cosmetic things, it looks good to me. 

Like we discussed offline, let's make sure we enhance the current fuzzers to also compare the results of this operator with a Window + limit to give us even more coverage before we enable this in Presto.

pedroerp:
Please add the "ready-to-merge" tag once the cosmetic items are fixed.

velox/exec/fuzzer/TopNRowNumberFuzzer.cpp
aditi-pandit: That was a good catch. Thanks.

velox/exec/TopNRowNumber.cpp
aditi-pandit: My bad. Have added it back.

aditi-pandit:
Fixed the review comments from Pedro. Please see comment above.

aditi-pandit:
Note: This code isn't used by Presto. It needs an optimizer rule to convert Window plan matching this case to TopNRank plan. We will use a config flag to turn on that Presto optimizer rule. By default it will be turned off so that Meta production isn't affected.

Issue #14080, opened 2025-07-11
Support merging spilled sorted files in multi rounds to limit the opened file num and memory usage


HolyLow:
### Description

Currently, when operators like HashAgg, Sort, Window trigger spilling, they would sort and spill the data into a bunch of files, and then merge the sorted files with a single `TreeOfLosers` in a single merging round. 
Though this implementation reduces the file IO (each file would be read/write only once), it might cause stability problem when the file set is large.

For example, if the agg dataset is very large, or the data is skewed, a single driver would spill a large mount of sorted files, which will be opened and read simultaneously. As each file would has a read buffer and each deserd stream has to deserd at least a RowVector in memory, the memory consumption will be proportional to num of spilled files. When the spilled file num is large,  the memory consumption would be large and causes OOM.

In our usecase, we encountered this error when agg reads spilled files:
```
"message": " allocate failed with 1.50MB from Memory Pool[op.4.0.6.Aggregation LEAF root[20250710_133605_9377_d96u6_6] parent[node.4] MMAP track-usage thread-safe]<max capacity 11.17GB unlimited capacity used 1.16GB available 4.61MB reservation [used 1.16GB, reserved 1.16GB, min 0B] counters [allocs 3808, frees 363, reserves 601, releases 21, collisions 2])> Failed to evict from cache state: AsyncDataCache:",
    "cause": {
      "type": "VeloxRuntimeError",
      "message": " allocate failed with 1.50MB from Memory Pool[op.4.0.6.Aggregation LEAF root[20250710_133605_9377_d96u6_6] parent[node.4] MMAP track-usage thread-safe]<max capacity 11.17GB unlimited capacity used 1.16GB available 4.61MB reservation [used 1.16GB, reserved 1.16GB, min 0B] counters [allocs 3808, frees 363, reserves 601, releases 21, collisions 2])> Failed to evict from cache state: AsyncDataCache:\nCache size: 0B tinySize: 0B large size: 0B\nCache entries: 0 read pins: 0 write pins: 0 pinned shared: 0B pinned exclusive: 0B\n num write wait: 0 empty entries: 0\nCache access miss: 0 hit: 0 hit bytes: 0B eviction: 0 savable eviction: 0 eviction checks: 0 aged out: 0 stales: 0\nPrefetch entries: 0 bytes: 0B\nAlloc Megaclocks 0\nAllocated pages: 2883583 cached pages: 0\n Operator: Aggregation[4] 2",
      "suppressed": [],
[...]

PR #14143, opened 2025-07-18
feat: support merging sorted files in multiple rounds

HolyLow:
Summary:
Support merging sorted files in multiple rounds, in order to cap the max opened files and corresponding memory consumption.

This MR solves https://github.com/facebookincubator/velox/issues/14080.

cc @xiaoxmeng @mbasmanova @duanmeng 

Any suggestion is welcome, thanks a lot.


velox/common/base/SpillConfig.cpp
```diff
@@ -34,6 +34,7 @@ SpillConfig::SpillConfig(
     uint64_t _maxSpillRunRows,
     uint64_t _writerFlushThresholdSize,
     const std::string& _compressionKind,
+    int32_t _numMaxMergeWays,
```
duanmeng: uint64_t?

velox/common/base/SpillConfig.h
```diff
@@ -71,6 +71,7 @@ struct SpillConfig {
       uint64_t _maxSpillRunRows,
       uint64_t _writerFlushThresholdSize,
       const std::string& _compressionKind,
+      int32_t numMaxMergeWays,
```
duanmeng: ditto

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+    RowVectorPtr bufferVector,
+    std::vector<const RowVector*>& bufferSpillSources,
+    std::vector<vector_size_t>& bufferSpillSourceRows,
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
+  std::vector<std::unique_ptr<SpillMergeStream>> streams;
+  streams.reserve(files.size());
+  for (auto& fileInfo : files) {
```
duanmeng: const auto&?

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
   return std::make_unique<TreeOfLosers<SpillMergeStream>>(std::move(streams));
 }
 
+namespace {
+SpillFileInfo mergeFiles(
+    std::vector<SpillFileInfo>& files,
```
duanmeng: const std::vector<SpillFileInfo>& files

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+    uint64_t readBufferSize,
+    uint64_t writeBufferSize,
+    vector_size_t batchSize,
+    RowVectorPtr bufferVector,
+    std::vector<const RowVector*>& bufferSpillSources,
+    std::vector<vector_size_t>& bufferSpillSourceRows,
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
```
duanmeng: VELOX_CHECK_GT(files.size(), 0)?

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+  std::vector<std::unique_ptr<SpillMergeStream>> streams;
+  streams.reserve(files.size());
+  for (auto& fileInfo : files) {
+    streams.push_back(FileSpillMergeStream::create(
+        SpillReadFile::create(fileInfo, readBufferSize, pool, spillStats)));
+  }
+  auto mergeStream =
```
duanmeng: const auto mergeStream

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+      std::make_unique<TreeOfLosers<SpillMergeStream>>(std::move(streams));
+  VELOX_CHECK(!files.empty());
+  auto type = files[0].type;
```
duanmeng: ditto

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+  auto startFileNum = files_.size();
+  std::priority_queue<
+      SpillFileInfo,
+      std::vector<SpillFileInfo>,
+      CompareFileBySize>
+      orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(mergeWayThreshold);
+  auto filePathPrefix = files_[0].path;
+  const size_t batchSize = 1'000;
```
duanmeng: Should it be outputSize_?

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+  if (mergeWayThreshold <= 1 || files_.size() <= mergeWayThreshold) {
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  auto startFileNum = files_.size();
+  std::priority_queue<
+      SpillFileInfo,
+      std::vector<SpillFileInfo>,
+      CompareFileBySize>
+      orderedFiles(files_.begin(), files_.end());
```
duanmeng: Do we need to prioritize the merge files by their size?

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+    VectorPtr tmp = std::move(bufferVector);
+    BaseVector::prepareForReuse(tmp, batchSize);
+    bufferVector = std::static_pointer_cast<RowVector>(tmp);
+    for (auto& child : bufferVector->children()) {
+      child->resize(batchSize);
+    }
+    int32_t outputRow = 0;
+    int32_t outputSize = 0;
+    bool isEndOfBatch = false;
+    while (stream != nullptr && outputRow + outputSize < bufferVector->size()) {
```
duanmeng: Should we extract this part as a separate method, and cover it by UT?

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,172 @@ SpillPartition::createOrderedReader(
...
+          bufferSpillSourceRows);
+      outputRow += outputSize;
+      outputSize = 0;
+    }
+
+    IndexRange range{0, outputRow};
+    writer->write(bufferVector, folly::Range<IndexRange*>(&range, 1));
+  }
+  auto resultFiles = writer->finish();
+  VELOX_CHECK_EQ(resultFiles.size(), 1);
```
duanmeng, line 488: Perhaps we should limit the size of the spill file and produce a sequence of files. cc @xiaoxmeng

duanmeng (commented):
@HolyLow Thanks for you contribution % some comments.

velox/common/base/SpillConfig.cpp
HolyLow: done.

velox/common/base/SpillConfig.h
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: Yes, because we want to minimize the sizes of the merged files. For example, we have 3 files with sizes 1MB, 2MB, 10MB, and we should choose to merge the 1MB file and 2MB file to minimize the file IO.

velox/exec/Spill.cpp
HolyLow: Refactored as requested, and UTs are added.

velox/exec/Spill.cpp
HolyLow: It would be better if we are able to split a sorted file into a group of sorted files, but under current abstraction, the file management would be very ugly. I think to achieve this, the file layer abstraction should be enhanced. A logical file layer should be added, which could be backed by a vector of physical files. But implementing this feature in this PR would be somehow out of scope, what do you think?

velox/exec/Spill.cpp
HolyLow: This is a batchSize for each gathering and writing, and no outputSize_ could be used in this method.

velox/exec/Spill.cpp
HolyLow: done.

HolyLow:
@duanmeng @xiaoxmeng Refactored as requested. Any suggestion is welcome.

velox/exec/Spill.cpp
HolyLow: This should not be const as it's non-const method would be used.

HolyLow:
@duanmeng @xiaoxmeng Any further comment?

HolyLow:
@mbasmanova @majetideepak @tanjialiang Could you kindly help review this PR, or find someone proper to review it? Thanks a lot.

mbasmanova:
@xiaoxmeng Meng, would you help move this PR forward?

HolyLow:
@xiaoxmeng Could you help review this PR? Thanks a lot!

HolyLow:
@duanmeng Do you have any further suggestions? @mbasmanova @xiaoxmeng It would be great if you could share your (whatever) thoughts on this patch. Thanks again.

duanmeng:
@HolyLow Sorry for the late reply—I just got back from a long vacation. I’ll take another look this week.

velox/exec/OperatorUtils.cpp
```diff
@@ -446,6 +446,43 @@ void gatherCopy(
...
+    TreeOfLosers<SpillMergeStream>* mergeTree,
+    int32_t& count,
+    std::vector<const RowVector*>& bufferSources,
+    std::vector<vector_size_t>& bufferSourceIndices) {
+  VELOX_CHECK_GE(bufferSources.size(), target->size());
+  VELOX_CHECK_GE(bufferSourceIndices.size(), target->size());
+  count = 0;
+  int32_t outputSize = 0;
+  bool isEndOfBatch = false;
+  SpillMergeStream* currentStream = mergeTree->next();
```
duanmeng: How about
```C++
while (count + outputSize < target->size()) {
    SpillMergeStream* currentStream = mergeTree->next();
    VELOX_CHECK_NOT_NULL(currentStream);
    ....
    currentStream->pop();
```

velox/exec/OperatorUtils.cpp
```diff
@@ -446,6 +446,43 @@ void gatherCopy(
...
+    currentStream->pop();
+    currentStream = mergeTree->next();
+  }
+  VELOX_CHECK_LE(count + outputSize, target->size());
+
+  if (FOLLY_LIKELY(outputSize != 0)) {
+    gatherCopy(target, count, outputSize, bufferSources, bufferSourceIndices);
+    count += outputSize;
+    outputSize = 0;
+  }
```
duanmeng: Perhaps you could make the `count` as a local variable and resize the target to `count`, and then you do not need the `int& count` param and `outputRow` outside this method.

velox/exec/Spill.cpp
duanmeng: Could you use constexpr see https://github.com/facebookincubator/velox/blob/main/velox/exec/Spiller.cpp#L202.

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,143 @@ SpillPartition::createOrderedReader(
...
+  return std::move(resultFiles[0]);
+}
+
+class CompareFileBySize {
+ public:
+  bool operator()(const SpillFileInfo& lhs, const SpillFileInfo& rhs) const {
+    return lhs.size > rhs.size;
+  }
```
duanmeng, line 495:
```C++
struct SpillFileInfoComp {
 public:
  bool operator()(const SpillFileInfo& lhs, const SpillFileInfo& rhs) const {
    return lhs.size > rhs.size;
  }
};
using SpillFileInfoHeap =
    std::priority_queue< SpillFileInfo, std::vector< SpillFileInfo >, SpillFileInfoComp>;
```

velox/exec/Spill.cpp
```diff
@@ -310,6 +311,143 @@ SpillPartition::createOrderedReader(
...
+  if (mergeWayThreshold <= 1 || files_.size() <= mergeWayThreshold) {
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  auto startFileNum = files_.size();
+  std::priority_queue<
+      SpillFileInfo,
+      std::vector<SpillFileInfo>,
+      CompareFileBySize>
+      orderedFiles(files_.begin(), files_.end());
```
duanmeng: `SpillFileInfoHeap orderedFiles`

velox/exec/OperatorUtils.h
```diff
@@ -135,6 +135,19 @@ void gatherCopy(
     const std::vector<vector_size_t>& sourceIndices,
     const std::vector<IdentityProjection>& columnMap = {});
 
+/// Merge sort with the mergeTree and gatherCopy the results into target.
+/// 'target' is the result RowVector, and the copying starts from row #0 up to
+/// row #target.size(). 'mergeTree' is the data source. 'count' is the actual
+/// row count that is copied to target. 'bufferSources' and
+/// 'bufferSourceIndices' are buffering vectors that could be reused across
+/// callings.
+void gatherMerge(
```
duanmeng: Perhaps, this method should be placed in `Spill.cpp`.

velox/exec/OperatorUtils.cpp
HolyLow: The `currentStream != nullptr` must remain in the loop condition to jump out of loop when the mergeTree is finished. But I got what you mean, and I refactored the code with for loop to avoid code duplication.

velox/exec/OperatorUtils.cpp
HolyLow: This is not proper. The `count` is not computed, but accumulated during the writing process. Besides, the `target->size()` is the upper bound limit on `count` as `target` serves as the output buffer. The dependency cannot be reversed.

velox/exec/Spill.cpp
HolyLow: Refactored with constexpr as suggested.

velox/exec/Spill.cpp
HolyLow: Refactored as suggested.

velox/exec/Spill.cpp
HolyLow: Refactored as suggested.

velox/exec/OperatorUtils.h
HolyLow: Moved the function and ut to `Spill.cpp` and `SpillTest.cpp`.

HolyLow:
@duanmeng Thanks for your comments, refactored as suggested. Any further comment is welcome.

duanmeng:
> @duanmeng Thanks for your comments, refactored as suggested. Any further comment is welcome.

@HolyLow LGTM

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,142 @@ SpillPartition::createOrderedReader(
...
+  if (mergeWayThreshold <= 1 || files_.size() <= mergeWayThreshold) {
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  auto startFileNum = files_.size();
+  SpillFileInfoHeap orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(mergeWayThreshold);
+  auto filePathPrefix = files_[0].path;
+  constexpr size_t batchSize = 1'000;
```
duanmeng: Perhaps setting the batch size to `1000` might be too large and could increase memory pressure during the final merge stage, since each stream holds a batch of that size. see https://github.com/facebookincubator/velox/pull/14445

HolyLow: Added new logic to limit the batch's size to under 64KB, which followed what https://github.com/facebookincubator/velox/pull/14445 did. 64KB should be small enough as 1k merge ways would consume up to 64MB memory, which is affordable in reading phase.

HolyLow:
@duanmeng Refactored as suggested. Any further comment is welcome, thanks a lot.

velox/common/base/SpillConfig.h
```diff
@@ -151,6 +152,12 @@ struct SpillConfig {
   /// CompressionKind when spilling, CompressionKind_NONE means no compression.
   common::CompressionKind compressionKind;
 
+  // The max merge ways when merging sorted files into a single ordered stream.
+  // Nums < 2 means unlimited. This is used to cap the max opened file num when
+  // merging spilled sorted files to avoid using too much memory and causing
+  // OOM.
```
tanjialiang: The max number of files to merge at a time when merging sorted files into a single ordered stream. 0 means unlimited. This is used to reduce memory pressure by capping the number of open files when merging spilled sorted files to

velox/common/base/SpillConfig.h
```diff
@@ -151,6 +152,12 @@ struct SpillConfig {
...
+  uint64_t numMaxMergeWays;
```
tanjialiang: Should we have a check of this setting such that this shall never be allowed to set as 1 which is semantically wrong.

velox/common/base/SpillConfig.h
tanjialiang: Add a note saying this is only applicable for ordered spill.

velox/common/base/SpillConfig.h
tanjialiang: s/numMaxMergeWays/numMaxMergeFiles/

velox/common/base/SpillConfig.h
```diff
@@ -71,6 +71,7 @@ struct SpillConfig {
       uint64_t _maxSpillRunRows,
       uint64_t _writerFlushThresholdSize,
       const std::string& _compressionKind,
+      uint64_t numMaxMergeWays,
```
tanjialiang: maybe uint32_t ?

velox/common/base/SpillConfig.h
tanjialiang: Shall we use tripple slash instead as this is public?

velox/core/QueryConfig.h
```diff
@@ -299,6 +299,13 @@ class QueryConfig {
   static constexpr const char* kSpillCompressionKind =
       "spill_compression_codec";
 
+  // The max merge ways when merging sorted files into a single ordered stream.
+  // Nums < 2 means unlimited. This is used to cap the max opened file num when
+  // merging spilled sorted files to avoid using too much memory and causing
+  // OOM.
+  static constexpr const char* kSpillNumMaxMergeWays =
```
tanjialiang: update comments and name and make it tripple slash

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,52 @@
...
 #include "velox/common/testutil/TestValue.h"
+#include "velox/exec/OperatorUtils.h"
 #include "velox/serializers/PrestoSerializer.h"
 
 using facebook::velox::common::testutil::TestValue;
 
 namespace facebook::velox::exec {
+
+namespace utils {
+void gatherMerge(
```
tanjialiang, line 36: This shall be in cpp file as it is only used in the cpp file

HolyLow:
@xiaoxmeng @tanjialiang Could you help review this PR? Thanks a lot. Any suggestion is welcome.

velox/exec/Spill.h
```diff
@@ -490,6 +517,22 @@ class SpillPartition {
...
       folly::Synchronized<common::SpillStats>* spillStats);
 
+  /// Create an ordered stream reader from this spill partition. If the
+  /// partition has more than #mergeWayThreshold files, the files will be
+  /// pre-merged recursively to make sure the final ordered reader reads no
+  /// more than #mergeWayThreshold files. This behavior is to avoid OOM problem
+  /// when opening and reading too many files at the same time. If
+  /// #mergeWayThreshold < 2, the merge way is unlimited.
+  std::unique_ptr<TreeOfLosers<SpillMergeStream>>
+  createOrderedReaderWithPreMerge(
```
tanjialiang: Should we change the original createOrderedReader method and let the spill config decides if divide-and-conquer takes place?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+};
+using SpillFileInfoHeap = std::priority_queue<
+    SpillFileInfo,
+    std::vector<SpillFileInfo>,
+    SpillFileInfoComp>;
+} // namespace
+
+std::unique_ptr<TreeOfLosers<SpillMergeStream>>
+SpillPartition::createOrderedReaderWithPreMerge(
+    int mergeWayThreshold,
```
tanjialiang: uint32_t

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+    uint64_t writeBufferSize,
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
+  if (mergeWayThreshold <= 1 || files_.size() <= mergeWayThreshold) {
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  auto startFileNum = files_.size();
```
tanjialiang: const auto numTotalFiles

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+    files.clear();
+  }
+
+  files_.clear();
+  while (!orderedFiles.empty()) {
+    files_.push_back(orderedFiles.top());
+    orderedFiles.pop();
+  }
+  auto endFileNum = files_.size();
+  VLOG(1) << "For file " << files_[0].path << "pre merge from #" << startFileNum
```
tanjialiang: We can probably export through stats. Might not need the logging here.

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+  constexpr size_t maxBatchRows = 1'000;
+  constexpr size_t maxBatchBytes = 64 * 1024;
+  RowVectorPtr bufferVector = std::static_pointer_cast<RowVector>(
+      BaseVector::create(files_[0].type, maxBatchRows, pool));
+  std::vector<const RowVector*> bufferSpillSources(maxBatchRows);
+  std::vector<vector_size_t> bufferSpillSourceRows(maxBatchRows);
+
+  // Recursively merge the files.
+  int round = 0;
+  for (; orderedFiles.size() > mergeWayThreshold; round++) {
```
tanjialiang: for(auto round = 0; ...

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+    int mergeFileNum = std::min(
```
tanjialiang: const auto

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+        fileCreateConfig);
+    orderedFiles.push(mergedFile);
+    files.clear();
+  }
+
+  files_.clear();
+  while (!orderedFiles.empty()) {
+    files_.push_back(orderedFiles.top());
+    orderedFiles.pop();
+  }
```
tanjialiang, line 547: anyway to incoperate this in the above loop?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+      ++numEstimations;
+      estimateRowSizeSum += estimateRowSize.value();
+    }
+  }
+
+  if (numEstimations == 0) {
+    return maxRows;
+  }
+
+  const auto estimateRowSize =
```
tanjialiang, line 403: local var name reuse

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
   return std::make_unique<TreeOfLosers<SpillMergeStream>>(std::move(streams));
 }
 
+namespace {
+size_t estimateOutputBatchRowSize(
```
tanjialiang: consider returning optional to align with estimate row size functions in the codebase

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+  SpillFileInfoHeap orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(mergeWayThreshold);
+  auto filePathPrefix = files_[0].path;
+  constexpr size_t maxBatchRows = 1'000;
+  constexpr size_t maxBatchBytes = 64 * 1024;
+  RowVectorPtr bufferVector = std::static_pointer_cast<RowVector>(
+      BaseVector::create(files_[0].type, maxBatchRows, pool));
+  std::vector<const RowVector*> bufferSpillSources(maxBatchRows);
+  std::vector<vector_size_t> bufferSpillSourceRows(maxBatchRows);
```
tanjialiang: passing all these as individual parameters make the code less readable. can we simplify this by defining a struct and wrapping them in the struct?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
+  VELOX_CHECK_GT(files.size(), 0);
+  std::vector<std::unique_ptr<SpillMergeStream>> streams;
+  streams.reserve(files.size());
+  for (const auto& fileInfo : files) {
+    streams.push_back(FileSpillMergeStream::create(
+        SpillReadFile::create(fileInfo, readBufferSize, pool, spillStats)));
+  }
+  auto batchRows =
```
tanjialiang:
```suggestion
  const auto batchRows =
```

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,172 @@ SpillPartition::createOrderedReader(
...
+  std::vector<const RowVector*> bufferSpillSources(maxBatchRows);
+  std::vector<vector_size_t> bufferSpillSourceRows(maxBatchRows);
+
+  // Recursively merge the files.
+  int round = 0;
+  for (; orderedFiles.size() > mergeWayThreshold; round++) {
+    int mergeFileNum = std::min(
+        static_cast<size_t>(mergeWayThreshold),
+        orderedFiles.size() + 1 - mergeWayThreshold);
+    auto roundPathPrefix =
```
tanjialiang: consider constructing the string in-place

velox/common/base/SpillConfig.h
HolyLow: Refactored as suggested.

velox/common/base/SpillConfig.h
HolyLow: done.

velox/common/base/SpillConfig.h
HolyLow: done.

velox/core/QueryConfig.h
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: This function declaration is placed in header file explicitly to assist testing it in UTs, as suggested by @duanmeng previously.

velox/exec/Spill.h
HolyLow: I am not sure if it is proper to directly change the original `createOrderedReader` method's signature, as this new method takes a bunch of parameters. Do you think I should rename this function to `createOrderedReader` directly and move the original `createOrderedReader` method to private zone?

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: It might not be proper to integrate into the above loop, as the loop handles pre-merging and this code produces the output to-be-merge tree.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: The naming here might cause confusion. This function is to calculate the num of rows in a batch, while in other places `estimateRowSize` would calculate the size of a single row. Returning optional here is not proper, and I've renamed the function to prevent the confusion.

velox/exec/Spill.cpp
HolyLow: Packed into `BatchingBufferParams`. Not sure if the naming is proper enough, looking forward to your suggestions.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: I think constructing the string in-place would make the code less readable yet bring no performance improvement. The pathPrefix would be assembled to the final filename in the underlying `SpillFile` interfaces, and the assembling methods are using fmt method, which is consistent here.

velox/exec/Spill.cpp
HolyLow: Removed as suggested.

HolyLow:
@tanjialiang Thanks for your review, refactored as suggested. Any further suggestion is welcome, thanks a lot.

HolyLow:
@tanjialiang Do you have any further comments? Thanks a lot if you could help review this.

HolyLow:
@mbasmanova Could you please help find someone proper to move this PR forward? There is no further progress for over a month. Thanks a lot.

HolyLow:
@tanjialiang @mbasmanova  Do you have any further comments? It would be greatly appreciated if you could help review this or help find someone to move this PR forward.

tanjialiang:
> @tanjialiang @mbasmanova Do you have any further comments? It would be greatly appreciated if you could help review this or help find someone to move this PR forward.

Sorry for the delay. I'll revisit the edits by end of this week. Thanks!

velox/common/base/SpillConfig.h
```diff
@@ -151,6 +152,13 @@ struct SpillConfig {
   /// CompressionKind when spilling, CompressionKind_NONE means no compression.
   common::CompressionKind compressionKind;
 
+  /// The max number of files to merge at a time when merging sorted files into
+  /// a single ordered stream. 0 means unlimited. This is used to reduce memory
+  /// pressure by capping the number of open files when merging spilled sorted
+  /// files to avoid using too much memory and causing OOM. Note that this is
+  /// only applicable for ordered spill.
+  uint32_t numMaxMergeFiles;
```
tanjialiang, line 169: Shall we also mention that this only applies to some specific sort merge spill scenarios not for all spill scenarios (e.g. not for hash join)

velox/exec/Spill.h
```diff
@@ -490,6 +517,22 @@ class SpillPartition {
       memory::MemoryPool* pool,
       folly::Synchronized<common::SpillStats>* spillStats);
 
+  /// Create an ordered stream reader from this spill partition. If the
+  /// partition has more than #numMaxMergeFiles files, the files will be
```
tanjialiang: s/#numMaxMergeFiles/numMaxMergeFiles/

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+std::unique_ptr<TreeOfLosers<SpillMergeStream>>
+SpillPartition::createOrderedReaderWithPreMerge(
+    uint32_t numMaxMergeFiles,
+    uint64_t readBufferSize,
+    uint64_t writeBufferSize,
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
+  if (numMaxMergeFiles <= 1 || files_.size() <= numMaxMergeFiles) {
```
tanjialiang: Maybe VELOX_CHECK it could not be equal to 1 and only make the condition == 0?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  SpillFileInfoHeap orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(numMaxMergeFiles);
+  const auto filePathPrefix = files_[0].path;
+  constexpr size_t maxBatchRows = 1'000;
+  constexpr size_t maxBatchBytes = 64 * 1024;
```
tanjialiang: These constants are not used elsewhere other than passed into params. Can we make them default values in the params?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+  BatchingBufferParams bufferParams(
+      maxBatchRows, maxBatchBytes, files_[0].type, pool);
+
+  // Recursively merge the files.
+  for (int round = 0; orderedFiles.size() > numMaxMergeFiles; round++) {
```
tanjialiang: uint32_t round = 0

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+    const int numMergeFiles = std::min(
```
tanjialiang: const uint64_t

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+        static_cast<size_t>(numMaxMergeFiles),
```
tanjialiang: static_cast<uint64_t>

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+        orderedFiles.size() + 1 - numMaxMergeFiles);
+    auto roundPathPrefix =
+        fmt::format("{}-merge-round-{}", filePathPrefix, round);
+    // Choose the top #numMergeFiles smallest files for merging to minimize IO.
```
tanjialiang:
```suggestion
    // Choose the top 'numMergeFiles' smallest files for merging to minimize IO.
```

velox/exec/Spill.cpp
tanjialiang: This can help to reduce the need of additional unnecessary local variables, improving the cleanness of the code. Normally when local variables are not referred to for more than once we might not prefer having this local variable.

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+    for (int i = 0; i < numMergeFiles; i++) {
```
tanjialiang: ditto uint32_t

velox/exec/Spill.h
tanjialiang: Yeah I think that works.

velox/exec/Spill.cpp
tanjialiang: It is okay to not test this single method. This method is used in the algorithm so it shall be automatically tested by the ordered reader itself.
Exposing something that is not publicly used is generally not preferred. Another way of facilitating tests is to use friend class. But this method does not fit as a member method, either.

If you really want to test then you can have another method called testGatherMerge() that calls gatherMerge(). This conforms with the convention

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+      const vector_size_t _maxBatchRows,
+      const size_t _maxBatchBytes,
+      const TypePtr& type,
+      memory::MemoryPool* pool)
+      : maxBatchRows(_maxBatchRows), maxBatchBytes(_maxBatchBytes) {
+    rowVector = std::static_pointer_cast<RowVector>(
+        BaseVector::create(type, maxBatchRows, pool));
+    spillSources.resize(maxBatchRows);
+    spillSourceRows.resize(maxBatchRows);
+  }
```
tanjialiang: Shall we not have constructor for this Params? We can compose the struct by directly accessing its member variables.

velox/exec/Spill.cpp
tanjialiang: That makes sense. Thanks!

velox/exec/Spill.cpp
tanjialiang, line 36: Can we have some documentation on top of this method as well?

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+      pathPrefix,
+      std::numeric_limits<uint64_t>::max(),
+      writeBufferSize,
+      fileCreateConfig,
+      updateAndCheckSpillLimitCb,
+      pool,
+      spillStats);
+
+  while (mergeTree->next()) {
+    VectorPtr tmp = std::move(bufferParams.rowVector);
```
tanjialiang: auto tmpRowVector

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+    BaseVector::prepareForReuse(tmp, batchRows);
+    bufferParams.rowVector = std::static_pointer_cast<RowVector>(tmp);
+    for (auto& child : bufferParams.rowVector->children()) {
+      child->resize(batchRows);
+    }
+    int32_t outputRow = 0;
+    utils::gatherMerge(
```
tanjialiang: maybe we can make it return numRows

velox/exec/Spill.cpp
```diff
@@ -310,6 +351,179 @@ SpillPartition::createOrderedReader(
...
+        bufferParams.rowVector.get(),
+        mergeTree.get(),
+        outputRow,
+        bufferParams.spillSources,
+        bufferParams.spillSourceRows);
+
+    IndexRange range{0, outputRow};
```
tanjialiang, line 484: const

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,52 @@
...
+#include "velox/exec/OperatorUtils.h"
 #include "velox/serializers/PrestoSerializer.h"
 
 using facebook::velox::common::testutil::TestValue;
 
 namespace facebook::velox::exec {
+
+namespace utils {
+void gatherMerge(
+    RowVector* target,
```
tanjialiang: RowVectorPtr&

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,52 @@
...
+    TreeOfLosers<SpillMergeStream>* mergeTree,
```
tanjialiang: TreeOfLosers<SpillMergeStream>&

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,52 @@
...
+       currentStream = mergeTree->next()) {
+    bufferSources[outputSize] = &currentStream->current();
+    bufferSourceIndices[outputSize] =
+        currentStream->currentIndex(&isEndOfBatch);
+    ++outputSize;
+    if (FOLLY_UNLIKELY(isEndOfBatch)) {
+      // The stream is at end of input batch. Need to copy out the rows before
+      // fetching next batch in 'pop'.
+      gatherCopy(target, count, outputSize, bufferSources, bufferSourceIndices);
+      count += outputSize;
```
tanjialiang: Is it because the next batch will make prev batch payload owners to be destroyed with references no longer held?

velox/common/base/SpillConfig.h
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: Extract a `testGatherMerge` method a suggested.

velox/exec/Spill.h
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.h
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: I think having a constructor would be better, as the vectors inside the Params should be initialized. Without a constructor, we have to init the vectors explicitly when creating a Params object, which is easy to bring error. What do you think? Looking forward to your further suggestions, thanks a lot.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: Refactored the parameter's name, but the `auto` cannot be used here because the type deduction would bring trouble in compilation.

velox/exec/Spill.cpp
HolyLow: The param cannot be const as the following calling only accepts non-const param.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: Yes, it is, and you are right.

HolyLow:
@tanjialiang Thanks for your comments, refactored the code as suggested. Any further suggestion would be greatly appreciated.

velox/exec/Spill.cpp
HolyLow: The function's parameter `count` actually serves as the numRows return value. There are several output values including the output vectors, so we place all of them to the function params.

tanjialiang:
> @tanjialiang Thanks for your comments, refactored the code as suggested. Any further suggestion would be greatly appreciated.

Thanks @HolyLow ! I'll take a look shortly

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+  if (FOLLY_LIKELY(outputSize != 0)) {
+    gatherCopy(
+        target.get(), count, outputSize, bufferSources, bufferSourceIndices);
+    count += outputSize;
+    outputSize = 0;
+  }
+}
+} // namespace
+
+namespace test {
```
tanjialiang: The method is prepended with test. We can remove this namespace

velox/exec/Spill.cpp
```diff
@@ -310,6 +370,179 @@ SpillPartition::createOrderedReader(
...
+      writeBufferSize,
+      fileCreateConfig,
+      updateAndCheckSpillLimitCb,
+      pool,
+      spillStats);
+
+  while (mergeTree->next()) {
+    VectorPtr tmpRowVector = std::move(bufferParams.rowVector);
+    BaseVector::prepareForReuse(tmpRowVector, batchRows);
+    bufferParams.rowVector = std::static_pointer_cast<RowVector>(tmpRowVector);
```
tanjialiang: Curious why we did this moving and repointing?

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+}
+} // namespace
+
+namespace test {
+void testGatherMerge(
+    RowVectorPtr& target,
+    TreeOfLosers<SpillMergeStream>& mergeTree,
+    int32_t& count,
+    std::vector<const RowVector*>& bufferSources,
+    std::vector<vector_size_t>& bufferSourceIndices) {
```
tanjialiang, line 89: Consider to use raw_vector with memory pool provided

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+void gatherMerge(
+    RowVectorPtr& target,
+    TreeOfLosers<SpillMergeStream>& mergeTree,
+    int32_t& count,
+    std::vector<const RowVector*>& bufferSources,
+    std::vector<vector_size_t>& bufferSourceIndices) {
+  VELOX_CHECK_GE(bufferSources.size(), target->size());
+  VELOX_CHECK_GE(bufferSourceIndices.size(), target->size());
+  count = 0;
+  int32_t outputSize = 0;
```
tanjialiang: Maybe we don't need this? Resize the reusable vectors each call and use size() to get the num output?

HolyLow:
@tanjialiang Do you have any further comments? Any suggestion is greatly appreciated.

tanjialiang:
> @tanjialiang Do you have any further comments? Any suggestion is greatly appreciated.

Sorry for the delay. I am looking at it now.

velox/exec/Spill.cpp
```diff
@@ -313,6 +373,179 @@ SpillPartition::createOrderedReader(
...
+      pool,
+      spillStats);
+
+  while (mergeTree->next()) {
+    VectorPtr tmpRowVector = std::move(bufferParams.rowVector);
+    BaseVector::prepareForReuse(tmpRowVector, batchRows);
+    bufferParams.rowVector = std::static_pointer_cast<RowVector>(tmpRowVector);
+    for (auto& child : bufferParams.rowVector->children()) {
+      child->resize(batchRows);
+    }
```
tanjialiang: Shall we just call resize on the parent vector? It should do the same thing

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+/// callings.
+void gatherMerge(
+    RowVectorPtr& target,
+    TreeOfLosers<SpillMergeStream>& mergeTree,
+    int32_t& count,
+    std::vector<const RowVector*>& bufferSources,
+    std::vector<vector_size_t>& bufferSourceIndices) {
+  VELOX_CHECK_GE(bufferSources.size(), target->size());
+  VELOX_CHECK_GE(bufferSourceIndices.size(), target->size());
+  count = 0;
```
tanjialiang: s/count/totalNumRows/

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+  int32_t outputSize = 0;
+  bool isEndOfBatch = false;
```
tanjialiang: s/outputSize/numBatchRows/

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,71 @@
...
+    gatherCopy(
+        target.get(), count, outputSize, bufferSources, bufferSourceIndices);
+    count += outputSize;
+    outputSize = 0;
+  }
+}
+} // namespace
+
+namespace test {
+void testGatherMerge(
```
tanjialiang: testingGatherMerge

velox/exec/Spill.cpp
```diff
@@ -313,6 +373,179 @@ SpillPartition::createOrderedReader(
   return std::make_unique<TreeOfLosers<SpillMergeStream>>(std::move(streams));
 }
 
+namespace {
+size_t estimateOutputBatchRows(
+    const std::vector<std::unique_ptr<SpillMergeStream>>& streams,
+    const vector_size_t maxRows,
+    const size_t maxBytes) {
```
tanjialiang: primative types do not need to be const qualified

velox/exec/Spill.cpp
```diff
@@ -313,6 +373,179 @@ SpillPartition::createOrderedReader(
...
+  size_t numEstimations{0};
+  int64_t estimateRowSizeSum{0};
```
tanjialiang: s/estimateRowSizeSum/totalEstimatedBytes/

velox/exec/Spill.cpp
tanjialiang: Then we should make it a class instead of struct.

tanjialiang:
Hi @HolyLow  overall looks good % some nits.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: The reason is that the BaseVector::prepareForReuse only accepts VectorPtr as the input parameter while bufferParams.rowVector is of type RowVectorPtr. To manually align the types and prevent possible wrong semantic, the moving and repointing must be carried out. This mechanism is used in existing code, such as https://github.com/facebookincubator/velox/blob/261692efe1f52258ce3d312ef739b874575000be/velox/exec/HashProbe.cpp#L781-L783 .

velox/exec/Spill.cpp
HolyLow: I get what you suggest, but we are not using the raw_vector here for two reasons. 
Firstly, in `gatherMerge` function we would call a helper function called `gatherCopy` which only accepts std::vector as input parameters rather than raw_vector. There are many existing codes that called `gatherCopy` already, so it would not be easy to change the function's signature. 
Secondly, the buffer vectors here would not consume a large amount memory as each batch is limited to no more than 1k rows, and the vector's size would be limited to 8KB, which would be acceptable to use temporarily.

velox/exec/Spill.cpp
HolyLow: Yes, we need this actually. The semantic of `outputSize` is actually `numBatchRows` as suggested below, and it cannot be replaced by calling `size()`. Sorry for the bad naming if it causes confusion, refactored as suggested.

velox/exec/Spill.cpp
HolyLow: Yes, you are right. Refactored as suggested, thanks.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: I am not sure if it is proper to make it a class. For convenience, we are accessing all the member parameters directly, so the parameters would better stay public. But if we are making all the params public, I fail to see the reason why we still name it a class. Personally, I think having a struct that has its own constructor is acceptable, and it is only used here as a temporary helper parameter.
What do you think? Any further suggestion is welcome, and I am open to further changes. Thanks a lot.

HolyLow:
@tanjialiang Thanks for your comments, refactored as suggested. Any further suggestion would be greatly appreciated.

velox/exec/Spill.cpp
tanjialiang: that makes sense. thanks for the explanation.

tanjialiang:
@HolyLow could you help to rebase and make sure all tests pass? Thanks

HolyLow:
@tanjialiang Rebased code and fixed UTs as suggested. There is still one failing UT in a specific environment setting, but it seems that the UT is about `SharedArbitrator` and is not related to this PR. Besides, the ut succeeds under all the other settings, so I think it is an unstable UT for some reason.

Any further suggestion would be greatly appreciated.

velox/exec/GroupingSet.cpp
```diff
@@ -1137,7 +1137,13 @@ bool GroupingSet::prepareNextSpillPartitionOutput() {
   VELOX_CHECK_NE(outputSpillPartition_, it->first.partitionNumber());
   outputSpillPartition_ = it->first.partitionNumber();
   merge_ = it->second->createOrderedReader(
-      spillConfig_->readBufferSize, pool_, spillStats_);
+      spillConfig_->numMaxMergeFiles,
```
xiaoxmeng: Can we pass spillConfig_ to createOrderedReader?

velox/exec/SortBuffer.cpp
```diff
@@ -488,7 +488,13 @@ void SortBuffer::prepareOutputWithSpill() {
 
   VELOX_CHECK_EQ(spillPartitionSet_.size(), 1);
   spillMerger_ = spillPartitionSet_.begin()->second->createOrderedReader(
-      spillConfig_->readBufferSize, pool(), spillStats_);
+      spillConfig_->numMaxMergeFiles,
```
xiaoxmeng: ditto

velox/exec/Spill.h
```diff
@@ -33,6 +33,23 @@
...
+class SpillMergeStream;
+
+/// Testing gatherMerge without exposing the interface in the header. Used in
+/// test only. gatherMerge merges & sorts with the mergeTree and gatherCopy the
+/// results into target. 'target' is the result RowVector, and the copying
+/// starts from row 0 up to row target.size(). 'mergeTree' is the data source.
+/// 'totalNumRows' is the actual num of rows that is copied to target.
+/// 'bufferSources' and 'bufferSourceIndices' are buffering vectors that could
+/// be reused across calls.
+void testingGatherMerge(
```
xiaoxmeng, line 46: Can we move this to test util file instead of production code?

velox/exec/Spill.h
```diff
@@ -479,6 +505,24 @@ class SpillPartition {
       memory::MemoryPool* pool,
       folly::Synchronized<common::SpillStats>* spillStats);
 
+  /// Create an ordered stream reader from this spill partition. If the
+  /// partition has more than numMaxMergeFiles files, the files will be
+  /// pre-merged recursively to make sure the final ordered reader reads no
+  /// more than numMaxMergeFiles files. This behavior is to avoid OOM problem
+  /// when opening and reading too many files at the same time. If
+  /// numMaxMergeFiles < 2, the merge way is unlimited.
+  std::unique_ptr<TreeOfLosers<SpillMergeStream>> createOrderedReader(
```
xiaoxmeng, line 514: why need two createOrderedReader

velox/exec/Spill.h
```diff
@@ -80,6 +97,15 @@ class SpillMergeStream : public MergeStream {
     return decoded_[index];
   }
 
+  // Returns the estimated row size based on the vector received from the
+  // merge source.
+  std::optional<int64_t> estimateRowSize() const {
```
xiaoxmeng, line 102: Why we need this?

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+  }
+
+  SpillFileInfoHeap orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(numMaxMergeFiles);
+  const auto filePathPrefix = files_[0].path;
+  BatchingBufferParams bufferParams(files_[0].type, pool);
+
+  // Recursively merge the files.
+  for (uint32_t round = 0; orderedFiles.size() > numMaxMergeFiles; round++) {
```
xiaoxmeng: ++round

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+  VELOX_CHECK_EQ(resultFiles.size(), 1);
+  return std::move(resultFiles[0]);
+}
+
+struct SpillFileInfoComp {
+  bool operator()(const SpillFileInfo& lhs, const SpillFileInfo& rhs) const {
+    return lhs.size > rhs.size;
+  }
+};
+using SpillFileInfoHeap = std::priority_queue<
```
xiaoxmeng:
```
s/SpillFileInfoHeap/SpillFileHeap/
```

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+
+    IndexRange range{0, outputRow};
+    writer->write(bufferParams.rowVector, folly::Range<IndexRange*>(&range, 1));
+  }
+  auto resultFiles = writer->finish();
+  VELOX_CHECK_EQ(resultFiles.size(), 1);
+  return std::move(resultFiles[0]);
+}
+
+struct SpillFileInfoComp {
```
xiaoxmeng:
```
s/SpillFileInfoComp/SpillFileCompare/
```

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
+  VELOX_CHECK_NE(numMaxMergeFiles, 1);
+  if (numMaxMergeFiles == 0 || files_.size() <= numMaxMergeFiles) {
+    return createOrderedReader(readBufferSize, pool, spillStats);
+  }
+
+  SpillFileInfoHeap orderedFiles(files_.begin(), files_.end());
```
xiaoxmeng: This heap keeps small file first?

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+  SpillFiles files;
+  files.reserve(numMaxMergeFiles);
+  const auto filePathPrefix = files_[0].path;
```
xiaoxmeng:
```
s/filePathPrefix/mergeFilePathPrefix/
```

xiaoxmeng (commented):
@HolyLow thanks for the change % minors.

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+  for (uint32_t round = 0; orderedFiles.size() > numMaxMergeFiles; round++) {
+    const uint64_t numMergeFiles = std::min(
+        static_cast<uint64_t>(numMaxMergeFiles),
+        static_cast<uint64_t>(orderedFiles.size() + 1 - numMaxMergeFiles));
+    // Choose the top 'numMergeFiles' smallest files for merging to minimize IO.
+    for (uint32_t i = 0; i < numMergeFiles; i++) {
+      files.push_back(orderedFiles.top());
+      orderedFiles.pop();
+    }
+    auto mergedFile = mergeFiles(
```
xiaoxmeng:
```
s/mergeFiles/mergeSpillFiles/
```

velox/exec/Spill.cpp
```diff
@@ -18,11 +18,78 @@
...
+    RowVectorPtr& target,
+    TreeOfLosers<SpillMergeStream>& mergeTree,
+    int32_t& totalNumRows,
+    std::vector<const RowVector*>& bufferSources,
+    std::vector<vector_size_t>& bufferSourceIndices) {
+  VELOX_CHECK_GE(bufferSources.size(), target->size());
+  VELOX_CHECK_GE(bufferSourceIndices.size(), target->size());
+  totalNumRows = 0;
+  int32_t numBatchRows = 0;
+  bool isEndOfBatch = false;
```
xiaoxmeng:
```
s/isEndOfBatch/endOfBatch/
```

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,177 @@ SpillPartition::createOrderedReader(
...
+      writeBufferSize,
+      fileCreateConfig,
+      updateAndCheckSpillLimitCb,
+      pool,
+      spillStats);
+
+  while (mergeTree->next()) {
+    VectorPtr tmpRowVector = std::move(bufferParams.rowVector);
+    BaseVector::prepareForReuse(tmpRowVector, batchRows);
+    bufferParams.rowVector = std::static_pointer_cast<RowVector>(tmpRowVector);
```
xiaoxmeng: use check utility in Casts.h

velox/exec/GroupingSet.cpp
HolyLow: Refactored as suggested.

velox/exec/SortBuffer.cpp
HolyLow: done.

velox/exec/Spill.h
HolyLow: I'm afraid this is not viable. The testing method has to call the `gatherMerge` function directly, and we still have to export the `gatherMerge` function to some header file in production code one way or another.

velox/exec/Spill.h
HolyLow: Renamed the private one to `createOrderedReaderInternal`.

velox/exec/Spill.h
HolyLow: This function supports estimate the row size of this merge stream. The previous reviewers suggested that we ought to keep track of the rows' sizes to avoid gathering too many large rows which would consume a great amount of memory. So this function is added to help provide the functionality.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: Yes, you are right. We decide to merge the small files first to reduce IO. As explained above, suppose we have three files of size 1MB, 10MB, 100MB and we limit the maxMergeFiles to 2. Then we have to do 2 rounds of merging. For the first round, merging 1MB & 10MB files first into a 11MB file brings less IO than merging 10MB & 100MB files first.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

HolyLow:
@xiaoxmeng Thanks for your review, refactored as suggested.

Any further suggestion would be greatly appreciated.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.h
tanjialiang: Normally we call private implementation version functionImpl. So it might be nicer to call it createOrderedReaderImpl

velox/exec/Spill.h
xiaoxmeng: I see. Does this need to be public?

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,174 @@ SpillPartition::createOrderedReader(
...
+SpillFileInfo mergeSpillFiles(
+    const std::vector<SpillFileInfo>& files,
+    const std::string& pathPrefix,
+    uint64_t readBufferSize,
+    uint64_t writeBufferSize,
+    BatchingBufferParams& bufferParams,
+    const common::UpdateAndCheckSpillLimitCB& updateAndCheckSpillLimitCb,
+    memory::MemoryPool* pool,
+    folly::Synchronized<common::SpillStats>* spillStats,
+    const std::string& fileCreateConfig) {
```
xiaoxmeng: can you keep the input args first? Like spillStats is an output arg?

velox/exec/Spill.cpp
```diff
@@ -313,6 +380,174 @@ SpillPartition::createOrderedReader(
...
+  if (numMaxMergeFiles == 0 || files_.size() <= numMaxMergeFiles) {
+    return createOrderedReaderInternal(
+        spillConfig.readBufferSize, pool, spillStats);
+  }
+
+  SpillFileHeap orderedFiles(files_.begin(), files_.end());
+  SpillFiles files;
+  files.reserve(numMaxMergeFiles);
+  const auto mergeFilePathPrefix = files_[0].path;
+  BatchingBufferParams bufferParams(files_[0].type, pool);
```
xiaoxmeng:
```
s/BatchingBufferParams/SpillFileMergeParams/
```

velox/exec/Spill.h
HolyLow: Yes, it does. This member function is used in a helper function `estimateOutputBatchRows` within Spill.cpp, so it might not proper to declare the function as a friend function which exports the helper function in the header file. Therefore we have to make the function public. Besides, this function provides functionality that might be used in other places, so it makes sense to make the function public. Still, I am open to further suggestions on this. Thanks a lot.

velox/exec/Spill.cpp
HolyLow: done.

velox/exec/Spill.cpp
HolyLow: done.

HolyLow:
@xiaoxmeng Thanks for your review, refactored as suggested. Any further suggestion would be greatly appreciated.

tanjialiang:
Hi @HolyLow could you help to rebase so that I can reimport and help with landing this one.

HolyLow:
@tanjialiang Rebased as suggested. It would be greatly appreciated that you kindly help reimport the PR.

tanjialiang:
> @tanjialiang Rebased as suggested. It would be greatly appreciated that you kindly help reimport the PR.

Hi @HolyLow , I am still seeing "This branch is out-of-date with the base branch"
The tests did not seem to run for this PR. Could you help to double check if this is rebased onto the latest main? Thank you!

HolyLow:
@tanjialiang Thanks for your reply. I've rebased to the main branch several times each time I replied you, but as there was a non-negligible delay when you replied me, there were more commits merged into the latest main branch and made this branch out-of-date again. I've also manually rebased the code to the main branch, but it still didn't work.
It seems that the CI policy has changed, which causes that almost all the PRs outside of Meta is blocked for some reason. No actual test run is triggered. Can you give me some further guidance on this? Thanks a lot.
CC @xiaoxmeng @mbasmanova

tanjialiang:
> @tanjialiang Thanks for your reply. I've rebased to the main branch several times each time I replied you, but as there was a non-negligible delay when you replied me, there were more commits merged into the latest main branch and made this branch out-of-date again. I've also manually rebased the code to the main branch, but it still didn't work. It seems that the CI policy has changed, which causes that almost all the PRs outside of Meta is blocked for some reason. No actual test run is triggered. Can you give me some further guidance on this? Thanks a lot. CC @xiaoxmeng @mbasmanova

Hi @HolyLow , I just got to know that it turned out to be some internal issues that disables the tests from running. I'll take it from here. Thanks for your attempts to help and sorry about the confusion.

tanjialiang:
There's an ongoing build issue. Folks are working to get it addressed. Wait for https://github.com/facebookincubator/velox/pull/15658 before merging

velox/exec/tests/utils/MergeTestBase.h
```diff
@@ -22,6 +22,7 @@
 
 #include <gtest/gtest.h>
 
+#include <exec/Spill.h>
```
tanjialiang: Hi @HolyLow , could you change this include to use double quote include? This does not compile internally. Thank you

#include "velox/exec/Spill.h"

HolyLow: done.

tanjialiang:
@HolyLow Looks like there are some new compilation issues. Would you mind taking a look?

HolyLow:
> @HolyLow Looks like there are some new compilation issues. 

@tanjialiang Fixed that again. PTAL.

PR #14145, opened 2025-07-18
refactor: Reduce usage of dynamic_cast in ExprCompiler

mbasmanova:
Summary:
ExprCompiler uses dynamic_cast to switch on expr type. This is inefficient and shows up in the profile.

Introduce ExprKind enum and ITypedExpr::kind() getter to allow for more efficient switching.

Also, introduce SpecialFormKind enum.

Also, remove unnecessary includes.

Differential Revision: D78553108


xiaoxmeng (approved):
@mbasmanova thanks for the optimizations!

PR #14146, opened 2025-07-18
feat: Collect iceberg data file statistics during insertion

PingLiuPing:
Collect iceberg data file statistics during insertion.
This PR only collect the statistics during insertion, the statistics are not been used for now. After the insertion and partition transforms PR been merged, will add that part into this PR.

### DIFF
--- a/velox/dwio/common/IcebergStatistics.h
+++ b/velox/dwio/common/IcebergStatistics.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/json/dynamic.h>
+
+namespace facebook::velox::dwio::common {
+
+// Iceberg data_file struct fields.
+struct IcebergDataFileStatistics {
+  int64_t numRecords;
+  std::unordered_map<int32_t, int64_t> columnsSizes;
+  std::unordered_map<int32_t, int64_t> valueCounts;
+  std::unordered_map<int32_t, int64_t> nullValueCounts;
+  std::unordered_map<int32_t, int64_t> nanValueCounts;
+  std::unordered_map<int32_t, std::string> lowerBounds;
+  std::unordered_map<int32_t, std::string> upperBounds;
+
+  // Split offsets for the data file. For example, all row
+  // group offsets in a Parquet file. Must be sorted ascending.
+  std::vector<uint64_t> splitOffsets;
+
+  IcebergDataFileStatistics() : numRecords(0) {}
+
+  folly::dynamic toJson() const;
+};
+
+} // namespace facebook::velox::dwio::common
--- a/velox/dwio/common/Options.h
+++ b/velox/dwio/common/Options.h
@@ -684,6 +684,15 @@ class ReaderOptions : public io::ReaderOptions {
   bool allowEmptyFile_{false};
 };
 
+/// Settings for collecting Iceberg column statistics. Holds the Iceberg
+/// source column ID (field_id) and whether to skip bounds collection for this
+/// field. For nested field, it contains child fields.
+struct IcebergStatsSettings {
+  int32_t id;
+  bool skipBounds;
+  std::vector<IcebergStatsSettings> children;
+};
+
 struct WriterOptions {
   TypePtr schema{nullptr};
   velox::memory::MemoryPool* memoryPool{nullptr};
@@ -705,6 +714,13 @@ struct WriterOptions {
 
   std::string sessionTimezoneName;
   bool adjustTimestampToTimezone{false};
+
+  /// This option controls whether collect iceberg data file statistics
+  /// during write. It will be set to true only by IcebergDataSink.
+  bool collectIcebergDataFileStats{false};
+
+  std::shared_ptr<std::vector<IcebergStatsSettings>> icebergStatsSetting{
+      nullptr};
 
   // WriterOption implementations can implement this function to specify how to
   // process format-specific session and connector configs.
--- a/velox/dwio/parquet/common/UnicodeUtil.h
+++ b/velox/dwio/parquet/common/UnicodeUtil.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace facebook::velox::parquet {
+
+class UnicodeUtil {
+ public:
+  static std::string truncateStringMin(
+      const char* input,
+      int32_t inputLength,
+      int32_t numCodePoints);
+
+  static std::string truncateStringMax(
+      const char* input,
+      int32_t inputLength,
+      int32_t numCodePoints);
+
+ private:
+  UnicodeUtil() = delete;
+};
+
+} // namespace facebook::velox::parquet
--- a/velox/dwio/parquet/thrift/ParquetThriftTypes.h
+++ b/velox/dwio/parquet/thrift/ParquetThriftTypes.h
@@ -466,12 +466,14 @@ typedef struct _Statistics__isset {
         min(false),
         null_count(false),
         distinct_count(false),
+        nan_count(false),
         max_value(false),
         min_value(false) {}
   bool max : 1;
   bool min : 1;
   bool null_count : 1;
   bool distinct_count : 1;
+  bool nan_count : 1;
   bool max_value : 1;
   bool min_value : 1;
 } _Statistics__isset;
@@ -489,6 +491,7 @@ class Statistics : public virtual apache::thrift::TBase {
         min(),
         null_count(0),
         distinct_count(0),
+        nan_count(0),
         max_value(),
         min_value() {}
 
@@ -516,6 +519,11 @@ class Statistics : public virtual apache::thrift::TBase {
    * count of distinct values occurring
    */
   int64_t distinct_count;
+
+  /**
+   * count of NaN values occurring
+   */
+  int64_t nan_count;
   /**
    * Min and max values for the column, determined by its ColumnOrder.
    *
@@ -534,6 +542,8 @@ class Statistics : public virtual apache::thrift::TBase {
   void __set_null_count(const int64_t val);
 
   void __set_distinct_count(const int64_t val);
+
+  void __set_nan_count(const int64_t val);
 
   void __set_max_value(const std::string& val);
 
@@ -555,6 +565,10 @@ class Statistics : public virtual apache::thrift::TBase {
     if (__isset.distinct_count != rhs.__isset.distinct_count)
       return false;
     else if (__isset.distinct_count && !(distinct_count == rhs.distinct_count))
+      return false;
+    if (__isset.nan_count != rhs.__isset.nan_count)
+      return false;
+    else if (__isset.nan_count && !(nan_count == rhs.nan_count))
       return false;
     if (__isset.max_value != rhs.__isset.max_value)
       return false;
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -20,9 +20,13 @@
 #include <arrow/table.h>
 #include "velox/common/base/Pointers.h"
 #include "velox/common/config/Config.h"
+#include "velox/common/encode/Base64.h"
 #include "velox/common/testutil/TestValue.h"
 #include "velox/core/QueryConfig.h"
+#include "velox/dwio/parquet/writer/arrow/ArrowSchema.h"
+#include "velox/dwio/parquet/writer/arrow/Metadata.h"
 #include "velox/dwio/parquet/writer/arrow/Properties.h"
+#include "velox/dwio/parquet/writer/arrow/Statistics.h"
 #include "velox/dwio/parquet/writer/arrow/Writer.h"
 #include "velox/exec/MemoryReclaimer.h"
 
@@ -193,47 +197,74 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
   }
 }
 
-std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
+std::shared_ptr<::arrow::Field> updateFieldRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const dwio::common::IcebergStatsSettings* icebergStatsSetting = nullptr) {
+  std::shared_ptr<::arrow::Field> newField{nullptr};
   if (type.isRow()) {
     auto& rowType = type.asRow();
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
     auto childrenSize = rowType.size();
     std::vector<std::shared_ptr<::arrow::Field>> newFields;
-    newFields.reserve(childrenSize);
-    for (auto i = 0; i < childrenSize; i++) {
-      newFields.push_back(updateFieldNameRecursive(
-          structType->fields()[i], *rowType.childAt(i), rowType.nameOf(i)));
+    newFields.reserve(rowType.size());
+    for (auto i = 0; i < rowType.size(); ++i) {
+      const auto* childSetting =
+          icebergStatsSetting ? &icebergStatsSetting->children.at(i) : nullptr;
+      newFields.push_back(updateFieldRecursive(
+          structType->fields()[i],
+          *rowType.childAt(i),
+          rowType.nameOf(i),
+          childSetting));
     }
-    return newField->WithType(::arrow::struct_(newFields));
+    newField = newField->WithType(::arrow::struct_(newFields));
   } else if (type.isArray()) {
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto listType =
         std::dynamic_pointer_cast<::arrow::BaseListType>(newField->type());
     auto elementType = type.asArray().elementType();
     auto elementField = listType->value_field();
-    return newField->WithType(
-        ::arrow::list(updateFieldNameRecursive(elementField, *elementType)));
+    const auto* childSetting =
+        icebergStatsSetting ? &icebergStatsSetting->children.at(0) : nullptr;
+
+    auto updatedElementField =
+        updateFieldRecursive(elementField, *elementType, name, childSetting);
+    newField = newField->WithType(::arrow::list(updatedElementField));
   } else if (type.isMap()) {
     auto mapType = type.asMap();
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto arrowMapType =
         std::dynamic_pointer_cast<::arrow::MapType>(newField->type());
-    auto newKeyField =
-        updateFieldNameRecursive(arrowMapType->key_field(), *mapType.keyType());
-    auto newValueField = updateFieldNameRecursive(
-        arrowMapType->item_field(), *mapType.valueType());
-    return newField->WithType(
-        ::arrow::map(newKeyField->type(), newValueField->type()));
+    const auto* keySetting =
+        icebergStatsSetting ? &icebergStatsSetting->children.at(0) : nullptr;
+    const auto* valueSetting =
+        icebergStatsSetting ? &icebergStatsSetting->children.at(1) : nullptr;
+    auto newKeyField = updateFieldRecursive(
+        arrowMapType->key_field(),
+        *mapType.keyType(),
+        mapType.nameOf(0),
+        keySetting);
+    auto newValueField = updateFieldRecursive(
+        arrowMapType->item_field(),
+        *mapType.valueType(),
+        mapType.nameOf(1),
+        valueSetting);
+    newField = newField->WithType(
+        std::make_shared<::arrow::MapType>(newKeyField, newValueField));
   } else if (name != "") {
-    return field->WithName(name);
+    newField = field->WithName(name);
   } else {
-    return field;
+    newField = field;
   }
+
+  if (icebergStatsSetting) {
+    newField = newField->WithMetadata(
+        arrow::arrow::FieldIdMetadata(icebergStatsSetting->id));
+  }
+  return newField;
 }
 
 std::optional<TimestampPrecision> getTimestampUnit(
@@ -350,6 +381,9 @@ Writer::Writer(
       getArrowParquetWriterOptions(options, flushPolicy_);
   setMemoryReclaimers();
   writeInt96AsTimestamp_ = options.writeInt96AsTimestamp;
+  dataFileStats_ = std::make_shared<dwio::common::IcebergDataFileStatistics>();
+  icebergStatsSetting_ = std::move(options.icebergStatsSetting);
+  collectIcebergStats_ = options.collectIcebergDataFileStats;
 }
 
 Writer::Writer(
@@ -445,8 +479,12 @@ void Writer::write(const VectorPtr& data) {
   std::vector<std::shared_ptr<::arrow::Field>> newFields;
   auto childSize = schema_->size();
   for (auto i = 0; i < childSize; i++) {
-    newFields.push_back(updateFieldNameRecursive(
-        arrowSchema->fields()[i], *schema_->childAt(i), schema_->nameOf(i)));
+    newFields.push_back(updateFieldRecursive(
+        arrowSchema->fields()[i],
+        *schema_->childAt(i),
+        schema_->nameOf(i),
+        this->icebergStatsSetting_ ? &this->icebergStatsSetting_->at(i)
+                                   : nullptr));
   }
 
   PARQUET_ASSIGN_OR_THROW(
@@ -490,6 +528,9 @@ void Writer::close() {
 
   if (arrowContext_->writer) {
     PARQUET_THROW_NOT_OK(arrowContext_->writer->Close());
+    if (collectIcebergStats_) {
+      collectIcebergDataFileStats(arrowContext_->writer->metadata());
+    }
     arrowContext_->writer.reset();
   }
   PARQUET_THROW_NOT_OK(stream_->Close());
@@ -500,6 +541,96 @@ void Writer::close() {
 void Writer::abort() {
   stream_->abort();
   arrowContext_.reset();
+}
+
+void Writer::collectIcebergDataFileStats(
+    const std::shared_ptr<arrow::FileMetaData>& fileMetadata) {
+  if (!icebergStatsSetting_) {
+    return;
+  }
+  VELOX_CHECK_NOT_NULL(fileMetadata);
+  dataFileStats_->numRecords = fileMetadata->num_rows();
+
+  std::unordered_set<int32_t> skipBoundsFields;
+
+  std::function<int32_t(const dwio::common::IcebergStatsSettings&)>
+      processFields =
+          [&skipBoundsFields, &processFields](
+              const dwio::common::IcebergStatsSettings& field) -> int32_t {
+    if (field.skipBounds) {
+      skipBoundsFields.insert(field.id);
+    }
+    if (field.children.empty()) {
+      return 1;
+    }
+    int32_t count = 0;
+    for (const auto& child : field.children) {
+      count += processFields(child);
+    }
+    return count;
+  };
+
+  // numColumns is not the number of columns in Iceberg table's schema,
+  // e.g., schema_->size(). It also contains the sub-fields when there are
+  // nested types in table's schema.
+  int32_t numColumns = 0;
+  for (const auto& field : *icebergStatsSetting_) {
+    numColumns += processFields(field);
+  }
+
+  std::unordered_map<int32_t, std::shared_ptr<arrow::Statistics>>
+      globalMinStats;
+  std::unordered_map<int32_t, std::shared_ptr<arrow::Statistics>>
+      globalMaxStats;
+
+  const auto numRowGroups = fileMetadata->num_row_groups();
+  for (auto i = 0; i < numRowGroups; ++i) {
+    const auto rgm = fileMetadata->RowGroup(i);
+    VELOX_CHECK_EQ(numColumns, rgm->num_columns());
+    dataFileStats_->splitOffsets.emplace_back(rgm->file_offset());
+
+    for (auto j = 0; j < numColumns; ++j) {
+      const auto columnChunkMetadata = rgm->ColumnChunk(j);
+      const auto fieldId = columnChunkMetadata->field_id();
+      const auto numValues = columnChunkMetadata->num_values();
+
+      dataFileStats_->valueCounts[fieldId] += numValues;
+      dataFileStats_->columnsSizes[fieldId] +=
+          columnChunkMetadata->total_compressed_size();
+
+      const auto columnChunkStats = columnChunkMetadata->statistics();
+      if (columnChunkStats->nan_count() > 0) {
+        dataFileStats_->nanValueCounts[fieldId] +=
+            columnChunkStats->nan_count();
+      }
+      dataFileStats_->nullValueCounts[fieldId] +=
+          columnChunkStats->null_count();
+
+      if (columnChunkStats->HasMinMax() &&
+          !skipBoundsFields.contains(fieldId)) {
+        if (globalMaxStats.find(fieldId) == globalMaxStats.end()) {
+          globalMinStats[fieldId] = columnChunkStats;
+          globalMaxStats[fieldId] = columnChunkStats;
+        } else {
+          globalMaxStats[fieldId] = arrow::Statistics::CompareAndGetMax(
+              globalMaxStats[fieldId], columnChunkStats);
+          globalMinStats[fieldId] = arrow::Statistics::CompareAndGetMin(
+              globalMinStats[fieldId], columnChunkStats);
+        }
+      }
+    }
+  }
+
+  for (const auto& [fieldId, minStats] : globalMinStats) {
+    const auto lowerBound = minStats->MinValue();
+    dataFileStats_->lowerBounds[fieldId] =
+        encoding::Base64::encode(lowerBound.data(), lowerBound.size());
+  }
+  for (const auto& [fieldId, maxStats] : globalMaxStats) {
+    const auto upperBound = maxStats->MaxValue();
+    dataFileStats_->upperBounds[fieldId] =
+        encoding::Base64::encode(upperBound.data(), upperBound.size());
+  }
 }
 
 void Writer::setMemoryReclaimers() {
--- a/velox/dwio/parquet/writer/Writer.h
+++ b/velox/dwio/parquet/writer/Writer.h
@@ -24,6 +24,7 @@
 #include "velox/dwio/common/Options.h"
 #include "velox/dwio/common/Writer.h"
 #include "velox/dwio/common/WriterFactory.h"
+#include "velox/dwio/parquet/writer/arrow/Metadata.h"
 #include "velox/dwio/parquet/writer/arrow/Types.h"
 #include "velox/dwio/parquet/writer/arrow/util/Compression.h"
 #include "velox/vector/ComplexVector.h"
@@ -195,6 +196,10 @@ class Writer : public dwio::common::Writer {
  private:
   // Sets the memory reclaimers for all the memory pools used by this writer.
   void setMemoryReclaimers();
+
+  // Collect Iceberg data file statistics from FileMetaData.
+  void collectIcebergDataFileStats(
+      const std::shared_ptr<arrow::FileMetaData>& fileMetadata);
 
   // Checks if the input data contains a nested wrapped vector or complex
   // vector. If so, flatten the input to make it compatible with
--- a/velox/dwio/parquet/writer/arrow/Statistics.cpp
+++ b/velox/dwio/parquet/writer/arrow/Statistics.cpp
@@ -35,10 +35,13 @@
 #include "arrow/visit_data_inline.h"
 
 #include "velox/common/base/Exceptions.h"
+#include "velox/dwio/parquet/common/UnicodeUtil.h"
 #include "velox/dwio/parquet/writer/arrow/Encoding.h"
 #include "velox/dwio/parquet/writer/arrow/Exception.h"
 #include "velox/dwio/parquet/writer/arrow/Platform.h"
 #include "velox/dwio/parquet/writer/arrow/Schema.h"
+#include "velox/type/DecimalUtil.h"
+#include "velox/type/HugeInt.h"
 
 using arrow::default_memory_pool;
 using arrow::MemoryPool;
@@ -547,6 +550,17 @@ TypedComparatorImpl<false, ByteArrayType>::GetMinMax(
   return GetMinMaxBinaryHelper<false>(*this, values);
 }
 
+template <typename T>
+std::string encodeDecimalToBigEndian(T value) {
+  uint8_t buffer[sizeof(T)];
+  if constexpr (std::is_same_v<T, int64_t>) {
+    *reinterpret_cast<int64_t*>(buffer) = ::arrow::bit_util::ToBigEndian(value);
+  } else if constexpr (std::is_same_v<T, int128_t>) {
+    *reinterpret_cast<int128_t*>(buffer) = DecimalUtil::bigEndian(value);
+  }
+  return std::string(reinterpret_cast<const char*>(buffer), sizeof(T));
+}
+
 template <typename DType>
 class TypedStatisticsImpl : public TypedStatistics<DType> {
  public:
@@ -590,9 +604,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
       int64_t num_values,
       int64_t null_count,
       int64_t distinct_count,
+      int64_t nan_count,
       bool has_min_max,
       bool has_null_count,
       bool has_distinct_count,
+      bool has_nan_count,
       MemoryPool* pool)
       : TypedStatisticsImpl(descr, pool) {
     TypedStatisticsImpl::IncrementNumValues(num_values);
@@ -607,6 +623,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
       has_distinct_count_ = false;
     }
 
+    if (has_nan_count) {
+      IncrementNaNValues(nan_count);
+    } else {
+      has_nan_count_ = false;
+    }
+
     if (!encoded_min.empty()) {
       PlainDecode(encoded_min, &min_);
     }
@@ -619,6 +641,11 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   bool HasDistinctCount() const override {
     return has_distinct_count_;
   };
+
+  bool HasNaNCount() const override {
+    return has_nan_count_;
+  };
+
   bool HasMinMax() const override {
     return has_min_max_;
   }
@@ -635,6 +662,12 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     num_values_ += n;
   }
 
+  void IncrementNaNValues(int64_t n) override {
+    if (n > 0) {
+      statistics_.nan_count += n;
+      has_nan_count_ = true;
+    }
+  }
   bool Equals(const Statistics& raw_other) const override {
     if (physical_type() != raw_other.physical_type())
       return false;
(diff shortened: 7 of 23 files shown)


velox/dwio/common/Statistics.h
```diff
@@ -596,4 +597,59 @@ struct RuntimeStatistics {
...
+struct IcebergDataFileStatistics {
+  int64_t numRecords;
+  std::unordered_map<int32_t, int64_t> columnsSizes;
+  std::unordered_map<int32_t, int64_t> valueCounts;
+  std::unordered_map<int32_t, int64_t> nullValueCounts;
+  std::unordered_map<int32_t, int64_t> nanValueCounts;
+  std::unordered_map<int32_t, std::string> lowerBounds;
+  std::unordered_map<int32_t, std::string> upperBounds;
+
+  IcebergDataFileStatistics() : numRecords(0) {}
```
jinchengchenghh: Why numRecords has default value?

velox/dwio/common/Statistics.h
```diff
@@ -596,4 +597,59 @@ struct RuntimeStatistics {
   }
 };
 
+struct IcebergDataFileStatistics {
```
jinchengchenghh: Comment on the class, and move it to IcebergStatistic.h and IcebergStatistic.cpp

velox/dwio/common/Statistics.h
```diff
@@ -596,4 +597,59 @@ struct RuntimeStatistics {
...
+
+  IcebergDataFileStatistics() : numRecords(0) {}
+
+  folly::dynamic toJson() const {
+    folly::dynamic json = folly::dynamic::object;
+    json["recordCount"] = numRecords;
+
+    folly::dynamic sizeJson = folly::dynamic::object;
+    for (const auto& pair : columnsSizes) {
+      sizeJson[folly::to<std::string>(pair.first)] = pair.second;
```
jinchengchenghh: Does folly cannot support int as key?

velox/dwio/common/Statistics.h
```diff
@@ -596,4 +597,59 @@ struct RuntimeStatistics {
...
+    }
+    json["nullValueCounts"] = std::move(nullCountJson);
+
+    folly::dynamic nanCountJson = folly::dynamic::object;
+    for (const auto& pair : nanValueCounts) {
+      nanCountJson[folly::to<std::string>(pair.first)] = pair.second;
+    }
+    json["nanValueCounts"] = std::move(nanCountJson);
+
+    folly::dynamic lowerBoundJson = folly::dynamic::object;
```
jinchengchenghh: Use #define or lambda function to reduce the duplicate code

velox/dwio/parquet/common/UnicodeUtil.h
```diff
@@ -0,0 +1,127 @@
...
+ private:
+  UnicodeUtil() = delete;
+
+ public:
+  static bool isCharHighSurrogate(char16_t ch) {
+    return (ch & 0xFC00) == 0xD800;
+  }
+
+  // Truncates a string to the specified number of Unicode code points.
+  static std::string truncateString(
```
jinchengchenghh: Is it similar to cappedByteLength?

velox/dwio/parquet/common/UnicodeUtil.h
```diff
@@ -0,0 +1,127 @@
...
+      int charLength;
+      utf8proc_codepoint(current, end, charLength);
+      current += charLength;
+      currentCodePoint++;
+    }
+    return input.substr(0, current - data);
+  }
+
+  static std::optional<std::string> truncateStringMin(
+      const std::optional<std::string>& input,
```
jinchengchenghh: Move implementation to cpp file and add comments

velox/dwio/parquet/thrift/ParquetThriftTypes.h
```diff
@@ -467,13 +467,15 @@ typedef struct _Statistics__isset {
         null_count(false),
         distinct_count(false),
         max_value(false),
-        min_value(false) {}
+        min_value(false),
+        nan_count(false) {}
```
jinchengchenghh: Move nan_count after distinct_count

velox/dwio/parquet/thrift/ParquetThriftTypes.h
```diff
@@ -467,13 +467,15 @@ typedef struct _Statistics__isset {
...
   bool max : 1;
   bool min : 1;
   bool null_count : 1;
   bool distinct_count : 1;
   bool max_value : 1;
   bool min_value : 1;
+  bool nan_count : 1;
```
jinchengchenghh: ditto

velox/dwio/parquet/thrift/ParquetThriftTypes.h
```diff
@@ -516,6 +519,11 @@ class Statistics : public virtual apache::thrift::TBase {
    * count of distinct values occurring
    */
   int64_t distinct_count;
+
+  /**
```
jinchengchenghh, line 523: //

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +508,62 @@ void Writer::abort() {
   arrowContext_.reset();
 }
 
+void Writer::collectIcebergDataFileStats(
+    const std::shared_ptr<arrow::FileMetaData>& fileMetadata) {
```
jinchengchenghh, line 547: const std::shared_ptr<arrow::FileMetaData>& fileMetadata -> const arrow::FileMetaData& fileMetadata

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +508,62 @@ void Writer::abort() {
...
+  VELOX_CHECK_NOT_NULL(fileMetadata);
```
jinchengchenghh, line 551: Use schema_ to get the numColumns

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +508,62 @@ void Writer::abort() {
...
+  globalMinStats.reserve(rgm->num_columns());
+  globalMaxStats.reserve(rgm->num_columns());
+  std::vector<bool> hasMinMax(rgm->num_columns(), false);
+
+  const auto numRowGroups = fileMetadata->num_row_groups();
+  for (auto i = 0; i < numRowGroups; ++i) {
+    rgm = arrowContext_->writer->metadata()->RowGroup(i);
+    std::vector<std::shared_ptr<arrow::Statistics>> columnStats;
+    for (auto j = 0; j < rgm->num_columns(); ++j) {
+      const auto numValues = rgm->ColumnChunk(j)->num_values();
```
jinchengchenghh: Does the column map key starts with 1? The key is fieldId in velox, how do you support complex data type?
```
  /**
   * Get the number of all values, including nulls, NaN and repeated.
   *
   * @return a Map of fieldId to the number of all values including nulls, NaN and repeated
   */
  public Map<Integer, Long> valueCounts() {
    return valueCounts;
  }

```

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +508,62 @@ void Writer::abort() {
...
+          globalMaxStats[j] = arrow::Statistics::CompareAndGetMax(
+              globalMaxStats[j], columnChunkStats);
+          globalMinStats[j] = arrow::Statistics::CompareAndGetMin(
+              globalMinStats[j], columnChunkStats);
+        }
+      }
+    }
+  }
+  for (auto j = 0; j < rgm->num_columns(); ++j) {
+    if (hasMinMax[j]) {
```
jinchengchenghh: Some data type is not needed to encode, check type first

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +767,60 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
 
+  std::string MinValue() const override {
+    if constexpr (std::is_same_v<T, int64_t>) {
+      if (descr_->logical_type()->is_decimal()) {
+        int64_t minValue = min_;
+        uint8_t buffer[sizeof(int64_t)];
+        // Convert to big-endian.
+        *reinterpret_cast<int64_t*>(buffer) =
+            ::arrow::bit_util::ToBigEndian(minValue);
+        return std::string(
```
jinchengchenghh: Do not need the minBytes? Use sizeof(int64_t)?

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +767,60 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     return s;
   }
 
+  std::string MinValue() const override {
+    if constexpr (std::is_same_v<T, int64_t>) {
```
jinchengchenghh, line 783: int128_t support

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -768,12 +850,28 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   int64_t num_values() const override {
     return num_values_;
   }
+  int64_t nan_count() const override {
```
jinchengchenghh, line 861: empty line above

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -783,6 +881,7 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   // a statistics thrift message which doesn't have the optional null_count,
   // `num_values_` may include null values.
   int64_t num_values_ = 0;
+  int64_t num_nulls_ = 0;
```
jinchengchenghh: Why do you need num_nulls_

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -803,6 +902,7 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   void ResetCounts() {
     this->statistics_.null_count = 0;
     this->statistics_.distinct_count = 0;
+    this->statistics_.nan_count = 0;
     this->num_values_ = 0;
```
jinchengchenghh, line 911: Also reset num_nulls_

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
           max_buffer_.get());
     }
   }
+
+  int64_t CountNaN(const T* values, const int64_t length) {
+    if constexpr (!std::is_floating_point_v<T>) {
+      return 0;
+    } else {
```
jinchengchenghh, line 954: remove else

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
           max_buffer_.get());
     }
   }
+
+  int64_t CountNaN(const T* values, const int64_t length) {
```
jinchengchenghh: const int64_t length -> int64_t length

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -224,13 +233,15 @@ class PARQUET_EXPORT Statistics {
   /// \param[in] descr the column schema
   /// \param[in] encoded_min the encoded minimum value
   /// \param[in] encoded_max the encoded maximum value
-  /// \param[in] num_values total number of values
+  /// \param[in] num_values total number of valuesg
```
jinchengchenghh: typo

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -224,13 +233,15 @@ class PARQUET_EXPORT Statistics {
...
   /// \param[in] null_count number of null values
   /// \param[in] distinct_count number of distinct values
+  /// \param[in] nan_count number of nan values
   /// \param[in] has_min_max whether the min/max statistics are set
   /// \param[in] has_null_count whether the null_count statistics are set
   /// \param[in] has_distinct_count whether the distinct_count statistics are
-  /// set \param[in] pool a memory pool to use for any memory allocations,
+  /// set \param[in] has_nan_count whether the nan_count statistics are set
```
jinchengchenghh, line 243: Why starts with set

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+  }
+
+  int64_t CountNaNSpaced(
+      const T* values,
+      const int64_t length,
+      const uint8_t* valid_bits,
+      const int64_t valid_bits_offset) {
+    if constexpr (!std::is_floating_point_v<T>) {
+      return 0;
+    } else {
```
jinchengchenghh, line 973: ditto

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      int64_t count = 0;
+      ::arrow::internal::VisitSetBitRunsVoid(
+          valid_bits,
+          valid_bits_offset,
+          length,
+          [&](int64_t position, const int64_t run_length) {
```
jinchengchenghh: drop const

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      }
+      return count;
+    }
+  }
+
+  int64_t CountNaNSpaced(
+      const T* values,
+      const int64_t length,
+      const uint8_t* valid_bits,
+      const int64_t valid_bits_offset) {
```
jinchengchenghh: ditto

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -841,6 +942,46 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+          count++;
+        }
+      }
+      return count;
+    }
+  }
+
+  int64_t CountNaNSpaced(
+      const T* values,
+      const int64_t length,
```
jinchengchenghh: ditto

velox/dwio/common/Statistics.h
PingLiuPing: Thanks.

velox/dwio/common/Statistics.h
PingLiuPing: Ok.

velox/dwio/parquet/common/UnicodeUtil.h
PingLiuPing: Yes, I change it to use cappedByteLength.

velox/dwio/parquet/common/UnicodeUtil.h
PingLiuPing: thanks

velox/dwio/parquet/thrift/ParquetThriftTypes.h
PingLiuPing: ok

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks. I think `const std::shared_ptr<arrow::FileMetaData>& fileMetadata` works well here, if you look at the method FileWriter::metadata, it returns `const std::shared_ptr<FileMetaData>` and I can directly use this value.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: thanks,

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Yes, when passing back these stats, the column filedId start from 1.

PingLiuPing: The filedId is unique in iceberg, it is Ok to use it as a key here. Even it is a complex data type, each sub field has a unique filedId.

velox/dwio/common/Statistics.h
PingLiuPing: It can compile, but during execution it throws exception
`
C++ exception with description "folly::toJson: JSON object key 1 was not a string when serializing key at "metrics"->"upperBounds"->1" thrown in the test body.`

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, actually all min and max should be stored as binary format. And that's why here I need to encode them.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Yes, I will add int128_t

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: thanks

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Missed this when refactoring the code. This is not needed.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: thanks.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks, I noticed this too, when I commit the code it is formatted aotmatically. You can also check the original line.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: thanks.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: thanks,

velox/dwio/common/Statistics.h
PingLiuPing: It acts as a default constructor.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Yes, this is different with murmur3 hash algorithm where each byte matters.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: The else cannot be removed because in else branch there is function call std::isnan().

velox/dwio/parquet/writer/Writer.cpp
jinchengchenghh: Only if in this case, we use the shared_ptr,
The change from `const std::shared_ptr<arrow::FileMetaData>& fileMetadata` to `const arrow::FileMetaData& fileMetadata` is typically made for **performance, code clarity, and flexibility**. Here's why:

### 1. **Avoid Unnecessary Indirection**
   - `std::shared_ptr` is a heap-allocated pointer wrapper. Passing it by reference (`&`) still involves pointer indirection (accessing the object requires dereferencing **twice**: first the shared_ptr reference, then the internal pointer).
   - **Direct reference access** (`const arrow::FileMetaData&`) is more efficient:  
     ✔️ Single dereference  
     ✔️ No atomic ref-count checks  
     ✔️ Better cache locality

### 2. **Eliminate Ownership Semantics Overhead**
   - `shared_ptr` implies **shared ownership**. If the function doesn’t need to extend the object’s lifetime or store the pointer, ownership semantics are redundant.
   - A raw reference (`&`) clearly signals:  
     ✔️ "This function only needs temporary access"  
     ✔️ "Object lifetime is managed elsewhere"

### 3. **Flexibility for Callers**
   - With `shared_ptr&`, callers **must** have a `shared_ptr` (even if they use `unique_ptr` or raw pointers).
   - With a direct `const &`, callers can pass:  
     ✔️ `shared_ptr.get()`  
     ✔️ `unique_ptr.get()`  
     ✔️ Stack objects (if applicable)  
     ✔️ Any existing reference  
     → **No ownership conversion needed**.

### 4. **Simpler Code**
   - **Syntax**: `fileMetadata->DoSomething()` becomes `fileMetadata.DoSomething()`.
   - **Intent**: Clearly communicates "I only need observation, not ownership."

---

### When to Keep `shared_ptr&`
Use `shared_ptr&` **only** if:
1. The function **stores a copy** of the `shared_ptr` (to extend lifetime).
2. Nullability is required (though `std::optional<const FileMetaData&>` is better post-C++17).

---

### Example Refactor
**Before** (with `shared_ptr`):
```cpp
void ProcessMetadata(const std::shared_ptr<arrow::FileMetaData>& fileMetadata) {
    std::cout << fileMetadata->num_rows() << std::endl;  // Indirection + ref-count
}
```

**After** (direct reference):
```cpp
void ProcessMetadata(const arrow::FileMetaData& fileMetadata) {
    std::cout << fileMetadata.num_rows() << std::endl;  // Direct access
}
```

**Caller Flexibility**:
```cpp
// Caller with shared_ptr
std::shared_ptr<arrow::FileMetaData> metadata = ...;
ProcessMetadata(*metadata);  // Dereference explicitly

// Caller with raw pointer
arrow::FileMetaData* raw_metadata = ...;
ProcessMetadata(*raw_metadata);  // Safe if non-null
```

### Key Takeaway
This change optimizes parameter passing when ownership isn’t needed. It’s a common best practice in C++ to prefer **observing references** (`const T&`) over smart pointers for parameter access unless explicit ownership transfer is required.

velox/dwio/parquet/writer/Writer.cpp
jinchengchenghh: But you only assign field id to top level columns, and the id is not same with Iceberg id, curious how the caller convert the id to Iceberg field id.

velox/dwio/parquet/writer/arrow/Statistics.cpp
jinchengchenghh: Use this format
if constexpr (!std::is_floating_point_v<T>) {
      return 0;
    } 
return xxx

PingLiuPing:
```
  int64_t CountNaN(const T* values, int64_t length) {
    if constexpr (!std::is_floating_point_v<T>) {
      return 0;
    } else {
      int64_t count = 0;
      for (auto i = 0; i < length; i++) {
        const auto val = SafeLoad(values + i);
        if (std::isnan(val)) {
          count++;
        }
      }
      return count;
    }
  }
```
Do you mean move this piece to a lambda
```
      int64_t count = 0;
      for (auto i = 0; i < length; i++) {
        const auto val = SafeLoad(values + i);
        if (std::isnan(val)) {
          count++;
        }
      }
```
And then calling this lambda in return statement?

If this is the case, I prefer the current implementation unchanged.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Nested type is out of scope now.
Here there are two assumptions:
1. The columns in the inserted row are continuous.
2. The table's schema source column ID start from 1.

I think 2 is true. Let me check if 1 is always true, if not I need to do lots of work to get the source column ID here.

PingLiuPing: There is problem during schema evolution.

PingLiuPing: I commit some code to fix the schema evolution issue for stats collection.
Some of the code change lies in HiveDataSink and IcebergDataSink. And also add a new field `columnId_` in HiveColumnHandle. 
when writing data to parquet file, I assume that the FileMetadata ColumnChunk array in RowGroup has same order with table schema. If not, more code changes required to get the correct columnChunk index by some mechanism.

velox/dwio/parquet/writer/arrow/Statistics.cpp
jinchengchenghh:
```
int64_t CountNaN(const T* values, int64_t length) {
    if constexpr (!std::is_floating_point_v<T>) {
      return 0;
    } 
    int64_t count = 0;
    for (auto i = 0; i < length; i++) {
      const auto val = SafeLoad(values + i);
      if (std::isnan(val)) {
        count++;
      }
    }
    return count;
  }
```

velox/dwio/parquet/writer/Writer.cpp
jinchengchenghh: I would suggest use the name for the column, for struct data type, we could use struct.a as key, and the upper layer could find the fieldId by name.
Field could be any unique value, such as the iceberg unit test
```
  public void testNestedStructProjection() throws Exception {
    Schema writeSchema =
        new Schema(
            Types.NestedField.required(0, "id", Types.LongType.get()),
            Types.NestedField.optional(
                3,
                "location",
                Types.StructType.of(
                    Types.NestedField.required(1, "lat", Types.FloatType.get()),
                    Types.NestedField.required(2, "long", Types.FloatType.get()))));
```

velox/dwio/common/IcebergStatistics.h
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <folly/json/dynamic.h>
+
+namespace facebook::velox::dwio::common {
+
+// Iceberg data_file struct fields
```
jinchengchenghh: End with .

velox/dwio/parquet/common/UnicodeUtil.h
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/common/base/Exceptions.h"
+#include "velox/expression/VectorFunction.h"
```
jinchengchenghh: Please check which header is in this file or in the cpp file

velox/dwio/parquet/common/UnicodeUtil.h
```diff
@@ -0,0 +1,47 @@
...
+#include "velox/external/utf8proc/utf8proc.h"
+#include "velox/functions/lib/string/StringImpl.h"
+
+namespace facebook::velox::parquet {
+
+class UnicodeUtil {
+ private:
```
jinchengchenghh: Move private after public

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +509,61 @@ void Writer::abort() {
   arrowContext_.reset();
 }
 
+void Writer::collectIcebergDataFileStats(
```
jinchengchenghh, line 546: Add an argument to decide if collect iceberg stats for each row. Do not affect the hive write path too much

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +766,78 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      int128_t minValue = min_;
+      uint8_t buffer[sizeof(int128_t)];
+      // Convert to big-endian.
+      *reinterpret_cast<int128_t*>(buffer) = DecimalUtil::bigEndian(minValue);
+      return std::string(
+          reinterpret_cast<const char*>(buffer), sizeof(int128_t));
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      std::string minStr(reinterpret_cast<const char*>(min_.ptr), min_.len);
+      const auto truncatedMin =
```
jinchengchenghh: Looks like the truncateStringMin argument always hasValue, why use std::optional? And the input should be min_.ptr and min_.len, don't construct the string

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +766,78 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     return s;
   }
 
+  std::string MinValue() const override {
+    if constexpr (std::is_same_v<T, int64_t>) {
```
jinchengchenghh, line 783: Extract the get min and max logic for decimal to anonymous namespace, only the argument min_ or max_ is different.

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +766,78 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      uint8_t buffer[sizeof(int128_t)];
+      // Convert to big-endian.
+      *reinterpret_cast<int128_t*>(buffer) = DecimalUtil::bigEndian(minValue);
+      return std::string(
+          reinterpret_cast<const char*>(buffer), sizeof(int128_t));
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      std::string minStr(reinterpret_cast<const char*>(min_.ptr), min_.len);
+      const auto truncatedMin =
+          UnicodeUtil::truncateStringMin(minStr, 16).value();
```
jinchengchenghh: Notice, this is the default behavior, but it can be change by table property
```
  public static final String DEFAULT_WRITE_METRICS_MODE = "write.metadata.metrics.default";
  public static final String DEFAULT_WRITE_METRICS_MODE_DEFAULT = "truncate(16)";
```

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -769,11 +868,28 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     return num_values_;
   }
 
+  int64_t nan_count() const override {
+    return statistics_.nan_count;
+  }
+
+  bool CompareMax(const Statistics& other) const override {
+    auto typedStats = dynamic_cast<const TypedStatisticsImpl<DType>*>(&other);
+    // Compare max values.
```
jinchengchenghh: Remove the meaningless comment

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -769,11 +868,28 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+    return comparator_->Compare(max_, typedStats->max_) ? false : true;
+  }
+
+  bool CompareMin(const Statistics& other) const override {
+    auto typedStats = dynamic_cast<const TypedStatisticsImpl<DType>*>(&other);
+    // Compare min values.
```
jinchengchenghh: ditto

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +509,61 @@ void Writer::abort() {
...
+  VELOX_CHECK_NOT_NULL(fileMetadata);
+  icebergDataFileStats_->numRecords = fileMetadata->num_rows();
+  auto numColumns = schema_->size();
+  std::vector<std::shared_ptr<arrow::Statistics>> globalMinStats;
+  std::vector<std::shared_ptr<arrow::Statistics>> globalMaxStats;
+  globalMinStats.reserve(numColumns);
+  globalMaxStats.reserve(numColumns);
+  std::vector<bool> hasMinMax(numColumns, false);
+
+  const auto numRowGroups = fileMetadata->num_row_groups();
```
jinchengchenghh, line 586: If the schema is 
```
struct<a, struct<b, c>>
```
How do you collect the stats of b?

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks, as I mentioned earlier there is `std::isnan` in the else branch. Remove it causing compiler error.

velox/dwio/common/IcebergStatistics.h
PingLiuPing: thanks.

velox/dwio/parquet/common/UnicodeUtil.h
PingLiuPing: Thanks. Deleted 
```
#include "velox/expression/VectorFunction.h"
```
And moved 
```
#include "velox/external/utf8proc/utf8proc.h"
#include "velox/functions/lib/string/StringImpl.h"
```
to cpp file.

velox/dwio/parquet/common/UnicodeUtil.h
PingLiuPing: ok.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks. Added a option `collectIcebergStats` in dwio::common::WriterOptions
And only when iceberg writes, this option will be set to true.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Ok

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks, this is where I get the value 16.
So far Presto doesn't support this property yet. Does Spark support it now?

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks, refactored UnicodeUtil.h.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: The source column IDs are:
a s b c
1 2 3 4
We need to collect these IDs: 1, 3, 4

These IDs will be passed to Writer, and from this function, the ID will be used as the key to store the stats.
And those map can be used directly for org.apache.iceberg.Metrics .

PingLiuPing:
@jinchengchenghh I noticed that Presto does not collect lower and upper bounds for MAP and LIST fields. However, it does collect column_sizes, value_counts, null_value_counts, and nan_value_counts for these types.
On the other hand, Presto does collect lower/upper bounds and other statistics for STRUCT fields.
Could you check how Spark behaves in this regard?

velox/dwio/parquet/writer/arrow/Statistics.cpp
jinchengchenghh: yes

jinchengchenghh:
This is a standard Iceberg Schema
```
  private static final Schema SCHEMA =
          // ROW(INT(), ROW(BOOLEAN(), BOOLEAN)))
      new Schema(
          required(1, "id", Types.IntegerType.get()),
          optional(2, "data", Types.StringType.get()),
          optional(
              3,
              "preferences",
              Types.StructType.of(
                  required(8, "feature1", Types.BooleanType.get()),
                  optional(9, "feature2", Types.BooleanType.get())),
              "struct of named boolean options"),
          required(
              4,
              "locations",
              Types.MapType.ofRequired(
                  10,
                  11,
                  Types.StructType.of(
                      required(20, "address", Types.StringType.get()),
                      required(21, "city", Types.StringType.get()),
                      required(22, "state", Types.StringType.get()),
                      required(23, "zip", Types.IntegerType.get())),
                  Types.StructType.of(
                      required(12, "lat", Types.FloatType.get()),
                      required(13, "long", Types.FloatType.get()))),
              "map of address to coordinate"),
          optional(
              5,
              "points",
              Types.ListType.ofOptional(
                  14,
                  Types.StructType.of(
                      required(15, "x", Types.LongType.get()),
                      required(16, "y", Types.LongType.get()))),
              "2-D cartesian points"),
          required(6, "doubles", Types.ListType.ofRequired(17, Types.DoubleType.get())),
          optional(
              7,
              "properties",
              Types.MapType.ofOptional(18, 19, Types.StringType.get(), Types.StringType.get()),
              "string map of properties"));

```
There is check
```
if (field != null && stats.hasNonNullValue() && shouldStoreBounds(column, fileSchema)) {
              Literal<?> min =
                  ParquetConversions.fromParquetPrimitive(
                      field.type(), column.getPrimitiveType(), stats.genericGetMin());
              updateMin(lowerBounds, fieldId, field.type(), min, metricsMode);
              Literal<?> max =
                  ParquetConversions.fromParquetPrimitive(
                      field.type(), column.getPrimitiveType(), stats.genericGetMax());
              updateMax(upperBounds, fieldId, field.type(), max, metricsMode);
            }

  // we allow struct nesting, but not maps or arrays
  private static boolean shouldStoreBounds(ColumnChunkMetaData column, Schema schema) {
    if (column.getPrimitiveType().getPrimitiveTypeName() == PrimitiveType.PrimitiveTypeName.INT96) {
      // stats for INT96 are not reliable
      return false;
    }

    ColumnPath columnPath = column.getPath();
    Iterator<String> pathIterator = columnPath.iterator();
    Type currentType = schema.asStruct();

    while (pathIterator.hasNext()) {
      if (currentType == null || !currentType.isStructType()) {
        return false;
      }
      String fieldName = pathIterator.next();
      currentType = currentType.asStructType().fieldType(fieldName);
    }

    return currentType != null && currentType.isPrimitiveType();
  }
```
Notice,   we allow struct nesting, but not maps or arrays, and if the column is all nulls, we also don't collect the stats.

PingLiuPing:
Thanks, I think nested struct is not an issue here. With the unique IDs we can collect stats for those fields.

Thanks, `shouldStoreBounds` confirm that we don't need to collect upper and lower bound for MAP and LIST fields. And this is not a special case for Presto. I need to adjust the code little bit for this.

> Notice, we allow struct nesting, but not maps or arrays, and if the column is all nulls, we also don't collect the stats.

To be more accurate, it doesn't collect upper and lower bounds when all values are NULL. It still collect other stats such as column_sizes , value_counts, null_value_counts. The current implementation has handled this case.

PingLiuPing:
After some investigation I don't find a straightforward way to disable collect lower and upper bounds for MAP and LIST. 
And since parquet has collected them already, it is just a matter of where we discard those information. Ideally I want to discard them in `parquet::Writer::collectDataFileStats`, but it requires the column metadata which requires too much code change.
I implement a workaround in IcebergDataSink, where I can get the column metadata i.e the column type. And then filtered the lower and upper bounds as needed. This happens just before sending back the commit metadata.
Let me know your thought on this @jinchengchenghh .

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -547,6 +550,18 @@ TypedComparatorImpl<false, ByteArrayType>::GetMinMax(
   return GetMinMaxBinaryHelper<false>(*this, values);
 }
 
+template <typename DecimalType>
```
jinchengchenghh: typename DecimalType -> T

velox/dwio/parquet/writer/Writer.h
```diff
@@ -196,6 +197,10 @@ class Writer : public dwio::common::Writer {
   // Sets the memory reclaimers for all the memory pools used by this writer.
   void setMemoryReclaimers();
 
+  // Collect Iceberg data file statistics from FileMetaData.
+  void collectDataFileStats(
```
jinchengchenghh: collectDataFileStats -> collectIcebergDataFileStats

velox/dwio/common/Options.h
```diff
@@ -705,6 +705,8 @@ struct WriterOptions {
 
   std::string sessionTimezoneName;
   bool adjustTimestampToTimezone{false};
+  bool collectIcebergStats{false};
```
jinchengchenghh: Add comment for the options

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +511,66 @@ void Writer::abort() {
...
+  for (auto i = 0; i < numRowGroups; ++i) {
+    auto rgm = fileMetadata->RowGroup(i);
+    // numColumns should match the actual number of columns in the Parquet file.
```
jinchengchenghh: remove this comment

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +511,66 @@ void Writer::abort() {
...
+    VELOX_CHECK_EQ(numColumns, rgm->num_columns());
+    for (auto j = 0; j < numColumns; ++j) {
```
jinchengchenghh, line 592: You could scan the input schema_, and get the skipColumns

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Ok, let's leave this to future enhancement. Need to add a option to velox parquet writer and map this option to Spark's table property.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Ok

velox/dwio/parquet/writer/Writer.h
PingLiuPing: Ok, I will change it back. Ying add some comments on this name and suggest not using `Iceberg` terminology in parquet writter.

velox/dwio/common/Options.h
PingLiuPing: Ok

velox/dwio/parquet/writer/Writer.h
jinchengchenghh: The statistic is tightly related to iceberg, the key is iceberg field id, so we cannot use a generic name for it.

jinchengchenghh:
There is also a option, use std::vector<ColumnSizes> to collect the stats, the index is the column number after flattened, let the upper data sink to convert to required statistic map, map the column index to the iceberg fieldId, the writer will not aware of the Iceberg, let the community to decide use which one, I would prefer this option, but I can also accept current implementation because I'm not sure the truncate length, encode strategy is common for other data lake

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -350,6 +353,9 @@ Writer::Writer(
       getArrowParquetWriterOptions(options, flushPolicy_);
   setMemoryReclaimers();
   writeInt96AsTimestamp_ = options.writeInt96AsTimestamp;
+  dataFileStats_ = std::make_shared<dwio::common::IcebergDataFileStatistics>();
```
jinchengchenghh, line 384: Move to initialize list and add const

PingLiuPing: I added dataFileStats_ in the base class. And there is no explicit constructor in base class.
Do you want me to add dataFileStats_ in parquet::Writer ? Doing this, I need to cast the writer to parquet::writer in connector layer to get the stats.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +511,68 @@ void Writer::abort() {
...
+    const std::shared_ptr<arrow::FileMetaData>& fileMetadata) {
+  VELOX_CHECK_NOT_NULL(fileMetadata);
+  dataFileStats_->numRecords = fileMetadata->num_rows();
+  // numColumns is not the number of columns in Iceberg table's schema,
+  // e.g., schema_->size(). It also contains the sub-fields when there are
+  // nested types in table's schema.
+  auto numColumns = sourceColumnIndices_->size();
+  std::vector<std::shared_ptr<arrow::Statistics>> globalMinStats;
+  std::vector<std::shared_ptr<arrow::Statistics>> globalMaxStats;
+  globalMinStats.reserve(numColumns);
```
p1rattttt: i think that you should use resize(numColumns) or just another constructor
std::vector<std::shared_ptr<arrow::Statistics>> globalMinStats(numColumns);
std::vector<std::shared_ptr<arrow::Statistics>> globalMaxStats(numColumns);
because later you do globalMinStats[j] = columnChunkStats; without increasing size

PingLiuPing: @p1rattttt  Thanks, this will lead to undefined behavior.
Will push the fix later.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
 std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
```
jinchengchenghh: Rename the function to updateFieldRecursive

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+  if (!name.empty()) {
+    newField = field->WithName(name);
+  }
   if (type.isRow()) {
-    auto& rowType = type.asRow();
-    auto newField = field->WithName(name);
+    const auto& rowType = type.asRow();
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
-    auto childrenSize = rowType.size();
```
jinchengchenghh, line 205: ditto

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+          structType->fields()[i],
+          *rowType.childAt(i),
+          rowType.nameOf(i),
+          childSetting));
     }
-    return newField->WithType(::arrow::struct_(newFields));
+    newField = newField->WithType(::arrow::struct_(newFields));
   } else if (type.isArray()) {
-    auto newField = field->WithName(name);
+    const auto& arrayType = type.asArray();
```
jinchengchenghh: ditto

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
-    auto elementType = type.asArray().elementType();
-    auto elementField = listType->value_field();
-    return newField->WithType(
-        ::arrow::list(updateFieldNameRecursive(elementField, *elementType)));
+    const auto* childSetting =
+        icebergFieldSetting ? &icebergFieldSetting->children.at(0) : nullptr;
+    auto updatedElementField = updateFieldNameRecursive(
+        listType->value_field(),
+        *arrayType.elementType(),
+        arrayType.nameOf(0),
```
jinchengchenghh: Use the function argument default value "" directly.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: thanks. Fixed.

jinchengchenghh:
The statistic returned by map with string, this is a copy, and return to json string, then parse json string, this is another cost, so maybe you need to measure the cost and test the performance. You may need to write the manifest natively and return the manifest path in the future as optimization.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: The issue resolved by adding a new struct IcebegNestedField. And it contains the field ID and the nested field IDs.

PR #14156, opened 2025-07-20
fix: When export to Arrow, NullType should have 0 buffer

WangGuangxin:
The NullType in Arrow should be 0 buffer. Otherwise, a check in Java side will throw exception `Could not load buffers for field xx: Null. error message: Expected 0 buffers for type Null, but found 2`.

The check can be found in arrow 15.0
https://github.com/apache/arrow/blob/a61f4af724cd06c3a9b4abd20491345997e532c0/java/c/src/main/java/org/apache/arrow/c/BufferImportTypeVisitor.java#L134


WangGuangxin:
@PHILO-HE @rui-mo Can you please take a look at this?

velox/vector/arrow/tests/ArrowBridgeArrayTest.cpp
```diff
@@ -113,6 +113,7 @@ class ArrowBridgeArrayExportTest : public testing::Test {
       const ArrowArray& arrowArray) {
     const bool isString =
         std::is_same_v<T, StringView> or std::is_same_v<T, std::string>;
+    const bool isNullType = std::is_same_v<T, UnknownValue>;
```
rui-mo: nit: In Spark, the mapped type is NullType, whereas in Presto, it is UnknownType. For consistency, Velox can still refer to it as unknown type.

https://github.com/prestodb/presto/blob/master/presto-common/src/main/java/com/facebook/presto/common/type/UnknownType.java

velox/vector/arrow/tests/ArrowBridgeArrayTest.cpp
```diff
@@ -231,6 +238,7 @@ class ArrowBridgeArrayExportTest : public testing::Test {
...
```
rui-mo: ditto

velox/vector/arrow/tests/ArrowBridgeArrayTest.cpp
```diff
@@ -1007,6 +1017,13 @@ TEST_F(ArrowBridgeArrayExportTest, dictionaryNested) {
   EXPECT_EQ(values.Value(2), 3);
 }
 
+TEST_F(ArrowBridgeArrayExportTest, nullType) {
```
rui-mo: ditto

rui-mo (commented):
Thanks.

velox/vector/arrow/Bridge.cpp
```diff
@@ -1112,7 +1114,7 @@ void exportConstantValue(
         ? sizeof(StringView)
         : vec.type()->cppSizeInBytes();
 
-    valuesVector = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
+    valuesVector = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH_ALL(
```
rui-mo: This change extends the support for UNKNOWN and OPAQUE types. Should `OPAQUE` type be excluded?

PHILO-HE (commented):
Looks good. Thanks.

zhli1142015:
Hello @WangGuangxin 
I observed the similar issue in our end, will you continue this PR to make sure it landing?

WangGuangxin:
> Hello @WangGuangxin I observed the similar issue in our end, will you continue this PR to make sure it landing?

ok, will update today

WangGuangxin:
@rui-mo @zhli1142015 Could you please help review this again?

velox/vector/arrow/Bridge.cpp
```diff
@@ -1112,19 +1114,24 @@ void exportConstantValue(
     // If this is a scalar type, then ConstantVector does not have a vector
     // inside. Wrap the single value in a flat vector with a single element to
     // export it to an ArrowArray.
-    size_t bufferSize = (vec.type()->isVarchar() || vec.type()->isVarbinary())
-        ? sizeof(StringView)
-        : vec.type()->cppSizeInBytes();
+    if (vec.typeKind() == TypeKind::UNKNOWN) {
+      valuesVector = BaseVector::create(vec.type(), 1, pool);
+      valuesVector->setNull(0, true);
```
rui-mo: Is `setNull` needed, because `BaseVector::createInternal` would allocate the null buffer as all-null?

```
case TypeKind::UNKNOWN: {
      BufferPtr nulls = allocateNulls(size, pool, bits::kNull);
      return std::make_shared<FlatVector<UnknownValue>>(
          pool,
          UNKNOWN(),
          std::move(nulls),
          size,
          nullptr,
          std::vector<BufferPtr>());
    }
```

WangGuangxin: yes, fixed

rui-mo:
@pedroerp Would you like to take a look? Thanks!

mbasmanova:
@WangGuangxin Would you rebase?

PR #14159, opened 2025-07-20
feat(iceberg): support sorting during insertion

PingLiuPing:
Support sorting during insertion
The commit start from bd5694ba0ab27ef103ac900bf2d8b38f3ba3b17c


velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/connectors/hive/TableHandle.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+struct IcebergNestedField {
```
yingsu00, line 22: nit: ColumnIdentity

PingLiuPing: Thanks, I think using IcebergNestedField is reasonable here because I want to represent iceberg `field`. And the field can be nested field. This is align to the iceberg spec concept. And also align to the iceberg java library implementation.

PR #14160, opened 2025-07-21
(feat): Remove cpr dependency

majetideepak:
cpr requires a specific libcurl version and is not portable. The usage does not require a new dependency and can
be replaced by libcurl which is present on most distros.


CMake/resolve_dependency_modules/curl.cmake
```diff
@@ -28,3 +28,9 @@ FetchContent_Declare(
   curl
   URL ${VELOX_CURL_SOURCE_URL}
   URL_HASH ${VELOX_CURL_BUILD_SHA256_CHECKSUM})
+
+set(BUILD_SHARED_LIBS ${VELOX_BUILD_SHARED})
```
assignUser: Should we just always statically link that? Or do we expose curl stuff through headers? (I hope not ...)

CMakeLists.txt
```diff
@@ -192,8 +192,8 @@ if(VELOX_ENABLE_BENCHMARKS_BASIC)
 endif()
 
 if(VELOX_BUILD_TESTING OR VELOX_BUILD_TEST_UTILS)
-  set(cpr_SOURCE BUNDLED)
-  velox_resolve_dependency(cpr)
+  set(curl_SOURCE BUNDLED)
+  velox_resolve_dependency(curl)
```
assignUser: Do we not also need it with faiss and no testing?

CMake/resolve_dependency_modules/gtest.cmake
```diff
@@ -31,5 +31,12 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(googletest)
 
+if(NOT TARGET GTest::gtest)
+  message(FATAL_ERROR "gtest not found")
+endif()
+if(NOT TARGET GTest::gmock)
+  message(FATAL_ERROR "gmock not found")
+endif()
+
```
assignUser: nit: remove after CI is green

CMake/resolve_dependency_modules/cudf.cmake
```diff
@@ -17,48 +17,48 @@ include_guard(GLOBAL)
 # 3.30.4 is the minimum version required by cudf
 cmake_minimum_required(VERSION 3.30.4)
 
-set(VELOX_rapids_cmake_VERSION 25.08)
+set(VELOX_rapids_cmake_VERSION 25.10)
```
czentgr, line 20: These changes are temporary to ensure it works fine?

CMakeLists.txt
```diff
@@ -212,9 +212,16 @@ if(VELOX_ENABLE_BENCHMARKS_BASIC)
   set(VELOX_BUILD_TEST_UTILS ON)
 endif()
 
+if(VELOX_ENABLE_CUDF)
+  message(STATUS "Building curl from source to satisfy cuDF curl version requirement")
+  set(curl_SOURCE BUNDLED)
+  velox_resolve_dependency(curl)
+  FetchContent_MakeAvailable(curl)
```
czentgr: Did we want to move this into the curl.cmake?

CMake/resolve_dependency_modules/cudf.cmake
majetideepak: Yes. I cherry-picked https://github.com/facebookincubator/velox/pull/15183 since it has the curl bundling required for cuDF. We don't have to bundle curl otherwise.

CMakeLists.txt
majetideepak: I will move it after rebase.

.github/workflows/linux-build-base.yml
```diff
@@ -206,7 +206,7 @@ jobs:
 
       - name: Run Tests
         run: |
-          cd _build/debug && ctest -j 8 --output-on-failure --no-tests=error
+          cd _build/debug && ctest -j 8 --output-on-failure --no-tests=error -E "_deps/.*"
```
karthikeyann: Is this added to avoid running tests inside _deps?
Are some of the dependencies building tests?

majetideepak: Yes, gflags starts building tests when we remove the global `unset(BUILD_TESTING)` from cpr.cmake
https://github.com/facebookincubator/velox/actions/runs/18730541644/job/53426189667?pr=14160

majetideepak: Some other package is likely caching the `BUILD_TESTING` variable

assignUser: Should we set build_testing after the deps resolution? or explictly set it to false in resolve modules within `block()`?

majetideepak: I am logging after each package to find the culprit.

CMake/resolve_dependency_modules/curl.cmake
majetideepak: I am worried if the system curl (dynamic) would be picked if we only have a static library since usually dynamic has a precedence over static.

majetideepak:
I moved this to draft to fix the build issues we hit by removing CPR.

CMakeLists.txt
majetideepak: We don't need curl for faiss. I thought system curl can be used if there is no testing and CUDF is disabled. But I see a point that this could lead to different behaviors.

assignUser:
Kind of dumb that we have to resort to this but lets use [--test-dir](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-test-dir) in our ctest invocation 
```diff
--- a/.github/workflows/linux-build-base.yml
+++ b/.github/workflows/linux-build-base.yml
@@ -137,7 +137,7 @@ jobs:
           wget https://repo1.maven.org/maven2/org/mockito/mockito-core/2.23.4/mockito-core-2.23.4.jar -O /usr/local/hadoop/share/hadoop/mapreduce/mockito-core-2.23.4.jar
 
           export CLASSPATH=`/usr/local/hadoop/bin/hdfs classpath --glob`
-          ctest -j 8 --label-exclude cuda_driver --output-on-failure --no-tests=error
+          ctest -j 8 --label-exclude cuda_driver --output-on-failure --no-tests=error --test-dir velox
 
   ubuntu-debug:
     runs-on: 8-core-ubuntu-22.04
diff --git a/Makefile b/Makefile
index b9548ea66..57670d83d 100644
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ benchmarks-basic-run:
 			${EXTRA_BENCHMARK_FLAGS}
 
 unittest: debug			#: Build with debugging and run unit tests
-	cd $(BUILD_BASE_DIR)/debug && ctest -j ${NUM_THREADS} -VV --output-on-failure
+	cd $(BUILD_BASE_DIR)/debug && ctest -j ${NUM_THREADS} -VV --output-on-failure --test-dir velox
 
 # Build with debugging and run expression fuzzer test. Use a fixed seed to
 # ensure the tests are reproducible.
diff --git a/scripts/docker/docker-command.sh b/scripts/docker/docker-command.sh
index 88faf22c9..475b7a184 100755
--- a/scripts/docker/docker-command.sh
+++ b/scripts/docker/docker-command.sh
@@ -16,4 +16,4 @@
 set -eu
 # Compilation and testing
 make
-cd _build/release && ctest -j"${NUM_THREADS}" --output-on-failure --no-tests=error
+cd _build/release && ctest -j"${NUM_THREADS}" --output-on-failure --no-tests=error --test-dir velox

```

majetideepak:
It was curl and duckdb that were the culprits so far.
I want to fix this property and avoid building tests and shared libs unnecessarily in dependencies.

majetideepak:
@assignUser, @czentgr, @kgpai can you take another look?

velox/exec/fuzzer/CMakeLists.txt
```diff
@@ -43,7 +43,8 @@ target_link_libraries(
   velox_exec_test_lib
   velox_expression_functions
   velox_presto_types
-  cpr::cpr
+  CURL::libcurl
+  Boost::regex
```
czentgr: This is used because it came in from CPR previously?

velox/exec/fuzzer/PrestoQueryRunner.cpp
```diff
@@ -493,38 +501,75 @@ std::vector<RowVectorPtr> PrestoQueryRunner::execute(
...
+
+  std::string url =
+      fmt::format("{}/v1/statement?binaryResults=true", coordinatorUri_);
+  curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
+  curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
+  curl_easy_setopt(curl, CURLOPT_POST, 1L);
+  curl_easy_setopt(curl, CURLOPT_POSTFIELDS, sql.c_str());
+  curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, sql.size());
+  curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, timeout_);
+  curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writeFunction);
```
czentgr, line 525: As a pre-caution I suggest we add
```
curl_easy_setopt(curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);
```
this is to head of code scanners. It won't do anything unless we use https.
Same for the other fetchNext.

velox/exec/fuzzer/PrestoQueryRunner.cpp
```diff
@@ -493,38 +501,75 @@ std::vector<RowVectorPtr> PrestoQueryRunner::execute(
...
+
+  std::string response;
+  curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response);
+
+  // Perform the request
+  CURLcode res = curl_easy_perform(curl);
+  VELOX_CHECK(
```
czentgr: `VELOX_CHECK_EQ(CURLE_OK, res, ...)`?

velox/exec/fuzzer/PrestoQueryRunner.cpp
```diff
@@ -493,38 +501,75 @@ std::vector<RowVectorPtr> PrestoQueryRunner::execute(
...
+  curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, timeout_);
+
+  // Capture the response body
+  curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writeFunction);
+  std::string response;
+  curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response);
+
+  // Perform GET request
+  CURLcode res = curl_easy_perform(curl);
+  VELOX_CHECK(
```
czentgr: VELOX_CHECK_EQ?

velox/exec/fuzzer/CMakeLists.txt
assignUser: I was wondering the same, a quick grep didn't show anything but I see re2 is also used.

assignUser (approved):
CMake lgtm

velox/exec/fuzzer/CMakeLists.txt
majetideepak: Rebase artifact. This is fixed.

PR #14177, opened 2025-07-22
refactor: Avoid dynamic allocation and atomic counter for scalar and unknown types

MBkkt:
Not sure about any performance improvements but pretty sure that current code do unnecessary stuff without simplicity or readability reasons

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -809,7 +809,7 @@ void OpaqueType::registerSerializationTypeErased(
 }
 
 ArrayTypePtr ARRAY(TypePtr elementType) {
-  return std::make_shared<ArrayType>(std::move(elementType));
+  return TypeFactory<TypeKind::ARRAY>::create(std::move(elementType));
 }
 
 RowTypePtr ROW(std::vector<std::string> names, std::vector<TypePtr> types) {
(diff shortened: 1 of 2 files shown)


velox/type/Type.cpp
```diff
@@ -835,13 +835,14 @@ RowTypePtr ROW(std::vector<TypePtr>&& types) {
...
-  return std::make_shared<MapType>(std::move(keyType), std::move(valType));
+  return TypeFactory<TypeKind::MAP>::create(
+      std::move(keyType), std::move(valType));
 }
 
 std::shared_ptr<const FunctionType> FUNCTION(
     std::vector<TypePtr>&& argumentTypes,
     TypePtr returnType) {
-  return std::make_shared<FunctionType>(
+  return std::make_shared<const FunctionType>(
```
mbasmanova, line 845: Is 'const' needed here?

mbasmanova (approved):
Thanks.

velox/type/Type.cpp
MBkkt: Otherwise type in signature and returned type is different and shared_ptr ctor will be called instead of nrvo

MBkkt:
@mbasmanova my code (constexpr part) doesn't compile for C++17 but it should for C++20 https://godbolt.org/z/1vx7zfoPj

I'm not sure, does velox support C++17? I saw this:
https://github.com/facebookincubator/velox/pull/10866

Looks like in CI it's still C++17 (from ubuntu and linux jobs)
```
-- CXX standard: 17
```

zhouyuan:
@kKPulla this patch is merged with red CI: 
https://github.com/facebookincubator/velox/pull/14177/checks

kKPulla:
> @kKPulla this patch is merged with red CI: https://github.com/facebookincubator/velox/pull/14177/checks

My bad. Somehow this signal did not show up when I am merging it in. Either way, does this PR fix this? cc: @zhztheplayer  https://github.com/facebookincubator/velox/pull/14310

PR #14178, opened 2025-07-22
feat: Support pushdown spec through ss table handle and the fallback eval processing

xiaoxmeng:
Summary:
Extends SequenceStoragePushdownSpec to receive the expression pushdown from the query optimizer. Adds struct as map to handle the conversion from map to row.
Extends sequence storage connector to handle the feature projection fallback from the sequence storage client in case the latter does not use thrift map encoding.

The follow up is to use koski udf to handle map to row conversion, and then integrate into koski query optimizer.

Differential Revision: D78467818


Yuhta (approved):
Velox part looks ok.  Let's wait the OSS build being fixed before merge though.

PR #14180, opened 2025-07-22
refactor: getCastOperator to accept TypePtr

HeidiHan0000:
Summary: For a new custom type (BigintEnumType which will be in a following PR), the cast operator requires the Type and its parameters to be able to fail on certain cases of cast. Refactoring getCastOperator and getCustomTypeCastOperator to accept this parameter as input.

Differential Revision: D78694010

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -930,7 +930,7 @@ CastOperatorPtr CastExpr::getCastOperator(const TypePtr& type) {
     return it->second;
   }
 
-  auto castOperator = getCustomTypeCastOperator(key);
+  auto castOperator = getCustomTypeCastOperator(type);
   if (castOperator == nullptr) {
     return nullptr;
   }
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -2068,7 +2068,11 @@ class CustomTypeFactory {
   /// should be treated as its underlying native type during type castings,
   /// return a nullptr. If a custom type does not support castings, throw an
   /// exception.
-  virtual exec::CastOperatorPtr getCastOperator() const = 0;
+  /// The TypePtr input parameter is used when the cast operation depends on the
+  /// Type and its properties. For example, when casting to an enum type, the
+  /// CastOperator needs to check if the input value exists in the enum type's
+  /// value map.
+  virtual exec::CastOperatorPtr getCastOperator(const TypePtr& type) const = 0;
 
   virtual AbstractInputGeneratorPtr getInputGenerator(
       const InputGeneratorConfig& config) const = 0;
@@ -2167,7 +2171,11 @@ bool unregisterCustomType(const std::string& name);
 /// Returns the custom cast operator for the custom type with the specified
 /// name. Returns nullptr if a type with the specified name does not exist or
 /// does not have a dedicated custom cast operator.
-exec::CastOperatorPtr getCustomTypeCastOperator(const std::string& name);
+/// The TypePtr input parameter is used when the cast operation depends on the
+/// Type and its properties. For example, when casting to an enum type, the
+/// CastOperator needs to check if the input value exists in the enum type's
+/// value map.
+exec::CastOperatorPtr getCustomTypeCastOperator(const TypePtr& type);
 
 /// Returns the input generator for the custom type with the specified name.
 AbstractInputGeneratorPtr getCustomTypeInputGenerator(
--- a/velox/expression/signature_parser/tests/ParseTypeSignatureTest.cpp
+++ b/velox/expression/signature_parser/tests/ParseTypeSignatureTest.cpp
@@ -61,7 +61,8 @@ class TypeFactory : public CustomTypeFactory {
     return type_;
   }
 
-  exec::CastOperatorPtr getCastOperator() const override {
+  exec::CastOperatorPtr getCastOperator(
+      const TypePtr& /*type*/) const override {
     return nullptr;
   }
 
--- a/velox/expression/tests/CastExprTest.cpp
+++ b/velox/expression/tests/CastExprTest.cpp
@@ -2874,7 +2874,8 @@ class BigintTypeWithCustomComparisonTypeFactory : public CustomTypeFactory {
   }
 
   // Type casting from and to TimestampWithTimezone is not supported yet.
-  exec::CastOperatorPtr getCastOperator() const override {
+  exec::CastOperatorPtr getCastOperator(
+      const TypePtr& /*type*/) const override {
     return BigintTypeWithCustomComparisonCastOperator::get();
   }
 
--- a/velox/expression/tests/CustomTypeTest.cpp
+++ b/velox/expression/tests/CustomTypeTest.cpp
@@ -61,7 +61,8 @@ class FancyIntTypeFactory : public CustomTypeFactory {
     return FancyIntType::get();
   }
 
-  exec::CastOperatorPtr getCastOperator() const override {
+  exec::CastOperatorPtr getCastOperator(
+      const TypePtr& /*type*/) const override {
     VELOX_UNSUPPORTED();
   }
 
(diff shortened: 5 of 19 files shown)


velox/expression/CastExpr.cpp
```diff
@@ -930,7 +930,7 @@ CastOperatorPtr CastExpr::getCastOperator(const TypePtr& type) {
     return it->second;
   }
 
-  auto castOperator = getCustomTypeCastOperator(key);
+  auto castOperator = getCustomTypeCastOperator(type);
```
mbasmanova, line 933: castOperators_ is a map keyed on type name. Since the same type name may produce different CastOperator instances, it is no longer sufficient use type name as the key.

HeidiHan0000: Thanks for pointing this out! I am considering using toString(), but this would mean for other types like ARRAY, MAP, DECIMAL etc, a different cast operator would be created for different parameters/children, which is not necessary. 

I can add a separate getKey(type) function which will return type->toString() for enum types and type->name() for all other types. But this would need to be part of the PR where I add the BigintEnumType since it does not exist yet. Does this sound okay?

kagamiori: Hi @HeidiHan0000, just curious why you need this change? Do you expect multiple custom types with the same name that should use different cast operators? Do you have a design doc about the BigintEnumType in Velox?
cc: @kgpai

HeidiHan0000: Hi @kagamiori thanks for the comment. Enum casts require a check to see if the value we are trying to cast to the enum actually exists in the enum as a value. 

For instance, consider `exampleType` as a BigintEnum with values {"value1": 1, "value2": 2}. Then `cast(5 as exampleType)` should fail with an error like "No value 5 in exampleType".
 
So it requires the values map of the enum, which is different for each enum type, although they have the same type name "BIGINT_ENUM". I can share with you a design doc offline

velox/type/Type.h
```diff
@@ -2167,7 +2167,7 @@ bool unregisterCustomType(const std::string& name);
 /// Returns the custom cast operator for the custom type with the specified
 /// name. Returns nullptr if a type with the specified name does not exist or
 /// does not have a dedicated custom cast operator.
```
kewang1024, line 2173: Also update the comment as well

velox/type/Type.h
```diff
@@ -2068,7 +2068,7 @@ class CustomTypeFactory {
   /// should be treated as its underlying native type during type castings,
   /// return a nullptr. If a custom type does not support castings, throw an
   /// exception.
-  virtual exec::CastOperatorPtr getCastOperator() const = 0;
+  virtual exec::CastOperatorPtr getCastOperator(const TypePtr& type) const = 0;
```
kewang1024, line 2075: update the comment

velox/expression/CastExpr.cpp
kewang1024: Thanks @mbasmanova for catching this issue

@HeidiHan0000 is it possible for us to return enum(name + mapping) as the name? the API seems to indicate this is a unique logical type name

```
 /// Returns unique logical type name. It can be
  /// different from the physical type name returned by 'kindName()'.
  virtual const char* name() const = 0;
```

Also since we're adding to `castOperators_` with the ephemeral enum types of different parameters, do we have one `CastExpr` per task and it would be destroyed after the task is finished?

mbasmanova: @HeidiHan0000 @kagamiori @kewang1024 

Type::name() returns the logical type name, but doesn't include parameters. It is DECIMAL, not DECIMAL(10, 3) or ROW, not ROW(a INTEGER, b VARCHAR). I expect it will be BIGINT_ENUM. Hence, we cannot use it as a key.

We need a unique identifier for the CastOperator. Currently, we assume that some types may use a single operator for all parameters, while other types may require different operators. The question is whether this is necessary. It might be better to require that a single instance of CastOperator handles all types that share the name. To achieve that we would need to modify CastOperator API to provide the full type information for both CAST source and target.

```
  /// Determines whether the cast operator supports casting to the custom type
  /// from the other type.
  virtual bool isSupportedFromType(const TypePtr& other) const = 0;

  /// Determines whether the cast operator supports casting from the custom type
  /// to the other type.
  virtual bool isSupportedToType(const TypePtr& other) const = 0;
```

We would change these to a single API: bool isSupported(from, to). We would also change castTo/From APIs to a single cast(from, to).

What do you think?

HeidiHan0000: @kewang1024  
>  is it possible for us to return enum(name + mapping) as the name? the API seems to indicate this is a unique logical type name

I'm not sure this would be possible, because for the getType(name, parameters) function, the same type->name() is used as a key for CustomTypeFactories stored in typeFactories(). And for registering CustomTypeFactories, we want them to all be registered under one name "BIGINT_ENUM"

HeidiHan0000: @mbasmanova Thanks for the suggestion, I think if castTo/From can take the TypePtr of the type which the castOperator is for, then that would solve our issue. In this case, the getCastOperator() function no longer needs to take in a TypePtr parameter so I will undo those parts, and then there will be refactoring required for all types with castOperators to use a cast(from, to) instead of the castFrom and castTo methods. Is this correct?

I'm not sure I understand why isSupportedTo/FromType would need to be updated though. Would you be able to clarify this part?

mbasmanova: > I'm not sure I understand why isSupportedTo/FromType would need to be updated though. Would you be able to clarify this part?

I believe these also need full type info for both to and from to decide whether cast is supported, no?

HeidiHan0000: @mbasmanova  Oh my bad, yes it is, thanks. 

Side question, for the isSupported function, how is it supposed to handle the case for casting to the same type? It would be a no-op, but would the isSupported function need to explicitly return true? Is there any other thing required for this case?

mbasmanova: I think we can document in the API that the caller must check if types are the same and not call isSupported. Hence, the implementation can simply assume that to != from.

HeidiHan0000: Got it, thanks for the clarification!

kagamiori: > @HeidiHan0000 @kagamiori @kewang1024
> 
> Type::name() returns the logical type name, but doesn't include parameters. It is DECIMAL, not DECIMAL(10, 3) or ROW, not ROW(a INTEGER, b VARCHAR). I expect it will be BIGINT_ENUM. Hence, we cannot use it as a key.
> 
> We need a unique identifier for the CastOperator. Currently, we assume that some types may use a single operator for all parameters, while other types may require different operators. The question is whether this is necessary. It might be better to require that a single instance of CastOperator handles all types that share the name. To achieve that we would need to modify CastOperator API to provide the full type information for both CAST source and target.
> 
> ```
>   /// Determines whether the cast operator supports casting to the custom type
>   /// from the other type.
>   virtual bool isSupportedFromType(const TypePtr& other) const = 0;
> 
>   /// Determines whether the cast operator supports casting from the custom type
>   /// to the other type.
>   virtual bool isSupportedToType(const TypePtr& other) const = 0;
> ```
> 
> We would change these to a single API: bool isSupported(from, to). We would also change castTo/From APIs to a single cast(from, to).
> 
> What do you think?

Hi @mbasmanova, yes, this makes sense to me. Combining the isSupportedFrom/To into one isSupported(from, to) is useful for general cases.

HeidiHan0000:
Upon further discussion, we found that this change is not needed. Originally, it was meant to serve 2 purposes for BigintEnumTypes:
1. In isSupportedTo and isSupportedFrom functions, to check if the other type is equivalent to the current BigintEnumType. But as mentioned in the above comments, this check is not required and the isSupported* functions can assume that toType != fromType.
2. For castTo and castFrom functions, as it requires knowledge of the values map of the enum type. This is also available from the input arguments to castTo/From, the input and result types can be retrieved with `(BaseVector&)input->type()`, and `TypePtr& resultType`

Closing the PR

PR #14181, opened 2025-07-22
feat: Add kLongEnumLiteral and kVarcharEnumLiteral TypeParameter kinds

HeidiHan0000:
Adding kLongEnumLiteral and kVarcharEnumLiteral TypeParameter kinds, both with a string enumName field and an unordered_map<string, int64_t/string> valuesMap, to be used as inputs to EumTypes (in following PRs)

Differential Revision: D78713455

### DIFF
--- a/velox/expression/SignatureBinder.cpp
+++ b/velox/expression/SignatureBinder.cpp
@@ -277,6 +277,9 @@ bool SignatureBinderBase::tryBind(
           return false;
         }
         break;
+      default:
+        // TODO: Add support for kLongEnumLiteral and kVarcharEnumLiteral.
+        break;
     }
   }
   return true;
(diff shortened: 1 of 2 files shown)


velox/expression/SignatureBinder.cpp
```diff
@@ -267,6 +267,9 @@ bool SignatureBinderBase::tryBind(
           return false;
         }
         break;
+      case TypeParameterKind::kLongEnumLiteral:
+      case TypeParameterKind::kVarcharEnumLiteral:
+        break;
```
kewang1024: can we leave this in the next PR?

kewang1024 (commented):
Thanks @HeidiHan0000, could you also update the title and description?

velox/expression/SignatureBinder.cpp
HeidiHan0000: Done, thanks!

kewang1024 (approved):

PR #14182, opened 2025-07-22
feat(type): Add BigintEnum type in Velox

HeidiHan0000:
Summary:
Adding BigintEnum type as a parametric type in Velox. This type will get registered once with name "BigintEnum", but the BIGINT_ENUM type takes in a std::vector<TypeParameters> which contain the enum name and values map. This way, each different enum type can be instantiated separate from the other enums with a different name and map.
Instances are stored in a static folly EvictingCacheMap with a set max size of 1000 which matches the Java implementation.
New parsing rules will be added in the following diff to recognize the string pattern that Java coordinator serializes the type info to, and to parse that into a string enumName and a std::map<std::string, int64_t> valuesMap.

Differential Revision: D78720602

### DIFF
--- a/velox/docs/develop/types.rst
+++ b/velox/docs/develop/types.rst
@@ -176,6 +176,7 @@ IPPREFIX                  ROW(HUGEINT,TINYINT)
 GEOMETRY                  VARBINARY
 TDIGEST                   VARBINARY
 QDIGEST                   VARBINARY
+BIGINT_ENUM               BIGINT
 ========================  =====================
 
 TIMESTAMP WITH TIME ZONE represents a time point in milliseconds precision
@@ -222,6 +223,16 @@ data for a given input set, and can be queried to retrieve approximate quantile 
 distribution. They may be merged without losing precision, and for storage and retrieval they may
 be cast to/from VARBINARY. The parameter type (BIGINT, REAL, or DOUBLE) represents
 the set of numbers that may be ingested by the quantile digest.
+
+BIGINT_ENUM(LongEnumParameter) type represents an enumerated value where the physical type is BIGINT.
+It takes one LongEnumParameter as parameter, which consists of a string name and a mapping of
+string keys to BIGINT values.
+There is a static cache which stores instances of different BIGINT_ENUM types. This is to treat each
+different enum type as a singleton. The LongEnumParameter is used as the key to retrieve the cached instance,
+and a new instance is only created if it has not been created with the given LongEnumParameter.
+Casting is permitted from any integer type to an enum type. Casting is only permitted from an enum type
+to a BIGINT type. Casting between different enum types is not permitted.
+Comparison operations are only allowed between values of the same enum type.
 
 Spark Types
 ~~~~~~~~~~~~
--- a/velox/functions/prestosql/TypeOf.cpp
+++ b/velox/functions/prestosql/TypeOf.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "velox/expression/VectorFunction.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
 #include "velox/functions/prestosql/types/BingTileType.h"
 #include "velox/functions/prestosql/types/GeometryType.h"
 #include "velox/functions/prestosql/types/HyperLogLogType.h"
@@ -59,7 +60,11 @@ std::string typeName(const TypePtr& type) {
       if (isBingTileType(type)) {
         return "bingtile";
       }
+      if (isBigintEnumType(*type)) {
+        return asBigintEnum(type)->enumName();
+      }
       return "bigint";
+
     case TypeKind::HUGEINT: {
       if (isUuidType(type)) {
         return "uuid";
--- a/velox/functions/prestosql/registration/RegistrationFunctions.cpp
+++ b/velox/functions/prestosql/registration/RegistrationFunctions.cpp
@@ -16,6 +16,7 @@
 #include <string>
 #include "velox/functions/prestosql/IPAddressFunctions.h"
 #include "velox/functions/prestosql/UuidFunctions.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
 
 namespace facebook::velox::functions {
 
@@ -135,6 +136,9 @@ void registerBitwiseFunctions(const std::string& prefix) {
 }
 
 void registerAllScalarFunctions(const std::string& prefix) {
+  // TODO: move registerBigintEnumType to registerEnumTypeFunctions once the
+  // enum functions are implemented
+  registerBigintEnumType();
   registerArithmeticFunctions(prefix);
   registerCheckedArithmeticFunctions(prefix);
   registerComparisonFunctions(prefix);
--- a/velox/functions/prestosql/tests/BigintEnumCastTest.cpp
+++ b/velox/functions/prestosql/tests/BigintEnumCastTest.cpp
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/functions/prestosql/tests/CastBaseTest.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+using namespace facebook::velox::test;
+using namespace facebook::velox::exec;
+
+namespace facebook::velox {
+namespace {
+
+class BigintEnumCastTest : public functions::test::CastBaseTest {
+ protected:
+  BigintEnumCastTest() {
+    registerBigintEnumType();
+    LongEnumParameter moodInfo(
+        "test.enum.mood", {{"CURIOUS", -2}, {"HAPPY", 0}});
+    moodEnum_ = BIGINT_ENUM(moodInfo);
+  }
+
+  BigintEnumTypePtr moodEnum_;
+};
+
+TEST_F(BigintEnumCastTest, castTo) {
+  // Cast different integer types to enum type.
+  testCast<int64_t, int64_t>(
+      BIGINT(), moodEnum_, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  testCast<int8_t, int64_t>(
+      TINYINT(), moodEnum_, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  testCast<int16_t, int64_t>(
+      SMALLINT(), moodEnum_, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  testCast<int32_t, int64_t>(
+      INTEGER(), moodEnum_, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  // Cast enum type to same enum type.
+  testCast<int64_t, int64_t>(
+      moodEnum_, moodEnum_, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+}
+
+TEST_F(BigintEnumCastTest, invalidCastTo) {
+  // Cast is only permitted from integer types to enum type.
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          VARCHAR(),
+          moodEnum_,
+          makeRowVector({makeNullableFlatVector<StringView>(
+              {"a"_sv, "b"_sv, std::nullopt})})),
+      "Cannot cast VARCHAR to test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}).");
+
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          BOOLEAN(),
+          moodEnum_,
+          makeRowVector(
+              {makeNullableFlatVector<bool>({true, false, std::nullopt})})),
+      "Cannot cast BOOLEAN to test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}).");
+
+  // Cast base type to enum type where the value does not exist in the enum.
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          BIGINT(),
+          moodEnum_,
+          makeRowVector({makeNullableFlatVector<int64_t>(
+              {0, 1, std::nullopt}, BIGINT())})),
+      "No value '1' in test.enum.mood");
+
+  // Cast enum type to different enum type.
+  std::unordered_map<std::string, int64_t> differentMap = {
+      {"CURIOUS", -2}, {"HAPPY", 3}};
+  LongEnumParameter differentEnumInfo("someEnumType", differentMap);
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          moodEnum_,
+          BIGINT_ENUM(differentEnumInfo),
+          makeRowVector({makeNullableFlatVector<int64_t>(
+              {0, 1, std::nullopt}, moodEnum_)})),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to someEnumType:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 3}).");
+
+  // Cast enum type to different enum type with same name.
+  LongEnumParameter sameNameDifferentMap("test.enum.mood", differentMap);
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          moodEnum_,
+          BIGINT_ENUM(sameNameDifferentMap),
+          makeRowVector({makeNullableFlatVector<int64_t>(
+              {0, 1, std::nullopt}, moodEnum_)})),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 3})");
+}
+
+TEST_F(BigintEnumCastTest, fromBigintEnum) {
+  // Cast enum type to base type.
+  testCast<int64_t, int64_t>(
+      moodEnum_, BIGINT(), {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  // Casting enum type to all other types including other integer types is not
+  // permitted.
+  auto enumInput = makeRowVector(
+      {makeNullableFlatVector<int64_t>({0, 1, std::nullopt}, moodEnum_)});
+  VELOX_ASSERT_THROW(
+      evaluateCast(moodEnum_, TINYINT(), enumInput),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to TINYINT.");
+
+  VELOX_ASSERT_THROW(
+      evaluateCast(moodEnum_, SMALLINT(), enumInput),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to SMALLINT.");
+
+  VELOX_ASSERT_THROW(
+      evaluateCast(moodEnum_, INTEGER(), enumInput),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to INTEGER.");
+
+  VELOX_ASSERT_THROW(
+      evaluateCast(moodEnum_, VARCHAR(), enumInput),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to VARCHAR.");
+
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          moodEnum_,
+          BOOLEAN(),
+          makeRowVector({makeNullableFlatVector<int64_t>(
+              {0, 1, std::nullopt}, moodEnum_)})),
+      "Cannot cast test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}) to BOOLEAN.");
+}
+} // namespace
+} // namespace facebook::velox
--- a/velox/functions/prestosql/tests/TypeOfTest.cpp
+++ b/velox/functions/prestosql/tests/TypeOfTest.cpp
@@ -15,6 +15,7 @@
  */
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
 #include "velox/functions/prestosql/types/BingTileType.h"
 #include "velox/functions/prestosql/types/GeometryType.h"
 #include "velox/functions/prestosql/types/HyperLogLogType.h"
@@ -26,16 +27,17 @@
 namespace facebook::velox::functions {
 namespace {
 
-class TypeOfTest : public functions::test::FunctionBaseTest {};
-
-TEST_F(TypeOfTest, basic) {
-  const auto typeOf = [&](const TypePtr& type) {
+class TypeOfTest : public functions::test::FunctionBaseTest {
+ public:
+  std::string typeOf(const TypePtr& type) {
     auto data = makeRowVector({
         BaseVector::createNullConstant(type, 1, pool()),
     });
     return evaluateOnce<std::string>("typeof(c0)", data).value();
-  };
+  }
+};
 
+TEST_F(TypeOfTest, basic) {
   EXPECT_EQ("boolean", typeOf(BOOLEAN()));
 
   EXPECT_EQ("tinyint", typeOf(TINYINT()));
@@ -53,21 +55,7 @@ TEST_F(TypeOfTest, basic) {
   EXPECT_EQ("varbinary", typeOf(VARBINARY()));
 
   EXPECT_EQ("timestamp", typeOf(TIMESTAMP()));
-  EXPECT_EQ("timestamp with time zone", typeOf(TIMESTAMP_WITH_TIME_ZONE()));
   EXPECT_EQ("date", typeOf(DATE()));
-
-  EXPECT_EQ("bingtile", typeOf(BINGTILE()));
-  EXPECT_EQ("geometry", typeOf(GEOMETRY()));
-
-  EXPECT_EQ("json", typeOf(JSON()));
-
-  EXPECT_EQ("HyperLogLog", typeOf(HYPERLOGLOG()));
-
-  EXPECT_EQ("tdigest(double)", typeOf(TDIGEST(DOUBLE())));
-
-  EXPECT_EQ("qdigest(bigint)", typeOf(QDIGEST(BIGINT())));
-  EXPECT_EQ("qdigest(real)", typeOf(QDIGEST(REAL())));
-  EXPECT_EQ("qdigest(double)", typeOf(QDIGEST(DOUBLE())));
 
   EXPECT_EQ("unknown", typeOf(UNKNOWN()));
 
@@ -83,5 +71,27 @@ TEST_F(TypeOfTest, basic) {
   VELOX_ASSERT_THROW(typeOf(OPAQUE<int>()), "Unsupported type: OPAQUE<int>");
 }
 
+TEST_F(TypeOfTest, customTypes) {
+  EXPECT_EQ("timestamp with time zone", typeOf(TIMESTAMP_WITH_TIME_ZONE()));
+  EXPECT_EQ("bingtile", typeOf(BINGTILE()));
+  EXPECT_EQ("geometry", typeOf(GEOMETRY()));
+
+  EXPECT_EQ("json", typeOf(JSON()));
+
+  EXPECT_EQ("HyperLogLog", typeOf(HYPERLOGLOG()));
+
+  EXPECT_EQ("tdigest(double)", typeOf(TDIGEST(DOUBLE())));
+
+  EXPECT_EQ("qdigest(bigint)", typeOf(QDIGEST(BIGINT())));
+  EXPECT_EQ("qdigest(real)", typeOf(QDIGEST(REAL())));
+  EXPECT_EQ("qdigest(double)", typeOf(QDIGEST(DOUBLE())));
+
+  std::unordered_map<std::string, int64_t> enumMap = {
+      {"CURIOUS", -2}, {"HAPPY", 0}};
+  LongEnumParameter moodInfo("test.enum.mood", enumMap);
+  LongEnumParameter otherInfo("someEnumType", enumMap);
+  EXPECT_EQ("test.enum.mood", typeOf(BIGINT_ENUM(moodInfo)));
+  EXPECT_EQ("someEnumType", typeOf(BIGINT_ENUM(otherInfo)));
+}
 } // namespace
 } // namespace facebook::velox::functions
--- a/velox/functions/prestosql/types/BigintEnumRegistration.cpp
+++ b/velox/functions/prestosql/types/BigintEnumRegistration.cpp
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/expression/CastExpr.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox {
+namespace {
+class BigintEnumCastOperator : public exec::CastOperator {
+ public:
+  static const std::shared_ptr<const CastOperator>& get() {
+    static const std::shared_ptr<const CastOperator> kInstance =
+        std::make_shared<const BigintEnumCastOperator>();
+
+    return kInstance;
+  }
+
+  // Casting is supported from all integer types.
+  bool isSupportedFromType(const TypePtr& other) const override {
+    return BIGINT()->equivalent(*other) || TINYINT()->equivalent(*other) ||
+        SMALLINT()->equivalent(*other) || INTEGER()->equivalent(*other);
+  }
+
+  // Casting is only supported to BIGINT type.
+  bool isSupportedToType(const TypePtr& other) const override {
+    return BIGINT()->equivalent(*other);
+  }
+
+  void castTo(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    switch (input.typeKind()) {
+      case TypeKind::TINYINT:
+        castFromInteger<int8_t>(input, context, rows, resultType, result);
+        break;
+      case TypeKind::SMALLINT:
+        castFromInteger<int16_t>(input, context, rows, resultType, result);
+        break;
+      case TypeKind::INTEGER:
+        castFromInteger<int32_t>(input, context, rows, resultType, result);
+        break;
+      case TypeKind::BIGINT:
+        castFromInteger<int64_t>(input, context, rows, resultType, result);
+        break;
+      default:
+        VELOX_UNREACHABLE(
+            "Cannot cast {} to {} type",
+            input.type()->toString(),
+            resultType->toString());
+    }
+  }
+
+  void castFrom(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->copy(&input, rows, nullptr);
+  }
+
+ private:
+  template <typename T>
+  void castFromInteger(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    auto enumType = asBigintEnum(resultType);
+    const auto* intVector = input.as<SimpleVector<T>>();
+    rows.applyToSelected([&](vector_size_t row) {
+      const int64_t intToCast = intVector->valueAt(row);
+      if (!enumType->containsValue(intToCast)) {
+        context.setStatus(
+            row,
+            Status::UserError(
+                "No value '{}' in {}", intToCast, enumType->enumName()));
+        return;
+      }
+      flatResult->set(row, intToCast);
+    });
+  }
+};
+
+class BigintEnumTypeFactory : public CustomTypeFactory {
+ public:
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK_EQ(
+        parameters.size(),
+        1,
+        "Expected exactly one type parameters for BigintEnumType");
+    VELOX_CHECK(
+        parameters[0].longEnumLiteral.has_value(),
+        "BigintEnumType parameter must be longEnumLiteral");
+    return BIGINT_ENUM(parameters[0].longEnumLiteral.value());
+  }
+
+  exec::CastOperatorPtr getCastOperator() const override {
+    return BigintEnumCastOperator::get();
+  }
+
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& /*config*/) const override {
+    return nullptr;
+  }
+};
+} // namespace
+
+void registerBigintEnumType() {
+  registerCustomType(
+      "bigint_enum", std::make_unique<const BigintEnumTypeFactory>());
+}
+} // namespace facebook::velox
--- a/velox/functions/prestosql/types/BigintEnumType.cpp
+++ b/velox/functions/prestosql/types/BigintEnumType.cpp
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <folly/Synchronized.h>
+
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox {
+
+namespace {
+std::unordered_map<int64_t, std::string> toFlippedMap(
+    const std::unordered_map<std::string, int64_t>& map,
+    const std::string& name) {
+  std::unordered_map<int64_t, std::string> flippedMap;
+  for (const auto& [key, value] : map) {
+    bool ok = flippedMap.emplace(value, key).second;
+    VELOX_USER_CHECK(
+        ok, "Invalid enum type {}, contains duplicate value {}", name, value);
+  }
+  return flippedMap;
+}
+
+std::string flippedMapToString(
+    const std::unordered_map<int64_t, std::string>& flippedMap) {
+  std::ostringstream oss;
+  oss << "{";
+  std::map<std::string, int64_t> sortedMap;
+  for (const auto& [key, value] : flippedMap) {
+    sortedMap[value] = key;
+  }
+  for (auto it = sortedMap.begin(); it != sortedMap.end(); ++it) {
+    if (it != sortedMap.begin()) {
+      oss << ", ";
+    }
+    oss << "\"" << it->first << "\"" << ": " << it->second;
+  }
+  oss << "}";
+  return oss.str();
+}
+
+} // namespace
+
+// Should only be called from get() to create a new instance.
+BigintEnumType::BigintEnumType(const LongEnumParameter& parameters)
+    : parameters_{TypeParameter(parameters)},
+      name_{parameters.name},
+      flippedMap_{toFlippedMap(parameters.valuesMap, name_)} {}
+
+std::string BigintEnumType::toString() const {
+  return fmt::format(
+      "{}:BigintEnum({})", name_, flippedMapToString(flippedMap_));
+}
+
+// A thread-safe LRU cache to store instances of BigintEnumType.
+using Cache = folly::EvictingCacheMap<
+    LongEnumParameter,
+    BigintEnumTypePtr,
+    LongEnumParameter::Hash>;
+
+BigintEnumTypePtr BigintEnumType::get(const LongEnumParameter& parameter) {
+  static const int maxCacheSize = 1000;
+  static folly::Synchronized<Cache> kCache{Cache(maxCacheSize)};
+  return kCache.withWLock([&](auto& cache) -> BigintEnumTypePtr {
+    auto it = cache.find(parameter);
+    if (it != cache.end()) {
+      return it->second;
+    }
+    // Can't use std::make_shared because calling private ctor.
+    auto instance =
+        std::shared_ptr<const BigintEnumType>(new BigintEnumType(parameter));
+    cache.insert(parameter, instance);
+    return instance;
+  });
+}
+
+folly::dynamic BigintEnumType::serialize() const {
+  folly::dynamic obj = folly::dynamic::object;
+  obj["name"] = "Type";
+  obj["type"] = name();
+  // parameters_[0].longEnumLiteral is assumed to have a value since it is
+  // constructed from a LongEnumParameter.
+  obj["kLongEnumParam"] =
+      parameters_[0].longEnumLiteral.value().serializeEnumParameter();
+  return obj;
+}
+
+} // namespace facebook::velox
--- a/velox/functions/prestosql/types/BigintEnumType.h
+++ b/velox/functions/prestosql/types/BigintEnumType.h
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <folly/container/EvictingCacheMap.h>
+
+#include "velox/type/Type.h"
+
+namespace facebook::velox {
+
+class BigintEnumType;
+using BigintEnumTypePtr = std::shared_ptr<const BigintEnumType>;
+
+/// Represents an enumerated value where the physical type is a bigint. Each
+/// enum type has a name and a set of string keys which map to bigint values,
+/// passed in as a LongEnumParameter TypeParameterKind.
+class BigintEnumType : public BigintType {
+ public:
+  static BigintEnumTypePtr get(const LongEnumParameter& parameter);
+
+  bool equivalent(const Type& other) const override {
+    return this == &other;
+  }
+
+  const char* name() const override {
+    return "BIGINT_ENUM";
+  }
+
+  const std::vector<TypeParameter>& parameters() const override {
+    return parameters_;
+  }
+
+  std::string toString() const override;
+
+  folly::dynamic serialize() const override;
+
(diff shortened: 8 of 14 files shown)


velox/docs/develop/types.rst
```diff
@@ -223,6 +224,13 @@ distribution. They may be merged without losing precision, and for storage and r
 be cast to/from VARBINARY. The parameter type (BIGINT, REAL, or DOUBLE) represents
 the set of numbers that may be ingested by the quantile digest.
 
+BIGINT_ENUM(LongEnumParameters) type represents an enumerated value where the physical type is BIGINT.
+It takes a vector of TypeParameters with 1 element as input. This TypeParameter is a LongEnumParameters,
```
kewang1024: NIT: It takes one LongEnumParameters as parameter

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+BigintEnumType::BigintEnumType(const std::vector<TypeParameter>& typeParameters)
+    : parameters_{typeParameters} {
+  VELOX_CHECK_EQ(
+      typeParameters.size(),
+      1,
+      "Expected exactly one type parameters for BigintEnumType");
+  VELOX_CHECK(
+      typeParameters[0].longEnumLiteral.has_value(),
+      "First parameter of BigintEnumType must be a long enum literal");
```
kewang1024: NIT: BigintEnumType parameter must be longEnumLiteral

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,88 @@
...
+  const std::string& enumName() const {
+    return name_;
+  }
+
+ private:
+  const std::vector<TypeParameter> parameters_;
+
+  std::string name_;
+  std::unordered_map<std::string, int64_t> longEnumMap_;
+  std::unordered_map<int64_t, std::string> flippedLongEnumMap_;
```
kewang1024: NIT:
map_
flippedMap_

velox/functions/prestosql/registration/RegistrationFunctions.cpp
```diff
@@ -135,6 +136,9 @@ void registerBitwiseFunctions(const std::string& prefix) {
 }
 
 void registerAllScalarFunctions(const std::string& prefix) {
+  // TODO: move registerBigintEnumType to registerEnumTypeFunctions once the
+  // enum functions are implemented
+  registerBigintEnumType();
```
kewang1024, line 141: do we need this as of now?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,176 @@
...
+ private:
+  // Cast is supported for all integer types.
+  // Casting to and from a different BigintEnumType is not supported
+  bool isCompatibleWith(const TypePtr& other) const {
+    return BIGINT()->equivalent(*other) || TINYINT()->equivalent(*other) ||
+        SMALLINT()->equivalent(*other) || INTEGER()->equivalent(*other);
+  }
+
+  template <typename T>
+  void castToType(
```
kewang1024: castToInteger

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,176 @@
...
+                value,
+                resultType->toString()));
+        return;
+      }
+      flatResult->set(row, static_cast<T>(value));
+    });
+  }
+
+  template <typename T>
+  void castFromType(
```
kewang1024: castFromInteger

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,176 @@
...
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    VELOX_CHECK(
+        isBigintEnumType(resultType),
+        "Invalid CAST TO type for BigintEnumType");
```
kewang1024: Move this check to `CastTo`

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    const auto* enumInts = input.as<SimpleVector<int64_t>>();
+    context.applyToSelectedNoThrow(rows, [&](auto row) {
+      int64_t value = enumInts->valueAt(row);
+      if (value < std::numeric_limits<T>::min() ||
+          value > std::numeric_limits<T>::max()) {
+        context.setStatus(
+            row,
+            Status::UserError(
+                "Value '{}' out of range for {}",
```
kewang1024: Cannot cast BigintEnum to {} type: Value '{}' out of range for {}

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,176 @@
...
+      const TypePtr& resultType,
+      VectorPtr& result) const {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    VELOX_CHECK(
+        isBigintEnumType(resultType),
+        "Invalid CAST TO type for BigintEnumType");
+    auto toType = std::dynamic_pointer_cast<const BigintEnumType>(resultType);
```
kewang1024: NIT: enumType

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,88 @@
...
+  folly::dynamic serializeEnumParameter(
+      const std::string& name,
+      const std::unordered_map<std::string, int64_t>& map) const;
+
+  bool containsValue(int64_t value) const {
+    return flippedLongEnumMap_.find(value) != flippedLongEnumMap_.end();
+  }
+
+  const std::string& enumName() const {
+    return name_;
```
kewang1024, line 55: return fmt::format("BigintEnum({})", name_);

velox/functions/prestosql/registration/RegistrationFunctions.cpp
HeidiHan0000: I needed it to test out queries for e2e. I removed it in the PR where I register the enum_key function.

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: This enumName() is used in the TypeOf function, and the Java TypeOf for these types only return the enum name, not the "BigintEnum" part. Would it be alright to keep as is?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& /*config*/) const override {
+    return nullptr;
+  }
+};
+} // namespace
+
+void registerBigintEnumType() {
+  registerCustomType(
+      "BIGINT_ENUM", std::make_unique<const BigintEnumTypeFactory>());
```
kewang1024: It seems that in our code base, other types are registered as low case

velox/functions/prestosql/TypeOf.cpp
```diff
@@ -59,7 +60,11 @@ std::string typeName(const TypePtr& type) {
       if (isBingTileType(type)) {
         return "bingtile";
       }
+      if (isBigintEnumType(type)) {
+        return (dynamic_pointer_cast<const BigintEnumType>(type))->enumName();
```
kagamiori: nit: You can add a helper method `asBigintEnum()` in BigintEnumType.h/cpp.

velox/type/Type.cpp
```diff
@@ -117,22 +149,35 @@ TypePtr Type::create(const folly::dynamic& obj) {
...
-      params.emplace_back(child);
+    if (obj.find("cTypes") != obj.items().end()) {
+      params.reserve(childTypes.size());
+      for (auto& child : childTypes) {
+        params.emplace_back(child);
+      }
     }
+    if (obj.find("kLongEnumParam") != obj.items().end()) {
+      params.emplace_back(deserializeEnumParam<int64_t>(obj["kLongEnumParam"]));
+    }
```
kagamiori: Is there a unit test that covers this change?

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,87 @@
...
+/// bigint values, passed in as a LongEnumParameter TypeParameterKind.
+class BigintEnumType : public BigintType {
+ public:
+  explicit BigintEnumType(const std::vector<TypeParameter>& typeParameters);
+
+  static BigintEnumTypePtr get(
+      const std::vector<TypeParameter>& typeParameters);
+
+  static std::string mapToString(
+      std::unordered_map<std::string, int64_t> longEnumMap);
```
kagamiori: Make this a private method. Also it doesn't need to be static?

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,87 @@
...
+
+  std::string toString() const override {
+    return fmt::format("{}:BigintEnum({}{})", name_, name_, mapToString(map_));
+  }
+
+  folly::dynamic serialize() const override;
+
+  folly::dynamic serializeEnumParameter(
+      const std::string& name,
+      const std::unordered_map<std::string, int64_t>& map) const;
```
kagamiori: Make this a private method too.

velox/functions/prestosql/types/tests/BigintEnumTypeTest.cpp
```diff
@@ -0,0 +1,59 @@
...
+  ASSERT_STREQ(BIGINT_ENUM(typeParameters)->kindName(), "BIGINT");
+  ASSERT_EQ(BIGINT_ENUM(typeParameters)->enumName(), "test.enum.mood");
+  ASSERT_EQ(BIGINT_ENUM(typeParameters)->parameters().size(), 1);
+  ASSERT_EQ(
+      BIGINT_ENUM(typeParameters)->toString(),
+      "test.enum.mood:BigintEnum(test.enum.mood{\"CURIOUS\": -2, \"HAPPY\": 0})");
+
+  ASSERT_TRUE(hasType("BIGINT_ENUM"));
+  ASSERT_EQ(
+      getType("BIGINT_ENUM", typeParameters), BIGINT_ENUM(typeParameters));
```
kagamiori: Could we also add a test case that checks two different bigint enum types are not equivalent or equal (i.e., Type::equivalent() and Type::operator==())? Also that the types got from the same type parameters are equal and equivalent?

velox/functions/prestosql/types/BigintEnumType.cpp
kagamiori: This constructor shouldn't be used directly. Let's make it private. Let's make it private. These two checks can also be removed since they're already checked in get().

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+  }
+  obj["valuesMap"] = valuesMap;
+  return obj;
+}
+
+folly::dynamic BigintEnumType::serialize() const {
+  folly::dynamic obj = folly::dynamic::object;
+  obj["name"] = "Type";
+  obj["type"] = name();
+  obj["kLongEnumParam"] = serializeEnumParameter(name_, map_);
```
kagamiori: nit: serializeEnumParameter is also a member method of BigintEnumType, so it can access name_ and map_ directly. You don't need to pass them.

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+folly::dynamic BigintEnumType::serializeEnumParameter(
+    const std::string& name,
+    const std::unordered_map<std::string, int64_t>& map) const {
+  folly::dynamic obj = folly::dynamic::object;
+  obj["enumName"] = name;
+  folly::dynamic valuesMap = folly::dynamic::array;
+  for (const auto& [key, value] : map) {
+    folly::dynamic pair = folly::dynamic::array(key, value);
+    valuesMap.push_back(std::move(pair));
+  }
```
kagamiori: Why do we serialize map as an array of arrays rather than a folly::dynamic::object?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return isCompatibleWith(other);
+  }
+
+  void castTo(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    VELOX_CHECK(
```
kagamiori: Use VELOX_USER_CHECK here. Also, maybe make the error message more informative, such as "Invalid cast from {} to {}" with the enum type names.

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+        VELOX_UNREACHABLE("Unsupported type: {}", input.type()->toString());
+    }
+  }
+
+ private:
+  // Cast is supported for all integer types.
+  // Casting to and from a different BigintEnumType is not supported
+  bool isCompatibleWith(const TypePtr& other) const {
+    return BIGINT()->equivalent(*other) || TINYINT()->equivalent(*other) ||
+        SMALLINT()->equivalent(*other) || INTEGER()->equivalent(*other);
```
kagamiori: Could you also check whether casting to/from boolean is allowed in Java? Because boolean is often considered an integral type as well.

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    const auto* enumInts = input.as<SimpleVector<int64_t>>();
+    context.applyToSelectedNoThrow(rows, [&](auto row) {
+      int64_t value = enumInts->valueAt(row);
+      if (value < std::numeric_limits<T>::min() ||
+          value > std::numeric_limits<T>::max()) {
+        context.setStatus(
+            row,
+            Status::UserError(
+                "Cannot cast BigintEnum to {} type: Value '{}' out of range for {}",
```
kagamiori: nit: shall we also give the enum type name in the error message?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return BIGINT_ENUM(parameters);
+  }
+
+  exec::CastOperatorPtr getCastOperator() const override {
+    return BigintEnumCastOperator::get();
+  }
+
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& /*config*/) const override {
+    return nullptr;
```
kagamiori, line 128: Do we have a plan to support bigint enum type in fuzzer later?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,177 @@
...
+      }
+      flatResult->set(row, intToCast);
+    });
+  }
+};
+
+class BigintEnumTypeFactory : public CustomTypeFactory {
+ public:
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK_EQ(parameters.size(), 1);
```
kagamiori: nit: This check is not needed since it's already checked inside BIGINT_ENUM().

kagamiori (commented):
Hi @HeidiHan0000, thank you for adding the support of bigint enum type! I left a few comments.

velox/type/Type.cpp
HeidiHan0000: Yes, the serde test in BigintEnumTypeTest comes here

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: I made it static because it is used in the static `get` function where the enum name and map is used as a concatenated string to search in the instances cache, so that it doesn't create a new one if it already exists

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: Good point, I've changed it to just use a folly::dynamic::object. Thanks!

velox/functions/prestosql/types/BigintEnumRegistration.cpp
HeidiHan0000: Thanks for calling it out, I checked and Java also does not allow it. I've added a test case for boolean casting.

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: I had previously made the constructor public because in a previous comment I was advised to use make_shared over shared_ptr, but I think this makes more sense. Please let me know if you think otherwise!

velox/functions/prestosql/types/BigintEnumRegistration.cpp
HeidiHan0000: Yes I will look more into it, probably in a following PR.

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,89 @@
...
+inline BigintEnumTypePtr BIGINT_ENUM(
+    const std::vector<TypeParameter>& typeParameters) {
+  return BigintEnumType::get(typeParameters);
+}
+
+FOLLY_ALWAYS_INLINE bool isBigintEnumType(const TypePtr& type) {
+  return dynamic_cast<const BigintEnumType*>(type.get()) != nullptr;
+}
+
+FOLLY_ALWAYS_INLINE std::optional<BigintEnumTypePtr> asBigintEnumType(
```
kagamiori: nit: Just make it return BigintEnumTypePtr. If the dynamic_pointer_cast fails, it will return a nullptr anyway.

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+std::string BigintEnumType::mapToString(
+    std::unordered_map<std::string, int64_t> valuesMap) {
+  // The values map should be sorted when converting to string for 2 reasons:
+  // 1. Printing the type will be consistent.
+  // 2. The enum name + values map is used as the key in the cache for get(), so
+  // the string should be consistent.
+  std::map<std::string, int64_t> sortedMap(valuesMap.begin(), valuesMap.end());
```
kagamiori: Curious, does BigintEnumType really need to store map_? It looks like map_ is used either through mapToString or serializeEnumParameters, both can just loop through flippedMap_?

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox {
+
+BigintEnumType::BigintEnumType(const std::vector<TypeParameter>& typeParameters)
+    : parameters_{typeParameters} {
+  name_ = typeParameters[0].longEnumLiteral.value().name;
+  map_ = typeParameters[0].longEnumLiteral.value().valuesMap;
+  for (const auto& [key, value] : map_) {
+    flippedMap_[value] = key;
```
kagamiori: Let's add a check that `flippedMap_.count(value) == 0` before `flippedMap_[value] = key`.

Let's also add a test case where there are multiple keys map to the same value. (Is this allowed?)

HeidiHan0000: I believe this is not allowed, I have added the check and test, thanks!

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: You are right, I removed the map_ and just added more variants of the mapToString function. Please let me know if this feels messier.

velox/functions/prestosql/tests/TypeOfTest.cpp
```diff
@@ -69,6 +70,19 @@ TEST_F(TypeOfTest, basic) {
   EXPECT_EQ("qdigest(real)", typeOf(QDIGEST(REAL())));
   EXPECT_EQ("qdigest(double)", typeOf(QDIGEST(DOUBLE())));
 
+  auto enumName = "test.enum.mood";
```
mbasmanova: This test is very long. Would you move the new logic into a separate test method?

velox/functions/prestosql/tests/TypeOfTest.cpp
```diff
@@ -69,6 +70,19 @@ TEST_F(TypeOfTest, basic) {
...
+  auto enumName2 = "someEnumType";
+  std::unordered_map<std::string, int64_t> enumMap = {
+      {"CURIOUS", -2}, {"HAPPY", 0}};
+  LongEnumParameter longEnumParameter(enumName, enumMap);
```
mbasmanova: longEnumParameter is long and hard to distinguish from longEnumParameter2

consider renaming to something like 'feeling' and 'mood'.

velox/functions/prestosql/tests/TypeOfTest.cpp
```diff
@@ -69,6 +70,19 @@ TEST_F(TypeOfTest, basic) {
   EXPECT_EQ("qdigest(real)", typeOf(QDIGEST(REAL())));
   EXPECT_EQ("qdigest(double)", typeOf(QDIGEST(DOUBLE())));
 
+  auto enumName = "test.enum.mood";
+  auto enumName2 = "someEnumType";
```
mbasmanova: drop this variable and enumName

velox/functions/prestosql/tests/TypeOfTest.cpp
```diff
@@ -69,6 +70,19 @@ TEST_F(TypeOfTest, basic) {
...
+  std::unordered_map<std::string, int64_t> enumMap = {
+      {"CURIOUS", -2}, {"HAPPY", 0}};
+  LongEnumParameter longEnumParameter(enumName, enumMap);
+  LongEnumParameter longEnumParameter2(enumName2, enumMap);
+  const std::vector<TypeParameter>& typeParameters = {
```
mbasmanova: do not store new values in a reference

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+class BigintEnumCastTest : public functions::test::CastBaseTest {
+ protected:
+  BigintEnumCastTest() {
+    registerBigintEnumType();
+  }
+};
+
+TEST_F(BigintEnumCastTest, toBigintEnum) {
+  auto enumName = "test.enum.mood";
```
mbasmanova: const

enumName - mood
enumName2 - feeling
enumName3 - <something else>

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+  std::unordered_map<std::string, int64_t> enumMap = {
+      {"CURIOUS", -2}, {"HAPPY", 0}};
+  LongEnumParameter longEnumParameter(enumName, enumMap);
```
mbasmanova: Is this variable needed?

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+  const std::vector<TypeParameter>& typeParameters = {
+      TypeParameter(longEnumParameter)};
+
+  // Cast base type to enum type
```
mbasmanova: Make sure all comments are full sentences. Start with a capital letter and end with a period.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+  testCast<int64_t, int64_t>(
+      BIGINT(),
+      BIGINT_ENUM(typeParameters),
```
mbasmanova: Let's add a variable for this type to avoid copy-paste.

```
auto moodEnum = BIGINT_ENUM(LongEnumParameter{"mood", {{"HAPPY", 1}, {"SAD", 2}}})
```

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+      {0, -2, std::nullopt},
+      {0, -2, std::nullopt});
+
+  // Cast TINYINT to enum type
```
mbasmanova: These comment are redundant. They simply repeat the code. Consider dropping.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+              {0, 1, std::nullopt}, BIGINT())})),
+      "No value '1' in test.enum.mood");
+
+  // Cast enum type to different enum type
+  std::string enumName2 = "someEnumType";
+  std::unordered_map<std::string, int64_t> enumMap2 = {
+      {"CURIOUS", -2}, {"HAPPY", 3}};
+  LongEnumParameter longEnumParameter2(enumName2, enumMap2);
+  auto typeParameters2 = {TypeParameter(longEnumParameter2)};
+  VELOX_ASSERT_THROW(
```
mbasmanova, line 90: Let's move negative tests into a separate test method.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+          BOOLEAN(),
+          BIGINT_ENUM(typeParameters),
+          makeRowVector(
+              {makeNullableFlatVector<bool>({true, false, std::nullopt})})),
+      "Cannot cast BOOLEAN to test.enum.mood:BigintEnum(test.enum.mood{\"CURIOUS\": -2, \"HAPPY\": 0}).");
+}
+
+TEST_F(BigintEnumCastTest, fromBigintEnum) {
+  auto enumName = "test.enum.mood";
+  std::unordered_map<std::string, int64_t> enumMap = {
```
mbasmanova: Consider creating test enum types in the ctor and store them in member variables. Then reuse across all tests.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+      BIGINT(),
+      {0, -2, std::nullopt},
+      {0, -2, std::nullopt});
+
+  // Cast enum type to TINYINT type
+  VELOX_ASSERT_THROW(
+      evaluateCast(
+          BIGINT_ENUM(typeParameters),
+          TINYINT(),
+          makeRowVector({makeNullableFlatVector<int64_t>(
```
mbasmanova: extract a temp variable to avoid copy-paste

velox/type/Type.h
```diff
@@ -392,13 +392,23 @@ using TypePtr = std::shared_ptr<const Type>;
 /// A struct which represents the parameters for a BigintEnumType.
 /// Consists of the name of the enum and a map of string keys to bigint values.
 struct LongEnumParameter {
+  LongEnumParameter(
```
mbasmanova, line 396: Do we need this ctor?

velox/type/Type.h
```diff
@@ -392,13 +392,23 @@ using TypePtr = std::shared_ptr<const Type>;
...
+      : name{std::move(_name)}, valuesMap{std::move(_valuesMap)} {}
+
   std::string name;
   std::unordered_map<std::string, int64_t> valuesMap;
 };
 
 /// A struct which represents the parameters for a VarcharEnumType.
 /// Consists of the name of the enum and a map of string keys to string values.
 struct VarcharEnumParameter {
+  VarcharEnumParameter(
```
mbasmanova: Same question.

velox/type/Type.h
```diff
@@ -392,13 +392,23 @@ using TypePtr = std::shared_ptr<const Type>;
 /// A struct which represents the parameters for a BigintEnumType.
```
mbasmanova: > A struct which represents

Drop this prefix as it is redundant. Start with an active verb. "Represents parameters for ..."

velox/type/Type.cpp
```diff
@@ -109,6 +109,35 @@ namespace {
...
+  auto valuesMap = obj["valuesMap"];
+
+  // Construct the values map
+  std::unordered_map<std::string, ValueType> map;
+  for (const auto& item : valuesMap.items()) {
+    std::string key = item.first.asString();
+    if constexpr (std::is_same_v<ValueType, int64_t>) {
+      int64_t value = item.second.asInt();
+      map.emplace(std::move(key), value);
+    } else if constexpr (std::is_same_v<ValueType, std::string>) {
```
mbasmanova: Can you add code that will go through this code path? Otherwise, this is untested and not clear if it works.

velox/type/Type.cpp
```diff
@@ -109,6 +109,35 @@ namespace {
...
+      std::string value = item.second.asString();
+      map.emplace(std::move(key), value);
+    }
+  }
+
+  // Construct the corresponding TypeParameter
+  if constexpr (std::is_same_v<ValueType, int64_t>) {
+    return TypeParameter(std::move(LongEnumParameter(enumName, map)));
```
mbasmanova: do we need std::move here?

Notice that TypeParameter's argument is a reference. You may want to change it to take arg by value, then move it.

velox/functions/prestosql/types/tests/BigintEnumTypeTest.cpp
```diff
@@ -0,0 +1,103 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/functions/prestosql/types/tests/TypeTestBase.h"
+
+namespace facebook::velox::test {
```
mbasmanova: tests go in anonymous namespace within the namespace that contains code being tested

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+#include <folly/container/EvictingCacheMap.h>
+
+#include "velox/type/Type.h"
+
+namespace facebook::velox {
+
+class BigintEnumType;
+using BigintEnumTypePtr = std::shared_ptr<const BigintEnumType>;
+
+/// BigintEnumType represents an enumerated value where the physical type is a
```
mbasmanova: drop 'BigintEnumType'

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+/// bigint. Each enum type has a name and a set of string keys which map to
+/// bigint values, passed in as a LongEnumParameter TypeParameterKind.
+class BigintEnumType : public BigintType {
+ public:
+  static BigintEnumTypePtr get(
+      const std::vector<TypeParameter>& typeParameters);
```
mbasmanova: Why is this a vector? Shouldn't it be just one LongEnumParameter value?

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+    return "BIGINT_ENUM";
+  }
+
+  const std::vector<TypeParameter>& parameters() const override {
+    return parameters_;
+  }
+
+  std::string toString() const override {
+    return fmt::format(
+        "{}:BigintEnum({}{})", name_, name_, flippedMapToString(flippedMap_));
```
mbasmanova: Do we need to include name_ twice? Why?

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  folly::dynamic serialize() const override;
+
+  bool containsValue(int64_t value) const {
+    return flippedMap_.find(value) != flippedMap_.end();
```
mbasmanova: flippedMap_.contains

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  const std::string& enumName() const {
+    return name_;
+  }
+
+ private:
+  explicit BigintEnumType(const std::vector<TypeParameter>& typeParameters);
+
+  static std::string flippedMapToString(
```
mbasmanova: Let's move these into free functions in .cpp

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+      const std::unordered_map<int64_t, std::string>& flippedMap);
+  static std::string mapToString(
+      const std::unordered_map<std::string, int64_t>& unorderedMap);
+  static std::string orderedMapToString(
+      const std::map<std::string, int64_t>& longEnumMap);
+
+  folly::dynamic serializeEnumParameter() const;
+
+  const std::vector<TypeParameter> parameters_;
+  std::string name_;
```
mbasmanova: const

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+  std::unordered_map<int64_t, std::string> flippedMap_;
```
mbasmanova: const

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+};
+
+inline BigintEnumTypePtr BIGINT_ENUM(
+    const std::vector<TypeParameter>& typeParameters) {
+  return BigintEnumType::get(typeParameters);
+}
+
+FOLLY_ALWAYS_INLINE bool isBigintEnumType(const TypePtr& type) {
+  return dynamic_cast<const BigintEnumType*>(type.get()) != nullptr;
```
mbasmanova: cast is expensive, let's first check type kind

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,91 @@
...
+}
+
+FOLLY_ALWAYS_INLINE BigintEnumTypePtr asBigintEnumType(const TypePtr& type) {
```
mbasmanova: asBigintEnumType -> asBigintEnum

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <folly/Synchronized.h>
+
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox {
+
+BigintEnumType::BigintEnumType(const std::vector<TypeParameter>& typeParameters)
+    : parameters_{typeParameters} {
+  name_ = typeParameters[0].longEnumLiteral.value().name;
```
mbasmanova: check that there is only one value in typeParameters

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,131 @@
...
+  for (const auto& [key, value] :
+       typeParameters[0].longEnumLiteral.value().valuesMap) {
+    VELOX_USER_CHECK(
```
mbasmanova, line 30: No need to lookup twice.

```
bool ok = flippedMap_.emplace(...).second;
VELOX_USER_CHECK(ok, ...);
```

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,131 @@
...
+    const std::unordered_map<std::string, int64_t>& unorderedMap) {
+  std::map<std::string, int64_t> sortedMap(
+      unorderedMap.begin(), unorderedMap.end());
+  for (const auto& [key, value] : unorderedMap) {
+    sortedMap[key] = value;
+  }
+  return orderedMapToString(sortedMap);
+}
+
+std::string BigintEnumType::orderedMapToString(
```
mbasmanova: the callers already sorted the map... why do we sort it again?

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,131 @@
...
+  static Cache kCache{
+      folly::EvictingCacheMap<std::string, BigintEnumTypePtr>(maxCacheSize)};
+
+  return kCache.withWLock([&](auto& cache) -> BigintEnumTypePtr {
+    auto it = cache.find(mapKey);
+    if (it != cache.end()) {
+      return it->second;
+    }
+
+    auto instance = std::shared_ptr<const BigintEnumType>(
```
mbasmanova: std::make_shared

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,131 @@
...
+        new BigintEnumType(typeParameters));
+    cache.insert(mapKey, instance);
+    return instance;
+  });
+}
+
+folly::dynamic BigintEnumType::serializeEnumParameter() const {
```
mbasmanova: Let's make this a method on LongEnumParameter

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+        asBigintEnumType(input.type())->toString());
+
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    const auto* enumInts = input.as<SimpleVector<int64_t>>();
+    context.applyToSelectedNoThrow(rows, [&](auto row) {
+      int64_t value = enumInts->valueAt(row);
+      flatResult->set(row, static_cast<int64_t>(value));
```
mbasmanova: Do we need static_cast ?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    auto enumType = asBigintEnumType(resultType);
+    const auto* intVector = input.as<SimpleVector<T>>();
+    context.applyToSelectedNoThrow(rows, [&](auto row) {
```
mbasmanova: no need to use applyToSelectedNoThrow; just rows.applyToSelected

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+      const int64_t intToCast = static_cast<int64_t>(intVector->valueAt(row));
+      if (!enumType->containsValue(intToCast)) {
+        context.setStatus(
+            row,
+            Status::UserError(
+                "No value '{}' in {}", intToCast, enumType->enumName()));
+        return;
+      }
+      flatResult->set(row, intToCast);
```
mbasmanova, line 104: We can avoid copying if 'input' vector is single-referenced and can be reused.

mbasmanova (commented):
@HeidiHan0000 Some comments.

velox/type/Type.h
HeidiHan0000: I thought I did, but now can't find what it was. Removed!

velox/type/Type.cpp
HeidiHan0000: Got it, I had it for the VarcharEnumTypes, but I removed it and added a TODO. I will add it back when I implement VarcharEnumTypes!

velox/type/Type.cpp
HeidiHan0000: Thanks, changed!

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: It is just one LongEnumParameter, but CustomTypeFactory::getType receives a `std::vector<TypeParameter>` (I think it is for decimal which accepts 2 TypeParameters of kind longLiteral for precision and scale?). Then `getType` eventually calls this `get` with the same `std::vector<TypeParameter>`, which matches the other parametric custom types' `get` functions.

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: This is to match the Presto format which looks like `test.enum.mood:BigintEnum(test.enum.mood{\"CURIOUS\": -2, \"HAPPY\": 0})`.

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: flippedMap_ is populated element by element in the constructor, so I couldn't mark it as const

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: This constructor is only called from the get function, which has the check for typeParameters.size()== 1 and that it is a longEnumParameter. Would adding the check here again be redundant, or would it be better to be redundant but safe? I added this as a comment for now, but please let me know what you think!

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: You are right, removed, thanks!

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: I had changed it back to this to make the constructor private, since it should not be constructed anywhere else other than from the get function which will store it in the instances cache. Would it be better to make it public and use make_shared?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
HeidiHan0000: Would you be able to clarify what reusing the input vector would look like? I would still need to go through rows to check the values, right?

velox/functions/prestosql/types/BigintEnumType.h
mbasmanova: I believe the caller of this API knows that there is only one parameter. Hence, I suggest to tighten this API to take a single parameter and update the caller.

```
inline BigintEnumTypePtr BIGINT_ENUM(
    const std::vector<TypeParameter>& typeParameters) {
  return BigintEnumType::get(typeParameters);
}
```

velox/functions/prestosql/types/BigintEnumType.h
mbasmanova: Is it important to match Presto's format? I assume not. If so, let's change to eliminate redundancy.

mbasmanova: The following is shorter and easier to read:

`test.enum.mood:BigintEnum({"CURIOUS": -2, "HAPPY": 0})`

velox/functions/prestosql/types/BigintEnumType.h
mbasmanova: let's introduce a helper function toFlippedMap and call it from the ctor. This would allow to mark this variable const.

velox/functions/prestosql/types/BigintEnumType.cpp
mbasmanova: Hmm... then change the ctor to take a single LongEnumParameter.

velox/functions/prestosql/types/BigintEnumType.cpp
mbasmanova: Got it. That's fine. Perhaps, add a short comment to help future readers.

// Can't use std::make_shared because calling private ctor.

velox/functions/prestosql/types/BigintEnumRegistration.cpp
kgpai: @HeidiHan0000  I think @mbasmanova  is saying that If input vector is singly referenced and is of same type as the result then we can just return that as is without copying value. We just need to ensure that the value in the the legal range of values for the enum. 

(Something like below) 

```
if (input->type() == resultType && input.use_count() == 1 && input->isFlatEncoding() && input->isWriteable()) {
 // We can reuse the vector instead of doing a copy. 
input->prepareForReuse();
input->resize(rows.end());
result = std::move(input); 
}
```

velox/docs/develop/types.rst
```diff
@@ -223,6 +224,13 @@ distribution. They may be merged without losing precision, and for storage and r
 be cast to/from VARBINARY. The parameter type (BIGINT, REAL, or DOUBLE) represents
 the set of numbers that may be ingested by the quantile digest.
 
+BIGINT_ENUM(LongEnumParameter) type represents an enumerated value where the physical type is BIGINT.
+It takes one LongEnumParameters as parameter, which consists of a string name and a mapping of
```
mbasmanova: typo: LongEnumParameters -> LongEnumParameter

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/functions/prestosql/tests/CastBaseTest.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+using namespace facebook::velox;
+using namespace facebook::velox::test;
+using namespace facebook::velox::exec;
+
+namespace {
```
mbasmanova, line 26: Make sure to put test into anonymous namespace inside the namespace that holds code being tested.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,157 @@
...
+ protected:
+  BigintEnumCastTest() {
+    registerBigintEnumType();
+    LongEnumParameter moodInfo(
+        "test.enum.mood", {{"CURIOUS", -2}, {"HAPPY", 0}});
+    std::vector<TypeParameter> moodTypeParams = {TypeParameter(moodInfo)};
+    moodEnum = BIGINT_ENUM(moodTypeParams);
+  }
+
+  BigintEnumTypePtr moodEnum;
```
mbasmanova: naming: moodEnum -> moodEnum_

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
```diff
@@ -0,0 +1,157 @@
...
+      SMALLINT(), moodEnum, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  testCast<int32_t, int64_t>(
+      INTEGER(), moodEnum, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  // Cast enum type to same enum type.
+  testCast<int64_t, int64_t>(
+      moodEnum, moodEnum, {0, -2, std::nullopt}, {0, -2, std::nullopt});
+
+  // Cast enum type to same enum type where the values map is in a different
```
mbasmanova: How does this work? Do we allow 2 different BIGINT_ENUM types with the same name? Is this intentional?

velox/functions/prestosql/tests/TypeOfTest.cpp
```diff
@@ -26,16 +27,18 @@
 namespace facebook::velox::functions {
 namespace {
 
-class TypeOfTest : public functions::test::FunctionBaseTest {};
+class TypeOfTest : public functions::test::FunctionBaseTest {
+ public:
+  const std::function<std::string(const TypePtr&)> typeOf =
```
mbasmanova: turn this into a method

```
std::string typeOf(const TypePtr& type) {...}
```

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/expression/CastExpr.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox {
+namespace {
+class BigintEnumCastOperator : public exec::CastOperator {
+ public:
+  static const std::shared_ptr<const CastOperator>& get() {
+    static const std::shared_ptr<const CastOperator> instance =
```
mbasmanova: naming: instance -> kInstance

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  void castFrom(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    VELOX_USER_CHECK(
```
mbasmanova: VELOX_USER_CHECK_EQ

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,148 @@
...
+        resultType->kind() == TypeKind::BIGINT,
+        "Cannot cast {} to {} type",
+        input.type()->toString(),
+        asBigintEnum(input.type())->toString());
```
mbasmanova: Do we need asBigintEnum ? Can we just call input.type()->toString() ? 

Looks like we use same type for 'to' and 'from' in this error message.

velox/functions/prestosql/types/BigintEnumRegistration.cpp
mbasmanova: Let's discuss this separately. For now, just note that copying values one by one is very slow.

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+      name_{typeParameters[0].longEnumLiteral.value().name} {
+  for (const auto& [key, value] :
+       typeParameters[0].longEnumLiteral.value().valuesMap) {
+    bool ok = flippedMap_.emplace(value, key).second;
+    VELOX_USER_CHECK(
+        ok, "Invalid enum type {}, contains duplicate value {}", name_, value);
+  }
+}
+
+std::string orderedMapToString(
```
mbasmanova: Make sure to put helper functions into anonymous namespace.

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+    const std::vector<TypeParameter>& typeParameters) {
+  VELOX_CHECK_EQ(
+      typeParameters.size(),
+      1,
+      "Expected exactly one type parameters for BigintEnumType");
+  VELOX_CHECK(
+      typeParameters[0].longEnumLiteral.has_value(),
+      "BigintEnumType parameter must be longEnumLiteral");
+
+  const auto& mapKey = typeParameters[0].longEnumLiteral.value().name +
```
mbasmanova: Using string as a key is inefficient. Why can't we use the type itself?

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,114 @@
...
+      mapToString(typeParameters[0].longEnumLiteral.value().valuesMap);
+
+  const int maxCacheSize = 1000;
```
mbasmanova: static

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,80 @@
...
+  const std::string name_;
+  std::unordered_map<int64_t, std::string> flippedMap_;
+};
+
+inline BigintEnumTypePtr BIGINT_ENUM(
+    const std::vector<TypeParameter>& typeParameters) {
+  return BigintEnumType::get(typeParameters);
+}
+
+FOLLY_ALWAYS_INLINE bool isBigintEnumType(const TypePtr& type) {
```
mbasmanova: change TypePtr to Type to allow calling this method without having access to a shared_ptr

velox/functions/prestosql/types/tests/BigintEnumTypeTest.cpp
mbasmanova: This should be facebook::velox, because this is where BigintEnumType is defined.

velox/functions/prestosql/types/tests/BigintEnumTypeTest.cpp
```diff
@@ -0,0 +1,96 @@
...
+ protected:
+  BigintEnumTypeTest() {
+    registerBigintEnumType();
+    LongEnumParameter moodInfo(
+        "test.enum.mood", {{"CURIOUS", -2}, {"HAPPY", 0}});
+    std::vector<TypeParameter> moodTypeParams = {TypeParameter(moodInfo)};
+    moodEnum = BIGINT_ENUM(moodTypeParams);
+  }
+
+  BigintEnumTypePtr moodEnum;
```
mbasmanova: naming: moodEnum -> moodEnum_

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: I agree. If it is not a big deal that the error messages may look different and that it doesn't match the exact format of the serialized Presto type, I can change it to this new format

mbasmanova: Let's change.

velox/functions/prestosql/tests/BigintEnumCastTest.cpp
HeidiHan0000: Removing since it is a duplicate check of the case `Different TypeParameters with same enumName and enumMap but in different order` in BigintEnumTypeTest

velox/functions/prestosql/types/BigintEnumRegistration.cpp
HeidiHan0000: Actually I had put these as a just in case safety check, but I don't think they are necessary and the caller should be failing these cases. Removed them

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: This cache is used to return the type if already created, so I'm not sure if using the type makes sense. Did you mean the LongEnumParameter? In that case, would defining a custom hash for LongEnumParameter (which has the std::string name and std::unordered_map values) be a better approach?

mbasmanova: > would defining a custom hash for LongEnumParameter (which has the std::string name and std::unordered_map values) be a better approach?

Yes, I think so. 

See `Variant::hash<TypeKind::MAP>` for an example of hashing map in a way that doesn't care about the order of entries.

```
template <>
uint64_t Variant::hash<TypeKind::MAP>() const {
  const auto& mapVariant = value<TypeKind::MAP>();
  uint64_t hash = bits::kNullHash;
  for (const auto& [key, value] : mapVariant) {
    const auto elementHash = bits::hashMix(key.hash(), value.hash());
    hash = bits::commutativeHashMix(hash, elementHash);
  }
  return hash;
}
```

mbasmanova (approved):

Issue #14188, opened 2025-07-22
Build is failing with pessimizing-move of std::move(zoomCheck)


majetideepak:
### Problem description

```
/__w/velox/velox/velox_main/./velox/functions/prestosql/BingTileFunctions.h:302:33: error: moving a local object in a return statement prevents copy elision [-Werror=pessimizing-move]
    302 |       return std::move(zoomCheck);
```
https://github.com/facebookincubator/velox/actions/workflows/scheduled.yml
Introduced here https://github.com/facebookincubator/velox/pull/14100

### System information

CI

### CMake log

```bash

```

jkhaliqi:
Change is being made in https://github.com/facebookincubator/velox/pull/14190

majetideepak:
Closed with #14190

PR #14189, opened 2025-07-22
fix: Remove std::move in return call

jkhaliqi:
fix: #14188


jkhaliqi:
Repeat of this PR: https://github.com/facebookincubator/velox/pull/14190

PR #14190, opened 2025-07-22
fix: Moving a local variable in return statements in BingTileFunctions.h is causing builds to fail

kevinwilfong:
Summary:
Builds are currently failing with the following error message

 error: moving a local object in a return statement prevents copy elision [-Werror=pessimizing-move]

referring to the lines recently added to BingTileFunctions.h which do

return std::move(zoomCheck);

The fix is to remove the std::move and let the compiler do it automatically.

D78198453 introduced the code in question.

Differential Revision: D78769633

### DIFF
--- a/velox/functions/prestosql/BingTileFunctions.h
+++ b/velox/functions/prestosql/BingTileFunctions.h
@@ -54,9 +54,9 @@ struct BingTileFunction {
           fmt::format("Bing tile Y coordinate {} cannot be negative", y));
     }
 
-    auto zoomCheck = checkBingTileZoom(zoom);
+    Status zoomCheck = checkBingTileZoom(zoom);
     if (FOLLY_UNLIKELY(!zoomCheck.ok())) {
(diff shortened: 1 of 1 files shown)


velox/functions/prestosql/BingTileFunctions.h
```diff
@@ -56,7 +56,7 @@ struct BingTileFunction {
 
     auto zoomCheck = checkBingTileZoom(zoom);
     if (FOLLY_UNLIKELY(!zoomCheck.ok())) {
-      return std::move(zoomCheck);
+      return zoomCheck;
```
kgpai, line 59: Can change the auto to status too to get copy ellision.

kgpai (approved):

kevinwilfong:
Fuzzer Jobs / Build builds PyVelox from trunk which is currently broken, so won't pass until my change lands.

Other jobs should pass

Issue #14122, opened 2025-07-16
Fuzzer produces mismatch for function with null input


peterenescu:
### Bug description

SOT Fuzzer may fail with apparent mismatch for functions accepting constant null as input:
```
I0716 15:43:25.460034 1878700 ExpressionFuzzerVerifier.cpp:389] ==============================> Started iteration 218 (seed: 4245448193)
I0716 15:43:25.460829 1878700 ExpressionVerifier.cpp:245] Executing expression 0 : greatest(20,"c0",85,null,"c0")
I0716 15:43:25.628309 1878700 PrestoQueryRunner.cpp:447] Execute presto sql: SELECT greatest(TINYINT '20', c0, TINYINT '85', cast(null as TINYINT), c0) as p0, row_number as p1 FROM (SELECT c0 as c0, row_number as row_number FROM (t_t_values))
fbcode/velox/exec/tests/utils/QueryAssertions.cpp:1186: Failure
Failed
Expected 10, got 10
10 extra rows, 10 missing rows
10 of extra rows:
	85 | 2
	85 | 3
	85 | 5
	85 | 7
	85 | 8
	87 | 1
	87 | 4
	96 | 6
	97 | 9
	124 | 0

10 of missing rows:
	null | 0
	null | 1
	null | 2
	null | 3
	null | 4
	null | 5
	null | 6
	null | 7
	null | 8
	null | 9

Unexpected results
```
A few peculiar things to note:
1. This behavior is particularly strange as expression evaluation should produce nulls and actually does. Unit testing and Prestissimo verification corroborates the Presto output. This discrepancy appears isolated to the fuzzer.
2. Building fuzzer and running locally may induce a failure, but retry on the same seed will not invoke the failure again. It seems highly transitory. 
3. Thirdly, and connected to the above point, the behavior is not consistent even within the same build. Take the following execution just a few iterations before, which matches the Presto output:
```
I0716 15:43:21.594882 1878700 ExpressionFuzzerVerifier.cpp:389] ==============================> Started iteration 207 (seed: 2592566165)
I0716 15:43:21.595443 1878700 ExpressionVerifier.cpp:245] Executing expression 0 : greatest(null,"c0")
I0716 15:43:21.596972 1878700 ExpressionVerifier.cpp:377] Common eval succeeded.
[...]

PR #14191, opened 2025-07-22
fix: ConstantTypedExpr::equals null behavior

peterenescu:
Summary:
Fixes ConstantTypedExpr::equals for constant expressions with null values. Currently, we simply return the value of BaseVector::compareNulls, which operators like a comparator (0 if same 1/-1 otherwise), which returns the opposite value we want (1 or true if same, 0 otherwise).

This will also solve the issues and reduce the flakiness we see in fuzzers because of this error: https://github.com/facebookincubator/velox/issues/14122

Differential Revision: D78771083

### DIFF
--- a/velox/core/Expressions.cpp
+++ b/velox/core/Expressions.cpp
@@ -364,7 +364,7 @@ bool equalsImpl(
 
   if (otherNull || thisNull) {
     return BaseVector::compareNulls(thisNull, otherNull, kEqualValueAtFlags)
-        .value();
+               .value() == 0;
   }
 
   return VELOX_DYNAMIC_TYPE_DISPATCH_ALL(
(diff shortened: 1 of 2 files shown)


kagamiori:
cc @mbasmanova

velox/core/Expressions.cpp
```diff
@@ -364,7 +364,9 @@ bool equalsImpl(
 
   if (otherNull || thisNull) {
     return BaseVector::compareNulls(thisNull, otherNull, kEqualValueAtFlags)
-        .value();
+                .value() == 0
+        ? true
+        : false;
```
kagamiori: nit: Just `return BaseVector::compareNulls(thisNull, otherNull, kEqualValueAtFlags).value() == 0;` should be enough.

kagamiori (approved):
LGTM. Thank you for the fix!

PR #14203, opened 2025-07-23
fix: Fix Spark decimal multiply and divide functions

rui-mo:
The `int256_t` type is needed when computing the intermediate result of large 
input values. To rescale down, the rescale factor can exceed the maximum value 
that `int128_t` can represent. The `DecimalUtil::kPowersOfTen` array is not 
large enough to hold these values, leading to unexpected result. This PR fixes 
this by adding `kLargeScalePowersOfTen`, which is sufficiently large for the 
decimal operation's rescale factor.

Reference: https://github.com/apache/arrow/blob/r-15.0.1/cpp/src/gandiva/decimal_xlarge.cc#L148-L164

### DIFF
--- a/velox/functions/sparksql/DecimalArithmetic.cpp
+++ b/velox/functions/sparksql/DecimalArithmetic.cpp
@@ -427,7 +427,7 @@ struct DecimalMultiplyFunction {
       return in;
     }
 
-    int256_t divisor = velox::DecimalUtil::kPowersOfTen[reduceBy];
+    int256_t divisor = DecimalUtil::getPowersOfTen(reduceBy);
     auto result = in / divisor;
     auto remainder = in % divisor;
     // Round up.
--- a/velox/functions/sparksql/DecimalUtil.h
+++ b/velox/functions/sparksql/DecimalUtil.h
@@ -186,8 +186,7 @@ class DecimalUtil {
         return R(-1);
       }
       int256_t aLarge = a;
-      int256_t aLargeScaledUp =
-          aLarge * velox::DecimalUtil::kPowersOfTen[aRescale];
+      int256_t aLargeScaledUp = aLarge * getPowersOfTen(aRescale);
       int256_t bLarge = b;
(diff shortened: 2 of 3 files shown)


velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -422,12 +422,35 @@ struct DecimalMultiplyFunction {
   }
 
  private:
+  static constexpr int32_t kMaxLargeScale = 2 * LongDecimalType::kMaxPrecision;
+
+  // Pre-compute the powers of ten for large scales. The maximum scale is
+  // 2 * LongDecimalType::kMaxPrecision, which is 76. The
+  // DecimalUtil::kPowersOfTen array is not large enough to hold these values.
+  static constexpr std::array<int256_t, kMaxLargeScale + 1>
```
dharanad: How about we move this to `velox/type/DecimalUtil.h` and replace other instance of `kPowersOfTen` with `kLargeScalePowersOfTen`

rui-mo: Thanks for your suggestion. I had the similar idea, but noticed only several computation needs the int256 type, and int128 was large enough for the rest. Which would you prefer, to replace all with the `kLargeScalePowersOfTen` or only replace those requires int256?

rui-mo: I'll take a look for this replacement.

dharanad: Replacing should suffice here

rui-mo: I discovered that this was the only rescale factor that needed to be of int256 type, thus replacing it alone should be sufficient.

rui-mo:
@jinchengchenghh Would you like to take a review? Thanks.

rui-mo:
@jinchengchenghh Would you please mark as ready-to-merge if no further comment? Thanks.

PR #14206, opened 2025-07-23
fix(fuzzer): Skip xxhash64 from Presto SOT

pedroerp:
Summary:
A new signature was recently added to xxhash64 in Velox, but not
supported in Presto Java. This makes Presto SOT fuzzer fail. Since we cannot
skip a specific signature, skipping the entire function for now until the new
signture is merged and released in Presto java:
.
https://github.com/prestodb/presto/pull/25521

Reviewed By: peterenescu

Differential Revision: D78830349


peterenescu (approved):

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -189,6 +189,9 @@ int main(int argc, char** argv) {
       "st_xmin",
       "st_ymax",
       "st_ymin",
+      // Skipping until the new signature is merged and released in Presto:
+      // https://github.com/prestodb/presto/pull/25521
+      "xxhash64",
```
kevinwilfong: Should we move it to the list that's only excluded when Presto is used as a source of truth (see line 242)?

PR #14211, opened 2025-07-24
build: Add a CMake linter hook to pre-commit

assignUser:
This PR introduces [ast-grep](https://ast-grep.github.io/) linting for CMake files to enforce best practices. This hook uses custom cmake language support that requires the compilation of a tree-sitter parser written in C. The hook handles this platform independent.

I created comprehensive rules for common CMake errors and best practices based on the style we have been enforcing in reviews. I will write up a proper styleguide as well (TODO). The rules come with test files to verify each linting rule's functionality.

I fixed some long standing issues and also set the matching rule to `error`  to avoid re-introduction of these patterns.

### Variable-to-Target Conversions
- Replaced variable-based dependency lists with proper target interfaces
- Converted `${DEPENDENCY_LIBS}` variables to proper targets in multiple locations
- Improved build isolation and dependency clarity

### Build Isolation Improvements
- Wrapped FetchContent dependency declarations in `block()` to isolate variable changes
- Prevents dependency-specific settings from leaking into the main build
- Makes builds more reproducible and prevents cross-contamination (see recent issue with CPR)

There is a rule to address implicit PUBLIC linking but that is set to warning for now. It even comes with an automatic fix but eh... ` 177 files changed, 2228 insertions(+), 2884 deletions(-)`

### DIFF
--- a/velox/exec/fuzzer/CMakeLists.txt
+++ b/velox/exec/fuzzer/CMakeLists.txt
@@ -28,21 +28,20 @@ add_library(
 
 target_link_libraries(
   velox_fuzzer_util
-  PRIVATE
-    velox_common_fuzzer_util
-    velox_vector_fuzzer
-    velox_core
-    velox_exec_test_lib
-    velox_expression_functions
-    velox_presto_types
-    cpr::cpr
-    Boost::regex
-    velox_type_parser
-    Folly::folly
-    velox_hive_connector
-    velox_dwio_dwrf_writer
-    velox_dwio_catalog_fbhive
-    velox_dwio_faulty_file_sink)
+  PUBLIC velox_common_fuzzer_util
+         velox_vector_fuzzer
+         velox_core
+         velox_exec_test_lib
+         velox_expression_functions
+         velox_presto_types
+         cpr::cpr
+         Boost::regex
+         velox_type_parser
+         Folly::folly
+         velox_hive_connector
+         velox_dwio_dwrf_writer
+         velox_dwio_catalog_fbhive
+         velox_dwio_faulty_file_sink)
 
 add_library(velox_aggregation_fuzzer_base AggregationFuzzerBase.cpp)
 


CMake/resolve_dependency_modules/geos.cmake
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 include_guard(GLOBAL)
 
-block() # This creates a separate scope so any changed variables don't affect
-        # the rest of the build.
+block(SCOPE_FOR VARIABLES) # This creates a separate scope so any changed
+                           # variables don't affect
+# the rest of the build.
```
kou, line 18:
```suggestion
# This creates a separate scope so any changed
# variables don't affect the rest of the build.
block(SCOPE_FOR VARIABLES)
```

CMake/resolve_dependency_modules/gflags.cmake
```diff
@@ -43,19 +44,8 @@ set(GFLAGS_IS_SUBPROJECT ON)
 
 # Workaround for https://github.com/gflags/gflags/issues/277
 unset(BUILD_SHARED_LIBS)
-if(DEFINED CACHE{BUILD_SHARED_LIBS})
-  set(CACHED_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
-  unset(BUILD_SHARED_LIBS CACHE)
-endif()
```
kou, line 49: I think that cache variables can't be guarded by `block(SCOPE_FOR VARIABLES)`.
Is it safe that we remove this?

scripts/ci/cmake-rules-test/__snapshots__/cmake-minimum-version-error-snapshot.yml
```diff
@@ -0,0 +1,105 @@
...
+snapshots:
+  ? |-
+    # Multiple version specifications (first one triggers)
+    cmake_minimum_required(VERSION 3.5)
+    cmake_minimum_required(VERSION 3.15)
+  : labels:
+      - source: cmake_minimum_required(VERSION 3.5)
+        style: primary
+        start: 55
+        end: 90
```
kou, line 24: Duplicated test?

kou (commented):
I understand your intention that we prevent deprecated CMake config (enforce recommended CMake config) automatically.

But it may be difficult for us to maintain raw ast-grep configurations by ourselves.

How about creating a standalone CMake lint tool that is a wrapper of ast-grep instead? The tool provides more simpler configurations than raw ast-grep configurations. We will use the simpler configurations. For example, making `velox_link_libraries()` as an alias of `target_link_libraries()`.

(Bonus: If we create a standalone CMake lint tool, we can use it in multiple projects such as Velox and Apache Arrow.)

Can we maintain the standalone CMake lint tool by multiple people for sustainability? For example, you can I can maintain it as the initial maintainers.

PR #14215, opened 2025-07-24
feat: Rewrite array flatten with vector function and zero copy optimization

xiaoxmeng:
Summary: Rewrite flatten array function with vector function and zero copy optimization. For Meta internal AI use case with sub-query authoring, this reduces the peak memory of flatten array udf from 218MB to 13MB, cpu cost from 13.34s to 1.54s.

Differential Revision: D78875977

### DIFF
--- a/velox/functions/prestosql/ArrayFlatten.cpp
+++ b/velox/functions/prestosql/ArrayFlatten.cpp
@@ -1,0 +1,242 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/prestosql/ArrayFunctions.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+class ArrayFlattenFunction : public exec::VectorFunction {
+ public:
+  static std::vector<exec::FunctionSignaturePtr> signatures() {
+    // array(array(T)) -> array(T)
+    return {exec::FunctionSignatureBuilder()
+                .returnType("array(T)")
+                .argumentType("array(array(T))")
+                .typeVariable("T")
+                .build()};
+  }
+
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& outputType,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    VELOX_CHECK_EQ(args.size(), 1);
+    const auto& arrayVector = args[0];
+    VELOX_CHECK_EQ(arrayVector->encoding(), VectorEncoding::Simple::ARRAY);
+    const auto* arrayVectorPtr = arrayVector->as<ArrayVector>();
+    auto state =
+        ProcessState::create(outputType, rows.end(), arrayVectorPtr, context);
+    rows.applyToSelected(
+        [&](vector_size_t row) { processOneInputArray(row, row, state); });
+    VectorPtr localResult;
+    if (state.flattenElementsConsecutive) {
+      localResult = flattenArray(state);
+    } else {
+      localResult = flattenArrayWithDictionary(rows, state);
+    }
+    context.moveOrCopyResult(localResult, rows, result);
+  }
+
+ private:
+  struct ProcessState {
+    TypePtr resultType;
+    vector_size_t numRows{0};
+
+    const vector_size_t* arrayOffsets{};
+    const vector_size_t* arraySizes{};
+
+    std::unique_ptr<exec::LocalSelectivityVector> innerArrayRowSelector;
+    std::unique_ptr<exec::LocalDecodedVector> decodedInnerArrayHolder;
+    const DecodedVector* decodedInnerArray{nullptr};
+    VectorPtr elements{nullptr};
+    const vector_size_t* elementOffsets{nullptr};
+    const vector_size_t* elementSizes{nullptr};
+
+    BufferPtr flattenArrayOffsets{nullptr};
+    vector_size_t* rawFlattenArrayOffsets{nullptr};
+    BufferPtr flattenArraySizes{nullptr};
+    vector_size_t* rawFlattenArraySizes{nullptr};
+
+    // Indicates if the flatten elements are consecutive in source base element
+    // vector. If so, we can directly reference the source element vector.
+    // Otherwise, we need to build a dictionary mapping on top of the source
+    // element vector.
+    bool flattenElementsConsecutive{true};
+    vector_size_t numFlattenElements{0};
+    BufferPtr flattenElementIndices{nullptr};
+    vector_size_t* rawFlattenElementIndices{nullptr};
+    VectorPtr flattenElementsWithDictionary{nullptr};
+
+    memory::MemoryPool* pool{nullptr};
+
+    std::pair<vector_size_t, vector_size_t> elementOffsetAndSize(
+        vector_size_t elementIndex) const {
+      return {elementOffsets[elementIndex], elementSizes[elementIndex]};
+    }
+
+    // @param arrayVector ARRAY(ARRAY(E)) vector.
+    static ProcessState create(
+        const TypePtr& resultType,
+        size_t numRows,
+        const ArrayVector* arrayVector,
+        exec::EvalCtx& context) {
+      VELOX_CHECK_NOT_NULL(arrayVector);
+      ProcessState state;
+      state.resultType = resultType;
+      state.numRows = numRows;
+      state.pool = context.pool();
+
+      state.arrayOffsets = arrayVector->rawOffsets();
+      state.arraySizes = arrayVector->rawSizes();
+
+      const auto innerArrayVector = arrayVector->elements();
+      state.innerArrayRowSelector =
+          std::make_unique<exec::LocalSelectivityVector>(
+              context, innerArrayVector->size());
+      state.innerArrayRowSelector->get()->setAll();
+      state.decodedInnerArrayHolder =
+          std::make_unique<exec::LocalDecodedVector>(
+              context, *innerArrayVector, *state.innerArrayRowSelector->get());
+      state.decodedInnerArray = state.decodedInnerArrayHolder->get();
+
+      auto* innerArrayVectorPtr =
+          state.decodedInnerArray->base()->as<ArrayVector>();
+      state.elements = innerArrayVectorPtr->elements();
+      state.elementOffsets = innerArrayVectorPtr->rawOffsets();
+      state.elementSizes = innerArrayVectorPtr->rawSizes();
+
+      state.flattenArrayOffsets = allocateIndices(state.numRows, state.pool);
+      state.rawFlattenArrayOffsets =
+          state.flattenArrayOffsets->asMutable<velox::vector_size_t>();
+
+      state.flattenArraySizes = allocateIndices(state.numRows, state.pool);
+      state.rawFlattenArraySizes =
+          state.flattenArraySizes->asMutable<velox::vector_size_t>();
+      return state;
+    }
+  };
+
+  VectorPtr flattenArrayWithDictionary(
+      const SelectivityVector& rows,
+      ProcessState& state) const {
+    createFlattenElementIndices(state);
+    vector_size_t flattenElementOffset{0};
+    rows.applyToSelected([&](vector_size_t row) {
+      processOneOutputRow(row, row, state, flattenElementOffset);
+    });
+    return std::make_shared<velox::ArrayVector>(
+        state.pool,
+        state.resultType,
+        nullptr,
+        state.numRows,
+        state.flattenArrayOffsets,
+        state.flattenArraySizes,
+        state.flattenElementsWithDictionary);
+  }
+
+  static void createFlattenElementIndices(ProcessState& state) {
+    VELOX_CHECK(!state.flattenElementsConsecutive);
+    VELOX_CHECK_NULL(state.flattenElementIndices);
+    VELOX_CHECK_NULL(state.flattenElementsWithDictionary);
+    state.flattenElementIndices =
+        allocateIndices(state.numFlattenElements, state.pool);
+    state.rawFlattenElementIndices =
+        state.flattenElementIndices->asMutable<vector_size_t>();
+    state.flattenElementsWithDictionary = BaseVector::wrapInDictionary(
+        nullptr,
+        state.flattenElementIndices,
+        state.numFlattenElements,
+        state.elements);
+  }
+
+  VectorPtr flattenArray(const ProcessState& state) const {
+    return std::make_shared<velox::ArrayVector>(
+        state.pool,
+        state.resultType,
+        nullptr,
+        state.numRows,
+        state.flattenArrayOffsets,
+        state.flattenArraySizes,
+        state.elements);
+  }
+
+  // @param inputRow The row index in the input ARRAY(ARRAY(E)) vector.
+  // @param outputRow The row index in the output ARRAY(E) vector.
+  void processOneInputArray(
+      vector_size_t inputRow,
+      vector_size_t outputRow,
+      ProcessState& state) const {
+    vector_size_t nextElementOffset{-1};
+    for (auto i = 0, offset = state.arrayOffsets[inputRow];
+         i < state.arraySizes[inputRow];
+         ++i, ++offset) {
+      if (FOLLY_UNLIKELY(state.decodedInnerArray->isNullAt(offset))) {
+        continue;
+      }
+      const auto elementArrayIndex = state.decodedInnerArray->index(offset);
+      const auto [elementOffset, numElements] =
+          state.elementOffsetAndSize(elementArrayIndex);
+      state.numFlattenElements += numElements;
+      state.rawFlattenArraySizes[outputRow] += numElements;
+      if (nextElementOffset == -1) {
+        state.rawFlattenArrayOffsets[outputRow] = elementOffset;
+      } else if (nextElementOffset != elementOffset) {
+        state.flattenElementsConsecutive = false;
+      }
+      nextElementOffset = elementOffset + numElements;
+    }
+  }
+
+  void processOneOutputRow(
+      vector_size_t inputRow,
+      vector_size_t outputRow,
+      const ProcessState& state,
+      vector_size_t& flattenElementOffset) const {
+    state.rawFlattenArrayOffsets[outputRow] = flattenElementOffset;
+    vector_size_t numFlattenElements{0};
+    for (auto i = 0, elementArrayOffset = state.arrayOffsets[inputRow];
+         i < state.arraySizes[inputRow];
+         ++i, ++elementArrayOffset) {
+      if (FOLLY_UNLIKELY(
+              state.decodedInnerArray->isNullAt(elementArrayOffset))) {
+        continue;
+      }
+      const auto elementArrayIndex =
+          state.decodedInnerArray->index(elementArrayOffset);
+      const auto [elementOffset, numElements] =
+          state.elementOffsetAndSize(elementArrayIndex);
+      for (auto elementIndex = elementOffset;
+           elementIndex < elementOffset + numElements;
+           ++elementIndex) {
+        state.rawFlattenElementIndices[flattenElementOffset++] = elementIndex;
+      }
+      numFlattenElements += numElements;
+    }
+    VELOX_CHECK_EQ(numFlattenElements, state.rawFlattenArraySizes[outputRow]);
+  }
+};
+
+} // namespace
+
+VELOX_DECLARE_VECTOR_FUNCTION(
+    udf_array_flatten,
+    ArrayFlattenFunction::signatures(),
+    std::make_unique<ArrayFlattenFunction>());
+
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/tests/ArrayFlattenTest.cpp
+++ b/velox/functions/prestosql/tests/ArrayFlattenTest.cpp
@@ -36,8 +36,8 @@ class ArrayFlattenTest : public FunctionBaseTest {
 
 /// Flatten integer arrays.
 TEST_F(ArrayFlattenTest, intArrays) {
-  const auto baseVector = makeArrayVector<int64_t>(
-      {{1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}, {6, 6}});
+  const auto baseVector = makeArrayVectorFromJson<int64_t>(
+      {"[1, 1]", "[2,2]", "[3, 3]", "[4,4]", "[5, 5]", "[6, 6]"});
 
   // Create arrays of array vector using above base vector.
   // [[1, 1], [2, 2], [3, 3]]
@@ -55,7 +55,7 @@ TEST_F(ArrayFlattenTest, intArrays) {
 }
 
 /// Flatten arrays with null.
-TEST_F(ArrayFlattenTest, nullArray) {
+TEST_F(ArrayFlattenTest, nullArrayWithConsecutiveBaseElements) {
   const auto baseVector = makeNullableArrayVector<int64_t>(
       {{{1, 1}},
        std::nullopt,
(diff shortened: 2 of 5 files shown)


velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& outputType,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    VELOX_CHECK_EQ(args.size(), 1);
+    const auto& arrayVector = args[0];
+    VELOX_CHECK(arrayVector->type()->isArray());
```
mbasmanova: These checks are not needed as they can be enforced generically based on signature.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+    velox::BaseVector::ensureWritable(rows, outputType, context.pool(), result);
```
mbasmanova: Consider instead to use context.moveOrCopy

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    exec::LocalDecodedVector decodedArray(context, *arrayVector, rows);
```
mbasmanova: single-arg deterministic functions receive their inputs as flat or constant, hence, no need to use decoded vector; also, FYI, there is DecodedArgs utility that helps with decoding

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+        }
+        const auto elementArrayIndex =
+            decodedElementArray->index(elementArrayOffset);
+        const auto elementOffset = elementOffsets[elementArrayIndex];
+        const auto numElements = elementSizes[elementArrayIndex];
+        numFlattenElements += numElements;
+        if (nextElementOffset == -1) {
+          state.rawFlattenArrayOffsets[row] = elementOffset;
+        } else if (nextElementOffset != elementOffset) {
+          needCopy = true;
```
mbasmanova: I believe the can always make this zero-copy by wrapping inner vector in a dictionary.

xiaoxmeng: > I believe the can always make this zero-copy by wrapping inner vector in a dictionary.

How we guarantee that all the concatenated inner arrays from one flatten array are always consecutive in base element array? There is no restriction on that?

velox/functions/prestosql/ArrayFlatten.cpp
xiaoxmeng: You mean do this at the end with local computed result?

xiaoxmeng: Updated.

velox/functions/prestosql/ArrayFlatten.cpp
xiaoxmeng: So the expression execution framework already does the input flattening?

mbasmanova: That's right.

fbcode/velox/expression/VectorFunction.h
```
  /// Single-argument deterministic functions may receive their only argument
  /// vector as flat or constant, but not dictionary encoded.
  ///
  /// Single-argument functions that specify null-in-null-out behavior, e.g.
  /// defaultNullBehavior is true, will never see a null row in 'rows'.
  /// Hence, they can safely assume that args[0] vector is flat or constant and
  /// has no nulls in specified positions.
```

velox/functions/prestosql/ArrayFlatten.cpp
mbasmanova: By wrapping the innermost element vector in a dictionary.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+#include "velox/functions/prestosql/ArrayFunctions.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+class ArrayFlattenFunction : public exec::VectorFunction {
+ public:
+  static std::vector<exec::FunctionSignaturePtr> signatures() {
+    // Define the function signature: array(array(T)) -> array(T)
```
mbasmanova: Drop "Define the function signature:" prefix. It is redundant.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+  };
+
+  VectorPtr flattenFlatArray(
+      const SelectivityVector& rows,
+      const ArrayVector* arrayVector,
+      const TypePtr& outputType,
+      exec::EvalCtx& context) const {
+    VELOX_CHECK_NOT_NULL(arrayVector);
+
+    ProcessState state = getProcessState(
```
mbasmanova: consider moving getProcessState to a static method in ProcessState

auto state = ProcessState::create(...)

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+        outputType,
+        rows.end(),
+        arrayVector,
+        arrayVector->mayHaveNulls(),
+        context);
+    rows.applyToSelected([&](vector_size_t row) {
+      if (FOLLY_UNLIKELY(arrayVector->isNullAt(row))) {
```
mbasmanova: This can never be true. This check is not needed. See documentation in fbcode/velox/expression/VectorFunction.h

```
  /// Single-argument functions that specify null-in-null-out behavior, e.g.
  /// defaultNullBehavior is true, will never see a null row in 'rows'.
  /// Hence, they can safely assume that args[0] vector is flat or constant and
  /// has no nulls in specified positions.
```

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+        state.flattenArrayHasNulls = true;
+        velox::bits::setNull(state.rawFlattenArrayNulls, row, true);
+        state.rawFlattenArraySizes[row] = 0;
+        return;
+      }
+      processOneInputArray(row, row, state);
+    });
+
+    if (state.needCopyElements) {
+      // TODO: avoid the actual element copy with dicitonary mapping.
```
mbasmanova: typo: dicitonary -> dictionary

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+  VectorPtr flattenConstantArray(
+      const SelectivityVector& rows,
+      const VectorPtr& constantVector,
+      const TypePtr& outputType,
+      exec::EvalCtx& context) const {
+    VELOX_CHECK(constantVector->isConstantEncoding());
+    auto* constArrayVector = constantVector->as<ConstantVector<ComplexType>>();
+    VELOX_CHECK_NOT_NULL(constArrayVector);
+    if (constArrayVector->isNullAt(0)) {
```
mbasmanova: This cannot happen. Hence, this check is not needed.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+        outputType,
+        /*numRows=*/1,
+        arrayVector,
+        /*arrayHasNull=*/false,
+        context);
+    processOneInputArray(arrayIndex, 0, state);
+
+    VectorPtr result;
+    if (state.needCopyElements) {
+      // TODO: avoid the actual element copy with dicitonary mapping.
```
mbasmanova: dicitonary -> dictionary

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+    return BaseVector::wrapInConstant(rows.end(), 0, result);
+  }
+
+  ProcessState getProcessState(
+      const TypePtr& resultType,
+      size_t numRows,
+      const ArrayVector* arrayVector,
+      bool arrayHasNull,
+      exec::EvalCtx& context) const {
+    VELOX_CHECK_NOT_NULL(arrayVector);
```
mbasmanova: The caller already checked this. Let's pass arrayVector by reference so we don't need this check.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+    state.elementArrayRowSelector->get()->setAll();
+    state.decodedElementArrayHolder =
+        std::make_unique<exec::LocalDecodedVector>(
+            context,
+            *elementArrayVector,
+            *state.elementArrayRowSelector->get());
+    state.decodedElementArray = state.decodedElementArrayHolder->get();
+
+    state.elements =
+        state.decodedElementArray->base()->as<ArrayVector>()->elements();
```
mbasmanova: > state.decodedElementArray->base()->as<ArrayVector>()
 
 let's make a temp variable to hold this to avoid multiple casts and improve readability

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+        state.decodedElementArray->base()->as<ArrayVector>()->rawOffsets();
+    state.elementSizes =
+        state.decodedElementArray->base()->as<ArrayVector>()->rawSizes();
+
+    const auto flattenArrayIndicesBytes =
+        state.numRows * sizeof(velox::vector_size_t);
+    state.flattenArrayOffsets = allocateIndices(state.numRows, state.pool);
+    state.rawFlattenArrayOffsets =
+        state.flattenArrayOffsets->asMutable<velox::vector_size_t>();
+    state.flattenArrayOffsets->setSize(flattenArrayIndicesBytes);
```
mbasmanova: this is not needed; allocateIndices takes care of that

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+    std::memset(state.rawFlattenArrayOffsets, 0, flattenArrayIndicesBytes);
```
mbasmanova: also not needed

```
/// Allocates a buffer to fit at least 'size' indices and initializes them to
/// zero.
inline BufferPtr allocateIndices(vector_size_t size, memory::MemoryPool* pool) {
  return AlignedBuffer::allocate<vector_size_t>(size, pool, 0);
}
```

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+    state.flattenArraySizes = allocateIndices(state.numRows, state.pool);
+    state.rawFlattenArraySizes =
+        state.flattenArraySizes->asMutable<velox::vector_size_t>();
+    state.flattenArraySizes->setSize(flattenArrayIndicesBytes);
+    std::memset(state.rawFlattenArraySizes, 0, flattenArrayIndicesBytes);
```
mbasmanova: not needed

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+      return state;
+    }
+
+    const auto nullsBytes = velox::bits::nbytes(state.numRows);
+    state.flattenArrayNulls = velox::allocateNulls(state.numRows, state.pool);
+    state.rawFlattenArrayNulls = state.flattenArrayNulls->asMutable<uint64_t>();
+    bits::fillBits(
+        state.rawFlattenArrayNulls, 0, state.numRows, velox::bits::kNotNull);
+    state.flattenArrayNulls->setSize(nullsBytes);
+    return state;
```
mbasmanova: I believe this logic is not need. Do you see it triggering in tests?

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,340 @@
...
+      if (FOLLY_UNLIKELY(
+              state.flattenElementHasNulls &&
+              state.decodedElementArray->isNullAt(elementArrayOffset))) {
+        continue;
+      }
+      const auto elementArrayIndex =
+          state.decodedElementArray->index(elementArrayOffset);
+      const auto elementOffset = state.elementOffsets[elementArrayIndex];
+      const auto numElements = state.elementSizes[elementArrayIndex];
+      flattenElements->copy(
```
mbasmanova: Copying is very expensive. Can we wrap in dictionary? If not, can we at least copy all ranges at once?

```
  /// Copy multiple ranges at once.  This is more efficient than calling `copy`
  /// multiple times, especially for ARRAY, MAP, and VARCHAR.
  virtual void copyRanges(
      const BaseVector* /*source*/,
      const folly::Range<const CopyRange*>& /*ranges*/) {
```

velox/functions/prestosql/tests/ArrayFlattenTest.cpp
```diff
@@ -77,4 +77,169 @@ TEST_F(ArrayFlattenTest, nullArray) {
...
+TEST_F(ArrayFlattenTest, nullArrayWithCopy) {
+  const auto baseVector = makeNullableArrayVector<int64_t>(
+      {{{1, 1}},
+       std::nullopt,
+       {{3, 3}},
+       {{}},
+       {{5, std::nullopt}},
+       {{std::nullopt, 6}}});
+
+  // Create arrays of array vector using above base vector.
```
mbasmanova, line 85: Use makeNestedArrayVectorFromJson

velox/functions/prestosql/tests/ArrayFlattenTest.cpp
```diff
@@ -77,4 +77,169 @@ TEST_F(ArrayFlattenTest, nullArray) {
...
+          {std::nullopt, 6, 5, std::nullopt}});
+  for (bool enabled : {false, true}) {
+    SCOPED_TRACE(fmt::format("peeling encoding: {}", enabled));
+    // Enforce the constant encoding flatten path to trigger.
+    execCtx_.testingPeelingEnabled(enabled);
+    testExpression("flatten(c0)", {constArrayOfArrays}, expected);
+  }
+}
+
+TEST_F(ArrayFlattenTest, fuzz) {
```
mbasmanova: Do we need this? Let's just run fuzzer biased towards this function.

mbasmanova (changes requested):
@xiaoxmeng Thank you for iterating. Have some further comments.

velox/core/QueryCtx.h
```diff
@@ -390,6 +390,12 @@ class ExecCtx {
     return optimizationParams_;
   }
 
+  void testingPeelingEnabled(bool enabled) {
```
mbasmanova: Do we need this? Why can't we just specify the desired config? CC: @bikramSingh91

velox/functions/prestosql/ArrayFlatten.cpp
xiaoxmeng: Good to know.

velox/core/QueryCtx.h
xiaoxmeng: ExecCtx used by the test facility is a member of base class FunctionBaseTest.h

velox/functions/prestosql/tests/ArrayFlattenTest.cpp
xiaoxmeng: I feel it might be good to have a dedicated fuzz test and catch some bug during the development.

velox/functions/prestosql/ArrayFlatten.cpp
xiaoxmeng: I tested and args[0] may have nulls returns true.

velox/functions/prestosql/ArrayFlatten.cpp
xiaoxmeng: At least args[0] may have nulls returns true from unit test

velox/functions/prestosql/tests/ArrayFlattenTest.cpp
mbasmanova: My concern is about maintenance for such tests. 

One can run fuzzer for just this function via "--only flatten" during development to catch bugs. Would you try this to see if that covers the need.

xiaoxmeng: Removed.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,303 @@
...
+        state.flattenArraySizes,
+        state.elements);
+  }
+
+  FOLLY_ALWAYS_INLINE void processOneInputArray(
+      vector_size_t inputRow,
+      vector_size_t outputRow,
+      ProcessState& state) const {
+    vector_size_t nextElementOffset{-1};
+    for (auto i = 0, elementArrayOffset = state.arrayOffsets[inputRow];
```
mbasmanova: nit: elementArrayOffset -> offset for readability

velox/functions/prestosql/tests/ArrayFlattenTest.cpp
```diff
@@ -77,4 +77,93 @@ TEST_F(ArrayFlattenTest, nullArray) {
...
   testExpression("flatten(c0)", {arrayOfArrays}, expected);
 }
+
+TEST_F(ArrayFlattenTest, nullArrayWithoutConsecutiveBaseElements) {
+  const auto baseVector = makeArrayVectorFromJson<int64_t>(
+      {"[1, 1]", "null", "[3, 3]", "[]", "[5, null]", "[null, 6]"});
+
+  // Create arrays of array vector using above base vector.
+  const auto numElements = 6;
+  auto reversedIndices = allocateIndices(numElements, pool());
```
mbasmanova: use makeIndicesInReverse

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,254 @@
...
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& outputType,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    VELOX_CHECK_EQ(args.size(), 1);
+    const auto& arrayVector = args[0];
+    VELOX_CHECK_EQ(arrayVector->encoding(), VectorEncoding::Simple::ARRAY);
+    VectorPtr localResult = flattenFlatArray(
```
mbasmanova: We no longer need flattenFlatArray method, right?

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,254 @@
...
+    BufferPtr flattenElementIndices{};
+    vector_size_t* rawFlattenElementIndices{};
+    VectorPtr flattenElementsWithDictionary{};
+
+    memory::MemoryPool* pool{nullptr};
+
+    static ProcessState create(
+        const TypePtr& resultType,
+        size_t numRows,
+        const ArrayVector* arrayVector,
```
mbasmanova, line 97: I find it difficult to follow the code because it is not clear what is array(array()) and what is the inner array() and what are the elements. Wondering if some comments could help.

perhaps, 

// @param arrayVector ARRAY(ARRAY(E)) vector.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,254 @@
...
+        exec::EvalCtx& context) {
+      VELOX_CHECK_NOT_NULL(arrayVector);
+      ProcessState state;
+      state.resultType = resultType;
```
mbasmanova, line 101: // ARRAY(ARRAY(E))

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,254 @@
...
+      state.numRows = numRows;
+      state.pool = context.pool();
+
+      state.arrayOffsets = arrayVector->rawOffsets();
+      state.arraySizes = arrayVector->rawSizes();
+
+      const auto elementArrayVector = arrayVector->elements();
```
mbasmanova: elementArrayVector is confusing to me; I assume this is the inner array vector. 

Would it be helpful to rename to innerArrayVector?

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,254 @@
...
+      state.elementArrayRowSelector->get()->setAll();
+      state.decodedElementArrayHolder =
+          std::make_unique<exec::LocalDecodedVector>(
+              context,
+              *elementArrayVector,
+              *state.elementArrayRowSelector->get());
+      state.decodedElementArray = state.decodedElementArrayHolder->get();
+
+      state.elements =
+          state.decodedElementArray->base()->as<ArrayVector>()->elements();
```
mbasmanova: Can we add a temp variable to hold state.decodedElementArray->base()->as<ArrayVector>() to avoid copy-paste?

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+      state.elementOffsets =
+          state.decodedElementArray->base()->as<ArrayVector>()->rawOffsets();
+      state.elementSizes =
+          state.decodedElementArray->base()->as<ArrayVector>()->rawSizes();
+
+      const auto flattenArrayIndicesBytes =
```
mbasmanova: This variable and setSize() call are not needed. Just,

```
      state.flattenArrayOffsets = allocateIndices(state.numRows, state.pool);
      state.rawFlattenArrayOffsets =
          state.flattenArrayOffsets->asMutable<velox::vector_size_t>();
```

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+          state.numRows * sizeof(velox::vector_size_t);
+      state.flattenArrayOffsets = allocateIndices(state.numRows, state.pool);
+      state.rawFlattenArrayOffsets =
+          state.flattenArrayOffsets->asMutable<velox::vector_size_t>();
+      state.flattenArrayOffsets->setSize(flattenArrayIndicesBytes);
+
+      state.flattenArraySizes = allocateIndices(state.numRows, state.pool);
+      state.rawFlattenArraySizes =
+          state.flattenArraySizes->asMutable<velox::vector_size_t>();
+      state.flattenArraySizes->setSize(flattenArrayIndicesBytes);
```
mbasmanova: ditto

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,242 @@
...
+    TypePtr resultType;
+    vector_size_t numRows{0};
+
+    const vector_size_t* arrayOffsets{};
+    const vector_size_t* arraySizes{};
+
+    std::unique_ptr<exec::LocalSelectivityVector> elementArrayRowSelector;
+    std::unique_ptr<exec::LocalDecodedVector> decodedElementArrayHolder;
+    const DecodedVector* decodedElementArray{};
+    VectorPtr elements{};
```
mbasmanova: I find {} confusing. It is not needed for shared_ptrs. For raw pointers, I feel it is better to explicitly initialize to nullptr

```
const DecodedVector* decodedElementArray{nullptr};
```

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,238 @@
...
+    VELOX_CHECK_EQ(args.size(), 1);
+    const auto& arrayVector = args[0];
+    VELOX_CHECK_EQ(arrayVector->encoding(), VectorEncoding::Simple::ARRAY);
+    const auto* arrayVectorPtr = arrayVector->as<ArrayVector>();
+    auto state =
+        ProcessState::create(outputType, rows.end(), arrayVectorPtr, context);
+    rows.applyToSelected(
+        [&](vector_size_t row) { processOneInputArray(row, row, state); });
+    VectorPtr localResult;
+    if (!state.flattenElementsConsecutive) {
```
mbasmanova: nit: it is easier to read if the check is "positive"

```
if (state.flattenElementsConsecutive) {
   ...
```

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,238 @@
...
+        state.resultType,
+        nullptr,
+        state.numRows,
+        state.flattenArrayOffsets,
+        state.flattenArraySizes,
+        state.elements);
+  }
+
+  void processOneInputArray(
+      vector_size_t inputRow,
```
mbasmanova, line 182: inputRow points to an array-of-arrays, right?

outputRow points to a corresponding flat array

Would be nice to clarify in the comments.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+  static void createFlattenElementIndices(ProcessState& state) {
+    VELOX_CHECK(!state.flattenElementsConsecutive);
+    VELOX_CHECK_NULL(state.flattenElementIndices);
+    VELOX_CHECK_NULL(state.flattenElementsWithDictionary);
+    state.flattenElementIndices =
+        allocateIndices(state.numFlattenElements, state.pool);
+    state.rawFlattenElementIndices =
+        state.flattenElementIndices->asMutable<vector_size_t>();
+    state.flattenElementIndices->setSize(
```
mbasmanova: setSize is not needed

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,238 @@
...
+         i < state.arraySizes[inputRow];
+         ++i, ++elementArrayOffset) {
+      if (FOLLY_UNLIKELY(
+              state.decodedElementArray->isNullAt(elementArrayOffset))) {
+        continue;
+      }
+      const auto elementArrayIndex =
+          state.decodedElementArray->index(elementArrayOffset);
+      const auto elementOffset = state.elementOffsets[elementArrayIndex];
+      const auto numElements = state.elementSizes[elementArrayIndex];
```
mbasmanova: This code seems to be repeated twice. Wondering if it can be replaced with a method call on state.

```
const auto {elementOffset, numElements} = state.elementOffsetAndSize(elementArrayOffset);
```

This would return {0, 0} if array is null.

velox/functions/prestosql/ArrayFlatten.cpp
```diff
@@ -0,0 +1,238 @@
...
+      for (auto i = 0, elementIndex = elementOffset; i < numElements;
```
mbasmanova: std::iota

mbasmanova (approved):
@xiaoxmeng Meng, thank you for iterating and implementing zero-copy in all cases. Looks great % a few optional nits.

PR #14224, opened 2025-07-25
feat: Make AssignUniqueId operator support barriered execution.

littlesamo:
Summary:
1. Make AssignUniqueId operator support barriered execution.
2. Add option to Cursor to optionally ensure loading vector. Currently query would fail execution via AssertQueryBuilder if only identify projection exists for multi split execution, since the vectors a lazy vectors and the access can be after file is closed. 

Differential Revision: D78876063

### DIFF
--- a/velox/exec/tests/AssignUniqueIdTest.cpp
+++ b/velox/exec/tests/AssignUniqueIdTest.cpp
@@ -16,31 +16,51 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/exec/PlanNodeStats.h"
 #include "velox/exec/tests/utils/AssertQueryBuilder.h"
-#include "velox/exec/tests/utils/OperatorTestBase.h"
+#include "velox/exec/tests/utils/HiveConnectorTestBase.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/exec/tests/utils/QueryAssertions.h"
 
-using namespace facebook::velox;
+namespace facebook::velox::exec {
+
 using namespace facebook::velox::test;
-using namespace facebook::velox::exec;
 using namespace facebook::velox::exec::test;
 
-class AssignUniqueIdTest : public OperatorTestBase {
+namespace {
+
+class AssignUniqueIdTest : public HiveConnectorTestBase {
  protected:
+  void SetUp() override {
+    HiveConnectorTestBase::SetUp();
+  }
+
   void verifyUniqueId(
       const std::shared_ptr<const core::PlanNode>& plan,
       const std::vector<RowVectorPtr>& input) {
     CursorParameters params;
     params.planNode = plan;
-
     auto result = readCursor(params);
-    auto numColumns = result.second[0]->childrenSize();
+    ASSERT_EQ(result.second[0]->childrenSize(), input[0]->childrenSize() + 1);
+    verifyUniqueId(input, result.second);
+
+    auto task = result.first->task();
+    // Verify number of memory allocations. There should be exactly one
+    // allocation (per thread of execution) for the values buffer of the
+    // unique ID vector. Memory should be allocated when producing first
+    // batch of output and re-used for subsequent batches.
+    auto stats = toPlanStats(task->taskStats());
+    ASSERT_EQ(1, stats.at(uniqueNodeId_).numMemoryAllocations);
+  }
+
+  void verifyUniqueId(
+      const std::vector<RowVectorPtr>& input,
+      const std::vector<facebook::velox::RowVectorPtr>& vectors) {
+    auto numColumns = vectors[0]->childrenSize();
     ASSERT_EQ(numColumns, input[0]->childrenSize() + 1);
 
     std::set<int64_t> ids;
     for (int i = 0; i < numColumns; i++) {
-      for (auto batch = 0; batch < result.second.size(); ++batch) {
-        auto column = result.second[batch]->childAt(i);
+      for (auto batch = 0; batch < vectors.size(); ++batch) {
+        auto column = vectors[batch]->childAt(i);
         if (i < numColumns - 1) {
           assertEqualVectors(input[batch]->childAt(i), column);
         } else {
@@ -59,15 +79,6 @@ class AssignUniqueIdTest : public OperatorTestBase {
     }
 
     ASSERT_EQ(totalInputSize, ids.size());
-
-    auto task = result.first->task();
-
-    // Verify number of memory allocations. There should be exactly one
-    // allocation (per thread of execution) for the values buffer of the unique
-    // ID vector. Memory should be allocated when producing first batch of
-    // output and re-used for subsequent batches.
-    auto stats = toPlanStats(task->taskStats());
-    ASSERT_EQ(1, stats.at(uniqueNodeId_).numMemoryAllocations);
   }
 
   core::PlanNodeId uniqueNodeId_;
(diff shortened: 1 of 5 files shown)


velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -166,3 +179,66 @@ TEST_F(AssignUniqueIdTest, taskUniqueIdLimit) {
       AssertQueryBuilder(plan).copyResults(pool()),
       "(16777216 vs. 16777216) Unique 24-bit ID specified for AssignUniqueId exceeds the limit");
 }
+
+TEST_F(AssignUniqueIdTest, barrier) {
+  std::shared_ptr<const RowType> rowType{
```
mbasmanova: RowTypePtr or auto

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -166,3 +179,66 @@ TEST_F(AssignUniqueIdTest, taskUniqueIdLimit) {
...
+  const int numSplits{5};
+
+  auto tempFiles = [&]() {
+    std::vector<RowVectorPtr> vectors;
+    std::vector<std::shared_ptr<TempFilePath>> tempFiles;
+
+    const int numRowsPerSplit{100};
+    for (int32_t i = 0; i < numSplits; ++i) {
+      auto vector = std::dynamic_pointer_cast<RowVector>(
+          BatchMaker::createBatch(rowType, numRowsPerSplit, *pool_));
```
mbasmanova: BatchMaker is deprecated. Would you use VectorMaker or better yet makeXxx methods from the base class?

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -166,3 +179,66 @@ TEST_F(AssignUniqueIdTest, taskUniqueIdLimit) {
...
+      vectors.push_back(vector);
+      tempFiles.push_back(TempFilePath::create());
+    }
+    writeToFiles(toFilePaths(tempFiles), vectors);
+    return tempFiles;
+  }();
+
+  auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
```
mbasmanova: not needed since plan is linear

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -166,3 +179,66 @@ TEST_F(AssignUniqueIdTest, taskUniqueIdLimit) {
...
+                         .endTableScan()
+                         .project({"c0"})
+                         .planNode();
+    return AssertQueryBuilder(inputPlan)
+        .splits(makeHiveConnectorSplits(tempFiles))
+        .serialExecution(true)
+        .copyResults(pool());
+  }();
+
+  // Create plan for testing
```
mbasmanova: Redundant comment. Let's remove.

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -166,3 +179,66 @@ TEST_F(AssignUniqueIdTest, taskUniqueIdLimit) {
...
+  auto plan = PlanBuilder(planNodeIdGenerator)
+                  .startTableScan()
+                  .outputType(rowType)
+                  .endTableScan()
```
mbasmanova: I believe this can be shortened to .tableScan(rowType)

velox/exec/tests/utils/AssertQueryBuilder.cpp
```diff
@@ -41,7 +41,7 @@ AssertQueryBuilder::AssertQueryBuilder(
 }
 
 AssertQueryBuilder::AssertQueryBuilder(DuckDbQueryRunner& duckDbQueryRunner)
-    : duckDbQueryRunner_{&duckDbQueryRunner} {}
+    : AssertQueryBuilder{nullptr, duckDbQueryRunner} {}
```
mbasmanova: Why this change?

mbasmanova (commented):
@littlesamo Thank you for the change.

> Add option to Cursor to optionally ensure loading vector. 

Should this be updated to reflect the code?

Is it necessary to use Hive table scan for the test?

velox/exec/tests/utils/AssertQueryBuilder.cpp
littlesamo: It is an left over from previous version. I will revert this part.

velox/exec/tests/utils/QueryAssertions.cpp
```diff
@@ -1345,11 +1345,12 @@ std::pair<std::unique_ptr<TaskCursor>, std::vector<RowVectorPtr>> readCursor(
...
   std::vector<RowVectorPtr> result;
   auto* task = cursor->task().get();
-
   while (!cursor->noMoreSplits()) {
     addSplits(cursor.get());
     while (cursor->moveNext()) {
-      result.push_back(cursor->current());
+      auto vector = cursor->current();
+      vector->loadedVector();
+      result.push_back(std::move(vector));
```
xiaoxmeng, line 1353: This might break ss lazy decode test with sequential execution mode per our offline discussion. Also this is not necessary for parallel execution mode which always materialize in sink operator. Maybe awe could add an option in CursorParameters to set to early materialize? Thanks!

xiaoxmeng (commented):
@littlesamo LGTM % minors on test. Thanks!

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -14,33 +14,54 @@
  * limitations under the License.
  */
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/dwio/common/tests/utils/BatchMaker.h"
```
mbasmanova: This is no longer needed.

velox/exec/tests/utils/QueryAssertions.cpp
mbasmanova: It seems that Cursor's implementation is inconsistent.

MultiThreadedTaskCursor always loads lazys, while SingleThreadedTaskCursor doesn't. Hence, when serialExecution=true, Cursor::next may return lazy vector that's not loaded.

I guess it might be useful to not load lazys in single-threaded execution and let the caller decide whether to do that or not, however, this behavior needs to be documented clearly in Cursor.

Now, QueryAssertions is a test utility which is collecting results into std::vector and the only safe way to do that is by loading lazies. Hence, this change seems correct.

@xiaoxmeng Which test are you referring to? "ss lazy decode test with sequential execution mode" - If this test has non-obvious behavior, then perhaps, it needs to use Cursor directly.

xiaoxmeng: SequenceStorageIndexSourceTest.LazyDecode but it seems like the test has been changed and we do not check the return vector now. But lazy decode is only required for SingleThreadedTaskCursor and we could just load there if we want to do this unconditionally.

velox/exec/Cursor.cpp
```diff
@@ -142,7 +142,6 @@ class TaskCursorBase : public TaskCursor {
       const std::shared_ptr<folly::Executor>& executor) {
     static std::atomic<int32_t> cursorId;
     taskId_ = fmt::format("test_cursor_{}", ++cursorId);
-
```
mbasmanova: unintentional change?

velox/exec/tests/AssignUniqueIdTest.cpp
```diff
@@ -16,31 +16,51 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/exec/PlanNodeStats.h"
 #include "velox/exec/tests/utils/AssertQueryBuilder.h"
+#include "velox/exec/tests/utils/HiveConnectorTestBase.h"
 #include "velox/exec/tests/utils/OperatorTestBase.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/exec/tests/utils/QueryAssertions.h"
 
-using namespace facebook::velox;
+namespace facebook::velox::exec::test {
```
mbasmanova: namespace should be facebook::velox::exec

mbasmanova (approved):

PR #14234, opened 2025-07-25
fix: Fix implicit conversion from long to double

oliver6782:
Summary: fix implicit conversion from long to double

Differential Revision: D78952502

### DIFF
--- a/velox/functions/prestosql/aggregates/sfm/SfmSketch.cpp
+++ b/velox/functions/prestosql/aggregates/sfm/SfmSketch.cpp
@@ -219,7 +219,8 @@ int64_t SfmSketch::cardinality() const {
     return 0;
   }
 
-  VELOX_CHECK_LE(guess, std::numeric_limits<int64_t>::max());
+  VELOX_CHECK_LE(
+      guess, static_cast<double>(std::numeric_limits<int64_t>::max()));
   // Clamp negative values to 0 before rounding to avoid undefined behavior
   // when casting negative values to unsigned types.
   double clampedGuess = std::max(0.0, guess);


amitkdutta (approved):
Thanks @oliver6782

czentgr:
@oliver6782 Why did you solve this by casting? The `guess` variable is a double and there is a numeric max for this type. There is no reason to use int64 at all. See https://github.com/facebookincubator/velox/pull/14236
This also avoids the cast.

velox/functions/prestosql/aggregates/sfm/SfmSketch.cpp
```diff
@@ -219,7 +219,8 @@ int64_t SfmSketch::cardinality() const {
     return 0;
   }
 
-  VELOX_CHECK_LE(guess, std::numeric_limits<int64_t>::max());
+  VELOX_CHECK_LE(
+      guess, static_cast<double>(std::numeric_limits<int64_t>::max()));
```
MBkkt, line 223: @oliver6782 Hi, unfortunately this check is incorrect, because `2^63 - 1` casted in double will be `2^63` so if double is equal to 2^63 this check will be true, but result of this function will be incorrect.

should be `VELOX_CHECK_LT`.
And maybe should be something like this, to make it clear from code
```cpp
  VELOX_CHECK_LT(
      guess, -static_cast<double>(std::numeric_limits<int64_t>::min()));
```

oliver6782: great catch, thanks

oliver6782: also see fix #14256

MBkkt: Looks good to me, thanks!

PR #13818, opened 2025-06-18
fix(build): Remove cudf from clang build

czentgr:
cudf does not officially support building with the clang compiler.
Previously, this was fine but the current clang build fails and we need
to remove it.

PR #14236, opened 2025-07-25
fix(build): Comparison of double to max int64

czentgr:
A double type value is compared to numeric_limits<int64_t> when it should be compared to numeric_limits<double>. Not using the correct max value leads to Clang build error
error: implicit conversion from 'long' to 'double' changes value from 9223372036854775807 to 9223372036854775808
[-Werror,-Wimplicit-const-int-float-conversion]

Also another Clang fix is here: https://github.com/facebookincubator/velox/pull/13818
However, this affects the PrestoC++ build CI because it builds with Clang.


oliver6782:
cardinality should return a 64-bit integer, so we need to make sure guess is less than the max value of 64-bit integer

PR #14239, opened 2025-07-25
fix(build): Fix cuDF and other default Makefile builds, test link failure, and tidy indentation

simoneves:
Firstly, on June 9, 2025, @assignUser pushed a change to the root `docker-compose.yml` file which changed the two definitions of `EXTRA_CMAKE_FLAGS` in the `adapters-cpp` and `adapters-cuda` services to use the YAML `>` syntax for removing line breaks from multi-line strings.

Commit diff here: https://github.com/facebookincubator/velox/commit/7c73c1106eb4fab48d0728c19361c8f9294d0c12#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3

For me, at least, this breaks the `Makefile` use of `${EXTRA_CMAKE_FLAGS}` in the `cmake-cudf` and `cmake-wave` rules, because the resulting environment variable inside the container is malformed, resulting in a failure like this:
```
seves@d5a7de6-lcedt:~/work/nvidia/velox$ docker compose run -it adapters-cuda bash
WARN[0000] /home/seves/work/nvidia/velox/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
bash-5.1# cd /velox/
bash-5.1# make cudf
make cmake-cudf BUILD_DIR=release BUILD_TYPE=release
make[1]: Entering directory '/velox'
make EXTRA_CMAKE_FLAGS="-DVELOX_ENABLE_PARQUET=ON -DVELOX_ENABLE_S3=ON
/bin/bash: -c: line 1: unexpected EOF while looking for matching `"'
/bin/bash: -c: line 2: syntax error: unexpected end of file
make[1]: *** [Makefile:107: cmake-cudf] Error 2
make[1]: Leaving directory '/velox'
make: *** [Makefile:135: cudf] Error 2
bash-5.1# 
```

Changing the definitions back to how they were before that commit resolves the problem.

This problem exists in the `rapidsai/velox` fork also.

It's possible this has not been noticed, because if you override that env var for the command, as indeed the `build.sh` script in the NV fork does, it works OK:
```
bash-5.1# EXTRA_CMAKE_FLAGS="foo" make cudf
```

Secondly, default builds of the cuDF variants will fail due to the lack of explicit CUDA architecture, as the default is `52` and cuDF requires at least `70`. This is resolved by setting the default to `native` for the cuDF build rule, unless overridden with `CUDA_ARCHITECTURES` in the build env.

It's possible that the cuDF variant test and/or benchmarks may also require `ENABLE_ARROW=on` and `ENABLE_PARQUET=on` as defaults but this is still being tested.

Finally, tidied up some indentation which was confusing.

FINALLY finally, added a missing lib to the link of `spark_expression_fuzzer_test` which was failing in a default build.


bdice:
There's something a little funny here. I wonder if there is a missing closing quote somewhere. It could be a red herring that this fixes the problem. See the Makefile here: https://github.com/facebookincubator/velox/blob/1d57657586ba74ce7c39a3e6ec073b3e738d5bc0/Makefile#L147-L152

In lines 147 and 152, I don't see a closing quote. I wonder if there's something like that causing a problem.

simoneves:
> In lines 147 and 152, I don't see a closing quote. I wonder if there's something like that causing a problem.

@bdice there ARE closing quotes there, but they're on the next line which is massively indented. Scroll right! :)

bdice:
Well... isn't that unusual. I did notice some whitespace but didn't scroll. Perhaps we should fix that too. 🙃

simoneves:
> Well... isn't that unusual. I did notice some whitespace but didn't scroll. Perhaps we should fix that too. 🙃

Done

Makefile
```diff
@@ -56,6 +56,8 @@ endif
 
 ifdef CUDA_ARCHITECTURES
 CMAKE_FLAGS += -DCMAKE_CUDA_ARCHITECTURES="$(CUDA_ARCHITECTURES)"
+else
+EXTRA_CMAKE_CUDA_FLAGS = -DCMAKE_CUDA_ARCHITECTURES="native"
```
simoneves, line 60: Default this to `native` (for non-cuDF build variants only) if not set in the external env.

Makefile
```diff
@@ -104,7 +106,7 @@ cmake-wave:
 	$(MAKE) EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DVELOX_ENABLE_WAVE=ON" cmake
 
 cmake-cudf:
-	$(MAKE) EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DVELOX_ENABLE_CUDF=ON" cmake
+	$(MAKE) EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DVELOX_ENABLE_CUDF=ON ${EXTRA_CMAKE_CUDA_FLAGS}" cmake
```
simoneves, line 109: See above (invokes this for cuDF variants only)

Makefile
```diff
@@ -118,13 +120,17 @@ release:				#: Build the release version
 	$(MAKE) build BUILD_DIR=release
 
 minimal_debug:			#: Minimal build with debugging symbols
-	$(MAKE) cmake BUILD_DIR=debug BUILD_TYPE=debug EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} -DVELOX_BUILD_MINIMAL=ON"
+	$(MAKE) cmake BUILD_DIR=debug BUILD_TYPE=debug \
+		EXTRA_CMAKE_FLAGS="${EXTRA_CMAKE_FLAGS} \
+			-DVELOX_BUILD_MINIMAL=ON"
```
simoneves, line 125: Tidy up indentations (x6) including two with way too many tabs!

docker-compose.yml
```diff
@@ -52,9 +52,7 @@ services:
     environment:
       NUM_THREADS: 8 # default value for NUM_THREADS
       CCACHE_DIR: /velox/.ccache
-      EXTRA_CMAKE_FLAGS: >
-        -DVELOX_ENABLE_PARQUET=ON
-        -DVELOX_ENABLE_S3=ON
+      EXTRA_CMAKE_FLAGS: -DVELOX_ENABLE_PARQUET=ON -DVELOX_ENABLE_S3=ON
```
simoneves: Undo use of YAML `>` syntax which results in a malformed env var in the container.

assignUser: Ok so after some investigation this was indeed caused by '>' as that leaves a newline at the end. But we can just replace it with `>-` which will NOT add that newline and fix the issue.

Makefile
assignUser: Should L58 also be EXTRA_CMAKE_CUDA_FLAGS for consistency?

Makefile
assignUser: Does the wave build also need the cuda arch?

Makefile
simoneves: This was my attempt at a separate chunk of config that could then ONLY be included in the `cudf` variants, rather than all of them, even if the others would ignore it.

Makefile
simoneves: > Does the wave build also need the cuda arch?

I actually don't know.

docker-compose.yml
simoneves: @assignUser done and confirmed working

velox/expression/fuzzer/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ add_executable(
 target_link_libraries(
   spark_expression_fuzzer_test
   velox_expression_fuzzer
+  velox_constrained_input_generators
```
simoneves, line 68: Missing lib causing a link failure of this test exe

docker-compose.yml
```diff
@@ -52,7 +52,7 @@ services:
     environment:
       NUM_THREADS: 8 # default value for NUM_THREADS
       CCACHE_DIR: /velox/.ccache
-      EXTRA_CMAKE_FLAGS: >
+      EXTRA_CMAKE_FLAGS: >-
```
simoneves: Tweaked syntax to avoid trailing newline on env var value, per @assignUser

Makefile
simoneves: It seems that it does. Will add.

simoneves: Added. I also note that the Wave build is very noisy, CUDA-wise:
* Deprecation warnings for `cudaDeviceSetSharedMemConfig()`in some tests
* Pre-processor parsing warnings for calls to `module->launch()` in various `.cu` files
* Missing return statement warnings in `GpuDecoder-inl.cuh`
* Several other general warnings

simoneves:
@assignUser @majetideepak ready for re-review

Makefile
simoneves: ...and now the `wave` variants also

simoneves:
Can somebody please help me interpret the Fuzzer check job failures, as I am not familiar. Not sure if those were triggered by my library change, or if I just got unlucky with some random behavior?

Edlt: Both seem to be triggered by an unrecognized geometry type in an ST function test:
```
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Unrecognized geometry type: 71
Retriable: False
Context: st_numpoints(GLC&"-|*FKAw\}X@ka2R$A>=,DV/$Q[_Zru=H6y9S7AD%MiZ'KI[gjk#]2o;/.|{:GEOMETRY)
```

bdice (approved):
This looks good to me! Thanks @simoneves.

assignUser (approved):
Nice, looks good! Thanks for the detailed PR description and comments, makes reviewing fun!

assignUser:
(the fuzzer issues are also on main, so unrelated :+1: )

simoneves:
Thanks for the approvals, @assignUser and @bdice. I synced my fork and re-based over @assignUser 's change to `docker-compose.yml` but then I tried another default `make cudf` and hit another test link fail, so I just pushed a fix for that too. Hope this PR is still acceptable. I assume I don't have permissions to land it myself...

kKPulla:
> Thanks for the approvals, @assignUser and @bdice. I synced my fork and re-based over @assignUser 's change to `docker-compose.yml` but then I tried another default `make cudf` and hit another test link fail, so I just pushed a fix for that too. Hope this PR is still acceptable. I assume I don't have permissions to land it myself...

Thanks Simon! @assignUser, @bdice - can one of you please take a quick look to make sure all is good in the latest revision and I will initiate the merge.

bdice (approved):
Thanks @simoneves! Looks good to me.

kKPulla:
@simoneves can you rebase on main as we had a CI breakage that prevented fuzzer signals from this PR? Thanks!

simoneves:
@kKPulla done

Issue #14183, opened 2025-07-22
[cuDF] Refactor ParquetConnector used to enable cudf reading of parquet


devavret:
### Description

Currently we support cudf based table scans using a separate set of Connector related classes (`ParquetConnector`, `ParquetDataSource`, `ParquetTableHandle`, `ParquetConnectorSplit`, ...). This has the following issues:
- **Integration issues**: Existence of a separate `ParquetConnector` requires one to choose GPU reading at plan generation. This inconsistent with how we handle other operators which replace operators out of an existing cpu based plan. This causes problems in integration e.g. in TpchBenchmark where cudf integration related code "spills out" into velox codebase. [Example in benchmark PR](https://github.com/devavret/velox/blob/aee030ed1ff1817c7f519ebcedf79fd028661931/velox/exec/tests/utils/PlanBuilder.cpp#L308-L334)
- **Duplicate Code**: `ParquetConnector`, `ParquetTableHandle`, `ParquetSplit` contain a reduced subset of functionality that already exists in corresponding `Hive*` classes
- **Code organization**: Logically, cudf's parquet reader is not a new type of `DataSource` (in Presto's definition of the term) but a file decoder which ideally should be usable by all `DataSource`s that may need to read parquet. Its parallel in the CPU reader realm is `facebook::velox::parquet::ParquetReader` which in case of hive is used by `hive::SplitReader` which is used by `HiveDataSource`

The goal is to push the cudf parquet integration closer to `ParquetReader` layer in order to reuse as much of existing infrastructure as possible.

### Current infrastructure:
```
[Plan]
HiveTableHandle (Determines which connector to use)

[Driver Pipeline]
TableScan ------------> HiveConnector
               via                    
          TableHandle               
TableScan ------------> HiveDataSource -> hive::SplitReader ---------------> ParquetReader
               via                                                via
          HiveConnector                                     gerReaderFactory
```
`HiveDataSource` is responsible for applying remaining filter on the next batch. The expression evaluator used in `HiveDataSource` expects Velox `RowVector`s on CPU memory and would not work on `CudfVector`s. This limits our cudf injection to the `DataSource` layer unless we can adapt the ExpressionEvaluator in `HiveDataSource` to work on `CudfVector`.

#### 1. Make existing `Parquet*` classes derive from `Hive*` classes
  - Change `ParquetConnector` to `CudfConnector : public HiveConnector`
[...]

PR #14248, opened 2025-07-28
feat(cuDF): Add a function to construct a cuDF AST from SubfieldFilters

devavret:
This PR contributes to the issue https://github.com/facebookincubator/velox/issues/14183 which pertains to seamlessly enabling cudf based file reading by translating existing Hive related objects to corresponding cuDF related ones.

`HiveTableHandle` has two members that contribute to filtered reading: a range filter of type `SubfieldFilters` and a remaining filter. In order to allow automatic construction of a `CudfHiveTableHandle` from a `HiveTableHandle`, the former needs to be able to convert the `SubfieldFilters`  object into a cuDF AST, which can be used by cuDF's parquet reader. This PR adds a function to do that conversion


Yuhta (approved):

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -820,4 +833,365 @@ bool ExpressionEvaluator::canBeEvaluated(
...
+      return tree.push(literal);
+    };
+
+    if (bigintRange->isSingleValue()) {
+      // Equal comparison: column = value. This value is the same as the
+      // lower/upper bound.
+      if (skipLowerBound || skipUpperBound) {
+        // If the singular value of this filter lies outside the range of the
+        // column's NativeT type then we want to be always false
+        return tree.push(Operation{Op::NOT_EQUAL, columnRef, columnRef});
```
karthikeyann, line 947: ~nit:~
nulls comparions return nulls. would `NULL_EQUAL` operator suffice here?

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -820,4 +833,365 @@ bool ExpressionEvaluator::canBeEvaluated(
...
+      } else {
+        auto const& literal = addLiteral(lower);
+        return tree.push(Operation{Op::EQUAL, columnRef, literal});
```
karthikeyann, line 950: Since the `columnRef` can be any integral type, we need to add `cast` operator if it is not int64 type, to compare with `int64` literal. Otherise AST will throw
```
E20250821 02:08:47.042325  9895 Exceptions.h:66] Line: /velox/velox/exec/Driver.cpp:574, Function:operator(), Expression:  Operator::getOutput failed for [operator: TableScan, plan node ID: 0]: CUDF failure at:/velox/_build/release/_deps/cudf-src/cpp/src/ast/expression_parser.cpp:244: An AST expression was provided non-matching operand types., Source: RUNTIME, ErrorCode: INVALID_STATE
```

karthikeyann (changes requested):

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
karthikeyann: This breaks Q16 in velox cudf tpch benchmark.

karthikeyann: The literal for AST comparison operator is int64.  For bigInt range, the column used to compare may not be only int64. it could be any other integral. AST will fail if the operands compared are not of same type. So, the column needs to be casted to int64.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
karthikeyann: null != null will return null for NOT_EQUAL.
null != null will return false for NULL_EQUAL

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
devavret: This is not supposed to happen as the `addLiteral` function only makes a literal of the same type as the column. See that the literal is of NativeT type which is the type of the column when dispatched. Moreover, this scenario is tested in unit tests https://github.com/devavret/velox/blob/33adf15ba797388f105befa99750a3f36a90d7d2/velox/experimental/cudf/tests/SubfieldFilterAstTest.cpp#L358-L378
Besides, this PR does not wire in this functionality in the table scans. Are you sure this conversion function was used?

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
devavret: if the value contained in this filter lies outside the type range of this column, we want to fail this filter always. This line is a stand-in for returning false always.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
karthikeyann: SubField Filter is used in TableScan. it's not used in FilterProject.

devavret: In this PR, it's not used in tablescan either.

velox/experimental/cudf/tests/SubfieldFilterAstTest.cpp
```diff
@@ -0,0 +1,593 @@
...
+  EXPECT_LE(scalars.size(), 2UL) << "Too many scalars for range filter";
+
+  // Execution validation
+  auto vec = makeTestVector(rowType, 100);
+  testFilterExecution(rowType, columnName, *filter, vec, expr);
+}
+
+TEST_F(SubfieldFilterAstTest, Int64InList) {
+  const std::string columnName = "c0";
+  auto rowType = ROW({{columnName, BIGINT()}});
```
karthikeyann: Updating this rowType from BIGINT to INTEGER , makes the test fail.

karthikeyann (changes requested):

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
karthikeyann: Yes. not this PR, I found the issue while testing https://github.com/facebookincubator/velox/pull/13695

devavret: oh okay that makes sense.

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -822,4 +833,440 @@ bool ExpressionEvaluator::canBeEvaluated(
...
+cudf::ast::expression const& createAstFromSubfieldFilter(
+    const common::Subfield& subfield,
+    const common::Filter& filter,
+    cudf::ast::tree& tree,
+    std::vector<std::unique_ptr<cudf::scalar>>& scalars,
+    const RowTypePtr& inputRowSchema) {
+  // First, create column reference from subfield
+  // For now, only support simple field references
+  if (subfield.path().empty() ||
+      subfield.path()[0]->kind() != common::kNestedField) {
```
zhouyuan, line 1129: @devavret 
it should be:  `common::SubfieldKind::kNestedField` ?

devavret: This seems to have resulted from a race between PR merges. https://github.com/facebookincubator/velox/pull/14724 merged yesterday but this PR was imported 2 days ago when `SubfieldKind` was not an enum class yet. This needs a fix now.

zhouyuan: Would you please help to make a fix for this? The CUDF build is in failure now

devavret: When resolving merge conflicts, I had to make this fix in this PR https://github.com/facebookincubator/velox/pull/14462. Should be in soon

PR #14263, opened 2025-07-29
fix: Remove duplication in top-level CMakeLists.txt

MBkkt:
cxx version was set twice in different value without comment, looks like typo

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 # limitations under the License.
 cmake_minimum_required(VERSION 3.28)
 message(STATUS "Building using CMake version: ${CMAKE_VERSION}")
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 set(CMAKE_CXX_EXTENSIONS ON) # Big Int is an extension
 message(STATUS "CXX standard: ${CMAKE_CXX_STANDARD}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 cmake_minimum_required(VERSION 3.28)
 message(STATUS "Building using CMake version: ${CMAKE_VERSION}")
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 set(CMAKE_CXX_EXTENSIONS ON) # Big Int is an extension
```
MBkkt, line 18: I also don't sure is it necessary or not now. Because for me with modern clang and libc++ it works with `set(CMAKE_CXX_EXTENSIONS OFF)`. But I don't build whole velox only this:
```cmake
set(VELOX_MONO_LIBRARY ON CACHE BOOL "" FORCE)
set(VELOX_ENABLE_EXEC ON CACHE BOOL "" FORCE)
set(VELOX_ENABLE_AGGREGATES ON CACHE BOOL "" FORCE)
set(VELOX_ENABLE_PRESTO_FUNCTIONS ON CACHE BOOL "" FORCE)
set(VELOX_ENABLE_SPARK_FUNCTIONS ON CACHE BOOL "" FORCE)
set(VELOX_ENABLE_EXPRESSION ON CACHE BOOL "" FORCE)
```

assignUser: It's better to be explicit with the standard, otherwise older compilers might decay to 17 which will cause issues when we start using 20 features :) 

`CMAKE_CXX_EXTENSIONS` defaults to `ON`, that's why you can remove it without issue. But again, being explicit about it is better as this plays the role of documentation here.

assignUser (approved):
Ah, this looks like a merge/rebase error. Good catch, thanks!

CMakeLists.txt
MBkkt: I mean interesting is everything compile with extensions OFF or not. Feels like it should compile with off, and it works for subset of velox, not sure about every target

MBkkt: But maybe you're right about older compilers, thanks

PR #14269, opened 2025-07-29
fix: Do not keep reference to input vector in FilterProject longer than needed

Yuhta:
Summary:
`FilterProject` always consumes all the input by the time we return
from `getOutput`, so there is no need to track number of rows consumed.  In the
current implementation, this was causing us to hold on to the input after
`getOutput` is returned and preventing memory reusing when the batch is processed by downstream operators.

Differential Revision: D79174768


mbasmanova (approved):
Thanks.

PR #14272, opened 2025-07-29
feat: collect iceberg data file statistics

PingLiuPing:
Full PR that support iceberg data file statistics.


velox/dwio/common/Statistics.h
```diff
@@ -18,6 +18,7 @@
 
 #include <folly/Hash.h>
 #include <folly/container/F14Map.h>
+#include <folly/json/dynamic.h>
```
jinchengchenghh: Remove

velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "velox/connectors/hive/TableHandle.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+struct IcebergNestedField {
+  int32_t id;
+  std::vector<IcebergNestedField> children;
+
+  IcebergNestedField(
```
jinchengchenghh: Remove the constructors.

velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -0,0 +1,52 @@
...
+  IcebergColumnHandle(
+      const std::string& name,
+      ColumnType columnType,
+      TypePtr dataType,
+      TypePtr hiveType,
+      const IcebergNestedField& nestedField,
+      std::vector<common::Subfield> requiredSubfields = {},
+      ColumnParseParameters columnParseParameters = {});
+
+  const IcebergNestedField nestedField() const;
```
jinchengchenghh: const IcebergNestedField&

velox/dwio/common/Statistics.h
PingLiuPing: thanks.

velox/connectors/hive/iceberg/IcebergColumnHandle.h
PingLiuPing: thanks

velox/connectors/hive/iceberg/IcebergColumnHandle.h
PingLiuPing: thanks.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +511,68 @@ void Writer::abort() {
   arrowContext_.reset();
 }
 
+void Writer::collectIcebergDataFileStats(
```
4ertus2: Wouldn't it be better to make separate IcebergStatsCollector outside of `parquet::Writer` with this logic? Actually it's not about parquet at all.

velox/dwio/common/Writer.h
```diff
@@ -92,6 +99,10 @@ class Writer {
   static void checkStateTransition(State oldState, State newState);
 
   State state_{State::kInit};
+  bool collectIcebergStats_{false};
+  std::shared_ptr<IcebergDataFileStatistics> dataFileStats_{nullptr};
+  std::shared_ptr<std::vector<std::pair<int32_t, bool>>> sourceColumnIndices_{
```
4ertus2: Could it help to save "PARQUET:field_id" for schema evolution?
According to https://iceberg.apache.org/spec/#parquet

> Column IDs are required to be stored as [field IDs](http://github.com/apache/parquet-format/blob/40699d05bd24181de6b1457babbee2c16dce3803/src/main/thrift/parquet.thrift#L459) on the parquet schema.

Is int32_t here an original Iceberg's field_id or not?

velox/dwio/common/Options.h
```diff
@@ -706,6 +706,13 @@ struct WriterOptions {
   std::string sessionTimezoneName;
   bool adjustTimestampToTimezone{false};
 
+  // This option controls whether collect iceberg data file statistics
+  // during write. It will be set to true only by IcebergDataSink.
+  bool collectIcebergStats{false};
```
4ertus2: There're more restrictions for Iceberg. I.e. according to https://iceberg.apache.org/spec/#parquet

> Column IDs are required to be stored as [field IDs](http://github.com/apache/parquet-format/blob/40699d05bd24181de6b1457babbee2c16dce3803/src/main/thrift/parquet.thrift#L459) on the parquet schema.

> Lists must use the [3-level representation](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists).

It could be one general `IcebergCompatible` option here to accumulate all these issues.

velox/connectors/hive/HiveDataSink.cpp
```diff
@@ -763,6 +757,32 @@ uint32_t HiveDataSink::appendWriter(const HiveWriterId& id) {
...
+
+  writerIndexMap_.emplace(id, writers_.size() - 1);
+  return writerIndexMap_[id];
+}
+
+std::shared_ptr<dwio::common::WriterOptions> HiveDataSink::createWriterOptions()
+    const {
   // Take the writer options provided by the user as a starting point, or
   // allocate a new one.
   auto options = insertTableHandle_->writerOptions();
```
4ertus2, line 788: It's not clear why `options` are not null here. `writerOptions` is an optional argument for `HiveInsertTableHandle` ctor.

velox/dwio/common/Writer.h
PingLiuPing: Thanks @4ertus2 

> Is int32_t here an original Iceberg's field_id or not?

Yes, it is the original icebergs's field ID, and it must be the field ID, because the stats map will be serialized to json and write to manifest.

> Could it help to save "PARQUET:field_id" for schema evolution?

Will have a look at this. Actually there are defects in current parquet reader during schema evolution. If we can save the iceberg original field ID in parquet file, possibly we can resolve this.

4ertus2:
```
@@ -451,8 +457,15 @@ void Writer::write(const VectorPtr& data) {
   std::vector<std::shared_ptr<::arrow::Field>> newFields;
   auto childSize = schema_->size();
   for (auto i = 0; i < childSize; i++) {
-    newFields.push_back(updateFieldNameRecursive(
-        arrowSchema->fields()[i], *schema_->childAt(i), schema_->nameOf(i)));
+    auto field = updateFieldNameRecursive(
+        arrowSchema->fields()[i], *schema_->childAt(i), schema_->nameOf(i));
+    if (sourceColumnIndices_) {
+      VELOX_CHECK(i < sourceColumnIndices_->size());
+      if (int32_t field_id = (*sourceColumnIndices_)[i].first) {
+        field = field->WithMetadata(arrow::arrow::FieldIdMetadata(field_id));
+      }
+    }
+    newFields.emplace_back(std::move(field));
   }
 
   PARQUET_ASSIGN_OR_THROW(
```
for filed_ids. `FieldIdMetadata` from velox/dwio/parquet/writer/arrow/ArrowSchema.cpp

velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/connectors/hive/TableHandle.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+struct IcebergNestedField {
+  int32_t id;
```
4ertus2, line 23: It's better to make it zero by default
```
int32_t id{};
```

velox/dwio/common/Writer.h
PingLiuPing: @4ertus2 , thanks, but can you clarify what is your questions here?

4ertus2: It's not a question. It's an easy way to add field_ids in parquet with this patch.

I've tested it in our fork with changes from your PR.

PingLiuPing: Oh, I have used similar approach to add filed_id. if you found any issues please let me know.
This PR contains accumulated Iceberg insertion related code. There are separate PRs to handle different features though. Since they are not merged, I have to include all the commits to run CI.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+    const auto& rowType = type.asRow();
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
-    auto childrenSize = rowType.size();
     std::vector<std::shared_ptr<::arrow::Field>> newFields;
-    newFields.reserve(childrenSize);
-    for (auto i = 0; i < childrenSize; i++) {
+    newFields.reserve(rowType.size());
+    for (auto i = 0; i < rowType.size(); ++i) {
+      const auto* childSetting =
```
jinchengchenghh: &icebergFieldSetting -> icebergFieldSetting

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
 std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const dwio::common::IcebergStatsSettings* icebergFieldSetting = nullptr) {
```
jinchengchenghh: icebergFieldSetting -> icebergStatsSetting

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+  auto newField = field;
+  if (!name.empty()) {
+    newField = field->WithName(name);
+  }
   if (type.isRow()) {
-    auto& rowType = type.asRow();
-    auto newField = field->WithName(name);
+    const auto& rowType = type.asRow();
```
jinchengchenghh: Do not include the unrelated change

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
 std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const dwio::common::IcebergStatsSettings* icebergFieldSetting = nullptr) {
+  auto newField = field;
```
jinchengchenghh: Do not extract newField, make minimal change

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -196,44 +200,67 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
   } else if (type.isMap()) {
-    auto mapType = type.asMap();
-    auto newField = field->WithName(name);
+    const auto& mapType = type.asMap();
     auto arrowMapType =
         std::dynamic_pointer_cast<::arrow::MapType>(newField->type());
-    auto newKeyField =
-        updateFieldNameRecursive(arrowMapType->key_field(), *mapType.keyType());
+    const auto* keySetting =
+        icebergFieldSetting ? &icebergFieldSetting->children.at(0) : nullptr;
```
jinchengchenghh: ditto

velox/dwio/common/Options.h
```diff
@@ -684,6 +684,15 @@ class ReaderOptions : public io::ReaderOptions {
   bool allowEmptyFile_{false};
 };
 
+// Settings for collecting Iceberg column statistics. Holds the Iceberg
```
jinchengchenghh: ///

velox/dwio/common/Options.h
```diff
@@ -706,6 +715,13 @@ struct WriterOptions {
   std::string sessionTimezoneName;
   bool adjustTimestampToTimezone{false};
 
+  // This option controls whether collect iceberg data file statistics
```
jinchengchenghh: ///

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +539,116 @@ void Writer::abort() {
...
+
+  // numColumns is not the number of columns in Iceberg table's schema,
+  // e.g., schema_->size(). It also contains the sub-fields when there are
+  // nested types in table's schema.
+  int32_t numColumns = 0;
+  for (const auto& field : *icebergStatsSetting_) {
+    numColumns += countLeafFields(field);
+  }
+
+  const std::function<const dwio::common::IcebergStatsSettings*(int32_t)>
```
jinchengchenghh: Maybe flatten the IcebergStatsSettings first, this will cause recursively search the settings for each field.

velox/dwio/common/Writer.h
PingLiuPing: @4ertus2  FYI, I think your patch cannot handle nested type, it only process top level fields.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, I need to make newField available in the function scope so that I can reuse this variable and then attach metadata to it. And by extract it to function scope I can just add 
```
  if (icebergStatsSetting) {
    newField = newField->WithMetadata(
        arrow::arrow::FieldIdMetadata(icebergStatsSetting->id));
  }
```
at the end of this function instead of adding this if statement in each if (type.xxx) branch.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, the parameter type  of updateFieldNameRecursive is 
`const dwio::common::IcebergStatsSettings* icebergStatsSetting = nullptr`
So I have to use pointer here.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -502,6 +539,116 @@ void Writer::abort() {
...
+  std::unordered_map<int32_t, std::shared_ptr<arrow::Statistics>>
+      globalMinStats;
+  std::unordered_map<int32_t, std::shared_ptr<arrow::Statistics>>
+      globalMaxStats;
+
+  const auto numRowGroups = fileMetadata->num_row_groups();
+  for (int i = 0; i < numRowGroups; ++i) {
+    const auto rgm = fileMetadata->RowGroup(i);
+    VELOX_CHECK_EQ(numColumns, rgm->num_columns());
+
```
p1rattttt: `dataFileStats_->splitOffsets.emplace_back(rgm->file_offset()); `

velox/dwio/common/IcebergStatistics.h
```diff
@@ -0,0 +1,38 @@
...
+// Iceberg data_file struct fields.
+struct IcebergDataFileStatistics {
+  int64_t numRecords;
+  std::unordered_map<int32_t, int64_t> columnsSizes;
+  std::unordered_map<int32_t, int64_t> valueCounts;
+  std::unordered_map<int32_t, int64_t> nullValueCounts;
+  std::unordered_map<int32_t, int64_t> nanValueCounts;
+  std::unordered_map<int32_t, std::string> lowerBounds;
+  std::unordered_map<int32_t, std::string> upperBounds;
+
```
p1rattttt: `std::vector<uint64_t> splitOffsets; `

velox/dwio/common/IcebergStatistics.cpp
```diff
@@ -0,0 +1,43 @@
...
+    return result;
+  };
+
+  json["columnSizes"] = mapToJson(columnsSizes);
+  json["valueCounts"] = mapToJson(valueCounts);
+  json["nullValueCounts"] = mapToJson(nullValueCounts);
+  json["nanValueCounts"] = mapToJson(nanValueCounts);
+  json["lowerBounds"] = mapToJson(lowerBounds);
+  json["upperBounds"] = mapToJson(upperBounds);
+
```
p1rattttt: folly::dynamic arr = folly::dynamic::array;

  for (const auto& offset : splitOffsets) {
    arr.push_back(offset);
  }
  json["splitOffsets"] = arr;

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks. Yes, it is better. I can remove this lambda completely and using a `set` to hold the fields that should be skipped.

p1rattttt (commented):
I think it would be good to also collect split offsets, as they can be very useful when reading Parquet files

PingLiuPing:
> Hi! I think it would be good to also collect split offsets, as they can be very useful when reading Parquet files

Thanks @p1rattttt , yes, it should be collected. I will add it.
I'm using Presto to test this and found that it doesn't collect split offsets.

Screenshot 2025-08-13 at 11 55 57

velox/dwio/common/IcebergStatistics.cpp
```diff
@@ -0,0 +1,49 @@
+/*
```
yingsu00: This should be file format(parquet,dwrf,etc.) agnostic and table format(hive, iceberg, etc) agnostic. I would say it's representing generic FileStatistics that is not specific to Iceberg. Connector specific notion should NOT appear in dwio::common. It's ok to rename it to FileStatistics.cpp/h, but it's better to put them in velox/dwio/common/Statistics.h, which contains a umbrella `Statistics` class, from which you can get the `ColumnStatistics`. You you put your new file level statistics as a new class `FileStatistics` there. A `getFileStatistics()` should be added to `Statstics` class, in analygous to `getColumnStatistics()`

velox/dwio/common/IcebergStatistics.h
```diff
@@ -0,0 +1,42 @@
...
+  // Split offsets for the data file. For example, all row
+  // group offsets in a Parquet file. Must be sorted ascending.
+  std::vector<uint64_t> splitOffsets;
```
yingsu00: "Split" is not a notion in a file format, but rather a concept in a data processing engine like Presto, Prestissimo. I think you meant RowGroup or Stripe offsets. Since Parquet doesn't have "Stripe"s, we can just have a pure interface function `std::vector<uint64_t> rowGroupOffsets()` here, and let Parquet and Dwrf have their overrides.

velox/dwio/common/IcebergStatistics.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#pragma once
+
+#include <folly/json/dynamic.h>
+
+namespace facebook::velox::dwio::common {
+
+// Iceberg data_file struct fields.
+struct IcebergDataFileStatistics {
```
yingsu00: Rename to `DataFileStatistics`

velox/dwio/common/Options.h
```diff
@@ -684,6 +684,15 @@ class ReaderOptions : public io::ReaderOptions {
   bool allowEmptyFile_{false};
 };
 
+/// Settings for collecting Iceberg column statistics. Holds the Iceberg
+/// source column ID (field_id) and whether to skip bounds collection for this
```
yingsu00: It is not exactly a "column" ID. It's field id and could be nested fields (struct members, list elements, map keys/values). "column" is only for top level fields so the comment is confusing. Can you please remove the "column ID (" part?

velox/dwio/common/Options.h
```diff
@@ -684,6 +684,15 @@ class ReaderOptions : public io::ReaderOptions {
...
+/// field. For nested field, it contains child fields.
+struct IcebergStatsSettings {
+  int32_t id;
```
yingsu00: IMO `fieldId` is more explanatory.

velox/dwio/common/Options.h
```diff
@@ -684,6 +684,15 @@ class ReaderOptions : public io::ReaderOptions {
   bool allowEmptyFile_{false};
 };
 
+/// Settings for collecting Iceberg column statistics. Holds the Iceberg
+/// source column ID (field_id) and whether to skip bounds collection for this
+/// field. For nested field, it contains child fields.
+struct IcebergStatsSettings {
```
yingsu00: Again, It's not appropriate to have "Iceberg" stuff in dwio::common. This is just some generic stats collection options. Could you please rename it to `StatsCollectionOptions` or an even more explanatory name?

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -742,6 +779,55 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+      }
+    }
+    if constexpr (std::is_same_v<T, int128_t>) {
+      return encodeDecimalToBigEndian(min_);
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      // TODO: 16 is default value. See DEFAULT_WRITE_METRICS_MODE_DEFAULT in
+      // org.apache.iceberg.TableProperties. Need to support this table
+      // property.
+      const auto truncatedMin = UnicodeUtil::truncateStringMin(
```
jinchengchenghh, line 802: Please return std::string_view here, otherwise, may copy and cause performance issue

velox/dwio/common/Options.h
PingLiuPing: Thanks Ying, fixed.

velox/dwio/common/Options.h
PingLiuPing: Thanks Ying. I think this is Iceberg specific because
1. The id maps to Icebeg field ID, which is a unique ID of all fields. But when the type of one field is nested type, here we don't need to collect stats for that field. Instead we only collect the sub-field stats.
2. For the second member `skipBonds`, this is also iceberg specific that we need to skip map and array type for bounds collection.

velox/dwio/common/Options.h
PingLiuPing: Yes, thanks.

velox/dwio/common/IcebergStatistics.cpp
PingLiuPing: @p1rattttt  I moved this field to the top level of the commit metadata. 
As you can see there is no such field in org.apache.iceberg.Metrics

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks.

velox/dwio/common/IcebergStatistics.h
PingLiuPing: Thanks, I split this into base class and Iceberg specific class.

velox/dwio/common/Options.h
PingLiuPing: Moved this to Iceberg connector

velox/dwio/common/IcebergStatistics.h
PingLiuPing: https://iceberg.apache.org/spec/#data-file-fields

velox/dwio/parquet/writer/arrow/Schema.h
```diff
@@ -505,6 +505,10 @@ class PARQUET_EXPORT ColumnDescriptor {
     return node_;
   }
 
+  int32_t field_id() const {
```
jinchengchenghh: Maybe you don't need to add the function, call schema_node()->field_id()

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -94,9 +97,15 @@ struct CompareHelper {
       "T is an unsigned numeric");
 
   constexpr static T DefaultMin() {
+    if constexpr (std::is_floating_point_v<T>) {
```
jinchengchenghh, line 100: I would suggest to add the INF process to GetMinMax, the DefaultMin is Inf looks not reasonable.

velox/dwio/common/IcebergStatistics.cpp
PingLiuPing: Initially I have looked the Statistics.h and found that this statistics is not used during writing. It is used for reader.

I deleted this file and move them to Iceberg connector.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks, I think `Inf` is just a special value that represent something, when we compare it with `normal` float point numbers we can get:

-Inf < lowest < max < Inf

So, I think somehow it is reasonable to make it as the minimal value.

velox/dwio/parquet/writer/arrow/Schema.h
PingLiuPing: thanks. removed.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, added collector in connector level.

velox/connectors/hive/iceberg/IcebergColumnHandle.h
PingLiuPing: Thanks, will change this when split this huge PR to smaller pieces.

velox/dwio/common/Options.h
PingLiuPing: Thanks, this is a good idea, will think about this when splitting the PRs.

PR #14274, opened 2025-07-30
feat(type): Add TypeParser logic to parse BigintEnumType

HeidiHan0000:
Summary:
This PR adds TypeParser logic in the Bison/Flex rules to be able to recognize a type string in the format of:
enumName:BigintEnum(enumName{"mapKey":0 ... })
which will call getType with the appropriate parameters to return a BigintEnumType.

Differential Revision: D78903388

### DIFF
--- a/velox/type/parser/TypeParser.yy
+++ b/velox/type/parser/TypeParser.yy
@@ -34,7 +34,7 @@
     #define yylex(x) scanner->lex(x)
 }
 
-%token               LPAREN RPAREN COMMA ARRAY MAP ROW FUNCTION DECIMAL
+%token               LPAREN RPAREN COMMA ARRAY MAP ROW FUNCTION DECIMAL LBRACE RBRACE
 %token <std::string> WORD VARIABLE QUOTED_ID
 %token <long long>   NUMBER
 %token YYEOF         0
@@ -149,7 +149,6 @@ named_type : type_single_word        { $$ = std::make_pair("", $1); }
            | type_with_spaces        { $$ = inferTypeWithSpaces($1, false); }
            | QUOTED_ID type          { $1.erase(0, 1); $1.pop_back(); $$ = std::make_pair($1, $2); }  // Remove the quotes.
            ;
-
 %%
 
 void facebook::velox::type::Parser::error(const std::string& msg) {
--- a/velox/docs/develop/types.rst
+++ b/velox/docs/develop/types.rst
@@ -176,6 +176,7 @@ IPPREFIX                  ROW(HUGEINT,TINYINT)
 GEOMETRY                  VARBINARY
 TDIGEST                   VARBINARY
 QDIGEST                   VARBINARY
+BIGINT_ENUM               BIGINT
 ========================  =====================
 
 TIMESTAMP WITH TIME ZONE represents a time point in milliseconds precision
@@ -222,6 +223,16 @@ data for a given input set, and can be queried to retrieve approximate quantile 
 distribution. They may be merged without losing precision, and for storage and retrieval they may
 be cast to/from VARBINARY. The parameter type (BIGINT, REAL, or DOUBLE) represents
 the set of numbers that may be ingested by the quantile digest.
+
+BIGINT_ENUM(LongEnumParameter) type represents an enumerated value where the physical type is BIGINT.
+It takes one LongEnumParameter as parameter, which consists of a string name and a mapping of
+string keys to BIGINT values.
+There is a static cache which stores instances of different BIGINT_ENUM types. This is to treat each
+different enum type as a singleton. The LongEnumParameter is used as the key to retrieve the cached instance,
+and a new instance is only created if it has not been created with the given LongEnumParameter.
+Casting is permitted from any integer type to an enum type. Casting is only permitted from an enum type
+to a BIGINT type. Casting between different enum types is not permitted.
+Comparison operations are only allowed between values of the same enum type.
 
 Spark Types
 ~~~~~~~~~~~~
--- a/velox/expression/tests/CustomTypeTest.cpp
+++ b/velox/expression/tests/CustomTypeTest.cpp
@@ -236,6 +236,7 @@ TEST_F(CustomTypeTest, getCustomTypeNames) {
       "TDIGEST",
       "QDIGEST",
       "SFMSKETCH",
+      "BIGINT_ENUM",
   };
 #ifdef VELOX_ENABLE_GEO
   expectedTypes.insert("GEOMETRY");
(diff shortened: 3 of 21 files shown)


velox/type/parser/ParserUtil.cpp
```diff
@@ -65,4 +65,35 @@ std::pair<std::string, std::shared_ptr<const Type>> inferTypeWithSpaces(
...
+  }
+
+  return result;
+}
+
+TypePtr getEnumType(
+    std::string enumType,
+    const std::string& enumName,
+    const std::string& valuesMap) {
+  std::vector<TypeParameter> params;
```
kewang1024: I know that we have a check in yy file, but for code integrity, we can add a velox check here for enumType

velox/type/parser/ParserUtil.cpp
```diff
@@ -65,4 +65,35 @@ std::pair<std::string, std::shared_ptr<const Type>> inferTypeWithSpaces(
       fieldName, typeFromString(allWords.substr(fieldName.size() + 1)));
 }
 
+std::unordered_map<std::string, int64_t> parseMapFromString(
+    const std::string& input) {
+  // Wrap with braces to make valid JSON
+  std::string jsonStr = "{" + input + "}";
```
kewang1024: why don't we move this logic into .yy file? basically the input you get here should be a valid json already

for example
enum_map_entries_json : LBRACE enum_map_entries RBRACE {$$ = "{" + $2 + "}";}

velox/type/parser/TypeParser.yy
```diff
@@ -149,6 +152,36 @@ named_type : type_single_word        { $$ = std::make_pair("", $1); }
...
+ * "test.enum.mood:BigintEnum(test.enum.mood{"CURIOUS":2, "HAPPY":0})"
+ * where "test.enum.mood" is the enum name, "BigintEnum" is the enum kind,
+ * and "CURIOUS":2, "HAPPY":0 are the enum values.
+ * These values are passed as parameters to BIGINT_ENUM type.
+ */
+enum_map_entries : enum_map_entry { $$ = $1; }
+            | enum_map_entries COMMA enum_map_entry { $$ = $1 + ", " + $3; }
+            ;
+
+enum_map_entry : QUOTED_ID COLON SIGNED_INT {  $$ = $1 + ": " + std::to_string($3); }
```
kewang1024: maybe we don't need the extra space after ":"?

velox/type/parser/TypeParser.yy
```diff
@@ -149,6 +152,36 @@ named_type : type_single_word        { $$ = std::make_pair("", $1); }
...
+               | QUOTED_ID COLON NUMBER     {  $$ = $1 + ": " + std::to_string($3); }
+               ;
+
+enum_kind : WORD { if ($1 != "BigintEnum" && $1 != "VarcharEnum" )
+                    {
+                        std::string msg = "Invalid type " + $1 + ", expected BigintEnum or VarcharEnum";
+                        error(msg.c_str());
+                    }
+                $$ = $1; }
+                ;
```
kewang1024: remove ";" to be consistent with below?

HeidiHan0000: Added ; to the below lines to make it consistent with the rest!

velox/type/parser/tests/TypeParserTest.cpp
```diff
@@ -367,5 +368,69 @@ TEST_F(TypeParserTest, fieldNames) {
            ARRAY(BIGINT())}));
 }
 
+TEST_F(TypeParserTest, enumTest) {
+  registerBigintEnumType();
+
+  auto enumMoodString =
```
mbasmanova: enumMoodString -> mood

velox/type/parser/tests/TypeParserTest.cpp
```diff
@@ -367,5 +368,69 @@ TEST_F(TypeParserTest, fieldNames) {
...
+  auto otherEnumString =
+      "someEnumType:BigintEnum(someEnumType{\"CURIOUSs\": -2, \"HAPPY\": 3})";
+  ASSERT_EQ(
+      parseType(otherEnumString)->toString(),
+      "someEnumType:BigintEnum(someEnumType{\"CURIOUSs\": -2, \"HAPPY\": 3})");
+
+  // Invalid enum type
+  auto invalidString1 =
+      "test.enum.mood:SmallintEnum(test.enum.mood{\"CURIOUS\":-2, \"HAPPY\":0})";
+  VELOX_ASSERT_THROW(
```
mbasmanova: Let's move negative tests into a separate test method.

velox/type/parser/tests/TypeParserTest.cpp
```diff
@@ -367,5 +368,69 @@ TEST_F(TypeParserTest, fieldNames) {
...
+      "Failed to parse type [test.enum.mood:SmallintEnum(test.enum.mood{\"CURIOUS\":-2, \"HAPPY\":0})]. Invalid type SmallintEnum, expected BigintEnum or VarcharEnum");
+
+  // Invalid format: missing ":"
+  auto invalidString2 = "testNoColon(test.enum.mood{“CURIOUS”:-2, “HAPPY”:0})";
+  VELOX_ASSERT_THROW(
+      parseType(invalidString2),
+      "Failed to parse type [testNoColon(test.enum.mood{“CURIOUS”:-2, “HAPPY”:0})]. syntax error, unexpected LBRACE, expecting COLON");
+
+  // Invalid format: missing "("
+  auto invalidString3 =
```
mbasmanova: Do we need these variables? Names like a1, a2, a3 are anti-pattern.

velox/type/parser/ParserUtil.cpp
```diff
@@ -65,4 +65,37 @@ std::pair<std::string, std::shared_ptr<const Type>> inferTypeWithSpaces(
...
+
+  for (const auto& item : obj.items()) {
+    std::string key = item.first.asString();
+    VELOX_CHECK(
+        item.second.isInt(),
+        "Invalid value {} for key {} in enum type",
+        item.second,
+        key);
+    int64_t value = item.second.asInt();
+    result.emplace(std::move(key), value);
```
mbasmanova: Add a check for duplicates.

velox/type/parser/ParserUtil.h
```diff
@@ -31,6 +31,11 @@ TypePtr customTypeWithChildren(
     const std::string& name,
     const std::vector<TypePtr>& children);
 
+TypePtr getEnumType(
```
mbasmanova: Please, document all APIs in header files.

velox/type/parser/tests/TypeParserTest.cpp
```diff
@@ -367,5 +369,84 @@ TEST_F(TypeParserTest, fieldNames) {
            ARRAY(BIGINT())}));
 }
 
+TEST_F(TypeParserTest, enumTest) {
```
mbasmanova: drop "Test" suffix; ditto other places

mbasmanova (commented):
Some comments.

velox/type/parser/ParserUtil.cpp
HeidiHan0000: Thanks for the note. I noticed that folly::parseJson will silently override duplicate keys when parsing JSON maps. So instead, I formatted the map as an array of pairs from the bison parser file so that the same folly::parseJson can be used to also detect and throw on duplicates. Please let me know if manually parsing would look cleaner than manipulating the input to use folly::parseJson. I have left comments explaining this and tests to check for duplicates

mbasmanova: Thanks, but I don't see these changes.

PR #14279, opened 2025-07-30
misc: Remove duplicate registration for Spark url function

zhli1142015:
Same functions are also registered in https://github.com/facebookincubator/velox/blob/f2db819d053f4e7092efbef873ba23ea885fd1c3/velox/functions/sparksql/registration/RegisterUrl.cpp#L22


zhli1142015:
cc @rui-mo and @jinchengchenghh  thanks.

jinchengchenghh (approved):

PR #14280, opened 2025-07-30
fix: Add ST_Numpoints to fuzzer exclusion lists

Sullivan-Patrick:
Summary: Forgot to do this on D78748278.

Reviewed By: jagill

Differential Revision: D79234143


HeidiHan0000 (approved):
Thanks! Also is "Unrecognized geometry type: 123" a legit error and do you plan to fix that?

kagamiori (approved):

Sullivan-Patrick:
> Thanks! Also is "Unrecognized geometry type: 123" a legit error and do you plan to fix that?

This type of issue is why we're excluding Geometry functions from the fuzzer (since it's challenging to make an input generator for Geometries). This failure happens due to a random Varbinary input which shouldn't be possible in the real world.

PR #14281, opened 2025-07-30
refactor(abfs): Separate client creations for different authentications and add AzureClientProvider interface

marin-ma:
1. Currently the read/write clients for abfs are created from the `AbfsConfig`.
This patch introduces the `AzureClientProvider` interface for creating 
read/write client, and the part of logic in `AbfsConfig` to handle the auth 
configurations are extracted out. 

2. Rename AbfsConfig to AbfsPath, which only preserves the logic that 
parses the abfs file path.

3. Separate the client creations for different auth types.

4. Add `AzureClientProviderFactories` to support registering the 
creator functions of `AzureClientProvider` for different accounts.

Part of the changes in this patch are extracted from https://github.com/facebookincubator/velox/pull/14045


marin-ma:
@zhli1142015 Could you help to review? Thanks!

velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h
```diff
@@ -0,0 +1,40 @@
...
+
+/// Provider interface for creating Azure Blob and Data Lake clients.
+class AzureClientProvider {
+ public:
+  virtual ~AzureClientProvider() = default;
+
+  explicit AzureClientProvider(const std::shared_ptr<AbfsPath>& path)
+      : abfsPath_(path) {}
+
+  virtual std::unique_ptr<AzureBlobClient> getBlobClient() = 0;
```
zhli1142015: Can we still use the original names, `getReadFileClient` and `getWriteFileClient`?

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,110 @@
...
+    const std::shared_ptr<AbfsPath>& abfsPath,
+    const config::ConfigBase& config) {
+  auto authTypeKey = fmt::format(
+      "{}.{}", kAzureAccountAuthType, abfsPath->accountNameWithSuffix());
+  std::string authType = kAzureSharedKeyAuthType;
+  if (config.valueExists(authTypeKey)) {
+    authType = config.get<std::string>(authTypeKey).value();
+  }
+
+  if (authType == kAzureSharedKeyAuthType) {
```
zhli1142015: nit: please rewrite these if statements using a switch statement.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+} // namespace
+
+void AzureClientProviderFactories::registerFactory(
+    const std::string& account,
+    const AzureClientProviderFactory& factory) {
+  azureClientFactoryRegistry().withWLock([&](auto& factories) {
+    auto it = factories.find(account);
+    VELOX_USER_CHECK(
```
zhli1142015: If the factory is already registered, should we overwrite it instead of throwing an error?

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return factory(abfsPath, config)->getBlobClient();
+  }
+  return defaultAzureClientProviderFactory(abfsPath, config)->getBlobClient();
+}
+
+std::unique_ptr<AzureDataLakeFileClient>
+AzureClientProviderFactories::getDataLakeFileClient(
+    const std::shared_ptr<AbfsPath>& abfsPath,
+    const config::ConfigBase& config) {
+  if (clientFactoryRegistered(abfsPath->accountName())) {
```
zhli1142015: nit: maybe implment the `clientFactoryRegistered` as `std::optional<> getClientFactory`.

velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h
zhli1142015: Shouldn’t the interface here be:
```
getReadFileClient(
    const std::shared_ptr<AbfsPath>& abfsPath,
    const config::ConfigBase& config);
```

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
zhli1142015: Should we skip checking the auth type here, and instead register different factories for each auth type when we register the factory?

velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h
marin-ma: `abfsPath` and `config` are passed to the constructor of each `AzureClientProvider` implementation, so we don't need to pass them again here.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
marin-ma: We can add an "overwrite" argument here to control the behaviour.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
marin-ma: The factories are registered for a specific account. If no factory was registered from the caller/user, then it checks and creates the factory from config.

Do you mean registering the factory for the account in this check to avoid entering the check next time?

marin-ma: Seems like the if statements cannot be avoided because the value of `authType` is string type.

marin-ma:
@zhli1142015 Could you help to review again? Thanks!

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
zhli1142015: My understanding is that at most one AzureClientProvider is registered in the factories for each account. Which provider should be used ought to be decided by the registration logic, not by the factories. I also don’t think we need to provide a defaultAzureClientProviderFactory here — we should register the appropriate provider in Gluten based on the supplied configs.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
zhli1142015: Yes we can add "overwrite" argument, but I think it should always be true. Do you have an example where it needs to be false?

velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h
zhli1142015: The AzureClientProvider is registered per account, but it looks like we're binding it to a path — could that be a problem? I suggest changing to the function signature shown above.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,112 @@
...
+  return azureClientFactoryRegistry().withRLock(
+      [&](const auto& factories) -> std::optional<AzureClientProviderFactory> {
+        if (auto it = factories.find(account); it != factories.end()) {
+          return it->second;
+        }
+        return std::nullopt;
+      });
+}
+
+std::unique_ptr<AzureBlobClient> AzureClientProviderFactories::getBlobClient(
```
zhli1142015: Please update the function name here to getReadFileClient also.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
marin-ma: If the provider to use fully relies on the caller's registration, should this configuration `fs.azure.account.auth.type` be removed?

marin-ma:
@zhli1142015  Could you help to review again? Thanks!

Seems like with the current change, `fs.azure.account.auth.type` can be removed.

velox/connectors/hive/storage_adapters/abfs/tests/AbfsFileSystemTest.cpp
```diff
@@ -174,31 +203,31 @@ TEST_F(AbfsFileSystemTest, fileHandleWithProperties) {
...
-      readData(readFile.get());
-    });
-    threads.emplace_back(std::move(thread));
-  }
-  startThreads = true;
-  for (auto& thread : threads) {
-    thread.join();
-  }
-}
+// TEST_F(AbfsFileSystemTest, multipleThreadsWithReadFile) {
```
zhli1142015: Why you need to comment this test?

velox/connectors/hive/storage_adapters/abfs/AbfsConfig.h
```diff
@@ -61,70 +56,48 @@ static constexpr const char* kAzureOAuthAuthType = "OAuth";
...
-
-  std::unique_ptr<BlobClient> getReadFileClient();
+// For performance, re-use SAS tokens until the expiry is within this number of
+// seconds
+static constexpr const char* kAzureSasTokenRenewPeriod =
+    "fs.azure.sas.token.renew.period.for.streams";
 
-  std::unique_ptr<AzureDataLakeFileClient> getWriteFileClient();
+/// Helper class to parse and extract information from a given ABFS path.
+class AbfsPath {
```
zhli1142015, line 58: nit: can you rename this file to AbfsPath?

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
zhli1142015: We still require this configuration — we will use it to register different providers for each account. We also need to implement the static void registerProvider(ConfigBase config) method. We should move the logic above into this method and call it from the appropriate place in Gluten.

velox/connectors/hive/storage_adapters/abfs/tests/AbfsFileSystemTest.cpp
marin-ma: unintentional change

marin-ma:
@zhli1142015 Could you help to review again? Thanks!

velox/connectors/hive/storage_adapters/abfs/AbfsPath.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h"
+
+#include <azure/identity/client_secret_credential.hpp>
```
zhli1142015: nit:  I think this include is not needed.

velox/connectors/hive/storage_adapters/abfs/AbfsPath.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+
+#include "velox/common/config/Config.h"
```
zhli1142015: same here.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,90 @@
...
+      account);
+}
+
+} // namespace
+
+void AzureClientProviderFactories::registerFactory(
+    const std::string& account,
+    const AzureClientProviderFactory& factory) {
+  azureClientFactoryRegistry().withWLock([&](auto& factories) {
+    auto it = factories.find(account);
```
zhli1142015: nit: I think this line is not needed.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+std::optional<AzureClientProviderFactory>
+AzureClientProviderFactories::getClientFactory(const std::string& account) {
+  return azureClientFactoryRegistry().withRLock(
+      [&](const auto& factories) -> std::optional<AzureClientProviderFactory> {
+        if (auto it = factories.find(account); it != factories.end()) {
+          return it->second;
+        }
+        return std::nullopt;
+      });
```
zhli1142015, line 60: nit: maybe call factoryNotRegistered in this method, and let it returns AzureClientProviderFactory instead of std::optional<AzureClientProviderFactory>

velox/connectors/hive/storage_adapters/abfs/AzureDataLakeFileClient.h
```diff
@@ -44,6 +44,6 @@ class AzureDataLakeFileClient {
   virtual void append(const uint8_t* buffer, size_t size, uint64_t offset) = 0;
   virtual void flush(uint64_t position) = 0;
   virtual void close() = 0;
-  virtual std::string getUrl() const = 0;
+  virtual std::string getUrl() = 0;
```
zhli1142015, line 47: Why the `const` is removed?

marin-ma: The url may not be a const value and depends on the implementation of the `AzureClientProvider`. For example if the sasToken is dynamically updated, the url will also be updated. https://github.com/facebookincubator/velox/pull/14045/commits/c95e567128710c44cfb4a11ad6c694879cc3a923#diff-e1d939ff0d59569c553be7d697c6900142ebc10c10ed30d31e071a9f63559727R111-R113

marin-ma:
@zhli1142015 Could you help to review again? Thanks!

marin-ma:
@kagamiori Could you help to take a look and import this patch? Thanks!

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.h
```diff
@@ -16,9 +16,35 @@
 
 #pragma once
 
+#include <functional>
+#include <memory>
+#include <string>
+
+namespace facebook::velox::config {
+class ConfigBase;
```
rui-mo, line 25: nit: make it surrounded it by empty lines.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -60,4 +65,53 @@ void registerAbfsFileSystem() {
 #endif
 }
 
+void registerAzureClientProvider(const config::ConfigBase& config) {
+#ifdef VELOX_ENABLE_ABFS
+  for (const auto& [key, value] : config.rawConfigs()) {
+    constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
+    if (key.find(authTypePrefix) == 0) {
+      std::string_view skey = key;
+      // extract the accountName after "fs.azure.account.auth.type.".
```
rui-mo: nit: capitalize the first for comments.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -60,4 +65,53 @@ void registerAbfsFileSystem() {
...
+      } else if (value == kAzureOAuthAuthType) {
+        AzureClientProviderFactories::registerFactory(
+            accountName, [](const std::string&) {
+              return std::make_unique<OAuthAzureClientProvider>();
+            });
+      } else if (value == kAzureSASAuthType) {
+        AzureClientProviderFactories::registerFactory(
+            accountName, [](const std::string&) {
+              return std::make_unique<FixedSasAzureClientProvider>();
+            });
```
rui-mo, line 99: Might could extract the lambda as a variable to handle different types, and call the `AzureClientProviderFactories::registerFactory` once.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.h
```diff
@@ -0,0 +1,104 @@
...
+#include "velox/connectors/hive/storage_adapters/abfs/AzureBlobClient.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureDataLakeFileClient.h"
+
+#include <memory>
+#include <string>
+
+namespace facebook::velox::filesystems {
+
+/// AzureClientProvider for Shared Key authentication.
```
rui-mo: nit: use // for single-line comments. ditto for the others.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.h
```diff
@@ -0,0 +1,104 @@
...
+#pragma once
+
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureBlobClient.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureDataLakeFileClient.h"
+
+#include <memory>
+#include <string>
```
rui-mo: Are these two includes needed?

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.cpp
```diff
@@ -0,0 +1,238 @@
...
+  VELOX_USER_CHECK(
+      config.valueExists(clientSecretKey),
+      "Config {} not found",
+      clientSecretKey);
+  VELOX_USER_CHECK(
+      config.valueExists(clientEndpointKey),
+      "Config {} not found",
+      clientEndpointKey);
+  auto clientEndpoint = config.get<std::string>(clientEndpointKey).value();
+  auto firstSep = clientEndpoint.find_first_of("/", /* https:// */ 8);
```
rui-mo: nit: use the /\*argName=\*/value format (note the lack of spaces).

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.h
```diff
@@ -0,0 +1,67 @@
...
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureBlobClient.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureClientProvider.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AzureDataLakeFileClient.h"
+
+#include <functional>
+#include <memory>
+#include <optional>
+#include <string>
```
rui-mo: Perhaps verify and remove the unnecessary includes.

velox/connectors/hive/storage_adapters/abfs/AbfsPath.cpp
```diff
@@ -0,0 +1,60 @@
...
+  accountNameWithSuffix_ = file.substr(firstAt + 1, firstSep - firstAt - 1);
+  auto firstDot = accountNameWithSuffix_.find_first_of(".");
+  accountName_ = accountNameWithSuffix_.substr(0, firstDot);
+}
+
+std::string AbfsPath::getUrl(bool withblobSuffix) const {
+  std::string accountNameWithSuffixForUrl(accountNameWithSuffix_);
+  if (withblobSuffix) {
+    // We should use correct suffix for blob client.
+    size_t start_pos = accountNameWithSuffixForUrl.find("dfs");
```
rui-mo: nit: use camel-case style for variable name.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.h
marin-ma: <memory> is for `std::unique_ptr` and <string> is for `std::string`

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.cpp
marin-ma: This inline comment is not an arg name.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.h
rui-mo: I assume \<string\> has been pulled in by `AbfsPath.h`, not sure about \<memory\>.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.cpp
marin-ma: Changed the magic number to a local variable

```
 static const std::size_t kHttpsPrefixLen = 8;
```

marin-ma:
@rui-mo Could you help to review again? Thanks!

rui-mo:
@pedroerp @majetideepak Do you have any further comment on this refactor? Thanks!

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -60,4 +76,44 @@ void registerAbfsFileSystem() {
...
+void registerAzureClientProvider(const config::ConfigBase& config) {
+#ifdef VELOX_ENABLE_ABFS
+  for (const auto& [key, value] : config.rawConfigs()) {
+    constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
+    if (key.find(authTypePrefix) == 0) {
+      std::string_view skey = key;
+      // Extract the accountName after "fs.azure.account.auth.type.".
+      auto remaining = skey.substr(authTypePrefix.size() + 1);
+      auto dot = remaining.find(".");
+      VELOX_CHECK_NE(
```
majetideepak: nit: `VELOX_USER_CHECK_NE`

velox/connectors/hive/storage_adapters/abfs/AzureClientProviders.h
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
majetideepak: nit: Let's name this `AzureClientProviderImpl.h/.cpp` to make it clear that the implementations are in this file.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -14,16 +14,32 @@
...
 
 namespace facebook::velox::filesystems {
 
 #ifdef VELOX_ENABLE_ABFS
+
+namespace {
+
+template <typename T>
+void registerProviderFactory(const std::string& accountName) {
+  AzureClientProviderFactories::registerFactory(
```
majetideepak: This is a single-line code function. Let's inline the body below to make it easy to read.

velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.cpp
```diff
@@ -0,0 +1,76 @@
...
+      factories;
+  return factories;
+}
+
+} // namespace
+
+void AzureClientProviderFactories::registerFactory(
+    const std::string& account,
+    const AzureClientProviderFactory& factory) {
+  azureClientFactoryRegistry().withWLock(
```
majetideepak: We should log if a factory is being overridden for a specific account. It will help aid debugging.

majetideepak (commented):
Some minor comments.

marin-ma:
@majetideepak Could you help to review again? Thanks!

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -14,16 +14,22 @@
  * limitations under the License.
  */
 
+#include "velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.h" // @manual
```
majetideepak, line 17: nit: This include is not required.

majetideepak:
@marin-ma can you rebase and fix the conflicts?

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
marin-ma: It's required here for the function signature.
```
void registerAzureClientProviderFactory(
    const std::string& account,
    const AzureClientProviderFactory& factory)
```

marin-ma:
@majetideepak Conflicts resolved. Could you help to review again? Thanks!

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
majetideepak: Does it fail if you remove this header? You need the header only if there is a use of this function in this file.

majetideepak: You need it for the `AzureClientProviderFactory` definition.

PR #14286, opened 2025-07-30
docs: Add monthly update for July 2025

mbasmanova:
Differential Revision: D79301147


mbasmanova:
> Nice, thank you. Can we have docs somewhere on how you generated this?

It would be good indeed. I used Claude Code with the following prompts.

```
# Prompts Used for July 2025 Monthly Update

## Prompt 1
I need to generate a monthly update for the changes done in the past month in this project. I need rst file similar to velox/docs/monthly-updates/may-2024.rst. First, look at this file and tell ne that you undersatnd the structure.

## Prompt 2
OK, look at git log changes in the past month, and try to categorize the changes into similar areas. You may need to look at commit description in addition to titles.

## Prompt 3
This is a good start. Now assume that the target audience is developers who do not work on this day to day but want to understand important changes added in the given period, like adding new functionality, fixing critical bugs, major platform upgrades, etc. Commits like build errors, tests, typos, logging are not important. Also within each category, I would like to see all additions first, then bug/issue fixes, then optimizations.

## Prompt 4
You need to follow the format of the rst file I mentioned, there are no subsections.

## Prompt 5
The example RST file had back ticks around PR numbers, and function names. Could you update it?

## Prompt 6
Store this in the file with the appropriate name

## Prompt 7
One NIT: geometry functions belong to presto functions

## Prompt 8
Can you also create two files, one that just has the list of all prompts I provided, and another is instructions to future you, so you can repeat the same process next time. Include this last prompt as well to the log.
```

PR #14287, opened 2025-07-30
fix: Reduce memory usage of constant expression

Yuhta:
Summary:
In `ConstantExpr` we are making deep copy of the constant vector.  This
was done to avoid multi-threaded (MT) access to `computeAndSetIsAscii`, but that
function was changed to be MT-safe later, so we can drop the deep copy to reduce
memory usage and initialization cost.

Differential Revision: D79177296

### DIFF
--- a/velox/expression/tests/ExpressionRunnerUnitTest.cpp
+++ b/velox/expression/tests/ExpressionRunnerUnitTest.cpp
@@ -131,9 +131,8 @@ TEST_F(ExpressionRunnerUnitTest, persistAndReproComplexSql) {
   auto reproExprs = ExpressionRunner::parseSql(
       reproSql, nullptr, pool_.get(), reproComplexConstants);
   ASSERT_EQ(reproExprs.size(), 1);
-  // Note that ConstantExpr makes a copy of sharedConstantValue_ to guard
-  // against race conditions, which in effect falttens the array.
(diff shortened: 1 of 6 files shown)


velox/expression/tests/ExpressionRunnerUnitTest.cpp
```diff
@@ -133,7 +133,8 @@ TEST_F(ExpressionRunnerUnitTest, persistAndReproComplexSql) {
   ASSERT_EQ(reproExprs.size(), 1);
   // Note that ConstantExpr makes a copy of sharedConstantValue_ to guard
   // against race conditions, which in effect falttens the array.
-  ASSERT_EQ(reproExprs[0]->toString(), "{3, 5, 1, 2}");
+  ASSERT_EQ(
+      reproExprs[0]->toString(), "{[0->2] 3, [1->4] 5, [2->0] 1, [3->1] 2}");
```
mbasmanova: Can we avoid hard-coding toString output? Can we instead check that the expression is constant and holds the right value?

Yuhta: I think the point of this test was to check `saveVectorToFile` keeps exactly structure of the input vectors (including the encodings and how indices are layered out in dictionary), so a string to string comparison is reasonable in this case.  In this sense #5035 was breaking this property and this change just restores it back.

Issue #14262, opened 2025-07-29
Fast path for Spark split function


NEUpanning:
### Description

If the delimiter is a single character and not a reserved character of regex, we can use `std::string::find()` instead of re2 library to locate the delimiter's index, which is likely more efficient.

Inspired by the split implementation of JDK. 
https://github.com/bpupadhyaya/openjdk-8/blob/master/jdk/src/share/classes/java/lang/String.java#L2317

NEUpanning:
Closed by [#14291](https://github.com/facebookincubator/velox/pull/14291)

PR #14291, opened 2025-07-31
feat(function): Add fast path for Spark split function

NEUpanning:
Add a fast path that locates the delimiter by iterating string instead of
using re2 library for Spark split function. The microbenchmark shows
there is a 5x performance gain.

```
[...]parksql/benchmarks/SplitBenchmark.cpp     relative  time/iter   iters/s
baseline_old                                                29.14s    34.32m
split_new                                       514.54%      5.66s   176.60m
```

Fixes [#14262](https://github.com/facebookincubator/velox/issues/14262)

### DIFF
--- a/velox/functions/sparksql/Split.h
+++ b/velox/functions/sparksql/Split.h
@@ -47,9 +47,13 @@ struct Split {
       const std::vector<TypePtr>& /*inputTypes*/,
       const core::QueryConfig& config,
       const arg_type<Varchar>* /*input*/,
-      const arg_type<Varchar>* /*delimiter*/,
+      const arg_type<Varchar>* delimiter,
       const arg_type<int32_t>* /*limit*/) {
     cache_.setMaxCompiledRegexes(config.exprMaxCompiledRegexes());
+    if (delimiter) {
+      initializeFastPath(delimiter->data(), delimiter->size());
+      isConstantDelimiter_ = true;
+    }
   }
 
   FOLLY_ALWAYS_INLINE void call(
@@ -72,12 +76,32 @@ struct Split {
       out_type<Array<Varchar>>& result,
       const arg_type<Varchar>& input,
       const arg_type<Varchar>& delimiter,
-      int32_t limit) const {
+      int32_t limit) {
     if (delimiter.empty()) {
       splitEmptyDelimiter(result, input, limit);
     } else {
       split(result, input, delimiter, limit);
     }
+  }
+
+  // Checks if a string represents an octal-escaped ascii character (e.g.,
+  // \123). If valid, converts it to its decimal value.
+  bool isOctalString(const char* str, char& decimal, size_t size) const {
+    if (size > 4 || str[0] != '\\') {
+      return false;
+    }
+    for (int i = 1; i < size; ++i) {
+      char ch = str[i];
+      if (ch < '0' || ch > '7') {
+        return false;
+      }
+    }
+    // The ascii octal number should not exceed 177.
+    if (std::stoi(str + 1) > 177) {
+      return false;
+    }
+    decimal = std::stoi(str + 1, nullptr, 8);
+    return true;
   }
 
   // When pattern is empty, split each character out. Since Spark 3.4, when
@@ -113,6 +137,33 @@ struct Split {
     }
   }
 
+  // Fast path if the delimiter is a
+  // (1)one-char String and this character is not one of the RegEx's meta
+  // characters ".$|()[{^?*+\\", or
+  // (2)two-char String and the first char is the backslash and the second is
+  // not the ascii digit or ascii letter, or
+  // (3)octal string.
+  void initializeFastPath(const char* delimiterStr, size_t delimiterSize) {
+    const std::string reservedChars = ".$|()[{^?*+\\";
+    if (delimiterSize == 1) {
+      singleCharDelimiter_ = delimiterStr[0];
+      if (reservedChars.find(singleCharDelimiter_) == std::string::npos) {
+        fastPath_ = true;
+      }
+    } else if (delimiterSize == 2 && delimiterStr[0] == '\\') {
+      singleCharDelimiter_ = delimiterStr[1];
+      if (!std::isdigit(singleCharDelimiter_) &&
+          !std::isalpha(singleCharDelimiter_)) {
+        fastPath_ = true;
+      }
+    } else if (isOctalString(
+                   delimiterStr, singleCharDelimiter_, delimiterSize)) {
+      fastPath_ = true;
+    } else {
+      fastPath_ = false;
+    }
+  }
+
   // Split with a non-empty delimiter. If limit > 0, The resulting array's
   // length will not be more than limit and the resulting array's last entry
   // will contain all input beyond the last matched regex. If limit <= 0,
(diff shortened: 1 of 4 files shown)


NEUpanning:
@rui-mo Could you help take a look? Thanks!

velox/functions/sparksql/Split.h
```diff
@@ -189,5 +255,8 @@ struct Split {
   }
 
   mutable facebook::velox::functions::detail::ReCache cache_;
+  bool fastPath_{false};
+  char ch_;
```
rui-mo: Would you please add description for the variable `ch_`? It looks not obvious.

velox/functions/sparksql/Split.h
```diff
@@ -72,14 +76,33 @@ struct Split {
...
-      int32_t limit) const {
+      int32_t limit) {
     if (delimiter.empty()) {
       splitEmptyDelimiter(result, input, limit);
     } else {
       split(result, input, delimiter, limit);
     }
   }
 
+  // Octal escaped character in the form \000.
```
rui-mo: Would you please revise the comment? Some example:

> Checks if a string represents an octal-escaped character (e.g., \123). If valid, converts it to its decimal value.

velox/functions/sparksql/Split.h
```diff
@@ -72,14 +76,33 @@ struct Split {
...
+      return false;
+    }
+    for (int i = 1; i < size; ++i) {
+      char ch = str[i];
+      if (ch < '0' || ch > '7') {
+        return false;
+      }
+    }
+    // Octal value must be less than 377.
+    if (std::stoi(str + 1) > 377) {
```
rui-mo: Can the value be equal to 377?

velox/functions/sparksql/Split.h
```diff
@@ -72,14 +76,33 @@ struct Split {
...
+      return false;
+    }
+    decimal = std::stoi(str + 1, nullptr, 8);
```
rui-mo, line 103: Would the octal value exceed 127, and is it safe to assign it to a char?

velox/functions/sparksql/Split.h
```diff
@@ -113,6 +136,32 @@ struct Split {
     }
   }
 
+  // Fast path if the delimiter is a
+  // (1)one-char String and this character is not one of the RegEx's meta
+  // characters ".$|()[{^?*+\\", or
```
rui-mo, line 142: Is there any reference for the characters?

velox/functions/sparksql/Split.h
```diff
@@ -113,6 +136,32 @@ struct Split {
...
+    const std::string reservedChars = ".$|()[{^?*+\\";
+    if (delimiterSize == 1) {
+      ch_ = delimiterStr[0];
+      if (reservedChars.find(ch_) == std::string::npos) {
+        fastPath_ = true;
+      }
+    } else if (delimiterSize == 2 && delimiterStr[0] == '\\') {
+      ch_ = delimiterStr[1];
+      if (((ch_ - '0') | ('9' - ch_)) < 0 && ((ch_ - 'a') | ('z' - ch_)) < 0 &&
+          ((ch_ - 'A') | ('Z' - ch_)) < 0) {
```
rui-mo: Would it improve the readability to use `!std::isdigit` and `!std::isalpha` for the checks?

rui-mo (commented):
Thanks for this optimization!

velox/functions/sparksql/Split.h
NEUpanning: In the implementation of JDK, Spark split function utilizes this.
https://github.com/bpupadhyaya/openjdk-8/blob/master/jdk/src/share/classes/java/lang/String.java#L2319

velox/functions/sparksql/Split.h
NEUpanning: Yes, 377 matches a ÿ character. I will change the comment.

velox/functions/sparksql/Split.h
NEUpanning: Thanks!

velox/functions/sparksql/Split.h
NEUpanning: Thanks! Fast path should only for single character delimiter, so the octal value should not exceed decimal 127. Updated.

rui-mo: > if (std::stoi(str + 1) > 177) {

Do you want to check against 127?

velox/functions/sparksql/Split.h
rui-mo: Would you please add the reference link in the comments?

velox/functions/sparksql/Split.h
```diff
@@ -113,6 +137,33 @@ struct Split {
     }
   }
 
+  // Fast path if the delimiter is a
+  // (1)one-char String and this character is not one of the RegEx's meta
```
rui-mo: nit: use lower case for `String` to be consistent.

velox/functions/sparksql/Split.h
```diff
@@ -113,6 +137,33 @@ struct Split {
...
+  // characters ".$|()[{^?*+\\", or
+  // (2)two-char String and the first char is the backslash and the second is
```
rui-mo: ditto

velox/functions/sparksql/benchmarks/SplitBenchmark.cpp
```diff
@@ -0,0 +1,108 @@
...
+BENCHMARK_RELATIVE(split, n) {
+  benchmark->run(n, SplitBenchmark::kSplitExpression);
+}
+
+} // namespace
+
+int main(int argc, char** argv) {
+  folly::Init init{&argc, &argv};
+  memory::MemoryManager::initialize(memory::MemoryManager::Options{});
+  LOG(ERROR) << "Seed: " << seed;
```
rui-mo: nit: might be better to use LOG(INFO)

rui-mo (commented):
Thanks! Just some nits.

velox/functions/sparksql/Split.h
rui-mo: It might be helpful to add a test for the decimal value overflow.

NEUpanning: > if (std::stoi(str + 1) > 177) {
Do you want to check against 127?

Check against decimal 127, which is octal 177.

NEUpanning:
@rui-mo Thanks for your review. The fuzzer test passed. Here is the command and test statistics.

`spark_expression_fuzzer_test --duration_sec=120 --logtostderr=1 --velox_fuzzer_enable_complex_types --only=split`

```
E20250808 15:02:31.572259 12361364 ExpressionFuzzerVerifier.cpp:259] split 30281 100.00% 918524
E20250808 15:02:31.572293 12361364 ExpressionFuzzerVerifier.cpp:488] Total test cases: 45208
E20250808 15:02:31.572309 12361364 ExpressionFuzzerVerifier.cpp:489] Total test cases verified in the reference DB: 23772
E20250808 15:02:31.574255 12361364 ExpressionFuzzerVerifier.cpp:491] Total test cases failed in both Velox and the reference DB: 21436
E20250808 15:02:31.574281 12361364 ExpressionFuzzerVerifier.cpp:493] Total test cases unsupported in the reference DB: 0
```

PR #14294, opened 2025-07-31
refactor(cudf): Add CudfHiveConnector

devavret:
Add CudfHiveConnector, which inherits from the HiveConnector.
We want to use the HiveConnector implementation as a base for the CudfConnector since it is well established, and it gives us the flexibility to fallback to CPU when needed.

This PR also disables cuDF based parquet writer for now to limit the scope of changes.

Fixes: #14183


velox/connectors/hive/HiveConnector.h
```diff
@@ -78,6 +80,23 @@ class HiveConnector : public Connector {
...
+      std::function<std::unique_ptr<facebook::velox::connector::DataSource>(
+          const RowTypePtr& /*outputType*/,
+          const ConnectorTableHandlePtr& /*tableHandle*/,
+          const connector::ColumnHandleMap& /*columnHandles*/,
+          FileHandleFactory* /*fileHandleFactory*/,
+          folly::Executor* /*executor*/,
+          ConnectorQueryCtx* /*connectorQueryCtx*/,
+          const std::shared_ptr<const HiveConfig>& /*hiveConfig*/)>;
+
+  static void registerDataSourceDelegate(DataSourceDelegate delegate);
```
Yuhta: Why do we need to do this?  Can we modify the plan directly to use the CuDF data source instead?

devavret: That's actually how it used to be but it's inconsistent with how the rest of the operators were replaced, which is via a driver adapter, without having to change the plan. Which is why @majetideepak originally requested this approach (offline) and then I made that into an issue https://github.com/facebookincubator/velox/issues/14183

devavret: Note that adding such hooks has precedent in https://github.com/facebookincubator/velox/blob/b61a68cb7ae009867a290af6664701f8a555a4d1/velox/connectors/hive/HiveDataSource.cpp#L542-L562
but this one has a different purpose.

karthikeyann: @Yuhta

Yuhta: Would you refactor Wave data source to use the same mechanism as well?  It's better to keep only one way to extend these things

devavret: Wave uses the hook to extract information out of an existing DataSource so that it can use that to create its own in a separate wave TableScan. Our approach is different: we modify the behaviour of the existing TableScan operator. I am hesitant to touch wave code to modify their approach.

majetideepak:
@devavret can you please rebase?

velox/connectors/hive/HiveConnector.h
```diff
@@ -20,6 +20,8 @@
 #include "velox/connectors/hive/HiveConfig.h"
 #include "velox/core/PlanNode.h"
 
+#include <functional>
```
majetideepak: clean up?

velox/experimental/cudf/connectors/parquet/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ add_library(
   OBJECT
   ParquetConfig.cpp
   ParquetConnector.cpp
```
majetideepak: Do we still need the ParquetConnector.h/.cpp?

velox/experimental/cudf/connectors/parquet/CudfHiveConnector.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+using namespace facebook::velox::connector;
+
+CudfHiveConnector::CudfHiveConnector(
+    const std::string& id,
+    std::shared_ptr<const facebook::velox::config::ConfigBase> config,
+    folly::Executor* executor)
+    : hive::HiveConnector(id, config, executor),
+      parquetConfig_(std::make_shared<ParquetConfig>(config)) {
+  LOG(INFO) << "cudf::Parquet connector " << connectorId() << " created.";
```
majetideepak: Let the client/user of `CudfHiveConnector` LOG this where necessary. Remove from here.

velox/experimental/cudf/connectors/parquet/CudfHiveConnector.cpp
```diff
@@ -0,0 +1,94 @@
...
+      tableHandle,
+      columnHandles,
+      &fileHandleFactory_,
+      executor_,
+      connectorQueryCtx,
+      hiveConfig_);
+}
+
+// TODO (dm): Add data sink
+// std::unique_ptr<DataSink> CudfHiveConnector::createDataSink(
```
majetideepak: Remove commented code.

velox/experimental/cudf/connectors/parquet/CudfHiveConnector.h
```diff
@@ -0,0 +1,80 @@
...
+      const RowTypePtr& outputType,
+      const ConnectorTableHandlePtr& tableHandle,
+      const ColumnHandleMap& columnHandles,
+      ConnectorQueryCtx* connectorQueryCtx) override final;
+
+  bool canAddDynamicFilter() const override {
+    return false;
+  }
+
+  // TODO (dm): What config to be returned from here?
```
majetideepak: Remove commented code.

velox/experimental/cudf/connectors/parquet/ParquetDataSource.cpp
```diff
@@ -262,8 +265,14 @@ void ParquetDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
       VELOX_CHECK_EQ(
           hiveSplit->fileFormat,
           dwio::common::FileFormat::PARQUET,
-          "Unsupported file format for conversion from HiveConnectorSplit to cuDF ParquetConnectorSplit");
-      return ParquetConnectorSplitBuilder(hiveSplit->filePath)
+          "Unsupported file format for conversion from HiveConnectorSplit to ParquetConnectorSplit");
+      // Remove "file:" prefix from the file path if present
+      std::string cleanedPath = hiveSplit->filePath;
+      constexpr std::string_view filePrefix = "file:";
```
majetideepak: Velox naming convention: `kFilePrefix`

velox/experimental/cudf/connectors/parquet/ParquetDataSource.h
```diff
@@ -62,6 +62,10 @@ class ParquetDataSource : public DataSource, public NvtxHelper {
     return completedRows_;
   }
 
+  const common::SubfieldFilters* getFilters() const override {
```
majetideepak, line 66: Is this used anywhere?

velox/experimental/cudf/exec/ExpressionEvaluator.cpp
```diff
@@ -24,6 +24,7 @@
 #include "velox/vector/ConstantVector.h"
 #include "velox/vector/VectorTypeUtils.h"
 
+#include <cudf/ast/expressions.hpp>
```
majetideepak: Do we need these new includes?

velox/experimental/cudf/tests/utils/ParquetConnectorTestBase.cpp
```diff
@@ -67,40 +71,41 @@ ParquetConnectorTestBase::ParquetConnectorTestBase() {
...
 
 void ParquetConnectorTestBase::TearDown() {
   // Make sure all pending loads are finished or cancelled before unregister
   // connector.
   ioExecutor_.reset();
-  facebook::velox::connector::unregisterConnector(kParquetConnectorId);
-  facebook::velox::connector::unregisterConnectorFactory(
-      facebook::velox::cudf_velox::connector::parquet::ParquetConnectorFactory::
-          kParquetConnectorName);
+  facebook::velox::connector::unregisterConnector("hive");
```
majetideepak: nit: use a constant `KCudfHiveConnectorId = "hive"` since it is used at multiple places.

velox/experimental/cudf/connectors/parquet/ParquetDataSource.h
devavret: Looks like the only place this is used is https://github.com/facebookincubator/velox/blob/94fbcae7e967ab28ba90893c8c8e2998b6fb12df/velox/exec/TableScan.cpp#L36

velox/experimental/cudf/connectors/hive/CudfHiveConnector.cpp
```diff
@@ -0,0 +1,74 @@
...
+    : ::facebook::velox::connector::hive::HiveConnector(id, config, executor),
+      cudfHiveConfig_(std::make_shared<CudfHiveConfig>(config)) {}
+
+std::unique_ptr<DataSource> CudfHiveConnector::createDataSource(
+    const RowTypePtr& outputType,
+    const ConnectorTableHandlePtr& tableHandle,
+    const ColumnHandleMap& columnHandles,
+    ConnectorQueryCtx* connectorQueryCtx) {
+  // If it's parquet then return CudfHiveDataSource
+  // If it's not parquet then return HiveDataSource
```
majetideepak, line 40: We can handle this inside `createSplitReader`. If the split is not Parquet, we return a HiveSplitReader.

majetideepak (approved):

majetideepak:
@karthikeyann, @bdice, any feedback on this?

devavret:
> @karthikeyann, @bdice, any feedback on this?

Most importantly @mhaseeb123

majetideepak:
We want to use the HiveConnector implementation as a base for the CudfConnector since it is well established, and it gives us the flexibility to fallback to CPU when needed.

In this refactor, we are planning to process the entire file if the HiveSplit contains start offset 0. This is because the cudf Parquet reader only works with the row numbers.

In the future, for the HiveConnectorSplit (fileByteStart, fileByteLength) to work with Cudf parquet reader (rowStart, rowCount), we need to extract the metadata first. We can use the cudf.io.read_parquet_metadata to get the RowGroup offset metadata and see which RowGroups lie in the range of the HiveSplit. We then can use the RowGroup row counts to translate. cuDF team should provide guidance on this.

velox/experimental/cudf/connectors/hive/CudfHiveConnector.h
```diff
@@ -45,29 +43,26 @@ class ParquetConnector final : public Connector {
...
-      ConnectorQueryCtx* connectorQueryCtx,
-      CommitStrategy commitStrategy) override final;
-
-  folly::Executor* executor() const override {
-    return executor_;
+  bool canAddDynamicFilter() const override {
+    return false;
   }
 
+  // TODO (dm): Re-add data sink
```
mhaseeb123, line 50: I understand the comment but is there a particular reason to remove the data sink from here in the first place? is it to limit the scope of the PR?

velox/experimental/cudf/exec/CMakeLists.txt
```diff
@@ -31,7 +31,14 @@ add_library(
...
   PUBLIC cudf::cudf
-  PRIVATE arrow velox_arrow_bridge velox_exception velox_common_base velox_cudf_vector velox_exec
+  PRIVATE
+    arrow
+    velox_arrow_bridge
+    velox_exception
+    velox_common_base
+    velox_cudf_vector
+    velox_exec
+    velox_cudf_parquet_connector
```
mhaseeb123: shouldn't this be named as `velox_cudf_hive_connector`

mhaseeb123:
>We can use the cudf.io.read_parquet_metadata to get the RowGroup offset metadata and see which RowGroups lie in the range of the HiveSplit. We then can use the RowGroup row counts to translate. cuDF team should provide guidance on this.

@majetideepak `cudf.io.read_parquet_metadata` probably won't work here as it only returns a limited amount of metadata info but I think we can probably add methods to read and reuse parquet metadata.

velox/experimental/cudf/connectors/hive/CMakeLists.txt
```diff
@@ -12,26 +12,26 @@
...
+  CudfHiveConfig.cpp
+  CudfHiveConnector.cpp
+  CudfHiveConnectorSplit.cpp
+  CudfHiveDataSource.cpp
+  CudfHiveDataSink.cpp
+  CudfHiveTableHandle.cpp
 )
 
-set_target_properties(velox_cudf_parquet_connector PROPERTIES CUDA_ARCHITECTURES native)
+set_target_properties(velox_hive_parquet_connector PROPERTIES CUDA_ARCHITECTURES native)
```
karthikeyann:
```suggestion
set_target_properties(velox_cudf_hive_connector PROPERTIES CUDA_ARCHITECTURES native)
```

velox/experimental/cudf/connectors/hive/CMakeLists.txt
```diff
@@ -12,26 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_library(velox_cudf_parquet_config ParquetConfig.cpp)
+add_library(velox_cudf_parquet_config CudfHiveConfig.cpp)
```
mhaseeb123:
```suggestion
add_library(velox_cudf_hive_config CudfHiveConfig.cpp)
```

velox/experimental/cudf/connectors/hive/CMakeLists.txt
```diff
@@ -12,26 +12,26 @@
...
 
 set_target_properties(velox_cudf_parquet_config PROPERTIES CUDA_ARCHITECTURES native)
```
mhaseeb123:
```suggestion
set_target_properties(velox_cudf_hive_config PROPERTIES CUDA_ARCHITECTURES native)
```

velox/experimental/cudf/connectors/hive/CMakeLists.txt
```diff
@@ -12,26 +12,26 @@
...
 
 target_link_libraries(velox_cudf_parquet_config velox_core velox_exception cudf::cudf)
```
mhaseeb123:
```suggestion
target_link_libraries(velox_cudf_hive_config velox_core velox_exception cudf::cudf)
```

mhaseeb123: Let's double check to ensure we are using the updated library names across other cmake files that maybe linking them (TableScanTests?)

velox/experimental/cudf/connectors/hive/CudfHiveConnector.h
devavret: Yes exactly.

devavret:
> > We can use the cudf.io.read_parquet_metadata to get the RowGroup offset metadata and see which RowGroups lie in the range of the HiveSplit. We then can use the RowGroup row counts to translate. cuDF team should provide guidance on this.
> 
> @majetideepak `cudf.io.read_parquet_metadata` probably won't work here as it only returns a limited amount of metadata info but I think we can probably add methods to read and reuse parquet metadata.

Besides, do we need to rely on cuDF to read the metadata? Arrow has an API to read file metadata. We can iterate over row groups and use this one to find if this one lies in range. https://github.com/apache/arrow/blob/916f62df7bedd40a4847306dc2be3265ee647c02/cpp/src/parquet/metadata.h#L228

karthikeyann:
Approved.

majetideepak:
@devavret can you fix the conflict with main branch by rebasing?

majetideepak:
@xiaoxmeng can you please help import this? Thanks!

PR #14298, opened 2025-07-31
build: Use uv in docs workflow and improve ccaching

assignUser:
Follow up to #14046


assignUser:
@kgpai I rebased to make sure everything still works

PR #14303, opened 2025-08-01
fix: Disable compressed write for Text Writer

YiChengLee03:
Summary: As title

Differential Revision: D79405675


kewang1024 (approved):

kewang1024:
NIT: Disable compressed write for text writer

PR #14305, opened 2025-08-01
Add fb_truncated_mean native function

natashasehgal:
Summary: Add native implementation for fb_truncated_mean native

Differential Revision: D79102738


velox/functions/lib/QuantileDigest.h
```diff
@@ -109,6 +109,16 @@ class QuantileDigest {
           CdfEntry>>
   getDistributionFunction(double rangeStart, double rangeEnd) const;
 
+  // Calculate the truncated mean between quantile bounds
+  // Returns null when empty
```
kagamiori: Hi @natashasehgal, could you add more explanation in the comment about what "truncated mean" is? Maybe add a small example if that would help understand it.

natashasehgal: Makes sense! Enhanced description and added a basic example

velox/functions/lib/QuantileDigest.h
```diff
@@ -1371,6 +1386,77 @@ QuantileDigest<T, Allocator>::getDistributionFunction(
   return cdf;
 }
 
+template <typename T, typename Allocator>
+template <typename ValueFunc>
+std::optional<double> QuantileDigest<T, Allocator>::getTruncatedMean(
+    double lowerQuantileBound,
+    double upperQuantileBound,
+    ValueFunc valueFunc) const {
```
kagamiori: What would be the ValueFunc in a real use case? Are we expecting different ValueFunc for different UDFs?

velox/functions/lib/QuantileDigest.h
```diff
@@ -1371,6 +1386,77 @@ QuantileDigest<T, Allocator>::getDistributionFunction(
   return cdf;
 }
 
+template <typename T, typename Allocator>
+template <typename ValueFunc>
+std::optional<double> QuantileDigest<T, Allocator>::getTruncatedMean(
```
kagamiori: To make this API more general, how about making it return a list of values and their nodeCounts for values within the [lowerQuantileBound, upperQuantileBound] range, and let the UDF decide what it would compute (e.g., mean, medium, stddev, etc.)?

velox/functions/lib/QuantileDigest.h
```diff
@@ -109,6 +109,21 @@ class QuantileDigest {
...
   getDistributionFunction(double rangeStart, double rangeEnd) const;
 
+  /// Get the approx truncated mean from the digest.
+  /// Computes the mean of values between the specified quantile bounds,
+  /// excluding outliers.
+  ///
+  /// Example: For data [1, 10, 11, 12, 100] with bounds [0.25, 0.75], this
+  /// excludes the outliers (1, 100) and computes the mean of the middle values
+  /// as 11.
+  template <typename ValueFunc = std::function<double(T)>>
```
kagamiori: Thanks for the example!

velox/functions/lib/QuantileDigest.h
natashasehgal: In this case the argument and return types are simple so I removed it

velox/functions/lib/QuantileDigest.h
natashasehgal: That's a very good idea. I did some refactoring so the other statistical types can use the core logic here. I've still kept the UDF straightforward and development isn't planned on TDigest/QDigest data types, it's mostly for existing workloads. How do the new changes look?

velox/functions/lib/QuantileDigest.h
```diff
@@ -1371,6 +1403,88 @@ QuantileDigest<T, Allocator>::getDistributionFunction(
...
+
+        double amountOverLower = std::max(sum - lowerRank, 0.0);
+        double amountOverUpper = std::max(sum - upperRank, 0.0);
+        // Constrain node count to the rank of the lower and upper bound
+        // quantiles
+        nodeCount = std::max(
+            0.0,
+            std::min(
+                std::min(nodeCount, amountOverLower),
+                nodeCount - amountOverUpper));
```
kagamiori: This logic is a bit hard to follow. So essentially we're trying to get the correct node count at both boundaries, right? Then what about something like this:
```
double sum = 0.0;
double previous;
postOrderTraverse(root_, [...](int32_t node) {
  double nodeCount = counts_[node];
  if (nodeCount == 0) {
    return true;
  }
  previous = sum;
  sum += nodeCount;
  if (sum <= lowerRank) {
    return true;
  } else if (previous < lowerRank) {
    nodeCount = sum - lowerRank;
  } else if (sum > upperRank) {
    nodeCount = upperRank - previous;
  }
  double value = ...;
  result.push_back({value, nodeCount});
  return previous < upperRank;
},
lefts_,
rights_);
```

velox/functions/lib/QuantileDigest.h
```diff
@@ -1371,6 +1403,88 @@ QuantileDigest<T, Allocator>::getDistributionFunction(
...
+
+  double totalValue = 0.0;
+  double totalCount = 0.0;
+
+  for (const auto& pair : valuesAndCounts) {
+    totalValue += pair.value * pair.count;
+    totalCount += pair.count;
+  }
+
+  return totalValue / totalCount;
```
kagamiori: I think this logic is more specific to the UDF. Let's make it a helper method in the UDF class instead. You can also add it to the QuantileDigestTest class as a helper method to allow testing.

natashasehgal: Makes sense, moved this to test

velox/functions/lib/QuantileDigest.h
natashasehgal: This was taken as exact Java logic so there's no parity issue. I did a similar modification now, and the tests pass so it's ok

This exact code didn't work, we don't always push to result in cases of zero, null etc

velox/functions/lib/tests/QuantileDigestTest.cpp
```diff
@@ -952,4 +976,196 @@ TEST_F(QuantileDigestTest, getDistributionFunction) {
   testGetDistributionFunction<float>();
 }
 
+TEST_F(QuantileDigestTest, truncatedMeanBaseCases) {
```
kagamiori, line 1142: Could you also add a unit test that checks the immediate results from getValuesInQuantileRange(), ideally with weighted counts across the boundaries? 

The rest of this PR looks good to me. Thanks!

velox/functions/lib/tests/QuantileDigestTest.cpp
```diff
@@ -952,4 +976,290 @@ TEST_F(QuantileDigestTest, getDistributionFunction) {
...
+      StlAllocator<int64_t>(allocator()), 0.01};
+  singleDigest.add(42, 5.0);
+
+  auto result5 = singleDigest.getValuesInQuantileRange(0.0, 1.0);
+  EXPECT_EQ(result5.size(), 1);
+  EXPECT_EQ(result5[0].value, 42);
+  EXPECT_NEAR(result5[0].count, 5.0, 1e-6);
+}
+
+TEST_F(QuantileDigestTest, getValuesInQuantileRangeWeighted) {
```
kagamiori: nit: Is this test case the same as TEST_F(QuantileDigestTest, getValuesInQuantileRange)? If so, let's remove it.

velox/functions/lib/tests/QuantileDigestTest.cpp
```diff
@@ -952,4 +976,290 @@ TEST_F(QuantileDigestTest, getDistributionFunction) {
...
+  digest.add(15, 4.0);
+
+  auto result1 = digest.getValuesInQuantileRange(0.1, 0.7);
+  EXPECT_GT(result1.size(), 0);
+
+  double totalWeight = 0.0;
+  for (const auto& pair : result1) {
+    totalWeight += pair.count;
+  }
+  EXPECT_NEAR(totalWeight, 6.0, 0.1);
```
kagamiori: Would it make sense to just check the content in result1 directly? Something like
```
std::vector<decltype(digest)::ValueCountPair> expected = {{1, 1.0}, {5, 3.0}, {10, 1.0}, {15, 1.0}};
EXPECT_EQ(result1, expected);
```
Also, I think the algorithm you implemented is unlikely to cause imprecision issue of floating-point. Do we really need EXPECT_NEAR at the checks? If we can use EXPECT_EQ directly, let's not use EXPECT_NEAR.

Also, what if `lowerQuantileBound * weightedCount_` or `upperQuantileBound * weightedCount_` are not integers? Could we have one test case of this?

velox/functions/lib/tests/QuantileDigestTest.cpp
natashasehgal: Ah, missed while refactoring

velox/functions/lib/tests/QuantileDigestTest.cpp
natashasehgal: Good point. I should actually use the value type in the ValueCount structure as well. Created an example with non-int values here

And more test cases for non-integer quantile/result, quantile digest of bigint and real data type

natashasehgal: I had issues doing this exact approach. There were allocator issues. Expected vector uses default allocator, but the vector in getValuesInQuantileRange uses rebindalloc, similar to the rest of the QuantileDigest library. I'm using a simple helper that should achieve the same thing

velox/functions/lib/tests/QuantileDigestTest.cpp
```diff
@@ -952,4 +988,345 @@ TEST_F(QuantileDigestTest, getDistributionFunction) {
...
+}
+
+TEST_F(QuantileDigestTest, truncatedMeanBoundaries) {
+  QuantileDigest<int64_t> digest{StlAllocator<int64_t>(allocator()), 0.01};
+  for (int64_t i = 1; i <= 10; i++) {
+    digest.add(i, 1.0);
+  }
+
+  // Test exact boundaries
+  EXPECT_NEAR(getTruncatedMean(digest, 0.0, 1.0).value(), 5.5, 0.01);
```
kagamiori: nit: Why use 0.01 as the error bound here while using 0.1 at other checks? Could we use the same error bound at all checks (except the ones that use 1e-10 and 1e5 for very large and very small values)?

kagamiori (approved):
LGTM. Left one nit. Thank you for adding this new API and iterating on it!

PR #14306, opened 2025-08-01
Fix microsecond getting truncated when using to_timestamp casting from String to Timestamp in Spark

kapilks:
Spark function `to_timestamp` currently truncates microseconds part.
This gives different result compares to Vanilla spark.

Issue: https://github.com/apache/incubator-gluten/issues/10314

### DIFF
--- a/velox/functions/lib/DateTimeFormatter.cpp
+++ b/velox/functions/lib/DateTimeFormatter.cpp
@@ -819,7 +819,8 @@ int32_t parseFromPattern(
     const char* end,
     Date& date,
     bool specifierNext,
-    DateTimeFormatterType type) {
+    DateTimeFormatterType type,
+    bool truncateFractionOfSecondToMillis) {
   if (curPattern.specifier == DateTimeFormatSpecifier::TIMEZONE_OFFSET_ID) {
     int64_t size;
     if (curPattern.minRepresentDigits < 3) {
@@ -910,7 +911,14 @@ int32_t parseFromPattern(
       // it as the whole number; otherwise, it pads the number with zeros.
       if (type != DateTimeFormatterType::STRICT_SIMPLE &&
           type != DateTimeFormatterType::LENIENT_SIMPLE) {
-        number *= std::pow(10, 3 - count);
+        if (truncateFractionOfSecondToMillis) {
+          number *= std::pow(10, 3 - count);
+          number *= util::kMicrosPerMsec;
+        } else {
+          number *= std::pow(10, 6 - count);
+        }
+      } else {
+        number *= util::kMicrosPerMsec;
       }
     } else if (
         (curPattern.specifier == DateTimeFormatSpecifier::YEAR ||
@@ -1093,7 +1101,7 @@ int32_t parseFromPattern(
         break;
 
       case DateTimeFormatSpecifier::FRACTION_OF_SECOND:
-        date.microsecond = number * util::kMicrosPerMsec;
+        date.microsecond = number;
         break;
 
       case DateTimeFormatSpecifier::WEEK_YEAR:
@@ -1582,12 +1590,26 @@ Expected<DateTimeResult> DateTimeFormatter::parse(
         if (i + 1 < tokens_.size() &&
             tokens_[i + 1].type == DateTimeToken::Type::kPattern) {
           if (parseFromPattern(
-                  tok.pattern, input, cur, end, date, true, type_) == -1) {
+                  tok.pattern,
+                  input,
+                  cur,
+                  end,
+                  date,
+                  true,
+                  type_,
+                  truncateFractionOfSecondToMillis_) == -1) {
             return parseFail(input, cur, end);
           }
         } else {
           if (parseFromPattern(
-                  tok.pattern, input, cur, end, date, false, type_) == -1) {
+                  tok.pattern,
+                  input,
+                  cur,
+                  end,
+                  date,
+                  false,
+                  type_,
+                  truncateFractionOfSecondToMillis_) == -1) {
             return parseFail(input, cur, end);
           }
         }
@@ -1816,7 +1838,8 @@ Expected<std::shared_ptr<DateTimeFormatter>> buildMysqlDateTimeFormatter(
 }
 
 Expected<std::shared_ptr<DateTimeFormatter>> buildJodaDateTimeFormatter(
-    const std::string_view& format) {
+    const std::string_view& format,
+    bool truncateFractionOfSecondToMillis) {
   if (format.empty()) {
     if (threadSkipErrorDetails()) {
       return folly::makeUnexpected(Status::UserError());
@@ -1826,6 +1849,7 @@ Expected<std::shared_ptr<DateTimeFormatter>> buildJodaDateTimeFormatter(
   }
 
   DateTimeFormatterBuilder builder(format.size());
+  builder.truncateFractionOfSecondToMillis(truncateFractionOfSecondToMillis);
   const char* cur = format.data();
   const char* end = cur + format.size();
 
--- a/velox/functions/lib/DateTimeFormatter.h
+++ b/velox/functions/lib/DateTimeFormatter.h
@@ -184,11 +184,13 @@ class DateTimeFormatter {
       std::unique_ptr<char[]>&& literalBuf,
       size_t bufSize,
       std::vector<DateTimeToken>&& tokens,
-      DateTimeFormatterType type)
+      DateTimeFormatterType type,
+      bool truncateFractionOfSecondToMillis)
       : literalBuf_(std::move(literalBuf)),
         bufSize_(bufSize),
         tokens_(std::move(tokens)),
-        type_(type) {}
+        type_(type),
+        truncateFractionOfSecondToMillis_(truncateFractionOfSecondToMillis) {}
 
   const std::unique_ptr<char[]>& literalBuf() const {
     return literalBuf_;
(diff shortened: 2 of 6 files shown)


velox/functions/lib/DateTimeFormatter.h
```diff
@@ -184,11 +184,13 @@ class DateTimeFormatter {
       std::unique_ptr<char[]>&& literalBuf,
       size_t bufSize,
       std::vector<DateTimeToken>&& tokens,
-      DateTimeFormatterType type)
+      DateTimeFormatterType type,
+      bool truncateFractionOfSecondToMillis)
```
rui-mo: Would you please add description for this parameter, including how the fraction part is handled when true and false?

rui-mo: Would it be clearer to provide a parameter for the truncating unit, like second, milli, micro? This looks not like a 0 or 1 option.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -36,7 +36,8 @@ Expected<std::shared_ptr<DateTimeFormatter>> getDateTimeFormatter(
       return buildSimpleDateTimeFormatter(format, /*lenient=*/true);
     default:
       return buildJodaDateTimeFormatter(
-          std::string_view(format.data(), format.size()));
+          std::string_view(format.data(), format.size()),
+          /* truncateFractionOfSecondToMillis */ false);
```
rui-mo: nit: /\*truncateFractionOfSecondToMillis=\*/

rui-mo (commented):
Thanks for the fix!

velox/functions/lib/DateTimeFormatter.h
kapilks: This is good point, will try to update with this change

kapilks: Updated, thanks

kapilks:
@rui-mo Thanks for review. I have addressed your comments. Can you please review again

velox/functions/lib/DateTimeFormatter.h
```diff
@@ -135,6 +135,8 @@ FOLLY_ALWAYS_INLINE bool isTimeUnit(const DateTimeUnit unit) {
   return unit < DateTimeUnit::kDay;
 }
 
+enum class FractionOfSecondPrecision { kSecond, kMillisecond, kMicrosecond };
```
rui-mo: You might could move the `TimestampUnit` to 'velox/type/Timestamp.h' and reuse it.

https://github.com/facebookincubator/velox/blob/5088c18b0fd36bca0174b7a9941f390f703baf49/velox/vector/arrow/Bridge.h#L28-L33

rui-mo (commented):
Thanks.

velox/functions/lib/DateTimeFormatter.h
kapilks: Timestamp.h has already `TimestampPrecision`. That I can use. That doens't have `second` option but it should be fine as it is not used as of now.
https://github.com/facebookincubator/velox/blob/a417b2b45cd18629970330d1a8ec51a55d1e63a7/velox/type/Timestamp.h#L33-L37

velox/functions/lib/DateTimeFormatter.h
```diff
@@ -237,13 +239,17 @@ class DateTimeFormatter {
...
+  TimestampPrecision fractionOfSecondPrecision_;
 };
 
 Expected<std::shared_ptr<DateTimeFormatter>> buildMysqlDateTimeFormatter(
     const std::string_view& format);
 
 Expected<std::shared_ptr<DateTimeFormatter>> buildJodaDateTimeFormatter(
-    const std::string_view& format);
+    const std::string_view& format,
+    // Default kMilliseconds for Presto, in Spark it is set to kMicroseconds
```
rui-mo: nit: move the comment above the function name. Some example:
 
/// @param fractionOfSecondPrecision The precision of the second fraction part. By
/// default, it is kMilliseconds to be compatible with Presto.

velox/functions/lib/DateTimeFormatterBuilder.cpp
```diff
@@ -173,6 +173,16 @@ DateTimeFormatterBuilder& DateTimeFormatterBuilder::appendFractionOfSecond(
   return *this;
 }
 
+DateTimeFormatterBuilder&
+DateTimeFormatterBuilder::setFractionOfSecondPrecision(
+    TimestampPrecision fractionOfSecondPrecision) {
+  VELOX_CHECK(
+      fractionOfSecondPrecision != TimestampPrecision::kNanoseconds,
```
rui-mo: It would be preferable to check against the actual supported types, in case in the future `second` might be added in the `TimestampPrecision`.

velox/functions/lib/DateTimeFormatterBuilder.h
```diff
@@ -248,6 +248,16 @@ class DateTimeFormatterBuilder {
   /// be 9870, with digit being 6 the formatted result will be 987000
   DateTimeFormatterBuilder& appendFractionOfSecond(size_t digits);
 
+  /// Sets the precision for the fraction of second part of timestamp,
+  ///
+  /// \param fractionOfSecondPrecision this can be set to kSecond, kMilliSecond,
```
rui-mo: nit: use @param

rui-mo: TimestampPrecision does not include kSecond, is the description correct? Please also use the correct enum names.

velox/functions/lib/DateTimeFormatterBuilder.h
```diff
@@ -248,6 +248,16 @@ class DateTimeFormatterBuilder {
...
+  /// kMicrosecond. The additional digits present in the fraction of second will
+  /// be truncated. e.g. 1999-01-01 01:59:00.123456, with precision set to
+  /// kMillisecond the result will be 123 or with kMicrosecond result will be
```
rui-mo: the result -> the second fraction part

velox/functions/lib/DateTimeFormatterBuilder.h
```diff
@@ -248,6 +248,16 @@ class DateTimeFormatterBuilder {
...
+  /// 123456
```
rui-mo: nit: add `.` at the end for comments.

rui-mo (commented):
Thanks. Would you please run the fuzzer for the get_timestamp function for 2 minutes to ensure it works?

kapilks:
> Thanks. Would you please run the fuzzer test for the get_timestamp function for 2 minutes to ensure it works?

Thanks for this recommendation. 

I have run the fuzzer for `get_timestamp`. The spark `to_timestamp` maps to this in velox.

image

kapilks:
@rui-mo Thanks for review. I have addressed all the comments.
Please take a look.

kapilks:
The failure in CI doesn't seem related

image

velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -910,7 +911,15 @@ int32_t parseFromPattern(
...
-        number *= std::pow(10, 3 - count);
+        if (fractionOfSecondPrecision == TimestampPrecision::kMilliseconds) {
+          number *= std::pow(10, 3 - count);
+          number *= util::kMicrosPerMsec;
+        } else if (
+            fractionOfSecondPrecision == TimestampPrecision::kMicroseconds) {
+          number *= std::pow(10, 6 - count);
+        }
+      } else {
+        number *= util::kMicrosPerMsec;
```
rui-mo, line 922: @NEUpanning Would you like to review the above change? Thanks.

velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -910,7 +911,15 @@ int32_t parseFromPattern(
       // it as the whole number; otherwise, it pads the number with zeros.
       if (type != DateTimeFormatterType::STRICT_SIMPLE &&
           type != DateTimeFormatterType::LENIENT_SIMPLE) {
-        number *= std::pow(10, 3 - count);
+        if (fractionOfSecondPrecision == TimestampPrecision::kMilliseconds) {
+          number *= std::pow(10, 3 - count);
+          number *= util::kMicrosPerMsec;
```
rui-mo, line 916: Any possibility that count is larger than 3 for milli precision? Do we need a check for it?

velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -910,7 +911,15 @@ int32_t parseFromPattern(
...
+        } else if (
+            fractionOfSecondPrecision == TimestampPrecision::kMicroseconds) {
+          number *= std::pow(10, 6 - count);
```
rui-mo, line 919: ditto, can count be larger than 6?

velox/functions/lib/DateTimeFormatter.cpp
```diff
@@ -910,7 +911,15 @@ int32_t parseFromPattern(
       // it as the whole number; otherwise, it pads the number with zeros.
```
rui-mo, line 911: Would you please update the comments accordingly?

rui-mo (commented):
Thanks!

velox/functions/lib/DateTimeFormatter.cpp
NEUpanning: @rui-mo It looks good to me. I see it has just moved from [1096L](https://github.com/facebookincubator/velox/pull/14306/files#diff-4ec398f7f5c4ab9996a57283368b3f08d1c76b39d031f7a552276ca812aba44aL1096).

rui-mo: Thanks for checking :)

PR #14314, opened 2025-08-01
feat: Add applyingRemainingFilter logic inside TableEvolutionFuzzer

juwentus1234:
Summary:
Use expression fuzzer to generate random filter expression and apply it into tableEvolutionFuzzer with schema evolution.
This is first draft to gurantee we are in the right direction.

Differential Revision: D79475106

### DIFF
--- a/velox/exec/tests/TableEvolutionFuzzer.cpp
+++ b/velox/exec/tests/TableEvolutionFuzzer.cpp
@@ -21,9 +21,13 @@
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/exec/tests/utils/QueryAssertions.h"
 #include "velox/exec/tests/utils/TempDirectoryPath.h"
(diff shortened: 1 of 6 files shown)


velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -716,4 +838,62 @@ void TableEvolutionFuzzer::createWriteTasks(
...
+    return;
+  }
+
+  std::vector<std::string> filterStrings;
+  for (const auto& expr : generatedRemainingFilters.expressions) {
+    auto filterString = expr->toString();
+
+    // Update column names in the filter string using columnNameMapping
+    for (const auto& [originalName, currentName] : columnNameMapping) {
+      // Simple string replacement - this is a basic approach
```
Yuhta, line 893: Add a `TODO` tag here

Yuhta (approved):
Have you checked usually how much data is filtered out by remaining filter?

PR #14317, opened 2025-08-02
feat: Add barrier support for Limit operator

xiaoxmeng:
Summary: Add task barrier support for Limit operator

This changes make limit to support barrier. Since limit doesn't buffer any data, it just needs to tell the driver framework (1) it supports barrier; (2) there is no need for draining processing.

Differential Revision: D79499011

### DIFF
--- a/velox/exec/tests/LimitTest.cpp
+++ b/velox/exec/tests/LimitTest.cpp
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 #include "velox/exec/OutputBufferManager.h"
+#include "velox/exec/PlanNodeStats.h"
+#include "velox/exec/tests/utils/AssertQueryBuilder.h"
 #include "velox/exec/tests/utils/HiveConnectorTestBase.h"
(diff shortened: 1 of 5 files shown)


velox/exec/tests/LimitTest.cpp
```diff
@@ -142,3 +145,183 @@ TEST_F(LimitTest, partialLimitEagerFlush) {
...
+          barrierExecution,
+          offset,
+          limit,
+          numExpectedBarriers,
+          numExpectedOutputRows,
+          numExpectedFinishedSplits,
+          numExpectedOutputBatches);
+    }
+  } testSettings[] = {
+      {true,
```
mbasmanova: I find it hard to understand these tests cases. Would it be possible to add a comment for each of these to explain what is being tested? Consider introducing a Builder class to create these structs to allow the reader to understand these better.

xiaoxmeng: Commented. Thanks!

juwentus1234 (approved):

Issue #14318, opened 2025-08-02
GCC compile error template-id  in constructor and std::in_place_t in initializer list


wecharyu:
### Problem description

There are 3 compile errors when using GCC 14:
```bash
 /root/workspace/velox/./velox/expression/VectorReaders.h:118:36: error: template-id not allowed for constructor in C++20 [-Werror=template-id-cdtor]
  118 |   explicit ConstantVectorReader<T>(ConstantVector<exec_in_t>& vector) {
      |                                    ^~~~~~~~~~~~~~~~~~~~~~~~~
/root/workspace/velox/./velox/expression/VectorReaders.h:118:36: note: remove the '< >'
/root/workspace/velox/./velox/expression/VectorReaders.h:163:32: error: template-id not allowed for constructor in C++20 [-Werror=template-id-cdtor]
  163 |   explicit FlatVectorReader<T>(FlatVector<exec_in_t>& baseVector)
      |                                ^~~~~~~~~~~~~~~~~~~~~
/root/workspace/velox/./velox/expression/VectorReaders.h:163:32: note: remove the '< >'
```
```bash
/root/workspace/velox/velox/vector/tests/FlatMapVectorTest.cpp:424:63: error: converting to 'std::in_place_t' from initializer list would use explicit constructor 'constexpr std::in_place_t::in_place_t()'
  424 |   auto vector = maker_.flatMapVectorNullable<int64_t, int64_t>({
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
  425 |       {{}},
      |       ~~~~~                                                    
  426 |       {std::nullopt},
      |       ~~~~~~~~~~~~~~~                                          
  427 |       {{{1, 0}}},
      |       ~~~~~~~~~~~                                              
  428 |       {{{1, 1}, {2, std::nullopt}}},
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                           
  429 |       {{{0, 0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}}},
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~      
  430 |   });
      |   ~~                                                           
In file included from /usr/include/c++/14/bits/stl_pair.h:62,
                 from /usr/include/c++/14/bits/stl_algobase.h:64,
[...]

PR #4358, opened 2023-03-20
Change StringView constructor to take signed length.

kgpai:
Currently, It is possible to pass a negative size to StringView which due to implict casting creates bad values. We change signature to ensure we dont get passed bad values. 
Fixes #4331. 

Note, that we do not add this check for the user defined literal _sv since it's guaranteed to be non negative (https://en.cppreference.com/w/cpp/language/user_literal) .

PR #14319, opened 2025-08-02
fix: GCC compile error template-id in constructor and std::in_place_t in initializer list

wecharyu:
Fix https://github.com/facebookincubator/velox/issues/14318.

### DIFF
--- a/velox/vector/tests/FlatMapVectorTest.cpp
+++ b/velox/vector/tests/FlatMapVectorTest.cpp
@@ -422,11 +422,17 @@ TEST_F(FlatMapVectorTest, sortedKeyIndices) {
 
 TEST_F(FlatMapVectorTest, toString) {
   auto vector = maker_.flatMapVectorNullable<int64_t, int64_t>({
-      {{}},
-      {std::nullopt},
-      {{{1, 0}}},
-      {{{1, 1}, {2, std::nullopt}}},
-      {{{0, 0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}}},
+      std::make_optional<
+          std::vector<std::pair<int64_t, std::optional<int64_t>>>>({}),
+      std::nullopt,
+      std::make_optional<
+          std::vector<std::pair<int64_t, std::optional<int64_t>>>>({{1, 0}}),
+      std::make_optional<
+          std::vector<std::pair<int64_t, std::optional<int64_t>>>>(
+          {{1, 1}, {2, std::nullopt}}),
+      std::make_optional<
+          std::vector<std::pair<int64_t, std::optional<int64_t>>>>(
+          {{0, 0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}}),
   });
 
   EXPECT_EQ(
--- a/velox/expression/VectorReaders.h
+++ b/velox/expression/VectorReaders.h
@@ -115,7 +115,7 @@ struct ConstantVectorReader {
 
   std::optional<exec_in_t> value;
 
-  explicit ConstantVectorReader<T>(ConstantVector<exec_in_t>& vector) {
+  explicit ConstantVectorReader(ConstantVector<exec_in_t>& vector) {
     if (!vector.isNullAt(0)) {
       value = *vector.rawValues();
     }
@@ -160,7 +160,7 @@ struct FlatVectorReader {
   const exec_in_t* values;
   FlatVector<exec_in_t>* vector;
 
-  explicit FlatVectorReader<T>(FlatVector<exec_in_t>& baseVector)
+  explicit FlatVectorReader(FlatVector<exec_in_t>& baseVector)
       : values(baseVector.rawValues()), vector(&baseVector) {}
 
   exec_in_t operator[](vector_size_t offset) const {


velox/vector/tests/FlatMapVectorTest.cpp
```diff
@@ -422,11 +422,17 @@ TEST_F(FlatMapVectorTest, sortedKeyIndices) {
 
 TEST_F(FlatMapVectorTest, toString) {
   auto vector = maker_.flatMapVectorNullable<int64_t, int64_t>({
-      {{}},
-      {std::nullopt},
-      {{{1, 0}}},
-      {{{1, 1}, {2, std::nullopt}}},
-      {{{0, 0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 5}}},
+      std::make_optional<
```
Yuhta: Use `optionalEmpty` for better readability: https://github.com/facebookincubator/velox/blob/75f9e498a62363a048a7854030feb816d0a65ee7/velox/common/testutil/OptionalEmpty.h#L31

wecharyu: Addressed.

velox/type/Type.h
```diff
@@ -838,6 +838,9 @@ class ShortDecimalType : public DecimalType<TypeKind::BIGINT> {
 
 class LongDecimalType : public DecimalType<TypeKind::HUGEINT> {
  public:
+  // Avoid hiding overloaded virtual function.
+  using DecimalType<TypeKind::HUGEINT>::toString;
```
czentgr, line 890: Can you please rebase? This has been addressed already https://github.com/facebookincubator/velox/blob/2191ecb312d0a1f9bf21972f200286e7e9894602/velox/type/Type.h#L906 by PR https://github.com/facebookincubator/velox/pull/14781

wecharyu: Done.

velox/dwio/common/SelectiveColumnReader.h
```diff
@@ -733,7 +733,7 @@ class SelectiveColumnReader {
 
 template <>
 inline void SelectiveColumnReader::addValue(const std::string_view value) {
-  const uint64_t size = value.size();
+  const uint32_t size = value.size();
```
wecharyu: It fix the [compile error](https://github.com/facebookincubator/velox/actions/runs/18688811569/job/53288733775) after change value type from `folly::StringPiece` to `std::string_view`. cc @pedroerp

czentgr: I do not think this is the correct fix.
The return type of value.size() is `size_t` (aka uint64_t currently).

Instead, the issue is on line 745 calling the constructor of `StringView`. The length argument there is
```
StringView(const char* data, int32_t len)
```
is an int32_t. I think that might be the issue given that the type change fixes it.
So I think we should fix the StringView constructor instead (adding new constructor?).
Can we check this?

wecharyu: Revert it, change the memcpy() calls in `StringView` constructor to fix it.

velox/type/StringView.h
```diff
@@ -69,7 +69,13 @@ struct StringView {
       // small string: inlined. Zero the last 8 bytes first to allow for whole
       // word comparison.
       value_.data = nullptr;
-      memcpy(prefix_, data, size_);
+      // copy first 4 bytes into prefix_.
+      size_t prefixSize = std::min<size_t>(size_, kPrefixSize);
```
czentgr: So interesting how this doesn't come up before. Fundamentally, len can be greater than sizeof(prefix_) + sizeof(value_). 

Now we have two memcpy. That's not ideal.
I'm wondering why the API in the constructor uses int32_t and not uint32_t in the first place. The member is a uint32_t. So negative length will just be huge if it overflows due to assignment from say size_t in the caller.

I'm trying to understand the initial error, where changing the input from size_t to uint32_t helped.

wecharyu: The int32_t is introduced in https://github.com/facebookincubator/velox/pull/4358 to avoid negative size, can we use `simd::memcpy` here to avoid this compiler warnning?

czentgr:
@wecharyu https://github.com/facebookincubator/velox/issues/15245 This is a known issue and also being investigated. We might be looking at a GCC bug.

czentgr:
@wecharyu can we remove the change in StringView for now and we can get the other changes in? It turns out there are some more issues with GCC14 (14.2.1) on CentOS9. Interestingly, the same C++ compiler on Fedora is fine with one of the MemoryPool tests. https://github.com/facebookincubator/velox/actions/runs/19142988275/job/54712980027?pr=15427 vs https://github.com/facebookincubator/velox/actions/runs/19141366948/job/54707064356 if you are interested.

wecharyu:
@czentgr Thanks for the inforamtion, I have reverted the change in StringView.

czentgr:
@wecharyu Thanks, can you please rebase and check on the failing test? It is probably unrelated but lets see if it is a known issue or not.

assignUser (commented):
LGTM

czentgr:
@Yuhta Can you please re-review when you get a chance?

czentgr:
@jinchengchenghh @rui-mo Can you please review when you get a chance?

jinchengchenghh:
If we have a CI to cover gcc14 build?

czentgr:
@jinchengchenghh Yes, we have GCC14 coverage through the fedora CI and now also linux-adapters CI.

Yuhta (approved):

PR #14322, opened 2025-08-04
fix: Fix install stemmer in macOS

zml1206:
Fix error `sed: 1: "Makefile": invalid command code M`.
On macOS, the sed command has some differences from the GNU sed found on Linux systems. macOS does not require the PIC flag.

### DIFF
--- a/scripts/setup-common.sh
+++ b/scripts/setup-common.sh
@@ -232,9 +232,7 @@ function install_stemmer {
   wget_and_untar https://snowballstem.org/dist/libstemmer_c-"${STEMMER_VERSION}".tar.gz stemmer
   (
     cd "${DEPENDENCY_DIR}"/stemmer || exit
-    if [[ "$(uname)" != "Darwin" ]]; then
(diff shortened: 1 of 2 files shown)


zml1206:
cc @assignUser Thanks.

scripts/setup-common.sh
```diff
@@ -232,7 +232,9 @@ function install_stemmer {
   wget_and_untar https://snowballstem.org/dist/libstemmer_c-"${STEMMER_VERSION}".tar.gz stemmer
   (
     cd "${DEPENDENCY_DIR}"/stemmer || exit
-    sed -i '/CPPFLAGS=-Iinclude/ s/$/ -fPIC/' Makefile
+    if [[ "$(uname)" != "Darwin" ]]; then
```
majetideepak: We don't want any customization in the `setup-common.sh` file. Please make a copy of install_stemmer inside setup-macos.sh. There should be other such examples.

zml1206: Updated.

zml1206:
Ping again gently @majetideepak @assignUser Thanks.

zml1206:
resolved by #14484

PR #14339, opened 2025-08-06
feat(geo): Implement NLJ-version of SpatialJoinBuild/Probe Operators

jagill:
Summary:
Implement a simple version of SpatialJoin that is really just a NLJ in
disguise.  This fills out the operator skeleton in exec and allows us to test
correctness.  We could in theory run spatial queries with these, although they
would be very inefficient.

Later commits will replace the SpatialJoinOperators with more efficient
versions with SpatialIndexes, as well as implement Left Joins.

Differential Revision: D79613518

### DIFF
--- a/velox/core/PlanNode.cpp
+++ b/velox/core/PlanNode.cpp
@@ -3064,9 +3064,6 @@ PlanNodePtr PartitionedOutputNode::create(
       deserializeSingleSource(obj, context));
 }
 
-// static
-const JoinType SpatialJoinNode::kDefaultJoinType = JoinType::kInner;
-
 SpatialJoinNode::SpatialJoinNode(
     const PlanNodeId& id,
     JoinType joinType,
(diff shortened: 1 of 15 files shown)


velox/core/PlanNode.cpp
```diff
@@ -1825,6 +1825,102 @@ PlanNodePtr NestedLoopJoinNode::create(
       outputType);
 }
 
+// static
+const JoinType SpatialJoinNode::kDefaultJoinType = JoinType::kInner;
+// static
+const TypedExprPtr SpatialJoinNode::kDefaultJoinCondition = nullptr;
```
MBkkt: `inline static` can be used since C++17 

Example: 
https://cppreference.com/w/cpp/language/static.html
`inline static const std::string class_name{"X"}; // Likewise`


But these can be even better with `static constexpr` and avoid runtime constants at all

mbasmanova (approved):

PR #14346, opened 2025-08-06
fix: Make Spark abs function ansi-compliant

### LINKED ISSUE #14345: Add support for ansi supported abs function for Spark
### Description

Add support for ansi supported abs function for Spark

nimesh1601:
This PR supports ansi-compliant behavior for the Spark abs function.
Fixes #14345.

### DIFF
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -2,12 +2,17 @@
 Mathematical Functions
 ======================
 
-.. spark:function:: abs(x) -> [same as x]
+.. spark:function:: abs(x) -> [same as x] (ANSI compliant)
 
-    *ANSI compliant* - Returns the absolute value of ``x``. When ``x`` is negative minimum
+    Returns the absolute value of ``x``. When ``x`` is negative minimum
     value of integral type returns the same value as ``x`` following
     the behavior when Spark ANSI mode is disabled and throws exception
-    when Spark ANSI mode is enabled.
+    when Spark ANSI mode is enabled. ::
+
+        SELECT abs(-42); -- 42
+        SELECT abs(3.14); -- 3.14
+        SELECT abs(-128); -- 128 (with ANSI mode disabled)
+        -- abs(-128) throws overflow exception with ANSI mode enabled for TINYINT
 
 .. spark:function:: acos(x) -> double
 
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -7,7 +7,7 @@ Mathematical Functions
     Returns the absolute value of ``x``. When ``x`` is negative minimum
     value of integral type returns the same value as ``x`` following
     the behavior when Spark ANSI mode is disabled and throws exception
-    when spark ansi mode is enabled.
+    when Spark ANSI mode is enabled.
 
 .. spark:function:: acos(x) -> double
 
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -12,7 +12,7 @@ Mathematical Functions
         SELECT abs(-42); -- 42
         SELECT abs(3.14); -- 3.14
         SELECT abs(-128); -- 128 (with ANSI mode disabled)
-        -- abs(-128) throws overflow exception with ANSI mode enabled for TINYINT
+        SELECT abs(-128); throws overflow exception with ANSI mode enabled for TINYINT
 
 .. spark:function:: acos(x) -> double
 
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -4,7 +4,7 @@ Mathematical Functions
 
 .. spark:function:: abs(x) -> [same as x]
 
-    Returns the absolute value of ``x``. When ``x`` is negative minimum
+    *ANSI compliant* - Returns the absolute value of ``x``. When ``x`` is negative minimum
     value of integral type returns the same value as ``x`` following
     the behavior when Spark ANSI mode is disabled and throws exception
     when Spark ANSI mode is enabled.
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -12,8 +12,7 @@ Mathematical Functions
         SELECT abs(-42); -- 42
         SELECT abs(3.14); -- 3.14
         SELECT abs(-128); -- 128 (with ANSI mode disabled)
-        SELECT abs(-128); throws overflow exception with ANSI mode enabled for TINYINT
-
+        SELECT abs(-128); -- Overflow exception (with ANSI mode enabled for TINYINT)
 .. spark:function:: acos(x) -> double
 
     Returns the inverse cosine (a.k.a. arc cosine) of ``x``.
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -5,8 +5,9 @@ Mathematical Functions
 .. spark:function:: abs(x) -> [same as x]
 
     Returns the absolute value of ``x``. When ``x`` is negative minimum
-    value of integral type, returns the same value as ``x`` following
-    the behavior when Spark ANSI mode is disabled.
+    value of integral type returns the same value as ``x`` following
+    the behavior when Spark ANSI mode is disabled and throws exception
+    when spark ansi mode is enabled.
 
 .. spark:function:: acos(x) -> double
 
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -33,31 +33,35 @@ template <typename TExec>
 struct AbsFunction {
   VELOX_DEFINE_FUNCTION_TYPES(TExec);
 
-  bool ansiEnabled_ = false;
-
   template <typename TInput>
   FOLLY_ALWAYS_INLINE void initialize(
       const std::vector<TypePtr>& /*inputTypes*/,
       const core::QueryConfig& config,
       const TInput* /*a*/) {
-    ansiEnabled_ = config.sparkAnsiEnabled();
+    ansiEnabled = config.sparkAnsiEnabled();
   }
 
   template <typename T>
-  FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
+  FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        if (ansiEnabled_) {
-          // In ANSI mode, throws an overflow error.
-          VELOX_USER_FAIL("Arithmetic overflow: abs({})", a);
+        if (ansiEnabled) {
+          // In ANSI mode, returns an overflow error.
+          if (threadSkipErrorDetails()) {
+            return Status::UserError();
+          }
+          return Status::UserError("Arithmetic overflow: abs({})", a);
         }
         // In ANSI off mode, returns the same negative minimum value.
         result = a;
-        return;
+        return Status::OK();
       }
     }
     result = std::abs(a);
+    return Status::OK();
   }
+
+  bool ansiEnabled = false;
 };
 
 template <typename T>
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -48,10 +48,10 @@ struct AbsFunction {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
         if (ansiEnabled_) {
-          // In ANSI mode, throws an overflow error
+          // In ANSI mode, throws an overflow error.
           VELOX_USER_FAIL("Arithmetic overflow: abs({})", a);
         }
-        // In ANSI off mode, returns the same negative minimum value
+        // In ANSI off mode, returns the same negative minimum value.
         result = a;
         return;
       }
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -32,9 +32,9 @@ namespace facebook::velox::functions::sparksql {
 template <typename TExec>
 struct AbsFunction {
   VELOX_DEFINE_FUNCTION_TYPES(TExec);
-  
+
   bool ansiEnabled_ = false;
-  
+
   template <typename TInput>
   FOLLY_ALWAYS_INLINE void initialize(
       const std::vector<TypePtr>& /*inputTypes*/,
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -31,13 +31,11 @@ namespace facebook::velox::functions::sparksql {
 // The abs implementation is used for primitive types except for decimal type.
 template <typename TExec>
 struct AbsFunction {
-  VELOX_DEFINE_FUNCTION_TYPES(TExec);
-
-  template <typename TInput>
+  template <typename T>
   FOLLY_ALWAYS_INLINE void initialize(
       const std::vector<TypePtr>& /*inputTypes*/,
       const core::QueryConfig& config,
-      const TInput* /*a*/) {
+      const T* /*a*/) {
     ansiEnabled_ = config.sparkAnsiEnabled();
   }
 
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -38,14 +38,14 @@ struct AbsFunction {
       const std::vector<TypePtr>& /*inputTypes*/,
       const core::QueryConfig& config,
       const TInput* /*a*/) {
-    ansiEnabled = config.sparkAnsiEnabled();
+    ansiEnabled_ = config.sparkAnsiEnabled();
   }
 
   template <typename T>
   FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        if (ansiEnabled) {
+        if (ansiEnabled_) {
           // In ANSI mode, returns an overflow error.
           if (threadSkipErrorDetails()) {
             return Status::UserError();
@@ -61,7 +61,8 @@ struct AbsFunction {
     return Status::OK();
   }
 
-  bool ansiEnabled = false;
+ private:
+  bool ansiEnabled_ = false;
 };
 
 template <typename T>
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -31,15 +31,32 @@ namespace facebook::velox::functions::sparksql {
 // The abs implementation is used for primitive types except for decimal type.
 template <typename TExec>
 struct AbsFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+  
+  bool ansiEnabled_ = false;
+  
+  template <typename TInput>
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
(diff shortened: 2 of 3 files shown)


nimesh1601:
@rui-mo Can you review this ?

rui-mo:
cc: @zhli1142015

nimesh1601:
@rui-mo Yes, currently it falls back to Vanilla Spark. However, I'm working on introducing support for ANSI mode through a separate flag. This will allow us to develop and test features specifically for ANSI compatibility. The PR for the flag and initial support for arithmetic functions is available here [Gluten PR](https://github.com/apache/incubator-gluten/pull/10357)

zhli1142015:
Thank you for your change. I understand that there are many places that need to be modified to support ANSI mode. Do you have an overall design specification, and is it necessary to introduce new functions everywhere instead of modifying the existing abs implementation to support ANSI mode?

By the way, in Gluten there is a FallbackOnANSIMode rule that triggers a fallback when ANSI mode is enabled. I noticed that your Gluten PR did not modify this rule—could you explain how you tested your changes?

nimesh1601:
> Thank you for your change. I understand that there are many places that need to be modified to support ANSI mode. Do you have an overall design specification, and is it necessary to introduce new functions everywhere instead of modifying the existing abs implementation to support ANSI mode?
> 
> By the way, in Gluten there is a FallbackOnANSIMode rule that triggers a fallback when ANSI mode is enabled. I noticed that your Gluten PR did not modify this rule—could you explain how you tested your changes?

@zhli1142015 The Fallback Rule is implemented in this [PR](https://github.com/apache/incubator-gluten/pull/10357/files#diff-cdb3a65a8971463cccf175143de5eb1e909f05ff6451895defc46acfd708b821), once this is landed, I can write and add tests for the same in the Gluten abs function [PR](https://github.com/apache/incubator-gluten/pull/10372). 
This function was tested locally by adding relevant changes shared in the above PR's. 

I don't have an overall design specification at the moment, but I can create one if there is none already. Can discuss with @PHILO-HE once. If there are any format/guidelines that we follow, can you share one? I can create a doc accordingly.

As per my understanding of the code, we can't access whether ansi is enabled or not in the function definition itself, that's why I created a separate function. Let me know if my understanding is incorrect

rui-mo:
> As per my understanding of the code, we can't access whether ansi is enabled or not in the function definition itself, that's why I created a separate function. Let me know if my understanding is incorrect

Hi @nimesh1601, we can access the query config in the function `initialize` method. For example, https://github.com/facebookincubator/velox/blob/a417b2b45cd18629970330d1a8ec51a55d1e63a7/velox/functions/sparksql/DateTimeFunctions.h#L152-L154.

In terms of ANSI support, I would be in favor of making the behavior configurable such that the majority of the code can be reused. @PHILO-HE @zhli1142015 How do you think?

philo-he:
@rui-mo, @nimesh1601, @zhli1142015, I previously reviewed Spark's code and identified two types of cases:
1) For most expressions, their ANSI behavior is governed by Spark's ANSI configuration.
2) In a few instances, the ANSI flag for an expression is explicitly set during its construction in Spark.

To align Velox with Spark's ANSI configuration, we can introduce a new configuration in Velox. This will be used for the first case to distinguish between ANSI and non-ANSI behavior in function implementation, as Rui mentioned. For the second case, we can implement the Velox function with the ANSI flag as an argument, passing the flag from Gluten to Velox.

I recommend merging the ANSI and non-ANSI implementations into a single function to minimize duplication. There may be only a few exceptions that necessitate separate implementations.

rui-mo:
@PHILO-HE Looks nice. Would you like to create a Velox issue to document our plan for ANSI support?

philo-he:
> @PHILO-HE Looks nice. Would you like to create a Velox issue to document our plan for ANSI support?

@rui-mo, we have created a GitHub issue in Gluten, where required tasks will be listed.

nimesh1601:
Thanks @PHILO-HE @rui-mo @zhli1142015. I will close this PR and will raise a new one as per our discussion above

rui-mo:
Hi @nimesh1601, I assume you could update this PR to respect the ANSI config after https://github.com/facebookincubator/velox/pull/14375. Thanks.

nimesh1601:
> Hi @nimesh1601, I assume you could update this PR to respect the ANSI config after #14375. Thanks.

sure

nimesh1601:
@rui-mo Can you review this ?

velox/docs/functions/spark/math.rst
```diff
@@ -5,8 +5,9 @@ Mathematical Functions
 .. spark:function:: abs(x) -> [same as x]
 
     Returns the absolute value of ``x``. When ``x`` is negative minimum
-    value of integral type, returns the same value as ``x`` following
-    the behavior when Spark ANSI mode is disabled.
+    value of integral type returns the same value as ``x`` following
+    the behavior when Spark ANSI mode is disabled and throws exception
+    when spark ansi mode is enabled.
```
philo-he: For consistency, please change "ansi" to "ANSI" and "spark" to "Spark".

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,15 +31,32 @@ namespace facebook::velox::functions::sparksql {
...
   FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        // To be compatible with Spark's ANSI off mode, when the input is
-        // negative minimum value, returns the same value as input instead of
-        // throwing an error.
-        result = a;
-        return;
+        if (ansiEnabled_) {
+          // In ANSI mode, throw an overflow error
```
philo-he: throw -> throws
Add a period at the end.

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,15 +31,32 @@ namespace facebook::velox::functions::sparksql {
...
+          VELOX_USER_FAIL("Arithmetic overflow: abs({})", a);
+        } else {
```
philo-he: omit "else".

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,15 +31,32 @@ namespace facebook::velox::functions::sparksql {
...
+          // To be compatible with Spark's ANSI off mode, when the input is
+          // negative minimum value, returns the same value as input instead of
+          // throwing an error.
```
philo-he: Please make the comment concise, like: In ANSI off mode, returns the same negative minimum value.

velox/functions/sparksql/tests/ArithmeticTest.cpp
```diff
@@ -761,6 +761,37 @@ TEST_F(ArithmeticTest, abs) {
...
+  // Floating point should still work normally (no overflow for floats)
+  EXPECT_EQ(abs<float>(-99999.9999f), 99999.9999f);
+  EXPECT_EQ(
+      abs<float>(std::numeric_limits<float>::lowest()),
+      std::numeric_limits<float>::max());
+  EXPECT_EQ(abs<double>(-99999.9999), 99999.9999);
+  EXPECT_EQ(
+      abs<double>(std::numeric_limits<double>::lowest()),
+      std::numeric_limits<double>::max());
+}
```
philo-he, line 789: Do we still need to test normal cases? It seems repetitive. Perhaps we could focus on testing only the divergent cases.

nimesh1601: As these tests are in ANSI mode, I think we still need basic tests, but have removed redundant tests

philo-he: I reviewed Spark's test code in the link below. For common cases, they test both ANSI on and off within a for loop, expecting identical behavior. For divergent cases, they use different expectations. This approach avoids duplicate test code. Could we adopt this methodology in Velox to test all ANSI-related functions?

https://github.com/apache/spark/blob/5a30d68a652075497d72985fffb31ee281b18c79/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ArithmeticExpressionSuite.scala#L512-L556

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,13 +31,27 @@ namespace facebook::velox::functions::sparksql {
...
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        // To be compatible with Spark's ANSI off mode, when the input is
-        // negative minimum value, returns the same value as input instead of
-        // throwing an error.
+        if (ansiEnabled_) {
+          // In ANSI mode, throws an overflow error
+          VELOX_USER_FAIL("Arithmetic overflow: abs({})", a);
+        }
+        // In ANSI off mode, returns the same negative minimum value
```
philo-he: Please add a period at the end of each complete sentence in comments. Ditto for other comments.

velox/functions/sparksql/tests/ArithmeticTest.cpp
nimesh1601: yeah, that make sense

velox/docs/functions/spark/math.rst
```diff
@@ -5,8 +5,9 @@ Mathematical Functions
 .. spark:function:: abs(x) -> [same as x]
```
philo-he: I am proposing a consistent way to help user identify ANSI compliant functions.
https://github.com/facebookincubator/velox/pull/14698

Let's hold on this PR until that PR is merged. Thanks.

nimesh1601: Does this work? or should I change something?

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,13 +31,27 @@ namespace facebook::velox::functions::sparksql {
...
   template <typename T>
   FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        // To be compatible with Spark's ANSI off mode, when the input is
-        // negative minimum value, returns the same value as input instead of
-        // throwing an error.
+        if (ansiEnabled_) {
+          // In ANSI mode, throws an overflow error.
+          VELOX_USER_FAIL("Arithmetic overflow: abs({})", a);
```
jinchengchenghh: return Status rather than void

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,13 +31,27 @@ namespace facebook::velox::functions::sparksql {
 // The abs implementation is used for primitive types except for decimal type.
 template <typename TExec>
 struct AbsFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  bool ansiEnabled_ = false;
```
jinchengchenghh: Struct member format should be bool ansiEnabled, and move it after the function

velox/functions/sparksql/tests/ArithmeticTest.cpp
```diff
@@ -735,30 +735,55 @@ TEST_F(ArithmeticTest, checkedDivide) {
...
-  EXPECT_EQ(abs<double>(-99999.9999), 99999.9999);
-  EXPECT_EQ(
-      abs<double>(std::numeric_limits<double>::lowest()),
-      std::numeric_limits<double>::max());
+
+  // Test abs with ANSI on.
+  queryCtx_->testingOverrideConfigUnsafe(
+      {{core::QueryConfig::kSparkAnsiEnabled, "true"}});
+
+  EXPECT_THROW(abs<int8_t>(std::numeric_limits<int8_t>::min()), VeloxUserError);
```
jinchengchenghh: VELOX_ASSERT_THROW

velox/docs/functions/spark/math.rst
philo-he: @nimesh1601, let's follow the proposed convention. Please mark it as ANSI-compliant, formatted like:

.. spark:function:: abs(x) -> [same as x] (ANSI compliant)

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,19 +31,37 @@ namespace facebook::velox::functions::sparksql {
...
         result = a;
-        return;
+        return Status::OK();
       }
     }
     result = std::abs(a);
+    return Status::OK();
   }
+
+  bool ansiEnabled = false;
```
jinchengchenghh: Please mark it as private, and name it ansiEnabled_ as this comment. https://github.com/facebookincubator/velox/pull/14774#discussion_r2336146815

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,19 +31,37 @@ namespace facebook::velox::functions::sparksql {
...
-  FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
+  FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        // To be compatible with Spark's ANSI off mode, when the input is
-        // negative minimum value, returns the same value as input instead of
-        // throwing an error.
+        if (ansiEnabled) {
+          // In ANSI mode, returns an overflow error.
+          if (threadSkipErrorDetails()) {
```
jinchengchenghh, line 48: Usually, we don't check threadSkipErrorDetails

velox/docs/functions/spark/math.rst
```diff
@@ -4,9 +4,10 @@ Mathematical Functions
 
 .. spark:function:: abs(x) -> [same as x]
 
-    Returns the absolute value of ``x``. When ``x`` is negative minimum
-    value of integral type, returns the same value as ``x`` following
-    the behavior when Spark ANSI mode is disabled.
+    *ANSI compliant* - Returns the absolute value of ``x``. When ``x`` is negative minimum
+    value of integral type returns the same value as ``x`` following
+    the behavior when Spark ANSI mode is disabled and throws exception
+    when Spark ANSI mode is enabled.
```
jinchengchenghh: Add an example like https://github.com/facebookincubator/velox/pull/14774/files#diff-becf36f6e870124614abb95932682195b98698d0ef9e2dfcaf4e31dbc643dd21R36

velox/functions/sparksql/Arithmetic.h
nimesh1601: Followed similar approach as here: https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/Arithmetic.h#L551

velox/docs/functions/spark/math.rst
```diff
@@ -2,12 +2,17 @@
...
+    Returns the absolute value of ``x``. When ``x`` is negative minimum
     value of integral type returns the same value as ``x`` following
     the behavior when Spark ANSI mode is disabled and throws exception
-    when Spark ANSI mode is enabled.
+    when Spark ANSI mode is enabled. ::
+
+        SELECT abs(-42); -- 42
+        SELECT abs(3.14); -- 3.14
+        SELECT abs(-128); -- 128 (with ANSI mode disabled)
+        -- abs(-128) throws overflow exception with ANSI mode enabled for TINYINT
```
jinchengchenghh: SELECT abs(-128); throws overflow exception for TINYINT (with ANSI mode enabled)

velox/docs/functions/spark/math.rst
```diff
@@ -2,11 +2,17 @@
...
-    value of integral type, returns the same value as ``x`` following
-    the behavior when Spark ANSI mode is disabled.
+    value of integral type returns the same value as ``x`` following
+    the behavior when Spark ANSI mode is disabled and throws exception
+    when Spark ANSI mode is enabled. ::
+
+        SELECT abs(-42); -- 42
+        SELECT abs(3.14); -- 3.14
+        SELECT abs(-128); -- 128 (with ANSI mode disabled)
+        SELECT abs(-128); throws overflow exception with ANSI mode enabled for TINYINT
```
rui-mo: nit:

> SELECT abs(-128); -- Overflow exception (with ANSI mode enabled for TINYINT)

nimesh1601:
@rui-mo can you take a look?

nimesh1601:
@rui-mo can you take a look ?

velox/functions/sparksql/tests/ArithmeticTest.cpp
```diff
@@ -735,30 +736,56 @@ TEST_F(ArithmeticTest, checkedDivide) {
...
       abs<int64_t>(std::numeric_limits<int64_t>::min()),
       std::numeric_limits<int64_t>::min());
-  EXPECT_EQ(abs<float>(-99999.9999f), 99999.9999f);
-  EXPECT_EQ(
-      abs<float>(std::numeric_limits<float>::lowest()),
-      std::numeric_limits<float>::max());
-  EXPECT_EQ(abs<double>(-99999.9999), 99999.9999);
-  EXPECT_EQ(
-      abs<double>(std::numeric_limits<double>::lowest()),
-      std::numeric_limits<double>::max());
```
rui-mo, line 761: Why we need to remove these tests?

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,19 +31,38 @@ namespace facebook::velox::functions::sparksql {
...
+  }
+
   template <typename T>
-  FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
+  FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
     if constexpr (std::is_integral_v<T>) {
       if (FOLLY_UNLIKELY(a == std::numeric_limits<T>::min())) {
-        // To be compatible with Spark's ANSI off mode, when the input is
-        // negative minimum value, returns the same value as input instead of
-        // throwing an error.
```
rui-mo, line 40: > when the input is negative minimum value, returns the same value as input

I assume above is compatible with Spark. Would you please confirm?

velox/functions/sparksql/tests/ArithmeticTest.cpp
nimesh1601: We didn't remove them, but moved them to common tests for both ansi on and off as suggested by @PHILO-HE .

velox/functions/sparksql/Arithmetic.h
nimesh1601: Yes, it's the same behaviour as spark

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,19 +31,38 @@ namespace facebook::velox::functions::sparksql {
 // The abs implementation is used for primitive types except for decimal type.
 template <typename TExec>
 struct AbsFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
rui-mo: nit: this seems not being used.

velox/functions/sparksql/Arithmetic.h
```diff
@@ -31,19 +31,38 @@ namespace facebook::velox::functions::sparksql {
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TInput* /*a*/) {
+    ansiEnabled_ = config.sparkAnsiEnabled();
+  }
+
   template <typename T>
-  FOLLY_ALWAYS_INLINE void call(T& result, const T& a) {
+  FOLLY_ALWAYS_INLINE Status call(T& result, const T& a) {
```
rui-mo, line 43: nit: use the same name for the template parameter in `initialize` and `call` to avoid confusion.

rui-mo (approved):
Thanks. Just two nits.

velox/functions/sparksql/tests/ArithmeticTest.cpp
rui-mo: I see, thanks!

nimesh1601:
@rui-mo, all comments done.

rui-mo:
@nimesh1601 Thank you.

nimesh1601:
@rui-mo Do I need to do anything to get it merged?

rui-mo:
@nimesh1601 I marked this change as ready-to-merge, and we'll need Meta's help to land it.

Issue #14056, opened 2025-07-09
Align Variant::hash() with BaseVector::hashValueAt()


mbasmanova:
### Description

Variant is used to represent a single value of either primitive or complex type. It is used to specify a constant in ConstantTypedExpr.

Currently, Variant::hash() API returns a different value than BaseVector::hashValueAt(). 

We'd like to align Variant::hash() with BaseVector::hashValueAt() to avoid confusion.

Related issues: #14017 #14018

PR #14348, opened 2025-08-06
Refactor Variant::hash to use BitUtil and BaseVector null hash

hotcodemacha:
fixes https://github.com/facebookincubator/velox/issues/14056

### Summary
This PR refactors the hashing logic in `Variant::hash` to use the `BitUtil`
hashing functions and the `BaseVector::kNullHash` constant, replacing
the previous `folly::Hash`‑based approach.  

### Changes
- Introduce `BaseVector::kNullHash` for variant-null hashing.
- Use `bits::hashMix` to combine child hashes in ARRAY and ROW variants.
- Use `bits::commutativeHashMix` for MAP entries to ensure order‑independence.
- Add required includes: `velox/common/base/BitUtil.h` and `velox/vector/BaseVector.h`.
- Remove `folly::Hash` combiners and related dependencies in `Variant.cpp`.

[Updated] Another commit:
- Introduced a shared bits::kNullHash constant to unify null hash handling across components, enabling BaseVector and Variant to reference the same value without cross-module dependency

- Refactored Variant hashing for arrays, rows, maps, and nulls to use the new bits::kNullHash, eliminating the need to include vector headers within the type module

- Added a comprehensive test validating that vectorToVariant(...).hash() returns the same value as hashValueAt() for flat, array, map, and row vectors


hotcodemacha:
@mbasmanova Please help with review. Thanks.

velox/type/Variant.cpp
```diff
@@ -20,6 +20,8 @@
 #include "velox/common/encode/Base64.h"
 #include "velox/type/DecimalUtil.h"
 #include "velox/type/FloatingPointUtil.h"
+#include "velox/common/base/BitUtil.h"
+#include "velox/vector/BaseVector.h"
```
mbasmanova: Can we avoid introduce dependency from "type" to "vector"?

mbasmanova (commented):
Thanks. Would it be possible to add a test to verify that BaseVector::hashValueAt(index) == BaseVector::variantAt(index).hash()?

velox/type/Variant.cpp
hotcodemacha: ack

hotcodemacha:
> Thanks. Would it be possible to add a test to verify that BaseVector::hashValueAt(index) == BaseVector::variantAt(index).hash()?

ack

mbasmanova (approved):
Thank you.

PR #14369, opened 2025-08-07
fix: Fix approximateTypeEncodingwidth to count Row with one in width

kagamiori:
Summary: The approximateTypeEncodingwidth() method should count the Row-type layer itself as 1 in width. Failing to do so could cause unit test failure of VectorFuzzerTest.randTypeByWidth.

Differential Revision: D79829209

### DIFF
--- a/velox/type/TypeEncodingUtil.cpp
+++ b/velox/type/TypeEncodingUtil.cpp
@@ -34,16 +34,16 @@ size_t approximateTypeEncodingwidth(const TypePtr& type) {
 
   switch (type->kind()) {
     case TypeKind::ARRAY:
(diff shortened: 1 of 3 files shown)


velox/type/TypeEncodingUtil.cpp
```diff
@@ -43,7 +43,7 @@ size_t approximateTypeEncodingwidth(const TypePtr& type) {
       for (const auto& child : type->asRow().children()) {
         fieldWidth += approximateTypeEncodingwidth(child);
       }
-      return fieldWidth;
+      return 1 + fieldWidth;
```
xiaoxmeng:
```
nit: fieldWidth + 1
```

PR #14370, opened 2025-08-07
feat: Add LongEnumParameter to SignatureBinder

HeidiHan0000:
Added LongEnumParameter to SignatureBinder so that BigintEnum(LongEnumParameter) can be used as a function argument (for the enum_key function in a later PR)

Differential Revision: D78731080

### DIFF
--- a/velox/expression/FunctionSignature.h
+++ b/velox/expression/FunctionSignature.h
@@ -34,7 +34,11 @@ std::string sanitizeName(const std::string& name);
 /// Return a list of primitive type names.
 const std::vector<std::string> primitiveTypeNames();
 
-enum class ParameterType : int8_t { kTypeParameter, kIntegerParameter };
+enum class ParameterType : int8_t {
+  kTypeParameter,
+  kIntegerParameter,
+  kEnumParameter,
+};
 
 /// Canonical names for functions that have special treatments in pushdowns.
 enum class FunctionCanonicalName {
@@ -86,6 +90,10 @@ class SignatureVariable {
 
   bool isIntegerParameter() const {
     return type_ == ParameterType::kIntegerParameter;
+  }
+
+  bool isEnumParameter() const {
+    return type_ == ParameterType::kEnumParameter;
   }
 
   bool operator==(const SignatureVariable& rhs) const {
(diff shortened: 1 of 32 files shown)


velox/expression/FunctionSignature.h
```diff
@@ -309,6 +317,15 @@ class FunctionSignatureBuilder {
     return *this;
   }
 
+  FunctionSignatureBuilder& enumVariable(
+      const std::string& name,
+      std::optional<std::string> constraint = std::nullopt) {
```
mbasmanova, line 322: Is there a use case for using the default value? If not, let's remove the default.

velox/expression/tests/SignatureBinderTest.cpp
```diff
@@ -852,6 +855,18 @@ TEST(SignatureBinderTest, customType) {
     testSignatureBinder(signature, {INTEGER()}, TIMESTAMP_WITH_TIME_ZONE());
   }
 
+  {
+    // bigint_enum(enumParameters) -> bigint
+    auto signature = exec::FunctionSignatureBuilder()
+                         .enumVariable("enumParameters")
```
mbasmanova: enumParameters -> E to make this a bit more realistic

mbasmanova (commented):
Let's update PR description to clarify that we allow only one enum parameter.

velox/expression/FunctionSignature.h
HeidiHan0000: For the test in SignatureBinderTest, I use the default constraint.. does that count?

mbasmanova: I remember you mentioned somewhere that you cannot use this without specifying a constrain. If so, it means the default value is not usable. Let's remove it.

HeidiHan0000:
> Let's update PR description to clarify that we allow only one enum parameter.

I'm not sure I understand, do you mean one enum parameter to bigint_enum?

mbasmanova:
> I'm not sure I understand, do you mean one enum parameter to bigint_enum?

We discussed earlier that it is currently not possible to define a function that takes 2 different enums: foo(enum1, enum2).

velox/expression/FunctionSignature.h
HeidiHan0000: For the enum_key function, in using the simple functions api, I wasn't sure how I could use the FunctionSignatureBuilder with the registerFunction api, so I was using the constraints to register the name of that parameter ("E1" or alike). Is that constraint and this constraint the same concept? Would you be able to advise on how I can use this enumVariable to register the parameter instead?

HeidiHan0000:
> > I'm not sure I understand, do you mean one enum parameter to bigint_enum?
> 
> We discussed earlier that it is currently not possible to define a function that takes 2 different enums: foo(enum1, enum2).

Got it. Is it okay if I update that in https://github.com/facebookincubator/velox/pull/14482 where I added the E1 and E2 (or just E)?

mbasmanova (approved):

Issue #3869, opened 2023-01-30
Any plan to enhance ANSI compliance?


lviiii:
### Description

When we enable ANSI mode and run Spark UTs with Velox backend, some special cases got mismatched results with Vanilla Spark. For example, dividing by zero in Vanilla Spark results in an exception, but the result offloaded to Velox depends on the input types, refer to [Divide Expression](https://facebookincubator.github.io/velox/functions/math.html#divide).

Background
Spark build an ANSI compliance for better data quality and easier migration from traditional DBMS. Full details of this dialect are documented in [Spark ANSI Compliance](https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html). When enable ANSI mode in Spark,
1. Spark will throw an exception at runtime instead of returning null results if the inputs to a SQL operator/function are invalid
2. Definite implicit cast syntax rules in ANSI mode, and disallow some invalid castings in ANSI SQL standard.

In the future, we also hope there are some ANSI enhancements in Velox. Is there any plan for this?

mbasmanova:
CC: @rui-mo Rui, is this something you can comment on?

mbasmanova:
@rui-mo @lviiii I assume ANSI compliance question applies to Spark functions only. It should be possible to provide two sets of Spark functions: legacy and ANSI-compliant. If we had to choose between these and provide only one version, then I would think providing ANSI-compliant set of functions is better. What do you think?

mbasmanova:
@rui-mo My impression is that ANSI compliance is limited to the function implementations: scalar, aggregate and window. 

> For example, the behavior when sum overflows in aggregation can be affected. 

Here, the problem is with the SUM aggregate function, not the aggregation operator itself. As you mentioned, we'll need to provide ANSI-compliance packages of Spark functions which include scalar, aggregate and possibly window functions.

FelixYBW:
@mbasmanova does presto has any ANSI compliance issue or plan?

mbasmanova:
@FelixYBW Binwei, PrestoSQL was designed to be ANSI compliant out of the box. It follows SQL specification closely.

FelixYBW:
In this way we should just use prestoSQL functions if Spark's ANSI configuration is set

lviiii:
I tested the case dividing by zero in functions/prestosql. The presto function returns "Infinity" value instead of expected exception error, which don't meet ANSI compliance. @mbasmanova @FelixYBW

mbasmanova:
[...]

PR #14375, opened 2025-08-08
feat: Add a configuration property for Spark ANSI mode

PHILO-HE:
This change prepares for supporting an ANSI-compliant dialect in Velox. The added
configuration property allows function implementations to adapt based on the user's
ANSI mode setting.

See ANSI support discussion: https://github.com/facebookincubator/velox/issues/3869.

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -363,9 +363,11 @@ class QueryConfig {
   static constexpr const char* kPrestoArrayAggIgnoreNulls =
       "presto.array_agg.ignore_nulls";
 
-  /// If true, Velox will use an ANSI-compliant dialect. For example, Velox will
-  /// throw a runtime exception instead of returning null results when the
-  /// inputs to a SQL operator or function are invalid.
+  /// If true, Spark function's behavior is ANSI-compliant, e.g. throws runtime
+  /// exception instead of returning null on invalid inputs.
+  /// Note: This feature is still under development to achieve full ANSI
+  /// compliance. Users can refer to the Spark function documentation to verify
+  /// the current support status of a specific function.
   static constexpr const char* kSparkAnsiEnabled = "spark.ansi_enabled";
 
   /// The default number of expected items for the bloomfilter.
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -36,11 +36,6 @@ class QueryConfig {
   static constexpr const char* kQueryMaxMemoryPerNode =
       "query_max_memory_per_node";
 
-  /// If true, Velox will use an ANSI-compliant dialect. For example, Velox will
-  /// throw a runtime exception instead of returning null results when the inputs
-  /// to a SQL operator or function are invalid.
-  static constexpr const char* kAnsiEnabled = "ansi_enabled";
-
   /// User provided session timezone. Stores a string with the actual timezone
   /// name, e.g: "America/Los_Angeles".
   static constexpr const char* kSessionTimezone = "session_timezone";
@@ -367,6 +362,11 @@ class QueryConfig {
   /// If true, array_agg() aggregation function will ignore nulls in the input.
   static constexpr const char* kPrestoArrayAggIgnoreNulls =
       "presto.array_agg.ignore_nulls";
+
+  /// If true, Velox will use an ANSI-compliant dialect. For example, Velox will
+  /// throw a runtime exception instead of returning null results when the
+  /// inputs to a SQL operator or function are invalid.
+  static constexpr const char* kSparkAnsiEnabled = "spark.ansi_enabled";
 
   /// The default number of expected items for the bloomfilter.
   static constexpr const char* kSparkBloomFilterExpectedNumItems =
@@ -869,10 +869,6 @@ class QueryConfig {
     return get<uint64_t>(kExprMaxCompiledRegexes, 100);
   }
 
-  bool ansiEnabled() const {
-    return get<bool>(kAnsiEnabled, false);
-  }
-
   bool adjustTimestampToTimezone() const {
     return get<bool>(kAdjustTimestampToTimezone, false);
   }
@@ -1028,6 +1024,10 @@ class QueryConfig {
 
   bool prestoArrayAggIgnoreNulls() const {
     return get<bool>(kPrestoArrayAggIgnoreNulls, false);
+  }
+
+  bool sparkAnsiEnabled() const {
+    return get<bool>(kSparkAnsiEnabled, false);
   }
 
   int64_t sparkBloomFilterExpectedNumItems() const {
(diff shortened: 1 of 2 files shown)


velox/core/QueryConfig.h
```diff
@@ -36,6 +36,11 @@ class QueryConfig {
   static constexpr const char* kQueryMaxMemoryPerNode =
       "query_max_memory_per_node";
 
+  /// If true, Velox will use an ANSI-compliant dialect. For example, Velox will
+  /// throw a runtime exception instead of returning null results when the inputs
+  /// to a SQL operator or function are invalid.
+  static constexpr const char* kAnsiEnabled = "ansi_enabled";
```
mbasmanova: I assume this is specific to Spark functions. Presto functions are ANSI compliant already.

If so, let's add spark. prefix like other Spark-specific configs.

There is also a question about how to implement this without causing confusion or wrong results. It seems this change would allow user to request ANSI mode, but the results will not be compiant.

mbasmanova (changes requested):

velox/core/QueryConfig.h
philo-he: @mbasmanova, thanks for your review. I will update the new config to indicate that it is intended for Spark use only.

There will be several upcoming PRs that rely on this configuration to adapt function implementations. The legacy mode and ANSI mode will coexist in Spark for a long period. We plan to introduce branching logic in relevant Velox functions, respecting this configuration to ensure ANSI compliance while minimizing code duplication. It's also possible to add separate ANSI-compliant functions if there isn't much shared code.

mbasmanova: This makes sense, but my question is about something else. How do we ensure that when ANSI is on the query returns correct results. If we land this change, then turning ANSI on will produce wrong results (same as with ANSI off). 

One option is to create an allow list of functions and special forms that support ANSI mode and fail the query if ANSI mode is ON, but query contains a function that's not in allow list. We would need to think about where to place such a check.

CC: @rui-mo

philo-he: @mbasmanova, currently there is a rule in Gluten to fall back to vanilla Spark for the entire query when ANSI is on. Our plan is to firstly support a few common ANSI-compliant functions in Velox before removing this rule to allow the offloading. After this rule is removed, it's quite possible that incorrect results may be produced for functions that either lack ANSI support or have existing bugs, as you pointed out. To address this, maybe we can maintain a blacklist in Gluten to prevent the offloading for these functions, similar to our current approach when ANSI is off. This approach appears more user-friendly, as it allows the query to be executed in fallback mode rather than failing on Velox side.

rui-mo: The ANSI mode-controlled scenarios are described in Spark's documentation at https://github.com/apache/spark/blob/v4.0.0/docs/sql-ref-ansi-compliance.md, and they have a variety of effects on the detailed implementation. Given this, it might be difficult to cover every aspect by adding a check in Velox.

Since Gluten currently fallback for ANSI on, it would more of an experimental feature. We can also add a document in Velox to list the ANSI-compatible functions when supporting them, which should be rather limited for now.

velox/docs/configs.rst
```diff
@@ -977,6 +977,11 @@ Spark-specific Configuration
      - Type
      - Default Value
      - Description
+   * - spark.ansi_enabled
+     - bool
+     - false
+     - If true, Velox will use an ANSI-compliant dialect. For example, Velox will throw a runtime exception instead of
+       returning null results when the inputs to a SQL operator or function are invalid.
```
rui-mo: Would you please mention current status of this config? I suppose it won't take effect for now.

mbasmanova: Do we have a sense of a timeline for updating ALL Spark functions to support this property? How many functions need updating? How do we plan to test this? Can we setup Fuzzer that runs with ansi on and compares the results with Spark?

I'm concerned about users starting to use this property and run into data correctness issues.

mbasmanova:
> We can also add a document in Velox to list the ANSI-compatible functions when supporting them

That would be helpful. Let's come up with a convention for how to indicate in the docs whether a function supports ANSI mode or not.

Can we add an API that takes an Expr tree and returns true if all expressions are ANSI-compliant or not? This would allow non-Gluten users to check compliance easily.

mbasmanova:
Does ANSI mode affect the behavior of any operators? Or is it only for functions?

velox/docs/configs.rst
philo-he: @mbasmanova, my current estimate is at least one quarter to support all related Spark functions, depending on the number of contributors involved. I am still listing the functions that require ANSI support. Like non-ANSI functions, we will port Spark test cases into Velox and enable fuzzer tests to ensure consistency with Spark.

philo-he:
> Can we add an API that takes an Expr tree and returns true if all expressions are ANSI-compliant or not? This would allow non-Gluten users to check compliance easily.

@mbasmanova, sounds great. I will evaluate the feasibility. Thanks for the suggestion.

> Does ANSI mode affect the behavior of any operators? Or is it only for functions?

To the best of my knowledge, ANSI mode primarily affects expressions. Maybe, one exception is implicit type coercion during write operations, which it can also influence. However, this effect is indirect—ANSI actually controls the underlying expressions that the write operation relies on.

PHILO-HE (commented):
@mbasmanova, I've updated this PR. Could you take a look again?

velox/core/QueryConfig.h
```diff
@@ -363,6 +363,11 @@ class QueryConfig {
   static constexpr const char* kPrestoArrayAggIgnoreNulls =
       "presto.array_agg.ignore_nulls";
 
+  /// If true, Velox will use an ANSI-compliant dialect. For example, Velox will
```
mbasmanova: It is not quite accurate to say "Velox will use" here because the setting applies only to Spark functions package, not to core engine. Let's change to something lilke

/// If true, Spark function's behavior is ANSI-compliant, e.g. throws runtime exception instead of returning null on invalid inputs.

Also, let's clarify that this setting is not implemented yet and provide guidance on how to tell which functions support this. Are we going to come up with some way to mark a function as ANSI-compliant in the docs?

philo-he: @mbasmanova, The suggestion makes sense. I've updated the PR accordingly. We will document the ANSI behavior in the existing documentation for the specific function as part of the ANSI-support PR. Thanks.

rui-mo:
> Let's come up with a standard way of doing that before PRs start showing up to ensure all PRs do that in a consistent easy to understand way.

@mbasmanova @PHILO-HE  I assume we could document the "standard way" at below with another PR, and follow this way in subsequent PRs.
https://github.com/facebookincubator/velox/blob/6bcbfe86f0a9ba76ffc1e5a79a8ad79607094a0f/velox/docs/spark_functions.rst?plain=1#L5

The standard way could be to add a label `(ANSI Compatible)` at the function signature if the ANSI ON/OFF behaviors are both supported, and document the details of differences. Example:

Screenshot 2025-08-27 at 17 07 17

How do you think?

mbasmanova:
@rui-mo I like this proposal.

PR #14379, opened 2025-08-08
fix(vector): Vector length should be vector_size_t instead of size_t

MBkkt:
Also some small improvements that was in the same file

### DIFF
--- a/velox/vector/ComplexVector.h
+++ b/velox/vector/ComplexVector.h
@@ -33,16 +33,16 @@ class RowVector : public BaseVector {
 
   RowVector(
       velox::memory::MemoryPool* pool,
-      std::shared_ptr<const Type> type,
+      const TypePtr& type,
       BufferPtr nulls,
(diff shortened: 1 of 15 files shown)


mbasmanova (approved):
Thanks!

MBkkt:
@mbasmanova I noticed that I missed some places with same issue, so I fixed them too

velox/vector/ComplexVector.h
```diff
@@ -33,16 +33,16 @@ class RowVector : public BaseVector {
...
-      size_t length,
+      vector_size_t length,
       std::vector<VectorPtr> children,
       std::optional<vector_size_t> nullCount = std::nullopt)
       : BaseVector(
             pool,
             type,
             VectorEncoding::Simple::ROW,
-            nulls,
+            std::move(nulls),
```
pedroerp, line 45: Good catch :)

PR #14383, opened 2025-08-08
fix: Fix crash by stackoverflow when grouping by nano-timestamp key

zhztheplayer:
TBD


zhztheplayer:
@duxiao1212 @amitkdutta @kewang1024

Would you like to review, thanks. This is a regression caused by https://github.com/facebookincubator/velox/pull/13494.

duxiao1212:
LGTM, thanks for the fix.

jinchengchenghh (approved):

velox/exec/VectorHasher.h
```diff
@@ -677,11 +677,17 @@ template <>
...
 }
-
 template <>
 inline uint64_t VectorHasher::valueId(Timestamp value) {
-  return value.getNanos() % 1'000'000 != 0 ? kUnmappable
-                                           : valueId(value.toMillis());
+  if (FOLLY_UNLIKELY(
+          value.getNanos() % Timestamp::kNanosecondsInMillisecond != 0)) {
+    // The timestamp is in nanosecond precision. The values are
+    // not mappable to uint64.
```
rui-mo: nit: I suppose microsecond precision is also possible, e.g. Spark current_timestamp function returns microseconds. Also, the microseconds are mappable to uint64, but cannot be mapped to millis without precision loss.

rui-mo (approved):

velox/exec/VectorHasher.h
zhztheplayer: > I suppose microsecond precision is also possible

That's a valid point to follow-up. I shall also confirm whether mapping the microseconds here impacts the hive writer partitioner's sanity.

zhztheplayer:
Thank you for your review @duxiao1212 @jinchengchenghh @rui-mo.

velox/exec/VectorHasher.h
rui-mo: Just clarify this is only one nit on comment. We might don't need to map timestamp to microsecond because the Spark partition key is of millisecond precision.

zhztheplayer: > Just clarify this is only one nit on comment.

In that case, may I know your suggestion about the comment? Just to confirm whether we were on the same page.

rui-mo: Perhaps revise the comment to something like below.

> The timestamp is in nanosecond or microsecond precision. The values are not mappable to milliseconds without precision loss.

zhztheplayer: > We might don't need to map timestamp to microsecond because the Spark partition key is of millisecond precision.

That's similar with what I wondered. Ideally hasher is to be general and to avoid specific handlings for caller side. The partitioned writers may need to pre-process the partition keys before sending them to the hasher.

mbasmanova (approved):
Thanks.

PR #14384, opened 2025-08-08
feat(gcs): Add config to support registering custom GcsOAuthCredentialsProvider

marin-ma:
Currently, only a service account credential file is supported for creating the GCS client. In real-world scenarios, users may instead rely on provided OAuth credentials and manage their expiration according to external rules.

This patch adds a `GcsOAuthCredentialsProvider` interface for custom providers that generate OAuth credentials, a config entry `hive.gcs.auth.access-token-provider` to enable using a self-defined `GcsOAuthCredentialsProvider`, and adds a new API `registerGcsOAuthCredentialsProvider` for registering the factory to use the custom `GcsOAuthCredentialsProvider` implementation.

### DIFF
--- a/velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
@@ -36,12 +36,40 @@ using namespace connector::hive;
 namespace gcs = ::google::cloud::storage;
 namespace gc = ::google::cloud;
 
+namespace {
+
 auto constexpr kGcsInvalidPath = "File {} is not a valid gcs file";
+
+folly::Synchronized<
+    std::unordered_map<std::string, GcsOAuthCredentialsProviderFactory>>&
+credentialsProviderFactories() {
+  static folly::Synchronized<
+      std::unordered_map<std::string, GcsOAuthCredentialsProviderFactory>>
+      factories;
+  return factories;
+}
+
+std::shared_ptr<GcsOAuthCredentialsProvider> getCredentialsProviderByName(
+    const std::string& providerName,
+    const std::shared_ptr<connector::hive::HiveConfig>& hiveConfig) {
+  return credentialsProviderFactories().withRLock([&](const auto& factories) {
+    const auto it = factories.find(providerName);
+    VELOX_CHECK(
+        it != factories.end(),
+        "GcsOAuthCredentialsProviderFactory for '{}' not registered",
+        providerName);
+    const auto& factory = it->second;
+    return factory(hiveConfig);
+  });
+}
+
+} // namespace
 
 class GcsFileSystem::Impl {
  public:
-  Impl(const config::ConfigBase* config)
-      : hiveConfig_(std::make_shared<HiveConfig>(
+  Impl(const std::string& bucket, const config::ConfigBase* config)
+      : bucket_(bucket),
+        hiveConfig_(std::make_shared<HiveConfig>(
             std::make_shared<config::ConfigBase>(config->rawConfigsCopy()))) {}
 
   ~Impl() = default;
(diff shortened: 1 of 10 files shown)


velox/connectors/hive/storage_adapters/gcs/GcsOAuthCredentialsProvider.h
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma once
+
+#include <google/cloud/storage/oauth2/credentials.h>
+
+namespace facebook::velox::filesystems {
+
+namespace gcs = ::google::cloud::storage;
+
+class GcsOAuthCredentialsProvider {
```
JkSelf, line 28: Could you help to add some comments for this class and method?

velox/connectors/hive/storage_adapters/gcs/GcsOAuthCredentialsProvider.h
```diff
@@ -0,0 +1,33 @@
...
+ public:
+  virtual ~GcsOAuthCredentialsProvider() = default;
+
+  virtual std::shared_ptr<gcs::oauth2::Credentials> getCredentials(
+      const std::string& bucket) = 0;
+};
+
+} // namespace facebook::velox::filesystems
```
JkSelf, line 36: nit: add new empty line.

velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
```diff
@@ -325,5 +363,20 @@ void GcsFileSystem::rmdir(std::string_view path) {
   }
 }
 
+void registerOAuthCredentialsProvider(
```
JkSelf, line 369: Could we directly move this register function into the RegisterGcsFileSystem.cpp file?

velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
```diff
@@ -36,12 +36,40 @@ using namespace connector::hive;
 namespace gcs = ::google::cloud::storage;
 namespace gc = ::google::cloud;
 
+namespace {
+
 auto constexpr kGcsInvalidPath = "File {} is not a valid gcs file";
 
+folly::Synchronized<
```
JkSelf, line 43: Could we move credentialsProviderFactories and getCredentialsProviderByName method into the GcsOAuthCredentialsProvider class?

JkSelf (commented):
LGTM except some minor comments.

velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
marin-ma: `GcsOAuthCredentialsProvider` only defines the credentials provider interface. ~Moved them into `RegisterGcsFileSystem.cpp` to simplify the code.~

I prefer to keep the provider registry in `GcsFileSystem.cpp` as it's only used for creating the GCSClient. If we separate `getCredentialsProviderByName` and `registerOAuthCredentialsProvider` into different files then `credentialsProviderFactories()` need to be exposed to both of them.

velox/connectors/hive/storage_adapters/gcs/GcsOAuthCredentialsProvider.h
marin-ma: There's already an empty line at the end of this file.

JkSelf:
@majetideepak Could you help to take a look? Thanks.

velox/connectors/hive/storage_adapters/gcs/GcsFileSystem.cpp
```diff
@@ -36,12 +36,40 @@ using namespace connector::hive;
...
+      std::unordered_map<std::string, GcsOAuthCredentialsProviderFactory>>
+      factories;
+  return factories;
+}
+
+std::shared_ptr<GcsOAuthCredentialsProvider> getCredentialsProviderByName(
+    const std::string& providerName,
+    const std::shared_ptr<connector::hive::HiveConfig>& hiveConfig) {
+  return credentialsProviderFactories().withRLock([&](const auto& factories) {
+    const auto it = factories.find(providerName);
```
majetideepak, line 59: nit: Let's also check that the providerName is not empty.

marin-ma:
@majetideepak Addressed comments. Can we mark this as ready to merge if no further comments? Thanks!

velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.cpp
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include "velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.h"
```
majetideepak, line 17: nit: We don't need to include this header.

majetideepak (approved):

velox/connectors/hive/storage_adapters/gcs/RegisterGcsFileSystem.cpp
marin-ma: Same as abfs, it's needed for the function signature. If it's not included then the compilation will fall with `VELOX_ENABLE_GCS` undefined.

peterenescu:
Looks like touches Hive connector, confirmed with @Yuhta that this good

PR #14393, opened 2025-08-09
fix: Use customized doubleCompare for comparing

skyelves:
Summary: fix: Use customized doubleCompare for comparing

Differential Revision: D79934792

### DIFF
--- a/velox/functions/prestosql/aggregates/NumericHistogramAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/NumericHistogramAggregate.cpp
@@ -58,7 +58,7 @@ class NumericHistogram {
       if (right != nullptr) {
         penalty_ = computePenalty(value, weight, right->value_, right->weight_);
       } else {
-        penalty_ = std::numeric_limits<double>::infinity();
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/aggregates/NumericHistogramAggregate.cpp
```diff
@@ -70,10 +70,12 @@ class NumericHistogram {
 
     struct Compare {
       bool operator()(const Entry* current, const Entry* other) const {
-        if (current->penalty_ == other->penalty_) {
+        auto penaltyComparison =
```
kevinwilfong: nit: const

kevinwilfong (approved):
Thanks for the fix!

kevinwilfong (changes requested):
Could you add a unit test to cover the case that was failing?

skyelves:
> Could you add a unit test to cover the case that was failing?

added

velox/functions/prestosql/aggregates/NumericHistogramAggregate.cpp
skyelves: resolved

PR #14394, opened 2025-08-09
Simplify dependencies in velox/runner/Runner.h

mbasmanova:
Summary: Extract BlockingReason and DriverStats into their own header files to avoid including Driver.h when only stats are needed.

Differential Revision: D79936021


xiaoxmeng (approved):
@mbasmanova thanks!

velox/exec/BlockingReason.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/exec/BlockingReason.h"
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::velox::exec {
+
+std::string blockingReasonToString(BlockingReason reason) {
```
MBkkt: `string_view` can be used?

mbasmanova: Replaced with VELOX_DECLARE_ENUM_NAME

PR #14399, opened 2025-08-09
misc: Use make_shared to create PeeledEncoding

MBkkt:
PeeledEncoding sizeof is 32 bytes (small), so even if it's used with weak_ptr it doesn't make sense to allocate control block in separate allocation

### DIFF
--- a/velox/expression/PeeledEncoding.cpp
+++ b/velox/expression/PeeledEncoding.cpp
@@ -26,7 +26,7 @@ namespace facebook::velox::exec {
     DecodedVector& decodedVector,
     bool canPeelsHaveNulls,
     std::vector<VectorPtr>& peeledVectors) {
-  auto peeledEncoding = std::make_shared<PeeledEncoding>();
+  auto peeledEncoding = std::make_shared<PeeledEncoding>(PrivateTag{});
   if (peeledEncoding->peelInternal(
           vectorsToPeel,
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
Thanks.

MBkkt:
@mbasmanova 
I noticed only after build error: this constructor was private.

It can be emulated with suggested pattern (constructor is impossible to call from public/protected scope, but it is written different way). 

Is such pattern ok for velox in this place?
I just want to clarify it, because the code that you approved was changed.

velox/expression/PeeledEncoding.cpp
```diff
@@ -26,7 +26,7 @@ namespace facebook::velox::exec {
     DecodedVector& decodedVector,
     bool canPeelsHaveNulls,
     std::vector<VectorPtr>& peeledVectors) {
-  std::shared_ptr<PeeledEncoding> peeledEncoding(new PeeledEncoding());
+  auto peeledEncoding = std::make_shared<PeeledEncoding>(PrivateTag{});
```
mbasmanova, line 29: This is a clever trick. I didn't know about it.

It feels an overkill here though. I suggest to keep the original code because it is shorter and simpler.

MBkkt: Ok

PR #14400, opened 2025-08-09
feat: Add DATE type support for TextReader

Black-Kamous:
Summary: Add DATE type support for TextReader, reading 'YYYY-MM-DD' format date in text file with Velox type 'DATE()' now returns int32_t days after epoch, instead of 'NULL'. Unit test is added in velox/dwio/text/tests/reader/TextReaderTest.cpp .


Black-Kamous:
@YiChengLee03 @zacw7  Would you please take a look at this? Thanks!

PR #14403, opened 2025-08-09
fix: Make access O(1) instead of O(n), it should reduce overhead for large projections

MBkkt:
…and keep it similar for small.

Memory overhead is just additional ~16.125 bytes per type in row.

Also I don't think lazy is good here, but I made it this way because it was TODO in previous attempt

It will be nice if exists some benchmarks for this

Context: https://github.com/facebookexperimental/verax/pull/118#discussion_r2264184373

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -410,14 +410,49 @@ RowType::RowType(std::vector<std::string>&& names, std::vector<TypePtr>&& types)
 }
 
 RowType::~RowType() {
-  if (auto* parameters = parameters_.load()) {
-    delete parameters;
-  }
+  auto* parameters = parameters_.load(std::memory_order_acquire);
+  delete parameters;
+
+  auto* nameToIndex = nameToIndex_.load(std::memory_order_acquire);
+  delete nameToIndex;
 }
 
-std::unique_ptr<std::vector<TypeParameter>> RowType::makeParameters() const {
-  return std::make_unique<std::vector<TypeParameter>>(
+const std::vector<TypeParameter>* RowType::ensureParameters() const {
+  auto newParameters = std::make_unique<std::vector<TypeParameter>>(
       createTypeParameters(children_));
+
+  std::vector<TypeParameter>* oldParameters = nullptr;
+  if (!parameters_.compare_exchange_strong(
+          oldParameters,
+          newParameters.get(),
+          std::memory_order_acq_rel,
+          std::memory_order_acquire)) {
+    return oldParameters;
+  }
+
+  return newParameters.release();
+}
+
+const RowType::NameToIndex* RowType::ensureNameToIndex() const {
+  auto newNameToIndex = std::make_unique<NameToIndex>();
+  newNameToIndex->reserve(names_.size());
+  for (uint32_t i = 0; const auto& name : names_) {
+    if (auto* oldNameToIndex = nameToIndex_.load(std::memory_order_acquire)) {
+      return oldNameToIndex;
+    }
+    newNameToIndex->emplace(NameIndex{name, i++});
+  }
+
+  NameToIndex* oldNameToIndex = nullptr;
+  if (!nameToIndex_.compare_exchange_strong(
+          oldNameToIndex,
+          newNameToIndex.get(),
+          std::memory_order_acq_rel,
+          std::memory_order_acquire)) {
+    return oldNameToIndex;
+  }
+
+  return newNameToIndex.release();
 }
 
 namespace {
@@ -448,10 +483,8 @@ std::string makeFieldNotFoundErrorMessage(
 } // namespace
 
 const TypePtr& RowType::findChild(folly::StringPiece name) const {
-  for (uint32_t i = 0; i < names_.size(); ++i) {
-    if (names_.at(i) == name) {
-      return children_.at(i);
-    }
+  if (auto i = getChildIdxIfExists(name)) {
+    return children_[*i];
   }
   VELOX_USER_FAIL(makeFieldNotFoundErrorMessage(name, names_));
 }
@@ -471,7 +504,7 @@ bool RowType::isComparable() const {
 }
 
 bool RowType::containsChild(std::string_view name) const {
-  return std::find(names_.begin(), names_.end(), name) != names_.end();
+  return nameToIndex().contains(NameIndex{name, 0});
 }
 
 uint32_t RowType::getChildIdx(std::string_view name) const {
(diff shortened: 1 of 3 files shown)


mbasmanova:
CC: @kKPulla @littlesamo

velox/type/Type.h
```diff
@@ -1082,6 +1083,34 @@ class MapType : public TypeBase<TypeKind::MAP> {
...
+  };
+
+  struct NameIndexHasher {
+    size_t operator()(const NameIndex& nameIndex) const {
+      folly::f14::DefaultHasher<std::string_view> hasher;
+      return hasher(std::string_view{nameIndex.data, nameIndex.size});
+    }
+  };
+
+  // TODO: Consider using absl::flat_hash_set instead.
```
Yuhta, line 1123: What's the advantage of this?

Yuhta (commented):
We need to be very careful about such changes, it caused memory regression last time someone implemented it : #8245

MBkkt:
@Yuhta I saw this, and it was different implementation, it was copy string (so every large field name memory was duplicated), and map itself was in a few times larger (with entry sizeof 48 (instead of 16) and map itself was F14VectorMap instead of F14ValueMap)

velox/type/Type.h
MBkkt: From my experience they work faster. 

This todo requires benchmarks, by default I used folly tables, because they're used a lot in velox.

MBkkt:
@mbasmanova maybe you know what do I need to do to move it forward?

mbasmanova:
@MBkkt This one is tricky as we need to make sure it won't cause a big SEV like last time.

@bikramSingh91 Do you know how can we test this to figure out if this is Ok to merge?

MBkkt:
@mbasmanova @bikramSingh91 Hi, maybe you have some updates for this PR?
If not what actions can be done, to move on?

bikramSingh91:
@MBkkt Apologies for the delay, I have reached out to our internal POC for recommendation on testing. WIll report back as soon as I hear back from them.

MBkkt:
@bikramSingh91 Hi, thanks for answer, I also noticed that you imported PR.

This PR is quite old, there's no conflict yet, but if it will be necessary to rebase, please reach me out

velox/type/Type.h
```diff
@@ -1151,16 +1180,19 @@ class RowType : public TypeBase<TypeKind::ROW> {
...
+    const auto* parameters = parameters_.load(std::memory_order_acquire);
+    if (parameters) [[likely]] {
+      return *parameters;
+    }
+    return *ensureParameters();
+  }
+
+  const NameToIndex& nameToIndex() const {
+    const auto* nameToIndex = nameToIndex_.load(std::memory_order_acquire);
+    if (nameToIndex) [[likely]] {
```
bikramSingh91, line 1204: nit: FOLLY_LIKELY for consistency, here and above

velox/type/Type.cpp
```diff
@@ -326,14 +326,49 @@ RowType::RowType(std::vector<std::string>&& names, std::vector<TypePtr>&& types)
...
+  return newParameters.release();
+}
+
+const RowType::NameToIndex* RowType::ensureNameToIndex() const {
+  auto newNameToIndex = std::make_unique<NameToIndex>();
+  newNameToIndex->reserve(names_.size());
+  for (uint32_t i = 0; const auto& name : names_) {
+    if (auto* oldNameToIndex = nameToIndex_.load(std::memory_order_acquire)) {
+      return oldNameToIndex;
+    }
```
bikramSingh91: i am guessing you added this to fail fast, but I wonder how likely is that scenario VS the cost doing a load on every iteration.

bikramSingh91 (commented):
@MBkkt FYI, the regression tests I ran that were added as part of that SEV came out all clear, so we should be good to move forward with this.

cc: @mbasmanova @Yuhta

velox/type/Type.h
mbasmanova: @bikramSingh91 We are trying to reduce dependency on folly, hence, prefer standard C++ over folly.

velox/type/Type.cpp
MBkkt: acquire atomic load from L1 cache without contention on x86 quite cheap (1 cycle?) + always false branch (1-2 cycles) compared to insertion string_view in flat hashset (from 20 cycles to "endless"?).

If contention "happens" we want to fail I think as fast as we can

MBkkt: Although I think unlikely here makes sense

bikramSingh91: @MBkkt makes sense, thanks for the explanation!
Feel free to add `unlikely` here.
Also, can you please rebase to latest. Although there are no conflicts but would be good to get fresh signals before merging.
Thanks for working through this with us!

MBkkt: @bikramSingh91 I rebased and added unlikely

bikramSingh91 (approved):

kgpai:
@assignUser  Did the centos9 image change. Interesting it says gh not available now: 
```
 Status: Downloaded newer image for ghcr.io/facebookincubator/velox-dev:centos9
  ghcr.io/facebookincubator/velox-dev:centos9
  ... 
  gh: Not Found (HTTP 404)
  ```

assignUser:
That's an API 404, fix is #14706

bikramSingh91:
Thank you @kgpai and @assignUser for looking into this.

@MBkkt do you mind doing another rebase to cover the fix in https://github.com/facebookincubator/velox/pull/14706 ?

MBkkt:
@bikramSingh91 I rebased, but mentioned PR isn't merged yet

bikramSingh91:
@MBkkt Oh, looks like its merged now, can you rebase again?
(at this point I really wish github had a rebase button)

MBkkt:
@bikramSingh91 

> Oh, looks like its merged now, can you rebase again?

rebased

>  (at this point I really wish github had a rebase button)

It doesn't have rebase button, but it can be enabled to have "merge main" button.
Maybe it makes sense to enable it instead?

I think it doesn't have rebase button, because rebase will break local branches.
(for merge it's ok, you just can call `git pull` later)

MBkkt:
@assignUser It seems like mentioned PR doesn't help with 404 for `fuzzer jobs / build`

assignUser:
Oh you uncovered another edge case because the PR is not from your fork but serenedb.

MBkkt:
> Oh you uncovered another edge case because the PR is not from your fork but serenedb.

@assignUser Fix https://github.com/facebookincubator/velox/pull/14805

MBkkt:
@bikramSingh91 I think all related jobs in github actions are passed with my patch for fuzzer job (fix for 404).

What is the next step?

bikramSingh91:
> @bikramSingh91 I think all related jobs in github actions are passed with my patch for fuzzer job (fix for 404).
> 
> What is the next step?

@MBkkt Thanks for following up on this. Leave the rest to me, I'll import this diff, wait for internal tests to succeed and initiate merge.

PR #7477, opened 2023-11-08
Unnest array of rows based on unnestArrayOfRows

aditi-pandit:
Velox unnests array of row type into a single column of the
element row type. Presto also supports unnesting such arrays
into multiple columns one for each child type of element row
type. Adds unnestArrayOfRows to add such unnesting support
in Unnest PlanNode.

Needed for https://github.com/prestodb/presto/issues/20643

PR #14408, opened 2025-08-10
feat: Unnest works for Row type

MBkkt:
I think idea is pretty clear and it should allow to unnest something like array<row> to just columns with easy (just two unnest, first to unnest array, second to unnest row)

TODO:
I need to write tests, to make sure it works :)

### DIFF
--- a/velox/core/PlanNode.cpp
+++ b/velox/core/PlanNode.cpp
@@ -1374,6 +1374,12 @@ UnnestNode::UnnestNode(
   std::vector<std::string> names;
   std::vector<TypePtr> types;
 
+  const auto expectedSize = replicateVariables_.size() + unnestNames_.size() +
+      static_cast<size_t>(ordinalityName_.has_value()) +
+      static_cast<size_t>(emptyUnnestValueName_.has_value());
+  names.reserve(expectedSize);
+  types.reserve(expectedSize);
+
   for (const auto& variable : replicateVariables_) {
     names.emplace_back(variable->name());
     types.emplace_back(variable->type());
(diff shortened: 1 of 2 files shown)


mbasmanova:
Are you looking to unnest array(row(a, b, c)) into 3 columns: a, b, c?

If so, I believe @aditi-pandit has been working on that in https://github.com/facebookincubator/velox/pull/7477

MBkkt:
@mbasmanova Yes, but I think this approach is more generic, and allows to do 
map<string, row> unnest too. And it's more granular 

So I think mentioned PR is more like optimization that can be compared in performance with two unnest (first unnest array, second unnest row).

mbasmanova:
> this approach is more generic, and allows to do map<string, row> unnest too.

UNNEST is reasonably well defined SQL clause. It would be nice to keep its implementation in Velox consistent with other engines. 

I wish UNNEST(map(k, row)) can be implemented using existing building blocks, but we seem to be lacking functions that allow to manipulate structs, e.g. [map_entries](https://facebookincubator.github.io/velox/functions/presto/map.html#map_entries) can be used to convert map(key, row(a, b, c)) to row(key, row(a, b, c)), but we don't have flatten function that would turn row(key, row(a, b, c)) into row(key, a, b, c). If we did, then unnest(map(k, row)) would be just unnest(flatten(map_entries(map(k, row)).

That said, it might be reasonable to extend UNNEST to handle both array(row) and map(k, row). However, it could be better to add flatten function to flatten nested structs similar to [arrays](https://facebookincubator.github.io/velox/functions/presto/array.html#flatten).

Reference: 
- [Querying ARRAY-type fields in a struct in BigQuery](https://cloud.google.com/bigquery/docs/arrays#querying_array-type_fields_in_a_struct)
- [Unnest in Presto](https://trino.io/docs/current/sql/select.html#unnest)

MBkkt:
Unnest here is part of execution and it's abstract.
Why do you consider execution that converts `row(row(c1, c2), row(c3, c4))` to `row(c1, c2, c3, c4)` as something bad, because it's non-standard? 

SQL semantics is different for different engines and should mostly controlled by user of execution engine (parser, catalog, optimizer). 
So I think it's completely ok if execution can do more than concrete implementation, because execution isn't specification of such systems, but just way to provide some implementation for them?

The simple example is custom nodes/operators or something specific like geo joins.

> However, it could be better to add flatten function to flatten nested structs similar to [arrays](https://facebookincubator.github.io/velox/functions/presto/array.html#flatten).

Sounds reasonable, but it feels like implementing such function more complicated task than extend unnest operator (< 50 loc)

MBkkt:
> we don't have flatten function that would turn row(key, row(a, b, c)) into row(key, a, b, c).

I understand that function looks for this better, but it can be done with unnest operator (with provided patch)

MBkkt:
If it feels like bad default I can add bool settings (like `unnestRow`) for node which will be false by default (although I think it's unnecessary, because user control which columns they want to unnest)

mbasmanova:
I noticed that DuckDB supports UNNEST for structs. It also supports recursive unnest:

https://duckdb.org/docs/stable/sql/query_syntax/unnest.html#unnesting-structs

PR #14416, opened 2025-08-11
misc: Add string encodings flag in the fuzzer runner

rui-mo:
Enables command-line configuration of the string encodings used in the 
generated vector.

### DIFF
--- a/velox/expression/fuzzer/FuzzerRunner.cpp
+++ b/velox/expression/fuzzer/FuzzerRunner.cpp
@@ -17,6 +17,7 @@
 #include "velox/expression/fuzzer/FuzzerRunner.h"
 
 #include "velox/expression/fuzzer/ExpressionFuzzer.h"
+#include "velox/expression/signature_parser/ParseUtil.h"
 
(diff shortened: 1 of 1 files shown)


velox/expression/fuzzer/FuzzerRunner.cpp
```diff
@@ -162,6 +163,12 @@ DEFINE_string(
     "of functions at every instance. Number of tickets must be a positive "
     "integer. Example: eq=3,floor=5");
 
+DEFINE_string(
+    string_encodings,
+    "0",
+    "Comma-separated list of string encodings to use in generated vector. "
```
jinchengchenghh, line 169: Could you add an example?

rui-mo: > Supported encodings: ASCII(0), UNICODE_CASE_SENSITIVE(1), EXTENDED_UNICODE(2), MATHEMATICAL_SYMBOLS(3), ALPHABETIC(4), NUMERIC(5).

These are examples: 0, 1, ..., 5.

PR #14440, opened 2025-08-12
fix: Improve tzdb to be compatible with old tzdata.zi

izchen:
Current tzdata.zi file parsing assumes `Rule`, `Zone`, and `Link` records appear in that order. Older tzdata versions may place `Zone` records before `Rule`. This PR relaxes that assumption to support parsing older tzdata.zi file.

### DIFF
--- a/velox/external/tzdb/tzdb.cpp
+++ b/velox/external/tzdb/tzdb.cpp
@@ -724,7 +724,8 @@ static void __parse_link(tzdb& __tzdb, std::istream& __input) {
   __tzdb.links.emplace_back(std::move(__name), std::move(__target));
 }
 
-static void __parse_tzdata(
+template <int recordType>
+static void __parse_tzdata_record_type(
     tzdb& __db,
     __rules_storage_type& __rules,
     std::istream& __input) {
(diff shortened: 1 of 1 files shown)


pedroerp:
@kevinwilfong could you take a look at this one?

FelixYBW:
Thank you for the fix. The issue makes much trouble to Gluten

velox/external/tzdb/tzdb.cpp
```diff
@@ -745,26 +746,52 @@ static void __parse_tzdata(
...
+        }
         break;
 
       default:
-        std::__throw_runtime_error("corrupt tzdb: unexpected input");
+        if constexpr (recordType == 'z') {
+          std::__throw_runtime_error("corrupt tzdb: unexpected input");
+        } else {
+          __skip_line(__input);
+        }
```
kevinwilfong: Why only throw in the 'z' case? It seems like if it's invalid in the 'z' case it'll be invalid in all cases so you may as well leave it as it was to fail fast

velox/external/tzdb/tzdb.cpp
```diff
@@ -745,26 +746,52 @@ static void __parse_tzdata(
...
 }
 
+static void __parse_tzdata(
+    tzdb& __db,
+    __rules_storage_type& __rules,
+    std::istream& __input) {
+  const std::string buf{std::istreambuf_iterator<char>(__input), {}};
+  { std::istringstream iss(buf); __parse_tzdata_record_type<'r'>(__db, __rules, iss); }
+  { std::istringstream iss(buf); __parse_tzdata_record_type<'z'>(__db, __rules, iss); }
+  { std::istringstream iss(buf); __parse_tzdata_record_type<'l'>(__db, __rules, iss); }
```
kevinwilfong, line 797: Can you add a comment here that the parsing logic assumes rules are parsed before zones.

kevinwilfong (commented):
Thanks for the fix!

Daydream0506:
Which versions of tzdata does this patch support?​

velox/external/tzdb/tzdb.cpp
izchen: added comment

velox/external/tzdb/tzdb.cpp
izchen: Only `Z` have continuation lines.

- When parsing `R` or `L`, unexpected tokens are typically Zone continuations → skip the line.
- When parsing `Z`, any token not matched by the earlier cases is invalid (continuations are consumed inside __parse_zone) → throw.

izchen: added comment

izchen:
> Which versions of tzdata does this patch support?​

Verified with **tzdata-2021e-1.el8.noarch**; I haven’t tested other versions yet.

izchen:
@kevinwilfong Thanks for the review.

kevinwilfong (approved):
Thanks!

izchen:
Thank you everyone!

PR #14445, opened 2025-08-12
fix: Respect output buffer size in merge with spill and improvements

xiaoxmeng:
Summary:
Current merge operator doesn't respect output batch byte size when read merged result from the streaming sources as well as the spilled sources. This cause memory increase when we do local spill as the number output rows could be large as the default is 1024. This PR fix this by respecting the output batch size in bytes by estimating row size from the merge source which use the averaged row size from the first batch from each streaming or spilled source as the estimation. This significantly reduce the peak memory of pyspark velox use case which leverage local spill to cap the memory usage.

This PR also add runtime stats to tell how much wall time spent in streaming sources read and spilled source reads plus some code refactor.

The followup is to add async spilled read error handling and consider to provide a separate config to tune the output batch size of local merge operator which is single threaded and might not necessary to follow the regular operator batch size. This is a tradeoff between memory and cpu efficiency though.

Differential Revision: D80068538

### DIFF
--- a/velox/docs/monitoring/stats.rst
+++ b/velox/docs/monitoring/stats.rst
@@ -193,6 +193,28 @@ These stats are reported only by IndexLookupJoin operator
      -
      - The number of lazy decoded result batches returned from the storage client.
 
+Merge
+-----
+These stats are reported only by Merge operator
+
+.. list-table::
+   :widths: 50 25 50
+   :header-rows: 1
+
+   * - Stats
+     - Unit
+     - Description
+   * - streamingSourceReadWallNanos
+     - nanos
+     - The number of a spillable operators that don't support spill because of
+       spill limitation. For instance, a window operator do not support spill
+       if there is no partitioning.
+   * - spilledSourceReadWallNanos
+     - nanos
+     - The running wall time of the merge operator reading from the spilled source to
+       produce the final output. This only applies when spilling is enabled for local
+       merge.
+
 Spilling
 --------
 These stats are reported by operators that support spilling.
--- a/velox/exec/Merge.cpp
+++ b/velox/exec/Merge.cpp
@@ -53,7 +53,9 @@ Merge::Merge(
           planNodeId,
           operatorType,
           spillConfig),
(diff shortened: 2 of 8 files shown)


velox/docs/monitoring/stats.rst
```diff
@@ -193,6 +193,32 @@ These stats are reported only by IndexLookupJoin operator
...
+
+   * - Stats
+     - Unit
+     - Description
+   * - streamingSourceReadWallNanos
+     - nanos
+     - The number of a spillable operators that don't support spill because of
+       spill limitation. For instance, a window operator do not support spill
+       if there is no partitioning.
+       The running wall time of a writer operator from creation to close.
```
kevinwilfong: Leftover copypasta?

velox/exec/Merge.cpp
```diff
@@ -196,12 +205,19 @@ void Merge::maybeStartNextMergeSourceGroup() {
   std::vector<std::unique_ptr<SourceStream>> cursors;
   cursors.reserve(sources.size());
   for (auto* source : sources) {
-    cursors.push_back(
-        std::make_unique<SourceStream>(source, sortingKeys_, outputBatchSize_));
+    cursors.push_back(std::make_unique<SourceStream>(
+        source, sortingKeys_, maxOutputBatchRows_));
   }
 
+  // TODO: considr to provide a config other than the regular operator batch
```
kevinwilfong: nit: considr -> consider

velox/exec/Merge.cpp
```diff
@@ -450,11 +533,15 @@ SpillMerger::SpillMerger(
       pool_(pool->shared_from_this()),
       sources_(createMergeSources(spillReadFilesGroup.size())),
       batchStreams_(createBatchStreams(std::move(spillReadFilesGroup))),
+      // TODO: considr to provide a config other than the regular operator batch
```
kevinwilfong: nit: considr -> consider

kevinwilfong (approved):
Thanks for the fix! Some minor nits but the logic looks good

PR #14451, opened 2025-08-13
feat(expr): Support non-ANSI Spark semantics for cast failures in complex types

kevinwilfong:
Summary:
In Spark with ANSI mode turned off, when casting the elements of a complex type from one
type to another, if the cast fails, that particularly element is null in the result.

This differs from the current semantics of Velox's CastExpr (with try cast enabled) which 
follows the ANSI standard, which is to return null as the result for the entire complex type.

E.g.
TRY_CAST(ARRAY('1', 'abc', '2') as ARRAY(BIGINT))

This currently returns NULL in Velox. In Spark this would return ARRAY(1, NULL, 2).  It's also 
worth noting Spark would return this whether it was a TRY_CAST or a CAST.

This change adds a setting in the CastHooks `applyTryCastRecursively` which when true
applies the Spark semantics where a failed cast only effects the value in a complex type that
caused the failure.  This is false in PrestoCastHooks and true in SparkCastHooks.

Differential Revision: D80135098

### DIFF
--- a/velox/functions/sparksql/tests/SparkCastExprTest.cpp
+++ b/velox/functions/sparksql/tests/SparkCastExprTest.cpp
@@ -881,5 +881,56 @@ TEST_F(SparkCastExprTest, boolToTimestamp) {
       }));
 }
 
+TEST_F(SparkCastExprTest, recursiveTryCast) {
+  // Test array elements.
+  testCast(
+      makeArrayVector<StringView>({
+          {"1", "2", "3"},
+          {"4", "a", "6"},
+          {"b", "c", "d"},
+      }),
+      makeNullableArrayVector<int64_t>({
+          {1, 2, 3},
+          {4, std::nullopt, 6},
+          {std::nullopt, std::nullopt, std::nullopt},
+      }));
+
+  // Test map values (Spark doesn't allow casting if the map keys can become
+  // null).
+  testCast(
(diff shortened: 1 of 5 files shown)


velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -881,5 +881,56 @@ TEST_F(SparkCastExprTest, boolToTimestamp) {
...
+          {"4", "a", "6"},
+          {"b", "c", "d"},
+      }),
+      makeNullableArrayVector<int64_t>({
+          {1, 2, 3},
+          {4, std::nullopt, 6},
+          {std::nullopt, std::nullopt, std::nullopt},
+      }));
+
+  // Test map values (Spark doesn't allow casting if the map keys can become
```
kgpai, line 898: nit: For completeness, can you add a test where it throws for an invalid map key ?

kevinwilfong: It's a query compile time check in Spark. Velox allows it, but Spark would never use it.  I mentioned it because I'm not sure what the behavior in Spark would be if multiple keys failed and got converted to null, it turns out Spark just avoids the question entirely.

zhouyuan:
Cc @PHILO-HE

PR #14457, opened 2025-08-13
fix: Fix result mismatch in merge join with filter when output batch stop at last row of right side match

ccat3z:
In a merge join with a filter, when the output batch has just finished iter the right side match and the left row does not match any right row, `applyFilter()` will call `onMiss(output index)` in next output batch. This causes a result mismatch. Here are two examples.

1. Output batch ends with current left row

```
left table `t`:

t_c0 | t_c1
2    | 5
2    | 3

right table `u`:

u_c0 | u_c1
2    | 3
2    | 4

sql:
select * from t left join u where t_c0 = u_c0 and t_c1 < u_c1
-- outputBatchSize = 2 
```

The output batch of `t_c0 = u_c0` is:

```
t_c0 | t_c1 | u_c0 | u_c1
# Batch 1
2    | 5    | 2    | 3
2    | 5    | 2    | 4
# Batch 2
2    | 3    | 2    | 3
2    | 3    | 2    | 4
```

Original 

2. aa


velox/exec/MergeJoin.h
```diff
@@ -424,18 +427,17 @@ class MergeJoin : public Operator {
...
     template <typename TOnMiss>
     void noMoreFilterResults(TOnMiss onMiss) {
-      if (!currentRowPassed_) {
+      if (!currentRowPassed_ && currentRow_ >= 0) {
         onMiss(currentRow_);
       }
 
+      // Output index is meanless for next batch, so reset it to -1.
       currentRow_ = -1;
-      currentRowPassed_ = false;
```
JkSelf, line 500: Just wondering — why was this line removed?

velox/exec/MergeJoin.h
```diff
@@ -424,18 +427,17 @@ class MergeJoin : public Operator {
...
-    // next batch of output will start with a new left-side row or there will
-    // be no more batches. Calls 'onMiss' for the last left-side row if the
-    // filter failed for all matches of that row.
+    // Called when all rows from the current output batch are processed. Calls
+    // 'onMiss' for the last left-side row if the filter failed for all matches
+    // of that row.
     template <typename TOnMiss>
     void noMoreFilterResults(TOnMiss onMiss) {
-      if (!currentRowPassed_) {
+      if (!currentRowPassed_ && currentRow_ >= 0) {
```
JkSelf, line 497: In which use cases would currentRow_ < 0 occur when calling the noMoreFilterResults method?

velox/exec/MergeJoin.cpp
```diff
@@ -1383,20 +1383,48 @@ RowVectorPtr MergeJoin::applyFilter(const RowVectorPtr& output) {
...
+      // contain a different key match.
+      if (!leftMatch_) {
+        onMiss(args...);
+        return;
+      }
+
+      // 2. leftMatch_ may not be nullopt, but may be related to a different
+      // (subsequent) left key. So we check if the last row in the batch has the
+      // same left row number as the last key match.
+      if (!joinTracker_->isCurrentLeftMatch(numRows - 1)) {
```
JkSelf: Could we combine these two if conditions into one, like in the previous version of the code?

JkSelf (commented):
@ccat3z Thanks for your fix. Leaving some comments.

velox/exec/MergeJoin.h
ccat3z: After changing noMoreFilterResults() to be called for each batch, the next batch may still has current left row, so `currentRowPassed_` which means current left row has passed record should not be reset.

velox/exec/MergeJoin.h
ccat3z: Because currentRowPassed_ is only updated when processFilterResult occur a new left row, if a batch does not call processFilterResult even once (e.g. no matching rows on the right side), it will occur currentRow_ < 0.

velox/exec/MergeJoin.cpp
ccat3z: Sure, I've updated the code.

JkSelf (approved):
LGTM. Thanks for your fix.

JkSelf:
@xiaoxmeng @pedroerp Could you help to review this PR? Thanks.

ccat3z:
@xiaoxmeng @pedroerp Could you help to review this PR? Thanks.

Issue #14149, opened 2025-07-18
[cuDF] Improve coverage and performance of expression evaluation


GregoryKimball:
### Description

**Is your feature request related to a problem? Please describe.**

The cuDF backend for Velox uses an [ExpressionEvaluator](https://github.com/facebookincubator/velox/blob/main/velox/experimental/cudf/exec/ExpressionEvaluator.h) to process operators like FilterProject. 

Internally, the expression evaluation maps the predicate in a `core::FilterNode` to a `cudf::ast::expression`. However, `cudf::ast::expression` only implements a subset of the operators supported by `core::FilterNode`. For instance, strings manipulation and ternary operator are not supported in `cudf::ast::expression`. 

As of July 2025, ExpressionEvaluator has the capability to precompute columns if an unsupported operator occurs in a leaf node of the expression tree. However, there are gaps where precomputing isn't yet supported and when an AST-incompatible node occurs in a branch node.

**Describe the solution you'd like**

| Project | Summary | Status | 
|--|--|--|
| Add expression evaluator that uses cuDF APIs without AST | Implement an expression evaluator that traverses the expression tree and calls one or more cuDF APIs per operator. This approach will be simpler and more feature-complete than using AST and non-AST APIs together as in the design as of July 2025. For AST-compatible expressions, we should expect that cuDF API per operator will be than the AST-based `cudf::compute_column` | |
| Add option to use cuDF [JIT transform](https://github.com/rapidsai/cudf/blob/55ad7934a4ea9902640620f5a7762cf49c0ecb0e/cpp/include/cudf/transform.hpp#L64) for evaluating compatible expressions | JIT transform in cuDF supports arithmetic, comparisons and several string manipulation functions [link](https://github.com/rapidsai/cudf/tree/branch-25.08/cpp/examples/string_transforms). We should provide tools in `ExpressionEvaluator` to detect if an expression is compatible with JIT transform, perform the code gen to convert the expression into a CUDA C++ UDF, and then execute the transform. | |  
| Add a function registry pattern | Evaluator functions like `pushExprToTree` [link](https://github.com/facebookincubator/velox/blob/24a3ec38bccad68677e53d831881ed7e30b6b632/velox/experimental/cudf/exec/ExpressionEvaluator.cpp#L390) contain large switch statements. We should prefer a design that improves how we map between expression names and cuDF functions. Potentially, we could use a cuDF function registry to build interoperability between Velox UDFs and cuDF UDFs. | |
[...]

PR #14462, opened 2025-08-13
feat(cudf): Add support for Nested precompute expressions cardinality and split

devavret:
Adds support for evaluation of nested expressions using standalone cuDF APIs. Previously, standalone cuDF APIs could only be used in the leaf node of expressions while the rest of the evaluation would be done using cuDF AST based compute_column API. 

Contributes to #14149


zoltan:
thanks! it's great that we are getting closer and closer to an out-of-the-box Velox cuDF experience!

karthikeyann (commented):
nit: move over more functions to registry mechanism and move the factories to new file.

Yuhta:
@devavret Can you resolve the merge conflict?

devavret:
@Yuhta done!

PR #14463, opened 2025-08-13
fix: Revert signature filtering due to bias fuzzer flakiness

peterenescu:
Summary:
A recent change increased biased fuzzer test coverage when the function under inspection is in the skip list. This change would allow a normal expression fuzzer execution rather than early termination if the function is not in the list of testable functions.

Unfortunately, this would not work without major changes to the expression fuzzer like removing flags `try-with-retry`  and disabling complex/variadic signatures due to irreconcilable differences between expression evaluation and Presto SQL execution (for example different `try` error handling between Presot and Velox).

In order to not lose coverage in the bias fuzzer, let's revert this change so we don't have flakiness unrelated to the diff changes at hand.

Differential Revision: D80191788


czentgr (approved):
Thanks. That's the one that makes the fuzzer CI is continuously failing?
If you could please add the original PR that added this originally? It only says "a recent change" but might be good to have a direct link than find it through blame.

PR #14468, opened 2025-08-14
fix: Update the config name and documentation of flatmap preserving option in hive connector

kKPulla:
Summary:
1. Prefixed the config name with "hive.",  to be consistent with other options in hive connector
2. Added an entry in the documentation about this new option

Differential Revision: D80227719

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -676,6 +676,12 @@ Each query can override the config by setting corresponding query session proper
      - bool
      - true
      - Reads timestamp partition value as local time if true. Otherwise, reads as UTC.
+   * - hive.preserve-flat-maps-in-memory
+     - hive.preserve_flat_maps_in_memory
+     - bool
+     - false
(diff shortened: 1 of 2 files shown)


velox/docs/configs.rst
```diff
@@ -676,6 +676,11 @@ Each query can override the config by setting corresponding query session proper
      - bool
      - true
      - Reads timestamp partition value as local time if true. Otherwise, reads as UTC.
+   * - hive.preserve-flat-maps-in-memory
+     - hive.preserve_flat_maps_in_memory
+     - bool
+     - false
+     - Whether to preserve flat maps in memory as FlatMapVectors instead of converting them to MapVectors.
```
pedroerp: let's clarify that this is only applied inside the DWRF and Nimble readers - not during during expression eval or anywhere else.

pedroerp (approved):
Thank you

PR #14477, opened 2025-08-14
refactor: Remove unused methods

MBkkt:
They're also unnecessary increase binary size and doesn't allow to update xsimd (https://github.com/xtensor-stack/xsimd/commit/ead311fda435d7401c2ad5b34985a032515a945b but "batch_bool instead" doesn't compile)

Maybe I missed something why these members needed, but from my experience compiler auto-vectorization works quite good (same performance numbers for vector similarity functions as manually optimized functions from well known libraries).
The only reason when manual intrinsic make sense is simd-specific algorithms like some bitpacking/utf8/etc algorithms or if whole code compiles with different architecture options (e.g. without avx512 or avx2 and you check in runtime that they're available)
But there's not a lot of sense in writing common loops with arithmetic operations and using simd "registers".

Also 
* I fixed some clang-tidy warnings in related files.
* I added `.h` in `-inl.h` because otherwise clangd (code completion) doesn't work

### DIFF
--- a/velox/docs/programming-guide/chapter02.rst
+++ b/velox/docs/programming-guide/chapter02.rst
@@ -327,7 +327,8 @@ We can now populate the strings using the setNoCopy method.
 .. code-block:: c++
 
     for (auto i = 0; i < 100; ++i) {
-     substr->setNoCopy(i, StringView(vector->valueAt(i).data(), 20));
+     const auto value = vector->valueAt(i);
+     substr->setNoCopy(i, StringView(value.data(), 20));
     }
 
     LOG(INFO) << substr->toString();
--- a/velox/dwio/dwrf/test/TestColumnReader.cpp
+++ b/velox/dwio/dwrf/test/TestColumnReader.cpp
@@ -1290,10 +1290,10 @@ TEST_P(StringReaderTests, testDictionaryWithNulls) {
     } else {
       EXPECT_FALSE(stringBatch->isNullAt(i));
       const char* expected = i < 98 ? "ORC" : "Owen";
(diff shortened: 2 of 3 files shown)


velox/vector/BiasVector.h
```diff
@@ -110,31 +106,17 @@ class BiasVector : public SimpleVector<T> {
...
-
-  /**
-   * Loads a SIMD vector of data at the virtual byteOffset given
-   * Note this method is implemented on each vector type, but is intentionally
-   * not virtual for performance reasons
-   *
-   * @param byteOffset - the byte offset to laod from
-   */
-  xsimd::batch<T> loadSIMDValueBufferAt(size_t index) const;
+  T valueAt(vector_size_t idx) const final;
```
mbasmanova, line 115: I believe valueAtFast is used in some legacy application. Let me check.

mbasmanova:
Seeing errors:

```
fbcode/velox/functions/prestosql/types/TimestampWithTimeZoneRegistration.cpp:86:34: error: call to deleted member function 'data'
   86 |         inputVector.valueAt(row).data(),
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

MBkkt:
@mbasmanova Yeah, this is funny issue, I think I fixed it.
It was because returned type was `const T` and it was allowed to call `const&` method, because `&` vs `&&` resolved before `const`. https://godbolt.org/z/3vj8hMj1n

mbasmanova:
@MBkkt I confirmed that valueAtFast is used internally. 176 buck targets fail.

MBkkt:
@mbasmanova Ok I will keep it

PR #14482, opened 2025-08-14
feat: Add enum_key Presto function

HeidiHan0000:
Summary: Add enum_key function using the SimpleFunctionAPI.

Differential Revision: D80284921

### DIFF
--- a/velox/functions/prestosql/EnumFunctions.h
+++ b/velox/functions/prestosql/EnumFunctions.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/common/base/Status.h"
+#include "velox/core/QueryConfig.h"
+#include "velox/functions/Macros.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+
+namespace facebook::velox::functions {
+
+template <typename TExec>
+struct EnumKeyFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& inputTypes,
+      const core::QueryConfig& /*config*/,
+      const arg_type<BigintEnum<E1>>* /*input*/) {
+    VELOX_USER_CHECK_EQ(
+        inputTypes.size(), 1, "Expected 1 input type for enum_key function.");
+    enumPtr_ = std::dynamic_pointer_cast<const BigintEnumType>(inputTypes[0]);
+    VELOX_USER_CHECK_NOT_NULL(
+        enumPtr_, "Input type for enum_key function must be a BigintEnumType.");
+  }
+
+  Status call(out_type<Varchar>& result, const int64_t& input) {
+    auto keyAt = enumPtr_->keyAt(input);
+    if (!keyAt.has_value()) {
+      return Status::UserError("Value '{}' not in enum 'BigintEnum'", input);
+    }
+    result = keyAt.value();
+    return Status::OK();
+  }
+
+ private:
+  BigintEnumTypePtr enumPtr_;
+};
+
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/registration/EnumFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/EnumFunctionsRegistration.cpp
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/Registerer.h"
+#include "velox/functions/prestosql/EnumFunctions.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+#include "velox/type/SimpleFunctionApi.h"
+
+namespace facebook::velox::functions {
+
+void registerEnumFunctions(const std::string& prefix) {
+  registerBigintEnumType();
+
+  registerFunction<EnumKeyFunction, Varchar, BigintEnum<E1>>(
+      {prefix + "enum_key"});
+}
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/tests/EnumFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/EnumFunctionsTest.cpp
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
+
+using namespace facebook::velox::functions::test;
+using namespace facebook::velox::test;
+
+namespace facebook::velox::functions {
+namespace {
+
+class EnumFunctionsTest : public FunctionBaseTest {
+ protected:
+  void SetUp() override {
+    FunctionBaseTest::SetUp();
+  }
+};
+
+TEST_F(EnumFunctionsTest, enumKey) {
+  LongEnumParameter moodInfo("test.enum.mood", {{"CURIOUS", -2}, {"HAPPY", 0}});
+  auto bigintEnum = BIGINT_ENUM(moodInfo);
+  const auto enumKey = [&](std::optional<int64_t> val, TypePtr inputType) {
+    return evaluateOnce<std::string>("enum_key(c0)", inputType, val);
+  };
+  EXPECT_EQ("CURIOUS", enumKey(-2, bigintEnum));
+  EXPECT_EQ(std::nullopt, enumKey(std::nullopt, bigintEnum));
+
+  VELOX_ASSERT_THROW(
+      enumKey(1, bigintEnum), "Value '1' not in enum 'BigintEnum'");
+
+  VELOX_ASSERT_THROW(
+      enumKey(1, BIGINT()),
+      "Scalar function signature is not supported: enum_key(BIGINT). Supported signatures: (bigint_enum(E1)) -> varchar.");
+
+  VELOX_ASSERT_THROW(
+      evaluateOnce<std::string>(
+          "enum_key(c0, c1)",
+          {bigintEnum, BIGINT()},
+          std::optional<int64_t>(0),
+          std::optional<int64_t>(5)),
+      "Scalar function signature is not supported: enum_key(test.enum.mood:BigintEnum({\"CURIOUS\": -2, \"HAPPY\": 0}), BIGINT). Supported signatures: (bigint_enum(E1)) -> varchar.");
+}
+} // namespace
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/types/BigintEnumType.cpp
+++ b/velox/functions/prestosql/types/BigintEnumType.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <folly/Synchronized.h>
+#include <folly/container/EvictingCacheMap.h>
 
 #include "velox/functions/prestosql/types/BigintEnumType.h"
 
@@ -62,6 +63,14 @@ BigintEnumType::BigintEnumType(const LongEnumParameter& parameters)
 std::string BigintEnumType::toString() const {
   return fmt::format(
       "{}:BigintEnum({})", name_, flippedMapToString(flippedMap_));
+}
+
+const std::optional<std::string> BigintEnumType::keyAt(int64_t value) const {
+  auto it = flippedMap_.find(value);
+  if (it != flippedMap_.end()) {
+    return it->second;
+  }
+  return std::nullopt;
 }
 
 // A thread-safe LRU cache to store instances of BigintEnumType.
--- a/velox/functions/prestosql/types/BigintEnumType.h
+++ b/velox/functions/prestosql/types/BigintEnumType.h
@@ -15,8 +15,6 @@
  */
 #pragma once
 
-#include <folly/container/EvictingCacheMap.h>
-
 #include "velox/type/Type.h"
 
 namespace facebook::velox {
@@ -50,6 +48,10 @@ class BigintEnumType : public BigintType {
   bool containsValue(int64_t value) const {
     return flippedMap_.contains(value);
   }
+
+  /// Returns the string key given an integer value. If the value does not exist
+  /// in the flippedMap_, return std::nullopt.
+  const std::optional<std::string> keyAt(int64_t value) const;
 
   const std::string& enumName() const {
     return name_;
--- a/velox/core/SimpleFunctionMetadata.h
+++ b/velox/core/SimpleFunctionMetadata.h
@@ -288,6 +288,19 @@ struct TypeAnalysis<LongDecimal<P, S>> {
   }
 };
 
+template <typename E>
+struct TypeAnalysis<facebook::velox::BigintEnumT<E>> {
+  void run(TypeAnalysisResults& results) {
+    results.stats.concreteCount++;
+
+    const auto e = E::name();
+    results.out << fmt::format("bigint_enum({})", e);
+    results.addVariable(exec::SignatureVariable(
+        e, std::nullopt, exec::ParameterType::kEnumParameter));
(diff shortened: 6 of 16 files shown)


mbasmanova (commented):
@HeidiHan0000 Would you add documentation for this function? 

Please, update the PR title to indicate that this is a Presto function. Remove "in Velox" as it is redundant. Do not repeat "function".

feat: Add enum_key Presto function

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+
+template <typename TExec>
+struct EnumKeyFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& inputTypes,
+      const core::QueryConfig& /*config*/,
+      const arg_type<BigintEnum<E1>>* input) {
+    VELOX_CHECK(
```
mbasmanova: VELOX_USER_CHECK_EQ

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+        inputTypes.size() == 1, "Expected 1 input type for enum_key function.");
+    const auto& inputEnum = static_cast<const BigintEnumType&>(*inputTypes[0]);
```
mbasmanova: dynamic_cast + check

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+    enumMap_ = inputEnum.flippedLongEnumMap();
+  }
+
+  void call(out_type<Varchar>& result, const arg_type<BigintEnum<E1>>& input) {
```
mbasmanova: arg_type<BigintEnum<E1>> -> int64_t

velox/functions/prestosql/registration/EnumFunctionsRegistration.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "velox/functions/prestosql/EnumFunctions.h"
+#include "velox/functions/prestosql/types/BigintEnumRegistration.h"
+#include "velox/functions/prestosql/types/BigintEnumType.h"
+#include "velox/type/SimpleFunctionApi.h"
+
+namespace facebook::velox::functions {
+
+void registerEnumFunctions(const std::string& prefix) {
+  registerBigintEnumType();
+  std::vector<exec::SignatureVariable> constraints = {
```
mbasmanova: Why do we need this? Is E1 not enough?

velox/functions/prestosql/tests/EnumFunctionsTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+using namespace facebook::velox::test;
+
+namespace facebook::velox::functions {
+namespace {
+
+class EnumFunctionsTest : public FunctionBaseTest {
+ protected:
+  void SetUp() override {
+    FunctionBaseTest::SetUp();
+    registerBigintEnumType();
```
mbasmanova: Is this needed?

velox/functions/prestosql/tests/EnumFunctionsTest.cpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+
+TEST_F(EnumFunctionsTest, enum_key) {
```
mbasmanova: naming: enum_key -> enumKey or basic

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -0,0 +1,44 @@
...
+    result = BigintEnumType::keyAt(input, enumMap_);
+  }
+
+ private:
+  std::unordered_map<int64_t, std::string> enumMap_;
```
mbasmanova: Can we store BigintEnumTypePtr itself and make keyAt an instance method? This would make the code simpler and avoid copying the map.

velox/functions/prestosql/tests/EnumFunctionsTest.cpp
HeidiHan0000: No, the SetUp takes care of this. Removed, thanks!

velox/functions/prestosql/registration/EnumFunctionsRegistration.cpp
HeidiHan0000: I found that this is required for it to be able to recognize the named SignatureVariable, like the E1 in "(bigint_enum(E1)) -> varchar". (I've changed it to use "E1" instead of "enumParameters" to make that more obvious.) Actually, would you be able to advise how I can use the FunctionSignatureBuilder with the registerFunction api?

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -0,0 +1,98 @@
...
+FOLLY_ALWAYS_INLINE BigintEnumTypePtr asBigintEnum(const TypePtr& type) {
+  return std::dynamic_pointer_cast<const BigintEnumType>(type);
+}
+
+// Type to use for inputs and outputs of simple functions, e.g.
+// arg_type<BigintEnum> and out_type<BigintEnum>.
+template <typename P>
+struct BigintEnumT {
+  using type = int64_t;
+  static constexpr const char* typeName = "bigint_enum(E1)";
```
mbasmanova: E1 doesn't seem right here. Should it be something like P::name() ?

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -66,6 +67,14 @@ std::string mapToString(
   return orderedMapToString(sortedMap);
 }
 
+const std::string& BigintEnumType::keyAt(int64_t value) const {
+  auto it = flippedMap_.find(value);
+  if (it != flippedMap_.end()) {
+    return it->second;
+  }
+  VELOX_USER_FAIL("Value \'{}\' not in enum 'BigintEnum'", value);
```
mbasmanova: no need to escape single quotes

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -0,0 +1,45 @@
...
+      const arg_type<BigintEnum<E1>>* /*input*/) {
+    VELOX_USER_CHECK_EQ(
+        inputTypes.size(), 1, "Expected 1 input type for enum_key function.");
+    enumPtr_ = std::dynamic_pointer_cast<const BigintEnumType>(inputTypes[0]);
+    VELOX_USER_CHECK_NOT_NULL(
+        enumPtr_, "Input type for enum_key function must be a BigintEnumType.");
+  }
+
+  void call(out_type<Varchar>& result, const int64_t& input) {
+    result = enumPtr_->keyAt(input);
```
mbasmanova: Can we make this function non-throwing? 

https://velox-lib.io/blog/optimize-try-more/#non-throwing-simple-functions

velox/docs/functions/presto/coverage.rst
```diff
@@ -353,7 +353,7 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with
...
     :func:`array_intersect`                   :func:`empty_approx_set`                  :func:`json_format`                       :func:`replace_first`                     st_within                                     :func:`classification_miss_rate`
     :func:`array_join`                        :func:`ends_with`                         :func:`json_parse`                        :func:`reverse`                           st_x                                          :func:`classification_precision`
-    array_least_frequent                      enum_key                                  :func:`json_size`                         rgb                                       st_xmax                                       :func:`classification_recall`
+    array_least_frequent                      :func:`enum_key`                                    :func:`json_size`                         rgb                                       st_xmax                                       :func:`classification_recall`
```
mbasmanova: Coverage should be updated in a separate PR to avoid merge conflicts.

velox/functions/prestosql/tests/EnumFunctionsTest.cpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+TEST_F(EnumFunctionsTest, enumKey) {
+  LongEnumParameter moodInfo("test.enum.mood", {{"CURIOUS", -2}, {"HAPPY", 0}});
+  const std::vector<TypeParameter>& moodParams = {TypeParameter(moodInfo)};
+  auto bigintEnum = BIGINT_ENUM(moodParams);
+
+  auto expected =
+      makeNullableFlatVector<StringView>({"HAPPY", "CURIOUS", std::nullopt});
+  auto result = evaluate(
```
mbasmanova: Would you use evaluateOnce for readability?

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: For the SignatureVariable, I used E1::name(), but for the typeName string here, I just used the string. Should I format it with E1::name() here as well?

velox/docs/functions/presto/coverage.rst
HeidiHan0000: Removed the change. But could you clarify what it would conflict with?

velox/functions/prestosql/EnumFunctions.h
HeidiHan0000: There isn't a case where keyAt will throw, so I just added a return Status::OK(); at the end. Is there a non-throwing version of the initialize function as well, or is that one okay to throw?

mbasmanova: > There isn't a case where keyAt will throw
 
 The implementation of keyAt suggest that it throws on invalid input. Is this not the case?

```
const std::string& BigintEnumType::keyAt(int64_t value) const {
  auto it = flippedMap_.find(value);
  if (it != flippedMap_.end()) {
    return it->second;
  }
  VELOX_USER_FAIL("Value '{}' not in enum 'BigintEnum'", value);
}
```

velox/functions/prestosql/types/BigintEnumType.h
mbasmanova: I assume we want different names for `BigintEnum<E1>` and `BigintEnum<E2>`. Currently, both  produce the same name: bigint_enum(E1)

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,143 @@
...
+        castFromInteger<int32_t>(input, context, rows, resultType, result);
+        break;
+      case TypeKind::BIGINT:
+        castFromInteger<int64_t>(input, context, rows, resultType, result);
+        break;
+      default:
+        VELOX_UNREACHABLE(
+            "Cannot cast {} to {} type",
+            input.type()->toString(),
+            asBigintEnum(result->type())->toString());
```
mbasmanova: Do we need asBigintEnum here ? Can we not just write result->type()->toString()?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,143 @@
...
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
+
+    const auto* enumInts = input.as<SimpleVector<int64_t>>();
+    rows.applyToSelected([&](vector_size_t row) {
```
mbasmanova: Can we replace this loop with

```
flatResult->copy(rows, enumInts, nullptr);
```

?

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,112 @@
...
+  std::unordered_map<int64_t, std::string> flippedMap;
+  for (const auto& [key, value] : map) {
+    bool ok = flippedMap.emplace(value, key).second;
+    VELOX_USER_CHECK(
+        ok, "Invalid enum type {}, contains duplicate value {}", name, value);
+  }
+  return flippedMap;
+}
+
+std::string orderedMapToString(
```
mbasmanova: consistency: orderedMapToString -> sortedMapToString (to match sortedMap)

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+std::string BigintEnumType::toString() const {
+  std::map<std::string, int64_t> sortedMap;
+  for (const auto& [key, value] : flippedMap_) {
+    sortedMap[value] = key;
+  }
+  return fmt::format("{}:BigintEnum({})", name_, orderedMapToString(sortedMap));
+}
+
+// A thread-safe LRU cache to store instances of BigintEnumType.
```
mbasmanova, line 76: Would you update documentation for this type to explain the usage of this cache?

velox/type/Type.cpp
```diff
@@ -104,6 +104,30 @@ namespace {
...
+TypeParameter deserializeEnumParam(const folly::dynamic& obj) {
+  auto enumName = obj["enumName"].asString();
+  VELOX_CHECK(obj["valuesMap"].isObject());
+  auto valuesMap = obj["valuesMap"];
+
+  // Construct the values map
+  std::unordered_map<std::string, ValueType> map;
+  for (const auto& item : valuesMap.items()) {
+    std::string key = item.first.asString();
+    if constexpr (std::is_same_v<ValueType, int64_t>) {
```
mbasmanova: Should we add 'else' block to throw? Otherwise, we seem to be silently dropping data.

velox/docs/functions/presto/coverage.rst
mbasmanova: Functions are added often. When multiple PRs add functions and update coverage file, they produce conflicts.

velox/functions/prestosql/types/BigintEnumType.cpp
HeidiHan0000: I was able to remove this function altogether after changing the cache to use LongEnumParameter instead of a stringified value

velox/functions/prestosql/types/BigintEnumRegistration.cpp
HeidiHan0000: You are right, works with resultType->toString()

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: Since I only have the enum_key with "E1" registered, then can I remove E2, and just use like a common "E" to register it, then here just set typeName = "bigint_enum(E)"?

velox/functions/prestosql/EnumFunctions.h
HeidiHan0000: I must have looked at something else, sorry for the confusion. I've updated it to return Status::UserError for that case. It seems that the test still passes with VELOX_ASSERT_THROW for that UserError, is this correct?

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  void castFrom(
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const SelectivityVector& rows,
+      const TypePtr& resultType,
+      VectorPtr& result) const override {
+    context.ensureWritable(rows, resultType, result);
+    auto* flatResult = result->asChecked<FlatVector<int64_t>>();
+    flatResult->clearNulls(rows);
```
mbasmanova: This can be removed since you are calling 'copy' later.

velox/functions/prestosql/types/BigintEnumRegistration.cpp
```diff
@@ -0,0 +1,139 @@
...
+    const auto* enumInts = input.as<SimpleVector<int64_t>>();
```
mbasmanova: this cast is not needed. you can just call flatResult->copy(input, rows, nullptr).

velox/functions/prestosql/types/BigintEnumType.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+} // namespace
+
+// Should only be called from get() to create a new instance.
+BigintEnumType::BigintEnumType(const LongEnumParameter& parameters)
+    : parameters_{TypeParameter(parameters)},
+      name_{parameters.name},
+      flippedMap_{toFlippedMap(parameters.valuesMap, name_)} {}
+
+std::string BigintEnumType::flippedMapToString() const {
```
mbasmanova: Let's turn this into a free standing function that takes a map as input.

velox/type/Type.h
```diff
@@ -2407,3 +2424,26 @@ struct fmt::formatter<
     return formatter<string_view>::format(k->toString(), ctx);
   }
 };
+
+namespace std {
+template <>
+struct hash<facebook::velox::LongEnumParameter> {
+  size_t operator()(const ::facebook::velox::LongEnumParameter& param) const {
+    auto name = param.name;
```
mbasmanova: Move the implementation to .cpp file. 

It doesn't have to be std::hash. Can just be a custom hash function. You can specify it as template parameter when creating a cache.

mbasmanova (approved):
@HeidiHan0000 Thank you for iterating. Looks good. Please, consider remaining comments.

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -53,10 +53,16 @@ class BigintEnumType : public BigintType {
     return flippedMap_.contains(value);
   }
 
+  const std::optional<std::string> keyAt(int64_t value) const;
```
mbasmanova, line 54: Please, document all APIs.

velox/functions/prestosql/types/BigintEnumType.h
```diff
@@ -53,10 +53,16 @@ class BigintEnumType : public BigintType {
...
+
   const std::string& enumName() const {
     return name_;
   }
 
+  const std::unordered_map<int64_t, std::string>& flippedLongEnumMap() const {
```
mbasmanova: Is this API needed?

velox/functions/prestosql/EnumFunctions.h
mbasmanova: Yes, the test would pass regardless of whether you throw directly or set error status. The difference is in performance. Please, read the blog post I shared earlier.

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: Removed

velox/functions/prestosql/types/BigintEnumType.h
HeidiHan0000: Updated it so the typeName will use the E::name() in the string, and kept both E1 and E2 for possible future use

velox/functions/prestosql/registration/EnumFunctionsRegistration.cpp
HeidiHan0000: I was able to implement the `run` function for`TypeAnalysis<facebook::velox::BigintEnumT<E>>` , so that the SignatureVariable will be added through the `analysis` step using the template arguments of registerFunctions, rather than pass them in as constraints and have them added that way. Please let me know how this looks!

mbasmanova (approved):

velox/type/parser/ParserUtil.cpp
```diff
@@ -65,4 +65,57 @@ std::pair<std::string, std::shared_ptr<const Type>> inferTypeWithSpaces(
...
+  for (const auto& pair : obj) {
+    VELOX_CHECK(
+        pair.isArray() && pair.size() == 2 && pair[0].isString() &&
+            pair[1].isInt(),
+        "Failed to parse map: {}, each element must be a [string key, int value] pair");
+
+    std::string key = pair[0].asString();
+    int64_t value = pair[1].asInt();
+
+    VELOX_CHECK(
```
mbasmanova: nit: VELOX_USER_CHECK

mbasmanova (approved):

PR #14484, opened 2025-08-14
fix(build): Fix install_stemmer on Macs

kevinwilfong:
Summary:
Macs typically have BSD sed which requires a suffix with the -i option, Linux typically uses
GNU sed where it is optional.

Update the sed command in install_stemmer to work with either sed distribution.

Differential Revision: D80298887


kgpai (approved):
Please wait for all the images to pass/build.

pratikpugalia (approved):

czentgr (approved):

PR #14485, opened 2025-08-14
fix(fuzzer) Restrict range of Json float values to avoid scientific notation

pratikpugalia:
Summary: Restrict Json Input fuzzer float values to ```[10^-3, 10^7]``` to avoid conversion to scientific notation in presto SOT tests. ```json_parse``` in presto uses java's toString which converts any floats that do not belong to  the above mentioned range into the scientific notation and leads to mismatches in fuzzer tests.

Differential Revision: D80122608


peterenescu (approved):
Awesome, thanks for the fix!

PR #14487, opened 2025-08-15
feat: Add Task::finishIfAllSplitGroupsAreDone()

spershin:
Summary:
This methos is required in Native Presto for delayed Task start.
In this case we don't start the task and can process 'no more splits' message
before the task started.
If the task had no splits at all and is in grouped execution, then the task
will never transition to the kFinished state after it is started.
By adding this method we enable Native Presto to call it after starting the
delayed (queued) Task and transition to the kFinished state in the corner case
described above.

Differential Revision: D80314361


tanjialiang:
Should we add some tests for this method?

spershin:
> Should we add some tests for this method?

Let me come up with something.

velox/exec/tests/TaskTest.cpp
```diff
@@ -3273,4 +3272,71 @@ DEBUG_ONLY_TEST_F(TaskTest, taskExecutionEndTime) {
...
+
+  // Emulate 'no more splits message' before the task is started.
+  task->noMoreSplits(groupedScanNodeId);
+
+  // Task would not be in the kFinished state even though all splits are done.
+  EXPECT_EQ(task->state(), TaskState::kRunning);
+
+  task->start(1);
+
+  // No more splits will be there, the task has done everything.
```
tanjialiang, line 3315: maybe check state before the call as well.

spershin: > maybe check state before the call first.

@tanjialiang 
We do it on the line 3311.

tanjialiang:
lgtm cc @xiaoxmeng to see if he has some comments.

velox/exec/tests/TaskTest.cpp
spershin: Ah, you mean not before the start() call, but before the finishIfAllSplitGroupsAreDone()...
The code reference does not show this well.

Added the check.

spershin:
Abandoning in favor of the https://github.com/prestodb/presto/pull/25801

Issue #14413, opened 2025-08-11
Stack overflow when grouping by nano-timestamp key


zhztheplayer:
### Bug description

Infinite recursion at code:

```
facebook::velox::exec::HashTable::prepareForGroupProbe HashTable.cpp:2179
```

When running a SQL that groups by a nano-precision timestamp. Example:

```sql
SELECT DISTINCT * FROM (
    SELECT CURRENT_TIMESTAMP() as ts FROM range(0, 1000)
) t;
```

```
std::mutex::lock std_mutex.h:98
std::lock_guard::lock_guard std_mutex.h:229
facebook::velox::memory::MallocAllocator::incrementUsageWithReservationFunc MallocAllocator.h:139
operator() MallocAllocator.cpp:29
std::__invoke_impl<…>(std::__invoke_other, struct {...} &) invoke.h:61
std::__invoke_r<…>(struct {...} &) invoke.h:114
std::_Function_handler::_M_invoke(const std::_Any_data &, unsigned int &, unsigned int &&, std::mutex &) std_function.h:290
std::function::operator() std_function.h:590
facebook::velox::ConcurrentCounter::Counter::update ConcurrentCounter.h:103
facebook::velox::ConcurrentCounter::update ConcurrentCounter.h:70
facebook::velox::memory::MallocAllocator::incrementUsageWithReservation MallocAllocator.h:131
facebook::velox::memory::MallocAllocator::incrementUsage MallocAllocator.h:119
facebook::velox::memory::MallocAllocator::allocateBytesWithoutRetry MallocAllocator.cpp:266
operator() MemoryAllocator.cpp:342
std::__invoke_impl<…>(std::__invoke_other, struct {...} &) invoke.h:61
std::__invoke_r<…>(struct {...} &) invoke.h:114
std::_Function_handler::_M_invoke(const std::_Any_data &, facebook::velox::memory::Allocation &) std_function.h:290
std::function::operator() std_function.h:590
facebook::velox::cache::AsyncDataCache::makeSpace AsyncDataCache.cpp:785
facebook::velox::memory::MemoryAllocator::allocateBytes MemoryAllocator.cpp:339
facebook::velox::memory::MemoryPoolImpl::allocate MemoryPool.cpp:528
facebook::velox::raw_vector::allocateData RawVector.h:208
facebook::velox::raw_vector::grow RawVector.h:235
facebook::velox::raw_vector::reserve RawVector.h:149
[...]

PR #14488, opened 2025-08-15
test: Set timestamp precision to microseconds back in Spark aggregation fuzzer

zhztheplayer:
The timestamp precision in SparkAggregationFuzzerTest.cpp was set to millis temporarily in https://github.com/facebookincubator/velox/pull/13494.

Since a related issue https://github.com/facebookincubator/velox/issues/14413 is now fixed we should be able to set the precision back to micros.


zhztheplayer:
Thanks @rui-mo

PR #14490, opened 2025-08-15
fix: Use microsecond timestamp precision in Spark aggregation fuzzer test

rui-mo:
Spark aggregation fuzzer test starts to use millisecond timestamp precision 
after 1d20f63 to avoid a deep recursion caused by microsecond timestamp. 
Following the 2e4a083 fix to VectorHasher, microsecond precision can be 
utilized to be consistent with Spark.


rui-mo:
Duplicate with https://github.com/facebookincubator/velox/pull/14488.

PR #14493, opened 2025-08-15
refactor(expression): Move constants into separate header

czentgr:
Collect the constants used to identify special forms into a new common header. This is used later to refer to the same constants where applicable instead of literals.

### DIFF
--- a/velox/expression/ExprConstants.h
+++ b/velox/expression/ExprConstants.h
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+namespace facebook::velox::expression {
+
+constexpr const char* kConjunt = "conjunct";
+constexpr const char* kAnd = "and";
+constexpr const char* kOr = "or";
+constexpr const char* kSwitch = "switch";
+constexpr const char* kIn = "in";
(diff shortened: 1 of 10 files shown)


czentgr:
@mbasmanova For the expression PR that you reviewed github.com/facebookincubator/velox/pull/13424 we've decided to break it up into three separate PRs (this PR, https://github.com/facebookincubator/velox/pull/14523, and https://github.com/facebookincubator/velox/pull/14524). This is the first one moving the constants to a common header to be used.
If you have some time please take a look.

velox/expression/ExprConstants.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+namespace facebook::velox::expression {
```
mbasmanova, line 18: This works.

Have we decided how to split the code between velox::expression, velox::core and velox::exec? What would be the rules for deciding what goes where? 

It might be nicer to introduce a class, BuiltinNames or SpecialFormNames or similar. Or maybe, SpecialFormKind enum w/ VELOX_DEFINE_ENUM_NAME.

mbasmanova (approved):

velox/expression/ExprConstants.h
czentgr: That's a good question. I notice lots of these functions are in the "exec" namespace but in the "expression" folder. So presumably, there is a logical difference envision (even if not used for namespace purposes). 
For these constants though, there are names for expressions so it made sense to put them in that namespace as opposed to exec. I didn't consider core.

Come to look at this `SpecialFormKind` defintion. It uses the same name but in upper case. However, we could compare to kind and not name. Occasionally, the name is needed and stored in the SpecialForm itself.
We could use the names from `SpecialFormKind` which are upper case and store them instead. One missing enum would be the TryCast name which is stored in the CastExpr SpecialForm. but seems like we are already redefining the names multiple times which we are trying to avoid.

On the other hand, the strings used in the `SpecialFormKind` enum could use our new ExprConstants (and convert them to lower case).

czentgr:
@kevinwilfong looks like there might be some internal usage of these constants causing these failures? Anything I can help with?

PR #14496, opened 2025-08-18
fix: Allow release of window build resource on window output completion

tanjialiang:
Summary: Window operator currently holds resources allocated by row container till destructor is called. The resource held by row container could have been released as soon as all output has been produced. This change applies the above in time release logic.

Differential Revision: D80433738


aditi-pandit (approved):
Thanks @tanjialiang.

PR #14501, opened 2025-08-18
feat(cudf): Add an option to register functions in cudf expression evaluator

devavret:
Depends on #14462
Contributes to #14149


devavret:
closing because changes rolled up into #14462

PR #14507, opened 2025-08-18
feat: Add config to set table scan driver count

kewang1024:


velox/core/QueryConfig.h
```diff
@@ -396,6 +396,11 @@ class QueryConfig {
   static constexpr const char* kSparkJsonIgnoreNullFields =
       "spark.json_ignore_null_fields";
 
+  /// The number of table scan operators per task. When set, it will override
+  /// the max driver count limit.
+  static constexpr const char* kTaskTableScanDriverCount =
```
aditi-pandit, line 401: @kewang1024 : Thanks for this code.

Don't follow the reason for this config ? How does the engine determine what value to use here ?

kewang1024: @aditi-pandit The reasoning behind it is we want to scale table scan drivers up without bumping up max-driver-per-task on prestissimo

PR #14510, opened 2025-08-18
feat(python): Support for $row_group_id

pedroerp:
Summary:
Adding support for $row_group_id special column on PyVelox table scan
API. Also a small refactor to get rid of a memory leak issue if the plan is
destructed before all vectors generated by it.

Differential Revision: D80498363

### DIFF
--- a/velox/python/plan_builder/plan_builder.cpp
+++ b/velox/python/plan_builder/plan_builder.cpp
@@ -30,6 +30,9 @@ PYBIND11_MODULE(plan_builder, m) {
   velox::py::initializeVeloxMemory();
   velox::py::registerAllResources();
 
+  static auto rootPool = velox::memory::memoryManager()->addRootPool();
+  static auto leafPool = rootPool->addLeafChild("py_velox_plan_builder_pool");
+
   // Need types to specify table scan schema output.
   py::module::import("pyvelox.type");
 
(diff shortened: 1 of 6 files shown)


velox/python/plan_builder/plan_builder.cpp
```diff
@@ -30,6 +30,9 @@ PYBIND11_MODULE(plan_builder, m) {
   velox::py::initializeVeloxMemory();
   velox::py::registerAllResources();
 
+  static auto rootPool = velox::memory::memoryManager()->addRootPool();
+  static auto leafPool = rootPool->addLeafChild("py_velox_plan_builder_pool");
```
kgpai, line 34: Maybe the name  ought to be configurable and passed by user ?

velox/python/runner/PyLocalRunner.cpp
```diff
@@ -38,6 +38,18 @@ std::mutex& taskRegistryLock() {
   return lock;
 }
 
+exec::Split preparedSplit(
+    std::shared_ptr<connector::ConnectorSplit> connectorSplit) {
+  // Adds special $row_group_id column to the split.
+  if (auto* hiveConnectorSplit =
+          dynamic_cast<connector::hive::HiveConnectorSplit*>(
+              connectorSplit.get())) {
+    hiveConnectorSplit->infoColumns["$row_group_id"] =
```
kgpai, line 47: nit: curious why do we do this ? i.e what is it really used for ?

kgpai (approved):

PR #14516, opened 2025-08-19
fix: Remove redundant partition value handling in Iceberg column adaptation

PingLiuPing:
Implement IcebergSplitReader::adaptColumns. Without overriding this method, current iceberg reader uses `SplitReader::adaptColumns` which is specific to Hive implementation. One difference between Hive and Iceberg is Iceberg spec requires all columns should be wrote to data files while in Hive, it only write non-partition columns to data file. So, there are special logic to handle this during read in `SplitReader::adaptColumns`. But in Iceberg this is not needed.
And for Hive, it populates the column data with partition value directly, this is correct for Hive only, but for Iceberg, there are different kinds transforms other than Identity transform. And we can not deduce the original column from the partition value.

### DIFF
--- a/velox/connectors/hive/SplitReader.cpp
+++ b/velox/connectors/hive/SplitReader.cpp
@@ -28,30 +28,28 @@ namespace facebook::velox::connector::hive {
 namespace {
 
 template <TypeKind kind>
-VectorPtr newConstantFromString(
+VectorPtr newConstantFromStringImpl(
     const TypePtr& type,
     const std::optional<std::string>& value,
-    vector_size_t size,
     velox::memory::MemoryPool* pool,
-    const std::string& sessionTimezone,
-    bool asLocalTime,
-    bool isPartitionDateDaysSinceEpoch = false) {
+    bool isLocalTimestamp,
+    bool isDaysSinceEpoch) {
   using T = typename TypeTraits<kind>::NativeType;
   if (!value.has_value()) {
-    return std::make_shared<ConstantVector<T>>(pool, size, true, type, T());
+    return std::make_shared<ConstantVector<T>>(pool, 1, true, type, T());
   }
 
   if (type->isDate()) {
     int32_t days = 0;
     // For Iceberg, the date partition values are already in daysSinceEpoch
     // form.
-    if (isPartitionDateDaysSinceEpoch) {
+    if (isDaysSinceEpoch) {
       days = folly::to<int32_t>(value.value());
     } else {
       days = DATE()->toDays(value.value());
     }
     return std::make_shared<ConstantVector<int32_t>>(
-        pool, size, false, type, std::move(days));
+        pool, 1, false, type, std::move(days));
   }
 
   if constexpr (std::is_same_v<T, int64_t> || std::is_same_v<T, int128_t>) {
@@ -62,28 +60,44 @@ VectorPtr newConstantFromString(
           StringView(value.value()), precision, scale, result);
       VELOX_USER_CHECK(status.ok(), status.message());
       return std::make_shared<ConstantVector<T>>(
-          pool, size, false, type, std::move(result));
+          pool, 1, false, type, std::move(result));
     }
   }
 
   if constexpr (std::is_same_v<T, StringView>) {
     return std::make_shared<ConstantVector<StringView>>(
-        pool, size, false, type, StringView(value.value()));
+        pool, 1, false, type, StringView(value.value()));
   } else {
     auto copy = velox::util::Converter<kind>::tryCast(value.value())
                     .thenOrThrow(folly::identity, [&](const Status& status) {
                       VELOX_USER_FAIL("{}", status.message());
                     });
     if constexpr (kind == TypeKind::TIMESTAMP) {
-      if (asLocalTime) {
+      if (isLocalTimestamp) {
         copy.toGMT(Timestamp::defaultTimezone());
       }
     }
     return std::make_shared<ConstantVector<T>>(
-        pool, size, false, type, std::move(copy));
+        pool, 1, false, type, std::move(copy));
   }
 }
 } // namespace
+
+VectorPtr newConstantFromString(
+    const TypePtr& type,
+    const std::optional<std::string>& value,
+    velox::memory::MemoryPool* pool,
+    bool isLocalTimestamp,
+    bool isDaysSinceEpoch) {
+  return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH_ALL(
+      newConstantFromStringImpl,
+      type->kind(),
+      type,
+      value,
+      pool,
+      isLocalTimestamp,
+      isDaysSinceEpoch);
+}
 
 std::unique_ptr<SplitReader> SplitReader::create(
     const std::shared_ptr<hive::HiveConnectorSplit>& hiveSplit,
@@ -424,16 +438,13 @@ std::vector<TypePtr> SplitReader::adaptColumns(
                iter != hiveSplit_->infoColumns.end()) {
       auto infoColumnType =
           readerOutputType_->childAt(readerOutputType_->getChildIdx(fieldName));
-      auto constant = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH_ALL(
-          newConstantFromString,
-          infoColumnType->kind(),
+      auto constant = newConstantFromString(
           infoColumnType,
           iter->second,
-          1,
           connectorQueryCtx_->memoryPool(),
-          connectorQueryCtx_->sessionTimezone(),
           hiveConfig_->readTimestampPartitionValueAsLocalTime(
-              connectorQueryCtx_->sessionProperties()));
+              connectorQueryCtx_->sessionProperties()),
+          false);
       childSpec->setConstantValue(constant);
     } else if (
         childSpec->columnType() == common::ScanSpec::ColumnType::kRegular) {
@@ -480,14 +491,10 @@ void SplitReader::setPartitionValue(
       "ColumnHandle is missing for partition key {}",
       partitionKey);
   auto type = it->second->dataType();
-  auto constant = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH_ALL(
-      newConstantFromString,
-      type->kind(),
+  auto constant = newConstantFromString(
       type,
       value,
-      1,
       connectorQueryCtx_->memoryPool(),
-      connectorQueryCtx_->sessionTimezone(),
       hiveConfig_->readTimestampPartitionValueAsLocalTime(
           connectorQueryCtx_->sessionProperties()),
       it->second->isPartitionDateValueDaysSinceEpoch());
--- a/velox/connectors/hive/SplitReader.h
+++ b/velox/connectors/hive/SplitReader.h
@@ -47,6 +47,36 @@ class MemoryPool;
 
 namespace facebook::velox::connector::hive {
 
+/// Creates a constant vector of size 1 from a string representation of a value.
+///
+/// Used to materialize partition column values and info columns (e.g., $path,
+/// $file_size) when reading Hive and Iceberg tables. Partition values are
+/// stored as strings in HiveConnectorSplit::partitionKeys and need to be
+/// converted to their appropriate types.
+///
+/// @param type The target Velox type for the constant vector. Supports all
+/// scalar types including primitives, dates, timestamps, and decimals.
+/// @param value The string representation of the value to convert, formatted
+/// the same way as CAST(x as VARCHAR). Date values must be formatted using ISO
+/// 8601 as YYYY-MM-DD. If nullopt, creates a null constant vector.
+/// @param pool Memory pool for allocating the constant vector.
+/// @param isLocalTimestamp If true and type is TIMESTAMP, interprets the string
+/// value as local time and converts it to GMT. If false, treats the value
+/// as already in GMT.
+/// @param isDaysSinceEpoch If true and type is DATE, treats the string value as
+/// an integer representing days since epoch (used by Iceberg). If false, parses
+/// the string as a date string in ISO 8601 format (used by Hive).
+///
+/// @return A constant vector of size 1 containing the converted value, or a
+/// null constant if value is nullopt.
+/// @throws VeloxUserError if the string cannot be converted to the target type.
+VectorPtr newConstantFromString(
+    const TypePtr& type,
+    const std::optional<std::string>& value,
+    velox::memory::MemoryPool* pool,
+    bool isLocalTimestamp,
+    bool isDaysSinceEpoch);
+
 struct HiveConnectorSplit;
 class HiveTableHandle;
 class HiveColumnHandle;
@@ -163,17 +193,24 @@ class SplitReader {
       VectorPtr& output,
       const std::vector<BaseVector::CopyRange>& ranges);
 
- private:
-  /// Different table formats may have different meatadata columns.
-  /// This function will be used to update the scanSpec for these columns.
-  std::vector<TypePtr> adaptColumns(
-      const RowTypePtr& fileType,
-      const std::shared_ptr<const velox::RowType>& tableSchema) const;
-
+  /// Sets a constant partition value on the scanSpec for a partition column.
+  /// Converts the partition key string value to the appropriate type and sets
+  /// it as a constant value in the scanSpec, so the column will be filled
+  /// with this constant value.
+  ///
+  /// @param spec The scan spec to set the constant value on.
+  /// @param partitionKey The name of the partition column.
   void setPartitionValue(
       common::ScanSpec* spec,
       const std::string& partitionKey,
       const std::optional<std::string>& value) const;
+
+ private:
+  /// Different table formats may have different meatadata columns.
+  /// This function will be used to update the scanSpec for these columns.
+  virtual std::vector<TypePtr> adaptColumns(
+      const RowTypePtr& fileType,
+      const RowTypePtr& tableSchema) const;
 
  protected:
   std::shared_ptr<const HiveConnectorSplit> hiveSplit_;
--- a/velox/connectors/hive/iceberg/IcebergSplitReader.cpp
+++ b/velox/connectors/hive/iceberg/IcebergSplitReader.cpp
@@ -141,4 +141,56 @@ uint64_t IcebergSplitReader::next(uint64_t size, VectorPtr& output) {
   return rowsScanned;
 }
 
+std::vector<TypePtr> IcebergSplitReader::adaptColumns(
+    const RowTypePtr& fileType,
+    const RowTypePtr& tableSchema) const {
+  std::vector<TypePtr> columnTypes = fileType->children();
+  auto& childrenSpecs = scanSpec_->children();
+  // Iceberg table stores all column's data in data file.
+  for (const auto& childSpec : childrenSpecs) {
+    const std::string& fieldName = childSpec->fieldName();
+    if (auto iter = hiveSplit_->infoColumns.find(fieldName);
+        iter != hiveSplit_->infoColumns.end()) {
+      auto infoColumnType = readerOutputType_->findChild(fieldName);
+      auto constant = newConstantFromString(
+          infoColumnType,
+          iter->second,
+          connectorQueryCtx_->memoryPool(),
+          hiveConfig_->readTimestampPartitionValueAsLocalTime(
+              connectorQueryCtx_->sessionProperties()),
+          false);
+      childSpec->setConstantValue(constant);
+    } else {
+      auto fileTypeIdx = fileType->getChildIdxIfExists(fieldName);
+      auto outputTypeIdx = readerOutputType_->getChildIdxIfExists(fieldName);
+      if (outputTypeIdx.has_value() && fileTypeIdx.has_value()) {
+        childSpec->setConstantValue(nullptr);
+        auto& outputType = readerOutputType_->childAt(*outputTypeIdx);
+        columnTypes[*fileTypeIdx] = outputType;
+      } else if (!fileTypeIdx.has_value()) {
+        // Handle columns missing from the data file in two scenarios:
+        // 1. Schema evolution: Column was added after the data file was
+        // written and doesn't exist in older data files.
+        // 2. Partition columns: Hive migrated table. In Hive-written data
+        // files, partition column values are stored in partition metadata
+        // rather than in the data file itself, following Hive's partitioning
+        // convention.
+        if (auto it = hiveSplit_->partitionKeys.find(fieldName);
+            it != hiveSplit_->partitionKeys.end()) {
+          setPartitionValue(childSpec.get(), fieldName, it->second);
+        } else {
+          childSpec->setConstantValue(BaseVector::createNullConstant(
+              tableSchema->findChild(fieldName),
+              1,
+              connectorQueryCtx_->memoryPool()));
+        }
+      }
+    }
+  }
+
+  scanSpec_->resetCachedValues(false);
+
+  return columnTypes;
+}
+
 } // namespace facebook::velox::connector::hive::iceberg
--- a/velox/connectors/hive/iceberg/IcebergSplitReader.h
+++ b/velox/connectors/hive/iceberg/IcebergSplitReader.h
@@ -50,6 +50,49 @@ class IcebergSplitReader : public SplitReader {
   uint64_t next(uint64_t size, VectorPtr& output) override;
 
  private:
+  /// Adapts the data file schema to match the table schema expected by the
+  /// query.
+  ///
+  /// This method reconciles differences between the physical data file schema
+  /// and the logical table schema, handling various scenarios where columns may
+  /// be missing, added, or need special treatment.
+  ///
+  /// @param fileType The schema read from the data file's metadata. This
+  /// represents the actual columns physically present in the Parquet/ORC file.
+  /// @param tableSchema The logical schema defined in the catalog (e.g., from
+  /// DDL). This represents the current table schema that queries expect.
+  ///
+  /// @return A vector of column types adapted to match the query's
+  /// expectations, with appropriate type conversions and constant values set
+  /// for missing or special columns.
+  ///
+  /// The method handles the following scenarios for each column in the scan
+  /// spec:
+  ///
+  /// 1. Info columns (e.g., $path, $data_sequence_number, $deleted)
+  ///    These are virtual columns that provide metadata about the file itself.
+  ///    Values are read from hiveSplit_->infoColumns map and set as constant
+  ///    values in the scanSpec so they're materialized for all rows.
+  ///
+  /// 2. Regular columns present in File:
+  ///    Column exists in both fileType and readerOutputType. Type is adapted
+  ///    from fileType to match the expected output type, handling schema
+  ///    evolution where column types may have changed.
+  ///
+  /// 3. Columns missing from File:
+  ///    a) Partition columns (hive-migrated tables):
+  ///       Column is marked as partition key in hiveSplit_->partitionKeys.
+  ///       In Hive-written Iceberg tables, partition column values are stored
+  ///       in partition metadata, not in the data file itself. Value is read
+  ///       from partition metadata and set as a constant.
+  ///    b) Schema evolution (newly added columns):
+  ///       Column was added to the table schema after this data file was
+  ///       written. Set as NULL constant since the old file doesn't contain
+  ///       this column.
+  std::vector<TypePtr> adaptColumns(
+      const RowTypePtr& fileType,
+      const RowTypePtr& tableSchema) const override;
+
   // The read offset to the beginning of the split in number of rows for the
   // current batch for the base data file
   uint64_t baseReadOffset_;
--- a/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
+++ b/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
@@ -28,6 +28,7 @@
 #include "velox/dwio/parquet/RegisterParquetReader.h"
 #endif
 
+#include <exec/tests/utils/AssertQueryBuilder.h>
 #include <folly/Singleton.h>
 
 using namespace facebook::velox::exec::test;
@@ -284,6 +285,30 @@ class HiveIcebergTest : public HiveConnectorTestBase {
     return splits;
   }
 
+  ColumnHandleMap makeColumnHandles(
+      const RowTypePtr& rowType,
+      const std::unordered_set<int>& partitionIndices = {}) {
+    ColumnHandleMap assignments;
+    for (auto i = 0; i < rowType->size(); ++i) {
+      const auto& columnName = rowType->nameOf(i);
+      const auto& columnType = rowType->childAt(i);
+      auto columnHandleType = partitionIndices.contains(i)
+          ? HiveColumnHandle::ColumnType::kPartitionKey
+          : HiveColumnHandle::ColumnType::kRegular;
+
+      assignments.insert(
+          {columnName,
+           std::make_shared<HiveColumnHandle>(
+               columnName,
+               columnHandleType,
+               columnType,
+               columnType,
+               std::vector<common::Subfield>{})});
+    }
+
+    return assignments;
+  }
+
 #ifdef VELOX_ENABLE_PARQUET
   std::vector<std::shared_ptr<ConnectorSplit>> createParquetDeleteFileAndSplits(
       const std::string& path,
@@ -299,9 +324,7 @@ class HiveIcebergTest : public HiveConnectorTestBase {
                     static_cast<vector_size_t>(deletedPositionSize),
                     [&](vector_size_t) { return path; }),
                 makeFlatVector<int64_t>(deletePositionsVec),
-            })},
-        config_,
-        flushPolicyFactory_);
+            })});
 
     IcebergDeleteFile icebergDeleteFile(
         FileContent::kPositionalDeletes,
@@ -348,11 +371,7 @@ class HiveIcebergTest : public HiveConnectorTestBase {
       // files. This is to make constructing DuckDB queries easier
       std::vector<RowVectorPtr> dataVectors =
           makeVectors(dataFile.second, startingValue);
-      writeToFile(
-          dataFilePaths[dataFile.first]->getPath(),
-          dataVectors,
-          config_,
-          flushPolicyFactory_);
+      writeToFile(dataFilePaths[dataFile.first]->getPath(), dataVectors);
 
       for (int i = 0; i < dataVectors.size(); i++) {
         dataVectorsJoined.push_back(dataVectors[i]);
@@ -409,11 +428,7 @@ class HiveIcebergTest : public HiveConnectorTestBase {
         totalPositionsInDeleteFile += positionsInRowGroup.size();
       }
 
-      writeToFile(
-          deleteFilePath->getPath(),
-          deleteFileVectors,
-          config_,
-          flushPolicyFactory_);
+      writeToFile(deleteFilePath->getPath(), deleteFileVectors);
 
       deleteFilePaths[deleteFileName] =
           std::make_pair(totalPositionsInDeleteFile, deleteFilePath);
@@ -554,7 +569,7 @@ class HiveIcebergTest : public HiveConnectorTestBase {
   }
 
   core::PlanNodePtr tableScanNode() {
-    return PlanBuilder(pool_.get()).tableScan(rowType_).planNode();
+    return PlanBuilder().tableScan(rowType_).planNode();
   }
 
   dwio::common::FileFormat fileFomat_{dwio::common::FileFormat::DWRF};
@@ -591,7 +606,7 @@ TEST_F(HiveIcebergTest, singleBaseFileSinglePositionalDeleteFile) {
 /// delete positions. The parameter passed to
 /// assertSingleBaseFileSingleDeleteFile is the delete positions.for the middle
 /// base file.
-TEST_F(HiveIcebergTest, MultipleBaseFilesSinglePositionalDeleteFile) {
+TEST_F(HiveIcebergTest, multipleBaseFilesSinglePositionalDeleteFile) {
   folly::SingletonVault::singleton()->registrationComplete();
 
   assertMultipleBaseFileSingleDeleteFile({0, 1, 2, 3});
@@ -754,7 +769,7 @@ TEST_F(HiveIcebergTest, positionalDeletesMultipleSplits) {
   assertMultipleSplits({1000, 9000, 20000}, 1, 0, 20000, 3);
 }
 
-TEST_F(HiveIcebergTest, testPartitionedRead) {
+TEST_F(HiveIcebergTest, partitionedRead) {
   RowTypePtr rowType{ROW({"c0", "ds"}, {BIGINT(), DateType::get()})};
   std::unordered_map<std::string, std::optional<std::string>> partitionKeys;
   // Iceberg API sets partition values for dates to daysSinceEpoch, so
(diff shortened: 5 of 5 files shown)


PingLiuPing:
@majetideepak  Can you help to review this PR, thank you very much.

PingLiuPing:
@zhli1142015 I think PR https://github.com/facebookincubator/velox/pull/12910 (a dependency of this PR) will be merged soon.  
Could you please review this PR as well? It fixes some scenarios when reading Iceberg tables.  
I have follow-up PRs planned to address issues with querying Iceberg tables that contains evolution schema.
Thanks.

PingLiuPing:
@mbasmanova 
Per our discussion in [15035](https://github.com/facebookincubator/velox/pull/15035), I plan to integrate the actual partition transform logic into that PR. To support this, I need to deliver this PR first to run the unit test.

The purpose of this PR is to address certain limitations in the Iceberg reader. When a partition column is defined in an Iceberg table, and when query such iceberg table we cannot reuse the Hive reader. This is because the Hive writer only writes non-partition columns to data files, and the partition column values are derived from the partition map during query phase (there is special logic in hive reader).

In contrast, Iceberg requires all columns (both partition and non-partition) to be written to data files, and hence during reading we do not need those special logic. And for most Iceberg partition transforms, it’s not possible to reconstruct the original column value from the partition value.

velox/connectors/hive/SplitReader.h
```diff
@@ -163,18 +163,18 @@ class SplitReader {
       VectorPtr& output,
       const std::vector<BaseVector::CopyRange>& ranges);
 
+  void setPartitionValue(
```
mbasmanova, line 203: Would you document this method?

velox/connectors/hive/SplitReader.h
```diff
@@ -163,18 +163,18 @@ class SplitReader {
...
+      const std::string& partitionKey,
+      const std::optional<std::string>& value) const;
+
  private:
   /// Different table formats may have different meatadata columns.
   /// This function will be used to update the scanSpec for these columns.
-  std::vector<TypePtr> adaptColumns(
+  virtual std::vector<TypePtr> adaptColumns(
       const RowTypePtr& fileType,
       const std::shared_ptr<const velox::RowType>& tableSchema) const;
```
mbasmanova: nit: RowTypePtr

velox/connectors/hive/iceberg/IcebergSplitReader.h
```diff
@@ -50,6 +50,10 @@ class IcebergSplitReader : public SplitReader {
   uint64_t next(uint64_t size, VectorPtr& output) override;
 
  private:
+  std::vector<TypePtr> adaptColumns(
```
mbasmanova, line 92: Would you document this method?

velox/connectors/hive/iceberg/IcebergSplitReader.h
```diff
@@ -50,6 +50,10 @@ class IcebergSplitReader : public SplitReader {
...
+      const RowTypePtr& fileType,
+      const std::shared_ptr<const velox::RowType>& tableSchema) const override;
```
mbasmanova: RowTypePtr

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -141,4 +141,43 @@ uint64_t IcebergSplitReader::next(uint64_t size, VectorPtr& output) {
...
 }
 
+std::vector<TypePtr> IcebergSplitReader::adaptColumns(
+    const RowTypePtr& fileType,
+    const std::shared_ptr<const velox::RowType>& tableSchema) const {
+  std::vector<TypePtr> columnTypes = fileType->children();
+  auto& childrenSpecs = scanSpec_->children();
+  // Iceberg table stores all column's data in data file.
+  for (auto i = 0; i < childrenSpecs.size(); ++i) {
+    auto* childSpec = childrenSpecs[i].get();
```
mbasmanova: Can this be a for-each loop?

for (const auto& child : childrenSpecs)

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -834,6 +834,323 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
   HiveConnectorTestBase::assertQuery(plan, splits, "SELECT 0, '2018-04-06'");
 }
 
+TEST_F(HiveIcebergTest, testReadDecimal) {
```
mbasmanova: drop 'test' prefix from test method names (please, update pre-existing method names in this file)

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -834,6 +834,323 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  RowTypePtr rowType{ROW({"c0", "price"}, {BIGINT(), DECIMAL(10, 2)})};
```
mbasmanova: auto rowType = ROW(...)

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -834,6 +834,323 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+        makeIcebergSplits(dataFilePath->getPath(), {}, partitionKeys);
+    splits.insert(splits.end(), icebergSplits.begin(), icebergSplits.end());
+    tempFiles.push_back(dataFilePath);
+
+    VectorPtr expectedPrice =
+        makeFlatVector<int64_t>({12345, 12345, 12345}, DECIMAL(10, 2));
+    if (i == 1) {
+      expectedPrice =
+          makeFlatVector<int64_t>({67890, 67890, 67890}, DECIMAL(10, 2));
+    }
```
mbasmanova: this code is confusing; consider rewriting

```
std::vector<int64_t> unscaledPartitionValues = {12345, 67890};
...
auto expectedPrice =
        makeFlatVector<int64_t>(3, [&](auto) {return unscaledPartitionValues[i];}, DECIMAL(10, 2));
```

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -834,6 +834,323 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  }
+
+  ColumnHandleMap assignments;
+  assignments.insert(
+      {"c0",
+       std::make_shared<HiveColumnHandle>(
+           "c0",
+           HiveColumnHandle::ColumnType::kRegular,
+           rowType->childAt(0),
+           rowType->childAt(0))});
```
mbasmanova: Add helper method to reduce copy-paste. Updating existing code in this file as well.

```
makeColumnHandle("c0", rowType->childAt(0));
makePartitionKeyHandle("price", rowType->childAt(1));
```

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -834,6 +834,323 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+
+  auto plan = PlanBuilder(pool_.get())
+                  .tableScan(rowType, {}, "", rowType, assignments)
+                  .planNode();
+
+  createDuckDbTable(allDataVectors);
+  assertQuery(plan, splits, "SELECT * FROM tmp", 0);
+}
+
+TEST_F(HiveIcebergTest, testAddNewColumn) {
```
mbasmanova: These tests a quite repetitive. Consider refactoring to reduce boiler plate and copy-paste.

jinchengchenghh:
Which error do you meet? Does it only occur in decimal column? Why?

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -141,4 +141,43 @@ uint64_t IcebergSplitReader::next(uint64_t size, VectorPtr& output) {
   return rowsScanned;
 }
 
+std::vector<TypePtr> IcebergSplitReader::adaptColumns(
+    const RowTypePtr& fileType,
+    const std::shared_ptr<const velox::RowType>& tableSchema) const {
+  std::vector<TypePtr> columnTypes = fileType->children();
```
jinchengchenghh, line 147: Please respect infoColumns, iceberg may also have infoColumns such as _delete.

jinchengchenghh:
I don't see error in Gluten unit test, do you have a unit test in presto to reproduce it?

PingLiuPing:
> I don't see error in Gluten unit test, do you have a unit test in presto to reproduce it?

@jinchengchenghh  With or without https://github.com/IBM/velox/pull/425?

PingLiuPing:
> Which error do you meet? Does it only occur in decimal column? Why?

Thanks, I think I need to change the PR description little bit. When I open this PR, https://github.com/facebookincubator/velox/pull/12910 has not been merged. And at that time when reading a decimal partition column there are errors.

But the second point in PR description stands. See https://github.com/facebookincubator/velox/pull/14516#issuecomment-3417192091.

jinchengchenghh:
In gluten, we test two mode, With or without https://github.com/IBM/velox/pull/425, both of them can pass all the tests.

PingLiuPing:
> In gluten, we test two mode, With or without [IBM#425](https://github.com/IBM/velox/pull/425), both of them can pass all the tests.

Ok, without [IBM#425](https://github.com/IBM/velox/pull/425), do you have a test case that querying a partitioned iceberg table?
The main purpose of this PR is overwrite adapColumns  so that avoid to using the constant partition value to construct the result.
Imagine there is a bucket transform and the bucket value is 4, there is no way to re-construct the original column value from 4.

jinchengchenghh:
Yes, this is a feature added long long ago, our customers also use it.

PingLiuPing:
> Yes, this is a feature added long long ago, our customers also use it.

Ok, sounds like Gluten has special logic to build 
```
  const std::unordered_map<std::string, std::optional<std::string>>
      partitionKeys;
``` 
and it can assure that the partition value is been reversed from the transformed value, e.g. from `4` to a `string`.

Can you help to find out this logic from Gluten? 

From presto I confirm that there is no such logic because the iceberg metadata only records the partition value and there is no way construct the original column value from the partition value. And the original column value is been wrote to data file and to get that we have read the data file.

jinchengchenghh:
Could you paste the error stack in presto? If presto really has this problem, we can merge it.

PingLiuPing:
@jinchengchenghh  Thanks for the comments.

When querying a partitioned table from Presto, it returns the correct data.

The reason is that the Presto coordinator node sends the `partitionKeys` information, and it does not match the logic here. The key in the `partitionKeys` map is always suffixed with the transform name, such as year, month, etc., while the fieldName is the original column name.

Because of this, for non-identity transforms, the if branch will not be executed. However, for identity transforms, the key in `partitionKeys` is the original column name, which matches fieldName, so it will call setPartitionValue.

And for the Gluten test you mentioned earlier, I think it does not contains any test that cover the identity transform with Decimal column. Because without https://github.com/facebookincubator/velox/pull/12910, the reading definitely has error.

I don’t think this inconsistency is correct. First, for Iceberg, we should ignore partition values here since the original value of  partitioned column is already written into the data files. Second, I found that the existing unit test for reading partitioned Iceberg tables is incorrect. In this case (https://github.com/facebookincubator/velox/blob/430c3fa22e03f5ace949b1accc9009b39bb04ff2/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp#L757-L835), it explicitly builds partitionKeys and uses them to create a split, then explicitly builds the column handle by marking which column is a partition column. This logic looks like it was copied from Hive, but from an Iceberg perspective, it’s not necessary. Iceberg’s partition transform is an hidden layer, and when querying Iceberg data, we only get the transformed values from the manifest.

Third, without changing the current implementation, any new test related to reading partitioned Iceberg tables must follow this pattern to build the plan:

```
connector::ColumnHandleMap assignments;
  assignments.insert(
      {"c0",
       std::make_shared<HiveColumnHandle>(
           "c0",
           HiveColumnHandle::ColumnType::kRegular,
           rowType->childAt(0),
           rowType->childAt(0))});

  std::vector<common::Subfield> requiredSubFields;
  HiveColumnHandle::ColumnParseParameters columnParseParameters;
  columnParseParameters.partitionDateValueFormat =
      HiveColumnHandle::ColumnParseParameters::kDaysSinceEpoch;
  assignments.insert(
      {"ds",
       std::make_shared<HiveColumnHandle>(
           "ds",
           HiveColumnHandle::ColumnType::kPartitionKey,
           rowType->childAt(1),
           rowType->childAt(1),
           std::move(requiredSubFields),
           columnParseParameters)});

  auto plan = PlanBuilder(pool_.get())
                  .tableScan(rowType, {}, "", nullptr, assignments)
                  .planNode();
```

otherwise, `partitionKeys_` in `SplitReader` will be empty, and the query will fail with the following error:

```
  auto it = partitionKeys_->find(partitionKey);
  VELOX_CHECK(
      it != partitionKeys_->end(),
      "ColumnHandle is missing for partition key {}",
      partitionKey);
```
Note that the column name should be crafted too if it is a partitioned column.

To me this is not necessary for Iceberg, what do you think?

jinchengchenghh:
https://github.com/facebookincubator/velox/pull/12910 has merged, and the test has enabled, https://github.com/apache/incubator-gluten/pull/10910
Please update the PR title and PR description, as the interface argument requirement, partitionKeys_ should be set, do you mean after this change, we don't need to set partitionKeys? CC @yingsu00  What do you think about that?

PingLiuPing:
> #12910 has merged, and the test has enabled, [apache/incubator-gluten#10910](https://github.com/apache/incubator-gluten/pull/10910) Please update the PR title and PR description, as the interface argument requirement, partitionKeys_ should be set, do you mean after this change, we don't need to set partitionKeys? CC @yingsu00 What do you think about that?

Yes, with this change we don't need to set `partitionKeys_` when reading iceberg table.

PingLiuPing:
> https://github.com/facebookincubator/velox/pull/12910 has merged, and the test has enabled, https://github.com/apache/incubator-gluten/pull/10910

I know https://github.com/facebookincubator/velox/pull/12910 has been merged. Initially I have wrote this in this PR description to indicate this PR can only be merged after 12910.
Do you mean you have case to cover identity transform on decimal column already? After 12910 or before 12910 been merged?

jinchengchenghh:
This PR is not only related to decimal type, it is related to identity column transform. 
testPartitionValueTypes in https://github.com/apache/incubator-gluten/pull/10910 has covered all the identity transform data types.

PingLiuPing:
Thanks, ok, it is merged after https://github.com/facebookincubator/velox/pull/12910.
Just worrying there are special logic from Gluten and I might missed something when implement this PR.
Now we are clear this.

One more thing, you mentioned `infoColumns` earlier, I don't found any usage from Presto. Is this been used by Gluten?

jinchengchenghh:
It is used in Hive, it is metadata column, now it is not supported in Iceberg, but need to support in the future, iceberg also supports query metadata column

PingLiuPing:
> It is used in Hive, it is metadata column, now it is not supported in Iceberg, but need to support in the future, iceberg also supports query metadata column

Thanks, let's focus on iceberg, this PR doesn't change anything in Hive.
Ok, I will keep infoColumn in `IcebergSplitReader::adaptColumns`

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks.
I added the document.

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/IcebergSplitReader.h
PingLiuPing: Thanks.
I added document to this method.

velox/connectors/hive/iceberg/IcebergSplitReader.h
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
PingLiuPing: Yes.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks, updated.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks, I re-implement the test to more accurately cover the code changes.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks. Added makeColumnHandles to create the column handle map.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

PingLiuPing:
@mbasmanova I addressed the review comments, could you have a look please. Thank you.

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -141,4 +141,60 @@ uint64_t IcebergSplitReader::next(uint64_t size, VectorPtr& output) {
   return rowsScanned;
 }
 
+std::vector<TypePtr> IcebergSplitReader::adaptColumns(
+    const RowTypePtr& fileType,
+    const std::shared_ptr<const velox::RowType>& tableSchema) const {
```
mbasmanova: RowTypePtr

velox/connectors/hive/SplitReader.h
```diff
@@ -52,6 +52,63 @@ class HiveTableHandle;
 class HiveColumnHandle;
 class HiveConfig;
 
+template <TypeKind kind>
+VectorPtr newConstantFromString(
```
mbasmanova, line 73: Can we avoid moving this template into the header file? I believe all called used it together with VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH_ALL. Can we expose a non-template API and have its implementation in the .cpp?

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -284,6 +284,33 @@ class HiveIcebergTest : public HiveConnectorTestBase {
...
+      const RowTypePtr& rowType,
+      const std::vector<int>& partitionColumnIndices = {}) {
+    ColumnHandleMap assignments;
+    std::unordered_set<int> partitionIndices(
+        partitionColumnIndices.begin(), partitionColumnIndices.end());
+
+    for (int i = 0; i < rowType->size(); ++i) {
+      const auto& columnName = rowType->nameOf(i);
+      const auto& columnType = rowType->childAt(i);
+      auto columnHandleType = partitionIndices.count(i) > 0
```
mbasmanova: nit: partitionIndices.contains(i)

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +836,160 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+
+  // Read with new schema (c0 and c2 only, c1 removed).
+  auto plan = PlanBuilder(pool_.get()).tableScan(newRowType).planNode();
+
+  // Expected result: c0 and c2 have values, c1 is not present.
+  std::vector<RowVectorPtr> expectedVectors;
+  expectedVectors.push_back(makeRowVector({"c0", "c2"}, {c0, c2}));
+
+  createDuckDbTable(expectedVectors);
+  assertQuery(plan, icebergSplits, "SELECT * FROM tmp", 0);
```
mbasmanova: No need to create DuckDB table here. Just run the query and compare results to expectedVectors directly. See velox/exec/tests/utils/AssertQueryBuilder.h

```
  /// Run the query and compare results with 'expected'.
  std::shared_ptr<Task> assertResults(const RowVectorPtr& expected);
```

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks for the suggestion.
I have added a wrapper function and moved it cpp file.

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: @mbasmanova  Thanks for the comments. I have removed `createDuckDbTable`.

velox/connectors/hive/SplitReader.h
mbasmanova: Thanks. Please, document the new API.

velox/connectors/hive/SplitReader.h
```diff
@@ -52,6 +52,15 @@ class HiveTableHandle;
 class HiveColumnHandle;
 class HiveConfig;
 
+VectorPtr newConstantFromString(
+    const TypePtr& type,
+    const std::optional<std::string>& value,
+    vector_size_t size,
```
mbasmanova: Do we have a use case for setting size to something other than 1?

velox/connectors/hive/SplitReader.h
```diff
@@ -52,6 +52,15 @@ class HiveTableHandle;
...
+    velox::memory::MemoryPool* pool,
+    const std::string& sessionTimezone,
+    bool asLocalTime,
+    bool isPartitionDateDaysSinceEpoch = false);
```
mbasmanova: Do we need default value for this parameter?

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -141,4 +141,58 @@ uint64_t IcebergSplitReader::next(uint64_t size, VectorPtr& output) {
...
+    const RowTypePtr& tableSchema) const {
+  std::vector<TypePtr> columnTypes = fileType->children();
+  auto& childrenSpecs = scanSpec_->children();
+  // Iceberg table stores all column's data in data file.
+  for (const auto& childSpec : childrenSpecs) {
+    const std::string& fieldName = childSpec->fieldName();
+    if (auto iter = hiveSplit_->infoColumns.find(fieldName);
+        iter != hiveSplit_->infoColumns.end()) {
+      auto infoColumnType =
+          readerOutputType_->childAt(readerOutputType_->getChildIdx(fieldName));
```
mbasmanova: readerOutputType_->findChild(fieldName)

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +837,164 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  auto icebergSplits = makeIcebergSplits(dataFilePath->getPath());
+
+  // Read with new schema (c0 and c2 only, c1 removed).
+  auto plan = PlanBuilder(pool_.get()).tableScan(newRowType).planNode();
+
+  // Expected result: c0 and c2 have values, c1 is not present.
+  std::vector<RowVectorPtr> expectedVectors;
+  expectedVectors.push_back(makeRowVector({"c0", "c2"}, {c0, c2}));
+
+  AssertQueryBuilder(plan, duckDbQueryRunner_)
```
mbasmanova: drop duckDbQueryRunner_ argument; it is not needed

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +837,164 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  auto newRowType = ROW({"c0", "c1", "c2"}, {BIGINT(), INTEGER(), VARCHAR()});
+
+  // Write data file with old schema (only c0).
+  std::vector<RowVectorPtr> dataVectors;
+  auto c0 = makeFlatVector<int64_t>({100, 200, 300});
+  dataVectors.push_back(makeRowVector({"c0"}, {c0}));
+
+  auto dataFilePath = TempFilePath::create();
+  writeToFile(
+      dataFilePath->getPath(), dataVectors, config_, flushPolicyFactory_);
```
mbasmanova: This code repeats. Consider adding helper method to shorten to 

```
writeToFile(dataFilePath, dataVectors);
```

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +837,164 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+
+  auto icebergSplits = makeIcebergSplits(dataFilePath->getPath());
+
+  // Read with new schema (c0, c1, and c2).
+  auto plan = PlanBuilder(pool_.get())
```
mbasmanova: Drop pool_.get() parameter. It is not needed.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +837,164 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  // Read with table schema including partition columns.
+  auto plan = PlanBuilder(pool_.get())
+                  .tableScan(tableRowType, {}, "", tableRowType, assignments)
+                  .planNode();
+
+  // Expected result: c0 and c1 from file, region and year from partition keys.
+  auto expectedRegion = makeFlatVector<std::string>({"US", "US", "US"});
+  auto expectedYear = makeFlatVector<int32_t>({2025, 2025, 2025});
+  std::vector<RowVectorPtr> expectedVectors;
+  expectedVectors.push_back(makeRowVector(
```
mbasmanova, line 920: consider

```
makeRowVector(tableRowType->names(), {
   makeFlatVector<std::string>({"US", "US", "US"}),
   makeFlatVector<int32_t>({2025, 2025, 2025}),
})
```

No need for expectedRegion and expectedYear variables. Also, no need to repeat column names.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +837,164 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  auto fileRowType = ROW({"c0", "c1"}, {BIGINT(), VARCHAR()});
+  auto tableRowType =
+      ROW({"c0", "c1", "c2", "region"},
+          {BIGINT(), VARCHAR(), INTEGER(), DECIMAL(38, 5)});
+
+  // Write data file with c0 and c1.
+  std::vector<RowVectorPtr> dataVectors;
+  auto c0 = makeFlatVector<int64_t>({100, 200});
+  auto c1 = makeFlatVector<std::string>({"a", "b"});
+  dataVectors.push_back(makeRowVector({"c0", "c1"}, {c0, c1}));
```
mbasmanova: c0, c1, c2... are the default names, hence,

```
makeRowVector({
   makeFlatVector<int64_t>({100, 200}),
   makeFlatVector<std::string>({"a", "b"}),
});
```

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
PingLiuPing: Thanks.

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks. After checking, I found all the places are 1. I have removed this parameter.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/SplitReader.h
PingLiuPing: @mbasmanova Thanks for the comment.
We should keep the default value as false. Since this parameter is used is used to differentiate the DATE type value in `setPartitionValue` between hive and iceberg.

velox/connectors/hive/SplitReader.h
PingLiuPing: Sure.

velox/connectors/hive/SplitReader.h
mbasmanova: I was asking whether we need this parameter to have default value or can we ask the caller to always explicitly specify the value? Default parameters make the code harder to read.

velox/connectors/hive/SplitReader.h
```diff
@@ -47,6 +47,37 @@ class MemoryPool;
 
 namespace facebook::velox::connector::hive {
 
+/// Creates a constant vector from a string representation of a value.
```
mbasmanova: nit: a constant vector of size 1

velox/connectors/hive/SplitReader.h
```diff
@@ -47,6 +47,37 @@ class MemoryPool;
...
+///
+/// This function is primarily used to materialize partition column values and
```
mbasmanova: > This function is primarily used

Drop "This function is primarily". Start with an active verb: Used to ...

velox/connectors/hive/SplitReader.h
```diff
@@ -47,6 +47,37 @@ class MemoryPool;
...
+/// info columns (e.g., $path, $file_size) when reading Hive/Iceberg tables.
```
mbasmanova: Hive/Iceberg -> Hive and Iceberg

velox/connectors/hive/SplitReader.h
```diff
@@ -47,6 +47,37 @@ class MemoryPool;
...
+/// Partition values are stored as strings in HiveConnectorSplit::partitionKeys
+/// and need to be converted to their appropriate types.
+///
+/// @param type The target Velox type for the constant vector. Supports all
+/// scalar types including primitives, dates, timestamps, and decimals.
+/// @param value The string representation of the value to convert, formatted
+/// the same way as CAST(x as VARCHAR). Date values must be formatted using ISO
+/// 8601 as YYYY-MM-DD. If nullopt, creates a null constant vector.
+/// @param pool Memory pool for allocating the constant vector.
+/// @param asLocalTime If true and type is TIMESTAMP, interprets the string
```
mbasmanova: Love this documentation. Detailed and clear. Thank you for taking the time to write it.

velox/connectors/hive/SplitReader.h
```diff
@@ -47,6 +47,37 @@ class MemoryPool;
...
+///     value as local time and converts it to GMT. If false, treats the value
+///     as already in GMT.
+/// @param isPartitionDateDaysSinceEpoch If true and type is DATE, treats the
```
mbasmanova: asLocalTime and isPartitionDateDaysSinceEpoch parameters control how string representation of Timestamp and Date values are interpreted. The comments make it very clear. Yet, the parameter names are not consistent. How about, isLocalTimestamp and isDaysSinceEpoch.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -284,6 +285,33 @@ class HiveIcebergTest : public HiveConnectorTestBase {
     return splits;
   }
 
+  ColumnHandleMap makeColumnHandles(
+      const RowTypePtr& rowType,
+      const std::vector<int>& partitionColumnIndices = {}) {
```
mbasmanova: nit: if you change the type of the argument to std::unordered_set, then you won't need to create the set inside the function while callers will stay unmodified. Doing this helps reduce the amount of code and helps readability.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -772,42 +790,19 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
-           columnParseParameters)});
-
-  auto plan = PlanBuilder(pool_.get())
-                  .tableScan(rowType, {}, "", nullptr, assignments)
-                  .planNode();
+  auto assignments = makeColumnHandles(rowType, {1});
 
-  HiveConnectorTestBase::assertQuery(
+  auto plan =
+      PlanBuilder(pool_.get()).tableScan(rowType, {}, "", nullptr).planNode();
```
mbasmanova: drop pool_.get() argument; ditto other places

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -831,11 +826,146 @@ TEST_F(HiveIcebergTest, testPartitionedRead) {
...
+  auto dataFilePath = TempFilePath::create();
+  writeToFile(dataFilePath->getPath(), dataVectors);
+  auto icebergSplits = makeIcebergSplits(dataFilePath->getPath());
+
+  // Expected result: c0 has values, c1 and c2 are NULL.
+  std::vector<RowVectorPtr> expectedVectors;
+  expectedVectors.push_back(makeRowVector(
+      {c0,
+       makeNullConstant(TypeKind::INTEGER, 3),
+       makeNullConstant(TypeKind::VARCHAR, 3)}));
```
mbasmanova: add a comma after the last vector to improve readability

apply the same refactoring to other calls to makeRowVector

mbasmanova (commented):
Looks great. Thank you for iterating.

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks,

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks.

velox/connectors/hive/SplitReader.h
PingLiuPing: Thanks.

velox/connectors/hive/SplitReader.h
PingLiuPing: Ah, got your point. Yes I agree with you. Let me delete the default value and explicitly set it in caller.

velox/connectors/hive/SplitReader.h
PingLiuPing: > How about, isLocalTimestamp and isDaysSinceEpoch.

Thanks for suggestion, yes, these are better names suited for this function.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks for the suggestion.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: Thanks.
There are places where a vector will be used twice so I kept the local variable.
I removed all the local vector variables and get them from the rowVector when they are been used in second place.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
PingLiuPing: @mbasmanova  Thanks.
Removed most of them, and there is one place left where this pool can not be removed.

```
  // Test filter on non-partitioned date column
  std::vector<std::string> filters = {"ds = date'2018-04-06'"};
  plan = PlanBuilder(pool_.get()).tableScan(rowType, filters).planNode();
```

when removed it, the test cases crashed at https://github.com/facebookincubator/velox/blob/6851fd345323d68dbebef425817606620c4c5824/velox/buffer/Buffer.h#L392 since the pool is a NULL pointer.

But by reading this case carefully I don't think it makes sense. 
For iceberg:
1. There is no difference between reading partitioned table and non-partitioned table. 
2. Even we want test reading partitioned table, we should first write partitioned iceberg table, but this case does not write partitioned table at all.

So probably this case can be deleted, what do you think?

velox/connectors/hive/SplitReader.h
PingLiuPing: Cheers.

velox/connectors/hive/SplitReader.cpp
```diff
@@ -80,11 +78,27 @@ VectorPtr newConstantFromString(
...
   }
 }
 } // namespace
 
+VectorPtr newConstantFromString(
+    const TypePtr& type,
+    const std::optional<std::string>& value,
+    velox::memory::MemoryPool* pool,
+    bool asLocalTime,
+    bool isPartitionDateDaysSinceEpoch) {
```
mbasmanova: Let's rename these arguments as well.

velox/connectors/hive/SplitReader.cpp
```diff
@@ -28,17 +28,15 @@ namespace facebook::velox::connector::hive {
 namespace {
 
 template <TypeKind kind>
-VectorPtr newConstantFromString(
+VectorPtr newConstantFromStringImpl(
```
mbasmanova, line 31: Let's rename arguments and remove default value.

mbasmanova (approved):
Thanks.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
mbasmanova: I'm not sure I follow.

velox/connectors/hive/SplitReader.cpp
```diff
@@ -62,29 +60,45 @@ VectorPtr newConstantFromString(
...
       }
     }
     return std::make_shared<ConstantVector<T>>(
-        pool, size, false, type, std::move(copy));
+        pool, 1, false, type, std::move(copy));
   }
 }
 } // namespace
 
+VectorPtr newConstantFromString(
```
Yuhta, line 86: Can we move this into the anonymous namespace?

mbasmanova: @Yuhta This is a public API now.

velox/connectors/hive/SplitReader.cpp
PingLiuPing: Sure. I missed this.

PR #14520, opened 2025-08-19
feat: Add convenient getters for Variant arrays and maps

mbasmanova:
Summary:
Allow for extracting a std::vector of primitive values from a Variant.

The new `variant.array<int32_t>()` returns `std::vector<int32_t>`, while existing `variant.array()` returns `std::vector<Variant>`.

Differential Revision: D80559182


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #14523, opened 2025-08-19
feat(expression): Add ExprOptimizer for expression optimization

pramodsatya:
This work proposes a Velox Expression Optimizer API and framework that can be used by engines to rewrite Velox expressions to a more optimal form for comparison and evaluation. Such a functionality is useful in the Presto optimizer to obtain constant folded expressions that can be used in other optimizer rules.

### Design

At present, Velox offers a expression evaluation API to the engines. (https://github.com/facebookincubator/velox/blob/main/velox/examples/ExpressionEval.cpp#L156has an example usage). The optimizer API can be thought of as a first phase of this expression evaluation. 

The optimizer API :

- Takes an input TypedExpr like the expression evaluation
- Performs several logical expression rewrites that simplify the TypedExpr
- Does constant folding of sub-trees of the expression if applicable

The logical expression rewrites and constant folding are interleaved as the logical rewrites attempt to simplify the expression by constant folding its inputs if applicable. 
- Returns a TypedExpr to the caller

The original expression evaluation API builds on this optimize step to further compile  the TypedExpr to one or more Expr which is a evaluatable form of the input expression. The actual Expression evaluation executes the Expr for each input row vector of the data.

Since a lot of the new ExprOptimizer code is common with the original expression evaluation API, the code is reorganized to abstract the common functions and use them in the original compileExpressions API. 

### ExprOptimizer API
The application can use function optimizeExpressions in ExprOptimizer.h. This function takes as input a vector of TypedExprPtr and returns a vector of TypedExprPtr. Each individual TypedExprPtr in the input vector is optimized to the TypedExprPtr in the result vector at the same index. The original expression is returned if no optimization applies to the input expression.  If say a constant folding results in an exception during the optimization, then the result is wrapped in a fail function in the original expression.

```

std::vector<core::TypedExprPtr> optimizeExpressions(
    const std::vector<core::TypedExprPtr>& expressions,
    const std::shared_ptr<core::QueryCtx>& queryCtx,
    memory::MemoryPool* pool);
```
optimizeExpressions comprises 2 parts : constant folding and logical expression rewriting.

**Constant Folding**
The first part of the expression optimization is constant folding. A new utility function for constant folding is added in ExprUtils.h. This function uses the Velox constant evaluation function to constant fold subtrees of the input expression and return the folded expressions.

```
core::TypedExprPtr constantFold(
    const std::shared_ptr<core::QueryCtx>& queryCtx,
    memory::MemoryPool* pool);
```

**Expression rewrites (Code in ExprRewriter.h)**
The second part of the expression optimization is to execute logical rewrites of the input expression. Velox already does some logical rewrites in Vector Functions and has an API for ExpressionRewrite viz https://github.com/facebookincubator/velox/blob/main/velox/expression/VectorFunction.h#L284

Engines like Presto have several rewrites in their optimizer logic, and there are many new ideas of such rewrites as well. 

To organize the logical rewrites, the framework externalizes the ExpressionRewrite registrations and de-registrations and adds a couple basic ones in this PR. Expression rewrites for special form expressions are registered along with the registration of the respective `FunctionCallToSpecialForm` and described in the next section. Application of these special form expression rewrites is controlled with config `expression.apply_special_form_rewrites` and they are disabled by default, the user of `ExprOptimizer` can enable these rewrites with this config.

### Logical rewrites

A simple logical rewrite for conjunct expressions `AND`, `OR` is added to `ExprOptimizer` in this PR, more special form expression rewrites will be added in a follow-up PR. Expression rewrites will utilize the constant folding API introduced in the previous section:

- If any input to `AND` expression is or can be constant folded to `false`, rewrite conjunct expression `AND` with constant expression `false`. Similarly, If any input to `OR` expression is or can be constant folded to `true`, rewrite conjunct expression `OR` with constant expression `true`. Prune inputs to `AND` and `OR` expressions that are or can be constant folded to `true` and `false` respectively. Flatten inputs to `AND`, `OR`; eg: `a AND (b AND (c AND d))` -> `AND(a, b, c, d)`.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -437,6 +437,7 @@ jobs:
                 --max_expression_trees_per_step 2 \
                 --retry_with_try \
                 --enable_dereference \
+                --enable_expression_optimizer \
                 --duration_sec $DURATION \
                 --minloglevel=0 \
                 --stderrthreshold=2 \
--- a/velox/core/Expressions.h
+++ b/velox/core/Expressions.h
@@ -189,6 +189,16 @@ class CallTypedExpr : public ITypedExpr {
       : ITypedExpr{ExprKind::kCall, std::move(type), std::move(inputs)},
         name_(std::move(name)) {}
 
+  /// @param type Return type.
+  /// @param name Name of the function or special form.
+  /// @param inputs List of input expressions.
+  template <typename... TypedExprs>
+  CallTypedExpr(TypePtr type, std::string name, TypedExprs... inputs)
+      : CallTypedExpr(
+            std::move(type),
+            std::vector<TypedExprPtr>{std::forward<TypedExprs>(inputs)...},
+            std::move(name)) {}
+
   virtual const std::string& name() const {
     return name_;
   }
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -685,6 +685,11 @@ class QueryConfig {
   /// username.
   static constexpr const char* kClientTags = "client_tags";
 
+  /// Special form expression rewrites are applied if this is true. False by
+  /// default.
+  static constexpr const char* kExprApplySpecialFormRewrites =
+      "expression.apply_special_form_rewrites";
+
   bool selectiveNimbleReaderEnabled() const {
     return get<bool>(kSelectiveNimbleReaderEnabled, false);
   }
@@ -1232,6 +1237,10 @@ class QueryConfig {
 
   int32_t maxNumSplitsListenedTo() const {
     return get<int32_t>(kMaxNumSplitsListenedTo, 0);
+  }
+
+  bool exprApplySpecialFormRewrites() const {
+    return get<bool>(kExprApplySpecialFormRewrites, false);
   }
 
   std::string source() const {
--- a/velox/expression/ExprCompiler.cpp
+++ b/velox/expression/ExprCompiler.cpp
@@ -15,9 +15,12 @@
  */
 
 #include "velox/expression/ExprCompiler.h"
+#include "velox/expression/ConjunctExpr.h"
 #include "velox/expression/ConstantExpr.h"
 #include "velox/expression/Expr.h"
 #include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprUtils.h"
 #include "velox/expression/FieldReference.h"
 #include "velox/expression/LambdaExpr.h"
 #include "velox/expression/RowConstructor.h"
@@ -88,17 +91,6 @@ struct Scope {
   }
 };
 
-// Utility method to check eligibility for flattening.
-bool allInputTypesEquivalent(const TypedExprPtr& expr) {
-  const auto& inputs = expr->inputs();
-  for (int i = 1; i < inputs.size(); i++) {
-    if (!inputs[0]->type()->equivalent(*inputs[i]->type())) {
-      return false;
-    }
-  }
-  return true;
-}
-
 std::optional<std::string> shouldFlatten(
     const TypedExprPtr& expr,
     const std::unordered_set<std::string>& flatteningCandidates) {
@@ -108,47 +100,11 @@ std::optional<std::string> shouldFlatten(
     // inputs are of the same type.
     if (call->name() == expression::kAnd || call->name() == expression::kOr ||
         (flatteningCandidates.count(call->name()) &&
-         allInputTypesEquivalent(expr))) {
+         expression::utils::allInputTypesEquivalent(expr))) {
       return call->name();
     }
   }
   return std::nullopt;
-}
-
-bool isCall(const TypedExprPtr& expr, const std::string& name) {
-  if (expr->isCallKind()) {
-    return expr->asUnchecked<core::CallTypedExpr>()->name() == name;
-  }
-  return false;
-}
-
-// Recursively flattens nested ANDs, ORs or eligible callable expressions into a
-// vector of their inputs. Recursive flattening ceases exploring an input branch
-// if it encounters either an expression different from 'flattenCall' or its
-// inputs are not the same type.
-// Examples:
-// flattenCall: AND
-// in: a AND (b AND (c AND d))
-// out: [a, b, c, d]
-//
-// flattenCall: OR
-// in: (a OR b) OR (c OR d)
-// out: [a, b, c, d]
-//
-// flattenCall: concat
-// in: (array1, concat(array2, concat(array2, intVal))
-// out: [array1, array2, concat(array2, intVal)]
-void flattenInput(
-    const TypedExprPtr& input,
-    const std::string& flattenCall,
-    std::vector<TypedExprPtr>& flat) {
-  if (isCall(input, flattenCall) && allInputTypesEquivalent(input)) {
-    for (auto& child : input->inputs()) {
-      flattenInput(child, flattenCall, flat);
-    }
-  } else {
-    flat.emplace_back(input);
-  }
 }
 
 ExprPtr getAlreadyCompiled(const ITypedExpr* expr, ExprDedupMap* visited) {
@@ -181,7 +137,7 @@ std::vector<ExprPtr> compileInputs(
     } else {
       if (flattenIf.has_value()) {
         std::vector<TypedExprPtr> flat;
-        flattenInput(input, flattenIf.value(), flat);
+        expression::utils::flattenInput(input, flattenIf.value(), flat);
         for (auto& input_2 : flat) {
           compiledInputs.push_back(compileExpression(
               input_2,
@@ -349,15 +305,6 @@ std::vector<VectorPtr> getConstantInputs(const std::vector<ExprPtr>& exprs) {
     }
   }
   return constants;
-}
-
-core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
-  for (auto& rewrite : expressionRewrites()) {
-    if (auto rewritten = rewrite(expr)) {
-      return rewritten;
-    }
-  }
-  return expr;
 }
 
 ExprPtr compileCall(
@@ -573,7 +520,9 @@ ExprPtr compileExpression(
     memory::MemoryPool* pool,
     const std::unordered_set<std::string>& flatteningCandidates,
     bool enableConstantFolding) {
-  auto rewritten = rewriteExpression(expr);
+  const auto queryCtx = core::QueryCtx::create(
+      nullptr, core::QueryConfig{config.rawConfigsCopy()});
+  auto rewritten = expression::rewriteExpression(expr, queryCtx, pool);
   if (rewritten.get() != expr.get()) {
     scope->rewrittenExpressions.push_back(rewritten);
   }
--- a/velox/expression/ExprOptimizer.cpp
+++ b/velox/expression/ExprOptimizer.cpp
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/expression/Expr.h"
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprUtils.h"
+
+namespace facebook::velox::expression {
+
+std::vector<core::TypedExprPtr> optimizeExpressions(
+    const std::vector<core::TypedExprPtr>& expressions,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  std::vector<core::TypedExprPtr> optimizedExpressions;
+  optimizedExpressions.reserve(expressions.size());
+
+  for (auto& expression : expressions) {
+    auto rewritten = expression::rewriteExpression(expression, queryCtx, pool);
+    auto optimizedExpression =
+        utils::constantFold(rewritten ? rewritten : expression, queryCtx, pool);
+    optimizedExpressions.push_back(std::move(optimizedExpression));
+  }
+  return optimizedExpressions;
+}
+
+} // namespace facebook::velox::expression
--- a/velox/expression/ExprOptimizer.h
+++ b/velox/expression/ExprOptimizer.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/core/Expressions.h"
+#include "velox/core/QueryCtx.h"
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/VectorFunction.h"
+
+namespace facebook::velox::expression {
+
+/// Rewrites registered expressions to optimize the expression and
+/// constant folds the result.
+/// The function returns:
+/// - a new constant expression (if constant folded)
+/// - a fail function call expression due to an error during constant
+///   evaluation (for example, constant folding the expression 5/0 would result
+///   in an error)
+/// - the original expression (if it cannot be constant folded, for example,
+///   due to presence of a field expression)
+/// - a new rewritten expression (if it cannot be constant folded further, for
+///   example, COALESCE(a, COALESCE(a, b)) can be simplified to COALESCE(a, b))
+///
+/// Additional fail function details:
+/// When the expression is attempted to be constant folded and an error occurs,
+/// 5/0 or a BIGINT value can't be cast to TINYINT due to overflow, the fail
+/// function may be returned. If that is the case, no further evaluation of the
+/// arguments is performed and the fail function is returned for the expression.
+std::vector<core::TypedExprPtr> optimizeExpressions(
+    const std::vector<core::TypedExprPtr>& expressions,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool);
+} // namespace facebook::velox::expression
--- a/velox/expression/ExprRewrite.cpp
+++ b/velox/expression/ExprRewrite.cpp
@@ -13,20 +13,27 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#pragma once
+
+#include <set>
+
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprRewriteRegistry.h"
 
 namespace facebook::velox::expression {
 
-constexpr const char* kConjunct = "conjunct";
-constexpr const char* kAnd = "and";
-constexpr const char* kOr = "or";
-constexpr const char* kSwitch = "switch";
-constexpr const char* kIn = "in";
-constexpr const char* kIf = "if";
-constexpr const char* kFail = "fail";
-constexpr const char* kCoalesce = "coalesce";
-constexpr const char* kCast = "cast";
-constexpr const char* kTryCast = "try_cast";
-constexpr const char* kTry = "try";
+core::TypedExprPtr rewriteExpression(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  const auto& rewriteRegistry = exec::expressionRewriteRegistry();
+  const auto& rewriteNames = rewriteRegistry.getExpressionRewriteNames();
+  for (const auto& name : rewriteNames) {
+    auto expressionRewriteFunc = *rewriteRegistry.getExpressionRewrite(name);
+    if (auto rewritten = expressionRewriteFunc(expr, queryCtx, pool)) {
+      return rewritten;
+    }
+  }
+  return nullptr;
+}
 
 } // namespace facebook::velox::expression
--- a/velox/expression/ExprRewriteRegistry.h
+++ b/velox/expression/ExprRewriteRegistry.h
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "folly/Synchronized.h"
+#include "velox/expression/ExprRewrite.h"
+
+namespace facebook::velox::exec {
+
+using ExprRewriteRegistryType = std::
+    unordered_map<std::string, std::unique_ptr<expression::ExpressionRewrite>>;
+class ExpressionRewriteRegistry {
+ public:
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerExpressionRewrite(
+      const std::string& name,
+      std::unique_ptr<expression::ExpressionRewrite> expressionRewrite);
+
+  /// Clears all registered expression re-writes.
+  void unregisterExpressionRewrites();
+
+  expression::ExpressionRewrite* FOLLY_NULLABLE
+  getExpressionRewrite(const std::string& name) const;
+
+  std::vector<std::string> getExpressionRewriteNames() const;
+
+ private:
+  folly::Synchronized<ExprRewriteRegistryType> rewriteRegistry_;
+};
+
+const ExpressionRewriteRegistry& expressionRewriteRegistry();
+
+void registerExpressionRewrite(
+    const std::string& name,
+    std::unique_ptr<expression::ExpressionRewrite> expressionRewrite);
+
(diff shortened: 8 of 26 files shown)


velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,137 @@
...
+      expr->type(), std::move(optimizedInputs), expr->name());
+}
+
+core::TypedExprPtr rewriteExpression(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  const auto& rewriteRegistry = exec::expressionRewriteRegistry();
+  const auto& rewriteNames =
+      exec::expressionRewriteRegistry().getExpressionRewriteNames();
```
czentgr: We can use the variable here or not have the variable.

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <vector>
```
czentgr: I wonder if this comes in from one of the the headers already and if we really need it.
We don't need it for the map but the map is also coming in already.

velox/expression/ExprCompiler.cpp
```diff
@@ -575,7 +550,9 @@ ExprPtr compileExpression(
     memory::MemoryPool* pool,
     const std::unordered_set<std::string>& flatteningCandidates,
     bool enableConstantFolding) {
-  auto rewritten = rewriteExpression(expr);
+  const auto queryCtx = core::QueryCtx::create(
+      nullptr, core::QueryConfig{config.rawConfigsCopy()});
+  auto rewritten = expression::rewriteExpression(expr, queryCtx, pool);
```
czentgr, line 525: This will rewrite our expression. Because we register the special forms in the normal setup we automatically get expression rewrite and don't need a separate fuzzer?
We can run with the normal expression fuzzer and get the generated expression compiled and rewritten.

pramodsatya: Reverted, thanks!

velox/core/Expressions.h
```diff
@@ -189,6 +189,16 @@ class CallTypedExpr : public ITypedExpr {
       : ITypedExpr{ExprKind::kCall, std::move(type), std::move(inputs)},
         name_(std::move(name)) {}
 
+  /// @param type Return type.
+  /// @param inputs List of input expressions.
```
czentgr: Nit: lets swap with next line to have same order as arguments.

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+          // AND (.., false, ..) -> false
+          return foldedExpr;
+        }
+        break;
+      case utils::ConstantEvalResult::IS_NOT_CONSTANT:
+        allInputsConstant = false;
+        optimizedInputs.push_back(inputExpr);
+        break;
+      default:
+        return expr;
```
czentgr: We should probably return nullptr here. Technically, we should never come here.

velox/expression/tests/ExprTest.cpp
```diff
@@ -3643,39 +3643,6 @@ class NoOpVectorFunction : public exec::VectorFunction {
 };
 } // namespace
 
-TEST_P(ParameterizedExprTest, applyFunctionNoResult) {
```
czentgr: What happened with these tests? Why were they removed?

pramodsatya: Thanks for pointing it out, they were no longer applicable with the rewrites enabled by default, but with the query config controlling special form rewrites these can be added back.

velox/core/QueryConfig.h
```diff
@@ -685,6 +685,11 @@ class QueryConfig {
   /// username.
   static constexpr const char* kClientTags = "client_tags";
 
+  /// Special form expression rewrites are applied if this is true. False by
+  /// default.
+  static constexpr const char* kExprApplySpecialFormRewrites =
```
czentgr, line 690: We might need to document this as well if we use this.

czentgr (commented):
@mbasmanova This is the second PR for the special form expression rewrite. It contains the framework for registration and includes one of the rewrites. The last PR is adding the other rewrites. If you have time please take a look. This PR incorporates all the previous feedback that applied to this code.

mbasmanova:
@czentgr Love the PR description. Thank you for writing it.

> Application of these special form expression rewrites is controlled with config expression.apply_special_form_rewrites and they are disabled by default, the user of ExprOptimizer can enable these rewrites with this config.

Curious, why do we need a flag for this? Can we not always apply the rewrite?

.github/workflows/scheduled.yml
```diff
@@ -437,6 +437,7 @@ jobs:
                 --max_expression_trees_per_step 2 \
                 --retry_with_try \
                 --enable_dereference \
+                --enable_expression_optimizer \
```
mbasmanova, line 440: What does this do? Would it make sense to extract this change into a separate PR?

velox/core/Expressions.h
```diff
@@ -189,6 +189,16 @@ class CallTypedExpr : public ITypedExpr {
       : ITypedExpr{ExprKind::kCall, std::move(type), std::move(inputs)},
         name_(std::move(name)) {}
 
+  /// @param type Return type.
+  /// @param name Name of the function or special form.
+  /// @param inputs List of input expressions.
+  template <typename... TypedExprs>
+  CallTypedExpr(TypePtr type, std::string name, TypedExprs... inputs)
```
mbasmanova, line 196: Nice. Would it be possible to extract this in a separate PR so it can land sooner?

velox/core/QueryConfig.h
```diff
@@ -1234,6 +1239,10 @@ class QueryConfig {
     return get<int32_t>(kMaxNumSplitsListenedTo, 0);
   }
 
+  bool exprApplySpecialFormRewrites() const {
```
mbasmanova, line 1242: Why do we need this?

mbasmanova (commented):
@czentgr Do you think you could convert PR description into a documentation article?

velox/expression/ExprCompiler.cpp
```diff
@@ -573,7 +520,9 @@ ExprPtr compileExpression(
     memory::MemoryPool* pool,
     const std::unordered_set<std::string>& flatteningCandidates,
     bool enableConstantFolding) {
-  auto rewritten = rewriteExpression(expr);
+  const auto queryCtx = core::QueryCtx::create(
+      nullptr, core::QueryConfig{config.rawConfigsCopy()});
```
mbasmanova, line 524: Why do we need to make a copy of configs?

velox/expression/ExprConstants.h
```diff
@@ -17,7 +17,7 @@
 
 namespace facebook::velox::expression {
 
-constexpr const char* kConjunt = "conjunct";
+constexpr const char* kConjunct = "conjunct";
```
mbasmanova, line 20: 'conjunct' is not a function or special form name, is it? How is this used? Why do we need this constant?

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+#pragma once
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::expression::utils {
+
+/// Indicates the result of evaluating an expression to a boolean constant.
+enum class ConstantEvalResult {
+  IS_NOT_CONSTANT = 0,
```
mbasmanova, line 24: naming: kCamelCase

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::expression::utils {
+
+/// Indicates the result of evaluating an expression to a boolean constant.
+enum class ConstantEvalResult {
```
mbasmanova, line 23: Since this is only about boolean constants, perhaps, rename to PredicateEvalResult

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+  IS_NOT_CONSTANT = 0,
+  IS_NULL,
+  IS_TRUE,
+  IS_FALSE
+};
+
+/// Helper function to evaluate an expression to a boolean constant.
```
mbasmanova, line 30: Would you clarify what happens if 'expr' is not constant?

naming: perhaps, evalConstantPredicate or tryEvalConstantPredicate?

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+/// Returns true if expr is of type CallTypedExpr and the expr name matches
+/// the passed name. Otherwise, returns false.
+bool isCall(const core::TypedExprPtr& expr, const std::string& name);
+
+/// This function attempts to constant fold expr using the queryCtx
+/// configuration. When successful it returns the constant expression. On
+/// failure, it depends if an exception was thrown. If an exception was thrown a
+/// call expression for the fail function is returned, otherwise expr is
+/// returned unchanged.
+core::TypedExprPtr tryConstantFold(
```
mbasmanova, line 42: Given that we have this function, why do we need evalExprAsConstant and why is that one doesn't take QueryCtx or pool? Should we use tryConstantFold, then simply check if result is constant and if it is whether it is null, true or false?

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool);
+
+/// Constant fold subtrees of the input expression and return the folded
```
mbasmanova, line 47: Would you clarify how is this different from tryConstantFold?

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+/// expression.
+core::TypedExprPtr constantFold(
+    const core::TypedExprPtr& expr,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool);
+
+bool allInputTypesEquivalent(const core::TypedExprPtr& expr);
```
mbasmanova, line 54: Please, document.

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+
+/// Recursively flattens nested ANDs, ORs or eligible callable expressions into
+/// a vector of their inputs. Recursive flattening ceases exploring an input
+/// branch if it encounters either an expression different from 'flattenCall' or
+/// its inputs are not the same type. Examples: flattenCall: AND in: a AND (b
```
mbasmanova, line 59: formating

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+/// AND (c AND d)) out: [a, b, c, d]
+///
+/// flattenCall: OR
+/// in: (a OR b) OR (c OR d)
+/// out: [a, b, c, d]
+///
+/// flattenCall: concat
+/// in: (array1, concat(array2, concat(array2, intVal))
+/// out: [array1, array2, concat(array2, intVal)]
+void flattenInput(
```
mbasmanova, line 69: I assume this is a refactoring. Wondering if it can be extracted into its own PR and landed sooner?

velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,74 @@
...
+    const core::TypedExprPtr& input,
+    const std::string& flattenCall,
+    std::vector<core::TypedExprPtr>& flat);
```
mbasmanova, line 72: Can we just return this vector and not use an output parameter?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "velox/core/Expressions.h"
+#include "velox/expression/ExprConstants.h"
+
+namespace facebook::velox::expression::utils {
+
+ConstantEvalResult evalExprAsConstant(const core::TypedExprPtr& expr) {
+  if (expr->isConstantKind()) {
+    auto constantExpr = expr->asUnchecked<core::ConstantTypedExpr>();
+    if (constantExpr->isNull()) {
+      return ConstantEvalResult::IS_NULL;
```
mbasmanova, line 27: kNull, kTrue, kFalse, kNotConstant

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+ */
+
+#include "velox/expression/ExprUtils.h"
+#include "velox/core/Expressions.h"
+#include "velox/expression/ExprConstants.h"
+
+namespace facebook::velox::expression::utils {
+
+ConstantEvalResult evalExprAsConstant(const core::TypedExprPtr& expr) {
+  if (expr->isConstantKind()) {
```
mbasmanova, line 24: nit: if (!expr->isConstantKind()) { return ...}

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+    memory::MemoryPool* pool) {
+  std::vector<core::TypedExprPtr> foldedInputs;
+  for (auto& input : expr->inputs()) {
+    foldedInputs.push_back(constantFold(input, queryCtx, pool));
+  }
+
+  core::TypedExprPtr result;
+  bool isField = false;
+  if (auto callExpr =
+          std::dynamic_pointer_cast<const core::CallTypedExpr>(expr)) {
```
mbasmanova, line 78: expr->asUnchecked

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+    result = std::make_shared<core::CallTypedExpr>(
```
mbasmanova, line 79: Don't we want to fold is foldedInputs are all constant?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+        callExpr->type(), foldedInputs, callExpr->name());
+  } else if (
+      auto castExpr =
+          std::dynamic_pointer_cast<const core::CastTypedExpr>(expr)) {
+    VELOX_CHECK(!foldedInputs.empty());
+    if (foldedInputs.at(0)->type() == expr->type()) {
```
mbasmanova, line 85: Should we compare Type objects, not pointers?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  core::TypedExprPtr result;
+  bool isField = false;
+  if (auto callExpr =
+          std::dynamic_pointer_cast<const core::CallTypedExpr>(expr)) {
+    result = std::make_shared<core::CallTypedExpr>(
+        callExpr->type(), foldedInputs, callExpr->name());
+  } else if (
+      auto castExpr =
+          std::dynamic_pointer_cast<const core::CastTypedExpr>(expr)) {
```
mbasmanova, line 83: Would you mention this re-write in the PR description and documentation of this function?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+  } else if (
+      auto constantExpr =
+          std::dynamic_pointer_cast<const core::ConstantTypedExpr>(expr)) {
+    return constantExpr;
+  } else if (
+      auto field =
+          std::dynamic_pointer_cast<const core::FieldAccessTypedExpr>(expr)) {
+    isField = true;
+    result = field;
+  } else if (
```
mbasmanova, line 100: This if block seems redundant

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+          expr->type(), foldedInputs, castExpr->isTryCast());
+    }
+  } else if (
+      auto constantExpr =
+          std::dynamic_pointer_cast<const core::ConstantTypedExpr>(expr)) {
+    return constantExpr;
+  } else if (
+      auto field =
+          std::dynamic_pointer_cast<const core::FieldAccessTypedExpr>(expr)) {
+    isField = true;
```
mbasmanova, line 98: should this check field->isInputColumn() to distinguish between input reference struct member access?

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+    result = field;
+  } else if (
+      auto concatExpr =
+          std::dynamic_pointer_cast<const core::ConcatTypedExpr>(expr)) {
+    result = concatExpr;
+  } else {
+    result = expr;
+  }
+
+  auto folded = !isField ? tryConstantFold(result, queryCtx, pool) : result;
```
mbasmanova, line 108: drop 'folded' variable

velox/expression/ExprUtils.cpp
```diff
@@ -0,0 +1,136 @@
...
+  if (isCall(input, flattenCall) && allInputTypesEquivalent(input)) {
+    for (auto& child : input->inputs()) {
+      flattenInput(child, flattenCall, flat);
+    }
+  } else {
+    flat.emplace_back(input);
+  }
+}
+
+// Utility method to check eligibility for flattening.
```
mbasmanova, line 125: Move this comment to .h

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+core::TypedExprPtr rewriteConjunctExpression(
+    const core::TypedExprPtr& input,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  if (!queryCtx->queryConfig().exprApplySpecialFormRewrites() ||
+      (!utils::isCall(input, kAnd) && !utils::isCall(input, kOr))) {
+    return nullptr;
+  }
+
+  auto expr = std::dynamic_pointer_cast<const core::CallTypedExpr>(input);
```
mbasmanova, line 31: asUnchecked?

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+  const bool isAnd = (expr->name() == kAnd) ? true : false;
+  // If all inputs are AND or OR then we can flatten the inputs into a vector
+  // before further optimizing.
+  auto canFlatten = utils::allInputTypesEquivalent(expr);
```
mbasmanova, line 35: This check is redundant. AND must have all inputs of the same type. Same for OR.

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+  std::vector<core::TypedExprPtr> flat;
+  if (canFlatten) {
+    utils::flattenInput(input, expr->name(), flat);
+  }
+
+  const auto& inputsToOptimize = canFlatten ? flat : expr->inputs();
```
mbasmanova, line 41: Was this logic copy-pasted from somewhere? We used to also handle flattening for 'concat'. Does that still work?

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+  bool allInputsConstant = true;
+  bool hasNullInput = false;
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  core::TypedExprPtr nullInput = nullptr;
+  for (const auto& inputExpr : inputsToOptimize) {
+    auto foldedExpr = utils::tryConstantFold(inputExpr, queryCtx, pool);
+    if (utils::isCall(foldedExpr, kFail)) {
+      return foldedExpr;
```
mbasmanova, line 49: That logic doesn't seem right. c OR fail(...) is not equivalent to fail(...)

velox/expression/SpecialFormRewrites.cpp
```diff
@@ -0,0 +1,91 @@
...
+        optimizedInputs.push_back(inputExpr);
+        break;
+      default:
+        return nullptr;
+    }
+  }
+
+  if (allInputsConstant && hasNullInput) {
+    return nullInput;
+  } else if (optimizedInputs.empty()) {
```
mbasmanova, line 83: drop else after if

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,58 @@
...
+#pragma once
+
+#include "folly/Synchronized.h"
+#include "velox/expression/ExprRewrite.h"
+
+namespace facebook::velox::exec {
+
+using ExprRewriteRegistryType = std::
+    unordered_map<std::string, std::unique_ptr<expression::ExpressionRewrite>>;
+class ExpressionRewriteRegistry {
```
mbasmanova, line 25: nit: Add empty line

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,58 @@
...
+  /// Clears all registered expression re-writes.
+  void unregisterExpressionRewrites();
+
+  expression::ExpressionRewrite* FOLLY_NULLABLE
+  getExpressionRewrite(const std::string& name) const;
+
+  std::vector<std::string> getExpressionRewriteNames() const;
+
+ private:
+  folly::Synchronized<ExprRewriteRegistryType> rewriteRegistry_;
```
mbasmanova, line 48: nit: rewriteRegistry_ -> registry_

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,58 @@
...
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerExpressionRewrite(
+      const std::string& name,
```
mbasmanova, line 36: What is this name for? Would you document?

.github/workflows/scheduled.yml
czentgr: This is running the fuzzer with the expression rewrite/optimization turned on. That means the external optimizeExpression API is called in the fuzzer separately prior to being compiled.

One reason we did this is because we don't want to initially apply all the special form rewrites automatically during compilation (to phase them in basically and let them sit with the fuzzer).
There is a new session property that we added to control the usage during compilation.

velox/core/Expressions.h
czentgr: Are there also other users for it (or would you plan on using it)? We have one user with the `fail` function generation.

velox/expression/ExprConstants.h
czentgr: It is the name of the rewrite for AND and OR expressions when registering `rewriteConjunctExpression`.

mbasmanova: I see. Do we ned this? It is confusing to mix function names with just names.

velox/core/Expressions.h
mbasmanova: This is a very helpful convenience ctor. It would be nice to make it available. It also helps to make this PR a bit smaller by extracting and landing independent pieces.

.github/workflows/scheduled.yml
mbasmanova: I believe there is a refactoring piece where you moved existing logic from the ExprCompiler into a util. Maybe we can extract that change into its own PR and land.

Then, there is logic to add new re-writes and that logic might be "experimental". We would probably want to enable it in Fuzzer separately and give heads up to @kagamiori .

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include "velox/core/Expressions.h"
+#include "velox/core/QueryCtx.h"
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/VectorFunction.h"
+
+namespace facebook::velox::expression {
+
+/// Rewrites registered expressions to optimize the expression and
```
mbasmanova, line 25: > Rewrites registered expressions

is this accurate? I believe we have registered rewrites, not expressions.

velox/expression/ExprOptimizer.h
```diff
@@ -0,0 +1,46 @@
...
+/// - a new constant expression (if constant folded)
+/// - a fail function call expression due to an error during constant
+///   evaluation (for example, constant folding the expression 5/0 would result
+///   in an error)
+/// - the original expression (if it cannot be constant folded, for example,
+///   due to presence of a field expression)
+/// - a new rewritten expression (if it cannot be constant folded further, for
+///   example, COALESCE(a, COALESCE(a, b)) can be simplified to COALESCE(a, b))
+///
+/// Additional fail function details:
```
mbasmanova, line 37: I assume it is possible that a subexpression is converted to a fail function, no?

velox/expression/ExprOptimizer.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+std::vector<core::TypedExprPtr> optimizeExpressions(
+    const std::vector<core::TypedExprPtr>& expressions,
+    const std::shared_ptr<core::QueryCtx>& queryCtx,
+    memory::MemoryPool* pool) {
+  std::vector<core::TypedExprPtr> optimizedExpressions;
+  optimizedExpressions.reserve(expressions.size());
+
+  for (auto& expression : expressions) {
+    auto rewritten = expression::rewriteExpression(expression, queryCtx, pool);
```
mbasmanova, line 31: Any particular reason rewriteExpression doesn't include constant folding?

velox/expression/ExprCompiler.cpp
czentgr: This is because later in the flow we use `tryEvaluateConstantExpression` that uses a shared_ptr for the QueryCtx. So we would need to make a copy of it from the raw pointer somewhere. The function itself uses the raw pointer from the shared_ptr. So it seems easy to change the API of `tryEvaluateConstantExpression` to a raw pointer for now and not have to deal with the copy (also our new APIs used a shared_ptr that we can change to raw pointer to avoid a copy). Later we could refactor to have queryCtx be a shared_ptr in all functions and declare a `QueryCtxPtr` to be used similar to what we do with other instances of shared_ptr.

mbasmanova: > it seems easy to change the API of tryEvaluateConstantExpression to a raw pointer

Let's do that. In a separate PR.

velox/expression/ExprOptimizer.cpp
czentgr: The expression compiler will also do constant fold after calling this rewrite function. We didn't want to touch the flow of the expression compiler (constant fold would be called twice I think) and this is the external API for PrestoC++ so we apply it here instead outside of the rewrite.

mbasmanova: > We didn't want to touch the flow of the expression compiler 

Why not?

velox/expression/ExprUtils.h
czentgr: `tryConstantFold` is a wrapper for `constantFold`. In case of an exception, the latter throws and the former replaces the exception with the fail function. We don't use constantFold by itself and can combine the two instead. We thought having these separate might be cleaner.

velox/expression/ExprUtils.h
czentgr: `evalExprAsConstant` is a helper function to look at the output of an attempt to constant fold this expression. It is not evaluating the expression itself. We do need a better name. 
We have this helper because for other special forms we will do the same thing and depending on the result of the constant fold we do different types of processing. This was part of a rewrite to remove code duplication in the special forms to unify the processing and looking at the result.

velox/expression/SpecialFormRewrites.cpp
czentgr: Yes, we will verify it still works. But that flattening occurs in the expression compiler and is unchanged by the code here. Here we are looking at and/or call expressions to flatten (nested ands or nested ors only).

pramodsatya:
Completed in other PRs.

PR #14524, opened 2025-08-19
feat(expression): Add special form expression rewrites

pramodsatya:
Depends on #14717. 

Adds following special form expression rewrites:

- Prunes inputs to `COALESCE` that are `NULL`. Deduplicate inputs to `COALESCE`, eg: `COALESCE(a, b, a)` -> `COALESCE(a, b)`. Flatten nested `COALESCE` expressions, eg: `COALESCE(COALESCE(a, COALESCE(b, c)), 1)` -> `COALESCE(a, b, c, 1)`. Simplify `COALESCE` to the first non `NULL` constant input if all inputs seen before were `NULL`. 
- `IF` expression has three inputs: `IF(condition, value1, value2)`. If the condition is or can be constant folded to `true` or `false`, simplify the conditional expression to `value1` or `value2` respectively.
- `SWITCH` expression should have following inputs: `SWITCH(condition1, value1, condition2, value2, ..., conditionN, valueN, defaultValue)`. Ignores conditions that can be constant folded to `false`. If `conditionN` can be constant folded to `true` and no other conditions referencing variables were encountered before, reduces expression to `valueN`.
- Remove values from `IN`-list that do not match with the value being compared, after both expressions are constant folded. Also deduplicates values from constant folded IN-list.

### DIFF
--- a/velox/expression/CMakeLists.txt
+++ b/velox/expression/CMakeLists.txt
@@ -56,6 +56,7 @@ velox_add_library(
   RowConstructor.cpp
   SimpleFunctionRegistry.cpp
   SpecialFormRegistry.cpp
+        SpecialFormRewrites.cpp
   SwitchExpr.cpp
   TryExpr.cpp
   VectorFunction.cpp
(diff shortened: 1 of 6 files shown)


czentgr:
@pramodsatya Please rebase, there is another PR that fixes the build issue on macOS.
Also I suspect the fail function is triggering the fuzzer failures because the expression rewrite is called which could produce the fail function.

pramodsatya:
Completed in other PRs.

PR #14536, opened 2025-08-20
fix: use generic types for QDigest distribution

jonhehir:
Summary: The recently added `QuantileDigest::getDistributionFunction` accepts doubles for the `rangeStart` and `rangeEnd` parameters, rather than the generic type `T` (which is what is ultimately used in the returned distribution function). This commit changes the input parameters to match the types used in the return value.

Reviewed By: kagamiori, chenkuei

Differential Revision: D80181764


kagamiori (approved):
LGTM. Thanks!

PR #14538, opened 2025-08-20
feat: Add a delegate function to HiveConnector to control what datasource gets created

devavret:
This PR adds a `DataSourceDelegate` along with registering functions HiveConnector to give control over what datasource gets created from it.

Context:
While integrating Velox with cudf, we'd like to use cuDF's parquet reader with the existing velox `TableScan` operator and `HiveConnector`. In order to do that, we would like to use the aforementioned delegate mechanism to create a CudfDataSource which will use cudf's file readers and cudf's expression evaluation for processing the remaining filter.

Contributes to https://github.com/facebookincubator/velox/issues/14183

### DIFF
--- a/velox/connectors/hive/HiveConnector.cpp
+++ b/velox/connectors/hive/HiveConnector.cpp
@@ -28,6 +28,12 @@ using namespace facebook::velox::exec;
 
 namespace facebook::velox::connector::hive {
 
+HiveConnector::DataSourceDelegate HiveConnector::dataSourceDelegate_ = nullptr;
+
+void HiveConnector::registerDataSourceDelegate(DataSourceDelegate delegate) {
+  dataSourceDelegate_ = std::move(delegate);
+}
+
 namespace {
 std::vector<std::unique_ptr<HiveConnectorMetadataFactory>>&
 hiveConnectorMetadataFactories() {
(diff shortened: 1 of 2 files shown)


mbasmanova (commented):
@devavret Curious, why not create a new connector? What is that you'd like to reuse from HiveConnector?

majetideepak:
@mbasmanova the cudf connector is reading parquet files on gpus today. We could extend it to other file formats in the future. In a way it is a custom HiveDataSource. Pushing this lower into the HiveConnector allows clients such as Prestissimo to use it seamlessly without any changes to the protocol or plan.

mbasmanova:
@majetideepak I assume changes are still required to register the delegate connector. If so, it seems cleaner to simply have a switch on some config to decide whether to use vanilla HiveConnector or Cudf-based one. Am I missing something?

majetideepak:
@mbasmanova the delegation here is only for the HiveDataSource. One alternative is to extend the CudfConnector from the HiveConnector and customize the CreateDataSource. We could then register the cudfConnector instead of the HiveConnector. Is this what you meant?

mbasmanova:
> One alternative is to extend the CudfConnector from the HiveConnector and customize the CreateDataSource. 

That might be a reasonable option.

devavret:
Closing because we're changing our approach to extend HiveConnector by making CudfHiveConnector its subclass

PR #14540, opened 2025-08-20
refactor: Make local merge source queue size configurable

Yuhta:
Summary:
Add the new query config `local_merge_source_queue_size` to control how
many vectors we buffer in a local merge source.  This can be used to control the
memory usage of the pipeline.

This change does not alter any behavior of existing applications, which continue to use the default value (2) same as before.

Differential Revision: D80626475

### DIFF
--- a/velox/exec/Merge.h
+++ b/velox/exec/Merge.h
@@ -290,6 +290,7 @@ class SpillMerger : public std::enable_shared_from_this<SpillMerger> {
           spillReadFilesGroup,
       vector_size_t maxOutputBatchRows,
       uint64_t maxOutputBatchBytes,
+      int mergeSourceQueueSize,
       const common::SpillConfig* spillConfig,
       const std::shared_ptr<folly::Synchronized<common::SpillStats>>&
           spillStats,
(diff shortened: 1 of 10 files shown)


velox/exec/Merge.h
```diff
@@ -293,7 +293,8 @@ class SpillMerger : public std::enable_shared_from_this<SpillMerger> {
       const common::SpillConfig* spillConfig,
       const std::shared_ptr<folly::Synchronized<common::SpillStats>>&
           spillStats,
-      velox::memory::MemoryPool* pool);
+      velox::memory::MemoryPool* pool,
+      int mergeSourceQueueSize);
```
xiaoxmeng: move mergeSourceQueueSize ahead of spill config?

velox/exec/tests/MergerTest.cpp
```diff
@@ -159,11 +159,13 @@ class MergerTest : public OperatorTestBase {
         pool());
   }
 
-  static std::vector<std::shared_ptr<MergeSource>> createMergeSources(int num) {
+  static std::vector<std::shared_ptr<MergeSource>> createMergeSources(
+      int numSources,
```
xiaoxmeng: nit: use size_t and the same for other use cases? Thanks!

xiaoxmeng (approved):
@Yuhta  LGTM % nits!

PR #14541, opened 2025-08-20
feat: Support flat map vector fuzzing

peterenescu:
Summary: Adds support for fuzzing flat map vectors. Currently, this only adds vector generation in VectorFuzzer and adds a small unit test to the filter functions suite to test in a more specified manner. For one, this will help test more complicated lambdas that fuzzer doesn't support. Additionally, while I continue to investigate exactly how we want to test in expression fuzzer this may be a good coverage test (since expression fuzzer doesn't support encoding testing and doesn't let vectorfuzzer know which function is testing, my playground is a bit hacky to get it to work locally).

Differential Revision: D80296393

### DIFF
--- a/velox/vector/fuzzer/VectorFuzzer.cpp
+++ b/velox/vector/fuzzer/VectorFuzzer.cpp
@@ -254,6 +254,36 @@ AbstractInputGeneratorPtr maybeGetCustomTypeInputGenerator(
 
   return nullptr;
 }
+
+template <TypeKind kind>
+VectorPtr extractDistinctKeys(VectorPtr allKeys, memory::MemoryPool* pool) {
+  using TCpp = typename TypeTraits<kind>::NativeType;
+
+  std::vector<TCpp> distinctKeys;
+  std::unordered_set<TCpp> keySet;
+  auto flatKeys = allKeys->asFlatVector<TCpp>();
+  auto result = BaseVector::create<FlatVector<TCpp>>(
+      allKeys->type(), allKeys->size(), pool);
+
+  auto index = 0;
+  for (int i = 0; i < flatKeys->size(); ++i) {
+    auto key = flatKeys->valueAt(i);
+    if (keySet.find(key) == keySet.end()) {
+      keySet.insert(key);
+      distinctKeys.push_back(key);
+      result->set(index++, key);
+    }
+  }
+
+  result->resize(index);
+  if (allKeys->getDistinctValueCount().has_value())
+    VELOX_CHECK_EQ(
+        result->size(),
+        allKeys->getDistinctValueCount().value(),
+        "extracted distinct keys size should match input distinct value count");
+  return result;
+}
+
 } // namespace
 
 VectorPtr VectorFuzzer::fuzzNotNull(
@@ -466,6 +496,17 @@ VectorPtr VectorFuzzer::fuzzFlat(
   }
 }
 
+VectorPtr VectorFuzzer::fuzzUniqueFlatNotNull(
+    const TypePtr& type,
+    vector_size_t size) {
+  // Generate a random vector of non-null keys.
+  auto allKeys = fuzzFlatNotNull(type, getElementsVectorLength(opts_, size));
+
+  // Extract and return the distinct keys.
+  return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
+      extractDistinctKeys, type->kind(), allKeys, pool_);
+}
+
 VectorPtr VectorFuzzer::fuzzMap(
     const TypePtr& keyType,
     const TypePtr& valueType,
@@ -531,19 +572,24 @@ VectorPtr VectorFuzzer::fuzzComplex(const TypePtr& type, vector_size_t size) {
     }
 
     case TypeKind::MAP: {
-      // Do not initialize keys and values inline in the fuzzMap call as C++
-      // does not specify the order they'll be called in, leading to
-      // inconsistent results across platforms.
-      const auto& keyType = type->asMap().keyType();
-      const auto& valueType = type->asMap().valueType();
-      auto length = getElementsVectorLength(opts_, size);
+      if (opts_.allowFlatMapVector && coinToss(opts_.flatMapRatio)) {
+        return fuzzFlatMap(
+            type->asMap().keyType(), type->asMap().valueType(), size);
+      } else {
+        // Do not initialize keys and values inline in the fuzzMap call as C++
+        // does not specify the order they'll be called in, leading to
+        // inconsistent results across platforms.
+        const auto& keyType = type->asMap().keyType();
+        const auto& valueType = type->asMap().valueType();
+        auto length = getElementsVectorLength(opts_, size);
 
-      auto keys = opts_.normalizeMapKeys || !opts_.containerHasNulls
-          ? fuzzNotNull(keyType, length)
-          : fuzz(keyType, length);
-      auto values = opts_.containerHasNulls ? fuzz(valueType, length)
-                                            : fuzzNotNull(valueType, length);
-      return fuzzMap(keys, values, size);
+        auto keys = opts_.normalizeMapKeys || !opts_.containerHasNulls
+            ? fuzzNotNull(keyType, length)
+            : fuzz(keyType, length);
+        auto values = opts_.containerHasNulls ? fuzz(valueType, length)
+                                              : fuzzNotNull(valueType, length);
+        return fuzzMap(keys, values, size);
+      }
     }
 
     default:
@@ -713,6 +759,32 @@ MapVectorPtr VectorFuzzer::fuzzMap(
       values);
 }
 
+FlatMapVectorPtr VectorFuzzer::fuzzFlatMap(
+    const TypePtr& keyType,
+    const TypePtr& valueType,
+    vector_size_t size) {
+  // All keys present in the entirety of the FlatMap column.
+  VectorPtr distinctKeys =
+      fuzzUniqueFlatNotNull(keyType, opts_.containerLength);
+
+  // inMaps buffer and mapValues vector should be the same size.
+  std::vector<VectorPtr> mapValues(distinctKeys->size());
+  std::vector<BufferPtr> inMaps(distinctKeys->size());
+  for (int i = 0; i < mapValues.size(); ++i) {
+    mapValues[i] = fuzz(valueType, size);
+    inMaps[i] = fuzzFlatMapInMap(size);
+  }
+
+  return std::make_shared<FlatMapVector>(
+      pool_,
+      MAP(keyType, valueType),
+      fuzzNulls(size),
+      size,
+      distinctKeys,
+      mapValues,
+      inMaps);
+}
+
 RowVectorPtr VectorFuzzer::fuzzInputRow(
     const RowTypePtr& rowType,
     const std::vector<AbstractInputGeneratorPtr>& inputGenerators) {
@@ -817,6 +889,26 @@ BufferPtr VectorFuzzer::fuzzIndices(
     rawIndices[i] = rand<vector_size_t>(rng_) % baseVectorSize;
   }
   return indices;
+}
+
+BufferPtr VectorFuzzer::fuzzFlatMapInMap(vector_size_t size) {
+  VELOX_CHECK_GE(size, 0);
+
+  // When inMap buffer is null, key is present in all rows.
+  if (size == 0 || coinToss(opts_.inMapNullRatio)) {
+    return nullptr;
+  }
+
+  auto inMap = AlignedBuffer::allocate<bool>(size, pool_, 0);
+  auto* mutableInMap = inMap->asMutable<uint64_t>();
+
+  for (auto row = 0; row < size; ++row) {
+    if (coinToss(opts_.inMapRatio)) {
+      bits::setBit(mutableInMap, row);
+    }
+  }
+
+  return inMap;
 }
 
 std::pair<int8_t, int8_t> VectorFuzzer::randPrecisionScale(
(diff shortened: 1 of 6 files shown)


velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -143,6 +144,9 @@ class VectorFuzzer {
 
     /// Data spec for randomly generated data.
     fuzzer::DataSpec dataSpec{false, false};
+
+    /// Toggle for functions that support flat map vector encodings.
+    bool supportsFlatMapVectors{false};
```
pedroerp: for consistency, maybe `allowFlatMapVector`

velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -262,6 +266,18 @@ class VectorFuzzer {
   MapVectorPtr
   fuzzMap(const VectorPtr& keys, const VectorPtr& values, vector_size_t size);
 
+  /// Generates a fuzzed flat map vector using the given key and value types.
+  ///
+  /// Specifically, this function will fuzz keys values and use them to generate
```
pedroerp: nit: you can remove the "specifically"

velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -382,6 +398,10 @@ class VectorFuzzer {
   /// pointing to (baseVectorSize-1).
   BufferPtr fuzzIndices(vector_size_t size, vector_size_t baseVectorSize);
 
+  /// Generate a random inMap buffer (for use with flat map vectors) of 'size'
+  /// number of rows. This should randomize the keys per row in the map.
+  BufferPtr fuzzInMap(vector_size_t size);
```
pedroerp: maybe for clarity call this `fuzzFlatMapInMap()`

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -819,6 +882,20 @@ BufferPtr VectorFuzzer::fuzzIndices(
   return indices;
 }
 
+BufferPtr VectorFuzzer::fuzzInMap(vector_size_t size) {
+  VELOX_CHECK_GE(size, 0);
+  auto inMap = AlignedBuffer::allocate<bool>(size, pool_, 0);
+  auto mutableInMap = inMap->asMutable<uint64_t>();
+
+  for (auto row = 0; row < size; ++row) {
+    if (coinToss(.25)) {
```
pedroerp: make this configurable as well; it could be useful to change this for test purposes.

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +749,33 @@ MapVectorPtr VectorFuzzer::fuzzMap(
       values);
 }
 
+FlatMapVectorPtr VectorFuzzer::fuzzFlatMap(
+    const TypePtr& keyType,
+    const TypePtr& valueType,
+    vector_size_t size) {
+  // generate a vector of keys and extract the distinct ones
```
pedroerp: nit: use capital case and period for comments (full sentences).

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +749,33 @@ MapVectorPtr VectorFuzzer::fuzzMap(
...
+  auto allKeys = fuzzFlatNotNull(keyType, getElementsVectorLength(opts_, size));
```
pedroerp: we could also make this configurable (`opt_.flatMapMaxDistinctKeys`) or something similar.

pedroerp: maybe we could create a `fuzzUniqueFlatNotNull()` function?

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +749,33 @@ MapVectorPtr VectorFuzzer::fuzzMap(
...
+  VectorPtr distinctKeys = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
+      extractDistinctKeys, keyType->kind(), allKeys, pool_);
+
+  // inMap buffer and mapValues vector should be the same size
+  std::vector<VectorPtr> mapValues(distinctKeys->size());
+  std::vector<BufferPtr> inMaps(distinctKeys->size());
```
pedroerp, line 772: let's also toss a coin and leave a few inMap buffers nullptr (this should mean all keys pairs are in the map)

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +749,33 @@ MapVectorPtr VectorFuzzer::fuzzMap(
...
+  for (int i = 0; i < mapValues.size(); ++i) {
+    mapValues[i] = fuzzFlatNotNull(valueType, size);
```
pedroerp: this doesn't need to be not null. Actually, it shoudn't

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +749,33 @@ MapVectorPtr VectorFuzzer::fuzzMap(
...
+    inMaps[i] = fuzzInMap(size);
+  }
+
+  return std::make_shared<FlatMapVector>(
+      pool_,
+      MAP(keyType, valueType),
+      nullptr,
```
pedroerp: we should also add top level nulls.

pedroerp (commented):
Thanks @peterenescu . A few comments inline

velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -143,6 +144,19 @@ class VectorFuzzer {
 
     /// Data spec for randomly generated data.
     fuzzer::DataSpec dataSpec{false, false};
+
+    /// FlatMap options:
+    ///
+    /// Toggle for functions that support flat map vector encodings.
+    bool allowFlatMapVector{false};
```
pedroerp, line 151: nit: blank line before the next comment.

velox/vector/fuzzer/VectorFuzzer.h
```diff
@@ -143,6 +144,19 @@ class VectorFuzzer {
...
+    /// Flat map vs regular map generation ratio.
+    double flatMapRatio{.5};
+    /// Probability of a particular key-value pair being in a particular
```
pedroerp: not sure I understand what this means by just reading the comment.

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -819,6 +891,26 @@ BufferPtr VectorFuzzer::fuzzIndices(
   return indices;
 }
 
+BufferPtr VectorFuzzer::fuzzFlatMapInMap(vector_size_t size) {
+  VELOX_CHECK_GE(size, 0);
```
pedroerp, line 895: perhaps we should return nullptr in this case? (if size is zero)

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -819,6 +891,26 @@ BufferPtr VectorFuzzer::fuzzIndices(
...
+
+  // When inMap buffer is null, key is present in all rows.
+  if (coinToss(opts_.inMapNullRatio)) {
+    return nullptr;
+  }
+
+  auto inMap = AlignedBuffer::allocate<bool>(size, pool_, 0);
+  auto mutableInMap = inMap->asMutable<uint64_t>();
```
pedroerp: nit: we usually use "auto*" to signal this is a pointer.

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +759,32 @@ MapVectorPtr VectorFuzzer::fuzzMap(
       values);
 }
 
+FlatMapVectorPtr VectorFuzzer::fuzzFlatMap(
+    const TypePtr& keyType,
+    const TypePtr& valueType,
+    vector_size_t size) {
+  // All keys present in the entirety of the FlatMap column.
+  VectorPtr distinctKeys =
+      fuzzUniqueFlatNotNull(keyType, getElementsVectorLength(opts_, size));
```
pedroerp: shouldn't this be `opts.containerLength` or `opts.complexElementsMaxSize`? This has to do with the number of key value pairs per record, not overall (times `size`).

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -713,6 +759,32 @@ MapVectorPtr VectorFuzzer::fuzzMap(
...
+
+  // inMaps buffer and mapValues vector should be the same size.
+  std::vector<VectorPtr> mapValues(distinctKeys->size());
+  std::vector<BufferPtr> inMaps(distinctKeys->size());
+  for (int i = 0; i < mapValues.size(); ++i) {
+    mapValues[i] = fuzzFlat(valueType, size);
```
pedroerp: for extra fuzz-ness, this doesn't need to be flat.

velox/vector/fuzzer/VectorFuzzer.cpp
```diff
@@ -254,6 +254,36 @@ AbstractInputGeneratorPtr maybeGetCustomTypeInputGenerator(
 
   return nullptr;
 }
+
+template <TypeKind kind>
+VectorPtr extractDistinctKeys(VectorPtr allKeys, memory::MemoryPool* pool) {
```
pedroerp, line 259: maybe take allKeys explicitly as a FlatVectorPtr

pedroerp (commented):
A few more comments

velox/vector/fuzzer/VectorFuzzer.h
peterenescu: How about
```
Probability of a particular inMap BufferPtr bit being set, 
signifying whether or not the corresponding key is present 
in that row's map.
```

velox/vector/fuzzer/VectorFuzzer.cpp
peterenescu: What is the benefit of such a change?
`extractDistinctKeys` is called from `fuzzUniqueFlatNotNull` which calls `fuzzFlatNotNull` and stores its value as a VectorPtr as `allKeys` (input to `extractDistinctKeys`):
```
VectorPtr VectorFuzzer::fuzzUniqueFlatNotNull(
    const TypePtr& type,
    vector_size_t size) {
  // Generate a random vector of non-null keys.
  auto allKeys = fuzzFlatNotNull(type, getElementsVectorLength(opts_, size));

  // Extract and return the distinct keys.
  return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
      extractDistinctKeys, type->kind(), allKeys, pool_);
}
```
If we expect a FlatVectorPtr as input to `extractDistinctKeys` we would need to template-ize its caller function `fuzzUniqueFlatNotNull` which I'm concerned might get a little messy and make it harder for future fuzz use cases to use it (as it would require a dispatch).

Perhaps there's a cleaner way to do so that I am unaware of.

pedroerp: not a big deal, but assuming this function only really works with FlatVectors (see line 264), it's probably a better API design to make it take FlatVector explicitly. But it's if this is going to make it less convenient to dispatch, if you can't use the macros we provide.

peterenescu: The only other way I can think is either add another helper function to dispatch from or dispatch directly from fuzzMap:
```
FlatMapVectorPtr VectorFuzzer::fuzzFlatMap(
    const TypePtr& keyType,
    const TypePtr& valueType,
    vector_size_t size) {
  // All keys present in the entirety of the FlatMap column.
  VectorPtr distinctKeys = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
      fuzzUniqueFlatNotNull, keyType->kind(), keyType, opts_.containerLength);
...
}

...

template <TypeKind kind>
VectorPtr VectorFuzzer::fuzzUniqueFlatNotNull(
    const TypePtr& type,
    vector_size_t size) {
  using T = typename TypeTraits<kind>::NativeType;

  auto flatKeys =
      std::dynamic_pointer_cast<FlatVector<T>>(fuzzFlatNotNull(type, size));
  return extractDistinctKeys<T>(flatKeys, pool_);
}

...

template <typename T>
VectorPtr extractDistinctKeys(
    FlatVectorPtr<T>& keys,
    memory::MemoryPool* pool) {
...
// same logic as before
...
}
```
The only drawback with this version is `fuzzUniqueFlatNotNull` becomes the only `fuzz...` function to take a template argument. I think I'd prefer the original but let me know your thoughts

pedroerp: thanks for thinking through this. Agreed that the original is probably better because it's more concise.

velox/vector/fuzzer/VectorFuzzer.cpp
peterenescu: Done, added in the fuzzFlatMapInMap function itself

velox/vector/fuzzer/VectorFuzzer.h
peterenescu: Done

velox/vector/fuzzer/VectorFuzzer.cpp
peterenescu: Sure, added below with coinToss

pedroerp: you probably want to remove this line then.

pedroerp (approved):
Thanks @peterenescu

PR #14543, opened 2025-08-21
feat: Add core::IExpr::replaceInputs API

mbasmanova:
Summary:
This API is helpful for re-writing expression trees in https://github.com/facebookexperimental/verax/pull/217

Specifically, it is used to split a SELECT with a mix of aggregations and scalar expressions into Agg followed by Project.

For example, SELECT sum(a) / count(1) FROM t is rewritten as

```
- Project: s / c
   - Agg: s := sum(a), c := count(1)
```

Differential Revision: D80670441

### DIFF
--- a/velox/parse/Expressions.h
+++ b/velox/parse/Expressions.h
@@ -23,10 +23,14 @@ namespace facebook::velox::core {
 
 class InputExpr : public IExpr {
  public:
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #14544, opened 2025-08-21
velox(feat): add Task/Operator/Driver Stats toString() utils

hdikeman:
Summary:
Needed a way to summarize plan execution for the native execution project. I was unable to find a TaskStats logging function

Adding Task/Operator/DriverStat logging utilities

The caller is responsible for synchronization, no special effort is made to ensure that the stat structures are not being modified during logging. Ideally a copy would be created and the logger invoked on the copy

Differential Revision: D80673402


mbasmanova:
@hdikeman There is printPlanWithStats - https://facebookincubator.github.io/velox/develop/debugging.html

hdikeman:
Thanks @mbasmanova ! That's probably what I want
Though I wonder if DriverStats get pulled into that? I don't see them referenced. I suppose they are probably all rolled into the Operator-level execution time stats though
Thanks again!

Issue #14489, opened 2025-08-15
Should VELOX_BUILD_TEST_UTILS apply to benchmark utils?


mbasmanova:
### Problem description

We are seeing that QueryBenchmarkBase and ExpressionBenchmarkBuilder.cpp do not get build when VELOX_BUILD_TEST_UTILS is ON, but VELOX_ENABLE_BENCHMARKS and VELOX_BUILD_TESTING is off. 

Is this a bug? Or do we need a separate option to add benchmark utils when not building benchmarks?

CC: @kgpai @feilong-liu @assignUser @czentgr 

### System information

n/a

### CMake log

```bash

```

assignUser:
For me it sounds logical to only build benchmark utils when benchmarks are built? What are they used for outside of benchmarks?

czentgr:
I'd have the same question as @assignUser . Why would they need to be built without benchmarks? Are they used for something inside Meta that is not reflected in OSS?

mbasmanova:
These are used for application-specific benchmarks. Just like test utils are used for app-specific tests.

See https://github.com/facebookexperimental/verax/blob/main/axiom/runner/tests/LocalRunnerTestBase.h

```
class LocalRunnerTestBase : public velox::exec::test::HiveConnectorTestBase {
```

mbasmanova:
To clarify, when building Axiom, we'd like to speed up the build by skipping building Velox tests and benchmarks, but we still need to build test utils and benchmark utils.

kgpai:
Should we move QueryBenchmarkBase and ExpressionBenchmarkBuilder.cpp to be part of VELOX_BUILD_TEST_UTILS and have VELOX_ENABLE_BENCHMARKS depend on VELOX_BUILD_TEST_UTILS ?

mbasmanova:
> Should we move QueryBenchmarkBase and ExpressionBenchmarkBuilder.cpp to be part of VELOX_BUILD_TEST_UTILS and have VELOX_ENABLE_BENCHMARKS depend on VELOX_BUILD_TEST_UTILS ?

I'm fine with that. @assignUser @czentgr Do you have an opinion?

czentgr:
I'm fine with that too. I suppose we don't have that many different categories of utils. I could see a scenario where the utils grow such that we get complaints that unrelated utils (e.g. benchmark utils not used in unit tests) are being built. I suppose for benchmarking it is only these two (for now).

PR #14548, opened 2025-08-21
build: Restrict building benchmark utilities to VELOX_BUILD_TEST_UTILS

mbasmanova:
Summary:
We'd like to build benchmark utilities without building any benchmarks.

Fixes https://github.com/facebookincubator/velox/issues/14489

Differential Revision: D80693228

### DIFF
--- a/velox/benchmarks/CMakeLists.txt
+++ b/velox/benchmarks/CMakeLists.txt
@@ -11,55 +11,60 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-add_subdirectory(basic)
 
-set(velox_benchmark_deps
-    velox_type
-    velox_vector
-    velox_vector_fuzzer
-    velox_expression
-    velox_parse_parser
-    velox_parse_utils
-    velox_parse_expression
+if(${VELOX_BUILD_TEST_UTILS})
+  set(velox_benchmark_deps
+      velox_type
+      velox_vector
+      velox_vector_fuzzer
+      velox_expression
+      velox_parse_parser
+      velox_parse_utils
+      velox_parse_expression
+      velox_serialization
+      Folly::folly
+      Folly::follybenchmark
+      ${DOUBLE_CONVERSION}
+      gflags::gflags
+      glog::glog)
+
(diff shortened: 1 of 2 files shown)


mbasmanova:
@assignUser @czentgr @feilong-liu @kgpai Would you help review this change?

velox/benchmarks/CMakeLists.txt
```diff
@@ -11,55 +11,59 @@
...
-  velox_serialization
-  velox_encode
-  velox_type
-  velox_type_fbhive
-  velox_caching
-  velox_vector_test_lib
-  Folly::folly
-  Folly::follybenchmark
-  fmt::fmt)
+if(${VELOX_BUILD_TEST_UTILS})
```
czentgr, line 15: Let's move this block before the addition of the subdirectories.
The components in the subdirectories depend on the `velox_benchmark_deps` target so it is logical to define it first.

velox/benchmarks/CMakeLists.txt
```diff
@@ -11,55 +11,59 @@
...
-    gflags::gflags
-    glog::glog)
-
-add_library(velox_benchmark_builder ExpressionBenchmarkBuilder.cpp)
-target_link_libraries(
-  velox_benchmark_builder ${velox_benchmark_deps})
-# This is a workaround for the use of VectorTestBase.h which includes gtest.h
-target_link_libraries(
-  velox_benchmark_builder GTest::gtest)
+if(${VELOX_ENABLE_BENCHMARKS_BASIC})
```
kgpai: Are there places that use benchmarks basic ? Should we just get rid of it  ?  @majetideepak @czentgr Is this used internally inside IBM ?

velox/benchmarks/CMakeLists.txt
mbasmanova: @czentgr Updated. Would you take another look?

velox/benchmarks/CMakeLists.txt
czentgr: I'm pretty certain we don't use this flags because we don't actually run the benchmarks anywhere.

velox/benchmarks/CMakeLists.txt
czentgr: Thanks, looks good and was already merged.

PR #14102, opened 2025-07-15
build: Update bundled Apache Arrow to 21.0.0 from 15.0.0

kou:
The latest Apache Arrow C++ release is 21.0.0.

21.0.0 includes many improvements and fixes. It also includes bundled zlib upgrade that fixes #14029. 

This update bundled Apache Arrow to 21.0.0.
This also removes patches. It will reduce maintenance costs.

Fixes #14029

PR #14549, opened 2025-08-21
build: Add missing find_package(Thrift) in FindArrow.cmake

### LINKED ISSUE #14547: Thrift dependecy is missing with system Apache Arrow
### Problem description

https://github.com/facebookincubator/velox/pull/14102 introduced this problem.

https://github.com/facebookincubator/velox/pull/14102#issuecomment-3208239392

> I'm running into multiple problems on macOS.
> 
> 1. link error not finding `-lthrift` for some executables. This is caused by the fact that `find_package(thrift)` and the alias mapping was removed from `FindArrow.cmake`. Some executableslibraries (e.g. `velox_dwio_parquet_thrift`) in the parquet area have `thrift` as link dependency. If CMake cannot find the library it just adds `-lthrift` and hopes for the best.
> 
> 
>    ```
>    ld: library 'thrift' not found
>    c++: error: linker command failed with exit code 1 (use -v to see invocation)
>    ```
> 
> 2. On macOS specifically, if there is still an old version of Arrow installed (e.g. 15) and we try to use the bundled arrow (to avoid the missing thrift issue for example) then we get link errors because the CMake prefix picks up the includes from the old version (they appear first in the search list) but links in the new version of the library.
>     Question: the README with the versions was not updated. Looks like you wanted this change to be backwards compatible to work with Arrow 15 and not force an upgrade?
> 
> 
> I don't understand yet why the CI worked fine on Linux. The install step doesn't install the dynamic library which would work if it uses `-lthrift`. So something is different compared to macOS perhaps. For macOS it is clear why it worked. It is because it uses the bundling and has no prior versions installed. I'm inclined to add the find_package back to the FindArrow.cmake (this solved my problem). What do you think? What was the reason it was removed?

### System information

macOS

### CMake log

```bash

```

kou:
https://github.com/facebookincubator/velox/pull/14102 removed `find_package(Thrift)` from `FindArrow.cmake` but it was wrong. If we remove `find_package(Thrift)` from `FindArrow.cmake`, Apache Thrift isn't found with system Apache Arrow.

We need Apache Thrift for our bundled Apache Parquet implementation. System Apache Arrow doesn't depend on Apache Thrift. So we need to find Apache Thrift explicitly with system Apache Arrow.

This also changes CMake target name for Apache Thrift to `thrift::thrift` from `thrift`. If we use no namespace CMake target name (`thrift`) and the CMake target doesn't exit, CMake treats it as `-lthrift`. If we use namespace CMake target (`thrift::thrift`) and the CMake target doesn't exist, CMake reports an error that `thrift::thrift` doesn't exist. It will help us to find this problem.

This also improves bundled Apache Arrow. We don't build bundled Apache Thrift in Apache Arrow when system Apache Thrift is found. It will reduce needless build time.

Fixes #14547

### DIFF
--- a/CMake/resolve_dependency_modules/arrow/CMakeLists.txt
+++ b/CMake/resolve_dependency_modules/arrow/CMakeLists.txt
@@ -15,8 +15,10 @@ project(Arrow)
 
 if(VELOX_ENABLE_ARROW)
   if(Thrift_FOUND)
+    set(ARROW_WITH_THRIFT OFF)
     set(THRIFT_SOURCE "SYSTEM")
   else()
+    set(ARROW_WITH_THRIFT ON)
     set(THRIFT_SOURCE "BUNDLED")
   endif()
   # Apache Arrow 21.0.0 uses archive.apache.org as the default mirror, which
@@ -31,7 +33,7 @@ if(VELOX_ENABLE_ARROW)
   set(ARROW_CMAKE_ARGS
       -DARROW_PARQUET=OFF
       -DARROW_DEPENDENCY_SOURCE=AUTO
-      -DARROW_WITH_THRIFT=ON
+      -DARROW_WITH_THRIFT=${ARROW_WITH_THRIFT}
       -DARROW_WITH_LZ4=ON
       -DARROW_WITH_SNAPPY=ON
       -DARROW_WITH_ZLIB=ON
@@ -51,8 +53,10 @@ if(VELOX_ENABLE_ARROW)
       -DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH})
   set(ARROW_LIBDIR ${ARROW_PREFIX}/install/${CMAKE_INSTALL_LIBDIR})
 
-  add_library(thrift STATIC IMPORTED GLOBAL)
-  if(NOT Thrift_FOUND)
+  if(Thrift_FOUND)
+    set(ARROW_THRIFT_LIB)
+  else()
+    add_library(thrift::thrift STATIC IMPORTED GLOBAL)
     set(THRIFT_SOURCE_ROOT ${ARROW_PREFIX}/src/arrow_ep-build/_deps/thrift-src)
     set(THRIFT_BUILD_ROOT ${ARROW_PREFIX}/src/arrow_ep-build/_deps/thrift-build)
     set(THRIFT_INCLUDE_DIR ${THRIFT_BUILD_ROOT}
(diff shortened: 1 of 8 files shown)


kou:
@czentgr Could you try this?

CMake/resolve_dependency_modules/arrow/CMakeLists.txt
```diff
@@ -15,8 +15,10 @@ project(Arrow)
 
 if(VELOX_ENABLE_ARROW)
   if(Thrift_FOUND)
+    set(ARROW_WITH_THRIFT OFF)
```
czentgr, line 18: I don't think this will work in all cases.
If the user somehow decides to bundle arrow (there is no bundling for thrift) but has the downlevel thrift (0.16) in the system the build will fail. Arrow 21 requires a newer thrift (I tried that).

My change was adding this at the bottom to FindArrow
```
find_package(Thrift)
if(Thrift_FOUND AND NOT TARGET thrift)
  add_library(thrift ALIAS thrift::thrift)
endif()
```
But we need it to force to use the bundled thrift if arrow is bundled to make them match properly.
Basically, we need to bundle Thrift if Arrow is bundled and override any variables set by the system found Thrift unless it matches the expected version maybe?

This wouldn't be a problem if we force users to use the latest version because then we only have to deal with one version combination and can rely on what should be in the system.

PingLiuPing (commented):
Tested this change on Macos, it works.

CMake/resolve_dependency_modules/arrow/CMakeLists.txt
kou: > I don't think this will work in all cases. If the user somehow decides to bundle arrow (there is no bundling for thrift) but has the downlevel thrift (0.16) in the system the build will fail. Arrow 21 requires a newer thrift (I tried that).

Hmm. It's strange. Apache Arrow requires Apache Thrift 0.11.0 or later: https://github.com/apache/arrow/blob/apache-arrow-21.0.0/cpp/cmake_modules/ThirdpartyToolchain.cmake#L1808-L1818

System Apache Thrift 0.16 should work. Could you share the error message you got?

FYI: Bundled Apache Arrow doesn't use Apache Thrift because we disable Apache Parquet build by `-DARROW_PARQUET=no`. `-DARROW_WITH_THRIFT=ON` just reuses Apache Thrift bundling feature in Apache Arrow. I think that we should create `CMake/resolve_dependency_modules/thrift/`. If you agree with it, I can do it in this PR.

> My change was adding this at the bottom to FindArrow
> 
> ```
> find_package(Thrift)
> if(Thrift_FOUND AND NOT TARGET thrift)
>   add_library(thrift ALIAS thrift::thrift)
> endif()
> ```

I agree that this is a minimal fix. But we can't detect similar problem if we keep using `thrift` (not `thrift::thrift`). Because `thrift` (no namespaced CMake target) is treated as `-lthrift` when `thrift` CMake target isn't defined. So I think that we should use namespaced Cmake target (`thrift::thrift`). 
 
> But we need it to force to use the bundled thrift if arrow is bundled to make them match properly. Basically, we need to bundle Thrift if Arrow is bundled and override any variables set by the system found Thrift unless it matches the expected version maybe?

Bundled Apache Arrow doesn't use Apache Thrift. Apache Thrift is used only for our forked Apache Parquet implementation in `velox/dwio/parquet/`. I think that `velox/dwio/parquet/` works with Apache Thrift 0.11.0 or later because the original Apache Parquet implementation in Apache Arrow still requires Apache Thrift 0.11.0 or later.

kou: https://github.com/facebookincubator/velox/pull/14567 is for building bundled Apache Thrift separately.

kou:
I close this in favor of #14942.

PR #14551, opened 2025-08-21
feat: Add hash and operator== for core::IExpr

mbasmanova:
Summary: These APIs are helpful for re-writing expression trees in https://github.com/facebookexperimental/verax/pull/217

Differential Revision: D80704925


velox/parse/IExpr.h
```diff
@@ -92,9 +114,38 @@ class IExpr {
...
+  bool compareAliasAndInputs(const IExpr& other) const {
+    if (alias() != other.alias() || inputs().size() != other.inputs().size()) {
+      return false;
+    }
+
+    for (size_t i = 0; i < inputs().size(); ++i) {
+      if (!(*inputs()[i] == *other.inputs()[i])) {
+        return false;
+      }
+    }
```
xiaoxmeng, line 126: Do we need to compare alias?

mbasmanova: Yes, I think so. It is part of the expression.

PR #14552, opened 2025-08-21
fix(build): Benchmark compilation

czentgr:
With the C++20 change using volatile is an error.
There was one benchmark that did use this. but it didn’t come up in the CI which didn’t seem to build the benchmarks at all. As a result the CI needed a fix to ensure proper compilation.

### DIFF
--- a/.github/workflows/linux-build-base.yml
+++ b/.github/workflows/linux-build-base.yml
@@ -109,10 +109,10 @@ jobs:
           else
             # cuDF (unsupported for Clang) and Faiss (link issue when using Clang)
             # are excluded for Clang compilation and need to be added back when using GCC.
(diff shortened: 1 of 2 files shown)


karthikeyann:
```suggestion
    volatile int total = 0;
for (auto i = 0; i < 1000; i++) {
auto functions = getFunctionSignatures();
auto size = functions.size();
total = total + size;
```
Keeping `volatile` keywords and replacing `+=` will fix the compilation error too.

czentgr:
@karthikeyann thanks, that is interesting. The reason I removed it was because I don't see why `volatile` was used here. `total` should not be optimized away because it is used later. In addition, if we did have such variable we can use folly::benchmark `doNotOptimizeAway` function to be more clear on what the intent was.

kou:
This caused CI failure: #14869

PR #14569, opened 2025-08-22
fix: macro VELOX_ASSERT_ERROR_STATUS has variable conflict

marin-ma:
Fix error "Local variable const Status status has already been defined or declared"
when using `VELOX_ASSERT_ERROR_STATUS` multiple times in the same scope.


marin-ma:
@rui-mo Could you help to review? Thanks.

jkhaliqi (approved):

PR #14574, opened 2025-08-22
fix: Ignore sfmsketch type in ExpressionFuzzer

kgpai:
Summary:
SFMSketch doesnt across two runs even with same seed doesnt produce the same sketch the size produced by sketch differs ; See : https://www.internalfb.com/code/fbsource/[82343deea988]/fbcode/velox/functions/prestosql/types/fuzzer_utils/SfmSketchInputGenerator.h?lines=40 . 

Certain tests in ExpressionFuzzer fail due to this , see diff : D78992756

Differential Revision: D80848847

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzer.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzer.cpp
@@ -451,6 +451,7 @@ bool ExpressionFuzzer::isSupportedSignature(
   if (usesTypeName(signature, "opaque") ||
       usesTypeName(signature, "interval day to second") ||
       usesTypeName(signature, "ipprefix") ||
+      usesTypeName(signature, "sfmsketch") ||
       (!options_.enableDecimalType && usesTypeName(signature, "decimal")) ||
       (!options_.enableComplexTypes && useComplexType) ||
       (options_.enableComplexTypes && usesTypeName(signature, "unknown"))) {


kagamiori:
Hi @kgpai, are there functions using the SFMSketch type? I wonder whether we should fix it instead of ignoring it in fuzzer.

kgpai:
@kagamiori There are functions using SFMSketchtype - they are all probablisitic and non determistic, and they should be fixed - however while we work on fixing this, we shouldnt let this block legitimate fuzzer signals by causing fuzzer to fail. 
Here is issue : https://github.com/facebookincubator/velox/issues/14643

velox/expression/fuzzer/ExpressionFuzzer.cpp
```diff
@@ -451,6 +451,7 @@ bool ExpressionFuzzer::isSupportedSignature(
   if (usesTypeName(signature, "opaque") ||
       usesTypeName(signature, "interval day to second") ||
       usesTypeName(signature, "ipprefix") ||
+      usesTypeName(signature, "sfmsketch") ||
```
kagamiori: nit: Could you add the github issue link as a comment for this skip so that we don't forget it?

kagamiori (approved):
LGTM. Thanks!

PR #14575, opened 2025-08-22
feat: Add fedora runs to CI

kgpai:
This PR adds a fedora build to our default set of linux builds. This should ensure that we continue to build and work on Fedora in the future.


assignUser:
I am currently pushing a new version of the image (as it would be after merging this PR) to make sure everything works as expected. Once that's up I'll push my changes to cause a build on it

assignUser:
https://github.com/facebookincubator/velox/actions/runs/17478225037/job/49643110427?pr=14575#step:7:5651

This error means the file is missing the fmt include, I assume previously it got it transitivly but fmt 11 moved things around

Issue #14469, opened 2025-08-14
Window should read the spilled data in batch to avoid performance problem of small window partitions


HolyLow:
### Description

We encountered a performance problem when using window spill. The window execution was extremely slow when unspilling compared with other operators with the same data volume and data distribution. So I profiled the cpu flamegraph and surprisingly found that the `HashAllocator::clear()` took about 50% of execution time.
<img width="5080" height="892" alt="Image" src="https://github.com/user-attachments/assets/22ae6503-27b2-4408-b54e-a08853987107" />
After digging into the code, it seems that window reads spilled data one window partition by another: 
https://github.com/facebookincubator/velox/blob/e971b986cb267eaa586151c7901fd8c48652f470/velox/exec/SortWindowBuild.cpp#L297-L337
For each window partition, the row container would be cleared before execution, which would call `HashAllocator::clear()`. If each window partition is extremely small (in our case only 1~2 rows per partition), the row container would be cleared per row, and the data initialization/release would be carried out per row! This behavior leads to poor performance in small window partition senario.
To avoid this bad case, window should read the spilled data in batch rather than in window partition granularity. For example, each reading pass would read at least 1k rows which contains at least 1 complete window partition.  The entire data batch would be consumed in memory, and the data initialization/release would only be required once per batch. In this way, the performance problem is solved, and the `HashAllocator::clear()` is no longer a hotspot in the flamegraph. **The E2E time improves from 2h12min to 31min in our case (4x speedup).**
I am willing to contribute the patch if the community is interested. Any suggestion is welcome. 
cc @mbasmanova @oerling @xiaoxmeng @duanmeng

amitkdutta:
CC: @aditi-pandit

xiaoxmeng:
@HolyLow thanks for the investigation. Looking forward the optimization and will help the review! cc @JkSelf @aditi-pandit @tanjialiang

JkSelf:
@HolyLow Thanks for the investigation.
Is this a common usage pattern across all callers of `RowContainer::clear()`? If so, would it be feasible to introduce a batched approach directly within the RowContainer itself, rather than implementing it separately in each caller?

HolyLow:
> [@HolyLow](https://github.com/HolyLow) Thanks for the investigation. Is this a common usage pattern across all callers of `RowContainer::clear()`? If so, would it be feasible to introduce a batched approach directly within the RowContainer itself, rather than implementing it separately in each caller?

[...]

PR #14585, opened 2025-08-25
perf: Read spilled Window data in batches to optimize for small window partitions

HolyLow:
Extend Window operator to read spilled data in batches of window partitions to improve performance in the presence of small partitions.
A new configuration setting window_spill_min_read_batch_rows with default value of 1'000 controls the minimum number of rows for a reading batch. Setting window_spill_min_read_batch_rows to 1 loads a single partition rather than a partition batch each time.

The preferred semantic would be to set a memory budget and load as many partitions that fit. This is not feasible at the moment because (1) estimating a single row's size is not efficient or accurate enough and might cause performance issues for variable-width data; (2) spilled data format doesn't include information about how many rows are present in any given window partition.

Fixes https://github.com/facebookincubator/velox/issues/14469

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -250,7 +250,10 @@ class QueryConfig {
   /// Window spilling flag, only applies if "spill_enabled" flag is set.
   static constexpr const char* kWindowSpillEnabled = "window_spill_enabled";
 
-  /// When Window spills, control the minimum number rows of a reading batch.
+  /// When processing spilled window data, read batches of whole partitions
+  /// having at least that many rows. Set to 1 to read one whole partition at a
+  /// time. Each driver processing the Window operator will process that much
+  /// data at once.
   static constexpr const char* kWindowSpillMinReadBatchRows =
       "window_spill_min_read_batch_rows";
 
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -331,6 +331,12 @@ Spilling
      - boolean
      - true
      - When `spill_enabled` is true, determines whether Window operator can spill to disk under memory pressure.
+   * - window_spill_min_read_batch_rows
+     - integer
+     - 1000
+     - When processing spilled window data, read batches of whole partitions having at least that many rows. Set to 1 to
+       read one whole partition at a time. Each driver processing the Window operator will process that much data at
+       once.
    * - row_number_spill_enabled
      - boolean
      - true
--- a/velox/exec/SortWindowBuild.cpp
+++ b/velox/exec/SortWindowBuild.cpp
@@ -365,13 +365,8 @@ void SortWindowBuild::loadNextPartitionBatchFromSpill() {
     numSpillReadBatches_--;
 
     auto lockedOpStats = opStats_->wlock();
-    auto pair = lockedOpStats->runtimeStats.try_emplace(
-        Window::kSpillWindowReadBatchNums, RuntimeMetric(numSpillReadBatches_));
-    if (!pair.second) {
-      lockedOpStats->runtimeStats[Window::kSpillWindowReadBatchNums].addValue(
-          numSpillReadBatches_);
-    }
-    numSpillReadBatches_ = 0;
+    lockedOpStats->runtimeStats[Window::kWindowSpillReadNumBatches] =
+        RuntimeMetric(numSpillReadBatches_);
   }
 }
 
--- a/velox/exec/SortWindowBuild.h
+++ b/velox/exec/SortWindowBuild.h
@@ -76,8 +76,9 @@ class SortWindowBuild : public WindowBuild {
   // Find the next partition start row from start.
   vector_size_t findNextPartitionStartRow(vector_size_t start);
 
-  // Reads next partition batch from spilled data into 'data_' and
-  // 'sortedRows_'.
+  // Load the next partition batch if needed. If current partition batch is not
+  // entirely consumed, return directly. Otherwise, read next partition batch
+  // from spilled data into 'data_' and set pointers in 'sortedRows_'.
   void loadNextPartitionBatchFromSpill();
 
   const size_t numPartitionKeys_;
@@ -126,7 +127,7 @@ class SortWindowBuild : public WindowBuild {
   // Used to sort-merge spilled data.
   std::unique_ptr<TreeOfLosers<SpillMergeStream>> merge_;
 
-  // Num of spill read partition batches.
+  // Number of batches of whole partitions read from spilled data.
   uint64_t numSpillReadBatches_ = 0;
 };
 } // namespace facebook::velox::exec
--- a/velox/exec/Window.h
+++ b/velox/exec/Window.h
@@ -67,9 +67,10 @@ class Window : public Operator {
   void reclaim(uint64_t targetBytes, memory::MemoryReclaimer::Stats& stats)
       override;
 
-  /// Window operator's read partition batch num.
-  static inline const std::string kSpillWindowReadBatchNums{
-      "spillWindowReadBatchNums"};
+  /// Runtime statistics holding total number of batches read from spilled data.
+  /// 0 if no spilling occurred.
+  static inline const std::string kWindowSpillReadNumBatches{
+      "windowSpillReadNumBatches"};
 
  private:
   // Used for k preceding/following frames. Index is the column index if k is a
--- a/velox/exec/tests/WindowTest.cpp
+++ b/velox/exec/tests/WindowTest.cpp
@@ -115,16 +115,19 @@ TEST_F(WindowTest, spill) {
   ASSERT_GT(stats.spilledPartitions, 0);
 }
 
-TEST_F(WindowTest, spillBatchRead) {
+TEST_F(WindowTest, spillBatchReadTinyPartitions) {
   const vector_size_t size = 1'000;
   const uint32_t minReadBatchRows = 100;
+  // Each tiny partition has 1 row.
+  const uint32_t partitionRows = 1;
   auto data = makeRowVector(
       {"d", "p", "s"},
       {
           // Payload.
           makeFlatVector<int64_t>(size, [](auto row) { return row; }),
           // Partition key.
-          makeFlatVector<int16_t>(size, [](auto row) { return row; }),
+          makeFlatVector<int16_t>(
+              size, [](auto row) { return row / partitionRows; }),
           // Sorting key.
           makeFlatVector<int32_t>(size, [](auto row) { return row; }),
       });
@@ -158,10 +161,64 @@ TEST_F(WindowTest, spillBatchRead) {
   ASSERT_GT(stats.spilledRows, 0);
   ASSERT_GT(stats.spilledFiles, 0);
   ASSERT_GT(stats.spilledPartitions, 0);
-  auto opStats = toOperatorStats(task->taskStats());
-  ASSERT_LE(
-      opStats.at("Window").runtimeStats[Window::kSpillWindowReadBatchNums].sum,
+  ASSERT_EQ(
+      stats.operatorStats.at("Window")
+          ->customStats[Window::kWindowSpillReadNumBatches]
+          .sum,
       size / minReadBatchRows);
+}
+
+TEST_F(WindowTest, spillBatchReadHugePartitions) {
+  const vector_size_t size = 1'000;
+  const uint32_t minReadBatchRows = 100;
+  // Each huge partition has 200 rows, which is larger than minReadBatchRows.
+  const uint32_t partitionRows = 200;
+  auto data = makeRowVector(
+      {"d", "p", "s"},
+      {
+          // Payload.
+          makeFlatVector<int64_t>(size, [](auto row) { return row; }),
+          // Partition key.
+          makeFlatVector<int16_t>(
+              size, [](auto row) { return row / partitionRows; }),
+          // Sorting key.
+          makeFlatVector<int32_t>(size, [](auto row) { return row; }),
+      });
+
+  createDuckDbTable({data});
+
+  core::PlanNodeId windowId;
+  auto plan = PlanBuilder()
+                  .values(split(data, 10))
+                  .window({"row_number() over (partition by p order by s)"})
+                  .capturePlanNodeId(windowId)
+                  .planNode();
+
+  auto spillDirectory = TempDirectoryPath::create();
+  TestScopedSpillInjection scopedSpillInjection(100);
+  auto task =
+      AssertQueryBuilder(plan, duckDbQueryRunner_)
+          .config(core::QueryConfig::kPreferredOutputBatchBytes, "1024")
+          .config(core::QueryConfig::kSpillEnabled, "true")
+          .config(core::QueryConfig::kWindowSpillEnabled, "true")
+          .config(
+              core::QueryConfig::kWindowSpillMinReadBatchRows, minReadBatchRows)
+          .spillDirectory(spillDirectory->getPath())
+          .assertResults(
+              "SELECT *, row_number() over (partition by p order by s) FROM tmp");
+
+  auto taskStats = exec::toPlanStats(task->taskStats());
+  const auto& stats = taskStats.at(windowId);
+
+  ASSERT_GT(stats.spilledBytes, 0);
+  ASSERT_GT(stats.spilledRows, 0);
+  ASSERT_GT(stats.spilledFiles, 0);
+  ASSERT_GT(stats.spilledPartitions, 0);
+  ASSERT_EQ(
+      stats.operatorStats.at("Window")
+          ->customStats[Window::kWindowSpillReadNumBatches]
+          .sum,
+      size / partitionRows);
 }
 
 TEST_F(WindowTest, spillUnsupported) {


HolyLow:
@aditi-pandit @xiaoxmeng Would you kindly help review this PR? Thanks a lot.

velox/exec/SortWindowBuild.cpp
```diff
@@ -295,13 +295,28 @@ void SortWindowBuild::noMoreInput() {
 }
 
 void SortWindowBuild::loadNextPartitionFromSpill() {
+  // Check if nextPartition is already loaded. If so, return directly.
+  if (currentPartition_ < static_cast<int>(partitionStartRows_.size() - 2)) {
+    return;
+  }
+
+  constexpr int kBatchSize = 1'000;
```
aditi-pandit: Make this configurable ?

velox/exec/SortWindowBuild.cpp
```diff
@@ -324,7 +339,10 @@ void SortWindowBuild::loadNextPartitionFromSpill() {
     }
 
     if (newPartition) {
-      break;
+      partitionStartRows_.push_back(sortedRows_.size());
```
aditi-pandit, line 348: Its possible that the last partition in sortedRows_ is very large, and you could overrun the 1000 row limit... Is there any other way to take care that we don't overrun ?

aditi-pandit (commented):
Thanks for the code. Have a high level question.

velox/exec/SortWindowBuild.cpp
HolyLow: This is a design decision. It is ok to make this detailed parameter configurable, but wouldn't it make the configurations too complex? Personally I don't think making everything configurable is a good design, especially for this kind of detail logic. Actually, in the spilling logic, batching is carried out almost everywhere, and none of them are configurable. In general 1k rows would be safe enough, and if in some corner cases 1k rows would saturate the memory and cause OOM, the OOM problem would more likely occur in the processing of merging files (as all the files should preload a complete vector in memory) rather than when processing WindowPartition. In this case, configuring the parameter here would not solve the problem.

But still, if you insist on making this parameter configurable, I could refactor the code. Looking forward to your thoughts, thanks a lot.

velox/exec/SortWindowBuild.cpp
HolyLow: You are right about overrunning the 1k row limit situation, but the 1k-row limit here doesn't serve as a upper bound limit. It is more like lower bound limit: we must load at least 1k rows. The upper bound limit would be `1k + sizeof(lastWindowPartition)`, and there is no way to limit the last partition's size. 
If a partition has more than 1k rows, we have to load them together in memory for later WindowPartition computation logic. Many window functions such as aggregation, n_percentile, lead/lag require that the entire partition is loaded, so we have to load the entire partition at once.

HolyLow:
@aditi-pandit Do you have any further comments? Thanks a lot.

HolyLow:
@aditi-pandit  @mbasmanova Could you please help review or  find someone proper to move this PR forward? There is no further progress for over a month. Thanks a lot.

HolyLow:
@mbasmanova This PR does not change the output data as this is an optimization rather than a bugfix, and existing WindowSpill related tests should cover the correctness of this change.

mbasmanova:
> This PR does not change the output data as this is an optimization rather than a bugfix, and existing WindowSpill related tests should cover the correctness of this change.

Got it. I wonder how can we protect this optimization from being compromised in the future. Can we add a test case that reproduces the original issue and check some runtime stat? Do we already have runtime stats that expose the issue? Can we add a check in the test against these stats to ensure we read in batches as opposed one partition at a time? If we do not have such stat, can we add it?

velox/exec/SortWindowBuild.cpp
```diff
@@ -295,13 +295,28 @@ void SortWindowBuild::noMoreInput() {
...
   sortedRows_.clear();
-  sortedRows_.shrink_to_fit();
+  sortedRows_.reserve(kBatchSize);
   data_->clear();
+  partitionStartRows_.clear();
+  partitionStartRows_.reserve(kBatchSize);
+  partitionStartRows_.push_back(0);
+  currentPartition_ = -1;
 
+  // Load at least kBatchSize rows and a complete partition. The rows might
```
mbasmanova: Would you update PR description to clarify this behavior? My reading of it suggests that "batch" is defined as 1K rows + the tail of the last partition.

velox/exec/SortWindowBuild.cpp
```diff
@@ -295,13 +295,28 @@ void SortWindowBuild::noMoreInput() {
 }
 
 void SortWindowBuild::loadNextPartitionFromSpill() {
```
mbasmanova: The existing comment is no longer correct. Please, update.

```
// Reads next partition from spilled data into 'data_' and 'sortedRows_'.
  void loadNextPartitionFromSpill();
```

velox/exec/SortWindowBuild.cpp
mbasmanova: > constexpr int kBatchSize = 1'000;

I agree with @aditi-pandit that it would be nice to make this configurable. The default value can be 1K, but having a way to tune this optimization (or even disable by setting the value to 1) is helpful as it allows to mitigate production issues without making a code change.

mbasmanova:
@HolyLow Thank you for the optimization. The change itself looks good to me. I'm just wondering if there is a way to protect this logic with a unit test that checks runtime stats.

velox/exec/SortWindowBuild.cpp
HolyLow: @aditi-pandit @mbasmanova Added a configuration as suggested.

velox/exec/SortWindowBuild.cpp
HolyLow: Refactored the param naming to clarify the meaning.

velox/exec/SortWindowBuild.cpp
HolyLow: Done.

velox/exec/SortWindowBuild.cpp
HolyLow: Done.

HolyLow:
@mbasmanova Thanks for your review, added the stats and UT as suggested. Any further suggestion is welcome, thanks a lot.

mbasmanova:
@HolyLow Thank you for iterating. Would you update PR description to describe all changes?

velox/common/base/SpillConfig.h
```diff
@@ -157,5 +158,8 @@ struct SpillConfig {
 
   /// Custom options passed to velox::FileSystem to create spill WriteFile.
   std::string fileCreateConfig;
+
+  /// The minimum #rows of a reading batch for window spilling.
```
mbasmanova: typos: The minimum number of rows to read when processing spilled window data.

@HolyLow 

I thought about this setting a bit. It is currently defined as the minimum number of rows to read. When set to 1'000, reading 900 row partitions leads to loading 1800 rows in memory. If a single partition barely fits, current implementation will allow the query to succeed, but the new implementation will lead to an OOM.

Hence, I suggest to change the semantic of this setting as follows.

Restore at a least one full partition from spill. Continue restoring full partitions until total number of rows reaches 1000. Do not load more than 1000 rows unless these all come from a single partition.

Also, I'm wondering if it would be better to set a limit in terms of bytes rather than number of rows.

What do you think?

HolyLow: @mbasmanova Typo fixed.

Changing the semantic would be nice in theory, and this semantic is what I want to do at first thought, but the implementation would encounter some problems. 
Firstly, there is no way to calculate the size of a partition before reading the entire partition into memory, so we are not able to decide whether to read next partition or not according to current row budget. 
Secondly, if we read an incomplete partition (e.g. first partition has 900 rows, and the second incomplete partition has 100 rows stored in memory and we do not read further within current batch), we must somehow transfer the incomplete partition to next batch. But this operation is not easy, or at least not easily efficient. The row data is stored in RowContainer, and transferring data between different batches's RowContainers might bring even more memory usage. The releaseRow interface of RowContainer might also cause more memory holes and increase memory usage, especially for non-fixed-size datatypes such as varchar.
So in general I am not sure how to efficiently (worst-case guaranteed) implement the proposed semantic. Maybe you could kindly provide some insights, thanks a lot.

As for limiting the batch by bytes, I think it totally makes sense, but the `estimateRowSize()` interface of RowContainer might not be accurate and efficient as needed, and might even cause performance issue if we estimate the byte size for each row. So again I am not sure if it is acceptable to estimateRowSize per row.

HolyLow:
@mbasmanova MR description updated, looking forward to your further suggestions. Thanks a lot.

velox/common/base/SpillConfig.h
mbasmanova: @HolyLow These are valid points. Would you update PR description to explain the preferred semantic (one partition or multiple partitions not exceeding memory budget) and clarify why this is not feasible at the moment? 

Curious, what kind of query shapes do you encounter in practice? Are these heavy on complex types and strings?

velox/core/QueryConfig.h
```diff
@@ -250,6 +250,10 @@ class QueryConfig {
   /// Window spilling flag, only applies if "spill_enabled" flag is set.
   static constexpr const char* kWindowSpillEnabled = "window_spill_enabled";
 
+  /// When Window spills, control the minimum number rows of a reading batch.
```
mbasmanova: Please, update documentation as well: https://facebookincubator.github.io/velox/configs.html

Let's clarify a bit.

/// When processing spilled window data, read batches of whole partitions having at least that many rows. Set to 1 to read one whole partition at a time. Each driver processing the Window operator will process that much data at once.

velox/exec/Window.h
```diff
@@ -67,6 +67,10 @@ class Window : public Operator {
   void reclaim(uint64_t targetBytes, memory::MemoryReclaimer::Stats& stats)
       override;
 
+  /// Window operator's read partition batch num.
```
mbasmanova: Let's clarify a bit.

/// Runtime statistics holding total number of batches read from spilled data. 0 if no spilling occurred.

Let's also rename for clarify and consistency.

- kWindowSpillReadNumBatches
- "windowSpillReadNumBatches"

velox/exec/SortWindowBuild.h
```diff
@@ -121,5 +125,8 @@ class SortWindowBuild : public WindowBuild {
 
   // Used to sort-merge spilled data.
   std::unique_ptr<TreeOfLosers<SpillMergeStream>> merge_;
+
+  // Num of spill read partition batches.
```
mbasmanova: // Number of batches of whole partitions read from spilled data.

velox/exec/SortWindowBuild.cpp
```diff
@@ -334,16 +358,24 @@ void SortWindowBuild::loadNextPartitionFromSpill() {
...
-        data_.get(), partition, inversedInputChannels_, sortKeyInfo_);
+  // No more partition batches. All data is consumed.
+  if (sortedRows_.empty()) {
+    partitionStartRows_.clear();
+    numSpillReadBatches_--;
+
+    auto lockedOpStats = opStats_->wlock();
+    auto pair = lockedOpStats->runtimeStats.try_emplace(
+        Window::kSpillWindowReadBatchNums, RuntimeMetric(numSpillReadBatches_));
+    if (!pair.second) {
```
mbasmanova: Why do we need this conditional logic? Aren't we setting these stats exactly once?

velox/exec/SortWindowBuild.cpp
```diff
@@ -334,16 +358,24 @@ void SortWindowBuild::loadNextPartitionFromSpill() {
...
+      lockedOpStats->runtimeStats[Window::kSpillWindowReadBatchNums].addValue(
+          numSpillReadBatches_);
+    }
+    numSpillReadBatches_ = 0;
```
mbasmanova: Why do we need to reset this value to 0?

velox/exec/SortWindowBuild.h
```diff
@@ -75,8 +76,9 @@ class SortWindowBuild : public WindowBuild {
   // Find the next partition start row from start.
   vector_size_t findNextPartitionStartRow(vector_size_t start);
 
-  // Reads next partition from spilled data into 'data_' and 'sortedRows_'.
-  void loadNextPartitionFromSpill();
+  // Reads next partition batch from spilled data into 'data_' and
```
mbasmanova: This description is not accurate. This method seems to be reading partitions in batches. Each call either reads a new batch of partitions and sets the "pointers" to the first partition or moves the "pointers" the next partition in current batch.

velox/exec/tests/WindowTest.cpp
```diff
@@ -115,6 +115,55 @@ TEST_F(WindowTest, spill) {
...
 
+TEST_F(WindowTest, spillBatchRead) {
+  const vector_size_t size = 1'000;
+  const uint32_t minReadBatchRows = 100;
+  auto data = makeRowVector(
+      {"d", "p", "s"},
+      {
+          // Payload.
+          makeFlatVector<int64_t>(size, [](auto row) { return row; }),
+          // Partition key.
```
mbasmanova, line 128: Just to clarify, are you testing single-row partitions here? Should we also test multi-row partitions that span across the 1K row boundaries?

velox/exec/tests/WindowTest.cpp
```diff
@@ -115,6 +115,55 @@ TEST_F(WindowTest, spill) {
...
+              "SELECT *, row_number() over (partition by p order by s) FROM tmp");
+
+  auto taskStats = exec::toPlanStats(task->taskStats());
+  const auto& stats = taskStats.at(windowId);
+
+  ASSERT_GT(stats.spilledBytes, 0);
+  ASSERT_GT(stats.spilledRows, 0);
+  ASSERT_GT(stats.spilledFiles, 0);
+  ASSERT_GT(stats.spilledPartitions, 0);
+  auto opStats = toOperatorStats(task->taskStats());
```
mbasmanova: Can we use stats.operatorStats.at("Window").customStats instead of calling toOperatorStats?

velox/core/QueryConfig.h
HolyLow: Updated the configs.rst file as suggested. Refactor the annotation as suggested.

velox/exec/Window.h
HolyLow: Done.

velox/exec/SortWindowBuild.h
HolyLow: Done.

velox/exec/SortWindowBuild.cpp
HolyLow: I was thinking that the opStats might be used across different pipelines, but it seems not to be the case. Removed the logic as suggested.

velox/exec/SortWindowBuild.cpp
HolyLow: I was writing this as some kind of protecting logic, in case this function is called multiple times incorrectly. Removed this as suggested.

velox/exec/SortWindowBuild.h
HolyLow: This function would only read the new partition batch data and update the pointers if needed, but would not move the pointers to the next partition of the current batch. The pointer moving operation is carried out elsewhere. But I got what you mean, and the annotation is updated.

velox/exec/tests/WindowTest.cpp
HolyLow: Added another UT with huge partitions.

velox/exec/tests/WindowTest.cpp
HolyLow: Done.

mbasmanova (approved):
Thanks.

velox/common/base/SpillConfig.h
HolyLow: PR updated as suggested, thanks a lot.

The query shapes we encountered typically have multiple varchar columns, which makes it hard to estimate size precisely.

HolyLow:
@mbasmanova MR description is updated, and code is refactored as suggested. Your detailed comments really make this PR much better, and I've learned a lot. Looking forward to your further suggestions with great appreciation.

velox/docs/configs.rst
```diff
@@ -335,6 +331,12 @@ Spilling
      - boolean
      - true
      - When `spill_enabled` is true, determines whether Window operator can spill to disk under memory pressure.
+   * - window_spill_min_read_batch_rows
+     - integer
+     - 1000
+     - When processing spilled window data, read batches of whole partitions having at least that many rows. Set to 1 to
```
aditi-pandit, line 337: @HolyLow : I had been following this review, but saw the final commit today. 

This is a bit misleading... Are you saying that we read either one full partition > 1000 rows or as many partitions that fit in 1000 rows. Though it seems like we could have cases where we have x partitions taking 999 rows but then the last partition read for the last row could have > 1000 rows so we eventually read 2000 rows  maybe ?

Also why are you saying set to 1 ? Didn't follow that part.

@mbasmanova

mbasmanova: @aditi-pandit Aditi, you are correct. We read one whole partition + n more partitions until we read at least 1000 rows. We may read more rows than 1000. 

> Set to 1 ...

This is advice to the user. They can set this config to 1 to ensure no more than 1 whole partition is read at a time. This would be the original behavior.

HolyLow: @aditi-pandit Yes, you are right about the > 1000 rows case. That's the reason we said it "having at least that many rows". Setting to 1 would actually disable the batching logic and fallback to the original read-one-partition-per-calling behavior. We explicitly add this note here in case users suffer from some bad cases and want to fallback to original behavior.

aditi-pandit: Thanks for the explanation. Makes sense. Misinterpret the "Set to 1". Appreciate it.

PR #14587, opened 2025-08-25
feat: Add page load time metric

rui-mo:
The time spent in page load is useful for the table scan profiling, and this PR 
adds the page load time metric in runtime stats.

### DIFF
--- a/velox/dwio/common/Statistics.h
+++ b/velox/dwio/common/Statistics.h
@@ -533,12 +533,34 @@ class Statistics {
 };
 
 struct ColumnReaderStatistics {
+  void clear() {
+    pageScanTime.store(0, std::memory_order_relaxed);
+    flattenStringDictionaryValues = 0;
+  }
+
+  void incPageScanTime(uint64_t v) {
+    pageScanTime.fetch_add(v, std::memory_order_relaxed);
+  }
+
   // Number of rows returned by string dictionary reader that is flattened
   // instead of keeping dictionary encoding.
   int64_t flattenStringDictionaryValues{0};
+
+  std::atomic<uint64_t> pageScanTime{0};
 };
 
 struct RuntimeStatistics {
+  void clear() {
+    skippedSplits = 0;
+    processedSplits = 0;
+    skippedSplitBytes = 0;
+    skippedStrides = 0;
+    processedStrides = 0;
+    footerBufferOverread = 0;
+    numStripes = 0;
+    columnReaderStatistics.clear();
+  }
+
   // Number of splits skipped based on statistics.
   int64_t skippedSplits{0};
 
@@ -591,6 +613,10 @@ struct RuntimeStatistics {
       result.emplace(
           "flattenStringDictionaryValues",
           RuntimeCounter(columnReaderStatistics.flattenStringDictionaryValues));
+    }
+    if (columnReaderStatistics.pageScanTime > 0) {
+      result.emplace(
+          "pageScanTime", RuntimeCounter(columnReaderStatistics.pageScanTime));
     }
     return result;
   }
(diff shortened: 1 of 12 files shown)


velox/dwio/common/Statistics.h
```diff
@@ -533,12 +533,35 @@ class Statistics {
...
+  void incPageLoadTime(uint64_t v) {
+    pageLoadTime.fetch_add(v, std::memory_order_relaxed);
+  }
+
   // Number of rows returned by string dictionary reader that is flattened
   // instead of keeping dictionary encoding.
   int64_t flattenStringDictionaryValues{0};
+
+  // Total time spent in loading pages, in nanoseconds.
+  std::atomic<uint64_t> pageLoadTime{0};
```
Yuhta: We don't need this to be atomic, as readers are not supposed to be shared between threads.  If we remove the atomic, we can also avoid the `clear()` method and just use `= {}` to reset these POD structs.

rui-mo: Thanks for this suggestion! I updated this PR and removed the  atomic. Would you please take another look?

FelixYBW:
@Yuhta any more comments? The metric is useful for GLuten customer workload analysis. The metric counts all the remote storage fetch time.

@pedroerp FYI

FelixYBW:
@pedroerp @Yuhta Can you take a look of the PR? No functional change, just add a metric.

velox/dwio/common/Statistics.h
```diff
@@ -537,6 +537,9 @@ struct ColumnReaderStatistics {
   // Number of rows returned by string dictionary reader that is flattened
   // instead of keeping dictionary encoding.
   int64_t flattenStringDictionaryValues{0};
+
+  // Total time spent in loading pages, in nanoseconds.
+  uint64_t pageLoadTime{0};
```
pedroerp: nit: it's usually a good practice to explicit about the granularity of the metric: `pageLoadTimeNs`

rui-mo: Updated, thanks.

rui-mo:
@majetideepak Would you please take a look when you get a chance? Thanks.

FelixYBW:
@pedroerp Can you merge the PR?

velox/dwio/common/Statistics.h
```diff
@@ -597,6 +600,11 @@ struct RuntimeStatistics {
           "flattenStringDictionaryValues",
           RuntimeCounter(columnReaderStatistics.flattenStringDictionaryValues));
     }
+    if (columnReaderStatistics.pageLoadTimeNs > 0) {
+      result.emplace(
+          "pageLoadTimeNs",
+          RuntimeCounter(columnReaderStatistics.pageLoadTimeNs));
```
kKPulla: I see ``result`` here is a map with value type as ``RuntimeMetric`` but we are adding ``RuntimeCounter`` here. Can you fix that? (This matches with a bunch of failure signals I see in CI internally too)

rui-mo: > std::unordered_map<std::string, RuntimeCounter> result;

@kKPulla I noticed that the value type of result is `RuntimeCounter`. Did I overlook something? Please let me know if I’m missing anything.

kKPulla: Do you mean in L569 https://github.com/facebookincubator/velox/blob/main/velox/dwio/common/Statistics.h#L566 ? Isn't it

> std::unordered_map<std::string, RuntimeMetric> result;

rui-mo: Sorry, I just noticed this had been changed via https://github.com/facebookincubator/velox/commit/c623eaddcc1f8544c8d5a229c2f142b6669b6d0e recently. Updated this PR to use `RuntimeMetric`, and thanks for the pointer!

FelixYBW:
Thank you, @kKPulla

PR #14589, opened 2025-08-25
feat: Add helper VELOX_USER_RETURN_XXX

jinchengchenghh:


jinchengchenghh:
Could you help review this PR? Thanks! @mbasmanova 
A question is do I need to update existing code like `VELOX_RETURN_IF(
        entryDelimiter.empty(), Status::UserError("entryDelimiter is empty"));`?

velox/common/base/Status.h
```diff
@@ -537,6 +605,41 @@ inline Status genericToStatus(Status&& st) {
   return std::move(st);
 }
 
+struct CompileTimeEmptyString {
```
mbasmanova: Is this a copy-paste from somewhere else? Would it be possible to refactor to avoid duplicating code?

mbasmanova (commented):
@jinchengchenghh Thank you for adding these macros. Would you update PR description to explain what macros are being added and how they are supposed to be used? Would you also clarify how they are similar/different than existing RETURN_IF macro?

mbasmanova:
> A question is do I need to update existing code like VELOX_RETURN_IF( entryDelimiter.empty(), Status::UserError("entryDelimiter is empty"));?

How many places use this macro? Is VELOX_RETURN_IF now being replaced with VELOX_USER_RETURN ? It would be nice to replace VELOX_RETURN_IF, but, perhaps, in a separate PR.

jinchengchenghh:
Occured 5 times in SplitToMap.h, will be replaced by VELOX_USER_RETURN or VELOX_USER_RETURN_EQ, I will create a new PR to do the refactor.

mbasmanova (approved):
Thanks!

Issue #14590, opened 2025-08-25
High overhead in json_extract() due to redundant calls to simdjsonParse()


xiaodouchen:
### Description

We've encountered a performance issue with `json_extract()` function, especially when processing highly complex JSON objects. By profiling our application, we captured a CPU flame graph that clearly shows `simdjsonParse()` is responsible for ~69% of the total CPU time. 
<img width="4560" height="1216" alt="Image" src="https://github.com/user-attachments/assets/1a3a24cf-081d-46ca-8e1d-e6326e9f9eca" />
After analyzing the code, we found that `simdjsonParse()` is executed 3 times within the json_extract implementation. This redundant parsing is causing a performance overhead. The relevant code is as follows:

https://github.com/facebookincubator/velox/blob/8d01456cf77a56d56c371ecc9509c5ae111157d8/velox/functions/prestosql/JsonFunctions.cpp#L750-L757 

https://github.com/facebookincubator/velox/blob/8d01456cf77a56d56c371ecc9509c5ae111157d8/velox/functions/prestosql/json/SIMDJsonExtractor.h#L124-L147

Refactoring this to parse the JSON only once and reuse the result should resolve the issue.

mbasmanova:
CC: @kgpai @amitkdutta

PR #14592, opened 2025-08-25
fix: Remove redundant `simdjsonParse()` in `JsonExtractImpl`

xiaodouchen:
Summary:
Remove redundant `simdjsonParse()` in `JsonExtractImpl`. Parse the JSON 
only once and reuse the result to avoid significant overhead.

This PR fixes https://github.com/facebookincubator/velox/issues/14590.


mbasmanova (approved):
Thanks.

velox/functions/prestosql/JsonFunctions.cpp
```diff
@@ -747,16 +747,17 @@ struct JsonExtractImpl {
...
 
     // Check for valid json
-    {
-      SIMDJSON_ASSIGN_OR_RAISE(auto jsonDoc, simdjsonParse(paddedJson));
-      simdjson::ondemand::document parsedDoc;
-      if (auto errorCode = simdjsonParse(paddedJson).get(parsedDoc);
-          errorCode != simdjson::SUCCESS) {
-        return errorCode;
-      }
+    auto jsonDoc = simdjsonParse(paddedJson);
```
kgpai: Thanks @xiaodouchen , but simdjsonParse isnt fully reliable when it comes to validating well formed json's. We recently moved to jsonParsingError in this code. Can you rebase and use that here ?

xiaodouchen: @kgpai Thanks for the suggestion. I've updated the code to call `jsonParsingError()` as you recommended.

However, I noticed that `jsonParsingError()` also contains its own call to `simdjsonParse()`. The main goal of this PR is to reuse the parsed JSON document and avoid redundant parsing. To achieve this, I've added a new overload for `jsonParsingError()` that accepts `simdjson::ondemand::document` as input.

This change ensures that `simdjsonParse()` is called only once across the entire process, eliminating the overhead of repeated parsing.

xiaodouchen:
@kgpai I noticed a number of other functions (IsJsonScalarFunction, JsonArrayContainsFunction, JsonArrayGetFunction, JsonExtractScalarFunction, JsonArrayLengthFunction) also have duplicate `simdjsonParse()` calls (once in the function body and again in `jsonParsingError(const simdjson::padded_string& json)`). This also leads to unnecessary parsing overhead.

Do you think this optimization should also be included in this PR to reduce parsing overhead, or would you prefer it in a separate one?

kgpai:
Thanks @xiaodouchen I think we can make it as a part of this change.

kevinwilfong:
I removed the ready-to-merge tag in light of the recent conversation, feel free to add it back once it's resolved

xiaodouchen:
> Thanks @xiaodouchen I think we can make it as a part of this change.

@kgpai @kevinwilfong  I've updated the PR with the changes. Ready for another review, thanks!

velox/functions/prestosql/JsonFunctions.cpp
```diff
@@ -29,6 +29,60 @@ namespace facebook::velox::functions {
...
+} // namespace
+
+simdjson::error_code jsonParsingError(simdjson::ondemand::document& doc) {
+  SIMDJSON_TRY(validate<simdjson::ondemand::document&>(doc));
+  if (!doc.at_end()) {
+    return simdjson::TRAILING_CONTENT;
+  }
+  // Rewind the document to go through a document more than once without
+  // rescanning all of the input data again.
+  doc.rewind();
```
kgpai, line 75: Thank you for this change, For my curiosity I was wondering if you might be able to find out how much difference this makes compared to calling parse again ? (Some simple local benchmark to get an idea would be very helpful).

xiaodouchen:
> Thanks so much @xiaodouchen , this is great. Would it be possible for you to add a small benchmark in JsonExprBenchmark.cpp so we can get an idea of how much this change saves us (For.eg have one benchmark that calls jsonParsingError with paddedString and the other with document ?).

@kgpai  Sure! I've added the benchmark in JsonExprBenchmark.cpp as you suggested.

I used IsJsonScalarFunction as the basis for these benchmarks due to its relatively simple logic. This gives us two test cases to compare:
* IsJsonScalarWithParseAgain: This simulates the old behavior, calling jsonParsingError with paddedString and performing an additional parse.
* IsJsonScalarWithoutParseAgain: This represents the new approach, calling jsonParsingError with the existing document.

I noticed that the JSON data constructed in the existing JsonExprBenchmark was quite simple. To get a more realistic performance comparison, I've run the benchmarks against two sets of data (env: CentOS, llvm):
* With the original simple JSON data, the results are as follows:
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
IsJsonScalarWithParseAgain(100_iters_10bytes_si            31.01us    32.25K
IsJsonScalarWithoutParseAgain(100_iters_10bytes 129.72%    23.90us    41.84K
IsJsonScalarWithParseAgain(100_iters_100bytes_s            85.70us    11.67K
IsJsonScalarWithoutParseAgain(100_iters_100byte 113.55%    75.47us    13.25K
IsJsonScalarWithParseAgain(100_iters_1000bytes_           618.14us     1.62K
IsJsonScalarWithoutParseAgain(100_iters_1000byt 111.05%   556.65us     1.80K
IsJsonScalarWithParseAgain(100_iters_10000bytes             6.26ms    159.84
IsJsonScalarWithoutParseAgain(100_iters_10000by 113.58%     5.51ms    181.55
```

* With a more complex and realistic JSON from simdjson's benchmark suite, which I've also added: https://github.com/simdjson/simdjson/blob/master/jsonexamples/twitter.json. The performance results are as follows:
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
IsJsonScalarWithParseAgain(100_iters_600KB_size            84.98ms     11.77
IsJsonScalarWithoutParseAgain(100_iters_600KB_s 125.84%    67.53ms     14.81
```

velox/functions/prestosql/JsonFunctions.cpp
xiaodouchen: I understand that the benchmark mentioned here is the same one discussed in the comment above. Therefore, I've added a single new benchmark case to JsonExprBenchmark.cpp to demonstrate the optimization's effect.

Beyond that, according to the description in simdjson's rewind() PR (https://github.com/simdjson/simdjson/pull/1539), rewind() is a good choice.

```
Performance note: the On Demand front-end does not materialize the parsed numbers and other values. 
If you are accessing everything twice, you may need to parse them twice. Thus the rewind functionality 
is best suited for cases where the first pass only scans the structure of the document.
```

velox/functions/prestosql/benchmarks/JsonExprBenchmark.cpp
```diff
@@ -177,6 +177,34 @@ struct FollyJsonSizeFunction {
   }
 };
 
+template <typename T>
+struct IsJsonScalarFunctionWithParseAgain {
```
kgpai: Hi @xiaodouchen one more minor request, I feel comparion with is_json_scalar doesnt give true scale of improvment since most jsons are deeply nested and much larger . 
Can you change this function so that it takes a template parameter with it decides to parse or call rewind . We can also then pass it larger jsons and get a much better idea of the performance (which i suspect will be better than the 25% we see now)
Thank you again !

xiaodouchen: Hi, @kgpai  thanks for the feedback!

Based on your comments, I've tried to update the benchmarks. Could you please take a look and let me know if this aligns with what you had in mind?

I've added two sets of comparison tests to better illustrate the performance impact:

#### Group 1: Comparing JsonParsingErrorWithPaddedString vs. JsonParsingErrorWithDocument

Results with simple JSON:
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
JsonParsingErrorWithPaddedString(100_iters_10by            30.91us    32.36K
JsonParsingErrorWithDocument(100_iters_10bytes_ 128.23%    24.10us    41.49K
JsonParsingErrorWithPaddedString(100_iters_100b            85.39us    11.71K
JsonParsingErrorWithDocument(100_iters_100bytes 115.37%    74.01us    13.51K
JsonParsingErrorWithPaddedString(100_iters_1000           608.79us     1.64K
JsonParsingErrorWithDocument(100_iters_1000byte 110.48%   551.04us     1.81K
JsonParsingErrorWithPaddedString(100_iters_1000             5.95ms    168.14
JsonParsingErrorWithDocument(100_iters_10000byt 110.29%     5.39ms    185.45
```

Results with complex JSON (twitter.json):
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
JsonParsingErrorWithPaddedString(100_iters_600K            84.63ms     11.82
JsonParsingErrorWithDocument(100_iters_600KB_si 124.83%    67.80ms     14.75
```

#### Group 2: Comparing SimdjsonParseWithParseAgain vs. SimdjsonParseWithRewind

This group measures the overall improvement, including avoiding the re-parse.

Results with simple JSON:
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
SimdjsonParseWithParseAgain(100_iters_10bytes_s            16.49us    60.66K
SimdjsonParseWithRewind(100_iters_10bytes_size) 164.29%    10.04us    99.65K
SimdjsonParseWithParseAgain(100_iters_100bytes_            27.48us    36.40K
SimdjsonParseWithRewind(100_iters_100bytes_size 176.30%    15.58us    64.17K
SimdjsonParseWithParseAgain(100_iters_1000bytes           134.71us     7.42K
SimdjsonParseWithRewind(100_iters_1000bytes_siz 184.75%    72.91us    13.72K
SimdjsonParseWithParseAgain(100_iters_10000byte             1.15ms    870.36
SimdjsonParseWithRewind(100_iters_10000bytes_si 189.72%   605.61us     1.65K
```
Results with complex JSON (twitter.json):
```
[...]osql/benchmarks/JsonExprBenchmark.cpp     relative  time/iter   iters/s
SimdjsonParseWithParseAgain(100_iters_600KB_siz            40.41ms     24.75
SimdjsonParseWithRewind(100_iters_600KB_size)   175.55%    23.02ms     43.44
```

#### Additional Observation:
Additionally, it seems the primary difference between these two groups is the extra call to `validate()` in `jsonParsingError()`, which appears to be very costly.

Let me know your thoughts, thanks!

kgpai: Thanks this is very informative - jsonParsingError is required though because it validates the full correctness of the json since just using simdjsonparse isnt reliable.

xiaodouchen:
@kgpai @kevinwilfong @mbasmanova 
Great to see the `ready-to-merge` label. Could you please merge this when you get a chance? Thanks!

zhouyuan:
Cc: @PHILO-HE

PR #14118, opened 2025-07-16
feat: Support Compression for Text Writer

YiChengLee03:
Summary: Support ZSTD, Gzip and  raw deflate compression

Differential Revision: D77546411

PR #14593, opened 2025-08-25
feat: Add gzip compression support

xin-zhang2:
Add gzip support for ZlibCompressor.

### DIFF
--- a/velox/dwio/common/compression/Compression.cpp
+++ b/velox/dwio/common/compression/Compression.cpp
@@ -35,6 +35,8 @@ using memory::MemoryPool;
 
 namespace {
 
+constexpr int kGzipCodec = 16;
+
 class ZstdCompressor : public Compressor {
  public:
   explicit ZstdCompressor(int32_t level) : Compressor{level} {}
@@ -57,7 +59,7 @@ ZstdCompressor::compress(const void* src, void* dest, uint64_t length) {
 
 class ZlibCompressor : public Compressor {
  public:
-  explicit ZlibCompressor(int32_t level);
+  explicit ZlibCompressor(int32_t level, int32_t windowBits, bool isGzip);
 
   ~ZlibCompressor() override;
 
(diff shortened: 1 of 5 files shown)


velox/dwio/common/compression/Compression.cpp
```diff
@@ -68,13 +68,19 @@ class ZlibCompressor : public Compressor {
...
   stream_.zalloc = Z_NULL;
   stream_.zfree = Z_NULL;
   stream_.opaque = Z_NULL;
   DWIO_ENSURE_EQ(
-      deflateInit2(&stream_, level_, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY),
+      deflateInit2(
+          &stream_,
+          level_,
+          Z_DEFLATED,
+          (isGzip ? 31 : -15),
```
zacw7: can we create named constants for these magic numbers?

velox/dwio/common/compression/Compression.cpp
```diff
@@ -150,9 +156,9 @@ ZlibDecompressor::ZlibDecompressor(
   zstream_.next_out = Z_NULL;
   zstream_.avail_out = folly::to<uInt>(blockSize);
   int zlibWindowBits = windowBits;
-  constexpr int GZIP_DETECT_CODE = 32;
   if (isGzip) {
-    zlibWindowBits = zlibWindowBits | GZIP_DETECT_CODE;
+    zlibWindowBits =
+        (zlibWindowBits < 0 ? -zlibWindowBits : zlibWindowBits) | 16;
```
zacw7: ditto

velox/dwio/common/compression/Compression.cpp
xin-zhang2: Thanks for the review.
15 is the default value of windowBits. A negative value indicates raw deflate, and adding 16 enables gzip wrapper.
I've updated the code to use the windowsBits in the compression option instead of -15, which I think is more reasonable, and also defined 16 as a constant named kGzipCodec.

sdruzkin:
Why do we need gzip when zlib does the same? Or is it to make it explicitly when it's already supported implicitly?

xin-zhang2:
@sdruzkin Thanks for the review. 
The current ZlibCompressor only performs raw deflate compression, and this PR adds the support for gzip. 
The initial motivation came from https://github.com/facebookincubator/velox/pull/14118, which was closed due to the lack of gzip support in createCompressor. I opened this PR to add gzip support so that we can move forward with TextWriter compression.

kewang1024:
Thanks @xin-zhang2 for working on it! Is our next step to add the support in text writer code once this is merged?

xin-zhang2:
@kewang1024 
Yes. Is @YiChengLee03 still working on this? If not, I can continue with it.

zacw7:
> @kewang1024 Yes. Is @YiChengLee03 still working on this? If not, I can continue with it.

No. Would be great if you can take it over. thx

kewang1024 (approved):
Thank you @xin-zhang2

PR #14598, opened 2025-08-26
feat(abfs): Support multi register AbfsFileSystem instance

### LINKED ISSUE #14580: Add support for multiple ABFS file system instances
### Description

Enable support for multiple ABFS file system instances, consistent with other supported storage types.

JkSelf:
Fix https://github.com/facebookincubator/velox/issues/14580

### DIFF
--- a/velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
@@ -30,16 +30,47 @@ namespace facebook::velox::filesystems {
 
 #ifdef VELOX_ENABLE_ABFS
 
-folly::once_flag abfsInitiationFlag;
+using FileSystemMap = folly::Synchronized<
+    std::unordered_map<std::string, std::shared_ptr<FileSystem>>>;
+
+FileSystemMap& fileSystems() {
+  static FileSystemMap instances;
+  return instances;
+}
 
 std::shared_ptr<FileSystem> abfsFileSystemGenerator(
     std::shared_ptr<const config::ConfigBase> properties,
     std::string_view filePath) {
-  static std::shared_ptr<FileSystem> filesystem;
-  folly::call_once(abfsInitiationFlag, [&properties]() {
-    filesystem = std::make_shared<AbfsFileSystem>(properties);
-  });
-  return filesystem;
+  auto cacheKeys = extractCacheKeyFromConfig(*properties);
+  // If the authType is not provided, use the accountName as cacheKey.
+  std::string cacheKey = cacheKeys.empty()
+      ? AbfsPath(filePath).accountName()
+      : cacheKeys[0].accountName + cacheKeys[0].authType;
+
+  // Check if an instance exists with a read lock (shared).
+  auto fs = fileSystems().withRLock(
+      [&](auto& instanceMap) -> std::shared_ptr<FileSystem> {
+        auto iterator = instanceMap.find(cacheKey);
+        if (iterator != instanceMap.end()) {
+          return iterator->second;
+        }
+        return nullptr;
+      });
+  if (fs != nullptr) {
+    return fs;
+  }
+
+  return fileSystems().withWLock(
+      [&](auto& instanceMap) -> std::shared_ptr<FileSystem> {
+        // Repeat the checks with a write lock.
+        auto iterator = instanceMap.find(cacheKey);
+        if (iterator != instanceMap.end()) {
+          return iterator->second;
+        }
+        auto fs = std::make_shared<AbfsFileSystem>(properties);
+        instanceMap.insert({cacheKey, fs});
+        return fs;
+      });
 }
 
 std::unique_ptr<velox::dwio::common::FileSink> abfsWriteFileSinkGenerator(


JkSelf:

majetideepak (changes requested):
@JkSelf We create a client for each ReadFile and WriteFile for ABFS. We don't need this change.

JkSelf:
@majetideepak Got it. Thanks for your pointing.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -19,21 +19,72 @@
 #include "velox/connectors/hive/storage_adapters/abfs/AbfsFileSystem.h" // @manual
 #include "velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h" // @manual
 #include "velox/dwio/common/FileSink.h"
+
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsConfig.h"
```
czentgr: Nit: to be consistent lets move it up and mark it as manual to keep it in the correct order.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -19,21 +19,72 @@
...
+
+        auto cacheKey = fmt::format(
+            "{}-{}",
+            abfsConfig.fileSystem(),
+            properties->get<std::string>(
+                "fs.azure.blob-endpoint", abfsConfig.accountNameWithSuffix()));
+
+        // Check if an instance exists with a read lock (shared).
+        auto fs = gcsFileSystems().withRLock(
+            [&](auto& instanceMap) -> std::shared_ptr<FileSystem> {
```
czentgr: const& for the instanceMap.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -19,21 +19,72 @@
...
+    FileSystem>(std::shared_ptr<const config::ConfigBase>, std::string_view)>
+abfsFileSystemGenerator() {
+  static auto filesystemGenerator =
+      [](std::shared_ptr<const config::ConfigBase> properties,
+         std::string_view filePath) {
+        auto abfsConfig = AbfsConfig(filePath, *properties);
+
+        auto cacheKey = fmt::format(
+            "{}-{}",
+            abfsConfig.fileSystem(),
```
czentgr: Did we discuss using a different cache key, maybe like account-authtype?

JkSelf: @czentgr Yes. Will update the cacheKey as `accountName-authType` after [PR#14930 ](https://github.com/facebookincubator/velox/pull/14930) merged.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -14,32 +14,77 @@
...
+            [&](auto& instanceMap) -> std::shared_ptr<FileSystem> {
+              // Repeat the checks with a write lock.
+              auto iterator = instanceMap.find(cacheKey);
+              if (iterator != instanceMap.end()) {
+                return iterator->second;
+              }
+
+              std::shared_ptr<AbfsFileSystem> fs;
+              if (properties != nullptr) {
+                fs = std::make_shared<AbfsFileSystem>(properties);
```
czentgr: This requires us to the have the ClientProviderFactory for the different auth types registered?

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -30,16 +30,47 @@ namespace facebook::velox::filesystems {
...
-  static std::shared_ptr<FileSystem> filesystem;
-  folly::call_once(abfsInitiationFlag, [&properties]() {
-    filesystem = std::make_shared<AbfsFileSystem>(properties);
-  });
-  return filesystem;
+  auto cacheKeys = extractCacheKeyFromConfig(*properties);
+  // If the authType is not provided, use the accountName as cacheKey.
+  std::string cacheKey = cacheKeys.empty()
+      ? AbfsPath(filePath).accountName()
+      : cacheKeys[0].accountName + cacheKeys[0].authType;
```
JkSelf, line 48: @majetideepak @czentgr Do we need to handle the multi CacheKeys case here?

denodo-research-labs:
Hi, Do you have a rough estimate on when this might be ready? Thanks!

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
czentgr: Yes. We know that if the type is given then account+type is unique and we can reuse the same client across multiple catalogs (using the same account-type).
I suppose the main problem is when no type is specified.

We also have multiple AzureClientProviders that are also unique with account+type. This change doesn't address this and is likely orthogonal but we'd like auto-registration of the `AzureClientProvider`.

Per properties instance you could have multiple account/type entries (according to the doc and which one is ultimately used depends in the ABFS path). But also we can have multiple catalog definitions that result in multiple file systems that need to be created.

Let's hope this makes some sense. It is somewhat convoluted. Maybe we need to have examples of the configuration and what objects get created and how they are known.

PR #14601, opened 2025-08-26
feat(type): Add nullable getters for nested arrays and maps in Variant

### LINKED ISSUE #14597: Add nullableArrayOfArrays and nullableMapOfArrays in the Variant class
> if we should add nullableArrayOfArrays<T>() and nullableMapOfArrays<K,V>() as well which supports null ? I
> 
> @furmak331 Thank would be nice. Let's put that in a separate PR though. Thanks. 

 _Originally posted by @mbasmanova in [#14577](https://github.com/facebookincubator/velox/issues/14577#issuecomment-3218887536)_

furmak331:
This PR enhances the variant class with new getters to support nullable nested array and map types, including:

`nullableArrayOfArrays<T>()`: Returns a std::vector<std::optional<std::vector<std::optional<T>>>> for arrays of arrays with possible nulls at both levels.
`nullableMapOfArrays<K, V>()`: Returns a std::map<K, std::optional<std::vector<std::optional<V>>>> for maps with array values, supporting nulls for both the map value and array elements.

Fixes #14597

### DIFF
--- a/velox/type/Variant.h
+++ b/velox/type/Variant.h
@@ -574,6 +574,25 @@ class Variant {
     return values;
   }
 
+  /// Returns a std::map of primitive keys to optional std::vector of optional
+  /// primitive values. Null values are returned as unset optional.
+  template <typename K, typename V>
+  std::map<K, std::optional<std::vector<std::optional<V>>>>
+  nullableMapOfArrays() const {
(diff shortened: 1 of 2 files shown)


mbasmanova (approved):
Thanks.

PR #14602, opened 2025-08-26
refactor: Connector::supportsSplitPreload should be const

MBkkt:


kevinwilfong:
Hi @MBkkt it looks like the Axiom project https://github.com/facebookexperimental/verax has an implementation of a Connector, and so we need to do this in a way that we can make the change here in a backwards compatible way, update Axiom, and then come back here to Velox and clean up the backwards compatibility.

Can you change your code to be
  
```
#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
  virtual bool supportsSplitPreload() {
    return false;
  }
#else
  virtual bool supportsSplitPreload() const {
    return false;
  }
#endif
```

Then on the Verax side we can make sure they enable VELOX_ENABLE_BACKWARD_COMPATIBILITY until they fix their Connector and then we can clean up this code 

This PR has an example https://github.com/facebookincubator/velox/pull/9680

MBkkt:
@kevinwilfong fixed

MBkkt:
@kevinwilfong Hi, I see some CI failed but it's internal to fb, can you clarify next steps please?

kevinwilfong:
@MBkkt sorry that's on me, I'll try to get that addressed today

velox/connectors/Connector.h
```diff
@@ -590,9 +590,15 @@ class Connector {
   /// ConnectorSplit in addSplit(). If so, TableScan can preload splits
   /// so that file opening and metadata operations are off the Driver'
   /// thread.
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
```
mbasmanova, line 593: @kevinwilfong Kevin, this change is not quite right. It allows existing code to compile without any changes, but it doesn't allow changing that code to start using the new API. In general, I don't think it is possible to make a change like this backwards-compatible without changing the mehod name.

CC: @MBkkt @Yuhta

MBkkt: Can we just break it?
I think this in reality needed only for fb monorepo.
In monorepo should be easy to update it once for all, isn't it?
For non fb dependencies it's already "updated"

I don't know, maybe I don't have enough context, to understand why is bad option, so it's just my opinion on this

mbasmanova: @MBkkt You are correct. We can let such changes temporary break Axiom GitHub and follow-up with "Advance Velox" PR to fix.

VELOX_ENABLE_BACKWARD_COMPATIBILITY was introduced to handle breaking changes for Prestissimo. That project is Github-First, e.g. Meta's monorepo stores a read-only copy which cannot be updated without updating Presto's GitHub first. Axiom is not GitHub-First, hence this technique doesn't apply.

MBkkt: It's not related, but is velox github first itself?

mbasmanova: No, Velox is not GitHub first. Source of truth for Velox is Meta's monorepo.

PR #14603, opened 2025-08-26
fix: Incorrect min max stats when the column value are infinity or -infinity

PingLiuPing:
When writing parquet files, if the value of a float or double column are both infinity or -infinity, the min and max stats of this column is wrong.
For example, suppose the column is float, and all values are -std::numeric_limits::infinity(), the min and max collected in TypeStatisticsImpl are [-Inf, std::numeric_limits::lowest()].
But the actual result should be [-Inf, -Inf].


PingLiuPing:
@majetideepak  can you please have a look? thanks.

majetideepak:
@PingLiuPing This code has been copied from the Arrow library. We need to check how Arrow handles this case.

PingLiuPing:
Yes, I have checked arrow code, https://github.com/apache/arrow/blob/main/cpp/src/parquet/statistics.cc#L82. And they didn't fix this yet.
I opened an issue https://github.com/apache/arrow/issues/47426, but no response yet.

majetideepak:
@PingLiuPing Can you check what DuckDB and Presto Java do?

PingLiuPing:
Thanks @majetideepak , will check and get back to you.

PingLiuPing:
@majetideepak  I checked duckdb implementation of parquet metadata.

From here it defines the default min and max value for numeric column. This is same to what arrow does.

https://github.com/duckdb/duckdb/blob/15fba9f47c2c547d8a9fa2cdb23f91e5eb8ea533/extension/parquet/include/writer/parquet_write_stats.hpp#L47-L54

https://github.com/duckdb/duckdb/blob/15fba9f47c2c547d8a9fa2cdb23f91e5eb8ea533/src/include/duckdb/common/limits.hpp#L24-L39

I think duckdb has same issue with us. See the stats update code:
https://github.com/duckdb/duckdb/blob/15fba9f47c2c547d8a9fa2cdb23f91e5eb8ea533/extension/parquet/include/writer/parquet_write_operators.hpp#L82-L95


For Presto java, I think it leverages the iceberg java library to write the parquet data file and hence it is compatible with iceberg spec. Adding a background of this issue, we found it from iceberg spark unit test. See https://github.com/apache/iceberg/blob/38b5d7a820d31d512918ea6d09d538be64551098/spark/v3.4/spark/src/test/java/org/apache/iceberg/spark/sql/TestAggregatePushDown.java#L770.

I tested this in Presto iceberg connector.

`presto:iceberg> insert into test_stats values(1.234,infinity(),1),(1.235, infinity(),2);`

 And then check the parquet metadat:

```
■■■■■■■■■■■■■■■■■■■■■■■■path_in_schema = list
■■■■■■■■■■■■■■■■■■■■■■■■■■■■c_double
■■■■■■■■■■■■■■■■■■■■■■■■codec = 2
■■■■■■■■■■■■■■■■■■■■■■■■num_values = 2
■■■■■■■■■■■■■■■■■■■■■■■■total_uncompressed_size = 46
■■■■■■■■■■■■■■■■■■■■■■■■total_compressed_size = 83
■■■■■■■■■■■■■■■■■■■■■■■■data_page_offset = 55
■■■■■■■■■■■■■■■■■■■■■■■■statistics = Statistics
■■■■■■■■■■■■■■■■■■■■■■■■■■■■max = b'\x00\x00\x00\x00\x00\x00\xf0\x7f'
■■■■■■■■■■■■■■■■■■■■■■■■■■■■min = b'\x00\x00\x00\x00\x00\x00\xf0\x7f'
■■■■■■■■■■■■■■■■■■■■■■■■■■■■max_value = b'\x00\x00\x00\x00\x00\x00\xf0\x7f'
■■■■■■■■■■■■■■■■■■■■■■■■■■■■min_value = b'\x00\x00\x00\x00\x00\x00\xf0\x7f'
```
b'\x00\x00\x00\x00\x00\x00\xf0\x7f', little endian, is +inf.

PingLiuPing:
@majetideepak  Could you please have a look? Thanks.

PingLiuPing:
@majetideepak  Could you please have a look? Thank you.

PingLiuPing:
@mbasmanova  This PR is part of the overall solution of writing iceberg table. It fixes a discrepancy between current parquet writer and Iceberg's implementation. Could you help take a look? Thank you.

mbasmanova:
@PingLiuPing Would you update PR description to summaries the discussion in this PR. Specifically, please, explain that this is an existing bug in Arrow and link the issue. Also, clarify how DuckDB and Presto handle this. If DuckDB has a bug as well, please, file an issue and link. Thanks.

Would you also clarify what is the impact of this bug? What doesn't work or which queries produce wrong results?

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -94,9 +94,15 @@ struct CompareHelper {
...
+    if constexpr (std::is_floating_point_v<T>) {
+      return std::numeric_limits<T>::infinity();
+    }
     return std::numeric_limits<T>::max();
   }
   constexpr static T DefaultMax() {
+    if constexpr (std::is_floating_point_v<T>) {
+      return -std::numeric_limits<T>::infinity();
+    }
     return std::numeric_limits<T>::lowest();
```
mbasmanova: this can be ::min now that floating point types are handled separately

PingLiuPing:
> @PingLiuPing Would you update PR description to summaries the discussion in this PR. Specifically, please, explain that this is an existing bug in Arrow and link the issue. Also, clarify how DuckDB and Presto handle this. If DuckDB has a bug as well, please, file an issue and link. Thanks.
> 
> Would you also clarify what is the impact of this bug? What doesn't work or which queries produce wrong results?

Thank you. I updated PR description.

mbasmanova:
@PingLiuPing Thank you for updating PR description. Would you also clarify what's the impact of this bug? Other than metadata looking incorrect, is there is problem with querying such tables? Also, would you open an issue in DuckDB repo and link it here?

majetideepak (approved):
Sorry for missing this PR. I likely mistook this for the other parquet writer PR.

PingLiuPing:
> @PingLiuPing Thank you for updating PR description. Would you also clarify what's the impact of this bug? Other than metadata looking incorrect, is there is problem with querying such tables? Also, would you open an issue in DuckDB repo and link it here?

@mbasmanova Thank you. I added following to explain the impact of this bug.

> This discrepancy can cause aggregation queries to fail in Gluten. By default, Spark enables Iceberg aggregation pushdown (spark.sql.iceberg.aggregate-push-down.enabled), see https://iceberg.apache.org/docs/latest/spark-configuration/#spark-sql-options. For queries like SELECT MIN(c1), MAX(c2) FROM T, Spark reads the column statistics directly from the manifest files. If the min/max statistics are incorrect, the query can return wrong results.


Opened an issue in duckdb https://github.com/duckdb/duckdb/issues/20014

mbasmanova:
@PingLiuPing 

> For queries like SELECT MIN(c1), MAX(c2) FROM T, Spark reads the column statistics directly from the manifest files. If the min/max statistics are incorrect, the query can return wrong results.

Thank you. This helps quite a bit.

PingLiuPing:
@mbasmanova  FYI, duckdb has fixed this issue https://github.com/duckdb/duckdb/issues/20014#issuecomment-3611004919

mbasmanova:
> @mbasmanova FYI, duckdb has fixed this issue [duckdb/duckdb#20014 (comment)](https://github.com/duckdb/duckdb/issues/20014#issuecomment-3611004919)

Nice. Super quick. Thank you for reporting.

PR #14610, opened 2025-08-26
feat: Add VarcharEnum support for enum_key Presto function

HeidiHan0000:
Summary: Added VarcharEnum type support for enum_key function. Includes adding VarcharEnumParameterVariableBindings to SignatureBinder.

Differential Revision: D80992639

### DIFF
--- a/velox/functions/prestosql/EnumFunctions.h
+++ b/velox/functions/prestosql/EnumFunctions.h
@@ -20,10 +20,11 @@
 #include "velox/core/QueryConfig.h"
 #include "velox/functions/Macros.h"
 #include "velox/functions/prestosql/types/BigintEnumType.h"
+#include "velox/functions/prestosql/types/VarcharEnumType.h"
 
 namespace facebook::velox::functions {
 
-template <typename TExec>
+template <typename TExec, typename EnumTypePtr>
 struct EnumKeyFunction {
   VELOX_DEFINE_FUNCTION_TYPES(TExec);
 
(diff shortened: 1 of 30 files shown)


velox/functions/prestosql/EnumFunctions.h
```diff
@@ -33,22 +34,46 @@ struct EnumKeyFunction {
...
+      const core::QueryConfig& /*config*/,
+      const arg_type<VarcharEnum<E1>>* /*input*/) {
+    VELOX_USER_CHECK_EQ(
+        inputTypes.size(), 1, "Expected 1 input type for enum_key function.");
+    enumPtr_ = inputTypes[0];
   }
 
   Status call(out_type<Varchar>& result, const int64_t& input) {
-    auto keyAt = enumPtr_->keyAt(input);
+    auto bigintEnumPtr = asBigintEnum(enumPtr_);
```
mbasmanova: dynamic_cast is expensive; can we void applying it to each row?

velox/functions/prestosql/EnumFunctions.h
```diff
@@ -47,8 +60,17 @@ struct EnumKeyFunction {
     return Status::OK();
   }
 
+  Status call(out_type<Varchar>& result, const std::string& input) {
+    auto keyAt = enumPtr_->keyAt(input);
+    if (!keyAt.has_value()) {
+      return Status::UserError("Value '{}' not in enum 'VarcharEnum'", input);
```
mbasmanova: perhaps, rename VarcharEnum with an actual type name since VarcharEnum doesn't provide enough specifics for debugging

HeidiHan0000: I changed it to print the whole type, thanks!

PR #14612, opened 2025-08-26
fix: Add check to isInMap before casting as mutable buffer

peterenescu:
Summary:
Add check to flatMapVector function isInMap to confirm inMaps[channel] is not null before casting as mutable buffer.

inMap is allowed to be null and is encouraged for short circuiting for keys that are present in every row; however, we must confirm that it is not null before casting as mutable buffer and attempting to reference individual bits.

Differential Revision: D81077301

### DIFF
--- a/velox/vector/FlatMapVector.h
+++ b/velox/vector/FlatMapVector.h
@@ -214,8 +214,12 @@ class FlatMapVector : public BaseVector {
     if (inMaps_.size() <= keyChannel) {
       return true;
     }
-    auto* inMap = inMapsAt(keyChannel)->asMutable<uint64_t>();
-    return inMap ? bits::isBitSet(inMap, index) : true;
+    if (auto& inMap = inMapsAt(keyChannel)) {
+      return bits::isBitSet(inMap->as<uint64_t>(), index);
+    } else {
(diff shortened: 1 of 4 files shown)


velox/vector/FlatMapVector.h
```diff
@@ -214,8 +214,13 @@ class FlatMapVector : public BaseVector {
     if (inMaps_.size() <= keyChannel) {
       return true;
     }
-    auto* inMap = inMapsAt(keyChannel)->asMutable<uint64_t>();
-    return inMap ? bits::isBitSet(inMap, index) : true;
+    if (auto inMap = inMapsAt(keyChannel)) {
+      auto* mutableInMap = inMap->asMutable<uint64_t>();
+      return bits::isBitSet(mutableInMap, index);
```
pedroerp: since you're at it, here it doesn't need to be mutable:

```
  return bits::isBitSet(inMap->as<uint64_t>(), index);
```

pedroerp (approved):
Thanks @peterenescu

velox/vector/FlatMapVector.h
peterenescu: Good to know! That'll help me cleanup some of the filter code

PR #14617, opened 2025-08-27
fix: Return non-empty distribution function on empty QDigest

jonhehir:
Summary:
On an empty Q-Digest, the distribution function currently returns an empty vector. While the concept of the distribution function is not particularly well defined for an empty set, it is more convenient to maintain the invariant that the distribution function over the range [rangeStart, rangeEnd] will always include the points (rangeStart, 0) and (rangeEnd, 1).

This change enforces that invariant in `QuantileDigest::getDistributionFunction`, so that an empty digest will return a vector with (rangeStart, 0) and (rangeEnd, 1). Non-empty digests add points in between these.

Differential Revision: D81058154


natashasehgal (approved):
LGTM

natashasehgal:
@kagamiori can you also take a quick look

kagamiori (approved):

PR #14621, opened 2025-08-27
misc: Remove unneeded use-count checking logic

lingbin:
`StirngIdMap::release()` ensures that the use-count(`numInUse`) 
of all entries in `idToEntry_` and `stringToId_` will never be equal 
to 0.  This is because once an entry's use-count reaches zero, it is 
removed from both maps.

This PR also includes two refactorings of 'StringIdMapTest.cpp':
- Remove unnecessary `make_unique`;
- Replace `const char*` with `std::string_view`, eliminating the need 
  for `strlen`.

### DIFF
--- a/velox/common/caching/StringIdMap.cpp
+++ b/velox/common/caching/StringIdMap.cpp
@@ -31,8 +31,8 @@ void StringIdMap::release(uint64_t id) {
   std::lock_guard<std::mutex> l(mutex_);
   auto it = idToEntry_.find(id);
   if (it != idToEntry_.end()) {
-    VELOX_CHECK_LT(
-        0, it->second.numInUse, "Extra release of id in StringIdMap");
+    VELOX_CHECK_GT(
+        it->second.numInUse, 0, "Extra release of id in StringIdMap");
     if (--it->second.numInUse == 0) {
       pinnedSize_ -= it->second.string.size();
       auto strIter = stringToId_.find(it->second.string);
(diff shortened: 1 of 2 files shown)


lingbin:
@majetideepak   Could you please review this PR? Thanks.

lingbin:
@majetideepak   Kindly ping.

majetideepak:
@lingbin I will let @xiaoxmeng provide feedback. I am not very familiar with this code.

lingbin:
@xiaoxmeng  Kindly ping. Could you please review this PR? It is not a big change.

velox/common/caching/StringIdMap.cpp
```diff
@@ -60,11 +60,11 @@ uint64_t StringIdMap::makeId(std::string_view string) {
   if (it != stringToId_.end()) {
     auto entry = idToEntry_.find(it->second);
     VELOX_CHECK(entry != idToEntry_.end());
-    if (++entry->second.numInUse == 1) {
-      pinnedSize_ += entry->second.string.size();
```
xiaoxmeng, line 64: So pinnedSize_ update is redundant?

xiaoxmeng (approved):
@lingbin LGTM. Thanks!

velox/common/caching/StringIdMap.cpp
lingbin: Yes, because there is no entry with a reference count of `0` in `idToEntry_`, `if (++entry->second.numInUse == 1)` will never be true.

majetideepak: @lingbin why did we not keep `++entry->second.numInUse` here like below?

lingbin: @majetideepak  `++entry->second.numInUse` is on L64 of the new file.  

The modifications made in `makeId()` and `recoverId()` are the same.

PR #14622, opened 2025-08-27
feat: Add Spark unix_timestamp support with timestamp and format arguments

### LINKED ISSUE #14624: Spark unix_timestamp not registered with arguments: (TIMESTAMP  VARCHAR)
### Bug description

Spark unix_timestamp not registered with arguments: (TIMESTAMP  VARCHAR) and is falling back in Gluten

### System information

NA

### Relevant logs

```bash

```

nimesh1601:
fixes #14624 Support unix_timestamp registered with arguments: (TIMESTAMP VARCHAR).

### DIFF
--- a/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
+++ b/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
@@ -332,17 +332,17 @@ TEST_F(DateTimeFunctionsTest, unixTimestampTimestampWithFormatInput) {
     return evaluateOnce<int64_t>(
         "unix_timestamp(c0, c1)", timestamp, formatStr);
   };
-  
+
   // Test that format parameter is ignored for timestamp input
   EXPECT_EQ(0, unixTimestamp(Timestamp(0, 0), "yyyy-MM-dd"));
   EXPECT_EQ(1, unixTimestamp(Timestamp(1, 990), "invalid-format"));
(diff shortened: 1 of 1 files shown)


nimesh1601:
@rui-mo can you review this ?

velox/docs/functions/spark/datetime.rst
```diff
@@ -423,6 +428,15 @@ These functions support TIMESTAMP and DATE input types.
         SELECT unix_timestamp(CAST(1739933174 AS TIMESTAMP)); -- 1739933174
         SELECT unix_timestamp(CAST(-1739933174 AS TIMESTAMP)); -- -1739933174
 
+.. spark:function:: unix_timestamp(timestamp, format) -> bigint
+   :noindex:
+
+    Returns the UNIX timestamp of the given ``timestamp`` in seconds. The ``format``
+    parameter is ignored for timestamp input, as per Spark behavior. ::
```
rui-mo, line 435: Hi @nimesh1601, thanks for working on the fix. Is the case that Spark ignore the 'format' in its implementation when the input is not string? If so, adding this signature in Velox might be unnecessary because we could modify Gluten code to not pass the 'format' parameter, and that should solve the fallback.

nimesh1601: Yes, Spark ignore 'format' in its implementation when the input is timestamp. Sure, can create a diff in gluten to avoid this fallback, added here to have all cases in one place.

nimesh1601:
Raised PR in Gluten : https://github.com/apache/incubator-gluten/pull/10567

PR #14623, opened 2025-08-27
feat: Add convenience API to create homogeneous ROW type

mbasmanova:
Summary:
Add convenience API to create a struct where all fields have the same type. Also, add an API to easily make a struct with a single field.

```
/// Returns a homogenous struct where all fields have the same type.
///
/// Example:
///
///   ROW({"a", "b", "c"}, REAL()) is a shortcut for
///     ROW({"a", "b", "c"}, {REAL(), REAL(), REAL()}).
RowTypePtr ROW(std::vector<std::string> names, const TypePtr& childType);

/// Returns a struct with a single field.
///
/// Example: ROW("a", BIGINT()) is a shortcut for ROW({{"a", BIGINT()}}).
RowTypePtr ROW(std::string name, TypePtr type);
```

Differential Revision: D81127629

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -866,31 +866,48 @@ ArrayTypePtr ARRAY(TypePtr elementType) {
   return TypeFactory<TypeKind::ARRAY>::create(std::move(elementType));
 }
 
+MapTypePtr MAP(TypePtr keyType, TypePtr valueType) {
+  return TypeFactory<TypeKind::MAP>::create(
+      std::move(keyType), std::move(valueType));
+}
+
 RowTypePtr ROW(std::vector<std::string> names, std::vector<TypePtr> types) {
   return TypeFactory<TypeKind::ROW>::create(std::move(names), std::move(types));
 }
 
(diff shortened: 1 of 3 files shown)


velox/type/Type.cpp
```diff
@@ -866,31 +866,48 @@ ArrayTypePtr ARRAY(TypePtr elementType) {
...
+MapTypePtr MAP(TypePtr keyType, TypePtr valueType) {
+  return TypeFactory<TypeKind::MAP>::create(
+      std::move(keyType), std::move(valueType));
+}
+
 RowTypePtr ROW(std::vector<std::string> names, std::vector<TypePtr> types) {
   return TypeFactory<TypeKind::ROW>::create(std::move(names), std::move(types));
 }
 
+RowTypePtr ROW(std::vector<std::string> names, TypePtr childType) {
```
Yuhta, line 878: We may need to use a different name to avoid ambiguity, something like `makeHomogenousRowType`

mbasmanova: I specifically want to keep the API short and simple. makeHomogenousRowType is way too long.

PR #14626, opened 2025-08-27
feat: Add P4HyperLogLog type

natashasehgal:
Summary:
Add data type P4HyperLogLog
Add casting of both from/to varbinary (its physical type)

Differential Revision: D81148399

### DIFF
--- a/velox/functions/prestosql/tests/utils/FunctionBaseTest.cpp
+++ b/velox/functions/prestosql/tests/utils/FunctionBaseTest.cpp
@@ -16,13 +16,17 @@
 #include "FunctionBaseTest.h"
 #include "velox/functions/FunctionRegistry.h"
 #include "velox/functions/prestosql/registration/RegistrationFunctions.h"
-#include "velox/functions/prestosql/types/QDigestRegistration.h"
+#include "velox/functions/prestosql/types/P4HyperLogLogRegistration.h"
 #include "velox/parse/TypeResolver.h"
 
 namespace facebook::velox::functions::test {
 
 void FunctionBaseTest::SetUpTestCase() {
   parse::registerTypeResolver();
+  // TODO: remove the registration of P4HyperLogLog here once it is registered
+  // through
+  // registerAllScalarFunctions().
+  registerP4HyperLogLogType();
   functions::prestosql::registerAllScalarFunctions();
   memory::MemoryManager::testingSetInstance(memory::MemoryManager::Options{});
 }
--- a/velox/functions/prestosql/types/P4HyperLogLogRegistration.cpp
+++ b/velox/functions/prestosql/types/P4HyperLogLogRegistration.cpp
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/prestosql/types/P4HyperLogLogRegistration.h"
+
+#include "velox/functions/prestosql/types/P4HyperLogLogType.h"
+#include "velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h"
+#include "velox/type/Type.h"
+
+namespace facebook::velox {
+namespace {
+class P4HyperLogLogTypeFactory : public CustomTypeFactory {
+ public:
+  TypePtr getType(const std::vector<TypeParameter>& parameters) const override {
+    VELOX_CHECK(parameters.empty());
+    return P4HYPERLOGLOG();
+  }
+
+  // P4HyperLogLog should be treated as Varbinary during type castings.
+  exec::CastOperatorPtr getCastOperator() const override {
+    return nullptr;
+  }
+
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& config) const override {
+    return std::make_shared<fuzzer::P4HyperLogLogInputGenerator>(
+        config.seed_, config.nullRatio_, config.pool_);
+  }
+};
+} // namespace
+void registerP4HyperLogLogType() {
+  registerCustomType(
+      "p4hyperloglog", std::make_unique<const P4HyperLogLogTypeFactory>());
+}
+} // namespace facebook::velox
--- a/velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
+++ b/velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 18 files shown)


velox/functions/prestosql/types/KHyperLogLogRegistration.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  // KHyperLogLog should be treated as Varbinary during type castings.
+  exec::CastOperatorPtr getCastOperator() const override {
+    return nullptr;
+  }
+
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& config) const override {
+    return std::make_shared<fuzzer::HyperLogLogInputGenerator>(
+        config.seed_, config.nullRatio_, config.pool_);
```
kagamiori: It feels like KHyperLogLog cannot use the HyperLogLog input generator directly since they have different structures? Please correct me if my understanding is wrong.
(You can add KHyperLogLog input generator in a separate PR if it's easier for you.)

velox/functions/prestosql/types/P4HyperLogLogRegistration.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  // P4HyperLogLog should be treated as Varbinary during type castings.
+  exec::CastOperatorPtr getCastOperator() const override {
+    return nullptr;
+  }
+
+  AbstractInputGeneratorPtr getInputGenerator(
+      const InputGeneratorConfig& config) const override {
+    return std::make_shared<fuzzer::HyperLogLogInputGenerator>(
+        config.seed_, config.nullRatio_, config.pool_);
```
kagamiori: Ditto. HyperLogLog input generator can potentially produce sparse HLL, but my understanding is P4HyperLogLog only allows dense HLL?

velox/functions/prestosql/types/tests/P4HyperLogLogTypeTest.cpp
```diff
@@ -0,0 +1,44 @@
...
+  ASSERT_EQ(P4HYPERLOGLOG()->toString(), "P4HYPERLOGLOG");
+
+  ASSERT_TRUE(hasType("P4HYPERLOGLOG"));
+  ASSERT_EQ(*getType("P4HYPERLOGLOG", {}), *P4HYPERLOGLOG());
+
+  ASSERT_FALSE(P4HYPERLOGLOG()->isOrderable());
+}
+
+TEST_F(P4HyperLogLogTypeTest, serde) {
+  testTypeSerde(P4HYPERLOGLOG());
```
kagamiori, line 42: Could you also add the cases of KHyperLogLog and P4HyperLogLog to TypeOfTest.cpp as well?

kagamiori (commented):
Hi @natashasehgal, thank you for adding the new types! It looks good to me except a few nits.

velox/functions/prestosql/types/KHyperLogLogRegistration.cpp
natashasehgal: Yes KHyperLogLog is different. I have removed it only adding P4 in this PR

velox/functions/prestosql/types/P4HyperLogLogRegistration.cpp
natashasehgal: Yes. The generator generates between 100-1000 values so it is always in DenseHll format

velox/functions/prestosql/tests/utils/FunctionBaseTest.cpp
```diff
@@ -26,6 +27,10 @@ void FunctionBaseTest::SetUpTestCase() {
   // TODO: remove the registraiton of QDigest here once it is registered through
   // registerAllScalarFunctions().
   registerQDigestType();
```
kagamiori: This can be removed now.

velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
```diff
@@ -0,0 +1,80 @@
...
+  P4HyperLogLogInputGenerator(
+      const size_t seed,
+      const double nullRatio,
+      memory::MemoryPool* pool);
+
+  variant generate() override;
+
+ private:
+  template <typename T>
+  variant generateTyped() {
```
kagamiori: Since this method is almost the same as HyperLogLogInputGenerator::generateTyped(), is it possible to make the HyperLogLogInputGenerator constructor receive an optional argument `minNumValues` and reuse that logic?

The test case in P4HyperLogLogInputGeneratorTest.cpp can then be moved to HyperLogLogInputGeneratorTest.cpp.

natashasehgal: Good idea. It is similar. Added that

velox/functions/prestosql/tests/utils/FunctionBaseTest.cpp
natashasehgal: I removed in PR - https://github.com/facebookincubator/velox/pull/14783
Will re-base this

velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
natashasehgal: Can you check the new implementation. I added an optional minNumValues and max values is 1000 for both HLL and P4HLL. Previously it was 100 for HLL. Should maxNumValues also be different for these two types?

velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#pragma once
+
+#include "velox/functions/prestosql/types/fuzzer_utils/HyperLogLogInputGenerator.h"
+
+namespace facebook::velox::fuzzer {
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
kagamiori: nit: No need for this now?

velox/functions/prestosql/types/fuzzer_utils/P4HyperLogLogInputGenerator.h
```diff
@@ -0,0 +1,34 @@
...
+class P4HyperLogLogInputGenerator : public HyperLogLogInputGenerator {
+ public:
+  P4HyperLogLogInputGenerator(
+      const size_t seed,
+      const double nullRatio,
+      memory::MemoryPool* pool);
+};
+#pragma GCC diagnostic pop
```
kagamiori: Ditto.

velox/functions/prestosql/types/fuzzer_utils/tests/HyperLogLogInputGeneratorTest.cpp
```diff
@@ -53,4 +54,29 @@ TEST_F(HyperLogLogInputGeneratorTest, generate) {
   }
 }
 
+TEST_F(HyperLogLogInputGeneratorTest, generateP4HyperLogLog) {
```
kagamiori: nit: I originally thought we would just use HyperLogLogInputGenerator for P4HyperLogLog too, so I suggested moving this unit test to HyperLogLogInputGeneratorTest.cpp. But it's good that we still keep a separate P4HyperLogLogInputGenerator. Then it's better to still have this test case in P4HyperLogLogInputGeneratorTest.cpp. Sorry for causing confusion.

kagamiori (approved):
LGTM. Thank you for adding the P4HyperLogLog with fuzzer support!

velox/functions/prestosql/types/fuzzer_utils/tests/HyperLogLogInputGeneratorTest.cpp
natashasehgal: Makes sense. I was going to add it to the same file, but inheritance seemed cleaner (one file per one data type). Added the P4HLL test to a separate class

PR #14628, opened 2025-08-27
feat: Support quantile stats

DanielMunozT:
Summary:
Support quantile stats.
This will break existing external (to meta) code that derives from `BaseStatsReporter` since the classes will become abstract. If we want to keep compatibility I need to make a default implementation for the methods that throws.

Differential Revision: D81152998

### DIFF
--- a/velox/common/base/StatsReporter.h
+++ b/velox/common/base/StatsReporter.h
@@ -117,6 +117,25 @@ class BaseStatsReporter {
       int64_t max,
       const std::vector<int32_t>& pcts) const = 0;
 
+  /// Register a quantile metric for quantile stats with export types,
+  /// quantiles, and sliding window periods.
+  /// @param key The key to identify the stat.
+  /// @param statTypes The list of stat types to export (e.g., AVG, SUM, COUNT).
+  /// @param pcts The quantile percentiles to track (as values between 0.0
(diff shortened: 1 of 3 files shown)


pedroerp:
Cc: @majetideepak to ensure this won't break anything from IBM.

mbasmanova (commented):
CC: @tanjialiang @xiaoxmeng

majetideepak:
CC: @czentgr

velox/common/base/StatsReporter.h
```diff
@@ -183,11 +224,44 @@ class DummyStatsReporter : public BaseStatsReporter {
...
+template <typename... Args>
+std::vector<StatType> statTypes(Args... args) {
+  return std::vector<StatType>{args...};
+}
+
+/// Create a vector of percentiles from variadic arguments.
+/// Usage: percentiles(0.5, 0.95, 0.99)
+template <typename... Args>
+std::vector<double> percentiles(Args... args) {
+  return std::vector<double>{static_cast<double>(args)...};
```
czentgr, line 255: Question, the `static_cast` here does the same thing as `std::forward<type>` if I am not mistaken.
So we should use this instead?

czentgr:
When this lands with the pure virtual functions we should advance Velox in PrestoC++ to handle pure virtual functions in the prometheus stats reporter.

PR #14632, opened 2025-08-27
feat: Add demo to show how to write udf with dynamic input and output

xiaoxmeng:
Summary: Add demo to show how to write a vector function in velox with dynamic input type and output type for feature injection to support map to row conversion at run time.

Differential Revision: D81176937


bikramSingh91 (approved):
accepting to unblock, added some comments internally, please take a look before merging

PR #13953, opened 2025-07-01
fix(cudf): Properly handle no input scenarios in CudfHashAggregation

devavret:
Previously `CudfHashAggregation` misbehaved if it never received any input during its lifetime by failing to ever set itself to finished, hence hanging the task. 
It would also produce the wrong output in case of no input during global aggregations, while we expect 1 row output with sentinel values depending on aggregation type (sum/min/max - null, count - 0)
his PR fixes both these issues.

PR #14640, opened 2025-08-28
fix(cuDF): Fix an exception when we get empty input for avg aggregation

devavret:
In https://github.com/facebookincubator/velox/pull/13953 we fixed the case where Single/Final aggregation received no input. That PR added the fix for sum/min/max/count but deferred the fix for avg. This PR adds that.


jinchengchenghh:
Please rebase to fix the red CI

karthikeyann:
@devavret Rebased the PR branch on main.

PR #14642, opened 2025-08-28
feat: Add enableFlatMap option in makeWriteTask function

juwentus1234:
Summary: Add enableFlatMap parameter in makeWriteTask function to default disable it.

Differential Revision: D81240622


velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -811,7 +812,7 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
 
         if (!hasUnsupportedMapKey(setup.schema->childAt(i))) {
           // %50 chance to enable flatmap for this map column.
-          if (folly::Random::oneIn(2)) {
+          if (enableFlatMap && folly::Random::oneIn(2)) {
```
Yuhta: The `oneIn` call is not taking seed?  That's probably why we have a repro issue

PR #14644, opened 2025-08-28
fix: Integer overflow for NUMBER in TypeParser

HeidiHan0000:
Summary:
As title, the added test initially failed with:
terminate called after throwing an instance of 'folly::ConversionError'

Differential Revision: D81189954


kewang1024:
@HeidiHan0000 thanks for the fix! Can you fix the format checks?

kewang1024 (approved):
Thank you @HeidiHan0000

PR #14646, opened 2025-08-29
fix: Make columns stats spec optional for table write plan builders

xiaoxmeng:
Summary: Fix pyspark-velox which doesn't set column stats spec

Differential Revision: D81299287


duanmeng (approved):
@xiaoxmeng LGTM, thanks.

PR #14647, opened 2025-08-29
feat(readers): Customize column loader to track average row size

HuamengJiang:
Summary:
Add a framework to complement the row size estimate heuristics, based on the retained vector sizes.

Currently this framework is used as a stop gap solution to still have functional row estimates when column stats are missing, and decoders couldn't provide a relatively cheap estimate. The current functionality gap in decoder row estimates caused various queries to run with super small batches (frequently just 10 rows), and vastly slowing down the downstream eval.

Reviewed By: Yuhta

Differential Revision: D80310282

### DIFF
--- a/velox/exec/TableScan.cpp
+++ b/velox/exec/TableScan.cpp
@@ -180,15 +180,16 @@ RowVectorPtr TableScan::getOutput() {
         }
         continue;
       }
-      const auto estimatedRowSize = dataSource_->estimatedRowSize();
-      readBatchSize_ =
-          estimatedRowSize == connector::DataSource::kUnknownRowSize
-          ? outputBatchRows()
-          : outputBatchRows(estimatedRowSize);
     }
     VELOX_CHECK(!needNewSplit_);
     VELOX_CHECK(!hasDrained());
 
+    const auto estimatedRowSize = dataSource_->estimatedRowSize();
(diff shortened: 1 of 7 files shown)


velox/exec/TableScan.cpp
```diff
@@ -180,15 +180,16 @@ RowVectorPtr TableScan::getOutput() {
...
-          estimatedRowSize == connector::DataSource::kUnknownRowSize
-          ? outputBatchRows()
-          : outputBatchRows(estimatedRowSize);
     }
     VELOX_CHECK(!needNewSplit_);
     VELOX_CHECK(!hasDrained());
 
+    const auto estimatedRowSize = dataSource_->estimatedRowSize();
+    // TODO: Expose this to operator stats.
+    LOG(INFO) << "estimatedRowSize = " << estimatedRowSize;
```
mbasmanova, line 189: @HuamengJiang @Yuhta Is this intentional? Seems spammy.

mbasmanova: Removing: [#14661](https://github.com/facebookincubator/velox/pull/14661)

Yuhta: Sorry I was asking @HuamengJiang removing the loggings before merging.
@HuamengJiang no pre-approval for you next time :(

PR #14653, opened 2025-08-29
feat: Support dynamic quantile stat (#14652)

DanielMunozT:
Summary:

This will break existing external (to meta) code that derives from BaseStatsReporter since the classes will become abstract. If we want to keep compatibility I need to make a default implementation for the methods that throws.

Differential Revision: D81360340


tanjialiang:
> This will break existing external (to meta) code that derives from BaseStatsReporter since the classes will become abstract. If we want to keep compatibility I need to make a default implementation for the methods that throws.

cc @majetideepak I think you guys previously added some custom reporter implementing this one in presto_cpp, right? Maybe @DanielMunozT  could you help to make corresponding changes in prestodb codebase afterwards?

tanjialiang (approved):

PR #14654, opened 2025-08-29
fix(build): Use sudo correct during setup for Ubuntu

czentgr:
This change avoids problems calling sudo directly which can result in sudo not being found running the script during a container build.


assignUser (approved):
Thanks!

assignUser:
#14928 should fix docker ci

kgpai:
@czentgr  Can you fix broken image build ?

kgpai:
@czentgr Pleae let me know when you think this is ready to merge. Removing label till its ready, add it back when you think its done and we will automatically be notified.

PR #14655, opened 2025-08-29
refactor(expression): Extract helper functions to ExprUtils

pramodsatya:


velox/expression/ExprUtils.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::expression::utils {
+
+/// Returns true if expr is of type CallTypedExpr and the expr name matches
+/// the passed name. Otherwise, returns false.
+bool isCall(const core::TypedExprPtr& expr, const std::string& name);
+
+/// Utility method to check eligibility for flattening.
```
mbasmanova: perhaps, clarify what this does, e.g.

/// @return true if all inputs have equivalent types.

mbasmanova (approved):
@czentgr Thanks. It would be nice to follow-up with unit tests for newly-extracted helper functions.

PR #14656, opened 2025-08-30
feat(connector): Retain configuration object in TPCH connector

hdikeman:
Summary:
In this change I am retaining the connector configuration provided by the factory method in the TPCH connector. These config are also made available via the connectorConfig() API, which was previously unimplemented.

Adding a test to check that the config provided to the factory method is properly retained and exposed by the TPCH Connector.

Differential Revision: D81365452


hdikeman:
Closing so I can split from the closed-source changes after offline conversation with Masha

PR #14658, opened 2025-08-30
refactor(expression): Formalize ExpressionRewrite framework

pramodsatya:
Expression rewrites are currently defined in `VectorFunction.cpp`, rewrites are registered only for scalar functions, and rewrites are applied during expression compilation. In the upcoming `ExpressionOptimizer` work (#14523), we intend to build on the existing `ExpressionRewrite` support and introduce more expression rewrites for special form expressions. 

This refactor formalizes the existing `ExpressionRewrite` framework so it can be expanded upon in the `ExpressionOptimizer`.

### DIFF
--- a/velox/expression/ExprRewriteRegistry.cpp
+++ b/velox/expression/ExprRewriteRegistry.cpp
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/FunctionSignature.h"
+
+namespace facebook::velox::expression {
+
+void ExprRewriteRegistry::registerRewrite(ExpressionRewrite rewrite) {
+  registry_.withWLock([&](auto& list) { list.push_back(std::move(rewrite)); });
+}
+
+void ExprRewriteRegistry::clear() {
+  registry_.withWLock([&](auto& list) { list.clear(); });
+}
+
+core::TypedExprPtr ExprRewriteRegistry::rewrite(
+    const core::TypedExprPtr& expr) {
+  core::TypedExprPtr result = expr;
+  registry_.withRLock([&](const auto& list) {
+    for (const auto& rewrite : list) {
+      VELOX_CHECK_NOT_NULL(rewrite);
+      if (auto rewritten = (rewrite)(expr)) {
+        result = rewritten;
+        break;
+      }
+    }
+  });
+
+  return result;
+}
+} // namespace facebook::velox::expression
--- a/velox/expression/ExprRewriteRegistry.h
+++ b/velox/expression/ExprRewriteRegistry.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <folly/Synchronized.h>
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::expression {
+
+/// An expression re-writer that takes an expression and returns an equivalent
+/// expression or nullptr if re-write is not possible.
+using ExpressionRewrite =
+    std::function<core::TypedExprPtr(const core::TypedExprPtr)>;
+
+class ExprRewriteRegistry {
+ public:
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. For each expression, rewrites are applied in the
+  /// order they were registered. The first rewrite that returns non-null result
+  /// terminates the re-write for that particular expression.
+  void registerRewrite(ExpressionRewrite rewrite);
+
+  /// Clears the registry to remove all registered rewrites.
+  void clear();
+
+  core::TypedExprPtr rewrite(const core::TypedExprPtr& expr);
+
+  static ExprRewriteRegistry& instance() {
+    static ExprRewriteRegistry kInstance;
+    return kInstance;
+  }
+
+ private:
+  folly::Synchronized<std::vector<ExpressionRewrite>> registry_;
+};
+} // namespace facebook::velox::expression
--- a/velox/expression/tests/ExprRewriteRegistryTest.cpp
+++ b/velox/expression/tests/ExprRewriteRegistryTest.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "gtest/gtest.h"
+
+namespace facebook::velox::expression {
+
+namespace {
+
+class ExprRewriteRegistryTest : public testing::Test {};
+
+TEST_F(ExprRewriteRegistryTest, basic) {
+  expression::ExprRewriteRegistry registry;
+  expression::ExpressionRewrite testRewrite =
+      [&](const core::TypedExprPtr& input) {
+        return std::make_shared<core::CallTypedExpr>(
+            input->type(), "test_expr", input, input);
+      };
+  registry.registerRewrite(testRewrite);
+
+  auto input =
+      std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(123));
+  const auto rewritten = registry.rewrite(input);
+  ASSERT_TRUE(rewritten->isCallKind());
+  ASSERT_TRUE(rewritten->type()->isBigint());
+  const auto rewrittenCall = rewritten->asUnchecked<core::CallTypedExpr>();
+  ASSERT_EQ(rewrittenCall->inputs().size(), 2);
+  ASSERT_EQ(rewrittenCall->name(), "test_expr");
+
+  registry.clear();
+  const auto rewriteAfterClear = registry.rewrite(input);
+  ASSERT_TRUE(*rewriteAfterClear == *input);
+}
+
+} // namespace
+} // namespace facebook::velox::expression
--- a/velox/functions/prestosql/Reduce.cpp
+++ b/velox/functions/prestosql/Reduce.cpp
@@ -16,6 +16,7 @@
 #include "velox/common/base/RuntimeMetrics.h"
 #include "velox/core/Expressions.h"
 #include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewriteRegistry.h"
 #include "velox/expression/VectorFunction.h"
 #include "velox/functions/lib/LambdaFunctionUtil.h"
 
@@ -423,7 +424,7 @@ VELOX_DECLARE_VECTOR_FUNCTION_WITH_METADATA(
     std::make_unique<ReduceFunction>());
 
 void registerReduceRewrites(const std::string& prefix) {
-  exec::registerExpressionRewrite(
+  expression::ExprRewriteRegistry::instance().registerRewrite(
       [prefix](const auto& expr) { return rewriteReduce(prefix, expr); });
 }
 
--- a/velox/expression/CMakeLists.txt
+++ b/velox/expression/CMakeLists.txt
@@ -42,6 +42,7 @@ velox_add_library(
   EvalCtx.cpp
   Expr.cpp
   ExprCompiler.cpp
+  ExprRewriteRegistry.cpp
   ExprToSubfieldFilter.cpp
   ExprUtils.cpp
   FieldReference.cpp
--- a/velox/expression/ExprCompiler.cpp
+++ b/velox/expression/ExprCompiler.cpp
@@ -18,13 +18,13 @@
 #include "velox/expression/ConstantExpr.h"
 #include "velox/expression/Expr.h"
 #include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewriteRegistry.h"
 #include "velox/expression/ExprUtils.h"
 #include "velox/expression/FieldReference.h"
 #include "velox/expression/LambdaExpr.h"
 #include "velox/expression/RowConstructor.h"
 #include "velox/expression/SimpleFunctionRegistry.h"
 #include "velox/expression/SpecialFormRegistry.h"
-#include "velox/expression/VectorFunction.h"
 
 namespace facebook::velox::exec {
 
@@ -305,15 +305,6 @@ std::vector<VectorPtr> getConstantInputs(const std::vector<ExprPtr>& exprs) {
   return constants;
 }
 
-core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
-  for (auto& rewrite : expressionRewrites()) {
-    if (auto rewritten = rewrite(expr)) {
-      return rewritten;
-    }
-  }
-  return expr;
-}
-
 ExprPtr compileCall(
     const TypedExprPtr& expr,
     std::vector<ExprPtr> inputs,
@@ -527,7 +518,7 @@ ExprPtr compileExpression(
     memory::MemoryPool* pool,
     const std::unordered_set<std::string>& flatteningCandidates,
     bool enableConstantFolding) {
-  auto rewritten = rewriteExpression(expr);
+  auto rewritten = expression::ExprRewriteRegistry::instance().rewrite(expr);
   if (rewritten.get() != expr.get()) {
     scope->rewrittenExpressions.push_back(rewritten);
   }
--- a/velox/expression/VectorFunction.cpp
+++ b/velox/expression/VectorFunction.cpp
@@ -245,13 +245,4 @@ bool registerVectorFunction(
       name, signatures, factory, metadata, overwrite);
 }
 
-std::vector<ExpressionRewrite>& expressionRewrites() {
-  static std::vector<ExpressionRewrite> rewrites;
-  return rewrites;
-}
-
-void registerExpressionRewrite(ExpressionRewrite rewrite) {
-  expressionRewrites().emplace_back(rewrite);
-}
-
 } // namespace facebook::velox::exec
--- a/velox/expression/VectorFunction.h
+++ b/velox/expression/VectorFunction.h
@@ -279,22 +279,6 @@ bool registerStatefulVectorFunction(
     VectorFunctionMetadata metadata = {},
     bool overwrite = true);
 
-/// An expression re-writer that takes an expression and returns an equivalent
-/// expression or nullptr if re-write is not possible.
-using ExpressionRewrite = std::function<core::TypedExprPtr(core::TypedExprPtr)>;
-
-/// Returns a list of registered re-writes.
-std::vector<ExpressionRewrite>& expressionRewrites();
-
-/// Appends a 'rewrite' to 'expressionRewrites'.
-///
-/// The logic that applies re-writes is very simple and assumes that all
-/// rewrites are independent. Re-writes are applied to all expressions starting
-/// at the root and going down the hierarchy. For each expression, rewrites are
-/// applied in the order they were registered. The first rewrite that returns
-/// non-null result terminates the re-write for this particular expression.
-void registerExpressionRewrite(ExpressionRewrite rewrite);
-
 } // namespace facebook::velox::exec
 
 // Private. Return the external function name given a UDF tag.
--- a/velox/expression/tests/CMakeLists.txt
+++ b/velox/expression/tests/CMakeLists.txt
@@ -23,6 +23,7 @@ add_executable(
   CustomTypeTest.cpp
   ExprCompilerTest.cpp
   ExprEncodingsTest.cpp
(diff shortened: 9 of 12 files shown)


velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,36 @@
...
+core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
+  const auto& rewriteRegistry = exec::expressionRewriteRegistry();
+  const auto& rewriteNames = rewriteRegistry.getExpressionRewriteNames();
+  for (const auto& name : rewriteNames) {
+    auto expressionRewriteFunc = *rewriteRegistry.getExpressionRewrite(name);
+    if (auto rewritten = expressionRewriteFunc(expr)) {
+      return rewritten;
+    }
+  }
+  return nullptr;
```
czentgr: I know we've talked about nullptr here in the other comment. But this changes the behavior.
I'm surprised it works because in ExprCompiler line 568 it compares the pointers and if they don't match adds the new expression to the scope->rewrittenExpressions list. And that has no negative consequences?

In line 572 it makes sure we don't pass on nullptr. 
I'm wondering if we only want to add to the scope->rewrittenExpressions list if `rewritten` is non-null.

velox/expression/ExprRewrite.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace facebook::velox::expression {
+
+/// An expression re-writer that takes an expression and returns an equivalent
+/// expression or nullptr if re-write is not possible.
+using ExpressionRewrite =
+    std::function<core::TypedExprPtr(const core::TypedExprPtr)>;
+
+/// Applies all registered expression rewrites to `expr` sequentially and
+/// constant folds the result. Returns a rewritten TypedExpr.
```
czentgr: We don't constant fold the final expression - at most we only rewrite and/or constant fold its inputs (it depends on `expressionRewriteFunc`). That was done in the external API which we don't have here yet.

czentgr (commented):
Can we add a small unit test? We can look at the unit tests for the other registries.

velox/expression/ExprRewrite.h
pramodsatya: Updated comment, thanks for pointing it out.

velox/expression/ExprRewrite.cpp
pramodsatya: Updated to return `expr` here, thanks!

pramodsatya (commented):
Thanks for the feedback @czentgr, added tests for `ExprRewriteRegistry` and changed the namespace to `expression` for consistency. Could you please take another look?

velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
+  const auto& rewriteRegistry = expressionRewriteRegistry();
+  const auto& rewriteNames = rewriteRegistry.getExpressionRewriteNames();
+  for (const auto& name : rewriteNames) {
+    auto expressionRewriteFunc = *rewriteRegistry.getExpressionRewrite(name);
```
czentgr: `expressionRewriteFunc` could be `nullptr` if the name didn't match. This is unlikely but possible with the current implementation if say the function was unregistered.

pramodsatya: Thanks for pointing it out, in this use-case `expressionRewriteFunc` should not be a `nullptr` since the rewrite name is also obtained from the registry, added a check to ensure it is not a `nullptr`.

czentgr: Yes, in theory we should get names that have registrations and then we will always have a non-nullptr. But there is a tiny theoretical race condition between getting the names and actually looping over them and making the call. Because we allow to unregister in theory this could have been called and the list is now empty and nullptr is returned.
One thing would be to use a VELOX_CHECK_NOT_NULL instead of line 29 if we never expect a nullptr to occur.

pramodsatya: Got it, updated to VELOX_CHECK_NOT_NULL, thanks!

czentgr:
@mbasmanova This is the expression rewrite framework as a separate PR to reduce the amount of overall changes. Please take a look when you have a chance. It's a subset of the previously reviewed larger PR that we are breaking up.

mbasmanova:
@czentgr The issues raised in #14734 might be relevant.

velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <set>
+
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
+  const auto& rewriteRegistry = expressionRewriteRegistry();
```
mbasmanova: nit: rewriteRegistry -> registry

velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,37 @@
...
+  const auto& rewriteNames = rewriteRegistry.getExpressionRewriteNames();
+  for (const auto& name : rewriteNames) {
+    auto expressionRewriteFunc = *rewriteRegistry.getExpressionRewrite(name);
```
mbasmanova: nit: expressionRewriteFunc -> rewrite

velox/expression/ExprRewrite.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace facebook::velox::expression {
+
+/// An expression re-writer that takes an expression and returns an equivalent
+/// expression or nullptr if re-write is not possible.
+using ExpressionRewrite =
+    std::function<core::TypedExprPtr(const core::TypedExprPtr)>;
+
+/// Applies all registered expression rewrites to `expr` sequentially, returns
+/// the rewritten expression.
```
mbasmanova: perhaps, clarify that it stops at the first re-write that modifies the expression

velox/expression/ExprRewriteRegistry.cpp
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/FunctionSignature.h"
+
+namespace facebook::velox::expression {
+
+namespace {
+ExpressionRewriteRegistry& expressionRewriteRegistryInternal() {
+  static ExpressionRewriteRegistry instance;
```
mbasmanova: naming: kInstance

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+#pragma once
+
+#include "folly/Synchronized.h"
+#include "velox/expression/ExprRewrite.h"
+
+namespace facebook::velox::expression {
+
+using ExprRewriteRegistryType = std::
+    unordered_map<std::string, std::unique_ptr<expression::ExpressionRewrite>>;
+class ExpressionRewriteRegistry {
```
mbasmanova: add empty line

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "folly/Synchronized.h"
+#include "velox/expression/ExprRewrite.h"
+
+namespace facebook::velox::expression {
+
+using ExprRewriteRegistryType = std::
```
mbasmanova: does this need to be visible publicly?

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerExpressionRewrite(
+      const std::string& name,
```
mbasmanova: Would you clarify that this 'name' is for?

Does it need to be unique?

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+ public:
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerExpressionRewrite(
```
mbasmanova: registerExpressionRewrite -> register (since ExpressionRewrite is part of the context / class name)

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+      const std::string& name,
+      std::unique_ptr<expression::ExpressionRewrite> expressionRewrite);
+
+  void unregisterExpressionRewrites();
```
mbasmanova: unregisterExpressionRewrites -> unregisterAll

velox/expression/ExprRewrite.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <set>
+
+#include "velox/expression/ExprRewrite.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr) {
+  const auto& rewriteRegistry = expressionRewriteRegistry();
+  const auto& rewriteNames = rewriteRegistry.getExpressionRewriteNames();
```
mbasmanova: Why do we need getNames + getRewrite(name)? Why not just get a list of rewrites directly?

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::vector<std::string> getExpressionRewriteNames() const;
+
+ private:
+  folly::Synchronized<ExprRewriteRegistryType> registry_;
+};
+
+const ExpressionRewriteRegistry& expressionRewriteRegistry();
+
+void registerExpressionRewrite(
```
mbasmanova: Do we need these functions?

velox/expression/tests/ExprRewriteRegistryTest.cpp
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "gtest/gtest.h"
+
+namespace facebook::velox::expression::test {
+
+class ExprRewriteRegistryTest : public testing::Test {};
```
mbasmanova, line 23: put this class into anonymous namespace inside same namespace as code being tested

```
namespace facebook::velox::expression {
namespace {
...
```

velox/expression/tests/ExprRewriteRegistryTest.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+TEST_F(ExprRewriteRegistryTest, basic) {
+  expression::ExpressionRewrite testRewrite =
+      [&](const core::TypedExprPtr& input) {
+        return std::make_shared<core::CallTypedExpr>(
+            input->type(),
+            std::vector<core::TypedExprPtr>{input, input},
```
mbasmanova: `std::vector<core::TypedExprPtr>` not needed if you use the new API you just added

velox/functions/prestosql/Reduce.cpp
```diff
@@ -421,8 +422,12 @@ VELOX_DECLARE_VECTOR_FUNCTION_WITH_METADATA(
     std::make_unique<ReduceFunction>());
 
 void registerReduceRewrites(const std::string& prefix) {
-  exec::registerExpressionRewrite(
-      [prefix](const auto& expr) { return rewriteReduce(prefix, expr); });
+  expression::registerExpressionRewrite(
+      "reduce",
+      std::make_unique<expression::ExpressionRewrite>(
```
mbasmanova: is this a unique_ptr over std::function?

velox/expression/ExprRewriteRegistry.h
czentgr: This was the name to identify the rewrite itself for debugging purposes. It was one of the prior comments that it would be good to have it for debugging even if unused?

velox/expression/ExprRewriteRegistry.h
czentgr: This follows a pattern used for other registries as well. This is the external API to add new rewrites to the registry.

velox/expression/ExprRewriteRegistry.h
czentgr: `register` seems to be a reserved keyword. It gives unqualified id compile error when we attempt to use it.
How about `registerRewrite`?

mbasmanova: Sounds good.

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerRewrite(const ExpressionRewrite& expressionRewrite);
+
+  void unregisterAll();
```
aditi-pandit: Nit : Maybe call this clear or clearRegistry()

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+#include "velox/expression/ExprRewrite.h"
+
+namespace facebook::velox::expression {
+
+class ExpressionRewriteRegistry {
+ public:
+  /// Appends a 'rewrite' to 'expressionRewrites'.
+  ///
+  /// The logic that applies re-writes is very simple and assumes that all
+  /// rewrites are independent. Re-writes are applied to all expressions
```
aditi-pandit: Don't entirely follow this... In the code I remember seeing that the rewrites were applied at the root and could traverse the tree if needed. The rewrite framework didn't do any traversal as such. Please can you state more clearly.

aditi-pandit (commented):
Thanks @pramodsatya. Overall the code looks good to me, and I don't have much comments beyond the original PR.

Just 2 minor nits.

velox/expression/ExprRewriteRegistry.h
pramodsatya: Thanks for pointing it out, this comment is borrowed from the original function in `VectorFunction.h` and the expression rewrite `starting from the root and traversing down the tree` being referred to here seems to be related to how the rewrites are applied in the expression compiler while recursively traversing down the tree. 
The rewrites for special form expressions (to be added in a follow-up as part of the expression optimizer PR) will be traversing the tree to rewrite the expression, but this is not applicable to the existing rewrites for functions.

Could I elaborate on this distinction in the follow-up PR where special form expression rewrites will be added?

pramodsatya:
Thanks for the feedback @mbasmanova, @aditi-pandit, addressed the review comments. Could you please take another look?

velox/expression/tests/ExprRewriteRegistryTest.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "gtest/gtest.h"
+
+namespace facebook::velox::expression::test {
```
mbasmanova: drop ::test suffix; this applies only to test utilities, not tests themselves

velox/expression/ExprRewriteRegistry.h
mbasmanova: @pramodsatya 

> Re-writes are applied to all expressions starting at the root and going down the hierarchy.

Aditi has a good point. Let's remove this sentence. Like you pointed out this doesn't apply to this method. It applies to the caller of this method (ExprCompiler).

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+  /// starting at the root and going down the hierarchy. For each expression,
+  /// rewrites are applied in the order they were registered. The first rewrite
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerRewrite(const ExpressionRewrite& expressionRewrite);
```
mbasmanova: perhaps, better take the rewrite by value and move

expressionRewrite -> rewrite

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+
+  void clear();
```
mbasmanova, line 39: Please, document this method.

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+
+  std::vector<const ExpressionRewrite*> getRewrites() const;
```
mbasmanova: getRewrites -> rewrites

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+
+ private:
+  folly::Synchronized<std::vector<ExpressionRewrite>> registry_;
+};
+
+const ExpressionRewriteRegistry& expressionRewriteRegistry();
+
+void registerRewrite(const ExpressionRewrite& expressionRewrite);
```
mbasmanova: Let's remove this function and change the callers to use expressionRewriteRegistry().registerRewrite just like they use getRewrites.

I would suggest to go a step further and replace expressionRewriteRegistry function with a static method in ExpressionRewriteRegistry class.

mbasmanova: ExpressionRewriteRegistry::instance()

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,46 @@
...
+  /// that returns non-null result terminates the re-write for this particular
+  /// expression.
+  void registerRewrite(const ExpressionRewrite& expressionRewrite);
+
+  void clear();
+
+  std::vector<const ExpressionRewrite*> getRewrites() const;
+
+ private:
+  folly::Synchronized<std::vector<ExpressionRewrite>> registry_;
```
mbasmanova, line 49: @czentgr @aditi-pandit What's the retional of using folly::Synchronized here?

I believe this allows to call register / getRewrites / clear APIs concurrently, but getRewrites returns raw pointers and if one thread calls clear while another works with these raw pointers, there will be a crash, no?

Should we simply state that registry is not thread safe? The application is expected to populated it at startup and not modify thereafter.

velox/expression/ExprRewrite.h
```diff
@@ -0,0 +1,32 @@
...
+
+/// An expression re-writer that takes an expression and returns an equivalent
+/// expression or nullptr if re-write is not possible.
+using ExpressionRewrite =
+    std::function<core::TypedExprPtr(const core::TypedExprPtr)>;
+
+/// Applies all registered expression rewrites to `expr`, returns the rewritten
+/// expression once a rewrite is successfully applied. Returns `expr` is none
+/// of the registered rewrites could be applied.
+core::TypedExprPtr rewriteExpression(const core::TypedExprPtr& expr);
```
mbasmanova: Any particular reason not to have this as a method of ExpressionRewriteRegistry class?

Would it make sense to combine this header file with ExprRewriteRegistry?

BTW, let's make sure header file matches class name: ExpressionRewriteRegistry -> ExprRewriteRegistry

velox/expression/ExprRewriteRegistry.h
czentgr: @mbasmanova This was modeled after the `SpecialFormRegistry`. 
It appears to have the same problem in returning a raw pointer to a function and can be technically unregistered. The question is why a thread would unregister. Maybe it might be more likely to unregister rewrites than the special form functions.

For rewrites we wouldn't expect a user to remove the rewrites for the lifetime of the program. There wouldn't be a point in doing this unless an implementation removes the rewrites after initialization and re-registers only wanted registrations. We don't have an external API to remove individual rewrites.
However, if we want to future proof this we can protect the rewrites under the read lock instead of returning some list to loop over. We would have multiple threads hold the read lock longer but this is a shared lock so shouldn't be much of a problem.

pramodsatya:
Thanks @mbasmanova, we have addressed the review comments. Could you please take another look?

velox/expression/ExprRewriteRegistry.h
```diff
@@ -0,0 +1,51 @@
...
+  /// terminates the re-write for that particular expression.
+  void registerRewrite(const ExpressionRewrite rewrite);
+
+  /// Clears the registry to remove all registered rewrites.
+  void clear();
+
+  core::TypedExprPtr rewrite(const core::TypedExprPtr& expr);
+
+  static ExprRewriteRegistry& instance() {
+    static ExprRewriteRegistry instance;
```
mbasmanova: naming: instance -> kInstance

velox/expression/tests/ExprRewriteRegistryTest.cpp
```diff
@@ -0,0 +1,50 @@
...
+      };
+  registry.registerRewrite(testRewrite);
+
+  auto input =
+      std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(123));
+  const auto rewritten = registry.rewrite(input);
+  ASSERT_TRUE(rewritten->isCallKind());
+  ASSERT_TRUE(rewritten->type()->isBigint());
+  const auto rewrittenCall =
+      dynamic_cast<const core::CallTypedExpr*>(rewritten.get());
```
mbasmanova: nit: `rewritten->asUnchecked<core::CallTypedExpr>()`

velox/expression/ExprRewriteRegistry.h
pramodsatya: Fixed, apologies for the oversight.

velox/expression/ExprRewriteRegistry.cpp
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/FunctionSignature.h"
+
+namespace facebook::velox::expression {
+
+void ExprRewriteRegistry::registerRewrite(const ExpressionRewrite rewrite) {
+  registry_.withWLock([&](auto& list) { list.push_back(std::move(rewrite)); });
```
kagamiori, line 22: Hi @pramodsatya, linter complains about this line because std::move() won't take effect since rewrite is const. Could you make a fix, either make rewrite non-const or remove std::move and pass rewrite by reference?

pramodsatya: Thanks @kagamiori, removed const qualifier for rewrite.

kagamiori:
Hi @pramodsatya, there is a new conflict preventing merging. Could you please rebase it onto the latest main? Thank you!

pramodsatya:
> Hi @pramodsatya, there is a new conflict preventing merging. Could you please rebase it onto the latest main? Thank you!

Rebased @kagamiori, thanks!

PR #14661, opened 2025-08-30
fix: Remove spammy log message from TableScan

mbasmanova:
Summary: Looks like this was introduced by accident in https://github.com/facebookincubator/velox/pull/14647

Differential Revision: D81392041


amitkdutta (approved):
Thanks @mbasmanova

PR #14665, opened 2025-09-01
fix(hdfs): Add close() method to disconnect the hdfs connection

JkSelf:
The HdfsFileSystem::Disconnect operation requires the JVM method definitions to be loaded within an active JVM process. Therefore, it must be invoked before the JVM shuts down.

To address this, we’ve introduced a new close() API that performs the
disconnect operation. Third-party applications, such as Gluten, can call this close() method prior to JVM termination to ensure proper cleanup.

### DIFF
--- a/velox/common/file/FileSystems.h
+++ b/velox/common/file/FileSystems.h
@@ -179,6 +179,8 @@ class FileSystem {
       std::string_view path,
       const FileOptions& options = {}) = 0;
 
+  virtual void close() {}
+
   /// Deletes the file at 'path'. Throws on error.
   virtual void remove(std::string_view path) = 0;
 
--- a/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp
@@ -55,12 +55,25 @@ class HdfsFileSystem::Impl {
         driver_->GetLastExceptionRootCause());
   }
 
-  ~Impl() {
-    LOG(INFO) << "Disconnecting HDFS file system";
-    int disconnectResult = driver_->Disconnect(hdfsClient_);
-    if (disconnectResult != 0) {
-      LOG(WARNING) << "hdfs disconnect failure in HdfsReadFile close: "
-                   << errno;
+  ~Impl() {}
+
+  // The HdfsFileSystem::Disconnect operation requires the JVM method
+  // definitions to be loaded within an active JVM process.
+  //  Therefore, it must be invoked before the JVM shuts down.
+
+  // To address this, we’ve introduced a new close() API that performs the
+  // disconnect operation. Third-party applications can call this close() method
+  // prior to JVM termination to ensure proper cleanup.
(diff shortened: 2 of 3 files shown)


JkSelf:

velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp
```diff
@@ -55,12 +55,25 @@ class HdfsFileSystem::Impl {
...
+  ~Impl() {}
+
+  // The HdfsFileSystem::Disconnect operation requires the JVM method
+  // definitions to be loaded within an active JVM process.
+  //  Therefore, it must be invoked before the JVM shuts down.
+
+  // To address this, we’ve introduced a new close() API that performs the
+  // disconnect operation. Third-party applications can call this close() method
+  // prior to JVM termination to ensure proper cleanup.
+  void close() {
```
jinchengchenghh, line 72: Could you also check the close status in ~Impl()? if not closed, call the close.

JkSelf: @jinchengchenghh Thanks for your review.

The JVM process has already exited when ~Impl() is called.
Therefore, calling close() within ~Impl() still results in the same [issue](https://github.com/apache/incubator-gluten/issues/8072).

jinchengchenghh: Could we add a log warning in ~Impl() when it is not closed.

JkSelf: @jinchengchenghh Updated.

JkSelf:

velox/common/file/FileSystems.h
```diff
@@ -179,6 +179,8 @@ class FileSystem {
       std::string_view path,
       const FileOptions& options = {}) = 0;
 
+  virtual void close() {}
```
majetideepak: Can we not implement this in the destructor of the FileSystem?

JkSelf: @majetideepak Thanks for your review. I’m afraid that won’t work in our case. We need to explicitly call the close() method before the JVM terminates. If we rely on the FileSystem destructor to handle this, we may encounter a core dump [issue](https://github.com/apache/incubator-gluten/issues/8072) because the JVM might have already exited by the time the destructor is invoked.

JkSelf:
@majetideepak Could you help to review again? Thanks.

JkSelf:
@majetideepak Could you help to review again? Thanks.

majetideepak:
@JkSelf Can we add `close()` only to the HdfsFileSystem class and not the base class? Are you able to work with that?

JkSelf:
> @JkSelf Can we add `close()` only to the HdfsFileSystem class and not the base class? Are you able to work with that?

@majetideepak Yes. Add the `close()` method only in HdfsFileSystem. Please help to review again. Thanks.

PR #14666, opened 2025-09-01
fix(hdfs): Nullify hdfsFile_ before checking success status

boneanxs:
driver_->CloseFile can fail if the task is interrupted. In that case, hdfsFile_ is not reset to nullptr. Later, when the HdfsWriteFile destructor runs, it invokes close() again, resulting in a double close. This leads to a fatal JNI error:

```java
25/08/25 08:12:06 INFO [dispatcher-Executor] Executor: Executor is trying to kill task 43.2 in stage 11.0 (TID 617), reason: another attempt succeeded
HdfsWriteFile.cpp:57, Function:close, Expression: success == 0 (-1 vs. 0) Failed to close hdfs file: IOException: Failed to shutdown streamer, Source: RUNTIME, ErrorCode: INVALID_STATE
```
Eventually, the JVM crashes with a segmentation fault during a JNI call to FSDataOutputStream.close():

```java
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fa2ec7e0c48, pid=174755, tid=0x00007fa220262700
#
# JRE version: OpenJDK Runtime Environment (8.0_252-b09) (build 1.8.0_252-b09)
# Java VM: OpenJDK 64-Bit Server VM (25.252-b09 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.so+0x68fc48]  jni_invoke_nonstatic(JNIEnv_*, JavaValue*, _jobject*, JNICallType, _jmethodID*, JNI_ArgumentPusher*, Thread*)+0x48
#
```

Stack trace excerpt showing the duplicate close() call path:
```c++
#8  invokeMethodOnJclass(..., "org/apache/hadoop/fs/FSDataOutputStream", "close", "()V")
#10 hdfsCloseFile(...)
#11 facebook::velox::HdfsWriteFile::close()
#12 facebook::velox::HdfsWriteFile::~HdfsWriteFile()
#14 facebook::velox::dwio::common::WriteFileSink::~WriteFileSink()
#15 facebook::velox::parquet::Writer::abort()
```

### DIFF
--- a/velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
@@ -54,6 +54,8 @@ HdfsWriteFile::~HdfsWriteFile() {
 
 void HdfsWriteFile::close() {
   int success = driver_->CloseFile(hdfsClient_, hdfsFile_);
+  common::testutil::TestValue::adjust(
+      "facebook::velox::connectors::hive::HdfsWriteFile::close", &success);
   hdfsFile_ = nullptr;
   VELOX_CHECK_EQ(
       success,
--- a/velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
@@ -54,12 +54,12 @@ HdfsWriteFile::~HdfsWriteFile() {
 
 void HdfsWriteFile::close() {
   int success = driver_->CloseFile(hdfsClient_, hdfsFile_);
+  hdfsFile_ = nullptr;
   VELOX_CHECK_EQ(
       success,
       0,
       "Failed to close hdfs file: {}",
       driver_->GetLastExceptionRootCause());
-  hdfsFile_ = nullptr;
 }
 
 void HdfsWriteFile::flush() {
--- a/velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
@@ -522,7 +522,7 @@ TEST_F(HdfsFileSystemTest, readFailures) {
   verifyFailures(driver, hdfs);
 }
 
-TEST_F(HdfsFileSystemTest, writeFilePreventsDoubleClose) {
+DEBUG_ONLY_TEST_F(HdfsFileSystemTest, writeFilePreventsDoubleClose) {
   common::testutil::TestValue::enable();
 
   int closeCallCount = 0;
@@ -536,11 +536,9 @@ TEST_F(HdfsFileSystemTest, writeFilePreventsDoubleClose) {
         }
       }));
 
-  const std::string_view path = "/test_double_close.txt";
-  auto writeFile = openFileForWrite(path);
+  auto writeFile = openFileForWrite("/test_double_close.txt");
 
-  const std::string_view data = "test data";
-  writeFile->append(data);
+  writeFile->append("test data");
   writeFile->flush();
 
   VELOX_ASSERT_THROW(writeFile->close(), "Failed to close hdfs file:");
--- a/velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
@@ -21,6 +21,7 @@
 #include "gtest/gtest.h"
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/common/testutil/TestValue.h"
 #include "velox/connectors/hive/storage_adapters/hdfs/HdfsReadFile.h"
 #include "velox/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.h"
 #include "velox/connectors/hive/storage_adapters/hdfs/tests/HdfsMiniCluster.h"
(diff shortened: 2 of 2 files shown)


boneanxs:
@rui-mo @majetideepak Could you please review this, thanks!

rui-mo:
@JkSelf Would you like to take a review? Thanks.

boneanxs:
> Thanks. Looks good. Would it be possible to use a UT to simulate the failure?

Hey @rui-mo, it looks tricky to make `driver_->CloseFile` return an error, The only approach I can think of is introducing a MockLibHdfsShim. Do you think testing it that way would be okay?

velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
```diff
@@ -54,12 +54,12 @@ HdfsWriteFile::~HdfsWriteFile() {
 
 void HdfsWriteFile::close() {
   int success = driver_->CloseFile(hdfsClient_, hdfsFile_);
+  hdfsFile_ = nullptr;
```
rui-mo, line 59: Based on current change, if `driver_->CloseFile` fails with exception, would the `hdfsFile_ = nullptr` be executed?

rui-mo:
> > Thanks. Looks good. Would it be possible to use a UT to simulate the failure?
> 
> Hey @rui-mo, it looks tricky to make `driver_->CloseFile` return an error, The only approach I can think of is introducing a MockLibHdfsShim. Do you think testing it that way would be okay?

One way to test this is to use `SCOPED_TESTVALUE_SET` to adjust the `success` code, which can then be tested in debug mode after `TestValue::enable()`.

velox/connectors/hive/storage_adapters/hdfs/HdfsWriteFile.cpp
boneanxs: No, if it fails with exception, the `hdfsFile_` won't be reset, but it looks the method rarely throw exceptions, unless the jvm is stopped and env is freed.

```c
int hdfsCloseFile(hdfsFS fs, hdfsFile file)
{
    int ret;
    // JAVA EQUIVALENT:
    //  file.close 

    //The interface whose 'close' method to be called
    CachedJavaClass cachedJavaClass;
    const char *interfaceShortName;

    //Caught exception
    jthrowable jthr;

    //Get the JNIEnv* corresponding to current thread
    JNIEnv* env = getJNIEnv();
    if (env == NULL) {
        errno = EINTERNAL;
        return -1;
    }

    //Sanity check
    if (!file || file->type == HDFS_STREAM_UNINITIALIZED) {
        errno = EBADF;
        return -1;
    }

    if (file->type == HDFS_STREAM_INPUT) {
        cachedJavaClass = JC_FS_DATA_INPUT_STREAM;
    } else {
        cachedJavaClass = JC_FS_DATA_OUTPUT_STREAM;
    }
  
    jthr = invokeMethod(env, NULL, INSTANCE, file->file,
            cachedJavaClass, "close", "()V");
    if (jthr) {
        interfaceShortName = (file->type == HDFS_STREAM_INPUT) ? 
            "FSDataInputStream" : "FSDataOutputStream";
        ret = printExceptionAndFree(env, jthr, PRINT_EXC_ALL,
                "%s#close", interfaceShortName);
    } else {
        ret = 0;
    }

    //De-allocate memory
    (*env)->DeleteGlobalRef(env, file->file);
    free(file);

    if (ret) {
        errno = ret;
        return -1;
    }
    return 0;
}
```

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
```diff
@@ -520,3 +521,36 @@ TEST_F(HdfsFileSystemTest, readFailures) {
       std::string(miniCluster->nameNodePort()));
   verifyFailures(driver, hdfs);
 }
+
+TEST_F(HdfsFileSystemTest, writeFilePreventsDoubleClose) {
```
rui-mo: I have an impression that `SCOPED_TESTVALUE_SET` only takes effect on the debug code. Do we need to write this test as `DEBUG_ONLY_TEST_F`?

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
```diff
@@ -520,3 +521,36 @@ TEST_F(HdfsFileSystemTest, readFailures) {
...
+  SCOPED_TESTVALUE_SET(
+      "facebook::velox::connectors::hive::HdfsWriteFile::close",
+      std::function<void(int*)>([&closeCallCount](int* success) {
+        ++closeCallCount;
+        if (closeCallCount == 1) {
+          *success = -1;
+        }
+      }));
+
+  const std::string_view path = "/test_double_close.txt";
```
rui-mo: Do we need this variable as it's used once?

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
```diff
@@ -520,3 +521,36 @@ TEST_F(HdfsFileSystemTest, readFailures) {
...
+  auto writeFile = openFileForWrite(path);
+
+  const std::string_view data = "test data";
```
rui-mo: ditto

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
boneanxs: sure

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
```diff
@@ -520,3 +521,34 @@ TEST_F(HdfsFileSystemTest, readFailures) {
...
+  VELOX_ASSERT_THROW(writeFile->close(), "Failed to close hdfs file:");
+
+  EXPECT_EQ(closeCallCount, 1);
+
+  // Destructor should not call close() again because hdfsFile_ is nullptr
+  // The closeCallCount should remain 1.
+  writeFile.reset();
+  EXPECT_EQ(closeCallCount, 1);
+
+  common::testutil::TestValue::disable();
```
rui-mo, line 553: Can we also verify that `hdfsFile_` has been set as nullptr after the exception is thrown?

rui-mo (approved):
Thanks! Just one minor suggestion.

velox/connectors/hive/storage_adapters/hdfs/tests/HdfsFileSystemTest.cpp
boneanxs: `writeFile.reset();` should ensure it, since it firstly check if `hdfsFile_` is available or not.

```c++
HdfsWriteFile::~HdfsWriteFile() {
  if (hdfsFile_) {
    close();
  }
}
```

boneanxs:
Hey @majetideepak can you help review this, thanks!

PR #14670, opened 2025-09-01
feat: Support Iceberg clustered writer

PingLiuPing:
There are two writer mode supported by iceberg, one is fanout (default), the other one clustered.
This PR implements clustered writer mode.
When using clustered input mode, the input data is assumed to be clustered/partitioned beforehand.

### DIFF
--- a/velox/connectors/hive/iceberg/IcebergDataSink.cpp
+++ b/velox/connectors/hive/iceberg/IcebergDataSink.cpp
@@ -31,6 +31,9 @@ namespace facebook::velox::connector::hive::iceberg {
 
 namespace {
 
+constexpr std::string_view kNotClusteredRowsErrorMsg =
+    "Incoming records violate the writer assumption that records are clustered by spec and \n by partition within each spec. Either cluster the incoming records or switch to fanout writers.\nEncountered records that belong to already closed files:\n";
+
 #define WRITER_NON_RECLAIMABLE_SECTION_GUARD(index)       \
   memory::NonReclaimableSectionGuard nonReclaimableGuard( \
       writerInfo_[(index)]->nonReclaimableSectionHolder.get())
@@ -208,7 +211,10 @@ IcebergDataSink::IcebergDataSink(
                     insertTableHandle->columnTransforms(),
                     hiveConfig->isPartitionPathAsLowerCase(
                         connectorQueryCtx->sessionProperties()))
-              : nullptr) {
+              : nullptr),
+      fanoutEnabled_(
+          hiveConfig_->fanoutEnabled(connectorQueryCtx_->sessionProperties())),
+      currentWriterId_(0) {
   if (isPartitioned()) {
     partitionData_.resize(maxOpenWriters_);
   }
@@ -325,8 +331,6 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
 }
 
 void IcebergDataSink::splitInputRowsAndEnsureWriters(RowVectorPtr input) {
-  VELOX_CHECK(isPartitioned());
-
   std::fill(partitionSizes_.begin(), partitionSizes_.end(), 0);
 
   const auto numRows = partitionIds_.size();
@@ -339,26 +343,7 @@ void IcebergDataSink::splitInputRowsAndEnsureWriters(RowVectorPtr input) {
     if (!partitionData_[index].empty()) {
       continue;
     }
-
-    std::vector<folly::dynamic> partitionValues(partitionChannels_.size());
-    auto icebergPartitionIdGenerator =
-        dynamic_cast<const IcebergPartitionIdGenerator*>(
-            partitionIdGenerator_.get());
-    VELOX_CHECK_NOT_NULL(icebergPartitionIdGenerator);
-    const RowVectorPtr transformedValues =
-        icebergPartitionIdGenerator->partitionValues();
-    for (auto i = 0; i < partitionChannels_.size(); ++i) {
-      auto block = transformedValues->childAt(i);
-      if (block->isNullAt(index)) {
-        partitionValues[i] = nullptr;
-      } else {
-        DecodedVector decoded(*block);
-        partitionValues[i] = VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
-            extractPartitionValue, block->typeKind(), &decoded, index);
-      }
-    }
-
-    partitionData_[index] = partitionValues;
+    buildPartitionData(index);
   }
 
   for (auto i = 0; i < partitionSizes_.size(); ++i) {
@@ -369,6 +354,11 @@ void IcebergDataSink::splitInputRowsAndEnsureWriters(RowVectorPtr input) {
   }
 }
 
+void IcebergDataSink::computePartition(const RowVectorPtr& input) {
+  VELOX_CHECK(isPartitioned());
+  partitionIdGenerator_->run(input, partitionIds_);
+}
+
 void IcebergDataSink::appendData(RowVectorPtr input) {
   checkRunning();
   if (!isPartitioned()) {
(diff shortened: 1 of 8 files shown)


velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -369,30 +354,92 @@ void IcebergDataSink::splitInputRowsAndEnsureWriters(RowVectorPtr input) {
...
 
+void IcebergDataSink::computePartition(const RowVectorPtr& input) {
+  VELOX_CHECK(isPartitioned());
+  partitionIdGenerator_->run(input, partitionIds_);
+}
+
 void IcebergDataSink::appendData(RowVectorPtr input) {
   checkRunning();
-  if (!isPartitioned() || partitionIdGenerator_->numPartitions() == 1) {
+  if (!isPartitioned()) {
```
jinchengchenghh, line 364: Why remove partitionIdGenerator_->numPartitions() == 1

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -369,30 +354,92 @@ void IcebergDataSink::splitInputRowsAndEnsureWriters(RowVectorPtr input) {
...
+    std::fill(partitionSizes_.begin(), partitionSizes_.end(), 0);
+    const auto numRows = input->size();
+    uint32_t index = 0;
+    for (auto row = 0; row < numRows; ++row) {
+      auto id = getIcebergWriterId(row);
+      index = ensureWriter(id);
+      if (currentWriterId_ != index) {
+        clusteredWrite(input, currentWriterId_);
+        closeWriter(currentWriterId_);
+        completedWriterId_.insert(currentWriterId_);
```
jinchengchenghh: Why reserve completedWriterId_?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -463,9 +510,11 @@ void IcebergDataSink::closeInternal() {
 
   if (state_ == State::kClosed) {
     for (int i = 0; i < writers_.size(); ++i) {
-      WRITER_NON_RECLAIMABLE_SECTION_GUARD(i);
-      writers_[i]->close();
-      dataFileStats_.push_back(writers_[i]->dataFileStats());
+      if (writers_[i]) {
```
jinchengchenghh, line 513: cluster write mode does not need to reserve the writers_ list.

velox/connectors/hive/iceberg/IcebergDataSink.h
```diff
@@ -127,6 +139,11 @@ class IcebergDataSink : public HiveDataSink {
       std::vector<std::unique_ptr<dwio::common::DataFileStatsSettings>>>
       statsSettings_;
   std::unique_ptr<DataFileStatsCollector> icebergStatsCollector_;
+
+  // Below are structures for clustered mode writer.
+  const bool fanoutEnabled;
```
jinchengchenghh: const bool fanoutEnabled_

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: `partitionIdGenerator_->numPartitions() == 1` does not work here. It only work after calling `partitionIdGenerator->run()`. And for iceberg, even there is only one partition we need build the partition data and create the special directory name.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Suppose this is a identity partition. And the input column value is [1,2,3,1]. By simply checking the current partition ID 1 is different with previous one 3 can not guarantee the clustered input.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Yes, you are right, I tried to re-use the same data structure and re-use the same index `i` here. I do implement a first version by only keep a single writer in the writers_ array but it is complicated than this version.

velox/connectors/hive/iceberg/IcebergDataSink.h
PingLiuPing: thanks

velox/connectors/hive/iceberg/IcebergDataSink.cpp
jinchengchenghh: Looks like here is a bug before this PR, please port it to first PR

velox/connectors/hive/iceberg/IcebergDataSink.h
```diff
@@ -0,0 +1,149 @@
...
+  std::vector<std::shared_ptr<dwio::common::DataFileStatistics>> dataFileStats_;
+  std::shared_ptr<
+      std::vector<std::unique_ptr<dwio::common::DataFileStatsSettings>>>
+      statsSettings_;
+  std::unique_ptr<DataFileStatsCollector> icebergStatsCollector_;
+
+  // Below are structures for clustered mode writer.
+  const bool fanoutEnabled_;
+  uint32_t currentWriterId_;
+  std::unordered_set<uint32_t> completedWriterId_;
```
jinchengchenghh: completedWriterId_ -> completedWriterIds_

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Ok, need to move the fix to https://github.com/facebookincubator/velox/pull/13874/files#diff-e816a7cc45e90fc219c6eef1ada327d2575c43dde7f7d8395fd7f5751a9fbff1R290

velox/connectors/hive/iceberg/IcebergDataSink.h
PingLiuPing: thanks

velox/connectors/hive/iceberg/IcebergDataSink.h
```diff
@@ -0,0 +1,149 @@
...
+  std::vector<std::vector<folly::dynamic>> partitionData_;
+
+  std::vector<std::shared_ptr<dwio::common::DataFileStatistics>> dataFileStats_;
+  std::shared_ptr<
+      std::vector<std::unique_ptr<dwio::common::DataFileStatsSettings>>>
+      statsSettings_;
+  std::unique_ptr<DataFileStatsCollector> icebergStatsCollector_;
+
+  // Below are structures for clustered mode writer.
+  const bool fanoutEnabled_;
```
jinchengchenghh, line 144: Move const member upfront

velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/connectors/hive/TableHandle.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+struct IcebergNestedField {
```
yingsu00, line 22: ColumnIdentity. Both Presto and Trino are using this name.

yingsu00, line 22: This is part of the ColumnHandle and need to support serialization.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -0,0 +1,630 @@
...
+          hiveConfig_->fanoutEnabled(connectorQueryCtx_->sessionProperties())),
+      currentWriterId_(0) {
+  if (isPartitioned()) {
+    partitionData_.resize(maxOpenWriters_);
+  }
+  const auto& inputColumns = insertTableHandle_->inputColumns();
+
+  std::function<IcebergDataFileStatsSettings(
+      const IcebergNestedField&, const TypePtr&, bool)>
+      buildNestedField = [&](const IcebergNestedField& f,
```
yingsu00, line 225: Would it be possible to make buildNestedField a private function instead of a lambda here? This is hard to read.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -0,0 +1,630 @@
...
+                             const TypePtr& type,
+                             bool skipBounds) -> IcebergDataFileStatsSettings {
+    VELOX_CHECK_NOT_NULL(type, "Input column type cannot be null.");
+    bool currentSkipBounds = skipBounds || type->isMap() || type->isArray();
+    IcebergDataFileStatsSettings field(f.id, currentSkipBounds);
```
yingsu00, line 230: The name "field" is confusing. Can we just name it statsSettings?

majetideepak:
@PingLiuPing Can you add more details on the iceberg clustered writes?

> When using clustered input mode, the input data is assumed to be clustered/partitioned beforehand.

In the current implementation, who handles the clustering and partitioning?

PingLiuPing:
@majetideepak , thanks for the comment.

> In the current implementation, who handles the clustering and partitioning?

It should be processed by the caller to guarantee the input data is partitioned first. For example,  for identity partition transform (same with Hive), and suppose the partition column type is integer. The valid input stream is 1,1,1,2,2,2,3,3,3. But if the input stream is 1,1,1,2,2,1, then the cluster mode will report error.

majetideepak:
I now see that this PR corresponds to the last commit and the other commits are part of other PRs.

velox/connectors/hive/iceberg/IcebergColumnHandle.h
PingLiuPing: @yingsu00  Thanks for the comments. This PR contains 7 commits but only the last commit is for clustered writer. I have to do this to pass the CI because it depends the previous commits but they are not merged.
Sorry for the confuse.

PingLiuPing:
@yingsu00  Thanks for the comments, I will address your comments when split these huge PRs to smaller pieces.

PR #14671, opened 2025-09-02
Add API to MapVector that projects keys into a RowVector

bikramSingh91:
Summary: Add an API to MapVector that accepts a vector of type MAP and a list of `keysToProject`, and returns a RowVector where each field corresponds to a key from `keysToProject`, with field names specified by `outputFieldNames`. For each row in the MapVector, the function extracts the values associated with the specified keys and assigns them to the corresponding fields in the output RowVector. If the map is null, a key is missing, or the value is null, the output field will also be null. Also accepts a boolean `replaceNulls` which when set to true, replaces these nulls with default values appropriate for the value type. Note that `replaceNulls` cannot be used when the value type is ROW or UNKNOWN. Only keys of types SMALLINT, INTEGER, and BIGINT are supported.

Differential Revision: D81465416


mbasmanova:
> default values appropriate for the value type

@bikramSingh91 Would you clarify what these are?

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,39 @@ class MapVector : public ArrayVectorBase {
...
+  /// For each row in the MapVector, the function extracts the values associated
+  /// with the specified keys and assigns them to the corresponding fields in
+  /// the output RowVector. If the map is null, a key is missing, or the value
+  /// is null, the output field will also be null. If 'replaceNulls' is set to
+  /// true, these nulls are replaced with default values appropriate for the
+  /// value type.
+  ///
+  /// Note:
+  /// - 'replaceNulls' cannot be used when the value type is ROW.
+  /// - Only keys of types SMALLINT, INTEGER, and BIGINT are supported.
```
mbasmanova: Why is this restriction?

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,39 @@ class MapVector : public ArrayVectorBase {
...
+  ///                          in the output RowVector.
+  /// @param replaceNulls      If true, replaces null output values with
+  ///                          type-specific defaults (not supported for ROW
+  ///                          or UNKNOWN types).
+  /// @return                  A RowVector with fields for each key in
+  ///                          keysToProject, named as specified in
+  ///                          outputFieldNames.
+  static VectorPtr toRowVector(
+      const BaseVector& map,
+      const std::vector<variant>& keysToProject,
```
mbasmanova: variant -> Variant

wondering if there is a way to avoid using Variant altogether

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,39 @@ class MapVector : public ArrayVectorBase {
...
+      const std::vector<std::string>& outputFieldNames,
+      bool replaceNulls,
+      const SelectivityVector& rows);
```
mbasmanova: Would you clarify what happens to rows that are not selected?

velox/vector/tests/VectorTest.cpp
```diff
@@ -4224,5 +4224,354 @@ TEST_F(VectorTest, estimateFlatSize) {
   // Test that the second call to prepareForReuse will not cause crash
   arrayVector->prepareForReuse();
 }
+
+class MapVectorToRowVectorTest : public testing::Test,
```
mbasmanova: Let's extract this test into its own file.

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,39 @@ class MapVector : public ArrayVectorBase {
...
+  /// @param outputFieldNames  The names to assign to the corresponding fields
+  ///                          in the output RowVector.
+  /// @param replaceNulls      If true, replaces null output values with
+  ///                          type-specific defaults (not supported for ROW
+  ///                          or UNKNOWN types).
+  /// @return                  A RowVector with fields for each key in
+  ///                          keysToProject, named as specified in
+  ///                          outputFieldNames.
+  static VectorPtr toRowVector(
+      const BaseVector& map,
```
mbasmanova: Would you clarify whether this has to be a MapVector or can be wrapped in Dict or Const?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+} // namespace
+
+VectorPtr MapVector::toRowVector(
+    const BaseVector& map,
+    const std::vector<variant>& keysToProject,
+    const std::vector<std::string>& outputFieldNames,
+    bool replaceNulls,
+    const SelectivityVector& rows) {
+  VELOX_CHECK_EQ(map.type()->kind(), TypeKind::MAP);
+  VELOX_CHECK(!keysToProject.empty());
```
mbasmanova: Would you document this requirement?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+}
+} // namespace
+
+VectorPtr MapVector::toRowVector(
+    const BaseVector& map,
+    const std::vector<variant>& keysToProject,
+    const std::vector<std::string>& outputFieldNames,
+    bool replaceNulls,
+    const SelectivityVector& rows) {
+  VELOX_CHECK_EQ(map.type()->kind(), TypeKind::MAP);
```
mbasmanova: _USER_CHECK ?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  VELOX_CHECK(!keysToProject.empty());
+  VELOX_CHECK_EQ(keysToProject.size(), outputFieldNames.size());
+
+  // Ensure that the key type and variant type are the same.
+  auto keyType = map.type()->asMap().keyType();
+  VELOX_CHECK_EQ(
+      keyType->kind(),
+      keysToProject[0].kind(),
```
mbasmanova: Should we verify all keys?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const std::vector<variant>& keysToProject,
+    const std::vector<std::string>& outputFieldNames,
+    bool replaceNulls,
+    const SelectivityVector& rows) {
+  VELOX_CHECK_EQ(map.type()->kind(), TypeKind::MAP);
+  VELOX_CHECK(!keysToProject.empty());
+  VELOX_CHECK_EQ(keysToProject.size(), outputFieldNames.size());
+
+  // Ensure that the key type and variant type are the same.
+  auto keyType = map.type()->asMap().keyType();
```
mbasmanova: const &

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  VELOX_CHECK_EQ(
+      keyType->kind(),
+      keysToProject[0].kind(),
+      "Key type and the type of keys to project are not the same");
+
+  if (keyType == SMALLINT()) {
```
mbasmanova: Do we support keys of type DATE?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const std::vector<variant>& keysToProject,
+    const std::vector<std::string> outputFieldNames,
+    bool replaceNulls,
+    const SelectivityVector& rows) {
+  using KeyType = typename TypeTraits<KeyKind>::NativeType;
+  folly::F14FastMap<KeyType, size_t> lookup;
+  lookup.reserve(keysToProject.size());
+  for (size_t i = 0; i < keysToProject.size(); ++i) {
+    VELOX_CHECK(
+        !keysToProject[i].isNull(), "Keys to project cannot contain null");
```
mbasmanova: Should we move this check earlier, before type-switch ?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const KeyType& key = keysToProject[i].value<KeyKind>();
+    // Key must be unique.
+    VELOX_CHECK_EQ(lookup.count(key), 0, "Duplicate keys not allowed");
```
mbasmanova: perhaps, include key value in the error message for easier troubleshooting

mbasmanova: avoid double lookup

```
book ok = lookup.insert({key, i}).second;
CHECK(ok);
```

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const std::vector<std::string> outputFieldNames,
+    bool replaceNulls,
+    const SelectivityVector& rows) {
+  using KeyType = typename TypeTraits<KeyKind>::NativeType;
+  folly::F14FastMap<KeyType, size_t> lookup;
+  lookup.reserve(keysToProject.size());
+  for (size_t i = 0; i < keysToProject.size(); ++i) {
+    VELOX_CHECK(
+        !keysToProject[i].isNull(), "Keys to project cannot contain null");
+    const KeyType& key = keysToProject[i].value<KeyKind>();
```
mbasmanova: since key is an integer, no need to use const &; just auto key = ...

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  auto mapKeys = mapBase->mapKeys();
+  auto decodedKeys = DecodedVector(*mapKeys);
+  auto mapValues = mapBase->mapValues();
+  auto decodedValues = DecodedVector(*mapValues);
+  auto offsets = mapBase->rawOffsets();
+  auto sizes = mapBase->rawSizes();
+  vector_size_t outputSize = rows.end();
+
+  for (size_t i = 0; i < keyToIndex.size(); ++i) {
+    children.push_back(
```
mbasmanova: reserve + emplace_back

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+        BaseVector::create(valueType, outputSize, mapBase->pool()));
+    flatChildren.push_back(children.back()->asFlatVector<ValueType>());
+    VELOX_CHECK_NOT_NULL(flatChildren.back());
+  }
+
+  static const ValueType defaultValue = ValueType();
+  std::vector<bool> visited(keyToIndex.size());
+
+  rows.applyToSelected([&](vector_size_t row) INLINE_LAMBDA {
```
mbasmanova: Isn't this lambda a bit large to inline?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const KeyType& key = keysToProject[i].value<KeyKind>();
+    // Key must be unique.
+    VELOX_CHECK_EQ(lookup.count(key), 0, "Duplicate keys not allowed");
+    lookup.insert({key, i});
+  }
+
+  std::vector<VectorPtr> children;
+  auto valueType = vector.type()->asMap().valueType();
+  if (valueType->isPrimitiveType() && valueType->kind() != TypeKind::UNKNOWN) {
+    children = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
```
mbasmanova: Is there significant perf benefit in having this separate path? Curious, how much does it save?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
   bits::fillBits(mutableRawNulls(), oldSize, newSize, bits::kNull);
 }
 
+namespace {
+template <typename KeyType, TypeKind ValueKind>
```
mbasmanova: It looks like ValueKind template parameter is not needed. It is only used to check if replaceNulls is supported or not and that check can happen before calling this function.

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    for (size_t field = 0; field < visited.size(); ++field) {
+      if (!visited[field]) {
+        if (replaceNulls) {
+          if constexpr (
+              ValueKind == TypeKind::ARRAY || ValueKind == TypeKind::MAP) {
+            auto arrayVectorBase =
+                children[field]->asUnchecked<ArrayVectorBase>();
+            arrayVectorBase->setOffsetAndSize(row, 0, 0);
+          } else {
+            VELOX_FAIL(
```
mbasmanova: This branch is not reachable because there is a check at the start of the function.

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    if (replaceNulls) {
+      VELOX_FAIL(
+          "Unsupported type for replacing nulls: {}",
+          TypeTraits<ValueKind>::TypeTraits::name);
+    }
+  }
+  DecodedVector decodedVector(vector);
+  auto mapBase = decodedVector.base()->asUnchecked<MapVector>();
+  auto valueType = mapBase->mapValues()->type();
+  auto mapKeys = mapBase->mapKeys();
```
mbasmanova: mapKeys variable is used only once; consider removing for readability (this block of code is quite dense)

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  auto decodedKeys = DecodedVector(*mapKeys);
```
mbasmanova: nit: consistency

DecodedVector decodedKeys(*mapKeys);

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  auto mapValues = mapBase->mapValues();
+  auto decodedValues = DecodedVector(*mapValues);
+  auto offsets = mapBase->rawOffsets();
```
mbasmanova: auto*

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
   bits::fillBits(mutableRawNulls(), oldSize, newSize, bits::kNull);
 }
 
+namespace {
+template <typename KeyType, TypeKind ValueKind>
+std::vector<VectorPtr> toRowVectorComplexImpl(
+    const BaseVector& vector,
+    folly::F14FastMap<KeyType, size_t> keyToIndex,
```
mbasmanova: nit: perhaps, add a @param keyToIndex to clarify that 'index' is unique and comes from [0, keyToIndex.size()) range.

pass map by const &

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  auto mapValues = mapBase->mapValues();
+  auto decodedValues = DecodedVector(*mapValues);
+  auto offsets = mapBase->rawOffsets();
+  auto sizes = mapBase->rawSizes();
+  vector_size_t outputSize = rows.end();
+  std::vector<bool> visited(keyToIndex.size());
+
+  std::vector<VectorPtr> children;
+  children.reserve(keyToIndex.size());
+  std::vector<std::vector<BaseVector::CopyRange>> copyRangesFromBase(
```
mbasmanova: nit: move this variable closer to where it is used

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+      keyToIndex.size());
+  for (size_t i = 0; i < keyToIndex.size(); ++i) {
+    children.push_back(
+        BaseVector::create(valueType, outputSize, mapBase->pool()));
+  }
+
+  rows.applyToSelected([&](vector_size_t row) INLINE_LAMBDA {
+    std::fill(visited.begin(), visited.end(), false);
+    if (!decodedVector.isNullAt(row)) {
```
mbasmanova: decodedVector -> decodedMap

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+      auto decodedIndex = decodedVector.index(row);
+      auto offset = offsets[decodedIndex];
+      auto size = sizes[decodedIndex];
+      for (auto j = offset; j < offset + size; ++j) {
+        if (decodedKeys.isNullAt(j) || decodedValues.isNullAt(j)) {
```
mbasmanova: Why do we skip map entries where value is null?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+
+    // Fill missing keys in the row with null or default values.
+    for (size_t field = 0; field < visited.size(); ++field) {
+      if (!visited[field]) {
+        if (replaceNulls) {
+          if constexpr (
+              ValueKind == TypeKind::ARRAY || ValueKind == TypeKind::MAP) {
+            auto arrayVectorBase =
+                children[field]->asUnchecked<ArrayVectorBase>();
+            arrayVectorBase->setOffsetAndSize(row, 0, 0);
```
mbasmanova: Can we do that in bulk before rows.applyToSelected loop?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+        if (decodedKeys.isNullAt(j) || decodedValues.isNullAt(j)) {
+          continue;
+        }
+        auto key = decodedKeys.valueAt<KeyType>(j);
+        auto it = keyToIndex.find(key);
+        if (it == keyToIndex.end()) {
+          continue;
+        }
+        auto index = it->second;
+        visited[index] = true;
```
mbasmanova: Should we check that visited[index] is not already true (e.g. fail if there are duplicate map keys)?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,266 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+        break;
+      default:
+        VELOX_FAIL("Unsupported type: {}", valueType->kindName());
+    }
+  }
+
+  std::vector<TypePtr> outputChildTypes(outputFieldNames.size(), valueType);
+  return std::make_shared<RowVector>(
+      vector.pool(),
+      ROW(outputFieldNames, std::move(outputChildTypes)),
```
mbasmanova: Use ROW(outputFieldNames, valueType) to create homogeneous struct type

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,39 @@ class MapVector : public ArrayVectorBase {
   std::shared_ptr<MapVector> update(
       const folly::Range<DecodedVector*>& others) const;
 
+  /// Retrieves the specified keys and their associated values from each row in
+  /// a MapVector, returning a RowVector where each field corresponds to a key
+  /// and contains its respective value.
+  ///
+  /// For each row in the MapVector, the function extracts the values associated
+  /// with the specified keys and assigns them to the corresponding fields in
+  /// the output RowVector. If the map is null, a key is missing, or the value
```
mbasmanova: > If the map is null

I'd expect the output struct to be null, not have all null fields. Is there a reason for the currently-implemented semantic?

velox/vector/ComplexVector.h
bikramSingh91: > wondering if there is a way to avoid using Variant altogether

using a variant was the only way I could think of making the API type agnostic and avoid templates like other Vector APIs.
Open to suggestions!

mbasmanova: ConstantVector is another way.

velox/vector/ComplexVector.h
bikramSingh91: As discussed offline, the key types are limited to this as the current use case only required these types. The current implementation can be extended trivially to add support for other primitive types, we would just need to add additional testing for it. I'll add a TODO for now and get back to extending it once I have more bandwidth.

velox/vector/ComplexVector.h
bikramSingh91: A ConstantVector might be too heavy weight to build:
For eg: 
- A ConstantVector<ArrayVector<T>> with a single row: This will involve creating an ArrayVector with single row (allocating buffers for offsets, sizes and a flatVector to hold the values) the wrapping it in a Constant encoding layer. 

How about:
- A FlatVector<T> with rows equal to the number of keys we need to project: this will include allocating a single bufferPtr. Then in the future we can also pass a ComplexVector to add support for complex keys

Note: In both cases, we would have to pass a VectorPtr to avoid templating the API. Not a big deal though.

Just curious, why would we want to avoid using a variant? do we want to avoid intertwining variants with the Vector classes? or is creating a variant heavyweight?

velox/vector/ComplexVector.cpp
bikramSingh91: they would be set to null/default value at the end

velox/vector/ComplexVector.cpp
bikramSingh91: This is an expectation of pre-proc, ideally there should not be any duplicate map keys, but to in order to make this least disruptive I kept this expectation. Let me add a boolean like `throwOnDuplicateKeys` to make this optional

velox/vector/ComplexVector.cpp
bikramSingh91: yup, since we are creating new vectors altogether, we can also just set nulls to all in case replace nulls is false. Same for primitive types. will make the change in both

velox/vector/ComplexVector.cpp
bikramSingh91: removed

velox/vector/ComplexVector.cpp
bikramSingh91: would this prevent inlining? is there anyway I can optimize the code here?

velox/vector/ComplexVector.cpp
bikramSingh91: moved

velox/vector/ComplexVector.cpp
bikramSingh91: I havent done a thorough analysis on this, but here is what I thought:
If I do both primitive and complex types the same way, that is, create an array of copyRanges for each child, then for primitive types I'll end up allocating [size(CopyRange) = 3 * sizeof(int32) = 12 bytes] * vector->size() memory whereas the values() buffer of each would actually be smaller in most cases. Also creating a new range will be ~ to assigning a value to the flat vector. So will save a bunch of memory and CPU cycles. 

let me know what you think

velox/vector/ComplexVector.cpp
bikramSingh91: Not yet, added a TODO

velox/vector/ComplexVector.h
bikramSingh91: this is an expectation of pre-proc, added a boolean flag to toggle this behavior.

bikramSingh91:
@mbasmanova  Addressed all comments, will add additional testing for newly added options (allowTopLevelNulls and throwOnDuplicateKeys) in the next revision

velox/vector/ComplexVector.h
mbasmanova: Variant was never designed to be used for production code paths. It is not integrated with memory management. It cannot be moved cheaply unless wrapped in a shared_ptr. It is very inefficient for complex types. A Variant array stores each element wrapped in a Variant. Extracting values from non-primitive Variant is also cumbersome and expensive.

> A FlatVector with rows equal to the number of keys we need to project

That sound good to me. Can we do that? This would be one of Flat/Array/Map/RowVector passed as VectorPtr, right? This way it can specify keys of both primitive and complex types.

velox/vector/ComplexVector.cpp
mbasmanova: I'm wary of adding significant amount of code without providing evidence that it significantly improves performance. Would it be possible to add a benchmark that simulates production use cases we are aware of and checking what's the difference between having fast path and not? Would it be possible to also run a few prod queries w/ and w/o fast path to check how much difference it makes e2e? 

It would also be Ok to provide a single path initially. Then, work on benchmarks and optionally fast path in a follow-up.

Thanks.

bikramSingh91:
@mbasmanova Addressed the remaining comments. Please take a look when you get a chance.

Highlights of the latest revision:
- Consolidated complex types and primitive types paths into one.
- Switched from using Variant to VectorPtr for `keysToProject`
- Added testing for newly added options (allowTopLevelNulls and throwOnDuplicateKeys)

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
   std::shared_ptr<MapVector> update(
       const folly::Range<DecodedVector*>& others) const;
 
+  struct ToRowVectorOptions {
+    // The list of keys to extract from each map. The type should be the same as
```
mbasmanova: > The type should be the same as the value type in the map.

typo: value -> key ?

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+    // the value type in the map. Should be the same size as 'outputFieldNames'.
+    // Cannot be empty.
+    VectorPtr keysToProject;
```
mbasmanova: perhaps, also clarify that keys cannot be null

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+
+    // The names to assign to the corresponding fields in the output RowVector.
+    // Should be the same size as 'keysToProject'.
+    std::vector<std::string> outputFieldNames;
```
mbasmanova: perhaps, clarify that names must not be empty and must be unique

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+
+    // When set to false, if the map is null, a key is missing, or the
```
mbasmanova: // -> /// for public docs

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+    // value is null, the output field will also be null.
+    // If set to true, replaces null output values with type-specific defaults
```
mbasmanova: Wondering if it is better to let the caller specify the default value.

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+    // like for primitive types, they will be set to the c++ default initialized
+    // value, like 0 for INTEGER and empty string for VARCHAR. For Array/Map
+    // they will be set to an empty array or map. Not supported for ROW or
```
mbasmanova: > Not supported for ROW

We can remove this restriction if we let the caller provide the default value. There is no reason to disallow default value for ROW, I think.

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+  /// Extracts the specified keys and their corresponding values from each row
+  /// (marked active in 'rows') of a vector of Map type.
+  /// The input MapVector may be wrapped in a dictionary or constant wrapper.
+  /// Returns a RowVector where each field corresponds to one of the specified
+  /// keys, containing its associated value. For each row in the MapVector:
+  ///   - The function looks up the value associated with the specified keys and
+  ///   assigns them to the corresponding fields in the output RowVector
+  ///   - If a key is missing, or the value is null, the
+  ///   corresponding output field will be null.
+  ///   - If 'options.replaceNulls' is true, any null output fields are replaced
```
mbasmanova: wondering if we can avoid duplicating these comments and just refer to options

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+  ///   - If 'options.allowTopLevelNulls' is true, a null map will result in a
+  ///   null row, otherwise the top level row will not be null but every field
+  ///   in the output row will be null.
+  ///   - If 'options.throwOnDuplicateKeys' is true, duplicate keys are not
+  ///   allowed and an exception will be thrown. Otherwise, the last value is
+  ///   used.
+  ///
+  /// Note:
+  /// - Only keys of types SMALLINT, INTEGER, and BIGINT are currently
+  /// supported. TODO: Support other key types.
```
mbasmanova: Let's create a GitHub issue and reference it here.

velox/vector/ComplexVector.h
```diff
@@ -722,6 +722,63 @@ class MapVector : public ArrayVectorBase {
...
+  /// - Rows marked as inactive in the 'rows' selectivity vector will be
```
mbasmanova: > will be considered valid

What does that mean?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    const BaseVector& map,
+    const ToRowVectorOptions& options,
+    const SelectivityVector& rows) {
+  VELOX_USER_CHECK_EQ(map.type()->kind(), TypeKind::MAP);
+  VELOX_USER_CHECK_NOT_NULL(options.keysToProject);
+  VELOX_USER_CHECK_GT(options.keysToProject->size(), 0);
+  VELOX_USER_CHECK_EQ(
+      options.keysToProject->size(), options.outputFieldNames.size());
+
+  // Ensure that the key type and variant type are the same.
```
mbasmanova: typo: variant ?

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  const auto& keyType = map.type()->asMap().keyType();
+  VELOX_USER_CHECK_EQ(
+      keyType,
+      options.keysToProject->type(),
+      "Key type and the type of keys to project are not the same",
+      keyType,
```
mbasmanova: missing placeholders in the error message

velox/vector/ComplexVector.cpp
mbasmanova: If we do not support DATE, then we probably should enhance this check beyond checking TypeKind.

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+      bits::fillBits(rawNulls, 0, outputSize, bits::kNull);
+    } else {
+      VELOX_DYNAMIC_TYPE_DISPATCH_ALL(
+          preFillDefaults, valueType->kind(), children.back());
+    }
+  }
+
+  auto outputNulls = (!options.replaceNulls && options.allowTopLevelNulls &&
+                      decodedMap.mayHaveNulls())
+      ? AlignedBuffer::allocate<bool>(
```
mbasmanova: allocateNulls

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+      bits::setNull(rawNulls, row, true);
+    }
+  });
+  for (size_t i = 0; i < children.size(); ++i) {
+    children[i]->copyRanges(decodedValues.base(), copyRangesFromBase[i]);
+  }
+
+  return std::make_shared<RowVector>(
+      vector.pool(),
+      ROW(options.outputFieldNames, valueType),
```
mbasmanova: let's add a check that names are not empty and unique

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  std::vector<bool> visited(keyToIndex.size());
+  std::vector<std::vector<BaseVector::CopyRange>> copyRangesFromBase(
+      keyToIndex.size());
+  rows.applyToSelected([&](vector_size_t row) INLINE_LAMBDA {
+    std::fill(visited.begin(), visited.end(), false);
+    if (!decodedMap.isNullAt(row)) {
+      auto decodedIndex = decodedMap.index(row);
+      auto offset = offsets[decodedIndex];
+      auto size = sizes[decodedIndex];
+      for (auto j = offset; j < offset + size; ++j) {
```
mbasmanova: j -> i

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+          continue;
+        }
+        auto key = decodedKeys.valueAt<KeyType>(j);
+        auto it = keyToIndex.find(key);
+        if (it == keyToIndex.end()) {
+          continue;
+        }
+        auto index = it->second;
+        if (visited[index]) {
+          VELOX_USER_CHECK(
```
mbasmanova: If this API will be used to implement a UDF, it will need to provide a no-throw behavior.

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+              !options.throwOnDuplicateKeys,
+              "Duplicate keys not allowed: {}",
+              key);
+          VELOX_CHECK(
```
mbasmanova: split this check into 2 for easier troubleshooting

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+    auto* rawOffsets =
+        arrayBaseVector->mutableOffsets(childSize)->asMutable<vector_size_t>();
+    auto* rawSizes =
+        arrayBaseVector->mutableSizes(childSize)->asMutable<vector_size_t>();
+    std::fill(rawOffsets, rawOffsets + childSize, 0);
+    std::fill(rawSizes, rawSizes + childSize, 0);
+  } else {
+    VELOX_USER_FAIL(
+        "Unsupported type for replacing nulls: {}",
+        TypeTraits<ValueKind>::TypeTraits::name);
```
mbasmanova: child->type()->toString() ? or TypeKindName::toName(child->typeKind())

velox/vector/ComplexVector.cpp
```diff
@@ -1781,4 +1782,163 @@ void RowVector::appendNulls(vector_size_t numberOfRows) {
...
+  } else if constexpr (
+      ValueKind == TypeKind::ARRAY || ValueKind == TypeKind::MAP) {
+    auto childSize = child->size();
+    auto arrayBaseVector = child->asChecked<ArrayVectorBase>();
+    VELOX_CHECK_NOT_NULL(arrayBaseVector);
+    auto* rawOffsets =
+        arrayBaseVector->mutableOffsets(childSize)->asMutable<vector_size_t>();
+    auto* rawSizes =
+        arrayBaseVector->mutableSizes(childSize)->asMutable<vector_size_t>();
+    std::fill(rawOffsets, rawOffsets + childSize, 0);
```
mbasmanova: This is the default. Do we need this logic at all?

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+#include "velox/vector/SelectivityVector.h"
+#include "velox/vector/SimpleVector.h"
+#include "velox/vector/TypeAliases.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
+
+namespace facebook::velox {
+namespace {
+
+// For testing Opaque types.
+struct NonPOD {
```
mbasmanova: This seems to be duplicated in multiple tests. Let's extract this into a separate file so it can be reused.

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+          TypeTraits<ValueKind>::TypeTraits::name);
+    }
+  }
+
+  // Helper function that verifies the result of MapVector::toRowVector by
+  // iterating through all rows of the result and looking up its corresponding
+  // value in the original MapVector (by looking up the key for that row), if it
+  // exists. If key is not found, ensures that its set to null or the default
+  // value depending on whether replaceNulls is set.
+  void verifyResult(
```
mbasmanova: This code seems to duplicate the code being tested, which makes it hard to ensure it is correct. Wondering if there is a way to test in a more black-box manner. For example,

make_row_from_map(m, keys, names) is equivalent to cast(row_constructor(m[k1], m[k2], ...) as row(name1, name2,...))

Can we base the test on this invariant?

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+      validMap->mapKeys(),
+      rowValues);
+  VELOX_ASSERT_THROW(
+      MapVector::toRowVector(*mapWithRowValues, options, rows),
+      "Unsupported type for replacing nulls: ROW");
+}
+
+TEST_F(MapVectorToRowVectorTest, duplicateKey) {
+  // Create a map like {1: 1, 2: 2, 2: 3}.
+  auto testCase = makeMapVector<int64_t, int64_t>(
```
mbasmanova: This is a bit hard to read. Can we not use something like

```
makeMapVector({
   {{1, 1}, {2, 2}, {2, 3}},
});
```

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+      .throwOnDuplicateKeys = true,
+      .allowTopLevelNulls = false};
+  SelectivityVector rows(testCase->size());
+  VELOX_ASSERT_THROW(
+      MapVector::toRowVector(*testCase, options, rows),
+      "Duplicate keys not allowed");
+
+  options.throwOnDuplicateKeys = false;
+  auto result = MapVector::toRowVector(*testCase, options, rows);
+  ASSERT_EQ(
```
mbasmanova: Can we provide expectedResult and use assertEqualVectors?

Perhaps, also check that de-selecting row with duplicate key avoids the error as well.

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+  auto validMap = makeMapVectorFromJson<int64_t, int64_t>({"{1:1, 2:2, 3:3}"});
+
+  // Input of typeKind other than Map used
+  MapVector::ToRowVectorOptions options{
+      .keysToProject = {},
+      .outputFieldNames = {},
+      .replaceNulls = false,
+      .throwOnDuplicateKeys = false,
+      .allowTopLevelNulls = false};
+  ASSERT_ANY_THROW(
```
mbasmanova: VELOX_ASSERT_THROW

velox/vector/tests/MapVectorToRowVectorTest.cpp
```diff
@@ -0,0 +1,473 @@
...
+      MapVector::toRowVector(*mapWithStringKeys, options, rows),
+      "Only SMALLINT, INTEGER, BIGINT keys are currently supported, instead got VARCHAR");
+
+  // Mismatched Type between keysToProject and map keys
+  VELOX_ASSERT_THROW(
+      MapVector::toRowVector(*validMap, options, rows),
+      "(BIGINT vs. VARCHAR) Key type and the type of keys to project are not the same");
+
+  // Keys to project are not empty
+  // Mismatched Type between keysToProject and map keys
```
mbasmanova: Comment doesn't seem to match code.

velox/vector/ComplexVector.cpp
bikramSingh91: This is just to ensure, that we do not rely on the default being this way since its not guaranteed as a part of the create API

velox/vector/ComplexVector.cpp
bikramSingh91: this is checking the TypePtr directly instead of the TypeKind, so it will catch anything other than the 3 types. I'll update the NYI message underneath to instead use the type() to make it clearer.

velox/vector/ComplexVector.h
bikramSingh91: Makes sense. That would include passing another VectorPtr like we are doing for the keysToProject.
Would it make sense to add this to the list of TODO improvements (along with support for other keys) to limit the scope of this PR?

velox/vector/ComplexVector.h
bikramSingh91: Updated

velox/vector/ComplexVector.cpp
bikramSingh91: @mbasmanova Does that mean we need a codepath to handle individual rows throwing? and keep moving forward with other rows?
for eg, if a map at a rowIdx has duplicate keys, we mark the whole output row as null/default value?

velox/vector/tests/MapVectorToRowVectorTest.cpp
bikramSingh91: I have added a verification path to this existing one that uses expression eval. However, for simplicity, I have only done it for the path where 'replaceNulls' and 'allowTopLevelNulls' is false. This will help us sanity check the manual verification path that we would need to use to verify the rest of the options. Additionally, we can also verify the path where 'replaceNulls' is true, by generating the `Coalesce` expression for all input key types, but it will be too cumbersome.

Let me know if this approach seems sufficient.

velox/vector/tests/MapVectorToRowVectorTest.cpp
bikramSingh91: copy-paste error, removed, thanks for catching this

velox/vector/ComplexVector.h
mbasmanova: That's fine, although, it would require changing the API. Hence, shouldn't be delayed.

velox/vector/ComplexVector.cpp
mbasmanova: Something like that, but there needs to be a way to distinguish rows with errors vs. null rows and allow to produce reasonable error message if errors are not suppressed.

Wondering if we would need to move away from an API on MapVector into a UDF.

bikramSingh91:
@mbasmanova Please take another look when you get a chance.
Highlights of the change:
- Moved implementation out as a Utility class inside `functions` folder.
- Converted it from a single API call to a stateful class that is templated on the keyType. This is to ensure we dont generate the keyToFieldIndex map on every invocation.
- Added an optional ability to pass EvalCtx so that it can be used both within and outside the context of expression eval. EvalCtx will be used to leverage the vector and decodedVector pool within that context and apply per-row error handling as required by expression evaluation.
- Added corresponding tests: With/Without EvalCtx, additional error paths, per-row error handling.

velox/vector/tests/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ add_executable(
   FlatMapVectorTest.cpp
   IsWritableVectorTest.cpp
   LazyVectorTest.cpp
+  MapVectorToRowVectorTest.cpp
```
mbasmanova: This change doesn't seem right.

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -0,0 +1,280 @@
...
+    std::unordered_set<std::string> fieldNamesSet;
+    for (const auto& fieldName : fieldNames) {
+      VELOX_USER_CHECK(!fieldName.empty(), "Field name cannot be empty");
+      auto ok = fieldNamesSet.insert(fieldName).second;
+      VELOX_USER_CHECK(
+          ok, "Duplicate field names are not allowed: {}", fieldName);
+    }
+  }
+
+  bool replaceNulls_{false};
```
mbasmanova: const

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -0,0 +1,280 @@
...
+                  fmt::format("Duplicate keys not allowed: {}", key);
+              if (evalCtx) {
+                evalCtx->setStatus(row, Status::UserError(errorMessage));
+              } else {
+                VELOX_USER_FAIL(errorMessage);
+              }
+              return;
+            }
+            VELOX_CHECK(!copyRangesFromBase[index].empty());
+            VELOX_CHECK(copyRangesFromBase[index].back().targetIndex == row);
```
mbasmanova: VELOX_CHECK_EQ

PR #14681, opened 2025-09-02
fix: Replace boost::regex with RE2 regex in url_extract_parameter and fb_url_extract_parameter Presto function in Velox

abhinavmuk04:
Summary:
Some implementations of functions in Velox rely on boost::regex as regular expression engine. That is not recommended. RE2 is the preferred engine.

boost relies on backtracking and it can lead to significantly slower performance compared to RE2. RE2 has O(n) complexity for a given string of length n, as it uses finite-state machines

Differential Revision: D81494279


velox/functions/prestosql/URIParser.h
```diff
@@ -16,9 +16,11 @@
 #pragma once
 
 #include <boost/regex.hpp>
+#include <re2/re2.h>
 #include "velox/type/StringView.h"
 
 namespace facebook::velox::functions {
+
 namespace detail {
 FOLLY_ALWAYS_INLINE StringView submatch(const boost::cmatch& match, int idx) {
```
kevinwilfong: It looks like the only place this is still used is in URLFunctions.cpp

Do you plan to update that as well? If not could you move this function there, that way we can get rid of this detail namespace in a header file and the dependency on boost in this file.

velox/functions/prestosql/URIParser.h
```diff
@@ -58,26 +60,35 @@ std::optional<StringView> extractParameter(
...
+          return StringView(matches[3].data(), matches[3].size());
         }
       }
+
+      // Move past this match to continue searching
+      pos = matches[0].end() - input.data();
+      if (pos == matches[0].data() - input.data()) {
+        // Avoid infinite loop on zero-width matches
+        ++pos;
+      }
```
kevinwilfong, line 85: Do we have a test case for this already? If not could you add one, seems like an easy edge case to miss.

velox/functions/prestosql/URLFunctions.cpp
```diff
@@ -15,42 +15,3 @@
  */
 
 #include "URLFunctions.h"
-#include <optional>
```
kevinwilfong, line 18: Nice! This was just dead code?

Can you just delete the file.

PR #14682, opened 2025-09-02
feat(connectors): Pass connector configuration during Connector construction

hdikeman:
Summary:
Previously for some connectors we were overriding the connectorConfig() method and returning a child class configuration field. Instead, we can create a constructor parameter for the connector config object and implement connectorConfig() directly in the interface class

Ran all the velox/connector/ tests to verify this change

Differential Revision: D81501098

### DIFF
--- a/velox/connectors/Connector.h
+++ b/velox/connectors/Connector.h
@@ -564,7 +564,10 @@ class ConnectorMetadata;
 
 class Connector {
  public:
-  explicit Connector(const std::string& id) : id_(id) {}
+  explicit Connector(
+      const std::string& id,
+      std::shared_ptr<const config::ConfigBase> config = nullptr)
+      : id_(id), config_(std::move(config)) {}
 
   virtual ~Connector() = default;
 
@@ -572,9 +575,8 @@ class Connector {
     return id_;
   }
 
-  virtual const std::shared_ptr<const config::ConfigBase>& connectorConfig()
-      const {
-    VELOX_NYI("connectorConfig is not supported yet");
+  const std::shared_ptr<const config::ConfigBase>& connectorConfig() const {
+    return config_;
   }
 
   /// Returns true if this connector would accept a filter dynamically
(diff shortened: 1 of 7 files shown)


velox/connectors/tpch/TpchConnector.cpp
```diff
@@ -35,7 +35,7 @@ TpchConnector::TpchConnector(
     const std::string& id,
     std::shared_ptr<const config::ConfigBase> config,
     folly::Executor* /*executor*/)
-    : Connector(id) {
+    : Connector(id), config_(std::move(config)) {
```
mbasmanova: Any particular reason not to pass config to the base class?

hdikeman: To avoid changing the common Connector constructor interface. The constructor just takes a connector ID at the moment

Given the number of callers, breaking the interface is likely not an option. Could add it with a default nullptr if that is preferable to keeping it as a child class variable?

mbasmanova: > Could add it with a default nullptr if that is preferable

Let's do that, store the config in the base class and use it to implement config getter. We can also change connectorConfig to not be virtual.

hdikeman: Sounds good

velox/connectors/Connector.h
```diff
@@ -564,17 +564,19 @@ class ConnectorMetadata;
...
 
   virtual ~Connector() = default;
 
   const std::string& connectorId() const {
     return id_;
   }
 
-  virtual const std::shared_ptr<const config::ConfigBase>& connectorConfig()
-      const {
-    VELOX_NYI("connectorConfig is not supported yet");
```
mbasmanova, line 577: perhaps, check if config_ is null and throw if so

zhouyuan:
@bikramSingh91 
the CI is red:
https://github.com/facebookincubator/velox/actions/runs/17413820825/job/49437278229

PR #14687, opened 2025-09-02
refactor: Make PlanBuilder connector agnostic part 1

yingsu00:
# What PR #14687 Does

The PR is part of the ongoing decoupling of Hive from Velox exec tests:

## Before: 
Many velox/exec unit tests (for TableScan, TableWriter, Exchange, etc.) directly created HiveTableHandle, HiveColumnHandle, or HivePartitionFunctionSpec. This meant the tests wouldn’t compile or run without the Hive connector being statically linked.

## Now: 
The PR replaces those direct Hive references with calls into the ConnectorObjectFactory and connector-agnostic interfaces.

For example, in tests that used to say:
```c++
auto handle = std::make_shared<HiveTableHandle>(...);
```

they now construct folly::dynamic options and call:
```c++
auto handle = factory.makeTableHandle("hive", tableName, cols, options);
```

Similarly, partitioning specs that used to directly instantiate `HivePartitionFunctionSpec` are now requested through:
```c++
factory.makePartitionFunctionSpec("hive", opts);
```

This ensures the exec layer and its tests depend only on the abstract connector API, not on Hive headers.

# Why Exec Tests Need to Be Connector Agnostic

- Build isolation: If every exec test pulls in Hive, any change in Hive (even trivial) will force a recompile of all exec tests.
- Future connectors: Iceberg, Hudi, DeltaLake all need to reuse exec operators. Tests should validate operator logic, not Hive specifics.
- Portability: Connector-agnostic tests can run even if a connector plugin isn’t present, as long as a minimal factory (real or mock) is registered.

This is part of a larger effort to make the connector plugin architecture. Whether or not to support dynamic loading of connectors is still under discussion, but making exec tests connector agnostic is a must-do item to make Velox compile with VELOX_ENABLE_HIVE_CONNECTOR=OFF. Exec tests that link against Hive types would make the build fail if VELOX_ENABLE_HIVE_CONNECTOR=OFF. This is what we have agreed on in https://github.com/facebookincubator/velox/issues/13698.

# Steps to Make Exec Tests Connector Agnostic

1. Introduce ConnectorObjectFactory: Define abstract factory methods: makeTableHandle, makeColumnHandle, makeInsertTableHandle, makePartitionFunctionSpec, etc.  https://github.com/facebookincubator/velox/pull/13798

2. Implement connector-specific factories: Hive (and later Iceberg) provide a HiveObjectFactory implementing these methods. https://github.com/facebookincubator/velox/pull/13798

3. Refactor PlanBuilder, including TableScanBuilder, TableWriterBuilder and PartitionFunctionSpec related code in exec test utils.  Replace direct Hive handle creation with factory calls. Update affected tests too.  (this PR)

4. Finish PlanBuilder refactor to remove Tpch and Tpcds specific code. Can be done as a new beginner task.
5. Refactor HiveConnectorTestBase, make it connector agnostic. We can potentially rename it as ConnectorTestBase and move it back to connectors/common/tests. Hive tests, Iceberg tests or other connector-specific tests can also inherit it.

Depends on https://github.com/facebookincubator/velox/pull/13798

### DIFF
--- a/velox/connectors/Connector.h
+++ b/velox/connectors/Connector.h
@@ -27,6 +27,7 @@
 #include "velox/common/file/TokenProvider.h"
 #include "velox/common/future/VeloxPromise.h"
 #include "velox/core/ExpressionEvaluator.h"
+#include "velox/core/PartitionFunction.h"
 #include "velox/type/Filter.h"
 #include "velox/vector/ComplexVector.h"
 
@@ -93,6 +94,8 @@ struct ConnectorSplit : public ISerializable {
 
 class ColumnHandle : public ISerializable {
  public:
+  enum class ColumnType { kPartitionKey, kRegular, kSynthesized };
+
   virtual ~ColumnHandle() = default;
 
   virtual const std::string& name() const = 0;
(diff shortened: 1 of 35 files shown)


velox/connectors/Connector.h
```diff
@@ -688,6 +718,59 @@ class ConnectorFactory {
       folly::Executor* ioExecutor = nullptr,
       folly::Executor* cpuExecutor = nullptr) = 0;
 
+  virtual std::shared_ptr<ConnectorSplit> makeConnectorSplit(
```
mbasmanova, line 721: @yingsu00 Connector is a generic API, but these new APIs are specific to Hive-like connectors. For example, these do not make sense for MySQL connector.

yingsu00:
Closing for now. Will open new PRs that introduce Iceberg standalone connector by Masha's suggestion.

PR #14693, opened 2025-09-03
fix: Centralize the places that we do lazy materialization

xiaoxmeng:
Summary: We currently handle the left hand side lazy load in different places in the code but load all the rows without selectivity. This is error prone and cause subtle difference between left/right side processing. This change lazy load them once at add input.

Differential Revision: D81526796

### DIFF
--- a/velox/exec/MergeJoin.cpp
+++ b/velox/exec/MergeJoin.cpp
@@ -216,8 +216,8 @@ bool MergeJoin::needsInput() const {
 void MergeJoin::addInput(RowVectorPtr input) {
   VELOX_CHECK_NULL(input_);
   input_ = std::move(input);
+  loadColumns(input_, *operatorCtx_->execCtx());
   leftRowIndex_ = 0;
-
   if (joinTracker_) {
     joinTracker_->resetLastVector();
   }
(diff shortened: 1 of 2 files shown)


velox/exec/MergeJoin.h
```diff
@@ -331,6 +331,10 @@ class MergeJoin : public Operator {
   // rows from the left side that have a match on the right.
   RowVectorPtr filterOutputForAntiJoin(const RowVectorPtr& output);
 
+  void clearLeftInput();
```
JkSelf, line 334: nit: Please add some comments for `clearLeftInput ` and `clearRightInput ` method.

JkSelf (approved):
Thanks for your fix. LGTM. Except one small nit.

velox/exec/MergeJoin.cpp
```diff
@@ -216,8 +216,8 @@ bool MergeJoin::needsInput() const {
 void MergeJoin::addInput(RowVectorPtr input) {
   VELOX_CHECK_NULL(input_);
   input_ = std::move(input);
+  loadColumns(input_, *operatorCtx_->execCtx());
```
Yuhta, line 219: Let's leave a TODO here to lazy loading both sides.  The old code seems special treat the left side trying to keep it lazy, this change removes that, which we think is a low risk regression at this stage, but we should consider putting it back in an improved form in the future.

pedroerp: @xiaoxmeng @Yuhta @JkSelf  this was done on purpose. If we eagerly materialize columns, we will have to materialize all rows including the ones that don't have any matches for join keys. The previous code would only materialize rows that have matches. For low selectivity joins, this can make a big difference.

There is little point in doing this today for the right side because it comes from a different pipeline, so we never get lazies here (lazies are materialized once we cross pipelines). 

Unless I'm missing something, I think we will have to roll this back.

PR #14702, opened 2025-09-03
Fix(tpcds): return const& from toTableName and avoid redundant map construction (#14678)

### LINKED ISSUE #14678: TPC-DS: toTableName should return const &
### Description

toTableName returns std::string causing unnecessary copy. Also, on each call it creates a full map of table names redundantly.

Let's fix this to avoid creating a map multiple times and copying the strings.

Consider using VELOX_DEFINE_ENUM_NAME macro.

CC: @majetideepak @assignUser @pramodsatya @aditi-pandit @pedroerp 

```
std::string toTableName(Table table) {
  auto inverted = invertTpcdsTableMap();
  auto it = inverted.find(table);
  if (it != inverted.end()) {
    return inverted.at(table);
  }
  VELOX_UNREACHABLE("Invalid TPC-DS table: {}", static_cast<uint8_t>(table));
}


std::unordered_map<std::string, Table> tpcdsTableMap() {
  static const std::unordered_map<std::string, Table> map{
      {"call_center", Table::TBL_CALL_CENTER},
      {"catalog_page", Table::TBL_CATALOG_PAGE},
      {"catalog_returns", Table::TBL_CATALOG_RETURNS},
      ....
  };
  return map;
}

std::unordered_map<Table, std::string> invertTpcdsTableMap() {
  std::unordered_map<Table, std::string> inverted;
  for (const auto& [key, value] : tpcdsTableMap()) {
    inverted.emplace(value, key);
  }
  return inverted;
}
```

suryaprakash0010:
### Summary
- Changed `toTableName()` to return `const std::string&` instead of `std::string` to avoid unnecessary string copies.
- Removed redundant map reconstruction (`invertTpcdsTableMap`) by introducing a static map that is built once and reused.
- Added `VELOX_DEFINE_ENUM_NAME(Table)` for consistent enum-to-string mapping.
- Updated `fromTableName()` to use `TableName::tryToTable` for efficient reverse lookups.
- Added a test (`test_tpcds_optimization.cpp`) to validate behavior.

### Benefits
- Eliminates repeated map rebuilding, improving performance.
- Avoids extra string allocations and copies.
- Aligns with Velox coding style and enum handling.
- Ensures correctness with new test coverage.

Closes #14678

### DIFF
--- a/test_tpcds_optimization.cpp
+++ b/test_tpcds_optimization.cpp
@@ -1,0 +1,81 @@
+#include "velox/tpcds/gen/TpcdsGen.h"
+#include <iostream>
+#include <chrono>
+
+using namespace facebook::velox::tpcds;
+
+int main() {
+    // Test basic functionality
+    std::cout << "Testing TPC-DS table name optimization...\n";
+    
+    // Test all table names
+    const auto tables = {
+        Table::TBL_CALL_CENTER,
+        Table::TBL_CATALOG_PAGE,
+        Table::TBL_CATALOG_RETURNS,
+        Table::TBL_CATALOG_SALES,
+        Table::TBL_CUSTOMER,
+        Table::TBL_CUSTOMER_ADDRESS,
+        Table::TBL_CUSTOMER_DEMOGRAPHICS,
+        Table::TBL_DATE_DIM,
+        Table::TBL_HOUSEHOLD_DEMOGRAPHICS,
+        Table::TBL_INCOME_BAND,
+        Table::TBL_INVENTORY,
+        Table::TBL_ITEM,
+        Table::TBL_PROMOTION,
+        Table::TBL_REASON,
+        Table::TBL_SHIP_MODE,
+        Table::TBL_STORE,
+        Table::TBL_STORE_RETURNS,
+        Table::TBL_STORE_SALES,
+        Table::TBL_TIME_DIM,
+        Table::TBL_WAREHOUSE,
+        Table::TBL_WEB_PAGE,
+        Table::TBL_WEB_RETURNS,
+        Table::TBL_WEB_SALES,
+        Table::TBL_WEB_SITE
+    };
+    
+    // Test correctness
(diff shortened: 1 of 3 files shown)


MBkkt (commented):
But `toTableName` and `fromTableName` isn't necessary?
`TableName::toTable(...)` and `TableName::toName(...);` can be used instead

`VELOX_DECLARE_ENUM_NAME` declare functions for this enum, `TableName::toName` and `TableName::toTable`, so they can be used instead `toTableName` and `fromTableName`

velox/tpcds/gen/TpcdsGen.cpp
```diff
@@ -821,20 +817,27 @@ const RowTypePtr getTableSchema(Table table) {
...
   }
   VELOX_UNREACHABLE("Invalid TPC-DS table: {}", static_cast<uint8_t>(table));
 }
 
-Table fromTableName(const std::string& tableName) {
-  auto map = tpcdsTableMap();
-  auto it = map.find(tableName);
-  if (it != map.end()) {
-    return it->second;
+Table fromTableName(const std::string_view& tableName) {
```
czentgr, line 837: We don't need const& for std::string_view. It is implicitly const and is designed to not use const& as is used for std::string.

test_tpcds_optimization.cpp
```diff
@@ -0,0 +1,81 @@
+#include "velox/tpcds/gen/TpcdsGen.h"
```
czentgr, line 1: Needs license header.

test_tpcds_optimization.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <iostream>
+#include <chrono>
+
+using namespace facebook::velox::tpcds;
+
+int main() {
```
pramodsatya, line 7: Please add tests in `tpcds/gen/tests/TpcdsGenTest.cpp`.

PR #14704, opened 2025-09-03
refactor: Add Presto TypeParser enum logic to velox/type

HeidiHan0000:
Summary:
There was a refactor to move the entire parser folder to functions/prestosqltypes/parser in D78837244 since this parser is Presto specific.
Then in D78903388 and D81118053, some functions were added to support enum types in Presto.

Then Presto was updated to point to the new folder in D81545524. But since Presto release takes longer, in order to roll enum types out for the moment, I have added the new enum parsing logic to the old folder. I will remove the folder entirely once Presto release is complete.

Differential Revision: D81606695

### DIFF
--- a/velox/core/SimpleFunctionMetadata.h
+++ b/velox/core/SimpleFunctionMetadata.h
@@ -301,6 +301,19 @@ struct TypeAnalysis<facebook::velox::BigintEnumT<E>> {
   }
 };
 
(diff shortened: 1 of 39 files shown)


kewang1024:
There seems to be build failure @HeidiHan0000

kgpai (approved):

PR #14705, opened 2025-09-03
feat: Implement ST_LineFromText, ST_LineString, ST_MultiPoint

Sullivan-Patrick:
Summary: Adds 3 geometry constructor UDFs to velox

Differential Revision: D80682501


velox/functions/prestosql/geospatial/GeometrySerde.h
```diff
@@ -349,6 +351,56 @@ class GeometryDeserializer {
   static const std::unique_ptr<geos::geom::Envelope> deserializeEnvelope(
       const StringView& geometry);
 
+  template <typename T>
+  static const std::unique_ptr<geos::geom::CoordinateArraySequence>
```
kgpai: I am curious is it correct to have a const std::unique_ptr ?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -1891,4 +1892,85 @@ struct GeometryUnionFunction {
...
+      const arg_type<Array<Geometry>>& input) {
+    std::unique_ptr<geos::geom::CoordinateArraySequence> coords =
+        geospatial::GeometryDeserializer::deserializePointsToCoordinate<
+            Geometry>(input, "ST_LineString", true);
+
+    std::unique_ptr<geos::geom::LineString> lineString;
+    if (input.size() < 2) {
+      lineString = factory_->createLineString();
+    } else {
+      lineString = factory_->createLineString(std::move(coords));
```
kgpai, line 1939: Coords is only used here so maybe we can move it here so as to not create the object when size < 2.

velox/functions/prestosql/geospatial/GeometrySerde.h
Sullivan-Patrick: Can you elaborate? I'm not aware of any unwanted behaviors the const qualifier may be causing. It doesn't add anything, but since we don't need to change the contents of the ptr at any point in the current code I thought it was a good idea to add.

velox/functions/prestosql/GeometryFunctions.h
Sullivan-Patrick: I did this originally actually, but changed it for consistency with Java. There, the checks for null array inputs/non-point geometries/empty geometries/repeated point sequences are all made before this check. Put another way, if we were to move the check up, an input of 1 empty geom would return an error in java and an empty linestring in c++.

velox/functions/prestosql/geospatial/GeometrySerde.h
kgpai: When you have a const std::unique_ptr , you cant do a std::move on it , so in this case after this function is called the subsequent function that has ownership of the unique_ptr<geos::geom::CoordinateArraySequence> wont be able to move it - which I have to ask why ? I suppose rather you wanted a std::unique_ptr<const geos::geom::CoordinateArraySequence> .

PR #14706, opened 2025-09-03
fix(build): Replace actor with owner for determining merge base

czentgr:
Previously the actor (the workflow initiator) was used to try and determine the merge base commit sha. However, a user pushing to a different repo becomes the owner and $OWNER:$HEAD_REF does not exist in that scenario. Instead, try to use the PR creator name. We’ve seen HTTP 404 from GH CLI when the actor does not match the repo name.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -124,9 +124,9 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
           INPUT_REF: ${{ inputs.ref || 'main' }}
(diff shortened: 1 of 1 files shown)


czentgr:
Will remove the last commit.
```
https://github.com/facebookincubator/velox/actions/runs/17502873870/job/49719721674?pr=14706
```
shows correct execution with the correct base commit of the branch used.

assignUser (commented):
Nice, once you pop the last commit we can merge this!

czentgr:
@assignUser Done.

PR #14715, opened 2025-09-03
fix: Add `ARRAY<ARRAY<VARCHAR>>` support to `array_frequency`

yanhanBazoooka:
Summary:
This change resolves compilation errors for array_frequency when called
with `ARRAY<ARRAY<VARCHAR>>` arguments, which was causing ~364 queries
per week to fail with "Scalar function not registered" errors.

**Problem:**
The original `ArrayFrequencyFunction` used `ArrayView<true, Varchar>` directly
as keys in `folly::F14FastMap`, but this type lacks the necessary hash and
equality operators for nested arrays.

**Solution:**
- Created a template specialization of `ArrayFrequencyFunction` for `ARRAY<ARRAY<VARCHAR>>`
- Implemented `ArrayVarcharKey` helper `struct` that converts `ArrayView` to a
  hashable representation using `std::vector<std::optional<std::string>>`
- Added custom `ArrayVarcharKeyHasher` with proper hash combining for elements
- Manually constructs output arrays using `keyWriter.add_item()` instead of
  `copy_from()` to handle type compatibility issues

**Changes:**
- velox/functions/prestosql/ArrayFunctions.h: Added specialized template
- velox/functions/prestosql/registration/ArrayFunctionsRegistration.cpp:
  Added registration for Array<Varchar> type
- velox/functions/prestosql/tests/ArrayFrequencyTest.cpp: Added test coverage
  for nested varchar arrays

Differential Revision: D81629709


amitkdutta:
@yanhanBazoooka Please remove the `draft` and submit the PR. Also take a look at fuzzer failure if its related.

han-yan01:
> @yanhanBazoooka Please remove the `fix` and submit the PR. Also take a look at fuzzer failure if its related.

Thanks! Correct me if I'm wrong: I don't think the fuzzer failure is related due to it failing caused by presto java native hashing method runtime error for null mapkeys. Velox's cpp hashing native methods takes null as map keys. Added function intended behavior stays the same.

velox/functions/prestosql/ArrayFunctions.h
```diff
@@ -610,6 +610,78 @@ struct ArrayFrequencyFunction {
...
+  // Helper struct to hash Array<Varchar>
+  struct ArrayVarcharKey {
+    std::vector<std::optional<std::string>> elements;
+
+    ArrayVarcharKey() = default;
+
+    explicit ArrayVarcharKey(const arg_type<Array<Varchar>>& array) {
+      for (const auto& item : array) {
+        if (item.has_value()) {
+          elements.emplace_back(item.value().getString());
```
kgpai: This will create a copy of the string . Can we try and avoid that ?

velox/functions/prestosql/ArrayFunctions.h
```diff
@@ -610,6 +610,78 @@ struct ArrayFrequencyFunction {
...
+    }
+  };
+
+  struct ArrayVarcharKeyHasher {
+    std::size_t operator()(const ArrayVarcharKey& key) const {
+      std::size_t hash = 0;
+      for (const auto& elem : key.elements) {
+        if (elem.has_value()) {
+          hash = folly::hash::hash_combine(
+              hash, std::hash<std::string>{}(elem.value()));
```
kgpai: Why cant we store the hash when we read an item from the array and store that along with a reference to item and a frequency count. Keep updating the frequency count on a collision or adding a new entry when no collision, this will be less expensive than maintaining a copy via ArrayVarcharKey.

Issue #14712, opened 2025-09-03
Failing S3 and GCS tests


bikramSingh91:
### Bug description

The following s3 and GCS tests have started failing on master since yesterday.
```
The following tests FAILED:
	424 - velox_s3insert_test (Timeout)
	425 - velox_s3read_test (Timeout)
	427 - velox_s3multiendpoints_test (Timeout)
	431 - velox_gcs_insert_test (Timeout)
	432 - velox_gcs_multiendpoints_test (Timeout)
```

Example: https://github.com/facebookincubator/velox/actions/runs/17425696652/job/49472269767

Disabling them in #14711 till they are fixed.

cc: @JkSelf @majetideepak 

### System information

Github CI

### Relevant logs

```bash

```

JkSelf:
@bikramSingh91 I was able to reproduce the issue locally and will work on a fix shortly. Thanks for reporting it.

JkSelf:
@bikramSingh91 
This issue can be fixed in https://github.com/facebookincubator/velox/pull/14718. Just curious — why wasn’t this failure caught by the CI tests in PR #14682?

bikramSingh91:
@JkSelf Thank you for addressing the issue and pointing out the offending diff. It appears that the CI did detect the problem, but it may have been overlooked. I will kindly reach out to the author to ensure more careful attention in the future.

PR #14718, opened 2025-09-04
fix: Enable S3 and GCS unit tests

JkSelf:
Fix https://github.com/facebookincubator/velox/issues/14712.

In PR #14682, the `config_ `object is moved into the `Connector `using `std::move(config_)`, which makes `config_ `unavailable for constructing the `FileHandleGenerator`—resulting in a null pointer. This fix resolves the NPE issue by using `hiveConfig_->config()` instead of the original `config_`.


JkSelf:
@mbasmanova @hdikeman @bikramSingh91 @majetideepak 
Could you help to review this PR? Thanks.

bikramSingh91 (approved):

Issue #13168, opened 2025-04-28
Memory leak for TableScan for Spark using Gluten


nimesh1601:
### Bug description

There is a memory leak in the TableScan Operator. Seems like as per #12701 relevant cancel API is being called in TableScan , but same has not been implemented in HiveDataSource

### System information

NA

### Relevant logs

```bash
E20250427 14:32:18.128706   176 VeloxMemoryManager.cc:401] Failed to release Velox memory manager after 43350ms as there are still outstanding memory resources. 
E20250427 14:32:18.128832   176 MemoryPool.cpp:435] [MEM] Memory leak (Used memory): Memory Pool[op.0.0.0.TableScan LEAF root[root] parent[node.0] MALLOC track-usage thread-safe]<unlimited max capacity unlimited capacity used 276.00KB available 748.00KB reservation [used 276.00KB, reserved 1.00MB, min 0B] counters [allocs 114189, frees 114184, reserves 0, releases 1, collisions 0])>
E20250427 14:32:18.128959   176 Exceptions.h:66] Line: cpp/velox/memory/VeloxMemoryManager.cc:102, Function:removePool, Expression: pool->reservedBytes() == 0 (1048576 vs. 0), Source: RUNTIME, ErrorCode: INVALID_STATE
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (1048576 vs. 0)
Retriable: False
Expression: pool->reservedBytes() == 0
Function: removePool
cpp/velox/memory/VeloxMemoryManager.cc
Line: 102
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN6gluten20ListenableArbitrator10removePoolEPN8facebook5velox6memory10MemoryPoolE
# 4  _ZN8facebook5velox6memory13MemoryManager8dropPoolEPNS1_10MemoryPoolE
[...]

PR #8205, opened 2023-12-30
Fix memory leak bug when directbufferedinput is enabled

FelixYBW:
Fix issue #8204 

The root cause is the speculative fetch of DirectBufferedInput. Like if we have a SQL:

`select A where B=0`

It may prefetch ColumnChunk A where all B acually isn't 0. Then when the task finished, the fetch is still pending and hold the memory block.

The solution is to wait all pending requests finish before descror return.

Issue #8204, opened 2023-12-30
MemoryLeak when DirectBufferedInput is enabled


FelixYBW:
### Bug description

It's very similar issue as #8198 but even we only enable DIrectBufferedInput. The error info is the same

```
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Bad memory usage track state: Memory Pool[op.0.0.0.TableScan LEAF root[WholeStageIterator_root] parent[node.0] MALLOC track-usage thread-safe]<unlimited max capacity unlimited capacity used 8.42MB available 592.00KB reservation [used 8.42MB, reserved 9.00MB, min 0B] counters [allocs 27507, frees 27497, reserves 0, releases 1, collisions 0])>
Retriable: False
Expression: (usedReservationBytes_ == 0) && (reservationBytes_ == 0) && (minReservationBytes_ == 0)
Function: ~MemoryPoolImpl
File: /home/binweiyang/gluten/ep/build-velox/build/velox_ep/velox/common/memory/MemoryPool.cpp
Line: 398
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox6memory14MemoryPoolImplD2Ev
# 4  _ZN6gluten18VeloxMemoryManagerD1Ev
# 5  _ZN6gluten18VeloxMemoryManagerD0Ev
# 6  0x00007f9c677bf427
```

The root cause is the speculative fetch of DirectBufferedInput. Like if we have a SQL:

`select A where B=0`
It may prefetch ColumnChunk A where all B acually isn't 0. Then when the task finished, the fetch is still pending and hold the memory block. 

The solution is to wait all pending requests finish before descror return.

In theory we should have a clean up method called from application, manually make sure all resources are released before objects destruct.

### System information

```
Velox System Info v0.0.2
Commit: 233928d1af9d6b9d216cc32380e7043d9a530b44
CMake Version: 3.16.3
[...]

PR #14722, opened 2025-09-04
Fix memory leak caused by asynchronous prefetch

j7nhai:


j7nhai:
please check this PR. @xiaoxmeng @tanjialiang  @fzhedu @zhztheplayer @majetideepak

j7nhai:
Hi, @nimesh1601 
In issue https://github.com/facebookincubator/velox/issues/13168, It looks that there is some memory leak for TableScan for Spark using Gluten. And this PR may fix it. You can try to run the task with this patch.

FelixYBW:
@pedroerp could you prioritize this PR? We noted the leak in our TPCDS test already

pedroerp:
> @pedroerp could you prioritize this PR? We noted the leak in our TPCDS test already

I pinged @xiaoxmeng and asked him to take a look. Thank you guys for looking into this.

j7nhai:
> @j7nhai thanks for the fix. So the pool is not destroyed on memory manager destruction? The async load is executed by folly executor and if we hold a reference to the pool in async load, why this caused the problem. The fix might introduce other memory pool lifecycle issue as discovered in Prestissimo use case. Can you explain a bit details that lead to the problem. I don't think hold a reference to the pool itself can cause issue.

It’s not the extra pool reference that causes the memory leak; there is indeed a memory leak issue here. When destructing, we need to wait for the asynchronous threads to exit to ensure that the memory allocated by the pool is properly released before the memory manager is destroyed (otherwise, there will be exceptions in Gluten’s Velox memory manager). **I mistakenly thought that your pool reference was intended to fix the memory leak issue**, but actually, it doesn’t solve it. What’s needed is to wait for the folly executor to finish execution in the destructor of DirectCoalescedLoad. 

So, maybe I can keep your feature—both fixes can coexist and are not conflicting.

j7nhai:
I have kept the previous features and only fixed the memory leak issue. Please take another look, thanks!  @xiaoxmeng @tanjialiang @fzhedu @zhztheplayer @majetideepak @FelixYBW

velox/dwio/common/DirectBufferedInput.h
```diff
@@ -101,6 +101,17 @@ class DirectCoalescedLoad : public cache::CoalescedLoad {
...
 
+  void cancel() override {
+    folly::SemiFuture<bool> waitFuture(false);
+    if (!loadOrFuture(&waitFuture)) {
+      waitFuture.wait();
+    }
+    for (auto& request : requests_) {
+      pool_->freeNonContiguous(request.data);
+    }
+    CoalescedLoad::cancel();
```
philo-he, line 112: Could you help confirm if it is better to call cancel at first?

j7nhai: It is incorrect to call the cancel function first, because doing so will immediately set state_ to cancelled, and as a result, the loadOrFuture function will not perform any actual waiting.

velox/dwio/common/DirectBufferedInput.h
```diff
@@ -101,6 +101,17 @@ class DirectCoalescedLoad : public cache::CoalescedLoad {
     return size;
   }
 
+  void cancel() override {
```
rui-mo, line 104: The `cancel()` function is invoked by `~DirectBufferedInput()`, which ultimately depends on the destructor of TableScan, as TableScan owns the `std::unique_ptr<connector::DataSource> dataSource_`. This dependency relationship is illustrated in [this discussion](https://github.com/facebookincubator/velox/pull/8205#issuecomment-1916026685). As a result, TableScan must be destroyed before the memory manager, but the current approach in this PR does not explicitly guarantee that order.

To ensure that the memory held by the load is always properly released, [PR #8205](https://github.com/facebookincubator/velox/pull/8205) triggers close() through the TableScan::close() call chain to address the similar [issue](https://github.com/facebookincubator/velox/issues/8204).

Would you like to share your opinions on the fixes? cc: @xiaoxmeng @FelixYBW Thanks.

FelixYBW:
@j7nhai We will pick this PR to oap/velox and use it in Gluten. FYI.

https://github.com/IBM/velox/pull/1388

@pedroerp @xiaoxmeng can you review the PR?

velox/dwio/common/DirectBufferedInput.h
FelixYBW: The solution is either free the resources manually from top to bottom or free it in object destruction. Either way works.

velox/dwio/common/DirectBufferedInput.h
```diff
@@ -101,6 +101,17 @@ class DirectCoalescedLoad : public cache::CoalescedLoad {
...
+    folly::SemiFuture<bool> waitFuture(false);
+    if (state() == State::kLoading && !loadOrFuture(&waitFuture)) {
+      waitFuture.wait();
```
xiaoxmeng, line 107: such wait can easily lead to deadlock in Prestissimo use case @kewang1024 @Yuhta

FelixYBW: Can you share more details? Let me check if Gluten has the issue or not.

xiaoxmeng: @FelixYBW  I explained this offline to @rui-mo

PR #14724, opened 2025-09-04
refactor: Modernize velox/type/Subfield.h

mbasmanova:
Summary:
- Turn SubfieldKind enum into enum class.
- Add Subfield::is(SubfieldKind) and Subfield::as<T> helper methods.
- Add Subfield::baseName() getter.
- Move method implementations to .cpp.

Differential Revision: D81680820

### DIFF
--- a/velox/connectors/hive/HiveConnectorUtil.cpp
+++ b/velox/connectors/hive/HiveConnectorUtil.cpp
@@ -110,8 +110,7 @@ void addSubfields(
       folly::F14FastMap<std::string, std::vector<SubfieldSpec>> required;
       for (auto& subfield : subfields) {
         auto* element = subfield.subfield->path()[level].get();
-        auto* nestedField =
-            dynamic_cast<const common::Subfield::NestedField*>(element);
+        auto* nestedField = element->as<common::Subfield::NestedField>();
         VELOX_CHECK(
             nestedField,
             "Unsupported for row subfields pruning: {}",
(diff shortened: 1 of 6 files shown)


velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -150,20 +149,18 @@ void addSubfields(
       std::vector<int64_t> longSubscripts;
       for (auto& subfield : subfields) {
         auto* element = subfield.subfield->path()[level].get();
-        if (dynamic_cast<const common::Subfield::AllSubscripts*>(element)) {
+        if (element->is(common::SubfieldKind::kAllSubscripts)) {
```
pedroerp, line 152: that is much nicer :)

pedroerp: FWIW, in llvm's codebase (IIRC) they provide these things as regular functions, and make sure they work across all important types:

```
auto* nestedField = cast_as<common::Subfield::NestedField>(element);
...
if (is<common::SubfieldKind::kAllSubscripts>(element)) {
...
```

velox/core/FilterToExpression.cpp
```diff
@@ -57,64 +57,35 @@ core::TypedExprPtr handleNullAllowed(
...
       }
 
-      case common::SubfieldKind::kLongSubscript:
-      case common::SubfieldKind::kStringSubscript:
-      case common::SubfieldKind::kAllSubscripts:
-        VELOX_NYI(
-            "All other SubfieldKind except kNestedField are not implemented in getTypedExprFromSubfield");
-        break;
+      default:
+        VELOX_NYI("SubfieldKind other than kNestedField are not supported");
```
pedroerp: maybe include the one you got in the error message?

velox/type/Subfield.cpp
```diff
@@ -51,4 +74,56 @@ Subfield Subfield::clone() const {
   return subfield;
 }
 
+bool Subfield::isPrefix(const Subfield& other) const {
+  if (path_.size() < other.path_.size()) {
+    for (size_t i = 0; i < path_.size(); ++i) {
+      if (!(*path_[i].get() == *other.path_[i].get())) {
```
pedroerp, line 80: can we not do `if (.. != ...) {`  ? If not, maybe we should provide operator!= as well

mbasmanova: @pedroerp We may need to follow-up to improve this.

Issue #1622, opened 2022-05-14
[Refactoring] Make Velox support Parquet thrift code generation


yingsu00:
This is a sub issue of https://github.com/facebookincubator/velox/issues/1533

We will make the Parquet thrift cpp code automatically generated. The thrift definition is from https://github.com/apache/parquet-cpp/blob/master/src/parquet/parquet.thrift . This is similar to Protobuf for DWRF: velox/dwio/dwrf/proto

Note that we may need to install fb-thrift in Velox. This requires us to update the setup scripts.

PR #14725, opened 2025-09-04
feat(parquet): Add NaN statistics to Parquet writer

PingLiuPing:
Add NaN statistic to Parquet writer

This change introduces a NaN statistic in the Parquet writer. Unlike other statistics (e.g., null_count, distinct_count), the NaN statistic is not written to the Parquet file footer. It is only reported to the Parquet writer caller when needed, such as when writing Iceberg Parquet data files.

### DIFF
--- a/velox/dwio/parquet/writer/arrow/Metadata.cpp
+++ b/velox/dwio/parquet/writer/arrow/Metadata.cpp
@@ -101,10 +101,12 @@ static std::shared_ptr<Statistics> MakeTypedColumnStats(
         metadata.num_values - metadata.statistics.null_count,
         metadata.statistics.null_count,
         metadata.statistics.distinct_count,
+        /*nan_count=*/0,
         metadata.statistics.__isset.max_value ||
             metadata.statistics.__isset.min_value,
         metadata.statistics.__isset.null_count,
-        metadata.statistics.__isset.distinct_count);
+        metadata.statistics.__isset.distinct_count,
+        /*has_nan_count=*/false);
   }
   // Default behavior
   return MakeStatistics<DType>(
@@ -114,9 +116,11 @@ static std::shared_ptr<Statistics> MakeTypedColumnStats(
       metadata.num_values - metadata.statistics.null_count,
       metadata.statistics.null_count,
       metadata.statistics.distinct_count,
+      /*nan_count=*/0,
       metadata.statistics.__isset.max || metadata.statistics.__isset.min,
       metadata.statistics.__isset.null_count,
-      metadata.statistics.__isset.distinct_count);
+      metadata.statistics.__isset.distinct_count,
+      /*has_nan_count=*/false);
 }
 
 std::shared_ptr<Statistics> MakeColumnStats(
@@ -1015,6 +1019,22 @@ class FileMetaData::FileMetaDataImpl {
     file_decryptor_ = file_decryptor;
   }
 
+  // Set NaN counts from the builder (called during Finish)
+  // This stores total NaN counts per field ID across all row groups.
+  void setNaNCounts(
+      std::unordered_map<int32_t, std::pair<int64_t, bool>> nan_counts) {
+    field_nan_counts_ = std::move(nan_counts);
+  }
+
+  // Get total NaN count for a specific field ID across all row groups.
+  std::pair<int64_t, bool> getNaNCount(int32_t fieldId) const {
+    auto it = field_nan_counts_.find(fieldId);
+    if (it != field_nan_counts_.end()) {
+      return it->second;
+    }
+    return {0, false};
+  }
+
  private:
   friend FileMetaDataBuilder;
   uint32_t metadata_len_ = 0;
@@ -1024,6 +1044,9 @@ class FileMetaData::FileMetaDataImpl {
   std::shared_ptr<const KeyValueMetadata> key_value_metadata_;
   const ReaderProperties properties_;
   std::shared_ptr<InternalFileDecryptor> file_decryptor_;
+  // Total NaN counts per field ID across all row groups: field_id ->
+  // (nan_count, has_nan_count).
+  std::unordered_map<int32_t, std::pair<int64_t, bool>> field_nan_counts_;
 
   void InitSchema() {
     if (metadata_->schema.empty()) {
@@ -1198,6 +1221,10 @@ void FileMetaData::AppendRowGroups(const FileMetaData& other) {
 std::shared_ptr<FileMetaData> FileMetaData::Subset(
     const std::vector<int>& row_groups) const {
   return impl_->Subset(row_groups);
+}
+
+std::pair<int64_t, bool> FileMetaData::getNaNCount(int32_t fieldId) const {
+  return impl_->getNaNCount(fieldId);
 }
 
 void FileMetaData::WriteTo(
@@ -1715,6 +1742,19 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
   // column metadata
   void SetStatistics(const EncodedStatistics& val) {
     column_chunk_->meta_data.__set_statistics(ToThrift(val));
+    // Store NaN count separately since it's not written to the parquet file.
+    if (val.has_nan_count) {
+      nan_count_ = val.nan_count;
+      has_nan_count_ = true;
+    }
+  }
+
+  int64_t nan_count() const {
+    return nan_count_;
+  }
+
+  bool has_nan_count() const {
+    return has_nan_count_;
   }
 
   void Finish(
@@ -1883,6 +1923,9 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
       owned_column_chunk_;
   const std::shared_ptr<WriterProperties> properties_;
   const ColumnDescriptor* column_;
+  // NaN count is stored separately since it's not written to the parquet file.
+  int64_t nan_count_ = 0;
+  bool has_nan_count_ = false;
 };
 
 std::unique_ptr<ColumnChunkMetaDataBuilder> ColumnChunkMetaDataBuilder::Make(
@@ -1968,6 +2011,14 @@ void ColumnChunkMetaDataBuilder::SetStatistics(
 
 int64_t ColumnChunkMetaDataBuilder::total_compressed_size() const {
   return impl_->total_compressed_size();
+}
+
+int64_t ColumnChunkMetaDataBuilder::nan_count() const {
+  return impl_->nan_count();
+}
+
+bool ColumnChunkMetaDataBuilder::has_nan_count() const {
+  return impl_->has_nan_count();
 }
 
 class RowGroupMetaDataBuilder::RowGroupMetaDataBuilderImpl {
@@ -2062,6 +2113,16 @@ class RowGroupMetaDataBuilder::RowGroupMetaDataBuilderImpl {
     return row_group_->num_rows;
   }
 
+  // Returns a map of field_id -> (nan_count, has_nan_count).
+  std::unordered_map<int32_t, std::pair<int64_t, bool>> nan_counts() const {
+    std::unordered_map<int32_t, std::pair<int64_t, bool>> result;
+    for (const auto& builder : column_builders_) {
+      int32_t field_id = builder->descr()->schema_node()->field_id();
+      result[field_id] = {builder->nan_count(), builder->has_nan_count()};
+    }
+    return result;
+  }
+
  private:
   void InitializeColumns(int ncols) {
     row_group_->columns.resize(ncols);
@@ -2119,6 +2180,11 @@ void RowGroupMetaDataBuilder::Finish(
   impl_->Finish(total_bytes_written, row_group_ordinal);
 }
 
+std::unordered_map<int32_t, std::pair<int64_t, bool>>
+RowGroupMetaDataBuilder::nan_counts() const {
+  return impl_->nan_counts();
+}
+
 // file metadata
 class FileMetaDataBuilder::FileMetaDataBuilderImpl {
  public:
@@ -2138,6 +2204,9 @@ class FileMetaDataBuilder::FileMetaDataBuilderImpl {
   }
 
   RowGroupMetaDataBuilder* AppendRowGroup() {
+    // Accumulate NaN counts from the previous row group before creating a new
+    // one.
+    accumulateNaNCountsFromCurrentRowGroup();
     row_groups_.emplace_back();
     current_row_group_builder_ = RowGroupMetaDataBuilder::Make(
         properties_, schema_, &row_groups_.back());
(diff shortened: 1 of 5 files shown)


PingLiuPing:
@majetideepak  Can you help review this PR? Thank you very much.

velox/dwio/parquet/thrift/ParquetThriftTypes.h
```diff
@@ -466,12 +466,14 @@ typedef struct _Statistics__isset {
...
         null_count(false),
         distinct_count(false),
+        nan_count(false),
         max_value(false),
         min_value(false) {}
   bool max : 1;
   bool min : 1;
   bool null_count : 1;
   bool distinct_count : 1;
+  bool nan_count : 1;
```
yingsu00: Just FYI: The Arrow Parquet community has been discussing NAN and there is a recent PR: https://github.com/apache/parquet-format/pull/514/files#top. The nan_count will be added as the 9th member of the Statistics class. Ideally this .cpp and .h file shall be auto generated, because editting this by hand is brittle and easy to get wrong. But I saw there were already some changes made to these generated cpp/h files, so I regard this as a temporary solution. Would you please add a comment here and reference https://github.com/apache/parquet-format/pull/514/files#top?

velox/dwio/parquet/thrift/ParquetThriftTypes.cpp
```diff
@@ -677,6 +677,11 @@ void Statistics::__set_distinct_count(const int64_t val) {
   __isset.distinct_count = true;
 }
 
+void Statistics::__set_nan_count(const int64_t val) {
```
yingsu00: This comment is for Statistics::read() on line 699 and Statistics::write() on line 776. If you add a new field to Statistics class, you'll have to update these two function implementation as well. Otherwise the nan count won't be read or written correctly.

velox/dwio/parquet/thrift/ParquetThriftTypes.h
PingLiuPing: @yingsu00  Thank you for the comments. Do you know the routine in velox that when `parquet.thrift` will be updated?
I will add a comment here by referencing https://github.com/apache/parquet-format/pull/514/files#top.

velox/dwio/parquet/thrift/ParquetThriftTypes.cpp
PingLiuPing: @yingsu00 Thanks, I get your point. As I mentioned in the PR description, this NaN statistic will not be written to the Parquet file footer; it’s only used to count NaNs during writing. Do you think it’s mandatory to include this field in the Parquet file?

velox/dwio/parquet/thrift/ParquetThriftTypes.h
PingLiuPing: By checking the latest `parquet.thrift`, we have missed field 7 & 8.

PingLiuPing: I added a comment at line 525 where this data member is been introduced.

yingsu00: @PingLiuPing It was an old version I downloaded years ago. We may want to update it and regenerate the cpp/h at some time in the near future.

velox/dwio/parquet/thrift/ParquetThriftTypes.cpp
yingsu00: > it’s only used to count NaNs during writing. 

It's good to have the conforming cpp/h file with the .thrift file. If we update .thrift but do not update the .cpp/.h, the code will be hard to maintain soon. The best way is to update .thrift and regenerate the cpp and h. If not doing it now, then at least keep the files agree with each other.

PingLiuPing: @yingsu00  Thanks. 
Since currently velox do not support auto generate these two files. Found this issue https://github.com/facebookincubator/velox/issues/1622 though. Do you agree to manually change `parquet.thrift` file first to match the changes in `ParquetThriftTypes.h `?
Regarding upgrade `parquet.thrift` file, I think we need to consider:
1. There are two commits each for `ParquetThriftTypes.h ` and `ParquetThriftTypes.cpp`. When generate new version files, make sure to pick the changes back.
2. Need to check arrow parquet reader and writer, there are lots of changes on the copied arrow writer and reader code. This might difficult to upgrade.

majetideepak:
@PingLiuPing As Ying mentioned, we should not edit the ParquetThrift files. We should wait for https://github.com/apache/parquet-format/pull/514/files#top to land.

>`this NaN statistic will not be written to the Parquet file footer; it’s only used to count NaNs during writing`

 Can you clarify this statement? What is the purpose of this during writing?

> Iceberg Parquet data files.

Are these different from regular Parquet files?

PingLiuPing:
Thanks @majetideepak for the comment.

Regarding https://github.com/apache/parquet-format/pull/514/files#top, the catch here is suppose it is been merged. But from velox point of view how to pickup the changes from parquet-format? There are changes been made to ParquetThriftTypes.h  and ParquetThriftTypes.cpp already. And should we also update arrow parquet writers?

> this NaN statistic will not be written to the Parquet file footer; it’s only used to count NaNs during writing

Since in the velox version of parquet, the NaN is not a supported stats field. I just count this field on the fly and report it to Iceberg stats collector. It will not be wrote to the actual parquet metadata. So the parquet data file is identical with other data files.

PingLiuPing:
@majetideepak  Regarding upgrade parquet version in velox, I see https://github.com/apache/parquet-format/pull/514/files#top, still not merged. And I think arrow also takes some time to support this.
Could you please advice what's the nest step? I can draft a PR to upgrade parquet version if needed.

majetideepak:
> report it to Iceberg stats collector. 

How does this work?

PingLiuPing:
> > report it to Iceberg stats collector.
> 
> How does this work?

Thanks,
The nan stats will be retrieved through iceberg stats collector.
You can refer to this code https://github.com/facebookincubator/velox/pull/14272/files#diff-a1e71e9d7b5dc2bccc4fbd79fdb26bd2b572398a59ce6dd1662b2d2a18eedc8dR90

PingLiuPing:
@majetideepak Regarding 

> @PingLiuPing As Ying mentioned, we should not edit the ParquetThrift files. We should wait for https://github.com/apache/parquet-format/pull/514/files#top to land.

https://github.com/apache/parquet-format/pull/514/ still not been merged.
Just to check what's our strategy on this? Should we wait still?
And suppose it get merged what's the next step for us? Should we also wait for arrow to adapt that changes?

And then what's the steps in velox to upgrade arrow?
Could you please share some thoughts on this and I can do some preparation work if required.

PingLiuPing:
@mbasmanova  Could you share some thoughts on Velox's strategy of upgrading parquet writer code? Thank you very much.

mbasmanova:
> @mbasmanova Could you share some thoughts on Velox's strategy of upgrading parquet writer code? Thank you very much.

@PingLiuPing What's the context for this question? Is there a summary I can read somewhere?

PingLiuPing:
@mbasmanova Thanks for looking at this.

Iceberg requires nan statistics to be collected during writing parquet data files. But current Velox parquet writer does not write such stats. The purpose of this PR is adding such support so that when writing parquet data files, nan stats can be collected. 

During review, we found that the nan stats is been discussed and will be introduced from  https://github.com/apache/parquet-format/pull/514.
But the question to us are:
1. Parquet community still not merged PR 514. 
2. Suppose parquet merged that PR, since Velox currently copy the parquet writer code from arrow, we still need to wait for arrow to support this.
3. Regarding the parquet writer code, Velox has made few changes to those code. It seems not a straightforward cherry-pick from arrow.

mbasmanova:
@PingLiuPing Thank you for the context. Velox has its own copy of Parquet writer code, right? If so, there is no "upgrading" to discuss. Am I missing something? Would you ask the question again?

CC: @majetideepak Deepak, you used to be the owner of anything Parquet related? Are you still the owner? If so, what is your take?

PingLiuPing:
@mbasmanova  Thanks.

> Velox has its own copy of Parquet writer code, right? If so, there is no "upgrading" to discuss. Am I missing something? Would you ask the question again?

Sorry for the earlier confusion. Yes, Velox maintains its own copy of the Parquet writer code, but that code was originally imported from Arrow. When I mentioned “upgrading,” I was referring to what Ying and Deepak suggested: the Parquet file format specification has evolved, and we may need to keep Velox’s Parquet writer aligned with the latest format definition.

This PR currently collects nan stats on the fly without writing them into the Parquet file metadata, so no upgrade is required. But since there hasn’t been feedback from other reviewers for quite a while, I want to clarify the direction:
1. Should we update Velox’s Parquet writer to match the latest Parquet format? If so, what is the general process for doing that?
2. If we decide not to upgrade, is this PR acceptable as-is?

majetideepak:
@PingLiuPing As you mentioned, Velox forked the Arrow Writer code and adapted it to Velox APIs and coding conventions. We can copy new writer improvements if the Arrow Parquet community made any.
We can also update the `parquet.thrift` format as well if there is a need.

> Should we update Velox’s Parquet writer to match the latest Parquet format? If so, what is the general process for doing that?

Can you clarify this further? What are the improvements made compared to the current state?

Regarding this change, you mentioned here https://github.com/facebookincubator/velox/pull/14725#issuecomment-3303510763 that another PR will be consuming this. Can we land that PR first?
We should not be modifying the generated ParquetThrift files. Can we keep the NaN stats in the writer separately for now?

PingLiuPing:
Thanks @majetideepak 

> Can you clarify this further? What are the improvements made compared to the current state?


Intems of statistics, the latest parquet.thrift has these two fields.

```
   /** If true, max_value is the actual maximum value for a column */
   7: optional bool is_max_value_exact;
   /** If true, min_value is the actual minimum value for a column */
   8: optional bool is_min_value_exact;
```

> Regarding this change, you mentioned here https://github.com/facebookincubator/velox/pull/14725#issuecomment-3303510763 that another PR will be consuming this. Can we land that PR first?

Thanks, but it would be great if we support this in the lowest layer first.


> We should not be modifying the generated ParquetThrift files. Can we keep the NaN stats in the writer separately for now?

Thanks, I can investigate this direction. But seems this way still seems not the best way as ideally the change should be made in ParquetThriftTypes.h and .cpp

PingLiuPing:
> Can we keep the NaN stats in the writer separately for now?

@majetideepak  Would you please elaborate more on this? Thank you.

PingLiuPing:
@majetideepak  I added `nan_count` to parquet.thrift and then generate .h and .cpp file based on that parquet.thrift file. Then I compared these two files with Velox `ParquetThriftTypes.h` and `ParquetThriftTypes.cpp`. I think what I have changed is similar to the file generate by thrift compiler, and to make it identical I copied `nan_count` related changes to `ParquetThriftTypes.h` and `ParquetThriftTypes.cpp`, other changes are irrelevant.
 Could you help have a look? Thanks

majetideepak:
@PingLiuPing We should not be modifying the `parquet.thrift` file directly. We risk breaking parquet compatibility.

majetideepak:
>Can we keep the NaN stats in the writer separately for now?
> @majetideepak Would you please elaborate more on this? Thank you.

The idea is to have a separate field in the `(non-thrift) metadata object` to store the NaN stats. But the usage of these NaN stats is not clear from this PR.

PingLiuPing:
> @PingLiuPing How do other tools like DuckDB write iceberg files without this NaN statistics present in the parquet.thrift file?

@majetideepak  Thanks. For DuckDB-iceberg, they do not write this stat yet. They also do not support other stats at the moment. They have a PR to collect some of the stats (not NaN) https://github.com/duckdb/duckdb-iceberg/pull/640.

PingLiuPing:
> The idea is to have a separate field in the (non-thrift) metadata object to store the NaN stats. 

Ok, I will try to see if this is possible.

> But the usage of these NaN stats is not clear from this PR.

The stats are not used inside velox but it should be collected and saved to iceberg manifest file. It is used by upstream engine when planning.

majetideepak:
>The stats are not used inside velox but it should be collected and saved to iceberg manifest file. It is used by upstream engine when planning.

I am reading this as the coordinator requires these stats for better planning. The NaN stats are not written to Parquet but to an Iceberg manifest file. We should aim to store these stats separately as they are not related to Parquet anyway.

PingLiuPing:
> I am reading this as the coordinator requires these stats for better planning. The NaN stats are not written to Parquet but to an Iceberg manifest file. We should aim to store these stats separately as they are not related to Parquet anyway.

Yes, with the current paquet.thrift, NaN is not write to parquet file. Once https://github.com/apache/parquet-format/pull/514/files#top  is merged and Velox parquet writer is upgraded NaN will be wrote to parquet file.

My initial design is just collect the NaN stats and report it back to coordinator node, and do not write NaN to parquet file.
My first version of code follows this design, but since it changed `ParquetThriftTypes.h` and `ParquetThriftTypes.cpp` and following the review comments and seems this is not the most correct way to do this. Nevertheless I noteced that `ParquetThriftTypes.h` and `ParquetThriftTypes.cpp` have been changed many times.

majetideepak:
> Nevertheless I noteced that ParquetThriftTypes.h and ParquetThriftTypes.cpp have been changed many times.

These files will anyway go away with this PR https://github.com/facebookincubator/velox/pull/14942

PingLiuPing:
> > Nevertheless I noteced that ParquetThriftTypes.h and ParquetThriftTypes.cpp have been changed many times.
> 
> These files will anyway go away with this PR #14942

Oh, thanks. Let me investigate if there are ways to do this without code change in ParquetThriftTypes.h and ParquetThriftTypes.cpp

PingLiuPing:
@majetideepak  I re-factored the code and removed all the changes relate to `parquet.thrift`. Would you have another look? Thank you.

PingLiuPing:
@majetideepak  Could you please have a look? Thank you.

PingLiuPing:
@majetideepak  Could you please have a look? Thank you.

PingLiuPing:
@rui-mo @mbasmanova Wondering if you could take a look at this PR when convenient.

mbasmanova (approved):
@PingLiuPing Thanks.

It is hard to read this code because coding style is inconsistent and so different from Velox. If you plan to continue to develop this code, please, consider updating Parquet writer wholesale to follow Velox coding style?

CC: @pedroerp

PingLiuPing:
> @PingLiuPing Thanks.
> 
> It is hard to read this code because coding style is inconsistent and so different from Velox. If you plan to continue to develop this code, please, consider updating Parquet writer wholesale to follow Velox coding style?
> 
> CC: @pedroerp

@mbasmanova  Thank you so much for reviewing.

That makes sense. I’ll open an issue to track this and draft a plan to the Parquet writer to better align with Velox coding style.

mbasmanova:
> I’ll open an issue to track this and draft a plan to the Parquet writer to better align with Velox coding style.

@PingLiuPing Thank you. I expect this can be done using an LLM relatively quickly.

PR #5946, opened 2023-08-01
Refactor compression classes into common package

nmahadevuni:
Move classes from dwrf package to dwio/common package. This can later be used for parquet too.

PR #14726, opened 2025-09-04
fix(dwio): Fix decompression in ZlibDecompressionStream readOrSkip

xin-zhang2:
Fixes https://github.com/facebookincubator/velox/issues/14740

### DIFF
--- a/velox/dwio/common/compression/Compression.cpp
+++ b/velox/dwio/common/compression/Compression.cpp
@@ -555,6 +555,8 @@ bool ZlibDecompressionStream::readOrSkip(const void** data, int32_t* size) {
     *size = static_cast<int32_t>(availSize);
     outputBufferPtr_ = inputBufferPtr_ + availSize;
     outputBufferLength_ = 0;
+    inputBufferPtr_ += availSize;
+    remainingLength_ -= availSize;
   } else {
     DWIO_ENSURE_EQ(
         state_,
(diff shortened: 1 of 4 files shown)


xin-zhang2:
@zacw7 @Yuhta 
I'm running the TextReader integration tests in Presto and found that the blockSize we chose in https://github.com/facebookincubator/velox/pull/14003#discussion_r2288323525 could cause memory allocation issues. This PR sets the output buffer to three times the compressed length initially and allow it to dynamically expand during decompression, which I believe is more reasonable. It also fixes a misuse of inflate in ZlibDecompressionStream.
Could you please take a look at your convenience? Thanks!

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -191,34 +191,15 @@ TextRowReader::TextRowReader(
...
     contents_->inputStream = contents_->input->loadCompleteFile();
     auto name = contents_->inputStream->getName();
     contents_->decompressedInputStream = createDecompressor(
         contents_->compression,
         std::move(contents_->inputStream),
-        blockSize,
+        // An estimated value used as the output buffer size for the zlib
+        // decompressor, and as the fallback value of the decompressed length
+        // for other decompressors.
+        3 * contents_->fileLength,
```
zacw7: shall we as define the magic number `3` a named constant?

xin-zhang2: @zacw7 Thanks for the review. I added a constant `kDecompressionBufferFactor` for it.

xin-zhang2:
> is it possible to add unit tests for this ?

@zacw7 Yes, I added a new case to test the zlib decompression when the blockSize is smaller than the decompressed length. dwio/common/compression is from the refactor in https://github.com/facebookincubator/velox/pull/5946, and all tests are still under dwio/dwrf/tests, so I added the case there as well.

zacw7 (approved):

xin-zhang2:
@Yuhta Could you please help review this PR when you have time? Thanks!

kKPulla:
@xin-zhang2 can you please rebase on main (on top of #15100 ) to make sure the signals turn green? Thanks!

@zacw7 I see you already imported the PR, will you be merging this or should I take over?

xin-zhang2:
@kKPulla Thanks. I’ve rebased, and all checks have passed. Can you take a look?

velox/dwio/common/compression/Compression.cpp
```diff
@@ -567,51 +569,52 @@ bool ZlibDecompressionStream::readOrSkip(const void** data, int32_t* size) {
...
+        outputBuffer_->extend(*size);
+        outputBufferPtr_ = outputBuffer_->data();
+        zstream_.next_out = reinterpret_cast<Bytef*>(
+            const_cast<char*>(outputBufferPtr_ + *size));
+        zstream_.avail_out = folly::to<uInt>(blockSize_);
+        result = inflate(&zstream_, Z_SYNC_FLUSH);
+        // Result handling adapted from https://zlib.net/zlib_how.html
+        switch (result) {
+          case Z_NEED_DICT:
+            result = Z_DATA_ERROR;
```
kKPulla, line 595: don't have much context on the change, but is the fall through expected behavior here? Asking because I see a bunch of errors in our internal jobs complaining about this (and suggesting to add a break), blocking the merge. 

cc: @Yuhta

xin-zhang2: Yes, the fall through is expected, and I've updated the code to use fall through attribute to prevent the compiler warnings. Can you re-import the PR?

PR #14727, opened 2025-09-04
feat: Add trailing zero seconds option

PingLiuPing:
Add skipTrailingZeroSeconds option to TimestampToStringOptions

This PR is part of the iceberg insertion feature. As per review comments we should split PR to smaller pieces.
This option allows omitting trailing zero seconds and the preceding colon from timestamp strings. This matches the iceberg specification.


velox/type/Timestamp.cpp
```diff
@@ -268,17 +268,22 @@ StringView Timestamp::tmToStringView(
...
   if (options.precision == TimestampToStringOptions::Precision::kMilliseconds) {
     nanos /= 1'000'000;
   } else if (
       options.precision == TimestampToStringOptions::Precision::kMicroseconds) {
     nanos /= 1'000;
   }
-  if (options.skipTrailingZeros && nanos == 0) {
+
+  // Second.
+  const bool shouldSkipSeconds =
```
jinchengchenghh: Remove the used only once variable

jinchengchenghh:
Remove from PR description

> As per review comments we should split PR to smaller pieces.

velox/type/Timestamp.cpp
PingLiuPing: Thanks, it is been used twice actually. Should I remove it?

velox/type/Timestamp.h
```diff
@@ -48,6 +48,12 @@ struct TimestampToStringOptions {
   /// '2000-01-01 12:21:56.129000' becomes '2000-01-01 12:21:56.129'.
   bool skipTrailingZeros = false;
 
+  /// Whether to skip trailing zeros of second part. E.g. when true,
+  /// '2000-01-01 12:21:00' becomes '2000-01-01 12:21'.
+  /// '2000-01-01 12:21:00.000' becomes '2000-01-01 12:21'.
+  /// '2000-01-01 12:21:00.123' will not be impacted by this option.
```
rui-mo: What is the behavior for '2000-01-01 12:21:50', would the result become '2000-01-01 12:21:5'?

PingLiuPing: Thanks.
'2000-01-01 12:21:50' will not be changed as there is valid seconds part, i.e `tmValue.tm_sec != 0`.

rui-mo: > This matches the iceberg partition name specification.

It might help to provide the specification in the PR description, thanks.

PingLiuPing: Yes, you are right. Let me find it and also add '2000-01-01 12:21:50' in the comment. And also a test case.

rui-mo: Perhaps also add test for case like '2000-01-01 12:21:00.500'.

PingLiuPing: Thanks, I added a similar case.

PR #14733, opened 2025-09-04
refactor: All function literals that registered without prefix should be listed in single place (with Spark)

MBkkt:
Also some of these literals actively used in code, so velox code sometimes rely on fact that these functions are literals.

Now it should be clear from code.

I didn't touch `experimental/` folder, benchmarks, tests and something like duckdb parser utils

### DIFF
--- a/velox/core/FilterToExpression.cpp
+++ b/velox/core/FilterToExpression.cpp
@@ -16,6 +16,7 @@
 
 #include "velox/core/FilterToExpression.h"
 #include "velox/core/Expressions.h"
+#include "velox/expression/ExprConstants.h"
 
 namespace facebook::velox::core {
 
@@ -30,7 +31,8 @@ core::TypedExprPtr createBooleanExpr(
   } else if (conditions.size() == 1) {
     return conditions[0];
   } else {
-    return std::make_shared<CallTypedExpr>(BOOLEAN(), conditions, "and");
+    return std::make_shared<CallTypedExpr>(
+        BOOLEAN(), conditions, expression::kAnd);
   }
 }
 
@@ -43,11 +45,11 @@ core::TypedExprPtr handleNullAllowed(
     bool nullAllowed) {
   if (nullAllowed) {
     // If nullAllowed=true: (expression) OR (field IS NULL)
-    auto isNullExpr =
-        std::make_shared<CallTypedExpr>(BOOLEAN(), "is_null", subfieldExpr);
+    auto isNullExpr = std::make_shared<CallTypedExpr>(
+        BOOLEAN(), expression::kIsNull, subfieldExpr);
 
     return std::make_shared<CallTypedExpr>(
-        BOOLEAN(), "or", expression, isNullExpr);
+        BOOLEAN(), expression::kOr, expression, isNullExpr);
   } else {
     // If nullAllowed=false: just return the expression
     return expression;
@@ -125,12 +127,13 @@ core::TypedExprPtr filterToExpr(
 
     case common::FilterKind::kIsNull:
       return std::make_shared<CallTypedExpr>(
-          BOOLEAN(), "is_null", subfieldExpr);
+          BOOLEAN(), expression::kIsNull, subfieldExpr);
 
     case common::FilterKind::kIsNotNull: {
-      auto isNullExpr =
-          std::make_shared<CallTypedExpr>(BOOLEAN(), "is_null", subfieldExpr);
-      return std::make_shared<CallTypedExpr>(BOOLEAN(), "not", isNullExpr);
+      auto isNullExpr = std::make_shared<CallTypedExpr>(
+          BOOLEAN(), expression::kIsNull, subfieldExpr);
+      return std::make_shared<CallTypedExpr>(
+          BOOLEAN(), expression::bug::kNot, isNullExpr);
     }
 
     case common::FilterKind::kBoolValue: {
@@ -139,7 +142,7 @@ core::TypedExprPtr filterToExpr(
           BOOLEAN(), variant(boolFilter->testBool(true)));
 
       auto eqExpr = std::make_shared<CallTypedExpr>(
-          BOOLEAN(), "eq", subfieldExpr, boolValue);
+          BOOLEAN(), expression::bug::kEq, subfieldExpr, boolValue);
 
       return handleNullAllowed(eqExpr, subfieldExpr, boolFilter->nullAllowed());
     }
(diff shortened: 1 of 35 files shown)


velox/functions/sparksql/ConcatWs.cpp
```diff
@@ -383,7 +384,7 @@ exec::ExprPtr ConcatWsCallToSpecialForm::constructSpecialForm(
       std::move(args),
       std::move(concatWsFunction),
       exec::VectorFunctionMetadataBuilder().defaultNullBehavior(false).build(),
-      kConcatWs,
+      expression::old::kConcatWs,
```
mbasmanova, line 387: SparkSQL is an extention to Velox. It uses its own Spark-compatible function names. CC: @rui-mo

velox/core/FilterToExpression.cpp
```diff
@@ -152,12 +154,13 @@ core::TypedExprPtr filterToExpr(
...
+          BOOLEAN(), expression::kIsNull, subfieldExpr);
 
     case common::FilterKind::kIsNotNull: {
-      auto isNullExpr =
-          std::make_shared<CallTypedExpr>(BOOLEAN(), "is_null", subfieldExpr);
-      return std::make_shared<CallTypedExpr>(BOOLEAN(), "not", isNullExpr);
+      auto isNullExpr = std::make_shared<CallTypedExpr>(
+          BOOLEAN(), expression::kIsNull, subfieldExpr);
+      return std::make_shared<CallTypedExpr>(
+          BOOLEAN(), expression::bug::kNot, isNullExpr);
```
Yuhta, line 136: @juwentus1234 Let's add a `prefix` parameter to the function

velox/functions/sparksql/ConcatWs.cpp
MBkkt: But these names are kind of reserved from common namespace.
Imagine we will need to develop some function that has same name, if we don't want to break spark, we cannot use same name. Because these names used without prefix.

But maybe I should separate this PR to few smaller, ok

MBkkt: So my point, is all names that are registered without prefix aren't really extensions, because they're reserved from non-prefixed name space.

MBkkt: https://github.com/facebookincubator/velox/pull/14734

@mbasmanova I made same PR but without touching spark

velox/dwio/common/MetadataFilter.cpp
```diff
@@ -163,19 +164,19 @@ std::unique_ptr<MetadataFilter::Node> MetadataFilter::Node::fromExpression(
...
   }
-  if (call->name() == "or") {
+  if (call->name() == expression::kOr) {
     auto lhs = fromExpression(*call->inputs()[0], evaluator, negated);
     auto rhs = fromExpression(*call->inputs()[1], evaluator, negated);
     return negated ? AndNode::create(std::move(lhs), std::move(rhs))
                    : OrNode::create(std::move(lhs), std::move(rhs));
   }
-  if (call->name() == "not") {
+  if (call->name() == expression::bug::kNot) {
```
Yuhta, line 179: This one we maybe need a singleton to retrieve the prefix.  Or we resolve the name (we have evaluator here) to make sure it is indeed NOT function

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -477,47 +478,47 @@ PrestoExprToSubfieldFilterParser::leafCallToSubfieldFilter(
 
   const auto* leftSide = call.inputs()[0].get();
 
-  if (call.name() == "eq") {
+  if (call.name() == expression::bug::kEq) {
```
Yuhta, line 481: I think we can add a functionality to evaluator to resolve function names and return some canonical names, so we can do these expression transformation for well known functions.

velox/expression/ExprConstants.h
```diff
@@ -15,16 +15,67 @@
...
+inline constexpr const char* kRoundDecimal = "decimal_round";
+inline constexpr const char* kAtLeastNNonNulls = "at_least_n_non_nulls";
+inline constexpr const char* kFromJson = "from_json";
+inline constexpr const char* kGetStructField = "get_struct_field";
+inline constexpr const char* kGetArrayStructFields = "get_array_struct_fields";
+inline constexpr const char* kConcatWs = "concat_ws";
+
+} // namespace old
+
+/// TODO: If you're using these literals you don't account that these functions
```
Yuhta, line 64: Shall we consider something like `call.name() == prefix + kNot` safe?

velox/expression/ExprToSubfieldFilter.cpp
MBkkt: Maybe such approach can be used?

```
bool isNotExpr(
    const core::CallTypedExpr* call,
    core::ExpressionEvaluator* evaluator) {
  if (!endWith(call->name(), "not")) {
    return false;
  }
  auto exprs = evaluator->compile(expr);
  VELOX_CHECK_EQ(exprs->size(), 1);
  auto& compiled = exprs->expr(0);
  return compiled->vectorFunction() &&
      compiled->vectorFunction()->getCanonicalName() ==
      exec::FunctionCanonicalName::kNot;
}

```

velox/expression/ExprConstants.h
MBkkt: I think yes? Because I don't see reason why no, the question is where we will lookup prefix

velox/dwio/common/MetadataFilter.cpp
MBkkt: https://github.com/facebookincubator/velox/blob/09d2f0c724b2e7c3325dc5c6bad34a16297cd374/velox/connectors/hive/HiveConnectorUtil.cpp#L818

velox/expression/ExprConstants.h
```diff
@@ -15,16 +15,67 @@
...
+inline constexpr const char* kTry = "try";
+
+/// spark special form functions without prefix
+inline constexpr const char* kMakeDecimal = "make_decimal";
+inline constexpr const char* kRoundDecimal = "decimal_round";
+inline constexpr const char* kAtLeastNNonNulls = "at_least_n_non_nulls";
+inline constexpr const char* kFromJson = "from_json";
+inline constexpr const char* kGetStructField = "get_struct_field";
+inline constexpr const char* kGetArrayStructFields = "get_array_struct_fields";
+inline constexpr const char* kConcatWs = "concat_ws";
```
rui-mo, line 60: These seem to be specific to Spark. How about moving them under sparksql to avoid exposing them in the common library? Thanks.

MBkkt: In such case more easy to just avoid this PR (I made same PR but without spark), the whole point was to have list names that doesn't have prefixes so cannot be easily used even for different runtime.

velox/dwio/common/MetadataFilter.cpp
Yuhta: This is an one-off function to do the right thing (with some unnecessary overhead, because we only need to resolve the function in root node, no need to do it recursively).  Wondering if we should have a formal interface on `ExpressionEvaluator` to do exactly this.

PR #14734, opened 2025-09-04
refactor: All function literals that registered without prefix should be listed in single place

MBkkt:
Also some of these literals actively used in code, so velox code sometimes rely on fact that these functions are literals. Now it should be clear from code.

This doesn't account spark.


velox/core/FilterToExpression.cpp
```diff
@@ -139,7 +142,7 @@ core::TypedExprPtr filterToExpr(
           BOOLEAN(), variant(boolFilter->testBool(true)));
 
       auto eqExpr = std::make_shared<CallTypedExpr>(
-          BOOLEAN(), "eq", subfieldExpr, boolValue);
+          BOOLEAN(), expression::bug::kEq, subfieldExpr, boolValue);
```
juwentus1234: expression::bug::kEq looks a bit strange, should we use something in bug for no bug logic for consistency

velox/core/FilterToExpression.cpp
```diff
@@ -164,7 +167,7 @@ core::TypedExprPtr filterToExpr(
               variant(static_cast<int32_t>(rangeFilter->lower())));
 
           conditions.push_back(std::make_shared<CallTypedExpr>(
-              BOOLEAN(), "gte", subfieldExpr, lower));
+              BOOLEAN(), expression::bug::kGte, subfieldExpr, lower));
```
juwentus1234: ditto

velox/core/FilterToExpression.cpp
MBkkt: We can only do something like prefix + "eq"?

MBkkt: prefix probably can be function parameter I think, so user of this function should specify it?

velox/core/FilterToExpression.cpp
mbasmanova: > velox/core/FilterToExpression.cpp
 
@MBkkt @juwentus1234 It looks like this is a test-only utility. If so, then we should just move it out of "prod" directory and keep hard-coded function names.

MBkkt:
@mbasmanova Can you help? What are next steps for this PR?

mbasmanova:
@MBkkt I don't think we need these changes. You pointed out a real problem where HiveConnector doesn't handle function names properly and I believe we have created an issue to address that.

MBkkt:
@mbasmanova Why not? The point of this change not only pointing to the issue, but also allow to easy fix it with patch on client side, while the proper fix isn't done. Is something wrong with such approach?
Maybe you dislike `bug` or `old` namespace name? I can consider different names. 

It also contains few improvements:
1. `row_constructor`,  `in`, `is_null` constants
2. using `ExprConstant` where it's needed instead of raw string literals. 
3. removing unused literals to `kFail`

Should I put these in a separate PR?

mbasmanova:
@MBkkt I think it would be helpful to update the code to replace hard-coded special form names with constants defined in ExprConstants.h. It would be nice to add "row_constructor" constant to ExprConstants.h as well. However, let's not add any other constants.

MBkkt:
@mbasmanova Done.

> However, let's not add any other constants.

I don't understand why, but ok, it's still better than nothing.

velox/expression/ExprConstants.h
```diff
@@ -15,16 +15,24 @@
  */
 #pragma once
 
+/// This file contains string literals for function names.
```
mbasmanova: Please, remove this comment. It is not correct.

velox/expression/ExprConstants.h
```diff
@@ -15,16 +15,24 @@
...
-constexpr const char* kAnd = "and";
-constexpr const char* kOr = "or";
-constexpr const char* kSwitch = "switch";
-constexpr const char* kIf = "if";
-constexpr const char* kFail = "fail";
-constexpr const char* kCoalesce = "coalesce";
-constexpr const char* kCast = "cast";
-constexpr const char* kTryCast = "try_cast";
-constexpr const char* kTry = "try";
+/// These functions registered without prefix and velox code rely on this.
```
mbasmanova: Please, remove this comment.

velox/expression/ExprConstants.h
```diff
@@ -15,16 +15,24 @@
...
+inline constexpr const char* kAnd = "and";
+inline constexpr const char* kOr = "or";
+inline constexpr const char* kSwitch = "switch";
+inline constexpr const char* kIf = "if";
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
```
mbasmanova: Please, remove this.

velox/expression/RowConstructor.h
```diff
@@ -29,6 +29,8 @@ class RowConstructorCallToSpecialForm : public FunctionCallToSpecialForm {
       bool trackCpuUsage,
       const core::QueryConfig& config) override;
 
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
```
mbasmanova: Do we need this? Why?

velox/expression/ExprConstants.h
MBkkt: Why is it not correct? But ok, I removed it

velox/expression/ExprConstants.h
MBkkt: Ok

velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
```diff
@@ -101,7 +101,7 @@ void registerGeneralFunctions(const std::string& prefix) {
 
   registerFailFunction({prefix + "fail"});
 
-  registerAllSpecialFormGeneralFunctions();
+  registerAllSpecialFormGeneralFunctions(prefix);
```
mbasmanova: Can we remove this change?

MBkkt:
@mbasmanova I applied your suggestions

mbasmanova (approved):
Thanks.

Issue #14736, opened 2025-09-05
undefined reference to facebook::velox::dwio::common::compression::createDecompressor


JkSelf:
### Problem description

Encounter the following link error with `make debug` command.
```
/usr/bin/ld: velox/dwio/text/reader/libvelox_dwio_text_reader.a(TextReader.cpp.o): in function `facebook::velox::text::TextRowReader::TextRowReader(std::shared_ptr<facebook::velox::text::FileContents>, facebook::velox::dwio::common::RowReaderOptions const&)':
/home/lab01/velox/velox/dwio/text/reader/TextReader.cpp:218: undefined reference to `facebook::velox::dwio::common::compression::createDecompressor(facebook::velox::common::CompressionKind, std::unique_ptr<facebook::velox::dwio::common::SeekableInputStream, std::default_delete<facebook::velox::dwio::common::SeekableInputStream> >, unsigned long, facebook::velox::memory::MemoryPool&, facebook::velox::dwio::common::compression::CompressionOptions const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, facebook::velox::dwio::common::encryption::Decrypter const*, bool, unsigned long)'

```

### System information

Velox System Info v0.0.2
Commit: 0d59d2505f08ec8a1163d52745061d49a6578bb6
CMake Version: 3.28.3
System: Linux-5.4.0-204-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.4.0
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib/python3.8/dist-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt

### CMake log

```bash

```

kou:
I think that this is duplicated of #14673.

JkSelf:
Duplicated with https://github.com/facebookincubator/velox/issues/14673.

PR #14674, opened 2025-09-02
fix(build): Fix wrong link target for velox_dwio_text_reader

kou:
Fixes #14673

This may be related to #14003. `velox/dwio/text/reader/TextReader.cpp` uses
`facebook::velox::dwio::common::compression::createDecompressor()` since #14003.
`facebook::velox::dwio::common::compression::createDecompressor()` is included in `velox_dwio_common_compression` not `velox_dwio_common`. So `velox_dwio_text_reader` needs to be linked to `velox_dwio_common_compression` not `velox_dwio_common`.

PR #14737, opened 2025-09-05
fix: Fix undefined reference to facebook::velox::dwio::common::compression::createDecompressor

JkSelf:
Fix https://github.com/facebookincubator/velox/issues/14736


xin-zhang2:
@JkSelf Thanks for the fix. There is another fix in https://github.com/facebookincubator/velox/pull/14674 that replaces  velox_dwio_common with velox_dwio_common_compression. Since the text reader uses files from both, I think your approach is more accurate. I also left a comment in that PR.

JkSelf:
Duplicated with https://github.com/facebookincubator/velox/pull/14674

PR #14739, opened 2025-09-05
feat: Add current_timezone function

anandamideShakyan:
In the past, we were not implementing functions like current_timezone that returned constants in Velox. This was because Presto Java would infer the constant value at the co-ordinator and fold it, so the worker was never passed this function in execution.

However, with recent changes for side-car and worker based constant folding implementation, we need native implementations for functions like current_timezone, so adding to Velox as well.

### DIFF
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -1782,6 +1782,21 @@ struct CurrentDateFunction {
 };
 
 template <typename T>
+struct CurrentTimezoneFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+  std::string tzName_;
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>&,
+      const core::QueryConfig& config) {
+    tzName_ = config.sessionTimezone();
+  }
+  FOLLY_ALWAYS_INLINE void call(out_type<Varchar>& result) {
+    result = std::string_view(tzName_);
+  }
+};
+
+template <typename T>
 struct TimeZoneHourFunction : public TimestampWithTimezoneSupport<T> {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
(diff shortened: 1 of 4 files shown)


velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
```diff
@@ -89,6 +89,8 @@ void registerSimpleFunctions(const std::string& prefix) {
       {prefix + "to_unixtime"});
 
   registerFromUnixtime(prefix + "from_unixtime");
+  registerFunction<CurrentTimezoneFunction, Varchar>({"current_timezone"});
```
MBkkt:
```suggestion
  registerFunction<CurrentTimezoneFunction, Varchar>({prefix + "current_timezone"});
```
shouldn't be prefix?

anandamideShakyan: Yes. I missed it. Thanks.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1528,6 +1528,21 @@ struct CurrentDateFunction {
   }
 };
 
+template <typename T>
+struct CurrentTimezoneFunction {
```
kgpai, line 1785: I am wondering if it is more efficient to implement this as a vector function that returns a constant vector with the tzName. Is it mainly used in constant folding ?

anandamideShakyan: Yes. It is mainly used in constant folding. Should I still be implementing it as a vector function ? Will the performance gains be that huge in that use case ?

kgpai: I think its better to implement as a vector function as it will be quite a lot more performant.

aditi-pandit: @kgpai : Agree that vector function returning ConstantVector is more performant... but if it is always constant folded then it seemed alright to make it a simple function. Its not ever going to be called during expression evaluation with vector inputs.

@mbasmanova :wdyt ?

kgpai: If its always constant folded then yes it doesnt matter ; lets go with simple function implementation. If that is the case we can do this for current_time, localtime etc.

PR #14741, opened 2025-09-05
feat: Support write field_id to parquet metadata SchemaElement

PingLiuPing:
Parquet writer does NOT write `field_id`  to file metadata.

As per apache-iceberg [parquet specification](https://iceberg.apache.org/spec/#parquet), it requires that iceberg column IDs should be stored as field-ids in parquet schema.

Solution:
Creates metadata value with the key `PARQUET:field_id` to determine the field_id when converting an arrow schema into a parquet schema. If there is no such metadata entry then the field_id will not be present. And we can use existing function `FieldIdMetadata` to create such metadata.
To add field_id to nested data types, we can leverage `updateFieldNameRecursive` method.

### DIFF
--- a/velox/dwio/parquet/writer/Writer.cpp
+++ b/velox/dwio/parquet/writer/Writer.cpp
@@ -22,6 +22,7 @@
 #include "velox/common/config/Config.h"
 #include "velox/common/testutil/TestValue.h"
 #include "velox/core/QueryConfig.h"
+#include "velox/dwio/parquet/writer/arrow/ArrowSchema.h"
 #include "velox/dwio/parquet/writer/arrow/Properties.h"
 #include "velox/dwio/parquet/writer/arrow/Writer.h"
 #include "velox/exec/MemoryReclaimer.h"
@@ -193,47 +194,70 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
   }
 }
 
-std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
+std::shared_ptr<::arrow::Field> updateFieldRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const ParquetFieldId* fieldId = nullptr) {
+  std::shared_ptr<::arrow::Field> newField{nullptr};
   if (type.isRow()) {
     auto& rowType = type.asRow();
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
     auto childrenSize = rowType.size();
     std::vector<std::shared_ptr<::arrow::Field>> newFields;
     newFields.reserve(childrenSize);
-    for (auto i = 0; i < childrenSize; i++) {
-      newFields.push_back(updateFieldNameRecursive(
-          structType->fields()[i], *rowType.childAt(i), rowType.nameOf(i)));
+    for (auto i = 0; i < childrenSize; ++i) {
+      const auto* childSetting = fieldId ? &fieldId->children.at(i) : nullptr;
+      newFields.push_back(updateFieldRecursive(
+          structType->fields()[i],
+          *rowType.childAt(i),
+          rowType.nameOf(i),
+          childSetting));
     }
-    return newField->WithType(::arrow::struct_(newFields));
+    newField = newField->WithType(::arrow::struct_(newFields));
   } else if (type.isArray()) {
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto listType =
         std::dynamic_pointer_cast<::arrow::BaseListType>(newField->type());
     auto elementType = type.asArray().elementType();
     auto elementField = listType->value_field();
-    return newField->WithType(
-        ::arrow::list(updateFieldNameRecursive(elementField, *elementType)));
+    const auto* childSetting = fieldId ? &fieldId->children.at(0) : nullptr;
+    auto updatedElementField =
+        updateFieldRecursive(elementField, *elementType, name, childSetting);
+    newField = newField->WithType(::arrow::list(updatedElementField));
   } else if (type.isMap()) {
     auto mapType = type.asMap();
-    auto newField = field->WithName(name);
+    newField = field->WithName(name);
     auto arrowMapType =
         std::dynamic_pointer_cast<::arrow::MapType>(newField->type());
-    auto newKeyField =
-        updateFieldNameRecursive(arrowMapType->key_field(), *mapType.keyType());
-    auto newValueField = updateFieldNameRecursive(
-        arrowMapType->item_field(), *mapType.valueType());
-    return newField->WithType(
-        ::arrow::map(newKeyField->type(), newValueField->type()));
+    const auto* keySetting = fieldId ? &fieldId->children.at(0) : nullptr;
+    const auto* valueSetting = fieldId ? &fieldId->children.at(1) : nullptr;
+    auto newKeyField = updateFieldRecursive(
+        arrowMapType->key_field(),
+        *mapType.keyType(),
+        mapType.nameOf(0),
+        keySetting);
+    auto newValueField = updateFieldRecursive(
+        arrowMapType->item_field(),
+        *mapType.valueType(),
+        mapType.nameOf(1),
+        valueSetting);
+    newField = newField->WithType(
+        std::make_shared<::arrow::MapType>(newKeyField, newValueField));
   } else if (name != "") {
-    return field->WithName(name);
+    newField = field->WithName(name);
   } else {
-    return field;
+    newField = field;
   }
+
+  if (fieldId) {
+    newField =
+        newField->WithMetadata(arrow::arrow::FieldIdMetadata(fieldId->fieldId));
+  }
+
+  return newField;
 }
 
 std::optional<TimestampPrecision> getTimestampUnit(
@@ -351,6 +375,7 @@ Writer::Writer(
   setMemoryReclaimers();
   writeInt96AsTimestamp_ = options.writeInt96AsTimestamp;
   arrowMemoryPool_ = options.arrowMemoryPool;
+  parquetFieldIds_ = options.parquetFieldIds;
 }
 
 Writer::Writer(
@@ -446,8 +471,11 @@ void Writer::write(const VectorPtr& data) {
   std::vector<std::shared_ptr<::arrow::Field>> newFields;
   auto childSize = schema_->size();
   for (auto i = 0; i < childSize; i++) {
-    newFields.push_back(updateFieldNameRecursive(
-        arrowSchema->fields()[i], *schema_->childAt(i), schema_->nameOf(i)));
+    newFields.push_back(updateFieldRecursive(
+        arrowSchema->fields()[i],
+        *schema_->childAt(i),
+        schema_->nameOf(i),
+        parquetFieldIds_ ? &parquetFieldIds_->at(i) : nullptr));
   }
 
   PARQUET_ASSIGN_OR_THROW(
--- a/velox/dwio/parquet/writer/Writer.h
+++ b/velox/dwio/parquet/writer/Writer.h
@@ -88,6 +88,13 @@ class LambdaFlushPolicy : public DefaultFlushPolicy {
   std::function<bool()> lambda_;
 };
 
+// Settings for specifying Parquet field IDs during write operations.
+// Used to explicitly control field ID assignment in the Parquet schema.
+struct ParquetFieldId {
+  int32_t fieldId;
+  std::vector<ParquetFieldId> children;
+};
+
 struct WriterOptions : public dwio::common::WriterOptions {
   // Growth ratio passed to ArrowDataBufferSink. The default value is a
   // heuristic borrowed from
@@ -115,6 +122,8 @@ struct WriterOptions : public dwio::common::WriterOptions {
   std::optional<std::string> createdBy;
 
   std::shared_ptr<arrow::MemoryPool> arrowMemoryPool;
+
+  std::shared_ptr<std::vector<ParquetFieldId>> parquetFieldIds;
 
   // Parsing session and hive configs.
 
@@ -213,6 +222,8 @@ class Writer : public dwio::common::Writer {
   std::shared_ptr<ArrowDataBufferSink> stream_;
 
   std::shared_ptr<ArrowContext> arrowContext_;
+
+  std::shared_ptr<std::vector<ParquetFieldId>> parquetFieldIds_;
 
   std::unique_ptr<DefaultFlushPolicy> flushPolicy_;
 
--- a/velox/dwio/parquet/writer/arrow/ArrowSchema.cpp
+++ b/velox/dwio/parquet/writer/arrow/ArrowSchema.cpp
@@ -64,6 +64,8 @@ using ParquetType = Type;
 
 namespace {
 
+static constexpr char FIELD_ID_KEY[] = "PARQUET:field_id";
+
 /// Increments levels according to the cardinality of node.
 void IncrementLevels(LevelInfo& current_levels, const schema::Node& node) {
   if (node.is_repeated()) {
@@ -342,16 +344,6 @@ static Status GetTimestampMetadata(
   }
 
   return Status::OK();
-}
-
-static constexpr char FIELD_ID_KEY[] = "PARQUET:field_id";
-
-std::shared_ptr<::arrow::KeyValueMetadata> FieldIdMetadata(int field_id) {
-  if (field_id >= 0) {
-    return ::arrow::key_value_metadata({FIELD_ID_KEY}, {ToChars(field_id)});
-  } else {
-    return nullptr;
-  }
 }
 
 int FieldIdFromMetadata(
@@ -1221,6 +1213,14 @@ Result<bool> ApplyOriginalMetadata(
 }
 
 } // namespace
+
+std::shared_ptr<::arrow::KeyValueMetadata> FieldIdMetadata(int field_id) {
+  if (field_id >= 0) {
+    return ::arrow::key_value_metadata({FIELD_ID_KEY}, {ToChars(field_id)});
+  } else {
+    return nullptr;
+  }
+}
 
 Status FieldToNode(
     const std::shared_ptr<Field>& field,
--- a/velox/dwio/parquet/writer/arrow/ArrowSchema.h
+++ b/velox/dwio/parquet/writer/arrow/ArrowSchema.h
@@ -195,5 +195,7 @@ struct PARQUET_EXPORT SchemaManifest {
   }
 };
 
+std::shared_ptr<::arrow::KeyValueMetadata> FieldIdMetadata(int32_t field_id);
+
 } // namespace arrow
 } // namespace facebook::velox::parquet::arrow
--- a/velox/dwio/parquet/writer/arrow/Metadata.cpp
+++ b/velox/dwio/parquet/writer/arrow/Metadata.cpp
@@ -389,6 +389,10 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
     return std::nullopt;
   }
 
+  inline int32_t field_id() const {
+    return descr_->schema_node()->field_id();
+  }
+
  private:
   mutable std::shared_ptr<Statistics> possible_stats_;
   std::vector<Encoding::type> encodings_;
@@ -533,6 +537,10 @@ int64_t ColumnChunkMetaData::total_uncompressed_size() const {
 
 int64_t ColumnChunkMetaData::total_compressed_size() const {
   return impl_->total_compressed_size();
+}
+
+int32_t ColumnChunkMetaData::field_id() const {
+  return impl_->field_id();
 }
 
 std::unique_ptr<ColumnCryptoMetaData> ColumnChunkMetaData::crypto_metadata()
--- a/velox/dwio/parquet/writer/arrow/Metadata.h
+++ b/velox/dwio/parquet/writer/arrow/Metadata.h
@@ -187,6 +187,7 @@ class PARQUET_EXPORT ColumnChunkMetaData {
   int64_t index_page_offset() const;
   int64_t total_compressed_size() const;
   int64_t total_uncompressed_size() const;
+  int32_t field_id() const;
   std::unique_ptr<ColumnCryptoMetaData> crypto_metadata() const;
   std::optional<IndexLocation> GetColumnIndexLocation() const;
   std::optional<IndexLocation> GetOffsetIndexLocation() const;


velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,103 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+      std::move(sink), writerOptions, rootPool_, schema);
+  writer->write(data);
+  writer->close();
+
+  std::string_view sinkData(sinkPtr->data(), sinkPtr->size());
+  auto raFile = std::make_shared<::arrow::io::BufferReader>(
+      std::make_shared<::arrow::Buffer>(
+          reinterpret_cast<const uint8_t*>(sinkData.data()), sinkData.size()));
+
+  auto fileReader = parquet::arrow::ParquetFileReader::Open(raFile);
```
yingsu00: @PingLiuPing Could you please add a test using the native Parquet reader?

PingLiuPing: @yingsu00  Thank you for the review comments. Will look into this.

I do have plan on a broader scope to fix the schema evolution problem on IcebergSplitReader. https://github.com/facebookincubator/velox/discussions/6065

PingLiuPing: @yingsu00  I added additional check by using native parquet reader.

PingLiuPing:
@majetideepak  Just realise you are the code owners of these files. Wonder could you please help take a look, thank you.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,115 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
   writeToFile(makeRowVector({wrappedVector}));
 };
 
+class ParquetWriterFieldIdTest : public ParquetWriterTest,
```
JkSelf: Would it be possible to directly use `ParquetWriterTest`? It seems we may not need to create a new test like `ParquetWriterFieldIdTest`.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,115 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+  auto readFile = std::make_shared<InMemoryReadFile>(sinkData);
+  auto input = std::make_unique<BufferedInput>(readFile, *leafPool_.get());
+  dwio::common::ReaderOptions readerOptions{leafPool_.get()};
+  auto veloxReader =
+      std::make_unique<ParquetReader>(std::move(input), readerOptions);
+  EXPECT_EQ(veloxReader->numberOfRows(), 1);
+  auto veloxRowType = veloxReader->rowType();
+  EXPECT_EQ(*veloxRowType, *schema);
+
+  auto raFile = std::make_shared<::arrow::io::BufferReader>(
```
JkSelf: nit: `raFile `-> `rawFile`?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,115 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+          reinterpret_cast<const uint8_t*>(sinkData.data()), sinkData.size()));
+
+  auto fileReader = parquet::arrow::ParquetFileReader::Open(raFile);
+  auto metadata = fileReader->metadata();
+  auto* descr = metadata->schema();
+  auto* root = descr->group_node();
+
+  ASSERT_EQ(root->field_count(), 4);
+
+  auto exp = [&](int32_t yes) { return GetParam() ? yes : -1; };
```
JkSelf: `yes `-> `expectedFieldId`

velox/dwio/parquet/tests/writer/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ target_link_libraries(
   velox_dwio_parquet_writer
   velox_dwio_parquet_reader
   velox_dwio_common_test_utils
+  velox_dwio_arrow_parquet_writer_test_lib
```
yingsu00: nit: reorder in alphabetic order

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,115 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+
+  // Top-level field IDs.
+  EXPECT_EQ(root->field(0)->field_id(), exp(10));
+  EXPECT_EQ(root->field(1)->field_id(), exp(20));
+  EXPECT_EQ(root->field(2)->field_id(), exp(30));
+  EXPECT_EQ(root->field(3)->field_id(), exp(40));
+
+  using GroupNode = parquet::arrow::schema::GroupNode;
+  auto* b = static_cast<const GroupNode*>(root->field(1).get());
```
JkSelf: nit: could we rename the parameters b, c, and m to more meaningful names?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,70 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
   }
 }
 
-std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
+std::shared_ptr<::arrow::Field> updateFieldRecursive(
```
JkSelf: `updateFieldRecursive `-> `updateFieldNameAndIdRecursive`

JkSelf (commented):
@PingLiuPing Thanks for the fix. Left some comments.

velox/dwio/parquet/writer/arrow/ArrowSchema.cpp
```diff
@@ -64,6 +64,8 @@ using ParquetType = Type;
 
 namespace {
 
+static constexpr char FIELD_ID_KEY[] = "PARQUET:field_id";
```
yingsu00: Why move it here?

velox/dwio/parquet/tests/writer/CMakeLists.txt
PingLiuPing: Thanks.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: @JkSelf  thanks for your comment. Yes, initially I write two test cases to cover with and without fieldId. But most of the code are duplicated so I decide to use parameterized test.
I revert this to two separate cases.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks. changed the name to `arrowBufferReader`

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks. This names corresponds to the column names actually. So, when the variable name is b we know we are referencing column `b`. Does this sound reasonable?

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks

velox/dwio/parquet/writer/arrow/ArrowSchema.cpp
PingLiuPing: Thanks, reverted.

PingLiuPing:
@JkSelf  I addressed your comments, could you please have another look. Thank you very much.

majetideepak:
@PingLiuPing How will Presto C++ use this functionality if field_ids are set as writer options?

PingLiuPing:
> @PingLiuPing How will Presto C++ use this functionality if field_ids are set as writer options?

@majetideepak Thanks for the comment.

For presto, it passes `IcebergColumnHandle` to Prestissimo. And there is a `ColumnIdentity` object in this struct, which include the id of the column. Here the id is the actual iceberg field ID.

```cpp
struct ColumnIdentity {
  int id = {};
  String name = {};
  TypeCategory typeCategory = {};
  List<ColumnIdentity> children = {};
};
```

And since iceberg guarantees that the field ID is unique, we can get this ID and write is as the parquet field_id.
Note that we cannot dynamically generate based on starter index and assign it to parquet field_id. From iceberg, the field_id is not continuous and it can be any unique number.

velox/dwio/parquet/writer/arrow/ArrowSchema.h
```diff
@@ -195,5 +195,7 @@ struct PARQUET_EXPORT SchemaManifest {
   }
 };
 
+std::shared_ptr<::arrow::KeyValueMetadata> FieldIdMetadata(int32_t field_id);
```
majetideepak: Let's follow Velox naming convention when adding new code.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -446,8 +471,11 @@ void Writer::write(const VectorPtr& data) {
   std::vector<std::shared_ptr<::arrow::Field>> newFields;
   auto childSize = schema_->size();
   for (auto i = 0; i < childSize; i++) {
-    newFields.push_back(updateFieldNameRecursive(
-        arrowSchema->fields()[i], *schema_->childAt(i), schema_->nameOf(i)));
+    newFields.push_back(updateFieldNameAndIdRecursive(
+        arrowSchema->fields()[i],
+        *schema_->childAt(i),
+        schema_->nameOf(i),
+        parquetFieldIds_ ? &parquetFieldIds_->at(i) : nullptr));
```
majetideepak: We should also check if the length of `parquetFieldIds_` is equal to schema size?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,70 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
 }
 
-std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
+std::shared_ptr<::arrow::Field> updateFieldNameAndIdRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const ParquetFieldId* fieldId = nullptr) {
+  std::shared_ptr<::arrow::Field> newField{nullptr};
```
majetideepak: don't need the explicit nullptr initialization. It is implicit.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,70 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
     auto childrenSize = rowType.size();
     std::vector<std::shared_ptr<::arrow::Field>> newFields;
     newFields.reserve(childrenSize);
-    for (auto i = 0; i < childrenSize; i++) {
-      newFields.push_back(updateFieldNameRecursive(
-          structType->fields()[i], *rowType.childAt(i), rowType.nameOf(i)));
+    for (auto i = 0; i < childrenSize; ++i) {
+      const auto* childSetting = fieldId ? &fieldId->children.at(i) : nullptr;
```
majetideepak, line 253: What does childSetting mean here? Do we need to check for `fieldId->children` size before access?

PingLiuPing:
@majetideepak  Thanks for the comment.
```
It is still not clear how the ColumnIdentity object is passed to the parquet Writer. Can you clarify?
Why are parquetFieldIds part of the WriterOptions?
```
Right, so in Prestissimo, when constructing the InsertTableHandle object, it requires the ColumnHandle object be to constructed first. And here we can transform the `protocol::iceberg::IcebergColumnHandle` to Velox ColumnHandle and save the field ID.
And the InsertTableHandle will be referenced in DataSink, and when appending data, we can get those field IDs.

The parquetFieldIds are included in WriterOptions because the writer itself doesn’t generate field IDs, it has no knowledge of them. Instead, these IDs need to be provided by the upstream caller. Since the current Parquet writer does not write field IDs, I propose making this an optional writer option. This way, we preserve the existing behavior while enabling Iceberg (or other use cases) to pass field IDs when needed.
Do you think this is the right approach, or would you suggest a better alternative?

velox/dwio/parquet/writer/arrow/ArrowSchema.h
PingLiuPing: Thanks. This function originally exist in anonymous space in file `velox/dwio/parquet/writer/arrow/ArrowSchema.cpp`, since I want to use it outside of this file I move it into namespace `facebook::velox::parquet::arrow::arrow`. And then adding a declaration here in the header file.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks.
Since here we are handling nested field (struct). chidlSetting means the fieldId of the struct child field.
Yes, I need to add sanity check here.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, make sense.

PingLiuPing:
@majetideepak  Thanks for your comments. I addressed them, could you please have another look?

PingLiuPing:
@mbasmanova  Wondering if you can have a look at this PR. The current Velox parquet writer does not write `field_id` to parquet metadata. But `field_id` is a mandatory field required by iceberg, see https://iceberg.apache.org/spec/#parquet

> Column IDs are required to be stored as [field IDs](http://github.com/apache/parquet-format/blob/40699d05bd24181de6b1457babbee2c16dce3803/src/main/thrift/parquet.thrift#L459) on the parquet schema.

velox/dwio/parquet/writer/arrow/ArrowSchema.h
mbasmanova: +1 to @majetideepak 's comment.

Can we rename FieldIdMetadata to fieldIdMetadata ?

velox/dwio/parquet/writer/Writer.h
```diff
@@ -88,6 +88,13 @@ class LambdaFlushPolicy : public DefaultFlushPolicy {
   std::function<bool()> lambda_;
 };
 
+// Settings for specifying Parquet field IDs during write operations.
```
mbasmanova: nit: drop "Settings for specifying"

Use /// for public API comments.

velox/dwio/parquet/writer/Writer.h
```diff
@@ -116,6 +123,8 @@ struct WriterOptions : public dwio::common::WriterOptions {
 
   std::shared_ptr<arrow::MemoryPool> arrowMemoryPool;
 
+  std::shared_ptr<std::vector<ParquetFieldId>> parquetFieldIds;
```
mbasmanova: Please, document.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,71 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
   }
 }
 
-std::shared_ptr<::arrow::Field> updateFieldNameRecursive(
+std::shared_ptr<::arrow::Field> updateFieldNameAndIdRecursive(
     const std::shared_ptr<::arrow::Field>& field,
     const Type& type,
-    const std::string& name = "") {
+    const std::string& name = "",
+    const ParquetFieldId* fieldId = nullptr) {
```
mbasmanova: Let's remove default values and pass these explicitly. This will make code easier to read.

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,71 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+        std::make_shared<::arrow::MapType>(newKeyField, newValueField));
   } else if (name != "") {
-    return field->WithName(name);
+    newField = field->WithName(name);
   } else {
-    return field;
+    newField = field;
   }
+
+  if (fieldId) {
```
mbasmanova: It is possible to restructure this code to 

1. compute newField
2. set fieldId
3. do more based on type

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,155 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+      ROW({"a", "b", "c", "m"},
+          {BIGINT(),
+           ROW({"x", "y"}, {INTEGER(), VARCHAR()}),
+           ARRAY(INTEGER()),
+           MAP(VARCHAR(), INTEGER())});
+
+  auto arrVec = makeArrayVector<int32_t>({{3}});
+  auto mapVec = makeMapVector<StringView, int32_t>({{{StringView("k"), 4}}});
+  auto data = makeRowVector({
+      makeFlatVector<int64_t>(1, [](auto) { return 1; }),
```
mbasmanova: Are we creating 1-row vector? Why? 

perhaps, use a version of makeFlatVector that takes a std::vector of values.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,155 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+      arrVec,
+      mapVec,
+  });
+
+  auto sink = std::make_unique<MemorySink>(
+      4 * 1024 * 1024,
+      dwio::common::FileSink::Options{.pool = leafPool_.get()});
+  auto* sinkPtr = sink.get();
+
+  parquet::WriterOptions writerOptions;
```
mbasmanova: Is this a copy-paste? Would it be possible to refactor?

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks. The test case is verifying the parquet data file metadata, so the number of rows doesn't matter.
I changed this to write multiple rows.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks, I revert the case to use parameterized test.

velox/dwio/parquet/writer/Writer.h
PingLiuPing: Thanks.

velox/dwio/parquet/writer/Writer.h
PingLiuPing: Thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks, refactored to this way. It is more clearer.

PingLiuPing:
@majetideepak  @mbasmanova Could you help to have another look please?

velox/dwio/parquet/writer/Writer.h
```diff
@@ -116,6 +123,12 @@ struct WriterOptions : public dwio::common::WriterOptions {
 
   std::shared_ptr<arrow::MemoryPool> arrowMemoryPool;
 
+  /// Optional field IDs to assign to columns in the Parquet schema.
```
mbasmanova, line 127: What would be the use case for not specifying field IDs?

velox/dwio/parquet/writer/Writer.h
```diff
@@ -116,6 +123,12 @@ struct WriterOptions : public dwio::common::WriterOptions {
...
+  /// If provided, the writer will use these IDs for the schema fields.
+  /// If not provided, the field_id will be -1.
+  /// The structure should match the schema hierarchy with nested children.
+  std::shared_ptr<std::vector<ParquetFieldId>> parquetFieldIds;
```
mbasmanova: Does this need to be a shared_ptr? Why std::vector doesn't work?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,65 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
     auto structType =
         std::dynamic_pointer_cast<::arrow::StructType>(newField->type());
     auto childrenSize = rowType.size();
     std::vector<std::shared_ptr<::arrow::Field>> newFields;
     newFields.reserve(childrenSize);
-    for (auto i = 0; i < childrenSize; i++) {
-      newFields.push_back(updateFieldNameRecursive(
-          structType->fields()[i], *rowType.childAt(i), rowType.nameOf(i)));
+    for (auto i = 0; i < childrenSize; ++i) {
+      VELOX_CHECK(!fieldId || i < fieldId->children.size());
```
mbasmanova: perhaps, move this check outside of the loop

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -351,6 +370,7 @@ Writer::Writer(
   setMemoryReclaimers();
   writeInt96AsTimestamp_ = options.writeInt96AsTimestamp;
   arrowMemoryPool_ = options.arrowMemoryPool;
+  parquetFieldIds_ = options.parquetFieldIds;
```
mbasmanova: Should we check that parquetFieldIds are aligned with schema?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -445,9 +465,13 @@ void Writer::write(const VectorPtr& data) {
       "facebook::velox::parquet::Writer::write", arrowSchema.get());
   std::vector<std::shared_ptr<::arrow::Field>> newFields;
   auto childSize = schema_->size();
+  VELOX_CHECK(!parquetFieldIds_ || childSize == parquetFieldIds_->size());
```
mbasmanova: for better error messages

```
if (parquetFieldIds_) {
   VELOX_CHECK_EQ(childSize, parquetFieldIds_->size());
}
```

velox/dwio/parquet/writer/Writer.h
```diff
@@ -88,6 +88,13 @@ class LambdaFlushPolicy : public DefaultFlushPolicy {
   std::function<bool()> lambda_;
 };
 
+/// Parquet field IDs during write operations.
+/// Used to explicitly control field ID assignment in the Parquet schema.
```
mbasmanova, line 93: Would you clarify what are the constraints on these values? I assume filed IDs must be unique. Do they need to be positive or sequential?

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -193,47 +194,65 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
-    return field->WithName(name);
-  } else {
-    return field;
-  }
+    const auto* keySetting = fieldId ? &fieldId->children.at(0) : nullptr;
+    const auto* valueSetting = fieldId ? &fieldId->children.at(1) : nullptr;
+    auto newKeyField = updateFieldNameAndIdRecursive(
+        arrowMapType->key_field(),
+        *mapType.keyType(),
+        mapType.nameOf(0),
```
mbasmanova: Before this change map keys, values and array elements didn't use 'name'. Now they do. Is this intentional? What's the effect of this change?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
   writeToFile(makeRowVector({wrappedVector}));
 };
 
+class ParquetWriterFieldIdTest : public ParquetWriterTest,
```
mbasmanova: Let's move this into a separate file. Ideally, we'd like to have one test class per file.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+                                 public ::testing::WithParamInterface<bool> {
+ protected:
+  static void SetUpTestCase() {}
```
mbasmanova: is this needed? why?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+};
+
+TEST_P(ParquetWriterFieldIdTest, fieldIds) {
+  // Schema with various complex types to test field ID assignment:
+  // - "p": BIGINT (primitive type).
+  // - "s": ROW (structure) with nested fields "x" (INTEGER) and "y" (VARCHAR).
+  // - "a": ARRAY of INTEGER.
+  // - "m": MAP from VARCHAR keys to INTEGER values.
```
mbasmanova: redundant comment; repeats the code

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+           ARRAY(INTEGER()),
+           MAP(VARCHAR(), INTEGER())});
+  constexpr int32_t kRows = 10;
+  auto data = makeRowVector({
+      makeFlatVector<int64_t>(kRows, [](auto row) { return row; }),
+      makeRowVector(
+          {makeFlatVector<int32_t>(kRows, [](auto row) { return row; }),
+           makeFlatVector<std::string>(kRows, [](auto) { return "z"; })}),
+      makeArrayVector<int32_t>(std::vector<std::vector<int32_t>>(kRows, {3})),
+      makeMapVector<std::string, int32_t>(
```
mbasmanova: use makeArrayVectorFromJson and makeMapVectorFromJson for readability

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+          std::vector<
+              std::vector<std::pair<std::string, std::optional<int32_t>>>>(
+              kRows, {{{"k", 4}}})),
+  });
+
+  auto sink = std::make_unique<MemorySink>(
+      4 * 1024 * 1024,
+      dwio::common::FileSink::Options{.pool = leafPool_.get()});
```
mbasmanova: this code repeats many times; extract a helper method

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+        ParquetFieldId{40, {ParquetFieldId{41, {}}, ParquetFieldId{42, {}}}},
+    };
+    writerOptions.parquetFieldIds =
+        std::make_shared<std::vector<ParquetFieldId>>(fieldIds);
+  }
+
+  auto writer = std::make_unique<parquet::Writer>(
+      std::move(sink), writerOptions, rootPool_, schema);
+  writer->write(data);
+  writer->close();
```
mbasmanova: this code repeats a lot; consider extracting a helper method

```
auto* sink = write(schema, data, writerOptions);
```

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -951,6 +955,122 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
...
+
+  std::string_view sinkData(sinkPtr->data(), sinkPtr->size());
```
mbasmanova: ditto; a lot of code repetition; please, refactor; it may be necessary to create a separate PR for the refactoring and land this before this one

velox/dwio/parquet/writer/Writer.h
PingLiuPing: This depends on parquet reader. For Hive, it uses column name for schema evolution and field IDs are not required.
For Iceberg and delta they use field IDs for schema evolution.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thank you, makes sense. `validateSchemaRecursive` can be enhanced to do this.

velox/dwio/parquet/writer/Writer.h
mbasmanova: Got it. Since writer doesn't control the reader, I assume it should produce files that can be read by ALL compliant readers, no?

mbasmanova: > For Iceberg and delta they use field IDs for schema evolution.

Would you update PR description to explain the impact of this bug? I assume there are cases when Iceberg and Delta connectors cannot read files produced by the writer.

PingLiuPing: Yes,  for example in this schema evolution case, current parquet reader cannot handle it. 

original column [C1, C2], write data file F1,  then delete C2, and then adding C2 back but with incompatible data types with original C2, and write file F2. When reading data from F2, match by name or position both cannot handle this.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: Thank you. This is a potential bug. After adding `name`, for map the column name of mapType.nameOf(0) is `key` and nameOf(1) is `value`. We might write additional or different metadata than default to parquet files. 
I revert the changes and keep the default value for `name`.

velox/dwio/parquet/writer/Writer.h
PingLiuPing: Thanks, changed to std::vector

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: When I run `velox_parquet_writer_test` from command line or from IDE, without this line, test case `TEST_P(ParquetWriterFieldIdTest, fieldIds)` will be ignored.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Thanks. Moved to new file.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: But when move this class to a separate file, this empty method can be delelted.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
PingLiuPing: Sure, I will submit a separate PR to refactor testcases in this file and the case I added in this PR as well.

mbasmanova: Let's first refactor existing tests, then add new test without repeating copy-paste.

PingLiuPing: Oh, got it. Misunderstood your first statement.

PingLiuPing: @mbasmanova  I submit this PR for refactoring. https://github.com/facebookincubator/velox/pull/15678

velox/dwio/parquet/writer/Writer.cpp
```diff
@@ -185,55 +188,100 @@ void validateSchemaRecursive(const RowTypePtr& schema) {
...
+    for (auto i = 0; i < childrenSize; ++i) {
+      const auto* childSetting = fieldId ? &fieldId->children.at(i) : nullptr;
+      newFields.push_back(updateFieldNameAndIdRecursive(
+          structType->fields()[i],
+          *rowType.childAt(i),
+          childSetting,
+          rowType.nameOf(i)));
     }
-    return newField->WithType(::arrow::struct_(newFields));
+    newField = newField->WithType(::arrow::struct_(newFields));
```
majetideepak, line 260: nit: why not `return` here similar to the old code?

PingLiuPing: Thanks. a single `return` at the function exit can handle all cases.

majetideepak: How does this help? You now have an extra line of code and hopefully the compiler optimizes out the extra update to `newField` and makes it similar to the old code.

PingLiuPing: @majetideepak  Thanks.
Cannot return at line 240, so we need to return at the function exit anyway.
Without adding return at the function exit there is compiler warning:
`warning: non-void function does not return a value in all control paths [-Wreturn-type]`

velox/dwio/parquet/writer/Writer.h
PingLiuPing: Thanks, clarified with "Each ID must be unique positive number, do not need to be sequential."

mbasmanova:
@PingLiuPing @majetideepak Curious, what's the status of this PR?

velox/dwio/parquet/writer/Writer.cpp
PingLiuPing: @majetideepak Does this sounds good to you? Open to any feedback. Thank you.

PingLiuPing:
> @PingLiuPing @majetideepak Curious, what's the status of this PR?

@mbasmanova  Thanks, after https://github.com/facebookincubator/velox/pull/15678 been merged, I rebased this PR. And make some adjustment to test to reuse some helper functions. Now it is pending on this comment https://github.com/facebookincubator/velox/pull/14741#discussion_r2586003015.

PR #14743, opened 2025-09-05
Use std::span instead of folly::Range

MBkkt:


mbasmanova (approved):
Thanks.

Yuhta:
There are still build failures

PR #14748, opened 2025-09-05
feat: Support VeloxQueryRunner in Expression Fuzzer

peterenescu:
Summary: Add expression fuzzer support for a new reference query runner: VeloxQueryRunner.

VeloxQueryRunner sends and receives serialized plans and serialized row vector output from LocalRunnerService, allowing fuzzer to compare expression evaluation across builds.

Differential Revision: D81618822

### DIFF
--- a/velox/expression/tests/ExpressionVerifier.cpp
+++ b/velox/expression/tests/ExpressionVerifier.cpp
@@ -434,6 +434,7 @@ ExpressionVerifier::verify(
             // Throws in case only one evaluation path throws exception.
             // Otherwise, return false to signal that the expression failed.
             if (exceptionCommonPtr && exceptionReference) {
+              LOG(INFO) << "Both paths threw exception.";
               verificationStates.push_back(VerificationState::kBothPathsThrow);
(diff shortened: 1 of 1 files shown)


velox/exec/fuzzer/CMakeLists.txt
```diff
@@ -27,6 +27,12 @@ add_library(
   PrestoSql.cpp
 )
 
+# Add VeloxQueryRunner only when remote functions are enabled
+if(VELOX_ENABLE_REMOTE_FUNCTIONS)
+  target_sources(velox_fuzzer_util PRIVATE VeloxQueryRunner.cpp)
+  target_link_libraries(velox_fuzzer_util FBThrift::thriftcpp2)
+endif()
+
```
kagamiori: @peterenescu What's our plan to resolve this? Can we find out what're missing for building VeloxQueryRunner in Github CI and find an approach to bridge the gap? We would want to run fuzzer with VeloxQueryRunner in Github CI eventually too.

Issue #14754, opened 2025-09-06
ArrowArrayStruct contains null data pointer for a buffer with non-zero computed size


wecharyu:
### Bug description

Writing parquet file could get this error sometimes:
```bash
25/08/27 07:03:36 INFO [Executor task launch worker for task 0.0 in stage 3.0 (TID 9767)] ShuffleBlockFetcherIterator: Started 49 remote fetches in 3143 ms
E20250827 07:04:16.583768 3615260 Exceptions.h:66] Line: /home/jenkins/agent/workspace/di-spark/Gluten/CI_spark_gluten/ep/build-velox/build/velox_ep/velox/exec/Driver.cpp:629, Function:operator(), Expression:  Operator::addInput failed for [operator: TableWrite, plan node ID: 3]: Invalid: ArrowArrayStruct contains null data pointer for a buffer with non-zero computed size, Source: RUNTIME, ErrorCode: INVALID_STATE
25/08/27 07:04:17 ERROR [Executor task launch worker for task 0.0 in stage 3.0 (TID 9767)] Utils: Aborting task
org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Operator::addInput failed for [operator: TableWrite, plan node ID: 3]: Invalid: ArrowArrayStruct contains null data pointer for a buffer with non-zero computed size
Retriable: False
Function: operator()
File: /home/jenkins/agent/workspace/di-spark/Gluten/CI_spark_gluten/ep/build-velox/build/velox_ep/velox/exec/Driver.cpp
Line: 629
```

### System information

```bash
CMake Version: 3.31.6
System: Linux-5.4.139-1.el7.elrepo.x86_64
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.5.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.5.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt
```

### Relevant logs

```bash

```

PR #14755, opened 2025-09-06
fix: ArrowArrayStruct contains null data pointer for a buffer with non-zero computed size

wecharyu:
fix https://github.com/facebookincubator/velox/issues/14754.


wecharyu:
@majetideepak @pedroerp Could you help review this patch?

majetideepak (approved):

majetideepak:
@wecharyu can you rebase with latest main? Thanks!

PR #14757, opened 2025-09-06
build(pyvelox): Exclude pyvelox_libraries component from default

### LINKED ISSUE #14756: Don't install pyvelox component by default
### Description

It doesn't really make sense to build the contentse of `velox/python` when not invoked through the python package build as that's their only use and it adds a pybind11 (at minimum) dependency to the minimal build.

assignUser:
Fixes #14756


assignUser:
Superseded by #14738

PR #14762, opened 2025-09-07
build(ci): Fix build report

assignUser:
I thought there was an issue with the report code breaking the build but it was actually our conbench instance. It seems to have issues with fetching Github metadata (I have already started investigating).

This is a workaround but should not cause any issues as we don't collect metrics in PRs.

### DIFF
--- a/.github/workflows/build-metrics.yml
+++ b/.github/workflows/build-metrics.yml
@@ -36,7 +36,7 @@ permissions:
 jobs:
   metrics:
     name: Linux ${{ matrix.link-type }} - ${{ matrix.type }} with adapters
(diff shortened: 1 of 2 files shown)


.github/workflows/build-metrics.yml
```diff
@@ -36,7 +36,7 @@ permissions:
 jobs:
   metrics:
     name: Linux ${{ matrix.link-type }} - ${{ matrix.type }} with adapters
-    if: ${{ github.repository == 'facebookincubator/velox' }}
+    if: false # ${{ github.repository == 'facebookincubator/velox' }}
```
assignUser: I will revert this before the merge with `[no ci]` to prevent a run.

PR #14765, opened 2025-09-08
feat(SelectivityIterator): Enable copy constructor for SelectivityIterator

akoshelev:
Summary:
I am trying to write a UDF function that will operate on std::range of input values to avoid copying them to a new container.

It works if I make`SelectivityIterator` copyable as ranges require that. I still need to build a wrapper around it to add iterator tags and iterator concepts, but it is not too bad as long as iterator logic is encapsulated in `SelectivityIterator`.

Differential Revision: D81847304

### DIFF
--- a/velox/vector/tests/SelectivityVectorTest.cpp
+++ b/velox/vector/tests/SelectivityVectorTest.cpp
@@ -360,6 +360,47 @@ TEST(SelectivityVectorTest, iterator) {
   EXPECT_EQ(count, bits::countBits(&contiguous[0], 0, 240));
 }
 
+TEST(SelectivityVectorTest, selectivityIteratorCopyConstructor) {
+  SelectivityVector vector(10);
+  vector.clearAll();
+  vector.setValid(1, true);
+  vector.setValid(3, true);
+  vector.setValid(5, true);
+  vector.setValid(7, true);
+  vector.updateBounds();
+
+  // Test non-advanced iterator
+  SelectivityIterator iter1(vector);
+  SelectivityIterator iter1Copy(iter1);
+
+  auto collectValues = [](SelectivityIterator& iter) {
+    std::vector<vector_size_t> values;
+    vector_size_t row;
+    while (iter.next(row)) {
+      values.push_back(row);
+    }
+    return values;
+  };
+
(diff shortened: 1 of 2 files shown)


velox/vector/tests/SelectivityVectorTest.cpp
```diff
@@ -360,6 +360,47 @@ TEST(SelectivityVectorTest, iterator) {
...
+  auto collectValues = [](SelectivityIterator& iter) {
+    std::vector<vector_size_t> values;
+    vector_size_t row;
+    while (iter.next(row)) {
+      values.push_back(row);
+    }
+    return values;
+  };
+
+  EXPECT_EQ(collectValues(iter1), std::vector<vector_size_t>({1, 3, 5, 7}));
```
pedroerp, line 385: could we just compare the iterators directly, if you provide the equality operator somewhere?

velox/vector/tests/SelectivityVectorTest.cpp
```diff
@@ -360,6 +360,47 @@ TEST(SelectivityVectorTest, iterator) {
...
+  iter2.next(row); // consumes 1
+  iter2.next(row); // consumes 3
+
+  SelectivityIterator iter2copy(iter2);
+
+  EXPECT_EQ(collectValues(iter2), std::vector<vector_size_t>({5, 7}));
+  EXPECT_EQ(collectValues(iter2copy), std::vector<vector_size_t>({5, 7}));
+
+  SelectivityIterator iter3 = iter2copy;
+  EXPECT_EQ(collectValues(iter3), std::vector<vector_size_t>());
```
pedroerp, line 401: I was expecting a test case using std::range to ensure it works now?

akoshelev: std::range concept came with C++20 - my impression was that Velox uses C++17 standard.

velox/vector/tests/SelectivityVectorTest.cpp
akoshelev: I am not sure I know how to do that. I thought comparing iterators directly would result in comparing the element they point to which is insufficient here. 
I could be wrong though - do you have a suggestion how to do that?

PR #14773, opened 2025-09-08
fix: Disable cpr CURL_ZLIB build in CACHE

liujiayi771:
CURL_ZLIB needs to be configured in the CACHE in order to properly disable its compilation. Otherwise, it will not be effectively turned off, because `NOT DEFINED CACHE{CURL_ZLIB}` will be true.

https://github.com/libcpr/cpr/blob/7fa467321f41c883ab83cc12d7f51c58653a0522/CMakeLists.txt#L213


liujiayi771:
@kgpai Could you help to review?

PR #14776, opened 2025-09-08
fix: Add `-Wno-restrict` for gcc-12.3.1 as well

zhouyuan:
Gcc-12 has a bug on (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329). Velox already adds the flag for building with 12.2.1. 
In Gluten community there is a requirement for Gcc-12.3.1 to run with openEuler. This patch adds the same flag for Gcc-12.3.1 as well


CMakeLists.txt
```diff
@@ -429,9 +429,10 @@ if(ENABLE_ALL_WARNINGS)
          -Wno-format-overflow \
          -Wno-strict-aliasing"
     )
-    # Avoid compiler bug for GCC 12.2.1
+    # Avoid compiler bug for GCC 12
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329
-    if(CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "12.2.1")
+    list(APPEND GCC_LIST "12.2.1" "12.3.1")
```
czentgr: There is GCC 12.4 and 12.5. Would it not be better to use `VERSION_GREATER_EQUAL 12.2`?

zhouyuan: Update.

zhouyuan:
@czentgr Thanks. I have rebased to the latest main branch. Please help to take a look.

CMakeLists.txt
```diff
@@ -438,9 +438,9 @@ if(ENABLE_ALL_WARNINGS)
          -Wno-format-overflow \
          -Wno-strict-aliasing"
     )
-    # Avoid compiler bug for GCC 12.2.1
+    # Avoid compiler bug for GCC 12
```
czentgr: Nit: technically "GCC 12.2 and later".

zhouyuan: done

CMakeLists.txt
```diff
@@ -438,9 +438,9 @@ if(ENABLE_ALL_WARNINGS)
          -Wno-format-overflow \
          -Wno-strict-aliasing"
     )
-    # Avoid compiler bug for GCC 12.2.1
+    # Avoid compiler bug for GCC 12.2 and later
```
czentgr, line 441: Because we are discussing moving the minimum GCC to GCC13 or 14 this would be addressed then. So lets make sure to set an upper bound for adding this warning as well.
We can also test it because the fedora CI uses GCC14.

```
if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "12.2" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "13")
```

zhouyuan: done

zhouyuan:
Not necessary as Velox will upgrade to gcc-14

PR #14784, opened 2025-09-08
refactor: Remove redundant calls to registerConnectorFactory

mbasmanova:
Summary:
Connector factories are used only by Prestissimo to create multiple instances of the same kind of connector for different catalogs. These factories are not needed for other use cases.

A follow-up would be to move connector factories out of Velox into Prestissimo.

Differential Revision: D81960874

### DIFF
--- a/velox/benchmarks/QueryBenchmarkBase.cpp
+++ b/velox/benchmarks/QueryBenchmarkBase.cpp
@@ -215,12 +215,9 @@ void QueryBenchmarkBase::initialize() {
       std::move(configurationValues));
 
   // Create hive connector with config...
-  connector::registerConnectorFactory(
-      std::make_shared<connector::hive::HiveConnectorFactory>());
+  connector::hive::HiveConnectorFactory factory;
   auto hiveConnector =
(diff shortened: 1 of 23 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

yingsu00:
@mbasmanova What about Gluten if we move the ConnectorFactories into Prestissimo? On the other hand, we want to promote Iceberg as a standalone connector, but the tests directly relying on Hive are obstacles to that goal. Please also note that VELOX_ENABLE_HIVE_CONNECTOR is a CMake option and could be set to false by a user in his/her environment. These tests directly depending on Hive would cause the build to fail. Gluten is an important user of Iceberg too. That's why I proposed https://github.com/facebookincubator/velox/pull/14687 to make these tests connector-agnostic. Would it be possible for you to read that PR to see if it makes sense? THanks.

mbasmanova:
@yingsu00 I expect that Gluten as well as Prestissimo can maintain their own connector factories. Velox doesn't use these, hence, there is no good reason for these to live inside the library.

Issue #14785, opened 2025-09-08
Change TIME in tpcds Tables.h


kgpai:
### Problem description

This line in [Tables.h](https://github.com/facebookincubator/velox/blob/main/velox/tpcds/gen/dsdgen/include/tables.h#L51)

```
#define TIME 18
```

Makes it impossible to add the TIME type in Velox.
Can it be changed to something like so :

```
#define TIMET 18
```
that we can add TIME as a type in Velox .

### System information

N/A

### CMake log

```bash
N/A
```

kgpai:
cc: @pramodsatya @aditi-pandit @majetideepak @czentgr @mbasmanova

pedroerp:
We should vendor all global symbols, I had to do something similar to TPC-H. Just rename it to something like `TPC_DS_TIME`

pramodsatya:
> We should vendor all global symbols, I had to do something similar to TPC-H. Just rename it to something like `TPC_DS_TIME`

Thanks @pedroerp, will add `TPC_DS` prefix to the dsdgen macros in a follow-up.

PR #14786, opened 2025-09-08
fix: Rename Tpcds table TIME macro

pramodsatya:
Resolves https://github.com/facebookincubator/velox/issues/14785.


velox/tpcds/gen/dsdgen/include/tables.h
```diff
@@ -48,7 +48,7 @@
 #define STORE 15
 #define STORE_RETURNS 16
 #define STORE_SALES 17
-#define TIME 18
+#define TIME_T 18
```
kgpai: nit: changed DATE to DATET, TIME to TIME_T - Should you change it to TIMET or DATE to DATE_T to be consistent?

pramodsatya: Changed to `TIMET`, was trying to follow _ convention from remaining macros and missed DATET, thanks for pointing it out.

Yuhta:
Hi @pramodsatya , the code in tpcds folder does not have full path for the `#include` preprocessors so it fails to build in Meta internal build.  Can you fix all the `#include` lines with the full path from the root of the project?  Thanks

pramodsatya:
> Hi @pramodsatya , the code in tpcds folder does not have full path for the `#include` preprocessors so it fails to build in Meta internal build. Can you fix all the `#include` lines with the full path from the root of the project? Thanks

Updated @Yuhta, could you please take another look?

pramodsatya:
Thanks for the feedback @mbasmanova, will ensure there are separate PRs for separate commits going forward. Looks like this is merged and will rename the macros with TPC_DS prefix in a follow-up.

PR #14796, opened 2025-09-09
feat: Update Expression.cc using Casts util function for type cast

juwentus1234:
Summary: Using utility casts.h function

Differential Revision: D81986594

### DIFF
--- a/velox/core/Expressions.cpp
+++ b/velox/core/Expressions.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "velox/core/Expressions.h"
+#include "velox/common/Casts.h"
 #include "velox/common/encode/Base64.h"
 #include "velox/vector/ComplexVector.h"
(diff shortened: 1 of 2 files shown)


velox/core/Expressions.cpp
```diff
@@ -465,7 +466,7 @@ TypedExprPtr FieldAccessTypedExpr::rewriteInputNames(
   VELOX_CHECK_EQ(1, newInputs.size());
   // Only rewrite name if input in InputTypedExpr. Rewrite in other
   // cases(like dereference) is unsound.
-  if (!std::dynamic_pointer_cast<const InputTypedExpr>(newInputs[0])) {
+  if (!is_instance_of<InputTypedExpr>(newInputs[0])) {
```
mbasmanova, line 469: @juwentus1234 @xiaoxmeng 

As discussed earlier, we should not use cast here. Instead, use `newInputs[0]->isInputKind()`. Ditto other places.

PR #14798, opened 2025-09-09
docs: Fix install prefix to use $USER instead of $USERNAME for better compatibility

PHILO-HE:


philo-he:
@majetideepak, could you take a look?

Issue #14800, opened 2025-09-09
Spark get_json_object returns non-compact JSON for objects/arrays


wecharyu:
### Bug description

Spark does not enable pretty printing for `get_json_object`, so it would output a compact json string:
https://github.com/apache/spark/blob/fa33ea000a0bda9e5a3fa1af98e8e85b8cc5e4d4/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/json/JsonExpressionEvalUtils.scala#L398-L403

But velox current implement will output a non-compact result:
```bash
= Results ==
!== Correct Answer - 1 ==   == Gluten Answer - 1 ==
 struct<>                   struct<>
![["v1","v2"]]              [["v1", "v2"]]
```

### System information

CMake Version: 3.31.6
System: Linux-5.4.139-1.el7.elrepo.x86_64
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.5.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.5.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

PR #14801, opened 2025-09-09
fix: Minify JSON objects/arrays in Spark get_json_object

wecharyu:
fix: https://github.com/facebookincubator/velox/issues/14800


rui-mo:
@PHILO-HE Would you like to review this change? Thanks.

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -123,31 +131,41 @@ struct GetJsonObjectFunction {
...
   // - Removes spaces after dots (e.g., "$. a" -> "$.a")
+  // - Removes trailing spaces after root symbol (e.g., "$ " -> "$")
   std::string normalizeJsonPath(StringView jsonPath) {
     // First, remove single quotes for bracket notation
     const std::string& path = removeSingleQuotes(jsonPath);
     if (path == "-1") {
       return path;
     }
 
+    std::string normalized = path;
```
rui-mo: Can we reuse the `path`?

wecharyu: Addressed.

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -123,31 +131,39 @@ struct GetJsonObjectFunction {
...
     // Use Boost regex to find and remove spaces after dots
     // Pattern: "dot + one or more spaces" -> "dot"
     static const boost::regex dotSpaceRegex("\\.\\s+");
-    return boost::regex_replace(path, dotSpaceRegex, ".");
+    path = boost::regex_replace(path, dotSpaceRegex, ".");
+
+    // Remove trailing spaces after $.
+    // Pattern: "$ + one or more spaces" -> "$"
+    static const boost::regex rootSpaceRegex("\\$\\s+$");
+    path = boost::regex_replace(path, rootSpaceRegex, "$");
```
philo-he: Could we combine this with the above to remove trailing space after '$' or '.' in single regex replacement?

velox/functions/sparksql/tests/GetJsonObjectTest.cpp
```diff
@@ -82,19 +84,19 @@ TEST_F(GetJsonObjectTest, basic) {
       getJsonObject(
           R"({"my": {"info": {"name": "Alice", "age": "5", "id": "001"}}})",
           "$.my.info"),
-      R"({"name": "Alice", "age": "5", "id": "001"})");
+      "{\"name\":\"Alice\",\"age\":\"5\",\"id\":\"001\"}");
```
philo-he: Why not keep using `R"(...)"` for raw string literals.

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -185,21 +201,59 @@ struct GetJsonObjectFunction {
...
+          }
+          result.append("\"");
+          result.append(key);
+          result.append("\":");
+          if (!extractStringResult(field.value(), result, false)) {
+            return false;
+          }
+          first = false;
+        }
+        result.append("}");
```
philo-he, line 303: It seems JSON specification doesn't require the extra space after comma must be removed. Even though Spark does this, I am wondering whether the current implementation in Velox is acceptable to Spark users.

velox/functions/sparksql/GetJsonObject.h
wecharyu: They are a little different, the `$` with space can not has other char but `.` could. For example `$.  a` is valid, but `$ .a` is invalid, so it seems not necessary to combine them.

velox/functions/sparksql/GetJsonObject.h
wecharyu: Some users have there own compare tool to check the result consistency between spark and gluten, this difference will cause some misleading in consistent checking.

velox/functions/sparksql/tests/GetJsonObjectTest.cpp
wecharyu: Revert.

velox/functions/sparksql/GetJsonObject.h
philo-he: @wecharyu, thanks for the reply. I understand many users view the result inconsistency as bugs. However, to achieve full compatibility with Spark, in many cases, including the case here, we may have to complicate code, even bring perf. overhead possibly. My point is, if the difference is acceptable, maybe user can just change the check tool to allow such difference.

wecharyu: The users always check the result consistency according to the output type, it's difficult to figure out where the result comes from and ignore it for `get_json_object` expression. 

I have run a benchmark test on this change, it improves performance for small outputs, with larger outputs incurring proportional overhead — a trade-off we make for consistency. @PHILO-HE Is such performance decline acceptable?
- Before
```bash
[...]benchmarks/GetJsonObjectBenchmark.cpp     relative  time/iter   iters/s
number_depth1                                             412.47ms      2.42
number_depth1_variable_path                               741.39ms      1.35
number_depth2                                             465.45ms      2.15
object_depth1_1col                                        461.17ms      2.17
object_depth3_4col                                        716.10ms      1.40
array_depth1_1element                                     459.90ms      2.17
array_depth4_20element                                    904.13ms      1.11
```
- After
```bash
[...]benchmarks/GetJsonObjectBenchmark.cpp     relative  time/iter   iters/s
number_depth1                                             239.33ms      4.18 (+42%)
number_depth1_variable_path                               267.14ms      3.74 (+64%)
number_depth2                                             288.82ms      3.46 (+38%)
object_depth1_1col                                        301.56ms      3.32 (+34%)
object_depth3_4col                                        622.49ms      1.61 (+13%)
array_depth1_1element                                     267.38ms      3.74 (+41%)
array_depth4_20element                                       1.22s   816.49m (-35%)
```

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -120,34 +127,90 @@ struct GetJsonObjectFunction {
...
+
+    if (state == kAfterDot) {
+      // Trailing dot is invalid.
+      return "-1";
     }
 
-    // Use Boost regex to find and remove spaces after dots
-    // Pattern: "dot + one or more spaces" -> "dot"
-    static const boost::regex dotSpaceRegex("\\.\\s+");
-    return boost::regex_replace(path, dotSpaceRegex, ".");
```
wecharyu, line 136: `regex_replace` is expensive especially when the `jsonPath` needs to be normalized for each row, so we replaced it with a state-machine implementation.

velox/functions/sparksql/GetJsonObject.h
philo-he: @wecharyu, thanks for conducting the benchmark tests. The overhead seems non-negligible. cc @mbasmanova if you have any comments.

mbasmanova:
@wecharyu 

> Supports trailing spaces in $ path, eg: '$ '

This change seems unrelated the format of the output. Would you extract it into a separate PR? Also, please, update the documentation: https://facebookincubator.github.io/velox/functions/spark/json.html#get_json_object

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -52,26 +51,38 @@ struct GetJsonObjectFunction {
...
+      return false;
+    }
     try {
+      // jsonPath is "$".
+      if (formattedJsonPath.empty()) {
+        if (jsonDoc.is_scalar()) {
+          // For scalar value, need to wrap it with a json object to make
+          // simdjson parse it correctly.
+          auto wrappedJson =
+              "{\"k\":" + std::string(json.data(), json.size()) + "}";
```
mbasmanova, line 71: There must be simpler way that doesn't require constructing extra string. Consider creating an issue in simdjson repo to ask for advice.

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -120,34 +127,90 @@ struct GetJsonObjectFunction {
     return result;
   }
 
-  // Normalizes the JSON path to be Spark-compatible:
-  // - Removes single quotes in bracket notation
-  // - Removes spaces after dots (e.g., "$. a" -> "$.a")
+  // Normalizes the JSON path to be Spark-compatible.
```
mbasmanova, line 130: Let's update the documentation to clarify the format of the output: https://facebookincubator.github.io/velox/functions/spark/json.html#get_json_object

In general, I'm not sure it is a good idea to try to restrict the format beyond it being a valid JSON, especially if doing so affects performance.

Users should be advised not to take a dependency on specific string representation of the JSON object.

Correctness testing is a challenge indeed. However, it is not limited to JSON strings. Any string representation that is not 100% defined is a problem. Specifically, the results of casting floating point values to string.

Prestissimo folks have developed a set of techniques to use for correctness verification. Perhaps, they can share some materials and advice. 

CC: @kagamiori  @amitkdutta @kgpai 

If we do need to provide a very specific format of the JSON output, then I suggest to introduce a configuration flag to enable / disable that functionality and document the format clearly.

Hope this helps.

philo-he: I agree with simply updating the correction verification tool to allow differences caused by whitespace.

FYI. According to https://www.rfc-editor.org/rfc/rfc8259, "Insignificant whitespace is allowed before or after any of the six structural characters."

PR #14804, opened 2025-09-09
feat: Add merge_hll Presto scalar function

natashasehgal:
Summary: Add merge_hll native function

Use Vector Function as Simple Function currently does not have functionality to provide memory allocator
Here we use query context memory pool, available from the vector function instead of creating a HashStringAllocator

Differential Revision: D82038217

### DIFF
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -129,6 +129,8 @@ std::unordered_set<std::string> skipFunctions = {
     "merge_sfm", // Fuzzer can generate sketches of different sizes.
     "element_at",
     "width_bucket",
+    // Varbinary output can't be compared exactly with Java.
+    "merge_hll",
     // Fuzzer and the underlying engine are confused about TDigest functions
     // (since TDigest is a user defined type), and tries to pass a
     // VARBINARY (since TDigest's implementation uses an
--- a/velox/functions/prestosql/HyperLogLogFunctions.h
+++ b/velox/functions/prestosql/HyperLogLogFunctions.h
@@ -78,4 +78,131 @@ struct EmptyApproxSetWithMaxErrorFunction {
     return true;
   }
 };
+
+class MergeHllFunction : public exec::VectorFunction {
+ public:
+  static std::vector<exec::FunctionSignaturePtr> signatures() {
+    return {exec::FunctionSignatureBuilder()
+                .returnType("hyperloglog")
+                .argumentType("array(hyperloglog)")
+                .build()};
+  }
+
+  void apply(
+      const SelectivityVector& rows,
+      std::vector<VectorPtr>& args,
+      const TypePtr& outputType,
+      exec::EvalCtx& context,
+      VectorPtr& result) const override {
+    VELOX_CHECK_EQ(args.size(), 1);
+
+    context.ensureWritable(rows, outputType, result);
+    auto* flatResult = result->as<FlatVector<StringView>>();
+
+    exec::LocalDecodedVector arrayDecoder(context, *args[0], rows);
+    auto decodedArray = arrayDecoder.get();
+    auto baseArray = decodedArray->base()->as<ArrayVector>();
+    auto rawSizes = baseArray->rawSizes();
+    auto rawOffsets = baseArray->rawOffsets();
+    auto indices = decodedArray->indices();
+    auto arrayElements = baseArray->elements();
+    auto flatArrayElements = arrayElements->as<FlatVector<StringView>>();
+
+    memory::MemoryPool* memoryPool = context.pool();
+    using SparseHll = common::hll::SparseHll<memory::MemoryPool>;
+    using DenseHll = common::hll::DenseHll<memory::MemoryPool>;
+
+    context.applyToSelectedNoThrow(rows, [&](vector_size_t row) {
+      if (decodedArray->isNullAt(row)) {
+        flatResult->setNull(row, true);
+        return;
+      }
+
+      auto arraySize = rawSizes[indices[row]];
+      auto arrayOffset = rawOffsets[indices[row]];
+
+      std::optional<int8_t> indexBitLength;
+      std::optional<SparseHll> sparseHll;
+      std::optional<DenseHll> denseHll;
+
+      for (vector_size_t i = 0; i < arraySize; ++i) {
+        auto elementIndex = arrayOffset + i;
+        if (arrayElements->isNullAt(elementIndex)) {
+          continue;
+        }
+
+        auto hllData = flatArrayElements->valueAt(elementIndex);
+        if (hllData.empty()) {
+          continue;
+        }
+
+        const bool isSparseHllData =
+            common::hll::SparseHlls::canDeserialize(hllData.data());
+        const bool isDenseHllData = !isSparseHllData &&
+            common::hll::DenseHlls::canDeserialize(hllData.data());
+
+        VELOX_USER_CHECK(
+            isSparseHllData || isDenseHllData, "Invalid HLL data format");
+
+        const auto hllIndexBitLength = isSparseHllData
+            ? common::hll::SparseHlls::deserializeIndexBitLength(hllData.data())
+            : common::hll::DenseHlls::deserializeIndexBitLength(hllData.data());
+
+        if (indexBitLength.has_value()) {
+          VELOX_USER_CHECK_EQ(
+              indexBitLength.value(),
+              hllIndexBitLength,
+              "Cannot merge HLLs with different indexBitLength");
+        } else {
+          indexBitLength = hllIndexBitLength;
+        }
+
+        if (isSparseHllData) {
+          if (denseHll.has_value()) {
+            SparseHll temp{hllData.data(), memoryPool};
+            temp.toDense(*denseHll);
+          } else {
+            if (!sparseHll.has_value()) {
+              sparseHll.emplace(memoryPool);
+              sparseHll->setSoftMemoryLimit(
+                  common::hll::DenseHlls::estimateInMemorySize(
+                      indexBitLength.value()));
+            }
+
+            sparseHll->mergeWith(hllData.data());
+            if (sparseHll->overLimit()) {
+              denseHll.emplace(indexBitLength.value(), memoryPool);
+              sparseHll->toDense(*denseHll);
+              sparseHll.reset();
+            }
+          }
+        } else {
+          if (sparseHll.has_value()) {
+            denseHll.emplace(indexBitLength.value(), memoryPool);
+            sparseHll->toDense(*denseHll);
+            sparseHll.reset();
+          } else if (!denseHll.has_value()) {
+            denseHll.emplace(indexBitLength.value(), memoryPool);
+          }
+          denseHll->mergeWith(hllData.data());
+        }
+      }
+
+      if (!sparseHll.has_value() && !denseHll.has_value()) {
+        flatResult->setNull(row, true);
+      } else if (sparseHll.has_value()) {
+        auto size = sparseHll->serializedSize();
+        char* buffer = flatResult->getRawStringBufferWithSpace(size);
+        sparseHll->serialize(indexBitLength.value(), buffer);
+        flatResult->setNoCopy(row, StringView(buffer, size));
+      } else {
+        auto size = denseHll->serializedSize();
+        char* buffer = flatResult->getRawStringBufferWithSpace(size);
+        denseHll->serialize(buffer);
+        flatResult->setNoCopy(row, StringView(buffer, size));
+      }
+    });
+  }
+};
+
 } // namespace facebook::velox::functions
--- a/velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
@@ -13,14 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include "velox/common/base/tests/GTestUtils.h"
 #include "velox/common/hyperloglog/SparseHll.h"
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
 #include "velox/functions/prestosql/types/HyperLogLogType.h"
 #define XXH_INLINE_ALL
 #include <xxhash.h>
 
-using facebook::velox::common::hll::DenseHll;
-using facebook::velox::common::hll::SparseHll;
+using namespace facebook::velox::common::hll;
 
 namespace facebook::velox {
 namespace {
@@ -50,6 +50,28 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
     return serialized;
   }
 
+  // Creates sparse HLL with values from start (inclusive) to end (exclusive).
+  // modValue applies modulo operation if non-zero.
+  std::string
+  createSparse(int8_t indexBitLength, int start, int end, int modValue = 0) {
+    SparseHll<> hll{&allocator_};
+    for (int i = start; i < end; i++) {
+      hll.insertHash(hashOne(modValue ? (i % modValue) : i));
+    }
+    return serialize(indexBitLength, hll);
+  }
+
+  // Creates dense HLL with values from start (inclusive) to end (exclusive).
+  // modValue applies modulo operation if non-zero.
+  std::string
+  createDense(int8_t indexBitLength, int start, int end, int modValue = 0) {
+    DenseHll<> hll{indexBitLength, &allocator_};
+    for (int i = start; i < end; i++) {
+      hll.insertHash(hashOne(modValue ? (i % modValue) : i));
+    }
+    return serialize(hll);
+  }
+
   std::shared_ptr<memory::MemoryPool> pool_{
       memory::memoryManager()->addLeafPool()};
   HashStringAllocator allocator_{pool_.get()};
@@ -77,12 +99,7 @@ TEST_F(HyperLogLogFunctionsTest, cardinalitySparse) {
     return evaluateOnce<int64_t>("cardinality(c0)", HYPERLOGLOG(), input);
   };
 
-  SparseHll<> sparseHll{&allocator_};
-  for (int i = 0; i < 1'000; i++) {
-    sparseHll.insertHash(hashOne(i % 17));
-  }
-
-  auto serialized = serialize(11, sparseHll);
+  auto serialized = createSparse(11, 0, 1000, 17);
   EXPECT_EQ(17, cardinality(serialized));
 }
 
@@ -91,13 +108,14 @@ TEST_F(HyperLogLogFunctionsTest, cardinalityDense) {
     return evaluateOnce<int64_t>("cardinality(c0)", HYPERLOGLOG(), input);
   };
 
-  DenseHll<> denseHll{12, &allocator_};
+  DenseHll<> expectedHll{12, &allocator_};
   for (int i = 0; i < 10'000'000; i++) {
-    denseHll.insertHash(hashOne(i));
+    expectedHll.insertHash(hashOne(i));
   }
+  auto expectedCardinality = expectedHll.cardinality();
 
-  auto serialized = serialize(denseHll);
-  EXPECT_EQ(denseHll.cardinality(), cardinality(serialized));
+  auto serialized = createDense(12, 0, 10'000'000);
+  EXPECT_EQ(expectedCardinality, cardinality(serialized));
 }
 
 TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
@@ -110,6 +128,133 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
       0,
       evaluateOnce<int64_t>(
           "cardinality(empty_approx_set(0.1))", makeRowVector(ROW({}), 1)));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHll) {
+  // Test merging two HLL instances with different value ranges.
+  const int8_t indexBitLength = 12;
+
+  std::vector<std::string> serializedHlls = {
+      createSparse(indexBitLength, 0, 10),
+      createSparse(indexBitLength, 10, 20)};
+
+  auto elements = makeFlatVector(serializedHlls, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  auto cardinality = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+
+  EXPECT_EQ(20, cardinality);
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllMixedSparseAndDense) {
+  // Test merging sparse and dense HLL instances.
+  const int8_t indexBitLength = 12;
+
+  std::vector<std::string> serializedHlls = {
+      createSparse(indexBitLength, 0, 50),
+      createDense(indexBitLength, 50, 2000)};
+
+  auto elements = makeFlatVector(serializedHlls, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  auto cardinality = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+
+  EXPECT_EQ(1978, cardinality);
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllWithNull) {
+  auto serialized = createSparse(12, 0, 10);
+  auto elements = makeNullableFlatVector<std::string>(
+      {serialized, std::nullopt, serialized}, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 3}, elements);
+
+  auto cardinality = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+  EXPECT_EQ(10, cardinality);
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllEmpty) {
+  auto input = makeRowVector(
+      {makeArrayVectorFromJson<std::string>({"null"}, ARRAY(HYPERLOGLOG()))});
+  auto result = evaluate("merge_hll(c0)", input);
+  ASSERT_TRUE(result->isNullAt(0));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllNullArray) {
+  auto input = makeRowVector({makeNullableArrayVector<std::string>(
+      {std::nullopt}, ARRAY(HYPERLOGLOG()))});
+  auto result = evaluate("merge_hll(c0)", input);
+  ASSERT_TRUE(result->isNullAt(0));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllAllNulls) {
+  auto input = makeRowVector({makeArrayVectorFromJson<std::string>(
+      {"[null, null, null]"}, ARRAY(HYPERLOGLOG()))});
+  auto result = evaluate("merge_hll(c0)", input);
+  ASSERT_TRUE(result->isNullAt(0));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllInvalidData) {
+  // Test with invalid HLL data format.
+  std::string invalidData = "invalid_hll_data";
+  auto elements = makeFlatVector<std::string>({invalidData}, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 1}, elements);
+
+  VELOX_ASSERT_THROW(
+      evaluate("merge_hll(c0)", makeRowVector({arrayVector})),
+      "Invalid HLL data format");
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllMismatchedIndexBitLength) {
+  const int8_t indexBitLength1 = 11;
+  const int8_t indexBitLength2 = 12;
+
+  std::vector<std::string> serializedHlls = {
+      createSparse(indexBitLength1, 0, 5),
+      createSparse(indexBitLength2, 5, 10)};
+
+  auto elements = makeFlatVector(serializedHlls, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  VELOX_ASSERT_THROW(
+      evaluate("merge_hll(c0)", makeRowVector({arrayVector})),
+      "Cannot merge HLLs with different indexBitLength");
+}
+
+TEST_F(
+    HyperLogLogFunctionsTest,
+    mergeHllMismatchedIndexBitLengthDenseVsSparse) {
+  const int8_t indexBitLength1 = 11;
+  const int8_t indexBitLength2 = 12;
+
+  std::vector<std::string> serializedHlls = {
+      createSparse(indexBitLength1, 0, 50),
+      createDense(indexBitLength2, 50, 2000)};
+
+  auto elements = makeFlatVector(serializedHlls, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  VELOX_ASSERT_THROW(
+      evaluate("merge_hll(c0)", makeRowVector({arrayVector})),
+      "Cannot merge HLLs with different indexBitLength");
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllDenseFirstThenSparse) {
+  const int8_t indexBitLength = 12;
+
+  std::vector<std::string> serializedHlls = {
+      createDense(indexBitLength, 0, 2000),
+      createSparse(indexBitLength, 2000, 2050)};
+
+  auto elements = makeFlatVector(serializedHlls, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  auto cardinality = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+
+  EXPECT_EQ(2035, cardinality);
 }
 
 } // namespace
--- a/velox/docs/functions/presto/hyperloglog.rst
+++ b/velox/docs/functions/presto/hyperloglog.rst
@@ -70,3 +70,10 @@ Functions
 
     Returns the ``HyperLogLog`` of the aggregate union of the individual ``hll``
     HyperLogLog structures.
+
+.. function:: merge_hll(array(HyperLogLog)) -> HyperLogLog
+
+    Returns the ``HyperLogLog`` of the union of an array of ``HyperLogLog`` structures.
+
+    * Returns ``NULL`` if the input array is ``NULL``, empty, or contains only ``NULL`` elements
+    * Ignores ``NULL`` elements and merges only valid ``HyperLogLog`` structures when the array contains a mix of ``NULL`` and non-null elements
--- a/velox/functions/prestosql/registration/HyperLogFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/HyperLogFunctionsRegistration.cpp
@@ -31,5 +31,10 @@ void registerHyperLogFunctions(const std::string& prefix) {
       Constant<double>>({prefix + "empty_approx_set"});
   registerFunction<EmptyApproxSetFunction, HyperLogLog>(
       {prefix + "empty_approx_set"});
+
+  exec::registerVectorFunction(
+      prefix + "merge_hll",
+      MergeHllFunction::signatures(),
+      std::make_unique<MergeHllFunction>());
 }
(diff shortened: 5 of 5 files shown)


velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,142 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+      std::optional<DenseHll> denseHll;
+
+      for (vector_size_t i = 0; i < arraySize; ++i) {
+        auto elementIndex = arrayOffset + i;
+        if (arrayElements->isNullAt(elementIndex)) {
+          continue;
+        }
+
+        auto hllData =
+            arrayElements->as<FlatVector<StringView>>()->valueAt(elementIndex);
```
mbasmanova: can be move this cast outside of the loop?

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,142 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+          if (common::hll::DenseHlls::canDeserialize(hllData.data())) {
+            indexBitLength = common::hll::DenseHlls::deserializeIndexBitLength(
+                hllData.data());
+          } else if (common::hll::SparseHlls::canDeserialize(hllData.data())) {
+            indexBitLength = common::hll::SparseHlls::deserializeIndexBitLength(
+                hllData.data());
+          }
+        }
+
+        if (common::hll::SparseHlls::canDeserialize(hllData.data())) {
```
mbasmanova: Can we restructure the code to avoid calling this twice? here and on L144

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,142 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+      }
+
+      if (!sparseHll && !denseHll) {
+        flatResult->setNull(row, true);
+      } else if (isSparse) {
+        auto size = sparseHll->serializedSize();
+        auto stringView =
+            StringView(flatResult->getRawStringBufferWithSpace(size), size);
+        sparseHll->serialize(
+            indexBitLength, const_cast<char*>(stringView.data()));
```
mbasmanova: const_cast is an anti-pattern; let's modify the code to remove it

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -52,7 +51,10 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
 
   std::shared_ptr<memory::MemoryPool> pool_{
       memory::memoryManager()->addLeafPool()};
-  HashStringAllocator allocator_{pool_.get()};
+  // For tests using HashStringAllocator
+  HashStringAllocator hsAllocator_{pool_.get()};
+  // For tests using MemoryPoolStlAllocator
+  memory::StlAllocator<uint8_t> allocator_{*pool_};
```
mbasmanova: Is this used?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +114,78 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  auto serialized1 = serialize(indexBitLength, sparseHll1);
+
+  // Create sparse HLL with elements 10-19 using memory::MemoryPool
+  common::hll::SparseHll<memory::MemoryPool> sparseHll2{pool_.get()};
+  for (int i = 10; i < 20; i++) {
+    sparseHll2.insertHash(hashOne(i));
+  }
+  auto serialized2 = serialize(indexBitLength, sparseHll2);
+
+  // Use makeFlatVector to create binary data properly
```
mbasmanova: Drop this comment. It is redundant.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +114,78 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
           "cardinality(empty_approx_set(0.1))", makeRowVector(ROW({}), 1)));
 }
 
+TEST_F(HyperLogLogFunctionsTest, mergeHll) {
+  // Create simple test using memory::MemoryPool to match merge_hll
```
mbasmanova: What does this comment refer to?

Make sure all comments are full sentences.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +114,78 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  // implementation
+  const int8_t indexBitLength = 12;
+
+  // Create sparse HLL with elements 0-9 using memory::MemoryPool
```
mbasmanova: Why are we using MemoryPool here? We should not try to match the impl of merge_hll because it should not depend on how HLLs were created in the first place.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Updated this and used full sentences now, was an older one

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Not anymore, removed

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: Did this a different way. Why is this an issue, is it incase casting fails?

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: Refactored

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -127,6 +127,7 @@ std::unordered_map<std::string, std::shared_ptr<ArgValuesGenerator>>
 std::unordered_set<std::string> skipFunctions = {
     "noisy_empty_approx_set_sfm", // Non-deterministic because of privacy.
     "merge_sfm", // Fuzzer can generate sketches of different sizes.
+    "merge_hll", // Fuzzer can generate sketches of different sizes.
```
mbasmanova: > Fuzzer can generate sketches of different sizes.

This just sounds that Fuzzer may generate invalid input. That should not be a problem. Why is it?

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+      if (decodedArray->isNullAt(row)) {
+        flatResult->setNull(row, true);
+        return;
+      }
+
+      auto arraySize = rawSizes[indices[row]];
+      auto arrayOffset = rawOffsets[indices[row]];
+
+      int8_t indexBitLength = -1;
+      bool isSparse = true;
```
mbasmanova: Do we need isSparse variable? Can we not just check if sparseHll.has_value()?

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+        if (hllData.empty()) {
+          continue;
+        }
+
+        bool isSparseHllData = false;
+        bool isDenseHllData = false;
+
+        if (indexBitLength < 0) {
+          if (common::hll::DenseHlls::canDeserialize(hllData.data())) {
+            isDenseHllData = true;
```
mbasmanova: It might be cleaner to write this as follows.

```
const bool isSparseHllData =
              common::hll::SparseHlls::canDeserialize(hllData.data());
const boo isDenseHllData = !isSparseHllData &&
              common::hll::DenseHlls::canDeserialize(hllData.data());

if (indexBitLength < 0) {
   if (isSparseHllData) {
       indexBitLength =...
    } else {
       indexBitLength =...
    }
}
```

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+            sparseHll->mergeWith(hllData.data());
+            if (sparseHll->overLimit()) {
+              denseHll.emplace(indexBitLength, memoryPool);
+              sparseHll->toDense(*denseHll);
+              sparseHll.reset();
+              isSparse = false;
+            }
+          } else {
+            SparseHll temp{hllData.data(), memoryPool};
+            temp.toDense(*denseHll);
```
mbasmanova: Does this code path work? Do we have test coverage for it?

Somehow, it seems like *denseHll will fail / crash here because denseHll is std::nullopt.

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+          }
+        } else if (common::hll::DenseHlls::canDeserialize(hllData.data())) {
```
mbasmanova: isDenseHllData

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+          auto hllIndexBitLength =
+              common::hll::DenseHlls::deserializeIndexBitLength(hllData.data());
+          VELOX_USER_CHECK_EQ(
```
mbasmanova: Vector functions should not throw. See https://facebookincubator.github.io/velox/develop/scalar-functions.html#try-expression-support

Please, fix and add test coverage.

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,149 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+            denseHll.emplace(indexBitLength, memoryPool);
+            if (sparseHll) {
+              sparseHll->toDense(*denseHll);
+              sparseHll.reset();
+            }
+            isSparse = false;
+          }
+          denseHll->mergeWith(hllData.data());
+        } else {
+          VELOX_USER_FAIL("Invalid HLL data format");
```
mbasmanova: ditto

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -52,7 +51,7 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
 
   std::shared_ptr<memory::MemoryPool> pool_{
       memory::memoryManager()->addLeafPool()};
-  HashStringAllocator allocator_{pool_.get()};
+  HashStringAllocator hsAllocator_{pool_.get()};
```
mbasmanova: Is this change needed?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +111,76 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
           "cardinality(empty_approx_set(0.1))", makeRowVector(ROW({}), 1)));
 }
 
+TEST_F(HyperLogLogFunctionsTest, mergeHll) {
+  // Test merging two HLL instances with different value ranges.
+  const int8_t indexBitLength = 12;
+
+  // Create sparse HLL with elements 0-9 using memory::MemoryPool
+  common::hll::SparseHll<memory::MemoryPool> sparseHll1{pool_.get()};
```
mbasmanova: Why can't we just use `SparseHll<> sparseHll{allocator_}` ?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +111,76 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  for (int i = 0; i < 10; i++) {
+    sparseHll1.insertHash(hashOne(i));
+  }
+  auto serialized1 = serialize(indexBitLength, sparseHll1);
```
mbasmanova: Avoid variable names a1, a2, a3... These are hard to read.

```
std::vector<std::string> serializedHlls;
{
  SparseHll<> hll{allocator_}
  for (int i = 0; i < 10; i++) {
    hll.insertHash(hashOne(i));
  }
  serializedHlls.emplace_back(serialize(indexBitLength, hll));
}

...for (int i = 10; i < 20; i++)...
```

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +111,76 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+    sparseHll2.insertHash(hashOne(i));
+  }
+  auto serialized2 = serialize(indexBitLength, sparseHll2);
+
+  auto elements = makeFlatVector<StringView>(
+      {StringView(serialized1), StringView(serialized2)}, HYPERLOGLOG());
+  auto arrayVector = makeArrayVector({0, 2}, elements);
+
+  auto cardinalityValue = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
```
mbasmanova, line 162: Let's also add tests for a mix of sparse and dense hlls.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +111,76 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+
+  // Should be close to 20 unique elements
+  EXPECT_GE(cardinalityValue, 18);
```
mbasmanova: Can we just use EXPECT_EQ? These algorithms are deterministic, no?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Not anymore removed

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Makes sense updated

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: Thanks for the link. Very useful. Updated to applyToSelectedNoThrow() instead of applyToSelected() does that make sense?

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: I added a test for this. We should only enter here now if denseHLL is not empty

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: yes updated

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -129,10 +129,10 @@ std::unordered_set<std::string> skipFunctions = {
     "merge_sfm", // Fuzzer can generate sketches of different sizes.
     "element_at",
     "width_bucket",
-    // Fuzzer and the underlying engine are confused about TDigest functions
-    // (since TDigest is a user defined type), and tries to pass a
-    // VARBINARY (since TDigest's implementation uses an
+    // functions (since TDigest is a user defined type), and tries to pass a
```
mbasmanova: The beginning of the sentence appears to be lost. Please, double check.

Is there an issue about this limitation in the Fuzzer? If not, let's create one. CC: @kagamiori

velox/functions/prestosql/HyperLogLogFunctions.h
mbasmanova: Thank you for adding a test. Did this code path work? Did you only add a test or also fixed this code path?

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,151 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+              indexBitLength,
+              hllIndexBitLength);
+
+          auto cardinality =
+              common::hll::SparseHlls::cardinality(hllData.data());
+          if (cardinality == 0) {
+            continue;
+          }
+
+          if (sparseHll.has_value() || !denseHll.has_value()) {
```
mbasmanova: This condition is a bit hard to read. Perhaps, it would be clearer if it was written as

```
if (denseHll.has_value()) {
   ...
} else {
   ...
}
```

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +112,227 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  }
+  {
+    SparseHll<> hll{&allocator_};
+    for (int i = 10; i < 20; i++) {
+      hll.insertHash(hashOne(i));
+    }
+    serializedHlls.emplace_back(serialize(indexBitLength, hll));
+  }
+
+  auto elements = makeFlatVector<StringView>(
```
mbasmanova: simplify to `makeFlatVector(serializedHlls, HYPERLOGLOG())`

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +112,227 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  std::vector<std::string> serializedHlls;
+
+  // Create sparse HLL with fewer elements (stays sparse)
+  {
+    SparseHll<> sparseHll{&allocator_};
+    for (int i = 0; i < 50; i++) {
+      sparseHll.insertHash(hashOne(i));
+    }
+    serializedHlls.emplace_back(serialize(indexBitLength, sparseHll));
+  }
```
mbasmanova: This code repeats multiple times. Consider introducing helper function.

```
serializedHlls.emplace_back(serialize(indexBitLength, makeSparseHll(0, 50)));
serializedHlls.emplace_back(serialize(indexBitLength, makeDenseHll(50, 2000)));
```

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: Agree updated

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
natashasehgal: Updated the comment thanks.

A Hyperloglog generator can be added, one exists for TDigest and QDigest now. 
The output varbinaries are compared directly in fuzzer for scalar functions. For aggregate functions there's support for custom comparison. 

Another solution Wei and I discussed is not having a function that generates varbinaries as the outer most function

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,151 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+          VELOX_USER_CHECK_EQ(
+              indexBitLength,
+              hllIndexBitLength,
+              "Cannot merge HLLs with different indexBitLength: {} vs {}",
+              indexBitLength,
+              hllIndexBitLength);
+
+          auto cardinality =
+              common::hll::SparseHlls::cardinality(hllData.data());
+          if (cardinality == 0) {
```
mbasmanova: Why do we need this check for SparseHll, but not for DenseHll? Is there a test that hits this condition?

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,151 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+            continue;
+          }
+
+          if (denseHll.has_value()) {
+            SparseHll temp{hllData.data(), memoryPool};
+            temp.toDense(*denseHll);
+          } else {
+            if (!sparseHll) {
```
mbasmanova: consistency: !sparseHll.has_value()

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,151 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+        }
+
+        if (isSparseHllData) {
+          auto hllIndexBitLength =
+              common::hll::SparseHlls::deserializeIndexBitLength(
+                  hllData.data());
+          VELOX_USER_CHECK_EQ(
+              indexBitLength,
+              hllIndexBitLength,
+              "Cannot merge HLLs with different indexBitLength: {} vs {}",
```
mbasmanova: drop ": {} vs {}", this is included automatically

velox/functions/prestosql/HyperLogLogFunctions.h
```diff
@@ -78,4 +78,151 @@ struct EmptyApproxSetWithMaxErrorFunction {
...
+            indexBitLength = common::hll::DenseHlls::deserializeIndexBitLength(
+                hllData.data());
+          }
+        }
+
+        if (isSparseHllData) {
+          auto hllIndexBitLength =
+              common::hll::SparseHlls::deserializeIndexBitLength(
+                  hllData.data());
+          VELOX_USER_CHECK_EQ(
```
mbasmanova: This check is duplicated in the 'if' and 'else' branches. Let's move it out. Let's also use std::optional for indexBitLength.

```
VELOX_USER_CHECK(isSparseHllData || isDenseHllData, "Invalid HLL data format");

const auto hllIndexBitLength = deserializeIndexBitLength(isSparseHllData, hllData.data());

if (indexBitLength.has_value()) {
    VELOX_USER_CHECK_EQ(indexBitLength.value(), hllIndexBitLength,...);
} else {
    indexBitLength = hllIndexBitLength;
}

```

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +125,217 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  // Test merging two HLL instances with different value ranges.
+  const int8_t indexBitLength = 12;
+
+  std::vector<std::string> serializedHlls;
+  {
+    SparseHll<> hll{&allocator_};
+    for (int i = 0; i < 10; i++) {
+      hll.insertHash(hashOne(i));
+    }
+    addSerializedHll(serializedHlls, indexBitLength, hll);
```
mbasmanova: What's the value of changing serializedHlls.emplace_back with addSerializedHll? It doesn't help reduce code duplication. Instead, try

```
std::string makeSparseHll(int start, int end, int indexBitLength) {
    SparseHll<> hll{&allocator_};
    for (int i = start; i < end; i++) {
      hll.insertHash(hashOne(i));
    }

    return serialize(indexBitLength, hll);
}

std::vector<std::string> serializedHlls;
serializedHlls.emplace_back(makeSparseHll(0, 10, indexBitLength));
serializedHlls.emplace_back(makeSparseHll(10, 20, indexBitLength));

```

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
mbasmanova: @natashasehgal Would you create a GitHub issue to describe this Fuzzer limitation in details. Let's figure out how to address it. 

CC: @kagamiori @peterenescu

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: You're right updated this

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: Good idea for index bit also optional instead of -1 as sentinel

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: It's an optimization. Empty sparse HLL are common, and cheap to check cardinality. The same is not true for DenseHLL. 

Java doesn't have this check, removing it

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
natashasehgal: https://github.com/facebookincubator/velox/issues/14964

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
natashasehgal: Updated message added new message

velox/functions/prestosql/HyperLogLogFunctions.h
natashasehgal: I added a test and tested with review comment that added denseHll has value check first

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
...
+  }
+
+  template <typename HllType>
+  void populateHll(HllType& hll, int start, int end, int modValue = 0) {
+    for (int i = start; i < end; i++) {
+      hll.insertHash(hashOne(modValue ? (i % modValue) : i));
+    }
+  }
+
+  void addSerializedHll(
```
mbasmanova: This function is not needed.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
     return serialized;
   }
 
+  // Helpers to create HLL for testing
+  template <bool IsSparse = true>
+  std::string
+  createHll(int8_t indexBitLength, int start, int end, int modValue = 0) {
```
mbasmanova: Let's add createSparse and createDense methods. These are more readable than `create<true>` and `create<false>`.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
     return serialized;
   }
 
+  // Helpers to create HLL for testing
```
mbasmanova: This comment is not very informative. Please, explain what this method actually does.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
...
+      DenseHll<> hll{indexBitLength, &allocator_};
+      for (int i = start; i < end; i++) {
+        hll.insertHash(hashOne(modValue ? (i % modValue) : i));
+      }
+      return serialize(hll);
+    }
+  }
+
+  template <typename HllType>
+  void populateHll(HllType& hll, int start, int end, int modValue = 0) {
```
mbasmanova: This function is used only one, hence, not needed. Let's remove.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -91,13 +118,12 @@ TEST_F(HyperLogLogFunctionsTest, cardinalityDense) {
...
-  for (int i = 0; i < 10'000'000; i++) {
-    denseHll.insertHash(hashOne(i));
-  }
+  DenseHll<> expectedHll{12, &allocator_};
+  populateHll(expectedHll, 0, 10000000);
+  auto expectedCardinality = expectedHll.cardinality();
 
-  auto serialized = serialize(denseHll);
-  EXPECT_EQ(denseHll.cardinality(), cardinality(serialized));
+  auto serialized = createHll<false>(12, 0, 10000000);
```
mbasmanova: 10000000 -> 10'000'000 for readability

How long does this test take? Let's make sure it doesn't take more than a second or two.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +138,151 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
           "cardinality(empty_approx_set(0.1))", makeRowVector(ROW({}), 1)));
 }
 
+TEST_F(HyperLogLogFunctionsTest, mergeHll) {
+  // Test merging two HLL instances with different value ranges.
+  const int8_t indexBitLength = 12;
+
+  std::vector<std::string> serializedHlls;
+  serializedHlls.emplace_back(createHll<true>(indexBitLength, 0, 10));
```
mbasmanova: shorten to

```
std::vector<std::string> serializedHlls = {
   createSparse(...),
   createSparse(...),
};
```

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +138,151 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+
+  auto cardinalityValue = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+
+  EXPECT_EQ(1978, cardinalityValue);
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllWithNull) {
+  auto serialized = createHll<true>(12, 0, 10);
+  auto elements = makeFlatVector<StringView>(
```
mbasmanova: Can you not do makeFlatVector({serialized, "", serialized}, HYPERLOGLOG())?

mbasmanova: Or, even makeNullableFlatVector({serialized, std::nullopt, serialized}, HYPERLOGLOG()) and remove setNull?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +138,151 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+      {StringView(serialized), StringView(), StringView(serialized)},
+      HYPERLOGLOG());
+  elements->setNull(1, true);
+  auto arrayVector = makeArrayVector({0, 3}, elements);
+
+  auto cardinalityValue = evaluateOnce<int64_t>(
```
mbasmanova: cardinalityValue -> cardinality

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +138,151 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+TEST_F(HyperLogLogFunctionsTest, mergeHllAllNulls) {
+  auto input = makeRowVector({makeArrayVectorFromJson<std::string>(
+      {"[null, null, null]"}, ARRAY(HYPERLOGLOG()))});
+  auto result = evaluate("merge_hll(c0)", input);
+  ASSERT_TRUE(result->isNullAt(0));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllInvalidData) {
+  // Test with invalid HLL data format
+  auto invalidData = std::string("invalid_hll_data");
```
mbasmanova: std::string invalidData = "invalid_hll_data";

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +138,151 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllAllNulls) {
+  auto input = makeRowVector({makeArrayVectorFromJson<std::string>(
+      {"[null, null, null]"}, ARRAY(HYPERLOGLOG()))});
+  auto result = evaluate("merge_hll(c0)", input);
+  ASSERT_TRUE(result->isNullAt(0));
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllInvalidData) {
+  // Test with invalid HLL data format
```
mbasmanova: Make sure all comments are full sentences. Start with a capital letter and end with a period.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Great idea

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: Updated please check comments

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
natashasehgal: This takes 1419 ms. Will keep it

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
     return serialized;
   }
 
+  // Creates serialized HLL with specified parameters for testing.
```
mbasmanova: drop "for testing" as it is redundant.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
...
+  // Uses sparse or dense format and populates with hash values from start to
```
mbasmanova: Would you document 'modValue' and specify whether start / end are inclusive or not?

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -50,6 +50,38 @@ class HyperLogLogFunctionsTest : public functions::test::FunctionBaseTest {
...
+  // end range.
+  template <bool IsSparse = true>
+  std::string
+  createHll(int8_t indexBitLength, int start, int end, int modValue = 0) {
+    if constexpr (IsSparse) {
```
mbasmanova: This function can be remove and relevant parts moved to createSparse and createDense. There is no code shared for different template parameters.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -77,12 +109,7 @@ TEST_F(HyperLogLogFunctionsTest, cardinalitySparse) {
     return evaluateOnce<int64_t>("cardinality(c0)", HYPERLOGLOG(), input);
   };
 
-  SparseHll<> sparseHll{&allocator_};
-  for (int i = 0; i < 1'000; i++) {
-    sparseHll.insertHash(hashOne(i % 17));
-  }
-
-  auto serialized = serialize(11, sparseHll);
+  auto serialized = createHll<true>(11, 0, 1000, 17);
```
mbasmanova: Do not use template directly. Use createSparse or createDense.

velox/functions/prestosql/tests/HyperLogLogFunctionsTest.cpp
```diff
@@ -112,5 +140,134 @@ TEST_F(HyperLogLogFunctionsTest, emptyApproxSet) {
...
+  auto cardinality = evaluateOnce<int64_t>(
+      "cardinality(merge_hll(c0))", makeRowVector({arrayVector}));
+
+  EXPECT_EQ(1978, cardinality);
+}
+
+TEST_F(HyperLogLogFunctionsTest, mergeHllWithNull) {
+  auto serialized = createHll<true>(12, 0, 10);
+  auto elements = makeNullableFlatVector<StringView>(
+      {StringView(serialized), std::nullopt, StringView(serialized)},
```
mbasmanova: same comments as elsewhere; please, make sure to apply comments to all relevant pieces of code

PR #14805, opened 2025-09-09
Fix CI

MBkkt:


MBkkt:
@assignUser I think I fixed it, will you look?

MBkkt:
Fixed as part of this PR https://github.com/facebookincubator/velox/pull/14403

assignUser:
@MBkkt thanks for the investigation and fix!

PR #14810, opened 2025-09-10
Bump default DWRF flat map key limit to 30k

sdruzkin:
Summary: Bump default DWRF flat map key limit to 30k to reduce FBETL oncall load since they are bumping the limit every time when they run into it anyway.

Differential Revision: D82073848


HuamengJiang (approved):

pedroerp (approved):

velox/dwio/dwrf/common/Config.cpp
```diff
@@ -182,7 +182,7 @@ Config::Entry<const std::vector<std::vector<std::string>>>
 
 Config::Entry<uint32_t> Config::MAP_FLAT_MAX_KEYS(
     "orc.map.flat.max.keys",
-    20000);
+    30000);
```
HuamengJiang, line 185: @sdruzkin, I recall there's a discussion on bumping to 25k instead. Could you remind me of the resolution there?

I'm also curious how much memory increase a 5K/10K per column bump tends to bring, in case you already have a measurement (from vader maybe?).

PR #14812, opened 2025-09-10
build(pyvelox): Don't install pyvelox_libraries with VELOX_BUILD_PYTHON_PACKAGE=NO

### LINKED ISSUE #14756: Don't install pyvelox component by default

kou:
Fixes #14756


assignUser (approved):
Thanks!

PR #14829, opened 2025-09-11
build: Remove needless `gflags::gflags` workaround

### LINKED ISSUE #14760: Improve GFlags CMake target handling
### Description

`find_package(gflags)` creates `gflags` CMake target not `gflags::gflags` CMake target by default. But we want to always use `gflags::gflags` as GFlags CMake target. So we have a workaround for this: https://github.com/facebookincubator/velox/blob/7274da53db330bb8ad691d394818248809c10ca8/CMakeLists.txt#L533-L541

The workaround creates internal `gflags_gflags` interface CMake target for portability. (It seems that this portability is for Presto C++.)

CMake 3.18 or later can create an alias CMake target for non-GLOBAL CMake target:

https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries

> Added in version 3.18: An ALIAS can target a non-GLOBAL Imported Target. Such alias is scoped to the directory in which it is created and below. The [ALIAS_GLOBAL](https://cmake.org/cmake/help/latest/prop_tgt/ALIAS_GLOBAL.html#prop_tgt:ALIAS_GLOBAL) target property can be used to check if the alias is global or not.

We require CMake 3.28 or later. So we may be able to improve this workaround.

kou:
Fixes #14760

It's needed for old CMake such as CMake 3.16. CMake 3.18 or later can create an alias CMake target for non-`GLOBAL` CMake target:

https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries

> Added in version 3.18: An ALIAS can target a non-GLOBAL Imported
> Target. Such alias is scoped to the directory in which it is created
> and below. The ALIAS_GLOBAL target property can be used to check if
> the alias is global or not.

Now, we require CMake 3.28 or later. So we can remove the `gflags_gflags` workaround.

I confirmed that this change works with Presto C++.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -544,9 +544,7 @@ if(NOT TARGET gflags::gflags)
   # target even when velox is built as a subproject which uses
   # `find_package(gflags)` which does not create a globally imported target that
   # we can ALIAS.
-  add_library(gflags_gflags INTERFACE)
-  target_link_libraries(gflags_gflags INTERFACE gflags)
-  add_library(gflags::gflags ALIAS gflags_gflags)
+  add_library(gflags::gflags ALIAS gflags)
 endif()
 
 if(${gflags_SOURCE} STREQUAL "BUNDLED")


kagamiori:
Hi @kou, could you please rebase this PR onto the latest main? This is needed for merging the PR. Thanks!

kou:
Sure! Done!

Issue #13942, opened 2025-06-30
Spark Aggregate Fuzzer crashes w/ Segmentation fault


mbasmanova:
### Description

https://github.com/facebookincubator/velox/actions/runs/15984885549/job/45089189142?pr=13938

```
E20250630 16:10:21.989187   148 AggregationFuzzerBase.cpp:678] Total functions: 18 (34 signatures)
E20250630 16:10:21.989313   148 AggregationFuzzerBase.cpp:682] Functions with at least one supported signature: 18 (100.00%)
E20250630 16:10:21.989331   148 AggregationFuzzerBase.cpp:688] Functions with no supported signature: 0 (0.00%)
E20250630 16:10:21.989353   148 AggregationFuzzerBase.cpp:691] Supported function signatures: 30 (88.24%)
E20250630 16:10:21.989377   148 AggregationFuzzerBase.cpp:697] Unsupported function signatures: 4 (11.76%)
I0630 16:10:27.286002   148 Compression.cpp:629] Initialized zstd compressor with compression level 7
*** Aborted at 1751325547 (Unix time, try 'date -d @1751325547') ***
*** Signal 11 (SIGSEGV) (0x0) received by PID 148 (pthread TID 0x7f1a44554740) (linux TID 148) (code: address not mapped to object), stack trace: ***
(error retrieving stack trace)
/__w/_temp/28b1a7ab-4689-4665-b054-18075571e5a3.sh: line 17:   148 Segmentation fault      (core dumped) ./spark_aggregation_fuzzer_test --seed ${random_seed} --duration_sec $DURATION --enable_sorted_aggregations=false --minloglevel=0 --stderrthreshold=2 --log_dir=/tmp/spark_aggregate_fuzzer_repro/logs --repro_persist_path=/tmp/spark_aggregate_fuzzer_repro
Error: Process completed with exit code 139.
```

CC: @rui-mo @kagamiori @bikramSingh91 

### Relevant logs

```bash

```

bikramSingh91:
Looks like this was recently enabled in https://github.com/facebookincubator/velox/pull/9559
@rui-mo do you mind looking into this if you have spark setup on your machine?  I was able to run the spark agg fuzzer without Spark as the source of truth so having it might be necessary to repro

rui-mo:
Yeah, I have the Spark test environment. I tried it on my end but was not able to repro the segmentation fault. I'll try more runs.

Update:

[...]

PR #14831, opened 2025-09-11
fix(fuzzer): Disable streaming aggregation for Spark fuzzer test

rui-mo:
Since the last workflow failures occurred in the streaming aggregation test, 
this PR temporarily disables it and increases the wait time for Spark connect
service to start.

https://github.com/facebookincubator/velox/issues/13942, https://github.com/facebookincubator/velox/issues/15993

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -603,7 +603,7 @@ jobs:
         run: |
           bash /opt/start-spark.sh
           # Sleep for 60 seconds to allow Spark server to start.
-          sleep 60
+          sleep 120
           mkdir -p /tmp/spark_aggregate_fuzzer_repro/logs/
           chmod -R 777 /tmp/spark_aggregate_fuzzer_repro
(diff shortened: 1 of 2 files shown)


rui-mo:
@mbasmanova I’m still unable to reproduce the failure locally. Since the last two workflow failures both occurred in the streaming aggregation test, I’m disabling it in this PR to see if that helps, and converting the PR to draft for now. Thanks.

rui-mo:
Hi @mbasmanova @kevinwilfong, regarding the Spark aggregate fuzzer issue, could we merge this PR first and observe its impact in future runs? This change disables the streaming tests and increases the service startup time, but if the workflow fails again, we may get clearer clues.

kgpai (approved):

kgpai:
The spark aggregate fuzzer only ran for 18 minutes , so not sure that is enough signal, but we know the error is when streaming is enabled, so I think we should be good to ship this.

PR #14834, opened 2025-09-11
refactor: Use 'cbegin()' to assign to 'const_iterator'

lingbin:
Using `cbegin()` makes the intent more explicit, although it's not 
actually necessary because `positions` is already a const vector.


lingbin:
@kagamiori  Could you please review this PR? It is a small change.

PR #14837, opened 2025-09-11
feat: Add filter support for index join which can't be converted into join conditions

xiaoxmeng:
Summary:
Add filter support in index join for filter which can't be converted into join conditions that can be pushdown to index source. The filter is executed on the lookup result before left join processing. This is to enable Meta AI data exploration query shapes

The followup is to consider use join match tracker inside index lookup to handle this logic to be consistent with other join type implementations.

Differential Revision: D82149399


zacw7 (approved):
Thanks for adding the support!

PR #14850, opened 2025-09-12
refactor: Readability improvements, avoid few allocations

MBkkt:
* fix: Some places have extra semicolon, remove it, make format better
* fix: Add missing includes
* refactor: type_index is just a raw pointer to type_info, so better to be passed by value
* refactor: Avoid allocations for static shared_ptr
* fix: Avoid using iterators as pointers
* fix: Avoid implicit cast
* refactor: Avoid explicit inline (it already was inline). Use constexpr for size-like stuff.

I found this during my other work, so it's convenient for me to make it single PR.
Also changes are small and well separated.

### DIFF
--- a/velox/type/OpaqueCustomTypes.h
+++ b/velox/type/OpaqueCustomTypes.h
@@ -57,9 +57,6 @@ class OpaqueCustomTypeRegister {
 
     static const TypePtr& get() {
       static const VeloxType kInstance;
-      // TODO: kInstancePtr needed to return const TypePtr& instead of TypePtr.
-      // In general it's not really necessary, but there's a lot of places that
-      // depends on this in fb internal repos. So for now we keep it like this.
       static const TypePtr kInstancePtr{TypePtr{}, &kInstance};
       return kInstancePtr;
     }


velox/type/OpaqueCustomTypes.h
```diff
@@ -55,9 +55,9 @@ class OpaqueCustomTypeRegister {
    public:
     VeloxType() : OpaqueType(std::type_index(typeid(T))) {}
 
-    static const TypePtr& get() {
-      static const TypePtr instance{new VeloxType()};
-      return instance;
+    static TypePtr get() {
+      static const VeloxType kInstance;
+      return {TypePtr{}, &kInstance};
```
mbasmanova: @MBkkt I'm seeing errors internally. I believe they are caused by this change.

```
error: non-const lvalue reference to type 'shared_ptr<...>' cannot bind to a temporary of type 'shared_ptr<...>'
   26 |   auto& typePtr = T::singletonTypePtr();
      |         ^         ~~~~~~~~~~~~~~~~~~~~~
```

MBkkt: Yes, you're right, should I wrap it in `VELOX_ENABLE_BACKWARD_COMPATIBILITY` before callsites fixes?

mbasmanova: I do not think it is feasible to change the callsites. Would you revert this specific change?

MBkkt: I can but I don't understand WDYT it's not feasible?

This place written in strange way, it should be written `const auto&` or `auto`, but not `auto&`.

mbasmanova: > I don't understand WDYT it's not feasible

Too many places to change. Many of these require extensive testing. I simply don't have bandwidth to do that now.

MBkkt: Probably should works now https://github.com/facebookincubator/velox/pull/14850/commits/c54fcb25cfa525e19b345aa604b300602b79312f

velox/type/OpaqueCustomTypes.h
```diff
@@ -55,9 +55,13 @@ class OpaqueCustomTypeRegister {
    public:
     VeloxType() : OpaqueType(std::type_index(typeid(T))) {}
 
-    static TypePtr get() {
+    static const TypePtr& get() {
       static const VeloxType kInstance;
-      return {TypePtr{}, &kInstance};
+      // TODO: kInstancePtr needed to return const TypePtr& instead of TypePtr.
+      // In general it's not really necessary, but there's a lot of places that
```
mbasmanova: Let's remove this TODO. I don't think many readers would understand why "In general it's not really necessary" and why anything needs changing here.

MBkkt: It's unnecessary because we create shared_ptr in `.data` that just holds pointer to another object in `.data`.
Simpler and probably faster will create this std::shared_ptr on every call (because everything will be inlined in practice).

I can rephrase it, would you suggest way how it will be understandable by many readers?

I think I want to keep TODO, because right now, it's the only place in velox with such pattern, every other places translated to code like it was before commit with "Apply review suggestions"

MBkkt: So the point is not that code `kInstancePtr` is something really bad.
But point is that the only place that differs from others and it's slightly worse.
And I hope this TODO prevent someone copy this code to some new code.

mbasmanova: I suggest to send an update to https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md, then reference newly-added guideline in this TODO.

It would be nice if you could remove the TODO from this PR, then re-introduce it in a PR that updates coding style. This will allow me to proceed with merging this PR without further delay.

MBkkt: Ok, removed todo

mbasmanova:
@MBkkt Would you rebase?

MBkkt:
@mbasmanova done

Issue #14833, opened 2025-09-11
Use parquet field_id for column matching and prune in IcebergSplitReader


PingLiuPing:
### Description

Problem

`IcebergSplitReader` currently reuses the Hive `SplitReader`, which relies on column name or positional matching to adapt Parquet fields with query columns.
This causes problems under certain schema evolution scenarios. For example, if a column is dropped and later re-added with the same name but a different type, the current matching logic becomes incorrect.
In Iceberg, column projection is defined by field IDs (see [Iceberg Spec: Column Projection](https://iceberg.apache.org/spec/#column-projection)). Using names or positions is not sufficient or correct.

See Iceberg Java code for the actual matching logic:

https://github.com/apache/iceberg/blob/d1771207c9040f1c8b6886665b56d1a972fe402a/parquet/src/main/java/org/apache/iceberg/parquet/ReadConf.java#L79-L89

Solution
As the first step, we should support using field_id matching. With this we will add more data structures to support iceberg table name-mapping properties, and then upstream engines can pass down the approperiate name-mapping objects. Finally we will implement the fallbackIds (assign field id to parquet field).

For first step, we can extend `ReaderOptions` with a new column matching mode that supports field ID–based matching. Add an additional parameter to pass in the requested field IDs (instead of only the requested type).
Then use field IDs for matching against the corresponding SchemaElement field IDs in Parquet metadata.

This will ensure correct column resolution even in cases of schema evolution (rename, drop/re-add with different type, etc.).

Yuhta:
In this case, we should receive `field_id` as field names and the engine (possible with an extra project node) renames it back to user facing names later.  Otherwise a lot of the prunings won't work as field names are supposed to be the unique identifier.  The only change needed is to match the field names with `field_id` in parquet file reader.

PingLiuPing:
@Yuhta Thanks for the suggestion. This approach indeed helps minimize the code changes in Velox.In [#14851](https://github.com/facebookincubator/velox/pull/14851), I’ve included two commits, the first follows my initial design, where field IDs are passed explicitly. This avoids changes to the query plan; the second takes the approach you suggested: treating the field_id as the column name so we can reuse the existing matching logic.

PR #14851, opened 2025-09-12
feat: Support using parquet field_id for column matching

PingLiuPing:
resolve https://github.com/facebookincubator/velox/issues/14833


velox/dwio/common/Options.h
```diff
@@ -550,11 +558,26 @@ class ReaderOptions : public io::ReaderOptions {
...
+  }
+
+  ReaderOptions& setParquetFieldIds(
+      std::shared_ptr<std::vector<ParquetFieldIdBinding>> fieldIds) {
+    parquetFieldIds_ = std::move(fieldIds);
+    return *this;
+  }
+
+  const std::shared_ptr<const std::vector<ParquetFieldIdBinding>>&
+  parquetFieldIds() const {
```
Yuhta: Where is this supposed to be used?

PingLiuPing: @Yuhta Thanks, they are relate to the first commit. Will remove them.

PR #14852, opened 2025-09-12
fix(build): ABFS connector install for macOS

czentgr:
The sed program usage only works on Linux and needs a fix for macOS. This is the same fix that was applied for libstemmer.


assignUser (approved):
Thanks!

PR #14856, opened 2025-09-12
refactor(test): Remove duplicate code in StringFunctionsTest

czentgr:
There are two test helper functions doing the same thing. Cleanup to use one implementation and also clean up a declaration that has no implementation.


majetideepak (approved):

majetideepak:
@czentgr can you rebase with latest main? Thanks.

PR #14873, opened 2025-09-16
feat(streaming): add streaming related support

shuai-xu:
As proposed in https://github.com/facebookincubator/velox/discussions/14793, we try to make velox can support streaming engine like Flink.
Streaming processing has some difference with batch. 
1. A streaming task may contains some streaming operators, and a record is passed down from one operator to its downstreams at once without persistant. So the pull based mode of velox that sink call its source to get the output and so on doesn't fit for streaming. Instead, we should use the pushed based mode, that source generate outputs continously, and then push the records to its targets.
2. Streaming processing need state and watermark, some operators may store the temporary result in state and then calculate a result with the state and new coming input. Watermark may be used to trigger a window or finish a partition.
In this pr, we create the base framework to support streaming.
The class StreamingTask is used to manage the pushed based processing, it return a result every time called next(). It also has some state related method like snapshot().
StreamingOperator is used to manager the processing logic. Velox now have many operators, but they are all stateless, we hope we can hold the state in StreamingOperator, and reuse the existing operator to computing with input and state. This may need to modify some operators a little, as some operators like Aggs may only output when input finish, but streaming never finish.
StreamingPlanNode is used to help build the streaming graph accoring pushed model.
The result return to streaming framework may be a RowVector or a Watermark, so we use StreamRecord as the result.

### DIFF
--- a/velox/experimental/streaming/StreamingOperator.h
+++ b/velox/experimental/streaming/StreamingOperator.h
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/exec/Operator.h"
+#include "velox/experimental/streaming/CombinedWatermarkStatus.h"
+
+namespace facebook::velox::streaming {
+
+/**
+ * This class is to support streaming engines. Unlike batch jobs,
+ * operators in streaming jobs may have several outputs,
+ * so the pull mode of velox that downstream call its upstream to get output
+ * will not fit for it. We need to use push mode that sources
+ * an output, and then push to its downstreams and so on.
+ *
+ * StreamingOperator is designed to hold the state and an velox operator,
+ * it gives the state and input to velox operator to calculate the result.
+ * For example, Streaming Top1 node gets the previous result from state,
+ * and then call TopN operator to get new result, and at last update the state.
+ */
+class StreamingOperator {
+ public:
+  StreamingOperator(
+      std::unique_ptr<exec::Operator> op,
+      std::vector<std::unique_ptr<StreamingOperator>> targets)
+      : operator_(std::move(op)),
+        targets_(std::move(targets)) {
+    isSink_ = operator_->operatorType() == "TableWrite";
+  }
+
+  virtual void initialize();
+
+  virtual bool isFinished();
+
+  virtual void addInput(RowVectorPtr input);
+
+  virtual void getOutput();
+
+  bool sourceEmpty();
+
+  virtual void close();
+
+  void processWatermark(long timestamp, int index);
+
+  void snapshotState(long checkpointId);
+
+  void notifyCheckpointComplete(long checkpointId);
+
+  void notifyCheckpointAborted(long checkpointId);
+
+  virtual std::string name() const {
+    return operator_->operatorType();
+  }
+
+ protected:
+  void pushOutput(RowVectorPtr output);
+  void pushWatermark(long timestamp, int index);
+  virtual void processWatermarkInternal(long timestamp) {}
+
+  virtual int numInputs() const {
+    return 1;
+  }
+
+  std::unique_ptr<exec::Operator>& op() {
+    return operator_;
+  }
+
+ private:
+  bool isSink() {
+    return isSink_;
+  }
+
+  std::unique_ptr<exec::Operator> operator_;
+  std::vector<std::unique_ptr<StreamingOperator>> targets_;
+  bool isSink_;
+  bool sourceEmpty_ = true;
+  std::shared_ptr<CombinedWatermarkStatus> combinedWatermarkStatus_;
+};
+
+using StreamingOperatorPtr = std::unique_ptr<StreamingOperator>;
+
+} // namespace facebook::velox::streaming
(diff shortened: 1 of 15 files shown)


pedroerp:
Thank you @shuai-xu. @xiaoxmeng and @duanmeng have more context on the driver/task framework so hopefully they can help with code review.

velox/experimental/streaming/StreamingPlanner.cpp
```diff
@@ -0,0 +1,201 @@
...
+//static
+StreamingOperatorPtr StreamingPlanner::nodeToStreamingOperator(
+    const core::PlanNodePtr& planNode,
+    exec::DriverCtx* ctx) {
+  auto streamingNode =
+      std::dynamic_pointer_cast<const StreamingPlanNode>(planNode);
+  VELOX_CHECK(streamingNode, "Not streaming node: {}", planNode->toString());
+  std::vector<StreamingOperatorPtr> targets;
+  std::unique_ptr<exec::Operator> op = std::move(nodeToOperator(streamingNode->node(), ctx));
+  for (auto target : streamingNode->targets()) {
```
duanmeng, line 70: Is this `PlanNode` not a root node of a plan fragment?

velox/experimental/streaming/StreamingPlanner.cpp
```diff
@@ -0,0 +1,201 @@
...
+    targets.push_back(std::move(nodeToStreamingOperator(target, ctx)));
+  }
+  return std::make_unique<StreamingOperator>(std::move(op), std::move(targets));
```
duanmeng, line 73: Is `StreamingOperator` a combination of an `Operator Chain` and downstream chains? Are the `op` and `targets` in the same `TaskManager`?

velox/experimental/streaming/StreamingPlanner.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "velox/experimental/streaming/StreamingPlanNode.h"
+
+namespace facebook::velox::streaming {
+
+static std::atomic<int> opId = 0;
+
+// static
+StreamingOperatorPtr StreamingPlanner::plan(
+    const core::PlanFragment& planFragment,
+    exec::DriverCtx* ctx) {
```
duanmeng, line 57: It seems that you use the same `DriverCtx` to create all the `operators`, but they belong to different pipelines.

velox/experimental/streaming/StreamingTask.h
```diff
@@ -0,0 +1,98 @@
...
+  StreamElementPtr popOutput();
+
+  std::unique_ptr<StreamingOperator> operatorChain_;
+
+  // A task may have multi outputs once run,
+  // store them and return one by one.
+  std::list<StreamElementPtr> pendings_;
+
+  // hold the driver only to avoid it be released.
+  std::shared_ptr<exec::Driver> driver;
```
duanmeng, line 95: Is it just a dummy driver helping to create operators in `StreamingPlanner::plan`?

velox/experimental/streaming/StreamingOperator.h
```diff
@@ -0,0 +1,96 @@
...
+ * so the pull mode of velox that downstream call its upstream to get output
+ * will not fit for it. We need to use push mode that sources
+ * an output, and then push to its downstreams and so on.
+ *
+ * StreamingOperator is designed to hold the state and an velox operator,
+ * it gives the state and input to velox operator to calculate the result.
+ * For example, Streaming Top1 node gets the previous result from state,
+ * and then call TopN operator to get new result, and at last update the state.
+ */
+class StreamingOperator {
```
duanmeng, line 35: Is `StreamingOperator` a combination of operators, states, and a watermark? What's the relation between `op` and `targets`? Do they belong to a different `operator chain`?

velox/experimental/streaming/StreamingTask.h
```diff
@@ -0,0 +1,98 @@
...
+      core::PlanFragment planFragment,
+      std::shared_ptr<core::QueryCtx> queryCtx);
+
+  void initOperators();
+
+  void initStateBackend();
+
+  StreamElementPtr popOutput();
+
+  std::unique_ptr<StreamingOperator> operatorChain_;
```
duanmeng, line 88: Should operatorChain contain several StreamingOperators?

velox/experimental/streaming/StreamingTask.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace facebook::velox::streaming {
+
+/**
+ * StreamingTask is used to support streaming engines such as Flink.
+ * It only runs in a streaming mode, which call next and get a result once,
+ * also, it needs to handle state operations  
+ * Make it based on Task to reuse context releated classes.
+ */
+class StreamingTask : public exec::Task {
```
duanmeng, line 31: Does `StreamTask` abstract a Flink task by encapsulating an operator chain?

velox/experimental/streaming/StreamingTask.h
shuai-xu: Yes, Driver is useless for flink as each Task in flink only has one operator chain.

velox/experimental/streaming/StreamingPlanner.cpp
shuai-xu: The operators in a planFragment for streaming belong to the same pipelines. One pipeline for flink contains several operators that data can be pass from one to another directly without shuffle.

velox/experimental/streaming/StreamingPlanner.cpp
shuai-xu: Yes, we parse the plan fragment from the root PlanNode, and then its targets and so on.

velox/experimental/streaming/StreamingPlanner.cpp
shuai-xu: StreamingOperator contains the pointers of its targets, they make a operator chain. We can access an operator chain from the source StreamingOperator, the op in a StreamingOperator is for reusing Velox operators as much as possible. All the StreamingOperators belong to the same TaskManager.

velox/experimental/streaming/StreamingTask.h
shuai-xu: Yes, an operator chain contains serveral StreamingOperators.

velox/experimental/streaming/StreamingTask.h
shuai-xu: Yes

velox/experimental/streaming/StreamingOperator.h
shuai-xu: Yes, StreamingOperator combine operator, state, and watermark, operator is for the computing logic, state is for historical data, and watermark is to trigger the timer. Targets is to chain the operators to make a operator chain, so when an operator has data, we can push it to its targets. All the operators in a task belong to the same operator chain.

PR #14874, opened 2025-09-16
refactor: Reuse SplitReader::createRowReader in Meta internal SplitReader implementation

Yuhta:
Summary: So that any future change on `createRowReader` will not be forgotten.

Differential Revision: D82544173

### DIFF
--- a/velox/connectors/hive/SplitReader.cpp
+++ b/velox/connectors/hive/SplitReader.cpp
@@ -170,7 +170,7 @@ void SplitReader::prepareSplit(
     return;
   }
 
-  createRowReader(std::move(metadataFilter), std::move(rowType));
+  createRowReader(std::move(metadataFilter), std::move(rowType), std::nullopt);
 }
 
 void SplitReader::setBucketConversion(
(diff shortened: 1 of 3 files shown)


velox/connectors/hive/SplitReader.cpp
```diff
@@ -368,7 +368,8 @@ bool SplitReader::checkIfSplitIsEmpty(
 
 void SplitReader::createRowReader(
     std::shared_ptr<common::MetadataFilter> metadataFilter,
-    RowTypePtr rowType) {
+    RowTypePtr rowType,
+    std::optional<bool> rowSizeTrackingEnabled) {
```
HuamengJiang, line 372: @Yuhta, If we make it a part of the interface, do we want some more generic option structs? Or is the plan to still delete this afterward?

Yuhta: It depends on how fast we can get rid of this option, if we add column stats fast enough we can remove this.  Otherwise we keep it there, until we add a second option then we change it to a struct

PR #14879, opened 2025-09-16
Add maximum entry count limit to AsyncDataCache

zacw7:
Summary:
This change adds a maximum entry count limit to the AsyncDataCache class in Velox. The limit is set in the constructor of the CacheShard class and is used to prevent the cache from exceeding a certain number of entries. If the limit is reached, the cache attempts to evict entries to make room for the new entry. If the eviction fails, an error is thrown. 
The default value of the max entry count limit is 10 millions. Assuming the memory overhead of one cache entry is ~36 bytes: 16 bytes (RawFileCacheKey) + 8 bytes (value pointer) + ~12 bytes (HashMap overhead), the default default caps the memory overhead for storing cache entries to ~360MB.

Differential Revision: D82559227


velox/common/caching/SsdCache.h
```diff
@@ -71,6 +73,10 @@ class SsdCache {
     /// Executor for async fsync in checkpoint.
     folly::Executor* executor;
 
+    /// Maximum number of SSD cache entries allowed. A value of 0 means no
+    /// limit. When the limit is reached, new entry writes will be skipped.
+    uint64_t maxEntries;
+
```
gggrace14, line 79: Include maxEntries in `toString()`?

velox/common/caching/SsdFile.cpp
```diff
@@ -345,6 +346,16 @@ void SsdFile::clearRegionEntriesLocked(const std::vector<int32_t>& regions) {
 
 void SsdFile::write(std::vector<CachePin>& pins) {
   process::TraceContext trace("SsdFile::write");
+
+  // Check entry count limit before writing
+  if (maxEntries_ > 0) {
+    std::shared_lock<std::shared_mutex> l(mutex_);
+    if (entries_.size() + pins.size() >= maxEntries_) {
+      ++stats_.writeSsdDropped;
+      return;
```
gggrace14, line 355: Hi @zacw7 , I'm curious how this maxEntries check works with the eviction? If I understand correct, this change is, after the server starts and once the ssd writes grow and hit this maxEntries_ , we will stop writing to the ssd cache. Will it be too aggressive? We don't need to give chance to eviction, that is `clearRegionEntriesLocked(candidates)`? Does it make sense to do this check after eviction, ex. getSpace()? In that case, after we evict several times, the entries_.size() could drop below maxEntries_ and we can continue to write. Is this understanding correct?

velox/common/caching/SsdFile.h
```diff
@@ -261,7 +261,8 @@ class SsdFile {
         bool _disableFileCow = false,
         bool _checksumEnabled = false,
         bool _checksumReadVerificationEnabled = false,
-        folly::Executor* _executor = nullptr)
+        folly::Executor* _executor = nullptr,
+        uint64_t _maxEntries = 0)
```
xiaoxmeng: put this ahead of executor.

velox/common/caching/SsdFile.h
```diff
@@ -621,6 +627,9 @@ class SsdFile {
   // Executor for async fsync in checkpoint.
   folly::Executor* executor_;
 
+  // Maximum number of SSD cache entries allowed in this file. 0 means no limit.
+  const uint64_t maxEntries_;
```
xiaoxmeng: put const member first.

folly::Executor* const executor_;

velox/common/caching/SsdFile.h
```diff
@@ -296,6 +298,10 @@ class SsdFile {
 
     /// Executor for async fsync in checkpoint.
     folly::Executor* executor;
```
xiaoxmeng: are these const members?

velox/common/caching/SsdFile.cpp
```diff
@@ -345,6 +346,16 @@ void SsdFile::clearRegionEntriesLocked(const std::vector<int32_t>& regions) {
 
 void SsdFile::write(std::vector<CachePin>& pins) {
   process::TraceContext trace("SsdFile::write");
+
+  // Check entry count limit before writing
+  if (maxEntries_ > 0) {
```
xiaoxmeng, line 351: put a comment in header say 0 means no limit?

velox/common/caching/SsdFile.cpp
```diff
@@ -345,6 +346,16 @@ void SsdFile::clearRegionEntriesLocked(const std::vector<int32_t>& regions) {
...
+    std::shared_lock<std::shared_mutex> l(mutex_);
+    if (entries_.size() + pins.size() >= maxEntries_) {
+      ++stats_.writeSsdDropped;
```
xiaoxmeng:
```
s/writeSsdDropped/writeExceedSsdEntryLimit or exceedSsdEntryLimit/
```

velox/common/caching/SsdCache.cpp
```diff
@@ -58,6 +59,9 @@ SsdCache::SsdCache(const Config& config)
   const uint64_t sizeQuantum = numShards_ * SsdFile::kRegionSize;
   const int32_t fileMaxRegions =
       bits::roundUp(config.maxBytes, sizeQuantum) / sizeQuantum;
+  // Distribute maxEntries across shards
+  const uint64_t maxEntriesPerShard =
+      maxEntries_ == 0 ? 0 : (maxEntries_ + numShards_ - 1) / numShards_;
```
xiaoxmeng: use divRoundUp?

velox/common/caching/SsdFile.cpp
zacw7: mentioned in the header.

velox/common/caching/SsdCache.h
```diff
@@ -182,6 +188,8 @@ class SsdCache {
   // Stats for selecting entries to save from AsyncDataCache.
   const std::unique_ptr<FileGroupStats> groupStats_;
   folly::Executor* const executor_;
+  // Maximum number of SSD cache entries allowed. 0 means no limit.
+  const uint64_t maxEntries_;
```
xiaoxmeng, line 192: Leave an empty line between const and non-const members? Thanks!

velox/common/caching/SsdFile.cpp
```diff
@@ -345,6 +346,16 @@ void SsdFile::clearRegionEntriesLocked(const std::vector<int32_t>& regions) {
 
 void SsdFile::write(std::vector<CachePin>& pins) {
   process::TraceContext trace("SsdFile::write");
+
+  // Check entry count limit before writing
+  if (maxEntries_ > 0) {
+    std::shared_lock<std::shared_mutex> l(mutex_);
+    if (entries_.size() + pins.size() >= maxEntries_) {
```
xiaoxmeng, line 353:
```
entries_.size() + pins.size() >? not =
```
Can you have a unit test to ensure this behavior? Thanks!

PR #14887, opened 2025-09-17
feat: Add support for casting TIME to VARCHAR

kgpai:
Summary: Adds support for casting TIME to VARCHAR. Casting accounts for the local time zone and also formats it in standard HH:mm:SS.SSS format.

Differential Revision: D82653023


kevinwilfong:
Nvm, sorry I forgot you already have a test for nulls, looks good!

PR #14889, opened 2025-09-17
feat: Add IConfig interface

codeworse:
It is useful to replace `ConfigBase` with `IConfig` (a config interface). In cases where there is already an external config with variables, it is inconvenient to separately create a `ConfigBase` for `QueryConfig`. It would be more convenient to create an interface through which `QueryConfig` can read the necessary variables.

`IConfig` has only two methods: `std::optional<std::string> access(const std::string&)` and `std::unordered_map<std::string, std::string> rawConfigsCopy()`

Example where it's helpful:

SereneDB has separate maps for global, session, and transactional variables (postgres semantics). They are all accessed via IConfig interface.

### DIFF
--- a/velox/common/config/IConfig.h
+++ b/velox/common/config/IConfig.h
@@ -34,10 +34,10 @@ class IConfig {
   template <typename T>
   std::optional<T> get(
       const std::string& key,
-      std::function<T(std::string, std::string)> toT = [](auto /* unused */,
-                                                          auto value) {
-        return folly::to<T>(value);
-      }) const {
+      const std::function<T(std::string, std::string)>& toT =
+          [](auto /* unused */, auto value) {
+            return folly::to<T>(value);
+          }) const {
     if (auto val = access(key)) {
       return toT(key, *val);
     }
@@ -48,10 +48,10 @@ class IConfig {
   T get(
       const std::string& key,
       const T& defaultValue,
-      std::function<T(std::string, std::string)> toT = [](auto /* unused */,
-                                                          auto value) {
-        return folly::to<T>(value);
-      }) const {
+      const std::function<T(std::string, std::string)>& toT =
+          [](auto /* unused */, auto value) {
+            return folly::to<T>(value);
+          }) const {
     if (auto val = access(key)) {
       return toT(key, *val);
     }


mbasmanova:
> We have separate maps for global, session, and transactional variables (postgres semantics)

@codeworse Welcome to the community? Would you introduce yourself? What project you are working on?

velox/dwio/dwrf/test/TestDecompression.cpp
```diff
@@ -934,11 +934,11 @@ class TestSeek : public ::testing::Test {
   std::shared_ptr<MemoryPool> pool_ = memory::memoryManager()->addLeafPool();
 };
 
-TEST_F(TestSeek, Zlib) {
-  auto codec = zlib::getCodec(
-      zlib::Options(zlib::Options::Format::RAW), COMPRESSION_LEVEL_DEFAULT);
-  runTest(*codec, CompressionKind_ZLIB);
-}
+// TEST_F(TestSeek, Zlib) {
```
mbasmanova: Why is this commented out?

velox/common/config/IConfig.h
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include <folly/Conv.h>
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+namespace facebook::velox::config {
+
+class IConfig {
```
mbasmanova, line 32: Would you document this class?

velox/core/tests/QueryConfigTest.cpp
```diff
@@ -29,7 +29,7 @@ class QueryConfigTest : public testing::Test {
 };
 
 TEST_F(QueryConfigTest, emptyConfig) {
-  auto queryCtx = QueryCtx::create(nullptr, QueryConfig{{}});
+  auto queryCtx = QueryCtx::create(nullptr, QueryConfig{});
```
mbasmanova: Is it possible to avoid this change? There might be a lot of places that would need updating. CC: @pedroerp

velox/dwio/dwrf/test/TestDecompression.cpp
MBkkt: It was my local diff (this test doesn't build locally for me), I removed this diff

velox/core/tests/QueryConfigTest.cpp
MBkkt: Ok, fixed.

codeworse:
@mbasmanova Hi! Yes, I'm working at SereneDB, doing velox/postgres integration

velox/common/config/IConfig.h
codeworse: updated

velox/common/config/IConfig.h
```diff
@@ -0,0 +1,70 @@
...
+#pragma once
+
+#include <folly/Conv.h>
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+namespace facebook::velox::config {
+
+/// IConfig - Read-only config interface
```
mbasmanova: Thank you for documenting this class. Why are the doc lines so short? Would you make them longer to match the rest of the codebase?

codeworse: refactored

MBkkt:
@bikramSingh91 Hi, I wanted to ask, is it ok that tests working 4 days?

velox/common/config/IConfig.h
```diff
@@ -0,0 +1,68 @@
...
+/// IConfig - Read-only config interface for accessing key-value parameters.
+/// Supports value retrieval by key and duplication of the raw config data.
+/// Can be used by velox::QueryConfig to access externally managed system
+/// configuration.
+class IConfig {
+ public:
+  template <typename T>
+  std::optional<T> get(
+      const std::string& key,
+      std::function<T(std::string, std::string)> toT = [](auto /* unused */,
```
bikramSingh91: @codeworse I am getting linter failures that highlight an "unnecessary copy" for `toT` , the suggested fix is to convert this function to a `const &`. Can you please make that change here and other API underneath that have the same parameter.

codeworse: make `const std::function<T(std::string, std::string)>& toT` or are you talking about `toT` parameters(make `T(const std::string&, const std::string&)`)?

codeworse: updated

bikramSingh91: I was referring to the function pointer. Thanks for the update

PR #14898, opened 2025-09-18
misc: Remove doy alias for Spark dayofyear function

zhli1142015:
In Spark, `dayofyear` function does not have a `doy` alias, so remove it.


rui-mo (commented):
Thanks. Just wondering what kind of issues we might encounter to have an unused name.

zhli1142015:
> Thanks. Just wondering what kind of issues we might encounter to have an unused name.

No issue relalted, just noticed this when looking another issue.

PR #14904, opened 2025-09-19
fix(misc): Slightly improve row size tracking query config documentation

HuamengJiang:
Summary: Clarify the row size tracking query config documentation a little bit.

Differential Revision: D82781040


amitkdutta (approved):
Thanks @HuamengJiang

PR #14906, opened 2025-09-19
fix(SMJ): verify SMJ comparison with two join keys 

Henry2SS:
TODO

### DIFF
--- a/velox/exec/tests/MergeJoinTest.cpp
+++ b/velox/exec/tests/MergeJoinTest.cpp
@@ -369,6 +369,134 @@ class MergeJoinTest : public HiveConnectorTestBase {
         std::bind(
             &MergeJoinTest::generateLazyInput, this, std::placeholders::_1));
   }
+
+  void testJoinTwoKeysWithNulls(
+      RowVectorPtr& leftVectors,
+      RowVectorPtr& rightVectors) {
+    auto leftFile = TempFilePath::create();
+    writeToFile(leftFile->path, leftVectors);
+    createDuckDbTable("t", {leftVectors});
+    auto rightFile = TempFilePath::create();
+    writeToFile(rightFile->path, rightVectors);
+    createDuckDbTable("u", {rightVectors});
+    auto joinTypes = {
+        core::JoinType::kInner,
+        core::JoinType::kLeft,
+        core::JoinType::kRight,
+        core::JoinType::kFull,
+    };
+
+    for (auto joinType : joinTypes) {
+      LOG(ERROR) << "running " << joinTypeName(joinType);
+      auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
+      core::PlanNodeId leftScanId;
+      core::PlanNodeId rightScanId;
+      auto op = PlanBuilder(planNodeIdGenerator)
+                    .tableScan(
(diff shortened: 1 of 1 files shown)


velox/exec/MergeJoin.cpp
```diff
@@ -247,7 +247,9 @@ int32_t MergeJoin::compare(
     if (!compare.has_value()) {
       // The SQL semantics of Presto and Spark will always return false if
       // comparing a NULL value with any other value.
-      return -1;
+      // If right side is null, and left side is not null, should return 1
+      // So the following process will catch up rightInput_ with input_
+      return batch->childAt(keys[i])->isNullAt(index) ? -1 : 1;
```
JkSelf: Could we consider changing the null handle mode to kNullAsValue [here](https://github.com/facebookincubator/velox/blob/d8a7a5e0984069785085d607920e20345b8cb3fe/velox/exec/MergeJoin.cpp#L239)? It appears that the `compareNulls  `method already handles this case appropriately: https://github.com/facebookincubator/velox/blob/d8a7a5e0984069785085d607920e20345b8cb3fe/velox/vector/BaseVector.h#L910-L928

Henry2SS: No. 
There is a case: left is null, and right is null. kNullAsValue will return a 0, which means left and right are matched.
But in Join, null doesn't match null.

JkSelf (approved):
Thanks for your fix. LGTM.

Henry2SS:
@JkSelf Need approval to run the CI. Thanks~

Henry2SS:
@JkSelf Hi, is this PR eligible for merging since the CI has passed?

velox/exec/MergeJoin.cpp
```diff
@@ -247,7 +247,9 @@ int32_t MergeJoin::compare(
     if (!compare.has_value()) {
```
xiaoxmeng, line 247: NYC: can you rename batch to leftBatch and otherBatch to rightBatch. The same for others? Thanks!

velox/exec/MergeJoin.cpp
```diff
@@ -247,7 +247,9 @@ int32_t MergeJoin::compare(
...
       // The SQL semantics of Presto and Spark will always return false if
       // comparing a NULL value with any other value.
-      return -1;
+      // If right side is null, and left side is not null, should return 1
```
xiaoxmeng: Can you rephrase the comment with the right semantics?

JkSelf:
@Henry2SS Could you please address the comments from @xiaoxmeng? Once those are resolved, we can proceed with merging this PR. Thanks.

velox/exec/MergeJoin.cpp
Henry2SS: done

velox/exec/MergeJoin.cpp
Henry2SS: Thanks for your suggestions.

Henry2SS:
> @Henry2SS Could you please address the comments from @xiaoxmeng? Once those are resolved, we can proceed with merging this PR. Thanks.

Done. Really appreciate your help.

Henry2SS:
@JkSelf Rebased onto `main` branch. Need approval to run CI. Thanks~

PR #14910, opened 2025-09-19
fix: Allow scientific when casting from decimal to string

boneanxs:
For absolute value less than 1e-6, Spark writes them in scientific notation. 
This PR aligns our behavior with Spark’s formatting(which uses the java
`BigDecimal.toString()`).

Java implementation: https://github.com/openjdk/jdk/blob/a306f88a8456be454f4954c7e4fb8a1273344b5b/src/java.base/share/classes/java/math/BigDecimal.java#L4133

Benchmark:

before:

```c++
cast##cast_decimal_to_inline_string                         8.87ms    112.68
cast##cast_short_decimal                                   12.22ms     81.85
cast##cast_small_decimal                                   14.68ms     68.12
cast##cast_long_decimal                                    30.76ms     32.51
```

after:
```c++
[...]hmarks/ExpressionBenchmarkBuilder.cpp     relative  time/iter   iters/s
cast_decimal_as_varchar##cast_decimal_to_inline             9.00ms    111.07
cast_decimal_as_varchar##cast_short_decimal                11.85ms     84.39
cast_decimal_as_varchar##cast_small_decimal                14.46ms     69.14
cast_decimal_as_varchar##cast_long_decimal                 29.86ms     33.49
```


boneanxs:
@rui-mo @jinchengchenghh @zhli1142015 Hey, could you please help review this? Thanks!

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -229,6 +229,111 @@ class DecimalUtil {
...
+      int32_t maxSize,
+      char* const startPosition) {
+    char* writePosition = startPosition;
+    if (unscaledValue == 0) {
+      *writePosition++ = '0';
+      if (scale > 6) {
+        *writePosition++ = 'E';
+        auto exp =
+            std::to_chars(writePosition, writePosition + maxSize - 2, -scale);
+        VELOX_DCHECK_EQ(
```
zhli1142015: VELOX_CHECK_EQ ?

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -229,6 +229,111 @@ class DecimalUtil {
...
+      if (unscaledValue < 0) {
+        *writePosition++ = '-';
+        unscaledValue = -unscaledValue;
+      }
+      int digits = countDigits(unscaledValue);
+      int32_t adjusted = digits - 1 - scale;
+      if (adjusted < -6) {
+        // Use scientific notation if the absolute value is less than 1e-6.
+        // This is consistent with Spark's behavior.
+        auto coeffBuf = std::unique_ptr<char[]>(new char[digits]);
```
zhli1142015: nit: std::make_unique<char[]>(digits)

velox/functions/sparksql/DecimalUtil.h
zhli1142015: I think you might need to use Status to return errors, just like the API below.
https://github.com/facebookincubator/velox/blob/d8a7a5e0984069785085d607920e20345b8cb3fe/velox/type/DecimalUtil.h#L519

velox/expression/CastHooks.h
```diff
@@ -76,5 +76,34 @@ class CastHooks {
...
   virtual Expected<Timestamp> castBooleanToTimestamp(bool seconds) const = 0;
+
+  /// Converts decimal to string type.
+  template <typename FromNativeType>
+  size_t castFromDecimalToString(
+      FromNativeType unscaledValue,
+      int32_t scale,
+      int32_t maxSize,
+      char* const startPosition) const {
+    if constexpr (std::is_same_v<FromNativeType, int64_t>) {
```
jinchengchenghh: Could we remove if constexpr? FromNativeType type is same with applyCastFromDecimalToString input type

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -229,6 +229,111 @@ class DecimalUtil {
...
+        }
+      }
+    } else {
+      if (unscaledValue < 0) {
+        *writePosition++ = '-';
+        unscaledValue = -unscaledValue;
+      }
+      int digits = countDigits(unscaledValue);
+      int32_t adjusted = digits - 1 - scale;
+      if (adjusted < -6) {
```
jinchengchenghh: Can we do some refactor for existing castToString?

boneanxs: Do you mean adding a new parameter (e.g., sciEnabled) to castToString, and still using CastHook, but passing different values to sparkCastHook/prestoCastHook?

jinchengchenghh: Yes, could you also check the Presto behavior?

rui-mo (commented):
Would you please also add benchmarks for Presto and Spark to ensure the performance? E.g. https://github.com/facebookincubator/velox/blob/main/velox/benchmarks/basic/CastBenchmark.cpp Thanks.

velox/functions/sparksql/DecimalUtil.h
boneanxs: The velox behavior follows the presto:

```sql
> select cast(cast(0.00000001234 as decimal(18, 17)) as varchar);
> 
0.00000001234000000
```
Spark

```sql
spark-sql> select cast(cast(0.00000001234 as decimal(18, 17)) as string);
1.234000000E-8
```

velox/functions/sparksql/DecimalUtil.h
boneanxs: Here I follow the same behavior with the `velox/velox/type/DecimalUtil.h#castToString`

velox/expression/CastHooks.h
```diff
@@ -76,5 +76,29 @@ class CastHooks {
 
   /// Converts boolean to timestamp type.
   virtual Expected<Timestamp> castBooleanToTimestamp(bool seconds) const = 0;
+
+  /// Converts decimal to string type.
+  template <typename FromNativeType>
+  size_t castFromDecimalToString(
```
jinchengchenghh: Could you add the comment to describe the return value?

velox/type/DecimalUtil.h
```diff
@@ -341,11 +341,22 @@ class DecimalUtil {
       T unscaledValue,
       int32_t scale,
       int32_t maxSize,
-      char* const startPosition) {
+      char* const startPosition,
+      bool sci = false) {
```
jinchengchenghh: please describe the argument and renamed to isScientific

velox/type/DecimalUtil.h
```diff
@@ -356,6 +367,41 @@ class DecimalUtil {
         *writePosition++ = '-';
         unscaledValue = -unscaledValue;
       }
+      int32_t adjusted = 0;
```
jinchengchenghh: Move adjusted to auto adjusted = digits - 1 - scale;

velox/type/DecimalUtil.h
jinchengchenghh: And update the function comment

velox/type/DecimalUtil.h
```diff
@@ -356,6 +367,41 @@ class DecimalUtil {
...
+          auto coeffBuf = std::make_unique<char[]>(digits);
+          auto [coeffEnd, coeffEc] = std::to_chars(
+              coeffBuf.get(), coeffBuf.get() + digits, unscaledValue);
+          VELOX_DCHECK_EQ(
+              coeffEc,
+              std::errc(),
+              "Failed to cast coeff to varchar: {}",
+              std::make_error_code(coeffEc).message());
+          int coeffLen = coeffEnd - coeffBuf.get();
+          *writePosition++ = coeffBuf[0];
```
jinchengchenghh: The maxSize does not consider scientific case, is it possible the actual size exceeds maxSize?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +367,41 @@ class DecimalUtil {
...
+          // This is consistent with Spark's behavior.
+          auto coeffBuf = std::make_unique<char[]>(digits);
+          auto [coeffEnd, coeffEc] = std::to_chars(
+              coeffBuf.get(), coeffBuf.get() + digits, unscaledValue);
+          VELOX_DCHECK_EQ(
+              coeffEc,
+              std::errc(),
+              "Failed to cast coeff to varchar: {}",
+              std::make_error_code(coeffEc).message());
+          int coeffLen = coeffEnd - coeffBuf.get();
```
jinchengchenghh: Is the coeffLen equal to digits?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +367,41 @@ class DecimalUtil {
...
         unscaledValue = -unscaledValue;
       }
+      int32_t adjusted = 0;
+      if (sci) {
+        int digits = countDigits(unscaledValue);
+        adjusted = digits - 1 - scale;
+        if (adjusted < -6) {
+          // Use scientific notation if the absolute value is less than 1e-6.
+          // This is consistent with Spark's behavior.
+          auto coeffBuf = std::make_unique<char[]>(digits);
```
jinchengchenghh: Please don't introduce the temp buffer, maybe you can check the digits > 1, and write to writePosition

boneanxs: If we Avoid the temporary buffer, we requires `std::memmove` to safely handle the overlapping memory regions when inserting the decimal point(`.`). it might be slower comparing to the extra temp buffer, especially coeffBuf is smaller in most cases?

velox/type/DecimalUtil.h
boneanxs: shouldn't be, the scientifc case is more compact, it only triggered for small magnitude decimals(absolute value < 1e-6).

Since maxStringViewSize(p, s) is calculated to fit the full decimal representation (which includes all the leading zeros), it will always have sufficient space for the much shorter scientific notation format.

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -192,6 +192,68 @@ TEST_F(SparkCastExprTest, decimalToIntegral) {
   testDecimalToIntegralCasts<int8_t>();
 }
 
+TEST_F(SparkCastExprTest, decimalToVarchar) {
+  auto expect64 = [&](const std::vector<std::optional<int64_t>>& values,
```
jinchengchenghh: Could you use like other tests to test?
```
    testCast(
        shortFlat,
        makeNullableFlatVector<T>(
            {-3,
             -2 /*-2.6 truncated to -2*/,
             -2 /*-2.3 truncated to -2*/,
             -2,
             -1,
             0,
             55,
             57 /*57.49 truncated to 57*/,
             57 /*57.55 truncated to 57*/,
             69,
             72,
             std::nullopt}));
```

velox/type/DecimalUtil.h
```diff
@@ -356,6 +370,39 @@ class DecimalUtil {
         *writePosition++ = '-';
         unscaledValue = -unscaledValue;
       }
+      if (isScientific) {
```
jinchengchenghh, line 379: Could you use unscaledValue < xxx to check if the number should be scientific format, then we would not need countDigits for other common normal numbers
Can `unscaledValue < (1e-6 * DecimalUtil::kPowersOfTen[scale]) )` cover the case?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +370,39 @@ class DecimalUtil {
...
+        const auto digits = countDigits(unscaledValue);
+        const auto adjusted = digits - 1 - scale;
+        if (adjusted < -6) {
+          // Use scientific notation if the absolute value is less than 1e-6.
+          // This is consistent with Spark's behavior.
+          auto coeffBuf = std::make_unique<char[]>(digits);
```
jinchengchenghh: `auto coeffBuf = std::vector<char>(digits)`

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -192,6 +192,68 @@ TEST_F(SparkCastExprTest, decimalToIntegral) {
...
+  expect64(
+      {123456789, -333333333, 0, 5, -9, std::nullopt},
+      DECIMAL(9, 2),
+      {"1234567.89", "-3333333.33", "0.00", "0.05", "-0.09", std::nullopt});
+
+  // Short decimal zero scale.
+  expect64({0}, DECIMAL(6, 0), {"0"});
+
+  // Short decimal extreme values (scientific notation for small magnitudes).
+  expect64(
```
jinchengchenghh: Please also add the long decimal scientific notation

velox/functions/sparksql/tests/SparkCastExprTest.cpp
boneanxs: Using the lambda helper function can reduce the same comparing codes

jinchengchenghh: rename to testShortDecimal

velox/type/DecimalUtil.h
jinchengchenghh: Please update the benchmark result, with this change, the benchmark result should be similar.

jinchengchenghh (approved):
Overall looks good, only some small nits

velox/type/DecimalUtil.h
boneanxs: done, updated to the description, It appeals a small improvement 👍

velox/functions/sparksql/tests/SparkCastExprTest.cpp
boneanxs: done

boneanxs:
@rui-mo @zhli1142015 Hi, any further comments? Thanks!

velox/expression/CastHooks.h
```diff
@@ -76,5 +76,30 @@ class CastHooks {
...
+      int64_t unscaledValue,
+      int32_t scale,
+      int32_t maxSize,
+      char* const startPosition) const = 0;
+
+  virtual size_t applyCastFromDecimalToString(
+      int128_t unscaledValue,
+      int32_t scale,
+      int32_t maxSize,
+      char* const startPosition) const = 0;
```
rui-mo: It appears that the only difference between Presto and Spark is the value of isScientific. Given that, could we define a virtual method like bool `isScientific()` and call it during cast? This would allow us to reuse most of the existing code while introducing minimal additional API.

velox/type/DecimalUtil.h
```diff
@@ -335,17 +335,31 @@ class DecimalUtil {
   /// @param scale The scale of decimal.
   /// @param maxSize The estimated max size of string.
   /// @param startPosition The start position to write from.
-  /// @return The actual size of the string.
+  /// @param isScientific Whether to format small magnitude decimals using
+  /// scientific notation (Spark-compatible).
```
rui-mo: Would you please provide an example in the comment?

velox/type/DecimalUtil.h
```diff
@@ -335,17 +335,34 @@ class DecimalUtil {
...
       int32_t scale,
       int32_t maxSize,
-      char* const startPosition) {
+      char* const startPosition,
+      bool isScientific = false) {
     char* writePosition = startPosition;
     if (unscaledValue == 0) {
       *writePosition++ = '0';
-      if (scale > 0) {
+      if (isScientific && scale > 6) {
```
rui-mo: nit: Could we define a const for the value `6` to improve maintainability?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +373,40 @@ class DecimalUtil {
         *writePosition++ = '-';
         unscaledValue = -unscaledValue;
       }
+      if (isScientific) {
+        if (scale >= 6 &&
+            unscaledValue < DecimalUtil::kPowersOfTen[scale - 6]) {
+          const auto digits = countDigits(unscaledValue);
+          const auto adjusted = digits - 1 - scale;
```
rui-mo: nit: perhaps move `adjusted` near its usage.

velox/type/DecimalUtil.h
```diff
@@ -356,6 +373,40 @@ class DecimalUtil {
...
+          // Use scientific notation if the absolute value is less than 1e-6.
+          // This is consistent with Spark's behavior.
+          auto coeffBuf = std::vector<char>(digits);
```
rui-mo: nit: does `coeff` mean `coefficient`?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +373,40 @@ class DecimalUtil {
...
+          VELOX_DCHECK_EQ(
+              coeffEc,
+              std::errc(),
+              "Failed to cast coeff to varchar: {}",
+              std::make_error_code(coeffEc).message());
+          *writePosition++ = coeffBuf[0];
+          if (digits > 1) {
+            *writePosition++ = '.';
+            size_t toCopy = digits - 1;
+            std::memcpy(writePosition, coeffBuf.data() + 1, toCopy);
```
rui-mo: Why do we write to an intermediate buffer first and then memcpy into `writePosition`? Can we write directly to `writePosition` instead?

rui-mo:
Hi @boneanxs, please also update the relevant documentation after https://github.com/facebookincubator/velox/pull/16124, thanks!

velox/type/DecimalUtil.h
boneanxs: The intermediate buffer is required since https://github.com/facebookincubator/velox/pull/14910#discussion_r2642748944

rui-mo: I see. Another possible approach might be to write the first character to `writePosition`, then determine whether to insert a `.`, followed by the remaining portion of the unscaled value. This is just an initial idea, though. I haven’t evaluated how feasible it is yet.

rui-mo: And this won’t block approval of the PR once the other minor comments are addressed.

boneanxs: > I see. Another possible approach might be to write the first character to `writePosition`, then determine whether to insert a `.`, followed by the remaining portion of the unscaled value. This is just an initial idea, though. I haven’t evaluated how feasible it is yet.

We can do so, but it appeals introducing some extra division/modulo operations, If I understand correctly, we need to:
1. calculate the leading digit by unscaledValue / 10^(digits-1)
2. calculate the remainder by `unscaledValue % divisor`
3. calculate the leading zero after the `.` by check the digits of remainder `countDigits(remainder)`

Comparing with the temporary buffer handling(which usually is a small buffer, it should be fast), not sure the new way is more performant

boneanxs:
> Hi @boneanxs, please also update the relevant documentation after #16124, thanks!

done

rui-mo:
> [ RUN      ] SparkCastExprTest.decimalToString
/__w/velox/velox/velox/vector/tests/utils/VectorTestBase.cpp:153: Failure
Value of: expected->equalValueAt(actual.get(), i, i)
  Actual: false
Expected: true
at 3: expected 0.0000000000, but got 0E-10

@boneanxs There's an UT failure. Can you please take a look? Thanks.

velox/type/DecimalUtil.h
```diff
@@ -356,6 +376,43 @@ class DecimalUtil {
...
       }
+      if (isScientific) {
+        if (scale >= kMinScientificNotationScale &&
+            unscaledValue < DecimalUtil::kPowersOfTen
+                                [scale - kMinScientificNotationScale]) {
+          // Use scientific notation if the absolute value is less than 1e-6.
+          // This is consistent with Spark's behavior.
+          const auto digits = countDigits(unscaledValue);
+          auto coefficientBuf = std::vector<char>(digits);
+          auto [coefficientEnd, coefficientEc] = std::to_chars(
```
kgpai: @boneanxs Apologies for the delay on this, our builds fail due to unused variable: 

```
 unused variable '[coefficientEnd, coefficientEc]' [-Werror,-Wunused-variable]
  387 |           auto [coefficientEnd, coefficientEc] = std::to_chars(
```

Can you fix these problems ?

boneanxs: done

kgpai:
@boneanxs Can you rebase against latest main ?

velox/type/DecimalUtil.h
```diff
@@ -356,6 +376,43 @@ class DecimalUtil {
...
+                                [scale - kMinScientificNotationScale]) {
+          // Use scientific notation if the absolute value is less than 1e-6.
+          // This is consistent with Spark's behavior.
+          const auto digits = countDigits(unscaledValue);
+          auto coefficientBuf = std::vector<char>(digits);
+          auto coefficient = std::to_chars(
+              coefficientBuf.data(),
+              coefficientBuf.data() + digits,
+              unscaledValue);
+          VELOX_DCHECK_EQ(
```
kgpai, line 391: @boneanxs coefficient is only used for dcheck, and fails in optimized build; Can you refactor so that the whole statement is part of DCHECK ?

boneanxs: done

rui-mo:
@boneanxs There arises an UT failure. Can you please take a look?

```
SparkCastExprTest.decimalToString
/__w/velox/velox/velox/vector/tests/utils/VectorTestBase.cpp:153: Failure
Value of: expected->equalValueAt(actual.get(), i, i)
  Actual: false
Expected: true
at 4: expected 5.5E-9, but got \0.\0E-9
```

kgpai:
@boneanxs Please `@' me when you think this is ready to merge.

boneanxs:
> @boneanxs Please `@' me when you think this is ready to merge.

@kgpai done, please take a look

zhouyuan:
@boneanxs thanks for the patch, it seems the behavior changed in recent spark: if ansi is on then it will use plain string
https://github.com/apache/spark/pull/37221/changes

PR #14913, opened 2025-09-19
misc: Enable implicit conversion from velox::StringView to std::string_view

kKPulla:
Summary: This is aligned with the conversion to folly::StringPiece that we allow to be done implicitly. Given std::string_view is also a view type, we should keep it consistent.

Differential Revision: D82840490


MBkkt (commented):
I think we are also need to remove cast to std::string.
Instead constructor for std::string should be used (implicit cast to string_view will be used)

Otherwise see errors in github CI

kKPulla:
> I think we are also need to remove cast to std::string. Instead constructor for std::string should be used (implicit cast to string_view will be used)
> 
> Otherwise see errors in github CI

@MBkkt Yeah the CI failure can be fixed by explicitly constructing std::string but I agree with you on removing implicit cast to std::string and initially that was in scope. But I did not include it here as it is not backwards compatible change as today we allow velox::StringView -> std::string implicit cast. 

I did not get a chance to get back to this, will address and update. Thanks for reviewing.

PR #14917, opened 2025-09-19
feat: Upgrade simdjson to 13.3.0

czentgr:
simdjson needs to be upgraded to work with Clang 20 and C++20 support.


assignUser:
So this doesn't require code changes, so is backwards compatible? That's nice, so we don't have to wait for meta to use this version internaly?

czentgr:
@assignUser That's right. No code changes, just one change in an error exception message seen in the tests.

kgpai:
@czentgr This change requires a separate change intnerally where we have to update simdjson, so give me some time for that. I will keep this thread posted on how that goes.

assignUser:
@kgpai As I understand it this is backwards compatible to the previous simdjson version, so it should be fine for us to use this version before the internal change lands as long as we don't introduce any use of backwards incompatible apis?

kgpai:
@assignUser Yes you are right, no changes in api, so we shouldnt need to upgrade anything internally.

velox/functions/prestosql/tests/JsonFunctionsTest.cpp
```diff
@@ -313,7 +313,9 @@ TEST_F(JsonFunctionsTest, jsonParse) {
       jsonParse(R"({"k1":})"), "The JSON document has an improper structure");
   VELOX_ASSERT_THROW(
       jsonParse(R"({:"k1"})"), "The JSON document has an improper structure");
-  VELOX_ASSERT_THROW(jsonParse(R"(not_json)"), "Problem while parsing an atom");
+  VELOX_ASSERT_THROW(
```
kgpai: @assignUser This test case fails internally though since the error messages are different.

kgpai: @czentgr Shall we just ensure that it throws and not check the error message ?

czentgr: @kgpai Yes, we can. Usually, the macro is used to make sure the correct exception message (if generated by Velox) is thrown. In this case, it is a third party message so can't really rely on the details. I can update it to just check for failure via ASSERT_THROW.

PR #14918, opened 2025-09-20
feat: Allow transferring memory to a given memory pool

kagamiori:
Summary:
This diff adds a new API
`bool MemoryPool::transferTo(MemoryPool* dest, void* buffer, int64_t size)`
that allows transferring the ownership of the memory starting from `buffer` of size 
`size` to the `dest` pool. This API requires the calling pool and the destination pool 
are both leaf pools and use the same allocator. If the requirements are satisfied, this 
API transfers the ownership and return true. If not, this API does nothing and return 
false.

Differential Revision: D82877273

### DIFF
--- a/velox/common/memory/MemoryPool.cpp
+++ b/velox/common/memory/MemoryPool.cpp
@@ -18,6 +18,7 @@
 
 #include <signal.h>
 
+#include "velox/common/Casts.h"
 #include "velox/common/base/Counters.h"
 #include "velox/common/base/StatsReporter.h"
 #include "velox/common/base/SuccinctPrinter.h"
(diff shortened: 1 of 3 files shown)


velox/common/memory/MemoryPool.cpp
```diff
@@ -583,18 +584,41 @@ void* MemoryPoolImpl::reallocate(void* p, int64_t size, int64_t newSize) {
...
   release(alignedSize);
 }
 
+bool MemoryPoolImpl::transferTo(MemoryPool* dest, void* buffer, int64_t size) {
+  if (!isLeaf() || !dest->isLeaf()) {
+    return false;
+  }
+  VELOX_CHECK_NOT_NULL(dest);
+  auto* destImpl = checked_pointer_cast<MemoryPoolImpl, MemoryPool>(dest);
+  VELOX_CHECK_NOT_NULL(destImpl);
```
xiaoxmeng: Remove VELOX_CHECK_NOT_NULL as checked_pointer_cast does the nullable check

PR #14359, opened 2025-08-07
feat(cudf): Large string support for importFromArrow

karthikeyann:
Arrow bridge in velox does not support large string conversion to string vector with 64 bit offsets in Velox.
This PR adds large strings support to importFromArrow

- importFromArrow large string support
- unit tests added for importing as Viewer and Owner.

Context: cuDF supports large string column. Exporting it to arrow and interop with Velox APIs required large string support via Arrow Bridge. So, large string support is added for importFromArrow.

PR #14921, opened 2025-09-22
fix: Fix Arrow larget strig import may overflow

jinchengchenghh:
After this PR https://github.com/facebookincubator/velox/pull/14359, Arrow supports large string data type, but the length data type is int64_t, it may overflow Velox StringView supports length int32_t, causing undefined behavior.

Further more, if velox should support large string, if user use LargeString data type, they may request the string size more that int32_t range.

### DIFF
--- a/velox/vector/arrow/Bridge.cpp
+++ b/velox/vector/arrow/Bridge.cpp
@@ -621,6 +621,16 @@ VectorPtr createStringFlatVector(
   bool shouldAcquireStringBuffer = false;
 
   for (size_t i = 0; i < length; ++i) {
+    if constexpr (!std::is_same_v<TOffset, int32_t>) {
+      auto rowBytes = offsets[i + 1] - offsets[i];
+      VELOX_CHECK(
+          rowBytes >= std::numeric_limits<int32_t>::min() &&
+          rowBytes <= std::numeric_limits<int32_t>::max(),
+          "Offset difference (rowBytes = {}) out of int32_t range at index {}",
+          rowBytes,
+          i);
+    }
+
     rawStringViews[i] =
         StringView(values + offsets[i], offsets[i + 1] - offsets[i]);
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
Could you help review? Thanks! @pedroerp

jinchengchenghh:
If we create a large string with size more than max int32, GCC/libstdc++ and Clang/libc++ can handle it, but debug mode or small heap limits might fail. This test will be a high memory load test, would you like me to add? @pedroerp

jinchengchenghh:
> If we create a large string with size more than max int32, GCC/libstdc++ and Clang/libc++ can handle it, but debug mode or small heap limits might fail. This test will be a high memory load test, would you like me to add? @pedroerp

Could you help take a look?  Thanks! @pedroerp

pedroerp (approved):
@jinchengchenghh fair enough. I also can't think of a good way to test this; and this seems safe and simple enough. Thanks

velox/vector/arrow/Bridge.cpp
```diff
@@ -621,8 +621,16 @@ VectorPtr createStringFlatVector(
   bool shouldAcquireStringBuffer = false;
 
   for (size_t i = 0; i < length; ++i) {
-    rawStringViews[i] =
-        StringView(values + offsets[i], offsets[i + 1] - offsets[i]);
+    const auto rowBytes = offsets[i + 1] - offsets[i];
+    if constexpr (!std::is_same_v<TOffset, int32_t>) {
+      VELOX_CHECK(
+          rowBytes >= 0 && rowBytes <= std::numeric_limits<int32_t>::max(),
```
mbasmanova, line 656: split into 2 separate checks to produce better error messages

mbasmanova:
@jinchengchenghh Would you also rebase?

PR #14927, opened 2025-09-22
fix(abfs): Fix AzureClientProviderFactory registration issue

xin-zhang2:
Register the AzureClientProvider when creating ABFS file system to prevent the "No AzureClientProviderFactory registered" error if `registerAzureClientProvider` is not explicitly called.

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -939,9 +939,6 @@ These semantics are similar to the `Apache Hadoop-Aws module <https://hadoop.apa
        "SharedKey": Uses the storage account name and key for authentication.
        "OAuth": Utilizes OAuth tokens for secure authentication.
        "SAS": Employs Shared Access Signatures for granular access control.
-       To create Azure clients with the configured authentication type, the caller must
-       register the corresponding Azure client provider from the configuration by calling
-       `registerAzureClientProvider`.
    * - fs.azure.account.key.<storage-account>.dfs.core.windows.net
      - string
      -


xin-zhang2:
@czentgr @majetideepak 
After further consideration, I think a singleton may not be suitable for the client provider of each auth type. This is because each ReadFile or WriteFile creates its own client, and the provider's memebers are initialized using a ReadFile/WriteFile-specifc path before the client is created. A singleton could therefore lead to potential data races.

For example, in `OAuthAzureClientProvider::getReadFileClient`, member `tokenCredential_` is used to create a `BlobClient` instance, while it is initialized in `OAuthAzureClientProvider::init` which depends on the `abfsPath` passed to it.
```
std::unique_ptr<AzureBlobClient> OAuthAzureClientProvider::getReadFileClient(
    const std::shared_ptr<AbfsPath>& abfsPath,
    const config::ConfigBase& config) {
  init(abfsPath, config);
  const auto url = abfsPath->getUrl(true);
  auto client = std::make_unique<BlobClient>(url, tokenCredential_);
  return std::make_unique<BlobClientWrapper>(std::move(client));
}

void OAuthAzureClientProvider::init(
    const std::shared_ptr<AbfsPath>& abfsPath,
    const config::ConfigBase& config) {
  auto clientIdKey = fmt::format(
      "{}.{}", kAzureAccountOAuth2ClientId, abfsPath->accountNameWithSuffix());
  auto clientSecretKey = fmt::format(
      "{}.{}",
      kAzureAccountOAuth2ClientSecret,
      abfsPath->accountNameWithSuffix());
  auto clientEndpointKey = fmt::format(
      "{}.{}",
      kAzureAccountOAuth2ClientEndpoint,
      abfsPath->accountNameWithSuffix());
  VELOX_USER_CHECK(
      config.valueExists(clientIdKey), "Config {} not found", clientIdKey);
  VELOX_USER_CHECK(
      config.valueExists(clientSecretKey),
      "Config {} not found",
      clientSecretKey);
  VELOX_USER_CHECK(
      config.valueExists(clientEndpointKey),
      "Config {} not found",
      clientEndpointKey);
  auto clientEndpoint = config.get<std::string>(clientEndpointKey).value();
  // Length of "https://".
  static const std::size_t kHttpsPrefixLen = 8;
  auto firstSep = clientEndpoint.find_first_of("/", kHttpsPrefixLen);
  authorityHost_ = clientEndpoint.substr(0, firstSep + 1);
  auto sedondSep = clientEndpoint.find_first_of("/", firstSep + 1);
  tenentId_ = clientEndpoint.substr(firstSep + 1, sedondSep - firstSep - 1);
  Azure::Identity::ClientSecretCredentialOptions options;
  options.AuthorityHost = authorityHost_;
  tokenCredential_ = std::make_shared<Azure::Identity::ClientSecretCredential>(
      tenentId_,
      config.get<std::string>(clientIdKey).value(),
      config.get<std::string>(clientSecretKey).value(),
      options);
}
```

What do you think?

velox/connectors/hive/storage_adapters/abfs/AbfsFileSystem.cpp
```diff
@@ -25,12 +25,14 @@
...
 #include "velox/connectors/hive/storage_adapters/abfs/AbfsWriteFile.h"
 #include "velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.h"
+#include "velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.h"
 
 namespace facebook::velox::filesystems {
 
 AbfsFileSystem::AbfsFileSystem(std::shared_ptr<const config::ConfigBase> config)
     : FileSystem(config) {
   VELOX_CHECK_NOT_NULL(config.get());
+  registerAzureClientProvider(*config_);
```
czentgr, line 35: The issue could be that this happens last - at the split reception.
Meanwhile a user might have registered a custom handler but for the same account that is passed through the config. If this happens to match one of the registered accounts they get replaced here assuming the auth type matches one of the previous providers. It basically means the auth type for a custom provider should not match one of the supported auth types.

Perhaps we need to check on the registration? But we don't know when this is called and when it is custom vs built-in.

This also changes the behavior and we need to update the documentation accordingly (see the additions made with the refactor.)

czentgr: Please also fix the type in the abstract and commit message "registeration".

xin-zhang2: Typo fixed.

Yes, both custom providers and built-in providers are registered via `AzureClientProviderFactories::registerFactory`, which will overwrite the previous registered provider for the same account.
https://github.com/facebookincubator/velox/blob/95787796185321cba2c6a2808bacb91a0a443acd/velox/connectors/hive/storage_adapters/abfs/AzureClientProviderFactories.h#L35-L40

I‘m not sure whether we need to define a priority for custom and built-in providers - for example, only checking the built-in ones when no custom provider is present? 

I removed that addition from the doc.

xin-zhang2:
@marin-ma 
I noticed the comment https://github.com/facebookincubator/velox/pull/14281#discussion_r2269441981 as well and discussed it with @czentgr and @majetideepak. We think it might be more reasonable to handle the registration within velox, especially since both Presto and Gluten are affected by this change. Given that the hive config is also passed to `abfsFileSystemGenerator` during the initialization of the `AbfsFileSystem` singleton, why not register it there? 
One possible reason I can think of for moving the registration to the caller side is to allow the caller to decide whether to create a provider from the config or use a custom provider for a specific account, but currently both Gluten and Presto register the providers from the config. Are there any other considerations?

velox/connectors/hive/storage_adapters/abfs/AbfsFileSystem.cpp
czentgr: This is the multi abfs instance support. We still need to register the client factories at some point if they don't exist for this account/auth type.
See https://github.com/facebookincubator/velox/pull/14598

PR #14928, opened 2025-09-23
fix(build): Bump Hadoop version for more reliable download

simoneves:
The archive.apache.org download server is very slow and unreliable. The existing download can take around an hour, if it doesn't fail completely.

This PR bumps the version to 3.3.6 which is available on the main dlcdn.apache.org download server which is MUCH faster and more reliable. This download takes approximately 30 seconds from an NV system.

The newer version will have three years of bug fixes also (3.3.0 was released in June 2020, 3.3.6 in June 2023).

### DIFF
--- a/scripts/setup-common.sh
+++ b/scripts/setup-common.sh
@@ -393,7 +393,7 @@ function install_azure-storage-sdk-cpp {
 
 function install_hdfs_deps {
   # Dependencies for Hadoop testing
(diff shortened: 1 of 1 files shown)


assignUser:
Actually this is breaking docker builds for most PRs modifying the images or setup scripts so let's merge it now and revert/change later if necessary.

mbasmanova:
This change seems to have broken CI: #15093

PR #14929, opened 2025-09-23
feat: Add reading flatmap as struct feature in TableEvolutionFuzzer

juwentus1234:
Summary: For columns which is Map and eligible to be writen as flatmap, implement the feature of reading it as struct.

Differential Revision: D83015746

### DIFF
--- a/velox/exec/tests/TableEvolutionFuzzer.cpp
+++ b/velox/exec/tests/TableEvolutionFuzzer.cpp
@@ -70,6 +70,14 @@ VectorFuzzer::Options makeVectorFuzzerOptions() {
   return options;
 }
 
+template <typename T>
+void removeFromVector(std::vector<T>& vec, const T& value) {
+  auto it = std::find(vec.begin(), vec.end(), value);
+  if (it != vec.end()) {
+    vec.erase(it);
+  }
+}
+
 bool hasUnsupportedMapKey(const TypePtr& type) {
   switch (type->kind()) {
     case TypeKind::MAP: {
@@ -219,6 +227,7 @@ std::vector<std::vector<RowVectorPtr>> runTaskCursors(
   }
   std::vector<std::vector<RowVectorPtr>> results;
   constexpr std::chrono::seconds kTaskTimeout(10);
+  results.reserve(futures.size());
   for (auto& future : futures) {
     results.push_back(std::move(future).get(kTaskTimeout));
   }
@@ -559,12 +568,17 @@ void TableEvolutionFuzzer::run() {
       2 * config_.evolutionCount - 1);
   RowVectorPtr finalExpectedData;
 
+  folly::F14FastMap<int, folly::F14FastSet<std::string>> globalMapColumnKeys;
+  std::vector<int> globallyConsistentColumnIndexVector;
+
   createWriteTasks(
       testSetups,
       bucketColumnIndices,
       tableOutputRootDir->getPath(),
       writeTasks,
-      finalExpectedData);
+      finalExpectedData,
+      globalMapColumnKeys,
+      globallyConsistentColumnIndexVector);
 
   auto executor = folly::getGlobalCPUExecutor();
   auto writeResults = runTaskCursors(writeTasks, *executor);
@@ -618,10 +632,20 @@ void TableEvolutionFuzzer::run() {
   }
 
   std::vector<std::shared_ptr<TaskCursor>> scanTasks(2);
-  scanTasks[0] =
-      makeScanTask(rowType, std::move(actualSplits), pushownConfig, false);
-  scanTasks[1] =
-      makeScanTask(rowType, std::move(expectedSplits), pushownConfig, true);
+  scanTasks[0] = makeScanTask(
+      rowType,
+      std::move(actualSplits),
+      pushownConfig,
+      false,
+      globalMapColumnKeys,
+      globallyConsistentColumnIndexVector);
+  scanTasks[1] = makeScanTask(
+      rowType,
+      std::move(expectedSplits),
+      pushownConfig,
+      true,
+      globalMapColumnKeys,
+      globallyConsistentColumnIndexVector);
 
   ScopedOOMInjector oomInjectorReadPath(
       [this]() -> bool { return folly::Random::oneIn(10, rng_); },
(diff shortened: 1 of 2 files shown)


velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -820,8 +838,69 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
           // %50 chance to enable flatmap for this map column.
           if (enableFlatMap && folly::Random::oneIn(2, rng)) {
             supportedMapColumnIndices.push_back(static_cast<uint32_t>(i));
-            VLOG(1) << "Write column " << setup.schema->nameOf(i)
```
Yuhta, line 823: Let's keep this log to know which columns in which setups are written as flatmaps

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -820,8 +838,69 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
...
-                    << " as flatmap";
+
+            // Extract actual keys from the map data
+            auto mapVector = data->childAt(i)->asChecked<MapVector>();
```
Yuhta: Use `DecodedVector` to avoid cast, fuzzer might generate encoded data in the future

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -820,8 +838,69 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
...
+              std::vector<std::string> keyVector;
+              auto keys = mapVector->mapKeys();
+
+              if (keys) {
+                // Use a set to collect unique keys
+                std::set<std::string> uniqueKeys;
+
+                // Convert all key types to string
+                for (vector_size_t j = 0; j < keys->size(); ++j) {
+                  if (!keys->isNullAt(j)) {
```
Yuhta: Let's also check the map row itself is not null

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -820,8 +838,69 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
...
+                      auto* keyVector = keys->asFlatVector<int16_t>();
+                      auto keyVal = keyVector->valueAt(j);
+                      keyStr = std::to_string(keyVal);
+                    } else if (keys->type()->isTinyint()) {
+                      auto* keyVector = keys->asFlatVector<int8_t>();
+                      auto keyVal = keyVector->valueAt(j);
+                      keyStr = std::to_string(keyVal);
+                    } else {
+                      // For other types, use toString
+                      keyStr = keys->toString(j);
```
Yuhta: Just throw for other types since they are not supported for flatmap

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -559,12 +560,17 @@ void TableEvolutionFuzzer::run() {
       2 * config_.evolutionCount - 1);
   RowVectorPtr finalExpectedData;
 
+  std::unordered_map<int, std::vector<std::string>> globalMapColumnKeys;
```
Yuhta: Just use `F14FastSet` as values and insert into it directly?

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -986,16 +1120,87 @@ void TableEvolutionFuzzer::createWriteTasks(
...
+          }
+        }
+      }
+    }
+    eachStepLocalSupportedMapColumns[i] = localSupportedMapColumns;
+  }
+
+  // Find globally consistent columns: only columns that can write as flatmap in
+  // ALL evolution steps
+  if (config_.evolutionCount == 0) {
```
Yuhta: Is this always greater than 0?

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -986,16 +1120,87 @@ void TableEvolutionFuzzer::createWriteTasks(
...
+        auto& stepColumns = eachStepLocalSupportedMapColumns[stepIndex];
+        if (std::find(stepColumns.begin(), stepColumns.end(), columnIndex) ==
+            stepColumns.end()) {
+          appearsInAllSteps = false;
+          break;
+        }
+      }
+
+      if (appearsInAllSteps) {
+        globallyConsistentColumnIndexVector.push_back(columnIndex);
```
Yuhta: We don't need to be so accurate for the valid column indices before writing.  Just pass in all columns and remove the ones that are not flatmaps inside `makeWriteTask`.

juwentus1234: I think this is needed as it filters out those only appears in some steps evolution but not in all, more safe and easy to debug logic.

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -822,7 +849,86 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeWriteTask(
...
+            SelectivityVector allRows(data->childAt(i)->size());
+            DecodedVector decodedMap(*data->childAt(i), allRows);
+            auto* mapVector = decodedMap.base()->asChecked<MapVector>();
+            if (mapVector->size() > 0) {
+              std::vector<std::string> keyVector;
+              auto keys = mapVector->mapKeys();
+
+              if (keys) {
+                // Use a set to collect unique keys
+                std::set<std::string> uniqueKeys;
```
Yuhta: How about making `globalMapColumnKeys` as `folly::F14FastMap<int, folly:F14FastSet<std::string>>` and insert into it directly?

juwentus1234: updated

Yuhta: Just do `auto& uniqueKeys = globalMapColumnKeys[i]` and you don't need to insert it twice

PR #14930, opened 2025-09-23
refactor(abfs): Extract the logic to get the accountName and authType from the config into common method

JkSelf:
We also need to include both `accountName `and `authType ` in the cache key to support multiple ABFS FileSystem instances. Therefore, this PR extracts the related code into a common method.

### DIFF
--- a/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
+++ b/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h"
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+
+namespace facebook::velox::filesystems {
+
+std::vector<CacheKey> extractCacheKeyFromConfig(
+    const config::ConfigBase& config) {
+  std::vector<CacheKey> cacheKeys;
+  for (const auto& [key, value] : config.rawConfigs()) {
+    constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
+    if (key.find(authTypePrefix) == 0) {
+      std::string_view skey = key;
+      // Extract the accountName after "fs.azure.account.auth.type.".
+      auto remaining = skey.substr(authTypePrefix.size() + 1);
+      auto dot = remaining.find(".");
+      VELOX_USER_CHECK_NE(
+          dot,
+          std::string_view::npos,
+          "Invalid Azure account auth type key: {}",
+          key);
+      auto accountName = std::string(remaining.substr(0, dot));
+      cacheKeys.emplace_back(CacheKey{accountName, value});
+    }
+  }
+  return cacheKeys;
+}
+
+} // namespace facebook::velox::filesystems
--- a/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
+++ b/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
@@ -26,17 +26,15 @@ std::vector<CacheKey> extractCacheKeyFromConfig(
   constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
   for (const auto& [key, value] : config.rawConfigs()) {
     if (key.find(authTypePrefix) == 0) {
-      std::string_view skey = key;
       // Extract the accountName after "fs.azure.account.auth.type.".
-      auto remaining = skey.substr(authTypePrefix.size() + 1);
+      auto remaining = std::string_view(key).substr(authTypePrefix.size() + 1);
       auto dot = remaining.find(".");
       VELOX_USER_CHECK_NE(
           dot,
           std::string_view::npos,
           "Invalid Azure account auth type key: {}",
           key);
-      std::string_view accountName = std::string(remaining.substr(0, dot));
-      cacheKeys.emplace_back(CacheKey{accountName, value});
+      cacheKeys.emplace_back(CacheKey{remaining.substr(0, dot), value});
     }
   }
   return cacheKeys;
--- a/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
+++ b/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
@@ -23,8 +23,8 @@ namespace facebook::velox::filesystems {
 std::vector<CacheKey> extractCacheKeyFromConfig(
     const config::ConfigBase& config) {
   std::vector<CacheKey> cacheKeys;
+  constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
   for (const auto& [key, value] : config.rawConfigs()) {
-    constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
     if (key.find(authTypePrefix) == 0) {
       std::string_view skey = key;
       // Extract the accountName after "fs.azure.account.auth.type.".
@@ -35,7 +35,7 @@ std::vector<CacheKey> extractCacheKeyFromConfig(
           std::string_view::npos,
           "Invalid Azure account auth type key: {}",
           key);
-      auto accountName = std::string(remaining.substr(0, dot));
+      std::string_view accountName = std::string(remaining.substr(0, dot));
       cacheKeys.emplace_back(CacheKey{accountName, value});
     }
   }
--- a/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
+++ b/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
@@ -18,15 +18,20 @@
 
 #include <azure/storage/common/storage_exception.hpp>
 #include <fmt/format.h>
-#include "velox/common/config/Config.h"
 #include "velox/common/file/File.h"
-#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
 
 namespace facebook::velox::filesystems {
 namespace {
 constexpr std::string_view kAbfsScheme{"abfs://"};
 constexpr std::string_view kAbfssScheme{"abfss://"};
 } // namespace
+
+class ConfigBase;
+
+struct CacheKey {
+  std::string accountName;
+  std::string authType;
+};
 
 inline bool isAbfsFile(const std::string_view filename) {
   return filename.find(kAbfsScheme) == 0 || filename.find(kAbfssScheme) == 0;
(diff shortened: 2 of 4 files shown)


JkSelf:

majetideepak:
CC: @czentgr @xin-zhang2

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
```diff
@@ -45,4 +47,26 @@ inline std::string throwStorageExceptionWithOperationDetails(
   VELOX_FAIL(errMsg);
 }
 
+inline void extractCacheKeyFromConfig(
+    const config::ConfigBase& config,
+    std::string& accountName,
```
majetideepak: Can we return a `std::pair<std::string, std::string>` ?

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
```diff
@@ -68,41 +68,29 @@ void registerAbfsFileSystem() {
...
-      auto dot = remaining.find(".");
-      VELOX_USER_CHECK_NE(
-          dot,
-          std::string_view::npos,
-          "Invalid Azure account auth type key: {}",
-          key);
-      auto accountName = std::string(remaining.substr(0, dot));
+  std::string accountName;
+  std::string authType;
+  extractCacheKeyFromConfig(config, accountName, authType);
```
xin-zhang2: The code logic is different from the original. If the config contains multiple `fs.azure.account.auth.type.` items for different accounts, we need to register for all accounts. The current `extractCacheKeyFromConfig` will only return the last one.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
```diff
@@ -45,4 +47,26 @@ inline std::string throwStorageExceptionWithOperationDetails(
   VELOX_FAIL(errMsg);
 }
 
+inline void extractCacheKeyFromConfig(
```
jinchengchenghh: remove inline, this should not be a top used function

JkSelf: @jinchengchenghh Yes. Removed.

velox/connectors/hive/storage_adapters/abfs/RegisterAbfsFileSystem.cpp
JkSelf: @xin-zhang2 Good catch. Resolved. Could you help to review again? Thanks.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
JkSelf: @majetideepak Add `struct CacheKey` to store the `accountName `and `authType`.

JkSelf:
@majetideepak @xin-zhang2 @jinchengchenghh Resolved all your comments. Could you help to review again? Thanks.

zhli1142015:
What's the scenario for using account and auth type as the cache key here? My understanding is that an account will eventually use only one type of auth to access it, even if the user has configured multiple types of auth.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+
+namespace facebook::velox::filesystems {
+
+std::vector<CacheKey> extractCacheKeyFromConfig(
+    const config::ConfigBase& config) {
+  std::vector<CacheKey> cacheKeys;
+  for (const auto& [key, value] : config.rawConfigs()) {
+    constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
```
xin-zhang2: authTypePrefix can be moved outside the for loop

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
```diff
@@ -0,0 +1,45 @@
...
+      // Extract the accountName after "fs.azure.account.auth.type.".
+      auto remaining = skey.substr(authTypePrefix.size() + 1);
+      auto dot = remaining.find(".");
+      VELOX_USER_CHECK_NE(
+          dot,
+          std::string_view::npos,
+          "Invalid Azure account auth type key: {}",
+          key);
+      auto accountName = std::string(remaining.substr(0, dot));
+      cacheKeys.emplace_back(CacheKey{accountName, value});
```
xin-zhang2: The construction of accountName can be moved into CacheKey{} to avoid an extra copy.

majetideepak:
> What's the scenario for using account and auth type as the cache key here? 

@zhli1142015 We need this to support multiple abfsfilesystem instances. Each instance will have a different config. Each config could use the same account name but with a different auth mechanism.

JkSelf:
> What's the scenario for using account and auth type as the cache key here? My understanding is that an account will eventually use only one type of auth to access it, even if the user has configured multiple types of auth.

@zhli1142015 Thanks for your review.
In your production environment, if the same account name is configured with multiple authentication types, is only one auth type active at a time? If so, how does the system determine which one to use?

zhli1142015:
We don't have such a complex scenario. We decide which provider a user uses based on the auth type config. We consider this unique per account because the config is stored in a dictionary and passed in during connector initialization.

My understanding is that the original FileSystem was a singleton, so AzureClientProviderFactories was also designed as a singleton. If we need to support multiple FileSystems, making AzureClientProviderFactories a property of FileSystem might be a more straightforward approach.

JkSelf:
> If we need to support multiple FileSystems, making AzureClientProviderFactories a property of FileSystem might be a more straightforward approach.

@zhli1142015 Thanks for your response.
If each account uses only one auth type, we could use just the accountName as the cache key without including the auth type. We could also register the `AzureClientProvider `directly when initializing the `AbfsFileSystem`. CC @majetideepak @xin-zhang2

majetideepak:
> We decide which provider a user uses based on the auth type config. We consider this unique per account because the config is stored in a dictionary and passed in during connector initialization.

@JkSelf This is their internal system implementation. But in practice, an account can have multiple auth mechanisms as Zhen mentioned earlier. Let's stick to account-name_auth-mechanism for safety.

@xin-zhang2, @czentgr and I also discussed the need to remove the `registerAzureClientProvider` call. When creating the fileSystem, we should also check if a custom clientProvider is provided. If not, we set the default factory (SharedKeyAzureClientProvider, OAuthAzureClientProvider, FixedSasAzureClientProvider) in the fileSystem and use that to create the read and write clients.

@zhli1142015 what do you think?

zhli1142015:
Thank you for your explanation. I'm ok with your proposal. 
We have some custom client factories that need to call Java methods via JNI to obtain tokens, so they are implemented in Gluten. In Hadoop, custom client providers are registered through configuration and initilized by reflection. However, we cannot use reflection here to initialize them, so we still need an API to allow registering instances of custom client factories in Gluten.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
czentgr: accountName can be a string_view still, and we can make a new constructor for CacheKey to then create the std::string to avoid one copy and only have the std::string created in the CacheKey constructor.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
JkSelf: @xin-zhang2 Yes. Moved outside the for loop.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
JkSelf: Add the constructor for CacheKey to reduce the copy.

```
CacheKey(std::string_view accountName, std::string_view authType)
      : accountName(accountName), authType(authType) {}
```

JkSelf:
@majetideepak @jinchengchenghh @xin-zhang2 @czentgr @zhli1142015 Could you help to review again? Thanks.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
```diff
@@ -26,6 +26,16 @@ constexpr std::string_view kAbfsScheme{"abfs://"};
 constexpr std::string_view kAbfssScheme{"abfss://"};
 } // namespace
 
+class ConfigBase;
+
+struct CacheKey {
```
jinchengchenghh, line 31: const std::string accountName;

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
```diff
@@ -26,6 +26,16 @@ constexpr std::string_view kAbfsScheme{"abfs://"};
...
+  std::string accountName;
+  std::string authType;
+
+  CacheKey(std::string_view accountName, std::string_view authType)
```
jinchengchenghh: Do we need the ctor?

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
```diff
@@ -0,0 +1,45 @@
...
+      // Extract the accountName after "fs.azure.account.auth.type.".
+      auto remaining = skey.substr(authTypePrefix.size() + 1);
+      auto dot = remaining.find(".");
+      VELOX_USER_CHECK_NE(
+          dot,
+          std::string_view::npos,
+          "Invalid Azure account auth type key: {}",
+          key);
+      std::string_view accountName = std::string(remaining.substr(0, dot));
+      cacheKeys.emplace_back(CacheKey{accountName, value});
```
jinchengchenghh: cacheKeys.emplace_back(CacheKey{remaining.substr(0, dot), value});

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+std::vector<CacheKey> extractCacheKeyFromConfig(
+    const config::ConfigBase& config) {
+  std::vector<CacheKey> cacheKeys;
+  constexpr std::string_view authTypePrefix{kAzureAccountAuthType};
+  for (const auto& [key, value] : config.rawConfigs()) {
+    if (key.find(authTypePrefix) == 0) {
+      std::string_view skey = key;
+      // Extract the accountName after "fs.azure.account.auth.type.".
+      auto remaining = skey.substr(authTypePrefix.size() + 1);
```
jinchengchenghh: std::string_view remaining = std::string_view(key).substr(authTypePrefix.size() + 1);
Remove the used only once variable

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h"
+#include "velox/common/config/Config.h"
+#include "velox/connectors/hive/storage_adapters/abfs/AbfsPath.h"
+
+namespace facebook::velox::filesystems {
+
+std::vector<CacheKey> extractCacheKeyFromConfig(
```
jinchengchenghh, line 23: Could we use std::pair?

JkSelf: @jinchengchenghh I think using CacheKey instead of std::pair makes the code easier to understand and maintain.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
JkSelf: @jinchengchenghh Yes. Updated.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp
JkSelf: @jinchengchenghh Yes. Updated.

velox/connectors/hive/storage_adapters/abfs/AbfsUtil.h
JkSelf: After making accountName a const std::string, we cannot directly use: cacheKeys.emplace_back(CacheKey{remaining.substr(0, dot), value}); This results in the following compilation error: 
```
/mnt/DP_disk3/jk/projects/fb-velox/velox/velox/connectors/hive/storage_adapters/abfs/AbfsUtil.cpp:37:55: error: could not convert ‘remaining.std::basic_string_view<char>::substr(0, dot)’ from ‘std::basic_string_view<char>’ to ‘const string’ {aka ‘const std::__cxx11::basic_string<char>’}
   37 |       cacheKeys.emplace_back(CacheKey{remaining.substr(0, dot), value});
      |                                       ~~~~~~~~~~~~~~~~^~~~~~~~
      |                                                       |
      |                                                       std::basic_string_view<char>
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-unknown-warning-option’ may have been intended to silence earlier diagnostics
ninja: build stopped: subcommand failed.
make[1]: *** [Makefile:112: build] Error 1

```
Without the constructor, we would need to explicitly convert the string_view: cacheKeys.emplace_back(CacheKey{std::string(remaining.substr(0, dot)), value}); This approach requires 2 copies. However, by adding the constructor, we reduce it to only 1 copy.

jinchengchenghh (approved):

JkSelf:
@majetideepak @czentgr @zhli1142015 Could you help to review again? Thanks.

majetideepak (approved):

zhli1142015 (approved):

PR #14935, opened 2025-09-23
feat: Add Spark div function

zhli1142015:
Add support for the Spark IntegralDivide function. Implement the div function 
for both integral and decimal types, and add the corresponding checked_div 
function to support ANSI mode.
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L850

### DIFF
--- a/velox/docs/functions/spark/decimal.rst
+++ b/velox/docs/functions/spark/decimal.rst
@@ -38,7 +38,7 @@ The HiveQL behavior:
 https://cwiki.apache.org/confluence/download/attachments/27362075/Hive_Decimal_Precision_Scale_Support.pdf
 
 Additionally, the computation of decimal division adapts to the allow-precision-loss flag,
-while the decimal addition, subtraction, and multiplication do not.
+while the decimal addition, subtraction, multiplication and integer division do not.
 
 Addition and Subtraction
 ~~~~~~~~~~~~~~~~~~~~~~~~
@@ -75,30 +75,13 @@ When allow-precision-loss is false:
     s = fractionalDigits
 
 Integer Division
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~
 
-Integer division operators return a ``bigint`` result and are not affected by the ``allow-precision-loss`` configuration.
-The input precision formula ``p1-s1+s2`` (capped at 38) determines the valid range for the operands.
-If ``p1-s1+s2`` exceeds 38, the effective precision is 38. For example, with ``p1``=38, ``s1``=1, ``s2``=2,
-the formula yields 39 but the effective precision used is 38.
+::
 
-.. spark:function:: div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
-
-    Returns the results of dividing x by y. Performs the integer division truncates toward zero.
-    Truncation occurs if the result is within the precision(min(p1-s1+s2, 38)) but exceeds the BIGINT range.
-    Division by zero or overflow results in null. ::
-
-        SELECT CAST(1 as DECIMAL(17, 3)) div CAST(2 as DECIMAL(17, 3)); -- 0
-        SELECT CAST(21 as DECIMAL(20, 3)) div CAST(20 as DECIMAL(20, 2)); -- 1
-        SELECT CAST(1 as DECIMAL(20, 3)) div CAST(0 as DECIMAL(20, 3)); -- NULL
-        SELECT CAST(99999999999999999999999999999999999 as DECIMAL(38, 1)) div CAST(0.001 as DECIMAL(7, 4)); -- 687399551400672280 // Result is truncated to int64_t.
-
-.. spark:function:: checked_div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
-
-    Returns the result of integer division of ``x`` by ``y``, truncating toward zero.
-    Truncation occurs if the result is within the precision(min(``p1-s1+s2``, 38)) but exceeds the BIGINT range.
-    Division by zero or overflow results in an error. This function operates in ANSI mode (error on invalid input).
-    Corresponds to Spark's operator ``div`` with ``spark.sql.ansi.enabled`` set to true.
+    precision = p1 - s1 + s2
+    p = precision == 0 ? 1 : min(38,  precision)
+    s = 0
 
 Decimal Precision and Scale Adjustment
 --------------------------------------
@@ -137,6 +120,28 @@ Decimal division uses a different formula:
     scale = fractionalDigits - (wholeDigits + fractionalDigits - 38) / 2 - 1
 
 Returns NULL when the actual result cannot be represented with the calculated decimal type.
+
+Arithmetic Functions
+--------------------
+
+.. spark:function:: div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
+
+    Performs integer division and returns the bigint result of dividing ``x`` by ``y``, truncating toward zero.
+    Truncation occurs if the result is within the result precision but exceeds the BIGINT range.
+    Division by zero or overflow results in NULL. Does not respect the ``allow-precision-loss`` configuration.
+    Corresponds to Spark's operator ``div`` with ``spark.sql.ansi.enabled`` set to false.  ::
+
+        SELECT CAST(1 as DECIMAL(17, 3)) div CAST(2 as DECIMAL(17, 3)); -- 0
+        SELECT CAST(21 as DECIMAL(20, 3)) div CAST(20 as DECIMAL(20, 2)); -- 1
+        SELECT CAST(1 as DECIMAL(20, 3)) div CAST(0 as DECIMAL(20, 3)); -- NULL
+        SELECT CAST(99999999999999999999999999999999999 as DECIMAL(38, 1)) div CAST(0.001 as DECIMAL(7, 4)); -- 687399551400672280 // Result is truncated to int64_t.
+
+.. spark:function:: checked_div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
+
+    Performs integer division and returns the bigint result of dividing ``x`` by ``y``, truncating toward zero.
+    Truncation occurs if the result is within the result precision but exceeds the BIGINT range.
+    Division by zero or overflow results in an error.
+    Corresponds to Spark's operator ``div`` with ``spark.sql.ansi.enabled`` set to true.
 
 Decimal Functions
 -----------------


velox/docs/functions/spark/math.rst
```diff
@@ -83,6 +83,13 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
+.. function:: checked_div(x, y) -> bigint
+
+    Returns the results of dividing (integer division performs truncation) x by y.
+    Supported types are DECIMAL and integral types. For DECIMAL types, x and y can have different precision and scale.
+    For integral types, x and y must have the same type.
+    Division by zero or overflow results in an error. Corresponds to Spark's operator ``div`` with ``failOnError`` as true.
```
jinchengchenghh: I see checked_divide, is it only difference with ANSI? If that, move it to checked_divide. See this as example https://github.com/facebookincubator/velox/pull/14346

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,127 @@ struct DecimalDivideFunction {
   uint8_t rPrecision_;
 };
 
+/// Decimal integral divide function implementation.
+struct DecimalIntegralDivideBase {
```
jinchengchenghh, line 503: Simple decimal function is supported, could you use simple function for it, an example is https://github.com/facebookincubator/velox/pull/13174

velox/docs/functions/spark/math.rst
zhli1142015: checked_divide implements ANSI-mode-compliant floating-point division, while checked_div implements ANSI-mode-compliant integral division (with truncation). They are different.
div and checked_div can not be merged together like abs. The current simple API does not support returning both Status and isResultNull. Moreover, there should be no case where we need to both return an error Status and set the result to null. Therefore, we need to provide a separate implementation for ANSI mode, just like other arithmetic functions.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: This is implemented using the simple API, and this class extracts the common logic.

velox/docs/functions/spark/math.rst
```diff
@@ -83,6 +83,13 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
+.. function:: checked_div(x, y) -> bigint
```
jinchengchenghh, line 86: Add (ANSI compliant)

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,127 @@ struct DecimalDivideFunction {
...
+    auto [bPrecision, bScale] = getDecimalPrecisionScale(*inputTypes[1]);
+    auto [rPrecision, rScale] =
+        computeResultPrecisionScale(aPrecision, aScale, bPrecision, bScale);
+    rPrecision_ = rPrecision;
+    aRescale_ = std::max<int8_t>(0, bScale - aScale);
+    bRescale_ = std::max<int8_t>(0, aScale - bScale);
+  }
+
+  template <typename A, typename B>
+  bool computeQuotient(int64_t& out, const A& a, const B& b) {
```
jinchengchenghh, line 517: Could we do a refactor for existing decimal divide to avoid copy too much code?

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,127 @@ struct DecimalDivideFunction {
...
+
+    if (!velox::DecimalUtil::valueInPrecisionRange(quotient, rPrecision_)) {
+      return false;
+    }
+    out = static_cast<int64_t>(quotient);
+    return true;
+  }
+
+ private:
+  static std::pair<uint8_t, uint8_t> computeResultPrecisionScale(
```
jinchengchenghh: The ANSI true or false should not affect the result precision and scale.

velox/functions/sparksql/Arithmetic.h
```diff
@@ -642,4 +642,54 @@ struct CheckedDivideFunction {
   }
 };
 
+template <typename TExec>
+struct IntegeralDivideFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
jinchengchenghh: VELOX_DEFINE_FUNCTION_TYPES is not used

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
```diff
@@ -714,5 +714,65 @@ TEST_F(DecimalArithmeticTest, ceil) {
            0},
           DECIMAL(19, 19))});
 }
+
+TEST_F(DecimalArithmeticTest, div) {
```
jinchengchenghh, line 737: Please use simple API to test the decimal type as iceberg bucket function

velox/docs/functions/spark/math.rst
jinchengchenghh: Div could always return bool.
For checked_div, could we implement callNullable and return Status?

jinchengchenghh:
And, if we respect ansi enabled config, we do not need to do other conversion in Gluten side.

velox/docs/functions/spark/math.rst
zhli1142015: --> For checked_div, could we implement callNullable and return Status?

We don't need to implement this, as checked_div would not return null.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: --> avoid copy too much code?
Could you point out which code you are referring to?

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: I agree, so it's implemented in DecimalIntegralDivideBase.

velox/functions/sparksql/Arithmetic.h
zhli1142015: Could you explain this comment? I might not have understood what it means.

zhli1142015:
> And, if we respect ansi enabled config, we do not need to do other conversion in Gluten side.

Here, with ANSI mode off, we require the call function to return a bool. With ANSI mode on, we require the call function to return a Status. So this cannot be achieved using Velox's config.
If it could, it would mean the following implementation, which might cause performance issues if we use try(div(ansi on)) .

image

velox/functions/sparksql/Arithmetic.h
jinchengchenghh: VELOX_DEFINE_FUNCTION_TYPES introduce arg_type<VARCHR>, we do not use the function type micro.

velox/functions/sparksql/Arithmetic.h
```diff
@@ -642,4 +642,44 @@ struct CheckedDivideFunction {
...
+    }
+
+    result = a / b;
+    return true;
+  }
+};
+
+template <typename TExec>
+struct CheckedIntegeralDivideFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
jinchengchenghh: ditto

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+    auto [bPrecision, bScale] = getDecimalPrecisionScale(*inputTypes[1]);
+    auto [rPrecision, rScale] =
+        computeResultPrecisionScale(aPrecision, aScale, bPrecision, bScale);
+    rPrecision_ = rPrecision;
+    aRescale_ = std::max<int8_t>(0, bScale - aScale);
+    bRescale_ = std::max<int8_t>(0, aScale - bScale);
+  }
+
+  template <typename A, typename B>
+  bool computeQuotient(int64_t& out, const A& a, const B& b) {
```
jinchengchenghh, line 517: Is the result always int64_t

velox/functions/sparksql/DecimalArithmetic.cpp
jinchengchenghh: rScale is not used, change to computeResultPrecision

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+    return true;
+  }
+
+ private:
+  static std::pair<uint8_t, uint8_t> computeResultPrecisionScale(
+      uint8_t aPrecision,
+      uint8_t aScale,
+      uint8_t bPrecision,
+      uint8_t bScale) {
+    auto intPrecision = aPrecision - aScale + bScale;
```
jinchengchenghh, line 551: If the aPrecision is 38, and aScale = 0, bScale = 0, the result should be long decimal, why the return type is always int64_t

velox/docs/functions/spark/math.rst
jinchengchenghh: Please add the comment the result scale is always 0

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+    }
+    out = static_cast<int64_t>(quotient);
+    return true;
+  }
+
+ private:
+  static std::pair<uint8_t, uint8_t> computeResultPrecisionScale(
+      uint8_t aPrecision,
+      uint8_t aScale,
+      uint8_t bPrecision,
```
jinchengchenghh: bPrecision is not used

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+    return DecimalUtil::bounded(intPrecision, 0);
+  }
+
+ protected:
+  uint8_t aRescale_;
+  uint8_t bRescale_;
+  uint8_t rPrecision_;
+};
+
+/// Decimal integral divide function that returns null on division by zero or
```
jinchengchenghh: Use // for private comment

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+/// overflow.
+template <typename TExec>
+struct DecimalIntegeralDivideFunction : DecimalIntegralDivideBase {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
jinchengchenghh: ditto

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
...
+  template <typename A, typename B>
+  bool call(int64_t& out, const A& a, const B& b) {
+    if (b == 0) {
+      return false;
+    }
+    return computeQuotient(out, a, b);
+  }
+};
+
+/// Decimal integral divide function that returns error on division by zero or
```
jinchengchenghh: ditto

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,118 @@ struct DecimalDivideFunction {
   uint8_t rPrecision_;
 };
 
+/// Decimal integral divide function implementation.
+struct DecimalIntegralDivideBase {
+  template <typename A, typename B>
+  void
+  initializeBase(const std::vector<TypePtr>& inputTypes, A* /*a*/, B* /*b*/) {
```
jinchengchenghh: Remove the argument A* /*a*/, B* /*b*/

jinchengchenghh:
I see, we cannot set the result to null even with callNullable, since Spark also uses separate Expression for ANSI mode, I agree we could use two functions for it. Thanks for your clarification.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: Updated, thanks.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: Yes, it's always int64_t.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: I don't know why, but in spark ut there is a similar case, overflow caused the wrong result returned.
https://github.com/apache/spark/blob/master/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ArithmeticExpressionSuite.scala#L742

jinchengchenghh: Current implementation overflow will cause exception, should we align with Spark?https://github.com/facebookincubator/velox/pull/14935/files#diff-4baee9f82f9c347f753972415d345941d2c2a110b608d480b090650171da096bR536

zhli1142015: If the result exceeds the range representable by rPrecision, it will either trigger an overflow  error (ANSI) or return null. If the result only exceeds the range representable by int64_t, the value will be truncated. We follow the same behavor here.
https://github.com/facebookincubator/velox/pull/14935/files#diff-257bd1947469dfd8494c7e19f535d4a3247b3df086287110e53740c199104073R785

https://github.com/facebookincubator/velox/pull/14935/files#diff-257bd1947469dfd8494c7e19f535d4a3247b3df086287110e53740c199104073R802

jinchengchenghh: Could you add the comment to computeQuotient?

jinchengchenghh: Maybe also the document to describe when the result is incorrect, would be better to add an example, SELECT XXX

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
```diff
@@ -714,5 +732,105 @@ TEST_F(DecimalArithmeticTest, ceil) {
...
+  EXPECT_EQ(
+      (div<int128_t, int128_t>(
+          DECIMAL(38, 0),
+          DECIMAL(38, 6),
+          HugeInt::parse("99999999999999999999999999999999999999"),
+          1)),
+      std::nullopt);
+}
+
+TEST_F(DecimalArithmeticTest, checked_div) {
```
jinchengchenghh: checked_div -> checkedDiv

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: Done.

velox/docs/functions/spark/math.rst
```diff
@@ -83,6 +83,14 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
+.. function:: checked_div(x, y) -> bigint (ANSI compliant)
+
+    Returns the result of integer division of x by y, truncating toward zero.
+    Supported types are DECIMAL and integral types. For DECIMAL types, x and y can have different precision and scale.
+    For integral types, x and y must have the same type.
+    The result type is BIGINT, truncation may occur if the result is within the result precision but exceeds the BIGINT range.
```
jinchengchenghh: truncation occurs when, 
describe what is the result precision since the result is bigint

velox/docs/functions/spark/math.rst
```diff
@@ -83,6 +83,14 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
+.. function:: checked_div(x, y) -> bigint (ANSI compliant)
+
+    Returns the result of integer division of x by y, truncating toward zero.
+    Supported types are DECIMAL and integral types. For DECIMAL types, x and y can have different precision and scale.
+    For integral types, x and y must have the same type.
+    Truncation may occur if the result is within the result precision(p1-s1+s2) but exceeds the BIGINT range.
```
jinchengchenghh: For decimal types, truncation occurs if the result is within the precision(p1-s1+s2) but exceeds the BIGINT range.
And also add the comment to `div`

velox/docs/functions/spark/math.rst
```diff
@@ -118,6 +126,20 @@ Mathematical Functions
...
     Converts angle x in radians to degrees.
 
+.. function:: div(x, y) -> bigint
+
+    Returns the results of dividing x by y. Performs the integer division truncates toward zero.
+    Supported types are DECIMAL and integral types. For DECIMAL types, x and y can have different precision and scale.
+    For integral types, x and y must have the same type.
+    Division by zero or overflow results in null. ::
+
+        SELECT 3 div 2; -- 1
```
jinchengchenghh, line 134: Add an incorrect result example

jinchengchenghh (approved):
If you have further comments, @rui-mo

velox/functions/lib/RegistrationHelpers.h
```diff
@@ -71,6 +71,15 @@ void registerUnaryIntegralWithTReturn(const std::vector<std::string>& aliases) {
   registerFunction<T, TReturn, int64_t>(aliases);
 }
 
+template <template <class> class T, typename TReturn>
+void registerBinaryIntegralWithTReturn(
+    const std::vector<std::string>& aliases) {
+  registerFunction<T, TReturn, int8_t, int8_t>(aliases);
+  registerFunction<T, TReturn, int16_t, int16_t>(aliases);
+  registerFunction<T, TReturn, int32_t, int32_t>(aliases);
+  registerFunction<T, TReturn, int64_t, int64_t>(aliases);
```
rui-mo: It seems to overlap with `registerBinaryScalar`.

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,111 @@ struct DecimalDivideFunction {
...
+  // inaccurate if it exceeds int64_t range.
+  template <typename A, typename B>
+  bool computeQuotient(int64_t& out, const A& a, const B& b) {
+    // Determine sign and convert to absolute values.
+    bool isNegative = (a < 0) != (b < 0);
+    int128_t absA = static_cast<int128_t>(a < 0 ? -a : a);
+    int128_t absB = static_cast<int128_t>(b < 0 ? -b : b);
+
+    // Scale values, checking for overflow.
+    int128_t scaledA, scaledB;
```
rui-mo: nit: declare one variable in each line.

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,111 @@ struct DecimalDivideFunction {
...
+ protected:
+  uint8_t aRescale_;
+  uint8_t bRescale_;
+  uint8_t rPrecision_;
+};
+
+// Decimal integral divide function that returns null on division by zero or
+// overflow.
+template <typename TExec>
+struct DecimalIntegeralDivideFunction : DecimalIntegralDivideBase {
```
rui-mo: typo: note the extra "e" in "Integeral", which should be "Integral".

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -499,6 +499,111 @@ struct DecimalDivideFunction {
...
+    auto bScale = getDecimalPrecisionScale(*inputTypes[1]).second;
+    rPrecision_ = computeResultPrecision(aPrecision, aScale, bScale);
+    aRescale_ = std::max<int8_t>(0, bScale - aScale);
+    bRescale_ = std::max<int8_t>(0, aScale - bScale);
+  }
+
+  // Computes the quotient of 'a' and 'b' and stores it in 'out'. Returns false
+  // if the result exceeds rPrecision_ or if overflow occurs during scaling.
+  // Following Spark's behavior, the result is truncated to int64_t and may be
+  // inaccurate if it exceeds int64_t range.
```
rui-mo: > may be inaccurate if it exceeds int64_t range.

Would you please clarify how would be result become inaccurate, and does the result match Spark?

velox/docs/functions/spark/math.rst
```diff
@@ -83,6 +83,14 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
+.. function:: checked_div(x, y) -> bigint (ANSI compliant)
+
+    Returns the result of integer division of x by y, truncating toward zero.
+    Supported types are DECIMAL and integral types. For DECIMAL types, x and y can have different precision and scale.
+    For integral types, x and y must have the same type.
+    For decimal types, truncation occurs if the result is within the precision(p1-s1+s2) but exceeds the BIGINT range.
```
rui-mo: Could you document the formula used to calculate the result precision? Currently, p1, s1, and s2 are undefined, which may cause confusion. It might also be helpful to move the documentation for decimal division to https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/spark/decimal.rst.

velox/functions/lib/RegistrationHelpers.h
zhli1142015: Updated, thanks.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: Good catch, thanks.

velox/functions/sparksql/DecimalArithmetic.cpp
zhli1142015: Below is an realted example, yes it matches the spark behavior.
`SELECT CAST(99999999999999999999999999999999999 as DECIMAL(38, 1)) div CAST(0.001 as DECIMAL(7, 4)); -- 687399551400672280`

velox/docs/functions/spark/math.rst
zhli1142015: Updated, thanks.

zhli1142015:
UT failure is not related, https://github.com/facebookincubator/velox/issues/15093

velox/docs/functions/spark/decimal.rst
```diff
@@ -122,6 +122,12 @@ Decimal Functions
 
         SELECT ceil(cast(1.23 as DECIMAL(3, 2))); -- 2 // Output type: decimal(2,0)
 
+.. spark:function:: checked_div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint (ANSI compliant)
```
rui-mo: Would it be more appropriate to place these in the `Decimal Operators` section above, since it also involves a division operation?

rui-mo: `ANSI compliant` usually means this function respects the ANSI configuration and supports both ANSI ON and OFF modes. For this function, could you please instead document which mode it adapts to?

rui-mo (commented):
Thanks, and sorry for the delay.

velox/docs/functions/spark/decimal.rst
```diff
@@ -74,6 +74,27 @@ When allow-precision-loss is false:
     p = wholeDigits + fractionalDigits
     s = fractionalDigits
 
+Integer Division
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. spark:function:: div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
+
+    Returns the results of dividing x by y. Performs the integer division truncates toward zero.
```
rui-mo: nit: refer to the variables \``x\`` and \``y\``.

velox/docs/functions/spark/decimal.rst
```diff
@@ -74,6 +74,27 @@ When allow-precision-loss is false:
...
+    Truncation occurs if the result is within the precision(p1-s1+s2) but exceeds the BIGINT range.
```
rui-mo: > p1-s1+s2

ditto

velox/docs/functions/spark/decimal.rst
```diff
@@ -74,6 +74,27 @@ When allow-precision-loss is false:
...
+        SELECT CAST(21 as DECIMAL(20, 3)) div CAST(20 as DECIMAL(20, 2)); -- 1
+        SELECT CAST(1 as DECIMAL(20, 3)) div CAST(0 as DECIMAL(20, 3)); -- NULL
+        SELECT CAST(99999999999999999999999999999999999 as DECIMAL(38, 1)) div CAST(0.001 as DECIMAL(7, 4)); -- 687399551400672280 // Result is truncated to int64_t.
+
+.. spark:function:: checked_div(x: decimal(p1, s1), y: decimal(p2, s2)) -> bigint
+
+    Returns the result of integer division of x by y, truncating toward zero.
+    Truncation occurs if the result is within the precision(p1-s1+s2) but exceeds the BIGINT range.
+    Division by zero or overflow results in an error. This function operates in ANSI mode (error on invalid input).
+    Corresponds to Spark's operator ``div`` with ``spark.sql.ansi.enabled`` set to true.
```
rui-mo: Will the above functions comply with the allow-precision-loss configuration? I’m asking to determine whether this is the right place to include their documentation. Thanks.

rui-mo (commented):
Thanks! Added some comments on the documentation.

velox/docs/functions/spark/math.rst
```diff
@@ -83,7 +83,13 @@ Mathematical Functions
     Returns the result of adding x to y. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``+`` with ``failOnError`` as true.
 
-.. function:: checked_divide(x, y) -> [same as x]
+.. function:: checked_div(x, y) -> bigint (ANSI compliant)
```
rui-mo: Is this function only for ANSI-on mode, meaning it is not fully ANSI-compliant because the ANSI configuration is not respected?

velox/docs/functions/spark/decimal.rst
```diff
@@ -74,6 +74,30 @@ When allow-precision-loss is false:
     p = wholeDigits + fractionalDigits
     s = fractionalDigits
 
+Integer Division
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Integer division operators return a ``bigint`` result and are not affected by the ``allow-precision-loss`` configuration.
+The input precision formula ``p1-s1+s2`` determines the valid range for the operands.
```
rui-mo: Could you clarify the behavior when `p1 - s1 + s2` exceeds 38? For example, when p1 = 38, s1 = 1, and s2 = 2?

zhli1142015:
LGTM, thanks

rui-mo (approved):

PR #14936, opened 2025-09-23
refactor: Use VELOX_USER_RETURN in Spark arithmetic functions

zhli1142015:
Refactor Spark checked_arithmetic functions to use VELOX_USER_RETURN for error handling.

### DIFF
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -558,18 +558,15 @@ struct RIntFunction {
 
 template <typename TExec>
 struct CheckedAddFunction {
-  VELOX_DEFINE_FUNCTION_TYPES(TExec);
   template <typename T>
   FOLLY_ALWAYS_INLINE Status call(T& result, const T& a, const T& b) {
     if constexpr (std::is_integral_v<T>) {
       T res;
-      bool overflow = __builtin_add_overflow(a, b, &res);
-      if (UNLIKELY(overflow)) {
-        if (threadSkipErrorDetails()) {
-          return Status::UserError();
-        }
-        return Status::UserError("Arithmetic overflow: {} + {}", a, b);
-      }
+      VELOX_USER_RETURN(
+          __builtin_add_overflow(a, b, &res),
+          "Arithmetic overflow: {} + {}",
+          a,
+          b);
       result = res;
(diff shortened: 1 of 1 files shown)


velox/functions/sparksql/Arithmetic.h
```diff
@@ -549,27 +549,22 @@ struct UnHexFunction {
...
     if constexpr (std::is_integral_v<T>) {
       T res;
-      bool overflow = __builtin_add_overflow(a, b, &res);
-      if (UNLIKELY(overflow)) {
-        if (threadSkipErrorDetails()) {
-          return Status::UserError();
-        }
-        return Status::UserError("Arithmetic overflow: {} + {}", a, b);
-      }
+      VELOX_USER_RETURN(
```
jinchengchenghh, line 565: Please don't change here, this function is commonly used in try_cast, this API may cause performance issue since it always return Error with detailed message

zhli1142015: Good point, but instread of reverting this change, I think we should add support for this pattern in these macros since other functions using them can be called together with the try() function also. This way, those functions can also benefit from the performance improvement you mentioned. Do you want to make this change, or should I handle it?

jinchengchenghh: Could you give any insights? Do we need VELOX_USER_RETURN (skipThreadError, ...) in function scope? I don't have strong preference. @mbasmanova

zhli1142015: I think the macro VELOX_USER_RETURN may be defined as something below:
image

jinchengchenghh: This is a low load check, so I agree we can add it to current API

mbasmanova: What are you trying to do here? Please, explain the problem you are trying to solve and how you want to solve it?

Please, check out https://velox-lib.io/blog/optimize-try-more/ for context.

zhli1142015:
Okay, I've split this PR into three smaller PRs
https://github.com/facebookincubator/velox/pull/14936
https://github.com/facebookincubator/velox/pull/14952
https://github.com/facebookincubator/velox/pull/14951

PR #14942, opened 2025-09-24
build: Use FBThrift instead of Apache Thrift

### LINKED ISSUE #13175: Add support for FBThrift in Parquet
### Discussed in https://github.com/facebookincubator/velox/discussions/6163

<div type='discussions-op-text'>

<sup>Originally posted by **pedroerp** August 18, 2023</sup>
We currently have a dependency on thrift from our native parquet reader. As we add support for remote function execution, the client <-> server communication will be done using thrift RPC as well. 

FBThrift is a newer and supposedly more efficient version than thrift, and among other things provides an async transport called "cpp2" (not available in thrift), which will be used for remote function execution. 

Linking thrift and fbthrift to the same binary causes symbols to conflict and end up causing crashes as the wrong version of symbols eventually get picked. There was a long discussion on Slack (check #builds-and-ci channel) on whether this could be solved using dynamic linking, but so solution was found. 

With this context, the proposal is to remove the thrift dependency from Velox (created due to parquet), and consolidate all of that on FBThrift. Despite their divergence, both projects seem to support TCompactProtocol (the ones used by Parquet), so, in theory (need some testing to validate), they should be compatible. FBThrift is also claimed to be more efficient, so there might be small performance gains in doing so.

The parquet generated cpp files are also today pre-compiled and checked in as part of the codebase; as part of this change we would also move them to be compiled at build time, like how it is done for remote functions. I've tested and all of them seems to be doable. 

There is also a thrift dependency that comes with Arrow (we use arrow's parquet writer), but this is nicely wrapped within arrow and doesn't leak its internal symbols. 

What do you all think?

UPDATE: just doubled checked with someone from the fbthrift team and the TCompactProtocol serialization needed for Parquet should be compatible. 

Cc: @assignUser @majetideepak @aditi-pandit @yingsu00 @kgpai @mbasmanova @xiaoxmeng @oerling </div>

kou:
Fixes #13175

See the associated issue why we want to move to FBThrift from Apache Thrift.

FBThrift and Apache Thrift has many incompatible API changes:

* `XXX.__isset.YYY` -> `XXX.YYY().has_value()` (or `bool(XXX.YYY())`): The latter implicit `bool(...)` syntax is used in this PR
* `XXX.__isset.YYY = true` -> `XXX.YYY().ensure()`
* `XXX.YYY` -> `XXX.YYY().value()` (or `*XXX.YYY()`): The latter `*...` syntax is used in this PR
* `XXX.__set_YYY(value)` -> `XXX.YYY() = value`
* `thrift::ThriftTransport`/`TCompactProcotolT` -> `folly::IOBuf`/`CompactProtocolReader`
* FBThrift doesn't support `optional` and default value that is used by `DataPageHeaderV2.is_compressed`: We need to use the default value manually in our code
* FBThrift raises an exception when we access to an optional field that isn't set (Apache Thrift doesn't have the check)
 
There are many changes in this PR but there is no logic change and no refactoring. This just converts API usage for FBThrift.


velox/dwio/parquet/reader/PageReader.cpp
```diff
@@ -92,16 +90,14 @@ PageHeader PageReader::readPageHeader() {
...
-  uint64_t readBytes;
-  readBytes = pageHeader.read(&protocol);
-
+  unsigned long readBytes = thrift::deserialize(
+      &pageHeader,
+      std::string_view(
+          reinterpret_cast<const char*>(bufferStart_),
+          bufferEnd_ - bufferStart_));
   pageDataStart_ = pageStart_ + readBytes;
+  bufferStart_ += readBytes;
```
kou: We need this because `thrift::ThriftTransport` does this internally with old code.

velox/dwio/parquet/reader/PageReader.cpp
```diff
@@ -301,21 +297,24 @@ void PageReader::prepareDataPageV2(const PageHeader& pageHeader, int64_t row) {
...
   auto levelsSize = repeatLength + defineLength;
   pageData_ += levelsSize;
-  if (pageHeader.data_page_header_v2.__isset.is_compressed &&
-      pageHeader.data_page_header_v2.is_compressed &&
-      (pageHeader.compressed_page_size - levelsSize > 0)) {
+  // parquet.thrift uses "7: optional bool is_compressed = true;" but
+  // FBThrift doesn't support "optional" and default value. (The
+  // default value isn't used for missing is_compressed.) So we need
+  // to use value_or(true) here.
+  if (pageHeader.data_page_header_v2()->is_compressed().value_or(true) &&
```
kou, line 304: This is a FBThrift limitation.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -228,39 +226,41 @@ void ReaderBase::loadFileMetaData() {
...
       "Invalid Parquet schema: Need at least one non-root column in the file");
+  // parquet.thrift says "The root of the schema does not have a
+  // repetition_type" but there are Parquet files that specify
+  // REQUIRED explicitly.
   VELOX_CHECK_EQ(
-      fileMetaData_->schema[0].repetition_type,
+      (*fileMetaData_->schema())[0].repetition_type().value_or(
+          thrift::FieldRepetitionType::REQUIRED),
       thrift::FieldRepetitionType::REQUIRED,
       "Invalid Parquet schema: root element must be REQUIRED");
```
kou, line 256: I think that we can remove this check but I keep this in this PR.

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -516,22 +521,23 @@ std::unique_ptr<ParquetTypeWithId> ReaderBase::getParquetColumnInfo(
...
               name);
       }
     } else {
-      if (schemaElement.repetition_type ==
-          thrift::FieldRepetitionType::REPEATED) {
-        if (schema[parentSchemaIdx].converted_type ==
-            thrift::ConvertedType::LIST) {
+      if (schemaElement.repetition_type() &&
+          *schemaElement.repetition_type() ==
+              thrift::FieldRepetitionType::REPEATED) {
```
kou, line 532: We can use `value_or()` here if we want:

```suggestion
      if (schemaElement.repetition_type().value_or(thrift::FieldRepetitionType::REPEATED) ==
              thrift::FieldRepetitionType::REPEATED) {
```

velox/dwio/parquet/thrift/CMakeLists.txt
```diff
@@ -12,5 +12,12 @@
...
-velox_link_libraries(velox_dwio_parquet_thrift arrow thrift Boost::headers fmt::fmt)
+include(FBThriftCppLibrary)
+
+add_fbthrift_cpp_library(velox_dwio_parquet_thrift_raw parquet.thrift)
+velox_add_library(velox_dwio_prquet_thrift)
+if(VELOX_MONO_LIBRARY)
+  target_link_libraries(velox velox_dwio_parquet_thrift_raw)
+else()
+  target_link_libraries(velox_dwio_parquet_thrift PUBLIC velox_dwio_parquet_thrift_raw)
+endif()
```
kou, line 22: We can cleanup this part once we cleanup `velox_add_library()` and its family.

velox/dwio/parquet/thrift/parquet.thrift
```diff
@@ -580,7 +582,12 @@ struct DataPageHeaderV2 {
   definition_levels_byte_length + repetition_levels_byte_length + 1 and compressed_page_size (included)
   is compressed with the compression_codec.
   If missing it is considered compressed */
-  7: optional bool is_compressed = 1;
+  // FBThrift rejects "optional" and default value. This annotation allows
+  // this usage but generated C++ code doesn't use the default value when
+  // "is_compressed" is missing. So our code must use
+  // is_compressed().value_or(true).
+  @thrift.AllowUnsafeOptionalCustomDefaultValue
```
kou, line 589: This is a FBThrift limitation.

velox/dwio/parquet/writer/arrow/PageIndex.cpp
```diff
@@ -572,8 +571,8 @@ class ColumnIndexBuilderImpl final : public ColumnIndexBuilder {
     state_ = BuilderState::kFinished;
 
     /// Clear null_counts vector because at least one page does not provide it.
-    if (!column_index_.__isset.null_counts) {
-      column_index_.null_counts.clear();
+    if (!column_index_.null_counts()) {
+      column_index_.null_counts().reset();
     }
```
kou, line 577: I think that we can remove this.

velox/dwio/parquet/writer/arrow/tests/FileDeserializeTest.cpp
```diff
@@ -105,13 +105,15 @@ static std::vector<Compression::type> GetSupportedCodecTypes() {
...
+    data_page_header_.encoding() =
         facebook::velox::parquet::thrift::Encoding::PLAIN;
-    data_page_header_.definition_level_encoding =
+    data_page_header_.definition_level_encoding() =
         facebook::velox::parquet::thrift::Encoding::RLE;
-    data_page_header_.repetition_level_encoding =
+    data_page_header_.repetition_level_encoding() =
         facebook::velox::parquet::thrift::Encoding::RLE;
 
+    data_page_header_v2_.is_compressed() = false;
```
kou, line 115: This is a FBThrift limitation.

velox/dwio/parquet/writer/arrow/tests/FileDeserializeTest.cpp
```diff
@@ -285,6 +288,7 @@ void TestPageSerde::TestPageSerdeCrc(
             data_size, static_cast<int32_t>(actual_size), checksum_opt));
       } else {
         if (write_data_page_v2) {
+          data_page_header_v2_.is_compressed() = (codec != nullptr);
```
kou, line 291: This is a FBThrift limitation.

velox/dwio/parquet/thrift/parquet.thrift
```diff
@@ -580,7 +582,13 @@ struct DataPageHeaderV2 {
...
   is compressed with the compression_codec.
   If missing it is considered compressed */
-  7: optional bool is_compressed = 1;
+  // FBThrift rejects "optional" and default value. This annotation allows
+  // this usage but generated C++ code doesn't use the default value when
+  // "is_compressed" is missing. So our code must use
+  // is_compressed().value_or(true).
+  // This annotation is available since FBThrift v2025.08.18.00.
+  // @thrift.AllowUnsafeOptionalCustomDefaultValue
+  7: optional bool is_compressed = true;
```
kou, line 590: This is a FBThrift limitation.

kou:
Hmm. Homebrew's Folly doesn't enable 128bit integer support.
We need to fix https://github.com/facebook/folly/issues/1666 before this.

majetideepak:
> Homebrew's Folly doesn't enable 128bit integer support.

@kou thanks for this work! We vendor (`run_and_time install_folly`) folly on MacOS. So this is not a blocker?

kou:
If we can use `install_{fizz,wangle,mvfst,fbthrift}` too, we can unblock it.

majetideepak:
> If we can use install_{fizz,wangle,mvfst,fbthrift} too, we can unblock it.

Yes, we do this inside `scripts/setup-macos.sh`

kou:
OK. We can build Velox on macOS by building FBThrift and related products by ourselves.

FYI: I opened a PR to Folly: https://github.com/facebook/folly/pull/2508

We need a patch for FBThrift to fix a link error: https://github.com/facebook/fbthrift/pull/675

https://github.com/facebookincubator/velox/actions/runs/18067953830/job/51413317250#step:7:5934

```text
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::detail::json::getTypeNameForTypeID(apache::thrift::protocol::TType)':
THeader.cpp:(.text+0x169): undefined reference to `apache::thrift::detail::json::throwUnrecognizedType()'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::detail::json::getTypeIDForTypeName(folly::Range<char const*>)::{lambda()#1}::operator()() const':
THeader.cpp:(.text+0x17d): undefined reference to `apache::thrift::detail::json::throwUnrecognizedType()'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::detail::json::clampSize(long)':
THeader.cpp:(.text+0x331): undefined reference to `apache::thrift::detail::json::throwNegativeSize(long)'
/usr/bin/ld: THeader.cpp:(.text+0x357): undefined reference to `apache::thrift::detail::json::throwExceededSizeLimit(long, long)'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::transport::THeader::untransform(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >, std::vector<apache::thrift::TTransform, std::allocator<apache::thrift::TTransform> >&)':
THeader.cpp:(.text+0x275d): undefined reference to `apache::thrift::rocket::CompressionAlgorithmSelector::fromTTransform(apache::thrift::TTransform const&)'
/usr/bin/ld: THeader.cpp:(.text+0x2780): undefined reference to `apache::thrift::rocket::CompressionManager::uncompressBuffer(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >&&, apache::thrift::CompressionAlgorithm)'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::transport::THeader::transform(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >, std::vector<apache::thrift::TTransform, std::allocator<apache::thrift::TTransform> >&, unsigned long)':
THeader.cpp:(.text+0x2998): undefined reference to `apache::thrift::rocket::CompressionAlgorithmSelector::fromTTransform(apache::thrift::TTransform const&)'
/usr/bin/ld: THeader.cpp:(.text+0x29bb): undefined reference to `apache::thrift::rocket::CompressionManager::compressBuffer(std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> >&&, apache::thrift::CompressionAlgorithm)'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::JSONProtocolWriterCommon::JSONProtocolWriterCommon(apache::thrift::ExternalBufferSharing)':
THeader.cpp:(.text._ZN6apache6thrift24JSONProtocolWriterCommonC2ENS0_21ExternalBufferSharingE[_ZN6apache6thrift24JSONProtocolWriterCommonC5ENS0_21ExternalBufferSharingE]+0x1b): undefined reference to `apache::thrift::detail::ProtocolBase::ProtocolBase()'
/usr/bin/ld: /usr/local/lib/libtransport.a(THeader.cpp.o): in function `apache::thrift::JSONProtocolReaderCommon::JSONProtocolReaderCommon(apache::thrift::ExternalBufferSharing)':
THeader.cpp:(.text._ZN6apache6thrift24JSONProtocolReaderCommonC2ENS0_21ExternalBufferSharingE[_ZN6apache6thrift24JSONProtocolReaderCommonC5ENS0_21ExternalBufferSharingE]+0x1b): undefined reference to `apache::thrift::detail::ProtocolBase::ProtocolBase()'
/usr/bin/ld: /usr/local/lib/librpcmetadata.a(RpcMetadata_types_serialization.cpp.o): in function `apache::thrift::SimpleJSONProtocolReader::skip(apache::thrift::protocol::TType, int)':
RpcMetadata_types_serialization.cpp:(.text._ZN6apache6thrift24SimpleJSONProtocolReader4skipENS0_8protocol5TTypeEi[_ZN6apache6thrift24SimpleJSONProtocolReader4skipENS0_8protocol5TTypeEi]+0x2b): undefined reference to `fLI::FLAGS_thrift_protocol_max_depth'
/usr/bin/ld: /usr/local/lib/librpcmetadata.a(RpcMetadata_types_serialization.cpp.o): in function `void apache::thrift::skip_n<apache::thrift::SimpleJSONProtocolReader, apache::thrift::protocol::TType>(apache::thrift::SimpleJSONProtocolReader&, unsigned int, std::initializer_list<apache::thrift::protocol::TType>, int)':
RpcMetadata_types_serialization.cpp:(.text._ZN6apache6thrift6skip_nINS0_24SimpleJSONProtocolReaderENS0_8protocol5TTypeEEEvRT_jSt16initializer_listIT0_Ei[_ZN6apache6thrift6skip_nINS0_24SimpleJSONProtocolReaderENS0_8protocol5TTypeEEEvRT_jSt16initializer_listIT0_Ei]+0x22): undefined reference to `fLI::FLAGS_thrift_protocol_max_depth'
```

kou:
Hmm. `velox_exec_test` crashed in `HashJoinTest.taskWaitTimeout`: https://github.com/facebookincubator/velox/blob/198082aeee10b4fd96666012f1cd43cc6e3e60be/velox/exec/tests/HashJoinTest.cpp#L7052

https://github.com/facebookincubator/velox/actions/runs/18129265004/job/51591711191?pr=14942#step:10:1920

```text
[ RUN      ] HashJoinTest.taskWaitTimeout
E0930 13:09:27.191798 218477 Exceptions.h:53] Line: /home/runner/work/velox/velox/velox/velox/exec/Task.cpp:3545, Function:reclaimTask, Expression:  Memory reclaim failed to wait for task test_cursor_549 to pause after 999.10ms with max timeout 999ms, Source: RUNTIME, ErrorCode: INVALID_STATE
E0930 13:09:27.192206 218477 Exceptions.h:53] Line: /home/runner/work/velox/velox/velox/velox/exec/Task.cpp:1164, Function:resume, Expression: !driver->isOnThread() && !driver->isTerminated() , Source: RUNTIME, ErrorCode: INVALID_STATE
E0930 13:09:27.192500 218477 ArbitrationParticipant.cpp:303] [MEM] Failed to reclaim from memory pool default_root_418, aborting it: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Memory reclaim failed to wait for task test_cursor_549 to pause after 999.10ms with max timeout 999ms
Retriable: False
Context: Operator: HashBuild[4] 2
Function: reclaimTask
File: /home/runner/work/velox/velox/velox/velox/exec/Task.cpp
Line: 3545
Stack trace:
...

*** Aborted at 1759237772 (Unix time, try 'date -d @1759237772') ***
*** Signal 6 (SIGABRT) (0x3e90001d51e) received by PID 120094 (pthread TID 0x7f13237ff640) (linux TID 219906) (maybe from PID 120094, UID 1001) (code: -6), stack trace: ***
    @ 0000000018d2c7a8 _ZN5folly10symbolizer12_GLOBAL__N_118innerSignalHandlerEiP9siginfo_tPv
                       /home/runner/work/velox/velox/velox/deps-download/folly/folly/debugging/symbolizer/SignalHandler.cpp:485
    @ 0000000018d2c88e _ZN5folly10symbolizer12_GLOBAL__N_113signalHandlerEiP9siginfo_tPv
                       /home/runner/work/velox/velox/velox/deps-download/folly/folly/debugging/symbolizer/SignalHandler.cpp:506
    @ 000000000004251f (unknown)
    @ 00000000000969fc pthread_kill
    @ 0000000000042475 raise
    @ 00000000000287f2 abort
    @ 00000000000a2b9d (unknown)
    @ 00000000000ae20b (unknown)
    @ 00000000000ae276 _ZSt9terminatev
    @ 00000000000ae1fd _ZSt17rethrow_exceptionNSt15__exception_ptr13exception_ptrE
    @ 0000000018d2454d _ZSt17rethrow_exceptionNSt15__exception_ptr13exception_ptrE
                       /home/runner/work/velox/velox/velox/deps-download/folly/folly/debugging/exception_tracer/ExceptionTracerLib.cpp:244
    @ 0000000010f78bbe _ZN8facebook5velox4exec23MultiThreadedTaskCursor14checkTaskErrorEv
                       /home/runner/work/velox/velox/velox/velox/exec/Cursor.cpp:360
    @ 0000000010f788c8 _ZN8facebook5velox4exec23MultiThreadedTaskCursor8moveNextEv
                       /home/runner/work/velox/velox/velox/velox/exec/Cursor.cpp:311
    @ 000000000e612efa _ZN8facebook5velox4exec4test10readCursorERKNS1_16CursorParametersESt8functionIFvPNS1_10TaskCursorEEEm
                       /home/runner/work/velox/velox/velox/velox/exec/tests/utils/QueryAssertions.cpp:1349
    @ 000000000e5442a0 _ZN8facebook5velox4exec4test18AssertQueryBuilder10readCursorEv
                       /home/runner/work/velox/velox/velox/velox/exec/tests/utils/AssertQueryBuilder.cpp:359
    @ 000000000e543170 _ZN8facebook5velox4exec4test18AssertQueryBuilder11copyResultsEPNS0_6memory10MemoryPoolERSt10shared_ptrINS1_4TaskEE
                       /home/runner/work/velox/velox/velox/velox/exec/tests/utils/AssertQueryBuilder.cpp:245
    @ 000000000e54be89 _ZN8facebook5velox4exec4test15runHashJoinTaskERKSt6vectorISt10shared_ptrINS0_9RowVectorEESaIS6_EERKS4_INS0_4core8QueryCtxEEbjPNS0_6memory10MemoryPoolEbRKS6_
                       /home/runner/work/velox/velox/velox/velox/exec/tests/utils/ArbitratorTestUtil.cpp:116
    @ 000000000d9a1c4c _ZZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvENKUlvE1_clEv
                       /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp:7117
    @ 000000000da4953a _ZSt13__invoke_implIvZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvEUlvE1_JEET_St14__invoke_otherOT0_DpOT1_
                       /usr/include/c++/11/bits/invoke.h:61
                       -> /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp
    @ 000000000da491e8 _ZSt8__invokeIZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvEUlvE1_JEENSt15__invoke_resultIT_JDpT0_EE4typeEOS7_DpOS8_
                       /usr/include/c++/11/bits/invoke.h:96
                       -> /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp
    @ 000000000da48de5 _ZNSt6thread8_InvokerISt5tupleIJZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvEUlvE1_EEE9_M_invokeIJLm0EEEEvSt12_Index_tupleIJXspT_EEE
                       /usr/include/c++/11/bits/std_thread.h:259
                       -> /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp
    @ 000000000da47b89 _ZNSt6thread8_InvokerISt5tupleIJZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvEUlvE1_EEEclEv
                       /usr/include/c++/11/bits/std_thread.h:266
                       -> /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp
    @ 000000000da41a43 _ZNSt6thread11_State_implINS_8_InvokerISt5tupleIJZN8facebook5velox4exec12_GLOBAL__N_133HashJoinTest_taskWaitTimeout_Test8TestBodyEvEUlvE1_EEEEE6_M_runEv
                       /usr/include/c++/11/bits/std_thread.h:211
                       -> /home/runner/work/velox/velox/velox/velox/exec/tests/HashJoinTest.cpp
    @ 00000000000dc252 (unknown)
    @ 0000000000094ac2 (unknown)
    @ 00000000001268bf (unknown)
Fatal signal handler. ThreadDebugInfo object not found.
```

* It seems that the hash join task raises an exception with large timeout
* It seems that signal handler by Folly is crashed to print backtrace for the (unhandled) exception

majetideepak:
@kou Is this ready for review?

.github/workflows/linux-build-base.yml
```diff
@@ -170,7 +170,18 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          source scripts/setup-ubuntu.sh && install_apt_deps && install_faiss_deps
+          source scripts/setup-ubuntu.sh
+          install_apt_deps
+          install_faiss_deps
```
kou, line 258: We don't need to use `&&` because `set -e` is enabled by default.

CMake/third-party/FBThriftCppLibrary.cmake
```diff
@@ -59,6 +59,8 @@ function(add_fbthrift_cpp_library LIB_NAME THRIFT_FILE)
     "${output_dir}/gen-cpp2/${base}_data.h"
     "${output_dir}/gen-cpp2/${base}_data.cpp"
     "${output_dir}/gen-cpp2/${base}_types.cpp"
+    "${output_dir}/gen-cpp2/${base}_types_binary.cpp"
+    "${output_dir}/gen-cpp2/${base}_types_compact.cpp"
```
kou, line 73: Apache Parquet implementation uses `CompactProtocol` and remote protocol uses `BinaryProtocol`.

CMake/third-party/FBThriftCppLibrary.cmake
```diff
@@ -106,14 +108,7 @@ function(add_fbthrift_cpp_library LIB_NAME THRIFT_FILE)
...
 
-  # Now emit the library rule to compile the sources
-  if(BUILD_SHARED_LIBS)
-    set(LIB_TYPE SHARED)
-  else()
-    set(LIB_TYPE STATIC)
-  endif()
-
-  add_library("${LIB_NAME}" ${LIB_TYPE} ${generated_sources})
+  add_library("${LIB_NAME}" STATIC ${generated_sources})
```
kou, line 147: We always use static library. We may want to use `OBJECT` here.

CMake/VeloxUtils.cmake
```diff
@@ -153,7 +153,7 @@ function(velox_add_library TARGET)
         )
       endif()
     endif()
-    # create alias for compatability
+    # create alias for compatibility
```
kou: This is a needless change. This just fixes a typo.

scripts/setup-versions.sh
```diff
@@ -19,7 +19,7 @@
 # The versions should match the declared versions in this file.
 
 # Build dependencies versions.
-FB_OS_VERSION="v2025.04.28.00"
+FB_OS_VERSION="v2025.09.15.00"
```
kou: We need recent FBThrift to use the `@thrift.AllowUnsafeOptionalCustomDefaultValue` annotation.

velox/common/dynamic_registry/tests/CMakeLists.txt
```diff
@@ -24,40 +24,69 @@ add_library(velox_overload_int_function_dynamic SHARED DynamicIntFunctionOverloa
 add_library(velox_overload_varchar_function_dynamic SHARED DynamicVarcharFunctionOverload.cpp)
 add_library(velox_function_non_default_dynamic SHARED DynamicFunctionNonDefault.cpp)
 
-set(CMAKE_DYLIB_TEST_LINK_LIBRARIES fmt::fmt Folly::folly glog::glog xsimd)
+set(DYLIB_TEST_LINK_LIBRARIES fmt::fmt glog::glog xsimd)
```
kou: Removing the `CMAKE_` prefix because this is not a CMake provided variable.
This is not a required change. This just improves readability.

velox/dwio/parquet/reader/CMakeLists.txt
```diff
@@ -30,14 +30,12 @@ velox_add_library(
 
 velox_link_libraries(
   velox_dwio_native_parquet_reader
-  velox_dwio_parquet_thrift
```
kou, line 33: We don't need this because `velox_dwio_parquet_common` already depends on `velox_dwio_parquet_thrift`.

velox/dwio/parquet/writer/arrow/tests/SchemaTest.cpp
```diff
@@ -2152,45 +2148,6 @@ TEST(TestSchemaNodeCreation, FactoryExceptions) {
...
-  node = PrimitiveNode::Make(
-      "timestamp",
-      Repetition::REQUIRED,
-      TimestampLogicalType::Make(true, LogicalType::TimeUnit::NANOS),
-      Type::INT64);
-  facebook::velox::parquet::thrift::SchemaElement timestamp_intermediary;
-  node->ToParquet(&timestamp_intermediary);
-  // ... corrupt the Thrift intermediary ....
-  timestamp_intermediary.logicalType.TIMESTAMP.unit.__isset.NANOS = false;
-  ASSERT_ANY_THROW(PrimitiveNode::FromParquet(&timestamp_intermediary));
```
kou, line 2231: We can't create broken Thrift data with FBThrift API.

velox/dwio/parquet/writer/arrow/CMakeLists.txt
```diff
@@ -45,7 +45,6 @@ velox_link_libraries(
   velox_dwio_arrow_parquet_writer_lib
   velox_dwio_arrow_parquet_writer_util_lib
   velox_dwio_parquet_common
-  velox_dwio_parquet_thrift
```
kou, line 48: We don't need this because `velox_dwio_parquet_common` already depends on `velox_dwio_parquet_thrift`.

velox/exec/tests/HashJoinTest.cpp
```diff
@@ -7065,7 +7065,7 @@ DEBUG_ONLY_TEST_F(HashJoinTest, taskWaitTimeout) {
   const auto expectedResult =
       runHashJoinTask(vectors, nullptr, false, numDrivers, pool(), false).data;
 
-  for (uint64_t timeoutMs : {1'000, 30'000}) {
+  for (uint64_t timeoutMs : {1'000, 60'000}) {
```
kou: I don't know why but 30 secs isn't enough with this change on CI machine (`8-core-ubuntu-22.04`). (This isn't needed on my local machine.)

`v2025.09.15.00` FB OS is slower than `v2025.04.28.00`?

I think that FBThrift isn't related because hash join doesn't use Parquet.

velox/exec/tests/HashJoinTest.cpp
```diff
@@ -7113,8 +7113,17 @@ DEBUG_ONLY_TEST_F(HashJoinTest, taskWaitTimeout) {
             "Memory reclaim failed to wait");
       } else {
         // We expect succeed on large time out or no timeout.
-        const auto result = runHashJoinTask(
-            vectors, queryCtx, false, numDrivers, pool(), true, expectedResult);
+        QueryTestResult result;
+        ASSERT_NO_THROW({
```
kou, line 7345: If we don't have this and timeout happens, the test is crashed.

CMakeLists.txt
```diff
@@ -389,9 +389,9 @@ endif()
 message("Setting CMAKE_CXX_FLAGS=${SCRIPT_CXX_FLAGS}")
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SCRIPT_CXX_FLAGS}")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D USE_VELOX_COMMON_BASE")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_VELOX_COMMON_BASE")
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D HAS_UNCAUGHT_EXCEPTIONS")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DHAS_UNCAUGHT_EXCEPTIONS")
```
kou, line 409: They are needless changes. They are just cosmetic changes.

velox/common/dynamic_registry/tests/CMakeLists.txt
assignUser: Shouldn't this just be an interface target?

velox/dwio/parquet/thrift/CMakeLists.txt
assignUser: I assume you mean because `velox_link_libraries` drops the link to `velox..._raw`? You can add it [here](https://github.com/facebookincubator/velox/blob/main/CMake/VeloxUtils.cmake#L174) to override that behavior.

assignUser:
Thanks for pushing this forward @kou!

velox/common/dynamic_registry/tests/CMakeLists.txt
kou: OK. I'll try it.

velox/dwio/parquet/thrift/CMakeLists.txt
kou: I want to use `INTERFACE` here... (The latest code is changed from this.)

velox/common/dynamic_registry/tests/CMakeLists.txt
kou: Done.

pedroerp:
Wow, can't believe this is actually happening, thanks @kou!

Any chance we could split this PR? It ended up getting very large and hard to review. One suggestion could be to first submit the C++ changes to move to fbthrift, then remove the thrift hookups from CI and build systems in a follow up.

kou:
> Any chance we could split this PR? It ended up getting very large and hard to review. One suggestion could be to first submit the C++ changes to move to fbthrift, then remove the thrift hookups from CI and build systems in a follow up.

Yes. But the large part of this PR is the C++ changes part. We can't mix FBThrift and Apache Thrift. So we can't split the C++ changes part... Is it acceptable? (The C++ changes part may still not be reviewable...)

My PR comments mark important parts in the C++ changes part. Other parts are  consistent conversions described in the PR description. This note may help your review.

(The Apache Thrift hookups part is small.)

JkSelf:
> Thanks @kou. The changes look great! I will run the TPC-H velox benchmarks over Parquet data just to ensure there is no perf regression.
> 
> CC: @JkSelf, @rui-mo who work on Gluten CC: @jaystarshot at Uber

@majetideepak @rui-mo I can help to verify the perf in TPC-H and TPC-DS in Gluten. Thanks.

JkSelf:
@majetideepak @kou @pedroerp 

I tried running the 1TB TPC-DS benchmark with this PR, but encountered the following exception:
```
Caused by: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Operator::getOutput failed for [operator: FilterProject, plan node ID: 1]: underflow
Retriable: False
Function: operator()
File: /mnt/DP_disk3/jk/projects/gluten/ep/build-velox/build/velox_ep/velox/exec/Driver.cpp
Line: 668
Stack trace:
# 0  facebook::velox::VeloxException::VeloxException(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)
# 1  void facebook::velox::detail::veloxCheckFail<facebook::velox::VeloxRuntimeError, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(facebook::velox::detail::VeloxCheckFailArgs const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
# 2  facebook::velox::exec::Driver::runInternal(std::shared_ptr<facebook::velox::exec::Driver>&, std::shared_ptr<facebook::velox::exec::BlockingState>&, std::shared_ptr<facebook::velox::RowVector>&)::{lambda()#10}::operator()() const [clone .cold]
# 3  facebook::velox::exec::Driver::runInternal(std::shared_ptr<facebook::velox::exec::Driver>&, std::shared_ptr<facebook::velox::exec::BlockingState>&, std::shared_ptr<facebook::velox::RowVector>&)
# 4  facebook::velox::exec::Driver::next(folly::SemiFuture<folly::Unit>*, facebook::velox::exec::Operator*&, facebook::velox::exec::BlockingReason&)
# 5  facebook::velox::exec::Task::next(folly::SemiFuture<folly::Unit>*)
# 6  gluten::WholeStageResultIterator::nextInternal()
# 7  gluten::WholeStageResultIterator::next()
# 8  Java_org_apache_gluten_vectorized_ColumnarBatchOutIterator_nativeHasNext
# 9  0x00007fb8cd60ff2f
# 10 0x00007fb8cd96e203

        at org.apache.gluten.vectorized.ColumnarBatchOutIterator.nativeHasNext(Native Method)
        at org.apache.gluten.vectorized.ColumnarBatchOutIterator.hasNext0(ColumnarBatchOutIterator.java:58)
        at org.apache.gluten.iterator.ClosableIterator.hasNext(ClosableIterator.java:36)
        ... 35 more

        at org.apache.gluten.iterator.ClosableIterator.hasNext(ClosableIterator.java:38)
        at scala.collection.convert.Wrappers$JIteratorWrapper.hasNext(Wrappers.scala:45)
        at org.apache.gluten.iterator.IteratorsV1$InvocationFlowProtection.hasNext(IteratorsV1.scala:154)
        at org.apache.gluten.iterator.IteratorsV1$ReadTimeAccumulator.hasNext(IteratorsV1.scala:122)
        at org.apache.gluten.iterator.IteratorsV1$IteratorCompleter.hasNext(IteratorsV1.scala:66)
        at org.apache.gluten.iterator.IteratorsV1$PayloadCloser.hasNext(IteratorsV1.scala:38)
        at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
        at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
        at org.apache.spark.shuffle.ColumnarShuffleWriter.internalWrite(ColumnarShuffleWriter.scala:140)
        at org.apache.spark.shuffle.ColumnarShuffleWriter.write(ColumnarShuffleWriter.scala:287)
        at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:104)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:54)
        at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:166)
```

And the core dump stack is in native libhdfs.so. 
```
RAX=0x00007f6ce64deff6 is an unknown value
RBX=0x00007f6ce5f14040 is an unknown value
RCX=0x00007f6cb69b03f1: <offset 0xc3f1> in /mnt/DP_disk3/jk/binary/hadoop-3.3.0/lib/native/libhdfs.so at 0x00007f6cb69a4000
RDX=0x00007f6cb69b076d: <offset 0xc76d> in /mnt/DP_disk3/jk/binary/hadoop-3.3.0/lib/native/libhdfs.so at 0x00007f6cb69a4000
RSP=0x00007f6cb591dab8 is an unknown value
RBP=0x00007f6cb69b076d: <offset 0xc76d> in /mnt/DP_disk3/jk/binary/hadoop-3.3.0/lib/native/libhdfs.so at 0x00007f6cb69a4000
RSI=0x00007f6d06d86fd8 is a global jni handle
RDI=0x00007f6ce5f14040 is an unknown value
R8 =0x0000000000000001 is an unknown value
R9 =0x00007f6ce5f14040 is an unknown value
R10=0x0000000000000008 is an unknown value
R11=0x0000000000000246 is an unknown value
R12=0x00007f6cb69b03f1: <offset 0xc3f1> in /mnt/DP_disk3/jk/binary/hadoop-3.3.0/lib/native/libhdfs.so at 0x00007f6cb69a4000
R13=0x00007f6cb69b2a20: <offset 0xea20> in /mnt/DP_disk3/jk/binary/hadoop-3.3.0/lib/native/libhdfs.so at 0x00007f6cb69a4000
R14=0x00007f6cb591db10 is an unknown value
R15=0x00007f6ce5f14040 is an unknown value
```

I have a question regarding this PR: Since Apache Hadoop uses Apache Thrift, and we're switching to FB Thrift in Velox, there may be API incompatibilities between Apache Thrift and FB Thrift. I believe this could be causing the above exception.

What are your thoughts on this?"

kou:
I'm not familiar with Apache Hadoop but Apache Hadoop uses Apache Thrift in Java not C++, right?
It seems that `libhdfs.so` https://github.com/apache/hadoop/tree/trunk/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs calls Java API instead of implement HDFS in C++.
So Apache Thrift and FBThrift incompatibility may not be related...

jaystarshot:
Parquet reader and writer use some arrow functionality in velox, would that part of arrow also be compatible?

kou:
Apache Arrow version isn't changed in this PR.

jaystarshot:
I mean arrow internally uses apache thrift ( atleast for parquet )so will that flow be compatible with fbthrift in the current arrow version?
Also what about future arrow upgrades is it guaranteed to be compatible with fbthrift.

kou:
I'm not sure whether this answers your question but...

* Velox forks the Apache Parquet implementation in Apache Arrow
* The forked Apache Parquet implementation uses Apache Arrow
* Apache Arrow uses Apache Thrift only for Apache Parquet. The Apache Arrow part used by the forked Apache Parquet implementation doesn't depend on Apache Thrift.

So we don't need to care about Apache Thrift/FBThrift compatibility for Apache Arrow. (We don't use Apache Thrift related code in Apache Arrow.)

Does this answer your question...?

JkSelf:
> I'm not familiar with Apache Hadoop but Apache Hadoop uses Apache Thrift in Java not C++, right? It seems that `libhdfs.so` https://github.com/apache/hadoop/tree/trunk/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfs calls Java API instead of implement HDFS in C++. So Apache Thrift and FBThrift incompatibility may not be related...

@kou Thanks for your reply.
I haven't identified the root cause yet, but this PR does indeed cause the core dump issue. I will investigate further and provide an update. Thank you.

kou:
Thanks.

If we can show stack trace for an exception here https://github.com/facebookincubator/velox/blob/d187cb0fbb3622b00a044836c119ba8c8ee00b34/velox/exec/Driver.cpp#L331-L336 , it will help us to know what happened...

JkSelf:
@kou 
I used a try-catch approach to locate the code that throws the exception. I found that the exception is triggered during the deserialization of the Thrift header [here](https://github.com/facebookincubator/velox/pull/14942/files#diff-58ef2bd79ff4173b5a384a3016d6f7bd7d4b237876251cc9d9910e162993eaaaR99-R103). The Parquet file used in my test was generated by Spark using Apache Thrift. When converting the Apache Thrift header to Facebook Thrift, I suspect there might be some incompatibilities.

kou:
Thanks!!! Could you provide the Apache Parquet file that has a problem? I'll debug it on my local machine.

JkSelf:
> Thanks!!! Could you provide the Apache Parquet file that has a problem? I'll debug it on my local machine.

@kou You can have a try https://github.com/JkSelf/velox/commit/d815f3544a5d677338d2184a8f1de5dcf1d9d83d

kou:
Thanks! I'll try it!

kou:
@JkSelf I tried reading all column chunk metadata by the following code:

```diff
diff --git a/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp b/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
index 64047967f..fd4f8c249 100644
--- a/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
+++ b/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
@@ -1789,3 +1789,23 @@ TEST_F(ParquetReaderTest, readerWithSchema) {
 
   EXPECT_EQ(reader.rowType()->toString(), schema->toString());
 }
+
+TEST_F(ParquetReaderTest, XXX) {
+  const std::string sample(getExampleFilePath("crash.parquet"));
+  dwio::common::ReaderOptions readerOptions{leafPool_.get()};
+  auto reader = createReader(sample, readerOptions);
+  auto type = reader->typeWithId();
+  auto fileMetaData = reader->fileMetaData();
+  for (int i = 0; i < fileMetaData.numRowGroups(); ++i) {
+    auto rowGroupMetaData = fileMetaData.rowGroup(i);
+    for (int j = 0; j < rowGroupMetaData.numColumns(); ++j) {
+      auto columnChunkMetaData = rowGroupMetaData.columnChunk(j);
+      if (columnChunkMetaData.hasStatistics()) {
+        auto columnStatistics = columnChunkMetaData.getColumnStatistics(type->childAt(j)->type(), rowGroupMetaData.numRows());
+        columnChunkMetaData.getColumnMetadataStatsMinValue();
+        columnChunkMetaData.getColumnMetadataStatsMaxValue();
+        columnChunkMetaData.getColumnMetadataStatsNullCount();
+      }
+    }
+  }
+}
```

But I couldn't reproduce the exception...

If you can use debugger, could you try getting backtrace for the exception?

JkSelf:
@kou Here is the stacktrace:
```

0# facebook::velox::parquet::PageReader::readPageHeader() in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 1# facebook::velox::parquet::PageReader::seekToPage(long) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 2# facebook::velox::parquet::PageReader::rowsForPage(facebook::velox::dwio::common::SelectiveColumnReader&, bool, bool, folly::Range<int const*>&, unsigned long const*&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 3# void facebook::velox::parquet::PageReader::readWithVisitor<facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true> >(facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true>&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 4# void facebook::velox::parquet::ParquetData::readWithVisitor<facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true> >(facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true>) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 5# void facebook::velox::parquet::IntegerColumnReader::readWithVisitor<facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true> >(folly::Range<int const*> const&, facebook::velox::dwio::common::ColumnVisitor<int, facebook::velox::common::IsNotNull, facebook::velox::dwio::common::ExtractToReader, true, true>) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 6# void facebook::velox::dwio::common::SelectiveIntegerColumnReader::readHelper<facebook::velox::parquet::IntegerColumnReader, facebook::velox::common::IsNotNull, true, facebook::velox::dwio::common::ExtractToReader>(facebook::velox::common::Filter const*, folly::Range<int const*> const&, facebook::velox::dwio::common::ExtractToReader) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 7# void facebook::velox::dwio::common::SelectiveIntegerColumnReader::processFilter<facebook::velox::parquet::IntegerColumnReader, true, true, facebook::velox::dwio::common::ExtractToReader>(facebook::velox::common::Filter const*, facebook::velox::dwio::common::ExtractToReader, folly::Range<int const*> const&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 8# void facebook::velox::dwio::common::SelectiveIntegerColumnReader::readCommon<facebook::velox::parquet::IntegerColumnReader, true>(folly::Range<int const*> const&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
 9# facebook::velox::parquet::IntegerColumnReader::read(long, folly::Range<int const*> const&, unsigned long const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
10# facebook::velox::dwio::common::SelectiveStructColumnReaderBase::read(long, folly::Range<int const*> const&, unsigned long const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
11# facebook::velox::parquet::StructColumnReader::read(long, folly::Range<int const*> const&, unsigned long const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
12# facebook::velox::dwio::common::SelectiveStructColumnReaderBase::next(unsigned long, std::shared_ptr<facebook::velox::BaseVector>&, facebook::velox::dwio::common::Mutation const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
13# facebook::velox::parquet::ParquetRowReader::Impl::next(unsigned long, std::shared_ptr<facebook::velox::BaseVector>&, facebook::velox::dwio::common::Mutation const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
14# facebook::velox::parquet::ParquetRowReader::next(unsigned long, std::shared_ptr<facebook::velox::BaseVector>&, facebook::velox::dwio::common::Mutation const*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
15# facebook::velox::connector::hive::SplitReader::next(unsigned long, std::shared_ptr<facebook::velox::BaseVector>&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
16# facebook::velox::connector::hive::HiveDataSource::next(unsigned long, folly::SemiFuture<folly::Unit>&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
17# facebook::velox::exec::TableScan::getOutput() in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
18# 0x00007EFD9AEA1660 in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
19# 0x00007EFD9AEA2258 in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
20# 0x00007EFD9AEAA294 in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
21# facebook::velox::exec::Driver::runInternal(std::shared_ptr<facebook::velox::exec::Driver>&, std::shared_ptr<facebook::velox::exec::BlockingState>&, std::shared_ptr<facebook::velox::RowVector>&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
22# facebook::velox::exec::Driver::next(folly::SemiFuture<folly::Unit>*, facebook::velox::exec::Operator*&, facebook::velox::exec::BlockingReason&) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
23# facebook::velox::exec::Task::next(folly::SemiFuture<folly::Unit>*) in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
24# gluten::WholeStageResultIterator::nextInternal() in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
25# gluten::WholeStageResultIterator::next() in /mnt/DP_disk3/jk/projects/gluten/cpp/build/releases/libvelox.so
26# gluten::ResultIterator::getNext() in ./cpp/build/velox/benchmarks/generic_benchmark
27# gluten::ResultIterator::hasNext() in ./cpp/build/velox/benchmarks/generic_benchmark
28# 0x0000558210EAEE2C in ./cpp/build/velox/benchmarks/generic_benchmark

```

kou:
Thanks! Could you use `-DCMAKE_BUILD_TYPE=Debug` not `-DCMAKE_BUILD_TYPE=Release` to show filename and line number in the backtrace?

kou:
(I'll try using `ParquetRowReader`.)

kou:
I tried this and confirmed that this calls `facebook::velox::parquet::PageReader::readPageHeader()` but I couldn't reproduce this. Hmm...

```diff
diff --git a/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp b/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
index 64047967f..e83cb3411 100644
--- a/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
+++ b/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
@@ -1789,3 +1789,44 @@ TEST_F(ParquetReaderTest, readerWithSchema) {
 
   EXPECT_EQ(reader.rowType()->toString(), schema->toString());
 }
+
+TEST_F(ParquetReaderTest, XXX) {
+  const std::string sample(getExampleFilePath("crash.parquet"));
+  dwio::common::ReaderOptions readerOptions{leafPool_.get()};
+  auto reader = createReader(sample, readerOptions);
+  auto type = reader->typeWithId();
+  auto fileMetaData = reader->fileMetaData();
+  auto rowType = ROW(
+    {"ws_sold_time_sk","ws_ship_date_sk","ws_item_sk","ws_bill_customer_sk","ws_bill_cdemo_sk","ws_bill_hdemo_sk","ws_bill_addr_sk","ws_ship_customer_sk","ws_ship_cdemo_sk","ws_ship_hdemo_sk","ws_ship_addr_sk","ws_web_page_sk","ws_web_site_sk","ws_ship_mode_sk","ws_warehouse_sk","ws_promo_sk","ws_order_number","ws_quantity","ws_wholesale_cost","ws_list_price","ws_sales_price","ws_ext_discount_amt","ws_ext_sales_price","ws_ext_wholesale_cost","ws_ext_list_price","ws_ext_tax","ws_coupon_amt","ws_ext_ship_cost","ws_net_paid","ws_net_paid_inc_tax","ws_net_paid_inc_ship","ws_net_paid_inc_ship_tax","ws_net_profit"},
+    {INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),BIGINT(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER(),INTEGER()}
+    );
+  RowReaderOptions rowReaderOpts;
+  auto scanSpec = makeScanSpec(rowType);
+  for (const auto &name : rowType->names()) {
+    scanSpec->
+      getOrCreateChild(facebook::velox::common::Subfield(name))->
+      setFilter(exec::lessThanOrEqual(std::numeric_limits<int32_t>::max()));
+  }
+  rowReaderOpts.setScanSpec(std::move(scanSpec));
+  auto rowReader = reader->createRowReader(rowReaderOpts);
+  auto parquetRowReader = dynamic_cast<ParquetRowReader*>(rowReader.get());
+  constexpr int kBatchSize = 1000;
+  auto result = BaseVector::create(rowType, kBatchSize, pool_.get());
+  for (int i = 0; i < fileMetaData.numRowGroups(); ++i) {
+    auto rowGroupMetaData = fileMetaData.rowGroup(i);
+    for (int j = 0; j < rowGroupMetaData.numColumns(); ++j) {
+      auto columnChunkMetaData = rowGroupMetaData.columnChunk(j);
+      if (columnChunkMetaData.hasStatistics()) {
+        auto columnStatistics = columnChunkMetaData.getColumnStatistics(type->childAt(j)->type(), rowGroupMetaData.numRows());
+        columnChunkMetaData.getColumnMetadataStatsMinValue();
+        columnChunkMetaData.getColumnMetadataStatsMaxValue();
+        columnChunkMetaData.getColumnMetadataStatsNullCount();
+      }
+    }
+
+    while (parquetRowReader->next(kBatchSize, result) > 0) {
+      // Do nothing
+    }
+    parquetRowReader->nextRowNumber();
+  }
+}
```

majetideepak:
>releases/libvelox.so

Gluten uses a shared library for Velox. My hunch is that they are building both thrift and fbthrift together.
@JkSelf Can you confirm this. Do you still use apache thrift in the build for Arrow or something else?

JkSelf:
@majetideepak The Arrow compilation still uses Apache Thrift. My question is whether Arrow supports compiling with Facebook Thrift (fbthrift) [here](https://github.com/apache/arrow/blob/ac9707663c160599fc46925a9d15d2cf030d110b/cpp/cmake_modules/ThirdpartyToolchain.cmake#L1633-L1709)?

JkSelf:
> Thanks! Could you use `-DCMAKE_BUILD_TYPE=Debug` not `-DCMAKE_BUILD_TYPE=Release` to show filename and line number in the backtrace?

Yes. My compile is debug mode.

`libvelox.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=ba14c4b9666dc6ae4937038751b8931fe8778c70, with debug_info, not stripped`

kou:
I'm an Apache Arrow developer. How did you build Apache Arrow? If Apache Arrow isn't built with Apache Parquet support, the bundled Apache Thrift isn't used.

kou:
> Yes. My compile is debug mode.

Oh! How did you get the backtrace? `gdb`? Could you show source location too? In `gdb`, I think that `bt` shows source location too.

JkSelf:
> I'm an Apache Arrow developer. How did you build Apache Arrow? If Apache Arrow isn't built with Apache Parquet support, the bundled Apache Thrift isn't used.

@kou You can refer https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh.

JkSelf:
> > Yes. My compile is debug mode.
> 
> Oh! How did you get the backtrace? `gdb`? Could you show source location too? In `gdb`, I think that `bt` shows source location too.

@kou I use `boost::stacktrace::stacktrace(); ` to print.

kou:
> You can refer https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh.

Thanks. Could you remove the followings?

* https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh#L56
* https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh#L71-L73

kou:
> I use `boost::stacktrace::stacktrace(); ` to print.

Hmm. The example in https://www.boost.org/doc/libs/latest/doc/html/stacktrace/getting_started.html#stacktrace.getting_started.how_to_print_current_call_stack shows source location...

Could you try `gdb` instead...?

majetideepak:
@JkSelf As @kou mentioned, you only need Thrift in Arrow for Parquet support. I believe Gluten uses the Arrow Parquet reader for benchmarking only. So it need not be part of the core library. Can you remove the thrift build in Arrow as @kou shared here https://github.com/facebookincubator/velox/pull/14942#issuecomment-3466493041 and see if you still see a crash?

JkSelf:
> > You can refer https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh.
> 
> Thanks. Could you remove the followings?
> 
> * https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh#L56
> * https://github.com/apache/incubator-gluten/blob/main/dev/build-arrow.sh#L71-L73

@kou @majetideepak I tried to remove the apache thrift dependency when compiling arrow. The issue still exists.

JkSelf:
@kou @majetideepak This issue can be reproduced within this [commit](https://github.com/JkSelf/velox/commit/5fcfd350881afc62a918ebc0d8d3c75fe3e817c9). It appears there are differences between Apache Thrift and fbthrift in how the page header is handled.

kou:
Thanks!!! I'll try it next week!

kou:
Thanks! I could reproduce this problem on my local machine.

I think that I understand this problem. In the Apache Thrift based implementation, we implement `ThriftTransport` that reads more data from IO. But FBThrift doesn't have a hook like it. So I removed it from the FBThrift based implementation. (The FBThrift based implementation reads data before we parse Thrift data.) But it seems that it's needed. I'm finding FBThrift API to do it... If we can customize `Folly::Cursor`/`Folly::IOBuf`, we can do it but we can't customize them...

kou:
Ah, we may be able to use `apache::thrift::CompactProtocolReaderWithRefill` instead of `apache::thrift::CompactProtocolReader` for this case. I'll try it.

kou:
@JkSelf I've migrated to `apache::thrift::CompactProtocolReaderWithRefill` based implementation. Could you try again? The test you provided works on my environment with the change.

kgpai:
cc: @JkSelf Does this work on Gluten, is there any blocker ?

JkSelf:
@kgpai I encountered a hang issue while running the 500GB Q94 TPC-DS benchmark. The failure data can be found [here](https://github.com/JkSelf/velox/commit/8306024dc626e76097becd0c5e12427c2fc5c0de). @kou, could you please take a look when you have a chance? Thanks.

peterenescu:
FYI likely blocks https://github.com/facebookincubator/velox/pull/15483 due to mismatched thrift libraries that causes Spark fuzzer crashes. https://github.com/facebookincubator/velox/issues/15916 may also be relevant
cc: @kgpai @majetideepak

peterenescu:
@kou were you able to take a look at and resolve the Gluten issue? I think there's also a fuzzer build error. I've began to import this PR internally and resolved all of the internal dependency and conflicts, so should be good to land this whenever the above two issues are resolved.
cc: @JkSelf

kou:
Sorry. I haven't enough bandwidth to take a look at the Gluten issue yet.

For the Fuzzer job failure, it seems that old FBThrift is used in the job:

https://github.com/facebookincubator/velox/actions/runs/19313090763/job/55238057140?pr=14942#step:14:1538

```text
[ERROR:/__w/velox/velox/velox/velox/dwio/parquet/thrift/parquet.thrift:589] Type `thrift.AllowUnsafeOptionalCustomDefaultValue` not defined.
```

peterenescu:
@kou I've resolved that in a fork of your change https://github.com/facebookincubator/velox/actions/runs/21049529120/job/60532317584?pr=16019
Looks like the error now is due to unfound cmake files, which is curious because only fuzzer build fails with this error
```
cc1plus: fatal error: /__w/velox/velox/velox/_build/debug/velox/dwio/parquet/thrift/gen-cpp2/parquet_types_binary.cpp: No such file or directory
compilation terminated.
```
Do you know if you'll bandwidth to investigate the Gluten issue?

kou:
It's strange. The file was generated in this PR: https://github.com/facebookincubator/velox/actions/runs/19313090875/job/55238057579?pr=14942#step:8:2370

```text
[963/3315] Building CXX object velox/dwio/parquet/thrift/CMakeFiles/velox_dwio_parquet_thrift_raw.dir/gen-cpp2/parquet_types_binary.cpp.o
```

Is the Fuzzer job really use the same FBThrift version?

> Do you know if you'll bandwidth to investigate the Gluten issue?

Sorry... I can't promise anything for now. At least I can't work on this in a few weeks. I have another tasks I need to work on.

peterenescu:
Is there a way to confirm? Removing 3 files from the cmake `generated_sources` list passes the fuzzer build
```
"${output_dir}/gen-cpp2/${base}_types_binary.cpp"
"${output_dir}/gen-cpp2/${base}_types_compact.cpp"
"${output_dir}/gen-cpp2/${base}_types_serialization.cpp"
```
I presume that means it's using the correct fbthrift as it's able to build the other files.

kou:
Hmm. Could you share how to update FBthrift in the Docker image that is used by the Fuzzer job?

peterenescu:
@kou I've added some changes, mainly branch conflicts fixes and fbthrift usage fixes to pass internal tests. Can you add me as a co-owner so I can publish changes and run CI?
cc: @kgpai

kou:
Thanks! I've invited you to my fork!

peterenescu:
Unfortunately still unable to push to this branch from internal, but that's okay. I just applied my merge fixes manually using the web editor, so hopefully that unblocks you for now. If for whatever reason there's an issue, I duplicated the code here: https://github.com/facebookincubator/velox/pull/16019.
cc: @kgpai

kou:
Oh... Could you share the error message from `git`?

peterenescu:
```
Peter Enescu (peterenescu) cannot export this diff because it is already associated with a pull request by another author: kou.
A diff can only be exported by the owner of the linked pull request. If you want to export to your own pull request, then first unlink the existing pull request
```
We can an internal tool to export from internal, which doesn't support force commit to different authors. One alternative is to use https://github.com/facebookincubator/velox/pull/16019 that you can override. Up to you, in any case we are blocked by Gluten regression so we can discuss after

kou:
Ah, I understand.

I'm OK that we continue this with #16019. I can close this in favor of #16019.

peterenescu:
Sounds good, sorry for the headache. https://github.com/facebookincubator/velox/pull/16019 is updated with all recent changes and merge conflicts resolved. Currently running CI changes. Feel free to take a look at your convenience.

kou:
No problem! I close this in favor of #16019.

PR #14952, opened 2025-09-25
docs: add ANSI compliant tag for Spark checked_arithmetic functions

zhli1142015:


velox/docs/functions/spark/math.rst
```diff
@@ -78,22 +78,22 @@ Mathematical Functions
     Returns ``x`` rounded up to the nearest integer.
     Supported types are: BIGINT, DOUBLE and DECIMAL.
 
-.. function:: checked_add(x, y) -> [same as x]
+.. function:: checked_add(x, y) -> [same as x] (ANSI compliant)
```
mbasmanova: Would you update PR description to explain why it is valid to add this without making any changes to the functions?

CC: @rui-mo

velox/docs/functions/spark/math.rst
```diff
@@ -78,22 +78,22 @@ Mathematical Functions
     Returns ``x`` rounded up to the nearest integer.
     Supported types are: BIGINT, DOUBLE and DECIMAL.
 
-.. function:: checked_add(x, y) -> [same as x]
+.. spark:function:: checked_add(x, y) -> [same as x] (ANSI compliant)
```
rui-mo, line 81: Does “ANSI compliant” mean that the function supports both ANSI ON and OFF behaviors and respects the ANSI configuration? If so, we could simply document that these functions are intended for ANSI ON mode. cc: @PHILO-HE

Issue #14943, opened 2025-09-24
Cudf initialize operator when driver initializing


jinchengchenghh:
### Bug description

After these two PRs https://github.com/facebookincubator/velox/pull/6395 https://github.com/facebookincubator/velox/pull/6409, it moves operator initialization from ctor to lazy initialization to avoid dead lock, and check if pool reservation bytes is 0 after driver initialization.

In cudf, manually initialize the operators by  [driver_.initializeOperators()](https://github.com/facebookincubator/velox/blob/main/velox/experimental/cudf/exec/ToCudf.cpp#L77) to get the expression information to decide if fallback, this will cause exception in unit test https://github.com/facebookincubator/velox/pull/14925

And I also meet similar exception in TPCDS Q67 test while window is not supported in cudf.
```
, Source: RUNTIME, ErrorCode: INVALID_STATE
25/09/24 13:33:16 ERROR TaskResources: Task 63 failed by error:
org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Unexpected memory pool allocations during task[Gluten_Stage_32_TID_63_VTID_42] driver initialization:
Top 1 leaf memory pool usages:
    op.2.0.0.Window usage 96.00KB reserved 1.00MB peak 96.00KB
 
root usage 96.00KB reserved 1.00MB peak 1.00MB
    task.Gluten_Stage_32_TID_63_VTID_42 usage 96.00KB reserved 1.00MB peak 1.00MB
        node.2 usage 96.00KB reserved 1.00MB peak 1.00MB
            op.2.0.0.Window usage 96.00KB reserved 1.00MB peak 96.00KB
 
 
Retriable: False
Function: next
/opt/gluten/ep/build-velox/build/velox_ep/velox/exec/Task.cpp
Line: 798
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
[...]

PR #6395, opened 2023-09-01
Move op initialization outside of FilterProject ctor path

tanjialiang:
Currently FilterProject operator initialization is in its constructor path. The constructor uses memory pool to compile expressions. This can lead to deadlock from memory arbitration in the following scenario:
1. T1: TaskManager::createOrUpdateTask()
2. T1: acquired PrestoTask::mutex_;
3. T1: acquired Task::mutex_;
4. T2: Something triggered arbitration, in arbitration calling Task::reclaim() trying to acquire Task::mutex_ but not able to acquire;
5. T1: triggered arbitration during FilterProject operator construction, got in arbitration queue, stuck in arbitration queue because current processing arbitration in T2 cannot proceed

DEADLOCK stuck in arbitration, and queue not able to move forward. Also PrestoTask mutex_ held forever.

The root cause of the above scenario is we hold Task::mutex_ while going into memory arbitration. We shall move the logic that uses memory pool out from the constructor to avoid such scenario.

PR #14925, opened 2025-09-22
feat(cudf): Support switch complex expression

jinchengchenghh:
We need to disable pool reserved bytes to avoid exception, so disable the test
```
query.TaskCursorQuery_0.0 usage 128B reserved 1.00MB peak 1.00MB
    task.test_cursor_1 usage 128B reserved 1.00MB peak 1.00MB
        node.1 usage 128B reserved 1.00MB peak 1.00MB
            op.1.0.0.FilterProject usage 128B reserved 1.00MB peak 128B


Retriable: False
Function: createAndStartDrivers
File: /opt/velox/velox/exec/Task.cpp
Line: 1021
```
And it is essential to register functions gt though it is in AST tree, otherwise, throws exception
```
Reason: Unsupported expression for recursive evaluation: gt
Retriable: False
Context: Operator: CudfFilterProject[1] 2
```

Some cudf API may not support the scalar, so we need to make the column, may need to add a common way like CudfFunction virtual function bool requestColumns(), in that case, convert the literal node to a column

PR #14956, opened 2025-09-25
fix(cudf): Fix initialize operator in driver creation

jinchengchenghh:
https://github.com/facebookincubator/velox/pull/6395 fixes a deadlock caused by allocating memory in driver creation, so we should not initialize operator in DriverAdapter.

Removed the driver_.initializeOperators().
Expose the filer node from FilterProject operator because when both project and filter exists, we can only get the project node id from op->planNodeId, then we cannot construct CudfFilterProject operator.
Move the CudfFilterProject initialization to function initialize().

Further more, if Cudf ExpressionEvaluator can get information from ITypedExpr, we can even remove compileExpression.

The cudf tests are broken, the test failed with or without this PR, other tests passed/
```
[ RUN      ] OrderByTest.singleKey
2: /opt/velox/velox/exec/tests/utils/QueryAssertions.cpp:1285: Failure
2: Failed
2: Expected keys: 999, actual: null
2: Note: DuckDB only supports timestamps of millisecond precision. If this test involves timestamp inputs, please make sure you use the right precision.
2: DuckDB query: SELECT * FROM tmp WHERE c0 % 2 >= 0 ORDER BY c0 DESC NULLS FIRST
```
Resolves: https://github.com/facebookincubator/velox/issues/14943

### DIFF
--- a/velox/experimental/cudf/exec/CudfFilterProject.cpp
+++ b/velox/experimental/cudf/exec/CudfFilterProject.cpp
@@ -154,8 +154,9 @@ void CudfFilterProject::initialize() {
     isIdentityProjection_ = true;
   }
 
-  auto lazyDereference = (dynamic_cast<const core::LazyDereferenceNode*>(project_.get()) !=
-          nullptr);
+  auto lazyDereference =
+      (dynamic_cast<const core::LazyDereferenceNode*>(project_.get()) !=
+       nullptr);
   VELOX_CHECK(!(lazyDereference && filter_));
   auto expr = exec::makeExprSetFromFlag(
       std::move(allExprs), operatorCtx_->execCtx(), lazyDereference);
@@ -178,8 +179,8 @@ void CudfFilterProject::initialize() {
     projectExprs = {expr->exprs().begin() + 1, expr->exprs().end()};
   }
 
-  projectEvaluator_ = ExpressionEvaluator(
-      hasFilter_ ? projectExprs : expr->exprs(), inputType);
+  projectEvaluator_ =
+      ExpressionEvaluator(hasFilter_ ? projectExprs : expr->exprs(), inputType);
 
   filter_.reset();
   project_.reset();
--- a/velox/experimental/cudf/exec/CudfFilterProject.h
+++ b/velox/experimental/cudf/exec/CudfFilterProject.h
@@ -68,7 +68,6 @@ class CudfFilterProject : public exec::Operator, public NvtxHelper {
   }
 
  private:
-
   bool allInputProcessed();
 
   // If true exprs_[0] is a filter and the other expressions are projections
--- a/velox/experimental/cudf/exec/ToCudf.cpp
+++ b/velox/experimental/cudf/exec/ToCudf.cpp
@@ -115,10 +115,13 @@ bool CompileState::compile(bool force_replace) {
           getPlanNode(filterProjectOp->planNodeId()));
       auto filterNode = filterProjectOp->filterNode();
       bool canBeEvaluated = true;
-      if (projectPlanNode && !ExpressionEvaluator::canBeEvaluated(projectPlanNode->projections())) {
+      if (projectPlanNode &&
+          !ExpressionEvaluator::canBeEvaluated(
+              projectPlanNode->projections())) {
         canBeEvaluated = false;
       }
-      if (canBeEvaluated && filterNode && !ExpressionEvaluator::canBeEvaluated({filterNode->filter()})) {
+      if (canBeEvaluated && filterNode &&
+          !ExpressionEvaluator::canBeEvaluated({filterNode->filter()})) {
         canBeEvaluated = false;
       }
       return canBeEvaluated;
--- a/velox/experimental/cudf/exec/ExpressionEvaluator.cpp
+++ b/velox/experimental/cudf/exec/ExpressionEvaluator.cpp
@@ -296,8 +296,8 @@ bool canBeEvaluated(const core::TypedExprPtr& expr) {
 
     case core::ExprKind::kCall: {
       const auto* call = expr->asUnchecked<core::CallTypedExpr>();
-      const auto name =
-          stripPrefix(call->name(), CudfConfig::getInstance().functionNamePrefix);
+      const auto name = stripPrefix(
+          call->name(), CudfConfig::getInstance().functionNamePrefix);
       if (supportedOps.count(name) || binaryOps.count(name) ||
           unaryOps.count(name)) {
         return std::all_of(


velox/exec/FilterProject.h
```diff
@@ -82,6 +82,10 @@ class FilterProject : public Operator {
   /// tracking is enabled via query config.
   OperatorStats stats(bool clear) override;
 
+  std::shared_ptr<const core::FilterNode> filterNode() const {
```
Yuhta: `const std::shared_ptr<const core::FilterNode>&`

jinchengchenghh:
Could you help review? Thanks!  @devavret

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -93,30 +103,20 @@ bool CompileState::compile(bool force_replace) {
...
 
-  auto isFilterProjectSupported = [](const exec::Operator* op) {
+  auto isFilterProjectSupported = [getPlanNode](const exec::Operator* op) {
     if (auto filterProjectOp = dynamic_cast<const exec::FilterProject*>(op)) {
-      auto info = filterProjectOp->exprsAndProjection();
-      return ExpressionEvaluator::canBeEvaluated(info.exprs->exprs());
+      auto projectPlanNode = std::dynamic_pointer_cast<const core::ProjectNode>(
+          getPlanNode(filterProjectOp->planNodeId()));
+      const auto exprs = exprsAndProjection(*filterProjectOp, projectPlanNode);
+      return ExpressionEvaluator::canBeEvaluated(exprs);
```
devavret: This can be simplified so that the exprsAndProjection function above is not needed
```suggestion
      auto const& projectPlanNode = std::dynamic_pointer_cast<const core::ProjectNode>(
          getPlanNode(filterProjectOp->planNodeId()));
      auto const& filterPlanNode = filterProjectOp->filterNode();
   
      return ExpressionEvaluator::canBeEvaluated(projectPlanNode->projections()) &&
                  ExpressionEvaluator::canBeEvaluated(filterPlanNode->filter());
```

velox/experimental/cudf/exec/CudfFilterProject.h
```diff
@@ -67,18 +67,49 @@ class CudfFilterProject : public exec::Operator, public NvtxHelper {
...
+  void initializeFilterProject();
+
+  /// Data for accelerator conversion.
+  struct Export {
+    const exec::ExprSet* exprs;
+    bool hasFilter;
+    const std::vector<exec::IdentityProjection>* resultProjections;
+  };
+
+  Export exprsAndProjection() const {
```
devavret: Is this needed in `CudfFilterProject`? This was used in `FilterProject` to extract information for replacement in wave.

Also, considering how initializing operators before replacement had its problems, perhaps wave also needs a better method to extract export information. @Yuhta thoughts?

velox/experimental/cudf/exec/CudfFilterProject.h
```diff
@@ -67,18 +67,49 @@ class CudfFilterProject : public exec::Operator, public NvtxHelper {
...
+
+  // Indices for fields/input columns that are both an identity projection and
+  // are referenced by either a filter or project expression. This is used to
+  // identify fields that need to be preloaded before evaluating filters or
+  // projections.
+  // Consider projection with 2 expressions: f(c0) AND g(c1), c1
+  // If c1 is a LazyVector and f(c0) AND g(c1) expression is evaluated first, it
+  // will load c1 only for rows where f(c0) is true. However, c1 identity
+  // projection needs all rows.
+  std::vector<column_index_t> multiplyReferencedFieldIndices_;
```
devavret: Cudf implementation doesn't need this yet.

velox/experimental/cudf/exec/CudfFilterProject.h
```diff
@@ -67,18 +67,49 @@ class CudfFilterProject : public exec::Operator, public NvtxHelper {
...
+  }
+
   bool allInputProcessed();
+
   // If true exprs_[0] is a filter and the other expressions are projections
   const bool hasFilter_{false};
+  const bool lazyDereference_;
+
+  std::unique_ptr<exec::ExprSet> exprs_;
+  int32_t numExprs_;
```
devavret: They don't appear to be used after initialize. They can be removed after consolidating the `initialize` and `initializeFilterProject` methods.

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -40,13 +41,64 @@ void debugPrintTree(
     debugPrintTree(input, indent + 2);
   }
 }
+
+bool checkAddIdentityProjection(
```
devavret, line 46: This seems to be copied from `FilterProject.cpp`. Can it be extracted from both places to reduce code duplication? @Yuhta would it be ok if we did this for `FilterProject.cpp`?

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -59,30 +111,106 @@ CudfFilterProject::CudfFilterProject(
...
   }
+
   projectEvaluator_ = ExpressionEvaluator(
       hasFilter_ ? projectExprs : info.exprs->exprs(), inputType);
+
+  filter_.reset();
+  project_.reset();
+}
+
+void CudfFilterProject::initializeFilterProject() {
```
devavret: The contents of this can simply be included in `CudfFilterProject::initialize()`

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -40,13 +41,64 @@ void debugPrintTree(
...
+      identityProjections.emplace_back(inputChannel, outputChannel);
+      return true;
+    }
+  }
+
+  return false;
+}
+
+// Split stats to attrbitute cardinality reduction to the Filter node.
+std::vector<exec::OperatorStats> splitStats(
```
devavret, line 66: Is this confirmed to be wired in? When we replace `FilterProject` with `CudfFilterProject`, did the stats splitter made for the former work for the latter? It would be nice if this was confirmed to be working, but in case it's not, this can be replaced with a TODO.

devavret (changes requested):
Thanks for this change. Requesting some cleanups but mostly, the approach looks good. I'd also request you to manually run all cudf unit tests to confirm this doesn't break other cudf based operators.

jinchengchenghh:
Now we only validate the function name, to validate consolidate, we must need to validate the input type and whether input argument is constant by constructing the CudfFunction, this is constructed from Expr, which is a big refactor, we need to discuss that further to make the decision.

If the plan does not contains the fallback , means all the operator has been replaced to Cudf operator, this exception should not occur, but that pool reserved check may not guarantee no memory allocation in driver creation, it should check peakBytes? @xiaoxmeng If that, we should not initialize operator.

jinchengchenghh:
Even if all the operators can be offloaded to GPU, we can still get this exception https://github.com/facebookincubator/velox/pull/14925#issuecomment-3402772925

velox/experimental/cudf/exec/ToCudf.cpp
jinchengchenghh: The projectPlanNode or filterPlanNode might be null, so I remove exprsAndProjection and do null check before

velox/experimental/cudf/exec/CudfFilterProject.cpp
jinchengchenghh: ParallelProject.cpp also copies from FilterProject.cpp, could we extract them to a new file?

velox/experimental/cudf/exec/CudfFilterProject.cpp
jinchengchenghh: I think it partially works for CudfFilterProject  https://github.com/facebookincubator/velox/pull/12895/files, The stats origins from pipelineStats, number of bytes is collected from RowVector `estimateFlatSize`, if we implement this function for CudfVector, we can collect the bytes stats, but we can collect the inputVectors number stats for now.
https://github.com/facebookincubator/velox/blob/main/velox/exec/Driver.cpp#L584
```
  const auto taskStats = tasks_[0]->taskStats();
  ASSERT_EQ(taskStats.pipelineStats.size(), 1);
  const auto& operators = taskStats.pipelineStats[0].operatorStats;
  EXPECT_GT(operators[1].getOutputTiming.wallNanos, 0);
  EXPECT_EQ(operators[0].outputPositions, 10000000);
```

jinchengchenghh: So I would prefer to keep this and implement CudfVector-> estimateFlatSize in the future to enable fully stats.

jinchengchenghh:
I describe the unit test result in the PR description, passed except the already broken test OrderByTest.singleKey @devavret

velox/experimental/cudf/exec/CudfFilterProject.cpp
devavret: CudfVector-> estimateFlatSize is already implemented. So you're saying this should work as expected

velox/experimental/cudf/exec/CudfFilterProject.h
devavret: Looks like `lazyDereference_` and `exprs_` also don't need to be stored.

velox/experimental/cudf/exec/CudfFilterProject.cpp
jinchengchenghh: yes

velox/experimental/cudf/exec/CudfFilterProject.h
jinchengchenghh: They are used in initialize   exprs_ = exec::makeExprSetFromFlag(
      std::move(allExprs), operatorCtx_->execCtx(), lazyDereference_);

jinchengchenghh: I moved them to initialize function.

devavret (approved):
Only one cleanup concern remains but it's not blocking. Otherwise LGTM

jinchengchenghh:
The failed test is not related to this PR, create an issue to track it https://github.com/facebookincubator/velox/issues/15188

jinchengchenghh:
The fedora debug CI also failed in this PR https://github.com/facebookincubator/velox/actions/runs/18548851948/job/52872103922?pr=15185

PR #14963, opened 2025-09-25
feat(cudf): Remove gflags in favor of CudfConfig

majetideepak:
gflags are not recommended to be part of any core library. The preferred way is to use configs, as they are
easier to change in a deployment.
This PR introduces a CudfConfig class and adds a singleton to be used in place of gflags.
Existing cuDF clients that depended on these gflags must either move to using this singleton
or can add these gflags in the client and translate to the singleton.
A helper method `initialize()` has been added to help with the translation.
See changes inside HashJoinTest.cpp and TableWriteTest.cpp

### DIFF
--- a/velox/experimental/cudf/CudfConfig.h
+++ b/velox/experimental/cudf/CudfConfig.h
@@ -25,16 +25,18 @@ struct CudfConfig {
   /// Clients must set the configs below before invoking registerCudf().
   static CudfConfig& getInstance();
 
+  /// Enable cudf by default.
+  /// Clients can disable here and enable it via the QueryConfig as well.
+  bool enabled{true};
   /// Enable debug printing.
   bool debugEnabled{false};
   /// Memory resource for cuDF.
   /// Possible values are (cuda, pool, async, arena, managed, managed_pool).
   std::string memoryResource{"async"};
+  /// The initial percent of GPU memory to allocate for memory resource.
+  int memoryPercent{50};
   /// Register all the functions with the functionNamePrefix.
   std::string functionNamePrefix;
-  /// The initial percent of GPU memory to allocate for memory resource for one
-  /// thread.
-  int memoryPercent{50};
   /// Force replacement of operators. Throws an error if a replacement fails.
   bool forceReplace{false};
 };
--- a/velox/experimental/cudf/CudfConfig.h
+++ b/velox/experimental/cudf/CudfConfig.h
@@ -17,26 +17,44 @@
 #pragma once
 
 #include <string>
+#include <unordered_map>
 
 namespace facebook::velox::cudf_velox {
 
 struct CudfConfig {
+  /// Keys used by the initialize() method.
+  static constexpr const char* kCudfEnabled{"cudf.enabled"};
+  static constexpr const char* kCudfDebugEnabled{"cudf.debug_enabled"};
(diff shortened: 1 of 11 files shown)


majetideepak:
CC: @zoltan

velox/experimental/cudf/CudfConfig.h
```diff
@@ -0,0 +1,42 @@
...
+  /// Enable debug printing.
+  bool debugEnabled{false};
+  /// Memory resource for cuDF.
+  /// Possible values are (cuda, pool, async, arena, managed, managed_pool).
+  std::string memoryResource{"async"};
+  /// Register all the functions with the functionNamePrefix.
+  std::string functionNamePrefix;
+  /// The initial percent of GPU memory to allocate for memory resource for one
+  /// thread.
+  int memoryPercent{50};
```
zoltan: this is the rmm pool's initial size and the pool is global - all drivers allocate from this pool. this is not per-thread.

majetideepak: I copied the existing comment. Does this apply to all memoryResource types?

zoltan: @bdice can keep me honest here but that is my understanding.

bdice: Yes, the memory resource should be used for the entire process, not just one thread.

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -402,12 +405,26 @@ void unregisterCudf() {
...
-  return isCudfRegistered;
+CudfConfig& CudfConfig::getInstance() {
+  static CudfConfig instance;
+  return instance;
 }
 
-bool cudfDebugEnabled() {
-  return FLAGS_velox_cudf_debug;
+void CudfConfig::initialize(
+    std::unordered_map<std::string, std::string>&& config) {
```
karthikeyann, line 414: Is it callers responsibility to call `CudfConfig::initialize(config)`?

majetideepak: Yes. This is a helper method for clients such as Presto C++, which will load a `cudf.properties` file.

velox/experimental/cudf/CudfConfig.h
```diff
@@ -0,0 +1,62 @@
...
+  bool enabled{true};
+
+  /// Enable debug printing.
+  bool debugEnabled{false};
+
+  /// Memory resource for cuDF.
+  /// Possible values are (cuda, pool, async, arena, managed, managed_pool).
+  std::string memoryResource{"async"};
+
+  /// The initial percent of GPU memory to allocate for memory resource.
```
bdice: Once #14997 goes in, a better description would be:
```suggestion
  /// The initial percent of GPU memory to allocate for pool or arena memory resources.
```

majetideepak:
@bdice I added a description to the PR. Thanks.

bikramSingh91:
@majetideepak do you mind doing a rebase as it seems to have conflicts with some recently merged changes

velox/experimental/cudf/CudfConfig.h
majetideepak: I updated the description!

majetideepak:
@bikramSingh91 Can you please import now? I resolved the conflict. Thanks.

PR #14967, opened 2025-09-25
feat: Add TIME support to millisecond function

han-yan01:
Summary:
Add support for TIME type to the millisecond() function in Velox to enable extracting milliseconds from TIME values.

**Changes:**
- Added new `call` method in `MillisecondFunction` that handles TIME type
- TIME values are stored as milliseconds since midnight, so extraction uses modulo operation to get milliseconds within the current second
- Registered the new function signature `MillisecondFunction<int64_t, Time>`
- Added comprehensive tests covering various TIME values including edge cases

**Planned:**
- support for TIME WITH TIMEZONE

Differential Revision: D83291159


velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -867,6 +867,21 @@ struct MillisecondFunction : public TimestampWithTimezoneSupport<T> {
...
+
+  FOLLY_ALWAYS_INLINE void call(int64_t& result, const arg_type<Time>& time) {
+    // TIME is stored as milliseconds since midnight, so extract milliseconds
+    // within the current second.
+    result = time % Timestamp::kMillisecondsInSecond;
+  }
+
+  // TODO: Add support for TIME WITH TIMEZONE once the type is implemented
+  // TIME WITH TIMEZONE uses lower 12 bits for timezone, upper 52 bits for time
+  // FOLLY_ALWAYS_INLINE void call(int64_t& result, const
```
kgpai: nit: this comment etc is not required .

kgpai (commented):
Looks good , can you remove comment and look into fuzzer failure.

han-yan01:
> Looks good , can you remove comment and look into fuzzer failure. 

Thanks!
Reading into the error message, it seems that is because Fuzzer failure is due to TIME not supported in Presto. Which will be triggered for the rest of other implementation. Would u like me to exclude this and future implementations from the fuzzer test or do we just ignore the error? Thanks.

kgpai:
``` No default Hive type provided for unsupported Hive type: time, Source: RUNTIME, ErrorCode: INVALID_STATE```  - 

Thanks @han-yan01 . We need to first add support for casting to bigint , so that we can add a type transform , to fix the fuzzer failure . Created an issue about it : https://github.com/facebookincubator/velox/issues/14969 . 

Would you be open to working on adding cast to bigint support (should be realtively easy)? Reach out to me if you need some pointers .

han-yan01:
> ` No default Hive type provided for unsupported Hive type: time, Source: RUNTIME, ErrorCode: INVALID_STATE` -
> 
> Thanks @han-yan01 . We need to first add support for casting to bigint , so that we can add a type transform , to fix the fuzzer failure . Created an issue about it : #14969 .
> 
> Would you be open to working on adding cast to bigint support (should be realtively easy)? Reach out to me if you need some pointers .

Sure thing! WIP

PR #14973, opened 2025-09-26
Fix buffer over-reading on simd::transpose

mblanco-denodo:
Fix buffer over-reading on simd::transpose that can cause invalid garbage values to be stored on output buffer on release builds

### DIFF
--- a/velox/common/base/SimdUtil.h
+++ b/velox/common/base/SimdUtil.h
@@ -480,8 +480,7 @@ inline void storeLeading(
 
 /// Stores elements of 'input' selected by 'indices' into 'output'. output[i] =
 /// input[indices[i]].
-/// Indices and output may be the same. May over-read indices but will not
-/// dereference indices that are not in range. Writes exactly indices.size()
+/// Indices and output may be the same. Writes exactly indices.size()
 /// elements of 'output'.
 template <typename TData, typename TIndex, typename A = xsimd::default_arch>
 inline void transpose(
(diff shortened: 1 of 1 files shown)


velox/common/base/SimdUtil.h
```diff
@@ -480,8 +480,7 @@ inline void storeLeading(
 
 /// Stores elements of 'input' selected by 'indices' into 'output'. output[i] =
 /// input[indices[i]].
-/// Indices and output may be the same. May over-read indices but will not
-/// dereference indices that are not in range. Writes exactly indices.size()
+/// Indices and output may be the same. Writes exactly indices.size()
```
Yuhta, line 483: Can you make a different version that does not overread if you need it?  We pad the buffer so in most of cases overread is fine, and your fix has impact on performance, so should be an opt in.

velox/common/base/SimdUtil.h
```diff
@@ -497,12 +496,20 @@ inline void transpose(
...
+#if XSIMD_WITH_AVX2
     auto mask = simd::leadingMask<TData>(numLeft);
     auto indexBatch = loadGatherIndices<TData, TIndex>(indices.data() + i);
     const auto values = simd::maskGather<TData, TIndex>(
         xsimd::broadcast<TData>(0), mask, input, indexBatch);
     storeLeading<TData, A>(values, mask, numLeft, output + i);
+#else
+    #pragma unroll
+    for (uint32_t k = 0; k < numLeft; ++k) {
+	register uint32_t offset = i + k;
```
MBkkt, line 508:
```suggestion
	uint32_t offset = i + k;
```
it's doing nothing, see https://en.cppreference.com/w/cpp/keyword/register.html

mblanco-denodo:
I am closing it as the issue involves only compilations for architectures that do not support BMI and BMI2

PR #14974, opened 2025-09-26
fix: Change array_sort lambda signature

kgpai:
Summary:
array_sort lambda signature is defined as function(T, T, bigint) , however Presto has this defined as function(T, T, int) . This causes the side car to fail array_sort queries that use this lambda. See details in [this issue](https://github.com/prestodb/presto/issues/26146). 

This change changes the signature to match Presto which fixes the side car.

Differential Revision: D83353566

### DIFF
--- a/velox/functions/lib/ArraySort.cpp
+++ b/velox/functions/lib/ArraySort.cpp
@@ -450,7 +450,7 @@ std::vector<std::shared_ptr<exec::FunctionSignature>> signatures(
             .typeVariable("T")
             .returnType("array(T)")
             .argumentType("array(T)")
-            .constantArgumentType("function(T,T,bigint)")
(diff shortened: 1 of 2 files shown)


kgpai:
@rui-mo @jinchengchenghh @zhli1142015 Does spark use the function(T,T,bigint) signature ? If so i can add a special case for spark.

jinchengchenghh:
Spark sort array is different with Presto, it does not involves the change of this PR. https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/registration/RegisterArray.cpp#L220

kgpai:
Thanks @jinchengchenghh , So i can take it that spark isnt affected by this change ?

jinchengchenghh:
Yes

Issue #14969, opened 2025-09-26
Add TimeIntermediateTypeTransform


kgpai:
### Description

We need to add Time Intermediate type transform to [PrestoQueryRunnerIntermediateTypeTransforms.pp](https://github.com/facebookincubator/velox/blob/main/velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp) so that it can convert to a type that can be persisted and tested by the Fuzzer. 
To do this we need to first support casting Time to bigint .

kgpai:
cc: @han-yan01

PR #14975, opened 2025-09-26
feat: Add support for casting TIME to BIGINT in Velox expressions

han-yan01:
Summary:
**Problem:**
- This change adds support for casting TIME values to BIGINT in Velox's expression casting system so that we can add a type transform , to fix the fuzzer failure.
issue: https://github.com/facebookincubator/velox/issues/14969

**Solution:**
- Added BIGINT case to the `CastExpr::castFromTime` function in CastExpr.cpp
- The implementation is straightforward since TIME is internally represented as int64_t (milliseconds since midnight), making it essentially a passthrough operation
- Added comprehensive test coverage for the new functionality

**Implementation Details:**
- TIME values are stored internally as milliseconds since midnight (int64_t)
- The cast simply returns the same numeric value without any transformation
- No timezone adjustments needed (unlike TIME to VARCHAR which requires timezone handling)

Differential Revision: D83356013

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -278,6 +278,31 @@ VectorPtr CastExpr::castFromTime(
       buffer->setSize(rawBuffer - buffer->asMutable<char>());
       return castResult;
     }
+    case TypeKind::BIGINT: {
+      // if input is constant, create a constant output vector
+      if (input.isConstantEncoding()) {
+        auto constantInput = input.as<ConstantVector<int64_t>>();
+        if (constantInput->isNullAt(0)) {
+          return BaseVector::createNullConstant(
+              toType, rows.end(), context.pool());
+        } else {
+          auto constantValue = constantInput->valueAt(0);
+          return std::make_shared<ConstantVector<int64_t>>(
+              context.pool(),
+              rows.end(),
+              false, // isNull
+              toType,
+              std::move(constantValue));
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


velox/expression/CastExpr.cpp
```diff
@@ -278,6 +278,13 @@ VectorPtr CastExpr::castFromTime(
       buffer->setSize(rawBuffer - buffer->asMutable<char>());
       return castResult;
     }
+    case TypeKind::BIGINT: {
+      auto* resultFlatVector = castResult->as<FlatVector<int64_t>>();
+      applyToSelectedNoThrowLocal(context, rows, castResult, [&](int row) {
```
kgpai, line 301: Can we be smarter about this, if the input is a constant vector maybe create another constant vector of int64 type, if its flat see if we can reuse and if not only then do the copy ?

han-yan01: Thanks for the review, updated for O(1) for constants and O(n) for normal copy.

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3166,5 +3166,54 @@ TEST_F(CastExprTest, timeToVarcharCast) {
   }
 }
 
+TEST_F(CastExprTest, timeToBigintCast) {
+  {
```
kgpai, line 3170: Can we add a test that exercise the constant input case (with and without null) ? This is likely to be the most common use case.

PR #14976, opened 2025-09-26
feat: Add ignore duplicates

### LINKED ISSUE #14897: Add `ignoreDuplicates` for `AggregateFunctionMetadata`
### Description

We should add `bool ignoresDuplicates{false};` in `AggregateFunctionMetadata`. 

It should be `false` for such functions like `sum`, `count`, etc

But it should be `true` for 
Presto
1. `min`, 
2. `max`
3. `bool_and`
4. `bool_or`
5. `bitwise_and_agg`
6. `bitwise_or_agg`
7. `set_agg`
8. `set_union`
9. `max_data_size_for_stats`

Spark
1. `collect_set`
2. `min`, 
3. `max`

```cpp
struct AggregateFunctionMetadata {
  /// True if results of the aggregation depend on the order of inputs. For
  /// example, array_agg is order sensitive while count is not.
  bool orderSensitive{true};

  /// Indicates if this is a companion function.
  bool companionFunction{false};
};
```

NOTES

1. I also think for Spark `first`, `last`, `first_ignore_null`, `last_ignore_null` `ignoreDuplicates` should be true, because they're not deterministic

MBkkt:
Fix #14897

### DIFF
--- a/velox/functions/lib/aggregates/BitwiseAggregateBase.h
+++ b/velox/functions/lib/aggregates/BitwiseAggregateBase.h
@@ -72,6 +72,7 @@ class BitwiseAggregateBase : public SimpleNumericAggregate<T, T, T> {
 template <template <typename U> class T>
 exec::AggregateRegistrationResult registerBitwise(
     const std::string& name,
+    bool ignoreDuplicates,
     bool withCompanionFunctions,
     bool onlyPrestoSignatures,
     bool overwrite) {
@@ -114,7 +115,7 @@ exec::AggregateRegistrationResult registerBitwise(
                 inputType->kindName());
         }
       },
-      {false /*orderSensitive*/, false /*companionFunction*/},
+      {.ignoreDuplicates = ignoreDuplicates, .orderSensitive = false},
       withCompanionFunctions,
       overwrite);
 }
--- a/velox/functions/prestosql/aggregates/TDigestAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/TDigestAggregate.cpp
@@ -355,8 +355,8 @@ void registerTDigestAggregate(
         return std::make_unique<TDigestAggregate<double>>(
             hasWeight, hasCompression, resultTypes);
       },
-      {true /*orderSensitive*/, false /*companionFunction*/},
-      false /*companionFunction*/,
+      {},
+      withCompanionFunctions,
       overwrite);
 }
 } // namespace facebook::velox::aggregate::prestosql
--- a/velox/exec/Aggregate.h
+++ b/velox/exec/Aggregate.h
@@ -505,6 +505,10 @@ using AggregateFunctionFactory = std::function<std::unique_ptr<Aggregate>(
     const core::QueryConfig& config)>;
 
 struct AggregateFunctionMetadata {
+  /// True if results of the aggregation ignore duplicate values.
+  /// For example, min and max ignore duplicates while sum does not.
+  bool ignoreDuplicates{false};
+
   /// True if results of the aggregation depend on the order of inputs. For
   /// example, array_agg is order sensitive while count is not.
   bool orderSensitive{true};
--- a/velox/exec/AggregateInfo.cpp
+++ b/velox/exec/AggregateInfo.cpp
@@ -81,11 +81,17 @@ std::vector<AggregateInfo> toAggregateInfo(
             arg->toString());
       }
     }
+    const auto& name = aggregate.call->name();
+    // 1. Ignore duplicates property
+    //    if aggregate function is not sensitive to duplicates.
+    // 2. Ignore sorting properties
+    //    if aggregate function is not sensitive to the order of inputs.
+    auto* entry = getAggregateFunctionEntry(name);
+    const auto& metadata = entry->metadata;
 
-    info.distinct = aggregate.distinct;
-    info.intermediateType = resolveAggregateFunction(
-                                aggregate.call->name(), aggregate.rawInputTypes)
-                                .second;
+    info.distinct = !metadata.ignoreDuplicates && aggregate.distinct;
+    info.intermediateType =
+        resolveAggregateFunction(name, aggregate.rawInputTypes).second;
 
     // Setup aggregation mask: convert the Variable Reference name to the
     // channel (projection) index, if there is a mask.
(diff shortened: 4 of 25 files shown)


velox/functions/lib/aggregates/BitwiseAggregateBase.h
```diff
@@ -72,6 +72,7 @@ class BitwiseAggregateBase : public SimpleNumericAggregate<T, T, T> {
 template <template <typename U> class T>
 exec::AggregateRegistrationResult registerBitwise(
     const std::string& name,
+    bool ignoreDuplicates,
```
mbasmanova, line 75: Isn't this flag always true? Do we need this parameter?

mbasmanova:
CI is red. Please, take a look. If the issue is not related to that PR, search issues to see if it is a known problem. If not, please, create new issue. Provide a link here. Thanks.

velox/functions/lib/aggregates/BitwiseAggregateBase.h
MBkkt: It's not true for "xor" and it's true for "and" and "or"

MBkkt:
@mbasmanova I think bug is fixed (needs to wait CI to confirm, but locally everything is good now)

MBkkt:
@mbasmanova I also updated PR description, is it ok now?

mbasmanova:
> For an example I'm not sure about presto "min_by" and "max_by", should I add mark them too?

Yes, I think these functions do not care about duplicate inputs.

mbasmanova:
> Fix typo in registerTDigestAggregate. This typo preventing adding companion aggregate functions.

CC: @natashasehgal

mbasmanova:
> Fix typo in registerTDigestAggregate. This typo preventing adding companion aggregate functions.

CC: @kagamiori

mbasmanova:
> Fix typo in registerTDigestAggregate. This typo preventing adding companion aggregate functions.

@MBkkt Would you extract this fix into a separate PR?

velox/functions/prestosql/aggregates/TDigestAggregate.cpp
```diff
@@ -355,8 +355,8 @@ void registerTDigestAggregate(
         return std::make_unique<TDigestAggregate<double>>(
             hasWeight, hasCompression, resultTypes);
       },
-      {true /*orderSensitive*/, false /*companionFunction*/},
-      false /*companionFunction*/,
+      {},
+      withCompanionFunctions,
```
MBkkt, line 359:
```suggestion
      false /*companionFunction*/,
```

MBkkt:
@mbasmanova I separated https://github.com/facebookincubator/velox/pull/14995 and added ignoreDuplicates=true + orderSensitive=false for min_by + max_by

mbasmanova:
@MBkkt min_by and max_by are order sensitive in the presence of ties.

MBkkt:
@mbasmanova Thanks, fixed

MBkkt:
@mbasmanova @bikramSingh91 I rebased PR because there was single line conflict + I fixed clang-format

mbasmanova:
@MBkkt CI is red. Please, take a look.

MBkkt:
@mbasmanova 

One fail is related to incorrect fix of merge conflict, I will fix this.

Another looks like this

For such query
```
Aggregation[2][SINGLE [g0, g1, g2, g3, g4] a0 := max(ROW["c0"],ROW["c1"]) distinct] -> g0:INTEGER, g1:ARRAY<INTEGER>, g2:MAP<BOOLEAN,VARCHAR>, g3:BOOLEAN, g4:TIMESTAMP WITH TIME ZONE, a0:ARRAY<REAL>
  -- Project[1][expressions: (c0:REAL, ROW["c0"]), (c1:BIGINT, ROW["c1"]), (g0:INTEGER, ROW["g0"]), (g1:ARRAY<INTEGER>, ROW["g1"]), (g2:MAP<BOOLEAN,VARCHAR>, ROW["g2"]), (g3:BOOLEAN, ROW["g3"]), (g4:TIMESTAMP WITH TIME ZONE, coalesce(try(parse_datetime(ROW["g4"],yyyy-MM-dd HH:mm:ss.SSS ZZZ)),parse_datetime(ROW["g4"],yyyy-MM-dd HH:mm:ss.SSS ZZ)))] -> c0:REAL, c1:BIGINT, g0:INTEGER, g1:ARRAY<INTEGER>, g2:MAP<BOOLEAN,VARCHAR>, g3:BOOLEAN, g4:TIMESTAMP WITH TIME ZONE
    -- Values[0][1000 rows in 10 vectors] -> c0:REAL, c1:BIGINT, g0:INTEGER, g1:ARRAY<INTEGER>, g2:MAP<BOOLEAN,VARCHAR>, g3:BOOLEAN, g4:VARCHAR
```

Output is different

```
 Expected 984, got 984
1 extra rows, 1 missing rows
1 of extra rows:
	null | [1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570] | [{"key":true,"value":null}] | true | null | [0.1676550954580307,0.1676550954580307]

1 of missing rows:
	null | [1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570,1433765570] | [{"key":true,"value":null}] | true | null | [0.1676550954580307]
```

Can you explain this query for me? It looks like overload `max(column, array_size)` should be used, but I don't understand how this overload can be used when second argument isn't constant.

MBkkt:
@mbasmanova I fixed bug. Bug was that Presto min/max/min_by/max_by can have additional argument that specify how many min/max elements do we want to return, for such overloads we cannot ignore distinct.

Right now function metadata available per name, not per overload, so best what I can do is just disable ignoreDuplicates for them.

https://github.com/facebookincubator/velox/pull/14976/commits/128324aaf86c8e0805c7bca0cef8deec6fed7130

mbasmanova:
@MBkkt Nice finding. Makes sense. Do we need to fix metadata for Spark's min_by/max_by as well?

MBkkt:
@mbasmanova 

> Do we need to fix metadata for Spark's min_by/max_by as well?

Spark min, max, min_by, max_by doesn't have TopN overload, so I added for all of them `ignoreDuplicates=true`

Issue #14978, opened 2025-09-27
HiveConnector drops some conjuncts from the remaining filter


mbasmanova:
### Bug description

Hive connector has logic to decompose remaining filter. That logic assumes that AND and OR have exactly 2 arguments. When these have more arguments the 3rd and remaining arguments are lost. 

```
    -- TableScan[0][table: lineitem, remaining filter: (and(gte(cast("l_receiptdate" as DATE),1994-01-01),lt(cast("l_receiptdate" as DATE),1995-01-01),lt("l_commitdate","l_receiptdate"),lt("l_shipdate","l_commitdate"))), data columns: ROW<l_orderkey:BIGINT,l_partkey:BIGINT,l_suppkey:BIGINT,l_linenumber:INTEGER,l_quantity:DOUBLE,l_extendedprice:DOUBLE,l_discount:DOUBLE,l_tax:DOUBLE,l_returnflag:VARCHAR,l_linestatus:VARCHAR,l_shipdate:DATE,l_commitdate:DATE,l_receiptdate:DATE,l_shipinstruct:VARCHAR,l_shipmode:VARCHAR,l_comment:VARCHAR>] ->
       Estimate: 9507.65 rows, 0B peak memory
       Input: 9525 rows (45.64KB, 7 batches), Raw Input: 60175 rows (2.43MB), Output: 9525 rows (45.64KB, 7 batches), Cpu time: 23.16ms, Wall time: 25.58ms, Blocked wall time: 39.07ms, Peak memory: 2.33MB, Memory allocations: 36, Threads: 16, Splits: 1, CPU breakdown: B/I/O/F (216.11us/0ns/22.89ms/51.72us)
```

HiveConnectorUtil.cpp

```
  if ((call->name() == expression::kAnd && !negated) ||
      (call->name() == expression::kOr && negated)) {
    auto lhs = extractFiltersFromRemainingFilter(
        call->inputs()[0], evaluator, negated, filters, sampleRate);
    auto rhs = extractFiltersFromRemainingFilter(
        call->inputs()[1], evaluator, negated, filters, sampleRate);
    if (!lhs) {
      return rhs;
    }
    if (!rhs) {
      return lhs;
    }
    return replaceInputs(call, {lhs, rhs});
  }
```

CC: @Yuhta @oerling @xiaoxmeng 

### System information

n/a

### Relevant logs

```bash

```

PR #14979, opened 2025-09-27
fix(hive): Fix subfield filter extraction for AND/OR more than 2 arguments

Yuhta:
Summary:
Hive connector has this behavior because Prestissimo only generates AND/OR with 2 arguments today.  In order to use it in different engines we need to eliminate this bug.  If optimizer is smart enough to extract subfields filters correctly, we don't even need to do this though: https://github.com/prestodb/presto/issues/26203

Fix https://github.com/facebookincubator/velox/issues/14978

Differential Revision: D83414367

### DIFF
--- a/velox/connectors/hive/HiveConnectorUtil.cpp
+++ b/velox/connectors/hive/HiveConnectorUtil.cpp
@@ -917,17 +917,23 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
 
   if ((call->name() == expression::kAnd && !negated) ||
       (call->name() == expression::kOr && negated)) {
-    auto lhs = extractFiltersFromRemainingFilter(
-        call->inputs()[0], evaluator, negated, filters, sampleRate);
-    auto rhs = extractFiltersFromRemainingFilter(
-        call->inputs()[1], evaluator, negated, filters, sampleRate);
-    if (!lhs) {
-      return rhs;
+    std::vector<core::TypedExprPtr> args;
+    args.reserve(call->inputs().size());
+    for (const auto& input : call->inputs()) {
+      if (auto arg = extractFiltersFromRemainingFilter(
+              input, evaluator, negated, filters, sampleRate)) {
+        args.push_back(std::move(arg));
+      }
+      // If extractFiltersFromRemainingFilter returns nullptr, it means
+      // everything in input is converted to filters.
     }
-    if (!rhs) {
-      return lhs;
+    if (args.empty()) {
+      return nullptr;
     }
-    return replaceInputs(call, {lhs, rhs});
+    if (args.size() == 1) {
+      return std::move(args[0]);
+    }
+    return replaceInputs(call, std::move(args));
   }
   if (!negated) {
     double rate = getPrestoSampleRate(expr, call, evaluator);
--- a/velox/connectors/hive/tests/HiveConnectorTest.cpp
+++ b/velox/connectors/hive/tests/HiveConnectorTest.cpp
@@ -21,6 +21,7 @@
 #include "velox/connectors/hive/HiveConfig.h"
 #include "velox/connectors/hive/HiveConnectorUtil.h"
 #include "velox/connectors/hive/HiveDataSource.h"
+#include "velox/expression/ExprConstants.h"
 #include "velox/expression/ExprToSubfieldFilter.h"
 
 namespace facebook::velox::connector::hive {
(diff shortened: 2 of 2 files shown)


MBkkt (commented):
Maybe it's better to implement proper fix?
Code is really simple for this:
```cpp
    std::vector<core::TypedExprPtr> args;
    args.reserve(call->inputs().size());
    for (const auto& input : call->inputs()) {
      auto arg = extractFiltersFromRemainingFilter(
          input, evaluator, negated, filters, sampleRate);
      if (arg) {
        args.emplace_back(std::move(arg));
      }
    }
    if (args.empty()) {
      return nullptr;
    }
    if (args.size() == 1) {
      return std::move(args[0]);
    }
    return replaceInputs(call, std::move(args));
```

MBkkt:
>  Disable it for now for more than 2 arguments, and we can
implement it if we find it is useful (if optimizer is smart enough we don't even
need this).

What do you mean? I think smart optimizer should generate flat AND/OR instead of binary tries

Yuhta:
@MBkkt Optimizer should extract the subfield filters so we don't need to do this during execution.  We added this to overcome some limitations in Presto optimizer, but if we have control of optimizer we should not need this at all.  All the information is available during planning.

velox/dwio/common/MetadataFilter.cpp
```diff
@@ -165,12 +165,18 @@ std::unique_ptr<MetadataFilter::Node> MetadataFilter::Node::fromExpression(
     return nullptr;
   }
   if (call->name() == expression::kAnd) {
+    if (call->inputs().size() != 2) {
+      return nullptr;
+    }
     auto lhs = fromExpression(*call->inputs()[0], evaluator, negated);
     auto rhs = fromExpression(*call->inputs()[1], evaluator, negated);
     return negated ? OrNode::create(std::move(lhs), std::move(rhs))
                    : AndNode::create(std::move(lhs), std::move(rhs));
```
MBkkt, line 171: @Yuhta I'm interested, can you explain for me, please?

Why does OrNode use `andBits` and AndNode use `orBits`?

Here:
```cpp
uint64_t* eval(LeafResults& leafResults, int size) const override {
    auto* l = lhs_->eval(leafResults, size);
    auto* r = rhs_->eval(leafResults, size);
    if (!l || !r) {
      return nullptr;
    }
    bits::andBits(l, r, 0, size);
    return l;
}
```

Yuhta: The set bit here means that row group needs to be filtered out

MBkkt: Thanks, I understand now

MBkkt:
@Yuhta btw I made changes for MetadataFilter https://github.com/facebookincubator/velox/pull/14981

velox/connectors/hive/tests/HiveConnectorTest.cpp
```diff
@@ -610,6 +611,26 @@ TEST_F(HiveConnectorTest, extractFiltersFromRemainingFilter) {
   ASSERT_TRUE(remaining);
   ASSERT_EQ(
       remaining->toString(), "not(lt(ROW[\"c2\"],cast(0 as DECIMAL(20, 0))))");
+
+  // parseExpr gives AND/OR with 2 arguments.  We need to construct the node
+  // manually to have more than 2.
+  expr = std::make_shared<core::CallTypedExpr>(
+      BOOLEAN(),
+      std::vector<core::TypedExprPtr>({
```
mbasmanova: nit: use ctor that takes variadic number of arguments

```
std::make_shared<core::CallTypedExpr>(
   BOOLEAN(),
   expression::kAnd,
   parseExpr("c0 > 0", rowType),
   parseExpr("c1 > 0", rowType),
   parseExpr("c2 > 0::decimal(20, 0)", rowType));
```

velox/connectors/hive/tests/HiveConnectorTest.cpp
```diff
@@ -610,6 +611,26 @@ TEST_F(HiveConnectorTest, extractFiltersFromRemainingFilter) {
...
+          parseExpr("c0 > 0", rowType),
+          parseExpr("c1 > 0", rowType),
+          parseExpr("c2 > 0::decimal(20, 0)", rowType),
+      }),
+      expression::kAnd);
+  filters.clear();
+  remaining = extractFiltersFromRemainingFilter(
+      expr, &evaluator, false, filters, sampleRate);
+  ASSERT_EQ(sampleRate, 1);
+  ASSERT_EQ(filters.size(), 3);
```
mbasmanova, line 627: nit: perhaps, use https://google.github.io/googletest/reference/matchers.html#container-matchers

velox/dwio/common/MetadataFilter.cpp
```diff
@@ -165,12 +165,18 @@ std::unique_ptr<MetadataFilter::Node> MetadataFilter::Node::fromExpression(
     return nullptr;
   }
   if (call->name() == expression::kAnd) {
+    if (call->inputs().size() != 2) {
```
mbasmanova: Why do we need this restriction? Can we not just make it work for arbitrary number of arguments? Why do we need this logic in the first place? Can we add a test?

mbasmanova (commented):
@Yuhta Thank you for a quick fix. Some questions.

velox/dwio/common/MetadataFilter.cpp
MBkkt: I made separate PR with that https://github.com/facebookincubator/velox/pull/14981

velox/connectors/hive/tests/HiveConnectorTest.cpp
Yuhta: There is no equality defined on filter objects, here we check filter are generated on certain fields (which is the most relevant part)

mbasmanova:
> If optimizer is smart enough to extract subfields filters correctly, we don't even need to do this though.

Is there a GitHub issue in Presto to request the optimizer enhancement? If not, let's create one and reference.

velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -917,17 +917,21 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
...
-    auto lhs = extractFiltersFromRemainingFilter(
-        call->inputs()[0], evaluator, negated, filters, sampleRate);
-    auto rhs = extractFiltersFromRemainingFilter(
-        call->inputs()[1], evaluator, negated, filters, sampleRate);
-    if (!lhs) {
-      return rhs;
+    std::vector<core::TypedExprPtr> args;
+    args.reserve(call->inputs().size());
+    for (const auto& input : call->inputs()) {
+      if (auto arg = extractFiltersFromRemainingFilter(
```
mbasmanova, line 923: Don't we need to have an 'else' branch? Aren't we dropping something silently?

Yuhta: If the result is `nullptr`, it means everything in `input` is converted to `filters`, so we are not dropping things silently

mbasmanova: Would you add a comment and a test case? Thanks.

Yuhta:
@mbasmanova Created an issue in Presto repo: https://github.com/prestodb/presto/issues/26203

velox/connectors/hive/HiveConnectorUtil.cpp
Yuhta: Added

mbasmanova (approved):
Thanks. Would you reference newly created Presto issue in this PR's description?

PR #14980, opened 2025-09-27
fix: Add missing includes

MBkkt:
Add missing includes
1. mutex needed because of std::lock_guard in velox/common/base/Portability.h
2. folly_always_inline in velox/common/base/CountBits.h

### DIFF
--- a/velox/common/base/CountBits.h
+++ b/velox/common/base/CountBits.h
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <folly/CPortability.h>
+
 namespace facebook::velox {
 
 // Copied from format.h of fmt.
--- a/velox/common/base/Portability.h
+++ b/velox/common/base/Portability.h
@@ -19,6 +19,7 @@
 #include <atomic>
 #include <cstddef>
 #include <cstdint>
+#include <mutex>
 #include <vector>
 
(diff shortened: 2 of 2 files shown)


velox/common/base/SimdUtil-inl.h
```diff
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#pragma once
```
mbasmanova: xxx-inl.h files are supposed to be included only from the corresponding xxx.h files. Hence, it seems that #pragma once and #include Xxx.h are not necessary.

mbasmanova: > Without others vscode code-autocomplete is broken

Is this why you need this? If so, can we split this PR into two? For the xxx-inl.h changes, let's also update coding guidelines. Otherwise, folks won't know why this is needed.

MBkkt: Yes, only because of this.

This code is almost plain text to any clangd based IDE I think

MBkkt: https://github.com/facebookincubator/velox/pull/14996
(and in this PR only real missing includes)

mbasmanova (approved):
Thanks.

PR #14981, opened 2025-09-27
fix: MetadataFilter construction from flat (multi) And/Or

MBkkt:


velox/dwio/common/MetadataFilter.cpp
```diff
@@ -68,86 +68,113 @@ class MetadataFilter::LeafNode : public Node {
...
+      bool conjuction,
+      std::vector<std::unique_ptr<Node>> args);
+
+  static std::unique_ptr<Node> fromExpression(
+      const std::vector<core::TypedExprPtr>& inputs,
+      core::ExpressionEvaluator* evaluator,
+      bool conjunction,
+      bool negated) {
+    std::vector<std::unique_ptr<Node>> args;
+    for (const auto& input : inputs) {
```
MBkkt, line 84:
```suggestion
    std::vector<std::unique_ptr<Node>> args;
    args.reserve(inputs.size());
    for (const auto& input : inputs) {
```

mbasmanova (commented):
@MBkkt Can we add a test?

velox/dwio/common/MetadataFilter.cpp
```diff
@@ -68,86 +68,118 @@ class MetadataFilter::LeafNode : public Node {
...
-      return lhs;
+      bool conjuction,
+      std::vector<std::unique_ptr<Node>> args);
+
+  static std::unique_ptr<Node> fromExpression(
+      const std::vector<core::TypedExprPtr>& inputs,
+      core::ExpressionEvaluator* evaluator,
+      bool conjunction,
+      bool negated) {
+    conjunction = conjunction ? !negated : negated;
```
Yuhta: More readable way: `conjunction = negated ? !conjunction : conjunction`

velox/dwio/common/MetadataFilter.cpp
```diff
@@ -68,86 +68,118 @@ class MetadataFilter::LeafNode : public Node {
...
-struct MetadataFilter::OrNode : Node {
-  static std::unique_ptr<Node> create(
-      std::unique_ptr<Node> lhs,
-      std::unique_ptr<Node> rhs) {
-    if (!lhs || !rhs) {
-      return nullptr;
+struct MetadataFilter::AndNode final : ConditionNode {
+  using ConditionNode::ConditionNode;
+
+  uint64_t* eval(LeafResults& leafResults, int size) const final {
```
Yuhta, line 123: Do we really need this subclass anymore?  Just check `conjunction_` in `eval` should be enough, like we did in `ConditionNode::fromExpression`.

MBkkt: Now it's compile time "condition" (compile bitAnd or bitOr), do you suggest to make it runtime?

velox/dwio/common/MetadataFilter.cpp
MBkkt: Fixed

MBkkt:
@mbasmanova I added tests (can be checked with `velox_dwrf_e2e_filter_test --gtest_filter="*metadata*"`)

velox/dwio/common/MetadataFilter.cpp
Yuhta: It's a runtime lookup in vtable anyway (once per `eval` call).  Another way is to make `condition` a template parameter

mbasmanova (commented):
@Yuhta Jimmy, would you help review?

I expected this utility to have its own light-weight unit test. Why are we testing it inside E2EFilterTest?

velox/dwio/common/MetadataFilter.cpp
MBkkt: I don't understand your point, now loop without runtime condition, vtable indirection is once per loop

> Another way is to make condition a template parameter

It will be same what I did but with different syntax

Yuhta: What I mean is this:
```c++
template <bool kConjunction>
class ConditionNode {
  uint64_t* eval(LeafResults& leafResults, int size) const final {
    if constexpr (kConjuncion) {
      // ...
    } else {
      // ...
    }
  }
};

static std::unique_ptr<Node> create(/* ... */) {
  // ...
  return conjunction ? std::make_unique<ConditionNode<true>>(std::move(args)) : std::make_unique<ConditionNode<false>>(std::move(args));
}
```
But yeah it's not that different, just save 2 subclasses with only 1 methods, it's more of a nit.

Yuhta:
@mbasmanova The logic to generate leaf node values is format dependent so I put it in E2E tests to be called on each file formats.  We could prefill the leaf values and test only the condition combination though.

velox/dwio/common/MetadataFilter.cpp
MBkkt: I understand this suggestion, I think it's a little worse because compiler will compile other methods (that don't require compile time if) twice, e.g. now `ConditionNode::addToScanSpec` but with your suggestion it will be `ConditionNode<false>::addToScanSpec` and `ConditionNode<true>::addToScanSpec`.

I first thought about `template<bool And>` class, then decided to do this a little different way, because dislike this duplication

MBkkt:
@mbasmanova this PR waiting for merge last two days, is it ok and it will be merged soon or I need to do something?

PR #14982, opened 2025-09-27
Remove old vector API

MBkkt:
There's multiple reasons to do this:
- Newer xsimd doesn't support current code
- We pay cost for compile this code (because it's generic headers)
- We sometimes pay cost even when this methods never called (Constant and Bias vector init additional buffers)
- There's no performance benefits compared to generic code written in a good way and compiler allowed to vectorize it (sse/avx/avx2/avx512 enabled for target).


velox/vector/BiasVector.h
```diff
@@ -123,15 +119,6 @@ class BiasVector : public SimpleVector<T> {
...
   }
 
-  /**
-   * Loads a SIMD vector of data at the virtual byteOffset given
-   * Note this method is implemented on each vector type, but is intentionally
-   * not virtual for performance reasons
-   *
-   * @param byteOffset - the byte offset to laod from
-   */
-  xsimd::batch<T> loadSIMDValueBufferAt(size_t index) const;
```
mbasmanova, line 133: This API is used internally.

mbasmanova: If it helps, we could add a macro to conditionally include this code. We'll then use that macro internally (the usage is in legacy code that cannot be modified due to legal reasons.)

MBkkt: Can you suggest name for this macro for legacy code?

MBkkt: I will put this code under it

mbasmanova: perhaps, VELOX_ENABLE_LOAD_SIMD_VALUE_BUFFER

MBkkt: @mbasmanova I did this, will you look?

.github/workflows/linux-build-base.yml
```diff
@@ -103,6 +103,7 @@ jobs:
             "-DVELOX_ENABLE_WAVE=ON"
             "-DVELOX_MONO_LIBRARY=ON"
             "-DVELOX_BUILD_SHARED=ON"
+            "-DVELOX_ENABLE_LOAD_SIMD_VALUE_BUFFER=ON"
```
mbasmanova: Do we need this change? If we can only have one run of CI, then, perhaps, we want to test non-legacy code, no? If we want to test legacy code, we'd need to add another CI run. Not sure it is worth it though.

CC: @kgpai

CMakeLists.txt
```diff
@@ -166,6 +166,11 @@ option(VELOX_BUILD_VECTOR_TEST_UTILS "Builds Velox vector test utilities" OFF)
 option(VELOX_BUILD_PYTHON_PACKAGE "Builds Velox Python bindings" OFF)
 option(VELOX_PYTHON_LEGACY_ONLY "Build only the pyvelox module" OFF)
 option(VELOX_BUILD_RUNNER "Builds velox runner" ON)
+option(
```
mbasmanova: I don't think we need this. Legacy app doesn't use cmake.

mbasmanova (commented):
@MBkkt Looks great % comments about CI and cmake. Would you update PR description to describe the new macro?

.github/workflows/linux-build-base.yml
MBkkt: I removed this

MBkkt:
@mbasmanova this PR waiting for merge last two days, is it ok and it will be merged soon or I need to do something?

mbasmanova:
@MBkkt I see that there is a pretty long queue of ready-to-merge PRs. I hope @kgpai will get to this PR soon though.

@kgpai Krishna, would you help merge this PR?

bikramSingh91:
This is causing internal failures for some projects that use this code. We'll need some time to fix it, so please expect some delay in merging

bikramSingh91:
Just an FYI, the internal failures have been resolved and I have initiated a land/merge. Once all CI tests pass this should merge.

PR #14983, opened 2025-09-27
fix: Fix race in MemoryPool in debug mode

kletkavrubashku:
Summary:
`MemoryPool::toString()` calls `dumpRecordsDbg()` without locking `debugAllocMutex_` mutex.
```
std::string MemoryPoolImpl::toString(bool detail) const {
  std::string result;
  {
    result = toStringLocked();
  }
  if (detail) {
    result += "\n" + treeMemoryUsage();
  }
  if (FOLLY_UNLIKELY(debugEnabled())) {
    result += "\n" + dumpRecordsDbg();
  }
  return result;
}
```
Let's have 2 versions for `dumpRecordsDbg()` method to avoid confusion.

Differential Revision: D83418591


tanjialiang (approved):
LGTM thanks for the fix

PR #14986, opened 2025-09-28
fix: Use correct C++ without UB (data race), that produce same or faster code

MBkkt:
Another argument is `annotate_ignore_thread_sanitizer_guard` should be used only to mark false positive issues, not the real one.

Benchmark results:
AMD Ryzen 9 9950X, bare metal
```
[...]elox/exec/benchmarks/AtomicsBench.cpp     relative  time/iter   iters/s
std_atomic_bool_write_seq_cst                              1.16min    14.36m
std_atomic_bool_write_release                                4.50s   222.17m
std_atomic_bool_write_relaxed                                4.48s   223.00m
one_way_flag_write                                         55.20ms     18.12
one_way_flag_unsafe_write                                  55.57ms     17.99
std_atomic_bool_read_seq_cst                               29.66ms     33.72
std_atomic_bool_read_acquire                               30.10ms     33.22
std_atomic_bool_read_relaxed                               30.00ms     33.33
one_way_flag_read                                          29.84ms     33.51
one_way_flag_unsafe_read                                   59.68ms     16.76
```

### DIFF
--- a/velox/exec/OneWayStatusFlag.h
+++ b/velox/exec/OneWayStatusFlag.h
@@ -16,53 +16,28 @@
 
 #pragma once
 
-#include <folly/synchronization/SanitizeThread.h>
 #include <atomic>
 
 namespace facebook::velox::exec {
 
-/// A simple one way status flag that uses a non atomic flag to avoid
-/// unnecessary atomic operations.
 class OneWayStatusFlag {
  public:
-  bool check() const {
-#if defined(__x86_64__)
-    folly::annotate_ignore_thread_sanitizer_guard g(__FILE__, __LINE__);
-    return fastStatus_ || atomicStatus_.load();
-#else
-    return atomicStatus_.load(std::memory_order_relaxed) ||
-        atomicStatus_.load();
-#endif
+  bool check() const noexcept {
+    return status_.load(std::memory_order_acquire);
   }
 
-  void set() {
-#if defined(__x86_64__)
-    folly::annotate_ignore_thread_sanitizer_guard g(__FILE__, __LINE__);
-    if (!fastStatus_) {
-      atomicStatus_.store(true);
-      fastStatus_ = true;
+  void set() noexcept {
+    if (!status_.load(std::memory_order_relaxed)) {
+      status_.store(true, std::memory_order_release);
     }
-#else
-    if (!atomicStatus_.load(std::memory_order_relaxed)) {
-      atomicStatus_.store(true);
-    }
-#endif
   }
 
-  /// Operator overload to convert OneWayStatusFlag to bool
-  operator bool() const {
+  explicit operator bool() const noexcept {
     return check();
   }
 
  private:
-#if defined(__x86_64__)
-  // This flag can only go from false to true, and is only checked at the end of
-  // a loop. Given that once a flag is true it can never go back to false, we
-  // are ok to use this in a non synchronized manner to avoid the overhead. As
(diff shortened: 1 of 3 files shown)


velox/exec/OneWayStatusFlag.h
```diff
@@ -16,53 +16,28 @@
...
-  operator bool() const {
+  explicit operator bool() const noexcept {
     return check();
   }
 
  private:
-#if defined(__x86_64__)
-  // This flag can only go from false to true, and is only checked at the end of
-  // a loop. Given that once a flag is true it can never go back to false, we
-  // are ok to use this in a non synchronized manner to avoid the overhead. As
```
MBkkt, line 61: C++ memory model is quite complicated, and such UB potentially can produce incorrect behavior for some code (probably non trivial synchronization)

velox/exec/benchmarks/CMakeLists.txt
```diff
@@ -158,3 +158,10 @@ target_link_libraries(
   velox_vector_fuzzer
   Folly::follybenchmark
 )
+
+add_executable(velox_atomics_benchmark AtomicsBench.cpp)
+
+target_link_libraries(
+  velox_atomics_benchmark
+  Folly::follybenchmark
+)
```
MBkkt:
```suggestion
target_link_libraries(velox_atomics_benchmark Folly::follybenchmark)
```
cmake format

Yuhta:
CC: @joeg

velox/exec/benchmarks/AtomicsBench.cpp
```diff
@@ -46,97 +70,133 @@ void runParallelUpdates(
...
+#if defined(__x86_64__) && !defined(TSAN_BUILD)
+
+BENCHMARK(one_way_flag_unsafe_write) {
+  OneWayStatusFlagUnsafe flag;
   runParallelUpdates(
       [&](size_t iters) {
         for (size_t i = 0; i < iters; ++i) {
           flag.set();
+          bool dummy{};
+          folly::doNotOptimizeAway(dummy);
```
Yuhta, line 137: So the whole loop was optimized away, and that's why this one was so fast

MBkkt: Right, but it won't happen in real life, in real life some optimizations like this is possible (regsiter instead of memory in cacheline access).
But it won't lead to real speedup, because other code that do multiple memory accessed executed

velox/exec/benchmarks/AtomicsBench.cpp
```diff
@@ -16,16 +16,40 @@
 
 #include <folly/Benchmark.h>
 #include <folly/init/Init.h>
-#include <gtest/gtest.h>
+#include <folly/synchronization/SanitizeThread.h>
```
bikramSingh91: internal linter has marked this as an unused include, can you please remove this if this indeed is the case?

MBkkt:
```suggestion
```

MBkkt: Yes, it's right, it was used, but now unused, I removed it.

MBkkt:
@bikramSingh91 this PR waiting for merge last two days, is it ok and it will be merged soon or I need to do something?

bikramSingh91:
@MBkkt it will be merged soon

bikramSingh91:
Merge showing warning of older base, just rebased and started the merge process again, if all goes well, should land today

PR #14987, opened 2025-09-28
refactor: Buffer counter should use weaker memory order

MBkkt:
ClickHouse use boost intrusive_ptr_counter for this.
It has `acq_rel` for fetch_sub/add and `acquire` for load

https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/COW.h

This PR also adjust some small issues in the Buffer classes


velox/buffer/Buffer.h
```diff
@@ -611,11 +615,11 @@ class NonPODAlignedBuffer : public Buffer {
...
  protected:
   NonPODAlignedBuffer(velox::memory::MemoryPool* pool, size_t capacity)
-      : Buffer(
+      : Buffer{
             pool,
             reinterpret_cast<uint8_t*>(this) + sizeof(*this),
             capacity,
-            false /*podType*/) {
+            false /*podType*/} {
     static_assert(sizeof(*this) == AlignedBuffer::kAlignment);
```
MBkkt, line 682: why sizeof instead of alignof?

MBkkt:
@Yuhta I saw you reviewed similar PR, can you review this please?

MBkkt:
@mbasmanova Maybe you can look? Or you know someone who will look?

According to this PR https://github.com/facebookincubator/velox/pull/12928 it can be significant improvement on arm

mbasmanova:
> This PR also adjust some small issues in the Buffer classes

@MBkkt Would you extract these changes into a separate PR?

mbasmanova:
> refactor: Buffer counter should use weaker memory order

Isn't the main goal to improve performance? If so, perhaps, 'refactor' is not quite the right type.

MBkkt:
@mbasmanova I changed the title and made separate PR with refactoring https://github.com/facebookincubator/velox/pull/15085

mbasmanova (approved):
Thanks.

mbasmanova (commented):
@MBkkt Curious, what kind of speed up you observe? Is this measurable / visible for e2e query execution?

kKPulla:
@MBkkt can you please rebase on main (on top of #15100 ) to make sure the signals turn green? Thanks!

MBkkt:
@kKPulla I rebased this PR

MBkkt:
@mbasmanova 

> Curious, what kind of speed up you observe? Is this measurable / visible for e2e query execution?

I just noticed that PR with similar idea wasn't merged because it was too relaxed.
This happened than I looked to the issue that was origin of  `OneWayStatusFlag` (I did this because I made `OneWayStatusFlag` correct in terms of C++ memory model).

I remembered that clickhouse using similar idea for their buffers (check unique or not).
So decided to check their code, because maybe I didn't remember it right.
But it was same as I remembered and what I think should be correct and faster.

Regards to speedup, I didn't make my own measurements, I'm waiting for arm machine...
But according to mentioned PR, for micro benchmark it should be something like this
>  VectorSlice benchmark shows up to 12% performance improvement, as measured on Arm Neoverse V1.

Unfortunately there's no evidence of e2e benchmarks in mentioned PR

PR #14988, opened 2025-09-28
refactor: Move driver creation into task init from the first next call (#13075)

xiaoxmeng:
Summary:

This is required for task barrier processing.

Differential Revision: D73299498


tanjialiang (approved):
LGTM thanks for the refactoring

PR #14990, opened 2025-09-28
fix: Fix incorrect initialization of empty 'folly::Promise' object

lingbin:
X-link: https://github.com/prestodb/presto/pull/26171

For an empty `folly::Promise` object that is expected to be overwritten,
we should use `folly::makeEmpty()` to initialize it (it is 'invalid')
instead of the default constructor (it will be 'valid' but
'not fulfilled', assigning to it will cause an exception. Creating an
exception triggers a stack unwind, which can saturate the CPU in
high-concurrency scenarios, **causing significant performance issues**.

### DIFF
--- a/velox/exec/tests/utils/LocalExchangeSource.cpp
+++ b/velox/exec/tests/utils/LocalExchangeSource.cpp
@@ -271,7 +271,7 @@ class LocalExchangeSource : public exec::ExchangeSource {
   }
 
   bool checkSetRequestPromise() {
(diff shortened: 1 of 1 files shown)


lingbin:
@mbasmanova  Could you review this PR?  
This issue was discovered in `PrestoExchangeSource` in presto_cpp(https://github.com/prestodb/presto/pull/26171), but also exists in `LocalExchangeSource` in Velox.

PR #14832, opened 2025-09-11
fix(TextReader): Fix incorret result for DECIMAL and DATE

xin-zhang2:
Fixes https://github.com/facebookincubator/velox/issues/14821

Issue #2125, opened 2022-07-27
Enable ASAN/TSAN/UBSAN/Unit tests for Linux & MacOS


kgpai:
Currently Unit tests arent run for MacOs builds ; this needs to be fixed. We should also enable ASAN/TSAN/UBSAN for Linux and MacOS.

mbasmanova:
@kgpai What's blocking us from enabling ASAN/TSAN/UBSAN in CI?

CC: @assignUser @MBkkt

MBkkt:
@mbasmanova from my side main issue in the environment, locally I build with clang-20, lld-20, compiler-rt 20, etc
And it works. In CI environment is quite unusual to me, some ancient gcc, etc.

I think @assignUser can explain it better :)

kgpai:
@mbasmanova  We are currently on gcc and need to move to clang-19+ and above to enable sanitizers. Clang <=14 does not work with C++ 20 and thus the need to clang-19+. We plan to move to clang-20 but that will also require updating simdjson (to 3.13) and fmt (to 11)  . We plan to do this by end of this quarter. 

cc: @czentgr

PR #14994, opened 2025-09-29
fix(build): Fix TextReader build errors with Clang 15

xin-zhang2:
Fixes TextReader build failures with Clang 15.

Capturing structured bindings in lambdas is a C++20 feature, and Clang added support for it starting in Clang 16.
https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html.
> Implemented [P1091R3](https://wg21.link/p1091r3) and [P1381R1](https://wg21.link/P1381R1) to support capturing structured bindings in lambdas ([#52720](https://github.com/llvm/llvm-project/issues/52720), [#54300](https://github.com/llvm/llvm-project/issues/54300), [#54301](https://github.com/llvm/llvm-project/issues/54301), [#49430](https://github.com/llvm/llvm-project/issues/49430)).

### DIFF
--- a/velox/dwio/text/reader/TextReader.cpp
+++ b/velox/dwio/text/reader/TextReader.cpp
@@ -1096,17 +1096,17 @@ void TextRowReader::readElement(
     case TypeKind::BIGINT:
       if (reqT->isShortDecimal()) {
         const std::string& str = getString(*this, isNull, delim);
-        auto [precision, scale] = getDecimalPrecisionScale(*reqT);
+        auto precisionScale = getDecimalPrecisionScale(*reqT);
         setValueFromString<int64_t>(
             str,
             data,
             insertionRow,
-            [precision, scale](const std::string& s) -> std::optional<int64_t> {
+            [precisionScale](const std::string& s) -> std::optional<int64_t> {
               int64_t v = 0;
               const auto status = DecimalUtil::castFromString(
                   StringView(s.data(), static_cast<int32_t>(s.size())),
-                  precision,
-                  scale,
+                  precisionScale.first,
+                  precisionScale.second,
                   v);
               return status.ok() ? std::optional<int64_t>(v) : std::nullopt;
             });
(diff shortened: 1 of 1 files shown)


MBkkt:
@mbasmanova @assignUser why does clang-15 required? Can we move on to newer clang?

I think even llvm libc++ supports only >= clang-18 (starting from clang-20, already clang-21, so we can move even to clang-19) https://github.com/llvm/llvm-project/commit/7f845cba2ccc2ab637b8e40fbafb9f83a2d67c70

clang major version updated faster than gcc major version

mbasmanova:
@xin-zhang2 Would you clarify why you need support for clang 15?

mbasmanova:
CC: @kgpai @assignUser

xin-zhang2:
@mbasmanova @MBkkt Thanks for the review.
The motivation for this PR is to address a [Prestissimo build failure](https://github.com/facebookincubator/velox/pull/14832#discussion_r2386240443) occurs with Clang 15. 
Upgrading Clang in Presto Github Action env could also fix the issue, but since the Velox README lists 15 as the minimum supported version https://github.com/facebookincubator/velox?tab=readme-ov-file#supported-os-and-compiler-matrix, I opened this PR to make it compile with Clang 15.

mbasmanova:
> Velox README lists 15 as the minimum supported version

Maybe we need to reconsider. @assignUser @kgpai @majetideepak @czentgr  What do you think?

czentgr:
@mbasmanova We are working on using newer CLang compilers. But we also need some more dependency upgrades - notably for simdjson and fmt. I have a PR for the former and @kgpai is working on the latter.
With that we can then go to newer CLang versions (plan it to go to CLang 20).

Until then we have the CI that has scheduled runs with CLang 15 which we added a while back. If we are ok with it failing for now until we go to a newer CLang that's fine. Or we can merge this PR and revert later to get it going?

xin-zhang2:
@czentgr We don't have to revert this PR if it is merged, since it works with both clang 15 and newer versions.

mbasmanova:
> We are working on using newer CLang compilers. But we also need some more dependency upgrades - notably for simdjson and fmt. 

@czentgr Thank you for the context. Is there a GitHub issue that is tracking that effort? Curious, why is it hard / taking long time?

I'm concerned about not being able to use C++ 20 features in Velox and not having CI in Velox to catch such usage. If we have to support clang15 in Velox, then we need to add CI that uses that. Otherwise, perfectly green PRs will keep breaking Prestissimo builds and require "blind" patching where the PR fixing somethings that appears to be working fine even without the changes.

What do you think?

czentgr:
@mbasmanova The CI does use Clang15 but it doesn't run for every PR (https://github.com/facebookincubator/velox/actions/workflows/scheduled.yml?query=event%3Aschedule). It has been broken for a few days now due to this issue. From a developer point of view it could be run manually, if there is concern but of course that is not that user friendly.

The effort is basically a roundabout way of making Clang the default in the CI and add additional sanity testing.
For this we have an issue: https://github.com/facebookincubator/velox/issues/2125 . This details the requirements a bit more.
Clang15 does support certain C++20 features but not all. At the time we found it supports the main desired ones (co-routines and chrono) so it was fine to go to C++20.

Either way, would you be ok with getting merged now and it possibly getting reverted later when we moved up compiler versions?

@xin-zhang2 There was reason for the code to have this syntax by the original author. While it works it might generate different machine code (could be interesting to see if it does). So we should revert it back if merged once we have addressed the compiler issue.

mbasmanova:
@czentgr Thank you for clarifying.

> would you be ok with getting merged now and it possibly getting reverted later when we moved up compiler versions?

That's Ok with me.

czentgr:
Fixed by https://github.com/facebookincubator/velox/pull/15004. Closing.

mbasmanova:
CC: @kgpai

PR #14995, opened 2025-09-29
fix: Register companion functions in registerTDigestAggregate

MBkkt:
Otherwise companion functions don't registered for TDigest aggregation functions

### DIFF
--- a/velox/functions/prestosql/aggregates/TDigestAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/TDigestAggregate.cpp
@@ -356,7 +356,7 @@ void registerTDigestAggregate(
             hasWeight, hasCompression, resultTypes);
       },
       {true /*orderSensitive*/, false /*companionFunction*/},
-      false /*companionFunction*/,
+      withCompanionFunctions,
       overwrite);
 }
 } // namespace facebook::velox::aggregate::prestosql


velox/functions/prestosql/aggregates/TDigestAggregate.cpp
```diff
@@ -356,7 +356,7 @@ void registerTDigestAggregate(
             hasWeight, hasCompression, resultTypes);
       },
       {true /*orderSensitive*/, false /*companionFunction*/},
-      false /*companionFunction*/,
+      withCompanionFunctions,
```
kagamiori, line 359: Hi @MBkkt, could you please clarify why you would like to register companion functions of tdigest_agg? In Velox, there exists a `merge(tdigest) -> tdigest` function (https://facebookincubator.github.io/velox/functions/presto/tdigest.html#merge). I wonder if this fulfills your use case.

MBkkt: I don't have usecase for now, I just noticed that registerTDigest have unused parameter, there's no comment or something about this.
So I make this patch.

If you believe that companion functions for TDigest never needed, let's remove this and unused bool parameter from this function + comment in it's caller

WDYT?

kagamiori: Hi @MBkkt, got it. Thank you for the explanation. I think the companion functions of tdigest_agg would be unnecessary for normal use cases, so we can remove the unused bool parameter and just pass false here. This is what we did with registerQDigestAggAggregate too.

MBkkt: Hi @kagamiori , I make PR with this https://github.com/facebookincubator/velox/pull/15027

PR #14996, opened 2025-09-29
fix: Add includes for -inl.h files

MBkkt:
We need to include `#pragma once` and `#include header.h` in header-int.h to enable autocomplete in clangd-based IDEs (e.g. VSCode).

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -299,6 +299,8 @@ macro names are always upper-snake-case. Also:
     the interface. Only if someone needs to understand your implementation
     should opening the `-inl.h` be necessary (or the .cpp file, for that
     matter).
+  * Add `#pragma once` and `#include header.h` in `header-inl.h`, otherwise
+    clangd based autocomplete won't work.
 
 ## Function Arguments
 
--- a/velox/common/base/SimdUtil-inl.h
+++ b/velox/common/base/SimdUtil-inl.h
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#pragma once
 
 #include <numeric>
+#include "velox/common/base/SimdUtil.h"
 
 #if XSIMD_WITH_NEON
 namespace xsimd::types {
--- a/velox/exec/VectorHasher-inl.h
+++ b/velox/exec/VectorHasher-inl.h
@@ -13,6 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#pragma once
+
+#include "velox/exec/VectorHasher.h"
+
 namespace facebook::velox::exec {
 
 template <typename T>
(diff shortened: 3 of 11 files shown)


mbasmanova:
@MBkkt Would you update PR description to explain why this is needed?

CODING_STYLE.md
```diff
@@ -299,6 +299,8 @@ macro names are always upper-snake-case. Also:
     the interface. Only if someone needs to understand your implementation
     should opening the `-inl.h` be necessary (or the .cpp file, for that
     matter).
+  * Add `#pragma once` and `#include header.h` in `header-inl.h`, otherwise
```
mbasmanova, line 302: CC: @pedroerp @xiaoxmeng @rui-mo @majetideepak @aditi-pandit @amitkdutta @Yuhta

MBkkt:
@mbasmanova done

mbasmanova (approved):
Thanks.

velox/common/base/SimdUtil-inl.h
```diff
@@ -13,8 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#pragma once
 
 #include <numeric>
+#include "velox/common/base/SimdUtil.h"
```
Yuhta: It's not included on purpose so that people cannot include `-inl.h` directly without including the public header file.

MBkkt: Do you have suggestions how to fix this other way?
image
vs
image

It's not only about red highlight, navigation, etc also doesn't work.

MBkkt: @Yuhta if you think this behavior is required it can be done this way (with additional boilerplate code)

In public header
```
#define ALLOW_IMPL
#include "-inl.h"
```

In impl header

```
#ifdef ALLOW_IMPL
#undef ALLOW_IMPL
#else
#error "You shouldn't include impl header directly"
#endif
```

Yuhta: Yeah we can use `ALLOW_IMPL`, let's name space it to `VELOX_ALLOW_INCLUDE_HEADER_IMPL` (sorry I am paranoid about macro name conflict in headers)

MBkkt: ok

MBkkt: @Yuhta I tried something like this. Unfortunately this doesn't work (it allows include implementation header directly)

image

I mean there's two options
1. It's impossible to include without include public header
2. It's errors for clangd
because clangd relies on same build options as real code.

Yuhta: Then this means we need to rely on code review to prevent people from including `-inl.h` directly.  Trade off between making `clangd` happy or making code review easier.

MBkkt: We can make precommit check, that if include starts with velox and ends with -inl.h
It should be in header with same name, WDYT?

Yuhta: Yeah I think that works

PR #15002, opened 2025-09-30
feat[Aggregate Function]: Add reservoir_sample aggregate function

maniloya:
# Summary:
Add reservoir_sample aggregate function

# Details:
Reservoir sample function allow you to select a fixed-size random sample from a dataset, ensuring each record has an equal chance of being chosen.

# Testing:
Added test cases for aggregate function.
Verified that all existing and new tests pass.

# Impact:
No impact on other functionalities.


velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+namespace {
+
+constexpr int32_t kUninitializedMarker = -1;
+
+struct ReservoirSampleAccumulator {
+  ValueList samples;
+  ValueList initialSamples;
+  int64_t initialSeenCount = 0;
+  int64_t processedCount = 0;
+  int32_t maxSampleSize = 0;
```
kgpai: nit: size_t instead of int32_t .

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+      }
+    }
+  }
+
+ private:
+  uint64_t computeValueHash(
+      const DecodedVector& decodedVector,
+      vector_size_t row) const {
+    // Compute hash based on the value to ensure deterministic sampling
+    if (decodedVector.isNullAt(row)) {
```
kgpai: nit: dont we need to validate that row is < size of decodedVector etc ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+      return 0; // Consistent hash for null values
+    }
+
+    auto baseVector = decodedVector.base();
+    auto index = decodedVector.index(row);
+
+    // Simple hash based on vector type and value
+    switch (baseVector->typeKind()) {
+      case TypeKind::BIGINT:
```
kgpai: Use TypeTraits<TypeKind>::nativeType to get valueAt instead of this switch statement.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+      case TypeKind::INTEGER:
+        return static_cast<uint64_t>(
+            baseVector->as<FlatVector<int32_t>>()->valueAt(index));
+      case TypeKind::SMALLINT:
+        return static_cast<uint64_t>(
+            baseVector->as<FlatVector<int16_t>>()->valueAt(index));
+      case TypeKind::TINYINT:
+        return static_cast<uint64_t>(
+            baseVector->as<FlatVector<int8_t>>()->valueAt(index));
+      case TypeKind::REAL: {
```
kgpai: curious why hash bits for real/double ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+      case TypeKind::VARCHAR: {
+        auto stringView =
+            baseVector->as<FlatVector<StringView>>()->valueAt(index);
+        return folly::Hash{}(stringView);
+      }
+      default:
+        // For complex types (ROW, ARRAY, etc.), use the value's hash if
+        // available Otherwise, use a combination of type and position
+        return bits::hashMix(
+            static_cast<uint64_t>(baseVector->typeKind()),
```
kgpai: This means that every complex type will give same hash , since you are accessing the the typeKind and not even the type here..

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+  bool initialized = false;
+
+  explicit ReservoirSampleAccumulator() {}
+
+  void addValueToReservoir(
+      const DecodedVector& decodedVector,
+      vector_size_t row,
+      HashStringAllocator* allocator) {
+    if (samples.size() < static_cast<size_t>(maxSampleSize)) {
+      samples.appendValue(decodedVector, row, allocator);
```
kgpai: nit; Velox mostly prefers if(..) { do something ; return;} rather than if (..) {..} else {...} so that its obvious that its an early return.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+        vector_size_t replaceIndex = randomPos - 1;
+
+        ValueList newSamples;
+        ValueListReader reader(const_cast<ValueList&>(samples));
+
+        for (vector_size_t i = 0; i < samples.size(); i++) {
+          if (i == replaceIndex) {
+            newSamples.appendValue(decodedVector, row, allocator);
+          } else {
+            auto tempVector = BaseVector::create(
```
kgpai: Are you going to be creating a new Vector for every sample ? Why create a temp vector and decode to append a value to newSamples ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+    return child->asArray().elementType();
+  }
+
+  vector_size_t calculateTotalSampleElements(char** groups, int32_t numGroups)
+      const {
+    vector_size_t total = 0;
+    for (int32_t i = 0; i < numGroups; i++) {
+      auto* accumulator = value<ReservoirSampleAccumulator>(groups[i]);
+      if (!accumulator->initialized) {
+        const_cast<ReservoirSampleAccumulator*>(accumulator)->initialized =
```
kgpai: Curious, what is the point of the initialized field if you just set it to true like this ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+    for (int32_t i = 0; i < numGroups; i++) {
+      auto* accumulator = value<ReservoirSampleAccumulator>(groups[i]);
+
+      rawOffsets[i] = elementIndex;
+      rawSizes[i] = accumulator->samples.size();
+
+      // Copy samples from ValueList to output vector
+      if (accumulator->samples.size() > 0) {
+        ValueListReader reader(const_cast<ValueList&>(accumulator->samples));
+        for (vector_size_t j = 0; j < accumulator->samples.size(); j++) {
```
kgpai: Do you need to worry about handling nulls from the valueList ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,759 @@
...
+    auto elementType = getElementTypeFromIntermediate();
+
+    auto rowType = ROW(
+        {{"initialSample", ARRAY(elementType)},
+         {"initialSeenCount", BIGINT()},
+         {"seenCount", BIGINT()},
+         {"maxSampleSize", INTEGER()},
+         {"sample", ARRAY(elementType)}});
+
+    // Calculate total elements for both arrays
```
kgpai: nit: Its much nicer to add comments on the business logic (for e.g why ROW type above has sample, seen count, etc, and how reservoir sampling works, compared to comments like this since these are more evident.

kgpai (commented):
It would be nicer to split up the accumulator code into a separate file from the aggregate function.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,156 @@
...
+  bool isInitialized() const {
+    return maxSampleSize >= 0;
+  }
+
+  void addValueToReservoir(
+      const DecodedVector& decodedVector,
+      vector_size_t row,
+      HashStringAllocator* allocator) {
+    if (samples.size() < maxSampleSize) {
+      // LOG(INFO) << "[46] value : " << decodedVector.valueAt<>(row);
```
kgpai: nit: please remove comment.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,156 @@
...
+
+  bool shouldReplaceSample(
+      const DecodedVector& decodedVector,
+      vector_size_t row,
+      int64_t& replaceIndex) {
+    uint64_t valueHash = computeValueHash(decodedVector, row);
+    uint64_t seed =
+        bits::hashMix(valueHash, static_cast<uint64_t>(processedCount));
+
+    std::mt19937_64 rng(seed);
```
kgpai: I am curious do you have to create a new seed everytime you need to check if a sample should be replaced ? If not rng /dist can be method variables and should make this much faster.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,156 @@
...
+    replaceIndex = randomPos - 1;
+    return true;
+  }
+
+  void replaceSampleAt(
+      const DecodedVector& newValue,
+      vector_size_t newValueRow,
+      int64_t replaceIndex,
+      HashStringAllocator* allocator) {
+    auto sampleSize = samples.size();
```
kgpai: This is going to be tremendously inefficient if we have to rewrite all the samples each time we need to replace a value. Can we try use RowContainer here ?

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    auto typeKind = decodedVector.base()->typeKind();
+
+    if (typeKind == TypeKind::BOOLEAN) {
+      return decodedVector.valueAt<bool>(row) ? 1 : 0;
+    }
+
+    return VELOX_DYNAMIC_TYPE_DISPATCH(
+        computeValueHashTyped, typeKind, decodedVector, row);
```
kgpai: What is this hash used for, it only looks at physical representation so custom types which might be same actually might hash different due to this.

kgpai (commented):
It might be good to split this into two PRs:
1. First just support non complex types , so that you can rewrite the value in a vector without having to rewrite appendlists.

2. if you need to support complex types, then use rowserializer to serialize the complex type to a bytestring (and back) and track that in an accumulator / vector .

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+#include "velox/vector/DecodedVector.h"
+
+namespace facebook::velox::aggregate::prestosql {
+
+struct ReservoirSampleAccumulator {
+  VectorPtr samples;
+  VectorPtr initialSamples;
+  int64_t initialSeenCount = 0;
+  int64_t processedCount = 0;
+  int32_t maxSampleSize = -1;
```
kgpai, line 34: nit: Its nicer to have it as optional<uint32_t> maxSampleSize;

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+
+  explicit ReservoirSampleAccumulator(memory::MemoryPool* pool) : pool_(pool) {}
+
+  bool isInitialized() const {
+    return maxSampleSize >= 0;
```
kgpai, line 39: nit: return maxSampleSize.has_value();

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+
+    int64_t replaceIndex = -1;
+    if (!shouldReplaceSample(decodedVector, row, replaceIndex)) {
+      return;
+    }
+
+    replaceValueAt(decodedVector, row, replaceIndex);
+  }
+
+  bool shouldReplaceSample(
```
kgpai, line 69: nit: any reason why shouldReplaceSample is public ?

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+      const DecodedVector& decodedVector,
+      vector_size_t row,
+      int64_t& replaceIndex) const {
+    std::mt19937_64 rng(static_cast<uint64_t>(processedCount));
```
kgpai: Do you need to create a new rng on every call to determine if you should replace a sample ? This construction is not cheap, You can instead create the uniform distribution each time and just reuse the rng ? Another way to do it is to just have an an rng between 0 to 1 and scale it each time with processedCount (?).

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+
+  void appendValue(
+      const DecodedVector& decodedVector,
+      vector_size_t row,
+      const TypePtr& type) {
+    if (!samples) {
+      samples = BaseVector::create(type, 0, pool_);
+    }
+    auto currentSize = samples->size();
+    samples->resize(currentSize + 1);
```
kgpai, line 94: Why not create a vector with maxSamplesize capacity by calling reserve rather than call resize all the time ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+  bool isFixedSize() const override {
+    return false;
+  }
+
+  void extractValues(char** groups, int32_t numGroups, VectorPtr* result)
+      override {
+    auto* pool = allocator_->pool();
+    auto arrayType = this->resultType()->childAt(1);
+    auto elementType = arrayType->asArray().elementType();
+    auto rowType = ROW({BIGINT(), arrayType});
```
kgpai: nit:This can be initialized when the aggregate is being constructed.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+    auto sizes = allocateSizes(numGroups, pool);
+    auto* rawOffsets = offsets->asMutable<vector_size_t>();
+    auto* rawSizes = sizes->asMutable<vector_size_t>();
+
+    vector_size_t offset = 0;
+    for (int32_t i = 0; i < numGroups; i++) {
+      auto* accumulator = value<ReservoirSampleAccumulator>(groups[i]);
+      auto sampleSize = accumulator->sampleCount();
+
+      processedCounts->set(i, accumulator->processedCount);
```
kgpai: nit: if you are setting rawoffset, rawsize, you can also get raw buffer value for processedCount and set it directly.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+        std::vector<VectorPtr>{processedCountVector, sampleVector});
+  }
+
+  void extractAccumulators(char** groups, int32_t numGroups, VectorPtr* result)
+      override {
+    auto* pool = allocator_->pool();
+    auto elementType = this->resultType()->childAt(1)->asArray().elementType();
+
+    auto rowType = ROW(
+        {{"Sample", ARRAY(elementType)},
```
kgpai: nit: name it "sample" to stick with naming convention.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
```diff
@@ -0,0 +1,108 @@
...
+namespace facebook::velox::aggregate::prestosql {
+
+struct ReservoirSampleAccumulator {
+  VectorPtr samples;
+  VectorPtr initialSamples;
+  int64_t initialSeenCount = 0;
+  int64_t processedCount = 0;
+  int32_t maxSampleSize = -1;
+
+  explicit ReservoirSampleAccumulator(memory::MemoryPool* pool) : pool_(pool) {}
```
kgpai: nit: I am not sure if its good practice for an accumulator to hold reference to memory pool - maybe its not a problem, but i need to dig into this a bit.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+      auto elementVector = arrayVector->elements();
+      accumulator->initialSamples =
+          BaseVector::create(elementType, arraySize, allocator_->pool());
+      accumulator->initialSamples->resize(arraySize);
+      accumulator->initialSamples->copy(
+          elementVector.get(), 0, arrayOffset, arraySize);
+      accumulator->samples =
+          BaseVector::create(elementType, arraySize, allocator_->pool());
+      accumulator->samples->resize(arraySize);
+      accumulator->samples->copy(
```
kgpai: nit: wont the copy resize samples ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+#include "velox/functions/prestosql/aggregates/ReservoirSampleAggregate.h"
+#include "velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h"
+#include "velox/type/Type.h"
+#include "velox/vector/FlatVector.h"
+
+namespace facebook::velox::aggregate::prestosql {
+
+namespace {
+
+class ReservoirSampleAggregate : public exec::Aggregate {
```
kgpai, line 27: nit: For such long classes, it helps if the class is defined first with its methods and then the implementations follow outside the class definition .

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+      auto* accumulator = value<ReservoirSampleAccumulator>(groups[row]);
+      auto decodedRow = decodedIntermediate.index(row);
+
+      int32_t maxSampleSize = maxSampleSizeVector->valueAt(decodedRow);
+      if (maxSampleSize < 0) {
+        return;
+      }
+
+      if (!accumulator->isInitialized()) {
+        new (accumulator) ReservoirSampleAccumulator(allocator_->pool());
```
kgpai: accumulator is still a valid object right, so why recreate a placement new on previously created accumulator ? Might this not be dangerous ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,663 @@
...
+    }
+
+    auto sampleElements = sampleArrayVector->elements();
+    DecodedVector decodedSampleElements(*sampleElements);
+
+    if (otherProcessedCount < accumulator->maxSampleSize) {
+      for (vector_size_t i = 0; i < otherSampleSize; i++) {
+        accumulator->processedCount++;
+        auto elementIndex = otherSampleOffset + i;
+        accumulator->addValueToReservoir(
```
kgpai: When merging two accumulators this call will end up being very expensive, and would need to be optimized..

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
maniloya: I think VectorPtr doesn't have reserve method.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
maniloya: appendValue() method in the accumulator is the place pool is being used, modified the code to pass as a function parameter.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
maniloya: I think VectorPtr->copy expects the memory to be available for copy and it doesn't resize

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
maniloya: used folly::ThreadLocalPRNG instead of defining a new rng everytime withing a thread.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
maniloya: updated the code to reuse the accumulator.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
maniloya: Removed the rng creation by using ThreadLocal PRNG from folly

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
maniloya: How in this case we are creating the BaseVector with required size, we can avoid resize here and all the cases that follow the same pattern. Updated the code.

mbasmanova:
@maniloya The original PR description has a lot of redundant text. The amount of useful information was less than 30%. The PR title repeated "aggregate" twice. I updated the title and description. Please, review these as well as contributing guidelines and follow.

velox/functions/prestosql/aggregates/tests/ReservoirSampleTest.cpp
```diff
@@ -0,0 +1,247 @@
...
+namespace facebook::velox::aggregate::prestosql {
+
+namespace {
+
+class ReservoirSampleTest : public AggregationTestBase {
+ protected:
+  void SetUp() override {
+    AggregationTestBase::SetUp();
+    disableTestStreaming();
+    disableTestIncremental();
```
mbasmanova, line 33: Why are you disabling these?

velox/functions/prestosql/fuzzer/WindowFuzzerTest.cpp
```diff
@@ -138,6 +138,7 @@ int main(int argc, char** argv) {
       "merge",
       // https://github.com/facebookincubator/velox/issues/14423
       "numeric_histogram",
+      "reservoir_sample",
```
mbasmanova, line 141: Why are you excluding these? 

It would be nice to mention and explain this and other gaps in test coverage in the PR description.

maniloya: Reservoir_sample contains randomness in the result generated so we are skipping the comparison with presto java version.

velox/functions/prestosql/aggregates/tests/ReservoirSampleTest.cpp
maniloya: This aggregate has intermediateResultType is different from finalResultType. This is creating issues in testing.

velox/functions/prestosql/aggregates/AggregateNames.h
```diff
@@ -72,6 +72,7 @@ const char* const kRegrSxy = "regr_sxy";
 const char* const kRegrSxx = "regr_sxx";
 const char* const kRegrSyy = "regr_syy";
 const char* const kRegrR2 = "regr_r2";
+const char* const kReservoirSample = "reservoir_sample";
```
mbasmanova, line 75: Please, add documentation: https://facebookincubator.github.io/velox/functions/presto/aggregate.html

velox/functions/prestosql/fuzzer/WindowFuzzerTest.cpp
mbasmanova: What's the source of randomness? Would you explain in detail? Can it be removed? It will be difficult to ensure correctness if the function is non-deterministic. 

CC: @gggrace14

velox/functions/prestosql/aggregates/tests/ReservoirSampleTest.cpp
mbasmanova: What kind of issues? Please, explain in detail and explore ways to address these as opposed to just lose test coverage.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
natashasehgal: What if you use ValueList instead of a VectorPtr. This stores variable-length list of values and doesn't require resizing

e.g https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp#L213

https://facebookincubator.github.io/velox/develop/aggregate-functions.html#accumulator

natashasehgal: Or even VectorPtr + pre-sizing instead of starting from size 0. If maxSampleSize is not large 

@kgpai any preference in the two approaches?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,722 @@
...
+#include "velox/functions/prestosql/aggregates/ReservoirSampleAggregate.h"
+#include "velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h"
+#include "velox/type/Type.h"
+#include "velox/vector/FlatVector.h"
+
+namespace facebook::velox::aggregate::prestosql {
+
+namespace {
+
+// Class definition
```
gggrace14: Remove this comment?

velox/functions/prestosql/fuzzer/WindowFuzzerTest.cpp
gggrace14: Mani, there're several typical sources of nondeterminism, like order of inputs getting to a node, environment dependency, etc. What is it currently in your case? Have you checked a couple of other "random" functions as examples? Do they have a way to make it "idempotent" as we once chatted about, which is good enough for fuzzer test? Or do they have similar source of nondeterminism?

maniloya: The randomness is from the sampling that the algorithm performs, due to the randomness there will be a result mismatch between java and cpp execution. Source of randomness originates when the reservoir is full and there is a new value that need to be added. New value is replaced with existing value at a random index. Will add ResultVerifier & InputGenerator, to do the following verification of results in the following way rather than directly comparing the results.
1. ProcessedCount returned should be same for both executions
2. Output samples for each group need to be present in the input values

I was premature to avoid this earlier. Thanks for the pointing it out.

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
maniloya: reservoir_sample requires replacement of values in the reservoir at index i. ValueList doesn't support replacement at index i, this would require us to copy entire Valuelist for a single replacement. 
Also we dont have control/estimate on how large the maxSampleSize is, it is defined by the users.
@kgpai

velox/functions/prestosql/fuzzer/WindowFuzzerTest.cpp
mbasmanova: > The randomness is from the sampling that the algorithm performs

Would you clarify some more? Typically, randomness can be controlled with a seed. Is this not the case here?

velox/functions/prestosql/aggregates/ReservoirSampleAccumulator.h
kgpai: Yes presizing would work , ValueList semantics wont work here. The maxSampleSize is still available when the udf is invoked so you can preallocate Vector size then.

PR #15010, opened 2025-09-30
fix: Addup expression stats in PlanNodeStats::operator+=()

MnO2:
Summary: The logic of adding up expressionStats is missing from [PlanNodeStats::operator+=()](https://github.com/facebookincubator/velox/blob/main/velox/exec/PlanNodeStats.cpp#L22). It is present in [PlanNodeStats::addTotals()](https://github.com/facebookincubator/velox/blob/main/velox/exec/PlanNodeStats.cpp#L94). Hence, the result is correct when adding up OperatorStats, but incorrect when adding up PlanNodeStats.

Differential Revision: D83536288

### DIFF
--- a/velox/exec/PlanNodeStats.cpp
+++ b/velox/exec/PlanNodeStats.cpp
@@ -59,6 +59,14 @@ PlanNodeStats& PlanNodeStats::operator+=(const PlanNodeStats& another) {
     }
   }
 
+  for (const auto& [name, exprStats] : another.expressionStats) {
+    auto const [it, inserted] =
+        this->expressionStats.try_emplace(name, exprStats);
+    if (!inserted) {
+      it->second.add(exprStats);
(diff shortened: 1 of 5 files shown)


velox/exec/PlanNodeStats.cpp
```diff
@@ -59,6 +59,14 @@ PlanNodeStats& PlanNodeStats::operator+=(const PlanNodeStats& another) {
     }
   }
 
+  for (const auto& [name, exprStats] : another.expressionStats) {
+    if (UNLIKELY(this->expressionStats.count(name) == 0)) {
+      this->expressionStats.insert(std::make_pair(name, exprStats));
```
MBkkt:
```suggestion
      this->expressionStats.emplace(name, exprStats);
```

PR #15012, opened 2025-10-01
qdigest_agg - Update intermediate type to VARBINARY

natashasehgal:
Differential Revision: D83618668


natashasehgal:
@kagamiori can you help confirm the Signature Change in fuzzer failing is ok. As it's an intentional signature change. And this test won't fail after the signature is changed

PR #15013, opened 2025-10-01
refactor: Expose a utility function in MakeRowFromMap

bikramSingh91:
Summary: This is a simple refactor to expose the utility function inside MakeRowFromMap that adds a default value to all rows of an input.

Differential Revision: D83629467


velox/functions/lib/MakeRowFromMap.h
```diff
@@ -23,6 +23,40 @@
 
 namespace facebook::velox::functions {
 
+namespace make_row_from_map {
```
mbasmanova: This is unconventional. We don't do this in the codebase. Let's put this helper function in a class named MakeRowFromMapDefaults or similar.

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -23,6 +23,40 @@
...
+// Utility function to fill an input vector of flat encoding with default values
```
mbasmanova: > Utility function to fill

Drop "Utility function" and start with an active verb: "Fills ..."

> input vector of flat encoding

this may need clarification as MapVector::encoding != flat. We can just say "input vector" and then clarify that the vector must be writable (which implies flat in a broad sense).

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -23,6 +23,40 @@
...
+// based on it type. For primitive types, it will be set to the c++ default,
+// like 0 for INTEGER and empty string for VARCHAR. For Array/Map they will be
+// set to an empty array or map. Not supported for ROW or UNKNOWN types.
+template <TypeKind ValueKind>
+static void fillWithDefaultValue(VectorPtr& vector) {
```
mbasmanova: We have 2 use cases for this logic: (1) make a flat vector will default values; (2) make a constant vector. Perhaps, add 2 separate functions: 
- MakeRowFromMapDefaults::createFlat(type, size) -> VectorPtr 
- MakeRowFromMapDefaults::createConstant(type, size) -> VectorPtr

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -23,6 +23,40 @@
...
+  VELOX_CHECK_NOT_NULL(vector);
+  if constexpr (
+      (TypeTraits<ValueKind>::isPrimitiveType ||
+       ValueKind == TypeKind::OPAQUE) &&
+      ValueKind != TypeKind::UNKNOWN) {
+    using NativeType = typename TypeTraits<ValueKind>::NativeType;
```
mbasmanova: NativeType -> T

velox/functions/lib/MakeRowFromMap.h
```diff
@@ -23,6 +23,64 @@
...
+// types, it will be set to the c++ default, like 0 for INTEGER and empty string
+// for VARCHAR. For Array/Map they will be set to an empty array or map. Not
+// supported for ROW or UNKNOWN types.
+class MakeRowFromMapDefaults {
+ public:
+  static VectorPtr createFlat(
+      const TypePtr& type,
+      const vector_size_t size,
+      memory::MemoryPool& pool,
+      VectorPool* vectorPool) {
```
mbasmanova, line 38: Would you document this API? specifically, let's clarify whether vectorPool can be null or not.

PR #15024, opened 2025-10-01
feat: Add simple function support for TIME WITH TIME ZONE

kgpai:
Summary: Adds support for simple functions to use TIME WITH TIME ZONE type.

Differential Revision: D83703684


kevinwilfong (approved):
Thanks!

PR #15027, opened 2025-10-02
fix: Do not allow registering companion functions for TDigest aggregate functions

MBkkt:
> I think the companion functions of tdigest_agg would be unnecessary for normal use cases, so we can remove the unused bool parameter and just pass false here. This is what we did with registerQDigestAggAggregate too.

https://github.com/facebookincubator/velox/pull/14995#discussion_r2395964880


kagamiori (approved):
Thank you!

PR #15028, opened 2025-10-02
fix(cache): Fix calculation of write rate in SsdCache

lingbin:
Time is measured in 'microseconds', not 'seconds', so we need to 
multiply by 1,000,000 when calculating write rate.


velox/common/caching/SsdCache.cpp
```diff
@@ -135,9 +136,10 @@ void SsdCache::write(std::vector<CachePin> pins) {
         // Typically occurs every few GB. Allows detecting unusually slow rates
         // from failing devices.
         VELOX_SSD_CACHE_LOG(INFO) << fmt::format(
-            "Wrote {}, {} bytes/s",
+            "Wrote {} to SSD, {} bytes/s",
             succinctBytes(bytes),
-            static_cast<float>(bytes) / (getCurrentTimeMicro() - startTimeUs));
+            static_cast<double>(bytes) * 1000'000 /
```
majetideepak: nit: ` 1'000'000`

lingbin: Fixed, thanks.

PR #15029, opened 2025-10-02
feat: Add TIME type support to Arrow bridge

PingLiuPing:
Export Velox TIME (int64 millis) as Arrow time32 (int32 millis, format "ttm").
Import Arrow time32/time64 to Velox TIME, supporting all time units.

This enables writing TIME to Parquet with TIME_MILLIS logical type and INT32
physical type.

Parquet metadata for TIME columns:
- Physical type: INT32
- Logical type: TIME

Example:
■■■■■■■■SchemaElement
■■■■■■■■■■■■type = 1
■■■■■■■■■■■■repetition_type = 1
■■■■■■■■■■■■name = c_time
■■■■■■■■■■■■converted_type = 7
■■■■■■■■■■■■logicalType = LogicalType
■■■■■■■■■■■■■■■■TIME = TimeType
■■■■■■■■■■■■■■■■■■■■isAdjustedToUTC = True
■■■■■■■■■■■■■■■■■■■■unit = TimeUnit
■■■■■■■■■■■■■■■■■■■■■■■■MILLIS = MilliSeconds

### DIFF
--- a/velox/vector/arrow/Bridge.cpp
+++ b/velox/vector/arrow/Bridge.cpp
@@ -286,6 +286,11 @@ const char* exportArrowFormatStr(
       }
       return "i"; // int32
     case TypeKind::BIGINT:
+      if (type->isTime()) {
+        // TIME is stored as milliseconds since midnight in Velox.
+        // Export as Arrow time32 with milliseconds unit.
+        return "ttm";
+      }
       return "l"; // int64
     case TypeKind::REAL:
       return "f"; // float32
@@ -384,6 +389,30 @@ struct Selection {
   std::optional<std::vector<std::pair<vector_size_t, vector_size_t>>> ranges_;
   vector_size_t total_;
 };
+
+// Gather values from time buffer. Nulls are skipped.
+// TIME is stored as int64_t milliseconds in Velox, converted to int32_t
+// milliseconds for Arrow time32.
+void gatherFromTimeBuffer(
+    const BaseVector& vec,
+    const Selection& rows,
+    Buffer& out) {
+  auto src = (*vec.values()).as<int64_t>();
+  auto dst = out.asMutable<int32_t>();
+  vector_size_t j = 0;
+  if (!vec.mayHaveNulls() || vec.getNullCount() == 0) {
+    // Convert int64 milliseconds to int32 milliseconds.
+    rows.apply(
+        [&](vector_size_t i) { dst[j++] = static_cast<int32_t>(src[i]); });
+    return;
+  }
+  rows.apply([&](vector_size_t i) {
+    if (!vec.isNullAt(i)) {
+      dst[j] = static_cast<int32_t>(src[i]); // int64 to int32 milliseconds.
+    }
+    j++;
+  });
+}
 
 // Gather values from timestamp buffer. Nulls are skipped.
 void gatherFromTimestampBuffer(
@@ -682,7 +711,8 @@ bool isFlatScalarZeroCopy(const TypePtr& type) {
   // mapped to Arrow Decimal128.
   // - Velox's Timestamp representation (2x 64bit values) does not have an
   // equivalent in Arrow.
-  return !type->isShortDecimal() && !type->isTimestamp();
+  // - Velox's TIME is in milliseconds, Arrow time64 is in microseconds.
+  return !type->isShortDecimal() && !type->isTimestamp() && !type->isTime();
 }
 
 // Returns the size of a single element of a given `type` in the target arrow
@@ -692,6 +722,9 @@ size_t getArrowElementSize(const TypePtr& type) {
     return sizeof(int128_t);
   } else if (type->isTimestamp()) {
     return sizeof(int64_t);
+  } else if (type->isTime()) {
+    // TIME is exported as Arrow time32 (int32_t).
+    return sizeof(int32_t);
   } else {
     return type->cppSizeInBytes();
   }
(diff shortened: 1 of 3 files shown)


velox/vector/arrow/Bridge.cpp
```diff
@@ -286,6 +286,11 @@ const char* exportArrowFormatStr(
       }
       return "i"; // int32
     case TypeKind::BIGINT:
+      if (type->isTime()) {
+        // TIME is stored as milliseconds since midnight in Velox.
+        // Export as Arrow time64 with microseconds unit.
+        return "ttu";
```
pedroerp: why don't we export using ttm (microsecond) to use the exact same value?

PingLiuPing: @pedroerp  Thank you for the comment.
I think you are saying why don't we using ttm (millisecond). I think that is reasonable too. The reason I'm using microsecond is I see Presto java and Trino both write time to parquet file with physical type `int64` and logical type `TimeType` with `TimeUnit=MicroSeconds`.

pedroerp: I think here it would make more sense to export as ttm by default, since this is the same precision used by TIME in Velox vectors. So the conversion can be simpler and zero-copy. 

Clients of this API can always use Arrow or have custom code to convert to other precisions.

PingLiuPing: @pedroerp  Thanks. I change it to ttm (millisecond) now.

velox/vector/arrow/Bridge.cpp
```diff
@@ -385,6 +390,30 @@ struct Selection {
   vector_size_t total_;
 };
 
+// Gather values from time buffer. Nulls are skipped.
+// TIME is stored as int64_t milliseconds in Velox, converted to int32_t
+// milliseconds for Arrow time32.
+void gatherFromTimeBuffer(
```
pedroerp, line 396: now that we don't want to do any conversion during the gather, seems like we don't need this function and can only rely on `gatherFromBuffer()`, no?

Unless you are also planning to implement something like `option.timestampUnit` to enable API users to choose the precision they want.

velox/vector/arrow/Bridge.cpp
```diff
@@ -682,7 +711,8 @@ bool isFlatScalarZeroCopy(const TypePtr& type) {
   // mapped to Arrow Decimal128.
   // - Velox's Timestamp representation (2x 64bit values) does not have an
   // equivalent in Arrow.
-  return !type->isShortDecimal() && !type->isTimestamp();
+  // - Velox's TIME is in milliseconds, Arrow time64 is in microseconds.
+  return !type->isShortDecimal() && !type->isTimestamp() && !type->isTime();
```
pedroerp, line 715: you also probably don't want this anymore

velox/vector/arrow/Bridge.cpp
```diff
@@ -1544,6 +1582,32 @@ TypePtr importFromArrow(const ArrowSchema& arrowSchema) {
 
 namespace {
 
+enum class TimeUnit { kSecond, kMilli, kMicro, kNano };
```
pedroerp, line 1585: do we also want to expose this to API users (through ArrowOptions)?

velox/vector/arrow/Bridge.cpp
PingLiuPing: @pedroerp Thanks for the comment.
I think this customised conversion is needed because velox stores time as int64, but we need to export it to int32 which is 4 bytes, we need to cast the value to int32.

velox/vector/arrow/Bridge.cpp
PingLiuPing: @pedroerp  Thanks for the comment.
This is a good point, but since currently velox TIME only indicates milliseconds precision. I don't see any requirement to export different precision TIME to arrow at the moment. But they are required when import value from arrow.

PingLiuPing:
@pedroerp  Could you help have another look please, thanks.

velox/vector/arrow/Bridge.cpp
pedroerp: You're right, thanks for clarifying.

velox/vector/arrow/Bridge.cpp
```diff
@@ -385,6 +390,30 @@ struct Selection {
...
 };
 
+// Gather values from time buffer. Nulls are skipped.
+// TIME is stored as int64_t milliseconds in Velox, converted to int32_t
+// milliseconds for Arrow time32.
+void gatherFromTimeBuffer(
+    const BaseVector& vec,
+    const Selection& rows,
+    Buffer& out) {
+  auto src = (*vec.values()).as<int64_t>();
```
pedroerp, line 400: could you do instead?

`auto src = vec.values()->as<int64_t>();`

velox/vector/arrow/Bridge.cpp
```diff
@@ -1835,6 +1899,126 @@ void setTimestamps(
...
+        break;
+      case TimeUnit::kMicro:
+        for (auto i = 0; i < length; ++i) {
+          if (!bits::isBitNull(rawNulls, i)) {
+            rawTimes[i] = static_cast<int64_t>(input[i]) / 1'000;
+          }
+        }
+        break;
+      case TimeUnit::kNano:
+        for (auto i = 0; i < length; ++i) {
```
pedroerp, line 1970: there is probably a way to extract these loops as a helper templated function with a constexpr if check on nullability.

PR #9822, opened 2024-05-15
Add checksum for ssd cache

zacw7:
This PR introduces an optional checksum feature for the SSD cache. When
enabled, a CRC-based checksum is calculated for each cache entry and
stored in the checkpoint file. Additionally, if read verification is
activated, the checksum is recalculated and verified against the stored
value when cache data is recovered from the checkpoint or loaded from
the SSD.
A new counter, `ssd_cache_read_corruptions`, is added to track the
number of corruptions detected due to checksum mismatches.

PR #15031, opened 2025-10-03
fix(cache): Fix comment of SsdRun

lingbin:
In #9822, an optional checksum feature was introduced to SsdCache, and 
the size of `SsdRun` is no longer 64 bytes. This PR corrects its 
comment and adds a description about the checksum.

No functional changes.

### DIFF
--- a/velox/common/caching/SsdFile.h
+++ b/velox/common/caching/SsdFile.h
@@ -32,16 +32,18 @@ class SsdFileTestHelper;
 class SsdCacheTestHelper;
 } // namespace test
 
-/// A 64 bit word describing a SSD cache entry in an SsdFile. The low 23 bits
-/// are the size, for a maximum entry size of 8MB. The high bits are the offset.
+/// The 'fileBits_' field is a 64 bit word describing a SSD cache entry in an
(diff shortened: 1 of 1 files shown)


lingbin:
@kKPulla Could you help review this PR? Thanks.

velox/common/caching/SsdFile.h
```diff
@@ -58,9 +60,11 @@ class SsdRun {
     checksum_ = other.checksum_;
   }
 
-  void operator=(SsdRun&& other) {
+  void operator=(SsdRun&& other) noexcept {
     fileBits_ = other.fileBits_;
     checksum_ = other.checksum_;
+    other.fileBits_ = 0;
```
majetideepak, line 66: Why do we reset the `other` values? We should allow the caller to do this based on the need.

lingbin: @majetideepak  It is the **Move** assignment operator. After this call, we should not use the "other" anymore(otherwise, it will be a "**use-after-move**" error), and we should make 'other' be an "**empty**" state.

PR #15038, opened 2025-10-03
feat: Add FlatMapVector support for Subscript

peterenescu:
Summary:
Adds FlatMapVector support for functions subscript or element_at. This will allow FlatMapVector-encoded maps to be read without conversion and have its values extracted from.

Additionally, this support will greatly increase performance as singular key values can be projected without full-map materialization.

Differential Revision: D83867312

### DIFF
--- a/velox/functions/lib/SubscriptUtil.cpp
+++ b/velox/functions/lib/SubscriptUtil.cpp
@@ -22,6 +22,7 @@
 #include "velox/common/memory/MemoryPool.h"
 #include "velox/functions/lib/SubscriptUtil.h"
 #include "velox/type/Type.h"
+#include "velox/vector/FlatMapVector.h"
 #include "velox/vector/TypeAliases.h"
 
 namespace facebook::velox::functions {
@@ -60,7 +61,7 @@ VectorPtr applyMapTyped(
     bool triggerCaching,
     std::shared_ptr<detail::LookupTableBase>& cachedLookupTablePtr,
     const SelectivityVector& rows,
-    const VectorPtr& mapArg,
+    const DecodedVector& decodedMap,
     const VectorPtr& indexArg,
     exec::EvalCtx& context) {
   static constexpr vector_size_t kMinCachedMapSize = 100;
@@ -85,10 +86,8 @@ VectorPtr applyMapTyped(
 
   // Get base MapVector.
   // TODO: Optimize the case when indices are identity.
-  exec::LocalDecodedVector mapHolder(context, *mapArg, rows);
-  auto decodedMap = mapHolder.get();
-  auto baseMap = decodedMap->base()->as<MapVector>();
-  auto mapIndices = decodedMap->indices();
+  auto baseMap = decodedMap.base()->as<MapVector>();
+  auto mapIndices = decodedMap.indices();
 
   // Get map keys.
   auto mapKeys = baseMap->mapKeys();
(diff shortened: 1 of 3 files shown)


velox/functions/prestosql/tests/ElementAtTest.cpp
```diff
@@ -232,6 +232,118 @@ TEST_F(ElementAtTest, mapWithDictionaryKeys) {
   }
 }
 
+TEST_F(ElementAtTest, flatMapTest) {
+  // Null map.
+  facebook::velox::test::assertEqualVectors(
```
pedroerp: nit: the common practice is to leave tests under a `facebook::velox::something::test` namespace (or simply `facebook::velox::test`), so here you can just use `test::assertEqualVectors`.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,6 +187,47 @@ VectorPtr applyMapTyped(
...
+    const VectorPtr& mapArg,
+    const VectorPtr& indexArg,
+    exec::EvalCtx& context) {
+  // Decode FlatMapVector.
+  exec::LocalDecodedVector decoder(context, *mapArg, rows);
+  auto decoded = decoder.get();
+  auto flatMap = decoded->base()->as<FlatMapVector>();
+
+  // Get vector projection.
+  if (auto projection = flatMap->projectKey(indexArg, 0)) {
```
pedroerp: I suppose this assumes that the index argument is a constant; do we support non-constant key argument for the other types of element_at() function?

For example, if instead of `element_at(col, 10)` you get `element_at(col, col2)`.

I think it's fine if we don't support it, but let's check for a constant and explicitly throw an exception in case you get something else.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,6 +187,47 @@ VectorPtr applyMapTyped(
...
+    // Define nulls and indices bufers.
+    BufferPtr indices =
+        AlignedBuffer::allocate<vector_size_t>(rows.size(), flatMap->pool());
+    BufferPtr nulls = allocateNulls(rows.size(), flatMap->pool());
+    auto mutableIndices = indices->asMutable<vector_size_t>();
+    auto rawNulls = nulls->asMutable<uint64_t>();
+    for (int i = 0; i < decoded->size(); i++) {
```
pedroerp: I think this function will be hot enough that it's worth to have two paths, one for the common case, when there are no wrappings, and one for the constant/dict case. 

The former one can be cheaper as you won't need to allocate and wrap the returned data in a dictionary, and you can also just reuse the same nulls buffer.

velox/functions/lib/SubscriptUtil.cpp
peterenescu: Good question, it looks like we do and I was missing the non-constant code path. Let me add

velox/functions/lib/SubscriptUtil.cpp
peterenescu: Makes sense. Let me add the two paths

velox/functions/lib/SubscriptUtil.cpp
pedroerp: Great. Let's keep them separated though, as the constant case will probably be the hot one.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -330,6 +434,20 @@ VectorPtr MapSubscript::applyMap(
   // Ensure map key type and second argument are the same.
   VELOX_CHECK(mapArg->type()->childAt(0)->equivalent(*indexArg->type()));
 
+  // Short-circuit for FlatMapVector encoding. FlatMapVector doesn't need to
+  // distinguish between primitive and complex types (where the former requires
+  // a type dispatch).
+  exec::LocalDecodedVector mapDecoder(context, *mapArg, rows);
```
pedroerp, line 397: can we avoid decoding the top level map vector more than once? I suppose we also have to decode it within the functions below.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,6 +187,109 @@ VectorPtr applyMapTyped(
...
+/// far simpler than the regular map encoding because FlatMapVector already
+/// supports feature projection. This implementation will serve as a fast-path
+/// execution for now-wrapped vectors.
+VectorPtr applyFlatMap(
+    const SelectivityVector& rows,
+    const VectorPtr& map,
+    const VectorPtr& elementAt,
+    exec::EvalCtx& context) {
+  // Decode input flat map vector.
+  exec::LocalDecodedVector mapDecoder(context, *map, rows);
```
pedroerp: seems like here you should just assert the vector is already decoded instead of having to decode it again.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,6 +187,109 @@ VectorPtr applyMapTyped(
...
+  exec::LocalDecodedVector elementAtDecoder(context, *elementAt, rows);
+  auto decodedElementAt = elementAtDecoder.get();
+
+  // Optimal use case: constant key. We can simply project the feature using the
+  // first value in the arg vector.
+  if (decodedElementAt->isConstantMapping()) {
+    if (auto projection = flatMap->projectKey(elementAt, 0)) {
+      return flatMap->projectKey(elementAt, 0);
+    }
+  } else {
```
pedroerp: if you get into this branch, then the optimization above won't help as much. Meaning that you could potentially organize the code as a single function as follows:

0. First decoded the vector. 
1. Then, if there are no wrappings and constant key, provide the fast-path at line 210 (hopefully this will be the common case)
2. Otherwise, a path that decodes the vector but still assumes a constant key (line 250).
3. A general slower path that uses the decoded vector wrapping and has to copy each element (line 273).

velox/functions/lib/SubscriptUtil.cpp
peterenescu: @pedroerp Yep that's why I kept it, to keep signature parity with the "regular" map implementation. I think there's a couple options here:
1. Refactor all functions to input a decoded vector (and pass the `apply`'s decoding into each).
2. Only refactor `applyFlatMap...` functions to take a decoded vector and leave the other former implementations alone.

pedroerp: If we always need to decode for all maps, then perhaps it's easier to decode on the parent function and propagate down the decoded vector. Unless there are cases where we optimize and don't need to fully decode it.

pedroerp: @peterenescu can you also propagate the decoded vector to applyMapTyped() and applyMapComplexType()? Otherwise we will double decode in these common situations.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,6 +187,77 @@ VectorPtr applyMapTyped(
...
+  // Decode input flat map vector.
+  auto flatMap = decodedMap.base()->as<FlatMapVector>();
+
+  // Decode input elementAt vector.
+  exec::LocalDecodedVector elementAtDecoder(context, *elementAt, rows);
+  auto decodedElementAt = elementAtDecoder.get();
+
+  // Optimal use case: unwrapped vector and constant key. We can simply project
+  // the feature using the first value in the arg vector.
+  if (decodedMap.isIdentityMapping() && decodedElementAt->isConstantMapping()) {
```
pedroerp: can we have this case by directly checking the encoding type of the elementAt vector? This should be the most common case, and here you don't necessarily need to decode the vector.

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,9 +185,76 @@ VectorPtr applyMapTyped(
...
+    const DecodedVector& decodedMap,
+    const VectorPtr& elementAt,
+    exec::EvalCtx& context) {
+  // Decode input flat map vector.
+  auto flatMap = decodedMap.base()->as<FlatMapVector>();
+
+  // Optimal use case: unwrapped vector and constant key. We can simply project
+  // the feature using the first value in the arg vector.
+  if (decodedMap.isIdentityMapping() && elementAt->isConstantEncoding()) {
+    if (auto projection = flatMap->projectKey(elementAt, 0)) {
```
pedroerp, line 203: Accepting to unblock, but you need to fetch the right idx from the constant vector (it's not always zero).

velox/functions/lib/SubscriptUtil.cpp
```diff
@@ -186,9 +185,76 @@ VectorPtr applyMapTyped(
...
+    auto mutableIndices = indices->asMutable<vector_size_t>();
+    auto rawNulls = nulls->asMutable<uint64_t>();
+    for (int i = 0; i < decodedMap.size(); i++) {
+      mutableIndices[i] = decodedMap.indices()[i];
+      if (decodedMap.isNullAt(i)) {
+        bits::setNull(rawNulls, i, true);
+      }
+    }
+
+    if (auto projection = flatMap->projectKey(elementAt, 0)) {
```
pedroerp, line 224: same here

velox/functions/lib/SubscriptUtil.cpp
peterenescu: Done!

PR #15040, opened 2025-10-03
build(ci): Free more space for docker builds

czentgr:


czentgr:
@assignUser Please review. We saw problems with some runners running out of disk space. This frees up more.

PR #15041, opened 2025-10-04
feat: Add support for casting VARCHAR to TIME

han-yan01:
Summary: Adds support for casting VARCHAR to TIME. Casting accounts for the local time zone.

Differential Revision: D83886564

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -309,18 +309,100 @@ VectorPtr CastExpr::castFromTime(
   }
 }
 
+VectorPtr CastExpr::castToTime(
+    const SelectivityVector& rows,
+    const BaseVector& input,
+    exec::EvalCtx& context,
+    const TypePtr& fromType) {
+  VectorPtr castResult;
+  context.ensureWritable(rows, TIME(), castResult);
+  (*castResult).clearNulls(rows);
+
+  switch (fromType->kind()) {
+    case TypeKind::VARCHAR: {
+      // Get session timezone
+      const auto* timeZone =
+          getTimeZoneFromConfig(context.execCtx()->queryCtx()->queryConfig());
+      // Get session start time
+      const auto startTimeMs =
+          context.execCtx()->queryCtx()->queryConfig().sessionStartTimeMs();
+      auto systemDay = std::chrono::milliseconds{startTimeMs} / kMillisInDay;
+
+      auto* inputVector = input.as<SimpleVector<StringView>>();
+      auto* resultFlatVector = castResult->as<FlatVector<int64_t>>();
+
+      applyToSelectedNoThrowLocal(context, rows, castResult, [&](int row) {
+        try {
+          const auto inputString = inputVector->valueAt(row);
+
+          // Parse the time string using the TimeType method
+          // This converts HH:MM:SS[.sss] format to milliseconds since midnight
+          int64_t parsedTimeMillis = TIME()->valueToTime(inputString);
+
+          // Validate time range (0 to 86399999 ms in a day)
+          if (parsedTimeMillis < 0 || parsedTimeMillis >= kMillisInDay) {
+            VELOX_USER_FAIL(
+                makeErrorMessage(input, row, TIME()) +
+                    " Time value {} is out of range",
+                parsedTimeMillis);
+          }
+
+          // Apply timezone conversion (reverse of TIME to VARCHAR)
+          // Interpret the parsed time as local time, convert to UTC
+          int64_t adjustedTime{0};
+          if (timeZone) {
+            // Create a local timestamp for the current day with the parsed time
+            auto localSystemTime =
+                systemDay.count() * kMillisInDay + parsedTimeMillis;
+
+            // Convert from local time to UTC
+            auto utcTime =
+                timeZone->to_sys(std::chrono::milliseconds{localSystemTime});
+            adjustedTime = (utcTime % kMillisInDay).count();
+          } else {
+            adjustedTime = parsedTimeMillis;
+          }
+
+          // Handle day wrapping
+          if (adjustedTime < 0) {
+            adjustedTime += kMillisInDay;
+          } else if (adjustedTime >= kMillisInDay) {
+            adjustedTime -= kMillisInDay;
+          }
+
+          resultFlatVector->set(row, adjustedTime);
+        } catch (const VeloxException& ue) {
+          if (!ue.isUserError()) {
+            throw;
+          }
+          VELOX_USER_FAIL(
+              makeErrorMessage(input, row, TIME()) + " " + ue.message());
+        } catch (const std::exception& e) {
+          VELOX_USER_FAIL(
+              makeErrorMessage(input, row, TIME()) + " " + e.what());
+        }
+      });
+
+      return castResult;
+    }
+    default:
+      VELOX_UNSUPPORTED(
+          "Cast from {} to TIME is not supported", fromType->toString());
+  }
+}
+
 namespace {
 void propagateErrorsOrSetNulls(
     bool setNullInResultAtError,
-    EvalCtx& context,
+    exec::EvalCtx& context,
     const SelectivityVector& nestedRows,
     const BufferPtr& elementToTopLevelRows,
     VectorPtr& result,
-    EvalErrorsPtr& oldErrors) {
+    exec::EvalErrorsPtr& oldErrors) {
   if (context.errors()) {
     if (setNullInResultAtError) {
-      // Errors in context.errors() should be translated to nulls in the top
-      // level rows.
+      // Errors in context.errors() should be translated to nulls in
+      // the top level rows.
       context.convertElementErrorsToTopLevelNulls(
           nestedRows, elementToTopLevelRows, result);
     } else {
@@ -347,8 +429,8 @@ VectorPtr CastExpr::applyMap(
     exec::EvalCtx& context,
     const MapType& fromType,
     const MapType& toType) {
-  // Cast input keys/values vector to output keys/values vector using their
-  // element selectivity vector
+  // Cast input keys/values vector to output keys/values vector using
+  // their element selectivity vector
 
   // Initialize nested rows
   auto mapKeys = input->mapKeys();
@@ -400,8 +482,8 @@ VectorPtr CastExpr::applyMap(
     }
   }
 
-  // Returned map vector should be addressable for every element, even those
-  // that are not selected.
+  // Returned map vector should be addressable for every element, even
+  // those that are not selected.
   BufferPtr sizes = input->sizes();
   if (newMapKeys->isConstantEncoding() && newMapValues->isConstantEncoding()) {
     // We extends size since that is cheap.
@@ -449,8 +531,8 @@ VectorPtr CastExpr::applyArray(
     exec::EvalCtx& context,
     const ArrayType& fromType,
     const ArrayType& toType) {
-  // Cast input array elements to output array elements based on their types
-  // using their linear selectivity vector
+  // Cast input array elements to output array elements based on their
+  // types using their linear selectivity vector
   auto arrayElements = input->elements();
 
   auto nestedRows =
@@ -473,8 +555,8 @@ VectorPtr CastExpr::applyArray(
         newElements);
   }
 
-  // Returned array vector should be addressable for every element, even those
-  // that are not selected.
+  // Returned array vector should be addressable for every element,
+  // even those that are not selected.
   BufferPtr sizes = input->sizes();
   if (newElements->isConstantEncoding()) {
     // If the newElements we extends its size since that is cheap.
@@ -519,8 +601,8 @@ VectorPtr CastExpr::applyRow(
   int numInputChildren = input->children().size();
   int numOutputChildren = toRowType.size();
 
-  // Extract the flag indicating matching of children must be done by name or
-  // position
+  // Extract the flag indicating matching of children must be done by
+  // name or position
   auto matchByName =
       context.execCtx()->queryCtx()->queryConfig().isMatchStructByName();
 
@@ -530,14 +612,16 @@ VectorPtr CastExpr::applyRow(
 
   EvalErrorsPtr oldErrors;
   if (setNullInResultAtError()) {
-    // We need to isolate errors that happen during the cast from previous
-    // errors since those translate to nulls, unlike exisiting errors.
+    // We need to isolate errors that happen during the cast from
+    // previous errors since those translate to nulls, unlike
+    // exisiting errors.
     context.swapErrors(oldErrors);
   }
 
   for (auto toChildrenIndex = 0; toChildrenIndex < numOutputChildren;
        toChildrenIndex++) {
-    // For each child, find the corresponding column index in the output
+    // For each child, find the corresponding column index in the
+    // output
     const auto& toFieldName = toRowType.nameOf(toChildrenIndex);
     bool matchNotFound = false;
 
@@ -711,8 +795,8 @@ void CastExpr::applyPeeled(
     };
 
     if (setNullInResultAtError()) {
-      // This can be optimized by passing setNullInResultAtError() to castTo and
-      // castFrom operations.
+      // This can be optimized by passing setNullInResultAtError() to
+      // castTo and castFrom operations.
 
       EvalErrorsPtr oldErrors;
       context.swapErrors(oldErrors);
@@ -748,6 +832,8 @@ void CastExpr::applyPeeled(
         toType->toString());
   } else if (fromType->isTime()) {
     result = castFromTime(rows, input, context, toType);
+  } else if (toType->isTime()) {
+    result = castToTime(rows, input, context, fromType);
   } else if (toType->isShortDecimal()) {
     result = applyDecimal<int64_t>(rows, input, context, fromType, toType);
   } else if (toType->isLongDecimal()) {
@@ -867,8 +953,8 @@ VectorPtr CastExpr::applyTimestampToVarcharCast(
     const auto stringView =
         Timestamp::tsToStringView(inputValue, options, rawBuffer);
     flatResult->setNoCopy(row, stringView);
-    // The result of both Presto and Spark contains more than 12 digits even
-    // when 'zeroPaddingYear' is disabled.
+    // The result of both Presto and Spark contains more than 12
+    // digits even when 'zeroPaddingYear' is disabled.
     VELOX_DCHECK(!stringView.isInline());
     rawBuffer += stringView.size();
   });
@@ -965,8 +1051,8 @@ void CastExpr::apply(
   context.moveOrCopyResult(localResult, *remainingRows, result);
   context.releaseVector(localResult);
 
-  // If there are nulls or rows that encountered errors in the input, add nulls
-  // to the result at the same rows.
+  // If there are nulls or rows that encountered errors in the input,
+  // add nulls to the result at the same rows.
   VELOX_CHECK_NOT_NULL(result);
   if (rawNulls || context.errors()) {
     EvalCtx::addNulls(
@@ -995,7 +1081,8 @@ void CastExpr::evalSpecialForm(
   } else {
     apply(rows, input, context, fromType, toType, result);
   }
-  // Return 'input' back to the vector pool in 'context' so it can be reused.
+  // Return 'input' back to the vector pool in 'context' so it can be
+  // reused.
   context.releaseVector(input);
 }
 
--- a/velox/expression/CastExpr.h
+++ b/velox/expression/CastExpr.h
@@ -210,6 +210,12 @@ class CastExpr : public SpecialForm {
       exec::EvalCtx& context,
       const TypePtr& toType);
 
+  VectorPtr castToTime(
+      const SelectivityVector& rows,
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const TypePtr& fromType);
+
   template <typename TInput, typename TOutput>
   void applyDecimalCastKernel(
       const SelectivityVector& rows,
--- a/velox/expression/tests/CastExprTest.cpp
+++ b/velox/expression/tests/CastExprTest.cpp
@@ -3269,5 +3269,250 @@ TEST_F(CastExprTest, timeToBigintCast) {
   }
 }
 
+TEST_F(CastExprTest, varcharToTimeCast) {
+  {
+    // Test casting VARCHAR to TIME with valid time strings
+    auto varcharVector = makeFlatVector<StringView>(
+        {"00:00:00.000",
+         "01:01:01.000",
+         "12:00:00.000",
+         "23:59:59.999",
+         "09:30:15.123",
+         "18:45:30.500"});
+
+    // Use makeCastExpr directly to avoid DuckDB parser issues
+    auto inputField =
+        std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0");
+    auto castExpr = makeCastExpr(inputField, TIME(), false);
+    auto result = evaluate(castExpr, makeRowVector({varcharVector}));
+
+    // TIME is milliseconds since midnight
+    auto expected = makeFlatVector<int64_t>(
+        {
+            0, // 00:00:00.000
+            3661000, // 01:01:01.000 = 3600*1000 + 60*1000 + 1*1000
+            43200000, // 12:00:00.000 = 12*3600*1000
+            86399999, // 23:59:59.999 = 23*3600*1000 + 59*60*1000 + 59*1000 +
+                      // 999
+            34215123, // 09:30:15.123 = 9*3600*1000 + 30*60*1000 + 15*1000 + 123
+            67530500 // 18:45:30.500 = 18*3600*1000 + 45*60*1000 + 30*1000 + 500
+        },
+        TIME());
+
+    assertEqualVectors(expected, result);
+  }
+
+  {
+    // Test casting VARCHAR to TIME with nulls
+    auto varcharVector = makeNullableFlatVector<StringView>(
+        {"00:00:00.000", std::nullopt, "12:00:00.000", std::nullopt});
+
+    auto inputField =
+        std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0");
+    auto castExpr = makeCastExpr(inputField, TIME(), false);
+    auto result = evaluate(castExpr, makeRowVector({varcharVector}));
+
+    auto expected = makeNullableFlatVector<int64_t>(
+        {0, std::nullopt, 43200000, std::nullopt}, TIME());
+
+    assertEqualVectors(expected, result);
+  }
+
+  {
+    // Test time formats without milliseconds
+    auto varcharVector =
+        makeFlatVector<StringView>({"00:00:00", "12:30:45", "23:59:59"});
+
+    auto inputField =
+        std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0");
+    auto castExpr = makeCastExpr(inputField, TIME(), false);
+    auto result = evaluate(castExpr, makeRowVector({varcharVector}));
+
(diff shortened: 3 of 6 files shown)


velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,250 @@ TEST_F(CastExprTest, timeToBigintCast) {
   }
 }
 
+TEST_F(CastExprTest, varcharToTimeCast) {
+  {
+    // Test casting VARCHAR to TIME with valid time strings
```
kgpai, line 3277: Can you explicitly set the session time zone to UTC. before running the tests.

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,250 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+            std::nullopt, // completely invalid
+            std::nullopt // missing seconds
+        },
+        TIME());
+
+    assertEqualVectors(expected, result);
+  }
+
+  {
+    // Test edge cases with leading/trailing spaces (should be handled by
```
kgpai: Is this true ? Leading/trailing spaces might be handled only when its part of the sql construct, otherwise it should throw. Can you validate this in Presto.

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,250 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+    });
+
+    auto inputField =
+        std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0");
+    auto tryCastExpr = makeCastExpr(inputField, TIME(), true);
+    auto result = evaluate(tryCastExpr, makeRowVector({varcharVector}));
+
+    auto expected = makeNullableFlatVector<int64_t>(
+        {
+            45045123, // 12:30:45.123 with spaces handled
```
kgpai: I do believe this is wrong, Presto will throw on trailing/leading spaces is my belief.

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,250 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+        },
+        TIME());
+
+    assertEqualVectors(expected, result);
+  }
+
+  {
+    // Test microsecond precision (should truncate to milliseconds)
+    auto varcharVector = makeFlatVector<StringView>({
+        "12:30:45.123456", // microseconds should be truncated
```
kgpai: Can you validate this behavior in Presto, I will be surprised if this is the case.

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,97 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+      result = result * 10 + (data[pos] - '0');
+      pos++;
+      digitCount++;
+    }
+
+    return digitCount >= minDigits;
+  };
+
+  // Skip leading whitespace
+  while (pos < timeStr.size() && data[pos] == ' ') {
```
kgpai: Dont think Presto does this - but please validate.

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,97 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  VELOX_USER_CHECK(
+      parseNumber(second, 1, 2), "Failed to parse second in time string");
+
+  // Parse optional fractional seconds
+  if (pos < timeStr.size() && data[pos] == '.') {
+    pos++;
+    int32_t fractionalPart = 0;
+    int32_t fractionalDigits = 0;
+
+    // Parse up to 6 digits for microseconds, but only keep first 3 for
```
kgpai: Again I dont think we support nano second precision .

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,97 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  };
+
+  // Skip leading whitespace
+  while (pos < timeStr.size() && data[pos] == ' ') {
+    pos++;
+  }
+
+  // Parse hour (1-2 digits)
+  VELOX_USER_CHECK(
+      parseNumber(hour, 1, 2), "Failed to parse hour in time string");
```
kgpai: If the second digit is unicode this will still return success - which isnt right, time to varchar will always be in hh:mm:SSS format .

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,100 @@ VectorPtr CastExpr::castFromTime(
...
+    const SelectivityVector& rows,
+    const BaseVector& input,
+    exec::EvalCtx& context,
+    const TypePtr& fromType) {
+  VectorPtr castResult;
+  context.ensureWritable(rows, TIME(), castResult);
+  (*castResult).clearNulls(rows);
+
+  switch (fromType->kind()) {
+    case TypeKind::VARCHAR: {
```
kgpai, line 319: nit move this to a separate function so we dont have a very large castToTime function.

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,100 @@ VectorPtr CastExpr::castFromTime(
...
+      applyToSelectedNoThrowLocal(context, rows, castResult, [&](int row) {
+        try {
+          const auto inputString = inputVector->valueAt(row);
+
+          // Parse the time string using the TimeType method
+          // This converts HH:MM:SS[.sss] format to milliseconds since midnight
+          int64_t parsedTimeMillis = TIME()->valueToTime(inputString);
+
+          // Validate time range (0 to 86399999 ms in a day)
+          if (parsedTimeMillis < 0 || parsedTimeMillis >= kMillisInDay) {
```
kgpai: This validation should be in the valueToTime and not here.

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,100 @@ VectorPtr CastExpr::castFromTime(
...
+            VELOX_USER_FAIL(
+                makeErrorMessage(input, row, TIME()) +
+                    " Time value {} is out of range",
+                parsedTimeMillis);
+          }
+
+          // Apply timezone conversion (reverse of TIME to VARCHAR)
+          // Interpret the parsed time as local time, convert to UTC
+          int64_t adjustedTime{0};
```
kgpai: I think everything related to timezone conversion should go in TIME's valueToTime function since it knows best about the semantics - The cast should only deal with casting logic.

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,100 @@ VectorPtr CastExpr::castFromTime(
...
+      VELOX_UNSUPPORTED(
+          "Cast from {} to TIME is not supported", fromType->toString());
+  }
+}
+
 namespace {
 void propagateErrorsOrSetNulls(
     bool setNullInResultAtError,
-    EvalCtx& context,
+    exec::EvalCtx& context,
```
kgpai: nit: why add exec:: if it works without it ..

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -68,7 +68,7 @@ intermediateTypeTransforms() {
           {JSON(), std::make_shared<JsonTransform>()},
           {TIME(),
            std::make_shared<IntermediateTypeTransformUsingCast>(
-               TIME(), BIGINT())},
```
kgpai, line 71: This means going from TIME() to BIGINT() right, why not add another one ?

velox/expression/tests/CastExprTest.cpp
han-yan01: Thanks for pointing that out: leading/trailing spaces will just fail the query with "cannot cast to TIME" Error. ref: 1355912979709395
Fixing it now.

velox/expression/tests/CastExprTest.cpp
han-yan01: Thanks for pointing that out: verified that Presto will throw for microseconds, only take exact milliseconds 
ref: 1355912979709395

velox/exec/fuzzer/PrestoQueryRunnerIntermediateTypeTransforms.cpp
```diff
@@ -66,9 +66,9 @@ intermediateTypeTransforms() {
            std::make_shared<IntermediateTypeTransformUsingCast>(
                SFMSKETCH(), VARBINARY())},
           {JSON(), std::make_shared<JsonTransform>()},
-          {TIME(),
-           std::make_shared<IntermediateTypeTransformUsingCast>(
-               TIME(), BIGINT())},
+          // {TIME(),
+          //  std::make_shared<IntermediateTypeTransformUsingCast>(
+          //      TIME(), BIGINT())},  // Disabled
```
kgpai: You do not need to disable this right, just change this to TIME(), VARCHAR() .

velox/exec/fuzzer/PrestoQueryRunner.cpp
```diff
@@ -338,10 +338,14 @@ bool PrestoQueryRunner::isSupported(const exec::FunctionSignature& signature) {
...
   //   - Expected differences between Presto Java and Velox C++ implementations
+  //
+  // time:
+  //   - Not supported by Fuzzer, will enable after intermediate type support
+  //   - Cannot be used as input/output types in Presto SQL
   return !(
       usesTypeName(signature, "interval year to month") ||
       usesTypeName(signature, "hugeint") ||
-      usesTypeName(signature, "geometry") ||
+      usesTypeName(signature, "geometry") || usesTypeName(signature, "time") ||
```
kgpai, line 344: I dont think is required - we do want to test time in Presto SOT using varchar as intermediate type.

velox/exec/tests/PrestoQueryRunnerTimeTransformTest.cpp
```diff
@@ -42,7 +42,7 @@ class PrestoQueryRunnerTimeTransformTest
 
 // Test that TIME is recognized as an intermediate type that needs
 // transformation
-TEST_F(PrestoQueryRunnerTimeTransformTest, isIntermediateOnlyType) {
+TEST_F(PrestoQueryRunnerTimeTransformTest, DISABLED_isIntermediateOnlyType) {
```
kgpai: If we are not having a special IntermediateTypeTransform for TIME we can remove these tests instead of just commenting things out.

velox/expression/CastExpr.h
```diff
@@ -210,6 +210,17 @@ class CastExpr : public SpecialForm {
       exec::EvalCtx& context,
       const TypePtr& toType);
 
+  VectorPtr castToTime(
+      const SelectivityVector& rows,
+      const BaseVector& input,
+      exec::EvalCtx& context,
+      const TypePtr& fromType);
+
+  VectorPtr castVarcharToTime(
```
kgpai: This need not be a public function, we should just have this as an anonymous inner function that is called by castToTime.

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,73 @@ VectorPtr CastExpr::castFromTime(
   }
 }
 
+VectorPtr CastExpr::castVarcharToTime(
```
kgpai: nit: castVarcharToTime need not be public member of CastExpr, rather lets have it inside an anonymous namespace that is called by castToTime.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -272,6 +272,7 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "cast(array(real)) -> array(varchar)",
     "cast(map(varchar,double)) -> map(varchar,varchar)",
     "cast(map(varchar,real)) -> map(varchar,varchar)",
+    "cast(varchar) -> time", // Disable VARCHAR to TIME casting
```
kgpai: Is there a reason why we want to disable this ?

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,139 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
   return StringView{startPos, kTimeToVarcharRowSize};
 }
 
+int64_t TimeType::valueToTime(const StringView& timeStr) const {
+  size_t pos = 0;
+  int32_t hour = 0, minute = 0, second = 0, millis = 0;
+  const char* data = timeStr.data();
+
+  // Helper function to check if a character is a digit
+  auto isDigit = [](char c) { return c >= '0' && c <= '9'; };
```
kgpai: nit: You can use std::isDigit .

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,139 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+
+  // Helper function to parse a number from a string at a given position
+  auto parseNumber =
+      [&](int32_t& result, int32_t minDigits, int32_t maxDigits) -> bool {
+    result = 0;
+    int32_t digitCount = 0;
+
+    while (pos < timeStr.size() && digitCount < maxDigits) {
```
kgpai: nit: cant you use str::stoi ?

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +309,73 @@ VectorPtr CastExpr::castFromTime(
...
+  auto* inputVector = input.as<SimpleVector<StringView>>();
+  auto* resultFlatVector = castResult->as<FlatVector<int64_t>>();
+
+  applyToSelectedNoThrowLocal(context, rows, castResult, [&](int row) {
+    try {
+      const auto inputString = inputVector->valueAt(row);
+
+      // Parse the time string using the TimeType method with timezone support
+      // This handles parsing, validation, and timezone conversion
+      int64_t result = TIME()->valueToTime(inputString, timeZone, systemDay);
```
kgpai: Isnt it better to pass in sessionStartTime and timeZone to valueToTime rather than systemDay ?

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,139 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+    const StringView& timeStr,
+    const tz::TimeZone* timeZone,
+    std::chrono::milliseconds systemDay) const {
+  // First parse the time string normally to get local time
+  int64_t parsedTimeMillis = valueToTime(timeStr);
+
+  // Apply timezone conversion if provided
+  if (timeZone) {
+    // Create a local timestamp for the current day with the parsed time
+    auto localSystemTime = systemDay.count() * kMillisInDay + parsedTimeMillis;
```
kgpai: You are dividing by kMillisInDay to get systemDay then multiplying again, you can just pass in sessionTime and then do something like this : 
`std::chrono::duration_cast<std::chrono::days>(sessionStartTime)` .

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,299 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+  }
+
+  {
+    // Test timezone-aware VARCHAR to TIME casting with America/Los_Angeles
+    // timezone When casting VARCHAR to TIME, the string is interpreted as local
+    // time and converted to UTC (the reverse of TIME to VARCHAR)
+    setSessionStartTimeAndTimeZone(
+        1756710000000, "America/Los_Angeles"); // 2025-09-01T00:00:00.000 (DST)
+
+    auto varcharVector = makeFlatVector<StringView>({
```
kgpai, line 3475: How does this handle Day light savings transition - 
Say you have time as '02:01:00' which is 2:01 am on Nov 2 PST . This is during day light savings transition time , and there is legally no 2:01 am (as we jump forward to 3:00 am) ,will converting this to UTC and back give us 2:01 am ?

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
han-yan01: trail and error attempt to make sure TIME is not included in this PR for testing

velox/type/Type.cpp
han-yan01: imo str::stoi is subpar here because granular control over each TIME component. Using it will loose the digit count constraint and actually complicates things. stoi also creates/copy substrings which is less efficient. wdyt?

velox/type/Type.cpp
han-yan01: Splendid optimization, thanks for pointing that out!

velox/expression/tests/CastExprTest.cpp
han-yan01: Testing shows it will crash:
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Cannot cast VARCHAR '02:01:00.000' to TIME.  2024-03-10 02:01:00.000 is in a gap between
2024-03-10 02:00:00 PST and
2024-03-10 03:00:00 PDT which are both equivalent to
2024-03-10 10:00:00 UTC

Would you like me to handle it more gracefully (ignore entry) or does this suffice?

han-yan01: wrote a test and validated that velox returns null for gap time
That's the graceful handling we are looking for right?

han-yan01: The gap time `'02:01:00'` will return NULL when converted during the DST spring-forward transition

velox/expression/CastExpr.cpp
han-yan01: I think with the existing time to varchar using system day I also need to use systemD to ensure TIME → VARCHAR → TIME round-trip conversions

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,610 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+    // Times outside the DST gap should round-trip correctly
+    EXPECT_EQ(result->valueAt(0), "01:30:00.000");
+    EXPECT_EQ(result->valueAt(1), "03:30:00.000");
+  }
+
+  {
+    // Test DST gap times with try_cast - gap times should return null
+    auto varcharVector = makeFlatVector<StringView>({
+        "01:30:00.000", // Before gap - should work
+        "02:01:00.000", // In the gap - should return null
```
kgpai, line 3626: It would be good if we have a test that explicitly ensures that we fail/throw for times in the gap. (try_cast null is fine, but it doesnt mean that we threw).

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,610 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+        "02:01:00.000", // Gap time - should throw exception
+    });
+
+    auto inputField =
+        std::make_shared<core::FieldAccessTypedExpr>(VARCHAR(), "c0");
+    auto castExpr =
+        makeCastExpr(inputField, TIME(), false); // regular cast, not try_cast
+
+    // Regular cast should throw exception for gap times
+    VELOX_ASSERT_THROW(
```
kgpai, line 3806: Oh nvm, I see that you have added that test here..

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+      components.minute);
+  VELOX_USER_CHECK(
+      !(components.second < 0 || components.second > 59),
+      "Invalid second value: {}",
+      components.second);
+}
+
+// Convert time components to milliseconds since midnight
+int64_t timeComponentsToMillis(const TimeComponents& components) {
+  int64_t result = static_cast<int64_t>(components.hour) * 3600000 +
```
kgpai: Use kMillisInHour  etc instead of magic numbers..

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+    fractionalDigits++;
+  }
+
+  return fractionalPart;
+}
+
+// Validate parsed time components
+void validateTimeComponents(const TimeComponents& components) {
+  VELOX_USER_CHECK(
+      !(components.hour < 0 || components.hour > 23),
```
kgpai: nit use (kHoursInDay -1) etc..

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+// Convert time components to milliseconds since midnight
+int64_t timeComponentsToMillis(const TimeComponents& components) {
+  int64_t result = static_cast<int64_t>(components.hour) * 3600000 +
+      static_cast<int64_t>(components.minute) * 60000 +
+      static_cast<int64_t>(components.second) * 1000 +
+      static_cast<int64_t>(components.millis);
+
+  // Validate time range (0 to 86399999 ms in a day)
+  VELOX_USER_CHECK(
+      !(result < 0 || result >= 86400000),
```
kgpai: nit: use kMillisInDay

kgpai: I see in some places we use kMillisInDay and other constants and some places we dont, lets try and be consistent.

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+
+struct TimeComponents {
+  int32_t hour = 0;
+  int32_t minute = 0;
+  int32_t second = 0;
+  int32_t millis = 0;
+};
+
+// Parse a number with min/max digit constraints from the given position
+bool parseNumber(
```
kgpai: nit: Is there a reason to use parseNumber etc instead of just using std::from_chars ?

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+
+// Skip a single character if it matches the expected character
+void skipExpectedChar(const StringView& str, size_t& pos, char expected) {
+  if (pos < str.size() && str.data()[pos] == expected) {
+    pos++;
+  }
+}
+
+// Parse fractional seconds (milliseconds)
+int32_t parseFractionalSeconds(const StringView& str, size_t& pos) {
```
kgpai, line 1568: I do believe all of this can be replaced with std::from_chars

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+
+int64_t TimeType::valueToTime(
+    const StringView& timeStr,
+    const tz::TimeZone* timeZone,
+    int64_t sessionStartTimeMs) const {
+  // First parse the time string normally to get local time
+  int64_t parsedTimeMillis = valueToTime(timeStr);
+
+  // Apply timezone conversion if provided
+  if (timeZone) {
```
kgpai: nit: make not having the timezone the unlikely case, 
if (FOLLY_UNLIKELY(!timezone)) {
return parsedTimeMillis;
} 
...

velox/type/Type.cpp
han-yan01: using >= kHoursinDay etc for better readability

velox/type/Type.cpp
```diff
@@ -1530,6 +1532,227 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  int32_t digitCount = 0;
+  while (digitEnd < end && std::isdigit(*digitEnd) && digitCount < maxDigits) {
+    digitEnd++;
+    digitCount++;
+  }
+
+  // Check minimum digits constraint
+  if (digitCount < minDigits) {
+    return false;
+  }
```
kgpai: I dont think any of this is required, all you would need to do is something like this :
```
bool parseNumber(const StringView& str,
    size_t& start_pos,
    int32_t& result,
   size_t& end_pos) {
   auto parseResult = std::from_chars(str.data()+ start_pos, str.data() + end_pos, result);
  if (parseResult.ec != std::errc{}) {
    return false;
  }
  return true;
  }
  ```

File: velox/type/Type.cpp [hunk 33]
```diff
@@ -1530,6 +1532,227 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  const char* start = str.data() + pos;
+  const char* end = str.data() + str.size();
+
+  // Find the end of digits, respecting max 6 digits constraint
+  const char* digitEnd = start;
+  int32_t fractionalDigits = 0;
+  while (digitEnd < end && std::isdigit(*digitEnd) && fractionalDigits < 6) {
+    digitEnd++;
+    fractionalDigits++;
+  }
```
kgpai: There is no need for this since if from_chars is not able to convert to a digit it will throw ..Thats why the ask to use std::from_chars.

velox/type/Type.cpp
```diff
@@ -1530,6 +1532,227 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+
+  // Only support exactly milliseconds, not microseconds
+  VELOX_USER_CHECK(
+      fractionalDigits <= 3,
+      "Value cannot be cast to time. Microseconds precision not supported");
```
kgpai: You can just do 'VELOX_USER_CHECK_LE(str.size()- pos, 3, "Microsecond precision not supported)'.

velox/type/Type.cpp
```diff
@@ -1530,6 +1532,227 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+    if (parseResult.ec != std::errc{}) {
+      return 0;
+    }
+  }
+
+  // Convert to milliseconds (pad with zeros if needed)
+  while (fractionalDigits < 3) {
+    fractionalPart *= 10;
+    fractionalDigits++;
+  }
```
kgpai: The format will always have 3 digits , so you do not need to do this; See: https://github.com/facebookincubator/velox/blob/394eeb730a4faff858a26313b19f0360f179d5a3/velox/type/Type.cpp#L1249

velox/type/Type.cpp
```diff
@@ -1530,6 +1532,227 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+// Convert time components to milliseconds since midnight
+int64_t timeComponentsToMillis(const TimeComponents& components) {
+  int64_t result = static_cast<int64_t>(components.hour) * kMillisInHour +
+      static_cast<int64_t>(components.minute) * kMillisInMinute +
+      static_cast<int64_t>(components.second) * kMillisInSecond +
+      static_cast<int64_t>(components.millis);
+
+  // Validate time range (0 to 86399999 ms in a day)
+  VELOX_USER_CHECK(
+      !(result < 0 || result >= 86400000),
```
kgpai: Shouldnt we use kMillisInDay here ?

velox/expression/CastExpr.cpp
```diff
@@ -309,18 +310,70 @@ VectorPtr CastExpr::castFromTime(
...
+          } catch (const std::exception& e) {
+            VELOX_USER_FAIL(
+                makeErrorMessage(input, row, TIME()) + " " + e.what());
+          }
+        });
+
+        return castResult;
+      };
+
+      return castVarcharToTime(rows, input, context);
```
kgpai: nit: Why the use for a lambda if its only going to be called once ?

velox/type/Type.cpp
```diff
@@ -1530,6 +1532,204 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  }
+
+  // Calculate actual digit count
+  int32_t fractionalDigits = parseResult.ptr - start;
+
+  // Convert to milliseconds (pad with zeros if needed)
+  while (fractionalDigits < 3) {
+    fractionalPart *= 10;
+    fractionalDigits++;
+  }
```
kgpai: Dont think this is required since you are guaranteed 3 fractional digits. Does java produce a case where you can have less than 3 fractionalDigits ?

PR #15042, opened 2025-10-04
fix: Remove the backward macro to fix presto failure

xiaoxmeng:
Differential Revision: D83901797


duanmeng (approved):
@xiaoxmeng LGTM % thanks for your fix.

juwentus1234 (approved):

Issue #15043, opened 2025-10-04
[Velox Prestissimo] NULLs corrupted in VARCHAR and BIGINT columns at scale


hitarth:
### Bug description

While testing **Parquet writer correctness** in Prestissimo and comparing results against Presto Java, we observed unexpected handling of `NULL` values in `VARCHAR` columns: some `NULL`s are being written as empty strings in the destination table.

For smaller tables (up to ~100M rows), the outputs match. However, once the row count grows beyond ~150M, we consistently see that a portion of `NULL` varchars are converted into empty strings.

Here’s a simplified table schema used for reproducing the issue:

```
          Column           |  Type   | Extra | Comment
---------------------------+---------+-------+---------
 device_os                 | varchar |       |
 begin_effective_timestamp | bigint  |       |
 randomization_unit_id     | varchar |       |
 bucket_id                 | integer |       |
 cohort_index              | integer |       |
```

We confirmed this with a simple repro: repeatedly creating chained copies of a table using

```sql
CREATE TABLE tmp.source_table AS 
SELECT * FROM tmp.<previous_table>;
```

When querying these chained copies, the number of `NULL` values decreases with each step, while the number of empty strings increases (screenshot attached).

To clarify, we have verified there are no issues in the `SELECT` portion of the queries. The problem seems isolated to the writer.

Additionally, this incorrect null handling isn’t limited to `VARCHAR`. In some cases, `BIGINT` columns that should contain `NULL` values instead show random numbers.

<img width="873" height="1043" alt="Image" src="https://github.com/user-attachments/assets/e6394768-cc80-4a26-94a9-6e159fa76d3d" />

### System information

Velox System Info v0.0.2
Commit: 0de7acae8e2b50e3fd7f5d61e3ae7159b2bc0bdd
CMake Version: 3.28.3
System: Linux-6.6.93+
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.4.0
CMake Prefix Path: /usr/local;/usr;/;/usr/local/lib/python3.9/dist-packages/cmake/data;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

kgpai:
cc: @majetideepak

acvictor:
[...]

PR #15044, opened 2025-10-04
fix: Null handling during velox to arrow conversion

hitarth:
Summary:
A bug in Velox’s Parquet writer caused null VARCHAR values to be written as empty strings, breaking data integrity. The issue was traced to inconsistent null count handling between exportValidityBitmap() and exportStrings(), which produced Arrow arrays with mismatched metadata. We resolved this by updating exportValidityBitmap() always to calculate null counts for the current row selection rather than relying on cached values. This ensures consistency across the export pipeline. The fix eliminates silent corruption with minimal performance impact.

Fixes: https://github.com/facebookincubator/velox/issues/15043

### DIFF
--- a/velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
+++ b/velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
@@ -31,6 +31,7 @@
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/exec/tests/utils/QueryAssertions.h"
 #include "velox/exec/tests/utils/TempDirectoryPath.h"
+#include "velox/vector/DecodedVector.h"
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -836,6 +837,88 @@ TEST_F(ParquetWriterTest, updateWriterOptionsFromHiveConfig) {
...
+  for (int64_t i = 0; i < kRows; ++i) {
+    if (testData[i].has_value()) {
+      stringViewTestData.push_back(
+          StringView(testData[i]->data(), testData[i]->size()));
+    } else {
+      stringViewTestData.push_back(std::nullopt);
+    }
+  }
+
+  auto stringVector = makeNullableFlatVector<StringView>(stringViewTestData);
```
pedroerp, line 888: as far as I remember, you don't need to create the StringView objects yourself. You can just pass the std::strings to makeNullableFlatVector(), and internally it will do the StringView conversions for you as it creates the vector underneat.

pedroerp:
> to calculate null counts for the current row selection rather than relying on cached values

@hitarth does this mean you are getting vectors with a cached (but wrong) null count? Where were they coming from?

velox/vector/arrow/Bridge.cpp
```diff
@@ -666,12 +666,25 @@ void exportValidityBitmap(
...
+    // For constant vectors, compute null count based on the constant value and
+    // selection size
+    if (vec.getNullCount().value() == 0) {
+      // Constant non-null vector: no nulls in any selection
+      out.null_count = 0;
+    } else {
+      // Constant null vector: all selected elements are null
+      out.null_count = rows.count();
+    }
   } else {
```
pedroerp, line 684: don't we want to use the cached values for non-constant vector anymore?

hitarth: Actually, you raise a good point that made me re-examine the logic. Let me clarify:
When !rows.changed() is true, we're exporting the entire vector, so the cached null count should be correct. The original code's logic there was sound.
However, I'm seeing null corruption in production (affecting both VARCHAR and BIGINT as shown in #15043), and this manifests during Parquet writes at scale.
Can you help me understand: Are there scenarios where vec.getNullCount() could return a stale/incorrect cached value even when we're exporting the full vector? Or could there be cases where rows.changed() returns false but we're actually working with a modified view of the vector?
I may need to add more detailed logging to capture the exact state when corruption occurs. The safer approach for now is to always compute, but I want to understand if there's a specific edge case I'm missing.

majetideepak:
@hitarth Do the two tests you added fail without the fix in `Bridge.cpp`? I am asking based on your latest comment above,

hitarth:
> @hitarth Do the two tests you added fail without the fix in `Bridge.cpp`? I am asking based on your latest comment above,
@majetideepak thanks for taking a look. I was not able to reproduce this in the unit test, I have explained the repro in the linked [issue](https://github.com/facebookincubator/velox/issues/15043) , I can remove the unit tests, if you think they are not needed. I had added it for sanity check.

majetideepak:
@hitarth If the tests cover code paths that are not currently tested, we can keep them.
I am mainly interested in reproducing this issue. I saw the reproducer in the linked issue, but it is not complete since the initial table data is missing.
Are you able to share a reproducer from the start that I can try on my end?

hitarth:
It happens in our production tables with write queries . We did verifier and checked that after disabling the cached null count we don't see the issue, and this is deployed uber production.

majetideepak:
@hitarth, Recomputing nullcounts introduces CPU overhead in the order of the total row count. I prefer that we avoid it if possible. Since you have a reproducer handy, I would like to get to the bottom of this issue.
Can you help log if it is always a dictionary-encoded vector that was flattened before export that causes this?
https://github.com/facebookincubator/velox/blob/main/velox/vector/arrow/Bridge.cpp#L1204
That should give us some direction. Thanks!

velox/vector/arrow/Bridge.cpp
pedroerp: @hitarth none of these cases should happen, as far as I can tell. We should figure out who is creating the corrupted vectors (with the inaccurate null counts, and fix things there).

PR #15052, opened 2025-10-06
fix: Fix sxtream failed test by supporting task cancellation in test

xiaoxmeng:
Summary: xstream runtime doesn't cleanup the non-started velox task. A quick fix from test is to add task cancellation for this.

Differential Revision: D83944606

### DIFF
--- a/velox/exec/tests/utils/QueryAssertions.cpp
+++ b/velox/exec/tests/utils/QueryAssertions.cpp
@@ -1441,6 +1441,15 @@ void waitForAllTasksToBeDeleted(uint64_t maxWaitUs) {
       folly::join("\n", pendingTaskStats));
 }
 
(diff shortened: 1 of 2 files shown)


velox/exec/tests/utils/QueryAssertions.cpp
```diff
@@ -1441,6 +1441,13 @@ void waitForAllTasksToBeDeleted(uint64_t maxWaitUs) {
       folly::join("\n", pendingTaskStats));
 }
 
+void cancelAllTasks() {
+  std::vector<std::shared_ptr<Task>> pendingTasks = Task::getRunningTasks();
```
duanmeng, line 1445: const auto?

gggrace14 (approved):

PR #15059, opened 2025-10-06
feat: add ExchangeAggregation operator

kagamiori:
Summary:
This diff adds a new ExchangeAggregation operator to perform LocalPartitionNode 
followed by HashAggregationNode(Final) in a query plan. LocalPartitionNode followed 
by HashAggregationNode(Final) is originally implemented as 
LocalPartition/LocalExchange followed by HashAggregation operator. But 
LocalPartition/LocalExchange operators suffer from high lock contention when the 
number of drivers is large. So we introduce the ExchangeAggregation operator to 
make query performance more scalable.

Differential Revision: D83116141


velox/exec/ExchangeAggregation.h
```diff
@@ -0,0 +1,208 @@
...
+    return needToConsume_.load();
+  }
+
+  ~PartitionBucket() = default;
+
+  void testingValidate();
+
+ private:
+  size_t bucketSize_;
+  std::vector<std::atomic<Partitioned*>> slots_;
```
Yuhta, line 103: Double check each element is aligned at cache line size (64 bytes) to avoid false sharing.  You may need to wrap the `Partitioned*` in a struct annotated with `alignas(64)` to fix this.

PR #15062, opened 2025-10-06
Add native longest_common_prefix function

natashasehgal:
Summary: Add native longest_common_prefix function

Differential Revision: D83945193


velox/functions/prestosql/StringFunctions.h
```diff
@@ -672,4 +672,93 @@ struct XxHash64StringFunction {
...
+      size_t leftSize = left.size();
+      size_t rightSize = right.size();
+
+      size_t leftPos = 0, rightPos = 0;
+      size_t commonByteLength = 0;
+
+      while (leftPos < leftSize && rightPos < rightSize) {
+        int leftCharBytes = 0, rightCharBytes = 0;
+
+        auto leftCodePoint = utf8proc_codepoint(
```
MBkkt: I think this can be faster, we can find it same way as for ascii.
And then do -4...0 symbols to position us in the end of codepoint.

velox/functions/prestosql/StringFunctions.h
```diff
@@ -672,4 +672,93 @@ struct XxHash64StringFunction {
...
+      out_type<Varchar>& result,
+      const arg_type<Varchar>& left,
+      const arg_type<Varchar>& right) {
+    if constexpr (isAscii) {
+      size_t minLength = std::min(left.size(), right.size());
+      const char* leftData = left.data();
+      const char* rightData = right.data();
+
+      size_t commonPrefixLength = 0;
+      for (size_t i = 0; i < minLength; ++i) {
```
MBkkt: I'm not sure velox use or not abseil.
But it can be faster https://github.com/abseil/abseil-cpp/blob/master/absl/strings/match.cc#L74

natashasehgal: Velox doesn't use abseil for dependency reasons

velox/functions/prestosql/StringFunctions.h
natashasehgal: This optimization was failing the unit tests for correctness. Ended up not adding it, as the optimization didn't seem worth it for now

velox/functions/prestosql/StringFunctions.h
MBkkt: It's not actually true.
1. Velox already depends on abseil, just implicitly, because re2, gtest, protobuf, etc
2. I send this link to you to you check algorithm, it's quite simple, but more efficien

natashasehgal: Thanks. I got some internal feedback to not add abseil implementation if it can be avoided. Efficiency vs. dependency

I have created an issue to look into this optimization - https://github.com/facebookincubator/velox/issues/15467

velox/functions/prestosql/StringFunctions.h
MBkkt: @natashasehgal 

I don't think so, it will be nice if you explain what are you trying or at least provide example of test which doesn't pass.

The main point of suggested algorithm is following, longest common utf8 prefix will be prefix for longest common byte prefix.
Because it's true you can first find longest common byte prefix and then make it shorter for 0..3 bytes

velox/functions/prestosql/StringFunctions.h
MBkkt: Thanks for creating the issue.

But as I already wrote
1. Velox already depends on abseil
2. You don't need to use abseil if you don't want, you can just write same algorithm, it's quite trivial hack

PR #15063, opened 2025-10-06
feat: Add native from_base32 functions

natashasehgal:
Differential Revision: D84003180

### DIFF
--- a/velox/common/encode/Base32.cpp
+++ b/velox/common/encode/Base32.cpp
@@ -1,0 +1,192 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/common/encode/Base32.h"
+
+#include <folly/Expected.h>
+#include <cctype>
+#include <cstdint>
+
+#include "velox/common/base/CheckedArithmetic.h"
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::velox::encoding {
+
+// Reverse lookup table for decoding. 255 means invalid character.
+// Only uppercase letters (A-Z) and digits 2-7 are valid per RFC 4648.
+// Lowercase letters are NOT supported (matching Google Guava's
+// BaseEncoding.base32()).
+constexpr const Base32::ReverseIndex kBase32ReverseIndexTable = {
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 0-15
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 16-31
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 32-47
+    255, 255, 26,  27,  28,  29,  30,  31,  255,
+    255, 255, 255, 255, 255, 255, 255, // 48-63 ('2'-'7')
+    255, 0,   1,   2,   3,   4,   5,   6,   7,
+    8,   9,   10,  11,  12,  13,  14, // 64-79 ('A'-'O')
+    15,  16,  17,  18,  19,  20,  21,  22,  23,
+    24,  25,  255, 255, 255, 255, 255, // 80-95 ('P'-'Z')
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 96-111 (lowercase 'a'-'o' - INVALID)
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 112-127 (lowercase 'p'-'z' - INVALID)
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 128-143
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 144-159
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 160-175
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 176-191
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 192-207
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, // 208-223
+    255, 255, 255, 255, 255, 255, 255, 255, 255,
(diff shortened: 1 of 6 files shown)


velox/common/encode/Base32.h
```diff
@@ -0,0 +1,147 @@
...
+  /// Encodes the specified text.
+  static std::string encode(folly::StringPiece text);
+
+  /// Encodes the specified IOBuf data.
+  static std::string encode(const folly::IOBuf* inputBuffer);
+
+  /// Encodes the specified number of characters from the 'input' and writes the
+  /// result to the 'outputBuffer'. The output must have enough space as
+  /// returned by the calculateEncodedSize().
+  static void encode(const char* input, size_t inputSize, char* outputBuffer);
```
kgpai: Shouldnt it return the size of bytes written to outputBuffer ? (or perhaps a stringview ?)

velox/common/encode/Base32.h
```diff
@@ -0,0 +1,147 @@
...
+
+  // Decodes the specified data using the provided reverse lookup table.
+  static folly::Expected<size_t, Status> decodeImpl(
+      const char* input,
+      size_t inputSize,
+      char* outputBuffer,
+      size_t outputSize,
+      const ReverseIndex& reverseIndex);
+
+  VELOX_FRIEND_TEST(Base32Test, checksPadding);
```
kgpai: nit: Why not expose the padding functions etc, so that we dont need the friend tests ?

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::velox::encoding {
+
+// Character set for Base32 encoding (RFC 4648)
+constexpr const Base32::Charset kBase32Charset = {
+    'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K',
+    'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
+    'W', 'X', 'Y', 'Z', '2', '3', '4', '5', '6', '7'};
```
kgpai: I presume this is correct since I dont see '1' etc..

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+    255, 255, 255, 255, 255, 255, 255, 255,
+    255, 255, 255, 255, 255, 255, 255, 255 // 240-255
+};
+
+// static
+folly::Expected<uint8_t, Status> Base32::base32ReverseLookup(
+    char encodedChar,
+    const ReverseIndex& reverseIndex) {
+  auto index = reverseIndex[static_cast<uint8_t>(encodedChar)];
+  if (index >= 0x20) {
```
kgpai: nit: why not write 32 since the value seems to be encoded in decimal ?

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+}
+
+// static
+size_t Base32::calculateEncodedSize(size_t inputSize, bool withPadding) {
+  if (inputSize == 0) {
+    return 0;
+  }
+
+  // Base32 encodes 5 bytes into 8 characters
+  size_t encodedSize = ((inputSize + Base32::kBinaryBlockByteSize - 1) /
```
kgpai: nit: Is there risk of overflow underflow due to division and multiplication ? Should you use checkedMultiply / divide.

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+    encodedSize = (bitsNeeded + 4) / 5; // bits per base32 character is 5
+  }
+
+  return encodedSize;
+}
+
+// static
+folly::Expected<size_t, Status> Base32::calculateDecodedSize(
+    const char* input,
+    size_t& inputSize) {
```
kgpai: const size_t& (as you are passing a reference and it doesnt seem you intend to mutate it here) (or just pass by value).

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+  // We need to convert to bytes (8 bits each)
+  size_t totalBits = validCharCount * 5;
+  size_t decodedSize = totalBits / 8;
+
+  return decodedSize;
+}
+
+// Forward declare template specializations
+template <>
+void Base32::encodeImpl<folly::StringPiece>(
```
kgpai: Curious why we are adding support for StringPiece and IOBUF etc when we only will use StringView from the udf  below. Is there somewhere where these are called ?

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+  auto outputPointer = outputBuffer;
+  auto inputData = input.data();
+  size_t inputPos = 0;
+
+  // Process full blocks of 5 bytes (40 bits).
+  // Each block is encoded into 8 Base32 characters (8 * 5 = 40 bits).
+  // We pack the 5 input bytes into a 64-bit integer, then extract
+  // 8 groups of 5 bits each and map them to the Base32 charset.
+  for (; inputSize > 4; inputSize -= 5, inputPos += 5) {
```
kgpai: How does this work if inputSize is not a multiple of 5 ? Is there just padding ?

kgpai: NVM, I see it below..

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+  // Process full 5-byte blocks
+  while (cursor.length() >= 5) {
+    uint64_t inputBlock = static_cast<uint64_t>(cursor.read<uint8_t>()) << 32;
+    inputBlock |= static_cast<uint64_t>(cursor.read<uint8_t>()) << 24;
+    inputBlock |= static_cast<uint64_t>(cursor.read<uint8_t>()) << 16;
+    inputBlock |= static_cast<uint64_t>(cursor.read<uint8_t>()) << 8;
+    inputBlock |= static_cast<uint64_t>(cursor.read<uint8_t>());
+
+    *outputPointer++ = charset[(inputBlock >> 35) & 0x1f];
```
kgpai: This code seems common across all the encodes, maybe can be refactored to some other function call ?

velox/common/encode/Base32.cpp
```diff
@@ -0,0 +1,490 @@
...
+    char* outputBuffer,
+    size_t outputSize,
+    const ReverseIndex& reverseIndex) {
+  if (inputSize == 0) {
+    return 0;
+  }
+
+  size_t outputPos = 0;
+  uint64_t accumulator = 0;
+  int bitsAccumulated = 0;
```
kgpai: should this be size_t ?

velox/common/encode/Base32.h
natashasehgal: Updated to return size_t

velox/common/encode/Base32.h
natashasehgal: Made it public, was following Base64

velox/common/encode/Base32.cpp
natashasehgal: size_t is better updated

velox/common/encode/Base32.cpp
natashasehgal: Kept it similar to Base64. Do you mean refactor it and use it across Base32 (separate from Base64)?

velox/common/encode/Base32.cpp
natashasehgal: I'd added this for consistency with Base64 implementation which also provides IOBuf encoding support

velox/common/encode/Base32.cpp
natashasehgal: Added checkedMultiply

velox/common/encode/Base32.cpp
natashasehgal: RFC 4648 standard excludes 0, 1, 8, 9

velox/common/encode/Base32.cpp
kgpai: No the code between IOBuf and StringPiece is pretty similar save for minor differences and can be refactored so that most of it is shared. But more than that I still dont see why we need IOBuf support if there is no immediate use case.

PR #15068, opened 2025-10-06
feat: Add support for minute with TIME type

han-yan01:
Summary:
- implementation validates TIME range [0, 86399999] (milliseconds since midnight)
- extracts minutes using optimized formula: `(time % kMillisInHour) / kMillisInMinute`

Differential Revision: D84019915


velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -319,7 +319,7 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "day_of_year",
     "week_of_year",
     "doy",
-    "minute",
+    // "minute",
```
kgpai: Please remove this instead of commenting.

PR #15075, opened 2025-10-07
feat(udf): Add map_except Velox UDF

abhinavmuk04:
Summary:
## Summary

Implemented the `map_except` Velox UDF that removes specified keys from a map, complementing the existing `map_subset` function.

## Implementation Details

**Function Signature:** `MAP_EXCEPT(MAP(K,V), ARRAY[K]) -> MAP(K,V)`

**Core Logic:** Returns a map containing all entries from the input map EXCEPT those whose keys are in the provided array.

**Key Changes:**

1. **MapExcept.h** - Created three optimized function variants:
   - `MapExceptPrimitiveFunction` - Fast path for primitive types with constant key optimization
   - `MapExceptVarcharFunction` - Fast path for string types with constant key optimization
   - `MapExceptFunction` - Generic implementation for complex types

2. **MapFunctionsRegistration.cpp** - Registered function for all supported types:
   - Primitive types: bool, int8_t, int16_t, int32_t, int64_t, float, double, Timestamp, Date
   - String types: Varchar
   - Complex types: Generic (arrays, rows, etc.)

3. **MapExceptTest.cpp** - Comprehensive test coverage:
   - Bigint keys with constant and non-constant arrays
   - Varchar keys with inline and non-inline strings
   - Array keys (complex types)
   - Float/Double NaN handling
   - TimestampWithTimeZone support
   - Null value preservation
   - Empty array/map edge cases
   - Error handling for null comparisons

4. **BUCK** - Added MapExcept.h to build configuration

5. **map.rst** - Added documentation with usage examples

## Design Decisions

- Followed the exact pattern from `map_subset` (D82028529) but inverted the key matching logic
- Used `if (searchKeys_.contains(entry.first)) { continue; }` to exclude matching keys
- Removed early termination optimization since we're excluding keys, not finding them
- Preserved null values in output maps
- Maintained NaN equality semantics for floating point types

## Testing

All tests pass including:
- Edge cases (empty maps, empty key arrays)
- Type coverage (primitives, strings, complex types)
- Special values (NaN, null, TimestampWithTimeZone)
- Both constant and non-constant key arrays

[Session trajectory link](https://www.internalfb.com/intern/devai/devmate/inspector/?id=T240488251-4078e193-06d0-460c-a59c-b8c01c5154b8)

Differential Revision: D83998639

### DIFF
--- a/velox/functions/prestosql/MapExcept.h
+++ b/velox/functions/prestosql/MapExcept.h
@@ -1,0 +1,192 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/expression/ComplexViewTypes.h"
+#include "velox/functions/Udf.h"
+#include "velox/type/FloatingPointUtil.h"
+
+namespace facebook::velox::functions {
+
+/// Fast path for constant primitive type keys: map_except(m, array[1, 2, 3]).
+template <typename TExec, typename Key>
+struct MapExceptPrimitiveFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& /*config*/,
+      const arg_type<Map<Key, Generic<T1>>>* /*inputMap*/,
+      const arg_type<Array<Key>>* keys) {
+    if (keys != nullptr) {
+      constantSearchKeys_ = true;
+      searchKeys_.clear();
+      initializeSearchKeys(*keys);
+    }
+  }
+
+  void call(
+      out_type<Map<Key, Generic<T1>>>& out,
+      const arg_type<Map<Key, Generic<T1>>>& inputMap,
+      const arg_type<Array<Key>>& keys) {
+    if (!constantSearchKeys_) {
+      searchKeys_.clear();
+      initializeSearchKeys(keys);
+    }
+
+    for (const auto& entry : inputMap) {
+      if (searchKeys_.contains(entry.first)) {
+        continue;
+      }
+
+      if (!entry.second.has_value()) {
+        auto& keyWriter = out.add_null();
+        keyWriter = entry.first;
+      } else {
+        auto [keyWriter, valueWriter] = out.add_item();
+        keyWriter = entry.first;
+        valueWriter.copy_from(entry.second.value());
+      }
+    }
+  }
+
+ private:
+  void initializeSearchKeys(const arg_type<Array<Key>>& keys) {
+    for (const auto& key : keys.skipNulls()) {
+      searchKeys_.emplace(key);
+    }
+  }
+
+  bool constantSearchKeys_{false};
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/MapExcept.h
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  void call(
+      out_type<Map<Key, Generic<T1>>>& out,
+      const arg_type<Map<Key, Generic<T1>>>& inputMap,
+      const arg_type<Array<Key>>& keys) {
+    if (!constantSearchKeys_) {
+      searchKeys_.clear();
+      initializeSearchKeys(keys);
```
zacw7, line 47: in which case search keys would be initialized here instead of inside `initialize()`?

velox/functions/prestosql/MapExcept.h
```diff
@@ -0,0 +1,188 @@
...
+    if (keys != nullptr) {
+      constantSearchKeys_ = true;
+
+      searchKeyStrings_.reserve(keys->size());
+      for (const auto& key : keys->skipNulls()) {
+        if (key.isInline()) {
+          searchKeys_.emplace(key);
+        } else if (!searchKeys_.contains(key)) {
+          searchKeyStrings_.push_back(key.str());
+          searchKeys_.emplace(StringView(searchKeyStrings_.back()));
```
zacw7, line 99: do we need to clear these vectors?

velox/functions/prestosql/MapExcept.h
```diff
@@ -0,0 +1,188 @@
...
+
+  void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& /*config*/,
+      const arg_type<Map<Varchar, Generic<T1>>>* /*inputMap*/,
+      const arg_type<Array<Varchar>>* keys) {
+    if (keys != nullptr) {
+      constantSearchKeys_ = true;
+
+      searchKeyStrings_.reserve(keys->size());
```
zacw7, line 93: why don't we need the same for searchKeys_?

velox/functions/prestosql/MapExcept.h
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+ private:
+  bool constantSearchKeys_{false};
+  folly::F14FastSet<StringView> searchKeys_;
+  std::vector<std::string> searchKeyStrings_;
+};
+
+struct MapExceptFunctionEqualComparator {
```
zacw7, line 138: move this after `struct MapExceptFunction {...}`

zacw7 (commented):
LG % nits.

velox/functions/prestosql/MapExcept.h
abhinavmuk04: @zacw7 Search keys are initialized (inside `call()`) when the `keys` array parameter is NOT constant and varies per row, requiring the search keys to be rebuilt for each invocation. When keys are constant (e.g., `array[1, 2, 3]`), they're initialized once in `initialize()` and reused for better performance.

velox/functions/prestosql/MapExcept.h
abhinavmuk04: @zacw7 Added!

velox/functions/prestosql/MapExcept.h
abhinavmuk04: @zacw7 Added!

velox/functions/prestosql/MapExcept.h
zacw7: bump this

velox/functions/prestosql/MapExcept.h
zacw7: i meant shall we do `searchKeys_.reserve(keys->size())` as well

zacw7 (changes requested):

velox/functions/prestosql/MapExcept.h
abhinavmuk04: @zacw7 IIUC, the F14FastSet template requires the complete definition of MapExceptFunctionEqualComparator (not just a forward declaration) to instantiate type traits like std::is_empty, making it hard to move the comparator after MapExceptFunction without causing compilation errors. Let me know

PR #15078, opened 2025-10-07
fix: TIME intermediate type transform as VARCHAR

han-yan01:
Summary:
per discussion with @kgpai , we are changing using VARCHAR as the intermediate type transform.

previously we were using BIGINT but Presto does not support bidirectional BIGINT casting, thus fuzzer will still not be working with TIME.

Presto allows bidirectional casting for VARCHAR, so with this change fuzzer should now support TIME.

Enables TIME type for fuzzer that was previously disabled.

Differential Revision: D84077334

### DIFF
--- a/velox/exec/fuzzer/PrestoQueryRunner.cpp
+++ b/velox/exec/fuzzer/PrestoQueryRunner.cpp
@@ -338,10 +338,6 @@ bool PrestoQueryRunner::isSupported(const exec::FunctionSignature& signature) {
   //   - Under development in Presto
   //   - Cannot be used as constant literals
   //   - Expected differences between Presto Java and Velox C++ implementations
-  //
-  // time:
-  //   - Not supported by Fuzzer, will enable after intermediate type support
-  //   - Cannot be used as input/output types in Presto SQL
   return !(
       usesTypeName(signature, "interval year to month") ||
       usesTypeName(signature, "hugeint") ||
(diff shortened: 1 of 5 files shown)


velox/expression/tests/CastExprTest.cpp
```diff
@@ -3269,5 +3269,424 @@ TEST_F(CastExprTest, timeToBigintCast) {
...
+    // Times outside the DST gap should round-trip correctly
+    EXPECT_EQ(result->valueAt(0), "01:30:00.000");
+    EXPECT_EQ(result->valueAt(1), "03:30:00.000");
+  }
+
+  {
+    // Test DST gap times with try_cast - gap times should return null
+    auto varcharVector = makeFlatVector<StringView>({
+        "01:30:00.000", // Before gap - should work
+        "02:01:00.000", // In the gap - should return null
```
kgpai, line 3626: Can you validate this behavior is what is shown by Presto ?

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,145 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
   return StringView{startPos, kTimeToVarcharRowSize};
 }
 
+int64_t TimeType::valueToTime(const StringView& timeStr) const {
+  size_t pos = 0;
+  int32_t hour = 0, minute = 0, second = 0, millis = 0;
+  const char* data = timeStr.data();
+
+  // Helper function to parse a number from a string at a given position
+  auto parseNumber =
```
kgpai, line 1539: I do think you can simplify and make it very easy to parse with something like this : 
``` 
// Validate timeStr is right length etc..
try {
 auto hours = std::stoi(timeStr.substr(0, 2)); 
 VELOX_CHECK_EQ(timeStr[2], ':');
 auto minutes = std::stoi(timeStr.substr(3, 5));
 VELOX_CHECK_EQ(timeStr[5], ':');
 auto seconds = std::stoi(timeStr.substr(6, 7));
 if (timeStr[7] == '.') {
  // get optional milliseconds
  }
  } catch (...) {
   VELOX_USER_FAIL("Couldnt parse time " , timeStr);
 }
 ```

PR #15081, opened 2025-10-07
feat: Add TIME support to date_diff function

han-yan01:
Summary:
- added `diffTime()` function that calculates differences between TIME values
- simple millisecond arithmetic with unit conversion (TIME = ms since midnight)
- supports millisecond, second, minute, hour (rejects date-related units)
- `getTimeUnit()` for TIME-specific unit validation
- `initialize()` and `call()` method overloads for `<Time, Time>` parameters
- only allows time-related units, rejects day/month/year

Differential Revision: D84103016

### DIFF
--- a/velox/functions/prestosql/DateTimeImpl.h
+++ b/velox/functions/prestosql/DateTimeImpl.h
@@ -61,9 +61,7 @@ FOLLY_ALWAYS_INLINE Timestamp fromUnixtime(double unixtime) {
       seconds, milliseconds * Timestamp::kNanosecondsInMillisecond);
 }
 
-FOLLY_ALWAYS_INLINE boost::int64_t fromUnixtime(
-    double unixtime,
-    int16_t timeZoneId) {
+FOLLY_ALWAYS_INLINE int64_t fromUnixtime(double unixtime, int16_t timeZoneId) {
   if (FOLLY_UNLIKELY(std::isnan(unixtime))) {
     return pack(0, timeZoneId);
   }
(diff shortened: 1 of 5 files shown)


velox/functions/prestosql/DateTimeImpl.h
```diff
@@ -202,6 +200,32 @@ int64_t diffDate(
...
+  }
+
+  // TIME values are milliseconds since midnight
+  int64_t diffMillis = toTime - fromTime;
+
+  switch (unit) {
+    case DateTimeUnit::kMillisecond:
+      return diffMillis;
+    case DateTimeUnit::kSecond:
+      return diffMillis / kMillisInSecond;
```
PingLiuPing, line 229: Is this supposed to be multiply?
For example, suppose the diffMillis = 1; and 1/1000 should be a float value.

velox/type/Type.cpp
```diff
@@ -1530,6 +1530,192 @@ StringView TimeType::valueToString(int64_t value, char* const startPos) const {
...
+  int32_t minute = 0;
+  int32_t second = 0;
+  int32_t millis = 0;
+};
+
+// Parse a number with min/max digit constraints from the given position
+bool parseNumber(
+    const StringView& str,
+    size_t& pos,
+    int32_t& result,
```
PingLiuPing: remove &.

han-yan01: This is for the cast varchar to time PR - 
Correct me if I'm wrong but I think they are necessary cuz `pos` is passed by reference so it can be modified, `result` is passed by reference so it can be modified and return the parsed value. These are for the `parseNumber' lambda function with limited scope to modify both arguments by reference per effort to refactor/shrink the size of the valueToTime function
cc @kgpai

velox/functions/prestosql/DateTimeImpl.h
han-yan01: thanks for the review Ping,

correct me if i'm wrong but:
i think the division is actually correct for this use case:
- `diffTime` returns an `int64_t`, not a floating-point value
- we want to calculate the difference in the specified time unit
- say you have a time difference of 1 millisecond and you want to know how many complete seconds that represents, the answer is 0 seconds (since 1ms < 1 second)

which is consistent with how other date/time diff functions work - they return the number of complete units of the specified type:
*   999 milliseconds = 0 complete seconds
*   1000 milliseconds = 1 complete second
*   1001 milliseconds = 1 complete second (truncated)

wdyt @kgpai ?

PingLiuPing: > we want to calculate the difference in the specified time unit

Got it, thanks for the explanation.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -5709,3 +5709,122 @@ TEST_F(DateTimeFunctionsTest, localtime) {
...
+      dateDiff(
+          "minute", morning, afternoon)); // 315.254... minutes truncated to 315
+  EXPECT_EQ(
+      5, dateDiff("hour", morning, afternoon)); // 5.254... hours truncated to 5
+  EXPECT_EQ(
+      315,
+      dateDiff(
+          "minute", morning, afternoon)); // 315.254... minutes truncated to 315
+  EXPECT_EQ(
+      5, dateDiff("hour", morning, afternoon)); // 5.254... hours truncated to 5
```
PingLiuPing, line 6199: Seems these are duplicate.

velox/type/Type.cpp
PingLiuPing: Should be fine give it is in anonymous namespace.

velox/functions/prestosql/DateTimeImpl.h
kgpai: Would it be better if we relied on std::chrono to do this for us , rather than native operators ?

kgpai (approved):

PR #15089, opened 2025-10-08
misc: Optimization - std::move in Driver Output

natashasehgal:
Summary: AI4P Optimization: std::move in Driver Output, to avoid ref count

Differential Revision: D82071734

### DIFF
--- a/velox/exec/Driver.cpp
+++ b/velox/exec/Driver.cpp
@@ -443,8 +443,10 @@ inline void addInput(Operator* op, const RowVectorPtr& input) {
 }
 
 inline void getOutput(Operator* op, RowVectorPtr& result) {
-  result = op->getOutput();
-  if (FOLLY_UNLIKELY(op->shouldDropOutput())) {
+  auto output = op->getOutput();
+  if (FOLLY_LIKELY(!op->shouldDropOutput())) {
+    result = std::move(output); // Use move semantics to avoid ref counting
+  } else {
     result = nullptr;
   }
 }


velox/exec/Driver.cpp
```diff
@@ -443,8 +443,10 @@ inline void addInput(Operator* op, const RowVectorPtr& input) {
...
 inline void getOutput(Operator* op, RowVectorPtr& result) {
-  result = op->getOutput();
-  if (FOLLY_UNLIKELY(op->shouldDropOutput())) {
+  auto output = op->getOutput();
+  if (FOLLY_LIKELY(!op->shouldDropOutput())) {
+    result = std::move(output); // Use move semantics to avoid ref counting
+  } else {
     result = nullptr;
   }
 }
```
MBkkt, line 452: @natashasehgal Hi, could you explain this Optimization for me, please?
Maybe I'm missing something but to me code looks more complicated and without any evidence of performance improvement.
In both versions of code only move ctor/assing are used.

In old code was
```
1.  result = op->getOutput();
2.  if (FOLLY_UNLIKELY(op->shouldDropOutput())) {
3.   result = nullptr;
```
1. move-assign to result => destroy result if it was
2. unlikely cond check
3. if 2 then remove output buffer from in 1. step
4. if not 2 then nothing to do

In new code
```
1. auto output = op->getOutput();
2. if (FOLLY_LIKELY(!op->shouldDropOutput())) {
3.   result = std::move(output); // Use move semantics to avoid ref counting
    } else {
4.   result = nullptr;
```
1. move-ctor
2. likely not cond check
3. if 2 then move-assign output to result => destroy result if it was 
4. if not 2 then remove result and then remove output


So both
1. new 123 is exactly same with old 124, and in general just same code but written in a different way
2. new 124 is exactly same with old 123...

natashasehgal: Hi thanks for asking. I had a discussion with @Yuhta , who explained this change did not lead to any optimization. I didn't revert the change as it's a no-op.

Issue #14456, opened 2025-08-13
Reduce and isolate folly dependency


MBkkt:
### Description

What are benefits of avoiding folly?
* A lot of easier and faster build and contribute
* Less size of velox binaries, especially in test/debug environment
* A lot of easier to use velox in non-facebook C++ project

### Tasks

- [ ] FOLLY_LIKELY/UNLIKEY can be replaced with standard attributes https://github.com/facebookincubator/velox/issues/14398
- [ ] Other `FOLLY_`-macros can have `VELOX_` alternative https://github.com/facebookincubator/velox/issues/13202
- [x] Replace `folly::Optional` with `std::optional` https://github.com/facebookincubator/velox/pull/14455
- [ ] Replace `folly::StringPiece` with `std::string_view` (*1)
- [ ] Replace `folly::Range` with `std::span` (*1) https://github.com/facebookincubator/velox/pull/14743
- [ ] Isolate `folly::F14` `Value|Vector|Node|Fast` `Map|Set` to separate header (make `using ValueSet = ...`, etc), so anyone can redefine them in their fork/environment (examples of alternatives what can be used: abseil, boost [fast one](https://bannalia.blogspot.com/2022/11/inside-boostunorderedflatmap.html), or even [clickhouse code](https://github.com/kitaisreal/hash-table-aggregation-benchmark), etc). Also in case of folly not available velox usings should be using to `std::unordered_map`. This also allows to really easy benchmark change map implementation.
- [ ] Replace `folly::IntrusiveList` usings with velox usings. (inside folly it's just using to boost intrusive list)
- [ ] `folly::call_once` should be using in velox. If folly not available it can be `std::call_once`, others can use `absl::call_once` (another optimized alternative)
- [ ] velox using for `folly::SharedMutex` (if no folly => std or boost)

### TODO (not clear tasks)

- [ ] What to do with `folly::dynamic`, maybe we can make it separate and folly will swap to use it? Folly already use boost, so it should be ok? Also this part of folly doesn't depend a lot on other folly parts.
- [ ] Maybe don't use `folly::to`, instead use own velox wrappers?
- [ ] Isolate folly string manipulation functions (split, join, maybe some other)
- [ ] folly::Hash/DefaultHasher/etc => velox usings
- [ ] Isolate `folly::Executor`,  they're probably not really required for single thread execution
- [ ] Isolate `folly::Future/Promise`, they're probably not really required for single thread execution 
[...]

PR #15096, opened 2025-10-09
refactor: Remove usage of folly::StringPiece in HiveTypeParser

pedroerp:
Summary:
std::string_view.

Differential Revision: D84209938

### DIFF
--- a/velox/type/fbhive/HiveTypeParser.cpp
+++ b/velox/type/fbhive/HiveTypeParser.cpp
@@ -23,10 +23,10 @@
 #include "velox/common/base/Exceptions.h"
 
 namespace facebook::velox::type::fbhive {
-
 namespace {
+
 /// Returns true only if 'str' contains digits.
-bool isPositiveInteger(const std::string& str) {
+bool isPositiveInteger(std::string_view str) {
   return !str.empty() &&
       std::find_if(str.begin(), str.end(), [](unsigned char c) {
         return !std::isdigit(c);
@@ -37,6 +37,17 @@ bool isSupportedSpecialChar(char c) {
   static std::unordered_set<char> supported{'_', '$', '#'};
   return supported.count(c) == 1;
 }
+
+/// Returns whether `prefix` is non-empty case-insensitive prefix of `input`.
+bool isPrefix(std::string_view input, std::string_view prefix) {
+  return prefix.size() > 0 && input.size() >= prefix.size() &&
+      std::equal(
+             prefix.begin(),
+             prefix.end(),
+             input.begin(),
+             folly::AsciiCaseInsensitive{});
+}
+
 } // namespace
 
 HiveTypeParser::HiveTypeParser() {
@@ -69,15 +80,15 @@ HiveTypeParser::HiveTypeParser() {
   setupMetadata<TokenType::EndOfStream, TypeKind::INVALID>();
 }
 
-TypePtr HiveTypeParser::parse(const std::string& ser) {
-  remaining_ = folly::StringPiece(ser);
+TypePtr HiveTypeParser::parse(std::string_view input) {
+  remaining_ = input;
   Result result = parseType();
   VELOX_CHECK(
       remaining_.size() == 0 || TokenType::EndOfStream == lookAhead(),
       "Input remaining after parsing the Hive type \"{}\"\n"
       "Remaining: \"{}\"",
-      ser,
-      remaining_.toString());
+      input,
+      remaining_);
   return result.type;
 }
 
@@ -88,8 +99,8 @@ Result HiveTypeParser::parseType() {
   if (!nt.isValidType()) {
     VELOX_FAIL(
         "Unexpected token {} at {}. typeKind = {}",
-        nt.value.toString(),
-        remaining_.toString(),
+        nt.value,
+        remaining_,
         nt.typeKind());
   }
 
@@ -98,12 +109,12 @@ Result HiveTypeParser::parseType() {
       eatToken(TokenType::LeftRoundBracket);
       Token precision = nextToken();
       VELOX_CHECK(
-          isPositiveInteger(precision.value.toString()),
+          isPositiveInteger(precision.value),
           "Decimal precision must be a positive integer");
       eatToken(TokenType::Comma);
       Token scale = nextToken();
       VELOX_CHECK(
-          isPositiveInteger(scale.value.toString()),
+          isPositiveInteger(scale.value),
           "Decimal scale must be a positive integer");
       eatToken(TokenType::RightRoundBracket);
       return Result{DECIMAL(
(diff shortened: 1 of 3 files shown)


velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -163,18 +164,19 @@ Result HiveTypeParser::parseType() {
...
       return ResultList{std::move(subTypeList), std::move(names)};
     }
 
-    folly::StringPiece fieldName;
+    std::string_view fieldName;
     if (hasFieldNames) {
       fieldName = eatToken(TokenType::Identifier, true).value;
       eatToken(TokenType::Colon);
-      names.push_back(fieldName.toString());
+      names.push_back(std::string(fieldName));
```
MBkkt: emplace_back instead of push_back(string

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -196,7 +198,7 @@ Token HiveTypeParser::eatToken(TokenType tokenType, bool ignorePredefined) {
     return token;
   }
 
-  VELOX_FAIL("Unexpected token " + token.remaining.toString());
+  VELOX_FAIL("Unexpected token ", token.remaining);
```
MBkkt: Is it correct? I thought something like this required
I hoped this checked in compile time, but maybe it's not...
```suggestion
  VELOX_FAIL("Unexpected token {}", token.remaining);
```

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -70,14 +71,14 @@ HiveTypeParser::HiveTypeParser() {
 }
 
 TypePtr HiveTypeParser::parse(const std::string& ser) {
```
MBkkt: parse input also can be string_view

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -119,18 +120,18 @@ Result HiveTypeParser::parseType() {
...
     return Result{scalarType};
   } else if (nt.isOpaqueType()) {
     eatToken(TokenType::StartSubType);
-    folly::StringPiece innerTypeName =
+    std::string_view innerTypeName =
         eatToken(TokenType::Identifier, true).value;
     eatToken(TokenType::EndSubType);
 
-    auto typeIndex = getTypeIdForOpaqueTypeAlias(innerTypeName.str());
+    auto typeIndex = getTypeIdForOpaqueTypeAlias(std::string(innerTypeName));
```
MBkkt, line 146: We probably want TODO/issue about make `getTypeIdForOpaqueTypeAlias` accept string_view and remove here string construction.

We want to write this explicit because it's easy to forget remove here string construction, when `getTypeIdForOpaqueTypeAlias` actually starts to accept string_view

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -206,40 +208,45 @@ Token HiveTypeParser::nextToken(bool ignorePredefined) {
...
     ++iter;
   }
 
   if (len > 0) {
-    return makeExtendedToken(getMetadata(TokenType::Identifier), sp, len);
+    return makeExtendedToken(getMetadata(TokenType::Identifier), sv, len);
   }
 
-  VELOX_FAIL("Bad Token at " + sp.toString());
+  VELOX_FAIL("Bad Token at ", sv);
```
MBkkt:
```suggestion
  VELOX_FAIL("Bad Token at {}", sv);
```
same here? I'm not sure, if I'm missing something and it's correct, please resolve this

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -206,40 +208,45 @@ Token HiveTypeParser::nextToken(bool ignorePredefined) {
...
+        std::string_view match(token);
         if (match.size() > 0 &&
-            sp.startsWith(match, folly::AsciiCaseInsensitive{})) {
-          return makeExtendedToken(metadata.get(), sp, match.size());
+            // If `sv` starts with `match`.
+            std::equal(
+                sv.begin(),
+                sv.end(),
+                match.begin(),
+                folly::AsciiCaseInsensitive{})) {
```
MBkkt: I think it's not same, because now 
you're checking that sv equal to match

https://en.cppreference.com/w/cpp/algorithm/equal.html
> For overloads (1-4), the second range has [std::distance](https://en.cppreference.com/w/cpp/iterator/distance.html)(first1, last1) elements.

```suggestion
            match.size() <= sv.size() &&
            std::equal(
                match.begin(),
                match.end(),
                sv.begin(),
                folly::AsciiCaseInsensitive{})) {
```

velox/type/fbhive/HiveTypeParser.cpp
MBkkt: Yes, I think it's wrong
https://godbolt.org/z/3KWjrP1eP

MBkkt: https://github.com/fmtlib/fmt/issues/492

pedroerp: Ugh, nice catch!!

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: No, you're right. Nice catch :)

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: Good point. This changes the external API, but should be reasonable harmless (assuming don't change the string contents).

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: I actually missed C++20 also has a starts_with() method for strings. Will make the change.

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: Will leave the TODO for now, but `getTypeIdForOpaqueTypeAlias()` cannot take a string_view too easily become it needs to lookup a map where std::string is the key. We would need to change the map type itself, I guess.

velox/type/fbhive/HiveTypeParser.cpp
MBkkt: I think `const std::string&` in arguments to `std::string_view` is probably almost always backwards compatible.

velox/type/fbhive/HiveTypeParser.cpp
MBkkt: > it needs to lookup a map where std::string is the key. We would need to change the map type itself, I guess.

You're right. We need to make it FastHashMap/NodeHashMap I think (in general any map with transparent lookup will suite for this, but I think better to use faster maps).

I have some plans about this, in short:
1. Create using for something like velox::FlatHashMap, velox::NodeHashMap
2. Publish real-data benchmarks about reliable Hash+HashMap implementation (abseil, folly, libc++, etc)
3. Replace most code with usings

Well, all this takes some time unfortunately...

velox/type/fbhive/HiveTypeParser.cpp
```diff
@@ -272,15 +285,15 @@ int8_t HiveTypeParser::makeTokenId(TokenType tokenType) const {
 
 TokenAndRemaining HiveTypeParser::makeExtendedToken(
     TokenMetadata* tokenMetadata,
-    folly::StringPiece sp,
+    std::string_view sv,
     size_t len) const {
-  folly::StringPiece spmatch(sp.cbegin(), sp.cbegin() + len);
-  sp.advance(len);
+  std::string_view spmatch(sv.cbegin(), sv.cbegin() + len);
```
MBkkt:
```suggestion
  std::string_view spmatch{sv.data(), len};
```

> https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view.html

It's also faster (at least in debug :) )


As I mentioned string_view is pointer + size, so pointer + size ctor a little better then 2 iterators

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: That would be nice. These hash maps are usually not on the very hot path for query execution, so I'm not sure they would affect things in a meaningful way. We did some hash table perf evaluation back in the day for dictionary encoding streams in the reader, as we care very much about those. I think in those cases we usually use folly's F14.

velox/type/fbhive/HiveTypeParser.cpp
pedroerp: Thanks. Will make the change.

PR #15097, opened 2025-10-09
perf: Optimize Window with SubPartitionedSortWindowBuild

HolyLow:
This PR optimizes Window execution by adding a intermediate step called SubPartitionedSortWindowBuild. With SubPartitionedSortWindowBuild, the data is firstly redistributed into pre-defined number of sub partitions, for example, 2. Then window partitioning and window computation is carried out sequentially, one sub partition at a time.
 
This PR adds a new config `window_num_sub_partitions` to decide how many sub partitions the window data is divided into. The default value is 1, which means we do not use sub partitioning.

As each sub partition only holds part of data, the execution time (with sort-based partitioning) or memory requirement (with prefixSort based partitioning, or potential hashTable based partitioning) is reduced. After a sub partition is consumed completely, its data is dropped.

Fixes https://github.com/facebookincubator/velox/issues/15098.

## Result
Benchmarking result of WindowRegionSortBenchmark, run on Intel(R) Xeon(R) 6982P-C (up to 3900MHz), ubuntu 22.04:
- Raw result output is in https://docs.google.com/document/d/1w04GHuEO3WsKDNCNiLAPxjlGYaJRL06OCNn4YAF3tZo/edit?usp=sharing
- Collected data and analysis is in https://docs.google.com/spreadsheets/d/1lI3CgTcAoQCeHO0KNIVf3ZIOWpcRRRJRD2vG9hq38pA/edit?usp=sharing
According to the results and analysis, the proper numSubPartitions would be 16~32 to achieve both performance improvement (6%~17%) and memory saving (20%~30% memory decrease) across all kinds of cases.

### DIFF
--- a/velox/exec/benchmarks/CMakeLists.txt
+++ b/velox/exec/benchmarks/CMakeLists.txt
@@ -136,10 +136,10 @@ target_link_libraries(
   Folly::follybenchmark
 )
 
-add_executable(velox_window_regionsort_benchmark WindowSubPartitionedSortBenchmark.cpp)
+add_executable(velox_window_sub_partitioned_sort_benchmark WindowSubPartitionedSortBenchmark.cpp)
 
 target_link_libraries(
-  velox_window_regionsort_benchmark
+  velox_window_sub_partitioned_sort_benchmark
   velox_aggregates
   velox_exec
   velox_exec_test_lib
--- a/velox/exec/benchmarks/WindowSubPartitionedSortBenchmark.cpp
+++ b/velox/exec/benchmarks/WindowSubPartitionedSortBenchmark.cpp
@@ -34,14 +34,56 @@ using namespace facebook::velox::test;
 using namespace facebook::velox::exec;
 using namespace facebook::velox::exec::test;
 
-static constexpr int32_t kNumVectors = 500;
-static constexpr int32_t kRowsPerVector = 1'0000;
+static constexpr int32_t kNumVectors = 200;
+static constexpr int32_t kRowsPerVector = 10'000;
 
 namespace {
+class BenchmarkRecorder {
+ public:
+  BenchmarkRecorder() = default;
+
+  void record(std::string name, uint64_t numBytes) {
+    // Only record the first apperance.
+    if (numBytesRecords_.count(name) == 0) {
+      numBytesRecords_[name] = {1, numBytes};
+      names_.push_back(name);
+    } else {
+      auto& record = numBytesRecords_[name];
+      record.numAppearance++;
+      record.totalCount += numBytes;
+    }
+  }
+
+  std::string report() {
+    std::string result = "name, memory(MB)\n";
+    for (auto& name : names_) {
+      auto& record = numBytesRecords_[name];
+      result += fmt::format(
+          "{}, {}MB\n",
+          name,
+          record.totalCount / 1024 / 1024 / record.numAppearance);
+    }
+    return result;
+  }
+
+ private:
+  struct Counter {
+    int32_t numAppearance{0};
+    uint64_t totalCount{0};
+  };
+  std::vector<std::string> names_;
+  std::unordered_map<std::string, Counter> numBytesRecords_;
+};
 
 class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
  public:
-  explicit WindowSubPartitionedSortBenchmark() {
+  WindowSubPartitionedSortBenchmark(
+      int32_t numVectors,
+      int32_t rowsPerVector,
+      std::shared_ptr<BenchmarkRecorder> recorder)
+      : numVectors_(numVectors),
+        rowsPerVector_(rowsPerVector),
+        recorder_(recorder) {
     memory::SharedArbitrator::registerFactory();
     HiveConnectorTestBase::SetUp();
     aggregate::prestosql::registerAllAggregateFunctions();
@@ -63,23 +105,23 @@ class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
     });
 
     VectorFuzzer::Options opts;
-    opts.vectorSize = kRowsPerVector;
+    opts.vectorSize = rowsPerVector_;
     opts.nullRatio = 0;
     VectorFuzzer fuzzer(opts, pool_.get(), FLAGS_fuzzer_seed);
     std::vector<RowVectorPtr> inputVectors;
-    for (auto i = 0; i < kNumVectors; ++i) {
+    for (auto i = 0; i < numVectors_; ++i) {
       std::vector<VectorPtr> children;
 
       // Generate key with a small number of unique values from a small range
       // (0-16).
       children.emplace_back(
           makeFlatVector<int32_t>(
-              kRowsPerVector, [](auto row) { return row % 17; }));
+              rowsPerVector_, [](auto row) { return row % 17; }));
 
       // Generate key with a small number of unique values from a large range
       // (300 total values).
       children.emplace_back(
-          makeFlatVector<int32_t>(kRowsPerVector, [](auto row) {
+          makeFlatVector<int32_t>(rowsPerVector_, [](auto row) {
             if (row % 3 == 0) {
               return std::numeric_limits<int32_t>::max() - row % 100;
             } else if (row % 3 == 1) {
@@ -97,7 +139,7 @@ class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
       // order.
       children.emplace_back(
           makeFlatVector<int32_t>(
-              kRowsPerVector, [](auto row) { return row; }));
+              rowsPerVector_, [](auto row) { return row; }));
 
       // Generate random values without nulls.
       children.emplace_back(fuzzer.fuzzFlat(INTEGER()));
@@ -132,19 +174,15 @@ class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
   void TestBody() override {}
 
   void run(
+      const std::string& recordName,
       const std::string& key,
       const std::string& aggregate,
-      bool subPartitionedSort = true) {
+      int32_t numSubPartitions) {
     folly::BenchmarkSuspender suspender1;
 
-    if ((subPartitionedSort && !lastRunSubPartitionedSort_) ||
-        (!subPartitionedSort && lastRunSubPartitionedSort_)) {
-      report();
-      windowNanos_.clear();
-      windowMems_.clear();
-    }
+    windowNanos_.clear();
+    windowMems_.clear();
 
-    lastRunSubPartitionedSort_ = subPartitionedSort;
     std::string functionSql = fmt::format(
         "{} over (partition by {} order by k_sort)", aggregate, key);
 
@@ -156,7 +194,7 @@ class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
                                   .planFragment();
 
     vector_size_t numResultRows = 0;
-    auto task = makeTask(plan, subPartitionedSort);
+    auto task = makeTask(plan, numSubPartitions);
     task->addSplit(
         tableScanPlanId,
         exec::Split(makeHiveConnectorSplit(sourceFilePath_->getPath())));
@@ -183,17 +221,19 @@ class WindowSubPartitionedSortBenchmark : public HiveConnectorTestBase {
         }
       }
     }
+    recorder_->record(recordName, windowMems_.peakTotalMemoryReservation);
     suspender2.dismiss();
     folly::doNotOptimizeAway(numResultRows);
   }
 
   std::shared_ptr<exec::Task> makeTask(
       core::PlanFragment plan,
-      bool subPartitionedSort) {
+      int32_t numSubPartitions) {
+    bool subPartitionedSort = numSubPartitions > 1;
     if (subPartitionedSort) {
       const std::unordered_map<std::string, std::string> queryConfigMap(
           {{core::QueryConfig::kWindowNumSubPartitions,
-            std::to_string(numSubPartitions_)}});
+            std::to_string(numSubPartitions)}});
       return exec::Task::create(
           "t",
           std::move(plan),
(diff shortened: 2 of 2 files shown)


pedroerp:
@HolyLow could you add a summary to the PR, describing in details what it does and how it works?

HolyLow:
@pedroerp Thanks for your reply. The detailed description is in https://github.com/facebookincubator/velox/issues/15098, and I've added summary as suggested.

Any further suggestion is welcome, thanks a lot.

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,197 @@
...
+// todo: eliminate this. this is not used.
+std::vector<CompareFlags> makeCompareFlags(
+    int32_t numPartitionKeys,
+    const std::vector<core::SortOrder>& sortingOrders) {
+  std::vector<CompareFlags> compareFlags;
+  compareFlags.reserve(numPartitionKeys + sortingOrders.size());
+
+  for (auto i = 0; i < numPartitionKeys; ++i) {
+    compareFlags.push_back({});
+  }
```
MBkkt:
```suggestion
  compareFlags.resize(numPartitionKeys + sortingOrders.size());
```

velox/exec/WindowBuild.h
```diff
@@ -70,14 +70,17 @@ class WindowBuild {
...
   }
 
   /// Releases the memory held by the window build. This is called by the
   /// window operator when all rows have been processed.
   void release() {
-    data_->clear();
+    // todo: why this..?
+    if (data_) {
+      data_->clear();
+    }
```
MBkkt: Because `data_.reset()` in ctor? here

```
  // todo: to make sure we won't use incorrect data, release the original
  // data container...
  data_.reset();
```

velox/exec/Window.cpp
```diff
@@ -55,16 +56,31 @@ Window::Window(
...
-        pool(),
-        common::PrefixSortConfig{
-            driverCtx->queryConfig().prefixSortNormalizedKeyMaxBytes(),
-            driverCtx->queryConfig().prefixSortMinRows(),
-            driverCtx->queryConfig().prefixSortMaxStringPrefixLength()},
-        spillConfig,
-        &nonReclaimableSection_,
-        spillStats_.get());
+    auto regionNum = operatorCtx_->driverCtx()->queryConfig().windowRegionNum();
+    if (regionNum > 1) {
```
MBkkt: `else if (const auto regionNum = operatorCtx_->driverCtx()->queryConfig().windowRegionNum(); numRegion > 1)`

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // todo: to make sure we won't use incorrect data, release the original
+  // data container...
+  data_.reset();
+  VELOX_CHECK_NOT_NULL(pool_);
+}
+
+void RegionSortWindowBuild::addInput(RowVectorPtr input) {
+  regionIdsBuffer_.resize(input->size());
+  regionPartitionFunction_->partition(*input, regionIdsBuffer_);
```
MBkkt: I don't know, maybe it's better to introduce non-parallel local partition node and operator instead?
Or even better, extend current local partition operator to N:M model.
N -- number of partitions to generate
M -- number of threads to execute
M divides N

velox/exec/RegionSortWindowBuild.cpp
HolyLow: Seems the compareFlags are redundant. Removed this part of code directly.

velox/exec/Window.cpp
HolyLow: I think fusing the region condition to the upper level if-else branch would make the code hard to read. In my thought, the sub-cases under different region settings should stay in the nested level. 
Yet it is good to fuse the local variable `regionNum` into if condition declaration. Refactored as suggested.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: I think there is no such mechanism in Velox currently, and it seems that the non-parallel local partition operation has no other use cases. So personally I think maybe we could leave this to future work.
Besides, fusing the local-partition operation inside of WindowBuild would also leave space for further optimization, e.g. use different algorithms for different regions.

velox/exec/WindowBuild.h
HolyLow: done.

HolyLow:
@mbasmanova Could you help review this PR which optimizes window with RegionBuild? Thanks a lot.

mbasmanova:
@HolyLow CI is red. Would you rebase to see if the errors go away? If not, please, investigate.

mbasmanova:
> Benchmarking result of WindowRegionSortBenchmark, run on Intel(R) Xeon(R) 6982P-C (up to 3900MHz), ubuntu 22.04:

These results are very long and not readable. Would you move these into a separate Google Sheet or Gist and provide a link along with an easy-to-understand summary?

Would you also provide guidance on how to choose the number of regions? 

How would one specify the number of regions? Is there a new config? Please, clarify in the PR description.

velox/core/QueryConfig.h
```diff
@@ -490,6 +490,12 @@ class QueryConfig {
   static constexpr const char* kDriverCpuTimeSliceLimitMs =
       "driver_cpu_time_slice_limit_ms";
 
+  /// Number of regions when building window partitions. If region number is
+  /// greater than 1, window would divide data into multiple regions and compute
+  /// each region separately to accelerate computation and reduce memory usage.
+  static constexpr const char* kWindowRegionNum =
```
mbasmanova: kWindowNumRegions + window_num_regions.

Please, update documentation as well.

velox/core/QueryConfig.h
```diff
@@ -1226,6 +1232,10 @@ class QueryConfig {
     return get<uint32_t>(kDriverCpuTimeSliceLimitMs, 0);
   }
 
+  uint32_t windowRegionNum() const {
```
mbasmanova: windowNumRegions

velox/exec/Window.cpp
```diff
@@ -55,17 +56,34 @@ Window::Window(
...
-        pool(),
-        common::PrefixSortConfig{
-            driverCtx->queryConfig().prefixSortNormalizedKeyMaxBytes(),
-            driverCtx->queryConfig().prefixSortMinRows(),
-            driverCtx->queryConfig().prefixSortMaxStringPrefixLength()},
-        spillConfig,
-        &nonReclaimableSection_,
-        &stats_,
-        spillStats_.get());
+    if (auto regionNum =
```
mbasmanova: I wonder if we need this new term "region". Is "region" effectively a set of window partitions? In other words, region is just a partition on top of window partitions, no?

Typically, region implies a contiguous chunk, but here this is not the case.

velox/exec/Window.cpp
```diff
@@ -55,17 +56,34 @@ Window::Window(
...
+            operatorCtx_->driverCtx()->queryConfig().windowRegionNum();
+        regionNum > 1) {
+      windowBuild_ = std::make_unique<RegionSortWindowBuild>(
```
mbasmanova: > RegionSortWindowBuild

Would PartitionedSortWindowBuild be a better name? CC: @aditi-pandit

velox/exec/Window.cpp
mbasmanova: regionNum -> numRegions

velox/exec/Window.cpp
```diff
@@ -55,17 +56,34 @@ Window::Window(
...
+          windowNode,
+          regionNum,
+          pool(),
+          common::PrefixSortConfig{
+              driverCtx->queryConfig().prefixSortNormalizedKeyMaxBytes(),
+              driverCtx->queryConfig().prefixSortMinRows(),
+              driverCtx->queryConfig().prefixSortMaxStringPrefixLength()},
```
mbasmanova: this code repeats; please, extract a helper function

velox/exec/SortWindowBuild.h
```diff
@@ -29,7 +29,7 @@ class SortWindowBuild : public WindowBuild {
   SortWindowBuild(
       const std::shared_ptr<const core::WindowNode>& node,
       velox::memory::MemoryPool* pool,
-      common::PrefixSortConfig&& prefixSortConfig,
+      const common::PrefixSortConfig& prefixSortConfig,
```
mbasmanova: this change seems unrelated; is it needed? Can it be extracted into a separate PR?

velox/exec/SortWindowBuild.h
```diff
@@ -46,6 +46,11 @@ class SortWindowBuild : public WindowBuild {
 
   void addInput(RowVectorPtr input) override;
 
+  void addInputRow(
```
mbasmanova: Why is this a public method? Please, document it.

velox/exec/SortWindowBuild.h
```diff
@@ -56,9 +61,9 @@ class SortWindowBuild : public WindowBuild {
 
   std::shared_ptr<WindowPartition> nextPartition() override;
 
- private:
   void ensureInputFits(const RowVectorPtr& input);
```
mbasmanova, line 63: Why is this now a public method?

velox/exec/RegionSortWindowBuild.h
```diff
@@ -0,0 +1,90 @@
...
+
+  void ensureInputFits(const RowVectorPtr& input);
+
+  const int32_t numRegions_;
+
+  const size_t numPartitionKeys_;
+
+  std::unique_ptr<HashPartitionFunction> regionPartitionFunction_;
+
+  std::vector<std::unique_ptr<SortWindowBuild>> windowBuilds_;
```
mbasmanova: Would you document these properties? Is this a list of SortWindowBuilds for each region? Is the size of this vector == numRegions_?

velox/exec/RegionSortWindowBuild.h
```diff
@@ -0,0 +1,90 @@
...
+  bool spilled_{false};
+
+  std::vector<uint32_t> regionIdsBuffer_;
+
+  int32_t currentRegion_ = -1;
+
+  memory::MemoryPool* const pool_;
+
+  // Config for Prefix-sort.
+  const common::PrefixSortConfig prefixSortConfig_;
```
mbasmanova: Please, group constant member variables together before non-constant.

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+        node,
+        pool,
+        prefixSortConfig_,
+        spillConfig,
+        nonReclaimableSection,
+        opStats,
+        spillStats);
+  }
+  data_.reset();
+  VELOX_CHECK_NOT_NULL(pool_);
```
mbasmanova: move this to the start of the method

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+    windowBuilds_[i] = std::make_unique<SortWindowBuild>(
+        node,
+        pool,
+        prefixSortConfig_,
+        spillConfig,
+        nonReclaimableSection,
+        opStats,
+        spillStats);
+  }
+  data_.reset();
```
mbasmanova: Why is this needed? Isn't it already null?

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+  regionIdsBuffer_.resize(input->size());
+  regionPartitionFunction_->partition(*input, regionIdsBuffer_);
+
+  for (auto i = 0; i < inputChannels_.size(); ++i) {
+    decodedInputVectors_[i].decode(*input->childAt(inputChannels_[i]));
+  }
+
+  ensureInputFits(input);
+
+  VELOX_CHECK_LT(currentRegion_, 0);
```
mbasmanova: Should we move this check to the start of this method?

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+  for (auto row = 0; row < input->size(); ++row) {
+    auto& windowBuild = windowBuilds_[regionIdsBuffer_[row]];
+    windowBuild->addInputRow(decodedInputVectors_, row, input->childrenSize());
```
mbasmanova: It seems like input->childrenSize() parameter is redundant. Isn't it == decodedInputVectors_.size()?

velox/exec/SortWindowBuild.cpp
```diff
@@ -85,6 +85,19 @@ void SortWindowBuild::addInput(RowVectorPtr input) {
   numRows_ += input->size();
 }
 
+void SortWindowBuild::addInputRow(
```
mbasmanova: I see you provide this API to be used in the RegionSortWindowBuild. However, this introduces duplication of logic with SortWindowBuild::addInput. Would you refactor addInput to use this new API to avoid the duplication?

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  numRows_ += input->size();
+}
+
+bool RegionSortWindowBuild::switchToNextRegion() {
```
mbasmanova: Please, document this method in the .h file.

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+  if (currentRegion_ >= 0) {
+    windowBuilds_[currentRegion_].reset();
+  }
+  currentRegion_++;
+  if (currentRegion_ >= numRegions_) {
+    return false;
+  }
+
+  VELOX_CHECK_NOT_NULL(windowBuilds_[currentRegion_]);
+  windowBuilds_[currentRegion_]->noMoreInput();
```
mbasmanova: Would you add a comment to explain why we call noMoreInput here as opposed to in RegionSortWindowBuild::noMoreInput. I assume this is because SortWindowBuild::noMoreInput starts processing by sorting the data, which allocates memory and we want to avoid allocating this extra memory for all regions at once.

velox/exec/tests/WindowTest.cpp
```diff
@@ -431,6 +431,88 @@ DEBUG_ONLY_TEST_F(WindowTest, valuesRowsStreamingWindowBuild) {
   ASSERT_TRUE(isStreamCreated.load());
 }
 
+TEST_F(WindowTest, regionBuild) {
+  const vector_size_t size = 1'000;
+  const int partitionNum = 37;
+  const int regionNum = 4;
```
mbasmanova: What happens if regionNum is a multiple of partitionNum, e.g. partitionNum = 2 and regionNum = 4? Will we end up with skewed regions? Should we have some protection against that?

velox/exec/tests/WindowTest.cpp
```diff
@@ -431,6 +431,88 @@ DEBUG_ONLY_TEST_F(WindowTest, valuesRowsStreamingWindowBuild) {
...
+
+  createDuckDbTable({data});
+
+  core::PlanNodeId windowId;
+  auto plan =
+      PlanBuilder()
+          .values(split(data, 10))
+          .window({"row_number() over (partition by p order by s desc)"})
+          .capturePlanNodeId(windowId)
+          .orderBy({"s"}, false)
```
mbasmanova: Why is it important to add this orderBy?

velox/exec/tests/WindowTest.cpp
```diff
@@ -431,6 +431,88 @@ DEBUG_ONLY_TEST_F(WindowTest, valuesRowsStreamingWindowBuild) {
...
+          .planNode();
+
+  AssertQueryBuilder(plan, duckDbQueryRunner_)
+      .config(core::QueryConfig::kPreferredOutputBatchBytes, "1024")
+      .config(core::QueryConfig::kWindowRegionNum, std::to_string(regionNum))
+      .assertResults(
```
mbasmanova, line 464: Shouldn't we use something like assertSortedResults? assertResults doesn't check the order of rows.

velox/exec/Window.cpp
HolyLow: @mbasmanova The naming here really made me crazy actually. The major concern is that the WindowPartition already uses the `Partition` semantic, and using `Partitioned` again might make the naming even more confusing, as it is hard to distinguish the two partition naming.

I agree that the current `Region` naming might not be that proper, but really I don't know which word would be more concise and accurate.

cc @aditi-pandit

mbasmanova: Let's ask LLM for suggestions.

velox/exec/RegionSortWindowBuild.cpp
mbasmanova: @MBkkt might have a good point. It seems that currently we would use LocalPartition to partition N ways and process on N threads concurrently. However, it seems that we would want to partition N ways, but process one partition at a time. It might indeed be better to add support for such partitioning without coupling this with WindowOperator logic. This would also help solve the naming problem :-) ... no longer we would need to come up with a better name for "region".

velox/exec/MultiGroupSortWindowBuild.h
```diff
@@ -0,0 +1,90 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/exec/HashPartitionFunction.h"
+#include "velox/exec/PrefixSort.h"
+#include "velox/exec/SortWindowBuild.h"
+#include "velox/exec/Spiller.h"
+#include "velox/exec/WindowBuild.h"
```
aditi-pandit: This include might not be needed

velox/exec/RegionSortWindowBuild.h
```diff
@@ -0,0 +1,90 @@
...
+
+  const size_t numPartitionKeys_;
+
+  std::unique_ptr<HashPartitionFunction> regionPartitionFunction_;
+
+  std::vector<std::unique_ptr<SortWindowBuild>> windowBuilds_;
+
+  bool spilled_{false};
+
+  std::vector<uint32_t> regionIdsBuffer_;
```
aditi-pandit: What does this represent ? Comments for the variables would be useful.

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+      numRegions_(numRegions),
+      numPartitionKeys_{node->partitionKeys().size()},
+      pool_(pool),
+      prefixSortConfig_(prefixSortConfig),
+      spillStats_(spillStats) {
+  std::vector<column_index_t> keyChannels(numPartitionKeys_);
+  for (int i = 0; i < numPartitionKeys_; i++) {
+    keyChannels[i] = inputChannels_[i];
+  }
+  regionPartitionFunction_ = std::make_unique<HashPartitionFunction>(
```
aditi-pandit: Did you consider using only a subset of all the partition keys ? If the number of partitions is very coarse, then just picking the first or second partition key to demarcate into regions might be sufficient.

Also did you consider simple modulo function based grouping vs using HashPartition function ?

velox/exec/RegionSortWindowBuild.cpp
aditi-pandit: Maybe call this method addDecodedRow instead ?

velox/exec/RegionSortWindowBuild.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  VELOX_CHECK_LT(currentRegion_, 0);
+  for (auto& windowBuild : windowBuilds_) {
+    windowBuild->ensureInputFits(input);
+  }
+}
+
+void RegionSortWindowBuild::spill() {
+  VELOX_CHECK_LT(currentRegion_, 0);
+  for (auto& windowBuild : windowBuilds_) {
```
aditi-pandit: Why are you spilling all regions ? Like HashAgg, we can choose to spill only the biggest region to free up memory.

velox/exec/Window.cpp
aditi-pandit: Maybe SortByGroupWindowBuild/SortByRegionWindowBuild or MultiGroupSortWindowBuild ?
@mbasmanova

aditi-pandit (commented):
Thanks @HolyLow for this code.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: @mbasmanova The idea of separating the intra-thread partitioning logic into a standalone component might seem ideal, but the implementation has many critical obstacles. 
Firstly, is the standalone intra-thread partitioning logic a new operator? If this is the case, we have to change the plan translation logic as well, and having this new operator in the final physical plan seems a bit confusing. 
Secondly, how are we going to transfer the data from the new component to window operator in one-partition-then-another manner? Currently, the execution model of an operator is to accept data via `addInput` and marks data ending via `noMoreInput`. But now we might have to have another interface like `finishOnePartition`, so that the window operator would consume and output the current data, clear the data and get ready for next batch. This is completely incompatible to current driver pipeline's execution model.
Thirdly, the data encoding flow brought by the new execution model might bring much overhead. Data is transferred between operators as pages, but the data partitioning logic and the sorting logic inside of Window are based on RowContainer, so we have to encode data from page to page to RowContainer, which brings more overhead than Region-based encoding from page to RowContainer directly. Though this could be overcome by using dictionary encoding, it still brings unnecessary overhead.
Last but not least, even if we do pay great effort to overcome the above challenges, the intra-thread partitioning logic is not used by any other operator or algorithm and cannot be reused. So it is still under doubt that whether it is worth the pain.

In conclusion, for simplicity and clarity I think it is better to wrap the logic inside of WindowBuild rather than expose it out of WindowOperator.

velox/exec/Window.cpp
HolyLow: Or maybe we could call this PrePartitionedWindowBuild, and call the "regions" subWindowBuilds. Do you think this is proper? @mbasmanova @aditi-pandit And it is also good to call them "groups" as @aditi-pandit suggests.

velox/exec/RegionSortWindowBuild.cpp
mbasmanova: @HolyLow Thank you for detailed response. I agree with you. I think it was useful to think about @MBkkt's proposal though.

velox/core/QueryConfig.h
```diff
@@ -490,6 +490,11 @@ class QueryConfig {
   static constexpr const char* kDriverCpuTimeSliceLimitMs =
       "driver_cpu_time_slice_limit_ms";
 
+  /// Number of groups when building window partitions. If group number is
```
mbasmanova: Perhaps, we could adopt the term 'sub-partition'.

```
/// Window operator can be configured to sub-divide window partitions on each thread of execution into groups of partitions for sequential processing. This setting specifies how many sub-partitions to create for each thread. 
kWindowNumSubPartitions
```

velox/exec/SortWindowBuild.h
```diff
@@ -46,6 +46,10 @@ class SortWindowBuild : public WindowBuild {
 
   void addInput(RowVectorPtr input) override;
 
+  void addDecodedInputRow(
+      std::vector<DecodedVector>& decodedInputVectors,
+      size_t row);
```
mbasmanova: size_t -> vector_size_t

velox/exec/SortWindowBuild.cpp
```diff
@@ -85,6 +85,18 @@ void SortWindowBuild::addInput(RowVectorPtr input) {
   numRows_ += input->size();
 }
 
+void SortWindowBuild::addDecodedInputRow(
```
mbasmanova, line 83: Would you refactor SortWindowBuild::addInput to use this new method?

velox/exec/MultiGroupSortWindowBuild.h
```diff
@@ -0,0 +1,92 @@
...
+
+  // Partition each row to the corresponding group.
+  std::unique_ptr<HashPartitionFunction> groupPartitionFunction_;
+
+  // WindowBuilds for each group.
+  std::vector<std::unique_ptr<SortWindowBuild>> groupWindowBuilds_;
+
+  bool spilled_{false};
+
+  // Buffers the groupIds for each row. Reused across addInput callings.
```
mbasmanova: callings -> calls

velox/core/QueryConfig.h
HolyLow: In this case, are we supposed to name the WindowBuild as `PartitionedSortWindowBuild`?

mbasmanova: perhaps, SubPartitionedSortWindowBuild

velox/core/QueryConfig.h
HolyLow: done.

velox/core/QueryConfig.h
HolyLow: done.

velox/exec/Window.cpp
HolyLow: Refactored as subPartitions.

velox/exec/Window.cpp
HolyLow: Refactored to SubPartitionedSortWindowBuild.

velox/exec/Window.cpp
HolyLow: done.

velox/exec/Window.cpp
HolyLow: done.

velox/exec/SortWindowBuild.h
HolyLow: Reverted as suggested.

velox/exec/SortWindowBuild.h
HolyLow: Refactored to addDecodedInputRow. The method is called by SubPartitionedSortWindowBuild, so I made it public. We could also make SubPartitionedSortWindowBuild a friend class of SortWindowBuild, but I am not sure which is better, whether to make the method public or to make the class friend.

velox/exec/SortWindowBuild.h
HolyLow: Same as above problem about whether making the method public or making the class a friend.

velox/exec/RegionSortWindowBuild.h
HolyLow: Done. Yes, you are right that the vector's size == numSubPartitions_.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: done.

velox/exec/RegionSortWindowBuild.h
HolyLow: done.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: `data_` is inited in the constructor of WindowBuild. For SubPartitionedSortWindowBuild, the `data_` holds no data, and I explicitly mark it as null to avoid any possible misreading.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: done.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: Yes, you are right. Eliminated the redundant parameter already.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: done.

velox/exec/SortWindowBuild.cpp
HolyLow: We do could refactor `addInput` to use this new API, but for performance issue I suggest not to do this. In `addInput` the data is added in columnar manner, which might have better performance in multi-column cases, while in `addDecodedRow` the data could only be added in row manner. I am not sure if it would hurt performance to change the original columnar-mannered `addInput` to use row-mannered `addDecodedRow`. Looking forward to your thoughts, thanks a lot!

velox/exec/RegionSortWindowBuild.cpp
HolyLow: done.

velox/exec/tests/WindowTest.cpp
HolyLow: If we have skewed sub partitions, we might not be able to achieve performance and memory gain. But we would not have any correctness issue or have worse performance, because in this case we actually kind of fallback to the non-sub-partition algorithm. 
The memory consumption might be slightly larger because of the empty sub partitions, but the increase should be minor. The computation overhead might come from the partition hashing computation, but it is O(N) and would not affect the performance of O(NlogN) sorting.
So in my opinion there is actually no need to protect from this scenario. Besides, it is not easy to really "protect" from this issue.

velox/exec/tests/WindowTest.cpp
HolyLow: This orderBy is indeed redundant. Removed as suggested, thanks.

velox/exec/tests/WindowTest.cpp
HolyLow: assertResults is what I need here. I misunderstood its semantic, and now I get it right. Thanks a lot.

velox/exec/MultiGroupSortWindowBuild.h
HolyLow: done.

velox/exec/RegionSortWindowBuild.h
HolyLow: done.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: Yes, you are right that it is sufficient to just picking 1-2 partition keys when partitions are coarse. But if the first two partition keys are not coarse enough, we would end up with skewed sub partitions. Besides, computing the hash for all sub partition keys would not be that painful, especially compared with encoding all the keys into RowContainer. So I tend to directly use all the partition keys to avoid pre-mature optimization.
For now I have no idea on simple modulo function you proposed here. It would be greatly appreciated if you could share your thoughts.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: Good idea. Thanks.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: In theory we could achieve this, but for now the WindowBuild's `spill()` interface doesn't indicate how many bytes we want to spill. So for simplicity we spill all the data here just as SortWindowBuild here. This could be optimized if we want to spill more precisely.

velox/core/QueryConfig.h
HolyLow: Refactored as suggested.

velox/exec/SortWindowBuild.h
HolyLow: done.

velox/exec/MultiGroupSortWindowBuild.h
HolyLow: done.

velox/exec/SortWindowBuild.cpp
HolyLow: We do could refactor `addInput` to use this new API, but for performance issue I suggest not to do this. In `addInput` the data is added in columnar manner, which might have better performance in multi-column cases, while in `addDecodedRow` the data could only be added in row manner. I am not sure if it would hurt performance to change the original columnar-mannered addInput to use row-mannered `addDecodedRow`. Looking forward to your thoughts, thanks a lot!

HolyLow:
> These results are very long and not readable. Would you move these into a separate Google Sheet or Gist and provide a link along with an easy-to-understand summary?

> Would you also provide guidance on how to choose the number of regions?

@mbasmanova Would do that, but might require more time. I'll let you know when this is done.

> Testing appears to be light. Why do you believe it is sufficient? 

@mbasmanova The tests in `WindowTest.cpp` contains tests for all kinds of WindowBuild, such as SortWindowBuild, RowsStreamingWindowBuild, PartitionStreamingWindowBuild, etc. So I think PrePartitionedWindowBuild is just another WindowBuild and could be tested in a similar way.

> I would expect that we simply change existing window tests to run with and without extra partitioning.

I had the same thought as you at the beginning, but it seems that the tests in `WindowTest.cpp` is not all about SortWindowBuild. Other kinds of WindowBuilds also have a large portion of tests, and for these tests it is not proper to test SubPartitionedSortWindowBuild alongside them.

> I would also expect changing WindowFuzzer to use regions in some of the runs.

@mbasmanova This totally makes sense, but I need some time to dig into it as I am not familiar with the fuzzer module. 

@mbasmanova @aditi-pandit Thanks for your comments, refactored as suggested. Any further guidance or suggestions would be greatly appreciated.

velox/exec/SortWindowBuild.cpp
mbasmanova: > In addInput the data is added in columnar manner

Why do you think so? I see that the data is added one row at a time, not one column at a time. Am I missing something?

velox/exec/RegionSortWindowBuild.cpp
mbasmanova: > WindowBuild's spill() interface doesn't indicate how many bytes we want to spill

Would you create a GitHub issue to explain why it might be helpful to extend this interface? Let's do this in a follow-up if this is useful.

CC: @tanjialiang @xiaoxmeng

velox/exec/tests/WindowTest.cpp
mbasmanova: When you start writing the guildelines for how to choose the region count, consider, discussing this topic there.

velox/exec/SortWindowBuild.cpp
HolyLow: You are right that both `addInput` and `addDecodedRow` are adding data in row manner. I misunderstood the code :-( 
Refactored as suggested, thanks.

velox/exec/SortWindowBuild.cpp
HolyLow: I've made a mistake here. Both addInput and addDecodedRow are adding data in row manner. I misunderstood the code :-(
Refactored as suggested, thanks.

velox/core/QueryConfig.h
```diff
@@ -490,6 +490,12 @@ class QueryConfig {
   static constexpr const char* kDriverCpuTimeSliceLimitMs =
       "driver_cpu_time_slice_limit_ms";
 
+  /// Window operator can be configured to sub-divide window partitions on each
+  /// thread of execution into groups of partitions for sequential processing.
+  /// This setting specifies how many sub-partitions to create for each thread.
+  static constexpr const char* kWindowNumSubPartitions =
```
mbasmanova, line 496: make sure to update documentation at https://facebookincubator.github.io/velox/configs.html

mbasmanova:
@kagamiori @aditi-pandit Would you like to provide some pointers / guidance on how to extend the fuzzer?

velox/exec/RegionSortWindowBuild.cpp
tanjialiang: @HolyLow  We can just add target bytes to spill() in WindowBuild, if it is not hard to partially spill (I assume not as you basically spill some regions and keep other regions in memory). If it needs more changes and plumbings then we can create a followup github issue for this.

HolyLow:
@mbasmanova benchmarking results are recorded in a google sheet as suggested.
Benchmarking result of WindowRegionSortBenchmark, run on Intel(R) Xeon(R) 6982P-C (up to 3900MHz), ubuntu 22.04:
- Raw result output is in https://docs.google.com/document/d/1w04GHuEO3WsKDNCNiLAPxjlGYaJRL06OCNn4YAF3tZo/edit?usp=sharing
- Collected data and analysis is in https://docs.google.com/spreadsheets/d/1lI3CgTcAoQCeHO0KNIVf3ZIOWpcRRRJRD2vG9hq38pA/edit?usp=sharing
According to the results and analysis, the proper numSubPartitions would be 16~32 to achieve both performance improvement (6%~17%) and memory saving (20%~30% memory decrease) across all kinds of cases.

Any further suggestion is greatly appreciated.

velox/exec/RegionSortWindowBuild.cpp
HolyLow: @tanjialiang Thanks for your advice. I've opened a new issue for the extending the spill interface of WindowBuild, see https://github.com/facebookincubator/velox/issues/15452. I think we could iterate under the new issue rather than this one, as the feature would require more tests and might take some time. Besides, it is also more preferred to keep each issue focused on one feature.

HolyLow:
> Do you think you can follow-up to extend the Window Fuzzer to cover the new functionality?

@mbasmanova Sure, I've opened a new issue in https://github.com/facebookincubator/velox/issues/15451. We could extend the fuzzer under the new issue.

PR #15105, opened 2025-10-09
refactor: Move ioStats and fileReadOps to FileStorageContext struct

vhsu14:
Summary:
X-link: https://github.com/facebookincubator/nimble/pull/276

Refactored read operation parameters to use a `ReadOptions` struct instead of separate `ioStats` and `fileReadOps` parameters. This addresses the problem where adding new parameters to read functions requires updating all implementations across the codebase.

Differential Revision: D84112628

### DIFF
--- a/velox/common/file/File.h
+++ b/velox/common/file/File.h
@@ -48,6 +48,21 @@
 
 namespace facebook::velox {
 
+struct FileStorageContext {
+  /// Stats for IO operations
+  filesystems::File::IoStats* ioStats{nullptr};
+
+  /// Options for file read operations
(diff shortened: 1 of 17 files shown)


velox/common/file/File.h
```diff
@@ -48,50 +48,51 @@
 
 namespace facebook::velox {
 
+struct ReadOptions {
```
Yuhta: This is not just options (input parameters) but including both options and statistics.  A better name would be `FileStorageContext`.  There is also `ReaderOptions` in `dwio` directory but that's true options and it's for file formats, so we emphasize storage (not format) and context (not option) here.

PR #15110, opened 2025-10-10
fix: Error propagation in element_at UDF

bikramSingh91:
Summary:
Currently the UDF statically caches error messages to be re-used
whenever it needs to throw. However, since Velox error messages
add additional context when generated, the cached exception also
ends up re-using the message which contains the expression and
the top level expression details of the very first error produced
since process start. The original intention of that caching was
to mitigate perf impact when running under try where exceptions
are regularly expected. However, since we now have a mechanism
to handle that using Status objects, we employ that mechanism
to the UDF to fix the aforementioned issue.

Also updated the error messages to provide more details.
Updated unit tests and fixed a copy-paste error in one of the unit
tests which was testing the wrong input.

Differential Revision: D84308956

### DIFF
--- a/velox/expression/EvalCtx.cpp
+++ b/velox/expression/EvalCtx.cpp
@@ -192,6 +192,25 @@ void EvalCtx::setStatus(vector_size_t index, const Status& status) {
   }
 }
 
+void EvalCtx::setStatuses(const SelectivityVector& rows, const Status& status) {
+  VELOX_CHECK(!status.ok(), "Status must be an error");
+  if (status.isUserError()) {
+    if (throwOnError_) {
+      VELOX_USER_FAIL(status.message());
+    }
+
+    if (captureErrorDetails_) {
+      auto veloxException = toVeloxUserError(status.message());
+      rows.applyToSelected(
+          [&](auto row) { addError(row, veloxException, errors_); });
+    } else {
+      rows.applyToSelected([&](auto row) { addError(row, errors_); });
+    }
+  } else {
+    VELOX_FAIL(status.message());
+  }
+}
+
 void EvalCtx::setError(
     vector_size_t index,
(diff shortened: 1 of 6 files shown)


velox/functions/lib/SubscriptUtil.h
```diff
@@ -336,17 +334,28 @@ class SubscriptImpl : public exec::Subscript {
...
       // If index is invalid, capture the error and mark all rows as failed.
       bool isZeroSubscriptError = false;
+      const I originalIndex = decodedIndices->valueAt<I>(0);
       const auto adjustedIndex =
-          adjustIndex(decodedIndices->valueAt<I>(0), isZeroSubscriptError);
+          adjustIndex(originalIndex, isZeroSubscriptError);
       if (isZeroSubscriptError) {
-        context.setErrors(rows, detail::zeroSubscriptError());
+        try {
+          VELOX_USER_FAIL(detail::kZeroSubscriptErrorMsg);
```
mbasmanova: This is going to bring back the performance issue, no? Can we use context.captureErrorDetails + setStatus or similar?

bikramSingh91: This path is only for when the input is a constant, so the exception would only be generated at most once for all rows (the whole batch), instead of being generated for every row which would cause the perf issue.

I dont really have a strong preference, so lemme add a utility method like setError but for status instead to address this

velox/expression/EvalCtx.cpp
```diff
@@ -192,6 +192,25 @@ void EvalCtx::setStatus(vector_size_t index, const Status& status) {
   }
 }
 
+void EvalCtx::setStatuses(const SelectivityVector& rows, const Status& status) {
```
Yuhta, line 195: Nice, new functions should prefer this over throwing exceptions directly

PR #15111, opened 2025-10-10
feat(dwio): Enable row size tracking for metalake separately

HuamengJiang:
Summary:
Due to extra integration uncertainty in the metalake path and some previous QB signals, we decide to separately control row size tracking for metalake with the session property.

Instead of adding an additional session property, we decide to extend the current one. However, due to changing the session property type breaking backward compatibility, we still ended up introducing a new query config. We will delete the deprecated bool session property as the 3rd diff in the stack.

Differential Revision: D84228406

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -716,17 +716,36 @@ class QueryConfig {
   /// username.
   static constexpr const char* kClientTags = "client_tags";
 
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
   /// Enable (reader) row size tracker as a fallback to file level row size
   /// estimates.
   static constexpr const char* kRowSizeTrackingEnabled =
       "row_size_tracking_enabled";
+#endif
+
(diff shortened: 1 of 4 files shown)


Yuhta (approved):

velox/core/QueryConfig.h
```diff
@@ -716,18 +716,37 @@ class QueryConfig {
...
     return get<bool>(kSelectiveNimbleReaderEnabled, false);
   }
 
+#ifdef VELOX_ENABLE_BACKWARD_COMPATIBILITY
   bool rowSizeTrackingEnabled() const {
     return get<bool>(kRowSizeTrackingEnabled, true);
   }
+#endif
+
+  RowSizeTrackingMode rowSizeTrackingMode() const {
```
aditi-pandit, line 746: @HuamengJiang : It would be great to expose std::string getters and setters as well for this value. 

It is exposed as a session property by Prestissimo and it will be better for engines to not have to reimplement the string conversions for this enum.

aditi-pandit (commented):
Thanks for this PR.

PR #15112, opened 2025-10-10
fix: The processing logic of the testStringToMap method of StringToMapTest

weixiuli:
The assertEqualVectors function of StringToMapTest  is  used to  compare two vectors for equality.

 * This function checks if the size, type, and values of the `expected` and
 * `actual` vectors are identical. If any mismatch is found, an assertion
 * failure is triggered.
 *
 * @param expected the vector containing the expected values.
 * @param actual the vector to compare against the expected vector.
 
So we should match the expected and actual with the parameters of the function.


jinchengchenghh:
The first argument should be expected, but this is not restrict checked in Velox, please check it in the new code, we don't expect to refactor it.

PR #15085, opened 2025-10-08
refactor: Improve Buffer

MBkkt:
In this PR I refactored buffer to
* Avoid extra Releaser copy
* Use no_unique_address for empty Releaser
* Avoid virtual isView
* Added missed override for AlignedBuffer
* Add noexcept for some simple methods
* Made some readability improvements: default dtor, nested namespaces, C++17 usings from stdlib

PR #15120, opened 2025-10-10
feat: Optimize the buffer isView check

xiaoxmeng:
Summary: Buffer isView is virtual method which is expensive and show at least 10% of cosco shuffle read operator which mostly come from compact row deserializer. This diff makes this check using a buffer type

Differential Revision: D84329308

### DIFF
--- a/velox/buffer/Buffer.h
+++ b/velox/buffer/Buffer.h
@@ -19,6 +19,8 @@
 #include <iomanip>
 
 #include <boost/intrusive_ptr.hpp>
+#include <fmt/format.h>
+#include <utility>
 #include "velox/common/base/BitUtil.h"
 #include "velox/common/base/CheckedArithmetic.h"
 #include "velox/common/base/Exceptions.h"
@@ -61,7 +63,26 @@ class Buffer {
   static inline constexpr bool is_pod_like_v =
       std::is_trivially_destructible_v<T> && std::is_trivially_copyable_v<T>;
 
-  virtual ~Buffer() {}
+  virtual ~Buffer() = default;
+
+  inline static constexpr uint8_t kPODBit = 0;
+  inline static constexpr uint8_t kPODMask = 1 << kPODBit;
+  inline static constexpr uint8_t kViewBit = 1;
+  inline static constexpr uint8_t kViewMask = 1 << kViewBit;
+  static_assert(kPODBit != kViewBit);
+
+  enum class Type : uint8_t {
+    kNonPOD = 0 << kPODBit | 0 << kViewBit,
+    kPOD = 1 << kPODBit | 0 << kViewBit,
+    kNonPODView = 0 << kPODBit | 1 << kViewBit,
+    kPODView = 1 << kPODBit | 1 << kViewBit,
+  };
+
+  static std::string typeString(Type type);
+
+  Type type() const {
+    return type_;
+  }
 
   void addRef() {
     referenceCount_.fetch_add(1, std::memory_order_acq_rel);
@@ -86,7 +107,7 @@ class Buffer {
   const T* as() const {
     // We can't check actual types, but we can sanity-check POD/non-POD
     // conversion. `void` is special as it's used in type-erased contexts
-    VELOX_DCHECK((std::is_same_v<T, void>) || podType_ == is_pod_like_v<T>);
+    VELOX_DCHECK((std::is_same_v<T, void>) || isPOD() == is_pod_like_v<T>);
     return reinterpret_cast<const T*>(data_);
   }
 
@@ -102,7 +123,7 @@ class Buffer {
     VELOX_CHECK(!isView());
     // We can't check actual types, but we can sanity-check POD/non-POD
     // conversion. `void` is special as it's used in type-erased contexts
-    VELOX_DCHECK((std::is_same_v<T, void>) || podType_ == is_pod_like_v<T>);
+    VELOX_DCHECK((std::is_same_v<T, void>) || isPOD() == is_pod_like_v<T>);
     return reinterpret_cast<T*>(data_);
   }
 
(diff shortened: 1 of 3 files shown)


velox/buffer/Buffer.h
```diff
@@ -800,5 +809,9 @@ class BufferView : public Buffer {
   Releaser const releaser_;
 };
 
+// Template specialization for fast BufferView type checking
+// Note: This is implemented inline within BufferView for template
+// specialization
```
duanmeng: move these comments above `BufferView`?

xiaoxmeng:
> @xiaoxmeng Hello, I already did this here: #15085 I also think layout in your PR is worse, because it increases sizeof then it possible to avoid this. I also think we don't need type, bool is enough

@MBkkt Hi, I didn't notice your change. I am optimizing our internal shuffle workload. I might want to keep using Type as it can be used for other use case as well. The size is not changed and I did update to move both types together for layout. They both supposed to be consts. I can help merge the refactor one.

velox/buffer/Buffer.h
```diff
@@ -142,8 +148,8 @@ class Buffer {
     return !isView() && unique();
   }
 
-  virtual bool isView() const {
-    return false;
+  inline bool isView() const {
+    return type_ == Type::kView;
```
MBkkt: check with != 0/== 0 (bool check in this case) works better than check with == 2.

Why do you want to have separate Type?

Even if it's required for some reason, better is have `kView = 0`

xiaoxmeng: This help indicating different type of buffers and the same for view check. This just remove the asMutable check cost from the cpu profiling

MBkkt:
@xiaoxmeng Understand, thanks for explaining.

I think your changes almost same now (in terms of isView) but have type behavior.
And it's better to make kView zero constant.

velox/buffer/Buffer.h
```diff
@@ -142,8 +148,8 @@ class Buffer {
     return !isView() && unique();
   }
 
-  virtual bool isView() const {
-    return false;
+  inline bool isView() const {
```
MBkkt:
```suggestion
  bool isView() const {
```
it's already inline (because written in class definition), this inline do nothing

velox/buffer/Buffer.h
```diff
@@ -61,7 +61,13 @@ class Buffer {
   static inline constexpr bool is_pod_like_v =
       std::is_trivially_destructible_v<T> && std::is_trivially_copyable_v<T>;
 
-  virtual ~Buffer() {}
+  virtual ~Buffer() = default;
+
+  enum class Type : uint8_t { kAligned = 0, kNonPODAligned = 1, kView = 2 };
```
MBkkt:
```suggestion
  enum class Type : uint8_t { kView = 0, kAligned = 1, kNonPODAligned = 2, };
```

velox/buffer/Buffer.h
```diff
@@ -264,19 +271,22 @@ class Buffer {
...
-  uint64_t capacity_ = 0;
-  std::atomic<int32_t> referenceCount_;
-  bool podType_ = true;
+
+  uint64_t size_{0};
+  uint64_t capacity_{0};
+  std::atomic_int32_t referenceCount_;
+
+  Type type_{Type::kAligned};
+  bool podType_{true};
```
MBkkt:
```suggestion
  const Type type_{Type::kAligned};
  const bool podType_{true};
```

xiaoxmeng: Ok. That's why I want to put const members first initially. Let me change to const and keep it in the middle.

velox/buffer/Buffer.h
```diff
@@ -142,8 +156,12 @@ class Buffer {
...
 
-  virtual bool isView() const {
-    return false;
+  bool isView() const {
+    return type_ == Type::kNonPODView || type_ == Type::kPODView;
+  }
+
+  bool isPOD() const {
+    return type_ == Type::kPOD || type_ == Type::kPODView;
   }
```
MBkkt, line 172: But such checks are more expensive.
Maybe at least `type_ >= Type::kPODView` for `isView()`?
Or make View/Pod bit flags in type_?

velox/buffer/Buffer.h
```diff
@@ -256,27 +274,28 @@ class Buffer {
...
-  uint64_t size_ = 0;
-  uint64_t capacity_ = 0;
-  std::atomic<int32_t> referenceCount_;
-  bool podType_ = true;
+
+  uint64_t size_{0};
+  uint64_t capacity_{0};
+  std::atomic_int32_t referenceCount_{0};
+
+  const Type type_{Type::kPOD};
```
MBkkt, line 301: Here 3 bytes are available after type_

velox/buffer/Buffer.h
MBkkt, line 172:
```cpp
  enum class Type : uint8_t {
    kNone = 0,
    kIsPOD = 1 << 0,
    kIsView = 1 << 1,
    kNonPOD = kNone,
    kPOD = kNone | kIsPOD,
    kNonPODView = kNone | kIsView,
    kPODView = kNone | kIsPOD | kIsView,
  };

  bool isView() const {
    return type_ & Type::kIsView;
  }

  bool isPOD() const {
    return type_ & Type::kIsPOD;
  }
```

MBkkt: But I still think two bool flags are better :(

MBkkt:
@xiaoxmeng Thanks!

velox/buffer/Buffer.h
```diff
@@ -245,7 +270,7 @@ class Buffer {
   virtual void copyFrom(const Buffer* other, size_t bytes) {
     VELOX_CHECK(!isView());
```
MBkkt, line 271: can you replace this line with podType == POD? 
It's kind of duplicated check

PR #15124, opened 2025-10-10
fix: Replace boost::regex with RE2 in parse_duration function

abhinavmuk04:
Summary:
Some implementations of functions in Velox rely on boost::regex as regular expression engine. That is not recommended. RE2 is the preferred engine.

boost relies on backtracking and it can lead to significantly slower performance compared to RE2. RE2 has O(n) complexity for a given string of length n, as it uses finite-state machines.

This change replaces boost::regex with RE2 in the ParseDurationFunction in DateTimeFunctions.h. The implementation now uses RE2::FullMatch instead of boost::regex_search, and directly captures the value and unit strings using RE2's API, simplifying the code by eliminating the need for boost::smatch.

Differential Revision: D84391289

### DIFF
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -16,10 +16,9 @@
 #pragma once
 
 #define XXH_INLINE_ALL
-#include <boost/regex.hpp>
+#include <re2/re2.h>
 #include <xxhash.h>
 #include <string_view>
-#include "velox/expression/DecodedArgs.h"
 #include "velox/expression/VectorFunction.h"
 #include "velox/functions/lib/DateTimeFormatter.h"
 #include "velox/functions/lib/TimeUtils.h"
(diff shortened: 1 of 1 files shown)


MBkkt:
@abhinavmuk04 Hi, nice changes!

I have some suggestions:
1. We can use raw-string literals to avoid complicating regex pattern
2. We can avoid `std::string` copy for matched strings (`valueStr` and `unit`)
3. Better to use `from_chars`, because it works faster and doesn't throw exception, instead return result 

```cpp
template <typename T>
struct ParseDurationFunction {

  std::unique_ptr<RE2> durationRegex_;

  FOLLY_ALWAYS_INLINE void initialize(
      const std::vector<TypePtr>& /*inputTypes*/,
      const core::QueryConfig& /*config*/,
      const arg_type<Varchar>* /*amountUnit*/) {
    durationRegex_ =
        std::make_unique<RE2>(R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)");
  }

  FOLLY_ALWAYS_INLINE void call(
      out_type<IntervalDayTime>& result,
    std::string_view valueStr;
    std::string_view unit;
    if (!RE2::FullMatch(
            std::string_view{amountUnit}, *durationRegex_, &valueStr, &unit)) {
      VELOX_USER_FAIL(
          "Input duration is not a valid data duration string: {}", amountUnit);
    }

    double value{};
    auto [_, error] = std::from_chars(
        valueStr.data(), valueStr.data() + valueStr.size(), value);
    if (error != std::errc{}) {
      VELOX_USER_FAIL(
          "Input duration value is not a valid number: {}", valueStr);
    }

    result = valueOfTimeUnitToMillis(value, unit);
  }
};
```

MBkkt:
@abhinavmuk04 will you check https://github.com/facebookincubator/velox/pull/15134 ?

abhinavmuk04:
> @abhinavmuk04 will you check #15134 ?

@MBkkt I see in this PR you seem to be handling the parse_duration function. Is this something you are working on? If so, I can let go of this PR

abhinavmuk04:
> > @abhinavmuk04 will you check #15134 ?
> 
> @MBkkt I see in this PR you seem to be handling the parse_duration function. Is this something you are working on? If so, I can let go of this PR

@MBkkt wanted to follow up on this

MBkkt:
@abhinavmuk04 Not really. If you will merge your PR, I will just rebase. If you don't merge this PR, I will wait for review for mine, to merge it and get rid of `boost::regex`.

Some explanation:
 
I just noticed your PR and thought about it will be nice to do 3 things:
1. I did some review suggestions to your PR (avoid string copies, use faster "to double conversion", single time automaton construction, etc)
2. Removed second usage of boost::regex
3. Removed boost::regex from CMake

kevinwilfong:
It looks like you had them in a previous commit (the one I was initially looking at) and accidentally reverted them.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1813,43 +1813,40 @@ template <typename T>
...
-      double value = std::stod(match[1].str());
-      std::string unit = match[2].str();
-      result = valueOfTimeUnitToMillis(value, unit);
-    } catch (std::out_of_range&) {
+
+    double value{};
+    auto [_, error] = std::from_chars(
+        valueStr.data(), valueStr.data() + valueStr.size(), value);
+    if (error == std::errc::result_out_of_range) {
       VELOX_USER_FAIL(
```
kevinwilfong, line 1957: It looks like there's a build failure, due to these two VELOX_USER_FAIL calls, you probably need to wrap valueStr in a string_view when you pass it here.

abhinavmuk04:
@MBkkt "Better to use from_chars, because it works faster and doesn't throw exception, instead return result" I don't think I can implement std::from_chars` - Cannot implement due to platform compatibility

std::from_chars for floating-point types is not available on macOS with libc++ (Clang's C++ standard library). It works on Linux but Velox needs cross-platform compatibility.

So it seems like we will keep this the same

MBkkt:
@abhinavmuk04 You're right, there's two options
1) use for this code from my PR (fast_float + some other fixes)
2) You can just revert + forget about this, I will try to merge PR with improvements  in a separate commit

abhinavmuk04:
> @abhinavmuk04 You're right, there's two options
> 
> 1. use for this code from my PR (fast_float + some other fixes)
> 2. You can just revert + forget about this, I will try to merge PR with improvements  in a separate commit

@MBkkt @kevinwilfong Cool, we can keep this implementation for now. Let me know if you make a separate commit, I can help with review.

PR #15134, opened 2025-10-11
refactor: Replace boost regex with re2

MBkkt:
re2 works better, we will also decrease binary size and count of dependencies

First commit inspired by this PR https://github.com/facebookincubator/velox/pull/15124

### DIFF
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -1927,8 +1927,10 @@ struct ParseDurationFunction {
   FOLLY_ALWAYS_INLINE void call(
       out_type<IntervalDayTime>& result,
       const arg_type<Varchar>& amountUnit) {
-    static const LazyRE2 kDurationRegex(
-        R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)");
+    static const LazyRE2 kDurationRegex{
+        .pattern_ = R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)",
+        .options_ = {},
+    };
     // TODO: Remove re2::StringPiece != std::string_view hacks.
     // It's needed because for some systems in CI,
     // re2 and abseil libraries are old.


velox/functions/sparksql/GetJsonObject.h
```diff
@@ -190,13 +188,9 @@ struct GetJsonObjectFunction {
...
-      case simdjson::ondemand::json_type::object: {
-        // For nested case, e.g., for "{"my": {"hello": 10}}", "$.my" will
-        // return an object type.
-        ss << rawResult;
-        result.append(ss.str());
-        return true;
-      }
+      // For nested case, e.g., for "{"my": {"hello": 10}}",
+      // "$.my" will return an object type.
+      case simdjson::ondemand::json_type::object:
```
rui-mo, line 251: cc: @PHILO-HE Would you like to take a look at the change to get_json_object? Thanks.

MBkkt: Just in case, change here, it was same code in two cases

philo-he: The changes make sense. Thanks.

MBkkt:
@kevinwilfong Can you review this changes, please?

majetideepak:
I will let Meta folks chime in as well.
CC: @kgpai

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1921,50 +1922,36 @@ template <typename T>
...
-        std::make_unique<RE2>(R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)");
-  }
-
   FOLLY_ALWAYS_INLINE void call(
       out_type<IntervalDayTime>& result,
       const arg_type<Varchar>& amountUnit) {
+    static const LazyRE2 kDurationRegex{
+        R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)"};
+    // TODO: Remove re2::StringPiece != std::string_view hacks.
+    // It's needed because system is old, update please.
```
kevinwilfong: Could you be more specific either here in the the PR comments about what "the system" is, are you referring to CI?

kevinwilfong:
@MBkkt  it looks like there are conflicts, could you rebase?

MBkkt:
@kevinwilfong rebased

velox/functions/prestosql/DateTimeFunctions.h
MBkkt: I updated comment.

Yes, in some system in CI, system abseil was build with `absl::string_view != std::string_view` and re2 was build with `re2::StringPiece != absl::string_view`

It's old unsupported version of abseil, because it's true only for C++14, but abseil drops C++14 support already.

MBkkt:
@czentgr Yes, but It's because this PR exists too long.
Initially it was two places with boost regex, but after multiple updates from main and zero comments to PR...

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1921,50 +1922,37 @@ template <typename T>
...
-      const arg_type<Varchar>* /*amountUnit*/) {
-    durationRegex_ =
-        std::make_unique<RE2>(R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)");
-  }
-
   FOLLY_ALWAYS_INLINE void call(
       out_type<IntervalDayTime>& result,
       const arg_type<Varchar>& amountUnit) {
+    static const LazyRE2 kDurationRegex{
+        R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)"};
```
kevinwilfong: Hi @MBkkt, it looks like our internal build is a little stricter than what runs here.

-Wmissing-field-initializers is showing that the options_ field isn't being initialized here, combined with -Werror this is causing the build to fail and I'm unable to merge

could you either change this to invoke the constructor rather than the list initializer {} -> () or add an initializer for options_ .options_={}

MBkkt: Hi @kevinwilfong 

1. Can you change in future (in a separate PR) to make github CI same as internal in terms of warnings?
2. I'm not sure it will help, but lets try this: (same as `.options_ =`, but without explicit field mention) https://github.com/facebookincubator/velox/pull/15134/commits/f09e68f2e9f76b4a09d9f1b78c06cd068c53dd9d

```cpp
// Silence warnings about missing initializers for members of LazyRE2.
#if defined(__GNUC__)
#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
#endif

// Helper for writing global or static RE2s safely.
// Write
//     static LazyRE2 re = {".*"};
// and then use *re instead of writing
//     static RE2 re(".*");
// The former is more careful about multithreaded
// situations than the latter.
//
// N.B. This class never deletes the RE2 object that
// it constructs: that's a feature, so that it can be used
// for global and function static variables.
class LazyRE2 {
 private:
  struct NoArg {};

 public:
  typedef RE2 element_type;  // support std::pointer_traits

  // Constructor omitted to preserve braced initialization in C++98.

  // Pretend to be a pointer to Type (never NULL due to on-demand creation):
  RE2& operator*() const { return *get(); }
  RE2* operator->() const { return get(); }

  // Named accessor/initializer:
  RE2* get() const {
    absl::call_once(once_, &LazyRE2::Init, this);
    return ptr_;
  }

  // All data fields must be public to support {"foo"} initialization.
  const char* pattern_;
  RE2::CannedOptions options_;
  NoArg barrier_against_excess_initializers_;

  mutable RE2* ptr_;
  mutable absl::once_flag once_;

 private:
  static void Init(const LazyRE2* lazy_re2) {
    lazy_re2->ptr_ = new RE2(lazy_re2->pattern_, lazy_re2->options_);
  }

  void operator=(const LazyRE2&);  // disallowed
};
```

kevinwilfong: It looks like that doesn't work (it complains about barrier_against_excess_initializers_ in that case)

Explicitly listing the fields does
{.pattern_ = R"(^\s*(\d+(?:\.\d+)?)\s*([a-zA-Z]+)\s*$)", .options_ = {}}

MBkkt: https://github.com/facebookincubator/velox/pull/15134/commits/155b0be41cc5c320020433a04655b9e64dd13766 something like this maybe can help with internal CI warning?

kevinwilfong: While that does work internally, unfortunately it looks like it breaks the Mac build here in OS

MBkkt: ok, https://github.com/facebookincubator/velox/pull/15134/commits/736cf4774ba57dcbaffb4c310c58083261c12d92

kevinwilfong: Thanks! Merging, sorry for the slog

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #define XXH_INLINE_ALL
+#include <fast_float/fast_float.h>
```
jinchengchenghh, line 19: Do you need to use velox_include_directories to include the fast-float header? If the environment does not have fast-float, the velox build will fail, I'm not sure why Velox does not build it from source. @MBkkt  CC @czentgr 
https://github.com/IBM/velox/actions/runs/18915929559/job/53999183287
```
FAILED: velox/buffer/CMakeFiles/velox.dir/__/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp.o 
/usr/bin/ccache /opt/rh/devtoolset-11/root/usr/bin/c++ -DAWS_MQTT_WITH_WEBSOCKETS -DAWS_SDK_VERSION_MAJOR=1 -DAWS_SDK_VERSION_MINOR=11 -DAWS_SDK_VERSION_PATCH=285 -DAWS_USE_EPOLL -DAZ_RTTI -DCURL_STATICLIB -DGEOS_INLINE -DGFLAGS_IS_A_DLL=0 -DNDEBUG -DSIMDJSON_EXPERIMENTAL_ALLOW_INCOMPLETE_JSON -DSIMDJSON_THREADS_ENABLED=1 -DUSE_UNSTABLE_GEOS_CPP_API -DVELOX_DISABLE_GOOGLETEST -DVELOX_ENABLE_ABFS -DVELOX_ENABLE_COMPRESSION_LZ4 -DVELOX_ENABLE_GCS -DVELOX_ENABLE_GEO -DVELOX_ENABLE_HDFS -DVELOX_ENABLE_PARQUET -DVELOX_ENABLE_S3 -I/work/. -I/work/velox/external/xxhash -I/work/_build/release -I/work/velox/tpch/gen/dbgen/include -I/work/velox/tpcds/gen/dsdgen/include -I/work/_build/release/_deps/simdjson-src/include -isystem /work/gluten/dev/vcpkg/vcpkg_installed/x64-linux-avx/include -isystem /work/velox -isystem /work/velox/external -isystem /work/gluten/dev/vcpkg/vcpkg_installed/x64-linux-avx/lib/pkgconfig/../../include -isystem /work/gluten/dev/vcpkg/vcpkg_installed/x64-linux-avx/include/geos -Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers -Wno-unknown-warning-option -mavx2 -mfma -mavx -mf16c -mlzcnt -mbmi2 -D USE_VELOX_COMMON_BASE -D HAS_UNCAUGHT_EXCEPTIONS -DFOLLY_CFG_NO_COROUTINES -Wall -Wextra -Wno-unused        -Wno-unused-parameter        -Wno-sign-compare        -Wno-ignored-qualifiers        -Wno-implicit-fallthrough          -Wno-class-memaccess          -Wno-comment          -Wno-int-in-bool-context          -Wno-redundant-move          -Wno-array-bounds          -Wno-maybe-uninitialized          -Wno-unused-result          -Wno-format-overflow          -Wno-strict-aliasing -Werror -O3 -DNDEBUG -std=gnu++20 -fPIC -fdiagnostics-color=always -ffp-contract=off -DS2N_KYBER512R3_AVX2_BMI2 -DS2N_STACKTRACE -DS2N_CPUID_AVAILABLE -DS2N_FEATURES_AVAILABLE -fPIC -DS2N_FALL_THROUGH_SUPPORTED -DS2N___RESTRICT__SUPPORTED -DS2N_MADVISE_SUPPORTED -DS2N_CLONE_SUPPORTED -MD -MT velox/buffer/CMakeFiles/velox.dir/__/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp.o -MF velox/buffer/CMakeFiles/velox.dir/__/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp.o.d -o velox/buffer/CMakeFiles/velox.dir/__/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp.o -c /work/velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
In file included from /work/velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp:18:
/work/./velox/functions/prestosql/DateTimeFunctions.h:19:10: fatal error: fast_float/fast_float.h: No such file or directory
   19 | #include <fast_float/fast_float.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

MBkkt: No I shouldn't, because library target already linked in cmake

jinchengchenghh: Maybe you need to call https://github.com/facebookincubator/velox/pull/12735/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR411 as this PR

MBkkt: I think you build folly in some specific way, because fast_float link provided from Folly::folly now I think.
Lets do this explicit

MBkkt: @jinchengchenghh https://github.com/facebookincubator/velox/pull/15322 does it help?

jinchengchenghh:
```
velox_set_source(fastfloat)
velox_resolve_dependency(fastfloat CONFIG REQUIRED)
```
This exists in folly build from resource, but if folly exists in the system folder, it will not call that.

So you may need to add it to Velox https://github.com/facebookincubator/velox/blob/main/CMakeLists.txt

jinchengchenghh: I don't think https://github.com/facebookincubator/velox/pull/15322 can work, we cannot find the header, not the library, and fast_float does not contains cpp file

> The fast_float library provides fast header-only implementations for the C++ from_chars functions

MBkkt: Ok

MBkkt: > I don't think https://github.com/facebookincubator/velox/pull/15322 can work, we cannot find the header, not the library, and fast_float does not contains cpp file

It's not correct to setup cmake header-only library with just include header.
For an example some macros can be missed

jinchengchenghh: This is copied from fast_float github project description, https://github.com/fastfloat/fast_float
> The fast_float library provides fast header-only implementations for the C++ from_chars functions

MBkkt: It's header only but doesn't mean you should just include needed directories

MBkkt: Btw in Velox CI exists setup with system folly, how is it different from IBM setup?

jinchengchenghh: IBM installs the folly by vcpkg, so the environment already has folly, but vcpkg does not support install fast-float version 8.0.2 https://github.com/apache/incubator-gluten/pull/10977, so fast-float does not exist in the environment. It does not call setup-script.sh

MBkkt: > so fast-float does not exist in the environment. It does not call setup-script.sh

So you don't have fast_float in vcpkg folly or how it works?

> but vcpkg does not support install fast-float version 8.0.2

Hmm, strange, on their site they're claim support 8.0.2
image

Maybe I'm missing something because I'm not working with C++ package managers and system libraries.

As an alternative solution we probably can update fast_float version? I don't think it's bad, but it's also affects folly

jinchengchenghh: Maybe because Gluten uses an old vcpkg version? I'm not sure for that.

But I think current https://github.com/facebookincubator/velox/pull/15322 can help, so we don't need to update fast_float version and care about Gluten vcpkg fast-float version.

MBkkt: Ok, lets try current approach, if it doesn't help, please ping me. And we will try something else

jinchengchenghh: Thanks for your in time fix!

Issue #15047, opened 2025-10-04
Run clang-tidy in CI


assignUser:
### Description

Currently it's only a manual hook as it needs cmake to have been run to generate compile_commands.json. So we likely want to run it as part of the adapters build to check all features.

Issue #15046, opened 2025-10-04
Ensure linter parity with Meta internal checks


assignUser:
### Description

Matching the internal linters makes it faster to merge PRs as no adjustments are needed when GH CI passes.

I think we are only lacking `clang-tidy` at the moment, which is currently a manual pre-commit-hook. But someone from Meta should compare against the internal setup 

cc @kgpai @pedroerp

pedroerp:
@assignUser wondering if you've seen this project:

https://github.com/platisd/clang-tidy-pr-comments

it's meant to expose clang-tidy output as PR comments.

assignUser:
<img width="1105" height="1031" alt="Image" src="https://github.com/user-attachments/assets/b9f9fc79-2e4f-4c5e-81a6-eadba19d4b3e" /> @pedroerp  https://github.com/facebookincubator/velox/pull/15284

pedroerp:
Love this! Thanks @assignUser 

Cc: @mbasmanova @kgpai @majetideepak

PR #15135, opened 2025-10-11
fix: Disable some noisy .clang-tidy checks

MBkkt:
Also see https://github.com/facebookincubator/axiom/pull/335

Disable checks that are not useful or requires change almost every code line.

### **altera-struct-pack-align**

Checks that packed data structures are used when cross-compiling to FPGAs. This check warns when a struct is not packed but could be packed for more efficient FPGA synthesis.

### **altera-unroll-loops**

Checks for opportunities to unroll loops to improve performance on FPGA targets. Suggests using pragma unroll or pragma unroll with a factor.

### **boost-use-ranges**

Suggests using Boost.Range algorithms instead of standard library algorithms where applicable for better performance and readability.

### **bugprone-easily-swappable-parameters**

Finds function declarations where parameters of the same type are adjacent, making it easy to accidentally swap them when calling the function. This can catch bugs where parameters are passed in the wrong order.

### **cppcoreguidelines-avoid-const-or-ref-data-members**

Warns about const or reference data members in classes, as they can make classes non-assignable and non-copyable, violating the C++ Core Guidelines' recommendations for regular types.

### **cppcoreguidelines-avoid-do-while**

Flags do-while loops as they can be error-prone compared to while or for loops, as recommended by the C++ Core Guidelines.

### **cppcoreguidelines-non-private-member-variables-in-classes**

Enforces that all data members in classes should be private, following object-oriented design principles from the C++ Core Guidelines.

### **cppcoreguidelines-pro-bounds-constant-array-index**

Flags array subscript operations that use non-constant indices, as they cannot be statically verified to be within bounds, violating the bounds safety guidelines.

### **llvm-qualified-auto**

Requires that `auto` variables have explicit qualifiers like `const` or `*` when their type is not immediately obvious from the initializer.

### **modernize-use-designated-initializers**

Suggests using C++20 designated initializers for aggregate initialization to make code more readable and maintainable.

### **performance-avoid-endl**

Warns about using `std::endl` instead of `'\n'` because `std::endl` flushes the stream buffer which can be slower, especially in loops.

### **performance-enum-size**

Suggests specifying explicit underlying types for enums to control their size and potentially improve performance and memory usage.

### **readability-function-cognitive-complexity**

Measures and flags functions that have high cognitive complexity (too many nested control structures, conditions, etc.) making them hard to understand and maintain.

### **readability-identifier-length**

Enforces minimum and maximum lengths for identifiers (variable names, function names, etc.) to improve code readability.

### **readability-math-missing-parentheses**

Suggests adding parentheses to mathematical expressions to clarify operator precedence and improve readability.

### **readability-qualified-auto**

Similar to `llvm-qualified-auto`, requires explicit qualifiers with `auto` to make the intended type more obvious to readers.


assignUser:
Your arguments sound sensible and I would assume that these are not run internally at meta (otherwise the errors would block merges) but we will have to cross check this with #15046 / #15047

MBkkt:
@assignUser In general I agree with that you said, but I also think in meta this check doesn't prevent merge now, at least from my perspective it looks like this

MBkkt:
But these checks are almost insane, whole vscode spammed by them in most velox files

mbasmanova:
CC: @pedroerp @kgpai

pedroerp:
> @assignUser In general I agree with that you said, but I also think in meta this check doesn't prevent merge now, at least from my perspective it looks like this

that's correct. The lint errors are just a warning internally.

.clang-tidy
```diff
@@ -1,20 +1,26 @@
 Checks: >
   *,
   -abseil-*,
+  -altera-struct-pack-align,
```
pedroerp, line 4: should this just be `-altera-*` ?

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
   -hicpp-special-member-functions,
   -hicpp-use-equals-default,
   -hicpp-vararg,
-  -hicpp-vararg,
   -llvm-header-guard,
   -llvm-include-order,
+  -llvm-qualified-auto,
```
pedroerp, line 36: I think this one is actually useful. I've seen many bugs in the past because these were not clear. Perhaps we should start enforcing this at least for new code?

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
   -modernize-avoid-c-arrays,
   -modernize-deprecated-headers,
+  -modernize-use-designated-initializers,
   -modernize-use-nodiscard,
   -modernize-use-trailing-return-type,
   -mpi-*,
   -objc-*,
   -openmp-*,
+  -performance-avoid-endl,
+  -performance-enum-size,
```
pedroerp, line 50: this also seems useful. I think we have this enabled internally

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
   -readability-avoid-const-params-in-decls,
   -readability-convert-member-functions-to-static,
+  -readability-function-cognitive-complexity,
```
pedroerp, line 53: we also have this one internally

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
+  -readability-identifier-length,
   -readability-implicit-bool-conversion,
   -readability-magic-numbers,
+  -readability-math-missing-parentheses,
```
pedroerp, line 57: this is also useful, I think.

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
+  -readability-qualified-auto,
```
pedroerp, line 58: that's the same as the llvm one above

.clang-tidy
MBkkt: Can we avoid this? It's just hightlight a lot of functions.
No-one never will really fix this, because it's completely harmless.

Some example

image

.clang-tidy
MBkkt: I don't think so, everyone knows about multiply/divide have higher priority.

https://clang.llvm.org/extra/clang-tidy/checks/readability/math-missing-parentheses.html

.clang-tidy
MBkkt: Yes, readability-qualified-auto vs llvm-qualified-auto are alises to same check.
Can we have issue about enable it?
Now it will be highlight too much code

.clang-tidy
MBkkt: > -performance-enum-size,

This one? I'm ok to enable it back

.clang-tidy
MBkkt: https://github.com/facebookincubator/velox/pull/15135/files#r2433480584

.clang-tidy
MBkkt: Sounds good to me

[altera-id-dependent-backward-branch](https://clang.llvm.org/extra/clang-tidy/checks/altera/id-dependent-backward-branch.html)
[altera-kernel-name-restriction](https://clang.llvm.org/extra/clang-tidy/checks/altera/kernel-name-restriction.html)
[altera-single-work-item-barrier](https://clang.llvm.org/extra/clang-tidy/checks/altera/single-work-item-barrier.html)
[altera-struct-pack-align](https://clang.llvm.org/extra/clang-tidy/checks/altera/struct-pack-align.html)
[altera-unroll-loops](https://clang.llvm.org/extra/clang-tidy/checks/altera/unroll-loops.html)

.clang-tidy
MBkkt: Also a lot of code written without care about this

MBkkt:
@pedroerp Main point of this PR for me personally is to actually get files in vscode with meaningful clang-tidy warnings.
So when I write code for velox, I can rely on this, instead of disable it completely.

I think some of this checks can be good for new code, but I think we need to cleanup old code first before enable them

MBkkt:
> I think some of this checks can be good for new code, but I think we need to cleanup old code first before enable them

It can be done partially automatically

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
   -llvm-include-order,
+  -llvm-qualified-auto,
   -llvmlibc-*,
   -misc-no-recursion,
-  -misc-no-recursion,
   -misc-non-private-member-variables-in-classes,
   -misc-unused-parameters,
   -modernize-avoid-c-arrays,
   -modernize-deprecated-headers,
+  -modernize-use-designated-initializers,
```
kgpai, line 43: Isnt this a nice thing to have , makes it very clear when initializing structs.

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
...
   -modernize-use-trailing-return-type,
   -mpi-*,
   -objc-*,
   -openmp-*,
+  -performance-avoid-endl,
+  -performance-enum-size,
   -readability-avoid-const-params-in-decls,
   -readability-convert-member-functions-to-static,
+  -readability-function-cognitive-complexity,
+  -readability-identifier-length,
```
kgpai, line 54: Also think this is a good thing to have , but I realize we have some big identifiers.

.clang-tidy
```diff
@@ -25,25 +31,31 @@ Checks: >
   -hicpp-special-member-functions,
   -hicpp-use-equals-default,
   -hicpp-vararg,
-  -hicpp-vararg,
   -llvm-header-guard,
   -llvm-include-order,
+  -llvm-qualified-auto,
   -llvmlibc-*,
   -misc-no-recursion,
-  -misc-no-recursion,
```
kgpai, line 33: Adding this would give a lot of warnings since we use recursion in many places , no ?

.clang-tidy
pedroerp: Sounds good. Could you create an Issue?

.clang-tidy
pedroerp: Yes, agreed. I've seen this warning internally, but I'm not sure if it's very actionable.

.clang-tidy
MBkkt: https://github.com/facebookincubator/velox/issues/15205

.clang-tidy
MBkkt: It's ok thing, but the problem right now almost all code written like
`X x{1};` not `X x{.x=1};` so I don't think it will produce meaningful results

.clang-tidy
MBkkt: > Also think this is a good thing to have , but I realize we have some big identifiers.

`auto it = map.find;` is completely ok, but this check is insane and complains about this.
Ofc it can be adjusted, but this check is too specific

.clang-tidy
MBkkt: I don't know, it was disabled, I didn't touch it, just remove duplicate

mbasmanova:
@MBkkt Would you rebase?

MBkkt:
@mbasmanova done

mbasmanova:
Seeing linter errors. CC: @pedroerp 

Screenshot 2025-10-23 at 1 30 14 PM

MBkkt:
@mbasmanova I think linter can be valid about this, but I didn't change this

PR #15137, opened 2025-10-12
refactor: Remove usage of folly::StringPiece in JsonExtractor

pedroerp:
Summary:
std::string_view.

Differential Revision: D84436547

### DIFF
--- a/velox/functions/prestosql/benchmarks/JsonExprBenchmark.cpp
+++ b/velox/functions/prestosql/benchmarks/JsonExprBenchmark.cpp
@@ -13,10 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/**
- * This file tests the performance of each JsonXXXFunction.call() with
- * expression framework and Velox vectors.
- */
+
 #include <folly/Benchmark.h>
 #include <folly/init/Init.h>
 #include "velox/functions/Registerer.h"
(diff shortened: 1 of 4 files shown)


velox/functions/prestosql/benchmarks/JsonExprBenchmark.cpp
```diff
@@ -159,9 +159,8 @@ struct FollyJsonSizeFunction {
       int64_t& result,
       const arg_type<Json>& json,
       const arg_type<Varchar>& jsonPath) {
-    const folly::StringPiece& jsonStringPiece = json;
-    const folly::StringPiece& jsonPathStringPiece = jsonPath;
-    auto extractResult = jsonExtract(jsonStringPiece, jsonPathStringPiece);
+    auto extractResult =
+        jsonExtract(std::string_view(json), std::string_view(jsonPath));
```
MBkkt, line 166: right now `velox::StringView` have implicit cast to `folly::StringPiece` and explicit to `std::string_view`

I think we need to change this before we change all code to string_view :)
Otherwise we will do explicit "casts" everywhere, this is not really readable.

I made these changes, but it doesn't fully work yet, but I will fix it.

pedroerp: yep, that's why I'm working on these PRs. The problem is that we couldn't really make the `std::string_view` cast implicit as well. So I'm first cleaning up all usages of `folly::StringPiece` (since we want remove them in any case), then remove the implicit cast to `folly::StringPiece`, and make the case to `std::string_view` implicit. 

It will take a bit more refactoring since we have a bunch of internal scalar functions that will need to be changed, but removing usages of `folly::StringPiece` will be a net win. 

But yes, for now we will have to add the explicit casts. I can add a TODO so we remember to remove them later.

PR #15140, opened 2025-10-13
fix: Rewrite tsan atomic to always use std atomic

### LINKED ISSUE #14985: We shouldn't hack TSAN, tsan dataraces are real, even if sometimes it's ok on x86
### Bug description

We shouldn't have such hacks

https://github.com/facebookincubator/velox/blob/8875b4d08b26be2743bb5e79efe0548bebd08042/velox/common/base/Portability.h#L54-L58

Because
1. it's UB in C++, we shouldn't justify this, because compiler can exploit it and produce unexpected code
2. it's difficult to understand where is "ok" places to have data races, instead of human logic we should rely on instruments, such as TSAN
3. It doesn't work on all architectures expected way
4. It isn't fastest way to do such things as counters for statistics (because same memory location invalidated from different threads)

Instead we should have special lightweight counters for statistics.
I believe we can start with simple fix:

1. replace tsan_atomic with just atomic, but with atomic with relaxed operations.
2. replace tsan_lock_guard with common lock_guard

If needed we can implement shared (per core (see rseq)/thread) counters for statistics.
There's a lot of opensource (apache/mit) implementation that can be copied, e.g. https://github.com/userver-framework/userver/blob/develop/core/include/userver/concurrent/striped_counter.hpp

CC @mbasmanova 

### System information

any

### Relevant logs

```bash

```

MBkkt:
Fixes https://github.com/facebookincubator/velox/issues/14985

### DIFF
--- a/velox/common/base/Portability.h
+++ b/velox/common/base/Portability.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <fmt/core.h>
 #include <atomic>
 #include <cstddef>
 #include <cstdint>
@@ -52,42 +53,80 @@ namespace facebook::velox {
 #endif
 #endif
 
-/// Define tsan_atomic<T> to be std::atomic<T?> for tsan builds and
-/// T otherwise. This allows declaring variables like statistics
-/// counters that do not have to be exact nor have synchronized
-/// semantics. This deals with san errors while not incurring the
-/// bus lock overhead at run time in regular builds.
-#ifdef TSAN_BUILD
+/// tsan_atomic is relaxed atomic without use rmw operations.
+/// This allows declaring variables like statistics counters that do not have to
+/// be exact nor have synchronized semantics.
+/// TODO: Use different name.
 template <typename T>
-using tsan_atomic = std::atomic<T>;
+class tsan_atomic : protected std::atomic<T> {
+ public:
+  using std::atomic<T>::atomic;
 
-template <typename T>
-inline T tsanAtomicValue(const std::atomic<T>& x) {
-  return x;
-}
+  T operator=(T v) noexcept {
+    this->store(v, std::memory_order_relaxed);
+    return v;
+  }
 
-/// Lock guard in tsan build and no-op otherwise.
-template <typename T>
-using tsan_lock_guard = std::lock_guard<T>;
+  operator T() const noexcept {
+    return this->load(std::memory_order_relaxed);
+  }
 
-#else
+  T operator+=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue + v;
+    this->store(newValue, std::memory_order_relaxed);
+    return newValue;
+  }
+  T operator-=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue - v;
+    this->store(newValue, std::memory_order_relaxed);
+    return newValue;
+  }
+  T operator&=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue & v;
+    this->store(newValue, std::memory_order_relaxed);
+    return newValue;
+  }
+  T operator|=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue | v;
+    this->store(newValue, std::memory_order_relaxed);
+    return newValue;
+  }
+  T operator^=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue ^ v;
+    this->store(newValue, std::memory_order_relaxed);
+    return newValue;
+  }
 
-template <typename T>
-using tsan_atomic = T;
+  T operator++(int) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue + T(1);
+    this->store(newValue, std::memory_order_relaxed);
+    return oldValue;
+  }
+  T operator--(int) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue - T(1);
+    this->store(newValue, std::memory_order_relaxed);
+    return oldValue;
+  }
 
-template <typename T>
-inline T tsanAtomicValue(T x) {
-  return x;
-}
-template <typename T>
-struct TsanEmptyLockGuard {
-  TsanEmptyLockGuard(T& /*ignore*/) {}
+  T operator++() noexcept {
+    return *this += T(1);
+  }
+  T operator--() noexcept {
+    return *this -= T(1);
+  }
 };
 
 template <typename T>
-using tsan_lock_guard = TsanEmptyLockGuard<T>;
-
-#endif
+inline T tsanAtomicValue(const tsan_atomic<T>& x) {
+  return x;
+}
 
 template <typename T>
 inline void resizeTsanAtomic(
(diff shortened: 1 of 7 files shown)


mbasmanova:
CI is red.

MBkkt:
@mbasmanova Looks like some unrelated failure to me, maybe I need to rebase.

Also always wanted to ask, can you enable for Velox and Axiom repo this setting:
It will allow to rebase/merge upstream from github UI
`https://github.com/<repo-org>/<repo-name>/settings`
image

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch

mbasmanova:
@MBkkt 

> can you enable for Velox and Axiom repo this setting:

done.

MBkkt:
@mbasmanova Would you review this PR, I hope Fedora OOM issue would be fixed with update to this commit https://github.com/facebookincubator/velox/commit/d187cb0fbb3622b00a044836c119ba8c8ee00b34

velox/common/base/Portability.h
```diff
@@ -52,42 +53,80 @@ namespace facebook::velox {
...
-template <typename T>
-using tsan_lock_guard = std::lock_guard<T>;
+  operator T() const noexcept {
+    return this->load(std::memory_order_relaxed);
+  }
 
-#else
+  T operator+=(T v) noexcept {
+    const auto oldValue = this->load(std::memory_order_relaxed);
+    const auto newValue = oldValue + v;
```
pedroerp: maybe a n00b question, but isn't this the same as

`this->store(*this + v, std::memory_order_relaxed)`

And maybe add a comment explaining how this is unsafe (but by design)?

velox/vector/FlatVector-inl.h
```diff
@@ -83,19 +83,19 @@ xsimd::batch<T> FlatVector<T>::loadSIMDValueBufferAt(size_t byteOffset) const {
 
 template <typename T>
 std::unique_ptr<SimpleVector<uint64_t>> FlatVector<T>::hashAll() const {
-  using len_type = decltype(BaseVector::length_);
+  const vector_size_t length = BaseVector::length_;
```
pedroerp, line 86: this also feels unrelated (but probably harmless).

velox/common/memory/MemoryPool.h
```diff
@@ -1036,7 +1036,7 @@ class MemoryPoolImpl : public MemoryPool {
   // object if not null. For example, a memory pool can reclaim the used memory
   // from a spillable operator through disk spilling. If null, we can't reclaim
   // memory from this memory pool.
-  std::unique_ptr<MemoryReclaimer> reclaimer_;
+  std::atomic<MemoryReclaimer*> reclaimer_;
```
pedroerp, line 1039: is this unrelated? Do we make changes to this unique_ptr from multiple threads? If this is a different fix, perhaps submit as a separate diff so we can isolate possible failures/breakages?

pedroerp (commented):
@MBkkt thanks for the cleanup! A couple of small comments, but overall it looks good

velox/common/base/Portability.h
MBkkt: > maybe a n00b question, but isn't this the same as

I think it's same, I just write it this more verbose way, I will fix it.

> And maybe add a comment explaining how this is unsafe (but by design)?

Ok

velox/vector/FlatVector-inl.h
MBkkt: Well, it works with tsan_atomic, but `BaseVector::length_` is tsan_atomic.
We don't want to tsan_atomic will be used for `for (tsan_atomic i = 0; ... ; ++i)`

velox/common/memory/MemoryPool.h
MBkkt: It's related, it was used with tsan_lock_guard, I think using real lock_guard in this place can be bad for performance. So I simplify this code with atomic instead.

velox/common/base/Portability.h
MBkkt: https://github.com/facebookincubator/velox/pull/15140/commits/c8d6e8d4b63da020ddcb226ff8161ffb7db90125

velox/vector/FlatVector-inl.h
MBkkt: `for (len_type i = 0; i < BaseVector::length_; ++i)`
this code now is
`for (tsan_atomic<...> i = 0; i < BaseVector::length_; ++i)`

velox/common/memory/MemoryPool.h
MBkkt:
```
 tsan_lock_guard<std::mutex> l(mutex_);
  return reclaimer_.get();
```
This code was just reclaimer.get() without TSAN.
It doesn't look good, so I fixed it with atomic and removed tsan_lock_guard here.
Because real lock_guard probably can be painful for performance

velox/vector/FlatVector-inl.h
pedroerp: I didn't realize BaseVector::length_ was tsan atomic. You're right. Ignore my comment :)

velox/common/memory/MemoryPool.h
pedroerp: But isn't this different, considering that now you're not acquiring ownership over MemoryReclaimer anymore? 

I mean, here you are potentially changing the semantics of these objects and their expectations of lifetime of MemoryReclaimer. If we want to do this, I think this would be better in a separate PR, as this may end up breaking something else down the road, depending on how these memory pools are used.

MBkkt: > I mean, here you are potentially changing the semantics of these objects and their expectations of lifetime of MemoryReclaimer. If we want to do this, I think this would be better in a separate PR, as this may end up breaking something else down the road, depending on how these memory pools are used.

For code without tsan, I didn't change semantics in a weaker side.
Without tsan this code was just `return reclaimer_.get();`
Now it's `return reclaimer_.load(std::memory_order_relaxed);` with and without tsan, how do I break something without tsan?

It can break things with tsan, but then it's actually was a hidden bug and it anyway needs to be fixed

MBkkt: I have three options:
1. just remove tsan_lock_guard => probably will be data race, otherwise why is it written?
2. replace tsan_lock_guard with real lock_guard => memory pools and this method used a lot => perf penalty
3. use atomic to avoid data race and hopes it was correct code (without race condition/etc), if not new tsan run will highlight this

MBkkt: For code without tsan, I made code a little stricter when it was.

MBkkt: https://github.com/facebookincubator/velox/pull/15179

I made separate PR with tsan_lock_guard if it helps

pedroerp: Maybe I'm missing something, but my point is that we used to hold a `unique_ptr` as a class member, which takes ownership over the object you acquired. Now you are just holding a (non-owning) pointer to MemoryReclaimer. 

Who is now responsible for guaranteeing it's kept alive? MemoryPool, or MemoryPool's caller?

pedroerp: Maybe here what you want need is atomic unique_ptr, but I don't think this exists in C++.

MBkkt: @pedroerp Atomic unique_ptr isn't needed, it can be much simpler: `MemoryPoolImpl` is owner.
Please consider to read this example: https://godbolt.org/z/KbzKPs6rT

PR #15141, opened 2025-10-13
vector grouping

duanmeng:


velox/exec/MergeJoin.h
```diff
@@ -21,6 +21,162 @@
 
 namespace facebook::velox::exec {
 
+/// VectorGrouping is a streaming Velox Operator that groups consecutive rows by
```
Yuhta, line 24: @mbasmanova This is exactly the same as the session aware rebatcher we have internally.  Should we beef this implementation up to a full fledged rebatcher operator in base we want to move rebatch inside Velox?

mbasmanova: @Yuhta We replaced session aware rebatcher with streaming aggregation.

Yuhta: Is there a possibility for the regular rebatcher to move into Velox in the furure?  If that's the case, it might make sense to have only one rebatcher implementation to handle all similar use cases.  It seems people in OSS community also have interests in doing something similar to session aware rebatcher.

mbasmanova: @duanmeng Do you want to share a bit about the motivating use case for this new plan node and operator?

duanmeng: @mbasmanova We need to re-batch our data by `userId` for a user-level training use case. The requirement is that each output batch must contain all rows for one, and only one, user.
For example, if the input vectors are `[1, 1, 2, 2]`,`[2, 3, 3, 4]`, the output must be: `[1, 1]`, `[2, 2, 2]`, `[3, 3]`, `[4]`

PR #15145, opened 2025-10-13
feat(cuDF): Add CudfExpression interface to allow nesting of cuDF AST and cuDF standalone functions

devavret:
This PR adds the following features in cuDF based expression evaluation.
1. `CudfExpression` class which is extended to implement cuDF based AST expression evaluation and standalone cuDF functions based evaluations. This allows for arbitrary nesting of AST and cudf functions. Previously only one level of nesting was supported and only AST could nest functions based evaluation. This class will be used in the future to add in JIT based evaluation without changing ast or function expressions.
2. Function registration: Now `CudfFunction`s can be registered without adding a corresponding entry in AST's pushExprToTree.
3. Expressions now return `ColumnOrView`. This puts the onus on the caller for materialization if the expression only refers to a field. This makes accessing nested fields zero-copy.
4. Expression evaluator registration: Now ASTExpression can be optionally not registered to force standalone function based evaluation. Registration also takes a priority value to prioritize one type of expression translation over the other. In case of duplicate functionality e.g when replacing a velox Expr arithmetic binary op, one can prioritize replacing with cudf AST over cudf::binary_operation

This PR also refactors SubfieldFilterToAST converter and ASTExpression into separate source files.

Contributes to #14149

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -1143,6 +1143,14 @@ Note: These configurations are experimental and subject to change.
      - string
      - ""
      - The prefix to use for the function names in cuDF.
+   * - cudf.ast_expression_enabled
+     - bool
+     - true
+     - If true, enable using cuDF AST-based expression evaluation when supported.
+   * - cudf.ast_expression_priority
+     - integer
+     - 100
+     - Priority of cuDF AST expressions. Higher value wins when multiple cuDF execution options are available for the same Velox expression. Standalone cuDF functions have priority 50. If enabled, with a default priority of 100, AST will be chosen as replacement for cudf execution.
    * - cudf.allow_cpu_fallback
      - bool
      - true


velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -76,14 +76,25 @@ CudfFilterProject::CudfFilterProject(
       debugPrintTree(expr);
     }
   }
-  std::vector<std::shared_ptr<velox::exec::Expr>> projectExprs;
   if (hasFilter_) {
     // First expr is Filter, rest are Project
-    filterEvaluator_ = ExpressionEvaluator({info.exprs->exprs()[0]}, inputType);
-    projectExprs = {info.exprs->exprs().begin() + 1, info.exprs->exprs().end()};
+    filterEvaluator_ = createCudfExpression(info.exprs->exprs()[0], inputType);
+    std::transform(
```
jinchengchenghh, line 179: Move projectEvaluators_ initialization out of if else branch, looks like they are same

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -159,15 +170,31 @@ void CudfFilterProject::filter(
...
   // Rearrange columns to match outputType_
   std::vector<std::unique_ptr<cudf::column>> outputColumns(outputType_->size());
   // computed resultProjections
   for (int i = 0; i < resultProjections_.size(); i++) {
-    VELOX_CHECK_NOT_NULL(columns[i]);
-    outputColumns[resultProjections_[i].outputChannel] = std::move(columns[i]);
+    auto& columnOrView = columns[i];
+    if (std::holds_alternative<std::unique_ptr<cudf::column>>(columnOrView)) {
+      // Move the owned column
+      outputColumns[resultProjections_[i].outputChannel] =
```
jinchengchenghh, line 294: Can we output std::vector<std::unique_ptr<cudf::column_view>> here?

velox/experimental/cudf/CudfConfig.h
```diff
@@ -58,6 +62,18 @@ struct CudfConfig {
...
+  bool astExpressionEnabled{true};
+
+  /// Priority of AST expression. Expression with higher priority is chosen for
+  /// a given root expression.
+  /// Example:
+  /// Priority of expression that uses individual cuDF functions is 50.
+  /// If AST priority is 100 then for a velox expression node that is supported
+  /// by both, AST will be chosen as replacement for cudf execution, if AST
+  /// priority is 25 then standalone cudf function is chosen.
+  int astExpressionPriority{100};
```
jinchengchenghh, line 77: Will AST be removed if all the expressions are supported in velox expression node?

devavret: On the contrary, by default AST will be preferred if operation is supported by it. This means we don't have to implement many binary ops as functions.

velox/experimental/cudf/exec/CudfFilterProject.cpp
devavret: They are not same. If filter exists, then projectEvaluators start from begin() + 1, otherwise they start from begin()

velox/experimental/cudf/exec/CudfFilterProject.cpp
devavret: Do you want to avoid materialization of column_view? This would require changed in CudfVector and would be out of scope in this PR

karthikeyann:
@paul-aiyedun @misiugodfrey @simoneves

velox/experimental/cudf/exec/CudfFilterProject.cpp
jinchengchenghh: OK, keep it as it now

velox/experimental/cudf/CudfConfig.h
jinchengchenghh: If the binary ops  is recursive in the other Expressions like complex switch, we need to register the  binary ops as functions.  In that way, we need to register all binary ops as functions.

jinchengchenghh (approved):

velox/experimental/cudf/tests/utils/ExpressionTestUtil.h
```diff
@@ -0,0 +1,40 @@
...
+#include "velox/expression/ExprCompiler.h"
+#include "velox/parse/ExpressionsParser.h"
+#include "velox/parse/TypeResolver.h"
+
+#include <memory>
+#include <string>
+
+namespace facebook::velox::cudf_velox::test_utils {
+
+inline std::shared_ptr<exec::Expr> compileExecExpr(
```
jinchengchenghh: Remove inline, usually the function called for each rows is inline.

devavret: This is inline to prevent ODR violation because function body is in the header

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -300,16 +327,6 @@ std::vector<std::unique_ptr<cudf::column>> CudfFilterProject::project(
...
-  for (auto i = 0; i < outputColumns.size(); ++i) {
-    const auto cudfOutputType =
-        cudf::data_type(cudf_velox::veloxToCudfTypeId(outputType_->childAt(i)));
-
-    if (outputColumns[i]->type() != cudfOutputType) {
-      outputColumns[i] =
-          cudf::cast(*(outputColumns[i]), cudfOutputType, stream);
-    }
-  }
-
```
karthikeyann, line 312: nit: is cast not required anymore?
Fixable in follow up PR.

jinchengchenghh: I remember this is to fix year function, after support CudfFunction, we will need a YearFunction for Spark, and throw exception in AST Tree for year when type mismatches. https://github.com/rapidsai/velox/pull/35

velox/experimental/cudf/expression/ExpressionEvaluator.cpp
```diff
@@ -0,0 +1,865 @@
...
+  }
+
+  ColumnOrView eval(
+      std::vector<ColumnOrView>& inputColumns,
+      rmm::cuda_stream_view stream,
+      rmm::device_async_resource_ref mr) const override {
+    VELOX_CHECK(!inputColumns.empty());
+    auto inputTableView = convertToTableView(inputColumns);
+    return cudf::hashing::murmurhash3_x86_32(
+        inputTableView, seedValue_, stream, mr);
```
jinchengchenghh, line 467: After remove the cast in the end of FilterProject, we need to cast from uint32_t column to int32_t column here

devavret: The cast is now in `FunctionExprespression::eval` which only casts at the end of expression evaluation. We can add a similar cast to `AstExpression::eval` too.

devavret: https://github.com/devavret/velox/blob/23ea2607eb4b5eaa54daf64e5c068dd80b45b3bf/velox/experimental/cudf/expression/ExpressionEvaluator.cpp#L734-L742

devavret: I added a missing cast in ASTExpression also https://github.com/facebookincubator/velox/pull/15145/commits/84b7792ec8cd0900db2cee5ebe4c65f52279caad

PR #15153, opened 2025-10-13
feat: (WIP) Add support for PROJ dependency

kgpai:
Summary: We are adding the PROJ library for spherical geometric co-ordinate conversions and transformations. When working with large distances PROJ is more accurate and preferred over geos. We need PROJ to add support for certain geographic udf's.

Differential Revision: D84546083


CMake/resolve_dependency_modules/proj.cmake
```diff
@@ -0,0 +1,56 @@
...
+set(BUILD_SHARED_LIBS OFF)
+set(BUILD_TESTING OFF)
+set(ENABLE_TIFF OFF)
+set(ENABLE_CURL OFF)
+set(BUILD_APPS OFF)
+set(BUILD_PROJ ON)
+set(BUILD_PROJINFO ON)
+set(BUILD_PROJSYNC ON)
+
+FetchContent_MakeAvailable(proj)
```
assignUser, line 51: Wrap in 
`block()`
`endblock()`

PR #15158, opened 2025-10-14
refactor: Remove unused 'cache_load_quantum' flag

lingbin:
The 'cache_load_quantum' gflag is not used anywhere, so we should 
remove its declaration.


kKPulla:
@lingbin can you update your branch to pull in latest changes on main as the current build seems to be failing as per signals above?

lingbin:
> @lingbin can you update your branch to pull in latest changes on main as the current build seems to be failing as per signals above?

@kKPulla  already rebased, thanks.

PR #15159, opened 2025-10-14
feat: Update fmt to 11.2.0

czentgr:


czentgr:
Removing the commit with the new dependency images. Run with the new dependency images was all green.

CMake/resolve_dependency_modules/README.md
```diff
@@ -29,7 +29,7 @@ by Velox. See details on bundling below.
 | double-conversion | 3.1.5           | No       |
 | xsimd             | 10.0.0          | Yes      |
 | re2               | 2024-07-02      | Yes      |
-| fmt               | 10.1.1          | Yes      |
+| fmt               | 11.2.0          | Yes      |
```
czentgr: In the main CMakeLists we only require version 9.0.0 and later. So this version is not directly enforced.
There was a problem with the previous upgrade. Not a problem in the CI later on when we use a newer CLang because that will use the new version.

kgpai: Should we add a comment regarding that though - that we should avoid api's that are not present in 9.0.0 .

kgpai (approved):

CMake/resolve_dependency_modules/README.md
czentgr: Or we can enforce it in the cmake. If an older version is found it will be using the new bundled version.
I think the reason that it is still 9.0.0 is for backwards compatibility with additional code (outside of this repo) that uses an older fmt version. Or this isn't necessary anymore?

kgpai: Yes it will use the newer bundled version ,which is fine, but should point out that code that is not backwards compatible with 9.0.0 risks chance of being reverted (Its unfair on our part to add this additionally on developers , but I believe in most cases they are compatible , so hopefully its not too bad.)

PR #15166, opened 2025-10-14
fix: Reduce NUM_THREADS in Fedora workflow to fix OOM

abhinavmuk04:
Summary:
The Fedora Debug workflow was experiencing out-of-memory failures. Reducing the number of threads from 8 to 4 decreases parallelism and frees up memory during the build process, resolving the OOM issue.

This is a common workaround when new UDFs or other additions cause memory pressure during the CI build process.

Differential Revision: D84653776


kgpai (approved):

czentgr:
@abhinavmuk04 The fix is not quite complete. The problem is with the link step. Currently, the fedora build is still flaky. I think the NUM_LINK_JOBS needs to be set to 2.

abhinavmuk04:
> @abhinavmuk04 The fix is not quite complete. The problem is with the link step. Currently, the fedora build is still flaky. I think the NUM_LINK_JOBS needs to be set to 2.

@czentgr I have noticed this too. Just put out a fix: https://github.com/facebookincubator/velox/pull/15206

PR #15170, opened 2025-10-15
feat: Support extra schemes for hdfs filesystem

### LINKED ISSUE #15169: Support more schemes for libhdfs
### Description

Currently we have supported jvm-based libhdfs, which means we can access any file system that implements hadoop FileSystem through its java client, like:  `s3:// `, `rbf://`, `customFs://`, etc.

wForget:
Add a `filesystems::setHdfsExtraSupportedSchemes` method to allow setting more supported schemes for the hdfs filesystem. For extra supported scheme, we set `scheme://authority` to hdfs client `nn`, for more detail see: https://github.com/facebookincubator/velox/blob/a0e2f842e5afe507c9b7f71ac22f3f3b70a1da2c/velox/external/hdfs/hdfs.h#L298-L302

closes #15169

### DIFF
--- a/velox/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.cpp
+++ b/velox/connectors/hive/storage_adapters/hdfs/RegisterHdfsFileSystem.cpp
@@ -23,6 +23,8 @@
 #include "velox/dwio/common/FileSink.h"
 #endif
 
+#include <string_view>
+
 namespace facebook::velox::filesystems {
 
 #ifdef VELOX_ENABLE_HDFS


wForget:
I have verified this patch on gluten-1.5.0. After combining it with the https://github.com/wForget/gluten/commit/906962339cbaa8ff9b7e695af1ade83d44e6ba05 patch in gluten and adding `spark.gluten.sql.columnar.backend.velox.hdfsExtraSupportedSchemes=qbfs` conf, gluten can access our internal custom hadoop file system.

image

philo-he:
@wForget, could you please explain the advantage of using Java client over the C++ client to access S3 data? If S3 scheme is supported by both HDFS filesystem and S3 filesystem in Velox, which filesystem will be actually used? According to `isPathSupportedByRegisteredFileSystems`, it seems using which one will depend on the registering order.

wForget:
> @wForget, could you please explain the advantage of using Java client over the C++ client to access S3 data? 

The original intention of this PR was just to allow us to use jvm-based libhdfs to access more hadoop FileSystem implementations. I have not compared the performance differences between native S3 and JVM-based clients.

> If S3 scheme is supported by both HDFS filesystem and S3 filesystem in Velox, which filesystem will be actually used? According to `isPathSupportedByRegisteredFileSystems`, it seems using which one will depend on the registering order.

In this case we can disable enabled_s3 at build and then set supported s3a schemes for hdfs.

zhouyuan:
@wForget are you are trying to support for some Hadoop Compatible File System (https://cwiki.apache.org/confluence/display/HADOOP2/HCFS)? 
I think a more general approach is to support project like:
https://github.com/apache/opendal

wForget:
> @wForget are you are trying to support for some Hadoop Compatible File System (https://cwiki.apache.org/confluence/display/HADOOP2/HCFS)? I think a more general approach is to support project like: https://github.com/apache/opendal

Yes, we use a custom HCFS in our company, and there is no C++ client, so we have to use jvm-based libhdfs to access it.

wForget:
> I think a more general approach is to support project like: https://github.com/apache/opendal

I don't think integrating OpenDAL conflicts with this PR. This PR simply extends jvm-based libdhdfs to support more HCFS. OpenDAL provides a unified file access interface and integrates with most file system clients, and we need to support OpenDAL as a new file system.

PR #15172, opened 2025-10-15
fix: Cleanup code to make it Clang20 compatible

czentgr:
When compiling with Clang20 (default in Centos9)
various compile errors occurred that are fixed by these changes.

For example,
call to 'memset' is a pointer to non-trivially copyable type 'type'.

### DIFF
--- a/velox/common/file/FileSystems.h
+++ b/velox/common/file/FileSystems.h
@@ -48,10 +48,10 @@ struct FileOptions {
   /// etc.
   static constexpr folly::StringPiece kFileCreateConfig{"file-create-config"};
 
-  std::unordered_map<std::string, std::string> values;
+  std::unordered_map<std::string, std::string> values{};
   memory::MemoryPool* pool{nullptr};
   /// If specified then can be trusted to be the file size.
-  std::optional<int64_t> fileSize;
+  std::optional<int64_t> fileSize{};
 
   /// Whether to create parent directories if they don't exist.
   ///
--- a/velox/common/memory/Scratch.h
+++ b/velox/common/memory/Scratch.h
@@ -79,14 +79,17 @@ class Scratch {
     // stringop-overflow warning when 'newCapacity' is 0.
     folly::assume(capacity_ >= 0);
     if (newCapacity > capacity_) {
-      Item* newItems =
-          reinterpret_cast<Item*>(::malloc(sizeof(Item) * newCapacity));
+      auto* newItems =
+          reinterpret_cast<uint8_t*>(::malloc(sizeof(Item) * newCapacity));
       if (fill_ > 0) {
         ::memcpy(newItems, items_, fill_ * sizeof(Item));
       }
-      ::memset(newItems + fill_, 0, (newCapacity - fill_) * sizeof(Item));
+      ::memset(
+          newItems + fill_ * sizeof(Item),
+          0,
+          (newCapacity - fill_) * sizeof(Item));
       ::free(items_);
-      items_ = newItems;
+      items_ = reinterpret_cast<Item*>(newItems);
       capacity_ = newCapacity;
     }
     fill_ = std::min(fill_, newCapacity);
--- a/velox/experimental/wave/common/HashTable.h
+++ b/velox/experimental/wave/common/HashTable.h
@@ -81,7 +81,7 @@ struct AllocationRange {
         firstRowOffset(roundUp64(capacity / rowSize) / 8),
         rowOffset(firstRowOffset),
         stringOffset(capacity) {
-    ::memset(reinterpret_cast<char*>(base), 0, firstRowOffset);
+    ::memset(reinterpret_cast<uint8_t*>(base), 0, firstRowOffset);
   }
 
   AllocationRange(AllocationRange&& other) {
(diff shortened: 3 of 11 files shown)


velox/expression/signature_parser/Scanner.h
```diff
@@ -36,6 +36,8 @@ class Scanner : public yyFlexLexer {
       : yyFlexLexer(&arg_yyin, &arg_yyout),
         outputType_(outputType),
         input_(input) {}
+  ~Scanner(){};
```
MBkkt:
```suggestion
  ~Scanner() = default;
```

assignUser:
@czentgr ci is missing clang?

czentgr:
> @czentgr ci is missing clang?

I was trying to use the system clang which is Clang-20 for centos9 which the dep image has. Maybe it didn't get installed into the dependency image? I assumed because I have it locally.

velox/common/file/FileSystems.h
```diff
@@ -48,10 +48,10 @@ struct FileOptions {
   /// etc.
   static constexpr folly::StringPiece kFileCreateConfig{"file-create-config"};
 
-  std::unordered_map<std::string, std::string> values;
+  std::unordered_map<std::string, std::string> values{};
   memory::MemoryPool* pool{nullptr};
   /// If specified then can be trusted to be the file size.
-  std::optional<int64_t> fileSize;
+  std::optional<int64_t> fileSize{std::nullopt};
```
MBkkt: I think it's not needed, because default ctor exists and do same

velox/expression/signature_parser/Scanner.h
MBkkt:
```suggestion

  ~Scanner() = default;
```

scripts/setup-centos-adapters.sh
```diff
@@ -28,7 +28,7 @@
 
 set -efx -o pipefail
 
-VELOX_CUDA_VERSION=${CUDA_VERSION:-"12.8"}
```
kgpai, line 31: cc: @karthikeyann @devavret

czentgr: @kgpai This is a commit I will revert but will update it in a separate PR soon. I'm finding out what we would need to get everything going with Clang 20 because I want to run in the CI clean and not miss anything.

karthikeyann: 13.0 is supported by cudf. `CUDA 12.0+` is cudf requirement.

czentgr:
@assignUser @MBkkt Please re-review when you get a chance. I'm done with the testing. With Clang 20 used in the CI temporarily no new issues came up. I removed the test change and this is the final intended changes.

velox/tpcds/gen/dsdgen/dist.cpp
```diff
@@ -371,7 +371,6 @@ int load_dists() {
   read_ptr = tpcds_idx + tpcds_idx_len - (entry_count * IDX_SIZE);
   for (int i = 0; i < entry_count; i++) {
     d_idx_t entry;
-    memset(&entry, 0, sizeof(const d_idx_t));
```
MBkkt, line 374: There's difference between new and old code, for an example if someone compares two different entries with memcmp it won't work now (because there're padding bytes inside d_idx_t which are not initialized by zero).

But I think there's no such code, so it should be ok

velox/experimental/wave/common/HashTable.h
```diff
@@ -92,7 +92,7 @@ struct AllocationRange {
 
   void operator=(AllocationRange&& other) {
     *this = other;
-    memset(&other, 0, sizeof(AllocationRange));
+    memset(reinterpret_cast<char*>(&other), 0, sizeof(AllocationRange));
```
MBkkt: NIT: `uint8_t*` I think here it's the same, but it's commonly better approach to use unsigned char instead of signed for such usages.

See example where it's not the same https://github.com/abseil/abseil-cpp/commit/930a70d12e059efb6fd1ffaf3d1d3dd906d23877

velox/common/memory/Scratch.h
```diff
@@ -79,14 +79,17 @@ class Scratch {
     // stringop-overflow warning when 'newCapacity' is 0.
     folly::assume(capacity_ >= 0);
     if (newCapacity > capacity_) {
-      Item* newItems =
-          reinterpret_cast<Item*>(::malloc(sizeof(Item) * newCapacity));
+      auto newItems =
```
MBkkt: NIT
```suggestion
      auto* newItems =
```

velox/experimental/wave/common/HashTable.h
czentgr: Yes agreed, I will change that. I think I did it because of line 84. But that would need to change too.

velox/tpcds/gen/dsdgen/dist.cpp
czentgr: Yes, that's a good point.
The other option is to reinterpret cast it to `uint8*` and then set it. I tried to avoid the cast but would be closer to the original code.

MBkkt (approved):

assignUser (approved):

PR #15173, opened 2025-10-15
fix(parquet): Fix reading array of row

rui-mo:
The following exception was observed when reading the complex column. The issue 
occurred because the requested ROW type was being compared with the converted 
VARCHAR type, leading to an incompatibility error. This PR resolves the issue 
by unwrapping one level of the array for repeated struct children, allowing the 
nested requested type (nestedStringColumn:VARCHAR) to be correctly compared 
with the converted type.

> velox/dwio/parquet/reader/ParquetReader.cpp:958, Function:convertType, Expression: !requestedType || isCompatible( requestedType, isRepeated, [](const TypePtr& type) { return type->kind() == TypeKind::VARCHAR; }) Converted type VARCHAR is not allowed for requested type ROW<nestedIntsColumn:ARRAY<INTEGER>,nestedStringColumn:VARCHAR>, Source: RUNTIME, ErrorCode: INVALID_STATE


rui-mo:
HI @majetideepak, would you like to review this fix? Thanks!

rui-mo:
Hi @majetideepak, would you please review this change at your convenience? Thanks!

PR #15174, opened 2025-10-15
refactor: Make DummySourceNode in trace public

tanjialiang:
For extensibility, make DummySourceNode for custom trace node to use.

### DIFF
--- a/velox/exec/TraceUtil.h
+++ b/velox/exec/TraceUtil.h
@@ -27,6 +27,45 @@
 
 namespace facebook::velox::exec::trace {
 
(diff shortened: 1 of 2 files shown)


velox/exec/TraceUtil.h
```diff
@@ -27,6 +27,45 @@
 
 namespace facebook::velox::exec::trace {
 
+static const std::vector<core::PlanNodePtr> kEmptySources;
```
duanmeng, line 30: `inline const std::vector<core::PlanNodePtr> kEmptySources{};` ?

duanmeng (approved):
@tanjialiang LGTM, thanks.

PR #15179, opened 2025-10-15
fix: Remove tsan_lock_guard

MBkkt:
https://github.com/facebookincubator/velox/issues/14985

Code was really strange for these usecases.
Because without tsan `tsan_lock_guard` was just noop.
So in this PR I added more synchronization for build without tsan.
And make it same for tsan vs non-tsan.
So I hope if tsan highlight some issues we will fix this

### DIFF
--- a/velox/common/memory/MemoryPool.cpp
+++ b/velox/common/memory/MemoryPool.cpp
@@ -439,15 +439,16 @@ MemoryPoolImpl::MemoryPoolImpl(
     std::shared_ptr<MemoryPool> parent,
     std::unique_ptr<MemoryReclaimer> reclaimer,
     const Options& options)
-    : MemoryPool{name, kind, parent, options},
+    : MemoryPool{name, kind, std::move(parent), options},
       manager_{memoryManager},
       allocator_{manager_->allocator()},
       arbitrator_{manager_->arbitrator()},
-      reclaimer_(std::move(reclaimer)),
+      reclaimer_{reclaimer.get()},
       // The memory manager sets the capacity through grow() according to the
       // actually used memory arbitration policy.
-      capacity_(parent_ != nullptr ? kMaxMemory : 0) {
+      capacity_{parent_ != nullptr ? kMaxMemory : 0} {
   VELOX_CHECK(options.threadSafe || isLeaf());
+  std::ignore = reclaimer.release();
 }
 
 MemoryPoolImpl::~MemoryPoolImpl() {
@@ -483,6 +484,8 @@ MemoryPoolImpl::~MemoryPoolImpl() {
   if (destructionCb_ != nullptr) {
     destructionCb_(this);
   }
+
+  delete reclaimer_.load(std::memory_order_relaxed);
 }
 
 MemoryPool::Stats MemoryPoolImpl::stats() const {
@@ -1089,13 +1092,12 @@ void MemoryPoolImpl::setReclaimer(std::unique_ptr<MemoryReclaimer> reclaimer) {
         parent_->name());
   }
   std::lock_guard<std::mutex> l(mutex_);
-  VELOX_CHECK_NULL(reclaimer_);
-  reclaimer_ = std::move(reclaimer);
+  VELOX_CHECK_NULL(reclaimer_.load(std::memory_order_relaxed));
+  reclaimer_.store(reclaimer.release(), std::memory_order_relaxed);
 }
 
 MemoryReclaimer* MemoryPoolImpl::reclaimer() const {
-  tsan_lock_guard<std::mutex> l(mutex_);
-  return reclaimer_.get();
+  return reclaimer_.load(std::memory_order_relaxed);
 }
 
 std::optional<uint64_t> MemoryPoolImpl::reclaimableBytes() const {
--- a/velox/common/memory/MemoryPool.h
+++ b/velox/common/memory/MemoryPool.h
@@ -1036,7 +1036,7 @@ class MemoryPoolImpl : public MemoryPool {
   // object if not null. For example, a memory pool can reclaim the used memory
   // from a spillable operator through disk spilling. If null, we can't reclaim
   // memory from this memory pool.
-  std::unique_ptr<MemoryReclaimer> reclaimer_;
+  std::atomic<MemoryReclaimer*> reclaimer_;
 
   // The memory cap in bytes to enforce.
   tsan_atomic<int64_t> capacity_;
--- a/velox/common/base/AsyncSource.h
+++ b/velox/common/base/AsyncSource.h
@@ -152,7 +152,8 @@ class AsyncSource {
   // If true, move() will not block. But there is no guarantee that somebody
   // else will not get the item first.
   bool hasValue() const {
-    tsan_lock_guard<std::mutex> l(mutex_);
+    // TODO: Can we do this without a lock?
+    std::lock_guard<std::mutex> l(mutex_);
     return item_ != nullptr || exception_ != nullptr;
   }
 
--- a/velox/common/base/Portability.h
+++ b/velox/common/base/Portability.h
@@ -66,10 +66,6 @@ inline T tsanAtomicValue(const std::atomic<T>& x) {
   return x;
 }
 
-/// Lock guard in tsan build and no-op otherwise.
-template <typename T>
-using tsan_lock_guard = std::lock_guard<T>;
-
 #else
 
 template <typename T>
@@ -79,13 +75,6 @@ template <typename T>
 inline T tsanAtomicValue(T x) {
   return x;
 }
-template <typename T>
-struct TsanEmptyLockGuard {
-  TsanEmptyLockGuard(T& /*ignore*/) {}
-};
-
-template <typename T>
-using tsan_lock_guard = TsanEmptyLockGuard<T>;
 
 #endif
 
--- a/velox/common/caching/AsyncDataCache.h
+++ b/velox/common/caching/AsyncDataCache.h
@@ -441,7 +441,8 @@ class CoalescedLoad {
   bool loadOrFuture(folly::SemiFuture<bool>* wait, bool ssdSavable = true);
 
   State state() const {
-    tsan_lock_guard<std::mutex> l(mutex_);
+    // TODO: Can we do this without a lock?
+    std::lock_guard<std::mutex> l(mutex_);
     return state_;
   }
 
--- a/velox/common/caching/SsdFile.h
+++ b/velox/common/caching/SsdFile.h
@@ -330,7 +330,8 @@ class SsdFile {
   /// Asserts that the region of 'offset' is pinned. This is called by the pin
   /// holder. The pin count can be read without mutex.
   void checkPinned(uint64_t offset) const {
-    tsan_lock_guard<std::shared_mutex> l(mutex_);
+    // TODO: Can we do this without a lock?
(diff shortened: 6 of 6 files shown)


velox/common/memory/MemoryPool.cpp
```diff
@@ -1083,13 +1086,12 @@ void MemoryPoolImpl::setReclaimer(std::unique_ptr<MemoryReclaimer> reclaimer) {
         parent_->name());
   }
   std::lock_guard<std::mutex> l(mutex_);
-  VELOX_CHECK_NULL(reclaimer_);
-  reclaimer_ = std::move(reclaimer);
+  VELOX_CHECK_NULL(reclaimer_.load(std::memory_order_relaxed));
+  reclaimer_.store(reclaimer.release(), std::memory_order_relaxed);
```
MBkkt, line 1096: maybe we want memory_order_release + acquire in load.
I'm not sure, lets run tsan for this and clarify.

Before my changes it wasn't even atomic, so in any case it's not worse.

It's just allow us to find issue if they were hidden

velox/common/memory/MemoryPool.h
```diff
@@ -1036,7 +1036,7 @@ class MemoryPoolImpl : public MemoryPool {
   // object if not null. For example, a memory pool can reclaim the used memory
   // from a spillable operator through disk spilling. If null, we can't reclaim
   // memory from this memory pool.
-  std::unique_ptr<MemoryReclaimer> reclaimer_;
+  std::atomic<MemoryReclaimer*> reclaimer_;
```
pedroerp, line 1039: @MBkkt what's the relationship between this PR and https://github.com/facebookincubator/velox/pull/15140?

My question about memory ownership (now that we moved from a unique_ptr to a naked atomic pointer) is still unanswered there, and I see here we make the same change. Are these supposed to be a stack of PRs?

MBkkt:
@pedroerp Could you review this change, please?

pedroerp:
@MBkkt I still don't understand the change from `std::unique_ptr<MemoryReclaimer>` to `std::atomic<MemoryReclaimer*>`, and what that means from a memory ownership perspective. 

It seems to me that what you are expecting is a an atomic unique_ptr, but I don't think that exists in C++. There is an atomic shared_ptr though.

MBkkt:
@pedroerp 

> what that means from a memory ownership perspective.

In single threaded code there's **no difference** between my code and the main branch. I just make it manually managed pointer, but because this inside class and we don't assign a lot into this variable it's simple.

Lets me explain, what is difference in case of multiple threads?
tsan_lock_guard without tsan will be noop.
So we can just remove all tsan_lock_guard and see data races in tsan run, and then fix them in a correct way.
I think it's correct way to handle this issue, but we don't have tsan in github CI.
So my plan is following:
1. remove data races by simple code (just put honest lock_guard + TODO)
2. then we will have TSAN, reiterate on this and optimize if possible

But the MemoryPool is specific place, because it's performance sensitive.
So I need instead of replace tsan_lock_guard with honest lock_guard make it somehow same cheap as it was without lock_guard at all and still be correct for C++ memory model.
The way how to handle it, is to make unique_ptr atomic.
But it's not possible, so instead of unique_ptr I use manually managed raw pointer (but it works same, you can check code paths). This is possible because we assign only under the lock_guard (it was like this) and it's correct.

So from user perspective you can imagine that here written magic `atomic<unique_ptr`.
But from implementation perspective we need manually written atomic raw pointer

> atomic shared_ptr though.

It's ineffective, compared to simple atomic pointer.
First shared_ptr is two pointers
Second, it's commonly implemented as common shared_ptr + global sharded locks


Atomic raw pointer, without seq_cst almost same effective as without atomic at all.
Only difference that compiler cannot remove atomic load and use already loaded value, but such places can be optimized manually

velox/common/memory/MemoryPool.cpp
```diff
@@ -1083,13 +1086,12 @@ void MemoryPoolImpl::setReclaimer(std::unique_ptr<MemoryReclaimer> reclaimer) {
         parent_->name());
   }
   std::lock_guard<std::mutex> l(mutex_);
-  VELOX_CHECK_NULL(reclaimer_);
-  reclaimer_ = std::move(reclaimer);
```
MBkkt, line 1100: @pedroerp here there's no difference, because 
1. in the old and new code `reclaimer_` is empty, so we don't need to delete it (same in both version)
2. `reclaimer` ownership transferred to `reclaimer_` (same in both version)

velox/common/memory/MemoryPool.cpp
```diff
@@ -1083,13 +1086,12 @@ void MemoryPoolImpl::setReclaimer(std::unique_ptr<MemoryReclaimer> reclaimer) {
...
+  VELOX_CHECK_NULL(reclaimer_.load(std::memory_order_relaxed));
+  reclaimer_.store(reclaimer.release(), std::memory_order_relaxed);
 }
 
 MemoryReclaimer* MemoryPoolImpl::reclaimer() const {
-  tsan_lock_guard<std::mutex> l(mutex_);
-  return reclaimer_.get();
```
MBkkt, line 1105: @pedroerp 
Here wasn't used the fact it's unique_ptr, it's just pointer "load" (in both version)

velox/common/memory/MemoryPool.cpp
```diff
@@ -439,15 +439,16 @@ MemoryPoolImpl::MemoryPoolImpl(
...
       arbitrator_{manager_->arbitrator()},
-      reclaimer_(std::move(reclaimer)),
+      reclaimer_{reclaimer.get()},
       // The memory manager sets the capacity through grow() according to the
       // actually used memory arbitration policy.
-      capacity_(parent_ != nullptr ? kMaxMemory : 0) {
+      capacity_{parent_ != nullptr ? kMaxMemory : 0} {
   VELOX_CHECK(options.threadSafe || isLeaf());
+  std::ignore = reclaimer.release();
 }
```
MBkkt, line 452: @pedroerp 

Here is three things possible

1. someone throw exception before construction of `reclaimer_` => `reclaimer` will be removed (same in both version
2. someone thro exception after construction of  `reclaimer_` => `reclaimer` but before end of `MemoryPoolImpl` ctor. In old code `reclaimer_` will be deleted but `reclaimer` will be empty. In new code `reclaimer` will be deleted but `reclaimer_` isn't (because it's raw pointer). So both version behavior will be same.
3. Ctor successed, in both version ownership transferred from `reclaimer` to  `reclaimer_`

velox/common/memory/MemoryPool.cpp
```diff
@@ -483,6 +484,8 @@ MemoryPoolImpl::~MemoryPoolImpl() {
   if (destructionCb_ != nullptr) {
     destructionCb_(this);
   }
+
+  delete reclaimer_.load(std::memory_order_relaxed);
 }
```
MBkkt, line 489: @pedroerp 

In old code, we destroyed  `reclaimer_` field inside `MemoryPoolImpl` after "end" of `~MemoryPoolImpl`.
In new code in the end we destroy `reclaimer_` manually.

So the two code is kind of same.

MBkkt:
@pedroerp I commented all changes about reclaimer, I hope it helps

velox/common/memory/MemoryPool.cpp
```diff
@@ -439,15 +439,16 @@ MemoryPoolImpl::MemoryPoolImpl(
...
       allocator_{manager_->allocator()},
       arbitrator_{manager_->arbitrator()},
-      reclaimer_(std::move(reclaimer)),
+      reclaimer_{reclaimer.get()},
       // The memory manager sets the capacity through grow() according to the
       // actually used memory arbitration policy.
-      capacity_(parent_ != nullptr ? kMaxMemory : 0) {
+      capacity_{parent_ != nullptr ? kMaxMemory : 0} {
   VELOX_CHECK(options.threadSafe || isLeaf());
+  std::ignore = reclaimer.release();
```
pedroerp, line 451: @MBkkt I may be missing something terribly obvious here, but wouldn't we first  acquire a non-owning reference to the object in line 446, and here release the only owning reference to MemoryReclaimer, which will cause it to get destructed right away?

In other words, once the constructor is finished, who owns MemoryReclaimer?

velox/common/memory/MemoryPool.h
MBkkt: > what's the relationship between this PR and #15140?

I separated part (about tsan lock guards)  of 15140 pr to this PR, to simplify review.

> Are these supposed to be a stack of PRs?

Yes the idea is separate PRs, you can put atomic PR into draft, because the plan is to merge this first, and then rebase tsan atomic PR.

velox/common/memory/MemoryPool.cpp
MBkkt: > In other words, once the constructor is finished, who owns MemoryReclaimer?

Once constructor finished destructor of whole MemoryPoolImpl will be called, so before it will be finished we transfer ovnership from unique_ptr argument to atomic inside MemoryPoolImpl

Why didn't we do this in init?

Because can be exception before constructor finished, in such case destructor won't be called, so unique ptr argument still should own this memory

pedroerp: > (..) we transfer ownership from unique_ptr argument to atomic inside MemoryPoolImp

atomic will not hold ownership. It's just an atomic raw pointer

MBkkt: > inside MemoryPoolImp

Ownership hold by MemoryPoolImpl

MBkkt: I don't understand what do you not understand.
Same code but simplified, maybe it will help: https://godbolt.org/z/KbzKPs6rT

MBkkt: @pedroerp 

I think I see the issue

> I may be missing something terribly obvious here, but wouldn't we first acquire a non-owning reference to the object in line 446, and here release the only owning reference to MemoryReclaimer, which will cause it to get destructed right away?

> here release the only owning reference to MemoryReclaimer

release implemented like this 
```cpp
auto* unique_ptr::release() { 
  auto* was = raw_ptr_inside_unique_ptr_;
  raw_ptr_inside_unique_ptr_ = nullptr; // nothing deleted!
  return was;
}
```

> which will cause it to get destructed right away?

So the answer on this question, no, it won't.

Why?

Because C++ doesn't have garbage collector, it only has raw memory, RAII and reference counter (commonly named shared_ptr, weak_ptr and intrusive_ptr), but unique_ptr doesn't have some magic reference counter.

unique_ptr is just RAII class (as any class in C++), in destructor it will call deleter for pointer if it's not null.
After `unique_ptr::release()` inside unique_ptr will be null, so to avoid memory leak, someone else should call delete. I do this inside MemoryPoolImpl destructor.

That how ownership transferred from unique_ptr argument to MemoryPoolImpl, we save pointer into atomic and call delete for it in MemoryPoolImpl destructor.

MBkkt: @pedroerp Hi, is it clear now?

MBkkt: @pedroerp Hello, I just wanted to notify you that these PR and issue that it tries to solve still actual.

MBkkt: @pedroerp Hello, I just wanted to notify you that these PR and issue that it tries to solve still actual.

PR #15196, opened 2025-10-16
[Do not review] feat(type): Add VarcharN and VarbinaryN type support

pdabre12:


Yuhta:
We don't have plan to add support of VarcharN at runtime.  It can exist in type for expression transformation purpose, but all the runtime operators and functions are not supposed to support it.  CC: @mbasmanova @kgpai

PR #15203, opened 2025-10-17
fix: ST_CoordDim failing with wrong return type

jkhaliqi:
ST_CoordDim return type should be tinyint based on : https://github.com/prestodb/presto/issues/26344

### DIFF
--- a/velox/docs/functions/presto/geospatial.rst
+++ b/velox/docs/functions/presto/geospatial.rst
@@ -413,7 +413,7 @@ Accessors
 
     Returns the minimum convex geometry that encloses all input geometries.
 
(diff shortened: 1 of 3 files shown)


jagill:
Interesting.  Presto docs have it returning bigint (which is also different than integer): https://prestodb.io/docs/current/functions/geospatial.html#ST_CoordDim-Geometry-bigint

Let me check the code...

jagill:
The documentation is wrong, the Presto code has return type tinyint.  I'll approve this but let's make sure the CI passes.

jagill (approved):

jkhaliqi:
@jagill sounds good: posted a PR for presto documentation update as well : https://github.com/prestodb/presto/pull/26355

PR #15206, opened 2025-10-17
fix: Use 16-core runner and increase NUM_THREADS to 8

abhinavmuk04:
Summary: Reduced NUM_THREADS from 4 to 2 in the linux-build-base GitHub workflow to prevent resource exhaustion during CI builds. This change helps ensure more stable builds by reducing concurrent compilation threads while maintaining the same memory and link job limits.

Differential Revision: D84923538

### DIFF
--- a/.github/workflows/linux-build-base.yml
+++ b/.github/workflows/linux-build-base.yml
@@ -209,7 +209,7 @@ jobs:
           cd _build/debug && ctest -j 8 --output-on-failure --no-tests=error
 
   fedora-debug:
-    runs-on: 8-core-ubuntu-22.04
+    runs-on: 16-core-ubuntu
     container: ghcr.io/facebookincubator/velox-dev:fedora
     # prevent errors when forks ff their main branch
     if: ${{ github.repository == 'facebookincubator/velox' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/linux-build-base.yml
```diff
@@ -248,7 +248,7 @@ jobs:
           fmt_SOURCE: BUNDLED
           simdjson_SOURCE: BUNDLED
           gRPC_SOURCE: SYSTEM
-          MAKEFLAGS: NUM_THREADS=4 MAX_HIGH_MEM_JOBS=4 MAX_LINK_JOBS=3
+          MAKEFLAGS: NUM_THREADS=2 MAX_HIGH_MEM_JOBS=4 MAX_LINK_JOBS=3
```
kgpai: This will make things probably too slow - Can you make num_threads 8 and instead change the machine we are running to 16-core-ubuntu (https://www.internalfb.com/intern/opensource/ci/github-actions/runners) .

czentgr: Yes, the problem wasn't the number of threads per se. It was the number of parallel link jobs. The job would fail at the end during the linking. That's a common issue I've seen locally with memory constraints. Only MAX_LINK_JOBS would make a difference in that scenario or have more memory.

PR #15207, opened 2025-10-17
fix(ci): Reduce link jobs number to prevent OOM for fedora build

czentgr:
Despite the change from PR https://github.com/facebookincubator/velox/pull/15166 the link step
can still fail due to running out of memory during the link step and makes the CI pipeline unstable.

We need to further reduce the number of parallel link steps. Compiling the code is usually not a problem.


.github/workflows/linux-build-base.yml
```diff
@@ -248,7 +248,7 @@ jobs:
           fmt_SOURCE: BUNDLED
           simdjson_SOURCE: BUNDLED
           gRPC_SOURCE: SYSTEM
-          MAKEFLAGS: NUM_THREADS=4 MAX_HIGH_MEM_JOBS=4 MAX_LINK_JOBS=3
+          MAKEFLAGS: NUM_THREADS=4 MAX_HIGH_MEM_JOBS=4 MAX_LINK_JOBS=2
```
kgpai, line 251: I replied on another PR (#15206 ) , but lets give it a bigger machine.

czentgr: Sure. That will work too. It completed with 2 link parallel jobs as well.

czentgr:
Closed in favor of increasing the size of the runner from #15206 .

PR #15210, opened 2025-10-17
feat: Add support for casting TIME to timestampWithTimezone

han-yan01:
Summary:
- Added `castFromTime()` function that treats TIME values as wall-clock time in the session timezone and converts to UTC
- Respects `adjustTimestampToTimezone` configuration flag:
  - When enabled (default): Converts TIME from session timezone to UTC
  - When disabled: Treats TIME values as already in UTC
- Implemented constant vector optimization for the common case where input is a constant TIME vector (null or non-null)

Differential Revision: D84947052


velox/functions/prestosql/types/TimestampWithTimeZoneRegistration.cpp
```diff
@@ -76,6 +106,29 @@ void castFromDate(
   });
 }
 
+void castFromTime(
```
kgpai: Doesnt look like this is used. Please remove it .

PR #15214, opened 2025-10-17
feat: Add TypeOf support for TIME and TIME WITH TIMEZONE

pratikpugalia:
Summary: This diff changes TypeOf for new types to be implicitly supported such as TIME and TIME WITH TIMEZONE. Apart from having types that require custom typeOf handling all other types should be supported by lower(type->name())

Differential Revision: D84879127

### DIFF
--- a/velox/functions/prestosql/TypeOf.cpp
+++ b/velox/functions/prestosql/TypeOf.cpp
@@ -15,140 +15,103 @@
  */
 #include "velox/expression/VectorFunction.h"
 #include "velox/functions/prestosql/types/BigintEnumType.h"
-#include "velox/functions/prestosql/types/BingTileType.h"
-#include "velox/functions/prestosql/types/GeometryType.h"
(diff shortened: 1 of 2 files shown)


velox/functions/prestosql/TypeOf.cpp
```diff
@@ -15,140 +15,100 @@
...
 #include "velox/functions/prestosql/types/QDigestType.h"
 #include "velox/functions/prestosql/types/TDigestType.h"
-#include "velox/functions/prestosql/types/TimestampWithTimeZoneType.h"
-#include "velox/functions/prestosql/types/UuidType.h"
 #include "velox/functions/prestosql/types/VarcharEnumType.h"
 
 namespace facebook::velox::functions {
 namespace {
 
 std::string typeName(const TypePtr& type) {
```
kgpai, line 31: Can you add a small comment here , that by default it should handle most types unless it has mixed casing etc..

PR #15217, opened 2025-10-17
fix: Ensure ConstantTypedExpr and variant types are compatible

pramodsatya:


velox/core/Expressions.h
```diff
@@ -61,7 +61,12 @@ class ConstantTypedExpr : public ITypedExpr {
   // Variant::null() value is supported.
   ConstantTypedExpr(TypePtr type, Variant value)
       : ITypedExpr{ExprKind::kConstant, std::move(type)},
-        value_{std::move(value)} {}
+        value_{std::move(value)} {
+    VELOX_CHECK_EQ(
+        value_.kind(),
```
mbasmanova: This check is not sufficient for complex types. Use isTypeCompatible instead. Please, add tests to ensure that ConstantTypedExpr(ARRAY(VARCHAR()), variant::array({1,2,3}) fails.

velox/expression/tests/ExprTest.cpp
```diff
@@ -2595,8 +2595,8 @@ TEST_P(ParameterizedExprTest, constantToString) {
 
 TEST_F(ExprTest, constantEqualsNullConsistency) {
   // Constant expr created using variant
-  auto nullVariantToExpr =
-      std::make_shared<core::ConstantTypedExpr>(VARCHAR(), Variant{});
+  auto nullVariantToExpr = std::make_shared<core::ConstantTypedExpr>(
```
mbasmanova, line 2598: This change is incorrect. Once you use isTypeCompatible it won't be needed.

pramodsatya: `Variant{}` is initialized with `TypeKind::INVALID` [here](https://github.com/facebookincubator/velox/blob/bfa3218ea40a828dd4758980c6e199f1a311e7ab/velox/type/Variant.h#L321), so the null variant needs to be initialized with matching type, otherwise the test fails with below error using `isTypeCompatible`:
```
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Expression type does not match variant type
Retriable: False
Expression: value_.isTypeCompatible(ITypedExpr::type())
Function: ConstantTypedExpr
```
Is this fine?

velox/core/Expressions.h
pramodsatya: Thanks @mbasmanova, updated to use `isTypeCompatible`. Could you PTAL?

velox/expression/tests/ExprTest.cpp
mbasmanova: @pramodsatya I see. This is interesting. We may need to look into the use of TypeKind::INVALID. It feels like this shouldn't exist in the first place. Especially since I assume Variant::isNull() returns true for Variant{}. Perhaps, we need another fix to replace TypeKind::INVALID with TypeKind::UNKNOWN in Variant{} constructor or just handle INVALID the same way as UNKNOWN everywhere else. Would you create an issue about this? The change in this PR is fine.

velox/core/Expressions.h
```diff
@@ -61,7 +61,11 @@ class ConstantTypedExpr : public ITypedExpr {
   // Variant::null() value is supported.
   ConstantTypedExpr(TypePtr type, Variant value)
       : ITypedExpr{ExprKind::kConstant, std::move(type)},
-        value_{std::move(value)} {}
+        value_{std::move(value)} {
+    VELOX_CHECK(
+        value_.isTypeCompatible(ITypedExpr::type()),
+        "Expression type does not match variant type");
```
mbasmanova: Let's include both types in the error message to help troubleshooting?

velox/core/tests/ConstantTypedExprTest.cpp
```diff
@@ -579,4 +580,21 @@ TEST_F(ConstantTypedExprTest, toStringComplexTypes) {
       << "toString mismatch for OPAQUE variant vs vector";
 }
 
+TEST_F(ConstantTypedExprTest, variantTypeCheck) {
+  VELOX_ASSERT_THROW(
+      std::make_shared<core::ConstantTypedExpr>(INTEGER(), "abc"),
```
mbasmanova: Let's introduce a helper lambda and a constant for error message to reduce copy-paste.

```
VELOX_ASSERT_THROW(makeConstant(INTEGER(), "abc"), ...), kTypeMismatch);
```

velox/core/tests/ConstantTypedExprTest.cpp
```diff
@@ -579,4 +580,21 @@ TEST_F(ConstantTypedExprTest, toStringComplexTypes) {
...
+      "Expression type does not match variant type");
+  VELOX_ASSERT_THROW(
+      std::make_shared<core::ConstantTypedExpr>(INTEGER(), variant(123LL)),
```
mbasmanova: Let's add tests that use null variants.

velox/expression/tests/ExprTest.cpp
pramodsatya: Thanks @mbasmanova, I agree the use of TypeKind::INVALID seems incorrect here and changing to TypeKind::UNKNOWN would be better: https://github.com/pramodsatya/velox/commit/b760e87401cda5b0b4e468354583342b82596c89. Should I open an issue for further discussion or shall I open a PR with this commit?

mbasmanova: @pramodsatya Please, open a PR. Thanks.

velox/core/tests/ConstantTypedExprTest.cpp
```diff
@@ -579,4 +579,34 @@ TEST_F(ConstantTypedExprTest, toStringComplexTypes) {
       << "toString mismatch for OPAQUE variant vs vector";
 }
 
+TEST_F(ConstantTypedExprTest, variantTypeCheck) {
+  auto testVariantExpr =
+      [&](Variant value, const TypePtr type, const TypePtr expectedType) {
```
mbasmanova: pass arguments by const &

velox/core/tests/ConstantTypedExprTest.cpp
```diff
@@ -579,4 +579,34 @@ TEST_F(ConstantTypedExprTest, toStringComplexTypes) {
...
+        VELOX_ASSERT_THROW(
+            createVariantExpr(type, value),
+            fmt::format(
+                "Expression type {} does not match variant type {}",
+                type->toString(),
+                expectedType->toString()));
+        if (type->isPrimitiveType()) {
```
mbasmanova: Would you add an else branch to verify that createVariantExpr succeeds?

velox/expression/tests/ExprTest.cpp
pramodsatya: Thanks @mbasmanova, opened https://github.com/facebookincubator/velox/pull/15250, please take a look when you get a chance

mbasmanova (approved):
Thanks.

PR #15220, opened 2025-10-18
refactor: Consolidate fuzzer pools in SortBufferTest

duanmeng:


velox/exec/tests/SortBufferTest.cpp
```diff
@@ -124,6 +124,8 @@ class SortBufferTest : public OperatorTestBase,
   folly::Random::DefaultGenerator rng_;
   std::unordered_map<std::string, RuntimeMetric> stats_;
   std::unique_ptr<TestRuntimeStatWriter> statWriter_;
+  const std::shared_ptr<memory::MemoryPool> fuzzerPool_ =
```
tanjialiang: Shall we put all const together at top and use init list way to initialize the value?

duanmeng: Fixed

PR #15221, opened 2025-10-20
[Do not review] perf(Cast) : Optimize simple casts like INTEGER->BIGINT

aditi-pandit:
Avoid the overhead of peeling.

This code is for an experiment. The ideal design should enhance PeelEncoding classes to avoid extra vectors for single vector cases like CAST.

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -33,6 +33,28 @@
 #include "velox/vector/LazyVector.h"
 #include "velox/vector/SelectivityVector.h"
 
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+
+// x86 includes
+#if defined(__x86_64__) || defined(_M_X64)
+#if defined(__has_include)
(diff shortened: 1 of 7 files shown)


velox/expression/CastExpr.cpp
```diff
@@ -917,6 +917,185 @@ void CastExpr::applyPeeled(
...
+    out[i + 3] = static_cast<int64_t>(in[i + 3]);
+  }
+  for (; i < n; ++i) {
+    out[i] = static_cast<int64_t>(in[i]);
+  }
+}
+
+#if (defined(__x86_64__) || defined(_M_X64)) && defined(__AVX2__)
+// If compiled with -mavx2, we can directly call AVX2 intrinsics.
+inline void castIntegerToBigintAvx2(
```
Yuhta, line 967: This is not needed, just do a for loop and compiler will use SIMD automatically: https://godbolt.org/z/P8EzYW1fG

PR #15225, opened 2025-10-20
refactor: Remove usage of folly::StringPiece in dwio

pedroerp:
Summary:
X-link: https://github.com/facebookincubator/nimble/pull/287

std::string_view.

Differential Revision: D85061784


pedroerp:
> Is `sizeof(std::string_view)` the same size as `sizeof(StringPiece)`? If they are same the change should be safe

yes, they are both 2 words each (16 bytes).

Issue #15198, opened 2025-10-16
Parquet reader/writer produces incorrect results for a VARCHAR column


mbasmanova:
### Bug description

I'm seeing a failure when reading parquet file. The reader produces a LazyVector of type INTEGER, but after loading the vector it comes out as a vector of type VARCHAR.

The repro writes a table with 3 columns: DOUBLE, INTEGER, VARCHAR. Then reads it back. It appears that when reading back, the second column contains the values for the third column. It is not clear whether a bug is in the writer / reader or somewhere else.

Note that the issue is only reproducible when Parquet is used; it is working with DWRF.

https://github.com/facebookincubator/axiom/pull/515/commits/383f074276911b9b77bbee1e9478d8f43059a114

```
TEST_F(WriteTest, xxx) {
  SCOPE_EXIT {
    metadata_->dropTableIfExists("test");
    metadata_->dropTableIfExists("test2");
  };

  // Replacing parquet with dwrf below "fixes" the issue.
  runCtas(
      "CREATE TABLE test WITH (file_format = 'parquet') AS "
      "SELECT rand() as key, l_linenumber, cast(rand() as varchar) as x FROM lineitem",
      60175);
}
```

```
E1016 14:19:26.884835 332333 LocalHiveConnectorMetadata.cpp:320] Column: 1
E1016 14:19:26.884845 332333 LocalHiveConnectorMetadata.cpp:321] [LAZY INTEGER: 1000 elements, no nulls]
E1016 14:19:26.884863 332333 LocalHiveConnectorMetadata.cpp:322] 0: [0->0] 0.16195357810530708
1: [1->1] 0.6270444808268258
2: [2->2] 0.06635384886238982
3: [3->3] 0.39917190475088826
4: [4->4] 0.8613166684274954
E1016 14:19:26.905043 332333 LocalHiveConnectorMetadata.cpp:323] [DICTIONARY VARCHAR: 1000 elements, no nulls]
...
Reason: Type mismatch: VARCHAR vs. INTEGER
Retriable: False
Expression: vector.type()->equivalent(*type_)
Function: addStats
File: fbcode/axiom/connectors/hive/StatisticsBuilder.cpp
Line: 108
```

The logging is produced by this code:

```
      auto data = dataSource->next(kBatchSize, ignore).value();
      ...

      LOG(ERROR) << "Got " << data->size() << " rows: " << data->toString();
      for (auto i = 0; i < data->childrenSize(); ++i) {
[...]

PR #15228, opened 2025-10-20
fix(parquet): Handle Parquet reserved keywords when used as the column names

majetideepak:
Parquet file schema inference uses reserved keywords such as `key_value, key, value, list, element, bag, array_element` to handle array and map fields.
The schema inference must handle these keywords differently if these names appear as the top-level column names.
Example: `fileSchema = ROW({"key", "element"}, {BIGINT(), ARRAY(INTEGER())});`

A workaround for older deployments would be to avoid using the above keywords as column names.

Fixes: #15198

### DIFF
--- a/velox/dwio/common/MetricsLog.h
+++ b/velox/dwio/common/MetricsLog.h
@@ -24,7 +24,7 @@ namespace facebook::velox::dwio::common {
 class MetricsLog {
  public:
   static constexpr std::string_view LIB_VERSION_STRING{"1.1"};
-  static constexpr folly::StringPiece WRITE_OPERATION{"WRITE"};
+  static constexpr std::string_view WRITE_OPERATION{"WRITE"};
 
   enum class MetricsType {
     HEADER,
--- a/velox/dwio/dwrf/common/Common.h
+++ b/velox/dwio/dwrf/common/Common.h
@@ -29,11 +29,11 @@
 namespace facebook::velox::dwrf {
 
 // Writer version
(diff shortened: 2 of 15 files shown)


Yuhta (approved):

majetideepak:
I will add the merge label after fixing the failing tests.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
```diff
@@ -1757,3 +1757,34 @@ TEST_F(ParquetReaderTest, fileColumnVarcharToMetadataColumnMismatchTest) {
...
+  // Create an in-memory writer.
+  auto sink = std::make_unique<MemorySink>(
+      1024 * 1024, dwio::common::FileSink::Options{.pool = leafPool_.get()});
+  auto sinkPtr = sink.get();
+  const auto data = makeRowVector(
+      {makeFlatVector<int64_t>({1}), makeFlatVector<double>({1.1})});
+  parquet::WriterOptions writerOptions;
+  writerOptions.memoryPool = leafPool_.get();
+
+  // key is a Parquet reserved keyword. Ensure we don't add it twice during the
```
mbasmanova: @majetideepak Deepak, thank you for debugging and fixing this issue. For my own understanding, the problem was triggered by using 'key' as the column name. A workaround could be to rename the column. Is this accurate? If so, would be nice to update PR description to clarify.

majetideepak: @mbasmanova Yes. A workaround would be to rename the column. I updated the description.

PR #15231, opened 2025-10-21
fix(cudf): Fix all nulls case for skipping filter

karthikeyann:
This PR fixes orderby test failure listed in PR https://github.com/facebookincubator/velox/pull/14956


kgpai:
@karthikeyann This change might have broken the adapters build, can you folks look / fix ?

Issue #5315, opened 2023-06-20
Spark sql avg agg function support decimal


liujiayi771:
### Description

Spark sql avg agg function's signatures is different from Presto, we need support Spark sql separately.

I have done this part of the work in Gluten's velox branch, and I will submit the cods in the near future.

PR #11999, opened 2025-01-02
fix: Allow companion functions when result type is not resolvable given intermediate type

rui-mo:
The registrations of partial and merge companion functions does not require the 
result type is resolvable given intermediate type. This PR removes the 
limitations for them.The registration of merge_extract companion function used 
to depend on the resolving, and this PR uses the passed-in result type rather 
than the resolved from intermediate type.

PR #15244, opened 2025-10-22
feat: Add support for Spark avg(decimal) aggregate function

liujiayi771:
Support Spark avg(decimal) using simple UDAF interface. The current 
implementation adheres to Spark’s default behavior, in which
`spark.sql.decimalOperations.allowPrecisionLoss` is enabled.

Resolve https://github.com/facebookincubator/velox/issues/5315.
Resolve https://github.com/facebookincubator/velox/discussions/9048.

### DIFF
--- a/velox/docs/functions/spark/aggregate.rst
+++ b/velox/docs/functions/spark/aggregate.rst
@@ -10,12 +10,12 @@ General Aggregate Functions
 .. spark:function:: avg(x) -> double|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
-    When x is of type DECIMAL(p, s), the result type is DECIMAL(p + 4, s + 4),
-    and the intermediate results are (sum, count) pairs represented as row(DECIMAL(p + 10, s), BIGINT).
-    The current implementation for DECIMAL matches Spark avg's default behavior with allowPrecisionLoss=true.
+    When ``x`` is of type DECIMAL(p, s), the result type is DECIMAL(p + 4, s + 4),
+    and the intermediate results are (sum, count) pairs represented as ROW(DECIMAL(p + 10, s), BIGINT).
+    The current implementation for DECIMAL matches Spark avg's default behavior with spark.sql.decimalOperations.allowPrecisionLoss=true.
     For all other input types, the result type is DOUBLE,
-    and the intermediate results are (sum, count) pairs represented as row(DOUBLE, BIGINT).
-    When all inputs are nulls, the intermediate result is row(0, 0),
+    and the intermediate results are (sum, count) pairs represented as ROW(DOUBLE, BIGINT).
+    When all inputs are nulls, the intermediate result is ROW(0, 0),
     and the final result is null.
 
 .. spark:function:: bit_xor(x) -> bigint
--- a/velox/functions/sparksql/DecimalArithmetic.cpp
+++ b/velox/functions/sparksql/DecimalArithmetic.cpp
@@ -455,7 +455,7 @@ struct DecimalDivideFunction {
     auto [aPrecision, aScale] = getDecimalPrecisionScale(*inputTypes[0]);
     auto [bPrecision, bScale] = getDecimalPrecisionScale(*inputTypes[1]);
     auto [rPrecision, rScale] =
-        DecimalUtil::computeResultPrecisionScale<allowPrecisionLoss>(
+        DecimalUtil::computeDivideResultPrecisionScale<allowPrecisionLoss>(
             aPrecision, aScale, bPrecision, bScale);
     rPrecision_ = rPrecision;
     aRescale_ = rScale - aScale + bScale;
--- a/velox/functions/sparksql/DecimalUtil.h
+++ b/velox/functions/sparksql/DecimalUtil.h
@@ -216,7 +216,7 @@ class DecimalUtil {
   /// exceed 38, we reduce both the number of fraction digits and whole digits
   /// to fit within this limit.
   template <bool allowPrecisionLoss>
-  static std::pair<uint8_t, uint8_t> computeResultPrecisionScale(
+  static std::pair<uint8_t, uint8_t> computeDivideResultPrecisionScale(
       uint8_t aPrecision,
       uint8_t aScale,
       uint8_t bPrecision,
--- a/velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
+++ b/velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
@@ -61,8 +61,8 @@ class DecimalAverageAggregate {
       const TypePtr& resultType) {
     VELOX_CHECK_EQ(argTypes.size(), 1);
     auto inputType = argTypes[0];
-    this->sumType = detail::getSumType(inputType);
-    this->resultType = resultType;
+    this->sumType_ = detail::getSumType(inputType);
+    this->resultType_ = resultType;
   }
 
   static bool toIntermediate(
@@ -204,17 +204,17 @@ class DecimalAverageAggregate {
       }
 
       auto [resultPrecision, resultScale] =
-          getDecimalPrecisionScale(*fn->resultType.get());
+          getDecimalPrecisionScale(*fn->resultType_.get());
       auto [sumPrecision, sumScale] =
-          getDecimalPrecisionScale(*fn->sumType.get());
+          getDecimalPrecisionScale(*fn->sumType_.get());
 
       // Spark use DECIMAL(20,0) to represent long value.
       static const uint8_t countPrecision = 20;
       static const uint8_t countScale = 0;
 
       auto [dividePrecision, divideScale] =
-          functions::sparksql::DecimalUtil::computeResultPrecisionScale<true>(
-              sumPrecision, sumScale, countPrecision, countScale);
+          functions::sparksql::DecimalUtil::computeDivideResultPrecisionScale<
+              true>(sumPrecision, sumScale, countPrecision, countScale);
       divideScale = std::max<uint8_t>(divideScale, resultScale);
       auto sumRescale = divideScale - sumScale + countScale;
       int128_t avg;
@@ -239,9 +239,8 @@ class DecimalAverageAggregate {
     }
   };
 
- private:
-  TypePtr resultType;
-  TypePtr sumType;
+  TypePtr resultType_;
+  TypePtr sumType_;
 };
 
 } // namespace facebook::velox::functions::aggregate::sparksql
--- a/velox/functions/sparksql/tests/DecimalUtilTest.cpp
+++ b/velox/functions/sparksql/tests/DecimalUtilTest.cpp
@@ -35,6 +35,19 @@ class DecimalUtilTest : public testing::Test {
     ASSERT_EQ(overflow, expectedOverflow);
     ASSERT_EQ(r, expectedResult);
   }
+
+  template <bool allowPrecisionLoss>
+  void testComputeDivideResultPrecisionScale(
+      const uint8_t aPrecision,
+      const uint8_t aScale,
+      const uint8_t bPrecision,
+      const uint8_t bScale,
+      std::pair<uint8_t, uint8_t> expected) {
+    ASSERT_EQ(
+        DecimalUtil::computeDivideResultPrecisionScale<allowPrecisionLoss>(
+            aPrecision, aScale, bPrecision, bScale),
+        expected);
+  }
 };
 } // namespace
 
@@ -72,4 +85,19 @@ TEST_F(DecimalUtilTest, bounded) {
   testBounded(40, 3, {38, 3});
   testBounded(44, 42, {38, 38});
 }
+
+TEST_F(DecimalUtilTest, computeDivideResultPrecisionScale) {
+  // Test with allowPrecisionLoss = true.
+  testComputeDivideResultPrecisionScale<true>(10, 2, 5, 1, {17, 8});
+  testComputeDivideResultPrecisionScale<true>(38, 10, 10, 5, {38, 6});
+  testComputeDivideResultPrecisionScale<true>(1, 0, 1, 0, {7, 6});
+  testComputeDivideResultPrecisionScale<true>(20, 2, 20, 2, {38, 18});
+
+  // Test with allowPrecisionLoss = false.
+  testComputeDivideResultPrecisionScale<false>(10, 2, 5, 1, {17, 8});
+  testComputeDivideResultPrecisionScale<false>(38, 10, 5, 3, {38, 11});
+  testComputeDivideResultPrecisionScale<false>(1, 0, 1, 0, {7, 6});
+  testComputeDivideResultPrecisionScale<false>(30, 5, 10, 5, {38, 11});
+}
+
 } // namespace facebook::velox::functions::sparksql::test


liujiayi771:
cc @rui-mo, @jinchengchenghh.

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -210,6 +210,23 @@ class DecimalUtil {
     }
   }
 
+  /// Get the precision and scale after the divide computation, which is itself
+  /// inspired to SQLServer's one, Spark also uses this formulation.
+  /// https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql
+  /// Assume Divide a(p1, s1) and b(p2, s2).
+  /// The result precision = p1 - s1 + s2 + max(6, s1 + p2 + 1)
+  /// The result scale = max(s1 + p2 + 1, s2 + 6)
+  static std::pair<uint8_t, uint8_t> dividePrecisionScale(
```
jinchengchenghh: Does it consider precision loss mode false or true?

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -210,6 +210,23 @@ class DecimalUtil {
...
+      const uint8_t aPrecision,
```
jinchengchenghh: remove const when passing by value

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -210,6 +210,23 @@ class DecimalUtil {
...
+      const uint8_t aScale,
+      const uint8_t bPrecision,
+      const uint8_t bScale) {
+    uint8_t intDig = aPrecision - aScale + bScale;
```
jinchengchenghh: This may be duplicated with https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/DecimalArithmetic.cpp#L477, please do a refactor. And test for it.

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -129,16 +166,25 @@ exec::AggregateRegistrationResult registerAverage(
...
+                // Spark has an optimization rule named DecimalAggregate that
+                // converts input data to a Long for average calculation when
+                // the input data is of type decimal and the precision is <= 11.
+                // Therefore, the precision of the decimal inputs received by
+                // the decimal average is guaranteed to be > 11. Since the
+                // precision of the intermediate sum is input precision + 10, it
+                // will definitely be > 21, which means the intermediate sum
+                // will always be a long decimal. We can directly assume that
+                // the physical type of the intermediate sum is int128_t.
+                VELOX_USER_CHECK_GT(p, 11);
```
jinchengchenghh: Is it user faced exception?

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -129,16 +166,25 @@ exec::AggregateRegistrationResult registerAverage(
                   AverageAggregate<int32_t, double, double>>(resultType);
             case TypeKind::BIGINT: {
               if (inputType->isShortDecimal()) {
-                return std::make_unique<DecimalAverageAggregateBase<int64_t>>(
-                    resultType);
+                auto [p, s] = getDecimalPrecisionScale(*inputType.get());
```
jinchengchenghh: The s is not used, auto precision = getDecimalPrecisionScale(*inputType.get()).first;

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -99,13 +124,25 @@ exec::AggregateRegistrationResult registerAverage(
...
-                           .build());
+  signatures.push_back(
+      exec::AggregateFunctionSignatureBuilder()
+          .integerVariable("a_precision")
+          .integerVariable("a_scale")
+          .integerVariable("i_precision", "min(38, a_precision + 10)")
+          .integerVariable("r_precision", "min(38, a_precision + 4)")
+          .integerVariable("r_scale", "min(38, a_scale + 4)")
+          .argumentType("DECIMAL(a_precision, a_scale)")
+          .intermediateType("ROW(DECIMAL(i_precision, a_scale), bigint)")
```
jinchengchenghh: bigint -> BIGINT

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+#include "velox/exec/SimpleAggregateAdapter.h"
+#include "velox/functions/sparksql/DecimalUtil.h"
+
+namespace facebook::velox::functions::aggregate::sparksql {
+
+static TypePtr getSumType(const TypePtr& rawInputType) {
+  if (rawInputType->isRow()) {
+    return rawInputType->childAt(0);
+  }
+  auto [p, s] = getDecimalPrecisionScale(*rawInputType.get());
```
jinchengchenghh, line 28: ditto

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+ public:
+  using InputType = Row<TInputType>;
+
+  using IntermediateType =
+      Row</*sum*/ int128_t,
+          /*count*/ int64_t>;
+
+  using OutputType = TResultType;
+
+  TypePtr resultType;
```
jinchengchenghh: Move the field to private scope

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+  TypePtr sumType;
+
+  /// Spark's decimal sum doesn't have the concept of a null group, each group
+  /// is initialized with an initial value, where sum = 0 and count = 0. The
+  /// final agg may fallback to being executed in Spark, so the meaning of the
+  /// intermediate data should be consistent with Spark. Therefore, we need to
+  /// use the parameter nonNullGroup in writeIntermediateResult to output a null
+  /// group as sum = 0, count = 0. nonNullGroup is only available when
+  /// default-null behavior is disabled.
+  static constexpr bool default_null_behavior_ = false;
```
jinchengchenghh, line 56: default_null_behavior_  -> kDefaultNullBehavior

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+  /// and the count number of the input values. If the count is not 0, then if
+  /// sum is nullopt that means an overflow has happened.
+  struct AccumulatorType {
+    std::optional<int128_t> sum{0};
+    int64_t overflow{0};
+    int64_t count{0};
+
+    DecimalAverageAggregate<TInputType, TResultType>* fn_;
+
+    static constexpr bool is_aligned_ = true;
```
jinchengchenghh, line 91: ditto

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+
+  /// This struct stores the sum of input values, overflow during accumulation,
+  /// and the count number of the input values. If the count is not 0, then if
+  /// sum is nullopt that means an overflow has happened.
+  struct AccumulatorType {
+    std::optional<int128_t> sum{0};
+    int64_t overflow{0};
+    int64_t count{0};
+
+    DecimalAverageAggregate<TInputType, TResultType>* fn_;
```
jinchengchenghh: fn, if it is private member, it should be fn_ and move to private scope

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+
+    static constexpr bool is_aligned_ = true;
+
+    AccumulatorType() = delete;
+
+    explicit AccumulatorType(
```
jinchengchenghh: remove explicit since the number of arguments is not 1

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+        HashStringAllocator* /*allocator*/,
+        DecimalAverageAggregate<TInputType, TResultType>* fn)
+        : fn_(fn) {}
+
+    std::optional<int128_t> computeFinalResult() const {
+      if (!sum.has_value()) {
+        return std::nullopt;
+      }
+      auto const adjustedSum =
```
jinchengchenghh: const auto

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+        return std::nullopt;
+      }
+
+      auto [resultPrecision, resultScale] =
+          getDecimalPrecisionScale(*fn_->resultType.get());
+      auto [sumPrecision, sumScale] =
+          getDecimalPrecisionScale(*fn_->sumType.get());
+
+      // Spark use DECIMAL(20,0) to represent long value.
+      const uint8_t countPrecision = 20, countScale = 0;
```
jinchengchenghh: Use 2 lines to initialize variable

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+    static constexpr bool is_aligned_ = true;
+
+    AccumulatorType() = delete;
+
+    explicit AccumulatorType(
+        HashStringAllocator* /*allocator*/,
+        DecimalAverageAggregate<TInputType, TResultType>* fn)
+        : fn_(fn) {}
+
+    std::optional<int128_t> computeFinalResult() const {
```
jinchengchenghh: Could we return Status here?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,251 @@
...
+    } else {
+      out.copy_from(std::make_tuple(0, 0));
+    }
+    return true;
+  }
+
+  /// This struct stores the sum of input values, overflow during accumulation,
+  /// and the count number of the input values. If the count is not 0, then if
+  /// sum is nullopt that means an overflow has happened.
+  struct AccumulatorType {
```
jinchengchenghh, line 84: Some functions maybe private scope

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -110,5 +126,108 @@ TEST_F(AverageAggregationTest, avgAllNulls) {
   assertQuery(plan, expected);
 }
 
+TEST_F(AverageAggregationTest, avgDecimal) {
+  int64_t kRescale = DecimalUtil::kPowersOfTen[4];
+  // Short decimal aggregation
```
jinchengchenghh: End with `.`

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -110,5 +126,108 @@ TEST_F(AverageAggregationTest, avgAllNulls) {
...
+  auto shortDecimal = makeNullableFlatVector<int64_t>(
+      {1'000, 2'000, 3'000, 4'000, 5'000, std::nullopt}, DECIMAL(12, 1));
+  auto shortDecimalInput = makeRowVector({shortDecimal});
+  auto shortDecimalExpected = makeRowVector(
+      {makeNullableFlatVector<int64_t>({3'000 * kRescale}, DECIMAL(16, 5))});
```
jinchengchenghh: makeConstant

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: This variable must match the name specified in SimpleAggregateAdapter.
https://github.com/facebookincubator/velox/blob/6ad19ed24fe74779b387f6c8279525548cff2efb/velox/exec/SimpleAggregateAdapter.h#L117

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: Many functions will be called in SimpleAggregateAdapter, computeFinalResult can be private.

zhouyuan:
related:  https://github.com/facebookincubator/velox/pull/11999

velox/functions/sparksql/DecimalUtil.h
liujiayi771: Yes, I will reuse this function.

velox/functions/sparksql/DecimalUtil.h
liujiayi771: @jinchengchenghh In Spark 3.3, the avg function ultimately uses `Divide` to compute the result. The current implementation should follow the default behavior of `allowPrecisionLoss=true`. Starting from Spark 3.4, the community introduced `DecimalDivideWithOverflowCheck` and uses Java `BigDecimal` to perform division. Since `BigDecimal` calculations do not have `allowPrecisionLoss`, the final result is converted to the target data type using toPrecision.

The current computation process should align with Spark 3.3’s default behavior but is not entirely consistent with Spark 3.4 and later. However, this approach fully passes Spark’s unit tests, so I believe we can retain the current computation flow. Additionally, once the core code is merged into upstream, we can continue optimizing it. What do you think?

liujiayi771:
Should import this change. https://github.com/oap-project/velox/pull/509/files. cc @zhli1142015

velox/functions/sparksql/DecimalUtil.h
jinchengchenghh: I agree, let us describe the support status in PR description.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: Both p and s are used here.

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -210,6 +210,33 @@ class DecimalUtil {
     }
   }
 
+  // When allowing precision loss, computes the result precision and scale
+  // following Hive's formulas. When denying precision loss, calculates the
+  // number of whole digits and fraction digits. If the total number of digits
+  // exceed 38, we reduce both the number of fraction digits and whole digits to
+  // fit within this limit.
```
rui-mo: nit: use /// for comments of public API.

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -74,6 +75,30 @@ class AverageAggregate
...
+    const TypePtr& resultType) {
+  auto inputType = argTypes[0];
+  VELOX_USER_CHECK(inputType->isDecimal());
+  auto avgType = getAvgType(inputType);
+  if (inputType->isShortDecimal()) {
+    if (avgType->isShortDecimal()) {
+      return std::make_unique<exec::SimpleAggregateAdapter<
+          DecimalAverageAggregate<int64_t, int64_t>>>(
+          step, argTypes, resultType);
+    } else {
```
rui-mo: nit: drop the 'else' after return.

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -74,6 +75,30 @@ class AverageAggregate
...
+      return std::make_unique<exec::SimpleAggregateAdapter<
+          DecimalAverageAggregate<int64_t, int128_t>>>(
+          step, argTypes, resultType);
+    }
+  } else {
```
rui-mo: ditto

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -164,28 +202,24 @@ exec::AggregateRegistrationResult registerAverage(
                   resultType);
             case TypeKind::DOUBLE:
             case TypeKind::ROW:
+              if (inputType->childAt(0)->isDecimal()) {
+                return std::make_unique<
+                    exec::SimpleAggregateAdapter<DecimalAverageAggregate<
+                        int128_t /* unused */,
```
rui-mo: nit: drip the whitespaces and use `/*unused*/` for consistency.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,254 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/exec/SimpleAggregateAdapter.h"
+#include "velox/functions/sparksql/DecimalUtil.h"
+
+namespace facebook::velox::functions::aggregate::sparksql {
+
+static TypePtr getSumType(const TypePtr& rawInputType) {
```
rui-mo: Would it be more appropriate to drop 'static' and use 'inline'?

rui-mo: Consider moving this function to the namespace detail if it's only used in this header.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,254 @@
...
+    return rawInputType->childAt(0);
+  }
+  auto [p, s] = getDecimalPrecisionScale(*rawInputType.get());
+  // In Spark,
+  // sum precision = input precision + 10
+  // sum scale = input scale
+  return DECIMAL(std::min<uint8_t>(LongDecimalType::kMaxPrecision, p + 10), s);
+}
+
+static TypePtr getAvgType(const TypePtr& rawInputType) {
```
rui-mo: Can this function be moved the anonymous namespace of AverageAggregate.cpp?

velox/functions/sparksql/aggregates/AverageAggregate.cpp
liujiayi771: When Spark invokes it, precision is guaranteed to be greater than 11, and the current implementation is based on this assumption. However, users can directly call Spark's decimal average in Velox (e.g., in unit tests) and might pass a precision less than 11, which would cause issues. Therefore, a user exception is used here.

velox/functions/sparksql/aggregates/AverageAggregate.cpp
```diff
@@ -74,6 +75,37 @@ class AverageAggregate
   }
 };
 
+TypePtr getAvgType(const TypePtr& rawInputType) {
+  auto [p, s] = getDecimalPrecisionScale(*rawInputType.get());
+  // In Spark,
+  // sum precision = input precision + 4
+  // sum scale = input scale + 4
```
rui-mo: It is a bit confusing because the function name is `getAvgType` but the comments refer to `sum`. Is intermediate type for sum is returned, and would you please add comments for this function to clarify?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/exec/SimpleAggregateAdapter.h"
+#include "velox/functions/sparksql/DecimalUtil.h"
+
+namespace facebook::velox::functions::aggregate::sparksql {
+namespace detail {
+inline TypePtr getSumType(const TypePtr& rawInputType) {
```
rui-mo, line 24: nit: make the function surrounded by empty line.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+  if (rawInputType->isRow()) {
+    return rawInputType->childAt(0);
+  }
+  auto [p, s] = getDecimalPrecisionScale(*rawInputType.get());
+  // In Spark,
+  // sum precision = input precision + 10
+  // sum scale = input scale
```
rui-mo: Would you please clarify this is the intermediate sum?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+  /// use the parameter nonNullGroup in writeIntermediateResult to output a null
+  /// group as sum = 0, count = 0. nonNullGroup is only available when
+  /// default-null behavior is disabled.
+  static constexpr bool default_null_behavior_ = false;
+
+  void initialize(
+      core::AggregationNode::Step step,
+      const std::vector<TypePtr>& argTypes,
+      const TypePtr& resultType) {
+    VELOX_CHECK_LE(argTypes.size(), 1);
```
rui-mo: Can be arg types be empty?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+                               /*count*/ int64_t>;
+
+  using OutputType = TResultType;
+
+  /// Spark's decimal sum doesn't have the concept of a null group, each group
+  /// is initialized with an initial value, where sum = 0 and count = 0. The
+  /// final agg may fallback to being executed in Spark, so the meaning of the
+  /// intermediate data should be consistent with Spark. Therefore, we need to
+  /// use the parameter nonNullGroup in writeIntermediateResult to output a null
+  /// group as sum = 0, count = 0. nonNullGroup is only available when
```
rui-mo, line 54: For null group, is the sum 0 or nullopt?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+      if (!nonNullGroup || count == 0) {
+        // In Spark, if all inputs are null, count will be 0, and the result of
+        // average value will be null.
+        return false;
+      }
+      auto finalResult = computeFinalResult();
+      if (finalResult.has_value()) {
+        out = static_cast<TResultType>(finalResult.value());
+        return true;
+      } else {
```
rui-mo: nit: drop 'else'

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+            out.template set_null_at<0>();
+            out.template get_writer_at<1>() = count;
+          }
+        }
+      }
+      return true;
+    }
+
+    bool writeFinalResult(bool nonNullGroup, exec::out_type<OutputType>& out) {
+      if (!nonNullGroup || count == 0) {
```
rui-mo, line 180: Does `count == 0` indicates that the result is null for empty input?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+      }
+
+      auto [resultPrecision, resultScale] =
+          getDecimalPrecisionScale(*fn->resultType.get());
+      auto [sumPrecision, sumScale] =
+          getDecimalPrecisionScale(*fn->sumType.get());
+
+      // Spark use DECIMAL(20,0) to represent long value.
+      constexpr uint8_t countPrecision = 20;
+      constexpr uint8_t countScale = 0;
```
rui-mo: Could be declared as static variables.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,245 @@
...
+      divideScale = std::max<uint8_t>(divideScale, resultScale);
+      auto sumRescale = divideScale - sumScale + countScale;
+      int128_t avg;
+      bool overflow = false;
+      functions::sparksql::DecimalUtil::
+          divideWithRoundUp<int128_t, int128_t, int128_t>(
+              avg, adjustedSum.value(), count, sumRescale, overflow);
+      if (overflow) {
+        return std::nullopt;
+      } else {
```
rui-mo: nit: drop 'else'

rui-mo (commented):
Thanks!

velox/functions/sparksql/aggregates/AverageAggregate.cpp
liujiayi771: This function is used to get the result type of avg, the comments should be avg precision = ... and avg scale = ..., consistent with https://github.com/apache/spark/blob/a8e35c407bc5340f83b35e5a2f0b0767c6baadb0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Average.scala#L69. I will change the function name to getAvgResultType.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: The size should always be 1.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: In velox, for null group, the sum should be meaningless. In both `writeIntermediateResult` and `writeFinalResult`, if `nonNullGroup` is false, the values of sum and count are ignored.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: Yes. This is consistent with https://github.com/apache/spark/blob/a8e35c407bc5340f83b35e5a2f0b0767c6baadb0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Average.scala#L107.

rui-mo: Do we have a test to cover this?

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
rui-mo: > /// If the count is not 0, then if sum is nullopt that means an overflow has happened.

I wonder if we could make the distinctions between them more clear for readers.

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -30,13 +30,27 @@ class AverageAggregationTest : public AggregationTestBase {
  protected:
   void SetUp() override {
     AggregationTestBase::SetUp();
-    registerAggregateFunctions("spark_");
+    disableTestStreaming();
```
rui-mo: Gluten supports streaming aggregate, and I think we need to enable the streaming tests. Consider adding a test function following `testAggregations` and just removes the tests for the `extract` companion function.

https://github.com/apache/incubator-gluten/commit/4b5d0c1909eaa1f11631d82af806a09a247c3004

liujiayi771: You're referring to `testAggregationsWithCompanion`? All cases in this function will use either `extractExpressions` (which Spark cannot register to produce) or the merge_extract function (which cannot generate a plan through Velox parser). Therefore, I've separately tested the partial, merge, and merge_extract functions in `AverageAggregationTest`. We can now remove `disableTestStreaming` as the tests using `testAggregations` for validation will enter `validateStreamingInTestAggregations` testing.

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
liujiayi771: I added 2 test cases to verify scenarios where the input is empty or when the intermediate data count is 0. However, neither case can construct the situation where count equals 0 in `writeFinalResult`. This is because both `combine` and `writeIntermediateResult` will return false when count is 0 (resulting in a null group). The count in `writeFinalResult` can only potentially become 0 in the extract function, but Spark cannot register this companion function. Nevertheless, I believe keeping the `count == 0` check in `writeFinalResult` is still appropriate.

rui-mo: > The count in writeFinalResult can only potentially become 0 in the extract function

It looks like `writeIntermediateResult` can only return true. And does this indicates `writeFinalResult` can only be called when count > 0 in Spark's scenario?

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -30,13 +30,54 @@ class AverageAggregationTest : public AggregationTestBase {
  protected:
   void SetUp() override {
     AggregationTestBase::SetUp();
-    registerAggregateFunctions("spark_");
+    disableTestIncremental();
+    registerAggregateFunctions("spark_", true);
+  }
+
+  void globalDecimalAverageTest(
```
rui-mo: nit: The function doesn’t seem to be specific to decimal — perhaps we could consider a more appropriate name, like `testGlobalAverage`?

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -30,13 +30,54 @@ class AverageAggregationTest : public AggregationTestBase {
...
+    testAggregations(
+        input,
+        {},
+        {"spark_avg(c0)"},
+        {},
+        expected,
+        /*config*/ {});
+  }
+
+  core::PlanNodePtr createMergeExtractPlan(
```
rui-mo: nit: perhaps rename as `createMergeExtractDecimal`

velox/functions/sparksql/aggregates/tests/AverageAggregationTest.cpp
```diff
@@ -30,13 +30,54 @@ class AverageAggregationTest : public AggregationTestBase {
...
+        std::vector<std::string>{"c0"},
+        std::vector{avgAggregate},
+        /*ignoreNullKeys=*/false,
+        std::move(child));
   }
 };
 
 TEST_F(AverageAggregationTest, avgAllNulls) {
-  vector_size_t size = 1'000;
-  // Have two row vectors a least as it triggers different code paths.
```
rui-mo, line 39: nit: do we need to remove above comment?

rui-mo (commented):
Pease update the corresponding documentation as well, thanks!

https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/spark/aggregate.rst


> The current implementation should follow the default behavior of allowPrecisionLoss=true.

It would be helpful to document these more clearly.

liujiayi771:
@rui-mo Thanks! The comments have been addressed.

velox/docs/functions/spark/aggregate.rst
```diff
@@ -10,10 +10,11 @@ General Aggregate Functions
 .. spark:function:: avg(x) -> double|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
-    When x is of type DECIMAL, the result type is DECIMAL,
-    and the intermediate results are varbinarys or (sum, count) pairs represented as row(decimal, bigint).
+    When x is of type DECIMAL(p, s), the result type is DECIMAL(p + 4, s + 4),
```
rui-mo: nit: refer to the variable \``x\``

velox/docs/functions/spark/aggregate.rst
```diff
@@ -10,10 +10,11 @@ General Aggregate Functions
...
+    and the intermediate results are (sum, count) pairs represented as row(DECIMAL(p + 10, s), BIGINT).
+    The current implementation for DECIMAL matches Spark avg's default behavior with allowPrecisionLoss=true.
```
rui-mo: Perhaps provide the full configuration name `spark.sql.decimalOperations.allowPrecisionLoss`.

velox/docs/functions/spark/aggregate.rst
```diff
@@ -10,10 +10,11 @@ General Aggregate Functions
...
     For all other input types, the result type is DOUBLE,
-    and the intermediate results are (sum, count) pairs represented as row(double, bigint).
+    and the intermediate results are (sum, count) pairs represented as row(DOUBLE, BIGINT).
```
jinchengchenghh: Other data type is uppercase, so ROW(DOUBLE, BIGINT) might be metter

velox/docs/functions/spark/aggregate.rst
```diff
@@ -10,10 +10,11 @@ General Aggregate Functions
 .. spark:function:: avg(x) -> double|decimal
 
     Returns the average (arithmetic mean) of all non-null input values.
-    When x is of type DECIMAL, the result type is DECIMAL,
-    and the intermediate results are varbinarys or (sum, count) pairs represented as row(decimal, bigint).
+    When x is of type DECIMAL(p, s), the result type is DECIMAL(p + 4, s + 4),
+    and the intermediate results are (sum, count) pairs represented as row(DECIMAL(p + 10, s), BIGINT).
```
jinchengchenghh: ROW(DECIMAL(p + 10, s), BIGINT)

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -455,7 +455,8 @@ struct DecimalDivideFunction {
     auto [aPrecision, aScale] = getDecimalPrecisionScale(*inputTypes[0]);
     auto [bPrecision, bScale] = getDecimalPrecisionScale(*inputTypes[1]);
     auto [rPrecision, rScale] =
-        computeResultPrecisionScale(aPrecision, aScale, bPrecision, bScale);
+        DecimalUtil::computeResultPrecisionScale<allowPrecisionLoss>(
```
jinchengchenghh: computeDivideResultPrecisionScale

velox/functions/sparksql/DecimalUtil.h
```diff
@@ -210,6 +210,33 @@ class DecimalUtil {
     }
   }
 
+  /// When allowing precision loss, computes the result precision and scale
+  /// following Hive's formulas. When denying precision loss, calculates the
+  /// number of whole digits and fraction digits. If the total number of digits
+  /// exceed 38, we reduce both the number of fraction digits and whole digits
+  /// to fit within this limit.
+  template <bool allowPrecisionLoss>
+  static std::pair<uint8_t, uint8_t> computeResultPrecisionScale(
```
jinchengchenghh: Please add a test to here https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/tests/DecimalUtilTest.cpp

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,247 @@
...
+              resultPrecision,
+              resultScale,
+              rescaledValue);
+      return status.ok() ? std::optional<int128_t>(rescaledValue)
+                         : std::nullopt;
+    }
+  };
+
+ private:
+  TypePtr resultType;
```
jinchengchenghh: resultType_ and sumType_

velox/functions/sparksql/aggregates/DecimalAverageAggregate.h
```diff
@@ -0,0 +1,247 @@
...
+              divideScale,
+              resultPrecision,
+              resultScale,
+              rescaledValue);
+      return status.ok() ? std::optional<int128_t>(rescaledValue)
+                         : std::nullopt;
+    }
+  };
+
+ private:
```
jinchengchenghh: Duplicated private

rui-mo:
@jinchengchenghh @zhli1142015 Do you have any further comment? If not, we may proceed to merge this PR. Thanks.

rui-mo:
@liujiayi771 Can you please rebase this PR? Thanks.

PR #15246, opened 2025-10-22
perf: Reduce type parameter initialization cost and memory usage

Yuhta:
Summary: In many cases (e.g. ARRAY and MAP) type parameters are fixed size, we don't need dynamic allocation for this.  This is a visible memory cost when we have a lot of expressions.

Differential Revision: D85052874

### DIFF
--- a/velox/functions/prestosql/types/EnumTypeBase.h
+++ b/velox/functions/prestosql/types/EnumTypeBase.h
@@ -43,7 +43,7 @@ class EnumTypeBase : public TPhysical {
     return this == &other;
   }
 
-  const std::vector<TypeParameter>& parameters() const override {
+  std::span<const TypeParameter> parameters() const override {
     return parameters_;
   }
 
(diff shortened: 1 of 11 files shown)


velox/functions/prestosql/types/EnumTypeBase.h
```diff
@@ -43,7 +43,7 @@ class EnumTypeBase : public TPhysical {
     return this == &other;
   }
 
-  const std::vector<TypeParameter>& parameters() const override {
+  folly::Range<const TypeParameter*> parameters() const override {
```
MBkkt: can you use std::span please?

MBkkt:
```suggestion
  std::span<const TypeParameter> parameters() const override {
```

MBkkt (commented):
Overall, I did similar and it seems like it worked, I just didn't have time to contribute it. Thanks!

MBkkt (approved):

PR #15252, opened 2025-10-22
feat(geo): Increase bing tile max zoom shift to 6

jagill:
Summary:
To protect against exponentially growing numbers of bing tiles,
we put a max zoom shift of 5.  This constrained a bing tile to be
shifted only 5 zoom levels, which increases the number by `4**5 == 2**10`.
When rolling out this feature, a user is getting errors because they were
using a shift of 6, previously allowed.

This makes the default 6, so one bing tile becomes `4**6 == 2**12`.

Differential Revision: D85278103

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -791,7 +791,7 @@ class QueryConfig {
   }
 
   uint8_t debugBingTileChildrenMaxZoomShift() const {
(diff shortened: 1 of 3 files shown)


jagill:
Using config variables instead.

jagill:
Config variables had problems.  This is a slower but safer way.

PR #15259, opened 2025-10-23
refactor: Use `re2::StringPiece` for `RE2::FullMatch()`

kou:
`RE2::FullMatch()` uses `absl::string_view` not `std::string_view`:

https://github.com/google/re2/blob/61c4644171ee6b480540bf9e569cba06d9090b4b/re2/re2.h#L411

`absl::string_view` may not be an alias of `std::string_view`. In the case, the following error is reported:

```text
In file included from velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp:18:
velox/functions/prestosql/DateTimeFunctions.h:1939:10: error: no matching function for call to 'FullMatch'
 1939 |     if (!RE2::FullMatch(
      |          ^~~~~~~~~~~~~~
/include/re2/re2.h:411:15: note: candidate function template not viable: no known conversion from 'std::string_view' (aka 'basic_string_view<char>') to 'absl::string_view' for 1st argument
  411 |   static bool FullMatch(absl::string_view text, const RE2& re, A&&... a) {
      |               ^         ~~~~~~~~~~~~~~~~~~~~~~
```

Old RE2 that is provided by CentOS Stream 9 doesn't accept `absl::string_view`.

Old RE2 uses `re2::StringPiece` for `RE2::FullMatch()` and new RE2 provides `re2::StringPiece` as an alias of `absl::string_view`. So we can use `re2::StringPiece` for both of old and new RE2.

We can drop support for old RE2 to always use `absl::string_view` but we use `re2::StringPiece` for now. It seems that RE2 will use `std::string_view` instead of `absl::string_view` eventually. For example, https://github.com/google/re2/commit/2a029e28d3080e4718581649ba9faad5347c3ff9 is a commit to migrate to `std::optional` from `absl::optional`.

We can revisit this after RE2 migrates to `std::string_view`.

Related: GH-15124

### DIFF
--- a/scripts/setup-centos9.sh
+++ b/scripts/setup-centos9.sh
@@ -70,7 +70,7 @@ function install_build_prerequisites {
 # Install dependencies from the package managers.
 function install_velox_deps_from_dnf {
   dnf_install libevent-devel \
-    openssl-devel libzstd-devel lz4-devel double-conversion-devel \
+    openssl-devel re2-devel libzstd-devel lz4-devel double-conversion-devel \
     libdwarf-devel elfutils-libelf-devel curl-devel libicu-devel bison flex \
     libsodium-devel zlib-devel gtest-devel gmock-devel xxhash-devel
 
@@ -97,7 +97,6 @@ function install_velox_deps {
   run_and_time install_conda
   run_and_time install_gflags
   run_and_time install_glog
-  run_and_time install_re2
   run_and_time install_snappy
   run_and_time install_boost
   run_and_time install_protobuf
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -1936,10 +1936,10 @@ struct ParseDurationFunction {
   FOLLY_ALWAYS_INLINE void call(
       out_type<IntervalDayTime>& result,
       const arg_type<Varchar>& amountUnit) {
-    absl::string_view valueStr;
-    absl::string_view unit;
+    re2::StringPiece valueStr;
+    re2::StringPiece unit;
     if (!RE2::FullMatch(
-            absl::string_view(amountUnit.data(), amountUnit.size()),
+            re2::StringPiece(amountUnit.data(), amountUnit.size()),
             *durationRegex_,
             &valueStr,
             &unit)) {
@@ -1950,7 +1950,7 @@ struct ParseDurationFunction {
     double value{};
     try {
       size_t pos = 0;
-      // Create temporary string from absl::string_view for stod
+      // Create temporary string from re2::StringPiece for stod
       std::string valueString(valueStr.data(), valueStr.size());
       value = std::stod(valueString, &pos);
       if (pos != valueString.size()) {


kou:
@abhinavmuk04 Could you take a look at this? This is a follow-up PR for your PR GH-15124.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1937,7 +1937,10 @@ struct ParseDurationFunction {
     re2::StringPiece valueStr;
     re2::StringPiece unit;
     if (!RE2::FullMatch(
-            std::string_view{amountUnit}, *durationRegex_, &valueStr, &unit)) {
+            absl::string_view(amountUnit.data(), amountUnit.size()),
```
kou: Oh, old RE2 that is used by our CentOS Stream 9 based image uses `StringPiece` not `absl::string_view`/`std::string_view`...

kou: Let's use `re2::StringPiece` instead for old RE2 and new RE2.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -1937,7 +1937,12 @@ struct ParseDurationFunction {
     re2::StringPiece valueStr;
     re2::StringPiece unit;
     if (!RE2::FullMatch(
-            std::string_view{amountUnit}, *durationRegex_, &valueStr, &unit)) {
+            // We can use absl::string_view() once we require RE2
```
mbasmanova: Can we not use std::string_view? CC: @pedroerp

mbasmanova:
>  old RE2 that is used by our CentOS Stream 9 based image 

Can we update the image instead? CC: @assignUser @kgpai

assignUser:
Yeah, using a more recent re2 seems like the better solution imo.

pedroerp:
I'm surprised there is no implicit conversion between std::string_view and absl's or RE2's. Is it just because we are using old versions of these libraries?

velox/functions/prestosql/DateTimeFunctions.h
pedroerp: I'm surprised these libraries don't take a std::string_view. Is this because we are using old versions?

In general, for now we need this explicit cast (velox::StringView->std::string_view), but we are working on making this conversion implicit.

kou: We can use `std::string_view` when:
* We use old RE2 (because `re2::StringPiece` provides an implicit cast from `std::string_view`) or
* We use new RE2 and Abseil that uses `std::string` as an alias of `absl::string_view` (because `absl::string_view` does not provide an implicit cast from `std::string_view`)

If we use new RE2 and Abseil that does not use `std::string` as an alias of `absl::string_view` (e.g. Debian's Abseil package), we can not use `std::string_view`. We need to use `absl::string_view`. FYI: `re2::StringPiece` is an alias of `absl::string_view` in new RE2.

FYI: This works with new RE2:

```diff
diff --git a/velox/functions/prestosql/DateTimeFunctions.h b/velox/functions/prestosql/DateTimeFunctions.h
index 09200e41f..aa86d196a 100644
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -1936,8 +1936,7 @@ struct ParseDurationFunction {
       const arg_type<Varchar>& amountUnit) {
     re2::StringPiece valueStr;
     re2::StringPiece unit;
-    if (!RE2::FullMatch(
-            std::string_view{amountUnit}, *durationRegex_, &valueStr, &unit)) {
+    if (!RE2::FullMatch(amountUnit, *durationRegex_, &valueStr, &unit)) {
       VELOX_USER_FAIL(
           "Input duration is not a valid data duration string: {}", amountUnit);
     }
diff --git a/velox/type/StringView.h b/velox/type/StringView.h
index 5c1bc00a0..6613882fd 100644
--- a/velox/type/StringView.h
+++ b/velox/type/StringView.h
@@ -25,6 +25,8 @@
 #include <folly/Range.h>
 #include <folly/dynamic.h>
 
+#include <absl/strings/string_view.h>
+
 #include <fmt/format.h>
 
 #include "velox/common/base/BitUtil.h"
@@ -213,10 +215,17 @@ struct StringView {
   }
 
   operator std::string_view() && = delete;
-  explicit operator std::string_view() const& {
+  operator std::string_view() const& {
     return std::string_view(data(), size());
   }
 
+#ifndef ABSL_USES_STD_STRING_VIEW
+  operator absl::string_view() && = delete;
+  operator absl::string_view() const& {
+    return absl::string_view(data(), size());
+  }
+#endif
+
   const char* begin() && = delete;
   const char* begin() const& {
     return data();
```

kou:
> I'm surprised there is no implicit conversion between std::string_view and absl's or RE2's. Is it just because we are using old versions of these libraries?

No. New `RE2::FullMatch()` uses `absl::string_view` and `absl::string_view` doesn't provide implicit/explicit conversion from `std::string_view`. If new Abseil is built as C++11 compatible like Debian's Abseil package does, `std::string_view` doesn't work.

velox/functions/prestosql/DateTimeFunctions.h
mbasmanova: PR title says "Use re2::StringPiece", but the code uses "absl::string_view". Other code (2 lines above) uses re2::StringPiece. It would be nice to limit the number of different string views used in this code.

scripts/setup-centos9.sh
```diff
@@ -70,7 +70,7 @@ function install_build_prerequisites {
 # Install dependencies from the package managers.
 function install_velox_deps_from_dnf {
   dnf_install libevent-devel \
-    openssl-devel re2-devel libzstd-devel lz4-devel double-conversion-devel \
+    openssl-devel libzstd-devel lz4-devel double-conversion-devel \
```
mbasmanova: Please, document this change in the PR description.

mbasmanova:
CI is red.

velox/functions/prestosql/DateTimeFunctions.h
kou: Sorry. I forgot to update PR title/description.

I've replaced all `re2::StringPiece` with `absl::string_view`.

scripts/setup-centos9.sh
kou: Sorry. I forgot to do it. Done.

velox/functions/prestosql/DateTimeFunctions.h
mbasmanova: I see more usages of re2::StringPiece in the repo. Should we replace all of them with absl::string_view? Might be simpler to just use re2::StringPiece everywhere, no?

kou:
The "Linux release with adapters" failure was fixed by https://github.com/facebookincubator/velox/pull/15272 . I've rebased.

The "Fuzzer Jobs" failure is caused because `setup-centos9.sh` change isn't used. I think that this is a limitation of the current CI configuration. (I can work on improving it if it's needed.)

velox/functions/prestosql/DateTimeFunctions.h
kou: Oh... Let's use `re2::StringPiece` for now because:
* RE2 will use `std::string_view` not `absl::string_view` eventually because RE2 migrated to `std::optional` from `absl::optional`: https://github.com/google/re2/commit/2a029e28d3080e4718581649ba9faad5347c3ff9
* `re2::StringPiece` works with old RE2 and new RE2

We can revisit this when RE2 migrates to `std::string_view` from `absl::string_view`.

mbasmanova (approved):
Thanks.

mbasmanova:
> Old RE2 that is provided by CentOS Stream 9 doesn't accept absl::string_view.

Do you expect CentOS Stream 9 to be supported going forward? If so, we may need to add CI for it. CC: @assignUser @kgpai

MBkkt:
this PR https://github.com/facebookincubator/velox/pull/15134 already exists more than two weeks
And more than two weeks ago fixed the same issue...

kou:
Oh, sorry. I missed the PR.

PR #15260, opened 2025-10-23
build: Don't use `CMAKE_INSTALL_LIBDIR` in bundled Apache Arrow

kou:
`CMAKE_INSTALL_LIBDIR` is defined by `GNUInstallDirs` but we don't include it. So it's not defined.

If it's not defined, the following error may be occurred:

```text
: && /bin/clang++-17 ... -shared -Wl,-soname,libvelox.so -o lib/libvelox.so ...  CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a ... && :
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
clang++-17: error: no such file or directory: 'CMake/resolve_dependency_modules/arrow/arrow_ep/install//libarrow.a'
```


kou:
@majetideepak Could you also review this?

PR #14564, opened 2025-08-21
build: Use the official Snappy CMake package

kou:
Other dependencies such as Apache Arrow may use system Snappy CMake package. If we use custom Snappy find module, it may be conflict with system Snappy CMake package.

Snappy 1.1.5 that was released on 2017-06-29 or later ships the CMake
package. So we don't need our find module.

Fixes #14562

PR #15262, opened 2025-10-23
fix: Don't try installing `FindSnappy.cmake`

kou:
GH-14564 removed `FindSnappy.cmake` because it's not necessary.

But our CMake package related code always tries installing `FindSnappy.cmake`:

```text
CMake Error at velox/buffer/cmake_install.cmake:93 (file):
  file INSTALL cannot find
  "CMake/FindSnappy.cmake": No such file or directory.
  velox/cmake_install.cmake:47 (include)
  cmake_install.cmake:52 (include)
```


majetideepak:
@HuamengJiang Done. Thanks for your help!

PR #15267, opened 2025-10-23
fix: HashJoins and HashAggregations may produce incorrect results on types that support custom comparison

kevinwilfong:
Summary:
HashJoins and HashAggregations use HashTable to track values. The HashTable can use 3 
different hash modes: hash, array, normalizedKey.  For the latter two, it relies on 
VectorHasher to map values to a unique int64 (unlike hash this can't have collisions). 
VectorHasher decides if it is able to do this based on the TypeKind (physical type).

This leads to bugs when the logical type supports custom comparison. In this case, the 
TypeKind may indicate that value IDs can be produced, but the logic for doing this mapping 
may map two different values which are equal using the Type's custom comparison to 
different int64s at which point we lose the ability to identify equivalent values.

Furthermore, in HashJoin there is an additional issue with hash mode, where when it hits a 
collision or does a lookup in the HashTable, it calls RowContainer's equals method which is 
missing support for custom comparison. So values that should be equal so again we lose the 
ability to identify equivalent values.

This change fixes these issues by:
1) In VectorHasher I changed typeSupportsValueIds to return false if the type supports 
custom comparison
2) In VectorHasher I updated the constructor to explicitly disable range and unique value 
(array/normalizedKey) based hashes if the type supports custom comparison.
3) I updated all functions in VectorHasher that rely on value IDs to do a DCHECK that the 
type supports value IDs or handle them if possible. DCHECK because the code is 
performance sensitive.
4) I removed the equals function from RowContainer. This was only used in one place in the 
code and the logic is equivalent to compare == 0. It did have one optimization to skip null 
checks if we know there are no nulls which I ported to compare.

Note that I removed the special handling for int128_t in VectorHasher's analyzeValue. That 
was compensating for the issues I identify here in some cases, but did not fix the general 
issue with types that are physically 128 bit integers being used as value IDs.  That fix is 
coming separately.

Differential Revision: D85291290

### DIFF
--- a/velox/exec/RowContainer.h
+++ b/velox/exec/RowContainer.h
@@ -669,18 +669,11 @@ class RowContainer {
 #endif
   void setProbedFlag(char** rows, int32_t numRows);
 
-  /// Returns true if 'row' at 'column' equals the value at 'index' in
-  /// 'decoded'. 'mayHaveNulls' specifies if nulls need to be checked. This is a
-  /// fast path for compare().
-  template <bool mayHaveNulls>
-  bool equals(
-      const char* row,
-      RowColumn column,
-      const DecodedVector& decoded,
-      vector_size_t index) const;
-
   /// Compares the value at 'column' in 'row' with the value at 'index' in
   /// 'decoded'. Returns 0 for equal, < 0 for 'row' < 'decoded', > 0 otherwise.
+  /// 'mayHaveNulls' specifies if nulls need to be checked. This is a fast path
+  /// for compare().
+  template <bool mayHaveNulls = true>
   int32_t compare(
       const char* row,
       RowColumn column,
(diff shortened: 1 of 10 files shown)


velox/exec/RowContainer.h
```diff
@@ -671,18 +671,11 @@ class RowContainer {
   void
   setProbedFlag(char** rows, int32_t numRows);
 
-  /// Returns true if 'row' at 'column' equals the value at 'index' in
-  /// 'decoded'. 'mayHaveNulls' specifies if nulls need to be checked. This is a
-  /// fast path for compare().
-  template <bool mayHaveNulls>
-  bool equals(
```
MBkkt, line 676: Hello, for types like strings equals can be implemented a lot of more efficient than compare == 0, does velox and your PR account this for
1) varchar/varbinary
2) custom types

kevinwilfong: Hi, yes, there is a set of options we pass to compare via the CompareFlags struct, one of those options indicates that although we're doing a comparison we really only care about equality, in which case it will exit early in the case of strings, complex types, and custom types based on these physical types as soon as it detects inequality.

This is the
```CompareFlags::equality(CompareFlags::NullHandlingMode::kNullAsValue))```
I pass to compare in the line where I replaced the call to equals in HashTable.cpp

In fact, the old equals function I deleted (see equalsNoNulls which it eventually calls) was still implemented using compare == 0, but without that flag, so this should be an improvement.

Comment (MBkkt) [reply velox/exec/RowContainer.h:676]: I understand, thanks!

It's not related to your PR but maybe you know, why such design (with compare == 0 + flags) was chosen instead of separate equals method?

Comment (MBkkt) [reply velox/exec/RowContainer.h:676]: For an example

```
  // If equalsOnly and maps have different sizes, we can bail fast.
  if (flags.equalsOnly && leftIndices.size() != rightIndices.size()) {
    int result = leftIndices.size() - rightIndices.size();
    return flags.ascending ? result : result * -1;
  }
```

here in case of equalsOnly we still tries to distinguish -1 and 1, although it doesn't really matter

Comment (MBkkt) [reply velox/exec/RowContainer.h:676]: Hmm, funny, I just checked random place and found this, but in other place code written at least like this
```
  if (flags.equalsOnly &&
      rawSizes_[index] != otherArray->rawSizes_[wrappedOtherIndex]) {
    return 1;
  }
```

kevinwilfong:
Addressed a couple lint warnings and rebased to pick up the fix for the broken CI

velox/exec/RowContainer.h
kevinwilfong: > It's not related to your PR but maybe you know, why such design (with compare == 0 + flags) was chosen instead of separate equals method?

This predates me, but my guess is code consolidation.  E.g. BaseVector supports both compare and equalValueAt and equalValueAt just reuses the compare function and handles setting the flag and comparing to 0 for you. This could have been done in a private helper function to avoid it leaking to the user.

In APIs that aren't user facing, we typically see only compare, e.g. ContainerRowSerde which is what's ultimately used here.

kevinwilfong:
Fixed RowContainerTest.unknown, it was creating a RowContainer with just NULL, then calling equals (now compare) with mayHaveNulls = false.

kgpai (approved):

PR #15268, opened 2025-10-24
feat: Implement aggregation pushdown feature in TableEvolutionFuzzer

juwentus1234:
Summary:
Add the feature of aggregation pushdown in TableEvolutionFuzzer

Actual Path: with pushdown
Expected Path: without pushdown

Differential Revision: D85380537

### DIFF
--- a/velox/exec/tests/TableEvolutionFuzzer.cpp
+++ b/velox/exec/tests/TableEvolutionFuzzer.cpp
@@ -48,6 +48,13 @@ DEFINE_bool(
     "up after failures. Therefore, results are not compared when this is "
     "enabled. Note that this option only works in debug builds.");
 
+DEFINE_int32(
+    aggregation_pushdown_frequency,
+    5,
+    "Controls the frequency of aggregation pushdown. The aggregation pushdown "
+    "is enabled with probability 1/N where N is this value. For example, "
+    "N=5 means 20% chance, N=2 means 50% chance.");
+
 namespace facebook::velox::exec::test {
 
 std::ostream& operator<<(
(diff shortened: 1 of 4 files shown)


velox/exec/tests/utils/PlanBuilder.h
```diff
@@ -35,9 +35,19 @@ enum class Table : uint8_t;
...
+};
+
 struct PushdownConfig {
   common::SubfieldFilters subfieldFiltersMap;
   std::string remainingFilter;
+  // Aggregation pushdown configuration
+  std::optional<AggregationConfig> aggregationConfig;
+  // If true, insert an identity Project node before aggregation to block
+  // pushdown
+  bool insertProjectToBlockPushdown = false;
```
Yuhta: We don't need this to be part of the config, just pass it as a separate parameter to `makeScanTask`

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -401,6 +401,84 @@ fuzzer::ExpressionFuzzer::FuzzedExpressionData generateRemainingFilters(
   return expressionFuzzer.fuzzExpressions(1);
 }
 
+// Generate random aggregation configuration for pushdown testing.
+// Only generates aggregations that are eligible for pushdown:
+// - Supported aggregate functions: min, max, sum
```
Yuhta, line 413: We maybe want to add others like bitwise AND/OR but it can be a different PR

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -627,22 +705,46 @@ void TableEvolutionFuzzer::run() {
...
         generatedRemainingFilters,
         columnNameMapping,
-        pushownConfig,
+        pushdownConfig,
         subfieldFilteredFields);
   }
 
+  // Enable aggregation testing (50% chance)
+  std::optional<AggregationConfig> aggConfig;
+  bool shouldTestAggregation = folly::Random::oneIn(2, rng_);
```
Yuhta: Let's make the probability a gflag and default to something like 0.2 (aggregation pushdown is relatively rare in prod so we don't need to test that so often as 0.5)

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -1064,17 +1166,38 @@ std::unique_ptr<TaskCursor> TableEvolutionFuzzer::makeScanTask(
...
-              newSchemaUsingStructReadingFlatMap, // Use struct schema for
-                                                  // flatmap reading
-              /*pushdownConfig=*/pushdownConfig,
-              tableSchema, // Original schema as dataColumns
-              {})
-          .planNode();
+
+  auto builder = PlanBuilder()
+                     .filtersAsNode(useFiltersAsNode)
+                     .tableScanWithPushDown(
```
Yuhta, line 1205: For aggregation pushdown to kick in, we cannot have filter on the aggregated columns, we should remove the subfield filters and remaining filters on these columns (or not to generate them in the first place).

velox/exec/tests/TableEvolutionFuzzer.cpp
```diff
@@ -628,23 +716,60 @@ void TableEvolutionFuzzer::run() {
...
+      // Check if the column name appears in the remaining filter
+      if (pushdownConfig.remainingFilter.find(name) != std::string::npos) {
+        allFilteredColumns.insert(name);
+      }
+    }
+  }
+
+  // Enable aggregation testing (20% chance)
+  std::optional<AggregationConfig> aggConfig;
+  bool shouldTestAggregation = folly::Random::oneIn(5, rng_);
```
Yuhta: Let's make a gflag for this default to 0.2

Yuhta (approved):
I think we are almost there, did you double check the aggregate pushdown got actually triggered during the run?  Specifically this line should be hit: https://github.com/facebookincubator/velox/blob/b3953b0bdfba6572b6303ccda793e537306245e6/velox/exec/AggregationHook.h#L234

PR #15270, opened 2025-10-24
fix: Wrong results in parquet due to error in column skipping in some cases

jaystarshot:
This PR fixes a critical bug in `SelectiveStructColumnReader::isChildMissing()` that causes filter columns to be incorrectly skipped when reading Parquet files, resulting in queries returning 0 rows instead of the expected results.

## Problem

When a column is:
1. Used in a filter expression (e.g., `WHERE cost_category LIKE '%Cloud%'`)
2. NOT included in the query output (e.g., `SELECT forecast_capacity_cost, monthstr, report_month`)
3. Part of a query that includes partition columns (report_month)

The `isChildMissing()` function incorrectly marks the filter column as "missing" and skips reading it from the Parquet file. This causes the filter to evaluate against NULL values, returning 0 rows instead of the filtered results.


velox/dwio/common/SelectiveStructColumnReader.cpp
```diff
@@ -514,21 +514,25 @@ void SelectiveStructColumnReaderBase::recordParentNullsInChildren(
...
+  if (isRoot_ || // If we're in the root struct, missing fields are handled
+                 // elsewhere
+      childSpec.channel() ==
+          velox::common::ScanSpec::kNoChannel || // Filter-only field
+      fileType_->type()->kind() ==
+          TypeKind::MAP) { // Flat map can't have missing fields
+    return false;
+  }
+
+  // Check if the field exists in the file schema by name, not by channel
```
Yuhta, line 528: Output schema is a super set of file schema in case of non-root field.  Currently the only data warehouse evolution mode supported in Velox is matching by position and appending to the end of struct, so I don't think there is anything wrong with the old code.

In your unit test I don't even see nested struct being used, so this path should not even be triggered.  There is probably something wrong in parquet reader that causes `isRoot_` becomes false by mistake.

jaystarshot: @Yuhta Thanks for the quick review!

> Output schema is a super set of file schema in case of non-root field.

In this case the output schema has the column **report_month** which is not a part of the file schema (its marked as constant / (partition column). 

I added some debug logs around this function for this test case and isRoot is is indeed 0 for the isChildMissing function calls.


```
=== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: forecast_capacity_cost
  channel: 0
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('forecast_capacity_cost'): 1
  OLD BUG LOGIC (channel >= size): 0
  NEW FIX LOGIC (!containsChild): 0

=== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: monthstr
  channel: 1
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('monthstr'): 1
  OLD BUG LOGIC (channel >= size): 0
  NEW FIX LOGIC (!containsChild): 0
=
== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: cost_category
  channel: 3
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('cost_category'): 1
  OLD BUG LOGIC (channel >= size): 1
  NEW FIX LOGIC (!containsChild): 0
  *** BUG DEMONSTRATED! Old logic disagrees with new logic ***


=== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: cost_category
  channel: 3
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('cost_category'): 1
  OLD BUG LOGIC (channel >= size): 1
  NEW FIX LOGIC (!containsChild): 0
  *** BUG DEMONSTRATED! Old logic disagrees with new logic ***


=== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: forecast_capacity_cost
  channel: 0
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('forecast_capacity_cost'): 1
  OLD BUG LOGIC (channel >= size): 0
  NEW FIX LOGIC (!containsChild): 0


=== isChildMissing DEBUG ===
  isRoot_: 0
  fieldName: monthstr
  channel: 1
  fileType size: 3
  fileType: ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE>
  containsChild('monthstr'): 1
  OLD BUG LOGIC (channel >= size): 0
  NEW FIX LOGIC (!containsChild): 0
```

but maybe i am missing something?

jaystarshot:
@Yuhta can you please re-review, the test case clearly shows the issue. It could be one of the special cases and this could be happening in dwrf too

velox/dwio/common/SelectiveStructColumnReader.cpp
Yuhta: `isRoot_` should be set to true in this case (some bug in parquet reader), that's why the result was not correct

jaystarshot: I see thanks! for velox-parquet the root field type is always a struct  ROW<cost_category:VARCHAR,monthstr:VARCHAR,forecast_capacity_cost:DOUBLE> from logs in this case. So should we override this function in parquet reader for now?

Yuhta: In DWRF we set `isRoot_` to true when creating the root level column reader: https://github.com/facebookincubator/velox/blob/2b51a3cdd081ff6b9a729a86996edc664d7b2c40/velox/dwio/dwrf/reader/DwrfReader.cpp#L175

PR #15273, opened 2025-10-24
docs(geo): Add documentation for bingtile and geometry types

jagill:
Differential Revision: D85431595


mbasmanova:
> As title

@jagill Please, avoid descriptions like this. They do not provide any useful information. Leave description empty if title fully describes the changes.

PR #15276, opened 2025-10-24
misc: Revert std::move in Driver getOutput

natashasehgal:
Revert no-op change based on https://github.com/facebookincubator/velox/pull/15089#discussion_r2418317391 as it does not lead to any performance improvement

Differential Revision: D85445121


pedroerp (approved):
Thanks @natashasehgal

Issue #15050, opened 2025-10-05
Add annotation output to license-header.py


assignUser:
### Description

It should be simple enough to emit an error annotation when running in GHA: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message

PR #15284, opened 2025-10-25
build(ci): Run clang-tidy and add license header annotations

assignUser:
It might seems like overkill to use the adapters image but for clang-tidy to run it needs `compile-commands.json`, so we need to run cmake, which needs the dependencies.

I use `--line-filter` to only output warnings on changed lines, other warnings would not show up as annotations anyway and take very long.

The clang-tidy job only runs when there are changes to C++ files. 

Closes: #15047 
Closes: #15050

### DIFF
--- a/scripts/checks/run-clang-tidy.py
+++ b/scripts/checks/run-clang-tidy.py
@@ -80,9 +80,17 @@ def tidy(args):
     lines = f"'--line-filter={line_filter}'" if args.commit is not None else ""
 
     ok = True
-    build_path = f"-p {args.p}" if args.p else ""
+    build_path = args.p or os.getenv("BUILD_PATH")
+    build_path_str = f"-p {build_path}" if build_path else ""
+
+    if build_path_str == "" and not os.path.isfile(
+        os.getcwd().join("compile_commands.json")
+    ):
+        print("compile_commands.json not found, skipping clang-tidy")
+        return 0
+
     status, stdout, stderr = util.run(
-        f"xargs clang-tidy --format-style=file -header-filter='.*' --quiet {build_path} {fix} {lines}",
+        f"xargs clang-tidy --format-style=file -header-filter='.*' --quiet {build_path_str} {fix} {lines}",
         input=filtered_files,
     )
 


.github/workflows/preliminary_checks.yml
```diff
@@ -38,13 +38,104 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: Get changed files
+        id: changes
+        env:
+          GH_TOKEN: ${{ github.token }}
+          BASE_REF: ${{ github.base_ref }}
+          HEAD_REF: ${{ github.head_ref }}
+          PR_OWNER: ${{ github.event.pull_request.head.repo.owner.login }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
```
assignUser, line 60: note: investigate how this behaves on push/other than pull_request

assignUser: nvm we only run this on PR

.github/workflows/preliminary_checks.yml
```diff
@@ -38,13 +38,104 @@ jobs:
...
+    container: ghcr.io/facebookincubator/velox-dev:adapters
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      - run: uv tool install clang-tidy==18.1.8
+
+      - name: Configure Maximal Build
```
majetideepak, line 96: Why do we need to configure the build?

assignUser: clang-tidy needs the compilation database ('compile_commands.json')

scripts/checks/run-clang-tidy.py
```diff
@@ -15,75 +15,12 @@
...
 import argparse
 import json
-import regex
+import re
 import sys
+import os
 
 import util
 
-CODE_CHECKS = """*
```
czentgr, line 23: For my understanding, this is replaced by the regex pattern that looks for warnings/errors?

assignUser: This is replaced by our .clang-tidy file.

czentgr (approved):
Thanks!

scripts/checks/license-header.py
```diff
@@ -228,7 +229,13 @@ def main():
             # If removing the header text, zero it out there.
             header_comment = ""
 
-        message(log_to, "Fix  : " + filepath)
+        if os.environ.get("GITHUB_ACTIONS"):
```
czentgr, line 232: I didn't realize this but this change causes build failures in PrestoC++. Namely, it re-generates the protocol in github and this adds the new error message to the output causing compile errors.
We do a regenerate in the CI to try and ensure java-cpp protocol compatibility.

We call "remove" explicitly.

PR #15291, opened 2025-10-27
feat: Add RowVector::sortIndicesByKeys

duanmeng:
Add a `RowVector::sortIndicesByKeys` API that sorts a `RowVector` by specified key columns
and makes the corresponding sort indices.

### DIFF
--- a/velox/vector/ComplexVector.h
+++ b/velox/vector/ComplexVector.h
@@ -231,6 +231,17 @@ class RowVector : public BaseVector {
 
   VectorPtr slice(vector_size_t offset, vector_size_t length) const override;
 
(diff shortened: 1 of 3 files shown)


velox/vector/ComplexVector.h
```diff
@@ -231,6 +231,11 @@ class RowVector : public BaseVector {
 
   VectorPtr slice(vector_size_t offset, vector_size_t length) const override;
 
+  void sortIndicesByKeys(
```
pedroerp, line 240: @duanmeng could you document this method?

duanmeng: Fixed

PR #15294, opened 2025-10-27
docs: Move 'task_splits_count' to 'Task Execution' category

lingbin:
The "task_splits_count" metric is currently categorized as 
"Memory Management" in the documentation and code. This is incorrect; 
it should be in the "Task Execution" category.


lingbin:
image

PR #15293, opened 2025-10-27
fix: Update clang-format to 21.1.2 (from 18.1.3) in CI

aditi-pandit:

PR #15295, opened 2025-10-27
fix: Update clang-format in pre-commit to fix broken CI

MBkkt:
reason: https://github.com/facebookincubator/velox/commit/21af185832a129f6be1da09ea1c44c856739c5aa

same: https://github.com/facebookincubator/axiom/pull/562


mbasmanova:
Thanks, but @aditi-pandit has already created #15293

MBkkt:
@mbasmanova I see, I will close this in such case

PR #15299, opened 2025-10-28
feat: Add s3 OCI endpoint region inference when region is not specified in the config

minhancao:
Add s3 OCI endpoint region inference when region is not provided in the config. Also sets path-style-access=true for when OCI endpoint is detected.

The Oracle Cloud Infrastucture's (OCI) S3 object storage requires the region to be provided and set path-style-access=true in order for it to work. 

Currently, there is only implementation for inferring region from the endpoint for AWS S3 when region is not provided in the config. This PR will implement region inference from endpoint for OCI S3 just like how it's inferred for AWS when region is not explicitly provided in the config.

### DIFF
--- a/velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
+++ b/velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
@@ -70,6 +70,15 @@ S3Config::S3Config(
       }
     }
   }
+
+  // Check if using OCI and not AWS.
+  // If OCI is used, path-style-access must be set to true.
+  auto endpointValue = endpoint();
+  if (endpointValue.has_value() &&
+      endpointValue.value().find(kOCIHostSuffix) != std::string::npos) {
+    config_[Keys::kPathStyleAccess] = "true";
+  }
+
   payloadSigningPolicy_ =
       properties->get<std::string>(kS3PayloadSigningPolicy, "Never");
 }
@@ -77,11 +86,11 @@ S3Config::S3Config(
 std::optional<std::string> S3Config::endpointRegion() const {
   auto region = config_.find(Keys::kEndpointRegion)->second;
   if (!region.has_value()) {
-    // If region is not set, try inferring from the endpoint value for AWS
-    // endpoints.
+    // If region is not set, try inferring from the endpoint value for AWS or
+    // OCI endpoints.
     auto endpointValue = endpoint();
     if (endpointValue.has_value()) {
-      region = parseAWSStandardRegionName(endpointValue.value());
+      region = parseStandardRegionName(endpointValue.value());
     }
   }
   return region;
--- a/velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
+++ b/velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
@@ -149,36 +149,59 @@ std::optional<folly::Uri> S3ProxyConfigurationBuilder::build() {
   return proxyUri;
 }
 
-std::optional<std::string> parseAWSStandardRegionName(
-    std::string_view endpoint) {
-  // The assumption is that the endpoint ends with
-  // ".amazonaws.com" or ".amazonaws.com/". That means for AWS we don't
-  // expect a port in the endpoint.
-  const std::string_view kAmazonHostSuffix = ".amazonaws.com";
-  auto index = endpoint.size() - kAmazonHostSuffix.size();
-  // Handle the case where the endpoint ends in a trailing slash.
-  if (endpoint.back() == '/') {
-    index--;
-  }
-  if (endpoint.rfind(kAmazonHostSuffix) != index) {
+bool isAWSEndpoint(const std::string_view& endpoint) {
+  return endpoint.find(kAWSHostSuffix) != std::string_view::npos;
+}
+bool isOCIEndpoint(const std::string_view& endpoint) {
+  return endpoint.find(kOCIHostSuffix) != std::string_view::npos;
+}
+
+std::optional<std::string> parseStandardRegionName(std::string_view endpoint) {
+  std::string_view hostSuffix;
+  if (isAWSEndpoint(endpoint)) {
+    hostSuffix = kAWSHostSuffix;
+  } else if (isOCIEndpoint(endpoint)) {
+    hostSuffix = kOCIHostSuffix;
+  } else {
     return std::nullopt;
   }
-  // Remove the kAmazonHostSuffix.
-  std::string_view endpointPrefix = endpoint.substr(0, index);
-  const re2::RE2 pattern("^(?:.+\\.)?s3[-.]([a-z0-9-]+)$");
+
+  if (endpoint.empty()) {
+    return std::nullopt;
+  }
+
+  // Remove trailing slash.
+  if (endpoint.back() == '/') {
+    endpoint.remove_suffix(1);
+  }
+
+  // Ensure endpoint ends with the correct suffix.
+  if (endpoint.size() <= hostSuffix.size() || !endpoint.ends_with(hostSuffix)) {
+    return std::nullopt;
+  }
+
+  // Strip the host suffix.
+  endpoint.remove_suffix(hostSuffix.size());
+
+  static const re2::RE2 awsPattern(R"(^(?:.+\.)?s3[-.]([a-z0-9-]+)$)");
+  static const re2::RE2 ociPattern(
+      R"(^(?:.+\.)?objectstorage(?:\.[a-z0-9-]+)*\.([a-z0-9-]+)$)");
+
   std::string region;
-  if (re2::RE2::FullMatch(endpointPrefix, pattern, &region)) {
-    // endpointPrefix is 'bucket.s3-[region]' or 'bucket.s3.[region]'
+  const re2::RE2& pattern =
+      (hostSuffix == kAWSHostSuffix) ? awsPattern : ociPattern;
+
+  if (re2::RE2::FullMatch(endpoint, pattern, &region)) {
     return region;
   }
 
-  index = endpointPrefix.rfind('.');
-  if (index != std::string::npos) {
-    // endpointPrefix was 'service.[region]'.
-    return std::string(endpointPrefix.substr(index + 1));
+  // Fallback: return the region that should be after the rightmost '.'.
+  // Ex: "foo.a3-reg.amazonaws.com" -> "foo.a3-reg" -> "a3-reg"
+  auto lastDot = endpoint.rfind('.');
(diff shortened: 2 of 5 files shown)


velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
```diff
@@ -70,6 +70,15 @@ S3Config::S3Config(
       }
     }
   }
+
+  // Check if using OCI and not AWS.
+  // If OCI is used, path-style-access must be set to true.
+  auto endpointValue = endpoint();
+  if (endpointValue.has_value() &&
+      endpointValue.value().find(".oraclecloud.com") != std::string::npos) {
```
czentgr: Let's make `.oraclecloud.com` and `.amazonaws.com` constants to be used in the various places.

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
```diff
@@ -81,7 +90,20 @@ std::optional<std::string> S3Config::endpointRegion() const {
     // endpoints.
     auto endpointValue = endpoint();
     if (endpointValue.has_value()) {
-      region = parseAWSStandardRegionName(endpointValue.value());
+      // Check if endpoint is a full AWS endpoint (e.g.
+      // "s3.us-west-2.amazonaws.com").
+      if (endpointValue.value().find(".amazonaws.com") != std::string::npos) {
+        region = parseAWSStandardRegionName(endpointValue.value());
+        // Check if endpoint is an OCI endpoint (e.g.
```
czentgr: Comment is in wrong place. We probably also don't need this comment here.
Instead, please update the comment on line 80/81 and make it generic for the code block.

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
```diff
@@ -81,7 +90,20 @@ std::optional<std::string> S3Config::endpointRegion() const {
...
+        // "objectstorage.us-ashburn-1.oraclecloud.com").
+      } else if (
+          endpointValue.value().find(".oraclecloud.com") != std::string::npos) {
+        region = parseOCIStandardRegionName(endpointValue.value());
+
+      } else {
+        // Otherwise, assume it's a standard AWS region name
+        region = parseAWSStandardRegionName(endpointValue.value());
```
czentgr: Return `std::nullopt`. We don't have to call the AWS parsing to return the same.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -182,4 +182,46 @@ std::optional<std::string> parseAWSStandardRegionName(
   return std::nullopt;
 }
 
+std::optional<std::string> parseOCIStandardRegionName(
+    std::string_view endpoint) {
+  const std::string_view kOCIHostSuffix = ".oraclecloud.com";
```
czentgr: See above comment on constants. These could be defined in the util header.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -182,4 +182,46 @@ std::optional<std::string> parseAWSStandardRegionName(
...
+
+  // Handle trailing slash
```
czentgr: Note, comments are full sentences and can be omitted if obvious in what the code does.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -182,4 +182,46 @@ std::optional<std::string> parseAWSStandardRegionName(
...
+  // Check that the suffix is at the end
+  size_t suffixStart = index - kOCIHostSuffix.size();
+  if (endpoint.substr(suffixStart, kOCIHostSuffix.size()) != kOCIHostSuffix) {
+    return std::nullopt;
+  }
+
+  // Remove the suffix
+  std::string_view endpointPrefix = endpoint.substr(0, suffixStart);
+
+  // Extract region after last dot
```
czentgr: i.e.. we have `service.[region]` or `namespace.compat.objectstorage.[region]` left (for the comment).

This is where the code diverges between AWS and OCI. I wonder if we could compress the algorithm into a single function as it is mostly the same. The difference being the starting point of the region. 
Maybe a template with the type as template argument and constexpr how to find the start based on the type?

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
minhancao: Ok will make these constants in S3Util.h since S3Config.cpp and S3Util.cpp will both use them.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
pramodsatya: +1, can be combined to single function `parseRegionName`

majetideepak:
@minhancao can you add some details about OCI and describe the issues that are being addressed here?

minhancao:
@majetideepak Added some more details to the PR body on top.

czentgr:
@minhancao Please address the build issues.

velox/connectors/hive/storage_adapters/s3fs/S3Util.h
```diff
@@ -32,6 +32,9 @@
 
 namespace facebook::velox::filesystems {
 
+const std::string_view kAWSHostSuffix{".amazonaws.com"};
```
majetideepak: constexpr?

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
```diff
@@ -70,18 +70,27 @@ S3Config::S3Config(
       }
     }
   }
+
+  // Check if using OCI and not AWS.
+  // If OCI is used, path-style-access must be set to true.
+  auto endpointValue = endpoint();
+  if (endpointValue.has_value() &&
+      endpointValue.value().find(kOCIHostSuffix) != std::string::npos) {
```
majetideepak, line 78: We should check if it ends with the suffix?

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -149,36 +149,45 @@ std::optional<folly::Uri> S3ProxyConfigurationBuilder::build() {
...
-  if (re2::RE2::FullMatch(endpointPrefix, pattern, &region)) {
-    // endpointPrefix is 'bucket.s3-[region]' or 'bucket.s3.[region]'
-    return region;
-  }
+  const re2::RE2& s3Pattern =
+      (hostSuffix == kAWSHostSuffix) ? awsPattern : ociPattern;
 
-  index = endpointPrefix.rfind('.');
-  if (index != std::string::npos) {
-    // endpointPrefix was 'service.[region]'.
```
majetideepak, line 177: Is this case handled?

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
majetideepak: We can add util functions `isOCIEndpoint` and `isAWSEndpoint`.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -149,36 +149,45 @@ std::optional<folly::Uri> S3ProxyConfigurationBuilder::build() {
...
-    return std::string(endpointPrefix.substr(index + 1));
+  if (RE2::FullMatch(endpoint, s3Pattern, &region)) {
```
czentgr: Lets keep the `re2` namespace for the function.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
minhancao: No, adding the fallback implementation back.

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -149,36 +149,59 @@ std::optional<folly::Uri> S3ProxyConfigurationBuilder::build() {
...
+  std::string_view hostSuffix;
+  if (isAWSEndpoint(endpoint)) {
+    hostSuffix = kAWSHostSuffix;
+  } else if (isOCIEndpoint(endpoint)) {
+    hostSuffix = kOCIHostSuffix;
+  } else {
+    return std::nullopt;
+  }
+
+  if (endpoint.empty()) {
```
czentgr, line 169: This block is not needed here. If it was empty it returned in line 166 as neither would match.

So if we want a shortcut move it to the top.

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
czentgr: And then use the utility function here as well :)

velox/connectors/hive/storage_adapters/s3fs/S3Util.cpp
```diff
@@ -149,36 +149,59 @@ std::optional<folly::Uri> S3ProxyConfigurationBuilder::build() {
...
-std::optional<std::string> parseAWSStandardRegionName(
-    std::string_view endpoint) {
-  // The assumption is that the endpoint ends with
-  // ".amazonaws.com" or ".amazonaws.com/". That means for AWS we don't
-  // expect a port in the endpoint.
-  const std::string_view kAmazonHostSuffix = ".amazonaws.com";
-  auto index = endpoint.size() - kAmazonHostSuffix.size();
-  // Handle the case where the endpoint ends in a trailing slash.
+bool isAWSEndpoint(const std::string_view& endpoint) {
+  return endpoint.find(kAWSHostSuffix) != std::string_view::npos;
```
czentgr, line 153: `ends_with()`? That is safer and available.

velox/connectors/hive/storage_adapters/s3fs/S3Config.cpp
czentgr: But we shouldn't do this here. We need to document it however. We need to add documentation. There is documentation for this option. We can add some

minhancao: @czentgr I'm not sure what you mean, can you clarify pls?

Did you mean to have the setting of path-style-access=true for OCI S3 storage somewhere else and not here?

And also document it somewhere that path-style-access must be set to true for OCI S3 storage?

PR #15302, opened 2025-10-28
docs: Fix 'Avg' type name in metrics.rst

lingbin:
According to the documentation, "Average" is not a metric type, 
it should be "Avg".

No functional change.


tanjialiang (approved):
LGTM! Thanks for the fix!

PR #15306, opened 2025-10-28
feat: Upgrade simdjson dependency to 4.1.0

kevyang:
Summary:
allow-large-files

This commit was generated using `mgt import`.
pristine code for third-party libraries:
third-party/simdjson

uuid_cc2796e6e0604da2800cc1275df8ef85

update simdjson requested as we were running into ubsan errors in aarch64. i have validated that they go away with this change.

Differential Revision: D85209914


kgpai:
@czentgr Looks like all the docker images were built successfully - Are you asking for building velox on an image with this version of simdjson baked in ?

czentgr:
> @czentgr Looks like all the docker images were built successfully - Are you asking for building velox on an image with this version of simdjson baked in ?

@kgpai Yes, that's right. Like we do for most other dependency upgrades. If we are comfortable skipping it that would be ok. As you said this is a header only library so we should have seen issues in the CI if there were any.

velox/functions/prestosql/tests/JsonCastTest.cpp
```diff
@@ -1122,12 +1122,12 @@ TEST_F(JsonCastTest, toDouble) {
       JSON(),
       DOUBLE(),
       {"Infinity"_sv},
-      "The JSON document has an improper structure");
+      "INCORRECT_TYPE: The JSON element does not have the requested type.");
```
czentgr: VELOX_ASSERT_THROW does a string find on the resulting error message so it doesn't need to fully match. Other error checks don't include the actual error type and just check the text:
```
The JSON element does not have the requested type
```
should be sufficient and would match other tests such as on line 1132ff.

I suggest to remove the added type error name at the beginning and focus only on the text following that.

velox/functions/prestosql/tests/JsonFunctionsTest.cpp
```diff
@@ -310,9 +310,11 @@ TEST_F(JsonFunctionsTest, jsonParse) {
       R"([{"response":"[\"fusil a peinture\",\"�uD83E\\uDE2Dau bois\"]"}])");
 
   VELOX_ASSERT_THROW(
-      jsonParse(R"({"k1":})"), "The JSON document has an improper structure");
+      jsonParse(R"({"k1":})"),
+      "INCORRECT_TYPE: The JSON element does not have the requested type.");
   VELOX_ASSERT_THROW(
-      jsonParse(R"({:"k1"})"), "The JSON document has an improper structure");
+      jsonParse(R"({:"k1"})"),
+      "TAPE_ERROR: The JSON document has an improper structure: missing or superfluous commas, braces, missing keys, etc.  This is a fatal and unrecoverable error.");
```
czentgr: This error should have matched with the "he JSON document has an improper structure" previous test as it is a substring in the message.

czentgr (approved):
Thanks!

czentgr:
@kevyang I suppose you'll re-import the new commit?

kevyang:
> @kevyang I suppose you'll re-import the new commit?

Thanks for the review! Are you referring to our internal velox repo? If so, I've updated the proposed change there already.

czentgr:
@kevyang Yes, indeed. But good if you made the change there already! Then we can tag it as ready-to-merge?

PR #15308, opened 2025-10-28
fix: Remove resolveAggregateFunction API

kKPulla:
Summary:
Removed the existing `resolveAggregateFunction` API that returned both result and intermediate types as a pair in favor of the new split APIs:
- `resolveResultType`: Returns only the final type
- `resolveIntermediateType`: Returns only the intermediate type

Differential Revision: D85456269


peterenescu (approved):

rui-mo (approved):
Thanks!

Issue #15304, opened 2025-10-28
Flaky test S3FileSystemRegistrationTest.fileHandle


jinchengchenghh:
### Bug description

CI failed in this PR https://github.com/facebookincubator/velox/pull/15020, but is not related to this PR.
```
[ RUN      ] S3FileSystemRegistrationTest.fileHandle
ERROR Unable to initialize console server: Specified port is already in use
      > Please ensure no other program uses the same address/port
E20251023 16:35:46.941802 81998 Exceptions.h:53] Line: /__w/velox/velox/velox/connectors/hive/storage_adapters/s3fs/S3ReadFile.cpp:72, Function:initialize, Expression:  Failed to get metadata for S3 object due to: 'Network connection'. Path:'s3://data3/test.txt', SDK Error Type:99, HTTP Status Code:-1, S3 Service:'Unknown', Message:'curlCode: 7, Couldn't connect to server', RequestID:''., Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
C++ exception with description "Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Failed to get metadata for S3 object due to: 'Network connection'. Path:'s3://data3/test.txt', SDK Error Type:99, HTTP Status Code:-1, S3 Service:'Unknown', Message:'curlCode: 7, Couldn't connect to server', RequestID:''.
Retriable: False
Function: initialize
/__w/velox/velox/velox/connectors/hive/storage_adapters/s3fs/S3ReadFile.cpp
Line: 72
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox11filesystems10S3ReadFile4Impl10initializeERKNS1_11FileOptionsE
# 4  _ZN8facebook5velox11filesystems12S3FileSystem15openFileForReadESt17basic_string_viewIcSt11char_traitsIcEERKNS1_11FileOptionsE
[...]

PR #7388, opened 2023-11-02
Add support for multiple S3FileSystems

majetideepak:
The endpoint value is used as the S3Identity.
Refactor Minio initialization to the base test class.
We now create a Minio endpoint per test.
If the tests are run by the same process, then multiple filesystems are created.
Extend TableWrite PlanBuilder API to specify the connectorId.

PR #15313, opened 2025-10-29
fix(s3): Make the MinioServer instance only be initialized once

JkSelf:
The free port is not truly available, and a race condition still exists as described in [ISSUE](https://github.com/facebookincubator/velox/issues/15304) and [code](https://github.com/facebookincubator/velox/blob/56a0190ca75ed4a14e7ca18975074b64be681999/velox/exec/tests/utils/PortUtil.h#L25-L27). So we need to ensure that the port usage is shared across multiple S3FileSystemTest instances. This PR moves the `MinioServer` initialization into the `SetUpTestSuite` method, ensuring it is initialized only once, which can help to eliminate the race condition and improves test stability.

### DIFF
--- a/velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemTest.cpp
+++ b/velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemTest.cpp
@@ -29,21 +29,22 @@ namespace {
 
 class S3FileSystemTest : public S3Test {
  protected:
-  static void SetUpTestCase() {
+  static void SetUpTestSuite() {
+    S3Test::SetUpTestSuite();
     memory::MemoryManager::testingSetInstance(memory::MemoryManager::Options{});
-  }
-
-  void SetUp() override {
-    S3Test::SetUp();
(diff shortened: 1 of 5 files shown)


velox/connectors/hive/storage_adapters/s3fs/tests/S3FileSystemTest.cpp
```diff
@@ -57,23 +58,107 @@ class MyCredentialsProvider : public Aws::Auth::AWSCredentialsProvider {
 
 } // namespace
 
-TEST_F(S3FileSystemTest, writeAndRead) {
```
JkSelf, line 60: Duplicated with `TEST_F(S3FileSystemTest, writeFileAndRead)`

majetideepak (changes requested):
@JkSelf We moved away from this singleton approach here https://github.com/facebookincubator/velox/pull/7388 to isolate each test.
Without this, there is a risk that one test could modify the S3 endpoint and not clean up, and another test could change its behavior based on this state. We could end up with flaky behavior here as well.
The freePort utility is used by many other tests. If that has a bug, we should ideally fix that.

PR #15316, opened 2025-10-29
fix: Throw error for unsupported compression kinds in TextReader

### LINKED ISSUE #15314: TextReader lacks support for SNAPPY, LZO, and LZ4 compression, leading to garbled output
### Bug description

CompressionKind lists NONE, ZLIB, SNAPPY, LZO, ZSTD, LZ4, and GZIP. TextReader only supports GZIP, ZLIB, ZSTD, and NONE. SNAPPY/LZO/LZ4 are not supported, and TextReader produces garbled output when reading files compressed with these formats. Are there any existing plans to implement full compression support in TextReader? If not, would a contribution that adds explicit error handling for unsupported types be welcome?

### System information

 N/A

### Relevant logs

```bash

```

xiaodouchen:
This PR fixes an issue where TextReader would produce garbled output when
trying to read files with unsupported compression formats (like SNAPPY,
LZO, LZ4). Instead of failing silently, TextReader will now check the 
compression kind upon initialization and throw an explicit exception if 
it is not supported. This prevents data corruption and provides clear 
feedback to the user.

This PR fixes #15314.


xin-zhang2 (approved):
Thanks for the fix.

xiaodouchen:
@xin-zhang2 Thanks.

@zacw7 @Yuhta @mbasmanova Could you please help take a review and merge it?

PR #15321, opened 2025-10-29
refactor(cudf): Use LOG() instead of std::cout for debug messages

misiugodfrey:
Changed the stdout logging in cudf to use LOG() instead.

First pass just directly changes std::out to LOG(INFO) (I didn't see any cases where I thought LOG(WARNING) or more were justified).

Open questions to resolve:
- Are there sections that we want to group together into one LOG() call (such as where we iterate through a loop)?
- Are there cases we want to use different logging levels than LOG(INFO) (none I could see, but would appreciate a second opinion)?
- Do we want to use the form `if debugEnabled then LOG` or use something like VLOG(), or a debug logging macro instead?

### DIFF
--- a/velox/docs/develop/operators.rst
+++ b/velox/docs/develop/operators.rst
@@ -947,7 +947,7 @@ results available before seeing all input.
 TopNRowNumberNode
 ~~~~~~~~~~~~~~~~~
 
-An optimized version of a WindowNode with a single row_number function and a
+An optimized version of a WindowNode with a single row_number, rank or dense_rank function and a
 limit over sorted partitions.
 
 Partitions the input using specified partitioning keys and maintains up to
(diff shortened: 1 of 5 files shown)


mbasmanova (approved):
Thanks. I would use VLOG(1) for these messages, but I don't have enough context to have an informed opinion.

misiugodfrey:
> Thanks. I would use VLOG(1) for these messages, but I don't have enough context to have an informed opinion.

I just spoke to @karthikeyann and it sounds like it would be preferred to keep this more explicitly under the `cudf.debug_enabled` flag (which is not currently tied to VLOG) and there is currently no motivation to split this into different verbosity levels.  So for now it sounds like the preference is as-is, vs VLOG.

majetideepak (approved):

kevinwilfong:
@misiugodfrey if you could address the formatting issues (see "Run Checks / pre-commit") I'd be happy to help get this merged

misiugodfrey:
@kevinwilfong Apologies, I thought I had fixed the formatting issues, but the check seems to be failing on formatting that was brought in by running clang-format using the formatting file I have in the fork.  I'm looking into what's going wrong on my end, as there seems to be some conflict between what the checks' format expects and what I am seeing.

Update: Looks like this was caused by a clang-format version difference between what I was using and what is used in pre-commit (I wasn't aware of the pre-commit process).  Should be in sync now.

kevinwilfong:
No worries, it looks good now

PR #15322, opened 2025-10-29
fix(build): Add explicit dependency for fastfloat instead of implicit from folly

MBkkt:
Otherwise if folly target for some reason doesn't provide fastfloat we will fail to build

### DIFF
--- a/CMake/resolve_dependency_modules/folly/CMakeLists.txt
+++ b/CMake/resolve_dependency_modules/folly/CMakeLists.txt
@@ -14,8 +14,8 @@
 project(Folly)
 cmake_minimum_required(VERSION 3.28)
 
-velox_set_source(fastfloat)
-velox_resolve_dependency(fastfloat CONFIG REQUIRED)
+velox_set_source(FastFloat)
+velox_resolve_dependency(FastFloat CONFIG REQUIRED)
 
 set(VELOX_FOLLY_BUILD_VERSION v2025.04.28.00)
 set(
--- a/velox/functions/prestosql/CMakeLists.txt
+++ b/velox/functions/prestosql/CMakeLists.txt
@@ -75,6 +75,7 @@ velox_link_libraries(
   velox_presto_types
   velox_functions_util
   Folly::folly
+  FastFloat::fast_float
   stemmer::stemmer
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,6 +618,9 @@ endif()
 velox_set_source(simdjson)
 velox_resolve_dependency(simdjson 3.13.0)
 
+velox_set_source(FastFloat)
+velox_resolve_dependency(FastFloat)
+
 velox_set_source(folly)
 velox_resolve_dependency(folly)
 


MBkkt:
@jinchengchenghh Does it work for you in IBM? I'm not sure will or not it works with IBM vcpkg setup.

I think changes are anyway better to be merged, because be explicit about dependency is better (also FastFloat system and Folly from source didn't work before these changes).

But I want to understand do we need something else to make it works for IBM or not

MBkkt:
@assignUser can you review this please?

CMake/resolve_dependency_modules/folly/CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 project(Folly)
 cmake_minimum_required(VERSION 3.28)
 
-velox_set_source(fastfloat)
-velox_resolve_dependency(fastfloat CONFIG REQUIRED)
+velox_set_source(FastFloat)
+velox_resolve_dependency(FastFloat CONFIG REQUIRED)
```
MBkkt, line 18: without this it's impossible to use system fast float.
Because only FindFastFloat.cmake provided

jinchengchenghh:
It could work for IBM.

CMake/resolve_dependency_modules/folly/CMakeLists.txt
jinchengchenghh: The https://github.com/facebookincubator/velox/blob/main/CMake/resolve_dependency_modules/fastfloat.cmake#L26 name it to FASTFLOAT, which name should we use? I'm not familiar with the Velox build.

jinchengchenghh: The fastfloat.cmake will replace FindFastFloat.cmake

MBkkt: I think there're two cases, both checked in Velox CI:
1. installed system fast-float, in such case FindFastFloat.cmake will be used
2. from source code build fast-float, in such case fastfloat.cmake will be used, because `velox_build_dependency` is case insensitive. 

> https://github.com/facebookincubator/velox/blob/main/CMake/resolve_dependency_modules/fastfloat.cmake#L26 name it to FASTFLOAT, 

This is expected, because for `velox_resolve_dependency_url` argument should be written in UPPER_CASE

velox/functions/prestosql/CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ velox_link_libraries(
   velox_presto_types
   velox_functions_util
   Folly::folly
+  FastFloat::fast_float
```
jinchengchenghh, line 78: fastfloat.cmake does not define the library, could we use it here?
```
add_library(FastFloat::fast_float INTERFACE IMPORTED)
set_target_properties(FastFloat::fast_float PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${FASTFLOAT_SOURCE_DIR}/include"
)
```

MBkkt: > fastfloat.cmake does not define the library

Why do you think so? From local and CI I see the opposite

image

jinchengchenghh: OK, maybe some Velox macros has defined the library.

MBkkt: I don't think it's velox macros, I think FetchContent_MakeAvailable call FastFloat root CMakeLists.txt and it defines the library/etc

jinchengchenghh:
I have verified in Gluten, this PR can fix the build issue https://github.com/apache/incubator-gluten/actions/runs/18937504151/job/54067911462?pr=10978 This blocks Gluten daily rebase, could you help prioritize this PR? Thanks! @majetideepak @kgpai @czentgr

velox/functions/prestosql/CMakeLists.txt
jinchengchenghh, line 78: The dependency is added here. @majetideepak

majetideepak: Thanks!

majetideepak (approved):

CMake/resolve_dependency_modules/folly/CMakeLists.txt
assignUser: ^ Yes exactly, sorry for the late answer. Thanks for the change, I didn't realize we where using fastfloat transitively!

PR #15325, opened 2025-10-30
feat(cudf): Use file buffered input source in `CudfHiveDataSource`

mhaseeb123:
## Description

This PR enables the use of a buffered input source for `CudfHiveDataSource` to benefit from `AsyncDataCache` if available.

## Checklist
- [x] New or existing tests cover the new code. (See https://github.com/facebookincubator/velox/pull/15391)
- [x] I am familiar with the contribution guide

### DIFF
--- a/velox/exec/tests/utils/OperatorTestBase.cpp
+++ b/velox/exec/tests/utils/OperatorTestBase.cpp
@@ -153,6 +153,7 @@ void OperatorTestBase::SetUp() {
 void OperatorTestBase::TearDown() {
   waitForAllTasksToBeDeleted();
   stopPeriodicStatsReporter();
+  executor_.reset();
   // There might be lingering exchange source on executor even after all tasks
   // are deleted. This can cause memory leak because exchange source holds
   // reference to memory pool. We need to make sure they are properly cleaned.
--- a/velox/tool/trace/AggregationReplayer.h
+++ b/velox/tool/trace/AggregationReplayer.h
@@ -29,6 +29,7 @@ class AggregationReplayer : public OperatorReplayerBase {
       const std::string& taskId,
       const std::string& nodeId,
       const std::string& operatorType,
+      const std::string& spillBaseDir,
       const std::string& driverIds,
       uint64_t queryCapacity,
       folly::Executor* executor)
@@ -38,6 +39,7 @@ class AggregationReplayer : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            spillBaseDir,
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/FilterProjectReplayer.h
+++ b/velox/tool/trace/FilterProjectReplayer.h
@@ -43,6 +43,7 @@ class FilterProjectReplayer : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            "",
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/HashJoinReplayer.h
+++ b/velox/tool/trace/HashJoinReplayer.h
@@ -29,6 +29,7 @@ class HashJoinReplayer final : public OperatorReplayerBase {
       const std::string& taskId,
       const std::string& nodeId,
       const std::string& operatorType,
+      const std::string& spillBaseDir,
       const std::string& driverIds,
       uint64_t queryCapacity,
       folly::Executor* executor)
@@ -38,6 +39,7 @@ class HashJoinReplayer final : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            spillBaseDir,
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/IndexLookupJoinReplayer.h
+++ b/velox/tool/trace/IndexLookupJoinReplayer.h
@@ -38,6 +38,7 @@ class IndexLookupJoinReplayer : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            "",
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/OperatorReplayerBase.cpp
+++ b/velox/tool/trace/OperatorReplayerBase.cpp
@@ -33,11 +33,12 @@ using namespace facebook::velox;
 
 namespace facebook::velox::tool::trace {
 OperatorReplayerBase::OperatorReplayerBase(
-    std::string traceDir,
-    std::string queryId,
-    std::string taskId,
-    std::string nodeId,
-    std::string nodeName,
+    const std::string& traceDir,
+    const std::string& queryId,
+    const std::string& taskId,
+    const std::string& nodeId,
+    const std::string& nodeName,
+    const std::string& spillBaseDir,
     const std::string& driverIds,
     uint64_t queryCapacity,
     folly::Executor* executor)
@@ -48,6 +49,7 @@ OperatorReplayerBase::OperatorReplayerBase(
       taskTraceDir_(
           exec::trace::getTaskTraceDirectory(traceDir, queryId_, taskId_)),
       nodeTraceDir_(exec::trace::getNodeTraceDirectory(taskTraceDir_, nodeId_)),
+      spillBaseDir_(spillBaseDir),
       fs_(filesystems::getFileSystem(taskTraceDir_, nullptr)),
       pipelineIds_(exec::trace::listPipelineIds(nodeTraceDir_, fs_)),
       driverIds_(
@@ -83,15 +85,17 @@ OperatorReplayerBase::OperatorReplayerBase(
 
 RowVectorPtr OperatorReplayerBase::run(bool copyResults) {
   auto queryCtx = createQueryCtx();
-  std::shared_ptr<exec::test::TempDirectoryPath> spillDirectory;
-  if (queryCtx->queryConfig().spillEnabled()) {
-    spillDirectory = exec::test::TempDirectoryPath::create();
+  std::shared_ptr<exec::test::TempDirectoryPath> localSpillDirectory;
+  if (queryCtx->queryConfig().spillEnabled() && spillBaseDir_.empty()) {
+    localSpillDirectory = exec::test::TempDirectoryPath::create();
   }
 
   TraceReplayTaskRunner traceTaskRunner(createPlan(), std::move(queryCtx));
   auto [task, result] =
       traceTaskRunner.maxDrivers(driverIds_.size())
-          .spillDirectory(spillDirectory ? spillDirectory->getPath() : "")
+          .spillDirectory(
+              localSpillDirectory != nullptr ? localSpillDirectory->getPath()
+                                             : spillBaseDir_)
           .run(copyResults);
   printStats(task);
   return result;
--- a/velox/tool/trace/OperatorReplayerBase.h
+++ b/velox/tool/trace/OperatorReplayerBase.h
@@ -29,11 +29,12 @@ namespace facebook::velox::tool::trace {
 class OperatorReplayerBase {
  public:
   OperatorReplayerBase(
-      std::string traceDir,
-      std::string queryId,
-      std::string taskId,
-      std::string nodeId,
-      std::string nodeName,
+      const std::string& traceDir,
+      const std::string& queryId,
+      const std::string& taskId,
+      const std::string& nodeId,
+      const std::string& nodeName,
+      const std::string& spillBaseDir,
       const std::string& driverIds,
       uint64_t queryCapacity,
       folly::Executor* executor);
@@ -63,6 +64,7 @@ class OperatorReplayerBase {
   const std::string nodeName_;
   const std::string taskTraceDir_;
   const std::string nodeTraceDir_;
+  const std::string spillBaseDir_;
   const std::shared_ptr<filesystems::FileSystem> fs_;
   const std::vector<uint32_t> pipelineIds_;
   const std::vector<uint32_t> driverIds_;
--- a/velox/tool/trace/OrderByReplayer.h
+++ b/velox/tool/trace/OrderByReplayer.h
@@ -29,6 +29,7 @@ class OrderByReplayer : public OperatorReplayerBase {
       const std::string& taskId,
       const std::string& nodeId,
       const std::string& operatorType,
+      const std::string& spillBaseDir,
       const std::string& driverIds,
       uint64_t queryCapacity,
       folly::Executor* executor)
@@ -38,6 +39,7 @@ class OrderByReplayer : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            spillBaseDir,
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/PartitionedOutputReplayer.cpp
+++ b/velox/tool/trace/PartitionedOutputReplayer.cpp
@@ -121,6 +121,7 @@ PartitionedOutputReplayer::PartitionedOutputReplayer(
           taskId,
           nodeId,
           operatorType,
+          "",
           driverIds,
           queryCapacity,
           executor),
--- a/velox/tool/trace/TableScanReplayer.h
+++ b/velox/tool/trace/TableScanReplayer.h
@@ -42,6 +42,7 @@ class TableScanReplayer final : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            "",
             driverIds,
             queryCapacity,
             executor) {}
--- a/velox/tool/trace/TableWriterReplayer.h
+++ b/velox/tool/trace/TableWriterReplayer.h
@@ -42,6 +42,7 @@ class TableWriterReplayer final : public OperatorReplayerBase {
             taskId,
             nodeId,
             operatorType,
+            "",
             driverIds,
             queryCapacity,
             executor),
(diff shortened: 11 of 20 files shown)


devavret:
Let's also add a cuDF config that disables this path in case we're working with local storage

mhaseeb123:
> Let's also add a cuDF config that disables this path in case we're working with local storage

Done!

~Update: There's already a `HiveConfig::kEnableFileHandleCache` available based on which the `fileHandleFactory_`'s `maxSize()` is returned. Using that check instead~

Update: Added an option `CudfHiveConfig::kUseBufferedInput` to enable/disable (default:enable) the use of buffered input data source

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
```diff
@@ -311,20 +390,63 @@ void CudfHiveDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
...
+      fileHandleCachePtr = fileHandleFactory_->generate(
+          fileHandleKey, &fileProperties, fsStats_ ? fsStats_.get() : nullptr);
+      VELOX_CHECK_NOT_NULL(fileHandleCachePtr.get());
+    } catch (const VeloxRuntimeError& e) {
+      LOG(WARNING) << fmt::format(
+          "Failed to generate file handle cache for file {}, using fallback instead",
+          split_->filePath);
+      return cudf::io::source_info{split_->filePath};
+    }
+
```
majetideepak, line 430: I also see a check for `cacheTTLController` in the Hive path. Don't we need that here?

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
```diff
@@ -311,20 +390,63 @@ void CudfHiveDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
 
 std::unique_ptr<cudf::io::chunked_parquet_reader>
 CudfHiveDataSource::createSplitReader() {
+  // Build source info for the chunked parquet reader
+  auto sourceInfo = [&]() {
+    // fileHandleFactory_ has a max size of 0 if the file handle cache is
+    // disabled
+    if (fileHandleFactory_->maxSize() == 0) {
```
majetideepak: I don't see this case in the HiveConnector path.

majetideepak:
> Update: There's already a HiveConfig::kEnableFileHandleCache available based on which the fileHandleFactory_'s maxSize() is returned. Using that check instead

Disabling filehandle cache is different from disabling the use of existing Velox filehandle factories. Let's not overload.

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
mhaseeb123: Right but if the fileHandleFactory_->maxSize() == 0 then it's a good early check to just use the file data source path

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
mhaseeb123: Thanks for pointing out. Added

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
mhaseeb123: Looked at the impl of `generate` in CachedFactory.h, it simply just bypasses the cache if not available so this check of `maxSize()` seems wrong as we would be forcing a file data source when we might not intend it. Removed in aee28b0

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
devavret: @lga-zurich reported once that this part breaks for s3 links.

velox/experimental/cudf/connectors/hive/CudfHiveConfig.cpp
```diff
@@ -133,6 +133,16 @@ cudf::data_type CudfHiveConfig::timestampTypeSession(
   return cudf::data_type(cudf::type_id{unit});
 }
 
+bool CudfHiveConfig::isUseBufferedInput() const {
```
jinchengchenghh: All the HiveConfig also needs a document.

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
```diff
@@ -81,6 +81,12 @@ class CudfHiveConfig {
   static constexpr const char* kTimestampTypeSession =
       "parquet.reader.timestamp_type";
 
+  // Whether to use async data cache aided buffered input source
+  static constexpr const char* kUseBufferedInput =
```
jinchengchenghh, line 88: Cudf supports many file formats, may need to separate the config for common CudfConfig and file format specified config in the future.

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
```diff
@@ -42,6 +46,76 @@
...
+ public:
+  explicit BufferedInputDataSource(
+      std::shared_ptr<facebook::velox::dwio::common::BufferedInput> input)
+      : input_(std::move(input)), fileSize_(input_->getReadFile()->size()) {}
+
+  [[nodiscard]] size_t size() const override {
+    return fileSize_;
+  }
+
+  std::unique_ptr<datasource::buffer> host_read(size_t offset, size_t size)
```
jinchengchenghh, line 66: host_read -> hostRead, and others

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
majetideepak: Yes. Let's not add Parquet to this config variable. Let's use `cudf.hive.async-data-cache-enabled`.
This is only a temporary config until we fix the extra memcpy and add the KvikioFS. After these, we can remove this.
Please add a comment for the same.
The AsyncDataCache currently has to be initialized via `setInstance()`. That will control enabling/disabling the cache.
http://github.com/facebookincubator/velox/blob/main/velox/common/caching/AsyncDataCache.cpp#L700

velox/experimental/cudf/connectors/hive/CudfHiveConfig.cpp
majetideepak: rename to `useAsyncDataCache`

majetideepak: I agree. We should add `cudf.hive.` prefix to these options, I feel.

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
mhaseeb123: I think this is an override of the function with the same name in `cudf::io::datasource`

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
mhaseeb123: >Cudf supports many file formats, may need to separate the config for common CudfConfig and file format specified config in the future.

Updated the config option to `cudf.hive.use-buffered-input` and updated comments

devavret: > `cudf.hive.async-data-cache-enabled`

This is basically a `use_velox_io`. it can be used with or without async data cache. although you need it turned on for async data cache to take effect in cudf's readers. you still also need to enable async data cache.

mhaseeb123: > Let's use cudf.hive.async-data-cache-enabled.

and

> The AsyncDataCache currently has to be initialized via setInstance(). That will control enabling/disabling the cache.
http://github.com/facebookincubator/velox/blob/main/velox/common/caching/AsyncDataCache.cpp#L700

I think `cudf.hive.use-buffered-input` is the correct option here. The file handle AsyncDataCache is initialized in `fileHandleFactory` of the HiveConnector at construction based on `HiveConfig:: kEnableFileHandleCache` config (default: on). Whereas in `CudfHiveDataSourece::createSplitReader()` we are only deciding if we want to use the buffered input data source (which may benefit from the file handle cache if available and enabled, and would still work without the cache).

mhaseeb123: >This is basically a use_velox_io. it can be used with or without async data cache. although you need it turned on for async data cache to take effect in cudf's readers. you still also need to enable async data cache.

This is okay too. Pretty flexible on the names here as long we are on the same page 😄

velox/experimental/cudf/connectors/hive/CudfHiveConfig.cpp
mhaseeb123: Agreed. I can do that. Just to note most of the `CudfHiveConfig` options are specific to the cudf's parquet reader so keeping `parquet` in their name makes sense. We need a proper nomenclature (and perhaps a proper class hierarchy) to support other cudf readers in the future.

mhaseeb123: I updated the cudf hive options using the following schema for now:

Prefix: `cudf.hive`: Applies to both cudf hive readers and writers
Prefix: `cudf.hive.reader`: Applies to multiple cudf hive readers
Prefix: `cudf.hive.writer`: Applies to multiple cudf hive writers
Prefix: `cudf.hive.reader/writer.<name>`: Applies to the specified `<name>` reader/writer

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
mhaseeb123: > Cudf supports many file formats, may need to separate the config for common CudfConfig and file format specified config in the future.

xref: https://github.com/facebookincubator/velox/pull/15325/files#r2480784073

devavret:
There's another small UX issue here. cudf hive configs need to go in hive.properties in presto. When these are shared between coordinator and worker the coordinator refuses to start citing unknown config. @majetideepak

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
```diff
@@ -72,20 +72,31 @@ class CudfHiveConfig {
   // Whether to allow reading matching select columns from mismatched CudfHive
   // files.
   static constexpr const char* kAllowMismatchedCudfHiveSchemas =
```
jinchengchenghh, line 74: Please use CudfParquetConfig : public CudfHiveConfig to separate them.

majetideepak:
>cudf hive configs need to go in hive.properties in presto. When these are shared between coordinator and worker the coordinator refuses to start citing unknown config

@devavret The HiveConfigs for the Coordinator and C++ Worker are different. So we should not be adding these to the coordinator.

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
majetideepak: `use_velox_io` will probably not be applicable when we add KvikioFS to Velox :).

majetideepak: Looking at the implementation, I think `cudf.hive.use-buffered-input` is accurate. 
```
explicit BufferedInputDataSource(
      std::shared_ptr<facebook::velox::dwio::common::BufferedInput> input)
```
There is also a DirectBufferedInput that can be used and is not related to Async Data Cache.
https://github.com/facebookincubator/velox/blob/ec12473162b9c18df3fd7355d0490bf799a60b48/velox/dwio/common/DirectBufferedInput.h#L113

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
mhaseeb123: Going to revert changes in this file in this PR and will open a follow up PR

velox/experimental/cudf/connectors/hive/CudfHiveConfig.cpp
mhaseeb123: > All the HiveConfig also needs a document.

In a follow up PR along with other suggested changes

velox/experimental/cudf/tests/TableScanTest.cpp
```diff
@@ -439,3 +481,54 @@ TEST_F(TableScanTest, splitOffset) {
           .assertEmptyResults(),
       "CudfHiveDataSource cannot process splits with non-zero offset");
 }
+
+TEST_F(TableScanTest, s3DataSource) {
```
mhaseeb123: New test with S3 data source

velox/experimental/cudf/tests/TableScanTest.cpp
```diff
@@ -439,3 +481,54 @@ TEST_F(TableScanTest, splitOffset) {
...
+  // Copy source file to destination bucket.
+  dest << src.rdbuf();
+  ASSERT_GT(dest.tellp(), 0) << "Unable to copy from source " << sourceFile;
+  dest.close();
+
+  // Reset the CudfHiveConnector config to use buffered data source
+  auto configOverride = std::unordered_map<std::string, std::string>{
+      {facebook::velox::cudf_velox::connector::hive::CudfHiveConfig::
+           kUseBufferedInput,
+       "true"}};
```
mhaseeb123: Just to answer your offline query @karthikeyann, If we set this to `false` instead, this test will try to use kvikio to do I/O and will fail as it doesn't have all the S3 config from `minioserver->hiveConfig()`

velox/experimental/cudf/tests/TableScanTest.cpp
devavret: perfect

devavret (approved):
This looks like what we need. I'll still wait for reviews from @lga-zurich @majetideepak and @zoltan.
Approving to unblock

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
```diff
@@ -81,6 +81,15 @@ class CudfHiveConfig {
   static constexpr const char* kTimestampTypeSession =
       "parquet.reader.timestamp_type";
 
+  // Whether to use the buffered input source for CudfHiveDataSource. The
+  // buffered source may benefit from the file handle `AsyncDataCache` in the
+  // connector's fileHandleFactory. This file handle cache is controlled
```
majetideepak: `The buffered source may benefit from the file handle AsyncDataCache in the connector's fileHandleFactory.`
The above comment is not accurate. AsyncDataCache is an in-memory cache and uses the FileHandle to read data from storage when required. We can say.
`Takes advantage of AsyncDataCache if enabled.`

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
```diff
@@ -81,6 +81,15 @@ class CudfHiveConfig {
   static constexpr const char* kTimestampTypeSession =
       "parquet.reader.timestamp_type";
 
+  // Whether to use the buffered input source for CudfHiveDataSource. The
```
majetideepak: `BufferedInput` source

velox/experimental/cudf/connectors/hive/CudfHiveConfig.h
```diff
@@ -148,6 +157,9 @@ class CudfHiveConfig {
   cudf::data_type timestampType() const;
   cudf::data_type timestampTypeSession(const config::ConfigBase* session) const;
 
+  bool isUseBufferedInput() const;
```
majetideepak: nit: `useBufferedInput` is sufficient.

velox/experimental/cudf/tests/TableScanTest.cpp
```diff
@@ -439,3 +481,54 @@ TEST_F(TableScanTest, splitOffset) {
           .assertEmptyResults(),
       "CudfHiveDataSource cannot process splits with non-zero offset");
 }
+
+TEST_F(TableScanTest, s3DataSource) {
+  auto minioServer = std::make_unique<MinioServer>();
+  minioServer->start();
+
+  const auto sourceFile = facebook::velox::test::getDataFilePath(
+      "velox/connectors/hive/storage_adapters/s3fs/tests",
```
majetideepak: nit: why is this the baseDir?
Can you point to examples directly?

velox/experimental/cudf/tests/TableScanTest.cpp
majetideepak: We should add a new file `S3TableScanTest.cpp` that uses the S3Test as the base class.
Since this test depends on s3fs, we need to set the correct dependencies.
I think we will be adding a lot more tests for S3 in the future.

velox/experimental/cudf/tests/TableScanTest.cpp
mhaseeb123: Copied it from `S3ReadTest.cpp` directly.

velox/experimental/cudf/tests/TableScanTest.cpp
mhaseeb123: Sounds good. Let me move this to that

mhaseeb123:
@devavret @majetideepak @zoltan @karthikeyann @lga-zurich  I have moved the S3 read test to this PR https://github.com/facebookincubator/velox/pull/15391 and addressed the remaining comments so far. Please let me know if there's anything else

velox/experimental/cudf/tests/TableScanTest.cpp
mhaseeb123: Moved to https://github.com/facebookincubator/velox/pull/15391

velox/experimental/cudf/tests/TableScanTest.cpp
mhaseeb123: https://github.com/facebookincubator/velox/pull/15391

karthikeyann (approved):

majetideepak (approved):

PR #15328, opened 2025-10-30
perf: Do not request data size if there is only  a single source as in persistent shuffle

xiaoxmeng:
Differential Revision: D85647156

### DIFF
--- a/velox/exec/tests/ExchangeClientTest.cpp
+++ b/velox/exec/tests/ExchangeClientTest.cpp
@@ -589,7 +589,7 @@ TEST_P(ExchangeClientTest, acknowledge) {
   SCOPED_TESTVALUE_SET(
       "facebook::velox::exec::test::LocalExchangeSource::pause",
       std::function<void(void*)>(([&numberOfAcknowledgeRequests](void*) {
-        numberOfAcknowledgeRequests++;
+        ++numberOfAcknowledgeRequests;
       })));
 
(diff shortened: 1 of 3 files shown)


velox/exec/tests/ExchangeClientTest.cpp
```diff
@@ -589,7 +589,7 @@ TEST_P(ExchangeClientTest, acknowledge) {
   SCOPED_TESTVALUE_SET(
       "facebook::velox::exec::test::LocalExchangeSource::pause",
       std::function<void(void*)>(([&numberOfAcknowledgeRequests](void*) {
-        numberOfAcknowledgeRequests++;
+        ++numberOfAcknowledgeRequests;
```
duanmeng, line 592: Curiously, is there any performance difference here?

tanjialiang (approved):

velox/exec/tests/ExchangeClientTest.cpp
tanjialiang: Post-increment (i++) may be less efficient because it often requires making a copy of the object to return the original value before incrementing.
Pre-increment (++i) usually just increments and returns a reference, avoiding the copy.

tanjialiang: But compiler should optimize both well if i is primitive type.

PR #15331, opened 2025-10-30
feat: Check join filter in PlanConsistencyChecker

mbasmanova:
Differential Revision: D85850879


xiaoxmeng (approved):
@mbasmanova thanks for the fix!

PR #15333, opened 2025-10-30
refactor: HivePartitionUtil for better extensibility

PingLiuPing:
Refactor HivePartitionUtil to use class-based design for extensibility

This PR refactors the partition value formatting logic from standalone
template functions to a class-based design, enabling better code reuse and
extensibility for different table formats (e.g., Iceberg).

### DIFF
--- a/velox/connectors/hive/HivePartitionUtil.h
+++ b/velox/connectors/hive/HivePartitionUtil.h
@@ -15,12 +15,63 @@
  */
 #pragma once
 
+#include <memory>
 #include "velox/vector/ComplexVector.h"
 
 namespace facebook::velox::connector::hive {
 
-std::vector<std::pair<std::string, std::string>> extractPartitionKeyValues(
-    const RowVectorPtr& partitionsVector,
-    vector_size_t row);
+FOLLY_ALWAYS_INLINE std::string formatDate(int32_t value) {
+  return DATE()->toString(value);
+}
+
+/// Converting partition values to their string representations.
+/// Provides virtual methods for formatting different data types
+/// according to Hive partitioning conventions.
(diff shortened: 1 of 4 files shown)


PingLiuPing:
With this PR, when formatting iceberg partition values, I can use following class and just overwrite int32, Timestamp and StringView types. Other types can share same formatting with HivePartitionUtil.

```
class IcebergPartitionUtil : public HivePartitionUtil {
 public:
  explicit IcebergPartitionUtil(TransformPtr transform)
      : transform_(transform) {}

  ~IcebergPartitionUtil() override = default;

  using HivePartitionUtil::toPartitionString;

  std::string toPartitionString(int32_t value, const TypePtr& type)
      const override;

  std::string toPartitionString(Timestamp value, const TypePtr& type)
      const override;

  std::string toPartitionString(StringView value, const TypePtr& type)
      const override;

 private:
  // The partition transform to apply when formatting values.
  TransformPtr transform_;
};
```

velox/connectors/hive/HivePartitionUtil.h
```diff
@@ -15,12 +15,78 @@
...
+ public:
+  virtual ~HivePartitionUtil() = default;
+
+  FOLLY_ALWAYS_INLINE std::string toPartitionString(
+      bool value,
+      const TypePtr& type) const {
+    return value ? "true" : "false";
+  }
+
+  FOLLY_ALWAYS_INLINE std::string toPartitionString(
```
jinchengchenghh: Here should be a template function

PingLiuPing: Yes, thanks.

PR #15340, opened 2025-10-31
fix(docs): Correct metric name for spill_writes_count

aditi-pandit:


velox/docs/monitoring/metrics.rst
```diff
@@ -520,7 +520,7 @@ Spilling
      - The distribution of the amount of time spent on serializing rows for
        spilling in range of [0, 600s] with 20 buckets. It is configured to report
        the latency at P50, P90, P99, and P100 percentiles.
-   * - spill_disk_writes_count
+   * - spill_writes_count
      - Count
      - The number of disk writes to spill rows.
```
majetideepak: Let's fix the comment  as well.
`The number of spill writes to storage, which is the number of write calls to the Velox filesystem.`

aditi-pandit: Modified that to "The number of Velox filesystem write calls to spill rows." to keep it concise.

aditi-pandit:
Thanks @majetideepak. Have updated the code.

PR #15341, opened 2025-10-31
fix: Install utf8proc headers

kou:
`velox/functions/lib/string/StringCore.h` includes `utf8procImpl.h`. So we need to install them.

This also updates `velox/.gitattributes`. Don't ignore `CMakeLists.txt` from diff because `external/**/CMakeLists.txt` are written by us.


kou:
@majetideepak Could you take a look at this?

kou:
@assignUser Could you review this?

kou:
Could someone reopen this? This is needed for Nimble.

Issue #15348, opened 2025-10-31
HiveTableHandle is missing handles for columns used in pushed down filters


mbasmanova:
### Bug description

TableScanNode includes TableHandle and a map of assignments. The assignments contain columns that are produced by the scan node. Pushed down filters may use columns that are not produces and therefore are not present in 'assignments'. Hence, TableHandle needs to include the necessary information about these columns. HiveTableHandle is missing that. In Presto Java, these properties are stored in HiveTableLayoutHandle.predicateColumns property. See more in https://github.com/facebookincubator/axiom/issues/243#issuecomment-3256078561

```
public class HiveTableLayoutHandle
        extends BaseHiveTableLayoutHandle
{
...
    private final Map<String, HiveColumnHandle> predicateColumns;

```

Repro:

```
TEST_F(TableScanTest, xxx) {
  auto data = makeVectors(1, 100, ROW({"a", "b", "c"}, BIGINT()));

  auto filePath = TempFilePath::create();
  writeToFile(filePath->getPath(), data);

  connector::ColumnHandleMap assignments = {
      {"x", regularColumn("a", BIGINT())},
  };

  auto split = exec::test::HiveConnectorSplitBuilder(filePath->getPath())
                   .partitionKey("ds", "2025-10-23")
                   .build();

  auto plan =
      PlanBuilder()
          .startTableScan()
          .outputType(ROW({"x"}, {BIGINT()}))
          .assignments(assignments)
          .dataColumns(ROW(
              {"a", "b", "c", "ds"}, {BIGINT(), BIGINT(), BIGINT(), VARCHAR()}))
          .subfieldFilter("ds = '2025-10-23'")
          .endTableScan()
          .planNode();

  auto results = AssertQueryBuilder(plan).split(split).copyResults(pool());

  LOG(ERROR) << results->toString();
}

Reason: ColumnHandle is missing for partition key ds
Retriable: False
Expression: it != partitionKeys_->end()
Context: Split [Hive: /tmp/velox_test_xNmLaz 0 - 18446744073709551615] Task test_cursor_1
Additional Context: Operator: TableScan[0] 0
Function: setPartitionValue
fbcode/velox/connectors/hive/SplitReader.cpp
Line: 481

# 4  facebook::velox::connector::hive::SplitReader::setPartitionValue(...
# 5  facebook::velox::connector::hive::SplitReader::adaptColumns(...
# 6  facebook::velox::connector::hive::SplitReader::getAdaptedRowType() const
[...]

PR #15351, opened 2025-10-31
feat: Add HiveTableHandle::columnHandles

Yuhta:
Summary:
Currently if a partitioning column is not projected out, but used in remaining filter, the engine still adds it to assignments in order for the connector to recognize it.  This is a hack and we should pass the full column handles inside the table handle.

Fix https://github.com/facebookincubator/velox/issues/15348

Differential Revision: D85979627

### DIFF
--- a/velox/connectors/hive/HiveDataSource.h
+++ b/velox/connectors/hive/HiveDataSource.h
@@ -37,7 +37,7 @@ class HiveDataSource : public DataSource {
   HiveDataSource(
       const RowTypePtr& outputType,
       const connector::ConnectorTableHandlePtr& tableHandle,
-      const connector::ColumnHandleMap& columnHandles,
+      const connector::ColumnHandleMap& assignments,
       FileHandleFactory* fileHandleFactory,
       folly::Executor* executor,
       const ConnectorQueryCtx* connectorQueryCtx,
@@ -146,6 +146,10 @@ class HiveDataSource : public DataSource {
     }
     return emptyOutput_;
   }
+
+  // Add the information from column handle to the corresponding fields in this
+  // object.
+  void processColumnHandle(const HiveColumnHandlePtr& handle);
 
   // The row type for the data source output, not including filter-only columns
   const RowTypePtr outputType_;
--- a/velox/connectors/hive/TableHandle.h
+++ b/velox/connectors/hive/TableHandle.h
@@ -147,7 +147,8 @@ class HiveTableHandle : public ConnectorTableHandle {
       common::SubfieldFilters subfieldFilters,
       const core::TypedExprPtr& remainingFilter,
       const RowTypePtr& dataColumns = nullptr,
-      const std::unordered_map<std::string, std::string>& tableParameters = {});
+      const std::unordered_map<std::string, std::string>& tableParameters = {},
+      std::vector<HiveColumnHandlePtr> columnHandles = {});
 
   const std::string& tableName() const {
     return tableName_;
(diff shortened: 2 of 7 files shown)


velox/connectors/hive/TableHandle.h
```diff
@@ -178,6 +179,11 @@ class HiveTableHandle : public ConnectorTableHandle {
     return tableParameters_;
   }
 
+  // Full schema including partitioning columns and data columns.
```
mbasmanova: Does this need to be full schema or only columns that are used in subfieldFilters and remainingFilter?

velox/connectors/hive/TableHandle.h
```diff
@@ -147,7 +147,8 @@ class HiveTableHandle : public ConnectorTableHandle {
       common::SubfieldFilters subfieldFilters,
       const core::TypedExprPtr& remainingFilter,
       const RowTypePtr& dataColumns = nullptr,
-      const std::unordered_map<std::string, std::string>& tableParameters = {});
+      const std::unordered_map<std::string, std::string>& tableParameters = {},
+      std::vector<HiveColumnHandlePtr> columnHandles = {});
```
mbasmanova, line 151: Would you document this new argument?

velox/connectors/hive/HiveDataSource.h
```diff
@@ -147,6 +147,8 @@ class HiveDataSource : public DataSource {
     return emptyOutput_;
   }
 
+  void processColumnHandle(const HiveColumnHandlePtr& handle);
```
mbasmanova, line 152: Would you add a comment to explain what this method is doing? Should we verify consistency between assignments and columnHandles? E.g. what if column x is reported as partition key in assignment and regular column in columnHandles?

velox/exec/tests/TableScanTest.cpp
```diff
@@ -6143,5 +6143,43 @@ TEST_F(TableScanTest, parallelUnitLoader) {
...
+  connector::ColumnHandleMap assignments = {
+      {"x", regularColumn("a", BIGINT())}};
+  auto split = exec::test::HiveConnectorSplitBuilder(filePath->getPath())
+                   .partitionKey("ds", "2025-10-23")
+                   .build();
+  auto plan = PlanBuilder()
+                  .startTableScan()
+                  .outputType(ROW({"x"}, {BIGINT()}))
+                  .assignments(assignments)
+                  .dataColumns(asRowType(data[0]->type()))
```
mbasmanova, line 6163: Why do we need both dataColumns and columnHandles?

velox/exec/tests/TableScanTest.cpp
```diff
@@ -6143,5 +6143,43 @@ TEST_F(TableScanTest, parallelUnitLoader) {
...
+                  .columnHandles({
+                      std::make_shared<HiveColumnHandle>(
```
mbasmanova: Can we use helper methods regularColumn, synthesizedColumn, partitionKey from velox/exec/tests/utils/HiveConnectorTestBase.h ?

velox/connectors/hive/TableHandle.h
Yuhta: Ideally full schema, but in practice we only look it up for the columns appearing in filters and remaining filter expression

velox/connectors/hive/HiveDataSource.h
Yuhta: Yeah that's something I have in mind but did not add it, let me add some very strict check for now since it's new

velox/exec/tests/TableScanTest.cpp
Yuhta: dataColumns are needed for schema evolution purpose as well, because column handle does not tell us the position of the column in file.  In this test case though we can probably skip it, but in prod usage we always supply it.

mbasmanova: Would be nice to document all pieces of information that go into HiveTableHandle.

velox/connectors/hive/TableHandle.h
Yuhta: I added the document on the accessor method below

velox/exec/tests/TableScanTest.cpp
Yuhta: I will beef up the comments on the `HiveTableHandle` accessors

PR #15352, opened 2025-10-31
build(ci): Trigger java image upload on push to main

assignUser:
The DOCKER_UPLOAD_CACHE envvar is used in the bake file to determine the output type (i.e. if to push or not).

It was missing from the java dependent image step and so they were not uploaded.


czentgr (approved):
Thanks.

PR #15354, opened 2025-11-01
misc(runtimestats): Add runtime metric to find the max cpu time of running drivers per operator and expose existing ioWaitWallNanos

shangm2:
Summary: as titled

Differential Revision: D85927678


velox/exec/OperatorUtils.cpp
```diff
@@ -130,7 +130,9 @@ bool shouldAggregateRuntimeMetric(const std::string& name) {
       "dataSourceAddSplitWallNanos",
       "dataSourceLazyWallNanos",
       "queuedWallNanos",
-      "flushTimes"};
+      "flushTimes",
+      "driverCpuTimeNanos",
```
tanjialiang, line 134: shall we use the defined constant? If possible we can make other entries to reuse their corresponding defined constants as well.

PR #15357, opened 2025-11-01
fix: Upgrade java version for docker images

kgpai:
Upgrading Presto java version also requires us to upgrade the  java version used in the images.


kgpai:
@czentgr I believe this is the only change that is required, the problem is not with the url for the cli.

amitkdutta (approved):
Thanks @kgpai

kagamiori (approved):

PR #15359, opened 2025-11-01
fix: Avoid throw exception in toSubfieldFilter

MBkkt:
Why?

Because in Axiom now code written like this
```cpp
    try {
      auto pair = velox::exec::toSubfieldFilter(typedExpr, evaluator);
      if (!pair.second) {
        remainingConjuncts.push_back(std::move(typedExpr));
        continue;
      }
      pushdownConjuncts.push_back(typedExpr);
    } catch (const std::exception&) {
      remainingConjuncts.push_back(std::move(typedExpr));
    }
  }
```

it's not only ineffective, it's inconvenient to debug (you cannot use break on throw, because this function will throw even when everything fine)

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -543,7 +543,9 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
       if (!right.second) {
         return {};
       }
-      VELOX_CHECK(left.first == right.first);
+      if (left.first != right.first) {
+        return {};
+      }
       return {
           std::move(left.first),
           makeOrFilter(std::move(left.second), std::move(right.second))};


velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -536,7 +536,13 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
...
     if (call->name() == "or") {
       auto left = toSubfieldFilter(call->inputs()[0], evaluator);
+      if (!left.second) {
+        return {};
+      }
       auto right = toSubfieldFilter(call->inputs()[1], evaluator);
+      if (!right.second) {
+        return {};
+      }
       VELOX_CHECK(left.first == right.first);
```
MBkkt:
```suggestion
      if (left.first != right.first) {
        return {};
      }
```

Yuhta:
Can we add a bool flag or another function (e.g. `toSubfieldFilterNoThrow`) to introduce this behavior?  Silently suppressing errors should not be the default behavior.

MBkkt:
@Yuhta This function already returns null in some cases (with implicit error) and in Axiom this error anyway ignored.

Are any other usages where this information is used?

Yuhta:
@MBkkt Ideally null should be returned only for unsupported expression (valid expression that cannot be converted to single field filters).  For invalid expression, we should fail early as soon as we can (unless there is some reason preventing us from doing so).

MBkkt:
@Yuhta ok, sounds reasonable to me, I will change this

MBkkt:
@Yuhta I think for this PR only for unsupported expressions null returned

Yuhta (approved):
I see, this method is a little weird and very limited.  Ideally people usually do is to extract a list of filters from a conjunctive expression.  It seems this method was created only for testing purpose so the behavior is a little weird.  We may need to revisit this method and exam where it is used later.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -536,8 +536,16 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
   if (auto call = asCall(expr.get())) {
     if (call->name() == "or") {
```
mbasmanova, line 537: I wonder if we should remove the logic for handling "or"... I don't think Axiom expects that logic.

Yuhta (changes requested):
Discussed with @mbasmanova and we probably should document this method first to make the behavior and intended usage clear before changing it.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -536,8 +536,16 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
...
       auto left = toSubfieldFilter(call->inputs()[0], evaluator);
+      if (!left.second) {
+        return {};
+      }
       auto right = toSubfieldFilter(call->inputs()[1], evaluator);
-      VELOX_CHECK(left.first == right.first);
+      if (!right.second) {
+        return {};
+      }
+      if (left.first != right.first) {
```
mbasmanova, line 546: Let's make sure to add tests for this logic.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -559,8 +567,7 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
       return std::make_pair(std::move(subfield), std::move(filter));
     }
   }
-  VELOX_UNSUPPORTED(
-      "Unsupported expression for range filter: {}", expr->toString());
+  return {};
```
mbasmanova, line 570: It might be clearer to change return type to std::optional and return std::nullptr.

MBkkt: I think it's worse.
What is semantics of nullptr filter in such case?

PR #15363, opened 2025-11-01
fix: Add missed handle backslash in unescapeJson

MBkkt:
Also return count of written bytes to avoid neediness of read json string twice (now user of this function can over-allocate memory).


MBkkt:
@kgpai Could you review please?

MBkkt:
@kgpai Could you review please?

MBkkt:
@kgpai Could you review please?

velox/functions/prestosql/json/JsonStringUtil.cpp
```diff
@@ -552,6 +552,7 @@ unescapeSizeForJsonFunctions(const char* input, size_t length, bool fully) {
         case 'f':
         case 'r':
         case 't':
+        case '\\':
```
kgpai, line 555: Can you add a test case that exercises this ?

velox/functions/prestosql/json/JsonStringUtil.cpp
```diff
@@ -706,6 +708,7 @@ void unescapeForJsonFunctions(
       }
     }
   }
+  return pos - output;
```
kgpai, line 711: Curious, unescapeSizeForJsonFunctions gives you size before calling this, so why do we need to return size ?

MBkkt: Because we don't want unescape json twice, instead we're using kind of over resize and then adjust size to real one

velox/functions/prestosql/json/JsonStringUtil.cpp
MBkkt: As I mentioned in slack, we have our own sqllogic-like tests for these and I don't have time now to write tests for velox.

If it helps there was similar fix for other function
https://github.com/facebookincubator/velox/pull/12442

PR #15337, opened 2025-10-30
feat(cuDF): Implement post-filtered cuDF hash joins with streaming probe sides

shrshi:
This PR uses the libcudf hash join class to implement filtered left, right and inner joins. It also streams the probe side table for all join types. Additional book-keeping is required in the case of right joins to ensure that all rows of the probe table partitions are present in the join output.

PR #14797, opened 2025-09-09
Add velox-cudf support for TopN

karthikeyann:
This PR adds operator CudfTopN to run TopN in GPU.
Unit tests are added to verify the output.

The CudfTopN stores top N values of each input batch, and if number of batches exceed 5, the batches are concatenated together and topN values are stored back to the batch vector. getOutput will concatenated all batches together and return topN values.

The tests are same as TopN, but Setup makes sure they run in GPU.
This PR has additional fix 2288ca913ca4584c273c6690e75acb4f1dd07250 for cudfFilterProject which affects filter results with null values.

PR #15364, opened 2025-11-01
fix: Fix compile error in CI

MBkkt:


MBkkt:
@devavret Yes, I already noticed, and mention you in #15337 
Will it be easy to fix or it's more easy to revert and apply fix later?

devavret:
@MBkkt its an easy fix. I'll make it in 2 hours if it's okay?

karthikeyann:
I have the fix already in PR #14797
I will make it as separate PR.

PR #14998, opened 2025-09-29
fix: Replace dashes with underscores in setup script function names

simoneves:
Two functions in `scripts/setup-common.sh` have names containing dashes/hyphens.

While technically legal in `bash`, it causes problems if these scripts are called from a Docker `RUN` command, as whatever script interpreter is invoked there is fussier, and you get a failure of the form...
```
scripts/setup-common.sh: line 348: `install_gcs-sdk-cpp': not a valid identifier
```

PR #15365, opened 2025-11-01
build(scripts): Replace dashes with underscores in function names

assignUser:
Continuation of #14998
---
Two functions in `scripts/setup-common.sh` have names containing dashes/hyphens.

While technically legal in `bash`, it causes problems if these scripts are called from a Docker `RUN` command, as whatever script interpreter is invoked there is fussier, and you get a failure of the form...
```
scripts/setup-common.sh: line 348: `install_gcs-sdk-cpp': not a valid identifier
```

### DIFF
--- a/scripts/setup-fedora.sh
+++ b/scripts/setup-fedora.sh
@@ -68,7 +68,7 @@ function install_velox_deps_from_dnf {
 
 function install_velox_deps {
   run_and_time install_velox_deps_from_dnf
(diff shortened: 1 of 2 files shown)


peterenescu:
@assignUser are we concerned about the [failing benchmark](https://github.com/facebookincubator/velox/actions/runs/19193908306/job/54872081010)?

peterenescu:
Ah looks like it's unrelated and WIP https://github.com/facebookincubator/velox/pull/14319

PR #15367, opened 2025-11-01
build(ci): Add Linter annotations

### LINKED ISSUE #15049: Investigate matchers to add lint annotations
### Description

Annotation on the files directly will make it easier to surface issues without contributors needing to go to the ci log.
docs: https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

assignUser:
Closes #15049 
Currently stacked on #15284 

This will create annotation for issues the linters find. If there are a lot of issues not all will show up as an annotation as there are pretty strict limits (10 errors and warnings each per step and 50 total per job).

### DIFF
--- a/.github/matchers/clang-tidy.json
+++ b/.github/matchers/clang-tidy.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "clang-tidy",
+            "pattern": [
+                {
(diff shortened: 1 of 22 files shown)


assignUser:
It's a bit weird that when ever the license-header script creates an annotation, any other annotations will be suppressed. I tried a lot of different variations and it only happens with this check (see zizmor and yamllint in the same file). 

It's a bit annoying but license-header issues should be rarer than the other ones so it's not to bad. I don't think it's worth spending more time on it.

assignUser:
This is ready for review and merge once #15047 is done.

kgpai:
Commenting to keep open.

PR #15372, opened 2025-11-03
feat: Add customizable post processor to hive columns

Yuhta:
Summary:
In some ML use cases, we need to normalize the flatmap data using some
customized logic on the output data from Hive connector.  Add a configuration
point to `HiveColumnHandle` for this purpose, and add `LazyVector::chain` to
preserve the laziness while applying the transformation on data.

Differential Revision: D86102068


xiaoxmeng (approved):
@Yuhta LGTM % nits. Thanks!

PR #5298, opened 2023-06-16
Add ValueStream operator

jinchengchenghh:
Add ValueStreamNode to fetch data from Gluten Velox based shuffle to avoid fetch all the data at once.

Resolves: https://github.com/facebookincubator/velox/issues/4649
Relevant: https://github.com/facebookincubator/velox/pull/3917

PR #15375, opened 2025-11-03
feat(cudf): Support user defined gpu operator

jinchengchenghh:
In Gluten, after Shuffle, add ValueStreamNode and ValueStream operator, https://github.com/apache/incubator-gluten/blob/main/cpp/velox/operators/plannodes/RowVectorStream.h.

After shuffle reader support output CudfVector, the ValueStream operator outputs cudf::table, so do not need to add CudfFromVelox operator, to achieve this,add CudfValueStream extends CudfOperator as other GPU operator that accepts CudfVector and output CudfVector.

### DIFF
--- a/velox/experimental/cudf/exec/CudfOperator.h
+++ b/velox/experimental/cudf/exec/CudfOperator.h
@@ -16,9 +16,9 @@
 
 #pragma once
 
-#include "velox/core/PlanNode.h"
-
 #include "velox/experimental/cudf/exec/NvtxHelper.h"
+
+#include "velox/core/PlanNode.h"
 
 namespace facebook::velox::cudf_velox {
 


velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -164,6 +164,7 @@ bool CompileState::compile(bool allowCpuFallback) {
                    exec::LocalPartition,
                    exec::LocalExchange,
                    exec::AssignUniqueId>(op) ||
+            isAnyOf<NvtxHelper>(op) ||
```
devavret: this change would allow unsupported filter project and join to also pass the check since they inherit NvtxHelper

devavret (changes requested):
If you want other operators to be treated as consumers/producers of cudfvector without having to make changes to them, consider emulating what we do with local exchange.

jinchengchenghh:
The exec::LocalExchange class is a Velox class, but the operator VectorStream is a class in Gluten, so we cannot add it directly.

velox/experimental/cudf/exec/ToCudf.cpp
jinchengchenghh: The CPU operator won't inherit NvtxHelper, the GPU operator does not run into the check

jinchengchenghh:
This is why the class is in Gluten repo, not in Velox repo https://github.com/facebookincubator/velox/pull/5298#pullrequestreview-1488394943

devavret:
@jinchengchenghh instead of reusing NvtxHelper, would it make sense to add new classes that can be inherited from to indicate if an operator takes/emits CudfVector? 

Edit: Maybe named VectorPassthroughOperator

jinchengchenghh:
What about CudfOperator? And inherits NvtxHelper. If we have other operator that outputs CudfVector in the future, it can also inherits CudfOperator. @devavret

devavret:
`CudfOperator` sounds fine. But there should at least be a distinction of whether it inputs `CudfVector` or outputs `CudfVector`. What do you think? For now, it's okay to just add one and we can add the other later.

jinchengchenghh:
Yes, now we can suppose CudfOperator accepts CudfVector and output CudfVector, if there are other cases, we can add them then

jinchengchenghh:
Address the comments, could you help review again? Thanks! @devavret

velox/experimental/cudf/exec/CudfOperator.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace facebook::velox::cudf_velox {
+
+/// The user defined operator will inherit this operator, the operator accepts
+/// CudfOperator and output CudfVector.
+class CudfOperator : public NvtxHelper {
+ public:
+  CudfOperator(int32_t operatorId, const PlanNodeId& nodeId)
+      : facebook::velox::cudf_velox::NvtxHelper(
+            nvtx3::rgb{160, 82, 45}, // Sienna
```
devavret: If possible, take this as constructor argument too. With default being grey.

PR #15382, opened 2025-11-04
feat: Add rewrites for SWITCH and IF special forms

pramodsatya:
Rewrites SWITCH and IF special form expressions which have either of the shapes:
1. `SWITCH(condition1, value1, condition2, value2, .., conditionN, valueN, .., optional<defaultValue>)`
2. `IF(condition, trueValue, falseValue)` 

Removes `(conditionN, valueN)` pairs from inputs when `conditionN` is false. 
If `conditionN` is true and all the previous conditions were false, the expression is simplified to `valueN`. 
If `conditionN` is true and there is at least one condition before `conditionN` that can not be evaluated without input data, modifies else clause to `valueN` and prunes rest of the inputs. 
Returns an optimized SWITCH expression with pruned inputs.

### DIFF
--- a/velox/core/Expressions.h
+++ b/velox/core/Expressions.h
@@ -151,6 +151,9 @@ class ConstantTypedExpr : public ITypedExpr {
 
   static TypedExprPtr create(const folly::dynamic& obj, void* context);
 
+  /// Returns a NULL constant expression of given type.
+  static TypedExprPtr makeNull(const TypePtr& type);
+
  private:
   const Variant value_;
   const VectorPtr valueVector_;
--- a/velox/expression/SwitchRewrite.cpp
+++ b/velox/expression/SwitchRewrite.cpp
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <set>
+
+#include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/ExprUtils.h"
+#include "velox/expression/SwitchRewrite.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr SwitchRewrite::rewrite(const core::TypedExprPtr& expr) {
+  if (!expr->isCallKind()) {
+    return nullptr;
+  }
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  if (!(callExpr->name() == kSwitch) && !(callExpr->name() == kIf)) {
+    return nullptr;
+  }
+
+  const auto& inputs = expr->inputs();
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+
+  // Iterate over all `(condition, value)` pairs. Handle `else` value at the
+  // end.
+  for (auto i = 0; i < numInputs - 1; i += 2) {
+    const auto& condition = inputs.at(i);
+    const auto& value = inputs.at(i + 1);
+
+    if (condition->isConstantKind()) {
+      const auto* constCondition =
+          condition->asUnchecked<core::ConstantTypedExpr>();
+      if (auto boolCondition = constCondition->toBool()) {
+        if (boolCondition.value()) {
+          // If this condition is true and all conditions before this are false,
+          // simplify `switch` expression to the corresponding `value`.
+          if (optimizedInputs.empty()) {
+            return value;
+          }
+          // If this condition is true and all conditions before this can only
+          // be evaluated at runtime, make this the new `else` value and stop
+          // checking further conditions.
+          optimizedInputs.push_back(value);
+          break;
+        } else {
+          // Skip false conditions.
+          continue;
+        }
+      } else {
+        // Skip NULL conditions.
+        continue;
+      }
+    } else {
+      optimizedInputs.push_back(condition);
+      optimizedInputs.push_back(value);
+    }
+  }
+
+  // Handle `else` value if present.
+  if (optimizedInputs.size() % 2 == 0 && numInputs % 2 == 1) {
+    const auto elseValue = inputs.at(numInputs - 1);
+    // Return `else` value if there are no conditions.
+    if (optimizedInputs.empty()) {
+      return elseValue;
+    }
+    optimizedInputs.emplace_back(elseValue);
+  }
+  // Return NULL if there are no conditions and `else` value is not present.
+  if (optimizedInputs.empty()) {
+    return core::ConstantTypedExpr::makeNull(expr->type());
+  }
+
+  return std::make_shared<core::CallTypedExpr>(
+      expr->type(), std::move(optimizedInputs), callExpr->name());
+}
+
+void SwitchRewrite::registerRewrite() {
+  expression::ExprRewriteRegistry::instance().registerRewrite(
+      expression::SwitchRewrite::rewrite);
+}
+
+} // namespace facebook::velox::expression
--- a/velox/expression/SwitchRewrite.h
+++ b/velox/expression/SwitchRewrite.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::expression {
+
+class SwitchRewrite {
+ public:
+  /// Rewrites SWITCH and IF special form expressions which have the shape
+  /// `SWITCH(condition1, value1, condition2, value2, .., conditionN, valueN, ..
+  /// .., optional<elseValue>)`, and `IF(condition, trueValue, falseValue)`
+  /// respectively. IF expression is rewritten to `trueValue` or `falseValue`
+  /// if `condition` is either true or false respectively. The conditions of
+  /// SWITCH expression are traversed in order and if the condition is:
+  ///  1. false, it is removed from the inputs along with its corresponding
+  ///     value.
+  ///  2. true: If there are no inputs before this, the corresponding value
+  ///     is returned. Otherwise, the corresponding value is the new else value
+  ///     and all subsequent inputs are ignored.
+  ///  3. not a constant boolean and it cannot be evaluated without input data,
+  ///     the condition and its corresponding value are retained in the inputs.
+  /// If there are no conditions in the expression now, the else value is
+  /// returned if it is present; otherwise NULL is returned. The else value is
+  /// added to the inputs otherwise and the optimized SWITCH expression with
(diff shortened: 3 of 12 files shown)


velox/expression/tests/ExprTest.cpp
```diff
@@ -1642,13 +1642,8 @@ TEST_P(ParameterizedExprTest, switchExpr) {
...
   assertEqualVectors(expected, result);
 
-  // No "else" clause and no match.
-  result = evaluate("case 0 when 100 then 1 when 200 then 2 end", vector);
-  expected = makeAllNullFlatVector<int64_t>(size);
-  assertEqualVectors(expected, result);
-
   result = evaluate("case c2 when 100 then 1 when 200 then 2 end", vector);
-  assertEqualVectors(expected, result);
+  // assertEqualVectors(expected, result);
```
aditi-pandit: Why is this commented out ?

pramodsatya: This test looks erroneous to me and I have fixed it now, could you please confirm if the fix is valid @aditi-pandit? 
I had commented it out and kept the PR draft as I wanted to debug this further and check if the fuzzer uncovers any additional failures resulting from this rewrite.

pramodsatya: Apologies for the oversight, reverted this.

pramodsatya (commented):
Thanks for the feedback @aditi-pandit, addressed the comments. Could you PTAL?

velox/expression/tests/SpecialFormRewriteTestBase.h
```diff
@@ -0,0 +1,63 @@
...
+    const auto rewritten =
+        expression::ExprRewriteRegistry::instance().rewrite(typedExpr);
+    const auto expectedExpr = makeTypedExpr(expected, type);
+    if (*rewritten != *expectedExpr) {
+      LOG(ERROR) << fmt::format("Input: {}", typedExpr->toString());
+      LOG(ERROR) << fmt::format("Rewritten: {}", rewritten->toString());
+      LOG(ERROR) << fmt::format("Expected: {}", expectedExpr->toString());
+    }
+
+    ASSERT_TRUE(*rewritten == *expectedExpr);
```
aditi-pandit: Use VELOX_CHECK_EQ here instead of ASSERT_TRUE.

aditi-pandit (commented):
Thanks @pramodsatya. Have just one minor comment.

velox/expression/tests/SpecialFormRewriteTestBase.h
pramodsatya: Thanks for the suggestion, using `VELOX_CHECK_EQ` fails here because it expects a formatter to be defined for `ITypedExpr`, is it fine to use `VELOX_CHECK`?
```
In file included from /Users/pramod/velox_dependency_install/include/fmt/format.h:41:
/Users/pramod/velox_dependency_install/include/fmt/base.h:2262:45: error: implicit instantiation of undefined template 'fmt::detail::type_is_unformattable_for<facebook::velox::core::ITypedExpr, char>'
 2262 |     type_is_unformattable_for<T, char_type> _;
      |                                             ^
/Users/pramod/velox_dependency_install/include/fmt/base.h:2235:44: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<facebook::velox::core::ITypedExpr, 0>' requested here
 2235 |   FMT_CONSTEXPR20 FMT_INLINE value(T& x) : value(x, custom_tag()) {}
      |                                            ^
/Users/pramod/velox_dependency_install/include/fmt/base.h:2808:12: note: in instantiation of function template specialization 'fmt::detail::value<fmt::context>::value<const facebook::velox::core::ITypedExpr, 0>' requested here
 2808 |   return {{args...}};
      |            ^
/Users/pramod/ws/velox/./velox/common/base/ExceptionHelper.h:54:33: note: in instantiation of function template specialization 'fmt::make_format_args<fmt::context, const facebook::velox::core::ITypedExpr, const facebook::velox::core::ITypedExpr, 2, 0, 255ULL>' requested here
   54 |   return fmt::vformat(fmt, fmt::make_format_args(args...));
      |                                 ^
/Users/pramod/ws/velox/./velox/expression/tests/SpecialFormRewriteTestBase.h:58:5: note: in instantiation of function template specialization 'facebook::velox::errorMessage<facebook::velox::core::ITypedExpr, facebook::velox::core::ITypedExpr>' requested here
   58 |     VELOX_CHECK_EQ(*rewritten, *expectedExpr, "err");
```

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,99 @@
...
+          // checking further `then` clauses.
+          hasOptimizedElseClause = true;
+          optimizedInputs.push_back(value);
+          break;
+        } else {
+          // Ignore `then` clauses with false value.
+          continue;
+        }
+      } else {
+        continue;
```
aditi-pandit, line 65: Don't understand what this continue is about. Can you add a comment about what it represents ?

aditi-pandit (commented):
Thanks @pramodsatya

aditi-pandit:
@mbasmanova : I did a round of review and this code looks about alright to me. PTAL.

velox/expression/SwitchRewrite.cpp
pramodsatya: This is for NULL conditions, added a comment, thanks.

velox/expression/SwitchRewrite.h
```diff
@@ -0,0 +1,36 @@
...
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::expression {
+
+class SwitchRewrite {
+ public:
+  /// Rewrites SWITCH and IF special form expressions which have the shape
+  /// SWITCH(condition1, value1, condition2, value2, .., conditionN, valueN, ..,
+  /// optional<defaultValue>), and IF(condition, trueValue, falseValue). Removes
+  /// `(conditionN, valueN)` pairs from inputs when `conditionN` is false. If
```
mbasmanova: > If `conditionN` is true and all the previous conditions were false,

This is confusing because we just said that step 1 is to remove all cases where condition is false.

I'd expect the following 2 steps:

- Remove all pairs where condition is false.
- Remove everything after the first condition which is true.
- If last pair's condition is true, replace it with an else.

velox/expression/SwitchRewrite.h
```diff
@@ -0,0 +1,36 @@
...
+  /// `conditionN` is true and all the previous conditions were false, the
+  /// expression is simplified to `valueN`. If `conditionN` is true and there is
+  /// at least one condition before `conditionN` that can not be evaluated
+  /// without input data, modifies else clause to `valueN` and prunes rest of
+  /// the inputs. Returns an optimized SWITCH expression with pruned inputs.
```
mbasmanova: > Returns an optimized SWITCH expression with pruned inputs.

This is redundant and not quite accurate. I assume if(true, a, b) returns a field reference expression 'a', not a switch expression.

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "velox/expression/SwitchRewrite.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr SwitchRewrite::rewrite(const core::TypedExprPtr& expr) {
+  if (!expr->isCallKind()) {
+    return nullptr;
+  }
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  // `if` is a special form of `switch` with a single condition.
```
mbasmanova: > special form

this term is overloaded here; consider changing to 

> IF is implemented as a SWITCH with a single case statement.

mbasmanova: Although, wondering why do we need this comment at all? Should it go into the header file where we would explain that rewrite applies to both SWITCH and IF?

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (!(callExpr->name() == kSwitch) && !(callExpr->name() == kIf)) {
+    return nullptr;
+  }
+
+  const auto& inputs = expr->inputs();
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  bool hasOptimizedElseClause = false;
+
+  // Iterate over all `(condition, value)` pairs. Handle `else` clause at end.
```
mbasmanova: typo: "at end" -> "at the end" or "last"

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,100 @@
...
+        // Skip NULL conditions.
+        continue;
+      }
+    } else {
+      optimizedInputs.push_back(condition);
+      optimizedInputs.push_back(value);
+    }
+  }
+
+  // Handle default `else` clause if present.
```
mbasmanova: drop "default" since it is redundant

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // If all conditions are false, return value from `else` clause.
+    if (optimizedInputs.empty()) {
+      return elseValue;
+    }
+    optimizedInputs.emplace_back(elseValue);
+  }
+  // Return NULL if all conditions are false and no default `else` clause is
+  // present.
+  if (optimizedInputs.empty()) {
+    return std::make_shared<core::ConstantTypedExpr>(
```
mbasmanova: nit: let's add a static helper method ConstantTypedExpr::makeNull(type) to reduce boiler plate

velox/expression/SwitchRewrite.cpp
```diff
@@ -0,0 +1,100 @@
...
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  // `if` is a special form of `switch` with a single condition.
+  if (!(callExpr->name() == kSwitch) && !(callExpr->name() == kIf)) {
+    return nullptr;
+  }
+
+  const auto& inputs = expr->inputs();
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  bool hasOptimizedElseClause = false;
```
mbasmanova: Do we need this boolean? It seems like it can be computed after the loop as optimizedInputs % 2 == 1.

velox/expression/tests/SpecialFormRewriteTestBase.h
```diff
@@ -0,0 +1,62 @@
...
+    const auto rewritten =
+        expression::ExprRewriteRegistry::instance().rewrite(typedExpr);
+    const auto expectedExpr = makeTypedExpr(expected, type);
+    if (*rewritten != *expectedExpr) {
+      SCOPED_TRACE(fmt::format("Input: {}", typedExpr->toString()));
+      SCOPED_TRACE(fmt::format("Rewritten: {}", rewritten->toString()));
+      SCOPED_TRACE(fmt::format("Expected: {}", expectedExpr->toString()));
+    }
+
+    VELOX_CHECK(*rewritten == *expectedExpr);
```
mbasmanova: Can we use ASSERT_TRUE here?

velox/expression/tests/SpecialFormRewriteTestBase.h
```diff
@@ -0,0 +1,62 @@
...
+  /// Validates special form expression rewrites that can short circuit
+  /// expression evaluation.
+  /// @param expr Input SQL expression to be rewritten.
+  /// @param expected Expected SQL expression after rewriting `expr`.
+  /// @param type Row type containing input fields referenced by `expr`.
+  void testRewrite(
+      const std::string& expr,
+      const std::string& expected,
+      const RowTypePtr& type = ROW({})) {
+    const auto typedExpr = makeTypedExpr(expr, type);
```
mbasmanova: please, move implementation to .cpp file

mbasmanova:
@aditi-pandit @pramodsatya Should we add a section to documentation about expression optimizer and all these awesome re-writes you are adding?

velox/expression/SwitchRewrite.cpp
pramodsatya: Removed this comment.

velox/expression/SwitchRewrite.cpp
pramodsatya: Reverted.

velox/expression/SwitchRewrite.h
pramodsatya: Updated accordingly, thanks.

mbasmanova:
> Would it be fine to add documentation for ExprOptimizer and the special form rewrites in a follow-up PR?

@pramodsatya Yup.

velox/core/Expressions.h
```diff
@@ -151,6 +151,8 @@ class ConstantTypedExpr : public ITypedExpr {
 
   static TypedExprPtr create(const folly::dynamic& obj, void* context);
 
+  static TypedExprPtr makeNull(const TypePtr& type);
```
mbasmanova, line 155: please, document

PR #15386, opened 2025-11-04
fix: Change columnHandles to filterColumnHandles in HiveTableHandle

Yuhta:
Summary:
Turns out the requirement to have full schema in column handles and
pointer equality is not pratical for engines.  Change the semantic of the new
field to be extra column handles that are not appearing in assignments already.

Differential Revision: D86209890

### DIFF
--- a/velox/connectors/hive/HiveDataSource.cpp
+++ b/velox/connectors/hive/HiveDataSource.cpp
@@ -50,6 +50,19 @@ bool shouldEagerlyMaterialize(
   return false;
 }
 
+void checkColumnHandleConsistent(
+    const HiveColumnHandle& x,
+    const HiveColumnHandle& y) {
+  VELOX_CHECK(
+      x.columnType() == y.columnType(),
+      "Inconsistent column handle: {}",
+      x.name());
+  VELOX_CHECK_EQ(
+      *x.dataType(), *y.dataType(), "Inconsistent column handle: {}", x.name());
+  VELOX_CHECK_EQ(
+      *x.hiveType(), *y.hiveType(), "Inconsistent column handle: {}", x.name());
+}
+
 } // namespace
 
 void HiveDataSource::processColumnHandle(const HiveColumnHandlePtr& handle) {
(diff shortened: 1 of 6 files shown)


velox/connectors/hive/HiveDataSource.cpp
```diff
@@ -91,29 +91,24 @@ HiveDataSource::HiveDataSource(
...
+    auto handle =
+        std::dynamic_pointer_cast<const HiveColumnHandle>(columnHandle);
+    VELOX_CHECK_NOT_NULL(
+        handle,
+        "ColumnHandle must be an instance of HiveColumnHandle for {}",
+        canonicalizedName);
     VELOX_CHECK(
-        assignmentHandles.empty(),
-        "Assignments must be a subset of column handles");
+        !extraColumnHandleNames.contains(handle->name()),
```
mbasmanova: Is this check necessary? Can we just require that overlap exists, the handles are identical (not pointer equality, but value equality)?

velox/connectors/hive/HiveDataSource.cpp
```diff
@@ -91,29 +91,24 @@ HiveDataSource::HiveDataSource(
...
-    for (const auto& [_, columnHandle] : assignments) {
-      assignmentHandles.insert(columnHandle.get());
-    }
-    for (auto& handle : hiveTableHandle_->columnHandles()) {
-      assignmentHandles.erase(handle.get());
-      processColumnHandle(handle);
-    }
+  folly::F14FastSet<std::string_view> extraColumnHandleNames;
+  for (auto& handle : hiveTableHandle_->extraColumnHandles()) {
+    processColumnHandle(handle);
```
mbasmanova: processColumnHandle assumes handles are "unique"... but there is not check for that; similarly, we are lacking checks that assignments have consistent handles, e.g. if same column is projected twice under different names the handles are the same.

velox/connectors/hive/TableHandle.cpp
```diff
@@ -190,6 +190,13 @@ folly::dynamic HiveTableHandle::serialize() const {
     tableParameters[param.first] = param.second;
   }
   obj["tableParameters"] = tableParameters;
+  if (!extraColumnHandles_.empty()) {
```
mbasmanova: Is this a fix for a serialization bug? Would you mention this in the PR description or extract into a separate PR?

Yuhta: Yes we missed this part, I mentioned in the PR description

velox/connectors/hive/HiveDataSource.cpp
Yuhta: Should we add a check here?  Or we should add check when we construct the plan node?

velox/connectors/hive/HiveDataSource.cpp
Yuhta: If we can ensure this it would be optimal.  Having duplicated objects are still correct but less optimal (more expensive to construct and more expensive to check).

mbasmanova: Agree. 'assignments' is something produced by the engine with some cooperation with the connector. TableHandle is produced by the connector. Having hard dependencies between these is not desirable as it is requires the engine to first produce 'assignments', then ask connector to produce TableHandle and provide assignments as inputs.

velox/connectors/hive/HiveDataSource.cpp
mbasmanova: Hive connector doesn't participate in construction of the plan node (at least not directly).

velox/connectors/hive/TableHandle.cpp
mbasmanova: Would you add a test for serde? It really is better extracted into a separate PR.

velox/connectors/hive/TableHandle.h
```diff
@@ -148,7 +148,7 @@ class HiveTableHandle : public ConnectorTableHandle {
       const core::TypedExprPtr& remainingFilter,
       const RowTypePtr& dataColumns = nullptr,
       const std::unordered_map<std::string, std::string>& tableParameters = {},
-      std::vector<HiveColumnHandlePtr> columnHandles = {});
+      std::vector<HiveColumnHandlePtr> extraColumnHandles = {});
```
mbasmanova: nit: filterColumnHandles might be clearer

velox/connectors/hive/HiveDataSource.cpp
```diff
@@ -91,29 +91,43 @@ HiveDataSource::HiveDataSource(
...
-  } else {
-    folly::F14FastSet<const connector::ColumnHandle*> assignmentHandles;
-    for (const auto& [_, columnHandle] : assignments) {
-      assignmentHandles.insert(columnHandle.get());
-    }
-    for (auto& handle : hiveTableHandle_->columnHandles()) {
-      assignmentHandles.erase(handle.get());
+  folly::F14FastMap<std::string_view, const HiveColumnHandle*> columnHandles;
+  // Column handled keyed on the column alias, the name used in the query.
+  for (const auto& [canonicalizedName, columnHandle] : assignments) {
```
mbasmanova, line 109: Why "canonicalized"? This is output name, no?

velox/connectors/hive/HiveDataSource.cpp
```diff
@@ -91,29 +91,43 @@ HiveDataSource::HiveDataSource(
...
+    auto handle =
+        std::dynamic_pointer_cast<const HiveColumnHandle>(columnHandle);
+    VELOX_CHECK_NOT_NULL(
+        handle,
+        "ColumnHandle must be an instance of HiveColumnHandle for {}",
+        canonicalizedName);
+    VELOX_CHECK(
+        !columnHandles.contains(handle->name()),
+        "Cannot map from same table column to different outputs in table scan; a project node should be used for such usage: {}",
```
mbasmanova: "for such usage" -> instead

velox/connectors/hive/HiveDataSource.cpp
```diff
@@ -91,29 +91,43 @@ HiveDataSource::HiveDataSource(
...
+        handle->name());
+    processColumnHandle(handle);
+    columnHandles[handle->name()] = handle.get();
```
mbasmanova: perhaps, combine checking 'contains' and inserting to avoid double lookup

mbasmanova (approved):
Thanks.

velox/connectors/hive/HiveDataSource.cpp
Yuhta: Yes the output name, it's called that since I joined velox team but I don't find it's too confusing so did not change it

PR #15388, opened 2025-11-04
fix: Back out "fix: Reduce lock contention in ScanTracker"

Yuhta:
Summary:
Original commit changeset: a6ed6383f125

Original Phabricator Diff: D84661649

Differential Revision: D86217393


amitkdutta (approved):
@Yuhta We should put the stack trace as well to show the revert reason.

Issue #15384, opened 2025-11-04
Consolidate toSubfieldFilter into ExprToSubfieldFilterParser


Yuhta:
### Description

Today `toSubfieldFilter` is a separate function from `ExprToSubfieldFilterParser` (because it exists before the parser).  To make a more unified interface, we should move `toSubfieldFilter` into `ExprToSubfieldFilterParser` as well, so that we can potentially making it virtual if needed.

Also it would be nice to register a parser directly instead of registering a parser factory.

CC: @mbasmanova

mbasmanova:
Note that this change is necessary because toSubfieldFilter is not language agnostic. It hard-codes the meaning of "not" function, which is not guaranteed to be the same across dialects (Presto, Spark, Postgres, etc.). On the other hand, ExprToSubfieldFilterParser is supposed to be language agnostic.

CC: @MBkkt

mbasmanova:
I believe we also do not want to register PrestoExprToSubfieldFilterParser as default parser and instead expect it to be registered along with all Presto functions and use 'prefix' as needed.

MBkkt:
@mbasmanova I think "not" shouldn't be language specific, I think it's convinient to have language agnostic "and", "or", "not". Because de morgan's laws.

PR #15390, opened 2025-11-04
refactor: Replace factory with singleton in ExprToSubfieldFilterParser

mbasmanova:
Summary: Part of https://github.com/facebookincubator/velox/issues/15384

Differential Revision: D86218253

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -112,13 +112,11 @@ toInt64List(const VectorPtr& vector, vector_size_t start, vector_size_t size) {
   return values;
 }
 
-static std::shared_ptr<ExprToSubfieldFilterParser> defaultParser =
-    std::make_shared<PrestoExprToSubfieldFilterParser>();
-
 } // namespace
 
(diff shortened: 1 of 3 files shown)


velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -112,13 +112,11 @@ toInt64List(const VectorPtr& vector, vector_size_t start, vector_size_t size) {
...
-static std::shared_ptr<ExprToSubfieldFilterParser> defaultParser =
-    std::make_shared<PrestoExprToSubfieldFilterParser>();
-
 } // namespace
 
-std::function<std::shared_ptr<ExprToSubfieldFilterParser>()>
-    ExprToSubfieldFilterParser::parserFactory_ = [] { return defaultParser; };
+std::shared_ptr<ExprToSubfieldFilterParser>
+    ExprToSubfieldFilterParser::parser_ =
+        std::make_shared<PrestoExprToSubfieldFilterParser>();
```
MBkkt, line 119: instead of static shared_ptr better to have static by value and aliasing shared ptr.

Yuhta: How should that be done?

PR #15392, opened 2025-11-04
refactor: Deprecate exec::toSubfieldFilter API

mbasmanova:
Summary:
Part of https://github.com/facebookincubator/velox/issues/15384

exec::toSubfieldFilter API appears language agnostic, but it assumes that function registered under name "not" has semantics of Presto's "not" function. This assumption is not valid and can lead to difficult to debug correctness issues.

The fix is to move that API to ExprToSubfieldFilterParser interface which has language-specific implementations.

Differential Revision: D86244462

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -50,10 +50,6 @@ T singleValue(const VectorPtr& vector) {
   return simpleVector->valueAt(0);
 }
 
-const core::CallTypedExpr* asCall(const core::ITypedExpr* expr) {
-  return dynamic_cast<const core::CallTypedExpr*>(expr);
-}
-
 common::BigintRange* asBigintRange(std::unique_ptr<common::Filter>& filter) {
   return dynamic_cast<common::BigintRange*>(filter.get());
 }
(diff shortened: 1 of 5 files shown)


velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -528,10 +534,11 @@ PrestoExprToSubfieldFilterParser::leafCallToSubfieldFilter(
   return nullptr;
 }
 
-std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
+std::pair<common::Subfield, std::unique_ptr<common::Filter>>
+PrestoExprToSubfieldFilterParser::toSubfieldFilter(
     const core::TypedExprPtr& expr,
     core::ExpressionEvaluator* evaluator) {
-  if (auto call = asCall(expr.get())) {
+  if (auto* call = expr->asUnchecked<core::CallTypedExpr>()) {
```
Yuhta: This is changed from a dynamic cast to a static one

mbasmanova: Let me add a check for kind.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -543,15 +551,11 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
     common::Subfield subfield;
     std::unique_ptr<common::Filter> filter;
     if (call->name() == "not") {
-      if (auto* inner = asCall(call->inputs()[0].get())) {
-        filter =
-            ExprToSubfieldFilterParser::getInstance()->leafCallToSubfieldFilter(
-                *inner, subfield, evaluator, true);
+      if (auto* inner = call->inputs()[0]->asUnchecked<core::CallTypedExpr>()) {
```
Yuhta: Here as well, should we just keep `asCall` and check kind there instead?

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: Actually, asUnchecked is a dynamic_cast:

```
  template <typename T>
  const T* asUnchecked() const {
    return dynamic_cast<const T*>(this);
  }
```

PR #15396, opened 2025-11-05
perf: Use SIMD for TIME based comparison operators (#15396)

pratikpugalia:
Summary:

Use simd for time based comparison as it is a primitive type  and can be simdized similar to interval_year_month, interval_day_time

Reviewed By: kgpai

Differential Revision: D86261353


pratikpugalia:
@claude /pr-review

PR #15399, opened 2025-11-05
feat: Add rewrite for COALESCE special form

pramodsatya:
Rewrites `COALESCE` special form expression. First, flattens nested `COALESCE` expressions in a recursive manner, `eg: COALESCE(a, COALESCE(b, c)) -> COALESCE(a, b, c)`. 

The inputs to `COALESCE` are then traversed in order and:
1. If the input is constant:    
a. and it is `NULL`, the input is removed.    
b. and if there are no inputs before this, the `COALESCE` expression is simplified to this constant input.    
c. otherwise, all subsequent inputs after this are removed. 
2. Otherwise, if the same input is seen before, it is removed i.e non-constant inputs to `COALESCE` are deduplicated.

If there is a single input to `COALESCE` after pruning inputs, `COALESCE` is simplified to this expression. 
Otherwise, returns an optimized `COALESCE` expression with pruned inputs.

### DIFF
--- a/velox/expression/CoalesceRewrite.cpp
+++ b/velox/expression/CoalesceRewrite.cpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <set>
+
+#include "velox/expression/CoalesceRewrite.h"
+#include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/ExprUtils.h"
+
+namespace facebook::velox::expression {
+
+core::TypedExprPtr CoalesceRewrite::rewrite(const core::TypedExprPtr& expr) {
+  if (!expr->isCallKind()) {
+    return nullptr;
+  }
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  if (callExpr->name() != kCoalesce) {
+    return nullptr;
+  }
+
+  // Deduplicate inputs to COALESCE and remove NULL inputs.
+  std::vector<core::TypedExprPtr> flat;
+  utils::flattenInput(expr, kCoalesce, flat);
+  folly::F14FastSet<
+      const core::ITypedExpr*,
+      core::ITypedExprHasher,
+      core::ITypedExprComparer>
+      uniqueInputs;
+  std::vector<core::TypedExprPtr> deduplicatedInputs;
+
+  for (const auto& input : flat) {
+    if (input->isConstantKind()) {
+      const auto* constantExpr = input->asUnchecked<core::ConstantTypedExpr>();
+      if (constantExpr->isNull()) {
+        // Drop NULL constant.
+        continue;
+      }
+
+      if (uniqueInputs.empty()) {
+        // Return first non-NULL constant input.
+        return input;
+      }
+
+      // Drop inputs after non-NULL constant input.
+      uniqueInputs.insert(input.get());
+      deduplicatedInputs.push_back(input);
+      break;
+    } else {
+      if (uniqueInputs.emplace(input.get()).second) {
+        deduplicatedInputs.push_back(input);
+      }
+    }
+  }
+
+  // Return NULL if all inputs to COALESCE are NULL.
+  if (uniqueInputs.empty()) {
+    return flat.front();
+  }
+  // If there is a single input to COALESCE, return this expression.
+  if (uniqueInputs.size() == 1) {
+    return deduplicatedInputs.front();
+  }
+
+  return std::make_shared<core::CallTypedExpr>(
+      expr->type(), std::move(deduplicatedInputs), kCoalesce);
+}
+
+void CoalesceRewrite::registerRewrite() {
+  expression::ExprRewriteRegistry::instance().registerRewrite(
+      expression::CoalesceRewrite::rewrite);
+}
+
+} // namespace facebook::velox::expression
--- a/velox/expression/tests/CoalesceRewriteTest.cpp
+++ b/velox/expression/tests/CoalesceRewriteTest.cpp
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include "velox/expression/tests/SpecialFormRewriteTestBase.h"
+
+namespace facebook::velox::expression {
+namespace {
+
+class CoalesceRewriteTest
+    : public expression::test::SpecialFormRewriteTestBase {};
+
+TEST_F(CoalesceRewriteTest, basic) {
+  auto type = ROW({"a", "b"}, {BIGINT(), BIGINT()});
+  testRewrite("coalesce(null::bigint, 1, a)", "1", type);
+  testRewrite(
+      "coalesce(null::bigint, 6 * a, 5 * b, 1, null::bigint)",
+      "coalesce(6 * a, 5 * b, 1)",
+      type);
+  testRewrite(
+      "coalesce(null::bigint, coalesce(null::bigint, a, coalesce(1, b)))",
+      "coalesce(a, 1)",
+      type);
+  testRewrite("coalesce(a, null::bigint, a)", "a", type);
+  testRewrite(
+      "coalesce(a, b, a, coalesce(coalesce(b, a), b))", "coalesce(a, b)", type);
+  testRewrite(
+      "coalesce(a, coalesce(2, coalesce(b, a)), b)", "coalesce(a, 2)", type);
+}
+
+} // namespace
+} // namespace facebook::velox::expression
--- a/velox/core/ITypedExpr.cpp
+++ b/velox/core/ITypedExpr.cpp
@@ -35,4 +35,14 @@ const auto& exprKindNames() {
 } // namespace
 
 VELOX_DEFINE_ENUM_NAME(ExprKind, exprKindNames);
+
+size_t ITypedExprHasher::operator()(const ITypedExpr* expr) const {
+  return expr->hash();
+}
+
+bool ITypedExprComparer::operator()(
+    const ITypedExpr* lhs,
+    const ITypedExpr* rhs) const {
+  return *lhs == *rhs;
+}
 } // namespace facebook::velox::core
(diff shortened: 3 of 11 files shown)


velox/expression/tests/CoalesceRewriteTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <gtest/gtest.h>
+
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/functions/prestosql/registration/RegistrationFunctions.h"
+#include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+
+namespace facebook::velox::expression {
+namespace {
+
+class CoalesceRewriteTest : public functions::test::FunctionBaseTest {
```
aditi-pandit: You should derived from SpecialFormRewriteTestBase

aditi-pandit:
@pramodsatya : Post rebase this PR is good for review.

pramodsatya:
Thanks @aditi-pandit. @mbasmanova, could you please help review this rewrite for `COALESCE` expressions?

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include "velox/expression/CoalesceRewrite.h"
+#include "velox/expression/ExprConstants.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/ExprUtils.h"
+
+namespace facebook::velox::expression {
+
+namespace {
+struct TypedExprHasher {
```
mbasmanova: Is this a copy-paste from velox/expression/ExprCompiler.cpp ? Let's refactor to avoid copy-paste. We can move these to the same header file where ITypedExpr is defined.

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,97 @@
...
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  if (callExpr->name() != kCoalesce) {
+    return nullptr;
+  }
+
+  // Deduplicate inputs to COALESCE and remove NULL inputs.
+  std::vector<core::TypedExprPtr> flat;
+  utils::flattenInput(expr, kCoalesce, flat);
+  folly::F14FastSet<core::TypedExprPtr, TypedExprHasher, TypedExprComparator>
+      deduplicatedInputSet;
```
mbasmanova: deduplicatedInputSet -> uniqueInputs

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,97 @@
...
+  std::vector<core::TypedExprPtr> deduplicatedInputList;
```
mbasmanova: deduplicatedInputList -> deduplicatedInputs

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  for (const auto& input : flat) {
+    if (input->isConstantKind()) {
+      const auto* constantExpr = input->asUnchecked<core::ConstantTypedExpr>();
+      if (!constantExpr->isNull()) {
```
mbasmanova: I find it hard to follow this flow. Consider restructuring.

```
if (constantExpr->isNull()) {
   // Drop NULL constant.
   continue;
}

if (deduplicatedInputSet.empty()) {
   // Return first non-NULL constant input.
   return input;
}

// Drop inputs after non-NULL constant input.
deduplicatedInputSet.insert(input);
deduplicatedInputList.push_back(input);
break;
```

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,97 @@
...
+          deduplicatedInputSet.insert(input);
+          deduplicatedInputList.push_back(input);
+          break;
+        }
+        // Return first non-NULL constant input if all inputs before this
+        // were NULL.
+        return input;
+      }
+    } else {
+      if (!deduplicatedInputSet.contains(input)) {
```
mbasmanova: avoid double lookup; call emplace and check result

velox/expression/CoalesceRewrite.cpp
pramodsatya: Updated accordingly, thanks.

velox/expression/CoalesceRewrite.cpp
pramodsatya: Thanks, restructured accordingly.

velox/expression/CoalesceRewrite.cpp
pramodsatya: Thanks, updated.

velox/expression/CoalesceRewrite.cpp
```diff
@@ -0,0 +1,87 @@
...
+        // Return first non-NULL constant input.
+        return input;
+      }
+
+      // Drop inputs after non-NULL constant input.
+      uniqueInputs.insert(input.get());
+      deduplicatedInputs.push_back(input);
+      break;
+    } else {
+      const auto [it, inserted] = uniqueInputs.emplace(input.get());
```
mbasmanova: it variable is not needed

const bool inserted = uniqueInputs.emplace(input.get()).second;

or

if (uniqueInputs.emplace(input.get()).second) {
...
}

mbasmanova (approved):
Thanks.

velox/expression/CoalesceRewrite.cpp
pramodsatya: Updated, thanks.

peterenescu:
@pramodsatya Are you able to run expression fuzzer locally on coalesce? It looks like we had some longer-running expression fuzzers fail with arithmetic error for simplified execution
```
Executing expression 0 : switch(map_key_exists(null,ceiling(coalesce(0.4492889642715454,0.05161726474761963))),transform_values(if("c0",{[382->530] [530->295] 644797815 => {[382->458] [458->383] [383->108] 0.05657687829807401, null, false, -805899563}}),lambda ROW<__a0:INTEGER,__a1:ROW<f0:DOUBLE,f1:INTERVAL DAY TO SECOND,f2:BOOLEAN,f3:INTERVAL DAY TO SECOND>> -> 2011-04-17),array_has_duplicates($internal$canonicalize(array_intersect(find_first({[47->37] {null}},array_sum({[400->715] [715->234] [234->565] 6976801220365335985, [401->317] [317->100] [100->151] 5371919366574234310, [402->701] [701->397] [397->300] 780637959026489352, [403->688] [688->606] [606->257] 6855833019528270910, null, ...3 more}),lambda ROW<__a0:ARRAY<ARRAY<VARCHAR>>> -> is_null("__a0"))))),"c1")
Common eval succeeded.
Only simplified path threw exception:
...
Error Code: ARITHMETIC_ERROR
Reason: integer overflow: 6976801220365335985 + 5371919366574234310
Retriable: False
Context: array_sum({[400->715] [715->234] [234->565] 6976801220365335985, [401->317] [317->100] [100->151] 5371919366574234310, [402->701] [701->397] [397->300] 780637959026489352, [403->688] [688->606] [606->257] 6855833019528270910, null, ...3 more}:ARRAY<BIGINT>)
Function: checkedPlus
```

pramodsatya:
> @pramodsatya Are you able to run expression fuzzer locally on coalesce? It looks like we had some longer-running expression fuzzers fail with arithmetic error for simplified execution
> 
> ```
> Executing expression 0 : switch(map_key_exists(null,ceiling(coalesce(0.4492889642715454,0.05161726474761963))),transform_values(if("c0",{[382->530] [530->295] 644797815 => {[382->458] [458->383] [383->108] 0.05657687829807401, null, false, -805899563}}),lambda ROW<__a0:INTEGER,__a1:ROW<f0:DOUBLE,f1:INTERVAL DAY TO SECOND,f2:BOOLEAN,f3:INTERVAL DAY TO SECOND>> -> 2011-04-17),array_has_duplicates($internal$canonicalize(array_intersect(find_first({[47->37] {null}},array_sum({[400->715] [715->234] [234->565] 6976801220365335985, [401->317] [317->100] [100->151] 5371919366574234310, [402->701] [701->397] [397->300] 780637959026489352, [403->688] [688->606] [606->257] 6855833019528270910, null, ...3 more}),lambda ROW<__a0:ARRAY<ARRAY<VARCHAR>>> -> is_null("__a0"))))),"c1")
> Common eval succeeded.
> Only simplified path threw exception:
> ...
> Error Code: ARITHMETIC_ERROR
> Reason: integer overflow: 6976801220365335985 + 5371919366574234310
> Retriable: False
> Context: array_sum({[400->715] [715->234] [234->565] 6976801220365335985, [401->317] [317->100] [100->151] 5371919366574234310, [402->701] [701->397] [397->300] 780637959026489352, [403->688] [688->606] [606->257] 6855833019528270910, null, ...3 more}:ARRAY<BIGINT>)
> Function: checkedPlus
> ```

Thanks @peterenescu, I ran the expression fuzzer locally with only `coalesce` for 2hrs+ and do not see any errors. From the logs, `integer overflow` error appears to be originating from `array_sum` function call in the expression and not from `coalesce`. Could you please share the fuzzer `seed` with which this error was seen so I can try to repro on my end and debug further? The `seed` would also help confirm if the error is resulting from this rewrite.

peterenescu:
@pramodsatya Seed is 3539781210. If that doesn't work, you should list all of the functions in the failure case above using the `--only` flag. Eventually it'll compound the same expression (you can ignore `$internal$...`)

mbasmanova:
@pramodsatya Would you confirm if you are able to reproduce Fuzzer failure @peterenescu reported?

peterenescu:
@pramodsatya You can try with flags
```
--duration_sec=1800 --enable_variadic_signatures --lazy_vector_generation_ratio=0.2 --common_dictionary_wraps_generation_ratio=0.3 --velox_fuzzer_enable_complex_types --velox_fuzzer_enable_column_reuse --velox_fuzzer_enable_expression_reuse --include_fb_functions=false --retry_with_try --enable_dereference --seed 3539781210
```
You'll have to confirm the commit ids match as this one (don't rebase locally). If you cannot reproduce, you can drop the seed and let run for a significant time or with the failing functions above using the `only` flag.

pramodsatya:
@peterenescu, @mbasmanova, I am unable to reproduce the error with the seed and fuzzer options you shared (removed arg `--include_fb_functions=false`). I haven't rebased and am on the same commit that was imported. Also ran the expression fuzzer with just the functions from the failing expression and it didn't help. Now running it for longer than `1800s` and will update if I can get a repro locally.

peterenescu:
[coalesce_repro.zip](https://github.com/user-attachments/files/23509857/coalesce_repro.zip)
Thanks for attempting, let us know your results. You may also want to try using above repro files from the failed seed: [wiki](https://facebookincubator.github.io/velox/develop/testing/expression-fuzzer.html#accurate-on-disk-reproduction)

pramodsatya:
> [coalesce_repro.zip](https://github.com/user-attachments/files/23509857/coalesce_repro.zip) Thanks for attempting, let us know your results. You may also want to try using above repro files from the failed seed: [wiki](https://facebookincubator.github.io/velox/develop/testing/expression-fuzzer.html#accurate-on-disk-reproduction)

Thanks a lot @peterenescu, I am now able to repro the error locally with this directory as repro path and triaging the error.

pramodsatya:
@peterenescu, I see the same error on `main` branch without this commit as well, the error results from the integer values in input array to `array_sum` causing an overflow when added. One solution could be to add `array_sum` to the list of skip functions in [`ExpressionFuzzerTest.cpp`](https://github.com/facebookincubator/velox/blob/2b51a3cdd081ff6b9a729a86996edc664d7b2c40/velox/expression/fuzzer/ExpressionFuzzerTest.cpp#L127). 
I feel it would be better to add an `InputGenerator` for `array_sum` say `ArraySumInputGenerator`, similar to `MapUnionSumInputGenerator`, that ensures the values in the integer array [do not overflow](https://github.com/facebookincubator/velox/blob/2b51a3cdd081ff6b9a729a86996edc664d7b2c40/velox/functions/prestosql/fuzzer/MapUnionSumInputGenerator.h#L32) when added. This would also require the custom input generators for functions to be wired into the expression fuzzer (this seems to be missing now).

Would it be fine to fix this fuzzer error in a follow-up PR?

peterenescu:
Hi @pramodsatya that sounds good. Please create a Github issue for that. Based on my local testing, it's a fairly rare and isolated to only array_sum as well. Thanks for confirming this occurred prior to your changes.

pramodsatya:
Thanks @peterenescu, opened https://github.com/facebookincubator/velox/issues/15486 to track this error.

PR #15404, opened 2025-11-05
feat: PartitionedOutputNode must distinguish the root of a plan

zoltan:
Currently when we translate a Presto plan to a Velox plan fragment, we do not have a way later on at the Velox layer to detect if a particular PartitionedOutputNode will be read by the coordinator as the final output of a plan or will be read by other workers.

This patch adds a new field to PartitionedOutputNode so we can signal this during plan translation inside Prestissimo.

In my particular case, This is needed to order to support exchange translation decisions in the cuDF exchange (coming in via a later PR), where the last PartitionedOutputNode has a single partition and will be read by the Presto coordinator, but all other PartitionedOutputNodes will be actually read by other workers via special, hardware accelerated Exchange. Without this change we have no way to distinguish these two at the pure Velox DriverAdapter translation level.

### DIFF
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -2587,7 +2587,8 @@ class PartitionedOutputNode : public PlanNode {
       PartitionFunctionSpecPtr partitionFunctionSpec,
       RowTypePtr outputType,
       VectorSerde::Kind serdeKind,
-      PlanNodePtr source);
+      PlanNodePtr source,
+      bool rootFragment = false);
 
   static std::shared_ptr<PartitionedOutputNode> broadcast(
       const PlanNodeId& id,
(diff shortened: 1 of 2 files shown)


velox/core/PlanNode.h
```diff
@@ -2721,6 +2723,10 @@ class PartitionedOutputNode : public PlanNode {
     return outputType_;
   }
 
+  const bool isRootFragment() const {
```
majetideepak, line 2726: no need to use const since the return is a scalar value here.
`bool isRootFragment() const {`
Similar to below
`bool isPartitioned() const {`

majetideepak:
@mbasmanova, @xiaoxmeng, any thoughts on this API change?

majetideepak:
On the Prestissimo side, the OutputNode is also mapped to the Velox PartitionedOutputNode. This field will help capture that information.
https://github.com/prestodb/presto/pull/26535/files

mbasmanova (commented):
@zoltan It sounds like you want to introduce a new plan node: OutputNode. That node is a special case of PartitionedOutputNode: it must be the root node and must have "gather" semantic.

zoltan:
we've opted for a nicer and more fine grained solution (to carry the transport information in the protocol), which is being implemented. that work makes this obsolete.

PR #15406, opened 2025-11-05
fix: Back out "perf: Do not request data size if there is only  a single source as in persistent shuffle"

kewang1024:
Summary:
Original commit changeset: 4424ea18f7d4

Original Phabricator Diff: D85647156

Reviewed By: amitkdutta, xiaoxmeng

Differential Revision: D86272441


amitkdutta (approved):
Thanks @kewang1024 for debugging a very urgent case and fiding the root cause.

PR #15407, opened 2025-11-05
misc: Use ioCounter for pageLoadTimeNs

rui-mo:
This PR uses `ioCounter` for `pageLoadTimeNs` to collect its min, max, count stats.

### DIFF
--- a/velox/common/io/IoStatistics.h
+++ b/velox/common/io/IoStatistics.h
@@ -43,6 +43,24 @@ struct OperationCounters {
 
 class IoCounter {
  public:
+  IoCounter& operator=(const IoCounter& other) noexcept {
+    if (this != &other) {
+      count_.store(
+          other.count_.load(std::memory_order_relaxed),
+          std::memory_order_relaxed);
+      sum_.store(
+          other.sum_.load(std::memory_order_relaxed),
+          std::memory_order_relaxed);
+      min_.store(
+          other.min_.load(std::memory_order_relaxed),
+          std::memory_order_relaxed);
+      max_.store(
+          other.max_.load(std::memory_order_relaxed),
+          std::memory_order_relaxed);
+    }
+    return *this;
+  }
+
   uint64_t count() const {
     return count_;
   }
--- a/velox/dwio/common/Statistics.h
+++ b/velox/dwio/common/Statistics.h
@@ -22,6 +22,7 @@
 
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/base/RuntimeMetrics.h"
+#include "velox/common/io/IoStatistics.h"
 #include "velox/dwio/common/exception/Exception.h"
 
 namespace facebook::velox::dwio::common {
@@ -540,7 +541,7 @@ struct ColumnReaderStatistics {
   int64_t flattenStringDictionaryValues{0};
 
   // Total time spent in loading pages, in nanoseconds.
-  uint64_t pageLoadTimeNs{0};
+  io::IoCounter pageLoadTimeNs;
 };
 
 struct RuntimeStatistics {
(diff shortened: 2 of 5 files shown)


velox/dwio/common/Reader.h
```diff
@@ -205,7 +205,8 @@ class Reader {
    * @return Row reader
    */
   virtual std::unique_ptr<RowReader> createRowReader(
-      const RowReaderOptions& options = {}) const = 0;
+      const RowReaderOptions& options = {},
+      std::shared_ptr<io::IoStatistics> ioStats = nullptr) const = 0;
```
Yuhta: Ideally format reader should not be aware of such storage level stats.  And other formats than Parquet does not need this.  Is there a way to extract the IO stats from the `BufferedInput` in parquet reader, instead of drilling a hole through format layer?

rui-mo: Thanks for reviewing! We discovered that simply changing the variable type to ioCounter is enough to gather these metrics.

velox/common/io/IoStatistics.h
```diff
@@ -73,6 +73,13 @@ class IoCounter {
     casLoop(max_, other.max_, std::less());
   }
 
+  void clear() {
```
Yuhta: Just do `ioCounter = {}` should be enough, there is no need to add an extra method for this

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -1336,8 +1336,8 @@ class ParquetRowReader::Impl {
   void updateRuntimeStats(dwio::common::RuntimeStatistics& stats) const {
     stats.skippedStrides += skippedStrides_;
     stats.processedStrides += rowGroupIds_.size();
-    stats.columnReaderStatistics.pageLoadTimeNs +=
-        columnReaderStats_.pageLoadTimeNs;
+    stats.columnReaderStatistics.pageLoadTimeNs.increment(
+        columnReaderStats_.pageLoadTimeNs.sum());
```
FelixYBW: should we use merge?

velox/dwio/common/Statistics.h
```diff
@@ -601,10 +619,10 @@ struct RuntimeStatistics {
           "flattenStringDictionaryValues",
           RuntimeMetric(columnReaderStatistics.flattenStringDictionaryValues));
     }
-    if (columnReaderStatistics.pageLoadTimeNs > 0) {
+    if (columnReaderStatistics.pageLoadTimeNs.sum() > 0) {
       result.emplace(
           "pageLoadTimeNs",
-          RuntimeMetric(columnReaderStatistics.pageLoadTimeNs));
+          RuntimeMetric(columnReaderStatistics.pageLoadTimeNs.sum()));
```
FelixYBW: Let's wait until #15408 merged, then update count,min,max

velox/dwio/parquet/reader/ParquetReader.cpp
rui-mo: `updateRuntimeStats` is called by `HiveDataSource::next` and eventually at https://github.com/facebookincubator/velox/blob/04faded0c6c38e5291970c8dc916820af68bff01/velox/exec/TableScan.cpp#L204

If using `merge` here, it's more like a batch-level updating for the min/max stats. Typically the other runtime stats are merged at task level at
https://github.com/facebookincubator/velox/blob/04faded0c6c38e5291970c8dc916820af68bff01/velox/exec/TableScan.cpp#L315


How do you think? Thanks.

velox/common/io/IoStatistics.h
rui-mo: Some tests and benchmarks need to reinitialize a `RuntimeStatistics` object. But after adding `IoCounter` to it, the assignment operator is implicitly deleted due to the use of atomic variables within `IoCounter`. To address this issue, `clear()` methods have been introduced as an alternative solution.

Yuhta: Can you add a copy assignment for `IoCounter` then?  Then there is no need to duplicate the initialization logic in all containing objects.

velox/dwio/parquet/reader/ParquetReader.cpp
FelixYBW: I see. So it's counted in columnreader level or page level. We care about single I/O on page level. So we should use merge.

velox/common/io/IoStatistics.h
rui-mo: Updated, thanks!

velox/dwio/parquet/reader/ParquetReader.cpp
rui-mo: Updated, thanks!

PR #15408, opened 2025-11-05
misc: Propagate full IO stats (count/min/max) to runtime metrics

rui-mo:
This PR enhances the propagation of IO statistics into RuntimeMetric by 
including not just the sum, but also count, min, and max values from the IO 
layer. This affects the IO-related metrics such as:`prefetchBytes`,
`ioWaitWallNanos`, `storageReadBytes`, `localReadBytes`, `ramReadBytes`. 
Previously, these Runtime metrics had a default count of 1, and min/max were 
equal to sum because no detailed IO stats were passed down. After this change, 
count/min/max now reflect real observed values from the IO layer.

The metrics `maxSingleIoWaitWallNanos` and `numStorageRead` are removed
because their functionality is represented by the `ioWaitWallNanos.max ` and
`storageReadBytes.count` respectively.

Users relying on these metrics for historical analysis, monitoring, or 
dashboards should be aware of this impact.


Yuhta (approved):
The change itself makes sense, the only caveat is it changes the semantics of these stats drastically (so that all the stats other than sum are not comparable before & after this change).  This is something worthing noting in the release note.

Yuhta:
@rui-mo The test failures are relevant, can you double check and fix?

peterenescu:
To be specific, there are four velox exec tests failing:
PrintPlanWithStatsTest (innerJoinWithTableScan, partialAggregateWithTableScan, tableWriterWithTableScan)
TableScanTest (statsBasedFilterReorderDisabled)

FelixYBW:
With this change, we may remove `maxSingleIoWaitWallNanos`, `numStorageRead`. They are shown in the `ioWaitWallNanos.max` and `storageReadBytes.count`

rui-mo:
Hi everyone, thanks for the suggestions! I've updated the PR description and resolved the unit test failure. Could you please review it again?

majetideepak:
This PR is causing the values to overflow since `IoCounter` uses `uint64_t` and `RuntimeMetric` uses `int64_t`.
@rui-mo will you be able to help with a fix?

Specifically due to `min_` default in the `IoCounter`.
```
std::atomic<uint64_t> min_{std::numeric_limits<uint64_t>::max()};
```

```
"TableScan.0.ioWaitWallNanos": {
                     "name": "TableScan.0.ioWaitWallNanos",
                     "unit": "NANO",
                     "sum": 0,
                     "count": 0,
                     "max": 0,
                     "min": -1000
                 }
```
See https://github.com/prestodb/presto/issues/26588#issuecomment-3530342769

rui-mo:
@majetideepak Thanks for pointing this out. I opened a PR that changes RuntimeMetrics counters from signed to unsigned (int64_t -> uint64_t) https://github.com/facebookincubator/velox/pull/15536. Let me know your thoughts, thanks.

PR #15410, opened 2025-11-05
feat(cudf): Support CudfVector childrenSize

jinchengchenghh:
Initialize the children vector with nullptr, then function childrenSize can get correct value.


jinchengchenghh:
Could you help review this PR? Thanks! @devavret @karthikeyann I need to wrap CudfVector to Gluten class VeloxColumnarBatch, so I need this function https://github.com/apache/incubator-gluten/blob/main/cpp/velox/memory/VeloxColumnarBatch.h#L30

devavret (approved):
Seems like an ok hack. since childrenSize() isn't virtual

Yuhta (approved):

PR #15424, opened 2025-11-06
feat: Add sampling rate to HiveTableHandle

mbasmanova:
Summary: Axiom optimizer will be able to parse the filter on top of scan to extract not only comparisons of subfields with constants, but also sampling rate (rand() < k). We need a place to store that in the HiveTableHandle.

Differential Revision: D86423635

### DIFF
--- a/velox/connectors/hive/TableHandle.cpp
+++ b/velox/connectors/hive/TableHandle.cpp
@@ -112,15 +112,20 @@ HiveTableHandle::HiveTableHandle(
     const core::TypedExprPtr& remainingFilter,
     const RowTypePtr& dataColumns,
     const std::unordered_map<std::string, std::string>& tableParameters,
-    std::vector<HiveColumnHandlePtr> filterColumnHandles)
+    std::vector<HiveColumnHandlePtr> filterColumnHandles,
+    double sampleRate)
     : ConnectorTableHandle(std::move(connectorId)),
       tableName_(tableName),
       filterPushdownEnabled_(filterPushdownEnabled),
       subfieldFilters_(std::move(subfieldFilters)),
       remainingFilter_(remainingFilter),
+      sampleRate_{sampleRate},
       dataColumns_(dataColumns),
       tableParameters_(tableParameters),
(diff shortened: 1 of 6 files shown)


velox/connectors/hive/TableHandle.h
```diff
@@ -140,12 +140,15 @@ using HiveColumnHandleMap =
...
  public:
+  /// @param sampleRate A non-negative sampling rate. 0.1 means 10% sampling.
+  /// >= 1.0 means no sampling. Default is no sampling.
   HiveTableHandle(
       std::string connectorId,
       const std::string& tableName,
       bool filterPushdownEnabled,
       common::SubfieldFilters subfieldFilters,
       const core::TypedExprPtr& remainingFilter,
+      double sampleRate = 1.0,
```
Yuhta: Can we add it to end of parameters list to avoid breaking?

velox/connectors/hive/TableHandle.cpp
```diff
@@ -118,9 +119,12 @@ HiveTableHandle::HiveTableHandle(
       filterPushdownEnabled_(filterPushdownEnabled),
       subfieldFilters_(std::move(subfieldFilters)),
       remainingFilter_(remainingFilter),
+      sampleRate_{sampleRate > 1.0 ? 1.0 : sampleRate},
```
Yuhta: Should we even allow larger than 1.0?  It should be (0, 1] I think

velox/connectors/hive/TableHandle.h
mbasmanova: Sure. Changed.

mbasmanova:
@Yuhta Thank you for review. Updated the PR to address comments. Would you take another look?

Yuhta:
@mbasmanova Can we just throw exception for >1 like we did for non-positive values?

mbasmanova:
> @mbasmanova Can we just throw exception for >1 like we did for non-positive values?

Sure. Updated.

PR #15427, opened 2025-11-06
fix(ci): Remove temporary GCC14 install

czentgr:


assignUser:
We set `CC` and `CXX` in the docker image and enabling gcctoolset14 doesn't override that.
```
bash-5.1# echo $CC
/opt/rh/gcc-toolset-12/root/bin/gcc
bash-5.1# source /opt/rh/gcc-toolset-14/enable
bash-5.1# echo $CC
/opt/rh/gcc-toolset-12/root/bin/gcc
```
So setting those to gcc 14 should fix this!

czentgr:
> We set `CC` and `CXX` in the docker image and enabling gcctoolset14 doesn't override that.
> 
> ```
> bash-5.1# echo $CC
> /opt/rh/gcc-toolset-12/root/bin/gcc
> bash-5.1# source /opt/rh/gcc-toolset-14/enable
> bash-5.1# echo $CC
> /opt/rh/gcc-toolset-12/root/bin/gcc
> ```
> 
> So setting those to gcc 14 should fix this!

Thanks! Derp, of course it had to be set somewhere.
I found it https://github.com/facebookincubator/velox/blob/1ad164070c13200bce272ff584da998c562604ca/scripts/docker/centos-multi.dockerfile#L89

czentgr:
@majetideepak Please review when you get a chance.

karthikeyann (approved):
Build with cudf works. Tested locally.

Yuhta (approved):

PR #15430, opened 2025-11-06
feat: Report number of lookups to the AsyncDataCache

minhancao:
Add runtime stats for number of lookups to the AsyncDataCache.

This is to track the total number of lookups to the memory cache. Currently, the existing metrics (memory_cache_num_hits + memory_cache_num_new) would not accurately track this due to memory_cache_num_hits only being incremented after the first entry access.

https://github.com/facebookincubator/velox/blob/main/velox/docs/monitoring/metrics.rst?plain=1#L335 https://github.com/facebookincubator/velox/blob/main/velox/common/caching/AsyncDataCache.cpp#L192

### DIFF
--- a/velox/common/base/Counters.cpp
+++ b/velox/common/base/Counters.cpp
@@ -162,6 +162,9 @@ void registerVeloxMetrics() {
   DEFINE_METRIC(
       kMetricMemoryCacheSumEvictScore, facebook::velox::StatType::SUM);
 
+  // Number of lookups since last counter retrieval.
+  DEFINE_METRIC(kMetricMemoryCacheNumLookups, facebook::velox::StatType::SUM);
+
   // Number of hits (saved IO) since last counter retrieval. The first hit to a
   // prefetched entry does not count.
   DEFINE_METRIC(kMetricMemoryCacheNumHits, facebook::velox::StatType::SUM);
--- a/velox/common/base/Counters.h
+++ b/velox/common/base/Counters.h
@@ -220,6 +220,9 @@ constexpr std::string_view kMetricMemoryCacheTotalPrefetchBytes{
 constexpr std::string_view kMetricMemoryCacheSumEvictScore{
     "velox.memory_cache_sum_evict_score"};
 
+constexpr std::string_view kMetricMemoryCacheNumLookups{
+    "velox.memory_cache_num_lookups"};
+
 constexpr std::string_view kMetricMemoryCacheNumHits{
     "velox.memory_cache_num_hits"};
 
--- a/velox/common/base/PeriodicStatsReporter.cpp
+++ b/velox/common/base/PeriodicStatsReporter.cpp
@@ -162,6 +162,7 @@ void PeriodicStatsReporter::reportCacheStats() {
   // Memory cache cumulative stats.
   const auto deltaCacheStats = cacheStats - lastCacheStats_;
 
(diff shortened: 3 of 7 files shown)


zacw7:
curious why is this needed?

minhancao:
@zacw7 This is needed for the tracking and calculation of the hit ratio % of the AsyncDataCache.

Hit ratio = number of hits / number of lookups to the AsyncDataCache

zacw7:
> @zacw7 This is needed for the tracking and calculation of the hit ratio % of the AsyncDataCache.
> 
> Hit ratio = number of hits / number of lookups to the AsyncDataCache

memory cache hit rate can be calculated by: memory_cache_num_hits / (memory_cache_num_hits + memory_cache_num_new) 
memory + ssd cache hit rate: (memory_cache_num_hits + ssd_cache_read_entries) / (memory_cache_num_hits + memory_cache_num_new)

minhancao:
@zacw7 Thanks for the equations! I have tried those equations before and they indeed would work for calculating the hit rate. 

However, I would also like to track the total number of lookups to the memory cache. (memory_cache_num_hits + memory_cache_num_new) would not accurately track this due to memory_cache_num_hits only being incremented after the first entry access. 

https://github.com/facebookincubator/velox/blob/main/velox/docs/monitoring/metrics.rst?plain=1#L335
https://github.com/facebookincubator/velox/blob/main/velox/common/caching/AsyncDataCache.cpp#L192

zacw7:
> @zacw7 Thanks for the equations! I have tried those equations before and they indeed would work for calculating the hit rate.
> 
> However, I would also like to track the total number of lookups to the memory cache. (memory_cache_num_hits + memory_cache_num_new) would not accurately track this due to memory_cache_num_hits only being incremented after the first entry access.
> 
> https://github.com/facebookincubator/velox/blob/main/velox/docs/monitoring/metrics.rst?plain=1#L335 https://github.com/facebookincubator/velox/blob/main/velox/common/caching/AsyncDataCache.cpp#L192

makes sense. thanks for the clarification.

zacw7 (approved):
LGTM. Thanks for adding this.

majetideepak (changes requested):
@minhancao Does the `numlookups` vary a lot compared to `memory_cache_num_hits + memory_cache_num_new` in practice? 
 If there is a lot of variation, then we can probably add a specific metric for first entry access or update num_hits in that path, since it is a hit in a way?
We should avoid adding new metrics that correlate with existing metrics.

minhancao:
@majetideepak Yes, there is a big constant difference between `numLookups` compared to `memory_cache_num_hits + memory_cache_num_new`.

`numLookups`:
image

`memory_cache_num_hits + memory_cache_num_new`:
image


I am for incrementing numHits for first entry access,  not sure why we don't do so since it would be considered a hit because you're accessing that entry in the cache...

czentgr:
@minhancao Please rebase and address review comments and fix conflicts.

minhancao:
@majetideepak Your review comments got lost when I rebased with main branch, please review this PR when you get the chance, thanks!

majetideepak (approved):

majetideepak:
@xiaoxmeng, any concerns with this?

majetideepak:
@zacw7 can you take another look as well? This is now straightforward.

PR #15433, opened 2025-11-06
feat: Add UNKNOWN key, value support for map_zip_with

natashasehgal:
Summary: Add UNKNOWN key, value support for map_zip_with

Differential Revision: D86463197


amitkdutta (approved):
Looks good. Thanks @natashasehgal

PR #15434, opened 2025-11-07
fix: Fix hash probe caused stuck driver issue

tanjialiang:
Summary:
This is a forward fix on the previous revert https://github.com/facebookincubator/velox/pull/15428

The fix makes sure that during spilling, we don't yield in the getOutputInternal() as it is unnecessary.

Differential Revision: D86478196


xiaoxmeng (approved):

PR #15437, opened 2025-11-07
Fix int128 serde

Enochack:
In our use cases, we need to serialize/deserialize 128 bit decimals which don't fit in 64 bits. But the current serde approach just treat int128 as int64 which doesn't work when there are more than 64 significant bits. Instead, int128 is treated as binary in this PR, which works for both `folly::dynamic` and JSON.

### DIFF
--- a/velox/type/Variant.cpp
+++ b/velox/type/Variant.cpp
@@ -682,7 +682,9 @@ folly::dynamic Variant::serialize() const {
       break;
     }
     case TypeKind::HUGEINT: {
-      objValue = value<TypeKind::HUGEINT>();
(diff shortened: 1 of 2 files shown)


Enochack:
> Please use human readable string for decimals JSON serialization. Doing base64 encoding adds unnecessarily complexity.

Sorry that maybe my description is misleading. The point is essentially int128 serde. The underlying int128 doesn't necessarily represent a decimal. For example, it could also a 128-bit integer. But we can't get what it represents here. The original approach doesn't work with int128 which has more than 64 significant bits, which is the PR to fix. Does this make sense?

PR #15035, opened 2025-10-03
feat: Support iceberg partition transform

PingLiuPing:
This is continuous of previous PR https://github.com/facebookincubator/velox/pull/14723.
Implements comprehensive partition transform functionality for
Iceberg tables, enabling data to be partitioned using various transform
functions including identity, bucket, truncate, and temporal transforms
(year, month, day, hour).

PR #15440, opened 2025-11-07
feat: Add support for evaluating Iceberg partition transforms

PingLiuPing:
Introduce infrastructure for evaluating Iceberg partition transforms.
- TransformExprBuilder converts Iceberg partition specifications into Velox expressions
- TransformEvaluator evaluates multiple transform expressions in a single pass using compiled ExprSet.

### DIFF
--- a/velox/connectors/hive/iceberg/PartitionSpec.h
+++ b/velox/connectors/hive/iceberg/PartitionSpec.h
@@ -20,6 +20,25 @@
 
 namespace facebook::velox::connector::hive::iceberg {
 
+inline constexpr char const* kIcebergFunctionPrefix{"iceberg_"};
+
+inline const std::string kBucketFunction =
+    std::string(kIcebergFunctionPrefix) + "bucket";
+inline const std::string kTruncateFunction =
+    std::string(kIcebergFunctionPrefix) + "truncate";
+inline const std::string kYearFunction =
+    std::string(kIcebergFunctionPrefix) + "years";
+inline const std::string kMonthFunction =
+    std::string(kIcebergFunctionPrefix) + "months";
+inline const std::string kDayFunction =
+    std::string(kIcebergFunctionPrefix) + "days";
+inline const std::string kHourFunction =
+    std::string(kIcebergFunctionPrefix) + "hours";
+
+/// Registers Iceberg partition transform functions with prefix
+/// kIcebergFunctionPrefix.
+void registerIcebergFunctions();
+
 /// Partition transform types.
 /// Defines how source column values are converted into partition keys.
 /// See https://iceberg.apache.org/spec/#partition-transforms.
@@ -113,6 +132,7 @@ struct IcebergPartitionSpec {
         case TransformType::kTruncate:
           return type;
       }
+      VELOX_UNREACHABLE("Unknown transform type");
     }
   };
 
--- a/velox/connectors/hive/iceberg/TransformEvaluator.cpp
+++ b/velox/connectors/hive/iceberg/TransformEvaluator.cpp
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/iceberg/TransformEvaluator.h"
+
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+TransformEvaluator::TransformEvaluator(
+    const std::vector<core::TypedExprPtr>& expressions,
+    const ConnectorQueryCtx* connectorQueryCtx)
+    : connectorQueryCtx_(connectorQueryCtx) {
+  VELOX_CHECK_NOT_NULL(connectorQueryCtx_);
+  exprSet_ = connectorQueryCtx_->expressionEvaluator()->compile(expressions);
+  VELOX_CHECK_NOT_NULL(exprSet_);
+}
+
+std::vector<VectorPtr> TransformEvaluator::evaluate(
+    const RowVectorPtr& input) const {
+  const auto numRows = input->size();
+  const auto numExpressions = exprSet_->exprs().size();
+
+  std::vector<VectorPtr> results(numExpressions);
+  SelectivityVector rows(numRows);
+
+  // Evaluate all expressions in one pass.
+  connectorQueryCtx_->expressionEvaluator()->evaluate(
+      exprSet_.get(), rows, *input, results);
+
+  return results;
+}
+
+} // namespace facebook::velox::connector::hive::iceberg
--- a/velox/connectors/hive/iceberg/TransformExprBuilder.h
+++ b/velox/connectors/hive/iceberg/TransformExprBuilder.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/connectors/hive/iceberg/PartitionSpec.h"
+#include "velox/expression/Expr.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+/// Converts Iceberg partition specification to Velox expressions.
+class TransformExprBuilder {
+ public:
+  /// Converts partition specification to a list of typed expressions.
+  ///
+  /// @param partitionSpec Iceberg partition specification containing transform
+  /// definitions for each partition field.
+  /// @param partitionChannels Column indices (0-based) in the input RowVector
+  /// that correspond to each partition field. Must have the same size as
+  /// partitionSpec->fields. Provides the positional mapping from partition spec
+  /// fields to input RowVector columns.
+  /// @param inputType The row type of the input data. This is necessary for
+  /// building expressions because the column names in partitionSpec reference
+  /// table schema names, which might not match the column names in inputType
+  /// (e.g., inputType may use generated names like c0, c1, c2). The
+  /// FieldAccessTypedExpr must be built using the actual column names from
+  /// inputType that will be present at runtime. The partitionChannels provide
+  /// the positional mapping to locate the correct columns.
+  /// @return Vector of typed expressions, one for each partition field.
+  static std::vector<core::TypedExprPtr> toExpressions(
+      const IcebergPartitionSpecPtr& partitionSpec,
+      const std::vector<column_index_t>& partitionChannels,
+      const RowTypePtr& inputType);
+};
+
+} // namespace facebook::velox::connector::hive::iceberg
--- a/velox/connectors/hive/iceberg/tests/TransformTest.cpp
+++ b/velox/connectors/hive/iceberg/tests/TransformTest.cpp
@@ -1,0 +1,328 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/common/encode/Base64.h"
+#include "velox/connectors/hive/iceberg/PartitionSpec.h"
+#include "velox/connectors/hive/iceberg/TransformEvaluator.h"
+#include "velox/connectors/hive/iceberg/TransformExprBuilder.h"
+#include "velox/connectors/hive/iceberg/tests/IcebergTestBase.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+namespace {
+
+class TransformTest : public test::IcebergTestBase {
+ protected:
+  void testTransform(
+      const IcebergPartitionSpecPtr& spec,
+      const RowVectorPtr& input,
+      const RowVectorPtr& expected) const {
+    std::vector<column_index_t> partitionChannels;
+    for (const auto& field : spec->fields) {
+      partitionChannels.push_back(input->rowType()->getChildIdx(field.name));
+    }
+    // Build and evaluate transform expressions.
+    auto transformExprs = TransformExprBuilder::toExpressions(
+        spec, partitionChannels, input->rowType());
+    auto transformEvaluator = std::make_unique<TransformEvaluator>(
+        transformExprs, connectorQueryCtx_.get());
+    auto result = transformEvaluator->evaluate(input);
+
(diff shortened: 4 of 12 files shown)


PingLiuPing:
@mbasmanova This PR continues from https://github.com/facebookincubator/velox/pull/15035#pullrequestreview-3427716555.
In next PR, I plan to include `IcebergPartitionIdGenerator` and do some refactor on Hive `PartitionIdGenerator`, I roughly count the lines which should be also around 700 lines. After that, all the preliminary work should be complete, and I’ll submit another PR to integrate them with IcebergDataSink.

mbasmanova:
Ci is red

velox/connectors/hive/iceberg/IcebergPartitionPath.h
```diff
@@ -0,0 +1,51 @@
...
+#pragma once
+
+#include "velox/connectors/hive/HivePartitionUtil.h"
+#include "velox/connectors/hive/iceberg/PartitionSpec.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+/// Converts a partition value to its string representation for use in
+/// partition directory path. The format follows the Iceberg specification
+/// for partition path encoding.
```
mbasmanova: Do you have a link to the spec? Would be nice to add here.

velox/connectors/hive/iceberg/IcebergPartitionPath.h
```diff
@@ -0,0 +1,51 @@
...
+class IcebergPartitionPath : public HivePartitionUtil {
+ public:
+  explicit IcebergPartitionPath(TransformType transformType)
+      : transformType_(transformType) {}
+
+  ~IcebergPartitionPath() override = default;
```
mbasmanova: not needed

velox/connectors/hive/iceberg/IcebergPartitionPath.h
```diff
@@ -0,0 +1,51 @@
...
+
+  using HivePartitionUtil::toPartitionString;
+
+  std::string toPartitionString(int32_t value, const TypePtr& type)
```
mbasmanova: perhaps, document the format of the result

velox/connectors/hive/iceberg/IcebergPartitionPath.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "velox/connectors/hive/HivePartitionUtil.h"
+#include "velox/connectors/hive/iceberg/PartitionSpec.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+/// Converts a partition value to its string representation for use in
+/// partition directory path. The format follows the Iceberg specification
+/// for partition path encoding.
+class IcebergPartitionPath : public HivePartitionUtil {
```
mbasmanova: Would it be possible to extract this class into a separate PR and add a test?

velox/connectors/hive/iceberg/TransformExprBuilder.h
```diff
@@ -0,0 +1,64 @@
...
+  /// Builds an expression tree for one partition transform. Identity transforms
+  /// become FieldAccessTypedExpr, while other transforms (bucket, truncate,
+  /// year, month, day, hour) become CallTypedExpr with appropriate function
+  /// names and parameters.
+  ///
+  /// @param field Partition field containing transform type, source column
+  /// type, and optional parameter (e.g., bucket count, truncate width).
+  /// @param inputFieldName Name of the source column in the input RowVector.
+  /// @return Typed expression representing the transform.
+  static core::TypedExprPtr toExpression(
```
mbasmanova: Do we need to expose this API?

velox/connectors/hive/iceberg/TransformEvaluator.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <mutex>
+
+#include "velox/expression/Expr.h"
+#include "velox/functions/iceberg/Register.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+namespace {
+
+constexpr char const* kIcebergFunctionPrefix{"iceberg_"};
```
mbasmanova: move this next to its only use

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,292 @@
...
+#include "velox/connectors/hive/iceberg/tests/IcebergTestBase.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+namespace {
+
+class TransformTest : public test::IcebergTestBase {
+ protected:
+  template <typename T>
+  VectorPtr createFlatVector(
```
mbasmanova: Why not use VectorMaker or helper methods from VectorTestBase?

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,292 @@
...
+        0, std::vector<IcebergPartitionSpec::Field>{field});
+    auto transformExprs = TransformExprBuilder::toExpressions(
+        spec, std::vector<column_index_t>{0}, asRowType(rowVector->type()));
+    auto transformEvaluator = std::make_unique<TransformEvaluator>(
+        transformExprs, connectorQueryCtx_.get());
+    auto resultVector = transformEvaluator->evaluate(rowVector);
+
+    // Create expected vector and verify results.
+    auto expectedVector = createFlatVector<OUT>(expectedValues);
+    ASSERT_EQ(resultVector[0]->size(), expectedValues.size());
```
mbasmanova: use assertEqualVectors or similar from VectorTestBase.h

velox/connectors/hive/iceberg/IcebergPartitionPath.h
PingLiuPing: Thank you for reviewing.

I will separate this to a new PR.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks. Changed to VectorMaker

velox/connectors/hive/iceberg/IcebergPartitionPath.h
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/TransformExprBuilder.h
PingLiuPing: Thanks. I moved it to anonymous namespace.

velox/connectors/hive/iceberg/TransformEvaluator.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/PartitionSpec.h
```diff
@@ -113,6 +113,7 @@ struct IcebergPartitionSpec {
         case TransformType::kTruncate:
           return type;
       }
+      VELOX_UNREACHABLE("Unknown transform type");
```
PingLiuPing, line 135: @mbasmanova  Have to add this to fix the building break on Linux. Macos is fine without this change.

velox/connectors/hive/iceberg/IcebergPartitionPath.h
PingLiuPing: Thanks for the comment.
I cannot find an explicit document about this. I will revise the comment.
I followed the Iceberg java code to implement this.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+namespace {
+
+class TransformTest : public test::IcebergTestBase {
+ protected:
+  template <typename IN, typename OUT>
+  void testTransform(
+      const IcebergPartitionSpec::Field& field,
+      const std::vector<std::optional<IN>>& inputValues,
+      const std::vector<std::optional<OUT>>& expectedValues,
```
mbasmanova: Can we replace these arguments with RowVectorPtr and use makeFlatVector and similar in the callers?

PingLiuPing: Thanks. Just updated.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+namespace facebook::velox::connector::hive::iceberg {
+
+namespace {
+
+class TransformTest : public test::IcebergTestBase {
+ protected:
+  void testTransform(
+      const IcebergPartitionSpec::Field& field,
+      const RowVectorPtr& inputRowVector,
```
mbasmanova: nit: inputRowVector -> input, expectedRowVector -> expected

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,368 @@
...
+      const RowVectorPtr& expectedRowVector) {
+    // Build and evaluate transform expressions.
+    auto spec = std::make_shared<const IcebergPartitionSpec>(
+        0, std::vector<IcebergPartitionSpec::Field>{field});
+    auto transformExprs = TransformExprBuilder::toExpressions(
+        spec,
+        std::vector<column_index_t>{0},
+        asRowType(inputRowVector->type()));
```
mbasmanova: input->rowType()

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,368 @@
...
+namespace {
+
+class TransformTest : public test::IcebergTestBase {
+ protected:
+  void testTransform(
+      const IcebergPartitionSpec::Field& field,
+      const RowVectorPtr& inputRowVector,
+      const RowVectorPtr& expectedRowVector) {
+    // Build and evaluate transform expressions.
+    auto spec = std::make_shared<const IcebergPartitionSpec>(
```
mbasmanova: Why do we test single field specs only? Why not test multi-field specs?

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,368 @@
...
+      makeRowVector({makeFlatVector<StringView>(
+          {StringView("test data"), StringView("")})}));
+  testTransform(
+      partitionSpec->fields[3],
+      makeRowVector(
+          {partitionSpec->fields[3].name},
+          {makeFlatVector<StringView>(
+              {StringView("\x01\x02\x03", 3), StringView("", 0)},
+              VARBINARY())}),
+      makeRowVector({makeFlatVector<StringView>(
```
mbasmanova: drop StringView or use std::string

```
makeFlatVector<std::string>({"foo", "bar"})
```

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thank you for the comment.
I added those test one by one when I implement the transformation logic one by one, and type by type.
I will refactor it to support test multiple transforms.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/TransformEvaluator.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+TransformEvaluator::TransformEvaluator(
+    const std::vector<core::TypedExprPtr>& expressions,
+    const ConnectorQueryCtx* connectorQueryCtx)
+    : connectorQueryCtx_(connectorQueryCtx) {
+  VELOX_CHECK_NOT_NULL(connectorQueryCtx_);
+
+  // Register Iceberg functions once for expression evaluation.
+  static std::once_flag registerFlag;
+  static constexpr char const* kIcebergFunctionPrefix{"iceberg_"};
```
mbasmanova: iceberg_ prefix is now hard-coded in 2 places: here and in TransformExprBuilder.cpp

Also, It seems better to make these functions available unconditionally, not only after running a CTAS or INSERT INTO query.

I suggest to move the logic for registering these functions to the same place where we register Iceberg connector and add a constant for prefix, which can be reused without hard-coding multiple times.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,362 @@
...
+      partitionSpec,
+      makeRowVector(
+          rowType->names(),
+          {makeFlatVector<int32_t>({1, -1}),
+           makeFlatVector<int64_t>({1L, -1L}),
+           makeFlatVector<std::string>({("test data"), ("")}),
+           makeFlatVector<std::string>({("\x01\x02\x03"), ("")}, VARBINARY()),
+           makeFlatVector<Timestamp>(
+               {Timestamp(0, 0), Timestamp(1609459200, 0)})}),
+      makeRowVector(
```
mbasmanova: Let's not copy-paste expected values for identity transform. Let's use 'input'.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,362 @@
...
+          {
+              makeNullableFlatVector<int32_t>({std::nullopt}),
+              makeNullableFlatVector<std::string>({std::nullopt}),
+              makeNullableFlatVector<std::string>({std::nullopt}, VARBINARY()),
+              makeNullableFlatVector<int32_t>({std::nullopt}, DATE()),
+              makeNullableFlatVector<Timestamp>({std::nullopt}),
+              makeNullableFlatVector<Timestamp>({std::nullopt}),
+              makeNullableFlatVector<Timestamp>({std::nullopt}),
+          }),
+      makeRowVector({
```
mbasmanova, line 108: ditto

velox/connectors/hive/iceberg/tests/TransformTest.cpp
```diff
@@ -0,0 +1,362 @@
...
+          makeFlatVector<int64_t>({100L, -200L, 0L}),
+          makeFlatVector<std::string>({"abcde", "测试dat", "x"}),
+          makeFlatVector<std::string>(
+              {"abc", "\x01\x02\x03", "\x05"}, VARBINARY()),
+      }));
+}
+
+TEST_F(TransformTest, multipleTransforms) {
+  const auto& rowType =
+      ROW({"c_int", "c_date", "c_varchar"}, {INTEGER(), DATE(), VARCHAR()});
```
mbasmanova: no need to use custom column names; use c0, c1, c2... and then drop rowType->name() argument from makeRowVector; ditto all other places

velox/connectors/hive/iceberg/TransformEvaluator.cpp
PingLiuPing: Thank you for the comment.
Iceberg connector does not exist at the moment. Can I register these functions when registering Hive connector?

mbasmanova: > Iceberg connector does not exist at the moment. 

Hmm... but then, how can it be used? Should we create one or at least add a function to register what's needed for Iceberg?

PingLiuPing: Currently it depends on `HiveConnector::createDataSink` to create an IcebergDataSink.

> Should we create one or at least add a function to register what's needed for Iceberg?

How about adding a function first in this PR, and adding IcebergConnector in a separate PR?

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/TransformTest.cpp
PingLiuPing: Thanks, this case tested all transforms on null value. Most of the expected vector type is different with input.

velox/connectors/hive/iceberg/TransformEvaluator.cpp
mbasmanova: > How about adding a function first in this PR, and adding IcebergConnector in a separate PR?

Sounds good.

mbasmanova (approved):
Thanks.

PingLiuPing:
@peterenescu  Could you please help to import this PR? Thank you.

PR #15441, opened 2025-11-07
perf: Add Filter::testStringView

Yuhta:
Summary: When testing filters against in-memory `StringView` vectors, we observe suboptimal performance when using the `testBytes` interface.  This is because `StringView` can compare the inline or prefix part in a few instructions before going into the byte-to-byte comparison loop, but `testBytes` does not use it.  Leverage this by adding a `testStringView` interface.  The gain is most significant for an equality filter with a small string (no longer than 12 bytes).

Differential Revision: D85272787

### DIFF
--- a/velox/type/Filter.h
+++ b/velox/type/Filter.h
@@ -200,6 +200,10 @@ class Filter : public velox::ISerializable {
     VELOX_UNSUPPORTED("{}: testBytes() is not supported.", toString());
   }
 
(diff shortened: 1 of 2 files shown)


kagamiori (approved):
Do we also need to optimize MultiRange filter?

Yuhta:
@kagamiori We definitely could, but the benefit is not as significant, and the chance of multi-range being push down is not high either.

PR #474, opened 2021-10-20
Reuse S3FileSystem instance

majetideepak:
Improve the S3FileSystem support by reusing the S3FileSystem instance. Rename S3FileTest.cpp to S3FileSystemTest.cpp

The following changes are also made.

- Clarify that the S3FileSystem owns the S3Client instance. Update the S3ReadFile API accordingly.

- Add a fileHandle test as it is used by the HiveConnector

- Fix a benign bug and re-factor the S3FileSystemTest

PR #15444, opened 2025-11-07
feat(ci): Add ASAN and UBSAN build and CI job

czentgr:


czentgr:
@kgpai @pedroerp FYI, the job was cancelled after 6h. But looks like we were close to finishing `473/474 Test #474: velox_tool_trace_test`. This job used a `16-core-ubuntu` machine that is also used for the fedora build. Now that it generated a ccache for this job/build I expect this to finish much faster. I probably will rebase and retry.

However, the sanitizer reported lots of hits. Alignment errors, type errors (specifically it complains about the generated parsers), memory leak errors in tests, ...

velox/exec/fuzzer/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ add_library(
 # https://github.com/facebookincubator/velox/issues/15414
 if(VELOX_ENABLE_REMOTE_FUNCTIONS)
   target_sources(velox_fuzzer_util PRIVATE VeloxQueryRunner.cpp)
-  target_link_libraries(velox_fuzzer_util FBThrift::thriftcpp2)
+  target_link_libraries(velox_fuzzer_util velox_local_runner_service_lib)
```
czentgr: This fixes a race condition I've seen popping up here and there. VeloxQueryRunner.cpp relies on a generated header but has no direct dependency. So in many cases the file will have been generated already but I've seen cases where it is not. So make sure that the dependency is defined to have stable builds.
This could be a separate PR if needed.

.github/workflows/scheduled.yml
```diff
@@ -1475,3 +1475,101 @@ jobs:
...
+      faiss_SOURCE: BUNDLED
+    steps:
+      - name: Free Disk Space
+        run: |
+          # Re-used from free-disk-space github action.
+          getAvailableSpace() { echo $(df -a $1 | awk 'NR > 1 {avail+=$4} END {print avail}'); }
+          # Show before
+          echo "Original available disk space: " $(getAvailableSpace)
+          # Remove DotNet.
+          rm -rf /host_usr/share/dotnet || true
```
kgpai: Curious where these are coming from ? Are the adapters images built with this ?

.github/workflows/scheduled.yml
```diff
@@ -1475,3 +1475,101 @@ jobs:
...
+            "-DVELOX_ENABLE_HDFS=ON"
+            "-DVELOX_ENABLE_S3=ON"
+            "-DVELOX_ENABLE_GCS=ON"
+            "-DVELOX_ENABLE_ABFS=ON"
+            "-DVELOX_MONO_LIBRARY=ON"
+            "-DVELOX_BUILD_SHARED=ON"
+            "-DVELOX_ENABLE_REMOTE_FUNCTIONS=ON"
+          )
+          clang++ --version
+          dnf install -y compiler-rt
```
kgpai: nit: lets add this to the centos image

.github/workflows/scheduled.yml
czentgr: These come from the github host. They don't come from the adapters image.

czentgr:
@kgpai Let me rebase. Then it will run with Clang 21 because Centos9 updated it recently.
We don't install any particular version but use the OS version that comes with it by default.

CMakeLists.txt
```diff
@@ -743,7 +747,26 @@ if("${TREAT_WARNINGS_AS_ERRORS}")
...
+  AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 20
+)
+  set(
+    CMAKE_CXX_FLAGS
+    "${CMAKE_CXX_FLAGS} -O1 -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer -DVELOX_ENABLE_ASAN_UBSAN_SANITIZERS"
+  )
+  # Skip alignment errors and member access check for now.
+  # https://github.com/facebookincubator/velox/issues/15811
+  # https://github.com/facebookincubator/velox/issues/15810
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address -fsanitize=undefined -fno-sanitize=alignment -fno-sanitize=vptr")
```
kgpai: Lets also add `-fno-sanitize=alignment` to turn off the noisy alignment issues.

.github/workflows/linux-build-base.yml
```diff
@@ -375,3 +375,110 @@ jobs:
...
+        run: ccache -s
+
+      - uses: apache/infrastructure-actions/stash/save@3354c1565d4b0e335b78a76aedd82153a9e144d4
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-asan-ubsan-adapters
+
+      - name: Run Tests
+        env:
+          LIBHDFS3_CONF: ${{ github.workspace }}/scripts/ci/hdfs-client.xml
```
kgpai, line 885: Lets also add LSAN_OPTIONS=detect_leaks=0 to turn off leak sanitizer (Its pretty noisy right now, and I think it fails more often than not due to our arena allocation/memory pooling).

kgpai (commented):
Looks good , Thanks for this !

.github/workflows/scheduled.yml
czentgr: There is a change to the setup-centos script as well. Once we get a new image I will remove this. But I need this now because the image hasn't been updated yet. Or we do this as a separate change and then get the new image. Either way, it is another PR.

CMakeLists.txt
czentgr: The above change already has `-fno-sanitize=alignment` and `-fno-sanitize=vptr`. You would need to scroll to the right (and I need to run the pre-commit to fix it).

kgpai:
@czentgr Increased the size of the machine for the resolve dependencies change , so we should be unblocked on landing this .

kgpai:
@czentgr  Rebase to latest main resolve_dependency should be green.

czentgr:
@kgpai Please take a look. The Ci is now clean. The Spark aggregate fuzzer issue is a known issue. We come in now at 119 mins (with cache). Adapters being the slowest still.

.github/workflows/linux-build-base.yml
```diff
@@ -383,3 +383,107 @@ jobs:
...
+            "-DVELOX_ENABLE_PARQUET=ON"
+            "-DVELOX_ENABLE_HDFS=ON"
+            "-DVELOX_ENABLE_S3=ON"
+            "-DVELOX_ENABLE_GCS=ON"
+            "-DVELOX_ENABLE_ABFS=ON"
+            "-DVELOX_BUILD_SHARED=ON"
+            "-DVELOX_ENABLE_REMOTE_FUNCTIONS=OFF"
+          )
+          clang++ --version
+          dnf install -y compiler-rt
```
kgpai: Can you add this in the setup-centos9 script ? I am trying to make the builds much more stable, so can we move the conda install etc to the scripts and rebuild docker images?

.github/workflows/linux-build-base.yml
```diff
@@ -383,3 +383,107 @@ jobs:
...
+          LSAN_OPTIONS: detect_leaks=0
+        working-directory: _build/release
+        run: |
+          # Can be removed after images are rebuild
+          if [ -f "/opt/miniforge/etc/profile.d/conda.sh" ]; then
+            source "/opt/miniforge/etc/profile.d/conda.sh"
+            conda activate adapters
+          fi
+          # Needed for HADOOP 3.3.6 minicluster. Can remove after updating to 3.4.2.
+          wget https://repo1.maven.org/maven2/org/mockito/mockito-core/2.23.4/mockito-core-2.23.4.jar -O /usr/local/hadoop/share/hadoop/mapreduce/mockito-core-2.23.4.jar
```
kgpai: Lets add all these to the docker images and rebuild.

czentgr:
@kgpai I migrated the job to use the new steps for build/test. Repeated run is now 60mins for build and test. 
Please take another look. I think it is ready to go after review.

kgpai:
Thanks for putting this together — having ASAN/UBSAN coverage in CI is a great addition, and the integration with the existing build/test/status pattern is clean. I had a few questions and small suggestions on the changes; nothing blocking, just hoping to better understand the intent in a couple of places.

**1. `linux-build-base.yml` — `VELOX_BUILD_SHARED=ON` vs. the comment above it**

The comment says `VELOX_BUILD_SHARED is turned off due producing a SEGV if the shared library is used. Revisit once PR 16019 lands that removes thrift.` — but the flag list right below has `-DVELOX_BUILD_SHARED=ON`. Could you clarify which one is right? If the SEGV is still a real risk we probably want `OFF` here; if PR #16019 has landed and the comment is stale, would you mind updating the comment so the next reader doesn't have to wonder?

**2. `CMakeLists.txt` — `VERSION_GREATER_EQUAL 15` → `VERSION_EQUAL 15` for `-latomic`**

This change drops `-latomic` linking for Clang 16 and newer. I wasn't sure if that was intentional or a side-effect of the sanitizer work — could you share the motivation? If the goal was just to keep Clang 20 from picking up `-latomic`, a `VERSION_LESS 16` (or an explicit version range) would preserve the behavior for Clang 16/17/18/19 builds that may still need it.

**3. `SkewedPartitionBalancerTest.cpp` — disabled `serializedRebalanceExecution`**

The comment notes the test returns `2 vs 1 expected` or times out under `-O1`. That feels like it could be a real signal (optimizer interaction or a timing-sensitive test) rather than a sanitizer artifact. Would you mind opening a tracking issue and referencing it next to the `#ifndef`, the way the alignment/vptr suppressions reference #15810 / #15811? That way it doesn't get forgotten.

**4. `SignatureParser.yy` — the `yyFlexLexer` dummy block**

A few small things on this one:
- The dummies throw raw `std::runtime_error` — would `VELOX_FAIL` be a better fit for codebase consistency?
- The `\"- dummy\"` suffix on the `yylex()` message stood out a bit; happy to leave it but figured I'd flag it.
- More substantively: since the typeinfo is only missing under sanitizers, this looks more like a linker visibility / link-order issue than a sanitizer issue per se. Did you consider an explicit-instantiation TU or an \`extern template\` outside the \`#ifdef\`? That would localize the workaround and not depend on the sanitizer macro. Totally fine to defer if you've already explored that and ruled it out — just curious what you found.

**5. `ExchangeClientTest.cpp` — `executor_->join()` in `TearDown`**

This looks like a real lifetime fix that ASAN surfaced. Could you add a sentence in the commit message about what ASAN reported here? And do you happen to know whether any other test fixtures using `executor_` may need the same change, or is this one specific to the exchange-client lifecycle? Not asking you to audit them all — just want to make sure we're not patching one symptom of a broader pattern.

Thanks again — happy to chat on any of these if it's easier than going back and forth in comments.

czentgr:
@kgpai I've pushed some changes based on your review comments

1. linux-build-base.yml — VELOX_BUILD_SHARED=ON vs. the comment above it

The comment was outdated. I fixed it.

2. CMakeLists.txt — VERSION_GREATER_EQUAL 15 → VERSION_EQUAL 15 for -latomic

The implementation for the atomic symbols can come from the gcc atomic library (needing to add -latomic) or the `compiler-rt` library. The latter was added for asan/ubsan which is why we don't need to link atomic anymore.
To make things a bit more clear I added a function to check if compiler-rt is available instead of the version check.

3. SkewedPartitionBalancerTest.cpp — disabled serializedRebalanceExecution

This test is only run if enabled in DEBUG mode. But we build in RELEASE mode with `-O1`. As a result the test is not run.
I will check if the issue still occurs.
Edit: The issue does not occur anymore and no changes are made to the test.


4. SignatureParser.yy — the yyFlexLexer dummy block

I made it its own translation unit and added it to the common library because it appears multiple times across the code base.

5. ExchangeClientTest.cpp — executor_->join() in TearDown

It seems that this is a common pattern to ensure we don't have left over threads. I see others use a reset() function which is not implemented for this executor. I suspect adding the asan/ubsan symbols changed the timing a bit and so we need to ensure we are cleaning up. This is a test only problem.

kgpai:
Spotted an inconsistency between the CMake gate and the CI step:

The gate in `CMakeLists.txt` requires `CMAKE_BUILD_TYPE STREQUAL "Release"`:

```cmake
if(VELOX_ENABLE_ASAN_UBSAN_SANITIZERS
   AND CMAKE_BUILD_TYPE STREQUAL "Release"
   AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang"
   AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 20)
```

But the workflow step `Make Asan/Ubsan Build (release plus -O1)` invokes `make debug`, which resolves to `BUILD_TYPE=Debug` in the Velox Makefile. So `CMAKE_BUILD_TYPE` ends up as `Debug` and the entire sanitizer block is skipped — no `-fsanitize=address`, no `-fsanitize=undefined`. The job ends up producing a vanilla Debug build under the ASAN/UBSAN job name.

Question: does the gate actually need to be Release-only? A Debug build with sanitizers would give better stack traces and line numbers in sanitizer reports, which is usually what you want when debugging the issues these tools flag.

czentgr:
@kgpai Thanks for the review. You were correct and it was a good catch. I missed to change this when switching from release back to debug. I addressed the issue. It should be good to go now.
We can see the sanitizer flags being set
```
FINAL CMAKE_CXX_FLAGS= -mavx2 -mfma -mavx -mf16c -mlzcnt -mbmi2 -D USE_VELOX_COMMON_BASE -D HAS_UNCAUGHT_EXCEPTIONS -Wall -Wextra -Wno-unused        -Wno-unused-parameter        -Wno-sign-compare        -Wno-ignored-qualifiers        -Wno-range-loop-analysis          -Wno-mismatched-tags          -Wno-nullability-completeness -Werror -O1 -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer -DVELOX_ENABLE_ASAN_UBSAN_SANITIZERS
```

kgpai (approved):

kgpai:
Please fix the pre-commit issue, looks good otherwise.

czentgr:
@kgpai Yes, the pre-commit issue in unrelated. I've pinged you about it.

czentgr:
@kgpai CI is green now after rebasing again. Please take another look.

kgpai:
@czentgr Should land soon.

PR #15445, opened 2025-11-08
fix: Fix incorrect scalar optimization in `gatherCopy`

duanmeng:
The optimized path for scalar targets in `gatherCopy` incorrectly assumed all
source vectors were flat, causing failures with encoded vectors such as `DictionaryVector`.
This PR adds another check to ensure the fast path is now only taken when the target is
scalar and all the sources are flattened.

### DIFF
--- a/velox/exec/OperatorUtils.cpp
+++ b/velox/exec/OperatorUtils.cpp
@@ -103,7 +103,11 @@ void gatherCopy(
     const std::vector<const RowVector*>& sources,
     const std::vector<vector_size_t>& sourceIndices,
     column_index_t sourceChannel) {
-  if (target->isScalar()) {
+  const bool flattenSources =
+      std::all_of(sources.begin(), sources.end(), [](const auto& source) {
+        return source->isFlatEncoding();
+      });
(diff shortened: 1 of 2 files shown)


velox/examples/metrics.cpp
```diff
@@ -0,0 +1,18 @@
+/*
```
xiaoxmeng: Why we need this?

velox/exec/OperatorUtils.cpp
```diff
@@ -102,8 +102,9 @@ void gatherCopy(
     vector_size_t count,
     const std::vector<const RowVector*>& sources,
     const std::vector<vector_size_t>& sourceIndices,
-    column_index_t sourceChannel) {
-  if (target->isScalar()) {
+    column_index_t sourceChannel,
```
xiaoxmeng: Can we flatten based on the source vector encoding? Thanks!

xiaoxmeng (commented):
@duanmeng thanks!

xiaoxmeng (approved):
@duanmeng thanks!

PR #15453, opened 2025-11-10
refactor: Compare sort flags count against column count instead of row count

duanmeng:
Corrects the sort validation logic by comparing the number of sort flags to the column count
instead of the row count.


tanjialiang (approved):
Thanks for the fix

velox/exec/SortBuffer.cpp
```diff
@@ -310,7 +310,7 @@ void SortBuffer::ensureSortFits() {
   }
 
   // The memory for std::vector sorted rows and prefix sort required buffer.
-  uint64_t sortBufferToReserve =
+  const uint64_t sortBufferToReserve =
```
tanjialiang: const auto?

PR #15455, opened 2025-11-10
feat: Add support for SpatialJoinFuzzer

kgpai:
Differential Revision: D86709638

This change adds support for SpatialJoinFuzzer. Spatial Join is sufficiently different from default Join fuzzing that it seemed better to keep it separate rather than shoe horn into our default join fuzzer. 

SpatialJoin has more limited capabilities than regular joins. It does not support:

* Multiple partition strategies
* Grouped execution
* Spilling
* Flipped join order (build/probe swap)

At the moment the fuzzer only generates simple ``Values -> SpatialJoin`` plans.

Limitations
===========

Apart from this the current iteration of the Spatial fuzzer has following limitations that I will try and address in later iterations.

1. Limited support for radius-based spatial joins (used only in ST_Distance predicate case)
2. No support for complex filter expressions beyond spatial predicates
3. No support for TableScan
4. No support for output column reordering/selection (uses all columns)

Another thing to note is that currently it doesnt support either Presto or DuckDB as a source of truth. Support for adding Presto as a reference query runner will come in future iterations.

### DIFF
--- a/velox/docs/develop/testing.rst
+++ b/velox/docs/develop/testing.rst
@@ -11,5 +11,6 @@ Testing Tools
     testing/join-fuzzer
     testing/memory-arbitration-fuzzer
     testing/row-number-fuzzer
+    testing/spatial-join-fuzzer
     testing/writer-fuzzer
     testing/spark-query-runner.rst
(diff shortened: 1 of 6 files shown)


aditi-pandit:
@jkhaliqi CCing Jacob as he has been testing Geospatial on IBM side.

jagill (approved):
Please fix lints.

We can iterate with more complicated geometry types in later PRs.

PR #10301, opened 2024-06-25
Improve the performance of hash Spark functions for scalar types through auto-vectorization

zhli1142015:
Improve the performance of hash Spark functions for scalar types through auto-vectorization.
We can observe a measurable speed-up with this change: 
TINYINT: ~4X, SMALLINT: ~4X, BIGINT: ~2X, REAL: ~4X, DOUBLE:~2X, DECIMAL(18, 6): ~2X, DECIMAL(38, 6): ~1.2X, TIMESTAMP: ~1.8X, VARCHAR: ~1.3X.
Benchmark result before this optimization:
```
============================================================================
[...]hmarks/ExpressionBenchmarkBuilder.cpp     relative  time/iter   iters/s
============================================================================
hash_ARRAY<MAP<INTEGER,VARCHAR>>##hash                      9.63ms    103.88
hash_ARRAY<MAP<INTEGER,VARCHAR>>##xxhash64                  9.38ms    106.66
----------------------------------------------------------------------------
hash_BIGINT##hash                                         352.84us     2.83K
hash_BIGINT##xxhash64                                     346.62us     2.88K
----------------------------------------------------------------------------
hash_DECIMAL(18, 6)##hash                                 353.29us     2.83K
hash_DECIMAL(18, 6)##xxhash64                             346.79us     2.88K
----------------------------------------------------------------------------
hash_DECIMAL(38, 6)##hash                                 787.34us     1.27K
hash_DECIMAL(38, 6)##xxhash64                             742.86us     1.35K
----------------------------------------------------------------------------
hash_DOUBLE##hash                                         370.08us     2.70K
hash_DOUBLE##xxhash64                                     371.87us     2.69K
----------------------------------------------------------------------------
hash_REAL##hash                                           349.01us     2.87K
hash_REAL##xxhash64                                       360.70us     2.77K
----------------------------------------------------------------------------
hash_ROW<f_map:MAP<INTEGER,VARCHAR>,f_array:ARR            17.70ms     56.50
hash_ROW<f_map:MAP<INTEGER,VARCHAR>,f_array:ARR            17.93ms     55.77
----------------------------------------------------------------------------
hash_SMALLINT##hash                                       331.26us     3.02K
hash_SMALLINT##xxhash64                                   340.99us     2.93K
----------------------------------------------------------------------------
hash_TIMESTAMP##hash                                      493.88us     2.02K
hash_TIMESTAMP##xxhash64                                  486.54us     2.06K
----------------------------------------------------------------------------
[...]

PR #15458, opened 2025-11-11
perf: Optimize basic numeric upcast

rui-mo:
When the row size is large (e.g., around 300,000,000), casting from a narrower 
integer type to a wider one—such as cast(integer as bigint)—can become time-
consuming.

This PR optimizes the numeric upcast by performing the cast directly on the 
raw values within loops, and drops the try-catch used for potential error handing. 
Since upcasts guarantee that the source value fits within the target type, overflow
handling is unnecessary in this case.

The performance gains are likely attributed to:
1) Eliminating try-catch blocks when error handling is unnecessary.
2) Improved auto-vectorization and lower function call overhead after replacing
`valueAt` and `set` with direct access.
3) Avoiding overflow checks.

Optimized conversions include: 
```
CAST(tinyint AS smallint)
CAST(tinyint AS integer)
CAST(tinyint AS bigint)
CAST(tinyint AS real)
CAST(tinyint AS double)
CAST(tinyint AS hugeint)

CAST(smallint AS integer)
CAST(smallint AS bigint)
CAST(smallint AS real)
CAST(smallint AS double)
CAST(smallint AS hugeint)

CAST(integer AS bigint)
CAST(integer AS real)
CAST(integer AS double)
CAST(integer AS hugeint)

CAST(bigint AS real)
CAST(bigint AS double)
CAST(bigint AS hugeint)

CAST(hugeint AS real)
CAST(hugeint AS double)

CAST(real AS double)
```

Before:

```
[...]hmarks/ExpressionBenchmarkBuilder.cpp     relative  time/iter   iters/s
numeric_upcast##cast_tinyint_nullable_as_smalli            1.05min    15.86m
numeric_upcast##cast_tinyint_as_smallint                   1.02min    16.28m
numeric_upcast##cast_tinyint_nullable_as_intege            1.17min    14.23m
numeric_upcast##cast_tinyint_as_integer                    1.08min    15.38m
numeric_upcast##cast_tinyint_nullable_as_bigint            1.70min     9.82m
numeric_upcast##cast_tinyint_as_bigint                     1.37min    12.20m
numeric_upcast##cast_tinyint_nullable_as_real              2.10min     7.93m
numeric_upcast##cast_tinyint_as_real                       2.18min     7.63m
numeric_upcast##cast_tinyint_nullable_as_double            2.49min     6.70m
numeric_upcast##cast_tinyint_as_double                     2.32min     7.18m
numeric_upcast##cast_tinyint_nullable_as_hugein            2.46min     6.78m
numeric_upcast##cast_tinyint_as_hugeint                    2.31min     7.21m
numeric_upcast##cast_smallint_nullable_as_integ            1.20min    13.94m
numeric_upcast##cast_smallint_as_integer                   1.09min    15.34m
numeric_upcast##cast_smallint_nullable_as_bigin            1.66min    10.05m
numeric_upcast##cast_smallint_as_bigint                    1.37min    12.17m
numeric_upcast##cast_smallint_nullable_as_real             2.20min     7.57m
numeric_upcast##cast_smallint_as_real                      2.31min     7.21m
numeric_upcast##cast_smallint_nullable_as_doubl            2.46min     6.78m
numeric_upcast##cast_smallint_as_double                    2.31min     7.22m
numeric_upcast##cast_smallint_nullable_as_hugei            2.59min     6.45m
numeric_upcast##cast_smallint_as_hugeint                   2.31min     7.21m
numeric_upcast##cast_integer_nullable_as_bigint            1.70min     9.83m
numeric_upcast##cast_integer_as_bigint                     1.37min    12.14m
numeric_upcast##cast_integer_nullable_as_real              1.72min     9.72m
numeric_upcast##cast_integer_as_real                       1.81min     9.19m
numeric_upcast##cast_integer_nullable_as_double            2.09min     7.98m
numeric_upcast##cast_integer_as_double                     1.94min     8.61m
numeric_upcast##cast_integer_nullable_as_hugein            2.09min     7.98m
numeric_upcast##cast_integer_as_hugeint                    1.94min     8.60m
numeric_upcast##cast_bigint_nullable_as_real               1.72min     9.67m
numeric_upcast##cast_bigint_as_real                        1.82min     9.16m
numeric_upcast##cast_bigint_nullable_as_double             2.09min     7.96m
numeric_upcast##cast_bigint_as_double                      1.94min     8.57m
numeric_upcast##cast_bigint_nullable_as_hugeint            2.08min     8.00m
numeric_upcast##cast_bigint_as_hugeint                     1.95min     8.56m
numeric_upcast##cast_hugeint_nullable_as_real              2.78min     6.00m
numeric_upcast##cast_hugeint_as_real                       2.49min     6.70m
numeric_upcast##cast_hugeint_nullable_as_double            2.57min     6.48m
numeric_upcast##cast_hugeint_as_double                     2.31min     7.21m
numeric_upcast##cast_real_nullable_as_double               2.20min     7.59m
numeric_upcast##cast_real_as_double                        1.92min     8.70m
```
After:

```
[...]hmarks/ExpressionBenchmarkBuilder.cpp     relative  time/iter   iters/s
numeric_upcast##cast_tinyint_nullable_as_smalli             25.15s    39.76m
numeric_upcast##cast_tinyint_as_smallint                    14.28s    70.05m
numeric_upcast##cast_tinyint_nullable_as_intege             36.05s    27.74m
numeric_upcast##cast_tinyint_as_integer                     18.55s    53.91m
numeric_upcast##cast_tinyint_nullable_as_bigint             54.77s    18.26m
numeric_upcast##cast_tinyint_as_bigint                      26.45s    37.81m
numeric_upcast##cast_tinyint_nullable_as_real               33.12s    30.19m
numeric_upcast##cast_tinyint_as_real                        20.59s    48.56m
numeric_upcast##cast_tinyint_nullable_as_double             55.68s    17.96m
numeric_upcast##cast_tinyint_as_double                      28.26s    35.38m
numeric_upcast##cast_tinyint_nullable_as_hugein             54.93s    18.21m
numeric_upcast##cast_tinyint_as_hugeint                     28.21s    35.45m
numeric_upcast##cast_smallint_nullable_as_integ             33.92s    29.48m
numeric_upcast##cast_smallint_as_integer                    18.45s    54.21m
numeric_upcast##cast_smallint_nullable_as_bigin             56.99s    17.55m
numeric_upcast##cast_smallint_as_bigint                     26.37s    37.92m
numeric_upcast##cast_smallint_nullable_as_real              32.48s    30.79m
numeric_upcast##cast_smallint_as_real                       19.67s    50.83m
numeric_upcast##cast_smallint_nullable_as_doubl             53.93s    18.54m
numeric_upcast##cast_smallint_as_double                     28.79s    34.73m
numeric_upcast##cast_smallint_nullable_as_hugei             53.07s    18.84m
numeric_upcast##cast_smallint_as_hugeint                    28.49s    35.10m
numeric_upcast##cast_integer_nullable_as_bigint            1.01min    16.51m
numeric_upcast##cast_integer_as_bigint                      27.04s    36.98m
numeric_upcast##cast_integer_nullable_as_real               35.55s    28.13m
numeric_upcast##cast_integer_as_real                        20.14s    49.66m
numeric_upcast##cast_integer_nullable_as_double             56.62s    17.66m
numeric_upcast##cast_integer_as_double                      28.39s    35.23m
numeric_upcast##cast_integer_nullable_as_hugein             57.36s    17.43m
numeric_upcast##cast_integer_as_hugeint                     28.49s    35.10m
numeric_upcast##cast_bigint_nullable_as_real                35.33s    28.31m
numeric_upcast##cast_bigint_as_real                         20.35s    49.14m
numeric_upcast##cast_bigint_nullable_as_double              56.84s    17.59m
numeric_upcast##cast_bigint_as_double                       28.41s    35.20m
numeric_upcast##cast_bigint_nullable_as_hugeint             58.48s    17.10m
numeric_upcast##cast_bigint_as_hugeint                      28.41s    35.20m
numeric_upcast##cast_hugeint_nullable_as_real              2.08min     8.01m
numeric_upcast##cast_hugeint_as_real                       2.01min     8.28m
numeric_upcast##cast_hugeint_nullable_as_double            1.89min     8.83m
numeric_upcast##cast_hugeint_as_double                     1.37min    12.13m
numeric_upcast##cast_real_nullable_as_double                59.15s    16.91m
numeric_upcast##cast_real_as_double                         29.15s    34.30m

```

### DIFF
--- a/velox/expression/CastExpr.cpp
+++ b/velox/expression/CastExpr.cpp
@@ -47,6 +47,38 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
   return nullptr;
 }
 
+// Returns true if casting from 'fromType' to 'toType' is a supported fast
+// upcast.
+bool isSupportedFastUpcast(const TypePtr& fromType, const TypePtr& toType) {
+  auto isIntegralType = [](const TypePtr& type) {
+    return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+        type == BIGINT() || type == HUGEINT();
+  };
+
+  auto isBasicNumericType = [&isIntegralType](const TypePtr& type) {
+    return isIntegralType(type) || type == REAL() || type == DOUBLE();
+  };
+
+  if (isIntegralType(fromType) && isBasicNumericType(toType)) {
+    if (fromType->cppSizeInBytes() < toType->cppSizeInBytes()) {
+      return true;
+    }
+    if (fromType == INTEGER() && toType == REAL()) {
+      return true;
+    }
+    if (fromType == BIGINT() || fromType == HUGEINT()) {
+      if (toType == REAL() || toType == DOUBLE()) {
+        return true;
+      }
+    }
+  }
+
+  if (fromType == REAL() && toType == DOUBLE()) {
+    return true;
+  }
+  return false;
+}
+
 } // namespace
 
 VectorPtr CastExpr::castFromDate(
@@ -897,6 +929,71 @@ void CastExpr::applyPeeled(
             fromType, toType, rows, context, input, result);
         break;
     }
+  } else if (isSupportedFastUpcast(fromType, toType)) {
+    switch (toType->kind()) {
+      case TypeKind::SMALLINT:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::SMALLINT,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      case TypeKind::INTEGER:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::INTEGER,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      case TypeKind::BIGINT:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::BIGINT,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      case TypeKind::REAL:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::REAL,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      case TypeKind::DOUBLE:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::DOUBLE,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      case TypeKind::HUGEINT:
+        result = VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH(
+            applyNumericUpcast,
+            TypeKind::HUGEINT,
+            fromType->kind(),
+            rows,
+            toType,
+            context,
+            input);
+        break;
+      default:
+        VELOX_UNREACHABLE();
+    }
   } else {
     switch (toType->kind()) {
       case TypeKind::MAP:
@@ -975,6 +1072,66 @@ VectorPtr CastExpr::applyTimestampToVarcharCast(
   // Update the exact buffer size.
   buffer->setSize(rawBuffer - buffer->asMutable<char>());
   return result;
+}
+
+template <TypeKind ToKind, TypeKind FromKind>
+VectorPtr CastExpr::applyNumericUpcast(
+    const SelectivityVector& rows,
+    const TypePtr& toType,
+    exec::EvalCtx& context,
+    const BaseVector& input) {
+  if constexpr (
+      FromKind != TypeKind::TINYINT && FromKind != TypeKind::SMALLINT &&
+      FromKind != TypeKind::INTEGER && FromKind != TypeKind::BIGINT &&
+      FromKind != TypeKind::HUGEINT && FromKind != TypeKind::REAL) {
+    VELOX_UNSUPPORTED(
+        "Cannot upcast from {} to {}", input.type(), toType->toString());
+  } else {
+    using ToNativeType = typename TypeTraits<ToKind>::NativeType;
+    using FromNativeType = typename TypeTraits<FromKind>::NativeType;
+
+    VectorPtr result;
+    context.ensureWritable(rows, toType, result);
+    (*result).clearNulls(rows);
+
+    if (input.isConstantEncoding()) {
+      auto constantInput = input.as<ConstantVector<FromNativeType>>();
+      if (constantInput->isNullAt(0)) {
+        return BaseVector::createNullConstant(
+            toType, rows.end(), context.pool());
+      }
+      auto constantValue = static_cast<ToNativeType>(constantInput->valueAt(0));
+      return std::make_shared<ConstantVector<ToNativeType>>(
+          context.pool(),
+          rows.end(),
+          /*isNull=*/false,
+          toType,
+          std::move(constantValue));
+    }
+
+    if (input.isFlatEncoding()) {
+      const auto simpleInput = input.asFlatVector<FromNativeType>();
+      auto flatResult = result->asFlatVector<ToNativeType>();
+
+      const FromNativeType* in =
+          simpleInput->template rawValues<FromNativeType>();
+      ToNativeType* out = flatResult->template mutableRawValues<ToNativeType>();
+
+      vector_size_t begin = rows.begin();
+      vector_size_t n = rows.end();
+      for (auto i = begin; i < n; ++i) {
+        if (rows.isValid(i)) {
+          // Converting large bigint/hugeint values to float/double directly may
+          // lose precision, but it's consistent with the implementation in
+          // velox/type/Conversions.h.
+          out[i] = static_cast<ToNativeType>(in[i]);
+        }
+      }
+      return result;
+    }
+
+    VELOX_FAIL("applyNumericUpcast only allows constant or flat input.");
+  }
 }
 
 template <typename TInput>
--- a/velox/expression/tests/CastExprTest.cpp
+++ b/velox/expression/tests/CastExprTest.cpp
@@ -4057,5 +4057,144 @@ TEST_F(CastExprTest, timestampToTimeCast) {
 
   testCast(TIMESTAMP(), TIME(), realTimestamps, expectedRealTime);
 }
+
+TEST_F(CastExprTest, numericUpcast) {
+  auto testNumericUpcast = [&](const VectorPtr& input,
+                               const VectorPtr& expected) {
+    std::vector<core::TypedExprPtr> inputs = {
+        std::make_shared<const core::FieldAccessTypedExpr>(
+            input->type(), "c0")};
+    auto castExpr =
+        std::make_shared<core::CallTypedExpr>(expected->type(), inputs, "cast");
+
+    testEncodings(castExpr, {input}, expected);
+  };
+
+  auto vectorSize = 1'000;
+
+  // Cast from TINYINT to SMALLINT.
+  const auto tinyintVector = makeFlatVector<int8_t>(vectorSize, [](auto i) {
+    return i % std::numeric_limits<int8_t>::max();
+  });
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<int16_t>(vectorSize, [](auto i) {
+        return i % std::numeric_limits<int8_t>::max();
+      }));
+
+  // Cast from TINYINT to INTEGER.
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<int32_t>(vectorSize, [](auto i) {
+        return i % std::numeric_limits<int8_t>::max();
+      }));
+
+  // Cast from TINYINT to BIGINT.
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<int64_t>(vectorSize, [](auto i) {
+        return i % std::numeric_limits<int8_t>::max();
+      }));
+
+  // Cast from TINYINT to REAL.
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<float>(vectorSize, [](auto i) {
+        return static_cast<float>(i % std::numeric_limits<int8_t>::max());
+      }));
+
+  // Cast from TINYINT to DOUBLE.
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<double>(vectorSize, [](auto i) {
+        return static_cast<double>(i % std::numeric_limits<int8_t>::max());
+      }));
+
+  // Cast from TINYINT to HUGEINT.
+  testNumericUpcast(
+      tinyintVector, makeFlatVector<int128_t>(vectorSize, [](auto i) {
+        return i % std::numeric_limits<int8_t>::max();
+      }));
+
+  // Cast from SMALLINT to INTEGER.
+  const auto smallintVector =
+      makeFlatVector<int16_t>(vectorSize, [](auto i) { return 1 + 2 * i; });
+  testNumericUpcast(
+      smallintVector,
+      makeFlatVector<int32_t>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from SMALLINT to BIGINT.
+  testNumericUpcast(
+      smallintVector,
+      makeFlatVector<int64_t>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from SMALLINT to REAL.
+  testNumericUpcast(
+      smallintVector, makeFlatVector<float>(vectorSize, [](auto i) {
+        return static_cast<float>(1 + 2 * i);
+      }));
+
+  // Cast from SMALLINT to DOUBLE.
+  testNumericUpcast(
+      smallintVector, makeFlatVector<double>(vectorSize, [](auto i) {
+        return static_cast<double>(1 + 2 * i);
+      }));
+
+  // Cast from SMALLINT to HUGEINT.
+  testNumericUpcast(
+      smallintVector, makeFlatVector<int128_t>(vectorSize, [](auto i) {
+        return static_cast<int128_t>(1 + 2 * i);
+      }));
+
+  // Cast from INTEGER to BIGINT.
+  const auto integerVector =
+      makeFlatVector<int32_t>(vectorSize, [](auto i) { return 1 + 2 * i; });
+  testNumericUpcast(
+      integerVector,
+      makeFlatVector<int64_t>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from INTEGER to REAL.
+  testNumericUpcast(
+      integerVector,
+      makeFlatVector<float>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from INTEGER to DOUBLE.
+  testNumericUpcast(
+      integerVector,
+      makeFlatVector<double>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from INTEGER to HUGEINT.
+  testNumericUpcast(
+      integerVector, makeFlatVector<int128_t>(vectorSize, [](auto i) {
+        return static_cast<int128_t>(1 + 2 * i);
+      }));
+
+  // Cast from BIGINT to REAL.
+  testNumericUpcast(
+      makeFlatVector<int64_t>(vectorSize, [](auto i) { return 1 + 2 * i; }),
+      makeFlatVector<float>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from BIGINT to DOUBLE.
+  testNumericUpcast(
+      makeFlatVector<int64_t>(vectorSize, [](auto i) { return 1 + 2 * i; }),
+      makeFlatVector<double>(vectorSize, [](auto i) { return 1 + 2 * i; }));
+
+  // Cast from BIGINT to HUGEINT.
+  testNumericUpcast(
+      makeFlatVector<int64_t>(vectorSize, [](auto i) { return 1 + 2 * i; }),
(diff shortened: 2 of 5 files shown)


rui-mo:
cc: @zhouyuan

velox/expression/CastExpr.cpp
```diff
@@ -955,6 +1037,58 @@ VectorPtr CastExpr::applyTimestampToVarcharCast(
...
+      out[i] = static_cast<ToNativeType>(in[i]);
+      out[i + 1] = static_cast<ToNativeType>(in[i + 1]);
+      out[i + 2] = static_cast<ToNativeType>(in[i + 2]);
+      out[i + 3] = static_cast<ToNativeType>(in[i + 3]);
+    }
+    for (; i < n; ++i) {
+      out[i] = static_cast<ToNativeType>(in[i]);
+    }
+    return result;
+  }
```
MBkkt: Hi, manual loop unrolling is actually worse when you use flags like sse4/avx/avx2 in compile time.
It's because compiler already doing vectorization for such loops, and if you have manual vectorization it will produce two vectorized loops both are same but additional checks/etc

You can compare here:
https://godbolt.org/z/38av815W5

velox/benchmarks/basic/IntegralUpcastBenchmark.cpp
```diff
@@ -0,0 +1,59 @@
...
+      nullptr);
+
+  benchmarkBuilder
+      .addBenchmarkSet(
+          "integral_upcast",
+          vectorMaker.rowVector(
+              {"c0", "c1", "c2"}, {tinyIntInput, smallIntInput, integerInput}))
+      .addExpression("cast_tinyint_as_smallint", "cast(c0 as smallint)")
+      .addExpression("cast_smallint_as_integer", "cast(c1 as integer)")
+      .addExpression("cast_integer_as_bigint", "cast(c2 as bigint)")
```
MBkkt:
```suggestion
      .addExpression("cast_tinyint_as_smallint", "cast(c0 as smallint)")
      .addExpression("cast_tinyint_as_integer", "cast(c0 as integer)")
      .addExpression("cast_smallint_as_integer", "cast(c1 as integer)")
      .addExpression("cast_tinyint_as_bigint", "cast(c0 as bigint)")
      .addExpression("cast_smallint_as_bigint", "cast(c1 as bigint)")
      .addExpression("cast_integer_as_bigint", "cast(c2 as bigint)")
```

I think it will be nice to have this 3 additional cases in the benchmark

velox/expression/tests/CastExprTest.cpp
```diff
@@ -3992,5 +3992,43 @@ TEST_F(CastExprTest, timeToTimestampCast) {
     assertEqualVectors(expected, result);
   }
 }
+
+TEST_F(CastExprTest, integeralUpcast) {
```
MBkkt: Same about tests, I think it's important to also check
1. tinyint => integer
2. tinyint => bigint
3. smallint => bigint

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,50 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
   return nullptr;
 }
 
+bool isIntegralType(const TypePtr& type) {
```
MBkkt: I think this optimization also should work for hugeint

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,50 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
...
+  switch (type->kind()) {
+    case TypeKind::TINYINT:
+    case TypeKind::SMALLINT:
+    case TypeKind::INTEGER:
+      return true;
+    case TypeKind::BIGINT: {
+      if (type->isDecimal()) {
+        return false;
+      }
```
MBkkt: I think this approach isn't correct, what if someone else created some new logical velox type that inherits from velox native types?

I think instead you need to check in each case something like
`return type == velox::BIGINT();`

https://github.com/facebookincubator/velox/blob/2302437e42362703852b3db5de0eb768cd0d3799/velox/type/Type.h#L793

velox/expression/CastExpr.cpp
```diff
@@ -955,6 +1037,58 @@ VectorPtr CastExpr::applyTimestampToVarcharCast(
   return result;
 }
 
+template <TypeKind ToKind, TypeKind FromKind>
+VectorPtr CastExpr::applyUpcastIntegral(
```
MBkkt: shouldn't this also works at least for:
1. tinyint, smallint => real
2. tinyint, smallint, integer, real => double


Maybe also 
1. some-larger-integer => real/double 
2. real/double to some-int

But I'm not sure how velox handle overflows in such case

MBkkt:
@rui-mo Can you explain it to me?

You mentioned in PR description that hotspot is `forEachSetBit`

But at the same time you wrote that this PR optimizes only **when rows are all selected.**

> This PR optimizes the integral upcast by performing the cast directly on the
raw values within loops when rows are all selected.

But in such case shouldn't be `forEachSetBit`, right?

Because Velox have such code
```cpp
template <typename Callable>
inline void SelectivityVector::applyToSelected(Callable func) const {
  if (isAllSelected()) {
    const auto end = end_;
    for (vector_size_t row = begin_; row < end; ++row) {
      func(row);
    }
  } else {
    bits::forEachSetBit(bits_.data(), begin_, end_, func);
  }
}
```

jinchengchenghh:
This optimization maybe similar to following
https://github.com/facebookincubator/velox/blob/4056e41e6efadd20622e92e1b04162d59276d63c/velox/connectors/hive/HivePartitionFunction.cpp#L119-L123

jinchengchenghh:
We may need to optimize applyToSelected for all the functions

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,49 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
...
 
+bool isBasicNumericType(const TypePtr& type) {
+  if (type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT() || type == REAL() || type == DOUBLE()) {
+    return true;
+  }
+  return false;
+}
+
+#define VELOX_DYNAMIC_BASIC_NUMERIC_TEMPLATE_TYPE_DISPATCH(             \
```
jinchengchenghh: Could we use VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH and check std::is_arithmetic_v<T>

velox/expression/CastExpr.cpp
```diff
@@ -875,6 +918,65 @@ void CastExpr::applyPeeled(
             fromType, toType, rows, context, input, result);
         break;
     }
+  } else if (
+      isBasicNumericType(fromType) && isBasicNumericType(toType) &&
+      ((fromType->cppSizeInBytes() < toType->cppSizeInBytes()) ||
+       (fromType == INTEGER() && toType == REAL()) ||
+       (fromType == BIGINT() && toType == DOUBLE()))) {
+    switch (toType->kind()) {
+      case TypeKind::SMALLINT:
```
jinchengchenghh, line 934: Could we use result = VELOX_DYNAMIC_BASIC_NUMERIC_TEMPLATE_TYPE_DISPATCH(
            applyNumericUpcast,
            toType->kind(),
            fromType->kind(),
            rows,
            toType,
            context,
            input);

velox/expression/CastExpr.cpp
rui-mo: Thanks for the pointer! I removed the loop unrolling.

velox/benchmarks/basic/IntegralUpcastBenchmark.cpp
rui-mo: Added all relevant cases in the benchmark, thanks.

velox/expression/tests/CastExprTest.cpp
rui-mo: Added all relevant cases in the test, thanks.

velox/expression/CastExpr.cpp
rui-mo: Hugeint is used to represent the decimal type. Casting an integer type to a decimal type requires rescaling to match the target scale, which is a different operation and needs special handling. Therefore, I excluded decimal from this optimization.

velox/expression/CastExpr.cpp
rui-mo: Thanks for the suggestion. Updated the type comparison.

velox/expression/CastExpr.cpp
rui-mo: Thanks for mentioning this. The type range follows the order `tinyint < smallint < integer < bigint < real < double`. I included all the upcast cases between these types.

For casts from real or double to integer types, potential overflow errors must be detected and either raised or result in a null return depending on configuration, and values like `NaN` and `Inf` might require special handling. Since this is a more complex scenario than simple upcasting, this PR does not include them.

velox/expression/CastExpr.cpp
MBkkt: Hugeint isn't only decimal. I think hugeint also just int128_t that can be.
See example int128 https://clickhouse.com/docs/sql-reference/data-types/int-uint

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,51 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
...
 }
 
+bool isIntegralType(const TypePtr& type) {
+  return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT();
+}
+
+bool isBasicNumericType(const TypePtr& type) {
+  return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT() || type == REAL() || type == DOUBLE();
```
MBkkt:
```suggestion
  return isIntegralType(type) || type == REAL() || type == DOUBLE();
```

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,51 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
   return nullptr;
 }
 
+bool isIntegralType(const TypePtr& type) {
+  return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT();
```
MBkkt: It will be nice to have HUGEINT here, not for decimal but for just int128

rui-mo: Is int128 used by any type other than the long decimal type in Presto? I assume it's only for the long decimal type.

rui-mo:
> @rui-mo Can you explain it to me?
> 
> You mentioned in PR description that hotspot is `forEachSetBit`
> 
> But at the same time you wrote that this PR optimizes only **when rows are all selected.**
> 
> > This PR optimizes the integral upcast by performing the cast directly on the
> > raw values within loops when rows are all selected.
> 
> But in such case shouldn't be `forEachSetBit`, right?
> 
> Because Velox have such code
> 
> ```c++
> template <typename Callable>
> inline void SelectivityVector::applyToSelected(Callable func) const {
>   if (isAllSelected()) {
>     const auto end = end_;
>     for (vector_size_t row = begin_; row < end; ++row) {
>       func(row);
>     }
>   } else {
>     bits::forEachSetBit(bits_.data(), begin_, end_, func);
>   }
> }
> ```

@MBkkt Sorry for the confusion. This PR improves performance in both scenarios. When not all rows are selected (for example, when null values are present and are deselected before casting), the performance hotspot is as follows.

has_null

And when all rows are selected, the hotspot is as follows.

no_null

Although Velox already applies the for-loop optimization you mentioned above when all rows are selected, `CastExpr::applyToSelectedNoThrowLocal` still includes a try-catch block for potential error handling, which is unnecessary in upcast cases. Eliminating this overhead is one of the reasons for the performance improvement in the all-selected scenario.

I will include both nullable and non-nullable benchmark results in the PR description to verify that both scenarios see performance improvements.

MBkkt:
@rui-mo Thanks, I understand now.

I think it will be useful to move this optimization to lambda that supplied by CastExpr to `SelectivityVector::applyToSelected`, but it's separate task

velox/expression/CastExpr.cpp
MBkkt: > Is int128 used by any type other than the long decimal type in Presto?

No, you're right.
But Velox can be used not only for Spark and Presto. 
For an example, in SereneDB we need to provide more types.

Yuhta:
I think `applyToSelected` is already optimized (see https://github.com/facebookincubator/velox/pull/10301/files#diff-3d8a3c26f9d059ef07d7c31f08b2c605c7e748206493887da6023d2a085f00b8), the code in https://github.com/facebookincubator/velox/blame/4056e41e6efadd20622e92e1b04162d59276d63c/velox/connectors/hive/HivePartitionFunction.cpp#L119-L123 is outdated and could be removed.

The only thing need is to avoid try catch when it is not necessary.

rui-mo:
@MBkkt @Yuhta Thanks for sharing your insights. I ran a few experiments to investigate where the performance improvement comes from. For CAST(integer AS bigint), the original performance is as follows:

```
numeric_upcast##cast_integer_nullable_as_bigint            1.74min     9.55m
numeric_upcast##cast_integer_as_bigint                     1.38min    12.10m
```

After removing the try-catch from `CastExpr::applyToSelectedNoThrowLocal` and replacing `callFollyTo` with `static_cast` (bypassing overflow checks and error handling), I observed a slight performance gain, shown below.

```
numeric_upcast##cast_integer_nullable_as_bigint            1.66min    10.03m
numeric_upcast##cast_integer_as_bigint                     1.26min    13.22m
```

I further removed the try-catch block from `CastExpr::applyCastKernel`, which resulted in the following performance.

```
numeric_upcast##cast_integer_nullable_as_bigint            1.33min    12.51m
numeric_upcast##cast_integer_as_bigint                      58.44s    17.11m
```

I then used perf record to identify the performance hotspot and found the results below, which show that the `valueAt` and `set` functions are the main time consumers.

Screenshot 2025-11-13 at 14 41 10

In summary, the performance gains are likely attributed to:

1) Eliminating try-catch blocks when error handling is unnecessary.
2) Improved auto-vectorization and lower function call overhead after replacing `valueAt` and `set` with direct access.
3) Avoiding overflow checks.

velox/expression/CastExpr.cpp
rui-mo: I added hugeint and included tests and benchmarks, thanks.

velox/expression/CastExpr.cpp
rui-mo: I changed to `VELOX_DYNAMIC_SCALAR_TEMPLATE_TYPE_DISPATCH`, thanks.

velox/expression/CastExpr.cpp
rui-mo: The dispatch function supports only one-parameter dispatching, and we need the switch cases here.

jinchengchenghh: Use fallthrough? maybe if is better

rui-mo: I didn't understand. Why do we need `fallthrough` here?

jinchengchenghh: We have checked isBasicNumericType before, so we may don't need to switch case again

jinchengchenghh: Ok, I know, here is compiler issue, though toType->kind(), == TypeKind::SMALLINT, we still need to expand

velox/expression/CastExpr.cpp
```diff
@@ -875,6 +884,76 @@ void CastExpr::applyPeeled(
             fromType, toType, rows, context, input, result);
         break;
     }
+  } else if (
+      isIntegralType(fromType) && isBasicNumericType(toType) &&
+      ((fromType->cppSizeInBytes() < toType->cppSizeInBytes()) ||
+       (fromType == INTEGER() && toType == REAL()) ||
+       (fromType == BIGINT() && toType == REAL()) ||
+       (fromType == BIGINT() && toType == DOUBLE()))) {
```
jinchengchenghh: Can cast int128 to double also benefit from this?

rui-mo: Yes, I added `CAST(hugeint AS real)` and `CAST(hugeint AS double)` as well as their tests and benchmarks. Thanks.

velox/expression/CastExpr.cpp
```diff
@@ -875,6 +884,78 @@ void CastExpr::applyPeeled(
...
         break;
     }
+  } else if (
+      isIntegralType(fromType) && isBasicNumericType(toType) &&
+      ((fromType->cppSizeInBytes() < toType->cppSizeInBytes()) ||
+       (fromType == INTEGER() && toType == REAL()) ||
+       (fromType == BIGINT() && toType == REAL()) ||
+       (fromType == BIGINT() && toType == DOUBLE()) ||
+       (fromType == HUGEINT() && toType == REAL()) ||
+       (fromType == HUGEINT() && toType == DOUBLE()))) {
```
MBkkt: real to double also should be ok?

MBkkt (commented):
Overall looks good to me.

I think more general approach to how avoid "try catch"/etc overhead can be developed separately

velox/expression/CastExpr.cpp
rui-mo: Nice catch, thanks. I added its support as well as tests and benchmarks.
I opened https://github.com/facebookincubator/velox/issues/15506 for the discussion of avoiding "try catch" in cast.

velox/expression/CastExpr.cpp
```diff
@@ -47,6 +47,38 @@ const tz::TimeZone* getTimeZoneFromConfig(const core::QueryConfig& config) {
...
+  auto isIntegralType = [](const TypePtr& type) {
+    return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+        type == BIGINT() || type == HUGEINT();
+  };
+
+  auto isBasicNumericType = [&isIntegralType](const TypePtr& type) {
+    return isIntegralType(type) || type == REAL() || type == DOUBLE();
+  };
+
+  if (isIntegralType(fromType) && isBasicNumericType(toType)) {
```
jinchengchenghh, line 62: Why the fromType cannot be isBasicNumericType?

MBkkt: https://github.com/facebookincubator/velox/pull/15458#discussion_r2518716257

> for the casts from real and double

I think Velox have some special handling of non finite floating point to integer cast?

jinchengchenghh: Make sense

PR #15460, opened 2025-11-11
fix: Start prefetch from the first split in table scan

rui-mo:
`getSplit()` reads a split, while `checkPreload()` preloads one. Previously, 
prefetching began with the second split. By calling `checkPreload()` before
`getSplit()`, prefetching can now start from the first split.


rui-mo:
cc: @zhli1142015 @FelixYBW

velox/exec/TableScan.cpp
```diff
@@ -528,4 +520,21 @@ void TableScan::close() {
       TableScan::kNumRunningScaleThreads,
       RuntimeCounter(scaledStats.numRunningDrivers));
 }
+
+void TableScan::createDatasource() {
+  if (connectorQueryCtx_ == nullptr) {
+    connectorQueryCtx_ = operatorCtx_->createConnectorQueryCtx(
+        connector_->connectorId(), planNodeId(), connectorPool_);
+  }
+  if (dataSource_ == nullptr) {
```
MBkkt: I think such asserts make sense?
```
VELOX_DCHECK_NULL(connectorQueryCtx_);
VELOX_DCHECK_NULL(dataSource_);
```

FelixYBW:
To be more clear, currently velox's split prefetch starts after the first split finished. If a partition only has 2 split, the second one won't be prefetched.

The PR is going to fix the issue. It starts the split prefetch as long as the first one starts to read. So the perf gain rely on the split number in a partition. If a partition has two split, the PR can boost perf by 2x almost.

The chart below is the traceview of main and IO threads in which the split and rowgroup are fetched.

image

@pedroerp

Yuhta:
> If a partition only has 2 split

This sounds like a bigger issue, the workload is doomed to be skewed with such a setup

FelixYBW:
> > If a partition only has 2 split
> 
> This sounds like a bigger issue, the workload is doomed to be skewed with such a setup

No, It's common in spark partitions. Like you generate 100 parquet file but use 50 partitions. Data isn't skewed.

Yuhta:
@FelixYBW A split is usually a stripe (row group) if the number of files is not large enough.  With one file per split, the split level preloading is not that useful anyway.  We need stripe level prefetch in this case (it's already there in Parquet).

FelixYBW:
> @FelixYBW A split is usually a stripe (row group) if the number of files is not large enough. With one file per split, the split level preloading is not that useful anyway. We need stripe level prefetch in this case (it's already there in Parquet).

In spark workload, it's very common one split has 1 to multiple rowgroups, each partition has part of parquet file to hundreds of parquet files. We can't assume the split number and row group number in each partition. We must enable both rowgroup prefetch and split prefetch in our workloads, it's one main reason Gluten outperform Spark in TPCDS workloads.

velox/exec/TableScan.cpp
rui-mo: If `checkPreload()` skips the data source creation, it can also be created in the `getSplit()`.

Yuhta:
> it's very common one split has 1 to multiple rowgroups

It's usually either 1 split per row group or 1 split per file, how did you generate your data?

FelixYBW:
> It's usually either 1 split per row group or 1 split per file, how did you generate your data?

Hi @Yuhta we can't make any assumption on this in Spark or how customer generate spark files. We noted the issue is from a customer workload.

What the PR hurt? In our viewpoint, it's just a bug fix to Velox. In our test, the worst case is it doesn't benefit performance.

@pedroerp your thought?

zhli1142015:
We also have a similar change internally. We discovered this issue while tuning the performance of the 1TB TPC-DS workload. In this dataset, due to partitioning, a large portion of the files end up being small—often smaller than Spark’s configured Parquet block (row-group) size. As a result, loading Parquet footers accounts for a significant portion of the scan operator’s time. Since Spark merges many of these small splits into the same task, the split prefetch feature becomes particularly beneficial in this scenario.

rui-mo:
Hi @Yuhta, the “1 split per row group or 1 split per file” pattern is true only for clean, controlled datasets. But in practice, a Parquet file can contain multiple splits when the file is larger than Spark’s `maxPartitionBytes`, when HDFS/object-store block boundaries force additional slicing, or when Spark intentionally subdivides large files for parallelism. A split can contain multiple row groups when its byte boundaries don’t align with Parquet row-group boundaries such as when row groups are smaller than the split size, when a file contains multiple row groups due to how it was written or compacted, or when row-group sizes are irregular.
In real production workloads, both situations occur frequently, which is why enabling both split-level and row-group–level prefetching is important for optimal performance.

Yuhta:
The change itself should be ok; it's just if one split contains many row groups, row group level prefetch would benefit much more than split level prefetch (i.e. the benefit of this change is not large).  The extreme scenario that fixing this would reduce the wall time by 2x would only happen if there is only one row group in the split AND there are only 2 splits in each partition, which almost never happen in prod.

FelixYBW:
> The change itself should be ok; it's just if one split contains many row groups, row group level prefetch would benefit much more than split level prefetch (i.e. the benefit of this change is not large). The extreme scenario that fixing this would reduce the wall time by 2x would only happen if there is only one row group in the split AND there are only 2 splits in each partition, which almost never happen in prod.

Could you approve the PR then? We needn't waste our time to argue on whether it happens or not. It may never happen in Meta but it does happen in Spark workloads.

Yuhta:
@FelixYBW It's not just about whether it happens or not; if that's the case, fixing the partitioning if you can (e.g. in TPC-DS you have control over the data), it would result in more than 2x improvement in this case.

velox/exec/TableScan.cpp
MBkkt: @rui-mo 

You wrote this (two times)
```
 if (dataSource_ == nullptr) {
    createDatasource();
  }
```

there's no other calls of  createDatasource(), so if inside  createDatasource() is just confuse reader

zhli1142015:
I think we should add a unit test to cover this change, so that we can prevent it from being unintentionally altered in the future.

rui-mo:
Let's hold on this PR for the test. Thanks everyone for the input.

velox/connectors/hive/HiveDataSource.h
```diff
@@ -62,10 +62,6 @@ class HiveDataSource : public DataSource {
 
   std::unordered_map<std::string, RuntimeMetric> getRuntimeStats() override;
 
-  bool allPrefetchIssued() const override {
```
Yuhta, line 65: What's the reason for dropping this?  We probably still need to keep this interface even it is not fully used today, as the stripe level prefetch is not finished yet for all file formats.

velox/exec/Task.cpp
```diff
@@ -1600,7 +1600,9 @@ void Task::addSplitLocked(
   ++taskStats_.numQueuedSplits;
 
   if (split.hasConnectorSplit()) {
-    VELOX_CHECK_NULL(split.connectorSplit->dataSource);
```
Yuhta, line 1602: In which case this would be violated?  If a data source is prefetched, it should not go through this path.

rui-mo: In unit tests, a split is usually created and reused, for example:

https://github.com/facebookincubator/velox/blob/75b2349f2c40ac121941ad61a47beb683ab5bcc7/velox/exec/tests/TableScanTest.cpp#L413

Without this change, tests failed during adding split at below.

https://github.com/facebookincubator/velox/blob/75b2349f2c40ac121941ad61a47beb683ab5bcc7/velox/exec/Task.cpp#L1603

I was wondering if we can keep its reusability if it won't impact the runtime code path.

velox/connectors/hive/HiveDataSource.h
rui-mo: The `dataSource_->allPrefetchIssued()` requires the split reader to be created, but since the split reader hasn't been created for the first split, we need to bypass this condition check.

https://github.com/facebookincubator/velox/blob/75b2349f2c40ac121941ad61a47beb683ab5bcc7/velox/connectors/hive/HiveDataSource.h#L65-L66

Since `checkPreload()` is the only function calling `allPrefetchIssued()` and I noticed that both DWRF and Parquet return true in `allPrefetchIssued()`, I was wondering if it’s still necessary to keep it. Do you have any suggestions on this?

Yuhta: This method is there to coordinate the interplay between split preload and stripe preload.  Since the DWRF/Nimble stripe preload is still WIP, we don't have much experience running both at same time.  It's better to keep it, and only drop it after we get enough experience running both at same time and confirm we don't need any coordination between them.  Can you change the check in table scan to shortcut this check when it is the first split?

velox/exec/Task.cpp
Yuhta: Splits cannot be reused due to this.  It's not a problem in prod code since we are not reusing splits, and we believe the check brings some value to prevent mistake in plan/runtime setup.  Can you fix the unit test to recreate the split (not the file though) when needed?

velox/connectors/hive/HiveDataSource.h
rui-mo: Thanks, I see. This method has been added back. Regarding the idea of “shortcut this check when it is the first split”, I noticed that once `splitPreloader_` is initialized during the first call, the `checkPreload` method becomes a no-op. So there’s no meaningful reason to keep this check for subsequent calls.

velox/exec/Task.cpp
rui-mo: Thanks for the clarification. I’ve updated the unit test to recreate the split from file path.

velox/connectors/hive/HiveDataSource.h
Yuhta: I see.  The situation how we use `allPrefetchIssued()` might need to be revisited later, but let's keep the method itself for now.

Yuhta:
@rui-mo Do you have a unit test specific for this change (other than the refactoring)?

rui-mo:
@Yuhta This test `TEST_F(TableScanTest, preloadSplits)` is specific for this change. Without this PR, the number of prefetched splits is 9 rather than 10.

velox/exec/TableScan.cpp
rui-mo: Thanks. This change has been revised.

Yuhta (approved):

rui-mo:
@kKPulla @Yuhta Could someone help re-import and merge this change? Thanks!

kKPulla:
> @kKPulla @Yuhta Could someone help re-import and merge this change? Thanks!

Sorry I forgot to update here. There are couple of internal tests that are failing, I will take a look at those today and try to land this by EOD.

kKPulla:
@rui-mo The test failures from the CI signals above are fixed in #16076 Can you please rebase? Thanks!

rui-mo:
@kKPulla The workflows work now. Thanks for your help.

rui-mo:
@Yuhta @kKPulla Thank you for taking the time to review and help land this change.

PR #15461, opened 2025-11-11
feat: Add Iceberg partition name generator

PingLiuPing:
Implement partition name generation for writing partitioned Iceberg tables, addressing the differences between Hive and Iceberg partition naming conventions.

1. Add IcebergPartitionNameGenerator class that extends PartitionNameGenerator to handle Iceberg-specific partition transforms (identity, bucket, truncate, year, month, day, hour) and generate Iceberg-compliant partition paths.
2. Updated IcebergDataSink to use the new partition name generator.
3. Refactor PartitionIdGenerator class to make it only generator partition IDs.
4. Test all transform types (identity, bucket, truncate, year, month, day, hour). Add tests for edge cases: null values, special characters, URL encoding, multiple transforms on same column, timestamp edge cases: leap years, negative timestamps, year boundaries, extreme dates (±10000).


PingLiuPing:
@mbasmanova This PR continues from https://github.com/facebookincubator/velox/pull/15035#pullrequestreview-3427716555. This is the last piece of the preliminary work for partition transforms. Next I’ll submit a PR to integrate them with IcebergDataSink.

velox/connectors/hive/PartitionIdGenerator.h
```diff
@@ -52,16 +54,52 @@ class PartitionIdGenerator {
...
+  // generation.
+  //
+  // @param input RowVector containing the partition column values (either
+  // original or transformed).
+  // @param result Output vector to store partition IDs, indexed by row number.
+  void computeAndSavePartitionIds(
+      const RowVectorPtr& input,
+      raw_vector<uint64_t>& result);
+
+  const std::vector<column_index_t> partitionChannels_;
```
mbasmanova, line 82: Would you put const member variables before the rest? Also, make sure all methods appear before member variables.

- public methods
- protected methods
- private methods
- protected variables
- private variables

velox/connectors/hive/iceberg/IcebergPartitionIdGenerator.h
```diff
@@ -0,0 +1,134 @@
...
+/// Computes partition keys by applying Iceberg partition transforms
+/// (e.g., bucket, truncate, year, month, day, hour, identity) to source columns
+/// and maps the transformed values to sequential partition IDs. Maintains a
+/// mapping from partition values to partition IDs and generates
+/// Iceberg-compliant partition path names for file organization.
+///
+/// Each unique combination of transformed partition keys is assigned
+/// a sequential ID (0, 1, 2, ...) up to a configurable maximum.
+/// Uses TransformEvaluator for batch evaluation of transforms.
+class IcebergPartitionIdGenerator : public PartitionIdGenerator {
```
mbasmanova: I assume the only difference from Hive is the logic for producing partition keys. If that's the case, then, perhaps, add a virtual method to return partition keys to PartitionIdGenerator and keep all logic for producing partition IDs there.

velox/connectors/hive/PartitionIdGenerator.h
PingLiuPing: Thanks, will do. Might need to adjust initializer list little bit.

velox/connectors/hive/PartitionIdGenerator.cpp
```diff
@@ -60,11 +60,30 @@ PartitionIdGenerator::PartitionIdGenerator(
...
+      partitionPathAsLowerCase_(partitionPathAsLowerCase),
+      pool_(pool) {
+  VELOX_USER_CHECK(
+      !partitionChannels_.empty(), "There must be at least one partition key.");
+}
+
 void PartitionIdGenerator::run(
     const RowVectorPtr& input,
     raw_vector<uint64_t>& result) {
+  result.resize(input->size());
```
mbasmanova: It seems we only need to provide a way to process 'input' before it is used to compute partition IDs.

PingLiuPing: Yes, you are right. The input here is different between Hive and Iceberg.
For Hive it uses the original value from input. But for Iceberg it should use the value after applying transformation.
I override the `run` method in IcebergPartitionIdGenerator and process the input there.
An alternative way is process this in the caller `HiveDataSink::computePartitionAndBucketIds`, I can override this method `HiveDataSink::computePartitionAndBucketIds`  in IcebergDataSink.

mbasmanova: Can we add a prepareInput(const RowVectorPtr& input) virtual method to PartitionIdGenerator with default implementation just returning 'input' as-is?

mbasmanova: > An alternative way is process this in the caller HiveDataSink::computePartitionAndBucketIds, I can override this method HiveDataSink::computePartitionAndBucketIds in IcebergDataSink.

That might be a good option as well.

velox/connectors/hive/iceberg/IcebergPartitionIdGenerator.h
PingLiuPing: @mbasmanova  Thank you for the comment.

There are other differences:
1. Partition name, `partitionName` should be override to support Iceberg specific requirement.
2. The logic to initialize hashers is different. For Hive it uses the child type of input RowType directly. But for Iceberg this should be the result type of iceberg transform.
3. And since 2, we should provide `IcebergPartitionSpec` to `IcebergPartitionIdGenerator`.
4. And since 2 I think it might makesense to perform the transform when generating the partition Ids. And this way we can re-use lots of logic from HiveDataSink (`partitionIdGenerator_`). 
5. `savePartitionValues` is different since for Iceberg we created a new RowVector that contains partition column only, the channel here is continuous.

velox/connectors/hive/PartitionIdGenerator.cpp
PingLiuPing: @mbasmanova  Thank you.

I updated the code with `add a prepareInput(const RowVectorPtr& input) virtual method to PartitionIdGenerator`.

velox/connectors/hive/PartitionIdGenerator.h
```diff
@@ -52,11 +54,16 @@ class PartitionIdGenerator {
   /// style. It is derived from the partitionValues_ at index partitionId.
   /// Partition keys appear in the order of partition columns in the table
   /// schema.
-  std::string partitionName(uint64_t partitionId) const;
+  virtual std::string partitionName(uint32_t partitionId) const;
 
- private:
-  static constexpr const int32_t kHasherReservePct = 20;
+ protected:
+  PartitionIdGenerator(
```
mbasmanova: It looks like ParitionIdGeneration does more than just generate partition IDs. Is this intentional? Perhaps, this needs to be revisited.

The logic of generating partition IDs seems to be the same between Hive and Iceberg. The only difference is that Hive can generate partition IDs from raw input, but Iceberg needs to pre-process the input before feeding it into partition ID generator. You had a idea of doing this pre-processing outside of partition ID generator, somewhere in the DataSink. Let's explore that option.

What are the responsibilities of PartitionIdGenerator outside of generating partition IDs? What would be a cleaner way to separate these out?

mbasmanova:
My reading of PartitionIdGenerator is the following.

> void run(const RowVectorPtr& input, raw_vector<uint64_t>& result)

'run' method takes input values and computes partition IDs. It relies on 'partitionChannels' specified in the ctor. In addition to assigning partition IDs to each row of input, this method also populates 'partitionValues_' RowVector. 

'partitionValues_' has one row per unique partition ID. Row N contains partition keys for for partition ID N.  

This API doesn't any any changes. IcebergDataSink just needs to call it on transformed data.

> std::string partitionName(uint64_t partitionId) const;

This API looks up partition keys in partitionValues_'s vector at row number partitionId and formats these into Hive partition path. This logic is different between Hive and Iceberg. We can change this API to not compute directory path, but to return a reference to partitionValues_ vector. Then, HiveDataSink can apply its own formatting and IcebergDataSink its own.

This way, PartitionIdGenerator can be a generic utility responsible for producing unique partition IDs and tracking partition key values for each partition. The callers (data sinks) will be responsible for formatting partition keys into paths and pre-processing inputs.

What do you think?

PingLiuPing:
@mbasmanova  Thank you for looking at this comprehensively.

Yes the current design of `PartitionIdGenerator` breaks the single responsibility principle.  I totally agree with your suggestions.
What I need to do are:
1.  Changing the current API `std::string partitionName(uint64_t partitionId) const;` to `const RowVectorPtr& partitionValues()`.
2. Enhance `HivePartitionUtil` to support generating partition name based on input `partitionId` and `partitionValues`.
3. Similarly enhance `IcebergPartitionPath` to support same.
4. Override `computePartitionAndBucketIds` in `IcebergDataSink`. Need to move  these three members from `IcebergPartitionIdGenerator` to `IcebergDataSink` for evaluating the transform.
```
  const IcebergPartitionSpecPtr partitionSpec_;
  const ConnectorQueryCtx* connectorQueryCtx_;
  const std::unique_ptr<TransformEvaluator> transformEvaluator_;
```
5. Adjust existing Hive connect testcase that depends on API `partitionName`.

mbasmanova:
I'm also seeing that HivePartitionUtil is named poorly. Perhaps, we could rename it to HivePartitionPath.

PingLiuPing:
> I'm also seeing that HivePartitionUtil is named poorly. Perhaps, we could rename it to HivePartitionPath.

Yes you are right. How about PartitionNameGenerator in corresponds to PartitionIdGenerator ?

mbasmanova:
> PartitionNameGenerator

Sounds good. Thanks.

velox/connectors/hive/tests/HivePartitionUtilTest.cpp
PingLiuPing, line 1: @mbasmanova  I renamed `HivePartitionUtil.h` to` PartitionNameGenerator.h`. But I didn't change this test file name yet. Let me know if it is necessary to change it.

velox/connectors/hive/PartitionIdGenerator.h
```diff
@@ -70,29 +72,32 @@ class PartitionIdGenerator {
 
   // Copies partition values of 'row' from 'input' into 'partitionId' row in
   // 'partitionValues_'.
-  void savePartitionValues(
-      uint64_t partitionId,
+  virtual void savePartitionValues(
```
mbasmanova: Why is savePartitionValues became virtual? What do we need to customize?

PingLiuPing: Thanks, just pushed a new commit, has reverted these changes.

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
```diff
@@ -0,0 +1,176 @@
...
+      partitionSpec_(partitionSpec) {
+  formatters_.reserve(partitionSpec_->fields.size());
+  for (const auto& field : partitionSpec_->fields) {
+    formatters_.emplace_back(
+        std::make_shared<const IcebergPartitionValueFormatter>(
+            field.transformType, partitionPathAsLowerCase));
+  }
+}
+
+std::string IcebergPartitionNameGenerator::partitionName(
```
mbasmanova: What's the difference between this logic and FileUtils::makePartName ?

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "velox/common/encode/Base64.h"
+#include "velox/functions/prestosql/URLFunctions.h"
+#include "velox/vector/SimpleVector.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+namespace {
+
+template <TypeKind Kind>
+std::pair<std::string, std::string> makePartitionKeyValueString(
```
mbasmanova: This looks like a copy-paste from existing code. Is there any difference?

PingLiuPing: There is a  difference when the partition column value is `null`, Hive returns `[c1, ""]`, Iceberg returns [c1, "null"].

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
PingLiuPing: @mbasmanova  Thank you for the comment.

For` FileUtils::makePartName`, it escapes following chars:

```
constexpr auto charsToEscape = folly::make_array(
    '"',
    '#',
    '%',
    '\'',
    '*',
    '/',
    ':',
    '=',
    '?',
    '\\',
    '\x7F',
    '{',
    '[',
    ']',
    '^');
```

But Iceberg follows Java URLEncoder.encode. Letters (a-z, A-Z), digits (0-9) and certain punctuation (-, _, ., *) are left unchanged, everything else will be percent-encoded.

A typical example is `+` .  See this test data `{"plus+test", "plus%2Btest"},`.

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
mbasmanova: I see. Is this the only difference? Wondering if there is a way to account for this difference w/o duplicating so much code.

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
mbasmanova: Got it. Also wondering if there is a way to account for this difference w/o duplicating so much code.

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
PingLiuPing: Yes, this is the only difference.
Refactored and just use base class template function.

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.cpp
PingLiuPing: @mbasmanova  Thank you for the comment.
I refactored this by adding an extra std::function parameter to `FileUtils::makePartName`.

velox/dwio/catalog/fbhive/FileUtils.h
```diff
@@ -29,6 +29,10 @@ namespace fbhive {
 
 class FileUtils {
  public:
+  /// Function type for encoding partition key/value strings.
+  /// Takes a string to encode and returns the encoded string.
+  using EncodeFunction = std::function<std::string(const std::string&)>;
```
mbasmanova, line 34: Let's extract FileUtils changes into a separate PR and add tests.

velox/connectors/hive/PartitionNameGenerator.h
```diff
@@ -61,17 +68,57 @@ class HivePartitionUtil {
   virtual std::string toPartitionString(Timestamp value, const TypePtr& type)
       const;
 
+  virtual std::string partitionName(
```
mbasmanova: Do we need both this method (partitionName) and extractPartitionKeyValues here?

velox/connectors/hive/PartitionNameGenerator.h
```diff
@@ -61,17 +68,57 @@ class HivePartitionUtil {
...
   /// @return A vector of (column_name, formatted_value) pairs.
-  static std::vector<std::pair<std::string, std::string>>
+  virtual std::vector<std::pair<std::string, std::string>>
   extractPartitionKeyValues(
       const RowVectorPtr& partitionsVector,
-      vector_size_t row);
-};
+      vector_size_t row) const;
+
+  /// Returns the Hive string representation for null partition values.
```
mbasmanova: Can we replace getNullPartitionValue virtual method with a 'nullValue' string provided to the ctor?

velox/connectors/hive/iceberg/IcebergPartitionNameGenerator.h
```diff
@@ -0,0 +1,136 @@
...
+  /// @param partitionValues RowVector containing transformed partition keys
+  /// for all partitions. Each row represents one unique partition, with
+  /// columns corresponding to partition fields in partitionSpec. Row at
+  /// partitionId contains the keys for this specific partition.
+  /// @return URL-encoded partition path string suitable for use in file paths.
+  std::string partitionName(
+      uint32_t partitionId,
+      const RowVectorPtr& partitionValues) const override;
+
+  class IcebergPartitionValueFormatter : public PartitionNameGenerator {
```
mbasmanova: It is confusing to have IcebergPartitionNameGenerator class extend PartitionNameGenerator and contain IcebergPartitionValueFormatter class which also extends PartitionNameGenerator.

PingLiuPing: Thank you for the comment.
I've been struggled when implementing this.
I want to re-use some of the `toPartitionString` method from `PartitionNameGenerator`.  But the implementations of toPartitionString in IcebergPartitionNameGenerator depends on TransformType. The TransformType is a concept of Iceberg, so we should not reference this from Hive directory (e.g., add TransformType as parameter of toPartitionString).  On the other hand transformType can only be known from `IcebergPartitionSepc`. And it is also strange to add `TransformType` and `IcebergPartitionSepc` both as data members of `IcebergPartitionNameGenerator`.

Let me see if this can be resolved by using your proposal in following.

PingLiuPing:
> Does this make sense?

@mbasmanova  Thank you for the proposal. Let me explore this and see if it is better approach.

velox/dwio/catalog/fbhive/FileUtils.h
PingLiuPing: @mbasmanova  Sure, see https://github.com/facebookincubator/velox/pull/15480

velox/connectors/hive/PartitionNameGenerator.h
PingLiuPing: Thanks.
Not necessarily. Let me refactor further.

PingLiuPing:
> > Does this make sense?
> 
> @mbasmanova Thank you for the proposal. Let me explore this and see if it is better approach.

@mbasmanova, thank you for the proposal.

I update the code and basically follow the idea you proposed.

In HiveDataSink, adding a new method to get the partition name, this method calls `HivePartitionName`
```
std::optional<std::string> getPartitionName(
      uint32_t partitionId)
```
And IcebergDataSink can override this to call `IcebergPartitionName`.

In class `IcebergPartitionName`, it re-uses the static method from `HivePartitionName` and just specialize 3 data types for special handing. And it also re-use the template `partitionKeyValues` in `HivePartitionName`.
 

The CI should fail. I will rebase the code once #15477 and #15480 merged.

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+#pragma once
+
+#include <memory>
+#include "velox/vector/ComplexVector.h"
+#include "velox/vector/SimpleVector.h"
+
+namespace facebook::velox::connector::hive {
+
+FOLLY_ALWAYS_INLINE std::string formatDate(int32_t value) {
+  return DATE()->toString(value);
```
mbasmanova: toString() APIs do not provide any guarantee about the format of the output, hence, should only be used for debugging / logging. Should this use something like toIso8601() ?

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+}
+
+/// Converting partition values to their string representations.
+/// Provides virtual methods for formatting different data types
```
mbasmanova: > Provides virtual methods

Is this correct?

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+/// according to Hive partitioning conventions.
+class HivePartitionName {
+ public:
+  /// Generic template for formatting simple types that just need string
```
mbasmanova: > simple types that just need string conversion.

What are these types? Asking, because it is confusing to see simple bool and integers special cased.

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  /// conversion. Specialized/overloaded for types that need special handling.
+  template <typename T>
+  FOLLY_ALWAYS_INLINE static std::string toName(T value, const TypePtr& type) {
+    return fmt::to_string(value);
+  }
+
+  FOLLY_ALWAYS_INLINE static std::string toName(
+      bool value,
+      const TypePtr& type) {
+    return value ? "true" : "false";
```
mbasmanova: Isn't fmt::to_string produce the same?

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  static std::string toName(int64_t value, const TypePtr& type);
+
+  static std::string toName(int128_t value, const TypePtr& type);
+
+  static std::string toName(StringView value, const TypePtr& type) {
+    return fmt::to_string(value);
```
mbasmanova: Why do we need this specialization?

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  FOLLY_ALWAYS_INLINE static std::string toName(
+      bool value,
+      const TypePtr& type) {
+    return value ? "true" : "false";
+  }
+
+  static std::string toName(int64_t value, const TypePtr& type);
```
mbasmanova, line 44: Please, document to explain why a simple integer needs specialization.

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  static std::string toName(int32_t value, const TypePtr& type);
+
+  static std::string toName(Timestamp value, const TypePtr& type);
+
+  /// Build partition key-value pairs from partition values.
+  /// Returns a vector of (key, value) pairs for all partition columns.
+  /// @tparam Fn A callable that converts a value to a partition
+  ///         string. Takes (value, type, columnIndex) and returns string.
```
mbasmanova: Why do we need to pass columnIndex to this callable?

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+    return fmt::to_string(value);
+  }
+
+  static std::string toName(int32_t value, const TypePtr& type);
+
+  static std::string toName(Timestamp value, const TypePtr& type);
+
+  /// Build partition key-value pairs from partition values.
+  /// Returns a vector of (key, value) pairs for all partition columns.
+  /// @tparam Fn A callable that converts a value to a partition
```
mbasmanova: Fn -> F or Func

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  ///         string. Takes (value, type, columnIndex) and returns string.
+  /// @param partitionId The partition ID (row index) to extract values from.
+  /// @param partitionValues RowVector containing partition values.
+  /// @param nullValueString The string to use for null values.
+  /// @param toPartitionName Callable to convert the value to a string.
```
mbasmanova: typo: "the value" -> "a value"

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+  /// @return A formatted partition directory name string. Null values are
+  /// represented as
+  ///         __HIVE_DEFAULT_PARTITION__.
+  static std::string partitionName(
+      uint32_t partitionId,
+      const RowVectorPtr& partitionValues,
+      bool partitionKeyAsLowerCase);
+};
+
+namespace {
```
mbasmanova: do not use anonymous namespace in header file; use 'detail' namespace.

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+// @param row The row index to extract the value from.
+// @param name The partition column name.
+// @param type The type of the partition column.
+// @param columnIndex The column index in the partition values.
+// @param nullValue The string to use for null values ("" for Hive, "null" for
+//        Iceberg).
+// @param toPartitionName Callable to convert the value to a string.
+// @return A pair of (column_name, formatted_value).
+template <TypeKind Kind, typename Fn>
+std::pair<std::string, std::string> makePartitionKeyValueString(
```
mbasmanova: This function can be simplified to return the value string and the caller can use that to make a pair with the name.

velox/connectors/hive/HivePartitionName.h
```diff
@@ -0,0 +1,177 @@
...
+    const BaseVector& partitionVector,
+    vector_size_t row,
+    const std::string& name,
+    const TypePtr& type,
+    int columnIndex,
+    const std::string& nullValue,
+    const Fn& toPartitionName) {
+  using T = typename TypeTraits<Kind>::NativeType;
+  if (partitionVector.as<SimpleVector<T>>()->isNullAt(row)) {
```
mbasmanova: This check can happen before calling the template. Doing this will help simplify this function further by removing 'nullValue' parameter.

velox/connectors/hive/iceberg/IcebergPartitionName.h
```diff
@@ -0,0 +1,114 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/connectors/hive/HivePartitionName.h"
```
mbasmanova, line 19: This include shouldn't be needed. Is it?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -50,14 +53,102 @@ IcebergDataSink::IcebergDataSink(
...
     const std::shared_ptr<const HiveConfig>& hiveConfig)
-    : HiveDataSink(
+    : IcebergDataSink(
           std::move(inputType),
           insertTableHandle,
           connectorQueryCtx,
           commitStrategy,
           hiveConfig,
+          [&insertTableHandle]() {
+            const auto& inputColumns = insertTableHandle->inputColumns();
```
mbasmanova: would you extract this logic into a helper method to comment what is does?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -50,14 +53,102 @@ IcebergDataSink::IcebergDataSink(
...
+                if (auto it = partitionKeyMap.find(field.name);
+                    it != partitionKeyMap.end()) {
+                  channels.push_back(it->second);
+                }
+              }
+            }
+            return channels;
+          }(),
+          [&insertTableHandle]() -> RowTypePtr {
+            const auto& partitionSpec = insertTableHandle->partitionSpec();
```
mbasmanova: ditto

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -91,4 +183,33 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
...
+      nullptr,
+      input->size(),
+      std::move(transformedColumns));
+  partitionIdGenerator_->run(transformedRowVector, partitionIds_);
+}
+
+std::optional<std::string> IcebergDataSink::getPartitionName(
+    uint32_t partitionId) const {
+  std::optional<std::string> partitionName;
+  if (icebergPartitionName_) {
```
mbasmanova: Should this be a check? Is it valid to call getPartitionName for unpartitioned table?

mbasmanova (commented):
@PingLiuPing Thank you for iterating. This is shaping up quite nicely. I have a few minor comments.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks. `DateType::toString` directly calls `DateType::toIso8601(days);`, seems the result should be good. But I get your point and let me change it to `DateType::toIso8601(days)` for non debugging code.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks. Forget to update this comment.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks for the comment.
columnIndex is not used for HivePartitionName, it will be used in IcebergPartitionName to locate the transform type of each partition column since Iceberg requires such information to format the partition value.
Needs to add it here as an parameter of template function.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/IcebergPartitionName.h
PingLiuPing: Thanks, it should be kept, I call `HivePartitionName::toName(value, type)` in the default template method `toName`

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Yes, it should a check. 
getPartitionName is been protected by `isPartitioned()`. So it is invalid to call getPartitionName for unpartitioned table.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Sure.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks,  should be removed after refactor from previous implementation.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks. You are right.

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks. refined

velox/connectors/hive/HivePartitionName.h
PingLiuPing: Thanks. Yes you are right.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Thanks, ok.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Ok.

PingLiuPing:
> @PingLiuPing Thank you for iterating. This is shaping up quite nicely. I have a few minor comments.

@mbasmanova Thank you very much for reviewing. Just addressed your comment,  please take another look whenever you get a chance.

velox/connectors/hive/HiveDataSink.h
```diff
@@ -530,7 +531,9 @@ class HiveDataSink : public DataSink {
       CommitStrategy commitStrategy,
       const std::shared_ptr<const HiveConfig>& hiveConfig,
       uint32_t bucketCount,
-      std::unique_ptr<core::PartitionFunction> bucketFunction);
+      std::unique_ptr<core::PartitionFunction> bucketFunction,
+      const std::vector<column_index_t>& partitionChannels,
```
mbasmanova, line 569: Would you document this ctor? Why do we have 2 ctors? Do we need both?

velox/connectors/hive/HiveDataSink.cpp
```diff
@@ -864,6 +868,18 @@ uint32_t HiveDataSink::appendWriter(const HiveWriterId& id) {
   return writerIndexMap_[id];
 }
 
+std::optional<std::string> HiveDataSink::getPartitionName(
+    uint32_t partitionId) const {
+  std::optional<std::string> partitionName;
+  if (isPartitioned()) {
```
mbasmanova: Should this be a VELOX_CHECK?

velox/connectors/hive/iceberg/IcebergDataSink.h
```diff
@@ -114,6 +116,61 @@ class IcebergDataSink : public HiveDataSink {
   /// @return Vector of JSON strings, one per writer, formatted according to
   /// Presto and Spark Iceberg commit protocol.
   std::vector<std::string> commitMessage() const override;
+
+ private:
+  IcebergDataSink(
```
mbasmanova, line 121: Why do we need this private ctor?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -15,10 +15,16 @@
...
 
 #include "velox/connectors/hive/iceberg/IcebergDataSink.h"
+
 #include "velox/common/base/Fs.h"
+#include "velox/connectors/hive/PartitionIdGenerator.h"
+#include "velox/connectors/hive/iceberg/TransformExprBuilder.h"
 
 namespace facebook::velox::connector::hive::iceberg {
 
+static constexpr std::string_view kDefaultIcebergFunctionPrefix{
```
mbasmanova, line 25: Is this hard-coded in multiple places?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -103,4 +245,33 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
...
+      nullptr,
+      input->size(),
+      std::move(transformedColumns));
+  partitionIdGenerator_->run(transformedRowVector, partitionIds_);
+}
+
+std::optional<std::string> IcebergDataSink::getPartitionName(
+    uint32_t partitionId) const {
+  VELOX_CHECK_NOT_NULL(icebergPartitionName_);
+  std::optional<std::string> partitionName =
```
mbasmanova: nit: partitionName variable is not needed; just return icebergPartitionName_->partitionNam...

velox/connectors/hive/tests/HivePartitionNameTest.cpp
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#include "velox/connectors/hive/HivePartitionUtil.h"
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/connectors/hive/HivePartitionName.h"
```
mbasmanova, line 17: Would you rename the file to HivePartitionNameTest?

velox/connectors/hive/tests/HivePartitionUtilTest.cpp
```diff
@@ -102,8 +108,11 @@ TEST_F(HivePartitionUtilTest, partitionName) {
 
       EXPECT_EQ(
           FileUtils::makePartName(
-              HivePartitionUtil::extractPartitionKeyValues(
-                  makePartitionsVector(input, partitionChannels), 0),
+              HivePartitionName::partitionKeyValues(
+                  0,
+                  makePartitionsVector(input, partitionChannels),
+                  /*nullValueString=*/"",
+                  toPartitionName()),
```
mbasmanova: this code repeats, perhaps, add a helper function partitionKeyValues(0, makePartitionsVector(input, partitionChannels))

velox/connectors/hive/iceberg/tests/PartitionNameTest.cpp
```diff
@@ -0,0 +1,480 @@
...
+#include "velox/connectors/hive/iceberg/TransformExprBuilder.h"
+#include "velox/connectors/hive/iceberg/tests/IcebergTestBase.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+using namespace facebook::velox;
+
+namespace {
+
+class PartitionNameGeneratorTest : public test::IcebergTestBase {
```
mbasmanova: Please, make sure file name matches test class name.

velox/connectors/hive/tests/HivePartitionNameTest.cpp
PingLiuPing: Sure.

velox/connectors/hive/iceberg/IcebergDataSink.cpp

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Thanks.
`kDefaultIcebergFunctionPrefix` ("$internal$.iceberg.") is added for registering those iceberg function in IcebergDataSink. This implies these functions will be registered twice with different prefix, one from upstream engine, one as internal usage.
`kDefaultTestIcebergFunctionNamePrefix` ("$internal$.test_iceberg.")  is added in IcebergTestBase.h and it is used to register those functions for testing purpose.

velox/connectors/hive/HiveDataSink.h
PingLiuPing: > Would you document this ctor?

Sure.

> Why do we have 2 ctors?

I think the first one is used as a simpler public API, it delegates the actual implementation to second constructor.
With this pattern, subclasses like `IcebergDataSink` can call the full constructor directly with custom initialisation logic while still re-use the base class initialisation. Probably make sense to keep them both.

velox/connectors/hive/HiveDataSink.cpp
PingLiuPing: Yes.

velox/connectors/hive/iceberg/IcebergDataSink.h
PingLiuPing: Thanks.

This is used to simplify the initialization list a bit, some data members initialization depends on the input. We can prepare the input in the public constructor and then do the initialization in this private constructor.

velox/connectors/hive/iceberg/tests/PartitionNameTest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -103,4 +245,31 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
...
+  const auto& transformedRowVector = std::make_shared<RowVector>(
+      connectorQueryCtx_->memoryPool(),
+      partitionRowType_,
+      nullptr,
+      input->size(),
+      std::move(transformedColumns));
+  partitionIdGenerator_->run(transformedRowVector, partitionIds_);
+}
+
+std::optional<std::string> IcebergDataSink::getPartitionName(
```
mbasmanova: Can we change the output type to std::string? The output is never std::nullopt, is it?

PingLiuPing: Thanks. It can be changed to std::string.

PR #15464, opened 2025-11-11
feat: Add numOperatorSplits runtime statistic

arhimondr:
Summary: To understand how splits are distributed across drivers

Differential Revision: D86765597


velox/exec/TableScan.cpp
```diff
@@ -512,20 +513,14 @@ void TableScan::close() {
     dataSource_->cancel();
   }
 
-  if (scaledController_ == nullptr) {
-    return;
-  }
+  stats_.wlock()->addRuntimeStat(
```
xiaoxmeng, line 516: Can we just report numSplits in stats_? They track the same sth? Thanks!

PR #15466, opened 2025-11-11
fix(ci): Replace macOS Intel build from CI

czentgr:
The macOS 13 images for github runners are
deprecated. On running with
macOS 15 and intel based images the CI pipeline
are very slow. Instead, we are building on arm for
debug and release.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -49,18 +49,16 @@ concurrency:
 jobs:
   macos-build:
     if: ${{ github.repository == 'facebookincubator/velox' }}
-    name: ${{ matrix.os }}
+    name: macos-15-${{ matrix.type }}
     strategy:
       fail-fast: false
       matrix:
-        # macos-13 = x86_64 Mac
-        # macos-15 = arm64 Mac and cmake 4.0
-        os: [macos-13, macos-15]
-    runs-on: ${{ matrix.os }}
+        # macos-15 = arm64 Mac and cmake 4.0 with 7GB RAM
+        type: [debug, release]
+    runs-on: macos-15
     env:
       CCACHE_DIR: ${{ github.workspace }}/ccache
-      # The arm runners have only 7GB RAM
-      BUILD_TYPE: ${{ matrix.os == 'macos-15' && 'Release' || 'Debug' }}
+      BUILD_TYPE: ${{ matrix.type }}
       INSTALL_PREFIX: /tmp/deps-install
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


czentgr:
@kgpai @majetideepak @assignUser The build for this takes really long. This is going to be the last supported intel version. it is so slow because it must be running with an emulator or something. It looks like it'll be like 3+h.
Maybe we should just stop supporting this? With ccache it should be better once available but still something to think about.

assignUser:
+1 for dropping Intel mac

kgpai:
@czentgr Yes sounds fair to me.

czentgr:
I've got a clean run. We can keep some of the changes because the code in some cases is technically not correct (e.g. the MOCK method fixes). 
The time is 176 mins arm mac (2h56 with ccache) vs 347 mins (5h47mins without ccache) intel mac. I think that is really close to getting killed by github as well (I've seen runners being killed with timeout). With the consensus I will change the PR to add some fixes identified by the compiler and remove the intel build otherwise.

czentgr:
@assignUser The clang-tidy failure here shows the fix for clang-tidy (I have a separate PR) and highlights the issue with protocol generated files if they happen to be changed.

czentgr:
I split this into a debug - release matrix instead. The debug build on arm is not slower than the release build apparently.
Initial run: both about 200mins. See https://github.com/facebookincubator/velox/actions/runs/19542428807/job/55952142955
Subsequent run with ccache (and no code changes) results in 147min debug and 197 min release build (see above). It did have a ccache it downloaded.

assignUser:
Yeah 6h is the limit for workflows. Wild, must really be emulation. Will have to take this info to arrow too...

Thanks for investigating @czentgr !

czentgr:
Silently, the macOS 13 runners have come back. I didn't find any comment on that. I suspect there were complaints about the performance of the macos15-intel runners.
We probably should still replace the intel run with the arm run and build debug and release. Eventually, these runners will go. There shouldn't be too many intel macs out there.

kgpai:
@czentgr Agree, lets remove the intel runners (since I dont see Apple selling these anymore - correct me if wrong).

velox/type/tests/TypeTest.cpp
```diff
@@ -678,10 +678,16 @@ class OpaqueWithMetadataType : public OpaqueType {
       : OpaqueType(std::type_index(typeid(OpaqueWithMetadata))),
         metadata(metadata) {}
 
+  bool operator==(const OpaqueWithMetadataType& other) const {
```
majetideepak: Is this related to this change?

czentgr: This was a change to make it work on intel macOS 15.
This is the better code imho and the issue was valid.

PR #15474, opened 2025-11-12
fix: Build failure caused by Add support for casting TIME to TimeWithTimezone

amberMZ:
Summary: Fix build failure caused by D84954694

Differential Revision: D86821163


han-yan01 (approved):

han-yan01:
closed due to root cause diff revert
thanks @amberMZ for taking a look at this!

PR #15477, opened 2025-11-12
feat: Add support for evaluating Iceberg partition transforms

PingLiuPing:
Introduce infrastructure for evaluating Iceberg partition transforms.
- TransformExprBuilder converts Iceberg partition specifications into Velox expressions
- TransformEvaluator evaluates multiple transform expressions in a single pass using compiled ExprSet.

### DIFF
--- a/velox/connectors/hive/iceberg/PartitionSpec.h
+++ b/velox/connectors/hive/iceberg/PartitionSpec.h
@@ -20,35 +20,6 @@
 
 namespace facebook::velox::connector::hive::iceberg {
 
-/// Prefix for registering Iceberg partition transform functions in Velox's
-/// function registry. Should follow the format "catalog.schema.function_name"
-/// to support namespace isolation and avoid issues such as
-/// https://github.com/prestodb/presto/pull/26584.
-/// The "iceberg.default." prefix is used to:
-/// - Clearly identify these as Iceberg-specific transform functions (bucket,
-///   truncate, years, months, days, hours).
-/// - Separate them from other function namespaces (e.g., Presto, Spark).
-/// - Use "default" as the schema name since Iceberg transforms are not tied to
-///   a specific schema but are part of the Iceberg specification.
-inline constexpr char const* kIcebergFunctionPrefix{"iceberg.default."};
-
-inline const std::string kBucketFunction =
-    std::string(kIcebergFunctionPrefix) + "bucket";
-inline const std::string kTruncateFunction =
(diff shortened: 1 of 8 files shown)


PingLiuPing:
@mbasmanova  PR https://github.com/facebookincubator/velox/pull/15440 been merged and then backed out due to the registered function name does not follow Prestissimo naming conventions i.e. catalog.schema.function_name. And it can cause Prestissimo worker node crash, see https://github.com/prestodb/presto/pull/26584#issuecomment-3518527017.

I updated the prefix to "iceberg.default.".
Tested this PR with Prestissimo. Works fine now.

Could you help have a look? Sorry for the trouble.

velox/connectors/hive/iceberg/PartitionSpec.h
```diff
@@ -20,6 +20,25 @@
 
 namespace facebook::velox::connector::hive::iceberg {
 
+inline constexpr char const* kIcebergFunctionPrefix{"iceberg.default."};
```
mbasmanova: >  the registered function name does not follow Prestissimo naming conventions i.e. catalog.schema.function_name

Where is this requirement documented and how is it tested at PR time? CC: @amitkdutta @tanjialiang @aditi-pandit 

Iceberg functions are not Prestissimo-specific... what will we do if different engines have conflicting requirements?

Let's add a comment to explain the choice of naming and add a link to a GitHub issue that provides more context. Let's also follow-up on the above questions.

PingLiuPing: @mbasmanova  The code to check this is https://github.com/prestodb/presto/blob/da5d5fcf3174604f35d85b96ceabad2f3eee2334/presto-native-execution/presto_cpp/main/functions/FunctionMetadata.cpp#L45

PingLiuPing: > Let's add a comment to explain the choice of naming and add a link to a GitHub issue that provides more context. Let's also follow-up on the above questions.

Thanks, added comments to explain the naming choice.

PingLiuPing: @mbasmanova  It seems I should remove the function registration out of Velox and let the upstream engines to register them during startup. This follow the same pattern with other functions as well.

But iceberg has some speciality since it needs the function name when evaluating the transformation. Wondering how we can get the prefix within Velox to build the function name.

PR #15478, opened 2025-11-12
feat(cuDF): Check signatures of functions in Expr before replacing FilterProject with CudfFilterProject

devavret:
This PR adds function signatures to cuDF functions and checks that they match the ones in velox before replacing FilterProject operator with CudfFilterProject.

This PR also reverts to using exec::Expr to evaluate replaceablility of FilterProject in driver adapter. This was previously changed in #14956 to use TypedExpr. This is now safe since the leaf pool used for memory allocations is no longer taken from the query pool but directly from the top level memory pool


velox/experimental/cudf/expression/AstExpression.cpp
```diff
@@ -157,20 +160,186 @@ const std::unordered_map<std::string, Op> binaryOps = [] {
...
+      return isOpAndInputsSupported(Op::CAST_TO_INT64, inputCudfDataTypes);
+    }
+    if (outputKind == TypeKind::DOUBLE) {
+      return isOpAndInputsSupported(Op::CAST_TO_FLOAT64, inputCudfDataTypes);
+    }
+    return false;
+  }
+
+  // Switch/If: only specific patterns supported in pure AST
+  if ((name == "switch" || name == "if") && len == 3) {
```
jinchengchenghh: May remove the switch branch, because the result is incorrect if the source column contains null, the disabled unit test case has confirmed that.

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -398,6 +398,7 @@ CudfHashJoinProbe::CudfHashJoinProbe(
     // create ast tree
     std::vector<PrecomputeInstruction> rightPrecomputeInstructions;
     std::vector<PrecomputeInstruction> leftPrecomputeInstructions;
+    static constexpr bool kAllowPureAstOnly = true;
```
jinchengchenghh, line 402: Why set it to true here? In what condition, do we only use ast tree?

velox/experimental/cudf/expression/AstExpression.cpp
```diff
@@ -157,20 +160,186 @@ const std::unordered_map<std::string, Op> binaryOps = [] {
...
+  if (name == "literal") {
+    auto type = expr->type();
+    return isSupportedLiteral(type);
+  }
+  if (auto fieldExpr = std::dynamic_pointer_cast<FieldReference>(expr)) {
+    const auto fieldName =
+        fieldExpr->inputs().empty() ? name : fieldExpr->inputs()[0]->name();
+    if (fieldExpr->field() == fieldName) {
+      return true;
+      // } else if (!allowPureAstOnly ) {
```
jinchengchenghh: remove?

velox/experimental/cudf/exec/CudfHashJoin.cpp
devavret: This is set to true in CudfHashJoin because for mixed joins (equi join + conditions) the cudf api only accepts a cudf AST. For FilterProject, this remains false

devavret:
> Looks like this validation is under AST tree, now it does not consider `registerFunction` path, am I right?

It does consider registerFunction path. Previously, we checked only function names and not signature before deciding to replace. This could cause issues like in AST implementation of switch/if since it only handled certain inputs. The signatures are also added for cudf functions used in FunctionExpression. https://github.com/facebookincubator/velox/pull/15478/files#diff-fc50bf5b0285fc08f7c5c8e47cec8a6264a62cb0ae2f3420284172382e44568dR796-R1012

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -415,7 +417,8 @@ CudfHashJoinProbe::CudfHashJoinProbe(
           probeType,
           buildType,
           leftPrecomputeInstructions,
-          rightPrecomputeInstructions);
+          rightPrecomputeInstructions,
+          kAllowPureAstOnly);
```
jinchengchenghh, line 422: After this PR https://github.com/facebookincubator/velox/pull/15337/files, only some join type still uses tree_, maybe optimize it later. CC @shrshi

velox/experimental/cudf/tests/FilterProjectTest.cpp
```diff
@@ -863,7 +901,13 @@ TEST_F(CudfFilterProjectTest, mixedLiteralProjection) {
...
 }
 
-TEST_F(CudfFilterProjectTest, dereference) {
+// This test checks for CudfExpression's ability to handle nested field
+// references. However, to test this, we need to disable CPU fallback, otherwise
+// the test could pass by using CPU, having not exercised the CudfExpression at
+// all. But this test relies on row_constructor to construct the nested fields,
+// which CudfExpression doesn't support. Disabling this until we have a better
+// test
+TEST_F(CudfFilterProjectTest, DISABLED_dereference) {
```
jinchengchenghh, line 946: I remember we support row_constructor in cudf CC @karthikeyann

devavret: I can see that AST supported "nested_column" but I don't see "row_constructor" in the old commits. https://github.com/facebookincubator/velox/blob/b9bc9fd9647e3b9cd192464ca1f5a18a5896854d/velox/experimental/cudf/exec/ExpressionEvaluator.cpp#L732-L733

jinchengchenghh: Yes, this is to support row_constructor https://github.com/rapidsai/velox/pull/31/files

devavret: I remember testing that and found that first project (row_constructor) falls back on cpu but the nested access can happen on gpu. IIUC, this test was to check that c1_c2.c1 can be processed on gpu. But if we allowed fallback, then both could run on cpu and test would pass. That's why I disabled it

jinchengchenghh (approved):

karthikeyann:
@xiaoxmeng could you check these build errors? 
They are unrelated to the change in this PR.

PR #15480, opened 2025-11-12
refactor: FileUtils::makePartName

PingLiuPing:
Refactor FileUtils::makePartName to support Iceberg partition name generation.
1. Add 'useDefaultPartitionValue' parameter to control whether
   empty partition values should be replaced with `__HIVE_DEFAULT_PARTITION__`
2. Add 'encodeFunc' parameter to allow custom encoding functions
   (defaults to escapePathName for Hive compatibility).
3. Add test coverage.

### DIFF
--- a/velox/dwio/catalog/fbhive/FileUtils.cpp
+++ b/velox/dwio/catalog/fbhive/FileUtils.cpp
@@ -158,46 +158,28 @@ std::string FileUtils::unescapePathName(const std::string& data) {
 
 std::string FileUtils::makePartName(
     const std::vector<std::pair<std::string, std::string>>& entries,
-    bool partitionPathAsLowerCase) {
-  size_t size = 0;
-  size_t escapeCount = 0;
-  std::for_each(entries.begin(), entries.end(), [&](auto& pair) {
-    auto keySize = pair.first.size();
-    VELOX_CHECK_GT(keySize, 0);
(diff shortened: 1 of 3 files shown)


velox/dwio/catalog/fbhive/FileUtils.cpp
```diff
@@ -22,6 +22,7 @@
 #include <folly/container/Array.h>
 
 #include "velox/common/base/Exceptions.h"
+#include "velox/functions/prestosql/URLFunctions.h"
```
mbasmanova: We don't want to create a dependency from dwio to presto functions. If we need to reuse this logic, let's move it somewhere under velox/common.

mbasmanova: Alternatively, FileUtils::escapeIcebergPathName logic can go into Iceberg connector. Perhaps, no need to have this exposed as a reusable function.

PingLiuPing: > Alternatively, FileUtils::escapeIcebergPathName logic can go into Iceberg connector. Perhaps, no need to have this exposed as a reusable function.

Yes, this is a better option. But if go with this option, I don't need to add this method into this PR. Just need to change the makePartName interface.

PR #15483, opened 2025-11-12
feat: Support VeloxQueryRunner in Github Expression Fuzzer (#15483)

### LINKED ISSUE #15414: Support VeloxQueryRunner in ExpressionFuzzer
### Description

Recently, a new reference query runner has been added to the fuzzer exec suite: VeloxQueryRunner. This runner sends and receives serialized query plans and RowVector output from a new service, LocalRunnerService, with the hope of comparing expression evaluation across different builds of Velox. This new runner comes with several benefits including apple-to-apple comparisons where we can directly compare Velox expression evaluation with itself, eliminating unavoidable cosmetic differences between expression evaluation and query execution (in fact we get to effectively reclaim the entire skip list from my testing). Additionally, it's much caster, roughly twice as fast as executing with Presto since we can leverage batch serialization and expression evaluation. Ultimately, this will allow us to deprecate Presto SOT fuzzers and use VeloxQueryRunner as our regression fuzzer test (although that is a discussion for a different thread). This will require several changes:
1. Currently LocalRunnerService depends on an internal build of Thrift. We will need to potentially refactor (more likely just use an external build) of FBThrift.
2. ExpressionFuzzerTest will need to be extended to support this new fuzzer. No work is required from the ExpressionVerifier, we can simply hook it up directly to main as we do PrestoQueryRunner. We should enable it using a flag `local_runner_url` similar to PrestoQueryRunner usage to easily manipulate execution using runtime flags.
3. We will need to select a release version of Velox, build it, and store the binary in some artifacts repo to download during PR fuzzer job setup.

peterenescu:
Summary:

Additionally resolves https://github.com/facebookincubator/velox/issues/15414

Add Regression Fuzzer (i.e. Expression Fuzzer with LocalRunnerService as a source-of-truth) to Fuzzer Job test suite. This fuzzer sends serialized plans to a thrift service build on the base commit to test solely on new changes.

Differential Revision: D86911550

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -222,6 +222,7 @@ jobs:
             -DVELOX_MONO_LIBRARY=ON
             -DVELOX_BUILD_PYTHON_PACKAGE=ON
             -DVELOX_PYTHON_LEGACY_ONLY=ON
+            -DVELOX_ENABLE_LOCAL_RUNNER_SERVICE=ON
             ${{ inputs.extraCMakeFlags }}
         run: |
           make python-venv
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -384,6 +384,13 @@ jobs:
           path: velox/_build/debug/velox/exec/fuzzer/velox_spatial_join_fuzzer
           retention-days: ${{ env.RETENTION }}
 
+      - name: Upload local runner service runner
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: local_runner_service_runner
+          path: velox/_build/debug/velox/exec/fuzzer/velox_local_runner_service_runner
+          retention-days: ${{ env.RETENTION }}
+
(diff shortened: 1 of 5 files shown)


.github/workflows/linux-build-base.yml
```diff
@@ -102,6 +102,7 @@ jobs:
             "-DVELOX_ENABLE_WAVE=ON"
             "-DVELOX_MONO_LIBRARY=ON"
             "-DVELOX_BUILD_SHARED=ON"
+            "-DVELOX_ENABLE_REMOTE_FUNCTIONS=ON"
```
kgpai: Lets not enable remote functions . This doesnt capture the root of the problem which is that fbthrift is a required dependency for fuzzers now.

kgpai (changes requested):

.github/workflows/linux-build-base.yml
peterenescu: Sounds good, I'll probably separate the logic in this PR to 1. update the build and then 2. to add a fuzzer job. Will follow up

PR #15488, opened 2025-11-13
feat: Add rewrite for IN special form

pramodsatya:
Rewrites IN special form when the `value` being searched for is constant. 

- Returns `true` if `value` is in the `IN-list`. 
- Returns NULL if the value is NULL. 
- Deduplicates NULLs from the `IN-list`. 
- Removes constant elements from `IN-list` that do not match with `value`.

Examples:

- 5 IN (a, 5) ==> true
- NULL IN (5, a, b) ==> NULL
- 5 IN (NULL, a, NULL) ==> 5 IN (a, NULL)
- 5 IN (1, 2, a, b) ==> 5 IN (a, b)

### DIFF
--- a/velox/functions/prestosql/InRewrite.cpp
+++ b/velox/functions/prestosql/InRewrite.cpp
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <set>
+
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/expression/ExprUtils.h"
+#include "velox/functions/prestosql/InRewrite.h"
+
+namespace facebook::velox::functions {
+
+core::TypedExprPtr InRewrite::rewrite(const core::TypedExprPtr& expr) {
+  // IN expression with a constant value can be rewritten.
+  if (expr->inputs().empty() || !expr->inputs()[0]->isConstantKind()) {
+    return nullptr;
+  }
+
+  if (!expr->isCallKind()) {
+    return nullptr;
+  }
+  static const char* kIn = "in";
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  if (callExpr->name() != kIn) {
+    return nullptr;
+  }
+
+  const auto& valueExpr = callExpr->inputs().at(0);
+  const auto* constValueExpr =
+      valueExpr->asUnchecked<core::ConstantTypedExpr>();
+  if (constValueExpr->isNull()) {
+    return core::ConstantTypedExpr::makeNull(BOOLEAN());
+  }
+
+  const auto& inputs = callExpr->inputs();
+  const auto numInputs = inputs.size();
+  bool hasNullInput = false;
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  optimizedInputs.emplace_back(valueExpr);
+
+  for (auto i = 1; i < numInputs; i++) {
+    const auto& input = inputs[i];
+    if (input->isConstantKind()) {
+      const auto* constInput = input->asUnchecked<core::ConstantTypedExpr>();
+      if (constInput->isNull()) {
+        if (!hasNullInput) {
+          optimizedInputs.push_back(input);
+          hasNullInput = true;
+        }
+      } else if (*constInput == *constValueExpr) {
+        return std::make_shared<core::ConstantTypedExpr>(BOOLEAN(), true);
+      }
+    } else {
+      optimizedInputs.push_back(input);
+    }
+  }
+
+  VELOX_CHECK(!optimizedInputs.empty(), "No inputs to IN after rewrite.");
+  return std::make_shared<core::CallTypedExpr>(
+      expr->type(), std::move(optimizedInputs), callExpr->name());
+}
+
+void InRewrite::registerRewrite() {
+  expression::ExprRewriteRegistry::instance().registerRewrite(
+      functions::InRewrite::rewrite);
+}
+
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/InRewrite.h
+++ b/velox/functions/prestosql/InRewrite.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::functions {
+
+class InRewrite {
+ public:
+  /// Rewrites IN special form function when the `value` being searched for is
+  /// constant and the IN-list is not constant. Returns `true` if `value` is in
+  /// the IN-list. Returns NULL if the IN-list contains a NULL. Removes constant
+  /// expressions from IN-list that are not equal to `value`.
+  static core::TypedExprPtr rewrite(const core::TypedExprPtr& expr);
+
+  static void registerRewrite();
+};
+
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
@@ -21,6 +21,7 @@
 #include "velox/functions/prestosql/Fail.h"
 #include "velox/functions/prestosql/GreatestLeast.h"
 #include "velox/functions/prestosql/InPredicate.h"
+#include "velox/functions/prestosql/InRewrite.h"
 #include "velox/functions/prestosql/Reduce.h"
 #include "velox/functions/prestosql/types/IPAddressType.h"
 #include "velox/functions/prestosql/types/TimestampWithTimeZoneType.h"
@@ -79,6 +80,8 @@ void registerAllSpecialFormGeneralFunctions() {
       bool,
       Generic<T1>,
       Variadic<Generic<T1>>>({"in"});
+  InRewrite::registerRewrite();
+
   VELOX_REGISTER_VECTOR_FUNCTION(udf_concat_row, expression::kRowConstructor);
   registerIsNullFunction("is_null");
 }
--- a/velox/functions/prestosql/tests/InRewriteTest.cpp
+++ b/velox/functions/prestosql/tests/InRewriteTest.cpp
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/core/Expressions.h"
+#include "velox/expression/ExprRewriteRegistry.h"
+#include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+class InRewriteTest : public functions::test::FunctionBaseTest {
+ protected:
+  static core::TypedExprPtr constant(
+      const TypePtr& type,
+      const Variant& value) {
+    return std::make_shared<core::ConstantTypedExpr>(type, value);
+  }
+
+  static core::TypedExprPtr field(
+      const TypePtr& type,
+      const std::string& name) {
+    return std::make_shared<core::FieldAccessTypedExpr>(type, name);
+  }
+
+  static core::TypedExprPtr makeIn(
+      std::initializer_list<core::TypedExprPtr> inputs) {
+    return std::make_shared<core::CallTypedExpr>(BOOLEAN(), "in", inputs);
+  }
+
+  bool testInRewrite(
+      const core::TypedExprPtr& expr,
+      const core::TypedExprPtr& expected) {
+    const auto rewritten =
+        expression::ExprRewriteRegistry::instance().rewrite(expr);
+    return *rewritten == *expected;
+  }
+};
+
+TEST_F(InRewriteTest, basic) {
+  // 1234 in (2, a, b, 1234) -> true.
+  auto in = makeIn(
+      {constant(INTEGER(), 1234),
+       constant(INTEGER(), 2),
+       field(INTEGER(), "a"),
+       field(INTEGER(), "b"),
+       constant(INTEGER(), 1234)});
+  auto expected = constant(BOOLEAN(), true);
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // 81 in (2, 4, a, b, 9) -> 81 in (a, b)
+  in = makeIn(
+      {constant(INTEGER(), 81),
+       constant(INTEGER(), 2),
+       field(INTEGER(), "a"),
+       constant(INTEGER(), 4),
+       field(INTEGER(), "b"),
+       constant(INTEGER(), 1234)});
+  expected = makeIn(
+      {constant(INTEGER(), 81), field(INTEGER(), "a"), field(INTEGER(), "b")});
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // 5 in (NULL, a, 5) -> true
+  in = makeIn(
+      {constant(BIGINT(), 5LL),
+       constant(BIGINT(), variant::null(TypeKind::BIGINT)),
+       field(BIGINT(), "a"),
+       constant(BIGINT(), 5LL)});
+  expected = constant(BOOLEAN(), true);
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // NULL in (5, NULL, a) -> NULL
+  in = makeIn(
+      {constant(BIGINT(), variant::null(TypeKind::BIGINT)),
+       constant(BIGINT(), 5LL),
+       constant(BIGINT(), variant::null(TypeKind::BIGINT)),
+       field(BIGINT(), "a")});
+  expected = constant(BOOLEAN(), variant::null(TypeKind::BOOLEAN));
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // NULL in ('foo', 'bar', a) -> NULL
+  in = makeIn(
+      {constant(VARCHAR(), variant::null(TypeKind::VARCHAR)),
+       constant(VARCHAR(), "foo"),
+       constant(VARCHAR(), "bar"),
+       field(VARCHAR(), "a")});
+  expected = constant(BOOLEAN(), variant::null(TypeKind::BOOLEAN));
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // 'hello' in ('foo', NULL, 'bar', a, NULL) -> 'hello' in (NULL, a)
+  in = makeIn(
+      {constant(VARCHAR(), "hello"),
+       constant(VARCHAR(), "foo"),
+       constant(VARCHAR(), variant::null(TypeKind::VARCHAR)),
+       constant(VARCHAR(), "bar"),
+       field(VARCHAR(), "a"),
+       constant(VARCHAR(), variant::null(TypeKind::VARCHAR))});
+  expected = makeIn(
+      {constant(VARCHAR(), "hello"),
+       constant(VARCHAR(), variant::null(TypeKind::VARCHAR)),
+       field(VARCHAR(), "a")});
+  ASSERT_TRUE(testInRewrite(in, expected));
+}
+
+} // namespace
+} // namespace facebook::velox::functions
--- a/velox/functions/prestosql/CMakeLists.txt
+++ b/velox/functions/prestosql/CMakeLists.txt
@@ -35,6 +35,7 @@ velox_add_library(
   FindFirst.cpp
   FromUtf8.cpp
   InPredicate.cpp
+  InRewrite.cpp
   JsonFunctions.cpp
   Map.cpp
   MapEntries.cpp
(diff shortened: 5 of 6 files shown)


velox/expression/tests/SpecialFormRewriteTestBase.cpp
```diff
@@ -33,21 +33,29 @@ void SpecialFormRewriteTestBase::TearDown() {
   expression::ExprRewriteRegistry::instance().clear();
 }
 
+void SpecialFormRewriteTestBase::testRewrite(
+    const core::TypedExprPtr& expr,
+    const core::TypedExprPtr& expected,
+    const RowTypePtr& type) {
```
aditi-pandit: type is not used in the function. Please remove the argument.

velox/expression/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+      constant(INTEGER(), 1234));
+  expected = std::make_shared<core::CallTypedExpr>(
+      BOOLEAN(),
+      "in",
+      constant(INTEGER(), 81),
+      field(INTEGER(), "a"),
+      field(INTEGER(), "b"));
+  testRewrite(in, expected, type);
+
+  // 'hello' in ('bar', a, 'hello', 'blah') -> true
```
aditi-pandit: Lets add a test which returns false for the evaluation.

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,68 @@
...
+      const auto* constInput =
+          inputs[i]->asUnchecked<core::ConstantTypedExpr>();
+      if (*constInput == *constValueExpr) {
+        return std::make_shared<core::ConstantTypedExpr>(BOOLEAN(), true);
+      }
+    } else {
+      optimizedInputs.push_back(inputs[i]);
+    }
+  }
+
```
aditi-pandit, line 68: If the In list comprised of only constants none of which matched the value being searched for then we can return false. In that case optimizedInputs would be empty here.

We should add that condition.

aditi-pandit (commented):
@pramodsatya : Looks good overall. Have few comments.

velox/functions/prestosql/InRewrite.cpp
czentgr: +1

velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
```diff
@@ -79,6 +80,8 @@ void registerAllSpecialFormGeneralFunctions() {
       bool,
       Generic<T1>,
       Variadic<Generic<T1>>>({"in"});
```
czentgr, line 82: You can use the added `kIn` constant here now and also in line 77 above.

velox/expression/tests/InRewriteTest.cpp
czentgr: +1

velox/functions/prestosql/InRewrite.cpp
pramodsatya: Thanks for pointing it out, the intent was to add this rewrite only for non constant IN-lists, since the expression will be constant folded by the `ExprOptimizer` when the IN-list is constant. The function for constant IN-list is also different and it is optimized already. 
Updated the checks and added a comment to indicate the same. 

Update: If the rewrite were to evaluate a constant IN expression, with say the `tryEvaluateConstantExpression` API, it would need a queryCtx and a memory pool to perform the evaluation. This breaks the neat abstraction provided by API `ExprRewriteRegistry::rewrite()`, which just takes in a `TypedExpr` and returns a rewritten `TypedExpr`. I feel it is better to retain the current expression optimization logic which delineates constant folding and logical rewrites.

velox/expression/tests/SpecialFormRewriteTestBase.cpp
pramodsatya: type is required by other tests in `ExprOptimizerTest.cpp`, the variant of `testRewrite` that takes in `TypedExpr` was added because Duckdb parser has a limitation that it cannot parse IN expressions with non constant IN-list.

velox/expression/tests/InRewriteTest.cpp
pramodsatya: With this rewrite, `false` cannot be returned since the IN-list is not constant and will have non constant expressions like field references that can only be evaluated at runtime.

pramodsatya (commented):
Thanks for the feedback @aditi-pandit, @czentgr, addressed the comments. Could you please take another look?

mbasmanova:
@pramodsatya What's the status of this PR?

pramodsatya:
@mbasmanova, it is currently pending review. Could you please take a look when you get a chance? This is the last of the special form rewrites we want to add. Thanks.

mbasmanova:
@pramodsatya CI is red. Would you take a look?

velox/expression/ExprConstants.h
```diff
@@ -21,6 +21,7 @@ inline constexpr const char* kAnd = "and";
 inline constexpr const char* kOr = "or";
 inline constexpr const char* kSwitch = "switch";
 inline constexpr const char* kIf = "if";
+inline constexpr const char* kIn = "in";
```
mbasmanova: Velox doesn't provide built-in IN special form. Hence, I don't think this constant belongs here.

velox/functions/prestosql/InRewrite.h
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/core/Expressions.h"
+
+namespace facebook::velox::expression {
+
+class InRewrite {
```
mbasmanova, line 20: Isn't this rewrite specific to Presto?

velox/expression/InRewrite.cpp
```diff
@@ -0,0 +1,73 @@
...
+  if (!expr->isCallKind()) {
+    return nullptr;
+  }
+  const auto* callExpr = expr->asUnchecked<core::CallTypedExpr>();
+  if (callExpr->name() != kIn) {
+    return nullptr;
+  }
+
+  const auto& valueExpr = callExpr->inputs().at(0);
+  if (!valueExpr->isConstantKind()) {
```
mbasmanova: perhaps, make this the first check; it is cheaper than dynamic cast

`expr->inputs().size() == 2 && expr->inputs()[0]->isConstantKind()`

velox/expression/InRewrite.cpp
```diff
@@ -0,0 +1,73 @@
...
+    return nullptr;
+  }
+  // IN expression with a constant IN-list can be evaluated directly and will
```
mbasmanova: "can be evaluated directly" -> "should be constant folded, hence, doesn't need a rewrite"

velox/expression/InRewrite.cpp
```diff
@@ -0,0 +1,73 @@
...
+  // not be rewritten.
+  const auto inList = callExpr->inputs().at(1);
```
mbasmanova: const &

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,73 @@
...
+  return std::make_shared<core::CallTypedExpr>(
```
mbasmanova, line 70: perhaps, add a check that optimizedInputs is not empty

velox/expression/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+};
+
+TEST_F(InRewriteTest, basic) {
+  // 1234 in (2, a, b, 1234) -> true.
+  const auto type = ROW({"a", "b"}, {BIGINT(), BIGINT()});
+  auto in = std::make_shared<core::CallTypedExpr>(
+      BOOLEAN(),
+      "in",
+      constant(INTEGER(), 1234),
```
mbasmanova: perhaps, add a helper function makeIn that takes an std::initializer_list

```
makeIn(constant(INTEGER(), 2), field...)
```

velox/expression/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  static core::TypedExprPtr field(
+      const TypePtr& type,
+      const std::string& name) {
+    return std::make_shared<core::FieldAccessTypedExpr>(type, name);
+  }
+};
+
+TEST_F(InRewriteTest, basic) {
```
mbasmanova: Would you also add a test for IN list with all constant inputs?

velox/expression/ExprConstants.h
pramodsatya: Reverted, thanks.

velox/expression/InRewrite.cpp
pramodsatya: Updated, thanks.

velox/functions/prestosql/InRewrite.h
pramodsatya: Yes, thanks for pointing it out, moved it to `velox/functions/prestosql`, is that fine?

velox/functions/prestosql/InRewrite.cpp
aditi-pandit: @mbasmanova : What do you think about how this rewrite is written ?

In my mind each rewrite should be self contained in Velox... So if we are looking to optimize the pattern 
const value IN (list of constant values + columns), then we should optimize for both the true and false cases. But here there is an assumption that only if we are able to find the value its fine, else this has to be from the columns and we should always have a single column left. To me this sounds very fragile.

Looking forward to your feedback.

mbasmanova: @aditi-pandit 

> we should optimize for both the true and false cases

Sure, but what is the "false" case? Would you share an example?

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "velox/functions/prestosql/InRewrite.h"
+
+namespace facebook::velox::functions {
+
+core::TypedExprPtr InRewrite::rewrite(const core::TypedExprPtr& expr) {
+  // IN expression with a constant value and with a constant IN-list should be
+  // constant folded, hence, doesn't need a rewrite. IN expression with a
+  // constant value and with a non-constant IN-list can be rewritten.
+  if ((expr->inputs().size() == 2 && expr->inputs()[1]->isConstantKind()) ||
+      (!expr->inputs().empty() && !expr->inputs()[0]->isConstantKind())) {
```
mbasmanova: I'm confused.

I expect that we check that first input is constant, no? What else do we need to check?

I'd expect a check like this:

```
if (expr->inputs().empty() || !expr->inputs()[0]->isConstantKind()) {
   return nullptr;
}
```

pramodsatya: We also need to ensure the IN-list is not constant, this is represented by the second input for non-generic IN function implementation. I have simplified this check to `expr->inputs[1]->isConstantKind()`, although ideally it should be as follows:
```
  const auto inList = callExpr->inputs().at(1);
  if (inList->isConstantKind() && callExpr->inputs().size() == 2) {
    const auto* inListConstantExpr = inList->asUnchecked<core::ConstantTypedExpr>();
    if (inListConstantExpr->hasValueVector()) {
      const auto* constantArray = inListConstantExpr->valueVector()->as<ConstantVector<ComplexType>>();
      if (constantArray && constantArray->type()->isArray()) {
        return nullptr;
      }
    }
  }
```
I figured it would suffice to check `expr->inputs().size() == 2 && expr->inputs[1]->isConstantKind()`, instead of checking the second expression is a constant with an array vector representing the IN-list. Is that fine?

velox/functions/prestosql/InRewrite.cpp
aditi-pandit: @mbasmanova : So if we have a IN like

Constant(5) IN {Constant(6), Constant(7)} the rule should evaluate to false. But it seems like we assume there will always be a column if we couldn't find the constant value in the constants in the IN list (so !optimizedInputs.empty()). Does the Presto optimizer already have the knowledge to give Velox this expression only under this form ? Even if it does, this seems fragile.

A simple solution is for the Velox rule to return false if !optimizedInputs.empty() so that it atleast seems more logically complete to cover the cases of searching a constant value in an IN list.

aditi-pandit:
> @pramodsatya Looks good to me. I'll approve once @aditi-pandit 's questions are resolved and she approves.

@mbasmanova : https://github.com/facebookincubator/velox/pull/15488#discussion_r2535757898

velox/functions/prestosql/InRewrite.cpp
mbasmanova: > We also need to ensure the IN-list is not constant,

Maybe not. Can we change the rewrite API to guarantee that it is never called on expression with all constant inputs?

mbasmanova: Also, wondering if it makes sense to rewrite const IN (a, b, NULL, ...) to NULL.

velox/functions/prestosql/InRewrite.cpp
mbasmanova: I assume the ExprOptimizer would constant fold an expression with all constant inputs. Is this not the case?

velox/functions/prestosql/InRewrite.cpp
pramodsatya: Thanks for the suggestion, updated rewrite API and added rewrite for IN-list with NULLs.

velox/expression/ExprRewriteRegistry.cpp
```diff
@@ -28,6 +28,15 @@ void ExprRewriteRegistry::clear() {
 
 core::TypedExprPtr ExprRewriteRegistry::rewrite(
     const core::TypedExprPtr& expr) {
+  // Do not rewrite expression if all inputs are constant.
+  const auto& inputs = expr->inputs();
+  if (std::all_of(
+          inputs.begin(), inputs.end(), [](const core::TypedExprPtr& input) {
+            return input->isConstantKind();
+          })) {
+    return expr;
```
mbasmanova: Should this be a CHECK ? The caller should not call this API with all constant inputs.

pramodsatya: Thanks for the suggestion, this is called by `ExprOptimizer` and by `ExprCompiler`. Adding this as a `VELOX_CHECK` would cause expression compilation to fail for expressions with all constant inputs when `ExprSet` is constructed with `enableConstantFolding=false`. Could I modify this assignment in `ExprCompiler.cpp` from
```
  auto rewritten = enableConstantFolding
      ? expression::optimize(expr, queryCtx, pool)
      : expression::ExprRewriteRegistry::instance().rewrite(expr);
```
to 
```
  auto rewritten = enableConstantFolding ? expression::optimize(expr, queryCtx, pool) : expr;
```

mbasmanova: Do you know why we need 'enableConstantFolding' flag? Where do we set it to 'false'?

If we change the meaning of the flag, then we may need to rename it as well. 

CC: @pedroerp @bikramSingh91

pramodsatya: I am not sure why this flag was introduced, I do not see any instances in the OSS codebase where it is set to `false`. Is it set to `false` internally at Meta? If not, would it be alright to remove this flag?

velox/functions/prestosql/InRewrite.cpp
aditi-pandit: @mbasmanova : I had some misinterpretations of this in that why did constant folding then not handle the case when the constant occurred in the input, why is that handled in this rewrite. 

But I'm also fine if we assert that !optimizedInputs.empty() as you are saying below that
there is a pre-requisite that input expr has some non-constant inputs.

My questions are clarified.

aditi-pandit:
@pramodsatya : My doubts are clarified now.

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,72 @@
...
+  const auto numInputs = inputs.size();
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  optimizedInputs.emplace_back(valueExpr);
+
+  for (auto i = 1; i < numInputs; i++) {
+    if (inputs[i]->isConstantKind()) {
+      const auto* constInput =
+          inputs[i]->asUnchecked<core::ConstantTypedExpr>();
+      if (constInput->isNull()) {
+        return core::ConstantTypedExpr::makeNull(BOOLEAN());
```
mbasmanova: Would you double check that "5 in (null, a 5)" becomes "true" and not "null"?

velox/expression/ExprRewriteRegistry.cpp
mbasmanova: @pramodsatya Let's create a PR to remove the flag and see what breaks.

velox/functions/prestosql/InRewrite.cpp
mbasmanova: Let's also check that "null in (5, null, a)" returns "null", not "true".

pramodsatya: Modified the rewrite to handle these cases and added tests, thanks.

velox/functions/prestosql/InRewrite.cpp
pramodsatya: @mbasmanova, would it be alright to move the rewrite API changes to https://github.com/facebookincubator/velox/pull/15541 ?

velox/functions/prestosql/InRewrite.cpp
pramodsatya: This check is added in https://github.com/facebookincubator/velox/pull/15569.

velox/expression/ExprRewriteRegistry.cpp
pramodsatya: This is being tracked in https://github.com/facebookincubator/velox/pull/15541.

pramodsatya:
Thanks @mbasmanova, @aditi-pandit. Rebased on top of the `ExprRewriteRegistry::rewrite` API change and fixed the tests. Could you PTAL?

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,79 @@
...
+    return core::ConstantTypedExpr::makeNull(BOOLEAN());
+  }
+
+  const auto& inputs = callExpr->inputs();
+  const auto numInputs = inputs.size();
+  bool hasNullInput = false;
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  optimizedInputs.emplace_back(valueExpr);
+
+  for (auto i = 1; i < numInputs; i++) {
```
mbasmanova, line 52: can we use for-each loop over callExpr->inputs() ?

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (inputs[i]->isConstantKind()) {
+      const auto* constInput =
+          inputs[i]->asUnchecked<core::ConstantTypedExpr>();
+      if (constInput->isNull() && hasNullInput == false) {
```
mbasmanova: "&& hasNullInput == false" - can we drop this condition? why do we need it?

velox/functions/prestosql/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,114 @@
...
+      std::initializer_list<core::TypedExprPtr> inputs) {
+    return std::make_shared<core::CallTypedExpr>(BOOLEAN(), "in", inputs);
+  }
+
+  void testInRewrite(
+      const core::TypedExprPtr& expr,
+      const core::TypedExprPtr& expected) {
+    const auto rewritten =
+        expression::ExprRewriteRegistry::instance().rewrite(expr);
+    ASSERT_TRUE(*rewritten == *expected);
```
mbasmanova: better return a boolean and move ASSERT_TRUE to the caller. This way the assertion will contain the "right" line in the test.

velox/functions/prestosql/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,114 @@
...
+  // NULL in (5, NULL, a) -> NULL
+  in = makeIn(
+      {constant(BIGINT(), variant::null(TypeKind::BIGINT)),
+       constant(BIGINT(), 5LL),
+       constant(BIGINT(), variant::null(TypeKind::BIGINT)),
+       field(BIGINT(), "a")});
+  expected = constant(BOOLEAN(), variant::null(TypeKind::BOOLEAN));
+  testInRewrite(in, expected);
+
+  // 'hello' in ('foo', 'bar', a, NULL) -> NULL
```
mbasmanova: Is this correct? Shouldn't this become 'hello' in (a, NULL) ? if 'a' happens to be 'hello', the result will be true, not null.

velox/functions/prestosql/InRewrite.cpp
pramodsatya: Dropped this, thanks.

velox/functions/prestosql/InRewrite.cpp
pramodsatya: The first expression from inputs is the IN-value and should be skipped since it is not in the IN-list. Modified to use an iterator instead of accessing vector elements by index. Is that fine?

velox/functions/prestosql/tests/InRewriteTest.cpp
pramodsatya: Thanks, updated the rewrite to fix this.

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,77 @@
...
+      valueExpr->asUnchecked<core::ConstantTypedExpr>();
+  if (constValueExpr->isNull()) {
+    return core::ConstantTypedExpr::makeNull(BOOLEAN());
+  }
+
+  const auto& inputs = callExpr->inputs();
+  bool hasNullInput = false;
+  std::vector<core::TypedExprPtr> optimizedInputs;
+  optimizedInputs.emplace_back(valueExpr);
+  for (auto it = std::next(inputs.begin()); it != inputs.end(); ++it) {
```
mbasmanova: I see that for-each loop can't be used here. Let's then go back to the original loop, but add a local variable to avoid repeating inputs[i]. Thanks.

velox/functions/prestosql/tests/InRewriteTest.cpp
```diff
@@ -0,0 +1,118 @@
...
+       constant(INTEGER(), 2),
+       field(INTEGER(), "a"),
+       constant(INTEGER(), 4),
+       field(INTEGER(), "b"),
+       constant(INTEGER(), 1234)});
+  expected = makeIn(
+      {constant(INTEGER(), 81), field(INTEGER(), "a"), field(INTEGER(), "b")});
+  ASSERT_TRUE(testInRewrite(in, expected));
+
+  // 5 in (NULL, a, 5) -> true
```
mbasmanova, line 75: Would you add a test for "5 in (1, 2, null, 4)"?

pramodsatya: Thanks, I considered this but it would fail the `rewrite` API check since all inputs are constant.

velox/functions/prestosql/InRewrite.cpp
pramodsatya: Updated, thanks.

velox/functions/prestosql/InRewrite.cpp
```diff
@@ -0,0 +1,80 @@
...
+      }
+    } else {
+      optimizedInputs.push_back(input);
+    }
+  }
+
+  if (hasNullInput) {
+    optimizedInputs.push_back(
+        core::ConstantTypedExpr::makeNull(optimizedInputs.front()->type()));
+  }
```
mbasmanova: nit: perhaps, replace this logic with 

```
if (constInput->isNull()) {
   if (!hasNullInput) {
      optimizedInputs.push_back(input);
   } else {
      hasNullInput = true;
   }
}
```

mbasmanova (approved):
@pramodsatya Looks good % one last nit.

velox/functions/prestosql/InRewrite.cpp
pramodsatya: Updated, thanks.

PR #15489, opened 2025-11-13
fix: Fix StringView remapping in transferOrCopyTo() when string buffers are copied

kagamiori:
Summary:
FlatVector::transferOrCopyTo() and ConstantVector::transferOrCopyTo() may copy the string 
buffers to new ones. However, there was a bug that they didn't udpate the StringViews in the 
vectors afterwards. StringViews must be updated if a string buffer is replaced by its copy, 
because StringViews contain pointers to addresses inside the string buffers.

This diff updates StringViews in FlatVector::transferOrCopyTo() and 
ConstantVector::transferOrCopyTo() when a copy of a string buffer occurs. It also updated the 
unit test to cover these situations. 

In addition, this diff adds documentation of the BaseVector::transferOrCopyTo() API. It also 
removes FlatMapVector::transferOrCopyTo() since it's not covered in the unit test yet.

Differential Revision: D86680882

### DIFF
--- a/velox/vector/BaseVector.h
+++ b/velox/vector/BaseVector.h
@@ -542,6 +542,11 @@ class BaseVector {
   virtual VectorPtr testingCopyPreserveEncodings(
       velox::memory::MemoryPool* pool = nullptr) const = 0;
 
+  /// Transfer or copy this vector and all its buffers recursively to 'pool'.
+  /// The transfer of a buffer is allowed if its original pool and 'pool' are
+  /// from the same MemoryAllocator and the buffer is not a BufferView. If a
+  /// buffer is not allowed to be transferred, it is copied to pool. After this
+  /// call, this vector and all its buffers are owned by 'pool'.
   virtual void transferOrCopyTo(velox::memory::MemoryPool* pool);
 
   /// Construct a zero-copy slice of the vector with the indicated offset and
--- a/velox/vector/ConstantVector.h
+++ b/velox/vector/ConstantVector.h
@@ -408,8 +408,14 @@ class ConstantVector final : public SimpleVector<T> {
     if (valueVector_) {
       valueVector_->transferOrCopyTo(pool);
     }
-    if (stringBuffer_ && !stringBuffer_->transferTo(pool)) {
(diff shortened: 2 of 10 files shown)


velox/vector/ConstantVector.h
```diff
@@ -408,8 +408,14 @@ class ConstantVector final : public SimpleVector<T> {
     if (valueVector_) {
       valueVector_->transferOrCopyTo(pool);
     }
-    if (stringBuffer_ && !stringBuffer_->transferTo(pool)) {
-      stringBuffer_ = AlignedBuffer::copy<char>(stringBuffer_, pool);
+    if constexpr (std::is_same_v<T, StringView>) {
+      if (stringBuffer_ && !stringBuffer_->transferTo(pool)) {
```
bikramSingh91, line 412: can you also add a tiny method comment for `virtual bool transferTo(velox::memory::MemoryPool* /*pool*/)` in buffer.h ?

velox/vector/BaseVector.h
```diff
@@ -542,6 +542,11 @@ class BaseVector {
   virtual VectorPtr testingCopyPreserveEncodings(
       velox::memory::MemoryPool* pool = nullptr) const = 0;
 
+  /// Transfer or copy this vector and all its buffers recursively to 'pool'.
+  /// The transfer of a buffer is allowed if its original pool and 'pool' are
+  /// from the same MemoryManager and the buffer is not a BufferSlice. If a
```
bikramSingh91: nit: I am probably being a bit pedantic but do you mean the same MemoryAllocator? I know the MemoryManager and MemoryAllocator are inherently tied and have 1:1 mapping for now, but I feel the distinction between the two is still meaningful to clarify. As in the future, the same memory manager might support multiple allocators like one for GPU size memory and one for RAM.

bikramSingh91 (approved):

velox/vector/BaseVector.h
kagamiori: Yes, I mean "the same MemoryAllocator" which would be the case if they use the same MemoryManager. It's a good point that one memory manager can have multiple allocators. Let me update the comment as you suggest. Thanks!

PR #15496, opened 2025-11-13
fix: FileUtils::makePartName should throw on empty input

Yuhta:
Summary:
Empty input was not allowed before
https://github.com/facebookincubator/velox/pull/15480.  It was not obvious but
there are downstream projects depending on this behavior.  Add the extra check
to ensure we throw in this case.

Differential Revision: D87006275

### DIFF
--- a/velox/dwio/catalog/fbhive/test/FileUtilsTests.cpp
+++ b/velox/dwio/catalog/fbhive/test/FileUtilsTests.cpp
@@ -19,8 +19,10 @@
 #include "velox/common/base/Exceptions.h"
 #include "velox/dwio/catalog/fbhive/FileUtils.h"
 
+namespace facebook::velox::dwio::catalog::fbhive {
+namespace {
+
 using namespace ::testing;
-using namespace facebook::velox::dwio::catalog::fbhive;
 
 TEST(FileUtilsTests, makePartName) {
   std::vector<std::pair<std::string, std::string>> pairs{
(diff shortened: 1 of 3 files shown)


velox/dwio/catalog/fbhive/test/FileUtilsTests.cpp
```diff
@@ -31,6 +33,7 @@ TEST(FileUtilsTests, makePartName) {
   ASSERT_EQ(
       FileUtils::makePartName(pairs, false),
       "ds=2016-01-01/FOO=__HIVE_DEFAULT_PARTITION__/a%0Ab%3Ac=a%23b%3Dc");
+  ASSERT_THROW(FileUtils::makePartName({}, false), VeloxException);
```
mbasmanova, line 36: Thank you. Would be nice to also update documentation for this API to clarify this behavior.

PingLiuPing (approved):
Thank you.

Agree this is not obvious. Revisited the Velox code and cannot find a scenario that the input key-value pairs is empty since when generating the partition name there should be at least one partition.  There is special handling for empty case already.

```
std::string FileUtils::extractPartitionName(const std::string& filePath) {
  const auto& partitionParts = extractPartitionKeyValues(filePath);
  return partitionParts.empty() ? "" : makePartName(partitionParts, false);
}
```

It's understandable we should check the input first before further processing. The input varies across different projects.

PR #15498, opened 2025-11-13
feat(cudf): Implement and default to automatic GPU detection

zoltan:
Instead of providing an enabled configuration option, we're explicitly enabling cuDF with "auto" mode.

New behavior:
- unset (default): auto-detect GPU availability, enable cuDF if GPU is present
- cudf.disabled=false: require cuDF, fail if no GPU is present
- cudf.disabled=true: force disable cuDF

This is a QoL improvement for mixed cluster deployments and enables
shipping a single image that works on systems with and without a GPU.

### DIFF
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -1133,10 +1133,10 @@ Note: These configurations are experimental and subject to change.
      - Type
      - Default Value
      - Description
-   * - cudf.enabled
-     - bool
-     - true
-     - If true, enable cuDF. By default, it is enabled if compiled with cuDF.
+   * - cudf.disabled
+     - string
+     - unset
+     - Controls cuDF usage. When unset (default), auto-detects GPU availability. Set to "false" to require cuDF enabled (fails if no GPU detected), or "true" to force disable cuDF.
    * - cudf.memory_resource
      - string
      - async
--- a/velox/experimental/cudf/exec/ToCudf.cpp
+++ b/velox/experimental/cudf/exec/ToCudf.cpp
@@ -64,6 +64,18 @@ bool isAnyOf(const Base* p) {
   return ((dynamic_cast<const Deriveds*>(p) != nullptr) || ...);
 }
 
+bool detectGpuAvailable() {
+  int deviceCount = 0;
+  auto error = cudaGetDeviceCount(&deviceCount);
+
+  if (error != cudaSuccess || deviceCount == 0) {
+    cudaGetLastError();
+    return false;
+  }
+
+  return true;
+}
+
 } // namespace
 
 bool CompileState::compile(bool allowCpuFallback) {
@@ -420,8 +432,8 @@ struct CudfDriverAdapter {
 
   // Call operator needed by DriverAdapter
   bool operator()(const exec::DriverFactory& factory, exec::Driver& driver) {
-    if (!driver.driverCtx()->queryConfig().get<bool>(
-            CudfConfig::kCudfEnabled, CudfConfig::getInstance().enabled) &&
+    if (driver.driverCtx()->queryConfig().get<bool>(
+            CudfConfig::kCudfDisabled, CudfConfig::getInstance().disabled) &&
         allowCpuFallback_) {
       return false;
     }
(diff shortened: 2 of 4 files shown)


velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -64,6 +64,18 @@ bool isAnyOf(const Base* p) {
   return ((dynamic_cast<const Deriveds*>(p) != nullptr) || ...);
 }
 
+bool detectGpuAvailable() {
+  int deviceCount = 0;
+  cudaError_t error = cudaGetDeviceCount(&deviceCount);
```
majetideepak: nit: auto :)

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -491,7 +503,20 @@ CudfConfig& CudfConfig::getInstance() {
 void CudfConfig::initialize(
     std::unordered_map<std::string, std::string>&& config) {
   if (config.find(kCudfEnabled) != config.end()) {
-    enabled = folly::to<bool>(config[kCudfEnabled]);
```
majetideepak, line 494: We have to change the config type to string.
I think the default semantics should be like auto unless explicitly disabled. 
So convert `enabled` -> `disabled`. disabled set to false or unset means auto.

zoltan: once the system is initialized, it must be true or false, can't be auto - that is why I've left the actual property as bool.

I can change the default to be auto.

zoltan: but of course to change _the default_, it needs to become a string.

majetideepak: We can leave it as a bool and default to infer if the config is not set. Empty string.

zoltan: I like to idea of moving to auto by default and then renaming enabled to disabled. I'll change it.

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -420,8 +432,8 @@ struct CudfDriverAdapter {
 
   // Call operator needed by DriverAdapter
   bool operator()(const exec::DriverFactory& factory, exec::Driver& driver) {
-    if (!driver.driverCtx()->queryConfig().get<bool>(
-            CudfConfig::kCudfEnabled, CudfConfig::getInstance().enabled) &&
+    if (driver.driverCtx()->queryConfig().get<bool>(
+            CudfConfig::kCudfDisabled, CudfConfig::getInstance().disabled) &&
         allowCpuFallback_) {
```
lga-zurich, line 437: If cudf is disabled, do we care whether cpu fallback is allowed or not? That will just make for another combination of conflicting configurations. I would say that if cudf is not enabled, the rest of the cudf parameters should be irrelevant.

karthikeyann (approved):

karthikeyann:
@simoneves after this PR is merged, config templates in https://github.com/rapidsai/velox-testing/blob/main/presto/docker/config/template/etc_worker/config_native.properties should be updated

velox/docs/configs.rst
```diff
@@ -1133,10 +1133,10 @@ Note: These configurations are experimental and subject to change.
      - Type
      - Default Value
      - Description
-   * - cudf.enabled
-     - bool
-     - true
-     - If true, enable cuDF. By default, it is enabled if compiled with cuDF.
+   * - cudf.disabled
```
jinchengchenghh, line 1136: Why not still use cudf.enabled with default true value? And always detect if gpu is available, if not available, fallback to cpu, and log a warning

zoltan: @majetideepak felt like this was a cleaner approach in the first PR review yesterday, so we switched over.

devavret: Double negatives do hurt my brain. So I have to agree with @jinchengchenghh here. 

Also, I feel like you'd want users to opt-in for cudf execution since it's still experimental.

lga-zurich: > Double negatives do hurt my brain

I also find it very confusing.

Also agree that having a `.disabled` parameter makes it feel like something you need to opt-out of.

Finally, if the user does set it to some value, I don't think `auto` should second-guess the user and make it run where it shouldn't.

zoltan: ok, how about I switch back the enabled, and then
* if it's unset, we'll default to auto
* if it's enabled, and we don't have a GPU, we'll fail
* if it's disabled, it's disabled
would you be all OK with that?

@devavret as for opt-in: the user already opts-in because the VELOX_ENABLE_CUDF is off by default, in my opinion. if you're compiling with cuDF *and not setting* enabled to anything explicitly, I feel like auto is a good choice. wdyt?

jinchengchenghh: The caller should detect the gpu and decide if registerCudf, rather than detect for each task/plan.

zoltan: what do you mean "the caller"?

we run the detection exactly one, when we read the configuration.

jinchengchenghh: In Spark, it can define a config as discoveryScript which detects the GPU by nvidia-smi, schedule the GPU task to GPU node

velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -64,6 +64,18 @@ bool isAnyOf(const Base* p) {
   return ((dynamic_cast<const Deriveds*>(p) != nullptr) || ...);
 }
 
+bool detectGpuAvailable() {
+  int deviceCount = 0;
+  auto error = cudaGetDeviceCount(&deviceCount);
```
jinchengchenghh, line 69: If this runs in a CPU node, does cuda install?

zoltan: good point. let me macro it around just like kvikio does.

velox/docs/configs.rst
zoltan: the current code does not detect it for each task/plan. it is detected at startup, once.

devavret: >as for opt-in: the user already opts-in because the VELOX_ENABLE_CUDF is off by default, in my opinion. if you're compiling with cuDF and not setting enabled to anything explicitly, I feel like auto is a good choice. wdyt?

If some unified build sets VELOX_ENABLE_CUDF so as to have one package that can run anywhere, having stuff automatically run on a gpu on machines that have it could be a source of some confusion at best and frustration at worst.

PR #15500, opened 2025-11-13
fix: Back out "feat: Add support for evaluating Iceberg partition transforms"

Yuhta:
Summary:
Original commit changeset: 3f839a3a8ea6

Original Phabricator Diff: D86893852

Differential Revision: D87015771


PingLiuPing:
@Yuhta  Could you please share some info about this PR? Is there any regressions caused by https://github.com/facebookincubator/velox/pull/15477/? Thank you very much.

Yuhta:
@PingLiuPing There is some build size problem but it seems false positive to blame this change so far.  I will keep an eye and this but so far you don't need to worry about it.

yingsu00 (commented):
@Yuhta Hi Jimmy, is this PR still needed?

PR #15501, opened 2025-11-14
docs: Document error handling in simple functions

kevinwilfong:
Summary:
Clarify that error handling is taken care of for the user in simple functions and no action is
needed on their part (unlike Vector functions). Likewise, emphasize the import of not
handling VeloxRuntimeErrors as not allowing them to fail the Task can lead to undefined
behavior.

Differential Revision: D87021815

### DIFF
--- a/velox/docs/develop/scalar-functions.rst
+++ b/velox/docs/develop/scalar-functions.rst
@@ -538,6 +538,20 @@ The code below shows an example of a function that concatenates a variable numbe
        }
      };
 
+Error Handling
+^^^^^^^^^^^^^^
+
+Simple functions do not need to handle exceptions in the "call" function (or any of
+its variations) or the "initialize" function. If an exception is thrown, the
+framework will ensure it is either propagated to the Task or the semantics of
+Velox's builtin TRY expression are respected if applicable.
+
+It is important that VeloxRuntimeErrors not be caught and ignored or wrapped in
+another exception type. This exception type represents a system error, and we cannot
+guarantee Velox's internal state for the function is valid after such an exception is
+thrown. They need to be able to propagate to the Task, not allowing them to do so can
+lead to undefined behavior.
+
 Vector Functions
 ----------------
 


velox/docs/develop/scalar-functions.rst
```diff
@@ -538,6 +538,20 @@ The code below shows an example of a function that concatenates a variable numbe
...
+Simple functions do not need to handle exceptions in the "call" function (or any of
+its variations) or the "initialize" function. If an exception is thrown, the
+framework will ensure it is either propagated to the Task or the semantics of
+Velox's builtin TRY expression are respected if applicable.
+
+It is important that VeloxRuntimeErrors not be caught and ignored or wrapped in
+another exception type. This exception type represents a system error, and we cannot
+guarantee Velox's internal state for the function is valid after such an exception is
+thrown. They need to be able to propagate to the Task, not allowing them to do so can
+lead to undefined behavior.
```
bikramSingh91, line 553: should we put this at the top as its super important when discussing how to handle exceptions.

velox/docs/develop/scalar-functions.rst
```diff
@@ -538,6 +538,20 @@ The code below shows an example of a function that concatenates a variable numbe
        }
      };
 
+Error Handling
+^^^^^^^^^^^^^^
+
+Simple functions do not need to handle exceptions in the "call" function (or any of
+its variations) or the "initialize" function. If an exception is thrown, the
+framework will ensure it is either propagated to the Task or the semantics of
+Velox's builtin TRY expression are respected if applicable.
```
bikramSingh91, line 547: If we are talking about error handling, we should mention the preferred method of exposing error conditions using Status object. In case that is not possible and the implementation has to throw, its allowed to capture exceptions if thrown by an API call inside the implementation in order to wrap in into a more meaningful exception (with possibly additional relevant context).  However, it should ensure it always propagates (or never catches) VeloxRuntimeErrors.

PR #15502, opened 2025-11-14
docs: Update Presto functions coverage map

pramodsatya:
Updated the complete list of Presto scalar functions using output from ```SHOW
FUNCTIONS``` command. 

Re-generated coverage maps using instructions from
```velox/functions/prestosql/coverage/README.md```.


aditi-pandit:
@mbasmanova : Its been a while since the Presto function coverage maps were updated, and we are actually very green in coverage now :) Would be great to update these.

Please can you help with the review.

mbasmanova (approved):

Issue #15245, opened 2025-10-22
Ubuntu Benchmarks build is failing


majetideepak:
### Problem description

https://github.com/facebookincubator/velox/actions/runs/18688811569/job/53288733775
Started with commit: 
https://github.com/facebookincubator/velox/commit/c95e64e50266dfeea85d2c678e1fa81ed8ef3d40

```
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: error: ‘void* __builtin___memcpy_chk(void*, const void*, long unsigned int, long unsigned int)’ writing 13 or more bytes into a region of size 12 overflows the destination [-Werror=stringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### System information

CI

### CMake log

```bash

```

majetideepak:
@pedroerp can you please take a look? Thanks.

pedroerp:
ugh, sorry. Looking

pedroerp:
@majetideepak interesting. I tried to compile using gcc on my laptop, and I got a similar error, but from a different part of the code:

```
In member function ‘void facebook::velox::CpuWallTiming::add(const facebook::velox::CpuWallTiming&)’,
    inlined from ‘void facebook::velox::exec::HashBuild::addRuntimeStats()’ at /home/pedroerp/Github/velox-1/velox/exec/HashBuild.cpp:923:37:
/home/pedroerp/Github/velox-1/./velox/common/time/CpuWallTimer.h:35:11: error: writing 16 bytes into a region of size 0 [-Werror=stringop-overflow=]
   35 |     count += other.count;
      |     ~~~~~~^~~~~~~~~~~~~~
```

Have we recently changed anything in the warning flags we pass to the compiler? Cc: @kgpai @czentgr

pedroerp:
Doing some research, it looks like this warning in buggy on older versions of gcc, and many projects have workaround to disable it:

https://github.com/google/flatbuffers/issues/5950
https://github.com/google/flatbuffers/pull/6020

We even have some hacky macros in Velox to go around this:

https://github.com/facebookincubator/velox/blob/main/velox/common/base/Macros.h#L21

[...]

PR #15503, opened 2025-11-14
WIP: fix: Suppress stringop-overflow warning

kou:
I just try this for #15245.


kou:
This doesn't fix anything...

PR #15508, opened 2025-11-14
feat(python): Support for JSON type

pedroerp:
Summary: Adding support for JSON types through the Python API.

Differential Revision: D87097991


python/test/test_type.py
```diff
@@ -43,6 +44,7 @@ def test_simple_types(self):
         self.assertTrue(isinstance(DOUBLE(), Type))
         self.assertTrue(isinstance(VARCHAR(), Type))
         self.assertTrue(isinstance(VARBINARY(), Type))
+        self.assertTrue(isinstance(JSON(), Type))
```
kgpai, line 47: JsonType extends VarcharType  , so assertTrue(JSON().equivalent(VARCHAR()) should be true I presume, can you assert that ?

pedroerp: @kgpai we don't expose .equivalent() through the Python API (not clear why it would be needed). We only expose an equality operator, so I added a test to ensure it correctly identifies VARCHAR() and JSON() as different types.

PR #15510, opened 2025-11-14
feat: Add map_keys_overlap function to Velox

abhinavmuk04:
Summary:
This diff implements the `map_keys_overlap(map(K,V), array(K)) -> boolean` function for Velox, which returns true if any key in the map matches any element in the given array, and false otherwise.

The function handles various key types through specialized implementations:
- Primitive types (bool, int8, int16, int32, int64, float, double, timestamp, date) use `MapKeysOverlapPrimitiveFunction` with `folly::F14FastSet` for efficient lookups
- Varchar keys use `MapKeysOverlapVarcharFunction` with optimized `StringView` comparisons
- Generic/complex key types use `MapKeysOverlapFunction` with custom equality comparator

Key behaviors:
- Returns false if either the map or array is empty
- Null keys in the array are ignored (not matched)
- Map values can be null (only keys are compared)
- Early returns true on first matching key

The function is excluded from Presto fuzzer tests as it's a Velox-only function not available in Presto.

Differential Revision: D87098829

### DIFF
--- a/velox/functions/prestosql/MapKeysOverlap.h
+++ b/velox/functions/prestosql/MapKeysOverlap.h
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/expression/ComplexViewTypes.h"
+#include "velox/functions/Udf.h"
+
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/MapKeysOverlap.h
```diff
@@ -0,0 +1,129 @@
...
+ */
+#pragma once
+
+#include "velox/expression/ComplexViewTypes.h"
+
+namespace facebook::velox::functions {
+
+template <typename TExec, typename Key>
+struct MapKeysOverlapPrimitiveFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
zacw7, line 25: don't we need `#include "velox/functions/Udf.h"` to be able to use this macro?

velox/functions/prestosql/MapKeysOverlap.h
```diff
@@ -0,0 +1,129 @@
...
+      const arg_type<Array<Generic<T1>>>& keys) {
+    if (inputMap.empty() || keys.empty()) {
+      out = false;
+      return;
+    }
+
+    MapKeysOverlapFunctionEqualComparator comparator;
+
+    for (const auto& entry : inputMap) {
+      for (const auto& key : keys) {
```
zacw7: can we use a set here?

zacw7 (approved):
LG % nit.

zacw7 (approved):
Thanks for adding this!

PR #15511, opened 2025-11-15
feat: S2 presto functions (#15511)

0xandr:
Summary:

S2 Geometry (http://s2geometry.io/) is a library for spherical geometry that decomposes the Earth's surface into a hierarchy of cells. Unlike planar tiling systems (e.g., Bing Tiles), S2 cells have near-uniform area across all latitudes and support 31 levels of hierarchical resolution, making them well-suited for spatial indexing, containment checks, and compact region coverings.

New S2 cell functions:

| Function | Signature | Description |
|---|---|---|
| `s2_cell_area_sq_km` | `(bigint) -> double` | Surface area of a cell in km² |
| `s2_cell_contains` | `(bigint, bigint) -> boolean` | Hierarchical containment check |
| `s2_cell_from_token` | `(varchar) -> bigint` | Convert hex token to cell ID |
| `s2_cell_level` | `(bigint) -> integer` | Get level (0-30) of a cell |
| `s2_cell_parent` | `(bigint, integer) -> bigint` | Get ancestor at a given level |
| `s2_cell_to_token` | `(bigint) -> varchar` | Convert cell ID to hex token |
| `s2_cells` | `(Geometry, integer) -> array(bigint)` | Cover geometry at fixed level |
| `s2_cells` | `(Geometry, integer, integer, integer) -> array(bigint)` | Compact mixed-level covering (similar to `geometry_to_dissolved_bing_tiles`) |

Design: all functions operate on cell IDs (BIGINT). Token conversion is
explicit via `s2_cell_from_token`/`s2_cell_to_token`. Level parameters and
return values use INTEGER consistently.

Build isolation:
S2 cell operations are isolated in a separate translation unit (`S2CellOperations.cpp`) to prevent a symbol conflict between s2geometry's `malloc_extension.h` and folly's `MallocImpl.h` — both define `nallocx` with incompatible types (function declaration vs function pointer). The isolation header (`S2CellOperations.h`) exposes only primitive-type wrappers, keeping s2geometry headers out of any translation unit that includes folly.

Dependencies: Adds a bundled dependency on s2geometry 0.12.0 (the first version to remove the glog dependency). Includes a one-line patch (`s2geometry-gcc12-max.patch`) to fix a `std::max` template deduction failure with GCC 12 in `s2cell_id.cc`, where `absl::bit_width` returns `unsigned long` but the other argument is `int`. s2geometry 0.13.1 was considered but requires abseil APIs (`absl::byteswap`, `absl::endian`, `SpinLock`) not present in Velox's bundled abseil `20240116` — upgrading s2geometry will be a simple version bump once Velox upgrades abseil. Includes a
one-line patch for a `std::max` template deduction failure with GCC 12.

All S2 functionality is gated behind `VELOX_ENABLE_GEO`.

Reviewed By: Sullivan-Patrick, pratikpugalia

Differential Revision: D78041102

### DIFF
--- a/CMake/resolve_dependency_modules/s2geometry.cmake
+++ b/CMake/resolve_dependency_modules/s2geometry.cmake
@@ -1,0 +1,54 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+include_guard(GLOBAL)
+
+list(PREPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/s2geometry)
+
+# This creates a separate scope so any changed variables don't affect
+# the rest of the build.
+block()
+  set(VELOX_S2GEOMETRY_BUILD_VERSION 0.12.0)
+  set(
+    VELOX_S2GEOMETRY_BUILD_SHA256_CHECKSUM
+    c09ec751c3043965a0d441e046a73c456c995e6063439a72290f661c1054d611
+  )
+  string(
+    CONCAT
+    VELOX_S2GEOMETRY_SOURCE_URL
+    "https://github.com/google/s2geometry/archive/refs/tags/"
+    "v${VELOX_S2GEOMETRY_BUILD_VERSION}.tar.gz"
+  )
+
+  velox_resolve_dependency_url(S2GEOMETRY)
+
+  FetchContent_Declare(
+    s2geometry
+    URL ${VELOX_S2GEOMETRY_SOURCE_URL}
+    URL_HASH ${VELOX_S2GEOMETRY_BUILD_SHA256_CHECKSUM}
+    OVERRIDE_FIND_PACKAGE
+    SYSTEM
+    EXCLUDE_FROM_ALL
+    PATCH_COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/s2geometry/s2geometry-gcc12-max.patch
+  )
+
+  list(APPEND CMAKE_MODULE_PATH "${s2geometry_SOURCE_DIR}/cmake")
+  set(BUILD_SHARED_LIBS OFF)
+  set(BUILD_TESTING OFF)
+  set(BUILD_TESTS OFF)
+  set(CMAKE_BUILD_TYPE Release)
+
+  FetchContent_MakeAvailable(s2geometry)
+
+  add_library(s2::s2 ALIAS s2)
+endblock()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -762,4 +762,10 @@ if(VELOX_ENABLE_GEO)
   velox_resolve_dependency(geos)
 endif()
 
+if(VELOX_ENABLE_GEO)
+  list(PREPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake/resolve_dependency_modules/s2geometry)
+  velox_set_source(s2geometry)
+  velox_resolve_dependency(s2geometry)
+endif()
+
 add_subdirectory(velox)
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -64,6 +64,8 @@ using facebook::velox::fuzzer::FuzzerRunner;
 using facebook::velox::fuzzer::JsonExtractArgValuesGenerator;
 using facebook::velox::fuzzer::JsonParseArgValuesGenerator;
 using facebook::velox::fuzzer::QDigestArgValuesGenerator;
+using facebook::velox::fuzzer::S2CellIdArgValuesGenerator;
+using facebook::velox::fuzzer::S2CellTokenArgValuesGenerator;
 using facebook::velox::fuzzer::SetDigestArgValuesGenerator;
 using facebook::velox::fuzzer::TDigestArgValuesGenerator;
 using facebook::velox::fuzzer::UnifiedDigestArgValuesGenerator;
@@ -136,7 +138,14 @@ std::unordered_map<std::string, std::shared_ptr<ArgValuesGenerator>>
          std::make_shared<SetDigestArgValuesGenerator>("jaccard_index")},
         {"intersection_cardinality",
          std::make_shared<SetDigestArgValuesGenerator>(
-             "intersection_cardinality")}};
+             "intersection_cardinality")},
+        {"s2_cell_area_sq_km", std::make_shared<S2CellIdArgValuesGenerator>()},
+        {"s2_cell_contains", std::make_shared<S2CellIdArgValuesGenerator>()},
+        {"s2_cell_level", std::make_shared<S2CellIdArgValuesGenerator>()},
+        {"s2_cell_parent", std::make_shared<S2CellIdArgValuesGenerator>()},
+        {"s2_cell_to_token", std::make_shared<S2CellIdArgValuesGenerator>()},
+        {"s2_cell_from_token",
+         std::make_shared<S2CellTokenArgValuesGenerator>()}};
 
 // TODO: List of the functions that at some point crash or fail and need to
 // be fixed before we can enable.
@@ -297,6 +306,7 @@ std::unordered_set<std::string> skipFunctions = {
     "convex_hull_agg",
     "geometry_union_agg",
     "localtime",
+    "s2_cells",
 };
 
 std::unordered_set<std::string> skipFunctionsSOT = {
@@ -483,6 +493,13 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "uniqueness_distribution(khyperloglog,bigint) -> map(bigint,double)",
     "merge_khll(array(khyperloglog)) -> khyperloglog",
     "pmod", // Not available in Presto
+    // S2 functions are not registered in the Presto instance used for fuzzing.
+    "s2_cell_area_sq_km",
+    "s2_cell_contains",
+    "s2_cell_from_token",
+    "s2_cell_level",
+    "s2_cell_parent",
+    "s2_cell_to_token",
 };
 
 int main(int argc, char** argv) {
--- a/velox/functions/prestosql/CMakeLists.txt
+++ b/velox/functions/prestosql/CMakeLists.txt
@@ -117,6 +117,7 @@ velox_add_library(
   RegexpReplace.h
   RegexpSplit.h
   RemapKeys.h
+  S2Functions.h
   SetDigestFunctions.h
   SfmSketchFunctions.h
   SplitPart.h
@@ -153,7 +154,14 @@ endif()
 
 if(VELOX_ENABLE_GEO)
   add_subdirectory(geospatial)
-  velox_link_libraries(velox_functions_prestosql_impl velox_functions_geo)
+
+  # S2 cell operations are isolated in a separate library to prevent a
+  # nallocx symbol conflict between s2geometry's malloc_extension.h and
+  # folly's MallocImpl.h.
+  velox_add_library(velox_s2_cell_operations S2CellOperations.cpp HEADERS S2CellOperations.h)
+  velox_link_libraries(velox_s2_cell_operations s2::s2 GEOS::geos)
+
+  velox_link_libraries(velox_functions_prestosql_impl velox_functions_geo velox_s2_cell_operations)
 endif()
 
 if(NOT VELOX_MONO_LIBRARY)
--- a/velox/functions/prestosql/S2CellOperations.cpp
+++ b/velox/functions/prestosql/S2CellOperations.cpp
@@ -1,0 +1,230 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/prestosql/S2CellOperations.h"
+
+#include <memory>
+
+#include <vector>
+
+#include <geos/geom/Geometry.h>
+#include <geos/geom/LineString.h>
+#include <geos/geom/LinearRing.h>
+#include <geos/geom/Polygon.h>
+#include <s2/s2cell.h>
+#include <s2/s2cell_id.h>
+#include <s2/s2earth.h>
+#include <s2/s2latlng.h>
+#include <s2/s2loop.h>
+#include <s2/s2point.h>
+#include <s2/s2polygon.h>
+#include <s2/s2polyline.h>
+#include <s2/s2region_coverer.h>
+
+namespace facebook::velox::functions {
+
+namespace {
+
+// Maps a GEOS coordinate (x=longitude, y=latitude) to a point on the
+// unit sphere.
+S2Point toS2Point(const geos::geom::Coordinate& coord) {
+  return S2LatLng::FromDegrees(coord.y, coord.x).ToPoint();
+}
+
+// Converts a GEOS linear ring to an S2Loop. Skips the closing vertex
+// (GEOS rings repeat the first vertex; S2 loops are implicitly closed).
+// Normalizes the loop so it encloses the smaller of the two spherical
+// regions it defines.
+std::unique_ptr<S2Loop> toS2Loop(const geos::geom::LinearRing& ring) {
+  auto coords = ring.getCoordinates();
+  auto numVertices = coords->getSize() - 1;
+  std::vector<S2Point> points;
+  points.reserve(numVertices);
+  for (size_t i = 0; i < numVertices; ++i) {
+    points.push_back(toS2Point(coords->getAt(i)));
+  }
+  auto loop = std::make_unique<S2Loop>(points);
+  loop->Normalize();
+  return loop;
+}
+
+// Converts a GEOS geometry to S2 objects and appends the cell covering.
+// A polygon valid in GEOS can be invalid on the sphere (e.g.,
+// self-intersecting edges after geodesic projection), which is why we
+// check S2Polygon::IsValid().
+std::string appendCovering(
+    const geos::geom::Geometry& geom,
+    int32_t level,
+    S2RegionCoverer& coverer,
+    std::vector<uint64_t>& result) {
+  if (geom.isEmpty()) {
+    return {};
+  }
+
+  auto typeId = geom.getGeometryTypeId();
+
+  if (typeId == geos::geom::GeometryTypeId::GEOS_POINT) {
+    result.push_back(
+        S2CellId(toS2Point(*geom.getCoordinate())).parent(level).id());
+
+  } else if (typeId == geos::geom::GeometryTypeId::GEOS_LINESTRING) {
+    auto coords = geom.getCoordinates();
+    if (coords->getSize() < 2) {
+      return "LineString must have at least 2 points";
+    }
+    std::vector<S2Point> points;
+    points.reserve(coords->getSize());
+    for (size_t i = 0; i < coords->getSize(); ++i) {
+      points.push_back(toS2Point(coords->getAt(i)));
+    }
+    S2Polyline polyline(points);
+    std::vector<S2CellId> covering;
+    coverer.GetCovering(polyline, &covering);
+    for (const auto& cellId : covering) {
+      result.push_back(cellId.id());
+    }
+
+  } else if (typeId == geos::geom::GeometryTypeId::GEOS_POLYGON) {
+    const auto& polygon = static_cast<const geos::geom::Polygon&>(geom);
+
+    auto* exteriorRing = polygon.getExteriorRing();
+    if (exteriorRing->getNumPoints() < 4) {
+      return "Polygon ring must have at least 4 points";
+    }
+
+    // Build S2Polygon from exterior ring and holes. The S2Polygon
+    // constructor determines hole nesting automatically.
+    std::vector<std::unique_ptr<S2Loop>> loops;
+    loops.reserve(1 + polygon.getNumInteriorRing());
+    loops.push_back(toS2Loop(*exteriorRing));
+    for (size_t i = 0; i < polygon.getNumInteriorRing(); ++i) {
+      loops.push_back(toS2Loop(*polygon.getInteriorRingN(i)));
+    }
+
+    S2Polygon s2polygon(std::move(loops));
+    if (!s2polygon.IsValid()) {
+      return "Polygon is not valid on the sphere";
+    }
+    std::vector<S2CellId> covering;
+    coverer.GetCovering(s2polygon, &covering);
+    for (const auto& cellId : covering) {
+      result.push_back(cellId.id());
+    }
+
+  } else {
+    for (size_t i = 0; i < geom.getNumGeometries(); ++i) {
+      auto error =
+          appendCovering(*geom.getGeometryN(i), level, coverer, result);
+      if (!error.empty()) {
+        return error;
+      }
+    }
+  }
+
+  return {};
+}
+
+} // namespace
+
+namespace {
+
+S2CellId s2CellId(int64_t cellId) {
+  return S2CellId(static_cast<uint64_t>(cellId));
+}
+
+int64_t fromS2CellId(S2CellId cellId) {
+  return static_cast<int64_t>(cellId.id());
+}
+
+} // namespace
+
+bool S2CellOp::isValid(int64_t cellId) {
+  return s2CellId(cellId).is_valid();
+}
+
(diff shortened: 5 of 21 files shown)


velox/functions/prestosql/CMakeLists.txt
```diff
@@ -131,6 +131,12 @@ velox_add_library(
   WordStem.h
 )
 
+# S2 cell operations are isolated in a separate library to prevent a
+# google::DEBUG_MODE symbol conflict between s2geometry's and glog's
+# log_severity.h headers.
+velox_add_library(velox_s2_cell_operations S2CellOperations.cpp)
```
pratikpugalia: You would have to add the headers `S2CellOperations.h` and `S2Functions.h` here!

More information [here](https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md?plain=1#L397)
```suggestion
velox_add_library(velox_s2_cell_operations S2CellOperations.cpp HEADERS S2CellOperations.h S2Functions.h)
```

CMakeLists.txt
```diff
@@ -762,4 +762,9 @@ if(VELOX_ENABLE_GEO)
   velox_resolve_dependency(geos)
 endif()
 
+# S2 geometry library for geospatial functions.
+# Always bundled since it is not commonly available as a system package.
+set(s2geometry_SOURCE BUNDLED)
```
pratikpugalia: should use velox_set_source which would anyway fallback to BUNDLED if not found in the system 

```suggestion
velox_set_source(s2geometry)
```

pratikpugalia:
LGTM otherwise from the perspective of build. cc: @mbasmanova

CMakeLists.txt
0xandr: This seems to fail the build. velox_set_source respects the global VELOX_DEPENDENCY_SOURCE env var, so when CI sets VELOX_DEPENDENCY_SOURCE=SYSTEM, s2geometry gets SYSTEM and CMake tries find_package(s2geometry REQUIRED) — which fails

velox/functions/prestosql/S2CellOperations.h
```diff
@@ -0,0 +1,47 @@
...
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <string_view>
+
+namespace facebook::velox::functions {
+
+/// Returns whether cellId is a valid S2 cell ID.
+bool s2CellIdIsValid(uint64_t cellId);
```
mbasmanova: put these into S2Cell class as static methods; drop s2Cell prefix

velox/functions/prestosql/S2CellOperations.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+int s2CellIdLevel(uint64_t cellId) {
+  return S2CellId(cellId).level();
+}
+
+uint64_t s2CellIdParent(uint64_t cellId, int level) {
+  return S2CellId(cellId).parent(level).id();
+}
+
+uint64_t s2CellIdFromToken(std::string_view token) {
```
mbasmanova: Why do we need wrappers for these one-liners?

velox/functions/prestosql/S2Functions.h
```diff
@@ -0,0 +1,168 @@
...
+struct S2CellIDAreaSqKmFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  FOLLY_ALWAYS_INLINE Status
+  call(out_type<double>& result, const arg_type<Varchar>& cellIdStr) {
+    if (cellIdStr.empty()) {
+      if (threadSkipErrorDetails()) {
+        return Status::UserError();
+      }
+      return Status::UserError("S2_CELL_AREA_SQ_KM: Empty cell ID");
```
mbasmanova: This repeats a lot. Consider extracting a helper.

velox/functions/prestosql/S2Functions.h
```diff
@@ -0,0 +1,168 @@
...
+#include "velox/functions/prestosql/S2CellOperations.h"
+
+namespace facebook::velox::functions {
+
+template <typename TExec>
+struct S2CellIDAreaSqKmFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  FOLLY_ALWAYS_INLINE Status
+  call(out_type<double>& result, const arg_type<Varchar>& cellIdStr) {
```
mbasmanova: `out_type<double>` -> double for readability

velox/functions/prestosql/S2Functions.h
```diff
@@ -0,0 +1,168 @@
...
+      }
+      return Status::UserError("S2_CELL_TOKEN_PARENT: Empty cell token");
+    }
+
+    if (level < 0 || level > 30) {
+      if (threadSkipErrorDetails()) {
+        return Status::UserError();
+      }
+      return Status::UserError(
+          "S2_CELL_TOKEN_PARENT: Expected level {} to be in [0,30] range",
```
mbasmanova: make sure to put runtime info at the end of the message

velox/functions/prestosql/S2Functions.h
```diff
@@ -0,0 +1,168 @@
...
+      if (threadSkipErrorDetails()) {
+        return Status::UserError();
+      }
+      return Status::UserError(
+          "S2_CELL_ID_PARENT: Expected level {} to be in [0,30] range", level);
+    }
+
+    std::string_view cellIdView(cellIdStr.data(), cellIdStr.size());
+
+    auto cellIdResult = folly::tryTo<uint64_t>(cellIdView);
```
mbasmanova: this code is duplicated; please dedup

velox/functions/prestosql/tests/S2FunctionsTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (expected.has_value()) {
+      ASSERT_TRUE(result.has_value());
+      EXPECT_EQ(expected.value(), result.value());
+    } else {
+      EXPECT_FALSE(result.has_value());
+    }
+  }
+};
+
+TEST_F(S2FunctionsTest, s2CellIdParent) {
```
mbasmanova: drop s2Cell prefixes; test name provides context

velox/functions/prestosql/tests/S2FunctionsTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include <gtest/gtest.h>
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+
+using facebook::velox::functions::test::FunctionBaseTest;
+
+class S2FunctionsTest : public FunctionBaseTest {
+ protected:
+  void testS2CellIdParent(
```
mbasmanova: make this a local lambda

velox/functions/prestosql/tests/S2FunctionsTest.cpp
```diff
@@ -0,0 +1,140 @@
...
+      const std::optional<std::string>& expected) {
+    auto result = evaluateOnce<std::string>(
+        "s2_cell_token_parent(c0, c1)", cellToken, level);
+    if (expected.has_value()) {
+      ASSERT_TRUE(result.has_value());
+      EXPECT_EQ(expected.value(), result.value());
+    } else {
+      EXPECT_FALSE(result.has_value());
+    }
+  }
```
mbasmanova: copy-paste; please dedup

CMakeLists.txt
pratikpugalia: My bad, this will work in CI when you add `install_s2geometry` to the CI setup scripts (`setup-fedora.sh, setup-ubuntu.sh, setup-macos.sh, setup-centos9.sh`), similar to how `install_geos` is defined in `setup-common.sh`.

pratikpugalia:
Additionally, should `s2geometry` be an optional dependency behind a CMake option (like `VELOX_ENABLE_GEO` for geos)? Currently it's added unconditionally. Users who don't need S2 would take the build-time hit for no benefit.

Wdyt @mbasmanova @kgpai ?

velox/functions/prestosql/S2Functions.h
0xandr: switched to using ints for function signatures, so this is obsolete

mbasmanova:
**Previous review comments (April 1) — not yet addressed:**

1. **S2CellOperations.h** — "put these into S2Cell class as static methods; drop s2Cell prefix" — still free functions in `s2` namespace, no class.

2. **S2Functions.h** — "This repeats a lot. Consider extracting a helper." — level validation (`if (level < 0 || level > 30)`) is still duplicated across `S2CellIdParentFunction`, `GeometryToS2CellIdsFunction`, and three times in `GeometryToDissolvedS2CellIdsFunction`.

3. **S2Functions.h** — "this code is duplicated; please dedup" — the envelope null check + deserialize pattern is still duplicated between `GeometryToS2CellIdsFunction` and `GeometryToDissolvedS2CellIdsFunction`.

4. **S2FunctionsTest.cpp** — "make this a local lambda" — `assertEqualResult` is still a class method.

5. **S2FunctionsTest.cpp** — "copy-paste; please dedup" — `testParent`, `testAreaSqKm`, `testFromToken`, `testToToken` still follow the same pattern and could be collapsed.

Could you address these or reply to the original comments explaining why they don't apply?

**New comments:**

6. **S2CellOperations.h** — The build isolation rationale (nallocx symbol conflict between s2geometry and folly) is well explained in the PR description, but it's not in the header file itself. Please add a comment explaining why this isolation layer exists, so future readers and maintainers understand the design choice without having to find the PR.

7. **S2Functions.h** — The `userError` and `validateCellId` helpers are in the `s2` namespace, which collides with the namespace used in `S2CellOperations.h`. Local helpers in a header should be in a `detail` namespace.

8. **Naming** — Since these are not existing Presto functions, we have freedom to choose better names. Please rename all functions:
   - Drop `_id` from the `s2_cell_id_*` prefix: `s2_cell_area_sq_km`, `s2_cell_contains`, `s2_cell_from_token`, `s2_cell_level`, `s2_cell_parent`, `s2_cell_to_token`
   - Rename the geometry functions to `s2_cells(geometry, level)` for fixed-level covering and use overloading for the dissolved variant: `s2_cells(geometry, min_level, max_level, max_cells)` — instead of a separate `geometry_to_dissolved_s2_cell_ids` function. This follows the pattern of `bing_tiles` and avoids a proliferation of function names.

9. **Documentation** — The intro in `geospatial.rst` needs more context:
   - Explain what a token is (a compact hex representation of the cell ID) and its relationship to the cell ID.
   - Explain the design choice that all functions operate on cell IDs (bigint), with `s2_cell_from_token` / `s2_cell_to_token` providing conversion. This helps users understand why there aren't token-based variants of every function.

10. **PR description** — Please add a summary of the new functions with their signatures and brief descriptions (similar to the docs in `geospatial.rst`). The current description focuses on build and dependency details but doesn't give readers a quick overview of what functions are being added and what they do.

0xandr:
@mbasmanova thanks for a thorough review, I'd like to follow up on the following:

> 3. **S2Functions.h** — "this code is duplicated; please dedup" — the envelope null check + deserialize pattern is still duplicated between `GeometryToS2CellIdsFunction` and `GeometryToDissolvedS2CellIdsFunction`.

I tried and it failed because out_type<Array<int64_t>> and arg_type<Geometry> are only available inside a struct that has VELOX_DEFINE_FUNCTION_TYPES(TExec), and adding another function templated on TExec seemed more trouble than worth. 

> 4. **S2FunctionsTest.cpp** — "make this a local lambda" — `assertEqualResult` is still a class method.

I don't think I can use ASSERT_TRUE inside a lambda, since it will terminate the lambda and not the test, can I? IIUC gtest macros use `return` internally, so these must be member functions (not lambdas) for ASSERT_* to abort the test correctly.

> 5. **S2FunctionsTest.cpp** — "copy-paste; please dedup" — `testParent`, `testAreaSqKm`, `testFromToken`, `testToToken` still follow the same pattern and could be collapsed.

I'm not sure I follow, these helpers take different arg types, check functions that return different types. I'm not sure how these can be collapsed? I guess I could use a templated function, but I doubt this will be shorter and more clear. Please advise.

> 8. **Naming** — Since these are not existing Presto functions, we have freedom to choose better names. Please rename all functions:
>    * Drop `_id` from the `s2_cell_id_*` prefix: `s2_cell_area_sq_km`, `s2_cell_contains`, `s2_cell_from_token`, `s2_cell_level`, `s2_cell_parent`, `s2_cell_to_token`

I considered dropping the ID however unlike Bing Tile functions that accept and return a custom BingTile type, these deal with raw int IDs, so it seemed not consistent to not mention ID. I understand however these functions can be overloaded, so I guess no harm in dropping _id prefix.

>    * Rename the geometry functions to `s2_cells(geometry, level)` for fixed-level covering and use overloading for the dissolved variant: `s2_cells(geometry, min_level, max_level, max_cells)` — instead of a separate `geometry_to_dissolved_s2_cell_ids` function. This follows the pattern of `bing_tiles` and avoids a proliferation of function names.

BingTile actually does have a `geometry_to_dissolved_bing_tiles()` function, I really only tried to follow an established convention here, but I am ok overloading instead.

mbasmanova:
Thanks for pointing out that `geometry_to_dissolved_bing_tiles` follows the same naming pattern — that's a good catch. Looking at it, I think the bing tile naming itself could be improved. Rather than following that precedent, let's establish a better convention going forward. For the S2 functions, let's go with `s2_cells(geometry, level)` and `s2_cells(geometry, min_level, max_level, max_cells)` as overloads.

@Sullivan-Patrick @kgpai — Let's also rename `geometry_to_dissolved_bing_tiles` for consistency. Please file a follow-up PR.

mbasmanova:
**3. Dedup envelope null check** — The duplicated code doesn't actually depend on `out_type` or `arg_type` — it only uses `StringView` (which `arg_type<Geometry>` resolves to). The `envelope->isNull()` check is not self-explanatory — a reader seeing this for the first time wouldn't know why a null envelope means the geometry should be skipped. Please extract a well-named, documented helper. Something like:

```cpp
/// Deserializes a geometry, returning nullptr for empty geometries
/// (those with null envelopes).
static std::unique_ptr<geos::geom::Geometry> deserializeNonEmpty(
    const StringView& geometry);
```

Then the caller becomes:
```cpp
auto geom = deserializeNonEmpty(geometry);
if (!geom) {
  return Status::OK();
}
```

This pattern is also duplicated in existing code (e.g., `geometry_to_bing_tiles`). @Sullivan-Patrick @kgpai — please help update the existing geometry functions to use this helper as a follow-up.

mbasmanova:
`CoveringResult` is a custom error reporting pattern. Velox provides `Expected<T>` (alias for `folly::Expected<T, Status>`) for functions that return a value or an error. Please replace `CoveringResult` with `Expected<std::vector<uint64_t>>` and rename to `S2CellOp::tryCovering` and `S2CellOp::tryDissolvedCovering`. The `try` prefix follows the Folly convention for functions that report errors without throwing.

mbasmanova:
`S2CellOperations.h` — The class and method comments use `//` but public APIs in headers should use `///`, and each public method needs a doc comment. Please review [CODING_STYLE.md](https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md) and make sure all changes in this PR comply.

mbasmanova:
**4. assertEqualResult / testXxx as lambdas** — `ASSERT_*` macros behave the same in member functions as in lambdas — they `return` from the current function, not from the caller. So the current member function approach doesn't provide any advantage over lambdas.

Each of `testParent`, `testAreaSqKm`, `testFromToken`, `testToToken` is used in only one test. Having them as class methods makes them look like shared utilities. Please make them local lambdas within each test.

CMake/resolve_dependency_modules/s2geometry.cmake
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+include_guard(GLOBAL)
+
+# This creates a separate scope so any changed variables don't affect
+# the rest of the build.
+block()
+  set(VELOX_S2GEOMETRY_BUILD_VERSION 0.12.0)
```
kgpai, line 21: Curious, does S2 Geometry have dependencies on geos - or is it independent?

pratikpugalia (approved):

mbasmanova:
Looks good — all previous feedback addressed. A few nits:

1. Test constants and the test class should be placed in an anonymous namespace inside `facebook::velox::functions`.

2. The test constants are magic numbers — please add comments explaining what each cell represents and how they relate to each other. For example, is `kLevel20CellId` a descendant of `kLevel14CellId`? Is `kLevel14ParentId` the parent of both? What location do they correspond to? A reader seeing these for the first time can't tell.

3. Token strings `"8085808a1b5"` and `"8085808b"` are repeated across tests. Extract these as named constants alongside the cell ID constants.

4. Remove the null-input test cases (e.g., `test(std::nullopt, std::nullopt)`). These functions use default-null-behavior, so null propagation is handled by the framework — no need to test it. Once null tests are removed, the test lambda parameters should be non-optional (e.g., `int64_t cellId, int32_t level, int64_t expected` instead of `const std::optional<int64_t>&`). Error cases should call `evaluateOnce` directly rather than going through the test lambda.

5. `lvl` — do not abbreviate. Use `level`.

6. Tests use `cast(cardinality(s2_cells(...)) as varchar)` which tests `cast`, `cardinality`, and `ST_GeometryFromText` in addition to `s2_cells`. Test the function directly — evaluate the array result and check it, rather than wrapping in unrelated functions.

mbasmanova:
Getting close. A few more items:

1. **Naming** — Please go over the code and simplify parameter names throughout. The function name already provides context, so `parentCellIdVal` → `parent`, `childCellIdVal` → `child`, `cellIdVal` → `cellId`, `cellToken` → `token`, etc.

2. **`S2CellOp` should use `int64_t`, not `uint64_t`** — The constant casting between `int64_t` and `uint64_t` throughout `S2Functions.h` is noisy and error-prone. Since the SQL type is `BIGINT` (int64_t), the `S2CellOp` API should use `int64_t` everywhere. Just document that the values are reinterpreted from unsigned S2 cell IDs.

3. **Drop `using facebook::velox::functions::test::FunctionBaseTest`** — The test is inside the `facebook::velox::functions` namespace already. Just use `test::FunctionBaseTest`.

4. **Simplify `makeFlatVector<StringView>({StringView(wkt)})`** — Try `makeFlatVector({wkt})`.

5. **Simplify `std::optional<std::string>(std::string(kLevel14Token))`** — This is very hard to read. Use `std::optional(std::string(kLevel14Token))` or similar.

6. **Drop unnecessary `tokenView` variable** — `S2CellOp::fromToken` takes `std::string_view`, so pass the token directly without creating an intermediate variable.

7. **Add `countCells` helper** — For tests that only verify the number of cells, use `EXPECT_EQ(1, countCells("POINT (0 0)", 10))` instead of `auto pointCells = cells(...); EXPECT_EQ(1, pointCells.size())`. Use `EXPECT_EQ(0, countCells(...))` instead of `EXPECT_TRUE(emptyCells.empty())` — it provides a better error message on failure.

velox/functions/prestosql/S2CellOperations.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <s2/s2polyline.h>
+#include <s2/s2region_coverer.h>
+
+namespace facebook::velox::functions {
+
+namespace {
+
+// Maps a GEOS coordinate (x=longitude, y=latitude) to a point on the
+// unit sphere.
+S2Point toS2Point(const geos::geom::Coordinate& coord) {
```
kgpai, line 43: nit: Please fix header related lints..

velox/functions/prestosql/S2CellOperations.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+// Converts a GEOS linear ring to an S2Loop. Skips the closing vertex
+// (GEOS rings repeat the first vertex; S2 loops are implicitly closed).
+// Normalizes the loop so it encloses the smaller of the two spherical
+// regions it defines.
+std::unique_ptr<S2Loop> toS2Loop(const geos::geom::LinearRing& ring) {
+  auto coords = ring.getCoordinates();
+  auto numVertices = coords->getSize() - 1;
+  std::vector<S2Point> points;
+  points.reserve(numVertices);
```
kgpai, line 55: nit: ensure numVertices is >=0

velox/functions/prestosql/S2CellOperations.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+Expected<std::vector<int64_t>> S2CellOp::tryCovering(
+    const geos::geom::Geometry& geom,
+    int32_t level) {
+  S2RegionCoverer::Options options;
+  options.set_min_level(level);
+  options.set_max_level(level);
+  S2RegionCoverer coverer(options);
+
+  std::vector<uint64_t> cellIds;
```
kgpai, line 203: nit: Will the vector of cellIds be signficant number , that might require it to be tracked ? If its not supposed to be very big , I suppose its fine, but otherwise might be need to use one of our allocators.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -297,6 +297,14 @@ std::unordered_set<std::string> skipFunctions = {
     "convex_hull_agg",
     "geometry_union_agg",
     "localtime",
+    // S2 functions are not available in the reference DB (DuckDB).
```
kgpai: While these functions may not be available in DDB, i dont think we use DDB for expression fuzzer testing (only aggregate testing ) - expression fuzzer uses presto as SOT (and if they are not available in Presto, then we should just add it to SOT skip list ). Removing it from here  will help us validate that the functions are battle hardened.

CMakeLists.txt
```diff
@@ -762,4 +762,11 @@ if(VELOX_ENABLE_GEO)
   velox_resolve_dependency(geos)
 endif()
 
+if(VELOX_ENABLE_GEO)
+  # S2 geometry library for geospatial functions.
+  # Always bundled since it is not commonly available as a system package.
+  set(s2geometry_SOURCE BUNDLED)
```
kgpai: Do not hardcode to BUNDLED, As this makes our builds flaky.

kgpai (commented):
It will be good to add an ArgValuesGenerator so that we produce non random values and test the functions out more thoroughly. Look at some examples like URLArgValuesGenerator etc. 

Another question is whether s2 functions will be used in some sort of spatial join ?

0xandr:
@kgpai will add an ArgValuesGenerator. Yes, functions will be used for spatial joins as well as dynamic roll-ups. E.g. `JOIN city_s2_cells ON s2_cell_id_contains(city_s2_cells.cell_id, object.s2_cell)` or `GROUP BY s2_cell_id_parent(s2_cell, level)` etc

CMake/resolve_dependency_modules/s2geometry.cmake
0xandr: s2geometry library does not, but the functions we add do, since we provide functions to translate a geometry into a set of s2 cell IDs

velox/functions/prestosql/S2CellOperations.cpp
0xandr: Don't mind adding it, but this is only called from appendCovering, where we ensure exterior ring size, and interior ring size is guaranteed by a valid GEOS LinearRing. Let me know how you'd like to porceed.

CMakeLists.txt
pratikpugalia: Thanks for the flag @kgpai, I Missed that this was not fixed on the second review.

@0xandr Please use velox_resolve_dependency and also  add install_s2geometry support in relevant scripts as listed in the previous comment [here](https://github.com/facebookincubator/velox/pull/15511#discussion_r3025131471). That should fix failing jobs. 
```
velox_set_source(s2geometry)
velox_resolve_dependency(geos)
```

velox/functions/prestosql/S2CellOperations.cpp
0xandr: The current callers write results into a Velox ArrayVector via result.add_item(), so I guess the data gets copied into tracked memory immediately? The temporary std::vector should be short-lived and freed when the function returns. Let me know if you'd like to pursue a different approach.

CMakeLists.txt
0xandr: tried to follow @pratikpugalia suggestion https://github.com/facebookincubator/velox/pull/15511#discussion_r3025131471

pratikpugalia: The CI failure is because the Docker images don't have s2geometry yet VELOX_DEPENDENCY_SOURCE=SYSTEM makes find_package hard-fail. Docker rebuild only happens on merge

Maybe Default s2geometry to AUTO/BUNDLED instead of velox_set_source and create a follow-up to velox_set_source once Docker images are rebuilt on merge to main.

Also patch question: Your BUNDLED resolver applies s2geometry-gcc12-max.patch but install_s2geometry in setup-common.sh doesn't, should the setup script also apply it for robustness?

cc: @kgpai

mbasmanova:
`S2CellOperations.h` — The `S2CellOp` methods (e.g., `areaSqKm`, `level`, `parent`) assume valid cell IDs but this isn't documented. Please add a note to the struct or method docs that callers must validate inputs via `isValid()` first, and that passing invalid cell IDs results in undefined behavior.

mbasmanova:
The `testThrows` lambda in `cellsDissolved` is used only twice and is a single-line body. Just inline the `VELOX_ASSERT_USER_THROW` calls directly.

velox/functions/prestosql/S2CellOperations.cpp
kgpai: As long as we are sure that this cant be negative etc, then thats fine (adding a debug check cant hurt though).

velox/functions/prestosql/S2CellOperations.cpp
kgpai: Sure, but we have seen cases where folks create temporary vectors of 10's hundred's of Mb's which still add to memory pressure and OOM, but since its not using our allocator doesnt get tracked. If these are supposed to be small, I suppose its not a problem.

scripts/setup-common.sh
```diff
@@ -280,6 +281,16 @@ function install_geos {
   fi
 }
 
+function install_s2geometry {
```
kgpai, line 284: Thanks for making these changes..

mbasmanova (approved):

Issue #15336, opened 2025-10-30
localArbitrationRunInParallelWithGlobalArbitration is flaky


xiaoxmeng:
### Bug description

455/458 Test #408: velox_parquet_e2e_filter_test ....................................................................   Passed  498.63 sec
456/458 Test #431: velox_aggregates_test ............................................................................   Passed  423.49 sec
457/458 Test #230: MockSharedArbitrationTest.localArbitrationRunInParallelWithGlobalArbitration .....................***Timeout 1500.01 sec
Note: Google Test filter = MockSharedArbitrationTest.localArbitrationRunInParallelWithGlobalArbitration
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from MockSharedArbitrationTest
[ RUN      ] MockSharedArbitrationTest.localArbitrationRunInParallelWithGlobalArbitration

458/458 Test #449: velox_exec_test ..................................................................................   Passed  3546.55 sec

99% tests passed, 1 tests failed out of 455

Total Test time (real) = 4158.83 sec

The following tests did not run:
	 99 - */MemoryAllocatorTest.allocContiguousVsize/* (Disabled)
	157 - */MemoryPoolTest.memoryLeakCheck/* (Disabled)
	291 - */SharedArbitrationTestWithThreadingModes.raceBetweenTaskTerminateAndReclaim/* (Disabled)

The following tests FAILED:
	230 - MockSharedArbitrationTest.localArbitrationRunInParallelWithGlobalArbitration (Timeout)
Errors while running CTest
Error: Process completed with exit code 8.

### System information

The failure comes from GCC build:

[Build with GCC / Ubuntu debug with resolve_dependency](https://github.com/facebookincubator/velox/actions/runs/18947002556/job/54101046963?fbclid=IwY2xjawNwjTFleHRuA2FlbQIxMQBicmlkETFDeTJ6YUFubXJ0Yk8zM0s3AR4Y_WglcrCVvKwJTm1f13Dn8YZ6PFwaN6F2WQEqEUTaJQL6ylkXq065UOlPAA_aem_1kIHZMppSbwy_VOODe7kkg#logs)

### Relevant logs

```bash

```

Jmc18134:
My PR #15518 has hit this same issue [here](https://github.com/facebookincubator/velox/actions/runs/19435642178/job/55622757885?pr=15518).

PR #15518, opened 2025-11-17
perf: Use FMT_COMPILE for float to string casting

Jmc18134:
Make the float to string casts use FMT_COMPILE, to avoid parsing the format string on every call. Also take a reference to the expected result, to avoid copying the string on every cast.

Together, these improve the performance of the basic benchmarks for real and double to scientific/standard notation casts.


Jmc18134:
The failing check seems to be an existing issue, #15336.

PR #15519, opened 2025-11-17
perf(decimal): Add helper for extracting decimal components efficiently

Jmc18134:
Move the logic for extracting integral and fractional parts from decimals to its own function with separate code paths for each scale. This lets the compiler apply strength reduction to division by the constant scale factors.

This improves the performance of the decimal cast basic benchmarks like cast_decimal_as_bigint and cast_decimal_to_inline_string.


Jmc18134:
To be clear about what the strength reduction is: When doing division by a constant divisor the compiler will often replace division instructions with multiplication by a precomputed reciprocal, which is usually much faster. Currently the divisions for getting decimal components aren't treated as constant (even though they are), which is what this patch fixes.

Issue #15512, opened 2025-11-15
Variant::toJson API lossy for ROW variant


ericcfu:
### Bug description

Variant::toJson() is lossy for ROW format, losing the entry names. The serialization is instead serialized as an ARRAY format, without the name of each entry. Instead, we should serialize with name-value pairs.

Not sure if we need backward compatibility, but the fix would be here https://github.com/facebookincubator/velox/blob/main/velox/type/Variant.cpp#L395.

Instead of only including the entry value, we could do something like.
`type.asRow().nameOf(i) + std::string(":") + row.at(i).toJson(type.childAt(i));`

### System information

.

### Relevant logs

```bash

```

PR #15524, opened 2025-11-17
fix: Non lossy json serialization for ROW variant

ericcfu:
Summary:
https://github.com/facebookincubator/velox/issues/15512

In this PR, we will change the JSON serialization of ROW variant to include the KV pair with the name of the field. This allows non lossy json serialization.

Differential Revision: D87148634

### DIFF
--- a/velox/type/Variant.cpp
+++ b/velox/type/Variant.cpp
@@ -16,6 +16,7 @@
 
 #include "velox/type/Variant.h"
 #include <cfloat>
+#include "folly/dynamic.h"
 #include "folly/json.h"
 #include "velox/common/base/BitUtil.h"
 #include "velox/common/encode/Base64.h"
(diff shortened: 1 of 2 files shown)


velox/type/Variant.cpp
```diff
@@ -392,7 +392,11 @@ std::string Variant::toJson(const Type& type) const {
         if (!first) {
           b += ",";
         }
+        b += "ROW{\"key\":";
```
mbasmanova: Let's use folly::dynamic to generate valid JSON.

velox/type/Variant.cpp
```diff
@@ -500,7 +504,7 @@ std::string Variant::toJsonUnsafe(const TypePtr& type) const {
         if (!first) {
           b += ",";
         }
-        b += "{\"key\":";
+        b += "ROW{\"name\":";
```
ericcfu: Need to revert this line, we only want this for ROW, no change to map.

PR #15527, opened 2025-11-17
Draft: calculate and use retained size with shared string buffers in local partition

kagamiori:
Differential Revision: D87259148

### DIFF
--- a/velox/exec/LocalPartition.cpp
+++ b/velox/exec/LocalPartition.cpp
@@ -16,7 +16,9 @@
 
 #include "velox/exec/LocalPartition.h"
 #include "velox/exec/Task.h"
+#include "velox/vector/DictionaryVector.h"
 #include "velox/vector/EncodedVectorCopy.h"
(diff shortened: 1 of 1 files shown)


velox/exec/LocalPartition.cpp
```diff
@@ -438,6 +440,120 @@ void LocalPartition::copy(
   target->copyRanges(input.get(), ranges);
 }
 
+template <TypeKind KIND>
+uint64_t retainedSizeWithSharedStringBuffers(
```
mbasmanova, line 444: Wondering if there is a way to avoid duplicating the logic of BaseVector::retainedSize and also make this code unit-testable.

CC: @bikramSingh91 @pedroerp @kgpai

mbasmanova: One option could be to extend BaseVector::retainedSize to produce the result and additional number of bytes used by the string buffers.

CC: @kagamiori

PR #15529, opened 2025-11-17
feat(cudf): Check signatures of aggregate functions before replacing with `cudf` variants

mattgara:
This PR builds on #15478 by checking for signatures of aggregate functions in velox and testing against their cuDF counterparts before replacing operators.


velox/experimental/cudf/expression/ExpressionEvaluator.cpp
```diff
@@ -914,6 +914,15 @@ bool registerBuiltinFunctions(const std::string& prefix) {
...
            .constantArgumentType("integer")
+           .build(),
+       FunctionSignatureBuilder()
+           .returnType("double")
+           .argumentType("double")
+           .build(),
+       FunctionSignatureBuilder()
+           .returnType("double")
+           .argumentType("double")
+           .constantArgumentType("integer")
```
devavret: These need to be removed as cudf::round_decimal does not support floating point types https://github.com/rapidsai/cudf/blob/1aedf672420ae11744f83bc0567322e0d18fadca/cpp/src/round/round.cu#L403-L405

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -960,4 +963,384 @@ bool CudfHashAggregation::isFinished() {
...
+  registerAggregationFunctionForStep(
+      prefix + "count", core::AggregationNode::Step::kSingle, countSignatures);
+  registerAggregationFunctionForStep(
+      prefix + "count", core::AggregationNode::Step::kPartial, countSignatures);
+  registerAggregationFunctionForStep(
+      prefix + "count", core::AggregationNode::Step::kFinal, countSignatures);
+  registerAggregationFunctionForStep(
+      prefix + "count",
+      core::AggregationNode::Step::kIntermediate,
+      countSignatures);
```
devavret: final and intermediate signatures are bingint -> bigint only

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -960,4 +963,384 @@ bool CudfHashAggregation::isFinished() {
...
+  // Partial step: avg(input_type) -> row(sum input_type, count bigint)
+  auto avgPartialSignatures = std::vector<exec::FunctionSignaturePtr>{
+      FunctionSignatureBuilder()
+          .returnType("row(smallint,bigint)")
+          .argumentType("smallint")
+          .build(),
+      FunctionSignatureBuilder()
+          .returnType("row(integer,bigint)")
+          .argumentType("integer")
+          .build(),
```
devavret, line 1236: first part of the struct might be bigint always. Since it stores a sum and the sum of any int type is stored as bigint in sum agg. Could you please confirm? I'm aware that return types are not matched yet for replacement consideration.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -960,4 +963,384 @@ bool CudfHashAggregation::isFinished() {
...
+  // Final step: avg(row(sum input_type, count bigint)) -> double
+  auto avgFinalSignatures = std::vector<exec::FunctionSignaturePtr>{
+      FunctionSignatureBuilder()
+          .returnType("double")
+          .argumentType("row(smallint,bigint)")
+          .build(),
+      FunctionSignatureBuilder()
+          .returnType("double")
+          .argumentType("row(integer,bigint)")
+          .build(),
```
devavret: likewise, I think these never happen. Could you please confirm?

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -960,4 +963,384 @@ bool CudfHashAggregation::isFinished() {
...
+  registerAggregationFunctionForStep(
+      prefix + "avg", core::AggregationNode::Step::kFinal, avgFinalSignatures);
+
+  // Intermediate step: avg(row(sum input_type, count bigint)) -> row(sum
+  // input_type, count bigint)
+  auto avgIntermediateSignatures = std::vector<exec::FunctionSignaturePtr>{
+      FunctionSignatureBuilder()
+          .returnType("row(double,bigint)")
+          .argumentType("row(double,bigint)")
+          .build()};
```
devavret, line 1269: Missing bigint sum type?

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -960,4 +963,384 @@ bool CudfHashAggregation::isFinished() {
...
+  for (const auto& groupingKey : groupingKeys) {
+    auto expandedKey = expandFieldReference(groupingKey, sourceNode);
+    std::vector<core::TypedExprPtr> exprs = {expandedKey};
+    if (!canBeEvaluatedByCudf(exprs, queryCtx)) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
devavret, line 1411: What does this do? Can grouping keys be expressions themselves?

velox/experimental/cudf/expression/ExpressionEvaluator.cpp
mattgara: I'm actually not sure how this snuck in as I've never actually modified this code myself, I think this was off an old version of a base branch, and should not be here! Good catch, removing this.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
mattgara: > first part of the struct might be bigint always. Since it stores a sum and the sum of any int type is stored as bigint in sum agg. Could you please confirm?

I don't believe it is `bigint`. See: https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/AverageAggregate.cpp#L30 and https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/AverageAggregate.cpp#L38-L45

Based on this for ints it should be double. I will update this to use the double only signatures.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
mattgara: Yes, I believe that is true, I will also swap these to use double only.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
mattgara: Based on the same reference above (see https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/AverageAggregate.cpp#L30) I believe this only uses double sum type.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
mattgara: This expands grouping key field refs back to their underlying projection expressions so we can check cuDF support when the group‑by keys are derived within a project node, which appears to be the expected behaviour for some types of queries. 

For example, see the [test](https://github.com/facebookincubator/velox/pull/15529/files#diff-c7710e4012be844ab2b80dcbb9904cbff25b833431a5d4fb63dbcc7e357b80b4R220) added in this PR that demonstrates this case. The test uses a query of the form 
```
SELECT c0 + c1 AS key1, length(c6) AS key2, sum(c2) FROM tmp GROUP BY key1, key2
```
where `c0 + c1` is an expression that is aliased as `key1` and `length(c6)` is aliased as `key2`. 

Without expanding the grouping key field refs back to their projection expressions, we’d only see key1/key2 as field refs and miss that they’re `c0 + c1` and `length(c6)`.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -977,4 +980,434 @@ bool CudfHashAggregation::isFinished() {
...
+          .returnType("row(double,bigint)")
+          .argumentType("integer")
+          .build(),
+      FunctionSignatureBuilder()
+          .returnType("row(double,bigint)")
+          .argumentType("bigint")
+          .build(),
+      FunctionSignatureBuilder()
+          .returnType("row(double,bigint)")
+          .argumentType("real")
```
jinchengchenghh, line 1243: This is the Spark and Presto data type difference, especially in Velox Presto, AVG(FLOAT) Intermediate type is double as Spark, not presto.

Function | Engine | Intermediate | Return
-- | -- | -- | --
COUNT | Both | BIGINT | BIGINT
SUM(INT) | Both | BIGINT | BIGINT
SUM(FLOAT) | Spark | DOUBLE | DOUBLE
SUM(FLOAT) | Presto | REAL | REAL
SUM(DECIMAL) | Spark | Widened | Widened
SUM(DECIMAL) | Presto | Same precision | Same precision
AVG(INT) | Both | BIGINT → DOUBLE | DOUBLE
AVG(FLOAT) | Spark | DOUBLE | DOUBLE
AVG(FLOAT) | Presto | REAL | REAL
AVG(DECIMAL) | Spark | Widened | Widened
AVG(DECIMAL) | Presto | Narrower | Widened

I will create a PR to register Spark aggregate function after this PR https://github.com/facebookincubator/velox/pull/16524 landing CC @devavret @karthikeyann

jinchengchenghh:
And the hash join filter also needs validation

PR #15532, opened 2025-11-18
fix: Max length estimation in DateTimeFormatter

bikramSingh91:
Summary:
Fixes the maxLength() utility function in DateTimeFormatter for TIMEZONE specifiers that print out the timezone abbreviation, that is, for 'z', 'zz' and 'zzz'. It increases the size for the case where some timezones do not have abbreviations but instead use GMT offsets like `GMT+05:00`.

Originally found by the expression fuzzer.

Differential Revision: D87290122


bikramSingh91:
Fixes: #15557

PR #15534, opened 2025-11-18
fix: Trim numeric suffix when casting string to real/double

### LINKED ISSUE #15533: Result mismatch with spark when casting string with numeric suffixes to double
### Bug description

Forwarded issue from: https://github.com/apache/incubator-gluten/issues/11082

test sql:

```
select cast('00123d' as double), cast('123f' as double);
```

spark result: `123.0, 123.0`
gluten result: `NULL, NULL`

In Java, when parsing string to float/double, the opt [FfDd]? suffix will be skipped.

https://github.com/openjdk/jdk/blob/f4305923fb6203089fd13cf3387c81e127ae5fe2/src/java.base/share/classes/jdk/internal/math/FloatingDecimal.java#L1409-L1412

### System information

spark 3.5 + gluten
centos7

### Relevant logs

```bash

```

wForget:
Makes the result of `casting string with numeric suffix to real/double` consistent with spark.

In Java, when parsing string to float/double, the opt [FfDd]? suffix will be skipped.
https://github.com/openjdk/jdk/blob/f4305923fb6203089fd13cf3387c81e127ae5fe2/src/java.base/share/classes/jdk/internal/math/FloatingDecimal.java#L1409-L1412

closes #15533

### DIFF
--- a/velox/expression/PrestoCastHooks.cpp
+++ b/velox/expression/PrestoCastHooks.cpp
@@ -127,12 +127,14 @@ Expected<T> doCastToFloatingPoint(const StringView& data) {
 
 Expected<float> PrestoCastHooks::castStringToReal(
     const StringView& data) const {
-  return doCastToFloatingPoint<float>(data);
+  auto trimmed = util::trimNumericSuffix(data);
+  return doCastToFloatingPoint<float>(trimmed);
 }
 
 Expected<double> PrestoCastHooks::castStringToDouble(
     const StringView& data) const {
-  return doCastToFloatingPoint<double>(data);
+  auto trimmed = util::trimNumericSuffix(data);
+  return doCastToFloatingPoint<double>(trimmed);
 }
 
 StringView PrestoCastHooks::removeWhiteSpaces(const StringView& view) const {
--- a/velox/expression/PrestoCastHooks.cpp
+++ b/velox/expression/PrestoCastHooks.cpp
@@ -127,14 +127,12 @@ Expected<T> doCastToFloatingPoint(const StringView& data) {
 
 Expected<float> PrestoCastHooks::castStringToReal(
     const StringView& data) const {
-  auto trimmed = util::trimNumericSuffix(data);
-  return doCastToFloatingPoint<float>(trimmed);
+  return doCastToFloatingPoint<float>(util::trimFloatSuffix(data));
 }
 
 Expected<double> PrestoCastHooks::castStringToDouble(
     const StringView& data) const {
-  auto trimmed = util::trimNumericSuffix(data);
-  return doCastToFloatingPoint<double>(trimmed);
+  return doCastToFloatingPoint<double>(util::trimFloatSuffix(data));
 }
 
 StringView PrestoCastHooks::removeWhiteSpaces(const StringView& view) const {
--- a/velox/expression/tests/CastExprTest.cpp
+++ b/velox/expression/tests/CastExprTest.cpp
@@ -1214,6 +1214,8 @@ TEST_F(CastExprTest, primitiveValidCornerCases) {
     testCast<std::string, float>("real", {"1.7E308"}, {kInf});
     testCast<std::string, float>("real", {"1."}, {1.0});
     testCast<std::string, float>("real", {"1"}, {1});
+    testCast<std::string, float>("real", {"1.2f"}, {1.2});
+    testCast<std::string, float>("real", {"1.2d"}, {1.2});
     testCast<std::string, float>("real", {"  1  "}, {1});
     testCast<std::string, float>("real", {"-Infinity"}, {-kInf});
     testCast<std::string, float>("real", {"+Infinity"}, {kInf});
--- a/velox/expression/tests/CastExprTest.cpp
+++ b/velox/expression/tests/CastExprTest.cpp
@@ -1223,6 +1223,8 @@ TEST_F(CastExprTest, primitiveValidCornerCases) {
     testCast<std::string, float>("real", {"  NaN  "}, {kNan});
     testCast<std::string, float>("real", {"  -NaN  "}, {kNan});
     testCast<std::string, float>("real", {"  +NaN  "}, {kNan});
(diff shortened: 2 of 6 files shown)


velox/functions/sparksql/specialforms/SparkCastHooks.cpp
```diff
@@ -110,13 +110,25 @@ Expected<int32_t> SparkCastHooks::castStringToDate(
...
-  return util::Converter<TypeKind::REAL>::tryCast(data);
+  auto trimmed = trimNumericSuffix(data);
+  return util::Converter<TypeKind::REAL>::tryCast(trimmed);
 }
 
 Expected<double> SparkCastHooks::castStringToDouble(
     const StringView& data) const {
-  return util::Converter<TypeKind::DOUBLE>::tryCast(data);
+  auto trimmed = trimNumericSuffix(data);
+  return util::Converter<TypeKind::DOUBLE>::tryCast(trimmed);
```
jinchengchenghh: Please update this change to common behavior, presto also aligns with this behavior
```
presto:default> select cast('123.0d' as double);
 _col0 
-------
 123.0 
(1 row)

Query 20251118_103153_00007_2tw34, FINISHED, 1 node
Splits: 17 total, 17 done (100.00%)
[Latency: client-side: 67ms, server-side: 51ms] [0 rows, 0B] [0 rows/s, 0B/s]
```

wForget: Thanks @jinchengchenghh, I have created a new common method in Conversions and also changed the cast logic of presto. Could you take another look?

velox/expression/tests/CastExprTest.cpp
```diff
@@ -1214,6 +1214,8 @@ TEST_F(CastExprTest, primitiveValidCornerCases) {
     testCast<std::string, float>("real", {"1.7E308"}, {kInf});
     testCast<std::string, float>("real", {"1."}, {1.0});
     testCast<std::string, float>("real", {"1"}, {1});
+    testCast<std::string, float>("real", {"1.2f"}, {1.2});
```
jinchengchenghh, line 1217: please also ad the double test

velox/expression/PrestoCastHooks.cpp
```diff
@@ -127,12 +127,14 @@ Expected<T> doCastToFloatingPoint(const StringView& data) {
 
 Expected<float> PrestoCastHooks::castStringToReal(
     const StringView& data) const {
-  return doCastToFloatingPoint<float>(data);
+  auto trimmed = util::trimNumericSuffix(data);
```
jinchengchenghh: doCastToFloatingPoint<float>(util::trimNumericSuffix(data))

velox/type/Conversions.h
```diff
@@ -248,6 +248,9 @@ struct Converter<TypeKind::BOOLEAN, void, TPolicy> {
 /// a StringView of the trimmed string.
 std::string_view trimWhiteSpace(const char* data, size_t length);
 
+/// Removes trailing numeric type suffix ('f', 'F', 'd', 'D') from the input.
```
jinchengchenghh: trimNumericSuffix -> trimFloatSuffix, also update the comment. int with d is not supported, right?

wForget: > int with d is not supported, right?

yes, `CAST(12d AS INT)` returns `NULL` on spark 3.5.

image

velox/expression/tests/CastExprTest.cpp
```diff
@@ -1214,13 +1214,17 @@ TEST_F(CastExprTest, primitiveValidCornerCases) {
...
+    testCast<std::string, float>("real", {"1.2d"}, {1.2});
     testCast<std::string, float>("real", {"  1  "}, {1});
     testCast<std::string, float>("real", {"-Infinity"}, {-kInf});
     testCast<std::string, float>("real", {"+Infinity"}, {kInf});
     testCast<std::string, float>("real", {" Infinity "}, {kInf});
     testCast<std::string, float>("real", {"  NaN  "}, {kNan});
     testCast<std::string, float>("real", {"  -NaN  "}, {kNan});
     testCast<std::string, float>("real", {"  +NaN  "}, {kNan});
+    testCast<std::string, double>("double", {"1.2f"}, {1.2});
```
jinchengchenghh, line 1228: Please add the test for 'F' and 'D'

wForget: Thanks, added

jinchengchenghh (approved):

velox/type/Conversions.h
```diff
@@ -248,6 +248,10 @@ struct Converter<TypeKind::BOOLEAN, void, TPolicy> {
 /// a StringView of the trimmed string.
 std::string_view trimWhiteSpace(const char* data, size_t length);
 
+/// Removes trailing floating-point type suffix ('f', 'F', 'd', 'D') from the
+/// input.
+StringView trimFloatSuffix(const StringView& view);
```
rui-mo, line 253: nit: StringView is fixed-width, and could just be passed by value.

velox/type/Conversions.cpp
```diff
@@ -99,4 +99,16 @@ std::string_view trimWhiteSpace(const char* data, size_t length) {
   return std::string_view(data + startIndex, endIndex - startIndex + charSize);
 }
 
+/// Removes trailing floating-point type suffix ('f', 'F', 'd', 'D') from the
+/// input.
+StringView trimFloatSuffix(const StringView& view) {
+  if (!view.empty()) {
+    char last = *(view.end() - 1);
+    if (last == 'f' || last == 'F' || last == 'd' || last == 'D') {
```
rui-mo, line 107: nit: FOLLY_UNLIKELY

rui-mo (commented):
Thanks! Just two nits.

velox/type/Conversions.h
Yuhta, line 253: We should inline this function since in most of the cases it's no-op

PR #15536, opened 2025-11-18
fix: Change count metric from signed to unsigned (int64_t -> uint64_t)

rui-mo:
Use an unsigned type for the count metric while keeping int64_t for value 
metrics. When the unit is kNone, negative values can be valid such as for delta-
type metrics, so int64_t remains appropriate. In contrast, `count` should 
always be non-negative. This PR also addresses potential overflow when 
converting unsigned metrics to `RuntimeMetrics`.

### DIFF
--- a/velox/common/base/RuntimeMetrics.h
+++ b/velox/common/base/RuntimeMetrics.h
@@ -23,6 +23,13 @@
 
 namespace facebook::velox {
 
+/// Converts unsigned bigint to signed, capping at int64_t max if overflow
+/// happens. Could be replaced by 'std::saturate_cast' since C++26.
+inline int64_t saturateCast(uint64_t value) {
+  return static_cast<int64_t>(std::min(
+      value, static_cast<uint64_t>(std::numeric_limits<int64_t>::max())));
+}
+
 struct RuntimeCounter {
   enum class Unit { kNone, kNanos, kBytes };
   int64_t value;
@@ -36,7 +43,7 @@ struct RuntimeMetric {
   // Sum, min, max have the same unit, count has kNone.
   RuntimeCounter::Unit unit;
   int64_t sum{0};
-  int64_t count{0};
+  uint64_t count{0};
   int64_t min{std::numeric_limits<int64_t>::max()};
   int64_t max{std::numeric_limits<int64_t>::min()};
 
@@ -51,7 +58,7 @@ struct RuntimeMetric {
 
   explicit RuntimeMetric(
       int64_t _sum,
-      int64_t _count,
+      uint64_t _count,
       int64_t _min,
       int64_t _max,
       RuntimeCounter::Unit _unit = RuntimeCounter::Unit::kNone)
--- a/velox/common/base/tests/RuntimeMetricsTest.cpp
+++ b/velox/common/base/tests/RuntimeMetricsTest.cpp
@@ -24,7 +24,7 @@ class RuntimeMetricsTest : public testing::Test {
   static void testMetric(
       const RuntimeMetric& rm1,
       int64_t expectedSum,
-      int64_t expectedCount,
+      uint64_t expectedCount,
       int64_t expectedMin = std::numeric_limits<int64_t>::max(),
       int64_t expectedMax = std::numeric_limits<int64_t>::min()) {
     EXPECT_EQ(expectedSum, rm1.sum);
@@ -82,6 +82,21 @@ TEST_F(RuntimeMetricsTest, basic) {
   ASSERT_EQ(
       timeRm.toString(),
       "sum:2.00us, count:1, min:2.00us, max:2.00us, avg: 2.00us");
+}
+
+TEST_F(RuntimeMetricsTest, saturateCast) {
+  auto maxUint64 = std::numeric_limits<uint64_t>::max();
+  RuntimeMetric rm{
+      saturateCast(maxUint64),
+      maxUint64,
+      saturateCast(maxUint64),
+      saturateCast(maxUint64)};
+
+  auto maxInt64 = std::numeric_limits<int64_t>::max();
+  EXPECT_EQ(rm.sum, maxInt64);
+  EXPECT_EQ(rm.count, maxUint64);
+  EXPECT_EQ(rm.min, maxInt64);
+  EXPECT_EQ(rm.max, maxInt64);
 }
 
 } // namespace facebook::velox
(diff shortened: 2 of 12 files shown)


majetideepak:
@rui-mo I see a test failing `[  FAILED  ] RuntimeMetricsTest.basic (0 ms)`.

@tanjialiang, @xiaoxmeng Do you have any thoughts on this PR?

rui-mo:
Thanks @majetideepak. The test should be fixed now.

velox/common/base/tests/RuntimeMetricsTest.cpp
```diff
@@ -25,8 +25,8 @@ class RuntimeMetricsTest : public testing::Test {
       const RuntimeMetric& rm1,
       int64_t expectedSum,
       int64_t expectedCount,
-      int64_t expectedMin = std::numeric_limits<int64_t>::max(),
-      int64_t expectedMax = std::numeric_limits<int64_t>::min()) {
+      int64_t expectedMin = std::numeric_limits<uint64_t>::max(),
+      int64_t expectedMax = std::numeric_limits<uint64_t>::min()) {
```
tanjialiang: assigning a uint64 max to int64

rui-mo: Thanks for catching this. I double-checked those to ensure unexpected overflow would not occur.

tanjialiang:
It just came to my mind: Should we keep value signed? When Unit is kNone, we might expect legit negative value metrics such as delta type of metrics.

counts should be non-negative this should be strictly true.

rui-mo:
@tanjialiang If it’s necessary to keep the value as `int64_t`, I’m OK with doing so. In that case, we’ll need special handling for range checks when passing uint64_t metrics to RuntimeMetrics. For example, IoCounter uses uint64_t, which could overflow an int64_t.

tanjialiang:
> @tanjialiang If it’s necessary to keep the value as `int64_t`, I’m OK with doing so. In that case, we’ll need special handling for range checks when passing uint64_t metrics to RuntimeMetrics. For example, IoCounter uses uint64_t, which could overflow an int64_t.

For most cases, we should be good. But yeah for the ones we already know that is doing a uint64 to int64 assignment we can add impose range checks. (and it should be okay if we miss some. most cases they won't overflow)

rui-mo:
@tanjialiang I’ve updated the PR based on your suggestion. `uint64_t` is now used only for count, and additional range checks have been added to prevent potential overflow. Please take another look when you have a moment. Thanks!

velox/common/base/RuntimeMetrics.h
```diff
@@ -22,6 +22,11 @@
 
 namespace facebook::velox {
 
+inline int64_t unsignedToSigned(uint64_t value) {
```
tanjialiang: Should we just use folly::to to enforce it does not overflow? I'm concerned The misuse of this method might introduce data corruption issues.

rui-mo: This should prevent overflow, as `std::min(value, INT64_MAX)` caps the value at `INT64_MAX`. Does this make sense? Thanks.

rui-mo:
Hi @tanjialiang, do you have any further comments? Thanks!

rui-mo:
Hi @tanjialiang, could you please take another look at your convenience? Thanks.

rui-mo:
Hi @Yuhta @tanjialiang, this PR fixes the overflow issue after propagating IO stats to runtime metrics, see https://github.com/facebookincubator/velox/pull/15408#issuecomment-3543451841. Could you please take further review? Thanks.

velox/common/base/RuntimeMetrics.h
Yuhta: There is a name for this called saturate_cast: https://en.cppreference.com/w/cpp/numeric/saturate_cast.html
If we use this name, it should not be misused.

rui-mo: @Yuhta Thanks for the helpful insight. I’ve renamed the function. Could you please take another look?
```
/// Converts unsigned bigint to signed, capping at int64_t max if overflow
/// happens. Could be replaced by 'std::saturate_cast' since C++26.
inline int64_t saturateCast(uint64_t value) {
  return static_cast<int64_t>(std::min(
      value, static_cast<uint64_t>(std::numeric_limits<int64_t>::max())));
}
```

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -1159,7 +1159,7 @@ CudfVectorPtr CudfHashAggregation::releaseAndResetPartialOutput() {
         std::string(exec::HashAggregation::kFlushTimes), RuntimeCounter(1));
     lockedStats->addRuntimeStat(
         std::string(exec::HashAggregation::kPartialAggregationPct),
-        RuntimeCounter(aggregationPct));
+        RuntimeCounter(static_cast<int64_t>(aggregationPct)));
```
Yuhta: Should this one be saturate cast as well?

rui-mo: I assumed `aggregationPct` would typically be <= 100, so I used static_cast. I've updated it to use a saturate cast to prevent any potential risk. Thanks.

peterenescu:
Hi @rui-mo you will also need to make a small change to presto https://github.com/prestodb/presto/pull/27276. We will probably need to: 
1. Add compatibility layer
2. Merge the Velox code
3. Update Presto side
4. Remove the compatibility layer in Velox
cc: @Yuhta

rui-mo:
@peterenescu Thanks for your help. I understand - that’s because RuntimeMetric in Presto also uses int64_t for count.

```
namespace facebook::presto::protocol {
struct RuntimeMetric {
  String name = {};
  RuntimeUnit unit = {};
  int64_t sum = {};
  int64_t count = {};
  int64_t max = {};
  int64_t min = {};
};
```

rui-mo:
@peterenescu Could you please let me know what I need to do to "add compatibility layer" in Velox?

Yuhta:
@rui-mo Maybe we change the presto side first to be `uint64_t`?  Will this error go away?

rui-mo:
Hi @peterenescu @Yuhta, based on the review comment from https://github.com/prestodb/presto/pull/27295#discussion_r2912834680, it seems that we could merge this PR into Velox first and apply the corresponding changes in Presto afterward. I would appreciate your thoughts on this approach. Thank you!

Yuhta:
@rui-mo The merge into Meta internal codebase is blocked unless the Presto side is fixed first.  If the changed code on Presto side works with both before & after with Velox, we should change Presto first; otherwise we may need to use `VELOX_ENABLE_BACKWARD_COMPATIBILITY`

PR #15539, opened 2025-11-18
fix(cudf): Fix empty vector handling in FilterProject and HashAggregation

devavret:
`cudf::table` does not handle the case where the data is empty but size is non-zero. This resulted in a bug when running queries with a count(*) with presto-cudf because the plan contains a Project node with no projections. Sample broken query: 
```sql
-- TPC-DS SF100
SELECT count(*) 
FROM date_dim 
WHERE d_month_seq BETWEEN 1200 AND 1211;
-- Expected: 366
-- Actual: 0 
```
This PR disables cuDF on Project and Aggregate plan nodes which expect to receive empty tables and on Project which expects to produce empty results.

### DIFF
--- a/velox/experimental/cudf/exec/ToCudf.cpp
+++ b/velox/experimental/cudf/exec/ToCudf.cpp
@@ -127,8 +127,9 @@ bool CompileState::compile(bool allowCpuFallback) {
       if (projectPlanNode) {
         if (projectPlanNode->sources()[0]->outputType()->size() == 0 ||
             projectPlanNode->outputType()->size() == 0) {
-          canBeEvaluated = false;
+          return false;
         }
+      }
 
       // Check filter separately.
       if (filterNode) {


velox/experimental/cudf/exec/ToCudf.cpp
```diff
@@ -123,6 +123,12 @@ bool CompileState::compile(bool allowCpuFallback) {
       auto projectPlanNode = std::dynamic_pointer_cast<const core::ProjectNode>(
           getPlanNode(filterProjectOp->planNodeId()));
       auto filterNode = filterProjectOp->filterNode();
+      bool canBeEvaluated = true;
+      if (projectPlanNode) {
+        if (projectPlanNode->sources()[0]->outputType()->size() == 0 ||
+            projectPlanNode->outputType()->size() == 0) {
+          canBeEvaluated = false;
+        }
```
karthikeyann:
```suggestion
          return false;
        }
      }
```

karthikeyann: @devavret please review this change.

devavret: This is now correct

PR #15544, opened 2025-11-18
feat: Initial support for Claude 

kgpai:
With this change we are hoping to leverage Claude for initial code reviews on PR's. At this moment the idea is to use this on a trial basis, monitor usage and cost. As we get more comfortable and understand its limitations and cost structure etc, we might consider expanding its usage and capabilities.

### DIFF
--- a/.github/workflows/preliminary_checks.yml
+++ b/.github/workflows/preliminary_checks.yml
@@ -23,8 +23,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write
-  id-token: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.repository }}-${{ github.head_ref || github.sha }}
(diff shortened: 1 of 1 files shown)


assignUser:
This is not possible in a safe way as you need the Claude API key but `pull_request` doesn't provide that for fork PRs (so they can't steal it :ninja: !). 

There is a trigger that would make this possible but would be unsafe to the max, leading to compromise of the API key at the minimum.

Sorry :( Github copilot might be a solution due to the different integration but even there I am not sure on the security implications given prompt injection (I haven't looked what permission that requires). There is also https://www.coderabbit.ai/pricing which seems to be free for open source projects but I don't know anything else about it.

CLAUDE.md
```diff
@@ -0,0 +1,458 @@
...
+- avx, avx2, sse (Intel) or Neon/Neon64 (ARM)
+
+**Using Clang on Linux:**
+```bash
+export USE_CLANG=true
+./scripts/setup-ubuntu.sh  # or setup-centos9.sh
+export CC=/usr/bin/clang-15
+export CXX=/usr/bin/clang++-15
+make
+```
```
MBkkt, line 458: is it still right? :/

aditi-pandit:
@kgpai : Presto uses Sourcery AI and it is quite helpful. https://www.sourcery.ai/ Maybe we could try it ?

PR #15545, opened 2025-11-18
feat: Add map to struct pushdown support to file readers

Yuhta:
Summary: This is a fallback mode to read flatmap as struct.  It is not used in Presto but needed in other ML compute engines.

Differential Revision: D87095482


xiaoxmeng (approved):
@Yuhta LGTM. Thanks!

PR #15548, opened 2025-11-18
fix: Do not request data size if there is only  a single source as in persistent shuffle and add query config

han-yan01:
Summary:
Sequel of D86278872 and D85647156
This diff's addition:
- implemented cleaner exposure of nomoresources through exchangequeue with api access, tsan fix
- add query config to enable this flag in sapphire velox (default: disabled)
- note that in production this diff won't be in effect as it is disabled unless otherwise enabled in query config

optimization logic in D85647156
no more sources flag setting logic credit D86278872

Differential Revision: D87038056


xiaoxmeng (approved):
@han-yan01 thanks!

Issue #15486, opened 2025-11-13
Mismatched expression evaluation for switch block


pramodsatya:
### Description

The following `integer overflow` was seen in `array_sum` function with expression fuzzer. 
An `InputGenerator` for `array_sum` say `ArraySumInputGenerator`, should be added that is similar to `MapUnionSumInputGenerator`, in order to ensure the values in the integer array [do not overflow](https://github.com/facebookincubator/velox/blob/2b51a3cdd081ff6b9a729a86996edc664d7b2c40/velox/functions/prestosql/fuzzer/MapUnionSumInputGenerator.h#L32) when added. 
The custom input generators for functions should also be wired into the expression fuzzer.

Alternatively, `array_sum` can be added to the list of skip functions in [`ExpressionFuzzerTest.cpp`](https://github.com/facebookincubator/velox/blob/2b51a3cdd081ff6b9a729a86996edc664d7b2c40/velox/expression/fuzzer/ExpressionFuzzerTest.cpp#L127). 

See https://github.com/facebookincubator/velox/pull/15399/#issuecomment-3523935299 for fuzzer repro path  to reproduce the error locally.

cc: @peterenescu @mbasmanova @aditi-pandit @czentgr 

### Error Reproduction

velox_expression_runner_test  --fuzzer_repro_path="/Users/pramod/Downloads/tmp/coalesce_repro/velox_expressionVerifier_5gRpp9" --sql_path="/Users/pramod/Downloads/tmp/coalesce_repro/velox_expressionVerifier_5gRpp9/sql" --mode=verify

### Relevant logs

```bash
[...]

PR #15549, opened 2025-11-18
fix: Unregister special forms SWITCH rewrite

peterenescu:
Summary:
SWITCH rewrite has correctness issues identified by expression fuzzer and causes behavior mismatch.

More information and investigation can be found here: https://github.com/facebookincubator/velox/issues/15486

Differential Revision: D87355432

### DIFF
--- a/velox/expression/tests/ExprOptimizerTest.cpp
+++ b/velox/expression/tests/ExprOptimizerTest.cpp
@@ -172,10 +172,6 @@ TEST_F(ExprOptimizerTest, lambdas) {
       "reduce(c0, (89.0E0 + 11.0E0), (s, x) -> s + x * (0.2E0 - 0.1E0), s -> (s < (100.0E0 + 1.0E0)))",
       "reduce(c0, 100.0E0, (s, x) -> s + x * 0.1E0, s -> (s < 101.0E0))",
       ROW({"c0"}, {ARRAY(DOUBLE())}));
-  testExpression(
-      "reduce(c0, 1 - 1, (s, x) -> s + x, s -> coalesce(s, abs(2 - 10)) * (5 * 2))",
-      "reduce(c0, 0, (s, x) -> s + x, s -> coalesce(s, 8) * 10)",
-      ROW({"c0"}, {ARRAY(SMALLINT())}));
 }
 
 TEST_F(ExprOptimizerTest, rewritesWithConstantFolding) {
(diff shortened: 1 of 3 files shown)


pramodsatya:
@peterenescu, could you please retain `SwitchRewrite.h/.cpp`? We are dependent on this rewrite in Presto C++ for optimizing Presto RowExpressions with the native engine (ref: https://github.com/prestodb/presto/pull/22927). The rewritten expression will be converted to a Presto RowExpression and it will **not** be compiled/evaluated in Velox.
cc: @czentgr @aditi-pandit @majetideepak

czentgr:
@pramodsatya If there is a correctness problem it won't work for PrestoC++ either. We have to investigate what the correctness issues are. I see you opened the issue. Looks to be a type issue where the fuzzer generates two large integer literals but on constant folding the resulting value would overflow the type. I would say this is expected and a technically a user error that the fuzzer should skip, no?

@peterenescu Instead of fully reverting it, can we just not register the rewrite when registering the special form? That way the changes are minimal and we still have an option to register it separately in PrestoC++?

peterenescu:
@czentgr The issue is not with the fuzzer but a behavior change that was introduced by the switch rewrite. More information can be found here: https://github.com/facebookincubator/velox/issues/15486#issuecomment-3555149199.

@pramodsatya As the investigation above identifies, there is a meaningful behavioral change that _will_ affect expression evaluation in Presto C++ as it uses Velox **_and_** any ongoing efforts in prestodb/presto#22927 to use Velox constant folding and expression rewrite in coordinator. Therefore, it would be prudent to test and certify this behavior change before we introduce it to Presto.

To clarify, we were hoping to revert _only_ SWITCH while we test and verify internally and add it back hopefully by next week.

pramodsatya:
Thanks @czentgr, @peterenescu. I agree the `SWITCH` rewrite introduces behavior change that will affect any expression evaluated in Presto C++. 
Clarifying a bit more about why we need this rewrite in Presto C++:
1. We intend to apply the `SWITCH` rewrite only via the Velox `ExprOptimizer`. The intent is to optimize a Presto expression using the existing Velox `ExprOptimizer` (constant folding + rewrites) and return back an optimized Presto expression to the coordinator. In the linked Presto PR, we are only adding the requisite Velox -> Presto expression conversion.
2. The Velox `TypedExpr` optimized via the `ExprOptimizer` will **not** be compiled and it will not be evaluated on any input data, mitigating the risk due to correctness issue that has been identified. Instead, the optimized `TypedExpr` will be converted back to a Presto expression and returned to the coordinator. Presto java already performs this optimization for `SWITCH`, so this rewrite will give us functional parity with Presto for expression optimization in C++.
3. All these changes will be made in the Presto C++ sidecar, which is an opt-in via config (`native-sidecar=false` by default). We are also adding a new endpoint `v1/expressions` for this change, so this will not affect **any** existing Presto C++ functionality.

We **should** definitely avoid applying `SWITCH` rewrite in Velox until the issue is resolved. But retaining `SwitchRewrite.h/.cpp` would give us the flexibility to test/add this in Presto C++ for our usecase without duplicating effort. Please let me know what you think.

velox/expression/tests/ExprOptimizerTest.cpp
```diff
@@ -172,10 +172,6 @@ TEST_F(ExprOptimizerTest, lambdas) {
       "reduce(c0, (89.0E0 + 11.0E0), (s, x) -> s + x * (0.2E0 - 0.1E0), s -> (s < (100.0E0 + 1.0E0)))",
       "reduce(c0, 100.0E0, (s, x) -> s + x * 0.1E0, s -> (s < 101.0E0))",
       ROW({"c0"}, {ARRAY(DOUBLE())}));
-  testExpression(
```
bikramSingh91: not all of these tests rely on the switch rewrite, can you add those back?

velox/expression/tests/SwitchRewriteTest.cpp
```diff
@@ -25,25 +25,36 @@ class SwitchRewriteTest : public expression::test::SpecialFormRewriteTestBase {
 };
 
 TEST_F(SwitchRewriteTest, basic) {
```
bikramSingh91: lets remove this whole test since its not adding any value till the switch rewrite is added.

velox/expression/tests/ExprOptimizerTest.cpp
```diff
@@ -323,8 +286,8 @@ TEST_F(ExprOptimizerTest, makeFailExpr) {
   // Primitive types.
   assertPrestoFailExpr("0 / 0", prestoFailCall(), ROW({}));
   assertPrestoFailExpr(
-      "if(a = 2 * 2, a / abs(-1 * 3), 0 / 0)",
-      fmt::format("if(a = 4, a / 3, {})", prestoFailCall()),
+      "if(false, a * abs(-1 * 3), 0 / 0)",
```
bikramSingh91: similarly, over here, lets remove this since its not adding any test coverage unless switch rewrite is added

bikramSingh91:
Update on next steps: https://github.com/facebookincubator/velox/issues/15486#issuecomment-3561050649

TLDR: we are closing the PR and will work on enhancing the fuzzer to handle this case

PR #15429, opened 2025-11-06
feat: Add KHyperLogLog Util functions

HeidiHan0000:
Summary:
Adding khll util functions as per the Java implementation. (https://github.com/prestodb/presto/tree/master/presto-main-base/src/main/java/com/facebook/presto/type/khyperloglog)

These will be used by KHLL udfs from https://prestodb.io/docs/current/functions/khyperloglog.html

Differential Revision: D86235906

PR #15550, opened 2025-11-18
feat: Add KHyperloglog Scalar UDFs

HeidiHan0000:
Summary:
Adds the scalar udfs of the khyperloglog type, based on the utils from https://github.com/facebookincubator/velox/pull/15429

Java implementation: https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/type/khyperloglog/KHyperLogLogFunctions.java
Presto docs: https://prestodb.io/docs/current/functions/khyperloglog.html

Differential Revision: D87008234


czentgr:
@HeidiHan0000 Can we close this? This appears to be a duplicate of https://github.com/facebookincubator/velox/pull/15744.

PR #15552, opened 2025-11-18
feat: Support matchRowFieldsByName in CastHooks

kevinwilfong:
Summary:
Currently, CastExpr determines whether to map row fields by name or by position when
casting between row types, based on a setting in the QueryConfig. This means that it must
apply uniformly to all CastExprs in a query.

We have a use case in Spark where the engine supports cast mapping row fields by position
by default, but a custom UDF enables casting mapping row fields by name. To enable this I'm
adding a function to the CastHooks interface that we check to determine how to map
fields. For PrestoCastHooks this defaults to the value from QueryConfig so there's no change
in behavior. For SparkCastHooks it looks at the value from the QueryConfig as well as a
constructor parameter and OR's them, preserving the existing behavior unless we explicitly
set the new behavior when constructing the instance of SparkCastHooks.

I changed the name in the CastHooks to matchRowFieldsByName instead of 
matchStructByName because we're matching the fields not the struct, and we refer to the
type as row in Velox not struct. Note that matchStructByName is hardcoded in the config
name string, so to not break backwards compatibility I did not change it there.

Differential Revision: D87371309

### DIFF
--- a/velox/functions/sparksql/specialforms/SparkCastHooks.h
+++ b/velox/functions/sparksql/specialforms/SparkCastHooks.h
@@ -16,7 +16,7 @@
 
 #pragma once
 
-#include "velox/core/QueryCtx.h"
+#include "velox/core/QueryConfig.h"
 #include "velox/expression/CastHooks.h"
 
 namespace facebook::velox::functions::sparksql {
@@ -26,7 +26,8 @@ class SparkCastHooks : public exec::CastHooks {
  public:
   explicit SparkCastHooks(
       const velox::core::QueryConfig& config,
-      bool allowOverflow);
+      bool allowOverflow,
+      bool matchRowFieldsByName = false);
 
   // TODO: Spark hook allows more string patterns than Presto.
   Expected<Timestamp> castStringToTimestamp(
@@ -78,6 +79,10 @@ class SparkCastHooks : public exec::CastHooks {
 
   exec::PolicyType getPolicy() const override;
 
+  bool matchRowFieldsByName() const override {
+    return matchRowFieldsByName_;
+  }
+
  private:
   // Casts a number to a timestamp. The number is treated as the number of
   // seconds since the epoch (1970-01-01 00:00:00 UTC).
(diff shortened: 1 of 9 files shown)


velox/functions/sparksql/specialforms/SparkCastHooks.h
```diff
@@ -78,6 +79,10 @@ class SparkCastHooks : public exec::CastHooks {
 
   exec::PolicyType getPolicy() const override;
 
+  bool matchRowFieldsByName() const override {
+    return matchRowFieldsByName_;
+  }
```
rui-mo, line 84: I see that this PR adds a `matchRowFieldsByName_` field in the cast hook, but I’m not able to find where this function is called. Could you please clarify how and where this will be used? Thanks.

kevinwilfong: @rui-mo  It's called at line 597 in velox/expression/CastExpr.cpp

kevinwilfong: Or did you mean where would we initialize SparkCastHooks without the default value for matchRowFieldsByName passed to the constructor?

I don't expect that to get used in the Cast expressions we use in OS Velox (or Gluten for that matter). This would allow us to register additional special forms (with a name other than "cast") that also map to CastExpr but have slightly different semantics than Spark's cast. We currently expose these in Spark through custom UDFs.

velox/functions/sparksql/tests/SparkCastHooksTest.cpp
```diff
@@ -0,0 +1,48 @@
...
+  auto testMatchRowFieldsByName = [](bool queryConfigValue,
+                                     bool castHooksValue,
+                                     bool expected) {
+    core::QueryConfig queryConfig(
+        {{core::QueryConfig::kCastMatchStructByName,
+          folly::to<std::string>(queryConfigValue)}});
+
+    EXPECT_EQ(
+        expected,
+        SparkCastHooks(queryConfig, false /* allowOverflow */, castHooksValue)
```
rui-mo, line 36: nit: perhaps rename `castHooksValue` as `matchRowFieldsByName`.

rui-mo, line 36: nit: `/*allowOverflow=*/false `

velox/functions/sparksql/specialforms/SparkCastHooks.h
rui-mo: Thank you for the clarification, I understand now. It seems this change won't impact the Spark cast. The information you've provided is very helpful. Would you like to document that for the `matchRowFieldsByName_` field?

PR #15561, opened 2025-11-19
refactor: Add Filter::as<T> convenience APIs

mbasmanova:
Differential Revision: D87426354

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -50,13 +50,12 @@ T singleValue(const VectorPtr& vector) {
   return simpleVector->valueAt(0);
 }
 
-common::BigintRange* asBigintRange(std::unique_ptr<common::Filter>& filter) {
-  return dynamic_cast<common::BigintRange*>(filter.get());
+bool isBigintRange(const std::unique_ptr<common::Filter>& filter) {
+  return filter->is(common::FilterKind::kBigintRange);
 }
 
-common::BigintMultiRange* asBigintMultiRange(
-    std::unique_ptr<common::Filter>& filter) {
-  return dynamic_cast<common::BigintMultiRange*>(filter.get());
+bool isBigintMultiRange(const std::unique_ptr<common::Filter>& filter) {
+  return filter->is(common::FilterKind::kBigintMultiRange);
 }
 
 template <typename T, typename U>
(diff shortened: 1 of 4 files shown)


velox/expression/ExprToSubfieldFilter.h
```diff
@@ -281,24 +281,48 @@ inline std::unique_ptr<common::Filter> in(
...
     const std::vector<std::string>& values,
     bool nullAllowed = false) {
   return std::make_unique<common::NegatedBytesValues>(values, nullAllowed);
 }
 
+inline std::unique_ptr<common::NegatedBytesValues> notIn(
+    std::initializer_list<std::string> values,
+    bool nullAllowed = false) {
+  return notIn(std::vector<std::string>(values), nullAllowed);
+}
```
MBkkt, line 324: I think such methods are unnecessary, `vector` has ctor from `initializer_list`.
Just change `const std::vector<std::string>&` to `std::vector<std::string>` and use `std::move` if you want to avoid additional copy

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -132,41 +137,37 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeNotEqualFilter(
...
+
+  if (typeKind == TypeKind::HUGEINT) {
     VELOX_NYI();
   } else {
     std::vector<std::unique_ptr<common::Filter>> filters;
     filters.emplace_back(std::move(lessThanFilter));
     filters.emplace_back(std::move(greaterThanFilter));
+
     return std::make_unique<common::MultiRange>(std::move(filters), false);
   }
```
MBkkt, line 172: This code is completely unnecessary makeOr should do this. Why do we implement disjunction manually here.


Another important point, that IMO we should use NegatedBigintRange for integers, not `... < x || x < ...`

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -50,25 +50,30 @@ T singleValue(const VectorPtr& vector) {
...
+  return filter->is(common::FilterKind::kBigintMultiRange);
 }
 
 template <typename T, typename U>
 std::unique_ptr<T> asUniquePtr(std::unique_ptr<U> ptr) {
   return std::unique_ptr<T>(static_cast<T*>(ptr.release()));
 }
 
+template <typename T>
+std::unique_ptr<common::BigintRange> asBigintRange(std::unique_ptr<T>& ptr) {
```
MBkkt: Why is it template?
```suggestion
std::unique_ptr<common::BigintRange> asBigintRange(std::unique_ptr<common::Filter>& ptr) {
```

mbasmanova: a left over from refactoring; will fix; thanks.

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: This logic may have been added before NegatedXxx filters existed. We may look into changing it in a follow-up. @Yuhta Jimmy, would you have any concerns about using NegatedBigintRange for !=?

PR #15565, opened 2025-11-19
feat: Enhance extractFiltersFromRemainingFilter to support an OR

mbasmanova:
Summary:
Add support for converting "f(a) OR g(a)" to subfield filter.

For example,

- a > 0 OR a < 10 becomes BigintRange(1, 9)

Differential Revision: D87378318

### DIFF
--- a/velox/connectors/hive/HiveConnectorUtil.cpp
+++ b/velox/connectors/hive/HiveConnectorUtil.cpp
@@ -930,6 +930,49 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
     }
     return replaceInputs(call, std::move(args));
   }
+
+  if ((call->name() == expression::kAnd && negated) ||
+      (call->name() == expression::kOr && !negated)) {
+    std::vector<std::unique_ptr<common::Filter>> disjuncts;
+    common::Subfield subfield;
+
+    for (const auto& input : call->inputs()) {
+      common::SubfieldFilters tmpFilters;
+      double tmpSampleRate = 1;
+      auto tmpRemaining = extractFiltersFromRemainingFilter(
+          input, evaluator, negated, tmpFilters, tmpSampleRate);
+
+      if (tmpRemaining != nullptr || tmpSampleRate != 1 ||
+          tmpFilters.size() != 1) {
+        disjuncts.clear();
+        break;
+      }
+
+      if (disjuncts.empty()) {
+        subfield = tmpFilters.begin()->first.clone();
+      } else if (!(subfield == tmpFilters.begin()->first)) {
+        disjuncts.clear();
+        break;
+      }
+
+      disjuncts.push_back(tmpFilters.begin()->second->clone());
+    }
+
+    if (!disjuncts.empty()) {
+      auto filter =
+          exec::ExprToSubfieldFilterParser::makeOrFilter(std::move(disjuncts));
+
+      auto it = filters.find(subfield);
+      if (it != filters.end()) {
+        filter = filter->mergeWith(it->second.get());
+      }
+
+      filters.insert_or_assign(std::move(subfield), std::move(filter));
+
+      return nullptr;
+    }
+  }
+
   if (!negated) {
     double rate = getPrestoSampleRate(expr, call, evaluator);
     if (rate != -1) {
@@ -937,6 +980,7 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
       return nullptr;
     }
   }
+
   return expr;
 }
 } // namespace
--- a/velox/connectors/hive/tests/HiveConnectorTest.cpp
+++ b/velox/connectors/hive/tests/HiveConnectorTest.cpp
@@ -694,5 +694,54 @@ TEST_F(HiveConnectorTest, prestoTableSampling) {
   ASSERT_TRUE(filters.empty());
 }
 
+#define VELOX_ASSERT_FILTER(expected, actual)   \
+  ASSERT_TRUE(expected->testingEquals(*actual)) \
+      << expected->toString() << " vs " << actual->toString();
+
+TEST_F(HiveConnectorTest, disjuncts) {
+  auto queryCtx = core::QueryCtx::create();
+  exec::SimpleExpressionEvaluator evaluator(queryCtx.get(), pool_.get());
+  auto rowType = ROW({"c0", "c1", "c2"}, {BIGINT(), BIGINT(), DECIMAL(20, 0)});
+
+  {
+    auto expr =
+        parseExpr("(c0 > 0 and c0 < 10) or (c0 > 5 and c0 < 15)", rowType);
+
+    SubfieldFilters filters;
+    double sampleRate = 1;
+    auto remaining = extractFiltersFromRemainingFilter(
+        expr, &evaluator, filters, sampleRate);
+    ASSERT_TRUE(remaining == nullptr);
+    ASSERT_EQ(sampleRate, 1);
+    ASSERT_EQ(filters.size(), 1);
+    VELOX_ASSERT_FILTER(exec::between(1, 14), filters.begin()->second);
+  }
+
+  {
+    auto expr = parseExpr("(c0 between -10 and 12)", rowType);
+
+    SubfieldFilters filters;
+    double sampleRate = 1;
+    auto remaining = extractFiltersFromRemainingFilter(
+        expr, &evaluator, filters, sampleRate);
+    ASSERT_TRUE(remaining == nullptr);
+    ASSERT_EQ(sampleRate, 1);
+    ASSERT_EQ(filters.size(), 1);
+    ASSERT_EQ(filters.begin()->first, Subfield("c0"));
+    VELOX_ASSERT_FILTER(exec::between(-10, 12), filters.begin()->second);
+
+    expr = parseExpr("(c0 > 0 and c0 < 10) or (c0 > 5 and c0 < 15)", rowType);
+    remaining = extractFiltersFromRemainingFilter(
+        expr, &evaluator, filters, sampleRate);
+    ASSERT_TRUE(remaining == nullptr);
+    ASSERT_EQ(sampleRate, 1);
+    ASSERT_EQ(filters.size(), 1);
+    ASSERT_EQ(filters.begin()->first, Subfield("c0"));
+    VELOX_ASSERT_FILTER(exec::between(1, 12), filters.begin()->second);
+  }
+}
+
+#undef VELOX_ASSERT_FILTER
+
 } // namespace
 } // namespace facebook::velox::connector::hive


velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
-        std::move(newRanges), false);
+  if (auto merged = tryMergeFloatingPointRanges<double>(disjuncts)) {
+    return merged;
   }
 
-  if (asBigintMultiRange(a) && asBigintRange(b)) {
-    return makeOrFilter(std::move(b), std::move(a));
+  if (auto merged = tryMergeFloatingPointRanges<float>(disjuncts)) {
+    return merged;
   }
```
MBkkt, line 679: why double before float?

velox/expression/ExprToSubfieldFilter.h
```diff
@@ -281,24 +282,48 @@ inline std::unique_ptr<common::Filter> in(
...
     const std::vector<std::string>& values,
     bool nullAllowed = false) {
   return std::make_unique<common::NegatedBytesValues>(values, nullAllowed);
 }
 
+inline std::unique_ptr<common::NegatedBytesValues> notIn(
+    std::initializer_list<std::string> values,
+    bool nullAllowed = false) {
+  return notIn(std::vector<std::string>(values), nullAllowed);
+}
```
MBkkt, line 325: Such functions is just slowdown compilation and increase binary size without any profit.
Why do you add them, what are you trying to achieve?

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: It doesn't really matter. No particular reason.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+      for (const auto& range :
+           filter->as<common::BigintMultiRange>()->ranges()) {
+        ranges.emplace_back(std::make_unique<common::BigintRange>(*range));
+      }
+    }
+  }
+
+  std::sort(ranges.begin(), ranges.end(), [](const auto& a, const auto& b) {
+    return a->lower() < b->lower();
+  });
```
MBkkt: I think such comparison is better
```
if (a->lower() != b->lower() {
   return a->lower() < b->lower();
}
return a->upper() > b->upper();
```
because in such case [1, 1000] will be less than [1, 10], [1, 100], etc.
So we will create correct range and will create less ranges at all

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+    for (auto& filter : disjuncts) {
+      values.emplace_back(filter->as<common::BigintRange>()->lower());
+    }
+
+    return common::createBigintValues(values, isNullAllowed(disjuncts));
+  }
+
+  if (!std::all_of(disjuncts.begin(), disjuncts.end(), [](const auto& filter) {
+        return isBigintRange(filter) || isBigintMultiRange(filter);
+      })) {
```
MBkkt: This code wants to account existence of `NegatedBigintRange`.
In most cases NegatedBigintRanges || some-other-bigint will leads to AlwaysTrue.
We want this to be optimized, right?

MBkkt: This code maybe better to account BigintValues and NegatedBigintValues, because code above can produce it.

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: Are you suggesting there is a bug in this logic? Any chance you could clarify what case doesn't work?

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: I'm not sure I follow. Any chance you could clarify a specific use case that doesn't work properly?

velox/expression/ExprToSubfieldFilter.cpp
MBkkt: I didn't say there's bug. It's just ineffective.
Imagine input like this `[1, 1], [1, 2], [1, 3], [1, 4] ... [1, 1e6 + 1]`
With current code in such case instead of single allocation, it will be 1e6 allocations

velox/expression/ExprToSubfieldFilter.cpp
MBkkt: `x != 1` -- to work better should be NegatedBigintRange instead of BigintMultiRange (a < 1 || 1 < a), do we agree on this?

If so we want expressions, like `x != 1 || x == 10` optimized (needs to merge NegatedBigintRange and BigintRange)

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: Just to confirm my understanding, the optimization would have 2 parts:

1- a change to comparator you proposed
2- a change to tryMergeOverlappingRanges to check if one range is a subset of another and return the other range without making a new one

Is that right?

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+      upperUnbounded,
+      upperExclusive,
+      /*nullAllowed=*/false);
+}
+
+template <typename T>
+std::unique_ptr<common::Filter> tryMergeFloatingPointRanges(
+    std::vector<std::unique_ptr<common::Filter>>& disjuncts) {
+  if (!std::all_of(disjuncts.begin(), disjuncts.end(), [](const auto& filter) {
+        return filter->is(common::FilterKind::kDoubleRange);
```
MBkkt: @mbasmanova it looks like a bug, it should be FloatRange for `T is float`?

@mbasmanova I also think mix of float ranges and double ranges can be merged into double range

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: I see. That would be a useful optimization, but I'd like to keep this PR relatevly small. We can implement this in a follow-up.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+
+    if (a->lowerUnbounded()) {
+      return true;
+    }
+
+    if (b->lowerUnbounded()) {
+      return false;
+    }
+
+    return a->lower() < b->lower();
```
MBkkt, line 647: Same what I said for bigint, it's better to sort in opposite order by upper, if lower is equal to each other.

Also, nit:

```
    if (a->lowerUnbounded() != b->lowerUnbounded()) {
      return a->lowerUbounded() > b->lowerUbounded();
    }
    return a->lower() < b->lower();
```
I think such code a lot of easier to read and follow, when something like tuple comparison written
1. check is equal or not
2. if not produce result
3. else try next member

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+    return std::move(newRanges.front());
+  }
+
+  std::vector<std::unique_ptr<common::Filter>> filters;
+  filters.reserve(newRanges.size());
+  for (auto& range : newRanges) {
+    filters.emplace_back(std::move(range));
+  }
+
+  return std::make_unique<common::MultiRange>(std::move(filters), nullAllowed);
```
MBkkt: This code mostly duplicated between this function and bigint function.
I think it can be single function, this will decrease chance of bug, because increase code coverage by a lot

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: > it looks like a bug, it should be FloatRange for T is float?

Nice catch. Thanks. Fixing.

> float ranges and double ranges can be merged into double range

No, that's not possible. The type of the range must match the type of the column.

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,253 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
-
-  if (asBigintRange(a) && asBigintMultiRange(b)) {
-    const auto& ranges = asBigintMultiRange(b)->ranges();
-    std::vector<std::unique_ptr<common::BigintRange>> newRanges;
-    newRanges.emplace_back(asUniquePtr<common::BigintRange>(std::move(a)));
-    for (const auto& range : ranges) {
-      newRanges.emplace_back(asUniquePtr<common::BigintRange>(range->clone()));
-    }
+    std::vector<std::unique_ptr<common::Filter>> disjuncts) {
+  VELOX_CHECK_GE(disjuncts.size(), 2);
```
MBkkt, line 667: We should check for cases when
1. if one of them is always true we want to return it
2. then if we have both isNull and isNotNull we need to return always true
3. then we want to remove everything that is always false
4. then we compute nullAllowed
5. then we can remove all isNull and all isNotNull

velox/expression/ExprToSubfieldFilter.cpp
MBkkt: Yes

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,257 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+    if (a.lower() < b.lower()) {
+      lowerExclusive = a.lowerExclusive();
+    } else if (a.lower() > b.lower()) {
+      lowerExclusive = b.lowerExclusive();
+    } else {
+      lowerExclusive = a.lowerExclusive() && b.lowerExclusive();
+    }
+  }
+
+  const T upper = upperUnbounded ? std::numeric_limits<double>::max()
```
MBkkt: T instead of double?

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -438,38 +439,257 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeBetweenFilter(
...
+  }
+
+  if (!b.upperUnbounded() && !a.lowerUnbounded() && b.upper() < a.lower()) {
+    return nullptr;
+  }
+
+  const bool lowerUnbounded = a.lowerUnbounded() || b.lowerUnbounded();
+  const bool upperUnbounded = a.upperUnbounded() || b.upperUnbounded();
+
+  const T lower = lowerUnbounded ? std::numeric_limits<double>::lowest()
```
MBkkt: T instead of double?

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: These might be useful additions, but I feel they are out of scope of this PR. Can go in a follow-up though.

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: I extracted common logic into mergeOverlappingDisjuncts helper function. Thanks.

velox/expression/ExprToSubfieldFilter.cpp
mbasmanova: I agree... except in this case I believe it is confusing 

> return a->lowerUbounded() > b->lowerUbounded();

One would wonder why we use greater than and not less than (like in a->lower() < b->lower()). I believe the answer is because true > false and true here means "lower".... super confusing.

PR #15569, opened 2025-11-19
fix: Do not rewrite expression if it only has constant inputs

pramodsatya:
Expression rewrite is invoked after constant folding, hence, rewrite should not expect a deterministic input expression with all constant inputs. Adds a check to ExprRewriteRegistry.

### DIFF
--- a/velox/expression/ExprRewriteRegistry.cpp
+++ b/velox/expression/ExprRewriteRegistry.cpp
@@ -28,6 +28,16 @@ void ExprRewriteRegistry::clear() {
 
 core::TypedExprPtr ExprRewriteRegistry::rewrite(
     const core::TypedExprPtr& expr) {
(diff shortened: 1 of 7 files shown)


velox/expression/ExprRewriteRegistry.cpp
```diff
@@ -28,6 +28,16 @@ void ExprRewriteRegistry::clear() {
 
 core::TypedExprPtr ExprRewriteRegistry::rewrite(
     const core::TypedExprPtr& expr) {
+  const auto& inputs = expr->inputs();
+  VELOX_CHECK(
```
mbasmanova, line 32: We may need to relax this check later if we get any re-writes for non-deterministic expressions.

czentgr (approved):

PR #15572, opened 2025-11-19
fix(ci): Clang-tidy does not find compile-commands.json

czentgr:
Clang-tidy is skipped because it cannot find
the compile-commands.json that are generated in the previous step.


czentgr:
@assignUser We saw that the clang-tidy step was skipped because it couldn't find the compile-commands.json. So I added the path.

But turns out there is another problem. When I tried it in one of my PRs one of the test files required a generated header. And that header is a protocol header and not built by `make cmake`. Only trying to build the actual target would it run the commands to generate it. I have no good idea on how to generate the files without a full rebuild.
See https://github.com/facebookincubator/velox/actions/runs/19514077568/job/55861131876?pr=15466

If we have command to build targets that would generate the files we might still build half of Velox.

kgpai:
Thanks @czentgr

kgpai:
>  I have no good idea on how to generate the files without a full rebuild.
See https://github.com/facebookincubator/velox/actions/runs/19514077568/job/55861131876?pr=15466

> If we have command to build targets that would generate the files we might still build half of Velox.

Clang tidy doesnt have to run as part of preliminary check, we can move it to the adapters build after everything is built ?

assignUser:
> Clang tidy doesnt have to run as part of preliminary check, we can move it to the adapters build after everything is built ?

That would also be a possibility

czentgr:
@assignUser I moved the clang-tidy to the adapters build and run it as the last step (to avoid having issues when we skip it for clang build for now because we don't build everything with clang).
If we assume there are no existing clang-tidy issues in the code base this should do it.

PR #15577, opened 2025-11-20
fix: Fix flaky test TableScanTest::statsBasedFilterReorderDisabled

juwentus1234:
Summary:
The test `statsBasedFilterReorderDisabled` was flaky because it expected only 1 storage read on the second query run, but sometimes got 1877+ reads instead due to cache miss.

Root cause: tableScanStats.customStats["storageReadBytes"].count returns numbers of reads which is not tableScanStats.customStats.count("storageReadBytes") which returns either 0 or 1
cache miss case should be the same as first read.

Differential Revision: D87501924

### DIFF
--- a/velox/exec/tests/TableScanTest.cpp
+++ b/velox/exec/tests/TableScanTest.cpp
@@ -5761,7 +5761,7 @@ TEST_F(TableScanTest, footerIOCount) {
   ASSERT_GT(stats.at("footerBufferOverread").sum, 0);
 }
 
-TEST_F(TableScanTest, statsBasedFilterReorderDisabled) {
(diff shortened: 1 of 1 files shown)


velox/exec/tests/TableScanTest.cpp
```diff
@@ -5790,8 +5790,8 @@ TEST_F(TableScanTest, statsBasedFilterReorderDisabled) {
   }
   createDuckDbTable(vectors);
 
-  for (auto disableReoder : {false}) {
-    SCOPED_TRACE(fmt::format("disableReoder {}", disableReoder));
+  for (auto disableReorder : {false}) {
```
xiaoxmeng: shall we test both enable and disable?

PR #15581, opened 2025-11-20
refactor: Add Iceberg connector

PingLiuPing:
Refactors the Iceberg connector implementation to separate it from Hive connector. Sets the foundation for independent evolution of both connectors.

### DIFF
--- a/velox/connectors/hive/HiveConnector.cpp
+++ b/velox/connectors/hive/HiveConnector.cpp
@@ -20,7 +20,6 @@
 #include "velox/connectors/hive/HiveDataSink.h"
 #include "velox/connectors/hive/HiveDataSource.h"
 #include "velox/connectors/hive/HivePartitionFunction.h"
-#include "velox/connectors/hive/iceberg/IcebergDataSink.h"
 
 #include <boost/lexical_cast.hpp>
 #include <memory>
@@ -74,16 +73,6 @@ std::unique_ptr<DataSink> HiveConnector::createDataSink(
     ConnectorInsertTableHandlePtr connectorInsertTableHandle,
     ConnectorQueryCtx* connectorQueryCtx,
     CommitStrategy commitStrategy) {
-  if (auto icebergInsertHandle =
-          std::dynamic_pointer_cast<const iceberg::IcebergInsertTableHandle>(
-              connectorInsertTableHandle)) {
-    return std::make_unique<iceberg::IcebergDataSink>(
-        inputType,
-        icebergInsertHandle,
-        connectorQueryCtx,
-        commitStrategy,
-        hiveConfig_);
-  }
   auto hiveInsertHandle =
       std::dynamic_pointer_cast<const HiveInsertTableHandle>(
           connectorInsertTableHandle);
--- a/velox/connectors/hive/iceberg/IcebergConnector.cpp
+++ b/velox/connectors/hive/iceberg/IcebergConnector.cpp
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/connectors/hive/iceberg/IcebergConnector.h"
+
+#include "velox/connectors/hive/HiveConnector.h"
+#include "velox/connectors/hive/iceberg/IcebergDataSink.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+const std::string_view kIcebergFunctionPrefixConfig{"presto.iceberg-namespace"};
+const std::string_view kDefaultIcebergFunctionPrefix{"$internal$.iceberg."};
+
+namespace {
+
+// Registers Iceberg partition transform functions with prefix.
+// NOTE: These functions are registered for internal transform usage only.
+// Upstream engines such as Prestissimo and Gluten should register the same
+// functions with different prefixes to avoid conflicts.
+void registerIcebergInternalFunctions(const std::string& prefix) {
+  static std::once_flag registerFlag;
+
+  std::call_once(registerFlag, [prefix]() {
+    functions::iceberg::registerFunctions(prefix);
+  });
+}
+
+} // namespace
+
+IcebergConnector::IcebergConnector(
+    const std::string& id,
+    std::shared_ptr<const config::ConfigBase> config,
+    folly::Executor* ioExecutor)
+    : HiveConnector(id, config, ioExecutor),
+      functionPrefix_(config->get<std::string>(
+          std::string(kIcebergFunctionPrefixConfig),
+          std::string(kDefaultIcebergFunctionPrefix))) {
+  registerIcebergInternalFunctions(functionPrefix_);
+}
+
+std::unique_ptr<DataSink> IcebergConnector::createDataSink(
+    RowTypePtr inputType,
+    ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+    ConnectorQueryCtx* connectorQueryCtx,
+    CommitStrategy commitStrategy) {
+  auto icebergInsertHandle =
+      checked_pointer_cast<const IcebergInsertTableHandle>(
+          connectorInsertTableHandle);
+
+  return std::make_unique<IcebergDataSink>(
+      inputType,
+      icebergInsertHandle,
+      connectorQueryCtx,
+      commitStrategy,
+      hiveConfig_,
+      functionPrefix_);
+}
+
+} // namespace facebook::velox::connector::hive::iceberg
--- a/velox/connectors/hive/iceberg/IcebergConnector.h
+++ b/velox/connectors/hive/iceberg/IcebergConnector.h
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/connectors/hive/HiveConnector.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+/// TODO Add IcebergConfig class and Move these configuration properties to
+/// IcebergConfig.h
+extern const std::string_view kIcebergFunctionPrefixConfig;
+extern const std::string_view kDefaultIcebergFunctionPrefix;
+
+/// Provides Iceberg table format support.
+/// - Creates HiveDataSource instances that use IcebergSplitReader for reading
+///   Iceberg tables with support for delete files and schema evolution.
+/// - Creates IcebergDataSink instances for writing data with Iceberg-specific
+///   partition transforms and commit metadata.
+class IcebergConnector final : public HiveConnector {
+ public:
+  IcebergConnector(
+      const std::string& id,
+      std::shared_ptr<const config::ConfigBase> config,
+      folly::Executor* ioExecutor);
+
+  /// Creates IcebergDataSink for writing to Iceberg tables.
+  ///
+  /// @param inputType The schema of the input data to write.
+  /// @param connectorInsertTableHandle Must be an IcebergInsertTableHandle
+  ///        containing Iceberg-specific write configuration.
+  /// @param connectorQueryCtx Query context for the write operation.
+  /// @param commitStrategy Strategy for committing the write operation. Only
+  ///        CommitStrategy::kNoCommit is supported for Iceberg tables. Files
+  ///        are written directly with their final names and commit metadata is
+  ///        returned for the coordinator to update the Iceberg metadata tables.
+  /// @return IcebergDataSink instance configured for the write operation.
+  std::unique_ptr<DataSink> createDataSink(
+      RowTypePtr inputType,
+      ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+      ConnectorQueryCtx* connectorQueryCtx,
+      CommitStrategy commitStrategy) override;
+
+ private:
+  const std::string functionPrefix_;
+};
+
+class IcebergConnectorFactory final : public ConnectorFactory {
+ public:
+  static constexpr const char* kIcebergConnectorName = "iceberg";
+
+  IcebergConnectorFactory() : ConnectorFactory(kIcebergConnectorName) {}
+
+  /// Creates a new IcebergConnector instance.
+  ///
+  /// @param id Unique identifier for this connector instance (typically the
+  /// catalog name).
+  /// @param config Connector configuration properties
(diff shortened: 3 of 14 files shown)


PingLiuPing:
@mbasmanova  This is a follow up PR from discussion https://github.com/facebookincubator/velox/pull/15440#discussion_r2510085483. Code changes in Prestissimo:  https://github.com/prestodb/presto/pull/26661

velox/connectors/hive/iceberg/tests/IcebergTestBase.cpp
```diff
@@ -36,6 +36,15 @@ void IcebergTestBase::SetUp() {
   functions::iceberg::registerFunctions(
       std::string(kDefaultTestIcebergFunctionNamePrefix));
 
+  // Register IcebergConnector.
+  IcebergConnectorFactory icebergFactory;
+  auto icebergConnector = icebergFactory.newConnector(
```
mbasmanova, line 39: Can we add a way to specify the prefix to use for transform functions? This way, the application can control the contents of the function registry better.

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
```diff
@@ -72,5 +88,26 @@ TEST_F(IcebergInsertTest, bigDecimal) {
 }
 #endif
 
+TEST_F(IcebergInsertTest, connectorConfiguration) {
```
mbasmanova: Why does this test belongs to 'insert' test? How do we verify that configuration properties do take effect and not just ignored?

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
```diff
@@ -72,5 +88,26 @@ TEST_F(IcebergInsertTest, bigDecimal) {
...
+          {"file-handle-cache-enabled", "true"},
+          {"cache-max-file-handles", "1000"}});
+
+  resetIcebergConnector(customConfig);
+
+  auto rowType = ROW({"c1", "c2"}, {INTEGER(), VARCHAR()});
+  test(rowType);
+}
+
+TEST_F(IcebergInsertTest, connectorProperties) {
```
mbasmanova: Does this test belong here?

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
```diff
@@ -36,9 +37,24 @@ class IcebergInsertTest : public test::IcebergTestBase {
...
+    auto plan = exec::test::PlanBuilder()
+                    .startTableScan()
+                    .connectorId(test::kIcebergConnectorId)
+                    .outputType(rowType)
+                    .endTableScan()
+                    .planNode();
     exec::test::AssertQueryBuilder(plan).splits(splits).assertResults(vectors);
   }
+
+  static void resetIcebergConnector(
```
mbasmanova: Should this go into xxxTestBase?

velox/connectors/hive/iceberg/IcebergDataSink.h
```diff
@@ -24,11 +24,8 @@
 
 namespace facebook::velox::connector::hive::iceberg {
 
-/// Registers Iceberg partition transform functions with prefix.
-/// NOTE: These functions are registered for internal transform usage only.
-/// Upstream engines such as Prestissimo and Gluten should register the same
-/// functions with different prefixes to avoid conflicts.
-void registerIcebergInternalFunctions(const std::string_view& prefix);
+static constexpr std::string_view kDefaultIcebergFunctionPrefix{
```
mbasmanova: I think we want to remove this constant altogether and let the application specify the prefix when registering the connector.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+/// Factory for creating IcebergConnector instances.
+class IcebergConnectorFactory final : public ConnectorFactory {
+ public:
+  static constexpr const char* kIcebergConnectorName = "iceberg";
+
+  IcebergConnectorFactory() : ConnectorFactory(kIcebergConnectorName) {}
+
+  std::shared_ptr<Connector> newConnector(
```
mbasmanova, line 79: Would you document this API? Specifically, what are ioExecutor and cpuExecutor are for and why they are optional?

velox/connectors/hive/iceberg/IcebergConnector.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+std::unique_ptr<DataSink> IcebergConnector::createDataSink(
+    RowTypePtr inputType,
+    ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+    ConnectorQueryCtx* connectorQueryCtx,
+    CommitStrategy commitStrategy) {
+  auto icebergInsertHandle =
+      std::dynamic_pointer_cast<const IcebergInsertTableHandle>(
+          connectorInsertTableHandle);
+  VELOX_USER_CHECK_NOT_NULL(
```
mbasmanova: Use checked_poiner_cast from velox/common/Casts.h

velox/connectors/hive/HiveConnector.cpp
```diff
@@ -74,16 +73,6 @@ std::unique_ptr<DataSink> HiveConnector::createDataSink(
     ConnectorInsertTableHandlePtr connectorInsertTableHandle,
     ConnectorQueryCtx* connectorQueryCtx,
     CommitStrategy commitStrategy) {
-  if (auto icebergInsertHandle =
```
mbasmanova, line 77: Nice.

velox/connectors/hive/iceberg/tests/IcebergTestBase.cpp
PingLiuPing: Thanks. This makes sense.
But `newConnector` is an interface from `ConnectorFactory`, we cannot adding a new parameter to this method to  specifying the prefix since the parameter doesn't mean anything to other connector. And I see Prestissimo calls this method for all types of connector with same parametes, so we cannot adding a new method either otherwise it requires special handling from caller.
Instead I hard-code the prefix in `IcebergConnectorFactory::newConnector` and pass it to IcebergConnector, then pass it to IcebergDataSink. This prefix is for internal use only.

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
PingLiuPing: Thanks. 
Moved them to a new test file.

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
PingLiuPing: Thanks, moved.

velox/connectors/hive/iceberg/tests/IcebergInsertTest.cpp
PingLiuPing: Moved to new test file.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: Thanks. Sure. `cpuExecutor` is not used by Hive and Iceberg connector.

velox/connectors/hive/iceberg/IcebergConnector.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergTestBase.cpp
mbasmanova: I think it is desirable to avoid hard-coding this prefix. Can we define connector-specific configuration property and use that instead?

PingLiuPing: Yes, I looked at this too. But seems this way requires a new entry in the catalog property file.

See https://github.com/prestodb/presto/blob/9374cd8672d8e76a2c9db0ee296a5c8a320e5b58/presto-native-execution/presto_cpp/main/PrestoServer.cpp#L1270-L1300

mbasmanova: Is this a problem?

PingLiuPing: It seems to me that this prefix is primarily for internal use, and therefore the application may not need to specify it explicitly in the catalog property file. And we still need some logic to amend this if the property is missing.

mbasmanova: The problem is that currently a single application (Prestissimo) imposes restrictions on the Connector which may be used in other applications. If these other applications impose their own restrictions that conflict with restrictions imposed by Prestissimo there will be no way to satisfy these. Hence, it would be best to avoid such restrictions. Makes sense?

PingLiuPing: Thank you. Yes, understood now.
Update the code to retrieve this prefix from config and setting the default prefix if not present.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,88 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/connectors/hive/HiveConnector.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+static constexpr std::string_view kIcebergFunctionPrefix{
```
mbasmanova: We probably want something similar to velox/connectors/hive/HiveConfig.h

PingLiuPing: Yes, agree. I actually added IcebergConfig.h and then removed it since I only need one property in this PR.
Can I open another PR to add IcebergConfig.h? It requires some changes on other Iceberg classes too.

mbasmanova: Sounds good. Let's then add a TODO here.

mbasmanova: The names kIcebergFunctionPrefix and kDefaultIcebergFunctionPrefix suggest that these are both prefixes... however one is the name of the config property whose value is the prefix and the other is the prefix. Perhaps, rename kIcebergFunctionPrefix to kIcebergFunctionPrefixConfig to avoid confusion.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,88 @@
...
+  /// @param connectorQueryCtx Query context for the write operation.
+  /// @param commitStrategy Strategy for committing the write operation.
+  /// @return IcebergDataSink instance configured for the write operation.
+  /// @throws VeloxUserError if connectorInsertTableHandle is not an
+  ///         IcebergInsertTableHandle.
+  std::unique_ptr<DataSink> createDataSink(
+      RowTypePtr inputType,
+      ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+      ConnectorQueryCtx* connectorQueryCtx,
+      CommitStrategy commitStrategy) override;
```
mbasmanova, line 54: Does Iceberg have support for all commit strategies? If not, let's clarify what is supported.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,88 @@
...
+
+  /// Creates IcebergDataSink for writing to Iceberg tables.
+  ///
+  /// @param inputType The schema of the input data to write.
+  /// @param connectorInsertTableHandle Must be an IcebergInsertTableHandle
+  ///        containing Iceberg-specific write configuration.
+  /// @param connectorQueryCtx Query context for the write operation.
+  /// @param commitStrategy Strategy for committing the write operation.
+  /// @return IcebergDataSink instance configured for the write operation.
+  /// @throws VeloxUserError if connectorInsertTableHandle is not an
```
mbasmanova: This comment is redundant. Let's remove.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,88 @@
...
+      RowTypePtr inputType,
+      ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+      ConnectorQueryCtx* connectorQueryCtx,
+      CommitStrategy commitStrategy) override;
+
+ private:
+  const std::string functionPrefix_;
+};
+
+/// Factory for creating IcebergConnector instances.
```
mbasmanova: Redundant comment.

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,88 @@
...
+  /// operations run synchronously.
+  /// @param cpuExecutor ConnectorFactory interface to support other connector
+  /// types that may need CPU-bound async work. Currently unused by
+  /// IcebergConnector.
+  /// @return Shared pointer to the newly created IcebergConnector instance
+  std::shared_ptr<Connector> newConnector(
+      const std::string& id,
+      std::shared_ptr<const config::ConfigBase> config,
+      folly::Executor* ioExecutor = nullptr,
+      folly::Executor* cpuExecutor = nullptr) override {
```
mbasmanova: linter will complain about this parameter being unused. Add [[maybe_unused]].

velox/connectors/hive/iceberg/tests/IcebergTestBase.cpp
```diff
@@ -57,6 +62,8 @@ void IcebergTestBase::TearDown() {
   opPool_.reset();
   root_.reset();
   queryCtx_.reset();
+  // Unregister IcebergConnector
```
mbasmanova: redundant comment

PingLiuPing: thanks.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: Thanks. Added this to HiveConnector as well.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: thanks.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: Yes.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: Thanks, only kNoCommit is supported.

velox/connectors/hive/iceberg/IcebergConnector.h
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/IcebergTestBase.h
```diff
@@ -29,8 +29,7 @@
 
 namespace facebook::velox::connector::hive::iceberg::test {
 
-constexpr std::string_view kDefaultTestIcebergFunctionNamePrefix{
-    "$internal$.test_iceberg."};
+static const std::string kIcebergConnectorId = "test-iceberg";
```
kagamiori: There is a linter complaint about ODR violation that variables shouldn't be defined in hearder file. Could we leave a declaration here and move the definition to .cpp file?

velox/connectors/hive/iceberg/IcebergConnector.h
```diff
@@ -0,0 +1,90 @@
...
+#include "velox/connectors/hive/HiveConnector.h"
+
+namespace facebook::velox::connector::hive::iceberg {
+
+/// TODO Add IcebergConfig class and Move these configuration properties to
+/// IcebergConfig.h
+static constexpr std::string_view kIcebergFunctionPrefixConfig{
+    "presto.iceberg-namespace"};
+static constexpr std::string_view kDefaultIcebergFunctionPrefix{
+    "$internal$.iceberg."};
```
kagamiori: Same linter complaint here. Could we move the definition to .cpp file?

kagamiori (commented):
Hi @PingLiuPing, got two linter complaints as below. Could you take a look? Thanks!

velox/connectors/hive/iceberg/tests/IcebergTestBase.h
PingLiuPing: Thanks, will fix this.

PR #15583, opened 2025-11-20
perf: Optimize copying distinct spilling vector to output vector in HashAggragtion

NEUpanning:
Currently, when it comes to copying a distinct spilling vector to the output
vector, the `GroupingSet::mergeNextWithoutAggregates` method leverages
`RowVector::copy`. However, this approach is not efficient, as it involves
operations which are not necessary for this scenario such as allocating the
source to the target index buffer, ensuring that the output vector is writable,
and copying ascii-ness info. This PR utilizes `OperatorUtils::gatherCopy`
instead.

### DIFF
--- a/velox/exec/GroupingSet.cpp
+++ b/velox/exec/GroupingSet.cpp
@@ -1286,7 +1286,7 @@ bool GroupingSet::mergeNextWithoutAggregates(
   bool newDistinct{true};
   int32_t numOutputRows{0};
   int32_t outputSize{0};
-  bool isEndOfBatch = false;
+  bool endOfBatch = false;
   prepareSpillResultWithoutAggregates(maxOutputRows, result);
 
   while (numOutputRows + outputSize < maxOutputRows) {
@@ -1308,7 +1308,7 @@ bool GroupingSet::mergeNextWithoutAggregates(
         numDistinctSpillFilesPerPartition_[outputSpillPartition_]) {
       newDistinct = false;
     }
-    auto index = stream->currentIndex(&isEndOfBatch);
+    auto index = stream->currentIndex(&endOfBatch);
     if (!next.second && newDistinct) {
       // Yield result for new distinct.
       spillSources_[outputSize] = &stream->current();
@@ -1316,7 +1316,7 @@ bool GroupingSet::mergeNextWithoutAggregates(
       ++outputSize;
     }
 
-    if (FOLLY_UNLIKELY(isEndOfBatch)) {
+    if (FOLLY_UNLIKELY(endOfBatch)) {
       // The stream is at end of input batch. Need to copy out the rows before
       // fetching next batch in 'pop'.
       gatherCopy(


velox/vector/FlatVector-inl.h
```diff
@@ -485,8 +485,8 @@ void FlatVector<T>::resize(vector_size_t newSize, bool setNotNull) {
 
 template <typename T>
 void FlatVector<T>::ensureWritable(const SelectivityVector& rows) {
-  auto newSize = std::max<vector_size_t>(rows.end(), BaseVector::length_);
   if (values_ && !values_->isMutable()) {
+    auto newSize = std::max<vector_size_t>(rows.end(), BaseVector::length_);
```
mbasmanova: Is this unrelated refactoring? Would you submit this as a separate PR?

velox/exec/GroupingSet.cpp
```diff
@@ -1306,13 +1314,39 @@ bool GroupingSet::mergeNextWithoutAggregates(
...
+    gatherCopy(
+        spillResultWithoutAggregates_.get(),
+        numOutputRows,
+        outputSize,
+        spillSources_,
+        spillSourceRows_);
+    numOutputRows += outputSize;
+  }
   spillResultWithoutAggregates_->resize(numOutputRows);
+  for (auto i = 0; i < spillResultWithoutAggregates_->childrenSize(); ++i) {
```
mbasmanova: Let's add RowVector::resizeRecursively or resizeWithChildren API to avoid hard-coding this logic in multiple places.

velox/vector/FlatVector-inl.h
NEUpanning: > Is this unrelated refactoring? 

No, I also noticed this issue when I perf hashagg spilling. I will submit another PR.

velox/vector/ComplexVector.h
```diff
@@ -244,11 +244,18 @@ class RowVector : public BaseVector {
...
+  /// increasing the size. Note : If the child is null, then it will stay null
+  /// after the resize.
+  void resize(vector_size_t newSize, bool setNotNull = true) override;
+
   /// Resizes the parent row container and also recursively resizes the
   /// children. Note that this function will throw if the children are not
   /// uniquely referenced by the parent when increasing the size.
   /// Note : If the child is null, then it will stay null after the resize.
-  void resize(vector_size_t newSize, bool setNotNull = true) override;
+  void resizeWithChildren(vector_size_t newSize, bool setNotNull = true);
```
mbasmanova: Let's move this into a separate PR.

velox/exec/GroupingSet.cpp
NEUpanning: @mbasmanova Thanks. I've added resizeWithChildren API. SortBuffer can also utilize this API, but I don't know why it needs to resize children vector since after calling`BaseVector::prepareForReuse` the size of children vector should be `batchSize`. Do you have any input?

https://github.com/facebookincubator/velox/blob/5a746f73f15d2b489d436d3c7c845e662fc250b8/velox/exec/SortBuffer.cpp#L383-L393

velox/vector/ComplexVector.h
```diff
@@ -244,11 +244,18 @@ class RowVector : public BaseVector {
   /// until the few places that require it are migrated over.
   void unsafeResize(vector_size_t newSize, bool setNotNull = true);
 
+  /// Resizes the parent row container and also recursively resizes the
```
mbasmanova: > Resizes the parent row container and also recursively resizes the children

is this already how this API works? If so, we don't need a new API, do we?

velox/exec/GroupingSet.cpp
mbasmanova: Might be an oversight. CC: @kewang1024 @xiaoxmeng

velox/vector/ComplexVector.h
NEUpanning: Thanks! You are right. It is not necessary to reduce children vector size.

velox/vector/FlatVector-inl.h
NEUpanning: https://github.com/facebookincubator/velox/pull/15586

velox/exec/GroupingSet.cpp
NEUpanning: I got why this happened and submitted a PR for it. https://github.com/facebookincubator/velox/pull/15600

velox/exec/GroupingSet.cpp
```diff
@@ -1280,12 +1285,15 @@ bool GroupingSet::mergeNextWithoutAggregates(
   // less than 'numDistinctSpillFilesPerPartition_'.
   bool newDistinct{true};
   int32_t numOutputRows{0};
+  int32_t outputSize{0};
+  bool isEndOfBatch = false;
```
mbasmanova: nit: isEndOfBatch -> endOfBatch

PR #15603, opened 2025-11-22
feat: Add current_time Presto function

anandamideShakyan:
In the past, we were not implementing functions like current_time that returned constants in Velox. This was because Presto Java would infer the constant value at the co-ordinator and fold it, so the worker was never passed this function in execution.

However, with recent changes for side-car and worker based constant folding implementation, we need native implementations for functions like current_time, so adding to Velox as well.

### DIFF
--- a/velox/docs/functions/presto/datetime.rst
+++ b/velox/docs/functions/presto/datetime.rst
@@ -184,6 +184,10 @@ Date and Time Functions
 
     Returns ``timestamp`` as a UNIX timestamp.
 
+.. function:: current_time() -> time with time zone
+
+    Returns the current time since midnight with the session timezoneReturns the current time since midnight with the session timezone, based on the query session start time.
+
 .. function:: current_timezone() -> varchar
 
     Returns the current session time zone as a varchar.
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -457,7 +457,8 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "$internal$split_to_map",
     "$internal$canonicalize",
     "$internal$contains",
-    "localtime", // localtime cannot be called with paranthesis:
+    "current_time", // current_time cannot be called with parenthesis
+    "localtime", // localtime cannot be called with parenthesis:
                  // https://github.com/facebookincubator/velox/issues/14937,
     "localtimestamp", // localtimestamp cannot be called with parenthesis
     "jarowinkler_similarity", // https://github.com/facebookincubator/velox/issues/15736
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -2131,4 +2131,39 @@ struct LocalTimestampFunction {
   Timestamp ts_;
 };
 
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    const tz::TimeZone* timeZone = getTimeZoneFromConfig(config);
+    // Java/Presto session always provides a timezone (TimeZoneKey is required).
+    VELOX_CHECK_NOT_NULL(timeZone);
+
+    auto sessionStartTimeMs = config.sessionStartTimeMs();
+
+    auto localMillis =
+        timeZone->to_local(std::chrono::milliseconds{sessionStartTimeMs});
+    auto localMillisSinceMidnight =
+        localMillis - std::chrono::floor<std::chrono::days>(localMillis);
+
+    auto currentOffset = localMillis.count() - sessionStartTimeMs;
+    // Safe since timezone offsets are bounded [-840, 840] minutes
+    auto currentOffsetMinutes =
+        static_cast<int16_t>(currentOffset / (60 * 1000));
+
+    auto encodedOffset = util::biasEncode(currentOffsetMinutes);
+    currentTime_ = util::pack(localMillisSinceMidnight.count(), encodedOffset);
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<TimeWithTimezone>& result) {
+    result = currentTime_;
+  }
+
+ private:
+  int64_t currentTime_{0};
+};
+
 } // namespace facebook::velox::functions
--- a/velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
@@ -90,6 +90,8 @@ void registerSimpleFunctions(const std::string& prefix) {
       {prefix + "to_unixtime"});
 
   registerFromUnixtime(prefix + "from_unixtime");
+  registerFunction<CurrentTimeFunction, TimeWithTimezone>(
+      {prefix + "current_time"});
   registerFunction<CurrentTimezoneFunction, Varchar>(
       {prefix + "current_timezone"});
   registerFunction<CurrentTimestampFunction, TimestampWithTimezone>(
--- a/velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
@@ -7076,6 +7076,31 @@ TEST_F(DateTimeFunctionsTest, dateAddDateVariableUnit) {
   assertEqualVectors(expected, result);
 }
 
+TEST_F(DateTimeFunctionsTest, currentTime) {
+  auto testCurrentTime = [&](int64_t sessionStartTime,
+                             const std::string& zone,
+                             int64_t expectedMillis,
+                             int16_t expectedOffset) {
+    setSessionStartTimeAndTimeZone(sessionStartTime, zone);
+
+    auto packed = evaluateOnce<int64_t>(
+        "current_time()",
+        makeRowVector(ROW({}), 1),
+        std::nullopt,
+        TIME_WITH_TIME_ZONE());
+
+    ASSERT_TRUE(packed.has_value());
+
+    EXPECT_EQ(util::unpackMillisUtc(packed.value()), expectedMillis);
+    EXPECT_EQ(util::unpackZoneOffset(packed.value()), expectedOffset);
+  };
+
+  testCurrentTime(1710064800000, "UTC", 36000000, 0);
+  testCurrentTime(1710064800000, "Asia/Kolkata", 55800000, 1170);
+  testCurrentTime(1705312800000, "America/Los_Angeles", 7200000, 361);
+  testCurrentTime(1717243200000, "America/Los_Angeles", 18000000, 421);
+}
+
 TEST_F(DateTimeFunctionsTest, currentTimezone) {
   {
     setQueryTimeZone("Asia/Kolkata");


velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2031,4 +2032,29 @@ struct LocalTimeFunction {
...
+      VELOX_USER_FAIL("Timezone cannot be null");
+    }
+    result_ = pack(currentTimeSinceMidnight_, timeZone_->id());
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<TimeWithTimezone>& result) {
+    result = result_;
+  }
+private:
+  int64_t currentTimeSinceMidnight_;
```
aditi-pandit: Why does this need to be a member variable ? Based on the logic doesn't seem like it would be needed.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2031,4 +2032,29 @@ struct LocalTimeFunction {
   int64_t localTimeSinceMidnight_;
 };
 
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+  int64_t result_{0};
+  const tz::TimeZone* timeZone_ = nullptr;
```
aditi-pandit: Does this need to be a member variable as well ? It can be a local variable within initialize() function.

velox/docs/functions/presto/datetime.rst
```diff
@@ -188,6 +188,10 @@ Date and Time Functions
 
         SELECT current_timezone;   -- Asia/Kolkata
 
+.. function:: current_time() -> time with time zone
```
aditi-pandit: This function is alphabetically before current_timezone. Please move it ahead of it.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6833,3 +6833,44 @@ TEST_F(DateTimeFunctionsTest, currentTimezone) {
...
+          throw;
+        }
+      },
+      VeloxException);
+
+  // Test with timezone America/Los_Angeles
+  auto laPacked = callCurrentTime(1710061200000, "America/Los_Angeles");
+  // Reuse TimestampWithTimezone: TIME WITH TIME ZONE uses the same
+  // packed (millis, zoneKey) format, so unpack() works.
+  auto la = TimestampWithTimezone::unpack(laPacked);
```
aditi-pandit: Can you add bunch more test cases in different timezones ?

aditi-pandit (commented):
Thanks @anandamideShakyan. Have few comments.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: You are right, it is not needed will remove it. I was following the implementation of local_time since current_time is just local_time + timezone. I also had a doubt about handling legacytimestamp like it is done in presto java. The velox implementation of local_time doesn't handle legacytimestamp case, also it doesn't handle the conversion to local clock time `ISOChronology localChronology = getChronology(properties.getTimeZoneKey());
        return localChronology.millisOfDay().get(properties.getSessionStartTime());` it simply returns ` the time from midnight in UTC.` .  Why do the implementations diverge, are they handled differently in velox , I guess because velox doesn't support legacytimestamp right ? So I don't need to handle that case here right ? Or am I missing something here ? @kgpai @aditi-pandit

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Changed to local variable.

velox/docs/functions/presto/datetime.rst
anandamideShakyan: Done.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2031,4 +2032,26 @@ struct LocalTimeFunction {
...
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+  int64_t result_{0};
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    auto sessionStartTimeMs = config.sessionStartTimeMs();
+    int64_t currentTimeSinceMidnight_ = sessionStartTimeMs % kMillisInDay;
```
czentgr: currentTimeSinceMidnight_ is not a member so should not have `_` in the name.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6816,6 +6816,62 @@ TEST_F(DateTimeFunctionsTest, dateAddDateVariableUnit) {
...
+    if (timeZone.has_value()) {
+      setSessionStartTimeAndTimeZone(sessionStartTime, timeZone.value());
+    } else {
+      setQuerySessionStartTime(sessionStartTime);
+    }
+
+    auto rowVector = makeRowVector({});
+    rowVector->resize(1);
+
+    auto result = evaluate("current_time()", rowVector);
```
czentgr: You should use `evaluateOnce<int64_t>` instead. That reduces the code quite a bit using this helper.

velox/functions/prestosql/DateTimeFunctions.h
czentgr: The code from the comment is code is from the Java `localtime` and not `current_time` implementation. The `current_time` implementation is here: https://github.com/prestodb/presto/blob/327032c0e8a01d6ce554847036d81a853a562e11/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/DateTimeFunctions.java#L114
It uses UTC and the timezone.
The `localtime` function would convert the session start time (in UTC) to local time using the time zone information.
Also `legacyTimestamp` doesn't come into the picture for `current_time`.

czentgr (commented):
Please also run the fuzzer with the new function and post the results. Instructions are here (step 7): https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md#adding-scalar-functions

Note, the CI does show fuzzer failures. please take a look.
The function is non-deterministic and needs to be excluded from the expression fuzzer?

anandamideShakyan:
I ran the expression fuzzer test , it ran successfully. 
Screenshot 2026-01-24 at 5 00 13 AM
To run the tests locally, I had to remove the function from the `skipFunctions` list in the `ExpressionFuzzerTest`. But in the CI this is causing the error, as the evaluation engines might be different in CI so outputs might be varying.

Screenshot 2026-01-25 at 5 52 50 PM

 That's why I  am adding this to `skipFunctions` list again.

velox/functions/prestosql/registration/DateTimeFunctionsRegistration.cpp
```diff
@@ -346,6 +346,8 @@ void registerSimpleFunctions(const std::string& prefix) {
       {prefix + "parse_duration"});
 
   registerFunction<LocalTimeFunction, Time>({prefix + "localtime"});
+  registerFunction<CurrentTimeFunction, TimeWithTimezone>(
```
aditi-pandit: Is this an alias for now ? If yes, then we should add the alias as well.

aditi-pandit:
@czentgr : PTAL.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6980,6 +6980,61 @@ TEST_F(DateTimeFunctionsTest, dateAddDateVariableUnit) {
...
+
+    return evaluateOnce<int64_t>(
+        "current_time()",
+        makeRowVector(ROW({}), 1),
+        std::nullopt,
+        TIME_WITH_TIME_ZONE());
+  };
+
+  // Test without timezone
+  EXPECT_THROW(
```
czentgr: This is expected to throw. 
Please use macro VELOX_ASSERT_THROW and compare straight to the expected error message ("Timezone cannot be null") for this scenario.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,26 @@ struct LocalTimeFunction {
   int64_t localTimeSinceMidnight_;
 };
 
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+  int64_t result_{0};
```
czentgr: Let's make this a private member of this class - similar to the other functions.

anandamideShakyan: Done.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
anandamideShakyan: Done.

anandamideShakyan:
@czentgr I have resolved the comments, PTAL again.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,28 @@ struct LocalTimeFunction {
...
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    auto sessionStartTimeMs = config.sessionStartTimeMs();
+    int64_t currentTimeSinceMidnight = sessionStartTimeMs % kMillisInDay;
+    const tz::TimeZone* timeZone_ = getTimeZoneFromConfig(config);
```
czentgr: timeZone_ is not a member. Please remove `_`.

velox/docs/functions/presto/coverage.rst
```diff
@@ -436,6 +436,7 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with
     table.coverage tr:nth-child(75) td:nth-child(4) {background-color: #6BA81E;}
     table.coverage tr:nth-child(75) td:nth-child(5) {background-color: #6BA81E;}
     table.coverage tr:nth-child(75) td:nth-child(7) {background-color: #6BA81E;}
+    table.coverage tr:nth-child(76) td:nth-child(1) {background-color: #6BA81E;}
```
czentgr: It was mentioned in another PR to maybe defer this update and put it in a separate PR.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -450,7 +450,8 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "$internal$split_to_map",
     "$internal$canonicalize",
     "$internal$contains",
-    "localtime", // localtime cannot be called with paranthesis:
+    "current_time", // current_time cannot be called with parenthesis
+    "localtime", // localtime cannot be called with parenthesis:
```
anandamideShakyan, line 461: fixed typo: parenthesis

velox/docs/functions/presto/coverage.rst
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Done.

anandamideShakyan:
@mbasmanova I am looking at your comments in the other PR. I will try to resolve them asap. Please review this PR as well in the meantime. Thanks.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,44 @@ struct LocalTimeFunction {
...
 
+template <typename T>
+struct CurrentTimeFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /* type */,
+      const core::QueryConfig& config) {
+    auto sessionStartTimeMs = config.sessionStartTimeMs();
+    int64_t currentTimeSinceMidnight = sessionStartTimeMs % kMillisInDay;
```
kgpai: Please use std::chrono to get time since midnight (rather than doing this arithmetic). See CastExpr::castToTime for this.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,44 @@ struct LocalTimeFunction {
...
+    const tz::TimeZone* timeZone = getTimeZoneFromConfig(config);
+    if (timeZone == nullptr) {
+      VELOX_USER_FAIL("Timezone cannot be null");
```
kgpai: nit: Can you validate that java doesnt allow the timezone to not be set ?

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,44 @@ struct LocalTimeFunction {
...
+    }
+
+    int64_t utcMillisOfDay = sessionStartTimeMs % kMillisInDay;
+    if (utcMillisOfDay < 0) {
+      utcMillisOfDay += kMillisInDay;
+    }
+
+    int32_t offsetAtInstant =
```
kgpai: You should first convert timestamp to local timezone and then use std::chrono to get time of day. Look at TimestampwithTimezoneRegistration for how it is done.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Done.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: Validated Java behavior — Presto session always has a TimeZoneKey (SqlFunctionProperties.getTimeZoneKey() is required). Updated to VELOX_CHECK and added comment.

velox/functions/prestosql/DateTimeFunctions.h
anandamideShakyan: In the java implementation we take the utcmillis and adjust it by difference of : ( dst offset at epoch and at offset at sessionStartMs), I have followed the same in my implementation. 
https://github.com/prestodb/presto/blob/873393fb0123fb8d0621da9fd877b19faf53b831/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/DateTimeFunctions.java#L118C9-L125C1

I have just done this `valueToSessionTimeZoneOffsetDiff()`  calculation below. 

Do you want me to convert to local wall time and store the offset between local and utc time as timezone key ? That would also be accurate because both will give the same values:
Whether we adjust utcmillisofday for offset difference and store the timezonekey, or we store the localmillis along with difference of utcmillis and localmillis(basically current time difference). I think this was done by Natasha in her PR. 

I still think my implementation follows the java implementation more closely, if you still want me to change it let me know.

kgpai: I think its dangerous to do this by hand , especially the computation of offset etc. . I think there are likely edge cases that this logic here is wrong esp when dealing with DST etc. Its far more correct to use timestamp and convert that to local timezone and then get time of day (and also more consistent).

anandamideShakyan: Ok then. I will use that approach.

anandamideShakyan:
@kgpai PTAL again. Thanks.

kgpai:
@anandamideShakyan Will do.

anandamideShakyan:
@kgpai I have addressed the comments. About std::chrono:: readability issue, I see that my implementation is consistent with velox coding style so I haven't made any changes( many other functions in the same file have implemented it without explicitly adding a namespace qualifier). Let me know what you think. PTAL again. Thanks. 

Also sorry if I’ve been nudging a bit too often for reviews — I appreciate your time and feedback.

mbasmanova:
@anandamideShakyan CI is red. Would you take a look?

@kgpai Any remaining comments?

anandamideShakyan:
@mbasmanova The failures are not related to my code changes, I have rebased the code and pushed so hopefully they will go away. PTAL again. Thanks.

kgpai:
@anandamideShakyan Looking.

PR #15605, opened 2025-11-23
feat: Add localtimestamp Presto function

anandamideShakyan:
In the past, we were not implementing functions like localtimestamp that returned constants in Velox. This was because Presto Java would infer the constant value at the co-ordinator and fold it, so the worker was never passed this function in execution.

However, with recent changes for side-car and worker based constant folding implementation, we need native implementations for functions like localtimestamp, so adding to Velox as well.

### DIFF
--- a/velox/docs/functions/presto/datetime.rst
+++ b/velox/docs/functions/presto/datetime.rst
@@ -161,6 +161,10 @@ Date and Time Functions
     using ``hours`` and ``minutes`` for the time zone offset.
     The offset must be in [-14:00, 14:00] range.
 
+.. function:: localtimestamp -> timestamp
+
+    Returns the timestamp as of the start of the query.
+
 .. function:: to_iso8601(x) -> varchar
 
     Formats ``x`` as an ISO 8601 string. Supported types for ``x`` are:
--- a/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
+++ b/velox/expression/fuzzer/ExpressionFuzzerTest.cpp
@@ -454,6 +454,7 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "$internal$contains",
     "localtime", // localtime cannot be called with paranthesis:
                  // https://github.com/facebookincubator/velox/issues/14937,
+    "localtimestamp", // localtimestamp cannot be called with parenthesis
     "jarowinkler_similarity", // https://github.com/facebookincubator/velox/issues/15736
     // Fuzzer and the underlying engine are confused about SetDigest functions
     // (since KHLL is a user defined type), and tries to pass a
--- a/velox/functions/prestosql/DateTimeFunctions.h
+++ b/velox/functions/prestosql/DateTimeFunctions.h
@@ -2113,4 +2113,22 @@ struct LocalTimeFunction {
   int64_t localTimeSinceMidnight_;
 };
 
+template <typename T>
+struct LocalTimestampFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config) {
+    ts_ = Timestamp::fromMillis(config.sessionStartTimeMs());
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<Timestamp>& result) {
+    result = ts_;
+  }
+
+ private:
+  Timestamp ts_;
+};
+
 } // namespace facebook::velox::functions
--- a/velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
@@ -6663,6 +6663,102 @@ TEST_F(DateTimeFunctionsTest, localtime) {
   EXPECT_EQ(localVal, 32400000); // 9 AM UTC
 }
 
+TEST_F(DateTimeFunctionsTest, localTimestamp) {
+  const auto localTimestamp = [&](int64_t sessionStartTime,
+                                  const std::optional<std::string>& timeZone) {
+    if (timeZone.has_value()) {
+      setSessionStartTimeAndTimeZone(sessionStartTime, timeZone.value());
+    } else {
+      setQuerySessionStartTime(sessionStartTime);
+    }
+
+    return evaluateOnce<Timestamp>(
+        "localtimestamp()", makeRowVector(ROW({}), 1));
+  };
+
+  // ============================================================
+  // 1. Basic behavior (no timezone)
+  // ============================================================
+
+  // Epoch — zero nanos
+  {
+    auto tsUtc = localTimestamp(0, std::nullopt);
+    ASSERT_TRUE(tsUtc.has_value());
+    EXPECT_EQ(tsUtc->getSeconds(), 0);
+    EXPECT_EQ(tsUtc->getNanos(), 0);
+  }
+
+  // ============================================================
+  // 2. Basic behavior with timezone
+  // ============================================================
+
+  // Exact second boundary
+  {
+    const int64_t millis = 1758499200000; // exact second
+    auto tsExact = localTimestamp(millis, "America/Los_Angeles");
+
+    ASSERT_TRUE(tsExact.has_value());
+    EXPECT_EQ(tsExact.value().getSeconds(), millis / 1000);
+    EXPECT_EQ(tsExact.value().getNanos(), 0);
+  }
+
+  // Fractional milliseconds (nanos test)
+  {
+    const int64_t millisWithFraction = 1758499200123; // +123 ms
+    auto tsWithNanos =
+        localTimestamp(millisWithFraction, "America/Los_Angeles");
+
+    ASSERT_TRUE(tsWithNanos.has_value());
+    EXPECT_EQ(tsWithNanos.value().getSeconds(), millisWithFraction / 1000);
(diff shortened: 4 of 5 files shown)


anandamideShakyan:
I ran the expression fuzzer test , it ran successfully. 
Screenshot 2026-01-24 at 4 48 04 AM
To run the tests locally, I had to remove the function from the `skipFunctions` list in the `ExpressionFuzzerTest`. But in the CI this is causing the error, as the evaluation engines might be different in CI so outputs might be varying.

Screenshot 2026-01-25 at 5 54 35 PM

 That's why I  am adding this to `skipFunctions` list again.

czentgr:
@anandamideShakyan Please rebase.

velox/functions/prestosql/DateTimeFunctions.h
```diff
@@ -2113,4 +2113,27 @@ struct LocalTimeFunction {
...
+      const core::QueryConfig& config) {
+    auto sessionStartTimeMs = config.sessionStartTimeMs();
+
+    // Convert ms → seconds + nanos
+    seconds_ = sessionStartTimeMs / 1000;
+    nanos_ = (sessionStartTimeMs % 1000) * 1'000'000;
+  }
+
+  FOLLY_ALWAYS_INLINE void call(out_type<Timestamp>& result) {
+    result = Timestamp{seconds_, nanos_};
```
czentgr: Why not save the Timestamp constructed already? Saves the construction and only has one member.

anandamideShakyan: Done. PTAL again.

velox/docs/functions/presto/coverage.rst
```diff
@@ -174,6 +174,7 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with
     table.coverage tr:nth-child(27) td:nth-child(5) {background-color: #6BA81E;}
     table.coverage tr:nth-child(27) td:nth-child(7) {background-color: #6BA81E;}
     table.coverage tr:nth-child(28) td:nth-child(1) {background-color: #6BA81E;}
+    table.coverage tr:nth-child(28) td:nth-child(3) {background-color: #6BA81E;}
```
mbasmanova: Remove changes to coverage.rst. These should come in their own PRs to reduce possibility of conflicts.

velox/docs/functions/presto/datetime.rst
```diff
@@ -201,6 +201,10 @@ Date and Time Functions
         SELECT current_timestamp;  -- 2025-07-17 14:53:12.123 Asia/Kolkata
         SELECT now();              -- 2025-07-17 14:53:12.123 Asia/Kolkata
 
+.. function:: localtimestamp -> timestamp
```
mbasmanova: Make sure functions appear in alphabetical order.

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6663,6 +6663,37 @@ TEST_F(DateTimeFunctionsTest, localtime) {
...
+    if (timeZone.has_value()) {
+      setSessionStartTimeAndTimeZone(sessionStartTime, timeZone.value());
+    } else {
+      setQuerySessionStartTime(sessionStartTime);
+    }
+
+    auto rowVector = makeRowVector({});
+    rowVector->resize(1);
+
+    auto result = evaluate("localtimestamp()", rowVector);
```
mbasmanova: use evaluateOnce

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6663,6 +6663,37 @@ TEST_F(DateTimeFunctionsTest, localtime) {
...
+
+  // Without timezone
+  Timestamp tsUtc = localTimestamp(0, std::nullopt);
+  EXPECT_EQ(tsUtc.getSeconds(), 0);
+  EXPECT_EQ(tsUtc.getNanos(), 0);
+
+  // With timezone : America/Los_Angeles
+  Timestamp tsLocal = localTimestamp(1758499200000, "America/Los_Angeles");
+  EXPECT_EQ(tsLocal.getSeconds(), 1758499200);
+  EXPECT_EQ(tsLocal.getNanos(), (1758499200000 % 1000) * 1'000'000);
```
mbasmanova: > 1758499200000 % 1000

This is confusing. Isn't it just 0?

Can we tests timestamps with and without nanos?

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -452,6 +452,7 @@ std::unordered_set<std::string> skipFunctionsSOT = {
     "$internal$contains",
     "localtime", // localtime cannot be called with paranthesis:
                  // https://github.com/facebookincubator/velox/issues/14937,
+    "localtimestamp", // localtime cannot be called with paranthesis
```
czentgr: typo.  "localtime cannot be called with paranthesis" -> "localtimestamp cannot be called with paranthesis"

velox/functions/prestosql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -6663,6 +6663,37 @@ TEST_F(DateTimeFunctionsTest, localtime) {
...
+}
```
kgpai, line 6760: Can you add a test when we are on the daylight savings boundary  (both forward and backwards case) and ensure that timestamp returned is valid ?

anandamideShakyan: I have added the test cases, I have a doubt though......
I see in java implementations of localtime and localtimetimestamp we store the local wall clock time, whereas in velox implementation of localtime we only return time from midnight in UTC, I followed the same approach here and basically returned the utc timestamp. 
Is there a difference in how velox and presto handle this? Do we just return the utc time/timestamp from velox and the conversion happens upstream ? 
Or I should use to_local().count() to convert the utc timestamp to local wall clock time and store that. Because right now we won't see a change in returned values for DST since we are directly storing the timestamp in seconds and nanoseconds without converting to local. Let me know if I am missing anything.

kgpai: > Is there a difference in how velox and presto handle this? Do we just return the utc time/timestamp from velox and the conversion happens upstream ?

Yes , Velox doesnt support legacy_timestamp based time (See https://github.com/prestodb/presto/issues/25957) . 
Like you pointed out for DST etc, timestamp implementation doesnt change (only the representation based on timezone).

anandamideShakyan:
@mbasmanova @czentgr PTAL again. Thanks.

czentgr (approved):
Thanks!

kgpai:
@anandamideShakyan Should land today. Thanks for your patience.

anandamideShakyan:
> @anandamideShakyan Should land today. Thanks for your patience.

No issues, meanwhile can you review https://github.com/facebookincubator/velox/pull/15603. Thanks.

kgpai:
Hi @anandamideShakyan I think this might require changes to expression fuzzer with presto as SOT. 

```
I0226 05:05:24.446590  1255 PrestoQueryRunner.cpp:485] Execute presto sql: DROP TABLE IF EXISTS t_f3e2e293_6f48_4058_9543_66c801513b9b
I0226 05:05:24.459189  1255 PrestoQueryRunner.cpp:485] Execute presto sql: CREATE TABLE t_f3e2e293_6f48_4058_9543_66c801513b9b(row_number) WITH (format = 'DWRF') AS SELECT cast(null as BIGINT)
I0226 05:05:24.522174  1255 PrestoQueryRunner.cpp:485] Execute presto sql: SELECT "$path" FROM t_f3e2e293_6f48_4058_9543_66c801513b9b
I0226 05:05:24.597878  1255 PrestoQueryRunner.cpp:485] Execute presto sql: DELETE FROM t_f3e2e293_6f48_4058_9543_66c801513b9b
I0226 05:05:24.621125  1255 PrestoQueryRunner.cpp:485] Execute presto sql: SELECT localtimestamp() as p0, row_number as p1 FROM (SELECT row_number as row_number FROM (t_f3e2e293_6f48_4058_9543_66c801513b9b))
E0226 05:05:24.631150  1255 Exceptions.h:53] Line: fbcode/velox/exec/fuzzer/PrestoQueryRunner.cpp:122, Function:throwIfFailed, Expression:  Presto query failed: 1 line 1:23: mismatched input ')'. Expecting: <integer>, Source: RUNTIME, ErrorCode: INVALID_STATE
W0226 05:05:24.631253  1255 PrestoQueryRunner.cpp:463] Query failed in Presto
E0226 05:05:24.631263  1255 ExpressionVerifier.cpp:445] Only reference path threw exception
E0226 05:05:24.631314  1255 Exceptions.h:53] Line: fbcode/velox/expression/tests/ExpressionVerifier.cpp:454, Function:verify, Expression:  Reference path throws for query: SELECT localtimestamp() as p0, row_number as p1 FROM (SELECT row_number as row_number FROM (t_f3e2e293_6f48_4058_9543_66c801513b9b)), Source: RUNTIME, ErrorCode: INVALID_STATE
I0226 05:05:24.631350  1255 ExpressionVerifier.cpp:628] Skipping persistence because repro path is empty.
```

anandamideShakyan:
@kgpai I have already added it to skipFunctionsSOT list, since these functions don't work with parenthesis. Do we need to do anything more ?

kgpai:
@anandamideShakyan Ah ok, Its an internal only fuzzer, so I suppose the change here doesnt affect it ; I will make the change on my end and then attempt to land.

PR #15611, opened 2025-11-24
feat(plan_builder): Support untyped filter expressions

pedroerp:
Summary:
Add plan builder API to enable users to use parsed expressions, in
addition to SQL snippets.

Differential Revision: D87718424

### DIFF
--- a/velox/exec/tests/PlanBuilderTest.cpp
+++ b/velox/exec/tests/PlanBuilderTest.cpp
@@ -13,10 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #include "velox/exec/tests/utils/PlanBuilder.h"
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/core/Expressions.h"
 #include "velox/exec/WindowFunction.h"
+#include "velox/exec/tests/utils/ExpressionBuilder.h"
 #include "velox/exec/tests/utils/HiveConnectorTestBase.h"
 #include "velox/exec/tests/utils/TestIndexStorageConnector.h"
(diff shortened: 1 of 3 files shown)


velox/exec/tests/utils/PlanBuilder.cpp
```diff
@@ -716,15 +716,20 @@ PlanBuilder& PlanBuilder::optionalFilter(const std::string& optionalFilter) {
   return filter(optionalFilter);
 }
 
-PlanBuilder& PlanBuilder::filter(const std::string& filter) {
+PlanBuilder& PlanBuilder::filter(const core::ExprPtr& filterExpr) {
```
mbasmanova, line 719: I assume we'll need this for all APIs that take SQL strings, no?

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -303,6 +305,45 @@ TEST_F(PlanBuilderTest, projectExpressions) {
...
+          .planNode()
+          ->toString(true, false),
+      expectation);
+
+  // Filter with untyped expression (same expression as above).
+  VELOX_CHECK_EQ(
+      PlanBuilder()
+          .tableScan("tmp", data)
+          .filter(
+              std::make_shared<core::CallExpr>(
```
mbasmanova: Can we add ExprAPI-style helper functions to make tests easier to read and maintain? I expect we'll be adding more APIs and tests like this.

pedroerp:
> Thanks. Do you plan to follow-up to add a version that takes ExprPtr for all methods that take SQL?

There are a whole bunch of them. Maybe we can add as needed?

PR #15612, opened 2025-11-24
fix: Fix flaky test freeUnusedCapacityWhenReclaimMemoryPool

duxiao1212:
Summary:
freeUnusedCapacityWhenReclaimMemoryPool was flaky

the main test thread reads stats immediately after joining the allocation thread, but the global arbitration controller updates the reclaimedUsedBytes_ stat asynchronously  later.

1. Main thread: allocThread.join() - waits for allocation to complete
2. Main thread: arbitrator_->stats() - reads stats (returns 0B), and assertion failed
3. Global arbitration controller thread: Updates reclaimedUsedBytes_ = 512MB

From system  causality perspective: stats must reflect the state of the system before other threads can observe it.

Differential Revision: D87809685

### DIFF
--- a/velox/exec/GroupingSet.cpp
+++ b/velox/exec/GroupingSet.cpp
@@ -15,6 +15,7 @@
  */
 #include "velox/exec/GroupingSet.h"
 #include "velox/common/testutil/TestValue.h"
+#include "velox/exec/OperatorUtils.h"
 #include "velox/exec/Task.h"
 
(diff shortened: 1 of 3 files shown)


tanjialiang (approved):
Thanks for the fix. LGTM

xiaoxmeng (approved):
@duxiao1212 thanks for the fix!

PR #15613, opened 2025-11-24
fix(cudf): S3 read test in cudf should depend on s3 being enabled

zoltan:
we should only build velox_cudf_s3_read_test if VELOX_ENABLE_S3 is true.


zoltan:
ping @czentgr @majetideepak

kagamiori:
Hi @zoltan, could you please rebase this PR onto the latest main? It's needed for merging it. Thanks!

PR #15615, opened 2025-11-24
feat(plan_builder): Add fluent expression builder API

pedroerp:
Summary:
Adding an API to allow developers to fluently create (untyped)
expression trees without having to rely on a SQL parser. Details and
examples provided in code comments and unit tests.

Differential Revision: D87828592

### DIFF
--- a/velox/exec/tests/ExpressionBuilderTest.cpp
+++ b/velox/exec/tests/ExpressionBuilderTest.cpp
@@ -1,0 +1,182 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/exec/tests/utils/ExpressionBuilder.h"
+
+#include <gtest/gtest.h>
+#include "velox/parse/ExpressionsParser.h"
+#include "velox/type/Variant.h"
+
+namespace facebook::velox::expr_builder::test {
+namespace {
+
+// Test convenience functions for downcasting.
+template <typename T>
+std::shared_ptr<const T> as(detail::ExprWrapper in) {
+  return std::dynamic_pointer_cast<const T>(in.expr());
+}
+
+template <typename T>
+bool is(detail::ExprWrapper in) {
+  return std::dynamic_pointer_cast<const T>(in.expr()) != nullptr;
+}
+
+// Parses a SQL expression using DuckDB.
+core::ExprPtr parseSql(const std::string& sql) {
+  return parse::parseExpr(sql, {});
+}
+
+TEST(ExpressionBuilderTest, columnReference) {
+  EXPECT_EQ(col("c0"), parseSql("c0"));
+  EXPECT_EQ(parseSql("c0"), col("c0"));
+  EXPECT_EQ("c0"_c, parseSql("c0"));
+
+  EXPECT_EQ(col("parent", "child"), parseSql("parent.child"));
+  EXPECT_EQ(col("parent").subfield("child"), parseSql("parent.child"));
+}
+
+TEST(ExpressionBuilderTest, literals) {
+  auto validate = [](detail::ExprWrapper expr,
+                     const TypePtr& expectedType,
+                     variant expectedValue) {
+    EXPECT_TRUE(is<core::ConstantExpr>(expr));
+    auto constant = as<core::ConstantExpr>(expr);
+    EXPECT_EQ(*constant->type(), *expectedType);
+    EXPECT_TRUE(constant->value().equalsWithEpsilon(expectedValue));
+  };
+
+  // Integer literal types.
+  validate(lit(123456L), BIGINT(), variant(123456L));
+  validate(lit(123), INTEGER(), variant(123));
+  validate(lit(int16_t(123)), SMALLINT(), variant(int16_t(123)));
+  validate(lit(int8_t(123)), TINYINT(), variant(int8_t(123)));
+
+  // Boolean.
+  validate(lit(true), BOOLEAN(), variant(true));
+  validate(lit(false), BOOLEAN(), variant(false));
+
+  // Floating point.
+  validate(lit(10.1f), REAL(), variant(10.1f));
+  validate(lit(10.1), DOUBLE(), variant(10.1));
+
+  // String.
+  validate(lit("str"), VARCHAR(), variant("str"));
+
+  // Null.
+  validate(lit(nullptr), UNKNOWN(), variant::null(TypeKind::UNKNOWN));
+}
+
+TEST(ExpressionBuilderTest, comparisons) {
+  // Make sure all combinations work, as long as at least one side is a
+  // ExprWrapper.
+  EXPECT_EQ(col("a") == lit(10L), parseSql("a = 10"));
+  EXPECT_EQ(lit(10L) == col("a"), parseSql("10 = a"));
+
+  EXPECT_EQ(col("a") == 10L, parseSql("a = 10"));
+  EXPECT_EQ(10L == col("a"), parseSql("10 = a"));
+
+  EXPECT_EQ(col("a") == col("b"), parseSql("a = b"));
+  EXPECT_EQ(col("a") == nullptr, parseSql("a = null"));
+
+  // Other comparisons.
+  EXPECT_EQ(col("a") != 1.1, parseSql("a != 1.1"));
+  EXPECT_EQ(col("a") != lit(1.1), parseSql("a != 1.1"));
+  EXPECT_EQ(col("a") > 42L, parseSql("a > 42"));
+  EXPECT_EQ(col("a") >= 42L, parseSql("a >= 42"));
+  EXPECT_EQ(col("a") < 42L, parseSql("a < 42"));
+  EXPECT_EQ(col("a") <= 42L, parseSql("a <= 42"));
+
+  EXPECT_EQ(!col("a"), parseSql("not a"));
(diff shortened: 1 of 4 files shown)


velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,350 @@
...
+
+#include "velox/parse/Expressions.h"
+
+namespace facebook::velox::expr_builder {
+
+/// Fluent Expression Builder.
+///
+/// This file contains fluent methods that make it convenient to create
+/// (untyped) expression trees. This provides similar functionality to a SQL
+/// parser, without briging dependency on external libraries or bringing leaked
```
mbasmanova: typo: "briging"

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,350 @@
...
+/// semantics from other systems.
+///
+/// The untyped expressions can then be turned into typed expressions ready for
+/// execution using type binding from `core::Expressions::inferTypes()`.
+///
+/// The API provided is as close to the actual expression trees as possible.
+/// Filters, arithmetics, function calls, literals, aliases, and more are
```
mbasmanova: What does "Filters" refer to here?

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,350 @@
...
+/// supported with this API.
+///
+/// For example, to create a column reference, you can:
+///
+/// > using namespace expr_builder;
+/// > ExprPtr e = col("c0");
```
mbasmanova: Is it necessary to use ExprPtr? Will auto also work? If not, perhaps, explain why.

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,350 @@
...
+/// Nested column references can be specified as either:
+///
+/// > ExprPtr e = col("parent", "child");
+/// > ExprPtr e = col("parent").subfield("child");
+///
+/// To debug the expression generated, you can simply:
+///
+/// > LOG(INFO) << *e;
+///
+/// Comparisons and other expressions can be fluently created using C++
```
mbasmanova, line 59: Would you clarify how these are translated to function calls? I assume we hard-code a mapping from C++ operator to a function name and that mapping may not work for some systems. Specifically, Spark seems to use different names for == and similar. See gluten/cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,350 @@
...
+/// overloaded operators:
+///
+/// > ExprPtr e = col("c") > 10; // "c > 10"
+/// > ExprPtr e = col("c") != "bar"; // "c != 'bar'"
+/// > ExprPtr e = col("c") == nullptr; // "c = null"
+///
+/// C++ literals are automatically converted into ConstantExpr (expression
```
mbasmanova, line 66: I believe this doesn't work when literal is the first expression. Does 1 < col("c") work? If not, perhaps, clarify.

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: I meant "comparisons". Will fix

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,387 @@
...
+/// > ExprPtr e = cast("str", VARBINARY());
+///
+/// Null checking filters:
+///
+/// > ExprPtr e = isNull(col("c")) // "c is null"
+/// > ExprPtr e = !isNull(col("c")) // "c is not null"
+///
+/// Conjuncts and "between":
+///
+/// > ExprPtr e = col("a") && col("b") || col("c"); // "a AND b OR c"
```
mbasmanova: this feel error prone; perhaps, use explicit parentheses to clarify the intended order of operations

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: yeah, auto also works. You can also just pass it to any function that takes a ExprPtr as expected. I'll clarify.

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,387 @@
...
+/// Both will generate "col + 5 * 100", which is "plus(col, multiply(5, 100))".
+///
+/// Generic function calls can be created using `call()`:
+///
+/// > ExprPtr e = call("func", 10); // "func(10)"
+///
+/// `call()` supports arbitrary parameters, which can be other expressions or
+/// (C++) literals.
+///
+/// All functions above can be nested and combined in arbitrary ways.
```
mbasmanova, line 130: Do we support lambdas?

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,387 @@
...
+///
+/// > ExprPtr e = 10L * col("c1") > call(
+///     "func", 3.4, col("g") / col("h"), call("j"));
+///
+/// is the same as "10 * c1 > func(3.4, g / h, j())".
+
+using core::ExprPtr;
```
mbasmanova: Can we avoid 'using' in a header file?

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: Will clarify. For now I'm just using the Presto function names "eq/neq/gt/lt/..."

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,387 @@
...
+
+class ExprWrapper;
```
mbasmanova, line 162: perhaps, put this into 'detail' namespace

velox/exec/tests/utils/ExpressionBuilder.h
```diff
@@ -0,0 +1,387 @@
...
+
+// Either builds a ConstantExpr (literal) based on a scalar value, or passes
```
mbasmanova: /// ?

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: yeah, that also works. As long as any of the operands is an expression wrapper. The only thing that won't work, naturally, is two C++ literals, like "1 + 10", because C++ will just constant fold them and you'll get a 11 literal expression. I have this documented in this header.

velox/exec/tests/ExpressionBuilderTest.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <gtest/gtest.h>
+#include "velox/parse/ExpressionsParser.h"
+#include "velox/type/Variant.h"
+
+namespace facebook::velox::expr_builder::test {
+namespace {
+
+// Test convenience functions for downcasting.
+template <typename T>
+std::shared_ptr<const T> as(ExprWrapper in) {
```
mbasmanova: Any particular reason not to make these methods of ExprWrapper ?

velox/exec/tests/utils/ExpressionBuilder.h
mbasmanova: > without briging dependency on external libraries or bringing leaked
> semantics from other systems.

Perhaps, revise this statement as it doesn't seem accurate.

velox/parse/Expressions.cpp
```diff
@@ -88,7 +88,11 @@ size_t CastExpr::localHash() const {
 
 std::string CastExpr::toString() const {
   return appendAliasIfExists(
-      "cast(" + input()->toString() + " as " + type_->toString() + ")");
+      fmt::format(
```
pedroerp, line 91: this was a legit bug; we were not printing try_cast() correctly.

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: not yet, but we can come up with a fluent lambda API. I can add as a next PR.

 We can' support a native actual C++ lambda because they get compiled right away.

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: Actually, I take that back. This is still semantic-agnostic; it just happens that some C++ operators will be mapped to a pre-fixed function name. Now, it's the user's responsibility to make sure that there is a function with the expected semantic  registered there. 

I'll clarify on the comments.

mbasmanova: > it just happens that some C++ operators will be mapped to a pre-fixed function name

This is a new requirement. Velox currently doesn't require that users register functions with specific names.

mbasmanova: Also, note that TRY and TRY_CAST are Presto-specific special forms. These don't even exist in Spark.

pedroerp: @mbasmanova 

> This is a new requirement. Velox currently doesn't require that users register functions with specific names.

This isn't a requirement here as well. It will just create the untyped expressions with those names. The functions themselves will only be needed if the user tries to execute them. 

> Also, note that TRY and TRY_CAST are Presto-specific special forms. These don't even exist in Spark.

Same here. This library will just create the untyped expressions (CastExpr).

velox/exec/tests/utils/ExpressionBuilder.h
pedroerp: Not really; it will follow operator precedence rules in C++, which are deterministic and well defined. I can add parenthesis  here for clarity though.

velox/exec/tests/ExpressionBuilderTest.cpp
pedroerp: These are only needed here for one specific test suite. If they are needed/useful for actually building expressions we could move, but I haven't seen a need for it yet.

velox/exec/tests/utils/ExpressionBuilder.h
mbasmanova: It would be nice to add parenthesis for clarity. Not everybody has these rules at the top of their head.

PR #14582, opened 2025-08-25
fix: Initialize errors only once in JsonCastOperator

xiaodouchen:
Summary:
Ensure that the `errors` array within the `JsonCastOperator` is initialized 
only once, avoiding significant overhead when casting multiple columns.

This PR fixes https://github.com/facebookincubator/velox/issues/14581

PR #15617, opened 2025-11-25
fix: Exception handling in cast from JSON

bikramSingh91:
Summary:
A recent change https://github.com/facebookincubator/velox/pull/14582
made the exception initialization codepath create and maintain a
static list of exceptions shared by all Cast to JSON expressions.
Since Velox error messages add additional context when generated,
the static exception ends up retaining the message which contains
the expression and the top level expression details of the very
first error produced. This then ends up being re-used across
expression objects and across different query plan resulting in
wrong error message being thrown.
This change reverts the original PR to ensure exceptions with the
expected context are thrown.

Test Plan:
Added unit tests for both cast to json and json_parse as both have
similar exception generation pattern.

Differential Revision: D87838044

### DIFF
--- a/velox/functions/prestosql/tests/JsonCastTest.cpp
+++ b/velox/functions/prestosql/tests/JsonCastTest.cpp
@@ -1822,3 +1822,18 @@ TEST_F(JsonCastTest, castFromJsonWithEscapingForSpecialUniocodeCharacters) {
   testCast(R"(["\u007F"])", "\u007F");
   testCast(R"(["\u008A"])", "\u008A");
 }
+
+TEST_F(JsonCastTest, uniqueErrorContextMessage) {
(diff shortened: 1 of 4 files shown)


velox/functions/prestosql/tests/JsonCastTest.cpp
```diff
@@ -1822,3 +1822,18 @@ TEST_F(JsonCastTest, castFromJsonWithEscapingForSpecialUniocodeCharacters) {
   testCast(R"(["\u007F"])", "\u007F");
   testCast(R"(["\u008A"])", "\u008A");
 }
+
+TEST_F(JsonCastTest, testUniqueErrorContextMessage) {
```
mbasmanova: naming: drop 'test' prefix

Yuhta (approved):

PR #15621, opened 2025-11-25
fix: Fix memory accounting of buffer mode in LocalPartition

kagamiori:
Summary:
LocalPartition in buffer-mode appends input rows to the corresponding partition buffer vectors and 
flush the buffer vectors to LocalExchangeQueue when their sizes exceeds a limit. This diff fixes two 
problem with the buffer-mode:
1. When a string buffer is referenced by multiple partition buffer vectors, the entire string buffer 
size is counted once per partition buffer vector. This makes the sizes of partition buffer vectors over 
estimated and prevents the buffer-mode from effectively reducing the number of vectors flushed to LocalExchangeQueues. This diff addresses this problem by amortizing the string buffer size when 
calculating the sizes of partition buffer vectors.
2. LocalPartition checks individual partition buffer vector size and decide whether to flush it. After 
we amortize the string buffer size, this could cause the partition buffers that have few rows to hold 
reference to string buffers for too long, hence not able to release the string buffers to reduce 
memory usage. This diff changes LocalPartition to check the total size of all partition buffers and 
flush all buffers once the total size exceeds the memory limit.

Differential Revision: D87805165

### DIFF
--- a/velox/exec/LocalPartition.cpp
+++ b/velox/exec/LocalPartition.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "velox/exec/LocalPartition.h"
+#include "velox/common/Casts.h"
 #include "velox/exec/Task.h"
 #include "velox/vector/EncodedVectorCopy.h"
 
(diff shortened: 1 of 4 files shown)


velox/exec/LocalPartition.cpp
```diff
@@ -438,49 +438,137 @@ void LocalPartition::copy(
...
-    vector_size_t size,
+    vector_size_t numPartitionRows,
     int partition,
-    const BufferPtr& indices,
-    const vector_size_t* rawIndices) {
+    const vector_size_t* rawIndices,
+    uint64_t& totalPartitionBufferSizeExcludingString,
+    uint64_t& totalPartitionStringBufferSize) {
+  VELOX_CHECK_GT(singlePartitionBufferSize_, 0);
+  if (partitionBuffers_.empty()) {
```
Yuhta: nit: This should be outside the loop over partitions

velox/exec/LocalPartition.cpp
```diff
@@ -438,49 +438,137 @@ void LocalPartition::copy(
...
+    targetIndex = partitionBuffer->size();
+  }
+  for (int i = 0; i < numPartitionRows; i++) {
+    copyRanges_[i] = {rawIndices[i], targetIndex, 1};
+    targetIndex++;
+  }
+
+  copy(
+      input,
+      folly::Range{copyRanges_.data(), static_cast<size_t>(numPartitionRows)},
```
Yuhta: nit: Just do `copy(input, copyRanges_, partitionBuffer)` should be enough once you resize the `copyRanges_` unconditionally

velox/exec/LocalPartition.cpp
```diff
@@ -438,49 +438,137 @@ void LocalPartition::copy(
...
-    const vector_size_t* rawIndices) {
+    const vector_size_t* rawIndices,
+    uint64_t& totalPartitionBufferSizeExcludingString,
+    uint64_t& totalPartitionStringBufferSize) {
+  VELOX_CHECK_GT(singlePartitionBufferSize_, 0);
+  if (partitionBuffers_.empty()) {
+    partitionBuffers_.resize(numPartitions_);
+  }
+  if (copyRanges_.size() < numPartitionRows) {
+    copyRanges_.resize(numPartitionRows);
```
Yuhta: Just resize it unconditionally, the capacity will not shrink

velox/exec/LocalPartition.cpp
```diff
@@ -438,49 +438,137 @@ void LocalPartition::copy(
...
+  } else {
+    nonEmptyPartitionCount = numPartitions_ -
+        std::count(numRowsPerPartition.begin(), numRowsPerPartition.end(), 0);
+  }
+  VELOX_CHECK_NE(
+      nonEmptyPartitionCount,
+      0,
+      "Input rows should be assigned to at least one partition");
+
+  // Calculat the partition buffer size across all partitions with amortized
```
Yuhta: typo: Calculat

velox/exec/LocalPartition.cpp
```diff
@@ -438,49 +438,137 @@ void LocalPartition::copy(
...
-    }
-
-    copy(
-        input,
-        folly::Range{copyRanges_.data(), static_cast<size_t>(size)},
-        partitionBuffer);
-
-    if (partitionBuffer &&
-        partitionBuffer->retainedSize() >= singlePartitionBufferSize_) {
       partitionData = std::dynamic_pointer_cast<RowVector>(partitionBuffer);
```
Yuhta: nit: We can use `checkedPointerCast` in `velox/common/Casts.h` now

kagamiori:
cc @arhimondr

PR #15625, opened 2025-11-25
fix: Do not wait for async load at ParallelUnitLoader destructor

kewang1024:
Summary:
**1. DEADLOCK**
We have observed hanging when parallelUnitLoader is turned on ( [stacktrace](https://www.internalfb.com/intern/everpaste/?handle=GGR-SCIae13bMmFZAKsB6AQpWsMJbsIXAAAB&phabricator_paste_number=2015904911)). It hanged because there is a circular dependency when memory arbitration is triggered (P2054964327)
- HashBuild triggers memory arbitration
- It acquires the **arbitration lock** and reclaim memory from TableScan
- That triggers TableScan's destruction which triggers ParallelUnitLoader destruction
- ParallelUnitLoader destructor waits for async load to finish
- the async load triggers memory arbitration again and tries to acquire the same **arbiration lock**

To break the dependency, we no longer wait for asyc load operation at ParallelUnitLoader destruction.

**2. LIFE CYCLE**
To make sure asyc load operation can still function after ParallelUnitLoader destruction and DwrfRowReader destruction:
- Capture DwrfUnit as shared_pointer to keep it alive after ParallelUnitLoader is destroyed
- Make DwrfUnit hold onto the used components:
   - Pass ReaderBase as shared pointer
    - Pass ColumnReaderStatistics as shared pointer
    - Pass columnReaderOptions as copy instead of reference

Differential Revision: D87828770

### DIFF
--- a/velox/dwio/common/ParallelUnitLoader.cpp
+++ b/velox/dwio/common/ParallelUnitLoader.cpp
@@ -17,6 +17,7 @@
 #include "velox/dwio/common/ParallelUnitLoader.h"
 #include <numeric>
 #include "velox/common/base/Exceptions.h"
+#include "velox/common/testutil/TestValue.h"
 #include "velox/common/time/Timer.h"
 
 namespace facebook::velox::dwio::common {
@@ -55,7 +56,9 @@ class ParallelUnitLoader : public UnitLoader {
       std::vector<std::unique_ptr<LoadUnit>> units,
       folly::Executor* ioExecutor,
       uint16_t maxConcurrentLoads)
-      : loadUnits_(std::move(units)),
+      : loadUnits_(
+            std::make_move_iterator(units.begin()),
+            std::make_move_iterator(units.end())),
         ioExecutor_(ioExecutor),
         maxConcurrentLoads_(maxConcurrentLoads) {
     VELOX_CHECK_NOT_NULL(ioExecutor, "ParallelUnitLoader ioExecutor is null");
@@ -74,7 +77,6 @@ class ParallelUnitLoader : public UnitLoader {
     for (auto& future : futures_) {
       if (future.valid()) {
         future.cancel();
-        future.wait();
       }
     }
   }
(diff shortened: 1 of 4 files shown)


velox/dwio/common/ParallelUnitLoader.cpp
```diff
@@ -55,7 +55,9 @@ class ParallelUnitLoader : public UnitLoader {
       std::vector<std::unique_ptr<LoadUnit>> units,
       folly::Executor* ioExecutor,
       uint16_t maxConcurrentLoads)
-      : loadUnits_(std::move(units)),
+      : loadUnits_(
+            std::make_move_iterator(units.begin()),
```
xiaoxmeng, line 60: can we use std::move() here?

velox/dwio/common/ParallelUnitLoader.cpp
```diff
@@ -141,9 +142,12 @@ class ParallelUnitLoader : public UnitLoader {
...
     VELOX_CHECK_LT(unitIndex, loadUnits_.size(), "Unit index out of bounds");
     VELOX_CHECK_NOT_NULL(ioExecutor_, "ParallelUnitLoader ioExecutor is null");
+    VELOX_DCHECK(!loadUnits_.empty(), "loadUnits_ should not be empty");
 
-    futures_[unitIndex] = folly::via(
-        ioExecutor_, [this, unitIndex]() { loadUnits_[unitIndex]->load(); });
+    // Capture shared_ptr by value to prevent use-after-free if
+    // ParallelUnitLoader is destroyed while async operation is running
+    auto unit = loadUnits_[unitIndex];
+    futures_[unitIndex] = folly::via(ioExecutor_, [unit]() { unit->load(); });
```
xiaoxmeng:
```
unit = loadUnits_[unitIndex]
```

velox/dwio/dwrf/reader/DwrfReader.cpp
```diff
@@ -141,9 +149,11 @@ void DwrfUnit::ensureDecoders() {
 
   preloaded_ = options_.preloadStripe();
 
+  // Create a temporary StripeReaderBase wrapper to fetch stripe
+  StripeReaderBase stripeReaderBase(readerBase_);
```
xiaoxmeng: can you create StripeReaderBase DwrfUnit from readerBase_?

xiaoxmeng (approved):
@kewang1024 LGTM % minors. Could you address before land? Thanks!

velox/dwio/common/ParallelUnitLoader.cpp
kewang1024: we can't because we need to convert each element from unique_ptr<LoadUnit> to shared_ptr<LoadUnit>

velox/dwio/common/ParallelUnitLoader.cpp
kewang1024: could you clarify the changes? it seems same as mine

PR #15627, opened 2025-11-25
feat(build): Add `SIMDJSON_SKIPUTF8VALIDATION` option for simdjson in setup script

philo-he:
In the existing code below, the `SIMDJSON_SKIPUTF8VALIDATION` option for simdjson is determined through the setting for `VELOX_SIMDJSON_SKIPUTF8VALIDATION`. https://github.com/facebookincubator/velox/blob/13ac97fe4441b983aa19747dde0bf7ae2aef4ea4/CMake/resolve_dependency_modules/simdjson.cmake#L36-L38

For consistency, we should also allow setting this build option in setup script. For Spark users, it must be enabled to allow invalid UTF8 encoding in JSON data, which is necessary for alignment with Spark.


philo-he:
@majetideepak, could you take a look?

PR #15628, opened 2025-11-25
refactor: Improve VeloxPromise API

mbasmanova:
Summary:
- Request context in makeVeloxContinuePromiseContract.
- Do not require context for empty promises.

Helps debug "Broken promise" exceptions.

Differential Revision: D87861045

### DIFF
--- a/velox/common/future/VeloxPromise.h
+++ b/velox/common/future/VeloxPromise.h
@@ -28,12 +28,15 @@ class VeloxPromise : public folly::Promise<T> {
   VeloxPromise() : folly::Promise<T>() {}
 
   explicit VeloxPromise(const std::string& context)
-      : folly::Promise<T>(), context_(context) {}
+      : folly::Promise<T>(), context_(context) {
+    if (context.empty()) {
+      LOG(WARNING)
+          << "PROMISE: VeloxPromise must be constructed with a context.";
+    }
+  }
 
-  VeloxPromise(
-      folly::futures::detail::EmptyConstruct,
-      const std::string& context) noexcept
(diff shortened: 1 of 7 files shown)


velox/exec/ExchangeQueue.h
```diff
@@ -197,7 +197,9 @@ class ExchangeQueue {
   }
 
   std::vector<ContinuePromise> clearAllPromisesLocked() {
-    std::vector<ContinuePromise> promises(promises_.size());
+    std::vector<ContinuePromise> promises;
+    promises.reserve(promises_.size());
```
MBkkt, line 201: it's bugfix? not refactor

before your changes it was vector with `size == promises_.size() * 2`, now it's `size == promises_.size()`

velox/common/future/VeloxPromise.h
```diff
@@ -28,12 +28,15 @@ class VeloxPromise : public folly::Promise<T> {
   VeloxPromise() : folly::Promise<T>() {}
```
MBkkt, line 28: remove default ctor? or at least deprecate it?

velox/exec/ExchangeQueue.h
mbasmanova: Yes, this is a bug fix.

velox/common/future/VeloxPromise.h
mbasmanova: I tried, but couldn't do this because the following usage pattern is quite common.

```
VeloxPromise p;
...
p = std::move(...);
```

PR #15634, opened 2025-11-26
feat(plan_builder): Alias support for expression builder

pedroerp:
Summary:
Adding support for alias() and .alias() in expression builder. It
requires a new API in IExpr because IExpr are always immutable.

Differential Revision: D87737347

### DIFF
--- a/velox/parse/Expressions.h
+++ b/velox/parse/Expressions.h
@@ -79,6 +79,10 @@ class FieldAccessExpr : public IExpr {
         name_, alias(), std::move(newInputs));
   }
 
+  ExprPtr withAlias(const std::string& alias) const override {
+    return std::make_shared<FieldAccessExpr>(name_, alias, inputs());
+  }
+
   ExprPtr dropAlias() const override {
     return std::make_shared<FieldAccessExpr>(name_, std::nullopt, inputs());
   }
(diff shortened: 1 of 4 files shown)


velox/parse/IExpr.h
```diff
@@ -85,6 +85,9 @@ class IExpr {
   /// Returns a copy of this expression with the given inputs.
   virtual ExprPtr replaceInputs(std::vector<ExprPtr> newInputs) const = 0;
 
+  /// Returns a copy of this expression with the the new alias added.
```
mbasmanova, line 88: nit: perhaps, clarify, that this can be called repeatedly and the last "alias" wins.

I assume there is no need to clear an alias, is there?

velox/parse/Expressions.h
```diff
@@ -33,6 +33,10 @@ class InputExpr : public IExpr {
     return std::make_shared<InputExpr>();
   }
 
+  ExprPtr withAlias(const std::string&) const override {
+    return std::make_shared<InputExpr>();
```
mbasmanova: Should we fail instead of ignoring 'alias'? Same re: lambda.

mbasmanova (approved):

velox/parse/IExpr.h
pedroerp: > I assume there is no need to clear an alias, is there?

there is an existing "dropAlias()" method below to achieve that. I think you added it for Verax at some point.

PR #15638, opened 2025-11-26
perf: Avoid string copy in IPAddressType

pedroerp:
Summary: Avoiding unnecessary string copy from StringView to std::string.

Differential Revision: D87908833


amitkdutta (approved):
Thanks @pedroerp

PR #15640, opened 2025-11-26
Batch serialization with preserve encoding option support in remote functions

zation99:
Summary:
I noticed that the constant encodings were not preserved when we are receiving thrift request from Presto for remote function calls. We use a lot of constant parameters in our remote funciton udf and if we don't preserve the constant encoding, it will more than double the payload size.
After debugging the issue, the fix is to use batch serializer (instead of iterative serializer) which preserves the encoding correctly. I noticed 60% reduced in payload size after testing this change locally for our remote function udf. This is just a simple case, in other cases we will have even more savings (more than 3x reduction in payload size).

Differential Revision: D87873415

### DIFF
--- a/velox/vector/VectorStream.cpp
+++ b/velox/vector/VectorStream.cpp
@@ -313,6 +313,23 @@ folly::IOBuf rowVectorToIOBuf(
   return std::move(*stream.getIOBuf());
 }
 
+folly::IOBuf rowVectorToIOBufUsingBatchSerializer(
+    const RowVectorPtr& rowVector,
+    memory::MemoryPool& pool,
+    VectorSerde* serde,
+    const VectorSerde::Options* options) {
+  // Use BatchVectorSerializer with provided options (e.g., preserveEncodings)
+  // to maintain encoding through serialization
+  auto serializer = serde->createBatchSerializer(&pool, options);
+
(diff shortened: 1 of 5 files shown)


velox/vector/VectorStream.cpp
```diff
@@ -313,6 +313,23 @@ folly::IOBuf rowVectorToIOBuf(
...
 }
 
+folly::IOBuf rowVectorToIOBufUsingBatchSerializer(
+    const RowVectorPtr& rowVector,
+    memory::MemoryPool& pool,
+    VectorSerde* serde,
+    const VectorSerde::Options* options) {
+  // Use BatchVectorSerializer with provided options (e.g., preserveEncodings)
+  // to maintain encoding through serialization
+  auto serializer = serde->createBatchSerializer(&pool, options);
```
pedroerp, line 323: this is a good catch, but I'm thinking we probably want to fix `rowVectorToIOBuf()` directly.

Check the comment on velox/vector/VectorStream.h:496. The "single-batch" serialization was meant to be the expected behavior of `rowVectorToIOBuf()`. Let's fix if there and keep a single version of the function. It seems like the remote function adapter is the only place where we use it in any case.

PR #15641, opened 2025-11-26
refactor: Add SimpleFunction signature with MemoryPool

natashasehgal:
Summary:
Add MemoryPool support to SimpleFunction `initialize()` method to enable memory tracking for scalar functions (e.g merge_hll) that use HashStringAllocator.

Some types that require this include Hyperloglog, KHyperloglog, SetDigest, SfmSketch etc. 

**Changes** 
SimpleFunctionMetadata/SimpleFunctionAdapter/VectorFunction
- New function signature with MemoryPool
- Function caller tries method with MemoryPool first then without. 

ExprCompiler
- Gets MemoryPool from Execution Context passes through function factories
- Create MemoryPool reference (with special no-op deleter so it doesn't try to delete memory it doesn't own)

ExprCompilerTest - Added test here

Differential Revision: D87932003

### DIFF
--- a/velox/core/SimpleFunctionMetadata.h
+++ b/velox/core/SimpleFunctionMetadata.h
@@ -881,14 +881,53 @@ class UDFHolder {
         (udf_has_callAscii_return_void && udf_has_call_return_bool)),
       "The return type for callAscii() must match the return type for call().");
 
-  // initialize():
-  static constexpr bool udf_has_initialize = util::has_method<
+  // Detects if initialize() is a template method using SFINAE.
+  // Template methods can match any signature via template parameter deduction,
+  // causing false positives in trait detection. We probe with a dummy type
+  // that's not in our expected signature to identify templates.
+  struct DummyProbeType {};
+
+  template <typename U, typename = void>
+  struct has_template_initialize : std::false_type {};
+
+  template <typename U>
+  struct has_template_initialize<
+      U,
+      util::detail::void_t<decltype(std::declval<U>().initialize(
+          std::declval<const std::vector<TypePtr>&>(),
+          std::declval<const core::QueryConfig&>(),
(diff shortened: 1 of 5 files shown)


velox/core/SimpleFunctionMetadata.h
```diff
@@ -881,15 +881,58 @@ class UDFHolder {
...
+          memory::MemoryPool*,
+          const exec_arg_type<TArgs>*...>::value;
+
+  // Combined trait for backward compatibility: true if ANY initialize exists
+  // This preserves the original meaning of udf_has_initialize
+  static constexpr bool udf_has_initialize =
+      udf_has_initialize_with_pool || udf_has_initialize_without_pool;
+
+  // Alias (deprecated - use udf_has_initialize_without_pool)
+  static constexpr bool udf_has_initialize_no_pool =
```
Yuhta: Just do not add this?

velox/expression/ExprCompiler.cpp
```diff
@@ -302,8 +303,11 @@ ExprPtr compileCall(
         resultType,
         folly::join(", ", inputTypes));
 
+    // Create a shared_ptr as the memory pool is managed by the ExecCtx
+    auto poolPtr = std::shared_ptr<memory::MemoryPool>(
```
Yuhta: If we cannot control the life cycle, we should just pass `memory::MemoryPool*` instead of `shared_ptr` to functions instead of faking it.

natashasehgal: Makes sense, using memory pool direclty

velox/core/SimpleFunctionMetadata.h
natashasehgal: Removed this

PR #15644, opened 2025-11-26
feat: Add fuzzer type transform for KHLL

natashasehgal:
Summary:
Add intermediate type transform for KHyperLogLog to enable fuzzer testing.

**Change**
Adds KHyperLogLog <-> VARBINARY transform for Presto Java compatibility. KHLL is an intermediate type only (used for computation). Storage in hive column tables is as VARBINARY.

Also excludes KHLL from isConstantExprSupported() since it cannot be used as a constant literal in SQL (e.g., WHERE column = KHYPERLOGLOG X'...').

Follows the same pattern as SetDigest (D87554118)

Differential Revision: D87946570


HeidiHan0000 (approved):
Thanks so much for the change!

PR #15645, opened 2025-11-26
feat(cudf): Add install_ucx helper

zoltan:
The upcoming hardware-accelerated exchange is based on ucx, so we need a helper to set it up.


scripts/setup-centos-adapters.sh
```diff
@@ -25,13 +25,49 @@
...
+      CUDA_FLAG="--with-cuda=/usr/local/cuda"
+    fi
+
+    mkdir build-linux && cd build-linux
+
+    ../contrib/configure-release --prefix="${INSTALL_PREFIX}" --with-sysroot --enable-cma \
+      --enable-mt --with-gnu-ld --with-rdmacm --with-verbs \
+      --without-go --without-java ${CUDA_FLAG}
+    make "-j${NPROC}"
+    ${SUDO} make install
```
czentgr: ${SUDO} is not required for CentOS.

zoltan: fixed!

scripts/setup-centos-adapters.sh
```diff
@@ -25,13 +25,49 @@
...
+#   UCX_VERSION from the env
 
 set -efx -o pipefail
 
 VELOX_CUDA_VERSION=${CUDA_VERSION:-"12.8"}
+VELOX_UCX_VERSION=${UCX_VERSION:-"1.19.0"}
 SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 source "$SCRIPT_DIR"/setup-centos9.sh
 
+function install_ucx {
```
jinchengchenghh, line 38: I don't see where to call the function.

zoltan: this will be called from presto for now. we're trying to get the individual pieces upstreamed.

jinchengchenghh (approved):

Issue #15648, opened 2025-11-26
IN rewrite causes behavior and result change


kagamiori:
### Bug description

The recently introduced writer for IN predicate (https://github.com/facebookincubator/velox/pull/15488) causes behavior change (null vs. exception) and result change (non-null vs. null). These changes are found by expression fuzzer.

#### Result change (bug)
The result change is due to a bug in InRewrite that it doesn't respect null-as-indeterminate when comparing complex-typed values.
https://github.com/facebookincubator/velox/blob/fc70dbe02bb6cb3eaa4ee25ed06dd022ca5b3136/velox/functions/prestosql/InPredicate.h#L45-L46

Below is a repro. This test fails with InRewrite and passes after disabling InRewrite.
```
TEST_F(InPredicateTest, aaa) {
  auto c0 = BaseVector::wrapInConstant(
      1, 0, makeMapVectorFromJson<int32_t, int32_t>({"{0: null}"}));
  auto c1 = BaseVector::wrapInConstant(
      1, 0, makeMapVectorFromJson<int32_t, int32_t>({"{0: 1}"}));
  auto c2 = makeMapVectorFromJson<int32_t, int32_t>({"{0: 1, 10: 2}"});

  auto expr = std::make_shared<core::CallTypedExpr>(
      BOOLEAN(),
      "in",
      std::make_shared<core::ConstantTypedExpr>(c0),
      std::make_shared<core::ConstantTypedExpr>(c1),
      std::make_shared<core::FieldAccessTypedExpr>(c2->type(), "c2"));
  auto result = evaluate(expr, makeRowVector({"c0", "c1", "c2"}, {c0, c1, c2}));
  assertEqualVectors(makeNullConstant(TypeKind::BOOLEAN, 1), result);
}
```

#### Behavior change
The rewrite for IN also introduced a behavior change about NULL value. In Velox, the IN function receives an argument `value` followed by a variadic argument list `inList`. This IN function does not have default-null behavior because a NULL in the inList doesn't imply a NULL result.
https://github.com/facebookincubator/velox/blob/fc70dbe02bb6cb3eaa4ee25ed06dd022ca5b3136/velox/functions/prestosql/InPredicate.h#L33-L36

On the other hand, InRewrite returns a NULL if `value` is NULL. This causes a behavior change on an expression like the following:
`in(null, subexpression_always_throw(...), ...)`
[...]

PR #15649, opened 2025-11-26
Revert "feat: Add rewrite for IN special form"

kagamiori:
Summary:
This diff reverts D87598240
This change introduced incorrect behavior as described in https://github.com/facebookincubator/velox/issues/15648.

Depends on D87598240

Differential Revision: D87955860


velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
```diff
@@ -80,8 +79,6 @@ void registerAllSpecialFormGeneralFunctions() {
       bool,
       Generic<T1>,
       Variadic<Generic<T1>>>({"in"});
-  InRewrite::registerRewrite();
```
pramodsatya, line 83: @kagamiori, could you please remove the registration of the rewrite and retain `InRewrite.h/.cpp`? Reference: https://github.com/facebookincubator/velox/pull/15549#issuecomment-3555328606 .

pramodsatya (changes requested):
Thanks @kagamiori, could you please remove just the registration of the rewrite until https://github.com/facebookincubator/velox/issues/15648 is resolved? That way it can still be used outside of Velox expression eval, please see discussion in https://github.com/facebookincubator/velox/pull/15549#issuecomment-3555328606.

velox/functions/prestosql/registration/GeneralFunctionsRegistration.cpp
kagamiori: Hi @pramodsatya, sorry that I just saw your message, but this PR is already merged. Could you create another PR to add InRewrite.h/.cpp back? (By the way, we found that IN after rewrite can produce incorrect result, so it's probably safer for your use case too to fix this bug before continue using it.)

pramodsatya: @kagamiori, no worries, I reintroduced the IN-rewrite here: https://github.com/facebookincubator/velox/pull/15663, avoiding the rewrite for expressions of form `NULL IN [a, b, ...]` in order to respect the `kNullAsIndeterminate` null compare flag and fix the result change. Could you please take a look when you get a chance?

PR #15652, opened 2025-11-27
fix: Make RowType::hashKind() thread-safe on AArch64

mpurbay-arm:
fix(type): Make RowType::hashKind() thread-safe on AArch64
    
    The lazy cache added in 6b5a5f155860 (“Optimize RowType::hashKind”) used a
    relaxed load of the “computed” flag and left the cached value uninitialized.
    On weakly ordered CPUs (ARM/AArch64) a reader can observe `hashKindComputed_`
    == true before the write to `hashKind_` is visible, returning a stale/garbage
    value. This manifested as a deterministic failure of:
      TypeTest.rowHashKindMultiThreaded when run with multiple cores.
    
    Even though `hashKind_` and `hashKindComputed_` are stored in a sequentially-consistent
    order, the loading of them in the reader thread are not sequentially-consistent,
    because of the relaxed load. Using acquire–release would fix this,
    but there’s an even cheaper simplification.
    
    Fix: remove the flag and use a single atomic with a zero sentinel:
      - hashKind_ is std::atomic_size_t initialized to 0 (0 = “not computed”)
      - readers do: load(relaxed); if nonzero, return it
      - first thread computes v and store(relaxed) v into hashKind_
    Relaxed accesses are sufficient here because all synchronization is confined to
    the same atomic object (no cross-atomic publish/subscribe needed). Multiple
    threads may compute v concurrently, but they produce the same value and one
    store wins—benign.
    
    Result: removes the race on ARM, keeps sizeof(RowType) smaller, and makes the
    hot path a single relaxed load + branch. No functional API changes.

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -570,11 +570,12 @@ bool RowType::equals(const Type& other) const {
 }
 
 size_t RowType::hashKind() const {
-  if (!hashKindComputed_.load(std::memory_order_relaxed)) {
-    hashKind_ = TypeBase<TypeKind::ROW>::hashKind();
-    hashKindComputed_ = true;
+  auto hashKind = hashKind_.load(std::memory_order_relaxed);
+  if (hashKind == 0) {
+    hashKind = TypeBase<TypeKind::ROW>::hashKind();
+    hashKind_.store(hashKind, std::memory_order_relaxed);
   }
-  return hashKind_;
+  return hashKind;
 }
 
 void RowType::printChildren(std::stringstream& ss, std::string_view delimiter)
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -1218,8 +1218,7 @@ class RowType : public TypeBase<TypeKind::ROW> {
   const std::vector<TypePtr> children_;
   mutable std::atomic<std::vector<TypeParameter>*> parameters_{nullptr};
   mutable std::atomic<NameToIndex*> nameToIndex_{nullptr};
-  mutable std::atomic_bool hashKindComputed_{false};
-  mutable std::atomic_size_t hashKind_;
+  mutable std::atomic_size_t hashKind_{0};
 };
 
 using RowTypePtr = std::shared_ptr<const RowType>;


velox/type/Type.cpp
```diff
@@ -570,11 +570,12 @@ bool RowType::equals(const Type& other) const {
 }
 
 size_t RowType::hashKind() const {
-  if (!hashKindComputed_.load(std::memory_order_relaxed)) {
-    hashKind_ = TypeBase<TypeKind::ROW>::hashKind();
-    hashKindComputed_ = true;
+  auto hash_kind = hashKind_.load(std::memory_order_relaxed);
+  if (hash_kind != 0) {
```
Yuhta: This condition is not correct.  Also use camel case (`hashKind`).

MBkkt: When zero is possible for (kind is always >= 0)?
```
size_t Type::hashKind() const {
  size_t hash = (int32_t)kind() + 1;
  for (auto& child : *this) {
    hash = hash * 31 + child->hashKind();
  }
  return hash;
}
```

mpurbay-arm: Thanks! You’re right—the condition should check for the uninitialized sentinel. hashKind_ uses 0 to mean “not computed yet,” and per Type::hashKind() we always produce a value ≥ 1 ((int32_t)kind() + 1 then accumulates positives), so zero is impossible for a real hash.
I’ve flipped the condition and switched to camelCase:

Yuhta (approved):

kKPulla:
@mpurbay-arm can you please rebase this change to the latest and make sure to fix the CI signals that are failing? I can help merge this shortly after.

mpurbay-arm:
@kKPulla  I’ve rebased the change onto the latest branch. CI is still showing a couple of failing tests, Please let me know if you spot anything I might be missing. Thanks!

kKPulla:
@mpurbay-arm Sorry for the delay here due to holidays, I will pick this back up today and will try to merge this in the next day or two. I will keep you posted if I have any blockers.

mpurbay-arm:
@kKPulla Thanks for looking into this patch. Quick follow-up — I’m still not seeing the code deployed yet. Is there anything you need from me to move it along? If there are any blockers, feel free to ping me.

kKPulla:
> @kKPulla Thanks for looking into this patch. Quick follow-up — I’m still not seeing the code deployed yet. Is there anything you need from me to move it along? If there are any blockers, feel free to ping me.

@mpurbay-arm Sorry for the delay. I had some failures internally that deemed to be unrelated. I rebased earlier today and waiting for all signals to turn green. I will keep you posted.

PR #15656, opened 2025-11-28
fix: Use 'MemoryManager::Options' instead of 'MemoryManagerOption'

lingbin:
'MemoryManagerOption' has been replaced with 'MemoryManager::Options',  
this PR fix the error in the doc.


tanjialiang (approved):
Thanks for the fix

PR #15657, opened 2025-11-29
feat: Enhance floating point value retrieval with type-specific handling in parquet reader

baibaichen:
## Summary
This PR enables the Parquet `FloatingPointColumnReader` to support reading `FLOAT` columns as `DOUBLE`, allowing for schema evolution where the requested type is wider than the file type to avoid result mismatch. It also enforces the bulk path (fast path) for these operations and adds comprehensive test coverage.

## Key Changes

1.  **Type Promotion Support**: Updated `ParquetColumnReader::build` to instantiate `FloatingPointColumnReader<float, double>` when the file type is `REAL` but the requested type is `DOUBLE`.
2.  **Always Enable Bulk Path**: Overridden `hasBulkPath()` in `FloatingPointColumnReader` to always return `true`.
3.  **Safe Bulk Path Selection**: Updated `useFastPath` in `DecoderUtil.h` to disable the fast path when there is a mismatch between the filter type and the data type (e.g., `FloatingPointRange<double>` on `float` data).
4.  **Comprehensive Testing**: Added `floatToDoubleComprehensive_*` tests covering various combinations of nulls, dictionary encoding, filters, and deletions.

### 1. Why Parquet Floating Point Readers Always Support Bulk Path

In this PR, we explicitly set `kHasBulkPath = true` for `FloatingPointColumnReader`, even when `TData` (file type, e.g., `float`) differs from `TRequested` (output type, e.g., `double`).

**Reasoning:**
The Parquet decoder architecture decouples the decoding phase from the type conversion phase. The buffer management in `SelectiveColumnReader` relies on the file type `TData`:

*   **Buffer Allocation**: `prepareRead<TData>()` initializes the `values_` buffer with `valueSize_ = sizeof(TData)`.
*   **Decoding**: The decoding process utilizes the fast path (bulk path) mechanisms if possible:
    *   **Dictionary Encoding**: `RleBpDataDecoder` executes `bulkScan`, which calls `DictionaryColumnVisitor::processRun` or `processRle`. These methods copy values from the dictionary (which holds `TData`) directly into the `values_` buffer (casted to `TData*`).
    *   **Direct Encoding**: Uses `fixedWidthScan` to read `TData` values directly.
    *   **Null Handling**: `dwio::common::nonNullRowsFromDense` is used to identify valid rows for decoding.
    Crucially, all these operations write `sizeof(TData)` bytes per value, matching the buffer allocation.
*   **Bulk Path Validity**: Since the decoder and the buffer both operate in the `TData` domain, the bulk path (fast path) logic—which efficiently decodes batches of data—remains valid and safe.
*   **Deferred Conversion**: Type conversion to `TRequested` is deferred until the `getValues()` call, where `upcastScalarValues<TData, TRequested>()` handles the promotion (e.g., allocating a new 8-byte buffer for doubles and converting the 4-byte floats).

This design allows us to leverage the performance benefits of the bulk path for `float` -> `double` reads without modification to the decoders.

#### Fixed in  `useFastPath`

`FloatingPointRange` explicitly forbids cross-type SIMD evaluation. For instance, calling `FloatingPointRange<float>::testValues` with a `double` batch triggers a `VELOX_FAIL` at runtime.

*   **The Fix**: `useFastPath` now returns `false` if:
    *   The filter is `FloatingPointRange<double>` but the data type is not `double`.
    *   The filter is `FloatingPointRange<float>` but the data type is not `float`.
*   **Impact**: For Float-to-Double promotion, the bulk path is used when there is no filter or the filter is compatible (e.g., `IsNull`, `IsNotNull`). If a `double` range filter is provided, the reader safely falls back to the non-bulk path.

### 2. Evolution of `TRequested` Usage

**Before this PR:**
`TRequested` served three main purposes:
1.  **Bulk Path Determination**: Used to decide if the fast path could be enabled (`std::is_same_v<TData, TRequested>`).
2.  **Result Container**: Defined the type of the output vector.
3.  **Null Handling**: Used as the type for default null values in `filterNulls`.

**After this PR:**
1.  **Bulk Path Determination**: No longer used for this check in Parquet. We explicitly enable bulk path for floating-point conversions because the decoder operates on `TData`.
2.  **Result Container**: Remains the primary type for the output vector.
3.  **Null Handling**: The usage of `filterNulls<TRequested>` remains unchanged.

**Why `filterNulls<TRequested>` was not changed to `filterNulls<TData>`:**
Although `filterNulls<TRequested>` writes `sizeof(TRequested)` bytes (e.g., 8 bytes for double) into a buffer allocated for `TData` (e.g., 4 bytes for float), we decided not to change this because:
1.  **Safety via Padding**: The `values_` buffer allocation includes `simd::kPadding`, which provides sufficient buffer space to prevent memory corruption even when writing larger null values at the end of the buffer.
2.  **Data Integrity**: While writing a larger null value might overwrite adjacent bytes, those bytes correspond to subsequent values. Since the current value is null, its specific bit pattern is irrelevant. If the subsequent value is valid, it will be written correctly afterwards, overwriting the "tail" of the previous null value.
3.  **DWRF Compatibility**: This code path is shared with DWRF, which may rely on the existing behavior.

### 3. Test Coverage

Added a comprehensive suite of tests in `ParquetReaderTest.cpp` to verify the correctness of float-to-double promotion under various scenarios. The test matrix covers:

*   **Basic Scenarios**: Simple float-to-double conversion, handling of nulls, parent struct nulls, and row deletions.
*   **Dictionary Encoding**: Verified both with and without dictionary encoding enabled.
*   **Filters**: Tested `IsNull`, `IsNotNull` and value filters to ensure they work correctly with type promotion.
*   **Comprehensive Matrix**: Refactored individual tests into a parameterized test suite `FloatToDoubleEvolutionTest` using GTest's `TEST_P`. This systematically covers all combinations of:
    *   **Nulls**: Presence vs. absence of null values.
    *   **Dictionary**: Enabled vs. disabled.
    *   **Filters**: No filter, `IsNull`, `IsNotNull`, and `>=`
    *   **Density**: Dense reading (contiguous) vs. Non-dense reading (with deletions/sparse reading).

This ensures that the bulk path logic holds up across different decoding paths (RLE/BP vs. Direct) and filtering strategies, while keeping the test code concise and maintainable.

### DIFF
--- a/velox/type/Filter.h
+++ b/velox/type/Filter.h
@@ -1856,9 +1856,9 @@ template <>
 inline xsimd::batch_bool<float> FloatingPointRange<double>::testValues(
     xsimd::batch<float> x) const {
   // Slow path for schema evolution:  handles the case where double filter is
-  // applied to float data (e.g., reading float column with double type request).
-  // Falls back to element-by-element testing as batch processing requires type
-  // matching.
+  // applied to float data (e.g., reading float column with double type
+  // request). Falls back to element-by-element testing as batch processing
+  // requires type matching.
   return Filter::testValues(x);
 }
 
--- a/velox/type/tests/FilterTest.cpp
+++ b/velox/type/tests/FilterTest.cpp
@@ -804,14 +804,16 @@ TEST(FilterTest, doubleRange) {
   filter = lessThanOrEqualDouble(1.2);
   {
     auto verifyFloat = [&](float x) { return filter->testFloat(x); };
-    float n8[] = {1.0f, std::nanf("nan"), 1.3f, 1e20f, -1e20f, 0.0f, 1.1f, 1.2f};
+    float n8[] = {
+        1.0f, std::nanf("nan"), 1.3f, 1e20f, -1e20f, 0.0f, 1.1f, 1.2f};
     checkSimd(filter.get(), n8, verifyFloat);
   }
 
   filter = greaterThanDouble(1.2);
   {
     auto verifyFloat = [&](float x) { return filter->testFloat(x); };
-    float n8[] = {1.0f, std::nanf("nan"), 1.3f, 1e20f, -1e20f, 0.0f, 1.1f, 1.2f};
+    float n8[] = {
+        1.0f, std::nanf("nan"), 1.3f, 1e20f, -1e20f, 0.0f, 1.1f, 1.2f};
     checkSimd(filter.get(), n8, verifyFloat);
   }
 }


baibaichen:
@majetideepak 

this pr is ready to review.

cc @zhouyuan

velox/dwio/common/SelectiveFloatingPointColumnReader.h
```diff
@@ -45,7 +45,29 @@ class SelectiveFloatingPointColumnReader : public SelectiveColumnReader {
   readCommon(int64_t offset, const RowSet& rows, const uint64_t* incomingNulls);
 
   void getValues(const RowSet& rows, VectorPtr* result) override {
-    getFlatValues<TData, TRequested>(rows, result, requestedType_);
+    switch (requestedType_->kind()) {
```
Yuhta: This method is called per batch, we should not do validation here unless it's not possible else where.  In this case `TData` and `TRequested` is created when we build the column reader, and the validation should be added there instead of this per batch call.

baibaichen: @Yuhta 

**Thank you for the suggestion. However, in this case, the per-batch switch statement is required.**

Theoretically, based on the current builder logic, `TRequested` should equal `TypeTraits<requestedType_->kind()>::NativeType`. However, `TRequested` is the type currently used in the Parquet decoder. If I create a reader instance like `FloatingPointColumnReader<float, double>`, it will trigger a runtime assertion, see  [here](https://github.com/facebookincubator/velox/blob/953c2cc8e9b7852c0c9c59c532d6ec339e9ccede/velox/dwio/common/ColumnVisitors.h#L1175)
```c++
  if constexpr (!kHasBulkPath) {
    // Only DWRF integer dictionary is using this, which should not disable bulk
    // path at decoder level.
    VELOX_UNREACHABLE();
  }
```  

Commenting out this assertion might seem acceptable, but referring to the implementation of  [`SelectiveColumnReader::getIntValues`](https://github.com/facebookincubator/velox/blob/953c2cc8e9b7852c0c9c59c532d6ec339e9ccede/velox/dwio/common/SelectiveColumnReader.cpp#L198), it appears better to ensure that the `TVector` parameter in `getFlatValues` always matches `requestedType_`.

baibaichen: @Yuhta Would introducing `TDecode` for internal decoding and limiting `TRequested` to `getValues` be acceptable?

``` c++
template <typename TData, typename TDecode, typename TRequested>
class FloatingPointColumnReader
    : public SelectiveFloatingPointColumnReader<TData, TDecode> {
 public:
  void getValues(const RowSet& rows, VectorPtr* result) override {
    getFlatValues<TData, TRequested>(rows, result, requestedType_);
  }
};
```

 Personally, I prefer keeping the current approach since it aligns with the existing codebase.

rui-mo:
I added a tracker for the Spark fuzzer issue: https://github.com/facebookincubator/velox/issues/15697, thanks.

velox/dwio/common/SelectiveFloatingPointColumnReader.h
Yuhta: `TRequested` is the output type, decoder should take both `TData` and `TRequested` and make the correct conversion.  I don't think a third type is needed here, just need to fix the decoder to make it work.

baibaichen:
@Yuhta I’ve updated the implementation. Could you please take a look?

velox/dwio/common/SelectiveFloatingPointColumnReader.h
```diff
@@ -35,9 +35,10 @@ class SelectiveFloatingPointColumnReader : public SelectiveColumnReader {
             params,
             scanSpec) {}
 
-  // Offers fast path only if data and result widths match.
+  // Offers a fast path only if data and result widths match.
+  static constexpr bool kHasBulkPath = std::is_same_v<TData, TRequested>;
```
Yuhta, line 39: nit: Leave a new line in-between

velox/dwio/common/DecoderUtil.h
```diff
@@ -419,7 +421,15 @@ bool useFastPath(Visitor& visitor) {
            is_same_v<typename Visitor::FilterType, velox::common::AlwaysTrue> ||
        !hasNulls || !visitor.allowNulls()) &&
       (std::is_same_v<typename Visitor::HookType, NoHook> || !hasNulls ||
-       Visitor::HookType::kSkipNulls);
+       Visitor::HookType::kSkipNulls) &&
+      (!std::is_same_v<
```
Yuhta: This should not be needed, if the filter is accepting `double`, the requested type must be `double` as well, and bulk path should not kick in when data type is `float` because `Visitor::kHasBulkPath = false` in this case

baibaichen: > This should not be needed, if the filter is accepting `double`, the requested type must be `double` as well, and bulk path should not kick in when data type is `float` because `Visitor::kHasBulkPath = false` in this case

This PR makes the Parquet `FloatingPointColumnReader` always set `Visitor::kHasBulkPath` to `true`. Therefore, when a `FloatingPointRange` filter is applied and the data type mismatches the requested type, we need to disable the fast path.

Yuhta: Then this is not the right place to add this condition, which decoder is Parquet `FloatingPointColumnReader` using?  The condition should be added there.  Also check on only `FloatingPointRange` is not enough, it should be disabled whenever there is a filter and type mismatch.

velox/dwio/common/SelectiveFloatingPointColumnReader.h
baibaichen: done

velox/dwio/common/DecoderUtil.h
baibaichen: > Then this is not the right place to add this condition, which decoder is Parquet `FloatingPointColumnReader` using? The condition should be added there. Also check on only `FloatingPointRange` is not enough, it should be disabled whenever there is a filter and type mismatch.

After looking into `MultiRange`, I realized we can support the type mismatch directly in the filter rather than disabling the bulk path. I've reverted the changes in `DecoderUtil.h`.

Specifically, I supported `inline xsimd::batch_bool<float> FloatingPointRange<double>::testValues(xsimd:: batch<float> x)`. This allows `FloatingPointRange<double>` to correctly process float data batches when there's a type mismatch, similar to how `MultiRange` handles different types.

velox/dwio/common/SelectiveFloatingPointColumnReader.h
Yuhta: Oh I meant an empty line between `kHasBulkPath` and `hasBulkPath()`, not `kHasBulkPath` and its comment

velox/type/Filter.h
```diff
@@ -1781,8 +1781,8 @@ inline xsimd::batch_bool<double> FloatingPointRange<double>::testValues(
 
 template <>
 inline xsimd::batch_bool<float> FloatingPointRange<double>::testValues(
-    xsimd::batch<float>) const {
-  VELOX_FAIL("Not defined for double filter");
+    xsimd::batch<float> x) const {
+  return Filter::testValues(x);
```
Yuhta, line 1862: Can you add some tests to `FilterTest` for this?  Also comment this is a slow path only triggered for schema evolution.

velox/dwio/common/DecoderUtil.h
Yuhta: Both `MultiRange` and `FloatingPointRange<double>::testValues(xsimd:: batch<float> x)` are very slow and only acts as fallback though (maybe the latter could have a faster implementation but probably not worth the trouble).

velox/type/Filter.h
baibaichen: Done

baibaichen:
@kgpai @Yuhta just checking in—this PR was imported a week ago. Does it need any further changes from my side? Thanks!

PR #15658, opened 2025-11-29
fix(build): Literal overload for long on macOS

czentgr:
PR #15615 added a test that uses the literal "10L" that on macOS resolves to the long integral type and for which there was no "lit" function overload resulting in ambiguous usage of lit.

### DIFF
--- a/velox/exec/tests/PlanBuilderTest.cpp
+++ b/velox/exec/tests/PlanBuilderTest.cpp
@@ -323,7 +323,7 @@ TEST_F(PlanBuilderTest, filter) {
   VELOX_CHECK_EQ(
       PlanBuilder()
           .tableScan("tmp", data)
-          .filter(col("c0") + 10L > 100L)
+          .filter(col("c0") + 10LL > 100LL)
           .planNode()
           ->toString(true, false),
       expectation);


baibaichen:
@czentgr @pedroerp 

I also encountered the `no formatter for IExpr` error in my local Ubuntu development environment. How did the CI/CD pipeline manage to pass compilation?

velox/parse/IExpr.h
PingLiuPing, line 93: A alternative way would be changing this line to:

```
    VELOX_FAIL("Unable to add alias to expression '{}'", this->toString());
```

czentgr:
@baibaichen You may notice the CI actions did not trigger builds. There was some emergency maintenance and they have not been restored. Meanwhile some PRs got merged as they appeared green and did build ok in the Meta internal build apparently. But outside of it does not build.
The actions should hopefully be restored soon.

velox/exec/tests/PlanBuilderTest.cpp
```diff
@@ -323,7 +323,7 @@ TEST_F(PlanBuilderTest, filter) {
   VELOX_CHECK_EQ(
       PlanBuilder()
           .tableScan("tmp", data)
-          .filter(col("c0") + 10L > 100L)
+          .filter(col("c0") + 10 > 100)
```
czentgr: On macOS this should use 10 or 10LL (100 or 100LL) because the `L` resolves to `long` on macOS which is its own type. We want to avoid special handling for linux vs macOS.

velox/parse/IExpr.h
czentgr: I left the formatter in because it helps in case IExpr is used in the formatter again not not just in this VELOX_FAIL.

czentgr:
The fuzzer job python build uses the main branch which has the fix for the formatter missing.

czentgr:
https://github.com/facebookincubator/velox/pull/15664 now added fmt formatters and supersedes some of the changes here. This unblocked the Linux build. That leaves the macOS build issue with the literal type 10L resolving to long which is special on macOS.

kagamiori:
Hi @czentgr, there is a test failure of PlanBuilderTest.filter in CI. We saw the same error internal on the imported diff. Could you take a look?

czentgr:
Issue was caused because the literal `10` resolves to an integer and the expression then gets wrapped in a cast that is not expected. I changed it to `10LL` to keep the original intent to resolve to a bigint. This also compiles ok on my macOS. It is just the `10L` that is a special literal resolving to long when used here.

czentgr:
Closing in favor of https://github.com/facebookincubator/velox/pull/15670.
Originally, I used an overload for `lit`. The other solution uses an overload for `toExprWrapper`.

PR #15663, opened 2025-12-01
feat: Add rewrite for IN special form

pramodsatya:
Rewrites IN special form when the `value` being searched for is constant. 

- Returns `true` if `value` is in the `IN-list`. 
- Removes constant elements from `IN-list` that do not match with `value`. 

Constant expression comparison for these 2 rules uses null-as-indeterminate null comparison mode in order to maintain consistency with the function implementation of IN predicate. 
Tests are added to ensure the correctness issue from https://github.com/facebookincubator/velox/issues/15648 is resolved.

Examples:

- 5 IN (NULL, a, 5) ==> true
- 5 IN (1, 2, a, b) ==> 5 IN (a, b)

Cherry-picked from https://github.com/facebookincubator/velox/pull/15488.

Depends on https://github.com/facebookincubator/velox/pull/15705, https://github.com/facebookincubator/velox/pull/15726.

### DIFF
--- a/velox/common/base/CompareFlags.h
+++ b/velox/common/base/CompareFlags.h
@@ -84,7 +84,7 @@ struct CompareFlags {
     ///       ex: (null, 1) = (null, 1) is indeterminate.
     ///
     ///       - If all fields compare results are true, then result is true.
-    ///       ex: (1, 1) = (1, 1) is indeterminate.
+    ///       ex: (1, 1) = (1, 1) is true.
     ///
     ///   4. Maps:
     ///     - Keys are compared first, if keys are not equal values are not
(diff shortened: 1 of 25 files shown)


aditi-pandit:
@pramodsatya : Thanks for the code. Code looks good. Please rebase to fix the build issues.

@mbasmanova : PTAL.

mbasmanova:
@pramodsatya 

> addressing concerns from https://github.com/facebookincubator/velox/issues/15648.

Would you expand this to explain what wasn't working before and how it is fixed now? Please, clarify which tests were added to verify the fix.

pramodsatya:
Thanks @mbasmanova, @aditi-pandit. I made changes to the IN rewrite to ensure constant expression comparison follows the same null handling mode, null-as-indeterminate, as the function implementation. This should effectively resolve https://github.com/facebookincubator/velox/issues/15648. I have verified the fix with reproducer attached in the issue, extended it into a test, and updated the PR description accordingly. 

The rewrite now relies on `ConstantTypedExpr` being able to support user-supplied null handling mode, and I have extracted this change into https://github.com/facebookincubator/velox/pull/15705. Could you please take another look?

PR #15667, opened 2025-12-02
feat(python): Add expression fluent API

pedroerp:
Summary:
Adding a fluent Python API to enable users to build expression without
having to necessarily rely on SQL. Following PySpark's syntax and philosophy,
where possible.

Differential Revision: D87718840

### DIFF
--- a/velox/python/expression/PyExpression.h
+++ b/velox/python/expression/PyExpression.h
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
+
+#include "velox/parse/Expressions.h"
+#include "velox/parse/IExpr.h"
+#include "velox/type/Variant.h"
+
+namespace facebook::velox::py {
+
+/// A thin wrapper around Velox untyped expressions.
+class PyExpr {
+ public:
+  explicit PyExpr(const core::ExprPtr expr = nullptr) : expr_(expr) {}
+
+  std::string toString() const {
+    if (!expr_) {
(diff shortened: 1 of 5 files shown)


velox/python/expression/PyExpression.h
```diff
@@ -0,0 +1,63 @@
...
+  explicit PyExpr(const core::ExprPtr expr = nullptr) : expr_(expr) {}
+
+  std::string toString() const {
+    if (!expr_) {
+      return "[nullptr]";
+    }
+    return expr_->toString();
+  }
+
+  core::ExprPtr expr() const {
```
kgpai, line 39: nit: make return type also const ?

velox/python/expression/expression.cpp
```diff
@@ -0,0 +1,192 @@
...
+    return velox::py::PyExpr::createConstant(
+        velox::DOUBLE(), velox::Variant(arg.cast<double>()));
+  }
+  // Strings.
+  else if (py::isinstance<py::str>(arg)) {
+    return velox::py::PyExpr::createConstant(
+        velox::VARCHAR(), velox::Variant(arg.cast<std::string>()));
+  }
+  // TODO: Support container types (arrays/list, maps, and structs).
+  throw py::type_error(
```
kgpai, line 58: FYI,  You can reuse/copy the code here (https://github.com/facebookincubator/velox/blob/main/velox/python/legacy/complex.cpp#L160) to add support for conversion of complex python types to velox vectors. It supports any kind of array right now, but is straightforward to add map/row support.

velox/python/expression/expression.cpp
```diff
@@ -0,0 +1,192 @@
...
+    return velox::py::PyExpr::createConstant(
+        velox::BOOLEAN(), velox::Variant(arg.cast<bool>()));
+  }
+  // None.
+  else if (py::isinstance<py::none>(arg)) {
+    return velox::py::PyExpr::createConstant(
+        velox::UNKNOWN(), velox::Variant::null(velox::TypeKind::UNKNOWN));
+  }
+  // Integers. For now we assume always bigint (64 bits).
+  else if (py::isinstance<py::int_>(arg)) {
```
kgpai, line 43: Does something like 'a' resolve to a int and thus bigint ?

velox/python/expression/expression.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+} // namespace
+
+PYBIND11_MODULE(expression, m) {
+  using velox::py::PyExpr;
+
+  py::class_<PyExpr>(m, "Expr")
+      .def("__str__", &PyExpr::toString)
+      // Binary arithmetic operators.
+      .def(
```
kgpai, line 79: Nice !

kgpai (approved):
Some nits, but looks good.

PR #15676, opened 2025-12-02
fix: Handle NULL values in comparison filter generation

baibaichen:
## Summary
This PR fixes an issue discovered in https://github.com/apache/spark/pull/45202 where Spark pushes down predicates like `b > null` to Velox.

## Problem
The `ExprToSubfieldFilterParser` currently does not check whether constant expressions are NULL. When NULL values are encountered, they are converted to default values, resulting in incorrect filter generation.

## Solution
If the constant value is NULL, return AlwaysFalse filter instead of nullptr.

This change correctly implements SQL three-valued logic: comparisons with NULL always result in NULL, which in filter context means no rows match. Returning AlwaysFalse explicitly represents this semantic, allowing optimizers to recognize and optimize away these conditions.

## Affected Operators
Comparison operators: =, !=, <, <=, >, >=

## Modified Functions
Modified functions in ExprToSubfieldFilterParser:
- makeNotEqualFilter
- makeEqualFilter
- makeGreaterThanFilter
- makeLessThanFilter
- makeLessThanOrEqualFilter
- makeGreaterThanOrEqualFilter

### DIFF
--- a/velox/expression/ExprToSubfieldFilter.cpp
+++ b/velox/expression/ExprToSubfieldFilter.cpp
@@ -136,6 +136,9 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeNotEqualFilter(
   if (!value) {
     return nullptr;
   }
+  if (value->isNullAt(0)) {
+    return std::make_unique<common::AlwaysFalse>();
+  }
 
   auto lessThanFilter = makeLessThanFilter(valueExpr, evaluator);
   if (!lessThanFilter) {
@@ -180,6 +183,9 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeEqualFilter(
   if (!value) {
     return nullptr;
   }
+  if (value->isNullAt(0)) {
+    return std::make_unique<common::AlwaysFalse>();
+  }
   switch (value->typeKind()) {
     case TypeKind::BOOLEAN:
       return boolEqual(singleValue<bool>(value));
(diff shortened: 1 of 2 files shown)


baibaichen:
@zhouyuan Could you suggest someone to review this PR? The failed fuzzer test seems unrelated to my update:

zhouyuan:
@rui-mo would you please help to take a look on this patch?

velox/expression/ExprToSubfieldFilter.cpp
```diff
@@ -136,6 +136,9 @@ std::unique_ptr<common::Filter> ExprToSubfieldFilterParser::makeNotEqualFilter(
   if (!value) {
     return nullptr;
   }
+  if (value->isNullAt(0)) {
+    return std::make_unique<common::AlwaysFalse>();
```
rui-mo, line 140: I understand the fix, and just have one question: some comparison functions do allow NULL, for example, `EqualNullSafe` returns true when comparing NULL with NULL. Should we clarify and document that these comparisons are not null-safe?

velox/expression/ExprToSubfieldFilter.cpp
rui-mo, line 373: BTW. I have a PR (https://github.com/facebookincubator/velox/pull/14082) that addresses a similar issue when the values contain NULL.

rui-mo:
cc: @Yuhta

velox/expression/ExprToSubfieldFilter.cpp
Yuhta: Those functions should be represented with different functions, not the well known eq, neq etc. recognized here

PR #5632, opened 2023-07-13
Add support for aggregations over distinct inputs

mbasmanova:
Add support for queries like SELECT count(distinct x) FROM t.

Inputs are first accumulated into SetAccumulators. When all input has been
received, unique values are aggregated.

PR #15679, opened 2025-12-03
fix: Destroy the previously created accumulator for AggregateWindow to fix memory leak

jiangjiangtian:
One example I found is as follows: if the method [initializeNewGroupsInternal](https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/aggregates/FirstLastAggregate.cpp#L120) in FirstLastAggregateBase is called multiple times and the function is not numeric, the previously created `SingleValueAccumulator` object will be overwritten without calling `destroy` to free its memory. This leads to memory leak and eventually causes OOM. This situation will occur in `Window` operator when there are multiple window frames to process.

So this PR fixes the issue by calling `destroy` on `rawSingleGroupRow_` when processing a new window frame.

### DIFF
--- a/velox/exec/AggregateWindow.cpp
+++ b/velox/exec/AggregateWindow.cpp
@@ -150,7 +150,6 @@ class AggregateWindowFunction : public exec::WindowFunction {
         // This is the start of a new incremental aggregation. So the
         // aggregate_ function object should be initialized.
         auto singleGroup = std::vector<vector_size_t>{0};
-        aggregate_->clear();
         aggregate_->destroy(folly::Range<char**>(&rawSingleGroupRow_, 1));
         aggregate_->initializeNewGroups(&rawSingleGroupRow_, singleGroup);
         aggregateInitialized_ = true;
@@ -337,7 +336,6 @@ class AggregateWindowFunction : public exec::WindowFunction {
       // TODO : Try to re-use previous computations by advancing and retracting
       // the aggregation based on the frame changes with each row. This would
       // require adding new APIs to the Aggregate framework.
-      aggregate_->clear();
       aggregate_->destroy(folly::Range<char**>(&rawSingleGroupRow_, 1));
       aggregate_->initializeNewGroups(&rawSingleGroupRow_, kSingleGroup);
       aggregateInitialized_ = true;
--- a/velox/exec/Aggregate.h
+++ b/velox/exec/Aggregate.h
@@ -302,6 +302,7 @@ class Aggregate {
 
     for (auto* group : groups) {
       group[initializedByte_] &= ~initializedMask_;
+      clearNull(group);
     }
   }
 
--- a/velox/functions/sparksql/window/tests/SparkWindowTest.cpp
+++ b/velox/functions/sparksql/window/tests/SparkWindowTest.cpp
@@ -115,8 +115,8 @@ class SparkAggregateWindowLimitMemoryTest
   static void SetUpTestCase() {
     OperatorTestBase::SetUpTestCase();
     OperatorTestBase::setupMemory(
-        256 << 20, // allocatorCapacity
-        256 << 20, // arbitratorCapacity
+        192 << 20, // allocatorCapacity
+        192 << 20, // arbitratorCapacity
         0, // arbitratorReservedCapacity
         0, // memoryPoolInitCapacity
         0, // memoryPoolReservedCapacity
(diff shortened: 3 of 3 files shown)


jinchengchenghh:
I find this code https://github.com/facebookincubator/velox/pull/5632/files#diff-b9ed2eb619a914241f656abe59881d169602eed734af2df443f6480d611850aaR122-R124
to call destroy directly, maybe your exception is caused by not calling the destroy directly.

Also, maybe fix that in virtual void initializeNewGroups(
      char** groups,
      folly::Range<const vector_size_t*> indices) {
    initializeNewGroupsInternal(groups, indices);

    for (auto index : indices) {
      groups[index][initializedByte_] |= initializedMask_;
    }
  }
This can help eliminate the memory leak issue.
Also, does this problem only occurs when isFixedSize() false?
```
  virtual bool isFixedSize() const {
    return true;
  }
```
The initializeNewGroupsInternal should only be called when the condition is satisfied.

jiangjiangtian:
> I find this code https://github.com/facebookincubator/velox/pull/5632/files#diff-b9ed2eb619a914241f656abe59881d169602eed734af2df443f6480d611850aaR122-R124 to call destroy directly, maybe your exception is caused by not calling the destroy directly.

Yes, now `AggregateWindow` will only create a new accumulator but without destroying it.

> 
> Also, maybe fix that in virtual void initializeNewGroups( char** groups, folly::Range<const vector_size_t*> indices) { initializeNewGroupsInternal(groups, indices);
> 
> ```
> for (auto index : indices) {
>   groups[index][initializedByte_] |= initializedMask_;
> }
> ```
> 
> } This can help eliminate the memory leak issue.

Do you means that if the `group` has been initialized, then this function should not call `initializeNewGroupsInternal` again? But when we process a new window frame, we need to initialize the `group` again to clear the result of the latest window frame. So I call `destroy` in `AggregateWindow`

> Also, does this problem only occurs when isFixedSize() false?

Maybe not. I find that [NonNumericArbitrary](https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp#L163) does not override `isFixedSize` but we have to call `destroy` on it.

> ```
>   virtual bool isFixedSize() const {
>     return true;
>   }
> ```
> 
> The initializeNewGroupsInternal should only be called when the condition is satisfied.

So this may be not correct.
Thanks for your reply.

jinchengchenghh:
I mean destroy the allocator in initializeNewGroups not initializeNewGroupsInternal, this can help all the group initialization issue, but looks like community prefers to directly call the destroy function, so let us align with the current solution, find one and solve one.

jinchengchenghh:
In the code and review, we may be carefully call the initializeNewGroups, decide if it should call destroy first

velox/exec/AggregateWindow.cpp
```diff
@@ -151,6 +151,7 @@ class AggregateWindowFunction : public exec::WindowFunction {
         // aggregate_ function object should be initialized.
         auto singleGroup = std::vector<vector_size_t>{0};
         aggregate_->clear();
+        aggregate_->destroy(folly::Range<char**>(&rawSingleGroupRow_, 1));
```
jinchengchenghh, line 153: As the clear function comment, do we need to also destroy the allocator_ in clear?
```
  // Clears state between reuses, e.g. this is called before reusing
  // the aggregation operator's state after flushing a partial
  // aggregation.
```

jiangjiangtian: I think we can't destroy allocator_ in clear because the allocator_ is shared between all window functions(https://github.com/facebookincubator/velox/blob/main/velox/exec/Window.cpp#L225-L232).
We can clear window functions here (https://github.com/facebookincubator/velox/blob/main/velox/exec/Window.cpp#L714-L719) and then clear allocator_.

jiangjiangtian:
@mbasmanova @xiaoxmeng @Yuhta Could you please help review this PR?

velox/exec/AggregateWindow.cpp
mbasmanova, line 153: Do we still need to call 'clear' if we are calling 'destroy'?

mbasmanova:
CC: @jiangjiangtian

velox/exec/AggregateWindow.cpp
jiangjiangtian: Maybe we need. `clear` clears some states of the aggregate function and when we start processing a new window frame, we need to clear the states.

aditi-pandit: @jiangjiangtian @mbasmanova I too feel its confusing to call both these. Why does clearing an aggregate not imply destroying all the groups ?

jiangjiangtian: Shall we add call to `destroy` in `clear`?

velox/exec/tests/WindowTest.cpp
```diff
@@ -1036,5 +1037,67 @@ DEBUG_ONLY_TEST_F(WindowTest, releaseWindowBuildInTime) {
               "ORDER BY d");
 }
 
+class SparkWindowTest : public WindowTest {
```
jiangjiangtian: The test has some problem, please ignore it now.

velox/exec/AggregateWindow.cpp
jiangjiangtian: @mbasmanova @aditi-pandit how do you think? should we add call to `destroy` in `clear`?

mbasmanova: @jiangjiangtian clear() is clearing the state, but doesn't free memory. Similar to std::vector::clear. destroy() frees up memory (and also clears the state). Similar to std::vector's destructor. I would imagine it is sufficient to call destroy(). Calling clear() before destroy() is redundant. Is this not the case?

jiangjiangtian: @mbasmanova Now destroy won't clear the states of `Aggregate` functions.  So should we call clear() in destroy()?

mbasmanova: > destroy won't clear the states

That's strange. Why would we need to explicitly clear the state if are releasing memory used by these. Perhaps, there are bugs in some implementations of destroy. What happens when you remove 'clear' call before 'destroy'?

jiangjiangtian: > What happens when you remove 'clear' call before 'destroy'?

Sorry, I have not tried it yet. But I find that 'clear' just sets the null count in `Aggregate` to 0 and the null count  is just used to determine whether we need to actually extract bits from 'group' when we check if the 'group' is null or when we want to clear null bits of the 'group'.
After reading some code, I realize that we have to call [clearNull](https://github.com/facebookincubator/velox/blob/main/velox/exec/Aggregate.h#L411) in `destroy`. I think that is what we need to do. Before we destroy the accumulator, we need to check whether it is null. If it is, we need to subtract it from the null count in `Aggregate` to keep it aligned with the number of null groups allocated. Now we don't perform this check, so maybe that's the bug in `destroy`.
Based on the investigation above, I agree with you that the 'clear' call is confusing. So if we call `clearNull` in `destroy`, we can resolve the confusion. What do you think? Thanks! cc @mbasmanova

mbasmanova: @jiangjiangtian Yes, I believe you are correct that destroy needs to clear the null flag.

jiangjiangtian:
@mbasmanova I iterate this PR based on https://github.com/facebookincubator/velox/pull/15680. Please take a look again, thanks!

jiangjiangtian:
> @jiangjiangtian Would you confirm that the test fails without the changes?

@mbasmanova Yes! The test will fail if I revert this PR or https://github.com/facebookincubator/velox/pull/15680.

PR #15680, opened 2025-12-03
fix: Call prepareReuse on argVectors in AggregateWindow to clear string buffers

jiangjiangtian:
Currently, after processing one window, `AggregateWindow` will only call `resize` on `argVectors_` and [resize](https://github.com/facebookincubator/velox/blob/main/velox/vector/FlatVector-inl.h#L449) does not clear string buffers if the new size is not 0. As a result, `AggregateWindow` may consume a large amount of memory over time.

This PR fixes the issue by calling `prepareReuse` on `argVectors_` when resetting a new partition to clear the string buffers.

### DIFF
--- a/velox/exec/AggregateWindow.cpp
+++ b/velox/exec/AggregateWindow.cpp
@@ -257,10 +257,10 @@ class AggregateWindowFunction : public exec::WindowFunction {
   void fillArgVectors(vector_size_t firstRow, vector_size_t lastRow) {
     vector_size_t numFrameRows = lastRow + 1 - firstRow;
     for (int i = 0; i < argIndices_.size(); i++) {
-      // Without the following call to `prepareForReuse`, if the type of `argVectors_[i]`
-      // is VARCHAR, then the string buffers will accumulate across calculations.
-      // As a result, memory consumption will increase over time. So we call
-      // `prepareForReuse` to clear string buffers timely.
+      // Without the following call to `prepareForReuse`, if the type of
+      // `argVectors_[i]` is VARCHAR, then the string buffers will accumulate
+      // across calculations. As a result, memory consumption will increase over
+      // time. So we call `prepareForReuse` to clear string buffers timely.
       argVectors_[i]->prepareForReuse();
       argVectors_[i]->resize(numFrameRows);
       // Only non-constant field argument vectors need to be populated. The
--- a/velox/functions/sparksql/window/tests/SparkWindowTest.cpp
+++ b/velox/functions/sparksql/window/tests/SparkWindowTest.cpp
@@ -109,7 +109,8 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
     SparkWindowTest,
     testing::ValuesIn(getSparkWindowTestParams()));
 
-class SparkAggregateWindowLimitMemoryTest : public exec::test::OperatorTestBase {
+class SparkAggregateWindowLimitMemoryTest
+    : public exec::test::OperatorTestBase {
  public:
   static void SetUpTestCase() {
     OperatorTestBase::SetUpTestCase();
@@ -136,9 +137,11 @@ class SparkAggregateWindowLimitMemoryTest : public exec::test::OperatorTestBase 
 // the limit and cause failure.
 // The capacity is calculated as:
 //   1. size of input data: 1 rows * 32KB (the length of the string) = 32KB
-//   2. size of RowContainer: ((1,024 rows * 32KB (data) + 1,024 rows * 32KB (Accumulators))) * 3 = 192MB
+//   2. size of RowContainer: ((1,024 rows * 32KB (data) + 1,024 rows * 32KB
+//      (Accumulators))) * 3 = 192MB
 //      (Accumulators won't be destroyed now)
-//   3. size of results: 10 rows * 32KB * 2 (column 'd' and the result column) = 640KB
+//   3. size of results: 10 rows * 32KB * 2 (column 'd' and the result column) =
+//      640KB
 //   4. other overheads
 //    Total: ~ 192MB
 // If we don't clear the string buffers in time, the size of the string buffers
@@ -149,18 +152,16 @@ TEST_F(SparkAggregateWindowLimitMemoryTest, clearStringBuffersInTime) {
   constexpr vector_size_t size = 1'024 * 3;
   constexpr vector_size_t resultSize = 10;
   // For this test, it is important to create a single-row partition.
-  // When we process a new partition, we will call 'fillArgVectors' in 'AggregateWindow'.
-  // Single-row partition can let us call 'fillArgVectors' as many as possible.
-  // If we don't call 'prepareForReuse' in 'fillArgVectors', the string buffers
-  // will accumulate and cause high memory usage.
+  // When we process a new partition, we will call 'fillArgVectors' in
+  // 'AggregateWindow'. Single-row partition can let us call 'fillArgVectors' as
+  // many as possible. If we don't call 'prepareForReuse' in 'fillArgVectors',
+  // the string buffers will accumulate and cause high memory usage.
   auto data = makeRowVector(
       {"d", "p"},
-      {
-          // Payload Data.
-          makeConstant(std::string(32 * 1024, 'a'), size),
-          // Partition key.
-          makeFlatVector<int16_t>(size, [](auto row) { return row; })
-      });
+      {// Payload Data.
+       makeConstant(std::string(32 * 1024, 'a'), size),
+       // Partition key.
+       makeFlatVector<int16_t>(size, [](auto row) { return row; })});
 
   createDuckDbTable({data});
 
@@ -169,27 +170,31 @@ TEST_F(SparkAggregateWindowLimitMemoryTest, clearStringBuffersInTime) {
                   .streamingWindow({"first(d) over (partition by p)"})
                   // We use LIMIT clause here to limit the result size to limit
                   // the memory consumed by results.
-                  // Besides, we set offset to 'size - resultSize' to make the Window operator
-                  // process all window partitions. In this way, if we don't clear
-                  // the string buffers during the window function processing,
-                  // the string buffers would accumulate to cause high memory usage.
+                  // Besides, we set offset to 'size - resultSize' to make the
+                  // Window operator process all window partitions. In this way,
+                  // if we don't clear the string buffers during the window
+                  // function processing, the string buffers would accumulate to
+                  // cause high memory usage.
                   // TODO: remove the use of LIMIT clause
-                  // after fixing https://github.com/facebookincubator/velox/issues/16210
+                  // after fixing
+                  // https://github.com/facebookincubator/velox/issues/16210
                   .limit(size - resultSize, resultSize, true)
                   .planNode();
 
-  const auto realResultSize = AssertQueryBuilder(plan)
-      .serialExecution(true)
-      // Because we set offset to 'size - resultSize' in the LIMIT clause,
-      // the Window operator will produce intermediate results that will be
-      // discarded by LIMIT clause.
-      // Set preferred output batch rows to 10 to limit the memory
-      // consumed by these intermediate results.
-      // By limiting the memory consumed by input and intermediate results,
-      // we can know that the reason why we can pass this test is that we clear
-      // the string buffers in time during the window function processing.
-      .config(core::QueryConfig::kPreferredOutputBatchRows, resultSize)
-      .countResults();
+  const auto realResultSize =
+      AssertQueryBuilder(plan)
+          .serialExecution(true)
+          // Because we set offset to 'size - resultSize' in the LIMIT clause,
+          // the Window operator will produce intermediate results that will be
+          // discarded by LIMIT clause.
+          // Set preferred output batch rows to 10 to limit the memory
+          // consumed by these intermediate results.
+          // By limiting the memory consumed by input and intermediate results,
+          // we can know that the reason why we can pass this test is that we
+          // clear the string buffers in time during the window function
+          // processing.
+          .config(core::QueryConfig::kPreferredOutputBatchRows, resultSize)
+          .countResults();
   ASSERT_EQ(realResultSize, resultSize);
 }
 


jiangjiangtian:
@mbasmanova Could you please help review this PR? Thanks!

velox/exec/AggregateWindow.cpp
```diff
@@ -116,6 +116,9 @@ class AggregateWindowFunction : public exec::WindowFunction {
     partition_ = partition;
 
     previousFrameMetadata_.reset();
+    for (auto& argVector : argVectors_) {
```
mbasmanova: perhaps, add a comment to explain why is this needed.

aditi-pandit: Curious why this is done when resetting the vectors for the partition vs for on a block by block basis in fillArgVectors ?

jiangjiangtian: Maybe it is more intuitive to call it in fillArgVectors, thanks!

jiangjiangtian: @mbasmanova @aditi-pandit I have added comment and unit test. Would you please take a look again?

velox/exec/AggregateWindow.cpp
```diff
@@ -257,6 +257,13 @@ class AggregateWindowFunction : public exec::WindowFunction {
   void fillArgVectors(vector_size_t firstRow, vector_size_t lastRow) {
     vector_size_t numFrameRows = lastRow + 1 - firstRow;
     for (int i = 0; i < argIndices_.size(); i++) {
+      // Without the following call to `prepareForReuse`, if the type of `argVectors_[i]`
+      // is VARCHAR, then the string buffers will accumulate across calculations.
+      // As a result, memory consumption will increase over time. So we call
+      // `prepareForReuse` to clear string buffers timely.
+      argVectors_[i]->prepareForReuse();
+      velox::common::testutil::TestValue::adjust(
```
mbasmanova: Avoid using TestValue unless absolutely necessary. This is quite intrusive and doesn't really test the behavior of the code, rather verifies that a specific code path is taken. This is hard to maintain with refactoring and other code evolutions.

I imagine we should be able to run a simple query that uses VARCHAR, set memory limit low and make sure things work correctly. We should be able to also verify that if we remove the fix, the test fails with an OOM.

jiangjiangtian: Thanks! I will try.

jiangjiangtian: @mbasmanova I rewrite the unit test, please take a look again, thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,54 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+    OperatorTestBase::setupMemory(
+        184 << 20, // allocatorCapacity
+        184 << 20, // arbitratorCapacity
+        0, // arbitratorReservedCapacity
+        184 << 20, // memoryPoolInitCapacity
+        0, // memoryPoolReservedCapacity
+        0, // memoryPoolMinReclaimBytes
+        0 // memoryPoolAbortCapacityLimit
+    );
+  }
```
mbasmanova, line 126: Would you add a comment to explain the choice of these values?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,54 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+
+  void SetUp() override {
+    WindowTestBase::SetUp();
+    WindowTestBase::options_.parseIntegerAsBigint = false;
```
mbasmanova: Is this needed?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,54 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+
+  createDuckDbTable({data});
+
+  auto plan = PlanBuilder()
+                  .values(split(data, 10))
+                  .streamingWindow({"first(d) over (partition by p0 order by s)"})
+                  .planNode();
+
+  AssertQueryBuilder(plan, duckDbQueryRunner_)
+      .config(core::QueryConfig::kPreferredOutputBatchBytes, "1024")
```
mbasmanova: Is this needed? Would you add a comment to explain what this test is designed to verify and how it achieves that?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,54 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+  }
+
+  void SetUp() override {
+    WindowTestBase::SetUp();
+    WindowTestBase::options_.parseIntegerAsBigint = false;
+    velox::functions::aggregate::sparksql::registerAggregateFunctions("");
+  }
+};
+
+TEST_F(SparkAggregateWindowTest, clearStringBuffersInTime) {
```
mbasmanova: What happens when this test runs without a fix?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,54 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+  const vector_size_t size = 1'024;
+  auto data = makeRowVector(
+      {"d", "p0", "s"},
```
mbasmanova: consistency: p0 vs. s

Let's rename p0 to p

jiangjiangtian: Thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: The test will fail.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Removed. Thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Removed. Thanks!

jiangjiangtian:
> @jiangjiangtian Thank you for iterating. Would you clarify how you designed the test to verify the bug fix?

@mbasmanova Thanks for your review! I add some comments about how I get the memory capacity. Please take a look.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,67 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+          // Partition key.
+          makeFlatVector<int16_t>(size, [](auto row) { return row; }),
+          // Sorting key.
+          makeFlatVector<int32_t>(size, [](auto row) { return row; }),
+      });
+
+  createDuckDbTable({data});
+
+  auto plan = PlanBuilder()
+                  .values(split(data, 1))
```
mbasmanova: > split(data, 1)

Do we need this call? Why would we want to split a vector into 1 vector? Are we just trying to create an std::vector with a single element? Let's do .values({data}) instead.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,67 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+                  .streamingWindow({"first(d) over (partition by p order by s)"})
```
mbasmanova: > order by s

Do we need this clause?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,67 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+                  .planNode();
+
+  AssertQueryBuilder(plan, duckDbQueryRunner_)
+      .serialExecution(true)
+      .assertResults(
+          "SELECT *, first(d) over (partition by p order by s) "
```
mbasmanova: Would this be the same as "SELECT *, d" FROM tmp"?

I assume it is important for this test to create single-row partitions. If so, would you comment about why.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,67 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+
+// This test limits the memory capacity to check that string buffers are cleared
+// in time during the aggregation window function processing.
+// Without timely clearing the string buffers, the memory usage would exceed
+// the limit and cause failure.
+// The capacity is calculated as:
+//   1. size of input data: 1,024 rows * 32KB (the length of every string) = 32MB
+//   2. size of RowContainer: 1,024 rows * 32KB = 32MB
+//   3. size of results: 1,024 rows * 32KB * 2 (column 'd' and the result column) * 1.5 (preferred size) = 96MB
+//   4. size of SingleValueAccumulator: 1,024 rows * 32KB = 32MB (We won't destroy it now)
```
mbasmanova: Isn't accumulator part of RowContainer? If that's the case, it would be more logical to combine #4 with #2, no?

My understanding is that RowContainer stores at least 64 KB for each row of input. 32KB for column 'd' and another 32KB for first(d) accumulator. With 1024 rows, this adds up to 64MB.

With single-row window partitions, we call fillArgVector for each row. Without the fix, the argVector would accumulate all strings using another 32MB. With the fix, the argVector would use something like 32KB (minimum size of a string buffer).

The input data is 32MB. The result vector is another 64MB. This adds up to something like 160MB (32 input + 64 result + 64 row container). But you are saying with padding and other overheads this is more like 224MB. That's a very large difference, which makes it difficult to make sure that 32MB savings from argVectors make a difference and can be reliably observed. Am I missing something?

I wonder if the test would be more reliable if we reduced the amount of memory used by input and result vectors. One way to do that could be to (1) change input vector for 'd' to constant; (2) change result vector size to 10 rows. This way, the memory used by input and result will be very small (KBs, not MBs). The total amount of memory would be dominated by RowContainer (64MB) and argVectors(32MB before the fix). We can triple the number of input rows, which would give us 192MB for RowContainer and 96MB for argVectors. Assuming that 192MB RowContainer + other small things fits in 256MB of total memory we should expect the test to pass with the fix and reliably fail without (192 + 96 = 288 > 256). What do you think?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,67 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
     SparkWindowTest,
     testing::ValuesIn(getSparkWindowTestParams()));
 
+class SparkAggregateWindowLimitMemoryTest : public WindowTestBase {
```
mbasmanova: Do we need WindowTestBase as a base class? What do we use from it? Can we use OperatorTestBase instead?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: For this case, yes. Because every partition only has one row.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: I think we need this clause here because we use streamingWindow here. It assumes that the input is sorted, so we use this clause here to indicate that the input is sorted by 's'.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Yes, I think you are right. Thanks.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: 1. Yes, accumulator is part of RowContainer. I will combine it.
2. When we allocate a string buffer to store a string of 32KB, we actually will allocate a string buffer of 48KB size because we will round the size up [here](https://github.com/facebookincubator/velox/blob/main/velox/buffer/Buffer.h#L396). So I say the size of result is 96MB in the comment. And I'm sorry that I made a mistake. The memory capacity may be 192MB. Thanks for your pointing!
3. For now, when I set the capacity to 200MB, I can't pass the test, either. I did some investigations and found that the memory pool that allocates memory for input reserved 64MB. The memory consumed by input affect us. So I think that reducing the amount of memory used by input and result vectors is a really good idea. I will try it. Thanks!

mbasmanova: @xiaoxmeng @tanjialiang  Meng, any chance you can help us understand the memory usage here?

tanjialiang: Hi @jiangjiangtian, I'm trying to understand the question: are you asking why memory pool that allocates memory for input reserved 64MB instead of 32MB?

xiaoxmeng: > Hi @jiangjiangtian, I'm trying to understand the question: are you asking why memory pool that allocates memory for input reserved 64MB instead of 32MB?

Yeah, is it the velox memory pool reserve more than expected? There is roundup in memory reservation.

jiangjiangtian: @tanjialiang @xiaoxmeng Thanks for your explanation.

jiangjiangtian: @mbasmanova I rewrite the unit test according to your idea. Could you please take a look? Thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
mbasmanova: Can you double check? I believe streaming window only requires that input is clustered on partitioned key (this is the case because each row is its own partition) and sorted within partition (this is also the case because each partition is just one row).

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,80 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+          // Sorting key.
+          makeFlatVector<int32_t>(size, [](auto row) { return row; }),
+      });
+
+  createDuckDbTable({data});
+
+  auto plan = PlanBuilder()
+                  .values({data})
+                  .streamingWindow({"first(d) over (partition by p order by s)"})
+                  .limit(size - resultSize, resultSize, true)
```
mbasmanova, line 181: What's the purpose of this LIMIT clause?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,80 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+  }
+};
+
+// This test limits the memory capacity to check that string buffers are cleared
+// in time during the aggregation window function processing.
+// Without timely clearing the string buffers, the memory usage would exceed
+// the limit and cause failure.
+// The capacity is calculated as:
+//   1. size of input data: 1 rows * 32KB (the length of the string) = 32KB
+//   2. size of RowContainer: ((1,024 rows * 32KB (data) + 1,024 rows * 32KB (Accumulators))) * 3 = 192MB
```
mbasmanova: > * 3

Would you clarify this? Why do we need 3 times more memory than there is data?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Limit the size of result. Because if we don't limit the result size, the result will consume a lot of memory.
Besides, I set the offset of LIMIT clause to `size - resultSize` to let the window operator process every window partition.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: I set the size of input to 1024 * 3.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
mbasmanova: I see. We can use countResults instead of assertResults to avoid accumulating results in memory. This would allow to remove the 'limit' and make the test a bit easier to reason about.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
mbasmanova: I still think we do not need 'order by' clauses in this test.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: 'countResults' will get all results and then count them.
I think I need add more comments on why we use LIMIT clause.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Sorry, I didn't see your reply. I will check it, thanks!

jiangjiangtian: Yes, streaming window only requires that input is clustered on partitioned key and sorted within partition. So it makes sense that we remove the 'order by' clause here, thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
mbasmanova: > 'countResults' will get all results and then count them.

Oh... I see. Any chance you could create an issue about this? It would be nice to fix.

> I think I need add more comments on why we use LIMIT clause.

Ok, let's do this and add a TODO with a link to GitHub issue to use countResults once it is fixed.

jiangjiangtian: Sure, thank you for advice.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,92 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+  // will accumulate and cause high memory usage.
+  auto data = makeRowVector(
+      {"d", "p", "s"},
+      {
+          // Payload Data.
+          makeConstant(std::string(32 * 1024, 'a'), size),
+          // Partition key.
+          makeFlatVector<int16_t>(size, [](auto row) { return row; }),
+          // Sorting key.
+          makeFlatVector<int32_t>(size, [](auto row) { return row; }),
```
mbasmanova: I assume we no longer need this column('s'), do we?

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,92 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+      // consumed by these intermediate results.
+      // By limiting the memory consumed by input and intermediate results,
+      // we can know that the reason why we can pass this test is that we clear
+      // the string buffers in time during the window function processing.
+      .config(core::QueryConfig::kPreferredOutputBatchRows, 10)
+      // TODO: remove the use of LIMIT clause and use `countResults`
+      // after fixing https://github.com/facebookincubator/velox/issues/16210
+      .assertResults(
+          fmt::format("SELECT *, first(d) over (partition by p) "
+          "FROM tmp ORDER BY s LIMIT 10 OFFSET {}", size - resultSize));
```
mbasmanova: We can drop 'ORDER BY s' here

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Yes, thanks.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
jiangjiangtian: Without 'ORDER BY', the order of results returned by duckDb is arbitrary. Because we limit the size of results to 10, the assert will fail.

mbasmanova: @jiangjiangtian True, but that's Ok. assertResults doesn't require the order of results to match.

jiangjiangtian: The content of results is also different because we use LIMIT clause here. Maybe I should use `countResults`.

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,89 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+      // Because we set offset to 'size - resultSize' in the LIMIT clause,
+      // the Window operator will produce intermediate results that will be
+      // discarded by LIMIT clause.
+      // Set preferred output batch rows to 10 to limit the memory
+      // consumed by these intermediate results.
+      // By limiting the memory consumed by input and intermediate results,
+      // we can know that the reason why we can pass this test is that we clear
+      // the string buffers in time during the window function processing.
+      .config(core::QueryConfig::kPreferredOutputBatchRows, 10)
+      // TODO: remove the use of LIMIT clause
```
mbasmanova: Let's move this TODO to the place that uses the limit: L176

Using countResults here is much cleaner. You may also simplify AssertQueryBuilder(plan, duckDbQueryRunner_) to AssertQueryBuilder(plan) since we no longer use DuckDB.

jiangjiangtian: Thanks!

velox/functions/sparksql/window/tests/SparkWindowTest.cpp
```diff
@@ -106,5 +109,89 @@ VELOX_INSTANTIATE_TEST_SUITE_P(
...
+      .serialExecution(true)
+      // Because we set offset to 'size - resultSize' in the LIMIT clause,
+      // the Window operator will produce intermediate results that will be
+      // discarded by LIMIT clause.
+      // Set preferred output batch rows to 10 to limit the memory
+      // consumed by these intermediate results.
+      // By limiting the memory consumed by input and intermediate results,
+      // we can know that the reason why we can pass this test is that we clear
+      // the string buffers in time during the window function processing.
+      .config(core::QueryConfig::kPreferredOutputBatchRows, 10)
```
mbasmanova: 10 -> resultSize

jiangjiangtian: Thanks.

kagamiori:
Hi @jiangjiangtian, there is a failure of the format check in https://github.com/facebookincubator/velox/actions/runs/21624362345/job/62385671302?pr=15680, as well as a few other failures in this PR. Could you take a look? Thanks!

jiangjiangtian:
> Hi @jiangjiangtian, there is a failure of the format check in https://github.com/facebookincubator/velox/actions/runs/21624362345/job/62385671302?pr=15680, as well as a few other failures in this PR. Could you take a look? Thanks!

I have fixed the format check.
The other failure seems to be not related to this PR.

kagamiori:
Hi @jiangjiangtian, could you also rebase the PR onto the latest main, because there is a merge conflict? Thanks!

jiangjiangtian:
> Hi @jiangjiangtian, could you also rebase the PR onto the latest main, because there is a merge conflict? Thanks!

@kagamiori I have rebased, thanks!

kagamiori:
Hi @jiangjiangtian, it seems there is a conflict again. Could you rebase again onto the latest main? Sorry for the delay and inconvenience.

jiangjiangtian:
> Hi @jiangjiangtian, it seems there is a conflict again. Could you rebase again onto the latest main? Sorry for the delay and inconvenience.

@kagamiori rebased.

kagamiori:
Not able to import this PR currently due to an incident of Github services. I'll retry once the github incident is resolved/mitigated.

kagamiori:
Hi @jiangjiangtian, sorry, could you rebase again to trigger the CI tests on this PR? It seems like no test ran due to some problem on the github side.

jiangjiangtian:
> Hi @jiangjiangtian, sorry, could you rebase again to trigger the CI tests on this PR? It seems like no test ran due to some problem on the github side.

Done.

PR #15681, opened 2025-12-03
fix: Propagate nulls from parent to child in exportRows

acvictor:
We observed production failures with ARITHMETIC_ERROR when exporting Velox RowVectors containing NULL parent rows to Arrow format during Delta Lake OPTIMIZE operations. The error occurred in toMicros() when processing TIMESTAMP fields:
```
org.apache.spark.sql.delta.schema.InvariantViolationException: Exception: VeloxUserError 
Error Source: USER 
Error Code: INVALID_ARGUMENT 
Reason: Could not convert Timestamp(124597086605856, 124597086605856) to microseconds, Exception: VeloxUserError 
Error Source: USER 
Error Code: ARITHMETIC_ERROR 
Reason: integer overflow: 124597086605856 * 1000000 
Retriable: False 
DisplayReason: False 
Function: checkedMultiply 
File: /__w/1/s/Velox/./velox/common/base/CheckedArithmetic.h 
Line: 51 
```
When a parent RowVector has NULL at position i, the child vector's data at position i contains uninitialized garbage. However, the child's null bitmap incorrectly indicates NOT NULL. The Arrow export code, based on the bitmap, attempted to read the garbage data, causing integer overflow in timestamp conversions.

If a null bitmap indicates NOT NULL, the data buffer MUST contain valid data. This PR propagates parent nulls through the entire Arrow export call chain in Bridge.cpp without modifying the source Velox vectors.


acvictor:
@majetideepak raised a draft here (for timestamp buffers only and can generalize if you think the approach makes sense.

acvictor:
@majetideepak can you please review this? Thanks!

kgpai:
Please convert out of draft for reviews. Also please update to latest base and fix failing CI Jobs.

acvictor:
> Please convert out of draft for reviews. Also please update to latest base and fix failing CI Jobs.

Thanks @kgpai, can you please take a look?

acvictor:
@kgpai @majetideepak can I please get your thoughts on this?

kgpai:
Apologies for delay @acvictor . Looking

PR #15684, opened 2025-12-03
fix: Expand operator with constants specified using ConstantVector

jinchengchenghh:
The constantExpr->value() only exists when the constant is of scalar type, consider the value vector for complex data type.
```
-- Expand[1][[1, null, null, n0_1, n0_2, n0_3, n0_4], [4, n0_0, {-1, {{-1}, {k => v}}}, null, null, null, null]] -> n1_0:INTEGER, n1_1:INTEGER, n1_2:ROW<c1:INTEGER,c2:ROW<a:ARRAY<INTEGER>,m:MAP<VARCHAR,VARCHAR>>>, n1_3:VARCHAR, n1_4:BIGINT, n1_5:INTEGER, n1_6:ROW<>
  -- ValueStream[0][] -> n0_0:INTEGER, n0_1:VARCHAR, n0_2:BIGINT, n0_3:INTEGER, n0_4:ROW<>
```
Returns null for complex constant before this fix.
```
VeloxColumnarToRowConverter 0: {4, 1, null, null, null, ...2 more}
```

Create the constant vector in Expand operator constructor to reuse it.

### DIFF
--- a/velox/exec/Expand.cpp
+++ b/velox/exec/Expand.cpp
@@ -64,9 +64,9 @@ void Expand::initialize() {
     return;
   }
   const auto numColumns = constantProjections_[0].size();
-  std::vector<VectorPtr> constantOutput;
-  constantOutput.reserve(numColumns);
   for (const auto& projections : constantProjections_) {
+    std::vector<VectorPtr> constantOutput;
+    constantOutput.reserve(numColumns);
     for (const auto& constant : projections) {
       if (constant) {
         constantOutput.push_back(constant->toConstantVector(pool()));
@@ -74,8 +74,8 @@ void Expand::initialize() {
         constantOutput.push_back(nullptr);
       }
     }
+    constantOutputs_.emplace_back(std::move(constantOutput));
   }
-  constantOutputs_.emplace_back(std::move(constantOutput));
 }
 
 bool Expand::needsInput() const {


velox/exec/Expand.cpp
```diff
@@ -86,16 +86,7 @@ RowVectorPtr Expand::getOutput() {
 
   for (auto i = 0; i < numColumns; ++i) {
     if (rowProjection[i] == kConstantChannel) {
-      const auto& constantExpr = constantProjection[i];
-      if (constantExpr->value().isNull()) {
```
jinchengchenghh, line 90: This check is not correct

JkSelf (approved):
LGTM. Thanks for your fix.

jinchengchenghh:
Hi, @mbasmanova , could you help review this PR? Thanks

velox/exec/tests/ExpandTest.cpp
```diff
@@ -39,6 +39,20 @@ class ExpandTest : public OperatorTestBase {
 };
 } // anonymous namespace
 
+TEST_F(ExpandTest, complexConstant) {
```
mbasmanova, line 40: move all tests into anonymous namespace

```
namespace {
class ExpandTest 
...

TEST_F(ExpandTest,...

TEST_F(ExpandTest,...

}
```

velox/exec/tests/ExpandTest.cpp
```diff
@@ -39,6 +39,20 @@ class ExpandTest : public OperatorTestBase {
...
+  auto data = makeRowVectorData(1);
+
+  createDuckDbTable({data});
+
+  auto arrayVector = makeArrayVector<int32_t>({{1, 2, 3}});
+  auto constantExpr = std::make_shared<core::ConstantTypedExpr>(arrayVector);
+
+  auto plan =
+      PlanBuilder().values({data}).expand({{constantExpr}}, {"c1"}).planNode();
```
mbasmanova: Can we do something like

```
.expand({
   {"k1", "k2", "a", "b", "[1, 2, 3] as c"},
}) 
```

velox/exec/Expand.cpp
```diff
@@ -86,16 +86,7 @@ RowVectorPtr Expand::getOutput() {
...
-        // Add null column.
-        outputColumns[i] = BaseVector::createNullConstant(
-            outputType_->childAt(i), numInput, pool());
-      } else {
-        // Add constant column.
-        outputColumns[i] = BaseVector::createConstant(
-            constantExpr->type(), constantExpr->value(), numInput, pool());
-      }
+      outputColumns[i] = constantProjection[i]->toConstantVector(pool());
```
mbasmanova: This fix looks good. I wonder if we can avoid making new vector for each output batch and instead create these once and reuse.

velox/exec/tests/ExpandTest.cpp
jinchengchenghh: I receive the exception after updated to this
```
unknown file: Failure
C++ exception with description "Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Scalar function doesn't exist: list_value.
Retriable: False
Function: resolveScalarFunctionType
File: /Users/chengchengjin/code/velox/velox/parse/TypeResolver.cpp
Line: 97
```

```
TEST_F(ExpandTest, complexConstant) {
  auto data = makeRowVectorData(1);

  auto arrayVector = makeArrayVector<int32_t>({{1, 2, 3}});
//   auto constantExpr = std::make_shared<core::ConstantTypedExpr>(arrayVector);

  auto plan = PlanBuilder()
                  .values({data})
                  .expand({
                      {"k1", "k2", "a", "b", "[1, 2, 3] as c"},
                  })
                  .planNode();
  auto children = data->children();
  children.push_back(arrayVector);
  auto expected = makeRowVector(children);
  assertQuery(plan, expected);
}
```

jinchengchenghh: Can not use array_constructor also, the list_value is parsed as array_constructor in here, so even if we resolve the parse issue, we still cannot use the projection to test expand node.
https://github.com/facebookincubator/velox/blob/a886644e93fb1aabacf28eb117884d3e98c0ce7d/velox/parse/QueryPlanner.cpp#L87
```
Error Code: INVALID_ARGUMENT
Reason: Unsupported projection expression in Expand plan node. Expected field reference or constant. Got: array_constructor(1,2,3) 
Retriable: False
Expression: TypedExprs::isFieldAccess(columnProjection) || TypedExprs::isConstant(columnProjection)
Function: ExpandNode
File: /Users/chengchengjin/code/velox/velox/core/PlanNode.cpp
```

jinchengchenghh: This can work well, I will update to this, thanks for your suggestion!
```
TEST_F(ExpandTest, complexConstant) {
  auto data = makeRowVectorData(1);
  auto children = data->children();
  auto arrayVector = makeArrayVector<int32_t>({{1, 2, 3}});
  auto complexConstants = makeRowVector({arrayVector});
  children.push_back(arrayVector);
  auto expected = makeRowVector({"k1", "k2", "a", "b", "c0"}, children);

//   auto constantExpr = std::make_shared<core::ConstantTypedExpr>(arrayVector);

  auto plan = PlanBuilder()
                  .values({expected})
                  .expand({
                      {"k1", "k2", "a", "b", "__complex_constant(c0) as c"}
                  }, complexConstants)
                  .planNode();

  assertQuery(plan, expected);
}
```

mbasmanova: I think I missed 'ARRAY' keyword in the original example. Something like this should work:

```
PlanBuilder(pool())
...
.expand({
   {"k1", "k2", "a", "b", "ARRAY[1, 2, 3] as c"},
}) 
```

jinchengchenghh: Thanks! Updated and works well.

velox/exec/tests/ExpandTest.cpp
```diff
@@ -37,7 +36,24 @@ class ExpandTest : public OperatorTestBase {
...
   }
 };
-} // anonymous namespace
+
+TEST_F(ExpandTest, complexConstant) {
+  auto otherColumns = makeRowVectorData(1);
+  auto children = otherColumns->children();
+  auto arrayVector = makeArrayVector<int32_t>({{1, 2, 3}});
+  children.push_back(arrayVector);
+  std::vector<VectorPtr> complexConstants(children.size());
```
mbasmanova: complexConstants not needed anymore

velox/exec/tests/ExpandTest.cpp
```diff
@@ -37,7 +36,24 @@ class ExpandTest : public OperatorTestBase {
...
+  complexConstants.back() = makeRowVector({arrayVector});
+  auto expected = makeRowVector({"k1", "k2", "a", "b", "c0"}, children);
+
+  auto plan = PlanBuilder(pool())
+                  .values({expected})
+                  .expand(
+                      {{"k1", "k2", "a", "b", "ARRAY[1, 2, 3] as c"}})
+                  .planNode();
+
+  assertQuery(plan, expected);
```
mbasmanova, line 62: It is a bit strange that 'expected' is used as both the input and output of the plan; is this intentional?

velox/exec/Expand.cpp
```diff
@@ -81,21 +80,12 @@ RowVectorPtr Expand::getOutput() {
   std::vector<VectorPtr> outputColumns(outputType_->size());
 
   const auto& rowProjection = fieldProjections_[rowIndex_];
-  const auto& constantProjection = constantProjections_[rowIndex_];
+  const auto& constantProjection = constantOutputs_[rowIndex_];
```
mbasmanova, line 104: constantOutputs_ contains vectors of size 1, but here we need to produce a vector of size numInput

I just realized that we cannot reuse the vector and may need to call BaseVector::wrapInConstant(numInput, 0, constantOutputs_[rowIndex_]).

velox/exec/tests/ExpandTest.cpp
```diff
@@ -37,7 +36,24 @@ class ExpandTest : public OperatorTestBase {
         });
   }
 };
-} // anonymous namespace
+
+TEST_F(ExpandTest, complexConstant) {
+  auto otherColumns = makeRowVectorData(1);
```
mbasmanova: Let's use size > 1. Otherwise, we won't be able to catch a bug mentioned above.

velox/exec/tests/ExpandTest.cpp
```diff
@@ -37,7 +36,22 @@ class ExpandTest : public OperatorTestBase {
...
+  auto data = makeRowVectorData(3);
+  auto children = data->children();
+  auto arrayVector =
+      makeArrayVector<int32_t>({{1, 2, 3}, {1, 2, 3}, {1, 2, 3}});
+  children.push_back(arrayVector);
+  auto expected = makeRowVector({"k1", "k2", "a", "b", "c"}, children);
+
+  auto plan = PlanBuilder(pool())
+                  .values({data})
+                  .expand({{"k1", "k2", "a", "b", "ARRAY[1, 2, 3] as c"}})
```
mbasmanova: Should we also test null value? 

> null::integer[]

jinchengchenghh:
After add more tests, I receive the exception which requires not allocate memory in Operator constructor, so I move the memory allocation to initialize CC @mbasmanova 
```
E20251222 16:23:01.206250 24729043 Exceptions.h:53] Line: /Users/chengchengjin/code/velox/velox/exec/Task.cpp:1054, Function:createAndStartDrivers, Expression:  Unexpected memory pool allocations during task[test_cursor_1] driver initialization: 
Top 1 leaf memory pool usages:
    op.1.0.0.Expand usage 768B reserved 1.00MB peak 768B

query.TaskCursorQuery_0.0 usage 768B reserved 1.00MB peak 1.00MB
    task.test_cursor_1 usage 768B reserved 1.00MB peak 1.00MB
        node.1 usage 768B reserved 1.00MB peak 1.00MB
            op.1.0.0.Expand usage 768B reserved 1.00MB peak 768B

, Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Unexpected memory pool allocations during task[test_cursor_1] driver initialization: 
```

kKPulla:
@jinchengchenghh I just imported the PR to merge and while I wait for all approvals and signals to pass, can you please take a look at the test failures from the CI signals above and fix if relevant?

jinchengchenghh:
Could you help import it? I have resolved the failed CI. Thanks! @kKPulla

PR #15688, opened 2025-12-03
feat: Add reservoir_sample aggregate function

gggrace14:
Reservoir sample function allows to select a fixed-size random sample from a dataset, ensuring each record has an equal chance of being chosen.

### DIFF
--- a/velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
@@ -1,0 +1,891 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/prestosql/aggregates/ReservoirSampleAggregate.h"
+
+#include <folly/Random.h>
+
+#include "velox/common/base/Exceptions.h"
+#include "velox/exec/Aggregate.h"
+#include "velox/functions/prestosql/aggregates/AggregateNames.h"
+#include "velox/vector/ComplexVector.h"
+#include "velox/vector/DecodedVector.h"
+#include "velox/vector/FlatVector.h"
+
+namespace facebook::velox::aggregate::prestosql {
+
+namespace {
+
+struct ReservoirSampleAccumulator {
+  VectorPtr samples = nullptr;
+  std::optional<int32_t> maxSampleSize = std::nullopt;
+  std::optional<int64_t> processedCount = std::nullopt;
+  VectorPtr initialSamples = nullptr;
+  // -1 indicates no initialSamples. Use -1 for simplicity for writing and
+  // reading intermediate results. Allowed values of initialSeenCount are -1, 0,
+  // and positive integers.
+  int64_t initialSeenCount = -1;
+
+  void initialize(
+      int32_t desiredSampleSize,
+      const TypePtr& elementType,
+      memory::MemoryPool* pool);
+
+  bool isInitialized() const {
+    if (samples == nullptr) {
+      return false;
+    }
+    VELOX_DCHECK(maxSampleSize.has_value());
+    VELOX_DCHECK(processedCount.has_value());
+    return true;
+  }
+
+  vector_size_t initialSampleCount() const {
+    return initialSamples ? initialSamples->size() : 0;
+  }
+
+  vector_size_t sampleCount() const {
+    return std::min<vector_size_t>(
+        processedCount.value_or(0), maxSampleSize.value_or(0));
+  }
+
+  /// Adds a single value to the reservoir sample using reservoir sampling
+  /// algorithm.
+  /// @param vector The vector containing the value to add.
+  /// @param row The row within the vector to add.
+  /// @param isNull Whether the value at the given row is null.
+  void addValueToReservoir(
+      const BaseVector* vector,
+      vector_size_t row,
+      bool isNull = false);
+
+  /// Adds multiple values from a vector to the reservoir sample using
+  /// reservoir sampling algorithm.
+  /// @param vector The vector containing the values to add.
+  /// @param offset Starting offset within the vector.
+  /// @param size Number of values to add from the vector.
+  void addValuesToReservoir(
+      const BaseVector* vector,
+      vector_size_t offset,
+      vector_size_t size);
+
+  /// Merges reservoir samples from an ArrayVector row into this accumulator.
+  /// @param otherSampleArrayVector ArrayVector containing the samples to merge.
+  /// @param otherRow Row within otherSampleArrayVector containing the samples
+  /// to merge.
+  /// @param otherProcessedCount Total number of processed input elements.
+  /// @param otherMaxSampleSize: Max sample size to maintain in the reservoir.
+  void mergeSamples(
+      const ArrayVector& otherSampleArrayVector,
+      vector_size_t otherRow,
+      int64_t otherProcessedCount,
+      int32_t otherMaxSampleSize,
+      memory::MemoryPool* pool);
+
+  /// Merges reservoir samples from another vector into this accumulator.
+  /// @param otherSampleVector Vector containing the samples to merge.
+  /// @param otherOffset Starting offset within otherSampleVector.
+  /// @param otherSize Number of samples to merge from otherSampleVector.
+  /// @param otherProcessedCount Total number of processed input elements.
+  /// @param otherMaxSampleSize: Max sample size to maintain in the reservoir.
+  void mergeSamples(
+      const VectorPtr& otherSampleVector,
+      vector_size_t otherOffset,
+      vector_size_t otherSize,
+      int64_t otherProcessedCount,
+      int32_t otherMaxSampleSize,
+      memory::MemoryPool* pool);
+
+  /// Sets the samples for this accumulator from a source vector.
+  /// @param sampleVector Source vector containing the samples to copy.
+  /// @param offset Starting offset within sampleVector to copy from.
+  /// @param size Number of samples to copy from sampleVector.
+  /// @param seenCount Total number of elements that have been processed.
+  /// @param desiredSampleSize Max sample size to maintain in the reservoir.
+  void setSamples(
+      const VectorPtr& sampleVector,
+      vector_size_t offset,
+      vector_size_t size,
+      int64_t seenCount,
+      int32_t desiredSampleSize,
+      memory::MemoryPool* pool);
+
+  /// Sets the initial samples for this accumulator from a source vector.
+  /// This is used to initialize the reservoir with a pre-existing sample set
+  /// and its associated processed count from a previous sampling operation.
+  /// @param initialSampleVector Source vector containing the initial samples to
+  /// copy.
+  /// @param offset Starting offset within initialSampleVector to copy from.
+  /// @param size Number of initial samples to copy from initialSampleVector.
+  /// @param initialProcessedCount Total number of elements that were processed
+  ///        to generate the initial samples.
+  void setInitialSamples(
+      const VectorPtr initialSampleVector,
+      vector_size_t offset,
+      vector_size_t size,
+      int64_t initialProcessedCount,
+      memory::MemoryPool* pool);
+};
+
+void ReservoirSampleAccumulator::initialize(
+    int32_t desiredSampleSize,
+    const TypePtr& elementType,
+    memory::MemoryPool* pool) {
+  VELOX_USER_CHECK_GT(desiredSampleSize, 0, "sample size must be positive");
+  maxSampleSize = desiredSampleSize;
+  samples = BaseVector::create(elementType, desiredSampleSize, pool);
+  VELOX_DCHECK(samples->isFlatEncoding());
+  processedCount = 0;
+
+  initialSamples = nullptr;
+  initialSeenCount = -1;
+}
+
+void ReservoirSampleAccumulator::setSamples(
+    const VectorPtr& sampleVector,
+    vector_size_t offset,
+    vector_size_t size,
+    int64_t seenCount,
+    int32_t desiredSampleSize,
+    memory::MemoryPool* pool) {
+  VELOX_DCHECK_NOT_NULL(sampleVector);
+  VELOX_DCHECK_GE(seenCount, 0);
+  VELOX_DCHECK(size == seenCount || size == desiredSampleSize);
+
+  if (samples == nullptr) {
+    samples = BaseVector::create(sampleVector->type(), desiredSampleSize, pool);
+  }
+  samples->copy(sampleVector.get(), 0, offset, size);
+  processedCount = seenCount;
+  maxSampleSize = desiredSampleSize;
+}
+
+void ReservoirSampleAccumulator::setInitialSamples(
+    const VectorPtr initialSampleVector,
+    vector_size_t offset,
+    vector_size_t size,
+    int64_t initialProcessedCount,
+    memory::MemoryPool* pool) {
+  if (initialProcessedCount <= 0) {
+    VELOX_USER_CHECK(
+        initialSampleVector == nullptr || size == 0,
+        "Initial state array must be null or empty "
+        "when initial processed count is <= 0.");
+  }
+
+  if (initialSeenCount >= 0) {
+    return;
+  }
+
+  if (initialSampleVector && size > 0) {
+    VELOX_USER_CHECK_GE(
+        initialProcessedCount,
+        size,
+        "InitialProcessedCount must be greater than or equal to the number "
+        "of positions in the initial sample.");
+  }
+
+  // When initialSamples == nullptr, initialSeenCount can be any of
+  // -1, 0, or positive.
+  initialSeenCount = initialProcessedCount;
+
+  if (initialProcessedCount <= 0 || initialSampleVector == nullptr) {
+    initialSamples = nullptr;
+    return;
+  }
+  initialSamples = BaseVector::create(initialSampleVector->type(), size, pool);
+  initialSamples->copy(initialSampleVector.get(), 0, offset, size);
+}
+
+void ReservoirSampleAccumulator::addValueToReservoir(
+    const BaseVector* vector,
+    vector_size_t row,
+    bool isNull) {
+  VELOX_CHECK(isInitialized());
+
+  processedCount = processedCount.value_or(0) + 1;
+
+  vector_size_t targetIndex = -1;
+  if (processedCount.value() <= maxSampleSize) {
+    targetIndex = processedCount.value() - 1;
+  } else {
+    const uint64_t random = folly::Random::rand64(processedCount.value());
+    if (random < samples->size()) {
+      targetIndex = random;
+    }
+  }
+
+  if (targetIndex >= 0) {
+    if (isNull) {
+      samples->setNull(targetIndex, true);
+    } else {
+      samples->copy(vector, targetIndex, row, 1);
+    }
+  }
+}
+
+void ReservoirSampleAccumulator::addValuesToReservoir(
+    const BaseVector* vector,
+    vector_size_t offset,
+    vector_size_t size) {
+  VELOX_CHECK(isInitialized());
+
+  SelectivityVector rows(maxSampleSize.value(), false);
+  std::vector<vector_size_t> toSource(maxSampleSize.value());
+
+  for (vector_size_t i = 0; i < size; i++) {
+    if (processedCount.value() < maxSampleSize) {
+      toSource[processedCount.value()] = offset + i;
+      rows.setValid(processedCount.value(), true);
+
+      processedCount = processedCount.value_or(0) + 1;
+    } else {
+      processedCount = processedCount.value_or(0) + 1;
+
+      const uint64_t random = folly::Random::rand64(processedCount.value());
+      if (random < samples->size()) {
+        toSource[random] = offset + i;
+        rows.setValid(random, true);
+      }
+    }
+  }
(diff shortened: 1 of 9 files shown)


velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    VELOX_DCHECK(processedCount.has_value());
+    return true;
+  }
+
+  vector_size_t initialSampleCount() const {
+    return initialSamples ? initialSamples->size() : 0;
+  }
+
+  vector_size_t sampleCount() const {
+    return processedCount.value_or(0) <= maxSampleSize.value_or(0)
```
kgpai: nit: Would something like this be wrong: return `min(processedCount.value_or(0), maxSampleSize.value_or(0))` ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+        ? processedCount.value_or(0)
+        : samples->size();
+  }
+
+  void addValueToReservoir(
+      const BaseVector* vector,
+      vector_size_t row,
+      bool isNull = false);
+
+  void mergeSamples(
```
kgpai, line 91: nit: can you add some comments for some of these functions..

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+  processedCount = 0;
+
+  initialSamples = nullptr;
+  initialSeenCount = -1;
+}
+
+void ReservoirSampleAccumulator::setSamples(
+    const VectorPtr& sampleVector,
+    vector_size_t offset,
+    vector_size_t size,
```
kgpai, line 160: nit: can you add some comment on offset/size - are the offset size into the sampleVector or samples vector ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    int64_t seenCount,
+    int32_t desiredSampleSize,
+    const HashStringAllocator* allocator) {
+  VELOX_CHECK_NOT_NULL(sampleVector);
+  VELOX_CHECK_GE(seenCount, 0);
+  VELOX_DCHECK(size == seenCount || size == desiredSampleSize);
+
+  // const VectorPtr& sampleElements = sampleArrayVector.elements();
+  samples = BaseVector::create(
```
kgpai: We already create the vector with desired sample size in initialize() why call create again ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+  maxSampleSize = desiredSampleSize;
+}
+
+void ReservoirSampleAccumulator::setInitialSamples(
+    const VectorPtr initialSampleVector,
+    vector_size_t offset,
+    vector_size_t size,
+    int64_t initialProcessedCount,
+    const HashStringAllocator* allocator) {
+  if (initialProcessedCount <= 0) {
```
kgpai, line 182: nit: You can add this if condition as part of the velox_user_check .

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    if (random < samples->size()) {
+      targetIndex = random;
+    }
+  }
+
+  if (targetIndex >= 0) {
+    if (isNull) {
+      samples->setNull(targetIndex, true);
+    } else {
+      samples->copy(vector, targetIndex, row, 1);
```
kgpai, line 235: I dont think we are supporting complex types, so in this case calling copy for mutating one value is rather expensive, rather just keep ref to rawbuffer and mutate the targetIndex directly.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+};
+
+void ReservoirSampleAccumulator::initialize(
+    int32_t desiredSampleSize,
+    const TypePtr& elementType,
+    const HashStringAllocator* allocator) {
+  VELOX_USER_CHECK_GT(desiredSampleSize, 0, "sample size must be positive");
+  maxSampleSize = desiredSampleSize;
+  samples =
+      BaseVector::create(elementType, desiredSampleSize, allocator->pool());
```
kgpai: nit: why pass along HSA , if all you are doing is getting its memory pool, why not just pass the memory pool ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+  VELOX_CHECK_EQ(
+      maxSampleSize.value(),
+      otherMaxSampleSize,
+      "maximum number of samples {} must be equal to that of other {}",
+      maxSampleSize.value(),
+      otherMaxSampleSize);
+
+  if (otherProcessedCount < maxSampleSize.value()) {
+    for (vector_size_t i = 0; i < size; ++i) {
+      addValueToReservoir(sampleVector.get(), offset + i);
```
kgpai: nit: maybe this can be done via bulk copy instead of calling 1 at a time.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    }
+    return;
+  }
+
+  if (processedCount.value() < maxSampleSize.value()) {
```
kgpai, line 331: nit: It shouldnt matter if otherProcessedCount < maxSampleSize doesnt it , all that matters is if processedCount < maxSamplesize and we copy over whats in the other accumulator till it hits max sample size then do sampling..

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    const VectorPtr tempSamples = std::move(samples);
+    const int64_t tempProcessedCount = processedCount.value();
+
+    setSamples(
```
kgpai, line 336: So we get all the samples from the other accumulator and then add the our current samples back ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+      sampleArrayVector.elements(),
+      sampleArrayVector.offsetAt(row),
+      sampleArrayVector.sizeAt(row),
+      otherProcessedCount,
+      otherMaxSampleSize,
+      allocator);
+}
+
+void ReservoirSampleAccumulator::mergeSamples(
+    const VectorPtr& sampleVector,
```
kgpai: nit: Should this be called otherSampleVector ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+  }
+
+  folly::ThreadLocalPRNG rng;
+  const VectorPtr oneShuffledVector =
+      shuffleByDictionary(samples, 0, sampleCount(), rng, allocator);
+  const VectorPtr otherShuffledVector =
+      shuffleByDictionary(sampleVector, offset, size, rng, allocator);
+  const int64_t oneProcessedCount = processedCount.value();
+
+  samples =
```
kgpai: Sorry but why are we creating another samples vector here again ? Why not just reuse the current samples vector.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+    vector_size_t size,
+    folly::ThreadLocalPRNG& randGen,
+    const HashStringAllocator* allocator) {
+  BufferPtr shuffledIndices = allocateIndices(size, allocator->pool());
+  vector_size_t* rawIndices = shuffledIndices->asMutable<vector_size_t>();
+
+  std::iota(rawIndices, rawIndices + size, offset);
+  std::shuffle(rawIndices, rawIndices + size, randGen);
+
+  return BaseVector::wrapInDictionary(
```
kgpai: I think the cost of calling BaseVector::copy because you maintain dictionaries will kill any benefit of dictionaries . You can just as well keep the vector flat and shuffle the raw buffer.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,819 @@
...
+                accumulator->initialSampleCount(),
+        "When a positive initial_processed_count is provided, "
+        "the size of the initial sample must be equal to desired_sample_size parameter.");
+
+    ReservoirSampleAccumulator finalAccumulator;
+    const VectorPtr nonNullInitialSample = BaseVector::getOrCreateEmpty(
+        accumulator->initialSamples,
+        accumulator->samples->type(),
+        allocator_->pool());
+    finalAccumulator.setSamples(
```
kgpai, line 519: Sorry I dont understand why not merge initial samples during any of the partial/intermediate or even final aggregation steps, why wait till you need to extract values ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Hi @kgpai  Krishna, I think one statement under the hood is, when running the reservoir sampling and doing the replacement, it matters whether a vector is one with elements to be replaced (one vector) or is one replacing (other vector). Suppose one vector has maxSampleSize or m elements, and I think the mergeSamples() intentionally makes sure that one vector always has m elements. According to the addValueToReservoir() logic after one merge, the probability for an element of one vector staying in the sample is 1 - 1/p, where p is processedCount. And the probability for an element of other vector being in the sample is m/p. If p>m, 1-1/p is higher.

I replicated the exact logic of Java version here and elsewhere. I suggest we respect java logic, since I think there's math justification behind it. I know the general algorithm of reservoir sampling and justification. When it comes to the distributed version where it does sampling separately and then merge, Java might refer to some algorithm.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: One thing is if we reuse the current `samples` which is actually behind `oneShuffledVector`, we will overwrite the content. And the overwrite pos moves faster than the read pos.

Another thing is I read from the doc of `extractValues()`, "This method should not have side effects, i.e., calling this method doesn't change the content of the accumulators."

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Exactly. Replicating Java's logic. I guess the intention is to make the vector of size maxSampleSize the one being replaced, while make the vector of size < maxSampleSize the replacing vector.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: I will wrap it with `if (samples == nullptr)`.

One caller of this `setSamples()` is mergeSamples() (line 264) where I std::move(samples) to avoid memory copy. So need to create here again.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: I will add an additional implementation to only update targetIndex and do bulk copy.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Revised

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Revised per offline discussion

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Done

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+  // When initialSamples == nullptr, initialSeenCount can be any of
+  // -1, 0, or positive.
+  initialSeenCount = initialProcessedCount;
+
+  if (initialProcessedCount <= 0 || initialSampleVector == nullptr) {
+    initialSamples = nullptr;
+    return;
+  }
+  initialSamples = BaseVector::create(initialSampleVector->type(), size, pool);
```
kgpai, line 209: Curious - can we just keep a reference to initialSampleVector rather that create a copy ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+          folly::Random::rand64(processedCount.value(), rng);
+      if (random < samples->size()) {
+        toSource[random] = offset + i;
+        rows.setValid(random, true);
+      }
+    }
+  }
+  rows.updateBounds();
+
+  samples->copy(vector, rows, toSource.data());
```
kgpai, line 267: Cant we just copy inside the for loop itself, by getting and mutating the rawbuffer ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+  SelectivityVector rows(maxSampleSize.value(), false);
+  std::vector<vector_size_t> toSource(maxSampleSize.value());
+
+  for (vector_size_t i = 0; i < size; i++) {
+    if (processedCount.value() < maxSampleSize) {
+      toSource[processedCount.value()] = offset + i;
+      rows.setValid(processedCount.value(), true);
+
+      processedCount = processedCount.value_or(0) + 1;
```
kgpai, line 254: nit: processedCount ++ is good enough, since in the if condition you already accessed .value() so it should have a value (if it doesnt it will crash there).

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+  if (!isInitialized()) {
+    initialize(otherMaxSampleSize, elementType, pool);
+  }
+
+  VELOX_DCHECK(
+      samples->type()->equivalent(*elementType),
+      "Samples to be merged are of different types.");
+  VELOX_CHECK_EQ(
+      maxSampleSize.value(),
```
kgpai, line 320: if max sample sizes are supposed to be same across accumulators why pass and check if they are same, isnt it going to be same constant used across the whole aggregate ?

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
kgpai: I am not sure whether there is any real benefit for this logic. IMV this is non determinsitic anyway and all we need to ensure is that all elements have the same probability of selection.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+  SelectivityVector otherRows(maxSampleSize.value(), false);
+  vector_size_t oneNext = 0;
+  vector_size_t otherNext = 0;
+
+  for (vector_size_t i = 0; i < maxSampleSize.value(); ++i) {
+    if (folly::Random::rand64(
+            processedCount.value() + otherProcessedCount, rng) <
+        processedCount.value()) {
+      oneRows.setValid(i, true);
+      toSourceRow.push_back(oneIndex[oneNext++]);
```
kgpai: May I ask why we need toSourceRow, you have already shuffled it, why keep another vector for mapping why not just determine count of elements to get from other vector say n, and copy first n elements ? The probably will be same since you are shuffling everything at first anyway.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+    } else {
+      otherRows.setValid(i, true);
+      toSourceRow.push_back(otherIndex[otherNext++]);
+    }
+  }
+  oneRows.updateBounds();
+  otherRows.updateBounds();
+
+  VectorPtr mergedSamples =
```
kgpai: I do believe we can avoid creating a new vector everytime we merge, since this will make many calls to the pool.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+
+  auto* initialSeenCountVector =
+      rowVector->childAt(1)->asChecked<FlatVector<int64_t>>();
+  initialSeenCountVector->resetNulls();
+
+  auto* seenCountVector =
+      rowVector->childAt(2)->asChecked<FlatVector<int64_t>>();
+  seenCountVector->resetNulls();
+
+  auto* maxSampleSizeVector =
```
kgpai, line 571: Not sure why we need this ,as its almost always the same and a constant across accumulators.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
```diff
@@ -0,0 +1,910 @@
...
+      initialSampleOffset += initialSampleSize;
+    }
+
+    initialSeenCountVector->set(i, accumulator->initialSeenCount);
+  }
+}
+
+template <bool kSingleGroup>
+void ReservoirSampleAggregate::addRaw(
+    std::conditional_t<kSingleGroup, char*, char**> group,
```
kgpai, line 628: Neat did not know this trick.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: If we keep a reference to the input vector, it could be a reference to an input *batch* that's already gone? The input batch could be gone when we're processing the next batch or after the addRawInputs() returns and before we enter the extractAccumulators()?

Also I find the Java behavior yesterday that each group could have a different initialSample array which is the first initialSample array that the group sees. That's too complex if we use a reference.

I can test the reference approach in the next version, but still feel it'slikely to be dangling reference.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Revised

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: One thing is the `samples` also contains nulls in addition to actual values. So I think we're not mutating a single raw buffer but two. Then it'll be the similar to mutating a Vector by the Vector::copy() func, given the logic of whether the nulls buffer has been preallocated or not.

Another aspect is, I thought you prefer bulk copy, from previous comment. If you prefer moving the samples->copy to the for-loop, that's the signature in line 233. I can change it back. Just let me know which version you prefer. I feel they're similar perf-wise, with bulk copy slightly better.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Krishna, I went back and checked after we discussed, I found that the Java behavior is to allow different maxSampleSize for different groups. So different accumulator can have different maxSampleSize values. Then we're not able to keep a constant as a data member of the ReservoirSampleAggregate class.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: We need to make sure all elements are selected with the same probability *globally*. However, we're running in a distributed manner and *locally* we proved that it matters whether a vector is being replaced (left) or replacing (right). The logic always makes the vector of maxSampleSize being replaced. It's likely that the algorithm enforces this logic *locally* to make sure the equal probability *globally*.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: We cannot copy *in place* on the `samlpes` vector, with shuffled index, `oneIndex`, for it. The behavior is undefined, since oneIndex has overlapping source indices and target indices. That is,
```
samples.copy(samples.get(), oneRows, oneIndex);
```
behavior is undefined. oneIndex has overlapping source indices and target indices.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Revised as suggested. I get rid of toSourceRow and reuse the shuffled oneIndex and otherIndex. Pre-compute the count of elements from each vector.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
gggrace14: Krishna, I went back and checked after we discussed, I found that the Java behavior is to allow different maxSampleSize for different groups. So different accumulator can have different maxSampleSize values. Then we're not able to keep a constant as a data member of the ReservoirSampleAggregate class.

velox/functions/prestosql/aggregates/ReservoirSampleAggregate.cpp
kgpai: If you keep a reference it wont 'go' away. It should be safe to keep the reference, its only a problem for things like stringview which do not have a shared ptr and that can cause problems.

PR #15695, opened 2025-12-04
feat: Support variant-to-vector for DECIMAL types

xiaodouchen:
Previously, variant-to-vector conversion did not support DECIMAL types.
I observed that DECIMAL can be handled the same way as other FixedWidth
types. Therefore, the template specialization for HUGEINT can be
removed, allowing the generic implementation to handle it.

### DIFF
--- a/velox/vector/VariantToVector.h
+++ b/velox/vector/VariantToVector.h
@@ -21,8 +21,7 @@ namespace facebook::velox {
 
 // Converts Variant `value` into a Velox vector using specified type.
 //
-// Supports all primitive types and complex types that do not contain DECIMAL
-// types.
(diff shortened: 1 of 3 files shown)


velox/vector/VariantToVector.h
```diff
@@ -21,8 +21,7 @@ namespace facebook::velox {
 
 // Converts Variant `value` into a Velox vector using specified type.
 //
-// Supports all primitive types and complex types that do not contain DECIMAL
-// types.
+// Supports all primitive types and complex types.
```
mbasmanova, line 24: Does this mean that ALL types are supported? If so, we can remove the comment. Otherwise, let's clarify what's still not supported.

PR #15702, opened 2025-12-04
docs: Add st_equals case for empty geometries

jkhaliqi:
ST_Equals for empty geometries should return true even if they are different geometry types.
closes: https://github.com/prestodb/presto/issues/26253

### DIFF
--- a/velox/docs/functions/presto/geospatial.rst
+++ b/velox/docs/functions/presto/geospatial.rst
@@ -135,7 +135,9 @@ function you are using.
 
 .. function:: ST_Equals(geometry1: Geometry, geometry2: Geometry) -> boolean
 
-    Returns ``true`` if the given geometries represent the same geometry.
+    Returns ``true`` if the given geometries represent the same geometry
+    according to ISO SQL/MM semantics. Also returns ``true`` if both geometries are empty,
+    regardless of their geometry types.
 
 .. function:: ST_Intersects(geometry1: Geometry, geometry2: Geometry) -> boolean
 


czentgr:
@jagill Can you please take a look at the issue and PR when you get a chance? Looking at the standard. Maybe Presto is wrong?

jagill:
Thanks for this investigation and apologies for taking so long to respond.

There is a mismatch between [DE-9IM model](https://en.wikipedia.org/wiki/DE-9IM) and the [ISO spec](https://www.iso.org/standard/60343.html). The [OGC standard](https://www.ogc.org/standards/sfs/) says it follows ISO semantics for ST_Equals.

The DE-9IM model states: "Two geometries are topologically equal if their interiors intersect and no part of the interior or boundary of one geometry intersects the exterior of the other" (intersection code `T*F**FFF*`). By this definition, two empty geometries (regardless of type) **are not equal**, since they do not have interiors and thus the interiors don't intersect.

However, the ISO spec says that `ST_Equals(g1, g2)` is equivalent to `ST_IsEmpty(ST_SymDifference(g1, g2))`. This evaluates to true for two empty geometries (regardless of type). By this definition, two empty geometries **are equal**.

I think either of these two are possible ways forward. Since we try to follow the ISO spec as closely as possible, and also it's more intuitive that two empty geometries are equal, I'd say that the current Velox behavior is correct (although we should document it), and we should change the current Presto behavior (and document it).

I'm uncomfortable with empty geometries being equal if and only if they are the same type, because the type of the empty geometry is not specified in any operator or spec, and different implementations vary widely in what they return. Thus non-semantic changes of the empty geometry type would result in different evaluations of ST_Equals, which is fragile.

jkhaliqi:
@jagill Thanks for the review! I have updated the velox docs to match the current implementation in this PR and changed presto to also give true for empty geometries regardless of the type here: https://github.com/prestodb/presto/pull/27015

czentgr:
@jkhaliqi please rebase. Let's see if the CI completes now.

czentgr (approved):

jainxrohit:
@jkhaliqi It seems its failing with build issues:
Error: Failed to FinalizeArtifact: Failed to make request after 5 attempts: Unexpected token '<', "<!DOCTYPE "... is not valid JSON

jkhaliqi:
Thank you! CI should be passing now @czentgr @jainxrohit

PR #15704, opened 2025-12-05
misc: Back out batch serialization with preserve encoding option support in remote functions (#15703)

avinaashgupta:
Summary:

Original commit changeset: 9976c4af1fc5

Original Phabricator Diff: D87873415

We are seeing crashes in both LogarithmLeafService and LogarithmFrontend P2067237549

Differential Revision: D88425172


pedroerp (approved):
Thank out. 

FYI: this exposes a crash somewhere else in the serialization code. Reverting it for now to ensure that main in clean until we have a proper fix.

PR #15705, opened 2025-12-05
feat: Enable constant expression comparison with custom null handling mode

pramodsatya:
Enables comparison of constant expressions with user-specified null handling mode. 
Allows constant expression comparison to use null-as-indeterminate null comparison semantics.

Required by https://github.com/facebookincubator/velox/pull/15663.

Depends on https://github.com/facebookincubator/velox/pull/15726.


velox/core/Expressions.cpp
```diff
@@ -252,11 +261,16 @@ bool equalsNoNulls<TypeKind::MAP>(
 
   size_t i = 0;
   for (const auto& [key, value] : mapValue) {
-    if (!equalsImpl(mapVector->mapKeys(), sortedIndices[i], key)) {
+    if (!equalsImpl(
+            mapVector->mapKeys(), sortedIndices[i], key, nullHandlingMode)) {
       return false;
```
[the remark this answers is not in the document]
pramodsatya: `equalsNoNulls` is only called when neither of the values being compared is null.

velox/core/Expressions.h
```diff
@@ -139,6 +139,25 @@ class ConstantTypedExpr : public ITypedExpr {
...
+  /// @param pool Memory pool used to convert constant expressions with variant
+  ///     to a constant vector for comparison with nullHandlingMode.
+  /// TODO: Add support for comparing Variants with kNullAsIndeterminate null
+  ///  comparison mode.
+  /// @return true if expressions are equal, false if not equal, std::nullopt
+  ///     if the result is indeterminate (only with kNullAsIndeterminate mode).
+  std::optional<bool> equals(
+      const ITypedExpr& other,
+      CompareFlags::NullHandlingMode nullHandlingMode,
+      memory::MemoryPool* pool) const;
```
mbasmanova: I don't think we want 'pool' argument for 'equals' APIs. There should be no memory allocations here.

mbasmanova: Instead of adding new API, let's just add an optional nullHandlingMode argument to existing 'equals' API.

pramodsatya: Thanks, I agree it is not suitable to add the `pool` argument here. Currently only vectors support comparison with custom `NullHandlingMode`. Since the constant in the expression can be represented either by a variant or a constant vector, it is required to convert variants to a constant vector for comparing with the right `NullHandlingMode`. The pool is needed to convert the variant to constant vector and solve the constant expression comparison problem.
Could you please share how to resolve this issue of comparing variant-vector and variant-variant from two constant expressions with the provided `NullHandlingMode`? 
cc: @aditi-pandit @czentgr

mbasmanova: @pramodsatya I think you need to add nullHandlingMode parameter to equalsImpl and introduce Variant::equals API that takes nullHandlingMode.

mbasmanova: CC: @bikramSingh91 @kagamiori

pramodsatya: Thanks @mbasmanova, working on adding Variant::equals API that takes in a nullHandlingMode, and will share the PR for review soon.

pramodsatya:
@aditi-pandit, could you please help review this change? It builds on https://github.com/facebookincubator/velox/pull/15726 and is required to enable the IN-rewrite.

pramodsatya:
@aditi-pandit, @mbasmanova, could you please help review this change?

Issue #1573, opened 2022-05-10
Adding validation logic in Substrait-to-Velox conversion


rui-mo:
A Substrait plan validation can be used to determine whether to offload the computing to Velox. If the validation fails, the compute will fallback to Spark for execution.

Instead of just trying the plan conversions and catching the exceptions, a separate validation logic is needed because there are some differences between the validation and the real conversion, including:

- In real conversion, a node can need a child node as its input, but the Substrait plan for its validation only contains itself, which can cause failure if we send this plan to the real conversion.
- In real conversion, the input types of a plan node can be deduced from its child node, but in validation we need to send them through the Substrait plan for a type validation.
- There are some code not needed for validation in real conversion.
- There are some special logic in the validation because current supporting and verifying is still on-going.

Currently the validation is strict, and the not-verified or unsupported cases in TPC-H would fallback. As we continue working on this, the conditions for a validation success would become more broad and general.

mbasmanova:
@rui-mo Rui, thank you for creating an issue to describe the new validation logic.

> There are some code not needed for validation in real conversion.

Would you give some examples? My reading of the attached PR suggests that one difference is that validation compiles the filter and project expressions using ExprSet and returns false if compilation throws (e.g. a function doesn't exist) while conversion doesn't have that logic.

> There are some special logic in the validation because current supporting and verifying is still on-going.

Would you give some examples?

rui-mo:
Thanks for your comment.
> There are some code not needed for validation in real conversion.

Generally, the code relating to the plan node creation can be not needed. For example, in aggregation node creation, because Spark's partial avg outputs two columns of sum and count, if detecting final avg, we need to add a project node before aggregation to combine them into the column of row(double, bigint). But in validation, this logic is not needed.

> There are some special logic in the validation because current supporting and verifying is still on-going.

For example, there are two kinds of filters in HiveTableHandle, the subfieldFilters and remainingFilter. Because we only supported the subfieldFilters, the ReadRel validation will exclude the cases cannot be handled by current TableScan node creation.

mbasmanova:
@rui-mo Rui, thank you for the clarifications. These are helpful. 

I realized that I'm lacking visibility into the end-to-end workflow you are working on.

I'm imagining you are integrating Spark's front-end with Velox-based evaluation engine. I'm picturing the following steps:

1. Generate query plan using Spark and convert it into Substrait plan.
2. Convert Substrait plan to Velox plan.
[...]

PR #15720, opened 2025-12-08
fix: Enable map_keys_overlap for expression fuzzer

abhinavmuk04:
Summary: Remove map_keys_overlap from skipFunctions to enable fuzzer testing. The function is already correctly skipped from SOT verification since it's Velox-only.

Differential Revision: D88680724


kevinwilfong (approved):
Thanks! This function should only be registered in Presto and the Spark fuzzer has been flaky recently so I'm not too concerned about the fuzzer failure, but if you could rebase on https://github.com/facebookincubator/velox/pull/1573 and get a clean run that would be great!

PR #15726, opened 2025-12-09
feat: Add variant comparison with null handling mode

pramodsatya:
Enables comparison of variants, with `Variant::equals` API, to use custom null handling mode. 
Required for constant expression comparison using `null-as-indeterminate` null comparison semantics (https://github.com/facebookincubator/velox/pull/15705).


velox/type/Variant.h
```diff
@@ -648,6 +649,10 @@ class Variant {
 
   bool operator<(const Variant& other) const;
 
+  std::optional<bool> equals(
```
aditi-pandit, line 665: Can you document this API ? In what cases does this method return std::nullopt ? Is there a reference for the interpretation of NullHandlingMode ? If yes, then please add it here.

velox/type/Variant.cpp
```diff
@@ -83,9 +82,12 @@ struct VariantEquality<TypeKind::ARRAY> {
     }
     for (size_t i = 0; i != aArray.size(); ++i) {
       // todo(youknowjack): switch outside the loop
-      bool result =
-          dispatchDynamicVariantEquality(aArray[i], bArray[i], NullEqualsNull);
-      if (!result) {
+      auto compareResult = dispatchDynamicVariantEquality(
+          aArray[i], bArray[i], nullHandlingMode);
+      if (!compareResult.has_value()) {
```
aditi-pandit: Can we just return compareResult itself ? Same comment in other places in this file.

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,34 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
 }
 
+TEST(VariantTest, equalsWithNullAsIndeterminate) {
+  auto testNullEquality = [](const TypeKind& kind) {
+    const auto null = Variant::null(kind);
+    ASSERT_EQ(
+        Variant::null(kind).equals(
+            null, CompareFlags::NullHandlingMode::kNullAsIndeterminate),
+        std::nullopt);
+    auto compareResult = Variant::null(kind).equals(
```
aditi-pandit: Can we reuse the above variable null (at line 804) here ?

velox/type/Variant.cpp
```diff
@@ -83,9 +82,12 @@ struct VariantEquality<TypeKind::ARRAY> {
...
+        return compareResult;
+      }
+      if (!compareResult.value()) {
```
aditi-pandit: I think we can combine both conditions to return compareResult...
if (!compareResult.value()) return false else return true

is also the same as compareResult.

pramodsatya: Sorry I don't follow, could you please clarify if this conforms to the comparison logic in https://github.com/facebookincubator/velox/blob/b88ce66a7b490886260d011c42e6a41d014adf54/velox/common/base/CompareFlags.h#L64 ?

pramodsatya: @aditi-pandit, consolidated the conditions to return compareResult using `value_or(false)`, is this fine?
```
    for (...) {
      if (!compareResult.value_or(false)) {
        return compareResult;
      }
    }
    return true;
```

velox/type/Variant.cpp
```diff
@@ -1108,10 +1127,7 @@ void Variant::verifyArrayElements(const std::vector<Variant>& inputs) {
 }
 
 bool Variant::equalsWithNullEqualsNull(const Variant& other) const {
-  if (other.kind_ != this->kind_) {
-    return false;
-  }
-  return dispatchDynamicVariantEquality(*this, other, true);
+  return this->equals(other);
```
czentgr: This function is now deprecated. Users can use `equals` with the same semantic (`kNullAsValue`).
We should document that change to equals because it handles the nested case properly.

Maybe we can clean up the usage of `equalsWithNullEqualsNull` in a separate PR?

pramodsatya: Updated, thanks. I didn't cleanup this method as I wasn't sure if this was being used internally at Meta. I can cleanup the usage in a follow-up PR after checking with Meta reviewers.

pramodsatya:
Thanks for the feedback @aditi-pandit, @czentgr, I have addressed the review comments. Could you please take another look?

velox/type/Variant.h
```diff
@@ -648,8 +649,25 @@ class Variant {
...
+  ///  - `true` or `false` when `nullHandlingMode` is `kNullAsValue` and the
+  ///     comparison is determinate.
+  ///
+  /// See `CompareFlags::NullHandlingMode` for the interpretation of the
+  /// null-handling semantics for different types.
+  std::optional<bool> equals(
+      const Variant& other,
+      CompareFlags::NullHandlingMode nullHandlingMode) const;
+
   bool equals(const Variant& other) const;
```
mbasmanova, line 672: Would you add a comment to this method to clarify that it is a shortcut for the other one with nullHandlingMode = ...?

velox/type/Variant.cpp
mbasmanova: I don't see any usage of equalsWithNullEqualsNull outside of Variant class. Let's clean up in this PR.

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,33 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
+  //  - indeterminate with kNullAsIndeterminate nullHandlingMode.
+  //  - true with kNullAsValue nullHandlingMode.
+  testNullEquality(TypeKind::SMALLINT);
+  testNullEquality(TypeKind::INTEGER);
+  testNullEquality(TypeKind::BIGINT);
+  testNullEquality(TypeKind::VARCHAR);
+  testNullEquality(TypeKind::VARBINARY);
+  testNullEquality(TypeKind::DOUBLE);
+  testNullEquality(TypeKind::TIMESTAMP);
+  testNullEquality(TypeKind::ARRAY);
```
mbasmanova, line 824: Can we add tests for comparing non-null complex types with nulls inside?

velox/type/Variant.cpp
```diff
@@ -25,19 +25,18 @@
...
-    const bool& enableNullEqualsNull);
+    CompareFlags::NullHandlingMode nullHandlingMode);
 
-template <bool nullEqualsNull>
-bool evaluateNullEquality(const Variant& a, const Variant& b) {
-  if constexpr (nullEqualsNull) {
-    if (a.isNull() && b.isNull()) {
-      return true;
-    }
+template <CompareFlags::NullHandlingMode nullHandlingMode>
```
mbasmanova: Why is this a template parameter? Seems like it can be a regular parameter, no?

pramodsatya: Thanks for the suggestion, I had refactored existing template parameter `nullEqualsNull` to `nullHandlingMode` to keep changes minimal earlier, made it a regular parameter now.

velox/type/Variant.cpp
pramodsatya: Reverted, thanks.

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,76 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
+  testNullEquality(TypeKind::MAP);
+  testNullEquality(TypeKind::ROW);
+
+  // Test comparing non-null complex types with nulls inside.
+  // Arrays with null elements compared to each other.
+  {
+    auto arr1 = Variant::array({Variant(1), Variant::null(TypeKind::INTEGER)});
```
mbasmanova: arr1, arr2, etc. names are anti-pattern.

consider adding a helper function to create arrays from std::optional<int32_t>

```
auto a = makeArray({1, std::nullopt});
auto b = makeArray({1, std::nullopt});
auto c = makeArray({2, std::nullopt});

EXPECT_THAT(a.equals(b, kNullAsValue), ::testing::Optional(true))
ASSERT_EQ(a.equals(b, kNullAsIndeterminate), std::nullopt);
...
```

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,76 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
   ASSERT_FALSE(Variant(sum1).equalsWithEpsilon(Variant(sum3)));
 }
 
+TEST(VariantTest, equalsWithNullHandlingMode) {
```
mbasmanova: This method is rather long. Consider splitting into test for primitive values and separate test for each complex type.

velox/type/Variant.cpp
```diff
@@ -83,10 +83,10 @@ struct VariantEquality<TypeKind::ARRAY> {
     }
     for (size_t i = 0; i != aArray.size(); ++i) {
       // todo(youknowjack): switch outside the loop
-      bool result =
-          dispatchDynamicVariantEquality(aArray[i], bArray[i], NullEqualsNull);
-      if (!result) {
-        return false;
+      auto compareResult = dispatchDynamicVariantEquality(
+          aArray[i], bArray[i], nullHandlingMode);
+      if (!compareResult.value_or(false)) {
```
kKPulla: When the nullHandlingMode is _kNullAsIndeterminate_ we shouldn't be doing this early return but instead loop through to see if we have any **exact mismatch** right? 

Ex: Comparing [null, 1] and [null, 2], I assume we want the result to be _false_ and not _std::nullopt_ ?

Ditto for MAP and ROW types

velox/type/Variant.h
```diff
@@ -617,7 +618,8 @@ class Variant {
 
   struct NullEqualsNullsComparator {
     bool operator()(const Variant& a, const Variant& b) const {
-      return a.equalsWithNullEqualsNull(b);
+      return a.equals(b, CompareFlags::NullHandlingMode::kNullAsValue)
```
kKPulla: _kNullAsValue_ should always return a boolean value right? If so, can we instead have a CHECK here instead of silently replacing with _false_?

velox/type/Variant.cpp
```diff
@@ -137,29 +137,40 @@ struct VariantEquality<TypeKind::MAP> {
...
+std::optional<bool> dispatchDynamicVariantEquality(
     const Variant& a,
     const Variant& b,
-    const bool& enableNullEqualsNull) {
-  if (enableNullEqualsNull) {
-    return VELOX_DYNAMIC_TYPE_DISPATCH_METHOD(
-        VariantEquality, equals<true>, a.kind(), a, b);
+    CompareFlags::NullHandlingMode nullHandlingMode) {
+  if (nullHandlingMode == CompareFlags::NullHandlingMode::kNullAsValue) {
+    return VELOX_DYNAMIC_TYPE_DISPATCH_METHOD_ALL(
```
kKPulla, line 189: Is this change to support ALL types intentional? Curious as this was not mentioned in the PR specifically.

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,76 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
+    auto arr1 = Variant::array({Variant(1), Variant::null(TypeKind::INTEGER)});
+    auto arr2 = Variant::array({Variant(1), Variant::null(TypeKind::INTEGER)});
+    auto arr3 = Variant::array({Variant(2), Variant::null(TypeKind::INTEGER)});
+
+    // Both are non-null arrays with null elements inside.
+    auto res1 = arr1.equals(arr2, CompareFlags::NullHandlingMode::kNullAsValue);
+    ASSERT_TRUE(res1.has_value() && res1.value());
+
+    // Arrays with different elements and null inside.
+    auto res2 = arr1.equals(arr3, CompareFlags::NullHandlingMode::kNullAsValue);
```
kKPulla: related to the comment about early return when null, we should test with arrays having elements after null too

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,76 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
+  testNullEquality(TypeKind::VARCHAR);
+  testNullEquality(TypeKind::VARBINARY);
+  testNullEquality(TypeKind::DOUBLE);
+  testNullEquality(TypeKind::TIMESTAMP);
+  testNullEquality(TypeKind::ARRAY);
+  testNullEquality(TypeKind::MAP);
+  testNullEquality(TypeKind::ROW);
+
+  // Test comparing non-null complex types with nulls inside.
+  // Arrays with null elements compared to each other.
```
kKPulla: nit: can we also add a test for MAP alongside ARRAY and ROW that you added already?

velox/type/Variant.cpp
pramodsatya: Thanks for pointing it out, updated accordingly.

velox/type/Variant.cpp
pramodsatya: Yes this is intentional and is added to fix failing tests such as `ConstantTypedExprTest.serdeComplexTypes`. The new equals implementation dispatches into `OPAQUE` and `UNKNOWN` types via the dynamic dispatcher calls for ARRAY/ROW/MAP types, and `VELOX_DYNAMIC_TYPE_DISPATCH_METHOD_ALL` is required to handle these cases.

pramodsatya:
Thanks for the feedback @kKPulla, @mbasmanova, I have addressed the review comments. Could you please take another look?

velox/type/tests/VariantTest.cpp
```diff
@@ -799,6 +799,156 @@ TEST(VariantTest, equalsWithEpsilonFloat) {
...
+      if (element.has_value()) {
+        arrayElements.emplace_back(element.value());
+      } else {
+        arrayElements.push_back(Variant::null(TypeKind::INTEGER));
+      }
+    }
+    return Variant::array(arrayElements);
+  };
+
+  auto a = makeArray({std::nullopt, 1});
```
kKPulla: nit: To avoid missing any edge cases in the future, probably we can mix and match these - say have non-null values before and after null? Ditto for other types too.

Also thanks for splitting tests for each type.

kKPulla:
> Thanks for the feedback @kKPulla, @mbasmanova, I have addressed the review comments. Could you please take another look?

Thanks for addressing the feedback and improving this API. Added one minor comment and once addressed, I will add the PR to merge queue.

velox/type/tests/VariantTest.cpp
pramodsatya: Thanks @kKPulla, updated. Could you PTAL?

kKPulla:
Thanks again for your contribution and addressing the feedback above. Added to the merge queue.

PR #15727, opened 2025-12-09
perf: Optimize toBitMask() and VectorHasher for Arm NEON

LDong-Arm:
This pull request implements `toBitMask()` for Arm NEON and optimizes how xsimd boolean operators and `toBitMask()` are combined. For details, see the commit messages.

### DIFF
--- a/velox/common/base/SimdUtil-inl.h
+++ b/velox/common/base/SimdUtil-inl.h
@@ -91,10 +91,24 @@ struct BitMask<T, A, 1> {
 
 #if XSIMD_WITH_NEON
   static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::neon&) {
-    alignas(A::alignment()) static const int8_t kShift[] = {
-        -7, -6, -5, -4, -3, -2, -1, 0, -7, -6, -5, -4, -3, -2, -1, 0};
-    int8x16_t vshift = vld1q_s8(kShift);
-    uint8x16_t vmask = vshlq_u8(vandq_u8(mask, vdupq_n_u8(0x80)), vshift);
+    static constexpr uint8x16_t vmask_const = {
+        0x01,
+        0x02,
+        0x04,
+        0x08,
+        0x10,
+        0x20,
+        0x40,
+        0x80, // lower half
+        0x01,
+        0x02,
+        0x04,
+        0x08,
+        0x10,
+        0x20,
+        0x40,
+        0x80}; // upper half
+    uint8x16_t vmask = vandq_u8(mask.data, vmask_const);
     return (vaddv_u8(vget_high_u8(vmask)) << 8) | vaddv_u8(vget_low_u8(vmask));
   }
 #endif
@@ -125,6 +139,15 @@ struct BitMask<T, A, 2> {
   }
 #endif
 
+#if XSIMD_WITH_NEON
+  static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::neon&) {
+    static constexpr uint16x8_t vmask_const = {
+        0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80};
+    uint16x8_t vmask = vandq_u16(mask.data, vmask_const);
+    return vaddvq_u16(vmask);
+  }
+#endif
+
   static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::generic&) {
     return genericToBitMask(mask);
   }
@@ -143,6 +166,14 @@ struct BitMask<T, A, 4> {
 #if XSIMD_WITH_SSE2
   static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::sse2&) {
     return _mm_movemask_ps(reinterpret_cast<__m128>(mask.data));
+  }
+#endif
+
+#if XSIMD_WITH_NEON
+  static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::neon&) {
+    static constexpr uint32x4_t vmask_const = {0x1, 0x2, 0x4, 0x8};
+    uint32x4_t vmask = vandq_u32(mask.data, vmask_const);
+    return vaddvq_u32(vmask);
   }
 #endif
 
@@ -170,6 +201,14 @@ struct BitMask<T, A, 8> {
 #if XSIMD_WITH_SSE2
   static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::sse2&) {
     return _mm_movemask_pd(reinterpret_cast<__m128d>(mask.data));
+  }
+#endif
+
+#if XSIMD_WITH_NEON
+  static int toBitMask(xsimd::batch_bool<T, A> mask, const xsimd::neon&) {
+    static constexpr uint64x2_t vmask_const = {0x1, 0x2};
+    const uint64x2_t vmask = vandq_u64(mask.data, vmask_const);
+    return vaddvq_u64(vmask);
   }
 #endif
 


velox/exec/HashTable.cpp
```diff
@@ -1128,8 +1128,7 @@ void HashTable<ignoreNullKeys>::insertForGroupBy(
       for (int64_t numProbedBuckets = 0; numProbedBuckets < numBuckets();
            ++numProbedBuckets) {
         MaskType free =
-            ~simd::toBitMask(
-                BaseHashTable::TagVector::batch_bool_type(tagsInTable)) &
+            ~simd::toBitMask(tagsInTable != TagVector::broadcast(0)) &
```
LDong-Arm: Note: Comparing a batch of tags against a zero batch is what `fullProbe()` and `toString()` in the same file (`HashTable.cpp`) already do, so the change here merely aligns how it is done.

LDong-Arm:
The fundamental idea here is that `toBitMask()` takes `xsimd::batch_bool` as input, and `xsimd::batch_bool` is free to have an architecture-dependent meaning (which is abstracted out from the user code). Any code that uses XSimd should only generate `xsimd::batch_bool` from comparison operators of `xsimd::batch` - it should _not_ create a raw value of `xsimd::batch_bool` directly.

This guarantees that `xsimd::batch_bool` on NEON has each vector lane equal to an "all ones" (true) or "all zeroes" (false) bitmask. (As far as I see, comparison instructions of x86's SSE and AVX also have this behaviour.) And "all ones" is a very nice property for vector-to-bitmask conversion. Previously, the difficulty in implementing an efficient `toBitMask` was due to assuming only the top-bit was set.

~~I have searched for all usage of `toBitMask()` in the Velox repo, and in all cases but one, the `xsimd::batch_bool` argument is already compliant - generated by comparison operators.~~ (Maybe this needs more rigorous checking.)  A "non-compliant" case I've spotted is `HashTable<ignoreNullKeys>::insertForGroupBy()`, and even the rest of the `HashTable` functions that use `toBitMask()` already does so in a compliant way, so I've included a fix. This was actually caught by a unit test failure.

LDong-Arm:
@Yuhta Please let me know what you think about the idea I've described above, and if I've missed any assumptions. Thanks.

velox/exec/VectorHasher-inl.h
```diff
@@ -28,14 +28,13 @@ bool VectorHasher::tryMapToRangeSimd(
   constexpr int kWidth = xsimd::batch<T>::size;
   for (; row + kWidth <= rows.end(); row += kWidth) {
     auto data = xsimd::load_unaligned(values + row);
-    int32_t gtMax = simd::toBitMask(data > allHigh);
-    int32_t ltMin = simd::toBitMask(data < allLow);
+    int32_t outOfRange = simd::toBitMask((data > allHigh) || (data < allLow));
```
Yuhta: We don't really need these change, do we?  Back then I measured in some cases the bitmask OR is faster than the SIMD one (depending on the situation of the instruction pipeline though), so it's better not to change this part unless necessary

velox/exec/HashTable.cpp
Yuhta: Can we make sure on x86 this change is not generating extra instructions?  Otherwise we may need to guard it with macro.

velox/exec/VectorHasher-inl.h
LDong-Arm: I'm using [the VectorHasher benchmark](https://github.com/facebookincubator/velox/blob/main/velox/exec/benchmarks/VectorHasherBenchmark.cpp) to check performance, and the SIMD OR (`simd::toBitMask((data > allHigh) || (data < allLow))`) significantly improves performance of GCC (I've tried 13 and 15) on **AMD Zen 4** for the following:

* `computeValueIdsBigintNoNulls`
* `computeValueIdsIntegerNoNulls`
* `computeValueIdsSmallintNoNulls`

The other sub-benchmarks are unaffected.

As for Clang 20 on AMD Zen 4, most sub-benchmarks are not really impacted by SIMD OR vs bitmask OR - maybe Clang does further optimisation to generate equivalent code. But in one case (`computeValueIdsLowCardinalityNotAllUsed`), the SIMD OR is indeed _slower_ than the bitmask OR (and both are faster than GCC).

But on Arm Neoverse V2, _both GCC and Clang_ benefit heavily from the SIMD OR, for the three `computeValueIds[Bigint,Integer,Smallint]NoNulls` sub-benchmarks. The other sub-benchmarks' scores are unaffected. I think it's because even as `toBitMask()` is optimised by this pull request, it still requires two NEON instructions (unlike AVX2 requiring one, `_mm256_movemask_...`), so calling `toBitMask()` as few times as possible is more beneficial here.

I think even for the same architecture, different CPUs may have different SIMD throughput relative to scalar operations.
So, which sub-benchmarks are more representative of real use cases? The `computeValueIds**NoNulls` ones, or the `computeValueIdsLowCardinalityNotAllUsed`, or something else? If the benefit/drawback of SIMD OR is ambiguous (highly situational and hardware dependent), I can remove this from the current PR.

velox/exec/HashTable.cpp
LDong-Arm: I'm thinking of adding a new `toBitMask(xsimd::batch)` and passing `tagsInTable` to it (because it's really a `xsimd::batch`), which will be efficient on x86.

The difference will be that
* `toBitMask(xsimd::batch_bool)` will strictly take a "real" `batch_bool`, and we can leverage any assumptions like `true == -1` and do architecture-dependent optimisations
* `toBitMask(xsimd::batch)` will take "loose" values like `tagsInTable`, which can be force cast to `toBitMask(xsimd::batch_bool)` on x86 to not waste instructions, but implemented as `toBitMask(value != xsimd::broadcast(0))` as a fallback (as this should always be valid).

Will update once I have this implemented.

velox/exec/VectorHasher-inl.h
Yuhta: It's mixed, let's put this change in a separate PR so if there is any significant effect, it can be isolated.  This PR should be neutral on x86 and pure improvement on arm.

velox/exec/HashTable.cpp
Yuhta: Yes we can do that

LDong-Arm: This turns out to be a bit trickier than I thought.
The x86 AVX `toBitMask()` based on `_mm256_movemask_epi8` builds a bitmask from the top bits of the 8-bit tags we pass here.

- For a normal used tag, the top bit is always set, whereas the lower 7 bits hold the actual value - in this case, checking top bit gives the same result as simply checking non-zero, and the non-zero check is easier to do with Arm Neon.
- For the special `static constexpr uint8_t kTombstoneTag = 0x7f`, the top bit is not set, so checking non-zero will give a different behavior.

~~Does `insertForGroupBy()` here ever see or care about tombstones? Please me know if you have an immediate answer to this, otherwise I'll dig further into the tombstone logic and come up with an update.~~

Update: This code path is only hit during `rehash()` on a freshly (re)allocated `table_`, so we don't need to handle `kTombstoneTag`.

velox/exec/VectorHasher-inl.h
LDong-Arm: Done. I've removed it from this PR.

velox/exec/HashTable.cpp
LDong-Arm: In the end, I kept it simple and just used `#if XSIMD_WITH_SSE2` to keep existing behavior on x86 and added a inline comment to the code. Please review, thanks.

LDong-Arm:
@Yuhta

The essence of this PR is to let each architecture implement `toBitMask()` in its _native_ boolean representation, rather than forcing the SSE/AVX-specific boolean (e.g. building bitmasks from high bits of elements). As long as `xsimd::batch_bool` is derived from comparison operators (like `>`, `!=`), its value is guaranteed to be a correct native boolean, thanks to the "magic" of xsimd. Reinterpreting `batch` to `batch_bool` only works on SSE/AVX and is actually optimal there, but it should be avoided on other architectures.

Below, I've done a check of _all_ calls to `toBitMask()` in Velox. The good news is, all other calls are already done in ways that are neutral to architecture. The only exceptions is `HashTable::insertForGroupBy()`, which this PR fixes as per our earlier discussion. The full report is below.

<details>

<summary>Click to expand the report</summary>

## Rigorous ways of creating xsimd::batch_bool on all architectures

The underlying representation of xsimd::batch_bool depends on architecture. To ensure xsimd::batch_bool and simd::toBitMask() work correctly on all supported architectures, we should generally rely on the xsimd API to create truly valid xsimd::batch_bool for us.

The xsimd API creates valid xsimd::batch_bool mainly in a few ways:
- declaring a new xsimd::batch_bool variable with an initial value
- results of comparisons on two plain xsimd::batch variables
    - examples: `>`, `<`, `!=`, `==`
- results of operations on an existing valid xsimd::batch_bool variables
    - for example, bitwise (`&`, `|`) operations to combine two existing `xsimd::batch_bool` variables
- loading from memory via xsimd::batch_bool::load_unaligned()
    - note: even if the loaded memory is not a proper ximd::batch_bool representation, xsimd::batch_bool::load_unaligned() internally converts the values to a valid xsimd::batch_bool by doing a logical comparison on the loaded elements with a vector filled with zeros (see `xsimd_common_memory.hpp` in xsimd).

## Optimization specific to x86

Directly reinterpreting xsimd::batch to batch::batch_bool
- This is a shortcut on the SSE/AVX family of extensions, in cases where we only care about the high bit of each element of xsimd::batch. The implementation of `simd::toBitMask()` for SSE/AVX uses instructions that create a bitmask by combining high bits of elements.
- The generic way to ensure correctness on all architectures is to compare the `xsimd::batch` input with a reference value

## Checking uses of `simd::toBitMask()` in the whole Velox

Below lists how `xsimd::batch_bool` is created and passed to `simd::toBitMask`.

⚠️ velox/exec/HashTable.cpp
- ✅ comparison operators on plain xsimd::batch
- ⚠️ force-cast from xsimd::batch - we should only do this on x86
  - **addressed by this pull request**

✅ velox/dwio/common/SelectiveColumnReader.cpp
- comparison operators on plain xsimd::batch

✅ velox/dwio/common/ColumnVisitors.h
- function arguments of the type xsimd::batch_bool
    - on the caller side, those function arguments are created with `simd::fromBitMask()` which works correctly on all architectures
        - `simd::fromBitMask()` is implemented based on `xsimd::batch_bool::load_unaligned()`, which internally ensures the output is correct on any architectures in a rigorous way.
- in processRun():
    - non-SVE (the check should ideally be SSE/AVX): force-cast from xsimd::batch
    - SVE (the check should ideally be non-SSE/non-AVX): comparison operators on plain xsimd::batch
- ℹ️ so, this already applies the x86-specific optimization while keeping things correct/neutral for Arm SVE - like what we've done with HashTable in this PR. This fast path is not yet enabled on any other architectures like Neon - if we extend support, we need to be careful and switch to better choices of macros: https://github.com/facebookincubator/velox/blob/51a0a3d75e54ec95bebbdab49dac9beb7a0671c1/velox/dwio/common/ColumnVisitors.h#L923 (maybe check SSE/AVX instead of SVE_BITS)

✅ velox/dwio/common/DecoderUtil.h
- created from various overloads of testValues() in velox/type/Filter.h
    - the overloads in velox/type/Filter.h are all based on comparison operators on plain xsimd::batch

✅ velox/dwio/dwrf/reader/SelectiveStringDirectColumnReader.cpp
- comparison operators on plain xsimd::batch

✅ velox/functions/lib/SIMDComparisonUtil.h
- comparison operators on plain xsimd::batch

✅ velox/exec/VectorHasher.cpp
- comparison operators on plain xsimd::batch

✅ velox/exec/VectorHasher-inl.h
- comparison operators on plain xsimd::batch

✅ velox/exec/RowContainer.cpp
- comparison operators on plain xsimd::batch

✅ velox/serializers/KeyEncoder.cpp
- comparison operators on plain xsimd::batch

✅ velox/type/StringView.cpp
- comparison operators on plain xsimd::batch

✅ velox/type/Filter.h
- all implementations create return values in semantically correct ways (i.e. using comparison operators on plain xsimd::batch inputs)

✅ velox/type/tests/FilterTest.cpp
- created from various overloads of testValues() in velox/type/Filter.h

✅ velox/type/tests/FilterBenchmark.cpp
- created from various overloads of testValues() in velox/type/Filter.h

✅ velox/common/memory/MmapAllocator.cpp
- comparison operators on plain xsimd::batch

✅ velox/common/base/BigintIdMap.h
- comparison operators on plain xsimd::batch
- results of simd::fromBitMask()
- bit-wise operators on two xsimd::batch_bool

✅ velox/common/base/tests/SimdUtilTest.cpp
- comparison operators on plain xsimd::batch

✅ velox/experimental/wave/vector/WaveVector.cpp
- comparison operators on plain xsimd::batch

✅ velox/experimental/wave/common/tests/CpuTable.h
- comparison operators on plain xsimd::batch

✅ velox/experimental/wave/common/tests/CudaTest.cpp
- comparison operators on plain xsimd::batch
- directly instantiated xsimd::batch_bool

</details>

LDong-Arm:
Hi @Yuhta , I have addressed your comments, could you please have a look? Thanks!

LDong-Arm:
Rebased

jainxrohit:
@LDong-Arm Can you please look in to the pre-commit checks failures?

LDong-Arm:
@jainxrohit I have updated the code based on the clang-format suggestion

Issue #15731, opened 2025-12-09
`extractFiltersFromRemainingFilter` fails with some bigint filters


marin-ma:
### Bug description

We encountered below in error in Gluten CI

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 86.0 failed 1 times, most recent failure: Lost task 0.0 in stage 86.0 (TID 81) (fc99d8effa46 executor driver): org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Source: USER
Error Code: UNSUPPORTED
Reason: Filter(MultiRange, deterministic, no nulls): testInt64Range() is not supported.
Retriable: False
Context: Split [Hive: file:///__w/incubator-gluten/incubator-gluten/backends-velox/target/scala-2.13/test-classes/tpch-data-parquet/lineitem/part-00000-6c374e0a-7d76-401b-8458-a8e31f8ab704-c000.snappy.parquet 0 - 1863237] Task Gluten_Stage_86_TID_81_VTID_20363
Function: testInt64Range
File: /work/ep/build-velox/build/velox_ep/./velox/type/Filter.h
Line: 241
```

The failure can be reproduced by the following unit test:

```
TEST_F(HiveConnectorTest, bigintFilters) {
  auto expr = parseExpr("c0 not in (1, 3) or c0 in (1, 2)", rowType);
  SubfieldFilters filters;
  double sampleRate = 1;
  auto remaining = extractFiltersFromRemainingFilter(
      expr, &evaluator, filters, sampleRate);
  ASSERT_EQ(sampleRate, 1);
  ASSERT_EQ(filters.size(), 1);
  ASSERT_EQ(filters.begin()->first, Subfield("c0"));
  ASSERT_TRUE(filters.begin()->second->testInt64Range(0, 2, false));
}
```

Output for the unit test:
```
C++ exception with description "Exception: VeloxUserError
Error Source: USER
Error Code: UNSUPPORTED
Reason: Filter(MultiRange, deterministic, no nulls): testInt64Range() is not supported.
Retriable: False
Function: testInt64Range
File: /Users/rong/workspace/github/apache/gluten/ep/build-velox/build/velox_ep/./velox/type/Filter.h
Line: 241
```

### System information

Velox System Info v0.0.2
Commit: 35b8a49e264d1813f631ab75ecdeebc8be296b31
CMake Version: 4.1.3
System: Darwin-24.5.0
[...]

PR #15732, opened 2025-12-09
perf: Optimize extractFiltersFromRemainingFilter for a mix of OR and NOT

marin-ma:
Currently only `FilterKind::kBigIntRange` and `FilterKind::kBigintmultiRange` are supported in merging. Without this fix, the rest of the types for bigint will be combined into a `MultiRange` filter, which doesn't support `testInt64Range` and `testInt64`. This patch ensures that these filter types are excluded.

Related issue: #15731

### DIFF
--- a/velox/exec/tests/TableScanTest.cpp
+++ b/velox/exec/tests/TableScanTest.cpp
@@ -2515,7 +2515,7 @@ TEST_F(TableScanTest, statsBasedSkippingWithoutDecompression) {
   auto assertQuery = [&](const std::string& filter) {
     auto rowType = asRowType(rowVector->type());
     return TableScanTest::assertQuery(
-        PlanBuilder(pool_.get()).tableScan(rowType, {filter}).planNode(),
+        PlanBuilder(pool_.get()).tableScan(rowType, {}, {filter}).planNode(),
         filePaths,
         "SELECT * FROM tmp WHERE " + filter);
   };
--- a/velox/exec/tests/utils/PlanBuilder.cpp
+++ b/velox/exec/tests/utils/PlanBuilder.cpp
@@ -214,10 +214,10 @@ std::pair<common::Subfield, std::unique_ptr<common::Filter>> toSubfieldFilter(
       auto left = toSubfieldFilter(call->inputs()[0], evaluator);
       auto right = toSubfieldFilter(call->inputs()[1], evaluator);
       VELOX_CHECK(left.first == right.first);
-      return {
-          std::move(left.first),
-          exec::ExprToSubfieldFilterParser::makeOrFilter(
-              std::move(left.second), std::move(right.second))};
+      auto filter = exec::ExprToSubfieldFilterParser::makeOrFilter(
+          std::move(left.second), std::move(right.second));
+      VELOX_CHECK_NOT_NULL(filter);
+      return {std::move(left.first), std::move(filter)};
     }
 
     if (call->name() == "not") {


velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -950,12 +963,18 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
 
       if (disjuncts.empty()) {
         subfield = tmpFilters.begin()->first.clone();
-      } else if (!(subfield == tmpFilters.begin()->first)) {
+      } else if (subfield != tmpFilters.begin()->first) {
+        disjuncts.clear();
+        break;
+      }
+
+      auto filter = tmpFilters.begin()->second->clone();
```
Yuhta: We can check it without clone

hdikeman:
Hi @marin-ma, apologies didn't see this one before I opened https://github.com/facebookincubator/velox/pull/15738

It is a fix to avoid some of the MultiRange int cases

Could you take a look? Based on your commit perhaps I also need to add support for negated bigint values/ranges? I hadn't seen that failure but lmk what you think, thanks

marin-ma:
@hdikeman Could you try the unit test I posted https://github.com/facebookincubator/velox/issues/15731 in with your fix? The unit test can reproduce the error we encountered in Gluten's CI.

Paste the unit test below:

```
TEST_F(HiveConnectorTest, bigintFilters) {
  auto expr = parseExpr("c0 not in (1, 3) or c0 in (1, 2)", rowType);
  SubfieldFilters filters;
  double sampleRate = 1;
  auto remaining = extractFiltersFromRemainingFilter(
      expr, &evaluator, filters, sampleRate);
  ASSERT_EQ(sampleRate, 1);
  ASSERT_EQ(filters.size(), 1);
  ASSERT_EQ(filters.begin()->first, Subfield("c0"));
  ASSERT_TRUE(filters.begin()->second->testInt64Range(0, 2, false));
}
```

velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -251,6 +251,19 @@ bool isSpecialColumn(
   return specialName.has_value() && name == *specialName;
 }
 
+bool supportsMergeWithOr(const common::FilterPtr& filter) {
+  switch (filter->kind()) {
+    case common::FilterKind::kBigintValuesUsingHashTable:
+    case common::FilterKind::kBigintValuesUsingBitmask:
+    case common::FilterKind::kNegatedBigintRange:
```
mbasmanova: Is there a reason why negated ranges cannot be combined into BigintMultiRange ? Is this just a bug / missing functionality in the filters?

velox/connectors/hive/HiveConnectorUtil.cpp
Yuhta, line 962: Actually let's check we don't generate any `MultiRange` here.  The performance of `MultiRange` is worse than OR expressions and the sole purpose of it is for correctness, so there is no point to combine them as an optimization step.

hdikeman:
> @hdikeman Could you try the unit test I posted #15731 in with your fix? The unit test can reproduce the error we encountered in Gluten's CI.
> 
> Paste the unit test below:
> 
> ```
> TEST_F(HiveConnectorTest, bigintFilters) {
>   auto expr = parseExpr("c0 not in (1, 3) or c0 in (1, 2)", rowType);
>   SubfieldFilters filters;
>   double sampleRate = 1;
>   auto remaining = extractFiltersFromRemainingFilter(
>       expr, &evaluator, filters, sampleRate);
>   ASSERT_EQ(sampleRate, 1);
>   ASSERT_EQ(filters.size(), 1);
>   ASSERT_EQ(filters.begin()->first, Subfield("c0"));
>   ASSERT_TRUE(filters.begin()->second->testInt64Range(0, 2, false));
> }
> ```

I would need to add handling for `kNegatedBigintValues[...]` I believe due to the `c0 not in (1, 3)`, I restricted the fix originally to just the two non-negated values filters

I think we could handle all the negated cases by just inverting the ranges before combining

But how do you think we should approach this @marin-ma? Your approach of leaving the bigint filters inside the remaining filter would also work, not sure if you know a compelling reason to go one way or the other

hdikeman:
@marin-ma feel free to take a look at https://github.com/facebookincubator/velox/pull/15738

I updated with negative filter handling as you mentioned

I'd be fine dropping if you think it's preferable to stuff everything into the remaining filter (as in this PR). Just let me know, thanks

hdikeman:
Hi @marin-ma I spoke with @Yuhta

Perhaps you can merge the fix in this PR and we can land https://github.com/facebookincubator/velox/pull/15738 later as needed? Jimmy pointed out there may be a few additional changes needed to polish up on that PR anyways

Sorry for the miscommunication here, I was not aware this was a recent regression, I thought this was an existing functionality gap, so I did not search for existing issues before opening https://github.com/facebookincubator/velox/pull/15738

velox/connectors/hive/HiveConnectorUtil.cpp
Yuhta: @marin-ma Let's check if we end up getting a `MultiRange` or `nullptr` we bail out the merging here, this will fix the error.  And @hdikeman will work on his change to improve `makeOrFilter` to generate efficient filters (no `MultiRange` for bigints) later.

marin-ma: @Yuhta `ExprToSubfieldFilterParser::makeOrFilter` doesn't return nullptr, and `MultiRange` is not the only returned type of filter for all cases.

Yuhta: It will be changed to return nullptr if we see the merging is not beneficial.  `MultiRange` is known to be slow so we should not accept that as optimization (which is the whole purpose of this function).

Yuhta (approved):

velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -962,6 +962,10 @@ core::TypedExprPtr extractFiltersFromRemainingFilter(
       auto filter =
           exec::ExprToSubfieldFilterParser::makeOrFilter(std::move(disjuncts));
 
+      if (filter == nullptr) {
+        return nullptr;
```
Yuhta: We need to return the original `expr` instead of `nullptr` here

marin-ma: Thanks for catching!

marin-ma: Returning null will affect another code path and result in the ut failure https://github.com/facebookincubator/velox/blob/main/velox/exec/tests/utils/PlanBuilder.cpp#L219-L220

```
[ RUN      ] TableScanTest.statsBasedSkippingWithoutDecompression
/__w/velox/velox/velox/exec/tests/utils/QueryAssertions.cpp:1103: Failure
Failed
Expected 12346, got 31234
18888 extra rows, 0 missing rows
10 of extra rows:
	"com.facebook.presto.orc.stream.01235"
	"com.facebook.presto.orc.stream.01236"
	"com.facebook.presto.orc.stream.01237"
	"com.facebook.presto.orc.stream.01238"
	"com.facebook.presto.orc.stream.01239"
	"com.facebook.presto.orc.stream.01240"
	"com.facebook.presto.orc.stream.01241"
	"com.facebook.presto.orc.stream.01242"
	"com.facebook.presto.orc.stream.01243"
	"com.facebook.presto.orc.stream.01244"

0 of missing rows:

Note: DuckDB only supports timestamps of millisecond precision. If this test involves timestamp inputs, please make sure you use the right precision.
DuckDB query: SELECT * FROM tmp WHERE c0 <= 'com.facebook.presto.orc.stream.01234' or c0 >= 'com.facebook.presto.orc.stream.20123'
/__w/velox/velox/velox/exec/tests/TableScanTest.cpp:2529: Failure
Expected equality of these values:
  size - 10'000
    Which is: 21234
  getTableScanStats(task).rawInputRows
    Which is: 31234
```

Yuhta: We need this change as well:
```diff
diff --git a/fbcode/velox/exec/tests/TableScanTest.cpp b/fbcode/velox/exec/tests/TableScanTest.cpp
--- a/fbcode/velox/exec/tests/TableScanTest.cpp
+++ b/fbcode/velox/exec/tests/TableScanTest.cpp
@@ -2515,7 +2515,7 @@
   auto assertQuery = [&](const std::string& filter) {
     auto rowType = asRowType(rowVector->type());
     return TableScanTest::assertQuery(
-        PlanBuilder(pool_.get()).tableScan(rowType, {filter}).planNode(),
+        PlanBuilder(pool_.get()).tableScan(rowType, {}, filter).planNode(),
         filePaths,
         "SELECT * FROM tmp WHERE " + filter);
   };
diff --git a/fbcode/velox/exec/tests/utils/PlanBuilder.cpp b/fbcode/velox/exec/tests/utils/PlanBuilder.cpp
--- a/fbcode/velox/exec/tests/utils/PlanBuilder.cpp
+++ b/fbcode/velox/exec/tests/utils/PlanBuilder.cpp
@@ -214,10 +214,10 @@
       auto left = toSubfieldFilter(call->inputs()[0], evaluator);
       auto right = toSubfieldFilter(call->inputs()[1], evaluator);
       VELOX_CHECK(left.first == right.first);
-      return {
-          std::move(left.first),
-          exec::ExprToSubfieldFilterParser::makeOrFilter(
-              std::move(left.second), std::move(right.second))};
+      auto filter = exec::ExprToSubfieldFilterParser::makeOrFilter(
+          std::move(left.second), std::move(right.second));
+      VELOX_CHECK_NOT_NULL(filter);
+      return {std::move(left.first), std::move(filter)};
     }

     if (call->name() == "not") {
```

PR #15733, opened 2025-12-09
fix: Skip `switch` function in Spark expression fuzzer

abhinavmuk04:
Summary:
The Spark expression fuzzer was failing intermittently due to the `like` function within the SWITCH function generating invalid escape strings (multi-character instead of single character). 

This diff adds `"switch"` to `skipFunctions` in `SparkExpressionFuzzerTest.cpp` to prevent these spurious failures until the underlying issue is resolved. Here we also excluded 'if' from Presto and Spark fuzzer.

Differential Revision: D88753778

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -442,7 +442,7 @@ jobs:
                 --velox_fuzzer_enable_expression_reuse \
                 --max_expression_trees_per_step 2 \
                 --retry_with_try \
-                --special_forms="and,or,cast,coalesce,if" \
+                --special_forms="and,or,cast,coalesce" \
                 --enable_dereference \
                 --duration_sec $DURATION \
                 --minloglevel=0 \
@@ -706,6 +706,7 @@ jobs:
                 --retry_with_try \
                 --enable_dereference \
                 --velox_fuzzer_enable_decimal_type \
+                --special_forms="and,or,cast,coalesce" \
                 --duration_sec $DURATION \
                 --minloglevel=0 \
                 --stderrthreshold=2 \


kevinwilfong:
I think the problem is with SWITCH, rather than like. There were some rewrites that were recently added to the expression optimizer that cause switch to behave inconsistently between common and simple eval.  SWITCH was recently removed from the Presto Fuzzer because of this

https://github.com/facebookincubator/velox/pull/15602

Looking at the CI result that prompted this, it looks like it was using the LIKE inside a SWITCH statement with a NULL constant, so it's likely the root cause

https://github.com/facebookincubator/velox/actions/runs/20045249887/job/57494090337?fbclid=IwY2xjawOlM59leHRuA2FlbQIxMQBicmlkETFzWU5rR0s5eU5xYWpkMEpHc3J0YwZhcHBfaWQBMAABHhHUqSpblqaj2kRLxz02h46TFxG3EknGP2Q70OWxJQWgnn1Cgbxg1vEetZlJ_aem_IVqYf6ZCBJNK4jGApSEQjg

kevinwilfong:
I don't think the issue in Presto is relevant, this code doesn't use Presto at all, and Velox should consistently throw an exception or not, or at least it would if not for that expression rewrite.

kevinwilfong (approved):
Thanks!

kevinwilfong:
heads up @peterenescu

kevinwilfong:
you would need to add

--special_forms="and,or,cast,coalesce,if" \

here

https://github.com/facebookincubator/velox/blob/528da3f8b39b5d3c36747b5d3afc45db00c9a126/.github/workflows/scheduled.yml#L692-L706

Similar to what was done for Presto in the PR I linked earlier

kevinwilfong (approved):
Looks good, could you update the title and description?

PR #15734, opened 2025-12-09
feat: Coercions for complex and generic types

### LINKED ISSUE #15749: Strange behavior with Unknown coercion in binding signatures
### Problem

Let's look at test:
```C++
TEST_F(NullableArrayViewTest, materialize) {
  auto result = evaluate(
      "array_constructor(1, 2, NULL, 4, NULL)",
      makeRowVector({makeFlatVector<int64_t>(1)}));

  DecodedVector decoded;
  exec::VectorReader<Array<int64_t>> reader(decode(decoded, *result.get()));

  ASSERT_EQ(
      reader[0].materialize(),
      std::vector<std::optional<int64_t>>(
          {1, 2, std::nullopt, 4, std::nullopt}));
}
```

It internally calls `SignatureBinder::tryBind()` without coercions. `array_constructor` has generic type `T` that is bound into corresponding integer type, meanwhile nulls (of type `Unknown`) are passed as actual arguments.
But the following test is failing:
```C++
TEST_F(NullableArrayViewTest, materializeStrange) {
  auto result = evaluate(
      "array_constructor(NULL, NULL, NULL, NULL, 4)",
      makeRowVector({makeFlatVector<int64_t>(1)}));

  DecodedVector decoded;
  exec::VectorReader<Array<int64_t>> reader(decode(decoded, *result.get()));

  ASSERT_EQ(
      reader[0].materialize(),
      std::vector<std::optional<int64_t>>(
          {std::nullopt, std::nullopt, std::nullopt, std::nullopt, 4}));
}
```
with the following reason:
```
Scalar function signature is not supported: array_constructor(UNKNOWN, UNKNOWN, UNKNOWN, UNKNOWN, BIGINT). Supported signatures: () -> array(unknown), (T...) -> array(T).
```

###  Expected behavior
Both tests launch status should be the same: fail or OK. For now it's inconsistent. I suppose both should be failed as calling `tryBind()` **without** coercions should coerce `Unknown` to arbitrary type.

### System information

```
Velox System Info v0.0.2
Commit: a8bfaca0adb3b0ba984395b35f5b4a8222881b48
CMake Version: 4.1.0
System: Linux-6.8.0-87-generic
Arch: x86_64
C++ Compiler: /usr/local/bin/c++
C++ Compiler Version: 20.1.8
20.1.8
C Compiler: /usr/local/bin/cc
C Compiler Version: 20.1.8
```

### Relevant logs

```bash

```

MBkkt:
Add support for coercions involving complex types (types with type parameters) and generic types. 
Previously, coercions were only supported for scalar types without parameters. 

### Complex types

Add coercion for type parameters. This enables queries like this where a function taking two arrays of the same type (array_intersect) is called with array(integer) and array(bigint). The argument of type array(integer) is automatically casted to array(bigint).

```
SQL> explain (type logical) select array_intersect(ARRAY[1, 2, 3], ARRAY[CAST(2 AS BIGINT)]);
- Project: -> ROW<expr:ARRAY<BIGINT>>
    expr := array_intersect(CAST(array_constructor(1, 2, 3) AS ARRAY<BIGINT>), array_constructor(CAST(2 AS BIGINT)))
  - Values: 1 rows -> ROW<>
```

Only case that doesn't work is cast unknown to complex type, e.g.
```
SQL> explain (type logical) select array_intersect(null, ARRAY[CAST(2 AS BIGINT)]);
Query failed: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Scalar function signature is not supported: array_intersect(UNKNOWN, ARRAY<BIGINT>).
```

It doesn't work because it's unclear what to do with type parameters, we only have their signature, we don't have actual parameters. To be able produce such implicit case we need to know how to convert parameter from signature to real one without actual type. I know fix for type parameters, but it's ugly and I don't know fix for non-type parameters (needs somehow find default value for them). So I think better to solve this separately.

### Generic types

Add a loop to find common super type for all arguments, then coerce each argument to the common super type. In the following example, there are 4 arguments with types UNKNOWN, INTEGER, UNKNOWN, DOUBLE. The common super type is DOUBLE, so all arguments are casted to DOUBLE.

```
SQL> explain (type logical) select ARRAY[null, 1, null, 2.0];
- Project: -> ROW<expr:ARRAY<DOUBLE>>
    expr := array_constructor(CAST(null AS DOUBLE), CAST(1 AS DOUBLE), CAST(null AS DOUBLE), 2)
  - Values: 1 rows -> ROW<>
```

### Uknown type

Add coercion from UNKNOWN to any type without parameters.

```
SQL> explain (type logical) select null + 1;
- Project: -> ROW<expr:INTEGER>
    expr := plus(CAST(null AS INTEGER), 1)
  - Values: 1 rows -> ROW<>
```

This also works with generic function signatures.

```
SQL> explain (type logical) select null = ARRAY[CAST(2 AS BIGINT)];
- Project: -> ROW<expr:BOOLEAN>
    expr := eq(CAST(null AS ARRAY<BIGINT>), array_constructor(CAST(2 AS BIGINT)))
  - Values: 1 rows -> ROW<>
```

### Precedence rules

Change cost of coercion to prefer generic function without coercions over a function with coercions.

### References
* [Data type precedence (Transact-SQL)](https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-type-precedence-transact-sql?view=sql-server-ver17)
* [Postgres: Type Conversion](https://www.postgresql.org/docs/7.3/typeconv.html) 

### Notes
- @mkornaukhov  came across some oddities while working on this PR, please check #15749 (they're related to case when we don't have coercions)

### Follow ups

1. Support implicit casts from unknown to complex types
2. Support implicit casts between types with non-type parameters, e.g. decimal
3. Fix #15749 and `isHomogeneousRow`
4. Support implicit cast registration
5. Support pseudo-types, it's needed for Postgres frontend in SereneDB. Pseudo-types have implicit casts to some types, but if it's generic it's still needed to be cast to some type (e.g. it's varchar for Postgres unknown (type of literal))
6. Optimize pre-existing `O(n^2)` in Switch/If/Coalesce expressions.
7. Find why actual type can be null in some tests: https://github.com/facebookincubator/velox/pull/15734#discussion_r2616567872

### DIFF
--- a/velox/core/SimpleFunctionMetadata.h
+++ b/velox/core/SimpleFunctionMetadata.h
@@ -20,7 +20,6 @@
 
 #include "velox/common/base/Exceptions.h"
 #include "velox/common/base/Status.h"
-#include "velox/core/CoreTypeSystem.h"
 #include "velox/core/Metaprogramming.h"
 #include "velox/core/QueryConfig.h"
 #include "velox/expression/FunctionSignature.h"
@@ -29,6 +28,12 @@
 #include "velox/type/Type.h"
 
 namespace facebook::velox::core {
+
+// This assumes we wont have signature longer than 1M argument.
+inline constexpr CallableCost kMaxFunctionArgs = 1'000'000;
+
+// This assumes we wont have function rank number greater than 4.
+inline constexpr CallableCost kMaxFunctionRank = 4;
 
 // Most UDFs are deterministic, hence this default value.
 template <class T, class = void>
@@ -141,7 +146,7 @@ struct TypeAnalysisResults {
     // in this case (1) is picked.
     // e.g: (Any, int) will be picked before (Any, Any)
     // e.g: Variadic<Array<Any>> is picked before Variadic<Any>.
-    uint32_t getRank() {
+    CallableCost getRank() const {
       if (!hasGeneric && !hasVariadic) {
         return 1;
       }
@@ -164,18 +169,20 @@ struct TypeAnalysisResults {
       VELOX_UNREACHABLE("unreachable");
     }
 
-    uint32_t computePriority() {
-      // This assumes we wont have signature longer than 1M argument.
-      return getRank() * 1000000 - concreteCount;
+    CallableCost computePriority() const {
+      const CallableCost rank = getRank();
+      VELOX_DCHECK_LE(rank, kMaxFunctionRank);
+      VELOX_DCHECK_LE(concreteCount, kMaxFunctionArgs);
+      return rank * kMaxFunctionArgs - concreteCount;
     }
   } stats;
 
   void addVariable(exec::SignatureVariable&& variable) {
-    if (!variablesInformation.count(variable.name())) {
-      variablesInformation.emplace(variable.name(), variable);
-    } else {
+    auto [it, emplaced] =
+        variablesInformation.try_emplace(variable.name(), variable);
+    if (!emplaced) {
       VELOX_CHECK(
-          variable == variablesInformation.at(variable.name()),
+          variable == it->second,
           "Cant assign different properties to the same variable {}",
           variable.name());
     }
@@ -438,7 +445,7 @@ class ISimpleFunctionMetadata {
   virtual std::string getName() const = 0;
   virtual bool isDeterministic() const = 0;
   virtual bool defaultNullBehavior() const = 0;
-  virtual uint32_t priority() const = 0;
+  virtual CallableCost priority() const = 0;
   virtual const std::shared_ptr<exec::FunctionSignature> signature() const = 0;
   virtual const TypePtr& resultPhysicalType() const = 0;
   virtual const std::vector<TypePtr>& argPhysicalTypes() const = 0;
@@ -533,7 +540,7 @@ class SimpleFunctionMetadata : public ISimpleFunctionMetadata {
     return CreateType<return_type>::create(signature());
   }
 
-  uint32_t priority() const override {
+  CallableCost priority() const override {
     return priority_;
   }
 
@@ -693,7 +700,7 @@ class SimpleFunctionMetadata : public ISimpleFunctionMetadata {
 
   const bool defaultNullBehavior_;
   exec::FunctionSignaturePtr signature_;
-  uint32_t priority_;
+  CallableCost priority_;
   TypePtr resultPhysicalType_;
   std::vector<TypePtr> argPhysicalTypes_;
 };
--- a/velox/expression/SignatureBinder.cpp
+++ b/velox/expression/SignatureBinder.cpp
@@ -108,34 +108,48 @@ bool checkNamedRowField(
     const TypeSignature& signature,
     const TypePtr& actualType,
     size_t idx) {
-  if (signature.rowFieldName().has_value() &&
-      (*signature.rowFieldName() != asRowType(actualType)->nameOf(idx))) {
-    return false;
+  if (!signature.rowFieldName()) {
+    return true;
   }
-  return true;
+  return *signature.rowFieldName() == asRowType(actualType)->nameOf(idx);
 }
 
 } // namespace
 
-bool SignatureBinder::tryBindWithCoercions(std::vector<Coercion>& coercions) {
-  return tryBind(true, coercions);
-}
-
-bool SignatureBinder::tryBind() {
-  std::vector<Coercion> coercions;
-  return tryBind(false, coercions);
-}
-
-bool SignatureBinder::tryBind(
-    bool allowCoercions,
-    std::vector<Coercion>& coercions) {
-  const auto numActualTypes = actualTypes_.size();
-
-  if (allowCoercions) {
-    coercions.clear();
-    coercions.resize(numActualTypes);
+bool SignatureBinder::coercibleToTypeVars(
+    const TypeSignature& signature,
+    const TypePtr& actualType) {
+  if (!actualType) {
+    return true;
   }
 
+  const auto& varName = signature.baseName();
+  if (auto variableIt = variables().find(varName);
+      variableIt != variables().end()) {
+    if (!variableIt->second.isTypeParameter()) {
+      return true;
+    }
+    auto& varType = typeVariablesBindings_[varName];
+    varType = TypeCoercer::leastCommonSuperType(varType, actualType);
+    return varType && variableIt->second.isEligibleType(*varType);
+  }
+
+  const auto& signatureParams = signature.parameters();
+  const auto& actualParams = actualType->parameters();
+
+  const auto paramsCnt = std::min(signatureParams.size(), actualParams.size());
+  for (size_t i = 0; i < paramsCnt; ++i) {
+    if (actualParams[i].kind == TypeParameterKind::kType &&
+        !coercibleToTypeVars(signatureParams[i], actualParams[i].type)) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool SignatureBinder::tryBind(std::vector<Coercion>* coercions) {
+  const auto numActualTypes = actualTypes_.size();
   const auto& formalArgs = signature_.argumentTypes();
   const auto numFormalArgs = formalArgs.size();
 
@@ -143,71 +157,48 @@ bool SignatureBinder::tryBind(
     if (numActualTypes < numFormalArgs - 1) {
       return false;
     }
-  } else {
-    if (numFormalArgs != numActualTypes) {
+    if (!coercions && !isAny(signature_.argumentTypes().back()) &&
+        numActualTypes > numFormalArgs) {
+      const auto& variableType = actualTypes_[numFormalArgs - 1];
+      if (!variableType) {
+        return false;
+      }
+      for (size_t i = numFormalArgs; i < numActualTypes; ++i) {
+        if (!actualTypes_[i] || !actualTypes_[i]->equivalent(*variableType)) {
+          return false;
+        }
+      }
+    }
+  } else if (numFormalArgs != numActualTypes) {
+    return false;
+  }
+
+  size_t numBindArgs = numActualTypes;
+  if (coercions) {
+    coercions->clear();
+    coercions->resize(numBindArgs);
+
+    // Phase 1: find least common var types
+    for (size_t i = 0; i < numActualTypes; ++i) {
+      const auto& formalArgSignature =
+          i < numFormalArgs ? formalArgs[i] : formalArgs.back();
+      if (!coercibleToTypeVars(formalArgSignature, actualTypes_[i])) {
+        return false;
+      }
+    }
+  } else if (numFormalArgs < numBindArgs) {
+    numBindArgs = numFormalArgs;
+  }
+
+  // Phase 2: check that bound types are coercible
+  for (size_t i = 0; i < numBindArgs; ++i) {
+    const auto& formalArgSignature =
+        i < numFormalArgs ? formalArgs[i] : formalArgs.back();
+    if (!SignatureBinderBase::tryBind(
+            formalArgSignature,
+            actualTypes_[i],
+            coercions ? &(*coercions)[i] : nullptr)) {
       return false;
-    }
-  }
-
-  if (allowCoercions && !variables().empty()) {
-    for (auto i = 0; i < numActualTypes; i++) {
-      if (actualTypes_[i]) {
-        const auto& typeSignature =
-            i < numFormalArgs ? formalArgs[i] : formalArgs[numFormalArgs - 1];
-
-        if (!tryBindVariablesWithCoercion(typeSignature, actualTypes_[i])) {
-          return false;
-        }
-      }
-    }
-  }
-
-  for (auto i = 0; i < numFormalArgs && i < numActualTypes; i++) {
-    if (actualTypes_[i]) {
-      if (allowCoercions) {
-        if (!SignatureBinderBase::tryBindWithCoercion(
-                formalArgs[i], actualTypes_[i], coercions[i])) {
-          return false;
-        }
-      } else {
-        if (!SignatureBinderBase::tryBind(formalArgs[i], actualTypes_[i])) {
-          return false;
-        }
-      }
-    } else {
-      return false;
-    }
-  }
-
-  if (signature_.variableArity()) {
-    if (!isAny(signature_.argumentTypes().back())) {
-      if (numActualTypes > numFormalArgs) {
-        if (allowCoercions) {
-          auto firstType = actualTypes_[numFormalArgs - 1];
-          if (coercions[numFormalArgs - 1].type != nullptr) {
-            firstType = coercions[numFormalArgs - 1].type;
-          }
-
-          for (auto i = numFormalArgs; i < numActualTypes; i++) {
-            if (auto cost =
-                    TypeCoercer::coercible(actualTypes_[i], firstType)) {
-              if (cost.value() > 0) {
-                coercions[i] = Coercion{firstType, cost.value()};
-              }
-            } else {
-              return false;
-            }
-          }
-
-        } else {
-          const auto& firstType = actualTypes_[numFormalArgs - 1];
-          for (auto i = numFormalArgs; i < numActualTypes; i++) {
-            if (!firstType->equivalent(*actualTypes_[i])) {
-              return false;
-            }
-          }
-        }
-      }
     }
   }
 
@@ -270,153 +261,53 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
   return true;
 }
 
-std::optional<bool> SignatureBinderBase::checkSetTypeVariable(
-    const exec::TypeSignature& typeSignature,
-    const TypePtr& actualType,
-    bool allowCoercion,
-    Coercion& coercion) {
-  const auto& baseName = typeSignature.baseName();
-
-  auto variableIt = variables().find(baseName);
-  if (variableIt == variables().end()) {
-    return std::nullopt;
-  }
-
-  if (allowCoercion) {
-    // Variables must be already set.
-    auto bindingIt = typeVariablesBindings_.find(baseName);
-    VELOX_CHECK(bindingIt != typeVariablesBindings_.end());
-
-    const auto& boundType = bindingIt->second;
-    const auto cost = TypeCoercer::coercible(actualType, boundType);
-    VELOX_CHECK(cost.has_value());
-
-    if (cost.value() > 0) {
-      coercion.type = boundType;
-      coercion.cost = cost.value();
-    }
-    return true;
-  }
-
-  // Variables cannot have further parameters.
-  VELOX_CHECK(
-      typeSignature.parameters().empty(),
-      "Variables with parameters are not supported");
-  const auto& variable = variableIt->second;
-  VELOX_CHECK(variable.isTypeParameter(), "Not expecting integer variable");
-
-  auto bindingIt = typeVariablesBindings_.find(baseName);
-  if (bindingIt != typeVariablesBindings_.end()) {
-    return bindingIt->second->equivalent(*actualType);
-  }
-
-  if (!variable.isEligibleType(*actualType)) {
-    return false;
-  }
-
-  typeVariablesBindings_[baseName] = actualType;
-  return true;
-}
-
-bool SignatureBinderBase::tryBindWithCoercion(
-    const exec::TypeSignature& typeSignature,
-    const TypePtr& actualType,
-    Coercion& coercion) {
-  return tryBind(typeSignature, actualType, true, coercion);
-}
-
-bool SignatureBinderBase::tryBind(
-    const exec::TypeSignature& typeSignature,
-    const TypePtr& actualType) {
-  Coercion coercion;
-  return tryBind(typeSignature, actualType, false, coercion);
-}
-
-bool SignatureBinder::tryBindVariablesWithCoercion(
-    const exec::TypeSignature& typeSignature,
-    const TypePtr& actualType) {
-  const auto& baseName = typeSignature.baseName();
-
-  auto variableIt = variables().find(baseName);
-  if (variableIt != variables().end()) {
-    // Variables cannot have further parameters.
-    VELOX_CHECK(
-        typeSignature.parameters().empty(),
-        "Variables with parameters are not supported");
-    const auto& variable = variableIt->second;
-    VELOX_CHECK(variable.isTypeParameter(), "Not expecting integer variable");
-
-    if (!variable.isEligibleType(*actualType)) {
-      return false;
-    }
-
-    auto bindingIt = typeVariablesBindings_.find(baseName);
-    if (bindingIt == typeVariablesBindings_.end()) {
-      typeVariablesBindings_[baseName] = actualType;
-      return true;
-    }
-
-    if (auto superType =
-            TypeCoercer::leastCommonSuperType(actualType, bindingIt->second)) {
-      typeVariablesBindings_[baseName] = superType;
-      return true;
-    }
-
-    return false;
-  }
-
-  if (typeSignature.isHomogeneousRow()) {
-    // TODO Add coercion support.
-    return true;
-  }
-
-  const auto& params = typeSignature.parameters();
-  if (params.size() != actualType->parameters().size()) {
-    return false;
-  }
-
-  for (auto i = 0; i < params.size(); i++) {
-    const auto& actualParameter = actualType->parameters()[i];
-    if (actualParameter.kind == TypeParameterKind::kType) {
-      if (!tryBindVariablesWithCoercion(params[i], actualParameter.type)) {
-        return false;
-      }
-    }
-  }
-
-  return true;
-}
-
 bool SignatureBinderBase::tryBind(
     const exec::TypeSignature& typeSignature,
     const TypePtr& actualType,
-    bool allowCoercion,
-    Coercion& coercion) {
-  coercion.reset();
+    Coercion* coercion) {
+  if (!actualType) {
+    return false;
+  }
+
   if (isAny(typeSignature)) {
     return true;
   }
 
-  if (auto result = checkSetTypeVariable(
-          typeSignature, actualType, allowCoercion, coercion)) {
-    return result.value();
-  }
-
-  // Type is not a variable.
   const auto& baseName = typeSignature.baseName();
-  auto typeName = boost::algorithm::to_upper_copy(baseName);
-  if (!boost::algorithm::iequals(typeName, actualType->name())) {
-    if (allowCoercion) {
-      if (auto availableCoercion =
-              TypeCoercer::coerceTypeBase(actualType, typeName)) {
-        coercion = availableCoercion.value();
-        return true;
+
+  if (auto variableIt = variables().find(baseName);
+      variableIt != variables().end()) {
+    VELOX_CHECK(
+        typeSignature.parameters().empty(),
+        "Variables with parameters are not supported");
+    VELOX_CHECK(variableIt->second.isTypeParameter());
+
+    auto& varType = typeVariablesBindings_[variableIt->second.name()];
+    if (!varType) {
+      if (!variableIt->second.isEligibleType(*actualType)) {
+        return false;
       }
+      varType = actualType;
     }
-    return false;
+    if (!coercion) {
+      return varType->equivalent(*actualType);
+    }
+    *coercion = TypeCoercer::coercible(actualType, varType);
+    return coercion->isPossible();
   }
 
   const auto& params = typeSignature.parameters();
+
+  // Type is not a variable.
+  auto typeName = boost::algorithm::to_upper_copy(baseName);
+  if (!boost::algorithm::iequals(typeName, actualType->name())) {
+    if (!coercion || !params.empty()) {
+      return false;
+    }
+    // TODO: Handle coercion from unknown to complex types with parameters.
+    *coercion = TypeCoercer::coerceTypeBase(actualType, typeName);
+    return coercion->isPossible();
+  }
 
   // Handle homogeneous row case: row(T, ...)
   if (typeSignature.isHomogeneousRow()) {
@@ -443,9 +334,18 @@ bool SignatureBinderBase::tryBind(
     }
 
     // TODO Add coercion support.
-    if (auto result = checkSetTypeVariable(
-            typeParam, actualChildType, /*allowCoercion=*/false, coercion)) {
-      return result.value();
+    const auto& paramBaseName = typeParam.baseName();
+    if (auto variableIt = variables().find(paramBaseName);
+        variableIt != variables().end()) {
+      auto& varType = typeVariablesBindings_[paramBaseName];
+      if (varType) {
+        return varType->equivalent(*actualChildType);
+      }
+      if (!variableIt->second.isEligibleType(*actualType)) {
+        return false;
+      }
+      varType = actualChildType;
+      return true;
     }
 
     return tryBind(typeParam, actualChildType);
@@ -456,12 +356,11 @@ bool SignatureBinderBase::tryBind(
     return false;
   }
 
-  std::vector<Coercion> paramCoercions;
-  if (allowCoercion) {
-    paramCoercions.resize(params.size());
-  }
+  CallableCost totalCost = 0;
+  std::vector<TypeParameter> newParameters;
+  newParameters.reserve(params.size());
 
-  for (auto i = 0; i < params.size(); i++) {
+  for (size_t i = 0; i < params.size(); i++) {
     const auto& actualParameter = actualType->parameters()[i];
     switch (actualParameter.kind) {
       case TypeParameterKind::kLongLiteral:
@@ -484,47 +383,37 @@ bool SignatureBinderBase::tryBind(
           return false;
         }
         break;
-      case TypeParameterKind::kType:
+      case TypeParameterKind::kType: {
         if (!checkNamedRowField(params[i], actualType, i)) {
           return false;
         }
-
-        if (allowCoercion) {
-          if (!tryBindWithCoercion(
-                  params[i], actualParameter.type, paramCoercions[i])) {
-            return false;
-          }
-
-        } else if (!tryBind(params[i], actualParameter.type)) {
+        Coercion parameterCoercion;
+        if (!tryBind(
+                params[i],
+                actualParameter.type,
+                coercion ? &parameterCoercion : nullptr)) {
           return false;
         }
-
-        break;
-    }
-  }
-
-  if (allowCoercion) {
-    const bool hasCoercion = std::ranges::any_of(
-        paramCoercions,
-        [](const auto& coercion) { return coercion.type != nullptr; });
-
-    if (hasCoercion) {
-      std::vector<TypeParameter> newParams;
-      newParams.reserve(params.size());
-      for (auto i = 0; i < params.size(); i++) {
-        if (paramCoercions[i].type != nullptr) {
-          newParams.push_back(
-              TypeParameter(paramCoercions[i].type, params[i].rowFieldName()));
-          coercion.cost += paramCoercions[i].cost;
-        } else {
-          newParams.push_back(actualType->parameters()[i]);
+        if (parameterCoercion.cost > 0) {
+          newParameters.emplace_back(
+              std::move(parameterCoercion.type), params[i].rowFieldName());
+          totalCost += parameterCoercion.cost;
+          continue;
         }
-      }
-
-      coercion.type = getType(typeName, newParams);
+      } break;
+    }
+    if (coercion) {
+      newParameters.emplace_back(actualParameter);
     }
   }
 
+  if (totalCost == 0) {
+    return true;
+  }
+  VELOX_DCHECK_NOT_NULL(coercion);
+  coercion->type = getType(typeName, newParameters);
+  coercion->cost = totalCost;
+  VELOX_DCHECK(!actualType->equivalent(*coercion->type));
   return true;
 }
 
--- a/velox/expression/SignatureBinder.h
+++ b/velox/expression/SignatureBinder.h
@@ -29,20 +29,15 @@ class SignatureBinderBase {
   /// Return true if actualType can bind to typeSignature and update bindings_
   /// accordingly. The number of parameters in typeSignature and actualType
   /// must match. Return false otherwise.
+  /// Non-null 'coercion' allows implicit type conversion if actualType
+  /// doesn't match typeSignature exactly.
+  /// @param coercion Type coercion necessary to bind actualType to
+  /// typeSignature if there is no exact match. 'coercion->type' is null if
+  /// there is exact match.
   bool tryBind(
       const exec::TypeSignature& typeSignature,
-      const TypePtr& actualType);
-
-  /// Like 'tryBind', but allows implicit type conversion if actualType
-  /// doesn't match typeSignature exactly.
-  ///
-  /// @param coercion Type coercion necessary to bind actualType to
-  /// typeSignature if there is no exact match. 'coercion.type' is null if
-  /// there is exact match.
-  bool tryBindWithCoercion(
-      const exec::TypeSignature& typeSignature,
       const TypePtr& actualType,
-      Coercion& coercion);
+      Coercion* coercion = nullptr);
 
   // Return the variables of the signature.
   auto& variables() const {
@@ -80,22 +75,10 @@ class SignatureBinderBase {
       const std::string& parameterName,
       const VarcharEnumParameter& params);
 
-  std::optional<bool> checkSetTypeVariable(
-      const exec::TypeSignature& typeSignature,
-      const TypePtr& actualType,
-      bool allowCoercion,
-      Coercion& coercion);
-
   /// Try to bind the integer parameter from the actualType.
   bool tryBindIntegerParameters(
       const std::vector<exec::TypeSignature>& parameters,
       const TypePtr& actualType);
-
-  bool tryBind(
-      const exec::TypeSignature& typeSignature,
-      const TypePtr& actualType,
-      bool allowCoercion,
-      Coercion& coercion);
 };
 
 /// Resolves generic type names in the function signature using actual input
@@ -115,14 +98,12 @@ class SignatureBinder : private SignatureBinderBase {
       : SignatureBinderBase{signature}, actualTypes_{actualTypes} {}
 
   /// Returns true if successfully resolved all generic type names.
-  bool tryBind();
-
-  /// Like 'tryBind', but allows implicit type conversion if actualTypes don't
+  /// Allows implicit type conversion if actualTypes don't
   /// match the signature exactly.
   /// @param coercions Type coercions necessary to bind actualTypes to the
   /// signature. There is one entry per argument. Coercion.type is null if no
   /// coercion is required for that argument.
-  bool tryBindWithCoercions(std::vector<Coercion>& coercions);
+  bool tryBind(std::vector<Coercion>* coercions = nullptr);
 
   /// Returns concrete return type or nullptr if couldn't fully resolve.
   TypePtr tryResolveReturnType() {
@@ -190,10 +171,12 @@ class SignatureBinder : private SignatureBinderBase {
           varcharEnumVariablesBindings);
 
  private:
-  bool tryBind(bool allowCoercions, std::vector<Coercion>& coercions);
-
-  bool tryBindVariablesWithCoercion(
-      const exec::TypeSignature& typeSignature,
+  // Traverse all types in signatures and deduce the least common type for each
+  // type parameter. Recursion is needed to traverse complex types like
+  // Array<Array<T>>, etc.
+  // Returns false iff some type parameters cannot be bound.
+  bool coercibleToTypeVars(
+      const TypeSignature& signature,
       const TypePtr& actualType);
 
   const std::vector<TypePtr>& actualTypes_;
--- a/velox/expression/SimpleFunctionRegistry.cpp
+++ b/velox/expression/SimpleFunctionRegistry.cpp
@@ -172,98 +172,88 @@ std::optional<SimpleFunctionRegistry::ResolvedSimpleFunction>
 SimpleFunctionRegistry::resolveFunction(
     const std::string& name,
     const std::vector<TypePtr>& argTypes,
-    bool allowCoercion,
-    std::vector<TypePtr>& coercions) const {
+    std::vector<TypePtr>* coercions) const {
   using Candidate = std::pair<const FunctionEntry*, TypePtr>;
-
-  std::optional<Candidate> selectedCandidate;
+  const FunctionEntry* selectedFunction{};
+  TypePtr selectedType;
+  auto selectedPriority = kImpossibleCoercionCost;
+  std::vector<Coercion> selectedCoercions;
+  size_t selectedCount = 0;
   registeredFunctions_.withRLock([&](const auto& map) {
-    if (const auto* signatureMap = getSignatureMap(name, map)) {
-      std::vector<std::pair<std::vector<Coercion>, Candidate>> candidates;
-      std::optional<uint32_t> priority;
+    const auto* signatureMap = getSignatureMap(name, map);
+    if (!signatureMap) {
+      return;
+    }
+    std::vector<Coercion> requiredCoercions;
 
-      for (const auto& [candidateSignature, functionEntry] : *signatureMap) {
-        SignatureBinder binder(candidateSignature, argTypes);
+    for (const auto& [candidateSignature, functionEntry] : *signatureMap) {
+      SignatureBinder binder(candidateSignature, argTypes);
 
-        std::vector<Coercion> requiredCoercions;
-
-        bool bound;
-        if (allowCoercion) {
-          bound = binder.tryBindWithCoercions(requiredCoercions);
-        } else {
-          bound = binder.tryBind();
-        }
-
-        if (bound) {
-          for (const auto& currentCandidate : functionEntry) {
-            const auto& m = currentCandidate->getMetadata();
-
-            // For variadic signatures, number of arguments in function call
-            // may be one less than number of arguments in the signature.
-            const auto numArgsToMatch =
-                std::min(argTypes.size(), m.argPhysicalTypes().size());
-
-            bool match = true;
-            for (auto i = 0; i < numArgsToMatch; ++i) {
-              auto argType = argTypes[i];
-              if (allowCoercion && requiredCoercions[i].type != nullptr) {
-                argType = requiredCoercions[i].type;
-              }
-              if (!physicalTypeMatches(argType, m.argPhysicalTypes()[i])) {
-                match = false;
-                break;
-              }
-            }
-
-            if (!match) {
-              continue;
-            }
-
-            auto resultType = binder.tryResolveReturnType();
-            VELOX_CHECK_NOT_NULL(resultType);
-
-            if (physicalTypeMatches(resultType, m.resultPhysicalType())) {
-              const auto currentPriority = m.priority();
-
-              if (!priority.has_value() || currentPriority < priority.value()) {
-                candidates.clear();
-                candidates.emplace_back(
-                    requiredCoercions,
-                    std::make_pair(currentCandidate.get(), resultType));
-                priority = currentPriority;
-              } else if (allowCoercion && currentPriority == priority.value()) {
-                candidates.emplace_back(
-                    requiredCoercions,
-                    std::make_pair(currentCandidate.get(), resultType));
-              }
-            }
-          }
-        }
+      if (!binder.tryBind(coercions ? &requiredCoercions : nullptr)) {
+        continue;
       }
+      for (const auto& currentCandidate : functionEntry) {
+        const auto& m = currentCandidate->getMetadata();
 
-      if (!candidates.empty()) {
-        if (allowCoercion) {
-          if (auto index = Coercion::pickLowestCost(candidates)) {
-            selectedCandidate = candidates[index.value()].second;
+        // For variadic signatures, number of arguments in function call
+        // may be one less than number of arguments in the signature.
+        const auto numArgsToMatch =
+            std::min(argTypes.size(), m.argPhysicalTypes().size());
 
-            const auto& selectedCoercions = candidates[index.value()].first;
-
-            coercions.clear();
-            for (const auto& coercion : selectedCoercions) {
-              coercions.push_back(coercion.type);
-            }
+        bool match = true;
+        for (auto i = 0; i < numArgsToMatch; ++i) {
+          auto argType = argTypes[i];
+          if (coercions && requiredCoercions[i].type != nullptr) {
+            argType = requiredCoercions[i].type;
           }
-        } else {
-          selectedCandidate = candidates[0].second;
+          if (!physicalTypeMatches(argType, m.argPhysicalTypes()[i])) {
+            match = false;
+            break;
+          }
+        }
+
+        if (!match) {
+          continue;
+        }
+
+        auto resultType = binder.tryResolveReturnType();
+        if (!resultType) {
+          continue;
+        }
+
+        if (physicalTypeMatches(resultType, m.resultPhysicalType())) {
+          // Coercion cost is required to match more suitable signature.
+          // For example, let's assume function has two signatures:
+          // 1. BIGINT -> BIGINT
+          // 2. Generic<T> -> BIGINT
+          // Then binding function to INTEGER should pick the second signature.
+          // So we need to ensure that any non-zero coercion cost is higher than
+          // the maximum possible cost of function signatures without coercion.
+          const auto coercionCost = Coercion::overallCost(requiredCoercions);
+          const auto currentPriority = m.priority() +
+              core::kMaxFunctionRank * core::kMaxFunctionArgs * coercionCost;
+
+          if (currentPriority < selectedPriority) {
+            selectedFunction = currentCandidate.get();
+            selectedType = resultType;
+            selectedPriority = currentPriority;
+            selectedCoercions = requiredCoercions;
+            selectedCount = 1;
+          } else if (currentPriority == selectedPriority) {
+            ++selectedCount;
+          }
         }
       }
     }
   });
-
-  return selectedCandidate.has_value()
-      ? std::make_optional(ResolvedSimpleFunction(
-            *(selectedCandidate->first), selectedCandidate->second))
-      : std::nullopt;
+  if (selectedCount != 1) {
+    if (coercions) {
+      coercions->clear();
+    }
+    return std::nullopt;
+  }
+  Coercion::convert(selectedCoercions, coercions);
+  return ResolvedSimpleFunction{*selectedFunction, selectedType};
 }
 
 } // namespace facebook::velox::exec
--- a/velox/expression/SimpleFunctionRegistry.h
+++ b/velox/expression/SimpleFunctionRegistry.h
@@ -156,23 +156,21 @@ class SimpleFunctionRegistry {
   std::optional<ResolvedSimpleFunction> resolveFunction(
       const std::string& name,
       const std::vector<TypePtr>& argTypes) const {
-    std::vector<TypePtr> coercions;
-    return resolveFunction(name, argTypes, false, coercions);
+    return resolveFunction(name, argTypes, nullptr);
   }
 
   std::optional<ResolvedSimpleFunction> resolveFunctionWithCoercions(
       const std::string& name,
       const std::vector<TypePtr>& argTypes,
       std::vector<TypePtr>& coercions) const {
-    return resolveFunction(name, argTypes, true, coercions);
+    return resolveFunction(name, argTypes, &coercions);
   }
 
  private:
   std::optional<ResolvedSimpleFunction> resolveFunction(
       const std::string& name,
       const std::vector<TypePtr>& argTypes,
-      bool allowCoercion,
-      std::vector<TypePtr>& coercions) const;
+      std::vector<TypePtr>* coercions) const;
 
   /// Registers a function with the given name and metadata. If an entry with
   /// the name already exists and 'overwrite' is true, the existing entry is
--- a/velox/expression/VectorFunction.cpp
+++ b/velox/expression/VectorFunction.cpp
@@ -86,56 +86,46 @@ TypePtr resolveVectorFunction(
   return nullptr;
 }
 
-namespace {
-bool hasCoercion(const std::vector<Coercion>& coercions) {
-  for (const auto& coercion : coercions) {
-    if (coercion.type != nullptr) {
-      return true;
-    }
-  }
-
-  return false;
-}
-} // namespace
-
 TypePtr resolveVectorFunctionWithCoercions(
     const std::string& functionName,
     const std::vector<TypePtr>& argTypes,
     std::vector<TypePtr>& coercions) {
-  coercions.clear();
+  std::vector<Coercion> selectedCoercions;
+  size_t selectedCount = 0;
 
   auto optionalType = applyToVectorFunctionEntry<TypePtr>(
-      functionName,
-      [&](const auto& /*name*/, const auto& entry) -> std::optional<TypePtr> {
-        std::vector<std::pair<std::vector<Coercion>, TypePtr>> candidates;
+      functionName, [&](const auto& /*name*/, const auto& entry) {
+        TypePtr selectedType;
+        auto selectedPriority = kImpossibleCoercionCost;
+        std::vector<Coercion> requiredCoercions;
         for (const auto& signature : entry.signatures) {
           exec::SignatureBinder binder(*signature, argTypes);
-          std::vector<Coercion> requiredCoercions;
-          if (binder.tryBindWithCoercions(requiredCoercions)) {
-            auto type = binder.tryResolveReturnType();
-            if (!hasCoercion(requiredCoercions)) {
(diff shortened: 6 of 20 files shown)


velox/type/TypeCoercer.cpp
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 #include "velox/type/TypeCoercer.h"
+#include <velox/type/Type.h>
+#include <optional>
```
MBkkt:
```suggestion
```

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+  return tryBind(nullptr);
+}
+
+bool SignatureBinder::resolveTypeVars(
+    const TypeSignature& signature,
+    const TypePtr& actualType,
+    Coercion* coercion) {
+  const auto& varName = signature.baseName();
+  if (variables().contains(varName)) {
+    auto& variableSignature = variables().at(varName);
```
MBkkt: Please make it single lookup?

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+      }
+      varType = actualType;
+    } else if (!varType->equivalent(*actualType)) {
+      if (!coercion) {
+        return false;
+      }
+      if (auto c = TypeCoercer::coercible(actualType, varType)) {
+        if (coercion) {
+          *coercion = c;
+        }
```
MBkkt:
```suggestion
        *coercion = c;
```

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+      } else if (auto c = TypeCoercer::coercible(varType, actualType)) {
+        if (!checkSignatureProperties(variableSignature, actualType)) {
+          return false;
+        }
```
MBkkt: Make it more flat?

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+    auto& varType = typeVariablesBindings_[varName];
+    if (!varType) {
+      if (!checkSignatureProperties(variableSignature, actualType)) {
+        return false;
+      }
+      varType = actualType;
+    } else if (!varType->equivalent(*actualType)) {
+      if (!coercion) {
+        return false;
+      }
```
MBkkt: make it more flat?

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+    }
+    return true;
+  }
+  const auto& signatureParams = signature.parameters();
+  const auto& actualParams = actualType->parameters();
+  if (signatureParams.empty()) {
+    return true;
+  }
```
MBkkt: Add comment about this case

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
 bool SignatureBinder::tryBind() {
-  std::vector<Coercion> coercions;
-  return tryBind(false, coercions);
+  return tryBind(nullptr);
+}
+
+bool SignatureBinder::resolveTypeVars(
+    const TypeSignature& signature,
+    const TypePtr& actualType,
+    Coercion* coercion) {
```
MBkkt: Maybe bool instead?

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+  return tryBind(&coercions);
 }
 
 bool SignatureBinder::tryBind() {
-  std::vector<Coercion> coercions;
-  return tryBind(false, coercions);
+  return tryBind(nullptr);
+}
+
+bool SignatureBinder::resolveTypeVars(
```
MBkkt: comment about why is it recursive?

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +507,39 @@ bool SignatureBinderBase::tryBind(
...
           return false;
         }
-
-        if (!tryBind(params[i], actualParameter.type)) {
-          // TODO Allow coercions for complex types.
+        Coercion childCoercion;
+        if (!tryBind(params[i], actualParameter.type, &childCoercion)) {
           return false;
         }
+        if (coercion && childCoercion.type != nullptr) {
```
MBkkt:
```suggestion
        if (coercion && childCoercion) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -155,23 +277,47 @@ bool SignatureBinder::tryBind(
...
+      const auto& formalArgSignature =
+          i < formalArgsCnt ? formalArgs[i] : formalArgs.back();
+      const auto& actualType = actualTypes_[i];
+      if (!checkTypeVarsCoercions(
+              formalArgSignature, actualType, &coercions[i])) {
+        return false;
+      }
+      if (coercions[i]) {
+        continue;
+      }
```
MBkkt: lets remove this and check this inside tryBind

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
-
-    if (!isAny(signature_.argumentTypes().back())) {
-      if (actualTypes_.size() > formalArgsCnt) {
-        auto& type = actualTypes_[formalArgsCnt - 1];
-        for (auto i = formalArgsCnt; i < actualTypes_.size(); i++) {
-          if (!type->equivalent(*actualTypes_[i]) &&
-              actualTypes_[i]->kind() != TypeKind::UNKNOWN) {
-            return false;
+    if (!coercionsPtr) {
+      if (!isAny(signature_.argumentTypes().back())) {
```
MBkkt:
```suggestion
      if (!coercionsPtr && !isAny(signature_.argumentTypes().back())) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -111,40 +116,157 @@ bool checkNamedRowField(
...
+  return true;
 }
 
-bool SignatureBinder::tryBind(
-    bool allowCoercions,
-    std::vector<Coercion>& coercions) {
-  if (allowCoercions) {
-    coercions.clear();
-    coercions.resize(actualTypes_.size());
+bool SignatureBinder::tryBind(std::vector<Coercion>* coercionsPtr) {
```
MBkkt: lets rename coercionsPtr to coercions

velox/expression/SignatureBinder.cpp
```diff
@@ -291,10 +422,12 @@ bool SignatureBinderBase::tryBind(
       boost::algorithm::to_upper_copy(std::string(actualType->name()));
 
   if (typeName != actualTypeName) {
-    if (allowCoercion) {
+    if (coercion) {
```
MBkkt:
```suggestion
if (typeName != actualTypeName && coercion) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +507,39 @@ bool SignatureBinderBase::tryBind(
...
+          newParameters.emplace_back(childCoercion.type);
+        } else {
+          newParameters.emplace_back(actualParameter.type);
+        }
         break;
+      }
+    }
+  }
+
+  if (coercion && hasCoercion) {
```
MBkkt: early return?

velox/expression/SignatureBinder.cpp
```diff
@@ -346,6 +479,11 @@ bool SignatureBinderBase::tryBind(
     return false;
   }
 
+  bool hasCoercion = false;
```
MBkkt:
```suggestion
  bool needsCoercion = false;
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
   }
 
+  int32_t cost = 0;
   for (auto i = 0; i < fromType->size(); i++) {
-    if (!coercible(fromType->childAt(i), toType->childAt(i))) {
-      return false;
+    if (auto c = coercible(fromType->childAt(i), toType->childAt(i))) {
+      cost += c.cost;
+    } else {
+      return Coercion(nullptr);
```
MBkkt:
```suggestion
      return {};
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
-    }
-
-    return false;
+    return TypeCoercer::coerceTypeBase(fromType, toType->name());
   }
 
   if (fromType->name() != toType->name() ||
       fromType->size() != toType->size()) {
-    return false;
+    return Coercion{nullptr};
```
MBkkt:
```suggestion
    return {};
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
+Coercion TypeCoercer::coerceTypeBase(
     const TypePtr& fromType,
     const std::string& toTypeName) {
   static const auto kAllowedCoercions = allowedCoercions();
   if (fromType->name() == toTypeName) {
-    return Coercion{.type = fromType, .cost = 0};
+    return Coercion{fromType, 0};
+  }
+  if (fromType == UNKNOWN()) {
+    // Cast Unknwon to complex type is not supported yet
```
MBkkt:
```suggestion
    // Cast Unknown to complex type is not supported yet
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
-      return false;
+    if (auto c = coercible(fromType->childAt(i), toType->childAt(i))) {
+      cost += c.cost;
+    } else {
+      return Coercion(nullptr);
     }
   }
 
-  return true;
+  return Coercion{toType, cost};
```
MBkkt:
```suggestion
  return {std::move(toType), cost};
```

velox/type/TypeCoercer.cpp
MBkkt: assert about different base name

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
 // static
-std::optional<Coercion> TypeCoercer::coerceTypeBase(
+Coercion TypeCoercer::coerceTypeBase(
     const TypePtr& fromType,
     const std::string& toTypeName) {
   static const auto kAllowedCoercions = allowedCoercions();
   if (fromType->name() == toTypeName) {
-    return Coercion{.type = fromType, .cost = 0};
+    return Coercion{fromType, 0};
+  }
```
MBkkt, line 81: DCHECK or CHECK

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
+    return Coercion{getType(toTypeName, {}), 0};
   }
 
   auto it = kAllowedCoercions.find({fromType->name(), toTypeName});
   if (it != kAllowedCoercions.end()) {
     return it->second;
   }
 
-  return std::nullopt;
+  return Coercion{nullptr};
```
MBkkt:
```suggestion
  return {};
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
     const TypePtr& fromType,
     const std::string& toTypeName) {
   static const auto kAllowedCoercions = allowedCoercions();
   if (fromType->name() == toTypeName) {
-    return Coercion{.type = fromType, .cost = 0};
+    return Coercion{fromType, 0};
+  }
+  if (fromType == UNKNOWN()) {
+    // Cast Unknwon to complex type is not supported yet
+    return Coercion{getType(toTypeName, {}), 0};
```
MBkkt:
```suggestion
    return Coercion{getType(toTypeName, {}), 1};
```

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +55,53 @@ allowedCoercions() {
...
 
 // static
-bool TypeCoercer::coercible(const TypePtr& fromType, const TypePtr& toType) {
-  if (fromType->isUnKnown()) {
-    return true;
+Coercion TypeCoercer::coercible(
+    const TypePtr& fromType,
+    const TypePtr& toType) {
+  if (fromType == UNKNOWN()) {
+    return Coercion{toType, 0};
```
MBkkt:
```suggestion
    return Coercion{toType, 1};
```

velox/type/TypeCoercer.cpp
MBkkt: should be `name()`

velox/expression/SignatureBinder.cpp
```diff
@@ -291,10 +422,12 @@ bool SignatureBinderBase::tryBind(
       boost::algorithm::to_upper_copy(std::string(actualType->name()));
```
MBkkt:
```suggestion
const auto& actualTypeName = actualType->name();
```

velox/expression/SignatureBinder.cpp
mkornaukhov: I don't think it's needed, returning value is always true

velox/type/TypeCoercer.cpp
mkornaukhov: toType is const ref

velox/type/TypeCoercer.cpp
mkornaukhov: It may be called in `coercible()` method

mkornaukhov:
I came across some oddities while working on this PR, please check #15749

velox/core/SimpleFunctionMetadata.h
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <folly/Likely.h>
+#include <velox/type/Cost.h>
```
MBkkt:
```suggestion
#include "velox/type/Cost.h"
```
+ should be below

velox/expression/fuzzer/ExpressionFuzzer.cpp
```diff
@@ -53,12 +53,13 @@ class FullSignatureBinder : public SignatureBinderBase {
...
       if (!SignatureBinderBase::tryBind(
-              signature_.argumentTypes()[i], argTypes[i])) {
+              signature_.argumentTypes()[i], argTypes[i], nullptr)) {
         return;
       }
     }
 
-    if (!SignatureBinderBase::tryBind(signature_.returnType(), returnType)) {
+    if (!SignatureBinderBase::tryBind(
+            signature_.returnType(), returnType, nullptr)) {
```
MBkkt: remove this diff

velox/type/Cost.h
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstdint>
+
+namespace facebook::velox {
+
+using CostT = int32_t;
```
MBkkt:
```suggestion
using Cost = int32_t;
```

velox/type/Cost.h
```diff
@@ -0,0 +1,30 @@
...
+
+// This assumes we wont have signature longer than 1M argument.
+constexpr CostT kRankCostStep = 1'000'000;
+
+// This assumes we wont have more ranks than 10.
+constexpr CostT kMinCoercionCost = 10 * kRankCostStep;
```
MBkkt:
```suggestion
inline constexpr CostT kRankCostStep = 1'000'000;

// This assumes we wont have more ranks than 10.
inline constexpr CostT kMinCoercionCost = 10 * kRankCostStep;
```

Also what does "ranks" means?

velox/type/TypeCoercer.cpp
```diff
@@ -35,11 +36,11 @@ allowedCoercions() {
   std::unordered_map<std::pair<std::string, std::string>, Coercion> coercions;
 
   auto add = [&](const TypePtr& from, const std::vector<TypePtr>& to) {
-    int32_t cost = 0;
+    int32_t cost = 1;
```
MBkkt:
```suggestion
    Cost cost = kMinCoercionCost + 1;
```

velox/expression/SimpleFunctionRegistry.cpp
```diff
@@ -223,7 +223,14 @@ SimpleFunctionRegistry::resolveFunction(
             VELOX_CHECK_NOT_NULL(resultType);
 
             if (physicalTypeMatches(resultType, m.resultPhysicalType())) {
-              const auto currentPriority = m.priority();
+              // Coercion cost is required to match more suitable signature. For
+              // example, let's assume function 'foo' has two signatures:
+              // 1. Integer -> Integer
+              // 2. Generic<T> -> Intger
+              // Then binding 'foo'to NULL (of type Unknown) should pick the
+              // second signature.
```
MBkkt:
```suggestion
              // Coercion cost is required to match more suitable signature. For
              // example, let's assume function has two signatures:
              // 1. BIGINT -> BIGINT
              // 2. Generic<T> -> BIGINT
              // Then binding function to INTEGER should pick the
              // second signature.
```

velox/expression/SignatureBinder.h
```diff
@@ -109,13 +98,14 @@ class SignatureBinder : private SignatureBinderBase {
...
-  /// Like 'tryBind', but allows implicit type conversion if actualTypes don't
+  /// Allows implicit type conversion if actualTypes don't
   /// match the signature exactly.
   /// @param coercions Type coercions necessary to bind actualTypes to the
   /// signature. There is one entry per argument. Coercion.type is null if no
   /// coercion is required for that argument.
+  bool tryBind(std::vector<Coercion>* coercions = nullptr);
+
+  /// Wrapper for 'tryBind'
   bool tryBindWithCoercions(std::vector<Coercion>& coercions);
```
MBkkt: Is it still needed?
```suggestion
```

velox/expression/SignatureBinder.cpp
```diff
@@ -337,7 +395,7 @@ bool SignatureBinderBase::tryBind(
         return true;
       }
     } else {
-      return tryBind(typeParam, actualChildType);
+      return tryBind(typeParam, actualChildType, nullptr);
```
MBkkt:
```suggestion
      return tryBind(typeParam, actualChildType);
```

velox/expression/SignatureBinder.cpp
```diff
@@ -346,6 +404,11 @@ bool SignatureBinderBase::tryBind(
     return false;
   }
 
+  bool needsCoercion = false;
+  int32_t totalCost = 0;
```
MBkkt:
```suggestion
  Cost totalCost = 0;
```

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +432,42 @@ bool SignatureBinderBase::tryBind(
...
+          newParameters.emplace_back(childCoercion.type);
+        } else {
+          newParameters.emplace_back(actualParameter.type);
+        }
         break;
+      }
     }
   }
+
+  if (coercion && needsCoercion) {
```
MBkkt:
```suggestion
  if (needsCoercion) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +432,42 @@ bool SignatureBinderBase::tryBind(
...
-        if (!tryBind(params[i], actualParameter.type)) {
-          // TODO Allow coercions for complex types.
+        Coercion childCoercion;
+        if (!tryBind(
+                params[i],
+                actualParameter.type,
+                coercion ? &childCoercion : nullptr)) {
           return false;
         }
+        if (coercion && childCoercion.type) {
```
MBkkt:
```suggestion
        if (childCoercion) {
```

velox/expression/SignatureBinder.cpp
MBkkt: I think it simplify the code

velox/type/Cost.h
mkornaukhov: Rank of precedence all function 'kinds'. See `SimpleFunctionMetadata.h`.

velox/type/TypeCoercer.cpp
mkornaukhov: +1 is not needed I suppose as here is prefix increment is used

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +438,42 @@ bool SignatureBinderBase::tryBind(
...
-        if (!tryBind(params[i], actualParameter.type)) {
-          // TODO Allow coercions for complex types.
+        Coercion childCoercion;
+        if (!tryBind(
+                params[i],
+                actualParameter.type,
+                coercion ? &childCoercion : nullptr)) {
           return false;
         }
+        if (childCoercion.type) {
```
MBkkt:
```suggestion
        if (childCoercion.cost > 0) {
```

In such case this condition 
```
    if (coercion->type->equivalent(*actualType)) {
      coercion->reset();
    }
```
is unnecessary

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +438,42 @@ bool SignatureBinderBase::tryBind(
...
+          needsCoercion = true;
+          totalCost += childCoercion.cost;
+          newParameters.emplace_back(childCoercion.type);
+        } else {
```
MBkkt:
```suggestion
        } else if (coercions) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +438,42 @@ bool SignatureBinderBase::tryBind(
...
+    std::vector<TypeParameter> typeParameters;
+    typeParameters.reserve(newParameters.size());
+    for (auto i = 0; i < newParameters.size(); ++i) {
+      typeParameters.emplace_back(newParameters[i], params[i].rowFieldName());
+    }
+    coercion->type = getType(typeName, typeParameters);
+    coercion->cost = totalCost;
+    if (coercion->type->equivalent(*actualType)) {
+      coercion->reset();
+    }
```
MBkkt: debug assert that they're not equal

velox/expression/SignatureBinder.cpp
```diff
@@ -369,18 +438,42 @@ bool SignatureBinderBase::tryBind(
...
+          newParameters.emplace_back(childCoercion.type);
+        } else {
+          newParameters.emplace_back(actualParameter.type);
+        }
         break;
+      }
     }
   }
+
+  if (needsCoercion) {
```
MBkkt:
```suggestion
  if (totalCost > 0) {
```

velox/expression/SignatureBinder.cpp
```diff
@@ -346,6 +410,11 @@ bool SignatureBinderBase::tryBind(
     return false;
   }
 
+  bool needsCoercion = false;
```
MBkkt:
```suggestion
```

velox/expression/SignatureBinder.cpp
```diff
@@ -229,72 +314,51 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
   const auto& baseName = typeSignature.baseName();
 
-  if (variables().count(baseName)) {
+  if (variables().contains(baseName)) {
     // Variables cannot have further parameters.
     VELOX_CHECK(
         typeSignature.parameters().empty(),
         "Variables with parameters are not supported");
-    auto& variable = variables().at(baseName);
+    const auto& variable = variables().at(baseName);
```
MBkkt: Make it single lookup please

velox/type/TypeCoercer.cpp
```diff
@@ -55,48 +56,54 @@ allowedCoercions() {
...
+  return {};
 }
 
 // static
-bool TypeCoercer::coercible(const TypePtr& fromType, const TypePtr& toType) {
-  if (fromType->isUnKnown()) {
-    return true;
+Coercion TypeCoercer::coercible(
+    const TypePtr& fromType,
+    const TypePtr& toType) {
```
MBkkt, line 100: this code can be called for same type (see If/SwitchExpr)
I think it's better to write
if (fromType->equal
```suggestion
Coercion TypeCoercer::coercible(
    const TypePtr& fromType,
    const TypePtr& toType) {
   if (fromType->equivalent(*toType)) {
      return {fromType, 0};
   }
```

velox/expression/SignatureBinder.cpp
```diff
@@ -229,72 +314,51 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
     }
 
-    if (variable.comparableTypesOnly() && !actualType->isComparable()) {
-      return false;
+    if (coercion) {
+      if (varType->equivalent(*actualType)) {
+        return true;
+      }
+      *coercion = TypeCoercer::coercible(actualType, varType);
+      return coercion->type != nullptr;
```
MBkkt:
```suggestion
      *coercion = TypeCoercer::coercible(actualType, varType);
      if (!*coercion) {
         return false;
      } else if (coercion->cost > 0) {
         coercion->reset();
      }
      return true;
```

velox/expression/SignatureBinder.cpp
```diff
@@ -229,72 +314,51 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
-    }
-
-    if (actualType->isUnKnown() && variable.knownTypesOnly()) {
-      return false;
-    }
-
-    if (variable.orderableTypesOnly() && !actualType->isOrderable()) {
-      return false;
+    auto& varType = typeVariablesBindings_[variable.name()];
+    if (!varType) {
```
MBkkt, line 286: this need only for no coercions case

velox/expression/SignatureBinder.cpp
```diff
@@ -155,19 +223,36 @@ bool SignatureBinder::tryBind(
...
+    for (size_t i = 0; i < formalArgsCnt && i < actualTypes_.size(); ++i) {
+      const auto& actualType = actualTypes_[i];
+      if (!actualType ||
+          !SignatureBinderBase::tryBind(formalArgs[i], actualType)) {
+        return false;
       }
-    } else {
+    }
+    return true;
+  }
```
MBkkt: lets merge this code and coercions code

velox/expression/SignatureBinder.cpp
```diff
@@ -318,55 +311,49 @@ bool SignatureBinderBase::tryBind(
...
-      if (varType->equivalent(*actualType)) {
-        return true;
-      }
-      *coercion = TypeCoercer::coercible(actualType, varType);
-      return coercion->type != nullptr;
+    if (!coercion) {
+      return varType->equivalent(*actualType);
     }
-    return varType->equivalent(*actualType);
+    *coercion = TypeCoercer::coercible(actualType, varType);
```
mkornaukhov, line 295: Here we should check for equivalence. If it is, then we should reset coercion. Otherwise, we will require, for example, `REAL -> REAL` coercion, that's what tests report.

mkornaukhov (dismissed):

velox/expression/SignatureBinder.cpp
```diff
@@ -229,80 +307,53 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
-    if (variable.comparableTypesOnly() && !actualType->isComparable()) {
-      return false;
+    if (!coercion) {
+      return varType->equivalent(*actualType);
     }
-
-    typeVariablesBindings_[baseName] = actualType;
-    return true;
+    *coercion = TypeCoercer::coercible(actualType, varType);
+    return static_cast<bool>(*coercion);
```
mkornaukhov:
```suggestion
    if (varType->equivalent(*actualType)) {
      return true;
    }
    *coercion = TypeCoercer::coercible(actualType, varType);
    return static_cast<bool>(*coercion);
```

mbasmanova (commented):
@MBkkt Thank you for working on this functionality. Would you update PR description to explain how coercions for complex types work and provide some examples that work and some examples that don't? Are there any limitations / assumptions in the implementation? Would be great to list these explicitly in the PR description. Thanks.

mbasmanova:
@MBkkt If possible, would you squash 30 commits into one. GitHub becomes slow when PR has so many commits.

MBkkt:
@mbasmanova PR description is updated, commits are rebased and squashed

mbasmanova:
> @mbasmanova PR description is updated, commits are rebased and squashed

@MBkkt Thank you. This is very helpful. For the items in Follow Up section, please, create GitHub issues and reference these.

velox/type/Cost.h
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstdint>
+#include <limits>
+
+namespace facebook::velox {
+
+using Cost = uint64_t;
```
mbasmanova: We probably do not want to introduce velox::Cost type. The name and location (top-level namespace) are too generic, but the usage is query specific. If we need this type, perhaps, move it to Coercion struct or rename to CoercionCost. Let's also remove type/Cost.h file. I assume it is not necessary.

velox/type/TypeCoercer.h
```diff
@@ -32,59 +37,35 @@ struct Coercion {
     return fmt::format("{} ({})", type->toString(), cost);
   }
 
-  void reset() {
-    type = nullptr;
-    cost = 0;
+  /// False if coercion is needed but not possible.
```
mbasmanova, line 51: The comment doesn't seem to match the code. cost == 0 means coercion is not needed, but this method returns 'false' when cost == 0. I don't think there is intuitive cast from Coercion to a boolean, hence, I suggest to remove this method altogether.

velox/type/TypeCoercer.h
```diff
@@ -32,59 +37,35 @@ struct Coercion {
...
-      return a.second < b.second;
-    });
-
-    if (costs[0].second < costs[1].second) {
-      return costs[0].first;
-    }
-
-    return std::nullopt;
-  }
+  /// Converts a list of valid Coercions into a list of TypePtr.
```
mbasmanova, line 64: Would you clarify what does it mean to 'convert' a Coercion to a Type? Are there any pre-requisites to calling this method? Please, document these. Why are we passing 'to' as a raw pointer and not returning a vector of types?

velox/type/TypeCoercer.h
```diff
@@ -32,59 +37,35 @@ struct Coercion {
...
 
 class TypeCoercer {
  public:
   /// Checks if the base of 'fromType' can be implicitly converted to a type
   /// with the given name.
+  /// Only types without type parameters are supported.
   ///
   /// @return "to" type and cost if conversion is possible.
-  static std::optional<Coercion> coerceTypeBase(
+  static Coercion coerceTypeBase(
```
mbasmanova, line 76: Why do you find it beneficial to remove the usage of std::optional? It seems convenient to have an explicit way to signal that coercion is not possible.

velox/type/TypeCoercer.h
```diff
@@ -32,59 +37,35 @@ struct Coercion {
...
+      std::vector<TypePtr>* to);
 };
 
 class TypeCoercer {
  public:
   /// Checks if the base of 'fromType' can be implicitly converted to a type
   /// with the given name.
+  /// Only types without type parameters are supported.
   ///
   /// @return "to" type and cost if conversion is possible.
```
mbasmanova: > /// @return "to" type and cost if conversion is possible.

This comment is no longer accurate. Would you update to clarify how the caller should use 'cost' to figure out whether coercion is possible or not?

velox/type/TypeCoercer.h
```diff
@@ -32,59 +37,35 @@ struct Coercion {
...
+  static Coercion coerceTypeBase(
       const TypePtr& fromType,
       const std::string& toTypeName);
 
   /// Checks if 'fromType' can be implicitly converted to 'toType'.
   ///
-  /// @return true if conversion is possible.
-  static bool coercible(const TypePtr& fromType, const TypePtr& toType);
+  /// @return "to" type and cost if conversion is possible.
+  static Coercion coercible(const TypePtr& fromType, const TypePtr& toType);
```
mbasmanova, line 83: I seems cleaner to just return the optional cost as the type is the same as toType, no?

velox/type/TypeCoercer.cpp
```diff
@@ -14,18 +14,36 @@
...
-      cost += coercion.cost;
+    if (!coercion) {
+      return kInvalidCost;
     }
+    cost += coercion.cost;
   }
+  if (cost == 0) {
+    return 0;
+  }
+  return kMaxFunctionRank * kMaxFunctionArgs + cost;
```
mbasmanova: Why are we adding kMaxFunctionRank * kMaxFunctionArgs to cost? Would you add a comment?

velox/type/tests/TypeCoercerTest.cpp
```diff
@@ -23,19 +23,27 @@ namespace {
...
+    VELOX_EXPECT_EQ_TYPES(coercion.type, toType);
+  }
 }
 
 void testBaseCoercion(const TypePtr& fromType) {
   auto coercion = TypeCoercer::coerceTypeBase(fromType, fromType->kindName());
-  ASSERT_TRUE(coercion.has_value());
-  VELOX_EXPECT_EQ_TYPES(coercion->type, fromType);
+  ASSERT_TRUE(coercion);
+  if (coercion.cost == 0) {
```
mbasmanova, line 37: Shouldn't this always be true? What would be an example where 'cost' is not 0?

velox/core/SimpleFunctionMetadata.h
```diff
@@ -141,7 +141,7 @@ struct TypeAnalysisResults {
     // in this case (1) is picked.
     // e.g: (Any, int) will be picked before (Any, Any)
     // e.g: Variadic<Array<Any>> is picked before Variadic<Any>.
-    uint32_t getRank() {
+    Cost getRank() const {
```
mbasmanova: It is a bit confusing to use Cost for rank. These are not the same. Can we keep these separate?

velox/experimental/cudf/expression/ExpressionEvaluator.cpp
```diff
@@ -118,7 +118,7 @@ static bool matchCallAgainstSignatures(
   for (const auto& sig : sigs) {
     std::vector<Coercion> coercions(n);
     exec::SignatureBinder binder(*sig, argTypes);
-    if (!binder.tryBindWithCoercions(coercions)) {
+    if (!binder.tryBind(&coercions)) {
```
mbasmanova: Is this a fix for a pre-existing bug? Let's move it into a separate PR.

CC: @karthikeyann

velox/expression/SimpleFunctionRegistry.h
```diff
@@ -156,23 +156,21 @@ class SimpleFunctionRegistry {
...
+    return resolveFunction(name, argTypes, &coercions);
   }
 
  private:
   std::optional<ResolvedSimpleFunction> resolveFunction(
       const std::string& name,
       const std::vector<TypePtr>& argTypes,
-      bool allowCoercion,
-      std::vector<TypePtr>& coercions) const;
+      std::vector<TypePtr>* coercions) const;
```
mbasmanova, line 173: This is a refactoring of existing code, right? If that's the case, would you extract it into a separate PR. This PR has a lot of non-trivial logic. It would be simpler to review and safer to land that separately from unrelated refactoring. Thanks.

velox/functions/tests/RegistryTestUtil.h
```diff
@@ -187,4 +187,26 @@ class VectorFuncFour : public velox::exec::VectorFunction {
   }
 };
 
+class VectorFuncFive : public velox::exec::VectorFunction {
```
mbasmanova, line 190: I know you are following the existing code, but this is quite confusing and hard to read. What does "Five" mean here? Would you add a comment to explain what's special about this function and, perhaps, choose a more descriptive name?

velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -115,6 +121,32 @@ class FunctionRegistryTest : public testing::Test {
...
+      const std::vector<TypePtr>& expectedCoercions) {
+    std::vector<TypePtr> coercions;
+    auto type = resolveFunctionWithCoercions(functionName, types, coercions);
+    VELOX_EXPECT_EQ_TYPES(type, expected);
+
+    if (expected != nullptr) {
+      EXPECT_EQ(types.size(), coercions.size());
+      EXPECT_EQ(expectedCoercions.size(), coercions.size());
+      for (auto i = 0; i < coercions.size(); ++i) {
+        if (expectedCoercions[i] == nullptr) {
```
mbasmanova: Can we use VELOX_EXPECT_EQ_TYPES here?

velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -772,10 +829,13 @@ TEST_F(FunctionRegistryTest, resolveFunctionWithCoercions) {
         },
         std::make_unique<DummyVectorFunction>());
 
-    testCannotResolve("foo", {ARRAY(TINYINT()), SMALLINT()});
+    testCoercions(
```
mbasmanova, line 825: Can we also test the following cases:

- foo(array(integer), integer) - no coercions
- foo(array(smallint), integer) - coercion for 1st arg
- foo(array(integer), smallint) - coercion for 2nd arg
- foo(array(integer), varchar) - can't resolve

velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -816,7 +885,70 @@ TEST_F(FunctionRegistryTest, resolveFunctionWithCoercions) {
              .build()},
         std::make_unique<DummyVectorFunction>());
 
-    testCannotResolve("foo", {TINYINT(), REAL()});
+    testCoercions("foo", {TINYINT(), REAL()}, REAL(), {REAL(), nullptr});
```
mbasmanova, line 894: Let's also test the case where 2nd argument needs coersion.

velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -816,7 +885,70 @@ TEST_F(FunctionRegistryTest, resolveFunctionWithCoercions) {
...
+  }
+
+  {
```
mbasmanova, line 903: Would you add a short comment to clarify what's special about this test case?

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
+  }
+  return true;
+}
+
 } // namespace
 
-bool SignatureBinder::tryBindWithCoercions(std::vector<Coercion>& coercions) {
-  return tryBind(true, coercions);
-}
+// Traverse all types in signatures and deduces the least common type for each
```
mbasmanova: typo: deduces -> deduce

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
+// type parameter. Recursion is needed to traverse complex types like Map<K, V>,
+// Array<Array<T>>, etc
+// Returns false iff some type parameters cannot be bound.
+bool SignatureBinder::coercibleToTypeVars(
+    const TypeSignature& signature,
+    const TypePtr& actualType) {
+  if (!actualType) {
```
mbasmanova, line 122: When would actualType be null?

velox/type/TypeCoercer.h
MBkkt: > The comment doesn't seem to match the code. cost == 0 means coercion is not needed, but this method returns 'false' when cost == 0

It's match the code.
There's three states:
1. coercion is needed and cast is needed, cost > 0 (type will be non-null)
2. coercion is possible but cast isn't need, cost == 0 (type will be null) 
3. coercion is impossible, cost == kInvalidCost (type will be null)

velox/type/TypeCoercer.h
MBkkt: > Would you clarify what does it mean to 'convert' a Coercion to a Type? 

Just extract type from each coercion in vector from Coercion to Type.
Type can be null, it's expected.

> Are there any pre-requisites to calling this method?

Coercion inside Coercions vector should be possible (cost != kInvalidCost).

> Why are we passing 'to' as a raw pointer and not returning a vector of types?

I want single place where I check to is nullptr or not

velox/type/TypeCoercer.h
MBkkt: Because Coercion already answer on this question.
In general I don't big fan of optional, they should be used only then they're really required.
For an example in velox a lot of code like `optional<smart_ptr>`, this is sad

velox/type/TypeCoercer.h
MBkkt: This is clarified in Coercion fields description, but yes I will update this

velox/type/TypeCoercer.h
MBkkt: I don't agree, we need cost for generic, and type is just for convenience on caller side

velox/type/TypeCoercer.cpp
MBkkt: Comment exists, I will move it to here.

It's needed because

> Change cost of coercion to prefer generic function without coercions over a function with coercions.

velox/type/tests/TypeCoercerTest.cpp
MBkkt: cost isn't zero than cast is needed or coercion is impossible

velox/core/SimpleFunctionMetadata.h
MBkkt: I want to use some using for `uint64_t`, what do you suggest instead Cost?
I also want to write in code that rank/priority/coercion cost use same type.
For this I need common `using T`, what `T` instead of `Cost` do you suggest? `FunctionCost`?

velox/experimental/cudf/expression/ExpressionEvaluator.cpp
MBkkt: It's not a bug, I just removed old internal signature, so this code changed.
I don't really want to separate this changes into multiple PRs

velox/expression/SimpleFunctionRegistry.h
MBkkt: Right.

> If that's the case, would you extract it into a separate PR. This PR has a lot of non-trivial logic. It would be simpler to review and safer to land that separately from unrelated refactoring. Thanks.

It's not simple to extract, refactoring is trivial

velox/functions/tests/RegistryTestUtil.h
MBkkt: I think, @mkornaukhov just to added test same way as it was written in this file, honestly speaking I have no idea what is written here, except it passes and tests needed lines of new code.

If you have code suggestion about tests, I will ok to apply it

velox/functions/tests/FunctionRegistryTest.cpp
MBkkt: @mkornaukhov will you add tests please?

velox/functions/tests/FunctionRegistryTest.cpp
MBkkt: @mkornaukhov will you add tests please?

velox/functions/tests/FunctionRegistryTest.cpp
MBkkt: @mkornaukhov will you add comment please?

velox/expression/SignatureBinder.cpp
MBkkt: In some Velox tests it's possible, we return null in such cases, but for some random reason we expect some variable info filled.

It's about case when funciton have function parameter (lambda), e.g. any_math

Even if we want to fix test, lets do this separately, we can create an issue

velox/type/Cost.h
MBkkt: Changed, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/type/TypeCoercer.h
MBkkt: Comments are updated, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/type/TypeCoercer.h
MBkkt: Comments are updated, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/type/TypeCoercer.h
MBkkt: Comments are updated, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/type/TypeCoercer.cpp
MBkkt: I move this in different place, because it's actually needed only for simple functions.

see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/core/SimpleFunctionMetadata.h
MBkkt: I renamed to `CallableCost`, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/functions/tests/FunctionRegistryTest.cpp
MBkkt: Applied, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

velox/expression/SignatureBinder.cpp
MBkkt: Fixed, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

MBkkt:
@mbasmanova I made update, to resolve some of your comments, see https://github.com/facebookincubator/velox/pull/15734/commits/6e44a9901b8929ebdc3e916f45362e802590036c

Mainly it's about comments and removing `Cost.h` and `velox::Cost`

For other comments I answered.

What is rest?

1. About more tests, I think @mkornaukhov will add more tests in Monday.
2. About issues for follow-ups, my suggestion is to create them from follow-ups before merge but after approve. Because it's still can change. For an example I think I find way to avoid Velox extensions for Postgres pseudo types.

I think other points to discuss, include:
1. different comments, if you don't like some of them or find them unclear, please write suggestion/question.
2. stylistic: optional vs non-optional, `bool + reference argument` vs `pointer`, etc. I think current code is a little more effective, but still correct and convenient, so if it's not against Velox codestyle I prefer to keep it like this. 
3. naming (classes/usings/etc): please just suggest name, I or @mkornaukhov will rename, it doesn't really matter for me.

velox/expression/SignatureBinder.cpp
MBkkt: I added "investigation of why actual type can be null" in follow-ups

velox/experimental/cudf/expression/ExpressionEvaluator.cpp
mbasmanova: This is not an internal signature, but rather public API. I see that coercion is not used here, hence, I believe it is a bug that this code uses tryBindWithCoercions. 

@karthikeyann Would you help take a look?

mbasmanova:
> optional vs non-optional, bool + reference argument vs pointer, etc. I think current code is a little more effective, but still correct and convenient, so if it's not against Velox codestyle I prefer to keep it like this.

I'm Ok with this change, but it needs to be a separate PR. Mixing refactoring and non-trivial functionality changes is an anti-pattern. It makes it difficult to review the code and hard to unland in case of unintended breakages. I saw you mentioned you don't want to split this PR. If that's still the case, I can take over the PR when I'm back from VeloxCon China. (We need this functionality in Axiom.) Let me know how you'd like to proceed.

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
+  }
+  return true;
+}
+
 } // namespace
 
-bool SignatureBinder::tryBindWithCoercions(std::vector<Coercion>& coercions) {
-  return tryBind(true, coercions);
-}
+// Traverse all types in signatures and deduce the least common type for each
```
mbasmanova: this comment belongs to .h file; please, move

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
 
+  return true;
+}
+
+bool SignatureBinder::tryBind(std::vector<Coercion>* coercions) {
   const auto& formalArgs = signature_.argumentTypes();
   const auto formalArgsCnt = formalArgs.size();
 
   if (signature_.variableArity()) {
+    VELOX_DCHECK_GE(formalArgsCnt, 1);
```
mbasmanova: I don't think we need this check. signature_ class should provide this guarantee.

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
+      for (size_t i = formalArgsCnt; i < actualTypes_.size(); ++i) {
+        if (!actualTypes_[i]) {
+          return false;
+        }
+        if (actualTypes_[i] == UNKNOWN()) {
+          continue;
+        }
+        if (!actualTypes_[i]->equivalent(*variableType)) {
+          return false;
         }
```
mbasmanova: Why did you need to change the code path that doesn't apply coercions? Was there a pre-existing bug? If so, this needs to be a separate PR.

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
       }
     }
-  } else {
-    if (formalArgsCnt != actualTypes_.size()) {
+  } else if (formalArgsCnt != actualTypes_.size()) {
+    return false;
+  }
+
+  // Phase 1: find least common var types
+  for (size_t i = 0; coercions && i < actualTypes_.size(); ++i) {
```
mbasmanova: This loop applies only if 'coercions' is not null, right? If so, it would be cleaner to put it inside an explicit 'if' statement.

```
if (coercions != nullptr) {
   for (...)
}
```

velox/expression/SignatureBinder.cpp
```diff
@@ -111,63 +118,134 @@ bool checkNamedRowField(
...
+    coercions->resize(bindArgsCnt);
+  } else if (formalArgsCnt < bindArgsCnt) {
+    bindArgsCnt = formalArgsCnt;
+  }
+
+  // Phase 2: check that bound types are coercible
+  for (size_t i = 0; i < bindArgsCnt; ++i) {
+    const auto& formalArgSignature =
+        i < formalArgsCnt ? formalArgs[i] : formalArgs.back();
+    if (!SignatureBinderBase::tryBind(
```
mbasmanova, line 197: In case when coercions are enabled, I believe this call to tryBind cannot return false. Wondering if it makes sense to have an explicit check for that.

velox/expression/SignatureBinder.cpp
```diff
@@ -229,80 +307,53 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
-      boost::algorithm::to_upper_copy(std::string(actualType->name()));
-
-  if (typeName != actualTypeName) {
-    if (allowCoercion) {
-      if (auto availableCoercion =
-              TypeCoercer::coerceTypeBase(actualType, typeName)) {
-        coercion = availableCoercion.value();
-        return true;
-      }
+  if (!boost::algorithm::iequals(typeName, actualType->name())) {
```
mbasmanova, line 303: I didn't know about boost::algorithm::iequals. This is nice improvement. Let's put it into a separate PR to avoid mixing pure refactoring with material changes.

velox/expression/SignatureBinder.cpp
```diff
@@ -229,80 +307,53 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
+
   if (isAny(typeSignature)) {
     return true;
   }
 
   const auto& baseName = typeSignature.baseName();
 
-  if (variables().count(baseName)) {
-    // Variables cannot have further parameters.
+  if (auto varIt = variables().find(baseName); varIt != variables().end()) {
```
mbasmanova: I assume this is an optimization to avoid double lookup in variables() map. This is nice. Let's move it into a separate PR to keep this PR as simple and short as possible.

velox/expression/SignatureBinder.cpp
```diff
@@ -229,80 +307,53 @@ bool SignatureBinderBase::checkOrSetIntegerParameter(
...
-        return true;
-      }
+  if (!boost::algorithm::iequals(typeName, actualType->name())) {
+    if (!coercion || !params.empty()) {
+      return false;
     }
-    return false;
+    // TODO: Handle coercion for complex types with parameters.
+    *coercion = TypeCoercer::coerceTypeBase(actualType, typeName);
+    return static_cast<bool>(*coercion);
```
mbasmanova: > static_cast<bool>(*coercion);

I feel code like this is hard to read an understand. Is there a way to simplify?

velox/type/TypeCoercer.h
```diff
@@ -19,10 +19,34 @@
...
 namespace facebook::velox {
 
-/// Type coercion necessary to bind a type to a signature.
+using CallableCost = uint64_t;
+
+// This assumes we wont have signature longer than 1M argument.
+inline constexpr CallableCost kMaxFunctionArgs = 1'000'000;
+
+// This assumes we wont have function rank number greater than 4.
+inline constexpr CallableCost kMaxFunctionRank = 4;
```
mbasmanova: These constants are specific to implementation detail of SimpleFunctionRegistry. Hence, I suggest to move these to velox/core/SimpleFunctionMetadata.h.

velox/expression/SignatureBinder.cpp
MBkkt: This isn't true, if coercions enabled this still can return false. Because even with coercions it's possible that we didn't find necessary coercion

velox/expression/SignatureBinder.cpp
MBkkt: I don't understand what is difficult about explicit cast to bool, but I think we have few options:
1) make cast to bool for Coercion class implicit
2) add method that can be called instead of explicit cast to bool (but it's needed only here)
3) replace explicit cast to bool with such expression: `!!*coercion`, it's quite common in assert/etc

velox/expression/SignatureBinder.cpp
MBkkt: >  signature_ class should provide this guarantee.

It's not true, I will change this

velox/expression/SignatureBinder.cpp
MBkkt: This code before this PR forget about is null check.
Also I think UNKNOWN check should check type instead of check type kind.

I think we can say it was bug, but no one checked it

velox/expression/SignatureBinder.cpp
MBkkt: ok

velox/type/TypeCoercer.h
MBkkt: I moved this, and one other constant from this common header, closer to their usage, see https://github.com/facebookincubator/velox/pull/15734/changes/861117750d189ac36413499955e1c28d07e1fb42

velox/expression/SignatureBinder.cpp
MBkkt: I followed second option, because I think you prefer it, see https://github.com/facebookincubator/velox/pull/15734/changes/861117750d189ac36413499955e1c28d07e1fb42

velox/expression/SignatureBinder.cpp
MBkkt: Applied, see https://github.com/facebookincubator/velox/pull/15734/changes/861117750d189ac36413499955e1c28d07e1fb42

velox/expression/SignatureBinder.cpp
MBkkt: I moved this check to function signature ctor, see https://github.com/facebookincubator/velox/pull/15734/changes/861117750d189ac36413499955e1c28d07e1fb42

velox/expression/SignatureBinder.cpp
MBkkt: Done, see https://github.com/facebookincubator/velox/pull/15734/changes/861117750d189ac36413499955e1c28d07e1fb42

MBkkt:
@mbasmanova I made separate refactoring PR and rebased this PR on refactoring PR, see https://github.com/facebookincubator/velox/pull/15769

I hope it's ok, if not, please feel free to get this changes in upstream as you wish

velox/expression/VectorFunction.cpp
```diff
@@ -102,39 +102,35 @@ TypePtr resolveVectorFunctionWithCoercions(
...
-        if (auto index = Coercion::pickLowestCost(candidates)) {
-          const auto& requiredCoercions = candidates[index.value()].first;
-          coercions.reserve(requiredCoercions.size());
-          for (const auto& coercion : requiredCoercions) {
-            coercions.push_back(coercion.type);
+          auto type = binder.tryResolveReturnType();
+          VELOX_CHECK_NOT_NULL(type);
+          const auto currentPriority = Coercion::overallCost(requiredCoercions);
+          if (currentPriority == 0) {
+            selectedCoercions = std::move(requiredCoercions);
```
bikramSingh91: nit: std::swap() for consistency with L125

velox/type/TypeCoercer.cpp
```diff
@@ -20,13 +20,26 @@ namespace facebook::velox {
...
-    }
+    VELOX_DCHECK(coercion);
+    cost += coercion.cost;
   }
   return cost;
 }
 
+void Coercion::convert(
+    const std::vector<Coercion>& from,
+    std::vector<TypePtr>* to) {
```
bikramSingh91, line 31: nit: is there any case where we pass a nullptr here? if not, can we just return the result in `to` rather then accepting it as an input param here

velox/expression/SimpleFunctionRegistry.cpp
```diff
@@ -235,16 +235,22 @@ SimpleFunctionRegistry::resolveFunction(
             selectedFunction = currentCandidate.get();
             selectedType = resultType;
             selectedPriority = currentPriority;
-            std::swap(selectedCoercions, requiredCoercions);
+            selectedCoercions = requiredCoercions;
+            selectedCount = 1;
+          } else if (currentPriority == selectedPriority) {
```
bikramSingh91, line 242: can you give an example of a case where there is a tie? Is there a test case for this code path?

velox/expression/SignatureBinder.cpp
```diff
@@ -104,70 +111,135 @@ bool checkNamedRowField(
...
+      return true;
+    }
+
+    return false;
+  }
+
+  const auto& signatureParams = signature.parameters();
+  const auto& actualParams = actualType->parameters();
+
+  const auto paramsCnt = std::min(signatureParams.size(), actualParams.size());
```
bikramSingh91, line 140: can you give an example of when the param count would differ and we still want to attempt coercion?

velox/expression/SignatureBinder.cpp
```diff
@@ -104,70 +111,135 @@ bool checkNamedRowField(
...
-    if (!isAny(signature_.argumentTypes().back())) {
-      if (actualTypes_.size() > formalArgsCnt) {
-        auto& type = actualTypes_[formalArgsCnt - 1];
-        for (auto i = formalArgsCnt; i < actualTypes_.size(); i++) {
-          if (!type->equivalent(*actualTypes_[i]) &&
-              actualTypes_[i]->kind() != TypeKind::UNKNOWN) {
-            return false;
-          }
-        }
+    if (!coercions && !isAny(signature_.argumentTypes().back()) &&
```
bikramSingh91, line 160: why does this require `coercions` to be a nullptr? can we not have variableArity and coercion together?

velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -816,7 +877,70 @@ TEST_F(FunctionRegistryTest, resolveFunctionWithCoercions) {
...
+        "foo",
+        {velox::exec::FunctionSignatureBuilder()
+             .returnType("integer")
+             .argumentType("integer")
+             .argumentType("integer")
+             .build()},
+        std::make_unique<DummyVectorFunction>());
+
+    testCoercions(
+        "foo", {INTEGER(), UNKNOWN()}, INTEGER(), {nullptr, INTEGER()});
```
bikramSingh91, line 963: can you add test cases :
- ROW types, with different children types, different num of children, different column names. 
- variableArity with coercion
- failing cases for the ones mentioned as limitation in the PR description

bikramSingh91 (commented):
There seem to multiple types of changes in this PR. I noticed you already separated out the refactoring part. However, the cost based changes intertwined with coercion changes are making it hard to follow and reason about the correctness. Can you please also separate out the cost specific changes so we can make sure every meaningful change to functionality is self contained and well tested before merging.

The order can look like:
1. Refactoring changes
2. Adding coercion support
3. Adding changes to costs of coercion and related decision making

velox/functions/tests/RegistryTestUtil.h
mkornaukhov: @mbasmanova 
Five means just the order, as well as all the previous ones.
It's just check that binding of vector function with complex generic signature is working with coercions.

velox/functions/tests/FunctionRegistryTest.cpp
mkornaukhov: There 2 complex types with same 2 type variables, I think it's important keys as some inaccuracies in implementation may make this test failing (I've faced it).

velox/functions/tests/FunctionRegistryTest.cpp
mkornaukhov: Could you explain what you mean in the third bullet?
First and second are added, see [491f898](https://github.com/facebookincubator/velox/pull/15734/commits/491f898ae24543b53417b145f0b64a98b3238a4b)

velox/expression/SimpleFunctionRegistry.cpp
mkornaukhov: Could you clarify the question?
When `currentPriority == selectedPriority` is possible? For example
```
signature foo(bigint, real)
signature foo(integer, double)

bind foo(integer(), real())
```
I do not know if there are tests covering this case, but this behavior is the same as before PR in the main (in case of choosing signatures with same rank).

velox/expression/SignatureBinder.cpp
mkornaukhov: Yes, I can.
```
signature array_construct(T, Variadic<T>)

bind array_construct(integer(), tinyint(), smallint(), bigint());
```

velox/expression/SignatureBinder.cpp
mkornaukhov: It's check for equivalence (but unknown are allowed, it's strange as for me, see  #15749).
In case of binding with coercions we will check for coercibility, not for equivalence.
Example
```
signature array_construct(T, Variadic<T>)

bind array_construct(integer(), tinyint(), smallint(), bigint());
```

In case of coercions we should allow it. In case of without coercions, we should check for equivalence.

velox/type/TypeCoercer.cpp
mkornaukhov: Yes, see SimpleFunctionRegistry.cpp, `SimpleFunctionRegistry::resolveFunction`.

velox/functions/tests/FunctionRegistryTest.cpp
mkornaukhov: Done in [491f898](https://github.com/facebookincubator/velox/pull/15734/commits/491f898ae24543b53417b145f0b64a98b3238a4b)

velox/functions/tests/FunctionRegistryTest.cpp
mkornaukhov: Done in [491f898](https://github.com/facebookincubator/velox/pull/15734/commits/491f898ae24543b53417b145f0b64a98b3238a4b)

MBkkt:
@bikramSingh91 

Thanks for looking this over. I see your goal of wanting maximum clarity in the changes.

For this contribution, we already structured it into the two parts: 
1) refactoring
2) functional changes 
We've tried to make sure the functional change is cohesive and testable in its current form.

Further splitting this functional changes would be a substantial piece of re-work. 
To be honest, I don't think the benefit would justify the effort required on our end at this point.

I'd really prefer we review and consider the PR as it stands. 
If the Velox process strictly requires a different structure, would it be possible for someone familiar with these code to handle the final split? I'm happy to provide context on the changes to make that easier.

Let me know how you'd like to proceed.

MBkkt:
I think PR description is outdated, because other PRs with same purpose was merged first, although created later.
This is sad.

Anyway this is updated version of PR with more than in the main tests.
I believe main difference in how coercions cost accounted and how generic types are resolved

PR #5442, opened 2023-06-28
Parquet write support combine batches to big row group

ulysses-you:
Parquet write would build row groups for each columnar batch, which causes a lot of small row groups.  It's due to each call on arrow WriteTable would build new row group.

This pr caches the input columnar batch and flush to arrow when:
- the cached numRows bigger than `rowsInRowGroup`
- the cached bytes bigger than `bytesInRowGroup`

Note, `bytesInRowGroup` is a new added option which is commonly used at java side.

PR #15751, opened 2025-12-11
feat: Flush row group by buffered bytes in parquet writer

wecharyu:
https://github.com/facebookincubator/velox/pull/5442 check `bytesInRowGroup` based on uncompressed bytes, which will cause the final compressed row group is much more smaller than config `bytesInRowGroup`. 

In this patch we flush row group based on buffered size on arrow, it could reduce the row group numbers and improve the read performance.

### DIFF
--- a/velox/dwio/parquet/writer/arrow/Properties.h
+++ b/velox/dwio/parquet/writer/arrow/Properties.h
@@ -398,7 +398,8 @@ class PARQUET_EXPORT WriterProperties {
     /// Specify the max number of rows to put in a single row group.
     /// Default 1Mi rows.
     Builder* max_row_group_length(int64_t max_row_group_length) {
-      ARROW_CHECK_GT(max_row_group_length, 0) << "max_row_group_length must be positive";
+      ARROW_CHECK_GT(max_row_group_length, 0)
+          << "max_row_group_length must be positive";
       max_row_group_length_ = max_row_group_length;
       return this;
     }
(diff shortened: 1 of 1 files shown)


xiaoxmeng:
@wecharyu sorry, but I merged this diff by mistake. It is now reverted by https://github.com/facebookincubator/velox/pull/16317. You can continue work on this. Thanks!

wecharyu:
@xiaoxmeng nvm, I'll align the code with [arrow-48467](https://github.com/apache/arrow/pull/48468).

velox/dwio/parquet/writer/arrow/Writer.cpp
```diff
@@ -492,16 +498,31 @@ class FileWriterImpl : public FileWriter {
...
+        double avg_row_size = buffered_bytes * 1.0 / group_rows;
+        batch_size = std::min(
+            batch_size,
+            static_cast<int64_t>(
+                (max_row_group_bytes - buffered_bytes) / avg_row_size));
+      }
+      if (batch_size > 0) {
+        RETURN_NOT_OK(WriteBatch(offset, batch_size));
+        offset += batch_size;
+      } else if (offset < batch.num_rows()) {
```
PingLiuPing, line 524: This is hard to read compare to original code, since it implies when "batch_size <= 0" && "offset < batch.num_rows()" it writes a new group.
Can we add a var such as `int64_t available_rows = max_row_group_length - group_rows;` and then adjust this value based on your new code 
while {
...

if (group_rows > 0) {
  if (buffered_bytes >= max_row_group_bytes) {
          available_rows = 0;
  } else if (buffered_bytes > 0) {
   double avg_row_size = buffered_bytes * 1.0 / group_rows;
   int64_t rows_by_bytes = 
              static_cast<int64_t>((max_row_group_bytes - buffered_bytes) / avg_row_size);
    available_rows = std::min(available_rows, rows_by_bytes);
}

if (available_rows <= 0) {
  RETURN_NOT_OK(NewBufferedRowGroup());
}
}

int64_t batch_size = std::min(available_rows, batch.num_rows() - offset);
RETURN_NOT_OK(WriteBatch(offset, batch_size));
offset += batch_size;
}

wecharyu:
Hi @PingLiuPing, I want first make https://github.com/apache/arrow/pull/48468 ready, then we can cherry-pick it and make little additional change in velox.

Issue #15551, opened 2025-11-18
Use bloom filter to reduce hash table lookups in HashProbe


kagamiori:
### Description

We're running TPC-H sf1000 queries on Prestissimo/Velox with in-memory tables. We noticed that for some TPC-H queries, such as q12, the probe side of an inner join is large (1.5B rows), while the join is selective (31M rows in output). Additionally, because the number of distinct values from each parallel build side exceeds VectorHasher::kMaxDistinct that is 100k, the join doesn't push down dynamic filter on the probe side. As a result, the query spends 23.5% CPU cycles in HashTable::joinNormalizedKeyProbe(). 

<img width="1304" height="395" alt="Image" src="https://github.com/user-attachments/assets/6945b283-60ee-49ce-84b8-b0010c1aea66" />

A optimization that @arhimondr proposed is to use a bloom filter before HashTable::joinProbe() so that values that do not pass the bloom filter do not need to probe the hash table.

kagamiori:
cc @mbasmanova

mbasmanova:
CC: @HolyLow @rui-mo @duanmeng

rui-mo:
cc: @zhouyuan

kagamiori:
Proof-of-concept PR by @arhimondr: https://github.com/facebookincubator/velox/pull/15566

PR #15752, opened 2025-12-11
feat: Dynamic Bloom filter pushdown from hash probe

Yuhta:
Summary:
Implement dynamic Bloom filter pushdown from hash probe.  This is controlled by query config `hash_probe_bloom_filter_pushdown_max_size` and turned off (set to 0) by default.

Also add a query config `hash_probe_dynamic_filter_pushdown_enabled` to disable dynamic filter generation from hash probe to help debug.

Fix https://github.com/facebookincubator/velox/issues/15551

Differential Revision: D88089324


kagamiori (approved):
LGTM. Thank you for adding the new feature to support bloom filter pushdown from join!

PR #15754, opened 2025-12-12
feat: Enable caching and re-use of HashTable in HashBuild operator

shrinidhijoshi:
Summary:

When Velox is used in materialized exchange systems like presto-on-spark, each task of a broadcast join is going to re-construct the same HashTable for the build side of the broadcast join.
We want to have the ability to construct the HashTable once and re-use it for all the tasks of the that broadcast join (planNodeId) that land on that worker.

Below are some design considerations addressed in this first version.

## Cache Table Structure

`Str` -> `HashTableCacheEntry`
`[QueryId:PlanNodeId]`  `[HashTable, buildComplete (true,false), builderTaskId, waiterTaskPromises]`

## Memory management
- The HashTables are cached under the Query memory pool. This is because the HT rows need to outlive the task / operator.
- Also we want to ensure that when query context is invalidated, the cached tables are also invalidated. This can happen in scenarios where the task runs into some issues during spilling
## Build synchronization
- **Among drivers of the same task** - there is no synchronization needed as all the drivers will build the partial table and the last driver will merge the tables. This part stays the same as existing hash build. But the last driver, along with setting the HT on the Join Bridge will also set put it in the cache
- **Among different tasks** - the first task whose driver initiates the entry creation, will set its taskId as the builderTaskId. Upon getting this entry when the other tasks check for a hash table, they see that a HashTable entry `buildComplete=False` and  `builderTaskId`  is a different task. They identify themselves as waiter tasks and queue a promise in that entry. So that they are notified when build is complete

## Post build / Join bridge transfer
- This is done as usual by the completing driver thread. It also sets 

## Eviction
- Currently the eviction has been left to the the user of the velox library by invoking the `flushEntry()` function on the cache. In future, we can use reference counting on probe operators and flush entries on memory pressure.

## Spilling
- Currently not supported. No plan in future. We just flush and rebuild
  
## Sample e2e runs
Internal benchmarks at Meta, for sample presto-on-spark queries show ~5x reduction in cost of jobs and ~2x in wall time
Screenshot 2025-12-23 at 3 19 56 PM
Differential Revision: D88284809

### DIFF
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -3191,7 +3191,8 @@ class HashJoinNode : public AbstractJoinNode {
       TypedExprPtr filter,
       PlanNodePtr left,
       PlanNodePtr right,
-      RowTypePtr outputType)
+      RowTypePtr outputType,
+      bool useHashTableCache = false)
       : AbstractJoinNode(
             id,
             joinType,
@@ -3201,7 +3202,8 @@ class HashJoinNode : public AbstractJoinNode {
             std::move(left),
             std::move(right),
             std::move(outputType)),
-        nullAware_{nullAware} {
+        nullAware_{nullAware},
+        useHashTableCache_{useHashTableCache} {
     validate();
 
     if (nullAware) {
@@ -3226,10 +3228,16 @@ class HashJoinNode : public AbstractJoinNode {
     explicit Builder(const HashJoinNode& other)
         : AbstractJoinNode::Builder<HashJoinNode, Builder>(other) {
       nullAware_ = other.isNullAware();
+      useHashTableCache_ = other.useHashTableCache();
     }
 
     Builder& nullAware(bool value) {
       nullAware_ = value;
+      return *this;
+    }
+
+    Builder& useHashTableCache(bool value) {
+      useHashTableCache_ = value;
       return *this;
     }
 
@@ -3259,11 +3267,13 @@ class HashJoinNode : public AbstractJoinNode {
           filter_.value_or(nullptr),
           left_.value(),
           right_.value(),
-          outputType_.value());
+          outputType_.value(),
+          useHashTableCache_.value_or(false));
     }
 
    private:
     std::optional<bool> nullAware_;
+    std::optional<bool> useHashTableCache_;
   };
 
   std::string_view name() const override {
@@ -3286,6 +3296,12 @@ class HashJoinNode : public AbstractJoinNode {
     return nullAware_;
   }
 
+  /// Returns whether hash table caching is enabled for broadcast joins.
+  /// Only used by Presto-on-Spark.
+  bool useHashTableCache() const {
+    return useHashTableCache_;
+  }
+
   folly::dynamic serialize() const override;
 
   static PlanNodePtr create(const folly::dynamic& obj, void* context);
@@ -3294,6 +3310,7 @@ class HashJoinNode : public AbstractJoinNode {
   void addDetails(std::stringstream& stream) const override;
 
   const bool nullAware_;
+  const bool useHashTableCache_;
 };
 
 using HashJoinNodePtr = std::shared_ptr<const HashJoinNode>;
--- a/velox/docs/develop/joins.rst
+++ b/velox/docs/develop/joins.rst
@@ -198,6 +198,11 @@ the join is executed using broadcast or partitioned strategy has no effect on
 the join execution itself. The only difference is that broadcast execution
 allows for dynamic filter pushdown while partitioned execution does not.
 
+HashJoinNode supports a ``useHashTableCache`` flag (used only by Presto-on-Spark)
+that enables caching of the hash table built for broadcast joins. When enabled,
+the first task to build the hash table stores it in a global cache, and subsequent
+tasks from same query reuse the cached table instead of rebuilding it.
+
 PartitionedOutput operator and OutputBufferManager support
 broadcasting the results of the plan evaluation. This functionality is enabled
 by setting boolean flag "broadcast" in the PartitionedOutputNode to true.
--- a/velox/exec/HashBuild.cpp
+++ b/velox/exec/HashBuild.cpp
@@ -18,6 +18,7 @@
 #include "velox/common/base/Counters.h"
 #include "velox/common/base/StatsReporter.h"
 #include "velox/common/testutil/TestValue.h"
+#include "velox/exec/HashTableCache.h"
 #include "velox/exec/OperatorUtils.h"
 #include "velox/exec/Task.h"
 #include "velox/exec/VectorHasher.h"
@@ -113,17 +114,105 @@ HashBuild::HashBuild(
   }
 
   tableType_ = hashJoinTableType(joinNode_);
-  setupTable();
-  setupSpiller();
+
   stateCleared_ = false;
 }
 
 void HashBuild::initialize() {
   Operator::initialize();
 
+  if (setupCachedHashTable()) {
+    return;
+  }
+
+  // Set up table and spiller now that cache state is initialized.
+  // This ensures tableMemoryPool() returns the cache's tablePool when enabled.
+  setupTable();
+  setupSpiller();
+
   if (isAntiJoin(joinType_) && joinNode_->filter()) {
     setupFilterForAntiJoins(keyChannelMap_);
   }
+}
+
+bool HashBuild::setupCachedHashTable() {
+  if (!joinNode_->useHashTableCache()) {
+    return false;
+  }
+
+  const auto& queryId = operatorCtx_->task()->queryCtx()->queryId();
+  cacheKey_ = fmt::format("{}:{}", queryId, planNodeId());
+
+  // Get or create the cache entry (which includes the pool).
+  // If another task is already building, future_ will be set.
+  auto* cache = HashTableCache::instance();
+  auto* queryCtx = operatorCtx_->task()->queryCtx().get();
+  cacheEntry_ = cache->get(cacheKey_, taskId(), queryCtx, &future_);
+  VELOX_CHECK_NOT_NULL(cacheEntry_);
+  VELOX_CHECK_NOT_NULL(cacheEntry_->tablePool);
+
+  // Check if table is already built.
+  if (cacheEntry_->buildComplete) {
+    noMoreInput();
+    return true;
+  }
+
+  // Check if we're a waiter task (future was set by get).
+  if (future_.valid()) {
+    setState(State::kWaitForBuild);
+    return true;
+  }
+
+  // This is the builder task - proceed with building.
+  return false;
+}
+
+bool HashBuild::getHashTableFromCache() {
+  if (!useHashTableCache()) {
+    return false;
+  }
+
+  if (!cacheEntry_->buildComplete) {
+    // Cache miss - we need to build the table.
+    stats_.wlock()->addRuntimeStat(
+        BaseHashTable::kHashTableCacheMiss, RuntimeCounter(1));
+    return false;
+  }
+
+  // Table already built by a previous task! Use it directly.
+  // Notify the bridge with the cached table.
+  // We pass a shared_ptr copy (not std::move) since the cache retains
+  // ownership.
+  joinBridge_->setHashTable(
+      cacheEntry_->table, {}, cacheEntry_->hasNullKeys, nullptr);
+  // Record cache hit metric.
+  stats_.wlock()->addRuntimeStat(
+      BaseHashTable::kHashTableCacheHit, RuntimeCounter(1));
+  return true;
+}
+
+void HashBuild::maybeSetHashTableInCache(
+    const std::shared_ptr<BaseHashTable>& table) {
+  if (!useHashTableCache()) {
+    return;
+  }
+  auto* cache = HashTableCache::instance();
+  cache->put(cacheKey(), table, joinHasNullKeys_);
+}
+
+bool HashBuild::receivedCachedHashTable() {
+  if (!useHashTableCache() || future_.valid()) {
+    return false;
+  }
+  // We were waiting on cached table from another task.
+  // Ensure that table is ready.
+  VELOX_CHECK(
+      cacheEntry_->buildComplete,
+      "Signalled that cache table is ready but it is not built yet.");
+  // Proceed through normal noMoreInput flow which will use the cache.
+  setRunning();
+  noMoreInput();
+  return true;
 }
 
 void HashBuild::setupTable() {
@@ -153,7 +242,7 @@ void HashBuild::setupTable() {
         true, // allowDuplicates
         true, // hasProbedFlag
         queryConfig.minTableRowsForParallelJoinBuild(),
-        pool());
+        tableMemoryPool());
   } else {
     // Right semi join needs to tag build rows that were probed.
     const bool needProbedFlag = joinNode_->isRightSemiFilterJoin();
@@ -166,7 +255,7 @@ void HashBuild::setupTable() {
           !dropDuplicates_, // allowDuplicates
           needProbedFlag, // hasProbedFlag
           queryConfig.minTableRowsForParallelJoinBuild(),
-          pool());
+          tableMemoryPool());
     } else {
       // Ignore null keys
       table_ = HashTable<true>::createForJoin(
@@ -175,7 +264,7 @@ void HashBuild::setupTable() {
           !dropDuplicates_, // allowDuplicates
           needProbedFlag, // hasProbedFlag
           queryConfig.minTableRowsForParallelJoinBuild(),
-          pool(),
+          tableMemoryPool(),
           queryConfig.hashProbeBloomFilterPushdownMaxSize());
     }
   }
@@ -319,6 +408,11 @@ void HashBuild::removeInputRowsForAntiJoinFilter() {
 
 void HashBuild::addInput(RowVectorPtr input) {
   checkRunning();
+
+  VELOX_CHECK(
+      !useHashTableCache() ||
+      (cacheEntry_->builderTaskId == taskId() && !cacheEntry_->buildComplete));
+
   ensureInputFits(input);
 
   TestValue::adjust("facebook::velox::exec::HashBuild::addInput", this);
@@ -656,6 +750,7 @@ void HashBuild::noMoreInput() {
   if (noMoreInput_) {
     return;
   }
+
   Operator::noMoreInput();
 
   noMoreInputInternal();
@@ -699,6 +794,10 @@ bool HashBuild::finishHashBuild() {
       promise.setValue();
     }
   };
+
+  if (getHashTableFromCache()) {
+    return true;
+  }
 
   if (joinHasNullKeys_ && isAntiJoin(joinType_) && nullAware_ &&
       !joinNode_->filter()) {
@@ -815,11 +914,16 @@ bool HashBuild::finishHashBuild() {
               spillStats);
         };
   }
+
+  // For hash table caching: the last driver caches the merged table.
+  std::shared_ptr<BaseHashTable> table = std::move(table_);
+  maybeSetHashTableInCache(table);
   joinBridge_->setHashTable(
-      std::move(table_),
+      table,
       std::move(spillPartitions),
       joinHasNullKeys_,
       std::move(tableSpillFunc));
+
   if (canSpill()) {
     stateCleared_ = true;
   }
@@ -871,7 +975,6 @@ void HashBuild::ensureTableFits(uint64_t numRows) {
 
 void HashBuild::postHashBuildProcess() {
   checkRunning();
-
   if (!canSpill()) {
     setState(State::kFinish);
     return;
@@ -1044,6 +1147,11 @@ BlockingReason HashBuild::isBlocked(ContinueFuture* future) {
     case State::kFinish:
       break;
     case State::kWaitForBuild:
+      if (receivedCachedHashTable()) {
+        break;
+      }
+      // We were waiting for peer drivers to finish - fall through to
+      // kWaitForProbe which has the same logic.
       [[fallthrough]];
     case State::kWaitForProbe:
       if (!future_.valid()) {
@@ -1125,6 +1233,11 @@ std::string HashBuild::stateName(State state) {
 
 bool HashBuild::canSpill() const {
   if (!Operator::canSpill()) {
+    return false;
+  }
+  // For Cached hash table, we don't support spill either by the
+  // task thats building or by the task that is re-using it
+  if (useHashTableCache()) {
     return false;
   }
   if (operatorCtx_->task()->hasMixedExecutionGroupJoin(joinNode_.get())) {
@@ -1216,6 +1329,19 @@ void HashBuild::reclaim(
     buildOp->table_->clear(true);
     buildOp->pool()->release();
   }
+}
+
+memory::MemoryPool* HashBuild::tableMemoryPool() const {
+  if (useHashTableCache()) {
+    // Cached hash tables use a leaf pool under the query pool (from cache
+    // entry). This allows the table to outlive the task while still supporting
+    // allocations.
+    VELOX_CHECK_NOT_NULL(cacheEntry_);
+    VELOX_CHECK_NOT_NULL(cacheEntry_->tablePool);
+    return cacheEntry_->tablePool.get();
+  }
+  // Regular joins use operator pool
+  return pool();
 }
 
 bool HashBuild::nonReclaimableState() const {
--- a/velox/exec/HashBuild.h
+++ b/velox/exec/HashBuild.h
@@ -17,6 +17,7 @@
 
 #include "velox/exec/HashJoinBridge.h"
 #include "velox/exec/HashTable.h"
+#include "velox/exec/HashTableCache.h"
 #include "velox/exec/Operator.h"
 #include "velox/exec/Spill.h"
 #include "velox/exec/Spiller.h"
@@ -106,6 +107,25 @@ class HashBuild final : public Operator {
 
   // Invoked to set up hash table to build.
   void setupTable();
+
+  // Sets up hash table caching if enabled. Returns true if the cached table
+  // is already available or if this operator should wait for another task
+  // to build it, in which case further initialization should be skipped.
+  // Returns false if this operator should proceed with building the table.
+  bool setupCachedHashTable();
+
+  // Checks if a cached hash table is available and uses it if so.
+  // Returns true if the cached table was used (build can be skipped).
+  // Returns false if we need to build the table (cache miss).
+  bool getHashTableFromCache();
+
+  // Called when waiting for a cached hash table from another task.
+  // Returns true if the cached table was received and noMoreInput was called.
+  bool receivedCachedHashTable();
+
+  // Stores the built hash table in the cache for reuse by other tasks.
+  // No-op if hash table caching is not enabled.
+  void maybeSetHashTableInCache(const std::shared_ptr<BaseHashTable>& table);
 
   // Invoked when operator has finished processing the build input and wait for
   // all the other drivers to finish the processing. The last driver that
@@ -212,6 +232,27 @@ class HashBuild final : public Operator {
   // Invoked to abandon build deduped hash table.
   void abandonHashBuildDedup();
 
+  // Returns true if this operator is using a cached hash table.
+  // When enabled, the hash table is built once and cached for reuse
+  // by other tasks within the same query and stage.
+  bool useHashTableCache() const {
+    return !cacheKey_.empty();
+  }
+
+  // Returns the hash table cache key for this operator.
+  // Only valid if useHashTableCache() returns true.
+  const std::string& cacheKey() const {
+    VELOX_CHECK(
+        useHashTableCache(),
+        "cacheKey() called when table caching is not enabled");
+    return cacheKey_;
+  }
+
+  // Determines the memory pool to use for the hash table.
+  // For cached hash tables, uses query-level pool so the table can
+  // outlive the task. For regular joins, uses operator pool.
+  memory::MemoryPool* tableMemoryPool() const;
+
   const std::shared_ptr<const core::HashJoinNode> joinNode_;
 
   const core::JoinType joinType_;
@@ -246,6 +287,15 @@ class HashBuild final : public Operator {
   tsan_atomic<bool> exceededMaxSpillLevelLimit_{false};
 
   State state_{State::kRunning};
+
+  // For hash table caching: the cache key passed in at construction.
+  // If set, this operator coordinates via HashTableCache.
+  // Key format: "queryId:planNodeId"
+  std::string cacheKey_;
+
+  // For hash table caching: cached entry containing the shared table and pool.
+  // Retrieved from HashTableCache.
+  std::shared_ptr<HashTableCacheEntry> cacheEntry_;
 
   // The row type used for hash table build and disk spilling.
   RowTypePtr tableType_;
--- a/velox/exec/HashJoinBridge.h
+++ b/velox/exec/HashJoinBridge.h
@@ -53,8 +53,9 @@ class HashJoinBridge : public JoinBridge {
   /// Invoked by the build operator to set the built hash table.
   /// 'spillPartitionSet' contains the spilled partitions while building
   /// 'table' which only applies if the disk spilling is enabled.
+  /// Accepts both unique_ptr (regular joins) and shared_ptr (broadcast joins).
   void setHashTable(
-      std::unique_ptr<BaseHashTable> table,
+      std::shared_ptr<BaseHashTable> table,
       SpillPartitionSet spillPartitionSet,
       bool hasNullKeys,
       HashJoinTableSpillFunc&& tableSpillFunc);
--- a/velox/exec/HashTableCache.cpp
+++ b/velox/exec/HashTableCache.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/exec/HashTableCache.h"
+
+#include <fmt/format.h>
+
+#include "velox/core/QueryCtx.h"
+
+namespace facebook::velox::exec {
+
+HashTableCache* HashTableCache::instance() {
+  static HashTableCache instance;
+  return &instance;
+}
+
+std::shared_ptr<HashTableCacheEntry> HashTableCache::get(
+    const std::string& key,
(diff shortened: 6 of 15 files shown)


velox/exec/HashJoinBridge.h
```diff
@@ -53,8 +53,9 @@ class HashJoinBridge : public JoinBridge {
   /// Invoked by the build operator to set the built hash table.
   /// 'spillPartitionSet' contains the spilled partitions while building
   /// 'table' which only applies if the disk spilling is enabled.
+  /// Accepts both unique_ptr (regular joins) and shared_ptr (broadcast joins).
```
xiaoxmeng, line 56: We only accept shared ptr now? Can we fix the comment?

velox/core/PlanNode.h
```diff
@@ -3186,7 +3186,8 @@ class HashJoinNode : public AbstractJoinNode {
       TypedExprPtr filter,
       PlanNodePtr left,
       PlanNodePtr right,
-      RowTypePtr outputType)
+      RowTypePtr outputType,
+      bool useCachedHashTable = false)
```
xiaoxmeng:
```
s/useCachedHashTable/useHashTableCache/
```
Can you comment that this is set if it is used by broadcast join for Presto on Spark?

velox/core/PlanNode.h
```diff
@@ -3186,7 +3186,8 @@ class HashJoinNode : public AbstractJoinNode {
       TypedExprPtr filter,
       PlanNodePtr left,
       PlanNodePtr right,
-      RowTypePtr outputType)
+      RowTypePtr outputType,
```
xiaoxmeng, line 3194: Can you also update the plan node docs?

velox/exec/HashBuild.h
```diff
@@ -244,6 +266,19 @@ class HashBuild final : public Operator {
 
   State state_{State::kRunning};
 
+  // For hash table caching: the cache key passed in at construction.
+  // If set, this operator coordinates via HashTableCache.
+  // Key format: "queryId:stageId:planNodeId"
```
xiaoxmeng: no stage id?

velox/exec/HashBuild.cpp
```diff
@@ -75,6 +76,22 @@ HashBuild::HashBuild(
   VELOX_CHECK(pool()->trackUsage());
   VELOX_CHECK_NOT_NULL(joinBridge_);
 
+  // Compute cache key if this is a cacheable join (needed before setupTable
+  // to determine which memory pool to use).
+  if (joinNode_->useCachedHashTable()) {
+    const auto& queryId = operatorCtx_->task()->queryCtx()->queryId();
+    cacheKey_ = fmt::format("{}:{}", queryId, planNodeId());
```
xiaoxmeng: do we need to keep cacheKey_ in the object? Is it only used here?

velox/exec/HashBuild.cpp
```diff
@@ -119,6 +137,26 @@ HashBuild::HashBuild(
...
+    VELOX_CHECK_NOT_NULL(cacheEntry_);
+
+    // Check if table is already built.
+    if (cacheEntry_->buildComplete.load(std::memory_order_acquire)) {
+      noMoreInput();
+      return;
+    }
+
+    // Check if we're a waiter task (future was set by getOrAwait).
+    if (cacheWaitFuture_.valid()) {
```
xiaoxmeng: Shall we first check on cacheWaitFuture_?

velox/exec/HashBuild.cpp
```diff
@@ -152,7 +190,7 @@ void HashBuild::setupTable() {
         operatorCtx_->driverCtx()
             ->queryConfig()
             .minTableRowsForParallelJoinBuild(),
-        pool());
+        getTableMemoryPool());
```
xiaoxmeng: Why we need getTableMemoryPool() here?

velox/exec/HashBuild.cpp
```diff
@@ -119,6 +137,26 @@ HashBuild::HashBuild(
 void HashBuild::initialize() {
   Operator::initialize();
 
+  // For hash table caching: handle cache coordination across tasks.
+  if (isUsingCachedTable()) {
+    VELOX_CHECK_NOT_NULL(cacheEntry_);
+
+    // Check if table is already built.
+    if (cacheEntry_->buildComplete.load(std::memory_order_acquire)) {
+      noMoreInput();
```
xiaoxmeng: Do we need returns false for needsInput if we use cached hash table?

velox/exec/HashBuild.cpp
```diff
@@ -321,6 +359,14 @@ void HashBuild::removeInputRowsForAntiJoinFilter() {
 
 void HashBuild::addInput(RowVectorPtr input) {
   checkRunning();
+
+  // For hash table caching: check if table is already cached by another task.
+  // If so, skip processing this input - we'll use the cached table.
+  if (isUsingCachedTable() &&
```
xiaoxmeng: We shall handle needsInput instead here?

velox/exec/HashBuild.cpp
```diff
@@ -702,6 +749,31 @@ bool HashBuild::finishHashBuild() {
     }
   };
 
+  // For hash table caching: check if table already exists in cache (from
+  // previous task). This check is done AFTER the barrier to ensure only the
+  // last driver handles it.
+  if (isUsingCachedTable()) {
+    VELOX_CHECK_NOT_NULL(cacheEntry_);
+
+    if (cacheEntry_->buildComplete.load(std::memory_order_acquire)) {
```
xiaoxmeng: We can just use the default memory access order.

velox/exec/HashBuild.cpp
```diff
@@ -702,6 +749,31 @@ bool HashBuild::finishHashBuild() {
...
+
+      // Record cache hit metric
+      stats_.wlock()->addRuntimeStat("hashTableCacheHit", RuntimeCounter(1));
+
+      return true;
+    }
+
+    // Cache miss - we need to build the table
+    // Record cache miss metric
+    stats_.wlock()->addRuntimeStat("hashTableCacheHit", RuntimeCounter(0));
```
xiaoxmeng:
```
hashTableCacheHit
hashTableCacheMiss
```

velox/exec/HashBuild.cpp
```diff
@@ -873,6 +963,16 @@ void HashBuild::ensureTableFits(uint64_t numRows) {
 void HashBuild::postHashBuildProcess() {
   checkRunning();
 
+  // For waiter tasks with a cached hash table, skip spill handling
+  // and transition directly to finish state. The hash table was already
+  // built by the builder task and is shared via the cache.
+  if (isUsingCachedTable() && cacheEntry_ &&
```
xiaoxmeng: in which case cacheEntry_ is null but isUsingCachedTable() is true?

velox/exec/HashBuild.cpp
```diff
@@ -873,6 +963,16 @@ void HashBuild::ensureTableFits(uint64_t numRows) {
...
+      cacheEntry_->buildComplete.load(std::memory_order_acquire) &&
+      cacheEntry_->builderTaskId.value() != taskId()) {
```
xiaoxmeng: We shouldn't support spill if cache hash table? So do we need to check if we are the builder here?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include <fmt/format.h>
+
+namespace facebook::velox::exec {
+
+std::atomic<HashTableCache*> HashTableCache::instance_{nullptr};
+std::mutex HashTableCache::instanceMutex_;
+
+HashTableCache* HashTableCache::getInstance() {
+  auto* instance = instance_.load(std::memory_order_acquire);
```
xiaoxmeng: We can directly use a lock here as it is not on hot path. instance_ is protected by lock and it doesn't need to be atomic variable for simplicity.

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+  return instance;
+}
+
+/// Retrieves or creates the memory pool for a cached table.
+/// This must be called while holding the tables_ lock.
+///
+/// @param key The unique identifier for the cached table
+/// @param queryPool The query-level memory pool to create child pool under
+/// @param entry The cache entry to store the pool in
+static void getOrCreateTablePool(
```
xiaoxmeng: put this into anonymous namespace and drop static?

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,83 @@
...
+      ContinueFuture* future);
+
+  /// Stores a built hash table and notifies waiting tasks.
+  void put(
+      const std::string& key,
+      std::shared_ptr<BaseHashTable> table,
+      bool hasNullKeys);
+
+  /// Removes a cache entry.
+  static void flushEntry(const std::string& cacheKey);
```
xiaoxmeng: Let's call it drop method and it shouldn't be static?

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,83 @@
...
+
+/// Global cache for hash tables shared across tasks within the same query.
+/// First task builds the table, subsequent tasks wait and reuse it.
+class HashTableCache {
+ public:
+  static HashTableCache* getInstance();
+
+  /// Gets or creates a cache entry. First caller becomes the builder.
+  /// Subsequent callers get a future to wait on.
+  std::shared_ptr<HashTableEntry> getOrAwait(
```
xiaoxmeng:
```
s/HashTableEntry/HashTableCacheEntry/
```

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+        queryPool->addLeafChild(fmt::format("cached_table_{}", key));
+  }
+}
+
+std::shared_ptr<HashTableEntry> HashTableCache::getOrAwait(
+    const std::string& key,
+    memory::MemoryPool* queryPool,
+    const std::string& taskId,
+    ContinueFuture* future) {
+  auto lockedTables = tables_.wlock();
```
xiaoxmeng: Let's have a lock for a HashTableCache object?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  auto it = lockedTables->find(key);
+  if (it == lockedTables->end()) {
+    // No entry exists - create a placeholder for this task to build the table.
+    auto entry = std::make_shared<HashTableEntry>();
+    entry->buildComplete.store(false, std::memory_order_release);
```
xiaoxmeng:
```
entry->buildComplete = false;
```

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+    getOrCreateTablePool(key, queryPool, entry);
+
+    lockedTables->insert({key, entry});
+    return entry; // Return entry with pool, table will be filled later
+  }
+
+  auto& entry = it->second;
+
+  // Ensure pool is created (in case entry was created without pool somehow)
+  getOrCreateTablePool(key, queryPool, entry);
```
xiaoxmeng: Is this required?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  // Check if build is complete
+  if (entry->buildComplete.load(std::memory_order_acquire)) {
+    return entry;
+  }
+
+  // Table is being built by another task - set up waiting if future provided
+  if (future != nullptr) {
```
xiaoxmeng: Shall we enforce future pointer is always set?

velox/exec/HashTableCache.cpp
xiaoxmeng:
```
s/getOrCreateTablePool/ensureCacheTablePool/
```

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+    const std::string& key,
+    memory::MemoryPool* queryPool,
+    std::shared_ptr<HashTableEntry>& entry) {
+  if (entry->tablePool == nullptr) {
+    entry->tablePool =
+        queryPool->addLeafChild(fmt::format("cached_table_{}", key));
+  }
+}
+
+std::shared_ptr<HashTableEntry> HashTableCache::getOrAwait(
```
xiaoxmeng: let's simply call it get with comments?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::shared_ptr<BaseHashTable> table,
+    bool hasNullKeys) {
+  std::vector<ContinuePromise> promises;
+
+  {
+    auto lockedTables = tables_.wlock();
+
+    auto it = lockedTables->find(key);
+
+    if (it == lockedTables->end()) {
```
xiaoxmeng: Let's add check here instead of tolerate this?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,144 @@
...
+          std::make_shared<HashTableEntry>(std::move(table), hasNullKeys);
+      lockedTables->insert({key, entry});
+      return;
+    }
+
+    auto& entry = it->second;
+
+    // Update the entry with the built table
+    entry->table = std::move(table);
+    entry->hasNullKeys = hasNullKeys;
```
xiaoxmeng, line 101: Is hasNullKeys required?

velox/exec/HashJoinBridge.h
shrinidhijoshi: This comment can be left as is because it trying to capture the behavior that `unique_ptr` is also accepted and will be cast to `shared_ptr` implicitly

velox/exec/HashBuild.cpp
shrinidhijoshi: Yup. Will fix. This is remnant from some refactoring I did to move code from initialize to constructor, so that makes this field redundant.

velox/exec/HashBuild.cpp
shrinidhijoshi: Current approach just takes hit=0 as miss so that we have one metric to show total hits . Are you suggesting we add 2 counters separate for hit and miss ?

velox/exec/HashBuild.cpp
shrinidhijoshi: You are right. This defensive check is redundant as we have
```
    VELOX_CHECK_NOT_NULL(cacheEntry_);
``` 
in the constructor

velox/exec/HashBuild.cpp
shrinidhijoshi: You are right. On second though, I have moved the `isUsingCachedHashTable` check into `canSpill` and got rid of this code

velox/exec/HashTableCache.cpp
shrinidhijoshi: Yeah. Sounds simpler. All callers are already passing non null

velox/exec/HashTableCache.cpp
shrinidhijoshi: IIUC, this is needed because the future tasks (cache hit) will need to info while setting the HT on the joinBridge ?

velox/exec/HashTableCache.cpp
shrinidhijoshi: Also I realized I introduced a bug while refactoring where only one driver was building Hashtable. Added extra check in get() call to not make other drivers wait, ensuring HashTable is built in parallel

velox/exec/HashTableCache.cpp
shrinidhijoshi: Makes sense. This is overkill. Will fix

velox/exec/HashTableCache.cpp
shrinidhijoshi: Good point. Cleaned up

velox/exec/HashTableCache.cpp
shrinidhijoshi: Makes sense

velox/exec/HashBuild.cpp
shrinidhijoshi: Yes. This one is to avoid calling `getOutput` of upstream operator

velox/exec/HashBuild.cpp
shrinidhijoshi: From previous comment, handling needsInput here might be late, as it will already have triggered getOutput on upstream operator (which we don't want) by driver framework

velox/docs/develop/joins.rst
```diff
@@ -198,6 +198,11 @@ the join is executed using broadcast or partitioned strategy has no effect on
 the join execution itself. The only difference is that broadcast execution
 allows for dynamic filter pushdown while partitioned execution does not.
 
+HashJoinNode supports a ``useHashTableCache`` flag (used only by Presto-on-Spark)
+that enables caching of the hash table built for broadcast joins. When enabled,
+the first task to build the hash table stores it in a global cache, and subsequent
+tasks reuse the cached table instead of rebuilding it.
```
xiaoxmeng:
```
tasks from the same query ...
```

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "velox/common/base/SpillConfig.h"
+#include "velox/exec/HashTable.h"
+#include "velox/exec/Spill.h"
+
+namespace facebook::velox::exec {
+
+/// Cached hash table entry with build coordination metadata.
+/// Note: Cached hash tables do not support spilling.
+struct HashTableCacheEntry {
```
xiaoxmeng, line 32:
```
struct HashTableCacheEntry {
  std::shared_ptr<BaseHashTable> table;
  bool hasNullKeys{false};
  std::shared_ptr<memory::MemoryPool> pool;
  std::string builderTaskId;
  tsan_atomic<bool> buildComplete{false};
  std::vector<ContinuePromise> buildPromises;
};
```

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,77 @@
...
+ public:
+  static HashTableCache* getInstance();
+
+  /// Gets or creates a cache entry. First caller becomes the builder.
+  /// Subsequent callers from different tasks get a future to wait on.
+  /// @param future Must be non-null; set if caller needs to wait.
+  std::shared_ptr<HashTableCacheEntry> get(
+      const std::string& key,
+      memory::MemoryPool* queryPool,
+      const std::string& taskId,
```
xiaoxmeng, line 63: put taskId ahead of query pool?

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,77 @@
...
+
+ private:
+  HashTableCache() = default;
+
+  folly::Synchronized<
+      std::unordered_map<std::string, std::shared_ptr<HashTableCacheEntry>>>
+      tables_;
+
+  static HashTableCache* instance_;
+  static std::mutex instanceMutex_;
```
xiaoxmeng: use folly_init once to initialize the singleton.

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,77 @@
...
+      std::shared_ptr<BaseHashTable> table,
+      bool hasNullKeys);
+
+  /// Removes a cache entry.
+  void drop(const std::string& key);
+
+ private:
+  HashTableCache() = default;
+
+  folly::Synchronized<
```
xiaoxmeng: I might use a lock to protect tables_ and cache entry mutations

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,138 @@
...
+    const std::string& taskId,
+    ContinueFuture* future) {
+  VELOX_CHECK_NOT_NULL(future, "future parameter must not be null");
+
+  auto lockedTables = tables_.wlock();
+
+  auto it = lockedTables->find(key);
+  if (it == lockedTables->end()) {
+    // No entry exists - create a placeholder for this task to build the table.
+    auto entry = std::make_shared<HashTableCacheEntry>();
```
xiaoxmeng:
```
std::make_shared<HashTableCacheEntry>(key, taskId, queryPool)
```

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // should proceed to build (they coordinate via JoinBridge, not here).
+  if (entry->builderTaskId == taskId) {
+    return entry;
+  }
+
+  // Table is being built by another task - set up waiting.
+  {
+    std::lock_guard<std::mutex> promiseLock(entry->promisesMutex);
+    // Double-check after acquiring lock
+    if (!entry->buildComplete) {
```
xiaoxmeng: We don't need this check if all apis are protected by hash table cache lock.

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,138 @@
...
+      entry->buildPromises.emplace_back(fmt::format("HashTableCache::{}", key));
+      *future = entry->buildPromises.back().getSemiFuture();
```
xiaoxmeng: use makeVeloxContinuePromiseContract?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,138 @@
...
+    auto lockedTables = tables_.wlock();
+
+    auto it = lockedTables->find(key);
+    VELOX_CHECK(
+        it != lockedTables->end(),
+        "Cache entry for key '{}' must be created by get() before put()",
+        key);
+
+    auto& entry = it->second;
+
```
xiaoxmeng:
```
VELOX_CHECK(!entry->buildComplete);
VELOX_CHECK_NULL(entry->table);
```

velox/exec/HashBuild.h
```diff
@@ -244,6 +266,19 @@ class HashBuild final : public Operator {
...
+  std::optional<std::string> cacheKey_;
```
xiaoxmeng:
```
std::string cacheKey_;
```

velox/exec/HashBuild.h
```diff
@@ -212,6 +213,27 @@ class HashBuild final : public Operator {
   // Invoked to abandon build deduped hash table.
   void abandonHashBuildDedup();
 
+  // Returns true if this operator is using a cached hash table.
+  // When enabled, the hash table is built once and cached for reuse
+  // by other tasks within the same query and stage.
+  bool isUsingCachedTable() const {
```
xiaoxmeng:
```
s/isUsingCachedTable/useHashTableCache/
```

velox/exec/HashBuild.h
```diff
@@ -212,6 +213,27 @@ class HashBuild final : public Operator {
...
+    return cacheKey_.has_value();
+  }
+
+  // Returns the hash table cache key for this operator.
+  // Only valid if isUsingCachedTable() returns true.
+  const std::string& getCacheKey() const {
```
xiaoxmeng:
```
s/getCacheKey/cacheKey() const 
```

velox/exec/HashBuild.h
```diff
@@ -212,6 +213,27 @@ class HashBuild final : public Operator {
...
+    VELOX_CHECK(
+        isUsingCachedTable(),
+        "getCacheKey() called when table caching is not enabled");
+    return cacheKey_.value();
+  }
+
+  // Determines the memory pool to use for the hash table.
+  // For cached hash tables, uses query-level pool so the table can
+  // outlive the task. For regular joins, uses operator pool.
+  memory::MemoryPool* getTableMemoryPool() const;
```
xiaoxmeng:
```
s/getTableMemoryPool/tableMemoryPool()/
```

velox/exec/HashBuild.cpp
```diff
@@ -75,6 +76,23 @@ HashBuild::HashBuild(
   VELOX_CHECK(pool()->trackUsage());
   VELOX_CHECK_NOT_NULL(joinBridge_);
 
+  // Compute cache key if this is a cacheable join (needed before setupTable
+  // to determine which memory pool to use).
+  if (joinNode_->useHashTableCache()) {
```
xiaoxmeng: initHashTableFromCache?

velox/exec/HashBuild.cpp
```diff
@@ -317,6 +354,13 @@ void HashBuild::removeInputRowsForAntiJoinFilter() {
 
 void HashBuild::addInput(RowVectorPtr input) {
   checkRunning();
+
+  // For hash table caching: check if table is already cached by another task.
```
xiaoxmeng:
```
VELOX_CHECK(!useHashTableCache() || entry->builderId == taskId);
```

velox/exec/HashBuild.cpp
```diff
@@ -698,6 +743,28 @@ bool HashBuild::finishHashBuild() {
     }
   };
 
+  // For hash table caching: check if table already exists in cache (from
+  // previous task). This check is done AFTER the barrier to ensure only the
+  // last driver handles it.
+  if (isUsingCachedTable()) {
+    if (cacheEntry_->buildComplete.load(std::memory_order_acquire)) {
```
xiaoxmeng:
```
if (cacheEntry_->buildComplete) {
```

velox/exec/HashBuild.cpp
xiaoxmeng: defined as static const members of the hash build operator as we do here: velox/exec/TableWriter.h?lines=153

velox/exec/HashBuild.cpp
xiaoxmeng:
```
if (getHashTableFromCache) {
   return true;
}
...
```

velox/exec/HashBuild.cpp
```diff
@@ -812,11 +879,29 @@ bool HashBuild::finishHashBuild() {
...
   }
-  joinBridge_->setHashTable(
-      std::move(table_),
-      std::move(spillPartitions),
-      joinHasNullKeys_,
-      std::move(tableSpillFunc));
+
+  // For hash table caching: the last driver caches the merged table.
+  if (isUsingCachedTable()) {
+    auto* cache = HashTableCache::getInstance();
```
xiaoxmeng:
```
setHastTableToCache()?
joinBridge_->setHashTable(
        std::move(table_),
        std::move(spillPartitions),
        joinHasNullKeys_,
        std::move(tableSpillFunc));
```

velox/exec/HashBuild.cpp
```diff
@@ -121,11 +123,74 @@ HashBuild::HashBuild(
...
+
+    // Record cache hit metric.
+    stats_.wlock()->addRuntimeStat(
+        BaseHashTable::kHashTableCacheHit, RuntimeCounter(1));
+
+    return true;
+  }
+
+  // Cache miss - we need to build the table.
+  stats_.wlock()->addRuntimeStat(
```
xiaoxmeng, line 189: nit: 
```
if (!cacheEntry_->buildComplete) {
   stats_.wlock()->addRuntimeStat(BaseHashTable::kHashTableCacheMiss, RuntimeCounter(1));
   return false;
}
 
/// hit case?
```

velox/exec/HashBuild.cpp
```diff
@@ -319,6 +384,14 @@ void HashBuild::removeInputRowsForAntiJoinFilter() {
 
 void HashBuild::addInput(RowVectorPtr input) {
   checkRunning();
+
+  VELOX_CHECK(!useHashTableCache() || cacheEntry_->builderTaskId == taskId());
+  // For hash table caching: check if table is already cached by another task.
+  // If so, skip processing this input - we'll use the cached table.
+  if (useHashTableCache() && cacheEntry_->buildComplete) {
```
xiaoxmeng: I think we shall never run into this? Let's remove or add check that !cacheEntry_->buildComplete

velox/exec/HashBuild.cpp
```diff
@@ -815,11 +893,21 @@ bool HashBuild::finishHashBuild() {
               spillStats);
         };
   }
+
+  // For hash table caching: the last driver caches the merged table.
+  std::shared_ptr<BaseHashTable> sharedTable = std::move(table_);
+  if (useHashTableCache()) {
```
xiaoxmeng: can we add a method say
```
setHashTableToCache to cover this?
```

velox/exec/HashBuild.cpp
```diff
@@ -815,11 +893,21 @@ bool HashBuild::finishHashBuild() {
               spillStats);
         };
   }
+
+  // For hash table caching: the last driver caches the merged table.
+  std::shared_ptr<BaseHashTable> sharedTable = std::move(table_);
```
xiaoxmeng:
```
s/sharedTable/table/
```

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,70 @@
...
+  bool hasNullKeys{false};
+  tsan_atomic<bool> buildComplete{false};
+  std::vector<ContinuePromise> buildPromises;
+};
+
+/// Global cache for hash tables shared across tasks within the same query.
+/// First task builds the table, subsequent tasks wait and reuse it.
+class HashTableCache {
+ public:
+  static HashTableCache* getInstance();
```
xiaoxmeng:
```
nit: s/getInstance/instance/
```

velox/exec/HashTableCache.h
```diff
@@ -0,0 +1,70 @@
...
+      bool hasNullKeys);
+
+  /// Removes a cache entry.
+  void drop(const std::string& key);
+
+ private:
+  HashTableCache() = default;
+
+  std::unordered_map<std::string, std::shared_ptr<HashTableCacheEntry>> tables_;
+  std::mutex cacheLock_;
```
xiaoxmeng: put lock ahead of its protected members?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,114 @@
...
+    memory::MemoryPool* queryPool,
+    ContinueFuture* future) {
+  VELOX_CHECK_NOT_NULL(future, "future parameter must not be null");
+
+  std::lock_guard<std::mutex> guard(cacheLock_);
+
+  auto it = tables_.find(key);
+  if (it == tables_.end()) {
+    // No entry exists - create a placeholder for this task to build the table.
+    auto entry = std::make_shared<HashTableCacheEntry>();
```
xiaoxmeng: Can we add ctor for HashTableCacheEntry?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  // If this is the builder task, don't wait - all drivers of the builder task
+  // should proceed to build (they coordinate via JoinBridge, not here).
+  if (entry->builderTaskId == taskId) {
+    return entry;
+  }
+
+  // Table is being built by another task - set up waiting.
+  // Double-check after acquiring lock
+  if (!entry->buildComplete) {
```
xiaoxmeng: no need to check this as we are protected by the lock?

velox/exec/HashTableCache.cpp
```diff
@@ -0,0 +1,114 @@
...
+    auto [promise, waitFuture] = makeVeloxContinuePromiseContract(
```
xiaoxmeng:
```
s/waitFuture/_future/
```

velox/exec/tests/HashJoinTest.cpp
```diff
@@ -8705,6 +8707,277 @@ DEBUG_ONLY_TEST_P(
   ASSERT_TRUE(spillTriggered.load());
 }
 
+// Non-parameterized test fixture for hash table caching tests.
+class HashTableCachingTest : public HashJoinTestBase {
```
xiaoxmeng: Consider to move HashJoinTestBase into a separate file say HashJoinTestBase.h and move HashJoinWithCacheTest into a separate test file.

velox/exec/tests/HashJoinTest.cpp
```diff
@@ -8705,6 +8707,277 @@ DEBUG_ONLY_TEST_P(
...
+
+    return AssertQueryBuilder(joinNode, duckDbQueryRunner_)
+        .queryCtx(queryCtx)
+        .maxDrivers(numDrivers)
+        .assertResults(
+            "SELECT t.t_k, t.t_v, u.u_k, u.u_v FROM t, u WHERE t.t_k = u.u_k");
+  };
+
+  // Run two tasks concurrently using async.
+  auto future1 = std::async(std::launch::async, runTask);
```
xiaoxmeng: Can we have a number of threads that continuously running for multiple iterations? Thanks!

velox/exec/HashBuild.cpp
```diff
@@ -815,11 +908,18 @@ bool HashBuild::finishHashBuild() {
               spillStats);
         };
   }
+
+  // For hash table caching: the last driver caches the merged table.
+  std::shared_ptr<BaseHashTable> table = std::move(table_);
+  if (useHashTableCache()) {
+    setHashTableInCache(table);
```
xiaoxmeng: Let's check useHashTableCache() inside setHashTableInCache? only set if it is true?

velox/exec/HashBuild.cpp
```diff
@@ -121,11 +123,92 @@ HashBuild::HashBuild(
...
+  return true;
+}
+
+void HashBuild::setHashTableInCache(std::shared_ptr<BaseHashTable> table) {
+  VELOX_CHECK(useHashTableCache());
+  auto* cache = HashTableCache::instance();
+  cache->put(cacheKey(), std::move(table), joinHasNullKeys_);
+}
+
+bool HashBuild::receiveCachedHashTable() {
```
xiaoxmeng:
```
s/receiveCachedHashTable/hasCachedHashTable or receivedCachedHashTable/
```

velox/exec/tests/HashJoinWithCacheTest.cpp
```diff
@@ -0,0 +1,307 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
```
xiaoxmeng, line 3: Rename to HashJoinWithCacheTest.cpp

velox/exec/tests/HashTableCachingTest.cpp
```diff
@@ -0,0 +1,307 @@
...
+  constexpr int kNumIterations = 5;
+  constexpr int kNumThreads = 10;
+  constexpr int kTotalTasks = kNumIterations * kNumThreads;
+
+  std::vector<std::shared_ptr<Task>> allTasks;
+  std::mutex tasksMutex;
+
+  for (int iter = 0; iter < kNumIterations; ++iter) {
+    // Each thread runs one task.
+    auto threadFunc = [&]() {
```
xiaoxmeng: As we discussed, we do multiple rounds inside each runTask. Thanks!

xiaoxmeng (approved):
@shrinidhijoshi thanks % minors. Please address before land.

velox/exec/tests/HashJoinWithCacheTest.cpp
```diff
@@ -0,0 +1,543 @@
...
+  const auto cacheKey = fmt::format("{}:{}", queryId, joinNodeId);
+  HashTableCache::instance()->drop(cacheKey);
+}
+
+// Tests that HashProbe cannot spill when using a cached hash table.
+// When useHashTableCache() is true, HashProbe::canSpill() returns false
+// because spilling would clear the cached table and corrupt it for other tasks.
+// This test verifies that HashProbe completes without spilling.
+TEST_F(HashJoinWithCacheTest, probeCannotSpillWithCachedTable) {
+  const std::string queryId =
```
xiaoxmeng, line 332: This test should use testvalue to check if canSpill() is set to false for both build and probe. Rename the test case to joinWithCachedTableCanNotSpill? This test will simply succeed no matter spill enabled or not?

PR #15755, opened 2025-12-12
refactor: Use std::move for std::function in PrestoSerializedPage ctor

lingbin:
This forces an unnecessary and potentially expensive copy of the  
`std::function` object. The cost is particularly high if the underlying  
callable object requires a heap allocation (e.g., a lambda with a  
large capture).  

No functional changes.


lingbin:
@pedroerp  Could you help review this PR? This is a simple change waiting for a review for a very long time.  
Already rebased, Thanks.

PR #15757, opened 2025-12-12
refactor(exchange): Rename ExchangeQueue::atEnd_ to noMoreInput_ for clarity

lingbin:
The `ExchangeQueue` class currently uses the name `atEnd` for both a
private member variable (`atEnd_`) and an output parameter in
`dequeueLocked()`, leading to potential ambiguity and causing confusion
for readers

- The member variable `atEnd_` signifies a **producer-side** state:
  no more input data will ever be enqueued. This happens when all
  exchange-sources complete or an error occurs. Note that the queue
  itself may still hold data to be consumed.

- The `atEnd` output parameter of `dequeueLocked()` signifies a
  **consumer-side** state: the queue is now empty and will remain
  so forever. It is the definitive signal for the consumer to stop
  dequeuing.

The overlap in naming is problematic. A developer reading `if (atEnd_)`
inside the implementation might mistakenly assume it's checking for the
consumer's end-of-stream condition, while it's actually checking if the
exchange sources have been exhausted(the queue itself may still have
data).

To resolve this confusion, this patch renames the member variable
`atEnd_` to `noMoreInput_`, which can unambiguously describes the
variable's true purpose: to indicate that the queue's input is
permanently closed and will receive no more new input data. This name
holds true for both normal completion and error scenarios.

The `atEnd` output parameter of `dequeueLocked()` is intentionally left
unchanged. Its name is conventional and idiomatic for consumer-facing
APIs that signal the end of a stream or collection. Once the internal
member variable's name is corrected, the parameter's name is no longer
ambiguous in context.

This is purely a naming refactor intended to improve code readability
and maintainability. No functional changes are introduced.

### DIFF
--- a/velox/exec/ExchangeQueue.cpp
+++ b/velox/exec/ExchangeQueue.cpp
@@ -27,7 +27,7 @@ void ExchangeQueue::noMoreSources() {
   {
     std::lock_guard<std::mutex> l(mutex_);
     noMoreSources_ = true;
-    promises = checkCompleteLocked();
+    promises = checkNoMoreInput();
   }
   clearPromises(promises);
 }
@@ -42,11 +42,12 @@ void ExchangeQueue::close() {
 }
 
 int64_t ExchangeQueue::minOutputBatchBytesLocked() const {
-  // always allow to unblock when at end
-  if (atEnd_) {
+  // Allow to unblock if no more input.
+  if (noMoreInput_) {
     return 0;
   }
-  // At most 1% of received bytes so far to minimize latency for small exchanges
+  // At most 1% of received bytes so far to minimize latency for small
+  // exchanges.
   return std::min<int64_t>(minOutputBatchBytes_, receivedBytes_ / 100);
 }
 
@@ -55,7 +56,7 @@ void ExchangeQueue::enqueueLocked(
     std::vector<ContinuePromise>& promises) {
   if (page == nullptr) {
     ++numCompleted_;
-    auto completedPromises = checkCompleteLocked();
+    auto completedPromises = checkNoMoreInput();
     promises.reserve(promises.size() + completedPromises.size());
     for (auto& promise : completedPromises) {
       promises.push_back(std::move(promise));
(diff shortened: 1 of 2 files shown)


velox/exec/ExchangeQueue.h
```diff
@@ -117,9 +119,9 @@ class ExchangeQueue {
     return clearAllPromisesLocked();
   }
 
-  std::vector<ContinuePromise> checkCompleteLocked() {
+  std::vector<ContinuePromise> trySetCompletedLocked() {
```
mbasmanova: checkCompleteLocked name seems to match the implementation better as this method checks if all sources have been drained? 

trySetCompletedLocked suggest that we are setting some 'completed' variable, but there is no such variable.

perhaps, keep the original name or rename to checkNoMoreInput?

velox/exec/ExchangeQueue.h
```diff
@@ -166,7 +168,10 @@ class ExchangeQueue {
   int numCompleted_{0};
   int numSources_{0};
   tsan_atomic<bool> noMoreSources_{false};
-  bool atEnd_{false};
+  // True if no more pages will be enqueued. This can be due to all sources
```
mbasmanova, line 171: Nice comment. Thanks.

velox/exec/ExchangeQueue.cpp
```diff
@@ -42,11 +42,12 @@ void ExchangeQueue::close() {
 }
 
 int64_t ExchangeQueue::minOutputBatchBytesLocked() const {
-  // always allow to unblock when at end
-  if (atEnd_) {
+  // Always allow to unblock when no more input data.
```
mbasmanova: drop 'data' from 'no more input data'.

// Allow to unblock if no more input.

lingbin: Done, thanks.

velox/exec/ExchangeQueue.h
lingbin: Thanks for your review.  I made this minor change(`checkCompleteLocked`--> `trySetCompletedLocked `) because the name `"checkXxx"` might lead people to think it's just a pure check (a const method with no side effects), but this method actually modifies state by setting `noMoreInput_` to `true` when conditions are met.

I have now renamed it to `checkNoMoreInput` as you suggested, thanks.

PR #11817, opened 2024-12-10
fix: Array_distinct under allocates Buffer for result elements with overlapping input arrays

kevinwilfong:
Summary:
When the input arrays to array_distinct have overlapping ranges of elements, it under counts the upper bound
of the cumulative size of the result arrays. Currently it uses the index of the last referred to element in the
elements Vector, which ignores duplicates in the case of overlapping ranges.

This change makes it so it sums the sizes of the input arrays instead. This should also prevent overallocating in
the case where a significant subset of the elements Vector is unused.

On a side note, there is a plan to ensure arrays do not overlap 
https://github.com/facebookincubator/velox/issues/10876 but until that is implemented and enforced, this fixes
a potential issue, and as mentioned, has the potential to save some memory at a small CPU cost in the case of
non-overlapping arrays.

Differential Revision: D67049306

PR #15765, opened 2025-12-13
misc: Check for overlapping ranges when validating Array/MapVectors

kevinwilfong:
Summary:
This would have made it easier to catch issues like the one fixed in
https://github.com/facebookincubator/velox/pull/15764

Differential Revision: D89099534


velox/functions/prestosql/tests/ArrayDistinctTest.cpp
```diff
@@ -432,43 +432,3 @@ TEST_F(ArrayDistinctTest, timestampWithTimezone) {
        std::nullopt},
       {pack(1, 0), pack(2, 1), pack(3, 2), std::nullopt, pack(4, 3)});
 }
-
-TEST_F(ArrayDistinctTest, overlappingRanges) {
```
bikramSingh91, line 436: just to confirm, we are removing this because we dont offer any guarantees in velox if arrays are overlapping, right?

kevinwilfong: That's right, I added this test about a year ago and noted in the PR that once we enforced that arrays are non-overlapping this would no longer be needed.

https://github.com/facebookincubator/velox/pull/11817

If I leave it in it simply fails validation of the ArrayVector

PR #15768, opened 2025-12-15
feat(exchange): Make Exchange client eager fetching behavior configurable

shrinidhijoshi:
Summary:
Currently, Exchange operator upon executing `addRemoteTaskId` will eagerly initiate exchange sources 
to start reading data.
But if the downstream operator is a HashBuild that is going to use a cached HT, this eager request is wasteful 
and can even cause excessive read I/O on the data source.

Adding a flag to turn off this behavior completely in cases when HT caching is going to be enabled.
For batch materialization systems, presto-on-spark for example, the eager exchange source fetching is
not of as much importance either way.

Differential Revision: D88900098

### DIFF
--- a/velox/exec/ExchangeClient.cpp
+++ b/velox/exec/ExchangeClient.cpp
@@ -53,7 +53,16 @@ void ExchangeClient::addRemoteTaskId(const std::string& remoteTaskId) {
       sources_.push_back(source);
       queue_->addSourceLocked();
       emptySources_.push(source);
-      requestSpecs = pickSourcesToRequestLocked();
+      // When lazyFetching_ is true, I/O will be triggered lazily when next() is
+      // called from Exchange::isBlocked(). This allows waiter tasks using
+      // cached hash tables to skip I/O entirely when the table is already
+      // cached - the HashBuild operator will finish before
+      // Exchange::isBlocked() is ever called, so no unnecessary data fetching
+      // occurs.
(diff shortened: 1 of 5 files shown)


velox/exec/ExchangeClient.h
```diff
@@ -189,6 +191,10 @@ class ExchangeClient : public std::enable_shared_from_this<ExchangeClient> {
   // when there is only one exchange source.
   const bool skipRequestDataSizeWithSingleSource_;
 
+  // If true, start fetching data immediately when remote tasks are added.
+  // If false, defer fetching until next() is called.
+  const bool eagerFetch_;
```
xiaoxmeng:
```
s/eagerFetch_/lazyFetch_/
```

velox/exec/ExchangeClient.cpp
```diff
@@ -53,14 +53,23 @@ void ExchangeClient::addRemoteTaskId(const std::string& remoteTaskId) {
...
+      // Exchange::isBlocked() is ever called, so no unnecessary data fetching
+      // occurs.
     }
   }
 
   // Outside of lock.
   if (toClose) {
     toClose->close();
-  } else {
+  } else if (!requestSpecs.empty()) {
```
xiaoxmeng: Let request() to handle the empty specs?

xiaoxmeng (approved):

PR #15769, opened 2025-12-15
refactor: Refactor SignatureBinder to use pointer for coercions and update related methods

MBkkt:
This PR refactors the SignatureBinder API to use optional pointer parameters for coercions instead of separate methods, and updates the type system to use CallableCost (uint64_t) instead of mixed integer types for cost tracking. The changes consolidate two separate methods (tryBind and tryBindWithCoercions) into a single method with an optional pointer parameter.

Key changes:
- Introduced CallableCost type alias (uint64_t) for consistent cost representation across the type coercion system
- Refactored SignatureBinder::tryBind to accept an optional std::vector<Coercion>* parameter instead of having separate methods
- Updated SimpleFunctionRegistry to use the new pointer-based API pattern
- Enhanced function signature validation to check that variadic signatures have at least one argument

### DIFF
--- a/velox/type/TypeCoercer.cpp
+++ b/velox/type/TypeCoercer.cpp
@@ -17,29 +17,33 @@
 
 namespace facebook::velox {
 
-int64_t Coercion::overallCost(const std::vector<Coercion>& coercions) {
-  int64_t cost = 0;
+CallableCost Coercion::overallCost(const std::vector<Coercion>& coercions) {
+  CallableCost cost = 0;
   for (const auto& coercion : coercions) {
     if (coercion.type != nullptr) {
       cost += coercion.cost;
     }
   }
-
   return cost;
 }
 
 namespace {
+
+// This is cost of CAST from UNKNOWN type to any other type.
+// This is the lowest for any implicit CAST.
+// Any other implicit CAST will have cost higher than this.
+constexpr CallableCost kNullCoercionCost = 1;
 
 std::unordered_map<std::pair<std::string, std::string>, Coercion>
 allowedCoercions() {
   std::unordered_map<std::pair<std::string, std::string>, Coercion> coercions;
 
   auto add = [&](const TypePtr& from, const std::vector<TypePtr>& to) {
-    int32_t cost = 0;
+    auto cost = kNullCoercionCost;
     for (const auto& toType : to) {
       coercions.emplace(
           std::make_pair<std::string, std::string>(
-              from->kindName(), toType->kindName()),
+              from->name(), toType->name()),
           Coercion{.type = toType, .cost = ++cost});
     }
   };
@@ -52,6 +56,7 @@ allowedCoercions() {
 
   return coercions;
 }
+
 } // namespace
 
 // static
@@ -60,7 +65,7 @@ std::optional<Coercion> TypeCoercer::coerceTypeBase(
     const std::string& toTypeName) {
   static const auto kAllowedCoercions = allowedCoercions();
   if (fromType->name() == toTypeName) {
-    return Coercion{.type = fromType, .cost = 0};
+    return Coercion{fromType, 0};
   }
 
   auto it = kAllowedCoercions.find({fromType->name(), toTypeName});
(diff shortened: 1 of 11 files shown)


MBkkt:
Don't forget to use "Hide whitespace" feature
image

velox/type/TypeCoercer.h
```diff
@@ -19,10 +19,12 @@
 
 namespace facebook::velox {
 
+using CallableCost = uint64_t;
```
mbasmanova, line 22: Let's remove this change from this PR.

We need more discussion about this type. CallableCost name suggests that this value represents the cost of calling a function, however, this is used to rank function signatures. This is confusing.

velox/type/TypeCoercer.h
```diff
@@ -19,10 +19,12 @@
...
+
 /// Type coercion necessary to bind a type to a signature.
 struct Coercion {
   TypePtr type;
-  int32_t cost{0};
+  CallableCost cost = 0;
```
mbasmanova, line 27: {0} seems better. Why did you change to = 0?

velox/type/TypeCoercer.cpp
```diff
@@ -17,29 +17,33 @@
...
-
   return cost;
 }
 
 namespace {
 
+// This is cost of CAST from UNKNOWN type to any other type.
+// This is the lowest for any implicit CAST.
+// Any other implicit CAST will have cost higher than this.
+constexpr CallableCost kNullCoercionCost = 1;
```
mbasmanova, line 35: These changes seem unrelated to changing tryBind APIs? Please, do not mix multiple changes in a single PR.

velox/type/TypeCoercer.h
MBkkt: We need to use **same type** otherwise it's just doesn't make sense and can lead to unintentional overflows in some places
I don't see the point of removing this.

velox/type/TypeCoercer.h
MBkkt: I think I changed it because of habits, when change type, I think one assign symbol is more readable when two brackets.

Same reason why almost no one wrote
```
auto localFoo{makeFoo()};
```

If you insist I can change this back

velox/type/TypeCoercer.cpp
MBkkt: > Please, do not mix multiple changes in a single PR.

Thanks for the review. I agree that small, focused PRs are ideal, but I need to explain why the current expectation creates a fundamental barrier for teams/contributors like we're.

For us, developing a complete feature (like the recent coercion logic) efficiently results in a themed PR of ~500 LOC. Requiring us to split this into 3-4 sequential, micro-PRs for upstream multiplies the work for the same outcome. I simply cannot justify this cost to our team. Our choice becomes: stop trying to contribute back or operate at an unsustainable pace.

I believe difficulties in contribution (this, internal CI, no public sanitizers, etc) are a one of main reasons why projects like DuckDB, ClickHouse, and DataFusion are developing significantly faster. Their contribution processes are designed to accept almost any useful PRs from externals. Sometimes the maintainers there work with contributors to refine and split changes after the functional value is agreed upon. This lowers the barrier to entry and speeds up overall progress.

In Velox/Axiom it looks quite opposite, almost any code can be contributed from Meta internals (I can provide really unique examples if you wish), and at the same time to contribute from outside people quite often needs to very long list of time consuming requirements.

It's not something bad, but it's opposite approach.

After some thinking how we can find compromise where contributors like us can contribute in Velox and Velox still follow it's contribution model, I want suggest to try such approach:

1. Externals contributors submit a complete, tested feature bundle as a Draft PR, it's probably commonly can be separated into multiple separate PR. But such separation probably will require careful extraction and rewriting, in other words non zero time.
2. The review focuses on: "Is this functionality correct and valuable for Velox?" Not in about how to split this.
3. Then, a maintainer or contributor who wants the feature in upstream most takes the lead on restructuring the bundle into the sequence of isolated PRs that fit Velox's pipeline.

I am convinced that for Velox/Axiom to thrive as a community-driven project, not just an internal Meta project, it needs a contribution model that works for companies of all sizes. The success of other similar projects proves this is possible.


WDYT?

MBkkt:
A lot of conflicts was introduced, so I close it

PR #15771, opened 2025-12-15
fix: Add missing include in Iceberg test

pedroerp:
The missing include broke builds in a fedora clang17 laptop.


PingLiuPing (approved):
Thanks @pedroerp .

Issue #15776, opened 2025-12-15
Upgrade FB_OS_VERSION > "v2025.04.28.00"


kgpai:
### Description

The current FB_OS_VERSION is more than 6 months old. Notably it fails to build on latest Mac's with Tahoe (issue : #15728) . 
Creating this issue to track upgrading FB_OS_VERSION , which we believe will also fix the mac build issues.

kgpai:
cc: @mbasmanova @czentgr @majetideepak @pratikpugalia @peterenescu @assignUser

mbasmanova:
CC: @pedroerp

mbasmanova:
Tried upgrading to v2025.04.28.00, but got build failures: https://github.com/facebookincubator/velox/issues/15728#issuecomment-3657580668

PR #15777, opened 2025-12-16
feat: Update FB_OS_VERSION

kgpai:
Summary:
Update FB_OS_VERSION . 
As per issue : https://github.com/facebookincubator/velox/issues/15776 .

Differential Revision: D89233855

### DIFF
--- a/CMake/resolve_dependency_modules/folly/CMakeLists.txt
+++ b/CMake/resolve_dependency_modules/folly/CMakeLists.txt
@@ -17,10 +17,11 @@ cmake_minimum_required(VERSION 3.28)
 velox_set_source(FastFloat)
 velox_resolve_dependency(FastFloat CONFIG REQUIRED)
 
-set(VELOX_FOLLY_BUILD_VERSION v2025.04.28.00)
+set(VELOX_FOLLY_BUILD_VERSION v2025.12.15.00)
+# TODO: Update checksum after verifying the new release tarball
 set(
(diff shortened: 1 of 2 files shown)


kgpai:
I am not exactly sure what breaks with the new FB_OS_VERSION, so created this PR to find out. We can add fixes here. 

cc: @pratikpugalia @peterenescu @mbasmanova @czentgr

CMake/resolve_dependency_modules/folly/CMakeLists.txt
```diff
@@ -17,10 +17,11 @@ cmake_minimum_required(VERSION 3.28)
 velox_set_source(FastFloat)
 velox_resolve_dependency(FastFloat CONFIG REQUIRED)
 
-set(VELOX_FOLLY_BUILD_VERSION v2025.04.28.00)
+set(VELOX_FOLLY_BUILD_VERSION v2025.12.15.00)
```
mbasmanova, line 20: @majetideepak @kgpai @czentgr Why do we need this change? Where can I get the hash for VELOX_FOLLY_BUILD_SHA256_CHECKSUM for v2025.06.30.00?

kgpai: You will have to download the release binary from folly releases and run sha256 checksum to get the hash. We also want the latest folly since there is an ask from folly to move away from folly::hardware_concurrency to folly::available_concurrency, so it makes sense to get it to work with latest releases.

PR #15782, opened 2025-12-16
refactor(geo): Move GeometrySerde to common/geospatial

jagill:
Summary:
This is a better place to import it from for SpatialJoinBuilder,
which lives in exec.

Differential Revision: D89239583


velox/functions/prestosql/aggregates/CMakeLists.txt
```diff
@@ -79,7 +79,11 @@ velox_link_libraries(
 
 if(${VELOX_ENABLE_GEO})
   velox_add_library(velox_aggregates PRIVATE GeometryAggregate.cpp)
```
unidevel, line 81: Is this correct? should it be changed to `target_sources`?

PR #15789, opened 2025-12-17
fix: Prevent integer overflow in PrefixSort memory estimation

zhli1142015:
Fix an integer overflow bug in `PrefixSort::maxRequiredBytes()` that causes
incorrect memory estimation, preventing spill from triggering and leading to OOM.

**Problem**
The return type of `PrefixSort::maxRequiredBytes()` is `uint32_t`, which can only
represent up to ~4GB. When sorting large datasets (e.g., 1 billion rows), the
actual memory requirement (~24GB for prefix buffer) overflows, returning a much
smaller value (~2.35GB).

This causes `ensureSortFits()` to underestimate memory needs. The reservation
succeeds with the incorrect small value, so spill is not triggered. When
`PrefixSort::sort()` later attempts to allocate the actual memory required,
the operator is already outside the `ReclaimableSectionGuard` scope and cannot
be reclaimed, resulting in OOM.

**Fix**
Change the return type of `PrefixSort::maxRequiredBytes()` from `uint32_t` to
`uint64_t`.

### DIFF
--- a/velox/exec/PrefixSort.cpp
+++ b/velox/exec/PrefixSort.cpp
@@ -310,7 +310,7 @@ void PrefixSort::extractRowAndEncodePrefixKeys(char* row, char* prefixBuffer) {
 }
 
 // static.
-uint32_t PrefixSort::maxRequiredBytes(
+uint64_t PrefixSort::maxRequiredBytes(
     const RowContainer* rowContainer,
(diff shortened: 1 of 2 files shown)


pedroerp (approved):
Thank you

duanmeng (approved):
LGTM % thanks.

PR #15790, opened 2025-12-17
[WIP] prototype: Add Avro reader

izchen:
Implemented a native Avro reader based on Apache avro-cpp (release-1.12.1).


velox/connectors/hive/HiveConnectorUtil.cpp
```diff
@@ -604,6 +604,8 @@ void configureReaderOptions(
 
     readerOptions.setFileFormat(hiveSplit->fileFormat);
   }
+
+  readerOptions.serDeOptions().parameters = hiveSplit->serdeParameters;
```
Yuhta: We don't forward this blindly

Yuhta:
This is a large feature that needs maintainers.  Do we have candidate for maintainers for this?

izchen:
> This is a large feature that needs maintainers. Do we have candidate for maintainers for this?

Thanks for raising this.

This feature is actively used and maintained by our team internally. 
I will be the primary point of contact and maintainer, and we have multiple engineers familiar with the code who can provide backup support.

velox/connectors/hive/HiveConnectorUtil.cpp
izchen: Good point. I've updated this so that only a whitelisted set of keys is forwarded, and only when `format == AVRO`.

izchen:
The `Ubuntu Benchmark` CI failed due to a false positive from static analysis, which is the same issue as reported in #15245. 
I plan to add `VELOX_SUPPRESS_STRINGOP_OVERFLOW_WARNING` to fix it.

izchen:
The `Collect Build Metrics` and `Run Checks / clang-tidy` CI jobs failed because avro-cpp is not installed in the `ghcr.io/facebookincubator/velox-dev:adapters` image. (On Ubuntu, it can be installed via `./scripts/setup-ubuntu.sh install_avro`)

mbasmanova:
> This feature is actively used and maintained by our team internally.

@izchen Welcome to the community. Which company are you from?

mbasmanova:
> based on avro-cpp (release-1.12.1).

Can we avoid introducing a new dependency and instead implement the reader in Velox directly? Dependency management is hard and makes building Velox difficult. 

CC: @pedroerp

izchen:
> Welcome to the community. Which company are you from?

Thanks! I'm from xiaohongshu.com

We've been running Velox in production for about 1.5 years to accelerate Spark workloads and have seen significant compute cost savings. We appreciate the community and look forward to contributing back.

izchen:
> > based on avro-cpp (release-1.12.1).
> 
> Can we avoid introducing a new dependency and instead implement the reader in Velox directly? Dependency management is hard and makes building Velox difficult.
> 
> CC: @pedroerp

@mbasmanova Agreed — dependency management is painful.

Avro itself is fairly complex:

- avro-cpp implementation is ~19k LOC (C++ + headers, excluding blanks/comments)
- Avro spec contains many implicit rules, especially around schema parsing and schema evolution.

Using avro-cpp helps us avoid re-implementing the Avro spec and reduces the risk of diverging from upstream behavior.

avro-cpp is also actively maintained. In 2025, the Apache Avro community merged 36 commits related to avro-cpp. A fully Velox-native Avro reader would be a significant development and maintenance effort.

Given this, we believe avro-cpp is the better choice. We’ve already mitigated the impact by:

- keeping the dependency optional (VELOX_ENABLE_AVRO via EXTRA_CMAKE_FLAGS, default OFF)
- providing an install_avro helper script.

Open to further discussion.

mbasmanova:
@izchen Thank you for clarifying. What are the limitations of use avro-cpp? How are we tracking memory usage? Are filter pushdown and lazy loading supported? Are we forced to copy data from the format produced by the reader into Velox vectors? Is it feasible to address these?

This PR is quite large. Is there a way to break it up into smaller pieces?

CI is red. Are you working on fixing it?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+  if (schemaIt == serdeParameters.end() || schemaIt->second.empty()) {
+    return std::nullopt;
+  }
+
+  ::avro::ValidSchema schema;
+  std::istringstream schemaStream(schemaIt->second);
+  try {
+    ::avro::compileJsonSchema(schemaStream, schema);
+  } catch (const std::exception& e) {
+    VELOX_USER_FAIL(
```
mbasmanova: Why are we swallowing this exception?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+        "Failed to parse '{}' value '{}': {}",
+        kAvroSchemaLiteralKey,
+        it->second,
+        e.what());
+  }
+  VELOX_USER_CHECK_GT(
+      parsedValue,
+      0,
+      "{} must be a positive integer, got {}",
```
mbasmanova: no need to include parsedValue in the message; it is included automatically

velox/dwio/common/Options.h
```diff
@@ -106,6 +108,7 @@ class SerDeOptions {
   bool lastColumnTakesRest;
   uint8_t escapeChar;
   bool isEscaped;
+  std::unordered_map<std::string, std::string> parameters;
```
Yuhta: Should just use `RowReaderOptions::serdeParameters`

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+  const auto schemaIt =
+      serdeParameters.find(std::string(kAvroSchemaLiteralKey));
+  if (schemaIt == serdeParameters.end() || schemaIt->second.empty()) {
+    return std::nullopt;
+  }
+
+  ::avro::ValidSchema schema;
+  std::istringstream schemaStream(schemaIt->second);
+  try {
+    ::avro::compileJsonSchema(schemaStream, schema);
```
Yuhta: Why do we need to accept external schema separate from file?  Is it the evolved current table schema?  If so, you need to get the requested types as a Velox type (`RowReaderOptions::requestedType`) and construct the Avro schema inside the reader.

izchen:
Thanks for the detailed comments! I’ll follow up on these. It may take some time.

izchen:
@mbasmanova Apologies for the delayed response.


### **How are we tracking memory usage?**


Apache avro-cpp does not provide visibility into or control over memory usage during reads. It does not offer precise memory accounting, nor does it support injecting a custom allocator or memory pool.

In practice, its memory behavior is predictable based on the implementation:

- **O(1) state**: schema, decoder state, current row data, and stream sync marker
- **Block buffers**: compressed_ and uncompressed, sized by the current Avro block (typically ~64 KB by default), i.e. **O(B)**
- **Decompression internals**: temporary buffers and state from zlib / zstd / snappy, also bounded by **O(1) or O(B)**

Overall, the transient peak memory usage is roughly the sum of the compressed and decompressed block sizes, plus small constant overheads. With default block sizes, this stays within a small multiple of the avro block size.



### **Are filter pushdown and lazy loading supported?**

Avro is a row-based format and does not provide block-level statistics or indexing . As a result, filtering and column access happen after data is read and decoded, so physical filter pushdown and lazy loading are not natively supported.



### **Are we forced to copy data from the format produced by the reader into Velox vectors?**

This is a valid concern. This is indeed a limitation of using avro-cpp. 

Avro-cpp’s reader APIs generally materialize values into intermediate objects, which introduces an extra copy or object construction step.

For fixed-width scalar types this is usually negligible, but for variable-length types such as string and bytes the impact is higher due to allocations and buffer copies.

We plan to optimize this by projecting the Avro schema from the Velox requested schema and constructing the reader with the projected schema. This is not storage-level column pruning and does not reduce decompression, but it avoids decoding and materializing unused variable-length fields and reduces the memory footprint of the resulting Velox vectors.



### **This PR is quite large. Is there a way to break it up into smaller pieces?**

Yes. I plan to split this PR into three smaller ones:

1. Build and dependency setup
2. Metadata handling, including Velox–Avro schema mapping and schema projection
3. Data processing, writing decoded Avro data into Velox vectors



### **CI is red. Are you working on fixing it?**


CI jobs failed because avro-cpp is not installed in the `ghcr.io/facebookincubator/velox-dev:adapters` image. This issue goes away once the PR is split and the changes are merged in order.

velox/dwio/avro/reader/AvroReader.h
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+#pragma once
+
+#include <memory>
+#include <optional>
+
+#include "velox/dwio/common/BufferedInput.h"
+#include "velox/dwio/common/Reader.h"
+#include "velox/dwio/common/ReaderFactory.h"
```
gggrace14: Are we able to minimize the includes in a header file and replace them with forward declaration of the actual classes we need just as you do below?

velox/dwio/avro/reader/AvroReader.h
```diff
@@ -0,0 +1,96 @@
...
+  std::shared_ptr<AvroFileContents> contents_;
+  dwio::common::RowReaderOptions options_;
+  std::unique_ptr<::avro::DataFileReader<::avro::GenericDatum>> reader_;
+  std::unique_ptr<::avro::GenericDatum> datum_;
+  uint64_t currentRow_;
+  bool atEnd_;
+  std::optional<int64_t> splitLimit_;
+  uint64_t rowSize_;
+  uint64_t estimatedRowVectorSize_;
+  bool readerPastSync() const;
```
gggrace14: We need to declare those data members const that are not mutated by the class member functions. Candidates are options_ and splitLimit_l

gggrace14:
> ### **This PR is quite large. Is there a way to break it up into smaller pieces?**
> Yes. I plan to split this PR into three smaller ones:
> 
> 1. Build and dependency setup
> 2. Metadata handling, including Velox–Avro schema mapping and schema projection
> 3. Data processing, writing decoded Avro data into Velox vectors


Hi @izchen , thank you for clarification and addressing the requests.

Are you working on splitting this PR into three as you mentioned?

@Yuhta  and me will continue to review, given our availability. We could continue to comment here, once you have them. Just to let you know.

Separately, one question. Learned from you the avro-cpp dep we're introducing is under active development. So this Avro reader component needs help from you and your coworker as component maintainer, esp. for maintaining the dep and upgrades, since we don't have active traffic to this component that could verify changes and upgrades. Does that sound good to you?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+      std::move(input),
+      options,
+      std::move(schema),
+      std::move(typeInfo),
+      std::move(reader),
+      avroScanBatchBytes);
+}
+
+std::optional<uint64_t> AvroReader::numberOfRows() const {
+  return std::nullopt;
```
gggrace14, line 1750: I suppose TableScan and certain DataSource will be the caller of AvroReader. If we don't implement numberOfRows() here, how will we populate the stats of input rows of TableScan?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+    return kDefaultAvroScanBatchBytes;
+  }
+
+  uint64_t parsedValue = 0;
+  try {
+    parsedValue = folly::to<uint64_t>(it->second);
+  } catch (const folly::ConversionError& e) {
+    VELOX_USER_FAIL(
+        "Failed to parse '{}' value '{}': {}",
+        kAvroSchemaLiteralKey,
```
gggrace14: I think you mean kAvroScanBatchBytesKey here. Copy error?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+        it->second,
+        e.what());
+  }
+  VELOX_USER_CHECK_GT(
+      parsedValue,
+      0,
+      "{} must be a positive integer, got {}",
+      kAvroSchemaLiteralKey,
```
gggrace14: I think you mean kAvroScanBatchBytesKey here. Copy error?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+    VectorPtr& result,
+    const dwio::common::Mutation* mutation) {
+  if (atEnd_ || size == 0) {
+    return 0;
+  }
+  const auto rowsToRead = nextReadSize(size);
+  SelectivityVector rows(rowsToRead);
+  if (result &&
+      (!result->type()->equivalent(*contents_->rowType) ||
+       result->size() != size)) {
```
gggrace14: Do we need to reset result and re-allocate memory every time result->size() != size? Is it too expensive? I think ensureWritable() includes result resize.

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+    writeDatum(resolved, writer.current());
+    writer.commit(true);
+    ++numRead;
+    ++currentRow_;
+  }
+  writer.finish();
+  rowVector->resize(numRead);
+  if (numRead > 0) {
+    const auto batchBytes = rowVector->estimateFlatSize();
+    rowSize_ += numRead;
```
gggrace14: Are currentRow_ and rowSize_ always the same? Is it possible to get rid of one of the two?

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,974 @@
...
+      for (const auto& [key, valueDatum] : entries) {
+        auto resolvedValue =
+            resolveDatum(*resolvedInfo.children.front(), valueDatum);
+        if (!resolvedValue.has_value()) {
+          auto& keyWriter = mapWriter.add_null();
+          keyWriter.copy_from(key);
+          continue;
+        }
+
+        auto [keyWriter, valueWriter] = mapWriter.add_item();
```
gggrace14: `auto&& [keyWriter, valueWriter] = mapWriter.add_item();` is better?

izchen:
Hi @gggrace14 and @Yuhta , thanks for the update and for continuing the review.

Yes — I’ve prepared the PR split as mentioned. During the split, I also did some minor refactoring, added more comments for readability, and double-checked reader behaviors against the Avro specification (https://avro.apache.org/docs/1.12.0/specification/) to keep the implementation aligned.

For now, I pushed the changes as three commits in this PR. If this overall direction looks aligned with your expectation, I will proceed to open three separate PRs in the same order.

On maintenance: sounds good to me. I’ll take ownership of maintaining this Avro reader component long-term, especially for dependency upgrades and follow-ups as avro-cpp continues to evolve. My coworker will also help as needed.

Thanks again for your support and review time!

izchen:
Hi @gggrace14 and @Yuhta, just following up here — wanted to make sure this didn’t get lost.

Whenever you have time, could you please take a look at the latest updates?

Happy to make further adjustments if needed. Thanks again!

Yuhta:
@izchen We can start by separate out and merge the build infrastructure first, the more formal review of code can be started after that.  For now my main question is about why do we need to pass a second schema in addition to the output type.

izchen:
> For now my main question is about why do we need to pass a second schema in addition to the output type.

@Yuhta, Thank you for raising this question.

According to the Avro specification, schemas are defined in JSON format, and each Avro data file stores its writer schema alongside the data. (https://avro.apache.org/docs/1.12.0/specification/#schema-declaration)

The specification also defines Schema Resolution, which describes how differences between the reader schema and the writer schema are handled. When the two schemas are not identical, Avro applies a well-defined set of resolution rules, including field matching, default value injection, type promotion, and field renaming. (https://avro.apache.org/docs/1.12.0/specification/#schema-resolution)

On the consumer side, Hive and Spark support Avro Schema Resolution as well. Through table properties, users can explicitly provide a reader schema that will be used to read all Avro files in the table. This backward-compatibility mechanism is commonly used in practice when schemas evolve over time.
 (https://hive.apache.org/docs/latest/user/avroserde/#specifying-the-avro-schema-for-a-table)
 (https://spark.apache.org/docs/latest/sql-data-sources-avro.html#data-source-option)

Yuhta:
@izchen The reader schema is already provided with `RowReaderOptions::requestedType`, adding extra side way to pass it would cause duplication and inconsistency if they end up different.  Please construct the reader schema using `RowReaderOptions::requestedType` instead of custom property.

izchen:
> @izchen The reader schema is already provided with `RowReaderOptions::requestedType`, adding extra side way to pass it would cause duplication and inconsistency if they end up different. Please construct the reader schema using `RowReaderOptions::requestedType` instead of custom property.

@Yuhta Thanks for the comment, and sorry for the late reply.

I updated the behavior to make the Avro reader respect `RowReaderOptions::requestedType` and `scanSpec`:

- If `RowReaderOptions::requestedType` is provided, the Avro reader derives the target read schema from it, recursively aligns it with the file schema by field name, and validates the types. If it is not provided, the reader falls back to the file schema.
- If `RowReaderOptions::scanSpec` is provided, the Avro reader respects its projection, filter, constant, and mutation semantics, including pruning and filtering inside nested row / array / map fields.
- If both `requestedType` and `scanSpec` are provided, `requestedType` constrains the readable fields and types, while `scanSpec` determines the final output fields, ordering, filter-only fields, and constant fields.

I have also rebased the code onto the latest main.

As discussed earlier, I will split this large PR into smaller ones. The first one is here: https://github.com/facebookincubator/velox/pull/16555. Could you please take a look when you have a chance? I’m happy to make further adjustments as needed.

izchen:
@zhouyuan Thanks for the attention here.

Just a quick note that I have updated the PR. I’d appreciate it if you could also take a look when you have a chance.

zhouyuan:
@izchen thanks for the update, I ran into this gap when doing some iceberg test 
Cc: @rui-mo

pedroerp:
@izchen looks like this one is now waiting on your updates. Are you still interesting in finalizing this PR? Avro support would indeed be a nice feature to add to Velox.

Yuhta:
@pedroerp The actual merging should be broken down to smaller PRs, we can either close or put this PR in "Request changes" mode

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,1792 @@
...
+    *len = static_cast<size_t>(size);
+    pushback_ = 0;
+    return true;
+  }
+
+  void backup(size_t len) override {
+    if (pushback_ > 0) {
+      const void* rawData = nullptr;
+      int32_t size = 0;
+      stream_->Next(&rawData, &size);
```
jja725: what's the expected behavior if we reach EOF and return negative here.

velox/dwio/avro/reader/AvroReader.h
```diff
@@ -0,0 +1,93 @@
...
+      const override;
+
+  std::unique_ptr<dwio::common::RowReader> createRowReader(
+      const dwio::common::RowReaderOptions& options = {}) const override;
+
+ private:
+  std::shared_ptr<AvroFileContents> contents_;
+};
+
+class AvroRowReader : public dwio::common::RowReader {
```
jja725, line 58: can we add some comments

velox/dwio/avro/reader/AvroReader.cpp
```diff
@@ -0,0 +1,1792 @@
...
+  BaseVector::ensureWritable(rows, rowType, &contents_->pool, result);
+  auto rowVector = std::static_pointer_cast<RowVector>(result);
+  rowVector->resize(rowsToRead);
+  exec::VectorWriter<Any> writer;
+  writer.init(*rowVector);
+  const auto* rootInfo = readSchema_->typeInfo.get();
+
+  vector_size_t numRead = 0;
+  while (numRead < rowsToRead) {
+    if (reader_->pastSync(splitLimit_) || !reader_->read(*datum_)) {
```
jja725, line 1905: is there a plan to do vector reader

jja725 (changes requested):
Thanks for this great contribution and it's wanted within Uber as well! We would test it out internally so see if there's any gap/fix we can contribute as well. Also Agree we want seperate PR to make review easier.

izchen:
@pedroerp @Yuhta @jja725 Thanks for taking a look. I’ll update this PR later this week.

izchen:
I reviewed and fixed a few correctness issues, and rebased the changes onto the latest main branch. I’ll split this PR into smaller ones shortly and create an issue to track them.

PR #15794, opened 2025-12-17
feat(function): Support 10 key-value pairs in Spark map function

miaoever:
Summary: Increased the maximum number of key-value pairs supported by the Spark SQL map function from 3 to 10 (20 arguments total) as we have some use  cases using up to 10 key-value pairs.

Differential Revision: D89391837

### DIFF
--- a/velox/functions/sparksql/tests/MapTest.cpp
+++ b/velox/functions/sparksql/tests/MapTest.cpp
@@ -149,6 +149,69 @@ TEST_F(MapTest, wide) {
       mapVector);
 }
 
+TEST_F(MapTest, tenKeyValuePairs) {
+  // Test map function with 10 key-value pairs (20 arguments).
+  auto key1 = makeFlatVector<int64_t>({1, 11, 21});
+  auto val1 = makeFlatVector<std::string>({"a", "aa", "aaa"});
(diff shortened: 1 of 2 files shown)


velox/functions/sparksql/tests/MapTest.cpp
```diff
@@ -149,6 +149,69 @@ TEST_F(MapTest, wide) {
       mapVector);
 }
 
+TEST_F(MapTest, tenKeyValuePairs) {
+  // Test map function with 10 key-value pairs (20 arguments).
+  auto key1 = makeNullableFlatVector<int64_t>({1, 11, 21});
```
rui-mo: nit: use `makeFlatVector`

rui-mo (commented):
Thanks.

velox/functions/sparksql/tests/MapTest.cpp
miaoever: Thanks @rui-mo, updated to use `makeFlatVector`.

rui-mo (approved):

velox/functions/sparksql/tests/MapTest.cpp
rui-mo: Thanks for the prompt update. I'll mark this as ready-to-merge once the workflow passes.

PR #15798, opened 2025-12-18
feat: Add biased NULL pattern generation to VectorFuzzer

pratikpugalia:
Summary:
This diff extends VectorFuzzer with biased NULL distribution patterns to improve fuzzing coverage of edge cases.

**Changes:**
- Updated `fuzzNulls()` to generate NULLs with 4 modes: `kRandom`(70%), `kHeadOnly`(10%), `kTailOnly`(10%), `kHeadAndTail`(10%). 
- Modified `fuzzFlat()` to use the pattern-aware `fuzzNulls()` instead of inline random NULL generation

**Why:**
Random NULL distribution may miss edge cases where NULLs are concentrated at specific positions (beginning, end, or both edges). These patterns can expose bugs emerging from these patterns.

Differential Revision: D89227420


kgpai (approved):
Looks good, thanks !

PR #15809, opened 2025-12-18
feat(function): Add varbinary variants for strpos and contains (#15809)

laurencer:
Summary:

Adds `strpos`, `strrpos` and `contains` for varbinary columns which will allow for efficient string searching in encoded text data where we don't want to pay the charset decoding price (i.e. we don't have to convert the column to utf-8 and just do byte matches as an approximation).

Reviewed By: kaikalur

Differential Revision: D89479354

### DIFF
--- a/velox/docs/functions/presto/binary.rst
+++ b/velox/docs/functions/presto/binary.rst
@@ -6,6 +6,10 @@ Binary Functions
 
     Computes the crc32 checksum of ``binary``.
 
+.. function:: contains(x, element) -> boolean
+
+     Returns true if the binary ``x`` contains the ``element``.
+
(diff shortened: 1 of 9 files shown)


zacw7 (changes requested):
could you pls add the doc here: https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/presto/binary.rst

laurencer:
> could you pls add the doc here: https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/presto/binary.rst

@zacw7 added docs that match the string and array variants

zacw7 (approved):
LGTM. Thanks for adding the functions.

PR #15812, opened 2025-12-18
refactor(text-reader): String usage cleanup

pedroerp:
Summary:
Some clean up on text reader usage of strings:
* Remove usage of StringPiece
* Use std::string_view instead of StringView where appropriate
* Avoid unecessary std::string allocations.
* Other cosmetic and formatting issues.

Differential Revision: D89502176

### DIFF
--- a/velox/dwio/text/reader/TextReader.cpp
+++ b/velox/dwio/text/reader/TextReader.cpp
@@ -15,13 +15,16 @@
  */
 
 #include "velox/dwio/text/reader/TextReader.h"
+
+#include <boost/algorithm/string/predicate.hpp>
+#include <string>
+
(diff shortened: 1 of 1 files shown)


velox/dwio/text/reader/TextReader.cpp
```diff
@@ -95,7 +96,7 @@ void resizeVector(
   }
 }
 
-bool endsWith(const std::string& str, const std::string& suffix) {
+bool endsWith(const std::string_view& str, const std::string_view& suffix) {
```
mbasmanova: std::ends_with ?

pedroerp: nice, even better :)

PR #12010, opened 2025-01-03
feat: Avoid small batches in Exchange

arhimondr:
Summary:
Prevent exchange client from unblocking to early. Unblocking to early impedes
effectiveness of page merging. When the cost of creating a vector is high (for
example for data sets with high number of columns) creating small pages can
make queries significantly less efficient.

For example it was observed that when network is congested and Exchange buffers
are not filled up as fast query may experience CPU efficiency drop up to 4x: T211034421

Differential Revision: D67615570

PR #15824, opened 2025-12-19
perf(exchange): Reduce awakened consumer count

lingbin:
Replace the 'while' loop with a simple 'if' condition in 
'ExchangeQueue::enqueueLocked()' to avoid unnecessary iterations when 
checking for waiting consumers. Because after all, only one page is 
added here, so at most **only one consumer** should be awakened. 
Otherwise, excess awakened consumers will block again immediately 
due to no available pages, which wastes resources. 

After this patch, we only wake up one waiting consumer at a time when 
accumulated data meets the minimum batch size requirement(see 
'minOutputBatchBytesLocked()'). This reduces the number of awakened 
consumers and improves performance.  

See [1] for discussion.

[1] https://github.com/facebookincubator/velox/pull/12010#discussion_r1954908872

### DIFF
--- a/velox/exec/ExchangeQueue.cpp
+++ b/velox/exec/ExchangeQueue.cpp
@@ -74,14 +74,7 @@ void ExchangeQueue::enqueueLocked(
 
   queue_.push_back(std::move(page));
   const auto minBatchSize = minOutputBatchBytesLocked();
-  while (!promises_.empty()) {
-    VELOX_CHECK_LE(promises_.size(), numberOfConsumers_);
-    const int32_t unblockedConsumers = numberOfConsumers_ - promises_.size();
-    const int64_t unasignedBytes =
-        totalBytes_ - unblockedConsumers * minBatchSize;
-    if (unasignedBytes < minBatchSize) {
-      break;
-    }
+  if (!promises_.empty() && totalBytes_ >= minBatchSize) {
     // Resume one of the waiting drivers.
     auto it = promises_.begin();
     promises.push_back(std::move(it->second));
--- a/velox/exec/tests/ExchangeClientTest.cpp
+++ b/velox/exec/tests/ExchangeClientTest.cpp
@@ -890,7 +890,7 @@ TEST_P(ExchangeClientTest, minOutputBatchBytesMultipleConsumers) {
           consumers.begin(),
           consumers.end(),
           [](auto& consumer) { return consumer.isReady(); }),
-      numConsumers);
+      1);
 
   auto pages = client->next(1, 1, &atEnd, &consumers[1]);
   ASSERT_EQ(1, pages.size());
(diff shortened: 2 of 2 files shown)


mbasmanova:
CC: @arhimondr

velox/exec/ExchangeQueue.cpp
```diff
@@ -74,14 +74,7 @@ void ExchangeQueue::enqueueLocked(
...
   const auto minBatchSize = minOutputBatchBytesLocked();
-  while (!promises_.empty()) {
-    VELOX_CHECK_LE(promises_.size(), numberOfConsumers_);
-    const int32_t unblockedConsumers = numberOfConsumers_ - promises_.size();
-    const int64_t unasignedBytes =
-        totalBytes_ - unblockedConsumers * minBatchSize;
-    if (unasignedBytes < minBatchSize) {
-      break;
-    }
+  if (!promises_.empty() && totalBytes_ >= minBatchSize) {
```
mbasmanova, line 77: Isn't it possible that there are > 1 pages in the queue_?

lingbin: @mbasmanova   Thank you for your review. Yes, there can be more than one page in the queue.   

Since a consumer consumes multiple pages at a time(`dequeueLocked()` return a `std::vector` at a time), I think the key here isn't the number of remaining pages in the queue_, but rather the total size of all the remaining pages in the queue_. 
Because only one page was added here, waking up one consumer is sufficient, the awakened consumer will consume all the data in the queue_.

mbasmanova: > the key here isn't the number of remaining pages in the queue_, but rather the total size of all the remaining pages in the queue_.

This is my understanding as well.

> Because only one page was added here, waking up one consumer is sufficient, the awakened consumer will consume all the data in the queue_.

I'm not sure I understand why "the awakened consumer will consume all the data in the queue_". I believe it is possible that there is more data in the queue then can be consumed in a single dequeueLocked call. Am I missing something?

```
...If multiple pages are available,
  /// returns as many pages as fit within 'maxBytes', but no fewer than one...

  std::vector<std::unique_ptr<SerializedPageBase>> dequeueLocked(
      int consumerId,
      uint32_t maxBytes,
      bool* atEnd,
      ContinueFuture* future,
      ContinuePromise* stalePromise);
```

lingbin: @mbasmanova  Thanks for your reply.  

If there is data in the queue_ but the consumer is still blocked, it means the data in the queue_ is less than `minExchangeOutputBatchBytes` (value: `2MB`).  Each consumer (the `Exchange` operator) consumes data with `maxBytes = preferredOutputBatchBytes` (value: `10MB`, bigger than `2MB`), so it will consume all the data.

https://github.com/facebookincubator/velox/blob/bbe1d1d9f82e3c2887678a790b06d3c5e9816a09/velox/exec/Exchange.cpp#L143-L144

lingbin:
@mbasmanova  @kagamiori  Kindly ping.

PR #15826, opened 2025-12-19
fix(build): Issues in tests found by macOS15 intel compiler

czentgr:
Several issues were found in test programs:
- Not directly comparable enums
- Ambiguous function calls involving derived function overridden implementation
- Incorrectly implemented derived function that don’t use const or override correctly I expect newer compiler to pick up on these eventually so let’s fix them now. These are test fixes and don’t affect engine code.


majetideepak:
@czentgr Is this PR still relevant? We dropped x86 support with macos?

czentgr:
@majetideepak Yes, we did drop the x86 support for macOS. These cleanups would make the code a bit more clean. I wonder if we will see some of these again in the future.

majetideepak:
@czentgr do you want to rebase?

PR #15829, opened 2025-12-19
fix: Find VeloxType singletons for CustomType when embedded in OpaqueType objects

AshishCbMisra:
Summary:
## Background
- This diff is to fix a bug in the Types::create() method for CustomType.
- The bug manifests in the SignatureBinder.bind() operation failing to match for a CustomType argument.
- The CustomType is basically an OpaqueType object. 
- The root cause is that two different registries store OpaqueType objects.
   - Custom Type Registry → VeloxType singleton (correct name())
   - OpaqueSerdeRegistry → Base OpaqueType (name() = "OPAQUE")

-  When deserializing the custom type in Type::create() method, the code does not find VeloxType singleton, hence created an OpaqueType.
-  The CustomType returns its name in its custom name in the name() and toString() methods, whereas the OpaqueType returns "OPAQUE<demangled-name>" in its name/toString() methods.

## In this Diff
- The primary problem is that the CustomType has a Velox singleton for its type.
   However, type is serialized and deserialized, the type produced is an Opaque type, with custom type inside.
- In this diff, the velox deserialization correctly finds and produces the VeloxType singleton.

## Implementation Changes
- Types.cpp
   - Implemented searching and deserializing CustomType within OpaqueTypes in Type::create() method.

## Testing Changes
- OpaqueCustomTypesTest.cpp
   - Added test to serialize/deserialize a CustomType and confirm that VeloxType singleton is returned.

Differential Revision: D89564769

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -203,8 +203,8 @@ TypePtr Type::create(const folly::dynamic& obj) {
   if (isDecimalName(typeName)) {
     return DECIMAL(obj["precision"].asInt(), obj["scale"].asInt());
   }
-  // Checks if 'typeName' specifies a custom type.
-  if (customTypeExists(typeName)) {
+
+  auto customTypeSingleton = [&](const std::string& typeName) {
     std::vector<TypeParameter> params;
     if (obj.find("cTypes") != obj.items().end()) {
       params.reserve(childTypes.size());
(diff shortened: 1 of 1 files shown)


AshishCbMisra:
Pull request https://github.com/facebookincubator/velox/pull/15843 fixes this issue. Hence this PR is being closed.

PR #15835, opened 2025-12-20
refactor: Move TableWriteTraits to velox/core

pedroerp:
Summary:
Moving TableWriteTraits to velox/core, as it is strictly related
to plan metadata, and not to execution. The fact that TableWriteTraits was in
velox/exec forced clients that only wanted to define plans to take a full
dependency on velox/exec.

Leaving an alias in the previous namespace to maintain backward compatibility.

Differential Revision: D89592214


pedroerp:
@mbasmanova we should do it in a follow up, but another idea would be to pre-define a set of output columns that are "supported" by table writer, and having users providing which ones they want (and their order) as part of the plan. 

Do you guys know how this work for Gluten today? @rui-mo @jinchengchenghh @JkSelf

rui-mo:
> @mbasmanova we should do it in a follow up, but another idea would be to pre-define a set of output columns that are "supported" by table writer, and having users providing which ones they want (and their order) as part of the plan.
> 
> Do you guys know how this work for Gluten today? @rui-mo @jinchengchenghh @JkSelf

Currently, Gluten only passes null to the output type via `TableWriteTraits::outputType(std::nullopt)`. cc @JkSelf in case you have additional context.

JkSelf:
@pedroerp @rui-mo 
Gluten only requires the second column fragments to construct the necessary Spark information, as shown here: https://github.com/apache/incubator-gluten/blob/a35a3bda13b17e6cf51dd1f2d2e89b7fb3ecae20/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarWriteFilesExec.scala#L47-L69. And it doesn't use the TableWriteTrait object in Gluten.

PR #15842, opened 2025-12-22
refactor(variant): Improve code organization and more tests

pedroerp:
Summary:
No functional changes. Adding test coverage missing for a variety of
methods (create, serialize/deserialize, container nested types). Re-organizing
and grouping test suite methods.

Re-organizing Variant's API to make it more natural (construction methods
first), and adding a documentation with examples for popular construction
methods.

Differential Revision: D89691635

### DIFF
--- a/velox/type/tests/VariantTest.cpp
+++ b/velox/type/tests/VariantTest.cpp
@@ -20,14 +20,529 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/type/tests/utils/CustomTypesForTesting.h"
 
-namespace facebook::velox {
-
+namespace facebook::velox::test {
 namespace {
-void testSerDe(const Variant& value) {
-  auto serialized = value.serialize();
-  auto copy = Variant::create(serialized);
 
-  ASSERT_EQ(value, copy);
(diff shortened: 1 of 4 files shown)


velox/type/tests/VariantTest.cpp
```diff
@@ -20,14 +20,529 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/type/tests/utils/CustomTypesForTesting.h"
 
-namespace facebook::velox {
-
+namespace facebook::velox::test {
```
mbasmanova, line 23: The convention is to put test in anonymous namespace within the namespace that contains code being tested.

pedroerp: Good question. I see most test code being added in a ::test nested namespace inside the namespace being tested, then inside an anonymous namespace.

We should probably discuss this and add something to our code guidelines.

pedroerp:
> Thanks. It would be nice to clarify that Variable holds information about physical type only. It doesn't hold information about logical type. Also, it doesn't store field names of a struct.

Thank you. I added a couple of sentences to document this as well.

PR #15843, opened 2025-12-22
feat: Implement one step registration for OpaqueCustomType

AshishCbMisra:
Summary:
## Background.
- Today, a CustomType (inner type) may be wrapped by an OpaqueType (outer type), forming a OpaqueCustomType.
- When an OpaqueCustomType is defined, it needs to be registered in two place.
   - Custom Type Registry - a mapping from type name to an instance of CustomTypeFactory   that provides an instance of the type
   - OpaqueSerdeRegistry - a mapping from type name to functions that serialize and deserialize opaque type values
- Today, this requires two separate calls to the API. These separate calls are inconvenient and error prone.

## Implementation
- OpaqueCustomTypes.h
   - This PR changes the OpaqueCustomTypeRegister::registerType() function to accept optional serialization/deserialization functions, thus eliminating the need for a separate API call.
   - By keep the function arguments optional, it provides backward compatibility with existing code.

## Testing Changes
- OpaqueCustomTypesTest.cpp
   - Implements unit-tests that ensure that serialization/deserialization functions can be provided in the registerType function optionally.
   - Unit-tests ensure that serialization/deserialization functions can be provided utmost once.

Differential Revision: D89686612

### DIFF
--- a/velox/type/OpaqueCustomTypes.h
+++ b/velox/type/OpaqueCustomTypes.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include "velox/common/base/Exceptions.h"
 #include "velox/type/SimpleFunctionApi.h"
 #include "velox/type/Type.h"
 
@@ -34,14 +35,19 @@ class CastOperator;
 template <typename T, const char* customTypeName>
 class OpaqueCustomTypeRegister {
  public:
-  static bool registerType() {
-    return facebook::velox::registerCustomType(
-        customTypeName, std::make_unique<const TypeFactory>());
-  }
+  using SerializeValueFunc = OpaqueType::SerializeFunc<T>;
+  using DeserializeValueFunc = OpaqueType::DeserializeFunc<T>;
 
-  static bool unregisterType() {
-    return facebook::velox::unregisterCustomType(customTypeName);
-  }
+  // @param serializeValueFunc Optional serialization function for values of
+  // type T.
+  // @param deserializeValueFunc Optional deserialization function for values of
+  // type T. Both serializeValueFunc and deserializeValueFunc must be specified
+  // or not together.
+  static bool registerType(
+      SerializeValueFunc serializeValueFunc = nullptr,
+      DeserializeValueFunc deserializeValueFunc = nullptr);
+
+  static bool unregisterType();
 
   // Type used in the simple function interface as CustomType<TypeT>.
   struct TypeT {
@@ -51,13 +57,16 @@ class OpaqueCustomTypeRegister {
 
   using SimpleType = CustomType<TypeT>;
 
+  class VeloxType;
+  using VeloxTypePtr = std::shared_ptr<const VeloxType>;
+
   class VeloxType : public OpaqueType {
    public:
     VeloxType() : OpaqueType(std::type_index(typeid(T))) {}
 
-    static const TypePtr& get() {
+    static const VeloxTypePtr& get() {
       static const VeloxType kInstance;
-      static const TypePtr kInstancePtr{TypePtr{}, &kInstance};
+      static const VeloxTypePtr kInstancePtr{TypePtr{}, &kInstance};
       return kInstancePtr;
     }
 
@@ -77,13 +86,20 @@ class OpaqueCustomTypeRegister {
     std::string toString() const override {
       return customTypeName;
     }
+
+    folly::dynamic serialize() const override {
+      folly::dynamic obj = folly::dynamic::object;
+      obj["name"] = "Type";
+      obj["type"] = customTypeName;
+      return obj;
+    }
   };
 
-  static const TypePtr& singletonTypePtr() {
+  static const VeloxTypePtr& singletonTypePtr() {
     return VeloxType::get();
   }
 
-  static const TypePtr& get() {
+  static const VeloxTypePtr& get() {
     return VeloxType::get();
   }
 
@@ -108,4 +124,34 @@ class OpaqueCustomTypeRegister {
     }
   };
 };
+
+template <typename T, const char* customTypeName>
+bool OpaqueCustomTypeRegister<T, customTypeName>::registerType(
+    SerializeValueFunc serializeValueFunc,
+    DeserializeValueFunc deserializeValueFunc) {
+  VELOX_USER_CHECK(
+      ((serializeValueFunc && deserializeValueFunc) ||
+       (!serializeValueFunc && !deserializeValueFunc)),
+      "Both serialization and deserialization functions need to be registered for custom type {}",
+      customTypeName);
+  if (registerCustomType(
+          customTypeName, std::make_unique<const TypeFactory>())) {
+    if (serializeValueFunc && deserializeValueFunc) {
+      OpaqueType::registerSerialization<T>(
+          customTypeName, serializeValueFunc, deserializeValueFunc);
+    }
+    return true;
+  }
+  return false;
+}
+
+template <typename T, const char* customTypeName>
+bool OpaqueCustomTypeRegister<T, customTypeName>::unregisterType() {
+  if (unregisterCustomType(customTypeName)) {
+    OpaqueType::unregisterSerialization(singletonTypePtr(), customTypeName);
+    return true;
+  }
+  return false;
+}
+
 } // namespace facebook::velox
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -907,6 +907,15 @@ std::shared_ptr<const OpaqueType> OpaqueType::deserializeExtra(
   return nullptr;
 }
 
+bool OpaqueType::unregisterSerialization(
+    const OpaqueTypePtr& opaqueType,
+    const std::string& persistentName) {
+  VELOX_CHECK_NOT_NULL(opaqueType);
+  auto& registry = OpaqueSerdeRegistry::get();
+  registry.reverse.erase(persistentName);
+  return registry.mapping.erase(opaqueType->typeIndex_);
+}
+
 void OpaqueType::clearSerializationRegistry() {
   auto& registry = OpaqueSerdeRegistry::get();
   registry.mapping.clear();
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -1362,6 +1362,15 @@ class OpaqueType : public TypeBase<TypeKind::OPAQUE> {
         deserializeTypeErased);
   }
 
+  // This function is used to remove the serialization/deserialization functions
+  // for a type. It reverses the state changes made by registerSerialization
+  // function.
+  // @return true if the functions are found and removed
+  // successfully. False otherwise.
+  static bool unregisterSerialization(
+      const std::shared_ptr<const OpaqueType>& opaqueType,
+      const std::string& persistentName);
+
   static void clearSerializationRegistry();
 
  protected:
@@ -1376,6 +1385,8 @@ class OpaqueType : public TypeBase<TypeKind::OPAQUE> {
       SerializeFunc<void> serialize = nullptr,
       DeserializeFunc<void> deserialize = nullptr);
 };
+
+using OpaqueTypePtr = std::shared_ptr<const OpaqueType>;
 
 using IntegerType = ScalarType<TypeKind::INTEGER>;
 using BooleanType = ScalarType<TypeKind::BOOLEAN>;
@@ -2371,6 +2382,7 @@ bool registerOpaqueType(const std::string& alias) {
 template <typename Class>
 bool unregisterOpaqueType(const std::string& alias) {
   auto typeIndex = std::type_index(typeid(Class));
+  OpaqueType::unregisterSerialization(OpaqueType::create<Class>(), alias);
   return getTypeIndexByOpaqueAlias().erase(alias) == 1 &&
       getOpaqueAliasByTypeIndex().erase(typeIndex) == 1;
 }
--- a/velox/type/tests/OpaqueCustomTypesTest.cpp
+++ b/velox/type/tests/OpaqueCustomTypesTest.cpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <folly/dynamic.h>
+#include <gtest/gtest.h>
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/type/OpaqueCustomTypes.h"
+#include "velox/type/Variant.h"
+
+namespace facebook::velox {
+namespace {
+
+class OpaqueCustomTypeTest : public testing::Test {
+ public:
+  OpaqueCustomTypeTest()
+      : testCustomType_(TestCustomTypeRegister::VeloxType::get()) {}
+
+ protected:
+  struct TestCustomType {
+    std::string name;
+
+    bool operator==(const TestCustomType& other) const {
+      return name == other.name;
+    }
+
+    folly::dynamic serialize() const {
+      folly::dynamic obj = folly::dynamic::object();
+      obj["name"] = name;
+      return obj;
+    }
+
+    static std::shared_ptr<TestCustomType> create(const folly::dynamic& obj) {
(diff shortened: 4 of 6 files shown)


velox/type/OpaqueCustomTypes.h
```diff
@@ -34,15 +34,26 @@ class CastOperator;
...
 class OpaqueCustomTypeRegister {
  public:
-  static bool registerType() {
-    return facebook::velox::registerCustomType(
-        customTypeName, std::make_unique<const TypeFactory>());
-  }
+  using SerializeFunc = OpaqueType::SerializeFunc<T>;
+  using DeserializeFunc = OpaqueType::DeserializeFunc<T>;
+
+  // Optional serialization and deserialization functions. If not provided, the
```
mbasmanova: Perhaps, clarify that these are for serializing values of type T, not for serializing the type itself. Perhaps, rename to serializeValue and deserializeValue.

velox/type/OpaqueCustomTypes.h
```diff
@@ -34,15 +34,26 @@ class CastOperator;
...
+  // the registerSerialization() can be invoked utmost once later.
+  static bool registerType(
+      SerializeFunc serialize = nullptr,
+      DeserializeFunc deserialize = nullptr);
 
   static bool unregisterType() {
     return facebook::velox::unregisterCustomType(customTypeName);
   }
 
+  // Can be invoked only if serialization/deserialization functions were not
```
mbasmanova: Do we need this method? I assume not.

velox/type/OpaqueCustomTypes.h
```diff
@@ -108,4 +119,27 @@ class OpaqueCustomTypeRegister {
     }
   };
 };
+
+template <typename T, const char* customTypeName>
+bool OpaqueCustomTypeRegister<T, customTypeName>::registerType(
+    SerializeFunc serialize,
+    DeserializeFunc deserialize) {
+  if (facebook::velox::registerCustomType(
```
mbasmanova: drop facebook::velox:: prefix

velox/type/OpaqueCustomTypes.h
```diff
@@ -108,4 +119,27 @@ class OpaqueCustomTypeRegister {
...
+          customTypeName, std::make_unique<const TypeFactory>())) {
+    if (serialize || deserialize) {
```
mbasmanova: Let's assert that either both are null or both are not null. It doesn't make sense to specify one, but not the other, does it?

velox/type/OpaqueCustomTypes.h
```diff
@@ -34,15 +34,26 @@ class CastOperator;
...
+  using SerializeFunc = OpaqueType::SerializeFunc<T>;
+  using DeserializeFunc = OpaqueType::DeserializeFunc<T>;
+
+  // Optional serialization and deserialization functions. If not provided, the
+  // the registerSerialization() can be invoked utmost once later.
+  static bool registerType(
+      SerializeFunc serialize = nullptr,
+      DeserializeFunc deserialize = nullptr);
 
   static bool unregisterType() {
```
mbasmanova: Let's update unregisterType to unregister serde functions as well.

Let's also add 'serialize' API to VeloxType.

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/type/OpaqueCustomTypes.h"
+
+using namespace facebook;
+using namespace facebook::velox;
+
+namespace {
+
+TEST(OpaqueCustomTypeTest, RegisterWithoutSerde) {
```
mbasmanova: test methods start with lowercase

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // This is the C++ type registered within the opaque capsule. This could be
+  // anything.
+  struct TestCustomType {};
```
mbasmanova: nit: consider using shorter easier to read names

FooType, fooTypeName, FooTypeRegister, etc.

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,87 @@
...
+  static constexpr char testCustomTypeName[] = "my_test_custom_type";
+  using TestCustomTypeRegister =
+      OpaqueCustomTypeRegister<TestCustomType, testCustomTypeName>;
+
+  // Register type without serialization/deserialization and build a TypePtr.
+  TestCustomTypeRegister::registerType();
+  auto testCustomType = TestCustomTypeRegister::VeloxType::get();
```
mbasmanova: Let's also test round-trip serde for the type itself.

velox/type/OpaqueCustomTypes.h
AshishCbMisra: I had built in this capability, to allow for serialization/deserialization functions to be bolted on later. That is, Custom opaque type being defined in a library, while the user of the library could later define serialization/deserialization because they needed it. 

It was a design that offered flexibility at no extra cost. What do you think?

velox/type/OpaqueCustomTypes.h
AshishCbMisra: Agreed. 
I had not figured out the correct response if one was null, but not the other. Assert is a great choice.

velox/type/OpaqueCustomTypes.h
mbasmanova: My preference to keep API surface as small as possible. Flexibility is nice, but it has non-trivial maintenance cost.

AshishCbMisra: Sounds good. I will remove the method.
It is an extra capability and can be added on later easily, if we ever need it.

velox/type/OpaqueCustomTypes.h
```diff
@@ -108,4 +121,38 @@ class OpaqueCustomTypeRegister {
...
+bool OpaqueCustomTypeRegister<T, customTypeName>::registerType(
+    SerializeValueFunc serializeValueFunc,
+    DeserializeValueFunc deserializeValueFunc) {
+  if (registerCustomType(
+          customTypeName, std::make_unique<const TypeFactory>())) {
+    const bool registerSerialization =
+        (serializeValueFunc && deserializeValueFunc) ||
+        (!serializeValueFunc && !deserializeValueFunc);
+    if (registerSerialization) {
+      using Type = typename TypeT::type::element_type;
```
mbasmanova: Is this not just T?

velox/type/OpaqueCustomTypes.h
```diff
@@ -108,4 +121,38 @@ class OpaqueCustomTypeRegister {
...
+      OpaqueType::registerSerialization<Type>(
+          customTypeName, serializeValueFunc, deserializeValueFunc);
+    } else {
+      unregisterType();
+      VELOX_USER_CHECK(
```
mbasmanova: Let's move this check to the start of this function.

```
VELOX_USER_CHECK((serializeValueFunc && deserializeValueFunc) ||
        (!serializeValueFunc && !deserializeValueFunc), ...);
        
if (registerCustomType(
          customTypeName, std::make_unique<const TypeFactory>())) {
   if (serializeValueFunc && deserializeValueFunc) {
      OpaqueType::registerSerialization<T>(
          customTypeName, serializeValueFunc, deserializeValueFunc);
   }

    return true;
 }        

return false;
```

velox/type/OpaqueCustomTypes.h
```diff
@@ -108,4 +121,38 @@ class OpaqueCustomTypeRegister {
...
+          false,
+          "Both serialization and deserialization functions need to be registered for custom type {}",
+          customTypeName);
+    }
+    return true;
+  }
+  return false;
+}
+template <typename T, const char* customTypeName>
```
mbasmanova, line 148: add empty line between methods

velox/type/OpaqueCustomTypes.h
```diff
@@ -77,13 +82,21 @@ class OpaqueCustomTypeRegister {
     std::string toString() const override {
       return customTypeName;
     }
+
+    folly::dynamic serialize() const override {
+      folly::dynamic obj = folly::dynamic::object;
+      obj["name"] = "Type";
+      obj["type"] = TypeTraits<TypeKind::OPAQUE>::name;
+      obj["opaque"] = customTypeName;
```
mbasmanova: Why not just

```
    obj["name"] = "Type";
    obj["type"] = customTypeName;
```

?

velox/type/OpaqueCustomTypes.h
```diff
@@ -34,14 +35,15 @@ class CastOperator;
...
-    return facebook::velox::registerCustomType(
-        customTypeName, std::make_unique<const TypeFactory>());
-  }
+  using SerializeValueFunc = OpaqueType::SerializeFunc<T>;
+  using DeserializeValueFunc = OpaqueType::DeserializeFunc<T>;
 
-  static bool unregisterType() {
-    return facebook::velox::unregisterCustomType(customTypeName);
-  }
+  // Optional serialization and deserialization functions for values of type T.
```
mbasmanova: /// + @param

```
/// @param serializeValueFunc Optional serialization function for values of type T.
/// @param deserializeValueFunc Optional deserialization function for values of type T. Both serializeValueFunc and deserializeValueFunc must be specified or not together.
```

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+  struct TestCustomType {};
+  static constexpr char testCustomTypeName[] = "my_test_custom_type";
+  using TestCustomTypeRegister =
+      OpaqueCustomTypeRegister<TestCustomType, testCustomTypeName>;
+
+  TestCustomTypeRegister::VeloxTypePtr testCustomType_;
+
+  void TearDown() override {
+    TestCustomTypeRegister::unregisterType();
+    ASSERT_EQ(getCustomType(testCustomTypeName, {}), nullptr);
```
mbasmanova, line 59: There should be no testing / assert logic in TearDown. This is too late.

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  void testTypeSerde() {
+    // Test serialization/deserialization.
+    auto serializedCustomType = testCustomType_->serialize();
+    EXPECT_EQ(testCustomType_->name(), serializedCustomType["opaque"]);
+    auto deserializedType = Type::create(serializedCustomType);
+
+    // TODO: Ideally this should be EXPECT_EQ() but it doesn't work
+    // because Type::create() currently does not return the singleton instance.
```
mbasmanova: This will work if you just fix 'serialize' method like I suggested.

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+    VELOX_ASSERT_THROW(
+        testCustomType_->getSerializeFunc(),
+        "No serialization function registered for my_test_custom_type");
+    VELOX_ASSERT_THROW(
+        testCustomType_->getDeserializeFunc(),
+        "No deserialization function registered for my_test_custom_type");
+  }
+
+  void testTypeSerde() {
+    // Test serialization/deserialization.
```
mbasmanova: redundant comment

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+TEST_F(OpaqueCustomTypeTest, registerWithSerde) {
+  // Register type with serialization/deserialization and build a TypePtr.
+  TestCustomTypeRegister::registerType(
+      /*serializeValueFunc=*/[](const std::shared_ptr<
+                                 TestCustomType>&) { return ""; },
+      /*deserializeValueFunc=*/
+      [](const std::string&) { return std::make_shared<TestCustomType>(); });
+
+  testTypeSerde();
```
mbasmanova, line 102: Can we verifiy that serializeValueFunc and deserializeValueFunc were actually registered and are functional?

velox/type/OpaqueCustomTypes.h
AshishCbMisra: Couple of reasons:
- It is the OpaqueCustom type. The representation should be as such: Opaque outer type, embedding the custom inner type. 
- It breaks backward compatibility. A server running one velox version would break when sending serialized type to another running a previous version.

mbasmanova: > It is the OpaqueCustom type. The representation should be as such: Opaque outer type, embedding the custom inner type.

Not really. OpaqueType here is just an implementation detail of the custom type. It doesn't / shouldn't place a restriction on how custom type chooses to serialize itself.

> It breaks backward compatibility. A server running one velox version would break when sending serialized type to another running a previous version.

There is no such guarantee in Velox to begin with. Let's discuss offline if this needs to be added as it would be non-trivial undertaking.

velox/type/Type.cpp
```diff
@@ -907,6 +907,20 @@ std::shared_ptr<const OpaqueType> OpaqueType::deserializeExtra(
...
 }
 
+bool OpaqueType::clearSerializationRegistry(
+    const OpaqueTypePtr& opaqueType,
+    const std::string& persistentName) {
+  VELOX_CHECK_NOT_NULL(opaqueType);
+  auto& registry = OpaqueSerdeRegistry::get();
+  auto it = registry.mapping.find(opaqueType->typeIndex_);
+  if (it != registry.mapping.end()) {
+    registry.mapping.erase(opaqueType->typeIndex_);
```
mbasmanova: perhaps, use the 'it' to erase (no need to lookup the value again)

velox/type/OpaqueCustomTypes.h
AshishCbMisra: Done.

velox/type/OpaqueCustomTypes.h
AshishCbMisra: >> There is no such guarantee in Velox to begin with
I see.  In that case, done. Implemented your recommendation. 
I considered the implications of the change and reasoned that the changes being done are only to serialization/deserialization format. The Type object that is returned after deserialization is a singleton (same as before).  Hence the change should be safe.
The only potential problems should be code that inspects the serialized type object (for eg: unit-tests).

velox/type/tests/OpaqueCustomTypesTest.cpp
AshishCbMisra: Done.

velox/type/Type.cpp
AshishCbMisra: Done.

velox/type/Type.cpp
```diff
@@ -907,6 +907,20 @@ std::shared_ptr<const OpaqueType> OpaqueType::deserializeExtra(
   return nullptr;
 }
 
+bool OpaqueType::clearSerializationRegistry(
+    const OpaqueTypePtr& opaqueType,
+    const std::string& persistentName) {
+  VELOX_CHECK_NOT_NULL(opaqueType);
+  auto& registry = OpaqueSerdeRegistry::get();
+  auto it = registry.mapping.find(opaqueType->typeIndex_);
```
mbasmanova: nit: seems like no need to call 'find' before calling 'erase'; just call 'erase' and check the result; or, perhaps, use 'it' to verify that 'persistentName' matches 'opaqueType'

velox/type/Type.h
```diff
@@ -1362,6 +1362,10 @@ class OpaqueType : public TypeBase<TypeKind::OPAQUE> {
         deserializeTypeErased);
   }
 
+  static bool clearSerializationRegistry(
```
mbasmanova: nit: clearSerializationRegistry -> unregisterSerialization for consistency

please, document this API

velox/type/tests/OpaqueCustomTypesTest.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+#include <folly/dynamic.h>
+#include <gtest/gtest.h>
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/type/OpaqueCustomTypes.h"
+#include "velox/type/Variant.h"
+
+using namespace facebook;
+using namespace facebook::velox;
```
mbasmanova: put the test class into same namespace as the code being tested; then remove these

AshishCbMisra: Done.

velox/type/Type.h
AshishCbMisra: Done.

velox/type/Type.cpp
AshishCbMisra: Done.

velox/type/tests/OpaqueCustomTypesTest.cpp
AshishCbMisra: Done.

PR #15848, opened 2025-12-23
feat: Allow subfield rename and deletion for ORC format

rui-mo:
The default behavior of the schema evolution for row type is matching by index. 
This PR supports subfield rename and deletion for ORC file format, controlled by
configuration `useColumnNamesForColumnMapping`.
Missing subfields are identified by matching the file type and requested type on
the names of subfileds, and NULL occupies the position of the missing subfields. 
Below table summarizes the results for difference cases.

| Column schema | Requested output schema | ORC result |
| ------------- | ------------- |  ------------- |
| row({"a", "c"}) | row({"a", "b", "c"})  |  row(a_val, NULL, c_val) |
| row({"a", "c"}) | row({"b"})  | row(NULL) |
| row({"a", "c"}) | row({"b", "d"})  | row(NULL, NULL) |
| row({"a", "c"}) | row({})  | row() |

This is a separation PR of https://github.com/facebookincubator/velox/pull/5962.

### DIFF
--- a/velox/connectors/hive/SplitReader.cpp
+++ b/velox/connectors/hive/SplitReader.cpp
@@ -440,10 +440,11 @@ std::vector<TypePtr> SplitReader::adaptColumns(
       auto fileTypeIdx = fileType->getChildIdxIfExists(fieldName);
       if (!fileTypeIdx.has_value()) {
         // Column is missing. Most likely due to schema evolution.
-        VELOX_CHECK(tableSchema, "Unable to resolve column '{}'", fieldName);
+        auto schema = tableSchema ? tableSchema : readerOutputType_;
+        VELOX_CHECK(schema, "Unable to resolve column '{}'", fieldName);
         childSpec->setConstantValue(
             BaseVector::createNullConstant(
-                tableSchema->findChild(fieldName),
+                schema->findChild(fieldName),
                 1,
                 connectorQueryCtx_->memoryPool()));
       } else {
--- a/velox/connectors/hive/HiveConnectorUtil.cpp
+++ b/velox/connectors/hive/HiveConnectorUtil.cpp
@@ -569,7 +569,7 @@ void configureReaderOptions(
       break;
     }
     default:
-      useColumnNamesForColumnMapping = false;
+      break;
   }
 
   readerOptions.setUseColumnNamesForColumnMapping(
--- a/velox/dwio/common/SelectiveFlatMapColumnReader.h
+++ b/velox/dwio/common/SelectiveFlatMapColumnReader.h
@@ -24,11 +24,13 @@ namespace facebook::velox::dwio::common {
 class SelectiveFlatMapColumnReader : public SelectiveStructColumnReaderBase {
  protected:
   SelectiveFlatMapColumnReader(
+      const dwio::common::ColumnReaderOptions& columnReaderOptions,
       const TypePtr& requestedType,
       const std::shared_ptr<const dwio::common::TypeWithId>& fileType,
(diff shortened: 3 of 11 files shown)


velox/connectors/hive/HiveConfig.h
```diff
@@ -85,6 +85,16 @@ class HiveConfig {
   static constexpr const char* kParquetUseColumnNamesSession =
       "parquet_use_column_names";
 
+  static constexpr const char* kOrcAllowEnhancedSchemaEvolution =
```
Yuhta: Just name it `match_fields_by_name`.  Also do we really want to differentiate this config by file format?  The usual case is the whole data warehouse should have one setting, no matter which file format is used in a particular partition.  Otherwise the management of metadata would be very messy, and data migration would be almost impossible.

Also how does it interplay with `orc_use_column_names` and `parquet_use_column_names`?  Do we want to deprecate these?

velox/connectors/hive/SplitReader.cpp
```diff
@@ -440,12 +440,22 @@ std::vector<TypePtr> SplitReader::adaptColumns(
       auto fileTypeIdx = fileType->getChildIdxIfExists(fieldName);
       if (!fileTypeIdx.has_value()) {
         // Column is missing. Most likely due to schema evolution.
-        VELOX_CHECK(tableSchema, "Unable to resolve column '{}'", fieldName);
+        TypePtr fieldType = nullptr;
+        if (baseReaderOpts_.allowEnhancedSchemaEvolution()) {
+          auto outputTypeIdx =
+              readerOutputType_->getChildIdxIfExists(fieldName);
```
Yuhta: We probably don't need this, `readerOutputType_` is just a subset of `tableSchema`

velox/dwio/common/Options.h
```diff
@@ -675,6 +680,10 @@ class ReaderOptions : public io::ReaderOptions {
     return useColumnNamesForColumnMapping_;
   }
 
+  bool allowEnhancedSchemaEvolution() const {
```
Yuhta: How about reuse `useColumnNamesForColumnMapping()`?

velox/connectors/hive/HiveConfig.h
rui-mo: Thanks for the prompt review. I’ve updated this PR to reuse the `useColumnNamesForColumnMapping` configuration. Previously, subfield rename and deletion would raise errors. They are now supported when `use_column_names` mode is enabled.

Regarding file format differentiation, it has been observed that Spark behaves differently for Parquet and ORC formats (see https://github.com/facebookincubator/velox/pull/5962#issue-1832550633). To ensure compatibility with both Presto and Spark when using the Parquet format, I propose adding an extra Parquet-specific configuration to handle this special NULL logic in the follow-up PR. For example, when reading a column of type `row({"a", "c"})` with a schema of `row({"b"})`, the default result is `row(null)`. With this configuration enabled, the result would instead be `null`. Does this approach make sense? Thanks.

velox/connectors/hive/SplitReader.cpp
rui-mo: `tableSchema` may be nullptr in some cases, so I updated the logic to use `readerOutputType_` for child access when `tableSchema` is unavailable. Please kindly let me know if this looks reasonable, thanks.

velox/dwio/common/Options.h
rui-mo: Updated, thanks.

PR #15855, opened 2025-12-24
fix(variant): Variant string conversion logic on creation

pedroerp:
Summary:
Fixing two issues with variant creation:
- Variant::create() was relying on implicit conversion from StringView into
  std::string. It also did not support construction with std::string_view.
  Adding overloads to make them explicit.
- Variant::typeWithCustomComparison() was also relying on implicit conversion.
  Making the conversion explicit.

Differential Revision: D89753870

### DIFF
--- a/velox/type/Variant.h
+++ b/velox/type/Variant.h
@@ -138,8 +138,7 @@ struct VariantTypeTraits<
     KIND,
     usesCustomComparison,
     std::enable_if_t<
-        TypeTraits<KIND>::isPrimitiveType && KIND != TypeKind::VARCHAR &&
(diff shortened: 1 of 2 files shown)


velox/type/Variant.h
```diff
@@ -489,10 +504,11 @@ class Variant {
   static Variant typeWithCustomComparison(
       typename TypeTraits<KIND>::NativeType input,
       const TypePtr& type) {
+    using variant_traits = detail::VariantTypeTraits<KIND, true>;
```
mbasmanova, line 507: naming: variant_traits -> VariantTraits

mbasmanova (approved):

PR #15860, opened 2025-12-26
fix: Disable SSD cache write when SSD is full

gggrace14:
When SSD has no space, SSD file writes take longer time while finally fail or succeed.
This makes SsdCache holds writesInProgress_ positive for longer time and blocks other
tasks to proceed, ex., time-sensitive cache ttl task. SSD no space state is not
recoverable given the SSD cache write pattern today.

This change detects SSD no space error early, put the SsdFile into kNoSpace state
and stops SSD file writes. It also stops future checkpointing, eviction and eviction logging.


xiaoxmeng (approved):
@gggrace14 thanks for the fix and iterations!

PR #15861, opened 2025-12-26
refactor: Make std::string conversions explicit

pedroerp:
Summary:
To be able to remove unsafe implicit conversions from StringView into
std::string, we first need to ensure all conversion being done implicitly are
not either avoided, or made explicit. Fixing a bunch of place throughout the
codebase.

Since we will also remove implicit conversion from StringView into
folly::StringPiece, making explicit conversion to std::string_view into
libraries that expect folly::StringPiece, since the former can be implicitly
converted to the latter.

Differential Revision: D89818861

### DIFF
--- a/velox/exec/fuzzer/PrestoQueryRunner.cpp
+++ b/velox/exec/fuzzer/PrestoQueryRunner.cpp
@@ -404,9 +404,10 @@ std::string PrestoQueryRunner::createTable(
 
   // Query Presto to find out table's location on disk.
   auto results = execute(fmt::format("SELECT \"$path\" FROM {}", name));
-
   auto filePath = extractSingleValue<StringView>(results);
-  auto tableDirectoryPath = fs::path(filePath).parent_path();
+
+  // TODO: Remove explicit std::string_view cast.
+  auto tableDirectoryPath = fs::path(std::string_view(filePath)).parent_path();
 
   // Delete the all-null row.
   execute(fmt::format("DELETE FROM {}", name));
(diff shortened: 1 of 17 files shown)


velox/exec/fuzzer/PrestoQueryRunner.cpp
```diff
@@ -404,9 +404,10 @@ std::string PrestoQueryRunner::createTable(
 
   // Query Presto to find out table's location on disk.
   auto results = execute(fmt::format("SELECT \"$path\" FROM {}", name));
-
   auto filePath = extractSingleValue<StringView>(results);
-  auto tableDirectoryPath = fs::path(filePath).parent_path();
+
+  // TODO: Remove explicit std::string_view cast.
```
jagill, line 409: Why add this just to put a TODO to remove it?

jagill (approved):
LGTM, one nit

velox/exec/fuzzer/PrestoQueryRunner.cpp
pedroerp: @jagill I'll write a post to summarize this soon, but the main thing we are doing is to (a) remove implicit conversion from StringView to std::string, and (b) enable implicit conversion from StringView to std::string_view. We can't do (b) before we do (a), because it makes conversions ambiguous. And doing (a) requires us to clean up essentially the entire codebase.

So for now I'm working on (a), but because (b) is not shipped yet, I'm tagging explicit conversions to std::string_view in places where needed, but adding this TODO to remind me to clean it up once (a) then (b) are shipped. 

In this specific case, this would work because fs::path takes both std::string and std::string_view. So this in fact was silently allocating a std::string under the covers. So now we are skipping the allocation by simply converting StringView to std::string_view (though this is a just test code).

PR #15863, opened 2025-12-26
perf(function): Avoid unnecessary string copies

pedroerp:
Summary:
Two items:
* Removing unnecessary copied from StringView into std::string in many places
  across Presto function implementations.
* In places where copies are needed, making them explicit since we will soon
  remove the unwanted implicit conversion.

Differential Revision: D89820646

### DIFF
--- a/velox/functions/prestosql/GeometryFunctions.h
+++ b/velox/functions/prestosql/GeometryFunctions.h
@@ -54,7 +54,7 @@ struct StGeometryFromTextFunction {
     GEOS_TRY(
         {
           geos::io::WKTReader reader;
-          geosGeometry = reader.read(wkt);
+          geosGeometry = reader.read(std::string(wkt));
         },
         "Failed to parse WKT");
     common::geospatial::GeometrySerializer::serialize(*geosGeometry, result);
@@ -184,7 +184,7 @@ struct StPolygonFunction {
     GEOS_TRY(
         {
           geos::io::WKTReader reader;
-          geosGeometry = reader.read(wkt);
+          geosGeometry = reader.read(std::string(wkt));
         },
         "Failed to parse WKT");
     auto validate = geospatial::validateType(
@@ -214,7 +214,8 @@ struct StRelateFunction {
     std::unique_ptr<geos::geom::Geometry> rightGeosGeometry =
         common::geospatial::GeometryDeserializer::deserialize(rightGeometry);
     GEOS_TRY(
-        result = leftGeosGeometry->relate(*rightGeosGeometry, relation);
+        result =
+            leftGeosGeometry->relate(*rightGeosGeometry, std::string(relation));
         , "Failed to check geometry relation");
 
     return Status::OK();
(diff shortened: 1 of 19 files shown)


velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -54,7 +54,7 @@ struct StGeometryFromTextFunction {
     GEOS_TRY(
         {
           geos::io::WKTReader reader;
-          geosGeometry = reader.read(wkt);
+          geosGeometry = reader.read(std::string(wkt));
```
jagill, line 57: fixme: We don't want to do this. These strings are often large (MBs). The reader takes a `const std::string &`. Can we make an explicit cast to that? Or skip this one?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -184,7 +184,7 @@ struct StPolygonFunction {
...
```
jagill, line 187: fixme: We don't want to do this. These strings are often large (MBs). The reader takes a const std::string &. Can we make an explicit cast to that? Or skip this one?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -1788,7 +1789,7 @@ struct GeometryFromGeoJsonFunction {
       out_type<Geometry>& result,
       const arg_type<Varchar>& geometry) {
     auto reader = geos::io::GeoJSONReader();
-    auto geosGeometry = reader.read(geometry);
+    auto geosGeometry = reader.read(std::string(geometry));
```
jagill, line 1792: fixme: We don't want to do this. These strings are often large (MBs). The reader takes a const std::string &. Can we make an explicit cast to that? Or skip this one?

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -1951,7 +1952,7 @@ struct StLineFromTextFunction {
     GEOS_TRY(
         {
           geos::io::WKTReader reader;
-          geosGeometry = reader.read(wkt);
+          geosGeometry = reader.read(std::string(wkt));
```
jagill, line 1955: fixme: We don't want to do this. These strings are often large (MBs). The reader takes a const std::string &. Can we make an explicit cast to that? Or skip this one?

jagill (approved):
Please fix the fixme comments to prevent significant geospatial regression.

velox/functions/prestosql/GeometryFunctions.h
pedroerp: @jagill this is already what happens today, only that the std::string conversion is done implicitly. It doesn't seem like the WKTReader API supports anything other than std::string:

https://github.com/libgeos/geos/blob/main/include/geos/io/WKTReader.h#L107-L117

velox/functions/prestosql/GeometryFunctions.h
pedroerp: @jagill same:

https://github.com/libgeos/geos/blob/main/include/geos/io/GeoJSONReader.h#L71

velox/functions/prestosql/GeometryFunctions.h
pedroerp: @jagill ditto

velox/functions/prestosql/GeometryFunctions.h
jagill: Well, that's the advantage of making the conversions explicit! Now I understand better.

pedroerp: Yep, that implicit conversion is super misleading but it's been tough to clean up. Looks like we're close now :)

PR #15867, opened 2025-12-29
refactor: Move KeyEncoder from velox/exec to velox/serializers

xiaoxmeng:
Summary: Move KeyEncoder library and tests to velox/serializers to enable access from nimble OSS build, which uses VELOX_BUILD_MINIMAL_WITH_DWIO and doesn't include velox_exec dependencies.

Differential Revision: D89872199


gggrace14 (approved):
Change from refactoring looks good to me!

PR #15869, opened 2025-12-29
refactor(simple-function): Make materialization of string-types explicit

pedroerp:
Summary:
Changing the internal trait API used for View wrappers in simple
functions. The API contains an internal `requiresMaterialization` flag for
types like Row, Maps, and Array, which need to be materialized (copied) into a
container. Strings were silently copied into std::string, which also naturally
qualifies as a materialization.

Changing the flag for string types to make it explicit that the StringView gets
materialized into a std::string. The code will now explicitly call
StringView::materialize() which will do the exact same an the implicit std::string
conversion.

Differential Revision: D89897907

### DIFF
--- a/velox/type/SimpleFunctionApi.h
+++ b/velox/type/SimpleFunctionApi.h
@@ -468,14 +468,14 @@ template <>
 struct MaterializeType<Varchar> {
   using nullable_t = std::string;
   using null_free_t = std::string;
-  static constexpr bool requiresMaterialization = false;
+  static constexpr bool requiresMaterialization = true;
 };
 
 template <>
 struct MaterializeType<Varbinary> {
   using nullable_t = std::string;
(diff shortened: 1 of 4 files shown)


pansatadru (approved):

velox/type/SimpleFunctionApi.h
```diff
@@ -468,14 +468,14 @@ template <>
 struct MaterializeType<Varchar> {
   using nullable_t = std::string;
   using null_free_t = std::string;
-  static constexpr bool requiresMaterialization = false;
+  static constexpr bool requiresMaterialization = true;
```
kgpai, line 471: Dont we need a materialize() for the element type if this is set to true ? Is that handled somewhere ?
Can you write a brief test ?

PR #15870, opened 2025-12-30
fix(parquet): Crash when writing all-null batches

boneanxs:
Fix a segmentation fault and buffer overflow when writing all-null Velox vectors to Parquet via the Arrow bridge by allocating a correctly sized dummy values buffer for all-null vectors.

PR #14755 partially fixed all-null buffers for certain types. However, `Bridge.cpp::exportValues` could still crash for types such as timestamp when `rows.changed()` was true and the all-null vector had no values buffer.

### DIFF
--- a/velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
+++ b/velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
@@ -761,30 +761,125 @@ TEST_F(ParquetWriterTest, dictionaryEncodedVector) {
 }
 
 TEST_F(ParquetWriterTest, allNulls) {
-  auto schema = ROW({"c0"}, {INTEGER()});
-  const int64_t kRows = 4096;
-  // Create a column with all elements being null.
-  auto nulls = makeNulls(kRows, [](auto /*row*/) { return true; });
-  auto flatVector = std::make_shared<FlatVector<int32_t>>(
-      pool_.get(),
-      schema->childAt(0),
-      nulls,
-      kRows,
-      /*values=*/nullptr,
-      std::vector<BufferPtr>());
-  auto data = std::make_shared<RowVector>(
-      pool_.get(), schema, nullptr, kRows, std::vector<VectorPtr>{flatVector});
+  auto rowType = ROW({
+      {"bool_col", BOOLEAN()},
+      {"tinyint_col", TINYINT()},
+      {"smallint_col", SMALLINT()},
+      {"int_col", INTEGER()},
+      {"bigint_col", BIGINT()},
+      {"float_col", REAL()},
+      {"double_col", DOUBLE()},
+      {"string_col", VARCHAR()},
+      {"binary_col", VARBINARY()},
+      {"date_col", DATE()},
+      {"timestamp_col", TIMESTAMP()},
+      {"decimal_col", DECIMAL(10, 2)},
+  });
 
-  auto* sinkPtr = write(data);
+  const int64_t kRows = 100;
+
+  // Create a nulls buffer with all bits set to null (0)
+  BufferPtr nulls = AlignedBuffer::allocate<bool>(kRows, leafPool_.get());
+  auto* rawNulls = nulls->asMutable<uint64_t>();
+  bits::fillBits(rawNulls, 0, kRows, bits::kNull);
+
+  auto vector = std::make_shared<RowVector>(
+      leafPool_.get(),
+      rowType,
+      nullptr,
+      kRows,
+      std::vector<VectorPtr>{
+          std::make_shared<FlatVector<bool>>(
+              leafPool_.get(),
+              BOOLEAN(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<int8_t>>(
+              leafPool_.get(),
+              TINYINT(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<int16_t>>(
+              leafPool_.get(),
+              SMALLINT(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<int32_t>>(
+              leafPool_.get(),
+              INTEGER(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<int64_t>>(
+              leafPool_.get(),
+              BIGINT(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<float>>(
+              leafPool_.get(),
+              REAL(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<double>>(
+              leafPool_.get(),
+              DOUBLE(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<StringView>>(
+              leafPool_.get(),
+              VARCHAR(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<StringView>>(
+              leafPool_.get(),
+              VARBINARY(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<int32_t>>(
+              leafPool_.get(),
+              DATE(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<Timestamp>>(
+              leafPool_.get(),
+              TIMESTAMP(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          makeFlatVector<int64_t>(
+              kRows, [](auto) { return 0; }, nullEvery(1), DECIMAL(10, 2)),
+      });
+
+  auto sinkPtr = write(vector);
 
   dwio::common::ReaderOptions readerOptions{leafPool_.get()};
   auto reader = createReaderInMemory(*sinkPtr, readerOptions);
+  auto rowReader = createRowReaderWithSchema(std::move(reader), rowType);
 
-  ASSERT_EQ(reader->numberOfRows(), kRows);
-  ASSERT_EQ(*reader->rowType(), *schema);
+  auto result = BaseVector::create(rowType, 0, leafPool_.get());
+  rowReader->next(kRows, result);
 
-  auto rowReader = createRowReaderWithSchema(std::move(reader), schema);
-  assertReadWithReaderAndExpected(schema, *rowReader, data, *leafPool_);
+  assertEqualVectorPart(vector, result, 0);
 }
 
 } // namespace
--- a/velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
+++ b/velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
@@ -2206,8 +2206,10 @@ Status WriteArrowSerialize(
       ctx->GetScratchData<ParquetCType>(array.length(), &buffer));
 
   SerializeFunctor<ParquetType, ArrowType> functor;
-  RETURN_NOT_OK(
-      functor.Serialize(checked_cast<const ArrayType&>(array), ctx, buffer));
+  if (array.null_count() != array.length()) {
+    RETURN_NOT_OK(
+        functor.Serialize(checked_cast<const ArrayType&>(array), ctx, buffer));
+  }
   bool no_nulls = writer->descr()->schema_node()->is_required() ||
       (array.null_count() == 0);
   if (!maybe_parent_nulls && no_nulls) {
--- a/velox/vector/arrow/Bridge.cpp
+++ b/velox/vector/arrow/Bridge.cpp
@@ -748,12 +748,15 @@ void exportValues(
   const auto& type = vec.type();
   out.n_buffers = 2;
 
-  if (!rows.changed() && isFlatScalarZeroCopy(type)) {
+  if (!rows.changed() && !vec.values()) {
     // Arrow does not allow a nullptr for the values buffer. If the input vector
     // has no values buffer (all-null case), allocate an empty buffer of size 0.
-    auto values =
-        vec.values() ? vec.values() : AlignedBuffer::allocate<uint8_t>(0, pool);
-    holder.setBuffer(1, values);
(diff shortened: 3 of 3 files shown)


boneanxs:
Hi @majetideepak, could you please help review this, thanks!

majetideepak:
@boneanxs Sorry for the delay. I will take a look today or tomorrow.
CC: @JkSelf, @rui-mo for additional review.

velox/vector/arrow/Bridge.cpp
rui-mo, line 910: Could you please explain why a segmentation fault could occur for timestamps when all values are NULL? I assumed the values buffer would be allocated correctly and that NULLs would be skipped.

rui-mo (commented):
Thanks.

velox/vector/arrow/Bridge.cpp
boneanxs: In Velox, all-null vectors can have **nullptr values buffers**, The `gatherFromTimestampBuffer` function unconditionally dereferences the values buffer:
```c++
void gatherFromTimestampBuffer(
    const BaseVector& vec,
    const Selection& rows,
    TimestampUnit unit,
    Buffer& out) {
  auto src = (*vec.values()).as<Timestamp>();
```
If `vec.values()` returns nullptr, then `*vec.values()` causes a segmentation fault.
The same issue exists in gatherFromTimeBuffer.

The test added can reproduce this.

velox/vector/arrow/Bridge.cpp
```diff
@@ -748,12 +748,15 @@ void exportValues(
   const auto& type = vec.type();
   out.n_buffers = 2;
 
-  if (!rows.changed() && isFlatScalarZeroCopy(type)) {
+  if (!rows.changed() && !vec.values()) {
```
rui-mo: It looks like this fix is equivalent to changing the condition to
`if (!rows.changed() && (!vec.values() || isFlatScalarZeroCopy(type)))`.
What about scenarios where rows has changed? If values is nullptr, could this still lead to a segmentation fault?

Should we allocate an empty values buffer whenever values is nullptr?

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
```diff
@@ -2206,8 +2206,10 @@ Status WriteArrowSerialize(
       ctx->GetScratchData<ParquetCType>(array.length(), &buffer));
 
   SerializeFunctor<ParquetType, ArrowType> functor;
-  RETURN_NOT_OK(
-      functor.Serialize(checked_cast<const ArrayType&>(array), ctx, buffer));
+  if (array.null_count() != array.length()) {
+    RETURN_NOT_OK(
+        functor.Serialize(checked_cast<const ArrayType&>(array), ctx, buffer));
```
rui-mo, line 2211: If we allocate the values buffer with the correct size instead of leaving it empty, the original code should work, is that correct? Since this code is copied from Arrow, I’m wondering whether we can adjust `exportValues` so that the exported vector is fully compatible with Arrow.

rui-mo (commented):
Thanks.

rui-mo:
cc: @boneanxs A soft reminder to update this PR if you are still working on this, thanks.

mbasmanova (changes requested):
per @rui-mo 's comment

boneanxs:
sorry for the delay, will address comments recently

velox/vector/arrow/Bridge.cpp
boneanxs: Upstream Arrow has merged apache/arrow#48692, which makes the Parquet writer handle all-null arrays with a null values buffer. Can we upgrade/backport that fix in Velox's Arrow dependency?

If so, I can avoid allocating a zero-length buffer(Looks the arrow community doesn't encourage a zero-length buffer from the comment). I will also handle this before the rows.changed() gather path, since that path would otherwise dereference vec.values().

rui-mo: > Can we upgrade/backport that fix in Velox's Arrow dependency?

@boneanxs Yes, it would be preferable, thanks.

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
```diff
@@ -2181,8 +2181,11 @@ Status writeArrowSerialize(
       ctx->getScratchData<ParquetCType>(array.length(), &buffer));
 
   SerializeFunctor<ParquetType, ArrowType> functor;
-  RETURN_NOT_OK(
-      functor.serialize(checked_cast<const ArrayType&>(array), ctx, buffer));
+  // Backport apache/arrow#48692: all-null arrays may have no values buffer.
```
rui-mo: Do we need to consider all-null case in the `writeArrowZeroCopy`? It still DCHECKs that a missing values buffer implies `length == 0`.

Suggest fix: allow `data.values() == nullptr` only when `data.length() == 0` or `array.null_count() == array.length()`, allocate dummy values for the all-null non-empty case, and fail fast for any partially non-null array without values.

velox/vector/arrow/Bridge.cpp
```diff
@@ -890,12 +890,21 @@ void exportValues(
   const auto& type = vec.type();
   out.n_buffers = 2;
 
+  // Arrow C Data requires a values buffer large enough for the exported
+  // length, even when every value is null. Its contents are never read in
+  // that case. Handle this before gathering so a changed selection cannot
+  // dereference the missing Velox values buffer.
+  if (!vec.values()) {
```
rui-mo, line 897: A missing values buffer with non-null value could also be exported. Consider adding an explicit check before allocating the dummy buffer, e.g. `VELOX_CHECK_EQ(out.null_count, out.length, ...)`.

velox/vector/arrow/Bridge.cpp
```diff
@@ -890,12 +890,21 @@ void exportValues(
...
+    auto values = type->isBoolean()
+        ? AlignedBuffer::allocate<bool>(out.length, pool)
+        : AlignedBuffer::allocate<uint8_t>(
+              out.length * getArrowElementSize(type, options), pool);
```
rui-mo: Consider using `checkedMultiply<size_t>()` to guard overflow.

velox/vector/arrow/Bridge.cpp
```diff
@@ -890,12 +890,21 @@ void exportValues(
...
+    holder.setBuffer(1, std::move(values));
```
rui-mo: nit: `std::move(values)` is misleading because setBuffer takes `const BufferPtr&`.

rui-mo (commented):
@boneanxs Thanks! I left some comments, and could you please update the PR description as well?

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
boneanxs: Good point. We now guarantee in exportValues that any non-empty all-null Velox vector is exported with a correctly sized dummy values buffer. Therefore, no need to backport changes from arrow, reverted it instead of making complicates here.

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -1953,29 +1953,125 @@ TEST_F(ParquetWriterTest, selectiveFlatteningMixedEncodings) {
...
+  auto rowReader = createRowReaderFromReader(*reader, rowType);
 
-  ASSERT_EQ(reader->numberOfRows(), kRows);
-  ASSERT_EQ(*reader->rowType(), *schema);
+  auto result = BaseVector::create(rowType, 0, leafPool_.get());
+  rowReader->next(kRows, result);
 
-  auto rowReader = createRowReaderFromReader(*reader, schema);
-  assertReadWithReaderAndExpected(schema, *rowReader, data, *leafPool_);
+  assertEqualVectorPart(vector, result, 0);
```
rui-mo, line 2074: nit: could we keeping using `assertReadWithReaderAndExpected` to verify the total read count?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -1953,29 +1953,125 @@ TEST_F(ParquetWriterTest, selectiveFlatteningMixedEncodings) {
...
-  auto data = std::make_shared<RowVector>(
-      pool_.get(), schema, nullptr, kRows, std::vector<VectorPtr>{flatVector});
+      std::vector<VectorPtr>{
+          std::make_shared<FlatVector<bool>>(
+              leafPool_.get(),
+              BOOLEAN(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
```
rui-mo, line 1990: nit: could we add a local helper such as `makeAllNullFlatVector<NativeType>(type)` and use it for all fixed-width cases?

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -1953,29 +1953,125 @@ TEST_F(ParquetWriterTest, selectiveFlatteningMixedEncodings) {
...
+      {"string_col", VARCHAR()},
+      {"binary_col", VARBINARY()},
+      {"date_col", DATE()},
+      {"timestamp_col", TIMESTAMP()},
+      {"decimal_col", DECIMAL(10, 2)},
+  });
+
+  const int64_t kRows = 100;
+
+  // Create a nulls buffer with all bits set to null (0)
```
rui-mo, line 1973: nit: comments end with a period.

velox/vector/arrow/Bridge.cpp
```diff
@@ -890,15 +890,30 @@ void exportValues(
...
-        vec.values() ? vec.values() : AlignedBuffer::allocate<uint8_t>(0, pool);
+  // Arrow C Data requires a values buffer large enough for the exported
+  // length, even when every value is null. Its contents are never read in
+  // that case. Handle this before gathering so a changed selection cannot
+  // dereference the missing Velox values buffer.
+  if (!vec.values()) {
+    VELOX_CHECK_EQ(
+        out.null_count,
+        out.length,
+        "Missing values buffer is only supported for all-null vectors.");
```
rui-mo, line 901: nit: consider adding a test to cover the exception path.

rui-mo (approved):
Thanks for the update! @boneanxs This looks good to me, and just added some nits.

rui-mo:
@PingLiuPing Would you please help take a look? Thanks!

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
rui-mo: @boneanxs Does the above mean that, if we backport the Arrow change, we can avoid the extra allocation for non-empty all-null vectors?

velox/vector/arrow/Bridge.cpp
```diff
@@ -890,15 +890,30 @@ void exportValues(
...
+    auto values = type->isBoolean()
```
PingLiuPing: 902 - 907 are duplicate with line 918-922.
Can we refactor the function to:
  
```
if (vec.values() && !rows.changed() && isFlatScalarZeroCopy(type, options)) {
    holder.setBuffer(1, vec.values());
    return;
  }

  // Otherwise we will need a new buffer and copy the data.
  auto size = getArrowElementSize(type, options);
  auto values = type->isBoolean()
      ? AlignedBuffer::allocate<bool>(out.length, pool)
      : AlignedBuffer::allocate<uint8_t>(
            checkedMultiply<size_t>(out.length, size), pool);
  if (!vec.values()) {
    // Arrow C Data requires a values buffer large enough for the exported
    // length, even when every value is null.
    VELOX_CHECK_EQ(
        out.null_count,
        out.length,
        "Missing values buffer is only supported for all-null vectors.");
  } else if (type->kind() == TypeKind::TIMESTAMP) {
    gatherFromTimestampBuffer(vec, rows, options.timestampUnit, *values);
  } else if (...TIME...) {
    gatherFromTimeBuffer(vec, rows, *values);
  } else {
    gatherFromBuffer(*type, *vec.values(), rows, options, *values);
  }
  holder.setBuffer(1, values);
```

velox/dwio/parquet/tests/writer/ParquetWriterTest.cpp
```diff
@@ -1953,29 +1953,125 @@ TEST_F(ParquetWriterTest, selectiveFlatteningMixedEncodings) {
...
+              nullptr,
+              std::vector<BufferPtr>{}),
+          std::make_shared<FlatVector<Timestamp>>(
+              leafPool_.get(),
+              TIMESTAMP(),
+              nulls,
+              kRows,
+              nullptr,
+              std::vector<BufferPtr>{}),
+          makeFlatVector<int64_t>(
```
PingLiuPing: for decimal_col, this does not test the null value buffers, because `makeFlatVector` allocate a buffer.

PingLiuPing (changes requested):
Thanks.

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
boneanxs: No, we can't, the method from arrow `::arrow::ImportRecordBatch` explicitly enforce the value buffer can't be null if there's a length. So from velox side, we'd have to make the extra allocation.

The arrow changes only take effect if we manually create an Arrow Array, which is not suitable here. I was mistaken earlier.

PingLiuPing (approved):
Thanks.
Please update the PR description, 2 has been reverted.

velox/dwio/parquet/writer/arrow/ColumnWriter.cpp
rui-mo: Thanks, makes sense.

jja725 (approved):
Thanks for the fix!

boneanxs:
> Thanks. Please update the PR description, 2 has been reverted.

done, @PingLiuPing

kKPulla:
@mbasmanova - can you please review and approve this PR as your request for changes is not letting me merge this? Thanks!

mbasmanova (changes requested):
Thank you.

**Bridge.cpp:** the `!vec.values()` branch exports `values` without writing into it, so the consumer receives whatever the allocator returned. Both gather branches beside it fill every slot. Should the exported bytes be defined here?

**Tests:**

- Use `allocateNulls(size, pool, bits::kNull)` (BaseVector.h:1132) instead of spelling out the allocate and the `fillBits` — all three new sites do it by hand, and `ParquetWriterTest.allNulls` moved away from `makeNulls`, which it used before this change.
- `arrayGapWithAllNullTimestampValues` is a long name, and looking at why, it is two independent conditions in one test. "The gap between the two arrays forces the child selection to change" credits the wrong one: TIMESTAMP is never zero-copy, so an all-null flat Timestamp vector reaches the gather path on its own, and the array and the gap change nothing for it. The gap matters for a type that would otherwise be zero-copy, say an all-null BIGINT child, and that combination reaches `gatherFromBuffer(*type, *vec.values(), ...)` today. Could these be two tests, each named for its own case?
- `missingValuesWithNonNulls` tests the `FlatVector` constructor, not the bridge — please drop it. The `VELOX_CHECK` in `exportValues` it was meant to cover stays untested; is a vector in that state reachable at all?

PR #15871, opened 2025-12-30
feat: Add support for TZDIR environment variable

boneanxs:
This PR adds support for the `TZDIR` environment variable in the `tzdb` implementation.

Previously, the timezone database discovery logic was limited to standard system paths (e.g., `/usr/share/zoneinfo`, `/usr/share/zoneinfo/uclibc`). This limitation caused issues in large scale data center since if we want to upgrade tzdata dependency for velox, we have to upgrade the system's.

With this change, `__libcpp_tzdb_directory` now checks for `TZDIR` first. This aligns Velox's behavior with other standard C++ libraries (like abseil-cpp) and provides users with greater flexibility to configure the timezone database location.

### DIFF
--- a/velox/docs/develop/timestamp.rst
+++ b/velox/docs/develop/timestamp.rst
@@ -139,6 +139,16 @@ generally more efficient, but std::chrono does not handle time zone offsets
 such as ``+09:00``.  Timezone offsets are only supported in the API version
 that takes a timezone ID.
 
+Timezone Database Lookup
+------------------------
+
+Velox uses the IANA Time Zone Database to handle timezone conversions. It looks for the database in the following order:
+
+1. The directory specified by the ``TZDIR`` environment variable.
+2. ``/usr/share/zoneinfo/uclibc`` (on Linux).
+3. ``/usr/share/zoneinfo`` (on Linux).
+4. The path pointed to by ``/etc/localtime`` (on macOS).
+
 Casts
 -----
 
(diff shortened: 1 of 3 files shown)


boneanxs:
@kevinwilfong @pedroerp Hi, could you please help review this? thanks!

pedroerp:
Thanks @boneanxs . Could you also save this change as a patch in velox/external/tzdb/patches/ to make it more convenient to reapply it if we need to update the base library?

Would be nice to add some comments to the code as well.

pedroerp:
Could you also add a note somewhere in this doc about how Velox finds the timezone database (and ho to configure it using an env variable)?

https://facebookincubator.github.io/velox/develop/timestamp.html

boneanxs:
> Could you also add a note somewhere in this doc about how Velox finds the timezone database (and ho to configure it using an env variable)?
> 
> https://facebookincubator.github.io/velox/develop/timestamp.html

done. @pedroerp

pedroerp (approved):
Thank you

boneanxs:
> Thank you

Thanks! Let me know if anything else is needed. @pedroerp

boneanxs:
@pedroerp hi, is there any new comments for this?

pedroerp:
@boneanxs my bad, I missed adding the "read-to-merge" tag. Let me merge this manually

Issue #15847, opened 2025-12-23
CI: clang-tidy: No such file or directory


mbasmanova:
### Problem description

Build with GCC / Linux release with adapters CI job fails with:

https://github.com/facebookincubator/velox/actions/runs/20450501599/job/58762349765

```
Run git config --global --add safe.directory /__w/velox/velox
xargs: clang-tidy: No such file or directory
Error: Process completed with exit code 1.
```

CC: @pedroerp 

### System information

n/a

### CMake log

```bash

```

mbasmanova:
CC: @czentgr @majetideepak @assignUser @kgpai

mbasmanova:
CC: @kKPulla

kKPulla:
From the error message, likely related PR that was merged recently #15572 

@czentgr can you please take a look? Thanks!

kgpai:
Closing due to fix : #15872

mbasmanova:
Thank you, @czentgr

PR #15872, opened 2025-12-30
fix(ci): Clang-tidy installation

czentgr:
When clang-tidy was moved to the adapters build and fixed the path
(PR #15572)it was missed that it required an install that was done in a previous step.
The PR itself did not run clang-tidy because these was no actual 
code change.
One change to the run clang-tidy script is to return 0 if no STDOUT was
generated. Previously, it returned ok only if it matches with a warning
output but without any warning or error it would return fail as well. 

Addresses: https://github.com/facebookincubator/velox/issues/15847


pedroerp:
Thanks @czentgr. Let me know when this is ready to go and I can import it.

czentgr:
@pedroerp Please take a look. This was a run with some debug and a file being processed: https://github.com/facebookincubator/velox/actions/runs/20680840814/job/59374813282.
It passed as expected.

PR #15873, opened 2025-12-30
refactor(geo): Reduce memcpy in GeometryCollection serialization

jagill:
Summary:
Writing a GeometryCollection requires writing the length of the child geometry
before writing the child geometry. Previously we did this by writing to a
temporary std::string, then writing the length of that string to the output and
then copying the std::string to the output. Recursive GeometryCollections made
this worse because the "output" was actually just the parent std::string.

This commit writes a placeholder size (0), recording the offset. It then writes
the geometry, then writes the resulting length at the offset. This prevents all
the memcpys.

Removing this greatly simplifies our code an allows us to remove
VarbinaryWriter and almost all templating, which futher allows us to move most
implementations into the cpp file.

Differential Revision: D89928986

### DIFF
--- a/velox/expression/StringWriter.h
+++ b/velox/expression/StringWriter.h
@@ -135,6 +135,20 @@ class StringWriter : public UDFOutputString {
     append(std::string_view(input));
   }
 
+  template <typename T>
+  void writeAt(const T& value, size_t pos) {
+    static_assert(
+        std::is_trivially_copyable_v<T>, "T must be trivially copyable");
+    const char* dataToCopy = reinterpret_cast<const char*>(&value);
+    writeAt(dataToCopy, pos, sizeof(T));
+  }
+
+  void writeAt(const char* value, size_t pos, size_t length) {
+    VELOX_DCHECK(!finalized_);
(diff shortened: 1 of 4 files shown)


velox/expression/StringWriter.h
```diff
@@ -135,6 +135,20 @@ class StringWriter : public UDFOutputString {
     append(std::string_view(input));
   }
 
+  template <typename T>
+  void writeAt(const T& value, size_t pos) {
```
mbasmanova, line 139: Would be nice to document and add unit tests for this new API.

velox/expression/StringWriter.h
```diff
@@ -135,6 +135,20 @@ class StringWriter : public UDFOutputString {
...
+    static_assert(
+        std::is_trivially_copyable_v<T>, "T must be trivially copyable");
+    const char* dataToCopy = reinterpret_cast<const char*>(&value);
+    writeAt(dataToCopy, pos, sizeof(T));
+  }
+
+  void writeAt(const char* value, size_t pos, size_t length) {
```
mbasmanova, line 146: Would it make sense to use std::string_view instead of char* + length?

Issue #2262, opened 2022-08-11
Add Presto functions


mbasmanova:
Velox includes many of the PrestoSQL functions, but a few are still missing. It would great to add these.

Function coverage map: https://facebookincubator.github.io/velox/functions/coverage.html 

A subset of missing functions that would be most helpful to add.

### Array functions

all_match (lambda function)
any_match (lambda function)
array_average
array_frequency
array_has_duplicates
array_normalize
array_remove
array_union
flatten
repeat
sequence
shuffle
zip_with (lambda function)

### Map functions

map_from_entries
map_normalize
map_zip_with (lambda function)

### JSON functions

is_json_scalar
json_array_contains
json_array_get
json_array_length
json_extract
json_format
json_parse
json_size

### String functions

split_to_map
split_to_multimap
strrpos

### Regular expression functions

regexp_split

### Date and Time functions

timezone_hour
timezone_minute
week
week_of_year

### Mathematical functions

truncate

mbasmanova:
CC: @majetideepak @aditi-pandit

pramodsatya:
Thanks for sharing the list of missing functions. Following functions have been added:
Date and Time functions
week, week_of_year: https://github.com/facebookincubator/velox/pull/2287

JSON functions
is_json_scalar: https://github.com/facebookincubator/velox/pull/2291
json_array_length: https://github.com/facebookincubator/velox/pull/2294

Working on the following functions:
Mathematical functions
truncate

JSON functions
json_array_contains
json_array_get

jwyles-ahana:
I am going to start working on array_union.

aditi-pandit:
> I am going to start working on array_union.

There was a prior PR https://github.com/facebookincubator/velox/pull/867 that was abandoned. Maybe you can check with @kagamiori about it.

kagamiori:
> > I am going to start working on array_union.
> 
> There was a prior PR for array_union #867 that was abandoned. Maybe you can check with @kagamiori about it.

@aditi-pandit Thank you for bringing up this! It was a PR I didn't finish. I originally planned to rewrite it as a simple function once it's supported. (I just confirmed that it's not supported yet currently.) Do you need this function soon? Please feel free to take it over, or let me know if you want me to finish #867.

aditi-pandit:
> 

There isn't an urgency for the array_union function rightaway. We were just picking stuff from Masha's list above.

jwyles-ahana:
[...]

PR #15881, opened 2026-01-02
feat: Add FNV hash functions (fnv1_32, fnv1_64, fnv1a_32, fnv1a_64)

joyhaldar:
Adds FNV-1 and FNV-1a hash functions for 32-bit and 64-bit variants.

**Functions added:**
- `fnv1_32`
- `fnv1_64`
- `fnv1a_32`
- `fnv1a_64`

**Testing:**
- Unit tests with values from Presto's `TestVarbinaryFunctions.java`
- Expression fuzzer passed

Part of #2262

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -69,6 +69,12 @@ jobs:
           # Remove Haskell
           sudo rm -rf /opt/ghc || true
           sudo rm -rf /usr/local/.ghcup || true
+          # Remove Powershell
+          sudo rm -rf /usr/local/share/powershell || true
+          # Remove Node
+          sudo rm -rf /usr/local/lib/node_modules || true
+          # Remove Go
+          sudo rm -rf /opt/hostedtoolcache/go || true
           # Show after
           echo "New available disk space: " $(getAvailableSpace)
 
--- a/velox/docs/functions/presto/array.rst
+++ b/velox/docs/functions/presto/array.rst
@@ -132,6 +132,13 @@ Array Functions
         SELECT array_min(ARRAY[{-1, -2, -3, nan()]); -- -1
         SELECT array_min(ARRAY[{infinity(), nan()]); -- Infinity
 
+.. function:: array_max_by(array(T), function(T, U)) -> T()
+
+    Applies the provided function to each element, and returns the element that gives the maximum value.
+    ``U`` can be any orderable type. ::
+
+        SELECT array_max_by(ARRAY ['a', 'bbb', 'cc'], x -> LENGTH(x)) -- 'bbb'
+
 .. function:: array_normalize(array(E), E) -> array(E)
 
     Normalizes array ``x`` by dividing each element by the p-norm of the array. It is equivalent to ``TRANSFORM(array, v -> v / REDUCE(array, 0, (a, v) -> a + POW(ABS(v), p), a -> POW(a, 1 / p))``, but the reduce part is only executed once. Returns null if the array is null or there are null array elements. If ``p`` is 0, then the input array is returned. Only REAL and DOUBLE types are supported.
--- a/velox/docs/functions/presto/coverage.rst
+++ b/velox/docs/functions/presto/coverage.rst
@@ -62,12 +62,14 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with 
     table.coverage tr:nth-child(7) td:nth-child(9) {background-color: #6BA81E;}
     table.coverage tr:nth-child(8) td:nth-child(1) {background-color: #6BA81E;}
     table.coverage tr:nth-child(8) td:nth-child(2) {background-color: #6BA81E;}
+    table.coverage tr:nth-child(8) td:nth-child(3) {background-color: #6BA81E;}
     table.coverage tr:nth-child(8) td:nth-child(5) {background-color: #6BA81E;}
     table.coverage tr:nth-child(8) td:nth-child(7) {background-color: #6BA81E;}
     table.coverage tr:nth-child(8) td:nth-child(9) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(1) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(2) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(3) {background-color: #6BA81E;}
+    table.coverage tr:nth-child(9) td:nth-child(4) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(5) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(7) {background-color: #6BA81E;}
     table.coverage tr:nth-child(9) td:nth-child(9) {background-color: #6BA81E;}
@@ -121,6 +123,7 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with 
     table.coverage tr:nth-child(18) td:nth-child(2) {background-color: #6BA81E;}
     table.coverage tr:nth-child(18) td:nth-child(4) {background-color: #6BA81E;}
     table.coverage tr:nth-child(18) td:nth-child(5) {background-color: #6BA81E;}
+    table.coverage tr:nth-child(18) td:nth-child(7) {background-color: #6BA81E;}
     table.coverage tr:nth-child(19) td:nth-child(1) {background-color: #6BA81E;}
     table.coverage tr:nth-child(19) td:nth-child(2) {background-color: #6BA81E;}
     table.coverage tr:nth-child(19) td:nth-child(3) {background-color: #6BA81E;}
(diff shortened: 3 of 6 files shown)


joyhaldar:
Hello @mbasmanova, @pedroerp, @xiaoxmeng, could you kindly take a look at this PR when you have some time, please?

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,76 @@ struct Murmur3X64_128Function {
...
+
+  FOLLY_ALWAYS_INLINE
+  void call(out_type<int32_t>& result, const arg_type<Varbinary>& input) {
+    static constexpr uint32_t kOffsetBasis = 0x811c9dc5;
+    static constexpr uint32_t kPrime = 0x01000193;
+    const auto* data = reinterpret_cast<const unsigned char*>(input.data());
+    int32_t hash = kOffsetBasis;
+    for (size_t i = 0; i < input.size(); ++i) {
+      hash *= kPrime;
+      hash ^= data[i];
```
majetideepak: won't `input.data()[i]` work here?

majetideepak: We can remove the reinterpret_cast above.

joyhaldar:
Reference links to Presto Java implementation:

- Implementation: https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/FnvHash.java
- Tests: https://github.com/prestodb/presto/blob/eb321f82ee3f0bab5f33b98f3cdbe2399df9170b/presto-main-base/src/test/java/com/facebook/presto/operator/scalar/TestVarbinaryFunctions.java#L415

velox/functions/prestosql/BinaryFunctions.h
joyhaldar: Thank you for your review Deepak.

I tried removing the `reinterpret_cast` but tests fail for bytes > 127 due to sign extension for bytes > 127.
For example, byte `0xF5` (245) becomes -11 when char is signed, which corrupts the hash.
For reference, Presto uses [getUnsignedByte()](https://github.com/prestodb/presto/blob/eb321f82ee3f0bab5f33b98f3cdbe2399df9170b/presto-main-base/src/main/java/com/facebook/presto/operator/scalar/FnvHash.java#L39) which does `getByte(index) & 0xFF`, [Slice.java](https://github.com/airlift/slice/blob/f344b077afaa6189b1b14ae6cc5260de42f5e0e1/src/main/java/io/airlift/slice/Slice.java#L234).
I believe `reinterpret_cast<const unsigned char*>` is the C++ way to achieve the same.

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,76 @@ struct Murmur3X64_128Function {
...
+struct Fnv1_32Function {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE
+  void call(out_type<int32_t>& result, const arg_type<Varbinary>& input) {
+    static constexpr uint32_t kOffsetBasis = 0x811c9dc5;
+    static constexpr uint32_t kPrime = 0x01000193;
+    const auto* data = reinterpret_cast<const unsigned char*>(input.data());
+    int32_t hash = kOffsetBasis;
+    for (size_t i = 0; i < input.size(); ++i) {
```
pramodsatya: nit: auto i

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,76 @@ struct Murmur3X64_128Function {
...
+      hash *= kPrime;
+      hash ^= data[i];
+    }
+    result = hash;
+  }
+};
+
+template <typename T>
+struct Fnv1a_32Function {
```
pramodsatya, line 577: Could you abstract the hash calculation logic for `Fnv1a_32Function` and `Fnv1_32Function` into a common function, say `fnv1_32(const unsigned char* data, bool alternate)`, to avoid code repetition? Only the order of product and xor operations seems to differ for the two hash functions.

velox/functions/prestosql/BinaryFunctions.h
joyhaldar: Done.

velox/functions/prestosql/BinaryFunctions.h
joyhaldar: Thank you for your review Pramod. I have made the changes as suggested by you.

I also named the boolean parameter `xorFirst` instead of `alternate` for better readability (for me).

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,85 @@ struct Murmur3X64_128Function {
   }
 };
 
+/// Computes FNV hash for the given data.
+/// FNV-1 multiplies then XORs, FNV-1a XORs then multiplies.
+/// @param data Pointer to input bytes
+/// @param size Number of bytes
+/// @param xorFirst If true, use FNV-1a, if false, use FNV-1
+template <typename HashType>
+FOLLY_ALWAYS_INLINE HashType computeFnvHash(
```
pramodsatya, line 530: nit: move helper function inside anonymous namespace.

joyhaldar: Thank you for your review Pramod. 
I noticed the existing `fromHex` helper function in this file is not in an anonymous namespace. I was trying to follow the same pattern. Would you still like me to move it to an anonymous namespace? I am open to your suggestions.

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,85 @@ struct Murmur3X64_128Function {
...
+    const unsigned char* data,
+    size_t size,
+    HashType offsetBasis,
+    HashType prime,
+    bool xorFirst) {
+  HashType hash = offsetBasis;
+  for (auto i = 0; i < size; ++i) {
+    if (xorFirst) {
```
majetideepak: This `bool xorFirst` can be a template parameter as well.
We can then use constexpr, and the compiler will get rid of the branching.

majetideepak: `if constexpr (XORFirst)` after we add `template <typename HashType, bool XORFirst>`

joyhaldar: Thank you for the review Deepak.

I have made the change as suggested by you, thank you again.

velox/functions/prestosql/BinaryFunctions.h
pramodsatya: Yes, it would be better not to expose this helper function externally outside of these hash functions, wdyt? You could maybe cleanup `fromHex` and other such helper functions in a subsequent PR as well?

joyhaldar: Yes I could definitely do that Pramod.

velox/functions/prestosql/BinaryFunctions.h
```diff
@@ -524,4 +524,84 @@ struct Murmur3X64_128Function {
   }
 };
 
+/// Computes FNV hash for the given data.
+/// FNV-1 multiplies then XORs, FNV-1a XORs then multiplies.
+/// @param data Pointer to input bytes
+/// @param size Number of bytes
+/// @param xorFirst If true, use FNV-1a, if false, use FNV-1
```
majetideepak: This is not a param anymore.

joyhaldar: Done, thank you Deepak. Simplified the docstring to match the style used elsewhere in the file.

majetideepak:
@joyhaldar The Bias Fuzzer is failing. Can you check if the error is related? Thanks.

```
Biased functions: fnv1_32=20,fnv1_64=20,fnv1a_32=20,fnv1a_64=20
E20260107 18:51:23.963265    67 Exceptions.h:53] Line: /__w/velox/velox/velox/velox/expression/fuzzer/FuzzerToolkit.cpp:149, Function:operator(), Expression: left->equalValueAt(right.get(), row, row) Different values at idx '12': '0.8703478442568804' vs. 'null', Source: RUNTIME, ErrorCode: INVALID_STATE
```

joyhaldar:
> @joyhaldar The Bias Fuzzer is failing. Can you check if the error is related? Thanks.
> 
> ```
> Biased functions: fnv1_32=20,fnv1_64=20,fnv1a_32=20,fnv1a_64=20
> E20260107 18:51:23.963265    67 Exceptions.h:53] Line: /__w/velox/velox/velox/velox/expression/fuzzer/FuzzerToolkit.cpp:149, Function:operator(), Expression: left->equalValueAt(right.get(), row, row) Different values at idx '12': '0.8703478442568804' vs. 'null', Source: RUNTIME, ErrorCode: INVALID_STATE
> ```

Hello @majetideepak ,

I ran the expression fuzzer locally for 1 hour with the exact same parameters and seed `28347` as CI, with FNV functions prioritized in the test pool. It passed with 590k test cases and no failures. The CI error shows a floating point comparison ('0.8703478442568804' vs 'null'), but FNV functions return `int32_t` or `int64_t`. I couldn't reproduce this failure locally, so it appears to be unrelated to my changes in my understanding. Could I please request you to try re-triggering the CI to see if it could be a flaky test, please?

PR #15882, opened 2026-01-02
fix: Varbinary Iceberg partition value

PingLiuPing:
When a `VARBINARY` column is used as partition column, for example by identity or truncate transform, the transformed value is also `VARBINARY`. Such value must be base64-encoded before returning to upstream engine.

### DIFF
--- a/velox/connectors/hive/iceberg/IcebergDataSink.cpp
+++ b/velox/connectors/hive/iceberg/IcebergDataSink.cpp
@@ -21,6 +21,7 @@
 #include <boost/uuid/uuid_io.hpp>
 
 #include "velox/common/base/Fs.h"
+#include "velox/common/encode/Base64.h"
 #include "velox/connectors/hive/PartitionIdGenerator.h"
 #include "velox/connectors/hive/iceberg/TransformExprBuilder.h"
 #include "velox/exec/OperatorUtils.h"
@@ -48,7 +49,8 @@ template <>
 folly::dynamic extractPartitionValue<TypeKind::VARBINARY>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<StringView>>()->valueAt(row).str();
+  return encoding::Base64::encode(
+      child->as<SimpleVector<StringView>>()->valueAt(row));
 }
 
 template <>
--- a/velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
+++ b/velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
@@ -16,6 +16,9 @@
 
 #include <filesystem>
 
+#include <folly/json.h>
+
+#include "velox/common/encode/Base64.h"
 #include "velox/connectors/hive/iceberg/tests/IcebergTestBase.h"
 #include "velox/exec/tests/utils/AssertQueryBuilder.h"
 #include "velox/exec/tests/utils/PlanBuilder.h"
@@ -263,6 +266,26 @@ class TransformE2ETest : public test::IcebergTestBase {
         value.data(), value.size());
     return ((hash & 0x7FFFFFFF) % numBuckets);
   }
+
+  folly::dynamic getPartitionValuesFromCommitMessage(
+      const RowVectorPtr& rowVector) {
+    auto outputDirectory = TempDirectoryPath::create();
+    std::vector<test::PartitionField> partitionTransforms = {
+        {0, TransformType::kIdentity, std::nullopt},
+    };
+    const auto dataSink = createDataSinkAndAppendData(
+        {rowVector}, outputDirectory->getPath(), partitionTransforms);
+
+    auto commitMessages = dataSink->commitMessage();
+    VELOX_CHECK_EQ(commitMessages.size(), 1);
+    auto commitData = folly::parseJson(commitMessages[0]);
+    auto partitionDataJson =
+        folly::parseJson(commitData["partitionDataJson"].asString());
+    auto partitionValues = partitionDataJson["partitionValues"];
+    VELOX_CHECK_EQ(partitionValues.size(), 1);
+    dataSink->close();
+    return partitionValues[0];
+  }
 };
 
 TEST_F(TransformE2ETest, identity) {
@@ -342,6 +365,46 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
 
   ASSERT_EQ(actualPartitionNames, expectedPartitionNames)
       << "Partition folder names do not match expected values";
+}
+
+TEST_F(TransformE2ETest, varbinaryPartitionCommitMessage) {
+  std::vector<std::string> testData = {
+      "binarydata\x01\x02\x03",
+      "",
+      ".-_*/?%#&=+,;@!$\\\"'()<>:|[]{}^~`\n\r\t\u00A9",
+  };
+
+  for (const auto& binaryData : testData) {
+    SCOPED_TRACE(testing::Message() << "binaryData: " << binaryData);
+    auto rowVector = makeRowVector({
+        makeConstant(binaryData, 1, VARBINARY()),
+    });
+
+    auto partitionValue = getPartitionValuesFromCommitMessage(rowVector);
+    ASSERT_EQ(
+        partitionValue.asString(),
+        encoding::Base64::encode(binaryData.data(), binaryData.size()));
+  }
+}
+
+TEST_F(TransformE2ETest, nullPartitionValue) {
+  const std::vector<TypeKind> typeKinds = {
+      TypeKind::BOOLEAN,
+      TypeKind::INTEGER,
+      TypeKind::BIGINT,
+      TypeKind::VARCHAR,
+      TypeKind::VARBINARY,
+      TypeKind::TIMESTAMP,
+  };
+
+  for (const auto& typeKind : typeKinds) {
+    auto rowVector = makeRowVector({
+        makeNullConstant(typeKind, 1),
+    });
+
+    auto partitionValue = getPartitionValuesFromCommitMessage(rowVector);
+    ASSERT_TRUE(partitionValue.isNull());
+  }
 }
 
 TEST_F(TransformE2ETest, bucket) {


velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -48,7 +49,9 @@ template <>
 folly::dynamic extractPartitionValue<TypeKind::VARBINARY>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<StringView>>()->valueAt(row).str();
+  return encoding::Base64::encode(
```
yingsu00, line 52: Does this work for null values (Iceberg does allow NULL partition value)? From velox/vector/SimpleVector.h we could see the valueAt could return undefined value for nulls.
```
  // Concrete Vector types need to implement this themselves.
  // This method does not do bounds checking. When the value is null the return
  // value is technically undefined (currently implemented as default of T)
  virtual const T valueAt(vector_size_t idx) const = 0;
```
Same question for empty values

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +347,36 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
       << "Partition folder names do not match expected values";
 }
 
+// Verify that VARBINARY partition values are base64 encoded in commitMessage.
+TEST_F(TransformE2ETest, varbinaryPartitionCommitMessage) {
```
yingsu00, line 370: Can you please add test case for nulls?
e.g.
`SELECT * FROM t WHERE part IS NULL`
or storage partition with null binary.
Null is legal in Iceberg partition specs.

yingsu00, line 370: Could u please also add tests for empty binary? Note that (VARBINARY '') is distinct from null.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +347,36 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
...
+  auto commitData = folly::parseJson(commitMessages[0]);
+  auto partitionDataJson =
+      folly::parseJson(commitData["partitionDataJson"].asString());
+  auto partitionValues = partitionDataJson["partitionValues"];
+  ASSERT_EQ(partitionValues.size(), 1);
+
+  ASSERT_EQ(
+      partitionValues[0].asString(),
+      encoding::Base64::encode(binaryData.data(), binaryData.size()));
+  dataSink->close();
```
yingsu00: Right now the test checks encoding only. It would be better to check round trip:
decode(base64) == original

velox/connectors/hive/iceberg/IcebergDataSink.cpp
yingsu00: I actually just noticed other `extractPartitionValue` functions don't handle nulls and empty values themselves. Would be best if other overload functions also handle these cases in a uniform way.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +347,36 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
...
+  // Get the commit message and verify the partition value is base64 encoded.
+  auto commitMessages = dataSink->commitMessage();
+  ASSERT_EQ(commitMessages.size(), 1);
+  auto commitData = folly::parseJson(commitMessages[0]);
+  auto partitionDataJson =
+      folly::parseJson(commitData["partitionDataJson"].asString());
+  auto partitionValues = partitionDataJson["partitionValues"];
+  ASSERT_EQ(partitionValues.size(), 1);
+
+  ASSERT_EQ(
```
yingsu00: The current test only tests one value. Would it make sense to test multiple values, preferably with duplicates?

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +347,36 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
       << "Partition folder names do not match expected values";
 }
 
+// Verify that VARBINARY partition values are base64 encoded in commitMessage.
+TEST_F(TransformE2ETest, varbinaryPartitionCommitMessage) {
+  const std::string binaryData = "binarydata\x01\x02\x03";
```
yingsu00: Can you please add tests for special chars like '+', '/', '='?

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
yingsu00, line 370: The name "E2E" is somewhat confusing there. This changes only has the "encode" part, not the "decode" part, which could happen in other engines like Presto Java or Spark. They may handle special chars like '+' differently. Shall we add real E2E tests in Presto to make sure the encoded data, especially those including such special chars, can be decoded by Presto Java successfully?

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +347,36 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
...
+      {rowVector}, outputDirectory->getPath(), partitionTransforms);
+
+  // Get the commit message and verify the partition value is base64 encoded.
+  auto commitMessages = dataSink->commitMessage();
+  ASSERT_EQ(commitMessages.size(), 1);
+  auto commitData = folly::parseJson(commitMessages[0]);
+  auto partitionDataJson =
+      folly::parseJson(commitData["partitionDataJson"].asString());
+  auto partitionValues = partitionDataJson["partitionValues"];
+  ASSERT_EQ(partitionValues.size(), 1);
```
yingsu00: Would it make sense to add a test to check if '=' padding was added?
(Base64 pads to a multiple of 4 using =).

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks Ying, I will add some test case in https://github.com/prestodb/presto/pull/26889.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
PingLiuPing: Thanks Ying, the NULL partition value has been processed. See test case here https://github.com/prestodb/presto/pull/26889/changes#diff-aac07d3490ecd298fdf7135c1139c02804d81ba252e75e7d42b13e75b56449cdR77-R78

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks Ying, this PR fix a bug that I introduced when refactoring the code to use Velox expression evaluation infrastructure. The idea is to verify that the binary partition value get encoded.
There are other testcases to cover the partition logic.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks, the current test logic should be sufficient.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks. Added.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks, added.

PingLiuPing: Thanks Ying, there are test cases to test NULL specifically. See https://github.com/facebookincubator/velox/blob/59b492a9bce45f487f24b5cbae7dc845ea3d0827/velox/connectors/hive/iceberg/tests/TransformTest.cpp#L81
https://github.com/facebookincubator/velox/blob/59b492a9bce45f487f24b5cbae7dc845ea3d0827/velox/connectors/hive/iceberg/tests/PartitionNameTest.cpp#L195
I will add a case here to verify the serialized partition value explicitly for a NULL value.

PingLiuPing:
@mbasmanova  Would you help to have a look at this PR? It fixes a bug that I introduced when refactoring the code to use Velox expression evaluation infrastructure, I forgot to pick-up this special logic. Thank you.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -48,7 +49,9 @@ template <>
...
+      child->as<SimpleVector<StringView>>()->valueAt(row).data(),
```
mbasmanova: nit: can we capture the value of ...valueAt(row) in a variable to avoid duplicate calls? Alternatively, would be nice to add a version of encoding::Base64::encode API that takes std::string_view.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -263,6 +266,35 @@ class TransformE2ETest : public test::IcebergTestBase {
...
+
+  // Helper to verify that a partition value is correctly Base64 encoded.
+  static void verifyPartitionValue(
+      const folly::dynamic& partitionValue,
+      const std::string& expectedData) {
+    ASSERT_EQ(
+        partitionValue.asString(),
+        encoding::Base64::encode(expectedData.data(), expectedData.size()));
+    ASSERT_EQ(
+        encoding::Base64::decode(partitionValue.asString()), expectedData);
```
mbasmanova: The seconds assertion is redundant as it simply verifies that decode(encode(expected)) == expected.

> verifyPartitionValue

This method is specific for VARBINARY partition values, but its name is generic. I suggest to remove this method altogether.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +376,60 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
...
+    auto rowVector = makeRowVector({
+        makeConstant(binaryData, 1, VARBINARY()),
+    });
+    std::vector<test::PartitionField> partitionTransforms = {
+        {0, TransformType::kIdentity, std::nullopt},
+    };
+    auto partitionValue =
+        getPartitionValuesFromCommitMessage(rowVector, partitionTransforms);
+    verifyPartitionValue(partitionValue, binaryData);
+  }
```
mbasmanova: This test appears to contains a lot of copy-paste. Let's create a list of values to test, then loop other these.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
```diff
@@ -344,6 +376,60 @@ TEST_F(TransformE2ETest, partitionNamingConventions) {
...
+}
+
+TEST_F(TransformE2ETest, nullPartitionValue) {
+  auto rowVector = makeRowVector({
+      makeNullConstant(TypeKind::VARBINARY, 1),
```
mbasmanova: Should we test more types?

velox/connectors/hive/iceberg/IcebergDataSink.cpp
I see `std::string Base64::encode(std::string_view text) ` exist already.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks, reverted the code.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks.

velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp
PingLiuPing: Thanks, added more data types.

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -48,7 +49,8 @@ template <>
 folly::dynamic extractPartitionValue<TypeKind::VARBINARY>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<StringView>>()->valueAt(row).str();
+  const auto value = child->as<SimpleVector<StringView>>()->valueAt(row);
+  return encoding::Base64::encode(value.data(), value.size());
```
mbasmanova: nit: Can we not just pass 'value' and let it be implicitly casted to std::string_view?

PingLiuPing: Thanks, let me rebase and try, since https://github.com/facebookincubator/velox/pull/15946 is merged, now we can implicitly cast StringView to std::string_view.

kevinwilfong:
@PingLiuPing Could you rebase on top of https://github.com/facebookincubator/velox/pull/15981 so we can get a clean test run (thanks for that fix!)

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -48,7 +49,8 @@ template <>
 folly::dynamic extractPartitionValue<TypeKind::VARBINARY>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<StringView>>()->valueAt(row).str();
+  return encoding::Base64::encode(
+      child->as<SimpleVector<StringView>>()->valueAt(row));
```
juwentus1234, line 53: nit: using asChecked

PR #15883, opened 2026-01-02
Fix minor typo in index join test base

x77a1:
Summary: Fix minor typo in index join test base

Differential Revision: D89994278


xiaoxmeng (approved):
@x77a1 thank you for the fix!

PR #15885, opened 2026-01-02
feat: Add runtime stats in RowReader for the filter to index conversion

xiaoxmeng:
Summary:
Add a runtime statistic `numIndexFilterConversions` to track when ScanSpec filters are successfully converted to index bounds for index-based filtering. This metric helps monitor the effectiveness of cluster index pruning by reporting the number of index columns that were converted from filters.

The stat is:
- Defined as a static constant `kNumIndexFilterConversions` in `RowReader` base class
- Updated via `addThreadLocalRuntimeStat` when filters are converted to index bounds
- Documented in the metrics.rst reference guide under a new "Row Reader" section

This enables observability into how often and how many columns are leveraging index-based filtering for row/stripe pruning.

Differential Revision: D90031012

### DIFF
--- a/velox/dwio/common/Reader.h
+++ b/velox/dwio/common/Reader.h
@@ -44,6 +44,13 @@ class RowReader {
  public:
   static constexpr int64_t kAtEnd = -1;
 
+  /// Runtime stat names.
+  /// Tracks the number of index columns that were converted from ScanSpec
+  /// filters to index bounds for index-based filtering (e.g., cluster index
+  /// pruning in Nimble).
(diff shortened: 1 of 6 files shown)


zacw7 (approved):
Thanks for adding the stats!

velox/dwio/common/Reader.h
```diff
@@ -44,6 +44,13 @@ class RowReader {
  public:
   static constexpr int64_t kAtEnd = -1;
 
+  // Runtime stat names.
```
zacw7: remove this line?

PR #15891, opened 2026-01-04
fix(function): Adjust index offset for reservoir_sample when merging samples

gggrace14:
Seeing the certain samples in the result do not belong to the group key.
Fix a bug when addIntermediateResults merge two samples where otherOffset
is not being added when preparing the source index mapping for copying
otherSampleVector. This potentially results in wrong samples being copied
when merging samples.


amitkdutta (approved):

amitkdutta:
Will be great if its possible to add a unit test

PR #15894, opened 2026-01-05
perf: Optimize stats collection in ExchangeClient::collectStatsLocked()

lingbin:
Optimize `ExchangeClient::collectStatsLocked()` to reduce hash table
lookups when collecting metrics from ExchangeSources. Instead of using
'stats.count() + stats.insert() + stats[name]' (3 lookups per entry),
use `try_emplace()` which provides both insertion and access in a single
operation (1 lookup per entry).

Alose use `insert_or_assign()` for the "set or overwrite" pattern.
This is more efficient than operator[] as it avoids a potentially
wasteful default-construction followed by an assignment.


lingbin:
@pedroerp   just a gentle ping. Is there anything you’d like me to address before it can be merged?  Thanks.

PR #15896, opened 2026-01-05
fix(ci): Get changes for scheduled run

czentgr:
Moving clang-tidy to adapters also caused the scheduled adapters run to fail because it is dependent on get-changes job to succeed.
If the job was scheduled skip the changes files determination and simply indicate to not run clang-tidy.

### DIFF
--- a/.github/workflows/linux-build-base.yml
+++ b/.github/workflows/linux-build-base.yml
@@ -46,27 +46,33 @@ jobs:
           HEAD_REF: ${{ github.head_ref }}
           PR_OWNER: ${{ github.event.pull_request.head.repo.owner.login }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          JOB_TRIGGER: ${{ github.event_name }}
         run: |
-          merge_base_commit=$(gh api -q '.merge_base_commit.sha' \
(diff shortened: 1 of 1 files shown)


majetideepak:
@czentgr can we exlude the push to main branch as well. Those CI jobs seem to be failing as well.

.github/workflows/linux-build-base.yml
```diff
@@ -46,27 +46,33 @@ jobs:
...
+              echo 'files<<EOF'
+              cat changed_files.txt
+              echo 'EOF'
+            } >> "$GITHUB_OUTPUT"
 
-          if grep -qE $cpp_files changed_files.txt; then
-            echo "run_clang_tidy=true" >> "$GITHUB_OUTPUT"
+            if grep -qE $cpp_files changed_files.txt; then
+              echo "run_clang_tidy=true" >> "$GITHUB_OUTPUT"
+            fi
```
majetideepak, line 73: Should we have an `else` here?

majetideepak: I am guessing not since it did not exist before.

kgpai:
@czentgr can you rebase against latest, I think there might be some race-condition/conflict ?

czentgr:
@kgpai I rebased. So we should be good to go once it is green.

kgpai:
@czentgr This diff has been failing to land for some wierd reason. Will land it today.

PR #15897, opened 2026-01-05
feat(build): Enable coroutines

czentgr:
Note, we need changes when Velox is updated in PrestoC++ to align and create a new dependency image.
This breaks PrestoC++ on update otherwise (CI, remote functions).


czentgr:
Commit to increase disk space because docker bake failed in a previous action due to insufficient disk space.
Created https://github.com/facebookincubator/velox/pull/15910 to handle it.

kgpai (approved):

PR #15913, opened 2026-01-07
feat(cudf): Add support for cudf::packed_table into CudfVector

dan13bauer:
A `CudfVector` is exchanged by packing the underlying `cudf::table` into a `cudf::packed_table`. This change allows to construct a `CudfVector` from a `cudf::packed_table` instead of using a `cudf::table`, avoiding a copy. CudfVector then exports a table view that either points to the underlying `cudf::table` or `cudf::packed_table`, whatever is available. In order to take full advantage of this change, this PR also changes some of the operators to use the table view rather than the table.

### DIFF
--- a/python/test/test_runner.py
+++ b/python/test/test_runner.py
@@ -54,12 +54,24 @@ class TestPyVeloxRunner(unittest.TestCase):
         plan_builder = PlanBuilder().values()
         LocalRunner(plan_builder.get_plan_node())
 
-    def test_executed_twice(self):
-        # Ensure the runner fails if it is executed twice.
-        plan_builder = PlanBuilder().values()
+    def test_execute_twice(self):
+        # Ensure a runner can be executed twice.
+        vector = to_velox(
+            pyarrow.record_batch([pyarrow.array(list(range(10)))], names=["c0"])
+        )
+
+        plan_builder = PlanBuilder().values([vector])
         runner = LocalRunner(plan_builder.get_plan_node())
-        runner.execute()
-        self.assertRaises(RuntimeError, runner.execute)
+
+        total_size = 0
+        for vector in runner.execute():
+            total_size += vector.size()
+        self.assertEqual(total_size, 10)
+
+        total_size = 0
+        for vector in runner.execute():
+            total_size += vector.size()
+        self.assertEqual(total_size, 10)
 
     def test_values(self):
         vectors = []
--- a/velox/common/serialization/DeserializationRegistry.h
+++ b/velox/common/serialization/DeserializationRegistry.h
@@ -53,6 +53,16 @@ struct is_templated_create<
     T,
     std::void_t<decltype(T::template create<T>(
         std::declval<folly::dynamic>()))>> : std::true_type {};
+
+template <class, class = void>
+struct is_templated_create_with_context : std::false_type {};
+
+template <class T>
+struct is_templated_create_with_context<
+    T,
+    std::void_t<decltype(T::template create<T>(
+        std::declval<folly::dynamic>(),
+        std::declval<void*>()))>> : std::true_type {};
 } // namespace detail
 
 template <class T>
@@ -62,6 +72,17 @@ void registerDeserializer() {
         T::getClassName(), T::template create<T>);
   } else {
     DeserializationRegistryForSharedPtr().Register(
+        T::getClassName(), T::create);
+  }
+}
+
+template <class T>
+void registerDeserializerWithContext() {
+  if constexpr (detail::is_templated_create_with_context<T>::value) {
+    DeserializationWithContextRegistryForSharedPtr().Register(
+        T::getClassName(), T::template create<T>);
+  } else {
+    DeserializationWithContextRegistryForSharedPtr().Register(
         T::getClassName(), T::create);
   }
 }
--- a/velox/connectors/hive/iceberg/IcebergDataSink.cpp
+++ b/velox/connectors/hive/iceberg/IcebergDataSink.cpp
@@ -40,14 +40,14 @@ folly::dynamic extractPartitionValue(
     const VectorPtr& child,
     vector_size_t row) {
   using T = typename TypeTraits<Kind>::NativeType;
-  return child->as<SimpleVector<T>>()->valueAt(row);
+  return child->asChecked<SimpleVector<T>>()->valueAt(row);
 }
 
 template <>
 folly::dynamic extractPartitionValue<TypeKind::VARCHAR>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<StringView>>()->valueAt(row).str();
+  return child->asChecked<SimpleVector<StringView>>()->valueAt(row).str();
 }
 
 template <>
@@ -55,14 +55,14 @@ folly::dynamic extractPartitionValue<TypeKind::VARBINARY>(
     const VectorPtr& child,
     vector_size_t row) {
   return encoding::Base64::encode(
-      child->as<SimpleVector<StringView>>()->valueAt(row));
+      child->asChecked<SimpleVector<StringView>>()->valueAt(row));
 }
 
 template <>
 folly::dynamic extractPartitionValue<TypeKind::TIMESTAMP>(
     const VectorPtr& child,
     vector_size_t row) {
-  return child->as<SimpleVector<Timestamp>>()->valueAt(row).toMicros();
+  return child->asChecked<SimpleVector<Timestamp>>()->valueAt(row).toMicros();
 }
 
 class IcebergFileNameGenerator : public FileNameGenerator {
--- a/velox/connectors/tpch/TpchConnector.h
+++ b/velox/connectors/tpch/TpchConnector.h
@@ -34,9 +34,27 @@ class TpchColumnHandle : public ColumnHandle {
  public:
   explicit TpchColumnHandle(const std::string& name) : name_(name) {}
 
-  const std::string& name() const {
+  const std::string& name() const override {
     return name_;
   }
+
+  folly::dynamic serialize() const override {
+    folly::dynamic obj = folly::dynamic::object;
+    obj["name"] = TpchColumnHandle::getClassName();
+    obj["columnName"] = name_;
+    return obj;
+  }
+
+  static std::shared_ptr<TpchColumnHandle> create(const folly::dynamic& obj) {
+    auto name = obj["columnName"].asString();
+    return std::make_shared<TpchColumnHandle>(name);
+  }
+
+  static void registerSerDe() {
+    registerDeserializer<TpchColumnHandle>();
+  }
+
+  VELOX_DEFINE_CLASS_NAME(TpchColumnHandle)
 
  private:
   const std::string name_;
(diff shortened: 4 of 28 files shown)


zoltan:
@dan13bauer can you document here please the before-after situation regarding query stability?

dan13bauer:
We see a positive impact on the exchange performance at the cost of higher memory usage. We're currently investigating the higher memory usage. Here's an analysis from Presto's statistics on the impact Exchange operator:
```
  Exchange Runtime Statistics Comparison

  I've analyzed all Exchange operators across both configurations. Here's the comparison:

  Exchange.1566 (Stage 0, Small Data - 175 positions)

  | Metric             | cudfvector_original | cudfvector_packedcolumns | Difference         |
  |--------------------|---------------------|--------------------------|--------------------|
  | driverCpuTimeNanos | 822,213 ns          | 567,270 ns               | -31% (faster)      |
  | inputBatches       | 2                   | 2                        | same               |
  | outputBatches      | 2                   | 2                        | same               |
  | queuedWallNanos    | 54,000 ns           | 75,000 ns                | +39%               |
  | inputDataSize      | 4,747 bytes         | 4,864 bytes              | +2.5%              |
  | getOutputWall      | 299,041 ns          | 9,940 ns                 | -97% (much faster) |

  Exchange.1565 (Stage 1, Small Data - 175 positions)

  | Metric             | cudfvector_original | cudfvector_packedcolumns | Difference         |
  |--------------------|---------------------|--------------------------|--------------------|
  | driverCpuTimeNanos | 1,199,503 ns        | 859,120 ns               | -28% (faster)      |
  | inputBatches       | 4                   | 4                        | same               |
  | outputBatches      | 4                   | 4                        | same               |
  | queuedWallNanos    | 88,000 ns           | 144,000 ns               | +64%               |
  | getOutputWall      | 535 μs              | 16.49 μs                 | -97% (much faster) |

  Exchange.1798 (Stage 2, Medium Data - 350 positions)

  | Metric             | cudfvector_original | cudfvector_packedcolumns | Difference         |
  |--------------------|---------------------|--------------------------|--------------------|
  | driverCpuTimeNanos | 2,175,570 ns        | 986,941 ns               | -55% (faster)      |
  | inputBatches       | 4                   | 4                        | same               |
  | outputBatches      | 4                   | 4                        | same               |
  | queuedWallNanos    | 127,000 ns          | 157,000 ns               | +24%               |
  | inputDataSize      | 9,494 bytes         | 10,112 bytes             | +6.5%              |
  | getOutputWall      | 1.42 ms             | 20.45 μs                 | -99% (much faster) |

  Exchange.1563 (Stage 3, LARGE Data - 6 billion positions)

  | Metric             | cudfvector_original            | cudfvector_packedcolumns       | Difference            |
  |--------------------|--------------------------------|--------------------------------|-----------------------|
  | driverCpuTimeNanos | 1,286,905,373 ns (1.29s)       | 10,276,112 ns (10.3ms)         | -99.2% (125x faster!) |
  | inputBatches       | 1,220                          | 1,220                          | same                  |
  | outputBatches      | 1,220                          | 1,220                          | same                  |
  | queuedWallNanos    | 13,165,000 ns                  | 5,751,000 ns                   | -56% (faster)         |
  | inputDataSize      | 263,999,547,196 bytes (246 GB) | 263,999,754,688 bytes (246 GB) | ~same                 |
  | getOutputWall      | 1.31 s                         | 4.34 ms                        | -99.7% (302x faster!) |

  Key Findings

  The cudfvector_packedcolumns configuration shows dramatic performance improvements in the Exchange operators:

  1. Small data exchanges (1566, 1565): 28-31% faster CPU time, 97% faster output operations
  2. Medium data exchange (1798): 55% faster CPU time, 99% faster output operations
  3. Large data exchange (1563): 99.2% faster CPU time - this is the most significant improvement!
    - Original: 1.29 seconds CPU time
    - Packed columns: 10.3 milliseconds CPU time
    - 125x speedup on the largest data exchange

  The improvement scales with data size, with the most dramatic gains seen when exchanging the 246 GB dataset (6 billion rows). This suggests the packed column \
format significantly reduces
  serialization/deserialization overhead in data exchange operations.
```

dan13bauer:
Analysis of the memory consumption below.
Task 3 and 4 are the tasks where the HashJoins are happening.
```
  1. Which configuration consumes more memory?
  ┌──────────────────────┬──────────────┬──────────────┬────────────────────────────────┐
  │        Metric        │ main-testing │ packed_table │           Difference           │
  ├──────────────────────┼──────────────┼──────────────┼────────────────────────────────┤
  │ Combined Peak Memory │ 18.36 GB     │ 36.52 GB     │ packed_table uses 2x more peak │
  ├──────────────────────┼──────────────┼──────────────┼────────────────────────────────┤
  │ Total Allocations    │ 1151.93 GB   │ 1101.81 GB   │ main-testing ~4.3% more        │
  ├──────────────────────┼──────────────┼──────────────┼────────────────────────────────┤
  │ Sum of Task Peaks    │ 26.51 GB     │ 55.94 GB     │ packed_table ~2x more          │
  └──────────────────────┴──────────────┴──────────────┴────────────────────────────────┘
  Key Finding: The packed_table configuration has ~2x higher peak memory (36.52 GB vs 18.36 GB combined across workers). However, the total allocation throughput is similar (~1.1 TB for both).

  Per-worker peak:
  - main-testing: 8.95 GB (w1) + 9.41 GB (w2)
  - packed_table: 18.34 GB (w1) + 18.18 GB (w2)

  2. Top memory-consuming tasks

  Both configurations have the same task pattern - tasks 4.0.0.0 and 4.0.1.0 dominate:
  ┌──────────────┬──────────────────┬─────────────────────┬─────────────────────┐
  │ Task Pattern │ Total Allocation │ Peak (main-testing) │ Peak (packed_table) │
  ├──────────────┼──────────────────┼─────────────────────┼─────────────────────┤
  │ *.4.0.0.0    │ ~276 GB          │ 8.80 GB             │ 16.76 GB            │
  ├──────────────┼──────────────────┼─────────────────────┼─────────────────────┤
  │ *.4.0.1.0    │ ~276 GB          │ 8.80 GB             │ 16.76 GB            │
  ├──────────────┼──────────────────┼─────────────────────┼─────────────────────┤
  │ *.5.0.x.0    │ ~136 GB each     │ ~1 GB               │ ~1 GB               │
  ├──────────────┼──────────────────┼─────────────────────┼─────────────────────┤
  │ *.3.0.x.0    │ ~136 GB each     │ 747 MB              │ 7.48 GB             │
  └──────────────┴──────────────────┴─────────────────────┴─────────────────────┘
  The critical difference: Tasks 4.0.0.0, 4.0.1.0, 3.0.0.0, and 3.0.1.0 have ~2x to 10x higher peak memory in packed_table while having identical total allocations.
  ```
  
Stage4_Query9

**Comment by dan13bauer** [comment]
The higher memory usage was caused by the new approach that frees memory when `CudfVector` goes out of scope since `CudfVector::release()` is no longer called. This was an issue in the HashJoin operator where both the input vector and a freshly allocated output vector existed simultaneously. Releasing the input vector after it is no longer needed solves this issue and the peak memory usage is now slightly lower with the new approach:
```
● Memory Analysis Summary

  1. Memory Consumption Comparison
  ┌────────────────────────────┬───────────────┬───────────────────┐
  │       Configuration        │ Combined Peak │ Total Allocations │
  ├────────────────────────────┼───────────────┼───────────────────┤
  │ main-testing (reference)   │ 18.36 GB      │ 1151.93 GB        │
  ├────────────────────────────┼───────────────┼───────────────────┤
  │ packed_table               │ 36.52 GB      │ 1101.81 GB        │
  ├────────────────────────────┼───────────────┼───────────────────┤
  │ packed_table_v2 (with fix) │ 17.62 GB      │ 1151.94 GB        │
  └────────────────────────────┴───────────────┴───────────────────┘
  Key findings:
  - packed_table has 2x higher peak memory than main-testing (36.52 GB vs 18.36 GB)
  - packed_table_v2 (with the early release fix) reduced peak memory to 17.62 GB - slightly better than main-testing
  - Total allocations are similar across all configurations (~1151 GB)

  2. Top Memory-Consuming Tasks

  The dominant tasks are 4.0.x.0 (HashJoin tasks):
  ┌─────────┬───────────────────┬───────────────────┬──────────────────────┐
  │  Task   │ main-testing Peak │ packed_table Peak │ packed_table_v2 Peak │
  ├─────────┼───────────────────┼───────────────────┼──────────────────────┤
  │ 4.0.0.0 │ 8.80 GB           │ 16.76 GB          │ 8.80 GB              │
  ├─────────┼───────────────────┼───────────────────┼──────────────────────┤
  │ 4.0.1.0 │ 8.80 GB           │ 16.76 GB          │ 8.80 GB              │
  ├─────────┼───────────────────┼───────────────────┼──────────────────────┤
  │ 3.0.0.0 │ 747 MB            │ 7.48 GB           │ 747 MB               │
  ├─────────┼───────────────────┼───────────────────┼──────────────────────┤
  │ 3.0.1.0 │ 747 MB            │ 7.48 GB           │ 747 MB               │
  └─────────┴───────────────────┴───────────────────┴──────────────────────┘
  The fix worked. The early release of cudfInput.reset() in HashJoin:
  - Reduced HashJoin peak memory from 16.76 GB → 8.80 GB (back to baseline)
  - Reduced tasks 3.0.x.0 from 7.48 GB → 747 MB (10x reduction)

  The overall combined peak memory went from 36.52 GB (packed_table) → 17.62 GB (packed_table_v2), which is now slightly better than main-testing's 18.36 GB.
  ```

velox/experimental/cudf/vector/CudfVector.h
```diff
@@ -29,32 +30,58 @@
...
   uint64_t estimateFlatSize() const override;
 
  private:
+  // Owned table - set when constructed from cudf::table, or materialized
+  // on release() when constructed from packed_table.
   std::unique_ptr<cudf::table> table_;
+
+  // Packed table - set when constructed from packed_table.
+  // Keeps the packed data alive while tabView_ references it.
+  std::unique_ptr<cudf::packed_table> packedTable_;
```
devavret: nit: these can be in a variant to clarify that the CudfVector holds only one of these

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -1262,27 +1266,32 @@ RowVectorPtr CudfHashJoinProbe::getOutput() {
...
       break;
     default:
       VELOX_FAIL("Unsupported join type: ", joinNode_->joinType());
   }
 
+  // Release input CudfVector to free GPU memory before creating output.
+  // This reduces peak memory from (input + output) to max(input, output).
+  // cudfInput must be released first since input_.reset() only decrements
+  // the refcount while cudfInput still holds a reference.
+  cudfInput.reset();
```
devavret, line 1294: your post https://github.com/facebookincubator/velox/pull/15913#issuecomment-3729430308 says that this brought down the excess memory consumption back to pre-packed_table levels but this doesn't seem to be caused by the packed_table change. It seems that this optimization was available before (for all except antiJoin where it was already made). Am I understanding this correctly @shrshi? That means there's a memory consumption regression still lurking.

dan13bauer: I believe this is due to the packed_table change. Previously, on line 1238, the memory of the `cudfInput` was moved out and then released, see here:
https://github.com/facebookincubator/velox/pull/15913/changes/BASE..833b1f2c2840a5ced52c5e0537ede22535194e0a#diff-f2c7c33d4cefb3267976ab8973a5ade89ea9cf7fee65c6e97952693766dbe8b5L1238
After the change to add `packed_table`, the memory wasn't released but referenced through the view and only released when the method terminated, thus having both memory for the `cudfInput` and `cudfOutput` vector allocated simultaneously, increasing peak memory allocation. Above `cudfInput.reset()` fixes that.

velox/experimental/cudf/vector/CudfVector.h
dan13bauer: Yes, true! Will fix that.

marios-angelis:
I executed both versions (default and packed_table) with the following 2 configurations:
1. 1 worker with single execution mode enabled.
2. 2 workers with cudfExchange enabled.

For some queries, the peak memory usage is not deterministic across runs (for example for Q4 is can go from 5 to 7GBs), but in general, from the following figures, I don't observe any memory usage regression in the packed_table version. For the two-worker configuration, the peak memory shown in the figures represents the combined peak memory of both workers.

memory_comparison_multigpu_plot
memory_comparison_single_gpu_plot

velox/experimental/cudf/vector/CudfVector.h
dan13bauer: Fix in https://github.com/facebookincubator/velox/pull/15913/commits/1844c6c006caf2a79326550f4ab37a56bc4edff7

velox/experimental/cudf/exec/CudfHashJoin.cpp
devavret: In the old code, `cudfInput` was released at line 1238 but not deallocated. It just transferred ownership. The deallocation still happened at the end of `CudfHashJoinProbe::getOutput` scope. `cudfInput` and `cudfOutput` have to be alive at the same time for joins to work and I don't think that has changed. `cudfInput` is a downcasted `input_` and `input_` is reset in the next line anyway so that shouldn't have mattered. I'm still confused as to how this reset is affecting memory usage. But I can approve because memory usage has demonstrably not regressed and i think the reset is redundant but it doesn't change the behaviour.

czentgr:
@dan13bauer Can you make a new commit/rebase and retry? There is a race condition that might happen.
I have a change in one of my PRs. Maybe I should separate this out and check in now if it starts affecting the CI too badly.

PR #15914, opened 2026-01-07
Extract NameToIndex from RowType into standalone header

Magoja:
Summary:
## TL/DR
We seemed to have introduced O(1) name lookup for `RowType`. However, we are exposing the implementation detail `F14FastSet` directly in the interface.
This creates unnecessary API dependencies that Velox shouldn't be responsible for.

Here is my suggestion:
- Create a wrapper, and provide the minimal interfaces.
- Put it in a separate header, so that we can easily unittest.
- In future, we can evolve it with the minimal interfaces

This is exactly what this diff tries to achieve.

I hope this is reasonable

Following is AI generated summary
---
Refactors the NameToIndex lookup structure from RowType into a separate header file to improve code organization and reusability. The NameToIndex class provides a memory-efficient mapping from string names to uint32_t indices, which was previously embedded as an implementation detail within RowType.

Differential Revision: D90140810

### DIFF
--- a/velox/type/NameToIndex.h
+++ b/velox/type/NameToIndex.h
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


kgpai (commented):
Thanks @Magoja -  Looks fine to me, I dont have any strong opinion on this. 
Asking other folks to also have a look..
cc: @Yuhta .

velox/type/NameToIndex.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include <folly/Hash.h>
+#include <folly/container/F14Set.h>
+
+namespace facebook::velox {
+
+/// A lookup structure that maps string names to uint32_t indices.
+/// This is written to decrease memory footprint.
+/// In general it can be replaced with Map<string_view, size_t>.
+// TODO: Consider using absl::flat_hash_set instead.
```
Yuhta, line 33: nit: Ensure `///` for public comment

Yuhta (approved):

PR #15918, opened 2026-01-07
docs: Update Presto functions coverage map

### LINKED ISSUE #15917: Docs show green with no hyperlink when documentation is present
### Bug description

[Expected behavior]
Green colored links on documentation should have a corresponding link when the doc page is also present.
 
[actual behavior]
Green colored links with documentation do not have the hyperlink present.

### System information

Velox System Info v0.0.2
Commit: 8a7549594043272208b8ca3df4e04756f59b61f9
CMake Version: 3.31.6
System: Darwin-24.6.0
Arch: arm64
C++ Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
C++ Compiler Version: 16.0.0.16000026
C Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
C Compiler Version: 16.0.0.16000026
CMake Prefix Path: /Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr;/opt/homebrew;/usr/local;/usr;/;/opt/homebrew;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local
  

### Relevant logs

```bash
https://facebookincubator.github.io/velox/functions/presto/coverage.html
hovering over something like st_union that has docs present in https://facebookincubator.github.io/velox/functions/presto/geospatial.html#ST_Union does not show up.
```

jkhaliqi:
Closes #15917 
Removed duplicate function filter() being shown in the coverage map.
Also updated functions that appeared green on the coverage map but were missing links, even though the documentation with links already existed.

### DIFF
--- a/velox/docs/functions/presto/array.rst
+++ b/velox/docs/functions/presto/array.rst
@@ -2,7 +2,7 @@
 Array Functions
 =============================
 
-.. function:: all_match(array(T), function(T, boolean)) → boolean
+.. function:: all_match(array(T), function(T, boolean)) -> boolean
 
     Returns whether all elements of an array match the given predicate.
 
@@ -11,12 +11,21 @@ Array Functions
         Returns NULL if the predicate function returns NULL for one or more elements and true for all other elements.
         Throws an exception if the predicate fails for one or more elements and returns true or NULL for the rest.
 
-.. function:: any_match(array(T), function(T, boolean)) → boolean
+.. function:: any_match(array(T), function(T, boolean)) -> boolean
 
     Returns whether at least one element of an array matches the given predicate.
 
         Returns true if one or more elements match the predicate;
         Returns false if none of the elements matches (a special case is when the array is empty);
+        Returns NULL if the predicate function returns NULL for one or more elements and false for all other elements.
+        Throws an exception if the predicate fails for one or more elements and returns false or NULL for the rest.
+
+.. function:: none_match(array(T), function(T, boolean)) -> boolean
+
+    Returns whether no elements of an array match the given predicate.
+
+        Returns true if none of the elements matches the predicate (a special case is when the array is empty);
+        Returns false if one or more elements match;
         Returns NULL if the predicate function returns NULL for one or more elements and false for all other elements.
         Throws an exception if the predicate fails for one or more elements and returns false or NULL for the rest.
 
@@ -140,6 +149,17 @@ Array Functions
 .. function:: array_normalize(array(E), E) -> array(E)
 
     Normalizes array ``x`` by dividing each element by the p-norm of the array. It is equivalent to ``TRANSFORM(array, v -> v / REDUCE(array, 0, (a, v) -> a + POW(ABS(v), p), a -> POW(a, 1 / p))``, but the reduce part is only executed once. Returns null if the array is null or there are null array elements. If ``p`` is 0, then the input array is returned. Only REAL and DOUBLE types are supported.
+
+.. function:: arrays_overlap(x, y) -> boolean
+
+    Tests if arrays ``x`` and ``y`` have any non-null elements in common.
+    Returns null if there are no non-null elements in common but either array contains null.
+    For REAL and DOUBLE, NANs (Not-a-Number) are considered equal.
+
+.. function:: array_union(x, y) -> array
+
+    Returns an array of the elements in the union of x and y, without duplicates.
+    For REAL and DOUBLE, NANs (Not-a-Number) are considered equal.
 
 .. function:: array_position(x, element) -> bigint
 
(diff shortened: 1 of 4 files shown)


jkhaliqi:
[html-docs.zip](https://github.com/user-attachments/files/24482197/html-docs.zip)

velox/docs/functions/presto/geospatial.rst
```diff
@@ -46,40 +46,40 @@ the coordinate order is (x, y). The details of both WKT and WKB can be found
 Geometry Constructors
 ---------------------
 
-.. function:: ST_GeometryFromText(wkt: varchar) -> geometry: Geometry
+.. function:: st_geometryfromtext(wkt: varchar) -> geometry: Geometry
```
czentgr, line 49: Why the changes? In the end they are documented as camel case because the names (mainly) come from the standard. Internally, they will be lower case but we should leave it matching to the Presto documentation aka as it was before.

velox/docs/functions/presto/array.rst
```diff
@@ -2,16 +2,16 @@
 Array Functions
 =============================
 
-.. function:: all_match(array(T), function(T, boolean)) → boolean
+.. function:: all_match(array(T), function(T, boolean)) -> boolean
```
czentgr, line 5: This change isn't in the provided html files. Still looks like it was using `→` and not `->` (`_build/html/functions/presto/array.html`).
I see other changes working in the provided html so not sure what's going on.

velox/docs/functions/presto/coverage.rst
```diff
@@ -456,85 +472,84 @@ Here is a list of all scalar, aggregate, and window functions from Presto, with
...
-    current_timezone                          :func:`ip_prefix_collapse`                :func:`quantile_at_value`                 :func:`st_linestring`                                                                   :func:`variance`
-    :func:`date`                              :func:`ip_prefix_subnets`                 :func:`quarter`                           :func:`st_multipoint`
-    :func:`date_add`                          :func:`ip_subnet_max`                     :func:`radians`                           :func:`st_numgeometries`
-    :func:`date_diff`                         :func:`ip_subnet_min`                     :func:`rand`                              :func:`st_numinteriorring`
+    :func:`abs`                               :func:`date_diff`                         :func:`ip_subnet_min`                     :func:`radians`                           :func:`st_multipoint`                         :func:`approx_distinct`                       :func:`cume_dist`
```
czentgr: I see that the first column must have a new entry so that date_diff shifts to the second column (previously at the bottom of the first column).
Which function was added? Or why is the shift necessary?

czentgr (commented):
The functions.rst in the main directory also has a list of the supported functions. It is very similar to the coverage list but only lists functions that are supported.
Can we check those too?

velox/docs/functions/presto/geospatial.rst
jkhaliqi: Change is needed since in order to get the hyperlink to show it needs to be an exact match. Velox by default lowercases all the functions so this would be `st_geometryfromtext` and Sphinx is case-insensitive for the :func: directive in some cases which I assume is these cases. We can also change
```
std::string toFuncLink(
    const std::string& name,
    const std::string& domain = "") {
  return fmt::format("{}:func:`{}`", domain, name);
}
```
To have the cases where it would change for ST functions so ST_GeomFromBinary would be 
```
std::string toFuncLink(
    const std::string& name,
    const std::string& domain = "") {
  // Map lowercase function names to their proper documentation capitalization
  static const std::unordered_map<std::string, std::string> functionNameMap = {
      // Geometry functions
      {"st_area", "ST_Area"},
      {"st_asbinary", "ST_AsBinary"},
      {"st_astext", "ST_AsText"},
      // ADD IN MORE THAT NEED TO CHANGE THE CASEING
    // Check if we have a mapped name, otherwise use the original
    auto it = functionNameMap.find(name);
    const std::string& displayName = (it != functionNameMap.end()) ? it->second : name;
  
    return fmt::format("{}:func:`{}`", domain, displayName);
  };
}
```
But without it the link will not show up.

velox/docs/functions/presto/array.rst
jkhaliqi: Right, formatting is funky. Seems like when the return type has a parenthesis it messes up the parenthesis in the parameters and changes the -> arrow type. for those cases still on the UI, but code seems consistent with correct amount of parenthesis and arrows using -> so hyperlink is present.

velox/docs/functions/presto/coverage.rst
jkhaliqi: So `filter()` showed up two times so i removed `filter` function from the `all_scalar_functions.txt` file hence removing it from showing up twice in the coverage map. Before this the `pinot_binary_decimal_to_double` function which is twice as big of a square then the other functions since it has a long name was present on line with `crc32` function line, but now it is present down one in line with `current_date` increasing that square size and hence moving the `date_diff` function around as well. I don't know why it went down instead of up, but confirmed no other functions were added or deleted besides that `filter` function which showed up twice before.

PR #15943, opened 2026-01-08
refactor(python): Simplify task iterator API

pedroerp:
Summary:
Refactoring and simplifying task iterator API to enable it to return a
custom object, not simply a py::iterator, so we can add more methods to it in
the future.

Differential Revision: D89949295

### DIFF
--- a/velox/python/runner/PyLocalRunner.cpp
+++ b/velox/python/runner/PyLocalRunner.cpp
@@ -54,16 +54,27 @@ exec::Split preparedSplit(
 
 namespace py = pybind11;
 
-PyVector PyTaskIterator::Iterator::operator*() const {
-  return PyVector{vector_, outputPool_};
+PyTaskIterator::PyTaskIterator(
+    std::shared_ptr<memory::MemoryPool> pool,
+    std::shared_ptr<exec::TaskCursor> cursor)
+    : outputPool_(std::move(pool)), cursor_(std::move(cursor)) {
+  if (outputPool_ == nullptr) {
+    throw std::runtime_error(
+        "Memory pool cannot be nullptr when constructing PyTaskIterator.");
+  }
+  if (cursor_ == nullptr) {
+    throw std::runtime_error(
+        "Cursor cannot be nullptr when constructing PyTaskIterator.");
+  }
 }
 
-void PyTaskIterator::Iterator::advance() {
-  if (cursor_ && cursor_->moveNext()) {
-    vector_ = cursor_->current();
-  } else {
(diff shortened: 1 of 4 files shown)


velox/python/runner/PyLocalRunner.cpp
```diff
@@ -54,16 +54,27 @@ exec::Split preparedSplit(
 
 namespace py = pybind11;
 
-PyVector PyTaskIterator::Iterator::operator*() const {
-  return PyVector{vector_, outputPool_};
+PyTaskIterator::PyTaskIterator(
```
kgpai, line 57: nit: This looks like it should be backwards compatible, but want to confirm, that it doesnt break existing python programs..

kgpai (approved):

velox/python/runner/PyLocalRunner.cpp
pedroerp: this should be ok; you cannot create this object from Python (which is the external API). Only runner.execute() creates iterators today.

ericcfu, line 57: Not so familiar with this code, is it possible for the constructor to throw exceptions? Unhandled exceptions within pybind during constructor/destructor may cause undefined behavior, is it worth adding special handling?

velox/python/runner/runner.cpp
```diff
@@ -35,9 +35,18 @@ PYBIND11_MODULE(runner, m) {
   static auto executor = std::make_shared<folly::CPUThreadPoolExecutor>(
       folly::hardware_concurrency());
 
-  // execute() returns an iterator to Vectors.
+  // TaskIterator iterates over Vectors.
   py::module::import("pyvelox.vector");
 
+  py::class_<velox::py::PyTaskIterator>(m, "TaskIterator")
```
ericcfu, line 43: Is it worth registering some common velox exceptions in this pybind? Or are we happy having them translated to RuntimeException? https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

pedroerp: Good question. The exception scenarios in this particular case should not happen unless there is something wrong with the library, so a generic runtime error should be fine, I think; but we can refactor later if needed.

velox/python/runner/PyLocalRunner.cpp
pedroerp: AFAIK, exceptions in the constructor should be ok, you just cannot throw in a destructor. Am I missing something?

PR #15944, opened 2026-01-08
refactor(trace): Move core API to separate library

pedroerp:
Summary:
Moving core trace API componenents to a separate library, and creating
a virtual abstraction to enable clients to specialize tracers.

Differential Revision: D89945731


velox/exec/fuzzer/MemoryArbitrationFuzzerRunner.cpp
```diff
@@ -28,6 +28,7 @@
 #include "velox/exec/fuzzer/MemoryArbitrationFuzzer.h"
 #include "velox/exec/fuzzer/PrestoQueryRunner.h"
 #include "velox/exec/fuzzer/ReferenceQueryRunner.h"
+#include "velox/serializers/PrestoSerializer.h"
```
duanmeng, line 31: Do we need this?

duanmeng (approved):
@pedroerp LGTM, one nit. Thanks for your refactor!

xiaoxmeng (approved):
@pedroerp thanks!

velox/exec/fuzzer/MemoryArbitrationFuzzerRunner.cpp
pedroerp: This is an actual missing include - the serializer class is used below. I think before this header was being pulled recursively by some other header dependencies, and the changes in this PR resulted in this not happening anymore.

I was also confused about that.

PR #15946, opened 2026-01-09
perf(string-view)!: Remove implicit conversion to std::string

pedroerp:
Summary:
Remove implicit conversions from velox::StringView to
std::string, to avoid silent unintended copies and allocations. You can still
to do the conversion, but now it needs to explicitly stated:

> StringView sv;
> auto str = std::string(sv);

Also removing implicit conversions to folly::StringPiece, and some conversions
would have been made ambiguous with the implicit conversion to
std::string_view.

The codebase has been cleaned up from relying on these conversion in
previous PRs. 

Reviewed By: Magoja

Differential Revision: D89950455

### DIFF
--- a/velox/type/StringView.h
+++ b/velox/type/StringView.h
@@ -32,17 +32,40 @@
 
 namespace facebook::velox {
 
-/// Variable length string or binary type for use in vectors. This has
-/// semantics similar to std::string_view or folly::StringPiece and
-/// exposes a subset of the interface. If the string is 12 characters
-/// or less, it is inlined and no reference is held. If it is longer, a
-/// reference to the string is held and the 4 first characters are
-/// cached in the StringView. This allows failing comparisons early and
-/// reduces the CPU cache working set when dealing with short strings.
+/// C++ container to store a variable length string (or binary type) optimized
+/// for use in Vectors.
 ///
-/// Adapted from TU Munich Umbra and CWI DuckDB.
+/// StringView provides a lightweight, non-owning view of string data with
+/// semantics similar to std::string or std::string_view. It implements the
+/// following space optimization:
 ///
-/// TODO: Extend the interface to parity with folly::StringPiece as needed.
+/// - Strings <= 12 bytes: Fully inlined, no heap allocation or pointer
+///     indirection.
+/// - Strings > 12 bytes: Stores pointer + caches first 4 bytes as prefix.
+///
+/// The prefix caching enables early comparison failures and reduces CPU cache
+/// pressure when working with large sets of strings.
+///
+/// Memory Layout (16 bytes total):
+/// - 4 bytes: size.
+/// - 4 bytes: prefix (first 4 chars, or part of inline data).
+/// - 8 bytes: either inline data continuation OR pointer to external data.
+///
+/// Key Characteristics:
+/// - Non-owning: Does not manage lifetime of referenced data.
+/// - Immutable: Provides const access to underlying data.
+/// - Efficient comparisons: Uses prefix and size for fast inequality checks.
+/// - Zero-overhead for small strings: Fully inlined with no allocations.
+///
+/// Conversions:
+/// - Implicit: from char*, std::string_view (const& only).
+/// - Explicit: from std::string, folly::StringPiece.
+///
+/// Safety Notes:
+/// - StringView does NOT own the data it references.
+/// - Callers must ensure referenced data outlives the StringView.
+/// - Conversion to views from temporaries is explicitly deleted to prevent
+///   dangling references to inlined data.
 struct StringView {
  public:
   using value_type = char;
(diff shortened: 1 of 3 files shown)


velox/type/StringView.h
```diff
@@ -186,37 +209,54 @@ struct StringView {
...
+  /// > StringView sv();
+  /// > folly::StringPiece sp = sv;
+  ///
+  /// Note that we also explicitly disable conversion from a temporary (rvalue)
+  /// because this could have resulted in a folly::StringPiece wrapping around a
+  /// temporary buffer, if the StringView was inlined:
+  ///
+  /// > folly::StringPiece sp(StringView()); // unsafe, won't compile.
+  ///
+  explicit operator folly::StringPiece() && = delete;
```
pedroerp, line 229: only explicit folly::StringPiece conversions.

velox/type/StringView.h
```diff
@@ -186,37 +209,54 @@ struct StringView {
...
-  operator std::string() const {
+  /// Similarly, conversion to std::string can only be done explicitly, since
+  /// this may result in an allocation and string copy.
+  ///
+  /// Note that in this case it is ok to enable conversion from a temporary
+  /// (rvalue) StringView since its contents will be copied into the
+  /// std::string.
+  ///
+  /// > std::string str(StringView()); // ok
+  explicit operator std::string() const {
```
pedroerp, line 242: same for std::string conversions.

velox/type/StringView.h
```diff
@@ -186,37 +209,54 @@ struct StringView {
...
-  std::string getString() const {
-    return *this;
-  }
-
-  std::string materialize() const {
-    return *this;
+  /// Conversion to std::string_view are allowed to be implicit, as long as they
+  /// are not from a temporary (rvalue) StringView, for the same reasons
+  /// described above.
+  /* implicit */ operator std::string_view() && = delete;
```
pedroerp, line 249: std::string_view conversions are now implicitly done.

velox/type/StringView.h
```diff
@@ -186,37 +209,54 @@ struct StringView {
...
+  /* implicit */ operator std::string_view() const& {
+    return std::string_view(data(), size());
   }
 
-  operator folly::dynamic() && = delete;
-  operator folly::dynamic() const& {
+  // TODO: Should make this explicit-only.
```
pedroerp, line 254: this is also sketchy; but we should clean up on a different PR.

velox/type/StringView.h
```diff
@@ -231,13 +271,28 @@ struct StringView {
...
+    return std::string(*this);
+  }
+
+  std::string getString() const {
+    return std::string(*this);
+  }
+
+  std::string materialize() const {
+    return std::string(*this);
+  }
```
MBkkt, line 285: I don't think it's something really bad, but do we need 3 aliases to do same?

Maybe add `TODO` about make it single alias?

velox/type/StringView.h
MBkkt, line 229: This actually wrong, should be `const &&` instead of `&&`
See example https://godbolt.org/z/KErT4foYh

velox/type/StringView.h
MBkkt, line 249: Same here, should be `const &&` instead of `&&`
See example https://godbolt.org/z/KErT4foYh

MBkkt (approved):
@pedroerp This PR is good, but before your PR in this code already was a compile time bug.

velox/type/StringView.h
pedroerp: Yeah, I didn't like it either, but it would a good amount of clean up to remove them. I'll add the TODO for now.

velox/type/StringView.h
pedroerp: Good catch. I'll add, but I think we need to remove both && and const&&, since the first would also have a weird semantic. For example:

```
StringView sv;
std::string_view sv2 = std::move(sv);
```

this may not be strictly unsafe, but it's weird that sv2 would point to a buffer stored inside an object that was moved-out.

MBkkt: As I mentioned in the example link, if you define `const &&` and `const &` it's just implicit way to also delete `non-const &&` and define `non-const &`, but you can prefer to be explicit about this

pedroerp: @MBkkt turns out making this change ended up breaking a bunch of other callsites. I merged this as-is so it fixes the std::string conversion problem, and will create an Issue for us to fix this on a separate PR.

Thank you for flagging this.

MBkkt: @pedroerp Hi, ok, make sense, would you share link to the issue for me please?

pedroerp: Issue: https://github.com/facebookincubator/velox/issues/16039

PR #15951, opened 2026-01-09
docs: Add docs for murmur3_x64_128

### LINKED ISSUE #15933: Add velox docs for murmur3_x64_128
### Description

Adding docs to Velox, will enhance the velox documentation as well.
Presto docs found at : https://prestodb.io/docs/current/functions/binary.html#murmur3_x64_128-binary-varbinary

jkhaliqi:
closes #15933 with implementation as seen from presto docs: https://prestodb.io/docs/current/functions/binary.html#murmur3_x64_128-binary-varbinary

### DIFF
--- a/velox/docs/functions/presto/binary.rst
+++ b/velox/docs/functions/presto/binary.rst
@@ -105,6 +105,11 @@ Binary Functions
 
     Computes the md5 hash of ``binary``.
 
+.. function:: murmur3_x64_128(binary) -> varbinary
+
+    Computes a 128-bit hash of ``binary`` that is equivalent to the 128-bit
+    MurmurHash3 algorithm, often called MurmurHash3_x64_128 or Murmur3F.
+
 .. function:: rpad(binary, size, padbinary) -> varbinary
     :noindex:
 


velox/docs/functions/presto/binary.rst
```diff
@@ -105,6 +105,10 @@ Binary Functions
 
     Computes the md5 hash of ``binary``.
 
+.. function:: murmur3_x64_128(binary) -> varbinary
+
+    Computes a hash of ``binary`` that is equivalent to C++ MurmurHash3_x64_128 (Murmur3F) of the same ``binary``.
```
PingLiuPing: Would you please help to explain what is "C++ MurmurHash3_x64_128 (Murmur3F)"?

jkhaliqi: Updated docs to be more explicit.

velox/docs/functions/presto/binary.rst
```diff
@@ -105,6 +105,12 @@ Binary Functions
 
     Computes the md5 hash of ``binary``.
 
+.. function:: murmur3_x64_128(binary) -> varbinary
+
+    Computes a hash of ``binary`` that is equivalent to the 128-bit MurmurHash3
+    algorithm for x64 platforms, as defined by the original C++ reference
```
czentgr: Is this a platform specific implementation? We can also build arm64/aarch64 and so is this a problem there?

jkhaliqi: https://github.com/aappleby/smhasher/blob/0ff96f7835817a27d0487325b6c16033e2992eb5/src/MurmurHash3.cpp#L5
Looks like original code this is implemented from works on all platforms but optimized for respective platforms.
https://github.com/facebookincubator/velox/blob/0abf000477abd0fea59adde8f0ac4cd5c9a20249/velox/functions/prestosql/BinaryFunctions.h#L520-L529
Velox Code implementation also looks like it is not platform specific.

velox/docs/functions/presto/binary.rst
```diff
@@ -105,6 +105,12 @@ Binary Functions
 
     Computes the md5 hash of ``binary``.
 
+.. function:: murmur3_x64_128(binary) -> varbinary
+
+    Computes a 128-bit hash of ``binary`` that is equivalent to the 128-bit
+    MurmurHash3 algorithm, as defined by the original C++ reference
```
PingLiuPing: > as defined by the original C++ reference

Since MurmurHash3_x64_128 is a standardized algorithm, we may not need to call out the C++ implementation explicitly.
It might be clearer to state that the output is bit-compatible with MurmurHash3_x64_128 (Murmur3F).

PR #15953, opened 2026-01-09
fix: Use EvictingCacheMap for compiled regular expressions

wecharyu:
Velox throw exception when cached compiled regex number hit the `expression.max_compiled_regexes`, but the regex number varies if it's from the input table column. For example, in the following query there are many distinct `location_path` as regex, it's very easy to hit the limit.
```sql
select split(dir_path, location_path)[1] as partition_name from test_tbl
```

In this patch we want to use the `folly::EvictingCacheMap` to store the cached regex with 2 advantages:
1. the cached regex number is still in control by `expression.max_compiled_regexes`.
2. task will not fail when where are many regex.

### DIFF
--- a/velox/functions/lib/Re2Functions.cpp
+++ b/velox/functions/lib/Re2Functions.cpp
@@ -43,18 +43,13 @@ Expected<RE2*> ReCache::tryFindOrCompile(const StringView& pattern) {
     return reIt->second.get();
   }
 
-  if (cache_.size() >= maxCompiledRegexes_) {
-    return folly::makeUnexpected(
-        Status::UserError("Max number of regex reached"));
-  }
-
   auto re = std::make_unique<RE2>(toStringPiece(pattern), RE2::Quiet);
   if (!re->ok()) {
     return folly::makeUnexpected(
         Status::UserError("invalid regular expression:{}", re->error()));
   }
 
-  auto [it, inserted] = cache_.emplace(key, std::move(re));
+  auto [it, inserted] = cache_.insert(key, std::move(re));
   VELOX_CHECK(inserted);
 
   return it->second.get();
@@ -916,14 +911,14 @@ class LikeWithRe2 final : public exec::VectorFunction {
 };
 
 // This function is constructed when pattern or escape are not constants.
-// It allows up to 'expression.max_compiled_regexes' different regular
-// expressions to be compiled throughout the query lifetime per expression and
-// thread of execution, note that optimized regular expressions that are not
+// It maintains an LRU cache with maximum capacity of
+// 'expression.max_compiled_regexes' compiled regular expressions per expression
+// and thread of execution, note that optimized regular expressions that are not
 // compiled are not counted.
 class LikeGeneric final : public exec::VectorFunction {
  public:
   explicit LikeGeneric(int64_t maxCompiledRegexes)
-      : maxCompiledRegexes_(maxCompiledRegexes) {}
+      : compiledRegularExpressions_(maxCompiledRegexes) {}
 
   void apply(
       const SelectivityVector& rows,
@@ -1040,11 +1035,6 @@ class LikeGeneric final : public exec::VectorFunction {
       return reIt->second.get();
     }
 
-    VELOX_USER_CHECK_LT(
-        compiledRegularExpressions_.size(),
-        maxCompiledRegexes_,
-        "Max number of regex reached");
-
     bool validEscapeUsage;
     auto regex = likePatternToRe2(pattern, escapeChar, validEscapeUsage);
     VELOX_USER_CHECK(
@@ -1057,17 +1047,16 @@ class LikeGeneric final : public exec::VectorFunction {
     checkForBadPattern(*re);
 
     auto [it, inserted] =
-        compiledRegularExpressions_.emplace(key, std::move(re));
+        compiledRegularExpressions_.insert(key, std::move(re));
     VELOX_CHECK(inserted);
 
     return it->second.get();
   }
 
-  mutable folly::F14FastMap<
+  mutable folly::EvictingCacheMap<
       std::pair<std::string, std::optional<char>>,
       std::unique_ptr<RE2>>
       compiledRegularExpressions_;
-  int64_t maxCompiledRegexes_;
 };
 
 void re2ExtractAll(
--- a/velox/common/caching/SimpleLRUCache.h
+++ b/velox/common/caching/SimpleLRUCache.h
@@ -109,8 +109,8 @@ struct SimpleLRUCacheStats {
 /// A simple LRU cache that allows each element to occupy an arbitrary amount of
 /// space in the cache. Useful when the size of the cached elements can vary a
 /// lot; if they are all roughly the same size something that only tracks the
-/// number of elements in the cache like common/datastruct/LRUCacheMap.h may be
-/// better.
+/// number of elements in the cache like folly/container/EvictingCacheMap.h may
+/// be better.
 ///
 /// NOTE:
 /// 1. NOT Thread-Safe: All the public calls modify internal structures and
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -113,8 +113,8 @@ class QueryConfig {
   static constexpr const char* kExprMaxArraySizeInReduce =
       "expression.max_array_size_in_reduce";
 
-  /// Controls maximum number of compiled regular expression patterns per
-  /// function instance per thread of execution.
+  /// Controls the capacity of the LRU cache for compiled regular expression
+  /// patterns per function instance per thread of execution.
   static constexpr const char* kExprMaxCompiledRegexes =
       "expression.max_compiled_regexes";
 
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -281,7 +281,7 @@ Expression Evaluation Configuration
    * - expression.max_compiled_regexes
      - integer
      - 100
-     - Controls maximum number of compiled regular expression patterns per batch.
+     - Controls the capacity of the LRU cache for compiled regular expression patterns per batch.
    * - debug_disable_expression_with_peeling
      - bool
      - false
--- a/velox/docs/functions/spark/regexp.rst
+++ b/velox/docs/functions/spark/regexp.rst
@@ -26,11 +26,11 @@ See https://github.com/google/re2/wiki/Syntax for more information.
     Note: The wildcard '%' represents 0, 1 or multiple characters and the
     wildcard '_' represents exactly one character.
 
-    Note: Each function instance allow for a maximum of ``expression.max_compiled_regexes``
-    (default 100) regular expressions to be compiled per thread of execution. Not all patterns
-    require compilation of regular expressions. Patterns 'hello', 'hello%', '_hello__%',
-    '%hello', '%__hello_', '%hello%', where 'hello', 'velox'
-    contains only regular characters and '_' wildcards are evaluated without
+    Note: Each function instance maintains an LRU cache with a maximum capacity of
+    ``expression.max_compiled_regexes`` (default 100) for compiled regular expressions
+    per thread of execution. Not all patterns require compilation of regular expressions.
+    Patterns 'hello', 'hello%', '_hello__%', '%hello', '%__hello_', '%hello%', where
+    'hello', 'velox' contains only regular characters and '_' wildcards are evaluated without
     using regular expressions. Only those patterns that require the compilation of
     regular expressions are counted towards the limit. ::
 
--- a/velox/functions/lib/Re2Functions.h
+++ b/velox/functions/lib/Re2Functions.h
@@ -18,6 +18,7 @@
 #include <memory>
 #include <vector>
 
+#include <folly/container/EvictingCacheMap.h>
 #include <re2/re2.h>
 #include "velox/expression/VectorFunction.h"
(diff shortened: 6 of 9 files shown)


velox/functions/lib/Re2Functions.cpp
```diff
@@ -43,18 +43,13 @@ Expected<RE2*> ReCache::tryFindOrCompile(const StringView& pattern) {
     return reIt->second.get();
   }
 
-  if (cache_.size() >= maxCompiledRegexes_) {
-    return folly::makeUnexpected(
-        Status::UserError("Max number of regex reached"));
-  }
```
rui-mo, line 49: Could you please check whether Presto throws an exception in this case? If it does, we may need a separate implementation for Spark.

rui-mo (commented):
Thanks! This idea looks good to me.

velox/functions/lib/Re2Functions.cpp
wecharyu: Yes, Presto also use `detail::ReCache` and it will throw exception. Would it make sense to support an LRU cache for Presto too?

rui-mo: Do you mean that Presto (Java) throws an exception? If so, this difference between Spark and Presto would need to be handled in Velox to ensure compatibility with both.

wecharyu: Sorry for the misunderstanding. I tested a similar SQL in Presto (Java), and it does not throw an exception.

rui-mo: So this patch would also help align Velox’s behavior with Presto (Java), right? cc: @Yuhta

rui-mo (approved):

wecharyu:
Hi @rui-mo, can we merge this patch if no more concerns? Thanks~

rui-mo:
Hi @wecharyu, the sparksql changes look good to me. Since this PR also affects PrestoSQL, the changes might need to be confirmed by the Meta team.

rui-mo (commented):
@mbasmanova @Yuhta Would you please review the changes related to prestosql? Thanks!

velox/functions/lib/Re2Functions.cpp
```diff
@@ -1057,17 +1047,16 @@ class LikeGeneric final : public exec::VectorFunction {
...
     auto [it, inserted] =
-        compiledRegularExpressions_.emplace(key, std::move(re));
+        compiledRegularExpressions_.insert(key, std::move(re));
     VELOX_CHECK(inserted);
 
     return it->second.get();
   }
 
-  mutable folly::F14FastMap<
+  mutable folly::EvictingCacheMap<
```
Yuhta, line 1056: When the capacity is reached, there is high chance the query would thrash the cache and become intolerably slow.  It's more preferred to fail the query immediately in this case instead of let it running, make the user waiting and wasting computing power.  So a regular map the more proper in this case.  If you think there is value in using a cache, you can put it behind the a flag and user should explicitly enable it in order to take the risk.

wecharyu: > When the capacity is reached, there is high chance the query would thrash the cache and become intolerably slow.

The slowness should be mainly caused by compiling regex, and it's unavoidable.

Take `Like` function as an example, vanilla Spark will cache the pattern if it's constant, and compile each pattern otherwise:
https://github.com/apache/spark/blob/a2edb559299d6d49406972e53a2df7dc155cae47/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala#L57-L76

Velox has a similar logic, for constant pattern it also compile only once, but for variable patterns, it use a cache map, it could improve performance but may fail the task directly.
https://github.com/facebookincubator/velox/blob/f760d54f90dbf60a992b89622cdbccd917f533cd/velox/functions/lib/Re2Functions.cpp#L2273-L2276

> It's more preferred to fail the query immediately in this case instead of let it running, make the user waiting and wasting computing power.

I don’t think regex compilation is expensive enough to justify failing the query upfront. The overhead is relatively small and expected as part of normal execution.

More importantly, this is behavior the user explicitly requested. It’s similar to a user starting a large write job: even if it will take significant time and resources, we don’t reject it preemptively—we execute it as requested.

Yuhta: No compilation of regex is very expensive when it is needed on each row.  The correct solution in this case is asking user to rewrite the query to avoid dynamic regex.  If they insist on wasting computing power, we can add an option to disable the usage of this map (no cache & no limit, same as spark) and they can opt in using query config.

mbasmanova: In practice, we found that users never write such queries intentionally. It is often a mistake. Hence, I suggest to not even add a config option.

rui-mo: Hi @Yuhta @mbasmanova, thanks for your insights! We noticed that the query below is valid in Spark and is used by customers. While its performance may be poor, the user should bear that cost if they truly need to run it. Do you think we could move the fix into the SparkSQL folder to make it specific to Spark workloads? This way, we wouldn’t introduce any configuration or logic changes to the shared code.

`select split(dir_path, location_path)[1] as partition_name from test_tbl`

mbasmanova: @rui-mo Rui, can you help us understand what is a valid use case for such a query? Typically, users run queries in shared setups. This means that a single query should not take outsized share of resources. It is not up to individual users to decide on running inefficient expensive queries.

> select split(dir_path, location_path)[1] as partition_name from test_tbl

Is this the use case? Does this require regex at all? Perhaps, this should be implemented without regex.

rui-mo: Just to clarify, 'valid' means the query executes in Spark without errors.
> select split(dir_path, location_path)[1] as partition_name from test_tbl

The delimiter (the second argument) in split function is a regular expression, and relies on the ReCache.
https://github.com/facebookincubator/velox/blob/8853645932621aeadb7fffdd9112b881ce4225f8/velox/functions/sparksql/Split.h#L212

rui-mo: > Typically, users run queries in shared setups. This means that a single query should not take outsized share of resources. It is not up to individual users to decide on running inefficient expensive queries.

That makes sense for a shared setup. @wecharyu Could you provide more details about your setup and how you’re running this query?

mbasmanova: > The delimiter (the second argument) in split function is a regular expression

Sure, but is it a constant re or dynamic? The latter doesn't make a lot of sense to me.

rui-mo: That might be where the issue comes from. Spark allows dynamic delimiter columns, see my test below.

```
scala> val df = Seq(
     |   ("a,b,c", ","),
     |   ("1|2|3", "\\|"),
     |   ("x:y:z", ":")
     | ).toDF("text", "delimiter")
scala> df.createOrReplaceTempView("tmp")
scala> val df2 = spark.sql("select split(text, delimiter) from tmp")
scala> df2.show()
+--------------------------+
|split(text, delimiter, -1)|
+--------------------------+
|                 [a, b, c]|
|                 [1, 2, 3]|
|                 [x, y, z]|
+--------------------------+
```

mbasmanova: I understand that this is allowed. The point I'm making this is an easy way to shot oneself in the foot. As we found out using this API with lots of different values for delimiter is always a user error.

wecharyu: > select split(dir_path, location_path)[1] as partition_name from test_tbl

Hi @rui-mo @mbasmanova, the above query is from our user's production job. This table records the hive table's partition metadata, the column is:
`dir_path`: the partition path
`location_path`:  the table path
Usually the hive partition path format is `${table path}/${partition name}`,  so user try to use `split(dir_path, location_path)` to get the partition name, and it's truly dynamic regex pattern.

> Could you provide more details about your setup and how you’re running this query?

We limit the acquired maximum resources of a Spark job and will kill the job that's not finished after 2 days, to avoid a job consuming too many resources.

mbasmanova: @wecharyu This is a very common use case. It should not use regexes though. location_path is not a regex and should not be interpreted as such. Simple string operation would work faster and more reliably.

CC: @amitkdutta

wecharyu: Yes, it could be improved. However, when migrating user tasks to Gluten, users are usually unwilling to modify their existing queries. Therefore, we try to preserve behavior that is as close as possible to vanilla Spark.

mbasmanova: @wecharyu Thank you for the detailed context. I appreciate the effort to support your users during migration — it's a real challenge when users are unwilling to modify existing queries.

That said, I'd like to highlight that this particular query has a correctness issue that goes beyond performance. `split(dir_path, location_path)` treats `location_path` as a **regular expression**, not a literal string. This means:

- If `location_path` contains `.` (e.g., `/data/warehouse/db.name/table`), the `.` matches **any character**, not a literal dot.
- If it contains `(`, `)`, `[`, `+`, `*`, `?`, or `$`, the regex may fail or produce wrong results.

The query happens to work for simple paths, but it is silently incorrect for paths that contain regex metacharacters. Switching to an evicting cache would make the query run without errors, but wouldn't fix the wrong results — and would make them harder to discover.

For this specific use case — extracting partition name from `${table_path}/${partition_name}` — a correct and faster alternative is:

```sql
substr(dir_path, length(location_path) + 2) as partition_name
```

I understand that asking users to rewrite queries is friction, especially during migration. A good middle ground might be to optimize the `split` function itself to detect when the delimiter is a plain string (no regex metacharacters) and use a literal string split instead of compiling a regex. This is similar to what the LIKE function already does — it analyzes patterns and uses fast paths (prefix match, suffix match, substring search) whenever possible, falling back to regex only when truly needed. This way, queries like `split(dir_path, location_path)` would just work — fast and correct — without users needing to change anything and without removing the safety net for cases that genuinely require regex.

CC: @rui-mo @amitkdutta @Yuhta

rui-mo: @mbasmanova Thank you for your valuable insights and for pointing out the potential error results in addition to the performance concerns. They are very helpful.

> A good middle ground might be to optimize the split function itself to detect when the delimiter is a plain string (no regex metacharacters) and use a literal string split instead of compiling a regex.

We cannot distinguish whether a string is intended as a regex pattern or a literal string by examining its content alone, because the same string can have different meanings depending on user intent. For example, "hello.world" could mean either the literal text "hello.world" or a regex pattern matching "hello" followed by any character and then "world". Since Spark’s [StringSplit](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/regexpExpressions.scala#L596) always treats the delimiter as a regex, Velox needs to do the same to ensure consistent behavior. Given the risk of masking errors and potential performance regressions, users may have to adjust their queries. cc: @wecharyu

mbasmanova: @rui-mo 

> We cannot distinguish whether a string is intended as a regex pattern or a literal string by examining its content alone

This is indeed the case. What I'm suggesting is to check if the string doesn't have any regex special chars and if not use fast path that doesn't involve compiling regex. E.g. 'hello' can go through the fast path, 'hello.world' cannot.

rui-mo: Thanks for the clarification. The Spark split function has a fast path, though it seems @wecharyu’s mentioned case doesn’t use it.

https://github.com/facebookincubator/velox/blob/d0724bd7a8b455dc789c56d3dbb392d5df72a474/velox/functions/sparksql/Split.h#L140-L146

mbasmanova: @rui I didn't know about this fast path. Would it be possible to document this in https://facebookincubator.github.io/velox/functions/spark/string.html#split (similar to https://facebookincubator.github.io/velox/functions/presto/regexp.html#like )

Can we review that fast path to see if it can be extended to cover more cases?

wecharyu: We also checked the Vanilla Spark's `StringSplit` code, it use jdk `String.split` which only supports one-char or two-char fastpath, we consider to expand the fast path to support pure string(no regex) split case (but it's another topic because other regex usages may also hit this limits, so we want first solve the number limit issue): https://github.com/openjdk/jdk/blob/3e9fc5d49e52d79bcd2bb75068ff7efb31f768fd/src/java.base/share/classes/java/lang/String.java#L3771-L3778

mbasmanova: @wecharyu Please, check out https://velox-lib.io/blog/

rui-mo: Indeed, we initially followed Java’s implementation and supported limited cases in the fast path. And it should be possible to expand the fast path to cover more scenarios where the delimiter is a plain string and not intended to be interpreted as a regular expression.

Based on the discussion above, I agree that having a large number of dynamic regular expressions is not a typical user scenario. Even the case mentioned by @wecharyu is not intended for regex matching, but rather behaves more like a plain string match. In such situation, expanding the fast path seems to be the most appropriate way to address the issue.

PR #15955, opened 2026-01-09
feat: Add BaseVector::createFromVariants convenience API

PingLiuPing:
Add a new helper function to create a vector from a list of Variant values.
Supports all primitive types and complex types (ARRAY, MAP, ROW).

### DIFF
--- a/velox/vector/BaseVector.h
+++ b/velox/vector/BaseVector.h
@@ -862,6 +862,16 @@ class BaseVector {
   /// Returns the value at specified index as a Variant.
   Variant variantAt(vector_size_t index) const;
 
+  /// Returns values for all rows as Variants.
+  std::vector<Variant> toVariants() const;
+
+  /// Creates a vector from a list of Variant values. The result is a flat
+  /// vector with one element per value.
+  static VectorPtr createFromVariants(
+      const TypePtr& type,
+      const std::vector<Variant>& values,
+      velox::memory::MemoryPool* pool);
+
   /// Returns string representation of the value in the specified row.
   virtual std::string toString(vector_size_t index) const;
 
--- a/velox/vector/tests/VariantToVectorTest.cpp
+++ b/velox/vector/tests/VariantToVectorTest.cpp
@@ -36,7 +36,7 @@ class VariantToVectorTest : public testing::Test, public test::VectorTestBase {
         type, Variant::null(type->kind()), 1, pool());
 
     EXPECT_TRUE(vector->isConstantEncoding());
-    EXPECT_EQ(vector->type()->toString(), type->toString());
+    VELOX_EXPECT_EQ_TYPES(vector->type(), type);
     EXPECT_TRUE(vector->isNullAt(0));
 
     auto copy = vector->variantAt(0);
@@ -51,7 +51,7 @@ class VariantToVectorTest : public testing::Test, public test::VectorTestBase {
     auto vector = BaseVector::createConstant(type, value, 1, pool());
 
     EXPECT_TRUE(vector->isConstantEncoding());
-    EXPECT_EQ(vector->type()->toString(), type->toString());
+    VELOX_EXPECT_EQ_TYPES(vector->type(), type);
     EXPECT_TRUE(expected->equalValueAt(vector.get(), 0, 0))
         << "Expected: " << expected->toString(0)
         << ", but got: " << vector->toString(0);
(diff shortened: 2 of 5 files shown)


velox/vector/BaseVector.h
```diff
@@ -604,6 +604,13 @@ class BaseVector {
       vector_size_t size,
       velox::memory::MemoryPool* pool);
 
+  /// Creates a vector from a list of Variant values.
+  /// Supports all primitive types and complex types (ARRAY, MAP, ROW).
+  static VectorPtr createFromConstants(
```
mbasmanova: > createFromConstants

Let's rename to createFromVariants, add toVariants getter and move next to variantAt. Let's clarify that the result is a flat vector.

```
Variant variantAt(vector_size_t index) const;

std::vector<Variant> toVariants() const;

static VectorPtr createFromVariants(...);
```

> Supports all primitive types and complex types (ARRAY, MAP, ROW).

This is redundant since it effectively means that the API supports all types.

velox/vector/tests/VariantToVectorTest.cpp
```diff
@@ -61,6 +61,28 @@ class VariantToVectorTest : public testing::Test, public test::VectorTestBase {
...
   }
+
+  void testFromConstants(
+      const TypePtr& type,
+      const std::vector<Variant>& values,
+      const VectorPtr& expected) {
+    auto vector = BaseVector::createFromConstants(type, values, pool());
+
+    EXPECT_FALSE(vector->isConstantEncoding());
+    EXPECT_EQ(vector->type()->toString(), type->toString());
```
mbasmanova: use VELOX_EXPECT_EQ_TYPES

velox/vector/tests/VariantToVectorTest.cpp
```diff
@@ -61,6 +61,28 @@ class VariantToVectorTest : public testing::Test, public test::VectorTestBase {
     EXPECT_TRUE(copy.isTypeCompatible(type));
     EXPECT_EQ(copy, value);
   }
+
+  void testFromConstants(
+      const TypePtr& type,
+      const std::vector<Variant>& values,
+      const VectorPtr& expected) {
```
mbasmanova: > const VectorPtr& expected

Seems redundant to provide 'expected' as we can simply verify that values[i] == vector->variantAt(i).

velox/vector/tests/VariantToVectorTest.cpp
```diff
@@ -217,5 +239,63 @@ TEST_F(VariantToVectorTest, opaque) {
...
+  testFromConstants(
+      ROW({"a", "b"}, {INTEGER(), DOUBLE()}),
+      {Variant::row({1, 1.0}), Variant::row({2, 2.0})},
+      makeRowVector({
+          makeFlatVector<int32_t>({1, 2}),
+          makeFlatVector<double>({1.0, 2.0}),
+      }));
+}
+
+TEST_F(VariantToVectorTest, createFromConstantsEmpty) {
```
mbasmanova: Perhaps, add some negative tests where variants do not have consistent types or do not match the 'type'.

velox/vector/BaseVector.h
PingLiuPing: Thanks.

velox/vector/tests/VariantToVectorTest.cpp
PingLiuPing: Thanks.

velox/vector/tests/VariantToVectorTest.cpp
PingLiuPing: Thanks, you are right.

velox/vector/tests/VariantToVectorTest.cpp
PingLiuPing: Thanks, added a case to test type mismatch scenario.

velox/vector/tests/VariantToVectorTest.cpp
```diff
@@ -217,5 +232,71 @@ TEST_F(VariantToVectorTest, opaque) {
...
+
+TEST_F(VariantToVectorTest, createFromVariantsMismatchedType) {
+  {
+    std::vector<Variant> values = {
+        Variant::create<int32_t>(1),
+        Variant::create<int32_t>(2),
+    };
+    EXPECT_THROW(
+        BaseVector::createFromVariants(VARCHAR(), values, pool()),
+        std::invalid_argument);
```
mbasmanova: Can we make sure to throw VeloxException? and use VELOX_ASSERT_THROW to verify the message?

PingLiuPing: Yes, this require some additional change to `Variant::throwCheckIsKindError`. And I also updated`Variant::throwCheckPtrError`. @mbasmanova

mbasmanova (approved):
Thanks.

kevinwilfong:
@PingLiuPing it looks like VariantOpaqueTest.opaque is expecting the exception to be of a certain type and is now failing, could you take a look?

PingLiuPing:
> @PingLiuPing it looks like VariantOpaqueTest.opaque is expecting the exception to be of a certain type and is now failing, could you take a look?

@kevinwilfong  Thank you, will check and update.

PingLiuPing:
@kevinwilfong  Fixed.

kevinwilfong:
@PingLiuPing could I get you to rebase this one as well on top of https://github.com/facebookincubator/velox/pull/15981 (again, thank you so much for that fix!)

PR #15961, opened 2026-01-09
refactor(serde): Initialize CompactRow member variables with default values

Magoja:
Summary:
Prevents potential undefined behavior from uninitialized member variables by adding value-initialization to rowNullBytes_ and valueBytes_.

Also without this, sanitizer keep triggering alerts. Let's silence them with this.

Differential Revision: D90397061

### DIFF
--- a/velox/row/CompactRow.h
+++ b/velox/row/CompactRow.h
@@ -153,9 +153,9 @@ class CompactRow {
   bool supportsBulkCopy_{false};
 
   // ROW type only. Number of bytes used by null flags.
-  size_t rowNullBytes_;
+  size_t rowNullBytes_{0};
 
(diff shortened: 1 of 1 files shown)


velox/row/CompactRow.h
```diff
@@ -153,9 +153,9 @@ class CompactRow {
   bool supportsBulkCopy_{false};
 
   // ROW type only. Number of bytes used by null flags.
-  size_t rowNullBytes_;
+  size_t rowNullBytes_{};
 
   // Fixed-width types only. Number of bytes used for a single value.
-  size_t valueBytes_;
+  size_t valueBytes_{};
```
pedroerp: @Magoja should we add an explicit "0" zero, for clarity?

PR #15966, opened 2026-01-10
fix(build): Arrow bundle might fail with missing boost

czentgr:
The error shows as CMake error:

```
  lib/cpp/CMakeLists.txt:22 (REQUIRE_BOOST_HEADERS)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:290 (message):
  Could NOT find Boost (missing: Boost_INCLUDE_DIR) (Required is at least
  version "1.56")
  /opt/homebrew/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:654 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/share/cmake/Modules/FindBoost.cmake:2438 (find_package_handle_standard_args)
  build/cmake/BoostMacros.cmake:23 (find_package)
  lib/cpp/CMakeLists.txt:22 (REQUIRE_BOOST_HEADERS)
```

On macOS thrift is not installed into the system and this triggers an arrow bundling which can run into the missing boost headers/libraries.

### DIFF
--- a/CMake/resolve_dependency_modules/arrow/CMakeLists.txt
+++ b/CMake/resolve_dependency_modules/arrow/CMakeLists.txt
@@ -20,6 +20,19 @@ if(VELOX_ENABLE_ARROW)
     set(THRIFT_SOURCE "BUNDLED")
   endif()
 
+  # Avoid issues in finding the boost headers and libraries
+  # by setting the BOOST_ROOT to the install prefix.
+  # The same logic is used in the setup script-common.sh to install boost.
+  if(NOT DEFINED ENV{BOOST_ROOT} OR "$ENV{BOOST_ROOT}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


CMake/resolve_dependency_modules/arrow/CMakeLists.txt
```diff
@@ -20,6 +20,15 @@ if(VELOX_ENABLE_ARROW)
     set(THRIFT_SOURCE "BUNDLED")
   endif()
 
+  # Avoid issues in finding the boost headers and libraries
+  # by setting the BOOST_ROOT to the install prefix.
+  # The same logic is used in the setup script-common.sh to install boost.
+  if("$ENV{INSTALL_PREFIX}" STREQUAL "")
```
majetideepak: Let's skip if BOOST_ROOT is already defined.

PR #15967, opened 2026-01-10
feat(build): Update FBOS to v2026.01.05.00

czentgr:


czentgr:
Requires https://github.com/facebookincubator/velox/pull/15897 because co-routines leak into fbthrift.
```
#19 2233.0 /build/deps-download/fbthrift/./thrift/lib/cpp2/async/BiDiStream.h:71:34: error: ‘AsyncGenerator’ in namespace ‘folly::coro’ does not name a template type
#19 2233.0    71 |   using InputType = folly::coro::AsyncGenerator<InputElement&&>;
```

czentgr:
Investigating the SEGV. It occurs ONLY in release builds and NOT debug builds. It doesn't occur if only the remote functions are build but all CMake flags as they are used in the CI.

I built with having the frame pointer preserved which gets the function information at least.

Stack frame
```
(gdb) info stack
#0  0x00007fc5fd2f2614 in apache::thrift::TLibraryException::TLibraryException(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /velox/_build/release/lib/libvelox.so
#1  0x00007fc5fd2f27f6 in apache::thrift::transport::TTransportException::TTransportException(apache::thrift::transport::TTransportException::TTransportExceptionType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) () from /velox/_build/release/lib/libvelox.so
#2  0x00007fc5fd339a1d in apache::thrift::RocketClientChannel::closeNow() () from /velox/_build/release/lib/libvelox.so
#3  0x00007fc5fd337737 in apache::thrift::RocketClientChannel::~RocketClientChannel() ()
   from /velox/_build/release/lib/libvelox.so
#4  0x00007fc5fd3377d2 in apache::thrift::RocketClientChannel::~RocketClientChannel() ()
   from /velox/_build/release/lib/libvelox.so
#5  0x00007fc5fd2f2049 in folly::DelayedDestruction::Destructor::operator()(folly::DelayedDestruction*) const ()
   from /velox/_build/release/lib/libvelox.so
#6  0x00007fc5fd2fb2e2 in std::unique_ptr<apache::thrift::RocketClientChannel, folly::DelayedDestruction::Destructor>::~unique_ptr() () from /velox/_build/release/lib/libvelox.so
#7  0x00007fc5fd2ec120 in apache::thrift::HeaderClientChannel::RocketUpgradeChannel::~RocketUpgradeChannel() ()
   from /velox/_build/release/lib/libvelox.so
#8  0x00007fc5fd2ec18e in apache::thrift::HeaderClientChannel::RocketUpgradeChannel::~RocketUpgradeChannel() ()
   from /velox/_build/release/lib/libvelox.so
#9  0x00007fc5fc64d04a in std::_Sp_counted_deleter<apache::thrift::ClientChannel*, folly::DelayedDestruction::Destructor, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() () from /velox/_build/release/lib/libvelox.so
#10 0x00007fc5f6778b3a in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x1cb26e0)
    at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/shared_ptr_base.h:347
#11 std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x1cb26e0)
    at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/shared_ptr_base.h:317
#12 0x00007fc5fd265656 in apache::thrift::Client<facebook::velox::functions::remote::RemoteFunctionService>::~Client() ()
   from /velox/_build/release/lib/libvelox.so
#13 0x00007fc5fc650868 in std::_Sp_counted_deleter<facebook::velox::functions::(anonymous namespace)::RemoteThriftFunction*, std::default_delete<facebook::velox::functions::(anonymous namespace)::RemoteThriftFunction>, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() () from /velox/_build/release/lib/libvelox.so
#14 0x00007fc5f6c8ab21 in facebook::velox::exec::Expr::~Expr() () from /velox/_build/release/lib/libvelox.so
#15 0x00007fc5f6c8c08b in std::vector<std::shared_ptr<facebook::velox::exec::Expr>, std::allocator<std::shared_ptr<facebook::velox::exec::Expr> > >::~vector() () from /velox/_build/release/lib/libvelox.so
#16 0x0000000000a71a2f in facebook::velox::functions::test::FunctionBaseTest::evaluate(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<facebook::velox::RowVector> const&, std::optional<facebook::velox::SelectivityVector> const&) ()
#17 0x00000000009ebb72 in facebook::velox::functions::(anonymous namespace)::RemoteFunctionTest_simple_Test::TestBody()
    ()
#18 0x00007fc5eb25ce0c in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) [clone .constprop.0] () from /lib64/libgtest.so.1.11.0
#19 0x00007fc5eb23d826 in testing::Test::Run() () from /lib64/libgtest.so.1.11.0
#20 0x00007fc5eb23d9f0 in testing::TestInfo::Run() () from /lib64/libgtest.so.1.11.0
#21 0x00007fc5eb23daf9 in testing::TestSuite::Run() () from /lib64/libgtest.so.1.11.0
#22 0x00007fc5eb24cfc5 in testing::internal::UnitTestImpl::RunAllTests() () from /lib64/libgtest.so.1.11.0
#23 0x00007fc5eb24a7c8 in testing::UnitTest::Run() () from /lib64/libgtest.so.1.11.0
#24 0x00000000009d4240 in main ()
```
At frame 2 we call the `closeNow` function. This is here: https://github.com/facebook/fbthrift/blob/13ee1e2b291a59e9bec5b5026818d7c17a7c96df/thrift/lib/cpp2/async/RocketClientChannel.cpp#L1294

```
  rocket::RocketClient::closeNow(
      transport::TTransportException(
          transport::TTransportException::INTERRUPTED, "Client shutdown."));
```
Before it is called with the TTransportException it needs to be constructed. That is where the segv occurs.
The constructor has two constant inputs. So there should be no problem in calling it.

The signal occurs
```
(gdb) p $_siginfo
$1 = {si_signo = 11, si_errno = 0, si_code = 1, _sifields = {_pad = {8, 0 <repeats 27 times>}, _kill = {si_pid = 8,
      si_uid = 0}, _timer = {si_tid = 8, si_overrun = 0, si_sigval = {sival_int = 0, sival_ptr = 0x0}}, _rt = {
      si_pid = 8, si_uid = 0, si_sigval = {sival_int = 0, sival_ptr = 0x0}}, _sigchld = {si_pid = 8, si_uid = 0,
      si_status = 0, si_utime = 0, si_stime = 0}, _sigfault = {si_addr = 0x8, _addr_lsb = 0, _addr_bnd = {_lower = 0x0,
        _upper = 0x0}}, _sigpoll = {si_band = 8, si_fd = 0}}}
```
We can see the address is 0x8 that is faulty. We can see this is stored in register rsi and rax in the frame 2 when calling the function with two arguments. I suspect that 
```
(gdb) info registers
rax            0x8                 8
rbx            0x1cbfbc0           30145472
rcx            0x7ffe419fa090      140729999401104
rdx            0x7ffe419fa080      140729999401088
rsi            0x8                 8
rdi            0x7ffe419fa0d0      140729999401168
rbp            0x7ffe419fa150      0x7ffe419fa150
rsp            0x7ffe419fa0c0      0x7ffe419fa0c0
r8             0x1cb6b80           30108544
r9             0x0                 0
r10            0x7fc5f0c7fa08      140488124922376
r11            0x7fc5fd2f25fc      140488333010428
r12            0x1cb11f0           30085616
r13            0x0                 0
r14            0x1cb1190           30085520
r15            0x100000001         4294967297
rip            0x7fc5fd339a1d      0x7fc5fd339a1d <apache::thrift::RocketClientChannel::closeNow()+133>
eflags         0x10206             [ PF IF RF ]
cs             0x33                51
ss             0x2b                43
ds             0x0                 0
es             0x0                 0
fs             0x0                 0
gs             0x0                 0
k0             0x810000            8454144
k1             0xff                255
k2             0x3ff               1023
k3             0x0                 0
k4             0x0                 0
k5             0x0                 0
k6             0x1e0               480
k7             0x0                 0
```

So looking at the assembler I can see that the arguments are in rax which contains the constructed string and register esi (not on the list) is set to 5 which is the value for TTransportException::INTERRUPTED.
So it means that the string variable is bad with address of 0x8 when it is supposed to point to a std::string that is constructed from the "Client shutdown."

### Frame 2

image

The highlighted line is the next line after the call to the constructor. 

### Frame 1

image

Corresponding code is here: https://github.com/facebook/fbthrift/blob/13ee1e2b291a59e9bec5b5026818d7c17a7c96df/thrift/lib/cpp/transport/TTransportException.h#L82
We can see the call to `getDefaultMessage` in the stack and so we are done calling that and pass the message to the constructor here: https://github.com/facebook/fbthrift/blob/13ee1e2b291a59e9bec5b5026818d7c17a7c96df/thrift/lib/cpp/Thrift.h#L126

This takes a `const&` as message. The message itself is either construction with `fmt::format` or moved by `getDefaultMessage`. 
```
  TTransportException(TTransportExceptionType type, std::string message)
      : apache::thrift::TLibraryException(
            getDefaultMessage(type, std::move(message))),
        type_(type),
        errno_(0),
        options_(0) {}
```

```
std::string TTransportException::getDefaultMessage(
    TTransportExceptionType type, std::string&& message) {
  if (message.empty() &&
      static_cast<size_t>(type) >= TTransportExceptionTypeSize::value) {
    return fmt::format(
        "TTransportException: (Invalid exception type '{}')",
        folly::to_underlying(type));
  } else {
    return std::move(message);
  }
}
```
We know message was not empty (it was `Client shutdown.`). So the input is moved to the return as input to a const&.

And the input is now bad to the string constructor for the assignment `message_(message)`.
```
  explicit TLibraryException(const std::string& message) : message_(message) {}
```

Which then causes the issue in frame 0 that is calling the string constructor.

### Frame 0

image

The offending instruction is highlighted and it is %rax that is 0x8 causing the problem.

It is not quite clear to me what happens. The most recent change was adding fmt::format. But how does this work with debug? Is something optimized away?
The original code has not changed in a long time. All that changed recently was using fmt::format for the path that is not taken.
We could go to a folly version that didn't have the fmt::format changes. I have no clue for now what the root cause is.

@kgpai FYI

kgpai:
Thanks @czentgr

kgpai:
@czentgr  Looks like we are taking too much memory when linking  velox/exec/fuzzer/tests/presto_sql_test . 

Can you looking into optimizing this part of the build ? (Maybe break it up into smaller targets so that the build succeeds ?)

PR #15968, opened 2026-01-12
feat: Add map_update UDF

abhinavmuk04:
Summary:
Implemented MAP_UPDATE UDF for Velox following task T248293691.

## Function Signature
```
MAP_UPDATE(MAP<K,V>, ARRAY<K>, ARRAY<V>) -> MAP<K,V>
```

## Key Features
- Updates values in a map for specified keys using provided arrays
- If a key exists in the input map, its value is updated with the new value
- If a key doesn't exist, it is added to the map
- Keys and values arrays must have the same length
- Duplicate keys in the keys array throw an error
- Null keys are ignored, null values are preserved
- For REAL and DOUBLE, NaNs are considered equal

## Implementation Details
- Three optimized variants: primitive types, varchar, and generic types
- Follows the same pattern as map_append and map_subset
- Added to fuzzer skipFunctionsSOT (Velox-only function, not in Presto)

Differential Revision: D90475605

### DIFF
--- a/velox/functions/prestosql/tests/MapUpdateTest.cpp
+++ b/velox/functions/prestosql/tests/MapUpdateTest.cpp
@@ -1,0 +1,715 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/tests/MapUpdateTest.cpp
zacw7, line 1: CMakeLists.txt?

PR #15970, opened 2026-01-12
feat: Add L2_NORM UDF

abhinavmuk04:
Summary:
Introduces a MAGNITUDE function to calculate the Euclidean norm (L2 norm) of an array or map values, supporting common mathematical and ML use cases.

## Function Signatures
- `magnitude(array(T)) -> double` where T is numeric (int8, int16, int32, int64, float, double)
- `magnitude(map(K, V)) -> double` where V is numeric

## Behavior
- Calculates sqrt(sum(x^2)) for array elements or map values
- Returns 0.0 for empty arrays/maps
- Null elements/values are skipped (not included in calculation)
- Returns null if input is null

## Implementation Details
- Added `ArrayMagnitudeFunction` for array inputs
- Added `MapMagnitudeFunction` for map inputs with various key/value type combinations
- Registered as Velox-only function (excluded from Presto SOT fuzzer tests)

Differential Revision: D90475957

### DIFF
--- a/velox/functions/prestosql/L2Norm.h
+++ b/velox/functions/prestosql/L2Norm.h
@@ -1,0 +1,179 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


zacw7 (commented):
I think the name `magnitude` is a bit ambigious. How about `l2_norm`?

BigQuery: [LP_NORM](https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/bigqueryml-syntax-lp-norm)
Oracle: [VECTOR_NORM](https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/vector_norm.html)

velox/functions/prestosql/L2Norm.h
```diff
@@ -0,0 +1,138 @@
...
+      out = 0.0;
+      return true;
+    }
+
+    double sumOfSquares = 0.0;
+
+    for (const auto& item : inputArray) {
+      if (item.has_value()) {
+        auto value = static_cast<double>(item.value());
+        sumOfSquares += value * value;
```
zacw7: is there risk of over/underflow?

PR #15971, opened 2026-01-12
feat: Add dot_product UDF

abhinavmuk04:
Summary:
Add a DOT_PRODUCT function to compute the dot product between two arrays or maps, enabling efficient vector operations for ML and analytics workloads.

## Function Signatures
- `dot_product(array(T), array(T)) -> bigint/double`
- `dot_product(map(K, V), map(K, V)) -> bigint/double`

## Key Features
- **Array dot product**: Computes sum of element-wise products of corresponding elements
- **Map dot product**: Multiplies values with matching keys and sums the results
- **Equal length requirement**: Arrays must have the same length (throws error otherwise)
- **Null handling**: Null elements are treated as zero, null arguments return null
- **Type support**: int8, int16, int32, int64, float, double for arrays; integer/varchar keys with integer/double values for maps
- **Overflow protection**: Uses checked arithmetic for integer operations

## Behavior Examples
```sql
SELECT dot_product(ARRAY[1, 2, 3], ARRAY[4, 5, 6]); -- 32 (1*4 + 2*5 + 3*6)
SELECT dot_product(ARRAY[1.0, 2.0], ARRAY[3.0, 4.0]); -- 11.0
SELECT dot_product(ARRAY[1, NULL, 3], ARRAY[4, 5, 6]); -- 22 (nulls treated as 0)
SELECT dot_product(MAP(ARRAY[1, 2], ARRAY[10, 20]), MAP(ARRAY[1, 2], ARRAY[3, 4])); -- 110
```

Differential Revision: D90476184

### DIFF
--- a/velox/functions/prestosql/DotProduct.h
+++ b/velox/functions/prestosql/DotProduct.h
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cstdint>
+
+#include <folly/CPortability.h>
+#include <folly/container/F14Map.h>
+
+#include "velox/common/base/Exceptions.h"
+#include "velox/functions/Macros.h"
+#include "velox/functions/lib/CheckedArithmetic.h"
+#include "velox/type/SimpleFunctionApi.h"
+
+namespace facebook::velox::functions {
+
+/// Computes the dot product of two arrays.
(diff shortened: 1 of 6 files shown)


velox/functions/prestosql/registration/ArrayFunctionsRegistration.cpp
```diff
@@ -448,5 +449,64 @@ void registerArrayFunctions(const std::string& prefix) {
       ParameterBinder<MapL2NormFunction, int64_t, double>,
       double,
       Map<int64_t, double>>({prefix + "l2_norm"});
+
+  // Register dot_product for integer arrays.
+  // Note: float and double array versions are registered in
+  // MathematicalFunctionsRegistration.cpp with different behavior (NaN for
```
zacw7: any change in MathematicalFunctionsRegistration.cpp?

velox/functions/prestosql/DotProduct.h
```diff
@@ -0,0 +1,121 @@
...
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
+
+  template <typename TOutput>
+  FOLLY_ALWAYS_INLINE bool call(
+      TOutput& out,
+      const arg_type<Map<K, V>>& map1,
+      const arg_type<Map<K, V>>& map2) {
+    TOutput sum = 0;
+
+    for (const auto& [key1, val1] : map1) {
```
zacw7, line 103: shall we use a map for key-value lookup to avoid the nested loop?

zacw7 (approved):
LGTM. Thanks for adding the UDF.

PR #15972, opened 2026-01-12
feat: Add array_union_sum aggregation function

abhinavmuk04:
Summary:
This adds a new `array_union_sum` aggregation function that performs element-wise sum of arrays, similar to the existing `map_union_sum` function.

The function:
- Unions arrays of different lengths (extending to max length)
- Sums corresponding elements at each position
- Treats null elements as 0
- Supports numeric types: tinyint, smallint, integer, bigint, real, double
- Includes overflow checking for integer types

A variadic signature array_union_sum(c1, c2, ..., cN) is also supported, where each argument is a scalar at a fixed array position. This avoids per-row array construction overhead when the number of positions is known at query time.

Array form:

SELECT array_union_sum(x) FROM (
    VALUES
        (ARRAY[1, 2, 3]),
        (ARRAY[10, 5, 4, 1]),
        (ARRAY[9, NULL, 5, 4])
) AS t(x);
-- ARRAY[20, 7, 12, 5]

Variadic form:

SELECT array_union_sum(c0, c1, c2) FROM (
    VALUES
        (1, 2, 3),
        (10, 5, 4),
        (9, NULL, 5)
) AS t(c0, c1, c2);
-- ARRAY[20, 7, 12]

This is a Velox-only function (not available in Presto), so it has been added to the fuzzer SOT exclusion list.

Differential Revision: D90476644

### DIFF
--- a/velox/docs/functions/presto/aggregate.rst
+++ b/velox/docs/functions/presto/aggregate.rst
@@ -307,6 +307,22 @@ Map Aggregate Functions
     Returns the union of all the input maps summing the values of matching keys in all
     the maps. All null values in the original maps are coalesced to 0.
 
+Array Aggregate Functions
+-------------------------
+
+.. function:: array_union_sum(array(T)) -> array(T)
+
+    Returns the element-wise sum of all input arrays.
+    Arrays of different lengths are extended to the longest with zeros.
+    Null elements are treated as 0. Supported types for T are:
+    TINYINT, SMALLINT, INTEGER, BIGINT, REAL and DOUBLE.
+    Overflow is checked for integer types.
+
+    A variadic form ``array_union_sum(c1, c2, ..., cN)`` is also supported,
+    where each argument is a scalar at a fixed array position. This avoids
+    per-row array construction overhead when the number of positions is known
+    at query time.
+
 Approximate Aggregate Functions
 -------------------------------
 
--- a/velox/functions/prestosql/aggregates/ArrayUnionSumAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/ArrayUnionSumAggregate.cpp
@@ -1,0 +1,616 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/functions/prestosql/aggregates/ArrayUnionSumAggregate.h"
+#include "velox/exec/Aggregate.h"
+#include "velox/expression/FunctionSignature.h"
+#include "velox/functions/prestosql/aggregates/AggregateNames.h"
+#include "velox/vector/FlatVector.h"
+
+namespace facebook::velox::aggregate::prestosql {
+
+namespace {
+
+template <typename S>
+struct Accumulator {
+  using SumsVector = std::vector<S, AlignedStlAllocator<S, 16>>;
+  SumsVector sums;
+  bool initialized{false};
+
+  explicit Accumulator(const TypePtr& /*type*/, HashStringAllocator* allocator)
+      : sums{AlignedStlAllocator<S, 16>(allocator)} {}
+
+  size_t size() const {
+    return sums.size();
+  }
+
+  void addValues(
+      const ArrayVector* arrayVector,
+      const VectorPtr& arrayElements,
+      vector_size_t row) {
+    auto elements = arrayElements->template as<SimpleVector<S>>();
+    auto offset = arrayVector->offsetAt(row);
+    auto arraySize = arrayVector->sizeAt(row);
+
+    if (!initialized) {
+      // First array: just hold onto the values directly, no aggregation needed.
+      sums.resize(arraySize, S{0});
+      for (auto i = 0; i < arraySize; ++i) {
+        if (!elements->isNullAt(offset + i)) {
+          sums[i] = elements->valueAt(offset + i);
+        }
+      }
+      initialized = true;
+      return;
+    }
+
+    if (static_cast<size_t>(arraySize) > sums.size()) {
+      sums.resize(arraySize, S{0});
+    }
+
+    for (auto i = 0; i < arraySize; ++i) {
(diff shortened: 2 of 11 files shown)


kaikalur:
Also mention about variadic version in the PR description

velox/functions/prestosql/aggregates/ArrayUnionSumAggregate.cpp
```diff
@@ -0,0 +1,616 @@
...
+      }
+    }
+    return size;
+  }
+
+  DecodedVector decodedArrays_;
+};
+
+/// Variadic version: array_union_sum(c1, c2, ..., cN) where each ci is a
+/// scalar. Each row's arguments are treated as positions in a virtual array,
```
kaikalur: Arguments are not positions - argument position determines the index into the array for agg and final result

velox/docs/functions/presto/aggregate.rst
```diff
@@ -307,6 +307,22 @@ Map Aggregate Functions
     Returns the union of all the input maps summing the values of matching keys in all
     the maps. All null values in the original maps are coalesced to 0.
 
+Array Aggregate Functions
+-------------------------
+
+.. function:: array_union_sum(array(T)) -> array(T)
```
mbasmanova, line 313: I'm making assumptions. Please, correct me as needed.

array_union_sum is a shortcut for unnest + sum(x) group by(ordinality) + array_agg(sum order by ordinality)

variadic version is a shortcut for array[sum(c1), sum(c2)...]

This equivalence determines the semantics of array_union_sum (Null elements are treated as 0).

If so, let's explain this clearly and show the equivalent SQL query. 

Let's add tests that verify array_union_sum's results by comparing with equivalent SQLs. Add regular tests as well as fuzzer tests.

CC: @kagamiori @kgpai

kaikalur: But also, we have seen users actually wanting to sum arrays (like embeddings) position-wise in a number of situations so this is not just a shorthand but valid first-class agg just like map_union_sum. So not sure if we need the somewhat confusing unnest explanation in the docs - especially because wrapping it back to array needs another agg

velox/functions/prestosql/aggregates/ArrayUnionSumAggregate.cpp
```diff
@@ -0,0 +1,616 @@
...
+    rows.applyToSelected([&](vector_size_t row) {
+      auto* group = groups[row];
+      clearNull(group);
+      auto tracker = trackRowSize(group);
+      auto* acc = value<AccumulatorType>(group);
+
+      for (size_t argIdx = 0; argIdx < args.size(); ++argIdx) {
+        if (!decodedArgs_[argIdx].isNullAt(row)) {
+          auto val = decodedArgs_[argIdx].template valueAt<S>(row);
+          if (val != S{0}) {
```
kaikalur: actually might be good to special case 1 as well - using switch - as we want to use this to batch COUNT aggs as well using this as an optimization.

kaikalur (changes requested):

kagamiori:
Hi @abhinavmuk04, thank you for adding the new function. We generally recommend implementing aggregation functions using the [simple aggregation API](https://facebookincubator.github.io/velox/develop/aggregate-functions.html#simple-function-interface) that is less error-prone than the vector function API. Wondering if you could give it a try? 

https://github.com/facebookincubator/velox/pull/16475 adds the support for variadic arguments in the simple aggregation API.

kaikalur:
> Hi @abhinavmuk04, thank you for adding the new function. We generally recommend implementing aggregation functions using the [simple aggregation API](https://facebookincubator.github.io/velox/develop/aggregate-functions.html#simple-function-interface) that is less error-prone than the vector function API. Wondering if you could give it a try?
> 
> #16475 adds the support for variadic arguments in the simple aggregation API.

Is it as performant as the vector function?The whole idea for this is to support heavy aggregations

kaikalur:
> > Hi @abhinavmuk04, thank you for adding the new function. We generally recommend implementing aggregation functions using the [simple aggregation API](https://facebookincubator.github.io/velox/develop/aggregate-functions.html#simple-function-interface) that is less error-prone than the vector function API. Wondering if you could give it a try?
> > #16475 adds the support for variadic arguments in the simple aggregation API.
> 
> Is it as performant as the vector function?The whole idea for this is to support heavy aggregations - like some of our prod queries that have 2k+ SUM calls - want to try to use this api for that for optimization

mbasmanova:
> Is it as performant as the vector function?T

I expect it will be. Let's experiment to find out.

velox/docs/functions/presto/aggregate.rst
kaikalur: > But also, we have seen users actually wanting to sum arrays (like embeddings) position-wise in a number of situations so this is not just a shorthand but valid first-class agg just like map_union_sum. So not sure if we need the somewhat confusing unnest explanation in the docs - especially because wrapping it back to array needs another agg

By this logic we dont need any aggs lol - everything can be done using reduce-agg - especailly since this agg requires unnest and agg back I will not call it a short-hand. It's a first class function @mbasmanova

kaikalur:
And if velox team doesn't want it, we can simply add it to presto

kagamiori:
> > > Hi @abhinavmuk04, thank you for adding the new function. We generally recommend implementing aggregation functions using the [simple aggregation API](https://facebookincubator.github.io/velox/develop/aggregate-functions.html#simple-function-interface) that is less error-prone than the vector function API. Wondering if you could give it a try?
> > > #16475 adds the support for variadic arguments in the simple aggregation API.
> > 
> > 
> > Is it as performant as the vector function?The whole idea for this is to support heavy aggregations - like some of our prod queries that have 2k+ SUM calls - want to try to use this api for that for optimization

I have some numbers with some further optimizations to the simple aggregation API. I'll share it shortly.

kagamiori:
Hi @kaikalur and @mbasmanova, https://github.com/facebookincubator/velox/pull/16512 adds a benchmark of aggregation functions using variadic arguments, between the vector-function implementation and the simple-function implementation. (See the last commit in this PR since it includes base commits from the stack.)
```
[...]mpleVariadicSumAggregateBenchmark.cpp     relative  time/iter   iters/s
variadicSumAggregateDefaultNull                            16.31ms     61.32
simpleVariadicSumDefaultNull                    88.479%    18.43ms     54.25
variadicSumAggregateNonDefaultNull                         17.07ms     58.60
simpleVariadicSumNonDefaultNull                 101.27%    16.85ms     59.34
```

For functions of non-default-null behavior (e.g., the function added in this PR), the simple-function implementation is as good as the vector-function one. For functions of default-null behavior (i.e., functions that ignore an input row if the row contains any null), the simple-function implementation is currently 12--13% slower, due to indirections when accessing values through the variadic VectorReader. It would need further optimizations to close this gap, but this PR is not blocked by this.

For functions not using variadic arguments, the simple-function implementation is known to behave as good as vector-function implementation.
Screenshot 2026-02-24 at 5 05 31 PM

PR #15978, opened 2026-01-12
feat: Add TRANSFORM_WITH_INDEX UDF

abhinavmuk04:
Summary:
Add a new TRANSFORM_WITH_INDEX function that allows transformation of array elements with access to their index, enabling more flexible feature engineering.

The existing TRANSFORM function only passes the element to the lambda, making it difficult to use indices in array transform operations. This new function passes both the element and its 1-based index to the lambda:

```sql
TRANSFORM_WITH_INDEX(arr, (elem, index) -> ...)
```

## Function Signature
```
transform_with_index(array(T), function(T, bigint, U)) -> array(U)
```

## Examples
```sql
SELECT transform_with_index(ARRAY [5, 6, 7], (x, i) -> x * i); -- [5, 12, 21]
SELECT transform_with_index(ARRAY ['a', 'b', 'c'], (x, i) -> concat(x, cast(i as varchar))); -- ['a1', 'b2', 'c3']
SELECT transform_with_index(ARRAY [10, 20, 30], (x, i) -> i); -- [1, 2, 3]
```

## Implementation Details
- Uses 1-based indexing for Presto compatibility
- Follows the same pattern as the existing TRANSFORM function
- Added to fuzzer exclusion lists as this is a Velox-only function not available in Presto

Differential Revision: D90478316

### DIFF
--- a/velox/functions/prestosql/TransformWithIndex.cpp
+++ b/velox/functions/prestosql/TransformWithIndex.cpp
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/expression/Expr.h"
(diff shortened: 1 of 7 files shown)


velox/functions/prestosql/TransformWithIndex.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "velox/expression/VectorFunction.h"
+#include "velox/functions/lib/LambdaFunctionUtil.h"
+#include "velox/functions/lib/RowsTranslationUtil.h"
+#include "velox/vector/FunctionVector.h"
+
+namespace facebook::velox::functions {
+namespace {
+
+// See documentation at https://prestodb.io/docs/current/functions/array.html
+// transform_with_index(array(T), function(T, bigint, U)) -> array(U)
```
zacw7: what documentation does it refer to?

zacw7 (approved):
LG. Please take a look at the lint warnings before landing. Thanks!

PR #15980, opened 2026-01-12
fix: Check invalid escaped char for Spark get_json_object function

zhli1142015:
Jackson checks if there are invalid escape chars in json body, but simdjson 
deosn't. We add extrack escaped sequence check to align the behavior with Spark.

https://github.com/FasterXML/jackson-core/blob/8744bd42770c9e277d995ef00fb518940efef3ef/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java#L2648-#L2699 

```
scala> spark.conf.set("spark.gluten.enabled", "false")

scala> spark.sql("""select get_json_object('{"unionId":"xxxx", "nickname":"\\๑"}', '$.unionId');""").collect
val res3: Array[org.apache.spark.sql.Row] = Array([null])

scala> spark.conf.set("spark.gluten.enabled", "true")

scala> spark.sql("""select get_json_object('{"unionId":"xxxx", "nickname":"\\๑"}', '$.unionId');""").collect
val res5: Array[org.apache.spark.sql.Row] = Array([xxxx])
```

### DIFF
--- a/velox/functions/sparksql/GetJsonObject.h
+++ b/velox/functions/sparksql/GetJsonObject.h
@@ -16,9 +16,8 @@
 
 #pragma once
 
-#include <cstring>
-
 #include <folly/Likely.h>
+#include <cstring>
 
 #include "velox/core/QueryConfig.h"
 #include "velox/functions/Macros.h"
@@ -290,6 +289,8 @@ struct GetJsonObjectFunction {
   // Note: We only search for '\' which is ASCII (0x5C) and cannot appear
   // as a continuation byte in valid UTF-8 (continuation bytes are 0x80-0xBF).
   // So we can safely scan byte-by-byte regardless of encoding.
+  // See the valid escape sequences in Jackson's JSON parser:
+  // https://github.com/FasterXML/jackson-core/blob/jackson-core-2.19.2/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java#L2648
   FOLLY_ALWAYS_INLINE bool hasInvalidEscapedChar(
       const char* json,
       size_t size) {


velox/functions/sparksql/GetJsonObject.h
```diff
@@ -277,6 +281,55 @@ struct GetJsonObjectFunction {
...
+
+    while ((pos = static_cast<const char*>(memchr(
+                pos, '\\', static_cast<size_t>(end - pos)))) != nullptr) {
+      size_t remaining = static_cast<size_t>(end - pos);
+      if (FOLLY_UNLIKELY(remaining < 2)) {
+        return false; // Incomplete escape at end, let parser handle it.
+      }
+
+      switch (pos[1]) {
+        case '"':
```
jinchengchenghh, line 308: add `[[fallthrough]];`

velox/functions/sparksql/tests/GetJsonObjectTest.cpp
```diff
@@ -136,6 +136,13 @@ TEST_F(GetJsonObjectTest, nullResult) {
           R"([{"my": {"info": {"name": "Alice"quoted""}}}, {"other": ["v1", "v2"]}])",
           "$[0].my.info.name"),
       std::nullopt);
+
+  // Invalid escape sequence.
+  EXPECT_EQ(getJsonObject(R"({"hello": "\x"})", "$.hello"), std::nullopt);
```
jinchengchenghh, line 153: Do we need to test all the paths? pos[1] == xxx?

jinchengchenghh (approved):
Thanks!

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -16,6 +16,11 @@
 
 #pragma once
 
+#include <cstring>
+
```
rui-mo: nit: remove the empty line.

velox/functions/sparksql/GetJsonObject.h
```diff
@@ -277,6 +286,62 @@ struct GetJsonObjectFunction {
...
+        case 'f':
+          [[fallthrough]];
+        case 'n':
+          [[fallthrough]];
+        case 'r':
+          [[fallthrough]];
+        case 't':
+          pos += 2;
+          break;
+        case 'u':
```
rui-mo, line 325: Could you please add a reference for the rule used to validate escaped char?

zhli1142015: Added, thanks.

PR #15984, opened 2026-01-12
refactor(geo): Replace VarbinaryWriter with exec::StringWriter

jagill:
Summary:
Replace the custom VarbinaryWriter template class with the standard
velox::exec::StringWriter. The private write methods remain templated
to support both exec::StringWriter (for the main API) and std::string
(for temporary buffers in writeGeometryCollection).

This change simplifies the serialization interface by using Velox's
standard StringWriter instead of a custom abstraction.

Differential Revision: D90505499

### DIFF
--- a/velox/common/geospatial/tests/GeometrySerdeTest.cpp
+++ b/velox/common/geospatial/tests/GeometrySerdeTest.cpp
@@ -19,31 +19,40 @@
 #include <geos/io/WKTReader.h>
 #include <geos/io/WKTWriter.h>
 #include <gtest/gtest.h>
+#include "velox/common/memory/Memory.h"
+#include "velox/expression/StringWriter.h"
 #include "velox/type/StringView.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
 
 using namespace ::testing;
 
-using namespace facebook::velox::common::geospatial;
+namespace facebook::velox::common::geospatial {
+namespace {
 
-void assertRoundtrip(const std::string& wkt) {
-  geos::io::WKTReader reader;
-  geos::io::WKTWriter writer;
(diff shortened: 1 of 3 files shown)


velox/common/geospatial/tests/GeometrySerdeTest.cpp
```diff
@@ -19,31 +19,48 @@
...
 #include <geos/io/WKTWriter.h>
 #include <gtest/gtest.h>
+#include "velox/common/memory/Memory.h"
+#include "velox/expression/StringWriter.h"
 #include "velox/type/StringView.h"
+#include "velox/vector/tests/utils/VectorTestBase.h"
 
 using namespace ::testing;
 
+using namespace facebook::velox;
```
mbasmanova: A convention is to put the test in anonymous namespace within the namespace that contains code being tested.

```
namespace facebook::velox::common::geospatial {
namespace {

// test...
}
}

File: velox/common/geospatial/tests/GeometrySerdeTest.cpp [hunk 2]
```diff
@@ -19,31 +19,48 @@
...
+    exec::StringWriter stringWriter(vector.get(), 0);
+    GeometrySerializer::serialize(*geometry, stringWriter);
+    stringWriter.finalize();
+
+    StringView readBuffer = vector->valueAt(0);
+    auto deserialized = GeometryDeserializer::deserialize(readBuffer);
+
+    EXPECT_TRUE(geometry->equals(deserialized.get()))
+        << std::endl
+        << "Input:" << std::endl
```
mbasmanova: Use SCOPED_TRACE

mbasmanova (approved):
Thanks.

PR #15985, opened 2026-01-12
feat(geo): Make ST_Centroid consistent with ISO spec

jagill:
Summary:
The ISO specification has ST_Centroid returning NULL on
empty geometries.  ST_Centroid(Geometry) currently returns an empty
geometry for empty geometries.  This is the same as Java, but
inconsistent with the spec and also with ST_Centroid(SphericalGeography).

This change makes Velox compatible with the ISO spec but incompatible
with Presto Java.

Differential Revision: D90506624

### DIFF
--- a/velox/functions/prestosql/GeometryFunctions.h
+++ b/velox/functions/prestosql/GeometryFunctions.h
@@ -589,10 +589,15 @@ template <typename T>
 struct StCentroidFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
-  FOLLY_ALWAYS_INLINE Status
-  call(out_type<Geometry>& result, const arg_type<Geometry>& input) {
+  FOLLY_ALWAYS_INLINE bool call(
+      out_type<Geometry>& result,
+      const arg_type<Geometry>& input) {
     std::unique_ptr<geos::geom::Geometry> geosGeometry =
         common::geospatial::GeometryDeserializer::deserialize(input);
+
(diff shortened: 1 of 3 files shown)


velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -605,17 +610,17 @@ struct StCentroidFunction {
         "ST_Centroid");
 
     if (!validate.ok()) {
-      return validate;
+      VELOX_USER_FAIL(validate.message());
```
mbasmanova: VELOX_USER_CHECK(validate.ok(), validate.message())

velox/functions/prestosql/GeometryFunctions.h
```diff
@@ -589,11 +589,16 @@ template <typename T>
 struct StCentroidFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
-  FOLLY_ALWAYS_INLINE Status
-  call(out_type<Geometry>& result, const arg_type<Geometry>& input) {
+  FOLLY_ALWAYS_INLINE bool call(
```
mbasmanova, line 592: It is unfortunate we cannot use no-throwing APIs. Is there a GitHub issue to add support for that? If not, let's create one.

jagill: Do you mean non-throwing APIs for Geos? We can open an issue but I have high confidence it will not be implemented.

PR #15937, opened 2026-01-08
build(cudf): Update cudf dependency to 26.02

bdice:
Updates cudf to a prerelease of version 26.02. This is needed to get some fixes like https://github.com/rapidsai/cudf/pull/20834. We may also need https://github.com/rapidsai/cudf/pull/20982 -- I will see if CI passes first, and update again if needed.

PR #15992, opened 2026-01-12
build(cudf): Pin cuDF dependencies and add script for updates

bdice:
This PR pins cuDF's dependencies: rapids-cmake, rmm, and kvikio.

This is a follow-up to #15937, which updates the cuDF dependency to 26.02. It pulls a recent cuDF commit from `main`. However, cuDF's core RAPIDS dependencies (rapids-cmake, RMM, KvikIO) are currently unpinned, meaning that they will continue to pull the latest from `main` while cuDF remains pinned. We would like to ensure that builds are stable and reproducible, which means we should pin these core cuDF dependencies.

To make future updates simpler, I added a script `scripts/update-cudf-deps.sh`. To update to a recent branch, developers can run:

```bash
./scripts/update-cudf-deps.sh --branch main
./scripts/update-cudf-deps.sh --branch release/26.02
```

A pain point that cuDF developers have faced is testing Velox with cuDF pull requests in progress. To support that use case, developers can run:

```bash
./scripts/update-cudf-deps.sh --pr <pr-number>
```

To use a specific cuDF commit with compatible dependency versions, developers can run:

```bash
./scripts/update-cudf-deps.sh --commit <sha>
```

This automatically finds compatible rapids-cmake, rmm, and kvikio versions by selecting the most recent main branch commit before the specified cuDF commit date, ensuring all dependencies are temporally compatible.

### DIFF
--- a/velox/experimental/cudf/exec/CudfLocalPartition.cpp
+++ b/velox/experimental/cudf/exec/CudfLocalPartition.cpp
@@ -169,11 +169,16 @@ void CudfLocalPartition::addInput(RowVectorPtr input) {
       VELOX_FAIL("Unsupported partition function");
     }();
 
-    VELOX_CHECK(partitionOffsets.size() == numPartitions_);
+    // cuDF partitioning APIs return num_partitions + 1 offsets where:
+    // - offsets[i] is the starting row index for partition i
+    // - offsets[num_partitions] is the total row count
+    VELOX_CHECK(partitionOffsets.size() == numPartitions_ + 1);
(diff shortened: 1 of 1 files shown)


karthikeyann (approved):

scripts/update-cudf-deps.sh
bdice, line 1: This is optional -- if Velox maintainers have reservations about including this script, we don't have to include it here. We can move it somewhere else.

Its purpose is to pin a compatible cudf dependency tree (rapids-cmake, rmm, kvikio) based on the desired cuDF branch/PR/commit hash.

PR #15995, opened 2026-01-12
feat: Add SetDigest and KHLL to Velox cast documentation

natashasehgal:
Add SetDigest and KHLL to Velox cast documentation. These can be cast from/to VARBINARY its physical representation.

Differential Revision: D90530119


HeidiHan0000 (approved):
LGTM, thank you for adding these!

PR #15997, opened 2026-01-12
docs: Remove special functions

jkhaliqi:
When viewing the velox docs https://facebookincubator.github.io/velox/functions.html for the function list that is generated, lists special forms as well, that are implemented as function such as in, not, cast, between, but these do not have docs and are not present in presto so we can remove them from the function list in velox docs as well.
[html.zip](https://github.com/user-attachments/files/24575957/html.zip)


czentgr:
@pedroerp Please take a look. We are updating the documentation and fix oddities.

czentgr (approved):

allenshen13 (approved):

pedroerp (approved):
Looks good to me. Thanks @jkhaliqi

PR #16000, opened 2026-01-13
refactor(geo): Move GeometrySerializer implementations to cpp file

jagill:
Summary:
Move GeometrySerializer method implementations from the header file
to the cpp file. This reduces compilation dependencies and keeps
the header cleaner with only declarations.

Changes:
- Move all GeometrySerializer non-template method implementations to cpp
- Remove FOLLY_ALWAYS_INLINE from moved methods
- Remove template parameters from private methods (now only use exec::StringWriter)
- Add geos/geom/Polygon.h include to cpp for dynamic_cast
- Keep appendBytes as a template in the header (inline for performance)

Differential Revision: D90589273


mbasmanova (approved):
Thanks.

PR #16002, opened 2026-01-13
fix: Fix spurious compiler error due to deep inlining

kgpai:
Summary:
Ubuntu release build gives the following error due to GCC's spurious deep inlining optimization when inlining creation of StringViews : 
```
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:33: error: ‘void* __builtin___memcpy_chk(void*, const void*, long unsigned int, long unsigned int)’ writing 13 or more bytes into a region of size 12 overflows the destination [-Werror=stringop-overflow=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This fixes it by letting the compiler know that its not possible to have length > 12 in the inlining case. 
Fixes https://github.com/facebookincubator/velox/issues/15245

Differential Revision: D90620860

### DIFF
--- a/velox/type/StringView.h
+++ b/velox/type/StringView.h
@@ -92,7 +92,17 @@ struct StringView {
       // small string: inlined. Zero the last 8 bytes first to allow for whole
       // word comparison.
       value_.data = nullptr;
+#ifdef __GNUC__
+      // Use asm volatile to prevent GCC's interprocedural analysis from
+      // incorrectly concluding that size could exceed kInlineSize. This is a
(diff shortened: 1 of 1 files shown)


kevinwilfong (approved):
Thanks!

PR #16004, opened 2026-01-13
fix(ci): Clang-tidy coroutines flag unused

czentgr:
When the coroutines were enabled a new flag was added for gcc to ensure the correct compilation.
But this is also an unknown compile flag to clang-tidy/clang and needs to be removed.

Let's also add the stdout output if a failure occurs for easier diagnostic.


majetideepak (approved):

czentgr:
Checked with a build that modified a file to see if clang-tidy works and it was successful now: https://github.com/facebookincubator/velox/actions/runs/20974196836/job/60285024022?pr=16004

Issue #15485, opened 2025-11-13
FlatMapVector not compatible with BaseVector::copy()


kagamiori:
### Bug description

Calling BaseVector::copy() on a FlatMapVector throws the following error. It's because BaseVector::create() called inside copy() always creates regular MapVector.

```
TEST_F(FlatMapVectorTest, zzz) {
  std::vector<std::string> baseData = {
      "{1:10, 2:20, 3:null}",
      "null",
      "{}",
      "{4:40, 5:null, 6:null}",
  };
  auto baseFlatMap = makeFlatMapVectorFromJson<int64_t, int32_t>(baseData);
  auto copy = BaseVector::copy(*baseFlatMap, pool());
}
```
```
Line: fbcode/velox/vector/ComplexVector.cpp:509, Function:copyRangesImpl, Expression: leafSource->encoding() == encoding() (FLAT_MAP vs. MAP), Source: RUNTIME, ErrorCode: INVALID_STATE
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: (FLAT_MAP vs. MAP)
Retriable: False
Expression: leafSource->encoding() == encoding()
Function: copyRangesImpl
File: fbcode/velox/vector/ComplexVector.cpp
Line: 509
Stack trace: ...
```

### System information

N/A

### Relevant logs

```bash

```

kagamiori:
cc @pedroerp

peterenescu:
Will be resolved by https://github.com/facebookincubator/velox/pull/16007

peterenescu:
`BaseVector::copy` uses `BaseVector::create` and `copyRanges` to perform deep copies. By design, neither of these are compatible with encodings and are designed to flatten.

Jimmy added support for deep encoding copy in `EncodedVectorCopy` and we just recently added support in all cases for FlatMapVector. If we want to preserve flat maps in memory, we should use that copy.

Closing

PR #16007, opened 2026-01-14
refactor(LocalMerge): Preserve encoding copy during template output vector creation

peterenescu:
Summary: Support k-way merging of flat map encodeded columns. This will require some minimal support from BaseVector::create to support generating template FlatMapVectors. This will prevent errors like we see in [https://github.com/facebookincubator/velox/issues/15485](https://l.facebook.com/l.php?u=https%3A%2F%2Fgithub.com%2Ffacebookincubator%2Fvelox%2Fissues%2F15485&h=AT2DhTcKC08_uO_B6kqHDP7jLEI7UlrAxB7mlopQzjOYDW7Hk6HetnBRUdwburF_1o8Hq8W-r5-E1gyYP-A-kIqa3YJMJ1CyFdxZ3vm3yKbrvSLztcm3FcsoLFVcSi8kqEa4l1Rbpct9jo13ZqCA3J5kIMZm) where copies fail due to mismatched vector encodings.

Differential Revision: D90622855

### DIFF
--- a/velox/exec/Merge.cpp
+++ b/velox/exec/Merge.cpp
@@ -386,10 +386,7 @@ RowVectorPtr SourceMerger::getOutput(
   VELOX_CHECK_GT(outputBatchRows_, 0);
 
   if (!output_) {
-    output_ = BaseVector::create<RowVector>(type_, outputBatchRows_, pool_);
-    for (auto& child : output_->children()) {
-      child->resize(outputBatchRows_);
-    }
+    output_ = createOutputVector();
   }
 
   for (;;) {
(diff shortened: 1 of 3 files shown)


velox/exec/Merge.cpp
```diff
@@ -435,6 +432,27 @@ RowVectorPtr SourceMerger::getOutput(
...
+      output_ = std::static_pointer_cast<RowVector>(
+          BaseVector::createEmptyLike(source, outputBatchRows_, pool_));
+      return;
+    }
+  }
+
+  // If a non-null stream cannot be found, default to generating row vector by
+  // type.
+  output_ = BaseVector::create<RowVector>(type_, outputBatchRows_, pool_);
+  for (auto& child : output_->children()) {
```
pedroerp: btw, do we even need this? Looks like `BaseVector::create()` will already create the children recursively with the same size as the parent, no?

velox/exec/Merge.cpp
```diff
@@ -386,10 +386,7 @@ RowVectorPtr SourceMerger::getOutput(
   VELOX_CHECK_GT(outputBatchRows_, 0);
 
   if (!output_) {
-    output_ = BaseVector::create<RowVector>(type_, outputBatchRows_, pool_);
-    for (auto& child : output_->children()) {
-      child->resize(outputBatchRows_);
-    }
+    createOutputVector();
```
pedroerp: nit: 

`
output_ = createOutputVector();
`

would make what this function does even more obvious.

pedroerp (approved):
Small comment, but overall looks good.

velox/exec/Merge.cpp
peterenescu: @pedroerp Yes that's true, I kept it the same as before but I suppose for conciseness, I can make it a one liner

velox/exec/Merge.cpp
peterenescu: Sure thing let me add

PR #16011, opened 2026-01-14
feat: Add Spark monthname function

yaooqinn:
Implements the monthname() function for Spark SQL which returns the three-letter abbreviated month name from a given date.

The implementation matches Spark's behavior by using tm_mon values (0-11) and returning standard three-letter abbreviations.

**Spark Reference:**
- JIRA: [SPARK-46515](https://issues.apache.org/jira/browse/SPARK-46515)


velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -813,6 +813,55 @@ struct DayNameFunction {
...
+        result.append("Sep");
+        break;
+      case 9:
+        result.append("Oct");
+        break;
+      case 10:
+        result.append("Nov");
+        break;
+      default:
+        result.append("Dec");
```
philo-he: Can we use a static string array to hold these names in order and use `tm.tm_mon` as index to get the name?

velox/docs/functions/spark/datetime.rst
```diff
@@ -253,6 +253,15 @@ These functions support TIMESTAMP and DATE input types.
 
         SELECT month('2009-07-30'); -- 7
 
+.. spark:function:: monthname(date) -> varchar
+
+    Returns the three-letter abbreviated month name from the given date (Jan, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep, Oct, Nov, Dec). ::
```
philo-he: Nit: Could break this code line into two for better readability. Long line is not recommended in Velox documentation.

yaooqinn:
Thank you for the review @PHILO-HE! I've addressed both comments:

1. Replaced the switch statement with a static `constexpr std::string_view kMonthNames[]` array and use `tm.tm_mon` as the index.
2. Split the long documentation line into two shorter lines for better readability.

velox/docs/functions/spark/datetime.rst
```diff
@@ -253,6 +253,16 @@ These functions support TIMESTAMP and DATE input types.
 
         SELECT month('2009-07-30'); -- 7
 
+.. spark:function:: monthname(date) -> varchar
+
+    Returns the three-letter abbreviated month name from the given date.
```
philo-he: Nit: from -> for

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -813,6 +813,31 @@ struct DayNameFunction {
...
+        "May",
+        "Jun",
+        "Jul",
+        "Aug",
+        "Sep",
+        "Oct",
+        "Nov",
+        "Dec"};
+    const auto tm = getDateTime(date);
+    result.append(kMonthNames[tm.tm_mon]);
```
philo-he, line 838: It seems `setNoCopy` can be used here to optimize.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -813,6 +813,31 @@ struct DayNameFunction {
...
 };
 
+template <typename T>
+struct MonthNameFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<Varchar>& result,
+      const arg_type<Date>& date) {
+    static constexpr std::string_view kMonthNames[] = {
```
philo-he: It seems Velox has a convention that prefers moving such static constexpr variable to struct body in function implementation.

yaooqinn:
Thanks @PHILO-HE for the additional feedback! I've addressed all three new comments:

1. Changed "from" to "for" in the documentation
2. Used `setNoCopy` instead of `append` for optimization  
3. Moved the static constexpr array to the struct body (as a private member)

PHILO-HE (approved):
Looks good. It seems unrelated code changes are included. Please revert them. Thanks.

yaooqinn:
Thanks @PHILO-HE! I've reverted the unrelated changes to Transform.cpp and TransformTest.cpp.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -813,6 +813,33 @@ struct DayNameFunction {
...
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<Varchar>& result,
+      const arg_type<Date>& date) {
+    const auto tm = getDateTime(date);
+    result.setNoCopy(StringView(kMonthNames_[tm.tm_mon]));
+  }
+
+ private:
+  static constexpr std::string_view kMonthNames_[] = {
```
philo-he: Nit: seems no need to make it private. Then, rename it to `kMonthNames` without underscore.

PHILO-HE (commented):
@rui-mo, please take a look.

yaooqinn:
Thanks @PHILO-HE! Removed the private section and renamed `kMonthNames_` to `kMonthNames`.

velox/functions/sparksql/DateTimeFunctions.h
zhli1142015: Is setNoCopy inappropriate here? It might be working fine only because the inputs are always inlined (size < 12). If performance is truly a concern, you may consider wrapping a constant vector with a dictionary vector.

yaooqinn:
Thanks @zhli1142015 for the feedback! You're right that `setNoCopy` is risky here - it only works because the 3-char month names happen to be inlined. I've switched back to `append()` which is safer and consistent with `DayNameFunction`.

velox/docs/functions/spark/datetime.rst
```diff
@@ -253,6 +253,16 @@ These functions support TIMESTAMP and DATE input types.
 
         SELECT month('2009-07-30'); -- 7
 
+.. spark:function:: monthname(date) -> varchar
+
+    Returns the three-letter abbreviated month name for the given date.
```
rui-mo: nit: use \`\`date\`\`  to reference to the input variable.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -813,6 +813,32 @@ struct DayNameFunction {
   }
 };
 
+template <typename T>
+struct MonthNameFunction {
```
rui-mo, line 817: Could you please add the reference of Spark's implementation in the PR description?

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
```diff
@@ -676,6 +676,24 @@ TEST_F(DateTimeFunctionsTest, dayNameDate) {
...
+  EXPECT_EQ("Jan", monthName(0));
+
+  // Test various months across different years.
+  EXPECT_EQ("Jan", monthName(parseDate("2017-01-27")));
+  EXPECT_EQ("Feb", monthName(parseDate("2008-02-20")));
+  EXPECT_EQ("May", monthName(parseDate("2011-05-06")));
+  EXPECT_EQ("Aug", monthName(parseDate("2023-08-20")));
+  EXPECT_EQ("Aug", monthName(parseDate("2023-08-21")));
+  EXPECT_EQ(
+      "Oct", monthName(parseDate("1582-10-15"))); // Gregorian calendar start.
```
rui-mo, line 694: Spark’s implementation states that it supports input dates starting from 1970-01-01. Could you please help clarify what the expected behavior is when the input date is earlier than that?
Also, what is the behavior when the input date falls before the start of the Gregorian calendar? If there are any special cases, it might be helpful to add unit tests and document these corner cases as well.

```
  /**
   * Returns the three-letter abbreviated month name for the given number of days since 1970-01-01.
   */
  def getMonthName(days: Int): UTF8String = {
    val monthName = Month
      .of(getMonth(days))
      .getDisplayName(TextStyle.SHORT, DateFormatter.defaultLocale)

    UTF8String.fromString(monthName)
  }
```

rui-mo (commented):
Thanks for the prompt update.

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
yaooqinn: There is no such a restriction in spark, the test case here are ported exactly from spark side

rui-mo: Could you please add a test for input date being earlier than the start of Gregorian calendar, and ensure the result is consistent with Spark?

yaooqinn: It seems irrelevant to this PR. You mean the existing `getDateTime` might have consistent issues?

rui-mo: No exactly, I'm just wondering if we could add test to ensure the monthname function returns identical result with Spark when the input date is before the start of Gregorian calendar.

rui-mo (approved):

velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
yaooqinn: monthname works as a map of month numeric days. If we want to assess that part, I think we shall start from there. Spark' datetimes' migration to java8 API and calendar switch has very large amount of inconsistenct behavior within spark as well

rui-mo: I understand your point now, thanks.

PR #16014, opened 2026-01-14
feat(function): Add randstr Spark function

yaooqinn:
Add Spark SQL `randstr(length, seed)` function that returns a random string of the specified length with characters chosen uniformly at random from the pool: 0-9, a-z, A-Z.

### DIFF
--- a/velox/docs/functions/spark/string.rst
+++ b/velox/docs/functions/spark/string.rst
@@ -293,10 +293,11 @@ String Functions
     Both ``length`` and ``seed`` must be non-null constants.
     ``length`` must be a non-negative integer (SMALLINT or INT).
     ``seed`` must be an integer (INT or BIGINT).
-    With the same ``seed`` and partition, the function produces a reproducible sequence
+    With the same ``seed`` and partition ID, the function produces a reproducible sequence
     of outputs, though each row receives a different value from the sequence as the
     internal generator advances.
-    The partition is determined by the ``spark_partition_id`` query configuration.
+    The partition ID is retrieved from the ``spark_partition_id`` query configuration.
+    It's consistent with Spark's internal assignment for tasks.
     Uses XORShift random number generator matching Spark's implementation.
     Note: Spark's analyzer always provides a seed (either user-specified or
     auto-generated), so only the seeded variant is implemented.
--- a/velox/functions/lib/XORShiftRandom.h
+++ b/velox/functions/lib/XORShiftRandom.h
@@ -17,7 +17,7 @@
 
 #include <cstdint>
 
-namespace facebook::velox::functions::sparksql {
+namespace facebook::velox::functions {
 
 /// XORShift random number generator matching Spark's XORShiftRandom.
 /// Based on Marsaglia, G. (2003). Xorshift RNGs. Journal of Statistical
@@ -134,4 +134,4 @@ class XORShiftRandom {
   int64_t seed_{0};
 };
 
-} // namespace facebook::velox::functions::sparksql
+} // namespace facebook::velox::functions
--- a/velox/functions/sparksql/RandStr.h
+++ b/velox/functions/sparksql/RandStr.h
@@ -19,7 +19,7 @@
 
 #include "velox/core/QueryConfig.h"
 #include "velox/functions/Macros.h"
-#include "velox/functions/sparksql/XORShiftRandom.h"
+#include "velox/functions/lib/XORShiftRandom.h"
 
 namespace facebook::velox::functions::sparksql {
 
@@ -47,8 +47,9 @@ struct RandStrFunction {
       const core::QueryConfig& config,
       const TLen* length,
       const TSeed* seed) {
-    VELOX_USER_CHECK_NOT_NULL(length, "length must not be null");
-    VELOX_USER_CHECK_NOT_NULL(seed, "seed must be a constant value");
+    VELOX_USER_CHECK_NOT_NULL(
+        length, "length argument must be a constant value");
+    VELOX_USER_CHECK_NOT_NULL(seed, "seed argument must be a constant value");
     VELOX_USER_CHECK_GE(
         static_cast<int64_t>(*length), 0, "length must be non-negative");
     length_ = static_cast<int32_t>(*length);
@@ -76,7 +77,7 @@ struct RandStrFunction {
 
  private:
   int32_t length_{0};
-  XORShiftRandom generator_;
+  functions::XORShiftRandom generator_;
 };
 
 } // namespace facebook::velox::functions::sparksql
--- a/velox/functions/sparksql/tests/XORShiftRandomTest.cpp
+++ b/velox/functions/sparksql/tests/XORShiftRandomTest.cpp
@@ -14,12 +14,14 @@
  * limitations under the License.
  */
 
-#include "velox/functions/sparksql/XORShiftRandom.h"
+#include "velox/functions/lib/XORShiftRandom.h"
 
 #include <gtest/gtest.h>
 
 namespace facebook::velox::functions::sparksql::test {
 namespace {
+
+using facebook::velox::functions::XORShiftRandom;
 
 class XORShiftRandomTest : public testing::Test {};
 


velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,27 @@ String Functions
...
+        SELECT randstr(0); -- ""
+
+.. spark:function:: randstr(length, seed) -> varchar
+    :noindex:
+
+    Returns a string of the specified ``length`` whose characters are chosen uniformly
+    at random from the following pool of characters: 0-9, a-z, A-Z.
+    ``length`` must be a constant non-negative integer (SMALLINT or INT).
+    ``seed`` must be a constant integer (INT or BIGINT).
+    With the same ``seed`` and partition, the function produces deterministic results. ::
```
jinchengchenghh: What's the `partition` specification?

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,121 @@
...
+//   to match Spark's per-partition determinism.
+// - Without seed, the function is non-deterministic and uses thread-local
+//   randomness.
+template <typename T>
+struct RandStrFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  static constexpr bool is_deterministic = false;
+
+  // Overload for: randstr(length)
```
jinchengchenghh: Remove this comment, or update it to The implementation of randstr(length).

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,121 @@
...
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& /*config*/,
+      const TLen* length) {
+    VELOX_CHECK_NOT_NULL(length, "length argument must be constant");
+    VELOX_USER_CHECK_GE(
+        static_cast<int64_t>(*length), 0, "length must be non-negative");
+    hasSeed_ = false;
+  }
+
+  // Overload for: randstr(length, seed)
```
jinchengchenghh: ditto

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,27 @@ String Functions
         SELECT overlay('Spark SQL', 'tructured', 2, 4); -- "Structured SQL"
         SELECT overlay('Spark SQL', '_', -6, 3); -- "_Sql"
 
+.. spark:function:: randstr(length) -> varchar
```
jinchengchenghh: Please describe the corresponding Spark function.
I could not find execute it,
```
spark-sql (default)> select randstr(5);
[UNRESOLVED_ROUTINE] Cannot resolve function `randstr` on search path [`system`.`builtin`, `system`.`session`, `spark_catalog`.`default`].; line 1 pos 7
```

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,27 @@ String Functions
...
+        SELECT randstr(5); -- "ceV0P" (random)
+        SELECT randstr(0); -- ""
+
+.. spark:function:: randstr(length, seed) -> varchar
+    :noindex:
+
+    Returns a string of the specified ``length`` whose characters are chosen uniformly
+    at random from the following pool of characters: 0-9, a-z, A-Z.
+    ``length`` must be a constant non-negative integer (SMALLINT or INT).
+    ``seed`` must be a constant integer (INT or BIGINT).
```
jinchengchenghh: Please describe the behavior when the seed is null

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,121 @@
...
+  static constexpr int kPoolSize = 62;
+
+  FOLLY_ALWAYS_INLINE void generateUnseeded(
+      out_type<Varchar>& out,
+      int32_t length) {
+    // length validation is performed in initialize for Constant args. Guard
+    // here as well for safety.
+    VELOX_USER_CHECK_GE(length, 0, "length must be non-negative");
+    out.resize(length);
+    for (int32_t i = 0; i < length; ++i) {
```
jinchengchenghh: for (auto

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,27 @@ String Functions
...
+
+    Returns a string of the specified ``length`` whose characters are chosen uniformly
+    at random from the following pool of characters: 0-9, a-z, A-Z.
+    ``length`` must be a constant non-negative integer (SMALLINT or INT).
+    The function is non-deterministic. ::
+
+        SELECT randstr(5); -- "ceV0P" (random)
+        SELECT randstr(0); -- ""
+
+.. spark:function:: randstr(length, seed) -> varchar
```
philo-he, line 289: Suggest unifying the two kinds of inputs in the following form to indicate seed arg is optional.

`spark:function:: randstr(length[, seed]) -> varchar`

yaooqinn:
Sorry for the late reply, all comments addressed. @jinchengchenghh @PHILO-HE, thank you for the input

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,124 @@
...
+// - seed (if provided) must be a constant INT/LONG.
+// - With seed provided, generator is initialized with (seed + sparkPartitionId)
+//   to match Spark's per-partition determinism.
+// - Without seed, the function is non-deterministic and uses thread-local
+//   randomness.
+template <typename T>
+struct RandStrFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  static constexpr bool is_deterministic = false;
```
jinchengchenghh, line 40: You need two struct, randstr(length, seed) is deterministic

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,124 @@
...
+
+  /// The implementation of randstr(length).
+  template <typename TLen>
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& /*config*/,
+      const TLen* length) {
+    VELOX_CHECK_NOT_NULL(length, "length argument must be constant");
```
jinchengchenghh: Please return Status, and use the APIs to check https://github.com/facebookincubator/velox/pull/14589

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,124 @@
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    VELOX_CHECK_NOT_NULL(length, "length argument must be constant");
+    VELOX_USER_CHECK_GE(
+        static_cast<int64_t>(*length), 0, "length must be non-negative");
+    const int64_t s = seed ? static_cast<int64_t>(*seed) : 0;
+    const int64_t partitionId = config.sparkPartitionId();
```
jinchengchenghh: remove the variable since it is used only once

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,124 @@
...
+      // Use thread-local randomness for non-deterministic behavior.
+      uint32_t r = folly::Random::rand32();
+      out.data()[i] = kPool[r % kPoolSize];
+    }
+  }
+
+  FOLLY_ALWAYS_INLINE void generateSeeded(
+      out_type<Varchar>& out,
+      int32_t length) {
+    VELOX_USER_CHECK_GE(length, 0, "length must be non-negative");
```
jinchengchenghh: Status

yaooqinn:
@jinchengchenghh Thanks for the review! I've addressed the feedback:

**Split into two structs** ✅
- Created `RandStrFunction` (unseeded) and `RandStrSeededFunction` (seeded)
- Updated registration to use appropriate struct for each signature
- Note: Both structs use `is_deterministic = false` because this matches Spark's `Nondeterministic` trait. Even with a seed, results vary by `sparkPartitionId`, so the function is not globally deterministic - only per-partition deterministic.

**Removed temporary variable** ✅
- Inlined the seed calculation directly in `generator_.seed()`

**Regarding Status API (lines 48 and 111)**:
The simple function framework does **not** currently support `Status` return from `initialize()` - it only detects methods returning `void` (see [SimpleFunctionMetadata.h L927-947](https://github.com/facebookincubator/velox/blob/main/velox/core/SimpleFunctionMetadata.h#L927-L947)).

The `VELOX_USER_RETURN_*` macros from [PR #14589](https://github.com/facebookincubator/velox/pull/14589) are designed for `call()` methods that return `Status`. Since we're validating constant arguments in `initialize()`, using `VELOX_USER_CHECK_GE` (which throws VeloxUserError) is the correct pattern, consistent with other Spark functions:
- [RegexFunctions.cpp L70](https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/RegexFunctions.cpp#L70) - `VELOX_USER_CHECK` in `initialize()`
- [ToJson.h L352](https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/ToJson.h#L352) - `VELOX_USER_CHECK` in `initialize()`

Other changes:
- Moved `kPool` and `kPoolSize` to file-scope anonymous namespace
- Changed selection to `generator_() % kPoolSize` to match Spark's modulo approach

jinchengchenghh:
Thanks for your detailed explanation.
For spark, there are all non-deterministic, but for Velox, same inputs and same config produce same results, so this should be determinism in Vleox, what do you think about that? @rui-mo 
```
Determinism
^^^^^^^^^^^

By default simple functions are assumed to be deterministic, e.g. given the
same inputs they always produce the same results. If this is not the case,
the function must define a static constexpr bool is_deterministic member:
```
> Note: Both structs use is_deterministic = false because this matches Spark's Nondeterministic trait. Even with a seed, results vary by sparkPartitionId, so the function is not globally deterministic - only per-partition deterministic.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,109 @@
...
+
+  static constexpr bool is_deterministic = false;
+
+  template <typename TLen, typename TSeed>
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    VELOX_CHECK_NOT_NULL(length, "length argument must be constant");
```
jinchengchenghh: VELOX_USER_CHECK_NOT_NULL

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <folly/Random.h>
+#include <random>
+
+#include "velox/core/QueryConfig.h"
+#include "velox/functions/Macros.h"
+
+namespace facebook::velox::functions::sparksql {
+
+namespace {
```
jinchengchenghh: Use namespace detail if anouymous namespace in header file

velox/functions/sparksql/RandStr.h
philo-he: `deterministic` is true means same result is produced for same inputs, which I assume also include input data in the processing row (although not used in this function). So maybe, this function is not deterministic when seed is provided. Please confirm it.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,109 @@
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    VELOX_CHECK_NOT_NULL(length, "length argument must be constant");
+    VELOX_USER_CHECK_GE(
+        static_cast<int64_t>(*length), 0, "length must be non-negative");
+    generator_.seed(
+        (seed ? static_cast<int64_t>(*seed) : 0) + config.sparkPartitionId());
```
philo-he: If seed is nullptr, it means it's not constant, which is not allowed, right? Please confirm if Spark allows seed is a column reference.

philo-he: In the added document and comments, like "If seed is null, it is treated as 0.",
if it means seed is NULL constant, according to my previous investigation, Velox will internally convert it to 0  when `static_cast<int64_t>(*seed)` is used.


In Velox, checking `const TSeed* seed` is nullptr or not determines if it is constant. However, it cannot distinguish if seed is pointing to NULL constant or non-NULL constant.

velox/functions/sparksql/registration/RegisterString.cpp
```diff
@@ -40,6 +41,32 @@ void registerSparkStringFunctions(const std::string& prefix) {
 namespace sparksql {
 void registerStringFunctions(const std::string& prefix) {
   registerSparkStringFunctions(prefix);
+  // randstr(length) - non-deterministic
+  registerFunction<RandStrFunction, Varchar, Constant<int16_t>>(
```
philo-he: With `Constant` used, do we still need to check if the arg is constant or not in initialize method? Please confirm.

PHILO-HE (commented):
Some comments. Please confirm if them make sense. Thanks.

philo-he:
> Thanks for your detailed explanation. For spark, there are all non-deterministic, but for Velox, same inputs and same config produce same results, so this should be determinism in Vleox, what do you think about that? @rui-mo
> 
> ```
> Determinism
> ^^^^^^^^^^^
> 
> By default simple functions are assumed to be deterministic, e.g. given the
> same inputs they always produce the same results. If this is not the case,
> the function must define a static constexpr bool is_deterministic member:
> ```
> 
> > Note: Both structs use is_deterministic = false because this matches Spark's Nondeterministic trait. Even with a seed, results vary by sparkPartitionId, so the function is not globally deterministic - only per-partition deterministic.

See comment: https://github.com/facebookincubator/velox/pull/16014#discussion_r2734629183. In my understanding, even in one partition, same input in different rows cannot always produce same result. So I think it should be non-deterministic.

philo-he:
One minor suggestion on code organization. Would it be better to move this function to `Rand.h`?

velox/functions/sparksql/RandStr.h
jinchengchenghh: What is input data in the processing row (although not used in this function) you mentioned? Could you clarify more, I don't catch the context.

yaooqinn:
## Responses to Review Comments

### Question 1: Determinism (Line 39)

> `deterministic` is true means same result is produced for same inputs... So maybe, this function is not deterministic when seed is provided. Please confirm it.

**Confirmed.** `is_deterministic = false` is correct for both seeded and unseeded cases.

Even with a constant seed, different rows produce different outputs because the RNG advances after each row. I verified this with Spark 4.0:

```sql
SELECT id, randstr(5, 42) FROM range(5);
+---+--------------+
| id|randstr(5, 42)|
+---+--------------+
|  0|         pll6Y|
|  1|         OIJNn|
|  2|         I2NS5|
|  3|         bEWFX|
|  4|         kbQpB|
+---+--------------+
```

Same inputs (`5, 42`), different outputs per row. This matches Spark's implementation where `RandStr extends Nondeterministic` unconditionally.

The sequence is **reproducible** (same results across runs with same seed), but not **deterministic** in Velox's definition (same inputs → same output per row).

---

### Question 2: Seed nullptr meaning (Line 86)

> If seed is nullptr, it means it's not constant, which is not allowed, right? Please confirm if Spark allows seed is a column reference.

With `Constant<TSeed>` wrapper in registration, the framework guarantees seed is always a constant expression. So `seed == nullptr` means **NULL constant**, not a column reference.

Spark does NOT allow seed as a column reference:
```sql
spark-sql> SELECT randstr(5, id) FROM range(5);
ERROR: [DATATYPE_MISMATCH.NON_FOLDABLE_INPUT] Cannot resolve "randstr(5, id)" 
due to data type mismatch: the input `seed` should be a foldable integer 
expression; however, got "id". SQLSTATE: 42K09
```

In our implementation:
- `seed == nullptr` → NULL constant (treated as seed = 0 per Spark semantics)
- `seed != nullptr` → non-NULL constant value

The current code handles this correctly.

---

### Question 3: Constant check in initialize (Line 45)

> With `Constant` used, do we still need to check if the arg is constant or not in initialize method? Please confirm.

No, we don't need to check constness - the `Constant<T>` wrapper in registration already guarantees it.

What we DO check is whether the constant is NULL (nullptr). The current code:
```cpp
if (seed == nullptr) {
  generator_.seed(0);  // NULL constant treated as 0
} else {
  generator_.seed(static_cast<int64_t>(*seed) + config.sparkPartitionId());
}
```

This is the correct pattern - we're handling NULL vs non-NULL constant values, not checking constness.

---

I've also updated the documentation to clarify the distinction between "reproducible" and "deterministic":
- Changed "produces deterministic results" → "produces a reproducible sequence of outputs"
- Changed example comment "(deterministic with seed)" → "(reproducible with seed)"

jinchengchenghh:
Could you verify the native code with Gluten can produce same results with JVM Spark?

yaooqinn:
Due to different RNG algorithms, I don't think the results are identical

**Spark**: Uses `XORShiftRandom` (Marsaglia xorshift algorithm) with seed hashing
**Velox**: Uses `std::mt19937` (Mersenne Twister)

However, this matches Velox's existing pattern - `rand()` also use `std::mt19937` 
instead of Spark's XORShiftRandom.

jinchengchenghh:
> Due to different RNG algorithms, I don't think the results are identical
> 
> **Spark**: Uses `XORShiftRandom` (Marsaglia xorshift algorithm) with seed hashing **Velox**: Uses `std::mt19937` (Mersenne Twister)
> 
> However, this matches Velox's existing pattern - `rand()` also use `std::mt19937` instead of Spark's XORShiftRandom.

Make sense!

velox/functions/sparksql/RandStr.h
philo-he: @jinchengchenghh, it means the batch data for Velox to handle row by row. Here, it is not the input for this function, but the function is called for each row to get the result. Considering two rows have same data, the randstr function will return different results very likely, which I think doesn't match the definition of deterministic.

jinchengchenghh: Thanks for you explain.

philo-he:
> What we DO check is whether the constant is NULL (nullptr). The current code:
> 
> if (seed == nullptr) {
>   generator_.seed(0);  // NULL constant treated as 0
> } else {
>   generator_.seed(static_cast<int64_t>(*seed) + config.sparkPartitionId());
> }
> This is the correct pattern - we're handling NULL vs non-NULL constant values, not checking constness.

@yaooqinn, according to my understanding, we can only know the constness by checking if it is nullptr. 

See the following `unpackInitialize` function and the relevant code calling it. Could you confirm again? Thanks.
https://github.com/facebookincubator/velox/blob/6280634669da3b164534677a4b5db621e1f26876/velox/expression/SimpleFunctionAdapter.h#L220

yaooqinn:
@PHILO-HE You're correct. I made a mistake at that point.

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,25 @@ String Functions
...
+    ``seed``, if provided, must be a constant integer (INT or BIGINT).
+    If ``seed`` is null, it is treated as 0.
+    With the same ``seed`` and partition, the function produces a reproducible sequence
+    of outputs, though each row receives a different value from the sequence as the
+    internal generator advances.
+    The partition is determined by the ``spark_partition_id`` query configuration.
+    Without ``seed``, the function is non-deterministic.
+    This function was added in Spark 4.0. ::
+
+        SELECT randstr(5); -- "ceV0P" (random)
```
zhli1142015: My understanding is that Spark provides a default seed if the user doesn't specify one. Therefore, is it necessary for us to register and implement a version without the seed parameter?

yaooqinn:
Thank you @zhli1142015 for raising this question! I did a deep investigation into Spark's source code.

**Spark's Internal Behavior:**
You're correct that Spark's analyzer (`ResolveRandomSeed` rule) generates a random seed when none is provided:
```scala
case e: ExpressionWithRandomSeed if e.seedExpression == UnresolvedSeed =>
  e.withNewSeed(random.nextLong())
```
So internally, Spark always uses a seeded generator - it just generates a new random seed at query analysis time if the user doesn't provide one.

**Velox's Existing Pattern:**
However, looking at how `rand()` is already implemented in Velox (`RandFunction`), it follows the same two-variant pattern:
- `rand()` → uses `folly::Random::randDouble01()` directly
- `rand(seed)` → uses seeded `generator_`

The `randstr` implementation follows this same established Velox pattern for consistency:
- `randstr(length)` → uses `folly::Random::rand32()` per character  
- `randstr(length, seed)` → uses seeded `std::mt19937`

**Conclusion:**
While the implementation differs slightly from Spark's internal approach, both produce the same external behavior (non-deterministic vs reproducible output). More importantly, the current implementation is **consistent with how `rand()` already works in Velox**. Changing `randstr` to only have a seeded variant would break this consistency.

Do you think we should keep the current approach for consistency with `rand()`, or would you prefer to align more closely with Spark's internal implementation?

zhli1142015:
Yes, Spark’s analyzer will always provide a seed. Gluten translates the RandStr expression into the corresponding Velox function at the execution layer. Since at that point the expression always contains a seed—either explicitly specified by the user or injected by Spark’s analyzer—it will only be mapped to the version that takes the seed parameter.
Thanks.

yaooqinn:
Thank you @zhli1142015 for the clarification! That makes complete sense.

I've removed the unseeded variant since Gluten will always map to the seeded version. Changes in 974fc2ae0:
- Removed `randstr(length)` signatures (kept only `randstr(length, seed)`)
- Removed `folly/Random.h` dependency
- Updated tests to use seeded variant only

The implementation is now simpler and aligns with how Gluten/Spark actually uses it.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,86 @@
...
+      const TLen* length,
+      const TSeed* seed) {
+    // With Constant<TLen>, length is guaranteed to be constant.
+    // nullptr means NULL constant, which is not allowed for length.
+    VELOX_USER_CHECK_NOT_NULL(length, "length must not be null");
+    VELOX_USER_CHECK_GE(
+        static_cast<int64_t>(*length), 0, "length must be non-negative");
+    // NULL seed constant (nullptr) is treated as 0 per Spark semantics.
+    generator_.seed(
+        (seed ? static_cast<int64_t>(*seed) : 0) + config.sparkPartitionId());
```
zhli1142015: I think we need to add a check to ensure seed is not null, which means seed is a constant.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /// Called for seeded variant. Uses the seeded generator for reproducibility.
+  template <typename TLen, typename TSeed>
+  FOLLY_ALWAYS_INLINE bool callNullable(
```
zhli1142015: I think we should implement the call function here.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,86 @@
...
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    // With Constant<TLen>, length is guaranteed to be constant.
+    // nullptr means NULL constant, which is not allowed for length.
+    VELOX_USER_CHECK_NOT_NULL(length, "length must not be null");
+    VELOX_USER_CHECK_GE(
+        static_cast<int64_t>(*length), 0, "length must be non-negative");
```
zhli1142015, line 54: My understanding is that both length and seed here should be constants, so we need to store them as properties of the struct in this function.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,86 @@
...
+    const auto len = static_cast<int32_t>(*length);
+    out.resize(len);
+    for (auto i = 0; i < len; ++i) {
+      out.data()[i] = detail::kPool[generator_() % detail::kPoolSize];
+    }
+    return true;
+  }
+
+ private:
+  std::mt19937 generator_;
```
zhli1142015: This is not strictly required, but I think it’s feasible. We could implement XORShiftRandom so that its behavior aligns with Spark. I understand that result consistency is important when this function is used in reproducible experiments. Also, the implementation of `XORShiftRandom` is not very complex.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,86 @@
...
+/// Note: Spark's analyzer always provides a seed (either user-specified or
+/// auto-generated), so only the seeded variant is needed.
+///
+/// Note: Even with a constant seed, different rows produce different outputs
+/// as the generator advances, so is_deterministic is set to false.
+template <typename T>
+struct RandStrFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  static constexpr bool is_deterministic = false;
```
zhli1142015, line 40: With a fixed seed and length, the result is deterministic, so I consider it a deterministic function at the moment.

yaooqinn:
Thank you @zhli1142015 for the detailed review! I've addressed all your comments in 2443e88ad:

**1. Check seed is not null (ensures constness)** ✅
Added `VELOX_USER_CHECK_NOT_NULL(seed, "seed must not be null")` which validates the seed is a constant (non-constant args result in nullptr).

**2. Use `call()` instead of `callNullable()`** ✅
Since both length and seed are guaranteed constants and validated in `initialize()`, switched to the simpler `call()` method.

**3. Store length as struct member** ✅
Now storing `length_` in `initialize()` and using it in `call()`, avoiding redundant parameter access.

**4. Implement XORShiftRandom** ✅
Added `XORShiftRandom.h` implementing Spark's XORShift algorithm with proper MurmurHash3 seed hashing. This ensures result compatibility with Spark.

**5. Regarding `is_deterministic`** 
Kept it as `false` per PHILO-HE's earlier clarification - even with fixed seed and length, different rows produce different values because the generator advances. This matches other similar functions like `uuid`, `rand`, and `monotonically_increasing_id`.

velox/functions/sparksql/registration/RegisterString.cpp
```diff
@@ -40,6 +41,27 @@ void registerSparkStringFunctions(const std::string& prefix) {
 namespace sparksql {
 void registerStringFunctions(const std::string& prefix) {
   registerSparkStringFunctions(prefix);
+  // randstr(length, seed) - Spark's analyzer always provides a seed
```
zhli1142015: nit: Comments should end with ‘.’.

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,82 @@
...
+  /// Initialize for seeded variant: randstr(length, seed).
+  /// Both length and seed must be non-null constants.
+  template <typename TLen, typename TSeed>
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    VELOX_USER_CHECK_NOT_NULL(length, "length must not be null");
+    VELOX_USER_CHECK_NOT_NULL(seed, "seed must not be null");
```
zhli1142015: nit: seed must be a contant value.

zhli1142015:
> Thank you @zhli1142015 for the detailed review! I've addressed all your comments in [2443e88](https://github.com/facebookincubator/velox/commit/2443e88ad96259d6cceee33660946cdf2ccdd89f):
> 
> **1. Check seed is not null (ensures constness)** ✅ Added `VELOX_USER_CHECK_NOT_NULL(seed, "seed must not be null")` which validates the seed is a constant (non-constant args result in nullptr).
> 
> **2. Use `call()` instead of `callNullable()`** ✅ Since both length and seed are guaranteed constants and validated in `initialize()`, switched to the simpler `call()` method.
> 
> **3. Store length as struct member** ✅ Now storing `length_` in `initialize()` and using it in `call()`, avoiding redundant parameter access.
> 
> **4. Implement XORShiftRandom** ✅ Added `XORShiftRandom.h` implementing Spark's XORShift algorithm with proper MurmurHash3 seed hashing. This ensures result compatibility with Spark.
> 
> **5. Regarding `is_deterministic`** Kept it as `false` per PHILO-HE's earlier clarification - even with fixed seed and length, different rows produce different values because the generator advances. This matches other similar functions like `uuid`, `rand`, and `monotonically_increasing_id`.

Thanks for the update, maybe we can consider to use the `XORShiftRandom` in the rand function in a separate PR also.

yaooqinn:
> Thanks for the update, maybe we can consider to use the XORShiftRandom in the rand function in a separate PR also.

+1, let's update it later. BTW, any principles for keep backward compatibility should I be aware of, it looks a breakingchange.

yaooqinn:
Addressed the nit comments in 8b5f18b9a:
- Added period at end of comment in RegisterString.cpp
- Changed error message to "seed must be a constant value" for clarity

velox/functions/sparksql/RandStr.h
```diff
@@ -0,0 +1,82 @@
...
+
+  /// Initialize for seeded variant: randstr(length, seed).
+  /// Both length and seed must be non-null constants.
+  template <typename TLen, typename TSeed>
+  FOLLY_ALWAYS_INLINE void initialize(
+      const std::vector<TypePtr>& /*inputTypes*/,
+      const core::QueryConfig& config,
+      const TLen* length,
+      const TSeed* seed) {
+    VELOX_USER_CHECK_NOT_NULL(length, "length must not be null");
```
philo-he: Could you revise the error message for consistency with seed? Or you can use "length argument" and "seed argument" to indicate function's args.

`length must be a constant value`

velox/functions/lib/XORShiftRandom.h
philo-he, line 1: Suggest moving to `velox/functions/lib` for common use.

velox/functions/sparksql/RandStr.h
philo-he, line 1: Could you move the code to Rand.h for convenience to maintain the shared logic?

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,26 @@ String Functions
...
         SELECT overlay('Spark SQL', '_', -6, 3); -- "_Sql"
 
+.. spark:function:: randstr(length, seed) -> varchar
+
+    Returns a string of the specified ``length`` whose characters are chosen uniformly
+    at random from the following pool of characters: 0-9, a-z, A-Z.
+    Both ``length`` and ``seed`` must be non-null constants.
+    ``length`` must be a non-negative integer (SMALLINT or INT).
+    ``seed`` must be an integer (INT or BIGINT).
+    With the same ``seed`` and partition, the function produces a reproducible sequence
```
philo-he: partition -> partition ID

I feel using partition may be misleading since it sounds like the partition itself must be same.

velox/docs/functions/spark/string.rst
```diff
@@ -286,6 +286,26 @@ String Functions
...
+    of outputs, though each row receives a different value from the sequence as the
+    internal generator advances.
+    The partition is determined by the ``spark_partition_id`` query configuration.
```
philo-he: Suggestion:

The partition ID is retrieved from the query config. It's consistent with Spark's internal assignment for tasks.

PR #16015, opened 2026-01-14
misc: Remove step to upload benchmark results to conbench

kevinwilfong:
Summary:
We have lost access to the conbench hosts managed by Voltron Data. Removing the step to 
upload benchmark results there. They are still available in the artifacts.

Differential Revision: D90694195


kevinwilfong:
See https://github.com/facebookincubator/velox/actions/runs/20982384603/job/60309942381 for an example of a recent failure

PR #16016, opened 2026-01-14
refactor: Modify aggregates under prestosql to take the name as part of registration

kevinwilfong:
Summary:
Currently the aggregates under prestosql hard code the names the aggregates are registered
with. This makes it hard to reuse these functions in other engines where they may have
different names.

This change modifies the registration functions to take a collection of names rather than just
the prefix, so the caller has the flexibility to use different names.

Differential Revision: D90706025

### DIFF
--- a/velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp
@@ -21,7 +21,6 @@
 #include "velox/exec/Aggregate.h"
 #include "velox/expression/FunctionSignature.h"
 #include "velox/functions/lib/KllSketch.h"
-#include "velox/functions/prestosql/aggregates/AggregateNames.h"
 #include "velox/vector/DecodedVector.h"
 #include "velox/vector/FlatVector.h"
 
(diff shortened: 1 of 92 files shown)


velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp
```diff
@@ -892,7 +890,7 @@ void addSignatures(
 } // namespace
 
 void registerApproxPercentileAggregate(
-    const std::string& prefix,
+    const std::vector<std::string>& names,
```
kgpai, line 893: Can we create an issue for replacing prefix with names ? Might external users be affected ?

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
```diff
@@ -469,18 +467,17 @@ void registerInternalArrayAggAggregate(
           .argumentType("E")
           .build()};
 
-  auto name = prefix + "$internal$array_agg";
   exec::registerAggregateFunction(
-      name,
```
kgpai, line 474: nit: Lets make sure the names passed here is '$internal$array_agg' as this is used in fuzzer validation.

velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp
kevinwilfong: https://github.com/facebookincubator/velox/issues/16043

velox/functions/prestosql/aggregates/ArrayAggAggregate.cpp
kevinwilfong: The fuzzers currently rely on calling

facebook::velox::aggregate::prestosql::registerInternalAggregateFunctions();

to ensure this function and $internal$count_distinct are registered, I've ensured that function continues to register these aggregates with the same name

kevinwilfong:
Rebased

PR #16019, opened 2026-01-14
build: Use FBThrift instead of Apache Thrift (#16019) (#16019)

peterenescu:
Summary:
Removes external Thrift dependency primarily used in Parquet as noted by [https://github.com/facebookincubator/velox/issues/13175](https://github.com/facebookincubator/velox/issues/13175).

*Issues noted below for which we'd like to move to FBThrift from Apache Thrift.*

FBThrift and Apache Thrift has many incompatible API changes:

* `XXX.__isset.YYY` -> `XXX.YYY().has_value()` (or `bool(XXX.YYY())`): The latter implicit `bool(...)` syntax is used in this PR
* `XXX.__isset.YYY = true` -> `XXX.YYY().ensure()`
* `XXX.YYY` -> `XXX.YYY().value()` (or `*XXX.YYY()`): The latter `*...` syntax is used in this PR
* `XXX.__set_YYY(value)` -> `XXX.YYY() = value`
* `thrift::ThriftTransport`/`TCompactProcotolT` -> `folly::IOBuf`/`CompactProtocolReader`
* FBThrift doesn't support `optional` and default value that is used by `DataPageHeaderV2.is_compressed`: We need to use the default value manually in our code
* FBThrift raises an exception when we access to an optional field that isn't set (Apache Thrift doesn't have the check)

There are many changes in this PR but there is no logic change and no refactoring. This just converts API usage for FBThrift.

(Copy of [https://github.com/facebookincubator/velox/issues/14942](https://github.com/facebookincubator/velox/pull/14942) from kou due to some push permission issues.)

Differential Revision: D90723225

Pulled By: peterenescu


.github/workflows/linux-build-base.yml
```diff
@@ -339,6 +353,30 @@ jobs:
...
+          fetch-depth: 2
+          path: velox
+          persist-credentials: false
+
+      - name: Install Dependencies
+        env:
+          VELOX_BUILD_SHARED: "ON"
+          VELOX_ARROW_CMAKE_PATCH: ${{ github.workspace }}/velox/CMake/resolve_dependency_modules/arrow/cmake-compatibility.patch
+        run: |
+          if git diff --name-only HEAD^1 HEAD | grep -q "scripts/setup-"; then
```
czentgr: Make sure this is only triggered on a `pull_request`.
Similar to https://github.com/facebookincubator/velox/pull/15896

.github/workflows/macos.yml
```diff
@@ -74,6 +80,16 @@ jobs:
           source scripts/setup-macos.sh
           install_build_prerequisites
           install_velox_deps_from_brew
+          # We can use Homebrew's FBThrift once Homebrew's Folly
+          # enables 128bit integer support. We need to fix
+          # https://github.com/facebook/folly/issues/1666 for it.
+          install_boost
+          install_fmt
+          install_folly
```
czentgr, line 92: The setup script for macOS also installs fast_float just like for Linux. So you probably need that here too.

CMake/resolve_dependency_modules/README.md
```diff
@@ -32,18 +32,17 @@ by Velox. See details on bundling below.
 | fmt               | 11.2.0          | Yes      | Used API must be fmt 9 compatible |
 | simdjson          | 4.1.0           | Yes      ||
 | faiss             | 1.11.0          | Yes      ||
-| folly             | v2025.04.28.00  | Yes      ||
-| fizz              | v2025.04.28.00  | No       ||
-| wangle            | v2025.04.28.00  | No       ||
-| mvfst             | v2025.04.28.00  | No       ||
-| fbthrift          | v2025.04.28.00  | No       ||
+| folly             | v2025.09.15.00  | Yes      ||
```
czentgr: I'm trying to upgrade folly to a later version (2026.01.05.00). We will have a conflict here. The only issue for now is the velox_remote_functions_client_test is failing. I'm investigating why. It shouldn't fail. See https://github.com/facebookincubator/velox/pull/15967

Why did you pick this version (or just because that was the latest at the time you started)?

czentgr (commented):
Can you please explain the rationale a bit more? Looks like we want Arrow to link to fbthrift instead of apache thrift. I suppose fbthrift is compatible with apache thrift for arrow usage?

majetideepak:
Is this similar to https://github.com/facebookincubator/velox/pull/14942?
I believe the Gluten team saw a regression 
See https://github.com/facebookincubator/velox/pull/14942#issuecomment-3660245151

velox/dwio/parquet/thrift/CMakeLists.txt
```diff
@@ -12,5 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(velox_dwio_parquet_thrift ParquetThriftTypes.cpp)
-velox_link_libraries(velox_dwio_parquet_thrift arrow thrift Boost::headers fmt::fmt)
+include(FBThriftCppLibrary)
+
+add_fbthrift_cpp_library(velox_dwio_parquet_thrift_raw parquet.thrift)
+add_library(velox_dwio_parquet_thrift INTERFACE)
+target_link_libraries(velox_dwio_parquet_thrift INTERFACE velox_dwio_parquet_thrift_raw)
```
kgpai, line 19: We should use velox_link_libraries here , no need for the VELOX_MONO check.

kgpai:
FYI - The fuzzer builds fail because it requires _binary_types.cpp file ,  which was added after this change to fbthrift: https://github.com/facebook/fbthrift/commit/9ec44312b08fb069b542d01f7371006642706418 . The version of fbthrift we use is before this version. Landing https://github.com/facebookincubator/velox/pull/15967 and rebasing this change on top should fix the problem.

velox/dwio/parquet/thrift/CMakeLists.txt
kou: We want to use `INTERFACE` here. `velox_link_libraries()` doesn't support `INTERFACE`.

(I think that we should remove `velox_XXX()` wrappers and use `target_link_libraries()`/... directly eventually for easy to maintain.)

peterenescu:
@kgpai @kou @rui-mo 
Re-based resolving merge conflicts. Should be able to unblock you. There are some install dependency issues following fb os upgrade to `2026.01.05`, but that likely shouldn't affect the Gluten testing.

scripts/setup-common.sh
```diff
@@ -71,6 +71,16 @@ function install_mvfst {
 
 function install_fbthrift {
   wget_and_untar https://github.com/facebook/fbthrift/archive/refs/tags/"${FB_OS_VERSION}".tar.gz fbthrift
+  # We can't use "git apply" here because this is a patch for FBThrift
+  # not Velox. "git apply" in Velox repository tries applying a patch
+  # to Velox.
+  if command -v patch >/dev/null 2>&1; then
+    (cd "${DEPENDENCY_DIR}/fbthrift" &&
+      curl -L https://github.com/facebook/fbthrift/pull/675.patch | patch -p1)
```
peterenescu:
```
+ cd /home/runner/work/velox/velox/velox/deps-download/fbthrift
+ curl -L https://github.com/facebook/fbthrift/pull/675.patch
+ patch -p1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  9599  100  9599    0     0  46407      0 --:--:-- --:--:-- --:--:-- 46597
patch: **** Only garbage was found in the patch input.
```
Linux build fails likely due to double patch. Can this be removed @kou?

kou: Yes. https://github.com/facebook/fbthrift/pull/675 has been merged and `FB_OS_VERSION="v2026.01.05.00"` includes the patch.

rui-mo:
Hi @peterenescu, I added a reproducible test case for the hang issue when reading a Parquet file.
Test branch: https://github.com/rui-mo/velox/tree/test_fbthrift
The test performs a pure scan of the TPC-H lineitem table. See: `TEST_F(ParquetTableScanTest, test)`.
The test data is too large (~220 MB) to upload directly to GitHub, so it can be downloaded here:
https://drive.google.com/file/d/1nigKsfKrfByS2d0V0hP7HEMv8G0woU_8/view?usp=share_link
Please note that not all lineitem Parquet files reproduce this issue — you’ll need to use the file provided above.

cc: @kgpai @kou

kgpai:
cc: @majetideepak ^ Might you be able to help with this ?

kou:
Thanks! It's very helpful! I'll try it once I have some time.

peterenescu:
@majetideepak @rui-mo @kou Hi folks any update here?
cc: @kgpai

rui-mo:
@peterenescu @kou Please let me know if you are able to reproduce the infinite loop issue mentioned in https://github.com/facebookincubator/velox/pull/16019#issuecomment-3900008203 when reading a Parquet file. Thank you.

czentgr:
@rui-mo We reproduced the infinite loop with the test provided. Thanks. It is indeed an issue with processing buffer boundaries during a page scan (seek to page).
We have a fix for the refiller that seems to have not been using the API correctly. With some changes we are able to pass the repro test.

```
       OK ] ParquetTableScanTest.test (6150 ms)
[----------] 1 test from ParquetTableScanTest (6150 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (6451 ms total)
[  PASSED  ] 1 test.
```

Perhaps you can take the fix and retry with gluten?

kou:
Wow! Thanks @czentgr !!!

velox/dwio/parquet/thrift/ParquetThrift.h
```diff
@@ -0,0 +1,160 @@
...
+      }
+      buffer = folly::IOBuf::copyBuffer(data, dataBytes);
+    } else {
+      // Return only the new data - the currentData bytes are already in the
+      // reader's view
+      if (dataBytes >= requestedBytes - currentDataBytes) {
+        return folly::IOBuf::wrapBuffer(data, dataBytes);
+      }
+      buffer = folly::IOBuf::copyBuffer(data, dataBytes);
+    }
```
kou: We may be able to simplify this:

```suggestion
    // Return only the new data - the currentData bytes are already in the
    // reader's view
    if (dataBytes >= requestedBytes - currentDataBytes) {
      return folly::IOBuf::wrapBuffer(data, dataBytes);
    }
    buffer = folly::IOBuf::copyBuffer(data, dataBytes);
```

czentgr: Thanks! Yes, you are right, these are equivalent. I updated the commit.

rui-mo:
@czentgr Just a quick heads-up: your fix passes the previously failing TPC-H query in my test. I’ll also run larger-scale TPC-H and TPC-DS workloads to verify in the next two days. Thanks!

czentgr:
@peterenescu With @rui-mo's update I think this PR needs now cleanup/rebase to get it ready for a merge.

peterenescu:
Awesome! Thanks folks, exporting now, let's see how red CI looks 😅

peterenescu:
Hi @kou looks like there are some build failures following resolving merge conflicts. Are you able to take a look? You should have repo permissions

kou:
OK. I'll take a look at it.

rui-mo:
Hi everyone, we encountered another core dump issue while running TPC-DS Q4.
>apache::thrift::CompactProtocolReaderWithRefill::readFieldBegin(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, apache::thrift::protocol::TType&, short&) [clone .constprop.0]+0x127

The program fails after the refiller returns an IOBuf object whose data pointer is null.
```
    if (!readData(&data, &dataBytes)) {
      return folly::IOBuf::wrapBuffer(nullptr, 0);
    }
```

The possible cause is that the `setInput` call inside `ProtocolReaderWithRefill::ensureBuffer` fails with the null data pointer.

```
  void ensureBuffer(uint32_t size) {
    if (this->protocol_.in_.length() >= size) {
      return;
    }

    auto avail = this->protocol_.in_.peekBytes();
    buffer_ = refiller_(avail.data(), avail.size(), totalBytesRead(), size);
    if (!buffer_) {
      throw std::runtime_error("Empty buffer returned when refilling");
    }
    setInput(buffer_.get());
  }
```

cc: @kou @peterenescu @czentgr @kgpai

peterenescu:
@kou Included your changes and re-exported but it looks like the builds still fail. CI didn't start on your run due to merge conflicts

czentgr:
@rui-mo Is there again a repro for this SEGV? This time it operates on parquet files you have stored somewhere. Since it is again in a table scan could we isolate a query with a single file we can repro it again?

Looks like from the ensureBuffer point of view the refiller is not allowed to return empty data. So we would need to understand how it got to request data but without reading anything on refill.
Is there a full stack as well?

rui-mo:
@czentgr @peterenescu I wasn’t able to reproduce this issue using a pure scan, but could always reproduce it when running TPC-DS q4. I wonder if you have a way to test the TPC-DS queries. I can upload my test data for reproduction if needed. I assume further tests will be needed to verify the functionality of this PR, as merging it without sufficient tests could be risky.
Here is the stack: 
```
0# std::_Function_handler<std::unique_ptr<folly::IOBuf, std::default_delete<folly::IOBuf> > (unsigned char const*, int, int, int), facebook::velox::parquet::thrift::deserialize<facebook::velox::parquet::thrift::PageHeader>(facebook::velox::parquet::thrift::PageHeader*, facebook::velox::dwio::common::SeekableInputStream*, unsigned char const*, unsigned long)::{lambda(unsigned char const*, int, int, int)#2}>::_M_invoke(std::_Any_data const&, unsigned char const*&&, int&&, int&&, int&&) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 1# apache::thrift::ProtocolReaderWithRefill<apache::thrift::CompactProtocolReader>::ensureBuffer(unsigned int) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 2# 0x00007F94E875F7AA in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 3# 0x00007F94E876732B in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 4# facebook::velox::parquet::thrift::DeserializeResult facebook::velox::parquet::thrift::deserialize<facebook::velox::parquet::thrift::PageHeader>(facebook::velox::parquet::thrift::PageHeader*, facebook::velox::dwio::common::SeekableInputStream*, unsigned char const*, unsigned long) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 5# facebook::velox::parquet::PageReader::readPageHeader() in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 6# facebook::velox::parquet::PageReader::seekToPage(long) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 7# facebook::velox::parquet::PageReader::rowsForPage(facebook::velox::dwio::common::SelectiveColumnReader&, bool, bool, folly::Range<int const*>&, unsigned long const*&) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 8# void facebook::velox::parquet::PageReader::readWithVisitor<facebook::velox::dwio::common::ColumnVisitor<long, facebook::velox::common::AlwaysTrue, facebook::velox::dwio::common::ExtractToReader, false, true> >(facebook::velox::dwio::common::ColumnVisitor<long, facebook::velox::common::AlwaysTrue, facebook::velox::dwio::common::ExtractToReader, false, true>&) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
 9# void facebook::velox::dwio::common::SelectiveIntegerColumnReader::readHelper<facebook::velox::parquet::IntegerColumnReader, facebook::velox::common::AlwaysTrue, false, facebook::velox::dwio::common::ExtractToReader>(facebook::velox::common::Filter const*, folly::Range<int const*> const&, facebook::velox::dwio::common::ExtractToReader) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
10# facebook::velox::parquet::IntegerColumnReader::read(long, folly::Range<int const*> const&, unsigned long const*) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
11# 0x00007F94EA848678 in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
12# facebook::velox::dwio::common::ColumnLoader::loadInternal(folly::Range<int const*>, facebook::velox::ValueHook*, int, std::shared_ptr<facebook::velox::BaseVector>*) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
13# facebook::velox::VectorLoader::load(folly::Range<int const*>, facebook::velox::ValueHook*, int, std::shared_ptr<facebook::velox::BaseVector>*) in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
14# facebook::velox::LazyVector::loadVectorInternal() const in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
15# facebook::velox::LazyVector::loadedVector() in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
16# gluten::WholeStageResultIterator::next() in /data/tmp/gluten-05c58574-9947-4add-be43-af118e8b3732/jni/993edfeb-984e-49d2-92ec-175011efbd06/gluten-6436771930786767068/linux/amd64/libvelox.so
```

rui-mo:
@czentgr @peterenescu This issue doesn’t appear to be a corner case. When I tested the first 10 TPC-DS queries, Q4, Q5, and Q9 all failed with the same core dump, while the others haven’t been verified yet. I’m still working on finding a simpler query to reproduce the issue. It would also be very helpful if anyone could help verify the TPC-DS queries. Thanks.

czentgr:
@rui-mo Thanks, it is interesting because again it is in the seeking page header logic. It must be hitting another edge case.
This is not coming from a plain table scan but an intermediate result written to disk? I assume you tried already to read the store_sales table as it comes up in all the failing queries (but also in other queries).

rui-mo:
Hi @czentgr @peterenescu, I’m finally able to reproduce this issue in Velox using a simple query plan. Would you like to take a try?
Unit test: https://github.com/rui-mo/velox/commit/093c6d463322319e6a60c6e3c9407140d9897e5a
Test file: https://drive.google.com/file/d/1Rdy_RjNTUWtoh9iHwYZ3eQPRgqUjyFw0/view?usp=share_link

BTW, it appears that removing `project` and `partialAggregation` from the unit test still reproduces the core dump.

.github/workflows/linux-build-base.yml
```diff
@@ -118,6 +117,7 @@ jobs:
             mkdir /tmp/build
             cd /tmp/build
             source /opt/rh/gcc-toolset-12/enable
+            export VELOX_ARROW_CMAKE_PATCH="${GITHUB_WORKSPACE}/CMake/resolve_dependency_modules/arrow/arrow-testing-boost.patch ${GITHUB_WORKSPACE}/CMake/resolve_dependency_modules/arrow/cmake-compatibility.patch"
```
czentgr, line 132: Arrow was updated from Arrow 15 to Arrow 18. I see some issues with dependencies of Arrow for testing and boost in the CI. Looks like this patch was supposedly fixing this?

kou: Oh, sorry. I missed the upgrade.

Yes. It's for fixing arrow-testing and Boost but it's incompleted yet.
I tried fixing it with the patch when I have time but I couldn't fix it entirely. I will retry it when I have time again but I don't have enough time for now. Sorry. If the patch has any trouble, you can remove the patch. I may re-add something when I retry.

czentgr: @kou Yes, we will have to review this.

czentgr:
@peterenescu @rui-mo I've been working on a fix for the issue with the Parquet reading part.
@rui-mo Thanks again for the simple repro. What scale factor are you running?

I have a fix for this. I ran a Presto TPCDS SF1k workload and it passed successfully.
It needs some cleanup though to get it ready.

rui-mo:
@czentgr Thanks for verifying TPC-DS. I noticed this issue on a smaller dataset (SF500).

czentgr:
@rui-mo Would you be able to do another run? I ran on the SF1k and had no problems. I will re-run with SF10k TPCDS.

rui-mo:
Thanks @czentgr for the fix! I’ll run another test to verify it. By the way, I noticed another potential incompatibility between FBThrift and Apache Thrift (see https://github.com/facebookincubator/velox/issues/13175#issuecomment-4091968756) and wonder if there might be other issues.

majetideepak:
@rui-mo There should not be any compatibility issues between fbthrift and thrift since both are expected to have the same on-the-wire byte order (big endian). If at all there is a difference, it's a bug which we can fix. Not moving to fbthrift is a blocker for multiple other Velox improvements.

rui-mo:
@czentgr I verified the previously failed queries as well as all TPC-DS queries (SF500) on my end using your fix, and they are now working correctly. Thank you again for your fix!

czentgr:
@rui-mo @peterenescu I've pushed the latest changes. I had to make an adjustment to the new arrow patch to work with arrow-flight which is used in Prestissimo.
But I've run both TPCDS power 1k and 10k successfully. Performance looks good. There is no regression.
@rui-mo Do you want to give it another spin on the Gluten side before we claim success?

rui-mo:
Thanks @czentgr! I should have time to run another round of tests this week.

.github/workflows/ubuntu-bundled-deps.yml
```diff
@@ -70,7 +70,11 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          source scripts/setup-ubuntu.sh && install_apt_deps && install_faiss_deps
+          source scripts/setup-ubuntu.sh && \
+          install_apt_deps && install_faiss_deps && \
+          install_fmt && install_boost && install_fast_float && \
+          install_folly && install_fizz && install_wangle && install_mvfst && \
+          install_fbthrift
```
kou: We can remove `&&` and `\` because `shell: bash` has `-e`.
See also: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell

```suggestion
          source scripts/setup-ubuntu.sh
          install_apt_deps
          install_faiss_deps
          install_fmt
          install_boost
          install_fast_float
          install_folly
          install_fizz
          install_wangle
          install_mvfst
          install_fbthrift
```

CMake/third-party/FBThriftCppLibrary.cmake
```diff
@@ -137,10 +138,7 @@ function(add_fbthrift_cpp_library LIB_NAME THRIFT_FILE)
     set(LIB_TYPE STATIC)
   endif ()
 
-  add_library(
-    "${LIB_NAME}" ${LIB_TYPE}
-    ${generated_sources}
-  )
+  add_library("${LIB_NAME}" STATIC ${generated_sources})
```
kou, line 134: If we always use `STATIC`, we can remove `set(LIB_TYPE ...)` above.

kou (commented):
Thanks for updating this!

rui-mo:
@czentgr I verified the TPC-H and TPC-DS workloads in Gluten with the new patch, and both are working well. Thanks!

.github/workflows/linux-build-base.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - name: Install Dependencies
         env:
           VELOX_BUILD_SHARED: "ON"
-          VELOX_ARROW_CMAKE_PATCH: ${{ github.workspace }}/CMake/resolve_dependency_modules/arrow/cmake-compatibility.patch
+          VELOX_FBTHRIFT_CMAKE_PATCH: ${{ github.workspace }}/CMake/resolve_dependency_modules/fbthrift/compactv1-protocol-refiller.patch
```
zzhao0, line 119: The patch is not needed if we use 
https://github.com/facebook/fbthrift/releases/tag/v2026.05.04.00
or after

czentgr: Yes, we decided to add this here for now and then we'll update FBOS.

velox/dwio/parquet/thrift/parquet.thrift
```diff
@@ -37,7 +37,9 @@
  * File format description for the parquet file format
  */
 
-namespace cpp parquet.format
+include "thrift/annotation/thrift.thrift"
```
zzhao0, line 40: It doesn't seem like this include is used

velox/dwio/parquet/thrift/parquet.thrift
```diff
@@ -579,8 +581,9 @@ struct DataPageHeaderV2 {
   Which means the section of the page between
   definition_levels_byte_length + repetition_levels_byte_length + 1 and compressed_page_size (included)
   is compressed with the compression_codec.
-  If missing it is considered compressed */
-  7: optional bool is_compressed = 1;
+  If missing it is considered compressed.
+  Note: Code must use is_compressed().value_or(true) to handle the default. */
+  7: optional bool is_compressed;
```
zzhao0, line 586: Why removing the default? shouldn't the default value be true instead of 1?

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -665,22 +672,23 @@ std::unique_ptr<ParquetTypeWithId> ReaderBase::getParquetColumnInfo(
...
       }
     } else {
-      if (schemaElement.repetition_type ==
-          thrift::FieldRepetitionType::REPEATED) {
-        if (schema[parentSchemaIdx].converted_type ==
-            thrift::ConvertedType::LIST) {
+      if (schemaElement.repetition_type() &&
+          *schemaElement.repetition_type() ==
+              thrift::FieldRepetitionType::REPEATED) {
+        auto converted_type = schema[parentSchemaIdx].converted_type();
```
zzhao0: Let's be consistent with camel casing convention

velox/dwio/parquet/thrift/parquet.thrift
kou: Because FBThrift doesn't accept it.

It's mentioned in the PR description too:

> FBThrift doesn't support optional and default value that is used by `DataPageHeaderV2.is_compressed`: We need to use the default value manually in our code

velox/dwio/parquet/thrift/parquet.thrift
kou: It was needed when I tried. But FBThrift is updated after I tried initially. So we can try removing it.

velox/dwio/parquet/reader/ParquetReader.cpp
kou: Oh, sorry. There are some snake_case variables in this file. So I broke consistency...

@czentgr Could you fix this?

velox/dwio/parquet/thrift/parquet.thrift
czentgr: Yes, actually there appears to have been a bug previously. The default for this is true. That is if no flag is set then it should be true.
But the old implementation (as you can see in the diff) checked for the presence of the flag and the value. But if it was not set then we should assume compressed and previously that evaluated to not compressed.

So there is a fix in this as well and is highlighted in the commit message as well.

zzhao0 (approved):

kou:
Thanks!!!

PR #16021, opened 2026-01-15
feat: Expose 'runningIsBlockedWallNanos' operator runtime metric

lingbin:
The 'isBlockedTiming' field was present in 'OperatorStats' but not
exposed in the stats output, making it difficult to debug performance
issues related to 'Operator::isBlocked()' method calls. This change
exposes the field to help identify when operators are spending
significant time evaluating the 'isBlocked()'.

This enhancement will aid in troubleshooting performance issues by
providing visibility into the timing of 'isBlocked()' method calls,
which were previously hidden.

### DIFF
--- a/velox/exec/Task.cpp
+++ b/velox/exec/Task.cpp
@@ -138,6 +138,8 @@ void addRunningTimeOperatorMetrics(exec::OperatorStats& op) {
       RuntimeMetric(op.getOutputTiming.wallNanos, RuntimeCounter::Unit::kNanos);
   op.runtimeStats[std::string(OperatorStats::kRunningFinishWallNanos)] =
       RuntimeMetric(op.finishTiming.wallNanos, RuntimeCounter::Unit::kNanos);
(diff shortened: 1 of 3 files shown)


lingbin:
@mbasmanova  Could you help take a look?

velox/exec/Task.cpp
```diff
@@ -138,6 +138,8 @@ void addRunningTimeOperatorMetrics(exec::OperatorStats& op) {
       RuntimeMetric(op.getOutputTiming.wallNanos, RuntimeCounter::Unit::kNanos);
   op.runtimeStats[std::string(OperatorStats::kRunningFinishWallNanos)] =
       RuntimeMetric(op.finishTiming.wallNanos, RuntimeCounter::Unit::kNanos);
+  op.runtimeStats["runningIsBlockedWallNanos"] =
```
mbasmanova: Would you add a constant for this string like OperatorStats::kRunningXxx?

lingbin: Added, thanks.

PR #16025, opened 2026-01-15
misc: Add root pool name and state to HiveDataSink non-reclaimable warning

duxiao1212:
Summary:
Currently, the warning message appears as follows:

```
W0113 16:13:36.849711 115699 [Driver31097] HiveDataSink.cpp:1351] Can't reclaim from hive writer pool op.8.1.3.TableWrite.prism_batch.part[130] which is under non-reclaimable section, reserved memory: 74.00MB
```

However, we lack context regarding the root pool and its states. The hive writer pool can enter a non-reclaimable status in multiple scenarios. To aid in diagnosis, we should expose additional information about these cases.

Differential Revision: D90782787


tanjialiang (approved):
LGTM Thanks

xiaoxmeng (approved):
@duxiao1212 thanks!

PR #16027, opened 2026-01-15
feat: Add custom fuzzer tests for array_subset UDF

abhinavmuk04:
Summary:
Add comprehensive fuzzer tests for the `array_subset` UDF that verify behavior matches the equivalent expression using existing UDFs:

```sql
transform(
  array_sort(array_distinct(
    filter(indices, i -> i IS NOT NULL AND i > 0 AND i <= cardinality(arr))
  )),
  i -> element_at(arr, i)
)
```

These tests use VectorFuzzer to generate random inputs and compare results, helping catch edge cases and behavior differences that might not be covered by manual unit tests.

The 10 new fuzzer tests cover:
- Flat vectors with no nulls
- Arrays with nulls and variable length
- Dictionary-encoded vectors
- High null ratio scenarios
- String arrays (VARCHAR)
- Nested arrays (ARRAY(ARRAY(INTEGER)))
- Empty array edge cases
- Various numeric types (BIGINT, DOUBLE, BOOLEAN)
- Constant-encoded vectors
- Large vectors (stress test with 1000 rows)

Differential Revision: D90788140

### DIFF
--- a/velox/functions/prestosql/tests/ArraySubsetTest.cpp
+++ b/velox/functions/prestosql/tests/ArraySubsetTest.cpp
@@ -15,6 +15,9 @@
  */
 
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+#include "velox/vector/fuzzer/VectorFuzzer.h"
+
+#include <random>
 
 using namespace facebook::velox;
 using namespace facebook::velox::test;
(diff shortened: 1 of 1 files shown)


velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,319 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+  }
+
+  // The equivalent SQL expression for array_subset using existing UDFs.
+  // array_subset(arr, indices) is equivalent to:
+  // transform(
+  //   array_sort(array_distinct(
+  //     filter(indices, i -> i IS NOT NULL AND i > 0 AND i <= cardinality(arr))
+  //   )),
+  //   i -> element_at(arr, i)
+  // )
```
kevinwilfong: nit: This comment is a copy of the string just below and the comment at the top of the file, maybe just say something like

The expression below is equivalent to array_subset(c0, c1) using existing UDFs.

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,319 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+    auto inputArray = data->childAt(0);
+    auto indices = data->childAt(1);
+    SelectivityVector nonNullRows(data->size());
+
+    for (vector_size_t i = 0; i < data->size(); ++i) {
+      if (inputArray->isNullAt(i) || indices->isNullAt(i)) {
+        nonNullRows.setValid(i, false);
+      }
+    }
+    nonNullRows.updateBounds();
```
kevinwilfong, line 504: This could be rewritten as

```
nonNullRows.deselectNulls(inputArray->rawNulls(), 0, inputArray->size());
nonNullRows.deselectNulls(indices->rawNulls(), 0, indices->size());
```

This reuses existing code, handles calling updateBounds for you so it's less error prone, and (not that important here) would be more performant.

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,319 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+    auto nonNullRows = getNonNullRows(data);
+
+    // Compare only non-null rows (null propagation is tested separately)
+    for (auto i = 0; i < data->size(); ++i) {
+      if (nonNullRows.isValid(i)) {
+        ASSERT_TRUE(expected->equalValueAt(result.get(), i, i))
+            << "Mismatch at row " << i << ": expected " << expected->toString(i)
+            << ", got " << result->toString(i);
+      }
+    }
```
kevinwilfong, line 545:
```
nonNullRows.applyToSelected(
  [&expected, &result](const vector_size_t row) {
    ASSERT_TRUE(expected->equalValueAt(result.get(), row, row)
            << "Mismatch at row " << row << ": expected " << expected->toString(row)
            << ", got " << result->toString(row);
  }
);
```

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,319 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+
+  VectorFuzzer fuzzer(options, pool());
+
+  for (auto i = 0; i < 10; ++i) {
+    // Generate random integer arrays
+    auto inputArray = fuzzer.fuzz(ARRAY(INTEGER()));
+
+    // Generate indices array with values that may include valid and invalid
+    // indices
+    auto indices = fuzzer.fuzz(ARRAY(INTEGER()));
```
kevinwilfong: This will generate random 32-bit integers, the chances that any of them will be in the range 1-10 is very close to 0. This would probably be a more effective test if you took finer control of how the indices Vector is generated.

E.g. you could use VectorMaker's flatVector function to make a FlatVector of integers using VectorFuzzer's randInRange function to generate the values (range could be 0-11 to prioritize valid indices with a few invalid ones)

kevinwilfong: I see this is done repeatedly throughout this file, so it'd be good to make it available as a helper function.

abhinavmuk04: @kevinwilfong just addressed all the changes! Thanks for the review.

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,313 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+ protected:
+  // Helper to flatten vectors for consistent comparison across encodings
+  template <typename T>
+  static VectorPtr flatten(const std::shared_ptr<T>& vector) {
+    SelectivityVector allRows(vector->size());
+    auto result =
+        BaseVector::create(vector->type(), vector->size(), vector->pool());
+    result->copy(vector.get(), allRows, nullptr);
+    return result;
+  }
```
kevinwilfong, line 479: Any reason not to use BaseVector::flattenVector?

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +457,313 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+  // Get a SelectivityVector that excludes rows where either input is null.
+  // This is because array_subset returns NULL when either input is NULL,
+  // but the equivalent expression may return empty array.
+  static SelectivityVector getNonNullRows(const RowVectorPtr& data) {
+    auto inputArray = data->childAt(0);
+    auto indices = data->childAt(1);
+    SelectivityVector nonNullRows(data->size());
+
+    // Use isNullAt() for reliable null checking across all vector encodings.
+    // rawNulls() can be unreliable with complex vector types from the fuzzer.
```
kevinwilfong: I think you mean if the vectors are encoded, but fair point

velox/functions/prestosql/tests/ArraySubsetTest.cpp
kevinwilfong: @abhinavmuk04 it looks like you missed this one.

Just to make sure I wasn't missing something I added a line to throw an exception right after the 

    if (validIndices.empty()) {
      return;
    }

In the ArraySubset UDF, and ran these new tests.  Not one of them failed, so you're pretty much just testing the logic to filter out indices outside the range of the array.

velox/functions/prestosql/tests/ArraySubsetTest.cpp
abhinavmuk04: Thanks for pointing these out. Especially the comment before.

velox/functions/prestosql/tests/ArraySubsetTest.cpp
```diff
@@ -456,5 +459,387 @@ TEST_F(ArraySubsetTest, nullTest) {
...
+        options.vectorSize, 8, options.containerLength, options.nullRatio);
+
+    // Wrap in dictionary encoding
+    auto inputArray = fuzzer.fuzzDictionary(baseInputArray, options.vectorSize);
+    auto indices = fuzzer.fuzzDictionary(baseIndices, options.vectorSize);
+
+    auto data = makeRowVector({inputArray, indices});
+
+    // Flatten for comparison
+    auto flatData = makeRowVector({flatten(inputArray), flatten(indices)});
```
kevinwilfong, line 660: nit: flatten is recursive, so you could just call

auto flatData = makeRowVector({inputArray, indices});
BaseVector::flatten(flatData);

and get rid of flatten altogether.

PR #16030, opened 2026-01-15
feat: Add custom fuzzer tests for map_append UDF

abhinavmuk04:
Summary:
Added custom fuzzer tests for the map_append UDF to catch edge cases and verify the function behaves correctly with random inputs.

The fuzzer tests use VectorFuzzer to generate random input data and verify properties of the result. Since map_append has complex error conditions (duplicate keys, existing keys, mismatched lengths), the tests use `try(map_append(...))` to gracefully handle errors and verify that:
- When the function succeeds, the result map is at least as large as the input map
- Null handling is correct

Added 7 fuzzer test cases covering:
- Different data types (INTEGER, BIGINT, VARCHAR, DOUBLE, SMALLINT)
- High null ratio scenarios (50%)
- Large vector sizes (200 rows)

Differential Revision: D90791159

### DIFF
--- a/velox/functions/prestosql/tests/MapAppendTest.cpp
+++ b/velox/functions/prestosql/tests/MapAppendTest.cpp
@@ -15,6 +15,7 @@
  */
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/functions/prestosql/tests/utils/FunctionBaseTest.h"
+#include "velox/vector/fuzzer/VectorFuzzer.h"
 
 using namespace facebook::velox::test;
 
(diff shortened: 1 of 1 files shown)


velox/functions/prestosql/tests/MapAppendTest.cpp
```diff
@@ -363,4 +364,168 @@ TEST_F(MapAppendTest, preserveMapOrder) {
...
+  opts.vectorSize = 50;
+  opts.nullRatio = 0.1;
+  opts.containerLength = 5;
+  VectorFuzzer fuzzer(opts, pool());
+
+  auto inputMap = fuzzer.fuzz(MAP(SMALLINT(), SMALLINT()));
+  auto keys = fuzzer.fuzz(ARRAY(SMALLINT()));
+  auto values = fuzzer.fuzz(ARRAY(SMALLINT()));
+  auto data = makeRowVector({inputMap, keys, values});
+  testMapAppendProperties(data);
```
zacw7, line 514: this code pattern looks identical among many other tests (only the type is different). consider refactoring to use a templated function or parameterized test.

PR #16033, opened 2026-01-15
feat(connector): Add TPCH connector serdes

hdikeman:
Summary:
TPCH does not currently have serialization/deserialization APIs for TableHandle, ColumnHandle, or ConnectorSplit. I used the Hive connector serdes as a guide to add equivalent serdes for the TPCH connector. These serdes are relatively trivial given the simplicity of the TPCH connector structures.

Also adding test coverage for each serde type in `TpchConnectorSerDeTest.cpp`

Differential Revision: D90791949

### DIFF
--- a/velox/connectors/tpch/TpchConnector.h
+++ b/velox/connectors/tpch/TpchConnector.h
@@ -34,9 +34,27 @@ class TpchColumnHandle : public ColumnHandle {
  public:
   explicit TpchColumnHandle(const std::string& name) : name_(name) {}
 
-  const std::string& name() const {
+  const std::string& name() const override {
     return name_;
   }
+
+  folly::dynamic serialize() const override {
+    folly::dynamic obj = folly::dynamic::object;
+    obj["name"] = TpchColumnHandle::getClassName();
+    obj["columnName"] = name_;
+    return obj;
+  }
+
+  static std::shared_ptr<TpchColumnHandle> create(const folly::dynamic& obj) {
+    auto name = obj["columnName"].asString();
+    return std::make_shared<TpchColumnHandle>(name);
+  }
+
+  static void registerSerDe() {
+    registerDeserializer<TpchColumnHandle>();
+  }
+
+  VELOX_DEFINE_CLASS_NAME(TpchColumnHandle)
 
  private:
   const std::string name_;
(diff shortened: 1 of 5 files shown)


velox/connectors/tpch/tests/CMakeLists.txt
```diff
@@ -25,6 +25,20 @@ target_link_libraries(
   GTest::gtest_main
 )
 
+add_executable(velox_tpch_connector_serde_test TpchConnectorSerDeTest.cpp)
```
mbasmanova: In cmake, unlike in buck, it is perferable to combine all tests into a single target.

@kgpai @czentgr Krishna, Christian, please, correct / clarify if needed.

hdikeman: Got it, good to know! Let me do that. Thanks!

velox/connectors/tpch/TpchConnector.h
```diff
@@ -34,10 +34,27 @@ class TpchColumnHandle : public ColumnHandle {
...
   explicit TpchColumnHandle(const std::string& name) : name_(name) {}
 
-  const std::string& name() const {
+  const std::string& name() const override {
     return name_;
   }
 
+  folly::dynamic serialize() const override {
+    folly::dynamic obj = folly::dynamic::object;
+    obj["name"] = "TpchColumnHandle";
```
mbasmanova: perhaps, use VELOX_DEFINE_CLASS_NAME to avoid hard-coding the name multiple times.

hdikeman: Let me take a look at that, thanks for the pointer

velox/connectors/tpch/tests/CMakeLists.txt
kgpai: Thanks @mbasmanova . 
@hdikeman it is preferable in CMake to have not as fine grained targets as when using BUCK . One reason is significant write amplification that happens for every target . For e.g on average each test executable is around 200Mb (https://facebookincubator.github.io/velox/bm-report/#binary-sizes) . A lot of code is duplicated here as velox is statically linked. Even otherwise in such a large project as Velox having too many targets make it very hard to manage things and its better to have things less granular.

hdikeman: Thanks for the additional context @kgpai, I combined the new TpchConnectorSerDeTest with the existing TpchConnectorTest target

PR #16042, opened 2026-01-16
fix: Reduce dependencies on Parquet in Iceberg connector

kevinwilfong:
Summary:
https://github.com/facebookincubator/velox/pull/15607 introduced a dependency on the 
Parquet writer in Velox's Iceberg connector. This is breaking the build of projects that depend 
on the Iceberg connector and the Facebook Thrift library due to the duplicate symbols found
in the Apache Thrift library Parquet's writer depends on.

Fortunately, the Iceberg connector just depends on the ParquetFieldId class used by the
Parquet writer, which is itself very simple and does not have additional depedendencies.

In order to preserve the changes to the Iceberg connector and resolve the build issues, I've
split out the ParquetFieldId class into its own header file so the Iceberg connector can simply
include this header file, rather than the whole header file for the Parquet writer. This gives us
the flexibility to avoid introducing a transitive dependency on Apache Thrift.

Differential Revision: D90874048

### DIFF
--- a/velox/connectors/hive/iceberg/IcebergColumnHandle.h
+++ b/velox/connectors/hive/iceberg/IcebergColumnHandle.h
@@ -19,7 +19,7 @@
 #include <vector>
 
 #include "velox/connectors/hive/TableHandle.h"
-#include "velox/dwio/parquet/writer/Writer.h"
+#include "velox/dwio/parquet/writer/ParquetFieldId.h"
 #include "velox/type/Subfield.h"
 #include "velox/type/Type.h"
 
--- a/velox/connectors/hive/iceberg/CMakeLists.txt
+++ b/velox/connectors/hive/iceberg/CMakeLists.txt
@@ -29,6 +29,7 @@ velox_add_library(
 velox_link_libraries(
   velox_hive_iceberg_splitreader
   velox_connector
+  velox_dwio_parquet_field_id
   velox_functions_iceberg
   Folly::folly
 )
--- a/velox/connectors/hive/iceberg/IcebergColumnHandle.cpp
+++ b/velox/connectors/hive/iceberg/IcebergColumnHandle.cpp
@@ -20,7 +20,7 @@
 
 #include "velox/connectors/hive/TableHandle.h"
 #include "velox/connectors/hive/iceberg/IcebergColumnHandle.h"
-#include "velox/dwio/parquet/writer/Writer.h"
+#include "velox/dwio/parquet/writer/ParquetFieldId.h"
 #include "velox/type/Subfield.h"
 #include "velox/type/Type.h"
 
--- a/velox/dwio/parquet/writer/CMakeLists.txt
+++ b/velox/dwio/parquet/writer/CMakeLists.txt
@@ -14,6 +14,8 @@
 
 add_subdirectory(arrow)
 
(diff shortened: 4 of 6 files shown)


velox/connectors/hive/iceberg/IcebergColumnHandle.h
```diff
@@ -19,7 +19,7 @@
 #include <vector>
 
 #include "velox/connectors/hive/TableHandle.h"
-#include "velox/dwio/parquet/writer/Writer.h"
+#include "velox/dwio/parquet/writer/ParquetFieldId.h"
```
kgpai, line 22: Iceberg will still link against the parquet target though right ? It might be good if this can be moved out of the parquet target so that there is no chance in future we again run into duplicate symbol issue.

kevinwilfong: Check the internal diff, I separated the targets.

kgpai: Ah neat, should we do that for CMake too ? :)

kevinwilfong: sure, I was assuming this was just a problem for us

kevinwilfong:
@PingLiuPing thanks for reaching out. The problem lies in the fact that Presto depends on the Iceberg connector, e.g.
https://github.com/prestodb/presto/blob/acc0f202bc0f228e8a2d85ac690ac53245894be0/presto-native-execution/presto_cpp/main/connectors/IcebergPrestoToVeloxConnector.cpp#L19

and the fact that Velox's Parquet code depends on Apache Thrift which his historically caused a lot of issues because Velox and Presto also depend on FBThrift in a handful of places https://github.com/facebookincubator/velox/issues/13175

We generally work around this by guarding places that use Parquet with #ifdef VELOX_ENABLE_PARQUET, e.g.
https://github.com/facebookincubator/velox/blob/f92870afa2936c055d3f7df9448cb237896f20c6/velox/dwio/parquet/RegisterParquetWriter.cpp#L17-L32

So that we can avoid including Apache Thrift if we also need FBThrift

In the case of your change, I wasn't sure how to target such a guard (I think I'd have to completely disable the IcebergConnector given how central it is to the interface), and given it was just a small, straight forward class it was depending on, pulling out that class into a dependency that doesn't require Apache Thrift seemed like the best course of action.

There is a change being worked on to change the Parquet code in Velox to depend on FBThrift instead of Apache Thrift https://github.com/facebookincubator/velox/pull/14942 you may want to consider waiting on that change before introducing any more dependencies.

PingLiuPing:
@kevinwilfong Thank you very much for the explanation. I now understand that this is caused by a Thrift conflict from Prestissimo. Previously, I thought there might be something special happening internally in meta.

I can test this locally when I make new changes under dwio/parquet.

kevinwilfong:
@PingLiuPing I suspect we have stricter requirements within Meta about not using Apache Thrift (naturally we use FBThrift pretty much exclusively) but we also use a different build set up which makes isolating these dependencies easier.

I say this just as a warning that you may not be able to detect all issues on your own. If you need help getting some early signal reach out to a Meta engineer.

PR #16050, opened 2026-01-16
feat(cudf): Add `approx_distinct` global aggregation support

### LINKED ISSUE #16024: [cuDF] Add approx_distinct aggregation support
### Description

Add approx_distinct global aggregation support to CudfHashAggregation operator. A global aggregation is a reduction over all column values.

Aggregation implementations are dependent on the operator step. An approx_distinct which is part of a CudfHashAggregation with step[partial] could output the serialized `approx_distinct_count`. 

The step[final] implementation could then merge the sketches received from partial and then extract the estimate.

Currently, when CUDF execution is enabled, queries/plans that include approx_distinct fail at runtime with a NOT_IMPLEMENTED error. 

### Repro
Query: SELECT approx_distinct(x) FROM t;

### Expected behavior
approx_distinct should execute successfully under CUDF execution.

PointKernel:
Closes #16024

This PR adds cudf-based approx_distinct global aggregation.


devavret:
@PointKernel if it's not too much trouble, can you depend this PR on https://github.com/facebookincubator/velox/pull/15529 because that adds some signature checks and we'd have to add some for `approx_distinct`.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -419,6 +421,162 @@ struct MeanAggregator : cudf_velox::CudfHashAggregation::Aggregator {
...
+    VELOX_CHECK(
+        isGlobal,
+        "ApproxDistinctAggregator currently only supports global aggregation");
+  }
+
+  void addGroupbyRequest(
+      cudf::table_view const& tbl,
+      std::vector<cudf::groupby::aggregation_request>& requests) override {
+    VELOX_UNSUPPORTED(
+        "approx_distinct should use doReduce, not groupby requests");
```
devavret:
```suggestion
        "approx_distinct is not supported as a group aggregation");
```

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -419,6 +421,162 @@ struct MeanAggregator : cudf_velox::CudfHashAggregation::Aggregator {
...
+  }
+
+  std::unique_ptr<cudf::column> makeOutputColumn(
+      std::vector<cudf::groupby::aggregation_result>& results,
+      rmm::cuda_stream_view stream) override {
+    VELOX_UNSUPPORTED(
+        "approx_distinct should use doReduce, not makeOutputColumn");
```
devavret:
```suggestion
        "approx_distinct is not supported as a group aggregation");
```

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -419,6 +421,162 @@ struct MeanAggregator : cudf_velox::CudfHashAggregation::Aggregator {
...
+  std::unique_ptr<cudf::column> doPartialReduce(
+      cudf::table_view const& input,
+      rmm::cuda_stream_view stream) {
+    auto inputTable = cudf::table_view({input.column(inputIndex)});
+
+    cudf::approx_distinct_count sketch{
+        inputTable,
+        precision_,
+        cudf::null_policy::EXCLUDE,
+        cudf::nan_policy::NAN_IS_NULL,
```
devavret: Q: Are these verified to be compatible with CPU version of approx_distinct? The test with nulls expects <= 6 which could mean that null is counted as distinct.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -419,6 +421,162 @@ struct MeanAggregator : cudf_velox::CudfHashAggregation::Aggregator {
...
+        rmm::device_buffer{},
+        0);
+
+    return cudf::make_lists_column(
+        1,
+        std::move(offsets_column),
+        std::move(child_column),
+        0,
+        rmm::device_buffer{},
+        stream);
```
devavret: IIUC, sketch is being copied from inside the cudf::approx_distinct_count into a list column.
1. Is it verified that list is the return type of cpu implementation too?
2. It would be nice if cudf let you bring your own allocation for the scketch

velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -419,6 +421,162 @@ struct MeanAggregator : cudf_velox::CudfHashAggregation::Aggregator {
...
+        "approx_distinct should use doReduce, not makeOutputColumn");
+  }
+
+  std::unique_ptr<cudf::column> doReduce(
+      cudf::table_view const& input,
+      TypePtr const& outputType,
+      rmm::cuda_stream_view stream) override {
+    VELOX_CHECK(
+        step != core::AggregationNode::Step::kIntermediate,
+        "ApproxDistinctAggregator does not support intermediate aggregation step");
```
devavret: It should be easy to support intermediate aggregation. It's probably just a merge sketch without the estimate call.

PointKernel: Of course, I intentionally disabled it since the issue explicitly asks for partial and final.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
PointKernel: No, they are not. Should we require strict parity with the CPU behavior? If so, we would need to align a broader set of components like hashers, finalizer logic, and other algorithmic details which could require significantly more effort than anticipated.

devavret: Not strict parity, we can assume that partial results generated by cudf aggregation will be consumed by cudf versions of intermediate/final so algorithmic details are not expected to match. However results are expected to match.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
devavret: Alright. My bad, I'll update the issue.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
PointKernel: > However results are expected to match.

There are two things we need to verify for this to work. First, we must ensure that cudf's row hashing logic exactly matches velox’s, including null/NaN handling, nested type behavior, etc. Second, cuco currently uses HLL++, whereas velox uses KHLL; we need to use the exact same HLL algorithm to achieve matching results.

devavret: NaN and null handling should be the only things expected to match for now. I don't think approximations need to match but counting of nulls and nans should match. I'm not sure what velox does in this case but it should be easy to find out in unit tests.

GregoryKimball:
@PointKernel thank you for working through this feature. It turns out this is a critical item for Presto GPU to be able to run "Analyze tables". And analyze tables is a blocker to get valid query plans. Would you please move this PR out of draft when you've addressed all of @devavret 's review comments?

PointKernel:
I’ve rebased on https://github.com/facebookincubator/velox/pull/15529, excluded nulls, and treated NaNs as valid values to match Velox behavior. This PR is now ready for review.

velox/experimental/cudf/exec/CudfHashAggregation.cpp
PointKernel: > It would be nice if cudf let you bring your own allocation for the scketch

Are you referring to the case where users know the sketch size in advance, allocate it accordingly, and pass the allocated data span to the cuco HLL constructor?

devavret: yes. This is for memory management purposes though and not an immediate requirement for this PR

PointKernel: It’s doable. We already have this ctor https://github.com/rapidsai/cudf/blob/080c04ae620bbc22d12648973310d26bc6496945/cpp/include/cudf/reduction/approx_distinct_count.hpp#L115, but the current behavior still uses the provided span only as a source: it allocates new storage and copies the data into memory owned by `cudf::approx_distinct_count`.

We could make `cudf::approx_distinct_count` non-owning, which would better support multi-processing use cases like this.

devavret:
@PointKernel Now that #15529 is merged, you should also add this as a supported operator in `registerStepAwareBuiltinAggregationFunctions` otherwise it (gpu acceleration) won't be used. I found a pre-existing bug in the tests where it allows falling back to CPU mode and that's why your tests were passing. It can be fixed by disallowing fallback with this diff: 
```diff
diff --git a/velox/experimental/cudf/tests/AggregationTest.cpp b/velox/experimental/cudf/tests/AggregationTest.cpp
index b289affde..09629fe83 100644
--- a/velox/experimental/cudf/tests/AggregationTest.cpp
+++ b/velox/experimental/cudf/tests/AggregationTest.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "velox/experimental/cudf/CudfConfig.h"
 #include "velox/experimental/cudf/exec/ToCudf.h"
 
 #include "velox/dwio/common/tests/utils/BatchMaker.h"
@@ -40,6 +41,7 @@ class AggregationTest : public OperatorTestBase {
   void SetUp() override {
     OperatorTestBase::SetUp();
     filesystems::registerLocalFileSystem();
+    cudf_velox::CudfConfig::getInstance().allowCpuFallback = false;
     cudf_velox::registerCudf();
   }
```

kgpai:
@devavret @PointKernel can you fix the pre-commit format violations ?

karthikeyann:
> @devavret @PointKernel can you fix the pre-commit format violations ?

@kgpai https://github.com/facebookincubator/velox/pull/16265 PR fixes the pre-commit violations. The pre-commit failures is unrelated to this PR.

PR #16058, opened 2026-01-19
test: Revamp Spark cast tests to support ANSI on/off differentiation

infvg:
This PR revamps the SparkCastTest to test casts with ON/OFF.

### DIFF
--- a/velox/functions/sparksql/tests/SparkCastExprTest.cpp
+++ b/velox/functions/sparksql/tests/SparkCastExprTest.cpp
@@ -31,6 +31,182 @@ class SparkCastExprTest : public functions::test::CastBaseTest {
     memory::MemoryManager::testingSetInstance(memory::MemoryManager::Options{});
   }
 
+  void setAnsiSupport(bool value) {
+    queryCtx_->testingOverrideConfigUnsafe(
+        {{core::QueryConfig::kSparkAnsiEnabled, std::to_string(value)}});
+  }
+
+  void testBoolToTimestamp() {
+    testCast(
+        makeFlatVector<bool>({true, false}),
+        makeFlatVector<Timestamp>({
+            Timestamp(0, 1000),
+            Timestamp(0, 0),
+        }));
+  }
+
+  template <typename T>
+  void testIntegralToTimestampCast() {
+    testCast(
+        makeNullableFlatVector<T>({
+            0,
+            1,
+            std::numeric_limits<T>::max(),
+            std::numeric_limits<T>::min(),
+            std::nullopt,
+        }),
+        makeNullableFlatVector<Timestamp>(
+            {Timestamp(0, 0),
+             Timestamp(1, 0),
+             Timestamp(std::numeric_limits<T>::max(), 0),
+             Timestamp(std::numeric_limits<T>::min(), 0),
+             std::nullopt}));
+  }
+
+  void testIntToTimestamp() {
+    // Cast bigint as timestamp.
+    testCast(
+        makeNullableFlatVector<int64_t>({
+            0,
+            1727181032,
+            -1727181032,
+            9223372036855,
+            -9223372036856,
+            std::numeric_limits<int64_t>::max(),
+            std::numeric_limits<int64_t>::min(),
+        }),
+        makeNullableFlatVector<Timestamp>({
+            Timestamp(0, 0),
+            Timestamp(1727181032, 0),
+            Timestamp(-1727181032, 0),
+            Timestamp(9223372036854, 775'807'000),
+            Timestamp(-9223372036855, 224'192'000),
+            Timestamp(9223372036854, 775'807'000),
+            Timestamp(-9223372036855, 224'192'000),
+        }));
+
+    // Cast tinyint/smallint/integer as timestamp.
+    testIntegralToTimestampCast<int8_t>();
+    testIntegralToTimestampCast<int16_t>();
+    testIntegralToTimestampCast<int32_t>();
+  }
+
+  void testTinyIntToBinary() {
+    testCast<int8_t, std::string>(
+        TINYINT(),
+        VARBINARY(),
+        {18,
+         -26,
+         0,
+         110,
+         std::numeric_limits<int8_t>::max(),
+         std::numeric_limits<int8_t>::min()},
+        {std::string("\x12", 1),
+         std::string("\xE6", 1),
+         std::string("\0", 1),
+         std::string("\x6E", 1),
+         std::string("\x7F", 1),
+         std::string("\x80", 1)});
+  }
+
+  void testSmallintToBinary() {
+    testCast<int16_t, std::string>(
+        SMALLINT(),
+        VARBINARY(),
+        {180,
+         -199,
+         0,
+         12300,
+         std::numeric_limits<int16_t>::max(),
+         std::numeric_limits<int16_t>::min()},
+        {std::string("\0\xB4", 2),
+         std::string("\xFF\x39", 2),
+         std::string("\0\0", 2),
+         std::string("\x30\x0C", 2),
+         std::string("\x7F\xFF", 2),
+         std::string("\x80\00", 2)});
+  }
+
+  void testIntegerToBinary() {
+    testCast<int32_t, std::string>(
+        INTEGER(),
+        VARBINARY(),
+        {18,
+         -26,
+         0,
+         180000,
+         std::numeric_limits<int32_t>::max(),
+         std::numeric_limits<int32_t>::min()},
+        {std::string("\0\0\0\x12", 4),
+         std::string("\xFF\xFF\xFF\xE6", 4),
+         std::string("\0\0\0\0", 4),
+         std::string("\0\x02\xBF\x20", 4),
+         std::string("\x7F\xFF\xFF\xFF", 4),
+         std::string("\x80\0\0\0", 4)});
+  }
+
+  void testBigIntToBinary() {
+    testCast<int64_t, std::string>(
+        BIGINT(),
+        VARBINARY(),
+        {123456,
+         -256789,
+         0,
+         180000,
+         std::numeric_limits<int64_t>::max(),
+         std::numeric_limits<int64_t>::min()},
+        {std::string("\0\0\0\0\0\x01\xE2\x40", 8),
+         std::string("\xFF\xFF\xFF\xFF\xFF\xFC\x14\xEB", 8),
+         std::string("\0\0\0\0\0\0\0\0", 8),
+         std::string("\0\0\0\0\0\x02\xBF\x20", 8),
+         std::string("\x7F\xFF\xFF\xFF\xFF\xFF\xFF\xFF", 8),
+         std::string("\x80\x00\x00\x00\x00\x00\x00\x00", 8)});
+  }
+
+  void testFloatToTimestamp() {
+    testCast(
+        makeFlatVector<float>({
+            0.0,
+            1727181032.0,
+            -1727181032.0,
+            std::numeric_limits<float>::max(),
+            std::numeric_limits<float>::min(),
+        }),
+        makeNullableFlatVector<Timestamp>({
+            Timestamp(0, 0),
+            Timestamp(1727181056, 0),
+            Timestamp(-1727181056, 0),
+            Timestamp(9223372036854, 775'807'000),
+            Timestamp(0, 0),
+        }));
+  }
+
+  void testDoubleToTimestamp() {
+    testCast(
+        makeFlatVector<double>({
+            0.0,
+            1727181032.0,
+            -1727181032.0,
+            9223372036855.999,
+            -9223372036856.999,
+            1.79769e+308,
+            std::numeric_limits<double>::max(),
+            -std::numeric_limits<double>::max(),
+            std::numeric_limits<double>::min(),
+        }),
+        makeNullableFlatVector<Timestamp>({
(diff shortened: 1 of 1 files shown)


velox/expression/CastExpr-inl.h
```diff
@@ -187,7 +187,15 @@ void CastExpr::applyCastKernel(
...
-        setResultOrError(castResult, row);
+        if (castResult.hasError()) {
+          setError(castResult.error().message());
+        } else {
+          if (castResult.value().has_value()) {
+            result->set(row, castResult.value().value());
+          } else {
+            result->setNull(row, true);
+          }
+        }
```
rui-mo: This change would affect many Presto and Spark casts. Could you please limit the scope to the Spark to_timestamp cast only? Otherwise, all affected behaviors would need to be verified for both Presto and Spark.

rui-mo (commented):
Thanks!

velox/expression/CastExpr-inl.h
infvg: For presto, it can only return an error or a value. Only the spark hook can return `std::nullopt` so the null branch shouldn't be reachable with Presto. Please let me know what you think

philo-he: I assume the concern is not regarding to the functionality, but the performance and code maintenance. If possible, we should avoid add such Spark-specific or Presto-specific code in the common place, especially in the hot code path. Can we find a better way for this?

infvg: IMO this isn't Spark or Presto specific in this case. Presto is already ANSI compliant: https://github.com/prestodb/presto/blob/3f588e546738182a53318350b64ce194a6fce215/presto-main-base/src/main/java/com/facebook/presto/type/TimestampOperators.java#L212
and and this change will just allow the legacy spark functionality to work. We can see the same logic for double -> timestamp above

infvg:
@PHILO-HE  By default velox sets ANSI to false while presto is set to true. I saw your comment said to verify the regular cases for ANSI on & divergences for ANSI off. I set ANSI on for the affected tests & added a new test for the ANSI off cases, please take a look.

velox/expression/CastExpr-inl.h
philo-he: @infvg, It seems the `setNull` logic here is only specific to Spark's non-ANSI behavior. Is there a way to isolate this logic? While we have similar patterns elsewhere, I'd love to see if we can avoid Spark-specific or Presto-specific branches. Correct me if my understanding is wrong.

philo-he:
> @PHILO-HE By default velox sets ANSI to false while presto is set to true. I saw your comment said to verify the regular cases for ANSI on & divergences for ANSI off. I set ANSI on for the affected tests & added a new test for the ANSI off cases, please take a look.

@infvg, I meant we may need to test regular cases for both ANSI ON and OFF for better coverage, and also add sufficient divergence cases whose test expectation depends on the ANSI setting. I note this test strategy also exists in Spark. It has `CastSuiteBase` class which holds regular test cases, and its two child classes with ANSI set to ON and OFF respectively ensures the regular test cases in super class pass under the two different settings.

velox/expression/CastExpr-inl.h
infvg: @PHILO-HE We could pass a pointer to the result & row through the hooks and let them set whichever result they need. Let me know if this approach is good & I'll update all the hooks

philo-he: @infvg, I noticed that `setResultOrError` calls `setError`, which sets NULL as result or sets errors in the context based on `setNullInResultAtError()`. Perhaps we should simply update `setNullInResultAtError()` to respect the ANSI configuration. This would allow us to keep the cast logic inside the hooks unchanged and delegate the Null handling to the centralized `setResultOrError` logic. What do you think of this approach?

https://github.com/facebookincubator/velox/blob/e181822262fe5694aed733b295c2d54df9a506bb/velox/expression/CastExpr-inl.h#L105

rui-mo: We could also consider making `setNullInResultAtError` protected and overriding it in `SparkCastExpr` to avoid affecting Presto.

infvg: I changed it to override the `setNullInResultAtError` for `SparkCastExpr`. I think this will add ansi support for a lot of casts by extension. Please take a look

infvg:
@PHILO-HE I revamped the spark cast test suite to match Spark more. Please take a look

velox/expression/CastExpr-inl.h
philo-he: @infvg, Another thought. In PR #16059, `isTryCast` is updated to respect ANSI setting, and it also checks types to control the enabling for specific cast conversions. Could you evaluate which implementation is more proper? cc @rui-mo

rui-mo: Could the fix in https://github.com/facebookincubator/velox/pull/16059 address the issue here? If so, I’d prefer that approach since it feels clearer and simpler.

infvg:
@PHILO-HE @rui-mo I changed this PR to focus on revamping the test suite first.

philo-he:
@infvg, thanks for the update. I am afraid those tests under ANSI ON setting are not actually verifying the behavior of ANSI ON since some executions still go to ANSI OFF path maybe. 

Can we firstly add some control logic in tests for separate cast when adding the support? See the code below for reference.
https://github.com/malinjawi/velox/blob/e0f2202b665e3a07c596ac11fc3618c872d8f7ea/velox/functions/sparksql/tests/SparkCastExprTest.cpp#L471

Maybe, we can consider to revamp the cast tests when all kinds of ANSI cast are supported.

infvg:
@PHILO-HE I've added ANSI ON/OFF for all the the existing tests, but not all cast combinations go have ANSI on enabled yet due to this:
https://github.com/malinjawi/velox/blob/e0f2202b665e3a07c596ac11fc3618c872d8f7ea/velox/functions/sparksql/specialforms/SparkCastExpr.cpp#L21C4-L30C2

Once we start populating isAnsiSupported for all combinations, we can populate each individual test. 
For this PR, I separated the ANSI on/off logic to make it simpler to add tests.

All the tests under `SparkCastExprTestAnsiOn` automatically have isAnsiEnabled config property set to true:
https://github.com/facebookincubator/velox/pull/16058/changes#diff-850f8e84234883567d77deec5584ae4dd4e7b7a1a2ff361119af1410be057b11R954-R960

 This way we don't need to set Ansi on/off in individual tests. This is similar to Spark's implementation of seperating the Ansi ON/OFF tests.

velox/functions/prestosql/tests/utils/FunctionBaseTest.h
```diff
@@ -61,6 +61,11 @@ class FunctionBaseTest : public testing::Test,
     });
   }
 
+  void setAnsiSupport(bool value) {
+    queryCtx_->testingOverrideConfigUnsafe(
+        {{core::QueryConfig::kSparkAnsiEnabled, std::to_string(value)}});
```
rui-mo: This might be confusing to set Spark configuration in the Presto test base.

velox/functions/sparksql/specialforms/SparkCastHooks.h
```diff
@@ -78,6 +78,10 @@ class SparkCastHooks : public exec::CastHooks {
 
   exec::PolicyType getPolicy() const override;
 
+  bool isAnsiEnabled() const {
```
rui-mo: I did not find a caller for this function. Is it needed?

infvg: This was a part of the previous ansi support commits - removed

velox/functions/prestosql/tests/utils/FunctionBaseTest.h
infvg: Moved to SparkCastExprTest

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -139,78 +144,1148 @@ class SparkCastExprTest : public functions::test::CastBaseTest {
...
+  testInvalidCast<std::string>(
+      "timestamp",
+      {"2015-03-18X"},
+      "Cannot cast VARCHAR '2015-03-18X' to TIMESTAMP. Unable to parse timestamp value: \"2015-03-18X\", expected format is (YYYY-MM-DD HH:MM:SS[.MS])");
+
+  testInvalidCast<std::string>(
+      "timestamp",
+      {"abdef"},
+      "Cannot cast VARCHAR 'abdef' to TIMESTAMP. Unable to parse timestamp value: \"abdef\", expected format is (YYYY-MM-DD HH:MM:SS[.MS])");
+  */
```
rui-mo: Hi @infvg, could you clarify the reason for these tests to be excluded?

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -139,78 +144,1148 @@ class SparkCastExprTest : public functions::test::CastBaseTest {
         }),
         makeFlatVector<T>(expected));
   }
+  void testStringToDate() {
```
rui-mo: And what's the reason for extracting them as test methods, rather than putting them inside a test? Are these methods reused? Thanks.

velox/functions/sparksql/tests/SparkCastExprTest.cpp
infvg: These are ANSI on new tests that I added but they need the ansi setting to be enabled for those type casts. Once we enable those, we can use these tests.

velox/functions/sparksql/tests/SparkCastExprTest.cpp
infvg: These are common cases that need to pass for both ansi on & off. They're used in both the ansi on/off tests below & any ansi specific tests are added there instead.
Eg:
https://github.com/facebookincubator/velox/pull/16058/changes#diff-850f8e84234883567d77deec5584ae4dd4e7b7a1a2ff361119af1410be057b11R976-R989
&
https://github.com/facebookincubator/velox/pull/16058/changes#diff-850f8e84234883567d77deec5584ae4dd4e7b7a1a2ff361119af1410be057b11R1270-R1281

velox/functions/sparksql/tests/SparkCastExprTest.cpp
rui-mo: Is this PR primarily intended as a refactor for the unit tests? If so, could you please create an issue to track these unsupported cases instead of annotating them directly in the code?

infvg: Sure, let me do that

infvg: Raised this issue:
https://github.com/facebookincubator/velox/issues/16579

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -139,450 +144,385 @@ class SparkCastExprTest : public functions::test::CastBaseTest {
...
-  testCast<std::string, int32_t>(
-      "date", {"-1-1-1"}, {std::nullopt}, VARCHAR(), DATE());
-  testCast<std::string, int32_t>(
-      "date", {"-11-1-1"}, {std::nullopt}, VARCHAR(), DATE());
-  testCast<std::string, int32_t>(
-      "date", {"-111-1-1"}, {std::nullopt}, VARCHAR(), DATE());
-  testCast<std::string, int32_t>(
-      "date", {"- 1111-1-1"}, {std::nullopt}, VARCHAR(), DATE());
-}
+  void testStringToDate() {
```
rui-mo: nit: Add empty line between methods. Please fix this through.

rui-mo: Could you organize these helper functions in a specific sequence rather than randomly? For instance, sort them based on `fromType` in this order: bool → integer → floating-point → decimal → timestamp → complex types.

rui-mo:
Please also fix the format issues: https://github.com/facebookincubator/velox/actions/runs/22550365695/job/65375388680?pr=16058.

velox/functions/sparksql/tests/SparkCastExprTest.cpp
infvg: Reordered to bool -> integer -> float -> double -> decimal -> timestamp -> date -> string -> misc tests (overflow, truncate, corner cases)

infvg: Fixed

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -93,22 +243,96 @@ class SparkCastExprTest : public functions::test::CastBaseTest {
...
+             "-0.00000120"}));
+
+    testCast(
+        makeNullableFlatVector<int64_t>(
+            {100, std::nullopt, 1230, std::nullopt}, DECIMAL(10, 2)),
+        makeNullableFlatVector<std::string>(
+            {"1.00", std::nullopt, "12.30", std::nullopt}));
+  }
+
+  void testDoubleToTimestamp() {
```
rui-mo: nit: move this above L184?

infvg: Moved

PR #6586, opened 2023-09-15
Vendor Arrow Parquet Writer

pedroerp:
As discussed in [0], this PR is a first step at vendoring Arrow's Parquet writer into Velox. This moves all the most important Parquet related classes, but still uses Arrow core underneath. The next PRs will continue removing the arrow dependencies, and moving them to Velox types, vectors, and buffers.

[0] - https://github.com/facebookincubator/velox/discussions/6193

PR #16060, opened 2026-01-19
feat: Support iceberg compatible min max stats

PingLiuPing:
Add support for computing Iceberg-compatible min/max statistics in Parquet writer.
Iceberg requires specific encoding formats for column statistics that differ from the standard Parquet encoding.

**Decimal types:** Encoded in big-endian format instead of little-endian, as required by Iceberg's binary comparison semantics.
See [binary-single-value-serialization](https://iceberg.apache.org/spec/#binary-single-value-serialization).

**String Types (BYTE_ARRAY with UTF8):**
**Lower Bound:** Truncates to the specified number of Unicode code points (default 16)
Simple truncation ensures the result is ≤ the original string.
**Upper Bound:** Truncates and increments to create an exclusive upper bound
If the string has ≤ truncateTo code points, returns the exact value;
If longer, truncates to truncateTo code points and increments the last code point.
Increments from the last code point backwards until a valid increment is found.
Returns std::nullopt if no valid upper bound exists (e.g., all code points are at max value U+10FFFF)
Properly handles Unicode surrogate ranges (U+D800-U+DFFF) by skipping them during increment.
**Other Types (INT64 non-decimal, FLOAT, DOUBLE, etc.):**
Use standard Parquet plain encoding (no special handling needed)

Added two new methods to the Statistics interface:
IcebergLowerBoundInclusive(int32_t truncateTo): Returns an encoded value guaranteed to be ≤ the actual minimum value
IcebergUpperBoundExclusive(int32_t truncateTo): Returns an encoded value guaranteed to be ≥ the actual maximum value, or std::nullopt if no valid upper bound can be computed

Added comparison methods to enable merging statistics across row groups:
MaxGreaterThan(const Statistics& other): Compares maximum values between two statistics objects
MinLessThan(const Statistics& other): Compares minimum values between two statistics objects

Added new utility functions in velox/functions/lib/Utf8Utils.h:
truncate(std::string_view input, int32_t numCodePoints): Truncates a UTF-8 string to at most N code points.
roundUp(std::string_view input, int32_t numCodePoints): Rounds up a UTF-8 string to produce an exclusive upper bound.

### DIFF
--- a/velox/dwio/parquet/writer/arrow/Statistics.cpp
+++ b/velox/dwio/parquet/writer/arrow/Statistics.cpp
@@ -18,6 +18,8 @@
 
 #include "velox/dwio/parquet/writer/arrow/Statistics.h"
 
+#include "velox/functions/lib/string/StringImpl.h"
+
 #include <algorithm>
 #include <cmath>
 #include <cstring>
@@ -39,6 +41,9 @@
 #include "velox/dwio/parquet/writer/arrow/Exception.h"
 #include "velox/dwio/parquet/writer/arrow/Platform.h"
 #include "velox/dwio/parquet/writer/arrow/Schema.h"
+#include "velox/functions/lib/Utf8Utils.h"
+#include "velox/type/DecimalUtil.h"
+#include "velox/type/HugeInt.h"
 
 using arrow::default_memory_pool;
 using arrow::MemoryPool;
@@ -554,6 +559,17 @@ TypedComparatorImpl<false, ByteArrayType>::GetMinMax(
   return GetMinMaxBinaryHelper<false>(*this, values);
 }
 
+template <typename T>
+std::string encodeDecimalToBigEndian(T value) {
+  uint8_t buffer[sizeof(T)];
+  if constexpr (std::is_same_v<T, int64_t>) {
+    *reinterpret_cast<int64_t*>(buffer) = ::arrow::bit_util::ToBigEndian(value);
+  } else if constexpr (std::is_same_v<T, int128_t>) {
+    *reinterpret_cast<int128_t*>(buffer) = DecimalUtil::bigEndian(value);
+  }
+  return std::string(reinterpret_cast<const char*>(buffer), sizeof(T));
+}
+
 template <typename DType>
 class TypedStatisticsImpl : public TypedStatistics<DType> {
  public:
@@ -749,6 +765,56 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     return s;
   }
 
+  std::string IcebergLowerBoundInclusive(int32_t truncateTo) const override {
+    if constexpr (std::is_same_v<T, int64_t>) {
+      if (descr_->logical_type()->is_decimal()) {
+        return encodeDecimalToBigEndian(min_);
+      }
+    }
+    if constexpr (std::is_same_v<T, int128_t>) {
+      return encodeDecimalToBigEndian(min_);
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      const auto truncatedMin = functions::stringImpl::truncateUtf8(
+          std::string_view(min_), truncateTo);
+      std::string s;
+      this->PlainEncode(
+          ByteArray(
+              truncatedMin.size(),
+              reinterpret_cast<const uint8_t*>(truncatedMin.data())),
+          &s);
+      return s;
+    }
+    return EncodeMin();
+  }
+
+  std::optional<std::string> IcebergUpperBoundExclusive(
+      int32_t truncateTo) const override {
+    if constexpr (std::is_same_v<T, int64_t>) {
+      if (descr_->logical_type()->is_decimal()) {
+        return encodeDecimalToBigEndian(max_);
+      }
+    }
+    if constexpr (std::is_same_v<T, int128_t>) {
+      return encodeDecimalToBigEndian(max_);
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      const auto truncatedMax = functions::stringImpl::roundUpUtf8(
+          std::string_view(max_), truncateTo);
+      if (!truncatedMax.has_value()) {
+        return std::nullopt;
+      }
+      std::string s;
+      this->PlainEncode(
+          ByteArray(
+              truncatedMax->size(),
+              reinterpret_cast<const uint8_t*>(truncatedMax->data())),
+          &s);
+      return s;
+    }
+    return EncodeMax();
+  }
+
   EncodedStatistics Encode() override {
     EncodedStatistics s;
     if (HasMinMax()) {
@@ -774,6 +840,18 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
   }
   int64_t num_values() const override {
     return num_values_;
+  }
+
+  bool MaxGreaterThan(const Statistics& other) const override {
+    const auto* typedOther =
+        dynamic_cast<const TypedStatisticsImpl<DType>*>(&other);
+    return comparator_->Compare(max_, typedOther->max_) ? false : true;
+  }
+
+  bool MinLessThan(const Statistics& other) const override {
+    const auto* typedOther =
+        dynamic_cast<const TypedStatisticsImpl<DType>*>(&other);
+    return comparator_->Compare(min_, typedOther->min_) ? true : false;
   }
 
  private:
--- a/velox/dwio/parquet/writer/arrow/Statistics.h
+++ b/velox/dwio/parquet/writer/arrow/Statistics.h
@@ -281,6 +281,42 @@ class PARQUET_EXPORT Statistics {
   /// \brief Plain-encoded maximum value
   virtual std::string EncodeMax() const = 0;
 
+  /// \brief Encoded lower bound value compatible with Iceberg.
+  ///
+  /// Returns an encoded value guaranteed to be <= the actual minimum value.
+  /// For string types, truncates to at most \p truncateTo Unicode code points.
+  /// For decimal types, encodes the value in big-endian format as required by
+  /// Iceberg's single-value serialization specification.
+  /// For all other data types, uses the same plain encoding as Parquet
+  /// (returns the exact encoded minimum value).
+  ///
+  /// @param truncateTo Maximum number of Unicode code points for string types.
+  virtual std::string IcebergLowerBoundInclusive(int32_t truncateTo) const = 0;
+
+  /// \brief Encoded upper bound value compatible with Iceberg.
+  ///
+  /// Returns an encoded value guaranteed to be >= the actual maximum value.
+  /// For string types:
+  /// - If the maximum value has <= \p truncateTo Unicode code points, returns
+  ///   the exact encoded maximum value (inclusive upper bound).
+  /// - If the maximum value has > \p truncateTo Unicode code points, truncates
+  ///   to \p truncateTo code points and increments the last code point to
+  ///   produce an exclusive upper bound that is greater than the maximum value.
+  /// - Returns std::nullopt if no valid upper bound can be computed (e.g.,
+  ///   all code points in the truncated portion are at the maximum Unicode
+  ///   value U+10FFFF). This allows distinguishing between "upper bound is
+  ///   empty string" and "no valid upper bound exists".
+  /// For decimal types, encodes the value in big-endian format as required by
+  /// Iceberg's single-value serialization specification.
+  /// For all other data types, uses the same plain encoding as Parquet
+  /// (returns the exact encoded maximum value).
+  ///
+  /// @param truncateTo Maximum number of Unicode code points for string types.
+  /// @return Encoded upper bound value, or std::nullopt if no valid upper
+  ///         bound can be computed.
+  virtual std::optional<std::string> IcebergUpperBoundExclusive(
+      int32_t truncateTo) const = 0;
+
   /// \brief The finalized encoded form of the statistics for transport
   virtual EncodedStatistics Encode() = 0;
 
@@ -292,6 +328,14 @@ class PARQUET_EXPORT Statistics {
 
   /// \brief Check two Statistics for equality
   virtual bool Equals(const Statistics& other) const = 0;
+
+  /// \brief Return true if this object's max is greater than the other's max
+  /// \param[in] other the Statistics object to compare against
+  virtual bool MaxGreaterThan(const Statistics& other) const = 0;
+
+  /// \brief Return true if this object's min is less than the other's min
+  /// \param[in] other the Statistics object to compare against
+  virtual bool MinLessThan(const Statistics& other) const = 0;
 
  protected:
   static std::shared_ptr<Statistics> Make(
--- a/velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
+++ b/velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
@@ -1849,5 +1849,362 @@ TEST(TestEncodedStatistics, CopySafe) {
 }
 */
 
+namespace {
+
+constexpr int32_t kTruncLen = 16;
+
+template <typename ParquetType, typename T>
+std::shared_ptr<TypedStatistics<ParquetType>> makeStats(
+    const ColumnDescriptor* descr,
+    std::initializer_list<T> values) {
+  auto stats = MakeStatistics<ParquetType>(descr);
+  std::vector<T> v(values);
+  stats->Update(v.data(), v.size(), 0);
+  return stats;
+}
+
+std::shared_ptr<TypedStatistics<ByteArrayType>> makeStats(
+    const ColumnDescriptor* descr,
+    std::initializer_list<std::string> values) {
+  auto stats = MakeStatistics<ByteArrayType>(descr);
+  std::vector<std::string> strings(values);
+  std::vector<ByteArray> byteArrays;
+  byteArrays.reserve(strings.size());
+  for (const auto& s : strings) {
+    byteArrays.push_back(ByteArrayFromString(s));
+  }
+  stats->Update(byteArrays.data(), byteArrays.size(), 0);
+  return stats;
+}
+
+} // namespace
+
+TEST(IcebergStatistics, decimalMinMaxValue) {
+  const NodePtr node = PrimitiveNode::Make(
+      "decimal_col",
+      Repetition::REQUIRED,
+      LogicalType::Decimal(10, 2),
+      Type::INT64);
+  ColumnDescriptor descr(node, 0, 0);
+
+  auto stats =
+      makeStats<Int64Type, int64_t>(&descr, {12345, -67890, 100, 50000});
+
+  ASSERT_TRUE(stats->HasMinMax());
+  EXPECT_EQ(stats->min(), -67890);
+  EXPECT_EQ(stats->max(), 50000);
+
+  const auto lowerBound = stats->IcebergLowerBoundInclusive(kTruncLen);
+  const auto upperBound = stats->IcebergUpperBoundExclusive(kTruncLen);
+
+  ASSERT_TRUE(upperBound.has_value());
+
+  // Verify the encoding is big-endian.
+  int64_t decodedMin = ::arrow::bit_util::FromBigEndian(
+      *reinterpret_cast<const int64_t*>(lowerBound.data()));
+  int64_t decodedMax = ::arrow::bit_util::FromBigEndian(
+      *reinterpret_cast<const int64_t*>(upperBound->data()));
+
+  EXPECT_EQ(decodedMin, -67890);
+  EXPECT_EQ(decodedMax, 50000);
+}
+
+TEST(IcebergStatistics, nonDecimalBounds) {
+  const NodePtr node =
+      PrimitiveNode::Make("int_col", Repetition::REQUIRED, Type::INT64);
+  ColumnDescriptor descr(node, 0, 0);
+
+  auto stats = makeStats<Int64Type, int64_t>(&descr, {100, -200, 300, 50});
+
+  ASSERT_TRUE(stats->HasMinMax());
+
+  // For non-decimal INT64, IcebergLowerBound should equal EncodeMin (plain
+  // encoding).
+  EXPECT_EQ(stats->IcebergLowerBoundInclusive(kTruncLen), stats->EncodeMin());
+  const auto upperBound = stats->IcebergUpperBoundExclusive(kTruncLen);
+  ASSERT_TRUE(upperBound.has_value());
+  EXPECT_EQ(*upperBound, stats->EncodeMax());
+}
+
+TEST(IcebergStatistics, byteArrayBounds) {
+  NodePtr node = PrimitiveNode::Make(
+      "string_col",
+      Repetition::REQUIRED,
+      Type::BYTE_ARRAY,
+      ConvertedType::UTF8);
+  ColumnDescriptor descr(node, 0, 0);
+
+  auto stats = makeStats(
+      &descr,
+      {"AAAAAAAAAAAAAAAAAAAAAAAAA", "ZZZZZZZZZZZZZZZZZZZZZZZZZ", "Hello"});
+
+  ASSERT_TRUE(stats->HasMinMax());
+
+  // IcebergLowerBound should be truncated to 16 characters.
+  const auto lowerBound = stats->IcebergLowerBoundInclusive(kTruncLen);
+  EXPECT_EQ(lowerBound, "AAAAAAAAAAAAAAAA");
+
+  // IcebergUpperBound should be truncated and incremented.
+  const auto upperBound = stats->IcebergUpperBoundExclusive(kTruncLen);
+  ASSERT_TRUE(upperBound.has_value());
+  // 'Z' (0x5A) incremented becomes '[' (0x5B).
+  EXPECT_EQ(*upperBound, "ZZZZZZZZZZZZZZZ[");
+}
+
+TEST(IcebergStatistics, byteArrayBoundsNoTruncation) {
+  NodePtr node = PrimitiveNode::Make(
+      "string_col",
+      Repetition::REQUIRED,
+      Type::BYTE_ARRAY,
+      ConvertedType::UTF8);
+  ColumnDescriptor descr(node, 0, 0);
+
+  // Create strings shorter than 16 characters.
+  auto stats = makeStats(&descr, {"apple", "zebra", "banana"});
+
+  ASSERT_TRUE(stats->HasMinMax());
+
+  // For short strings, IcebergLowerBound/IcebergUpperBound should be the same
+  // as EncodeMin/EncodeMax.
+  EXPECT_EQ(stats->IcebergLowerBoundInclusive(kTruncLen), stats->EncodeMin());
+  const auto upperBound = stats->IcebergUpperBoundExclusive(kTruncLen);
+  ASSERT_TRUE(upperBound.has_value());
+  EXPECT_EQ(*upperBound, stats->EncodeMax());
+}
+
+TEST(IcebergStatistics, byteArrayBoundsUnicode) {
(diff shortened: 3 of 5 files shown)


PingLiuPing:
@mbasmanova Would you please have a look at this PR at your convenience?

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace facebook::velox::parquet {
+
+class UnicodeUtil {
+ public:
+  static std::string_view truncateStringMin(
```
mbasmanova: Would you document these APIs? 

Would it make sense to use std::string_view for the input? 

Would you check fbcode/velox/functions/lib/Utf8Utils.h in case it already has the required functionality?

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "velox/functions/lib/string/StringImpl.h"
+
+namespace facebook::velox::parquet {
+
+std::string_view UnicodeUtil::truncateStringMin(
+    const char* input,
+    int32_t inputLength,
+    int32_t numCodePoints) {
+  const auto truncatedLength = functions::stringImpl::cappedByteLength<false>(
+      StringView(input, inputLength), numCodePoints);
```
mbasmanova: Can we avoid using StringView here (it will cause a copy for short strings)? Can we use std::string_view instead?

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
```diff
@@ -0,0 +1,81 @@
...
+    return std::string(input, inputLength);
+  }
+
+  // Try to increment the last code point.
+  for (auto i = numCodePoints - 1; i >= 0; --i) {
+    const char* current = input;
+    int32_t currentCodePoint = 0;
+
+    // Find the i-th code point position.
+    while (current < input + truncatedLength && currentCodePoint < i) {
```
mbasmanova: This logic seems to repeat the logic inside cappedByteLength. Can we avoid scanning the string twice?

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
```diff
@@ -0,0 +1,81 @@
...
+      int32_t charLength;
+      utf8proc_codepoint(current, input + truncatedLength, charLength);
+      current += charLength;
+      currentCodePoint++;
+    }
+
+    if (current >= input + truncatedLength)
```
mbasmanova: Make sure to wrap body of 'if' in {}

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
```diff
@@ -0,0 +1,81 @@
...
+      continue;
+
+    int32_t charLength;
+    const auto codePoint =
+        utf8proc_codepoint(current, input + truncatedLength, charLength);
+    auto nextCodePoint = codePoint + 1;
+
+    // Check if the incremented code point is valid.
+    if (nextCodePoint != 0 && utf8proc_codepoint_valid(nextCodePoint)) {
```
mbasmanova: What if the code point is not valid?

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,12 @@ class PARQUET_EXPORT Statistics {
   /// \brief Plain-encoded maximum value
   virtual std::string EncodeMax() const = 0;
 
+  /// \brief Compatible minimum value with iceberg
```
mbasmanova: typos: Minimum value compatible with Iceberg.

ditto MaxValue

Would you expand the comment to explain what "compatible with Iceberg" means?

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -293,6 +299,24 @@ class PARQUET_EXPORT Statistics {
...
+      const std::shared_ptr<Statistics>& stats1,
+      const std::shared_ptr<Statistics>& stats2) {
+    return stats1->CompareMax(*stats2) ? stats1 : stats2;
+  }
+
+  static std::shared_ptr<Statistics> CompareAndGetMin(
+      const std::shared_ptr<Statistics>& stats1,
+      const std::shared_ptr<Statistics>& stats2) {
+    return stats1->CompareMin(*stats2) ? stats1 : stats2;
+  }
```
mbasmanova: Would you document these APIs? How do you plan to use them?

mbasmanova:
@majetideepak @yingsu00 Ying, Deepak, would one of you help review this Iceberg-specific PR?

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
PingLiuPing: Thanks for asking this. After revisit Iceberg Java implementation, on such scenario it should return std::nullopt.

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
PingLiuPing: Thanks. 
Added the document.
Yes, it can use std::string_view  for the input, because ByteArray supports std::string_view() operator.
I chekced `velox/functions/lib/Utf8Utils.h`  and it does not has the required function currently.

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
PingLiuPing: Yes, it can use std::string_view

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
PingLiuPing: thanks.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks. Added more coments.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: @mbasmanova 
These two helper APIs are used to compare min and max values across row groups and to derive the data-file–level min and max values. 
See this https://github.com/facebookincubator/velox/pull/16066/changes#diff-c852cba96188f835c605c7808c5ecdbddd9e9daa0f587e47ddba63f948042e20R109-R112

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.cpp
PingLiuPing: Thanks. Re-implemented this function to just scan the input string once.

velox/dwio/parquet/writer/arrow/Statistics.h
mbasmanova: Thank you for clarifying. 

It seems inefficient and confusing to return Statistics object from these APIs. Wouldn't it be better to return only min or max value?

Also these APIs return new shared_ptrs causing unnecessary increment / decrement of the reference count. Can they return a reference instead?

mbasmanova: > CompareAndGetMax

Why can't we just call these min and max? Do we need CompareAndGet prefix?

PingLiuPing: @mbasmanova  Thank you for the comments.
I refactored the code and I found `CompareAndGetMax` and `CompareAndGetMin` can be deleted. Just keep `CompareMax` and `CompareMin`. Using Compare prefix here is because this class already has min() and max() method.

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,25 @@ class PARQUET_EXPORT Statistics {
   /// \brief Plain-encoded maximum value
   virtual std::string EncodeMax() const = 0;
 
+  /// \brief Encoded minimum value compatible with Iceberg.
```
mbasmanova: @PingLiuPing Is this the code we copied from somewhere? It doesn't seem to follow Velox coding style. Should we wholesale update it to comply?

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,25 @@ class PARQUET_EXPORT Statistics {
...
+  ///
+  /// For string types, truncates to a configurable number of Unicode code
```
mbasmanova: > configurable number

How do we configure that?

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,25 @@ class PARQUET_EXPORT Statistics {
...
+  /// points (default 16). For decimal types, encodes the value in big-endian
```
mbasmanova: The comment explains about string and decimal types. What about other types?

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,25 @@ class PARQUET_EXPORT Statistics {
...
+  /// format as required by Iceberg's single-value serialization specification.
+  virtual std::string MinValue() const = 0;
+
+  /// \brief Encoded maximum value compatible with Iceberg.
+  ///
+  /// For string types, truncates to a configurable number of Unicode code
+  /// points (default 16) and increments the last code point to produce an
+  /// upper bound that is greater than all values with the same prefix.
```
mbasmanova: > upper bound that is greater than all values with the same prefix.

It looks like that string values are producing not the max value, but a minimum value that's strictly greater than max. However, for other types it produces just max. Is this intentional? Given these non-intuitive semantics, perhaps, rename the methods to MinIcebergValue and MaxIcebergValue.

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -293,6 +312,14 @@ class PARQUET_EXPORT Statistics {
   /// \brief Check two Statistics for equality
   virtual bool Equals(const Statistics& other) const = 0;
 
+  /// \brief Return true if this object's max is greater than the other's max
+  /// \param[in] other the Statistics object to compare against
+  virtual bool CompareMax(const Statistics& other) const = 0;
```
mbasmanova: Are these methods Iceberg-specific or not?

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Looking at the git history, yes, the code under writer/arrow folder are both imported from Arrow. See https://github.com/facebookincubator/velox/pull/6586.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: I left a TODO in the implementation of this method. I plan to add a configuration property for this.

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,25 @@ class PARQUET_EXPORT Statistics {
   /// \brief Plain-encoded maximum value
   virtual std::string EncodeMax() const = 0;
 
+  /// \brief Encoded minimum value compatible with Iceberg.
+  ///
+  /// For string types, truncates to a configurable number of Unicode code
+  /// points (default 16). For decimal types, encodes the value in big-endian
+  /// format as required by Iceberg's single-value serialization specification.
+  virtual std::string MinValue() const = 0;
```
mbasmanova: I wonder if these new getters are producing lower and upper bounds rather than min and max values. Perhaps, we could rename these for clarity and document that these values are guaranteed to be <= min and > max.

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
```diff
@@ -0,0 +1,63 @@
...
+  /// Truncates a UTF-8 encoded string to at most 'numCodePoints' Unicode code
+  /// points. Returns a string_view pointing to the truncated portion of the
+  /// input string. This is used for computing Iceberg lower bound statistics,
+  /// as the truncated string is guaranteed to be less than or equal to the
+  /// original string in lexicographic order.
+  ///
+  /// @param input The UTF-8 encoded input string.
+  /// @param numCodePoints Maximum number of Unicode code points to retain.
+  /// @return A string_view of the truncated string.
+  static std::string_view truncateStringMin(
```
mbasmanova: > truncateStringMin

This is just truncate, no? No need for StringMin suffix.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks. Other types should use same plain encoding with Parquet.
Updated comments.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks, make sense.

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
```diff
@@ -0,0 +1,63 @@
...
+  ///   increment code points from the last to the first. Returns immediately
+  ///   on the first successful increment.
+  /// - If no code point can be incremented (e.g., all are at max value
+  ///   U+10FFFF), returns std::nullopt.
+  ///
+  /// @param input The UTF-8 encoded input string.
+  /// @param numCodePoints Maximum number of Unicode code points to retain.
+  /// @return A new string containing the result, or std::nullopt if no valid
+  ///         upper bound can be found.
+  static std::optional<std::string> truncateStringMax(
```
mbasmanova: I believe this function doesn't really truncate the string, but rather returns a string that's >= input and has no more than numCodePoints characters. This seems like a rounding up behavior. Perhaps, we can use that name for clarity.

truncateStringMin -> truncate
truncateStringMax -> roundUp

Unicode processing is rather expensive, but checking if a string is all-ASCII is cheap. It is typically cheaper to check + take ascii path, then use unicode path. Would it make sense to do that here?

It might be better to move these utilities next to other string processing functions.

velox/dwio/parquet/writer/arrow/Statistics.h
mbasmanova: Do you know if Arrow Parquet Writer supports the functionality you are adding? 

@pedroerp @majetideepak Should we re-import Parquet Writer? If not, would it make sense to wholesale update it to match Velox coding style?

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -749,6 +763,58 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
...
+    if constexpr (std::is_same_v<T, int64_t>) {
+      if (descr_->logical_type()->is_decimal()) {
+        return encodeDecimalToBigEndian(min_);
+      }
+    }
+    if constexpr (std::is_same_v<T, int128_t>) {
+      return encodeDecimalToBigEndian(min_);
+    }
+    if constexpr (std::is_same_v<T, ByteArray>) {
+      // TODO: 16 is default value. See DEFAULT_WRITE_METRICS_MODE_DEFAULT in
```
mbasmanova: I assume we wouldn't want to access config here, but rather take the limit as an argument no?

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks. These two APIs are currently used only by Iceberg. However, the comparison is based on the actual `max_` and `min_` values. For example, for string types, `max_` and `min_` are not truncated.

Therefore, these methods are not Iceberg-specific and can be used more generally to compare statistics.

mbasmanova: Got it.

The semantics of CompareMax appear to be >, while CompareMin is <. This is non-intuitive. Would it make sense to rename these methods for clarity?

CompareMax -> MaxGreaterThen(other)
CompareMin -> MinLessThen(other)

velox/dwio/parquet/writer/arrow/Statistics.cpp
```diff
@@ -776,6 +842,18 @@ class TypedStatisticsImpl : public TypedStatistics<DType> {
     return num_values_;
   }
 
+  bool CompareMax(const Statistics& other) const override {
```
mbasmanova: Let's add tests for these methods.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
```diff
@@ -1849,5 +1849,195 @@ TEST(TestEncodedStatistics, CopySafe) {
...
+      "decimal_col",
+      Repetition::REQUIRED,
+      LogicalType::Decimal(10, 2),
+      Type::INT64);
+  ColumnDescriptor descr(node, 0, 0);
+
+  auto stats = MakeStatistics<Int64Type>(&descr);
+
+  std::vector<int64_t> values = {12345, -67890, 100, 50000};
+  stats->Update(values.data(), values.size(), 0);
```
mbasmanova: This logic repeats a lot. Perhaps, add a template helper that takes a list of values and returns a Statistics object.

auto stats = makeStats<int64_t>({12345, -67890, 100, 50000});

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks, yes you raise a good point. These aren't exact min/max values, they are lower and upper bounds.
Renamed to `IcebergLowerBound` and `IcebergUpperBound` and updated comments.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks. Makes sense.

velox/dwio/parquet/writer/arrow/Statistics.h
mbasmanova: > Renamed to IcebergLowerBound and IcebergUpperBound

Thanks. Perhaps, we can clarify further that lower is inclusive, but upper is exclusive?

IcebergLowerBoundInclusive, IcebergUpperBoundExclusive

PingLiuPing: Thanks.

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
PingLiuPing: Thanks

velox/dwio/parquet/writer/arrow/util/UnicodeUtil.h
PingLiuPing: Thanks.
Renamed the function.
Refactored the code and adding a fast-path for ascii.
Moved to velox/functions/lib/UnicodeUtil.h

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Thanks. added

velox/dwio/parquet/writer/arrow/Statistics.cpp
PingLiuPing: Yes, makes sense, and the call site is in Iceberg connector layer, make more sense to retrieve the config property there.

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,36 @@ class PARQUET_EXPORT Statistics {
...
+  /// Iceberg's single-value serialization specification.
+  /// For all other data types, uses the same plain encoding as Parquet
+  /// (returns the exact encoded minimum value).
+  ///
+  /// @param truncateTo Maximum number of Unicode code points for string types.
+  virtual std::string IcebergLowerBoundInclusive(int32_t truncateTo) const = 0;
+
+  /// \brief Encoded upper bound value compatible with Iceberg.
+  ///
+  /// Returns an encoded value guaranteed to be > the actual maximum value
```
mbasmanova: > guaranteed to be > the actual maximum value

This doesn't seem to match the implementation. It looks like if 'max' is <= 16 chars, then it is returned as is. The rounding happens only if 'max' > 16 chars.

velox/dwio/parquet/writer/arrow/Statistics.h
```diff
@@ -281,6 +281,36 @@ class PARQUET_EXPORT Statistics {
...
+  /// greater than the maximum value. Returns an empty string if no valid upper
+  /// bound can be computed (e.g., all code points are at the maximum Unicode
+  /// value U+10FFFF).
+  /// For decimal types, encodes the value in big-endian format as required by
+  /// Iceberg's single-value serialization specification.
+  /// For all other data types, uses the same plain encoding as Parquet
+  /// (returns the exact encoded maximum value).
+  ///
+  /// @param truncateTo Maximum number of Unicode code points for string types.
+  virtual std::string IcebergUpperBoundExclusive(int32_t truncateTo) const = 0;
```
mbasmanova: If min = max = <empty string> then IcebergUpperBoundExclusive will return <empty string>, which means "unbounded", which is incorrect. It seems that we may need to return std::optional<std::string> to allow to distinguish between unbounded and max == <empty string>.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks, refined the comments.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: Thanks, yes you are right. 
Updated the code and added unit test.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
```diff
@@ -1849,5 +1849,380 @@ TEST(TestEncodedStatistics, CopySafe) {
...
+  ColumnDescriptor descr(node, 0, 0);
+
+  auto stats =
+      makeStats<Int64Type, int64_t>(&descr, {12345, -67890, 100, 50000});
+
+  ASSERT_TRUE(stats->HasMinMax());
+  EXPECT_EQ(stats->min(), -67890);
+  EXPECT_EQ(stats->max(), 50000);
+
+  std::string lowerBound =
```
mbasmanova: consistency: auto (to match upperBound)

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
```diff
@@ -1849,5 +1849,380 @@ TEST(TestEncodedStatistics, CopySafe) {
...
+  auto stats = makeStats(
+      &descr,
+      {"AAAAAAAAAAAAAAAAAAAAAAAAA", "ZZZZZZZZZZZZZZZZZZZZZZZZZ", "Hello"});
+
+  ASSERT_TRUE(stats->HasMinMax());
+
+  // IcebergLowerBound should be truncated to 16 characters.
+  std::string lowerBound =
+      stats->IcebergLowerBoundInclusive(kDefaultTruncateLength);
+  EXPECT_EQ(lowerBound.size(), 16);
```
mbasmanova: This check seems redundant.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
```diff
@@ -1849,5 +1849,380 @@ TEST(TestEncodedStatistics, CopySafe) {
...
+  EXPECT_EQ(lowerBound, "AAAAAAAAAAAAAAAA");
+
+  // IcebergUpperBound should be truncated and incremented.
+  auto upperBound = stats->IcebergUpperBoundExclusive(kDefaultTruncateLength);
+  ASSERT_TRUE(upperBound.has_value());
+  EXPECT_EQ(upperBound->size(), 16);
```
mbasmanova: This check seems redundant.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
PingLiuPing: @mbasmanova  Thanks, updated.

velox/dwio/parquet/writer/arrow/Statistics.h
PingLiuPing: @mbasmanova I looked into the Arrow [C++ Parquet code](https://github.com/apache/arrow/tree/main/cpp/src/parquet) and their GitHub issues, and I couldn’t find support for this use case. While Arrow C++ can write Parquet files, I couldn’t find equivalent support for collecting Iceberg-compatible statistics like what we do here, or extension points to customize stats collection.

velox/functions/lib/UnicodeUtil.h
```diff
@@ -0,0 +1,63 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace facebook::velox::parquet {
+
+class UnicodeUtil {
```
Yuhta: Can you merge these functions to `Utf8Utils.h`?  Or move then to dwio/parquet directory if they are only used in parquet

PingLiuPing: @Yuhta  Thanks for the comment. I merged them into Utf8utils.h. Originally this file is moved from velox/dwio/parquet/arrow/util to velox/functions/lib/ per review comments.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/writer/arrow/tests/StatisticsTest.cpp
PingLiuPing: Thanks.

PingLiuPing:
> @PingLiuPing Thanks. Please, update PR description to match latest code.

@mbasmanova  Updated the PR description. Thank you.

velox/functions/lib/Utf8Utils.cpp
```diff
@@ -14,10 +14,20 @@
  * limitations under the License.
  */
 #include "velox/functions/lib/Utf8Utils.h"
+
 #include "velox/common/base/Exceptions.h"
 #include "velox/external/utf8proc/utf8procImpl.h"
+#include "velox/functions/lib/string/StringImpl.h"
```
Yuhta: It turns out a circular dependency between StringImpl and Util8Utils.  Let's move the new functions to `StringImpl.h` instead and name them with utf8 in the name so it's clear what they are for.

PingLiuPing: @Yuhta Thanks. Updated.

velox/functions/lib/string/StringImpl.h
```diff
@@ -814,4 +814,147 @@ FOLLY_ALWAYS_INLINE bool initcap(TOutString& output, const TInString& input) {
   }
 }
 
+namespace {
```
Yuhta: Anonymous namespace should not be used in header files, can you change it to something like `detail`?

PingLiuPing: Thanks @Yuhta , updated.

PR #16069, opened 2026-01-20
feat: Add Python code generation for RemoteFunction.thrift

cwalv:
Summary:
This enables Python thrift types for RemoteFunctionPage and related structs.
Required for Python sink support in XStream pipelines, since CustomSink.thrift
depends on RemoteFunction.thrift for the RemoteFunctionPage type used in
SinkWriteRequest.

Differential Revision: D91050192

### DIFF
--- a/velox/functions/remote/if/RemoteFunction.thrift
+++ b/velox/functions/remote/if/RemoteFunction.thrift
@@ -18,6 +18,7 @@ package "facebook.com/velox/functions"
 
 namespace cpp2 facebook.velox.functions.remote
 namespace java.swift com.facebook.spark.remotefunctionserver.api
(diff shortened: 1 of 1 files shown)


kunigami:
for some reason I don't see the BUCK changes on the files changed. do we need to re-export?

cwalv:
> for some reason I don't see the BUCK changes on the files changed. do we need to re-export?

It looks like the BUCK files aren't in public repo: https://github.com/facebookincubator/velox/blob/e78673aafc3da3012dca1a303a534b36303d07aa/velox/functions/remote/if/RemoteFunction.thrift

kunigami:
interesting! I hadn't noticed that before.

PR #16074, opened 2026-01-20
refactor(trace): Make tracing extensible

pedroerp:
Summary:
Reorganizing the tracing code to make it extensible for different
types of trace writers.

* Make trace writer creation extensible. 
* Create base class and move file dump tracer configs to specialized class.
* Untangle trace setup code from operators/tasks into tracing specialized classes.

Differential Revision: D91059289


velox/exec/trace/TraceConfig.h
```diff
@@ -27,34 +27,32 @@ class Operator;
 
 namespace facebook::velox::exec::trace {
 
-/// The callback used to update and aggregate the trace bytes of a query. If the
-/// query trace limit is set, the callback return true if the aggregate traced
-/// bytes exceed the set limit otherwise return false.
-using UpdateAndCheckTraceLimitCB = std::function<void(uint64_t)>;
+class TraceConfig {
```
xiaoxmeng, line 30: Shall we call this TraceBuilder and it seems that its function has changed with the refactor?

pedroerp: +1. I was thinking about calling it `TraceCtx` instead. Let me rename it in a separate PR since this require chances in a bunch of callsites.

PR #16077, opened 2026-01-20
feat: Implement file rollover at configurable size limit

prashantgolash:
Summary:
**Context**:

Large file causes warm storage local throttling issues due to hot blocks (It triggers FFRs). In such cases, the footer block is accessed by lot of workers, given most of them are assigned split from the same file and have to read footer to do further processing. More analysis / details in this [doc](https://docs.google.com/document/d/1JMeStVrE1Kqx2JA-mQwmwgr8rt9I3ZfR4f8IPC8CGMI/edit?tab=t.0#bookmark=kix.mdel4wv8eoh5). Also few relevant tasks: T249501068 and T241325745

**Fix:**
In this proposal I am introducing a default config / session property which can be used to rotate writer and write smaller files when they exceed the limit. From our [earlier analysis](https://docs.google.com/document/d/1JMeStVrE1Kqx2JA-mQwmwgr8rt9I3ZfR4f8IPC8CGMI/edit?tab=t.0#bookmark=id.vhojfp8w1lto) p99.5 <= 1GB and in these cases we haven't see throttling issues.

Differential Revision: D89820483

### DIFF
--- a/velox/connectors/hive/HiveConfig.cpp
+++ b/velox/connectors/hive/HiveConfig.cpp
@@ -211,6 +211,14 @@ uint64_t HiveConfig::sortWriterFinishTimeSliceLimitMs(
       kSortWriterFinishTimeSliceLimitMsSession,
       config_->get<uint64_t>(kSortWriterFinishTimeSliceLimitMs, 5'000));
 }
+uint64_t HiveConfig::maxTargetFileSizeBytes(
+    const config::ConfigBase* session) const {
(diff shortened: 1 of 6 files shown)


prashantgolash:
> Reviewed in the Meta internal tool. Looks good.
> 
> @prashantgolash , please ensure that two check failures are not related to the change.

Thanks Sergey. I verified that test failures are not related to my change and it seems they are also failing on other pulls

PR #16090, opened 2026-01-22
 test case velox_exec_test spillPartitionBitsOverlap to debug seed-dependent behaviour

mpurbay-arm:
- Add INJECT_BASE_SEED seeding in VectorFuzzer (VectorFuzzer.h:170–179) so we can reproduce the same fuzzer output per seed on a given platform.
  - Log fuzzer encoding decisions in VectorFuzzer::fuzz (ManuVF.fuzzBranch type=... tookConstant=0/1) to see when BIGINT keys take the constant path.
  - Log per-batch join output in HashProbe::getOutputInternal (ManuHP.listJoinResults / ManuHP.afterFilter) to track how many rows each batch contributes.
  - Log final result rows and batch count in AssertQueryBuilder (ManuAQB resultRowCount rows=... batches=...) to correlate seeds with total join cardinality (3–6K vs 3.3M–4.4M rows).

  This commit only adds Velox-side, test-only instrumentation; it does not change hash join or spill semantics. In a separate Folly change we switched to a consistent RNG so behaviour is now aligned across ARM/AMD; this patch
  helps analyse the remaining seed-dependent behaviour of this benchmark.

### DIFF
--- a/velox/exec/HashProbe.cpp
+++ b/velox/exec/HashProbe.cpp
@@ -1198,6 +1198,8 @@ RowVectorPtr HashProbe::getOutputInternal(bool toSpillOutput) {
           folly::Range(mapping.data(), outputBatchSize),
           folly::Range(outputTableRows, outputBatchSize),
           operatorCtx_->driverCtx()->queryConfig().preferredOutputBatchBytes());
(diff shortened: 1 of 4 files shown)


mpurbay-arm:
closing it for now.

PR #16091, opened 2026-01-22
test: Delete an iceberg redundant test case

PingLiuPing:
In this test case `TEST_F(HiveIcebergTest, partitionedRead)`, the way to construct and write the partitioned data are wrong. The data is not partitioned actually.
The verification of writing partitioned iceberg table and reading from partitioned iceberg table has been added in TransformE2ETest.cpp. And in this file it verifies all iceberg partition transforms.

### DIFF
--- a/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
+++ b/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
@@ -757,63 +757,6 @@ TEST_F(HiveIcebergTest, positionalDeletesMultipleSplits) {
   assertMultipleSplits({1000, 9000, 20000}, 1, 0, 20000, 3);
 }
 
-TEST_F(HiveIcebergTest, partitionedRead) {
-  RowTypePtr rowType{ROW({"c0", "ds"}, {BIGINT(), DateType::get()})};
-  std::unordered_map<std::string, std::optional<std::string>> partitionKeys;
-  // Iceberg API sets partition values for dates to daysSinceEpoch, so
-  // in velox, we do not need to convert it to days.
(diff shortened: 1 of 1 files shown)


PingLiuPing:
@yingsu00 Could you please have a look at this small PR? 
I cannot find you in the reviewers list.

mbasmanova:
> The way to construct and write the partitioned data are wrong. The data is not partitioned actually.

Curious, what is wrong in this test? Did we also miss verification that the result is partitioned?

PingLiuPing:
> > The way to construct and write the partitioned data are wrong. The data is not partitioned actually.
> 
> Curious, what is wrong in this test? Did we also miss verification that the result is partitioned?

@mbasmanova The purpose of this test should be to verify reading a partitioned Iceberg table. However, when this test was originally written, the Iceberg connector did not support writing or partitioning. As a workaround, the test mocked partitioning by explicitly generating separate data files and manually constructing splits with partition keys, implicitly assuming an identity transform.
The correct approach is to use IcebergDataSink to write data. For a partitioned table, we should build a PartitionSpec with the appropriate partition fields and transforms, then pass it to IcebergDataSink.

Yes, we should verify the result is partitioned. And it should be done by verifying the data file directory name. 

In TransformE2ETest.cpp, it has verified both writing and reading path for partitioned iceberg table.
For example 
https://github.com/facebookincubator/velox/blob/12f04b45f1c7e4e40fd651501c2aed8a8573b266/velox/connectors/hive/iceberg/tests/TransformE2ETest.cpp#L291-L320

mbasmanova:
@PingLiuPing Thank you for explaining.

PR #16092, opened 2026-01-22
test: Validate Spark string-to-date cast

malinjawi:
Add test coverage and mark string-to-date cast as ANSI compliant.

The implementation already exists in SparkCastHooks. This PR adds:
- Test validating ANSI mode behavior (exception vs null)
- Documentation marking cast as ANSI compliant

Verified against Spark behavior:
- ANSI ON: Invalid dates throw exception
- ANSI OFF: Invalid dates return NULL

Issue: https://github.com/apache/incubator-gluten/issues/10134.

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -284,6 +284,14 @@ class QueryConfig {
   /// output rows.
   static constexpr const char* kMaxOutputBatchRows = "max_output_batch_rows";
 
+  /// Initial output batch size in rows for MergeJoin operator. When non-zero,
+  /// the batch size starts at this value and is dynamically adjusted based on
+  /// the average row size of previous output batches. When zero (default),
+  /// dynamic adjustment is disabled and the batch size is fixed at
+  /// preferredOutputBatchRows.
+  static constexpr const char* kMergeJoinOutputBatchStartSize =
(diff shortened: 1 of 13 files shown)


velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -233,6 +233,34 @@ TEST_F(SparkCastExprTest, invalidDate) {
...
+       "2015-03-18 123", /*Trailing content*/
+       "  1970-01-01  ", /*Whitespace*/
+       "2015", /*Year only*/
+       "2015-03", /*Year-month*/
+       "1970-1-1"}, /*Single digit month/day*/
+      {0, 16512, 16512, 16512, 0, 16436, 16495, 0},
+      VARCHAR(),
+      DATE());
+
+  // Invalid dates return null.
```
rui-mo: Do we have tests covering the error path? Thanks!

malinjawi: @rui-mo I've added comprehensive tests covering the error path with ANSI mode ON. Used scope guard pattern to ensure ANSI mode is properly restored after the test. Ready for re-review. Thanks!

Issue #16094, opened 2026-01-22
run-clang-tidy.py fails in Linux release with adapters build


kgpai:
### Problem description

[See this run](https://github.com/facebookincubator/velox/actions/runs/21239265565/job/61113490891) . It seems clang-tidy is failing due to not getting the commit : 

```
usage: run-clang-tidy.py [-h] [--commit COMMIT] [--fix FIX] [-p P]
                         FILES [FILES ...]
run-clang-tidy.py: error: argument --commit: expected one argument
```

### System information

N/A

### CMake log

```bash

```

kgpai:
cc: @czentgr  @assignUser

PR #16098, opened 2026-01-22
fix(ci): Trigger to run clang-tidy

czentgr:
When the clang-tidy step was moved to the adapters build, the check to whether or not run clang-tidy determined the get-changes job was not ported. This resulted in running the clang-tidy job on the push trigger which has no commit range resulting in failure.

Addresses: https://github.com/facebookincubator/velox/issues/16094

### DIFF
--- a/.github/workflows/linux-build-base.yml
+++ b/.github/workflows/linux-build-base.yml
@@ -198,7 +198,7 @@ jobs:
       # need to run clang-tidy for that.
       # Let's also run this as last step so that if skipped it doesn't affect subsequent steps.
       - name: Install and run clang-tidy
-        if: ${{ ! inputs.use-clang }}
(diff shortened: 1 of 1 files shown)


velox/connectors/hive/storage_adapters/s3fs/tests/S3MultipleEndpointsTest.cpp
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
```
kgpai, line 16: nit: some previous change ?

czentgr: This was a change to trigger a change in a cpp file which should turn run-clang-tidy to true. I can see it ran in the CI. So I'm removing this change now.

PR #16099, opened 2026-01-23
feat: Add resolveVectorFunctionWithMetadataWithCoercions

marxhxxx:
Summary:
Add a new function `resolveVectorFunctionWithMetadataWithCoercions` that combines the functionality of `resolveVectorFunctionWithMetadata` and `resolveVectorFunctionWithCoercions`. This function returns both the return type and metadata for a vector function while also supporting type coercions when no signature matches the argument types exactly.

This is useful for callers that need function metadata (e.g., deterministic flag) in addition to the resolved return type when working with implicit type coercions.

Additionally, refactored `resolveVectorFunctionWithCoercions` to reuse `resolveVectorFunctionWithMetadataWithCoercions` internally, reducing code duplication.

Differential Revision: D91273969

### DIFF
--- a/velox/functions/tests/FunctionRegistryTest.cpp
+++ b/velox/functions/tests/FunctionRegistryTest.cpp
@@ -282,6 +282,60 @@ class FunctionRegistryTest : public testing::Test {
 
     return builder.build();
   }
+
+  void testVectorFunctionWithMetadataCoercions(
+      const std::string& name,
+      const std::vector<TypePtr>& argTypes,
+      const TypePtr& expectedReturnType,
+      const std::vector<TypePtr>& expectedCoercions,
+      bool expectedDeterministic) {
+    std::vector<TypePtr> coercions;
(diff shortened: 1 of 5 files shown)


velox/functions/tests/FunctionRegistryTest.cpp
```diff
@@ -1050,5 +1050,113 @@ TEST_F(FunctionRegistryTest, ipPrefixRegistration) {
...
+    EXPECT_EQ(coercions[1], nullptr);
+  }
+
+  // Test that metadata is correctly returned for non-deterministic function.
+  // This verifies we're returning actual function metadata, not defaults.
+  {
+    std::vector<TypePtr> coercions;
+    auto result = resolveVectorFunctionWithMetadataWithCoercions(
+        "bar_nondet", {TINYINT(), TINYINT()}, coercions);
+    EXPECT_TRUE(result.has_value());
```
mbasmanova: Would be nice to reduce copy-paste using helper functions similar to testCoercions.

PR #16101, opened 2026-01-23
docs: Add documentation for array_top_n function

allenshen13:
### Add documentation for array_top_n function with both signatures:

- array_top_n(array(T), n) - returns top n elements by natural ordering
- array_top_n(array(T), n, function(T, U)) - returns top n elements sorted by transform function
- Includes examples covering common use cases: basic usage, n larger than array size, null handling, and transform-based sorting (negation, absolute value, string length).

Depends on https://github.com/facebookincubator/velox/pull/16048

### DIFF
--- a/velox/exec/fuzzer/CMakeLists.txt
+++ b/velox/exec/fuzzer/CMakeLists.txt
@@ -12,6 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Generate Thrift library for LocalRunnerService early since velox_fuzzer_util
+# depends on it when VELOX_ENABLE_REMOTE_FUNCTIONS is enabled.
+if(VELOX_ENABLE_REMOTE_FUNCTIONS)
+  include(FBThriftCppLibrary)
+  add_fbthrift_cpp_library(
+    local_runner_service_thrift
+    if/LocalRunnerService.thrift
+    SERVICES
+    LocalRunnerService
+  )
+  target_compile_options(local_runner_service_thrift PRIVATE -Wno-error=deprecated-declarations)
+endif()
+
 add_library(
   velox_fuzzer_util
   ReferenceQueryRunner.cpp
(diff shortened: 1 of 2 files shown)


velox/docs/functions/presto/array.rst
```diff
@@ -252,6 +252,34 @@ Array Functions
...
+    Returns the top ``n`` elements of the array in descending order according to the natural ordering of the elements.
+    If ``n`` is larger than the size of the array, returns all elements sorted in descending order.
+    If ``n`` is zero, returns an empty array. Null elements are placed at the end of the result. ::
+
+        SELECT array_top_n(ARRAY [1, 100, 2, 5, 3], 3); -- [100, 5, 3]
+        SELECT array_top_n(ARRAY [1, 100], 5); -- [100, 1]
+        SELECT array_top_n(ARRAY [1, NULL, 2], 3); -- [2, 1, NULL]
+        SELECT array_top_n(ARRAY [1, 2, 3], 0); -- []
+
+.. function:: array_top_n(array(T), n, function(T, U)) -> array(T)
```
mbasmanova, line 266: I don't believe we have this function in Velox, do we?

CC: @Yuhta

velox/docs/functions/presto/array.rst
```diff
@@ -252,6 +252,34 @@ Array Functions
     Returns the sum of all non-null elements of the array. If there is no non-null elements, returns 0. The behaviour is similar to aggregation function sum().
     T must be coercible to double. Returns bigint if T is coercible to bigint. Otherwise, returns double.
 
+.. function:: array_top_n(array(T), n) -> array(T)
+
+    Returns the top ``n`` elements of the array in descending order according to the natural ordering of the elements.
```
mbasmanova, line 257: @allenshen13 Would you clarify how ties are resolves? Is this a stable sort or not?

velox/docs/functions/presto/array.rst
Yuhta: Yeah that's right, let's put this PR on hold

mbasmanova: This PR should be limited to adding documentation for an existing function, the one that doesn't take a lambda argument.

velox/docs/functions/presto/array.rst
allenshen13: With array_top_n you can have primitives, where equal values are completely indistinguishable (can't tell which position they come from). Even the complex types which can be compared, the full values are compared. This means that there is no way to have equal but distinguishable elements for the user to see. However, the underlying implementation uses std::priority_queue which doesn't guarentee any order among equals. The internal heap operations don't guarantee preservation of insertion order for equal elements. 

In this case is it even worth mentioning stability if the user can't observe it? Is there some complex type that could be considered equal and distinguishable?

mbasmanova: @allenshen13 Good point. If there is no observable behavior, no need to mention it.

PR #16104, opened 2026-01-23
fix(cudf): Fix companion function kind added result type

jinchengchenghh:
After this PR https://github.com/facebookincubator/velox/pull/11999/files#diff-58637420124ffe154b91ddfe97854192bc53cca18543a4046db8ebcb04edb7fd, the companion function name is updated from count_merge_extract to count_merge_extract_double, so update cudf logic as well.

### DIFF
--- a/velox/experimental/cudf/exec/CudfHashAggregation.cpp
+++ b/velox/experimental/cudf/exec/CudfHashAggregation.cpp
@@ -468,8 +468,8 @@ core::AggregationNode::Step getCompanionStep(
     return core::AggregationNode::Step::kPartial;
   }
 
-  // The format is count_merge_extract_BIGINT.
-  if (kind.find("_merge_extract_") != std::string::npos) {
+  // The format is count_merge_extract_BIGINT or count_merge_extract.
+  if (kind.find("_merge_extract") != std::string::npos) {
     return core::AggregationNode::Step::kFinal;
   }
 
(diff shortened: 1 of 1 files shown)


velox/experimental/cudf/exec/CudfHashAggregation.cpp
```diff
@@ -461,24 +455,40 @@ static const std::unordered_map<std::string, core::AggregationNode::Step>
...
+  if (kind.ends_with("_merge")) {
+    return core::AggregationNode::Step::kIntermediate;
+  }
+
+  if (kind.ends_with("_partial")) {
+    return core::AggregationNode::Step::kPartial;
+  }
+
+  // The format is count_merge_extract_BIGINT.
+  if (kind.find("_merge_extract_") != std::string::npos) {
```
rui-mo: This may not apply for all aggregate functions. Velox only uses a suffix to differentiate aggregate functions when the intermediate types are the same. As a result, some `merge_extract` functions may have no suffix and still end with `merge_extract`. To keep things simple, you might could replace "\_merge_extract\_" with "_merge_extract".

https://github.com/facebookincubator/velox/blob/cfddf73e5473dd3a7ee412a538e68bb472a9688d/velox/exec/AggregateCompanionAdapter.cpp#L422-L429

jinchengchenghh: Thanks for your advice, applied.

PR #16106, opened 2026-01-23
fix(build): Remove velox_dwio_parquet_field_id from Iceberg link library

PingLiuPing:
Resolve prestissimo building error. When parquet is disabled, `velox_dwio_parquet_field_id` can not be found.

See https://github.com/prestodb/presto/actions/runs/21270527640/job/61219536504?pr=26982.

### DIFF
--- a/velox/connectors/hive/iceberg/CMakeLists.txt
+++ b/velox/connectors/hive/iceberg/CMakeLists.txt
@@ -29,10 +29,13 @@ velox_add_library(
 velox_link_libraries(
   velox_hive_iceberg_splitreader
   velox_connector
-  velox_dwio_parquet_field_id
   velox_functions_iceberg
   Folly::folly
 )
+
+if(VELOX_ENABLE_PARQUET)
+  velox_link_libraries(velox_hive_iceberg_splitreader velox_dwio_parquet_field_id)
+endif()
 
 if(${VELOX_BUILD_TESTING})
   add_subdirectory(tests)
--- a/velox/connectors/hive/iceberg/IcebergColumnHandle.cpp
+++ b/velox/connectors/hive/iceberg/IcebergColumnHandle.cpp
@@ -20,7 +20,7 @@
 
 #include "velox/connectors/hive/TableHandle.h"
 #include "velox/connectors/hive/iceberg/IcebergColumnHandle.h"
-#include "velox/dwio/parquet/writer/ParquetFieldId.h"
+#include "velox/dwio/parquet/ParquetFieldId.h"
 #include "velox/type/Subfield.h"
 #include "velox/type/Type.h"
 
--- a/velox/connectors/hive/iceberg/IcebergColumnHandle.h
+++ b/velox/connectors/hive/iceberg/IcebergColumnHandle.h
@@ -19,7 +19,7 @@
 #include <vector>
 
 #include "velox/connectors/hive/TableHandle.h"
-#include "velox/dwio/parquet/writer/ParquetFieldId.h"
+#include "velox/dwio/parquet/ParquetFieldId.h"
 #include "velox/type/Subfield.h"
 #include "velox/type/Type.h"
 
--- a/velox/dwio/parquet/CMakeLists.txt
+++ b/velox/dwio/parquet/CMakeLists.txt
@@ -25,6 +25,7 @@ endif()
 
 velox_add_library(velox_dwio_parquet_reader RegisterParquetReader.cpp)
 velox_add_library(velox_dwio_parquet_writer RegisterParquetWriter.cpp)
+velox_add_library(velox_dwio_parquet_field_id INTERFACE ParquetFieldId.h)
 
 if(VELOX_ENABLE_PARQUET)
   velox_link_libraries(velox_dwio_parquet_reader velox_dwio_native_parquet_reader xsimd)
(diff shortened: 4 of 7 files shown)


PingLiuPing:
I tried in Presto and it works so far, see https://github.com/prestodb/presto/actions/runs/21283121234/job/61257698164?pr=26908

velox/connectors/hive/iceberg/CMakeLists.txt
```diff
@@ -29,7 +29,6 @@ velox_add_library(
 velox_link_libraries(
   velox_hive_iceberg_splitreader
   velox_connector
-  velox_dwio_parquet_field_id
```
kgpai, line 32: Does iceberg not need parquet's field id ? As per this I was under impression that it does :https://github.com/facebookincubator/velox/pull/16042/files

kevinwilfong:
@PingLiuPing how does https://github.com/prestodb/presto/actions/runs/21283121234/job/61257698164?pr=26908 demonstrate the fix?

https://github.com/prestodb/presto/pull/26982 hasn't been merged, and https://github.com/prestodb/presto/pull/26908 doesn't appear to contain the change to advance Velox

More generally, while I appreciate that Velox builds without that dependency, I'm not sure how

IcebergColumnHandle now depends on ParquetFieldId unconditionally
https://github.com/facebookincubator/velox/blob/e0230a0669dbaff16262121963c797f0f5b7d7a9/velox/connectors/hive/iceberg/IcebergColumnHandle.h#L22

Is the test not triggering hitting that code path, so it's getting optimized away? What happens if we try to use this code and Parquet is disabled in Velox?

Would we be better off adding something like

```
if(VELOX_ENABLE_PARQUET)
  velox_link_libraries(velox_hive_iceberg_splitreader velox_dwio_parquet_field_id)
endif()
```

or should we disable Iceberg if Parquet is disabled given the hard dependency?

PingLiuPing:
@kevinwilfong  Thanks for the comments.

I advanced velox and include the changes in this PR, see this commit https://github.com/prestodb/presto/commit/c2e3038ea4b9a0775bdcff4097af7bda8e1c07d8#diff-753c310d92de3307f8946f327961233e18e643ebfa88e337bd9542d365b259f4, after confirming the prestissimo CI works, I opened this PR.
And then revert the velox changes in https://github.com/prestodb/presto/pull/26908. Because https://github.com/prestodb/presto/pull/26982 has picked the commit from 26908.

> IcebergColumnHandle now depends on ParquetFieldId unconditionally

Yes, you are right. My understanding is that `velox/dwio/parquet/writer/ParquetFieldId.h` is a simple header-only file and does not depend on any other headers, especially those from Parquet.
Although its parent directory velox/dwio/parquet/writer/ is conditionally compiled, this header is referenced via its full path and therefore can be included unconditionally (Velox root CMakeLists.txt adds the entire velox/ directory to the include path).

> Is the test not triggering hitting that code path, so it's getting optimized away? 

IcebergColumnHandle.cpp is unconditionally compiled. 

> What happens if we try to use this code and Parquet is disabled in Velox?

I tried this, and the compiling is successful. 

```
if(VELOX_ENABLE_PARQUET)
  velox_link_libraries(velox_hive_iceberg_splitreader velox_dwio_parquet_field_id)
endif()
```

Yes, I tried this previously as well, and I think it’s not necessary since we can simply remove it.

Let’s see whether this change breaks anything in Meta. I’m open to taking the approach you suggested.

kevinwilfong:
Thanks for the explanation, that sounds reasonable

> Velox root CMakeLists.txt adds the entire velox/ directory to the include path

That's an important detail I was missing

This is guaranteed not to break anything in Meta as we don't use CMake (hence my limited familiarity)

kgpai:
@PingLiuPing I think depending on the root path for is inherently brittle . One other option is to add the velox_dwio_parquet_field_id target in velox/dwio/parquet/CMakeLists.txt so its always present irrespective of whether VELOX_ENABLE_PARQUET is on/off. This way we can maintain the interface dependency.

PingLiuPing:
> @PingLiuPing I think depending on the root path for is inherently brittle . One other option is to add the velox_dwio_parquet_field_id target in velox/dwio/parquet/CMakeLists.txt so its always present irrespective of whether VELOX_ENABLE_PARQUET is on/off. This way we can maintain the interface dependency.

@kgpai , currently there is a cmake target `velox_dwio_parquet_field_id` in `dwio/parquet/writer/CMakeLists.txt `. But the parent folder `dwio/parquet/writer` is conditionally compiled 
if(VELOX_ENABLE_PARQUET)
  ...
  add_subdirectory(writer)
endif()

Should we move ParquetFieldId.h from `dwio/parquet/writer/ParquetFieldId.h` to `dwio/parquet/ParquetFieldId.h` ?
And then link this target from Iceberg connector

```
if(VELOX_ENABLE_PARQUET)
  velox_link_libraries(velox_hive_iceberg_splitreader velox_dwio_parquet_field_id)
endif()
```

kgpai:
@PingLiuPing Yes thats my suggestion . That way we do not have to rely on just path resolution for this to work.

PingLiuPing:
> @PingLiuPing Yes thats my suggestion . That way we do not have to rely on just path resolution for this to work.

Sounds good. Let me change this.

PingLiuPing:
@kgpai @kevinwilfong  Updated code, would you take a look please?

PingLiuPing:
The uni test failure in Linux is not related. I see same errors from other PR too.

czentgr:
@amitkdutta @kgpai @kevinwilfong Please take another look when you get a chance. This is one of the PRs blocking the Velox update in Prestissimo.

PR #16108, opened 2026-01-23
fix: Load reused lazy vector in filter project

rui-mo:
In queries like:

```
SELECT c0, sum(c1), sum(c2)
FROM (
  SELECT a as c0, b AS c1, b AS c2
  FROM t
) GROUP BY c0
```

The first sum consumes the LazyVector via a hook-based load. The values are 
streamed directly into the aggregate, and the vector is not materialized. The 
second sum detects the input is already loaded and attempts to read 
materialized data. This leads to invalid access, resulting in incorrect results 
or a crash.

This PR loads the lazy vector in the filter project output when multiple output
vectors share the same underlying LazyVector.

### DIFF
--- a/velox/dwio/parquet/tests/reader/ParquetTableScanTest.cpp
+++ b/velox/dwio/parquet/tests/reader/ParquetTableScanTest.cpp
@@ -1610,6 +1610,36 @@ TEST_F(ParquetTableScanTest, inFilter) {
           "SELECT name FROM tmp where name not in ('alex', 'leo', 'mary', null, 'victor')");
 }
 
+TEST_F(ParquetTableScanTest, reusedLazyVectors) {
+  const std::vector<std::string> columnNames = {"a", "b"};
+  std::vector<RowVectorPtr> data;
+  for (auto row = 0; row < 10; ++row) {
+    data.emplace_back(makeRowVector(
+        columnNames,
+        {
+            makeFlatVector<int64_t>({static_cast<int64_t>(row % 5)}),
+            makeFlatVector<int64_t>({static_cast<int64_t>(row)}),
+        }));
+  }
+  const auto expectedRowVector = makeRowVector(
+      {makeFlatVector<int64_t>({0, 1, 2, 3, 4}),
+       makeFlatVector<int64_t>({5, 7, 9, 11, 13}),
+       makeFlatVector<int64_t>({5, 7, 9, 11, 13})});
+
+  const auto filePath = TempFilePath::create();
+  WriterOptions options;
+  writeToParquetFile(filePath->getPath(), data, options);
+
+  const auto plan = PlanBuilder()
+                        .tableScan(ROW(columnNames, {BIGINT(), BIGINT()}))
+                        .project({"a as c1", "b as c2", "b as c3"})
+                        .singleAggregation({"c1"}, {"sum(c2)", "sum(c3)"})
+                        .planNode();
+  AssertQueryBuilder(plan)
+      .split(makeSplit(filePath->getPath()))
+      .assertResults(expectedRowVector);
+}
+
 int main(int argc, char** argv) {
   testing::InitGoogleTest(&argc, argv);
   folly::Init init{&argc, &argv, false};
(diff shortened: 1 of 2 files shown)


Yuhta:
This is a bug in optimizer, we should not calculate sum(b) twice

rui-mo:
@Yuhta In a query like the one below, `SUM(c0)` and `SUM(c1)` are treated as two distinct aggregate expressions, even if c0 and c1 are aliases of the same underlying input. In more complex queries, the optimizer may not be able to reliably trace each column back to its source, and therefore cannot safely determine whether the two aggregates are truly identical. Moreover, if pushdown is extended to other aggregate functions beyond `SUM`, a similar issue could arise in cases like`SUM(c0), AVG(c1)`.

For above reasons, it would be ideal for Velox to remain correct even when the plan includes multiple aggregates over the same underlying source. How do you think? Thanks!

```
SELECT c0, sum(c1), sum(c2)
FROM (
  SELECT a as c0, b AS c1, b AS c2
  FROM t
) GROUP BY c0
```

Yuhta:
It is a reasonable requirement for optimizer to remove aliasings.  On Velox side we already have a check that if a `LazyVector` is already loaded, an exception would be thrown instead of crash or causing correctness issue.  This is better than eagerly materializing them and hiding potential performance issues.

https://github.com/facebookincubator/velox/blob/1d4ca7f1083b117ef84337239510ec2ef4f37b30/velox/vector/LazyVector.cpp#L259

rui-mo:
> It is a reasonable requirement for optimizer to remove aliasings.

@Yuhta I agree it’s reasonable for the optimizer to collapse simple alias chains, and we could consider that as a separate improvement. However, I don’t think execution correctness should depend on alias removal being complete. In more complex plans, provenance can be lost or obscured, and multiple aggregates may still end up consuming the same underlying lazy source (e.g. SUM and AVG, or two SUMs over the same input through different projections). 

>  On Velox side we already have a check that if a LazyVector is already loaded, an exception would be thrown instead of crash or causing correctness issue. 

However, in this case the exception is not triggered, and we end up with an incorrect result without any additional signal. This happens because the first SUM marks `allLoaded_` as true, so the second SUM does not attempt to load the input again and instead directly accesses the invalid materialized vector.

Yuhta:
`SUM(a)`, `AVG(a)` should be handled properly already.

> the first SUM marks allLoaded_ as true, so the second SUM does not attempt to load the input again and instead directly accesses the invalid materialized vector.

That is the bug that needs to be fixed.  Can you change your implementation to throw an error if we detect same lazy vector is being supplied to multiple columns?  Also aggregation is not the only place things can go wrong if same lazy vector is put in multiple columns, so it's also better to move this check to output of filter project to avoid that thing being generated in first place.

rui-mo:
@Yuhta Thanks for the suggestions. I’ve revised this PR accordingly. When you have a chance, could you please take another look? Thanks!

rui-mo:
Hi @Yuhta, after make the change accordingly, I noticed that in Velox it is possible for a lazy vector to be supplied to multiple columns. You could refer to the tests below as examples.

```
TEST_P(HashJoinTest, innerJoinForTypeWithCustomComparisonAndSmallVector)
TEST_F(SqlTest, tableScan)
TEST_F(TableScanTest, duplicateFieldProject)
```

These tests make me wonder whether forbidding this behavior is the right fix, since from the planner’s perspective it seems like a normal and valid case.
From my understanding, the issue may not be the lazy vector being reused across multiple columns, but rather that the pushdown loads the lazy vector through the hook and marks it as loaded when it hasn’t actually been materialized.
Given that, would it make sense to materialize the lazy vector once we detect it’s referenced more than once, before we decide to load it through the hook (i.e., disabling hook pushdown for that column as https://github.com/facebookincubator/velox/commit/2bbdf3aecef1baa9b528297a19f2fcb0a5889af0 did)?
I’d really appreciate your thoughts on this. cc: @mbasmanova

Yuhta:
@rui-mo It's not possible to detect if it is referenced multiple times from the lazy vector itself or from the pushdown site.  The only place we can do something when it is forked.  Another option is we materialize the lazy vector as soon as we detect it is forked in `FilterProject`.

PR #16111, opened 2026-01-23
fix(cudf): Slow S3 IO performance

majetideepak:
Use IO executor to schedule the chunk reads async.
Set `connector.num-io-threads-hw-multiplier` to 2 or 4 from Presto.

### DIFF
--- a/velox/experimental/cudf/connectors/hive/CudfHiveDataSourceHelpers.cpp
+++ b/velox/experimental/cudf/connectors/hive/CudfHiveDataSourceHelpers.cpp
@@ -26,10 +26,30 @@
 #include <cudf/io/parquet.hpp>
 #include <cudf/io/types.hpp>
 
+#include <folly/futures/Future.h>
+
 #include <list>
 #include <string>
 #include <unordered_map>
 #include <vector>
+
+namespace {
+template <typename T>
+std::future<T> toStdFuture(folly::Future<T> follyFuture) {
+  auto promise = std::make_shared<std::promise<T>>();
+  auto stdFuture = promise->get_future();
+
+  std::move(follyFuture).thenTry([promise](folly::Try<T>&& result) mutable {
+    if (result.hasValue()) {
+      promise->set_value(std::move(result.value()));
+    } else {
+      promise->set_exception(result.exception().to_exception_ptr());
+    }
+  });
+
+  return stdFuture;
+}
+} // namespace
 
 namespace facebook::velox::cudf_velox::connector::hive {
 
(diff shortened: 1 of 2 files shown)


velox/experimental/cudf/connectors/hive/CudfHiveDataSourceHelpers.cpp
```diff
@@ -78,8 +98,28 @@ std::future<size_t> BufferedInputDataSource::host_read_async(
...
+                      auto hostBuffer = this->host_read(offset, size);
+                      CUDF_CUDA_TRY(cudaMemcpyAsync(
+                          dst,
+                          hostBuffer->data(),
+                          hostBuffer->size(),
+                          cudaMemcpyHostToDevice,
+                          stream.value()));
+                      return hostBuffer->size();
+                    });
+  return toStdFuture(std::move(future));
```
mhaseeb123, line 118: I think this is exactly the same as what is currently happening

Main cudf reader: https://github.com/rapidsai/cudf/blob/6d453dfe114b0f864611eba3e0b4f1a81e353b98/cpp/src/io/parquet/reader_impl_preprocess_utils.cu#L203-L214

Hybrid scan reader: https://github.com/facebookincubator/velox/blob/0cd99c637cfcd181c08cffa96e51ac32b841372d/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp#L264-L273

majetideepak: Yes, but chunks are read sequentially, which is very slow from S3. The idea is to use the async path in the `if` block above with this change.
I will share the performance numbers shortly.

majetideepak:
First 6 TPC-H query performance after clearing cache before each query.
Original (seconds) | This PR fix (seconds)
-- | --
61 | 11
15​ | 5
128​ | 19
36​ | 8
109​ | 16
64​ | 7

majetideepak:
@zoltan can you share numbers from your run? Thanks.

velox/experimental/cudf/connectors/hive/CudfHiveDataSourceHelpers.cpp
```diff
@@ -78,8 +98,28 @@ std::future<size_t> BufferedInputDataSource::host_read_async(
   });
 }
 
+std::future<size_t> BufferedInputDataSource::device_read_async(
+    size_t offset,
+    size_t size,
+    uint8_t* dst,
+    rmm::cuda_stream_view stream) {
+  VELOX_CHECK(input_->executor() != nullptr, "IO executor is not initialized");
+  auto future = folly::via(input_->executor())
```
devavret, line 107: Since `BufferedInput` is not pure virtual, is there a danger of getting a `nullptr` for executor?

zoltan:
I've compiled g6e and g7e performance data on AWS. let me share it in a bit.

zoltan (approved):
AWS g6e.4xlarge, against S3:

1 driver:

pre-fix: Cold: 1196s / Warm: 40s
with the fix as-is: Cold: 630s / Warm: 33s
with x2 threads: Cold: 498s / Warm: 33s
with x4 threads: Cold: 424s / Warm: 33s

2 driver:

with the fix: Cold: 467s / Warm: 28s
with x2 threads: Cold: 426s / Warm: 25s
with x4 threads: Cold: 373s / Warm: 26s

velox/experimental/cudf/connectors/hive/CudfHiveDataSourceHelpers.cpp
majetideepak: No danger here.

Issue #16041, opened 2026-01-16
Complete Presto SQL function coverage


czentgr:
### Description

This issue is to record a push to complete the Presto SQL function coverage in Velox.
Meta has completed conversion of the Java usage to PrestoC++ (as recently announced by Amit). 

In IBM we have both Presto and PrestoC++ as offering but are looking to a PrestoC++ only offering. As such we want to finalize the built-in functions that are necessary to claim full compatibility. That's what we view to complete the offering.

Therefore, we reviewed what function would be needed to be added to close the coverage of the functions.
Some functions are considered to be useful for a narrow use-case (e.g. ML/AI) and if other functions are needed they can be implemented using the user-defined functions capability in PrestoC++.

Most SQL functions have been implemented by now. 
Part of this issue is also resolving missing documentation and documentation links for existing functions.

We reviewed the function coverage and list missing functions here and indicate which functions we deem necessary to implement. Some of the function we still need some coordination/conversation because of the various efforts being made (e.g. the spatial functions).

Functions that are already in progress are not listed.

Function listed as "Prioritized by IBM" means we are focused on implemented but anyone can pick up the issue. If it is not picked up by the time we are ready to implement it, we are going ahead and implement them.

### SQL inlined (scalar) functions
| Name | Comment |
|-------|-----------|
| array_least_frequent | Prioritized by IBM |
| array_split_into_chunks| Prioritized by IBM |
| array_top_n | Missing variant with 3 arguments (custom sorting lambda), Prioritized by IBM|
| map_top_n_keys | Missing variant with 3 arguments (custom sorting lambda), Prioritized by IBM, in progress |
| map_top_n_values | Missing variant with 3 arguments (custom sorting lambda), Prioritized by IBM |
| key_sampling_percent | Prioritized by IBM |

### Scalar functions

| Name | Comment |
|-------|-----------|
| bar | Will not implement, not standard function |
| classify | Will not implement, ML function |
| color | Will not implement |
| features | Will not implement, ML function |
| google_polyline_decode | Prioritized by IBM |
| google_polyline_encode | Prioritized by IBM |
| l2_squared | Will not implement, ML function |
| merge_khll | Discussion |
| myanmar_font_encoding | Will not implement |
| myanmar_normalize_unicode | Will not implement |
| pinot_binary_decimal_to_double | Will not implement |
| regress | Will not implement, ML function |
| reidentification_potential | Will not implement, ML function |
| render | Will not implement |
| rgb | Will not implement |
[...]

PR #16113, opened 2026-01-23
feat: Add key_sampling_percent scalar function

jkhaliqi:
Implement key_sampling_percent as part of #16041

### DIFF
--- a/velox/functions/prestosql/StringFunctions.h
+++ b/velox/functions/prestosql/StringFunctions.h
@@ -924,4 +924,23 @@ struct JaroWinklerSimilarityFunction {
   }
 };
 
+/// key_sampling_percent(varchar) -> double
+/// Returns a value between 0.0 and 1.0 using the hash of the given input
+/// string.
+template <typename T>
+struct KeySamplingPercentFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<double>& result,
+      const arg_type<Varchar>& string) {
+    static constexpr auto kTypeLength = sizeof(int64_t);
+    int64_t hash = XXH64(string.data(), string.size(), 0);
+    // Bitwise reinterpretation to match Java's
+    // Double.longBitsToDouble(xxhash64(...)) semantics.
+    // This may produce NaN, which is expected and allowed.
+    memcpy(&result, &hash, kTypeLength);
+    result = fmod(fabs(result), 100) / 100;
+  }
+};
 } // namespace facebook::velox::functions
--- a/velox/functions/prestosql/tests/StringFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/StringFunctionsTest.cpp
@@ -2481,3 +2481,21 @@ TEST_F(StringFunctionsTest, xxHash64FunctionVarchar) {
   // Non-ASCII strings
   EXPECT_EQ(8176744303664166369, xxhash64("日本語"));
 }
+
+TEST_F(StringFunctionsTest, keySamplingPercent) {
+  const auto keySamplingPercent =
+      [&](const std::optional<std::string>& string) {
+        return evaluateOnce<double>("key_sampling_percent(c0)", string);
+      };
+
+  EXPECT_EQ(std::nullopt, keySamplingPercent(std::nullopt));
+  EXPECT_EQ(0.56, keySamplingPercent("abc"));
+  EXPECT_EQ(6.11561179120687E-153, keySamplingPercent("abcdefghskwkjadhwd"));
+  EXPECT_EQ(2.393674127734674E-93, keySamplingPercent("001yxzuj"));
+  EXPECT_EQ(0.48, keySamplingPercent("56wfythjhdhvgewuikwemn"));
+  EXPECT_EQ(0.7520703125, keySamplingPercent("special_#@,$|%/^~?{}+-"));
+  EXPECT_EQ(0.4, keySamplingPercent("     "));
+  EXPECT_EQ(0.28, keySamplingPercent(""));
+  EXPECT_EQ(
+      4.143659858002825E-274, keySamplingPercent("Hello World from Velox!"));
+}
--- a/velox/docs/functions/presto/string.rst
+++ b/velox/docs/functions/presto/string.rst
@@ -372,3 +372,8 @@ Unicode Functions
 .. function:: to_utf8(string) -> varbinary
 
     Encodes ``string`` into a UTF-8 varbinary representation.
+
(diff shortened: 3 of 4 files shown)


aditi-pandit:
@jkhaliqi @czentgr : This is a SQL function in Java and there is previous work https://github.com/facebookincubator/velox/pull/8481 where we decided not to add this function in native. Have we debugged a SQL statement with key_sampling_percent and understand why it didn't work ?

jkhaliqi:
@aditi-pandit Thanks for the review, so it looks like we want these implemented now so this test can now pass [https://github.com/prestodb/presto/blob/4be75f05db0be6b49291593202c21e5edbb034b8/p[…]t/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java](https://github.com/prestodb/presto/blob/feb403bd5825a374a3dda7e29fd594d8294243da/presto-native-sidecar-plugin/src/test/java/com/facebook/presto/sidecar/TestNativeSidecarPlugin.java#L628).
Previously they wanted it removed because of this issue https://github.com/prestodb/presto/issues/22085 where the error would state 
`Query 20240304_044118_00007_4g98f failed: Scalar function name not registered: presto.default.key_sampling_percent, called with arguments: (VARCHAR).`
But now this has been fixed with this PR: https://github.com/prestodb/presto/pull/22095
So instead retrying that would give
```
Query 20260129_215630_00004_hwx54 failed:  Scalar function name not registered: presto.default.$operator$hash_code, called with arguments: (BIGINT).
```
Which is the same error we would get without this code so it is not a problem of adding in this PR now.
cc: @czentgr

velox/functions/prestosql/tests/StringFunctionsTest.cpp
```diff
@@ -2481,3 +2481,20 @@ TEST_F(StringFunctionsTest, xxHash64FunctionVarchar) {
   // Non-ASCII strings
   EXPECT_EQ(8176744303664166369, xxhash64("日本語"));
 }
+
+TEST_F(StringFunctionsTest, keySamplingPercent) {
+  const auto keySamplingPercent = [&](std::optional<std::string> string) {
```
czentgr: Make the argument a const&. We can save a copy this way.

czentgr:
Please rebase and run the expression fuzzer with the function and post the output into the PR.
Reference: https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md#adding-scalar-functions step 7.

Please also check the Presto Bias fuzzer.

velox/functions/prestosql/StringFunctions.h
```diff
@@ -924,4 +924,20 @@ struct JaroWinklerSimilarityFunction {
...
+template <typename T>
+struct KeySamplingPercentFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<double>& result,
+      const arg_type<Varchar>& string) {
+    static constexpr auto kTypeLength = sizeof(int64_t);
+    int64_t hash = XXH64(string.data(), string.size(), 0);
+    memcpy(&result, &hash, kTypeLength);
```
czentgr, line 942: We are copying a `int64` value into a `double` typed variable and then call abs on it?
Calling abs on a double is problematic because it has its own function called fabs. But why use abs anyway and then in a double type variable?
It doesn't make much functional sense to do this.

XXH64 produces results of type `XXH64_hash_t` which looks to be a uint_64. So we should be using uint anyway, then we don't need the abs/fabs at all.

So shouldn't
```
result = static_cast<double>(result % 100) / 100.0;
```
generate the equivalent result?

Did we check with Java values?

jkhaliqi: According to this conversation seems like the java values were mismatching when the above change was not happening. from_ieee754_64(xxhash64(...)) in Java reinterprets the hash bits as a double (equivalent to Double.longBitsToDouble) and can produce NaN. The memcpy is required to match this behavior in C++. As seen from: https://github.com/facebookincubator/velox/pull/8481#discussion_r1492055368

jkhaliqi:
```
E20260206 15:13:32.682310 539451 ExpressionFuzzerVerifier.cpp:220] ==============================> Top 1 by number of rows processed
E20260206 15:13:32.682317 539451 ExpressionFuzzerVerifier.cpp:222] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260206 15:13:32.682322 539451 ExpressionFuzzerVerifier.cpp:226] key_sampling_percent 99133 100.00% 1585143
E20260206 15:13:32.682332 539451 ExpressionFuzzerVerifier.cpp:233] ==============================> Bottom 1 by number of rows processed
E20260206 15:13:32.682337 539451 ExpressionFuzzerVerifier.cpp:235] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260206 15:13:32.682343 539451 ExpressionFuzzerVerifier.cpp:240] key_sampling_percent 99133 100.00% 1585143
E20260206 15:13:32.682349 539451 ExpressionFuzzerVerifier.cpp:253] ==============================> All stats sorted by number of times the function was chosen
E20260206 15:13:32.682354 539451 ExpressionFuzzerVerifier.cpp:255] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260206 15:13:32.682359 539451 ExpressionFuzzerVerifier.cpp:259] key_sampling_percent 99133 100.00% 1585143
E20260206 15:13:32.682366 539451 ExpressionFuzzerVerifier.cpp:488] Total test cases: 136220
E20260206 15:13:32.682371 539451 ExpressionFuzzerVerifier.cpp:489] Total test cases verified in the reference DB: 136220
E20260206 15:13:32.682377 539451 ExpressionFuzzerVerifier.cpp:491] Total test cases failed in both Velox and the reference DB: 0
E20260206 15:13:32.682384 539451 ExpressionFuzzerVerifier.cpp:493] Total test cases unsupported in the reference DB: 0
```

velox/functions/prestosql/StringFunctions.h
czentgr: Ok, looks like this is intentional. `fabs` is the correct function to use for the double type that is filled with the hash value. So this should be good enough.

czentgr (approved):
Thanks! Looks like CI isn't triggered. Maybe more of the github problems?
@majetideepak Can you please review since you reviewed the original? This is part of adding the remaining inlined SQL functions as native functions.

velox/functions/prestosql/StringFunctions.h
majetideepak: std::abs() or fabs both work.

PR #16122, opened 2026-01-25
feat: Add needsIndexSplit support to IndexLookupJoin through hive connector

xiaoxmeng:
Summary:
This diff extends the IndexLookupJoin operator to support index sources that require splits before performing lookups.

**ConnectorTableHandle interface**: Added `needsIndexSplit()` virtual method that returns true if the table handle requires splits for index lookup. Default returns false for backward compatibility.

**IndexSource interface**: Added `addSplits()` method to allow passing splits to the index source after collection.

**IndexLookupJoin operator**:
   - Added split collection logic in `collectIndexSplits()` that blocks until all splits are received
   - The operator waits for splits before accepting probe input when `needsIndexSplit()` is true
   - Splits are passed to the index source via `addSplits()` after the no-more-splits signal

**Task split registration**: Modified `numSourceNodes()` to register both probe and index scan nodes for split collection when `needsIndexSplit()` is true.

**Test extension**: 
   - Extended `TestIndexTableHandle` with `needsIndexSplit` parameter
   - Added `TestIndexConnectorSplit` class for testing
   - Added unit tests for split handling behavior with facility of test value

Differential Revision: D91331636


velox/exec/tests/IndexLookupJoinTest.cpp
```diff
@@ -282,6 +307,8 @@ TEST_P(IndexLookupJoinTest, planNodeAndSerde) {
                     .planNode();
```
zacw7, line 314: set needsIndexSplit() in the plan?

PR #16128, opened 2026-01-26
perf(TextReader): Switch parsing to fast_float, replace std::string with std::vector<char>

pashandor789:
# Changes

1) I've chosen fast_float, benchmarked it and it gave 15% boost 
2) Also i've deleted checks - they seem to me very weird because they're supposed to be checked by parsing library
3) I've added some boolean variants (t, f, T, F, 0, 1) because they're possible for CSV format and presto does the same
4) I've changed std::string -> std::vector<char> and replaced append with push_back since std::vector<> doesn't have SSO checks and doesn't make string null terminated after every append - that change gives +16% to previous boost

# Benchmark

I used clickbench data to test on. I did it on one-thread and also used truncated version (1 GB instead of 70 GB). 

here are the results : 

main version - 0 change
Average time: 14219ms

sscan --> from_chars 
Average time: 12031ms

std::vector<char> 
Average time: 10580ms

Total boost : +25%

### DIFF
--- a/velox/dwio/text/reader/TextReader.cpp
+++ b/velox/dwio/text/reader/TextReader.cpp
@@ -16,7 +16,7 @@
 
 #include "velox/dwio/text/reader/TextReader.h"
 
-#include <boost/algorithm/string/predicate.hpp>
+#include <fast_float/fast_float.h>
 #include <string>
 
 #include "velox/common/encode/Base64.h"
@@ -40,8 +40,6 @@ static constexpr std::string_view kTextfileCompressionExtensionLz4{".lz4"};
 static constexpr std::string_view kTextfileCompressionExtensionLzo{".lzo"};
 static constexpr std::string_view kTextfileCompressionExtensionSnappy{
     ".snappy"};
-
-static std::string emptyString = std::string();
 
 constexpr const int32_t kDecompressionBufferFactor = 3;
 
@@ -473,7 +471,7 @@ bool TextRowReader::isNone(DelimType delim) {
   return (delim == DelimTypeNone);
 }
 
-std::string&
+std::string_view
 TextRowReader::getString(TextRowReader& th, bool& isNull, DelimType& delim) {
   if (th.atEOL_) {
     delim = DelimTypeEOR; // top-level EOR
@@ -481,7 +479,7 @@ TextRowReader::getString(TextRowReader& th, bool& isNull, DelimType& delim) {
 
   if (th.isEOEorEOR(delim)) {
     isNull = true;
-    return emptyString;
+    return {};
   }
 
   bool wasEscaped = false;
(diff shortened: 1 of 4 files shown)


velox/dwio/text/reader/TextReader.cpp
```diff
@@ -758,52 +760,18 @@ T TextRowReader::getInteger(TextRowReader& th, bool& isNull, DelimType& delim) {
...
-      return 0;
-    }
-  }
-
-  if (!std::is_same<T, int64_t>::value) {
-    if (static_cast<int64_t>(static_cast<T>(v)) != v) {
-      isNull = true;
-      return 0;
-    }
-  }
```
zacw7, line 796: This code is to make sure no overflow. Does the library handle it?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -758,52 +760,18 @@ T TextRowReader::getInteger(TextRowReader& th, bool& isNull, DelimType& delim) {
     return 0;
   }
 
-  // Test if s is not acceptable integer format for
-  // the warehouse, for cases accepted by stol().
-  char c = str[0];
-  if (c != '-' && !std::isdigit(static_cast<unsigned char>(c))) {
```
zacw7, line 764: It seems we only allow the integer starts with '-' or digit. Does fast_float accept leading spaces?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -841,62 +818,6 @@ bool TextRowReader::getBoolean(
...
-  bool isShortInf = boost::algorithm::iequals(s, kShortInfinityStringView);
-
-  bool isNegInf = boost::algorithm::iequals(s, kNegInfinityStringView);
-  bool isShortNegInf =
-      boost::algorithm::iequals(s, kShortNegInfinityStringView);
-
-  return (!isNaN && !isInf && !isShortInf && !isNegInf && !isShortNegInf);
-}
-
-void trimStringInPlace(std::string& s) {
```
zacw7, line 872: Does the library allow leading/trailing whitespaces?

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -758,52 +760,18 @@ T TextRowReader::getInteger(TextRowReader& th, bool& isNull, DelimType& delim) {
...
-      if (i == scanPos && str[i] == '.') {
-        continue;
-      }
-      if (str[i] >= '0' && str[i] <= '9') {
-        continue;
-      }
-      isNull = true;
-      return 0;
-    }
-  }
```
zacw7, line 789: The current logic does decimal truncation. example, 123.45 (integer) is read as 123. I wonder if the library behaves the same.

zacw7 (commented):
Thanks @pashandor789 for drafting the changes.
1. There are a few potential behavior changes I put in the comments could you please help to verify.
2. Could you share more details about the benchmarking?

velox/dwio/text/reader/TextReader.cpp
pashandor789: yes of course, i've made variable `value` type of `uint64_t`
and above there's a check `ec != std::errc{}`. in case overflow occurs appropirate overflow error

velox/dwio/text/reader/TextReader.cpp
pashandor789: by default not, i'm gonna turn it on

velox/dwio/text/reader/TextReader.cpp
pashandor789: yes, answered below

pashandor789:
@zacw7 hey! 

about benchmarking - i's testing truncated clickbench data ~ 1 GB. Before i got 14 seconds after it 12 seconds.

velox/dwio/text/reader/TextReader.cpp
pashandor789: by the way it doesn't
but presto doesn't have the same behaviour either 


```
docker exec presto presto-cli --server localhost:8080 --execute "SELECT CAST('1.1' AS INTEGER);"
Query 20260127_122442_00005_r8u9a failed: Cannot cast '1.1' to INT
```

pashandor789:
@zacw7 Hello again i've changed description to make it more clear

velox/dwio/text/reader/TextReader.cpp
zacw7: This is the Textfile reader behavior in HiveConnector and it has nothing to do with cast. Maybe keep the logic to make the behavior consistent.

pashandor789: but the behavior of this thing is system specific for instance in PG they use their type parsers and i think so about presto

zacw7: or is it possible to make it configurable?

pashandor789: it can 
shall i add it ?

zacw7: yes please

PR #16131, opened 2026-01-26
build(cuda): Upgrade to CUDA 12.9

bdice:
This PR upgrades Velox's adaptors images to CUDA 12.9. This is now a [build-time requirement for cuDF](https://github.com/rapidsai/cudf/pull/21186).

cuDF is still compatible with CUDA 12.2+ at runtime, only the build-time requirement has increased.


Yuhta:
Note that we still have CUDA 12.8 as prod version in Meta, so the code in wave and breeze must be 12.8 compatible (cudf integration is ok).

karthikeyann:
cc @jinchengchenghh @zoltan @majetideepak @czentgr

Issue #16039, opened 2026-01-16
StringView should delete const&& string conversions


pedroerp:
### Description

Following up on a discussion on https://github.com/facebookincubator/velox/pull/15946, in StringView we should delete the `const&&` string conversion method, not only the `&&`. 

One specific callsite where this breaks is SimpleVector::valueAt():

https://github.com/facebookincubator/velox/blob/main/velox/vector/SimpleVector.h#L150

Because `valueAt()` returns const T, excerpts such as:

`
auto sv = std::string_view(vector->valueAt(0));
`

are still allowed. This would call a const&& conversion operator, which is not deleted (only the non-const && is), and  leave `sv` with a dangling reference to the internal buffer of the temporary that was just copied by `valueAt()`. 

To clean this up, there are a few options:

1. Simply make the change in StringView and clean up callsites appropriately.
2. Modify SimpleVector::valueAt() to return a `const StringView&` instead of `const StringView`.
3. Modify SimpleVector::valueAt() to return a non-const value instead; there is no value in setting it as const if it's always a copy.

The second may be slightly more efficient as it prevents copies, enabling the caller to either keep the reference and make a local copy.

Thoughts?

Cc: @MBkkt @mbasmanova @majetideepak @czentgr @xiaoxmeng @Yuhta @kKPulla @jagill @kevinwilfong @magoja

Magoja:
Can we just remove `const` from the function?
```
virtual T valueAt(vector_size_t idx) const = 0;
```
Why do we even return `const T` in the first place?

pedroerp:
> Why do we even return const T in the first place?

I don't know. Maybe to signal the caller that there is no point in changing the objects returned? But point taken, a third option would be to just remove that const from valueAt()'s return value.

MBkkt:
@pedroerp 

Right now `valueAt` return `const T` or `T` instead of `const T&` only because of BiasVector.
I know this because it was very inconvenient, maybe you or someone else have idea how can we modify/remove/adjust BiasVector?

MBkkt:
@Magoja 

> Why do we even return const T in the first place?

This is typical typo in Velox/Axiom, I noticed that there're multiple places where `const T` returned instead of `const T&`, I don't know why, I think it was from beginning.


Exists clang-tidy warning about this, we can fix it

pedroerp:
@MBkkt @Magoja I can try to look into BiasVectors, but the question really is: would we like to return `T` or `const T&`?

[...]

PR #16132, opened 2026-01-26
perf: Make valueAt return `const StringView&` instead of `const StringView`

MBkkt:
This avoids copy and needed to fix https://github.com/facebookincubator/velox/issues/16039

### DIFF
--- a/velox/vector/SimpleVector.h
+++ b/velox/vector/SimpleVector.h
@@ -144,12 +144,15 @@ class SimpleVector : public BaseVector {
     stats_ = std::move(stats);
   }
 
+  // Type of value returned depends on size of T. If T can fit in a machine
+  // word, return by value, otherwise return by const reference. This is an
+  // optimization to avoid copying large types.
+  using TValueAt = std::conditional_t<sizeof(T) <= sizeof(void*), T, const T&>;
+
   // Concrete Vector types need to implement this themselves.
   // This method does not do bounds checking. When the value is null the return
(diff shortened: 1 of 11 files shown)


velox/vector/SimpleVector.h
```diff
@@ -147,7 +147,9 @@ class SimpleVector : public BaseVector {
   // Concrete Vector types need to implement this themselves.
   // This method does not do bounds checking. When the value is null the return
   // value is technically undefined (currently implemented as default of T)
-  virtual const T valueAt(vector_size_t idx) const = 0;
+  using ValueAtType =
```
pedroerp: this makes sense. Can you just add a comment here explaining why we do this, so that we can return references to large types and prevent a copy?

pedroerp: also, we usually use the "T" prefix for types, so this should be "TValueAt"

MBkkt: > also, we usually use the "T" prefix for types, so this should be "TValueAt"

I didn't see this, can you share example for me please?

MBkkt: I find few examples, but also a lot of place without this `T` prefix

MBkkt: @pedroerp I fixed what you asked https://github.com/facebookincubator/velox/pull/16132/changes/e1415a93b839e7cfd723d96b58af741bd8ff9ba3

PR #16137, opened 2026-01-27
build: Split iceberg target

PingLiuPing:
The Iceberg connector is tightly coupled with Parquet. However, there is a known issue when referencing Parquet symbols from the connector layer, which can result in duplicate symbols from Thrift. Several workarounds have been adopted to mitigate this problem.

One way to address this cleanly is to conditionally compile the Iceberg connector only when VELOX_ENABLE_PARQUET is enabled.

This makes sense to some extent. The Iceberg specification supports three file formats: Avro, Parquet, and ORC. Velox currently supports ORC and Parquet, and Parquet is the de facto standard in most Iceberg deployments. Therefore, enabling Iceberg only when VELOX_ENABLE_PARQUET is set should be acceptable.

Gluten and Prestissimo all enable parquet by default. 

Let’s see how the discussion evolves.

@mbasmanova @yingsu00 @zhouyuan

### DIFF
--- a/velox/connectors/hive/CMakeLists.txt
+++ b/velox/connectors/hive/CMakeLists.txt
@@ -36,7 +36,7 @@ velox_add_library(
 
 velox_link_libraries(
   velox_hive_connector
-  PUBLIC velox_hive_iceberg_splitreader
+  PUBLIC velox_hive_iceberg_reader
   PRIVATE velox_common_io velox_connector velox_dwio_catalog_fbhive velox_hive_partition_function
 )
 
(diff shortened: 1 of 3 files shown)


mbasmanova:
@majetideepak @yingsu00 @Yuhta @pedroerp Any thoughts? Seems non-ideal for a connector to be tightly-coupled with a specific file format.

Yuhta:
@mbasmanova @PingLiuPing As far as I see the prod code is not coupled, it's more of a preference when writing test code.  Can we keep the prod code independent to parquet, and only build test code when parquet is enabled?

PR #16139, opened 2026-01-27
perf: Using Batch serialization in RemoteFunction when preserveEncoding is true

avinaashgupta:
Summary:
Right now constant encoding is not preserved during serialization of Row vectors in RemoteFunction client when creating the thrift request. We use a lot of constant parameters in our remote funciton udf and if we don't preserve the constant encoding, it will more than double/triple the payload size.

After debugging the issue, the fix is to use batch serializer (instead of iterative serializer) which preserves the encoding correctly. I noticed >70% reduced in payload size after testing this change locally for our remote function udf. This is just a simple case, in other cases we will have even more savings (more than 3x reduction in payload size).

Reviewed By: Sullivan-Patrick

Differential Revision: D91602429

### DIFF
--- a/velox/functions/remote/client/RemoteVectorFunction.cpp
+++ b/velox/functions/remote/client/RemoteVectorFunction.cpp
@@ -36,7 +36,12 @@ RemoteVectorFunction::RemoteVectorFunction(
     const RemoteVectorFunctionMetadata& metadata)
     : functionName_(functionName),
       serdeFormat_(metadata.serdeFormat),
-      serde_(getSerde(serdeFormat_)) {
+      serde_(getSerde(serdeFormat_)),
+      serdeOptions_(
+          metadata.preserveEncoding
+              ? getSerdeOptions(serdeFormat_, metadata.preserveEncoding)
+              : nullptr),
+      preserveEncoding_(metadata.preserveEncoding) {
   std::vector<TypePtr> types;
   types.reserve(inputArgs.size());
   serializedInputTypes_.reserve(inputArgs.size());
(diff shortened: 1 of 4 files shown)


velox/functions/remote/client/RemoteVectorFunction.cpp
```diff
@@ -91,8 +93,17 @@ void RemoteVectorFunction::applyRemote(
   requestInputs->pageFormat() = serdeFormat_;
 
   // TODO: serialize only active rows.
-  requestInputs->payload() = rowVectorToIOBuf(
-      remoteRowVector, rows.end(), *context.pool(), serde_.get());
+  if (preserveEncoding_) {
+    // Create PrestoOptions with preserveEncodings enabled to maintain
+    // constant encoding through serialization
+    serializer::presto::PrestoVectorSerde::PrestoOptions options;
```
pedroerp: could we instantiate this in a more general way? Ideally you would keep a pointer to the base class, and instantiate the concrete options object once, based on the serialization type. Right now it's coupled to Presto, so it will break it someone else tries to use with a non-Presto format.

velox/functions/remote/client/RemoteVectorFunction.cpp
```diff
@@ -91,8 +92,18 @@ void RemoteVectorFunction::applyRemote(
   requestInputs->pageFormat() = serdeFormat_;
 
   // TODO: serialize only active rows.
-  requestInputs->payload() = rowVectorToIOBuf(
-      remoteRowVector, rows.end(), *context.pool(), serde_.get());
+  if (preserveEncoding_) {
+    auto options = getSerdeOptions(serdeFormat_, preserveEncoding_);
```
pedroerp: Nice, this is better. Since this won't change from batch to batch, could you instantiate this once in the constructor and keep it as a member, so we don't need to create it over and over?

pedroerp: Or rather, maybe the serialization options should just be part of `RemoteVectorFunctionMetadata`

PR #16143, opened 2026-01-27
fix(cudf): Fix precompile pool name in CudfFilterProject.cpp

karthikeyann:
A single query can have multiple tasks from Presto GPU. so the name has to be unique. So, here we are letting addLeafPool generate unique names for each call.

Related: https://github.com/facebookincubator/velox/pull/16056

### DIFF
--- a/velox/experimental/cudf/exec/CudfFilterProject.cpp
+++ b/velox/experimental/cudf/exec/CudfFilterProject.cpp
@@ -105,8 +105,8 @@ bool canBeEvaluatedByCudf(
     return true;
   }
 
-  auto precompilePool = memory::memoryManager()->addLeafPool(
-      "", /*threadSafe*/ false);
+  auto precompilePool =
+      memory::memoryManager()->addLeafPool("", /*threadSafe*/ false);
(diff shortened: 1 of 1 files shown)


karthikeyann:
cc @paul-aiyedun

velox/experimental/cudf/exec/CudfFilterProject.cpp
```diff
@@ -105,8 +105,8 @@ bool canBeEvaluatedByCudf(
     return true;
   }
 
-  auto precompilePool = memory::memoryManager()->addLeafPool(
-      "cudf-expr-precompile" + queryCtx->queryId(), /*threadSafe*/ false);
+  auto precompilePool =
```
jinchengchenghh, line 108: Do we need to drop the pool when not used? May create many pools and each pool may reserve some memory

jinchengchenghh: The pool will be dropped after calling the precompilePool destruction, so every thing is good.

jinchengchenghh (approved):

PR #16146, opened 2026-01-28
feat: Handle limit 0

natashasehgal:
Handle limit 0 case

Differential Revision: D91646381

### DIFF
--- a/velox/core/PlanNode.h
+++ b/velox/core/PlanNode.h
@@ -4279,6 +4279,8 @@ class LimitNode : public PlanNode {
   // @param isPartial Boolean indicating whether Limit node generates partial
   // results on local workers or finalizes the partial results from `PARTIAL`
   // nodes.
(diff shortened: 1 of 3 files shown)


MBkkt:
@natashasehgal In our Axiom version we workaround this like this https://github.com/serenedb/axiom/commit/0d8a8e207ed54b089cfe32517cd585cf4817f662

natashasehgal:
Thanks @mbasmanova @MBkkt will close this and add the change in Axiom so we are not executing

natashasehgal:
> @natashasehgal In our Axiom version we workaround this like this [serenedb/axiom@0d8a8e2](https://github.com/serenedb/axiom/commit/0d8a8e207ed54b089cfe32517cd585cf4817f662)

@MBkkt This approach looks good to me, is there any plan to merge into`main`? If not I can take it up

PR #16147, opened 2026-01-28
feat: Support reset() for BufferedInput and derived classes for hive index source

xiaoxmeng:
Summary:
Implement the `reset()` method for `BufferedInput` and all its derived classes (`DirectBufferedInput` and `CachedBufferedInput`) to support reusing the same buffered input instance across different operations.

This is needed for index lookup workflows that reuse the same `BufferedInput` across different index lookups. For instance, Nimble file format with cluster index supports index lookup and needs to reset the buffered input state between lookups.

The reset implementation:
- **BufferedInput**: Clears all transient state (`regions_`, `offsets_`, `buffers_`, `enqueuedToBufferOffset_`) and the allocation pool
- **DirectBufferedInput**: Calls parent's reset, cancels pending coalesced loads, and clears `requests_` and stream-to-load mappings
- **CachedBufferedInput**: Calls parent's reset, cancels pending coalesced loads, and clears `requests_` and load mappings

The cancellation logic follows the same pattern as the destructors to ensure proper cleanup of async operations.

Differential Revision: D91388633

### DIFF
--- a/velox/common/file/File.h
+++ b/velox/common/file/File.h
@@ -268,6 +268,7 @@ class InMemoryReadFile : public ReadFile {
   void setShouldCoalesce(bool shouldCoalesce) {
     shouldCoalesce_ = shouldCoalesce;
   }
+
   bool shouldCoalesce() const final {
     return shouldCoalesce_;
   }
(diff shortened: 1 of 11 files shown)


tanjialiang (approved):
Thanks

PR #16148, opened 2026-01-28
docs(ci): Add CI coverage overview for builds

czentgr:


.github/workflow_overview.md
```diff
@@ -0,0 +1,43 @@
...
+Files: `.github/workflows/linux-build.yml` and `.github/workflows/linux-build-base.yml`
+
+The main workflow jobs are implemented in the `linux-build-base.yml` file. This file is included by the `linux-build.yml` file. The `linux-build.yml` file is the main workflow file that is triggered by a push to the main branch or a pull request. It is also triggered by a schedule to run the workflow on a regular basis. The `linux-build-base.yml` file is included by the `linux-build.yml` file to avoid code duplication.
+
+#### PR and merge to main trigger coverage
+
+| Name | Build type | Adapters | GPU | Compiler | Image | Size |
+| ---- | ---------- | -------- | --- | -------- | ----- | ---- |
+| adapters | release | Y | Y | GCC14 | Centos9 | 32 GB / 16 CPU |
+| ubuntu-debug | debug | N | GCC11 | None | 16GB / 8 CPU |
```
kgpai, line 22: This line is missing a column , as it shows GCC11 under gpu. 
Also I think we need a line for whether we run tests or not , which is the main point here. 
Instead of adapters we should probably have a column for 'coverage' and there have adapters - cudf (other optional stuff) etc..

PR #16149, opened 2026-01-28
fix: Enhance StringView by deleting implicit conversions from nullptr and rvalue references

### LINKED ISSUE #16039: StringView should delete const&& string conversions

MBkkt:
Fix #16039

### DIFF
--- a/velox/type/StringView.h
+++ b/velox/type/StringView.h
@@ -117,8 +117,9 @@ struct StringView {
   /// >  StringView sv = "literal";
   /// >  std::optional<StringView> osv = "literal";
   ///
-  /* implicit */ StringView(const char* data)
+  /* implicit */ StringView(const char* FOLLY_NONNULL data)
       : StringView(data, strlen(data)) {}
+  /* implicit */ StringView(std::nullptr_t) = delete;
 
   explicit StringView(const folly::fbstring& value)
       : StringView(value.data(), value.size()) {}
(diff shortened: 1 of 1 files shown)


pedroerp (approved):
Thank you @MBkkt

PR #16151, opened 2026-01-28
fix(cudf): Update default host compiler for adapters-cuda service to gcc-14

karthikeyann:
The default host compiler is set to gcc-12 in adapters-cpp service.
Update default host compiler for adapters-cuda service to gcc-14.
cudf needs gcc-13+

### DIFF
--- a/velox/common/file/File.h
+++ b/velox/common/file/File.h
@@ -268,6 +268,7 @@ class InMemoryReadFile : public ReadFile {
   void setShouldCoalesce(bool shouldCoalesce) {
     shouldCoalesce_ = shouldCoalesce;
   }
+
   bool shouldCoalesce() const final {
     return shouldCoalesce_;
   }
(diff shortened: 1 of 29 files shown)


docker-compose.yml
```diff
@@ -76,6 +76,9 @@ services:
             - driver: nvidia
               count: 1
               capabilities: [gpu]
+    environment:
+      CC: /opt/rh/gcc-toolset-14/root/bin/gcc # Default to gcc-14
+      CXX: /opt/rh/gcc-toolset-14/root/bin/g++ # Default to gcc-14
```
karthikeyann, line 81:
```suggestion
      CXX: /opt/rh/gcc-toolset-14/root/bin/g++ # Default to gcc-14
      CUDAHOSTCXX: /opt/rh/gcc-toolset-14/root/bin/g++ # Default to gcc-14
```

PR #16161, opened 2026-01-29
feat: Allow EncodedVectorCopy to generate FlatMapVector in non-NULL vectors (#16161)

peterenescu:
Summary:

Adds support to EncodedVectorCopy to copy FlatMapVector into non-NULL vectors. Will work in conjunction with D91092336 to support the full FlatMapVector copy capability. Currently has lack of support for wrapped FlatMapVectors due to limitations caused by `BaseVector::create` noted in `createTemplateVector` below. A pervasive solution to solve this will need to be investigated.

Reviewed By: Yuhta

Differential Revision: D91278343


peterenescu:
Adds full support for encoded copy https://github.com/facebookincubator/velox/issues/15485

PR #16162, opened 2026-01-29
feat(functions): Add map_top_n_keys and map_top_n_values

jkhaliqi:
Add `map_top_n_keys` and `map_top_n_values` with transform lambda argument as part of https://github.com/facebookincubator/velox/issues/16041

### DIFF
--- a/velox/docs/functions/presto/map.rst
+++ b/velox/docs/functions/presto/map.rst
@@ -262,6 +262,34 @@ Map Functions
         SELECT map_top_n_keys(map(ARRAY['a', 'b', 'c'], ARRAY[1, 2, 3]), 2) --- ['c', 'b']
         SELECT map_top_n_keys(map(ARRAY['a', 'b', 'c'], ARRAY[1, 2, 3]), 0) --- []
 
+.. function:: map_top_n_keys(map(K,V), n, function(K,V,U)) -> array(K)
+
+    Returns the top N keys of the map sorted by the transform function in descending order.
+    The lambda function is applied to each (key, value) pair and produces an orderable value of type U.
+    Keys are then sorted based on these transformed values, and the top N keys are returned.
+    U must be an orderable type. ``n`` must be a non-negative BIGINT value. ::
+
+        SELECT map_top_n_keys(MAP(ARRAY[1, 2, 3], ARRAY[10, 20, 30]), 2, (k, v) -> k); -- [3, 2]
+        SELECT map_top_n_keys(MAP(ARRAY[1, 2, 3], ARRAY[10, 20, 30]), 2, (k, v) -> v); -- [3, 2]
+        SELECT map_top_n_keys(MAP(ARRAY[1, 2, 3], ARRAY[10, 20, 30]), 2, (k, v) -> -k); -- [1, 2]
+
+    .. note::
+
+        Due to a signature mismatch between the Java and C++ implementations, certain queries
+        with lambda functions may fail when using Presto Java without sidecar mode enabled.
+        For example, the following query will fail in Presto Java::
+
+            SELECT map_top_n_keys(MAP(ARRAY['cat', 'leopard', 'mouse'], ARRAY[1, 2, 3]), 2, (k, v) -> length(k));
+            -- Error: Unexpected parameters for function map_top_n_keys
+
+        This query works correctly in Velox (C++) but requires sidecar mode to be enabled
+        when running through Presto Java to delegate execution to the Velox engine.
+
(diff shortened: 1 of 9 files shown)


mbasmanova:
@jkhaliqi Make sure to use transform lambda, not a comparator lambda. See discussion in https://github.com/facebookincubator/velox/pull/16048

CC: @allenshen13 @majetideepak @czentgr

jkhaliqi:
@mbasmanova Thanks for the review just to clarify. For the `array_sort` problem there was an issue with sorting `nulls` where the comparator must handle NULLs explicitly to avoid incorrect sorting. Now with, `map_top_n_keys` we can not have `null` elements as key values and also with the [documentation](https://prestodb.io/docs/current/functions/map.html#id0) it states the function `returns -1, 0, or 1 as the first key is less than, equal to, or greater than the second key. If the comparator function returns other values (including NULL), the query will fail and raise an error` which holds true from this example:
```
presto:tpch> WITH input(c0) AS ( VALUES (MAP(ARRAY[3,2,1], ARRAY[1,1,1])) ) SELECT map_top_n_keys(c0, 2, (x, y) -> CAST(IF(x < y, NULL, 1) AS INTEGER)) FROM input;
Query 20260130_180222_00009_ghts2 failed: Lambda comparator must return either -1, 0, or 1
```
Is the change more so for the `efficient way to do that in a vectorized engine` part of the post, or is there another correctness concern I might be missing?

mbasmanova:
@jkhaliqi Thank you for the clarification. Comparator lambdas tend to be verbose and easy to get wrong, and in Velox there isn’t an efficient way to implement them. We’re also aiming for consistency in function signatures and semantics across the codebase, so I’d prefer that we don’t add comparator lambda functions.

aditi-pandit:
@jkhaliqi : Its okay to only have transform lambda versions in Presto C++. We should just document that we will not add the comparator version. 

@mbasmanova : Sorry for the delay in looking at these PRs. I've been swamped lately. 

It would be great if we have some writeup explaining why we don't have Comparator lambda's in Velox. Maybe you wrote something already.. But i couldn't find it. Can you help with phrasing something ? That would be really helpful to users.

mbasmanova:
@aditi-pandit I wrote a blog post some time ago: https://velox-lib.io/blog/array-sort/

@kgpai and @amitkdutta Amit and Krishna also mentioned this in https://prestodb.io/blog/2026/01/22/presto-vs-prestissimo-known-differences-and-workarounds/

jkhaliqi:
```
E20260227 13:26:01.856446 4457452 ExpressionFuzzerVerifier.cpp:220] ==============================> Top 1 by number of rows processed
E20260227 13:26:01.856453 4457452 ExpressionFuzzerVerifier.cpp:222] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260227 13:26:01.856459 4457452 ExpressionFuzzerVerifier.cpp:226] map_top_n_keys 5209 100.00% 483760
E20260227 13:26:01.856468 4457452 ExpressionFuzzerVerifier.cpp:233] ==============================> Bottom 1 by number of rows processed
E20260227 13:26:01.856474 4457452 ExpressionFuzzerVerifier.cpp:235] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260227 13:26:01.856480 4457452 ExpressionFuzzerVerifier.cpp:240] map_top_n_keys 5209 100.00% 483760
E20260227 13:26:01.856487 4457452 ExpressionFuzzerVerifier.cpp:253] ==============================> All stats sorted by number of times the function was chosen
E20260227 13:26:01.856492 4457452 ExpressionFuzzerVerifier.cpp:255] Format: functionName numTimesSelected proportionOfTimesSelected numProcessedRows
E20260227 13:26:01.856498 4457452 ExpressionFuzzerVerifier.cpp:259] map_top_n_keys 5209 100.00% 483760
E20260227 13:26:01.856505 4457452 ExpressionFuzzerVerifier.cpp:488] Total test cases: 7816
E20260227 13:26:01.856510 4457452 ExpressionFuzzerVerifier.cpp:489] Total test cases verified in the reference DB: 7816
E20260227 13:26:01.856516 4457452 ExpressionFuzzerVerifier.cpp:491] Total test cases failed in both Velox and the reference DB: 0
E20260227 13:26:01.856523 4457452 ExpressionFuzzerVerifier.cpp:493] Total test cases unsupported in the reference DB: 0
```

allenshen13 (commented):
For test cases, maybe we can test NULL values in the map or create a lambda that returns null for some entries? Maybe have an n that is greater than map size?

PR #16163, opened 2026-01-29
fix(cudf): Fix some join type does not need AST tree

jinchengchenghh:
After PR https://github.com/facebookincubator/velox/pull/15337, some join types does not require the AST tree, it uses cudf ExpressionEvaluator which supports more expressions to evaluate the expression.

After this PR, TPCDS Q4 can run successfully.

### DIFF
--- a/velox/experimental/cudf/exec/CudfHashJoin.cpp
+++ b/velox/experimental/cudf/exec/CudfHashJoin.cpp
@@ -401,16 +401,19 @@ CudfHashJoinProbe::CudfHashJoinProbe(
     std::vector<PrecomputeInstruction> leftPrecomputeInstructions;
     static constexpr bool kAllowPureAstOnly = true;
     if (joinNode_->isRightJoin() || joinNode_->isRightSemiFilterJoin()) {
-      createAstTree(
-          exprs.exprs()[0],
-          tree_,
-          scalars_,
-          buildType,
-          probeType,
-          rightPrecomputeInstructions,
-          leftPrecomputeInstructions,
-          kAllowPureAstOnly);
-    } else {
+      if (joinNode_->isRightSemiFilterJoin()) {
+        createAstTree(
+            exprs.exprs()[0],
+            tree_,
+            scalars_,
+            buildType,
+            probeType,
+            rightPrecomputeInstructions,
+            leftPrecomputeInstructions,
+            kAllowPureAstOnly);
+      }
+
+    } else if (joinNode_->isAntiJoin() || joinNode_->isLeftSemiFilterJoin()) {
       createAstTree(
           exprs.exprs()[0],
           tree_,
@@ -423,7 +426,7 @@ CudfHashJoinProbe::CudfHashJoinProbe(
     }
     if (leftPrecomputeInstructions.size() > 0 ||
         rightPrecomputeInstructions.size() > 0) {
-      VELOX_NYI("Filters that require precomputation are not yet supported");
+      VELOX_NYI("Filters that require precomputation are not yet supported, join type is {}", core::JoinTypeName::toName(joinNode_->joinType()));
     }
   }
 }


jinchengchenghh:
@shrshi Could you help review this PR? Thanks!

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -401,16 +401,19 @@ CudfHashJoinProbe::CudfHashJoinProbe(
...
+            tree_,
+            scalars_,
+            buildType,
+            probeType,
+            rightPrecomputeInstructions,
+            leftPrecomputeInstructions,
+            kAllowPureAstOnly);
+      }
+
+    } else if (joinNode_->isAntiJoin() || joinNode_->isLeftSemiFilterJoin()) {
```
shrshi, line 416: In this case, the AST is constructed only for semi-filter joins and anti joins. However, I think they conflict with the changes in https://github.com/facebookincubator/velox/pull/16070. In this PR, without materializing the intermediate joined output, the filter is directly applied on the inner/left/right join indices using the AST . 
https://github.com/facebookincubator/velox/blob/ba98ecd49ab0b7a5da8521d0d42e8a9be6be0eec/velox/experimental/cudf/exec/CudfHashJoin.cpp#L665

Can you help me understand why this change is needed for Q4?

jinchengchenghh: Q4 throws exception the `switch` expression is not supported.
I missed this PR https://github.com/facebookincubator/velox/pull/16070/changes#diff-f2c7c33d4cefb3267976ab8973a5ade89ea9cf7fee65c6e97952693766dbe8b5R670, now all the join types require AST tree, This parameter `kAllowPureAstOnly` restrict all the filter can be represented by the AST tree, in Velox + cudf, we also have expression evaluator, the latter can evaluate more expressions by registering the cudf functions. So if the filter cannot be represented by AST tree, the join will fail, if we don't need to construct the `tree_`, we can support more queries. 
https://github.com/facebookincubator/velox/blob/27b3e068b891282cfa5005b311c94a794e0feda6/velox/experimental/cudf/exec/CudfHashJoin.cpp#L412

https://github.com/facebookincubator/velox/blob/27b3e068b891282cfa5005b311c94a794e0feda6/velox/experimental/cudf/expression/ExpressionEvaluator.cpp#L353

Could you elaborate a way to use expression evaluator instead of AST tree?

shrshi: Can we adopt a hybrid approach here where we fallback to the expression evaluator in case the operation is not supported by the AST? My proposal is that we continue using `createAstTree` in `CudfHashJoinProbe` to generate AST plus pre-compute instructions. We then execute the pre-computed instructions to materialize columns before invoking the AST kernel. We currently don't support filters that require pre-computation but I can work on implementing the pre-compute execution logic. Finally, the AST would reference the precomputed columns for unsupported subexpressions. 
This way, we can retain AST’s performance advantage for supported subgraphs by minimizing intermediate table materialization.
cc @karthikeyann

shrshi: @jinchengchenghh pre-computation has been enabled in PR https://github.com/facebookincubator/velox/pull/16212 for filters that need the expression evaluator. Would this help with TPCDS Q4?

jinchengchenghh: Thanks for your efforts! I think it helps TPCDS Q4.

I agree your proposed framework can resolve the expression issue. I have some concerns about the double filter and column abandon and liveness.

For

```sql
WHERE (a + b > 10) AND (unsupported_fn(c) > 0)
```

in a hybrid setup:

* If `unsupported_fn(c) > 0` is precomputed into a temp column `p0`
* And the AST is built without a proper liveness pass

then after the filter:

* `a`, `b`, `c`, and `p0` are all **logically dead**
* But they may still appear as **live inputs** in the AST tree
* AST evaluates p0 twice

This causes:

* Unnecessary materialization of columns
* Extra memory traffic
* In join contexts, potential correctness issues if columns are dropped too early or kept too long

**Key rule to avoid this:**

* Pre-compute should *only* materialize columns, never apply filtering
* After AST rewrite, run a liveness / column-pruning pass
* Columns used only by the filter must be dropped immediately after the filter stage

In short:

> Hybrid AST + precompute is safe only if the AST reflects *true post-filter liveness*, not the original expression tree.

We may accept current optimization and do some further optimization later.

Thank you again!

PR #16165, opened 2026-01-29
fix(macos): Install gflags and glog from source instead of Homebrew

### LINKED ISSUE #16135: Mac failed with build folly with glog and gflag
### Problem description

We I re-execute setup-macos.sh, folly failed with
```
Imported target "Folly::folly" includes non-existent path
    "/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include"
  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
-- Generating done (0.8s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
make[1]: *** [cmake] Error 1
make: *** [release] Error 2
ls /Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include
ls: /Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include: No such file or directory
```
Then I find my SDKs are
```
ls /Library/Developer/CommandLineTools/SDKs/
MacOSX.sdk	MacOSX13.3.sdk	MacOSX13.sdk	MacOSX14.4.sdk	MacOSX15.4.sdk	MacOSX15.sdk	MacOSX26.2.sdk	MacOSX26.sdk

```
Maybe because my mac is upgraded, so the old folly library is cannot used any more, so I try to set up script again, then I face this issue
```
In file included from /Users/chengchengjin/code/spark41/gluten/ep/build-velox/build/velox_ep/deps-download/folly/folly/cli/ProgramOptions.cpp:23:
/opt/homebrew/include/glog/logging.h:60:4: error: <glog/logging.h> was not included correctly. See the documentation for how to consume the library.
   60 | #  error <glog/logging.h> was not included correctly. See the documentation for how to consume the library.
      |    ^
In file included from /Users/chengchengjin/code/spark41/gluten/ep/build-velox/build/velox_ep/deps-download/folly/folly/cli/ProgramOptions.cpp:23:
In file included from /opt/homebrew/include/glog/logging.h:63:
/opt/homebrew/include/glog/flags.h:45:4: error: <glog/flags.h> was not included correctly. See the documentation for how to consume the library.
   45 | #  error <glog/flags.h> was not included correctly. See the documentation for how to consume the library.
      |
```
After remove glog and gflags by 
```
brew uninstall glog
brew uninstall gflags
```
I can build velox, it might because https://github.com/facebookincubator/velox/blob/main/CMake/resolve_dependency_modules/glog/glog-no-export.patch, brew installed glog does not have this patch.

Further more, because folly cannot build by set up script, fizz, wangle cannot be built in environment too.

### System information

Velox System Info v0.0.2
Commit: db7cebe806da815e1a748fce78b6af1223962e9f
CMake Version: 3.31.6
System: Darwin-25.2.0
Arch: arm64
C++ Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
C++ Compiler Version: 17.0.0.17000603
C Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
C Compiler Version: 17.0.0.17000603
CMake Prefix Path: /Library/Developer/CommandLineTools/SDKs/MacOSX26.2.sdk/usr;/opt/homebrew;/usr/local;/usr;/;/Applications/CMake.app/Contents;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local
  

### CMake log

```bash

```

yaooqinn:
## Summary

This PR fixes the macOS build failure caused by Homebrew's glog 0.7.x breaking API changes.

## Problem

Homebrew recently upgraded glog to version 0.7.1 which has breaking API changes - the `GLOG_EXPORT` macro requires CMake target linking which breaks direct header includes. This causes folly compilation to fail with the error:

```
<glog/logging.h> was not included correctly. See the document in the comment of logging.h for detail.
```

The root cause is that Velox pins glog to `v0.6.0` in `setup-versions.sh`, but the macOS setup script uses Homebrew which installs the latest incompatible version.

## Solution

This change:
- Removes `gflags` and `glog` from `MACOS_VELOX_DEPS` (Homebrew packages)
- Adds `install_gflags` function for macOS (similar to other setup scripts)
- Installs both gflags (v2.2.2) and glog (v0.6.0) from source using the pinned versions from `setup-versions.sh`
- Updates the CI workflow to call `install_gflags` and `install_glog`

This approach is consistent with how other Linux setup scripts handle these dependencies.

## Testing

This fix has been validated based on feedback from issue #16135 where contributors confirmed that:
- Uninstalling Homebrew's glog and building from source fixes the issue
- The pinned v0.6.0 version works correctly

Fixes #16135

### DIFF
--- a/scripts/setup-macos.sh
+++ b/scripts/setup-macos.sh
@@ -41,8 +41,9 @@ export OS_CXXFLAGS
 export CMAKE_POLICY_VERSION_MINIMUM="3.5"
 
 DEPENDENCY_DIR=${DEPENDENCY_DIR:-$(pwd)}
-MACOS_VELOX_DEPS="bison flex gflags glog googletest icu4c libevent libsodium lz4 openssl protobuf@21 simdjson snappy xz xxhash zstd"
-
+# gflags and glog are installed from source to ensure version compatibility.
+# Homebrew's glog 0.7.x has breaking API changes that are incompatible with folly.
+MACOS_VELOX_DEPS="bison flex googletest icu4c libevent libsodium lz4 openssl protobuf@21 simdjson snappy xz xxhash zstd"
 MACOS_BUILD_DEPS="ninja cmake"
 
 SUDO="${SUDO:-""}"
@@ -103,6 +104,11 @@ function install_velox_deps_from_brew {
   done
 }
 
+function install_gflags {
+  wget_and_untar https://github.com/gflags/gflags/archive/"${GFLAGS_VERSION}".tar.gz gflags
+  cmake_install_dir gflags -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=ON -DBUILD_gflags_LIB=ON
+}
+
 function install_s3 {
   install_aws_deps
 
@@ -151,7 +157,6 @@ function install_faiss {
     install_faiss_deps
 
     wget_and_untar "https://github.com/facebookresearch/faiss/archive/refs/tags/v${FAISS_VERSION}.tar.gz" faiss
-
     local cmake_args
     cmake_args=(
       -DFAISS_ENABLE_GPU=OFF
@@ -177,6 +182,8 @@ function install_velox_deps {
   run_and_time install_ranges_v3
   run_and_time install_double_conversion
   run_and_time install_re2
+  run_and_time install_gflags
+  run_and_time install_glog
   run_and_time install_boost
   run_and_time install_fmt
   run_and_time install_fast_float
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -74,6 +74,8 @@ jobs:
           source scripts/setup-macos.sh
           install_build_prerequisites
           install_velox_deps_from_brew
+          install_gflags
+          install_glog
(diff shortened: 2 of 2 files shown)


scripts/setup-macos.sh
```diff
@@ -103,6 +104,11 @@ function install_velox_deps_from_brew {
   done
 }
 
+function install_gflags {
+  wget_and_untar https://github.com/gflags/gflags/archive/"${GFLAGS_VERSION}".tar.gz gflags
+  cmake_install_dir gflags -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=ON -DBUILD_gflags_LIB=ON
```
czentgr, line 109: Suggest also setting `BUILD_TESTING=OFF`

czentgr:
Thanks for the quick fix. This explains the CI failures for macOS as well.

PR #16171, opened 2026-01-29
fix(build): Missing typename in Clang 15

czentgr:
Clang 15 doesn’t support the "down with typename" C++20 feature where typename still needs to be explicit for a dependent type in a return type declaration.
This broke the scheduled CI run for Clang 15 and Prestissimo.

PR introducing the dependent type as return type: https://github.com/facebookincubator/velox/pull/16132

### DIFF
--- a/velox/vector/BiasVector-inl.h
+++ b/velox/vector/BiasVector-inl.h
@@ -98,7 +98,8 @@ std::unique_ptr<SimpleVector<uint64_t>> BiasVector<T>::hashAll() const {
 }
 
 template <typename T>
-SimpleVector<T>::TValueAt BiasVector<T>::valueAtFast(vector_size_t idx) const {
+typename SimpleVector<T>::TValueAt BiasVector<T>::valueAtFast(
+    vector_size_t idx) const {
   switch (valueType_) {
     case TypeKind::INTEGER:
       return bias_ + reinterpret_cast<const int32_t*>(rawValues_)[idx];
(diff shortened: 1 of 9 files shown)


czentgr:
> Why do we need clang 15 support?

Because Clang 15 is the officially minimum supported compiler (see Velox readme). With the addition of asan/ubsan and jobs to use newer Clang and GCC we can determine a new minimum version to support. The current supported minimum is Clang 15 and GCC11.
However, we need to decide as community on a new set of minimum versions. For example, Prestissimo uses the same minimums and would need to follow suit.
Note the scheduled Velox CI for Clang 15 that ensures we support the minimum is also currently broken.

MBkkt:
@czentgr I know but the updating the minimum supported version of compilers was discussed about six months ago. Quite sad that there's no progress on this.

czentgr:
@MBkkt I'm only aware of moving to newer compilers in CI and overall but not necessarily also changing the minimum officially. What was the outcome of the discussion? I suppose it didn't result in an action item on the OSS side. But it is certainly something we are looking at now that we are moving up to newer compilers.

MBkkt:
@czentgr 

> changing the minimum officially

If minimum supported compiler isn't presented in CI it will be broken constantly

czentgr:
@MBkkt This was a trade off to not have too many jobs running for a job. The consensus was to do a scheduled run instead. Usually, the CI isn't broken. In the last year we have maybe two or three times when it got broken. Given the amount of commits since it has not been a problem. Just this one is one of the occurrences.

PR #16175, opened 2026-01-30
feat: Add hive index reader for Nimble file format

xiaoxmeng:
Summary:
This diff introduces HiveIndexReader, a new component that enables index-based lookup functionality for Hive tables stored in Nimble file format. The reader provides efficient point lookups by leveraging Nimble's cluster indexing capabilities.

HiveIndexReader: New class that wraps Nimble's SelectiveNimbleReader for index-based reads. Supports:
Setting probe data via setRequest() for index lookups
Iterating through matching rows with hasNext() and next()
Integration with AsyncDataCache for caching
IoStatistics tracking for monitoring cache hits

HiveTableHandle extension: Added indexColumns parameter to specify indexed columns, along with new APIs:
supportsIndexLookup() - Returns true if table has index columns configured
needsIndexSplit() - Returns true (index lookups require split info)
indexColumns() - Returns the list of indexed column names

SelectiveNimbleReader improvements:
Added reset() method to reuse the reader for multiple index lookups
Separated initIndex() (one-time setup) from setIndexBounds() (per-query setup)
Added split stripe boundary tracking for proper reset behavior

Test utilities: Renamed and enhanced TabletIndexTestUtils to IndexTestUtils with additional helpers for generating test data and writing indexed Nimble files.

The followup is to support skip list to handle batch lookup request in one pass of selective reader scan

Differential Revision: D91415139

### DIFF
--- a/velox/connectors/Connector.h
+++ b/velox/connectors/Connector.h
@@ -342,26 +342,26 @@ class IndexSource {
   }
 
   /// Represents a lookup request for a given input.
-  struct LookupRequest {
+  struct Request {
     /// Contains the input column vectors used by lookup join and range
     /// conditions.
     RowVectorPtr input;
 
-    explicit LookupRequest(RowVectorPtr input) : input(std::move(input)) {}
+    explicit Request(RowVectorPtr input) : input(std::move(input)) {}
   };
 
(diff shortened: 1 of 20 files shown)


HuamengJiang (approved):
Comments on internal diff.

PR #8650, opened 2024-02-02
Create utility file for encoding

Joe-Abraham:
Add a new utility file which will share the common logics for decoding and encoding based upon the base variants.

- Refactoring base64
- New presto functions to_base32 and from_base32

Will be raised as subsequent PRs.

PR #16176, opened 2026-01-30
fix: Add new utility file BaseEncoderUtils.h 

allenshen13:
### Summary
Extracts common encoding/decoding utilities shared between Base64 and Base32 into a new BaseEncoderUtils class, as proposed in #8650 and addressing reviewer feedback from that PR. Credit to @Joe-Abraham 

This PR refactors shared Base64/Base32 encoding utilities into BaseEncoderUtils. It should be reviewed and merged before #16235, which builds on top of it. The two PRs share the same commit history — this PR contains only the refactor commit, while #16235 adds the Base32 encoding feature on top.

## Changes

**`BaseEncoderUtils.h` (new)** — Shared utility class with:
- `isPadded`, `numPadding` — padding detection
- `checkForwardIndex`, `checkReverseIndex` — compile-time charset/reverse-index validation
- `reverseLookup` — character-to-value mapping with error handling
- `calculateEncodedSize`, `calculateDecodedSize` — generic size computation

**`Base64.h/.cpp`** — Refactored to delegate to `BaseEncoderUtils` for shared logic (no behavioral change).

### DIFF
--- a/velox/common/encode/Base32.h
+++ b/velox/common/encode/Base32.h
@@ -52,13 +52,10 @@ class Base32 {
       const char* input,
       const size_t inputSize);
 
-  // Padding character used in encoding.
-  static const char kPadding = '=';
-
   // Constants defining the size in bytes of binary and encoded blocks for
-  // Base32 encoding. Size of a binary block in bytes (5 bytes = 40 bits)
+  // Base32 encoding. Size of a binary block in bytes (5 bytes = 40 bits).
   static const int kBinaryBlockByteSize = 5;
-  // Size of an encoded block in bytes (8 bytes = 40 bits)
+  // Size of an encoded block in bytes (8 bytes = 40 bits).
   static const int kEncodedBlockByteSize = 8;
 
  private:
--- a/velox/common/encode/BaseEncoderUtils.h
+++ b/velox/common/encode/BaseEncoderUtils.h
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <string_view>
+
+#include "velox/common/base/Status.h"
+
+namespace facebook::velox::encoding {
+
+/// Describes the binary and encoded block sizes for a base encoding codec.
+struct CodecBlockSizes {
+  int binaryBlockByteSize;
+  int encodedBlockByteSize;
+};
+
+static constexpr CodecBlockSizes kBase64BlockSizes{3, 4};
+static constexpr CodecBlockSizes kBase32BlockSizes{5, 8};
+
+/// Shared utility functions for base encoding schemes (e.g. Base64, Base32).
+/// Provides common operations for padding detection, charset validation,
+/// reverse lookup, and encoded size calculation.
+namespace utils {
+
+/// Padding character used in base encoding schemes.
+inline constexpr char kPadding = '=';
+
+/// Sentinel value in reverse index tables indicating an invalid character.
+inline constexpr uint8_t kInvalidBase = 255;
+
+/// Error message for invalid encoded input string length.
+inline constexpr std::string_view kInvalidInputLengthError =
+    "decode() - invalid input string length.";
+
+/// Checks if the encoded input ends with padding character(s).
+inline bool isPadded(const char* input, size_t inputSize) {
+  return inputSize > 0 && input[inputSize - 1] == kPadding;
+}
+
+/// Counts the number of trailing padding characters in encoded input.
+inline size_t numPadding(const char* input, size_t inputSize) {
+  size_t count{0};
+  while (count < inputSize && input[inputSize - 1 - count] == kPadding) {
+    ++count;
+  }
+  return count;
+}
+
+/// Validates that each character in the charset maps correctly through the
+/// reverse index table back to its original index.
+/// @param index The current index to validate (iterates down to 0).
+/// @param charset The character set used for encoding.
+/// @param reverseIndex The reverse lookup table mapping characters to
+/// indices.
+template <typename Charset, typename ReverseIndex>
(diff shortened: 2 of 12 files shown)


velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,211 @@
...
+  /// Sentinel value in reverse index tables indicating an invalid character.
+  static constexpr uint8_t kInvalidBase = 255;
+
+  /// Checks if the encoded input ends with padding character(s).
+  static bool isPadded(std::string_view input) {
+    return !input.empty() && input.back() == kPadding;
+  }
+
+  /// Counts the number of trailing padding characters in encoded input.
+  static size_t numPadding(std::string_view input) {
```
Joe-Abraham: In my opinion, it is not logical to modify the input argument to `std::string_view` if the input type is not changed from `const char*` to `std::string_view` throughout the entire flow. In the current code, there are numerous conversions between `const char*` and `std::string_view`. Let me know what you think.

allenshen13: Yep, that makes sense. I have pushed the changes, let me know what you think.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,209 @@
...
+  /// @param inputSize The length of the encoded input. Modified on return to
+  ///   exclude padding characters.
+  /// @param binaryBlockByteSize Number of bytes per binary block
+  ///   (3 for Base64, 5 for Base32).
+  /// @param encodedBlockByteSize Number of characters per encoded block
+  ///   (4 for Base64, 8 for Base32).
+  static Expected<size_t> calculateDecodedSize(
+      const char* input,
+      size_t& inputSize,
+      int binaryBlockByteSize,
```
czentgr: For this function and calculateEncodedSize we want to check the input.
input not nullptr, binaryBlockByteSize and encodedBlockByteSize are greater than 0.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,209 @@
...
+      return 0;
+    }
+
+    // Check if the input string is padded.
+    if (isPadded(input, inputSize)) {
+      // If padded, ensure that the string length is a multiple of the encoded
+      // block size.
+      if (inputSize % encodedBlockByteSize != 0) {
+        return folly::makeUnexpected(
+            Status::UserError("decode() - invalid input string length."));
```
czentgr: The string is used multiple times, so it could be a static const char*.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,209 @@
...
+  /// Performs reverse lookup of an encoded character to its numeric value.
+  /// @param encodedChar The encoded character to look up.
+  /// @param reverseIndex The reverse index table for the encoding scheme.
+  /// @param base The base of the encoding (e.g. 64 for Base64, 32 for
+  ///   Base32). Characters mapping to values >= base are considered invalid.
+  template <typename ReverseIndex>
+  static Expected<uint8_t> reverseLookup(
+      char encodedChar,
+      const ReverseIndex& reverseIndex,
+      uint8_t base) {
```
czentgr: charsetSize instead of base?

czentgr (commented):
The other PR also has these changes to show what they look like but we need to make an order on the merge.

I see that the other PR was updated to have the code reviewed BaseEncoderUtils.

What about this PR?

allenshen13:
@czentgr The BaseEncoderUtils changes in this PR are stale — the reviewed/updated version lives in #16235 which includes all the feedback from both PRs. Should I keep both as separate PRs and update this PR or close this one and address your latest comments in the other PR.

velox/common/encode/BaseEncoderUtils.h
allenshen13: Added VELOX_CHECK_NOT_NULL(input) and VELOX_CHECK_GT for both block sizes.

velox/common/encode/BaseEncoderUtils.h
allenshen13: Added kInvalidInputLengthError constant to BaseEncoderUtils.h. Is there a benefit to using static const char* instead of std:string_view?

velox/common/encode/BaseEncoderUtils.h
allenshen13: Yep, that is better.

czentgr:
Okay, that's why I confused. I had a previous message to kee the updated new encoder util here and update them in the https://github.com/facebookincubator/velox/pull/16235 PR. But in THIS oPr there are also base32 changes. 
We don't want them here. Instead, the changes from this PR for a commit in the base32 PR to how how the utils are re-used.
But no base32 changes should be here.

velox/common/encode/Base32.h
```diff
@@ -22,6 +22,7 @@
 #include <string>
 
 #include "velox/common/base/Status.h"
+#include "velox/common/encode/BaseEncoderUtils.h"
```
czentgr: Can we move this to the cpp? Same for the Base64 one.
Nothing from the utils is used in the header, right?

czentgr (commented):
@majetideepak Please also take a look.

velox/common/encode/Base32.h
allenshen13: Yep, nothing from utils is used in .h files

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,158 @@
...
+  int encodedBlockByteSize;
+};
+
+static constexpr CodecBlockSizes kBase64BlockSizes{3, 4};
+static constexpr CodecBlockSizes kBase32BlockSizes{5, 8};
+
+/// Shared utility functions for base encoding schemes (e.g. Base64, Base32).
+/// Provides common operations for padding detection, charset validation,
+/// reverse lookup, and encoded size calculation.
+class BaseEncoderUtils {
```
czentgr: Instead of making this a class, can we add a namespace like velox::encoding:utils
and then group the functions in it and replace the class itself?
These are all public functions and the class doesn't encapsulate data/object.

I think this was one of the previous review comments.

allenshen13:
Re: Base32 changes in this PR
I wanted to clarify the rationale for including the Base32 changes here rather than deferring them to https://github.com/facebookincubator/velox/pull/16235.

This PR's purpose is to extract shared encoding utilities into encoding::utils. The Base32 changes here are purely mechanical migrations — replacing inline code (e.g., Base32::kPadding, hardcoded 32) with calls to the new shared utilities (utils::kPadding, kCharsetSize). No new Base32 functionality is added.

czentgr (approved):
@Yuhta If you get a chance perhaps you can review this PR please?

allenshen13:
Superseded: Base32::encode() landed upstream independently, so the shared-utils refactor is no longer needed. to_base32 is now a standalone function PR (#16235).

Issue #16179, opened 2026-01-30
[cuDF] Join Filters do not check that the functions they use are registered


simoneves:
### Bug description

e.g. TPCH Q7 and Q19 where there is a filter with nested boolean operations: WHERE a AND b AND (c OR d)

We get "Unsupported expression for recursive evaluation: and" (or "or")

### System information

Velox System Info v0.0.2
Commit: 59b492a9bce45f487f24b5cbae7dc845ea3d0827
CMake Version: 4.2.1
System: Linux-6.14.0-37-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 13.3.0
13.3.0
C Compiler: /usr/bin/cc
C Compiler Version: 13.3.0
13.3.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

### Relevant logs

```bash

```

devavret:
Fixed by https://github.com/facebookincubator/velox/pull/16180

PR #16180, opened 2026-01-30
fix(cudf): Check that join filter expressions can be evaluated by CUDF

simoneves:
Per https://github.com/facebookincubator/velox/issues/16179

CUDF join evaluation must also include their filters

### DIFF
--- a/velox/experimental/cudf/exec/ToCudf.cpp
+++ b/velox/experimental/cudf/exec/ToCudf.cpp
@@ -172,7 +172,7 @@ bool CompileState::compile(bool allowCpuFallback) {
     return true;
   };
 
-  auto isJoinSupported = [getPlanNode](const exec::Operator* op) {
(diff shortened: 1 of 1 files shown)


devavret (approved):

karthikeyann:
@simoneves are all TPCH benchmarks running with Presto GPU and Velox cudf tpch benchmark?

simoneves:
> @simoneves are all TPCH benchmarks running with Presto GPU and Velox cudf tpch benchmark?

Benchmark behavior is unchanged. TPCH Q15 still fails in the simple "integrity tests" because it outputs a row that fails some floating-point comparison. I will double-check pure Velox. Not sure why this particular issue didn't show up earlier, but it was definitely failing in my DECIMAL dev branch with Q7 and Q19 which have nested boolean operations.

simoneves:
@karthikeyann pure Velox is fine too, at least in terms of the tests our CI currently runs

PR #16181, opened 2026-01-31
feat: Improve textReader performance

### LINKED ISSUE #16010: TextReader is inefficient and parses bytes char by char
### Description

The textreader currently is too slow, performance of the textReader can be improved by 
1. Line by line parsing
2. Using Precompiled type switching
3. optimized buffer copy to result directly

jaystarshot:
Fixes #16010

```
Before
[...]/tests/reader/TextReaderBenchmark.cpp     relative  time/iter   iters/s
SyntheticFile_BatchSize_1k                                  24.75s    40.40m
SyntheticFile_BatchSize_10k                     100.71%     24.58s    40.69m
SyntheticFile_BatchSize_50k                     101.30%     24.43s    40.93m
SyntheticFile_AllColumns                                    24.82s    40.29m
SyntheticFile_HalfColumns                       141.66%     17.52s    57.08m
SyntheticFile_SingleColumn                      215.40%     11.52s    86.80m

```

```
WARNING: Benchmark running in DEBUG mode
[...]/tests/reader/TextReaderBenchmark.cpp     relative  time/iter   iters/s
SyntheticFile_BatchSize_1k                                  14.26s    70.12m
SyntheticFile_BatchSize_10k                     99.399%     14.35s    69.70m
SyntheticFile_BatchSize_50k                     98.298%     14.51s    68.93m
SyntheticFile_AllColumns                                    14.70s    68.01m
SyntheticFile_HalfColumns                       217.90%      6.75s   148.20m
SyntheticFile_SingleColumn                      2140.2%   686.98ms      1.46
WARNING: Benchmark running in DEBUG mode
```

**Timestamp semantics also changed (timezone conversion removed).**
The new TIMESTAMP setter now writes the parsed value directly, without converting to the default timezone. Previously, TEXT timestamps were adjusted via toGMT(defaultTimezone), so this shifts every naive timestamp by the local offset.

**Map values treat empty string as NULL for all types.**

**zero‑copy line parsing**: The reader now scans for \n in stream buffers and passes std::string_view slices directly to setters, instead of byte‑by‑byte parsing and building strings.

**Precompiled setters:** Type dispatch happens once in initializeColumnSetters(); per‑field parsing calls the compiled function directly.

**Simplified row loop:** No nested delimiter state machine; processLine() splits fields with findDelimiter() and writes directly.

**Better I/O efficiency:** Uses a 1MB SeekableFileInputStream buffer for uncompressed text, reducing read overhead.

**Less parsing overhead for primitives:** Numeric parsing is centralized with parseIntegerWithDecimal, and booleans use a fast case‑insensitive check.

### DIFF
--- a/velox/dwio/text/reader/TextReader.cpp
+++ b/velox/dwio/text/reader/TextReader.cpp
@@ -15,109 +15,70 @@
  */
 
 #include "velox/dwio/text/reader/TextReader.h"
-
-#include <boost/algorithm/string/predicate.hpp>
-#include <string>
-
+#include <glog/logging.h>
 #include "velox/common/encode/Base64.h"
 #include "velox/dwio/common/exception/Exceptions.h"
+#include "velox/type/Conversions.h"
+#include "velox/type/TimestampConversion.h"
 #include "velox/type/fbhive/HiveTypeParser.h"
 
 namespace facebook::velox::text {
-namespace {
 
 using common::CompressionKind;
-
-using dwio::common::EOFError;
 using dwio::common::RowReader;
 using dwio::common::verify;
+using folly::AsciiCaseInsensitive;
+using folly::StringPiece;
 
-static constexpr std::string_view kTextfileCompressionExtensionGzip{".gz"};
-static constexpr std::string_view kTextfileCompressionExtensionDeflate{
-    ".deflate"};
-static constexpr std::string_view kTextfileCompressionExtensionZst{".zst"};
-static constexpr std::string_view kTextfileCompressionExtensionLz4{".lz4"};
-static constexpr std::string_view kTextfileCompressionExtensionLzo{".lzo"};
-static constexpr std::string_view kTextfileCompressionExtensionSnappy{
-    ".snappy"};
+constexpr const char* kTextfileCompressionExtensionGzip = ".gz";
+constexpr const char* kTextfileCompressionExtensionDeflate = ".deflate";
+constexpr const char* kTextfileCompressionExtensionZst = ".zst";
+constexpr const char* kTextfileCompressionExtensionLz4 = ".lz4";
+constexpr const char* kTextfileCompressionExtensionLzo = ".lzo";
+constexpr const char* kTextfileCompressionExtensionSnappy = ".snappy";
 
-static std::string emptyString = std::string();
-
+namespace {
 constexpr const int32_t kDecompressionBufferFactor = 3;
 
-void resizeVector(
-    BaseVector* FOLLY_NULLABLE data,
-    const vector_size_t insertionIdx) {
-  if (data == nullptr) {
-    return;
-  }
-
-  auto dataSize = data->size();
-  if (dataSize == 0) {
-    data->resize(10);
-  } else if (dataSize <= insertionIdx) {
-    if (data->type()->kind() == TypeKind::ARRAY) {
-      auto oldSize = dataSize;
-      auto newSize = dataSize * 2;
-      data->resize(newSize);
-
-      auto arrayVector = data->asChecked<ArrayVector>();
-      auto rawOffsets = arrayVector->offsets()->asMutable<vector_size_t>();
-      auto rawSizes = arrayVector->sizes()->asMutable<vector_size_t>();
-
-      auto lastOffset = oldSize > 0 ? rawOffsets[oldSize - 1] : 0;
-      auto lastSize = oldSize > 0 ? rawSizes[oldSize - 1] : 0;
-      auto newOffset = oldSize > 0 ? lastOffset + lastSize : 0;
-
-      for (auto i = oldSize; i < newSize; ++i) {
-        rawSizes[i] = 0;
-        rawOffsets[i] = newOffset;
-      }
-    } else if (data->type()->kind() == TypeKind::MAP) {
-      auto oldSize = dataSize;
-      auto newSize = dataSize * 2;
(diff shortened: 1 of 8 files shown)


jaystarshot:
cc: @skadilover  @mbasmanova (needed to redesign current one)

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -209,1333 +146,943 @@ TextRowReader::TextRowReader(
...
 }
 
-uint64_t TextRowReader::next(
-    uint64_t rows,
-    VectorPtr& result,
-    const Mutation* mutation) {
-  if (atEOF_) {
-    return 0;
+template <typename T>
+static std::optional<T> parseIntegerWithDecimal(std::string_view str) {
```
MBkkt, line 191: This function is ineffective consider function from this PR: https://github.com/facebookincubator/velox/pull/16128

MBkkt:
@pashandor789 can you measure this on clickbench tsv (non-generated dataset)?

pashandor789:
Can you take this Code?
https://github.com/facebookincubator/velox/pull/16128
We ve got real Time improvement by +25%

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -15,111 +15,72 @@
...
+    compressionOptions.format.zlib.windowBits = -15;
+  } else if (endsWith(filename, kTextfileCompressionExtensionZst)) {
     kind = CompressionKind::CompressionKind_ZSTD;
   } else {
     kind = CompressionKind::CompressionKind_NONE;
   }
 }
 
+// Parse boolean using folly's case-insensitive comparison
+std::optional<bool> parseBoolean(std::string_view value) {
```
MBkkt, line 73: This function is ineffective consider function from this PR: #16128

jaystarshot:
@pashandor789 @MBkkt  thanks for the quick review are you able to do your benchmark here and report perf? https://github.com/facebookincubator/velox/pull/16128#discussion_r2730225096
I have the same issue as zac here (coming from legacy presto) for ints but maybe we can use your function for non ints?

Regarding the boolean, i have changed to directly parse the string itself so i think its a bit different now?

jaystarshot:
i used your fast float function for REAL , float etc
(my benchmark was using  

{"bool_col", BOOLEAN()},
      {"int_col", INTEGER()},
      {"bigint_col", BIGINT()},
      {"real_col", REAL()},
      {"double_col", DOUBLE()},
      {"timestamp_col", TIMESTAMP()},
      {"string_col", VARCHAR()},
      {"binary_col", VARBINARY()},**

```
WARNING: Benchmark running in DEBUG mode
[...]/tests/reader/TextReaderBenchmark.cpp     relative  time/iter   iters/s
SyntheticFile_BatchSize_1k                                   6.86s   145.84m
SyntheticFile_BatchSize_10k                     100.88%      6.80s   147.13m
SyntheticFile_BatchSize_50k                     99.008%      6.93s   144.39m
SyntheticFile_AllColumns                                     6.67s   149.89m
SyntheticFile_HalfColumns                       184.16%      3.62s   276.04m
SyntheticFile_SingleColumn                      1366.5%   488.22ms      2.05
```

I see a lot more improvement

MBkkt:
@jaystarshot Hi, thanks for working on this, we have similar thoughts about what needs to be done.

We can use fast_float for integers, and still have this unusual decimal => integer logic.
It's possible as Pavel mentioned in his PR.

About float parsing yes it can be used and if folly doesn't use it, it should speedup float/double parsing by a lot

MBkkt:
@jaystarshot I think we can agreed on process this PR first, and then Pavel will "rebase" his changes regarding fast_float code for integers/etc

velox/dwio/text/reader/TextReader.cpp
```diff
@@ -209,1335 +175,972 @@ TextRowReader::TextRowReader(
...
+      fileType->toString(),
+      reqType->toString());
 }
 
-uint64_t TextRowReader::seekToRow(uint64_t rowNumber) {
-  VELOX_CHECK_GT(
-      rowNumber, currentRow_, "Text file cannot seek to earlier row");
+bool TextRowReader::loadBuffer() {
+  // Get next chunk from stream - ZERO COPY, just get pointer
+  const void* data = nullptr;
```
skadilover, line 613: I'm very curious about the percentage of each of the new optimizations involved, for example, what is the percentage of performance improvement from introducing zero-copy?

xin-zhang2:
@jaystarshot Thanks for working on this! I'm following this PR since we're running some benchmarks on the text reader and it would be very helpful. Just wondering if there have been any updates.

aditi-pandit:
@jaystarshot : Loading from text files is a very important part of  the TPC-DS benchmark.. So these optimizations are critical to IBM as well. Do you plan to work on this further ?

jaystarshot:
Hi @aditi-pandit This one and my other PRs (in parquet) were waiting for review and I have now lost all context and don't have time to follow up.
I am okay with someone else picking this and others up (can add coauthor if my work is useful)

jja725:
I can pick this up and coauthor with @jaystarshot

jaystarshot:
@jja725 Another improvement in our internal PR is to remove filtering from the text reader as filtering at text reader is anyway not beneficial and needs a memory copy so that memory copy can be eliminated, you can add that too

PR #16186, opened 2026-01-31
feat: Improve PlanConsistencyChecker error messages

mbasmanova:
Summary:
Add ExceptionContextSetter to each visit method in PlanConsistencyChecker to show plan node details when a consistency check fails. This helps identify which plan node caused the failure.

Example error message with these changes (the Context field includes the details of the plan node that failed validation):

```
Error Code: INVALID_ARGUMENT
Reason: Wrong type of input column: b, BOOLEAN vs. INTEGER
Retriable: False
Expression: *type == *expectedType
Context: Plan node: -- Filter[3][expression: "b"] -> a:BOOLEAN, b:INTEGER, c:DOUBLE

Function: checkInputs
File: fbcode/velox/core/PlanConsistencyChecker.cpp
Line: 347
Stack trace:
Stack trace has been disabled. Use --velox_exception_user_stacktrace_enabled=true to enable it.
```

Differential Revision: D91978791

### DIFF
--- a/velox/core/PlanConsistencyChecker.cpp
+++ b/velox/core/PlanConsistencyChecker.cpp
@@ -15,10 +15,17 @@
  */
 
 #include "velox/core/PlanConsistencyChecker.h"
+#include "velox/common/base/Exceptions.h"
 
 namespace facebook::velox::core {
 
 namespace {
+
+// Returns a message describing the plan node for exception context.
+std::string planNodeMessage(VeloxException::Type /*exceptionType*/, void* arg) {
+  auto* node = static_cast<const PlanNode*>(arg);
+  return fmt::format("Plan node: {}", node->toString(/*detailed=*/true));
+}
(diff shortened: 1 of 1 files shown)


velox/core/PlanConsistencyChecker.cpp
```diff
@@ -15,11 +15,18 @@
...
 
 namespace facebook::velox::core {
 
 namespace {
 
+// Returns a message describing the plan node for exception context.
+std::string planNodeMessage(VeloxException::Type /*exceptionType*/, void* arg) {
+  auto* node = static_cast<const PlanNode*>(arg);
+  return fmt::format("Plan node: {}", node->toString(/*detailed=*/true));
+}
```
srsuryadev, line 28: :nit probably static private

amitkdutta (approved):

PR #16190, opened 2026-02-01
fix(TopNRowNumber): Rank with peer computation

aditi-pandit:
When decrementing rank values when outputting to memory, the ranks drop by the number of peers of the lower rank row than the higher one. 

Also fix the computation of numTopRankRows to avoid using the priority queue container vector as that does not necessarily maintain the order of ranks.

### DIFF
--- a/velox/exec/TopNRowNumber.cpp
+++ b/velox/exec/TopNRowNumber.cpp
@@ -578,7 +578,7 @@ TopNRowNumber::TopRows* TopNRowNumber::nextPartition() {
 
 template <core::TopNRowNumberNode::RankFunction TRank>
 void TopNRowNumber::computeNextRankInMemory(
-    const TopRows& partition,
+    TopRows& partition,
     vector_size_t outputIndex) {
   if constexpr (TRank == core::TopNRowNumberNode::RankFunction::kRowNumber) {
     nextRank_ -= 1;
@@ -587,21 +587,19 @@ void TopNRowNumber::computeNextRankInMemory(
 
   // This is the logic for rank() and dense_rank().
   // If the next row is a peer of the current one, then the rank remains the
-  // same, but the number of peers is incremented.
+  // same.
   if (comparator_.compare(outputRows_[outputIndex], partition.rows.top()) ==
       0) {
-    numPeers_ += 1;
     return;
   }
 
   // The new row is not a peer of the current one. So dense_rank drops the
-  // rank by 1, but rank drops it by the number of peers (which is then
-  // reset).
+  // rank by 1, but rank drops by the number of peers of the new top
+  // row (new rank) in TopRows queue.
   if constexpr (TRank == core::TopNRowNumberNode::RankFunction::kDenseRank) {
     nextRank_ -= 1;
   } else {
-    nextRank_ -= numPeers_;
-    numPeers_ = 1;
+    nextRank_ -= partition.numTopRankRows();
   }
 }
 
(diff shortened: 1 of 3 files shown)


velox/exec/TopNRowNumber.cpp
```diff
@@ -578,7 +578,7 @@ TopNRowNumber::TopRows* TopNRowNumber::nextPartition() {
 
 template <core::TopNRowNumberNode::RankFunction TRank>
 void TopNRowNumber::computeNextRankInMemory(
-    const TopRows& partition,
+    TopRows& partition,
```
pedroerp, line 581: why can't this be const anymore? I don't see where you modify this below.

aditi-pandit: This function now uses partition.numTopRankRows() which needs non const variable.

Ideally we should be able to do this logic without mutable structures, but priority_queue doesn't really offer methods beyond top of queue member access. So the only option was to call pop() and check top() and then reinsert elements back.

velox/exec/TopNRowNumber.cpp
```diff
@@ -1110,18 +1108,29 @@ char* TopNRowNumber::TopRows::removeTopRankRows() {
 
 vector_size_t TopNRowNumber::TopRows::numTopRankRows() {
   VELOX_CHECK(!rows.empty());
+
+  tempTopRankRows.clear();
```
xiaoxmeng, line 1113: shall we have SCOPED_EXIT here to clear tempTopRankRows on exit?

velox/exec/TopNRowNumber.cpp
```diff
@@ -587,21 +587,19 @@ void TopNRowNumber::computeNextRankInMemory(
 
   // This is the logic for rank() and dense_rank().
   // If the next row is a peer of the current one, then the rank remains the
-  // same, but the number of peers is incremented.
+  // same.
```
xiaoxmeng, line 590: Does the peer counting has bug? Why we can't use the current tracking and it seems that numTopRankRows is a bit expensive than explicitly tracking the peers? thanks!

aditi-pandit: @xiaoxmeng : 
The peer counting bug is the fix at line 602.

To give an example:
So say the input had sort keys : 1, 2, 2, 3. 
The ranks for these should be: 1, 2, 2, 4

When emitting output rows, we are going from top of the queue to lower. We know that the top rank found so far is 4. 
So first row output is (3, 4). Peers of 3 = 1

We need to compute the rank of the new top of the queue (which is 2). Since 2 != 3 we know its a new rank value. All rows with sort key 2 will have the same rank, and that rank value its incremented by the number of peers when computing the rank of the current row which is 4. As there are 2 peers of 2 then their rank should be 4 - 2 = 2.
So next output is (2, 2)

The new row at top of queue is 2. As its not a new sort key the rank remains the same. So next output is (2, 2). Peers of 2 = 2

The new top of queue is 1. As 1 != 2 then its a new rank. The new rank is incremented by the number of its peers to obtain the current rank 2. As number of peers is 1, so new rank is 1.

If we were counting peers of current row (old code) then we would output
1, 3, 3, 4 which is incorrect. 

We can't keep all output rows until we see next peer value when outputting, so finding number of top rank rows is the best approach.

We end up doing the numTopRankRows only once for each distinct sort key (not each row) so its not very expensive actually.

velox/exec/TopNRowNumber.cpp
aditi-pandit: Done.

xiaoxmeng (approved):
@aditi-pandit thanks for the explanation!

PR #16194, opened 2026-02-02
misc: Optimize Spark dayname function

philo-he:
This PR simplifies the code path. The core calculation is extracted from
`getDateTime` with unrelated calculation removed.

### DIFF
--- a/velox/functions/sparksql/DateTimeFunctions.h
+++ b/velox/functions/sparksql/DateTimeFunctions.h
@@ -794,7 +794,7 @@ struct DayNameFunction {
     if (dayOfWeek < 0) {
       dayOfWeek += 7;
     }
-    result.setNoCopy(StringView(kDayNames[dayOfWeek]));
+    result.append(StringView(kDayNames[dayOfWeek]));
   }
 };
 
--- a/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
+++ b/velox/functions/sparksql/tests/DateTimeFunctionsTest.cpp
@@ -665,6 +665,8 @@ TEST_F(DateTimeFunctionsTest, dayNameDate) {
 
   EXPECT_EQ(std::nullopt, dayName(std::nullopt));
   EXPECT_EQ("Thu", dayName(0)); // 1970-01-01
+  EXPECT_EQ("Wed", dayName(-1)); // 1969-12-31
+  EXPECT_EQ("Tue", dayName(-2)); // 1969-12-30
   EXPECT_EQ("Sun", dayName(parseDate("2023-08-20")));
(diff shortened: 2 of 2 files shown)


philo-he:
@yaooqinn, thanks for the review.
@zhli1142015, could you also take a look? This is a small optimization.

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -783,33 +783,18 @@ template <typename T>
 struct DayNameFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
+  static constexpr std::string_view kDayNames[] =
```
jinchengchenghh, line 786: Move to private scope

velox/functions/sparksql/DateTimeFunctions.h
```diff
@@ -783,33 +783,18 @@ template <typename T>
...
-        result.append("Thu");
-        break;
-      case 5:
-        result.append("Fri");
-        break;
-      default:
-        result.append("Sat");
-        break;
+    // Based on the fact that the Unix epoch is Thursday.
+    auto dayOfWeek = (4 + date) % 7;
```
jinchengchenghh, line 793: Could you add more tests to cover the upper bound and lower bound.


Currently, we cannot throw the exception. And what if the input argument is a negative value?

```
  VELOX_USER_CHECK(
      Timestamp::epochToCalendarUtc(seconds, dateTime),
      "Date is too large: {} days",
      days);
```

velox/functions/sparksql/DateTimeFunctions.h
philo-he: Given that it's static constexpr, maybe we can keep it public. Thanks.

velox/functions/sparksql/DateTimeFunctions.h
philo-he: Thanks for the comment. This exception comes from an internal check that ensures the correct calculation for year, but not related to the calculation for day name here. So I assume we don't need to throw exceptions when date is out of some range.

BTW, I noticed in Spark the following code was called internally for implementing this function, where we can see the logic is quite same without any limits on the date range.

https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/src/java.base/share/classes/java/time/LocalDate.java#L831

In addition, I have added two tests to cover negative values. Thanks.

philo-he:
@jinchengchenghh, could you take a look again? Thanks.

rui-mo:
> 2 - velox_base_test (Timeout)

The workflow failure is irrelevant to this change.

PR #16195, opened 2026-02-02
feat: Add IPADDRESS support to Arbitrary aggregate function

pratikpugalia:
Summary:
The arbitrary() aggregate function was throwing NOT_IMPLEMENTED when used
with IPADDRESS type. This was caught by the aggregation fuzzer.

This change adds a check for isIPAddressType() for HUGEINT type in
ArbitraryAggregate, allowing IPADDRESS values to be properly handled.

Differential Revision: D92026533


kgpai:
Can you validate that this is implemented in Presto ?

kgpai (approved):

PR #16206, opened 2026-02-02
feat(cursor): Integrate debug cursor into TaskCursor

pedroerp:
Summary:
Integrating debugger cursor into the TaskCursor API, hiding its
implementation details, and exposing a single "breakpoint" APIs to users.

Differential Revision: D92091567


velox/exec/Cursor.cpp
```diff
@@ -491,10 +502,91 @@ class SingleThreadedTaskCursor : public TaskCursorBase {
...
+    queryCtx_->setTraceCtxProvider(
+        [&](core::QueryCtx&, const core::PlanFragment&) {
+          return std::make_unique<TaskDebuggerTraceCtx>(
+              params.breakpoints, traceState_);
+        });
+
+    task_ = Task::create(
+        taskId_,
+        std::move(planFragment_),
+        0,
```
duanmeng: `/*destination*/ 0`?

velox/exec/Cursor.cpp
```diff
@@ -507,6 +599,146 @@ class SingleThreadedTaskCursor : public TaskCursorBase {
...
+  /// (which produces intermediate results) and the cursor (which consumes
+  /// them).
+  struct TraceState {
+    /// Promise used to signal the tracer to continue after a partial result
+    /// has been consumed.
+    ContinuePromise tracePromise{ContinuePromise::makeEmpty()};
+
+    /// The most recent intermediate result from a traced operator.
+    RowVectorPtr traceData;
+  } traceState_;
```
duanmeng: It might be cleaner to separate the struct definition from the variable declaration.

velox/exec/Cursor.cpp
```diff
@@ -507,6 +599,146 @@ class SingleThreadedTaskCursor : public TaskCursorBase {
...
+      ContinueFuture future = ContinueFuture::makeEmpty();
+
+      if (auto vector = task_->next(&future)) {
+        current_ = vector;
+        return true;
+      }
+
+      // When we hit a tracing point, the driver will return nullptr, set a
+      // future, and the trace implementation will capture state in traceState_.
+      if (traceState_.traceData) {
```
duanmeng, line 627: Is `traceState_.traceData` intended to be set by `TaskDebuggerTraceInputWriter`?

velox/exec/Cursor.cpp
```diff
@@ -507,6 +599,146 @@ class SingleThreadedTaskCursor : public TaskCursorBase {
...
+        if (isStep) {
+          current_ = traceState_.traceData;
+          return true;
```
duanmeng, line 630: So we just return here and leave it to the next `advance` call to signal the task driver to unblock?

duanmeng (approved):
@pedroerp LGTM, thanks % two nits.

velox/exec/Cursor.cpp
pedroerp: this should actually be `params.destination` !

velox/exec/Cursor.cpp
pedroerp: Yes, though they all run on the same thread, so no synchronization is needed.

velox/exec/Cursor.cpp
pedroerp: Exactly. This is what gives the "interactive" experience, as the cursor user can block the pipeline and manually inspect the traced data, until they decide to call next() to unblock the pipeline. 

Also, the driver is not really blocked; it's just sleeping off-thread :)

pedroerp:
Thank you @duanmeng. PR is updated

PR #16207, opened 2026-02-02
feat: Support unknown in coerceTypeBase

marxhxxx:
Summary: Add support for UNKNOWN type coercion in coerceTypeBase. When the source type is UNKNOWN, it can be coerced to following types TINYINT(), BOOLEAN(), SMALLINT(), INTEGER(), BIGINT(), REAL(), DOUBLE(), VARCHAR(), VARBINARY(). Cost of the coercion is higher according to the list order.

Differential Revision: D92099013


mbasmanova (approved):
Thanks!

PR #16211, opened 2026-02-03
feat(sparksql): Add transform with index parameter support

yaooqinn:
## Summary
Adds Spark-specific `transform` function that supports the 2-arg lambda signature `transform(array, (x, i) -> expr)` with element + index.

## Background
Spark's transform function supports two signatures:
1. `transform(array, x -> expr)` - element only
2. `transform(array, (x, i) -> expr)` - element + index

Velox's Presto implementation only supports signature #1. This PR adds a Spark-specific implementation that supports both.

## Changes
- Created `velox/functions/sparksql/Transform.cpp` with `SparkTransformFunction`
- Uses runtime arity detection to provide index vector when needed
- Uses 32-bit integer for index (matching Spark's IntegerType)
- Added 6 test cases covering both signatures and null handling

## Test Plan
```sql
-- Spark behavior this enables:
spark-sql> SELECT transform(array(1, 2, 3), (x, i) -> x + i);
[1, 3, 5]
```

All 6 new tests pass:
- `elementOnly` - basic transform without index
- `elementWithIndex` - transform using element + index  
- `withNulls` - null handling in arrays
- `nestedArrays` - nested array transformation
- `indexOnly` - using only index (ignore element)
- `emptyArray` - empty array edge case

### DIFF
--- a/velox/functions/sparksql/Transform.cpp
+++ b/velox/functions/sparksql/Transform.cpp
@@ -57,7 +57,8 @@ class TransformFunction : public TransformFunctionBase {
     // function(T, integer, U) has 3 children (input T, index integer, output U)
     // -> 2 input args.
     if (args[1]->type()->size() == 3) {
-      lambdaArgs.push_back(createIndexVector(flatArray, numElements, context));
+      lambdaArgs.push_back(
+          createIndexVector(flatArray, numElements, context.pool()));
     }
   }
 
@@ -68,8 +69,7 @@ class TransformFunction : public TransformFunctionBase {
   static VectorPtr createIndexVector(
       const ArrayVectorPtr& flatArray,
       vector_size_t numElements,
-      exec::EvalCtx& context) {
-    auto* pool = context.pool();
+      memory::MemoryPool* pool) {
     auto indexVector =
         BaseVector::create<FlatVector<int32_t>>(INTEGER(), numElements, pool);
 


velox/functions/sparksql/tests/TransformTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+  void testTransform(
+      const std::string& expression,
+      const VectorPtr& input,
+      const VectorPtr& expected) {
+    auto result = evaluate(expression, makeRowVector({input}));
+    assertEqualVectors(expected, result);
+  }
+};
+
+// Test transform with element-only lambda: transform(array, x -> x * 2)
```
jinchengchenghh: End with `.`

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "velox/expression/Expr.h"
+#include "velox/expression/VectorFunction.h"
+#include "velox/functions/lib/LambdaFunctionUtil.h"
+#include "velox/functions/lib/RowsTranslationUtil.h"
+#include "velox/vector/FunctionVector.h"
+
+namespace facebook::velox::functions::sparksql {
+namespace {
+
+/// Spark's transform function supporting both signatures:
```
jinchengchenghh: supporting -> supports

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,171 @@
...
+/// 1. transform(array, x -> expr) - element only
+/// 2. transform(array, (x, i) -> expr) - element + index (Spark-specific)
+///
+/// See Spark documentation:
+/// https://spark.apache.org/docs/latest/api/sql/index.html#transform
+class SparkTransformFunction : public exec::VectorFunction {
```
jinchengchenghh: Please remove Spark prefix since it has been in sparksql namespace

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,171 @@
...
+ public:
+  void apply(
```
jinchengchenghh: Could we reuse some code of udf_transform, move some of them to common path  `velox/functions/lib`?

yaooqinn:
I analyzed the code reuse opportunity. The common utilities are already shared in `velox/functions/lib/`:
- `flattenArray()` from `LambdaFunctionUtil.h`
- `toElementRows()`, `getElementToTopLevelRows()` from `RowsTranslationUtil.h`
- `toWrapCapture()`, `addNullsForUnselectedRows()` from `LambdaFunctionUtil.h`

The Spark-specific additions are:
1. Index detection via `functionType->size() == 3`
2. `createIndexVector()` helper for the `(x, i)` signature

Moving the remaining lambda iteration loop to lib would add complexity without significant benefit since it's ~15 lines of straightforward code. The current structure follows the same pattern as other Spark-specific lambda functions like `filter`.

If you'd prefer a different approach (e.g., a templated base class), I'm happy to refactor.

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,171 @@
...
+    // function(T, U) has 2 children (input T, output U) -> 1 input arg.
+    // function(T, integer, U) has 3 children (input T, index integer, output U)
+    // -> 2 input args.
+    auto functionType = args[1]->type();
+    bool withIndex = functionType->size() == 3;
+
+    std::vector<VectorPtr> lambdaArgs = {flatArray->elements()};
+
+    // If lambda expects index, create index vector.
+    if (withIndex) {
```
jinchengchenghh: Looks like the only change with presto TransformFunction is here, am I right? So I would prefer a base TransformFunction and add a virtual addIndexVectors(lambdaArgs) with default empty implementation,

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,171 @@
...
+        outputType,
+        std::move(newNulls),
+        rows.end(),
+        flatArray->offsets(),
+        flatArray->sizes(),
+        newElements);
+    context.moveOrCopyResult(localResult, rows, result);
+  }
+
+  static std::vector<std::shared_ptr<exec::FunctionSignature>> signatures() {
```
jinchengchenghh, line 32: And the signatures, call the super function signatures() and add the spark specified function signature

zhli1142015:
I think the documentation for the Spark-related functions needs to be updated.

yaooqinn:
>I think the documentation for the Spark-related functions needs to be updated.

@zhli1142015 the doc has this element-index pattern already

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,109 @@
...
+      exec::EvalCtx& context,
+      std::vector<VectorPtr>& lambdaArgs) const override {
+    // Check the lambda function type to see if it expects 2 input arguments.
+    // function(T, U) has 2 children (input T, output U) -> 1 input arg.
+    // function(T, integer, U) has 3 children (input T, index integer, output U)
+    // -> 2 input args.
+    auto functionType = args[1]->type();
+    bool withIndex = functionType->size() == 3;
+
+    if (withIndex) {
```
zhli1142015: nit: if (args[1]->type()->size() == 3)

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,109 @@
...
+      lambdaArgs.push_back(createIndexVector(flatArray, numElements, context));
+    }
+  }
+
+ private:
+  /// Creates an index vector where each element contains its position within
+  /// its respective array. For example, if we have arrays [a, b] and [c, d, e],
+  /// the index vector will be [0, 1, 0, 1, 2].
+  /// Spark uses IntegerType (32-bit) for the index.
+  static VectorPtr createIndexVector(
```
zhli1142015, line 69: I think you can use getElementToTopLevelRows here, no need to implment this function.

zhli1142015:
> > I think the documentation for the Spark-related functions needs to be updated.
> 
> @zhli1142015 the doc has this element-index pattern already

I see, thanks for the clarification.

yaooqinn:
Thanks @zhli1142015 for the review!

**Comment 1 (inline check)**: Done in the latest commit.

**Comment 2 (use getElementToTopLevelRows)**: `getElementToTopLevelRows` and `createIndexVector` serve different purposes:
- `getElementToTopLevelRows`: Creates element → source row mapping (e.g., `[0, 0, 1, 1, 1]` for arrays of size [2, 3])
- `createIndexVector`: Creates per-array element indices (e.g., `[0, 1, 0, 1, 2]` for arrays of size [2, 3])

Spark's `transform(array, (x, i) -> ...)` needs the element index within each array (starting at 0 for each array), not the row index. So we need a separate implementation here.

velox/functions/prestosql/Transform.cpp
```diff
@@ -13,88 +13,17 @@
...
-    // array(T), function(T, U) -> array(U)
-    return {exec::FunctionSignatureBuilder()
-                .typeVariable("T")
-                .typeVariable("U")
-                .returnType("array(U)")
-                .argumentType("array(T)")
-                .argumentType("function(T, U)")
-                .build()};
-  }
+class TransformFunction : public TransformFunctionBase {
```
jinchengchenghh: Could we remove the class and register TransformFunctionBase in Presto?

velox/functions/sparksql/registration/RegisterArray.cpp
```diff
@@ -52,6 +51,12 @@ void registerSparkArrayFunctions(const std::string& prefix) {
 }
 
 namespace sparksql {
+
+// Register Spark-specific transform with index support.
+void registerSparkTransform(const std::string& prefix) {
+  VELOX_REGISTER_VECTOR_FUNCTION(udf_spark_transform, prefix + "transform");
```
jinchengchenghh: Don't add a new function to only call one line function

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,106 @@
...
+            .typeVariable("T")
+            .typeVariable("U")
+            .returnType("array(U)")
+            .argumentType("array(T)")
+            .argumentType("function(T, integer, U)")
+            .build());
+    return sigs;
+  }
+
+ protected:
```
jinchengchenghh: private

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // function(T, U) has 2 children (input T, output U) -> 1 input arg.
+    // function(T, integer, U) has 3 children (input T, index integer, output U)
+    // -> 2 input args.
+    if (args[1]->type()->size() == 3) {
+      lambdaArgs.push_back(createIndexVector(flatArray, numElements, context));
+    }
+  }
+
+ private:
+  /// Creates an index vector where each element contains its position within
```
zhli1142015: nit: Comments for private methods should be written using //.

zhli1142015 (approved):

yaooqinn:
Thanks @jinchengchenghh @zhli1142015 for the additional feedback! All addressed in the latest commit:

1. **Remove Presto TransformFunction class**: Done - now directly registers `TransformFunctionBase`
2. **Inline registration**: Done - removed `registerSparkTransform()` wrapper, directly call `VELOX_REGISTER_VECTOR_FUNCTION`
3. **Change protected to private**: Done - `addIndexVector` is now private
4. **Use // for private method comments**: Done - changed `///` to `//` for private methods

velox/functions/sparksql/Transform.cpp
```diff
@@ -0,0 +1,106 @@
...
+  /// its respective array. For example, if we have arrays [a, b] and [c, d, e],
+  /// the index vector will be [0, 1, 0, 1, 2].
+  /// Spark uses IntegerType (32-bit) for the index.
+  static VectorPtr createIndexVector(
+      const ArrayVectorPtr& flatArray,
+      vector_size_t numElements,
+      exec::EvalCtx& context) {
+    auto* pool = context.pool();
```
jinchengchenghh: Ideally, we should add pool as the function argument, not `context`.

yaooqinn:
Thanks @jinchengchenghh! Done - `createIndexVector` now takes `memory::MemoryPool*` instead of `exec::EvalCtx&`.

yaooqinn:
Thank you @jinchengchenghh @zhli1142015 for the review

PR #16212, opened 2026-02-03
feat(cudf): Enable precomputation support for join filters

shrshi:
## Summary

Enable precomputation support for join filters in CUDF hash join for inner and left joins. Previously, join filters using non-AST operations (e.g., string functions like `lower()`, `upper()`, `length()`) threw `VELOX_NYI`. This change allows these filters to work by precomputing the required columns before filter evaluation.

## Changes

**AstExpression.h/cpp**
- Exposed `precomputeSubexpressions()` function for use by hash join
- Added `findExpressionSide()` helper to determine which table (probe/build) an expression references
- Fixed bug where precomputation always used the left schema, even for expressions referencing right-side fields

**CudfHashJoin.h/cpp**
- Added member variables to store precompute instructions and row types
- Added helper functions `tableViewToColumns()` and `createExtendedTableView()`
- Modified `innerJoin()` and `leftJoin()` to apply precomputation before filter evaluation
- Updated `filteredOutputIndices()` to accept extended views for filtering while using original views for output gathering
- Changed `kAllowPureAstOnly` from `true` to `false` to enable precomputation path

**HashJoinTest.cpp**
- Added tests for inner joins with string function filters (`lower`, `upper`, `length`)
- Added test combining precomputation with AST-supported operations
- Added test for left join with string function filter

## Test plan
- [x] `innerJoinWithStringFunctionFilter` - inner join with `lower()`
- [x] `innerJoinWithUpperFunctionFilter` - inner join with `upper()`
- [x] `innerJoinWithLengthFunctionFilter` - inner join with `length()`
- [x] `innerJoinWithMixedFilterPrecomputation` - string function + arithmetic comparison
- [x] `leftJoinWithStringFunctionFilter` - left join with `lower()`

### DIFF
--- a/velox/experimental/cudf/expression/AstExpressionUtils.h
+++ b/velox/experimental/cudf/expression/AstExpressionUtils.h
@@ -585,57 +585,6 @@ cudf::ast::expression const& AstContext::pushExprToTree(
   }
 }
 
-#if 0
-std::vector<ColumnOrView> precomputeSubexpressions(
-    std::vector<std::unique_ptr<cudf::column>>& inputTableColumns,
-    const std::vector<PrecomputeInstruction>& precomputeInstructions,
-    const std::vector<std::unique_ptr<cudf::scalar>>& scalars,
-    const RowTypePtr& inputRowSchema,
-    rmm::cuda_stream_view stream) {
-  std::vector<ColumnOrView> precomputedColumns;
-  precomputedColumns.reserve(precomputeInstructions.size());
-
-  for (const auto& instruction : precomputeInstructions) {
-    auto
-        [dependent_column_index,
-         ins_name,
-         new_column_index,
-         nested_dependent_column_indices,
-         cudf_expression] = instruction;
-
-    // If a compiled cudf node is available, evaluate it directly.
-    if (cudf_expression) {
-      auto result = cudf_expression->eval(
-          inputTableColumns,
-          stream,
-          cudf::get_current_device_resource_ref(),
-          /*finalize=*/true);
-      precomputedColumns.push_back(std::move(result));
-      continue;
-    }
-    if (ins_name.rfind("fill", 0) == 0) {
-      auto scalarIndex =
-          std::stoi(ins_name.substr(5)); // "fill " is 5 characters
-      auto newColumn = cudf::make_column_from_scalar(
-          *static_cast<cudf::string_scalar*>(scalars[scalarIndex].get()),
-          inputTableColumns[dependent_column_index]->size(),
-          stream,
-          cudf::get_current_device_resource_ref());
-      precomputedColumns.push_back(std::move(newColumn));
-    } else if (ins_name == "nested_column") {
-      // Nested column already exists in input. Don't materialize.
-      auto view = inputTableColumns[dependent_column_index]->view().child(
-          nested_dependent_column_indices[0]);
-      precomputedColumns.push_back(view);
-    } else {
-      VELOX_FAIL("Unsupported precompute operation " + ins_name);
-    }
-  }
-
-  return precomputedColumns;
-}
-#endif
-
 int AstContext::findExpressionSide(
     const std::shared_ptr<velox::exec::Expr>& expr) const {
   for (const auto* field : expr->distinctFields()) {


velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -399,7 +444,7 @@ CudfHashJoinProbe::CudfHashJoinProbe(
     // create ast tree
     std::vector<PrecomputeInstruction> rightPrecomputeInstructions;
     std::vector<PrecomputeInstruction> leftPrecomputeInstructions;
-    static constexpr bool kAllowPureAstOnly = true;
+    static constexpr bool kAllowPureAstOnly = false;
```
jinchengchenghh: We may remove this config if we don't need it here CC @devavret

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -421,10 +466,9 @@ CudfHashJoinProbe::CudfHashJoinProbe(
           rightPrecomputeInstructions,
           kAllowPureAstOnly);
     }
-    if (leftPrecomputeInstructions.size() > 0 ||
-        rightPrecomputeInstructions.size() > 0) {
-      VELOX_NYI("Filters that require precomputation are not yet supported");
-    }
+    // Store precompute instructions for use during join execution
+    leftPrecomputeInstructions_ = std::move(leftPrecomputeInstructions);
```
jinchengchenghh: Could we use the exposed API to execute the expression? https://github.com/facebookincubator/velox/blob/da0d2fdba364c5e14ea98536ecae57d3d7562041/velox/experimental/cudf/tests/CudfFunctionBaseTest.h#L37-L41

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -50,6 +50,47 @@
...
+  columns.reserve(tableView.num_columns());
+  for (cudf::size_type i = 0; i < tableView.num_columns(); ++i) {
+    columns.push_back(
+        std::make_unique<cudf::column>(
+            tableView.column(i),
+            stream,
+            cudf::get_current_device_resource_ref()));
+  }
+  return columns;
+}
```
devavret: This doesn't seem right. Materializing columns because precomputeSubexpressions takes vector of columns. Is there a way to modify precomputeSubexpressions to take table_view instead?

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -384,6 +425,10 @@ CudfHashJoinProbe::CudfHashJoinProbe(
     }
   }
 
+  // Store row types for precomputation
+  probeType_ = probeType;
+  buildType_ = buildType;
```
devavret: IIUC these are not mutated so no need to copy them. can these be init in constructor list

velox/experimental/cudf/exec/CudfHashJoin.cpp
devavret: I do think this was the only place it was needed so this could be removed

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -746,10 +828,42 @@ std::vector<std::unique_ptr<cudf::table>> CudfHashJoinProbe::leftJoin(
...
+      rightOwnedColumns = tableViewToColumns(rightTableView, stream);
+      rightPrecomputed = precomputeSubexpressions(
+          rightOwnedColumns,
+          rightPrecomputeInstructions_,
+          scalars_,
+          buildType_,
+          stream);
+      extendedRightView =
+          createExtendedTableView(rightTableView, rightPrecomputed);
+    }
```
devavret: Can this be cached? Since this is done on every probe, there's no reason to recompute the precompute columns of build side every time.

velox/experimental/cudf/exec/CudfHashJoin.cpp
shrshi: Yes, I've made the changes in CudfFunctionTestBase to directly accept column views

velox/experimental/cudf/expression/AstExpression.cpp
```diff
@@ -347,8 +346,38 @@ struct AstContext {
...
+
+  // Recursively check children
+  for (const auto& input : expr->inputs()) {
+    int side = findExpressionSide(input);
+    if (side >= 0) {
+      return side;
+    }
+  }
+  return -1;
+}
```
devavret: I think there was an easier way of doing this. Rather than recursing through expr, there's a `Expr::distinctFields()` that should give you a list of all fields it's referencing. Then you can lookup those in inputRowSchema sides.

velox/experimental/cudf/tests/HashJoinTest.cpp
```diff
@@ -3534,6 +3534,221 @@ TEST_F(HashJoinTest, semiProjectOverLazyVectors) {
...
+      .buildKeys({"u_c0"})
+      .buildVectors(std::move(buildVectors))
+      .buildProjections({"c0 AS u_c0", "c1 AS u_len"})
+      .joinFilter("length(t_str) = u_len")
+      .joinOutputLayout({"t_c0", "t_str", "u_c0", "u_len"})
+      .referenceQuery(
+          "SELECT t.c0, t.c1, u.c0, u.c1 FROM t, u "
+          "WHERE t.c0 = u.c0 AND length(t.c1) = u.c1")
+      .run();
+}
```
devavret: Do these test differ in something other than precompute operation type? If not then these could be redundant.

devavret (approved):

PR #16217, opened 2026-02-03
feat(parquet): Support read TIME_MILLIS parquet type

PingLiuPing:
Support read TIME type from parquet data file.

Added TimeColumnReader class to handle TIME_MILLIS (INT32, milliseconds) parquet type.
Modified ParquetColumnReader.cpp to instantiate the appropriate TimeColumnReader based on the underlying Parquet physical type.

### DIFF
--- a/velox/dwio/common/tests/utils/BatchMaker.cpp
+++ b/velox/dwio/common/tests/utils/BatchMaker.cpp
@@ -141,6 +141,18 @@ VectorPtr BatchMaker::createVector<TypeKind::BIGINT>(
     MemoryPool& pool,
     std::mt19937& gen,
     std::function<bool(vector_size_t /*index*/)> isNullAt) {
+  if (type->isTime()) {
+    return createScalar<int64_t>(
+        size,
+        gen,
+        [&gen]() {
+          // TIME is milliseconds since midnight.
+          return Random::rand64(0, 86400000, gen);
+        },
+        pool,
+        isNullAt,
+        type);
+  }
   return createScalar<int64_t>(
       size,
       gen,
--- a/velox/dwio/parquet/reader/Metadata.cpp
+++ b/velox/dwio/parquet/reader/Metadata.cpp
@@ -26,6 +26,20 @@ inline const T load(const char* ptr) {
   return ret;
 }
 
+inline std::optional<int64_t> decodeInt64Stat(const std::string& bytes) {
+  switch (bytes.size()) {
+    case sizeof(int64_t):
+      return load<int64_t>(bytes.data());
+    case sizeof(int32_t):
+      // Parquet stores Time types as int32_t (for milliseconds), but Velox's
+      // Time type is always int64_t, so we need to load sizeof(int32_t) and
+      // cast to int64_t.
+      return static_cast<int64_t>(load<int32_t>(bytes.data()));
+    default:
+      return std::nullopt;
+  }
+}
+
 template <typename T>
 inline std::optional<T> getMin(const thrift::Statistics& columnChunkStats) {
   return columnChunkStats.__isset.min_value
@@ -42,6 +56,24 @@ inline std::optional<T> getMax(const thrift::Statistics& columnChunkStats) {
       : (columnChunkStats.__isset.max
              ? std::optional<T>(load<T>(columnChunkStats.max.data()))
              : std::nullopt);
+}
+
+template <>
+inline std::optional<int64_t> getMin(
+    const thrift::Statistics& columnChunkStats) {
+  return columnChunkStats.__isset.min_value
+      ? decodeInt64Stat(columnChunkStats.min_value)
+      : (columnChunkStats.__isset.min ? decodeInt64Stat(columnChunkStats.min)
+                                      : std::nullopt);
+}
+
+template <>
+inline std::optional<int64_t> getMax(
+    const thrift::Statistics& columnChunkStats) {
+  return columnChunkStats.__isset.max_value
+      ? decodeInt64Stat(columnChunkStats.max_value)
+      : (columnChunkStats.__isset.max ? decodeInt64Stat(columnChunkStats.max)
+                                      : std::nullopt);
 }
 
 template <>
--- a/velox/dwio/parquet/reader/PageReader.cpp
+++ b/velox/dwio/parquet/reader/PageReader.cpp
@@ -393,7 +393,7 @@ void PageReader::prepareDictionary(const PageHeader& pageHeader) {
           ? sizeof(float)
           : sizeof(double);
       auto numBytes = dictionary_.numValues * typeSize;
-      if ((type_->type()->isShortDecimal() || type_->type()->isTime()) &&
+      if (type_->type()->isShortDecimal() &&
           parquetType == thrift::Type::INT32) {
         auto veloxTypeLength = type_->type()->cppSizeInBytes();
         auto numVeloxBytes = dictionary_.numValues * veloxTypeLength;
@@ -421,7 +421,7 @@ void PageReader::prepareDictionary(const PageHeader& pageHeader) {
         }
         stats_.pageLoadTimeNs.increment(readUs * 1'000);
       }
-      if ((type_->type()->isShortDecimal() || type_->type()->isTime()) &&
+      if (type_->type()->isShortDecimal() &&
           parquetType == thrift::Type::INT32) {
         auto values = dictionary_.values->asMutable<int64_t>();
         auto parquetValues = dictionary_.values->asMutable<int32_t>();
--- a/velox/dwio/parquet/reader/ParquetColumnReader.cpp
+++ b/velox/dwio/parquet/reader/ParquetColumnReader.cpp
@@ -27,6 +27,7 @@
 #include "velox/dwio/parquet/reader/RepeatedColumnReader.h"
 #include "velox/dwio/parquet/reader/StringColumnReader.h"
 #include "velox/dwio/parquet/reader/StructColumnReader.h"
+#include "velox/dwio/parquet/reader/TimeColumnReader.h"
 #include "velox/dwio/parquet/reader/TimestampColumnReader.h"
 #include "velox/dwio/parquet/thrift/ParquetThriftTypes.h"
 
@@ -40,6 +41,12 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
     ParquetParams& params,
     common::ScanSpec& scanSpec) {
   auto colName = scanSpec.fieldName();
+
+  if (fileType->type()->isTime()) {
+    VELOX_CHECK(fileType->type()->equivalent(*TIME()));
+    return std::make_unique<TimeColumnReader>(
+        requestedType, fileType, params, scanSpec);
+  }
 
   switch (fileType->type()->kind()) {
     case TypeKind::INTEGER:
--- a/velox/dwio/parquet/reader/ParquetReader.cpp
+++ b/velox/dwio/parquet/reader/ParquetReader.cpp
@@ -996,10 +996,27 @@ TypePtr ReaderBase::convertType(
             requestedType->toString());
         return VARCHAR();
       }
+      case thrift::ConvertedType::TIME_MILLIS:
+        VELOX_CHECK_EQ(
+            schemaElement.type,
+            thrift::Type::INT32,
+            "TIME_MILLIS converted type can only be set for value of thrift::Type::INT32");
+        VELOX_CHECK(
+            !requestedType ||
+                isCompatible(
+                    requestedType,
+                    isRepeated,
+                    [](const TypePtr& type) {
+                      return type->equivalent(*TIME());
+                    }),
+            kTypeMappingErrorFmtStr,
+            "TIME",
+            requestedType->toString());
+        return TIME();
+
       case thrift::ConvertedType::MAP:
       case thrift::ConvertedType::MAP_KEY_VALUE:
       case thrift::ConvertedType::LIST:
-      case thrift::ConvertedType::TIME_MILLIS:
       case thrift::ConvertedType::TIME_MICROS:
       case thrift::ConvertedType::JSON:
       case thrift::ConvertedType::BSON:
--- a/velox/dwio/parquet/reader/TimeColumnReader.h
+++ b/velox/dwio/parquet/reader/TimeColumnReader.h
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/dwio/parquet/reader/IntegerColumnReader.h"
+#include "velox/dwio/parquet/reader/ParquetColumnReader.h"
+#include "velox/dwio/parquet/thrift/ParquetThriftTypes.h"
+
+namespace facebook::velox::parquet {
+
+/// Column reader for Parquet TIME type.
+/// Handles conversion from Parquet TIME_MILLIS (INT32, milliseconds) to
+/// Velox TIME type (BIGINT, milliseconds).
+class TimeColumnReader : public IntegerColumnReader {
+ public:
+  TimeColumnReader(
+      const TypePtr& requestedType,
+      const std::shared_ptr<const dwio::common::TypeWithId>& fileType,
+      ParquetParams& params,
+      common::ScanSpec& scanSpec)
+      : IntegerColumnReader(requestedType, fileType, params, scanSpec) {
+    const auto typeWithId =
+        std::static_pointer_cast<const ParquetTypeWithId>(fileType_);
+    if (auto logicalType = typeWithId->logicalType_) {
+      VELOX_CHECK(logicalType->__isset.TIME);
+      const auto unit = logicalType->TIME.unit;
+      VELOX_CHECK(
+          unit.__isset.MILLIS,
+          "TIME precision other than milliseconds is not supported");
+    } else if (auto convertedType = typeWithId->convertedType_) {
+      VELOX_CHECK(
+          convertedType == thrift::ConvertedType::type::TIME_MILLIS,
+          "TIME converted type other than TIME_MILLIS is not supported");
+    } else {
+      VELOX_NYI("Logical type and converted type are not provided for TIME.");
+    }
+  }
+
+  void read(
+      int64_t offset,
+      const RowSet& rows,
+      const uint64_t* /*incomingNulls*/) override {
+    // Parquet stores TIME as INT32 (TIME_MILLIS).
+    // Velox represents TIME as BIGINT (8-byte).
+    // Use the physical width: TIME_MILLIS is INT32 (4 bytes).
+    VELOX_WIDTH_DISPATCH(4, prepareRead, offset, rows, nullptr);
+    readCommon<IntegerColumnReader, true>(rows);
+    readOffset_ += rows.back() + 1;
+  }
+};
+
+} // namespace facebook::velox::parquet
--- a/velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
+++ b/velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
@@ -692,6 +692,52 @@ TEST_F(E2EFilterTest, date) {
       20);
 }
 
+TEST_F(E2EFilterTest, time) {
+  struct {
+    parquet::arrow::Encoding::type encoding;
+    bool enableDictionary;
+    bool keepNulls;
+  } testCases[] = {
+      {parquet::arrow::Encoding::kPlain, false, true},
+      {parquet::arrow::Encoding::kPlain, true, true},
+      {parquet::arrow::Encoding::kDeltaBinaryPacked, false, false},
+      {parquet::arrow::Encoding::kDeltaBinaryPacked, false, true},
+  };
+
+  for (const auto& testCase : testCases) {
+    options_.encoding = testCase.encoding;
+    bool enableDictionary = testCase.enableDictionary;
+    bool keepNulls = testCase.keepNulls;
+    SCOPED_TRACE(
+        fmt::format(
+            "Encoding: {}, Dictionary: {}, KeepNulls: {}",
+            static_cast<int>(options_.encoding),
+            enableDictionary,
+            keepNulls));
+
+    options_.enableDictionary = enableDictionary;
+    options_.dataPageSize = 4 * 1024;
+    const int valMax = enableDictionary ? 1000 : 86399999;
+
+    testWithTypes(
+        "time_val:time",
+        [&]() {
+          makeIntDistribution<int64_t>(
+              "time_val",
+              0, // min
+              valMax, // max
+              22, // repeats
+              19, // rareFrequency
+              0, // rareMin
+              valMax, // rareMax
+              keepNulls); // keepNulls
+        },
+        false,
+        {"time_val"},
+        20);
+  }
+}
+
 TEST_F(E2EFilterTest, combineRowGroup) {
   rowsInRowGroup_ = 5;
   rowType_ = ROW({"c0"}, {INTEGER()});
--- a/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
+++ b/velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
@@ -1913,6 +1913,69 @@ TEST_F(ParquetReaderTest, columnStatisticsMultipleRowGroups) {
   EXPECT_EQ(intStats->getMaximum(), 50);
 }
 
+TEST_F(ParquetReaderTest, readTimeMillis) {
+  // Write TIME data using the parquet writer.
+  // The writer exports Velox TIME as Arrow time32 with milliseconds unit,
+  // which maps to Parquet TIME_MILLIS (INT32).
+  const auto rowType = ROW({"time_col"}, {TIME()});
+
+  auto data = makeRowVector(
+      rowType->names(),
+      {makeNullableFlatVector<int64_t>(
+          {1, 1'000, 3'600'000, std::nullopt, 43'200'000, 86'399'999},
+          TIME())});
+  auto sink = write(data);
+
+  dwio::common::ReaderOptions readerOpts{leafPool_.get()};
+  auto reader = createReaderInMemory(*sink, readerOpts);
+
+  EXPECT_EQ(reader->numberOfRows(), 6ULL);
+
+  auto type = reader->typeWithId();
+  EXPECT_EQ(type->size(), 1ULL);
+  auto col0 = type->childAt(0);
+  EXPECT_TRUE(col0->type()->isTime());
+
+  auto rowReaderOpts = getReaderOpts(rowType);
+  rowReaderOpts.setScanSpec(makeScanSpec(rowType));
+  auto rowReader = reader->createRowReader(rowReaderOpts);
+
+  assertReadWithReaderAndExpected(rowType, *rowReader, data, *leafPool_);
+}
+
+TEST_F(ParquetReaderTest, readTimeWithMultipleColumns) {
+  const auto rowType =
+      ROW({"id", "time_col", "name"}, {INTEGER(), TIME(), VARCHAR()});
+
+  auto data = makeRowVector(
+      rowType->names(),
+      {
+          makeFlatVector<int32_t>({1, 2, 3, 4, 5}),
+          makeFlatVector<int64_t>(
+              {0, 1'000, 3'600'000, 43'200'000, 86'399'999}, TIME()),
+          makeFlatVector<StringView>({"a", "b", "c", "d", "e"}),
+      });
+
+  auto sink = write(data);
+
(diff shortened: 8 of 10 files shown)


PingLiuPing:
@majetideepak Could you please help review this PR in your role as a maintainer? Thank you.

PingLiuPing:
@rui-mo Would you mind taking a look at this PR? Thank you.

velox/dwio/parquet/reader/TimeColumnReader.h
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    if (isMicros_) {
+      VectorPtr resultVector = *result;
+      auto rawValues =
+          resultVector->asUnchecked<FlatVector<int64_t>>()->mutableRawValues();
+      for (auto i = 0; i < numValues_; ++i) {
+        if (!resultVector->isNullAt(i)) {
+          // Convert microseconds to milliseconds.
+          rawValues[i] = rawValues[i] / 1000;
```
rui-mo: Spark’s time type uses microsecond precision, so I assume the current implementation is only suitable for Presto.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
```diff
@@ -1792,6 +1792,68 @@ TEST_F(ParquetReaderTest, readerWithSchema) {
   EXPECT_EQ(reader.rowType()->toString(), schema->toString());
 }
 
+TEST_F(ParquetReaderTest, readTimeMillis) {
```
rui-mo, line 1916: You may need to extend the E2EFilterTest to ensure filter pushdown and different encodings work.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -47,6 +48,22 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
     case TypeKind::SMALLINT:
     case TypeKind::TINYINT:
     case TypeKind::HUGEINT:
+      if (fileType->type()->isTime()) {
```
rui-mo: Should we only check `fileType->type()->isTime()` for INTEGER and BIGINT types?

velox/dwio/parquet/reader/TimeColumnReader.h
PingLiuPing: @rui-mo  Thank you for the comments.
Actually this is for aligning with Velox TIME type, Velox TIME type represents milliseconds since midnight.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Thanks, updated.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
PingLiuPing: Thanks @rui-mo

I fixed a bug for kDeltaBinaryPacked encoding when read time value from parquet.

Some background:
Velox TIME is stored as int64_t (BIGINT) internally, representing milliseconds since midnight. During write to parquet file, the flow is:
Velox -> Arrow export (Bridge.cpp), Velox TIME is exported as **Arrow time32** with milliseconds unit (format string "ttm").
Arrow -> Parquet schema (ArrowSchema.cpp)
Arrow TIME32 maps to Parquet kInt32 with TIME_MILLIS logical type.
So Parquet physical type for TIME is INT32, not INT64.

DeltaBpDecoder accumulates deltas using **64 bit** arithmetic, which is correct for INT64 data but causes incorrect overflow behavior for INT32 data.

velox/dwio/common/tests/utils/BatchMaker.cpp
```diff
@@ -144,7 +144,13 @@ VectorPtr BatchMaker::createVector<TypeKind::BIGINT>(
   return createScalar<int64_t>(
       size,
       gen,
-      [&gen]() { return Random::rand64(gen); },
+      [&gen, &type]() {
+        if (type->isTime()) {
```
Yuhta: Can we move this branching out side of the loop?

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
rui-mo: >  Parquet physical type for TIME is INT32, not INT64.

It appears that TIME_MILLIS uses int32 and TIME_MICROS uses int64.
https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L90-L104

velox/dwio/parquet/reader/TimeColumnReader.h
rui-mo: Ah, thanks for pointing that out. We’ll need to extend support to Spark as well. Since Velox’s TimeType extends BigintType, adding this support shouldn’t be too complicated, and we’ll take care of it in this PR: https://github.com/facebookincubator/velox/pull/16468.

If you’d like, we’d be happy to take care of the microsecond support after the time type has been extended.

velox/dwio/parquet/reader/TimeColumnReader.h
```diff
@@ -0,0 +1,83 @@
...
+  void read(
+      int64_t offset,
+      const RowSet& rows,
+      const uint64_t* /*incomingNulls*/) override {
+    // Parquet stores TIME as INT32 (TIME_MILLIS) or INT64 (TIME_MICROS).
+    // Velox represents TIME as BIGINT (8-byte).
+    // Use the physical width: TIME_MILLIS is INT32 (4 bytes),
+    // TIME_MICROS is INT64 (8 bytes).
+    VELOX_WIDTH_DISPATCH(isMicros_ ? 8 : 4, prepareRead, offset, rows, nullptr);
+    readCommon<IntegerColumnReader, true>(rows);
```
yingsu00, line 61: DO we need to do conversion (x/1000) for TIME_MICROS here? If the reader simply reads the raw Parquet TIME_MICROS value into Velox TIME without conversion, the interpretation will be wrong. For example, Parquet encodes a logical value of 12:00:00 as 
12h × 3600s × 1,000,000 = 43,200,000,000 microseconds, so the raw integer in the file is: 43,200,000,000. However Velox TIME is defined as "int64 milliseconds since midnight" so 43,200,000,000 would be interpreted as 500 days.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
```diff
@@ -1792,6 +1792,68 @@ TEST_F(ParquetReaderTest, readerWithSchema) {
...
+  EXPECT_EQ(type->childAt(0)->type()->kind(), TypeKind::INTEGER);
+  EXPECT_TRUE(type->childAt(1)->type()->isTime());
+  EXPECT_EQ(type->childAt(2)->type()->kind(), TypeKind::VARCHAR);
+
+  auto rowReaderOpts = getReaderOpts(rowType);
+  rowReaderOpts.setScanSpec(makeScanSpec(rowType));
+  auto rowReader = reader->createRowReader(rowReaderOpts);
+
+  assertReadWithReaderAndExpected(rowType, *rowReader, data, *leafPool_);
+}
```
yingsu00, line 1977: Can you please add a TIME_MICROS file test?

yingsu00: Can you please also add some tests for the out of range values? E.g. -1

velox/common/base/SimdUtil-inl.h
```diff
@@ -1317,7 +1335,11 @@ template <typename T, typename A>
 struct Filter<T, A, 2> {
   static xsimd::batch<T, A>
   apply(xsimd::batch<T, A> data, int mask, const xsimd::generic&) {
-    return genericPermute(data, byteSetBits[mask]);
+    if constexpr (xsimd::batch<T, A>::size <= 8) {
+      return genericPermute(data, byteSetBits[mask]);
+    } else {
+      return genericFilter(data, mask);
```
Yuhta: Which platform are we talking about?  Velox is not well tested under AVX512 yet (individual compilation unit could enable, but not globally).  If we want to enable it there, it should be a separate change.

PingLiuPing: @Yuhta Thanks, this is some debug code that I accidentally pushed here. I'll revert these changes.

velox/dwio/common/DirectDecoder.h
```diff
@@ -95,7 +95,8 @@ class DirectDecoder : public IntDecoder<isSigned> {
                                int128_t>) {
         toSkip = visitor.process(super::template readInt<int128_t>(), atEnd);
       } else {
-        toSkip = visitor.process(super::template readInt<int64_t>(), atEnd);
+        toSkip = visitor.process(
+            super::template readInt<typename Visitor::DataType>(), atEnd);
```
Yuhta: This does not seem have any effect.  `readInt()` either return int128 or int64, anything other than that is just an implicit cast.
Actually for anything other than these 2, it will cause bug in `readLittleEndianFromBigEndian`, so this should not be changed, and we probably should add compile time restriction in the super class for the types.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
PingLiuPing: Yes, for milliseconds, the physical type is int32_t

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
PingLiuPing: > Can you please add a TIME_MICROS file test?

Thanks Ying, I will add reading of TIME_MICROS in separate PR (should be simpler than this). And before that I will submit a PR to write TIME_MICROS first.

PingLiuPing: > Can you please also add some tests for the out of range values? E.g. -1

Thanks Ying, there is no such validness check for Time type, meaning with invalid value like -1 this test can pass. And I check other types too, there are no such check as well.

The rationale are:
Vector Creation: makeNullableFlatVector<int64_t>({-1}, TIME()) stores the literal int64_t value -1 in memory. Velox doesn't validate the bounds of TIME values at creation.
Arrow Export: When write() exports the Velox vector to an Arrow array, velox/vector/arrow/Bridge.cpp's gatherFromTimeBuffer takes the int64_t and runs dst[j] = static_cast<int32_t>(src[i]);. The 64-bit -1 becomes a 32-bit -1.
Parquet Write: Arrow's Parquet writer writes the 32-bit -1 to the file as plain Parquet INT32 encoding without checking if it falls between [0, 86400000000).
Parquet Read: During reading, Velox's Parquet reader reads the INT32 values. The -1 is read back and cast back up to Velox's int64_t (TIME type).
Comparison: The test's assertReadWithReaderAndExpected compares the results memory-by-memory. Since -1 round-tripped perfectly back and forth without constraints complaining, they match, and the test passes!
Neither the Arrow bridge nor the Parquet readers/writers strictly police the [0, 86400000) semantics for TIME_MILLIS at the encoding/decoding layer.

velox/dwio/parquet/tests/reader/ParquetReaderTest.cpp
PingLiuPing: @rui-mo  Thanks for the reminder, I have added a testcase to E2EFilterTest.

velox/dwio/parquet/reader/TimeColumnReader.h
PingLiuPing: Thanks @yingsu00 , I will submit another PR to support reading TIME_MICROS, before that I need to submit a PR to support writing TIME_MICROS first.

velox/dwio/common/tests/utils/BatchMaker.cpp
PingLiuPing: Thanks, moved out.

velox/dwio/common/DirectDecoder.h
PingLiuPing: Thanks @Yuhta , removed these change.

PingLiuPing: @Yuhta Your comments have been answered, could you help to take another look at your convenience? Thank you.

Yuhta: The common change looks good, I will leave it to @rui-mo and @majetideepak for the Parquet specific part

PingLiuPing: @rui-mo @majetideepak Could you help take a look at your convenience?

velox/dwio/parquet/reader/ParquetReader.cpp
```diff
@@ -976,10 +976,25 @@ TypePtr ReaderBase::convertType(
...
+        VELOX_CHECK_EQ(
+            schemaElement.type,
+            thrift::Type::INT32,
+            "TIME_MILLIS converted type can only be set for value of thrift::Type::INT32");
+        VELOX_CHECK(
+            !requestedType ||
+                isCompatible(
+                    requestedType,
+                    isRepeated,
+                    [](const TypePtr& type) { return type->isTime(); }),
```
rui-mo: nit: use `equivalent` to match against TIME() type, because the isTime will be extended to allow other TimeType.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -44,6 +45,12 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
   switch (fileType->type()->kind()) {
     case TypeKind::INTEGER:
     case TypeKind::BIGINT:
+      if (fileType->type()->isTime()) {
```
rui-mo: Should we check for `fileType->type()->isTime()` only when file type is integer?

velox/dwio/parquet/reader/Metadata.cpp
```diff
@@ -44,6 +58,24 @@ inline std::optional<T> getMax(const thrift::Statistics& columnChunkStats) {
              : std::nullopt);
 }
 
+template <>
+inline std::optional<int64_t> getMin(
+    const thrift::Statistics& columnChunkStats) {
+  return columnChunkStats.__isset.min_value
+      ? decodeInt64Stat(columnChunkStats.min_value)
```
rui-mo, line 65: Could you clarify why we need to enforce the min/max values as int64? Would it be possible to return them as int32?

velox/dwio/parquet/reader/TimeColumnReader.h
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  void read(
+      int64_t offset,
+      const RowSet& rows,
+      const uint64_t* /*incomingNulls*/) override {
+    // Parquet stores TIME as INT32 (TIME_MILLIS).
+    // Velox represents TIME as BIGINT (8-byte).
+    // Use the physical width: TIME_MILLIS is INT32 (4 bytes).
+    VELOX_WIDTH_DISPATCH(4, prepareRead, offset, rows, nullptr);
```
rui-mo, line 60: Is it correct to prepare the value buffer as integers? It seems that we need to read int32_t values as int64_t, so should we allocate the value buffer based on the size of the bigint type?

velox/dwio/parquet/reader/Metadata.cpp
PingLiuPing: @rui-mo Thanks, the reason is that velox stores Time as int64_t, I added a comment in line 34-36.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Thanks.

The fileType->type() here is the Velox type (BIGINT/TIME), not the Parquet physical type (INT32).

velox/dwio/parquet/reader/ParquetReader.cpp
PingLiuPing: Thanks.

velox/dwio/parquet/reader/TimeColumnReader.h
PingLiuPing: Thanks.
                                                                                                                                                The 4 in VELOX_WIDTH_DISPATCH controls the template type T used in prepareRead<T>, which determines:                                              
  1. valueSize_ is set to sizeof(T) = 4 (line 82 of SelectiveColumnReaderInternal.h)                                                                
  2. The value buffer is allocated with ensureValuesCapacity<T> using 4-byte slots (line 88)                                                        
                                                                                                                                                  This matches the Parquet physical format — TIME_MILLIS is stored as INT32 (4 bytes), so the decoder will write 4-byte int32 values into the buffer.
The widening from int32 to int64 happens later in getValues() -> getIntValues(). Since the requested type is BIGINT, it hits the TypeKind::BIGINT, and since valueSize_ is 4, it calls getFlatValues<int32_t, int64_t>(...)

PingLiuPing: @rui-mo Thanks for reviewing, I addressed your comments and rebased the code.

rui-mo: I see, thanks for clarifying. BTW, why is `fileType_`  int64 instead of int32 for the TIME millisecond type, and is that correct? If the `fileType_` is int32, we may don't need to override the `read` function at all.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
rui-mo: In this case, should we only add this check for BIGINT case? There is also one concern on if it is correct to have the bigint file type, because this information is extracted by `ReaderBase::getParquetColumnInfo` through Parquet.

velox/dwio/parquet/reader/TimeColumnReader.h
```diff
@@ -0,0 +1,67 @@
...
+          "TIME precision other than milliseconds is not supported");
+    } else if (auto convertedType = typeWithId->convertedType_) {
+      VELOX_CHECK(
+          convertedType == thrift::ConvertedType::type::TIME_MILLIS,
+          "TIME converted type other than TIME_MILLIS is not supported");
+    } else {
+      VELOX_NYI("Logical type and converted type are not provided for TIME.");
+    }
+  }
+
```
rui-mo, line 52: Do we need to override the `hasBulkPath` function? For example, in the floating point column reader, fast path is provided only when the data and result types match.

```
  // Offers a fast path only if data and result widths match.
  static constexpr bool kHasBulkPath = std::is_same_v<TData, TRequested>;

  bool hasBulkPath() const override {
    return kHasBulkPath;
  }
```

rui-mo: I did a quick check and found that ColumnVisitor already supports different integer widths, so overriding hasBulkPath may not be needed.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Thanks, Hm.., interesting, let me find out.

PingLiuPing: @rui-mo The var name here is misleading, 
here is the impl of `fileType->type()`

```
  const std::shared_ptr<const velox::Type>& type() const {
    return type_;
  }
```

the type_ is initialized as the first param from 
```
  TypeWithId(
      std::shared_ptr<const velox::Type> type,
      std::vector<std::unique_ptr<TypeWithId>>&& children,
      uint32_t id,
      uint32_t maxId,
      uint32_t column);
```

And it is also the first param of ParquetTypeWithId ctor.
And from `ReaderBase::getParquetColumnInfo` when ParquetTypeWithId is constructed, the first param is returned from `ReaderBase::convertType` which is TIME().

velox/dwio/parquet/reader/TimeColumnReader.h
PingLiuPing: Thanks, as explained above, here the `type_` does not indicates the parquetType/logicalType/converted_type etc. It is the velox type.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -44,6 +45,12 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
...
+        return std::make_unique<TimeColumnReader>(
```
rui-mo: nit: add `VELOX_DCHECK(type->equivalent(*TIME()));` to ensure only TIME_MILLI can be enabled.

velox/dwio/parquet/reader/Metadata.cpp
```diff
@@ -26,6 +26,20 @@ inline const T load(const char* ptr) {
   return ret;
 }
 
+inline std::optional<int64_t> decodeInt64Stat(const std::string& bytes) {
+  switch (bytes.size()) {
+    case sizeof(int64_t):
+      return load<int64_t>(bytes.data());
+    case sizeof(int32_t):
```
rui-mo, line 33: Would this change impact how metrics are handled for other types? For instance, if a type is int32_t and is meant to be read as int32_t, would this cause its metrics to be converted to int64_t instead? Is that correct?

velox/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -44,6 +45,12 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
   switch (fileType->type()->kind()) {
     case TypeKind::INTEGER:
     case TypeKind::BIGINT:
```
rui-mo, line 53: Should we match only `case TypeKind::BIGINT`, rather than matching both INTEGER and BIGINT?

velox/dwio/parquet/reader/TimeColumnReader.h
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/common/base/SimdUtil.h"
```
rui-mo: nit: is this header needed?

velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
```diff
@@ -692,6 +692,51 @@ TEST_F(E2EFilterTest, date) {
...
 
+TEST_F(E2EFilterTest, time) {
+  struct {
+    parquet::arrow::Encoding::type encoding;
+    bool enableDictionary;
+    bool keepNulls;
+  } testCases[] = {
+      {parquet::arrow::Encoding::kPlain, false, true},
+      {parquet::arrow::Encoding::kPlain, true, true},
+      {parquet::arrow::Encoding::kDeltaBinaryPacked, false, false},
```
rui-mo, line 703: Do we have test case like `kDeltaBinaryPacked, false, true`?

velox/dwio/parquet/reader/ParquetColumnReader.cpp
rui-mo: I see. Thanks for clarifying!

velox/dwio/parquet/reader/TimeColumnReader.h
PingLiuPing: Thanks.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Thanks for the comment. The current implementation is likely ok. Otherwise which one of the following is better?

```
    case TypeKind::INTEGER:
      return std::make_unique<IntegerColumnReader>(
          requestedType, fileType, params, scanSpec);
    case TypeKind::BIGINT:
      if (fileType->type()->isTime()) {
        VELOX_DCHECK(fileType->type()->equivalent(*TIME()));
        return std::make_unique<TimeColumnReader>(
            requestedType, fileType, params, scanSpec);
      }
      return std::make_unique<IntegerColumnReader>(
          requestedType, fileType, params, scanSpec);
```


```
    case TypeKind::BIGINT:
      if (fileType->type()->isTime()) {
        VELOX_DCHECK(fileType->type()->equivalent(*TIME()));
        return std::make_unique<TimeColumnReader>(
            requestedType, fileType, params, scanSpec);
      }
    [[fallthrough]];
    case TypeKind::INTEGER:
      return std::make_unique<IntegerColumnReader>(
          requestedType, fileType, params, scanSpec);
```

velox/dwio/parquet/reader/Metadata.cpp
PingLiuPing: @rui-mo  Thanks. The decodeInt64Stat function is only used by the getMin<int64_t> and getMax<int64_t> template specializations. Those specializations are only called for TypeKind::BIGINT (line 148-149).                                                                                           
                                                                                                                                                    For TypeKind::INTEGER (int32_t types), the code calls the generic template getMin<int32_t>/getMax<int32_t> which uses load<int32_t> directly — decodeInt64Stat is never involved.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: thanks, added it temporarily, will remove it in next PR when supporting reading time_micro

velox/dwio/parquet/tests/reader/E2EFilterTest.cpp
PingLiuPing: Thanks, added.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
rui-mo, line 43: Can we move the special handling for `if (fileType->type()->isTime())` here and keep the switch-case unchanged for better readability?

velox/dwio/parquet/reader/PageReader.cpp
```diff
@@ -421,7 +421,7 @@ void PageReader::prepareDictionary(const PageHeader& pageHeader) {
         }
         stats_.pageLoadTimeNs.increment(readUs * 1'000);
       }
-      if ((type_->type()->isShortDecimal() || type_->type()->isTime()) &&
+      if (type_->type()->isShortDecimal() &&
           parquetType == thrift::Type::INT32) {
```
rui-mo, line 425: Which branch does TIME fall into now? Is the original logic correct to read int32 values and converting them to int64? Thanks!

PingLiuPing: @rui-mo Thanks, when I rebased the code today I noticed that the test I added for time is failed, and eventually I found that the bytes here is wrong. Now for time type it will not fall into any branch here.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Moved, thanks.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
```diff
@@ -41,6 +42,11 @@ std::unique_ptr<dwio::common::SelectiveColumnReader> ParquetColumnReader::build(
     common::ScanSpec& scanSpec) {
   auto colName = scanSpec.fieldName();
 
+  if (fileType->type()->equivalent(*TIME())) {
```
rui-mo: nit: keep using `isTime` and add a VELOX_DCHECK to ensure the type is TIME(). Otherwise, the current implementation would cause other time types to fall into `IntegerColumnReader`, which is incorrect.

velox/dwio/parquet/reader/PageReader.cpp
rui-mo: Sorry, I still don’t understand why it wasn’t necessary to expand the int32 values to int64 to store them correctly in Velox. Since TIME_MILLI values are stored as continuous int32 values, I assumed we would need to read `dictionary_.numValues * 4` bytes and expand them to `dictionary_.numValues * 8` bytes.

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Thanks. I think TIME() represents a single type and using `equivalent` can block other non TIME() type already.

velox/dwio/parquet/reader/PageReader.cpp
PingLiuPing: If we widen dictionary INT32 values to INT64 for TIME here and since `TimeColumnReader::read()` always dispatches with width 4, when dictionary-encoded values are accessed, we use 4 to index into the dictionary buffer. But the dictionary values are now 8 bytes wide (after widening).

velox/dwio/parquet/reader/ParquetColumnReader.cpp
rui-mo: Yes, only `TIME()` will enter this branch, while the other time types falls into the switch-case and end up in `IntegerColumnReader`, which is incorrect. I think you need to make sure an exception is thrown for the other time types.

velox/dwio/parquet/reader/PageReader.cpp
rui-mo: Thanks for clarifying!

velox/dwio/parquet/reader/ParquetColumnReader.cpp
PingLiuPing: Oh, get your point, so in such case we should use VELOX_CHECK instead?

rui-mo: I assume DCHECK is fine for now, before supporting reading the other time types.

rui-mo:
Would you please update the PR description to match the implementation? Thanks.

PingLiuPing:
@Yuhta  Could you help to take a final look? Thank you.

Issue #16216, opened 2026-02-03
Velox rejects valid IANA timezone America/Coyhaique present in system tzdb


ReemaAlzaid:
### Bug description

Velox fails to accept a valid IANA timezone (America/Coyhaique) that is present in the system tzdb (tzdata ≥ 2025b), causing Spark + Gluten queries to fail at runtime with a VeloxUserError

This issue appears when running on newer Linux distributions (e.g. CentOS 9) where tzdata includes recently added IANA zones

### System information

OS: CentOS 9

tzdata: 2025c-1.el9

Velox: (current main / version used by Gluten)

Java: 17

Spark: 4.1

Gluten backend: Velox

### Relevant logs

```bash
org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Code: INVALID_ARGUMENT
Reason: session 'session_timezone' set with invalid value 'America/Coyhaique'
Expression: tz::getTimeZoneID(*tz, false) != -1
File: velox/core/QueryConfig.cpp
```

kgpai:
@ReemaAlzaid This has to be updated here : https://github.com/prestodb/presto/blob/master/presto-common/src/main/resources/com/facebook/presto/common/type/zone-index.properties before we can update the mapping in Velox. I would suggest raise an issue with Presto to fix this and then we can fix it here.

ReemaAlzaid:
Hey @kgpai 
oh I thought it must be updated here first? https://github.com/facebookincubator/velox/pull/16221/changes#diff-aa88ed4d6ad1e56e21b0a999f5415fedc003e6c2acf7e2ed80ac390a0085dfe6

kgpai:
@ReemaAlzaid No its auto generated - Please ensure Presto is updated first.

n0r0shi:
There is an open Presto PR (prestodb/presto#26981) that adds `America/Coyhaique` as timezone ID 2234 to
   `zone-index.properties`.

   I've cut a draft PR on the Velox side to add the same timezone entry to `TimeZoneDatabase.cpp`. It will be ready for review once the Presto PR is merged, to keep the ID assignment consistent.

ReemaAlzaid:
Hey @kgpai @pedroerp
the changes has been merged in presto as seen here https://github.com/prestodb/presto/blob/master/presto-common/src/main/resources/com/facebook/presto/common/type/zone-index.properties#L2243

so we can merge this pr https://github.com/facebookincubator/velox/pull/16221

PR #16221, opened 2026-02-03
fix: Add America/Coyhaique timezone

ReemaAlzaid:
fixes issue #16216, https://github.com/apache/incubator-gluten/issues/11597
Added America/Coyhaique timezone with ID 2234


velox/type/tz/TimeZoneDatabase.cpp
```diff
@@ -2265,6 +2265,7 @@ const std::vector<std::pair<int16_t, std::string>>& getTimeZoneEntries() {
         {2231, "Pacific/Kanton"},
         {2232, "Europe/Kyiv"},
         {2233, "America/Ciudad_Juarez"},
+        {2234, "America/Coyhaique"},
```
pedroerp, line 2268: this file is automatically generated based on the timezone definition on the presto file listed in the comment section. Is this new entry consistent with the file in presto?

kgpai: Please update here : https://github.com/prestodb/presto/blob/master/presto-common/src/main/resources/com/facebook/presto/common/type/zone-index.properties before making changes here.

ReemaAlzaid:
Hey @kgpai @pedroerp 
the changes has been merged in presto as seen here https://github.com/prestodb/presto/blob/master/presto-common/src/main/resources/com/facebook/presto/common/type/zone-index.properties#L2243

so we can merge this pr

ReemaAlzaid:
Kind Reminder @kgpai @pedroerp

ReemaAlzaid:
@pedroerp

rui-mo (commented):
Thanks for the fix. The Presto fuzzer failure might be related if Velox supports the new timezone, while the used Presto version hasn't included it. In that case, the Presto version used by PrestoQueryRunner would need to be advanced.

zhouyuan:
@kgpai @pedroerp could you please help to take a look? It's simple diff. Thanks.

kgpai (approved):
Thanks !

pratikpugalia:
@ReemaAlzaid Before landing could we bump presto separately to the version(likely 0.297) which supports this change, otherwise this will break fuzzers. 

cc: @kgpai

ReemaAlzaid:
@pratikpugalia bumped up the version

kgpai:
@ReemaAlzaid The bump up of the version should be another PR which we will land before this so that we can rebuild the docker images. Can you create another PR which I will prioritize landing ?

scripts/docker/java.dockerfile
```diff
@@ -15,7 +15,7 @@
 
 # Global arg default to share across stages
 ARG SPARK_VERSION=3.5.1
-ARG PRESTO_VERSION=0.295
+ARG PRESTO_VERSION=0.297
```
kgpai: Create separate PR for this please.

ReemaAlzaid: here is the pr for it
https://github.com/facebookincubator/velox/pull/17430

pedroerp:
@ReemaAlzaid @kgpai I'm assuming we are blocked on this one because the Presto update was reverted?

ReemaAlzaid:
> @ReemaAlzaid @kgpai I'm assuming we are blocked on this one because the Presto update was reverted?

yes

PR #16234, opened 2026-02-04
build(cuda): Add CUDA_VERSION build arg to adapters dockerfile

bdice:
## Summary
We would like to make it easier to configure the CUDA version used in building Velox containers.

This PR adds a `CUDA_VERSION` build arg to the adapters stage in `centos-multi.dockerfile` which allows overriding the CUDA version at build time. As before, it defaults to 12.9.

This mirrors the approach in https://github.com/prestodb/presto/pull/27074.

### DIFF
--- a/scripts/docker/centos-multi.dockerfile
+++ b/scripts/docker/centos-multi.dockerfile
@@ -133,7 +133,7 @@ COPY scripts/setup-centos-adapters.sh /
 ARG CUDA_VERSION
 ENV CUDA_VERSION=${CUDA_VERSION:-12.9}
 
-RUN bash /setup-centos-adapters.sh install_cuda ${CUDA_VERSION} && \
+RUN bash /setup-centos-adapters.sh install_cuda && \
       dnf clean all
 
 RUN bash /setup-centos-adapters.sh install_adapters_deps_from_dnf && \


scripts/docker/centos-multi.dockerfile
```diff
@@ -130,6 +130,9 @@ FROM centos9 AS adapters
 
 COPY scripts/setup-centos-adapters.sh /
 
+ARG CUDA_VERSION
+ENV CUDA_VERSION=${CUDA_VERSION:-12.9}
```
bdice, line 134: This environment variable is read by `bash /setup-centos-adapters.sh install_cuda` in the next step.

kgpai:
@bdice Folks looks like the docker publish scripts arent working : https://github.com/facebookincubator/velox/actions/runs/21961775674/job/63448645977 so your cuda version will not be updated. 
Can you folks fix the script ? I am going to revert this change in the meantime - We can reland this after the fix.

karthikeyann:
@kgpai could you elaborate, what is the issue with the script?

bdice:
Looks like this:
```
ERROR: ghcr.io/facebookincubator/velox-dev:buildx.build.warnings-arm64: not found
```
Nothing I changed here is arm64-specific. Do the images build successfully without this change? It’s not a network error or something temporary?

karthikeyann:
@kgpai Triaged the issue.
This error comes because of `buildx.build.warnings` artifact appears from "[Build ci Image on arm64](https://github.com/facebookincubator/velox/actions/runs/21961775674/job/63440855278)" job, and it is generated from
https://github.com/facebookincubator/velox/actions/runs/21961775674/job/63440855278#step:8:321

It comes from 
```
          "filename": "scripts/docker/centos-multi.dockerfile",
              "line": 167
```
which is
https://github.com/facebookincubator/velox/blob/6dbd60423836567f234afe2f7af4fb2b5c1ad758/scripts/docker/centos-multi.dockerfile#L167-L167

This seems to be unrelated to this line change, only connection is this PR changes the same file `scripts/docker/centos-multi.dockerfile` but different line.

Fix could be either
```
ENV LD_LIBRARY_PATH="/usr/local/lib:/usr/local/lib64:${LD_LIBRARY_PATH:-}"
```
or
```
ENV LD_LIBRARY_PATH=/usr/local/lib:/usr/local/lib64
```

Or Suggested alternative fix in issue https://github.com/facebookincubator/velox/issues/16386

bdice:
To add to @karthikeyann's analysis -- the Dockerfile warning is the trigger, but the reason it breaks the build is a bug in the "Export digests" step of `docker.yml`. The [jq filter](https://github.com/facebookincubator/velox/blob/9ea254547368dd82637c7c1e59f3df2471271482/.github/workflows/docker.yml#L149-L153) iterates over all top-level keys in the bake metadata and creates a file for each one. When buildx emits a `buildx.build.warnings` key in the metadata (due to the `LD_LIBRARY_PATH` lint warning), that key gets turned into a file alongside the real targets (`adapters`, `centos9`, `pyvelox`). Then the [merge job's `for file in *` loop](https://github.com/facebookincubator/velox/blob/9ea254547368dd82637c7c1e59f3df2471271482/.github/workflows/docker.yml#L190) tries to create a multi-arch manifest for it and fails.

Fixing the Dockerfile warning resolves the immediate issue, but the jq filter should also be hardened so that any future buildx metadata key (e.g., `buildx.build.ref`, `buildx.build.provenance`) doesn't cause the same failure. The real image targets all have an `image.name` field in their metadata value -- the filter could select only entries where that field exists:

```jq
'def skip: ["ubuntu", "fedora"];
  . | to_entries[] | select(.value | has("image.name")) |
  .key | split("-")[0] |
  if . as $name | skip | index($name) != null then empty else . end'
```

kgpai:
Cool Thank you @bdice & @karthikeyann ; Not sure when these warnings started appearing in the digest , but as suggested it makes sense to use ld.conf and also fix the jq script.

PR #11149, opened 2024-10-02
refactor: Update Base64 as non-throwing API

Joe-Abraham:
Follow-up PR: https://github.com/facebookincubator/velox/pull/10371

This PR introduces the following changes:

 1. Refactored Base64 APIs as Non-Throwing
	- The Base64 APIs have been refactored to be non-throwing, aligning with the performance optimizations discussed in the Velox blog post: ["Further Optimizing TRY_CAST and TRY"](https://velox-lib.io/blog/optimize-try-more/).
	- By eliminating exceptions in these functions, we aim to significantly improve performance, particularly in cases where a high percentage of rows might trigger exceptions.
	- The primary benefit comes from reducing the overhead of exception handling, leading to more efficient query execution.
 2. Scoped Change to `BinaryFunctions.h`
	- The non-throwing behavior is applied only to function calls originating from `BinaryFunctions.h`.
	- Calls from other files will continue using the throwing version, ensuring that functionality remains unchanged elsewhere.

Changing the `char` to `std::string_view` and `std::string` will be done in subsequent PR.

PR #8652, opened 2024-02-02
feat: Add presto function 'to_base32'

Joe-Abraham:
Add presto function 'to_base32'

- Part of https://github.com/facebookincubator/velox/issues/7294
- Dependencies https://github.com/facebookincubator/velox/pull/8650

PR #16235, opened 2026-02-04
feat: Add to_base32 Presto function

allenshen13:
Adds the `to_base32` Presto scalar function (`varbinary -> varchar`), complementing the existing `from_base32`. Encodes using `Base32::encode()` with the RFC 4648 alphabet (A-Z, 2-7).

### DIFF
--- a/velox/functions/prestosql/tests/BinaryFunctionsTest.cpp
+++ b/velox/functions/prestosql/tests/BinaryFunctionsTest.cpp
@@ -1013,4 +1013,39 @@ TEST_F(BinaryFunctionsTest, fromBase32) {
       fromBase32("NBSWY3DPEB3W64"), "Invalid input length 14");
 }
 
+TEST_F(BinaryFunctionsTest, toBase32) {
+  const auto toBase32 = [&](std::optional<std::string> value) {
+    return evaluateOnce<std::string>("to_base32(cast(c0 as varbinary))", value);
+  };
+
+  // Null handling.
+  EXPECT_EQ(std::nullopt, toBase32(std::nullopt));
+
+  // Empty input.
+  EXPECT_EQ("", toBase32(""));
+
+  // Basic encoding tests covering all trailing-byte counts (1-5 bytes).
+  EXPECT_EQ("ME======", toBase32("a"));
+  EXPECT_EQ("MFRA====", toBase32("ab"));
+  EXPECT_EQ("MFRGG===", toBase32("abc"));
+  EXPECT_EQ("MFRGGZA=", toBase32("abcd"));
+  EXPECT_EQ("MFRGGZDF", toBase32("abcde"));
+
+  // Longer strings.
+  EXPECT_EQ("NBSWY3DPEB3W64TMMQ======", toBase32("hello world"));
+
+  // Special characters and binary data.
+  const auto fromHex = [&](std::optional<std::string> value) {
+    return evaluateOnce<std::string>("from_hex(c0)", value);
+  };
+  EXPECT_EQ("75H36UA=", toBase32(fromHex("FF4FBF50")));
+
+  // Round-trip test: decode(encode(x)) == x.
+  const auto fromBase32 = [&](const std::optional<std::string>& value) {
+    return evaluateOnce<std::string>("from_base32(c0)", VARCHAR(), value);
+  };
+  auto encoded = toBase32("hello world");
+  EXPECT_EQ("hello world", fromBase32(encoded));
+}
+
 } // namespace
--- a/velox/docs/functions/presto/binary.rst
+++ b/velox/docs/functions/presto/binary.rst
@@ -140,6 +140,10 @@ Binary Functions
 
     Computes the 64-bit SpookyHashV2 hash of ``binary``.
 
+.. function:: to_base32(binary) -> varchar
+
+    Encodes ``binary`` into a base32 string representation using the `RFC 4648 <https://www.rfc-editor.org/rfc/rfc4648#section-6>`_ alphabet (A-Z, 2-7).
+
 .. function:: to_base64(binary) -> varchar
 
     Encodes ``binary`` into a base64 string representation.
--- a/velox/functions/prestosql/BinaryFunctions.h
+++ b/velox/functions/prestosql/BinaryFunctions.h
@@ -337,6 +337,18 @@ struct FromBase32Function {
 };
 
 template <typename T>
+struct ToBase32Function {
+  VELOX_DEFINE_FUNCTION_TYPES(T);
+
+  FOLLY_ALWAYS_INLINE void call(
+      out_type<Varchar>& result,
+      const arg_type<Varbinary>& input) {
+    result.append(
+        encoding::Base32::encode(std::string_view(input.data(), input.size())));
+  }
+};
+
+template <typename T>
 struct FromBase64UrlFunction {
   VELOX_DEFINE_FUNCTION_TYPES(T);
 
--- a/velox/functions/prestosql/registration/BinaryFunctionsRegistration.cpp
+++ b/velox/functions/prestosql/registration/BinaryFunctionsRegistration.cpp
@@ -66,6 +66,9 @@ void registerSimpleFunctions(const std::string& prefix) {
   registerFunction<FromBase32Function, Varbinary, Varbinary>(
       {prefix + "from_base32"});
 
+  registerFunction<ToBase32Function, Varchar, Varbinary>(
+      {prefix + "to_base32"});
+
   registerFunction<ToBase64UrlFunction, Varchar, Varbinary>(
       {prefix + "to_base64url"});
   registerFunction<FromBase64UrlFunction, Varbinary, Varchar>(


velox/functions/prestosql/tests/BinaryFunctionsTest.cpp
```diff
@@ -1013,4 +1011,39 @@ TEST_F(BinaryFunctionsTest, fromBase32) {
...
+  EXPECT_EQ("MFRGGZDF", toBase32("abcde"));
+
+  // Longer strings.
+  EXPECT_EQ("NBSWY3DPEB3W64TMMQ======", toBase32("hello world"));
+  EXPECT_EQ("GEZDGNA=", toBase32("1234"));
+  EXPECT_EQ("NZXQ====", toBase32("no"));
+  EXPECT_EQ("O5SQ====", toBase32("we"));
+  EXPECT_EQ("MRRDE===", toBase32("db2"));
+  EXPECT_EQ("MNQWWZI=", toBase32("cake"));
+  EXPECT_EQ("NNSWK3Q=", toBase32("keen"));
```
PingLiuPing: Seems redundant, let's add some special chars for testing.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,211 @@
...
+    if (inputSize == 0) {
+      return 0;
+    }
+
+    // Check if the input string is padded.
+    if (isPadded(input)) {
+      // If padded, ensure that the string length is a multiple of the encoded
+      // block size.
+      if (inputSize % encodedBlockByteSize != 0) {
+        return folly::makeUnexpected(
```
PingLiuPing: VELOX_USER_FAIL

velox/common/encode/tests/BaseEncoderUtilsTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+#include "velox/common/encode/BaseEncoderUtils.h"
+
+#include <gtest/gtest.h>
+#include "velox/common/base/tests/GTestUtils.h"
+
+namespace facebook::velox::encoding {
+
+class BaseEncoderUtilsTest : public ::testing::Test {};
```
PingLiuPing: Tests should be inside anonymous namespace.

PingLiuPing: Are we only testing static method here? If so, let's just use TEST()

velox/common/encode/tests/BaseEncoderUtilsTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+          .error());
+
+  // Single extra byte (invalid for Base64 unpadded).
+  inputSize = 1;
+  EXPECT_EQ(
+      Status::UserError("decode() - invalid input string length."),
+      BaseEncoderUtils::calculateDecodedSize("A", inputSize, 3, 4).error());
+}
+
+TEST_F(BaseEncoderUtilsTest, calculateDecodedSizeBase32) {
```
PingLiuPing: calculateDecodedSizeBase64 and calculateDecodedSizeBase32 repeats lots of code. It can be simplified by adding a helper function.

velox/common/encode/tests/Base32Test.cpp
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+#include "velox/common/encode/Base32.h"
+
+#include <gtest/gtest.h>
+
+namespace facebook::velox::encoding {
+
+class Base32Test : public ::testing::Test {};
```
PingLiuPing: Tests goes to anonymous namespace.

velox/common/encode/tests/Base32Test.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+TEST_F(Base32Test, encode) {
```
PingLiuPing: TEST_F to TEST

velox/common/encode/Base32.h
PingLiuPing, line 55: Do you aware of any specific reason that this method return `folly::Expected<size_t, Status> ` instead of a size?

Joe-Abraham: The intend is to make base32 as non throwing API, which results in better performance, more info in - https://velox-lib.io/blog/optimize-try-more/

Reference PR where improvements are done to base64 https://github.com/facebookincubator/velox/pull/11149.

velox/functions/prestosql/tests/BinaryFunctionsTest.cpp
allenshen13: Added binary data test using fromHex("FF4FBF50") matching the Base64 test pattern

velox/common/encode/BaseEncoderUtils.h
allenshen13: Skipped due to @Joe-Abraham comment.

velox/common/encode/tests/BaseEncoderUtilsTest.cpp
allenshen13: Yep, good point

velox/common/encode/Base32.h
PingLiuPing: @Joe-Abraham  Thank you for the explanation.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,211 @@
...
+  /// @param includePadding Whether to include padding in the output.
+  /// @param binaryBlockByteSize Number of bytes per binary block
+  ///   (3 for Base64, 5 for Base32).
+  /// @param encodedBlockByteSize Number of characters per encoded block
+  ///   (4 for Base64, 8 for Base32).
+  static size_t calculateEncodedSize(
+      size_t inputSize,
+      bool includePadding,
+      int binaryBlockByteSize,
+      int encodedBlockByteSize) {
```
PingLiuPing: How about adding a tiny struct like this:
```

struct CodecBlockSizes {
  int binaryBlockByteSize;
  int encodedBlockByteSize;
};

static constexpr CodecBlockSizes kBase64BlockSizes{3, 4};
static constexpr CodecBlockSizes kBase32BlockSizes{5, 8};
```
This way it can improve the readability when calling `calculateDecodedSize` and `calculateEncodedSize`.
And you can simplify the test based on kBase64BlockSizes and kBase32BlockSizes.
What do you think?

PingLiuPing (approved):
Thanks, LGTM.
Left one nit.

velox/common/encode/BaseEncoderUtils.h
allenshen13: Yep, definitely better that way, changes have been made

velox/common/encode/Base32.cpp
```diff
@@ -63,12 +63,123 @@ constexpr const Base32::ReverseIndex kBase32ReverseIndexTable = {
...
+    // 4 bytes (32 bits) -> 7 chars + 1 padding
+    size_t encodedChars = (inputSize * 8 + 4) / 5;
+
+    for (size_t i = 0; i < encodedChars; ++i) {
+      *outputPointer++ = charset[(block >> (35 - 5 * i)) & 0x1f];
+    }
+
+    // Pad to the next multiple of 8.
+    size_t paddingCount = Base32::kEncodedBlockByteSize - encodedChars;
+    for (size_t i = 0; i < paddingCount; ++i) {
```
czentgr: Instead of individually setting each byte we can use memset to write the padding bytes. That saves the loop.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /// Calculates the decoded output size from encoded input, accounting for
+  /// padding. Also strips padding from 'inputSize' so that the caller can
+  /// iterate over only the non-padding characters.
+  /// @param input The encoded input string (used to detect trailing padding).
+  /// @param inputSize The length of the encoded input. Modified on return to
+  ///   exclude padding characters.
+  /// @param blockSizes The binary and encoded block sizes for the codec.
+  static Expected<size_t> calculateDecodedSize(
```
czentgr: Is there a reason we add this all in the header instead of a cpp file?

allenshen13: Everything I did is based off Joe's PRs: https://github.com/facebookincubator/velox/pull/8650 and https://github.com/facebookincubator/velox/pull/8652. I think it was just grouped with the constexpr template functions. I moved calculateDecodedSize, calculateEncodedSize, reverseLookup to a new BaseEncoderUtils.cpp file.

Joe-Abraham: @allenshen13, I believe it would be a good idea to put the logic in a C++ file instead of keeping everything in a header file.

velox/common/encode/BaseEncoderUtils.h
```diff
@@ -0,0 +1,154 @@
...
+      std::string_view input,
```
Joe-Abraham: In my opinion, it is not logical to modify the input argument to std::string_view if the input type is not changed from const char* to std::string_view throughout the entire flow. In the current code, there are numerous conversions between const char* and std::string_view. Let me know what you think.

Joe-Abraham (commented):
Thanks @allenshen13, I have one suggestion. Otherwise, this looks good to me.

velox/common/encode/Base32.cpp
```diff
@@ -63,12 +65,121 @@ constexpr const Base32::ReverseIndex kBase32ReverseIndexTable = {
...
+  return utils::calculateEncodedSize(inputSize, withPadding, kBase32BlockSizes);
+}
+
+// static
+std::string Base32::encode(std::string_view text) {
+  return encode(text.data(), text.size());
+}
+
+// static
+std::string Base32::encode(const char* input, size_t inputSize) {
```
Joe-Abraham: Is there any usage of this function?

velox/common/encode/Base32.cpp
```diff
@@ -63,12 +65,121 @@ constexpr const Base32::ReverseIndex kBase32ReverseIndexTable = {
...
+        kBase32ReverseIndexTable),
+    "kBase32ReverseIndexTable has incorrect entries.");
+
+// static
+size_t Base32::calculateEncodedSize(size_t inputSize, bool withPadding) {
+  return utils::calculateEncodedSize(inputSize, withPadding, kBase32BlockSizes);
+}
+
+// static
+std::string Base32::encode(std::string_view text) {
```
Joe-Abraham: Is anyone calling this function?

velox/common/encode/Base32.cpp
```diff
@@ -63,12 +65,121 @@ constexpr const Base32::ReverseIndex kBase32ReverseIndexTable = {
...
+      block |= static_cast<uint64_t>(static_cast<uint8_t>(input[i]))
+          << (32 - 8 * i);
+    }
+
+    // Number of encoded characters: ceil(inputBits / 5).
+    // 1 byte (8 bits)  -> 2 chars + 6 padding
+    // 2 bytes (16 bits) -> 4 chars + 4 padding
+    // 3 bytes (24 bits) -> 5 chars + 3 padding
+    // 4 bytes (32 bits) -> 7 chars + 1 padding
+    size_t encodedChars = (inputSize * 8 + 4) / 5;
```
Joe-Abraham: I believe it’s beneficial to introduce variables instead of relying on hardcoded values as arbitrary numbers?

Joe-Abraham (commented):
Thanks @allenshen13, few nits.

czentgr:
@allenshen13 Please also run the expression fuzzer for the function as outlined https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md#adding-scalar-functions - Step 7 and post the result in the PR as a comment.

Joe-Abraham (approved):
Thanks, @allenshen13. LGTM!

allenshen13:
Running expression fuzzer:

to_base32 in isolation: 

> E20260621 22:45:16.698035 3586195 ExpressionFuzzerVerifier.cpp:271] to_base32 234053 100.00% 2908185
> E20260621 22:45:16.698045 3586195 ExpressionFuzzerVerifier.cpp:504] Total test cases: 321656
> E20260621 22:45:16.698056 3586195 ExpressionFuzzerVerifier.cpp:505] Total test cases verified in the reference DB: 321656
> E20260621 22:45:16.698067 3586195 ExpressionFuzzerVerifier.cpp:507] Total test cases failed in both Velox and the reference DB: 0
> E20260621 22:45:16.698078 3586195 ExpressionFuzzerVerifier.cpp:509] Total test cases unsupported in the reference DB: 0

PR #16239, opened 2026-02-04
feat: Add isDefaultNullBehavior API

mbasmanova:
Summary:
Add `isDefaultNullBehavior` function to the Velox function registry API. This function allows callers to check if a registered function has default null behavior (null in any input produces null output).

This API is needed by the Axiom optimizer to determine whether filter predicates can be pushed through outer joins.

Differential Revision: D92271292

### DIFF
--- a/velox/functions/FunctionRegistry.cpp
+++ b/velox/functions/FunctionRegistry.cpp
@@ -109,6 +109,25 @@ std::optional<bool> isDeterministic(const std::string& functionName) {
   return true;
 }
 
+std::optional<bool> isDefaultNullBehavior(const std::string& functionName) {
+  const auto simpleFunctions =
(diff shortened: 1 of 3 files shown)


velox/functions/FunctionRegistry.cpp
```diff
@@ -109,6 +109,25 @@ std::optional<bool> isDeterministic(const std::string& functionName) {
   return true;
 }
 
+std::optional<bool> isDefaultNullBehavior(const std::string& functionName) {
+  const auto simpleFunctions =
+      exec::simpleFunctions().getFunctionSignaturesAndMetadata(functionName);
```
srsuryadev, line 114: Let's address minor nits

PR #16240, opened 2026-02-04
fix: Fix the memory pool hierarchy of writer node

wecharyu:
`insertTableHandle_->writerOptions()` returns a shared `WriterOptions` object. Previously, `memoryPool` was only set when null, so after writer 0 initialized it, later writers reused writer 0's pool. As a result, all writers' memory gets attributed to writer 0's pool while other writers' pools show no usage, making it impossible to identify which writer is consuming memory.

Fix: set `memoryPool` unconditionally, matching the existing pattern for `nonReclaimableSection`.

### DIFF
--- a/velox/connectors/hive/iceberg/tests/CMakeLists.txt
+++ b/velox/connectors/hive/iceberg/tests/CMakeLists.txt
@@ -160,13 +160,8 @@ if(NOT VELOX_DISABLE_GOOGLETEST)
     GTest::gtest_main
   )
 
-  add_executable(
-    velox_hive_writer_options_adapter_test WriterOptionsAdapterTest.cpp
-  )
-  add_test(
-    velox_hive_writer_options_adapter_test
-    velox_hive_writer_options_adapter_test
-  )
+  add_executable(velox_hive_writer_options_adapter_test WriterOptionsAdapterTest.cpp)
+  add_test(velox_hive_writer_options_adapter_test velox_hive_writer_options_adapter_test)
 
   target_link_libraries(
     velox_hive_writer_options_adapter_test
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -606,9 +606,7 @@ std::shared_ptr<dwio::common::WriterOptions> HiveDataSink::createWriterOptions(
     options->schema = getNonPartitionTypes(dataChannels_, inputType_);
   }
 
-  if (options->memoryPool == nullptr) {
-    options->memoryPool = writerInfo_[writerIndex]->writerPool.get();
-  }
+  options->memoryPool = writerInfo_[writerIndex]->writerPool.get();
 
   if (!options->compressionKind) {
     options->compressionKind = insertTableHandle_->compressionKind();
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -606,8 +606,6 @@ std::shared_ptr<dwio::common::WriterOptions> HiveDataSink::createWriterOptions(
     options->schema = getNonPartitionTypes(dataChannels_, inputType_);
   }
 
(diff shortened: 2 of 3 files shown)


wecharyu:
Before: 
image
After:
image

PingLiuPing (approved):
Thanks.
This change should not affect Prestissimo either.

wecharyu:
@PingLiuPing Can we merge this patch if no more concerns?

PingLiuPing:
@wecharyu Would you please rebase the branch to latest main and run the CI again. Current CI is red.

PingLiuPing:
@wecharyu I see CI still red, I have restarted the CI. If the CI fail again please rebase the code.

velox/connectors/hive/iceberg/tests/CMakeLists.txt
```diff
@@ -160,13 +160,8 @@ if(NOT VELOX_DISABLE_GOOGLETEST)
     GTest::gtest_main
   )
 
-  add_executable(
```
wecharyu, line 163: fix: https://github.com/facebookincubator/velox/actions/runs/25773932792/job/75702596538

PingLiuPing:
@Yuhta @mbasmanova Could you help take a look at this PR?

wecharyu:
@PingLiuPing CI is green after rebase.

PingLiuPing:
> @PingLiuPing CI is green after rebase.

Yes, let's wait component maintainers to have a final look.

mbasmanova:
Thank you for the fix. The change looks good, a few comments.

The PR description is a bit vague. Consider something like:

> `insertTableHandle_->writerOptions()` returns a shared_ptr to the same `WriterOptions` object. When `createWriterOptions` is called for the first writer, it sets `memoryPool` on this shared object. For subsequent writers, the null guard (`if (options->memoryPool == nullptr)`) prevents updating `memoryPool` to the correct per-writer pool. As a result, all writers' memory gets attributed to writer 0's pool while other writers' pools show no usage, making it impossible to identify which writer is consuming memory. Fix: set `memoryPool` unconditionally, matching the existing pattern for `nonReclaimableSection`.

The `memoryPool` assignment is currently in the "Only overwrite options in case they were not already provided" section, but it's now set unconditionally. Consider moving it next to `nonReclaimableSection` (line 624), which is also set unconditionally for the same reason, and updating the comment there to cover both per-writer fields.

Is it possible to add a test for this? If not, could you describe how you tested the change in the PR description and file a GitHub issue to extend the testing framework to allow verifying memory pool hierarchy in multi-writer scenarios? CC: @tanjialiang @xiaoxmeng

wecharyu:
Added a unit test for this change, pls take a look again. Thanks. @mbasmanova

mbasmanova (approved):
Thank you for adding the test! Can you confirm the test fails without the fix? A couple of nits.

Inline the vectors into `makeRowVector` to avoid naming them, and simplify the lambdas:
```cpp
dataSink->appendData(makeRowVector({
    makeFlatVector<int64_t>(200, folly::identity),
    makeFlatVector<StringView>(
        200, [](auto row) { return fmt::format("part_{}", row % 2); }),
}));
```

Test name `sharedWriterOptionsUsePerWriterMemoryPool` → `perWriterMemoryPool`. Also, `sharedWriterOptions` → `writerOptions` or just `options`. `dwrfPool` — consider a format-neutral name (e.g. `innerPool`) since the fix applies to all writer formats.

wecharyu:
Yes, test would fail before this fix. @mbasmanova 
```bash
[ RUN      ] HiveDataSinkTest.perWriterMemoryPool
I0513 22:33:20.771877 3876351 Compression.cpp:660] Initialized zstd compressor with compression level 3
/root/workspace/velox/velox/connectors/hive/tests/HiveDataSinkTest.cpp:1988: Failure
Expected equality of these values:
  writerPoolNames.size()
    Which is: 1
  2

[  FAILED  ] HiveDataSinkTest.perWriterMemoryPool (16 ms)
```

PR #16243, opened 2026-02-05
feat: Add BaseVector::createLike to support empty vector creating matching source vector

peterenescu:
Summary: Adds function `BaseVector::createLike` to BaseVector's vector creation suite to allow empty vector creation matching a source vector. This will help in creating template vectors preserving flat map encoding and will adds capability for other encoding preserve template vectors.

Differential Revision: D92333447

### DIFF
--- a/velox/vector/BaseVector.h
+++ b/velox/vector/BaseVector.h
@@ -746,6 +746,17 @@ class BaseVector {
     return vector ? vector : create(type, 0, pool);
   }
 
+  /// Creates a vector matching the structure of a source vector, preserving
+  /// FLAT_MAP encoding where the source uses it. For nested types (ROW, ARRAY,
+  /// MAP), it recursively handles children to preserve encoding at each level.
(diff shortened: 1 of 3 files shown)


velox/vector/BaseVector.h
```diff
@@ -746,6 +746,15 @@ class BaseVector {
     return vector ? vector : create(type, 0, pool);
   }
 
+  /// Creates a vector matching the structure of a source vector, preserving
+  /// FLAT_MAP encoding where the source uses it. For nested types (ROW, ARRAY,
+  /// MAP), it recursively handles children to preserve encoding at each level.
+  /// BaseVector::create builds from type, without reference to encodings.
```
pedroerp: can you also clarify that this will still flatten dictionary, constant, and other encodings?

pedroerp (approved):

velox/vector/BaseVector.h
peterenescu: Yes good call, thanks!

PR #16248, opened 2026-02-05
feat: Add ANSI mode support for Spark CAST(string as long/int/short/byte)

malinjawi:
Add ANSI mode validation for casting VARCHAR to BIGINT, INTEGER, SMALLINT, and 
TINYINT in Spark dialect.

In ANSI mode:
- Decimal strings (e.g., '123.45') throw errors instead of truncating
- Invalid strings throw errors instead of returning NULL
- Overflow conditions throw errors instead of returning NULL

In non-ANSI mode (default):
- Decimal strings are truncated to integers
- Invalid strings return NULL
- Overflow returns NULL

Key changes:
- Update `isAnsiSupported() `to return true for string-to-integer casts
- Set `isTryCast=false` when ANSI mode is enabled for supported casts
- Add comprehensive tests covering all 4 integer types
- Mark cast as ANSI compliant in documentation

Test coverage matches Spark's `CastWithAnsiOffSuite` and `CastWithAnsiOnSuite`:
- Valid integers (with signs, whitespace)
- Decimal strings (truncate vs throw)
- Invalid strings (empty, non-numeric, mixed)
- Overflow cases (all integer type boundaries)

Addresses: [apache/incubator-gluten#10134](https://github.com/apache/incubator-gluten/issues/10134)

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -83,6 +83,24 @@ line width, indentation and ordering (for includes, using directives and etc). 
   code paths, and are not recommended for use in production environments. For
   example, `debug_disable_expression_with_peeling` is used to disable peeling
   optimization employed in expression evaluation.
+* **Do not abbreviate** variable, function, or class names. Use full,
+  descriptive names. Clarity is more important than brevity.
+
+  | ❌ Avoid | ✅ Prefer |
+  |----------|-----------|
+  | `outputCol` | `outputColumn` |
+  | `val` | `value` |
+  | `idx` | `index` |
+  | `agg` | `aggregation` |
+  | `sel` | `selectivity` |
+  | `rowCount` | `numRows` |
+
+  * Well-established abbreviations in the domain are acceptable (e.g., `id`,
+    `url`, `sql`, `expr`).
+  * Loop indices like `i`, `j`, `k` are acceptable for simple loops.
+  * Iterator variables named `it` are acceptable.
+  * `numXxx` naming pattern is acceptable and preferred over `xxxCount` or
+    `xxxCnt` (e.g., `numRows`, `numKeys`).
 
 ## Comments
 
@@ -154,8 +172,36 @@ About comment style:
 
   * Include enough context in the comment itself to make clear what will be
     done, without requiring any references from outside the code.
-  * Do not include the author’s username. If required, this can always be
+  * Do not include the author's username. If required, this can always be
     retrieved from git blame.
+* **Start comments with active verbs**, not "This class…" or "This method…".
+
+  | ❌ Avoid | ✅ Prefer |
+  |----------|-----------|
+  | `/// This class builds query plans.` | `/// Builds query plans.` |
+  | `/// This method computes selectivity.` | `/// Computes selectivity.` |
+  | `/// This function returns the cost.` | `/// Returns the cost.` |
+
+* **Avoid redundant comments** that simply repeat what the code already says.
+  Comments should explain *why*, not *what*.
+
+```cpp
+// ❌ Avoid - comment just repeats the code
+// Increment the counter.
+++counter;
+
+// ❌ Avoid - comment states the obvious
+// Return the result.
+return result;
+
+// ✅ Prefer - no comment needed when code is self-explanatory
+++counter;
+return result;
+
+// ✅ Prefer - comment explains WHY, not WHAT
+// Use a larger buffer to avoid repeated reallocations for typical queries.
+buffer.reserve(1024);
+```
 
 ## Asserts and CHECKs
 
@@ -234,8 +280,14 @@ About comment style:
 
 * Always use `nullptr` if you need a constant that represents a null pointer
   (`T*` for some `T`); use `0` otherwise for a zero value.
-* For large literal numbers, use ‘ to make it more readable, e.g:  `1’000’000`
-  instead of `1000000`.
+* For large literal numbers, use `'` to improve readability. Use digit
+  separators for numbers with 4 or more digits.
+
+  | ❌ Avoid | ✅ Prefer |
+  |----------|-----------|
+  | `10000` | `10'000` |
+  | `1000000` | `1'000'000` |
+  | `100` | `100` (no separator needed) |
 * For floating point literals, never omit the initial 0 before the decimal
   point (always `0.5`, not `.5`).
 * File level variables and constants should be defined in an anonymous
@@ -416,3 +468,15 @@ using TypePtr = std::shared_ptr<const Type>;
 using ContinueFuture = folly::SemiFuture<bool>;
 using ContinuePromise = VeloxPromise<bool>;
 ```
+
+## API Design
+
+* **Keep the public API surface small.** Do not expose methods or types that
+  callers don't need. Fewer public symbols mean less coupling and easier
+  evolution.
+* **Prefer free functions in .cpp over class methods.** If a helper doesn't
+  need access to class state or doesn't need to be called from outside the
+  translation unit, make it a free function (typically in an anonymous
+  namespace) rather than a static or private method.
+* **Keep method implementations in .cpp.** Except for trivial one-liners,
+  define methods in the .cpp file to keep headers small and reduce build times.
--- a/python/test/test_debugger_runner.py
+++ b/python/test/test_debugger_runner.py
@@ -254,3 +254,60 @@ class TestPyVeloxDebuggerRunner(unittest.TestCase):
         self.assertEqual(len(received_sizes), self._num_batches)
         # Each vector should have batch_size rows.
         self.assertTrue(all(s == self._batch_size for s in received_sizes))
+
+    def test_step_with_plan_id(self):
+        """step(plan_id) should only stop at the matching breakpoint."""
+        runner = LocalDebuggerRunner(self._plan_node)
+        runner.set_breakpoint(self._node_ids[3])
+        runner.set_breakpoint(self._node_ids[5])
+        runner.set_breakpoint(self._node_ids[8])
+
+        it = runner.execute()
+
+        # Step targeting node_ids[5] should skip node_ids[3] and stop at
+        # node_ids[5].
+        it.step(self._node_ids[5])
+        self.assertEqual(it.at(), self._node_ids[5])
+
+        # Step targeting node_ids[8] should skip node_ids[5] (remaining) and
+        # stop at node_ids[8].
+        it.step(self._node_ids[8])
+        self.assertEqual(it.at(), self._node_ids[8])
+
+        # Step with no filter (default) stops at the next breakpoint or task
+        # output.
+        it.step()
+        self.assertEqual(it.at(), "")
+
+    def test_step_with_plan_id_counts(self):
+        """step(plan_id) should only produce vectors from the matching
+        breakpoint plus task outputs."""
+        runner = LocalDebuggerRunner(self._plan_node)
+        runner.set_breakpoint(self._node_ids[3])
+        runner.set_breakpoint(self._node_ids[5])
+
+        it = runner.execute()
+        total_size = 0
+
+        # Only step to node_ids[5], skipping node_ids[3].
+        while True:
+            try:
+                vector = it.step(self._node_ids[5])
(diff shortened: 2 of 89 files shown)


velox/docs/functions/spark/conversion.rst
```diff
@@ -80,9 +80,11 @@ Valid examples
 From strings
 ^^^^^^^^^^^^
 
+*(ANSI compliant)*
+
 Casting a string to an integral type is allowed if the string represents a number within the range of result type.
-Casting from strings that represent floating-point numbers truncates the decimal part of the input value.
-Casting from invalid input values throws.
+Casting from strings that represent floating-point numbers truncates the decimal part of the input value when ANSI mode is disabled.
+Casting from invalid strings throws an error when ANSI mode is enabled, or returns NULL when ANSI mode is disabled.
```
philo-he: Suggest revising it for brevity:

Casting from invalid strings throws an error when ANSI mode is enabled; returns NULL otherwise.

velox/type/Type.h
```diff
@@ -657,6 +657,12 @@ class Type : public Tree<const TypePtr>, public velox::ISerializable {
 
   bool containsUnknown() const;
 
+  /// Returns true if this type is any integer type (TINYINT, SMALLINT, INTEGER,
+  /// or BIGINT).
+  bool isAnyInteger() const {
+    return isTinyint() || isSmallint() || isInteger() || isBigint();
```
philo-he: It seems this is essentially a check for the underlying implementation type, not the logical type. For example, `isInteger()` may also return true for Date type since the underlying implementation type is integer. Could you please confirm it?

malinjawi: Thanks! Updated the docs to the shorter phrasing, and removed i`sAnyInteger()` since it was based on underlying storage kinds. In `SparkCastExpr` I now explicitly gate ANSI support to true integer types (`TINYINT/SMALLINT/INTEGER/BIGINT`) and exclude `DATE/TIME/INTERVAL` to avoid logical-type overlap. Let me know if this works.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -17,13 +17,29 @@
...
+namespace {
+
+bool isSparkIntegralType(const TypePtr& type) {
+  if (type->isTinyint() || type->isSmallint()) {
+    return true;
+  }
+  if (type->isInteger()) {
+    return !type->isDate() && !type->isIntervalYearMonth();
+  }
+  if (type->isBigint()) {
```
rui-mo: nit: perhaps exclude decimal also.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -17,13 +17,29 @@
...
 
 namespace facebook::velox::functions::sparksql {
+namespace {
+
+bool isSparkIntegralType(const TypePtr& type) {
+  if (type->isTinyint() || type->isSmallint()) {
+    return true;
+  }
+  if (type->isInteger()) {
+    return !type->isDate() && !type->isIntervalYearMonth();
```
rui-mo: Can we compare objects like `type == INTEGER()` to avoid the comparison between type kinds?

malinjawi: Updated the check to use pointer equality (`type == TINYINT()/SMALLINT()/INTEGER()/BIGINT()`), which avoids kind-based checks and implicitly excludes `DATE/TIME/INTERVAL` and `decimal`.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -46,21 +54,27 @@ exec::ExprPtr SparkCastCallToSpecialForm::constructSpecialForm(
...
+
+  // For string-to-integer casts with ANSI enabled, we need to disallow
+  // decimal points. This is controlled by allowOverflow: when false, it
+  // uses SparkTryCastPolicy (truncate=false, no decimals allowed).
   // The distinction between CAST (ANSI off) and TRY_CAST is limited to
   // overflow handling, which is managed by the 'allowOverflow' flag in
   // SparkCastHooks.
-  const bool isTryCast =
-      !config.sparkAnsiEnabled() || !isAnsiSupported(fromType, type);
+  const bool allowOverflow = !(config.sparkAnsiEnabled() && ansiSupported);
```
philo-he: This is equivalent to `isTryCast`. Can we directly use it instead?

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -17,13 +17,21 @@
...
+
+} // namespace
 
 bool SparkCastCallToSpecialForm::isAnsiSupported(
     const TypePtr& fromType,
     const TypePtr& toType) {
-  // String to Boolean cast supports ANSI mode.
-  if (fromType->isVarchar() && toType->isBoolean()) {
-    return true;
+  // String to Boolean or Integer types support ANSI mode
```
philo-he: Nit: keep using a period at the end.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -46,21 +54,27 @@ exec::ExprPtr SparkCastCallToSpecialForm::constructSpecialForm(
       compiledChildren.size());
 
   const auto& fromType = compiledChildren[0]->type();
+  const bool ansiSupported = isAnsiSupported(fromType, type);
```
philo-he: If `ansiSupported` is used only once after code update, maybe we can remove it.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -17,13 +17,21 @@
 #include "velox/functions/sparksql/specialforms/SparkCastExpr.h"
 
 namespace facebook::velox::functions::sparksql {
+namespace {
+
+bool isSparkIntegralType(const TypePtr& type) {
```
philo-he: suggest renaming it to `isIntegralType`, since the code logic is not specific to Spark.

velox/docs/functions/spark/conversion.rst
```diff
@@ -80,9 +80,11 @@ Valid examples
 From strings
 ^^^^^^^^^^^^
 
+*(ANSI compliant)*
+
 Casting a string to an integral type is allowed if the string represents a number within the range of result type.
-Casting from strings that represent floating-point numbers truncates the decimal part of the input value.
-Casting from invalid input values throws.
+Casting from strings that represent floating-point numbers truncates the decimal part of the input value when ANSI mode is disabled.
+Casting from invalid strings throws an error when ANSI mode is enabled; returns NULL otherwise.
```
philo-he: It may be ambiguous to readers who may feel that invalid strings also contain floating-point-number sting.

Suggestion:
```
Casting from strings that represent floating-point numbers truncates the
decimal part of the input value when ANSI mode is disabled; throws an
error otherwise.

Casting from other invalid strings returns NULL when ANSI mode is disabled;
throws an error otherwise.
```

velox/functions/sparksql/tests/SparkCastExprTest.cpp
```diff
@@ -1145,5 +1145,105 @@ TEST_F(SparkCastExprTest, recursiveTryCast) {
       }));
 }
 
+TEST_F(SparkCastExprTest, stringToIntegerAnsi) {
```
philo-he: Could you check if the added test cases overlap with the existing tests: `primitiveValidCornerCases` and `fromString`?

Maybe, we can merge the test and modify `primitiveValidCornerCases`, `fromString` to get relevant cases tested under different ANSI settings.

PHILO-HE (commented):
Some minor comments. Thanks.

velox/functions/sparksql/tests/SparkCastExprTest.cpp
malinjawi: Thanks for the feedback @PHILO-HE ! 

All comments addressed:
- Replaced `allowOverflow` with `isTryCast`
- Added period to comment
- Removed `ansiSupported` variable
- Renamed to `isIntegralType`
- Improved docs - separated floating-point vs invalid strings
- Merged tests into `primitiveValidCornerCases` and `primitiveInvalidCornerCases`

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -17,18 +17,21 @@
...
-  if (fromType->isVarchar() && toType->isBoolean()) {
-    return true;
-  }
-
-  // String to Date cast supports ANSI mode.
-  if (fromType->isVarchar() && toType->isDate()) {
-    return true;
+  // String to Boolean, Integer, or Date types support ANSI mode.
+  if (fromType->isVarchar()) {
+    return toType->isBoolean() || isIntegralType(toType) || toType->isDate();
```
philo-he: Can we enable date type and its relevant tests in another PR?

malinjawi: Thanks for the review @PHILO-HE ! Just to clarify - Date type ANSI support was already merged in a [previous PR](https://github.com/facebookincubator/velox/pull/16092). This refactoring just combined the existing separate if-statements into one line.

Since Date support already exists and works, would you still prefer to remove it from the refactored code? Happy to do so to keep scope tight, just want to confirm since it would need to be re-added later.

philo-he: @malinjawi, I see. Then, it makes sense to do this change. Please kindly ignore my comment at the above. Thanks for the clarification.

PHILO-HE (approved):
Looks good. cc @rui-mo

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -46,21 +49,25 @@ exec::ExprPtr SparkCastCallToSpecialForm::constructSpecialForm(
...
   // the specific cast operation doesn't support ANSI mode.
-  // The distinction between CAST (ANSI off) and TRY_CAST is limited to
-  // overflow handling, which is managed by the 'allowOverflow' flag in
-  // SparkCastHooks.
   const bool isTryCast =
       !config.sparkAnsiEnabled() || !isAnsiSupported(fromType, type);
 
+  // For string-to-integer casts with ANSI enabled, we need to disallow
+  // decimal points. This is controlled by allowOverflow: when false, it
+  // uses SparkTryCastPolicy (truncate=false, no decimals allowed).
```
rui-mo: I would suggest avoiding comments tied to a specific cast here, since this flag applies to all casts. Perhaps move above to the certain cast.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
```diff
@@ -71,13 +78,19 @@ exec::ExprPtr SparkTryCastCallToSpecialForm::constructSpecialForm(
   VELOX_CHECK_EQ(
       compiledChildren.size(),
       1,
-      "TRY CAST statements expect exactly 1 argument, received {}.",
+      "TRY_CAST statements expect exactly 1 argument, received {}.",
       compiledChildren.size());
+
+  // TRY_CAST always uses allowOverflow=false to:
+  // 1. Return NULL on overflow (instead of wrapping)
+  // 2. Return NULL on invalid string formats like "125.5" for integers
```
rui-mo: Part of the above description is true only for the string-to-int cast. Perhaps keep the general comments here, and move the specific description to certain cast.

velox/functions/sparksql/specialforms/SparkCastExpr.cpp
malinjawi: Thanks @rui-mo . I’ve removed the string‑specific wording from the general `allowOverflow / TRY_CAST `comments, and moved the string‑to‑integral explanation into the `isAnsiSupported()`

rui-mo (approved):
Thanks!

PR #16253, opened 2026-02-05
feat(prestosql): Add ceil(DECIMAL) PrestoSQL function

simoneves:
For whatever, reason the Velox CPU/C++/Prestissimo implementation of `CEIL()` for `DECIMAL` types was never done.

This PR adds that, plus associated tests, which will also allow a corresponding implementation in Velox GPU CUDF.

The implementation is basically an adaptation of `FLOOR()` (which WAS done) with math based on the Java version.


velox/functions/prestosql/DecimalFunctions.cpp
```diff
@@ -561,6 +585,35 @@ void registerDecimalFloor(const std::string& prefix) {
       ShortDecimal<P1, S1>>({prefix + "floor"}, constraints);
 }
 
+void registerDecimalCeil(const std::string& prefix) {
+  std::vector<exec::SignatureVariable> constraints = {
+      exec::SignatureVariable(
+          P2::name(),
+          fmt::format(
+              "min(38, {p} - {s} + min({s}, 1) + 1)", // @TODO seves 1/28/26 verify the +1
```
simoneves: Anyone more familiar with the precision expression mechanism, please confirm or deny my naive assumption that, unlike `floor`, one would always need to leave space for one more digit, because `ceil(9.1) = 10.0` etc.

jinchengchenghh:
Could you help review this PR? Thanks! @mbasmanova ws

jinchengchenghh:
I'm familiar with SparkSql functions, so I ask @mbasmanova to help review Presto functions, @simoneves

mbasmanova (commented):
@simoneves Please, review https://github.com/facebookincubator/velox/blob/main/CONTRIBUTING.md#adding-scalar-functions and follow.

Please, update PR title to: feat: Add ceil(DECIMAL) Presto function

velox/functions/prestosql/DecimalFunctions.cpp
```diff
@@ -288,6 +288,30 @@ struct DecimalFloorFunction {
   uint8_t scale_;
 };
 
+template <typename TExec>
+struct DecimalCeilFunction {
```
mbasmanova, line 291: Please, add documentation to https://facebookincubator.github.io/velox/functions/presto/decimal.html

velox/functions/prestosql/DecimalFunctions.cpp
mbasmanova: @simoneves Please, consult Presto's implementation: presto-main-base/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java

```
    @ScalarFunction(value = "ceiling", alias = "ceil")
    @Description("round up to nearest integer")
    public static final class Ceiling
    {
        private Ceiling() {}

        @LiteralParameters({"p", "s", "rp"})
        @SqlType("decimal(rp,0)")
        @Constraint(variable = "rp", expression = "p - s + min(s, 1)")
        public static long ceilingShort(@LiteralParameter("s") long numScale, @SqlType("decimal(p, s)") long num)
        {
            long rescaleFactor = Decimals.longTenToNth((int) numScale);
            long increment = (num % rescaleFactor > 0) ? 1 : 0;
            return num / rescaleFactor + increment;
        }

        @LiteralParameters({"p", "s", "rp"})
        @SqlType("decimal(rp,0)")
        @Constraint(variable = "rp", expression = "p - s + min(s, 1)")
        public static Slice ceilingLong(@LiteralParameter("s") long numScale, @SqlType("decimal(p, s)") Slice num)
        {
            Slice tmp;
            if (isNegative(num)) {
                tmp = add(num, DECIMAL_HALF_UNSCALED_FOR_SCALE[(int) numScale]);
            }
            else {
                tmp = add(num, DECIMAL_ALMOST_HALF_UNSCALED_FOR_SCALE[(int) numScale]);
            }
            return rescale(tmp, -(int) numScale);
        }

        @LiteralParameters({"p", "s", "rp"})
        @SqlType("decimal(rp,0)")
        @Constraint(variable = "rp", expression = "p - s + min(s, 1)")
        public static long ceilingLongShort(@LiteralParameter("s") long numScale, @SqlType("decimal(p, s)") Slice num)
        {
            return unscaledDecimalToUnscaledLong(ceilingLong(numScale, num));
        }
    }
```

simoneves: That would have been the smart thing to do... (blush) Thanks for the nudge.

simoneves:
> Please, update PR title to: feat: Add ceil(DECIMAL) Presto function

@mbasmanova I do not seem to have permissions to update the title. There is not the usual [Edit] button. I have noticed this with both Velox and Presto upstream PRs that I have made.

mbasmanova:
> @mbasmanova I do not seem to have permissions to update the title. There is not the usual [Edit] button. I have noticed this with both Velox and Presto upstream PRs that I have made.

Could you check if you are logged in to GitHub? You may be not.

velox/functions/sparksql/registration/RegisterMath.cpp
```diff
@@ -80,6 +80,7 @@ void registerMathFunctions(const std::string& prefix) {
   registerFunction<sparksql::FloorFunction, int64_t, double>(
       {prefix + "floor"});
   registerDecimalFloor(prefix);
+  registerDecimalCeil(prefix);
```
rui-mo: @simoneves Just FYI. Spark already has the decimal ceil implementation: velox/functions/sparksql/DecimalCeil.cpp.

simoneves:
> Could you check if you are logged in to GitHub? You may be not.

I was, but now I see the [Edit] button, and it looks like somebody changed the title for me? Thank you...

mbasmanova:
@simoneves I updated the title for you.

velox/functions/sparksql/registration/RegisterMath.cpp
simoneves: Oops. Duplicate removed. Thank you.

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: Done

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: Done

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: @mbasmanova I made the math and the precision expression match the Java, but I remain unconvinced that the precision expression is correct in the first place. As I said before, what about the 9.x -> 10.0 case where one extra digit is required? What am I missing?

simoneves:
Also adding `DecimalArithmeticTest.ceil`  (WIP)

velox/functions/prestosql/DecimalFunctions.cpp
mbasmanova: @simoneves I can take a look sometime later, but not sure exactly when. To unblock, I suggest to (1) discuss with Claude; (2) open an issue in PrestoDB repo to discuss in a broader forum. Hope this helps.

simoneves:
Unit test added. Ready for re-review.

mbasmanova:
> Unit test added. Ready for re-review.

@simoneves Don't we need to resolve https://github.com/facebookincubator/velox/pull/16253#discussion_r2775635086 before proceeding with the review?

simoneves:
> @simoneves Don't we need to resolve [#16253 (comment)] before proceeding with the review?

Fair. This one does not block the DECIMAL PR. I will open a Presto issue for discussion.

simoneves:
@mbasmanova after re-reading the code and the docs, I have realized my mistake. Actually two mistakes.

Since `floor()` and `ceil()` actually round towards -INF and +INF respectively, they BOTH would naively seem to require an additional digit in some cases. However, since the output value is an integer, the implementations actually return a value with a scale of zero, and hence that additional digit slot is never actually required, as any fractional part in the original (which would be at least one digit) does not need to be retained. The precision expression `p - s + min(s, 1)` is indeed correct in both cases.

Actually, three mistakes, because my CUDF implementation in the other PR does not currently implement the above behavior properly, so I'll go and fix that next!

I therefore retract my concern in this PR and re-declare it ready for re-review! :)

mbasmanova:
@simoneves Thank you for digging in and resolving the question. CI is red. Would you take a look?

velox/docs/functions/presto/decimal.rst
```diff
@@ -257,6 +257,14 @@ Decimal Functions
 
         pr = min(38, p - s + min(s, 1))
 
+.. function:: ceil(x: decimal(p, s)) -> r: decimal(pr, 0)
```
mbasmanova: Make sure functions are listed in alphabetical order.

velox/functions/prestosql/DecimalFunctions.cpp
```diff
@@ -288,6 +288,30 @@ struct DecimalFloorFunction {
...
+      const std::vector<TypePtr>& inputTypes,
+      const core::QueryConfig& /*config*/,
+      A* /*a*/) {
+    scale_ = getDecimalPrecisionScale(*inputTypes[0]).second;
+  }
+
+  template <typename R, typename A>
+  void call(R& out, const A& a) {
+    const auto rescaleFactor = DecimalUtil::kPowersOfTen[scale_];
+    // the reverse of floor
```
mbasmanova: Make sure all comments are full sentences. Start with a capital letter and end with a period.

velox/functions/prestosql/DecimalFunctions.cpp
```diff
@@ -288,6 +288,30 @@ struct DecimalFloorFunction {
   uint8_t scale_;
 };
 
+template <typename TExec>
+struct DecimalCeilFunction {
+  VELOX_DEFINE_FUNCTION_TYPES(TExec);
```
mbasmanova, line 292: not needed

velox/functions/prestosql/DecimalFunctions.cpp
```diff
@@ -561,6 +585,35 @@ void registerDecimalFloor(const std::string& prefix) {
       ShortDecimal<P1, S1>>({prefix + "floor"}, constraints);
 }
 
+void registerDecimalCeil(const std::string& prefix) {
```
mbasmanova: Is this a near-copy-paste of registerDecimalFloor? Perhaps, extract a reusable template helper.

```
registerDecimalFloorOrCeil<DecimalCeilFunction>(prefix + "ceil")
```

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
```diff
@@ -637,6 +637,77 @@ TEST_F(DecimalArithmeticTest, floor) {
           DECIMAL(19, 19))});
 }
 
+TEST_F(DecimalArithmeticTest, ceil) {
```
mbasmanova: This test is not very readable. Would you clarify what is being tested and how to read the hard-coded values? Is there a way to improve readability?

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: WIP (non-compiling) implementation unintentionally pushed. Ignore.

simoneves:
@mbasmanova all done except for your point about the test I added. All I did was copy the `floor` test that's right above it, and then the numbers from the Spark equivalent test. In other words, not to argue, but it's far from the only test with that level of (un-)readability. They could probably all use more comments.

velox/docs/functions/presto/decimal.rst
simoneves: Done

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: Done

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: (see below)

velox/functions/prestosql/DecimalFunctions.cpp
simoneves: Done

simoneves:
I meant to do a merge-update on this branch, as it was stale, but by force of habit, I did a re-base onto main instead.

Apologies for breaking protocol for PRs in active review.

Other than the open question about whether it should be in the remit of this PR to reformat and comment the tests beyond the one I added, I believe this to be final.

mbasmanova:
> All I did was copy the floor test that's right above it, and then the numbers from the Spark equivalent test. 

Copy-paste is an anti-pattern. Let's not do that (even if someone did that before).

It sounds like tests for floor and ceil could be combined. That could help with readability as well. I'd expect something like this:

ASSERT_EQ(floor(10.3), 10);
ASSERT_EQ(ceil(10.3), 11);
...

simoneves:
@mbasmanova OK, I combined the tests

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
```diff
@@ -565,76 +565,104 @@ TEST_F(DecimalArithmeticTest, roundN) {
...
-  testDecimalExpr<TypeKind::BIGINT>(
-      {makeFlatVector<int64_t>(
-          {DecimalUtil::kShortDecimalMax, DecimalUtil::kShortDecimalMin},
-          DECIMAL(18, 0))},
-      "floor(c0)",
-      {makeFlatVector<int64_t>(
-          {DecimalUtil::kShortDecimalMax, DecimalUtil::kShortDecimalMin},
-          DECIMAL(18, 0))});
+  // inputs
+  auto const ssIn1 = makeFlatVector<int64_t>(
```
mbasmanova: ssIn1, ssIn2,... - what do these names mean? Can we use more descriptive names?

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
```diff
@@ -565,76 +565,104 @@ TEST_F(DecimalArithmeticTest, roundN) {
...
+  auto const ssOutC1 =
+      makeFlatVector<int64_t>({0, 1, 0, 1, 0, 1, 0, 1, 0}, DECIMAL(2, 0));
+  auto const ssOutC2 = makeFlatVector<int64_t>(
+      {123, -123, 124, -123, 124, -123, 124, -123, 124, -123, 124, -123},
+      DECIMAL(4, 0));
+  auto const ssOutC3 = makeFlatVector<int64_t>(
+      {DecimalUtil::kShortDecimalMax, DecimalUtil::kShortDecimalMin},
+      DECIMAL(18, 0));
+  // test
+  testDecimalExpr<TypeKind::BIGINT>({ssOutF1}, "floor(c0)", {ssIn1});
```
mbasmanova: Let's re-write the test to test one value / row at a time. Batch testing makes it hard to read / understand the test.

simoneves: Respectfully, the other 12 tests in this file all have this structure, as do the equivalent ones for Spark. That doesn't make it right, but I do not believe it should be in the remit of this PR, which was supposed to be a quick fix for something missing in Velox CPU and arguably outside the remit of CUDF decimal support, to change that beyond what I have already done, which would then make it different from all the others. By all means let's raise an issue to refactor ALL of them, but as a separate task and PR.

mbasmanova: The problem is that I cannot review this test because I cannot understand it. Sorry, I cannot just stamp the PR without making sure the code changes are correct and test coverage is sufficient. I would imagine Claude can re-write the test in no time. Hence, I don't understand the push back.

simoneves: The alternative is just to not add a test for `ceil` here at all, because it is already tested in the Spark tests, and indeed there is actually no existing `floor` test in that group.

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
simoneves: Done

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
simoneves: I retract that last suggestion. The function implementation in the Spark case is also separate, so that test does NOT test this new implementation for Presto.

simoneves: It does raise the question as to why Presto Velox ONLY had `floor` and Spark Velox ONLY had `ceil` :)

simoneves:
@mbasmanova I have rewritten the test in a new style, and opened a new issue #16464 to refactor the others in the same style once approved. It won't let me assign that issue to myself, but please feel free to.

velox/functions/prestosql/tests/DecimalArithmeticTest.cpp
```diff
@@ -565,76 +583,164 @@ TEST_F(DecimalArithmeticTest, roundN) {
...
+TEST_F(DecimalArithmeticTest, floorAndCeil) {
+  // short DECIMAL(3,2) -> short DECIMAL(2.0)
+  {
+    const auto inType = DECIMAL(3, 2);
+    const auto outType = DECIMAL(2, 0);
+    auto testFloor = makeDecimalExprTester<int64_t, TypeKind::BIGINT>(
+        inType, outType, "floor(c0)");
+    auto testCeil = makeDecimalExprTester<int64_t, TypeKind::BIGINT>(
+        inType, outType, "ceil(c0)");
+    testFloor(0, 0);
```
mbasmanova, line 602: Thank you for re-writing the tests.

How should I read this? 

x in (0, 100): floor is 0, ceil is 1
x = 0: floor = ceil = 0
x in (-100, 0): floor is -1,  ceil is 0
x = -100: floor = ceil = -1

Something like this? Would it make sense to group the test cases like so and add short comments? Perhaps, combine floor and ceil test in a single call with 3 args: x, floor(x), ceil(x).

What's the thinking behind choosing specific values?

simoneves: I did consider a combined style like `test(input, expected_floor, expected_ceil)` to reduce LOC but that would have required an intermediate lambda (or three) to wrap the two calls to the base class maker function (hopefully very re-usable for other tests, most of which are only for single operators), which I figured overall was uglier.

As for the choice of input and expected output values, I simply used the ones from the existing test. I can add more comments to explain the expected behavior (due to the scale change) in each group?

simoneves: @mbasmanova I expanded the comments for each test group to explain the expected behavior for that input and output scale combo.

kgpai:
@simoneves This change breaks the bias fuzzer . I think you might need to add ArgValueGenerator for ceil . See ExpressionFuzzerTest.cpp

simoneves:
> @simoneves This change breaks the bias fuzzer . I think you might need to add ArgValueGenerator for ceil . See ExpressionFuzzerTest.cpp

@kgpai Sorry, I missed that part. I have pushed a change, although cannot test it locally as I don't have permissions to `wget` the zip file of signatures. If the CI job still fails, I'll need help fetching a copy of that file.

velox/expression/fuzzer/ExpressionFuzzerTest.cpp
```diff
@@ -75,8 +75,9 @@ std::unordered_map<std::string, std::shared_ptr<ArgTypesGenerator>>
         {"minus", std::make_shared<PlusMinusArgTypesGenerator>()},
         {"multiply", std::make_shared<MultiplyArgTypesGenerator>()},
         {"divide", std::make_shared<DivideArgTypesGenerator>()},
-        {"floor", std::make_shared<FloorAndRoundArgTypesGenerator>()},
-        {"round", std::make_shared<FloorAndRoundArgTypesGenerator>()},
+        {"floor", std::make_shared<FloorCeilRoundArgTypesGenerator>()},
+        {"ceil", std::make_shared<FloorCeilRoundArgTypesGenerator>()},
```
simoneves, line 80: Added this line. The rest of the fuzzer changes are due to me renaming the TypesGenerator class.

simoneves: @kgpai I don't know if the Bias Fuzzer is happy now or not, as it didn't seem to run on this last commit. However, the Spark Aggregate Fuzzer failed, which I don't think it did before. I'm not familiar with any of the fuzzer stuff. Is that now due to my "fix", or perhaps an unrelated glitch?

kgpai:
@simoneves Presto bias fuzzer ran : https://github.com/facebookincubator/velox/actions/runs/22366724982/job/64743705125?pr=16253 and succeeded, Presto bias aggregation fuzzer is only for aggregates which ceil is not.

kgpai:
https://github.com/facebookincubator/velox/pull/16502 is landing soon and if you rebase on top that should fix some of the test failures we are seeing. I think the spark aggrgate fuzzer failure is not due to your change.

simoneves:
Updated again from `main` to bring in the Velox-side test fixes. Will hopefully go green now.

karthikeyann:
@kgpai please import this PR

PR #16258, opened 2026-02-05
perf(cudf): Fix several usages of default stream

devavret:
Operations on default stream synchronize operations on other streams and lead to poor utilization of GPU resources. This PR finds and fixes several such usages in cudf execution.


velox/experimental/cudf/connectors/hive/CudfHiveDataSource.h
```diff
@@ -158,6 +159,8 @@ class CudfHiveDataSource : public DataSource, public NvtxHelper {
   std::vector<std::unique_ptr<cudf::scalar>> subfieldScalars_;
   cudf::ast::tree subfieldTree_;
   common::SubfieldFilters subfieldFilters_;
+  // Cached combined subfield filter expression owned by 'subfieldTree_'.
+  cudf::ast::expression const* subfieldFilterExpr_{nullptr};
```
mhaseeb123, line 163: Do we want to use shared_ptr here instead of raw ptr?

velox/experimental/cudf/vector/CudfVector.cpp
```diff
@@ -94,6 +95,17 @@ std::pair<uint64_t, std::unique_ptr<cudf::table>> getTableSize(
       totalBytes, std::make_unique<cudf::table>(std::move(columnsOut)));
 }
 
+void logDefaultStreamIfNeeded(
```
mhaseeb123, line 98: This is a nice util

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.h
devavret: The expression object is actually owned by the `cudf::ast::tree` object so we don't need ownership here. Just a pointer to the actual top level expression out of that tree. I think typically that's tree.back() but I used this just to future proof against that assumption

xiaoxmeng:
@devavret  Can you fix the build failures?

devavret:
@xiaoxmeng I don't see any build failures in this PR. Can you post the link please. I'll take a look if it's caused by this.

PR #16259, opened 2026-02-05
feat(cudf): Use BufferedInput enqueue/load APIs

majetideepak:
The `enqueue/load` APIs handle read coalescing, tracking, and give more flexibility to load all the chunks in
one shot together, instead of reading chunks independently. The load API handles prefetch, preload, and uses the IO executor.
With the following configs in hive catalog
```
cudf.hive.use-buffered-input=true
cudf.hive.use-experimental-reader=true
parquet.reader.chunk-read-limit=4294967296
parquet.reader.pass-read-limit=8294967296
load-quantum=4194304
```
and `connector.num-io-threads-hw-multiplier=2` in config.properties, 
the execution time for all the queries TPCH-SF100 from S3 on g6e reduced from 133 seconds to 75 seconds.

### DIFF
--- a/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
+++ b/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
@@ -254,9 +254,9 @@ std::optional<RowVectorPtr> CudfHiveDataSource::next(
                 cudf::get_current_device_resource_ref());
             // Directly read the column chunk data to the device buffer if
             // supported
-            if (auto cachedInput =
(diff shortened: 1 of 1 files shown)


majetideepak:
@mhaseeb123 thanks for the approval. I am validating the results again from Zoltan's CI runner. Will post results here shortly.

majetideepak:
I verified this in Zoltan's CI as well. Overall time is around 75 seconds with this fix.

karthikeyann:
@xiaoxmeng could this be merged?

PR #16263, opened 2026-02-06
feat(operator): Velox barrier handling for MixedUnion (#16184)

davidgoode:
Summary:

Following along with D71925671, implement barrier handling so we can reuse the operator tree.

Differential Revision: D90407169


xiaoxmeng (approved):
@davidgoode thanks for the iterations!

PR #16268, opened 2026-02-06
feat: Skip iceberg positional delete files by position bounds

joyhaldar:
When reading Iceberg tables with positional deletes, we can skip loading delete files if their positions are before the current split.

The delete file's upperBounds metadata contains the max deleted position for the delete pos column (field ID `2147483545`, defined in [IcebergMetadataColumns.h](https://github.com/facebookincubator/velox/blob/6beae8654f678667cc3410d606a5096b865ba2a3/velox/connectors/hive/iceberg/IcebergMetadataColumns.h#L48)). If this value is less than [splitOffset_](https://github.com/facebookincubator/velox/blob/6beae8654f678667cc3410d606a5096b865ba2a3/velox/connectors/hive/iceberg/IcebergSplitReader.cpp#L74), the delete file can't affect any rows in the current split.

We skip the delete file if:
- `upperBound < splitOffset_` -> all deletes are before this split

Reference: https://iceberg.apache.org/spec/#position-delete-files

Added unit test for upper bound check.


majetideepak:
CC: @yingsu00

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
```diff
@@ -894,6 +894,62 @@ TEST_F(HiveIcebergTest, partitionColumnsFromHive) {
...
+  IcebergDeleteFile deleteFile(
+      FileContent::kPositionalDeletes,
+      deleteFilePath->getPath(),
+      dwio::common::FileFormat::DWRF,
+      3,
+      testing::internal::GetFileSize(
+          std::fopen(deleteFilePath->getPath().c_str(), "r")),
+      {},
+      {},
+      {{posColumn->id, "-1"}});
```
PingLiuPing: Is it possible to simulate a real upper bound in test?

PingLiuPing (commented):
@joyhaldar Thanks for the code. The enhancement looks solid. One comment left in the tests.

velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
joyhaldar: Thank you for reviewing my PR Ping. 

I have changed the upper bound to `2` as the max deleted position. The test now reads from the middle of the file and the split point is after the delete file's bounds.

PingLiuPing (approved):
Thanks @joyhaldar

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -78,6 +79,15 @@ void IcebergSplitReader::prepareSplit(
   for (const auto& deleteFile : deleteFiles) {
     if (deleteFile.content == FileContent::kPositionalDeletes) {
       if (deleteFile.recordCount > 0) {
+        // Skip the delete file if all delete positions are before this split.
+        auto posColumn = IcebergMetadataColumn::icebergDeletePosColumn();
+        if (auto posUpperBoundIt = deleteFile.upperBounds.find(posColumn->id);
+            posUpperBoundIt != deleteFile.upperBounds.end()) {
+          const int64_t deleteUpperBound = std::stoll(posUpperBoundIt->second);
+          if (deleteUpperBound < static_cast<int64_t>(splitOffset_)) {
```
majetideepak: Should we also check if `deleteLowerbound > splitOffset_ + splitLength_`?

joyhaldar: Thank you for your review Deepak. 

Added the lower bound check using `splitRows_` from `baseReader_->numberOfRows()`. Also added a test for it. Thank you for helping me on Slack!

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -72,12 +74,32 @@ void IcebergSplitReader::prepareSplit(
...
+            continue;
+          }
+        }
+        // Skip if all deletes are after this split.
+        if (auto posLowerBoundIt = deleteFile.lowerBounds.find(posColumn->id);
+            posLowerBoundIt != deleteFile.lowerBounds.end()) {
+          const int64_t deleteLowerBound = std::stoll(posLowerBoundIt->second);
+          if (deleteLowerBound >=
```
PingLiuPing: lowerBound is inclusive.
>= to >

joyhaldar: Thank you for your review @PingLiuPing.

My thought process was the following:

If `splitOffset_=0` and `splitRows_=100`, the split will cover rows `0-99`. So `splitOffset_ + splitRows_ = 100` is the first position outside the split. If `lowerBound=100`, all deletes are outside, so we should skip. That's why I used `>=`.

Also [the spec](https://iceberg.apache.org/spec/#position-delete-files) does say,
> A data row is deleted if there is an entry in a position delete file for the row's file and position in the data file, starting at 0.

But please correct me if I am incorrect.

PingLiuPing: > starting at 0.

Thanks, we should use `>=` therotically.

However, I’m struggling to understand how this situation could actually occur. Could you please elaborate?

` if (deleteLowerBound >= static_cast<int64_t>(splitOffset_ + splitRows_))
`
splitRows_ represents the total number of rows in the data file, and splitOffset_ is the starting ordinal offset of this split within the file.
Suppose  splitRows_ = 100, offset = 2.
How can we create a positional delete file with deleted row's position is 102?

joyhaldar: You're right, I think I made a mistake. I used `baseReader_->numberOfRows()` which returns total file rows, not split rows if I understand correctly. I was trying to get rows from the RowGroup, but I may have implemented it incorrectly.

If the lower bound check isn't useful in practice, should I remove it for now? Or would you be able to guide towards the correct way to get the split's row count?

PingLiuPing: I don't see any existing API to get the row numbers from split. Let's remove this for now. Thank you.

joyhaldar: Thank you again for the review. I have removed the lower bound check.

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
joyhaldar: Apologies, I misunderstood your suggestion. I have removed the lower bound check as there's no API to get split row count. Kept only the upper bound check.

velox/connectors/hive/iceberg/PositionalDeleteFileReader.cpp
```diff
@@ -57,9 +57,6 @@ PositionalDeleteFileReader::PositionalDeleteFileReader(
   VELOX_CHECK(deleteFile_.content == FileContent::kPositionalDeletes);
   VELOX_CHECK(deleteFile_.recordCount);
 
-  // TODO: check if the lowerbounds and upperbounds in deleteFile overlap with
```
majetideepak, line 60: Are we not checking only for one of the bounds? Keep the comment for the other bound?

velox/connectors/hive/iceberg/IcebergSplitReader.cpp
```diff
@@ -78,6 +79,15 @@ void IcebergSplitReader::prepareSplit(
   for (const auto& deleteFile : deleteFiles) {
     if (deleteFile.content == FileContent::kPositionalDeletes) {
       if (deleteFile.recordCount > 0) {
+        // Skip the delete file if all delete positions are before this split.
+        auto posColumn = IcebergMetadataColumn::icebergDeletePosColumn();
+        if (auto posUpperBoundIt = deleteFile.upperBounds.find(posColumn->id);
+            posUpperBoundIt != deleteFile.upperBounds.end()) {
+          const int64_t deleteUpperBound = std::stoll(posUpperBoundIt->second);
```
majetideepak: no need of const for scalars.

majetideepak: Convention is to use `folly::to<uint64_t>(posUpperBoundIt->second)`. You can skip the cast below then?

joyhaldar: Thank you Deepak, switched to `folly::to<uint64_t>` and removed the cast below.

velox/connectors/hive/iceberg/PositionalDeleteFileReader.cpp
joyhaldar: Thank you Deepak, I have updated the comment to keep the `TODO` for the lower bound check [here](https://github.com/joyhaldar/velox/blob/7052c25abfa9d0faf9c59d89934464d13c7b747d/velox/connectors/hive/iceberg/IcebergSplitReader.cpp#L83).

PR #16270, opened 2026-02-06
feat(expr): Add error code filtering to TryExpr

pradeepvaka:
Summary:
Add support for filtering catchable error codes in TryExpr. This allows
TRY() to catch only specific error codes and rethrow others.

**Changes:**
- Add `kTryCatchableErrorCodes` config in Velox QueryConfig
- Modify TryExpr to filter errors based on configured catchable codes:
  - Empty config = catch all user errors (backward compatible)
  - Specific codes = only catch matching errors, rethrow others
- Enable captureErrorDetails only when filtering is needed (performance)

**Design decisions:**
- Error codes matched by exact string (e.g., "ARITHMETIC_ERROR")
- Non-user errors are never caught regardless of configuration
- Non-Velox exceptions are treated as catchable

Related change in Presto Java https://github.com/prestodb/presto/pull/26976/

Differential Revision: D92117739

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -843,6 +843,12 @@ class QueryConfig {
   static constexpr const char* kJoinBuildVectorHasherMaxNumDistinct =
       "join_build_vector_hasher_max_num_distinct";
 
+  /// Comma-separated list of additional error codes that TRY() should catch.
+  /// TRY() always catches user errors. When specified, also catches errors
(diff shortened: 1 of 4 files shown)


kgpai:
Thanks @pradeepvaka Can you create an issue regarding this as its a pretty fundamental change that changes very core functionality using a query config. Please add motivation for this change and alternatives considered, risk etc. 
Thank you!

Issue #16087, opened 2026-01-21
Improve CastHooks and CastPolicies performance/usability


kevinwilfong:
### Description

CastHooks allow engines to configure how the CastExpr handles converting between built in types provided by Velox. When the interface was introduced it provided hooks for casting between 3 pairs of types, as well as for trimming strings, and a couple boolean flags. This has proven to be very useful, it now provides hooks for 8 pairs of types (if we lump the classes of integer and floating point types into two classes) and exposes new classes like TimestampToStringOptions and notably PolicyType. Other than complex and custom types, there's very few cast operations that don't rely on CastHooks.

CastHooks is an interface, so the calls we make to it are virtual, which leads to less than optimal performance. This has been handled in 2 ways:
1) Ensuring the call is made once per batch, a single virtual call per batch does not introduce much overhead. Unfortunately, this has not been maintained and we make virtual calls to CastHooks per row in several places now. E.g. https://github.com/facebookincubator/velox/blob/7b10ae2ec40e648b33258443499d04320972ba9b/velox/expression/CastExpr-inl.h#L138-L139
2) Adding CastPolicies to velox/type/Conversions.h This has been done for Presto and Spark, and leveraging templates/switch statements allows us to make a virtual call once per batch and change the per row behavior at compile time. Unfortunately, this is not scalable, we cannot add separate CastPolicies for every engine that use Velox in core Velox code.
https://github.com/facebookincubator/velox/blob/7b10ae2ec40e648b33258443499d04320972ba9b/velox/type/Conversions.h#L36-L61

The primary challenges with CastHooks are that new functions are added adhoc as needed by engines and the structure of applyCastKernel in CastExpr. The latter makes it difficult to change the behavior per engine without either per-row virtual calls or hard coded CastPolicies, and the former has led to us doing both.

I'd like to propose a more fully fleshed out interface that gives engines the ability to customize cast according to their individual quirks, share implementations where they overlap, and make it less tempting to introduce per row virtual calls.

CastKernel (from the applyCastKernel function in CastExpr) provides 3 sets of functions:
* For logical types (Date, IntervalDayTIme, Time, and Decimal) castTo<X> and castFrom<X> functions
* For physical types (Boolean, Integer types, Real, Double, Varbinary, Varchar, Timestamp) castTo<X> functions
* A couple of the helper functions in CastHooks applyTryCastRecursively and timestampToStringOptions because they're used in casting complex types, and in casting custom types respectively.

[...]

PR #16274, opened 2026-02-06
refactor: CastKernel [2/11] Port cast to/from Date to CastKernel

kevinwilfong:
Summary:
This change is part of a stack to address the performance and usability issues 
with CastExpr/CastHooks/CastPolicy and different engine semantics. Please 
see https://github.com/facebookincubator/velox/issues/16087 for a fuller 
description of the motivation. The stack will be landed as a unit, I've broken it 
up to try to make it easier to review.

This change introduces the castToDate and castFromDate functions to 
CastKernel and implements them in PrestoCastKernel and SparkCastKernel.

I have removed what I can of the relevant code in CastExpr and the CastHooks 
to help with seeing mapping to where the new code came from, it was just 
moved and slightly refactored. Note that I couldn't remove the 
removeWhiteSpaces functions from the Hooks, but it is copied from there. 
They will deleted from the Hooks once it's possible later in the stack.

Differential Revision: D91834120

### DIFF
--- a/velox/expression/CMakeLists.txt
+++ b/velox/expression/CMakeLists.txt
@@ -52,6 +52,7 @@ velox_add_library(
   LambdaExpr.cpp
   PeeledEncoding.cpp
   PrestoCastHooks.cpp
+  PrestoCastKernel.cpp
   RegisterSpecialForm.cpp
   RowConstructor.cpp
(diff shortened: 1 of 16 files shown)


kgpai (approved):
Please rebase against latest main.

Issue #16286, opened 2026-02-06
Build break in cudf code - undefined symbol serialized


czentgr:
### Problem description

Adapters CI starts failing with latest commits.

E.g. https://github.com/facebookincubator/velox/actions/runs/21765814012/job/62801260775?pr=16113

The issue is that an overriden function was declared but not implemented.
The PR that declared (but not defined) `folly::dynamic serialize() const override;` was https://github.com/facebookincubator/velox/pull/14968.

### System information

n/a

### CMake log

```bash
/opt/rh/gcc-toolset-14/root/usr/libexec/gcc/x86_64-redhat-linux/14/ld: velox/experimental/cudf/connectors/hive/libvelox_cudf_hive_connector.a(CudfHiveConnectorSplit.cpp.o):(.data.rel.ro._ZTVN8facebook5velox10cudf_velox9connector4hive22CudfHiveConnectorSplitE[_ZTVN8facebook5velox10cudf_velox9connector4hive22CudfHiveConnectorSplitE]+0x10): undefined reference to `facebook::velox::cudf_velox::connector::hive::CudfHiveConnectorSplit::serialize() const'
collect2: error: ld returned 1 exit status
```

bdice:
I think this is resolved by #16287 - please reopen if more is needed.

PR #14968, opened 2025-09-26
feaf(cudf): Add `start`, `length` and `infoColumns` fields to `CudfHiveConnectorSplit`

mhaseeb123:
## Description

Closes https://github.com/rapidsai/velox/issues/16 (duplicate of https://github.com/facebookincubator/velox/pull/14129)

This PR adds `start`, `length` and `infoColumns` fields to the `CudfHiveConnectorSplit` to allow splitting parquet file read across multiple splits. These fields also allow direct conversion from `HiveConnectorSplit` to `CudfHiveConnectorSplit`.

## Checklist
- [ ] Update RAPIDS/cuDF pinning to the latest commit or `25.10` release to pull in backend changes from https://github.com/rapidsai/cudf/pull/19991

PR #16287, opened 2026-02-07
fix(build): Define serialize function in CudfHiveConnectorSplit

czentgr:
Addresses: https://github.com/facebookincubator/velox/issues/16286


mhaseeb123:
Thanks for adding this. Looks like I think I forgot to copy over the definition from an earlier PR to PR https://github.com/facebookincubator/velox/pull/14968

czentgr:
@mhaseeb123 lets see if we get a clean build (and only tests failing with known issues) before adding ready-to-merge. I had a typo in the code that caused another failure.

mhaseeb123:
@czentgr thanks for the update. Please let me know if anything else broke because of #14968 (it had been up for a while before being merged so it could likely have been outdated)

karthikeyann (approved):

czentgr:
The pyvelox build error is unrelated. The test failures are known. This fix is good to go.

czentgr:
@kgpai FYI

PR #16288, opened 2026-02-07
build: Reduce run checks line limit

peterenescu:
Summary: Reduce the run checks line limit to 80 to be consistent with internal default linter

Differential Revision: D92549522


peterenescu:
@assignUser any concerns with this change?

peterenescu:
@assignUser done! Please take a look

karthikeyann:
+1 for 100 column limit.

Issue #16199, opened 2026-02-02
[cuDF] Companion functions not supported anymore in aggregations


devavret:
### Bug description

Hash aggregation supports adding and using a companion aggregation function which is generally the function + overriding the step. Meaning a count_merge acts like a count in intermediate step even if it appears in a `HashAggregation` operator with a different step. Since merging #15529 , this gets rejected for GPU support and falls back to CPU execution. There is also a bug in the AggregationTest for cudf where it allows cpu fallback, silently failing to use GPU and hence not testing all functionality properly. This should use be disallowed in unit tests. https://github.com/facebookincubator/velox/pull/16050#issuecomment-3835272738

### System information

Irrelevant for this issue

### Relevant logs

```bash
[ RUN      ] AggregationTest.CompanionAggs
E20260202 13:46:05.321277 2053115 Exceptions.h:53] Line: /velox/velox/experimental/cudf/exec/ToCudf.cpp:424, Function:compile, Expression: condition Replacement with cuDF operator failed, Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
C++ exception with description "Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Replacement with cuDF operator failed
Retriable: False
Expression: condition
Function: compile
/velox/velox/experimental/cudf/exec/ToCudf.cpp
Line: 424
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorEPKcEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox10cudf_velox12CompileState7compileEb
# 4  _ZNSt17_Function_handlerIFbRKN8facebook5velox4exec13DriverFactoryERNS2_6DriverEENS1_10cudf_velox17CudfDriverAdapterEE9_M_invokeERKSt9_Any_dataS5_S7_
[...]

PR #16289, opened 2026-02-07
fix(cudf): Allow companion aggregations in step-aware validation

mattgara:
This PR fixes companion aggregate function registration  in step-aware validation. 

This PR addresses the issue raised in #16199 .


devavret:
@jinchengchenghh Can you please review. Companion aggregations should've been broken without this change.

jinchengchenghh (approved):
Thanks for your fix!

PR #16291, opened 2026-02-07
feat(python): Hook support in LocalDebuggerRunner

pedroerp:
Summary:
Adding `set_hook(plan_node_id, hook_func)` support in
LocalDebuggerRunner to enable users debugging data issues in the DAG to add a
callback that inspects the vector, then decides whether the control loop should
stop and produce the vector as output, or keep going. Example:

```
def hook(vector):
    # inspect vector
    return True # or False, to ignore and keep advancing.


runner = LocalDebuggerRunner(plan_node)
runner.set_hook(some_node_id, hook)
runner.execute()
...
```

Reviewed By: xiaoxmeng

Differential Revision: D92328798

### DIFF
--- a/python/test/test_debugger_runner.py
+++ b/python/test/test_debugger_runner.py
@@ -169,3 +169,88 @@ class TestPyVeloxDebuggerRunner(unittest.TestCase):
 
         it.step()
         self.assertEqual(it.at(), self._node_ids[3])
+
+    def test_hook_always_stop(self):
+        """Hook that always returns True should behave like set_breakpoint."""
+        runner = LocalDebuggerRunner(self._plan_node)
+        runner.set_hook(self._node_ids[3], lambda v: True)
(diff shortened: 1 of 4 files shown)


python/test/test_debugger_runner.py
```diff
@@ -169,3 +169,89 @@ def test_iterator_at(self):
...
+        """Hook that conditionally stops based on vector content."""
+        runner = LocalDebuggerRunner(self._plan_node)
+
+        # Track how many times the hook is called.
+        call_count = [0]
+
+        def conditional_hook(vector):
+            call_count[0] += 1
+            # Stop only on odd calls.
+            return call_count[0] % 2 == 1
```
duanmeng: Perhaps, we can use a more Pythonic way,
```Python
call_count = 0
def conditional_hook(vector):
    nonlocal call_count
    call_count += 1
    # Stop only on odd calls.
    return call_count % 2 == 1
```

PR #16292, opened 2026-02-07
fix: Rename Type::isUnKnown() to isUnknown()

mbasmanova:
Summary: Rename the `isUnKnown()` method to `isUnknown()` to fix the typo (capital 'K' should be lowercase 'k'). This brings the method naming in line with other similar methods like `isUnscaledLongDecimal()`, `isOpaque()`, etc.

Differential Revision: D92610008


natashasehgal (approved):
Thanks for fixing this

Issue #16298, opened 2026-02-08
fix(cudf) Revert CudfHiveConnectorSplit change as it breaks Presto


simoneves:
### Bug description

PR #14968 unfortunately breaks Velox GPU in the context of Presto, possibly because the split start value now parsed from Presto is bogus (TBD).

Parquet file reading is therefore broken, which breaks all TPCH tests etc.

That PR commit should be reverted pending investigation.

### System information

N/A

### Relevant logs

```bash

```

PR #16299, opened 2026-02-08
fix(cudf): Revert "feat(cudf): Add `start`, `length` and `infoColumns` fields to `CudfHiveConnectorSplit` (#14968)"

simoneves:
This reverts commit 4e09b55218b2e452c31d160712032f90c372d0e3.

Per issue #16298 

Rebased and extended to first revert #16287 

I don't seem to have permissions to edit the title of this PR

_UPDATE_

To be replaced by https://github.com/facebookincubator/velox/pull/16312

### DIFF
--- a/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
+++ b/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
@@ -451,6 +451,10 @@ void CudfHiveDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
           hiveSplit->fileFormat,
           dwio::common::FileFormat::PARQUET,
           "Unsupported file format for conversion from HiveConnectorSplit to CudfHiveConnectorSplit");
+      VELOX_CHECK_EQ(
+          hiveSplit->start,
+          0,
+          "CudfHiveDataSource cannot process splits with non-zero offset");
       // Remove "file:" prefix from the file path if present
       std::string cleanedPath = hiveSplit->filePath;
       constexpr std::string_view kFilePrefix = "file:";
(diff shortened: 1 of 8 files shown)


karthikeyann:
@simoneves build break is fixed by PR https://github.com/facebookincubator/velox/pull/16287

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
```diff
@@ -573,17 +572,17 @@ void CudfHiveDataSource::setupCudfDataSourceAndOptions() {
...
           .timestamp_type(cudfHiveConfig_->timestampType())
           .build();
 
-  // Set num_bytes only if available
-  if (split_->size() != std::numeric_limits<uint64_t>::max()) {
-    readerOptions_.set_num_bytes(split_->size());
+  // Set num_rows only if available
+  if (cudfHiveConfig_->numRows().has_value()) {
+    readerOptions_.set_num_rows(cudfHiveConfig_->numRows().value());
   }
```
mhaseeb123, line 586: Hi @karthikeyann, @simoneves mentioned that he noticed empty output being produced by the reader with these changes as the `split_->start` and `split_->length` seem a bit suspicious or maybe just are off by 1. Since this hasn't merged yet, I would recommend not reverting the whole PR and instead just comment out two lines where we actually use these fields until we have answers to why the input values seem incorrect (L575 where we set skip_bytes and L586 where we set num_bytes). 

@simoneves can you please confirm if just doing this would solve the problem you are seeing?

simoneves:
Proposing replacing this with #16312

velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
simoneves: Implemented in #16312 and testing now

simoneves:
Reverted to DRAFT as the intention is to replace this with #16312

simoneves:
Replaced by #16335

PR #16300, opened 2026-02-08
fix: Fix flaky SkewedPartitionRebalancerTest.serializedRebalanceExecution

tanjialiang:
Summary:
The test was flaky because background thread might have already finished rebalancing before main thread starts. This results in 2 rebalancing happening.

The fix adds an additional EventCount to make sure when the main thread starts to rebalance, the background thread's rebalancing hasn't finished.

Reviewed By: xiaoxmeng

Differential Revision: D92632923

### DIFF
--- a/velox/common/base/tests/SkewedPartitionBalancerTest.cpp
+++ b/velox/common/base/tests/SkewedPartitionBalancerTest.cpp
@@ -330,10 +330,18 @@ DEBUG_ONLY_TEST_F(SkewedPartitionRebalancerTest, serializedRebalanceExecution) {
   SkewedPartitionRebalancerTestHelper helper(balancer.get());
   folly::EventCount rebalanceWait;
   std::atomic_bool rebalanceWaitFlag{true};
+
+  // Used to signal that the background thread has entered the test value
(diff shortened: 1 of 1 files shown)


velox/common/base/tests/SkewedPartitionBalancerTest.cpp
```diff
@@ -344,6 +352,12 @@ DEBUG_ONLY_TEST_F(SkewedPartitionRebalancerTest, serializedRebalanceExecution) {
 
   std::thread rebalanceThread([&]() { balancer->rebalance(); });
 
+  // Wait for the background thread to enter the test value callback and block
+  // there. This ensures that when we call rebalance() from the main thread,
+  // rebalancing_ is already true and our rebalance() call will return early.
+  mainThreadRebalancerWait.await(
+      [&] { return mainThreadRebalancerWaitFlag.load(); });
```
zhztheplayer, line 359: The condition may need to be `!mainThreadRebalancerWaitFlag.load()` to match the notifying statement `mainThreadRebalancerWaitFlag = false`.

PR #16302, opened 2026-02-08
feat: Support decimal type for the Spark checked_add and checked_subtract functions

### LINKED ISSUE #16301: Add checked_add and checked_subtract for Spark decimal types
### Description

Part of #3869

  ## Description

  PR #10083 added `checked_add`, `checked_subtract`, `checked_multiply`, and
  `checked_divide` for integer and floating-point types to support Spark's ANSI mode.
  However, other arithmetic "checked" operators with decimal type were not included.

  In Spark's ANSI mode (`spark.sql.ansi.enabled=true`), decimal arithmetic overflow
  should raise an error rather than returning null. Decimal addition and subtraction
  go through `CheckOverflow(BinaryArithmetic, decimalType, nullOnOverflow=false)`,
  which requires checked variants that throw on overflow.

  Currently, the existing `add`/`subtract` for decimal types return null on overflow
  (matching Spark's default non-ANSI behavior). We need decimal `checked_add`/
  `checked_subtract` that return an error status on overflow instead. My implementation should be aligned to what `checked_divide` has done.

  ## Scope

  - `checked_add` for all decimal input combinations (short/long x short/long)
  - `checked_subtract` for all decimal input combinations (short/long x short/long)
  - Both `allowPrecisionLoss` and `denyPrecisionLoss` variants

n0r0shi:
## Summary

  - Add `checked_add` and `checked_subtract` for decimal types that throw on
    overflow instead of returning null, enabling Spark ANSI mode support for
    decimal arithmetic

  ## Test plan

  - [x] Normal cases: all 4 input type combinations (short+short, short+long,
        long+short, long+long)
  - [x] Edge cases: add/subtract zero, negative numbers
  - [x] Precision capped at 38, no overflow (small values and near-boundary)
  - [x] Positive overflow throws error
  - [x] Negative overflow throws error
  - [x] Boundary overflow (max +/- 1) throws error

Closes #16301

### DIFF
--- a/velox/functions/sparksql/DecimalArithmetic.cpp
+++ b/velox/functions/sparksql/DecimalArithmetic.cpp
@@ -181,7 +181,14 @@ struct DecimalAddSubtractBase {
 
     // Scale up the whole part and scale down the fraction part to combine them.
     fraction = reduceScale(TResult(fraction), higherScale - rScale);
-    const auto whole = TResult(aWhole) + TResult(bWhole) + TResult(carryToLeft);
+    // Use __builtin_add_overflow to avoid undefined behavior from signed
+    // integer overflow when both whole parts are large.
+    TResult whole;
+    if (__builtin_add_overflow(TResult(aWhole), TResult(bWhole), &whole) ||
+        __builtin_add_overflow(whole, TResult(carryToLeft), &whole)) {
+      overflow = true;
+      return 0;
+    }
     return decimalAddResult(whole, TResult(fraction), rScale, overflow);
   }
 


velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -781,6 +841,11 @@ void registerDecimalAdd(const std::string& prefix) {
   registerDecimalBinary<AddFunctionDenyPrecisionLoss>(
       prefix + "add" + kDenyPrecisionLoss,
       makeConstraints(rPrecision, rScale, false));
+  registerDecimalBinary<CheckedAddFunctionAllowPrecisionLoss>(
+      prefix + "checked_add", makeConstraints(rPrecision, rScale, true));
+  registerDecimalBinary<CheckedAddFunctionDenyPrecisionLoss>(
+      prefix + "checked_add" + kDenyPrecisionLoss,
```
rui-mo, line 854: Is the `allowPrecisionLoss` in Spark taking effect for the checked arithmetic functions (on decimal type) as well, could you please confirm?

https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L444-L452

rui-mo (commented):
Thanks for this work! Please also update the relevant documentation.

https://github.com/facebookincubator/velox/blob/main/velox/docs/functions/spark/decimal.rst#arithmetic-functions

velox/functions/sparksql/DecimalArithmetic.cpp
n0r0shi: Yes. `allowPrecisionLoss` (`spark.sql.decimalOperations.allowPrecisionLoss`) and ANSI mode (`spark.sql.ansi.enabled`) are orthogonal settings in Spark.

  Regardless of the ANSI setting, the result type of decimal arithmetic is determined by `allowPrecisionLoss`, which controls whether `adjustPrecisionScale` or `bounded` is used in
  `resultDecimalType`. Then, if overflow occurs for that result type, Spark returns NULL when ANSI is off, or throws when ANSI is on.

  Specifically, the `precision` and `scale` used in [`checkDecimalOverflow`](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L445-L446) come from pattern-matching against [`dataType`](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L243), which is a lazy val that calls [`resultDecimalType`](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L224). That is where `allowPrecisionLoss` determines the result precision and scale. So the overflow check is always performed against the type that `allowPrecisionLoss` has already shaped.

  That is why we register both variants here.

n0r0shi:
Updated the doc. Unchecked add and subtract were missing, so I added them as well. Please let me know if there's anything else needed for this PR.

Thanks! @rui-mo

velox/functions/sparksql/DecimalArithmetic.cpp
rui-mo: Thanks for providing these details. Is the concept you described the same as the existing decimal_div function, see https://github.com/facebookincubator/velox/pull/16323#pullrequestreview-3778613526? I’m a bit confused about why another decimal_divide is needed.

rui-mo: > the result type of decimal arithmetic is determined by allowPrecisionLoss, which controls whether adjustPrecisionScale or bounded is used in
resultDecimalType. Then, if overflow occurs for that result type, Spark returns NULL when ANSI is off, or throws when ANSI is on.

Yes, I understand that `allowPrecisionLoss` takes effect for decimal add/subtract etc., but I haven't confirmed if it also applied to check checked arithmetic functions. The [checkDecimalOverflow](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L445-L446) you mentioned is also from a Add function.

n0r0shi: @rui-mo For your 1st question — Spark has two separate division operators:
  - [`Divide`](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L783):  for `/`
  - [`IntegralDivide`](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L854C12-L854C26) for the SQL `DIV` expression

  Spark's `Divide` maps to Velox's `divide`, and `IntegralDivide` maps to Velox's `div`. Their ANSI-mode counterparts are `checked_divide` and `checked_div`, respectively.

  These non-decimal versions already existed before my change [[link](https://github.com/facebookincubator/velox/blame/main/velox/functions/sparksql/registration/RegisterMath.cpp#L133-L135 )]

Regardless of whether this naming convention is ideal, I don't intend to change them. My PR #16323 simply adds the decimal version of `checked_divide`, since all arithmetic operators need a dedicated decimal implementation. `checked_div` already has a Decimal counterpart.

n0r0shi: @rui-mo For `allowPrecisionLoss` — it applies to checked arithmetic functions as well. In Spark, there are no separate "checked" expression classes. The same expression handles both ANSI and
  non-ANSI behavior, controlled by the [`EvalMode` flag](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L854-L857).

  Then Gluten reads this value and maps it to either the checked or non-checked version of the Velox operator [[link](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L154)].

  

  This is an existing pattern and I am following it. Since `allowPrecisionLoss` shapes the result type at the Spark level before Gluten does this mapping, it applies identically to both
  paths. That is why we register both `allowPrecisionLoss` and `denyPrecisionLoss` variants for the checked functions. `IntegralDivide` always returns `LongType` (result scale is 0), so
  precision loss is not applicable for the existing `checked_div`.

rui-mo: Thanks! The information you provided really helped clarify the relationship between checked arithmetic functions and ordinary arithmetic functions, as well as the relationship between `checked_div` and `checked_divide`. Would it be possible to document this for future reference?

> Spark's Divide maps to Velox's divide, and IntegralDivide maps to Velox's div. Their ANSI-mode counterparts are checked_divide and checked_div, respectively.

Perhaps add above to the PR description of https://github.com/facebookincubator/velox/pull/16323 to prevent confusion.

> For allowPrecisionLoss — it applies to checked arithmetic functions as well. In Spark, there are no separate "checked" expression classes. The same expression handles both ANSI and
non-ANSI behavior, controlled by the [EvalMode flag](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L854-L857).

Perhaps we could add documentation around line 6 to provide more context for `allowPrecisionLoss`. This would help clarify how checked arithmetic functions enable `allowPrecisionLoss` in Spark.

> IntegralDivide always returns LongType (result scale is 0), so
precision loss is not applicable for the existing checked_div

I’d recommend adding the above to the documentation around line 139 for checked_div.

velox/docs/functions/spark/decimal.rst
```diff
@@ -143,6 +143,40 @@ Arithmetic Functions
     Division by zero or overflow results in an error.
     Corresponds to Spark's operator ``div`` with ``spark.sql.ansi.enabled`` set to true.
 
+.. spark:function:: add(x: decimal(p1, s1), y: decimal(p2, s2)) -> r: decimal(p3, s3)
```
rui-mo, line 135: nit: please follow the alphabetical order to update the documentation.

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
```diff
@@ -88,6 +88,25 @@ class DecimalArithmeticTest : public SparkFunctionBaseTest {
       std::optional<U> u) {
     return evaluateOnce<int64_t>("checked_div(c0, c1)", {tType, uType}, t, u);
   }
+
+  template <typename T, typename U>
+  std::optional<int128_t> checked_add(
```
rui-mo: nit: Velox adopts the camel-case name style. Would you please fix them? Thanks.

velox/functions/sparksql/DecimalArithmetic.cpp
```diff
@@ -790,6 +855,11 @@ void registerDecimalSubtract(const std::string& prefix) {
   registerDecimalBinary<SubtractFunctionDenyPrecisionLoss>(
       prefix + "subtract" + kDenyPrecisionLoss,
       makeConstraints(rPrecision, rScale, false));
+  registerDecimalBinary<CheckedSubtractFunctionAllowPrecisionLoss>(
+      prefix + "checked_subtract", makeConstraints(rPrecision, rScale, true));
+  registerDecimalBinary<CheckedSubtractFunctionDenyPrecisionLoss>(
+      prefix + "checked_subtract" + kDenyPrecisionLoss,
+      makeConstraints(rPrecision, rScale, false));
```
rui-mo, line 869: Do we have tests to cover both allow-precision-loss and deny-precision-loss?

rui-mo (commented):
Thanks. I added some nits and this change LGTM overall.

velox/functions/sparksql/DecimalArithmetic.cpp
n0r0shi: Thanks.
  - Updated the PR description of https://github.com/facebookincubator/velox/pull/16323.
  - The remaining items are addressed in b3a57e021c518f7d3e2e5ef2f2be104b338984c4.

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
n0r0shi: Done in b3a57e021c518f7d3e2e5ef2f2be104b338984c4

velox/functions/sparksql/DecimalArithmetic.cpp
n0r0shi: Thank you for catching this! The original tests only covered the default (allow-precision-loss) variant and I missed the "deny" variant. Added in b3a57e021c518f7d3e2e5ef2f2be104b338984c4.

velox/docs/functions/spark/decimal.rst
n0r0shi: Thanks. Addressed in a263e60d4bd1471d6d80b418d61895951988f389

n0r0shi:
@rui-mo I believe I've addressed all the review comments. We'll proceed with the `checked_*` approach for `add` and `subtract`, as aligned in the discussion on #16361. Please let me know if there's anything else that needs to be updated. Thanks!

velox/docs/functions/spark/decimal.rst
```diff
@@ -8,6 +8,12 @@ The result may exceed maximum allowed precision of 38.
 
 Second stage caps precision at 38 and either reduces the scale or not depending on allow-precision-loss flag.
 
+The allow-precision-loss flag applies to both regular and checked (ANSI mode) arithmetic functions.
+In Spark, there are no separate checked expression classes. The same expression (e.g., ``Add``)
+handles both ANSI and non-ANSI behavior, controlled by an ``EvalMode`` flag. In Velox, the checked
+variants are registered as separate functions (e.g., ``checked_add``, ``checked_subtract``).
```
rui-mo: nit: perhaps mention that they are registered as separate functions to support the TRY evaluation mode.

rui-mo:
To fix the below issue, the arg types generator needs to be extended for the checked decimal functions like `velox/functions/sparksql/fuzzer/AddSubtractArgTypesGenerator.h`.

```
Reason: Cannot generate argument types for checked_add with return type DECIMAL(9, 0).
Retriable: False
Expression: it != argTypesGenerators_.end()
Function: generateExpressionFromSignatureTemplate
File: /__w/velox/velox/velox/velox/expression/fuzzer/ExpressionFuzzer.cpp
Line: 1055
```

rui-mo:
Please also fix the format issues, thanks.

velox/docs/functions/spark/decimal.rst
n0r0shi: addressed in 8dc879f74923262fad989a3dadad104be8d9383b

rui-mo: @n0r0shi Thanks. There are some workflow failures:
https://github.com/facebookincubator/velox/actions/runs/22347057553/job/64672777202?pr=16302 and format issues. Would you please take a look?

n0r0shi: I just added a fix in 28d89e9cde1614200cff74a31eefcc58dffd6576. I will check the next run, thanks.

n0r0shi:
Fixed the format issue and `AddSubtractArgTypesGenerator` in 28d89e9cde1614200cff74a31eefcc58dffd6576. Thanks!

n0r0shi:
There are 3 failures at this point, as far as I checked
- [Fuzzer Jobs / Spark Aggregate Fuzzer (pull_request)](https://github.com/facebookincubator/velox/actions/runs/22457771866/job/65062075498?pr=16302)
```
20260226 14:12:32.165205 19022 File.cpp:254] preadv failed with error: Bad file descriptor
```
this can be related to https://github.com/facebookincubator/velox/issues/16543

- [Linux Build using GCC / Build with GCC / Linux release with adapters (pull_request)](https://github.com/facebookincubator/velox/actions/runs/22457771934/job/65050526188?pr=16302)
- [Linux Build using GCC / Build with GCC / Ubuntu debug with resolve_dependency (pull_req..](https://github.com/facebookincubator/velox/actions/runs/22457771934/job/65050470988?pr=16302)
```
[  FAILED  ] TaskTest.addSplitAfterBarrier
```

which looks relevant to https://github.com/facebookincubator/velox/issues/16510

So it doesn't look like my PR has introduced these failures. Please let me know otherwise @rui-mo

rui-mo:
@n0r0shi Can you please rebase to latest main? The Linux test failures should be fixed, and the Aggregate fuzzer issue is going to be fixed in https://github.com/facebookincubator/velox/pull/16548.

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
```diff
@@ -88,6 +88,25 @@ class DecimalArithmeticTest : public SparkFunctionBaseTest {
...
+  }
+
+  template <typename T, typename U>
+  std::optional<int128_t> checkedSubtract(
+      const TypePtr& tType,
+      const TypePtr& uType,
+      std::optional<T> t,
+      std::optional<U> u) {
+    return evaluateOnce<int128_t>(
+        "checked_subtract(c0, c1)", {tType, uType}, t, u);
```
rui-mo: Please also review this comment: https://github.com/facebookincubator/velox/pull/16307#discussion_r2862388875, and consider adding try_* tests.

n0r0shi: Added `try_*` tests in 69afcb2. Also consolidated the default and `denyPrecisionLoss` cases — they now share a common helper that tests the same core scenarios, while `denyPrecisionLoss` retains additional test cases for its specific behavior.

zhli1142015 (approved):

rui-mo (approved):
Thanks!

n0r0shi:
Some checks failed but they don't appear related to my change — do I need to take any action?

kKPulla:
> Some checks failed but they don't appear related to my change — do I need to take any action?

Yeah the failures are for the newly added tests failing in debug mode

```
Note: Google Test filter = DecimalArithmeticTest.checkedSubtractDenyPrecisionLoss
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DecimalArithmeticTest
[ RUN      ] DecimalArithmeticTest.checkedSubtractDenyPrecisionLoss

fbcode/velox/functions/sparksql/DecimalArithmetic.cpp:184:40: runtime error: signed integer overflow: 0x4b3b4ca85a86c47a098a223fffffffff + 0x4b3b4ca85a86c47a098a223fffffffff cannot be represented in type '__int128'
    #0 0x7fa39b4562bb in __int128 facebook::velox::functions::sparksql::(anonymous namespace)::DecimalAddSubtractBase::addLargeNonNegative<__int128, __int128, __int128>(__int128, __int128, unsigned char, unsigned char, unsigned char, bool&) fbcode/velox/functions/sparksql/DecimalArithmetic.cpp:184

...

SUMMARY: UndefinedBehaviorSanitizer: signed-integer-overflow fbcode/velox/functions/sparksql/DecimalArithmetic.cpp:184:40 
```

We also have another internal CI that detects build speed regression for sparksql/ directory that failed. @kgpai any thoughts for that?

n0r0shi:
It looks like the addition of whole parts at this line is undefined behavior when the sum overflows `int128_t`, and my test cases are the first to hit that path with values large enough to trigger it:
```
  const auto whole = TResult(aWhole) + TResult(bWhole) + TResult(carryToLeft);
```

  I can wrap this with `__builtin_add_overflow` to detect the overflow safely:

```
  TResult whole;
  if (__builtin_add_overflow(TResult(aWhole), TResult(bWhole), &whole) ||
      __builtin_add_overflow(whole, TResult(carryToLeft), &whole)) {
    overflow = true;
    return 0;
  }
```
  Does that look right to you? @kKPulla

n0r0shi:
@kKPulla @kgpai  Pushed the change above. Let me know if there's anything I should check regarding the build speed regression.

kKPulla:
> @kKPulla @kgpai Pushed the change above. Let me know if there's anything I should check regarding the build speed regression.

Thanks for the update. I am reimporting to run the CI again to confirm, will keep you posted.

kKPulla:
@n0r0shi @rui-mo The test failures seem to have been fixed but the build speed regression is still there. Looking further into the change, it is very likely because of the templatizing we are doing in this PR. If you have any ways to optimize build speed for sparksql functions, would appreciate if you can look into it as we are trying to reduce build speeds generally in Velox. cc: @kgpai

n0r0shi:
Thanks for looking into this, @kKPulla @kgpai.

The regression comes from 20 new `SimpleFunctionAdapter` instantiations (4 `registerDecimalBinary` calls × 5 type combinations each), following the exact same pattern as the existing non-checked add/subtract/multiply/divide in this file. I looked into reducing the cost but the only options would sacrifice runtime performance or change the registration framework architecture.
 
These checked functions are necessary for Spark ANSI and TRY mode support. Build time improvements are worth pursuing separately at the framework level, but I don't think individual feature work should be blocked by costs inherent to the existing registration pattern.

Let me know how you'd like to proceed.

kgpai:
@kKPulla Lets go ahead and merge this and pursue build speed improvements as a separate overall topic.

PR #16308, opened 2026-02-09
fix: Use XORShiftRandom in Spark rand function for compatibility

yaooqinn:
## Summary
Migrate Spark's `rand()` function from `std::mt19937` to `XORShiftRandom` to 
match Spark's actual random number generation algorithm.

## Background
This is a follow-up to #16014 as @zhli1142015 suggested.
The seeded `rand(seed)` now uses the same XORShift algorithm that Spark uses, 
enabling result consistency between Velox and Spark when a seed is provided.

## Changes
- Add `nextDouble()` to `XORShiftRandom` matching Java's `Random.nextDouble()`
(53 bits of randomness)
- Update `RandFunction` to use `XORShiftRandom` instead of `std::mt19937`
- Add separate `initialize()` for unseeded `rand()` with random seed from
`folly::Random`
- Add tests for `nextDouble()` range and determinism

## Breaking Change
⚠️ This is a **behavioral change** - seeded `rand(seed)` will produce different results
than before, but now matches Spark's output.

## Test Plan
All existing tests pass and new tests are added.
- `XORShiftRandomTest.nextDoubleRange` - validates [0.0, 1.0) range
- `XORShiftRandomTest.nextDoubleDeterminism` - same seed produces same sequence
- `XORShiftRandomTest.nextDoubleDifferentSeeds` - different seeds produce different
sequences


velox/functions/lib/XORShiftRandom.h
```diff
@@ -44,6 +44,14 @@ class XORShiftRandom {
         (static_cast<uint32_t>(next(31)) * static_cast<uint64_t>(bound)) >> 31);
   }
 
+  /// Returns a random double in [0.0, 1.0) matching Java's Random.nextDouble().
```
rui-mo, line 47: Would you please add a link to Java's implementation to improve the maintainability?

yaooqinn: Addressed, thank you very much for the careful look

yaooqinn:
Thank you @rui-mo for the help.

PR #16304, opened 2026-02-08
fix: Hive unit test flushPolicyWithParquet

PingLiuPing:
Fix hive UT error https://github.com/facebookincubator/velox/actions/runs/21786681207/job/62859220521?pr=16295

`/home/runner/work/velox/velox/velox/./velox/dwio/parquet/writer/arrow/Properties.h:407:  Check failed: (maxRowGroupBytes) > (0) maxRowGroupBytes must be positive`

PR #16310, opened 2026-02-09
fix(test): Fix HiveDataSinkTest.flushPolicyWithParquet crash

yaooqinn:
## Summary

`HiveDataSinkTest.flushPolicyWithParquet` has been crashing with:

```
Check failed: (maxRowGroupBytes) > (0) maxRowGroupBytes must be positive
```

## Root Cause

PR #15751 added `ARROW_CHECK_GT(maxRowGroupBytes, 0)` validation in Arrow's `WriterProperties::Builder`. The test at line 1152 creates `DefaultFlushPolicy(1234, 0)` where `0` is the `bytesInRowGroup` parameter. This value flows through to `WriterProperties::Builder::maxRowGroupBytes(0)`, triggering the assertion failure (`SIGABRT`).

## Fix

Replace `bytesInRowGroup=0` with `128 * 1024 * 1024` (the default value). This preserves the test's intent of controlling flushing by row count only (1234 rows), while satisfying the positive-value constraint.


FelixYBW:
duplicated with #16304

PR #16312, opened 2026-02-09
fix(cudf): Disable passing of split values to Parquet reader pending investigation into Presto breakage in lieu of reverting #14968/#16287

simoneves:
PRs #14968 (and follow-up fix in #16287) add support for reading splits in Parquet files.

That code is in itself correct, but the incoming `start` value in the Presto context appear to be incorrect.

To allow easier investigation into that, rather than reverting the entire change, this PR simply disables passing of those values to the Parquet reader, thus avoiding the problem.

Posted on behalf of @mhaseeb123

### DIFF
--- a/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
+++ b/velox/experimental/cudf/connectors/hive/CudfHiveDataSource.cpp
@@ -635,8 +635,9 @@ CudfHybridScanReaderPtr CudfHiveDataSource::createExperimentalSplitReader() {
   if (not pageIndexByteRange.is_empty()) {
     auto const pageIndexBytes = dataSource_->host_read(
         pageIndexByteRange.offset(), pageIndexByteRange.size());
-    exptSplitReader->setup_page_index(cudf::host_span<uint8_t const>{
-        pageIndexBytes->data(), pageIndexBytes->size()});
+    exptSplitReader->setup_page_index(
+        cudf::host_span<uint8_t const>{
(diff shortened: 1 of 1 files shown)


bdice (approved):

zoltan:
why not simply just disable hive splitting for files?

simoneves:
> why not simply just disable hive splitting for files?

@zoltan If you mean `hive.file-splittable=false` that is already the `velox-testing` default, and does not appear to resolve this. If the "split start" value is passed from this new code to the Parquet reader, the file is not read properly (but with no apparent error) and the query outputs zero rows. This is the case with any query that reads a Parquet file, most obviously the TPCH tests.

majetideepak:
Ideally, we need to use VELOX_FORWARD_COMPATIBILITY or VELOX_BACKWARD_COMPATIBILITY for compatibility issues.
Do we know the root cause of Presto failures?

simoneves:
> Ideally, we need to use VELOX_FORWARD_COMPATIBILITY or VELOX_BACKWARD_COMPATIBILITY...

I considered that, but decided to keep in separate in case we had to invoke one or other of those for other reasons while this got fixed.

> Do we know the root cause of Presto failures?

Not yet but investigating. The specific problem is that Presto is sending values for `start` which seem incorrect given the schema (e.g. 5).

mhaseeb123:
Folks, I have found the bug (connector Id not propagated to the split) in the `CudfHiveConnectorTestBase.cpp`. Working on a fix now.

mhaseeb123:
Superseded by #16335 16335

PR #16307, opened 2026-02-09
feat: Support decimal type for Spark checked_multiply function

n0r0shi:
## Summary

- Add `checked_multiply` for decimal types that throws on overflow instead
  of returning null, enabling Spark ANSI mode support for decimal multiplication

## Test plan

- [x] Normal cases: all 4 input type combinations (short+short, short+long,
      long+short, long+long)
- [x] Edge cases: multiply by zero, negative numbers
- [x] Precision capped at 38, no overflow (small values and near-boundary)
- [x] Positive overflow throws error
- [x] Negative overflow throws error
- [x] Boundary overflow (max * 2, etc.) throws error

Closing https://github.com/facebookincubator/velox/issues/16306

PR #16323, opened 2026-02-10
feat: Add checked_divide for Spark decimal arithmetic

### LINKED ISSUE #16322: Add checked_divide for Spark decimal types"
### Description

  Part of #3869

   ## Description

   PR #10083 added `checked_add`, `checked_subtract`, `checked_multiply`, and
   `checked_divide` for integer and floating-point types to support Spark's ANSI
   mode. PR #14935 later added `checked_divide` for integer types specifically. However,
   decimal `checked_divide` was not included.

   In Spark's ANSI mode (`spark.sql.ansi.enabled=true`), decimal arithmetic overflow
   should raise an error rather than returning null. Decimal division goes through
   `CheckOverflow(BinaryArithmetic, decimalType, nullOnOverflow=false)`, which
   requires a checked variant that throws on overflow.

   Currently, the existing `divide` for decimal types returns null on overflow or
   division by zero (matching Spark's default non-ANSI behavior). We need a decimal
   `checked_divide` that returns an error status on overflow and division by zero
   instead.

   This follows the same pattern established by:
   - #16301 / PR #16302 — `checked_add` and `checked_subtract` for decimal
   - #16306 / PR #16307 — `checked_multiply` for decimal

   ## Scope

   - `checked_divide` for all decimal input combinations (short/long × short/long)
   - Both `allowPrecisionLoss` and `denyPrecisionLoss` variants
   - Division by zero throws "Division by zero"
   - Overflow (e.g., dividing max decimal by a tiny value) throws "Decimal overflow
    in divide"

n0r0shi:
Closes #16322

  ## Summary

  - Add `checked_divide` for decimal types that throws on overflow and division by zero instead of returning null, enabling Spark ANSI mode support for decimal division

  ## Test plan

  - [x] Normal cases: all 4 input type combinations (short/short, long/long, short/long, long/short)
  - [x] Division by zero throws error for all type combinations
  - [x] Overflow throws error (max 38-digit decimal divided by tiny value)
  - [x] Both `allowPrecisionLoss` and `denyPrecisionLoss` variants registered

 ## Notes on naming conventions

  In Spark, `Divide` (true division) maps to Velox's `divide`, and `IntegralDivide` (integer division) maps to Velox's `div`. Their ANSI-mode counterparts are `checked_divide` and `checked_div`, respectively.

  In Spark, there are no separate "checked" expression classes. The same expression handles both ANSI and non-ANSI behavior, controlled by the [EvalMode flag](https://github.com/apache/spark/blob/branch-3.5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L854-L857).
In Velox, the checked variants are registered as separate functions.

### DIFF
--- a/velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
+++ b/velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
@@ -181,6 +181,26 @@ class DecimalArithmeticTest : public SparkFunctionBaseTest {
         "checked_multiply", tType, uType, t, u, errorMessage);
   }
 
+  template <typename T, typename U>
+  std::optional<int128_t> checkedDivide(
+      const TypePtr& tType,
+      const TypePtr& uType,
+      std::optional<T> t,
+      std::optional<U> u) {
+    return checkedDecimalArithmetic("checked_divide", tType, uType, t, u);
+  }
+
+  template <typename T, typename U>
+  void assertErrorForCheckedDivide(
+      const TypePtr& tType,
+      const TypePtr& uType,
+      std::optional<T> t,
+      std::optional<U> u,
+      const std::string& errorMessage) {
+    assertErrorForCheckedDecimalArithmetic(
+        "checked_divide", tType, uType, t, u, errorMessage);
+  }
+
   /// Runs the common normal and overflow test cases for checked add.
   void testCheckedAddCommon(const std::string& suffix = "") {
     const std::string func = "checked_add" + suffix;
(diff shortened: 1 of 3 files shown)


rui-mo (commented):
Is there any difference from the `checked_divide` functions added in https://github.com/facebookincubator/velox/pull/14935? If yes, could you please update the documentation to clarify? Thanks.

n0r0shi:
Thanks for the review @rui-mo. They're different functions for different Spark operators; the names are admittedly close, which is why the question is fair:

  - **`checked_div`** (added in #14935 by @zhli1142015) — corresponds to Spark's `IntegralDivide` (the `DIV` keyword). Result type is decimal with scale 0 (effectively an integer-rounding result; Spark's `IntegralDivide.dataType` is always `LongType` at the row level).

  - **`checked_divide`** (this PR) — corresponds to Spark's `Divide` (the `/` operator). Result type follows the Spark decimal precision/scale rules described above, with `allowPrecisionLoss` honored.

  The distinction is already noted in `decimal.rst`:

  - Existing `checked_div`: *"Corresponds to Spark's operator `div` with `spark.sql.ansi.enabled` set to true"*
  - New `checked_divide`: *"Corresponds to Spark's operator `/` with `spark.sql.ansi.enabled` set to true"*

  Happy to add a cross-reference note in either docstring if you think that'd help future readers avoid the same confusion — let me know.

n0r0shi:
Just rebased on current main and pushed. The test helpers in `DecimalArithmeticTest.cpp` are now refactored to use the shared `checkedDecimalArithmetic` / `assertErrorForCheckedDecimalArithmetic` templates (and thin `checkedDivide` / `assertErrorForCheckedDivide` wrappers) — consistent with the pattern introduced in #16302 and #16307.

velox/functions/sparksql/tests/DecimalArithmeticTest.cpp
```diff
@@ -1231,5 +1251,52 @@ TEST_F(DecimalArithmeticTest, checkedMultiply) {
...
+
+  // Overflow throws (and try() returns null).
+  // Dividing a large value by a very small value can overflow.
+  assertErrorForCheckedDivide<int128_t, int128_t>(
+      DECIMAL(38, 0),
+      DECIMAL(38, 6),
+      HugeInt::parse("99999999999999999999999999999999999999"),
+      1,
+      "Decimal overflow in divide");
+}
```
rui-mo, line 1300: Would you please add tests for the `checked_divide_deny_precision_loss` as well?

rui-mo (commented):
Thanks!

PR #16328, opened 2026-02-10
feat: Support RESPECT NULLS for Spark collect_list aggregate function

yaooqinn:
## Summary

Add a new QueryConfig option `spark.collect_list.ignore_nulls` (default: `true`) to control null handling in the Spark `collect_list` aggregate function.

When set to `false` (RESPECT NULLS), null values are included in the output array instead of being skipped. This aligns with Spark 4.2+ RESPECT NULLS / IGNORE NULLS syntax ([SPARK-55256](https://issues.apache.org/jira/browse/SPARK-55256)).

## Changes

- **`velox/core/QueryConfig.h`**: Add `kSparkCollectListIgnoreNulls` config constant and `sparkCollectListIgnoreNulls()` accessor
- **`velox/exec/SimpleAggregateAdapter.h`**: Add public `fn()` accessor to allow factory lambdas to configure FUNC members after construction (follows the pattern used by Presto's array_agg)
- **`velox/functions/sparksql/aggregates/CollectListAggregate.cpp`**:
  - Add `ignoreNulls_` member to `CollectListAggregate`
  - Pass config to `AccumulatorType` via the fn pointer
  - Remove static `toIntermediate()` (cannot access runtime config; see note below)
  - Factory lambda reads config and sets `ignoreNulls_` via `fn()` accessor
- **`velox/functions/sparksql/aggregates/tests/CollectListAggregateTest.cpp`**: Add 3 new tests:
  - `respectNulls`: global aggregation with nulls included
  - `respectNullsGroupBy`: group-by aggregation with nulls included
  - `respectNullsAllNulls`: all-null input produces array of nulls (not empty array)
- **`velox/docs/functions/spark/aggregate.rst`**: Updated documentation

## Note on `toIntermediate` removal

The original `toIntermediate()` was a static method that always skipped nulls. Since it's called as `FUNC::toIntermediate()` via SFINAE detection in `SimpleAggregateAdapter`, it cannot access instance-level `ignoreNulls_` config. Removing it ensures correctness for RESPECT NULLS at the cost of using the standard accumulator path for partial aggregation.

## Test Plan

All 7 tests pass (4 existing + 3 new):
```
[  PASSED  ] 7 tests.
  CollectListAggregateTest.groupBy
  CollectListAggregateTest.global
  CollectListAggregateTest.ignoreNulls
  CollectListAggregateTest.allNullsInput
  CollectListAggregateTest.respectNulls
  CollectListAggregateTest.respectNullsGroupBy
  CollectListAggregateTest.respectNullsAllNulls
```

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -503,7 +503,7 @@ class QueryConfig {
 
   /// If true, collect_list aggregate function will ignore nulls in the input.
   /// Defaults to true to match Spark's default behavior. Set to false to
-  /// include nulls (RESPECT NULLS).
+  /// include nulls (RESPECT NULLS). Introduced in Spark 4.2 (SPARK-55256).
   static constexpr const char* kSparkCollectListIgnoreNulls =
       "spark.collect_list.ignore_nulls";
 


velox/exec/SimpleAggregateAdapter.h
```diff
@@ -589,6 +589,12 @@ class SimpleAggregateAdapter : public Aggregate {
   std::vector<DecodedVector> inputDecoded_;
   DecodedVector intermediateDecoded_;
 
+ public:
+  FUNC* fn() {
```
Yuhta: This is hacky.  @kagamiori Is there a way to pass query config to an aggregate function created by `SimpleAggregateAdapter`?

yaooqinn: Thanks for the feedback @Yuhta! I extended `SimpleAggregateAdapter` to support passing `QueryConfig` through the framework. Added a SFINAE-detected `initialize(step, argTypes, resultType, config)` overload — if `FUNC` defines it, the config is forwarded during construction. The `fn()` accessor is removed.

cc @kagamiori

yaooqinn:
Thank you for the check @Yuhta

velox/functions/sparksql/aggregates/CollectListAggregate.cpp
```diff
@@ -38,23 +38,27 @@ class CollectListAggregate {
   /// behavior can only be achieved when the default-null behavior is disabled.
   static constexpr bool default_null_behavior_ = false;
 
-  static bool toIntermediate(
```
rui-mo, line 41: Is `toIntermediate` called only if `supportsToIntermediate()` returns true? I wonder if the `toIntermediate` is taking effect in the first place.

velox/core/QueryConfig.h
```diff
@@ -501,6 +501,12 @@ class QueryConfig {
   static constexpr const char* kSparkJsonIgnoreNullFields =
       "spark.json_ignore_null_fields";
 
+  /// If true, collect_list aggregate function will ignore nulls in the input.
+  /// Defaults to true to match Spark's default behavior. Set to false to
+  /// include nulls (RESPECT NULLS).
```
rui-mo: Could you please also document which Spark version introduced this configuration? Thanks.

rui-mo (commented):
Thanks!

velox/functions/sparksql/aggregates/CollectListAggregate.cpp
yaooqinn: Yes, `toIntermediate` was taking effect — the SFINAE trait `support_to_intermediate` in `SimpleAggregateAdapter.h` detects it via `decltype(&T::toIntermediate)`, and since the original was a valid static member, `supportsToIntermediate()` returned `true` and the adapter called it during partial aggregation.

It was intentionally removed because `toIntermediate()` is static and cannot access the runtime `ignoreNulls_` config. Without it, partial aggregation falls back to the accumulator path (`addInput` + `extractValues`), which correctly respects the config. Added a comment explaining this.

velox/core/QueryConfig.h
yaooqinn: Done. Added "Introduced in Spark 4.2 (SPARK-55256)" to the config comment.

rui-mo (approved):
Thanks!

velox/functions/sparksql/aggregates/CollectListAggregate.cpp
rui-mo: I understand. Thanks for the explanation!

PR #16331, opened 2026-02-10
fix(cudf): Fix velox_cudf_s3_read_test

simoneves:
This test, unlike it's Hive CPU equivalent, was not using the prescribed helper function to locate files in the source code tree by relative path.

This test has also had issues in the past with a RMM shutdown crash, and is currently disabled in `velox_testing` CI, but I am re-testing it in the hope that the RMM problem has since been resolved.

Update: Test runs successful. RMM problem gone away. This test can then be re-enabled in `velox-testing` CI.

### DIFF
--- a/velox/experimental/cudf/tests/S3ReadTest.cpp
+++ b/velox/experimental/cudf/tests/S3ReadTest.cpp
@@ -65,8 +65,9 @@ class S3ReadTest : public S3Test, public ::test::VectorTestBase {
 } // namespace
 
 TEST_F(S3ReadTest, s3ReadTest) {
-  const auto sourceFile =
-      "/velox/velox/dwio/parquet/tests/examples/int.parquet";
+  const auto sourceFile = test::getDataFilePath(
+      "velox/experimental/cudf/tests",
+      "../../../dwio/parquet/tests/examples/int.parquet");
   const char* bucketName = "data";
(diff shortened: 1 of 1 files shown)


velox/experimental/cudf/tests/S3ReadTest.cpp
```diff
@@ -65,8 +65,9 @@ class S3ReadTest : public S3Test, public ::test::VectorTestBase {
 } // namespace
 
 TEST_F(S3ReadTest, s3ReadTest) {
-  const auto sourceFile =
-      "/velox/velox/dwio/parquet/tests/examples/int.parquet";
+  const auto sourceFile = test::getDataFilePath(
+      "velox/experimental/cudf/tests",
+      "../../../dwio/parquet/tests/examples/int.parquet");
```
mhaseeb123, line 70: IIRC (and JFYI) this only works if the test is run via `ctest` while being in `velox/_build/release` directory.

simoneves: Right, but the same is true for at least a couple of the other tests, and that's where CI runs them.

simoneves: The implementation of `getDataFilePath` has some Facebook trickery in it, which presumably is intended to make such tests work in some other build system of theirs too.

karthikeyann:
@peterenescu could you import this PR?

PR #16332, opened 2026-02-10
fix: Print verbose errors in cudf deps updater

mhaseeb123:
## Description

This PR updates the `update-cudf-deps.sh` to not silently exit without any message or effect when, for example, `read` fails due to GH rate limiting and throws a verbose message such as

```bash
Error: Failed to fetch VERSION file from branch main
  Try: `gh auth login` (or set GH_TOKEN) before running this script
  Create a GitHub personal access token at https://github.com/settings/tokens/new
  (Select scope: public_repo)
```

### DIFF
--- a/scripts/update-cudf-deps.sh
+++ b/scripts/update-cudf-deps.sh
@@ -61,12 +61,12 @@ fetch_github_api() {
     echo "Error: gh api failed" >&2
     echo "Response: $response" >&2
   else
-    local curl_opts="-s --max-time 30 -w %{http_code}"
+    local curl_opts=(-s --max-time 30 -w '%{http_code}')
     if [[ -n $GH_TOKEN ]]; then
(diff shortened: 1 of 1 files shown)


bdice (approved):
Recently I used this approach for a “maybe curl, maybe gh” style. Might help simplify the code if you think using gh for authentication when available is sufficient, rather than taking an environment variable.

```bash
fetch_github_api() {
    local endpoint=$1
    if command -v gh &>/dev/null; then
        gh api "${endpoint}" 2>/dev/null && return
    fi
    curl -sf "https://api.github.com/${endpoint}"
}
```

mhaseeb123:
Let's merge this @karthikeyann @majetideepak ?

bdice:
I fixed a small issue with pre-commit. Should be good to merge now.

mhaseeb123:
Thanks @bdice

karthikeyann:
Thank you for updating this.

PR #16339, opened 2026-02-11
feat(sparksql): Add Spark-specific sequence function for integer types

yaooqinn:
## Summary

Add Spark `sequence` function support for all integer types (`tinyint`, `smallint`, `integer`, `bigint`) by reusing the shared `SequenceFunction<T, K>` template with Spark-specific signatures and registration.

## Changes

- Extract `SequenceFunction` template into `lib/Sequence.h` (declaration) and `lib/Sequence.cpp` for reuse across dialects
- Add `sparksql/SequenceFunction.cpp` with Spark signatures and explicit instantiations for `int8_t` and `int16_t` types
- Keep Presto-specific signatures and registration in `prestosql/Sequence.cpp`
- Add 12 Spark-specific tests covering all integer types, null handling, and error cases

## Spark Behavior

| Expression | Result |
|---|---|
| `sequence(1, 5)` | `[1, 2, 3, 4, 5]` |
| `sequence(cast(1 as tinyint), cast(5 as tinyint))` | `[1, 2, 3, 4, 5]` |
| `sequence(1, 10, 3)` | `[1, 4, 7, 10]` |
| `sequence(5, 1)` | `[5, 4, 3, 2, 1]` |

### DIFF
--- a/velox/connectors/hive/iceberg/IcebergMetadataColumns.h
+++ b/velox/connectors/hive/iceberg/IcebergMetadataColumns.h
@@ -28,6 +28,11 @@ struct IcebergMetadataColumn {
   std::shared_ptr<const Type> type;
   std::string doc;
 
+  // Position delete file's metadata column ID, see
+  // https://iceberg.apache.org/spec/#position-delete-files.
+  static constexpr int32_t kPosId = 2'147'483'545;
+  static constexpr int32_t kFilePathId = 2'147'483'546;
+
   IcebergMetadataColumn(
       int _id,
       const std::string& _name,
@@ -37,7 +42,7 @@ struct IcebergMetadataColumn {
 
   static std::shared_ptr<IcebergMetadataColumn> icebergDeleteFilePathColumn() {
     return std::make_shared<IcebergMetadataColumn>(
-        2147483546,
+        kFilePathId,
         "file_path",
         VARCHAR(),
         "Path of a file in which a deleted row is stored");
@@ -45,7 +50,7 @@ struct IcebergMetadataColumn {
 
   static std::shared_ptr<IcebergMetadataColumn> icebergDeletePosColumn() {
     return std::make_shared<IcebergMetadataColumn>(
-        2147483545,
+        kPosId,
         "pos",
         BIGINT(),
         "Ordinal position of a deleted row in the data file");
--- a/velox/connectors/hive/iceberg/IcebergSplitReader.cpp
+++ b/velox/connectors/hive/iceberg/IcebergSplitReader.cpp
@@ -16,6 +16,9 @@
 
 #include "velox/connectors/hive/iceberg/IcebergSplitReader.h"
 
+#include <folly/lang/Bits.h>
+
+#include "velox/common/encode/Base64.h"
 #include "velox/connectors/hive/iceberg/IcebergDeleteFile.h"
 #include "velox/connectors/hive/iceberg/IcebergMetadataColumns.h"
 #include "velox/connectors/hive/iceberg/IcebergSplit.h"
@@ -82,11 +85,19 @@ void IcebergSplitReader::prepareSplit(
         // Skip the delete file if all delete positions are before this split.
         // TODO: Skip delete files where all positions are after the split, if
         // split row count becomes available.
-        auto posColumn = IcebergMetadataColumn::icebergDeletePosColumn();
-        if (auto posUpperBoundIt = deleteFile.upperBounds.find(posColumn->id);
-            posUpperBoundIt != deleteFile.upperBounds.end()) {
-          auto deleteUpperBound = folly::to<uint64_t>(posUpperBoundIt->second);
-          if (deleteUpperBound < splitOffset_) {
+        if (auto iter =
+                deleteFile.upperBounds.find(IcebergMetadataColumn::kPosId);
+            iter != deleteFile.upperBounds.end()) {
+          auto decodedBound = encoding::Base64::decode(iter->second);
+          VELOX_CHECK_EQ(
+              decodedBound.size(),
+              sizeof(uint64_t),
+              "Unexpected decoded size for positional delete upper bound.");
+          uint64_t posDeleteUpperBound;
+          std::memcpy(
+              &posDeleteUpperBound, decodedBound.data(), sizeof(uint64_t));
+          posDeleteUpperBound = folly::Endian::little(posDeleteUpperBound);
+          if (posDeleteUpperBound < splitOffset_) {
             continue;
           }
         }
--- a/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
+++ b/velox/connectors/hive/iceberg/tests/IcebergReadTest.cpp
@@ -15,7 +15,9 @@
  */
 
 #include <folly/Singleton.h>
+#include <folly/lang/Bits.h>
 #include "velox/common/base/tests/GTestUtils.h"
+#include "velox/common/encode/Base64.h"
 #include "velox/common/file/FileSystems.h"
 #include "velox/connectors/hive/HiveConnectorSplit.h"
 #include "velox/connectors/hive/iceberg/IcebergConnector.h"
@@ -920,7 +922,13 @@ TEST_F(HiveIcebergTest, skipDeleteFileByPositionUpperBound) {
        makeFlatVector<int64_t>({0, 1, 2})})};
   writeToFile(deleteFilePath->getPath(), deleteVectors);
 
-  // upperBound "2" is the max position in the delete file.
+  // upperBound "2" is the max position in the delete file. Iceberg stores
+  // long bounds as 8-byte little-endian binary, then Base64 encodes them.
+  uint64_t upperBound = 2;
+  auto upperBoundLE = folly::Endian::little(upperBound);
+  auto encodedUpperBound = encoding::Base64::encode(
+      std::string_view(
+          reinterpret_cast<const char*>(&upperBoundLE), sizeof(upperBoundLE)));
   IcebergDeleteFile deleteFile(
       FileContent::kPositionalDeletes,
       deleteFilePath->getPath(),
@@ -930,7 +938,7 @@ TEST_F(HiveIcebergTest, skipDeleteFileByPositionUpperBound) {
           std::fopen(deleteFilePath->getPath().c_str(), "r")),
       {},
       {},
-      {{posColumn->id, "2"}});
+      {{posColumn->id, encodedUpperBound}});
 
   auto plan = PlanBuilder()
                   .startTableScan()
--- a/velox/exec/fuzzer/FuzzerUtil.cpp
+++ b/velox/exec/fuzzer/FuzzerUtil.cpp
@@ -24,6 +24,7 @@
 #include "velox/exec/fuzzer/DuckQueryRunner.h"
 #include "velox/exec/fuzzer/PrestoQueryRunner.h"
 #include "velox/expression/SignatureBinder.h"
+#include "velox/functions/prestosql/types/IPAddressType.h"
 #include "velox/functions/prestosql/types/IPPrefixType.h"
 
 using namespace facebook::velox::dwio::catalog::fbhive;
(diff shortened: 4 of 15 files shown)


velox/functions/sparksql/SequenceFunction.cpp
```diff
@@ -0,0 +1,92 @@
...
+    const core::QueryConfig& /* config */) {
+  switch (inputArgs[0].type->kind()) {
+    case TypeKind::BIGINT:
+      return std::make_shared<SparkSequenceFunction<int64_t>>();
+    case TypeKind::INTEGER:
+      return std::make_shared<SparkSequenceFunction<int32_t>>();
+    case TypeKind::SMALLINT:
+      return std::make_shared<SparkSequenceFunction<int16_t>>();
+    case TypeKind::TINYINT:
+      return std::make_shared<SparkSequenceFunction<int8_t>>();
```
rui-mo: > Spark's sequence supports all IntegralType (tinyint, smallint,
integer, bigint), while Presto's only supports bigint. Adding
Spark-specific types directly to Presto's implementation would pollute
its function namespace. 

Is the above the only difference between Spark and Presto? If so, can we support Spark signatures by registering them with the types required by Spark, and reuse the original implementation?

velox/functions/prestosql/Sequence.h
```diff
@@ -0,0 +1,286 @@
...
+  VELOX_FAIL("Unexpected start/end type for argument INTERVAL_YEAR_MONTH");
+}
+
+inline int128_t getStepCount(int32_t start, int32_t end, int32_t step) {
+  return diffDate(DateTimeUnit::kMonth, start, end) / step + 1;
+}
+
+inline int128_t getStepCount(Timestamp start, Timestamp end, int32_t step) {
+  return diffTimestamp(DateTimeUnit::kMonth, start, end) / step + 1;
+}
```
rui-mo: Do we need to expose above functions as APIs? It looks like Spark only requires the access to `SequenceFunction` for registration.

velox/docs/functions/spark/array.rst
```diff
@@ -292,6 +292,21 @@ Array Functions
         SELECT shuffle(array(0, 0, 0), 0); -- [0, 0, 0]
         SELECT shuffle(array(1, NULL, 1, NULL, 2), 0); -- [2, 1, NULL, NULL, 1]
 
+.. spark:function:: sequence(start, stop) -> array(T)
+                    sequence(start, stop, step) -> array(T)
```
rui-mo, line 286: nit: follow the alphabetical order to add the documentation.

rui-mo (commented):
Thanks for the update.

yaooqinn:
@rui-mo Both comments addressed:
1. Moved helper functions (`toInt64`, `add`, `getStepCount`) into a `detail` namespace — only `SequenceFunction` is exposed as public API.
2. Fixed the documentation order — `sequence` now comes before `shuffle` alphabetically.

velox/functions/lib/Sequence.h
```diff
@@ -0,0 +1,299 @@
...
+
+} // namespace detail
+
+/// Sequence function that generates an array of values from start to stop
+/// (inclusive) with an optional step. Supports integer types (tinyint,
+/// smallint, integer, bigint), date, and timestamp.
+template <typename T, typename K>
+class SequenceFunction : public exec::VectorFunction {
+ public:
+  void apply(
```
jinchengchenghh, line 28: Use this format then you don't need to move so much code from cpp file to header file.
Header
```
template <typename T, typename K>
class SequenceFunction : public exec::VectorFunction {
public:
  void apply(...);
};

```
cpp file
```
template <typename T, typename K>
void SequenceFunction<T, K>::apply(...) {
  // implementation
}

// Explicit instantiations:
template class SequenceFunction<int64_t, int64_t>;
template class SequenceFunction<double, double>;

```

velox/functions/prestosql/Sequence.cpp
```diff
@@ -90,164 +114,184 @@ int128_t getStepCount(Timestamp start, Timestamp end, int32_t step) {
...
+  }
+  // Handle Date
+  VELOX_USER_CHECK(
+      step % kMillisInDay == 0,
+      "sequence step must be a day interval if start and end values are dates");
+  return step / kMillisInDay;
+}
+
+// Explicit instantiations for Presto types.
+template class SequenceFunction<int64_t, int64_t>;
```
jinchengchenghh: Please move these two files to folder lib, only left prestosql specified code and spark specified code to each folder

velox/functions/sparksql/SequenceFunction.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "velox/functions/prestosql/Sequence.h"
+
+namespace facebook::velox::functions::sparksql {
+namespace {
+
+std::vector<std::shared_ptr<exec::FunctionSignature>>
+sparkSequenceSignatures() {
```
jinchengchenghh: remove spark prefix

velox/functions/sparksql/SequenceFunction.cpp
```diff
@@ -0,0 +1,104 @@
...
+          .returnType("array(tinyint)")
+          .argumentType("tinyint")
+          .argumentType("tinyint")
+          .argumentType("tinyint")
+          .build(),
+  };
+  return signatures;
+}
+
+std::shared_ptr<exec::VectorFunction> makeSparkSequence(
```
jinchengchenghh: ditto

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,294 @@
...
+  }
+  // Handle Date
+  VELOX_USER_CHECK(
+      step % kMillisInDay == 0,
+      "sequence step must be a day interval if start and end values are dates");
+  return step / kMillisInDay;
+}
+
+// Explicit instantiations for Presto types.
+template class SequenceFunction<int64_t, int64_t>;
```
jinchengchenghh, line 288: Move to presto folder

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,294 @@
...
+template class SequenceFunction<int32_t, int64_t>;
+template class SequenceFunction<int32_t, int32_t>;
+template class SequenceFunction<Timestamp, int64_t>;
+template class SequenceFunction<Timestamp, int32_t>;
+// Explicit instantiations for Spark integer types.
+template class SequenceFunction<int8_t, int8_t>;
```
jinchengchenghh, line 294: Move to Spark folder

jinchengchenghh:
Please update the PR description

yaooqinn:
Rebased onto main. All 3 CI failures are unrelated to this PR:

1. **Linux GCC (ubuntu-cpp / linux.gcc)** — `TaskTest.addSplitAfterBarrier` race condition, tracked by #16510
2. **Linux GCC (other / linux.gcc)** — Same `TaskTest.addSplitAfterBarrier` failure, tracked by #16510
3. **Spark Aggregate Fuzzer Run** — `preadv` "Bad file descriptor" crash in `ReadFileInputStream::read`, filed #16543

yaooqinn:
@rui-mo @jinchengchenghh, anything more to address or ready to ship?

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,295 @@
+/*
```
rui-mo, line 1: `-inl.h` headers should not be included by file other than `Sequence.h`, see https://github.com/facebookincubator/velox/blob/f760d54f90dbf60a992b89622cdbccd917f533cd/velox/expression/CastExpr.h#L359

I assume what is needed here is `SequenceBase.h`.

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,295 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+#include "velox/expression/DecodedArgs.h"
+#include "velox/functions/lib/Sequence.h"
+#include "velox/functions/prestosql/DateTimeImpl.h"
+
+namespace facebook::velox::functions {
+namespace detail {
```
jinchengchenghh: Why not is Sequence.cpp?

rui-mo: `SequenceBase.h` includes `Sequence.h`, which looks a bit strange. Consider only exposing `SequenceFunction` in a single header for reuse. For the remaining implementations, place them in the .cpp file whenever possible. Otherwise, put them inside the detail namespace.

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,307 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/lib/Sequence.h"
+#include "velox/expression/DecodedArgs.h"
+#include "velox/functions/prestosql/DateTimeImpl.h"
+
+namespace facebook::velox::functions {
+namespace detail {
```
rui-mo: Use anonymous namespace in the cpp file.

yaooqinn:
@rui-mo All review comments have been addressed (anonymous namespace, removed -inl.h include). Could you take another look when you get a chance? Thanks!

velox/functions/lib/Sequence.cpp
rui-mo: @yaooqinn Can you please address this comment?

velox/functions/lib/Sequence.cpp
rui-mo: This one as well.

yaooqinn:
@rui-mo These two comments were already addressed in the previous push — the `VELOX_DECLARE_STATEFUL_VECTOR_FUNCTION` macros are now in `prestosql/Sequence.cpp` (for Presto's `udf_sequence`) and `sparksql/SequenceFunction.cpp` (for Spark's `udf_spark_sequence`). `lib/Sequence.cpp` only contains the shared template implementation and explicit instantiations — no registration macros. The threads show as outdated because the code has changed since the comments were made. Thanks!

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+// Explicit instantiations for all supported types.
+// Presto types:
+template class SequenceFunction<int64_t, int64_t>;
+template class SequenceFunction<int32_t, int64_t>;
+template class SequenceFunction<int32_t, int32_t>;
+template class SequenceFunction<Timestamp, int64_t>;
+template class SequenceFunction<Timestamp, int32_t>;
+// Spark-specific types:
+template class SequenceFunction<int8_t, int8_t>;
```
rui-mo, line 294: Can we move this to Spark folder? I assume we don't need to declare Spark template class in the lib folder.

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,297 @@
...
+  // Handle Date
+  VELOX_USER_CHECK(
+      step % kMillisInDay == 0,
+      "sequence step must be a day interval if start and end values are dates");
+  return step / kMillisInDay;
+}
+
+// Explicit instantiations for all supported types.
+// Presto types:
+template class SequenceFunction<int64_t, int64_t>;
```
rui-mo, line 288: Ditto. Consider moving to Prestosql folder.

velox/functions/lib/Sequence.cpp
rui-mo: @yaooqinn Sorry, I may have misunderstood. The implementation has now moved to the CPP file, and we need instantiations here.

yaooqinn:
also cc @kgpai

velox/functions/lib/Sequence.cpp
```diff
@@ -0,0 +1,297 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/lib/Sequence.h"
+#include "velox/expression/DecodedArgs.h"
+#include "velox/functions/prestosql/DateTimeImpl.h"
```
kKPulla: @yaooqinn looks like this is introducing a circular dependency of sorts as the lib/ folder is supposed to be a holder for common logic that prestosql/ can depend on but we have it the other way here. If you need this, we might have to move this from prestosql/ too probably?

cc: @Yuhta @kevinwilfong thoughts?

yaooqinn:
@kKPulla Good catch! Fixed the circular dependency — moved `diffDate()` from `prestosql/DateTimeImpl.h` to `lib/DateTimeUtil.h` where the other shared datetime utilities (`diffTimestamp`, `addToDate`, `addToTimestamp`) already live. `lib/Sequence.cpp` now includes only `lib/DateTimeUtil.h` — no more prestosql/ dependency from lib/.

@rui-mo Regarding the Spark instantiation comment — as you noted, the explicit template instantiations must stay in `lib/Sequence.cpp` since that's where the template implementation lives. The registration macros are already in the appropriate folders (`prestosql/Sequence.cpp` and `sparksql/SequenceFunction.cpp`).

PR #16344, opened 2026-02-11
feat: Add Spark TIMESTAMP_NTZ type

rui-mo:
The Spark timestamp type represents a specific instant in time with microsecond 
precision. Its value is based on the proleptic Gregorian calendar in UTC+00:00, 
making it adaptive to time zones.
In contrast, Spark’s `timestamp without time zone type` represents a local time 
with microsecond precision that is independent of any time zone. Neither type 
includes an explicit time zone parameter, and the key distinction is that one 
is time zone–aware while the other is not.
In Velox, the Timestamp type is time zone–aware because it already supports 
time zone adjustment, aligning with the semantics of Spark’s timestamp type. 
This PR introduces a new custom type to represent Spark’s TIMESTAMP_NTZ type,
and adds support for CAST(string AS timestamp_ntz) to ensure the type works
correctly.

Spark's implementation: https://github.com/apache/spark/blob/master/sql/api/src/main/scala/org/apache/spark/sql/types/TimestampNTZType.scala

### DIFF
--- a/velox/functions/sparksql/types/CMakeLists.txt
+++ b/velox/functions/sparksql/types/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(velox_spark_types TimestampNTZRegistration.cpp)
+velox_add_library(velox_spark_types TimestampNTZCastUtil.cpp TimestampNTZRegistration.cpp)
 
 velox_link_libraries(velox_spark_types velox_type velox_memory velox_expression)
 
--- a/velox/functions/sparksql/types/TimestampNTZCastUtil.cpp
+++ b/velox/functions/sparksql/types/TimestampNTZCastUtil.cpp
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "velox/functions/sparksql/types/TimestampNTZCastUtil.h"
+#include "velox/expression/EvalCtx.h"
+#include "velox/functions/lib/string/StringImpl.h"
+#include "velox/type/TimestampConversion.h"
+
+namespace facebook::velox::functions::sparksql {
+namespace {
+
+void setError(
+    exec::EvalCtx& context,
+    vector_size_t row,
+    const std::string& errorDetails) {
+  if (context.captureErrorDetails()) {
+    context.setStatus(row, Status::UserError("{}", errorDetails));
+  } else {
+    context.setStatus(row, Status::UserError());
+  }
+}
+
+StringView removeWhiteSpaces(const StringView& view) {
+  StringView output;
+  stringImpl::trimUnicodeWhiteSpace<true, true, StringView, StringView>(
+      output, view);
+  return output;
+}
+
+} // namespace
+
+void castFromString(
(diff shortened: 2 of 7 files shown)


velox/functions/sparksql/CMakeLists.txt
```diff
@@ -67,3 +67,4 @@ if(${VELOX_ENABLE_BENCHMARKS})
 endif()
 
 add_subdirectory(registration)
+add_subdirectory(types)
```
jinchengchenghh, line 70: The type is in sparksql space, can we support it in velox ArrowBridge later? 

Spark Type | Arrow Representation
-- | --
TimestampType (with TZ semantics) | timestamp(unit, "UTC")
TimestampNTZType | timestamp(unit, null)

rui-mo: The importing from and exporting to Arrow seem to work with the bigint kind, so we might don't need extra logic if we treat them as bigints, rather than timestamp type.

jinchengchenghh: Do we need to update here? https://github.com/facebookincubator/velox/blob/c14f6e3805ea9593dec3770227d79eb43e64aca9/velox/vector/arrow/Bridge.cpp#L313-L314

rui-mo: This type will currently fall into case TypeKind::BIGINT, which I expect should work for Gluten. While we’ll need to open a separate PR to verify, add tests and enable the Arrow conversions.

jinchengchenghh: The partial project now uses arrow conversion

rui-mo: What is "partial project"?

jinchengchenghh: Is timestamp without timezone in arrow format is `tsu:` ?https://arrow.apache.org/docs/dev/format/CDataInterface.html#data-type-description-format-strings, but this conversion will lost some information, velox(TIMESTAMP_NTZ) -> arrow(bigint)-> velox(bigint), this may make error in some case

> The timezone string is appended as-is after the colon character :, without any quotes. If the timezone is empty, the colon : must still be included.

jinchengchenghh: So we may disable the TIMESTAMP_NTZ support in arrow conversion

jinchengchenghh: Gluten partial project

rui-mo: @jinchengchenghh I understand your point. In this case, we’ll need a workaround to preserve its distinguishing characteristic. For example, the schema could use `tsu`, allowing it to be differentiated from the regular timestamp type during Arrow importing and exporting.

The header of this custom type needs to be introduced into `velox/vector/arrow/Bridge.cpp`, or we can consider adding an API calling custom conversion.

rui-mo: It looks better to me to add custom conversions like cast, because this helps keep the workaround within sparksql scope. 


```
exec::CastOperatorPtr getCustomTypeCastOperator(const std::string& name) {
  auto factory = getTypeFactory(name);
  if (factory) {
    return factory->getCastOperator();
  }

  return nullptr;
}
```

jinchengchenghh: Would you add this custom conversion in this PR or a follow up PR?

rui-mo: I’d prefer to add this in a separate PR. After landing the initial support, we have several follow-up items to address (see https://github.com/apache/incubator-gluten/issues/11622), and some teammates will be joining to help work through them. Does this make sense to you? Thanks.

jinchengchenghh: Sounds good!

kKPulla:
@rui-mo can you please make sure to rebase on top of https://github.com/facebookincubator/velox/pull/16533 for clean CI signals as there was a test failure fix that landed yesterday as part of it.

rui-mo:
@kKPulla I rebased this PR. Thanks.

kKPulla:
Sorry to bother again. Seems like the previous fix did not fully work. Can you try rebasing again on top of main to include https://github.com/facebookincubator/velox/pull/16539

rui-mo:
@kKPulla This change was on top of https://github.com/facebookincubator/velox/pull/16539, thanks.

PR #16349, opened 2026-02-11
fix(macos): Install protobuf@21 from source instead of Homebrew

### LINKED ISSUE #16309: build: protobuf@21 has been disabled in brew
### Problem description

Installing protobuf@21 using homebrew might result in failure because it has been disabled: https://formulae.brew.sh/formula/protobuf@21

We need to swap to the install steps used in the Linux variants.

### System information

n/a

### CMake log

```bash

```

jkhaliqi:
closes #16309

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -76,6 +76,7 @@ jobs:
           install_velox_deps_from_brew
           install_gflags
           install_glog
+          install_protobuf
           # Needed for faiss to find BLAS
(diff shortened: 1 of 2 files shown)


PingLiuPing (approved):
Thanks @jkhaliqi

mbasmanova (changes requested):
[deleted]

mbasmanova (approved):

PR #16350, opened 2026-02-11
feat: Add IcebergConfig

PingLiuPing:
Add an IcebergConfig wrapper class for Iceberg‑specific settings and wire it into IcebergConnector.
Move function prefix config/defaults into the new class.


mbasmanova (approved):
Thanks.

PR #16356, opened 2026-02-12
feat: Add checked_unaryminus for Spark ANSI mode integer overflow detection

### LINKED ISSUE #16355: Add checked_unaryminus for Spark ANSI mode integer overflow detection
### Description

In Spark ANSI mode, UnaryMinus on integer types should throw an arithmetic overflow error when negating the minimum value (e.g., -(-128) for TINYINT, -(-2147483648) for INT). This is because two's complement integers have an asymmetric range — MIN has no positive counterpart.

  Velox currently has no checked variant of unary minus, so Gluten cannot enforce this ANSI behavior and must fall back to Spark.

  Proposed Change

  Add a checked_unaryminus Spark function that:
  - For integral types, throws an error when the input equals std::numeric_limits<T>::min() (i.e., INT8_MIN, INT16_MIN, INT32_MIN, INT64_MIN)
  - For non-integral types, behaves identically to regular negation (decimals have a symmetric range and cannot overflow on negation)
  - Returns a Status to support TRY() wrapping (returns null instead of throwing)

  Covers TINYINT, SMALLINT, INTEGER, and BIGINT.

  Spark relies on `Math.negateExact` which has the same logic to throw an error.
```
    public static long negateExact(long a) {
        if (a == Long.MIN_VALUE) {
            throw new ArithmeticException("long overflow");
        }

        return -a;
    }
```

n0r0shi:
## Summary

  Add `checked_unaryminus` Spark function that throws on integer overflow when negating
  the minimum value of integral types (TINYINT, SMALLINT, INTEGER, BIGINT). This matches
  Spark's ANSI mode behavior where `UnaryMinus` uses `Math.negateExact()`.

  - Returns `Status` to support `TRY()` wrapping (returns null instead of throwing)
  - For non-integral types, behaves identically to regular negation

  ## Test plan

  - Normal negation for all integer types (positive, negative, zero)
  - Overflow detection: negating `MIN_VALUE` throws for int8, int16, int32, int64
  - `TRY(checked_unaryminus(MIN_VALUE))` returns null

  Closes #16355

### DIFF
--- a/velox/docs/functions/spark/math.rst
+++ b/velox/docs/functions/spark/math.rst
@@ -105,6 +105,11 @@ Mathematical Functions
     Returns the result of subtracting y from x. The types of x and y must be the same.
     For integral types, overflow results in an error. Corresponds to Spark's operator ``-`` with ``failOnError`` as true.
 
+.. function:: checked_unaryminus(x) -> [same as x]
+
+    Returns the negative of ``x``. For integral types, throws an error when ``x`` is the minimum value
+    of the type (e.g., ``-128`` for tinyint). Corresponds to Spark's unary ``-`` operator with ``failOnError`` as true.
+
 .. spark:function:: cos(x) -> double
 
     Returns the cosine of ``x``.
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -666,4 +666,21 @@ struct CheckedIntegralDivideFunction {
   }
 };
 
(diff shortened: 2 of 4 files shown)


n0r0shi:
I've added a doc, ready for a review. Thanks.

n0r0shi:
Closing as this is covered by #16361, which implements unaryminus ANSI support via the config-based approach. Since unaryminus doesn't have a try_* variant, config-based is the right fit here. See discussion in #16361.

PR #16359, opened 2026-02-12
feat: Add checked_element_at for Spark ANSI mode array/map access

### LINKED ISSUE #16358: Add checked_element_at for Spark ANSI mode array/map access
### Description

Summary

  In Spark ANSI mode, `ElementAt` should throw instead of returning null when:
  - Accessing an array with an out-of-bounds index
  - Accessing a map with a key that doesn't exist

  Velox currently has no checked variant of `element_at`, so Gluten cannot enforce this ANSI behavior and must fall back to Spark.

  Proposed Change

  Add a `checked_element_at` function that reuses the existing SubscriptImpl template infrastructure with different flags:
  - `allowOutOfBound = false` — throws on array index out of bounds
  - `throwOnMissingMapKey = true` — throws on missing map key

  Both error paths integrate with Velox's `TRY()` mechanism (returns null instead of throwing when wrapped in TRY).

  Spark Reference

  - [Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala) uses `failOnError` parameter to control ANSI behavior.

n0r0shi:
## Summary

  Add `checked_element_at` that throws on array out-of-bounds access and missing
  map keys, instead of returning null. This matches Spark's ANSI mode behavior
  where `ElementAt` with `failOnError=true` raises exceptions.

  Reuses the existing `SubscriptImpl` template with `allowOutOfBound=false` and
  `throwOnMissingMapKey=true`. Both error paths support `TRY()` wrapping.

  ## Test plan

  - Array: normal access, out-of-bounds throws, `TRY()` returns null
  - Map: normal access, missing key throws, `TRY()` returns null

  Closes #16358

### DIFF
--- a/velox/docs/functions/spark/array.rst
+++ b/velox/docs/functions/spark/array.rst
@@ -230,6 +230,24 @@ Array Functions
         SELECT concat(array(1, 2), array(1, 2), array(1, null)); -- [1, 2, 1, 2, 1, NULL]
         SELECT concat(array(array(1, 2)), array(array(1, null))); -- [[1, 2], [1, NULL]]
 
+.. spark:function:: element_at(array(E), index) -> E
+
+    Returns element of ``array`` at given 1-based ``index``. If ``index`` < 0, accesses elements
+    from the last to the first. Throws an error if ``index`` is 0. Returns ``NULL`` if the index
+    exceeds the length of the array.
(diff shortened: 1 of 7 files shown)


n0r0shi:
Added a doc. Ready for a review.

PR #16365, opened 2026-02-12
fix: Iceberg commit message when writer rotated

PingLiuPing:
Fixes commit message correctness for multi-file writes by updating Iceberg writer rotation handling and tracking per-file row counts and emitting one commit task per rotated file.
Fix parquet writer does not honor smaller values of `max-target-file-size` config.

These are the adjustments required after #16077.

### DIFF
--- a/velox/connectors/hive/HiveDataSink.h
+++ b/velox/connectors/hive/HiveDataSink.h
@@ -426,6 +426,8 @@ struct HiveFileInfo {
   std::string targetFileName;
   /// Size of the file in bytes.
   uint64_t fileSize{0};
+  /// Number of rows in the file.
+  uint64_t numRows{0};
 };
 
 struct HiveWriterInfo {
@@ -449,8 +451,11 @@ struct HiveWriterInfo {
   const std::shared_ptr<memory::MemoryPool> writerPool;
   const std::shared_ptr<memory::MemoryPool> sinkPool;
   const std::shared_ptr<memory::MemoryPool> sortPool;
-  int64_t numWrittenRows = 0;
-  int64_t inputSizeInBytes = 0;
+  /// Total rows written by this writer across all files.
+  uint64_t numWrittenRows = 0;
+  /// Rows written to the current file; reset to 0 when the file is finalized.
+  uint64_t currentFileWrittenRows{0};
+  uint64_t inputSizeInBytes = 0;
   /// File sequence number for tracking multiple files written due to size-based
   /// splitting. Incremented each time the writer rotates to a new file.
   /// Used to generate sequenced file names (e.g., file_1.orc, file_2.orc).
--- a/velox/connectors/hive/iceberg/IcebergDataSink.cpp
+++ b/velox/connectors/hive/iceberg/IcebergDataSink.cpp
@@ -315,30 +315,36 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
           insertTableHandle_);
 
   for (auto i = 0; i < writerInfo_.size(); ++i) {
-    const auto& info = writerInfo_.at(i);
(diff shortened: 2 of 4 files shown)


velox/connectors/hive/HiveDataSink.h
```diff
@@ -426,6 +426,8 @@ struct HiveFileInfo {
   std::string targetFileName;
   /// Size of the file in bytes.
   uint64_t fileSize{0};
+  /// Number of rows in the file.
+  uint64_t rowCount{0};
```
mbasmanova: nit: rowCount -> numRows

velox/connectors/hive/iceberg/IcebergDataSink.cpp
```diff
@@ -315,30 +315,36 @@ std::vector<std::string> IcebergDataSink::commitMessage() const {
...
-    const auto& info = writerInfo_.at(i);
-    VELOX_CHECK_NOT_NULL(info);
+    const auto& writerInfo = writerInfo_.at(i);
+    VELOX_CHECK_NOT_NULL(writerInfo);
+
     // Following metadata (json format) is consumed by Presto CommitTaskData.
     // It contains the minimal subset of metadata.
     // TODO: Complete metrics is missing now and this could lead to suboptimal
     // query plan, will collect full iceberg metrics in following PR.
     // clang-format off
```
mbasmanova: // clang-format off

this should go inside the loop

PingLiuPing:
> @PingLiuPing Is it possible to add a test to protect this behavior?

@mbasmanova Thank you for the quick review.

I attempted to add a test for this behavior. And found that triggering writer rotation requires writing a large amount of data, since the default Parquet flush policy rotates only after 1,024,000 rows or 128MB.

One possible approach is to allow IcebergInsertTableHandle to accept WriterOptions, so that tests can inject customized writer options with a smaller flush threshold. For example:
```
auto writerOptions = std::make_shared<parquet::WriterOptions>();
writerOptions->compressionKind = common::CompressionKind::CompressionKind_NONE;
writerOptions->enableDictionary = false;
writerOptions->flushPolicyFactory = []() {
  return std::make_unique<parquet::DefaultFlushPolicy>(256, 8 * 1024);
};
```

This would allow us to trigger writer rotation with a small dataset in tests.

Let me explore whether there is a cleaner or more appropriate approach before proceeding.

PingLiuPing:
@mbasmanova 

Ideally, we should be able to set kMaxTargetFileSizeSession to an arbitrary value and rely on the writer to rotate once the file size reaches that limit.
But, rotation depends on `ioStats_->rawBytesWritten()`, which only advances when the writer flushes.
I think this configuration does not currently work correctly for smaller values (e.g., values smaller than the default Parquet flush threshold bytesInRowGroup_). In such cases, setting kMaxTargetFileSizeSession to a smaller value does not result in generating files close to the expected size.

This is likely because Parquet’s flush logic is driven by row group thresholds (row count or row group size), rather than the target file size:
```
bool shouldFlush(
    const dwio::common::StripeProgress& stripeProgress) override {
  return stripeProgress.stripeRowCount >= rowsInRowGroup_ ||
         stripeProgress.stripeSizeEstimate >= bytesInRowGroup_;
}
```
This issue may also exist in the DWRF writer, although its flush policy is more complex. Parquet’s behavior is relatively straightforward since it is purely based on row count and row group size.

Would it make sense for me to open a separate PR to address this issue and add proper test coverage there?

velox/connectors/hive/iceberg/IcebergDataSink.cpp

velox/connectors/hive/HiveDataSink.h
PingLiuPing: Thanks.

PR #16384, opened 2026-02-13
feat(cursor): Support parallel execution in TaskDebuggerCursor

pedroerp:
Summary:
Adding support for parallel (multi-driver) execution in
TaskDebuggerCursor. Parallel execution has a consumer callback that will add
task output to the output queue, and block until the client has consumed it.

Creating a new cursor class and sharing the common logic through a base task
cursor class.

Differential Revision: D93172900

### DIFF
--- a/velox/exec/Cursor.cpp
+++ b/velox/exec/Cursor.cpp
@@ -535,138 +535,19 @@ class SingleThreadedTaskCursor : public TaskCursorBase {
   std::exception_ptr error_;
 };
 
-/// A debugging cursor for interactive task execution.
+/// Common base class for debugging cursors that support breakpoints.
 ///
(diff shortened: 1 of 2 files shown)


velox/exec/Cursor.cpp
```diff
@@ -772,10 +671,288 @@ class TaskDebuggerCursor : public TaskCursorBase {
...
+  void setNoMoreSplits() override {
+    VELOX_CHECK(!noMoreSplits_);
+    noMoreSplits_ = true;
+  }
+
+  bool noMoreSplits() const override {
+    return noMoreSplits_;
+  }
+
+  void setError(std::exception_ptr error) override {
```
duanmeng, line 712: nit: `const std::exception_ptr&` to align with the Task::setError

PR #16388, opened 2026-02-13
refactor: Extract operator type strings into constants

jainxrohit:
Summary:
Replace hardcoded operator type string literals scattered across operator constructors and comparison sites with centralized `static constexpr const char*` constants in a new `OperatorType.h` header.

Test-only operators (e.g., "Pauser", "Throw") are intentionally excluded and can be addressed in a follow-up.

Differential Revision: D93187082


PingLiuPing (approved):
Thanks.

PR #16389, opened 2026-02-13
fix(parquet): Max target file size not working

PingLiuPing:
Previously, writer rotation did not behave as expected when max-target-file-size was set to a value smaller than the default flush policy threshold.
Parquet only updates write stats on row group flush, so small target sizes were ignored.
For example when max-target-file-size < kDefaultBytesInRowGroup, we could not generating files close to the expected size.
We now cap Parquet’s row group byte threshold so stats advance during writes and rotation can happen.

And fixes a side effect where during writer rotation it could leave an empty file (for parquet it is an empty file with 0B size, for dwrf, it is a non-empty header only file).
The root cause is that prior to introducing writer rotation, writers were created strictly in response to actual input. After enabling rotation, closing a full writer would immediately create a new writer, regardless of whether more input was available. If no additional data arrived, it creates an empty file.
This fix makes the rotation lazy: after closing a full file it only advance the file sequence number but wait to open the next writer until new input arrives, which avoids creating an empty file.

### DIFF
--- a/velox/connectors/hive/HiveDataSink.cpp
+++ b/velox/connectors/hive/HiveDataSink.cpp
@@ -618,6 +618,10 @@ void HiveDataSink::write(size_t index, RowVectorPtr input) {
   WRITER_NON_RECLAIMABLE_SECTION_GUARD(index);
   auto dataInput = makeDataInput(dataChannels_, input);
 
+  if (writers_[index] == nullptr) {
+    writers_[index] = createWriterForIndex(index);
+  }
+
   writers_[index]->write(dataInput);
   writerInfo_[index]->inputSizeInBytes += dataInput->estimateFlatSize();
   writerInfo_[index]->numWrittenRows += dataInput->size();
@@ -681,7 +685,6 @@ void HiveDataSink::rotateWriter(size_t index) {
   VELOX_CHECK_LT(index, writerInfo_.size());
 
   auto& info = writerInfo_[index];
-  const auto& originalParams = info->writerParameters;
 
   // Close the writer first to flush all data including footer.
   writers_[index]->close();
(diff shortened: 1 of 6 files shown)


FelixYBW:
is it related to #15751 fix?

PingLiuPing:
> is it related to #15751 fix?

@FelixYBW No, this PR fixes a scenario that when `max-target-file-size < kDefaultBytesInRowGroup`, the writer rotation does not working and hence it cannot generate a file that close to max-target-file-size.

PingLiuPing:
@prashantgolash Could you help take a look at this part? 

> And fixes a side effect where during writer rotation it could leave an empty file (for parquet it is an empty file with 0B size, for dwrf, it is a non-empty header only file).

However, dwrf reader is able to read "non-empty header only file" without error, is this behavior (empty file) intentional by design?

prashantgolash:
> @prashantgolash Could you help take a look at this part?
> 
> > And fixes a side effect where during writer rotation it could leave an empty file (for parquet it is an empty file with 0B size, for dwrf, it is a non-empty header only file).
> 
> However, dwrf reader is able to read "non-empty header only file" without error, is this behavior (empty file) intentional by design?

I earlier thought about this but it worked for DWRF as you find and I realized this is corner case. Looks like for parquet we may need to fix this. Thanks for making the change. I will look into the diff in more detail by tomorrow EOD

velox/connectors/hive/HiveDataSink.cpp
prashantgolash, line 450: For case when `ensureFiles` is true, we will unconditionally create Writer. I think writer close should write footer file but that is not explicit from Writer contract. What will be behavior with parquet writer? Do we need to check for this also in finalizeWriter path to make sure this is explicitly handled.

The property intends to have files. Per comment
```
  // When this option is set the HiveDataSink will always write a file even
      // if there's no data. This is useful when the table is bucketed, but the
      // engine handles ensuring a 1 to 1 mapping from task to bucket.
```

prashantgolash (commented):
Lgtm overall, have a comment around ensureFiles property

velox/connectors/hive/HiveDataSink.cpp
prashantgolash: Also I think we need to rebase this on your other diff?

velox/connectors/hive/HiveDataSink.cpp
```diff
@@ -1049,28 +1032,10 @@ uint32_t HiveDataSink::appendWriter(const HiveWriterId& id) {
           std::move(writerPool),
           std::move(sinkPool),
           std::move(sortPool)));
-  // Set current file names for the initial file (sequence 0).
-  auto& newInfo = writerInfo_.back();
-  newInfo->currentWriteFileName = newInfo->writerParameters.writeFileName();
```
Yuhta, line 1053: Does this change the file name? We don't need sequence number in this case

PingLiuPing: @Yuhta Thanks for the comment.
When sequence number is 0, `makeSequencedFileName` returns original file name.

https://github.com/facebookincubator/velox/blob/fe064ad1bb0ee1912b3ebebb3b4685c491da20be/velox/connectors/hive/HiveDataSink.cpp#L48-L53

velox/connectors/hive/HiveDataSink.cpp
PingLiuPing: Thanks @prashantgolash 

> What will be behavior with parquet writer?

For parquet, it will generate an empty file (0 bytes).

>  Do we need to check for this also in finalizeWriter path to make sure this is explicitly handled.

I think I understand your point here, I'm trying to figure out if this is valid requirement for parquet. I have a fix for this to make it align with dwrf behaviour, basically I'm passing `ensureFiles` to writer option and during parquet writer close(), check this option together with other states and then write the metadata to parquet file.

How about land this PR first and I'll submit another PR to fix this if necessary.

PingLiuPing: Opened an issue to track this https://github.com/facebookincubator/velox/issues/16527

PingLiuPing:
@Yuhta Could you help to take another look at your convenience? Thank you.

PR #16390, opened 2026-02-13
fix: Allow FlatMapVector to generate nested flat map values

peterenescu:
Summary: Allow nested flat maps to copy and preserve encodings. Prior to this change, copy of `MAP<type1, MAP<type2, type3>>` would always generate an inner map of type `MAP`, which may break copying into a `FLAT_MAP`.

Differential Revision: D93186699

### DIFF
--- a/velox/vector/FlatMapVector.cpp
+++ b/velox/vector/FlatMapVector.cpp
@@ -546,7 +546,8 @@ void FlatMapVector::copyRanges(
       // Then we allocate a new key values vector and in map buffer.
       inMapsAt(channel, true) =
           AlignedBuffer::allocate<bool>(size(), pool(), false);
-      mapValues_.back() = BaseVector::create(valueType(), size(), pool());
+      mapValues_.back() = BaseVector::createEmptyLike(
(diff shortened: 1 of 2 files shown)


velox/vector/FlatMapVector.cpp
```diff
@@ -546,7 +546,23 @@ void FlatMapVector::copyRanges(
       // Then we allocate a new key values vector and in map buffer.
       inMapsAt(channel, true) =
           AlignedBuffer::allocate<bool>(size(), pool(), false);
-      mapValues_.back() = BaseVector::create(valueType(), size(), pool());
```
pedroerp, line 549: @peterenescu wouldn't it be better to better to use the `createLike()` API here so this logic is encapsulated?

peterenescu: @pedroerp I think we should shelve `createLike` until we have a little more clarity over what we want that API to look like (retain all encodings, or just flat map and if so is that the right name or entry file for it). For now, I think this is a potential bug that shouldn't be blocked by that discussion

PR #16400, opened 2026-02-15
feat: Resubmit add max scan rows limit for index lookup

xiaoxmeng:
Summary:
X-link: https://github.com/facebookincubator/nimble/pull/492

Original commit changeset: 5846b9ddaf6c

Original Phabricator Diff: D93313029

Differential Revision: D93348953


HuamengJiang (approved):
Resubmission diff.

PR #16401, opened 2026-02-15
feat: Make barrier processing thread-safe under multithread execution mode

xiaoxmeng:
Summary:
This diff fixes a data race detected by ThreadSanitizer (TSAN) in the barrier processing code under multi-threaded execution mode.

The race condition occurs between:
- **Write**: `Driver::startBarrier()` setting `barrier_ = BarrierState{}`
- **Read**: `Driver::hasBarrier()` (via `isDraining()`) reading `barrier_.drainingOpId.has_value()`

These accesses can happen concurrently from different driver threads.

**Fix approach:**
1. Added an atomic flag `hasBarrier_` to track whether barrier processing is active
2. Changed `hasBarrier()` to read from the atomic flag with `memory_order_acquire`
3. Updated `startBarrier()` to set the atomic flag with `memory_order_release` after initializing `barrier_`
4. Updated `finishBarrier()` to clear the atomic flag before resetting `barrier_`
5. Removed `std::optional` wrapper from `barrier_` since the atomic flag now tracks the active state. Note that `barrier_` is only meaningful when `hasBarrier_` is true.
6. Added `waitForAllTasksToBeDeleted()` in `barrierAfterNoMoreSplits` test to ensure all driver threads complete before the test iteration ends

The acquire-release memory ordering ensures proper synchronization: any thread that reads `hasBarrier_` as `true` is guaranteed to see the fully initialized `barrier_` state.

Differential Revision: D93355327

### DIFF
--- a/velox/exec/Cursor.cpp
+++ b/velox/exec/Cursor.cpp
@@ -18,8 +18,11 @@
 
 #include <folly/system/HardwareConcurrency.h>
 #include <filesystem>
+#include <optional>
 
 #include "velox/common/file/FileSystems.h"
+#include "velox/common/future/VeloxPromise.h"
+#include "velox/exec/BlockingReason.h"
 #include "velox/vector/EncodedVectorCopy.h"
 
 namespace facebook::velox::exec {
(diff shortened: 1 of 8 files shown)


duanmeng (approved):
@xiaoxmeng LGTM, thanks for your help.

PR #16353, opened 2026-02-12
feat: Add decimal remainder and checked_remainder for Spark arithmetic

n0r0shi:
Closes #16354 

## Summary

  - Add `remainder` and `checked_remainder` for decimal types to support
    Spark's `%` operator on decimals. The non-checked variant returns null
    on overflow or division by zero, while the checked variant throws for
    Spark ANSI mode.
  - Uses int128 fast path with `__builtin_mul_overflow` for rescaling,
    falling back to int256 only when the rescaled operands overflow int128.
  - Both `allowPrecisionLoss` and `denyPrecisionLoss` variants registered.

  ## Test plan

  - [x] Normal cases: all 4 input type combinations (short/short, long/long,
        short/long, long/short)
  - [x] Division by zero returns null (non-checked) / throws (checked)
  - [x] Negative operand combinations
  - [x] Remainder with different scales (rescaling required)
  - [x] Overflow detection via precision range check

PR #10083, opened 2024-06-06
Add checked_add, checked_divide, checked_multiply, checked_subtract Spark functions

zhli1142015:
Implement checked_add, checked_divide, checked_multiply, checked_subtract Spark 
functions with non-throwing Simple Functions.

These functions are used to implment `try_add`, `try_subtract`, `try_multiply`, 
`try_divide` functions. `try_XX` functions are `RuntimeReplaceable` which are 
converted to [`try(XX(failOnError=true))`](https://github.com/apache/spark/blob/branch-3.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/TryEval.scala#L81C23-L81C69) during planning. For example `try_add` 
is converted to  `try(add(failOnError=true))`. We add `checked_xx` functions in velox 
to implement `xx(failOnError=true)`.

try_add: https://spark.apache.org/docs/latest/api/sql/#try_add
try_subtract: https://spark.apache.org/docs/latest/api/sql/#try_subtract
try_multiply: https://spark.apache.org/docs/latest/api/sql/#try_multiply
try_divide: https://spark.apache.org/docs/latest/api/sql/#try_divide

Overflow check for types in Spark: https://github.com/apache/spark/blob/94912920b0e92c9470bdb27a409799d0fe48ff69/sql/catalyst/src/main/scala/org/apache/spark/sql/types/numerics.scala#L35C5-L35C18

PR #16403, opened 2026-02-15
Add checked_remainder for integer and floating-point types

n0r0shi:
## Summary
  - Add `checked_remainder` for integer and floating-point types that throws on division by zero, matching Spark's ANSI mode behavior for the `%` operator
  - The decimal version was added in #16353; this adds the integer and floating-point counterpart
  - Unlike `checked_divide`, remainder does not overflow on `INT_MIN % -1` (result is 0), so only division by zero is checked

  ## Test plan
  - Unit tests were added in `ArithmeticTest.checkedRemainder`

Closing https://github.com/facebookincubator/velox/issues/16404

### DIFF
--- a/velox/functions/sparksql/Arithmetic.h
+++ b/velox/functions/sparksql/Arithmetic.h
@@ -666,4 +666,40 @@ struct CheckedIntegralDivideFunction {
   }
 };
 
+template <typename TExec>
+struct CheckedRemainderFunction {
+  template <
+      typename TInput,
+      typename std::enable_if_t<!std::is_floating_point_v<TInput>, int> = 0>
+  FOLLY_ALWAYS_INLINE Status
+  call(TInput& result, const TInput a, const TInput b) {
+    VELOX_USER_RETURN_EQ(b, 0, "Division by zero");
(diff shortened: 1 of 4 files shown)


velox/functions/sparksql/Arithmetic.h
```diff
@@ -666,4 +666,40 @@ struct CheckedIntegralDivideFunction {
   }
 };
 
+template <typename TExec>
+struct CheckedRemainderFunction {
```
rui-mo, line 670: Hi @n0r0shi, just a note regarding ANSI support: the standard approach would be to control the behavior via query config, see https://github.com/facebookincubator/velox/pull/14346. Functions like checked_add were specifically introduced to support [try(XX(failOnError=true))](https://github.com/apache/spark/blob/branch-3.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/TryEval.scala#L81C23-L81C69), see https://github.com/facebookincubator/velox/pull/10083. This means we don’t need to add a checked version for every arithmetic function if the goal is only to support ANSI behavior. cc: @zhli1142015

n0r0shi:
Superseded by #17885 after rebase + config-based refactor per review feedback.

PR #16408, opened 2026-02-16
perf(dwio): Improve rowLoop performance to recognize more dense cases

Yuhta:
Summary:
In the current code when we do the dense loop on `rows` in `rowLoop`,
we stops earlier than necessary.  This results in forwarding most of the work to
the slow `sparseN` calls in case `begin` is not 0, and degrades the performance.
Fix this by considering `begin` when doing dense loop.

Differential Revision: D93424377

### DIFF
--- a/velox/common/base/BitUtil.h
+++ b/velox/common/base/BitUtil.h
@@ -753,7 +753,7 @@ inline uint64_t nextPowerOfTwo(uint64_t size) {
   return 2 * lower;
 }
 
-inline bool isPowerOfTwo(uint64_t size) {
+constexpr bool isPowerOfTwo(uint64_t size) {
   return (size & (size - 1)) == 0;
 }
 
--- a/velox/dwio/common/StreamUtil.h
+++ b/velox/dwio/common/StreamUtil.h
@@ -76,19 +76,19 @@ inline bool isDense(const T* values, int32_t size) {
   return (values[size - 1] - values[0] == size - 1);
 }
 
-template <typename Dense, typename Sparse, typename SparseN>
+template <int kStep, typename Dense, typename Sparse, typename SparseN>
 void rowLoop(
     const int32_t* rows,
     int32_t begin,
     int32_t end,
-    int32_t step,
     Dense dense,
     Sparse sparse,
     SparseN sparseN) {
+  static_assert(bits::isPowerOfTwo(kStep));
   int32_t i = begin;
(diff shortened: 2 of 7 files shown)


velox/common/base/BitUtil.h
```diff
@@ -753,7 +753,7 @@ inline uint64_t nextPowerOfTwo(uint64_t size) {
   return 2 * lower;
 }
 
-inline bool isPowerOfTwo(uint64_t size) {
+constexpr bool isPowerOfTwo(uint64_t size) {
```
xiaoxmeng, line 756: How does this help? Thanks!

velox/dwio/common/StreamUtil.h
```diff
@@ -76,19 +76,19 @@ inline bool isDense(const T* values, int32_t size) {
...
-    int32_t step,
     Dense dense,
     Sparse sparse,
     SparseN sparseN) {
+  static_assert(bits::isPowerOfTwo(kStep));
   int32_t i = begin;
-  auto firstPartial = (end - begin) & ~(step - 1);
-  for (; i < firstPartial; i += step) {
-    if (isDense(&rows[i], step)) {
+  auto firstPartial = (end - begin) & ~(kStep - 1);
```
xiaoxmeng:
```
nit: const auto firstPartial
```

velox/dwio/common/StreamUtil.h
```diff
@@ -76,19 +76,19 @@ inline bool isDense(const T* values, int32_t size) {
   return (values[size - 1] - values[0] == size - 1);
 }
 
-template <typename Dense, typename Sparse, typename SparseN>
+template <int kStep, typename Dense, typename Sparse, typename SparseN>
```
xiaoxmeng, line 79:
```
s/kStep/Step/?
```

xiaoxmeng (approved):
@Yuhta thanks for the optimizations % nits.

velox/common/base/BitUtil.h
Yuhta: This change is done so that I can use it in a `static_assert`

velox/dwio/common/StreamUtil.h
Yuhta: The naming convention for compile time constant values are starting with `k`, this is a `int` value not a type so this naming.

PR #16413, opened 2026-02-17
feat: Add conversion from microseconds to ISO-8601 string

PingLiuPing:
Add a helper function to TimeType to convert microseconds to compact iso8601 string.

### DIFF
--- a/velox/type/Type.cpp
+++ b/velox/type/Type.cpp
@@ -1603,6 +1603,33 @@ int64_t TimeType::valueToTime(
   return adjustedTime;
 }
 
+// static
+std::string TimeType::toCompactIso8601(int64_t microseconds) {
+  // Ensure the value is within valid TIME range.
+  VELOX_USER_CHECK(
+      !(microseconds < 0 || microseconds >= (util::kMicrosPerHour * 24)));
+
+  int64_t hours = microseconds / util::kMicrosPerHour;
+  microseconds %= util::kMicrosPerHour;
+  int64_t minutes = microseconds / util::kMicrosPerMinute;
+  microseconds %= util::kMicrosPerMinute;
+  int64_t seconds = microseconds / util::kMicrosPerSec;
+  int64_t remainingMicros = microseconds % util::kMicrosPerSec;
+
+  std::string result = fmt::format("{:02d}:{:02d}", hours, minutes);
+  if (seconds > 0 || remainingMicros > 0) {
+    result += fmt::format(":{:02d}", seconds);
+    if (remainingMicros > 0) {
+      if (remainingMicros % 1000 == 0) {
+        result += fmt::format(".{:03d}", remainingMicros / 1000);
+      } else {
+        result += fmt::format(".{:06d}", remainingMicros);
+      }
+    }
+  }
+  return result;
+}
+
 std::string stringifyTruncatedElementList(
     size_t size,
     const std::function<void(std::stringstream&, size_t)>& stringifyElement,
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -1622,6 +1622,32 @@ class TimeType final : public BigintType {
     return true;
   }
 
+  /// Converts microseconds since midnight to a compact ISO-8601 time string by
+  /// following rules:
+  /// - Base format is HH:MM.
+  /// - Seconds are included only if seconds > 0 or fractional seconds > 0.
+  /// - Fractional seconds are included only if microseconds > 0.
+  /// - Fractional precision is 3 digits (milliseconds) if microseconds are
+  ///   divisible by 1000, otherwise 6 digits (microseconds).
+  /// The output will be one of the following ISO-8601 formats:
+  /// - HH:mm
+  /// - HH:mm:ss
+  /// - HH:mm:ss.SSS
+  /// - HH:mm:ss.SSSSSS
+  /// Examples (microseconds -> string):
+  /// - 0 -> "00:00"
+  /// - 1 -> "00:00:00.000001"
+  /// - 100'000 -> "00:00:00.100"
+  /// - 1'000'000 -> "00:00:01"
+  /// - 1'000 -> "00:00:00.001"
+  /// - 38'000 -> "00:00:00.038"
+  /// - 38'001 -> "00:00:00.038001"
+  /// - 38'100 -> "00:00:00.038100"
+  /// - 29'288'000'000 -> "08:08:08"
+  /// - 36'775'038'000 -> "10:12:55.038"
+  /// - 86'399'999'999 -> "23:59:59.999999"
+  static std::string toCompactIso8601(int64_t microseconds);
+
   // When casting from TIME to varchar , the resultant varchar will always
   // be 12 bytes long (HH:MM:SS.mmm).
   static const size_t kTimeToVarcharRowSize = 12;
--- a/velox/type/tests/TypeTest.cpp
+++ b/velox/type/tests/TypeTest.cpp
@@ -18,6 +18,7 @@
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/type/CppToType.h"
 #include "velox/type/SimpleFunctionApi.h"
+#include "velox/type/TimestampConversion.h"
 #include "velox/type/TypeEncodingUtil.h"
 #include "velox/type/tests/utils/CustomTypesForTesting.h"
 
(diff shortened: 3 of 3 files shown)


velox/type/Type.cpp
```diff
@@ -1603,6 +1603,26 @@ int64_t TimeType::valueToTime(
   return adjustedTime;
 }
 
+// static
+std::string TimeType::toIso8601(int64_t microseconds) {
+  constexpr int64_t kMicrosPerSecond = 1'000'000LL;
+  constexpr int64_t kMicrosPerMinute = 60LL * kMicrosPerSecond;
+  constexpr int64_t kMicrosPerHour = 60LL * kMicrosPerMinute;
```
mbasmanova: Don't we have these constants somewhere else already?

velox/type/Type.cpp
```diff
@@ -1603,6 +1603,26 @@ int64_t TimeType::valueToTime(
...
+
+  int64_t hours = microseconds / kMicrosPerHour;
+  microseconds %= kMicrosPerHour;
+  int64_t minutes = microseconds / kMicrosPerMinute;
+  microseconds %= kMicrosPerMinute;
+  int64_t seconds = microseconds / kMicrosPerSecond;
+  int64_t remainingMicros = microseconds % kMicrosPerSecond;
+
+  if (remainingMicros == 0) {
```
mbasmanova: Would you document this logic?

velox/type/Type.h
```diff
@@ -1622,6 +1622,13 @@ class TimeType final : public BigintType {
     return true;
   }
 
+  /// Converts microseconds since midnight to an ISO-8601 time string
+  /// in the format HH:MM:SS.ssssss.
+  /// Trailing zeros in the fractional part are preserved.
```
mbasmanova: >  Trailing zeros in the fractional part are preserved.

This is confusing as sometimes the functions returns nothing after period.

velox/type/tests/TypeTest.cpp
```diff
@@ -1238,3 +1238,33 @@ TEST(TypeTest, time) {
...
+  constexpr int64_t kMicrosPerMinute = 60LL * kMicrosPerSecond;
+  constexpr int64_t kMicrosPerHour = 60LL * kMicrosPerMinute;
+
+  EXPECT_EQ("00:00:00", TimeType::toIso8601(0));
+  EXPECT_EQ("00:00:01", TimeType::toIso8601(kMicrosPerSecond));
+  EXPECT_EQ("00:00:00.000001", TimeType::toIso8601(1));
+  EXPECT_EQ("00:00:01.000002", TimeType::toIso8601(2));
+  EXPECT_EQ(
+      "10:12:55.038000",
+      TimeType::toIso8601(
```
mbasmanova: perhaps, add a helper lambda

```
EXPECT_EQ("10:12:55.038000", toIso8601(10, 12, 55, 38'000));
```

mbasmanova:
@PingLiuPing Curious, where do you plan to use this?

> Example: 36775038000 -> "10:12:55.038000", 29288000000 -> "08:08:08".

Does this follow the standard? Would you add a reference to explain when to add fractional part and when not?

velox/type/tests/TypeTest.cpp
PingLiuPing: Thanks. That's cleaner.

velox/type/Type.cpp
PingLiuPing: Thanks. Found them in `TimestampConversion.h`

velox/type/Type.cpp
PingLiuPing: Thanks. I’ve refined the documentation. I also checked the Iceberg Java implementation and found that it relies on LocalTime.toString(). I’ve aligned our implementation with the Java behavior.

PingLiuPing:
> @PingLiuPing Curious, where do you plan to use this?
> 
> > Example: 36775038000 -> "10:12:55.038000", 29288000000 -> "08:08:08".
> 
> Does this follow the standard? Would you add a reference to explain when to add fractional part and when not?

@mbasmanova  Thank you.
This helper function is needed to support the TIME type as an Iceberg Identity partition column. It is not used at the moment, but once https://github.com/facebookincubator/velox/pull/16217 lands, I plan to submit a follow up PR to add support for TIME as a partition column. without #16217 I cannot add corresponding testcase.

velox/type/tests/TypeTest.cpp
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 #include "velox/type/Type.h"
+
+#include <type/TimestampConversion.h>
```
mbasmanova: remove extra new lines around this include

velox/type/tests/TypeTest.cpp
```diff
@@ -1238,3 +1241,24 @@ TEST(TypeTest, time) {
...
+      [](int64_t hours, int64_t minutes, int64_t seconds, int64_t micros) {
+        return TimeType::toIso8601(
+            hours * util::kMicrosPerHour + minutes * util::kMicrosPerMinute +
+            seconds * util::kMicrosPerSec + micros);
+      };
+
+  EXPECT_EQ("00:00", toIso8601(0, 0, 0, 0));
+  EXPECT_EQ("00:00:00.000001", toIso8601(0, 0, 0, 1));
+  EXPECT_EQ("00:00:00.100", toIso8601(0, 0, 0, 100'000));
+  EXPECT_EQ("00:00:01", toIso8601(0, 0, 1, 0));
```
mbasmanova, line 1254: Is it part of ISO standard to not include .000 here? Can you link / quote relevant parts from the standard?

velox/type/Type.h
```diff
@@ -1622,6 +1622,27 @@ class TimeType final : public BigintType {
     return true;
   }
 
+  /// Converts microseconds since midnight to an ISO-8601 time string.
+  /// Matches Java LocalTime::toString formatting rules:
```
mbasmanova: Iso8601 is a well-defined standard. We should follow that, not "Java LocalTime::toString formatting rules". If we need the latter and it is not the same as Iso8601, then we need to change the name of the API.

PingLiuPing: @mbasmanova Thank you, I understand your point.
I’ve renamed the function to `toCompactIso8601`.
This function may return strings with variable length depending on the input, but each output is a valid ISO-8601 representation of the time.

velox/type/tests/TypeTest.cpp
PingLiuPing: That is not part of the standard. Updated the comment to explain this, this aligns the Java implementation that output the "shortest representation".

velox/type/tests/TypeTest.cpp
PingLiuPing: Thanks.

velox/type/tests/TypeTest.cpp
```diff
@@ -1238,3 +1239,24 @@ TEST(TypeTest, time) {
...
+  EXPECT_EQ("00:00:00.001", toIso8601(0, 0, 0, 1'000));
+  EXPECT_EQ("00:00:00.038", toIso8601(0, 0, 0, 38'000));
+  EXPECT_EQ("00:00:00.038001", toIso8601(0, 0, 0, 38'001));
+  EXPECT_EQ("00:00:00.038100", toIso8601(0, 0, 0, 38'100));
+  EXPECT_EQ("08:08:08", toIso8601(8, 8, 8, 0));
+  EXPECT_EQ("10:12:55.038", toIso8601(10, 12, 55, 38'000));
+  EXPECT_EQ("23:59:59.999999", toIso8601(23, 59, 59, 999'999));
```
kgpai, line 1261: Do we check if hours is always < 24 ? Are values greater than this legal ?

velox/type/Type.cpp
```diff
@@ -1603,6 +1603,29 @@ int64_t TimeType::valueToTime(
   return adjustedTime;
 }
 
+// static
+std::string TimeType::toCompactIso8601(int64_t microseconds) {
```
kgpai, line 1607: Do we need to validate that microseconds less than max value in 24 hours ?

PingLiuPing: @kgpai Thanks for pointing that out, I'll add it.

velox/type/tests/TypeTest.cpp
PingLiuPing: Once we check the input is within 24*kMicrosPerHour, then the output hours should be < 24.

kgpai:
@PingLiuPing This should land soon - Thanks for your patience.

PR #16416, opened 2026-02-17
feat(sparksql): Support RESPECT NULLS for Spark collect_set aggregate function

yaooqinn:
## Summary

Add a new field `ignoreNulls` to control null handling in the Spark
`collect_set` aggregate function, following the same pattern as `collect_list`
(#16328). When set to `false` (RESPECT NULLS), null values are included in the 
result set instead of being skipped. This aligns with Spark 4.2+ RESPECT NULLS 
/ IGNORE NULLS syntax ([SPARK-55533](https://issues.apache.org/jira/browse/SPARK-55533)).


velox/core/QueryConfig.h
```diff
@@ -519,6 +519,12 @@ class QueryConfig {
   static constexpr const char* kSparkCollectListIgnoreNulls =
       "spark.collect_list.ignore_nulls";
 
+  /// If true, collect_set aggregate function will ignore nulls in the input.
+  /// Defaults to true to match Spark's default behavior. Set to false to
+  /// include nulls (RESPECT NULLS). Introduced in Spark 4.2 (SPARK-55533).
+  static constexpr const char* kSparkCollectSetIgnoreNulls =
+      "spark.collect_set.ignore_nulls";
```
rui-mo: Hi @yaooqinn, does Spark provide centralized configuration rather than function-level configuration?

jinchengchenghh: I see in the PR https://github.com/apache/spark/pull/54329/changes, it is semantic level 
df.selectExpr("sort_array(collect_set(b) IGNORE NULLS)"), Seq(Row(Seq(2, 4))))

yaooqinn:
@jinchengchenghh missing tag read-to-merge?

jinchengchenghh:
Please also update the configs.rst

yaooqinn:
Done. Updated `configs.rst` with both `spark.collect_set.ignore_nulls` and `spark.collect_list.ignore_nulls` (which was also missing from a previous PR). Thanks @jinchengchenghh!

velox/docs/configs.rst
```diff
@@ -1163,6 +1158,16 @@ Spark-specific Configuration
      - bool
      - true
      - If true, ignore null fields when generating JSON string. If false, null fields are included with a null value.
+   * - spark.collect_list.ignore_nulls
+     - bool
+     - true
+     - If true, ``collect_list`` aggregate function ignores null inputs. If false, null values
```
jinchengchenghh: update to like  
```
:spark:func:`bloom_filter_agg` function
```

zhli1142015 (approved):

velox/core/QueryConfig.h
zhli1142015: If there are two collect_set expressions in the same query — one that ignores nulls and one that does not — what would the behavior be?

rui-mo: @yaooqinn If this is a parameter in Spark, why would we need a context level configuration? Can we add another field to be consistent with Spark?

```
case class CollectSet(
    child: Expression,
    mutableAggBufferOffset: Int = 0,
    inputAggBufferOffset: Int = 0)
    inputAggBufferOffset: Int = 0,
    ignoreNulls: Boolean = true)
```

rui-mo: How would you use this configuration in Gluten, and how to map Spark’s `ignoreNulls` value to the Velox configuration and update it for different calls?

yaooqinn: This is a valid concern, I missed this. Sorry about that, let me add a field-scope behavior gate.

rui-mo: Thank you for confirming. Please take your time, but let’s pause the import process until this issue is resolved.

yaooqinn:
Thanks @rui-mo @zhli1142015 @jinchengchenghh for the feedback. I've reworked the implementation to use a per-expression constant boolean argument instead of QueryConfig:

- Removed `kSparkCollectSetIgnoreNulls` from `QueryConfig.h`
- Added a 2-arg signature: `collect_set(x, ignoreNulls)` where `ignoreNulls` is a constant boolean
- Default behavior (1-arg): `collect_set(x)` ignores nulls (`ignoreNulls=true`)
- RESPECT NULLS (2-arg): `collect_set(x, false)` includes nulls (`ignoreNulls=false`)
- This is consistent with Spark's per-expression `CollectSet(child, ignoreNulls: Boolean = true)`
- In Gluten, the Substrait plan can simply pass the `ignoreNulls` boolean from the Spark expression as the second argument

This resolves the concern about two `collect_set` expressions in the same query with different null handling — each expression independently controls its behavior via the constant argument.

All 8 CollectSet tests pass.

velox/docs/functions/spark/aggregate.rst
```diff
@@ -65,9 +65,15 @@ General Aggregate Functions
     instead of an empty array.
 
 .. spark:function:: collect_set(x) -> array<[same as x]>
+                    collect_set(x, ignoreNulls) -> array<[same as x]>
```
rui-mo: Do we need to provide the single-arg signature, given that Spark always provides a value for "ignoreNulls"?

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
```diff
@@ -18,93 +18,118 @@
...
+         const std::vector<TypePtr>& argTypes,
+         const TypePtr& resultType,
+         const core::QueryConfig& /*config*/)
           -> std::unique_ptr<exec::Aggregate> {
-        VELOX_CHECK_EQ(argTypes.size(), 1);
-
         const bool isRawInput = exec::isRawInput(step);
+        // For raw input, argTypes contains [T] or [T, BOOLEAN].
+        // For intermediate/final input, argTypes contains [ARRAY(T)].
+        const bool ignoreNulls = isRawInput ? argTypes.size() == 1 : true;
```
rui-mo: This might cause unexpected behavior if the second argument is provided but is false.

yaooqinn:
Addressed review comments:

**Re: single-arg signature (aggregate.rst:68)**: Kept both 1-arg and 2-arg signatures for backward compatibility. The 1-arg form `collect_set(x)` defaults to `ignoreNulls=true`. Spark/Gluten callers will always use the 2-arg form.

**Re: unexpected behavior with constant boolean (CollectSetAggregate.cpp:124)**: Reworked to create a `SparkCollectSetAggregate` wrapper class that reads the actual constant boolean value from `args[1]` at runtime in `addRawInput` / `addSingleGroupRawInput`. This correctly handles both `collect_set(x, true)` (ignores nulls) and `collect_set(x, false)` (respects nulls). The `ignoreNulls_` member is also used for intermediate/final aggregation steps (`addIntermediateResults` / `addSingleGroupIntermediateResults`) to ensure consistency when partial aggregation is abandoned.

Added a new test `explicitIgnoreNullsTrue` to verify `collect_set(c0, true)` correctly ignores nulls. All 9 CollectSet tests pass.

jinchengchenghh:
We only need collect_set(T, boolean) -> array(T)  -- explicit ignoreNulls flag since the latter will always send by Gluten

yaooqinn:
> We only need collect_set(T, boolean) -> array(T) -- explicit ignoreNulls flag since the latter will always send by Gluten

OK, I was thinking the velox as an independent project to keep the backward compatibility

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
```diff
@@ -18,19 +18,256 @@
...
+            *baseArray,
+            decodedIndex,
+            SBase::decodedElements_,
+            SBase::allocator_);
+      }
+    });
+  }
+
+ private:
+  bool ignoreNulls_;
```
rui-mo: Would you mind clarifying the relationship between `ignoreNulls_ ` (which is always set to false during registration) and `args[1]->as<ConstantVector<bool>>()->valueAt(0)`? I’m not sure why we need `ignoreNulls_`. Could we rely solely on the runtime argument instead?

rui-mo: The current design may be a bit confusing. Would you mind considering avoiding additional workarounds? It might be better to implement this in a more straightforward way.

yaooqinn: Good question. We cannot rely solely on `args[1]` because it is only available in `addRawInput`/`addSingleGroupRawInput`. In the intermediate/final aggregation steps (`addIntermediateResults`, `addSingleGroupIntermediateResults`, `toIntermediate`), the `args` vector contains only the intermediate representation `array(T)` -- there is no boolean constant.

The `ignoreNulls_` member acts as a bridge:
1. Factory sets `ignoreNulls_=false` (conservative default, since the factory only receives `argTypes` but not constant values).
2. `addRawInput` updates `ignoreNulls_` from `args[1]` -- caching the actual user-provided value.
3. `toIntermediate()` reads `ignoreNulls_` to decide whether to emit empty arrays for null elements (preventing null leakage in abandon-partial-agg mode).
4. `addIntermediateResults` reads `ignoreNulls_` to filter nulls when merging intermediate results.

In partial-to-final mode, the final node never calls `addRawInput`, so it uses the factory default `false` -- which is safe because the partial step already filtered nulls from the intermediate data.

rui-mo: The issue seems to be that we’re unable to initialize the aggregate using a constant argument. Do you think revamping the initialize method would help? For example, in the simple function, a is valid if a constant is provided.


```
  template <typename A>
  void initialize(
      const std::vector<TypePtr>& inputTypes,
      const core::QueryConfig& /*config*/,
      A* /*a*/) {
    inputType_ = inputTypes[0];
  }
```

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
```diff
@@ -18,19 +18,261 @@
...
+      return std::make_unique<SparkCollectSetAggregate<StringView>>(
+          resultType, ignoreNulls);
+    case TypeKind::ARRAY:
+      [[fallthrough]];
+    case TypeKind::ROW:
+      return std::make_unique<SparkCollectSetAggregate<ComplexType>>(
+          resultType, ignoreNulls);
+    case TypeKind::UNKNOWN:
+      return std::make_unique<SparkCollectSetAggregate<UnknownValue>>(
+          resultType, ignoreNulls);
```
rui-mo: If `ignoreNulls` is always false, do we need to make it a variable for the `SparkCollectSetAggregate`? Can we declare `ignoreNulls_` with a default value `false` instead?

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
rui-mo: Another concern is that, with the current implementation, if there’s a fallback in the plan as shown below and  Spark's `ignoreNulls` is true, would step 4 produce an incorrect result? It seems like it doesn’t have access to the `ignoreNulls` value from the Spark plan.

(1) vanilla partial aggregate -> (2) r2c -> (3) columnar shuffle -> (4) gluten final aggregate

rui-mo:
Another possible approach might be to register two versions of the collect_set aggregate function with different names in Velox, allowing Gluten to select the appropriate one based on the `ignoreNulls` value from Spark.

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
yaooqinn: Thanks for the suggestion! I've created a separate PR #16624 that adds a `Aggregate::setConstantInputs()` virtual hook to the base class. It's called after `Aggregate::create()` in `toAggregateInfo()` with the constant input values, allowing aggregates to read constant arguments at initialization time.

This PR now uses that hook: `SparkCollectSetAggregate::setConstantInputs()` reads the boolean constant directly, replacing the previous runtime-caching approach. The design is much simpler now — no more indirection through `addRawInput`.

yaooqinn: Good catch on the fallback scenario. This is actually safe: when `ignoreNulls=true`, step (1) vanilla Spark partial aggregate already filters null elements from the intermediate data. So by the time step (4) Gluten final aggregate receives the intermediate `array(T)`, there are no null elements in it. The final node uses `ignoreNulls_=false` (default), which calls `addValues` -- but since the intermediate data contains no nulls, the result is correct.

In the reverse case (`ignoreNulls=false`), step (1) keeps nulls in the intermediate data, and step (4) with default `false` also keeps them -- also correct.

The key design invariant is: the default `false` (conservative, keeps nulls) is always safe for intermediate/final nodes because the partial node already enforces the null-filtering contract.

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
yaooqinn: Addressed in the latest push -- `ignoreNulls_` now defaults to `false` and is set via the new `setConstantInputs()` hook at initialization time.

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
rui-mo: > The key design invariant is: the default false (conservative, keeps nulls) is always safe for intermediate/final nodes because the partial node already enforces the null-filtering contract.

Thanks, this makes sense to me.

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
```diff
@@ -18,19 +18,250 @@
...
+  using Base = SetAggAggregate<T, false, false, AccumulatorType>;
+  using SBase = SetBaseAggregate<T, false, false, AccumulatorType>;
+
+ public:
+  explicit SparkCollectSetAggregate(const TypePtr& resultType)
+      : Base(resultType) {}
+
+  void setConstantInputs(
+      const std::vector<VectorPtr>& constantInputs) override {
+    if (constantInputs.size() > 1 && constantInputs[1] != nullptr) {
```
rui-mo: Can we assert `constantInputs.size()` is 2 because the size of constant inputs should aligns with that of the arguments?

```
  /// Optional constant inputs. The size of this vector matches the size of
  /// 'inputs'. Non-constant inputs have null entries.
  std::vector<VectorPtr> constantInputs;
```

Also assert `constantInputs[1]` is valid, ensured by the signature `.argumentType("T")
          .constantArgumentType("boolean")`.

yaooqinn: Done. Added `VELOX_CHECK_NOT_NULL(constantInputs[1])` assertion. Note that in the intermediate/final step, `constantInputs` has fewer entries (only the intermediate column), so we early-return when `size < 2` rather than asserting size == 2.

rui-mo:
To prevent the Spark aggregate fuzzer from failing, can you please use `VELOX_USER_CHECK_NOT_NULL`? User exceptions are permitted by the fuzzer, whereas runtime exceptions are not.

```
E20260304 09:16:25.856997   151 ExceptionTracer.cpp:227] exception stack complete
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
  what():  Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Retriable: False
Expression: constantInputs[1] != nullptr
Function: setConstantInputs
File: /__w/velox/velox/velox/velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
```

velox/functions/sparksql/aggregates/CollectSetAggregate.cpp
yaooqinn: Updated -- reverted to null-safe check instead of `VELOX_CHECK_NOT_NULL`. The Spark Aggregate Fuzzer showed that companion functions (e.g., `merge_collect_set`) may call `setConstantInputs` with `constantInputs[1] == nullptr` when the second argument is not a constant in that context. The null check handles all cases safely.

peterenescu:
Noticed agg fuzzer is still failing, let me know if this is still WIP

rui-mo:
@peterenescu We will need to fix the fuzzer before importing, thanks!

rui-mo:
@yaooqinn I ran the Spark aggregate fuzzer test locally and noticed it fails because Velox produces collect_set with two-column inputs, whereas Spark’s collect_set only supports a single column. The fuzzer test fails after Spark throws Analysis exception.

Do you think it would make sense to add back the single-argument registration and skip the two-argument test in the fuzzer? This should fix the failure while still ensuring collect_set is tested against Spark.

rui-mo:
@yaooqinn Just a gentle reminder to address the fuzzer issue so we can land this change, thanks!

yaooqinn:
Added back the single-arg `collect_set(T)` signature (defaults to `ignoreNulls=true`). Now both signatures are registered:

- `collect_set(T) -> array(T)` — Spark default (ignoreNulls=true)
- `collect_set(T, boolean) -> array(T)` — explicit ignoreNulls flag

This should fix the fuzzer issue since Spark's `collect_set` will match the single-arg signature. The 2-arg form is only used when Gluten explicitly passes the ignoreNulls constant.

rui-mo:
@yaooqinn Thanks. You might also need to exclude the two-arg signature in SparkAggregationFuzzerTest.cpp , like `"intersection_cardinality(khyperloglog,khyperloglog) -> bigint"` to avoid that signature being picked for test.

velox/docs/functions/spark/aggregate.rst
```diff
@@ -64,10 +64,15 @@ General Aggregate Functions
     behavior). In this mode, an all-null input produces an array of nulls
     instead of an empty array.
 
-.. spark:function:: collect_set(x) -> array<[same as x]>
+.. spark:function:: collect_set(x, ignoreNulls) -> array<[same as x]>
```
rui-mo: nit: use collect_set(x[ , ignoreNulls]) to make `ignoreNulls` optional.

yaooqinn:
Addressed:

1. **Added back single-arg signature** `collect_set(T)` — fuzzer can now test this against Spark
2. **Added `collect_set` to `skipFunctions` in SparkAggregationFuzzerTest.cpp** — the 2-arg `collect_set(T, boolean)` signature isn't supported by Spark's SQL parser, so skipping the Spark verification. The Velox-only agg fuzzer still tests both signatures.
3. **Doc nit**: Updated to `collect_set(x [, ignoreNulls])` with optional bracket notation.

@rui-mo

velox/functions/sparksql/fuzzer/SparkAggregationFuzzerTest.cpp
```diff
@@ -75,6 +75,9 @@ int main(int argc, char** argv) {
   // following names.
   std::unordered_set<std::string> skipFunctions = {
       "bloom_filter_agg",
+      // Velox registers a 2-arg collect_set(T, boolean) signature that Spark
+      // doesn't support. The fuzzer may pick this signature and fail.
+      "collect_set",
```
rui-mo, line 80: @yaooqinn This skips both signatures. I intended to skip only the two-argument one.

`collect_set(T,boolean) -> array(T)`

yaooqinn:
The aggregation fuzzer doesn't support per-signature skip (unlike the expression fuzzer which has string-based signature exclusion). The `skipFunctions` set operates at function-name level, skipping all signatures.

Two options:
1. Keep `collect_set` in `skipFunctions` (current) — skips both sigs from Spark comparison but existing unit tests cover both forms
2. Add per-signature skip support to the agg fuzzer (framework change, separate PR)

I'll go with option 1 for now. The 1-arg form is thoroughly tested in `CollectSetAggregateTest` (9 tests). Happy to add per-signature skip as a follow-up if needed.

@rui-mo

rui-mo:
Thanks @yaooqinn. I took another look and confirmed that per-signature skip isn’t currently supported.
Also, do you plan to address the similar issue with `kSparkCollectListIgnoreNulls`? It seems that this configuration should be removed for the same reasons mentioned above.

yaooqinn:
Thanks @rui-mo for the thorough review and confirmation on the per-signature skip! 🙏

Created #16839 to track the `collect_list` follow-up — will apply the same pattern (remove `kSparkCollectListIgnoreNulls` config, add constant boolean argument).

Issue #16352, opened 2026-02-12
Add barrier support under multi-threaded execution mode


xiaoxmeng:
### Description

Currently, the barrier support only works under single-threaded execution mode. For some use case like streaming, it requires the multi-threaded execution mode support. To support this, we need to handle the case that there are multiple operators instantiated from the same plan node to ensure that each operator gets the barrier signal exactly once.

cc @Yuhta @mbasmanova @pedroerp @duanmeng

Yuhta:
Do we have a plan or design how we are going to do it?  The node level states in task are quite implicit today, because it is organized by driver -> operator.  I imagine we would need a struct to represent state of each node (not operator) in task.

xiaoxmeng:
This work has finished and used in Meta internal shadow env for streaming use case.

PR #16419, opened 2026-02-17
feat: Make 'QueryAssertions::readCursor' can hold and check future returned by addSplit

duanmeng:
Make 'QueryAssertions::readCursor' can hold and check future returned by addSplit.
Part of #16352

### DIFF
--- a/velox/exec/tests/utils/QueryAssertions.cpp
+++ b/velox/exec/tests/utils/QueryAssertions.cpp
@@ -1360,13 +1360,31 @@ std::pair<std::unique_ptr<TaskCursor>, std::vector<RowVectorPtr>> readCursor(
     const CursorParameters& params,
     std::function<void(TaskCursor*)> addSplits,
     uint64_t maxWaitMicros) {
+  return readCursorAsync(
+      params,
+      [addSplitsVoid = std::move(addSplits)](TaskCursor* cursor) {
+        addSplitsVoid(cursor);
+        return ContinueFuture::makeEmpty();
+      },
+      maxWaitMicros);
+}
+
+std::pair<std::unique_ptr<TaskCursor>, std::vector<RowVectorPtr>>
+readCursorAsync(
+    const CursorParameters& params,
+    std::function<ContinueFuture(TaskCursor*)> addSplits,
+    uint64_t maxWaitMicros) {
   auto cursor = TaskCursor::create(params);
   // 'result' borrows memory from cursor so the life cycle must be shorter.
   std::vector<RowVectorPtr> result;
   auto* task = cursor->task().get();
   cursor->start();
+  auto future = ContinueFuture::makeEmpty();
   while (!cursor->noMoreSplits()) {
-    addSplits(cursor.get());
+    if (future.valid()) {
+      future.wait();
+    }
+    future = addSplits(cursor.get());
     while (cursor->moveNext()) {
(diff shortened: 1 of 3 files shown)


velox/connectors/hive/HiveIndexReader.cpp
```diff
@@ -23,6 +23,7 @@
 #include "velox/connectors/hive/HiveConnectorUtil.h"
 #include "velox/connectors/hive/TableHandle.h"
 #include "velox/dwio/common/ReaderFactory.h"
+#include "velox/serializers/KeyEncoder.h"
```
pedroerp: is this an unrelated change?

velox/exec/tests/utils/QueryAssertions.cpp
```diff
@@ -1358,15 +1361,19 @@ bool testingMaybeTriggerAbort(exec::Task* task) {
 
 std::pair<std::unique_ptr<TaskCursor>, std::vector<RowVectorPtr>> readCursor(
     const CursorParameters& params,
-    std::function<void(TaskCursor*)> addSplits,
+    std::function<ContinueFuture(TaskCursor*)> addSplits,
```
pedroerp, line 1375: instead of adding boilerplate to every callsite, maybe we could provide two signatures of this function (one taking a lambda returning void, and one returning a future), and make the former call the latter by adding a lambda wrapping around the other lambda, returning an empty future?

duanmeng: > instead of adding boilerplate to every callsite, maybe we could provide two signatures of this function (one taking a lambda returning void, and one returning a future), and make the former call the latter by adding a lambda wrapping around the other lambda, returning an empty future?

@pedroerp Thanks for your suggestion. If we provide two overloads like:
```C++
readCursor(..., std::function<void(TaskCursor*)>)
readCursor(..., std::function<ContinueFuture(TaskCursor*)>)
```
Then a lambda that returns `ContinueFuture` is also convertible to `std::function<void(...)>` (the return value is simply discarded). As a result, overload resolution cause ambiguous call error. 

I've added a new API `readCursorAsync(..., std::function<ContinueFuture(TaskCursor*)>)` and wrapped it in readCursor, so that we can reuse the code and avoid ambiguous errors.

velox/connectors/hive/HiveIndexReader.cpp
duanmeng: This is to fix the linker error,
```C++
Undefined symbols for architecture arm64:
  "facebook::velox::serializer::IndexBounds::set(facebook::velox::serializer::IndexBound, facebook::velox::serializer::IndexBound)", referenced from:
      facebook::velox::connector::hive::HiveIndexReader::buildRequestIndexBounds(std::__1::shared_ptr<facebook::velox::RowVector> const&) in HiveIndexReader.cpp.o
ld: symbol(s) not found for architecture arm64
```
Should I fix it in a separate PR?

velox/exec/tests/utils/QueryAssertions.cpp
pedroerp: Ouch, didn't know type resolution would ignore return types even for std::function. TIL!

Takes for looking into this.

pedroerp (approved):
Thank you @duanmeng

PR #16423, opened 2026-02-17
feat(cudf): Add cuDF join fuzzer with CI integration

shrshi:
This PR introduces fuzz testing for GPU-accelerated hash join operations in velox-cudf. The fuzzer continuously generates random join operations, executes them on GPU via cuDF, and verifies correctness against DuckDB as a reference database.

### Why refactor into JoinFuzzerBase instead of adding flags to the existing fuzzer?

We considered two approaches for adding cuDF join fuzzing:

**Option A: Configuration flags in existing fuzzer**
Add flags like `--use-cudf`, `--cudf-types`, `--cudf-join-types` to the existing `velox_join_fuzzer` binary.

**Option B: Base class with executor-specific subclasses**
Extract shared logic into `JoinFuzzerBase`, let each executor (CPU, cuDF) implement its own fuzzer.

We chose **Option B** for two reasons:

1. **Avoids configuration explosion.** A flag-based approach requires validating every combination: cuDF-supported types × cuDF-supported join types × cuDF-specific behaviors (no spilling, no merge joins, no grouped execution). Each executor has different constraints, and encoding all of them as flags creates a maintenance burden. With inheritance, each subclass simply declares what it supports - invalid combinations are impossible by construction.

2. **Avoids conditional dependencies.** A unified binary would need to link against cuDF libraries and call `registerCudf()` conditionally. This pollutes the core Velox build with optional GPU dependencies. With separate binaries, `velox_join_fuzzer` has zero cuDF dependencies, and `velox_cudf_join_fuzzer` links only what it needs.

The refactoring is minimal - subclasses override just three methods:
```cpp
std::vector<core::JoinType> getSupportedJoinTypes() const override;
std::vector<TypePtr> getSupportedTypes() const override;
void verify(core::JoinType joinType) override;
```

### Changes

**1. JoinFuzzerBase refactoring (`velox/exec/fuzzer/`)**
- Extracted common join fuzzing logic into `JoinFuzzerBase` abstract base class
- CPU `JoinFuzzer` now extends `JoinFuzzerBase`
- Added `velox_join_fuzzer_base` CMake library target

**2. cuDF Join Fuzzer (`velox/experimental/cudf/tests/`)**
- `CudfJoinFuzzer.cpp/h` - GPU fuzzer implementation extending `JoinFuzzerBase`
- `CudfJoinFuzzerRunner.cpp` - CLI entry point with configuration flags
- Supports join types: INNER, LEFT, FULL, ANTI, LEFT_SEMI_FILTER, LEFT_SEMI_PROJECT
- Supports scalar types: BOOLEAN, TINYINT, SMALLINT, INTEGER, BIGINT, REAL, DOUBLE, VARCHAR

**3. CI Integration (`.github/workflows/linux-build-base.yml`)**
- New `cudf-join-fuzzer` job that runs after `cudf-tests` passes
- Runs for 15 minutes on GPU runner
- Archives logs on failure for debugging/reproduction


shrshi:
Motivation for the fuzzer refactoring:
1. General extensibility to other operator executors:  Every fuzzer follows the same testing pattern of generating random input vectors, constructing and executing the query plan, and verifying results. The operator-specific infrastructure is depends on the executor, the supported types and algorithms. The `JoinFuzzerBase` customization points illustrate this - the subclasses need to override just three class functions. 
2. Better maintainability over configuration setting based design: We can adopt a configuration-based approach to turn on fuzz-testing in Velox-cuDF. However, this runs into issues like configuration explosion - e.g. cuDF supported types, supported join types, operator specific testing, spilling, etc. - that is hard to validate and maintain. Moreover, such a design will require cuDF registration to happen from the main Velox fuzzer binary causing conditional cuDF dependencies.

velox/experimental/cudf/tests/CudfJoinFuzzerRunner.cpp
```diff
@@ -0,0 +1,135 @@
...
+    "Initial seed for random number generator used to reproduce previous "
+    "results (0 means start with random seed).");
+
+DEFINE_int32(steps, 10, "Number of plans to generate and test.");
+
+DEFINE_int32(
+    duration_sec,
+    0,
+    "For how long it should run (in seconds). If zero, "
+    "it executes exactly --steps iterations and exits.");
```
devavret: There are `steps` and `duration_sec` flag in the CPU join fuzzer too. Can we make these common in the base?

velox/experimental/cudf/tests/CudfJoinFuzzerRunner.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+DEFINE_double(
+    null_ratio,
+    0.1,
+    "Chance of adding a null value in a vector "
+    "(expressed as double from 0 to 1).");
+
+DEFINE_int64(allocator_capacity, 8L << 30, "Allocator capacity in bytes.");
+
+DEFINE_int64(arbitrator_capacity, 6L << 30, "Arbitrator capacity in bytes.");
```
devavret, line 82: We don't need these in cudf execution right?

velox/experimental/cudf/tests/CudfJoinFuzzer.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // DuckDB's SQL conversion.
+    return {
+        core::JoinType::kInner,
+        core::JoinType::kLeft,
+        core::JoinType::kFull,
+        core::JoinType::kAnti,
+        core::JoinType::kLeftSemiFilter,
+        core::JoinType::kLeftSemiProject,
+    };
+  }
```
devavret: nit: is there any way to use the list we have in operator adapter so we don't have a duplicate list we have to maintain.

velox/exec/fuzzer/JoinFuzzer.cpp
```diff
@@ -150,214 +128,30 @@ class JoinFuzzer {
...
-      // output size, reduce the batch size to 10% to control the output size
-      // while still covering this case.
-      input.push_back(
-          vectorFuzzer_.fuzzRow(inputType, FLAGS_batch_size / 10, false));
-    } else {
-      input.push_back(vectorFuzzer_.fuzzInputRow(inputType));
-    }
+  /// Helper to access our extended Stats.
+  Stats& stats() {
+    return static_cast<Stats&>(*stats_);
```
devavret, line 139: nit: why the change to accessor? conversely, why not use the accessor in CudfJoinFuzzer?

devavret (commented):
I only reviewed the cpp stuff. Tagging @bdice for reviewing the CI changes.

devavret:
Requesting reviews from @mbasmanova @pedroerp and @duanmeng  who are listed as fuzzer maintainers but weren't auto assigned.

velox/experimental/cudf/tests/CudfJoinFuzzer.cpp
shrshi: Yep, refactored so that we maintain a single source of truth; the code now does a lookup of the supported types in cuDF hash join implementation.

velox/experimental/cudf/tests/CudfJoinFuzzerRunner.cpp
shrshi: We do need these options in cuDF execution since we are using Velox memory pools for vector generation and DuckDB query execution.

velox/exec/fuzzer/JoinFuzzer.cpp
shrshi: The stats() accessor in JoinFuzzer is necessary since the class derives from the base Stats struct in JoinFuzzerBase and adds `numCrossProduct` member. OTOH, CudfJoinFuzzer does not extend the base class. Nested loop joins are currently not supported in cuDF-Velox, so it makes sense to keep the inheritance-based design here,

shrshi:
Waiting on https://github.com/facebookincubator/velox/pull/17035 to merge to fix CI test failure.

velox/exec/fuzzer/JoinFuzzerBase.cpp
```diff
@@ -0,0 +1,353 @@
...
+  seed(rng_());
+}
+
+int32_t JoinFuzzerBase::randInt(int32_t min, int32_t max) {
+  return boost::random::uniform_int_distribution<int32_t>(min, max)(rng_);
+}
+
+core::JoinType JoinFuzzerBase::pickJoinType() {
+  const auto& joinTypes = getSupportedJoinTypes();
+  const size_t idx = randInt(0, joinTypes.size() - 1);
```
karthikeyann, line 115: If `getSupportedJoinTypes()` is ever empty, `joinTypes.size() - 1` underflows `size_t`, `randInt(0, -1)` is UB, and `joinTypes[idx]` reads OOB. Could we add `VELOX_USER_CHECK(!joinTypes.empty(), "No supported join types")` here, or assert once in `go()` before the loop?

velox/exec/fuzzer/JoinFuzzerBase.cpp
```diff
@@ -0,0 +1,353 @@
...
+  for (int i = 0; i < static_cast<int>(core::JoinType::kNumJoinTypes); ++i) {
+    auto joinType = static_cast<core::JoinType>(i);
+    if (isTargetSupported(joinType) && isReferenceSupported(joinType)) {
+      supportedTypes.push_back(joinType);
+    }
+  }
+  return supportedTypes;
+}
+
+std::string JoinFuzzerBase::joinTypeName(core::JoinType joinType) const {
```
karthikeyann: `JoinTypeName::toName(joinType)` already exists via `VELOX_DECLARE_ENUM_NAME(JoinType)` in `velox/core/PlanNode.h`, and it covers `kCountingLeftSemiFilter`/`kCountingAnti` which this switch returns `"UNKNOWN"` for. Could the virtual + switch be replaced with a direct `JoinTypeName::toName(joinType)` call?

velox/exec/fuzzer/JoinFuzzerBase.cpp
```diff
@@ -0,0 +1,353 @@
...
+          exec::MemoryReclaimer::create())),
+      vectorFuzzer_{getFuzzerOptions(), pool_.get()},
+      referenceQueryRunner_{std::move(referenceQueryRunner)},
+      stats_{std::make_unique<Stats>()} {
+  filesystems::registerLocalFileSystem();
+
+  std::unordered_map<std::string, std::string> hiveConfig = {
+      {connector::hive::HiveConfig::kNumCacheFileHandles, "1000"}};
+
+  connector::hive::HiveConnectorFactory factory;
```
karthikeyann: `CudfJoinFuzzer::verify()` only uses `PlanBuilder().values(...)` — never table scans — but the base ctor unconditionally registers `HiveConnector`. The cuDF subclass now takes a hive-connector dependency it doesn't exercise. Could hive registration move to `JoinFuzzer`'s ctor (CPU-only), or sit behind a virtual hook the cuDF subclass can opt out of?

velox/exec/fuzzer/JoinFuzzerBase.cpp
```diff
@@ -0,0 +1,353 @@
...
+    std::chrono::duration<double> elapsed =
+        std::chrono::system_clock::now() - startTime;
+    return elapsed.count() >= FLAGS_duration_sec;
+  }
+  return i >= FLAGS_steps;
+}
+
+} // namespace
+
+std::string JoinFuzzerBase::makePercentageString(size_t value, size_t total) {
```
karthikeyann: nitpick: per `CODING_STYLE.md` (free functions in anonymous namespace > static class methods), and matching the pre-refactor pattern, this could live in the anonymous namespace in `JoinFuzzerBase.cpp` instead of as a `static` class member.

shrshi:
Closing this PR in favour of #18230

PR #16430, opened 2026-02-17
perf: Presto's ArbitraryAggregate could store the base Vector in place of the encoded Vector in clustered mode

kevinwilfong:
Summary:
Currently, when ArbitraryAggregate is used in clustered mode (used in streaming 
aggregations) for each group it stores the Vector the first non-null value came from and its
position. As part of doing this it decodes the Vector to see if the value is null.

When we produce the output from ArbitraryAggregate we either copy the values for each 
group, wrap the Vector in a dictionary if all results in a batch come from the same Vector, or
return a slice of the Vector directly if the results in a batch appear sequentially in the same
Vector. This means we either need to decode the Vector again to do the copy, or will likely
decode it at some point downstream when we further process the results.

Given we've already done the work of decoding it, we can store the base Vector in the
ArbitraryAggregate's state, along with the decoded index, so that we don't need to decode it
again. This saves us a little bit of additional work.

The down side is that if the results appear sequentially in the encoded Vector we may now
dictionary encode instead of taking a slice, but on the other hand if the results appear
sequentially in the base Vector we will now take a slice instead of dictionary encoding.

Differential Revision: D93531319

### DIFF
--- a/velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
+++ b/velox/functions/prestosql/aggregates/ArbitraryAggregate.cpp
@@ -290,7 +290,9 @@ class NonNumericArbitrary : public exec::Aggregate {
       const std::vector<VectorPtr>& args,
       const folly::Range<const vector_size_t*>& groupBoundaries) override {
     VELOX_CHECK(clusteredInput_);
-    decoded_.decode(*args[0]);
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@kevinwilfong thanks for the optimization!

PR #16431, opened 2026-02-18
fix(cudf): Fix stream synchronization issue with CudfTopN

mattgara:
* Add test that reproduces issue

### DIFF
--- a/velox/experimental/cudf/exec/CudfTopN.cpp
+++ b/velox/experimental/cudf/exec/CudfTopN.cpp
@@ -17,6 +17,7 @@
 #include "velox/experimental/cudf/CudfNoDefaults.h"
 #include "velox/experimental/cudf/exec/CudfTopN.h"
 #include "velox/experimental/cudf/exec/GpuResources.h"
+#include "velox/experimental/cudf/exec/Utilities.h"
 
 #include <cudf/detail/copy.hpp>
 #include <cudf/detail/gather.hpp>
--- a/velox/experimental/cudf/tests/TopNTest.cpp
+++ b/velox/experimental/cudf/tests/TopNTest.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "velox/experimental/cudf/CudfQueryConfig.h"
+#include "velox/experimental/cudf/CudfConfig.h"
 #include "velox/experimental/cudf/exec/CudfConversion.h"
 #include "velox/experimental/cudf/exec/ToCudf.h"
 
@@ -284,7 +284,7 @@ TEST_F(TopNTest, numericTopNSynchronization) {
   // sync.
   AssertQueryBuilder(plan, duckDbQueryRunner_)
       .config(cudf_velox::CudfFromVelox::kGpuBatchSizeRows, batchSize)
-      .config(cudf_velox::CudfQueryConfig::kCudfTopNBatchSize, 1)
+      .config(cudf_velox::CudfConfig::kCudfTopNBatchSize, 1)
       .assertResults(
           fmt::format(
               "SELECT c0, c1, c2 FROM tmp ORDER BY c1 DESC, c0 LIMIT {}",


velox/experimental/cudf/exec/CudfTopN.cpp
```diff
@@ -73,9 +74,19 @@ CudfVectorPtr CudfTopN::mergeTopK(
...
   for (const auto& batch : topNBatches) {
+    if (!batch) {
+      continue;
+    }
     tableViews.push_back(batch->getTableView());
+    inputStreams.push_back(batch->stream());
   }
+  // Ensure all upstream batch-producing streams are visible on the merge
+  // stream.
+  cudf::detail::join_streams(inputStreams, stream);
```
wence-, line 91: Also need to join the inout streams to stream in the opposite order after launching work on stream that reads them so they are not deallocated before that work completes

mattgara: Good point, added.

velox/experimental/cudf/exec/CudfTopN.cpp
```diff
@@ -73,11 +74,23 @@ CudfVectorPtr CudfTopN::mergeTopK(
...
   auto mergedTable =
       cudf::merge(tableViews, sortKeys_, columnOrder_, nullOrder_, stream, mr);
+  // Ensure input-stream deallocations don't race with merge reads.
+  joinStreamsBack(inputStreams, stream);
```
wence-: why not just cudf::detail::join_streams with the argument order reversed?

karthikeyann (commented):
This seems like repeatable pattern for all Barrier operators.
@mattgara can we do similar investigation for other barrier operators such as HashBuild, OrderBy, and HashAggregation, and ensure stream synchronization across all inputs is done similar to this PR?

@devavret Looking back into our discussion about using multiple streams within a same operator across batches; It increases intra-pipeline GPU concurrency. Could this cause hidden dependency across pipelines because we use a stream pool, and those streams gets reused?

velox/experimental/cudf/exec/CudfTopN.cpp
mattgara: Sure, I believe the resulting stream order is equivalent. I originally mirrored the event‑based pattern used [here](https://github.com/facebookincubator/velox/blob/main/velox/experimental/cudf/exec/CudfHashJoin.cpp#L619), but `cudf::detail::join_streams` with reversed args expresses the same dependency more cleanly. I’ll switch to the `join_streams` approach, thanks.

mattgara:
> @mattgara can we do similar investigation for other barrier operators such as HashBuild, OrderBy, and HashAggregation, and ensure stream synchronization across all inputs is done similar to this PR?

Makes sense, can you create an issue describing this, and I can look into it. Let's not block this PR on that work.

velox/experimental/cudf/exec/CudfTopN.cpp
```diff
@@ -73,11 +76,31 @@ CudfVectorPtr CudfTopN::mergeTopK(
...
+    std::for_each(
+        inputStreams.begin(),
+        inputStreams.end(),
+        [&stream](rmm::cuda_stream_view inputStream) {
+          if (inputStream.value() != stream.value()) {
+            cudf::detail::join_streams(
+                std::vector<rmm::cuda_stream_view>{stream}, inputStream);
+          }
+        });
+  }
```
karthikeyann: `  if (!inputStreams.empty()) {` is not required.
STL algorithms such as std::for_each automatically will avoid this zero size issue. So, it's safe to remove this If condition.

karthikeyann: Consider creating a new utility for this,
```
void streams_wait_for_stream(host_span<rmm::cuda_stream_view const> streams, rmm::cuda_stream_view stream)
{
  auto const event = event_for_thread();
  CUDF_CUDA_TRY(cudaEventRecord(event, stream));
  std::for_each(streams.begin(), streams.end(), [&](auto& strm) {
    CUDF_CUDA_TRY(cudaStreamWaitEvent(strm, event, 0));
  });
}
```
@bdice do we have similar utility in libcudf? (I could not find one)

karthikeyann (changes requested):
Suggesting to create new cuda stream utility.

velox/experimental/cudf/exec/CudfTopN.cpp
karthikeyann: `inputStream.value() != stream.value()` is this check required?
It could be effectively no-op. @bdice

bdice: We have `join_streams` here:

https://github.com/rapidsai/cudf/blob/c0c03305c03a6feb8c6458485747167a4664aaaa/cpp/src/utilities/stream_pool.cpp#L185-L192

And its dual, `fork_streams`:

https://github.com/rapidsai/cudf/blob/c0c03305c03a6feb8c6458485747167a4664aaaa/cpp/src/utilities/stream_pool.cpp#L174-L183

Those are the implementations we should copy here (they're in `cudf::detail` so not public cuDF APIs).

bdice: @karthikeyann asked me about this. Sorry, I missed some context here.

Does this sound correct?

1. Input vectors have their own streams
2. We have an input `stream` for mergeTopK
3. We join_streams to ensure all the data is ready to read (`stream` waits on the input vectors’ streams)
4. We do work on `stream`, nothing else queues on the input vectors’ streams
5. We want to finish some work on `stream` before the input vectors go out of scope, **this only requires a `stream` sync**. We already know the input streams are not needed anymore after joining them to stream. What’s the need to join them again?

@wence- I apologize if I’ve overlooked something in your previous comment.

karthikeyann: @bdice Yes `stream` sync gets the job done, by waiting for work on stream to finish before going out of scope. But **it does not have not wait**. it needs to ensure that inputStreams don't proceed until queued up work on `stream` is done, and work on `stream` can continue async.

bdice: Aha. Then yes, we need something like `fork_streams` but rather than creating a set of streams that are synced to one stream, it has a set of input streams wait on one stream. So it's a reverse-join (many streams wait on one stream).

That would be a new function, just copy the `join_streams` implementation but invert the dependency relationship.

karthikeyann: I am unblocking this PR. We can refactor this after this utility is created.

mattgara: I'll refactor to match your suggestion for now, we can replace with utility once created in cudf.

mhaseeb123 (approved):

velox/experimental/cudf/exec/Utilities.cpp
```diff
@@ -287,9 +287,19 @@ std::vector<std::unique_ptr<cudf::table>> getConcatenatedTableBatched(
   return outputTables;
 }
 
+void streamsWaitForStream(
+    const std::vector<rmm::cuda_stream_view>& streams,
+    rmm::cuda_stream_view stream) {
+  CudaEvent event(cudaEventDisableTiming);
```
karthikeyann: @mattgara CudaEvent creation is costly operation. That's why in my PR https://github.com/facebookincubator/velox/pull/16542 I created one cudaevent per operator instance.
So, creating an event for this operator will amortize the cost. 

I would seek @bdice for best practice.
join_streams uses a CudaEvent per thread that is cached. does sharing the event across drivers a good idea because each driver might be executed by different thread each time ? Same event * could * be used by  (same thread but by different operator, and different streams).

velox/experimental/cudf/exec/Utilities.cpp
```diff
@@ -305,12 +315,12 @@ CudaEvent::CudaEvent(CudaEvent&& other) noexcept : event_(other.event_) {
...
 
 const CudaEvent& CudaEvent::recordFrom(rmm::cuda_stream_view stream) const {
-  cudaEventRecord(event_, stream.value());
+  CUDF_CUDA_TRY(cudaEventRecord(event_, stream.value()));
   return *this;
 }
 
 const CudaEvent& CudaEvent::waitOn(rmm::cuda_stream_view stream) const {
-  cudaStreamWaitEvent(stream.value(), event_, 0);
+  CUDF_CUDA_TRY(cudaStreamWaitEvent(stream.value(), event_, 0));
```
karthikeyann, line 204: 👍

velox/experimental/cudf/exec/Utilities.cpp
mattgara: Okay, updated to follow the pattern you highlighted here.

velox/experimental/cudf/exec/CudfTopN.cpp
karthikeyann: @wence- please re-review this PR again.

kKPulla:
@mattgara Can you please address the build failures in the CI signals so I can import and merge the PR? Thanks!

mattgara:
> @mattgara Can you please address the build failures in the CI signals so I can import and merge the PR? Thanks!

Sorry about that, should be fixed now, thanks!

PR #16435, opened 2026-02-18
feat: Add checked_sum aggregate for Spark ANSI mode overflow detection

n0r0shi:
## Summary

  - Add `checked_sum` aggregate for Spark ANSI mode (`spark.sql.ansi.enabled=true`)
  - **Integers**: reuse `SumAggregateBase` with `Overflow=false` (Presto's checked arithmetic
    via `checkedPlus()`), which throws on integer overflow.
  - **Decimals**: add `throwOnOverflow` template parameter to `DecimalSumAggregate` — allows
    intermediate overflow (tracked via overflow counter), throws only if the final result
    exceeds decimal precision, matching Spark's `CheckOverflowInSum` semantics

  ## Test plan

  - [x] `checkedDecimalSumOverflow` — two max-precision decimals overflow at final result
  - [x] `checkedDecimalSumAllNull` — all-null inputs return null (no false overflow)
  - [x] `checkedDecimalSumNormal` — normal decimal summation
  - [x] `checkedIntegerSumOverflow` — two INT64_MAX values throw on overflow
  - [x] `checkedIntegerSumNormal` — normal integer summation
  - [x] `checkedIntegerSumAllNull` — all-null inputs return null

Closing https://github.com/facebookincubator/velox/issues/16436

### DIFF
--- a/velox/functions/sparksql/aggregates/DecimalSumAggregate.h
+++ b/velox/functions/sparksql/aggregates/DecimalSumAggregate.h
@@ -25,7 +25,11 @@ namespace facebook::velox::functions::aggregate::sparksql {
 /// final output type of final aggregation.
 /// @tparam ResultPrecision The precision of the result type, used for checking
 /// overflow.
-template <typename TInputType, typename TSumType, uint8_t ResultPrecision>
+template <
+    typename TInputType,
+    typename TSumType,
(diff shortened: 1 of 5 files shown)


philo-he:
@n0r0shi, I assume we still need this PR. Do we need to re-open it?

PR #16437, opened 2026-02-18
docs: Update Spark rand documentation

rui-mo:
Correct the Spark rand documentation so that the examples return the proper results.
Follow-up for: https://github.com/facebookincubator/velox/commit/8830b83f2f52b6828b3145be64003f42981fc22b


rui-mo:
In Gluten, discovered inconsistent results during the Velox rebase: https://github.com/apache/incubator-gluten/pull/11624.

PR #16439, opened 2026-02-18
feat: Add checked_avg aggregate for Spark ANSI mode decimal overflow

### LINKED ISSUE #16438: Add checked_avg aggregate for Spark ANSI mode decimal overflow detection
## Summary

  Spark's ANSI mode (`spark.sql.ansi.enabled=true`) requires `AVG` to throw an error on
  decimal overflow instead of returning null. Velox currently only has `avg` with Spark's
  default non-ANSI behavior (returns null on overflow).

  This only applies to decimal inputs — integer/float inputs return `DOUBLE`, which
  cannot overflow (IEEE 754 goes to `Infinity`).

  ## Expected Behavior

  - `checked_avg` on decimal inputs throws when the final result overflows the target
    decimal precision, instead of returning null
  - Intermediate overflow is allowed (tracked via overflow counter in the int128 accumulator)

  ## Proposed Approach

  - Add `throwOnOverflow` template parameter to existing `DecimalAverageAggregate`
  - Register as `checked_avg` (decimal signatures only) alongside existing `avg`
  - Same pattern as `checked_sum` (#16436)

n0r0shi:
## Summary

  - Add `checked_avg` aggregate for Spark ANSI mode (`spark.sql.ansi.enabled=true`)
  - Decimal-only — integer/float AVG returns `DOUBLE` which cannot overflow
  - Add `throwOnOverflow` template parameter to existing `DecimalAverageAggregate`
  - On decimal overflow: throws instead of returning null
  - Same pattern as `checked_sum` (#16435)

  Closes #16438

  ## Test plan

  - [x] `checkedDecimalAvgOverflow` — 10 max-precision decimals overflow at final result
  - [x] `checkedDecimalAvgNormal` — normal decimal average computation

### DIFF
--- a/velox/functions/sparksql/aggregates/AverageAggregate.cpp
+++ b/velox/functions/sparksql/aggregates/AverageAggregate.cpp
@@ -85,6 +85,7 @@ TypePtr getAvgResultType(const TypePtr& rawInputType) {
       std::min<uint8_t>(LongDecimalType::kMaxPrecision, s + 4));
 }
 
+template <bool throwOnOverflow = false>
 std::unique_ptr<exec::Aggregate> getDecimalAverageAggregate(
     core::AggregationNode::Step step,
     const std::vector<TypePtr>& argTypes,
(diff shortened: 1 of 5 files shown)


rui-mo (commented):
@n0r0shi I don’t think we need to add more checked functions for the same reason mentioned here: https://github.com/facebookincubator/velox/pull/16403#discussion_r2812221809. Pease take a look.

n0r0shi:
Superseded by #17937 after rebase + config-based refactor per review feedback.

PR #16443, opened 2026-02-18
fix: Fix HashJoinTest.taskWaitTimeout flakness

duxiao1212:
Summary:
The test may fail when multiple HashBuild and HashProbe operators are in non-reclaimable state

```
W0218 07:23:24.665766 341086 HashBuild.cpp:1310] Can't reclaim from hash build operator, state_[RUNNING], nonReclaimableSection_[1], op.4.1.0.HashBuild, usage: 22.50MB
W0218 07:23:24.665939 341086 HashProbe.cpp:1894] Can't reclaim from hash probe operator, state_[WAIT_FOR_BUILD], nonReclaimableSection_[0], inputSpiller_[nullptr], table_[nullptr], table_ numDistinct[nullptr], op.4.0.3.HashProbe, usage: 0B, node pool reservation: 128.00MB
```

Eventually, the process fails with the following error:

```
terminate called after throwing an instance of 'facebook::velox::VeloxRuntimeError'
what(): Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: MEM_CAP_EXCEEDED
```

Differential Revision: D93619111

### DIFF
--- a/velox/exec/tests/HashJoinTest.cpp
+++ b/velox/exec/tests/HashJoinTest.cpp
@@ -7340,12 +7340,25 @@ DEBUG_ONLY_TEST_P(HashJoinTest, taskWaitTimeout) {
                 expectedResult),
             "Memory reclaim failed to wait");
       } else {
-        // We expect succeed on large time out or no timeout.
-        const auto result = runHashJoinTask(
-            vectors, queryCtx, false, numDrivers, pool(), true, expectedResult);
-        auto taskStats = exec::toPlanStats(result.task->taskStats());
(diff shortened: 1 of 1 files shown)


duxiao1212:
> @duxiao1212 thanks for the fix. Is it possible to avoid the memory allocation failure by designing the test properly? Thanks!

Thanks for the feedback. I think a proper fix would be non-trivial give the tight memory limits and 4 drivers with parallel build (.Also, given that the frequency is quite low (about 10 out of 100), I think the existing solution is sufficient.

PR #16444, opened 2026-02-18
feat(cuDF): Add CudfPlanNodeChecker

pramodsatya:
## Overview
Adds a `CudfPlanNodeChecker` to enable early validation of cuDF execution eligibility at the Presto coordinator level.

## Changes

- **Extract Plan Node Validations**: Centralizes per-plan-node `core::PlanNode` eligibility checks from `OperatorAdapters` into a standalone `CudfPlanNodeChecker`, which the operator adapters now delegate to.
- **Coordinator-callable API**: Exposes one `isXxxNodeSupported` function per supported plan node, taking a plain `core::PlanNode` (with optional `QueryCtx`/`MemoryPool`) so plans can be validated without an execution context.

## Motivation

This change enables fail-fast validation of Velox plan nodes at the Presto coordinator **before** query execution on workers, reducing latency and improving user experience. It allows Presto to make informed routing decisions for cuDF-compatible queries.

- Enables https://github.com/prestodb/presto/pull/27273

## Benefits
- Fail-fast validation at the coordinator level
- Avoids duplicating eligibility logic between operator adapters and the coordinator
- Supports pre-execution plan analysis and routing decisions


pramodsatya:
@majetideepak, could you please review this change?

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
```diff
@@ -0,0 +1,264 @@
...
+#include "velox/expression/Expr.h"
+#include "velox/expression/FunctionSignature.h"
+#include "velox/expression/SignatureBinder.h"
+#include "velox/type/TypeCoercer.h"
+
+namespace facebook::velox::cudf_velox {
+
+namespace {
+
+bool canBeEvaluatedByCudfImpl(const core::TypedExprPtr& expr, bool deep) {
```
devavret: We used to have this previously https://github.com/facebookincubator/velox/pull/15145/changes#diff-7f0825f174264a1018d43ddb6a30ae61547e3fa6cf35426750217553aee5bcf3R158 but we replaced it with a pure exec::Expr based approach in https://github.com/facebookincubator/velox/pull/15478/changes#diff-a9e0889b9b497374bbf5760e21709dd4eeae0b7a161902ea938086483c262061R101-R123 because "compiling" TypedExpr to exec::Expr may change the expression. (add casts, remove common subexpressions, fold constants) and this may change what the plan node checker sees and the operator has to work with. Can this function here have a similar approach where it compiles to exec::Expr first and then checks?

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    if (aggregate.distinct) {
+      return false;
+    }
+
+    if (aggregate.mask) {
+      return false;
+    }
+
```
devavret: please retain comments from original code

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  auto stepIt = funcIt->second.find(step);
+  if (stepIt == funcIt->second.end()) {
+    return false;
+  }
+
+  // Validate against step-specific signatures from registry
+  return matchTypedCallAgainstSignatures(call, stepIt->second);
+}
```
devavret: I just saw that the origin of this doesn't use the raw input types so it's been dropped from here. @mattgara Can you confirm this was intentional? I see from the docstring that the type values may be different between call.inputs() and aggregation.rawInputTypes https://github.com/facebookincubator/velox/blob/8853645932621aeadb7fffdd9112b881ce4225f8/velox/core/PlanNode.h#L1100-L1107

devavret (changes requested):
In the companion Presto PR, I see that it throws whenever it encounters plan nodes which have a cudf implementation but aren't fully supported (e.g. aggregation node not supported because of missing gpu implementation of aggregations). Currently, we let the whole operator fallback to CPU but the Task is allowed to run. Is that not desired anymore? Do you want a task to fully fallback to CPU if it's partially unsupported?

Requesting changes for one clarification from @pramodsatya and one from @mattgara.

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
pramodsatya: Thanks, I was not aware this implementation existed before. On the Presto side during Presto->Velox plan conversion, the Presto->Velox expression [conversion](https://github.com/prestodb/presto/blob/8a4fae0f8cc96ce30cd27075ebc604ebe855eaf3/presto-native-execution/presto_cpp/main/types/PrestoToVeloxExpr.h#L24) returns a `core::TypedExpr`; and expression compilation happens later in Velox. This is reasonable since Velox only exposes the `ITypedExpr` API to engines like Presto and `exec::Expr` is an internal Velox representation. So if we want to validate the project expressions in `core::ProjectNode` at Presto plan checker level with sidecar, we need to validate the `ITypedExpr`s.

Could you please share what changes to the expression during compilation are required by `cuDF` operators? Could these changes be introduced at the `ITypedExpr` level using expression [rewrites](https://github.com/facebookincubator/velox/blob/48d371d8ab0541ce5613a1f46583e102671c0049/velox/expression/ExprRewriteRegistry.h#L43)?

pramodsatya:
> In the companion Presto PR, I see that it throws whenever it encounters plan nodes which have a cudf implementation but aren't fully supported (e.g. aggregation node not supported because of missing gpu implementation of aggregations). Currently, we let the whole operator fallback to CPU but the Task is allowed to run. Is that not desired anymore? Do you want a task to fully fallback to CPU if it's partially unsupported?
> 
> Requesting changes for one clarification from @pramodsatya and one from @mattgara.

Thanks for the feedback @devavret, the companion Presto PR is incomplete and needs more changes, I will share the new Presto PR for review by EOD. The Velox side changes will however remain the same so I opened this PR for review. 

With the native plan checker in Presto sidecar, we want to determine at the Presto layer whether the query can be executed with CPU or GPU operators in its entirety (and then route it to the right node group), without having to fallback to CPU if it's partially unsupported. However we do not want to impact the existing operator level fallback to CPU, this will still be configurable with `cudf.allow_cpu_fallback`. Is that alright? 

Update: Presto PR for cuDF plan checker integration with sidecar: https://github.com/prestodb/presto/pull/27273 (WIP and will open for review after integration testing)

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
mattgara: > @mattgara Can you confirm this was intentional? I see from the docstring that the type values may be different between call.inputs() and aggregation.rawInputTypes

Yes, removing `rawInputTypes` here is fine and actually a cleanup I’ve been hoping for.

The raw inputs and inputs do differ as the comments says; however, we do want to use the inputs in this function. In this validation path we should use the current step’s input types (`call.inputs()`) to correctly handle `kFinal`/`kIntermediate`. 

Sorry for the confusion, the parameter was likely leftover from a previous merge/rebase that I failed to notice.

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
pramodsatya: @devavret, my apologies I had missed context from the commit you shared previously. The intent with TypedExpr check here was to avoid the overhead added by compilation and keep cuDF evaluation eligibility checks lightweight. `queryCtx` is used in `canBeEvaluatedByCudf` checks, however it appears to be passed along as a parameter only for compiling TypedExpr to an exec::Expr (please correct me if I am wrong), and not to retrieve a specific `QueryConfig` that determines cuDF evaluation eligibility. 

The optimizations to expr tree during compilation such as constant folding and cast optimization, which used to previously happen with exec::Expr, now happens at the TypedExpr layer using the ExprOptimizer, so it can be carried over to this check as well without having to compile the TypedExpr. Avoiding expression compilation will keep the cost of expression validation, and in turn plan node validation, lightweight. Common subexpressions are detected during compilation but removed during expression evaluation so it should not introduce an inconsistency. 

Could you please share an example of an expression that can introduce inconsistencies between what the plan node checker sees with this new check and what the operator gets? I would like to investigate such potential inconsistencies and harden this check further.

@majetideepak, could you please share your thoughts on this?

(Linking intended usage in Presto: https://github.com/prestodb/presto/pull/27273)

majetideepak:
@devavret do you have any thoughts on using TypedExpr for plan validation? This is the public API and it makes sense to use this for validation.

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
devavret: Sorry for the late reply. 

>The optimizations to expr tree during compilation such as constant folding and cast optimization, which used to previously happen with exec::Expr, now happens at the TypedExpr layer using the ExprOptimizer

I didn't know about this. If this is true then we can avoid ever having to compile the expression even during cudf evaluation. We should then switch over out expression translation to working on optimized TypedExpr. 

>Common subexpressions are detected during compilation but removed during expression evaluation so it should not introduce an inconsistency

We don't make use of this detection but were hoping to some day. I suppose this means that we'll have a harder time implementing our own CSE.

>Could you please share an example of an expression that can introduce inconsistencies between what the plan node checker sees with this new check and what the operator gets?

If we switch over to translating TypedExpr then there shouldn't be any inconsistencies. It just needs someone to make the change and then run benchmarks to look for regressions. That work would have to block this PR but after that, we shouldn't need any change here in this PR

majetideepak: >We should then switch over out expression translation to working on optimized TypedExpr.

@devavret I doubt this is possible since the expression translation needs more fields that are only available in Expr.

devavret: @majetideepak Can you please elaborate on the missing fields?

velox/experimental/cudf/plan/CudfPlanNodeChecker.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+// TODO: Fix tests for this helper function so it is not exposed outside of the
+//   aggregation node plan checker.
+/// Step-aware aggregation validation function
+bool canAggregationBeEvaluatedByCudf(
+    const core::CallTypedExpr& call,
+    core::AggregationNode::Step step) {
+  // Check against step-aware aggregation registry
+  auto& stepAwareRegistry = getStepAwareAggregationRegistry();
+  auto funcIt = stepAwareRegistry.find(call.name());
```
mattgara, line 179: I believe this drops planner-validation support for Spark-style companion aggregate names. [The implementation on `main`](https://github.com/facebookincubator/velox/blob/18fe4451e2c892adc4373986ecdb4bc6c32caa91/velox/experimental/cudf/exec/CudfHashAggregation.cpp#L1978-L1994) first checks and normalizes the `originalName` then derives the `companionStep`, while this implementation omits this two-step transform.

We have separated Spark and Presto test suites into their own homes, could we add a Spark-side regression that constructs a `CallTypedExpr` named `sum_partial` and asserts `canAggregationBeEvaluatedByCudf(call, kSingle)`, with corresponding `_merge` and `_merge_extract` cases? [The existing mixed-step companion test on `main`](https://github.com/facebookincubator/velox/blob/18fe4451e2c892adc4373986ecdb4bc6c32caa91/velox/experimental/cudf/tests/AggregationTest.cpp#L537-L562) is an end-to-end example.

velox/experimental/cudf/plan/CudfPlanNodeChecker.h
```diff
@@ -0,0 +1,47 @@
...
+// Step-aware aggregation validation function
+bool canAggregationBeEvaluatedByCudf(
+    const core::CallTypedExpr& call,
+    core::AggregationNode::Step step);
+
+// AggregationNode validation: check if aggregation can be evaluated by CUDF
+bool isAggregationNodeSupported(const core::AggregationNode* aggregationNode);
+
+// HashJoinNode validation: check if join can be evaluated by CUDF
+bool isHashJoinNodeSupported(const core::HashJoinNode* joinNode);
```
mattgara, line 45: This appears to allow a plan to pass coordinator validation only for cuDF execution to reject it later.

The code path I'm thinking of would have this pattern.

```
isValidPlanNode(LocalPartitionNode*) -> true
```
but
```
LocalPartitionAdapter::canRunOnGPU()
  -> CudfLocalPartition::shouldReplace() -> false.
```

That is, an unsupported `LocalPartitionNode` is accepted by the checker but rejected during execution.  If this is an issue, please add this as a regression test.

velox/experimental/cudf/plan/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+target_link_libraries(
+  velox_cudf_plan
+  PRIVATE
+    velox_cudf_hive_connector
+    velox_core
+    velox_exception
+    velox_expression
+    velox_cudf_expression_registry
+    velox_cudf_exec
```
mattgara, line 25: This seems to be introducing a cycle in the dependency graph. Here we make `velox_cudf_plan` dependent on `velox_cudf_exec`, while `velox/experimental/cudf/exec/CMakeLists.txt:47` introduces the reverse dependency.

Can we please separate this into a cycle-free dependency graph by splitting out functionality or by other means?

mattgara (commented):
LGTM overall, found three issues potentially looking into.

pramodsatya:
@mattgara, thanks for the feedback. This PR is currently blocked on the cudf-velox expression subsystem refactor in https://github.com/facebookincubator/velox/pull/17108, would really appreciate your feedback on this PR as well. 

I will rebase this PR and address your feedback once #17108 lands.

karthikeyann:
@pramodsatya Thank you for making this to make coordinator scheduling better. 
is this PR still required? The dependent PR you mentioned is taken care of.
Would you like to refresh this PR?

velox/experimental/cudf/tests/CudfPlanNodeCheckerTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  // variance has no cuDF aggregator (unlike stddev, which aliases the
+  // supported stddev_samp), so the node must fall back to CPU.
+  auto plan = PlanBuilder()
+                  .values({inputVector_})
+                  .singleAggregation({"c0"}, {"variance(c2)"})
+                  .planNode();
+  auto aggregationNode =
+      std::dynamic_pointer_cast<const core::AggregationNode>(plan);
+  ASSERT_NE(aggregationNode, nullptr);
+  EXPECT_FALSE(isAggregationNodeSupported(aggregationNode.get()));
```
pramodsatya: more unsupported tests are needed

velox/experimental/cudf/tests/CudfPlanNodeCheckerTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  void SetUp() override {
+    pool_ = memory::memoryManager()->addLeafPool();
+    queryCtx_ = core::QueryCtx::create();
+    functions::prestosql::registerAllScalarFunctions();
+    aggregate::prestosql::registerAllAggregateFunctions();
+    cudf_velox::CudfConfig::getInstance().allowCpuFallback = false;
+    cudf_velox::registerCudf();
+    parse::registerTypeResolver();
+
+    inputVector_ = makeRowVector(
```
pramodsatya: against convention? arent input vectors constructed in tests?

velox/experimental/cudf/exec/OperatorAdapters.cpp
```diff
@@ -194,39 +180,34 @@ class FilterProjectAdapter : public OperatorAdapter {
...
-              "FilterProject empty input type with filter or projections, PlanNode id: {}",
-              planNode->id());
-          return false;
-        }
-      }
+    // A fused FilterProject over an empty input type can only stay on GPU when
+    // it neither filters nor projects. The per-node checks below do not see the
+    // fused pairing, so guard that combination here.
+    if (projectPlanNode &&
+        projectPlanNode->sources()[0]->outputType()->size() == 0 &&
```
pramodsatya: isProjectNiodeSupported already covers this case, use that helper and add an additional case for `filterNode && projectPlanNode->sources()[0]->outputType()->size() == 0`, with a comment as to why that is needed

velox/experimental/cudf/tests/CudfPlanNodeCheckerTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+                  .values({inputVector_})
+                  .singleAggregation({"c0"}, {"variance(c2)"})
+                  .planNode();
+  auto aggregationNode =
+      std::dynamic_pointer_cast<const core::AggregationNode>(plan);
+  ASSERT_NE(aggregationNode, nullptr);
+  EXPECT_FALSE(isAggregationNodeSupported(aggregationNode.get()));
+}
+
+TEST_F(CudfPlanNodeCheckerTest, aggregationNodeWithQueryContext) {
```
pramodsatya: test adds no value as no queryCtx based decisions are demonstrated here, can an aggregate function whose behavior or eligibility changes based on queryCtx be used here?
test coverage is lacking in general and needs to be improved.

Issue #14446, opened 2025-08-12
Prestissimo decimal type is incorrect when inserting values in Parquet


jkhaliqi:
### Bug description

Prestissimo decimal type is incorrect when inserting values in Parquet. Decimal precision, scale are written incorrectly to Parquet. Need to change presto velox parquet-tools inspect output on parquet tools created by with decimal to be similar to presto java.

To Reproduce
```
create table decimaltest(col0 DECIMAL(10, 2));
CREATE TABLE
insert into decimaltest values(1200.00);
INSERT: 1 row

parquet meta file.parquet
```
Before the PR the output for prestissimo would be showing the `col0` type as `DECIMAL(6,2)` instead of `DECIMAL(10,2)`.
```
Created by: parquet-mr version 2.6.0 (build parquet-cpp-velox)
Properties: (none)
Schema:
message schema {
  optional int32 col0 (DECIMAL(6,2));
}
```
Expected behavior
Java writer does not have this issue. It uses the right precision, scale. AND with this PR we are getting the same output for the DECIMAL type and scale.

```
Created by: parquet-mr version 1.13.1 (build db4183109d5b734ec5930d870cdae161e408ddba)
Properties: (none)
Schema:
message hive_schema {
  optional fixed_len_byte_array(5) col0 (DECIMAL(10,2));
}
```
This issue is caused by the way casting is done for the Decimal type. Based on the velox plan for a smaller number we see no cast, which causes problems since the value being inserted is not being casted to the table type which can be bigger. For example 
```
create table col(col0 DECIMAL(10, 2)); 
insert into col values(12.23);
```
the plan shows
```
presto:tpch> explain (type distributed) insert into test values(12.23);
                                                                                                               Query Plan                                                                                                               
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Fragment 0 [COORDINATOR_ONLY]                                                                                                                                                                                                          
     Output layout: [rows_1]                                                                                                                                                                                                            
     Output partitioning: SINGLE []                                                                                                                                                                                                     
     Output encoding: COLUMNAR                                                                                                                                                                                                          
     Stage Execution Strategy: UNGROUPED_EXECUTION                                                                                                                                                                                      
     - Output[PlanNodeId 6][rows] => [rows_1:bigint]                                                                                                                                                                                    
             rows := rows_1                                                                                                                                                                                                             
[...]

PR #16452, opened 2026-02-18
feat: Add equivalent check for decimal type

jkhaliqi:
Update Decimal type validation to allow precision values that are greater than or equal to the required precision, as long as the scale remains the same.

### DIFF
--- a/velox/type/Type.h
+++ b/velox/type/Type.h
@@ -828,13 +828,23 @@ class DecimalType : public ScalarType<KIND> {
   static constexpr uint8_t kMaxPrecision = KIND == TypeKind::BIGINT ? 18 : 38;
   static constexpr uint8_t kMinPrecision = KIND == TypeKind::BIGINT ? 1 : 19;
 
-  inline bool equivalent(const Type& other) const override {
+  inline bool equals(const Type& other) const override {
     if (!Type::hasSameTypeId(other)) {
       return false;
     }
     const auto& otherDecimal = static_cast<const DecimalType<KIND>&>(other);
     return (
         otherDecimal.precision() == precision() &&
+        otherDecimal.scale() == scale());
+  }
+
+  inline bool equivalent(const Type& other) const override {
+    if (!Type::hasSameTypeId(other)) {
+      return false;
+    }
+    const auto& otherDecimal = static_cast<const DecimalType<KIND>&>(other);
+    return (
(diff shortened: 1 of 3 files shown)


velox/type/Type.h
```diff
@@ -838,6 +838,16 @@ class DecimalType : public ScalarType<KIND> {
         otherDecimal.scale() == scale());
   }
 
+  inline bool equivalent(const Type& other) const override {
+    if (!Type::hasSameTypeId(other)) {
+      return false;
+    }
+    const auto& otherDecimal = static_cast<const DecimalType<KIND>&>(other);
+    return (
+        otherDecimal.precision() >= precision() &&
```
Yuhta, line 847: This is not well-defined, `x.equivalent(y)` should always be the same as `y.equivalent(x)`.  What problem are you trying to resolve?

Yuhta (changes requested):

velox/type/Type.h
```diff
@@ -838,6 +838,16 @@ class DecimalType : public ScalarType<KIND> {
         otherDecimal.scale() == scale());
   }
 
+  inline bool equivalent(const Type& other) const override {
```
PingLiuPing, line 841: Would you add document to explain the difference with `equals`, and why such API is required.

velox/type/Type.h
jkhaliqi: The problem stems from: https://github.com/facebookincubator/velox/issues/14446. The presto planner does not do a cast for ShortDecimal, but was casting LongDecimal. Now we have a presto PR https://github.com/prestodb/presto/pull/26737/changes that uses the original table values. With this if we have a table value `decimal 10,2` but insert a `decimal 4,2` it will fail in velox check https://github.com/facebookincubator/velox/blob/b6cce020dcccff65cabb6f6167324bb743b6bf05/velox/dwio/parquet/writer/Writer.cpp#L487-L489. This should not be the case since smaller decimal types should be able to fit inside of larger types for example a `decimal (4,2)` can fit inside of a `decimal(10,2)`, but the reverse should not be allowed. Should the name be different from equivalent and possibly something like `decimalFitsInto` instead?

Yuhta: The check there is correct, there is a missing cast in your plan to cast `decimal(4,2)` into `decimal(10,2)`.  Writer should not do the cast implicitly.

PR #16454, opened 2026-02-19
fix(cudf): Replace default memory resource and stream with explicit output MR and data stream in Cudf operators

karthikeyann:
- Add a compile-time guard header that triggers errors when cudf APIs fall back to defaulted stream and mr arguments, catching unintended usage at build time (GNU compilers only).
- Introduce `cudf_velox::get_output_mr()` memory resource for allocating output vectors of Cudf operators, enabling memory tracking and passive spill of outputs via managed memory.
- Update Cudf Operators to use explicit data stream and get_output_mr() / temp MR instead of defaults. 
- Update connectors CudfHive* 
(expressions in follow up PR)
- Refactor Utilities.h/cpp to GpuResources.h/cpp (with mr and stream related methods), Now Utilities.h has only concatenation related methods.

Only place where default stream is explicitly used is `CudfFromVelox::close()`
`  cudf::get_default_stream(cudf::allow_default_stream).synchronize();`

This allows partial memory tracking, and passive selective spill support of output vectors. 
This allows removal of all default stream calls which are hard to track.

### DIFF
--- a/velox/core/QueryConfig.h
+++ b/velox/core/QueryConfig.h
@@ -223,6 +223,14 @@ class QueryConfig {
   static constexpr const char* kAggregationCompactionUnusedMemoryRatio =
       "aggregation_compaction_unused_memory_ratio";
 
+  /// If true, enables lightweight memory compaction before spilling during
+  /// memory reclaim in aggregation. When enabled, the aggregation operator
+  /// will try to compact aggregate function state (e.g., free dead strings)
+  /// before resorting to spilling.
+  /// Disabled by default.
+  static constexpr const char* kAggregationMemoryCompactionReclaimEnabled =
+      "aggregation_memory_compaction_reclaim_enabled";
+
   static constexpr const char* kAbandonPartialTopNRowNumberMinRows =
       "abandon_partial_topn_row_number_min_rows";
 
@@ -927,6 +935,10 @@ class QueryConfig {
 
   double aggregationCompactionUnusedMemoryRatio() const {
     return get<double>(kAggregationCompactionUnusedMemoryRatio, 0.25);
+  }
+
+  bool aggregationMemoryCompactionReclaimEnabled() const {
+    return get<bool>(kAggregationMemoryCompactionReclaimEnabled, false);
   }
 
   int32_t abandonPartialTopNRowNumberMinRows() const {
--- a/velox/docs/configs.rst
+++ b/velox/docs/configs.rst
@@ -546,6 +546,13 @@ Aggregation
        The value is in the range of [0, 1). Currently only applies to approx_most_frequent
        aggregate with StringView type during global aggregation. May be extended
        to other aggregation types on-demand.
+   * - aggregation_memory_compaction_reclaim_enabled
+     - bool
+     - false
+     - If true, enables lightweight memory compaction before spilling during
+       memory reclaim in aggregation. When enabled, the aggregation operator
+       will try to compact aggregate function state (e.g., free dead strings)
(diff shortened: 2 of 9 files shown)


karthikeyann:
@shrshi please review this PR

velox/experimental/cudf/exec/GpuResources.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+} // namespace facebook::velox::cudf_velox
+
+// This must NOT be in a file that includes CudfNoDefaults.h, because
+// CudfNoDefaults.h redeclares cudf::get_default_stream() with
+// __attribute__((error)). The overload below calls the real function.
+namespace cudf {
+
+struct allow_default_stream_t {};
```
shrshi: Nit: The `allow_default_stream_t` struct and the `allow_default_stream` constant are defined in both this file and `CudfNoDefaults.h`. Can we extract these definitions into a shared header? It would safeguard against one definition changing without the other

velox/experimental/cudf/exec/CudfHashJoin.cpp
```diff
@@ -631,10 +649,18 @@ std::unique_ptr<cudf::table> CudfHashJoinProbe::filteredOutput(
...
+      leftIndicesCol,
+      oobPolicy,
+      stream,
+      cudf_velox::get_output_mr());
+  auto rightResult = cudf::gather(
+      rightTableView,
+      rightIndicesCol,
+      oobPolicy,
+      stream,
+      cudf_velox::get_output_mr());
```
shrshi: Can these results be in temp mr since they haven't been filtered yet?

karthikeyann: I used `get_output_mr()` here with assumption that there might be optimization to move this table if the filter returned all true. Is that not the case anymore? That should be a good optimization that we could do to reduce memory pressure.

velox/experimental/cudf/exec/GpuResources.cpp
karthikeyann: updated.

devavret (approved):

velox/experimental/cudf/exec/CudfHashJoin.cpp
shrshi: We don't have this optimization. This function `filteredOutput` is only used in `rightJoin` right now, so we will only need to change how the filter is applied there https://github.com/facebookincubator/velox/blob/e26bf136f1b2b809d306d5a6f6c0dd94b4aac0fa/velox/experimental/cudf/exec/CudfHashJoin.cpp#L937

I'll open a PR for it tomorrow

karthikeyann:
@kgpai could this PR be imported?

kKPulla:
can you please rebase on top of https://github.com/facebookincubator/velox/pull/16533 once it lands for clean CI signals

karthikeyann:
@kKPulla rebased and checks completed.

mhaseeb123 (approved):

kKPulla:
Sorry to bother again. Seems like the previous fix did not fully work. Can you try rebasing again on top of main to include https://github.com/facebookincubator/velox/pull/16539

kKPulla:
@karthikeyann can you resolve conflicts above as it looks like one of the other PRs that I merged https://github.com/facebookincubator/velox/pull/16528 conflicted with this.

karthikeyann:
@kKPulla updated.

